Issue #447, opened 2018-06-25
ingressroute.spec.virtualhost.name: "" should be illegal


davecheney:
It should be illegal to submit an ingressroute spec with a blank fqdn
```
Spec: ingressroutev1.IngressRouteSpec{
                                        VirtualHost: &ingressroutev1.VirtualHost{
                                                Fqdn: "",
                                        },
```
^ Should be rejected. All virtualhosts must have a name.

stevesloka:
The CRD validation in #441 cannot validate since it's only required on a `root` IngressRoute. What should Contour do if this is found? Reject the IngressRoute and set a status on the object?

davecheney:
Yup, we'll reject it.

PR #482, opened 2018-06-29
Fqdn validation

stevesloka:
If any Root IngressRoute defined does not supply the `fqdn` value, then that object is invalid and should be skipped. Additionally, the status of the object is set to `invalid`. 

**(NOTE: This is currently blocked until #469 is merged.)**

Example:

```yaml
status:
  currentStatus: invalid
  description: Spec.VirtualHost.Fqdn must be specified!
```

// Fixes #447

PR #501, opened 2018-07-03
internal/dag: validate ingressroutes and produce a list that contains a status entry for each

### LINKED ISSUE #490: Validate IngressRoute resources and report status
When contour is processing IngressRoute resources, it needs to validate that the resource is valid.

The following is an initial list of validations that we should perform:
- [x] Negative weight provided in the route definition
- [x] Invalid port number provided for service
- [x] ~~Invalid Load Balancing strategy provided~~ (This is covered by the CRD validation)
- [x] ~~Route delegates to a non-existent namespace~~ (Delegating to a non-existent namespace or IR does not invalidate the parent)
- [x] Orphaned route
- [x] Prefix in Root does not match route in delegated route
- [x] Delegation chain produces a cycle
- [x] Root IngressRoute created in non-allowed namespace
- [x] A given Route both delegates and has a list of services

Some of these can potentially be done upfront before the resource even hits the DAG. Others will most likely have to be performed while building the DAG.

alexbrand:
Fixes #490 

The first step towards updating the status of IngressRoute (IR) resources is adding the validation that will determine the status.

Validation is performed while we are recomputing the DAG. There might be some validation that can be done before, but I think this will do for now. The `recompute` method of the DAG now also returns a list of statuses, each containing the IR in question, its status and a human readable message that explains the status.

Currently, the statuses are printed out to STDOUT instead of written up to the API server. This will be handled in a follow-up PR. 

This PR adds the following status checks for IRs:
- Negative weight provided in the route definition
- Invalid port number provided for service
- Prefix in parent does not match route in delegated route
- Root IR created in a namespace other than the root namespaces
- A given Route of an IR both delegates to another IR and has a list of services
- Orphaned route
- Delegation chain produces a cycle
- Root IR must specify fqdn

Specific check notes:

**Orphan detection**
When recomputing the DAG, initialize an empty map that keeps track of the non-root IngressRoutes that we have seen and mark them as orphaned. Whenever we follow an edge to a non-root IngressRoute, we mark that IngressRoute as not orphaned. Once we have finished processing all the IngressRoutes, we can determine which IngressRoutes are orphaned by iterating over the map.

**Cycle detection**
Before this PR, we were using a map to keep track of the IngressRoutes we had already visited. This worked, but prevented us from knowing the path we took to reach the IngressRoute that we are processing. Instead of using a map, switch to using a stack so that we can use to build the delegation path we have taken, and thus return it in the validation error message in the case of a cycle.

Additionally, the cycle check needs to be performed before we follow an edge that produces a cycle, so that we can mark the source IngressRoute as invalid, instead of the target.

### DIFF
--- a/internal/dag/dag.go
+++ b/internal/dag/dag.go
@@ -145,11 +145,11 @@ func (d *DAG) Recompute() {
 	d.mu.Lock()
 	defer d.mu.Unlock()
 	version := d.dag.version
-	var verrs []validationError
-	d.dag, verrs = d.recompute()
+	var status []ingressrouteStatus
+	d.dag, status = d.recompute()
 	d.dag.version = version + 1
-	for _, ve := range verrs {
-		fmt.Printf("%v", ve)
+	for _, s := range status {
+		fmt.Printf("%v\n", s)
 	}
 }
 
@@ -213,7 +213,7 @@ func (sm *serviceMap) insert(svc *v1.Service, port *v1.ServicePort) *Service {
 }
 
 // recompute builds a new *dag.dag.
-func (d *DAG) recompute() (dag, []validationError) {
+func (d *DAG) recompute() (dag, []ingressrouteStatus) {
 	sm := serviceMap{
 		services: d.services,
 	}
@@ -361,7 +361,7 @@ func (d *DAG) recompute() (dag, []validationError) {
 	}
 
 	// process ingressroute documents
-	var verrs []validationError
+	var status []ingressrouteStatus
 	orphaned := make(map[meta]bool)
 	for _, ir := range d.ingressroutes {
 		if ir.Spec.VirtualHost == nil {
@@ -374,7 +374,7 @@ func (d *DAG) recompute() (dag, []validationError) {
 
 		// ensure root ingressroute lives in allowed namespace
 		if !d.rootAllowed(ir) {
-			verrs = append(verrs, validationError{object: ir, msg: "root IngressRoute cannot be defined in this namespace"})
+			status = append(status, ingressrouteStatus{object: ir, status: "invalid", msg: "root IngressRoute cannot be defined in this namespace"})
 			continue
 		}
 
@@ -392,7 +392,7 @@ func (d *DAG) recompute() (dag, []validationError) {
 		var visited []*ingressroutev1.IngressRoute
 		prefixMatch := ""
 		ve := d.processIngressRoute(ir, prefixMatch, visited, host, service, vhost, orphaned)
-		verrs = append(verrs, ve...)
+		status = append(status, ve...)
 	}
 
 	var _d dag
@@ -407,28 +407,29 @@ func (d *DAG) recompute() (dag, []validationError) {
 		if orph {
 			ir, ok := d.ingressroutes[meta]
 			if ok {
-				verrs = append(verrs, validationError{object: ir, msg: "this IngressRoute is not part of a delegation chain from a root IngressRoute"})
+				status = append(status, ingressrouteStatus{object: ir, status: "orphaned", msg: "this IngressRoute is not part of a delegation chain from a root IngressRoute"})
 			}
 		}
 	}
-	return _d, verrs
+	return _d, status
 }
 
-func (d *DAG) processIngressRoute(ir *ingressroutev1.IngressRoute, prefixMatch string, visited []*ingressroutev1.IngressRoute, host string, service func(m meta, port intstr.IntOrString) *Service, vhost func(host string, port int) *VirtualHost, orphaned map[meta]bool) []validationError {
+func (d *DAG) processIngressRoute(ir *ingressroutev1.IngressRoute, prefixMatch string, visited []*ingressroutev1.IngressRoute, host string, service func(m meta, port intstr.IntOrString) *Service, vhost func(host string, port int) *VirtualHost, orphaned map[meta]bool) []ingressrouteStatus {
 	visited = append(visited, ir)
 	defer func() {
 		visited = visited[:len(visited)-1]
 	}()
 
+	var status []ingressrouteStatus
 	for _, route := range ir.Spec.Routes {
 		// route cannot both delegate and point to services
 		if len(route.Services) > 0 && route.Delegate.Name != "" {
-			return []validationError{{object: ir, msg: fmt.Sprintf("route %q: cannot specify services and delegate in the same route", route.Match)}}
+			return []ingressrouteStatus{{object: ir, status: "invalid", msg: fmt.Sprintf("route %q: cannot specify services and delegate in the same route", route.Match)}}
 		}
 		// base case: The route points to services, so we add them to the vhost
 		if len(route.Services) > 0 {
 			if !matchesPathPrefix(route.Match, prefixMatch) {
-				return []validationError{{object: ir, msg: "the path prefix does not match the parent's path prefix"}}
+				return []ingressrouteStatus{{object: ir, status: "invalid", msg: "the path prefix does not match the parent's path prefix"}}
 			}
 			r := &Route{
 				path:   route.Match,
@@ -436,10 +437,10 @@ func (d *DAG) processIngressRoute(ir *ingressroutev1.IngressRoute, prefixMatch s
 			}
 			for _, s := range route.Services {
 				if s.Port < 1 || s.Port > 65535 {
-					return []validationError{{object: ir, msg: fmt.Sprintf("route %q: service %q: port must be in the range 1-65535", route.Match, s.Name)}}
+					return []ingressrouteStatus{{object: ir, status: "invalid", msg: fmt.Sprintf("route %q: service %q: port must be in the range 1-65535", route.Match, s.Name)}}
 				}
 				if s.Weight != nil && *s.Weight < 0 {
-					return []validationError{{object: ir, msg: fmt.Sprintf("route %q: service %q: weight must be greater than zero", route.Match, s.Name)}}
+					return []ingressrouteStatus{{object: ir, status: "invalid", msg: fmt.Sprintf("route %q: service %q: weight must be greater than zero", route.Match, s.Name)}}
 				}
 				m := meta{name: s.Name, namespace: ir.Namespace}
 				if svc := service(m, intstr.FromInt(s.Port)); svc != nil {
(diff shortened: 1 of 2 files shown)


stevesloka:
What happens to the children of a parent IngressRoute that is invalid? Do they become orphaned or should they be marked as invalid because of a parent issue?

alexbrand:
I'll add a test case for that. The parent would be considered invalid, which would orphan all of the children assuming they are not reachable via other parents.

stevesloka:
Closing will be resolved in #482

stevesloka:
Fixes #447

alexbrand:
The following table outlines conditions, status and corresponding message.

| Condition | Status | Message |
|-----------|--------|----------|
| Valid IR | valid | valid IngressRoute |
| invalid port in service | invalid | route "/foo": service "home": port must be in the range 1-65535 |
| Root IR outside of root namespace | invalid | root IngressRoute cannot be defined in this namespace |
| Prefix does not match parent's prefix | invalid | the path prefix does not match the parent's path prefix |
| Invalid weight provided | invalid | route "/foo": service "home": weight must be greater than zero
| missing FQDN in root IR | invalid | Spec.VirtualHost.Fqdn must be specified | 
| route creates cycle in DAG | invalid | route creates a delegation cycle: roots/self -> roots/self
| route creates a cycle in DAG | invalid | route creates a delegation cycle: roots/parent -> roots/child -> roots/parent |
| route lists services and also delegates | invalid | route "/foo": cannot specify services and delegate in the same route |
| orphaned ingressroute | orphaned | this IngressRoute is not part of a delegation chain from a root IngressRoute

internal/dag/dag.go
```diff
@@ -351,18 +356,28 @@ func (d *DAG) recompute() dag {
 	}
 
 	// process ingressroute documents
+	var status []ingressrouteStatus
+	orphaned := make(map[meta]bool)
 	for _, ir := range d.ingressroutes {
 		if ir.Spec.VirtualHost == nil {
-			// delegate ingressroute, skip it
+			// delegate ingress route. mark as orphaned if we haven't reached it before.
```
davecheney: I don't think this needs a conditional. The logic that is here now says

_if not found, then mark as found_

which i think can be simplified into
```
orphaned[meta{name: ir.Name, namespace: ir.Namespace}] = true
```

internal/dag/dag.go
```diff
@@ -145,8 +146,12 @@ func (d *DAG) Recompute() {
 	d.mu.Lock()
 	defer d.mu.Unlock()
 	version := d.dag.version
-	d.dag = d.recompute()
+	var status []ingressrouteStatus
+	d.dag, status = d.recompute()
 	d.dag.version = version + 1
+	for _, s := range status {
+		fmt.Printf("%v\n", s)
```
davecheney: This is fine for a proof of concept but we can't land this into master with this fmt.Printf in here as it will generate large amounts of log spam.

internal/dag/dag.go
```diff
@@ -373,9 +388,10 @@ func (d *DAG) recompute() dag {
 			}
 		}
 
-		visited := make(map[meta]bool)
+		var visited []*ingressroutev1.IngressRoute
```
davecheney: this can be written as
```
ve := d.processIngressRoute(ir, prefixMatch, nil, host, service, vhost, orphaned)
```
because the value of `visited` before and after the call to processIngressRoute is unchanged.

internal/dag/dag.go
```diff
@@ -386,28 +402,45 @@ func (d *DAG) recompute() dag {
...
 
-func (d *DAG) processIngressRoute(ir *ingressroutev1.IngressRoute, prefixMatch string, visited map[meta]bool, host string, service func(m meta, port intstr.IntOrString) *Service, vhost func(host string, port int) *VirtualHost) {
-	// check if we have already visited this ingressroute. if we have, there is a cycle in the dag.
-	if visited[meta{name: ir.Name, namespace: ir.Namespace}] {
-		// TODO(abrand): Handle the cycle. Invalidate IngressRoute and set status?
-		return
-	}
+func (d *DAG) processIngressRoute(ir *ingressroutev1.IngressRoute, prefixMatch string, visited []*ingressroutev1.IngressRoute, host string, service func(m meta, port intstr.IntOrString) *Service, vhost func(host string, port int) *VirtualHost, orphaned map[meta]bool) []ingressrouteStatus {
+	visited = append(visited, ir)
+	defer func() {
```
davecheney: I don't think this code is doing what its supposed to, on exit from the function the callers value of visited will be unchanged. The memory beyond what the callers' copy of visited may contain `ir`, if there was sufficient space to append it without reallocating the slice, but the caller won't be able to see that.

internal/dag/dag.go
```diff
@@ -424,14 +457,30 @@ func (d *DAG) processIngressRoute(ir *ingressroutev1.IngressRoute, prefixMatch s
...
 			if ok {
+				// dest is not an orphaned route, as there is an IR that points to it
+				orphaned[meta{name: dest.Name, namespace: dest.Namespace}] = false
+
+				// ensure we are not following an edge that produces a cycle
+				var path []string
+				for _, vir := range visited {
+					path = append(path, fmt.Sprintf("%s/%s", vir.Namespace, vir.Name))
+				}
+				for _, vir := range visited {
```
davecheney: this makes me think that visited should probably be a map, not a slice. You're looking for in the set of visited elements to see if we've already passed this way before. The order is not important, only set membership. 

A map might be a better choice for visited and would make my comments above mute.

internal/dag/dag.go
alexbrand: The problem is that we can overwrite a `false` that is already there. For example, in a scenario where A delegates to B, assuming we process A before B, we would reach B from A and mark it as orphaned = false. Once we are done processing A, we move onto processing B, and we don't want to overwrite orphaned = false.

internal/dag/dag.go
alexbrand: We were using a map but switched to a slice so that we can build up the delegation path and include it in the error msg in case of a cycle. There's more context in the "Cycle Detection" section of the PR description.

internal/dag/dag.go
alexbrand: Ah nice - Thanks!

internal/dag/dag.go
alexbrand: Ah definitely. I see how this would break down. Thanks @davecheney.

Do you have any suggestions here? My first thought is to build a stack type that has push/pop methods that we can use here, but wondering if there's a simpler approach.

alexbrand:
@davecheney Addressed your comments. PTAL. 

I had to add a new struct to keep track of visited IngressRoutes and avoid the potential bug you raised. Interested in your thoughts on that approach.

alexbrand:
Also added validation to ensure that both load balancing strategy and health check are consistent across all reference to the same service inside a single IngressRoute.

internal/dag/dag.go
```diff
@@ -391,29 +403,69 @@ func (d *DAG) recompute() dag {
 		_d.roots = append(_d.roots, svh)
 	}
 
-	return _d
+	for meta, orph := range orphaned {
+		if orph {
+			ir, ok := d.ingressroutes[meta]
+			if ok {
```
stevesloka: What if this fails? Should we log out the error?

internal/dag/dag.go
```diff
@@ -391,29 +403,69 @@ func (d *DAG) recompute() dag {
...
+		visited.pop()
+	}()
+
+	// validate load balancing strategies are consistent across all references to the same service
+	lbstrat := make(map[string]string)
+	for _, route := range ir.Spec.Routes {
+		for _, svc := range route.Services {
+			s, ok := lbstrat[svc.Name]
+			if ok && svc.Strategy != s {
+				return []ingressrouteStatus{{object: ir, status: "invalid", msg: fmt.Sprintf("different load balancing strategies were specified for service %q", svc.Name)}}
```
stevesloka: nit: Should we include the namespace in the error so it's clear where the issue is?

internal/dag/dag.go
```diff
@@ -391,29 +403,69 @@ func (d *DAG) recompute() dag {
...
-	if visited[meta{name: ir.Name, namespace: ir.Namespace}] {
-		// TODO(abrand): Handle the cycle. Invalidate IngressRoute and set status?
-		return
+	// validate health check spec is consistent across all references to the same service
+	healthChecks := make(map[string]*ingressroutev1.HealthCheck)
+	for _, route := range ir.Spec.Routes {
+		for _, svc := range route.Services {
+			hc, ok := healthChecks[svc.Name]
+			if ok && svc.HealthCheck != hc {
+				return []ingressrouteStatus{{object: ir, status: "invalid", msg: fmt.Sprintf("different health checks were specified for service %q", svc.Name)}}
```
stevesloka: nit: Should we include the namespace in the error so it's clear where the issue is?

internal/dag/dag.go
```diff
@@ -391,29 +403,69 @@ func (d *DAG) recompute() dag {
...
+		// route cannot both delegate and point to services
+		if len(route.Services) > 0 && route.Delegate.Name != "" {
+			return []ingressrouteStatus{{object: ir, status: "invalid", msg: fmt.Sprintf("route %q: cannot specify services and delegate in the same route", route.Match)}}
+		}
 		// base case: The route points to services, so we add them to the vhost
 		if len(route.Services) > 0 {
 			if !matchesPathPrefix(route.Match, prefixMatch) {
-				// TODO: set status
-				return
+				return []ingressrouteStatus{{object: ir, status: "invalid", msg: "the path prefix does not match the parent's path prefix"}}
```
stevesloka: Would be good to include the parent path that's delegated.

internal/dag/dag_test.go
```diff
@@ -3382,6 +3383,537 @@ func TestMatchesPathPrefix(t *testing.T) {
...
+			t.Fatalf("stack is empty")
+		}
+		if got.Name != w {
+			t.Fatalf("wanted %v, but got %v", w, got)
+		}
+	}
+}
+
+func recurseIngressRouteStack(s *ingressrouteStack, i int) {
+	if i == 3 {
```
stevesloka: ??

alexbrand:
After talking to @stevesloka, we decided to punt on load balancing strategy and health check validation to avoid blowing up the scope of this PR more than it already is.

alexbrand:
Chatted with @davecheney. We don't need the stack for keeping track of the ingress routes we have already visited. The reason is that we already get that behavior for free when we pass the slice of visited ingress routes recursively. When we pass the slice, we make a copy of the slice and any modifications done to the slice will not be visible to the caller.

davecheney:
Committing on Alex's behalf.

Issue #450, opened 2018-06-25
ingressroute: add documentation


rosskukulinski:
We need to document the new IngressRoute API and provide some working examples.

PR #516, opened 2018-07-05
ingressroute documentation

rosskukulinski:
Adds documentation for IngressRoute CRD.

Signed-off-by: Ross Kukulinski <ross@kukulinski.com>

### DIFF
--- a/docs/ingressroute.md
+++ b/docs/ingressroute.md
@@ -1,0 +1,613 @@
+# IngressRoute
+
+The [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) object was added to Kubernetes in version 1.1 to describe properties of a cluster-wide reverse HTTP proxy. Since that time, the Ingress object has not progressed beyond the beta stage, and its stagnation inspired an [explosion of annotations](https://github.com/kubernetes/ingress-nginx/blob/master/docs/user-guide/nginx-configuration/annotations.md) to express missing properties of HTTP routing.
+
+The goal of the `IngressRoute` Custom Resource Definition (CRD) is to expand upon the functionality of the Ingress API to allow for a richer user experience as well as solve shortcomings in the original design.
+
+At this time, Contour is the only Kubernetes Ingress Controller to support the IngressRoute CRD, though there is nothing that inherently prevents other controllers from supporting the design.
+
+## Key IngressRoute Benefits
+
+- Safely supports multi-team Kubernetes clusters, with the ability to limit which Namespaces may configure virtual hosts and TLS credentials
+- Enables delegation of routing configuration for a path or domain to another Namespace
+- Accepts multiple services within a single route and load balances traffic across them
+- Natively allows defining service weighting and load balancing strategy without annotations
+- Validation of IngressRoute objects at creation time and status reporting for post-creation validity
+
+## Ingress to IngressRoute
+
+A minimal Ingress object might look like:
+
+```yaml
+# ingress.yaml
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: basic
+spec:
+  rules:
+  - host: foo-basic.bar.com
+    http:
+      paths:
+      - backend:
+          serviceName: s1
+          servicePort: 80
+```
+
+This Ingress object, named `basic`, will route incoming HTTP traffic with a `Host:` header for `foo-basic.bar.com` to a Service named `s1` on port `80`.  Implementing similar behavior using an IngressRoute looks like this:
+
+```yaml
+# ingressroute.yaml
+apiVersion: contour.heptio.com/v1beta1
+kind: IngressRoute
+metadata: 
+  name: basic
+spec: 
+  virtualhost:
+    fqdn: foo-basic.bar.com
+  routes: 
+    - match: /
+      services: 
+        - name: s1
+          port: 80
+```
+
+**Lines 1-4**: As with all other Kubernetes objects, an IngressRoute needs apiVersion, kind, and metadata fields. Note that the IngressRoute API is currently considered beta.
+
+**Line 6-7**: The presence of the `virtualhost` field indicates that this is a root IngressRoute that is the top level entry point for this domain.  The `fqdn` field specifies the fully qualified domain name that will be used to match against `Host:` HTTP headers.
+
+**Lines 8-9**: IngressRoutes must have one or more `routes`, each of which must have a path to match against (e.g. `/blog`) and then one or more `services` which will handle the HTTP traffic.  
+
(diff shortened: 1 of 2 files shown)


rosskukulinski:
Fixes #450 when merged.

docs/ingressroute.md
```diff
@@ -0,0 +1,297 @@
...
+# ingress-tls.secret.yaml
+apiVersion: v1
+data:
+  tls.crt: base64 encoded cert
+  tls.key: base64 encoded key
+kind: Secret
+metadata:
+  name: testsecret
+  namespace: default
+type: Opaque
```
stevesloka: We should use the `kubernetes.io/tls` type for TLS secrets: https://github.com/kubernetes/api/blob/master/core/v1/types.go#L4944

docs/ingressroute.md
```diff
@@ -0,0 +1,297 @@
...
+# domain-aliases.ingressroute.yaml
+apiVersion: contour.heptio.com/v1beta1
+kind: IngressRoute
+metadata: 
+  name: domain-aliases
+  namespace: default
+spec: 
+  virtualhost:
+    fqdn: bar.com
+    aliases:
```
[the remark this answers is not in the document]
rosskukulinski: updated with `(Not supported in beta.1)` in Aliases header

docs/ingressroute.md
```diff
@@ -0,0 +1,613 @@
...
+      - www.bar.com
+  routes: 
+    - match: /
+      services: 
+        - name: s1
+          port: 80
+```
+
+##### TLS
```
stevesloka: TLS is not implemented in beta.1

docs/ingressroute.md
```diff
@@ -0,0 +1,613 @@
...
+        - name: s2-def-strategy
+          port: 80
+        - name: s3-def-strategy
+          port: 80
+          strategy: Random # Overrides default LB algorithm for only this Service
+```
+
+#### Active Health Checking
+
+While Kubernetes' service discovery system provides active health checking through [Pod Readiness Probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/), there are scenarios where additional active health checking is desired.  For example, when the Kubernetes API server is down (either for regular maintenance or an outage), Pods that are failing Readiness Checks will not be removed from the corresponding Endpoint objects.  Envoy will continue to try to route traffic to these failing pods which results in user-facing HTTP errors or timeouts.
```
stevesloka: Should this section describe Liveliness Probes instead of Readiness?

rosskukulinski: I believe I'm correct in referencing Readiness Probes. 

> The kubelet uses liveness probes to know when to restart a Container. 
> The kubelet uses readiness probes to know when a Container is ready to start accepting traffic.

rosskukulinski:
@stevesloka updated & rebased from your comments.  Double check me on Liveness vs Readiness probes.

docs/ingressroute.md
stevesloka: Readiness only happens at the start of the container, once it passes it's never called again. Liveliness is called continually as the container runs to make sure it's still healthy.

rosskukulinski: Steve's CKA and CKAD have been revoked :)

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#define-readiness-probes

rosskukulinski:
@Bradamant3 is there anything else that I need to do to wire up the new ingressroute.md file into the documentation site?  We don't need to actually update the site until 0.6 (still a few weeks out), but I figured I could do some of the low-hanging fruit now.

docs/ingressroute.md
```diff
@@ -0,0 +1,613 @@
...
+
+The [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) object was added to Kubernetes in version 1.1 to describe properties of a cluster-wide reverse HTTP proxy. Since that time, the Ingress object has not progressed beyond the beta stage, and its stagnation inspired an [explosion of annotations](https://github.com/kubernetes/ingress-nginx/blob/master/docs/user-guide/nginx-configuration/annotations.md) to express missing properties of HTTP routing.
+
+The goal of the `IngressRoute` Custom Resource Definition (CRD) is to expand upon the functionality of the Ingress API to allow for a richer user experience as well as solve shortcomings in the original design.
+
+At this time, Contour is the only Kubernetes Ingress Controller to support the IngressRoute CRD, though there is nothing that inherently prevents other controllers from supporting the design.
+
+## Key IngressRoute Benefits
+
+- Safely supports multi-team Kubernetes clusters, with the ability to limit which Namespaces may configure virtual hosts and TLS credentials
```
davecheney: full stop at the end of each bullet point.

docs/ingressroute.md
```diff
@@ -0,0 +1,613 @@
...
+        - name: s1-def-strategy
+          port: 80
+        - name: s2-def-strategy
+          port: 80
+        - name: s3-def-strategy
+          port: 80
+          strategy: Random # Overrides default LB algorithm for only this Service
+```
+
+#### Active Health Checking
```
davecheney: Suggestion: maybe rephrase this as "Per upstream health checking", then we can avoid talking about Kubernete's pod health check and just talk about the healthcheck: key in the ingressroute doc.

docs/ingressroute.md
```diff
@@ -0,0 +1,613 @@
...
+
+It is possible for IngressRoute objects to exist that have not been delegated to by another IngressRoute.  These objects are considered "orphaned" and will be ignored by Contour in determining ingress configuration.
+
+### Restricted root namespaces
+
+IngressRoute delegation allows for Administrators to limit which users/namespaces may configure routes for a given domain, but it does not restrict where root IngressRoutes may be created. Contour has an enforcing mode which accepts a list of namespaces where root IngressRoutes are valid. Only users permitted to operate in those namespaces can therefore create IngressRoutes with the `virtualhost` field.
+
+This restricted mode is enabled in Contour by specifying a command line flag, `--ingressroute-root-namespaces`, which will restrict Contour to only searching the defined namespaces for root IngressRoutes.  This CLI accepts a comma separated list of namespaces where IngressRoutes are valid (e.g. `--ingressroute-root-namespaces=default,kube-system,my-admin-namespace`).
+
+IngressRoutes with a defined `virtualhost` field that are not in one of the allowed root namespaces will be flagged as `invalid` and will be ignored by Contour.
```
davecheney: Suggestion: please add a note that this is only supported for IngressRoute crds, --ingressroute-root-namespaces does not affect the operation of v1beta1.Ingress objects.

PR #519, opened 2018-07-06
Fixedbugs/518

davecheney:


internal/contour/cluster.go
```diff
@@ -1,4 +1,4 @@
-// Copyright © 2018 Heptio
+// Copyright © 2019 Heptio
```
stevesloka: 2019?

davecheney: There's some key combination that I hit in vim that autoincrements the first thing that looks like a number. I'll fix this and merge.

PR #532, opened 2018-07-09
Write Status to Kubernetes IngressRoute object after DAG update 

### LINKED ISSUE #524: Update IngressRouteStatus back to API Server
Currently, #469 takes the resulting IngressRouteStatus and print's it to stdout. We need to push that status back to the IngressRoute object in the k8s api. 

I plan to pick up the client-go lister to allow for us to check the status of the IngressRoute object without requiring a round trip to the server.

stevesloka:
Fixes #524 by updating IngressRoute Status fields after a DAG update. It only updates when the status changes (e.g. msg != ir.msg or desc != ir.desc)

### DIFF
--- a/internal/contour/adapter.go
+++ b/internal/contour/adapter.go
@@ -17,9 +17,9 @@
 package contour
 
 import (
-	"fmt"
-
 	"github.com/heptio/contour/internal/dag"
+	"github.com/heptio/contour/internal/k8s"
+	"github.com/sirupsen/logrus"
 	"k8s.io/api/extensions/v1beta1"
 )
 
@@ -36,6 +36,9 @@ type DAGAdapter struct {
 	ListenerCache
 	RouteCache
 	ClusterCache
+
+	IngressRouteStatus *k8s.IngressRouteStatus
+	logrus.FieldLogger
 }
 
 func (d *DAGAdapter) OnAdd(obj interface{}) {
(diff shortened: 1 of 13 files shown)


alexbrand:
@stevesloka Looks like a vscode artifact made it into the commit.

internal/k8s/status.go
```diff
@@ -0,0 +1,65 @@
...
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// IngressRouteStatus allows for updating the object's Status field
+type IngressRouteStatus struct {
+	Client clientset.Interface
+}
+
+// SetStatus sets the IngressRoute status field to an Valid or Invalid status
+func (irs *IngressRouteStatus) SetStatus(msg, desc string, existing *ingressroutev1.IngressRoute) error {
```
alexbrand: nit: thoughts on renaming `msg` to `status`?

internal/contour/adapter.go
```diff
@@ -76,7 +77,7 @@ func (d *DAGAdapter) OnDelete(obj interface{}) {
 
 func (d *DAGAdapter) setIngressRouteStatus(statuses dag.IngressrouteStatus) {
 	for _, s := range statuses.GetStatuses() {
-		fmt.Println(fmt.Sprintf("DAGVer: %d IR: %s Namespace: %s Status: %s Msg: %s", statuses.GetVersion(), s.GetIngressRouteName(), s.GetIngressRouteNamespace(), s.GetStatus(), s.GetMsg()))
+		d.IngressRouteStatus.SetStatus(s.GetMsg(), s.GetStatus(), s.GetObject())
```
alexbrand: Do we need to handle this error?

internal/k8s/status.go
stevesloka: I renamed `msg` to `status` and did that back through the dag as well.

internal/contour/adapter.go
stevesloka: I wasn't sure if we wanted to pipe through a logger but I did that and have this error handled now.

davecheney:
LGTM. Thank you.

Issue #525, opened 2018-07-06
IngressRoute Metrics


rosskukulinski:
Contour prometheus metrics will break down:

1. [ ] the count of IngressRoutes by namespace
2. [ ] count of root domains by namespace 
3. [ ] count of orphaned ingressroutes by namespace
4. [ ] count of valid routes by namespace.
5. [ ] count of invalid routes by namespace

(1) = (3 + 4 + 5)

PR #539, opened 2018-07-11
All: add prometheus metric handler

davecheney:
Updates #525

- Rework admin/debug http API handler
- rename flags to http-address and http-port, the previous ones had a
space and were probably never used by anyone
- add /heath and /metrics endpoints on http API handler

This adds 6 new dependencies over beta.1 but only a few hundred kb to the image size.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -23,6 +23,12 @@
   revision = "2efee857e7cfd4f3d0138cc3cbb1b4966962b93a"
 
 [[projects]]
+  branch = "master"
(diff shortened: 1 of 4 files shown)


davecheney:
Thanks for your branch, now beta.1 is cut do you want to propose it against contour:master.

https://github.com/stevesloka/contour/blob/metrics/internal/contour/metrics.go#L34-L39

Maybe if NewMetrics took a registry as a parameter rather than creating one internally it would give us a hook to access it for testing.

PR #541, opened 2018-07-11
Fix ingressroute docs re: root-namespaces CLI flag

rosskukulinski:
Short term fix to update documentation to reflect current behavior of the `--ingressroute-root-namespaces` CLI flag which must be specified multiple times, one per namespace.

Current documentation recommends to use comma-separated values which does not work.  #540 is for the longer-term UX work.

Signed-off-by: Ross Kukulinski <ross@kukulinski.com>

### DIFF
--- a/docs/ingressroute.md
+++ b/docs/ingressroute.md
@@ -607,7 +607,7 @@ IngressRoute delegation allows for Administrators to limit which users/namespace
 Contour has an enforcing mode which accepts a list of namespaces where root IngressRoutes are valid.
 Only users permitted to operate in those namespaces can therefore create IngressRoutes with the `virtualhost` field.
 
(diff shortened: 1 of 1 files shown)


docs/ingressroute.md
```diff
@@ -607,7 +607,7 @@ IngressRoute delegation allows for Administrators to limit which users/namespace
 Contour has an enforcing mode which accepts a list of namespaces where root IngressRoutes are valid.
 Only users permitted to operate in those namespaces can therefore create IngressRoutes with the `virtualhost` field.
 
-This restricted mode is enabled in Contour by specifying a command line flag, `--ingressroute-root-namespaces`, which will restrict Contour to only searching the defined namespaces for root IngressRoutes. This CLI accepts a comma separated list of namespaces where IngressRoutes are valid (e.g. `--ingressroute-root-namespaces=default,kube-system,my-admin-namespace`).
+This restricted mode is enabled in Contour by specifying a command line flag, `--ingressroute-root-namespaces`, which will restrict Contour to only searching the defined namespaces for root IngressRoutes. This CLI flag can be passed multiple times, one per Namespace where root IngressRoutes are valid (e.g. `--ingressroute-root-namespaces=default,kube-system,my-admin-namespace`).
```
alexbrand: We should update the example as well

rosskukulinski:
Ahh, good catch @alexbrand. Updated

PR #549, opened 2018-07-17
internal/dag: move processIngressRoute func to a struct

### LINKED ISSUE #521: Refactor and improve dag's processIngressRoute function
The current implementation of the processIngressRoute function works, but it takes a rather long list of parameters. Most of these parameters can actually live on a struct given that they are either accumulating results or remaining unchanged while recursing.

alexbrand:
Refactored the `processIngressRoute` function onto a method on a struct. The method only has parameters that change across recursive calls, while the rest of params now live on the struct.

Fixes #521

### DIFF
--- a/internal/dag/dag.go
+++ b/internal/dag/dag.go
@@ -393,8 +393,15 @@ func (d *DAG) recompute() (dag, IngressrouteStatus) {
 		}
 
 		prefixMatch := ""
-		ve := d.processIngressRoute(ir, prefixMatch, nil, host, service, vhost, orphaned)
-		status = append(status, ve...)
+		irp := ingressRouteProcessor{
+			host:          host,
+			service:       service,
+			vhost:         vhost,
+			ingressroutes: d.ingressroutes,
+			orphaned:      orphaned,
+		}
+		sts := irp.process(ir, prefixMatch, nil)
+		status = append(status, sts...)
 	}
 
 	var _d dag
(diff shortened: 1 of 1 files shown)


internal/dag/dag.go
```diff
@@ -416,7 +423,28 @@ func (d *DAG) recompute() (dag, IngressrouteStatus) {
...
+func (d *DAG) rootAllowed(ir *ingressroutev1.IngressRoute) bool {
+	if len(d.IngressRouteRootNamespaces) == 0 {
+		return true
+	}
+	for _, ns := range d.IngressRouteRootNamespaces {
+		if ns == ir.Namespace {
+			return true
+		}
+	}
+	return false
```
davecheney: here's a trick; you don't need to do it for this pr

```
for _, ns := range d.IngressRouteRootNamespaces {
		if ns == ir.Namespace {
			return true
		}
	}
	return len(d.IngressRouteRootNamespaces) == 0
```
If `len(d.IngressRouteRootNamespaces)` == 0 then the for loop will execute zero times, and the condition will return true, if there was at least one non matching value in `d.IngressRouteRootNamespaces` then the condition will be false and the function will return false.

davecheney: I think we have this kind of construct in a few places in contour.

davecheney:
Thank you for doing this.

Issue #568, opened 2018-07-21
Minor syntactical change proposal


stephenmoloney:
***Issue:***

A line like this:

```yaml
args: ["-c", "/config/contour.yaml", "--service-cluster", "cluster0", "--service-node", "node0", "-l", "info", "--v2-config-only"]
```

is a bit long for my liking with the potential to be even longer if using amendments.

***Suggestion:***

Syntax changes to break the arguments over multiple lines - example outlined below:

```yaml
       - image: gcr.io/heptio-images/contour:master
        imagePullPolicy: Always
        ports:
        - containerPort: 8000
          name: contour
        name: contour
        command: ["contour"]
        args:
        - serve
        - --incluster
        - --envoy-http-port=80
        - --envoy-https-port=443
```
I'm happy to make a PR for this proposal....

davecheney:
Sure thing. I actually thought that some of the manifests used that format. It doesn’t help that the apiserver will normalise it onto one line if you round trip through it via kubectl get -o yaml, but please do send a pr. 

A lot of those arguments are required by envoy, maybe in the future they can be made optional.

stephenmoloney:
ref #578

PR #578, opened 2018-07-23
Reduce long one liner argument statements into multiple shorter lines

stephenmoloney:
What does this MR do?

- Reduce long one liner args into multiple lines

Why is this MR needed?

- Enhance readability
- Easier readbility if adding additional optional arguments


stephenmoloney:
@davecheney 

A question I have is, should I also convert short one liners like

```
args: ["foo", "bar"]
```

into 

```
args:
- foo
- bar
```

in the name of consistency?

davecheney:
Fixes #568

PR #600, opened 2018-08-02
internal/dag: add websocket support to ingressroute

### LINKED ISSUE #595: Support enabling websockets via the IngressRoute resource
Contour supports an annotation on Kubernetes Ingress resources that enables websockets for specific routes. We should support this in the IngressRoute resource, either via a new field or an annotation.

Additionally, we should figure out and document how to make the AWS ELB work with websockets. We did some quick testing with a hacked version of contour (that defaulted all routes to `use_websockets=true`), and ran into issues with the ELB. Specifically, the websocket client (chrome) was getting a 403 error when attempting to open up websocket connections.

alexbrand:
Adds a new field to the Routes object inside the IngressRoute CRD called `EnableWebsockets`, which controls whether a client can upgrade to a websocket connection on the given route.

Fixes #595

### DIFF
--- a/apis/contour/v1beta1/ingressroute.go
+++ b/apis/contour/v1beta1/ingressroute.go
@@ -53,9 +53,12 @@ type TLS struct {
 type Route struct {
 	// Match defines the prefix match
 	Match string `json:"match"`
-	// Service are the services to proxy traffic
+	// Services are the services to proxy traffic
 	Services []Service `json:"services"`
+	// Delegate specifies that this route should be delegated to another IngressRoute
 	Delegate `json:"delegate"`
+	// Enables websocket support for the route
+	EnableWebsockets bool `json:"enableWebsockets"`
 }
 
 // Service defines an upstream to proxy traffic to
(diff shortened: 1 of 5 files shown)


stevesloka:
Overall this LGTM with a couple comments:

1. Is there any reason to not default this on for all routes?
2. It might be nice to as well have a default like we do for HealthChecks / LBAlgo to set it at the root and let it cascade down (probably as a new issue).

alexbrand:
Thanks @stevesloka. You raise a good question in number 1. My assumption is that we default websockets to be disabled because Envoy does so. I did some research and found some context around this in the envoy repo: (https://github.com/envoyproxy/envoy/issues/319#issuecomment-308494230):
> We should explicitly specify in the route configuration whether a route is a WS route or not. Meaning, we will not support attempting to detect if it is WS all the way back to the backend. If we receive a WS upgrade over HTTP/1.1, we will check the route table to see if the matched route is WS, and if not, fail the request with 400 or some other code.

I still don't quite understand the reasoning behind having to mark a route as WS though. Curious if @dfc has any thoughts here.

PR #604, opened 2018-08-07
internal/contour: ignore endpoint updates if both old and new have zero endpoints

### LINKED ISSUE #602: Scaling a deployment to zero replicas does not remove endpoints from EDS
I think the problem is here: https://github.com/heptio/contour/blob/master/internal/contour/endpointstranslator.go#L72

When scaling a deployment to zero, the Endpoints object has zero subsets, which causes Contour to basically ignore the update.

#### Steps to reproduce:

1. Create a deployment
```
kubectl run nginx --image=nginx --expose --port 80
```

2. Create an ingressroute for the deployment

```
kind: IngressRoute
metadata:
  name: nginx
spec:
  routes:
  - match: /
    services:
    - name: nginx
      port: 80
  virtualhost:
    fqdn: example.com
```

3. Verify that nginx is accessible via the ingressroute

4. Scale the deployment to zero replicas:
```
kubectl scale deploy nginx --replicas=0
```

Expected result: The cluster is removed from EDS.

Actual result: The cluster remains in the EDS response, and can be verified with `contour cli eds | grep nginx`.

alexbrand:
Fixes #602

```
Before this change, we were ignoring endpoint update events that had
zero endpoints in the updated object. This was a problem because we were
not handling the case where we went from N endpoints to zero endpoints.

With this change, we only ignore the update if we are going from zero
endpoints to zero endpoints. Otherwise, we handle the update like any
other.

Signed-off-by: Alexander Brand <alexbrand09@gmail.com>
```

### DIFF
--- a/internal/contour/endpointstranslator.go
+++ b/internal/contour/endpointstranslator.go
@@ -69,8 +69,9 @@ func (e *EndpointsTranslator) addEndpoints(ep *v1.Endpoints) {
 }
 
 func (e *EndpointsTranslator) updateEndpoints(oldep, newep *v1.Endpoints) {
-	if len(newep.Subsets) < 1 {
-		// if there are no endpoints in this object, ignore it
+	if len(newep.Subsets) == 0 && len(oldep.Subsets) == 0 {
+		// if there are no endpoints in this object, and the old
+		// object also had zero endpoints, ignore this update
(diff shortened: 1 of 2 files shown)


davecheney:
Thanks for the quick work.

PR #605, opened 2018-08-08
Add support for Alias in IngressRoute

### LINKED ISSUE #451: ingressroute.spec.virtualhost.aliases support or drop?
Figure out how to add support for virtualhost aliases or drop them from the spec.

stevesloka:
Fixes #451 by adding support for `Aliases` inside IngressRoute documents. Any alias defined is appended to the RDS `domains` array. 

### DIFF
--- a/internal/contour/route.go
+++ b/internal/contour/route.go
@@ -163,9 +163,13 @@ func (v *routeVisitor) Visit() map[string]*v2.RouteConfiguration {
 			ingress_http.VirtualHosts = append(ingress_http.VirtualHosts, vhost)
 		case *dag.SecureVirtualHost:
 			hostname := vh.FQDN()
-			domains := []string{hostname}
+			aliases := vh.Aliases()
+			domains := append(aliases, hostname)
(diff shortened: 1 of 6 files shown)


internal/contour/route.go
```diff
@@ -113,9 +113,13 @@ func (v *routeVisitor) Visit() map[string]*v2.RouteConfiguration {
 		switch vh := vh.(type) {
 		case *dag.VirtualHost:
```
alexbrand: I think we need to update the SecureVirtualHost case as well, to support Aliases for HTTPS.

stevesloka: I should have added a comment on this, TLS won't work until #596 is merged. If we get that merged happy to update, or maybe make a new Issue to give TLS alias support.

alexbrand: Ah! Got it.

stevesloka:
@davecheney I rebased and also added suppport for aliases in Secure VHosts since that was merged. Thanks!

PR #607, opened 2018-08-08
internal/dag: invalidate ingressroutes that use the same fqdn

### LINKED ISSUE #601: Root IngressRoutes that define the same FQDN can result in conflicting ingress configuration
Currently, it is possible to define multiple root IngressRoutes that specify the same FQDN. Conflict can arise when two (or more) IngressRoutes define the same FQDN + path combination but point the route to a different service. For example:

```
kind: IngressRoute
metadata: 
  name: numberone
spec: 
  virtualhost:
    fqdn: example.com
  routes: 
    - match: /
      services: 
        - name: service1
          port: 80
---
kind: IngressRoute
metadata: 
  name: numbertwo
spec: 
  virtualhost:
    fqdn: example.com
  routes: 
    - match: /
      services: 
        - name: service2
          port: 80
```

One potential solution is to enforce that a given FQDN can only be used in a single Root IngressRoute. 

xref: https://github.com/heptio/gimbal/issues/209

alexbrand:
Pre-process the IngressRoutes in the dag builder to filter IngressRoutes that re-use a given FQDN, and mark them as invalid. The IngressRoutes that pass this pre-processing step are then further processed as usual.

Fixes #601

### DIFF
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -17,6 +17,7 @@ package dag
 
 import (
 	"fmt"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


rosskukulinski:
@alexbrand does this also cover domain aliases?

PR #633, opened 2018-08-15
internal/metrics: Add cache to metrics to allow for deletion of unneeded metrics

### LINKED ISSUE #612: IngressRoute gauge metrics not being cleared
Contour outputs a few `gauge` type Prometheus metrics related to IngressRoutes.  Namely:

* contour_ingressroute_invalid_total
* contour_ingressroute_valid_total
* contour_ingressroute_orphaned_total
* contour_ingressroute_total

Due to the nature of the Prometheus library, once a gauge metric is set, it will remain at that value until changed.  Currently we do NOT set these metrics to 0 if a corresponding ingressroute is removed (or changes state from valid->invalid, invalid->valid, orphaned->valid, orphaned->invalid, invalid->orphaned, valid->orphaned).

This results in misleading graphs (like the Gimbal `Valid IngressRoutes` graph), showing old Virtualhosts that no longer exist.

We should set gauges to 0 if the corresponding ingressroute or virtualhost or namespace is gone.  If possible, we should also delete the gauge from Prometheus (if that's possible - it may _not_ be).

stevesloka:
Fixes #612 by adding a metric cache to the metrics package. After an update, metrics are processed and any metrics that are not touched during an update are removed which fixes the issue where when values change, they would stay reporting when are really invalid. 

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -32,6 +32,9 @@ type Metrics struct {
 
 	CacheHandlerOnUpdateSummary prometheus.Summary
 	ResourceEventHandlerSummary *prometheus.SummaryVec
+
+	// Keep a local cache of metrics for comparison on updates
+	metricCache *IngressRouteMetric
 }
 
 // IngressRouteMetric stores various metrics for IngressRoute objects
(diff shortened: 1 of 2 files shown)


rosskukulinski:
Can't vouch for @stevesloka's code, but my quick smoke testing via Grafana is 👍

PR #634, opened 2018-08-15
internal/contour: Filter out duplicates for updates

### LINKED ISSUE #533: internal/adapter: Filter out duplicate updates before DAG rebuild of IngressRoute's
An update of an IngressRoute status will cause new updates of the DAG. Need to add some checks to see if the update should be a no-op and dropped.

stevesloka:
Fixes #533 by ignoring updates for duplicate updates.

### DIFF
--- a/internal/contour/k8s.go
+++ b/internal/contour/k8s.go
@@ -17,6 +17,8 @@
 package contour
 
 import (
+	"reflect"
+
 	"github.com/heptio/contour/internal/dag"
 	"github.com/heptio/contour/internal/metrics"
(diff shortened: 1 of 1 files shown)


stevesloka:
The original intent of this was to filter out duplicate updates to `IngressRoute` objects due to the Status update. Do you want to scope to just that place to help limit the surface area? I don't know how often we'll get updates for other objects from the api server due to the fact the queue resync interval is disabled.

davecheney:
I'm going to rebase and merge this now. Thank you.

PR #641, opened 2018-08-18
internal/dag: Enforce TLS

### LINKED ISSUE #456: Ingressroute needs support for enforcing TLS
Ingressroute needs support for the equiv. of Ingresses `ingress.kubernetes.io/force-ssl-redirect: true`

This should be a property on the spec.virtualhost.tls object, it cannot be specified on a delegate. 
```
spec:
  virtualhost:
     fqdn: "example.com"
     tls:
       force-ssl-redirect: true
```
force-ssl-redirect defaults to false.

Needs internal/dag and internal/e2e tests to verify that force-ssl-redirect: true generates 301 entries for all http routes.

stevesloka:
Fixes #456 by implementing 301 redirects when a Root IngressRoute defines a TLS section. Adds a `permitInsecure` to the Route section allowing for specific routes to server insecure traffic.

> Since defining a TLS section of a root IngressRoute tells Contour that it should set up a TLS listener and serve the provided TLS certificate/key, it's inherent that all requests be served over TLS. This results in any request to an insecure endpoint will receive a 301 http status code informing the client to redirect to the secure endpoint. No additional parameters need to be set for this functionality other than specifying TLS on the IngressRoute.
> 
> Additionally, it may be necessary to serve specific routes over an insecure endpoint. An example would be the challenges sent from LetsEncyrpt. Specific routes can set the permitInsecure parameter which will let that route serve insecure or secure traffic (Meaning no 301 redirects).

### DIFF
--- a/apis/contour/v1beta1/ingressroute.go
+++ b/apis/contour/v1beta1/ingressroute.go
@@ -36,8 +36,6 @@ type VirtualHost struct {
 	// are described in fqdn and aliases, the tls.secretName secret must contain a
 	// matching certificate
 	TLS *TLS `json:"tls"`
-	// If set to true, this virtual host will only be accessible via HTTPS.
-	HTTPSOnly bool `json:"httpsOnly"`
 }
 
 // TLS describes tls properties. The CNI names that will be matched on
@@ -60,6 +58,9 @@ type Route struct {
 	Delegate `json:"delegate"`
 	// Enables websocket support for the route
 	EnableWebsockets bool `json:"enableWebsockets"`
+	// Allow this path to respond to insecure requests over HTTP which are normally
+	// not permitted when a `virtualhost.tls` block is present.
+	PermitInsecure bool `json:"permitInsecure"`
 }
 
 // Service defines an upstream to proxy traffic to
--- a/design/ingressroute-design.md
+++ b/design/ingressroute-design.md
@@ -94,6 +94,8 @@ spec:
     service:
     - name: google-static
       port: 9000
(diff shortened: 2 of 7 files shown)


apis/contour/v1beta1/ingressroute.go
```diff
@@ -63,6 +61,8 @@ type Route struct {
 	Delegate `json:"delegate"`
 	// Enables websocket support for the route
 	EnableWebsockets bool `json:"enableWebsockets"`
+	// Allows this path to receive http requests even if the VHost specifies TLS only
```
davecheney: nit: s/Allows/Allow

davecheney: Try this as a replacement
```
// Allow this path to respond to insecure requests over HTTP which are normally
// not permitted when a `virtualhost.tls` block is present.
```

design/ingressroute-design.md
```diff
@@ -74,6 +74,8 @@ spec:
       secretName: google-tls
       # optional: minimum TLS version this vhost should negotiate
       minimumProtocolVersion: "1.3"
+      # optional: enforces all requests are secured via https by returning an http 301 status code redirecting to https
+      enforceTLS: true
```
davecheney: nit: you're probably already doing this but this should probably be deleted so that there is no way of saying "don't TLS me bro", when spec.virtualhost.tls is present.

stevesloka: yeah that's left over from the original design, thanks!

stevesloka:
@davecheney this is ready for review!

PR #645, opened 2018-08-22
Fix examples for the IngressRoute documentation

jelmersnoeck:
This example was missing the `services:` key, which would cause failure
when trying to submit the example to a kubernetes cluster. The change
is on [line 294](https://github.com/heptio/contour/compare/master...jelmersnoeck:correct-yaml-spec?expand=1#diff-d1dacc562d1dac587ae8c3e7aa565e6bR294).

It also [sets a valid name](https://github.com/heptio/contour/compare/master...jelmersnoeck:correct-yaml-spec?expand=1#diff-d1dacc562d1dac587ae8c3e7aa565e6bR255) for the Disable HTTP example. The `disableHttp`
name contains a capitalized character which isn't allowed by the 
validation regex:

`[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*`

I'm not sure if this is desired - I couldn't find a guideline for this - but I've
also added a commit (https://github.com/heptio/contour/commit/c5fd114fdae7eaf42c8f10ec6e331d6a3e86d65d) which removes trailing whitespaces from
the documentation file. If this is not desired, I can remove this commit.

### DIFF
--- a/docs/ingressroute.md
+++ b/docs/ingressroute.md
@@ -242,28 +242,6 @@ The TLS **Minimum Protocol Version** a vhost should negotiate can be specified b
   - 1.2
   - 1.1 (Default)
 
-#### Disable HTTP
-
-IngressRoutes support disabling HTTP at the VHost level, so that the listener is only exposed over HTTPS. This is achieved by setting the `httpsOnly` field to `true`.
-
-This functionality is equivalent to the `kubernetes.io/ingress.allow-http: false` annotation supported in the Ingress resource.
-
-```yaml
-apiVersion: contour.heptio.com/v1beta1
-kind: IngressRoute
-metadata:
-  name: disableHttp
-spec:
-  virtualhost:
-    fqdn: foo-basic.bar.com
-    httpsOnly: true
-  routes:
-    - match: /
-      services:
-        - name: s1
-          port: 80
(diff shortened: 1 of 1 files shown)


rosskukulinski:
Hi @jelmersnoeck! Thanks for the contribution!  I'll let @davecheney respond re: trailing whitespaces commit.

jelmersnoeck:
@stevesloka I'll look into replacing the "Disable HTTP" example with a "Permit Insecure" example?

docs/ingressroute.md
```diff
@@ -521,6 +500,26 @@ spec:
           port: 80
 ```
 
+#### Permit Insecure
+
+IngressRoutes support disabling HTTPS at the route level, so that the path responds to insecure requests over HTTP which are normally not permitted when a `virtualhost.tls` block is present.
```
stevesloka: Small nit, the `permitInsecure` doesn't disable HTTPS, only allows HTTP.

jelmersnoeck:
@stevesloka done!

stevesloka:
LGTM! Thanks for the PR @jelmersnoeck!

PR #648, opened 2018-08-23
Fixup docs on how to setup CertManager with LetsEncrypt

stevesloka:
The docs on how to setup CertManager were a bit out of date. This updates the way to deploy CertManager as well as the Let's Encrypt server URLs.

### DIFF
--- a/docs/cert-manager.md
+++ b/docs/cert-manager.md
@@ -124,7 +124,7 @@ cd cert-manager
 Then deploy cert-manager:
 
 ```
-kubectl -n cert-manager apply -f docs/deploy/rbac/
+kubectl apply -f contrib/manifests/cert-manager/with-rbac.yaml
 ```
 
 When cert-manager is up and running you should see something like:
@@ -163,7 +163,7 @@ spec:
     http01: {}
     privateKeySecretRef:
       name: letsencrypt-staging
-    server: https://acme-staging.api.letsencrypt.org/directory
+    server: https://acme-staging-v02.api.letsencrypt.org/directory
 ```
 
 replacing `user@example.com` with your email address.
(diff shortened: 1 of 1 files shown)


glerchundi:
Does `cert-manager` work with contour's `IngressRoute` CRD or just with plain-old `Ingress`es?

rosskukulinski:
Hi @glerchundi!

If you need to use HTTP-01 validation, as of today, `cert-manager` only works with the existing Ingress support in Contour.  Alternatively, you could use cert-manager's Certificate resource using DNS-01 validation to provide certs for IngressRoute.

Longer-term, we're collaborating with the cert-manager team to [add support for alternative ACME challenge solvers](https://github.com/jetstack/cert-manager/issues/783) (e.g. IngressRoute)

stevesloka:
@glerchundi, to expand a bit on Ross's comment, you can use them together by specifying the `Certificate` CRD manually. In that CRD you can tell cert-manager to use the `contour` IngressClass which will tell cert-manager to create the needed http response paths for Let's Encrypt based upon your requested fqdn's. Once this is complete you can reference the secret created by cert-manager in your IngressRoute.

Here's an example which I'm using for my [site today](https://stevesloka.com) with IngressRoutes:
```
apiVersion: certmanager.k8s.io/v1alpha1
kind: Certificate
metadata:
  name: containersteve-com
  namespace: default
spec:
  secretName: containersteve-secret
  issuerRef:
    name: letsencrypt-prod
    kind: ClusterIssuer
  commonName: stevesloka.com
  dnsNames:
  - stevesloka.com
  - www.stevesloka.com
  acme:
    config:
    - http01:
        ingressClass: contour
      domains:
      - stevesloka.com
      - www.stevesloka.com
```

As @rosskukulinski mentioned, IngressRoute's do not support this natively, but this is an easy workaround for now. I'm going to document this all properly soon!

rosskukulinski:
@stevesloka I think you should also update the [FAQ](https://github.com/heptio/contour/blob/master/docs/troubleshooting.md#cant-make-kube-lego-work-with-contour) to indicate that kube-lego is deprecated in favor of cert-manager. Then you can point to your new documentation!

glerchundi:
Thanks @stevesloka @rosskukulinski, as we're using google cloud dns (supported by cert-manager) and it seems to be fairly easy to protect/authorize a concrete recordset entry, we'll use this one + upgrade our current ingresses to the new ingressroutes.

Thanks for your pointers,

Issue #644, opened 2018-08-21
0.6.0 Docs Update


rosskukulinski:
There have been a number of modifications/removals/changes to our original planned IngressRoute design spec, examples, and docs leading up to 0.6.  This is a tracking issue for needed docs & example updates to ensure consistency across the project.

- [x] Virtual Host Aliases were removed from the spec (#652)
- [x] Add example using multiple IR to work around removed Aliases feature (#649)
- [x] TLS & permitInsecure (#652)

README.md
- [x] Second paragraph should expand on why we created IR and then point to IngressRoute documentation. (#659)
- [x] “This is an early release” — Remove! (#659)
- [x] Update design/roadmap.md file (#658)
- [x] Contour Kubernetes version still 1.7? (#659)
- [x] “See also the k8s documentation for Services & Ingress” (#659)
- [x] Add ZenHub contributor section (steal from Ark) (#659)

Docs/README.md
- [x] Move IngressRoute API link to under “Deployment Options” (#659)
- [x] Should Deployment Options have an “Advanced Deployment” where contour & envoy are not in the same Pod? (#659)

Docs/annotations.md
- [x] We should link to the IngressRoute doc form the top level and comment that Ingress annotations are eliminated by using IngressRoute API.  Service annotations still apply (PR #655)

Docs/prometheus.md
- [x] Is this still accurate? Especially the admin socket public? (PR #654)

Docs/minikube.md
- [x] Update minikube (‘qinikube-example’???) Kuard example to be IngressRoute.  (PR #660)

docs/about.md
- [x] Update to include translating IngressRoute objects (#666)
- [x] Update to include how Ingress -> DAG -> CDS/SDS/RDS (#666)

troubleshooting.md
- [x] Update https://github.com/heptio/contour/blob/master/docs/troubleshooting.md#cant-make-kube-lego-work-with-contour (#PR 656)
- [x] Should we have a doc for making DOT notation output of DAG? (PR #656)

Docs/deploy-options.md
- [x] kuard example should use IngressRoute https://github.com/heptio/contour/blob/master/docs/deploy-options.md#test (PR #664)

Docs/faq.md
- [x] Idea: “Ingress” or “IngressRoute”? In FAQ (PR #657)

rosskukulinski:
I haven't done a consistency review, but anything else that jumps to mind @davecheney @stevesloka @alexbrand?

cc: @Bradamant3

alexbrand:
[...]

PR #649, opened 2018-08-24
docs: Add example showing how multiple Root IngressRoute objects 

stevesloka:
The original IngressRoute design spec had support for an `Aliases` field. We've since removed that and this example shows how to accomplish the same functionality with multiple root IngressRoutes. 

It demonstrates how to have `bar.com`, as well as `www.bar.com`,  have the same delegate route which enables multiple FQDN's to be processed equally. 

Updates #644 


alexbrand:
@stevesloka can you please rebase this? Thanks!

PR #651, opened 2018-08-27
deployment: remove Aliases from CRD printer

alexbrand:
The aliases field has been removed from the IngressRoute CRD. This PR removes Aliases from the CRD "additional printer columns".


alexbrand:
Updates #644

PR #666, opened 2018-09-05
Add docs on DAG & IngressRoute objects

stevesloka:
Updates #644 

### DIFF
--- a/docs/about.md
+++ b/docs/about.md
@@ -1,6 +1,6 @@
 # About Contour and Envoy
 
-At the highest level Contour is a proxy that fulfills the Envoy Cluster Discovery Service (CDS) API by translating Kubernetes Ingress objects into JSON configuration fragments.
+At the highest level Contour is a proxy that fulfills the Envoy Cluster Discovery Service (CDS) API by translating Kubernetes Ingress objects and IngressRoute objects into Envoy configuration objects.
 
 ## Introduction to Envoy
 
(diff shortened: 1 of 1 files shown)


docs/about.md
```diff
@@ -1,6 +1,6 @@
 # About Contour and Envoy
 
-At the highest level Contour is a proxy that fulfills the Envoy Cluster Discovery Service (CDS) API by translating Kubernetes Ingress objects into JSON configuration fragments.
+At the highest level Contour is a proxy that fulfills the Envoy Cluster Discovery Service (CDS) API by translating Kubernetes Ingress objects and IngressRoute objects into JSON configuration fragments.
```
alexbrand: Now that we no longer support the JSON API, I think we can update "JSON configuration fragments" as well. Maybe something like: "into Envoy configuration objects" ?

docs/about.md
```diff
@@ -11,10 +11,18 @@ Unlike other ingress controllers, Envoy supports continual reconfiguration witho
 
 ## How Contour works
 
+Contour supports both Ingress objects and IngressRoute objects which are ways to define traffic routes into a cluster. 
+They are different in how the objects are structured as well as how they are implemented, however, are identical at their core intent, to simply route traffic.
```
alexbrand: nit: the intent of the Ingress and IngressRoute objects is to configure the routing of ingress traffic, not to actually route traffic.

docs/about.md
stevesloka: :+1:

docs/about.md
stevesloka: :+1:

PR #703, opened 2018-09-25
gopkg.toml: freshen dependencies

davecheney:
Pin a few dependencies that we don't want to upgrade too yet.


davecheney:
Fixed in #725

Issue #443, opened 2018-06-22
Upgrade to envoy 1.7.0


davecheney:
Low priority issue.

alvaroaleman:
Very naive question: What does this require aside from changing the manifests?

PR #707, opened 2018-09-25
all: update to Envoy 1.7.0

davecheney:
Updates #443

Fix configuration generated by Contour to avoid triggering warnings
under Envoy 1.7.0.

This is _not_ backward compatible with Envoy 1.6.0.


davecheney:
Fixes #443

Issue #704, opened 2018-09-25
internal/generated: Allow users to consume IngressRoute client


stevesloka:
**Describe the solution you'd like**

Users who want to write controllers for their clusters which manipulate `IngressRoute` objects should be able to re-use the internal client Contour uses. Right now they cannot because the generated client exists within the `internal` package of Contour. 

Also, users should not have to depend on all the entire Contour repo as a dependency. One possible solution would be to move the generated client as well as the API spec to a new repo. 

This is a result of a Slack conversation with @amoskyler.

davecheney:
I agree! We should do this.

The only question I have is should we _just_ export our IngressRoute types, or should we combine them with the others from ark and sonobuoy.

davecheney:
Here's what I think we should do for 0.7 -- we can continue to iterate on it later if it turns out to be the wrong choice.

For 0.7 let's move the client out of internal/. If there are some good patterns we can follow from other projects I'll keep my curmudgeonly views on package naming to myself for the duration :P

PR #714, opened 2018-09-27
apis: Fix IngressRoute struct missing omitempty

### LINKED ISSUE #713: apis: IngressRoute struct missing `omitempty` for values

The IngressRoute CRD struct is missing `omitempty` on many of the non-required values. When interacting with this via golang, errors are thrown if all the values aren't specified:

```
spec.routes.services.healthCheck in body must be of type object: "null"
spec.routes.services.strategy in body should be one of [RoundRobin WeightedLeastRequest Random RingHash Maglev]
spec.routes.delegate.name in body should match '^[a-z0-9]([-a-z0-9]*[a-z0-9])?$'
spec.routes.delegate.namespace in body should match '^[a-z0-9]([-a-z0-9]*[a-z0-9])?$'  file="ingress_route.go:37"
```

If users create IngressRoute object via kubectl errors are not found due to differences in how it creates the resources.

stevesloka:
This fixes #713 by updating the `IngressRoute` struct to add `omitempty` to json params so that any params not supplied do not throw errors when integrating with golang. 

Other change is the `Delegate` struct is now a pointer so that we can add `omitempty` and make that optional. Tests / builder.go have been updated to reflect change.

### DIFF
--- a/apis/contour/v1beta1/ingressroute.go
+++ b/apis/contour/v1beta1/ingressroute.go
@@ -21,7 +21,7 @@ import (
 type IngressRouteSpec struct {
 	// Virtualhost appears at most once. If it is present, the object is considered
 	// to be a "root".
-	VirtualHost *VirtualHost `json:"virtualhost"`
+	VirtualHost *VirtualHost `json:"virtualhost,omitempty"`
 	// Routes are the ingress routes
 	Routes []Route `json:"routes"`
 }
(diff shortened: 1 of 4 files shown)


amoskyler:
Per your recommendation in the Kubernetes slack I copied `ingressroute.go` into my vendors folder and gave it a try. It resolved the omitempty issues.

davecheney:
We're not going to backport this to 0.6.1 because this PR is based on the move of ingressroute to apis/ in #704

Issue #720, opened 2018-10-02
IngressRoute does not honor kubernetes.io/ingress.class


pimvanpelt:
**What steps did you take and what happened:**
Running two contour pools, one called `contour-prod` and the other `contour-staging` (each setting the appropriate flag in `--ingress-class-name`

Two problems that signal that the annotation is not honored:

### Problem 1

Setting `annotations.kubernetes.io/ingress.class: contour-staging` on an IngressRoute applies this to all running contour instances, not the specific one selected.

```
$ kubectl -n contour get svc
NAME              TYPE           CLUSTER-IP     EXTERNAL-IP      PORT(S)                      AGE
contour-prod      LoadBalancer   10.109.26.45   194.126.235.48   80:30113/TCP,443:32570/TCP   3h42m
contour-staging   LoadBalancer   10.108.80.15   192.168.8.128    80:31054/TCP,443:30366/TCP   3h45m

$ kubectl get ingressroute --all-namespaces
NAMESPACE   NAME            FQDN          TLS SECRET   FIRST ROUTE   STATUS   STATUS DESCRIPTION
nginx       nginx-staging   k8s.ipng.nl                /             valid    valid IngressRoute
```

This IngressRoute is as expected, created on nginx-staging:
```
$ curl -I -H "Host: k8s.ipng.nl" 192.168.8.128
HTTP/1.1 200 OK
server: envoy
date: Tue, 02 Oct 2018 22:23:09 GMT
content-type: text/html
content-length: 612
last-modified: Tue, 03 Jul 2018 13:27:08 GMT
etag: "5b3b79ac-264"
accept-ranges: bytes
x-envoy-upstream-service-time: 5
```

This IngressRoute is NOT expected, it is not annotated to be created on nginx-prod:
```
$ curl -I -H "Host: k8s.ipng.nl" 194.126.235.48
HTTP/1.1 200 OK
server: envoy
date: Tue, 02 Oct 2018 22:24:15 GMT
content-type: text/html
content-length: 612
last-modified: Tue, 03 Jul 2018 13:27:08 GMT
etag: "5b3b79ac-264"
accept-ranges: bytes
x-envoy-upstream-service-time: 4
```

### Problem 2
Creating two IngressRoutes, one with contour-staging and one with contour-prod annotated, yields a conflict because both picked them up.

```
$ kubectl get ingressroute --all-namespaces
[...]

Issue #705, opened 2018-09-25
internal/grpc: TestGRPCStreaming test flake


davecheney:
```
--- FAIL: TestGRPCStreaming (0.41s)
    entry.go:146: time="2018-09-25T13:52:11+10:00" level=info msg=stream_wait connection=1 error_detail=nil resource_names="[]" response_nonce= type_url=type.googleapis.com/envoy.api.v2.Cluster version_info=
    entry.go:146: time="2018-09-25T13:52:11+10:00" level=info msg=response connection=1 count=0 error_detail=nil resource_names="[]" response_nonce= type_url=type.googleapis.com/envoy.api.v2.Cluster version_info=
    entry.go:146: time="2018-09-25T13:52:11+10:00" level=info msg=stream_wait connection=1 error_detail=nil resource_names="[]" response_nonce= type_url=type.googleapis.com/envoy.api.v2.Cluster version_info=
    entry.go:146: time="2018-09-25T13:52:11+10:00" level=error msg="stream terminated" connection=1 error="context deadline exceeded"
    --- FAIL: TestGRPCStreaming/StreamClusters (0.10s)
        server_test.go:78: expected "DeadlineExceeded", got "Unknown"
    entry.go:146: time="2018-09-25T13:52:11+10:00" level=info msg=stream_wait connection=1 error_detail=nil resource_names="[]" response_nonce= type_url=type.googleapis.com/envoy.api.v2.ClusterLoadAssignment version_info=
    entry.go:146: time="2018-09-25T13:52:11+10:00" level=info msg=response connection=1 count=1 error_detail=nil resource_names="[]" response_nonce= type_url=type.googleapis.com/envoy.api.v2.ClusterLoadAssignment version_info=
    entry.go:146: time="2018-09-25T13:52:11+10:00" level=info msg=stream_wait connection=1 error_detail=nil resource_names="[]" response_nonce= type_url=type.googleapis.com/envoy.api.v2.ClusterLoadAssignment version_info=
    entry.go:146: time="2018-09-25T13:52:11+10:00" level=error msg="stream terminated" connection=1 error="context canceled"
    entry.go:146: time="2018-09-25T13:52:11+10:00" level=info msg=stream_wait connection=1 error_detail=nil resource_names="[]" response_nonce= type_url=type.googleapis.com/envoy.api.v2.Listener version_info=
[...]

PR #739, opened 2018-10-05
Supports contour.heptio.com/ingress.class annotation for IngressRoutes

amoskyler:
I would like to propose the following PR as a solution to #720 

It works by checking for IngressRoute objects and testing for a `contour.heptio.com/ingress.class` annotation `validIngressClass`

Not sure if this is the correct approach, but my tests suggest it is functional.

### DIFF
--- a/internal/contour/k8s.go
+++ b/internal/contour/k8s.go
@@ -19,6 +19,7 @@ package contour
 import (
 	"reflect"
 
+	ingressroutev1 "github.com/heptio/contour/apis/contour/v1beta1"
 	"github.com/heptio/contour/internal/dag"
 	"github.com/heptio/contour/internal/metrics"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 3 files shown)


davecheney:
Thank you for this PR. I am going to hold off assigning a milestone for review until there is a decision on #720 on how to proceed.

amoskyler:
Hey @davecheney thanks for the tip - I ended up murdering the original rebase, sorry for all the updates. I've resolved it, so there is only a single commit.

amoskyler:
I see the grpc test is failed in server_test.go.

Do you have a suggestion on the best way to debug it? I'm a bit surprised that it fails here, as it appears to be very similar to one of the tests in rds_tests.go

davecheney:
I've just restarted the build. That flake is #705 and is really annoying, I'm trying to get it locked down for the 0.7 release.

PR #742, opened 2018-10-07
Changing StatsAddress flag from Int to String to prevent Init container crash loop.

josebiro:
And fixing deployment defaults for prometheus scraping to match reality.

This is a small bugfix, so didn't file an issue. Let me know if that's not appropriate.

All tests pass. Built and tested via docker and confirmed that:
- init container is no longer crash looping with the '--stats-address=0.0.0.0' flag set (it did previously, even though 0.0.0.0 is the default)
- the generated contour.yaml file produces the correct output.
- prometheus when configured with port 8002 can scrap metrics.


alexbrand:
Thanks @josebiro. I think this looks good, but do you mind adding a test case to internal/envoy/config_test.go where we set the stats address and port?

josebiro:
Test added. Checks should pass, everything is signed off, but for some reason github went from passing to waiting.

josebiro:
Rebased; feel free to merge at your will.

davecheney:
Thank you very much.

PR #759, opened 2018-10-16
internal: Enable Envoy healthchecks

### LINKED ISSUE #695: Healthcheck Envoy

Allow the ability to health check Envoy by using the health check filter (https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/http/health_check/v2/health_check.proto#envoy-api-msg-config-filter-http-health-check-v2-healthcheck). 

Updating the config passed to Envoy we could add this to the stats listener (or could add a new listener, but was trying to avoid having too many) which enables the health check filter on Envoy itself:

```
http_filters:
  - name: envoy.health_check
                     config:
                        endpoint: "/healthz"
                        pass_through_mode: false
```

Once implemented curling will return a 200 or 503:
```sh
$ curl -i  http://localhost:8002/healthz                                                                     
HTTP/1.1 200 OK
x-envoy-upstream-healthchecked-cluster: cluster0
date: Fri, 21 Sep 2018 02:37:25 GMT
server: envoy
content-length: 0
---
$  curl -i  http://localhost:8002/healthz                                                                     
HTTP/1.1 503 Service Unavailable
x-envoy-upstream-healthchecked-cluster: cluster0
date: Fri, 21 Sep 2018 02:48:57 GMT
server: envoy
content-length: 0
```

// cc @rdrgmnzs

stevesloka:
Fixes #695 by enabling a `/healthz` endpoint in Envoy to enable Kubernetes readiness probes. This also removes the ability to enable metrics since I'm reusing the same port to reduce the number of ports that we need to configure for Envoy. 


deployment/deployment-grpc-v2/02-contour.yaml
```diff
@@ -50,8 +56,6 @@ spec:
         command: ["contour"]
         args:
         - bootstrap
-        # Uncomment the statsd-enable to enable prometheus metrics
-        #- --statsd-enable
         # Uncomment to set a custom stats emission address and port
         #- --stats-address=0.0.0.0
         #- --stats-port=8002
```
rosskukulinski: @stevesloka should we change the commented out stats-port to something besides 8002?

docs/prometheus.md
```diff
@@ -9,9 +9,6 @@ avoid exposing the entire admin interface to Prometheus (and other workloads in
 the cluster), Contour configures a static listener that sends traffic to the
 stats endpoint and nowhere else.
 
-To enable the static listener, set the `--statsd-enabled` flag on the Contour
```
rosskukulinski: Should we add docs here for --stats-address and --stats-port?

rosskukulinski:
@stevesloka metrics can still be enabled, just that the end-user must now specify an address & port. Right?

alexbrand:
Will this cause issues if statsd is not deployed? I believe that flag was introduced when we added the ability to configure a statsd sink.

alexbrand:
Although it seems like it was controlling more than just the statsd piece.

stevesloka:
My idea to reuse the stats listener was to reduce the number of listeners. I could rever this and introduce another port if folks think that makes sense. 

This now "enables" the stats listener, if the forwarder isn't turned on then its sort of a no-op. Specifying the the IP/Port is a way to override the default (if someone would like to).

docs/prometheus.md
stevesloka: Yup good point @rosskukulinski, I think we missed those docs when this was introduced initally.

alexbrand:
I think reusing the port is fine. What _might_ be an issue is unconditionally enabling Envoy's statsd feature where it pushes metrics to a statsd sink. What if statsd is not running? Does envoy error out? Is it just going to continue trying to push metrics even though nothing is listening for them?

An alternative would be to move the "if statsdEnabled` check from old line 120 to old line 145 in `internal/envoy/config.go`.

stevesloka:
Yeah, we could do that. Right now just enabling the statsd interface will send UDP to nowhere. But if you don't want that turned on then I agree we should probably allow it to be disabled, however, the Prometheus stats interface is still available.

rosskukulinski:
+1 for disabling statsd-sink feature.  It's a CPU hog at scale.

stevesloka:
@alexbrand @rosskukulinski I'm retesting this now, but should be updated so that the `--enable-statsd` now just enables the statsd piece.

davecheney:
@stevesloka ping, is this ready to merge?

stevesloka:
yeah should be good now @davecheney, I just rebased it against master.

davecheney:
LGTM. Thank you.

PR #781, opened 2018-10-26
deployment: Added preStop hook to stop Envoy from receiving new requests

### LINKED ISSUE #696: Handle graceful restart of Envoy

When a new version of Envoy needs to be rolled out to the cluster or if the replica count needs to be scaled down, we should handle that better so requests are not lost. 

Envoy handles this already when the healthcheck fail endpoint is triggered, but k8s will just sigterm and inflight requests will get errors. 

See: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/envoy-users/HI1M8X4vHTQ/mcpQPKFaBQAJ

Ref: https://github.com/heptio/gimbal/issues/144

stevesloka:
Fixes #696 by adding a `preStop` lifecycle hook to the Envoy pod.

The process is as follows:
1. Shutdown is received
2. preStop hook will curl envoy and set its state to unhealthy
3. Envoy healthcheck will fail to stop requests from Kubernetes
4. Kubernetes will wait 30s to terminate (allowing requests to drop)

I ended up adding another container to the pod which includes `curl` since Envoy doesn't have it included in the container. 

@davecheney setting this to `WIP` since this needs a bit of discussion as well as a proper container (i.e. not from my personal account). 

### DIFF
--- a/deployment/deployment-grpc-v2/02-contour.yaml
+++ b/deployment/deployment-grpc-v2/02-contour.yaml
@@ -49,6 +49,10 @@ spec:
         volumeMounts:
         - name: contour-config
           mountPath: /config
+        lifecycle:
+          preStop:
+            exec:
+              command: ["wget", "-qO-", "http://localhost:9001/healthcheck/fail"] 
       initContainers:
       - image: gcr.io/heptio-images/contour:master
         imagePullPolicy: Always
--- a/deployment/ds-grpc-v2/02-contour.yaml
+++ b/deployment/ds-grpc-v2/02-contour.yaml
@@ -50,6 +50,10 @@ spec:
         volumeMounts:
         - name: contour-config
           mountPath: /config
+        lifecycle:
+          preStop:
+            exec:
+              command: ["wget", "-qO-", "http://localhost:9001/healthcheck/fail"] 
       initContainers:
       - image: gcr.io/heptio-images/contour:master
         imagePullPolicy: Always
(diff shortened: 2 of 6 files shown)


stevesloka:
Thanks for the suggestion @rochacon! I was looking for something like that, and didn't know the image had `wget`. I've updated the PR with your comments.

rochacon:
@stevesloka cool! I've been testing this and I still see some dropped connections but less than before, maybe there is still some other config to tune to achieve "zero downtime" rollouts.

I've noticed that both `deployment/render/daemonset-rbac.yaml` and `deployment/render/deployment-rbac.yaml` files are still with the custom image, maybe we should update it so this can get merged soon?

Thank you for this!

stevesloka:
Good catch, I forgot to run the `render.sh` (which is why the CI build failed). That's all fixed up now. 

I think to successfully drain all the connections we just need to wait longer. I've read some posts in the Envoy mailing list that set a 30s timeout which we're almost at with the default for the pod termination. Good to hear that you're seeing success! We'll need to add some docs around it I think to better understand the setup requirements.

deployment/deployment-grpc-v2/02-contour.yaml
```diff
@@ -49,6 +49,10 @@ spec:
         volumeMounts:
         - name: contour-config
           mountPath: /config
+        lifecycle:
+          preStop:
+            exec:
+              command: ["wget", "-qO-", "http://localhost:9001/healthcheck/fail"]
```
phylake: isn't this supposed to be a [POST](https://www.envoyproxy.io/docs/envoy/v1.9.0/operations/admin#post--healthcheck-fail)?

azuretek: I've confirmed this doesn't work, it returns a 400 error.

Adding `--post-data=''` is the fix.

davecheney:
@azuretek please open a new issue. Thank you.

Issue #787, opened 2018-10-29
[Feature] TLS TCP session proxying


davecheney:
This is a high level ticket to track the addition of TLS forwarding to Contour 0.8.

## What is being added?

Support for handling TLS encrypted TCP sessions is planned for Contour 0.8. This will terminate the TLS connection at the Envoy edge and forward the decrypted TCP flow to a backend pod making up a service. _[Note: this is my best understanding of the way Envoy 1.7.0 works, this may change_].

## How is it being added?

The DAG is being extended to differentiate between `dag.HTTPService` objects (#786), and `dag.TCPService` objects (#785). Once this is complete it will be possible to model the required configuration Envoy requires.

On the frontend the `ingressroute.spec` stanza will be extended with a new `forward` key. Something like
```
spec:
  virtualhost:
    fqdn: secure.example.com
    tls:
      secretname: "required"
  forward:
    services:
    - name: s1
      port: 80
      weight: 10
    - name: s2
      port: 8080
      weight: 20
    - name: s3
      port: 80
      weight: 70
      strategy: Random # Overrides default LB policy
```
- `services` is a list of `Service` objects, similar to `spec.routes.services`. There may be many entries in the `services` list, but only one `services` key may be present.
- `name` is the name of a service _in this Ingressroute's namespace_. 
- `port` is the port on the Service object. This port is expecting _NON TLS_ traffic. TLS demux is handled at the Envoy layer.
- `weight`, weighting is supported with the same logic as `spec.route.services`.
- `strategy`, load balancing strategies are supported per backend service.

Only one of `virtualhost` or `forward` keys may be present. `forward` is ignored if `spec.virtualhost.tls` is missing.
         
## Out of scope

- Raw TCP forwarding is not in scope _for Contour 0.8_. All connections proxied through Envoy must begin with a TLS SNI handshake. The SNI handshake is required to establish the hostname for the incoming connection. If TLS is not used then it is not clear what value using Contour adds over a traditional service load balancer.
- TLS forwarding whereby the TLS connection is forwarded _encrypted_ to the backend service is not in scope _for Contour 0.8_. This might be tracked as #15. 
- UDP forwarding is not in scope. It is unlikely Contour will ever attempt this feature.
- Delegation from a _Root_ ingressroute to another is not in scope.

alexbrand:
[...]

PR #806, opened 2018-11-15
internal/contour: add support for TCP Forwarding

davecheney:
Updates #787

This PR adds preliminary support for forwarding TLS encapsulated TCP
sessions.

This is supported only on ingressroute objects with a new spec.forward
key. The presence of spec.forward causes contour to ignore the contents
of spec.routes, although spec.routes must be present to pass validation.

spec.forward permits multiple services, although currently only one is
used.

contour-dag


berlinsaint:
thanks a lot

PR #809, opened 2018-11-15
Allow TLS 1.0 as a TLS minimumProtocolVersion version in an IngressRoute.

### LINKED ISSUE #808: TLS 1.0 cannot be enabled through an IngressRoute.
Specify an IngressRoute with `minimumProtocolVersion: "1.0"`.  The CustomResourceDefinition validation rejects the version.  Changing the CRD to accept the new string still does not allow TLS 1.0 to function.

To enable TLS 1.0 support for the route.

There is substantial reason that TLS >= 1.1 is the current recommended minimum for server-side TLS support, and the default configuration of contour should not enable the less secure TLS 1.0.  Nonetheless, there still exists large legacy software deployments that require TLS 1.0 support.  For example, we have several customers with government-provided stipulations that TLS 1.0 must be supported.

While best-practice is to avoid TLS 1.0, real world requirements prevent it from being killed off entirely just yet.  If a specific route requires legacy support, we should allow it to be manually configured.

Envoy itself supports TLS 1.0.  The DAG code ignores it as `internal/dag/build.go::compute()` does not map the string `"1.0"` to `auth.TlsParameters_TLSv1_0`  (any unrecognized string is mapped as TLS 1.1).  Completing the mapping allows TLS 1.0 to function.

As an aside, I haven't seen if/where contour allows changing of TLS ciphers.  That could be a different issue.  However, the default ciphers are sufficient for 1.0 support.

Issue #217 changed envoy/contour's default TLS to version 1.1.  This does not conflict with those changes.

I will provide a patch for consideration.

**Environment:**
All; verified on:
- Contour version: v0.7.0
- Kubernetes version: (use `kubectl version`): Client Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.2", GitCommit:"bb9ffb1654d4a729bb4cec18ff088eacc153c239", GitTreeState:"clean", BuildDate:"2018-08-08T16:31:10Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"11+", GitVersion:"v1.11.2-gke.9", GitCommit:"869cf55e493b5bc10ae080bdb3ac60338e868eff", GitTreeState:"clean", BuildDate:"2018-10-02T20:45:12Z", GoVersion:"go1.10.3b4", Compiler:"gc", Platform:"linux/amd64"}
- Kubernetes installer & version: GKE
- Cloud provider or hardware configuration: 1.11.2-gke.9
- OS (e.g. from `/etc/os-release`): container-optimized OS

smartyjohn:
Legacy software may not support TLS 1.1+.  This allows specifying a route to enable TLS 1.0.  The default behavior does *not* change, and remains TLS 1.1.

Patch to close #808.

### DIFF
--- a/deployment/common/common.yaml
+++ b/deployment/common/common.yaml
@@ -66,6 +66,7 @@ spec:
                         - "1.3"
                         - "1.2"
                         - "1.1"
+                        - "1.0"
             strategy:
               type: string
               enum:
--- a/deployment/render/daemonset-rbac.yaml
+++ b/deployment/render/daemonset-rbac.yaml
@@ -69,6 +69,7 @@ spec:
                         - "1.3"
                         - "1.2"
                         - "1.1"
+                        - "1.0"
             strategy:
               type: string
               enum:
--- a/deployment/render/deployment-rbac.yaml
+++ b/deployment/render/deployment-rbac.yaml
@@ -69,6 +69,7 @@ spec:
                         - "1.3"
                         - "1.2"
                         - "1.1"
+                        - "1.0"
             strategy:
               type: string
               enum:
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -394,6 +394,8 @@ func (b *builder) compute() *DAG {
 					svhost.MinProtoVersion = auth.TlsParameters_TLSv1_3
 				case "1.2":
 					svhost.MinProtoVersion = auth.TlsParameters_TLSv1_2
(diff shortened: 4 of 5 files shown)


davecheney:
As discussed in https://github.com/heptio/contour/issues/808#issuecomment-439222360 this PR is not accepted.

Issue #370, opened 2018-05-09
[feature] Add rate limiting support


dunjut:
From [annotations.md](https://github.com/heptio/contour/blob/master/docs/annotations.md), there are several limitations currently supported by contour, e.g. `max-connections` and `max-requests`.

In some scenario, we may also need ratelimits. For example, we don't expect envoy to send more than 1k requests per second to service foo, otherwise which would cause service foo to overload. This is pretty useful when service foo are actually doing multimedia jobs (transcoding/thumbnail/watermark/etc) which has estimable QPS capacity.

However, I'm not clear about how to make ratelimit globally restricted by envoy cluster, since all previous limitations are of per instance.

Hope this issue is worth discussing.

davecheney:
Thanks for raising this issue. To be honest, this isn't high on my priority list because I don't see how to implement this cleanly with a deployment of more than one contour pod as each will keep its own counter of requests in fly. You could probably do some back of the envelope math and set the rate limit to N over the number of pods in your contour deployment, but that's pretty fragile.

jhohertz:
Just dropping a reference into this ticket... The [Lyft reference implementation for rate limiting in envoy](https://github.com/lyft/ratelimit) would require some support to work with Contour, to:

- allow configuration of the ratelimit gRPC service into the data-plane
- some annotation sugar for Ingress (and I guess IngressRoutes too?) to allow setting the ratelimit configuration at each ingress instance.

That reference implementation is built to address the very kinds of issues raised earlier in the ticket around how to apply the limits globally vs. per-pod.

davecheney:
@jhohertz thanks for the update. I had a 5 second glance at the link you included to make sure I understood how this was implemented -- it looks like another xDS api, RLS. That looks pretty straight forward.

I'm not sure I can look at this for the 0.8 timeline, which is due at the end of next week. This is probably a medium to large (depending on your knowledge of Envoy and Contour) piece of work, but it should be straight forward to implement.

stevesloka:
I did some work on implementing this feature and there's a roadblock in how RateLimiting is currently implemented. Right now, RateLimiting can only be configured via the bootstrap config. There is an open issue to design the implementation to where it can be configured dynamically as we'd want in Contour (https://github.com/envoyproxy/envoy/issues/4669). 

[...]

PR #843, opened 2019-01-04
RateLimit design doc

stevesloka:
Updates #370 to provide a design doc implementation for RateLimiting. 

_Note: I have much of this developed on a branch, but want to get a good design doc agreed upon first._

### DIFF
--- a/design/ratelimit-design.md
+++ b/design/ratelimit-design.md
@@ -1,0 +1,175 @@
+# Envoy Rate Limiting Support
+
+The rate limit service configuration specifies the global rate limit service Envoy should talk to when it needs to make global rate limit decisions.
+If no rate limit service is configured, a “null” service will be used which will always return OK if called.
+
+Envoy's support of RateLimiting requires a rate limiting service to be exposed and support the gRPC IDL specified in [rls.proto](https://www.envoyproxy.io/docs/envoy/v1.9.0/api-v2/service/ratelimit/v2/rls.proto#envoy-api-file-envoy-service-ratelimit-v2-rls-proto).
+
+## Goals
+
+- Allow RateLimits to be applied to routes
+- Use Lyft's [ratelimiting implementation](https://github.com/lyft/ratelimit)
+
+## Non-goals
+
+- Support additional RateLimiting implementations
+- For the initial implementation, do not tightly integrate the RateLimiting service into Contour
+- Contour will not write the RateLimiter configuration (future versions could to make the integration smoother)
+
+## High-level Design
+
+RateLimiting will be disabled by default and will will be required to be enabled.
+The RateLimiting implementation will determine if the request sent to Envoy should be serviced or not.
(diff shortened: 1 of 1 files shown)


design/ratelimit-design.md
```diff
@@ -0,0 +1,154 @@
...
+      unit: second
+      requests_per_unit: 0
+```
+
+### IngressRoute Design
+
+The `IngressRoute` spec will be modified to allow specific routes to enable rate limiting.
+A new struct will be added to allow users to define what type of rate limiting should be applied to the route.
+
+The `rateLimit` struct will have two possible values:
```
davecheney: This feels like an envoy implementation detail, is there a way we can infer this value from the rate limit parameters supplied by the user?

design/ratelimit-design.md
```diff
@@ -0,0 +1,154 @@
...
+## Non-goals
+
+- Support additional RateLimiting implementations
+- For the initial implementation, do not tightly integrate the RateLimiting service into Contour
+- Contour will not write the RateLimiter configuration (future versions could to make the integration smoother)
+
+## High-level Design
+
+The RateLimiting implementation will determine if the request sent to Envoy should be serviced or not.
+Contour will require configuration to enable the rate limiting HTTP filter as well as pointing Envoy to the rate limit implementation service.
```
davecheney: We should probably only enable the rate limiting filter anywhere in the ingressroute chain has a rate limit stanza.

stevesloka: Yes I agree. I think looking at this now I never allowed for a spot to enable the service. That should probably be an arg to Contour as you'd only want to set it up once. Having that existing would be the trigger for contour to enable the filter, etc.

stevesloka:
> i'm guessing its some global param we'll have to pass to contour 

I think we could add an arg to Contour which would be the url to the rate limit service. The only downside to the argument is that users must have access to the deployment to add this, but that might be ok. 

> then do we reject ingressroute objects that have a rate limit specified, or do we accept them, without a rate limit

For similar items, we've been just setting status on the IngressRoute, so could do that for this as well. But what would we do for Ingress resources? We could log an error, or add a `status` annotation to the resource which seems werid.

design/ratelimit-design.md
stevesloka: Thought about this some more, I made this a map[string]string so that users could define whatever values they want. I think this is better and a good catch on the design.

stevesloka:
The other issue that I've come across now in looking at the implementation is the need for a way to restart Envoy in the event the bootstrap configuration needs to be changed. 

For example, if we disable rate limiting by default, and then later on enable, that would require a new bootstrap configuration to be applied to Envoy. Same would happen if we needed to change the server/port names. 

I could configure the cluster piece dynamically since it's just a cluster, but need to test if leaving rate_limiting enabled and referencing a bogus cluster is ok.

davecheney:
@stevesloka feel free to rebase this and merge it any time you're ready.

PR #849, opened 2019-01-10
internal/envoy: build bootstrap configuration programatically

### LINKED ISSUE #738: Generate bootstrap configuration programatically
Envoy 1.8.0 (#737) removes completely the support for the v1 JSON configuration option. We've dropped support long ago, but sport several command line flags in our deployment manifests which may need to be revised

https://github.com/heptio/contour/blob/master/deployment/deployment-grpc-v2/02-contour.yaml#L42

We should take the opportunity to freshen up the way bootstrap configs are generated by moving away from text/template to serialising the protobuf of the envoy config types.

This work will be supported by #708.

davecheney:
Fixes #738

This PR replaces the text/template yaml construction with protobuf -> json conversion. This should insulate us a little more from changes to the serialised spec of the bootstrap protobuf by giving us some more type safety -- if we keep our versions of go-control-plane in sync with envoy then we'll get compile errors when envoy deprecate fields.

As a side effect the configuration format is written in json, not yaml format. We long ago remove the requirement for the file to end in .yaml, and that was only to differentiate it from the version that ended in .json. Envoy doesn't care about the extension, but forcing the user to use `.json` would cause deployment failures during upgrade -- so, we're going to change our deployment and upgrade guides to mention `config.json`, but won't be that strict about enforcing it.

TODO:
- [x] Plumb `envoy.BootstrapConfig` variables into envoy.Bootstrap function
- [x] Unit tests


davecheney:
Moving to 0.10, too risky late in the 0.10 cycle

PR #850, opened 2019-01-13
tcpproxy: add tls passthrough

### LINKED ISSUE #15: SSL Passthrough
Hello, it would be good to know whether Contour supports ssl passthrough and if it doesn't - whether it would be possible/reasonable to add it. 

NGINX ingress controller has this ` ingress.kubernetes.io/ssl-passthrough: "true"` annotation, it might make sense to keep the format the same to make it easier to switch between ingresses.

```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    kubernetes.io/ingress.class: contour
    ingress.kubernetes.io/ssl-passthrough: "true"
  name: relay-ingress
  namespace: default
```

glerchundi:
Allow an IngressRoute for TCP forwarding but without a `tls` section
which implicitly configures it as a TLS Passthrough. TLS Passthrough
only reads the very first TLS packets and routes the traffic without
doing any kind of TLS termination. The chosen `virtualhost` is
discovered by reading the TLS SNI (Server Name Indication) extension.

Closes #15

### DIFF
--- a/docs/ingressroute.md
+++ b/docs/ingressroute.md
@@ -776,6 +776,10 @@ spec:
 
 The `spec.tcpproxy` key indicates that this _root_ IngressRoute will forward all de-encrypted TCP traffic to the backend service.
 
+In case `spec.virtualhost.tls` is not present, TLS is not going to be terminated
+on Envoy and will be forwarded to the specified services, where the termination
+would happen. This is called SSL/TLS Passthrough.
+
 ### Limitations
 
 The current limitations are present in Contour 0.8. These will be addressed in later Contour versions.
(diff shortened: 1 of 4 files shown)


glerchundi:
PTAL @davecheney we tested it in our environment and it should work as expected. This is the first of two MR we'll submit (the other will add support for delegation in TCP proxies).

Kudos to @igarciaolaizola for making most of the hard work by testing it empirically and fixing the a subtle bug that prevented this from working before. 👏

davecheney:
@glerchundi thank you very much for working on this. I really appreciate you taking the lead and pushing for this change.

We (waves at heptio in general) are at an all hands this week, but I will try to find the time to review and merge this and your related PRs

glerchundi:
No hurries at all take your time as we're not blocked by this. Thanks!

Issue #813, opened 2018-11-19
Support TLS verification of backend services


robbiemcmichael:
Contour should support TLS verification of backend services with a CA certificate. This feature is supported in Envoy, we just need to add the `trustedCA` field to the TLS `validation_context`: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/ssl#enabling-certificate-verification

robbiemcmichael:
Some personal opinion on implementation of this:

- Ideally the CA certificate should be configurable via a configmap, similar to how the TLS certificate and private key can be configured via a secret
- The ingress route that defines the backend service should define the configmap which is used for TLS verification (configmap would be in the same namespace as this ingress route)
- If TLS verification is enabled for a service in the ingress route then the `contour.heptio.com/upstream-protocol.h2` annotation should not be required on the upstream service - This is probably the most controversial as most ingress controllers for Kubernetes use service annotations for this, but I've always disliked needing to add it to the service as it should be the client (i.e. Envoy in this case) that specifies whether they are expecting to use HTTP or HTTPS

----

I'm working on the implementation of this feature for my own fork of Contour as we require TLS verification for an internal project, but I'd like to get some feedback so that hopefully it can be merged back into the main project eventually.

I'd like to extend the ingress route resource to support specifying the configmap for the CA certificate. I'm not sure how open you are to extending the resource, but from a technical standpoint it's not too bad as the `tlsVerification` field is optional. What I have right now is:

```yaml
apiVersion: contour.heptio.com/v1beta1
kind: IngressRoute
metadata:
  name: example
spec:
  routes:
  - match: /
    services:
    - name: service
      port: 443
      tlsVerification:
        configMapName: example-ca
```

The main thing I dislike about the above is that if the same CA is being used for every service then there would be a lot of unnecessary repetition. The different levels at which the CA could potentially be specified are:

1. At the ingress route level, all routes/services use the same CA - I don't like this as it seems possible that someone would want to use different CAs for some of the routes
2. At the route level, every service under a particular route uses the same CA - Seems like an improvement on the first option 
3. At the service level, every service can have a different CA (as shown above) - Gets repetitive if specifying multiple services under a single route
4. Some combination of the above options, which are listed in order of increasing precedence

stevesloka:
Hey, @robbiemcmichael thanks for the detailed work here! I agree with you on the points you mentioned regarding the effects of putting the CA on each route. 

[...]

PR #852, opened 2019-01-14
TLS verification of backend services

robbiemcmichael:
@davecheney It was good to meet you at the YOW! Conference. I briefly talked to you about implementing TLS verification for backend services; I finally got around to finishing off a sample implementation which verifies both the CA and hostname of the certificate.

This PR addresses issue #813. The implementation is similar to the design I suggested in that issue, however the ingress route schema has changed slightly to accommodate hostname verification. The question I had around the granularity of specifying the TLS verification settings (i.e. should it be one per ingress route, route or service?) is still my main concern and it would be good to get your thoughts. I'm happy to rework this PR based on any design concerns or code review comments.

### DIFF
--- a/internal/contour/cluster.go
+++ b/internal/contour/cluster.go
@@ -97,13 +97,7 @@ func visitClusters(root dag.Vertex) map[string]*v2.Cluster {
 
 func (v *clusterVisitor) visit(vertex dag.Vertex) {
 	switch service := vertex.(type) {
-	case *dag.HTTPService:
-		name := envoy.Clustername(&service.TCPService)
-		if _, ok := v.clusters[name]; !ok {
-			c := envoy.Cluster(service)
-			v.clusters[c.Name] = c
-		}
-	case *dag.TCPService:
+	case dag.Service:
 		name := envoy.Clustername(service)
 		if _, ok := v.clusters[name]; !ok {
 			c := envoy.Cluster(service)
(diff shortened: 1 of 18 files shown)


robbiemcmichael:
Updated my branch to remove some modifications to the deployment YAML file that were used for testing and accidentally left in as well as to add sign-offs to all commit messages.

internal/contour/cluster.go
```diff
@@ -97,13 +97,7 @@ func visitClusters(root dag.Vertex) map[string]*v2.Cluster {
...
 func (v *clusterVisitor) visit(vertex dag.Vertex) {
 	switch service := vertex.(type) {
-	case *dag.HTTPService:
-		name := envoy.Clustername(&service.TCPService)
-		if _, ok := v.clusters[name]; !ok {
-			c := envoy.Cluster(service)
-			v.clusters[c.Name] = c
-		}
-	case *dag.TCPService:
+	case dag.Service:
```
davecheney: This is good cleanup, but unrelated to this change. Can you please send it as its own PR. Thank you.

robbiemcmichael: It wasn't completely unrelated to this PR, I had to change `Clustername` to take a `Service` as a parameter since the hash needed to include [fields that are only present in an `HTTPService`](https://github.com/heptio/contour/pull/852/commits/b7be44985e81059703fd5c7f003a872c820d3b7a#diff-0f385e580f0bf79e437d3ad4d0b52f20R134).

The cleanup here was just a nice byproduct of changing that function, but I can still split that part out into a separate PR if you'd like.

davecheney:
Thank you for working on this. I'm sorry that it took so long to resolve and that we ended up duplicating some of your work.

I'm going to close this as a duplicate of #1045

PR #856, opened 2019-01-22
sticky sessions support

### LINKED ISSUE #361: Add sticky session annotation
At the moment the `lb_strategy` for each cluster defaults to whatever Envoy chooses. We should make this configurable via an annotation on the ingress object. Specifically i've been asked if we can support sticky session load balancing.

- [ ] choose an annotation name; `contour.heptio.com/lb-strategy:` sounds good.
- [ ] decide if it applies to the whole vhost, or only the routes mentioned in the ingress, hence you can have several ingresses with the same vhost and alter the lb strategy on a per path basis.

phylake:
> This project operates according to the _talk, then code_ rule.

#361 hasn't received much attention from the maintainers so I decided to code.

- As I understand it, IngressRoute is where new development is happening so I didn't add annotation support
- I added tests everywhere I could find tests for `ingressroutev1.IngressRoute`
- I noticed the CRD is lagging behind in terms of properties so I'm unsure if I needed updates there

closes #361

### DIFF
--- a/deployment/render/daemonset-rbac.yaml
+++ b/deployment/render/daemonset-rbac.yaml
@@ -111,6 +111,36 @@ spec:
                       namespace:
                         type: string
                         pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?$ # DNS-1123 label
+                  hashPolicy:
+                    type: array
+                    properties:
+                      terminal:
+                        type: boolean
+                      header:
+                        type: object
+                        required:
+                          - headerName
+                        properties:
+                          headerName:
+                            type: string
+                      cookie:
+                        type: object
+                        required:
+                          - name
+                        properties:
+                          name:
+                            type: string
+                          ttl:
+                            type: string
+                          path:
+                            type: string
+                      connectionProperties:
+                        type: object
+                        required:
+                          - sourceIp
+                        properties:
+                          sourceIp:
+                            type: boolean
                   services:
                     type: array
                     items:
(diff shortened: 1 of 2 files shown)


davecheney:
Thank you for your PR. I'll review this during the 0.10 milestone

docs/ingressroute.md
```diff
@@ -553,6 +553,43 @@ spec:
...
+    - match: /
+      hashPolicy:
+        # Precisely one of header, cookie, or connectionProperties must be set
+      - connectionProperties:
+          sourceIp: true
+        # Precisely one of header, cookie, or connectionProperties must be set
+        cookie:
+          name: some-cookie
+          path: cookie/path
+          ttl: 60s
```
Starblade42: If I pass TTL as a go duration string such as `60s`, the `hashPolicy` for my route doesn't get set.

However, if I set it as an int such as `60`, then it does get set, but in nanoseconds: `0.000000060s`. I believe that this is because the `ttl` value hasn't been parsed by `time.ParseDuration` before being used to construct the  `HashPolicyCookie`.

I also noticed that the cookie TTL doesn't appear in the tests, so we should add it.

apis/contour/v1beta1/ingressroute.go
```diff
@@ -50,6 +52,30 @@ type TLS struct {
 	MinimumProtocolVersion string `json:"minimumProtocolVersion,omitempty"`
 }
 
+type HashPolicyHeader struct {
```
davecheney: Each public type and field needs to have a comment so that these types are documented in godoc.

docs/ingressroute.md
phylake: I'll add the parsing support and additional tests. I'm not seeing tests that parse the CRD which would have caught this

Issue #854, opened 2019-01-22
Ingressroute status updates can cause update loop


dbason:
**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

Added non root IngressRoutes.  Then added a root IngressRoute delegating to those.  This caused a relatively high CPU and Memory usage (150%cpu usage and 2GB RAM - container is limited to 2GB).  This seems quite high for a single IngressRoute

**What did you expect to happen:**
Envoy continuing as normal

**Anything else you would like to add:**
After adding the IngressRoute with delegates it seems the envoy container is constantly logging the following message:
```[2019-01-22 02:20:05.816][1][info][upstream] source/server/lds_api.cc:80] lds: add/update listener 'ingress_https'```

**Environment:**

- Contour version: 1.8.1
- Kubernetes version: (use `kubectl version`): ```Client Version: version.Info{Major:"1", Minor:"8", GitVersion:"v1.8.3", GitCommit:"f0efb3cb883751c5ffdbe6d515f3cb4fbe7b7acd", GitTreeState:"clean", BuildDate:"2017-11-08T18:39:33Z", GoVersion:"go1.8.3", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.3", ```GitCommit:"a4529464e4629c21224b3d52edfe0ea91b072862", GitTreeState:"clean", BuildDate:"2018-09-09T17:53:03Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}
- Kubernetes installer & version: Rancher 2.1.5
- Cloud provider or hardware configuration: VMWare virtual servers
- OS (e.g. from `/etc/os-release`): CentOS 7

stevesloka:
That's interesting @dbason. How many services/endpoints do you have in the cluster? 

In performance testing, we've encountered high memory, but that was in the thousands of services range due to some gRPC implementations that should be resolved in Envoy soon (or could be already), however, seems like that listener update is the culprit. 

Could you share the `IngressRoute` yaml to look at?

dbason:
@stevesloka  here it is

`apiVersion: contour.heptio.com/v1beta1
kind: IngressRoute
metadata:
  name: ir-messenger-root-zero01
  namespace: zero
spec:
  virtualhost:
    fqdn: root.domain.com
    tls:
      secretName: dds-tls-secret
  routes:
    - match: /
      services:
        - name: s0
[...]

PR #861, opened 2019-01-25
internal/contour: filter out status update OnUpdate

davecheney:
Updates #854

Depending on the order in which parts of an ingressroute document are
processed the invalid status message generated may differ over runs.
Specifically two or more invalid components of an ingressroute will
patch status in a random order, defeating patch's "is this the same"
check and causing Contour to see the update to the ingressroute's
document, thus triggering a new OnUpdate event and the cycle starts
again.

For the interim filter out fields which are known to change -- status,
and metadata.resourceversion, during the OnUpdate comparision.

### DIFF
--- a/cmd/contour/contour.go
+++ b/cmd/contour/contour.go
@@ -99,6 +99,7 @@ func main() {
 	metrics := metrics.NewMetrics(registry)
 
 	reh := contour.ResourceEventHandler{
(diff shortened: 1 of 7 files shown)


rosskukulinski:
Didn't do a code review, but the problem description makes sense as to why the perf issues existed.

PR #866, opened 2019-01-25
Allow VirtualHost port to be defined so that Envoy will serve traffic on

### LINKED ISSUE #661: Add support for ports other than 80/443
I would like to see contour support ports outside of 80 and 443 for Virtualservices and Routes. Allowing virtualservices to only be created for two ports is highly restrictive. Looking at the source code it looks like ports 80 and 443 are hard coded, variables could easily be used for ports and allow for greater flexibility.

Use case:
I have a bare-metal K8s environment sitting behind a firewall. Firewall rules/portforwarding for ports 80 and 443 are restricted as many people are using the network behind the firewall. I've enabled port forwarding from an unused port ("30000" in this instance) to the Contour loadbalancer. When I try to route to my server from outside the firewall I have to include port 30000 to get the routing to work properly, but this will set my `Host` header to be `FQDN:30000`. Because of the way the `Host` header is set I cannot route to my service via contour.

stevesloka:
Fixes #661 by allowing the ports which Vhosts are configured to be dynamic. Previously, they were hardcoded to 80/443. 

@davecheney the VHost struct I added to the dag package seems out of place, not sure if you have any thoughts on how to pass that into the builder.go cleaner. What I have works, just smells funny to me. Open for additional thoughts about it. 

### DIFF
--- a/cmd/contour/contour.go
+++ b/cmd/contour/contour.go
@@ -22,6 +22,8 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/heptio/contour/internal/dag"
+
 	clientset "github.com/heptio/contour/apis/generated/clientset/versioned"
 	"github.com/heptio/contour/internal/contour"
 	"github.com/heptio/contour/internal/debug"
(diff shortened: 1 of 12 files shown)


davecheney:
Thanks. I'll look at this on monday once 0.9 is out the door.

joshrosso:
I've completed functional validation on this PR. It has enabled us to run nginx on `80/443` and contour on `8080/8443` by including the following flags.

```
        - --envoy-http-virtualhost-port=8080
        - --envoy-https-virtualhost-port=8443
        - --envoy-http-port=8080
        - --envoy-https-port=8443
```

Not implying this approach is the right design, just calling out it works as expected.

davecheney:
I'm not sure this is the right fix for #866. The problem in #866 is the long festering issue that envoy considers `Host: example.com` to be different from `Host: example.com:80`.

I'm not sure we should add this workaround.

stevesloka:
This is essentially removing the hard-coded `80/443` and allowing it to be configurable. I agree there could be better solutions, this was a short term fix which didn't change the functionality any. Can discuss further in #661

davecheney:
Superseded by #886

PR #873, opened 2019-01-29
RateLimiting implementation

stevesloka:
Updates #370 by implementing RateLimiting for Contour. This PR only implements for `IngressRoutes`, a future PR could add support for Ingress. 

I also need to add docs to explain how to configure/setup, that will come next, but wanted to agree on the impl and get a review first. 

### DIFF
--- a/internal/contour/listener_test.go
+++ b/internal/contour/listener_test.go
@@ -17,14 +17,14 @@ import (
 	"reflect"
 	"testing"
 
-	"github.com/envoyproxy/go-control-plane/envoy/api/v2"
+	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
 	"github.com/envoyproxy/go-control-plane/envoy/api/v2/auth"
 	"github.com/envoyproxy/go-control-plane/envoy/api/v2/listener"
 	ingressroutev1 "github.com/heptio/contour/apis/contour/v1beta1"
 	"github.com/heptio/contour/internal/envoy"
 	"github.com/heptio/contour/internal/metrics"
 	"github.com/prometheus/client_golang/prometheus"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/api/extensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
@@ -59,7 +59,8 @@ func TestListenerVisit(t *testing.T) {
 				Name:    ENVOY_HTTP_LISTENER,
 				Address: envoy.SocketAddress("0.0.0.0", 8080),
 				FilterChains: []listener.FilterChain{
-					filterchain(false, envoy.HTTPConnectionManager(ENVOY_HTTP_LISTENER, DEFAULT_HTTP_ACCESS_LOG)),
+					filterchain(false, envoy.HTTPConnectionManager(ENVOY_HTTP_LISTENER, DEFAULT_HTTP_ACCESS_LOG,
+						DEFAULT_RATE_LIMIT_SERVICE_DOMAIN, DEFAULT_RATE_LIMIT_SERVICE_STAGE, DEFAULT_RATE_LIMIT_SERVICE_FAILURE_MODE_DENY)),
 				},
 			}),
 		},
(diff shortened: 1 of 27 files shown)


internal/contour/listener_test.go
```diff
@@ -17,14 +17,14 @@ import (
 	"reflect"
 	"testing"
 
-	"github.com/envoyproxy/go-control-plane/envoy/api/v2"
+	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
```
davecheney: you shouldn't need to rename this import

internal/contour/listener_test.go
```diff
@@ -17,14 +17,14 @@ import (
...
 	"github.com/envoyproxy/go-control-plane/envoy/api/v2/auth"
 	"github.com/envoyproxy/go-control-plane/envoy/api/v2/listener"
 	ingressroutev1 "github.com/heptio/contour/apis/contour/v1beta1"
 	"github.com/heptio/contour/internal/envoy"
 	"github.com/heptio/contour/internal/metrics"
 	"github.com/prometheus/client_golang/prometheus"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
davecheney: or this one.

internal/dag/dag.go
```diff
@@ -18,7 +18,7 @@ package dag
 import (
 	"time"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
davecheney: here and everywhere else

internal/dag/dag.go
```diff
@@ -81,6 +81,9 @@ type Route struct {
 
 	// Indicates that during forwarding, the matched prefix (or path) should be swapped with this value
 	PrefixRewrite string
+
+	// Defines ratelimiting configuration if specified
+	RateLimitConfiguration map[string]string
```
davecheney: Is there a stronger type that can be used here?

internal/e2e/lds_test.go
```diff
@@ -73,7 +73,7 @@ func TestNonTLSListener(t *testing.T) {
 				Name:    "ingress_http",
 				Address: envoy.SocketAddress("0.0.0.0", 8080),
 				FilterChains: []listener.FilterChain{
-					filterchain(false, envoy.HTTPConnectionManager("ingress_http", "/dev/stdout")),
+					filterchain(false, envoy.HTTPConnectionManager("ingress_http", "/dev/stdout", "contour", 0, false)),
```
davecheney: Whenever I see test fixtures that include zero values for types, 0, and false, its a sure sign that we need 

a. either a helper
b. the API for the thing we're testing, envoy.HTTPConnectionManager has been stretched too far.

Could you try to dry this up with a helper for the moment, I don't think any of those parameters change apart from the name of the route, `ingress_{http,https}`

internal/envoy/route.go
```diff
@@ -41,6 +45,14 @@ func RouteRoute(r *dag.Route, services []*dag.HTTPService) *route.Route_Route {
 		)
 	}
 
+	rlActions := RateLimitActions(r)
```
davecheney: Could you change RateLimitActions to return a []*route.RateLimit, then do something like
```
ra.RateLimits = RateLimitActions(r)
```
If RateLimitActions returns nil, that will slot in just nicely as the 'not set' value for ra.RateLimits.

internal/envoy/route.go
```diff
@@ -137,6 +149,53 @@ func PrefixMatch(prefix string) route.RouteMatch {
 	}
 }
 
+// RateLimitStage returns stage configured for route. If none is provided
+// then the default is used
+func RateLimitStage(r *dag.Route) uint32 {
```
davecheney: Can this return a *types.UInt32Value directly, rather than requiring the caller to box it?

Also, I haven't gotten to the end of the diff yet, but if RateLimitState and RateLimitActions aren't called directly outside `internal/envoy` perhaps they can be made private to the package.

internal/dag/dag.go
stevesloka: There is, but was trying to not make it specific to the Lyft version of RateLimiting. Folks could implement their own service. Let me look and see if I can make it more specific but still generic.

internal/contour/listener_test.go
stevesloka: I think vscode is trying to be fancy. Let me see if I can turn that off.

internal/e2e/rds_test.go
```diff
@@ -2250,6 +2250,206 @@ func TestLoadBalancingStrategies(t *testing.T) {
...
+						Cluster: "default/kuard/80/da39a3ee5e",
+					},
+					RequestHeadersToAdd: []*core.HeaderValueOption{{
+						Header: &core.HeaderValue{
+							Key:   "x-request-start",
+							Value: "t=%START_TIME(%s.%3f)%",
+						},
+						Append: bv(true),
+					}},
+					RateLimits: routeclusterRatelimit(map[string]string{
```
drobinson123: This test is flaky due to Go's random map iteration order.

drobinson123:
Is there anything I can do to assist with getting this over the line?

I've rebased the branch to master, and added a helper for HTTPConnectionManager: https://github.com/zxvdr/contour/tree/new_rate_limit

davecheney:
@drobinson123 its not clear if we'll be able to include this feature for 1.0. We're having a meeting on Wednesday to prioritise. I will follow up on #370

stevesloka:
Updated design doc here: https://github.com/projectcontour/contour/pull/2283

stevesloka:
Ratelimiting is still top of mind for me, will come up with a new design doc and re-implement this since it's super stale.

PR #875, opened 2019-02-05
internal/envoy: only enable stats_sink if statsd enabled

davecheney:
Update #738


davecheney:
Updates #877

PR #886, opened 2019-02-11
internal/dag: allow external secure,insecure ports to be configured

### LINKED ISSUE #661: Add support for ports other than 80/443

davecheney:
Fixes #661

This PR exposes two new `contour serve` command line flags,
```
 --envoy-external-http-port
 --envoy-external-https-port
```
Which allow the external port, that is the port that will be presented
to with world via an ELB, NAT, or firewall, to be configured. This is
necessary to work around envoyproxy/envoy#1269 which mandates Contour add
explicit configuration of the port a HTTP/HTTPS request will arrive at
envoy.

By default the flags default to 80 and 443 respectively. If the ELB,
NAT, or firewall presenting Contour's Service on your network edge does
not use 80 and 443, you will need to supply the correct values via the
flags provided.

To avoid confusion this PT renames the existing `--envoy-http-port` and
`--envoy-https-port flags` to `--envoy-service-http-port` and
`--envoy-service-https-port` respectively. These flags default to 8080
and 8443 and must match the values for the http and https ports in the
Contour Service document. These flags are unrelated to the external
ports that HTTP/HTTPS requests will arrive on.

If you are deploying Contour in a Daemonset using `HostNeworking: true` you may
need to adjust `--envoy-service-http-port` and `--envoy-service-https-port`
to avoid port conflicts with other services bound to the Node's network
interface.


alexbrand:
Looks like it does indeed affect Gimbal, as we were using the `--envoy-http-port` flag.

joshrosso:
I'll be validating functionality this week.

Thanks for the work on this @stevesloka @davecheney.

Issue #410, opened 2018-05-25
Add handling for a default TLS certificate


mattalberts:
### Feature Request: Add handling for a default TLS certificate
 * Search Terms: "is:issue is:open TLS"

Add handling to allow use of a default TLS certificate, following these guidelines
 * The certificate is specified as a ```<namespace>/<name>``` reference to a secret.
 * The default secret is applied to Ingress specs that specify tls without providing spec.tls.secretName
 * Existing matching rules against spec.tls.hosts are still satisfied.
 * An Ingress spec.tls.secretName overrides the default secret.
 * The default secret does not apply to the default host "*"

Hopefully the guidelines don't sound to dogmatic. 😄  

All I'm try to capture is a loosening of ```validTLSSpecforVhost``` to only reject a TLS spec if both the default secret and  ```spec.tls.secretName``` are not set.
 * https://github.com/heptio/contour/blob/master/internal/contour/virtualhost.go#L234-L237

#### Example Depolyment|Container Definition
```yaml
        - name: contour
          image: gcr.io/heptio-images/contour:latest
          imagePullPolicy: Always
          command: ["contour"]
          args:
            - serve
            - --incluster
            - --ingress-class-name=contour
            - --default-tls-secret=$(POD_NAMESPACE)/contour-ingress
          env:
            - name: POD_NAME
              valueFrom:
                fieldRef:
                  fieldPath: metadata.name
            - name: POD_NAMESPACE
              valueFrom:
                fieldRef:
                  fieldPath: metadata.namespace
```

#### Example Ingress Definition
```yaml
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: contour-echo
  labels:
    ingress: contour
  annotations:
    kubernetes.io/ingress.class: contour
spec:
  tls:
  - hosts:
    - echo.127.0.0.1.xip.io
  rules:
  - host: echo.127.0.0.1.xip.io
    http:
      paths:
      - path: /
        backend:
          serviceName: echo
          servicePort: 80
---
```
> * we specify tls with a host but leave the secretName empty to use the default secret.
> * the behavior is similar to both the nginx and haproxy ingress controllers.

davecheney:
Thank you for raising this issue. 

As it stands now I don’t want to add this because it may violate some parts of the ingress spec. Can you please explain the use case for this feature.

mattalberts:
@davecheney 

[...]

PR #889, opened 2019-02-13
design: add TLSCertificateDelegation design draft

davecheney:
Updates #410 

### DIFF
--- a/design/secret-delegation.md
+++ b/design/secret-delegation.md
@@ -1,0 +1,84 @@
+# Secret delegation specficiation
+
+_Status_: Draft, in review.
+
+This document outlines a specification for delegating of permission to use the contents of a k8s Secret object from the owner's namespace to another namespace.
+
+## Goals
+
+- Permit wildcard certificates to be referenced across namespaces.
+
+## Non-goals
+
+- To implement a notion of a default, or fallback, TLS certificate available to all Ingress/IngressRoute objects.
+- To permit other classes of kubernetes objects to be referenced across namespaces.
+
+## Background
+
+Currently the Secret containing the TLS certificate must be co-located with the Ingress or root IngressRoute object referencing the secret.
+This requirement complicates deployment patterns where wildcard TLS certificates are used, specifically the use of a wildcard certificate to secure a number of subdomains.
+e.g. presenting foo.example.com using the certifiate for \*.example.com when the Ingress/IngressRoute object for foo.example.com does not share the same namespace as the Secret containing \*.example.com.
+
+This proposal introduces a specification, in the form of a simple CRD, whereby the permission to access a secret is delegated from the owning namespace to one or more other namespaces.
+
+## High-Level Design
+
+- The definition of a Secret's name is extended to recognize a namespace prefix, ie. `secretName: kube-system/wildcard-tls` means the secret `wildcard-tls` in the `kube-system` namespace.
+- A SecretDelegation CRD creates a mapping between a Secret in the owners namespace and the permission to reference that secret from a different namespace.
+
+## Detailed Design
+
+The implementation of this design is in three parts; the addition of a SecretDelegation CRD, and the modifications to the interpretation the tls stanza in Ingress and IngressRoute objects.
+
+### SecretDelegation object
+
+The SecretDelegation object provides a mapping of Secret objects from the SecretDelegation's namespace to the supplied target namespaces.
+
+```
+apiVersion: vmware.com/v1
+kind: SecretDelegation
+metadata:
+  name: wildcards
+  namespace: kube-system
+spec:
+  delegations:
+  - secretName: example-com-wildcard
+    targetNamespaces:
+    - dev-example
+    - www-example
+  - secretName: google-com
+    targetNamespaces: ["finance"]
+  - secretName: dev-wildcard
+    targetNamespaces: "*"
+```
+
+In this example permission to reference `kube-system/example-com-wildcard` is delegated to Ingress/IngressRoute objects in the `dev-example` and `www-example` namespaces, `kube-system/google-com` is delegated to Ingress/IngressRoute's in the `finance` namespace, and `kube-system/dev-wildcard` is delegated to _all_ namespaces.  
+
+### Ingress extensions
+
+To support this feature an extension to the `spec.tls.secretName` key will be recognized by Contour.
+If the `spec.tls.secretName` field contains a value with a forward slash, ie `namespace1/wildcard` the Secret object referenced will be `wildcard` in the namespace `namespace1`.
+
+If the appropriate secret delegation is in place Contour will use the fully qualified secret name as if it were in the same namespace as the Ingress object.
+
+_Note_: `kubectl` currently permits `spec.tlc.secretName` to contain a forward slash (`/`) but it is currently interpreted by Contour as part of the Secret object's name, not a separator.
+
+### IngressRoute extensions
+
+To support this feature an extension to the `spec.virtualhost.tls.secretName` key will be recognized by Contour.
+If the `spec.virtualhost.tls.secretName` field contains a value with a forward slash, ie `namespace1/wildcard` the Secret object referenced will be `wildcard` in the namespace `namespace1`.
+
+If the appropriate secret delegation is in place Contour will use the fully qualified secret name as if it were in the same namespace as the IngressRoute root object.
+
+## Alternatives Considered
+
+Alternative designs that extended the IngressRoute specification to allow referencing Secrets by name _and_ namespace were rejected because there was no way to effectively prevent anyone with the permission to construct an IngressRoute object in their own namespace from utilizing the TLS certificate from another namespace.
+While it would not be possible for the author to read the contents of the other namespace's secret--only Contour would have that permission--this would allow an attacker to present a certificate from a namespace they do not have permission to read as their own.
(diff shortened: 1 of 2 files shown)


alexbrand:
I like it! 

The only thing that is not fully clicking is developer UX. How does a developer of an application know where to find the wildcard cert? Does it have to be communicated out of band?

joshrosso:
I'll review this tonight.

davecheney:
Yes, it would have to be communicated out of band. I don’t have a better solution at the moment other than to note that to be able to use kubectl to list secrets in a namespace you have to have permission to read the secret itself (so I’ve been told) and if that is true then secret delegation isn’t making the problem worse, but it’s not improving it either.

alexbrand:
Yep, roger that. I am wondering if there is a design, however, where the developers don't need to worry about it. 

What are your thoughts on the nginx approach to this? The operator can specify a default cert, and any ingress/ingressroute that doesn't specify a secret name in the TLS block gets assigned the default cert.

davecheney:
> What are your thoughts on the nginx approach to this? The operator can specify a default cert, and any ingress/ingressroute that doesn't specify a secret name in the TLS block gets assigned the default cert.

My concerns with this is

a. why is it restricted to one certificate? why can't have two or more? 
b. in a multi team environment, how can I prevent someone from binding to the default certifcate?

design/secret-delegation.md
```diff
@@ -0,0 +1,84 @@
...
+In this example permission to reference `kube-system/example-com-wildcard` is delegated to Ingress/IngressRoute objects in the `dev-example` and `www-example` namespaces, `kube-system/google-com` is delegated to Ingress/IngressRoute's in the `finance` namespace, and `kube-system/dev-wildcard` is delegated to _all_ namespaces.  
+
+### Ingress extensions
+
+To support this feature an extension to the `spec.tls.secretName` key will be recognized by Contour.
+If the `spec.tls.secretName` field contains a value with a forward slash, ie `namespace1/wildcard` the Secret object referenced will be `wildcard` in the namespace `namespace1`.
+
+If the appropriate secret delegation is in place Contour will use the fully qualified secret name as if it were in the same namespace as the Ingress object.
+
+_Note_: `kubectl` currently permits `spec.tlc.secretName` to contain a forward slash (`/`) but it is currently interpreted by Contour as part of the Secret object's name, not a separator.
```
joshrosso: `.tlc.`->`.tls.`.

design/secret-delegation.md
```diff
@@ -0,0 +1,84 @@
...
+Currently the Secret containing the TLS certificate must be co-located with the Ingress or root IngressRoute object referencing the secret.
+This requirement complicates deployment patterns where wildcard TLS certificates are used, specifically the use of a wildcard certificate to secure a number of subdomains.
+e.g. presenting foo.example.com using the certifiate for \*.example.com when the Ingress/IngressRoute object for foo.example.com does not share the same namespace as the Secret containing \*.example.com.
+
+This proposal introduces a specification, in the form of a simple CRD, whereby the permission to access a secret is delegated from the owning namespace to one or more other namespaces.
+
+## High-Level Design
+
+- The definition of a Secret's name is extended to recognize a namespace prefix, ie. `secretName: kube-system/wildcard-tls` means the secret `wildcard-tls` in the `kube-system` namespace.
+- A SecretDelegation CRD creates a mapping between a Secret in the owners namespace and the permission to reference that secret from a different namespace.
```
joshrosso: `owners`->`owner's`

design/secret-delegation.md
```diff
@@ -0,0 +1,84 @@
...
+While it would not be possible for the author to read the contents of the other namespace's secret--only Contour would have that permission--this would allow an attacker to present a certificate from a namespace they do not have permission to read as their own.
+In the case of a wildcard certificate this is benficial--it's actually what we want--but also opens up the possibility, when combined with DNS spoofing, of presenting an alternate site using the _real_ SSL certificate, leading to cookie hijacking and MITM attacks.
+	
+## Security Considerations
+
+Delegation is a necessary security measure because it allows namespace owners to explicitly delegate the permission to reference secrets in their namespace without granting permission to actually _read_ the contents of the certificate.
+
+Permission to use secret delegation is restricted via RBAC and by default is not enabled.
+To create a secret delegation CRD the author must have permission to create the secret delegation object in the souce Namespace.
```
joshrosso: `souce`->`source`

rosskukulinski:
I like the approach.  We should also document the impact on IngressRoute `status` field.  If an IR references a secret in another namespace, we need to inform via `status` that permission is not allowed.  Same for a secret that doesn't exist.

@mattmoyer I'd appreciate your take on this PR from a security perspective.

design/secret-delegation.md
```diff
@@ -0,0 +1,84 @@
...
+
+The implementation of this design is in three parts; the addition of a SecretDelegation CRD, and the modifications to the interpretation the tls stanza in Ingress and IngressRoute objects.
+
+### SecretDelegation object
+
+The SecretDelegation object provides a mapping of Secret objects from the SecretDelegation's namespace to the supplied target namespaces.
+
+```
+apiVersion: vmware.com/v1
+kind: SecretDelegation
```
mattmoyer: I would try to name this so it's clear that you're delegating for the purposes of ingress only, not necessarily for arbitrary uses of the Secret.

In other words, granting a user in another namespace permission to serve traffic behind a TLS certificate with a particular name isn't the same as allowing them to access the secret directly (e.g., by mounting it into a Pod). This new object is meant to allow the former but not the latter.

design/secret-delegation.md
```diff
@@ -0,0 +1,84 @@
...
+metadata:
+  name: wildcards
+  namespace: kube-system
+spec:
+  delegations:
+  - secretName: example-com-wildcard
+    targetNamespaces:
```
mattmoyer: It might be worth documenting how `targetNamespaces` interacts with how IngressRoute describes restricted delegation of particular vhosts.

If I have a wildcard certificate for `*.example.com`, I might intend to allow it to be referenced by the `foo` namespace, but _only_ for the `foo.example.com` vhost. I believe that latter restriction could be enforced with the existing IngressRoute mechanisms but it's not called out explicitly here.

design/secret-delegation.md
davecheney: Yup, Joe bought that up as well. I think the easiest way to make this clearer is to call this proposal "TLS Certificate Delegation"

design/secret-delegation.md
davecheney: > It might be worth documenting how `targetNamespaces` interacts with how IngressRoute describes restricted delegation of particular vhosts.

I'd like that to be orthogonal. I have an alternative proposal that creates a coupling between wildcards and delegation.

> If I have a wildcard certificate for `*.example.com`, I might intend to allow it to be referenced by the `foo` namespace, but _only_ for the `foo.example.com` vhost. I believe that latter restriction could be enforced with the existing IngressRoute mechanisms but it's not called out explicitly here.

If you want to do that, and you don't trust the owner of foo.example.com with your certificate, then, the IngressRoute pattern of placing the root vhost for foo.example.com in one namespace, along with the certificate (a wildcard certificate can be used for a named subdomain without issue), and then delegating the responsibility for all the paths under foo.example.com to another namespace via IngressRoute's `prefix: "/"` delegation.

design/tls-certificate-delegation.md
```diff
@@ -0,0 +1,86 @@
...
+For example, presenting foo.example.com using the certifiate for \*.example.com when the Ingress/IngressRoute object for foo.example.com does not share the same namespace as the Secret containing \*.example.com.
+
+This proposal introduces a specification, in the form of a simple CRD, whereby the permission to access the contents of a secret containing a TLS certificate is delegated from the owning namespace to one or more other namespaces.
+
+## High-Level Design
+
+- The definition of a Secret's name is extended to recognize a namespace prefix, ie. `secretName: kube-system/wildcard-tls` means the secret `wildcard-tls` in the `kube-system` namespace.
+- A `TLSCertificateDelegation` CRD grants the permission to reference the contents of a Secret in the owner's namespace to an Ingress controller operating in the context of an Ingress or IngressRoute object from a different namespace for the purpose of retrieving the TLS certicate.
+
+## Detailed Design
```
rosskukulinski: @davecheney could you also provide additional details on how the `status` field in IngressRoute would be changed in the case of missing (orphaned?) delegation or any other failure conditions.

jcrowthe:
This topic (sharing a wildcard cert) has been on my mind, and has become a high priority for my team. In fact, we were designing a solution and we were planning on submitting a similar proposal with a slightly different implementation. However this proposal has the benefit of stronger RBAC around delegation, and so is a better fit.

Regarding #897, I believe it's subdomain delegation method doesn't actually address the TLS issue. As en example use case, my team deploys hundreds of services behind a single wildcard certificate. In the subdomain delegation model, we would be required to modify a root IngressRoute for each new service -- which would require administrator involvement for each new service onboarding. This is only different from the existing functionality (where we provide a dedicated root IngressRoute object for each new service) in that we would have only one IngressRoute root object rather than hundreds. In my opinion, subdomain delegation is relevant to the shared TLS cert problem but doesn't actually solve it. This TLSCertificateDelegation CRD model, however, directly addresses the shared TLS cert problem by providing the ability to delegate usage of TLS secrets to other namespaces on a cert by cert, and namespace by namespace basis.

In the end I believe both #897 and this proposal not only can but should coexist. Outside of the TLS context, #897 is an improvement to delegation, while inside the TLS context, this proposal is a direct solution to the problem.

I'll make a brief note on #897 as well.

PR #897, opened 2019-02-18
design: add Subdomain Delegation draft

davecheney:
Updates #410

Subdomain delegation is an alternative to TLS Certificate Delegation (#889).
Subdomain delegation introduces the notion of a "root" IngressRoute
object which holds the top level domain definition, ie, .mycompany.com,
along with a TLS stanza block that applies to all delegated subdomains
unless specifically overridden.

### DIFF
--- a/design/subdomain-delegation.md
+++ b/design/subdomain-delegation.md
@@ -1,0 +1,65 @@
+# Subdomain Delegation
+
+_Status_: Draft, in review.
+
+This document outlines an addition to the [IngressRoute][0] specification to permit subdomains to inherit domain wide parameters such as wildcard TLS certificates without exposing that configuration to the specific subdomain.
+
+## Goals
+
+- Permit a number of subdomains to be served using a wildcard TLS certificates without requiring the IngressRoute record for that subdomain to be a part of the namespace holding the TLS wildcard certificate's Secret.
+
+## Non-goals
+
+- To support hosting a Kubernetes service for a wildcard domain (eg, \*.whatever.net) itself. 
+
+## Background
+
+Currently the Secret containing the TLS certificate must be co-located in the same namespace as the Ingress or root IngressRoute object referencing that secret.
+This requirement complicates deployment patterns where wildcard TLS certificates are used, specifically the use of a wildcard certificate to secure a number of subdomains where the Ingress/IngressRoute records for those subdomains do not share the same namespace as the wildcard TLS certificate.
+For example, presenting foo.example.com using the certificate for \*.example.com when the Ingress/IngressRoute object for foo.example.com does not share the same namespace as the Secret containing \*.example.com.
+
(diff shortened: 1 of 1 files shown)


design/subdomain-delegation.md
```diff
@@ -0,0 +1,65 @@
...
+
+Currently the Secret containing the TLS certificate must be co-located in the same namespace as the Ingress or root IngressRoute object referencing that secret.
+This requirement complicates deployment patterns where wildcard TLS certificates are used, specifically the use of a wildcard certificate to secure a number of subdomains where the Ingress/IngressRoute records for those subdomains do not share the same namespace as the wildcard TLS certificate.
+For example, presenting foo.example.com using the certificate for \*.example.com when the Ingress/IngressRoute object for foo.example.com does not share the same namespace as the Secret containing \*.example.com.
+
+This proposal introduces a new type of delegation, subdomain delegation, which permits the root IngressRoute object to define a TLS certificate to be used for a set of IngressRoute resources denoted as subdomains.
+
+## High-Level Design
+
+- `spec.virtualhost.fqdn` can be treated as a domain _suffix_, if the `spec.delegate` key is present.
```
rosskukulinski: spec.delegate and spec.routes are mutually exclusive, correct?  You cannot have a domain _suffix_ and spec.routes.

design/subdomain-delegation.md
```diff
@@ -0,0 +1,65 @@
...
+
+```
+apiVersion: contour.heptio.com/v1beta1
+kind: IngressRoute
+metadata:
+  name: domain-root
+  namespace: mycompany
+spec:
+  virtualhost:
+    fqdn: .mycompany.com
```
rosskukulinski: Is `foo.bar.mycompany.com` an acceptable subdomain?  I wanted to be clear on the definition of `suffix` in this context

rosskukulinski:
I particularly like this spec as the subdomain delegation is likely a common use-case and this allows to having _one_ IR for a domain suffix rather than one IR _per subdomain_

jcrowthe:
I believe this proposal should be separate from #889. Specifically, I don't believe this proposal actually addresses the shared TLS cert issue, but is only side-along to the problem. I've commented on the other PR with more details. See https://github.com/heptio/contour/pull/889#issuecomment-465302084

design/subdomain-delegation.md
davecheney: That's a good point. Practically if this feature is used to delegate TLS subdomains it can only really be used for a single level. But from the POV of a data model it seems weird to add that restriction.

rosskukulinski: I think it's also interesting to work through chained delegations like we can do in IngressRoute

davecheney:
Deferred til 0.11

Issue #862, opened 2019-01-25
Encrypt gRPC traffic between Contour and Envoy


scottslowe:
**Feature we'd like to see added:**
When using TLS termination for Ingress/IngressRoute objects with Contour, certificates have to be passed to Envoy over the gRPC connection. Given the sensitive nature of TLS certificates, some security-conscious customers want the gRPC connection between the Contour pods and the Envoy pods to be encrypted (cc @stevesloka)

**Environment:**
This would apply to potentially any environment running Contour.

- Contour version: N/A
- Kubernetes version: (use `kubectl version`): Any supported
- Kubernetes installer & version: Any supported
- Cloud provider or hardware configuration: Any supported
- OS (e.g. from `/etc/os-release`): Any supported

davecheney:
Yes, I want to add this as well. It is a blocker to moving away from the
default contour and envoy in the same pod deployment model.

vaamarnath:
I am willing to pick this issue. Is this work something that a newbie can pick to work on?

From my understanding of contour architecture and context provided here, the work involved is to add mTLS (maybe just TLS?) for communication between contour and envoy. I can put in a design document if my understanding of the enhancement request is correct.

davecheney:
Yes please, a design document would be great. Or even a sketch of the major pieces you want to change here in this issue will also be fine. 

One of the goals of this ticket is not just to encrypt the transmission between envoy and contour, but to add a layer of authorisation. At the moment, if you can connect to contour’s endpoint you can ask it for any tls secret it has learned. This isn’t great, which is why by default we only bind to the container loopback. 

The big things your design should cover is. 

- how to enable this feature in the bootstrap config we pass to envoy on startup. If this is optional or always on, etc
- how to authenticate the envoy connecting too contour. 
- how to roll and invalidate those credentials. 
- how to secure those credentials

vaamarnath:
Thanks for your reply @davecheney. Before I start putting down my approach into a design document, I would like to run the approach with you at a high level. Here is what I am thinking now:

1. Use mTLS between envoy and contour to have authorisation. 
2. TLS certs can be user provided (is this interruptive to deployment experience for an user?). Another option is to have these certificates generated by an init container as part of the bootstrapping process. The certificates can be signed by an intermediary CA which itself is signed by Kubernetes CA. 
[...]

Issue #898, opened 2019-02-18
Implement Secret Discovery Service 


vaamarnath:
Implement Secret Discovery Service in Contour.

This enhancement request came out of the discussion on #862 for supporting rotating and invalidating certificates using SDS. More info about SDS is available here: https://www.envoyproxy.io/docs/envoy/latest/configuration/secret.

khgandhi:
@vaamarnath we have a requirement to have SDS implementation in contour as well. Do you have a design that you can share with the details ?

vaamarnath:
@khgandhi The current requirement for SDS came in through #862. There is no design available for this issue at this point in time. 

\cc @davecheney

bmadhavan:
@davecheney If you are ok, I can pick up this task. I can come up with a rough design doc and a list of possible changes in a couple of days.

davecheney:
@bmadhavan sure, feel free to discuss the design here or in a design document. Please note that I will be travelling until the 1st of April and may not be able to reply during that time.

bmadhavan:
ok. Sure.

bmadhavan:
@davecheney I have created an initial design draft for SDS support. Can you please take a look.

davecheney:
Due to the need to ship Contour 0.11 to address a security issue in Envoy 1.9.0 this has been bumped to the 0.12 milestone.

davecheney:
With #901 falling off the 0.12 milestone @vaamarnath and I have talked and the plan is to implement the basic SDS plumbing and we'll revisit the integration points with the rest of Contour later. 

I was going to sit down and write down the steps to add SDS to `internal/grpc` but it turns out we already have 99% of the design from @bmadhavan in #960. I've polished that up and you can find [here](https://github.com/heptio/contour/blob/master/design/sds-design.md).

I'm not going to mark this issue as 0.12 as @vaamarnath nor I can commit to taking this one by the 0.12 deadline, but consider the design for adding SDS up for grabs. 

Note to implementors: please work incrementally, one PR per high level design point. Please don't send one giant ass PR for the whole thing, that's not cool. You should also add some tests to `internal/e2e` which are really the func tests of `internal/grpc`.

Thanks

Dave

mattalberts:
@vaamarnath @davecheney Thanks for the notes.  I've been building up the full solution over the last few days, adding tests, and verifying functionality.  I'll split up the change set into a few PRs, mirroring the suggestion to provide 1 per major design point.

davecheney:
[...]

Issue #881, opened 2019-02-07
Split Contour and Envoy containers into their own pods


davecheney:
This ticket describes a long term plan to change the default mechanism for deploying Contour.

Currently the default deployment for Contour colocates a contour process and an envoy container in the same pod. This is convenient, envoy can talk over the container loopback to contour for its configuration, but it is not the only way to deploy contour. For instance Gimbal deploys contour and envoy in their own pods; one as a deployment, the other as a daemonset. This is the mechanism I want to move towards as the default way of deploying Contour. Specifically

- Contour is deployed as a classic deployment
- Envoy is deployed as a daemonset

This change would deliver the following benefits

- We no longer need one Contour per Envoy, this will reduce load on the API server. Ideally I would like to deploy only one Contour and rely on the kubelet's restart mechanism to handle upgrades and outages of the control plane. If this were achieved then it would avoid the need for leader election when updating status, and possibly mitigate status update loops like #854 
- Upgrades and configuration changes to Contour's deployment would not affect the Envoy pods, this would address issues where a change to Contour's flags causes a rolling drop of all traffic in flight across envoy's because the kubelet kills all the containers in the pod during reload.
- If configuration changes to the control plane no longer caused data plane outages, this would reduce the need to move contour's configuration into a configmap. The configmap being a way to hold changing configuration data which contour would have to watch and learn to reconfigure itself without exiting. This means contour doesn't need to watch config maps, and we avoid having to teach contour to restart itself internally, a procedure which will surely lead to bug reports in the field. 
- The crash of a contour process would no longer cause its sibling enovy to be killed.

This proposal is not without downside

- Currently the traffic between envoy and contour is not secured as we rely on the container loopback address. This would require an encrypted and authenticated channel between contour and envoy. #862 

This change would not remove the ability for users to deploy contour colocated with envoy in the same pod. We're just not going to push it as the default option any more.

so0k:
Based on the docs we are running contour in it's own deployment and using the Kube injected Environment variables for envoy to get xds config from contour. As highlighted in this issue, it seems the downside to this deployment is unsecured communication between envoy and contour in the kube cluster... this is not really highlighted in the docs

perhaps I could do a PR to the docs to add a caveat regarding the communication?

side note, couldn't we use mTLS from istio to secure communication between envoy and contour until this feature gets added?

davecheney:
Rather than fixing the docs I want to make the communication between contour and envoy secure. 

This is a high priority item for me.

[...]

PR #901, opened 2019-02-19
design: add secure communication between contour, envoy

vaamarnath:
Updates #862

### DIFF
--- a/design/secure-communication-design.md
+++ b/design/secure-communication-design.md
@@ -1,0 +1,88 @@
+# Secure Communication between Contour and Envoy
+
+_Status_: Draft, in review.
+
+This document outlines how to secure the communication between Contour
+and Envoy containers.
+
+## Goals
+
+- Achieve secure communication between Contour and Envoy
+- Authenticate Envoy connecting to Contour
+- Rotating and invalidating the credentials
+
+## Non-goals
+
+- Support for dynamic credentials
+- Support for use of Secret Discovery Service for credential management
+
+## Background
+
+Currently, Contour and Envoy containers are co-located within a pod.
+Current security model is relying on the fact that Contour exposes the management
+server on loopback interface which prevents any unauthorized access from outside
+the pod.
+
+With the need to move away from the [current deployment pattern of co-locating
+Contour and Envoy][0], it is necessary to secure the communication channel between
+the containers.
+
+## High level design
+
+- Use mutual TLS to secure communication channel between Contour and Envoy
+- Bootstrap phase of Contour will generate necessary TLS certificates
+
+## Detailed design
+
+### Default options for secure communication
+
+By default, mutual TLS is enabled to secure the communication between Contour and Envoy
+containers. The default `contour bootstrap` will generate the bootstrap configuration that
+enables secure communication between the containers. If `contour bootstrap` is passed with
+the parameter `--disable-tls`, the bootstrap configuration generated will make use of insecure
+communication between the containers.
+
+### Bootstrapping process
+
+This bootstrap phase is independent of `contour bootstrap` and uses `contour bootstrap-tls`.
+`contour bootstrap-tls` is run as an init container to bootstrap the TLS configuration.
+The bootstrap phase will create three pairs of the TLS certificates. The details about
+each pair of certificates are mentioned below:
+1. The first pair is used by Contour to serve HTTPS.
+2. The second pair is used by Envoy to authenticate with Contour.
+3. The last pair serves as the CA certificate which signs the above certificates.
+
+Each of these TLS certificates will be stored as Kubernetes secrets in namespace `heptio-contour`. The
+secrets will be named as `contour-tls`, `envoy-tls` and `contour-ca`, respectively. This ensures that
+even when the `contour` and `envoy` containers are not part of the same pod, the certificates are
+available for consumption by each of those.
+
+Note: In order for Contour to create the Kubernetes secrets in the `heptio-contour` namespace, the
+existing RBAC needs to be enhanced to grant permissions to `create` secrets in the namespace.
+
+### Setting up secure communication channel
+
+1. Contour container will pick the certificates from the Kubernetes secret `contour-tls` which is
+mounted at `/etc/tls/contour` and pick `contour-ca` mounted at `/etc/tls/ca` to start listening for
+HTTPS traffic.
+2. The bootstrap configuration yaml of Envoy will have [`tls_context`][1] added as part of `cluster` object.
+Envoy container will pick the certificates from the Kubernetes secret `envoy-tls` which is mounted at
+`/etc/tls/envoy` and pick `contour-ca` mounted at `/etc/tls/ca` to connect to Contour as configured by the
+bootstrap configuration file.
+
+### Rotation and invalidation of credentials
+
+This design assumes that to rotate or invalidate the credentials, Contour has to be redeployed which will
+regenerate a new set of certificates.
+
+## Future enhancements
+
+As discussed in [issue 881][2], this design uses static credentials and to support dynamic
+credentials, [Secret Discovery Service][3] needs to be implemented by Contour. This work
+is tracked [here][4].
+
+[0]: https://github.com/heptio/contour/issues/881
+[1]: https://www.envoyproxy.io/docs/envoy/v1.9.0/intro/arch_overview/ssl.html#enabling-certificate-verification
+[2]: https://github.com/heptio/contour/issues/862#issuecomment-464601450
+[3]: https://www.envoyproxy.io/docs/envoy/v1.9.0/configuration/secret
+[4]: https://github.com/heptio/contour/issues/898


design/secure-communication-design.md
```diff
@@ -0,0 +1,79 @@
...
+containers. If `contour bootstrap` is passed with the parameter `--disable-tls`,
+the communication between Contour and Envoy will not be secured.
+
+The bootstrap phase will create three pairs of the TLS certifications. The details about
+each pair of certificates are mentioned below:
+1. The first pair is used by Contour to serve HTTPS (named `contour.crt`, `contour.key`)
+2. The second pair is used by Envoy to authenticate with Contour (named `envoy.crt`, `envoy.key`)
+3. The last pair serves as the CA certificate which signs the above certificates (named `ca.crt`, `ca.key`)
+
+These certificates will be placed under `/config/tls/`. The bootstrap configuration yaml
```
davecheney: minor k8s nit, the /config volume is traditionally mounted as a tmpfs. I'm not sure if it it would be a good idea to have a second volume, say `/tls/` or `/certs/`, rather than potentially mounting `/config/tls` on top of `/config`. 

/cc @alexbrand @stevesloka @mauilion 

I think it might be safer to assume the certificates are in one volume and the config in another. It costs very little from a resource perspective to do this, and then you don't need to worry about customers who don't use `contour bootstrap`.

design/secure-communication-design.md
```diff
@@ -0,0 +1,79 @@
...
+of Envoy will have [`tls_context`][1] added as part of `cluster` object.
+
+### Setting up secure communication channel
+
+1. Contour container will pick the certificates `contour.crt`, `contour.key` and `ca.crt` from
```
davecheney: Is this going to be hard coded in the config passed to the envoy container?

vaamarnath: I have updated the design to make use of Kubernetes secrets instead of using the TLS certificate as files.

design/secure-communication-design.md
vaamarnath: With the change to make use of Kubernetes secrets, the TLS certs are not mounted at `/etc/tls/` and not using `/config/` anymore.

vaamarnath:
@davecheney I have updated the design document with how the bootstrap process creates the TLS certificates and stores those in Kubernetes as secrets. I have suggested to use `contour bootstrap-tls` to bootstrap these TLS certificates independent of the `contour bootstrap`. 

PTAL of the updated design doc.

design/secure-communication-design.md
```diff
@@ -0,0 +1,88 @@
...
+the parameter `--disable-tls`, the bootstrap configuration generated will make use of insecure
+communication between the containers.
+
+### Bootstrapping process
+
+This bootstrap phase is independent of `contour bootstrap` and uses `contour bootstrap-tls`.
+`contour bootstrap-tls` is run as an init container to bootstrap the TLS configuration.
+The bootstrap phase will create three pairs of the TLS certificates. The details about
+each pair of certificates are mentioned below:
+1. The first pair is used by Contour to serve HTTPS.
```
stevesloka: I'm not sure generating certs that envoy uses to serve TLS is in the scope of this issue. Those should be part of an `Ingress` or `IngressRoute` object.

design/secure-communication-design.md
```diff
@@ -0,0 +1,88 @@
...
+mounted at `/etc/tls/contour` and pick `contour-ca` mounted at `/etc/tls/ca` to start listening for
+HTTPS traffic.
+2. The bootstrap configuration yaml of Envoy will have [`tls_context`][1] added as part of `cluster` object.
+Envoy container will pick the certificates from the Kubernetes secret `envoy-tls` which is mounted at
+`/etc/tls/envoy` and pick `contour-ca` mounted at `/etc/tls/ca` to connect to Contour as configured by the
+bootstrap configuration file.
+
+### Rotation and invalidation of credentials
+
+This design assumes that to rotate or invalidate the credentials, Contour has to be redeployed which will
```
stevesloka: What happens if I run 2 instances of Contour in the cluster? They will each overwrite the other's secrets. Maybe we need to have Contour check if the secrets exist, then they don't auto-generate. This would also allow users to pass their own certs if they wanted without using the auto-generated.

design/secure-communication-design.md
```diff
@@ -0,0 +1,88 @@
...
+secrets will be named as `contour-tls`, `envoy-tls` and `contour-ca`, respectively. This ensures that
+even when the `contour` and `envoy` containers are not part of the same pod, the certificates are
+available for consumption by each of those.
+
+Note: In order for Contour to create the Kubernetes secrets in the `heptio-contour` namespace, the
+existing RBAC needs to be enhanced to grant permissions to `create` secrets in the namespace.
+
+### Setting up secure communication channel
+
+1. Contour container will pick the certificates from the Kubernetes secret `contour-tls` which is
```
stevesloka: Same as the previous comment, each Ingress or IngressRoute would specify the tls secret to use.

design/secure-communication-design.md
```diff
@@ -0,0 +1,88 @@
...
+mounted at `/etc/tls/contour` and pick `contour-ca` mounted at `/etc/tls/ca` to start listening for
+HTTPS traffic.
+2. The bootstrap configuration yaml of Envoy will have [`tls_context`][1] added as part of `cluster` object.
+Envoy container will pick the certificates from the Kubernetes secret `envoy-tls` which is mounted at
+`/etc/tls/envoy` and pick `contour-ca` mounted at `/etc/tls/ca` to connect to Contour as configured by the
+bootstrap configuration file.
+
+### Rotation and invalidation of credentials
```
stevesloka: If we regen new certs, how do we roll that change out to the Envoy instances? In the deployment model of a colocated Contour with Envoy, this is easy, but when you split apart Envoy from Contour, you'll have a separate daemonset to roll the Envoy pods.

design/secure-communication-design.md
stevesloka: I like having the config items separate. This also allows us to mount the secrets and watch for updates (in the future), without restarting Contour or Envoy.

design/secure-communication-design.md
```diff
@@ -0,0 +1,88 @@
...
+By default, mutual TLS is enabled to secure the communication between Contour and Envoy
+containers. The default `contour bootstrap` will generate the bootstrap configuration that
+enables secure communication between the containers. If `contour bootstrap` is passed with
+the parameter `--disable-tls`, the bootstrap configuration generated will make use of insecure
+communication between the containers.
+
+### Bootstrapping process
+
+This bootstrap phase is independent of `contour bootstrap` and uses `contour bootstrap-tls`.
+`contour bootstrap-tls` is run as an init container to bootstrap the TLS configuration.
```
stevesloka: Is this a separate init container (so a 2nd one)? It might be best to just have `bootstrap` do both tasks, creating the secrets and then the bootstrap config. I'm not sure why we'd want to create certs outside of the bootstrap unless this was a CLI command to set up a cluster (i.e. run from an admin workstation and not inside an init container).

design/secure-communication-design.md
```diff
@@ -0,0 +1,88 @@
...
+The bootstrap phase will create three pairs of the TLS certificates. The details about
+each pair of certificates are mentioned below:
+1. The first pair is used by Contour to serve HTTPS.
+2. The second pair is used by Envoy to authenticate with Contour.
+3. The last pair serves as the CA certificate which signs the above certificates.
+
+Each of these TLS certificates will be stored as Kubernetes secrets in namespace `heptio-contour`. The
```
stevesloka: There might be a chicken & egg problem here. If we utilize Kubernetes secrets, the pod won't start until the secret exists. We may need to do something like how Sonobuoy does it by having a `contour gen` command which would give you the secrets that would be applied to cluster.

design/secure-communication-design.md
```diff
@@ -0,0 +1,88 @@
...
+containers. The default `contour bootstrap` will generate the bootstrap configuration that
+enables secure communication between the containers. If `contour bootstrap` is passed with
+the parameter `--disable-tls`, the bootstrap configuration generated will make use of insecure
+communication between the containers.
+
+### Bootstrapping process
+
+This bootstrap phase is independent of `contour bootstrap` and uses `contour bootstrap-tls`.
+`contour bootstrap-tls` is run as an init container to bootstrap the TLS configuration.
+The bootstrap phase will create three pairs of the TLS certificates. The details about
```
stevesloka: What are the params on the certs that are created? How long are they good for? Do we need to worry about adding DNS to the SANs of the cert. Right now when we deploy a split model, we grab the service IP from Kubernets ENV var to know how Envoy will find Contour's gRPC endpoint. If it's over tls we'll need to figure out how to connect to Contour over dns (I had issues last time doing this), and we'd need to generate proper SANs on the cert.

design/secure-communication-design.md
vaamarnath: I am not sure if I understand your review comment here. Let me try to drive my idea here. It is to serve Contour the xDS APIs over HTTPS and Envoy talks to Contour over mTLS. To setup the mTLS communication, I am proposing to create a CA cert and get two pairs of the certs signed with the CA cert. One pair of the cert is used by Contour to serve HTTPS and Envoy uses the other pair to connect to Contour over mTLS.

design/secure-communication-design.md
vaamarnath: Oh yeah, I need to consider this as well in the design. I will work on it.

design/secure-communication-design.md
vaamarnath: Please see my previous comment.

design/secure-communication-design.md
vaamarnath: Yes, we will have to document how to handle the case where Contour and Envoy are not colocated. As per the initial discussion I had with @davecheney on #862, we agreed that we can implement and address this as part of #881 where it will be required to maybe have SDS in place (#898).

design/secure-communication-design.md
vaamarnath: I believe there are use cases where the users might not run `contour bootstrap` and provide the bootstrap configuration separately. I thought it would be a better idea to have it a separate init container for better control on these cases.

design/secure-communication-design.md
vaamarnath: As far as I know, as the init containers brings up the Kubernetes secrets, the pod will be able to start.

design/secure-communication-design.md
vaamarnath: Yes, I will update the design doc with more details about the cert expiration and the SANs.

And about the split model cases, I thought will be worked on as part of #881.

design/secure-communication-design.md
stevesloka: You are 100% correct, please ignore my comment, I forgot this was mTLS and not just TLS.

design/secure-communication-design.md
stevesloka: 👍

design/secure-communication-design.md
stevesloka: I don't think SDS would solve this issue 100%. I think there's still a timing issue that needs a blocker somehow. Please see comment below on mounting secret from init container. 

Also, we already have a split deployment example as well as Gimbal uses this model by default.

design/secure-communication-design.md
stevesloka: My initial thought was that in the contour yaml, we'd reference a secret. This would allow any instance of Contour to block from starting until the secret exists. If this is the case, I think we need some external process to generate the secret which Contour blocks on starting until that secret is available. The Contour deployment resource needs to reference a secret, it won't start until that secret is available. 

If Contour doesn't mount the secret as a volume, then I'm still not sure how multiple instances of Contour can coordinate together and not overwrite each other. It would work if you bring up a single instance of Contour at a time, but we can't guarantee folks will do that.

design/secure-communication-design.md
stevesloka: As commented previously, I think we need to think about the split model in this design.

design/secure-communication-design.md
stevesloka: It would take a larger discussion, but we could look at pulling in some of the kubebuilder bits to simply the CRD work, as well as give us leader election. I haven't used before though so can't speak to how well it works: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/manager.go#L104-L114

stevesloka: There may be a way we can do leader election with a configmap lock (https://github.com/kubernetes/client-go/blob/master/examples/leader-election/main.go). 

Looks like that example is broken, I'm working to fix it now.

vaamarnath:
I apologise for not being able to work on this issue for about a week. I am intending to work on this in next few days, and I would like to know if @davecheney has any additional comments to the current design.

design/secure-communication-design.md
vaamarnath: Sure, I will consider this for this design.

design/secure-communication-design.md
vaamarnath: Like I had mentioned before, we can have the init container to create the necessary secrets for us. I agree that we might want this to performed by one of the pods and not all of the pods in split deployment scenarios. This can definitely make use of the leader election provided by the Kubernetes client-go. 

For reference, for how init containers can create secrets: https://github.com/kelseyhightower/certificate-init-container.

design/secure-communication-design.md
vaamarnath: @stevesloka you are right that SDS will not solve this issue. I did a bit of reading and this is what I found
> The connection between Envoy proxy and SDS server has to be secure. One option is to run the SDS server on the same host and use Unix Domain Socket for the connection. Otherwise it requires mTLS between the proxy and SDS server. In this case, the client certificates for the SDS connection must be statically configured.

Reference: https://www.envoyproxy.io/docs/envoy/v1.10.0/configuration/secret

design/secure-communication-design.md
vaamarnath: I have addressed this already. I hope that answers your question.

vaamarnath:
I will continue working on the design as I find more spare time this week and it would be great if we agree on the approach for statically configuring the mTLS between contour and envoy. My suggestion is to generate the necessary TLS certificates using an init container. I am working on how to handle the cases of split deployment scenarios which would need consensus between the different contour pods on who is responsible for generating the TLS certificates.

vaamarnath:
Thank you for being patient with me on this design. 

> The problem with this proposal is the difficulty generating, then distributing the bootstrap client/server certs so that Envoy can communicate with a central Contour SDS server to receive additional certificates. This gives rise to questions like
> 
> * how to we generate those certs--maybe an init container
> * ok, if they are the key to obtaining all the other keys, how to we secure and rotate those keys, etc.
> 
I agree this is not a great user experience.

> The solution to both of those problems appears to me to be moving the SDS server from contour back to the envoy container as a sidecar process.
> 
> Would this work.

This is an interesting idea to purse to achieve a better user experience.

> 
> a. we add a new mode to contour, say, `contour servesds` which runs an SDS _only_ server. It is hard coded to talk only on localhost, or perhaps a unix domain socket -- whichever is more reasonable for a k8s pod.
> b. Instead of running a full contour in the same pod as each envoy, we deploy on contour in servesds mode with each envoy pod.
> 

To make sure that we are on the same page, I am repeating what you said here. The suggestion is to have a new way to run just server SDS server of Contour over unix domain socket which resides alongside with every instance of Envoy.

> This isn't a perfect solution, the secrets are still being read by an serviceAccount that the Envoy pod is running in, but at least we can move away from the Envoy + Contour co-habitation model.
> 
> How does that sound?

This idea is wonderful. I understand with the minimal SDS server only mode of Contour alongside with the Envoys makes it easier to push the TLS certificates to each of the Envoys. But, does this approach solve how the corresponding TLS certificates are pushed to Contour pods? Are you suggesting that when the certificates need to be rotated, the Contour pods initiate the regeneration of the TLS certificates and store the certificates as Kubernetes secrets? The minimal SDS server only Contour containers pick this change by watching the corresponding Kubernetes secrets and push the new TLS certificates to the Envoys? I guess this should solve the issue at hand. Right?

> Do you have time to incorporate this change? If not, that's fine, I can take a pass next week.

I am happy to work on this in next few days.

davecheney:
> To make sure that we are on the same page, I am repeating what you said here. The suggestion is to have a new way to run just server SDS server of Contour over unix domain socket which resides alongside with every instance of Envoy.

Yes, I suggest one of

a. `contour serve` starts up and writes a Secret containing the CA and client key, k8s writes this secret to a temporary mount, `contour bootstrap` configures Envoy to read the CA and key on startup so that Envoy can communicate with `contour serve`.
b. Similar to a, `contour serve` writes the ca and key, a sidecar process on each envoy serves the ca and key from the Secret object using the SDS protocol over a loopback socket (or unix domain socket if they are reasonable to use under k8s). `contour bootstrap` configures Envoy to use this sidecar SDS process to obtain the client key to connect to `contour serve`.

> But, does this approach solve how the corresponding TLS certificates are pushed to Contour pods?

I'm not sure what you mean. The idea I have is `contour serve` generates a CA and client/server key pair. It will serve gRPC using the server key and store the client key in a k8s Secret. Envoy directly (via a filesystem mount) or indirectly, via an SDS sidecar, retrieves the client key and connects to Contour. 

The main `contour serve` process will not implement SDS. There is no need. Contour will continue to serve key material for TLS listeners inline in LDS responses.

davecheney:
@vaamarnath a further simplification that occurred to me

The underlying problem is not that we want to add SDS support to Contour. The underlying problem isn't #862 either. The problem is currently if Contour is listening on a public address, anyone can connect to it and extract private keys via LDS.

I propose that rather than encrypting the communication between Envoy and Contour, we stop sending secrets via LDS. Instead we implement SDS as a sidecar on each Envoy pod. That will mean we don't transmit secrets in the clear via LDS between Contour and Envoy, we only communicate the name of a reference which can be used to obtain a secret via SDS over a loopback connetion from a sidecar running along side each envoy pod.

vaamarnath:
> The underlying problem is not that we want to add SDS support to Contour. The underlying problem isn't #862 either. The problem is currently if Contour is listening on a public address, anyone can connect to it and extract private keys via LDS.

Oh this is interesting. I always thought that, it was the objective to have encrypted traffic between Contour and Envoy.

> I propose that rather than encrypting the communication between Envoy and Contour, we stop sending secrets via LDS. Instead we implement SDS as a sidecar on each Envoy pod. That will mean we don't transmit secrets in the clear via LDS between Contour and Envoy, we only communicate the name of a reference which can be used to obtain a secret via SDS over a loopback connetion from a sidecar running along side each envoy pod.

This approach seems reasonable and I can give a shot to writing down the design for this new approach. Sounds good @davecheney?

davecheney:
Given there is no immediate way forward on a the design after several attempts I've decided to remove the 0.12 milestone from this issue.

After a bit of discussion with @vaamarnath we feel that doing the groundwork by implementing SDS support in `internal/grpc` is a useful step forward as we will need that for this design, and my spider senses tell me we may make use of it for #813 and #499.

@vaamarnath is going to track SDS support in #898

vaamarnath:
Closing this PR in favour of https://github.com/heptio/contour/pull/1186.

Issue #437, opened 2018-06-12
Feature request: CORS support


glerchundi:
As far as I can see, in v2 API, it can be implemented only as a http filter: https://github.com/envoyproxy/data-plane-api/blob/master/envoy/config/filter/network/http_connection_manager/v2/http_connection_manager.proto.

In order to have secure enough defaults I think this should be disabled by default and let people enable explicitly with a new command-line parameter.

The implementation is fairly easy I can send a PR if you feel this can get into contour.

WDYT @davecheney? 

---
[@rosskukulinski copy/paste] Proposed spec:

[...]
    tls:
      # required, the name of a secret in the current namespace
      secretName: google-tls
      # other properties like cipher suites may be added later
    # if present describes the CORS policy.
    cors:
      # Specifies the origins that will be allowed to do CORS requests.
      allowOrigin:
        - www.google.com
        - google.com
      # Specifies the content for the *access-control-allow-methods*
      # header.
      allowMethods:
        - GET
        - POST
      # Specifies the content for the *access-control-allow-headers*
      # header.
      allowHeaders:
        - Content-Type
      # Specifies the content for the *access-control-expose-headers*
      # header.
      exposeHeaders:
        - Content-Type
      # Specifies the content for the *access-control-max-age* header.
      maxAge: 24h
      # Specifies whether the resource allows credentials.
      allowCredentials: true
  strategy: RoundRobin # (Optional) LB Algorithm to apply to all services, defaults for all services
[...]

glerchundi:
I know that you probably don't want to infect ingress with annotations but IMHO the cleanest way to do this is to allow the user specify each cors policies in there with annotations, something like this:

```
contour.heptio.com/enable-cors
contour.heptio.com/cors-allow-origin
contour.heptio.com/cors-allow-methods
contour.heptio.com/cors-allow-headers
contour.heptio.com/cors-allow-credentials
contour.heptio.com/cors-max-age
```

This directly maps to envoy `CorsPolicy` and allows to do everything related to CORS.

For now and until defined in the `IngressRoute`, this is only valid for standard k8s `Ingress` objects.

Regards,

davecheney:
Thanks for raising this issue.

TBH I'm going to punt on this til IngressRoute (project) is done. I don't want to see another seven annotations on the ingress document.

glerchundi:
@davecheney what do you think about approaching this now once `IngressRoute` (project) seems to be settled?

davecheney:
[...]

PR #906, opened 2019-02-26
design: add cors policy to ingressroute

glerchundi:
This adds the possibility to configure CORS policies on a per route
basis.

Although it's possible to add CORS policy on a virtualhost level, it
seems to be much more explicit to use it on a per route basis, better
aligned with how parameters like these are already defined in contour
and it also covers all the use-cases (In spite of not being too DRY).

References:
* Documentation: https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route.proto?highlight=shadow#route-corspolicy
* Protocol Buffers specification:
  * Model: https://github.com/envoyproxy/envoy/blob/caa804febeda4a7b43ff269010807cb5c0567129/api/envoy/api/v2/route/route.proto#L392
  * On RouteAction: https://github.com/envoyproxy/envoy/blob/caa804febeda4a7b43ff269010807cb5c0567129/api/envoy/api/v2/route/route.proto#L753
  * On VirtualHost: https://github.com/envoyproxy/envoy/blob/caa804febeda4a7b43ff269010807cb5c0567129/api/envoy/api/v2/route/route.proto#L110

Signed-Off-By: Gorka Lerchundi Osa <glertxundi@gmail.com>

### DIFF
--- a/design/ingressroute-design.md
+++ b/design/ingressroute-design.md
@@ -91,6 +91,26 @@ spec:
       port: 9000
     # If enforceTLS is specified, allows any request to this path to serve insecure requests
     permitInsecure: true
+    # if present describes the CORS policy.
+    cors:
+      # Specifies the origins that will be allowed to do CORS requests.
+      allowOrigin:
+        - www.google.com
+        - google.com
+      # Specifies the content for the *access-control-allow-methods* header.
+      allowMethods:
+        - GET
+        - POST
+      # Specifies the content for the *access-control-allow-headers* header.
+      allowHeaders:
+        - Content-Type
+      # Specifies the content for the *access-control-expose-headers* header.
+      exposeHeaders:
+        - Content-Type
+      # Specifies the content for the *access-control-max-age* header.
+      maxAge: 86400
+      # Specifies whether the resource allows credentials.
+      allowCredentials: true
   - match: /finance
     # delegate delegates the matching route to another IngressRoute object.
     # This delegates the responsibility for /finance to the IngressRoute matching the delegate parameters
@@ -175,6 +195,12 @@ The upstream host can return 503 if it wants to immediately notify downstream ho
 It is also important to note these are health checks that Envoy is implementing and are separate from any other system such as those that exist in Kubernetes.
 
 _Note: Passive health checking is implemented via Outlier detection and is used to dynamically determine whether some number of hosts in an upstream cluster are performing unlike others and removing them from the healthy load balancing set. Passive checking is not included yet but will be in a future release._
+
+## CORS
+
+Cross-Origin Resource Sharing (CORS) is a mechanism that uses additional HTTP headers to tell a browser to let a web application running at a domain have permission to access selected resources from a different domain. 
+
+This can be configured on a per route basis by using the [CorsPolicy](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route.proto?highlight=shadow#envoy-api-msg-route-corspolicy) struct defined by Envoy.
 
 ## Delegation rules
 


glerchundi:
Hope it's what you were expecting @davecheney!

davecheney:
I’m sorry but this is not what I was expecting. 

I’d like to understand

A. Why does the load balancer need to be involved in setting these headers. Why can’t the application maintain them?
B. What is special about them? Can they be generalised into a general header add/remove facility?

Please understand I am as frustrated with the progress of this issue as you are. I can see cors is an important feature but as I know nothing about it I lack the background to make an informed decision about a feature I have to support in perpetuity. 

I strongly recommend you follow the design document process, an example for tls certificate delectation was recently landed.

glerchundi:
> A. Why does the load balancer need to be involved in setting these headers. Why can’t the application maintain them?

Surely the app could maintain them. But then you're coupling your application with a concrete domain or multiple domains that it should be allowed to access. Which is real pain as most of the times the running app/team doesn't know from which domain is going to be accessed.

As the knowledge of which domains are exposed to the outside is on Contour/Envoy it makes perfect sense to not propagate that info. to lower layers and keep everything on the edge. I am making my point?

> B. What is special about them? Can they be generalised into a general header add/remove facility?

No, because you'll cover only one case which isn't the common one on the current browsers. There are two types of CORS the [simple](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Simple_requests) (which can be covered by creating a generic add/remove headers mechanism) one and a [preflight](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Preflighted_requests) variant where another type of request comes into play on the overall process, the [`OPTIONS`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS) request. It therefore cannot be covered by simple add/remove facility.

This feature is enabled on Envoy by adding another filter on the HTTP filter chain (not surprisingly called [CORS filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http_filters/cors_filter)). The headers included on the simple variant or in the `OPTIONS` response are configured by defining the policy on the route being accessed (the `CorsPolicy` parameter in `VirtualHost` or `RouteAction`).

> Please understand I am as frustrated with the progress of this issue as you are.

Sorry for that, it's being hard.

> I can see cors is an important feature but as I know nothing about it I lack the background to make an informed decision about a feature I have to support in perpetuity.

I strongly recommend you to read the post I sent to you and try to digest it because it is by far, the simplest and best resource I found about this topic.

> I strongly recommend you follow the design document process, an example for tls certificate delectation was recently landed.

I'm going to stop now and will try to come back with recovered strength next time.

yoitsro:
Hey @glerchundi @davecheney. Thanks for your work on this so far. I understand there seems to be some differences of opinion here, but allow me to add another use case for why CORS support should be added.

I'm attempting to use grpc-web with contour, but due to a lack of CORS support, this is what happens:

1. The request from the browser fires the preflight OPTIONS request to see if the current domain is allowed to access this URL
2. The request hits contour/envoy but responds with a 503. I think this is due to the fact that the HTTP 1.1 OPTIONS verb doesn't translate to anything in gRPC so envoy can't direct the request to anywhere.
3. The request is returned with a 503 error
4. The original GET/POST request is never triggered from the browser.

Having CORS support would be ideal right now.

aberasarte:
We are facing the issues @yoitsro described as well so, CORS support would definitely be a great addition for us and for anyone else trying to publish an API that is going to be consumed from third-party web applications.

As this seems to be a bit blocked, would you find appropriate if I start from the beginning sending a design document that takes the TLS certificate delegation design document as an example? It would add some background about what CORS is and it would introduce a design proposal for achieving this in Contour.

@davecheney WDYT?

glerchundi:
Feel free to take the lead of this proposal if @davecheney agrees, cannot afford this ATM.

davecheney:
@aberasarte yes, please send a design document referencing #437.

I'm going to close this PR as I can't see it making forward progress in the current form.

PR #912, opened 2019-03-03
Set DrainconnectionsOnHostRemoval to true for any cluster w/healthchecks

### LINKED ISSUE #603: Envoy does not remove endpoint after it's been removed from the Kubernetes Endpoint object when health checking is enabled
When active health checking is enabled, Envoy does not seem to remove endpoints from the ClusterLoadAssignment when endpoints are removed from the Kubernetes Endpoints object. 

There is an Envoy Cluster option that I suspect is related to this:
> **drain_connections_on_host_removal**
> (bool) If this cluster uses EDS or STRICT_DNS to configure its hosts, immediately drain connections from any hosts that are removed from service discovery.
>
> This only affects behavior for hosts that are being actively health checked. If this flag is not set to true, Envoy will wait until the hosts fail active health checking before removing it from the cluster.

(https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/cds.proto)

xref: https://github.com/heptio/gimbal/issues/208

#### Steps to reproduce:
1. Create a deployment:
```
k run kuard --image=gcr.io/kuar-demo/kuard-amd64:1 --expose --port 8080
```

2. Create an ingressroute:
```
kind: IngressRoute
metadata:
  name: kuard
spec:
  routes:
  - match: /
    services:
    - healthCheck:
        healthyThresholdCount: 5
        intervalSeconds: 5
        path: /
        timeoutSeconds: 2
        unhealthyThresholdCount: 3
      name: kuard
      port: 8080
  virtualhost:
    fqdn: example.com
```

3. Verify you can reach kuard via the ingressroute

4. Edit the kuard pod, and remove the `run: kuard` label to remove this pod from the Endpoints object. This will trigger the creation of a new pod, and the Kubernetes Endpoints object will be updated to contain the address of the new pod.

Expected result: Envoy removes the other endpoint from the ClusterLoadAssignment

Actual result: Envoy maintains the old endpoint in the ClusterLoadAssignment, and sends traffic to both the old pod and the new one.

stevesloka:
Fixes #603 by setting DrainConnectionsOnHostRemoval to true if the cluster has healthchecks enabled. 

//cc https://github.com/heptio/gimbal/issues/208

### DIFF
--- a/internal/contour/cluster_test.go
+++ b/internal/contour/cluster_test.go
@@ -17,7 +17,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/envoyproxy/go-control-plane/envoy/api/v2"
+	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
 	"github.com/envoyproxy/go-control-plane/envoy/api/v2/cluster"
 	"github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
 	"github.com/google/go-cmp/cmp"
@@ -25,7 +25,7 @@ import (
 	"github.com/heptio/contour/internal/envoy"
 	"github.com/heptio/contour/internal/metrics"
 	"github.com/prometheus/client_golang/prometheus"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/api/extensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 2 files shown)


stevesloka:
/hold

I need to test this quickly to validate that it actually does what it says.

davecheney:
@stevesloka it would be great if you could write an e2e test for this, taking the ingressroute from #603 and checking that contour emits the right envoy data structure.

davecheney:
Milestone removed, please re-add the 0.10 milestone when this PR is ready. Thanks.

stevesloka:
:+1: I'll get those e2e's in, thanks!

davecheney:
@stevesloka i'm going to merge this fix for 0.10. I've opened #926 to track the missing test

PR #920, opened 2019-03-05
deployment: switch to envoyproxy/envoy

davecheney:


mattalberts:
Is the debian image the preferred (as in, would it be to my benefit to switch to the non-alpine version in any of the release versions)?

davecheney:
Either will work fine

Issue #815, opened 2018-11-20
IngressRoute support for custom request timeout


gtriggiano:
**Describe the solution you'd like**
I'd like to be able to configure custom request timeouts per IngressRoute.

It seems from docs that you can use annotations on ingresses, but according to [this comment](https://github.com/heptio/contour/pull/655#discussion_r213334220) 

>  IngressRoutes don't really use annotations for anything

Docs regarding IngressRoute contain no reference to a _requests timeout config parameter_.

**Environment:**

- Kubernetes version: v1.10.3 (EKS)
- Cloud provider or hardware configuration: AWS

**Contour daemon set**
```yaml
apiVersion: extensions/v1beta1
kind: DaemonSet
metadata:
  labels:
    app: contour
  name: contour
  namespace: heptio-contour
spec:
  selector:
    matchLabels:
      app: contour
  template:
    metadata:
      annotations:
        prometheus.io/format: prometheus
        prometheus.io/path: /stats
        prometheus.io/port: "9001"
        prometheus.io/scrape: "true"
      creationTimestamp: null
      labels:
        app: contour
    spec:
      containers:
      - args:
        - serve
        - --incluster
        command:
        - contour
        image: gcr.io/heptio-images/contour:latest
        imagePullPolicy: Always
        name: contour
        ports:
        - containerPort: 8000
          hostPort: 8000
          name: contour
          protocol: TCP
      - args:
        - --config-path /config/contour.yaml
        - --service-cluster cluster0
        - --service-node node0
        - --log-level info
        - --v2-config-only
        command:
        - envoy
        image: docker.io/envoyproxy/envoy-alpine:v1.7.0
        name: envoy
        ports:
        - containerPort: 8080
          hostPort: 8080
          name: http
          protocol: TCP
        - containerPort: 8443
          hostPort: 8443
          name: https
          protocol: TCP
        volumeMounts:
        - mountPath: /config
          name: contour-config
      dnsPolicy: ClusterFirst
      hostNetwork: true
      initContainers:
      - args:
        - bootstrap
        - /config/contour.yaml
        command:
        - contour
        image: gcr.io/heptio-images/contour:latest
        imagePullPolicy: Always
        name: envoy-initconfig
        volumeMounts:
        - mountPath: /config
          name: contour-config
      restartPolicy: Always
      serviceAccount: contour
      serviceAccountName: contour
      volumes:
      - emptyDir: {}
        name: contour-config
```

jipperinbham:
[...]

PR #925, opened 2019-03-08
design: Support Custom Request Timeout

prasoontelang:
Updates #815

Signed-off-by: Prasoon Telang <telangp@vmware.com>

### DIFF
--- a/design/custom-request-timeout.md
+++ b/design/custom-request-timeout.md
@@ -1,0 +1,457 @@
+# Executive Summary
+
+**Status**: _Draft_
+
+This document describes the design of a new resource in IngressRoute for custom request timeout and retries. 
+
+This new resource will be integrated into Contour 0.11.
+
+# Goals
+
+- Generic resource name for configuring custom request timeouts per IngressRoute
+- Generic resource name for configuring custom retry attempts per IngressRoute
+
+# Non-goals
+
+- Even though Virtual Hosts in Envoy supports Retry Policy, this design doc will not address it.
+
+# Background
+
+Contour supports custom request timeout and custom retry attempts via [Ingress Annotations](https://github.com/heptio/contour/blob/master/docs/annotations.md).
+The same functionality has been requested via IngressRoute as well.
+
+# High-level design
+
+At a high level, this document proposes the changes in the IngressRoute's CRD to also support custom request
+timeouts and retries. It covers:
+
+- Fields added for timeout and retries in the YAML file.
+- A high-level coverage on the struct changes required in the code.
+- Arguments passed for Envoy API calls.
+
+## Proposed YAML fields
+
+```yaml
+apiVersion: contour.heptio.com/v1beta1
+kind: IngressRoute
+metadata:
+    name: github.com
+    namespace: prod
+spec:
+    virtualhost:
+        fqdn: www.github.com
+        routes:
+        - match: /contour
+          # the timeout applied to the requests to "/contour" only, independent of other routes.
+          timeoutPolicy:
+              # the request timeout applied on this route, it is the maximum time that a client will
+              # await a response to its request.
+              request: 1s
+              # the maximum time that an idle connection is kept open when there are no active requests
+              idle: 2s
+          # the retryPolicy described here are local to match and apply for "/contour" only
+          retryPolicy:
+              # the number of retries it should make on failure, in conjunction with status codes
+              # mentioned for this route.
+              count: "7"
+              # the conditions for which there should be a reattempt, with the number of retries
+              # mentioned in count for this route.
+              onStatusCodes:
+              - "501"
+              - "502"
+              # timeout between the requests, if greater than request timeout, this field is ignored
+              perTryTimeout: 150ms
+        - match: /static-site
+          # the timeout applied for the requests to "/static-site" only, independent of other routes.
+          timeoutPolicy:
+              request: 200ms
+              idle: 1s
+          # retry for this route is independent of "/contour" retryPolicy.
+          retryPolicy:
+          	count: "2"
+          	onStatusCodes:
+          	- 50x
+```
+
+The naming conventions of the proposed YAML fields are inspired from [HTTP Timeouts](https://tools.ietf.org/id/draft-thomson-hybi-http-timeout-00.html#rfc.section.4).
+
+These proposed YAML resources are defined with being generic as motivation so that any ingress controller can use these 
+fields of IngressRoute and setup the ingress, not just limited to contour-envoy only.
+
+# Detailed Design
+
+## Reading from YAML
+The ingressroute spec struct will be updated to contain Timeout and Retry struct members
+
+
+```go
+// contour/apis/contour/v1beta1/ingressroute.go
+
+// Route contains the set of routes for a virtual host
+type Route struct {
+    [... other members ...]
+	// The timeout policy for this route
+	TimeoutPolicy *TimeoutPolicy `json:"timeoutPolicy,omitempty"`
+	// The retry policy for this route
+	RetryPolicy *RetryPolicy `json:"retryPolicy,omitempty"`
+}
+
+// TimeoutPolicy define the attributes associated with timeout
+type TimeoutPolicy struct {
+	// Timeout for receiving a response from the server after processing a request from client
+	Request	*JsonDuration `json:"request"`
+	// Timeout for an idle connection to terminate when there are no active requests
+	Idle *JsonDuration `json:"idle"`
+}
+
+// RetryPolicy define the attributes associated with retrying policy
+type RetryPolicy struct {
+	// NumRetries is maximum allowed number of retries. This is specified as string because 
+	// there isn't a straightforward way to determine if it was an erroneous input and default
+	NumRetries	string `json:"count"`
+	// Perform retry on failed requests with the matched status codes or aggregated as 5xx
+	OnStatusCodes []string `json:"onStatusCodes"`
+	// PerTryTimeout specifies the timeout per retry attempt. Ignored if OnStatusCodes are empty
+	PerTryTimeout *JsonDuration `json:"perTryTimeout"`
+}
+
+// new struct to parse from JSON and load directly as time.Duration 
+type JsonDuration struct {
+	*time.Duration
+}
+
+// to Unmarshal bytes of JSON into JsonDuration type
+func (d *JsonDuration)UnmarshalJSON(b []byte) (err error) {
+	
+	timeStr := strings.Trim(string(b), `"`)
+	duration, err := time.ParseDuration(timeStr)
+	
+	if err != nil && timeStr == "infinity" {
+		duration = -1
+		err = nil
+	}
+	
+	if err == nil {
+		// only if correctly parsed, we update the Duration. Else, Duration is initialized as nil
+		d.Duration = &duration
+	}
+	return
+}
+
+// to Marshal JsonDuration into JSON bytes type (unused)
+func (d JsonDuration)MarshalJSON() (b []byte, err error) {
+	return []byte(fmt.Sprintf(`"%s"`, d.String())), nil
+}
+
+// a wrapper function to interpret all forms of input received from the YAML file for contour
+func (d *JsonDuration) Time() (timeout time.Duration, valid bool) {
+	if d == nil {
+		// this means the timeout field (like request, idle, etc.) was not specified in the YAML file
+		// not specifying the timeout field is a valid input
+		timeout, valid = 0, true
+	} else if d.Duration == nil {
+		// this means the timeout field was incorrectly specified, hence it couldn't be casted to time.Duration
+		// incorrectly parsed timeout is an invalid input
+		timeout, valid = -1, false
+	} else {
+		// timeout was correctly parsed, hence valid input
+		timeout, valid = *d.Duration, true
+	}
+	
+	return
+}
+```
+
+## Intermediate form in the DAG for the routes
+We store unmarshalled structs from ingress' and ingressroute's yaml into a common dag struct route for both. 
+
+Consolidate timeout and retry members of dag's Route struct to their own respective struct. Though the names
+`TimeoutPolicy` and `RetryPolicy` are reused, this isn't the same as described in the previous section.
+
+```go
+// Approach: contour/internal/dag/dag.go
+
+type TimeoutPolicy struct {
+    Request *time.Duration
+    Idle    *time.Duration
+    [... more members can be added based on requirements ...]
+}
+
+type RetryPolicy struct {
+    NumRetries      int
+    // []string casted to []uint32 list, if 50x is present, generate []uint32{500, 501, ..., 509}
+    OnStatusCodes   []uint32
+    PerTryTimeout   *time.Duration
+    [... more members can be added based on requirements ...]
+}
+
+type Route struct {
+    Prefix          string
+    object          interface{}
+    httpServices    map[servicemeta]*HTTPService
+    HTTPSUpgrade    bool
+    Websocket       bool
+    PrefixRewrite   string
+
+    // new member replacing current Timeout
+    TimeoutPolicy   *TimeoutPolicy
+    // new member replacing current RetryOn, NumRetries, PerTryTimeout
+    RetryPolicy     *RetryPolicy
+}
+```
+
+This will involve slightly more changes in the code and their unit tests to be corrected, but hopefully,
+the future additions if made can be neatly encapsulated.
+
+## Communicating from DAG to Envoy
+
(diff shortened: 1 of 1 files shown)


prasoontelang:
There was a failure in the Travis CI build. It failed to clone golang tools. I couldn't retrigger the job as I don't see the button on top right (maybe due to permission issue). Any leads on how do I resolve it?

```
(54/54) Failed to write golang.org/x/tools@master
grouped write of manifest, lock and vendor: error while writing out vendor tree: failed to write dep tree: failed to export golang.org/x/tools: failed to fetch source for https://go.googlesource.com/tools: unable to get repository: Cloning into '/home/travis/gopath/pkg/dep/sources/https---go.googlesource.com-tools'...
POST git-upload-pack (691 bytes)
fatal: remote error: Internal Server Error
__ fatal: The remote end hung up unexpectedly __
__ : command failed: [git clone --recursive -v --progress https://go.googlesource.com/tools /home/travis/gopath/pkg/dep/sources/https---go.googlesource.com-tools]: exit status 128 __
The command "go get -u github.com/golang/dep/cmd/dep && dep ensure -vendor-only -v" failed and exited with 1 during .
Your build has been stopped.
```

vaamarnath:
@prasoontelang The easiest way I know of is to perform a `git amend` on your local branch and force push the changes again to your upstream fork. This will re-trigger the tests on Travis CI.

davecheney:
Thank you for this change. I'll review it next week when the 0.11 cycle opens.

design/custom-request-timeout.md
```diff
@@ -0,0 +1,256 @@
...
++---------+      +---------+       +----------+
+
+| Client  |----->| Proxy   |------>| Server   |
+
++---------+      +---------+       +----------+
+
+
+### Route.Timeout
+
+- `Request` : The time that spans between the point at which complete client request has been processed by the proxy, and when the
```
unicell: Could you link Envoy docs for the field definition? Less moving parts that way. Thanks!

prasoontelang: Will do that.

design/custom-request-timeout.md
```diff
@@ -0,0 +1,256 @@
...
+**Status**: _Draft_
+
+This document describes the design of a new resource in IngressRoute for custom request timeout and retries. 
+
+This new resource will be integrated into Contour 0.11.
+
+# Goals
+
+- Generic resource name for configuring custom request timeouts per IngressRoute
+- Generic resource name for configuring custom retry attempts per IngressRoute
```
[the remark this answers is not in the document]
prasoontelang: Thanks! I will address that as well. I was kinda focused on the issue's "Describe the solution you'd like". Would be nice to address both the cases

design/custom-request-timeout.md
```diff
@@ -0,0 +1,256 @@
...
+          # the timeout applied to the requests in this route
+          timeoutPolicy:
+              # the request timeout applied on this route, it is the maximum time that a client will await
+              # a response to its request.
+              request: 500ms
+              # the maximum time that an idle connection is kept open when there are no active requests
+              idle: 2s
+          retryPolicy:
+              # the number of retries it should make on failure
+              maxRetries: "7"
```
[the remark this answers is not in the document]
prasoontelang: Nice catch. Will do that.

design/custom-request-timeout.md
```diff
@@ -0,0 +1,256 @@
...
+              # the conditions for which it should reattempt
+              onStatusCodes:
```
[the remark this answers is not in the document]
prasoontelang: Your assumptions are correct.

I believe the user should be able to specify `5xx` and `4xx` in the YAML file. When some of the "xx" is included, the contour should map it to `"5xx"` or `"retriable-4xx"` respectively against `retry_on` for Envoy.  If `50x` or `40x` is specified, contour should expand that to `500,501,502..,509` and `400,401,402...,409` just like the specific status codes, which will be mapped to `retriable_status_codes`.

design/custom-request-timeout.md
```diff
@@ -0,0 +1,256 @@
...
+- Generic resource name for configuring custom retry attempts per IngressRoute
+
+# Non-goals
+
+- None
+
+# Background
+
+Contour supports custom request timeout and custom retry attempts via Ingress. The same functionality has been
+requested via IngressRoute as well.
```
davecheney: Please link to the documentation or issue for Ingress so its clear to the reader what we're replicating in IngressRoute.

design/custom-request-timeout.md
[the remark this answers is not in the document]
davecheney: Nice. I like how you separated timeoutPolicy and retryPolicy into separate objects, this feels right.

design/custom-request-timeout.md
```diff
@@ -0,0 +1,256 @@
...
+	// The request timeout for this route
+	Timeout TimeoutPolicy `json:"timeoutPolicy,omitempty"`
+	// The retry attempts for this route
+	Retry RetryPolicy `json:"retryPolicy,omitempty"`
+}
+
+// TimeoutPolicy defines the attributes associated with timeout
+type TimeoutPolicy struct {
+	// Timeout for receiving a response from the server after processing a request from client in milliseconds
+	Request	string `json:"request"`
```
davecheney: Can these be made time.Duration's? Does k8s support that?

design/custom-request-timeout.md
```diff
@@ -0,0 +1,256 @@
...
+	// Timeout for an idle connection to terminate when there are no active requests
+	Idle string `json:"idle"`
+}
+
+// RetryPolicy defines the attributes associated with retrying policy
+type RetryPolicy struct {
+	// MaxRetries is maximum allowed number of retries
+	MaxRetries	int `json:"maxRetries"`
+	// Perform retry on failed requests with the matched status codes or aggregated as 5xx
+	OnStatusCodes []string `json:"onStatusCodes"`
```
davecheney: I know we support this in Ingress, but nit picking about status codes feels really low level. What if we just said that retries only happen on 50x status codes?

design/custom-request-timeout.md
```diff
@@ -0,0 +1,256 @@
...
+	// Timeout for receiving a response from the server after processing a request from client in milliseconds
+	Request	string `json:"request"`
+	// Timeout for an idle connection to terminate when there are no active requests
+	Idle string `json:"idle"`
+}
+
+// RetryPolicy defines the attributes associated with retrying policy
+type RetryPolicy struct {
+	// MaxRetries is maximum allowed number of retries
+	MaxRetries	int `json:"maxRetries"`
```
davecheney: The documentation should call out what will happen if this is not supplied. ie, If not supplied, retries will not be performed.

davecheney: Is there a better name than 'MaxRetries'? There is no MinRetries value, so the Max prefix is weirdly redundant. What about 'Count' or just 'retries'?

design/custom-request-timeout.md
```diff
@@ -0,0 +1,256 @@
...
+	// Perform retry on failed requests with the matched status codes or aggregated as 5xx
+	OnStatusCodes []string `json:"onStatusCodes"`
+	// PerTryTimeout specifies the timeout per retry attempt. Ignored if OnStatusCodes are empty
+	PerTryTimeout string `json:"perTryTimeout"`
```
davecheney: If we don't let people provide an empty set of OnStatusCodes then PerTryTimeout will always be consulted. In that case we should call out the default value if not supplied; ie, its reasonable for people to write
````
retryPolicy:
  maxRetries: 6
```
We should make Contour behave sensibly in this instance.

design/custom-request-timeout.md
```diff
@@ -0,0 +1,256 @@
...
+type Route struct {
+    [... other members ...]
+    IdleTimeout time.Duration
+}
+
+```
+
+### Approach 2
+
+Consolidate timeout and retry members of dag's Route struct to their own respective struct.
```
davecheney: Yes please. Even though I wrote elsewhere that I don't want to reuse the k8s types in the dag your idea of separating the two policy objects is excellent and we should reproduce it in the dag data model.

design/custom-request-timeout.md
```diff
@@ -0,0 +1,256 @@
...
+type Route struct {
+    Prefix          string
+    object          interface{}
+    httpServices    map[servicemeta]*HTTPService
+    HTTPSUpgrade    bool
+    Websocket       bool
+    PrefixRewrite   string
+
+    // new member replacing current Timeout
+    Timeout         RouteTimeout
```
davecheney: There needs to be some discussion of what happens if the timeout key is not supplied. I think at a minimum this should be a pointer, `Timeout *RouteTimeout`.

Same for Retry

design/custom-request-timeout.md
```diff
@@ -0,0 +1,256 @@
...
+### Approach 2
+
+Consolidate timeout and retry members of dag's Route struct to their own respective struct.
+
+```golang
+// Approach 2: contour/internal/dag/dag.go
+
+type RouteTimeout struct {
+    Request time.Duration
+    Idle    time.Duration
```
davecheney: Please have a look at the discussions when this was added to Ingress. There is some messiness about dealing with 'Infinite' timeout, which envoy interprets as 0, and 'not set' which is _not_ the zero value of this field.

Sadly this grossness will have to be replicated here because people _will_ ask for infinite timeout.

design/custom-request-timeout.md
prasoontelang: Great catch! k8s have fields with "seconds" in the resource names: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go 

So they just read it as seconds. We could be dealing with milliseconds, so we could do the following by using `JsonDuration` in the playground example:

https://play.golang.org/p/ZzSZ9SHhqjL 

I would prefer using JsonDuration approach as it keeps the code clean, but k8s doesn't have it.

design/custom-request-timeout.md
prasoontelang: From the comment of @unicell , I have changed struct fields to numRetries. 

I will keep the YAML field as "count", which will unmarshal into `numRetries` struct member.

design/custom-request-timeout.md
prasoontelang: Good point for keeping it simple. Just to be sure, contour would accept the following?

- `501` -> `501` only
- `501,502` -> `501,502` only
- `50x` -> contour will expand it to `500,501,502,503,504,505,506,507,508,509` before feeding to envoy?

design/custom-request-timeout.md
prasoontelang: Noted. I will mention this more clearly in the follow up for this design doc.

design/custom-request-timeout.md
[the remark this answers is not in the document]
prasoontelang: based on the recent feedback for keeping ingressroute generic for all controllers, ingressroute for now will only have `50x` or {500->509} based status codes only.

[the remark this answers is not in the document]
unicell: `retry_on` provides difference schematics than `retriable_status_codes`. For example:

> 5xx
> Envoy will attempt a retry if the upstream server responds with any 5xx response code, or does not respond at all (disconnect/reset/read timeout). (Includes connect-failure and refused-stream) 

5xx retry policy consists of cases where upstream returns 5xx in response as well as those with no response code at all.

But I understand that probably we don't want to tie Contour with Envoy too closely per issues/587.

design/custom-request-timeout.md
unicell: On keeping things simple, I feel Envoy's `retry_on` policy[1] provides exactly of that. So that user can simply say retry on 5xx that covers all the 5xx in response code as well as those no response code cases as well.

[1] https://www.envoyproxy.io/docs/envoy/latest/configuration/http_filters/router_filter.html#x-envoy-retry-on

design/custom-request-timeout.md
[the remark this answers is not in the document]
unicell: This is the part I don't really get. If I understand it correctly, the `timeoutPolicy` here will be mapped to RouteAction.timeout  and RouteAction.idle_timeout in Envoy API. So making it into its own object, just to make it look cleaner, but not for later to be used or referred from other objects, correct?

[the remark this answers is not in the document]
prasoontelang: your understanding about `timeoutPolicy` to RouteAction.timeout and RouteAction.idle_timeout is correct. Within contour, timeout will be in its own object only for cleanliness, while populating to Envoy API it will be mapped accordingly.

design/custom-request-timeout.md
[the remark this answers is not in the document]
prasoontelang: Sorry for not making it clear in the design doc. A user can specify

```
  retryPolicy:
    onStatusCodes:
      - "505"
      - "509"
      - "501"
```

Since contour sees them all as numbers, it will set "retriable-status-codes" for `retry_on`, and actually fill in uint32 type status codes for `retriable_status_codes`. For anyother ingress controller using ingressroute, they should treat them as status codes to perform retry.

However, if contour sees anything like the following:

```
retryPolicy:
  onStatusCodes:
    - "50x"
    - "503"
```

Contour (or any ingress controller) sees that "50x" has been specified and it should expand to [500->509]. To configure for envoy, since we don't want to retry`5xx` (ie status codes >= 510 which is outside user's configuration), we will repeat the step as in previous example of setting "retriable-status-codes" for `retry_on` and then setting uint32 type status codes to `retriable_status_codes`.

design/custom-request-timeout.md
prasoontelang: I like your proposal but I am confused between keeping `50x` or `5xx` on YAML. If we are keeping `50x` then I feel mapping it to `5xx` could be misleading.

design/custom-request-timeout.md
```diff
@@ -0,0 +1,362 @@
...
+kind: IngressRoute
+metadata:
+    name: github.com
+    namespace: prod
+spec:
+    virtualhost:
+        fqdn: www.github.com
+        # this retryPolicy is applied globally to all routes under this virtual host.
+        # if a retryPolicy is redefined in a route, that policy gets preference for that route only.
+        retryPolicy:
```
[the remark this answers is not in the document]
prasoontelang: I just had a look at a retry pattern (src: https://docs.microsoft.com/en-us/azure/architecture/patterns/retry )

> **Cancel.** If the fault indicates that the failure isn't transient or is unlikely to be successful if repeated, the application should cancel the operation and report an exception. For example, an authentication failure caused by providing invalid credentials is not likely to succeed no matter how many times it's attempted.

> **Retry.** If the specific fault reported is unusual or rare, it might have been caused by unusual circumstances such as a network packet becoming corrupted while it was being transmitted. In this case, the application could retry the failing request again immediately because the same failure is unlikely to be repeated and the request will probably be successful.

`4xx` are user based errors and `5xx` are server based. `4xx` if retried, from my understanding, will continue to cause error unless the user corrects it. By complying with the highlighted pattern, in my opinion we shouldn't retry for most of `4xx`. 

However, there are status codes such as [409](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409) and [429](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) seem retriable.

Regarding your second statement of defining a set of retryPolicies (if I understand this correctly, it is like having custom names for a group of predefined status codes, just like the Envoy API), I fear that it will lead to Envoy [Issue 4103](https://github.com/envoyproxy/envoy/issues/4103). There may come a scenario where users might request new codes to include, eventually leading to numerically accepting status codes.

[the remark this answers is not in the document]
davecheney: I would prefer not to see anything route or path related added at the virtual host level. The Ingressroute crd is underdeveloped on this point by the virtualhost section is _not_ the catch all for routes. For this iteration, please do not try to implement a per virtualhost default setting. Leave it on the route.

davecheney:
@prasoontelang thank you for your patience. I've marked this document as approved with the following caveats

- Support for idle timeout has been dropped. We can revisit this when there is specific customer demand.
- Retry count is defined to be a number, the problems of the zero value for `NumRetries` coming from a failed string parsing, or actually being the number 0 are resolved by specifying that 0 is the "not set" value for `NumRetries`. This means you cannot explicitly say
```
       timeout:
         count: 0
```
But it turns out that leaving the field out has the same effect.
- Retry on status code support is hard coded to "50x". We can revisit this when there is specific customer demand.

If I've misunderstood the effect of any of these simplifications please let me know.

Due to the need to ship Contour 0.11 quickly to address the security issues in Envoy 1.9.0 I've moved #815 to the 0.12 milestone. As the 0.12 cycle will start this week, as soon as 0.11 is shipped, the net effect of this should be minimal.

Issue #592, opened 2018-07-26
internal/contour: make HoldoffNotifier delay constants configurable


davecheney:
#577 introduced a HoldoffNotifier type which is hard coded to a 100ms/500ms delay. This makes it unsuitable for use with our e2e tests so it is not currently used in the e2e path.

Investigate making the holdoff delays fields on the HoldoffNotifier itself and possibly use smaller values in the e2e tests.

alexbrand:
Started looking into this. The challenge here I think is getting it wired into the e2e tests. If we make the delay > 0, it seems like we will have to make our tests assert multiple times with a timeout, and somehow have access to the delay and maxDelay params to know what value to use as timeout. 

Open to ideas if there's a better approach here.

davecheney:
If this is hard then skip it, testing timeouts is messy. I’d prefer to test the hold off notifier in isolation.

rosskukulinski:
I think this is worth resolving for 0.6 to pay down tech debt.

rosskukulinski:
@davecheney I'm +1 for that -- Reading back over the thread I realize now there are more important things to tackle.

rosskukulinski:
moved to 0.7

davecheney:
Removing milestone, this is moderate tech debt, but I don't think qualifies for a planned release.

andrewsykim:
Opened a PR for this here https://github.com/heptio/contour/pull/927

davecheney:
Closing. The need to configure these constants is not important enough.

PR #927, opened 2019-03-08
make HoldoffNotifier delay constants configurable

andrewsykim:
Signed-off-by: Andrew Sy Kim kim.andrewsy@gmail.com

Fixes https://github.com/heptio/contour/issues/592 so we can use HoldoffNotifier in e2e tests. 

cc @davecheney

### DIFF
--- a/internal/contour/holdoff.go
+++ b/internal/contour/holdoff.go
@@ -26,8 +26,8 @@ import (
 )
 
 const (
-	holdoffDelay    = 100 * time.Millisecond
-	holdoffMaxDelay = 500 * time.Millisecond
+	defaultHoldoffDelay    = 100 * time.Millisecond
+	defaultHoldoffMaxDelay = 500 * time.Millisecond
 )
 
 // A HoldoffNotifier delays calls to OnChange in the hope of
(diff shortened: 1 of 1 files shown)


internal/contour/holdoff.go
```diff
@@ -42,16 +42,36 @@ type HoldoffNotifier struct {
...
+
+func (hn *HoldoffNotifier) SetHoldoffMaxDelay(t time.Duration) {
+	hn.holdoffMaxDelay = t
 }
 
 func (hn *HoldoffNotifier) OnChange(builder *dag.Builder) {
 	hn.mu.Lock()
 	defer hn.mu.Unlock()
+
+	if hn.holdoffDelay == 0 {
```
andrewsykim: This check is only needed once, I was thinking of removing this check by using a constructor func but this seemed cleaner.

davecheney: What happens if someone calls `SetHoldoffDelay(0)`, their value will be overridden with the default.

internal/contour/holdoff.go
```diff
@@ -42,16 +42,36 @@ type HoldoffNotifier struct {
...
+
+	holdoffDelay    time.Duration
+	holdoffMaxDelay time.Duration
+}
+
+func (hn *HoldoffNotifier) SetHoldoffDelay(t time.Duration) {
+	hn.holdoffDelay = t
+}
+
+func (hn *HoldoffNotifier) SetHoldoffMaxDelay(t time.Duration) {
```
davecheney: What happens if these functions are called after the notifier is started? Its either going to be a data race or the updated values will be ignored.

andrewsykim:
Good point @davecheney. How do you feel about having a constructor func for HoldoffNotifier that sets the defaults in that case?

davecheney:
I'd prefer to punt on this issue til its assigned a milestone. There's more than enough to do in 0.11 and 0.12

davecheney:
Closing based on https://github.com/heptio/contour/pull/927#issuecomment-472251124, please resend this when #592 is assigned a milestone.

PR #928, opened 2019-03-09
E2E tests drain connections host removal

### LINKED ISSUE #926: Need e2e test for #912
#912 landed late in the 0.10 cycle without an e2e test. It would be awesome if someone could write one so we don't revert this fix in the future.

Updates #603

vaamarnath:
Fixes #926


davecheney:
Thank you. I'll look at this on monday

PR #929, opened 2019-03-11
use shared informer factories for watchers

andrewsykim:
Uses informer factories to start client-go watchers. 

Other notable changes:
* block on `WaitForCacheSync` before running informers
* group all core informers & contour informers under the same workgroup - I can see why this isn't desirable, feel free to close this PR in that case :)


davecheney:
Thank you for this change, I appreciate it, but I don't plan on merging it without discussion. Please see our contributing guidelines and open an issue to discuss this change first. Thank you.

davecheney:
Sorry, I didn't mean to close this. Let's talk about _in an issue_, not on this PR first.

cmd/contour/contour.go
```diff
@@ -165,12 +168,15 @@ func main() {
...
 		client, contourClient := newClient(*kubeconfig, *inCluster)
 
-		wl := log.WithField("context", "watch")
-		k8s.WatchServices(&g, client, wl, &reh)
-		k8s.WatchIngress(&g, client, wl, &reh)
-		k8s.WatchSecrets(&g, client, wl, &reh)
-		k8s.WatchIngressRoutes(&g, contourClient, wl, &reh)
-		k8s.WatchTLSCertificateDelegations(&g, contourClient, wl, &reh)
+		// resync timer disabled for Contour
+		coreInformers := coreinformers.NewSharedInformerFactory(client, time.Duration(0))
```
davecheney: you don't need to convert 0 to time.Duration, the literal constant is automatically converted to the type required by the function call.

cmd/contour/contour.go
```diff
@@ -181,7 +187,35 @@ func main() {
...
-		k8s.WatchEndpoints(&g, client, wl, et)
+		coreInformers.Core().V1().Endpoints().Informer().AddEventHandler(et)
+
+		g.Add(func(stop <-chan struct{}) error {
+			log := log.WithField("context", "coreinformers")
+
+			log.Println("waiting for cache sync from core informers")
+			coreInformers.WaitForCacheSync(stop)
+
+			log.Println("starting core shared informer")
```
davecheney: log.Println("started") is probably good enough given the context captured in the logger.

cmd/contour/contour.go
```diff
@@ -181,7 +187,35 @@ func main() {
...
+			defer log.Println("stopping core shared informer")
+			coreInformers.Start(stop)
+
+			<-stop
+			return nil
+		})
+
+		g.Add(func(stop <-chan struct{}) error {
+			log := log.WithField("context", "contourinformers")
```
davecheney: same as above.

It might be worth breaking this out into a helper function that takes an interface with two methods, `WaitForCacheSync` and `Start`.

davecheney:
Fixes #930

cmd/contour/contour.go
andrewsykim: Updated. This also required the helper func to accept the stop channel and the logger.

internal/k8s/status.go
```diff
@@ -11,6 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package k8s contains helpers for setting the IngressRoute status
```
andrewsykim: I think this package might be better named `ingressroute` given it no longer contains watchers. I can do this in a follow-up PR

Issue #598, opened 2018-07-31
Switch from Dep to go modules


davecheney:
Requires #597

https://go-review.googlesource.com/c/go/+/124384 introduces support for a `v8.0.0+incompatible` annotation in the `go.mod` file which should provide a bridge between semver dependencies and go 1.11 Semantic Import Versioning requirement. This will let us update `go.mod` without having to wait for `k8s.io/client-go` to rewrite the repository.

davecheney:
Removing milestone, #691 is not going to happen for 0.7 and we might skip Go 1.11 completely.

davecheney:
Revisit after #848

davecheney:
This issue is open for anyone who wants to try it.

vaamarnath:
I can pick it up @davecheney

vaamarnath:
I have chosen to vendor the dependencies using `go mod vendor`. I have made the changes accordingly.

PR #935, opened 2019-03-12
gopkg.toml: upgrade to client-go v9.0.0 / kubernetes v1.12.6

### LINKED ISSUE #934: Upgrade to k8s.io/client-go 1.12.x
Upgrade from client-go 1.11.x to 1.12.x

vaamarnath:

Fixes #934

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -56,10 +56,16 @@
     "envoy/api/v2/listener",
     "envoy/api/v2/route",
     "envoy/config/accesslog/v2",
+    "envoy/config/bootstrap/v2",
     "envoy/config/filter/accesslog/v2",
     "envoy/config/filter/network/tcp_proxy/v2",
+    "envoy/config/metrics/v2",
+    "envoy/config/overload/v2alpha",
+    "envoy/config/ratelimit/v2",
+    "envoy/config/trace/v2",
     "envoy/service/load_stats/v2",
     "envoy/type",
+    "envoy/type/matcher",
     "pkg/util",
   ]
   pruneopts = ""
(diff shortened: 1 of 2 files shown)


Gopkg.lock
```diff
@@ -56,10 +56,16 @@
     "envoy/api/v2/listener",
     "envoy/api/v2/route",
     "envoy/config/accesslog/v2",
+    "envoy/config/bootstrap/v2",
```
vaamarnath: These dependencies got added when I run `dep ensure` after updating `Gopkg.toml`. I double-checked if it is genuine by running `dep check` and looks like these were missing from `Gopkg.lock` earlier.

```
➜  contour git:(client-go-update-1.12) ✗ dep check
# Gopkg.lock is out of sync:
github.com/envoyproxy/go-control-plane/envoy/config/bootstrap/v2: imported or required, but missing from Gopkg.lock's input-imports
github.com/envoyproxy/go-control-plane/envoy/config/metrics/v2: imported or required, but missing from Gopkg.lock's input-imports
```

davecheney: Who can say, dep is pretty buggy

davecheney:
@vaamarnath i'll merge this today. If you want to raise a p2 issue for `dep check` complaining, please go for it. I don't think its a high priority, we'll probably move away from dep by 0.12

vaamarnath:
@davecheney Yeah, I see there is an issue (#598) open already for moving to go modules. I will not raise an issue for `dep check` complaining for now.

davecheney:
I’m not sure what dep check is upset about, but it’s probably been upset for a long time and nothing bad has happened. I don’t see any warnings during the Travis ci builds so I think we can live with this debt til 0.12

PR #937, opened 2019-03-12
design: sticky sessions support

rohandvora:
Updates #361

### DIFF
--- a/design/sticky-session-design.md
+++ b/design/sticky-session-design.md
@@ -1,0 +1,343 @@
+# Sticky session support
+
+**Status**: _draft_
+
+This document proposes adding support for sticky session in contour.
+
+## Goals
+
+- Enabling support for sticky-sessions in Contour using
+    - Cookie
+    - Source IP
+    - Header
+- Adding support for sticky-sessions in IngressRoute CRD. 
+
+## Non-goals
+
+- Making options in IngressRoute very specific to Envoy.
+
+## Background
+
+Sticky-sessions allow requests to be routed to the same host based on certain parameters of the request.
+The parameters can be a cookie, the source IP, a custom header, etc.
+
+#### Sticky sessions in Envoy 
+Envoy supports sticky sessions by using a hash-based load balancing strategy (RingHash, Maglev). 
+Specifying a **hash policy** for such load-balancers will enable sticky sessions. Hash policy 
+is specified per route. **For sticky sessions, each route should only have one service
+associated to the route**. This is because if multiple clusters are associated per route (weighted clusters),
+requests will be forwarded to the cluster based on weights and stickiness will be 
+enforced in that cluster from that point. This defeats the purpose of using sticky sessions since a session will not
+be pinned to a pod. 
+
+## High-Level Design
+In contour, sticky sessions will be enabled per service. 
+If a service has sticky sessions enabled, it should be the **only service**
+under that route. Sticky sessions **do not** work with multiple services (weighted clusters).
+This means that sticky sessions cannot be used along with blue/green deployments.
+
+### Sticky-sessions configuration
+To enable sticky-sessions in a service, the `strategy` needs to be set to *StickySession*.
+By default, sticky-session will be based on **source-ip**. `stickySessionConfig` can 
+be provided in the service to specify the parameter for stickiness.
+
+#### Cookie-based stickiness
+This allows a request to be routed based on a cookie in the http request.
+The following parameters can be specified under the strategy options for cookie-based stickiness:
+- **name**: The name of the cookie to look for to determine stickiness.
+- **path**: The path for the cookie.
+```yaml
+apiVersion: contour.heptio.com/v1beta1
+kind: IngressRoute
+metadata: 
+  name: basic
+  namespace: default
+spec: 
+  virtualhost:
+    fqdn: foo-basic.bar.com
+  routes: 
+    - match: /       
+      services: 
+        - name: s1
+          port: 80  
+          strategy: StickySession
+          stickySessionConfig:
+            cookie:
+              name: name
+              path: cookie-path
+```                
+
+#### Header-based stickiness
+This allows a request to be routed based on a header in the http request.
+The following parameters can be specified under the strategy options for header-based stickiness:
+- **name**: The name of the header to look for to determine stickiness.
+```yaml
+apiVersion: contour.heptio.com/v1beta1
+kind: IngressRoute
+metadata: 
+  name: basic
+  namespace: default
+spec: 
+  virtualhost:
+    fqdn: foo-basic.bar.com
+  routes: 
+    - match: /       
+      services: 
+        - name: s1
+          port: 80  
+          strategy: StickySession
+          stickySessionConfig:
+            header:
+              name: header-name
+```  
+
+#### Connection-based stickiness
+This allows a request to be routed based on a connection property of the incoming request. Currently supported
+property is the source-IP of the incoming request.
+The following parameters can be specified under the strategy options for connection-based stickiness:
+- **property**: The property of the connection to be used to determine stickiness. Currently supported property:   
+    - IP
+```yaml
+apiVersion: contour.heptio.com/v1beta1
+kind: IngressRoute
+metadata: 
+  name: basic
+  namespace: default
+spec: 
+  virtualhost:
+    fqdn: foo-basic.bar.com
+  routes: 
+    - match: /       
+      services: 
+        - name: s1
+          port: 80  
+          strategy: StickySession
+          stickySessionConfig:
+            connection:
+              property: IP
+```  
+
+## Detailed Design
+
+Changes will be made to the ingressRoute CRD, the DAG, and envoy-specific internal code.
+
+### IngressRoute
+
+- deployment/common/common.yaml will be modified to include change in the ingressroute CRD.
+```yaml
+spec:
+  group: contour.heptio.com
+  version: v1beta1
+  scope: Namespaced
+  names:
+    plural: ingressroutes
+    kind: IngressRoute
+  additionalPrinterColumns: ...
+  validation:
+    openAPIV3Schema:
+      properties:
+        spec:
+          properties:
+            ...
+            routes:
+              type: array
+              items:
+                ...
+                properties:
+                  ...
+                  services:
+                    type: array
+                    items:
+                      type: object
+                      ...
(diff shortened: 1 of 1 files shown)


design/sticky-session-design.md
```diff
@@ -0,0 +1,170 @@
...
+metadata: 
+  name: basic
+  namespace: default
+spec: 
+  virtualhost:
+    fqdn: foo-basic.bar.com
+  routes: 
+    - match: /
+      sticky-session:
+        source: cookie/header/connection # There should be exactly one of these set.
```
davecheney: This is good but I think we can tighten this up quite a bit using different data structures to represent the different options; that also might mean we don't have to implement each of these options initailly.

My suggestion is 

```
routes:
  - match: /
     sticky-session:
       cookie:
          name: // name of cookie
          path: // cookie path
```
Then that opens the door for coming back and adding a the other types under their own key

design/sticky-session-design.md
```diff
@@ -0,0 +1,170 @@
...
+        name: name # This can be cookie name, header name, or connection property
+        ttl: time-to-live # This is only valid when source = cookie.
```
davecheney: Why does this need to be included? the cookie is generated by the backend application, we're just routing on this.

design/sticky-session-design.md
```diff
@@ -0,0 +1,170 @@
...
+kind: IngressRoute
+metadata: 
+  name: basic
+  namespace: default
+spec: 
+  virtualhost:
+    fqdn: foo-basic.bar.com
+  routes: 
+    - match: /
+      sticky-session:
```
davecheney: We already have the 'strategy:' key defined, why not add "sticky-session" as a new option?

Also, a route can have multiple services, which have a weighting -- not a load balancing strategy. Is it possible to apply stick sessions across multiple clusters -- i don't think it is so it won't be possible to apply sticky sessions across a weighted set of clusters.

design/sticky-session-design.md
```diff
@@ -0,0 +1,170 @@
...
+        ttl: time-to-live # This is only valid when source = cookie. 
+        path: cookie-path # This is only valid when source = cookie. 
+      services: 
+        - name: s1
+          port: 80  
+          strategy: RingHash # Optional. Default: RingHash
+```          
+
+For sticky sessions to be enabled on any route, the service associated to the route needs to have the 
+load balancing strategy set to a hash-based strategy (like RingHash or Maglev). The following cases need to be considered:
```
davecheney: If this is mandatory, then we shouldn't let users choose it. See my suggestion about `strategy: stick-session`.

design/sticky-session-design.md
```diff
@@ -0,0 +1,170 @@
...
+
+- No service strategy is specified: In this case, contour will set the service's strategy to "RingHash" by default.
+- Service strategy specified, but not set to a Hash-based strategy: In this case, there will be no effect of declaring sticky session.
```
davecheney: I'd prefer that we didn't let users create valid -- ie, fit through validation -- configurations that don't do what they want. This puts more load on having to generate good status messages, which we're not doing very well in Contour at the moment, so that would be solving two problems at once. I'd prefer it if the CRD data structures tried to prevent users supplying invalid configuration.

design/sticky-session-design.md
```diff
@@ -0,0 +1,170 @@
...
+
+- The Route structure in dag.go will be changed to add the sticky session config
+```go
+type Route struct {
+	...
+	// Indicates that during forwarding, the matched prefix (or path) should be swapped with this value
+	PrefixRewrite string
+
+	// Stores Sticky session Config
+	StickySession ingressroutev1.StickySessionConfig
```
davecheney: I'd prefer it if the k8s types didn't leak into the dag. I know I haven't been scrupulous in avoiding this, but it should be a goal that the dag is built from k8s types, not built out of k8s types.

design/sticky-session-design.md
rohandvora: Sure, this seems right.

design/sticky-session-design.md
rohandvora: If we specify TTL to [envoy](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route.proto#envoy-api-msg-route-routeaction-hashpolicy-header), it will generate the cookie if not present with the specified TTL. We can either mention this in the design or not support the generation part, whichever seems the right approach.

design/sticky-session-design.md
rohandvora: Sure, I understand the motive of not using k8s types in the dag. I will change this.

design/sticky-session-design.md
rohandvora: We can use the 'strategy' key to specify 'sticky-session' as an option. However since strategy is of string type, we'd need another key to specify options for this strategy. Do you suggest we have another key: *strategy-options* (or any other key perhaps) under which we could specify the option to use: cookie/ip/header?

design/sticky-session-design.md
rohandvora: Seems right. I'm not sure if users have a strong preference for using one hash-based strategy against another. If that's not a concern, we could take care of that in the implementation and not have users to specify the hash type.

Secondly, in the context of [Envoy](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/load_balancing/load_balancers#ring-hash), the types RingHash and Maglev (which contour supports as a strategy) would behave as random without a hash policy. That's another thing to consider as well.

design/sticky-session-design.md
```diff
@@ -0,0 +1,170 @@
...
+## Detailed Design
+
+### IngressRoute
+- The following changes will be made in *ingressroute.go*:   
+```go
+
+type Route struct {
+	...
+	// Indicates that sticky session will be configured for the route
+	StickySession StickySessionConfig
```
stevesloka: nit: This would need a `json` annotation.

rohandvora: Got it

design/sticky-session-design.md
```diff
@@ -0,0 +1,343 @@
...
+### Sticky-sessions configuration
+To enable sticky-sessions in a service, the `strategy` needs to be set to *StickySession*.
+By default, sticky-session will be based on **source-ip**. `stickySessionConfig` can 
+be provided in the service to specify the parameter for stickiness.
+
+#### Cookie-based stickiness
+This allows a request to be routed based on a cookie in the http request.
+The following parameters can be specified under the strategy options for cookie-based stickiness:
+- **name**: The name of the cookie to look for to determine stickiness.
+- **path**: The path for the cookie.
```
alexbrand: What is this `path` parameter used for?

rohandvora: It is used by envoy to set a path for the [cookie](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route.proto.html?highlight=strip%20headers#envoy-api-msg-route-routeaction-hashpolicy-cookie). This is used when no cookie of the specified name is found in the request. In that case, envoy will generate a cookie of that name under the specified "path".

rohandvora: Now that I think of it, if we support this in contour, then we should also include the TTL field. If a cookie is not found, a new cookie will be generated by envoy with the TTL and path set.

rohandvora: Sure. It is an optional feature provided by Envoy to set the cookie in case it is missing. We can use [passive](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route.proto.html?highlight=strip%20headers#route-routeaction-hashpolicy-cookie) cookie affinity to disable this behaviour.

design/sticky-session-design.md
```diff
@@ -0,0 +1,343 @@
...
+      services: 
+        - name: s1
+          port: 80  
+          strategy: StickySession
+          stickySessionConfig:
+            header:
+              name: header-name
+```  
+
+#### Connection-based stickiness
```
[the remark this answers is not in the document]
rohandvora: Sure

design/sticky-session-design.md
```diff
@@ -0,0 +1,343 @@
...
+    fqdn: foo-basic.bar.com
+  routes: 
+    - match: /       
+      services: 
+        - name: s1
+          port: 80  
+          strategy: StickySession
+          stickySessionConfig:
+            header:
+              name: header-name
```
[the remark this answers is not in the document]
rohandvora: Sure. I'll think of a way.

design/sticky-session-design.md
```diff
@@ -0,0 +1,343 @@
...
+
+## High-Level Design
+In contour, sticky sessions will be enabled per service. 
+If a service has sticky sessions enabled, it should be the **only service**
+under that route. Sticky sessions **do not** work with multiple services (weighted clusters).
+This means that sticky sessions cannot be used along with blue/green deployments.
+
+### Sticky-sessions configuration
+To enable sticky-sessions in a service, the `strategy` needs to be set to *StickySession*.
+By default, sticky-session will be based on **source-ip**. `stickySessionConfig` can
```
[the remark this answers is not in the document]
rohandvora: I understand your point. If we use cookie by default, can you suggest the default cookie name to use?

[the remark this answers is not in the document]
davecheney: >  If we use cookie by default, can you suggest the default cookie name to use?

This will have to be supplied by the customer, it's their cookie we're piggybacking on top of.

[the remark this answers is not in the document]
rohandvora: I meant that if in case a user does not provide any configuration associated with the sticky-session, what should be the default behaviour? Should we set an error or use a default parameter to hash on?
Note that the user will need to mention the strategy as _StickySession_ and additionally provide _stickySessionConfig_.

design/sticky-session-design.md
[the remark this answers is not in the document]
rohandvora: The cookie header can have multiple cookies. We can't hash on the entire header. One option is to use something like:
```yaml
header:
    name: cookie-name
    cookie: true
```
However, this doesn't seem intuitive enough.

design/sticky-session-design.md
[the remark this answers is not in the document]
randmonkey: I think it would be OK to provide the option of source-IP based sticky sessions, but do not set it as default.

davecheney:
Closed as duplicate #1157

PR #943, opened 2019-03-14
Gopkg.lock: fix dep check warning

davecheney:
Dep check was upset about something. I don't know what, it doesn't seem
to be critical, but this fixes the warning so better safe than sorry.


vaamarnath:
Ah, you are right! I should have re-run `dep check` after fixing the first set of complaints from `dep`. 

LGTM.

Issue #517, opened 2018-07-05
Validation: Set status on IngressRoute if TLS secret invalid


stevesloka:
If an IngressRoute specifies a secret used for TLS certificates and that secret does not exist, the `status` field on the IR should be updated with an appropriate message. 

Current implementation ignores the secret if issue: https://github.com/heptio/contour/blob/master/internal/dag/dag.go#L375

// ref https://github.com/heptio/contour/issues/458

rosskukulinski:
I think we should push to 0.7.0 as non-essential to cleanup IngressRoute work.  I believe most IngressControllers don't take an action or update status if there's an issue with Secrets, so we're not any worse than the standard.

davecheney:
If you have an implementation you can contribute I’d be grateful for the help.

so0k:
I'm seeing strange behaviour, I'm using `ingress` not `ingressRoute`, using cert-manager to get TLS secrets.

I have about 20 ingress manifests, and I had 4 TLS secrets that failed (acme challenge provider misconfiguration), contour seems to not provide any config to envoy. (cert list was empty, 8080/8443 listeners were not opened, none of the ingress worked). 

```
$ k version --short
Client Version: v1.11.6
Server Version: v1.11.6

gcr.io/heptio-images/contour:v0.8.1
docker.io/envoyproxy/envoy-alpine:v1.7.0
```

contour logs are not showing anything...
```
time="2019-02-08T12:28:40Z" level=info msg=stream_wait connection=120 context=grpc error_detail=nil resource_names="[ingress_http]" response_nonce= type_url=type.googleapis.com/envoy.api.v2.RouteConfiguration version_info=
time="2019-02-08T12:28:40Z" level=info msg=response connection=120 context=grpc count=1 error_detail=nil resource_names="[ingress_http]" response_nonce= type_url=type.googleapis.com/envoy.api.v2.RouteConfiguration version_info=
time="2019-02-08T12:28:40Z" level=info msg=stream_wait connection=120 context=grpc error_detail=nil resource_names="[ingress_http]" response_nonce= type_url=type.googleapis.com/envoy.api.v2.RouteConfiguration version_info=
```

is this expected behaviour? is this linked to this tracking issue on handling invalid TLS secrets for `ingress` resources?

PR #958, opened 2019-03-22
internal/dag: adds validation for TLS secret for IngressRoute

vaamarnath:
Sets the status on the IngressRoute as invalid if TLS secret is invalid

Updates #517

### DIFF
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -600,17 +600,25 @@ func (b *builder) computeIngressRoutes() {
 
 		var enforceTLS, passthrough bool
 		if tls := ir.Spec.VirtualHost.TLS; tls != nil {
-			// attach secrets to TLS enabled vhosts
-			m := splitSecret(tls.SecretName, ir.Namespace)
-			if sec := b.lookupSecret(m); sec != nil && b.delegationPermitted(m, ir.Namespace) {
-				svhost := b.lookupSecureVirtualHost(host)
-				svhost.Secret = sec
-				svhost.MinProtoVersion = minProtoVersion(ir.Spec.VirtualHost.TLS.MinimumProtocolVersion)
-				enforceTLS = true
-			}
 			// passthrough is true if tls.secretName is not present, and
 			// tls.passthrough is set to true.
-			passthrough = tls.SecretName == "" && tls.Passthrough
+			if tls.SecretName == "" {
+				passthrough = tls.Passthrough
+			} else {
+				// attach secrets to TLS enabled vhosts
+				m := splitSecret(tls.SecretName, ir.Namespace)
+				if b.delegationPermitted(m, ir.Namespace) {
+					sec := b.lookupSecret(m)
+					if sec == nil {
+						b.setStatus(Status{Object: ir, Status: StatusInvalid, Description: "TLS secret not found"})
+						continue
+					}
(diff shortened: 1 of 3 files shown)


vaamarnath:
I was not quite sure about the expected validation to be formed for the TLS secret for IngressRoutes. I have right now added the check to ensure that the secret exists and if not, the Status on the IngressRoute is set as invalid.

internal/e2e/lds_test.go
```diff
@@ -350,16 +350,11 @@ func TestIngressRouteTLSListener(t *testing.T) {
 	rh.OnDelete(s1)
 	assertEqual(t, &v2.DiscoveryResponse{
 		VersionInfo: "0",
-		Resources: []types.Any{
```
vaamarnath: I had to remove this validation with the changes I did. Is this an expected change? Have I made a mistake by removing this as when the secret is deleted?

internal/dag/builder.go
```diff
@@ -600,17 +600,25 @@ func (b *builder) computeIngressRoutes() {
...
-			if sec := b.lookupSecret(m); sec != nil && b.delegationPermitted(m, ir.Namespace) {
-				svhost := b.lookupSecureVirtualHost(host)
-				svhost.Secret = sec
-				svhost.MinProtoVersion = minProtoVersion(ir.Spec.VirtualHost.TLS.MinimumProtocolVersion)
-				enforceTLS = true
-			}
 			// passthrough is true if tls.secretName is not present, and
 			// tls.passthrough is set to true.
-			passthrough = tls.SecretName == "" && tls.Passthrough
+			if tls.SecretName == "" {
```
stevesloka: The check previously was to set pass through if the secret was empty and PassThrough was `true`, seems that logic is gone now and it's only checking for secretname being empty string.

internal/e2e/lds_test.go
stevesloka: I don't think you should have to remove this with the changes you made. At least this comparison is looking at the listener object. Double check that if statement that I commented on previously, it might be causing this error.

internal/e2e/lds_test.go
```diff
@@ -350,16 +350,11 @@ func TestIngressRouteTLSListener(t *testing.T) {
...
-				FilterChains: filterchain(envoy.HTTPConnectionManager("ingress_http", "/dev/stdout")),
-			}),
-		},
-		TypeUrl: listenerType,
-		Nonce:   "0",
+		Resources:   []types.Any{},
+		TypeUrl:     listenerType,
+		Nonce:       "0",
 	}, streamLDS(t, cc))
+	streamLDS(t, cc)
```
davecheney: I think this might be a mistake.

internal/e2e/lds_test.go
davecheney: Yeah, we can't remove this test, if it no longer passes then we need to dig into the change made in internal/dag and write a test to locate where the mistake is.

davecheney:
Moving this to the 0.12 milestone as #517 has moved to 0.12.

davecheney:
@vaamarnath ping. please update this or close it. Thank you.

vaamarnath:
@davecheney I am sorry that I couldn't respond to the comments given in this PR. I will close this PR for now and pick it up later when I find time.

PR #960, opened 2019-03-25
Initial draft of design for SDS support on Contour

bmadhavan:
Updates #898

Signed-off-by: Madhavan <bmadhavan@ebay.com>

### DIFF
--- a/design/sds-design.md
+++ b/design/sds-design.md
@@ -1,0 +1,68 @@
+# Support Secret Discovery Service on Contour
+
+_Status_: Draft
+
+This document outlines what changes are needed on Contour to support SDS. [0]
+
+## Goals
+
+- Implement SDS server as a gRPS service
+- Support auth.SdsSecretConfig configuration on Listener & Cluster under auth.CommonTlsContext
+- Support contour cli for SDS
+
+## Non-goals
+
+- NA
+
+## Unknowns
+
+- Changes to be done to take care of TLS delegation functionality
+- Any backward compatibility requirement
+
+## Background
+
+Currently, contour supports fetching TLS cert/keys from k8s secrets and populates
+the values for DownstreamTlsContext in LDS & UpstreamTlsContext in CDS.
+Now that Envoy 1.8 and later releases support Secret Discovery Service (SDS) to fetch 
+secrets remotely, we want to add support in Contour to parse SDS secret config
+and stream secrets using SDS GRPc service. Please refer to [1] to understand more on 
+how SDS works on Envoy.
+
+## High level design
+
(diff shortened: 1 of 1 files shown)


design/sds-design.md
```diff
@@ -0,0 +1,68 @@
+# Support Secret Discovery Service on Contour
+
+_Status_: Draft
+
+This document outlines what changes are needed on Contour to support SDS. [0]
+
+## Goals
+
+- Implement SDS server as a gRPS service
```
unicell: I still don't have the full picture regarding how things should work end to end. For instance, in the certe rotation scenario, how does SDS management server we plan to add here, works with some existing cert management tool, to complete the whole flow.

Meanwhile, is it possible to add initial support to Contour so that it can enable SDS config in Envoy, but pointing to 3rd party SDS management server, rather than implementing in Contour codebase itself. For example, run Contour for LDS that points to Istio `node_agent` and `citadel` for providing SDS.

@bmadhavan @davecheney comments?

design/sds-design.md
```diff
@@ -0,0 +1,68 @@
...
+- Support auth.SdsSecretConfig configuration on Listener & Cluster under auth.CommonTlsContext
```
unicell: No matter how big the scope is (for the question above), the change mentioned above most likely is what we need, as an initial step, for enabling SDS. And that change can be addressed first no matter we decide to bake SDS server into Contour or not.

It's going to be a change somehow similar to the below one on `Istio` side

- https://github.com/istio/api/pull/547
- https://github.com/istio/istio/pull/6453

@bmadhavan @davecheney Do we agree on that?

davecheney:
@bmadhavan thank you for your PR. Have you seen #901. Please work with @vaamarnath to consolidate your proposals. Thank you.

design/sds-design.md
bmadhavan: True. This change of supporting `auth.SdsSecretConfig configuration on Listener & Cluster` is a task that has to be done irrespective of which SDS server to use. I can submit a PR for those changes first.

davecheney:
@bmadhavan thank you for working on this, however I am closing this PR.

The design in #901 continues to evolve and we are converging on a simpler initial implementation, https://github.com/heptio/contour/issues/862#issuecomment-481543474.

Please follow and contribute to the discussion on #862.

davecheney:
Hello. After kind of running around on a design for secure envoy/contour communication in #901 I'm going to resurrect this design, with a few simplification, as its an excellent plan of action for how to achieve #898. Hopefully @bmadhavan and @vaamarnath can collaborate together on this.

PR #968, opened 2019-04-01
Allow '/' in tls secretName for certificate delegations

arminbuerkle:
Fixes heptio/contour#965

As described in the issue, the current regex does not allow namespaced tls delegation secrets as `secretName` which was introduced in contour v0.10.0.

This implementation is not perfect since it allows multiple '/' which should not be valid. But IMO it's better than removing the regex validation until something better is in place.

### DIFF
--- a/deployment/common/common.yaml
+++ b/deployment/common/common.yaml
@@ -57,7 +57,7 @@ spec:
                   properties:
                     secretName:
                       type: string
-                      pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$ # DNS-1123 subdomain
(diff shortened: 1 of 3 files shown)


deployment/common/common.yaml
```diff
@@ -57,7 +57,7 @@ spec:
                   properties:
                     secretName:
                       type: string
-                      pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$ # DNS-1123 subdomain
+                      pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?([\.\/][a-z0-9]([-a-z0-9]*[a-z0-9])?)*$ # Modified DNS-1123 subdomain
```
davecheney: Please remove this comment, its not accurate. It was never a subdomain.

davecheney:
Thank you. Please rebase to master and I'll merge this.

davecheney:
LGTM. Thank you.

PR #970, opened 2019-04-01
Add example deployments for prometheus / grafana

### LINKED ISSUE #969: Example metrics
Gimbal has a good example of how metrics can be exposed from Contour & Envoy. It would be great to bring them over to the Contour repo to allow users to also benefit.

stevesloka:
Fixes #969 by adding example deployments for prometheus & grafana. It updates the `ds-hostnet-split` example to match quickstart defined in [/docs/prometheus.md](github.com/heptio/contour/docs/prometheus.md) so that users can follow along and end up with a working solution.

This also adds sample dashboards to Grafana which provide metrics for both Contour & Envoy.

### DIFF
--- a/deployment/ds-hostnet-split/03-envoy.yaml
+++ b/deployment/ds-hostnet-split/03-envoy.yaml
@@ -18,8 +18,8 @@ spec:
       annotations:
         prometheus.io/scrape: "true"
         prometheus.io/port: "8002"
-        prometheus.io/path: "/stats"
-        prometheus.io/format: "prometheus"
+        prometheus.io/statsdport: "9102"
+        prometheus.io/path: "/stats/prometheus"
       labels:
         app: envoy
     spec:
@@ -57,7 +57,20 @@ spec:
         lifecycle:
           preStop:
             exec:
-              command: ["wget", "-qO-", "http://localhost:9001/healthcheck/fail"] 
+              command: ["wget", "-qO-", "http://localhost:9001/healthcheck/fail"]
+      - name: statsd-sink
+        image: prom/statsd-exporter:v0.6.0
+        command: 
+            - "/bin/statsd_exporter"
+        args:
+            - "-statsd.mapping-config=/etc/statsd/statsd.yaml"
+        ports:
+        - containerPort: 9102
+          protocol: TCP
+          name: metrics
+        volumeMounts:
+        - name: statsd
+          mountPath: /etc/statsd
       dnsPolicy: ClusterFirstWithHostNet
       hostNetwork: true
       initContainers:
@@ -68,6 +81,7 @@ spec:
         - $(CONTOUR_SERVICE_HOST)
         - --xds-port
         - $(CONTOUR_SERVICE_PORT)
+        - --statsd-enabled
         command:
         - contour
         image: gcr.io/heptio-images/contour:master
@@ -80,4 +94,7 @@ spec:
       volumes:
         - name: contour-config
           emptyDir: {}
+        - name: statsd
+          configMap:
+            name: envoy-statsd
       restartPolicy: Always
--- a/deployment/prometheus/02-prometheus-configmap.yaml
+++ b/deployment/prometheus/02-prometheus-configmap.yaml
@@ -1,0 +1,357 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: prometheus
+  namespace: contour-monitoring
+data:
+  prometheus.yml: |-
+    # A scrape configuration for running Prometheus on a Kubernetes cluster.
+    # This uses separate scrape configs for cluster components (i.e. API server, node)
+    # and services to allow each to use different authentication configs.
+    #
+    # Kubernetes labels will be added as Prometheus labels on metrics via the
+    # `labelmap` relabeling action.
+    #
+    # If you are using Kubernetes 1.7.2 or earlier, please take note of the comments
+    # for the kubernetes-cadvisor job; you will need to edit or remove this job.
+
+    rule_files:
+    - '/etc/prometheus-rules/*.rules'
+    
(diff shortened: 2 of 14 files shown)


stevesloka:
Thanks to @jipperinbham fixed a quick issue with the statsd forwarder. =)

deployment/prometheus/02-prometheus-configmap.yaml
```diff
@@ -0,0 +1,357 @@
...
+        regex: __meta_kubernetes_pod_label_(.+)
+      - source_labels: [__meta_kubernetes_namespace]
+        action: replace
+        target_label: kubernetes_namespace
+      - source_labels: [__meta_kubernetes_pod_name]
+        action: replace
+        target_label: kubernetes_pod_name
+      metric_relabel_configs:
+      - source_labels: [envoy_cluster_name]
+        regex: '(.+?)\/(.+?)\/(.*)'
```
jipperinbham: need to replace `\/` with `_` here as well and the two spots below

davecheney:
I'm not qualified to review this but i'm glad to see these prometheus dashboards included.

deployment/ds-hostnet-split/03-envoy.yaml
```diff
@@ -57,7 +57,20 @@ spec:
         lifecycle:
           preStop:
             exec:
-              command: ["wget", "-qO-", "http://localhost:9001/healthcheck/fail"] 
+              command: ["wget", "-qO-", "http://localhost:9001/healthcheck/fail"]
+      - name: statsd-sink
+        image: prom/statsd-exporter:v0.6.0
+        command: 
+            - "/bin/statsd_exporter"
```
rata: Just curious: envoy 1.10.0 has just been released (like 2 hours ago) and supports histograms in prometheus export now (https://github.com/envoyproxy/envoy/pull/5601 you shared it on slack a few days ago :-D) . Does the sidecar add more metrics if envoy 1.10 is used?

Also, I'm not sure when contour plans to upgrade to envoy 1.10, so even if it does, this can be totally relevant :-)

davecheney:
@stevesloka @rata @jipperinbham can you please finalise the review of this PR. I'd like it to either land or move back into development by EOD April 12. Thank you.

deployment/ds-hostnet-split/03-envoy.yaml
stevesloka: @rata yes Envoy v1.10 has Histograms enabled so we can remove the statsd forwarder bit. I'm not sure when we'll upgrade to v1.10, we typically wait a bit to let any issues shake out. I'd like to still merge this for v1.9.1 and then we'll update with a future PR once we move to envoy v1.10

stevesloka:
I think this is good to merge @davecheney, there is an issue with how the metrics are visualized which is tracked in #1008

deployment/ds-hostnet-split/03-envoy.yaml
```diff
@@ -18,8 +18,8 @@ spec:
       annotations:
         prometheus.io/scrape: "true"
         prometheus.io/port: "8002"
-        prometheus.io/path: "/stats"
-        prometheus.io/format: "prometheus"
+        prometheus.io/statsdport: "9102"
+        prometheus.io/path: "/stats/prometheus"
```
rata: Am I missing something or this will scrape only envoy metrics? stastd, I think, will gather only envoy and so will port `8002`. Right?

I think metrics for the contour container (on endpoint `/metrics` and port `8000`) are not gathered by this. Am I missing something? Hopefully I am :)

I will try to create a cluster and see if those are missing or not, it's too late here now to try this :-(

deployment/ds-hostnet-split/03-envoy.yaml
rata: Thanks a lot for the info! :)

deployment/ds-hostnet-split/03-envoy.yaml
stevesloka: @rata Contour metrics are scraped on the Contour deployment. This example is applied to the split model where Contour & Envoy are not colocated in the same pod.

rata: Ohh, I see. Silly me, thanks! :)

davecheney:
@stevesloka can you merge this or are you blocked on #1008?

deployment/prometheus/03-prometheus-deployment.yaml
```diff
@@ -0,0 +1,128 @@
...
+  annotations:
+    prometheus.io/scrape: 'true'
+  name: prometheus
+  namespace: contour-monitoring
+spec:
+  ports:
+  - protocol: TCP
+    name: prometheus
+    port: 9090
+  - protocol: TCP
```
alexbrand: Should we remove this? Doesn't seem like we should be targetting the alertmanager in the prometheus service.

stevesloka:
@rata thanks for the reply! I agree with you that this needs some work. We did most of this work in [Gimbal](https://github.com/heptio/gimbal/) which only uses the split-model deployment because of network perf reasons. My goal with this was to enable folks to start to look at what we had for metrics and stir inspiration (like you've been doing). =)

I think once we can move to Envoy 1.10 then we can drop the statsd_forwarder and make all this much simpler. I'd like to merge this, then open a new issue to track the work that needs to be done once we're at Envoy v1.1.0.

deployment/prometheus/03-prometheus-deployment.yaml
stevesloka: Thanks for the review @alexbrand, I just dropped the alertmanager reference.

rata:
@stevesloka sounds good to me. Let me know if I can help with the related issues :-)

PR #972, opened 2019-04-02
deployment/render.sh: Run from any CWD

rata:
This just makes it possible to run the script from any current working
directory. Currently it should be run from /development only.

Signed-off-by: Rodrigo Campos <rodrigo@kinvolk.io>

### DIFF
--- a/deployment/render.sh
+++ b/deployment/render.sh
@@ -1,5 +1,8 @@
 #! /bin/sh
 
+# Run correctly from any CWD
+cd $(dirname $0)
+
 ls ds-grpc-v2/*.yaml | xargs cat common/gen-warning.yaml > render/daemonset-rbac.yaml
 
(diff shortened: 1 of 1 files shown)


rata:
@davecheney sorry, I'm quite new to contour and not sure I follow. Can you please elaborate?

In any case, I'm okay closing this PR. I just didn't follow the reasoning. If you can elaborate so I don't do the same in the future it would be very helpful :-)

davecheney:
> In any case, I'm okay closing this PR. I just didn't follow the reasoning. If you can elaborate so I don't do the same in the future it would be very helpful :-)

To put it simply, this ain't that broke, compared to many other things that are, and the workaround for its moderately brokenness isn't that bad, so let's put effort into more important thinks.

Thanks

Dave

PR #978, opened 2019-04-05
Add upstream support for HTTP/1.1 + TLS

### LINKED ISSUE #406: Add upstream support for HTTP/1.1 + TLS annotation to the service
We need a way to explicitly configure the L7 protocol for a service port to HTTP/1.1 + TLS. Something like
```
contour.heptio.com/upstream-protocol.https: "admin, 443"
```
---
Update, I think this annotation should be
```
contour.heptio.com/upstream-protocol.tls: "admin, 443"
```

stevesloka:
Fixes #406 by adding a service annotation (tls) which allows for requests which terminate at Envoy to proxy via tls to an upstream.


internal/envoy/auth.go
```diff
@@ -16,10 +16,17 @@ package envoy
 import "github.com/envoyproxy/go-control-plane/envoy/api/v2/auth"
 
 // UpstreamTLSContext creates an ALPN h2 enabled TLS Context.
-func UpstreamTLSContext() *auth.UpstreamTlsContext {
+func UpstreamTLSContext(protocol string) *auth.UpstreamTlsContext {
```
davecheney: I'm concerned about adding a parameter here because its a singular parameter, not a list. Ideally these protocols are the _alpn_ protocols, but we're obverloading that behaviour on line 27 to use "tls" as a magic marker to return a blank list of alpn protocols.

Try this instead

```go
func UpstreamTLSContext(protocols ...string) *auth.UpstreamTlsContext {
    return &auth.UpstreamTlsContext{
        CommonTlsContext: &auth.CommonTlsContext{
            AlpnProtocols: protocols,
        },
   },
}
```
Then in see next comment

internal/envoy/cluster.go
```diff
@@ -43,9 +43,12 @@ func Cluster(s dag.Service) *v2.Cluster {
 
 func httpCluster(service *dag.HTTPService) *v2.Cluster {
 	c := cluster(&service.TCPService)
+
 	switch service.Protocol {
+	case "tls":
+		c.TlsContext = UpstreamTLSContext(service.Protocol)
```
davecheney: Then we can make this method clearer but removing the fallthrough.

case "tls":
     c.TlsContext = UpstreamTLSContext()
case "h2":
     c.TlsContext = UpstreamTLSContext("h2")
case "h2c":
     c.TlsContext = UpstreamTLSContext("h2")
     c.Http2ProtocolOptions = &core.Http2ProtocolOptions{}
}
```

Issue #983, opened 2019-04-05
Envoy CVE / Upgrade to v1.9.1


stevesloka:
**What steps did you take and what happened:**
There are two CVE's discovered with Envoy v1.9.0. You can read about the details [here](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/envoy-announce/VoHfnDqZiAM/zx996ROUBgAJ
): 

> CVE-2019-9900: Those most likely to be affected will have HTTP/1.1 traffic from untrusted endpoints and will make use of header based matching for access control or routing (including ext_authz, rate limiting service and similar inbuilt filters).

> CVE-2019-9901: Any deployment where path-based access control or routing is utilized is likely affected, in particular where routing decisions are intended to block or allow access (e.g. disabling an /admin handler). This will also depend on upstream path normalization semantics. If you use Envoy’s RBAC or ext_authz filter, together with prefix path matching or suffix header matching, it is likely that you are directly affected. If you have custom filters that rely on the request path or headers, you should will need to perform additional risk assessment.

Since Contour v0.10.0 and earlier does not expose Envoy's RBAC or  ext_authz filter it's likely that there risk is low, however, it would still be good to upgrade. 

There are two items this issue needs to address:

- [x] Upgrade manifests to Envoy v1.9.1
- [x] Enable path normalization via either the HTTP Connection Manager configuration HttpConnectionManager.normalize_path field or the http_connection_manager.normalize_path runtime option to mitigate CVE-2019-9901

PR #984, opened 2019-04-05
Upgrade Envoy to v1.9.1

stevesloka:
Updates #983 by updating the example deployment files to use Envoy v1.9.1

### DIFF
--- a/deployment/ds-hostnet-split/03-envoy.yaml
+++ b/deployment/ds-hostnet-split/03-envoy.yaml
@@ -33,7 +33,7 @@ spec:
         - node0
         command:
         - envoy
-        image: docker.io/envoyproxy/envoy-alpine:v1.9.0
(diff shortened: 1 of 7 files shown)


deployment/ds-hostnet-split/03-envoy.yaml
```diff
@@ -33,7 +33,7 @@ spec:
         - node0
         command:
         - envoy
-        image: docker.io/envoyproxy/envoy-alpine:v1.9.0
+        image: docker.io/envoyproxy/envoy:v1.9.1
```
so0k: any reason why you're switching from alpine to ubuntu based envoy image for the hostnet deployment?

davecheney:
You can stay with the alpine image if you prefer. These deployments are just guides.

Issue #876, opened 2019-02-05
Switch from proto.struct to proto.any configuration for bootstrap configuration


davecheney:
go-control-plane v0.6.4 (?) introduced a new way of specifying opaque filter config; rather than a proto.struct, use a proto.any. Both can be constructed by encoding an gRPC proto message, but the former involves an icky conversion via jsonpb.

The latter should give us more type safety -- we avoid using the jsonpb/struct hack in internal/envoy by constructing struct types directly, which is no better than the previous mechanism of using text/template. 

Switching to the config_any mechanism should give us more insulation from the changes to the grpc data types.

Blocked on:

- [x] #999 
- [x] https://github.com/envoyproxy/go-control-plane/issues/175

vaamarnath:
@davecheney I am willing to pick this work. I would love if you could give me some direction on how to proceed with this work. 
I have gone through the code in `internal/contour/config.go` which has the template `yamlConfig`. I saw that it is used to create the yaml configs. Though, I didn't see where `WriteYAML()` is being invoked to understand the execution flow. Is it invoked from the envoy for configuring itself? For fixing this issue, is the requirement to change this implementation to protobuf using proto.any?

davecheney:
This issue is already assigned, please choose another.

davecheney:
Bumping til 0.12. @vaamarnath you're welcome to work on this, but please not until the 0.12 cycle.

davecheney:
As far as I can see this is complete, all filter configurations are using TypedConfig.

davecheney:
Opps, one more. Will send a final PR

PR #989, opened 2019-04-08
go.mod: upgrade to go-control-plane v0.6.9

### LINKED ISSUE #933: Update envoyproxy/go-control-plane
See if there is a newer version of go-control-plane. Beware that the latest go-control-plane may require a newer envoy than 0.11's envoy 1.9.0.

davecheney:
Fixes #933

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.12
 
 require (
 	cloud.google.com/go v0.35.1 // indirect
-	github.com/envoyproxy/go-control-plane v0.6.8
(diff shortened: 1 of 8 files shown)


davecheney:
This is related to #983 and #876. Switching to a typed configuration allows us to write unit tests to assert that `http_connection_manager.normalize_path` is always enabled.

davecheney:
Merge if this is compatible with Envoy 1.9.1. Defer to 0.12 milestone if we need Envoy 1.10

PR #997, opened 2019-04-09
Dockerize CRD code generation to fix regression with Go modules

### LINKED ISSUE #996: make updategenerated is broken after Go module changes

With a clean clone (no `vendor` folder`), `make updategenerated` fails because of

```
Updating CRD generated code...
hack/update-generated-crd-code.sh: line 22: hack/../vendor/k8s.io/code-generator/generate-groups.sh: No such file or directory
Makefile:95: recipe for target 'updategenerated' failed
make: *** [updategenerated] Error 127
```

`make updategenerated` passes without error

**Environment:**

- Contour version:

unicell:
Fixes #996

Signed-off-by: Qiu Yu <unicell@gmail.com>

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,5 @@
 /.git
 /vendor
 /_integration
-/hack
 /contour
 *.test
--- a/Makefile
+++ b/Makefile
@@ -91,9 +91,15 @@ render:
 	@echo Rendering deployment files...
 	@(cd deployment && bash render.sh)
 
-updategenerated:
+pwd := $(shell pwd)
+repo_dir := /go/src/github.com/heptio/contour
+crd_gen := docker run --rm -v $(pwd)/apis:$(repo_dir)/apis -w $(repo_dir) crd_gen_container:latest hack/update-generated-crd-code.sh
+updategenerated: crd_gen_container
 	@echo Updating CRD generated code...
-	@(bash hack/update-generated-crd-code.sh)
+	@$(crd_gen) 2>/dev/null
+
+crd_gen_container:
+	docker build . -f ./hack/crd-gen-dockerfile -t crd_gen_container:latest
 
 gofmt:
 	@echo Checking code is gofmted
--- a/hack/crd-gen-dockerfile
+++ b/hack/crd-gen-dockerfile
@@ -1,0 +1,19 @@
+FROM golang:1.12.1 AS build
+
+ENV CODE_GEN_VER 1.12.6
+ENV CODE_GEN_PATH $GOPATH/src/k8s.io/code-generator
+RUN mkdir -p ${CODE_GEN_PATH} && \
+    curl -sSLO https://github.com/kubernetes/code-generator/archive/kubernetes-${CODE_GEN_VER}.tar.gz && \
+    tar xzf kubernetes-${CODE_GEN_VER}.tar.gz -C ${CODE_GEN_PATH} --strip-components=1
+
+WORKDIR ${CODE_GEN_PATH}
+ENV GOFLAGS -mod=vendor
+RUN bash -c "go install ./cmd/{defaulter-gen,client-gen,lister-gen,informer-gen,deepcopy-gen}"
+
+WORKDIR /go/src/github.com/heptio/contour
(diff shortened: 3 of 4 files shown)


davecheney:
Thank you for your PR. In future, please do not send code until the design has been discussed on the issue. FWIW, adding a `go mod vendor` before running update generated would probably have worked.

unicell:
@davecheney Thanks for the quick comment! Please consider it as a PoC as I hit the issue myself and tried several ways to remediate. Hence sent out the PR for further comments / feedbacks.

`go mod vendor` ignores all non-go files, hence the `generate-groups.sh` file from `k8s.io/code-generator` is gone missing. Meanwhile, code-generator project itself not seem to be go module ready[1], so figured Docker approach could be a good workaround till there's a better solution.

[1] https://github.com/kubernetes/code-generator/issues/57

davecheney:
> go mod vendor ignores all non-go files, hence the generate-groups.sh file from k8s.io/code-generator is gone missing.

Sigh, I really wish dependency managers would get out of the business of deciding what is and is not source code.

unicell:
Thanks for the comments. Fair enough points.

I understand that we don't want to carry the maintenance burden for a less infrequent thing. But before we close on this and move to the document approach, feel like several other points worth to be mentioned just to ensure that we all on the same page.

The whole idea of the proposed change is to have

1. repeatable build

For instance, same `code-generator` version has to be used, no excess code generated after running the code generation by anyone who cloned the project, etc.

2. easy to repeat build process

After the introduction of Go modules, and the fact of `code-generator` itself not Go module ready, it requires Contour to be placed under GOPATH[1]. The dependency of code-generator itself needs to be properly ensured (for that specific version of code-generator). Plus the fact `generate-groups.sh` actually runs `go install`[2].

All that craziness and the knowledge for getting things to build can be contained in a Docker image. If we choose to document all these details into text, that'll be lengthy for all the details in this PR, but not runnable.

I can close the PR if you still feel document is the right way to do. Optionally, if you don't like the way this change is touching the makefile targets, I can also add Dockfile to apis/generated and add document regarding how to use that Docker image for CRD generation and leave Makefile as is.

[1] https://github.com/kubernetes/code-generator/issues/57
[2] https://github.com/kubernetes/code-generator/blob/kubernetes-1.12.6/generate-groups.sh#L53

davecheney:
Thank you for your reply.

Fundamentally this debate is why we have our [_talk, then code_](https://github.com/heptio/contour/blob/master/CONTRIBUTING.md#before-you-submit-a-pull-request) policy. I am grateful for your contributions, but now we're in the situation where you have done some work which you are emotionally invested in and I am in the position of feeling like the bad guy because I am refusing work you have already done. 

If you had discussed your implementation on #996 then I would have told you that I don't want to see docker mixed into this. I may have mentioned that I'm not really impressed with go modules and this bug was something that made me consider reverting to dep. 

@glerchundi has proposed #1010 which is a simpler fix and offers less maintenance burden in the future. I'm not happy that they did this without declaring their intentions on #996, I do not wish to create a combative environment on this project.

I've explained that compared to things which we do all the time during a development cycle, like update the envoy grpc definitions, regenerating the CRD definitions happens far less frequently. In an ideal world rebuilding the generated code is something that might happen on each build so that we never check it in, but being pragmatic the generated code has never once caused a problem when bumping the kubernetes dependencies and the envoy gRPC definitions are a constant pain in the rear. Given the velocity goals of this project -- small incremental releases every month or so -- I just can't get that excited about investing engineering time in being able to hermetically reproduce the CRD definitions.

As I said previously I would be perfectly fine with a simple written process in a markdown file, I think that is a reasonable tradeoff of complexity for its infrequent usage. However @glerchundi has proposed #1010 which is an overall simpler solution than this PR. 

However, before I review #1010 I want to conclude this discussion. You may not agree with my decision here, but I hope to at least explain my reasoning clearly, and hopefully direct your attentions to issues which will have a higher payoff for Contour users.

unicell:
Thank you Dave for the detailed explanation. It totally makes sense to me. I'll close this PR and instead watch #1010 for how it goes.

davecheney:
Thank you for your understanding.

Issue #959, opened 2019-03-24
Contour dont start, started to fail after automatic restart


vongohren:
**What steps did you take and what happened:**
I did nothing, it just started to fail on prod. It happened after an automatic restart.
Dev was working fine, but then I restarted that as well, then it started to fail with the same problems. So I have not done anything for this to happen

**What did you expect to happen:**
I exepcted the clusters to run and the contour container to work smoothly.

**Anything else you would like to add:**
This is the logs i get
```
 time="2019-03-24T12:17:31Z" level=info msg="args: [serve --incluster]"
 E 
 time="2019-03-24T12:17:31Z" level=info msg=started context=grpc
 E 
 time="2019-03-24T12:17:31Z" level=info msg="waiting for cache sync" context=coreinformers
 E 
 time="2019-03-24T12:17:31Z" level=info msg=started context=coreinformers
 E 
 time="2019-03-24T12:17:31Z" level=info msg="waiting for cache sync" context=contourinformers
 E 
 time="2019-03-24T12:17:31Z" level=info msg=started context=contourinformers
 E 
 time="2019-03-24T12:17:31Z" level=info msg=started address="127.0.0.1:6060" context=debugsvc
 E 
 time="2019-03-24T12:17:31Z" level=info msg=started address="0.0.0.0:8000" context=metricsvc
 E 
 time="2019-03-24T12:17:31Z" level=info msg="forcing update" context=HoldoffNotifier last update=2562047h47m16.854775807s
 E 
 pkg/mod/k8s.io/client-go@v0.0.0-20190226174127-78295b709ec6/tools/cache/reflector.go:95: Failed to list *v1beta1.TLSCertificateDelegation: tlscertificatedelegations.contour.heptio.com is forbidden: User "system:serviceaccount:heptio-contour:contour" cannot list tlscertificatedelegations.contour.heptio.com at the cluster scope E 
 pkg/mod/k8s.io/client-go@v0.0.0-20190226174127-78295b709ec6/tools/cache/reflector.go:95: Failed to list *v1beta1.TLSCertificateDelegation: tlscertificatedelegations.contour.heptio.com is forbidden: User "system:serviceaccount:heptio-contour:contour" cannot list tlscertificatedelegations.contour.heptio.com at the cluster scope E 
 log: exiting because of error: log: cannot create log: open /tmp/contour.contour-74cc4d58c6-gt9f8.unknownuser.log.ERROR.20190324-121731.1: no such file or directory
 E 
[...]

PR #1004, opened 2019-04-09
Always let glog write to stderr not files

unicell:
Related to #959

Client-go uses glog even though Contour by itself doesn't. In the
failure scenario, such as CRD types not registered, glog in client-go
attempts to log to files under /tmp, which may not even exist (e.g in
`scratch` Docker image) or not accessible (container started with
non-root user etc.). And when that happens it'll crash the whole Contour
process which should not happen.

This change overrides the glog flag to let it always dumping to stderr,
avoid logging to files which is not desired in container environment

Signed-off-by: Qiu Yu <unicell@gmail.com>


davecheney:
God damn i hate glog. Thank you.

PR #1009, opened 2019-04-10
Removed duplicate "will"

lostllama:


davecheney:
I'm sorry to do this, but you need to sign your commit, even for such a tiny change.

Otherwise @stevesloka can roll this into #873

lostllama:
@davecheney Uh, I'll leave it to you guys. I made the edit via Chrome and can't really be bothered checking out the repo to remove a single word :)

davecheney:
Thanks. I understand.

davecheney:
I'm going to merge this without a CLA. There is no way I cannot create a competing patch which is not derived from this patch as its a one word change.

PR #1010, opened 2019-04-10
modules: pin dependencies and fix updategenerated

### LINKED ISSUE #996: make updategenerated is broken after Go module changes

glerchundi:
Closes #996

### DIFF
--- a/hack/update-generated-crd-code.sh
+++ b/hack/update-generated-crd-code.sh
@@ -15,10 +15,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-HACK_DIR=$(dirname "${BASH_SOURCE}")
-REPO_ROOT=${HACK_DIR}/..
-
-${REPO_ROOT}/vendor/k8s.io/code-generator/generate-groups.sh \
+VERSION=$(go list -m all | grep k8s.io/code-generator | rev | cut -d"-" -f1 | cut -d" " -f1 | rev)
+TMP_DIR=$(mktemp -d)
+git clone https://github.com/kubernetes/code-generator.git ${TMP_DIR}
+(cd ${TMP_DIR} && git reset --hard ${VERSION} && go mod init)
+${TMP_DIR}/generate-groups.sh \
   all \
   github.com/heptio/contour/apis/generated \
   github.com/heptio/contour/apis \
--- a/Makefile
+++ b/Makefile
@@ -56,13 +56,13 @@ local: $(LOCAL_BOOTSTRAP_CONFIG)
 		--service-cluster cluster0
 
 staticcheck:
-	@go get honnef.co/go/tools/cmd/staticcheck
+	go install honnef.co/go/tools/cmd/staticcheck
 	staticcheck \
 		-checks all,-ST1003 \
 		$(PKGS)
 
 misspell:
-	@go get github.com/client9/misspell/cmd/misspell
+	go install github.com/client9/misspell/cmd/misspell
 	misspell \
 		-i clas \
 		-locale US \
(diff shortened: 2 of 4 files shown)


davecheney:
@glerchundi thank you for your PR. 

While I am thankful for your PR, I prefer it to #997, @unicell already was working on this change and I don't want to encourage duelling PRs, that is not productive.

Please let me resolve #997 with @unicell and I'll review this changer afterwards.

glerchundi:
Ups, i didn't noticed that there was a PR already sent. 

> @unicell already was working on this change and I don't want to encourage duelling PRs, that is not productive.

I agree, feel free to do whatever you want with this PR.

hack/update-generated-crd-code.sh
```diff
@@ -15,13 +15,14 @@
...
 
-HACK_DIR=$(dirname "${BASH_SOURCE}")
-REPO_ROOT=${HACK_DIR}/..
-
-${REPO_ROOT}/vendor/k8s.io/code-generator/generate-groups.sh \
+VERSION=$(go list -m all | grep k8s.io/code-generator | rev | cut -d"-" -f1 | cut -d" " -f1 | rev)
+TMP_DIR=$(mktemp -d)
+git clone https://github.com/kubernetes/code-generator.git ${TMP_DIR}
+(cd ${TMP_DIR} && git reset --hard ${VERSION})
+${TMP_DIR}/generate-groups.sh \
```
unicell: Please note, if `contour` is checked out outside of $GOPATH, `code-generator` would still generate the CRD code into $GOPATH/src/github.com/heptio/contour/apis anyway due to https://github.com/kubernetes/code-generator/issues/57

glerchundi: Yep, but i think this is assumed as per the [contributing guidelines](https://github.com/heptio/contour/blob/master/CONTRIBUTING.md).

unicell: Cool, then we should be good

stevesloka:
I'm trying to make some changes to the CRD and regen a new client but running into issues. 

I took the code from this PR and added to my project. When I run a `make updategenerated` I get the following output:

```
Generating deepcopy funcs
Generating clientset for contour:v1beta1 at github.com/heptio/contour/apis/generated/clientset
**EvalSymlinks: too many links**
Generating listers for contour:v1beta1 at github.com/heptio/contour/apis/generated/listers
Generating informers for contour:v1beta1 at github.com/heptio/contour/apis/generated/informers
```

Everything seems ok, but did notice an issue on client-go where a similar problem (i.e. too many links) was found and it was a version mismatch: https://github.com/kubernetes/code-generator/issues/37

glerchundi:
@davecheney fixed by using `go mod init` (good catch!) and tested that it works in 1.12 also. I tried to avoid the use of `GO111MODULE` in order to be this a little bit more future-proof.

glerchundi:
@stevesloka i cannot reproduce, can you look at that for me?

stevesloka:
I don't think the $VERSION variable returns the right version, for me, it returns `generator`. We need to pin the generator to the same version as apimachinery / api & client-go. I set VERSION to `kubernetes-1.14.1` and it worked finally in addition to setting versions in go.mod as well for the previously mentioned

unicell:
Since @stevesloka is mentioning the $VERSION variable, not sure if it is a concern, the code-generator dependency we're adding to the `go.mod` will be removed by `go mod tidy` as code-generator not really being referenced from the go code.

glerchundi:
@stevesloka i tried to use the same version that appears on the go.mod. Go
modules support semver tags or directly the sha (prefixed by a
v0.0.0-somekindoftimestamp-sha). That snippet tries to extract the semver
or the sha. I cannot remember which version the sha I referenced on the go
modules points to but, if i’m not wrong, it was the same version used by
the other k8s dependencies, 1.12.6?

That being said, i dont know if there is a real problem in pointing the
generator to a newer version than the one used by the other k8s deps. as it
seems to fix, at least on your environment.

glerchundi:
@unicell go mod tidy is not going to purge those deps as these appear in
the tools.go file this PR adds.

stevesloka:
Testing this again I think this is working. I wiped my repo and started again and got my CRD's working. Thanks!

glerchundi:
You're welcome, thank you all for making this project as strong and reliable as possible with these kind of reviews!

PR #1011, opened 2019-04-10
remove AES128-* and AES256-* from permitted ciphers list

### LINKED ISSUE #823: Are you open to making TLS ciphers configurable?
Generally I'm reluctant to enter the custom cipher configuration game - the hunt for the perfect configuration to get an `A+` on qualys ssl labs is a mugs game.

However, we've hit a situation where we'd prefer to sacrifice some backwards compatibility to get perfect forward secrecy in place for all connections. Rats.

Are you open to a patch to make the ciphers enabled in envoy configurable?

The following comment in the ingress route design doc (`design/ingressroute-design.md`) sounds promising:

> other properties like cipher suites may be added later

We're currently still using contour 0.7 with traditional Ingress objects, but I guess we'll have to switch to using IngressRoute eventually, and if they make configurable ciphers an option then that'd be a bonus.

yob:
The full list of ciphers that envoy permits by default are listed at [here](https://www.envoyproxy.io/docs/envoy/v1.9.1/api-v2/api/v2/auth/cert.proto.html?highlight=ciphers).

The defaults rate an "A" on qualy's ssl labs, however the AES128-* and AES256-* ciphers are marked as "weak". This appears to be for a combination of two reasons:

1. They don't support [Forward Secrecy](https://en.wikipedia.org/wiki/Forward_secrecy
)
2. They are easy to implement wrongly, exposing users to the [ROBOT attack](https://robotattack.org/
). Presumably boringssl (which envoy uses) has implemented them correctly, but it seems the general [recommendation is to avoid them if possible, just in case](https://community.qualys.com/thread/17971-tlsrsawithaes256cbcsha-comes-to-be-weak-cipher#comment-39790).

Removing these ciphers doesn't improve the overall qualy's score, but it does remove the "weak" flags from the list of supported ciphers.

In practice, this is unlikely to hurt client compatibility for users of contour. Most of the clients that might need to use the removed ciphers (like windows XP, java 6+7, and android <= 4.3) are unable to connect without TLS 1.0, which contour never enables.

This cipher list is also ignored when a connection negotiates TLS 1.3. I'm not sure why that is, but the [envoy docs seems pretty clear](https://www.envoyproxy.io/docs/envoy/v1.10.0/api-v2/api/v2/auth/cert.proto.html?highlight=ciphers):

> this setting has no effect when negotiating TLS 1.3

Closes #823

### DIFF
--- a/internal/envoy/auth.go
+++ b/internal/envoy/auth.go
@@ -18,6 +18,32 @@ import (
 	"github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
 )
 
+var (
+	// This is the list of default ciphers used by contour 1.9.1. A handful are
+	// commented out, as they're arguably less secure. They're also unnecessary
+	// - most of the clients that might need to use the commented ciphers are
+	// unable to connect without TLS 1.0, which contour never enables.
+	//
+	// This list is ignored if the client and server negotiate TLS 1.3.
+	//
+	// The commented ciphers are left in place to simplify updating this list for future
+	// versions of envoy.
+	ciphers = []string{
+		"[ECDHE-ECDSA-AES128-GCM-SHA256|ECDHE-ECDSA-CHACHA20-POLY1305]",
+		"[ECDHE-RSA-AES128-GCM-SHA256|ECDHE-RSA-CHACHA20-POLY1305]",
+		"ECDHE-ECDSA-AES128-SHA",
+		"ECDHE-RSA-AES128-SHA",
+		//"AES128-GCM-SHA256",
+		//"AES128-SHA",
+		"ECDHE-ECDSA-AES256-GCM-SHA384",
+		"ECDHE-RSA-AES256-GCM-SHA384",
+		"ECDHE-ECDSA-AES256-SHA",
+		"ECDHE-RSA-AES256-SHA",
+		//"AES256-GCM-SHA384",
+		//"AES256-SHA",
+	}
+)
+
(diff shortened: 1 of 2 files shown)


yob:
Here are some screenshots of relevant qualys tests.

1. [Vanilla contour 0.11.0](https://user-images.githubusercontent.com/8132/55888463-bea1d180-5bf2-11e9-946f-efc85a5d7b8b.png) 
2. [This branch](https://user-images.githubusercontent.com/8132/55888493-cc575700-5bf2-11e9-9738-16f20ab7ed7b.png)

Note the ciphers section in the test for this branch lists no "weak" ciphers:

Screenshot from 2019-04-11 00-44-13

Also note the "Not simulated clients (Protocol mismatch)" list is identical in both tests. Those old clients can't connect without TLS 1.0, so knocking out some ciphers they might've depended on hasn't effected their ability to connect to contour.

davecheney:
@yob thank you for picking this up and your detailed analysis. I'm going to run this PR by @mattmoyer and @xstevens to see what they thing but in principal this delivers on the goals of #823 -- better SSL labs score without having to delegation this problem to each user to solve themselves.

yob:
Thanks @davecheney.

If it turns out some users need these ciphers, let me know. Maybe we could add an opt-in annotation `tls-ciphers-list: compatible` or something, that enables a few extra ciphers without requiring users to manage the list themselves.

Issue #70, opened 2017-11-28
Support adding / removing headers


davecheney:
Add support for adding / removing headers at the Ingress, namespace, and default level

drobinson123:
I'm interested in implementing ingress level support. Is this annotation format acceptable?

```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: example
  annotations:
    contour.heptio.com/response-headers-to-add: |
      Strict-Transport-Security "max-age=31536000; includeSubDomains"
      Public-Key-Pins: max-age=2592000; pin-sha256="E9CZ9INDbd+2eRQozYqqbQ2yXLVKB9+xcprMF+44U1g="
    contour.heptio.com/response-headers-to-remove: |
      X-Powered-By
spec:
  ...
```

davecheney:
Thanks for offering to work on this. My initial reaction to the annotation syntax is it’s pretty ugly, yaml embedded in annotations is quite verbose and not something people are going to get right first time. 

One way around this might be to encode the header name in the annotation key, ie

contour.heptio.com/add-header-max-age: “...”

Where add-header- is a known prefix and there is a form for converting the suffix into a http header. This should work as http headers are not case sensitive. 

The other question is should this be per ingress or per route. I can see an argument for per route.

drobinson123:
Encoding the header in the annotation will do, I have no preference.

Headers are appended to requests/responses in [a specific order](https://www.envoyproxy.io/docs/envoy/v1.6.0/configuration/http_conn_man/headers#config-http-conn-man-headers-custom-request-headers), so we may be able to do both ingress and route. I'll experiment and see how it could work.

One other question, I assume we only care about response headers? I don't have any use-cases for adding headers to requests.

davecheney:
I see the need for this, but tbh I’m turned off by how ugly this will be to fit into the ingress object.

zxvdr:
@davecheney, can you please provide some guidance regarding the annotation format? Unless I'm mistaken, the format I used in https://github.com/heptio/contour/pull/408 is the format [you suggested](https://github.com/heptio/contour/issues/70#issuecomment-387575336).

kunickiaj:
If you're using contour for TLS termination x-forwarded-proto is needed in many cases -- couldn't find a way to set that.

kellyaa:
+1 on wanting to see this happen. Right now I have a need to customize the `Host` header for route when proxying a request.

davecheney:
[...]

PR #1012, opened 2019-04-11
design: add CORS support

aberasarte:
Updates #437

Signed-off-by: Aritz Berasarte <raskasso@hotmail.com>

### DIFF
--- a/design/cors-configuration.md
+++ b/design/cors-configuration.md
@@ -1,0 +1,180 @@
+# CORS configuration
+
+This document proposes a way to enable CORS (Cross-origin resource sharing) policies and therefore, allow access to resources in a cluster from a webpage deployed at a different domain.
+
+## Goals
+
+- To allow configuration of CORS related policies, letting Contour process cross domain requests and add all the needed response headers.
+
+## Background
+
+Nowadays, most modern browsers don't allow requests to a domain from a webpage that has been fetched from a different domain. This is a protection mechanism against CSRF (Cross-Site Request Forgery) attacks. To explain what a CSRF attack is, it is necessary to explain how authentication and user sessions have been handled traditionally by most web applications:
+- When a user tries to access a protected application (an online bank for instance), they are asked to enter their credentials on a login screen.
+- When the credentials are sent to the server, these are validated against a database, an external service or whatever.
+- The server generates a token which is stored in a cookie. This way, the browser will send the cookie with the token automatically in successive requests and the server will be able to validate that the user has already been authenticated without querying the database.
+
+Knowing how authentication works, a malicious webpage could try to force a user to execute unwanted actions on a web application in which they're currently authenticated using some social engineering. This could be an attack example:
+- A user who is authenticated on an online bank (`www.mybank.com`) receives an email saying that they won a brand new car. To get the prize, they only have to visit `www.malicious-site.com/winner` and fill out a form with some personal data.
+- When the user goes to the malicious web page, a Javascript code is executed in the background sending an AJAX request to `www.mybank.com/transfer` which is the URL used for doing money transfers on the online bank.
+- As the user is already authenticated on `mybank.com`, the authentication cookie is sent automatically.
+- The web application checks the cookie and as it's valid, the money transfer is performed on the victim's behalf.
+
+To avoid this kind of scenario, web browsers apply the same-origin policy.
+
+### Same-origin policy
+
+Under the same-origin policy, a web browser allows scripts contained on a web page to access data on another server, but only if both share the same origin. An origin is defined as a combination of URI scheme, host name, and port number. Thanks to same-origin policy, attacks like the one explained previously are prevented by the browser itself because `www.mybank.com` and `www.malicious-site.com` don't share the same origin.
+
+However, the way web applications are developed has evolved, and nowadays it’s very frequent to separate the frontend from the backend, deploying them independently. For instance, the frontend could be a Javascript single page application deployed on a CDN (`myfrontend.com`) and the backend, a microservices cluster deployed somewhere else (`mybackend.com`).
+
+As the Javascript application needs to send requests to the API exposed by the backend and they are hosted on different domains, the web browser will prevent any communication between them due to the same-origin policy. This is where CORS comes into play.
+
+### CORS
+
+CORS is a mechanism that allows bypassing the same-origin policy for trusted sources. This is the way it works in its simplest form:
+- Every time a cross-origin AJAX request is about to send, the browser sets the `Origin` request header with the web page's origin as the value.
+- The server checks the `Origin` header and if the origin is allowed, it sets the `Access-Control-Allow-Origin` header to the `Origin` value.
+- When the response reaches the browser, it verifies that the value of the `Access-Control-Allow-Origin` header matches the origin of the tab the request  originated from. If it doesn't match, it throws an error.
+
+The following are the criteria that define a simple request:
+- Requests only use the GET or POST HTTP methods. If the POST method is used, then Content-Type can only be one of the following: application/x-www-form-urlencoded, multipart/form-data, or text/plain.
+- Requests do not set custom headers, such as X-Other-Header.
+
+If the content of the request doesn't meet the criteria above, the browser first checks whether the actual request should be sent. This is done by sending a special request (called preflight request) to the server in advance. A preflight request first sends an HTTP request to the resource using the OPTIONS method with the following headers:
+- `Origin`: Specifies the domain that would like access to the resource. This is inserted by the browser in a cross-origin request.
+- `Access-Control-Request-Method`: The HTTP method to be used in the actual request from the browser.
+- `Access-Control-Request-Headers`: The custom headers to be sent in the actual cross-origin request.
+
+In response to a preflight request, the server sends a response with the following headers:
+- `Access-Control-Allow-Origin`: Specifies the domains allowed to access the resource.
+- `Access-Control-Allow-Credentials`: Indicates whether browser credentials can be used to make the actual request (cookies for instance).
+- `Access-Control-Expose-Headers`: Allows headers to be exposed to the browser.
+- `Access-Control-Max-Age`: Specifies how long preflight request results can be cached.
+- `Access-Control-Allow-Methods`: Indicates which methods are allowed when making an actual request.
+- `Access-Control-Allow-Headers`: Indicates which headers can be used in the actual request.
+
+When the browser gets the preflight response, it checks if the origin is allowed and if the HTTP method and headers of the main request are in the list returned by the server. If so, it sends the main request, which will be a regular cross-origin request, it will include the `Origin` header and the response will contain `Access-Control-Allow-Origin` once again.
+
+This proposal introduces a way to set all the CORS related configuration in Contour, letting Envoy do all the heavy work.
+
+## High-Level Design
+
+Envoy supports [CORS via a filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/cors_filter.html) and it can be configured [using the API](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-corspolicy). The changes proposed in this document will allow the configuration of CORS policies in Contour.
+
+At a high level the proposed changes will imply:
+- Adding new fields at virtual host level to configure the CORS policy in the YAML.
+- Changing some structs in the code.
+- Enabling the Envoy CORS filter.
+
+### Proposed YAML fields
+
+```yaml
+apiVersion: projectcontour.io/v1
+kind: HTTPProxy
+metadata:
+ name: google apis
+ namespace: prod
+spec:
+ virtualhost:
+   fqdn: www.googleapis.com
+   # The CORS policy described here will apply to all the routes of the virtual host.
+   corsPolicy:
+     # Which domains can access the matched resources in a cross-site manner.
+     allowOrigin:
+       - "*"
+     # Which HTTP methods are allowed for cross-origin requests (case-sensitive).
+     allowMethods:
+       - GET
+       - POST
+     # The headers the server is going to accept (case-insensitive).
+     allowHeaders:
+       - cache-control
+       - content-type
+       - custom-header
+     # The non simple headers the client will be able to access (case-insensitive).
+     exposeHeaders:
+       - Content-Length
+       - Content-Range
+     # Whether the server allows sending credentials (cookies for instance) in cross-origin requests.
+     allowCredentials: true
+     # the amount of time the preflight response will be cached. It's expresed in the Go duration format. If not supplied, browser default values will apply.
+     maxAge: 10m
+   routes:
+     - conditions:
+       - prefix: /analytics
+       services:
+         - name: analytics-api
+           port: 9999
+```
+
+The names and types of the proposed fields are inspired by the [HTTP headers](https://www.w3.org/TR/2010/WD-cors-20100727/#syntax). The `maxAge` is going to be parsed into a `time.Duration` as this is going to be the type used in the Dag. If no value is provided, [browser defaults](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age#Directives) will apply.
+
+## Detailed Design
+
+### Reading from YAML
+The virtual host struct will be updated to contain the CORS related information:
+
+```go
+// contour/apis/contour/v1beta1/projectcontour/v1/httpproxy.go
+// VirtualHost appears at most once. If it is present, the object is considered
+// to be a "root".
+type VirtualHost struct {
+  [... other members ...]
+
+  // Specifies the CORS policy to apply for the matched path.
+  // +optional
+  CorsPolicy *CorsPolicy `json:"corsPolicy,omitempty"`
+}
+// CorsPolicy allows setting de CORS policy
+type CorsPolicy struct {
+	// Specifies whether the resource allows credentials.
+	AllowCredentials bool `json:"allowCredentials"`
+	// AllowOrigin specifies the origins that will be allowed to do CORS requests.
+	AllowOrigin []string `json:"allowOrigin"`
+	// AllowMethods specifies the content for the *access-control-allow-methods* header.
+	AllowMethods []string `json:"allowMethods"`
+	// AllowHeaders specifies the content for the *access-control-allow-headers* header.
+	AllowHeaders []string `json:"allowHeaders"`
+	// ExposeHeaders Specifies the content for the *access-control-expose-headers* header.
+	ExposeHeaders []string `json:"exposeHeaders"`
+	// MaxAgeSeconds specifies the content for the *access-control-max-age* header.
+	MaxAge string `json:"maxAge"`
+}
+```
+
+The common DAG struct will be updated accordingly:
+
+```go
+// contour/internal/dag/dag.go
+// A VirtualHost represents a named L4/L7 service.
+type VirtualHost struct {
+	[... other members ...]
+
+	CorsPolicy *CorsPolicy
+}
+```
+
+### Enabling the CORS filter in Envoy
+In order to enable the CORS filter in Envoy we will update `contour/internal/envoy/route.go` and map the values from DAG's route to [protobuf](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route.proto.html?highlight=shadow#route-corspolicy).
+
+## Alternatives Considered
+
+### Delegate all the CORS related logic to the applications
+All the CORS related tasks could be handled in each application using an application level middleware or similar. However, it isn't something people usually want to do at this level for the following reasons: 
+
+- This kind of configuration doesn't usually rely on developers because they are mainly focused on developing the services (the what), and not on which domains those services are going to be deployed on (the where), nor which security rules should be applied. This is usually a task for the Ops/DevOps/DevSecOps teams and that's the reason why all the reverse proxies out there provide this functionality alongside SSL termination, gzip compression and so on.
+- In microservices driven architectures probably all the microservices will share the same CORS policy. It's much easier to manage this for multiple services if the policy isn't embedded in the application.
+
+
+### Enable CORS with a single flag and apply sane defaults
+It would be great if we could just enable a flag and apply some sane defaults. Unfortunately, it's not easy to find some defaults that would work for most of the proposed fields:
+
+- **allowOrigin:** We could avoid this field and use '*' as the default value. However, as this opens the route for any domain, people would probably want to be able to set more restrictive values.
+- **allowMethods:** We could allow all the methods (GET, PUT, POST, DELETE, PATCH, OPTIONS) and get rid of this field.
+- **allowHeaders:** This is a very application specific setting. For instance, if a service is going to be consumed by JQuery clients, you'll probably want to allow the `X-Requested-With` header while for grpc-web services, you'll want to allow headers like `x-grpc-web` or `grpc-timeout`. 
+- **exposeHeaders:** This is a very application specific setting as well. You might want to expose different headers depending on the application and the technologies used.
+- **allowCredentials:** Setting a default value for this field is tricky because it has [some implications](https://www.w3.org/TR/2014/REC-cors-20140116/#supports-credentials).
+- **MaxAge:** We could get rid of this field and set a default value. As a reference, Firefox specifies a 24 hours default value while Chromium sets it to 10 minutes.
+
+### Develop a generic response header setting facility and handle CORS headers with it
+This approach wouldn't be valid for managing all the CORS related logic. Some of the headers like `Access-Control-Allow-Headers` are only set in response to preflight requests (method=OPTIONS). In addition to this, we have to take into account that CORS is not just about headers, some server side logic is needed as well. For example, the server should return the `Access-Control-Allow-Origin` header with the `Origin` value sent by the browser only if that `Origin` is allowed.


davecheney:
Thank you for your PR @aberasarte, especially for your thought design document.

I have several areas of concern for accepting this feature. They are all basically the same; the surface area of the CORS configuration values is too large. To put this in context, this design adds 6 new fields under the `cors:` key. These map to eight filter configuration values exposed by Envoy. This one feature would add more API surface area to IngressRoute than the Timeout and Retry, and the sticky sessions designs combined. More over at first reading none of the CORS configuration values have a sensible default, all the fields will have to be present, and configured with correct (not just correctly typed) values, for CORS to work. This is the major concern that I have.

Contour is a commercially supported product, if you're an essential PKS customer you can ring up our CRE team and they are expected to be able to debug a problem using any feature we add to Contour. Implementing CORS according to this design means they have to become experts on the safe and correct values for each of those fields. This is a long way of me saying, I have to consider the long term supportability of every feature we add to Contour.

So, how can we decrease the support cost of adding this feature? My initial reaction would be, _can I have a `cors: true` field_ that enables CORS for that route? I'm expecting that the immediate response would be "no, because we need to configure the X value". This is a good response because it means we can perhaps make some of the fields in the proposed design constant. It also like to do a survey of the common values that CORS filters in Envoy, and perhaps nginx and other web servers -- the field names seem to be direct analogs of the W3C document and the browser headers. If these values are always set to the same thing, then we have a path to arguing that they are actually constants and we don't need to let people configure them. I made the same argument in #823 and it seems to be holding so far.

The second approach I would take is to ask why CORS needs to be its own thing, can it be de sugared into a generic response header setting facility, see https://github.com/heptio/contour/issues/70#issuecomment-482058133. What is so special about CORS headers that they need their own filter and configuration? Reading the [W3C document](https://www.w3.org/TR/2014/REC-cors-20140116/) there are a stack of wtfs in there. The server seem to be repeating back to the client what it just told the server; the `Origin:` request header vs the `Access-Control-Allow-Origin: ` response header. I don't expect either one of us to justify the W3C's decisions, but if Contour is going to be responsible for configuring Envoy to set these headers rather than the backend service doing it, then this puts us back into the area of supportability, with the extra drop of difficulty that we might have to release security patches if Contour has bugs in configuring Envoy to operate securely--and equally we have to do a security release if the bug is on the Envoy side.

The third approach is to push this back to the backend application to set these headers. Why can't the backend application manage its own CORS settings? What are the pieces of information that Envoy has as a proxy that the backend lacks that prevent it from setting CORS headers correctly? If there is information that the backend service lacks, maybe we can change Contour to provide that information. If setting CORS at the Envoy middle box layer is a convenience, then that has to be weighed against the long term support cost of this feature.

Thank you for your time spend on the design document and your investment in continuing the discussion. I wanted to be clear that I am not saying _no_, but I have to consider the wider costs of adding any feature to Contour, and this feature is a big one; it has a high API surface area, and a potential security risk when misconfigured.

aberasarte:
Thanks @davecheney for your feedback, I really appreciate it.

> Contour is a commercially supported product, if you're an essential PKS customer you can ring up our CRE team and they are expected to be able to debug a problem using any feature we add to Contour. Implementing CORS according to this design means they have to become experts on the safe and correct values for each of those fields. This is a long way of me saying, I have to consider the long term supportability of every feature we add to Contour.

This is a really good point. CORS could be seen as and advanced feature and I didn't take into account its implications for the support team. They would need to acquire some basic knowledge of the topic and be able to deal with CORS related issues so, I understand what you mean.

> So, how can we decrease the support cost of adding this feature? My initial reaction would be, can I have a cors: true field that enables CORS for that route? I'm expecting that the immediate response would be "no, because we need to configure the X value". This is a good response because it means we can perhaps make some of the fields in the proposed design constant. It also like to do a survey of the common values that CORS filters in Envoy, and perhaps nginx and other web servers -- the field names seem to be direct analogs of the W3C document and the browser headers. If these values are always set to the same thing, then we have a path to arguing that they are actually constants and we don't need to let people configure them.

It would be great if we could just enable a flag and apply some sane defaults. Unfortunately, I don't think it's feasible to find default values that would work for most of the proposed fields:

- **allowOrigin:** We could avoid this field and use '*' as the default value. However, as this opens the route for any domain, I think people should be able to set more restrictive values.
- **allowMethods:** I'd rather make people specify the allowed methods but I guess that we could just allow all of them (GET, PUT, POST, DELETE, PATCH, OPTIONS) and get rid of this field.
- **allowHeaders:** This is a very application specific setting. For instance, if your service is going to be consumed by JQuery clients, you'll probably want to allow the `X-Requested-With` header while for  grpc-web services, you'll want to allow headers like `x-grpc-web` or `grpc-timeout`. 
- **exposeHeaders:** This is a very application specific setting as well. You might want to expose different headers depending of the application and the technologies used.
- **allowCredentials:** I would encourage people to only enable this one if they really need it (for example, if they need to allow the client to send cookies). Setting a default value for this field is tricky because it has [some implications](https://www.w3.org/TR/2014/REC-cors-20140116/#supports-credentials).
- **MaxAge:** I guess that we could get rid of this field and set a default value. As a reference, Firefox specifies a 24 hours default value while Chromium sets it to 10 minutes.

> The second approach I would take is to ask why CORS needs to be its own thing, can it be de sugared into a generic response header setting facility, see #70 (comment). What is so special about CORS headers that they need their own filter and configuration? 

I'm afraid that this approach wouldn't be valid for managing all the CORS related logic. Some of the headers like `Access-Control-Allow-Headers` are only set in response to preflight requests (method=OPTIONS). In addition to this, we have to take into account that CORS is not just about headers, some server side logic is needed as well. For example, the server should return the `Access-Control-Allow-Origin` header with the `Origin` value sent by the browser only if that `Origin` is allowed.

> The third approach is to push this back to the backend application to set these headers. Why can't the backend application manage its own CORS settings? 

Yes, it could be managed in the backend applications. There are middlewares and filters available for most of the frameworks and languages allowing developers to set CORS policies at application level. As this is a valid approach, it isn't something you usually want to do for the following reasons:

- This kind of configuration doesn't usually rely on developers because they are mostly focused on developing the services and not on which domains those services are going to be deployed on, nor which security rules should be applied. In fact, you could want to apply different CORS policies depending on where the application is going to be deployed on. This is the reason why this kind of tasks are usually handled at proxy level and managed for the Ops/DevOps/DevSecOps teams.
- If you are deploying multiple microservices you'll probably want to apply the same CORS policy to all of them. It's much easier to manage this for multiple services when the policy isn't embedded in the application.

I've updated the design document with some of the points discussed here.

sylwit:
For unblocking this feature and having a more simple draft, I would only implement those 2 values
`allowOrigin` and `allowMethods`. It should be enough for a lot of people.

I think a good default value for `allowOrigin` would be to reuse the vhost domains instead of allowing all domains which is less secure.

I agree for the to use all verbs also
allowMethods: I'd rather make people specify the allowed methods but I guess that we could just allow all of them (GET, PUT, POST, DELETE, PATCH, OPTIONS) and get rid of this field.

Here is an example: https://github.com/envoyproxy/envoy/blob/master/examples/cors/backend/front-envoy.yaml#L19

aberasarte:
Hi @davecheney just wondering if you've had a chance to think about my feedback and if you have any further comments.

igolaizola:
Any update on this?

davecheney:
I’m sorry we won’t be looking at this for 0.13. 

I will push again for a decision to include or not include this in the 0.14 cycle starting in a few weeks.

davecheney:
I have asked our product management for a decision on CORS support. I expect to make the next update on this ticket in two weeks.

aberasarte:
Thanks for the status update.

davecheney:
Thank you for your patience. I have unlocked #437 and will place an update there.

annismckenzie:
This shouldn't have been closed? :/

jpeach:
Agreed, re-opening.

glerchundi:
Is it possible for a contour maintainer to give a timeline for this? Not having the possibility to configure CORS behaviour makes the proxy unusable for use cases involving multiple domains & API services which is a quite common scenario (SPA + API).

The PR was submitted more than a year ago and I think it could project a false sensation of how the maintainers & community are engaged in reality.

I would expect at least a yes with a timeline or a rotund no to avoid false expectations.

Keep up the fantastic job you're doing here and hope it finally gets into CNCF!

yoitsro:
@glerchundi It's probably not so useful for this PR, but I'm trying to keep your branch up to date with main here: https://github.com/desanaio/contour/tree/feature/cors

Surprisingly, there haven't been that many conflicts/issues!

We require this functionality so for now, this seems like the only way to go.

youngnick:
Hi @glerchundi, thanks for your interest in this PR. Currently addressing this isn't on [our roadmap](https://github.com/projectcontour/community/blob/master/ROADMAP.md), mainly because there's an unknown amount of work to finish off the design.

The design here does a great job in laying out how the idea would work for a single HTTPProxy, but one of the perennial problems we've had with adding things to HTTPProxy is around what happens when you have a lot of them, if and how you default the settings, and if the cluster administrator wants to allow the settings to be set by HTTPProxies.

The example here is when we added the `request_timeout` - we allowed HTTPProxy users to set the request timeout to infinite, and this produced problems for some cluster administrators, who did not want to allow that. We're working on addressing this one, as well as adding more timeouts under #2225, this is just to illustrate that when adding a new feature into HTTPProxy, it's important to consider the cluster administrator experience as well as the application owner one.

I can appreciate that not having CORS available is a problem  for application owners, absolutely, and I would like to have some way to configure CORS in Contour. However, what I want to understand before just giving the go-ahead here is:
- Is there any way we can cut the number of fields to configure? Every item we add into HTTPProxy adds to the testing matrix, is there a way to make the testing of this feature easier? Are there any ways that a malicious application developer could use these fields to do something unsavoury with another HTTPProxy running through the same Envoy? Do we need to have an enumerated list of options for any values or are strings okay?
- what are the ways in which the needs of the cluster administrator and application owner interact? Is there a case where a cluster adminstrator would need to limit what the permissible values could be (The obvious one from what I can see is that maybe a cluster administrator might want to disallow an allowed origin of `*`). Should we allow this feature in without having cluster administrator overrides in place? I can't currently answer this because I don't know enough about the problem domain.
- Should there be a way to default the settings across all HTTPProxies? Would a cluster admin want to do that? If so, how?

What all of these questions add up to is that there's a bunch of prep work to really be able to comment properly here. I've love for this feature to end up delivered in Contour, but there needs to be more work done in understanding the context around using HTTPProxy in a shared proxy.

To get down to the real stuff, this work currently isn't on the roadmap because no one has followed up before now. We have limited resources, and try our best to assign them where they'll do the most good. @yoitsro, if you're carrying a fork because of this feature, then yes, let's get it on the roadmap and see what we can do.

To get this issue on the roadmap, please come to one of the community meetings, detailed on our [community page](https://projectcontour.io/community/), and tell us about it. I'd love to be proved wrong about the amount of work here, and get this feature added and done.

design/cors-configuration.md
```diff
@@ -0,0 +1,188 @@
...
+### Enabling the CORS filter in Envoy
+
+The final step will be to enable the Envoy's CORS filter. This we'll be done in the `HTTPConnectionManager` function of `contour/internal/listener.go`.
+
+## Alternatives Considered
+
+### Delegate all the CORS related logic to the applications
+All the CORS related tasks could be handled in each application using an application level middleware or similar. However, it isn't something people usually want to do at this level for the following reasons: 
+
+- This kind of configuration doesn't usually rely on developers because they are mainly focused on developing the services (the what), and not on which domains those services are going to be deployed on (the where), nor which security rules should be applied. This is usually a task for the Ops/DevOps/DevSecOps teams and that's the reason why all the reverse proxies out there provide this functionality alongside SSL termination, gzip compression and so on.
```
jpeach, line 172: Perhaps this argues for specifying the CORS policy somewhere that is not a `Route`? The inclusion model for `HTTPProxy` is more or less that cluster or infrastructure operators set up the `HTTPProxy` and some sort of inclusion chain, then application developers configure the `Route` objects.

It seems like here you are arguing that CORS policy should be centralized rather than distributed.

Maybe the trade-off here is to centralize the policy, but allow it to be disabled on specific routes?

design/cors-configuration.md
```diff
@@ -0,0 +1,188 @@
...
+apiVersion: contour.heptio.com/v1beta1
+kind: IngressRoute
+metadata:
+ name: google
+ namespace: prod
+spec:
+ virtualhost:
+   fqdn: www.google.com
+   routes:
+   - match: /finance
```
jpeach, line 81: Does this structure still make sense for the `HTTPProxy`API?

design/cors-configuration.md
```diff
@@ -0,0 +1,188 @@
...
+    // Indicates the CORS policy to apply
+    CorsPolicy *ingressroutev1.CorsPolicy
+}
+```
+
+### Communicating from DAG to Envoy
+
+Once the DAG has been built, we need to send the CORS related configuration to Envoy. To do so, we will update the `RouteRoute` function of `contour/internal/envoy/route.go` and map the values from DAG's route to [protobuf](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route.proto.html?highlight=shadow#route-corspolicy).
+
+One important thing to take into account is that in Envoy 1.10 the [corsPolicy.enabled has been deprecated](https://www.envoyproxy.io/docs/envoy/v1.10.0/api-v2/api/v2/route/route.proto#envoy-api-field-route-corspolicy-enabled). As we are still using Envoy 1.9, we'll start using the deprecated field to enable CORS and change it once we upgrade to 1.10.
```
jpeach, line 161: I think that we can use the `filter_enabled` field now.

design/cors-configuration.md
```diff
@@ -0,0 +1,188 @@
...
+### Reading from YAML
+The ingressroute spec struct will be updated to contain the CORS related information:
+
+```go
+// contour/apis/contour/v1beta1/ingressroute.go
+// Route contains the set of routes for a virtual host
+type Route struct {
+   [... other members ...]
+   // Specifies the CORS policy to apply for the matched path.
+    CorsPolicy *CorsPolicy `json:"corsPolicy,omitempty"`
```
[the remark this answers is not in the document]
aberasarte: You are right, probably the most common use case would be to apply the same policy across a virtual host but I would love to get feedback from the community about this.

design/cors-configuration.md
aberasarte: > Maybe the trade-off here is to centralize the policy, but allow it to be disabled on specific routes?

Sounds good to me.

yoitsro:
Paging @glerchundi. Since this was originally your PR which I've just been rebasing, please could you help with some of these comments? 🤞

design/cors-configuration.md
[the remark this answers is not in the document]
yoitsro: Having used this fork, I can confirm that yes, it would be good to apply the same policy across a virtual host. It perhaps doesn't give as much flexibility, but it's also a complete pain duplicating the CORS setup for each route.

design/cors-configuration.md
youngnick: We've done similar things before (a centralized policy that can be disabled), I also like this pattern.

design/cors-configuration.md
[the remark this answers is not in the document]
youngnick: Would this then make the CorsPolicy (maybe `OriginPolicy` is a more generic name) a peer to the TLS policy on the virtualhost struct? If so, I like that, because it gives one place to set this information - the root HTTPProxy, with the caveat that there should be a way to disable on nonroots.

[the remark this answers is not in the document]
jpeach: I am not confident that I understand the use cases that would not be possible if the policy is moved to the virtual host. Would be great if someone could explain to me in small words :)

glerchundi:
First of all sorry for being late to the party and thanks for all your comments @youngnick & review @jpeach. And @yoitsro for taking me into account ❤️ 

I've talked to @aberasarte over Slack and these are our conclusions:
- The one who is driving this feature request starting with the design doc is @aberasarte as he made a HUGE effort of converting my initial and incomplete design into a reviewable document 👏 
- Except for a few parameters, there aren't sane defaults for CORS. And it's not good to do so as this is a bad practice even from a security standpoint, see Zoom's issue. Correct here if I'm wrong @aberasarte 
- The current proposal is outdated as it uses `IngressRoute` for the spec & the implementation is based on a Envoy API version where things have changed about how CORS is configured (it's runtime fraction based now). The idea would be to adapt it but @aberasarte wanted to agree on some changes before proceeding as it is gonna take a big amount of his time to do so.
- The delegation think wasn't thought when the design doc was created, I think the best way to proceed would be to also provide some guidance to @aberasarte before doing anything.
- Regarding to where should we put the `corsPolicy`, the vision of the maintainers as of the time the design was created was to put everything at the `route` level. But we agree with you @jpeach that it makes much more sense at a `virtualhost` level.

WDYT?

jpeach:
> First of all sorry for being late to the party and thanks for all your comments @youngnick & review @jpeach. And @yoitsro for taking me into account ❤️
> 
> I've talked to @aberasarte over Slack and these are our conclusions:
> 
> * The one who is driving this feature request starting with the design doc is @aberasarte as he made a HUGE effort of converting my initial and incomplete design into a reviewable document 👏

👍 

> * Except for a few parameters, there aren't sane defaults for CORS. And it's not good to do so as this is a bad practice even from a security standpoint, see Zoom's issue. Correct here if I'm wrong @aberasarte
> * The current proposal is outdated as it uses `IngressRoute` for the spec & the implementation is based on a Envoy API version where things have changed about how CORS is configured (it's runtime fraction based now). The idea would be to adapt it but @aberasarte wanted to agree on some changes before proceeding as it is gonna take a big amount of his time to do so.
> * The delegation think wasn't thought when the design doc was created, I think the best way to proceed would be to also provide some guidance to @aberasarte before doing anything.

I'm not sure what you mean by the "delegation thing". Maybe we need a separate discussion (on slack or a github issue) to create a consensus here?

> * Regarding to where should we put the `corsPolicy`, the vision of the maintainers as of the time the design was created was to put everything at the `route` level. But we agree with you @jpeach that it makes much more sense at a `virtualhost` level.
> 
> UPDATE: I've forgotten to say that we're also keeping our own fork just for the sake of having CORS implemented.
> 
> WDYT?

This plan sounds good to me. I'm happy to shepherd the work, but I'm also glad for @aberasarte and @glerchundi to be the primary owners of the feature since I don't have a strong sense of the practical impact of some of the choices that we will need to make. I'm available on slack to discuss, or ping me on any issues where you need feedback.

aberasarte:
I'm happy to update the design doc adapting it to the `HTTPProxy` API and moving the configuration to the virtual host. I would like to get the greenlight on the amount of fields we are going to add though. Mi proposal would look as follows:

* **allowOrigins:** The origins that will be allowed to do cross-domain requests .
* **allowHeaders:** The headers that will be allowed to be sent in the cross-domain requests.
* **exposeHeaders:** The returned headers that are going to be readable from Javascript code.
* **allowCredentials:** Whether browser credentials can be used to perform cross-domain requests (cookies for instance).

In the other hand, the rest of the CORS related fields wouldn't be configurable. We would set the following defaults instead:

* **AllowMethods:** `GET`, `PUT`, `POST`, `DELETE`, `PATCH`, `OPTIONS`.
* **MaxAge:** 24 hours.

WDYT?

jpeach:
> I'm happy to update the design doc adapting it to the `HTTPProxy` API and moving the configuration to the virtual host. I would like to get the greenlight on the amount of fields we are going to add though. Mi proposal would look as follows:
> 
> * **allowOrigins:** The origins that will be allowed to do cross-domain requests .
> * **allowHeaders:** The headers that will be allowed to be sent in the cross-domain requests.
> * **exposeHeaders:** The returned headers that are going to be readable from Javascript code.
> * **allowCredentials:** Whether browser credentials can be used to perform cross-domain requests (cookies for instance).
> 
> In the other hand, the rest of the CORS related fields wouldn't be configurable. We would set the following defaults instead:
> 
> * **AllowMethods:** `GET`, `PUT`, `POST`, `DELETE`, `PATCH`, `OPTIONS`.
> * **MaxAge:** 24 hours.
> 
> WDYT?

I think that we should bite the bullet and add all 6 fields.

We should think carefully about whether any can be defaulted so that they could commonly be omitted. Unfortunately, we can't use kubebuilder defaults annotations until we move to CRD v1, so that probably means defaulting field inside Contour and just documenting that in the API docs.

design/cors-configuration.md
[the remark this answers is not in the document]
aberasarte: > Would this then make the CorsPolicy (maybe `OriginPolicy` is a more generic name) a peer to the TLS policy on the virtualhost struct?

Even though I agree that `OriginPolicy` is a more generic name, I think that it could confuse a lot of people that would be looking for the `Cors` term. Here is a chart from Google trends comparing the popularity of "Same Origin Policy" and "CORS".

Captura de pantalla 2020-07-22 a las 18 29 23


> If so, I like that, because it gives one place to set this information - the root HTTPProxy, with the caveat that there should be a way to disable on nonroots.

We could add a `disableCors` property at route level. WDYT?

design/cors-configuration.md
```diff
@@ -0,0 +1,188 @@
...
+}
+
+// CorsPolicy allows for setting de CORS policy
+type CorsPolicy struct {
+    // Specifies whether the resource allows credentials.
+    AllowCredentials bool `json:"allowCredentials"`
+    // AllowOrigin specifies the origins that will be allowed to do CORS requests.
+    AllowOrigin []string `json:"allowOrigin"`
```
aberasarte, line 130: `allow_origin` [has been deprecated](https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.12.0.html?highlight=allow_origin#deprecated) in favor of `allow_origin_string_match` which is a [StringMatcher](https://github.com/envoyproxy/envoy/blob/549164c42cae84b59154ca4c36009e408aa10b52/api/envoy/type/matcher/string.proto#L20). I'm not sure about the type that would suit better here, the existing [MatchCondition](https://github.com/projectcontour/contour/blob/master/internal/dag/dag.go#L54) is somehow related...

design/cors-configuration.md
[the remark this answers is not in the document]
youngnick: Adding a `disableCors` at route level is a plausible fix, year, but then we also need to consider if cluster admins may want to not allow that (basically, do we need a `disableDisableCors` setting at the root). Welcome to the administrative override design space!

I don't know enough about how this is usually used to know how important being able to disable CORS on some routes is versus the cluster admin's ability to ensure it's enforced everywhere.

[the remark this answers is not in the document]
aberasarte: > I am not confident that I understand the use cases that would not be possible if the policy is moved to the virtual host. Would be great if someone could explain to me in small words :)

The only case I could think of is the following one:  you have a public facing API and therefore, you want to give access to everyone that wants to use it. For instance,  if your API is deployed at `https://www.googleapis.com`, you need to set a cors policy where the allowed origins are `*` (because any domain could make requests to it). In this case, setting the policy at host level would be enough. But what if for whatever reason you want to block cross origin requests for a specific path like `/internal-api`? In this case, you could keep the aforementioned cors policy at host level but disable it for the `/internal-api` route. 

IMHO, the route level cors policies are edge cases and we could avoid them initially.

design/cors-configuration.md
aberasarte: > I'm not sure about the type that would suit better here

I'd rather avoid to mimic Envoy's `StringMatcher` in Contour. I think it's too complicated and we would add too many parameters (exact, prefix, sufix, regex...). IMO defining our type as a String and translating it into an `exact` StringMatcher  would be good enough. I think this approach is reasonable because `*` [is always treated as a wildcard](https://github.com/envoyproxy/envoy/blob/d6e2fd0185ca620745479da2c43c0564eeaf35c5/source/extensions/filters/http/cors/cors_filter.cc#L142) in Envoy even for `exact` matchers. What does all this mean?

The following configuration would allow any origin:

```
virtualhost:
   fqdn: company.com
   corsPolicy:
     allowOrigin:
       - "*"

```

But if I wanted to allow any subdomain, I couldn't set that this way:

```
virtualhost:
   fqdn: company.com
   corsPolicy:
     allowOrigin:
       - "*.company.com"

```

I would need to set all the allowed subdomains explicitly instead:

```
virtualhost:
   fqdn: company.com
   corsPolicy:
     allowOrigin:
       - "foo.company.com"
       - "bar.company.com"
...
```

youngnick:
Hi everyone, I've been a bit absent recently, can I get an update on where this PR is at? It seems like we're waiting on it being updated to use `HTTPProxy`? @aberasarte, will you be able to update it?

Also, there are a couple of outstanding queries - I think that the ability to disable CORS for routes is vital, but there's an interaction with the administrative requirement to be able to restrict those changes. For similar problems, @skriss is investigating admission control; we may be able to punt the administrative limits until after we make some decisions there.

@glerchundi, I know you're carrying a fork with CORS, and would love for you not to have to do that. I think that we're getting to a better place with `*Policy` structs in HTTPProxy, and this work should be pretty implementable once the design is done.

glerchundi:
> Hi everyone, I've been a bit absent recently, can I get an update on where this PR is at? It seems like we're waiting on it being updated to use HTTPProxy? @aberasarte, will you be able to update it?

I talked to @aberasarte and he'll update it, but he is on vacation and  wouldn't be able to get to it until September!

> Also, there are a couple of outstanding queries - I think that the ability to disable CORS for routes is vital, but there's an interaction with the administrative requirement to be able to restrict those changes. For similar problems, @skriss is investigating admission control; we may be able to punt the administrative limits until after we make some decisions there.

Would it be possible to clarify why is vital? IMHO i didn't consider it blocking as it can be addressed in a follow-up revision of this design proposal. Perhaps you could extend the comment with a use case you consider essential.

> @glerchundi, I know you're carrying a fork with CORS, and would love for you not to have to do that.

Yep!

> I think that we're getting to a better place with *Policy structs in HTTPProxy, and this work should be pretty implementable once the design is done.

I think the `*Policy` pattern you all designed is a fantastic approach. Currently we've a fork rebased on top of v1.7.0 branch that follows this approach but at the `route` level because that was the initial design. Now that everyone is more or less on the same page about feeling that CORS should be applied at the virtualhost level, we're a little bit closer to get it into `main` ;) finally included!

youngnick:
>Would it be possible to clarify why is vital? IMHO i didn't consider it blocking as it can be addressed in a follow-up revision of this design proposal. Perhaps you could extend the comment with a use case you consider essential.
Ah, sorry, I wasn't clear. I meant "vital to have eventually", on the assumption that *someone* will want to turn it off - that's been my experience for every feature we've added so far. :) I agree that it's not blocking for this proposal as it stands.

I'm following up on PRs, because we're looking at creating a new `main` branch to be the mainline for the repo (#2744) - which will make handling this tricky. I was hoping to get this in sooner than that - the timeline there is by the end of August. I don't think it's a huge deal, because we can either cherry pick the commit from `master` after we switch, or close this PR and redo against `main`. Probably the former is better as it keeps the discussion history more discoverable.

glerchundi:
> Ah, sorry, I wasn't clear. I meant "vital to have eventually", on the assumption that someone will want to turn it off - that's been my experience for every feature we've added so far. :) I agree that it's not blocking for this proposal as it stands.

Perfect! I'm with you then. Let's see what happens with the first version and we can invest time getting to that eventually. Thanks for clarifying.

> I'm following up on PRs, because we're looking at creating a new main branch to be the mainline for the repo (#2744) - which will make handling this tricky. I was hoping to get this in sooner than that - the timeline there is by the end of August. I don't think it's a huge deal, because we can either cherry pick the commit from master after we switch, or close this PR and redo against main. Probably the former is better as it keeps the discussion history more discoverable.

Ok. No worries. Creating the design doc is much more costly than moving PRs/branches around. As the time @aberasarte will get back to this again seems to be conflicting with your `main` timeline, don't consider this blocking.

design/cors-configuration.md
aberasarte: We'll start setting the policy at `HTTPProxy` level and see how things go. We could add a property at route level to disable the policy in following PRs.

aberasarte:
Thanks everyone for your review and for making this long awaited feature possible.

I've updated the design document. I think that all the comments have been addressed, PTAL.

I would like to see this merged before starting working in the implementation.

glerchundi:
Thanks for taking the time to update the design doc. @aberasarte even being on vacation!!!

LGTM

design/cors-configuration.md
```diff
@@ -0,0 +1,179 @@
...
+```yaml
+apiVersion: projectcontour.io/v1
+kind: HTTPProxy
+metadata:
+ name: google
+ namespace: prod
+spec:
+ virtualhost:
+   fqdn: www.google.com
+   # The CORS policy described here is local to match and apply for "/finance" only.
```
youngnick, line 80: With the way that HTTPProxy works, this implies that all Routes under this virtualhost will have the same CORS policy.

I'm not familiar enough with how CORS is used for real to see if that's bad. I suspect it may lead to unintended consequences maybe?

Sorry to bikeshed more on this one.

aberasarte: I updated the example, hopefully it's going to be easyer now to understand the way CORS policy will be applied.

The key here is that usually the APIs are exposed in their own domain. If we take a look at the way [Google publishes its APIs](https://developers.google.com/apis-explorer) we see that they use the domain https://www.googleapis.com:

- https://www.googleapis.com/analytics/ for analytics related APIs
- https://www.googleapis.com/blogs/ for blogger related APIs
...

All the APIs at https://www.googleapis.com need the same CORS Policy, so having the ability to set the policy at virtual host level comes in handy. IMHO it would be weird that someone needed to set a different policy for a certain route under the same domain. If that happened, a workaround could be to use a subdomain for that API, something like https://internal.googleapis.com (this is not a real API). For some reason, Google uses subdomains for some of their APIs (see https://bigquery.googleapis.com for instance) but I don't think they do this because of CORS.

For that reason, I think that is pretty safe to start with the HTTPProxy approach and see how the policy is being used. If we see that disabling the policy at route level is necessary we could add it afterwards. 

Just in case, @yoitsro, @glerchundi, do you know any real world example where different paths need different CORS policies for the same domain?

yoitsro: In our experience, CORS at the virtual host level is completely fine. As you say, if someone does need to have a different policy at a route level, that'd be out of the ordinary.

glerchundi:
Same for us, virtualhost works perfectly as we have an SPA hosted in a
specific domain (app.company.com) which is different from the API domain (
api.company.com).

I also agree with you two that we can start as it is designed in the doc.
right now and do whatever we want afterwards (cors in route level or just
allow disabling it at route level as well).

Hope it helps Nick

glerchundi:
Thanks everyone for getting this happen, specially to @aberasarte for creating the design doc. and @jpeach + @youngnick for investing the time to review it.

Issue #432, opened 2018-06-10
Provide support for integrating with an external auth service.


larkinkevin:
- Ambassador supports authentication via the "AuthService" manifest (and a user-provided external auth service that satisfies a very simple API).
- It appears that Envoy likes this approach, because they're planning to incorporate it:  https://github.com/envoyproxy/envoy/issues/2828.
- If/when Envoy provides this functionality, it would be great if Contour could expose it to users.

ihrwein:
I wanted to raise a similar issue then managed to convince myself that although this feature would be useful it's not aligned with this project's goals (note that I'm just somebody from the Internet who found this project some days ago.).

My reasoning is that different teams can have different authentication requirements. Contour is to have one Ingress that delegates the traffic to team-managed services. Those services can be simple services or API gateways that hides other services and those API gateways would be a better place for authentication. Just my two cents.

larkinkevin:
I can certainly see where you're coming from @ihrwein.  I just opened this issue to track the feature request (I figured I should at least make the ask).

Will totally understand if the Contour folks decide this is not a feature they want to support (because it's not handled by the Ingress spec).

rosskukulinski:
Hi @ihrwein and @larkinkevin - thanks for opening this issue.  I think this is an interesting feature request and something that could be a useful addition to the new IngressRoute CRD we're heads down on at the moment.

We haven't taken a close look at the APIs exposed via https://github.com/envoyproxy/envoy/issues/2828, but once it lands in an Envoy release, we'll take another peak.

mattmoyer:
+1 for this feature. I'd love to have some Contour backends protected by the `ext_authz` filter in the style of an API gateway.

Envoy docs for the feature, since it doesn't look like those have been linked here yet:
https://www.envoyproxy.io/docs/envoy/latest/configuration/http_filters/ext_authz_filter.html

pims:
I’d be happy to contribute a PoC for ext_authz filter configuration if the contour team still believes it's a valuable feature to add.

Are you dead set on having this only available through the new IngressRoute custom CRD or should I also consider how to integrate it with the native ingress definition?

rosskukulinski:
I think this may also be related to adding similar functionality as to https://github.com/heptio/contour/issues/68

davecheney:
Removing the 0.11 milestone as external auth support is not available in a shipping envoy 1.9

pims:
[...]

PR #1014, opened 2019-04-11
design: Adds external-auth support

simonswine:
This document outlines a specification to allow Contour to make use of the [external authentication] http_filter of Envoy to implement authentication.

This is an early version of a design draft. I have created an ugly PoC for it and this represents my learnings and my suggested approach. Happy for any feedback!

The feature is described in #432 

[external authentication]:(https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/ext_authz_filter)

### DIFF
--- a/design/external-authentication.md
+++ b/design/external-authentication.md
@@ -1,0 +1,216 @@
+# External Authentication
+
+_Status_: Draft
+
+This document outlines a specification to allow Contour to make use of the
+[external authentication] http_filter of Envoy to implement authentication.
+
+## Goals
+
+- Allow usage of Basic Auth on an IngressRoute using a password database in
+  `htpasswd` format stored in a secret
+
+- Allow usage of External Authentication API similar to the
+  [nginx-ingress external auth] configuration
+
+- Re-usability of the authentication configurations across multiple
+  IngressRoutes in a Namespace
+
+## Non-goals
+
+- Delegations of authentication configurations like TLSCertiticateDelegation
+  provides
+
+- Handling stateful authentication flows
+
+- Handling authentication of non HTTP traffic
+
+- Handling authentication on Kubernetes Ingress resources
+
+
+## Background
+
+Envoy 1.7.0 introduced an [external authentication] http_filter which can be
+configured via a GRPC API. Users request to hand-off authentication from their
+applications into the ingress controller. Nginx-Ingress provides similar
+features using annotations.
+
+## High-Level Design
+
+* IngressRoutes can reference an new Authentication CRD on a per route basis
+
+* Authentication CRDs define the method of authentication. Initially two
+  methods: Basic and RequestService are supported
+
+* Envoy's ingress listeners should configure Contour as authentication service
+  and Contour should implement the GRPC API call [CheckRequest]
+
+## Detailed Design
+
+### Changes to the CRDs
+
+To select an authentication method we can configure a reference on
+IngressRoutes like that:
+
+```
+apiVersion: contour.heptio.com/v1beta1
+kind: IngressRoute
+[...]
+spec:
+  routes:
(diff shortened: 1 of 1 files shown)


alexbrand:
Thanks for taking the time to write this up @simonswine!

I am very curious as to what others think, but in my mind, Contour does not seem like the right place to implement the authentication/authorization server. Contour is an ingress controller, and adding the auth concerns seems to muddy its responsibilities and security posture.

With that said, from a user's perspective, it might be painful to have to implement an external authN/Z server if I wanted to use this. I wonder if the Open Policy Agent would make sense here? It looks like there is an (improperly named) OPA adapter/plugin that implements the Envoy Ext Auth API: https://github.com/open-policy-agent/opa-istio-plugin

Something else that is a bit unclear to me is: what are we trying to solve? The proposal talks about Basic Auth and OAuth, but #432 seems to be about exposing the right knobs so that users can bring their own AuthN/Z server. I think it might be beneficial to discuss auth use cases to try to come up with a solution.

glerchundi:
Glad to see someone pushing this forward!

I’m with @alexbrand here, implementing the authentication logic in Contour
seems like to much IMHO.

I would vote for a one, and only, way to represent an external auth that
directly implements the CheckRequest API:

The good parts:
* Less logic
* Less code to maintain
* Separation of concerns
* Very flexible for the end-user. They are in the position to decide what
kind of auth methods are going to be implemented/deployed.
* Less API surface, easier to give support

The bad parts:
* Tied to Envoy, probably not that easy to make Contour vendor agnostic in
case there is an idea to make it compatible with other proxies like nginx
* More components involved so no that easy to deploy for simple use cases.

Thats it, thanks again for working on this!

stevesloka:
I kind of like having the separation with the `Authentication` CRD, it allows us to specify how something could auth centrally, then apply that to one or more routes. 

I'm wondering though if we should apply auth at a Root IngressRoute level, where all routes within a single `fqdn` should use the same auth or if that's too broad of a statement. By doing this we could have many Root IngressRoutes all reference the same `Authentication` OIDC provider. If that provider changed we'd update just a single CRD which would roll across any other applications.

glerchundi:
Why not directly allow configuring the [`CheckRequest`](https://github.com/envoyproxy/data-plane-api/blob/1935b52f94f7889ad9f538a17250e78cffd0af27/envoy/service/auth/v2/external_auth.proto#L25-L29) API instead of creating an abstraction? Really easy to sketch on the current `IngressRoute` and less API surface to be supported.

Once you make this configurable through the `IngressRoute` this allows other kind of deployments (separation of concerns for the win):
* Envoy/Contour requesting authorization to a external gRPC service.
* Envoy/Contour + (New CRDs those `Authentication` ones + Envoy/Contour implementing the authorization API): In this case Envoy would ask to Contour (which implements the ext_authz API) if the request is authorized or not.

Also, it seems like managed providers as well as other open source projects are embracing this 'Universal' API, for example: Google with its Traffic Director which implements `xDS v2`.

WDYT?

stevesloka:
Yup that's what I was suggesting @glerchundi, I see a way to configure the external gRPC service. I did something similar with my work on RateLimiting if you look at that PR. I showed how to use the Lyft implementation, but users could still integrate their own if they wanted.

glerchundi:
Whoop!

Fantastic, i misunderstood what you were suggesting. We’re on the same path
then.

Thanks!

glerchundi:
@simonswine do you plan to make changes to this PR in the short term? I'm interested in making some progress and I've some time in the next two months. So in case you feel I can take this one let me know.

michmike:
this PR is now stale. removing from the board so that when we prioritize #432, we will consider all options (including this design)

PR #1017, opened 2019-04-12
deployment: explicitly set the listening ports for doc purposes

### LINKED ISSUE #1003: Document the flags that can be set on the contour binary
Contour exposes a set of flags that can be used to configure the behavior of the ingress controller. Historically, I have resorted to looking at the source code under `cmd/` to find the flags I need to tweak. 

Ideally, these flags are documented for users that might not understand Go.

alexbrand:
There have been multiple cases in which users ask how to configure
Envoy's HTTP/HTTPS listening ports. By explicitly setting the
configuration flags in the deployment samples, users should have a
better sense of the flags that are available.

Fixes #1003

### DIFF
--- a/deployment/deployment-grpc-v2/02-contour.yaml
+++ b/deployment/deployment-grpc-v2/02-contour.yaml
@@ -25,7 +25,7 @@ spec:
         imagePullPolicy: Always
         name: contour
         command: ["contour"]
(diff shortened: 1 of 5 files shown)


vaamarnath:
LGTM

PR #1022, opened 2019-04-15
internal/contour: holdoff send if pending changes

mattalberts:
The holdoff notifier ensures synchronization between a forced and
delayed update, but does not indicate a need to flush.  As a result, if
a delayed update is blocked by a forced update, it will send resend
after the forced update releases the lock.  This change introduces a
pending change counter and a max pending threshold to condition
the update.

Updates #499

### DIFF
--- a/internal/contour/holdoff.go
+++ b/internal/contour/holdoff.go
@@ -27,8 +27,9 @@ import (
 )
 
 const (
-	holdoffDelay    = 100 * time.Millisecond
-	holdoffMaxDelay = 500 * time.Millisecond
+	holdoffDelay      = 100 * time.Millisecond
+	holdoffMaxDelay   = 500 * time.Millisecond
+	holdoffMaxPending = 500
 )
 
 // A HoldoffNotifier delays calls to OnChange in the hope of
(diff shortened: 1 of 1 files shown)


mattalberts:
> This PR will conflict #1021

davecheney:
Thank you for this PR. Can you please explain the scenario this PR is preventing. Thank you.

davecheney:
Thank you for your PR. I'm sorry I did not understand the PR description. Can you please explain the scenario you are preventing.

mattalberts:
That's because I (clearly) didn't grammar well yesterday. I'm sorry.  In light of the description that you just left on #499, I think you have a pretty good handle on what's causing the memory growth. (happy to cancel this PR, the window of opportunity is so small)

While looking for ways to reduce the outbound gRPC traffic, I walked the code back up to the `HoldoffNotifier` looking for possible races that would break the "only send on change" clause of the xDS protocol description. If I read this correctly:

> `timer.Stop` - if t.Stop returns false, then the timer has already expired and the function f has been started in its own goroutine; Stop does not wait for f to complete before returning

This was my working assumption:
- `func (hn *HoldoffNotifier) OnChange(builder *dag.Builder)`
  - acquires `hn.mu.Lock()`
  - calls `timer.Stop`, returns false because f was just called while we acquired the lock
- `time.AfterFunc(holdoffDelay, func() {})`
  - blocked on `hn.mu`
- `func (hn *HoldoffNotifier) OnChange(builder *dag.Builder)`
  - `hn.Notifier.OnChange(builder)`
  - releases `hn.mu.Unlock()`
- `time.AfterFunc(holdoffDelay, func() {})`
  - acquires `hn.mu.Lock()`
  - `hn.Notifier.OnChange(builder)` <- nothing has changed
 
Assuming nothing has changed between releasing the lock and acquiring the lock, `time.AfterFunc(holdoffDelay, func() {})` will trigger send of the same data.  Even as I describe it, it feels insignificant. Though it was nice having the change count in the logs.

davecheney:
@mattalberts i'm still processing your previous comment. 

On thing that occured to me is the pending atomic counter is useful in itself. How would you feel about pulling that out into its own PR?

mattalberts:
I can do that!

davecheney:
@mattalberts ping, would you please rebase this to master and we can talk about the rest of the change.

mattalberts:
I absolutely can!  Sorry, I missed the message (blame easter)

stevesloka:
@davecheney this is out of date again, curious what the next steps are based upon your last comment.

mattalberts:
I can rebase it again. Though, i think this is of limited value.  The sort order of the LDS entries is likely more import to reducing the overall impact of sending an update (i say decline or close :P)

davecheney:
@mattalberts to close this out, In 0.14 I rewrote the holdoff handler to superceed this change.

PR #1027, opened 2019-04-17
internal/contour: track pending changes in holdoff notifier

mattalberts:
Adds an atomic counter tracking the number of changes compacted by the
holdoff notifier.

Updates #499
Updates #1022

### DIFF
--- a/internal/contour/holdoff.go
+++ b/internal/contour/holdoff.go
@@ -18,6 +18,7 @@ package contour
 
 import (
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/heptio/contour/internal/dag"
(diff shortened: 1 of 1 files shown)


internal/contour/holdoff.go
```diff
@@ -63,9 +66,22 @@ func (hn *HoldoffNotifier) OnChange(builder *dag.Builder) {
...
-		hn.WithField("last update", time.Since(hn.last)).Info("performing delayed update")
+		hn.WithField("last_update", time.Since(hn.last)).WithField("pending", hn.pending.reset()).Info("performing delayed update")
 		hn.Notifier.OnChange(builder)
 		hn.last = time.Now()
 		hn.Metrics.SetDAGRebuiltMetric(hn.last.Unix())
 	})
 }
+
+// counter holds an atomically incrementing counter.
+type counter uint64
```
mattalberts: I considered extracting this into its own package (either `internal/atomic` or `internal/safe` to allow for reuse by `xds.go`.  Something like `pending safe.Uint64`.  I'm happy to do that with this PR or as a secondary one to collapse similar use in other areas.

davecheney: Let's follow Atwood's rule of three here and wait for two more use cases before refactoring.

PR #1040, opened 2019-04-23
design: add a strawman design for backend TLS verification

davecheney:
Updates #813

This PR adds a draft design for validation of backend services accessed
over TLS.

### DIFF
--- a/design/tls-backend-verification.md
+++ b/design/tls-backend-verification.md
@@ -1,0 +1,166 @@
+# TLS backend verification
+
+Status: Draft
+
+Contour 0.11 added the ability to communicate via TLS between Envoy and pods in the backend service.
+This proposal describes the facility for Envoy to verify the backend service's certificate.
+
+## Goals
+
+- Allow IngressRoute authors to assert that the backend service Envoy is communicating with is using a certificate signed by a certificate authority of their choice.
+- Allow IngressRoute authors to assert that the backend service Envoy is communicating with presents a subjectAltName of their choice.
+
+## Non Goals
+
+- Authentication via client side certificates is not in scope.
+- Support for k8s ingress documents.
+
+## Background
+
+Contour 0.11 added the `contour.heptio.com/upstream-protocol.tls` Service annotation which told Contour to use TLS when communicating with the members of the backend cluster, the Service's pods.
+However, this connection is not validated, Envoy will accept any certificate the client presents.
+
+This proposal seeks to improve the validation of TLS connections .
+
+## High-Level Design
+
+At a high level I propose the following:
+
+1. The CA to be used to validate the certificate presented by the backend should be packaged in a Secret.
+The secret will be placed in the namespace of the IngressRoute.
+
+2. IngressRoute's `spec.routes.services.service` entry will grow a new subkey, `validation`.
+
+3. The `validation` key will have a mandatory `caSecret` key which is the name of the secret where the ca to be validated is stored.
+Certificate Delegation is not in scope.
+
+4. The `validation` key will have an optional `subjectname` key which is expected to be present in the subjectAltName of the presented certificate.
+If `subjectname` is not present, any certificate with a valid chain to the supplied CA is considered valid.
+
+5. If `spec.routes.services[].validation` is present, `spec.routes.services[].{name,port}` must point to a service with a matching `contour.heptio.com/upstream-protocol.tls` Service annotation.
+If the Service annotation is not present or incorrect, the route is rejected with an appropriate status message.
+
+### Sample YAML
+
+```
+apiVersion: contour.heptio.com/v1beta1
+kind: IngressRoute
+metadata:
+  name: secure-backend
+spec:
+  virtualhost: 
+    fqdn: www.example.com  
+  routes:
+    - match: /
+      services:
+        - name: service
+          port: 8443
+          validation:
+            caSecret: my-certificate-authority
+            subjectname: backend.example.com 
+```
+
+## Detailed Design
+
+Implementation will consist of these steps:
+
+### CAs in Secrets
+
+The store of CA information is an opaque kubernetes secret.
+The secret will be stored in the same namespace as the corresponding IngressRoute.
+TLS certificate delegation is not in scope for this proposal.
+
+The secret object should contain one entry named `ca.key`, the constents will be the CA public key material.
+
+Example:
+```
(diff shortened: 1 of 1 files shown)


stevesloka:
@davecheney the proposal looks good to me. The only downside (that was discussed in the issue) is the duplication of reconfiguring the same caSecret on each service. The scenario might be I'm using the same CA across multiple services, but I've still got to re-duplicate that logic everywhere.

design/tls-backend-verification.md
```diff
@@ -0,0 +1,166 @@
...
+
+Implementation will consist of these steps:
+
+### CAs in Secrets
+
+The store of CA information is an opaque kubernetes secret.
+The secret will be stored in the same namespace as the corresponding IngressRoute.
+TLS certificate delegation is not in scope for this proposal.
+
+The secret object should contain one entry named `ca.key`, the constents will be the CA public key material.
```
robbiemcmichael: This should probably be named `ca.crt`. It does contain a public key, but it also contains other information required for certificate validation (the expiration date, a list of supported extensions, reference to a certificate revocation list, etc).

design/tls-backend-verification.md
```diff
@@ -0,0 +1,166 @@
...
+Envoy, acting as the client, is configured to talk to a set of IP addresses obtained from k8s Endpoint objects--there are no hostnames in play inside a cluster--representing pods for that service.
+Envoy is configured by Contour to perform a TLS handshake when connecting to a pod.
+This design will add the following constraints to Envoy's TLS handshake.
+
+a. That the certificate presented is signed by the CA who's public key we present to Envoy via the IngressRoute `spec.routes[].backends[]verification.caSecret` key.
+
+b. That the certificate presented contains a subjectAltName that is present in the IngressRoute `spec.routes[].backends[].verification.subjectName` key.
+
+The problem is that both of these checks exist _inside the security boundary of the application which we are attempting to validate_.
+Because the certificate is almost certainly going to be signed by a company CA, not one of the public CAs, checking the CA presented by the server matches the material supplied to the client, Envoy, is little more than a shared secret.
```
robbiemcmichael: This might be true when you generate your own CA and use that to sign the certificate, but it's not true for a large enterprise PKI which is really not that dissimilar from a public CA (the private CA is managed by another team, a concept of needing to have ownership of the subject names to get a certificate signed by the CA, even needing to pay for internal certificates).

design/tls-backend-verification.md
```diff
@@ -0,0 +1,166 @@
...
+It does not assert that the certificate _is_ signed by a trusted CA, only that the certificate _is_ signed by a key to which the client has a matching public component.
+This is further weakened by the observation that the service which we are connecting too, the ingressroute, and the caSecret are in the _same namespace_.
```
robbiemcmichael: I don't know if this is a design choice of Contour, but there's no technical reason ingressroutes couldn't be changed to refer to services in other namespaces.

design/tls-backend-verification.md
```diff
@@ -0,0 +1,166 @@
...
+It is likely that the secret that the service we are connecting too is also in the same namespace.
+
+SubjectAltName verification is similarly diluted.
+There is no compensating control that the client will use the public DNS infrastructure to resolve the IP address to connect too as in the earlier example.
+Instead Envoy will connect to a set of IP addresses controlled by the namespace--Ingressroute and endpoint documents live in the same namespace--and ask the endpoint that answers what it's name is.
+SubjectAltName becomes just a shared secret, albeit weaker that the CA verification, because its just string matching, there's no crypto in there.
+
+To be clear, verification does offer some improvements over simply doing a TLS handshake, but I cannot convince myself that it is as secure as the public TLS infrastructure.
+The same CA, certificate, and subject name parameters can be reused across multiple clusters with exactly the same resultant validation as someone who went to the effort to issue a certificate per service.
```
robbiemcmichael: It's not really any less secure than public CA (as long as your private CA is managed properly), it's just not as obviously necessary as you have greater trust in your own private network. The two major benefits in this case are:
- Ensures that traffic doesn't get routed to the wrong pod even if there is a bug in Envoy
- Is the only way to be confident that your connection isn't subject to a MITM attack

design/tls-backend-verification.md
```diff
@@ -0,0 +1,166 @@
...
+However, this connection is not validated, Envoy will accept any certificate the client presents.
+
+This proposal seeks to improve the validation of TLS connections .
+
+## High-Level Design
+
+At a high level I propose the following:
+
+1. The CA to be used to validate the certificate presented by the backend should be packaged in a Secret.
+The secret will be placed in the namespace of the IngressRoute.
```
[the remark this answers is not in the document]
stevesloka: Is there anything a configmap does better than a secret for this use-case?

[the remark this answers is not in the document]
robbiemcmichael: They don't do anything that secrets don't do, but you could use the same logic to say that Kubernetes shouldn't support configmaps at all.

It looks like the reasons for implementing it with secrets are about discouraging abuse of configmaps for other use cases. It's your code base so I don't have any strong opinions on that front, just wanted to make a note that CA certificates generally shouldn't be considered private material.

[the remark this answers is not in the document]
stevesloka: Gotcha, thanks for clarifying @robbiemcmichael. I appreciate the review on this! =)

design/tls-backend-verification.md
```diff
@@ -0,0 +1,166 @@
...
+  virtualhost: 
+    fqdn: www.example.com  
+  routes:
+    - match: /
+      services:
+        - name: service
+          port: 8443
+          validation:
+            caSecret: my-certificate-authority
+            subjectname: backend.example.com
```
stevesloka: To be consistent, should name this: `subjectName`.

design/tls-backend-verification.md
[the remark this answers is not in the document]
davecheney: I agree that a CA is not private material. The justification for using a secret is contour is already watching all secrets.

Storing this information in a config map means contour needs to watch all config maps, adding load to the API server. Additionally I'm very scared that this will create a precedent for storing arbitrary contour configuration in one or more config maps. I do not want to do this unless absolutely necessary and only after rigourous investigation of other options.

design/tls-backend-verification.md
davecheney: good catch.

PR #1043, opened 2019-04-24
Fix typo in statsd-enabled flag for contour deployment manifest

zhilingc:
Changed the flag from `statsd-enable` to `statsd-enabled`.

Didn't open an issue since its a small typo fix, hope that's ok.
Ref: https://github.com/heptio/contour/blob/master/cmd/contour/contour.go#L59

### DIFF
--- a/deployment/deployment-grpc-v2/02-contour.yaml
+++ b/deployment/deployment-grpc-v2/02-contour.yaml
@@ -58,8 +58,8 @@ spec:
         command: ["contour"]
         args:
         - bootstrap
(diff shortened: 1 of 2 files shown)


stevesloka:
Thanks for the fix @zhilingc! I'm hoping we can do away with this flag soon. =)

Could you please rebase your two commits into one as well as sign the commit? Thanks!

zhilingc:
@stevesloka done :).

zhilingc:
thanks @stevesloka! & done, should be rebased now.

Issue #956, opened 2019-03-21
Reenable the make staticcheck target


davecheney:
staticcheck OOMs in go module mode, see dominikh/go-tools#419.

Reenable the staticcheck target in make check once the bug is fixed in staticcheck.

davecheney:
The upstream issue has been improved. If someone wants to address this issue the first step is

go get -u honnef.co/go/tools/cmd/staticcheck

inside the heptio/contour directory -- make sure you are running go 1.12 or later.

Then readd the `staticcheck` target to `make test`

The fix any errors reported

Then wrap the whole in thing in a PR that closes this issue.

PR #1044, opened 2019-04-24
Re-enable staticcheck

rohandvora:
Adds back staticcheck to make check target.

Updates: #956

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ PROJECT = contour
 REGISTRY ?= gcr.io/heptio-images
 IMAGE := $(REGISTRY)/$(PROJECT)
 SRCDIRS := ./cmd ./internal ./apis
-PKGS := $(shell go list -mod=readonly ./cmd/... ./internal/...)
+PKGS := $(shell GO111MODULE=on go list -mod=readonly ./cmd/... ./internal/...)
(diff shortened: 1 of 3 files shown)


internal/contour/holdoff.go
```diff
@@ -76,9 +76,6 @@ func (hn *HoldoffNotifier) OnChange(builder *dag.Builder) {
 // counter holds an atomically incrementing counter.
 type counter uint64
 
-func (c *counter) get() uint64 {
```
stevesloka: Why did this get deleted?

rohandvora: The static check tool was complaining that this function was not being used. And since inc() and reset() themselves return the updated value, get() was not being used in the code.

Makefile
```diff
@@ -2,7 +2,7 @@ PROJECT = contour
 REGISTRY ?= gcr.io/heptio-images
 IMAGE := $(REGISTRY)/$(PROJECT)
 SRCDIRS := ./cmd ./internal ./apis
-PKGS := $(shell go list -mod=readonly ./cmd/... ./internal/...)
+PKGS := $(shell GO111MODULE=on go list -mod=readonly ./cmd/... ./internal/...)
```
davecheney: This shouldn't be needed if you're running Go 1.12 or later. What error did this fix?

PR #1045, opened 2019-04-25
Implement upstream tls backend verification with ca cert and required san

### LINKED ISSUE #813: Support TLS verification of backend services

stevesloka:
Fixes #813 by implementing tls backend verification via CA cert as well as an optional subject alt name verification on the cert. 


stevesloka:
I still need to do a full test of this on a live cluster, but wanted to get the code out in front of eyes. One thing I want to do is const the `ca.crt` in the key unless we go with a dynamic route. 

Also, I'm aware the design still isn't merged but thought this was enough to get us started and could easily adapt.

stevesloka:
I've been doing some testing of this and I think it needs some more design thought:

1. If I specify the tls annotation on the service, I can immediately proxy tls to the upstream without verification. This may not be a desire that administrators want to allow, nothing forces a user to specify the ca cert. I wonder if we should rename the `validation` struct we added to the route's service and add what functionality the tls annotation is doing. (@davecheney)
2. If the cert is missing then the same for #1 above applies. I think we should also be setting the ingressroute status and reject the change since it's invalid. 
3. I need to further test, but I'm not sure if rotating a cert happened automatically which it should

internal/dag/dag.go
```diff
@@ -93,6 +93,17 @@ type RetryPolicy struct {
 	PerTryTimeout time.Duration
 }
 
+// UpstreamValidation defines how to validate the certificate on the upstream service
+type UpstreamValidation struct {
+	// Certificate holds a reference to the Secret containing the CA to be used to
+	// verify the upstream connection.
+	Certificate *Secret
```
robbiemcmichael: I think it would be good for the for the field name to be clear that it's a _CA_ certificate.

stevesloka: Yeah I agree that more clear.

internal/dag/builder.go
```diff
@@ -171,13 +176,25 @@ func (b *builder) addHTTPService(svc *v1.Service, port *v1.ServicePort, strategy
...
 		Protocol: protocol,
 	}
 	b.services[s.toMeta()] = s
 	return s
 }
 
+func (b *builder) addUpstreamValidation(uv *ingressroutev1.UpstreamValidation, svc *v1.Service) *UpstreamValidation {
+	if uv != nil {
+		val := &UpstreamValidation{
+			Certificate: b.lookupSecret(meta{name: uv.CASecret, namespace: svc.Namespace}),
```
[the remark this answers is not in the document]
stevesloka: yes that is correct. You can't reference a service from a different namespace from an IngressRoute and I need a way to see what namespace the IngressRoute is currently located in.

stevesloka:
I think this is ready to go! =)

apis/contour/v1beta1/ingressroute.go
```diff
@@ -141,6 +143,14 @@ type RetryPolicy struct {
 	PerTryTimeout string `json:"perTryTimeout,omitempty"`
 }
 
+// UpstreamValidation defines how to verify the backend service's certificate
+type UpstreamValidation struct {
+	// Name of the Kubernetes secret be used to validate the certificate presented by the backend
+	CACertificate string `json:"caSecret"`
+	// Ket which is expected to be present in the 'subjectAltName' of the presented certificate
+	SubjectName string `json:"subjectName,omitempty"`
```
davecheney: Please remove omit empty, I decided at the last minute that validation of the CA is not sufficient. We'll make both of these parameters mandatory in 0.12

docs/ingressroute.md
```diff
@@ -273,6 +273,48 @@ spec:
           permitInsecure: true
 ```
 
+#### Upstream TLS
+
+An IngressRoute route can proxy to an upstream TLS connection by first annotating the upstream Kubernetes service with: `contour.heptio.com/upstream-protocol.tls: "443,https"`.
+This annoation tells Contour which port should be used for the TLS connection.
+In this example, the upstream service is named `https` and uses port `443`.
+Additionally, it is possible for Envoy to verify the backend service's certificate.
+The service of an `IngressRoute` can optionally specify a `validation` struct which has a manditory `caSecret` key as well as an optional `subjectName`.
```
davecheney: s/optional/manadatory

internal/dag/builder.go
```diff
@@ -126,17 +126,18 @@ func (b *builder) lookupTCPService(m meta, port intstr.IntOrString, strategy str
...
-		namespace:   m.namespace,
-		port:        int32(port.IntValue()),
-		strategy:    strategy,
-		healthcheck: healthcheckToString(hc),
+		name:               m.name,
+		namespace:          m.namespace,
+		port:               int32(port.IntValue()),
+		strategy:           strategy,
+		healthcheck:        healthcheckToString(hc),
+		upstreamvalidation: upstreamValidationToString(uv),
```
davecheney: Sorry, this is not correct. The upstream validation is not a property of the service, it is a propery of the cluster -- the path from the route to the service.

internal/dag/builder.go
```diff
@@ -171,13 +176,26 @@ func (b *builder) addHTTPService(svc *v1.Service, port *v1.ServicePort, strategy
...
 	return s
 }
 
+func (b *builder) addUpstreamValidation(uv *ingressroutev1.UpstreamValidation, svc *v1.Service) *UpstreamValidation {
+	if uv != nil {
+		val := &UpstreamValidation{
+			CACertificate: b.lookupSecret(meta{name: uv.CACertificate, namespace: svc.Namespace}),
+			SubjectName:   uv.SubjectName,
+		}
+		fmt.Println("-------- val found: ", val)
```
davecheney: debug code, please delete

internal/dag/builder.go
```diff
@@ -171,13 +176,26 @@ func (b *builder) addHTTPService(svc *v1.Service, port *v1.ServicePort, strategy
...
 			HealthCheck:        hc,
+			UpstreamValidation: b.addUpstreamValidation(uv, svc),
 		},
 		Protocol: protocol,
 	}
 	b.services[s.toMeta()] = s
 	return s
 }
 
+func (b *builder) addUpstreamValidation(uv *ingressroutev1.UpstreamValidation, svc *v1.Service) *UpstreamValidation {
```
davecheney: please don't call this add, it doesn't add anything to the state of the builder. lookupUpstreamValidation might be a better name, but perhaps a different design is needed.

internal/dag/dag.go
```diff
@@ -20,7 +20,7 @@ import (
 
 	"github.com/envoyproxy/go-control-plane/envoy/api/v2/auth"
 	ingressroutev1 "github.com/heptio/contour/apis/contour/v1beta1"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
davecheney: goimports turd

internal/dag/dag.go
```diff
@@ -93,6 +93,17 @@ type RetryPolicy struct {
 	PerTryTimeout time.Duration
 }
 
+// UpstreamValidation defines how to validate the certificate on the upstream service
+type UpstreamValidation struct {
+	// CACertificate holds a reference to the Secret containing the CA to be used to
+	// verify the upstream connection.
+	CACertificate *Secret
+
+	// SubjectName holds an optional subject name which Envoy will check against the
```
davecheney: s/optional/required

internal/e2e/cds_test.go
```diff
@@ -18,14 +18,14 @@ import (
 	"testing"
 	"time"
 
-	"github.com/envoyproxy/go-control-plane/envoy/api/v2"
+	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
```
davecheney: goimports turds here and elsewhere

internal/envoy/auth.go
```diff
@@ -47,12 +47,34 @@ var (
 // UpstreamTLSContext creates an auth.UpstreamTlsContext. By default
 // UpstreamTLSContext returns a HTTP/1.1 TLS enabled context. A list of
 // additional ALPN protocols can be provided.
-func UpstreamTLSContext(alpnProtocols ...string) *auth.UpstreamTlsContext {
-	return &auth.UpstreamTlsContext{
+func UpstreamTLSContext(cert []byte, subjaltname []string, alpnProtocols ...string) *auth.UpstreamTlsContext {
```
davecheney: s/cert/ca

internal/envoy/auth.go
```diff
@@ -47,12 +47,34 @@ var (
...
 	}
+
+	validation := validationContext(cert, subjaltname)
+	if validation != nil {
+		context.CommonTlsContext.ValidationContextType = validation
+	}
+	return context
+}
+
+func validationContext(cert []byte, subjaltname []string) *auth.CommonTlsContext_ValidationContext {
```
davecheney: two things

1. please put the guard clause first; ie

```go
if cert == nil {
   // no validation required
   return nil
```

b. please don't push this logic into validationContext. If its not supposed to be called with cert == nil, then don't call it with cert == nil. The reason I ask this is no validationContext _sometimes_ returns nil depending on its inputs, that means we have nils floating around the data structures, this is how you get java :). We should at all times be following these two rules

1. Don't use nil to indicate a failure
2. Reduce the propagation of nils through the data structures.

internal/envoy/cluster.go
```diff
@@ -51,6 +60,22 @@ func Cluster(c *dag.Cluster) *v2.Cluster {
 	}
 }
 
+func upstreamValidationCACert(upstream *dag.HTTPService) []byte {
+	if upstream.UpstreamValidation != nil {
```
davecheney: Please change these conditions so that the success case proceeds down the page. The guard clause should be 
```go
if upstreamValidation == nil {
   // right, it's nil, what next
  return nil
}
return upstream.UpstreamValidation.CACertificate.Object.Data[CACertificateKey]

File: internal/envoy/cluster.go [hunk 14]
```diff
@@ -147,6 +172,12 @@ func Clustername(cluster *dag.Cluster) string {
 		}
 		buf += hc.Path
 	}
+	if uv := service.UpstreamValidation; uv != nil {
+		buf += uv.CACertificate.Object.ObjectMeta.Name
+		if len(uv.SubjectName) > 0 {
```
davecheney: you can elide this check

buf += "" is just buf

internal/e2e/cds_test.go
stevesloka: Is there a way to disable this? I can turn off the imports, but it's nice to manage for me.

davecheney: I'm not sure, it just started to happen to me when I upgraded to vim-go 1.20. Maybe we need to file an issue upstream about goimports being silly.

apis/contour/v1beta1/tlscertificatedelegation.go
```diff
@@ -40,7 +40,7 @@ type CertificateDelegation struct {
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
-// TLSCertificateDelefgation is an TLS Certificate Delegation CRD specificiation.
+// TLSCertificateDelefgation is an TLS CACertificate Delegation CRD specificiation.
```
davecheney: Please revert this, I think it was an unintended change.

robbiemcmichael: The comment also has a typo: TLSCertificateDele**f**gation

davecheney:
@stevesloka ping, this is over to you to respond to review comments. Please ping me when it's ready for the next round of review.

internal/e2e/cds_test.go
stevesloka: This is apparently by design: https://github.com/golang/go/issues/30051#issuecomment-461913682

I'll turn off goimports for now and fix up manually.

stevesloka:
@davecheney I think this is ready for another look. The travis build keeps failing, not sure why, I'll kick it again in a bit.

PR #1046, opened 2019-04-25
internal/grpc: adds SDS and registers xds for sds

mattalberts:
Addresses the first high-level design bullet point, creatign a gRPC
server for SDS by adding an SDS cache interface, implementing
FetchSecrets and StreamSecrets, and registering the server as an SDS
handler.

Updates #898

### DIFF
--- a/internal/grpc/resources.go
+++ b/internal/grpc/resources.go
@@ -17,6 +17,7 @@ import (
 	"sort"
 
 	"github.com/envoyproxy/go-control-plane/envoy/api/v2"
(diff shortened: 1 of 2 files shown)


mattalberts:
I've left the block to register the resource type out until the second bullet point on the high-level design

```
				secretType: &SDS{
					Cache: cacheMap[secretType],
				},
```
> once `cacheMap` carries a `secretType`

mattalberts:
@stevesloka rebased and ready :)

PR #1047, opened 2019-04-25
internal/contour: add SecretCache and link to grpc

mattalberts:
Addresses the second high-level design bullet point, creating a
SecretCache to be used by Contour to internally identify any changes.
The cache is also registered as a resource type for the gRPC server,
allowing it to respond to requests.
    
Updates: #898
Signed-off-by: Matt Alberts <malberts@cloudflare.com>

PR #1048, opened 2019-04-25
cmd/contour: adds sds cli

mattalberts:
Addresses the third high-level design bullet point, support sds in the
contour cli for debugging.
    
Updates: #898

### DIFF
--- a/cmd/contour/cli.go
+++ b/cmd/contour/cli.go
@@ -29,6 +29,7 @@ const (
 	clusterType  = typePrefix + "Cluster"
 	routeType    = typePrefix + "RouteConfiguration"
 	listenerType = typePrefix + "Listener"
+	secretType   = typePrefix + "auth.Secret"
 )
 
 type Client struct {
--- a/cmd/contour/contour.go
+++ b/cmd/contour/contour.go
@@ -76,6 +76,8 @@ func main() {
 	lds.Arg("resources", "LDS resource filter").StringsVar(&resources)
 	rds := cli.Command("rds", "watch routes.")
 	rds.Arg("resources", "RDS resource filter").StringsVar(&resources)
+	sds := cli.Command("sds", "watch secrets.")
+	sds.Arg("resources", "SDS resource filter").StringsVar(&resources)
 
 	serve := app.Command("serve", "Serve xDS API traffic")
 	inCluster := serve.Flag("incluster", "use in cluster configuration.").Bool()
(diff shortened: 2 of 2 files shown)


mattalberts:
The review and commit order would be
 1. #1046 
 1. #1047 
 1. #1048

internal/dag/dag.go
```diff
@@ -293,6 +293,16 @@ func (s *Secret) Data() map[string][]byte {
 	return s.Object.Data
 }
 
+// Cert returns the secret's tls certificate
+func (s *Secret) Cert() []byte {
+	return s.Object.Data[v1.TLSCertKey]
```
stevesloka: Could there be a case where the cert is not in the secret?

internal/dag/dag.go
```diff
@@ -293,6 +293,16 @@ func (s *Secret) Data() map[string][]byte {
...
+}
+
+// PrivateKey returns the secret's tls private key
+func (s *Secret) PrivateKey() []byte {
+	return s.Object.Data[v1.TLSPrivateKeyKey]
```
stevesloka: Same here if the key is nil?

internal/dag/dag.go
mattalberts: It's plausible.  It is assumed to be valued when setting up the downstream context; but on review, that isn't guaranteed. 
 - `envoy.DownstreamTLSContext(data[v1.TLSCertKey], data[v1.TLSPrivateKeyKey], vh.MinProtoVersion, alpnProtos...)`
 - https://github.com/heptio/contour/blob/master/internal/contour/listener.go#L265

It feels like the the structure should be verified while creating a `SecureVirtualHost` vertex.

internal/dag/dag.go
mattalberts: Agreed. if the key doesn't exist, the return would be populated as and empty `[]byte`.

internal/dag/dag.go
mattalberts: These got marked as outdated when I rebased. They were brought in with #1047.

stevesloka: ok no biggie, there's another issue where we get the cert and not the key and it causes issues, but let's not try and fix that in this (just been on my mind).

internal/dag/dag.go
stevesloka: :+1:

internal/dag/dag.go
mattalberts: Its okay! I'm glad you mentioned it.  I remember looking at it while I drafted the PR, when trying to decide on introducing the helper functions on *dag.Secret.  Is there a already an issue for that?  I can clean them all up at one time in another PR :).

mattalberts: I bet its this one, #1051  :)

Issue #999, opened 2019-04-09
Upgrade go-control-plane to latest


davecheney:
Upgrade to go-control-plane 0.7.1 to pick up the grpc definitions for the recent CVE fix.

PR #1055, opened 2019-04-30
Updated go-control-plane to v0.7.1

rohandvora:
The newer version of go-control-plane has some fields removed like
RouteAction.RequestHeadersToAdd[] which were deprecated in v0.6.9.
Due to this, the existing implementation which added headers at the
weightedCluster and RouteAction layer was changed. Headers are now
added at the route layer as per discussion with @stevesloka.

#Fixes: #999


stevesloka:
Updates #876

Issue #1051, opened 2019-04-26
Contour sends invalid configuration to Envoy when TLS secret is not well formed.


mattmoyer:
We are running into an issue using Contour with [cert-manager](https://github.com/jetstack/cert-manager).

The setup:
- We create a cert-manager Certificate object.
- The cert-manager controller provisions a private key and starts requesting the certificate from Let's Encrypt.
- It saves the private key into the certificate Secret, in the following form:
   ```yaml
   apiVersion: v1
   kind: Secret
   metadata:
     name: my-tls-secret
     namespace: my-namespace
     labels:
       certmanager.k8s.io/certificate-name: my-tls-cert
   type: kubernetes.io/tls
   data:
     ca.crt: null
     tls.crt: null
     tls.key: # a valid base64 encoded private key [....]
   ```
- Let's Encrypt denies our certificate request based on their rate limiting policy, so the Secret stays in this half-created state for an extended period of time.
- Contour sees that the Ingress and associated Secret exist, and attempts to configure Envoy with the new values.
- Envoy rejects the configuration and emits an error log:
   ```
   [2019-04-26 16:32:29.677][000001][warning][config] [bazel-out/k8-opt/bin/source/common/config/_virtual_includes/grpc_mux_subscription_lib/common/config/grpc_mux_subscription_impl.h:70] gRPC config for type.googleapis.com/envoy.api.v2.Listener rejected: Proto constraint validation failed (ListenerValidationError.FilterChains[i]: ["embedded message failed validation"] | caused by FilterChainValidationError.TlsContext: ["embedded message failed validation"] | caused by DownstreamTlsContextValidationError.CommonTlsContext: ["embedded message failed validation"] | caused by CommonTlsContextValidationError.TlsCertificates[i]: ["embedded message failed validation"] | caused by TlsCertificateValidationError.CertificateChain: ["embedded message failed validation"] | caused by field: "specifier", reason: is required): name: "ingress_https"
   ```
- It then dumps the entire configuration including private keys to the log file 😬. The `tls_context` section that I think is causing problems looks like this (note empty `certificate_chain` field):
   ```
   filter_chains {
     filter_chain_match {
       server_names: "myhost.example.com"
     }
     tls_context {
       common_tls_context {
         tls_params {
           tls_minimum_protocol_version: TLSv1_1
           tls_maximum_protocol_version: TLSv1_3
         }
         tls_certificates {
           certificate_chain {
           }
           private_key {
[...]

PR #1056, opened 2019-05-01
internal/dag: Fix secret lookup

rohandvora:
Fixes secret lookup in Dag builder. It returns nil if certificate or
private key is empty/null in the kubernetes secret.

Fixes: #1051

### DIFF
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -206,6 +206,9 @@ func (b *builder) lookupSecret(m meta) *Secret {
 	if !ok {
 		return nil
 	}
+	if !validSecret(sec) {
+		return nil
+	}
 	s := &Secret{
 		Object: sec,
 	}
@@ -624,6 +627,11 @@ func (b *builder) rootAllowed(ir *ingressroutev1.IngressRoute) bool {
 		}
 	}
 	return false
+}
+
+// validSecret returns true if the Secret contains certificate and private key material.
+func validSecret(s *v1.Secret) bool {
+	return len(s.Data[v1.TLSCertKey]) > 0 && len(s.Data[v1.TLSPrivateKeyKey]) > 0
 }
 
 func (b *builder) processRoutes(ir *ingressroutev1.IngressRoute, prefixMatch string, visited []*ingressroutev1.IngressRoute, host string, enforceTLS bool) {
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -1266,6 +1266,15 @@ func TestDAGInsert(t *testing.T) {
 		Data: secretdata("certificate", "key"),
 	}
 
+	// Invalid cert in the secret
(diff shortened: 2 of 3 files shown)


internal/dag/builder.go
```diff
@@ -209,11 +209,14 @@ func (b *builder) lookupSecret(m meta) *Secret {
 	s := &Secret{
 		Object: sec,
 	}
-	if b.secrets == nil {
-		b.secrets = make(map[meta]*Secret)
+	if s.Cert() != nil && len(s.Cert()) > 0 && s.PrivateKey() != nil && len(s.PrivateKey()) > 0 {
```
stevesloka: Could we move this to a helper method, would make it easier to test the logic and clean up the logic flow.

rohandvora: Sure

internal/dag/builder.go
```diff
@@ -209,11 +209,14 @@ func (b *builder) lookupSecret(m meta) *Secret {
 	s := &Secret{
 		Object: sec,
 	}
-	if b.secrets == nil {
-		b.secrets = make(map[meta]*Secret)
+	if s.isValid() {
```
davecheney: Please move the success case outside the condition

```go
if !s.isValid() {
    return nil
}
if b.secrets == nil {
    b.secrets = make(map[meta]*Secret)
}
b.secrets[s.toMeta()] = s
return s
```
ref: https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

internal/dag/dag.go
```diff
@@ -303,6 +303,10 @@ func (s *Secret) PrivateKey() []byte {
 	return s.Object.Data[v1.TLSPrivateKeyKey]
 }
 
+func (s *Secret) isValid() bool {
+	return s.Cert() != nil && len(s.Cert()) > 0 && s.PrivateKey() != nil && len(s.PrivateKey()) > 0
```
davecheney: Thank you but I don't think this is right.

We should not be creating an invalid secret object and placing it in the dag. Either we filter this out here

https://github.com/heptio/contour/blob/master/internal/dag/builder.go#L570

which is not my preference because it encodes knowledge that secret objects are only present at some levels.

Or we filter it out earlier in `lookupSecret`. I'd like to try that. Can you please make this a function

```go
// validSecret returns true if the Secret contains 
func validSecret(s *v1.Secret) bool {
    return len(s.Data[v1.TLSCertKey]) > 0 && len(s.Data[v1.TLSPrivateKeyKey]) > 0
}
```
Also, len(s.Cert()) will work fine without the nil check.

davecheney:
Thank you for working on this, a few small changes needed.

internal/dag/dag.go
rohandvora: Sure. Thanks for the feedback!

PR #1057, opened 2019-05-01
Remove extra annotation from sample deployment

rsyvarth:
Removes an unnecessary annotation (`steve: "sloka"`) from the example `ds-hostnet-split` deployment. The change was added in https://github.com/heptio/contour/pull/1034

Signed-off-by: Robert Syvarth <robert.syvarth@gmail.com>


rsyvarth:
Done

Issue #1019, opened 2019-04-15
Document ingressroute request and timeout parameters


davecheney:
Fixes #815 

Document the use of ingressroute's request and timeout fields.

rohandvora:
I can take this up.

stevesloka:
Closed via #1042

PR #1067, opened 2019-05-07
internal/dag: Missing retry count implies 1 retry

davecheney:
Updates #1019

The retry policy documentation (#1042) states that a `retryPolicy:`
stanza with no fields implies a retry count of 1 -- if you don't want to
retry, don't add a `retryPolicy:` key.

This PR addresses this issue, and breaks out the retryPolicy logic into
a new file with associated tests to isolate this buisiness logic.

### DIFF
--- a/internal/dag/policy.go
+++ b/internal/dag/policy.go
@@ -1,0 +1,39 @@
+// Copyright © 2019 Heptio
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


internal/dag/policy.go
```diff
@@ -0,0 +1,39 @@
...
+	"time"
+
+	"github.com/heptio/contour/apis/contour/v1beta1"
+)
+
+func retryPolicyIngressRoute(rp *v1beta1.RetryPolicy) *RetryPolicy {
+	if rp == nil {
+		return nil
+	}
+	perTryTimeout, _ := time.ParseDuration(rp.PerTryTimeout)
```
vaamarnath: Shouldn't you check the error returned from `time.ParseDuration` call?

davecheney: It's not specified in the documentation that ParseDuration returns a zero duration in an error condition, but its been like that since pre Go 1.0 so we can utilise this here. We have tests that will catch if the behaviour of ParseDuration ever changes -- not that it will.

PR #1070, opened 2019-05-07
internal/envoy: listener idle timeouts

### LINKED ISSUE #1054: Listener IdleTimeout
Browsers such as Chrome have a tendency to keep TCP connections open after the HTTP request-response cycle is done to be reused by future requests and improve the UX. Since Envoy's default is *no timeout*, connections can stay open indefinitely. In high-traffic environments, this can cause Envoy's memory consumption to grow very quickly and become unmanageable.

Reported by @rata & Johanan Liebermann in #contour slack (https://kubernetes.slack.com/archives/C8XRH2R4J/p1556480572023000), they provided a diff of what change they made to validate the issue:

```
diff --git a/internal/envoy/listener.go b/internal/envoy/listener.go
index 9001322..97bd94e 100644
--- a/internal/envoy/listener.go
+++ b/internal/envoy/listener.go
@@ -103,6 +103,7 @@ func HTTPConnectionManager(routename, accessLogPath string) listener.Filter {
                                        "access_log":         accesslog(accessLogPath),
                                        "use_remote_address": {Kind: &types.Value_BoolValue{BoolValue: true}}, // TODO(jbeda) should this ever be false?
                                        "normalize_path":     {Kind: &types.Value_BoolValue{BoolValue: true}},
+                                       "idle_timeout":       sv("60s"),
                                },
                        },
                },
```

We should allow this value to be customized via an arg that gets passed to Contour. Also, this should be applied to all listeners that get created (i.e. http/https). It might be good to come up with a sane default that would apply to a majority of environments.

mattalberts:
Hardcodes the listener idle_timeout to 60s.

Fixes #1054

### DIFF
--- a/internal/envoy/listener.go
+++ b/internal/envoy/listener.go
@@ -103,6 +103,8 @@ func HTTPConnectionManager(routename, accessLogPath string) listener.Filter {
 					"access_log":         accesslog(accessLogPath),
 					"use_remote_address": {Kind: &types.Value_BoolValue{BoolValue: true}}, // TODO(jbeda) should this ever be false?
 					"normalize_path":     {Kind: &types.Value_BoolValue{BoolValue: true}},
+					"idle_timeout":       sv("60s"),
+					"request_timeout":    sv("15s"),
 				},
 			},
 		},


davecheney:
@mattalberts thanks for working on this. I'm 100% on board with idle timeout, that's a no brainer. I'm on the fence about the request timeout limit, I think 15 seconds is reasonable, but if there are people with large uploads then this will blow them up. I _really_ don't want to make this value configurable because its complicated to make it configurable per vhost, and I also don't want to expose it to ingressroute users.

I'm wondering if we can get away with just setting the idle timeout for now; that will close idle connections. It would be good to have a request body timeout as well, but I'm wondering if we can get away with it for now. What do you think?

davecheney:
Always when confronted with problems like this I consult what other web servers do, 

https://nginx.org/en/docs/http/ngx_http_core_module.html#client_body_timeout

nginx is sort of inconsistent here, they have a 60s timeout for headers -- which is protection against slowloris, but use a 60s timeout between reads on the client body, so if you want to hold the connection open for a week spoon feeding a byte at a time that'll be just fine.

I looked quickly but couldn't find anything for how apache limits the read time of a request; only the number of bytes read.

haproxy follows nginx's lead and specifics an inter packet timeout; the defaults aren't obvious, but 50000ms was spotted once.

I think that for 0.12 we should take idle=60 and leave request timeout unspecified.

mattalberts:
I was worried about a hard-coded `request_time` impacting current use-cases. Given other web-server behavior and it being not configurable for this iteration, I think you made the right choice (only take the `idle_timeout` )

PR #1077, opened 2019-05-10
internal/dag: respect force-ssl-redirect over allow-http annotation

### LINKED ISSUE #1023: allow-http and force-ssl annotations have surprising/confusing combined behaviour
I'm not convinced that this is a bug, but I think it's worth discussion.

_Step 1_

Initially, we had an `Ingress` created several months ago with the following annotation:

```
annotations:
  kubernetes.io/ingress.allow-http: "false"
```

Insecure HTTP endpoints were responding with 404:

```
HTTP/1.1 404 Not Found
date: Tue, 16 Apr 2019 05:16:24 GMT
server: envoy
content-length: 0
```

This seems correct, as [the docs](https://github.com/heptio/contour/blob/master/docs/annotations.md#standard-kubernetes-ingress-annotations) have this to say:

> Instructs Contour to not create an Envoy HTTP route for the virtual host. The Ingress exists only for HTTPS requests.

_Step 2_

We added the `force-ssl-redirect` annotation, as we wanted insecure HTTP requests to be redirected to secure endpoints.

```
annotations:
  kubernetes.io/ingress.allow-http: "false"
  ingress.kubernetes.io/force-ssl-redirect: "true"
```

Insecure endpoints still returned a 404, rather than an SSL redirect:

```
HTTP/1.1 404 Not Found
date: Tue, 16 Apr 2019 05:23:36 GMT
server: envoy
content-length: 0
```

This caused some confusion for us - especially as we had forgotten the `allow-http` annotation was set.

_Step 3_

We suspected that `allow-http` being present was getting in the way of `force-ssl-redirect` having the desired effect, so we removed it:

```
annotations:
  ingress.kubernetes.io/force-ssl-redirect: "true"
```

Now we get the redirect as desired:

```
HTTP/1.1 301 Moved Permanently
location: https://www.example.com/api/health_check
date: Tue, 16 Apr 2019 05:28:58 GMT
server: envoy
content-length: 0
```

I'm not entirely sure, but it was surprising.

These annotations are arguably simply incompatible, and in an ideal world I think it should not be possible to specify them both.  Semantically, it doesn't make sense to have both of them - they contradict each other. But in this case we're using a standard `Ingress` resource, not an `IngressRoute` so according to my (limited) understanding, it would not be trivial to implement validation of that in Contour at deploy- / apply-time. I do not feel comfortable suggesting run-time validation and failure, because then consumers of Contour would have to be closely watching the logs to identify mistakes.

Since both annotations are grouped under `Standard Kubernetes Ingress annotations` in the Contour docs, this pain point could potentially be documented in that section. If that's the way forward, I'm happy to draft some copy for the docs and a put up a PR.

Alternatively, Contour could just ensure that `force-ssl-redirect` always wins, so the `allow-http` annotation basically becomes a no-op when they're both specified. This would be the correct behaviour in all use cases that I can think of, but that may be naive.

:)

**Environment:**

- Contour version: `v0.11.0`
- Kubernetes version: (use `kubectl version`): `v1.14.0`
- Kubernetes installer & version: `1.11.7-gke.6`
- OS (e.g. from `/etc/os-release`): nodes are running CoreOS, Contour and Envoy are both running on the standard images from manifest included in the Contour release

ceralena:
Fixes #1023

The behaviour when these two annotations are both specified is somewhat
surprising: `allow-http: false` meant that no routes were not registered
for port 80, even if `force-ssl-redirect: true` was set - leading to a
404 where a 3xx with an upgrade to `https` was expected.

With this commit, we prioritize respecting `force-ssl-redirect`: if this
annotation is specified and set to `true`, we always register a port 80
route for the ingress, even if `allow-http: false`, so that the forced
upgrade can take effect.

Signed-off-by: Cera Davies <ceralena.davies@gmail.com>

### DIFF
--- a/docs/annotations.md
+++ b/docs/annotations.md
@@ -16,6 +16,7 @@ However, Contour still supports a number of annotations on the Ingress resources
  - `ingress.kubernetes.io/force-ssl-redirect`: Requires TLS/SSL for the Ingress to Envoy by setting the [Envoy virtual host option require_tls](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route.proto.html#envoy-api-field-route-virtualhost-require-tls)
  - `kubernetes.io/ingress.allow-http`: Instructs Contour to not create an Envoy HTTP route for the virtual host. The Ingress exists only for HTTPS requests. Specify `"false"` for Envoy to mark the endpoint as HTTPS only. All other values are ignored.
 
+The `ingress.kubernetes.io/force-ssl-redirect` annotation takes precedence over `kubernetes.io/ingress.allow-http`. If they are set to `"true"` and `"false"` respectively, Contour *will* create an Envoy HTTP route for the Virtual host, and set the `require_tls` virtual host option.
 
 ## Contour specific Ingress annotations
 
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -472,7 +472,7 @@ func (b *builder) computeIngresses() {
 				}
 
 				// should we create port 80 routes for this ingress
-				if httpAllowed(ing) {
+				if tlsRequired(ing) || httpAllowed(ing) {
 					b.lookupVirtualHost(host).addRoute(r)
 				}
 
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -179,6 +179,26 @@ func TestDAGInsert(t *testing.T) {
 			}},
 		},
 	}
+	i6c := &v1beta1.Ingress{
+		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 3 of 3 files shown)


ceralena:
There was a little confusion in the issue discussion, but I'm fairly confident that this is the desired behaviour that came out of it - so I figured I'd go ahead and create this PR to validate it. Happy to go back if this is wrong. Thanks!

davecheney:
Assigning to @rohandvora to review.

ceralena:
Thanks for the feedback. I've squashed a commit to update the docs and rebased off `master`.

youngnick:
Sorry @ceralena, can I ask for one last rebase? Then I can merge this in, thanks for the great fix!

PR #1078, opened 2019-05-10
Validate secret and set status on IngressRoute if secret is invalid 

### LINKED ISSUE #517: Validation: Set status on IngressRoute if TLS secret invalid

stevesloka:
Fixes #517 by setting the status of an IngressRoute if the TLS secret is not found or malformed. 

Example:
```
Status:
  Current Status:  invalid
  Description:     TLS Secret not found or is malformed
```
 
### DIFF
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -21,7 +21,7 @@ import (
 	"strconv"
 	"strings"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/api/extensions/v1beta1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 
@@ -197,6 +197,7 @@ func (b *builder) addTCPService(svc *v1.Service, port *v1.ServicePort, strategy 
 	return s
 }
 
+// lookupSecret returns a kuberntes secret if found or nil if not
 func (b *builder) lookupSecret(m meta, validate func(*v1.Secret) bool) *Secret {
 	if s, ok := b.secrets[m]; ok {
 		return s
@@ -515,7 +516,9 @@ func (b *builder) computeIngressRoutes() {
 		if tls := ir.Spec.VirtualHost.TLS; tls != nil {
 			// attach secrets to TLS enabled vhosts
 			m := splitSecret(tls.SecretName, ir.Namespace)
-			if sec := b.lookupSecret(m, validSecret); sec != nil && b.delegationPermitted(m, ir.Namespace) {
+			sec := b.lookupSecret(m, validSecret)
+			secretInvalidOrNotFound := sec == nil
+			if sec != nil && b.delegationPermitted(m, ir.Namespace) {
 				svhost := b.lookupSecureVirtualHost(host)
 				svhost.Secret = sec
 				svhost.MinProtoVersion = minProtoVersion(ir.Spec.VirtualHost.TLS.MinimumProtocolVersion)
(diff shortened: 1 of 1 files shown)


internal/dag/builder.go
```diff
@@ -215,7 +213,7 @@ func (b *builder) lookupSecret(m meta, validate func(*v1.Secret) bool) *Secret {
 		b.secrets = make(map[meta]*Secret)
 	}
 	b.secrets[s.toMeta()] = s
-	return s
+	return s, false
```
davecheney: Is this correct? if we return the secret then it _is_ found.

internal/dag/builder.go
```diff
@@ -515,15 +513,21 @@ func (b *builder) computeIngressRoutes() {
 		if tls := ir.Spec.VirtualHost.TLS; tls != nil {
 			// attach secrets to TLS enabled vhosts
 			m := splitSecret(tls.SecretName, ir.Namespace)
-			if sec := b.lookupSecret(m, validSecret); sec != nil && b.delegationPermitted(m, ir.Namespace) {
+			sec, secretInvalidOrNotFound := b.lookupSecret(m, validSecret)
+			if sec != nil && b.delegationPermitted(m, ir.Namespace) {
```
davecheney: This is the only place that the second boolean parameter is begin used. Would this workinstead

```go
sec := b.lookupSecret(m, validSecret)
secretInvalidOrNotFound := sec == nil
```

stevesloka: Yup that works! For some reason, I was making that more complicated than it needed to be.

internal/dag/builder.go
```diff
@@ -197,15 +197,13 @@ func (b *builder) addTCPService(svc *v1.Service, port *v1.ServicePort, strategy
...
 func (b *builder) lookupSecret(m meta, validate func(*v1.Secret) bool) *Secret {
 	if s, ok := b.secrets[m]; ok {
 		return s
 	}
 	sec, ok := b.source.secrets[m]
-	if !ok {
-		return nil
-	}
-	if !validate(sec) {
+	if !ok || !validate(sec) {
```
davecheney: I'm not sure this is an improvement in readability, i had to squint real hard to convince myself that if ok is true then the logic falls through to validate(sec)

Given that both validate functions, validSecret and validCA will work just fine, ie not panic, but return false if passed a nil. You could rewrite this as

```go
sec := b.source.secrets[m]
if !validate(sec) {
       return nil
}
```

davecheney: opps, my bad, this wasn't correct. if sec == nil then both validate functions will panic. My recommendation is to roll back the change to this function. thanks for the comment,

stevesloka: I've revered this back to the way it was

internal/dag/builder.go
```diff
@@ -197,6 +197,7 @@ func (b *builder) addTCPService(svc *v1.Service, port *v1.ServicePort, strategy
 	return s
 }
 
+// lookupSecret returns a kuberntes secret if found or nil if not
```
davecheney: s/kuberntes/kubernetes/

PR #1080, opened 2019-05-10
design: JSON access logs

azuretek:
Updates https://github.com/heptio/contour/issues/624

Signed-off-by: Abi X Renhart (abi.renhart@outreach.io)

### DIFF
--- a/design/json-access-logs-design.md
+++ b/design/json-access-logs-design.md
@@ -1,0 +1,74 @@
+# JSON Access Logs
+
+Status: Draft
+
+This document proposes a change to re-configure the default access log format to a structure JSON format.
+
+## Goals
+
+- Create an accepted format for JSON structured access logs produced by envoy.
+- Allow backends like Elasticsearch to consume structured access log data for easy ingestion and reporting.
+
+## Non Goals
+
+- We are not converting all envoy log output to JSON, just the access logs.
+
+## Background
+
+Access logs are currently only output using the default access log format that envoy provides.
+The default format has several issues, but the biggest issues are that it makes querying more expensive and harder to craft for the average user.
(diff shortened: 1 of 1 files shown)


davecheney:
Thank you. It’s the weekend in Australia. I’ll respond on Monday.

davecheney:
Thank you for taking the time to write this up. We’re still debating it internally, I’ll commit to replying in full by the end of the week.

davecheney:
I'm sorry I was not able to make the friday deadline. I'm sorry but I will not be able to give you and ETA til later in the week.

tyrken:
While not an active Contour user (we went back to Nginx ingress for weak/unimportant reasons),
I feel I can wade in after too many years sorting out logging at medium sized companies.

1. I don't think it should be the default - stick with what you have now.

3. I'm afraid there is no standard JSON format, and you would indeed "MUST" make it fully flexible.  I would suggest making minimal-ideally-zero difference between the Envoy "Format Dictionary" config and how you surface it in Contour.  Lean on the Envoy designers to handle backwards compatibility or new changes, don't burden your CRE team/future-maintainers with translation.  Just include perhaps this design’s suggestion as a documented config example and link to the Envoy docs.

2. I guestimate 90% of users (certainly all my historical use-cases) would be happy with cluster/ingress-controller level config.  If you know of heavy multi-tenant Contour users then it may have to go down to ingress level – but I’m sure that could be left to a v2 of this feature.  Note that if done in a v2, different JSON formats cope with the “munged into /dev/stdout” issue just fine by design.

I would also add that my personal vote would be for flexible JSON, not diverting time to adding CLF.

davecheney:
@youngnick has put together this proposal, #1485 which is our prefered design.

PR #1082, opened 2019-05-13
docs: add how we work document

davecheney:
The team is growing 🎉 it seems like a good time to write down the way we work.

### DIFF
--- a/docs/how-we-work.md
+++ b/docs/how-we-work.md
@@ -1,0 +1,30 @@
+This page captures how we work on Contour.
+
+- When you pick up an issue, assign it to yourself.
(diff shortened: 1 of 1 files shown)


davecheney:
ping, still looking for a review.

davecheney:
Cool. Thanks everyone for your input. As Nick pointed out, this is a starting point for a discussion, but you gotta start somewhere.

davecheney:
> The only other thing that I thought of would be adding something around using the fixes <issue> / updates <issue> to auto-close things.

I'm *sure* that this is mentioned elsewhere in the CONTRIBUTING file. I'll make a note to double check.

PR #1089, opened 2019-05-14
internal/envoy: convert TCPProxy to TypedConfig

davecheney:
Updates #876
Updates #499

Convert TCPProxy listener filter to a TypedConfig. This gives us
protection if Envoy changes the proto definition of the filter types in
the future. It also reduces the background churn of idempotent, but not
bit-for-bit identical messages between Contour and Envoy.

This has lots of pleasent follow on effects; sorting of TCP Proxy
members is less gross, and we remove all uses of the odeous
messageToStruct via JSON hack.

Also, add tests for FileAccessLog and HTTPConnectionManager. They're
simple, but by adding a test we can then be sure that we can use them in
other functions as we have a test to assert what they are returning --
and if we change it in the future.

### DIFF
--- a/internal/e2e/e2e.go
+++ b/internal/e2e/e2e.go
@@ -19,7 +19,6 @@ import (
 	"testing"
 
 	"github.com/envoyproxy/go-control-plane/envoy/api/v2"
-	accesslog "github.com/envoyproxy/go-control-plane/envoy/config/accesslog/v2"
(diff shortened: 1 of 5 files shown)


davecheney:
> /lgtm Nice to see all the deleted code. =)

It's always a good day when you can delete code.

hank

PR #1094, opened 2019-05-15
Remove statsd from boostrap & examples

### LINKED ISSUE #1086: Remove --statsd-enabled and related code
There should be no code for now unsued flags, like `--statsd-enabled` and friends that is not needed since https://github.com/heptio/contour/pull/1085 (if/when it is merged).

I don't see a reason to not remove it, but would like some opinions first just in case (following the talk then code project recommendation :))

**Anything else you would like to add:** No

**Environment:**

- Contour version: master (before 0.13 release)

stevesloka:
Fixes #1086 by removing statsd sink from bootstrap config since it's no longer needed with Envoy >v1.10. Also removes `--statsd+` flags which are no longer needed.


internal/envoy/config.go
```diff
@@ -36,11 +36,11 @@ type ConfigWriter struct {
 	AdminPort int
 
 	// StatsAddress is the address that the /stats path will listen on.
```
rata: contour metrics path is `/metrics`. Is the comment outdated or am I missing something? :)

internal/envoy/config.go
```diff
@@ -36,11 +36,11 @@ type ConfigWriter struct {
...
-	// Defaults to 0.0.0.0 and is only enabled if StatsdEnabled is true.
+	// Defaults to 0.0.0.0
 	StatsAddress string
 
 	// StatsPort is the port that the /stats path will listen on.
```
rata: Idem

internal/envoy/config.go
rata: Oh, this is the envoy part. This is now `/stats/prometheus`

davecheney: yeah, its a bit confusing, this is the address of the listener we push to envoy for it to serve it's stats on.

The comment could probably be improved with ^^ that text.

stevesloka: :+1:

Issue #825, opened 2018-12-03
Support of external endpoints as targets for proxying


anapsix:
**Describe the solution you'd like**
Would love to be able to define `Ingress` and/or `IngressRoute` objects, that would allow specifying endpoints outside of k8s cluster as targets for proxying.
Unlike #334, this feature would not rely on Service object, and would achieve described functionality directly.

Manifest could look like this, for example:
```
---
apiVersion: contour.heptio.com/v1beta1
kind: IngressRoute
metadata:
  name: weight-shifting
  namespace: default
spec:
  virtualhost:
    fqdn: weights.bar.com
  routes:
    - match: /
      hosts:
        - name: s1
          host: s1.bar.com
          port: 80
          weight: 10
        - name: s2
          host: s2.bar.com
          port: 80
          weight: 90
```

davecheney:
Thank you for raising this feature request, I've added it to the backlog.

stevesloka:
Hey @anapsix we have support for external name services now (https://github.com/heptio/contour/blob/master/docs/ingressroute.md#externalname)!

Issue #334, opened 2018-04-19
[feature] support Services with ExternalName or ExternalIP components


davecheney:
The service API supports so called _external_ services. This issue is a placeholder for 

- [ ] investigate how `service.spec.external{Name,IP}` interacts with the endpoints api
- [X] investigate adding support for ExternalName or IP to contour ([design doc](https://github.com/heptio/contour/blob/master/design/external-names.md))

No milestone as of yet.

PR #1095, opened 2019-05-15
Design doc for externalNames

stevesloka:
Updates #334 & #825 by providing an initial design doc for proxying traffic to dns endpoints instead of Kubernetes endpoints.

### DIFF
--- a/design/external-names.md
+++ b/design/external-names.md
@@ -1,0 +1,51 @@
+# External Names Service
+
+_Status_: Draft
+
+This document looks to allow Contour to process Kubernetes service types `ExternalName` so that traffic can be proxied to dns hosts rather than only pod endpoints.
+
+## Goals
+
+- Allow for Contour to proxy traffic via Envoy to external dns endpoints
+- Utilize built-in Kubernetes object types (i.e. ExternalName service)
+
+## Non-goals
+
+- `externalIPs` will not be used, users who want to proxy to custom IP's should build their own service/endpoint object
+- Utilize a new CRD or ConfigMap style mapping
+- Only support HTTP at the moment, TCP will be addressed in a further design doc
+
+## Background
+
+Contour currently watches Kubernetes services and endpoints, streaming them to via Envoy xDS CDS & EDS endpoints.
+Kubernetes pods are identified by their presence in the Endpoint document and streamed to Envoy via EDS.
+Requests to envoy are routed directly to Kubernetes pods (i.e. endpoints).
+This proposal looks to add a second way to identify the members of a Envoy clustr object, using a DNS entry to supply the cluster members rather than EDS.
+
+## High-level Design
+
+Contour will need to watch for and process Kubernetes service types of `ExternalName` (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#servicespec-v1-core).
+When configuring a Route, Contour will use the contents of the `spec.externalName` of the service as well as the port from the `spec.ports` section of the Kubernetes service.
+
+If ports are not supplied for the Kubernetes service, then the IngressRoute status will be set accordingly with an error message.
+
+Allowing for TLS proxying and validation should still be applied as it exists already so users can proxy to a TLS upstream as well as validate the certificate+subject alt name of the upstream service.
+
+## Detailed Design
+
+- The `TCPService` struct in `internal/dag/dag.go` will have a new `externalName string` field added.
+- When services are added in `internal/dag/builder.go`, if `svc.Spec.Type == v1.ServiceTypeExternalName` then the value of `svc.Spec.ExternalName` will be set to the TCPService struct's `externalName`.
+- Since the service does not have any Kubernetes Endpoints, we cannot rely on having an EDS cluster to reference from the CDS cluster.
+For this type of service (i.e. `ExternalName`) when creating the cluster in `internal/envoy/cluster.go`, instead of setting a `EdsClusterConfig`, we'll configure a `LoadAssignment` which should allow us to specify the endpoint defined in `externalName` dynamically.
+
+## Alternatives
+
+Potentially, we could avoid using a Service of type `ExternalName` by simply just extending the `IngressRoute` CRD as proposed here: https://github.com/heptio/contour/issues/825#issue-386869426
+
+The downside to this approach is it couldn't be used with Ingress resources.
+
+## Security Considerations
+
+Potentially someone could use an `ExternalName` service to reference services in another namespace by setting the `externalName` (ex: `other-service.namespace-other.svc.cluster.local`).
+
+This scenario exists in a cluster regardless if Contour is deployed or not, other security restrictions should be applied if users want to avoid this from happening.


design/external-names.md
```diff
@@ -0,0 +1,47 @@
...
+This document looks to allow Contour to process Kubernetes service types `ExternalName` so that traffic can be proxied to dns hosts rather than only pod endpoints.
+
+## Goals
+
+- Allow for Contour to proxy traffic via Envoy to external dns endpoints
+- Utilize built-in Kubernetes object types (i.e. ExternalName service)
+
+## Non-goals
+
+- `externalIPs` will not be used, users who want to proxy to custom IP's should build their own service/endpoint object
```
davecheney: +1 for calling it out. I think we could relax this restriction later one, but for now this is a reasonable thing to place out of scope.

design/external-names.md
```diff
@@ -0,0 +1,47 @@
...
+- Utilize a new CRD or ConfigMap style mapping
+
+## Background
+
+Contour currently watches Kubernetes services and endpoints, streaming them to via Envoy xDS CDS & EDS endpoints.
+Requests to envoy are routed directly to Kubernetes pods (i.e. endpoints).
```
davecheney: Possibly worth adding;

Kubernetes pods are identified by their presence in the Endpoint document and streamed to Envoy via EDS.

design/external-names.md
```diff
@@ -0,0 +1,47 @@
...
+This proposal looks to implement Kubernetes service type `ExternalName` which allows for an `externalName` which is represented as a valid RFC-1123 hostname.
```
davecheney: Possibly;

This proposal looks to add a second way to identify the members of a Envoy Cluster object, using a DNS entry to supply the cluster members rather than EDS.

design/external-names.md
```diff
@@ -0,0 +1,47 @@
...
+
+## High-level Design
+
+Contour will need to watch for and process Kubernetes service types of `ExternalName` (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#servicespec-v1-core).
+When configuring EDS endpoints, for this service type Contour will pass the contents of the `externalName` field on the Kubernetes service to EDS instead of a PodIP.
```
davecheney: I'm not sure this is correct. Does the endpoint controller create an Endpoint document for services that use ExternalName? If there is no endpoint there will be no entry in EDS.

What we'll do instead is record the DNS name as a new kind of dag object (see below for a suggestion) and the cluster visitor will configure the cluster discovery type to use DNS (https://www.envoyproxy.io/docs/envoy/v1.9.0/api-v2/api/v2/cds.proto#envoy-api-enum-cluster-discoverytype) rather than EDS.

design/external-names.md
```diff
@@ -0,0 +1,47 @@
...
+
+Additionally, users will need to configure ports on the service so Contour will understand how to proxy requests to the external endpoint.
```
davecheney: we're not really changing this; users still need to supply a Port variable to identify the DNS:PORT pair of the external service.

design/external-names.md
davecheney: Please also add we're only doing this for HTTP at the moment. We can do TCP later on if people ask.

design/external-names.md
```diff
@@ -0,0 +1,47 @@
...
+
+Potentially, we could avoid using a Service of type `ExternalName` by simply just extending the `IngressRoute` CRD as proposed here: https://github.com/heptio/contour/issues/825#issue-386869426
+
+The downside to this approach is it couldn't be used with Ingress resources.
+
+## Security Considerations
+
+Potentially someone could use an `ExternalName` service to reference services in another namespace by setting the `externalName` (ex: `other-service.namespace-other.svc.cluster.local`).
+
+This scenario exists in a cluster regardless if Contour is deployed or not, other security restrictions should be applied if users want to avoid this from happening.
```
davecheney: Agreed, namespace network isolation is outside the scope of contour.

davecheney:
LGTM. Feel free to merge when you're happy with the revisions.

design/external-names.md
```diff
@@ -0,0 +1,47 @@
...
+- The `TCPService` struct in `internal/dag/dag.go` will have a new `externalName string` field added.
+- When services are added in `internal/dag/builder.go`, if `svc.Spec.Type == v1.ServiceTypeExternalName` then the value of `svc.Spec.ExternalName` will be set to the TCPService struct's `externalName`.
+- Since the service does not have any Kubernetes Endpoints, we cannot rely on having an EDS cluster to reference from the CDS cluster.
+For this type of service (i.e. `ExternalName`) when creating the cluster in `internal/envoy/cluster.go`, instead of setting a `EdsClusterConfig`, we'll configure a `LoadAssignment *ClusterLoadAssignment` which should allow us to specify the endpoint defined in `externalName` dynamically.
+
+## Alternatives
+
+Potentially, we could avoid using a Service of type `ExternalName` by simply just extending the `IngressRoute` CRD as proposed here: https://github.com/heptio/contour/issues/825#issue-386869426
+
+The downside to this approach is it couldn't be used with Ingress resources.
```
mattmoyer: Using a Service also enables using existing annotations like `contour.heptio.com/upstream-protocol.h2` to control the upstream connection parameters the same way you would in the usual case, which feels like a big win.

design/external-names.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Security Considerations
```
mattmoyer: It might also be worth noting that users should probably pair this with the new upstream TLS validation feature (added in https://github.com/heptio/contour/pull/1045).

In a large percentage of deployments, this externalName upstream connection is more likely to be transiting untrusted networks.

design/external-names.md
davecheney: yeah, we keep all the logic that we've written to control Envoy's Cluster configuration (https://www.envoyproxy.io/docs/envoy/v1.9.0/api-v2/api/v2/cds.proto#cluster) and reuse the k8s service object, this proposal boils down to configuring the cluster with a static configuration -- a DNS entry -- rather than pointing back to Contour for an EDS record.

design/external-names.md
davecheney: Yeah, this will give users the opportunity to have TLS between clusters.

design/external-names.md
stevesloka: Oh good catch, when I started this, in my head I was going to send a dns entry to EDS instead of a IP, but later found the other way which I referenced later. Let me update.

design/external-names.md
stevesloka: I call it out because it's not required, I can create a service like this and k8s api is cool with it. I'll add a callout that we'll set the status on the IngressRoute object if a service is found but no ports.

```
apiVersion: v1
kind: Service
metadata:
  labels:
    run: nginx
  name: nginx
  namespace: default
spec:
  externalName: stevesloka.com
  type: ExternalName
status:
  loadBalancer: {}
  type: ExternalName
```

davecheney: good point. its probably worth stating it explicitly then -- the hosts come from `spec.externalName`, the Port comes from the `backend.servicePort` record.

We might also be able to use SRV records here, but that's out of scope for this proposal.

PR #1098, opened 2019-05-16
internal/envoy: move StatsListener generation to helper

davecheney:
Move the generation of the StatsListener to a helper inside
internal/envoy. This defines what the canonical definition of the stats
listener is, in one place, with a test. It also removes a lot of the
implementation plumbing from internal/contour and internal/e2e.

### DIFF
--- a/internal/contour/listener.go
+++ b/internal/contour/listener.go
@@ -16,14 +16,8 @@ package contour
 import (
 	"sync"
 
-	"github.com/gogo/protobuf/types"
-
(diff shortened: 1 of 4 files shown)


davecheney:
The goal for the `internal/envoy` package is it contains the knowledge for "make me a gRPC thing". It doesn't know why you want a stats listener, it doesn't much care either. By having this logic in helpers in their own package we test them in isolation of the business logic that calls it. As well as having the logic for helpers in one place, it also means we can use those helpers in both code and tests elsewhere because we assert that they work as specified in the contract of the test.

PR #1100, opened 2019-05-17
Set idle timeout for TCP proxy connections

### LINKED ISSUE #1074: tcp proxy connections need idle timeout setting
#1070 set the idle timeout for http connections to 60 seconds. We should do the same for TCP connections. The timeout should probably be longer than 60 seconds as the semantics for tcp connections are different from the keep alive behaviour of an idle http connection.

youngnick:
Fixes #1074.

Sets the idle timeout for TCP proxy connections to two and a half hours.

Also includes some small refactoring to make it so that changing the magic
numbers for the TCP timeout and the HTTP timeout is only required in one place,
tests included.


internal/envoy/listener.go
```diff
@@ -28,6 +28,18 @@ import (
 	"github.com/heptio/contour/internal/dag"
 )
 
+// HTTPDefaultIdleTimeout sets the idle timeout for HTTP connections
+// here and in the e2e tests.
+const HTTPDefaultIdleTimeout = 60
```
davecheney: Rather than HTTPDefaultIdleTimeout * time.Second lower down, why not multiply this constant by time.Second on this line then it is clear this is 60seconds, not 60 something’s which have the be scaled at the place they are used.

internal/envoy/listener.go
```diff
@@ -28,6 +28,18 @@ import (
...
+
+// TCPDefaultIdleTimeout sets the idle timeout in seconds for
+// connections through a TCP Proxy type filter.
+// It's defaulted to two and a half hours for reasons documented at
+// https://github.com/heptio/contour/issues/1074
+// Set to 9001 because now it's OVER NINE THOUSAND.
+// Exported so the same value can be used here and in e2e tests.
+const TCPDefaultIdleTimeout = 9001
```
davecheney: Same as above

internal/envoy/listener.go
```diff
@@ -28,6 +28,18 @@ import (
 	"github.com/heptio/contour/internal/dag"
 )
```
davecheney: Rather than documenting where this is used or what it used for, comment on _why_ 60 seconds is the number that was chosen. 

Documenting how a public constant can be used is not possible because anyone can import and use that value without you even knowing.

internal/envoy/listener.go
youngnick: That makes sense. Will do.

stevesloka:
@youngnick just needs a rebase. =)

PR #1101, opened 2019-05-17
Fix small linting errors.

youngnick:

### DIFF
--- a/internal/envoy/cluster.go
+++ b/internal/envoy/cluster.go
@@ -21,7 +21,7 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/envoy/cluster.go
```diff
@@ -261,6 +261,8 @@ func u32nil(val int) *types.UInt32Value {
 	}
 }
 
+// ClusterCommonLBConfig is a helper to return Cluster_CommonLbConfig
```
davecheney: // ClusterCommobLBConfig creates a *v2.Cluster_CommonLbConfig with HealthyPanicThreshold disables.

PR #1102, opened 2019-05-18
internal/envoy: Support services type externalName

### LINKED ISSUE #334: [feature] support Services with ExternalName or ExternalIP components

stevesloka:
Fixes #334 by allowing routes to proxy traffic to DNS names specified in the service.spec.externalName field of a service. When defined, this skips EDS and specifies the dns endpoint directly on the cluster.

### DIFF
--- a/internal/envoy/cluster.go
+++ b/internal/envoy/cluster.go
@@ -24,6 +24,7 @@ import (
 	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
 	envoy_cluster "github.com/envoyproxy/go-control-plane/envoy/api/v2/cluster"
 	"github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
+	endpoint "github.com/envoyproxy/go-control-plane/envoy/api/v2/endpoint"
 	envoy_type "github.com/envoyproxy/go-control-plane/envoy/type"
 	"github.com/gogo/protobuf/types"
 	"github.com/heptio/contour/internal/dag"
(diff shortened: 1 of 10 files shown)


internal/envoy/cluster.go
```diff
@@ -106,6 +114,36 @@ func cluster(cluster *dag.Cluster, service *dag.TCPService) *v2.Cluster {
 	return c
 }
 
+func ClusterLoadAssignConfig(cluster string, service *dag.TCPService) *v2.ClusterLoadAssignment {
```
davecheney: Please call this StaticClusterLoadAssignment. It is a bit of a mouthful, but see the suggested comment.

davecheney: // StaticClusterLoadAssignment creates a *v2.ClusterLoadAssignment pointing to the external DNS address of the service.

davecheney: the `cluster` parameter appears unused

internal/envoy/cluster.go
```diff
@@ -78,14 +79,21 @@ func upstreamValidationSubjectAltName(c *dag.Cluster) string {
...
+		ConnectTimeout: 250 * time.Millisecond,
+		LbPolicy:       lbPolicy(service.LoadBalancerStrategy),
+		CommonLbConfig: ClusterCommonLBConfig(),
+		HealthChecks:   edshealthcheck(service),
+	}
+
+	// If routing to service type `ExternalName` then specify upstream here and not via EDS
+	if len(service.ExternalName) > 0 {
+		c.LoadAssignment = ClusterLoadAssignConfig("contour", service)
+		c.ClusterDiscoveryType = ClusterDiscoveryType(v2.Cluster_STRICT_DNS)
```
davecheney: this is a tiny nit, but as this alternation does the same thing, just with different values, i think it would be clearer if the assignments to c.LoadAssignment and c.ClusterDiscoveryType were in the same order.

Maybe consider using a switch, rather than an if, again to make it clearer that this is a selection, not a conditional. eg

```go
switch len(service.ExternalName) {
case 0:
   // external name not set, cluster will be discovered via EDS
   c.ClusterDiscoveryType = ClusterDiscoveryType(v2.Cluster_EDS)
   c.EdsClusterConfig = edsconfig("contour", service)
default:
  // external name set, use hard coded DNS name
  c.ClusterDiscoveryType = ClusterDiscoveryType(v2.Cluster_STRICT_DNS)
  c.LoadAssignment = ClusterLoadAssignConfig("contour", service)
}
```

stevesloka:
Thanks for the comments @davecheney, please take a look again, I've addressed them.

internal/envoy/cluster.go
stevesloka: Great point, it's not a this or that scenario, so this is more clear.

internal/envoy/cluster.go
```diff
@@ -78,14 +79,23 @@ func upstreamValidationSubjectAltName(c *dag.Cluster) string {
...
+
+	switch len(service.ExternalName) {
+	case 0:
+		// external name not set, cluster will be discovered via EDS
+		c.ClusterDiscoveryType = ClusterDiscoveryType(v2.Cluster_EDS)
+		c.EdsClusterConfig = edsconfig("contour", service)
+	default:
+		// external name set, use hard coded DNS name
+		c.ClusterDiscoveryType = ClusterDiscoveryType(v2.Cluster_STRICT_DNS)
+		c.LoadAssignment = StaticClusterLoadAssignment(service)
```
davecheney: nice!

PR #1103, opened 2019-05-18
Adding jekyll site | Adding netlify.toml

lspringer:
Signed-off-by: Lee Springer <lee@smalltalk.agency>


davecheney:
Hello,

I wanted to give an update to all the participants. I've marked this PR for review in the 0.14 cycle because it will not be merged before we release 0.13 on the 20th.

In marking this PR for review in 0.14 it does not mean that this PR will be merged, but it is a placeholder for the following conversations:

- The body of this PR contains many unlinked files which were perhaps part of the older github pages version of the site. This should be removed.
- No analytics or site tracking present.
- Content is based on 0.11, which is two releases behind.
- A workflow for how edits to the site will occur. I've stated that I _do_ want the `site/` directory to live within this repository--that is preferable to documentation being maintained in a second repository--but we cannot have a PR workflow that leads to a publicly accessible version of the site being published by netlify. The way forward I see here is to only push to netlify on merge to master. This may result in a few "whoops, fixing site layout" PRs every now and then, but that is preferable to content which we do not control being published to a URL that _looks_ like projectcontour.io.

/cc @VMmore @timh @jonasrosland

jonasrosland:
Hi Dave!

Thank you for the update, and I'm looking forward to working through the changes that needs to be made.

Some questions and answers:

> The body of this PR contains many unlinked files which were perhaps part of the older github pages version of the site. This should be removed.
The content here was was copied over from the old docs directory, so the table of contents need to be updated and links to old content removed.

> No analytics or site tracking present.
Do you want analytics present, or do you want to make sure they are not there?

> Content is based on 0.11, which is two releases behind.
The PR was based on 0.11, and thanks to the fast release cycle of Contour it is now behind :)

>The way forward I see here is to only push to netlify on merge to master. 
Yup, as we've discussed previously, this is already set up. You won't have automatic builds of pull requests on the contour repository, only when merged to master.

I am also working on better documentation regarding how to run this locally and contribute back to the website. That could then live in the docs under a "Contributing" segment.

davecheney:
Closed by #1198

PR #1104, opened 2019-05-20
Makefile: promote unparam to check target

davecheney:
Promote mvdan's unparam to run as part of the make check target. This
would have spotted a condition in a recent PR where a paremeter was
declared in the signture but wasn't used in the body of the function.

Skip the api/ directory as the generated files aren't lint clean -- but
we don't really care because a computer wrote them for us.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ test-race: | test
 vet: | test
 	go vet ./...
 
-check: test test-race vet gofmt staticcheck misspell unconvert ineffassign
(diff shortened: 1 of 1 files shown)


davecheney:
> LGTM. Like the use of $(PKGS).

Some tools can use package names, others take directories, yet more take files. Hopefully tools like golangci-lint can help, but they came along after I set down these foundations so aren't the biggest dumpster fire atm

Issue #947, opened 2019-03-15
Header based routing


pbarker:
Expose header based routing https://www.envoyproxy.io/docs/envoy/v1.5.0/api-v1/route_config/route#headers

davecheney:
@pbarker thanks for raising this issue. Could you describe a little more about what you need this for. Said another way, if Contour let you route on a header, rather than the URL path, what would that enable you to do?

pbarker:
So, our specific use case is wanting to route to 'feature pods' based on a header. This would entail deploying a downstream service with a new feature and wanting to run tests on it before generally routing traffic to it. Not sure if thats in the scope for contour

stevesloka:
@pbarker so you would want something like this? Say you have 2 services in your cluster named `service-foo` and `service-bar`. 

Requests to `example.com` with no header would route to `service-foo` and requests to `example.com1 with a header of `X-Contour-App: bar` would route to `service-bar`?

I think this could be doable by adding a header to the DAG. Right now Contour builds a dag based upon the fqdn + path (e.g. example.com + /). We could take into account a header as well which would make the three unique. Any delegate of the root IngressRoute that has a header match would also need to match a header (if applicable).

pbarker:
awesome, love a good DAG. I'm also wondering on a general level whether contour would consider supporting a plugin ecosystem. Has that idea been put out there?

davecheney:
>  I'm also wondering on a general level whether contour would consider supporting a plugin ecosystem. Has that idea been put out there?

I'd like to avoid plugins as long as possible.

rosskukulinski:
But that's not to say plugins are off the table, especially if there are additional engineering and/or PM resources as well as sufficient user-story & design docs?  Thoughts @davecheney?

davecheney:
I'm sorry to be so firm, but Go plugins are weakly supported (linux yes, macos, sorta, windows, nope, lets not even talks about non intel platforms) and have incredibly tight coupling because the plugin and the program its being plugged into are built from an identical toolchain. The developer who originally added plugin support to Go (for the Google Fuschia project) left Google and not one jot of engineering effort has been spent on plugins since. I can't use plugins as a basis for Contour, it's too risky.

davecheney:
With that said, I'm not saying no to header based routing. What has to happen is we have to write up the proposal as a design document, once we have a design we'll have a guide how much work it will take and can figure out how to fit it in the schedule.

[...]

PR #1111, opened 2019-05-21
wip: design: Header routing design doc

stevesloka:
Updates #947 by adding a Header based routing design document. 

### DIFF
--- a/design/header-routing.md
+++ b/design/header-routing.md
@@ -1,11 +1,18 @@
-# Header Based Routing
+# v1.IngressRoute Design
 
 _Status_: Draft
 
-The goal of this feature is to allow Contour to route traffic by more than just fqdn or path matching, but by also routing based upon a header which exists on the request.
-The scope of this design doc looks to improve on the IngressRoute.v1beta1 design and incorporate changes to enable not just path based routing, but also header based routing as a first class citizen.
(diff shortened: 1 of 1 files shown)


stevesloka:
I have a question on if Contour should be able to route on **only** a Header or does it require a path to go with it. This design requires a path which I've outlined in the design.

design/header-routing.md
```diff
@@ -0,0 +1,71 @@
+# Hewader Based Routing
```
pbarker: sp

design/header-routing.md
```diff
@@ -0,0 +1,70 @@
...
+The IngressRoute spec will be updated to allow for a `header` field to be added which will allow for a set of key/value pairs to be applied to the route.
+In the following example, you can define for path `/foo` and send traffic to different services for the same path but which contain different headers.
+
+```
+spec:
+  routes:
+  - match: /foo
+      header:
+        name: "x-header"
+        value: "A"
```
davecheney: `value: "A"` here looks like an exact match. What about if I wanted to match on a prefix, or substring -- ie, the header value must contain a certain string.

design/header-routing.md
```diff
@@ -0,0 +1,70 @@
...
+    services:
+    - name: backend-A
+        port: 9999
+  - match: /foo
```
davecheney: What about the use case where I only want to match on header only? Does it make sense to always have to combine this with a path?

design/header-routing.md
stevesloka: Are you suggesting something like "A*" or "*A*" or even "AAAA*AAAA"?

stevesloka:
Closing this because the IngressRoutev1 design doc implements.

PR #1120, opened 2019-05-23
Update `make local` target to work with local `contour serve`

### LINKED ISSUE #1114: Local dev loop improvements
Currently we have a `make local` target that runs a local Envoy that will generate a Contour bootstrap file, and then pass it to Envoy. However, you already need to be running a `contour serve` on your machine. This issue is to figure out how to do something a bit more straightfoward, which should make testing things locally a bit easier.

This is a step towards solving #900, but won't completely solve it.

Must:
- start a Contour and Envoy container locally somehow, with the Envoy set up to talk to the Contour.
- Have some way to add new configuration to Contour. (Could be 'just use out-of-cluster').

May:
- make updating the running version of contour easy.

youngnick:
Fixes #1114 

Turns out that the easiest way to do this is to pass a host IP to the `contour bootstrap` invocation inside the `make local` target. Obviously this means that `contour serve` must be listening on that IP, which means that it should either be run with `contour serve --xds-address=0.0.0.0` or with a local IP on the host in place of 0.0.0.0. (Listening on all interfaces carries obvious security concerns).

I'm not sure where the best place to document this process is, willing to be educated on that front.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ vendor/
 .idea/
 .DS_Store
 go.sum
+localenvoyconfig.yaml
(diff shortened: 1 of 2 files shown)


stevesloka:
Also, might be nice to add `localenvoyconfig.yaml` to the `.gitignore`

PR #1131, opened 2019-05-29
Add flag to set Contour log level and reduce default xds logs verbosity

rochacon:
Contour outputs debug logs from `internal/grpc/xds.go:xdsHandler.stream` main loop by default, which may be quite verbose (in a small cluster I've seen around 400 log lines each couple of seconds).

In an attempt to reduce this verbosity by default, this PR adds a new flag to Contour: `--log-level`, which defaults to `info`, and sets the `internal/grpc/xds.go:xdsHandler.stream` main loop logs to `log.Debug`, reducing the amount of logs for each loop. The old behavior can be restored by setting `--log-level=debug` on Contour startup.
I've also fixed a typo in the comments related to the before mentioned log section.

### DIFF
--- a/cmd/contour/contour.go
+++ b/cmd/contour/contour.go
@@ -48,6 +48,7 @@ var ingressrouteRootNamespaceFlag string
 func main() {
 	log := logrus.StandardLogger()
 	app := kingpin.New("contour", "Heptio Contour Kubernetes ingress controller.")
+	logLevel := app.Flag("log-level", "Contour log level.").Default("info").String()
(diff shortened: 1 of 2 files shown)


davecheney:
Thank you for your PR. 

We use a talk then code philosophy on this project. Please raise an issue and we’ll talk about the problem before jumping to a solution.

rochacon:
Oops, sorry about that @davecheney. I'm closing this for now, the discussion issue is linked above.

Issue #283, opened 2018-03-13
Support health checking of Contour


cten:
Most load balancers have the feature of checking whether an endpoint is healthy or not. In HAProxy a common way to do this is use the HEAD method and expect a specific return code. 
Can a health endpoint be added to Contour, whether its a /healthz path or just supporting the HEAD method on the base path doesn't really matter to me. 
Thanks

stevesloka:
Sorry, misread this issue. My previous comment was health checking endpoints, not Contour itself.

stevesloka:
To clarify this Issue, the description is to health check Contour, there's a new issue opened to track health checking Envoy: #695

bpownow:
+1 on this issue. Thanks @stevesloka!

davecheney:
Contour has supported  a `/heath` endpoint on the debug listener, `0.0.0.0:8000` since release 0.6.

To close this issue

- [ ] Does `/health` need to be renamed to `/healthz`? 
- [ ] `/health`/`/healthz` should be woven through the example `deployments/`

cten:
That works for me, healthz might be more consistent with other apps but as
long as there is a health endpoint that's documented in good.
Thanks

stevesloka:
I do agree that `/healthz` is a little more consistent, but today the health check endpoint always returns `200 OK` (https://github.com/heptio/contour/blob/master/internal/metrics/metrics.go#L217-L222). 

I think we need to expand on this and wire in some additional checks to determine if Contour is actually healthy or not. For instance, what if Contour can't reach the k8s api server, etc?

PR #1132, opened 2019-05-29
Rename Contour health check endpoint for consistency, add probes to examples

rochacon:
This PR renames Contour health check endpoint from `/health` to `/healthz` to improve consistency with other components and adds liveness and readiness probes for this endpoint in the examples directory.

Related to #283

:warning: even though I've been using this endpoint in my environment, I've not tested all the examples manifests.

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -219,7 +219,7 @@ func (svc *Service) Start(stop <-chan struct{}) error {
 }
 
 func registerHealthCheck(mux *http.ServeMux, client *kubernetes.Clientset) {
-	mux.HandleFunc("/health", func(w http.ResponseWriter, r *http.Request) {
+	healthCheckHandler := func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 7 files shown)


internal/metrics/metrics.go
```diff
@@ -215,7 +215,7 @@ func (svc *Service) Start(stop <-chan struct{}) error {
 }
 
 func registerHealthCheck(mux *http.ServeMux) {
-	mux.HandleFunc("/health", func(w http.ResponseWriter, r *http.Request) {
+	mux.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
```
stevesloka: We should enable both `/healthz` and `/health` to give us a way to deprecate the old path.

davecheney: SGTM. Please add a TODO and point it to an issue to remove the old endpoint in 0.14 or 0.15

rochacon: Done

Issue #1118, opened 2019-05-22
Rename deployment/ to example/


davecheney:
Internally we always talk about the deployment/ directory as examples. They are samples designed to give the reader a starting point. 

Externally I think this is not as clear and suggests that the deployment/ directory is _the_ way to deploy contour. Sadly, given the current tooling available in the k8s ecosystem I don’t think it is possible to provide — let alone *unit test* — vetted, bulletproof, yaml manifests. 

The short term solution to this I think is to rename deployments/ to examples/.

stevesloka:
We'll also need to coordinate the jump shortcut links which point to the `/deployments` directory today (e.g. `https://j.hept.io/contour-deployment-rbac`).

davecheney:
@rochacon you're go to file a second PR that removes `deployment/`

PR #1133, opened 2019-05-29
Copy deployment/ -> examples/

rochacon:
Partially solves #1118, jump URLs need to be fixed after merge.


davecheney:
@rochacon ping

rochacon:
@davecheney done, restored the `deployment` directory. Should I open the removal PR now or you prefer to patch the links first?

davecheney:
Go for it, we won’t land til I can get the shortlinks changed over.

Issue #1125, opened 2019-05-28
[spike] Investigate if we can use :* syntax in domain matching


davecheney:
One of the oldest issues we have on this project is related to how Envoy handles `Host: ` headers which include a `:port` suffix. #210, #398, #353 and many more.

[A recent comment](https://github.com/envoyproxy/envoy/issues/886#issuecomment-492824085) suggests that we may be able to specify
```
domains: []string{"host", "host:*"}
```
instead of tracking and hard coding the port the listener connected to this routing is using, can we use the `:*` syntax instead?

This is a spike to see if this is possible, if it will work for both HTTP and HTTPS, and if so scope the work to remove completely or partially the mechanisms for configuring and tracking the listening port's number throughout Contour.

_nb_: The resolution of this ticket is either closing this issue with a comment if it is not possible to use this syntax. Bonus points for replying to the original commenter. If this is something we can use, please scope the work necessary in one or more new issues -- we may choose to do it over several releases, especially if deprecating flags are involved.

youngnick:
Okay, I have a branch that will change the VirtualHost generation so that it will always add the `hostname` and `hostname:*` to the `domains` part of the config, instead of using the given port.

Things I've been able to confirm:
- Envoy accepts the config
- Traffic routes correctly to bare hostname requests (ie Host header is `example.com`)
- Traffic routes correctly to hostnames with ports (ie Host header is `example.com:80` or `example.com:443`, or even `example.com:31273`)
- Kubernetes Services that are not available on whatever the value of `--envoy-service-http-port` or `--envoy-service-https-port` still work.

Changes that we'll need to make to actually make this a thing:
- Fix all the tests to expect `example.com:*` instead of `example.com:<port>`
- Remove the wiring that brings the ports down to the VIrtualHost. This will include deprecating some command line flags, like `--envoy-service-http-port` and `--envoy-service-https-port`, at least.
- remove `minikube.md` and update documentation with minikube and kind examples.
- Other stuff I don't know about.

davecheney:
Nice result. Make it so.

Issue #210, opened 2018-02-01
Non standard ingress ports break routing


davecheney:
This issue was reported on #contour on the k8s slack. In essence, when using minikube the service load balancer will allocate a random high port as the ingress. This is problematic because this port is not a _well known_ (RFC 2616 gibberish) port so `curl(1)` or browsers will include the port number in the `Host:` header. This causes Envoy to misroute the request because the `domains` file the RDS `virtualhost` entry does not contain the matching `:port` suffix.

The problem is the port minikube chooses is not easily predictable, so it is not simply a matter of including various permutations of `hostname:port` in the `virtualhost.domains` array.

Supporting Envoy ticket, envoyproxy/envoy#1269.

### Workarounds

The workaround if you are using minikube is to either force the `Host:` header with something like
```
curl -H "Host: example.com" -v http://example.com:31847
```
Or use a OSX specific extension to curl
```
curl -v --connect-to example.com:80:example.com:31847 http://example.com/
```

sevein:
As a workaround, I've also been using `socat` to set up a TCP proxy so I can access from my browser:

    $ socat TCP-LISTEN:80,fork TCP4:example.com:31847

This way the browser sends the `Host: example.com` header as expected by Envoy.
It should work as long as `example.com` resolves to `127.0.0.1`.

davecheney:
I'm going to close this issue as there is nothing more to be done on the contour side for the moment.

- minikube specific issues are documented per #353 
- Host headers with explicit :port suffixes are handled per #398

bitsofinfo:
https://github.com/envoyproxy/envoy/issues/886

youngnick:
The spike in #1125 has meant that we can actually do this now. Reopening.

youngnick:
Since the above PRs have been merged, it is now possible to use non-standard ingress ports as much as you like - any port in the Host header will now match.

This means that Minikube and kind will now work just fine for testing and development. Win!

Issue #1135, opened 2019-05-30
dag/builder: Clean up magic constant use to distinguish between secure and insecure listeners


youngnick:
Currently in the DAG (at https://github.com/heptio/contour/blob/a70c99a17d958a6645cadcabfcf4bd0b2794e6ef/internal/dag/builder.go#L260), we build two Listener structs, and store them in a map[int]*Listener (https://github.com/heptio/contour/blob/a70c99a17d958a6645cadcabfcf4bd0b2794e6ef/internal/dag/builder.go#L68).

The int key is only ever one of the two magic numbers 80 or 443 - 'insecure' or 'secure' respectively. Previously to #1125 , these were whatever the passed values of `envoy-external-http-port` and `envoy-external-https-port` were, but now they are no longer required they are being reverted to 80 and 443.

This ticket is to cover going back over the code and changing those numbers to be something more meaningful - perhaps making it a map[string]*Listener, or even a map[bool]*Listener, since either of those options are better than arbitrary numbers.

PR #1137, opened 2019-05-30
Remove ExternalPort functions, now Envoy provides new functionality.

youngnick:
Updates #1125 
Updates #210 

This update removes the functionality of the `--envoy-external-http-port` and `--envoy-external-https-port` flags, as we are now able to pass `www.example.com` and `www.example.com:*` to Envoy as allowed `Host:` headers, which makes them not required.

If you need to make your services available on a non-standard load balancer port (ie not 80 for http and 443 for https) upstream of Envoy, that no longer requires any configuration. As long as the traffic arrives at Envoy, it will be routed by the Host header, and will match any external port.

Note that this should also allow the easier use of minikube and kind for Contour testing, another PR will follow with documentation for both use cases.


cmd/contour/contour.go
```diff
@@ -140,6 +138,13 @@ func main() {
 	serve.Flag("ingress-class-name", "Contour IngressClass name").StringVar(&reh.IngressClass)
 	serve.Flag("ingressroute-root-namespaces", "Restrict contour to searching these namespaces for root ingress routes").StringVar(&ingressrouteRootNamespaceFlag)
 
+	// The following flags are no-ops, and the variables are used to print a message that they don't do anything
+	// any more.
+	var externalHTTPPort int
```
davecheney: Please add
```
// TODO(youngnick) remove these for 0.14, see #nnnn 
```
Where `nnnn` is an issue to remove these flags

cmd/contour/contour.go
```diff
@@ -163,6 +168,14 @@ func main() {
 		log.Infof("args: %v", args)
 		var g workgroup.Group
 
+		// Functionality warnings for non-functional flags.
+		if externalHTTPPort != 0 {
+			log.Infof("Due to changes to upstream Envoy, setting envoy-external-http-port to %d no longer does anything, your Ingress will work without it. This notice will be removed in two releases.", externalHTTPPort)
```
davecheney: shorter is better

log.Warn("--envoy-external-http-port will be removed in 0.14")

cmd/contour/contour.go
```diff
@@ -163,6 +168,14 @@ func main() {
...
+		}
+		if externalHTTPSPort != 0 {
+			log.Infof("Due to changes to upstream Envoy, setting envoy-external-https-port to %d no longer does anything, your Ingress will work without it. This notice will be removed in two releases.", externalHTTPSPort)
```
davecheney: same

cmd/contour/contour.go
```diff
@@ -163,6 +168,14 @@ func main() {
 		log.Infof("args: %v", args)
 		var g workgroup.Group
 
+		// Functionality warnings for non-functional flags.
```
davecheney: s/Functionality/Deprecation warning for deprecated flags.

internal/contour/route_test.go
```diff
@@ -237,7 +237,7 @@ func TestRouteVisit(t *testing.T) {
 					Name: "ingress_http",
 					VirtualHosts: []route.VirtualHost{{
 						Name:    "www.example.com",
-						Domains: []string{"www.example.com", "www.example.com:80"},
+						Domains: []string{"www.example.com", "www.example.com:*"},
```
davecheney: suggestion: the value of `Domains` can be determined fully from its input, try making a small helper to dry this up.
```go
func domains(hostname string) []string {
 return []string{hostname, hostname+":*"}
}
```

internal/dag/builder.go
```diff
@@ -39,17 +39,8 @@ const (
 type Builder struct {
 	KubernetesCache
 
-	// ExternalInsecurePort is the port that HTTP
-	// requests will arrive at the ELB or NAT that
-	// presents Envoy at the edge network.
-	// If not supplied, defaults to 80.
-	ExternalInsecurePort int
+	// TODO: Remove this layer of indirection?
```
davecheney: // TODO(youngnick) merge Builder and Kubernetes cache? See #nnnn

davecheney: The TODO doesn't mean you'll do it, it just means you're the person who knows best what that comment means.

internal/dag/builder_test.go
```diff
@@ -1664,9 +1664,7 @@ func TestDAGInsert(t *testing.T) {
 			),
 		},
 		"insert ingress w/ tls with different secure port": {
```
davecheney: Is this test testing anything anymore?

youngnick: No, and it was broken. Oops, thanks, removed.

internal/dag/builder.go
```diff
@@ -229,7 +220,7 @@ func (b *builder) lookupSecret(m meta, validate func(*v1.Secret) bool) *Secret {
 }
 
 func (b *builder) lookupVirtualHost(name string) *VirtualHost {
-	l := b.listener(b.externalInsecurePort())
+	l := b.listener(80)
```
stevesloka: Won't this break if we wanted to serve on a different port, say 8080 since the listener is now hardcoded to 80/443?

youngnick: I thought that too, but this doesn't actually configure the listener at all! It's just a magic number that chooses whether or not the listener in the dag is the secure or insecure one. The actual listening ports are determined purely by the Ingress object. That's why I opened #1135, to clean this up.

youngnick: I've tested this with Ingress documents that listen on random ports, and it all works just fine.

stevesloka: :+1:  :tada:

internal/dag/builder_test.go
```diff
@@ -3685,10 +3662,7 @@ func TestDAGIngressRouteUniqueFQDNs(t *testing.T) {
 
 	for name, tc := range tests {
 		t.Run(name, func(t *testing.T) {
-			b := Builder{
-				ExternalInsecurePort: 10080,
-				ExternalSecurePort:   10443,
-			}
+			b := Builder{}
```
davecheney: nit, you can write 
```go
var b Builder
```
This is perhaps a matter of taste but I use the `var` form to indicate to the reader that we're declaring a `Builder` without explicitly initalising it. This leaves the `Builder{ ... }` form for when explicit initialisation of `Builder`'s fields is needed.

internal/dag/builder_test.go
```diff
@@ -1663,29 +1663,6 @@ func TestDAGInsert(t *testing.T) {
 				},
 			),
 		},
-		"insert ingress w/ tls with different secure port": {
```
davecheney: Around line 1375 you'll find
```go
        tests := map[string]struct {
                *Builder
                objs []interface{}
                want []Vertex
        }{
```
now there is no need to provide a custom `*Builder` as part of the fixture -- as there is nothing to customise -- this field can probably be removed and the logic on like 3688 cleaned up.

internal/e2e/rds_test.go
```diff
@@ -309,7 +309,7 @@ func TestEditIngressInPlace(t *testing.T) {
 				Name: "ingress_http",
 				VirtualHosts: []route.VirtualHost{{
 					Name:    "hello.example.com",
-					Domains: []string{"hello.example.com", "hello.example.com:80"},
+					Domains: []string{"hello.example.com", "hello.example.com:*"},
```
davecheney: perhaps copy the `domains` helper from `contour/route_test.go` and clean this up a little.

internal/dag/builder_test.go
youngnick: I read this in one of your blogs, ( 😄 ) will do.

internal/dag/builder_test.go
youngnick: Removed the `*Builder` and just made a new Builder in the actual execution loop down below.

internal/e2e/rds_test.go
youngnick: Good idea, done.

PR #1143, opened 2019-06-03
Update docs with details about minikube and kind use.

### LINKED ISSUE #210: Non standard ingress ports break routing

### LINKED ISSUE #1125: [spike] Investigate if we can use :* syntax in domain matching

youngnick:
Closes #1125 
Updates #210 
(doesn't close #210 so I can put a proper summary in there before closing it).

Now that the code changes for #1125 are merged, this PR updates the docs with instructions for testing or developing Contour with minikube and kind.


docs/cert-manager.md
```diff
@@ -16,7 +16,7 @@ This tutorial shows you how to securely deploy an HTTPS web application on a Kub
 - A DNS domain that you control, where you host your web application
 - Administrator permissions for all deployment steps
 
-**NOTE:** A Minikube cluster is not supported because of the complexities of NAT and port forwarding.
+**NOTE:** To use a local cluster like `minikube` or `kind`, see the instructions in [the deployment docs][7] for getting access.
```
davecheney: S/for getting access//

Issue #1005, opened 2019-04-10
Session affinity support


timh:
part of epic #361

PR #1152, opened 2019-06-04
all: remove Maglev and RingHash as supported load balancer strategies

### LINKED ISSUE #1030: LBStrategy has no effect in contour v0.11.0
0. create a service with 2 endpoints in k8s.
1. create an `IngressRoute` resource and set the `strategy` field of its services to `RingHash`.
2. access the domain of `IngressRoute`, and the two endpoints(pods) are chosen randomly.
3. dump the configuration of envoy, `lb_policy` field is set correctly for the `cluster`, but no `hash_policy` is set in the corresponding  `RouteAction`. So that no hash can be computed and backends are chosen randomly, thus the `RingHash` strategy takes no effect.
Also, `Maglev` strategy takes no effect for the same reason
some `hash_policy` in envoy is set by default or by some field in `spec` of `IngressRoute` resource, so that the hash strategy can take effect. The `hash_policy` should support the policies supported in envoy

**Environment:**

- Contour version: v0.11.0
- Kubernetes version: (use `kubectl version`): v1.11.6
- Kubernetes installer & version: kubeadm v1.11
- Cloud provider or hardware configuration: bare metal, Intel Xeon CPU E5-2609
- OS (e.g. from `/etc/os-release`): Ubuntu 16.04 LTS, kernel 4.4.0.116

### LINKED ISSUE #1150: ingressroute: remove maglev and ringhash from the list of supported policies
ingressroute supports a number of load balacing strategies. Among then, ringhash and maglev are supported -- you can set that value -- however the value does nothing without additional configuration. 

As we don't provide any way to supply that configuration, and the main use for the additional hash value for maglev and ringhash would be for session affinity, #856, I propose that those values be removed from the set of supported configurations and replaced with a stick session support as a first class citizen.

There is no need for deprecation cycle, those values could never be used properly.

davecheney:
Fixes #1150
Fixes #1030
Updates #1005

RingHash and Maglev are two balancing / affinity strategies that can be
configured per Cluster, however they are not useful by themselves.
Without the ability to configure the hash key, which is usually some for
of session af session or cookie, these strategies are not useful and
cannot be used correctly.

As such, remove them from the list of selectable options. Support for
session affinity will be added later.

### DIFF
--- a/design/ingressroute-design.md
+++ b/design/ingressroute-design.md
@@ -159,8 +159,6 @@ The following list are the options available to choose from:
 
 - **RoundRobin:** Each healthy upstream host is selected in round robin order
 - **WeightedLeastRequest:** The least request load balancer uses an O(1) algorithm which selects two random healthy hosts and picks the host which has fewer active requests. _Note: This algorithm is simple and sufficient for load testing. It should not be used where true weighted least request behavior is desired_
-- **Ring hash:** The ring/modulo hash load balancer implements consistent hashing to upstream hosts
-- **Maglev:** The Maglev load balancer implements consistent hashing to upstream hosts
 - **Random:** The random load balancer selects a random healthy host
 
 More documentation on Envoy's lb support can be found here: [https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/load_balancing.html](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/load_balancing.html)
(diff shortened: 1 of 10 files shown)


design/ingressroute-design.md
```diff
@@ -159,8 +159,6 @@ The following list are the options available to choose from:
 
 - **RoundRobin:** Each healthy upstream host is selected in round robin order
 - **WeightedLeastRequest:** The least request load balancer uses an O(1) algorithm which selects two random healthy hosts and picks the host which has fewer active requests. _Note: This algorithm is simple and sufficient for load testing. It should not be used where true weighted least request behavior is desired_
-- **Ring hash:** The ring/modulo hash load balancer implements consistent hashing to upstream hosts
```
stevesloka: This is a bit non-related to this PR, but have a quick question:

Should the design doc be the goal of what we'd want, then the `docs` section describes what get's implemented?  Or do we only design what we'll implement?

davecheney: good point. I'll revert this.

PR #1153, opened 2019-06-04
internal/dag: move HealthCheck to dag.Cluster

### LINKED ISSUE #1145: internal/dag: move remaining fields from dag.Service to dag.Cluster
`dag.Cluster` was introduced in 0.12 to record properties of the _path_ from a route to a specific k8s Service object. This was done in part for Upstream Validation support.

This issue is a placeholder to track the work of moving the remaining fields from `dag.Service` to `dag.Cluster`

davecheney:
Fixes #1145

Move HealthCheck parameter to dag.Cluster.

### DIFF
--- a/internal/dag/dag.go
+++ b/internal/dag/dag.go
@@ -103,15 +103,6 @@ type UpstreamValidation struct {
 	SubjectName string
 }
 
-func (r *Route) addHTTPService(s *HTTPService, strategy string, weight int, uv *UpstreamValidation) {
-	r.Clusters = append(r.Clusters, &Cluster{
-		Upstream:             s,
-		LoadBalancerStrategy: strategy,
-		Weight:               weight,
-		UpstreamValidation:   uv,
-	})
-}
-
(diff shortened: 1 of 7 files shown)


internal/dag/dag.go
```diff
@@ -278,6 +265,8 @@ type Cluster struct {
 	// The load balancer type to use when picking a host in the cluster.
 	// See https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/cds.proto#envoy-api-enum-cluster-lbpolicy
 	LoadBalancerStrategy string
+
+	HealthCheck *ingressroutev1.HealthCheck
```
stevesloka: Should we also not tie the IngressRoute struct to the dag?

davecheney: Yes, I want to redo this type into something like HealthCheckPolicy. I do want to remove the depdendency on the ingressroute type, but this PR is just moving the field. I'll do the rest in a followup

stevesloka: ok sounds good

PR #1154, opened 2019-06-05
Add commit guidelines

### LINKED ISSUE #1136: Update CONTRIBUTING.md with team coding standards
Traditionally, contributions to the project have used a few informal standards around commit style, PR style, and so on. This issue is to document those and put them in CONTRIBUTING.md.

youngnick:
Fixes #1136 

Just a small change with some rules of thumb. I've tried to make it clear these are guidelines rather than firm rules, suggestions on wording welcomed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -89,8 +89,10 @@ with a colon as delimiter. For example 'docs:', 'internal/(packagename):', 'desi
 tend to squash before opening the PR, then have PR feedback as
 extra commits.
 - If master has moved on, you'll need to rebase before we can merge,
-so merging upstream master or rebasing from upstream before opening your PR will probably save you some time.
-- PRs *must* include a `Fixes #NNNN` or `Updates #NNNN` comment.
+so merging upstream master or rebasing from upstream before opening your
+PR will probably save you some time.
+- PRs *must* include a `Fixes #NNNN` or `Updates #NNNN` comment. Remember that
+`Fixes` will close the associated issue, and `Updates` will link the PR to it.
 
 #### Commit message template
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -78,24 +78,48 @@ If you plan to submit a pull request for anything more than a typo or obvious bu
 Depending on the size of the feature you may be expected to first write a design proposal.
 A proposal template is [available here](https://github.com/heptio/contour/tree/master/design/design-document-tmpl.md)
 
-### Rules of thumb for commits
-
-Generally, we try to have our commits stick to the following rules of thumb:
+### Commit message and PR guidelines
 
 - Have a short subject on the first line and a body. The body can be empty.
 - Use the imperative mood (ie "If applied, this commit will (subject)" should make sense).
 - There must be a DCO line ("Signed-off-by: David Cheney <cheneyd@vmware.com>"), see [DCO Sign Off](#dco-sign-off) below
-- Try to put a summary of the main area affected by the commit at the start, with a colon as delimiter. For example 'docs:', 'internal/(packagename):', 'design:' or something similar.
-
-We won't turn away commits for *not* being like this, this is what we tend to do right now.
-
-## Rules of thumb for PRs
-
-- Try to keep your number of commits low. Most of us tend to squash
-before opening the PR, then have PR feedback as extra commits.
+- Put a summary of the main area affected by the commit at the start,
+with a colon as delimiter. For example 'docs:', 'internal/(packagename):', 'design:' or something similar.
+- Try to keep your number of commits in a PR low. Generally we
+tend to squash before opening the PR, then have PR feedback as
+extra commits.
 - If master has moved on, you'll need to rebase before we can merge,
 so merging upstream master or rebasing from upstream before opening your PR will probably save you some time.
-- PR titles generally use similar rules of thumb for commit subjects.
+- PRs *must* include a `Fixes #NNNN` or `Updates #NNNN` comment.
+
+#### Commit message template
+
+```
+<packagename>: <imperative mood short description>
+
+Updates #NNNN
+Fixes #MMMM
+
+Signed-off-by: Your Name you@youremail.com
+
+<longer change description/justification>
+
+```
+
+#### Sample commit message
+
+```
+internal\contour: Add quux functions
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -78,6 +78,25 @@ If you plan to submit a pull request for anything more than a typo or obvious bu
 Depending on the size of the feature you may be expected to first write a design proposal.
 A proposal template is [available here](https://github.com/heptio/contour/tree/master/design/design-document-tmpl.md)
 
+### Rules of thumb for commits
```
davecheney: s/ Rules of thumb for commits/Commit Message Guidelines

I'd like to avoid _rule of thumb_ as (as I found out recently) in some jurisdictions it doesn't mean what I think you meant. https://en.wikipedia.org/wiki/Rule_of_thumb. I know you didn't mean any harm by this, but we should probably avoid this term.

CONTRIBUTING.md
```diff
@@ -78,6 +78,25 @@ If you plan to submit a pull request for anything more than a typo or obvious bu
...
+
+Generally, we try to have our commits stick to the following rules of thumb:
```
davecheney: delete this line

CONTRIBUTING.md
```diff
@@ -78,6 +78,25 @@ If you plan to submit a pull request for anything more than a typo or obvious bu
...
+
+- Have a short subject on the first line and a body. The body can be empty.
+- Use the imperative mood (ie "If applied, this commit will (subject)" should make sense).
+- There must be a DCO line ("Signed-off-by: David Cheney <cheneyd@vmware.com>"), see [DCO Sign Off](#dco-sign-off) below
+- Try to put a summary of the main area affected by the commit at the start, with a colon as delimiter. For example 'docs:', 'internal/(packagename):', 'design:' or something similar.
+
+We won't turn away commits for *not* being like this, this is what we tend to do right now.
```
davecheney: Delete, we dont' need to write down a rule that says we won't enforce the rule we just wrote down. This is a guide, that's enough.

CONTRIBUTING.md
```diff
@@ -78,6 +78,25 @@ If you plan to submit a pull request for anything more than a typo or obvious bu
 Depending on the size of the feature you may be expected to first write a design proposal.
 A proposal template is [available here](https://github.com/heptio/contour/tree/master/design/design-document-tmpl.md)
 
+### Rules of thumb for commits
+
+Generally, we try to have our commits stick to the following rules of thumb:
+
+- Have a short subject on the first line and a body. The body can be empty.
```
davecheney: I'd like to see something that can be copy pasted as a template

Fixes (or Updates): #NNNN


Signed-off-by: Your Name <you@youremail.com>
```
packagename: what will happen when


```

Then bullet points explaining what we meant

- The subject line should explain what will happen if this change is applied. Use the imperative mood (ie "If applied, this commit will (subject)" should make sense). The subject line should _not_ explain _how_ the change does what it does. For instance "fixes typo" is better than "change spelling of misspelled word".
- The body of the commit message should describe _why_ the change is needed and give background if necessary. The body should not spend time describing _how_ the change does what it does, that's already evident from the diff of the change.

CONTRIBUTING.md
```diff
@@ -78,6 +78,25 @@ If you plan to submit a pull request for anything more than a typo or obvious bu
...
+- Use the imperative mood (ie "If applied, this commit will (subject)" should make sense).
+- There must be a DCO line ("Signed-off-by: David Cheney <cheneyd@vmware.com>"), see [DCO Sign Off](#dco-sign-off) below
+- Try to put a summary of the main area affected by the commit at the start, with a colon as delimiter. For example 'docs:', 'internal/(packagename):', 'design:' or something similar.
+
+We won't turn away commits for *not* being like this, this is what we tend to do right now.
+
+## Rules of thumb for PRs
```
davecheney: I'd like to see PR and commits merged. From the github PoV if you have a branch with one commit the commit message becomes the PR. If you have a branch with a lot of commits, then I don't care about the various commits, they're probably junk like "wip", and "fixed again", so I'll ask that you squash them. The important part of the guide is the PR.

CONTRIBUTING.md
```diff
@@ -78,6 +78,25 @@ If you plan to submit a pull request for anything more than a typo or obvious bu
...
+
+- Try to keep your number of commits low. Most of us tend to squash
+before opening the PR, then have PR feedback as extra commits.
+- If master has moved on, you'll need to rebase before we can merge,
```
davecheney: PR's _must_ include a `Fixes #NNNN` or `Updates #NNNN` comment.

CONTRIBUTING.md
youngnick: whoa, okay, I had no idea about that history. Will avoid in the future, thanks.

CONTRIBUTING.md
youngnick: Added a template and a sample.

CONTRIBUTING.md
davecheney: Yeah, it was a surprise to me as well. I figure it’s better to be safe than sorry.

CONTRIBUTING.md
```diff
@@ -78,6 +78,49 @@ If you plan to submit a pull request for anything more than a typo or obvious bu
...
+- Use the imperative mood (ie "If applied, this commit will (subject)" should make sense).
+- There must be a DCO line ("Signed-off-by: David Cheney <cheneyd@vmware.com>"), see [DCO Sign Off](#dco-sign-off) below
+- Put a summary of the main area affected by the commit at the start,
+with a colon as delimiter. For example 'docs:', 'internal/(packagename):', 'design:' or something similar.
+- Try to keep your number of commits in a PR low. Generally we
+tend to squash before opening the PR, then have PR feedback as
+extra commits.
+- If master has moved on, you'll need to rebase before we can merge,
+so merging upstream master or rebasing from upstream before opening your PR will probably save you some time.
+- PRs *must* include a `Fixes #NNNN` or `Updates #NNNN` comment.
```
stevesloka: This is a super small nit, but might be nice to put a quick explanation of when you would use `fix` vs `update`. Update won't close an issue.

youngnick: Good point, will do.

PR #1156, opened 2019-06-06
examples: simplify validation for spec.virtualhost.fqdn

### LINKED ISSUE #1117: contour.heptio-contour invalid fqdn
Attempted to use contour for internal traffic by specifying an fqdn of contour.heptio-contour: {"fqdn":"contour.heptio-contour"}

This results in an error:
validation failure list:
spec.virtualhost.fqdn in body should match '^([a-zA-Z0-9]+(-[a-zA-Z0-9]+)*\.)+[a-z0-9]{2,}$'

I expected this to pass because it is a valid name.

I think the validation test is overly conservative and should allow a hyphen in the last element to allow this.  

I've worked around this by using the cluster IP, but this has to be set per deployment which is not too nice.  

**Environment:**

- Contour version:

davecheney:
Fixes #1117

We've tried a few times to find a regex that fits the domain names that
people want to use but there are always edge cases.

Give up and simply require that the fqdn field is required. Ultimately
if the string isn't a valid DNS name, nothing will be able to route to
it, so the cost of writing nonsense here is low.

### DIFF
--- a/examples/common/common.yaml
+++ b/examples/common/common.yaml
@@ -49,10 +49,11 @@ spec:
         spec:
           properties:
             virtualhost:
+              required:      
+                - fqdn
               properties:
                 fqdn:
                   type: string
-                  pattern: ^([a-zA-Z0-9]+(-[a-zA-Z0-9]+)*\.)+[a-z0-9]{2,}$
                 tls:
                   properties:
                     secretName:
--- a/examples/render/daemonset-rbac.yaml
+++ b/examples/render/daemonset-rbac.yaml
@@ -52,10 +52,11 @@ spec:
         spec:
           properties:
             virtualhost:
+              required:      
+                - fqdn
               properties:
                 fqdn:
                   type: string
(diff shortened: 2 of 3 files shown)


examples/common/common.yaml
```diff
@@ -49,10 +49,11 @@ spec:
         spec:
           properties:
             virtualhost:
+              required:      
+                - fqdn
```
stevesloka: I think making this required breaks any non-root ingressroutes.

davecheney: Thanks for spotting that. Maybe we just drop the pattern all together. At best all I think we can say about this field is, if it is present, it shouldn't be blank. Which is basically saying nothing.

stevesloka: Turns out when you remove the regex, you can make an ingressroute fqdn like: `*.containersteve.com` which works because Envoy accepts wildcards, but this could break delegation.

davecheney:
Abandoned. This isn’t the right approach.

PR #1157, opened 2019-06-06
design: add session affinity draft

### LINKED ISSUE #1005: Session affinity support

davecheney:
Fixes #1005

### DIFF
--- a/design/session-affinity.md
+++ b/design/session-affinity.md
@@ -1,0 +1,157 @@
+# Session Affinity
+
+Status: _Draft_
+
+Session Affinity, sometimes called _Sticky Sessions_, is a pattern whereby the incoming request indicates the backend server it wishes to handle the request.
+Sticky sessions are very much not a cloud native pattern, however when migrating applications to Kubernetes it may be required to support session affinity if the application requires them.
+
+This proposal describes how session affinity support will be added to Contour.
+
+## Goals
+
+- Support session affinity via a session cookie managed transparently by Contour.
+
+## Non Goals
+
+- Session Affinity via application supplied Header or Cookie.
+- Session Affinity via client IP.
+
+## Background
+
+Session Affinity can be enabled in Envoy via three methods; header, cookie, and client IP.
+Of these three methods, this proposal will only implement cookie based session affinity.
+
+Header based affinity is similar to cookies--cookies are headers after all--however header based routing assumes the remote client supplies this header, which further assumes it has knowledge of the application deployment.
+That is to say, for a remote client to request that a specific pod in a deployment handle its requests, that requires a priori knowledge of that pods existence.
+A very uncloud native design.
+
+If the IngressRoute author wants to route _via_ a header, we are working on that as part of the next round of routing improvements.
+
+Client IP affinity will also not be offered.
+This is because retaining the remote IP of a TCP connection once it has passed through load balancers, network translation, an so on is subtle, difficult to configure, and can break silently.
+The end result is traffic which was expected to be hashed randomly ends up being hashed against the internal IP address of the next hop router.
+Because of the difficulty in reliably preserving client IP addresses, and the unpredictable nature of the failure to do so, we won't be adding client IP affinity in this design.
+
+## High-Level Design
+
+### User interface
+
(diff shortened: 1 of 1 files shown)


davecheney:
Here's a design for sticky sessions. I've also spiked this in a branch and it works as advertised with the following interesting features:

1. The hash policy, where we set the cookie details, applies to the route. However the hashing of the backend applies _per_ cluster and a cluster load balancing strategy. The end result is you can create a route weighted cluster that sets a session affinity cookie only once the request hits the backend with the cookie strategy will it acquire the session cookie and grow an affinity for the specific backend member. 

In normal use with a single backend -- I cannot think how someone could write an application that required session affinity and spread it across two or more groups of backend servers -- this is not visible. However it is called out in the design doc as reusing the `strategy` key like this does not preclude users configuring a mixed set of strategies for multiple backends, and it works!

2. Cookie attributes. Given there is no way to reuse a session cookie provided by the application (believe me, I spent days trying to do this) we always configure a cookie named `X-Contour-Session-Affinity`. The `X-Contour` prefix gives us a good guarantee that we're not conflicting with an application set value. 

The `X-Contour-Session-Affinity` cookie is a _session_ cookie, it has no `Expires` or `Max-Age` attribute. This means the cookie will expire when the browser session is closed. This seems reasonable as the fragility with session affinity means there is little value in persisting this cookie for days or weeks -- it does not represent a login token -- only a handle to in memory state on the target backend.

The Cookie has a path of `/`, which means when set it applies to all routes on the vhost. I think this is the right decision for a few reasons. 

a. It is reasonable that an application may consist of several routes to the same backend, a backend which requires session persistence; say `/cart` and `/checkout`. Visits to `/checkout` should arrive at the same server as `/cart`, so they must share the same session affinity token.
b. The session affinity token is just a random string. This string, when hashed, will select a server arranged in a ring (modulo) list. If we present the same string to a different cluster which consists of the same backends then the value will hash the same and the result _should_ land on the same server (but I really need to test this more).

design/session-affinity.md
```diff
@@ -0,0 +1,157 @@
...
+- Session Affinity via application supplied Header or Cookie.
+- Session Affinity via client IP.
+
+## Background
+
+Session Affinity can be enabled in Envoy via three methods; header, cookie, and client IP.
+Of these three methods, this proposal will only implement cookie based session affinity.
+
+Header based affinity is similar to cookies--cookies are headers after all--however header based routing assumes the remote client supplies this header, which further assumes it has knowledge of the application deployment.
+That is to say, for a remote client to request that a specific pod in a deployment handle its requests, that requires a priori knowledge of that pods existence.
```
stevesloka: nit: s/priori/prior/g

youngnick: further nit: or 'a priori' should be italic, since it's a Latin phrase.

design/session-affinity.md
```diff
@@ -0,0 +1,157 @@
...
+- Path: `/`. The cookie applies to all routes on this virtual host in the hope that other `strategy: cookie` backends, assuming they dispatch to the same set of servers will share the same affinity cookie.
+
+## Alternatives Considered
+
+Most applications that require session affinity already use a session cookie.
+Various other sticky session implementations, notably nginx and apache, were studied to see if it was possible to reuse the session cookie supplied by the application however this proved fruitless as: 
+
+a. The Apache/Tomcat jvmroute session persistence method requires each backend application instance to know its own name and co-ordinate with the load balancer. It is infeasable to attempt this inside a Kubernetes cluster.
+b. Using the session id or key as the ring hash suffered from the bootstrapping problem of the first request being routed to backend A, which sets a session cookie, however that session cookie will _not_ hash to backend A and cause subsequent requests to arrive at the wrong server, possibly repeating this behavior. Co-ordinating the session cookie creation between the backend application and Envoy is infeasble.
```
youngnick: Nit: s/infeasble/infeasible/

design/session-affinity.md
davecheney: wow TIL. I'm so here for the corrective punctuation advice!

PR #1158, opened 2019-06-06
internal/metrics: Implemented healthcheck for Contour

### LINKED ISSUE #283: Support health checking of Contour

stevesloka:
Fixes #283 by implementing health check for Contour. It does this by using the kubernetes client Contour uses and checks the server version. This api is simple enough to test the connection to the api server, and doesn't require understanding namespaces or anything additional from the cluster.
 
I think some future PR's should move this out of the `metrics` package. It was initially added there so Prometheus would be able to scrape metrics, but now does a bit more.

// cc #1132

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -18,6 +18,8 @@ import (
 	"fmt"
 	"net/http"
 
+	"k8s.io/client-go/kubernetes"
+
 	"github.com/heptio/contour/internal/httpsvc"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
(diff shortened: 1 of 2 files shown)


internal/metrics/metrics.go
```diff
@@ -203,19 +204,29 @@ func (m *Metrics) SetIngressRouteMetric(metrics IngressRouteMetric) {
...
 	return svc.Service.Start(stop)
 }
 
-func registerHealthCheck(mux *http.ServeMux) {
+func registerHealthCheck(mux *http.ServeMux, client *kubernetes.Clientset) {
 	mux.HandleFunc("/health", func(w http.ResponseWriter, r *http.Request) {
+		// Try and lookup Kubernetes server version as a quick and dirty check
+		_, err := client.ServerVersion()
+		if err != nil {
+			w.WriteHeader(http.StatusServiceUnavailable)
```
davecheney: Could you also please include the error message. we get from client.ServerVersion()?

There is a helper in the http package for this
```go
msg := fmt.Sprintf("Failed Kubernetes Check: %v", err)
http.Error(w, msg, http.StatusServiceUnavailable)
return
```

Issue #1006, opened 2019-04-10
Implement sticky session support based on design


timh:
Implement sticky session support based on design in PR #1005

davecheney:
A design draft for how we plan to implement this feature was added in #1005. TL;DR
```yaml
apiVersion: contour.heptio.com/v1beta1
kind: IngressRoute
metadata:
  name: httpbin
  namespace: default
spec:
  virtualhost:
    fqdn: httpbin2.davecheney.com
  routes:
  - match: /
    services:
    - name: httpbin
      port: 8080
      strategy: cookie
```
The `strategy: cookie` setting tells Envoy to use a cookie that it manages itself to route requests to the same backend.

PR #1163, opened 2019-06-11
internal/envoy: implement hash cookie policy session affinity

davecheney:
Updates #1006

### DIFF
--- a/examples/common/common.yaml
+++ b/examples/common/common.yaml
@@ -70,7 +70,7 @@ spec:
                 - RoundRobin
                 - WeightedLeastRequest
                 - Random
-                - cookie
+                - Cookie
             healthCheck:
               type: object
               required:
@@ -128,7 +128,7 @@ spec:
                             - RoundRobin
                             - WeightedLeastRequest
                             - Random
-                            - cookie
+                            - Cookie
                         healthCheck:
                           type: object
                           required:
--- a/examples/common/common.yaml
+++ b/examples/common/common.yaml
@@ -70,6 +70,7 @@ spec:
                 - RoundRobin
                 - WeightedLeastRequest
                 - Random
+                - cookie
             healthCheck:
               type: object
               required:
@@ -127,6 +128,7 @@ spec:
                             - RoundRobin
                             - WeightedLeastRequest
                             - Random
+                            - cookie
                         healthCheck:
                           type: object
                           required:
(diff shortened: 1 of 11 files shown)


examples/common/common.yaml
```diff
@@ -70,6 +70,7 @@ spec:
                 - RoundRobin
                 - WeightedLeastRequest
                 - Random
+                - cookie
```
stevesloka: Should we make the casing the same for these options? Either PascalCase or something?

internal/envoy/route.go
```diff
@@ -41,21 +42,40 @@ func RouteRoute(r *dag.Route, clusters []*dag.Cluster) *route.Route_Route {
...
+// hashPolicy returns a slice of hash policies iff at least one of the route's
+// clusters supplied uses the `cookie` load balancing stategy.
+func hashPolicy(r *dag.Route) []*route.RouteAction_HashPolicy {
+	for _, c := range r.Clusters {
+		if c.LoadBalancerStrategy == "cookie" {
+			return []*route.RouteAction_HashPolicy{{
+				PolicySpecifier: &route.RouteAction_HashPolicy_Cookie_{
+					Cookie: &route.RouteAction_HashPolicy_Cookie{
+						Name: "X-Contour-Session-Affinity",
+						Ttl:  duration(0),
```
stevesloka: What does this ttl duration do? How long the cookie is valid?

examples/common/common.yaml
davecheney: That’s a good question. I thought about it some when I chose `cookie`, IMO the pascal case Of the other options are weird because they are a protrusion of envoy’s naming scheme into our world. Lower case cookie was me sort of saying that this isn’t an envoy assigned strategy, but one we synthesise ourselves. 

What I think we should do in the future is relax the marching on that field, if possible, and match on the word reguardless of case. But having said all that, I’ll make it `Cookie` to be consistent with the other values.

internal/envoy/route.go
davecheney: If the ttl field is omitted envoy will not emit the cookie - it basically says “use the cookie with this name, I don’t manage it”. We don’t want that, we want envoy to own the cookie so we have to provide a ttl. 

Ttl 0 says to the browser this is a session cookie, it’s lifespan is controlled by the browser, not envoy. It lives as long as the browser wants which seemed like the only reasonable choice.

examples/common/common.yaml
stevesloka: I agree I don't like the pascal case all that much. =)

davecheney: Yeah, it just feels out of place in k8s, the predominant style is either a pithy lower case word or camelCase. I’m not sure if calling it `cookie` struck a blow for good, or just chaos

PR #1171, opened 2019-06-13
internal/dag: rebuild should not trigger on unrelated secrets and services

### LINKED ISSUE #1166: internal/dag: rebuild should not trigger on unrelated secrets
Inside a k8s cluster there are many secret objects, only some of them may be referenced by a valid ingress object. 

Because contour must watch all secret objects it may see an unrelated secret changing and this will trigger a rebuild of the dag. The end result _should_ be a noop, but it would be better to avoid this activity in the first place.

Contour should only trigger a dag rebuild if the secret is referenced from a valid ingress object.

stevesloka:
Fixes #1166 & #1172 by adding a check to see when a secret/service is added or updated if any Ingress or IngressRoute references. If the secret/service is not referenced, then skip any dag rebuild, however, still add the secret/service to the cache so that if later the Ingress or IngressRoute object later references, then the secret/service will be available.

### DIFF
--- a/internal/contour/k8s.go
+++ b/internal/contour/k8s.go
@@ -54,6 +54,10 @@ type Notifier interface {
 	// OnChange is called to notify the callee that the
 	// contents of the *dag.KubernetesCache have changed.
 	OnChange(*dag.KubernetesCache)
+
+	// ShouldUpdate is called to determine if the object changing
+	// is referenced from an Ingress / IngressRoute object
+	ShouldUpdate(obj interface{}) bool
 }
 
 func (reh *ResourceEventHandler) OnAdd(obj interface{}) {
@@ -64,7 +68,9 @@ func (reh *ResourceEventHandler) OnAdd(obj interface{}) {
 	}
 	reh.WithField("op", "add").Debugf("%T", obj)
 	reh.Insert(obj)
-	reh.update()
+	if reh.Notifier.ShouldUpdate(obj) {
+		reh.update()
+	}
 }
 
 func (reh *ResourceEventHandler) OnUpdate(oldObj, newObj interface{}) {
@@ -89,7 +95,9 @@ func (reh *ResourceEventHandler) OnUpdate(oldObj, newObj interface{}) {
 		reh.WithField("op", "update").Debugf("%T", newObj)
 		reh.Remove(oldObj)
 		reh.Insert(newObj)
-		reh.update()
+		if reh.Notifier.ShouldUpdate(newObj) {
+			reh.update()
+		}
 
 	}
 }
@@ -100,7 +108,9 @@ func (reh *ResourceEventHandler) OnDelete(obj interface{}) {
 	// no need to check ingress class here
 	reh.WithField("op", "delete").Debugf("%T", obj)
 	reh.Remove(obj)
-	reh.update()
+	if reh.Notifier.ShouldUpdate(obj) {
+		reh.update()
+	}
 }
 
 func (reh *ResourceEventHandler) update() {
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -35,10 +35,10 @@ const (
 	StatusOrphaned = "orphaned"
 )
 
-// BuildDAG returns a new DAG from the supplied KubernetesCache.
-func BuildDAG(kc *KubernetesCache) *DAG {
+// BuildDAG returns a new DAG/ReferencedCache from the supplied KubernetesCache.
+func BuildDAG(kc *KubernetesCache) (*DAG, ReferencedCache) {
 	builder := &builder{source: kc}
-	return builder.compute()
+	return builder.compute(), builder.cache
 }
 
 // A builder holds the state of one invocation of Builder.Build.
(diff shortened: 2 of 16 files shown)


stevesloka:
I struggled to come up with an exact test based upon where I implemented the change. I'd like to test this behavior, am going to think about how to implement a specific test.

internal/e2e/sds_test.go
```diff
@@ -169,12 +169,12 @@ func TestSDSShouldNotIncrementVersionNumberForUnrelatedSecret(t *testing.T) {
 	// TODO(dfc) 1166: currently Contour will rebuild all the xDS tables
```
davecheney: please delete the todo, this PR fixes it.

davecheney: Please delete the skip on line 167

stevesloka: 👍

stevesloka: 👍

internal/contour/k8s.go
```diff
@@ -89,7 +95,11 @@ func (reh *ResourceEventHandler) OnUpdate(oldObj, newObj interface{}) {
 		reh.WithField("op", "update").Debugf("%T", newObj)
 		reh.Remove(oldObj)
 		reh.Insert(newObj)
-		reh.update()
+		// If secret updated, but is not used in an Ingress or IngressRoute resource then skip
+		// notifying that the cache has changed, but still allow it to be stored for later use
+		if reh.validSecret(newObj) {
```
davecheney: I think this logic needs to be propogated to OnDelete, deleting an unrelated secret shouldn't cause an update.

stevesloka: Ahh ok, my initial thought was we wanted to always rebuild, but you're right, an un-referenced secret still shouldn't cause an update.

internal/dag/cache.go
```diff
@@ -123,3 +123,27 @@ func (kc *KubernetesCache) remove(obj interface{}) {
...
+			}
+		}
+	}
+	for _, ingress := range kc.ingressroutes {
+		// If VirtualHost or TLS is not defined then continue
+		if ingress.Spec.VirtualHost == nil || ingress.Spec.VirtualHost.TLS == nil {
+			continue
+		}
+		if ingress.Spec.VirtualHost.TLS.SecretName == secret.GetName() &&
+			ingress.GetNamespace() == secret.GetNamespace() {
```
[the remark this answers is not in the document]
stevesloka: I've updated to check the TLS cert delegation in the cache, if the secret is referenced then I assume that it's in use somewhere.

stevesloka:
@davecheney I addressed the comments, ready for re-review. Thanks!

stevesloka:
Also, I updated this PR to add the services bit (#1172) but kept them in separate commits.

internal/contour/k8s.go
```diff
@@ -125,6 +138,21 @@ func (reh *ResourceEventHandler) validIngressClass(obj interface{}) bool {
 	}
 }
 
+// validObjectUpdate returns true iff secret that changed is referenced from a
+// *v1beta1.Ingress or ingressroutev1.IngressRoute resource
+func (reh *ResourceEventHandler) validObjectUpdate(obj interface{}) bool {
```
davecheney: suggestion: why not call the method; `shouldUpdate`?

internal/dag/cache.go
```diff
@@ -123,3 +123,69 @@ func (kc *KubernetesCache) remove(obj interface{}) {
...
+				if path.Backend.ServiceName == service.GetName() && ingress.GetNamespace() == service.GetNamespace() {
+					return true
+				}
+			}
+		}
+	}
+	for _, ingress := range kc.ingressroutes {
+		for _, route := range ingress.Spec.Routes {
+			for _, routeSvc := range route.Services {
+				if routeSvc.Name == service.GetName() && ingress.GetNamespace() == service.GetNamespace() {
```
davecheney: I'm not sure this is right. This will say that any service is referenced if it is present in a ingressroute even if the ingressroute is not referenced from a root.

stevesloka:
@davecheney @youngnick this is ready for another look.

internal/dag/builder.go
```diff
@@ -556,7 +599,7 @@ func defaultBackendRule(be *v1beta1.IngressBackend) v1beta1.IngressRule {
 }
 
 // DAG returns a *DAG representing the current state of this builder.
-func (b *builder) DAG() *DAG {
+func (b *builder) DAG(secrets map[Meta]*v1.Secret, services map[Meta]*v1.Service) *DAG {
```
davecheney: Passing these two values into DAG so they can be passed back out as fields on DAG feels wrong. These fields are already avaialble to DAG's caller, builder.compute, so they can be returned directly from dag.compute without weaving them through DAG. I think this gives extra weight to the comments ealier about not keeping around the whole dag so we can access two additional fields added to DAG. Just keep the two fields directly, preferably wrapped in some object that has a ShouldUpdate method.

phylake:
I really appreciate you working on this @stevesloka

stevesloka:
One thing I didn't do from the last review is that I'm still exporting `dag.Meta`. I can stop that and then create the same struct and create a copy method. It's bad that it's leaking through, but it's the same struct so that's why I did it, curious your thoughts.

davecheney:
@stevesloka I've done some work on this today, would you mind if i took this PR over?

stevesloka:
@davecheney yup, np

PR #1175, opened 2019-06-19
Update IngressRoute validation regex and add test yaml

### LINKED ISSUE #755: Allow ingress routes for hostnames without a TLD defined
I'd like to be able to create an ingress route with a simple hostname, (instead of|as well as) a fully qualified domain name.

For example, assigning an ingress route on any of the following names:
```
g
grafana
grafana.example.com
```

This is useful for services on a local LAN with a search domain configured because Envoy won't respond for services if the fully qualified name doesn't match exactly, and in a search path lookup, the TLD is not appended to the request.

The following CRD validator seems to work as I'd like and preserves the current validation logic:
```
fqdn:
  type: string
  pattern: ^([a-zA-Z0-9]{1,})$|^([a-zA-Z0-9]+(-[a-zA-Z0-9]+)*\.)+[a-z]{2,}$
```
It also manages to not crash the API server, which is ... um... better than previous experiments.

**Environment:**
- Contour version: 0.6.0
- Kubernetes version: (use `kubectl version`): v1.12.1
- Kubernetes installer & version: kubeadm v1.12.1
- Cloud provider or hardware configuration: homelab mixed amd64 and arm32/64.
NAME="Ubuntu"
VERSION="18.04.1 LTS (Bionic Beaver)"

### LINKED ISSUE #1117: contour.heptio-contour invalid fqdn

youngnick:
Update IngressRoute validation regex

Fixes #755
Fixes #1117

### DIFF
--- a/_integration/ingressroute-naming-test.yaml
+++ b/_integration/ingressroute-naming-test.yaml
@@ -1,0 +1,71 @@
+---
+apiVersion: contour.heptio.com/v1beta1
+kind: IngressRoute
+metadata: 
+  name: contour-validname
+  namespace: heptio-contour
+spec: 
+  virtualhost:
+    fqdn: contour.local
+  routes: 
+    - match: /
+      services: 
+        - name: contour
+          port: 8080
+
+---
+apiVersion: contour.heptio.com/v1beta1
+kind: IngressRoute
+metadata: 
+  name: contour-shortname
+  namespace: heptio-contour
+spec: 
+  virtualhost:
+    fqdn: contour
+  routes: 
+    - match: /
+      services: 
+        - name: contour
+          port: 8080
+---
+apiVersion: contour.heptio.com/v1beta1
+kind: IngressRoute
+metadata: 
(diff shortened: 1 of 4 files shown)


davecheney:
@youngnick pro tip: don't forget to make this PR for the milestone it should be reviewed in.

davecheney:
@youngnick good to do, land this and we can cross it off the 0.13 milestone.

Issue #1176, opened 2019-06-19
internal/grpc: properly handle Envoy ACK/NACK protocol


davecheney:
Envoy's stream protocol states that after each DiscoveryResponse message the subsequent DiscoveryRequest will contain a ACK or NACK for the previous response. Contour currently doesn't pay attention to this ACK/NACK protocol, but for better compliance with the XDS spec we should.

phylake:
NACKs happen (at least) when config validation fails. Envoy erroneously throws exceptions in its config constructors ([example](https://github.com/envoyproxy/envoy/blob/v1.10.0/source/common/router/config_impl.cc#L376)) which will leak memory (Destructors aren't run if exceptions are thrown in the ctor body. See Item 10 in More Effective C++ for more detail) for anything allocated in the constructor body. In the worst case Contour continues to send the same resources and gets NACKs back while Envoy leaks memory until it's oom-killed. If memory leaks and how much will be dependent on which object in the deeply nested config classes throws an exception and where.

Contour needs to mitigate this for xDS with resource hints by not sending the resource anymore if its NACKed. I don't know what mitigation it could do for LDS/CDS other than keep the previous set of resources around so they can be sent if the most recent version is NACKed.

davecheney:
For 0.15 I've landed #1350 and I plan to leave this issue here for the moment.

Sending invalid updates to Envoy is a problem -- we shouldn't do that, but I don't know what to do with the signal once we did send an invalid update. Logging is about the least worst thing we can do. Trying to cordon the offending update is complicated, it depends on which table it came through. Crashing contour is a non starter, so logging is about the best I can think to do at the moment. Hopefully logging at error level will give us a signal from the field which updates are causing the problem and we'll know if its something we can systematically filter out on the Contour side or if we need to build a framework to trace NACK'd responses from Envoy back to their original k8s object.

jpeach:
xref https://github.com/envoyproxy/go-control-plane/issues/46

sunjayBhatia:
> Crashing contour is a non starter, so logging is about the best I can think to do at the moment. Hopefully logging at error level will give us a signal from the field

Have we gotten any signal on this from users? Another option in addition to logging would be to increment a metric to make this issue more visible to operators

jpeach:
> > Crashing contour is a non starter, so logging is about the best I can think to do at the moment. Hopefully logging at error level will give us a signal from the field
> 
> Have we gotten any signal on this from users?

[...]

Issue #273, opened 2018-03-12
Improve gRPC XDS handshaking


davecheney:
Currently contour plays fast and loose with the xDS protocol[0]; we don't really check the sequence numbers sent to envoy, and we ignore the ack back. 

This should be improved for 0.5.

[0]: https://github.com/envoyproxy/data-plane-api/blob/master/XDS_PROTOCOL.md

rosskukulinski:
@davecheney is this still an open issue?

davecheney:
Yes. There is a delta between the go-control-plane implementation of the
xDS protocol, which was written after ours, and our own. Closing that gap
is important to make sure we implement the nuances of xDS correctly.

davecheney:
I'm going to bump this to 0.7 and we'll cover it with the general upgrades of the project dependencies early in the cycle.

rosskukulinski:
Sounds good, thanks

davecheney:
Lowering priority

davecheney:
As one of our oldest issues its time to come clean and recognise that this work is not going to happen any time soon. I’m going to close this issue and will open a new one when and if required.

lrouquette:
I'm surprised this is not a high priority item, especially in light of [this comment](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#when-to-send-an-update). From what I've seen while watching these logs for quite a while now, not respecting the protocol and [this issue](https://github.com/heptio/contour/issues/426) are the likely reasons for envoy to [OOM](https://github.com/heptio/contour/issues/599). Any thoughts?

davecheney:
@lrouquette yup, we know we send too many updates to Envoy. This is being tracked in the context of #499.

This ticket existed from a time before envoyproxy/go-control-plane existed and was intended to be a placeholder to unify those two implementations. Simply put, that never happened, so this ticket has been closed.

davecheney:
@lrouquette thank you for pointing out the mistake in the way we're handing the XDS transaction. I'm preparing a PR now that should correct the defect.

PR #1177, opened 2019-06-19
internal/grpc: handle one response per request

davecheney:
Updates #499
Updates #273
Updates #1176

The [XDS spec](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#when-to-send-an-update) says that Envoy will always initiate a stream with a
discovery request, and expects the management server to respond with
only one discovery response. After that, Envoy will initiate another
discovery request containing an ACK or a NACK from the previous
response.

Currently Contour ignores the ACK/NACK, this is #1176, however after
inspection of the current code it is evident that we're also not waiting
for Envoy to send the next discovery request.

This PR removes the inner `for {}` loop that would continue to reuse the
initial discovery request until the client disconnected. The previous
code was written in a time when we'd just implemented filtering and it
was possible for the filter to return no results, hence the inner loop
was--incorrectly--trying to loop until there was a result to return.

Huge thanks to @lrouquette who pointed this out.

### DIFF
--- a/internal/grpc/xds.go
+++ b/internal/grpc/xds.go
@@ -86,6 +86,8 @@ func (xh *xdsHandler) stream(st grpcStream) (err error) {
 			return err
 		}
 
+		// TODO(dfc) issue 1176: handle xDS ACK/NACK
+
 		// from the request we derive the resource to stream which have
 		// been registered according to the typeURL.
 		r, ok := xh.resources[req.TypeUrl]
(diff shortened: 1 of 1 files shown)


davecheney:
There is also some evidence that this PR improves the peak memory usage we're seeing with #499. I don't have a lot of evidence to support it, but my theory is the heap growth observed in 0.13.0.beta.1 are discovery requests that envoy issues in response to each discovery reply sent by contour. Unscientifically I believe as prior to this PR Contour was not calling `stream.Recv` to fetch the next discovery request, they were stacking up--or at least were not free'd--on the Envoy side until the connection was closed.

PR #1193, opened 2019-06-26
Update CONTRIBUTING, thanks @odacremolbap

### LINKED ISSUE #1191: docs: Contributing nits on Pulling and Building
**Go command missing?**

Probably a typo at https://github.com/heptio/contour/blob/master/CONTRIBUTING.md#fetch-the-source
```sh
go github.com/heptio/contour
```
I guess there should be a `get` in there

**Building**

- Default `make` target executes `install`, which will create put the binary in `$GOPATH/bin/contour` instead of  `current working directory`
- It would also be worth replacing `go build` with `go build ./cmd/contour/` when mentioning building.

https://github.com/heptio/contour/blob/master/CONTRIBUTING.md#building

**Container**
Nothing wrong here, but I like having a bash file to source when working that will setup my local env. With contour, setting `REGISTRY` sounds like a handy way for pushing images:
```sh
REGISTRY=docker.io/pmercado make push
```

https://github.com/heptio/contour/blob/master/CONTRIBUTING.md#build-an-image

(_Very opinionated_) I like finding something like a `localdev/` entry in `.gitignore` so I can `source local/my-env-file`, specially for complex setups.

youngnick:
Fix a `go` command missing.
Update Building section with some corrections.
Add some detail about using `make push` to push a Docker image.

Fixes #1191 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ vendor/
 .DS_Store
 go.sum
 localenvoyconfig.yaml
+localdev/
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,3 @@ vendor/
 .DS_Store
 go.sum
 localenvoyconfig.yaml
-localdev/
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,7 @@ This section describes how to build Contour from source.
 Contour uses [`go modules`][2] for dependency management.
 
 ```
-go github.com/heptio/contour
+go get github.com/heptio/contour
 ```
 
 Go is very particular when it comes to the location of the source code in your `$GOPATH`.
@@ -44,10 +44,8 @@ To build Contour, run:
 make
 ```
 
-This produces a `contour` binary in your current working directory.
+This uses a `go install` and produces a `contour` binary in your `$GOPATH/bin` directory.
 
-_TIP_: You may prefer to use `go install` rather than `go build` to cache build artifacts and reduce future compile times.
-In this case the binary is placed in `$GOPATH/bin/contour`.
 
 ### Running the unit tests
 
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -6,3 +6,4 @@ vendor/
 .DS_Store
 go.sum
 localenvoyconfig.yaml
+localdev/
```
odacremolbap: Thanks a lot!
TBH I didn't expect the `localdev/` at `.gitignore`, that's a folder that is useful to me, but probably some other people don't use it, or use a different name.

What about adding a comment above it to clarify it's purpose?
```

# Developer's personal hack files
localdev/
```
What about adding a comment above it to indicate the purpose?

Also, should `localenvoyconfig.yaml` be removed and the `Makefile` adapted to:

https://github.com/heptio/contour/blob/5046e28f053d1ff88e174fe4593a89bcd0c2ce0e/Makefile#L6


```
LOCAL_BOOTSTRAP_CONFIG = localdev/envoyconfig.yaml
```

stevesloka: Where does `localdev` come from? I don't see it referenced anywhere other than the .gitignore.

odacremolbap: if was a _light request_ here https://github.com/heptio/contour/issues/1191

I find it handy having my local envs and scripts at some folder inside the repo.
I saw that you are already using `localenvoyconfig.yaml`, the `localdev/` is a proposal to put in there any local dev assets, that might include the envoy config.

stevesloka: That's fine, but I don't see `localdev` being used in this PR. It's referenced in the .gitignore but nowhere else (unless I'm missing it).

odacremolbap: oh I see!

Yes, the PR is not using it, `localdev` would be that folder where you put handy scripts that are useful only at your local box, env vars pointing to your container registry, cloud creds if needed ... there should be no reference  to it at code.

But then, back at the issue I didn't thought about moving the `localenvoyconfig.yaml` to that directory because I expected some back and forth regarding that entry at the `.gitignore` ... since that's part of the current PR, I think it would be ok ask for another commit here that would modify `Makefile` to look for that file inside that local folder instead of having it hanging at the root.

Not sure if I explained it ok, and if that makes sense.

I'm at the slack channel if you want to reach out

youngnick: Good question @stevesloka! Yes, `localdev/` is listed only in the .gitignore file, no where else. I really like @odacremolbap's suggestion about comments and moving the envoyconfig.yaml files in there, I will do that later today. Maybe `local/` may be a better name though.

davecheney:
Sorry but no. A developers personal preference for where they place their source code is as varied as the day is long. I don’t want to set a precedent for one preference as that commits us to a path of including others, or worse, rejecting them for equally arbitrary reasons.

.gitignore
davecheney: Please revert this change. If this is your personal preference then you can add it to a global ~/.gitignore in your workspace.

youngnick: Will do, thanks for the feedback.

Issue #1075, opened 2019-05-09
Routing NG


davecheney:
When IngressRoute was introduced a year ago (july 2018) we chose to only support prefix matching for routes. This was mostly a time to market decision, but also reflected the fact that we didn't know what other ways customers wanted to match routes on -- we knew that envoy also supported other methods, but without a signal from our userbase, blindly adding support for all of the existing mechanisms felt like throwing the problem over the wall to our users to figure out what worked best.

A year later there is clear signal that prefix routing is too limited. This is not a surprise. What is more interesting is customers want to route not just on prefix, substring, and regex -- the three we identified last year -- but also header matching, source ip, user agent, and many more.

This issue is a placeholder for a larger discussion about revising ingressroute's route matching. This will be part of, and intended to be delivered before Contour 1.0.

davecheney:
Delivered in beta.1

PR #1195, opened 2019-06-27
IngressRoute v1 design document

### LINKED ISSUE #947: Header based routing

stevesloka:
Fixes #947, updates #1075 by providing an initial design for IngressRoute v1. 


design/ingressroute-v1-design.md
```diff
@@ -0,0 +1,320 @@
...
+
+In practice the linkage, or delegation, from root to vertex, is performed with a specific type of route action. You can think of it as routing traffic to another ingress route for further processing, instead of routing traffic directly to a service.
+
+## Detailed Design
+
+### Changes from v1beta1
+
+Here are some changes that are directly different from the v1beta1 design:
+
+- `services` under the `match` struct are renamed to `backends`
```
youngnick: Looks like you may need to `s/services/backends/g` - some of the sample IngressRoutes below are still using `services:`.

stevesloka: 👍

design/ingressroute-v1-design.md
```diff
@@ -0,0 +1,320 @@
...
+The IngressRoute spec will be updated to allow for a `header` field to be added which will allow for a set of key/value pairs to be applied to the route. Additionally, the path `match` moves to it's own `path` variable.
+
+**NOTE**: *Routing on a `header` value* still requires a path prefix to be specified. If pure header delegation is requested, then Contour will configure a wildcard path to match all paths resulting in a header match only across any path.
+
+#### Header Values
+
+IngressRoutev1 will support routing from various headers. It will implement commonly used headers such as user-agent & cookie since those will require a regex style implementation. By implementing them as speific types, we can apply those regex's automatically.
+
+- **header:** This is a generic header type and allows for matching custom headers
+- **cookie**: This is a built-in type and matches a `cookie` in the header while being implemented as a "contains" regex
```
davecheney: Cookies are just headers with a fixed value. Can we simplify cookie routing into a header called cookie and substring matching for key=value?

design/ingressroute-v1-design.md
```diff
@@ -0,0 +1,320 @@
...
+
+```yaml
+spec:
+  routes:
+  - match:
+      path: /foo
+      - header: x-header
+        value: a
+    delegate:
+      name: headera-ir
```
davecheney: What would the child ingressroute look like? Would it have the same match: stanza as the parent following the example we have in it beta1 where every path in the child ingressroute must share the same prefix as the parent?

design/ingressroute-v1-design.md
```diff
@@ -0,0 +1,320 @@
...
+
+**NOTE**: *Routing on a `header` value* still requires a path prefix to be specified. If pure header delegation is requested, then Contour will configure a wildcard path to match all paths resulting in a header match only across any path.
+
+#### Header Values
+
+IngressRoutev1 will support routing from various headers. It will implement commonly used headers such as user-agent & cookie since those will require a regex style implementation. By implementing them as speific types, we can apply those regex's automatically.
+
+- **header:** This is a generic header type and allows for matching custom headers
+- **cookie**: This is a built-in type and matches a `cookie` in the header while being implemented as a "contains" regex
+- **user-agent**: This is a built-in type and matches the `user-agent` in the header while being implemented as a "contains" regex
```
davecheney: User agent is the same as cookie, its another header, and people will absolutely have to use regex here because of the trash fire that is the user agent header.

Cookie and user-agent feel like helpers. If we give people flexible enough ways to match on headers generically, they can implement Cookie and User-agent themselves. I think this would be enough for IngressRoute 1 and we can add cookie and user-agent, and perhaps others, when needed, post contour 1.0.

design/ingressroute-v1-design.md
stevesloka: We could, I was just looking for ways to simplify the IngressRoute document. We're adding more cases and tests in Contour, but the user has a better UX. We could also start here and design the other later.

stevesloka: > Cookie and user-agent feel like helpers. If we give people flexible enough ways to match on headers generically, they can implement Cookie and User-agent themselves. I think this would be enough for IngressRoute 1 and we can add cookie and user-agent, and perhaps others, when needed, post contour 1.0.

👍

stevesloka:
I'm going to close this PR for now. I was a bit overzealous in what this design could tackle. I won't delete the branch just yet and will 100% take the feedback given into the next rev, but I think to leave this as-is will do more harm and be confusing for folks.

PR #1197, opened 2019-06-27
docs: Update kind deployment configuration

### LINKED ISSUE #1196: docs: update deploy-options.md for kind
Since the 0.4.0 release of kind, it's now possible to expose ports to the local machine when creating a cluster. This eliminates the need to `port-forward` to the Envoy pod when testing Contour.

stevesloka:
Fixes #1196 by updating kind docs for new 0.4.0 version. 

### DIFF
--- a/docs/deploy-options.md
+++ b/docs/deploy-options.md
@@ -64,12 +64,30 @@ minikube service -n heptio-contour contour --url
 The response is always an IP address, for example `http://192.168.99.100:30588`. This is used as CONTOUR_IP in the rest of the documentation.
 
 ### kind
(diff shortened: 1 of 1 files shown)


stevesloka:
Good catch on the path to the config file @davecheney! I've updated to make that more clear. =)

davecheney:
@stevesloka given we have `examples/` which has a variety of examples, maybe we could include a sample kind config there?

stevesloka:
@davecheney yup that's a good point. I added the file to examples directory in this PR: https://github.com/heptio/contour/pull/1202

PR #1201, opened 2019-06-28
Add gRPC over TLS support

youngnick:
Updates #862
Updates #881

- Make it clear that all three TLS params are required if you supply any in all three modes.
- Update `contour serve` variables to make it clear they're not client certs.
- Remove redundant test now that all three params are required.
- Fix a small linting thing in `cli.go` by adding comments to exported functions.
- Add secure-local Make target as a helper for local development. This needs more documentation, to be completely finished (basically, documenting how to make a CA and issue the certs correctly).
- Add env var support for all the TLS flags.


.gitignore
```diff
@@ -6,3 +6,5 @@ vendor/
 .DS_Store
 go.sum
 localenvoyconfig.yaml
+securelocalenvoyconfig.yaml
```
davecheney: This is ok for this PR, but in the future i'd prefer to not have a process that poops random files into the root of the repository. Please take this as a todo to try to eliminate these in the future.

Makefile
```diff
@@ -45,7 +47,21 @@ ifeq ($(TAG_LATEST), true)
 endif
 
 $(LOCAL_BOOTSTRAP_CONFIG): install
-	contour bootstrap --xds-address $(LOCALIP) --xds-port=8001 $@
+	contour bootstrap --xds-address $(LOCALIP) --xds-port=8001
```
davecheney: This line lost the `$@` on the end, could you please restore it.

cmd/contour/cli.go
```diff
@@ -15,41 +15,96 @@ package main
...
+	CAFile      string
+	ClientCert  string
+	ClientKey   string
 }
 
 func (c *Client) dial() *grpc.ClientConn {
-	conn, err := grpc.Dial(c.ContourAddr, grpc.WithInsecure())
-	check(err)
+
+	var conn *grpc.ClientConn
```
davecheney: You can eliminate these local declarations as follows

davecheney: Just delete `var err error`, each line below that does something like `conn, err := ...` has declared a new err inside the nested block, effective shadowing it.

cmd/contour/cli.go
```diff
@@ -15,41 +15,96 @@ package main
...
+			// override?
+			// The ServerName here needs to be one of the SANs available in
+			// the serving cert used by contour serve.
+			ServerName:   "contour",
+			Certificates: []tls.Certificate{certificate},
+			RootCAs:      certPool,
+		})
+
+		// Create a connection with the TLS credentials
+		conn, err = grpc.Dial(c.ContourAddr, grpc.WithTransportCredentials(creds))
```
davecheney: you could do, 

conn, err := grpc.Dial(...)
check(err)
return conn
}

cmd/contour/cli.go
```diff
@@ -15,41 +15,96 @@ package main
...
+		check(err)
+	} else {
+		conn, err = grpc.Dial(c.ContourAddr, grpc.WithInsecure())
+		check(err)
+	}
+
```
davecheney: But I think the better way to do this would be to understand the role of the parameter to grpc.Dial, this is a vararg argument, I think in the godoc it's called `options`; so I would do something like this
```go
var options []grpc.DialOption

switch {
case c.CAFile != "" || c.ClientCert != "" || c.ClientKey != "" {
    // build creds
    options = append(options, grpc.WithTransportCredentials(creds))
default:
   // insecure
   options = append(options, grpc.WithInsecure())
}
conn, err := grpc.Dial(c.ContourAddr, options...)
check(err)
return conn
```

cmd/contour/contour.go
```diff
@@ -49,19 +53,30 @@ func main() {
 	log := logrus.StandardLogger()
 	app := kingpin.New("contour", "Heptio Contour Kubernetes ingress controller.")
 	var config envoy.BootstrapConfig
+
+	// Set up a zero-valued tls.Config, we'll use this to tell if we need to do
+	// any TLS setup for the 'serve' command.
+	tlsconfig := tls.Config{}
```
davecheney: prefer
```
var tlsconfig tls.Config
```

cmd/contour/contour.go
```diff
@@ -296,3 +333,32 @@ func getEnv(key, fallback string) string {
...
+	}
+
+	ca, err := ioutil.ReadFile(caFile)
+	if err != nil {
+		return err
+	}
+	certPool := x509.NewCertPool()
+
+	if ok := certPool.AppendCertsFromPEM(ca); !ok {
+		return err
```
davecheney: err will be nil there because you checked it on line 347. You'll probably have to do
```
if ...; !ok {
    return fmt.Errorf("Unable to append certificate to pool")
}
```
perhaps there is something in the `ca` value which you can reerence, or not, just use the value of caFile

internal/envoy/bootstrap.go
```diff
@@ -85,9 +87,65 @@ func Bootstrap(c *BootstrapConfig) *bootstrap.Bootstrap {
...
+		if !(c.GrpcClientCert != "" && c.GrpcClientKey != "" && c.GrpcCABundle != "") {
+			log.Fatal("You must supply all three TLS parameters - --envoy-cafile, --envoy-cert-file, --envoy-key-file, or none of them.")
+		}
+		b.StaticResources.Clusters[0].TlsContext = grpcUpstreamTLSContext(c.GrpcCABundle, c.GrpcClientCert, c.GrpcClientKey)
+	}
+
 	return b
 }
 
+func grpcUpstreamTLSContext(cafile string, certfile string, keyfile string) *auth.UpstreamTlsContext {
```
davecheney: nit, you can elide the repeated string declaration like this
```go
func grpcUpstreamTLSContext(cafile, certfile, keyfile string) *auth.UpstreamTlsContext
```

davecheney: Suggestion; it might be better to call this somethign like `UpstreamFileTLSContext` or something to make it clear that it returns an UpstreamTLSContext that expects files to be present on disk, not SDS

PR #1203, opened 2019-07-01
 update split deployments with TLS secured gRPC

### LINKED ISSUE #862: Encrypt gRPC traffic between Contour and Envoy

### LINKED ISSUE #1127: deployment: Envoy use Contour dns name for gRPC connection
Updating #1122, in the split deployment model, Envoy utilizes the built-in service environment variables (https://github.com/heptio/contour/blob/master/deployment/ds-hostnet-split/03-envoy.yaml#L67 && https://github.com/heptio/contour/blob/master/deployment/ds-hostnet-split/03-envoy.yaml#L69).

If the deployment or daemonset is created before the service exists, then these variables won't exist. 

We should look to have Envoy use the internal dns name of `contour` instead of the env variable. Last time I attempted this I ran into Envoy issues resolving that name properly. 

Also, if we switch to this model, the port cannot be automatically determined. Users will need to keep those numbers in sync (just noting so we can document properly).

youngnick:
- Add `make gencerts` and `make applycerts` to Makefile
- Update `ds-hostnet-split` to expect the new certs.
- Add documentation for gRPC with TLS including example.
- Change split deployments to use the 'contour' name instead of the service env vars. Fixes #1127 

Fixes #862
Updates #881

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ SRCDIRS := ./cmd ./internal ./apis
 PKGS := $(shell GO111MODULE=on go list -mod=readonly ./cmd/... ./internal/...)
 LOCAL_BOOTSTRAP_CONFIG = localenvoyconfig.yaml
 SECURE_LOCAL_BOOTSTRAP_CONFIG = securelocalenvoyconfig.yaml
+PHONY = gencerts
 
 TAG_LATEST ?= false
 # Used to supply a local Envoy docker container an IP to connect to that is running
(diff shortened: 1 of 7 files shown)


davecheney:
sgtm

youngnick:
> Should the secret type use `kubernetes.io/tls` instead of `generic`? This way we could have the cert/key's be consistent inside the secret which makes users generating their own a bit easier.

TIL about that type of secret. I'll get right on that, then merge.

PR #1209, opened 2019-07-03
go.mod: upgrade to client-go v11.0.0

### LINKED ISSUE #1183: Upgrade to client-go v11.0.x / kubernetes 1.14.x
Upgrade to client-go v11.0.x and kubernetes 1.14.x 

Description for doing this with go modules is here https://github.com/kubernetes/client-go/blob/master/INSTALL.md#go-modules

We will stay on client-go 11 / k8s 1.14 for 0.14, and consider client-go 12 for 0.15.

Note for implementors: you will need to regenerate the `api/` directory, also be watchful for failing tests, something may have changed with status setting.

davecheney:
Fixes #1183

### DIFF
--- a/apis/generated/clientset/versioned/clientset.go
+++ b/apis/generated/clientset/versioned/clientset.go
@@ -28,8 +28,6 @@ import (
 type Interface interface {
 	Discovery() discovery.DiscoveryInterface
 	ContourV1beta1() contourv1beta1.ContourV1beta1Interface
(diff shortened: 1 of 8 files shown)


davecheney:
Thanks. I think we'll do v12 next release, the backwards compatibility tables for client-go aren't super clear so I'm worried about sticking to close the bleeding edge when we don't have a regression test suite across (or even a list of) k8s versions we support.

Issue #1215, opened 2019-07-04
Create a subcommand for provisioning TLS certs for encrypted gRPC


youngnick:
This issue covers creating a helper subcommand to provision TLS certs for the user to assist with the process of setting up encrypted gRPC.

This subcommand isn't *required* to make the feature work, it's a convenience so that people don't need `openssl` installed on their laptops to be able to get encrypted traffic between Contour and Envoy working. 

The subcommand will have two parts:
- a generation part that will generate a CA keypair, a Contour keypair, and an Envoy keypair.
- an output part that will put those items either: into a Kubernetes cluster as a set of `kubernetes.io/tls` Secrets, as rendered YAML files, or as individual PEM files.

Working name for the subcommand is `tlsbootstrap`, but would appreciate naming thoughts.

youngnick:
This will fulfill Phase 2 of #881.

PR #1224, opened 2019-07-05
Add `certgen` subcommand with output to PEM format.

youngnick:
Updates #1215 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ $(LOCAL_BOOTSTRAP_CONFIG): install
 	contour bootstrap --xds-address $(LOCALIP) --xds-port=8001 $@
 
 $(SECURE_LOCAL_BOOTSTRAP_CONFIG): install
-	contour bootstrap --xds-address $(LOCALIP) --xds-port=8001 --envoy-cafile /config/certs/CA.cert --envoy-cert-file /config/certs/client.cert --envoy-key-file /config/certs/client.key $@
+	contour bootstrap --xds-address $(LOCALIP) --xds-port=8001 --envoy-cafile /config/certs/CAcert.pem --envoy-cert-file /config/certs/envoycert.pem --envoy-key-file /config/certs/envoykey.pem $@
 
 secure-local: $(SECURE_LOCAL_BOOTSTRAP_CONFIG)
 	docker run \
@@ -133,6 +133,7 @@ cleancerts:
 
 certs/CAkey.pem:
 	@echo No CA keypair present, generating
+	@mkdir -p certs
 	openssl req -x509 -new -nodes -keyout certs/CAkey.pem \
 		-sha256 -days 1825 -out certs/CAcert.pem \
 		-subj "/O=Project Contour/CN=Contour CA"
--- a/Makefile
+++ b/Makefile
@@ -131,9 +131,11 @@ applycerts: gencerts
 cleancerts:
 	@kubectl delete secret -n heptio-contour cacert contourcert envoycert
 
-certs/CAkey.pem:
-	@echo No CA keypair present, generating
+certs:
 	@mkdir -p certs
+
+certs/CAkey.pem: | certs
+	@echo No CA keypair present, generating
 	openssl req -x509 -new -nodes -keyout certs/CAkey.pem \
 		-sha256 -days 1825 -out certs/CAcert.pem \
 		-subj "/O=Project Contour/CN=Contour CA"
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -133,6 +133,7 @@ cleancerts:
 
 certs/CAkey.pem:
 	@echo No CA keypair present, generating
+	@mkdir -p certs
```
davecheney: makefile pro tip
```
certs/CAkey.pem: | certs
   ...

certs:
     @mkdir -p certs
```
this makes the _presence_ of `certs/` a prereq for certs/CAKey.pem, but make won't be picky about timestamps

cmd/contour/certgen.go
```diff
@@ -0,0 +1,40 @@
+// Copyright © 2018 Heptio
```
davecheney: 2019

cmd/contour/contour.go
```diff
@@ -96,6 +99,16 @@ func main() {
 	switch kingpin.MustParse(app.Parse(args)) {
 	case bootstrap.FullCommand():
 		doBootstrap(bootstrapCtx)
+	case certgenApp.FullCommand():
+		generatedCerts, err := certgen.GenerateCerts(certgenConfig)
+		if err != nil {
```
davecheney: check(err)

cmd/contour/contour.go
```diff
@@ -96,6 +99,16 @@ func main() {
...
+			// TODO(youngnick) placeholder, fix this.
+			log.Fatal(err)
+		}
+		err = certgen.OutputCerts(certgenConfig, generatedCerts)
```
davecheney: same

internal/certgen/certgen.go
```diff
@@ -0,0 +1,331 @@
+// Copyright © 2018 Heptio
```
davecheney: 2019

internal/certgen/certgen.go
```diff
@@ -0,0 +1,331 @@
...
+	"io"
+	"io/ioutil"
+	"math/big"
+	"os"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+const keySize = 2048
```
davecheney: please add a comment to explain why 2048 is the magic number

internal/certgen/certgen.go
```diff
@@ -0,0 +1,331 @@
...
+	log := certgenConfig.FieldLogger
+
+	if certgenConfig.OutputPEM {
+		// TODO(youngnick): Should we sanitize this value?
+		err := os.MkdirAll(certgenConfig.OutputDir, 0755)
+		if err != nil {
+			return err
+		}
+
+		log.Infof("Outputting certs to PEM files in %s/", certgenConfig.OutputDir)
```
davecheney: I'd drop the logger here and just use straigth fmt.Printf for two reasons

1. log goes to stderr
2. nobody needs timestamps in their "here's your files" line

internal/certgen/certgen_test.go
```diff
@@ -0,0 +1,89 @@
+// Copyright © 2018 Heptio
```
davecheney: 2019

internal/certgen/certgen_test.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func verifyCert(certPEM []byte, roots *x509.CertPool, dnsname string) error {
+	block, _ := pem.Decode(certPEM)
+	if block == nil {
+		return fmt.Errorf("Failed to decode %s certificate from PEM form", dnsname)
+	}
+
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
```
davecheney: why not return err here, it probably says more than the message on line 77

internal/certgen/certgen_test.go
```diff
@@ -0,0 +1,89 @@
...
+		return errors.New("Failed to parse an x509 certificate out of the Contour cert")
+	}
+
+	opts := x509.VerifyOptions{
+		DNSName: dnsname,
+		Roots:   roots,
+	}
+	if _, err = cert.Verify(opts); err != nil {
```
davecheney: why not just

return cert.Verify(opts)

if cert.Verify works then it'll return nil, like line 88. If it fails, the err message probably already says "verification failed"

cmd/contour/certgen.go
youngnick: 🤦‍♂

internal/certgen/certgen_test.go
youngnick: Verify returns two values, the first is the complete certificate chain (which we don't care about here), so I'll need to leave this unless there's way to throw away the first value that's shorter than this.

PR #1225, opened 2019-07-05
Patch cert manager

vincentdeity:
Updated documentation for cert-manager.md to actually apply the ingress after updating ingress.


davecheney:
@@vincentdeity ping, i'm sorry to nag but we can't take your change without DCO signoff.

PR #1227, opened 2019-07-07
internal/dag: reject * from IngressRoute FQDN

### LINKED ISSUE #1167: internal/dag: spec.virtualhost.fqdn should reject values with a *
Wildcards in the form of `*.something.com` are permitted by envoy but prohibited by the ingressroute and ingress spec. Well, the former is firm but the latter is undefined.

We need a test to assert that an ingress or ingressroute with a `*` in the fqdn field is rejected. In fact we should define the set of valid values for the fqdn field in contour. It's ok to also have that validation included in the crd validation, but as the latter can be modified or removed, we need to have the canonical check inside Contour.

ref: https://github.com/heptio/contour/pull/1156#discussion_r292139552

odacremolbap:
Disallow wildcarded FQDNs for `IngressRoutes`

- Will check and reject any FQDN containing `*` at DAG build time
- Wont check Ingress

Implementation details

- CDR untouched, there is a regex already in place that does this check among others
- Tests for `IngressRoutes` using `*` have been replaced with `bar.com`. `foo.com` is been used at other tests, but wanted to make it easy for reviewers to diff
- Added a test case at `TestDAGIngressRouteStatus`

Fixes #1167

Signed-off-by: Pablo Mercado <odacremolbap@gmail.com>

### DIFF
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -484,6 +484,11 @@ func (b *builder) computeIngressRoutes() {
 			continue
 		}
 
+		if strings.Contains(host, "*") {
+			b.setStatus(Status{Object: ir, Status: StatusInvalid, Description: fmt.Sprintf("Spec.VirtualHost.Fqdn %q cannot use wildcards", host), Vhost: host})
+			continue
+		}
+
 		var enforceTLS, passthrough bool
 		if tls := ir.Spec.VirtualHost.TLS; tls != nil {
 			// attach secrets to TLS enabled vhosts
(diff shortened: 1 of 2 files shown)


odacremolbap:
Rejecting `*` for ingresses might need some clarification:

- `Ingress` objects without FQDN have historically targeted every upstream service defined in the spec. That configuration at envoy means setting a `*` domain.
- rejecting `*` will means that current behaviour where non informed `Spec.Rules[].Host` at `Ingress` object is translated [here](https://github.com/heptio/contour/blob/6a855c2e1144da613baaf06a487211ff49021216/internal/dag/builder.go#L437), will be noop
- If we noop on `Ingress` objects that have no `Host` we will need to update its status with a message, probably adding an `IngressStatus` [here](https://github.com/heptio/contour/blob/6a855c2e1144da613baaf06a487211ff49021216/internal/k8s/status.go#L27)

If that sounds right, I'm happy to start with that. If this requires some more design I'm ok dropping this PR

davecheney:
Let's break this out. please open a new issue for rejecting * from Ingress,
lets leave this PR as IngressRoute only.

PR #1230, opened 2019-07-08
internal/certgen: handle error from writeCertsPEM

davecheney:


youngnick:
Actually, I'll fold this into #1232 .

PR #1232, opened 2019-07-08
Add `--yaml` and `--kube` output options to `certgen` subcommand

youngnick:
Updates #1215
Updates #881

### DIFF
--- a/docs/deploy-seperate-pods.md
+++ b/docs/deploy-seperate-pods.md
@@ -17,12 +17,17 @@ This configuration has several advantages:
     - `contourcerts`: must contain `contourcert.pem` and `contourkey.pem` keys that contain a certificate and key for Contour. The certificate must be valid for the name `contour` either via CN or SAN.
     - `envoycerts`: must contain `envoycert.pem` and `envoykey.pem` keys that contain a certificate and key for Envoy.
 
+For detailed instructions on how to configure the required certs manually, see the [step-by-step TLS HOWTO][2].
+
 ## Deploy Contour
 
 1. [Clone the Contour repository][1] and cd into the repo.
-2. Ensure you have `openssl` installed (or at least something that provides that command.`)
-2. Generate the keypairs required to secure the TLS connection between Contour and Envoy. You can do that using `contour gencerts --pem` and then `make applycerts`, assuming that your kubectl will connect to the cluster you want. If not, follow the [detailed generation directions][2]. 
-3. Run `kubectl apply -f examples/ds-hostnet-split/`
+2. Run `kubectl apply -f examples/ds-hostnet-split/`
+
+This will:
+- set up RBAC
+- run a Kubernetes Job that will generate one-year validity certs and put them into `heptio-contour`
+- Install Contour and Envoy in a separate fashion.
 
 **NOTE**: The current configuration exposes the `/stats` path from the Envoy Admin UI so that Prometheus can scrape for metrics.
 
--- a/examples/ds-hostnet-split/03-contour.yaml
+++ b/examples/ds-hostnet-split/03-contour.yaml
@@ -35,7 +35,7 @@ spec:
         - --xds-port=8001
         - --envoy-service-http-port=80
         - --envoy-service-https-port=443
-        - --contour-cafile=/ca/CAcert.pem
+        - --contour-cafile=/ca/cacert.pem
         - --contour-cert-file=/certs/tls.crt
         - --contour-key-file=/certs/tls.key
         command: ["contour"]
--- a/internal/certgen/certgen.go
+++ b/internal/certgen/certgen.go
@@ -161,12 +161,10 @@ func NewContourCerts() *ContourCerts {
 // writeCertPEMs writes out all the PEMs for all the certs, using
 // the stored filenames.
 func (cc *ContourCerts) writeCertPEMs(outputDir string) error {
-	err := cc.CA.writePEMs(outputDir)
-	if err != nil {
+	if err := cc.CA.writePEMs(outputDir); err != nil {
 		return err
 	}
-	err = cc.Contour.writePEMs(outputDir)
-	if err != nil {
+	if err := cc.Contour.writePEMs(outputDir); err != nil {
 		return err
 	}
 	return cc.Envoy.writePEMs(outputDir)
@@ -197,18 +195,15 @@ func (cc *ContourCerts) writeCACertKube(namespace string, client *kubernetes.Cli
 // the generated Secrets into the Kube `namespace`
 func (cc *ContourCerts) writeSecretYAMLs(outputDir, namespace string) error {
 	// First, write out just the CA Cert secret.
-	err := cc.writeCACertYAML(outputDir, namespace)
-	if err != nil {
+	if err := cc.writeCACertYAML(outputDir, namespace); err != nil {
 		return err
 	}
 	// Next, write out the full CA keypair
-	err = cc.CA.writeTLSYAML(outputDir, namespace)
-	if err != nil {
+	if err := cc.CA.writeTLSYAML(outputDir, namespace); err != nil {
 		return err
 	}
 	// Next, Contour's keypair
-	err = cc.Contour.writeTLSYAML(outputDir, namespace)
-	if err != nil {
+	if err := cc.Contour.writeTLSYAML(outputDir, namespace); err != nil {
 		return err
 	}
 	return cc.Envoy.writeTLSYAML(outputDir, namespace)
@@ -218,18 +213,15 @@ func (cc *ContourCerts) writeSecretYAMLs(outputDir, namespace string) error {
 // `namespace`.
 func (cc *ContourCerts) writeToKube(namespace string, client *kubernetes.Clientset) error {
 	// First, write out just the CA Cert secret.
-	err := cc.writeCACertKube(namespace, client)
-	if err != nil {
+	if err := cc.writeCACertKube(namespace, client); err != nil {
 		return err
 	}
 	// Next, write out the full CA keypair
-	err = cc.CA.writeTLSKube(namespace, client)
-	if err != nil {
+	if err := cc.CA.writeTLSKube(namespace, client); err != nil {
 		return err
 	}
 	// Next, Contour's keypair
-	err = cc.Contour.writeTLSKube(namespace, client)
-	if err != nil {
+	if err := cc.Contour.writeTLSKube(namespace, client); err != nil {
 		return err
 	}
 	return cc.Envoy.writeTLSKube(namespace, client)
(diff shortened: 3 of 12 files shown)


internal/certgen/certgen.go
```diff
@@ -60,11 +69,15 @@ type Config struct {
...
-	Contour *keyPair
-	Envoy   *keyPair
+type certData struct {
+	Filename string
+	Data     []byte
+}
+
+// writePEM writes a certificate out to its filename in outputDir.
+func (cd *certData) writePEM(outputDir string) error {
+	return writePEMFile(outputDir+"/"+cd.Filename, cd.Data)
```
davecheney: nit: although we don't support and proably never will support building for windows, its probably not a bad idea to future proof this with path/filepath.join(outputDir, cd.Filename). If for no other reason than someone will probably add this later and why have to review two PRs, eh?

internal/certgen/certgen.go
```diff
@@ -28,7 +28,16 @@ import (
 	"io/ioutil"
 	"math/big"
 	"os"
+	"path"
 	"time"
+
+	corev1 "k8s.io/api/core/v1"
```
davecheney: It's time to start splitting these files up. This is for a variety of reasons. The first is obviously length, but it's not just an arbitrary file length, its more than this file contains various operations that are related but independant in their use and operation.

For example we have functions that make TLS certificates, others that write out certificates to a file system, and some that create a k8s client to write certificates to k8s.

The hint is when a file called `certgen.go` grows imports required to write things to k8s, you know its doing too many things.

I'd suggest organising the functions in this file along either functional boundaries; eg, certs.go for things that generate certs, k8s.go for things that talk to k8s, and so on -- or along import boundaries, eg. group functions by the imports they require -- generally this has the same result.

internal/certgen/certgen.go
```diff
@@ -113,37 +158,84 @@ func NewContourCerts() *ContourCerts {
 	}
 }
 
-func (kp *keyPair) writePEMs(outputDir string) error {
-	err := kp.Cert.writePEM(outputDir)
+// writeCertPEMs writes out all the PEMs for all the certs, using
+// the stored filenames.
+func (cc *ContourCerts) writeCertPEMs(outputDir string) error {
```
davecheney: Does writeCertPEMs create outputDir? It probably should

internal/certgen/certgen.go
```diff
@@ -73,9 +86,40 @@ type keyPair struct {
...
+	err := kp.Cert.writePEM(outputDir)
+	if err != nil {
+		return err
+	}
+	return kp.Key.writePEM(outputDir)
+}
+
+// ContourCerts holds all three keypairs required, the CA, the Contour, and Envoy keypairs.
+type ContourCerts struct {
+	CA               *keyPair
```
davecheney: I feel strongly that CA should not be a keyPair, there should be no reason that the private Key of the CA that signed the Envoy and Contour certs should not be forgotten as soon as those certs have been signed. This will ensure that nobody can use the CA in the contour namespace to mint additional certs.

internal/certgen/certgen.go
```diff
@@ -60,11 +69,15 @@ type Config struct {
 	OutputPEM bool
 }
 
-// ContourCerts holds all three keypairs required, the CA, the Contour, and Envoy keypairs.
-type ContourCerts struct {
-	CA      *keyPair
-	Contour *keyPair
-	Envoy   *keyPair
+type certData struct {
+	Filename string
```
davecheney: I think that Filename can probably be inferred from the context in which the certificate is used.

davecheney: This is because, at least in the k8s secret context, certData.Filename is strongly correlated with the name (key) of the secret inside the secret object, and also with the names in deployment yamls, while you _could_ change the filename certData is written too, in reality, people probably won't and I sure don't want to have to document all the places where they need to chase it.

internal/certgen/certgen.go
```diff
@@ -326,3 +487,48 @@ func bigIntHash(n *big.Int) []byte {
...
+		return nil
+	}
+	return fmt.Errorf("can't overwrite %s", filename)
+
+}
+
+func writeSecret(filename string, secret *corev1.Secret) error {
+	s := json.NewYAMLSerializer(json.DefaultMetaFactory, scheme.Scheme, scheme.Scheme)
+	_, err := os.Stat(filename)
+	if err != nil {
```
davecheney: putting the success case _inside_ the conditional block isn't considered good form, its better to use a guard clause to bail out early if the precondition is not met.

In this case I'd recommend making a helper, say, `createFile(path, force) (f, err)` where createFile creates path iff it does not exist and force is false, otherwise returns with an appropriate error message, this will let you write something like this

f, err := createFile(filename, false) // never overwrite
if err != nil {
    return err
}

internal/certgen/certgen.go
```diff
@@ -326,3 +487,48 @@ func bigIntHash(n *big.Int) []byte {
...
+		// Can't stat the file, we'll create it
+		f, err := os.Create(filename)
+		if err != nil {
+			return err
+		}
+		defer f.Close()
```
davecheney: Sadly you cannot use defer here as we're writing to f and operating systems may not report an error to write until we try to close (which implicitly flushes the file).

internal/certgen/certgen.go
```diff
@@ -326,3 +487,48 @@ func bigIntHash(n *big.Int) []byte {
...
+		err = s.Encode(secret, f)
+		if err != nil {
```
davecheney: If the encode failed we should delete the partially written file on disk, otherwise people who've scripted this and are using the presence of the file, like make does, will get screwed.

internal/certgen/certgen.go
```diff
@@ -317,6 +412,72 @@ func newCA(r io.Reader, cn string, expiry time.Time) ([]byte, []byte, error) {
...
+				"app": "contour",
+			},
+		},
+		Data: map[string][]byte{
+			path.Base(certfilename): certPEM,
+		},
+	}
+}
+
+func newClient(kubeconfig string, inCluster bool) (*kubernetes.Clientset, error) {
```
davecheney: i'm not sure this function belongs in this package. What the other functions that rely on newClient need is the *kubernetes.ClientSet. It feels super weird that internal/certgen needs to know if its being called from a process which is running inside or outside a cluster.

Probably the way to solve this is to move newClient's caller, Gencerts into cmd/contour then it can reuse the existing NewClient function there

internal/certgen/certgen.go
```diff
@@ -113,37 +158,84 @@ func NewContourCerts() *ContourCerts {
...
+	err = cc.Contour.writeTLSKube(namespace, client)
+	if err != nil {
+		return err
+	}
+	return cc.Envoy.writeTLSKube(namespace, client)
 }
 
-// GenerateCerts performs the actual cert generation steps and then returns the certs for the output functions.
+// GenerateCerts performs the actual cert generation steps and then returns the certs for the output function.
 func GenerateCerts(certConfig *Config) (*ContourCerts, error) {
```
davecheney: I'm pretty sure this needs to move to cmd/contour. certConfig is 100% the cli flags we pass into `contour gencerts` so its probably better than certConfig never leaks out of cmd/contour.

internal/certgen/certgen.go
```diff
@@ -196,7 +288,10 @@ func OutputCerts(certgenConfig *Config, certs *ContourCerts) error {
 		}
 
 		fmt.Printf("Outputting certs to PEM files in %s/\n", certgenConfig.OutputDir)
-		certs.writeCertPEMs(certgenConfig.OutputDir)
+		err = certs.writeCertPEMs(certgenConfig.OutputDir)
```
davecheney: As an added bonus, if you move GenerateCerts into cmd/contour you can use the check helper 'cos its totally fine for cmd/contour to blow up on error whereas its not cool for a library to completely dump the program its called from.

Then that would give you an easy way to call `check(os.MkdirAll(certgenConfig.OutputDir)) and know that the output directory existed beforehand.

internal/certgen/certgen.go
```diff
@@ -196,7 +288,10 @@ func OutputCerts(certgenConfig *Config, certs *ContourCerts) error {
...
+		if err != nil {
+			return err
+		}
 	}
 
 	if certgenConfig.OutputYAML {
```
davecheney: This feels like selection; it probably shouldn't be a set of if cases where only one matches.

internal/certgen/certgen.go
```diff
@@ -326,3 +487,48 @@ func bigIntHash(n *big.Int) []byte {
...
+			return err
+		}
+		fmt.Printf("Created %s\n", filename)
+		return nil
+	}
+	return fmt.Errorf("can't overwrite %s", filename)
+
+}
+
+func writeSecret(filename string, secret *corev1.Secret) error {
```
davecheney: To make testing easier, writeSecret should take an os.Writer where it will serialise the secret too. This means the caller is responsible for creating, and closing that file, while we can write a test for writeSecret without having to poop a file on disk.

internal/certgen/certgen.go
youngnick: I intended that you could pass from one to three, and get all of the results you asked for.

internal/certgen/output.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+// writeSecret writes out a given Secret to a file.
+func writeSecret(f *os.File, secret *corev1.Secret) error {
+	s := json.NewYAMLSerializer(json.DefaultMetaFactory, scheme.Scheme, scheme.Scheme)
+	return s.Encode(secret, f)
+}
+
+func createFile(filepath string, force bool) (f *os.File, err error) {
+
```
davecheney: This can be written a different way that closes the (arguably tiny) race between Stat and Create
```go
flags := os.O_RDWR|os.O_CREATE|os.O_TRUNC
if !force {
   flags = flags | os.O_EXCL
}
err := os.OpenFile(name, flags, 0666)
```

youngnick: TIL.

docs/deploy-seperate-pods.md
```diff
@@ -17,12 +17,17 @@ This configuration has several advantages:
...
 
+For detailed instructions on how to configure the required certs manually, see the [step-by-step TLS HOWTO][2].
+
 ## Deploy Contour
 
 1. [Clone the Contour repository][1] and cd into the repo.
-2. Ensure you have `openssl` installed (or at least something that provides that command.`)
-2. Generate the keypairs required to secure the TLS connection between Contour and Envoy. You can do that using `contour gencerts --pem` and then `make applycerts`, assuming that your kubectl will connect to the cluster you want. If not, follow the [detailed generation directions][2]. 
-3. Run `kubectl apply -f examples/ds-hostnet-split/`
+2. Run `kubectl apply -f examples/ds-hostnet-split/`
```
davecheney: What's the upgrade path for someone who had previously deployed this example in their 0.13 cluster? Can they kubectl apply -f again; pick up the new contour image version and the certificate job will also deploy?

If this is a bit sketchy maybe we should clone ds-hostnet-split, add a new example, then we can deprecate ds-hostnet-split in 0.15?

examples/ds-hostnet-split/03-contour.yaml
```diff
@@ -35,7 +35,7 @@ spec:
         - --xds-port=8001
         - --envoy-service-http-port=80
         - --envoy-service-https-port=443
-        - --contour-cafile=/ca/CAcert.pem
+        - --contour-cafile=/ca/cacert.pem
```
davecheney: thanks, mixed case will make you curse yourself when you mix windows and mac; CapsLock, just don't do it.

docs/deploy-seperate-pods.md
youngnick: I just confirmed that the upgrade path is just `kubectl -f apply` again - although I just noticed I forgot to change the certgen image spec to `gcr.io/heptio-images/contour:master`, I'll fix that.

If we want to change the Job spec later, we may need to do something clever, as Job Pod specs are immutable. But this works just fine for upgrading from no security to with security.

Issue #1090, opened 2019-05-14
Add support for external client certificate validation (mTLS)


uablrek:
**Describe the solution you'd like**

Enable client certificate validation for external clients ([mTLS](https://en.wikipedia.org/wiki/Mutual_authentication))

Since mTLS is done for a virtual server, not a path, annotations in service manifests can not be used. Annotations in the `Ingress` manifest would be possible but be awkward. So the client validation configuration will only be available in the `IngressRoute` objects.

Example;
```yaml
apiVersion: contour.heptio.com/v1beta1
kind: IngressRoute
metadata:
  name: www
  namespace: default
spec:
  virtualhost:
    fqdn: foo.com
    tls:
      secretName: server-secret
      clientValidation:
        secretName: client-secret
  routes:
    - match: /
      services:
        - name: s1
          port: 80
```

**Anything else you would like to add:**

See also #433 

Relevant envoy documentation; https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/auth/cert.proto.html?highlight=require_client_certificate

Example;
```yaml
      tls_context:
        require_client_certificate: true
        common_tls_context:
          validation_context:
            verify_certificate_spki: j1fFFSBuF7S/Hzk5xsIq7N7wvVm8xdMpMnt0OUsz5rs=
```

In the example Subject Public Key Information (SPKI) is used to validate the client certificate. This does not have to be secret but is generated outside the cluster without any risk of exposing the private key and can be specified in clear text in the annotation.

The spki is generated as described in the envoy doc with;
```
openssl x509 -in default/cert/kahttp.crt -noout -pubkey \
  | openssl pkey -pubin -outform DER \
  | openssl dgst -sha256 -binary \
  | openssl enc -base64
```

**Environment:**
N/A

davecheney:
Thank you for raising this issue. At the moment there is no plan to add mTLS between Envoy and the backend pod. I am marking this issue as needs-product for input from our product managers.

uablrek:
This is *not* about mTLS between Envoy and the backend POD.

It is about providing the possibility to require that an client outside k8s authenticate itself with a client certficate.

The "downstream" term I got from the Envoy documentation https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/terminology

davecheney:
I'm sorry I misunderstood your request, Envoy's nomenclature is unusual.

[...]

PR #1233, opened 2019-07-08
Client certificate validation (mTLS) design proposal

uablrek:
Implementation proposal for #1090 

See also PR https://github.com/heptio/contour/pull/1226

### DIFF
--- a/design/tls-client-verification.md
+++ b/design/tls-client-verification.md
@@ -15,6 +15,19 @@ Status: Draft
 
 ## Background
 
+In TLS (and https) only the server is authenticated with a
+certificate, for instance you as a client can be sure that you speak
+with your bank and not some malicious site. But sometimes also the client
+must be authenticated. As noted in
+[wikipedia](https://en.wikipedia.org/wiki/Mutual_authentication)
+client certification (mTLS) is not very common for end-users but is
+more widespread for business-to-business (B2B) applications (which
+may use gRPC and REST APIs).
+
+I can't give a real-life example but it is easy to imagine cases where
+client validation is necessary, for instance for an admin interface to
+a server that is accesses by automated clients.
+
 Client certificate validation (mTLS) is supported by Envoy. It should
 be possible for `contour` users to utilize this feature.
 
--- a/design/tls-client-verification.md
+++ b/design/tls-client-verification.md
@@ -62,7 +62,7 @@ spec:
         spkis:
           - 2IEpPESU/mmC30tPsnOfbGKdwKdQfN/wZw1QWpjGlmk=
         subjectAltNames:
-          - Joe
+          - server.example.com
   routes:
     - match: /
       services:
--- a/design/tls-client-verification.md
+++ b/design/tls-client-verification.md
@@ -1,174 +1,116 @@
-# Client certificate validation (mTLS)
+# External client certificate validation (mutual TLS authentication)
 
 Status: Draft
 
+This document outlines how to add support for authentication of external
+clients by validating their client certificates.
+
 ## Goals
 
-- Allow client certificate validation (mTLS) on contour (performed by Envoy)
-- Allow various ways for client certificate validation; Spki, Hash and CA
-- Document mTLS configuration
+- Allow Contour to be used to protect the backend services from access of
+  unauthorized external clients.
+- Allow authentication of external clients by having Envoy validate that
+  client certificates are signed by trusted CA (Certificate Authority)
+- Allow configuration of trusted CA certificate(s) for validating the client
+  certificates.
 
 ## Non Goals
 
-- Support for k8s ingress documents.
-- Repeat configuration details described in Envoy documentation (make references)
+- Fine grained authorization on level of individual request
+  (example of non-goal: only client X can access resource Y on backend
+  service Z)
 
 ## Background
 
-In TLS (and https) only the server is authenticated with a
-certificate, for instance you as a client can be sure that you speak
-with your bank and not some malicious site. But sometimes also the client
-must be authenticated. As noted in
-[wikipedia](https://en.wikipedia.org/wiki/Mutual_authentication)
-client certification (mTLS) is not very common for end-users but is
-more widespread for business-to-business (B2B) applications (which
-may use gRPC and REST APIs).
-
-mTLS is used is to restrict access to sensiteve services to validated
-clients only. Some services do not provide client validation
-themselves then a reversed proxy (envoy/contour) is used, for
-instance;
-
-* Elasticsearch did not support authentication, [until very
-  recently](https://www.elastic.co/blog/tips-to-secure-elasticsearch-clusters-for-free-with-encryption-users-and-more). The
-  common pattern was to put reverse proxy in front of Elasticsearch to
-  handle TLS and authentication
-
-* Prometheus does not support either
-  [TLS](https://prometheus.io/docs/guides/tls-encryption/) or
-  [authentication](https://prometheus.io/docs/guides/basic-auth/). Common
-  pattern is to put reverse proxy in front.
-
-
-
-Client certificate validation (mTLS) is supported by Envoy. It should
-be possible for `contour` users to utilize this feature.
+In TLS (and HTTPS) often only the client authenticates the server.  TLS
+supports optional authentication of client which is referred to as
+*mutual TLS authentication* or *mTLS*.  Mutual TLS authentication is typically
+used in machine-to-machine (M2M) communication e.g. to protect sensitive REST
+APIs.  The application acting as TLS client authenticates itself towards the
+TLS server by using x509 certificate and by providing proof of posession of
+the corresponding private key.
 
 
 ## High-Level Design
 
-At a high level I propose the following:
+Envoy supports following options for certificate based client authentication:
 
-1. A new record "clientValidation" is added in spec.virtualhost.tls
-   in the IngressRoute.
+1. Verify that a chain of trust can be established from the presented client
+   certificate to the configured trusted root CA certificate (validation of
+   certificate)
+2. In addition to 1, verify that the subject alternative name of the client is
+   one of the names listed in the configuration (validation of client identity)
+3. Verify that the client certificate hash or subject public key hash matches
+   with configured hash (hash pinning)
 
-2. The `clientValidation` contains configuration for client
-   certificate validation. Many ways of client certificate validation
-   may be specified (all ways supported by Envoy).
+Option 1) is proposed to be implemented in this document.  It is sufficient
+for the very simplest mutual TLS authentication use cases.
 
+At a high level following CRD change is proposed:
 
-### Sample YAML
+*  A new record `clientValidation` is added in `spec.virtualhost.tls`
+   in the `HTTPProxy`.
+
+*  The `clientValidation` contains parameter `caSecret` which is a reference
+   to a secret containing trusted CA certificate for validating client
+   certificates.
+
+Sample YAML
 
 ```
-apiVersion: contour.heptio.com/v1beta1
-kind: IngressRoute
+apiVersion: projectcontour.io/v1
+kind: HTTPProxy
 metadata:
-  name: kahttp
+  name: mutual-tls-example
   namespace: default
 spec:
   virtualhost:
-    fqdn: kahttp.com
+    fqdn: example.com
     tls:
-      secretName: contour-secret
+      secretName: server-credentials
       clientValidation:
-        secretName: clientsecret
-        spkis:
-          - 2IEpPESU/mmC30tPsnOfbGKdwKdQfN/wZw1QWpjGlmk=
-        subjectAltNames:
-          - server.example.com
+        caSecret: ca-cert-for-client-validation
   routes:
-    - match: /
-      services:
-        - name: kahttp
+    - services:
+        - name: service
           port: 80
 ```
 
 
-## Detailed Design
+### Trusted CAs in Secrets
 
-Since this design proposal required some "learning-by-doing" most of this
-is implemented by https://github.com/heptio/contour/pull/1226
+The same approach shall be followed for configuring trusted CA certificates as
+is used currently to store the CA certificates for backend (Envoy upstream)
+validation:
 
-Unit-tests must be added on all appropriate places.
-
-
-### CAs in Secrets
-
-
-The store of CA information is an opaque kubernetes secret.
-The secret will be stored in the same namespace as the corresponding IngressRoute.
-TLS certificate delegation is not in scope for this proposal.
-
-The secret object should contain one entry named `ca.key`, the constents will be the CA public key material.
+The CA certificate is stored in an opaque Kubernetes secret.  The secret will
+be stored in the same namespace as the corresponding `HTTPProxy` object.
+The secret object shall contain entry named `ca.crt`.  The constents shall
+be the CA certificate in PEM format.  The file may contain "PEM bundle",
+that is, a list of CA certificates concatenated in single file.
 
 Example:
 ```
-% kubectl create secret generic my-certificate-authority --from-file=./ca.key
+% kubectl create secret generic ca-cert-for-client-validation --from-file=./ca.crt
 ```
 
-Contour already subscribes to Secrets in all namespaces so Secrets will be piped through to the `dag.KubernetsCache` automatically.
+TLS certificate delegation is not in scope for CA certificates.
 
 
-### Changes in APIs
+## Detailed Design
 
-The new configuration item `clientValidation` must be parsed and the
-CRD's must be updated to validate the new item.
+The new configuration item `spec.virtualhost.tls.clientValidation` must be
+parsed and the CRD's must be updated to validate the new item.
 
-
-### Changes to the DAG
-
-A new typed will be added to the `dag` package, `ClientValidation`
-to capture the validation parameters. It will be added to `SecureVirtualHost`.
-
-```go
-package dag
-
-type ClientValidation struct {
-	// The CA for client validation.
-	*Secret
-	// SPKIs used to validate the client certificate
-	Spkis []string
-	// Hashes used to validate the client certificate
-	Hashes []string
-	// Alternative subject names
-	SubjectAltNames []string
-}
-```
-
-### Changes to internal/envoy
-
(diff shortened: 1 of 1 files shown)


design/tls-client-verification.md
```diff
@@ -0,0 +1,231 @@
...
+less is may be wise to mimic this approach in `contour`, for instance
+herd-code forwaring of the entire certificate rather than the hash in
+phase 1.
+
+
+## High-Level Design
+
+At a high level I propose the following:
+
+1. A new record "clientValidation" is added in spec.virtualhost.tls
```
davecheney: You lost me here. The previous section talked about XFCC being between the end user and and backend application server. Now this is talking about a change to the tls settings for the whole virtual host. Is this a problem with the problem statement? Is this a security issue? If so what is the threat model? Is this an authentication issue? It's not clear to me at the moment.

design/tls-client-verification.md
```diff
@@ -0,0 +1,231 @@
...
+   in the IngressRoute.
+
+2. The `clientValidation` contains configuration for client
+   certificate validation. Many ways of client certificate validation
+   may be specified (all ways supported by Envoy).
+
+3. For phase 1 the configuration of what to forward in the XFCC header is
+   hard-coded to SANITIZE_SET and certificate hash.
+
+The design of phase 2, the XFCC configuration, is dependent on the
```
davecheney: I'd like to see an issue for something like XFCC should be supported by Contour. If that's the actual goal of this proposal then lets agree on that first before getting side tracked into implementation issues.

design/tls-client-verification.md
uablrek: I think you may be right. I will remove the XFCC header from the proposal since it may be confusing. It can be handled separately later on in another issue.

design/tls-client-verification.md
uablrek: Please see the comment above

uablrek:
To clearify; the XFCC header is not necessary for mTLS. It is a way of informing the backend that client validation has been conducted with a certain certificate.

So, XFCC is informational data to the backend, it has nothing to do with security.

uablrek:
More about XFCC; Different client certificates may be used, e.g an "admin" certificate or "user", having different rights. Then it is not sufficient for the server that the client is validated as any client, the server must know *which* client that is validated. Then the XFCC is used to forward the necessary client certificate details.

Not all applications needs client certificate details, only those who uses different types of client certificates, and the way of forwarding client certificate details is not standard and varies between proxies which of course makes it hard for applications

uablrek:
@davecheney Can you please re-check the document?

davecheney:
I’m sorry I ran out of time this week to reply. I’ll try to review on Monday.

design/tls-client-verification.md
stevesloka: Hey @uablrek, just taking a look here as well. 

I'm curious about @davecheney's comment earlier about if this is a vhost setting or a per route setting. Could I have each route in my cluster use a different cert based upon the upstream it's calling?

design/tls-client-verification.md
```diff
@@ -0,0 +1,143 @@
+# Client certificate validation (mTLS)
+
+Status: Draft
+
+## Goals
+
+- Allow client certificate validation (mTLS) on contour (performed by Envoy)
+- Allow various ways for client certificate validation; Spki, Hash and CA
```
stevesloka: Can only one type be specified? Envoy allows multiple, but I'm not sure of the use-cases that would drive picking multiple.

> When both: verify_certificate_hash and verify_certificate_spki are specified, a hash matching value from either of the lists will result in the certificate being accepted.

uablrek: Me neither to be honest. I rely on that Envoy knows what they are doing and try to allow all Envoy configurations for mTLS also in `contour`. That is also a reason for the non-goal;
> Repeat configuration details described in Envoy documentation (make references)
I will pass the question on to a security expert, but they tend to be busy...

uablrek:
@stevesloka 

I have improved the "Background" chapter. Please notify me if you think it should be more.

We want to use `contour` but a stopping issue is lack of support for mTLS. A proposal is to use `istio` as it also used Envoy and have mTLS support, but I would not like to go that way. I do not know the use cases in our product that require mTLS, just that they do.

design/tls-client-verification.md
uablrek: @stevesloka I asked the very same question in the Envoy slack :smile: The ansver is; No client certificate validation is per vhost, not for routes. The reason is that the route is in the data, but the client validation is performed before data is decrypted.

uablrek: For the XFCC header on the other hand it seems possible to forward different certificate details to dirrefent routes. But as @davecheney it is much better to add XFCC support as a different item, so I do not mention it in the design document any more.

uablrek:
@davecheney @stevesloka Can you please take another look at this after the updates?

I the proposal is acceptable I can continue with the implementation.

design/tls-client-verification.md
tsaarni: I think using multiple types makes most sense in combination of a `trusted_ca` and `verify_subject_alt_name`. In this case the administrator wants to uniquely authenticate a particular client(s) before allowing them to establish TLS connection.  For example a client with identity "CN=Joe" issued by certain trusted CA (or a sub-CAs of the trusted CA).  This is by far the most common use case for client certificates in general.  However, like discussed in PR comments, in some cases the specific client identity might not matter and one would then specify  `trusted_ca` alone.

Combination of `verify_certificate_spki` and `verify_certificate_hash` does not make sense to me, since it is likely that administrator picks just one way to pin a specific certificate: either administrator wants to identify client(s) by hash of public key or hash of the complete certificate file, not both at the same time.  Since this is an act of whitelisting a specific client certificate, it would be bit redundant or at least confusing, to use `trusted_ca` in combination with either of these two methods. When a client identified by a specific cert is explicitly marked as allowed to establish a TLS connection, validating that it is also issued by a trusted CA should not generally matter (though I have not tested if this combination is supported by Envoy). Obviously Envoy will still require the client to demonstrate proof-of-possession of the private key associated to the client certificate, in addition to validating the hash of the public part.

In theory one could imagine a use case where some clients are allowed to establish a connection because they are issued by `trusted_ca` and others because they are specifically pinned/whitelisted by their `verify_certificate_spki` and `verify_certificate_hash` fingerprint but I don't think this is very common scenario.

uablrek: @tsaarni I have ignored `verify_subject_alt_name` since I did not understand it's use. From your comment I see it's important. Can the configuration look something like this?;
```
      clientValidation:
        secretName: clientsecret
        subjectAltNames:
          - Joe
```
The list of `subjectAltNames` would be passed to Envoy in the `verify_subject_alt_name` field.

uablrek: Envoy doc; https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/auth/cert.proto.html?highlight=require_client_certificate#auth-certificatevalidationcontext

design/tls-client-verification.md
```diff
@@ -0,0 +1,156 @@
...
+In TLS (and https) only the server is authenticated with a
+certificate, for instance you as a client can be sure that you speak
+with your bank and not some malicious site. But sometimes also the client
+must be authenticated. As noted in
+[wikipedia](https://en.wikipedia.org/wiki/Mutual_authentication)
+client certification (mTLS) is not very common for end-users but is
+more widespread for business-to-business (B2B) applications (which
+may use gRPC and REST APIs).
+
+I can't give a real-life example but it is easy to imagine cases where
```
tsaarni: Here are some common example scenarios for protecting REST APIs:

* Elasticsearch did not support authentication, [until very recently](https://www.elastic.co/blog/tips-to-secure-elasticsearch-clusters-for-free-with-encryption-users-and-more). The common pattern was to put reverse proxy in front of Elasticsearch to handle TLS and authentication
* Prometheus does not support either [TLS](https://prometheus.io/docs/guides/tls-encryption/) or [authentication](https://prometheus.io/docs/guides/basic-auth/). Common patter is to put reverse proxy in front.

uablrek: Thanks!

design/tls-client-verification.md
tsaarni: Correction to my comment:  My example "CN=Joe" (X509 Common Name field) was for subjectName and subjectAltName would rather be something like email "joe@example.com", instead of just "joe".   Obvious from documentation, Envoy only supports validation of subjectAltName field. 

One complication is that X509 subjectAltName is not just free text: there is type involved (see list [here]( https://en.wikipedia.org/wiki/Subject_Alternative_Name) and definition of [GeneralName in RFC5280]( https://tools.ietf.org/html/rfc5280#section-4.2.1.6).  However Envoy documentation does not seem to specify what format it supports?

I have small doubt that maybe it just supports URIs since this feature might be added to Envoy due to Istio, which uses subjectAltName in the [SPIFFE certs](https://github.com/spiffe/spiffe/blob/master/standards/X509-SVID.md) and it leaves subjectName empty.   I might try to find this from envoy code.

uablrek: Ok thanks for the update. I will change from "Joe" to an URI, "service.example.com".

design/tls-client-verification.md
stevesloka: Ok so to clarify, is this looking to mTLS a public client calling into Envoy with a backend service? Or to secure communication from Envoy to a backend service inside the cluster?

design/tls-client-verification.md
```diff
@@ -0,0 +1,161 @@
...
```
stevesloka: I'd like to challenge this statement: "I can't give a real-life example but it is easy to imagine cases". 

You're looking to add this feature, so there's a need to have this, right? We should be adding real-life use-cases to back up the proposal, without those, then I question why we are adding this feature.

design/tls-client-verification.md
```diff
@@ -0,0 +1,161 @@
...
+
+Since this design proposal required some "learning-by-doing" most of
+phase 1 is implemented by https://github.com/heptio/contour/pull/1226
+
+Unit-tests must be added on all appropriate places.
+
+
+### CAs in Secrets
+
+Same as for [TLS backend verification](tls-backend-verification.md).
```
stevesloka: What is the same? How communication is secured? How secrets are named?

design/tls-client-verification.md
uablrek: This is *not* about cluster internal Envoy-backend communication. It is mTLS used to validate an external client accessing the service through contour/envoy from outside the cluster.

design/tls-client-verification.md
tsaarni: Just out of curiosity I checked what types of subjectAltName's Envoy supports as values of `verify_subject_alt_name`.  The code is [here](https://github.com/envoyproxy/envoy/blob/81654d18dfe211d739082c922776cf18bbf03275/source/extensions/transport_sockets/tls/context_impl.cc#L489).   The value can be DNS name, URI or IPv4/6 address.

design/tls-client-verification.md
uablrek: They are named, used and created in the same way.

I removed the reference and copied the chapter from tls-backend-verification.md instead to make it clear.

uablrek: Communication is secured with TLS (encryption), same as always. mTLS is just a way to limit the access to the server to validated clients.

design/tls-client-verification.md
uablrek: I have updated the examples with the ones in the comment above https://github.com/heptio/contour/pull/1233#discussion_r306218224

uablrek:
@stevesloka @davecheney Can you please have another look at this?

I have added use-cases as requested.

Please note that mTLS is not something I have invented myself and try to impose on contour. mTLS is a well established standard for client authentication over TLS, and it has support in envoy. I just would like to make it available for contour users since we need client authentication for some REST interfaces.

uablrek:
Here is another use-case, client validation using mTLS for helm-charts with an example fow to configure the `nginx` ingress controller for mTLS; https://hub.helm.sh/charts/drgrove/mtls

tsaarni:
I would like to continue with this PR (which we agreed with @uablrek).  The design doc now only contains proposal for validating external client certificate with trusted CA certificate, all other proposals were removed.

I'd be grateful for any feedback!

youngnick:
Oops, I also missed that CI hasn't passed, looks like there's a misspelling in the document. @tsaarni, could you run `make check` in your local repo and fix any errors? Once the CI is green and @davecheney has given the 👍 , then I'll squash and merge for you.

design/tls-client-verification.md
```diff
@@ -0,0 +1,125 @@
+# External client certificate validation (mutual TLS authentication)
```
davecheney: s/(mutual TLS authentication)//

design/tls-client-verification.md
```diff
@@ -0,0 +1,125 @@
...
+
+Status: Draft
+
+This document outlines how to add support for authentication of external
```
davecheney: Do not wrap lines. One line per sentence please. This makes it easier to comment a specific line, and reduces the diff size for comments.

design/tls-client-verification.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Non Goals
+
+- Fine grained authorization on level of individual request
+  (example of non-goal: only client X can access resource Y on backend
+  service Z)
+
+## Background
+
+In TLS (and HTTPS) often only the client authenticates the server.  TLS
```
davecheney: Please remove all the words about mutual TLS authentication, as I understand it this proposal is only about client side certificate authentication. If you want to add mutual TLS as a non goal, please do so in the previous section.

design/tls-client-verification.md
```diff
@@ -0,0 +1,125 @@
...
+
+Option 1) is proposed to be implemented in this document.  It is sufficient
+for the very simplest mutual TLS authentication use cases.
+
+At a high level following CRD change is proposed:
+
+*  A new record `clientValidation` is added in `spec.virtualhost.tls`
+   in the `HTTPProxy`.
+
+*  The `clientValidation` contains parameter `caSecret` which is a reference
```
davecheney: What is the justification for a new key, do you think we'll have to add additional keys under `clientValidation` in the future?

design/tls-client-verification.md
```diff
@@ -0,0 +1,125 @@
...
+be the CA certificate in PEM format.  The file may contain "PEM bundle",
+that is, a list of CA certificates concatenated in single file.
+
+Example:
+```
+% kubectl create secret generic ca-cert-for-client-validation --from-file=./ca.crt
+```
+
+TLS certificate delegation is not in scope for CA certificates.
+
```
davecheney: remove newline

design/tls-client-verification.md
```diff
@@ -0,0 +1,125 @@
...
+## Background
+
+In TLS (and HTTPS) often only the client authenticates the server.  TLS
+supports optional authentication of client which is referred to as
+*mutual TLS authentication* or *mTLS*.  Mutual TLS authentication is typically
+used in machine-to-machine (M2M) communication e.g. to protect sensitive REST
+APIs.  The application acting as TLS client authenticates itself towards the
+TLS server by using x509 certificate and by providing proof of posession of
+the corresponding private key.
+
```
davecheney: remove newline

design/tls-client-verification.md
```diff
@@ -0,0 +1,125 @@
...
+## Detailed Design
+
+The new configuration item `spec.virtualhost.tls.clientValidation` must be
+parsed and the CRD's must be updated to validate the new item.
+
+Client certificate validation is enabled in Envoy by setting
+`auth.DownstreamTlsContext.RequireClientCertificate` value to `true` and by
+adding trusted CA certificates to `auth.CommonTlsContext.ValidationContextType`.
+
```
davecheney: remove newline

design/tls-client-verification.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Alternatives Considered
+
+To use annotation in the `Ingress` object was considered to
+clumsy. Annotation must on `Ingress` level and refer to an individual
+virtual host.
+
```
davecheney: remove newline

design/tls-client-verification.md
tsaarni: I wish that in future we could add verification of subject alternative name of the client (option 2. in the list above).

design/tls-client-verification.md
tsaarni: I have now removed references to mutual TLS authentication and mTLS.

design/tls-client-verification.md
tsaarni: I have now reformatted the document accordingly.

design/tls-client-verification.md
tsaarni: I have now removed references to mutual TLS authentication and mTLS.

design/tls-client-verification.md
tsaarni: Done!

design/tls-client-verification.md
tsaarni: Done!

design/tls-client-verification.md
tsaarni: Done!

design/tls-client-verification.md
tsaarni: Done!

tsaarni:
Thanks @davecheney and @youngnick for the review! I have now corrected the document accordingly. If there is further comments, I'll gladly do updates!

tsaarni:
I wonder what your thinking is about this feature currently:  Would it be something that could be accepted? Is there some changes or investigations that I could do?

jpeach:
> On 23 Jan 2020, at 3:33 am, Tero Saarni <notifications@github.com> wrote:
> 
> I wonder what your thinking is about this feature currently: Would it be something that could be accepted? Is there some changes or investigations that I could do?

This is approved, so I'll do a final review pass and land it in the next couple of days.

design/tls-client-verification.md
jpeach: We should use `UpstreamValidation` (with a new name) to describe both upstream and downstream TLS validation policy.

PR #1249, opened 2019-07-17
internal/dag: Reject invalid service configuration if referenced service is missing

### LINKED ISSUE #520: Reject invalid service configuration
If a service is configured which does not exist in the cluster, Contour will still pass that to Envoy and the result will be the caller will receive 503 errors. Contour should not allow a route to be created which points to an invalid service.

stevesloka:
Fixes #520 by rejecting any IngressRoute object that references a Kubernetes service that doesn't exist or is malformed. When this situation is encountered, the status of the IngressRoute object will be updated and the object will be ignored.

### DIFF
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -659,7 +659,7 @@ func (b *builder) processRoutes(ir *ingressroutev1.IngressRoute, prefixMatch str
 				s := b.lookupHTTPService(m, intstr.FromInt(service.Port))
 
 				if s == nil {
-					b.setStatus(Status{Object: ir, Status: StatusInvalid, Description: "Service referenced is invalid or missing"})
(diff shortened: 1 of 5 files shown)


internal/dag/builder.go
```diff
@@ -656,20 +656,25 @@ func (b *builder) processRoutes(ir *ingressroutev1.IngressRoute, prefixMatch str
...
-						Upstream:             s,
-						LoadBalancerStrategy: service.Strategy,
-						Weight:               service.Weight,
-						HealthCheck:          service.HealthCheck,
-						UpstreamValidation:   uv,
-					})
+				s := b.lookupHTTPService(m, intstr.FromInt(service.Port))
+
+				if s == nil {
+					b.setStatus(Status{Object: ir, Status: StatusInvalid, Description: "Service referenced is invalid or missing"})
```
odacremolbap: Since there might be a number of services, probably adding the service name/port to the error message would be worth it.

stevesloka: Good suggestion @odacremolbap! I've updated that message to add the service name/port.

PR #1256, opened 2019-07-19
example: replace wget with bash magic

### LINKED ISSUE #1254: preStop hook for envoy is broken in v0.13.0
Looking at the event logs of my kubernetes cluster, I see the following:
```
2m37s       Warning   FailedPreStopHook      Pod          Exec lifecycle hook ([wget -qO- --post-data='' http://localhost:9001/healthcheck/fail]) for Container "envoy" in Pod "contour-5484dfb84b-2czgv_proxy(9ee76b6b-a7ae-11e9-aa10-d690536598ee)" failed - error: command 'wget -qO- --post-data='' http://localhost:9001/healthcheck/fail' exited with 126: , message: "OCI runtime exec failed: exec failed: container_linux.go:344: starting container process caused \"exec: \\\"wget\\\": executable file not found in $PATH\": unknown\r\n"
```

davecheney:
Fixes #1254

Sadly envoyproxy/proxy does not ship with wget, or curl, so resort to
bash-fu to hit Envoy's healthcheck/fail endpoint.

### DIFF
--- a/examples/deployment-grpc-v2/02-contour.yaml
+++ b/examples/deployment-grpc-v2/02-contour.yaml
@@ -69,7 +69,14 @@ spec:
         lifecycle:
           preStop:
             exec:
-              command: ["wget", "-qO-", "--post-data=''", "http://localhost:9001/healthcheck/fail"] 
+              command:
+              - bash
+              - -c
+              - --
+              - echo
+              - -ne
+              - "POST /healthcheck/fail HTTP/1.1\r\nHost: localhost\r\nConnection: close\r\n\r\n"
+              - '>/dev/tcp/localhost/9001'
       initContainers:
       - image: gcr.io/heptio-images/contour:master
         imagePullPolicy: Always
--- a/examples/ds-grpc-v2/02-contour.yaml
+++ b/examples/ds-grpc-v2/02-contour.yaml
@@ -70,7 +70,14 @@ spec:
         lifecycle:
           preStop:
             exec:
-              command: ["wget", "-qO-", "--post-data=''", "http://localhost:9001/healthcheck/fail"] 
+              command:
+              - bash
+              - -c
+              - --
+              - echo
+              - -ne
+              - "POST /healthcheck/fail HTTP/1.1\r\nHost: localhost\r\nConnection: close\r\n\r\n"
+              - '>/dev/tcp/localhost/9001'
       initContainers:
       - image: gcr.io/heptio-images/contour:master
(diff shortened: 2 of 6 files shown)


rochacon:
The Alpine based image has `wget` in it due to busybox. What if we revert #920?


```
# docker run --rm docker.io/envoyproxy/envoy:v1.10.0 wget --help
/docker-entrypoint.sh: 16: exec: wget: not found

# docker run --rm docker.io/envoyproxy/envoy-alpine:v1.10.0 wget --help
BusyBox v1.29.3 (2019-01-24 07:45:07 UTC) multi-call binary.

Usage: wget [-c|--continue] [--spider] [-q|--quiet] [-O|--output-document FILE]
  [--header 'header: value'] [-Y|--proxy on/off] [-P DIR]
  [-S|--server-response] [-U|--user-agent AGENT] [-T SEC] URL...

Retrieve files via HTTP or FTP

  --spider  Only check URL existence: $? is 0 if exists
  -c    Continue retrieval of aborted transfer
  -q    Quiet
  -P DIR    Save to DIR (default .)
  -S        Show server response
  -T SEC    Network read timeout is SEC seconds
  -O FILE    Save to FILE ('-' for stdout)
  -U STR    Use STR for User-Agent header
  -Y on/off  Use proxy
```

stevesloka:
There is some internal guidance around not using alpine based images which is why the switch happened. You are free to use in your project if you'd like.

stevesloka:
@davecheney I played around with this today and got this to work, but it's still a bit wonky: 
```
exec 5<>/dev/tcp/localhost/9001 && echo -ne "POST /healthcheck/fail HTTP/1.1\r\nHost: localhost\r\n\r\n" >&5 && cat <&5
```

examples/deployment-grpc-v2/02-contour.yaml
```diff
@@ -58,7 +58,15 @@ spec:
...
           preStop:
             exec:
-              command: ["wget", "-qO-", "--post-data=''", "http://localhost:9001/healthcheck/fail"] 
+              command:
+              - bash
+              - -c
+              - --
+              - echo
+              - -ne
+              - "POST /healthcheck/fail HTTP/1.1\r\nHost: localhost\r\nConnection:
```
youngnick: Don't know how this newline got in, but I think it will work without it. Weird.

davecheney: Done. PTAL

PR #1265, opened 2019-07-23
cmd/contour: Update how defaults are applied to args passed to serve command

### LINKED ISSUE #1260: cmd/contour: Kubeconfig arg always overwrites config file
Looking through how the new config file work was done, I was testing it out by creating a config file to set the `kubeconfig` param. 

What I found is that since there's a default value set on the flag, it always overwrites the config file. When I take out the default value it works as expected and the value of the config file is respected.

Here's my config file:
```
{
  "kubeconfig": "/Users/stevesloka/.kube/aws-cluster"
}
```

Here's the command I used to start contour:
```
$ make install && contour serve -c config.json  --xds-address=0.0.0.0        
```

stevesloka:
Fixes #1260 by changing how default values are passed to the contour serve args. This allows the config file to specify what has been set in the struct and also allows users to override them with args or ENV vars respectively. 

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -38,7 +38,7 @@ import (
 	"github.com/heptio/contour/internal/workgroup"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


davecheney:
merged in 46f7924

PR #1272, opened 2019-07-24
internal/contour: Allow circuitbreaking values to be defaulted from Config file

### LINKED ISSUE #1192: configurable default circuit breaking config
Contour allows circuit breaking fields such as max connections to be configured via annotations. If unset then the Envoy defaults are used. These defaults are much too low, and were one of the callouts at EnvoyCon with regard to sensible defaults in Envoy.

Providing a way to increase these for all IngressRoutes would be ideal. As operators of Contour we'd like to provide some more sensible defaults rather than go out to dozens of service teams and ask them to provide the annotations.

stevesloka:
Fixes #1192 by allowing maxconnection, maxpendingrequest, maxrequest, and maxretries to be set as global defaults for a Contour cluster from a config file. Specifying values using
the service annotation still overrides.

Sample configfile looks like this:
```json
{
  "circuitbreaker": {
    "max-connections": 100,
    "max-pending-requests": 200,
    "max-requests": 300,
    "max-retries": 400
  }
}
```

Contour can be run like this: 
```bash
$ contour serve -c config.json 
```

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -38,7 +38,7 @@ import (
 	"github.com/heptio/contour/internal/workgroup"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
-	kingpin "gopkg.in/alecthomas/kingpin.v2"
+	"gopkg.in/alecthomas/kingpin.v2"
 	coreinformers "k8s.io/client-go/informers"
 )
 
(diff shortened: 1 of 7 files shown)


stevesloka:
Working on another PR that involves config items, I opted to move these params to their own struct.

stevesloka:
Sure thing, I've asked about good defaults in the issue.

PR #1273, opened 2019-07-26
internal/contour: Allow tls-min-protocol-version to be defined from a config file

### LINKED ISSUE #980: Configure a Global TLS min protocol version for the entire cluster
Introduce a new global min TLS version configuration which can be set during the contour installation. With this global min TLS configuration "spec.virtualhost.tls.minimumProtocolVersion" would only take effect when the value is higher than the one mentioned during deployment. 

We are aware of the setting "spec.virtualhost.tls.minimumProtocolVersion" which users need to embed in their IngressRoute definition. Present default is TLS 1.1 and it may not be the standard in many firms. Thus we have to mutate and validate each IngressRoute resource for exception. 

**Environment:**

- Contour version: all
- Kubernetes version: (use `kubectl version`): all
- Kubernetes installer & version: all
- Cloud provider or hardware configuration: all
- OS (e.g. from `/etc/os-release`): all

Blocked on:
- [ ] #1130

stevesloka:
Fixes #980 by allowing min tls version to be set from a Config file. This defaults for all Ingress/IngressRoute resources unless overridden by annoation or TLS spec in IngressRoute.


cmd/contour/serve.go
```diff
@@ -151,6 +151,12 @@ type serveContext struct {
 	httpsAddr      string
 	httpsPort      int
 	httpsAccessLog string
+
+	tlsConfig `json:"tls"`
```
sudeeptoroy: We should also support flags so that it can be set during deployment.
similar flags: envoy-service-https-port

internal/contour/listener.go
```diff
@@ -122,6 +127,22 @@ func (lvc *ListenerVisitorConfig) httpsAccessLog() string {
 	return DEFAULT_HTTPS_ACCESS_LOG
 }
 
+// minProtocolVersion returns min protocol version if specified via annotation first
```
sudeeptoroy: once set by the operator.. only applicable values should be protocols higher or equal to the one set by the operator
example: lets set operator sets the min to 1.2 users cannot use protocol 1.1
below logic does not seems to achieve this..

internal/dag/builder.go
```diff
@@ -308,17 +308,18 @@ func isBlank(s string) bool {
 	return len(strings.TrimSpace(s)) == 0
 }
 
-// minProtoVersion returns the TLS protocol version specified by an ingress annotation
+// MinProtoVersion returns the TLS protocol version specified by an ingress annotation
```
sudeeptoroy: if we can set the min Version via flag/config during startup we can contain the logic around this function.. i.e return a version equal or higher then the one configured during startup..

sudeeptoroy: const (
        DEFAULT_MIN_TLS_VERSION = auth.TlsParameters_TLSv1_1
)

var minTlsProtocol auth.TlsParameters_TlsProtocol = DEFAULT_MIN_TLS_VERSION

var tlsProtocols = map[string]auth.TlsParameters_TlsProtocol {
        "1.1": auth.TlsParameters_TLSv1_1,
        "1.2": auth.TlsParameters_TLSv1_2,
        "1.3": auth.TlsParameters_TLSv1_3,
}

func SetClusterMinTlsVersion(proto string) {
        p, ok := tlsProtocols[proto]
        if !ok {
                p = DEFAULT_MIN_TLS_VERSION
        }
        minTlsProtocol = p
}
func getClusterMinTlsVersion() auth.TlsParameters_TlsProtocol {
        return minTlsProtocol
}

func minProtoVersion(version string) auth.TlsParameters_TlsProtocol {
        p, ok := tlsProtocols[version]
        if !ok || p < getClusterMinTlsVersion() {
                p = getClusterMinTlsVersion()
        }
        return p
}

davecheney:
I'm going to clean this up and push.

I've reverted the logic that lets annotations downgrade the configured version, that defeats the purpose of having this feature.

davecheney:
Closed c807d1f

PR #1303, opened 2019-08-08
internal/dag: Add config file item `allowPermitInsecure`

### LINKED ISSUE #864: Permit Insecure in multi-team environments
When a TLS cert is defined on an `IngressRoute`, Contour will automatically configure Envoy to return a 301 redirect response to any insecure request. Users can then optionally allow insecure traffic by setting the `permitInsecure` field on a route. 

This can introduce a security risk since administrators may not want to allow users who have been delegated, to be able to serve insecure routes from a Root delegation. 

// cc @scottslowe 

Blocked:
- [x] #1130

stevesloka:
Fixes #864 by adding config items `allowPermitInsecure` which if set to false, disables the ability to use `permitInsecure` from an IngressRoute.

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -61,21 +61,22 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 	)
 	// Set defaults for parameters which are then overridden via flags, ENV, or ConfigFile
 	ctx = serveContext{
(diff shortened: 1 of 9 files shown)


stevesloka:
Good idea on the name, I didn't really like the one that I had originally. I got it all refactored and added updated a bit more.

davecheney:
Closed  c5ac052

Issue #732, opened 2018-10-04
websockets broken when used with ingressroute + multiple clusters


davecheney:
This is a tech debt bug related to #726 

Currently envoy does not support combining websockets with route.Route_WeightedClusters, envoyproxy/envoy#3684

This is slated to be addressed in Envoy 1.9.0; ~6 months from now; so Contour 0.11 timeframe.

In the mean time; we should reject ingressroute configurations that combine multiple clusters with websockets with an appropriate message.

davecheney:
Bumping to 0.8, there's no bandwidth to address this properly in 0.7.

davecheney:
Fixed #1327

PR #1327, opened 2019-08-13
internal/dag: Set IngressRoute status error if multiple services are configured with websockets

stevesloka:
Updates #732 by setting IngressRoute status has multiple services and enabled WebSockets. This is currently not supported in Envoy yet.

### DIFF
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -511,6 +511,12 @@ func (b *Builder) processRoutes(ir *ingressroutev1.IngressRoute, prefixMatch str
 			return
 		}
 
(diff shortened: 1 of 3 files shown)


stevesloka:
There is no fix upstream yet. This was a suggestion in the issue in the meantime. I'll add a test to cover the new functionality.

stevesloka:
@davecheney I've added a test to cover. When there are multiple upstreams for a route and websockets is enabled, then the route is rejected.

stevesloka:
Good catch @davecheney, I've added an e2e test.

Issue #1172, opened 2019-06-14
internal/dag: rebuild should not trigger on unrelated services


stevesloka:
Inside a k8s cluster there are many service objects, only some of them may be referenced by a valid ingress object.

Because contour must watch all service objects it may see an unrelated service changing and this will trigger a rebuild of the dag. The end result should be a noop, but it would be better to avoid this activity in the first place.

Contour should only trigger a dag rebuild if the service is referenced from a valid ingress object.

stevesloka:
Unassigned myself since @davecheney was going to reimplement.

PR #1329, opened 2019-08-14
internal/contour: rewrite ResourceEventHandler

davecheney:
Updates #1166
Updates #1172

Rewrite ResourceEventHandler to own the KubernetesCache and mediate
modifications to the cache via channels.

This has the following advantages:

- dag.KubernetesCache becomes single threaded -- it no longer needs a
lock to prevent concurrent modification because there is only one
modifying goroutine; the one running EventHandler.run.
- The operation of the update/holdoff/rebuild loop is simpler to
explain. At any time run's main select loop is doing one of four things:

  1. Waiting for an event.
  2. Stopping.
  3. Processing an OnAdd/Update/Delete event from the informer, which
may include immediately rebuilding the DAG if the maximum holdoff has
been exceeded.
  4. Responding to a timer event from a previous holdoff.

- The holdoff now delays properly on first run, we no longer see a
decades long first response time.
- internal/e2e tests now test with a random holdoff delay to properly
exercise that code path. To mitigate the additional time waiting for
holdoff internal/e2e tests now use t.Parallel() to run with as many CPUs
as possible.
- EventHandler.OnAdd/Update/Delete events now return immediately, no
longer blocking the informer. Because of this the metrics collected per
event have been removed -- they are no longer representative of anything
useful.


davecheney:
@stevesloka @youngnick please review carefully. This change exposed a lot of implicit sequencing in the `internal/e2e` tests -- they assumed that once `rh.OnAdd/Update/Delete` returned, the result had been processed all the way through the grpc watcher. This was never true in production, when the holdoff timers were set, and was only true previously because the holdoff timers were set to zero 0 -- effectively forcing every informer event to cause a full DAG rebuild.

Now that contour.EventHandler processes events via a channel -- even if the holdoff delay is set to zero -- there is no relationship between the send on the `update` channel and the call to `updateDAG` completing. Rather than go down the route of sleeping after sending to the EventHandler, which is wasteful and inaccurate (timeouts are either too long and waste time, or too short and make tests flakey), the EventHandler exposes a Sequence channel which can be used to receive a notification that the event has been fully processed. Inside the internal/e2e tests we wait for that sequence to fire which restores the 'happens before' relationship that the internal/e2e tests expect without resorting to timeouts and retries. This also means that we can set reasonable values for holdoff and max holdoff and test that code path directly. In fact, internal/e2e now randomises those values to stress the tests and the event handler.

davecheney:
> Overall this looks good to me. Possibly, we could rename some of the methods like `bump` to be more clear, but that's very nit-y.

That's a good point, initially `bump` was called `notify`, it was more professional sounding, but wasn't really descriptive what it was notifying -- the observer, or the cache handler. I'll take another pass today and try to make it clearer.

PR #1331, opened 2019-08-14
cmd/contour: added support to set port redirect on envoy

### LINKED ISSUE #1300: Redirect: default HTTP->HTTPS redirect does not work if envoy is deployed on non-default host ports
http redirect url is not composed correctly when envoy is deployed on non standard hostport (!80, !443).
Deployment steps:
1. Reference deployment: https://github.com/heptio/contour/blob/master/examples/ds-hostnet-split/03-envoy.yaml 
2. Change hostports to different value; example:
        - containerPort: 80
          hostPort: 8000
          name: http
          protocol: TCP
        - containerPort: 443
          hostPort: 6443
          name: https
          protocol: TCP

developer's steps:
1. Create ingressroute with redirect enabled for the http url; use wildcard
2. Curl url
Example: curl -L http://redirect-example.com:8000

Issue:
Redirect 301 is received with a wrong redirect port.
example 
Received redirect url: https://redirect-example.com:8000 
Expected url: https://redirect-example.com:6443

**Environment:**

- Contour version: all

sudeeptoroy:
Fixes #1300 by adding support to set port redirect on envoy for cases where contour is serving non standard ports; new configuration parameter is added.  

Signed-off-by: Sudeepto Roy sudeepto.roy@gmail.com


davecheney:
@sudeeptoroy thank you for working on this. Regrettably I am not going to land this change as is because we first need to _talk_ about the problem and agree on a solution on the issue before moving to code. I will close this PR now and respond on #1300

PR #1333, opened 2019-08-15
Add `--insecure` flag, with tests for the TLS flag logic.

youngnick:
The TLS flag logic got complex enough that I wanted to ensure we had tests for it.

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -118,7 +118,7 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 	serve.Flag("contour-cafile", "CA bundle file name for serving gRPC with TLS").Envar("CONTOUR_CAFILE").StringVar(&ctx.caFile)
 	serve.Flag("contour-cert-file", "Contour certificate file name for serving gRPC over TLS").Envar("CONTOUR_CERT_FILE").StringVar(&ctx.contourCert)
 	serve.Flag("contour-key-file", "Contour key file name for serving gRPC over TLS").Envar("CONTOUR_KEY_FILE").StringVar(&ctx.contourKey)
-	serve.Flag("insecure", "Allow serving without TLS secured gRPC.").BoolVar(&ctx.PermitInsecureGRPC)
+	serve.Flag("insecure", "Allow serving without TLS secured gRPC").BoolVar(&ctx.PermitInsecureGRPC)
 	serve.Flag("ingressroute-root-namespaces", "Restrict contour to searching these namespaces for root ingress routes").StringVar(&ctx.rootNamespaces)
 
 	serve.Flag("ingress-class-name", "Contour IngressClass name").StringVar(&ctx.ingressClass)
@@ -175,7 +175,8 @@ type serveContext struct {
 	httpsPort      int
 	httpsAccessLog string
 
-	PermitInsecureGRPC bool
+	// PermitInsecureGRPC disables TLS on Contour's gRPC listener.
+	PermitInsecureGRPC bool `json:"-"`
 
 	tlsConfig `json:"tls"`
 
@@ -223,9 +224,7 @@ func (ctx *serveContext) verifyTLSFlags() error {
 	if !(ctx.caFile != "" && ctx.contourCert != "" && ctx.contourKey != "") {
 		return errors.New("you must supply all three TLS parameters - --contour-cafile, --contour-cert-file, --contour-key-file, or none of them")
 	}
-
 	return nil
-
 }
 
 // ingressRouteRootNamespaces returns a slice of namespaces restricting where
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -18,6 +18,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
+	"errors"
 	"io/ioutil"
 	"log"
 	"net"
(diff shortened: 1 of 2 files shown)


cmd/contour/serve.go
```diff
@@ -173,8 +175,13 @@ type serveContext struct {
 	httpsPort      int
 	httpsAccessLog string
 
+	// serveInsecure d `json:"serveinsecure"`ver gRPC for serving.
```
davecheney: I think your editor ate this line, I'd prefer to call it something like this
```
// DisableCertificateAuth disables mTLS between Envoy and Contour.
DisableCertificateAuth bool `json:"-"`
```
Let's not make this configurable by the config file -- disable json --users must pass `--insecure` in every invocation of `contour serve`.

cmd/contour/serve.go
```diff
@@ -213,6 +218,29 @@ func (ctx *serveContext) tlsconfig() *tls.Config {
 	}
 }
 
+// verifytlsflags indicates if the TLS flags are set up correctly.
+// This needs to handle four cases:
+// 1. Flags indicate no TLS required, and are configured correctly. Returns false and nil.
+// 2. Flags indicate no TLS required, but no `--insecure` present. Returns false and error.
+// 2. Flags indicate TLS required, and are not configured correctly. Returns false and error.
```
davecheney: s/2/3, 3/4,etc

cmd/contour/serve.go
```diff
@@ -213,6 +218,29 @@ func (ctx *serveContext) tlsconfig() *tls.Config {
...
+// 3. Flags indicate TLS required, and are configured correctly. Returns true and nil.
+func (ctx *serveContext) verifytlsflags() (bool, error) {
```
davecheney: nit, i know i'm not very consistent with this with helper funcs, but please rename this to `verifyTLSFlags`

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

cmd/contour/serve_test.go
```diff
@@ -64,3 +64,50 @@ func TestServeContextIngressRouteRootNamespaces(t *testing.T) {
 		})
 	}
 }
+
+func TestServeContextTLSParams(t *testing.T) {
+	tests := map[string]struct {
+		ctx     serveContext
+		iserror bool
```
davecheney: nit: maybe call this parameter `wanterror` or `expecterror`

cmd/contour/serve_test.go
```diff
@@ -64,3 +64,50 @@ func TestServeContextIngressRouteRootNamespaces(t *testing.T) {
...
+		"tls not supplied, no insecure flag": {
+			ctx:     serveContext{},
+			iserror: true,
+			proceed: false,
+		},
+	}
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			proceed, err := tc.ctx.verifytlsflags()
+			_, iserror := err.(error)
```
davecheney: then maybe write this as
```
goterror := err != nil
```

cmd/contour/serve.go
```diff
@@ -213,6 +218,29 @@ func (ctx *serveContext) tlsconfig() *tls.Config {
 	}
 }
 
+// verifytlsflags indicates if the TLS flags are set up correctly.
```
davecheney: I think this function is doing more than one thing, it should probably do less and the things that it stops doing move elsewhere. Specifically the conflation of handling "tls flags are wrong, but its ok because we said --insecure". What I'd recommend is 

- a function that validates that the TLS parameters are correct, returning an error if not
- a function that decides to validate the tls parameters or not, which is probably using `DisableCertificateAuth` above.

after that point the logic to serve TLS or not is governed by `DisableCertificateAuth`. If `DisableCertificateAuth` is true, then we don't consult the TLS flags, if it is false then we use the TLS flags as they're validated correct above.

youngnick: That makes much more sense, thanks.

cmd/contour/serve.go
```diff
@@ -116,7 +118,7 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 	serve.Flag("contour-cafile", "CA bundle file name for serving gRPC with TLS").Envar("CONTOUR_CAFILE").StringVar(&ctx.caFile)
 	serve.Flag("contour-cert-file", "Contour certificate file name for serving gRPC over TLS").Envar("CONTOUR_CERT_FILE").StringVar(&ctx.contourCert)
 	serve.Flag("contour-key-file", "Contour key file name for serving gRPC over TLS").Envar("CONTOUR_KEY_FILE").StringVar(&ctx.contourKey)
-
+	serve.Flag("insecure", "Allow serving without TLS secured gRPC.").BoolVar(&ctx.PermitInsecureGRPC)
```
davecheney: Remove period at the end of the description.

cmd/contour/serve.go
```diff
@@ -173,8 +175,12 @@ type serveContext struct {
 	httpsPort      int
 	httpsAccessLog string
 
+	PermitInsecureGRPC bool
```
davecheney: // PermitInsecureGRPC disables TLS on Contour's gRPC listener.
PermitInsecureGRPC bool `json:"-"`

cmd/contour/serve.go
```diff
@@ -213,6 +214,20 @@ func (ctx *serveContext) tlsconfig() *tls.Config {
...
+	if ctx.caFile == "" && ctx.contourCert == "" && ctx.contourKey == "" {
+		return errors.New("no TLS parameters and --insecure not supplied. You must supply one or the other")
+	}
+	// If one of the three TLS commands is not empty, they all must be not empty
+	if !(ctx.caFile != "" && ctx.contourCert != "" && ctx.contourKey != "") {
+		return errors.New("you must supply all three TLS parameters - --contour-cafile, --contour-cert-file, --contour-key-file, or none of them")
+	}
+
+	return nil
+
```
davecheney: delete line

PR #1343, opened 2019-08-16
examples: mount contour ConfigMap on /config/contour.json

### LINKED ISSUE #1342: Mount configmap for contour's config file into contour serve container
Now we have the config file, we should make that available as a configmap in `examples/` so that users can edit the config file as required.

Our config map is also necessary as a hook for leader election.

davecheney:
Fixes #1342 

- rename contour-config to envoy-config
- rename contour.json to envoy.json
- bind /config/contour.json to all contour pods


davecheney:
PTAL, mo' yaml errors

Issue #1204, opened 2019-07-02
Contour should move to a leader election model


davecheney:
Currently each Contour pod watches the API server and produces its own copy of the dag which it passes to a listening Envoy.

Once #881 is completed we will be able to decoupled the lifecycle of Contour and Envoy. This will allow us in turn to point Envoy to the ClusterIP of the active Contour instance. By deploying several contours, but having only one active at a time.

Related:
- [ ] #1178 

Blocked:
- [x] #881 
- [ ] #1185

stevesloka:
I think we have enough work done on #881 to unblock this issue.

davecheney:
Moving to beta1 milestone.

youngnick:
Here's a status update on this one (note that this feature requires a split deployment to work):

Implementing a version of leader election into `contour serve` is relatively straightforward. I have a rough version working already. Currently it will:
- use the `contour` ConfigMap to hold the leader election annotation (used for the distributed lock coordination by the standard client-go election library.
- have the 'serve gRPC' goroutine block until it becomes the leader
- if deposed as the leader, the instance of contour will shut itself down.

As far as `contour serve` goes,  this gives us the ability to have only a single manager contour running and serving gRPC, which will help with all the Envoys having a consistent view of the world.

To make this work properly, though, we need to change the Readiness probe for the deployment to check the availability of the gRPC port. This is easily done inside the Deployment spec, but means that the RollingUpdate of the deployment when you change it will never finish, as not all pods will become Ready. So, we either need to fudge the readiness probe for a short period when the pod starts up (not ideal) or move away from using a RollingUpdate for the Deployment (also not ideal). I will think about this more.

youngnick:
After some thought and discussion with @davecheney, we've decided on this approach:
- Move the Contour Deployment to 'Recreate' for its update strategy. This will mean that on updates to the Deployment spec, Kube will delete and recreate the pods. Until #1280 is complete, this means that the risk of Contour doing a partial update to Envoy and removing listeners briefly is still present. We've decided to proceed as this is no change from the current state.
- Readiness will be attached to a TCP probe of the gRPC port. So, only one contour Pod (the leader) will be Ready at any one time.
- I will add a `--disable-leader-election` flag for the single-pod case, going by the principle of least surprise. Remember that the single-pod deployment case is moving to deprecated in 0.15 though.

davecheney:
> I will add a --disable-leader-election flag for the single-pod case

[...]

PR #1353, opened 2019-08-19
Add leader election

youngnick:
- add `--disable-leader-election` command line flag.
- Leader election parameters, including ConfigMap name and namespace, are not tunable yet.

Updates #1204 
`serve.go` needs cleanup, but this establishes baseline functionality.

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -397,7 +397,6 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	// Generate the event recorder to send election events to the logs.
 	// Set up the event bits
 	eventBroadcaster := record.NewBroadcaster()
-	eventBroadcaster.StartLogging(log.Infof)
 	// Broadcast election events to the config map
 	eventBroadcaster.StartRecordingToSink(&clientcorev1.EventSinkImpl{Interface: clientcorev1.New(client.CoreV1().RESTClient()).Events("")})
 	eventsScheme := runtime.NewScheme()
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -452,7 +452,7 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	// AddContext will generate its own context.Context and pass it in,
 	// managing the close process when the other goroutines are closed.
 	g.AddContext(func(electionCtx context.Context) {
-		log := log.WithField("context", "leaderelect")
+		log := log.WithField("context", "leaderelection")
 		if !ctx.EnableLeaderElection {
 			log.Info("Leader election disabled")
 			// if leader election is disabled, signal the gRPC goroutine
@@ -466,10 +466,10 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		log.WithFields(logrus.Fields{
 			"configmapname":      ctx.LeaderElectionConfig.Name,
 			"configmapnamespace": ctx.LeaderElectionConfig.Namespace,
-		}).Info("Started leader election")
+		}).Info("started")
 
 		le.Run(electionCtx)
-		log.Info("Finishing leader election")
+		log.Info("stopped")
 	})
 
 	g.Add(func(stop <-chan struct{}) error {
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -383,13 +383,78 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	}
 	g.Add(debugsvc.Start)
 
-	// Leader Election starts here.
+	// step 12. Setup leader election
 
+	// leaderOK will block gRPC startup until it's closed.
 	leaderOK := make(chan struct{})
+	// deposed is closed by the leader election callback when
+	// we are deposed as leader so that we can clean up.
 	deposed := make(chan struct{})
+	// leadingDisabled
 	leadingDisabled := make(chan struct{})
 
-	g.Add(func(stop <-chan struct{}) error {
+	// Set up the leader election
+	// Generate the event recorder to send election events to the logs.
+	// Set up the event bits
+	eventBroadcaster := record.NewBroadcaster()
+	eventBroadcaster.StartLogging(log.Infof)
+	// Broadcast election events to the config map
+	eventBroadcaster.StartRecordingToSink(&clientcorev1.EventSinkImpl{Interface: clientcorev1.New(client.CoreV1().RESTClient()).Events("")})
+	eventsScheme := runtime.NewScheme()
+	// Need an eventsScheme
+	err := coreV1.AddToScheme(eventsScheme)
+	check(err)
+	// The recorder is what will record events about the resource lock
+	recorder := eventBroadcaster.NewRecorder(eventsScheme, coreV1.EventSource{Component: "contour"})
+
+	// Figure out the resource lock ID
+	resourceLockID, isPodNameEnvSet := os.LookupEnv("POD_NAME")
+	if !isPodNameEnvSet {
+		resourceLockID = uuid.New().String()
+	}
+
+	// Generate the resource lock.
+	// TODO(youngnick) change this to a Lease object instead
+	// of the configmap once the Lease API has been GA for a full support
+	// cycle (ie nine months).
+	rl, err := resourcelock.New(
+		resourcelock.ConfigMapsResourceLock,
+		ctx.LeaderElectionConfig.Namespace,
+		ctx.LeaderElectionConfig.Name,
+		client.CoreV1(),
+		coordinationClient,
+		resourcelock.ResourceLockConfig{
+			Identity:      resourceLockID,
+			EventRecorder: recorder,
+		},
+	)
+	check(err)
+
+	// Make the leader elector, ready to be used in the Workgroup.
+	le, err := leaderelection.NewLeaderElector(leaderelection.LeaderElectionConfig{
+		Lock:          rl,
+		LeaseDuration: ctx.LeaderElectionConfig.LeaseDuration,
+		RenewDeadline: ctx.LeaderElectionConfig.RenewDeadline,
+		RetryPeriod:   ctx.LeaderElectionConfig.RetryPeriod,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: func(ctx context.Context) {
+				log.WithFields(logrus.Fields{
+					"lock":     rl.Describe(),
+					"identity": rl.Identity(),
+				}).Info("elected leader")
+				close(leaderOK)
+			},
+			OnStoppedLeading: func() {
+				// The context being canceled will trigger a handler that will
+				// deal with being deposed.
+				close(deposed)
+			},
+		},
+	})
+	check(err)
+	// AddContext will generate its own context.Context and pass it in,
+	// managing the close process when the other goroutines are closed.
+	g.AddContext(func(electionCtx context.Context) {
 		log := log.WithField("context", "leaderelect")
 		if ctx.DisableLeaderElection {
 			log.Info("Leader election disabled")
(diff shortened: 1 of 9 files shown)


examples/ds-hostnet-split/03-contour.yaml
```diff
@@ -40,7 +42,8 @@ spec:
         - --contour-key-file=/certs/tls.key
         - --config-path=/config/contour.yaml
         command: ["contour"]
-        image: gcr.io/heptio-images/contour:master
+        # image: gcr.io/heptio-images/contour:master
```
davecheney: please revert this

cmd/contour/serve.go
```diff
@@ -178,17 +195,31 @@ type serveContext struct {
...
-	TlsConfig `yaml:"tls"`
+	TLSConfig `yaml:"tls"`
 
 	// DisablePermitInsecure disables the use of the
 	// permitInsecure field in IngressRoute.
-	DisablePermitInsecure bool `json:"disablePermitInsecure"`
+	DisablePermitInsecure bool `yaml:"disablePermitInsecure"`
+
+	DisableLeaderElection bool
+	LeaderElectionConfig  `yaml:"leaderelection"`
```
davecheney: If you don't want this to be configurable via yaml, use ``yaml:"-"``

cmd/contour/serve.go
```diff
@@ -352,6 +383,92 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+		log := log.WithField("context", "leaderelect")
+		if ctx.DisableLeaderElection {
+			log.Info("Leader election disabled")
+			// if leader election is disabled, signal the gRPC goroutine
+			// to start serving and block this goroutine forever.
+			close(leaderOK)
+			<-leadingDisabled
+		}
+
+		log.Info("Started leader election")
```
davecheney: please consider log.WithField(...).WithField(...).Info("started leader election") so we don't have to take a PR from the structured logging fans.

cmd/contour/serve.go
```diff
@@ -352,6 +383,92 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+		log.Infof("Leader election params: %#v", ctx.LeaderElectionConfig)
+
+		// Generate the event recorder to send election events to the logs.
+		// Set up the event bits
+		eventBroadcaster := record.NewBroadcaster()
+		eventBroadcaster.StartLogging(log.Infof)
```
davecheney: how much does this log? People complain when contour outputs periodic information.

cmd/contour/serve.go
```diff
@@ -352,6 +383,92 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+		err := coreV1.AddToScheme(eventsScheme)
+		check(err)
+		// The recorder is what will record events about the resource lock
+		recorder := eventBroadcaster.NewRecorder(eventsScheme, coreV1.EventSource{Component: "contour"})
+
+		// Figure out the resource lock ID
+		var resourceLockID string
+		resourceLockID, isPodNameEnvSet := os.LookupEnv("POD_NAME")
+		if !isPodNameEnvSet {
+			resourceLockID = uuid.New().String()
```
davecheney: Why make this configurable? If a UUID will work when Pod name is not set, why not just use the uuid always?

cmd/contour/serve.go
```diff
@@ -352,6 +383,92 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+			Lock:          rl,
+			LeaseDuration: ctx.LeaderElectionConfig.LeaseDuration,
+			RenewDeadline: ctx.LeaderElectionConfig.RenewDeadline,
+			RetryPeriod:   ctx.LeaderElectionConfig.RetryPeriod,
+			Callbacks: leaderelection.LeaderCallbacks{
+				OnStartedLeading: func(ctx context.Context) {
+					log.WithFields(logrus.Fields{
+						"lock":     rl.Describe(),
+						"identity": rl.Identity(),
+					}).Info("started leading")
```
davecheney: s/started leading/elected leader/

(started leader election above is too close to started leading)

cmd/contour/serve.go
```diff
@@ -352,6 +383,92 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+					// The context being cancelled will trigger a handler that will
+					// deal with being deposed.
+					close(deposed)
+				},
+			},
+		})
+		check(err)
+		// Running the leader election requires a context.Context, but we're
+		// not using them, so we'll just pass in an empty one.
+		le.Run(context.Background())
```
davecheney: What causes le.Run to exit? I'm guessing its the context being canceled. group.Run won't return until this function returns so we might need to make sure the context is canceled.

cmd/contour/serve.go
```diff
@@ -352,6 +383,92 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+		return nil
+	})
+
+	g.Add(func(<-chan struct{}) error {
+		// If we get deposed as leader, shut it down.
+
+		<-deposed
```
davecheney: You should use a select here because one of two things could happen, stop could be called, or deposed could be called. As written this function won't stop until it is deposed, and won't be deposed until it has been elected, none of which may happen. try this

```
g.Add(func( stop <- chan struct{}) error {
    select {
    case <- stop:
        // shut down
    case <- deposed:
        log.Info("deposed as leader")
    }
    return nil
}
```

cmd/contour/serve.go
```diff
@@ -361,6 +478,7 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	// step 13. create grpc handler and register with workgroup.
 	g.Add(func(stop <-chan struct{}) error {
 		log := log.WithField("context", "grpc")
+		<-leaderOK
```
davecheney: As above, we need to select on leaderOK or stop because this function might need to stop before it is elected.

cmd/contour/serve.go
youngnick: It means that you can see which pod is the leader both by inspecting the annotations on the ConfigMap and checking the Event stream on the ConfigMap. It can be difficult to tell which pod is the leader without that.

cmd/contour/serve.go
youngnick: When the state changes only, no periodic logs.

cmd/contour/serve.go
youngnick: I've had a crack at this, PTAL.

cmd/contour/serve.go
```diff
@@ -352,6 +383,99 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	}
 	g.Add(debugsvc.Start)
 
+	// Leader Election starts here.
+
+	leaderOK := make(chan struct{})
+	deposed := make(chan struct{})
+	leadingDisabled := make(chan struct{})
+
+	g.Add(func(stop <-chan struct{}) error {
```
davecheney: I suspect that 392-464 could be executed in the context of the current goroutine, not handed off to the group that would leave
```
		// Running the leader election requires a context.Context, but we're
		// not using them, so we'll just pass in an empty one.
		le.Run(electionCtx)
		return nil
```
to be run in a goroutine. This probably means the context handling needs to be redone. Perhaps we need a better way to integrate contexts with workgroups.

cmd/contour/serve.go
```diff
@@ -352,6 +383,99 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+						"lock":     rl.Describe(),
+						"identity": rl.Identity(),
+					}).Info("elected leader")
+					close(leaderOK)
+				},
+				OnStoppedLeading: func() {
+					// The context being canceled will trigger a handler that will
+					// deal with being deposed.
+					close(deposed)
+					electionCtx.Done()
```
davecheney: Unlike sync.WaitGroup, Done does not signal that the context is done, it returns a channel which will receive a signal when the context is canceled.

cmd/contour/serve.go
```diff
@@ -352,6 +383,99 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+				},
+			},
+		})
+		check(err)
+		// Running the leader election requires a context.Context, but we're
+		// not using them, so we'll just pass in an empty one.
+		le.Run(electionCtx)
+		return nil
```
davecheney: https://github.com/heptio/contour/pull/1358 might help

davecheney:
I'm sorry to come back to the issue of --disable-leader-election. If I run
```
(~/src/github.com/heptio/contour) % contour serve --insecure
INFO[0000] args: [serve --insecure]                     
INFO[0000] started                                       context=contourEventHandler
INFO[0000] started                                       address="0.0.0.0:8000" context=metricsvc
INFO[0000] waiting for cache sync                        context=coreinformers
INFO[0000] started                                       context=coreinformers
INFO[0000] waiting for cache sync                        context=contourinformers
INFO[0000] started                                       context=contourinformers
INFO[0000] started                                       address="127.0.0.1:6060" context=debugsvc
INFO[0000] Started leader election                       configmapname=contour configmapnamespace=heptio-contour context=leaderelect
I0820 16:15:42.370290   29372 leaderelection.go:235] attempting to acquire leader lease  heptio-contour/contour...
INFO[0001] forcing update                                context=contourEventHandler last_update=1.276664652s outstanding=1
I0820 16:15:44.054823   29372 leaderelection.go:245] successfully acquired lease heptio-contour/contour
INFO[0001] elected leader                                identity=2fa7e781-40c8-44e6-8931-a66752afba63 lock=heptio-contour/contour
INFO[0001] started                                       address="127.0.0.1:8001" context=grpc
INFO[0001] performing delayed update                     context=contourEventHandler last_update=463.49089ms outstanding=187
```
It works fine without disabling leadership election. Can we just ship with leadership election always on?

cmd/contour/serve.go
```diff
@@ -352,15 +383,127 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+			<-leadingDisabled
+		}
+
+		log.WithFields(logrus.Fields{
+			"configmapname":      ctx.LeaderElectionConfig.Name,
+			"configmapnamespace": ctx.LeaderElectionConfig.Namespace,
+		}).Info("Started leader election")
+
+		// Running the leader election requires a context.Context, but we're
+		// not using them, so we'll just pass in an empty one.
```
davecheney: comment is out of date

cmd/contour/serve.go
```diff
@@ -352,15 +383,127 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+				// deal with being deposed.
+				close(deposed)
+			},
+		},
+	})
+	check(err)
+	// AddContext will generate its own context.Context and pass it in,
+	// managing the close process when the other goroutines are closed.
+	g.AddContext(func(electionCtx context.Context) {
+		log := log.WithField("context", "leaderelect")
```
davecheney: suggestion: if you call this `context=leaderelection` then

cmd/contour/serve.go
```diff
@@ -352,15 +383,127 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+			// The Workgroup will handle leaving this running until everything
+			// else closes down.
+			close(leaderOK)
+			<-electionCtx.Done()
+		}
+
+		log.WithFields(logrus.Fields{
+			"configmapname":      ctx.LeaderElectionConfig.Name,
+			"configmapnamespace": ctx.LeaderElectionConfig.Namespace,
+		}).Info("Started leader election")
```
davecheney: this becomes `started`, and

cmd/contour/serve.go
```diff
@@ -352,15 +383,127 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+
+		le.Run(electionCtx)
+		log.Info("Finishing leader election")
```
davecheney: that becomes `stopped` which is in keeping with the rest of the members of this group.

PR #1371, opened 2019-08-22
internal/dag: do not trigger update for unreferenced secrets

### LINKED ISSUE #1166: internal/dag: rebuild should not trigger on unrelated secrets

davecheney:
Fixes #1166

Teach dag.Cache.Insert to return false if the Secret being inserted
is not referenced by an Ingress or IngressRoute object in the cache.

The implementation is O(N(ingress)+N(ingressroutes)+O(tlsdelegations)).

### DIFF
--- a/internal/dag/cache.go
+++ b/internal/dag/cache.go
@@ -59,7 +59,7 @@ func (kc *KubernetesCache) Insert(obj interface{}) bool {
 			kc.secrets = make(map[Meta]*v1.Secret)
 		}
 		kc.secrets[m] = obj
-		return true
+		return kc.secretTriggersRebuild(obj)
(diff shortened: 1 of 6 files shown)


davecheney:
This PR cuts the number of updates _significantly_, there are a lot of secrets in k8s that contour doesn't care about.

Issue #1165, opened 2019-06-12
SDS: secret should not be present if the ingress contains no valid routes


davecheney:
Currently if a TLS enabled ingress is present in the DAG and there is a matching secret then SDS will present that secret _even_ if the Ingress has no valid routes.

This should be changed so that SDS will not present a certificate unless the ingress has at least one valid route.

davecheney:
This is a straight forward bug, we're not excluding secure virtual hosts that have no routes. Once they are excluded the secrets visitor will not be able to locate the unneeded secrets.

davecheney:
The fix for this is straight forward -- make sure we exclude SecureVirtualHosts that lack valid routes, but making this change has exposed a lot of issues downstream of the dag. I have to postpone this to p1 in 1.0-beta.1.

PR #1378, opened 2019-08-23
internal/dag: do not create routes if there is no matching service

davecheney:
Updates #1165

When processing Ingress documents the dag.Builder would create routes
even if there was no matching Service available. These empty routes were
then filtered out in the contour.Route visitor.

After this change the empty routes are no longer created at the dag
level and don't need to be filtered out at the visitor level. It also
appears this filtering was not as good as it should have been and there
were cases where listeners were being created when there were routes
without any valid clusters backing them.

### DIFF
--- a/internal/contour/listener_test.go
+++ b/internal/contour/listener_test.go
@@ -149,6 +149,19 @@ func TestListenerVisit(t *testing.T) {
 						},
 					},
 				},
+				&v1.Service{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "kuard",
+						Namespace: "default",
+					},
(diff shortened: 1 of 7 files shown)


davecheney:
For release notes: this change affects virtualhosts which previously presented routes which had no valid clusters. Prior to this change if you had an ingress (not ingressroute) document that pointed to a service which was missing the route would be present in contour but would always return 503. With this change the route will not be presented at all. 

One side effect of this is if this was the only one route on the last virtual host keeping the listener open then after this change contour will not present the listener rather than presenting the listner for the soul purpose of return a 503 because the backend service is misconfigured.

Issue #1361, opened 2019-08-20
cmd/contour: serve workgroup does not exit cleanly


davecheney:
```
diff --git a/cmd/contour/serve.go b/cmd/contour/serve.go
index e8657aa..ce7a47d 100644
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -386,6 +386,12 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
                return s.Serve(l)
        })
 
+       g.Add(func(<-chan struct{}) error {
+               time.Sleep(250 * time.Millisecond)
+               log.WithField("context", "poison").Error("shutting down")
+               return nil
+       })
+
        // step 14. GO!
        return g.Run()
 }
```
This diff should cause `contour serve` to exit almost immediately on startup. It doesn't. One of the group functions is not returning immediately on the stop channel closing.

PR #1382, opened 2019-08-23
cmd/contour: add SIGTERM handler for graceful shutdown

### LINKED ISSUE #1364: cmd/contour: hook SIGHUP to gracefully shut down contour serve
`contour serve` runs until one of the members of `serve`'s workgroup exits, shutting down the rest of the group.

For testing we sometimes manually hack in poison pill style jobs which terminate the group after a period of time #1361. It would be useful to be able to a graceful server shutdown rather than ^C the process. I propose we hook up SIGHUP to terminate the group.

alexbrand:
Adds a SIGTERM handler to `contour serve`.

Fixes #1364

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -538,17 +538,20 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	})
 
 	// step 15. Setup SIGTERM handler
-	g.Add(func(<-chan struct{}) error {
+	g.Add(func(stop <-chan struct{}) error {
 		c := make(chan os.Signal, 1)
 		signal.Notify(c, syscall.SIGTERM)
-		<-c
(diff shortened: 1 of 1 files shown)


alexbrand:
Yeah, my bad! Fixed now.

cmd/contour/serve.go
```diff
@@ -535,8 +537,18 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		return s.Serve(l)
 	})
 
-	// step 15. GO!
+	// step 15. Setup SIGTERM handler
+	g.Add(func(<-chan struct{}) error {
```
davecheney: This worker won't shut down until it receives a signal. Patch in the change from #1361 and check to see if this worker blocks shutdown of the group.

davecheney:
@alexbrand i'll rebase this for you so I can get the PR off the backlog.

PR #1388, opened 2019-08-26
Fix leader election config and TLS config loading of partial values.

### LINKED ISSUE #1367: Partial configuration in the config file is not defaulted correctly for child structs
# What happened
When prototyping putting leader election configuration in the configuration file, I found the following import from YAML did not work as I expected.

For the following struct:
```go
type LeaderElectionConfig struct {
	LeaseDuration time.Duration `yaml:"lease-duration"`
	RenewDeadline time.Duration `yaml:"renew-deadline"`
	RetryPeriod   time.Duration `yaml:"retry-period"`
	Namespace     string        `yaml:"configmap-namespace"`
	Name          string        `yaml:"configmap-name"`
}
```
This struct is included under a YAML field of `leaderelection`. So a full config, set to the defaults set elsewhere in `cmd/contour/serve.go` would look like this:
```yaml
leaderelection:
  lease-duration: 15s
  renew-deadline: 10s
  retry-period: 2s
  configmap-name: contour
  configmap-namespace: heptio-contour
```

This would produce a `LeaderElectionConfig` struct that had everything set correctly.

However, the following snippet:

```yaml
leaderelection:
  configmap-name: contour
  configmap-namespace: heptio-contour

```

Would produce a `LeaderElectionConfig` with only the `Namespace` and `Name` fields set, and the `time.Duration` fields set to their zero value. As there is no further input sanitation in the leader election code, this will produce a panic when used.

# What I expected

I expected that I would only need to set items I wished to change from defaults (in this case, `Name` and `Namespace`), and that defaults would be populated without needing to be in the configuration file.

youngnick:
Fixes #1367

Also add some tests for the YAML parsing of the config file.

Changed those tests to use `github.com/go-test/deep` so I could actually find the errors.

### DIFF
--- a/cmd/contour/servecontext.go
+++ b/cmd/contour/servecontext.go
@@ -1,0 +1,225 @@
+// Copyright © 2019 Heptio
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"crypto/rand"
+	"crypto/tls"
+	"crypto/x509"
+	"errors"
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/heptio/contour/internal/contour"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+)
+
+type serveContext struct {
+	// contour's kubernetes client parameters
+	InCluster  bool   `yaml:"incluster,omitempty"`
+	Kubeconfig string `yaml:"kubeconfig,omitempty"`
+
+	// contour's xds service parameters
+	xdsAddr                         string
+	xdsPort                         int
+	caFile, contourCert, contourKey string
+
+	// contour's debug handler parameters
+	debugAddr string
+	debugPort int
+
+	// contour's metrics handler parameters
+	metricsAddr string
+	metricsPort int
+
+	// ingressroute root namespaces
+	rootNamespaces string
+
+	// ingress class
+	ingressClass string
+
+	// envoy's stats listener parameters
+	statsAddr string
+	statsPort int
+
+	// envoy's listener parameters
+	useProxyProto bool
+
+	// envoy's http listener parameters
+	httpAddr      string
+	httpPort      int
+	httpAccessLog string
+
+	// envoy's https listener parameters
+	httpsAddr      string
+	httpsPort      int
+	httpsAccessLog string
+
+	// PermitInsecureGRPC disables TLS on Contour's gRPC listener.
+	PermitInsecureGRPC bool `yaml:"-"`
+
+	TLSConfig `yaml:"tls,omitempty"`
+
+	// DisablePermitInsecure disables the use of the
+	// permitInsecure field in IngressRoute.
+	DisablePermitInsecure bool `yaml:"disablePermitInsecure,omitempty"`
+
+	EnableLeaderElection bool
+	LeaderElectionConfig `yaml:"leaderelection,omitempty"`
+}
+
+// newServeContext returns a serveContext initialized to defaults.
+// Supply one or more functions to modify the defaults.
+func newServeContext(options ...func(*serveContext)) *serveContext {
+	// Set defaults for parameters which are then overridden via flags, ENV, or ConfigFile
+	ctx := serveContext{
+		Kubeconfig:            filepath.Join(os.Getenv("HOME"), ".kube", "config"),
+		xdsAddr:               "127.0.0.1",
+		xdsPort:               8001,
+		statsAddr:             "0.0.0.0",
+		statsPort:             8002,
+		debugAddr:             "127.0.0.1",
+		debugPort:             6060,
+		metricsAddr:           "0.0.0.0",
+		metricsPort:           8000,
+		httpAccessLog:         contour.DEFAULT_HTTP_ACCESS_LOG,
+		httpsAccessLog:        contour.DEFAULT_HTTPS_ACCESS_LOG,
+		httpAddr:              "0.0.0.0",
+		httpsAddr:             "0.0.0.0",
+		httpPort:              8080,
+		httpsPort:             8443,
+		PermitInsecureGRPC:    false,
+		DisablePermitInsecure: false,
+		EnableLeaderElection:  false,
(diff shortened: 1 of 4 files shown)


cmd/contour/servecontext.go
```diff
@@ -0,0 +1,200 @@
...
+	// PermitInsecureGRPC disables TLS on Contour's gRPC listener.
+	PermitInsecureGRPC bool `yaml:"-"`
+
+	TLSConfig `yaml:"tls,omitempty"`
+
+	// DisablePermitInsecure disables the use of the
+	// permitInsecure field in IngressRoute.
+	DisablePermitInsecure bool `yaml:"disablePermitInsecure,omitempty"`
+
+	EnableLeaderElection bool
```
davecheney: this needs a yaml:"-" to avoid it being exposed to the config file.

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/envoyproxy/go-control-plane v0.8.3
 	github.com/envoyproxy/protoc-gen-validate v0.0.14 // indirect
 	github.com/evanphx/json-patch v4.1.0+incompatible
+	github.com/go-test/deep v1.0.3
```
davecheney: I think this might have leaked in while you were working

cmd/contour/servecontext.go
```diff
@@ -0,0 +1,200 @@
...
+	LeaderElectionConfig `yaml:"leaderelection,omitempty"`
+}
+
+// newServeContext returns a serveContext initialized to defaults.
+// Supply one or more functions to modify the defaults.
+func newServeContext(options ...func(*serveContext)) *serveContext {
```
davecheney: I see what you're doing here passing in a function to change the fields in serveContext for testing, but its odd that this will leak into non text contexts.

I suggest something simpler. The test is only interested in does yaml decoding affect fields which are public -- you don't need to set the ones that yaml can't touch. Rather than passing in a option func style try something like this

want := func() *serveContext {
ctx := newServeContext()
ctx.foo = 1
ctx.bar =2
return ctx
}()

youngnick: Okay, I see. That is simpler, thanks.

cmd/contour/servecontext.go
```diff
@@ -0,0 +1,214 @@
...
+		httpAccessLog:         contour.DEFAULT_HTTP_ACCESS_LOG,
+		httpsAccessLog:        contour.DEFAULT_HTTPS_ACCESS_LOG,
+		httpAddr:              "0.0.0.0",
+		httpsAddr:             "0.0.0.0",
+		httpPort:              8080,
+		httpsPort:             8443,
+		PermitInsecureGRPC:    false,
+		DisablePermitInsecure: false,
+		EnableLeaderElection:  false,
+		LeaderElectionConfig:  *newLeaderElectionConfig(),
```
davecheney: Please don't make this a pointer, make it a value

cmd/contour/servecontext.go
```diff
@@ -0,0 +1,214 @@
...
+type LeaderElectionConfig struct {
+	LeaseDuration time.Duration `yaml:"lease-duration,omitempty"`
+	RenewDeadline time.Duration `yaml:"renew-deadline,omitempty"`
+	RetryPeriod   time.Duration `yaml:"retry-period,omitempty"`
+	Namespace     string        `yaml:"configmap-namespace,omitempty"`
+	Name          string        `yaml:"configmap-name,omitempty"`
+}
+
+// newLeaderElectionConfig returns a defaulted LeaderElectionConfig.
+func newLeaderElectionConfig() *LeaderElectionConfig {
```
davecheney: I suggest dropping this function and doing the config inline
```
LeaderElectionConfig: LeaderElectionConfig {
        LeaseDuration: time.Second * 15,
	RenewDeadline: time.Second * 10,
	RetryPeriod:   time.Second * 2,
	Namespace:     "heptio-contour",
	Name:          "contour",
}

File: cmd/contour/servecontext_test.go [hunk 6]
```diff
@@ -100,3 +104,98 @@ func TestServeContextTLSParams(t *testing.T) {
...
+	tests := map[string]struct {
+		yamlIn []byte
+		want   func() *serveContext
+	}{
+		"empty configuration": {
+			yamlIn: []byte(``),
+			want:   newServeContext,
+		},
+		"defaults in yaml": {
+			yamlIn: []byte(`
```
davecheney: make this a string then convert it to a []byte in in the test function below

toptip: avoid the preceeding newline with this
```
    yamlIn: ""+
`incluster: false
...

File: cmd/contour/servecontext_test.go [hunk 7]
```diff
@@ -100,3 +104,98 @@ func TestServeContextTLSParams(t *testing.T) {
...
+			want: newServeContext,
+		},
+		"tls configuration only": {
+			yamlIn: []byte(`
+tls:
+  minimum-protocol-version: 1.2
+`),
+			want: func() *serveContext {
+				ctx := newServeContext()
+				ctx.TLSConfig = TLSConfig{
```
davecheney: ctx.TLSConfig.MinimumProtocolVersion = 1.2

cmd/contour/servecontext_test.go
```diff
@@ -100,3 +104,98 @@ func TestServeContextTLSParams(t *testing.T) {
...
+				ctx.LeaderElectionConfig.RenewDeadline = 500 * time.Second
+				ctx.LeaderElectionConfig.RetryPeriod = 60 * time.Second
+				return ctx
+			},
+		},
+	}
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			got := newServeContext()
+			err := yaml.Unmarshal(tc.yamlIn, got)
```
davecheney: do the []byte(string) conversion here.

cmd/contour/servecontext_test.go
```diff
@@ -100,3 +104,98 @@ func TestServeContextTLSParams(t *testing.T) {
...
+			checkErr(t, err)
+			want := tc.want()
+
+			if diff := cmp.Diff(*want, *got, cmp.AllowUnexported(serveContext{})); diff != "" {
```
davecheney: Maybe we should just export ServeContext. It's in `package main` so nobody can import it anyway. If we export it then the comments will show up in godoc which is probably a good thing.

cmd/contour/servecontext_test.go
youngnick: If I do this, then goimports insists that the first line of YAML must be indented.
I don't think the beginning newline is hurting anything, and it makes the inlined YAML much more readable.

cmd/contour/servecontext_test.go
youngnick: AllowUnexported is actually talking about allowing comparison of the unexported fields of `serveContext` here.
Not sure about changing all of them to be exported just for this test case.

davecheney:
No worries. But beware some formats (yaml is ok) crap out on the preceding newline.

cmd/contour/servecontext_test.go
davecheney: true, ok, skip this and lets table the discussion of the value of go-cmp/cmp until after 1.0.

PR #1391, opened 2019-08-27
internal/dag: filter out secure virtual hosts without routes

### LINKED ISSUE #1165: SDS: secret should not be present if the ingress contains no valid routes

davecheney:
Fixes #1165

Filter out SecureVirtualHost objects that have a secret but no routes.
This will exclude the secret attached to those objects from SDS until
there are valid routes for that vhost.

### DIFF
--- a/internal/dag/dag.go
+++ b/internal/dag/dag.go
@@ -186,8 +186,10 @@ func (s *SecureVirtualHost) Visit(f func(Vertex)) {
 }
 
 func (s *SecureVirtualHost) Valid() bool {
-	// A SecureVirtualHost is valid if it has a Secret or a TCPProxy.
-	return s.Secret != nil || s.TCPProxy != nil
+	// A SecureVirtualHost is valid if either
+	// 1. it has a secret and at least one route.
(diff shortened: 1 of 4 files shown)


internal/dag/dag.go
```diff
@@ -186,8 +186,10 @@ func (s *SecureVirtualHost) Visit(f func(Vertex)) {
 }
 
 func (s *SecureVirtualHost) Valid() bool {
-	// A SecureVirtualHost is valid if it has a Secret or a TCPProxy.
-	return s.Secret != nil || s.TCPProxy != nil
+	// A SecureVirtualHost is valid if either
+	// 1. it has a secret and at least one route.
+	// 2. it has a tcpproxy, because the cpproxy backend may negotiate TLS inself.
```
youngnick: Nit: 'itself'

PR #1398, opened 2019-08-28
internal/contour: sort endpoints within a ClusterLoadAssignment

davecheney:
Fixes ##1159

Simplify the conversion of Endpoint document to ClusterLoadAssignment
set. This simplification makes it easier to sort the list of endpoints
without a ClusterLoadAssignment. ClusterLoadAssignments are sorted
themselves on the way out of EndpointTranslator.Query and Contents.

The simplification takes advantage of the fact that a service namespace,
name, port and port name tupple only appers once per EndpointSet. The
same port and port name (if there is more than one port) cannot appear
on the service document more than once therefore for each endpoint
subset, we'll never see the same port mentioned more than once.


davecheney:
@youngnick @stevesloka please review carefully, the Endpoint spec competes with the Ingress spec for vagueness.

PR #1400, opened 2019-08-28
Upgrade document suggestions

rbankston:
The document looks really really comprehensive. The only thing I wanted to call out was a pattern I have seen folks use of updating the namespace to something custom for architecture reasions. My suggestion was to call out that specific use case.


youngnick:
I did consider putting something like that in there already, but felt that it wasn't actually 'easy' enough any more.

It is pretty important that the whole set of YAMLs be applied this time, so if people have changed the namespace, they'll need to global search-and-replace `heptio-contour` in their chosen example. So I figured that they would be able to do that for the commands too.

rbankston:
Hey @youngnick I tried to do something similar will see if I can wordsmith something better around the search and replace possibly just a note at the bottom that it also applies to the easy way.

davecheney:
Also, please sign your commit, `git commit --amend -s && git push -f`. Thanks.

davecheney:
@rbankston ping!

rbankston:
Hey @davecheney sorry for the delays. Updated to a note at the bottom of the easy way to use search and replace in the editor.

davecheney:
Thanks for the update but I would prefer not to add addenda to the easy way — it’s not easy if you have to think about the namespace a thing just deployed into. 

Instead, what if we called out that the namespace is hard coded in the not so easy way section and left it to the reader to figure out what to do. Alternatively we could remove the namespace from the manifests and suggest a -n flag in our installation instructions.

rbankston:
@davecheney something like "These example deployments all deploy into the heptio-contour namespace by default"?

youngnick:
Merging as discussed with @davecheney

PR #1404, opened 2019-08-29
internal/dag: exclude any secrets other than tls and CA cert from cache

### LINKED ISSUE #1372: internal/dag: KubernetesCache should ignore non kubernetes.io/tls type secrets
Post #1371 `dag.KubernetesCache` stores, but does not notify, all `v1.Secret`s. Contour does not use non `kubernetes.io/tls` type secrets so we don't need to cache them at all.

davecheney:
Fixes #1372

Do not store secrets in the cache unless they are tls or CA certs.
Determining if a CA cert is referenced is too complex so any update to a
CA cert will trigger a rebuild.

### DIFF
--- a/internal/dag/cache.go
+++ b/internal/dag/cache.go
@@ -54,6 +54,11 @@ type Meta struct {
 func (kc *KubernetesCache) Insert(obj interface{}) bool {
 	switch obj := obj.(type) {
 	case *v1.Secret:
+		if _, hasCA := obj.Data["ca.crt"]; obj.Type != v1.SecretTypeTLS && !hasCA {
+			// ignore everything but kubernetes.io/tls secrets
+			// and secrets with a ca.crt key.
+			return false
+		}
 		m := Meta{name: obj.Name, namespace: obj.Namespace}
 		if kc.secrets == nil {
 			kc.secrets = make(map[Meta]*v1.Secret)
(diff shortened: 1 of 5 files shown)


internal/dag/cache.go
```diff
@@ -54,6 +54,11 @@ type Meta struct {
 func (kc *KubernetesCache) Insert(obj interface{}) bool {
 	switch obj := obj.(type) {
 	case *v1.Secret:
+		if _, hasCA := obj.Data["ca.crt"]; obj.Type != v1.SecretTypeTLS && !hasCA {
```
stevesloka: We should verify we have this rule documented somewhere. I know `kubernetes.io/tls` secrets aren't even well documented, so I could see folks creating generic secrets with the proper key (e.g. `ca.crt`) and having them be ignored.

davecheney: >  so I could see folks creating generic secrets with the proper key (e.g. ca.crt) and having them be ignored.

I think the logic is the opposite. If I created a secret that had a key `ca.crt` then every time it was touched it would trigger a dag rebuild.

davecheney:
> LGTM, although I'm not sure of how much a standard ca.crt is for the CA key name. Maybe worth documenting that somehwere?

I've raised https://github.com/heptio/contour/issues/1415

internal/dag/cache.go
stevesloka: OH right, I misread that.

davecheney: It is super subtle, and kind of annoying that we didn't also guard this with a Secret Type so we don't have to rely on the presence of the key.

PR #1417, opened 2019-08-30
go.mod,Dockerfile: upgrade to Go 1.13rc2

davecheney:
Updates #1137


stevesloka:
Just curious, why are we following the RC builds of 1.13 so closely?

davecheney:
Its a way of getting access to go 1.13 -- and catching problems with
it -- now, rather than in a few weeks. The release candidates bake
inside google for a few weeks before being released to the world so
they aren't fancy words for beta, they're the real deal. Once go 1.13
is released, the upgrade should be seamless.

PR #1424, opened 2019-09-02
internal: Initial implementation of HTTPLoadBalancer in builder + dag

stevesloka:


davecheney:
@stevesloka i've started to pull this apart into some smaller PRs, eg  https://github.com/heptio/contour/pull/1432

I'll keep working on this tomorrow.

apis/projectcontour/v1alpha1/httploadbalancer.go
```diff
@@ -78,6 +78,8 @@ type TLS struct {
 
 // Route contains the set of routes for a virtual host
 type Route struct {
+	// Includes allow for specific routing configuration to be appended to another HTTPLoadBalancer in another namespace
+	Includes []Include `json:"includes,omitempty"`
```
davecheney: I don't think this should be here -- a route doesn't include another route, routes (plural) can be included from another httplb document at the spec level.

Issue #1385, opened 2019-08-26
Leader election should be multi-read, single write style


youngnick:
The current implementation of leader election in Contour is as follows:
- Contour pods elect a leader using a Configmap as the distributed lock. (Hard-coded to `contour` in `heptio-contour`).
- All pods block before opening the gRPC port until one becomes leader.
- In order to ensure that only the leader actually serves traffic, the Pod's readinessProbe is configured to look for the gRPC  port.

Importantly, this means that non-leader Pods are not Ready (as far as Kubernetes is concerned), which ensures that their Endpoint is not present in the Endpoints that the Contour Service uses, and so no traffic from the Envoy Daemonset arrives at non-leader xDS ports. If this was not the case, and we used the metrics port for readinessProbes (as we do for livenessProbes), then Envoy would fail at connecting to Contour n-1 out of n times. (where n is the number of Contour pods in the Deployment).

The problems with this current method are:
- The Contour Deployment can't use the RollingUpdate strategy (as that requires that Pods become Ready for the Deployment to complete its rolling update).
- Kubernetes tooling assumes that most Pods will be Ready most of the time, and that unReady pods are an exceptional condition that should be looked at.

This ticket covers moving the leader election pattern to the same pattern used by the core Kubernetes components like the apiserver:
- n Pods run, and serve information from their serving endpoints
- only the leader writes back changes to controlled objects.

In the case of Contour these changes will mainly be status updates and events.

The downside of moving to this model is that the current single-reader, single-writer model ensures that all Envoys will receive the same messages in the same order, preventing partial configuration states at the Envoy end. I anticipate that #1280 will help to mitigate this in the future.

Blocked

- [ ] #1425

youngnick:
This work is currently in two parts:
- Changing the Status to be more compatible with this idea (#1425 )
- Setting up leader election to no longer block the gRPC server from opening its port.

I'm covering the latter in this issue.

youngnick:
This is blocked at the moment by the StatusWriter work in #1425 - I'm unassigning myself and leaving this one to @davecheney once he's finished there.

davecheney:
There are too many other higher priority issues ahead of this one. Bumping to rc.1

davecheney:
@youngnick I don't want to block this issue on #1425. If leader election is working -- a leader is elected amongst running contours -- and we've reverted the deployment changes that caused use to move away from `RollingUpdate`, then I'm fine to close this issue. 

Can you please action this soon. Thanks

[...]

PR #1428, opened 2019-09-02
Update leader election in prep for changing to multi-read, single-write

youngnick:
Updates #1385 

- Leader election on by default.
- Leader election configmap now `leader-elect` not `contour` (will be created if it doesn't exist)
- Added RBAC for HttpLoadBalancer CRDs.
- Updated `contour` deployment.

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -113,7 +113,7 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 	serve.Flag("envoy-service-https-port", "Kubernetes Service port for HTTPS requests").IntVar(&ctx.httpsPort)
 	serve.Flag("use-proxy-protocol", "Use PROXY protocol for all listeners").BoolVar(&ctx.useProxyProto)
 
-	serve.Flag("enable-leader-election", "Enable leader election mechanism").BoolVar(&ctx.EnableLeaderElection)
+	serve.Flag("disable-leader-election", "Enable leader election mechanism").BoolVar(&ctx.DisableLeaderElection)
 	return serve, ctx
 }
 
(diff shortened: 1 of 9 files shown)


cmd/contour/serve.go
```diff
@@ -113,7 +113,7 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 	serve.Flag("envoy-service-https-port", "Kubernetes Service port for HTTPS requests").IntVar(&ctx.httpsPort)
 	serve.Flag("use-proxy-protocol", "Use PROXY protocol for all listeners").BoolVar(&ctx.useProxyProto)
 
-	serve.Flag("enable-leader-election", "Enable leader election mechanism").BoolVar(&ctx.EnableLeaderElection)
+	serve.Flag("disable-leader-election", "Enable leader election mechanism").BoolVar(&ctx.DisableLeaderElection)
```
davecheney: please fix the description to match the flag

cmd/contour/serve.go
```diff
@@ -300,10 +300,9 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	// managing the close process when the other goroutines are closed.
 	g.AddContext(func(electionCtx context.Context) {
 		log := log.WithField("context", "leaderelection")
-		if !ctx.EnableLeaderElection {
+		if ctx.DisableLeaderElection {
```
davecheney: Please add a TODO that we can avoid setting up leadership election if its disable; just close leader OK and don't call AddContext.

davecheney: Same as line 312 which I can't add a comment on because github.

cmd/contour/serve.go
```diff
@@ -353,33 +352,26 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
-			log.Info("started")
-			defer log.Info("stopped")
-
-			go func() {
-				<-stop
-				s.Stop()
-			}()
-
-			return s.Serve(l)
+		addr := net.JoinHostPort(ctx.xdsAddr, strconv.Itoa(ctx.xdsPort))
```
davecheney: please update the comment on line 343

cmd/contour/servecontext.go
```diff
@@ -80,8 +80,8 @@ type serveContext struct {
 	// permitInsecure field in IngressRoute.
 	DisablePermitInsecure bool `yaml:"disablePermitInsecure,omitempty"`
 
-	// EnableLeaderElection should only be set by command line flag.
-	EnableLeaderElection bool `yaml:"-"`
+	// DisableLeaderElection should only be set by command line flag.
```
davecheney: s/should/can/

examples/common/rbac.yaml
```diff
@@ -58,6 +58,15 @@ rules:
   - put
   - post
   - patch
+- apiGroups: ["projectcontour.io"]
```
davecheney: looks unrelated

cmd/contour/serve.go
youngnick: Oops.

PR #1433, opened 2019-09-02
Use Envoy's /ready endpoint for readiness probes

### LINKED ISSUE #1277: Envoy v1.11: Investigate using `/ready` endpoint for probes
Mentioned in #1242, Envoy v1.11.0 introduces a new /ready admin endpoint which improves readiness probes, this might be a nice thing to change for Contour 0.15 too, maybe even improving smoother Envoy rollouts as discussed on #1245 

// cc @rochacon

rochacon:
This PR enables Envoy's 1.11+ `/ready` endpoint for readiness probes.

Closes #1277

### DIFF
--- a/examples/contour/03-envoy.yaml
+++ b/examples/contour/03-envoy.yaml
@@ -56,7 +56,7 @@ spec:
           protocol: TCP
         readinessProbe:
           httpGet:
-            path: /healthz
+            path: /ready
             port: 8002
           initialDelaySeconds: 3
           periodSeconds: 3
--- a/examples/deployment-grpc-v2/02-contour.yaml
+++ b/examples/deployment-grpc-v2/02-contour.yaml
@@ -69,7 +69,7 @@ spec:
               fieldPath: metadata.name
         readinessProbe:
           httpGet:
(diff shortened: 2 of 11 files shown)


ravilr:
we use the envoy.health_check filter today to achieve graceful connection draining during envoy shutdown/rolling deploys, i.e. issue an /healthcheck/fail and then wait for connections to drain.
Does /ready admin endpoint have the same behavior ?

rochacon:
@stevesloka I've left he filter on purpose, in the intent to use the `/ready` only for the `readinessProbe`, but given it is in the same listener it doesn't make sense to keep it. I'll test the change and update the PR later today.

@ravilr yes, both the health check filter and [`/ready`](https://www.envoyproxy.io/docs/envoy/v1.11.1/operations/admin.html?highlight=ready#get--ready) behave quite similar.

stevesloka:
Thanks @rochacon, I think if we switch to `/ready`, there's no need for the `healthz` filter to be applied. One less thing to configure. =)

davecheney:
Nick, would you please merge this for the OP

PR #1435, opened 2019-09-02
Add RBAC for new CRDs

youngnick:
Note that having the ClusterRole reference something that doesn't exist should be okay.

Noticed when I applied the new CRDs to my GKE cluster.


examples/ds-hostnet-split-httploadbalancer/02-rbac.yaml
```diff
@@ -58,6 +58,15 @@ rules:
   - put
   - post
   - patch
+- apiGroups: ["projectcontour.io"]
```
davecheney: Is this file a symlink to ..common/rbac.yaml and github isn't showing me that? If not, could it be made so?

youngnick: It is not, and definitely could be.

davecheney:
Also, you need to run the `./examples/render.sh` step

PR #1437, opened 2019-09-03
Dockerfile: upgrade to Go 1.13

### LINKED ISSUE #1377: Upgrade to Go 1.13
Go 1.13 will be released in the next 30 days, we should upgrade to 1.13rc1 next monday for the start of 1.0-beta.1 and prepare for 1.13 final for the 1.0-beta.1 release.

Blocked:
- [x] https://github.com/golang/go/issues/34054

davecheney:
Fixes #1377

BREAKING CHANGE

We build with -mod=readonly so that the build will fail if anything is
out of date in our go.mod file. golang/go#30667 changed the behaviour of
the go tool to require that go.mod AND go.sum be up to date so this
means I cannot use my hack of .gitignoring the go.sum file.

This means when you pull this commit you will probably have to delete
your local go.sum working copy. If we're lucky then you shouldn't need
to change it after that. If we're not lucky, the bloody thing will
always be dirty and we'll waste a lot of time telling people to revert
their changes to this file. C'est la vie.


davecheney:
Won't pass til Go 1.13 has been released and travis and docker have updated images. ETA for Go 1.13 is before the end of the week.

Issue #1348, opened 2019-08-16
IngressRoute that references a secret which has not been delegated is deemed valid


alexbrand:
**What steps did you take and what happened:**
1. Install contour:
```shell
kubectl apply -f https://raw.githubusercontent.com/heptio/contour/v0.14.2/examples/render/deployment-rbac.yaml
```

2. Create a TLS secret in the `heptio-contour` namespace

3. Create an IngressRoute outside of the `heptio-contour` namespace. In its TLS block, the IngressRoute should reference the secret that was created in the `heptio-contour` namespace. For example:
```yaml
apiVersion: contour.heptio.com/v1beta1
kind: IngressRoute
metadata:
  name: app-with-tls-delegation
  namespace: app-with-tls-delegation
spec:
  virtualhost:
    fqdn: app-with-tls-delegation.127.0.0.1.nip.io
    tls:
      secretName: heptio-contour/ssl-cert
  routes:
  - match: '/'
    services:
    - name: app-with-tls-delegation
      port: 80
```

4. Observe IngressRoute status:
```shell
$ kubectl get ingressroute
NAME                      FQDN                                       TLS SECRET                FIRST ROUTE   STATUS   STATUS DESCRIPTION
app-with-tls-delegation   app-with-tls-delegation.127.0.0.1.nip.io   heptio-contour/ssl-cert   /             valid    valid IngressRoute
```

**What did you expect to happen:**
The IngressRoute should be deemed invalid because the referenced TLS certificate is in another namespace, and there is no TLSCertificateDelegation resource granting access to the cert.

**Anything else you would like to add:**
By inspecting the DAG, it seems like the HTTPS listener is not setup:

```
digraph DAG {
rankdir="LR"
"0xc0006671a0" [shape=record, label="{listener|:80}"]
"0xc0006671c0" [shape=record, label="{http://app-with-tls-delegation.127.0.0.1.nip.io}"]
"0xc0000afc70" [shape=record, label="{prefix|/}"]
"0xc000500dc0" [shape=record, label="{cluster|{app-with-tls-delegation/app-with-ef7df1/80/da39a3ee5e|weight 0}}"]
"0xc00012d9d0" [shape=record, label="{httpservice|app-with-tls-delegation/app-with-tls-delegation:80}"]
"0xc000500dc0" -> "0xc00012d9d0"
"0xc0000afc70" -> "0xc000500dc0"
"0xc0006671c0" -> "0xc0000afc70"
[...]

PR #1442, opened 2019-09-03
internal/dag: report certificate delegation failure correctly

### LINKED ISSUE #1347: TcpProxy IngressRoute status is not set when it references an un-delegated TLS cert
1. Create a secret in the `heptio-contour` namespace called `default-ssl-cert`.

2. Create an IngressRoute that references the secret:
```yaml
kind: IngressRoute
metadata:
  name: sample-app
  namespace: tls-cert-delegation
spec:
  virtualhost:
    fqdn: 127.0.0.1.nip.io
    tls:
      secretName: heptio-contour/default-ssl-cert
  tcpproxy:
    services:
    - name: sample-app
      port: 80
```

3. Get the IngressRoute and see status:
```shell
$ kubectl get ingressroute
NAME         FQDN               TLS SECRET                        FIRST ROUTE   STATUS   STATUS DESCRIPTION
sample-app   127.0.0.1.nip.io   heptio-contour/default-ssl-cert
```

The status should reflect that the user does not have access to the referenced certificate. To avoid information disclosure, perhaps the message should be something like: "The referenced certificate does not exist, or you do not have access to it"

**Environment:**

- Contour version: v0.14.1
```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.0", GitCommit:"641856db18352033a0d96dbc99153fa3b27298e5", GitTreeState:"clean", BuildDate:"2019-03-26T00:04:52Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.0", GitCommit:"e8462b5b5dc2584fdcd18e6bcfe9f1e4d970a529", GitTreeState:"clean", BuildDate:"2019-06-25T23:41:27Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
```

davecheney:
Fixes #1347

### DIFF
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -389,26 +389,7 @@ func (b *Builder) computeIngressRoutes() {
 			continue
 		}
 
-		var enforceTLS, passthrough bool
-		if tls := ir.Spec.VirtualHost.TLS; tls != nil {
-			// attach secrets to TLS enabled vhosts
-			m := splitSecret(tls.SecretName, ir.Namespace)
-			sec := b.lookupSecret(m, validSecret)
-			if sec != nil && b.delegationPermitted(m, ir.Namespace) {
-				svhost := b.lookupSecureVirtualHost(host)
-				svhost.Secret = sec
-				svhost.MinProtoVersion = MinProtoVersion(ir.Spec.VirtualHost.TLS.MinimumProtocolVersion)
-				enforceTLS = true
-			}
-			// passthrough is true if tls.secretName is not present, and
-			// tls.passthrough is set to true.
-			passthrough = isBlank(tls.SecretName) && tls.Passthrough
-
-			// If not passthrough and secret is invalid, then set status
-			if sec == nil && !passthrough {
-				b.setStatus(Status{Object: ir, Status: StatusInvalid, Description: fmt.Sprintf("TLS Secret [%s] not found or is malformed", tls.SecretName)})
-			}
-		}
+		enforceTLS, passthrough := b.configureSecureVirtualHost(ir)
 
 		switch {
 		case ir.Spec.TCPProxy != nil && (passthrough || enforceTLS):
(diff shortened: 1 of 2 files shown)


davecheney:
@alexbrand this is the current behaviour for #1347. Could you please review and let me know it differs from what you expect. Thank you.

alexbrand:
My bad @davecheney. It looks like the issue only happens when the certificate exists in the other namespace. I missed that in the reproduction steps.

davecheney:
@alexbrand i believe I have fixed the issue, PTAL.

internal/dag/builder.go
```diff
@@ -354,11 +354,15 @@ func (b *Builder) computeIngressRoutes() {
...
-				svhost.MinProtoVersion = MinProtoVersion(ir.Spec.VirtualHost.TLS.MinimumProtocolVersion)
-				enforceTLS = true
+			if sec != nil {
+				if b.delegationPermitted(m, ir.Namespace) {
+					svhost := b.lookupSecureVirtualHost(host)
+					svhost.Secret = sec
+					svhost.MinProtoVersion = MinProtoVersion(ir.Spec.VirtualHost.TLS.MinimumProtocolVersion)
+					enforceTLS = true
+				} else {
+					b.setStatus(Status{Object: ir, Status: StatusInvalid, Description: fmt.Sprintf("%s: certificate delegation not permitted", tls.SecretName)})
```
alexbrand: Do we need a `continue` after this, to stop processing the IngressRoute?

alexbrand:
I believe this will also fix #1348

internal/dag/builder.go
davecheney: technically yes, but the next condition, `sec == nil` will prevent the other 1/2 of this function running. Having said that, this is a horrid piece of nested logic. Let me see if I can clean it up some so we can use `continue` or `return` after a status setting.

Issue #1434, opened 2019-09-02
Migrate :master image building from jenkins to github actions


davecheney:
Currently our `:master` and tagged (see [./docs/tagging.md]) builds are built using an internal jenkins instance that is going away soon. We need to move the service that builds our `:master` and release builds elsewhere. Additionally, our gcr.io image repository contains heptio branding so we need to move away from that.

TODO:
- [x] register `projectcontour` on the docker hub, figure out how to distribute write privs to the team
- [x] set up a `on commit to master` build that builds `docker.io/projectcontour/contour:master`
- [x] figure out how to build from a tag and tag an image as `:latest`
- [ ] grep through all the documentation and update all the `gcr.io` images
- [x] update release documentation (on internal wiki -- ask @davecheney) - BONUS POINTS will be awarded if the release documentation is moved to this repo.

Out of scope:

- [ ] Do not disable jenkins jobs, they can continue to push to gcr.io, 0.15 will be the last image we publish there.

youngnick:
Got the first two parts done with an Action, although testing leaves something to be desired.

Looks like it's possible to trigger builds on release creation as well, so that should help. I've requested access to the beta on my account, so once I have that I can do some testing with other repos to try and nail down how the release thing works.

youngnick:
In the meantime, I will do a PR and get rid of all the gcr.io images and change them to docker hub ones.

PR #1445, opened 2019-09-04
Move gcr.io/heptio/ to projectcontour/ for master only

youngnick:
Updates #1434 


davecheney:
I'd like to wait on this a few days to make sure the \:master builds are working well.

youngnick:
No problem. Let's leave it to soak then.

PR #1447, opened 2019-09-05
internal/dag: Add support for include with HTTPLoadBalancers

stevesloka:
This adds support for `Includes` in a Root `HTTPLoadBalancer`. Currently, only includes support for path prefix.


apis/projectcontour/v1alpha1/httploadbalancer.go
```diff
@@ -79,7 +79,7 @@ type TLS struct {
 // Route contains the set of routes for a virtual host
 type Route struct {
 	// Condition defines additional routing parameters on the route
-	Condition `json:"condition,omitempty"`
+	*Condition `json:"condition,omitempty"`
```
davecheney: nit: embedding is cool, but this is probably better written in this instance as
```
type Route struct {

       Condition *Condition
}
```
this will prevent the compiler letting us call Route.(some field on Condition)

internal/dag/builder.go
```diff
@@ -889,11 +865,21 @@ func (b *Builder) processTCPProxy(ir *ingressroutev1.IngressRoute, visited []*in
...
-func conditionPath(c projcontour.Condition) string {
-	if c.Prefix == "" {
+func conditionPath(routeCondition, includeCondition *projcontour.Condition) string {
+	pathPrefix := ""
+
+	if includeCondition != nil {
+		pathPrefix = includeCondition.Prefix
+	}
+	if routeCondition != nil {
+		pathPrefix += routeCondition.Prefix
```
davecheney: suggestion, use path.Join() not += because the former will deal with trailing / properly.

internal/dag/builder_test.go
```diff
@@ -1517,7 +1517,7 @@ func TestDAGInsert(t *testing.T) {
 				Fqdn: "example.com",
 			},
 			Routes: []projcontour.Route{{
-				Condition: projcontour.Condition{
+				Condition: &projcontour.Condition{
```
davecheney: I think this would work ok if Condition wasn't, uh, conditional. As in every Route should have a condition, given the only condition you can have is Prefix at the moment. 

In the future when we make Condition a set, []Condition, then it will be possible to create a route that has no conditions -- matches everything, and we'll have to figure out if we want to permit that, but even then the type will be
```
type Route struct {
    Conditions []Condition // not []*Condition
}
```

stevesloka: Right now I have it such that you don't need to specify a condition. If you're delegated `/blog` for instance, seems redundant to have to specify `/` in the child route.

stevesloka: Having multiple conditions on the same route would need some discussion. I'm not sure how you would handle multiple paths in the same route.

PR #1451, opened 2019-09-05
Add workflows for dealing with tagged releases

youngnick:
Updates #1434


.github/workflows/build_tag_no_latest.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+on:
+  push:
+    tags:
+      - 'v[0-9]*.[0-9]*.[0-9]*beta*'
+      - 'v[0-9]*.[0-9]*.[0-9]*alpha*'
+      - 'v[0-9]*.[0-9]*.[0-9]*rc*'
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
stevesloka: Should we pick a specific version to rule out potential `latest` issues?

.github/workflows/build_tag_no_latest.yaml
```diff
@@ -0,0 +1,22 @@
+name: Build and push a release without :latest
+
+on:
+  push:
+    tags:
+      - 'v[0-9]*.[0-9]*.[0-9]*beta*'
```
davecheney: if this is a regex it might be better to write this as 
```
^v[\d]+\.[\d]+\.[\d]+(-(alpha|beta|rc)\.[\d]+)$
```

.github/workflows/build_tag_with_latest.yaml
```diff
@@ -0,0 +1,23 @@
+name: Build and push a release with :latest
+
+on:
+  push:
+    tags:
+      - 'v[0-9]*.[0-9]*.[0-9]'
+      - 'v[0-9]*.[0-9]*.[0-9][0-9]'
+      - 'v[0-9]*.[0-9]*.[0-9][0-9][0-9]'
```
davecheney:
```
^v[\d]+\.[\d]+\.[\d]+$
```

.github/workflows/build_tag_no_latest.yaml
youngnick: I don't think so at the moment, this is only for doing Docker builds, so all the actual building is done in the go Docker image anyway.

PR #1463, opened 2019-09-05
RateLimiting implementation

### LINKED ISSUE #370: [feature] Add rate limiting support

zxvdr:
Fixes #370 

### DIFF
--- a/apis/contour/v1beta1/ingressroute.go
+++ b/apis/contour/v1beta1/ingressroute.go
@@ -48,6 +48,8 @@ type Route struct {
 	TimeoutPolicy *projcontour.TimeoutPolicy `json:"timeoutPolicy,omitempty"`
 	// // The retry policy for this route
 	RetryPolicy *projcontour.RetryPolicy `json:"retryPolicy,omitempty"`
+	// RateLimitConfiguration defines ratelimiting options which are passed to the rate limiting implementation service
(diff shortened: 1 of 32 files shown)


davecheney:
@zxvdr Thank you for your contribution, however this appears to be #873 pushed as your own branch. I'm going to close this as its probably an error. If you would like to discuss rate limiting please do so on #873 or #843

zxvdr:
@davecheney, it wasn't an error. [I asked](https://kubernetes.slack.com/archives/C8XRH2R4J/p1562879190127800) @stevesloka the status of #873 and offered to help get it over the line. He suggested opening a new PR. This is #873 rebased with most of the previous review comments addressed. Apologies for omitting these details earlier.

PR #1466, opened 2019-09-06
Add release documentation.

youngnick:
Updates #1434.


docs/making-a-release.md
```diff
@@ -0,0 +1,85 @@
...
+```sh
+% git checkout -b release-0.15
+```
+
+If you are doing a patch release on an existing branch, skip this step and just checkout the branch instead.
+
+This branch is used for all the release types, alpha through final.
+Each on of the release types is just a different tag on the release branch./
+
+## Alpha, beta, and RC releases
```
davecheney: \## Alpha, beta, and release candidates

PR #1467, opened 2019-09-06
Update troubleshooting instructions for split-by-default.

### LINKED ISSUE #1416: Troubleshooting document needs to be updated for split deployment
The troubleshooting document is biased towards the previous non split deployment. We should update the documentation to bias towards split deployments.

/cc @youngnick

youngnick:
Fixes #1416 

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -9,45 +9,50 @@ For example, if you have not configured any TLS ingress objects then Contour doe
 Because the HTTP and HTTPS listeners both use the same code, if you have no ingress objects deployed in your cluster, or if no ingress objects are permitted to talk on HTTP, then Envoy does not listen on port 8080 (80 in the service object).
 
 To test whether Contour is correctly deployed you can deploy the kuard example service:
-```
-$ kubectl apply -f https://j.hept.io/contour-kuard-example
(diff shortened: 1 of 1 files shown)


docs/troubleshooting.md
```diff
@@ -59,49 +64,24 @@ The following is an example of a DAG that maps `http://kuard.local:80/` to the
...
-```
+Because Contour secures its communications with Envoy using Secrets in the cluster, the easiest way is to run `contour cli` commands _inside_ the pod.
+Do this is via `kubectl exec`:
+
+```sh
 # Get one of the pods that matches the examples/daemonset
 CONTOUR_POD=$(kubectl -n heptio-contour get pod -l app=contour -o jsonpath='{.items[0].metadata.name}')
 # Do the port forward to that pod
-kubectl -n heptio-contour exec $CONTOUR_POD -c contour contour cli lds
+kubectl -n heptio-contour exec $CONTOUR_POD -c contour -- contour cli lds --cafile=/ca/cacert.pem --cert-file=/certs/tls.crt --key-file=/certs/tls.key
```
davecheney: this isn't so bad, we can live with it, especially as people can copy paste from here

PR #1471, opened 2019-09-07
internal/dag: Allow for multiple levels of includes

### LINKED ISSUE #1458: internal/dag: Detect delegation cycle within an httplb
If an httlb contains an include which adds `Conditions` to another httplb. A cucle can occur if that delegated httplb adds an include which points back to the original httlb. 

Similar check in IngressRoute: https://github.com/heptio/contour/blob/master/internal/dag/builder.go#L697-L702

### LINKED ISSUE #1470: internal/dag: HTTPLoadBalancer doesn't delegate multiple levels
If I have an httlb which passes an include off to another httplb, which then delegates again, the 3rd level won't get processed.

stevesloka:
Fixes #1470 by allowing multiple levels of includes to be processed within the builder. Also fixes #1458 by checking for a delegation cycle with Includes. 

### DIFF
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -253,11 +253,8 @@ func (b *Builder) validHTTPLoadBalancers() []*projcontour.HTTPLoadBalancer {
 			sort.Strings(conflicting) // sort for test stability
 			msg := fmt.Sprintf("fqdn %q is used in multiple HTTPLoadBalancers: %s", fqdn, strings.Join(conflicting, ", "))
 			for _, httplb := range httplbs {
-				//b.setStatus(Status{Object: httplb, Status: StatusInvalid, Description: msg, Vhost: fqdn})
-				// TODO (sas) Make Status work with both IngressRoutes & HTTPLoadBalancers
-				fmt.Printf("Name: %s Namespace: %s Status: %s Description: %s VHost: %s\n", httplb.Name, httplb.Namespace, StatusInvalid, msg, fqdn)
+				b.setStatus(Status{Object: httplb, Status: StatusInvalid, Description: msg, Vhost: fqdn})
 			}
-
 		}
 	}
 	return valid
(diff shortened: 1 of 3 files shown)


internal/dag/builder.go
```diff
@@ -493,6 +484,37 @@ func (b *Builder) computeHTTPLoadBalancers() {
 	}
 }
 
+// mergeConditions merges any two conditions when they are delegated
+func mergeConditions(delegatedCondition, includeCondition *projcontour.Condition) *projcontour.Condition {
```
davecheney: suggestion: the signature of this function, and its name tell us it's working with conditions. I think its probably less to read to just write
```
func mergeConditions(delegate, include *projcontour.Condition) *projcontour.Condition
```

internal/dag/builder.go
```diff
@@ -493,6 +484,37 @@ func (b *Builder) computeHTTPLoadBalancers() {
...
+	if delegatedCondition == nil {
+		return includeCondition
+	}
+
+	returnCondition := delegatedCondition.DeepCopy()
+	returnCondition.Prefix = delegatedCondition.Prefix + includeCondition.Prefix
+	return returnCondition
```
davecheney: if the `Condition` suffix goes away maybe calling this `result` would be better, and avoid the conflict with the `return` keyword.

internal/dag/builder.go
```diff
@@ -493,6 +484,37 @@ func (b *Builder) computeHTTPLoadBalancers() {
...
+			switch {
+			//	case ir.Spec.TCPProxy != nil && (passthrough || enforceTLS):
+			//		b.processTCPProxy(ir, nil, host)
+			case httplb.Spec.Routes != nil:
+				fmt.Println("processing an include: ", delegatedHTTPLb.Name)
+				b.processRoutes(delegatedHTTPLb, host, mergeConditions(delegatedCondition, &include.Condition), enforceTLS)
+			}
+
+			// Loop over any includes in the delegated httlb
+			if len(delegatedHTTPLb.Spec.Includes) > 0 {
```
davecheney: I think this needs some logic, similar to processIngressRoutes and processTCP that prevents it from including itself in a loop. We need to have this before this PR is landed because an infinite loop will stall the dag builder.

stevesloka:
please hold on review, adding tests and found an issue.

internal/dag/builder.go
davecheney: thank you

Issue #1425, opened 2019-09-02
Rework how statuses are written during dag rebuild


davecheney:
Currently status writing is handled as part of the `CacheHandler.OnUpdate` operation. The `dag.DAG` contains a `Statuses` method which returns status objects collected during the dag build. This approach has several limitations

1a. Status writing causes the cache handler to have a reference to the Contour k8s Client so it can write status back to the API server. Ideally the CacheHandler would have no knowledge of k8s at this point--its role is to manage the `internal/contour` visitors and hold their output.
1b. Visitation occurs after DAG build so any problems encountered by the visitors should by unrelated to k8s objects as the DAG abstracts said objects.
2. Status update has always been a second class citizen. Because and continuing this fact is there are few tests of status because they occur outside the dag build cycle, and the ones that we do have are specific to IngressRoute. There are no e2e tests of status.
3. We only report status for IngressRoute objects, not Ingress, not HTTPLoadBalancer, etc.
4. Status updates are tied to dag regeneration and this has a long history of producing duplicates, not to mention multiple contours trying to update the same object.

I propose introducing a layer of abstraction conceptual equivalent to a logger. To set the context, with Logrus we pass down into various things a FieldLogger, which is an object which collects bits of context along the way. At the bottom of the call chain, when something wants to log it doesn't need to have all the context of the various pieces that were added to the logger's context on the way down, they're already present in the FieldLogger.

I'm proposing something similar for status, a status logger or writer which collects context as it is used. The primary context is obviously the k8s object that is being worked on at the moment, but as we dig deeper into the object it might be a route, a service connected to a route, etc. There are also aspects of the builder pattern as we want to collect multiple units of status per object then emit them once

The overall API might look something like this
```go
type StatusWriter interface {
        WithObject(obj interface{}) ObjectStatusWriter
}

type ObjectStatusWriter {
      
        // various logging methods
  
        // Commit writes the colletive status back to the API server
        Commit() error
}
```
This would let us decouple anything wanting to write status from the actual status writing operation itself. `WithObject(obj).Commit()` would be a noop if there was no status to write. It would also let us test status writing directly to the fake k8s client rather than a side effect of dag building.

stevesloka:
@davecheney is this all fixed up with #1475?

PR #1475, opened 2019-09-09
internal/dag: introduce StatusWriter

davecheney:
Updates #1425 

Introduce a StatusWriter type and use it mostly within the builder. Fixing _orphaned_ status writing will be more complicated.

### DIFF
--- a/internal/contour/metrics.go
+++ b/internal/contour/metrics.go
@@ -18,7 +18,6 @@ package contour
 
 import (
 	ingressroutev1 "github.com/heptio/contour/apis/contour/v1beta1"
-	projcontour "github.com/heptio/contour/apis/projectcontour/v1alpha1"
 	"github.com/heptio/contour/internal/dag"
 	"github.com/heptio/contour/internal/metrics"
 )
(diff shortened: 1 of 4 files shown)


internal/contour/metrics.go
```diff
@@ -31,29 +30,18 @@ func calculateIngressRouteMetric(statuses map[dag.Meta]dag.Status) metrics.Ingre
...
-			metricValid[metrics.Meta{VHost: v.Vhost, Namespace: namespace}]++
+			metricValid[metrics.Meta{VHost: v.Vhost, Namespace: v.Object.GetObjectMeta().GetNamespace()}]++
 		case dag.StatusInvalid:
-			metricInvalid[metrics.Meta{VHost: v.Vhost, Namespace: namespace}]++
+			metricInvalid[metrics.Meta{VHost: v.Vhost, Namespace: v.Object.GetObjectMeta().GetNamespace()}]++
 		case dag.StatusOrphaned:
-			metricOrphaned[metrics.Meta{Namespace: namespace}]++
+			metricOrphaned[metrics.Meta{Namespace: v.Object.GetObjectMeta().GetNamespace()}]++
+		}
+		if v.Object.(*ingressroutev1.IngressRoute).Spec.VirtualHost != nil {
```
stevesloka: Need to check for typei= of `v.Object` here, if I use an `HTTPLoadBalancer` I get an exception since the type isn't an `IngressRoute`.

davecheney: Ahh, thank you. Good catch, I need to type assert here.

PR #1479, opened 2019-09-10
all: rename HTTPLoadBalancer to HTTPProxy

davecheney:
Blame Dave.

- examples/ds-hostnet-split-httploadbalancer has been removed. We
shouldn't be adding more different ways to deploy contour, we need to be
centralising on example/contour.


stevesloka:
> We shouldn't be adding more different ways to deploy contour, we need to be
> centralising on example/contour.

The reason I did this originally was so that folks wouldn't need to deploy the alpha/beta versions of `HTTPProxy` if they didn't want to.

PR #1482, opened 2019-09-10
internal/dag: AddHTTPProxy tests

### LINKED ISSUE #1459: internal/dag: Detect root httlb delegatation to another root
An root httplb (has a VirtualHost.Fqdn defined) cannot delegate to another root httplb. 

Similar implementation in IngressRoute: https://github.com/heptio/contour/blob/master/internal/dag/builder.go#L682-L685

stevesloka:
Add HTTPProxy tests to dag/builder.go

Fixes #1459

### DIFF
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -444,7 +444,7 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
 		return
 	}
 
-	// ensure root ingressroute lives in allowed namespace
+	// ensure root httpproxy lives in allowed namespace
 	if !b.rootAllowed(proxy.Namespace) {
 		sw.SetInvalid("root HTTPProxy cannot be defined in this namespace")
 		return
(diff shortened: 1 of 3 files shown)


internal/dag/builder.go
```diff
@@ -753,7 +756,7 @@ func (b *Builder) processIngressRoutes(sw *ObjectStatusWriter, ir *ingressroutev
 			sw.Commit()
 		}
 	}
-	sw.SetValid()
+	sw.SetValid("valid IngressRoute")
```
davecheney: please revert this. The name of the object can be inferred from the type stored in the statuswriter

stevesloka: I guess I can make it `WithValue(...)`, but the problem is that `IngressRoute` is hard-coded in the `SetValid()` method: https://github.com/heptio/contour/pull/1482/files#diff-942106236f0db9d26ef4d33d498122eeL66

PR #1490, opened 2019-09-12
Add integration test design document.

### LINKED ISSUE #1155: design: e2e/integration test rig
Contour currently has standard unit tests as well as some e2e tests that test the configuration that gets passed to Envoy, however, there aren't any full integration tests which spin up an instance of Contour & Envoy. 

The purpose of this issue is to gather a list of use-cases / tests to decide if the effort to build/maintain a full e2e test suite shows us value.

youngnick:
Fixes #1155.

### DIFF
--- a/design/integration-test-design.md
+++ b/design/integration-test-design.md
@@ -1,0 +1,103 @@
+# Contour integration test design
+
+Status: Draft
+
+This document outlines a new testing process and rig for Contour.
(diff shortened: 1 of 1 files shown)


youngnick:
Closing out this PR as it's very stale, @jpeach is doing much better work, and we won't be going ahead with this approach.

PR #1492, opened 2019-09-12
internal/dag: fix build

davecheney:
Fix e2e/rds test failure caused by virtual host that shouldn't be there
because it was part of an invalid delegation chain.

The fix required moving the check for delegation from a root to another
root earlier in the process which required changing some fixtures.
Specifically a non root can never delegate to a root, either directly or
indirectly so we can't use that as a test fixture.


davecheney:
Self approving to fix the build. @stevesloka please review when you can. Thanks!

PR #1494, opened 2019-09-12
internal/dag: Fixed route status not setting from within an HTTPProxy

stevesloka:
A new status writer was getting passed to `processRoutes`, but there was no `Commit()`, so status wasn't getting set back to the object in the API.


davecheney:
@stevesloka please rebase and try again. I landed a bunch of PRs this afternoon which I hope will fix the shoddy status api I wrote. I think it should be a bit easier to use now.

PR #1501, opened 2019-09-14
apis/projectcontour: Proxy header spec

stevesloka:
Updating the `Condition` spec's support for headers. This design allows for a `MatchType` which can be one of: 
- exact
- regex
- prefix
- suffix

The absence of a `value` means that the header must be present in the request regardless of its value. Also, `invert` may be set to `true` meaning the match result will be inverted before checking.


stevesloka:
But where does the `key` for the header get specified? 

Going with your suggestion we'd have:

- headerMatch (exactmatch)
- headerNotMatch (neg-match)
- headerContains (regex)
- headerNotContains (neg-regex)
- headerStartsWith (prefix)
- headerNotStartsWith (neg-prefix)
- headerEndsWith (suffix)
- headerNotEndsWith (neg-suffix)

stevesloka:
I kind of merged your ideas with what I had, let me know if you think this works:

```
spec:
  routes:
  - conditions:
     - headersContain: 
         - x-header: foo
```

davecheney:
Closing, we wen't with a different design.

PR #1509, opened 2019-09-16
internal/envoy: always preserve X-Request-Id

### LINKED ISSUE #1487: Preserve incoming X-Request-Id headers
For my use-case, I have Contour running as an ingress to my cluster but I have a service that routes HTTP requests to the Contour endpoint. The routing service generates a `X-Request-Id` header, but Envoy in it's default configuration will sanitize it and create a new request ID making log correlation difficult. I'd like to have some way I can set Envoy via Contour to not sanitize that header.

Envoy seems to permit setting this with the [`preserve_external_request_id`](https://www.envoyproxy.io/docs/envoy/v1.11.1/api-v2/config/filter/network/http_connection_manager/v2/http_connection_manager.proto.html#envoy-api-field-config-filter-network-http-connection-manager-v2-httpconnectionmanager-preserve-external-request-id) setting in HttpConnectionManager.

This is configured when the listener is created so my initial thought was to add a flag to Contour, but I'd love to hear more suggestions and feedback.

Not that I can think of.

**Environment:**

- Contour version: 0.14.2
- Kubernetes version: (use `kubectl version`): 1.14.6
- Kubernetes installer & version: Kubeadm 1.14
- OS (e.g. from `/etc/os-release`): Ubuntu 18.04.3 LTS

davecheney:
Fixes #1487

Tell Envoy to preserve the X-Request-Id if presented.


davecheney:
@jessedearing this will be available in `docker.io/projectcontour/contour:master` as soon as it lands and 1.0.0-beta.1 at the end of the week. Let me know if you'd like this to be backported to 0.15.1

Issue #624, opened 2018-08-13
Add support for structured logs


TrevinTeacutter:
It would be nice to have access logs that are structured as we use splunk for our logging aggregation and extracting key/value pairs is much nicer in splunk with structured logs than it is with regex. Could this be some sort of flag to enable this or did I just miss that this option is already enabled somewhere?

Blocked
- [x] #1130

davecheney:
Thank you for raising this issue. The access logs are generated directly from envoy. At the moment we have flags to control their output destination, but their format is currently defaulting to whatever envoy uses by default -- something akin to apache2 log.

https://github.com/heptio/contour/blob/master/internal/contour/listener.go#L338

Looking at 

https://www.envoyproxy.io/docs/envoy/v1.6.0/configuration/access_log#config-access-log-format

It looks like its possible to redefine the output of the logging format.

I'm wary of exposing the full configuration syntax of this to end users, it's too close to showing how the sausage is made. With that said, when you say _structured logging_, do you mean JSON?

TrevinTeacutter:
Yes, I mean JSON. We recently turned JSON access logs on for Traefik in our old DC/OS clusters and they have been useful as well as capable of including/redacting/excluding headers of requests coming through the ingress making correlation really easy, stuff that normally doesn't get included in apache2 logs for brevity's sake.

TrevinTeacutter:
Yes, I mean JSON. We recently turned JSON access logs on for Traefik in our old DC/OS clusters and they have been useful as well as capable of including/redacting/excluding headers of requests coming through the ingress making correlation really easy, stuff that normally doesn't get included in apache2 logs for brevity's sake.

01deyishu:
Maybe support define log format by configmap is a good way , just like nginx-ingress-controller use log-format-upstream parameter, also istio use logentry is a good practice.

davecheney:
We're not going to get to this for the 0.7 release. Removing milestone.

azuretek:
Has there been any agreement on this? Right now this is making Contour a less usable replacement for Nginx, having Envoy metrics available is great but getting full access log data would go a long way to making this a viable option.

davecheney:
No there has been no agreement because, apart from a request for json formatted logs, no details have been provided.

azuretek:
For our use case it would be nice if we could add an annotation or even just load a ConfigMap that will add a structure like so to every listener filter chain.

```
access_log: [
  {
    name: "envoy.file_access_log",
    config: {
[...]

PR #1511, opened 2019-09-16
Add support for JSON logging.

youngnick:
Updates #624

This PR contains the code for JSON logging only. The ticket will be closed when the documentation is merged as well.

### DIFF
--- a/internal/contour/listener.go
+++ b/internal/contour/listener.go
@@ -19,6 +19,7 @@ import (
 
 	envoy_api_v2_auth "github.com/envoyproxy/go-control-plane/envoy/api/v2/auth"
 	envoy_api_v2_listener "github.com/envoyproxy/go-control-plane/envoy/api/v2/listener"
+	envoy_api_v2_accesslog "github.com/envoyproxy/go-control-plane/envoy/config/filter/accesslog/v2"
 
 	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
 	"github.com/envoyproxy/go-control-plane/pkg/cache"
@@ -36,6 +37,7 @@ const (
 	DEFAULT_HTTPS_ACCESS_LOG       = "/dev/stdout"
 	DEFAULT_HTTPS_LISTENER_ADDRESS = DEFAULT_HTTP_LISTENER_ADDRESS
 	DEFAULT_HTTPS_LISTENER_PORT    = 8443
+	DEFAULT_ACCESS_LOG_TYPE        = "clf"
 )
 
 // ListenerVisitorConfig holds configuration parameters for visitListeners.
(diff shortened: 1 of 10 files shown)


internal/contour/listener.go
```diff
@@ -127,6 +139,68 @@ func (lvc *ListenerVisitorConfig) httpsAccessLog() string {
...
+	return DEFAULT_ACCESS_LOG_TYPE
+}
+
+// accesslogFields returns the access log fields that should be configured
+// for Envoy, or a default set if not configured.
+func (lvc *ListenerVisitorConfig) accesslogFields() []string {
+	if lvc.AccessLogFields != nil {
+		return lvc.AccessLogFields
+	}
+	return []string{
```
davecheney: Please move this to a public variable next to the translation map. This is so we can reference it as the canonical source from the release notes.

internal/contour/listener.go
```diff
@@ -127,6 +139,68 @@ func (lvc *ListenerVisitorConfig) httpsAccessLog() string {
...
+		"user_agent",
+		"x_forwarded_for",
+	}
+}
+
+func (lvc *ListenerVisitorConfig) newInsecureAccessLog() []*envoy_api_v2_accesslog.AccessLog {
+	var accesslogger []*envoy_api_v2_accesslog.AccessLog
+	switch lvc.accesslogType() {
+	case "json":
+		accesslogger = envoy.FileAccessLogJSON(lvc.httpAccessLog(), lvc.accesslogFields())
```
davecheney: case "json":
      return envoy.FileAccessLogJson...
default:
     return envoy.FileAccessLog...
}

internal/contour/listener.go
```diff
@@ -127,6 +139,68 @@ func (lvc *ListenerVisitorConfig) httpsAccessLog() string {
...
+	default:
+		accesslogger = envoy.FileAccessLog(lvc.httpAccessLog())
+	}
+	return accesslogger
+}
+
+func (lvc *ListenerVisitorConfig) newSecureAccessLog() []*envoy_api_v2_accesslog.AccessLog {
+	var accesslogger []*envoy_api_v2_accesslog.AccessLog
+	switch lvc.accesslogType() {
+	case "json":
```
davecheney: same as above

internal/contour/listener.go
```diff
@@ -209,6 +285,7 @@ func (l listenersByName) Less(i, j int) bool {
 	return l[i].(*v2.Listener).Name < l[j].(*v2.Listener).Name
 }
 
+// TypeURL returns the cache's Listener Type.
```
davecheney: nit: this is why I don't pay attention to the lint warnings ... this style of comment adds little beyond shutting the linter up.

internal/contour/listener.go
youngnick: I wasn't sure if that style was okay, thanks.

internal/contour/listener.go
youngnick: agreed, removed.

PR #1516, opened 2019-09-17
all: rename heptio-contour namespace projectcontour

### LINKED ISSUE #1505: Renaming heptio-contour namespace
The current namespace contour is deployed into contains the name of a defunct company. What should we rename the namespace?

- [ ] `contour`
- [x] `projectcontour`
- [ ] `vmware-contour`
- [ ] D. none of the above.

---
Blocked
- [ ] #1418

davecheney:
Fixes #1505

Update examples, documentation, and tests.


docs/grpc-tls-howto.md
```diff
@@ -86,13 +86,13 @@ Like the contour cert, this CSR uses the file [_integration/cert-envoy.ext](./_i
...
+kubectl create secret -n projectcontour tls contourcert --key=./certs/contourkey.pem --cert=./certs/contourcert.pem
+kubectl create secret -n projectcontour tls envoycert --key=./certs/envoykey.pem --cert=./certs/envoycert.pem
 ```
 
 Note that we don't put the CA **key** into the cluster, there's no reason for that to be there, and that would create a security problem. That also means that the `cacert` secret can't be a `tls` type secret, as they must be a keypair.
 
 # Conclusion
 
-Once this process is done, the certificates will be present as Secrets in the `heptio-contour` namespace, as required by `examples/ds-hostnet-split`.
+Once this process is done, the certificates will be present as Secrets in the `projectcontour` namespace, as required by `examples/ds-hostnet-split`.
```
youngnick: Update to `examples/contour`?

docs/troubleshooting.md
```diff
@@ -23,7 +23,7 @@ To access it from your workstation use `kubectl port-forward` like so,
 
 ```sh
 # Get one of the pods that matches the Envoy daemonset
-ENVOY_POD=$(kubectl -n heptio-contour get pod -l app=envoy -o name | head -1)
+ENVOY_POD=$(kubectl -n projectcontour get pod -l app=envoy -o name | head -1)
 # Do the port forward to that pod
 kubectl -n heptio-contour port-forward $ENVOY_POD 9001
```
youngnick: There are a few more instances of `heptio-contour` in this file, like this one

docs/grpc-tls-howto.md
davecheney: oooh, good catch, i'll grep for deceased examples.

PR #1517, opened 2019-09-17
Add documentation for structured logs.

### LINKED ISSUE #624: Add support for structured logs

youngnick:
Fixes #624.

Also update a small bug in config file definition that can lead to parsing weirdness.


youngnick:
That's a bad habit of mine, agreed.

PR #1518, opened 2019-09-17
examples: Update HTTPProxy CRD / validation

stevesloka:
Updates #1516 which removed the working version of the HTTPProxy CRD. 


stevesloka:
Ahh we're missing RBAC as well, one second.

Issue #1476, opened 2019-09-09
Upgrading guide needs to be updated for 0.15 -> 1.0.0-beta1


davecheney:
We need to update the upgrading guide to cover upgrading from 0.15 to 1.0.0-beta1.

TODO:
- [x] Cover the move from gcr to docker hub, noting that 0.15 is available in both locations.
- [x] split deployment is the new normal 
- [x] election is opt out, not opt in
- [x] :new: http load balancer
- [x] `/healthz` for readiness is now `/ready`
- [x] Deprecation notice on IngressRoute documentation.

PR #1521, opened 2019-09-17
docs: update upgrading.md for Contour 1.0.0-beta.1

davecheney:
Updates #1476

Draft, first cut of the upgrade documentation.

### DIFF
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -2,6 +2,98 @@
 
 This document describes the changes needed to upgrade your Contour installation.
 
+## Upgrading Contour 0.15 to 1.0.0-beta.1
+
+Contour 1.0.0-beta.1 changes the namespace Contour is deployed too, promotes leader election to on by default, and introduces a new version of the IngressRoute CRD, now called HTTPProxy.
+
+### Beta release
+
+Contour 0.15.0 remains the current stable release.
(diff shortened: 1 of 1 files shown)


docs/upgrading.md
```diff
@@ -2,6 +2,98 @@
...
+
+### Recommended Envoy version
+
+The recommended version of Envoy remains unchanged from Contour 0.15.
+Ensure the Envoy image version is `docker.io/envoyproxy/envoy:v1.11.1`.
+
+### Leader Election
+
+Contour 1.0.0-beta.1 enables leader election by default.
+No specific configuration is required
```
youngnick: No specific configuration is needed for this feature, but it will use a ConfigMap named `leader-elect` in the `projectcontour` namespace. These details can be changed using the config file.

(or something).

youngnick: Also, we need to note that leader election now does *not* block the gRPC endpoint from becoming available, that is, all Contours now serve xDS.

davecheney: @youngnick do you want to take a stab at this; you can push to this branch.

davecheney: Or we merge as is and you send a followup PR. :shrug:

youngnick: I'll do a followup PR.

PR #1529, opened 2019-09-17
Fix problems with example deployment

### LINKED ISSUE #1527: Contour config file example breaks deployment
If I deploy the sample Contour deployment (i.e. `kubectl apply -f examples/contour`), Contour fails to start and throws this error:

```
kubectl logs -f contour-754f4dd657-2rmk4 -n projectcontour                                              [14:35:17]
contour: error: EOF, try --help
```

Taking out the reference to the configfile allows Contour to run.

youngnick:
- Config file must have at least one value uncommented. Fixes #1527.
- Certgen Job must have imagePullPolicy set to Always for namespace change to take effect.

### DIFF
--- a/examples/contour/01-contour-config.yaml
+++ b/examples/contour/01-contour-config.yaml
@@ -12,8 +12,8 @@ data:
     # kubeconfig: /path/to/.kube/config
     #
     # disable ingressroute permitInsecure field
-    # disablePermitInsecure: false
-    # tls:
+    disablePermitInsecure: false
+    tls:
     #   minimum TLS version that Contour will negotiate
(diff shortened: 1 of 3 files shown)


examples/contour/01-contour-config.yaml
```diff
@@ -22,7 +22,7 @@ data:
     #   configmap-namespace: leader-elect
     ### Logging options
     # Default setting
-    # accesslog-format: clf
+    accesslog-format: clf
```
davecheney: Is this a valid setting?

youngnick: Yep, anything not 'json' will set the logging to the current logging format.

Issue #1462, opened 2019-09-05
docs: Update docs for httploadbalancer


stevesloka:
**Describe the solution you'd like**
Need to finalize docs and fill in the large missing pieces that are conflicting with current implementations. 

TODO post #1530 
- [x] Create example directory
- [x] Document Conditions
- [x] Update examples with correct `conditions:` syntax.
- [x] Document `timeoutPolicy.idle`
- [x] Document inclusion

PR #1530, opened 2019-09-17
docs: add draft HTTPProxy documentation

davecheney:
Updates #1462

Document HTTPProxy using the IngressRoute docs as a base.

### DIFF
--- a/docs/httpproxy.md
+++ b/docs/httpproxy.md
@@ -8,7 +8,7 @@ The goal of the `HTTPProxy` (previously `IngressRoute`) Custom Resource Definiti
 ## Key HTTPProxy Benefits
 
 - Safely supports multi-team Kubernetes clusters, with the ability to limit which Namespaces may configure virtual hosts and TLS credentials.
-- Enables delegation of routing configuration for a path or domain to another Namespace.
+- Enables including of routing configuration for a path or domain from another HTTPProxy, possibly in another Namespace.
 - Accepts multiple services within a single route and load balances traffic across them.
 - Natively allows defining service weighting and load balancing strategy without annotations.
 - Validation of HTTPProxy objects at creation time and status reporting for post-creation validity.
@@ -19,7 +19,7 @@ A minimal Ingress object might look like:
 
 ```yaml
 # ingress.yaml
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1beta1
 kind: Ingress
 metadata:
   name: basic
@@ -85,7 +85,7 @@ basic     24s
 Describing HTTPProxy:
 
 ```bash
-$ kubectl describe httpprox basic
+$ kubectl describe httpproxy basic
 Name:         basic
 Namespace:    default
 Labels:       <none>
@@ -118,7 +118,7 @@ httpproxy "basic" deleted
 
 ## HTTPProxy API Specification
 
-TODO(dfc) create example directory
+TODO(youngnick) create example directory
 
 There are a number of [working examples](/examples/example-workload/httpproxy) of HTTPProxy objects in the `examples/example-workload` directory.
 
@@ -162,7 +162,7 @@ spec:
           servicePort: 80
 ```
 
-must be represented by two different HTTPproxy objects:
+must be represented by two different HTTPProxy objects:
 
 ```yaml
 # httpproxy-name.yaml
@@ -250,9 +250,10 @@ If the `tls.secretName` property contains a slash, eg. `somenamespace/somesecret
 See TLS Certificate Delegation below for more information.
 
 The TLS **Minimum Protocol Version** a vhost should negotiate can be specified by setting the `spec.virtualhost.tls.minimumProtocolVersion`:
-  - 1.3
-  - 1.2
-  - 1.1 (Default)
+
+- 1.3
+- 1.2
+- 1.1 (Default)
 
 A HTTPProxy can be configured to permit insecure requests to specific Routes.
 In this example, any request to `foo2.bar.com/blog` will not receive a 301 redirect to HTTPS, but the `/` route will:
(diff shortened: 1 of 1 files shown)


docs/httpproxy.md
```diff
@@ -0,0 +1,833 @@
...
+    fqdn: foo-basic.bar.com
+  routes:
+    - conditions:
+        prefix: /
+      services:
+        - name: s1
+          port: 80
+```
+
+**Lines 1-4**: As with all other Kubernetes objects, an HTTPProxy needs apiVersion, kind, and metadata fields. Note that the HTTPProxy API is currently considered beta.
```
youngnick: It's currently alpha?

docs/httpproxy.md
```diff
@@ -0,0 +1,833 @@
...
+        prefix: /blog
+      permitInsecure: true
+      services: 
+        - name: s2
+          port: 80
+```
+
+#### Upstream TLS
+
+A HTTPProxy can proxy to an upstream TLS connection by first annotating the upstream Kubernetes service with: `contour.heptio.com/upstream-protocol.tls: "443,https"`.
```
youngnick: Annotation name update?

docs/httpproxy.md
```diff
@@ -0,0 +1,833 @@
...
+    - name: httpbin
+      port: 8080
+      strategy: Cookie
+```
+##### Limitations
+
+Session affinity is based on the premise that the backend servers are robust, do not change ordering, or grow and shrink according to load.
+None of these properties are guaranteed by a Kubernetes cluster and will be visible to applications that rely heavily on session affinity.
+
+Any pertibation in the set of pods backing a service risks redistributing backends around the hash ring.
```
youngnick: nit: 'perturbation'

docs/httpproxy.md
```diff
@@ -0,0 +1,833 @@
...
+```bash
+$ kubectl get httpproxy
+NAME      AGE
+basic     24s
+```
+
+Describing HTTPProxy:
+
+```bash
+$ kubectl describe httpprox basic
```
stevesloka: s/httpprox/httpproxy/g

docs/httpproxy.md
```diff
@@ -0,0 +1,833 @@
...
+              | 178.91.123.132  |
+bar.foo.com --|                 |-> bar.foo.com s2:80
+```
+
+Unlike Ingress, HTTPProxy only support a single root domain per HTTPProxy object.
+As an example, this Ingress object:
+
+```yaml
+# ingress-name.yaml
+apiVersion: extensions/v1beta1
```
stevesloka: Should we switch the examples to use the new group? (e.g. `networking.k8s.io/v1beta1`)

docs/httpproxy.md
```diff
@@ -0,0 +1,833 @@
...
+          servicePort: 80
+  - host: bar1.bar.com
+    http:
+      paths:
+      - backend:
+          serviceName: s2
+          servicePort: 80
+```
+
+must be represented by two different HTTPproxy objects:
```
stevesloka: nit: s/HTTPproxy/HTTPProxy/g

docs/httpproxy.md
```diff
@@ -0,0 +1,833 @@
...
+apiVersion: projectcontour.io/v1alpha1
+kind: HTTPProxy
+metadata:
+  name: name-example-foo
+  namespace: default
+spec:
+  virtualhost:
+    fqdn: foo1.bar.com
+  routes:
+    - conditions:
```
stevesloka: Conditions on route are singular (e.g. `condition`). Also, they aren't an array for the Route spec since you can only apply a single `Condition` to a specific route: https://github.com/projectcontour/contour/blob/master/apis/projectcontour/v1alpha1/httpproxy.go#L82

docs/httpproxy.md
stevesloka: I'd actually like to get rid of this annotation. We already specify which service/port to use, just need to add a field to the spec to upgrade the connection. 

My idea previously was to remove the `name` field of `service` and put the protocol in there (https://github.com/projectcontour/contour/pull/1195/files#diff-fc42c2911ab7ef4fd20292f823da57ebR116). This removes the need for annotations on services as well as can remove the `TCPProxy` spec.

docs/httpproxy.md
```diff
@@ -0,0 +1,833 @@
...
+This facility allows the owner of a TLS certificate to delegate, for the purposes of referencing the TLS certificate, permission to Contour to read the Secret object from another namespace.
+
+```yaml
+apiVersion: projectcontour.io/v1alpha1
+kind: TLSCertificateDelegation
+metadata:
+  name: example-com-wildcard
+  namespace: www-admin
+spec:
+  delegations:
```
stevesloka: Are we going to keep the `delegation` nomenclature for certificates? I still use it in words to describe the process, but it's different now from IngressRoute.

docs/httpproxy.md
```diff
@@ -0,0 +1,833 @@
...
+      services:
+        - name: s1
+          port: 80
+```
+
+In this example, the permission for Contour to reference the Secret `example-com-wildcard` in the `admin` namespace has been delegated to HTTPProxy objects in the `example-com` namespace.
+
+### Conditions
+
+Each Route entry in a HTTPProxy may contain one or more conditions.
```
stevesloka: Routes can't have multiple, but the spec can.

docs/httpproxy.md
```diff
@@ -0,0 +1,833 @@
...
+
+TODO(dfc) document conditions
+
+#### Multiple Routes
+
+HTTPProxy must have at least one route defined.
```
stevesloka: This isn't 100% true for `root` proxies since they can define includes and no routes.

docs/httpproxy.md
```diff
@@ -0,0 +1,833 @@
...
+```
+##### Limitations
+
+Session affinity is based on the premise that the backend servers are robust, do not change ordering, or grow and shrink according to load.
+None of these properties are guaranteed by a Kubernetes cluster and will be visible to applications that rely heavily on session affinity.
+
+Any pertibation in the set of pods backing a service risks redistributing backends around the hash ring.
+This is an unavoidable consiquence of Envoy's session affinity implementation and the pods-as-cattle approach of Kubernetes.
+
+#### Per-Upstream Active Health Checking
```
stevesloka: We should probably have a note about how healthchecks work with multiple upstreams and the dangers involved.

davecheney:
@youngnick I don't have time to finish this. Would you please address the comments and land this today so we can discuss it on the community call tomorrow. Thank you.

docs/httpproxy.md
youngnick: Resolving this one because this is not the case any more.

PR #1531, opened 2019-09-18
Remove all references to CLF, use Envoy's default instead.

### LINKED ISSUE #1124: Envoy should output CLF access logs by default
Currently we configure Envoy to http access logs to `/dev/stdout`. Because we do not specify any logging format,[Envoy's default format](https://www.envoyproxy.io/docs/envoy/latest/configuration/access_log#config-access-log-default-format) is chosen for us.

This is not ideal for two reasons

1. The format is not from [CLF](https://en.wikipedia.org/wiki/Common_Log_Format) which is emitted by competitors like nginx and apache. CLF is a well understood format that is handled natively by most log aggregation platforms.
2. Because the default format is not CLF, the requirement for end users to write their own parser for Envoy's default format drives them towards other solutions like JSON logging. See #624 and #1080 for discussion on JSON logging as an option, or default.

Putting aside #624 we _should_ change to emitting CLF. There is no reason to use Envoy's format and hopefully by offering CLF as a default we can lessen the calls for changing the logging to something else that is easier (than Envoy's default format) for log aggregation platforms to ingest.

youngnick:
Update #1124.


design/envoy-json-logging.md
```diff
@@ -68,9 +68,9 @@ logging:
 
 #### `format`
 
-There are two valid options here: `json` and `clf`.
+There are two valid options here: `json` and `envoy`.
 
-`clf` is the default, and will set Envoy logs to CLF format.
+`envoy` is the default, and will set Envoy logs to Envoy's [default format](https://www.envoyproxy.io/docs/envoy/v1.5.0/configuration/access_log#default-format).
```
davecheney: s/v1.5.0/latest

davecheney: `envoy` will use the Envoy's [default format].

internal/envoy/accesslog.go
```diff
@@ -80,7 +80,7 @@ func FileAccessLog(path string) []*accesslog.AccessLog {
 		ConfigType: &accesslog.AccessLog_TypedConfig{
```
davecheney: Please rename this to `FileAccessLogEnvoy`

internal/envoy/accesslog.go
```diff
@@ -80,7 +80,7 @@ func FileAccessLog(path string) []*accesslog.AccessLog {
...
 			TypedConfig: toAny(&accesslogv2.FileAccessLog{
 				Path: path,
-				// TODO(dfc) FileAccessLog_Format elided.
+				// AccessLogFormat elided, see https://github.com/projectcontour/contour/issues/1124 for why.
```
davecheney: Access log format left blank to defer to envoy's default log format.

PR #1533, opened 2019-09-18
internal/contour: clean up route visitor tests

davecheney:
Clean up the route visitor tests by introducing helpers from the envoy
package that we use elsewhere.

### DIFF
--- a/internal/envoy/accesslog_test.go
+++ b/internal/envoy/accesslog_test.go
@@ -14,8 +14,6 @@
 package envoy
 
 import (
-	"encoding/json"
-	"fmt"
 	"testing"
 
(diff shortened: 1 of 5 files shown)


internal/envoy/accesslog_test.go
```diff
@@ -111,11 +109,6 @@ func TestJSONFileAccessLog(t *testing.T) {
 	for name, tc := range tests {
 		t.Run(name, func(t *testing.T) {
 			got := FileAccessLogJSON(tc.path, tc.headers)
-			output, err := json.Marshal(got)
```
youngnick: Again, oops. Sorry.

Issue #1514, opened 2019-09-16
HTTPProxy not pushing updates to Envoy


stevesloka:
**What steps did you take and what happened:**
When Contour first spins up, it builds a DAG and the config is pushed down to Envoy, however, if an HTTPProxy changes, the DAG is rebuilt, but Envoy isn't updated dynamically.

davecheney:
Can you verify this with a `contour cli` session running in another window and the editing a HTTPProxy object?

davecheney:
Oops. This was my fault. The underlying cause is shadowing of `hints` here https://github.com/projectcontour/contour/blob/master/internal/contour/cond.go#L64

The result is if you register with a hint, which envoy does for RDS, EDS, and CDS(?) then `hints` would be set to the hint registered during `Register`, line 64 would not trigger because the length of things hints was > 0, however the following lines would also not trigger because the map `m` built from the outer hints was empty.

The end result is `contour cli rds` works, but `contour cli rds ingress_https` would only receive one update. 

Fix coming up.

PR #730, opened 2018-10-04
append x-request-start header to incoming HTTP requests

yob:
Some tracing applications support reporting on queuing time for incoming requests. Generally this is implemented by adding a `X-Request-Start` timestamp header to the requests as early as possible in the request path.

[New Relic](https://docs.newrelic.com/docs/apm/applications-menu/features/request-queue-server-configuration-examples) and [datadog](https://github.com/DataDog/dd-trace-rb/blob/master/docs/GettingStarted.md#http-request-queuing) are two examples of tools that support this header.

The [envoy docs](https://www.envoyproxy.io/docs/envoy/latest/configuration/http_conn_man/headers#custom-request-response-headers) say:

> Custom request/response headers can be added to a request/response at
> the weighted cluster, route, virtual host, and/or global route
> configuration level

I've opted to add them at the weighted cluster level, based on a [slack discussion with @stevesloka](https://kubernetes.slack.com/archives/C8XRH2R4J/p1538487259000100). I'm happy to change that if requested though - my envoy knowledge is pretty low, and I'm not sure what the tradeoffs of each location are.

The header is added to all requests unconditionally, as suggested in the original github issue.

Fixes #646

PR #1534, opened 2019-09-18
internal/envoy: move x-request-start header to RouteConfiguration

davecheney:
This PR fell out of debugging #1514.

A year ago #730 added the x-request-start header to each incoming
request. At the time this was done by adding the hedaer on each
route. This is verbose to see in the configuration and can be done
once at the enclosing RouteConfiguration.

This change means empty RouteConfigurations will have a
RequestHeadersToAdd field, but this is of little consiquence as it
applies to no routes as the list of virtualhosts is empty.

Another side effect of this change is the x-request-start header will be
added to redirects, however as this is a _request_ header, it is never
presented downstream during 301 processing so again it should have
little impact.

### DIFF
--- a/internal/contour/route.go
+++ b/internal/contour/route.go
@@ -90,13 +90,19 @@ type routeVisitor struct {
 }
 
 func visitRoutes(root dag.Vertex) map[string]*v2.RouteConfiguration {
+	headers := envoy.Headers(
+		envoy.AppendHeader("x-request-start", "t=%START_TIME(%s.%3f)%"),
+	)
+
 	rv := routeVisitor{
 		routes: map[string]*v2.RouteConfiguration{
 			"ingress_http": {
-				Name: "ingress_http",
+				Name:                "ingress_http",
+				RequestHeadersToAdd: headers,
 			},
 			"ingress_https": {
-				Name: "ingress_https",
+				Name:                "ingress_https",
+				RequestHeadersToAdd: headers,
 			},
 		},
 	}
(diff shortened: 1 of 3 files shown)


davecheney:
Much cleaner :)
```json
"route_config": {
      "name": "ingress_https",
      "virtual_hosts": [
       {
        "name": "httpbin.davecheney.com",
        "domains": [
         "httpbin.davecheney.com",
         "httpbin.davecheney.com:*"
        ],
        "routes": [
         {
          "match": {
           "prefix": "/kuard2"
          },
          "route": {
           "cluster": "default/kuard/80/da39a3ee5e"
          }
         },
         {
          "match": {
           "prefix": "/kuard"
          },
          "route": {
           "cluster": "default/kuard/80/da39a3ee5e"
          }
         },
         {
          "match": {
           "prefix": "/"
          },
          "route": {
           "cluster": "default/httpbin/8080/da39a3ee5e"
          }
         }
        ]
       },
       {
        "name": "httpbin2.davecheney.com",
        "domains": [
         "httpbin2.davecheney.com",
         "httpbin2.davecheney.com:*"
        ],
        "routes": [
         {
          "match": {
           "prefix": "/foo"
          },
          "route": {
           "cluster": "default/httpbin/8080/da39a3ee5e"
          }
         },
         {
          "match": {
           "prefix": "/bar"
          },
          "route": {
           "cluster": "default/httpbin/8080/da39a3ee5e"
          }
         }
        ]
       }
      ],
      "request_headers_to_add": [
       {
        "header": {
         "key": "x-request-start",
         "value": "t=%START_TIME(%s.%3f)%"
        },
        "append": true
       }
      ]

PR #1535, opened 2019-09-18
internal: Header routing implementation with HTTPProxy

stevesloka:
This PR includes the initial header routing implementation with HTTPProxy. Requests can now be routed based upon header in the request. 

This PR depends on #1501 which is the spec of header routing in HTTPProxy. 

Also note, this PR is missing full E2E tests as well as handling header routing with `Includes`. Those will come in further PR's, but this lays the groundwork now.


stevesloka:
I can clean up the header spec, that's the one I've been looking for feedback on. 

Why do you think that Conditions should be an array in the route? You can only have a single condition specified on a route match right? Otherwise, someone could set two different path matches for the same route.

davecheney:
@stevesloka gonna close this based on the f2f discussions we had yesterday and the changes overnight.

PR #1539, opened 2019-09-19
Revert Contour deployment to Rolling Update

youngnick:
Also set maxSurge so that new pods will be created before old pods are removed.

### DIFF
--- a/examples/contour/03-contour.yaml
+++ b/examples/contour/03-contour.yaml
@@ -8,7 +8,11 @@ metadata:
 spec:
   replicas: 2
   strategy:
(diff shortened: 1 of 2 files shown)


examples/contour/03-contour.yaml
```diff
@@ -8,7 +8,11 @@ metadata:
 spec:
   replicas: 2
   strategy:
-    type: Recreate
+    type: RollingUpdate
+    rollingUpdate:
+      # This value  of maxSurge means that during a rolling update
```
davecheney: extra whitespace

PR #1546, opened 2019-09-19
internal/contour: Merge conditions for pathPrefix

stevesloka:
Merge all conditions for a route when multiple are defined for allowing for a final full pathPrefix.

### DIFF
--- a/internal/contour/route.go
+++ b/internal/contour/route.go
@@ -14,6 +14,7 @@
 package contour
 
 import (
+	"path"
 	"sort"
 	"sync"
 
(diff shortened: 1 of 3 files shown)


internal/contour/route.go
```diff
@@ -128,13 +129,18 @@ func (v *routeVisitor) visit(vertex dag.Vertex) {
 					if len(route.Conditions) < 1 {
 						return
 					}
-					switch cond := route.Conditions[0].(type) {
-					case *dag.PrefixCondition:
-						rr := envoy.Route(envoy.RoutePrefix(cond.Prefix), envoy.RouteRoute(route))
+
+					// Merge all pathPrefix conditions for this route
+					mergedPathPrefix := mergePathPrefixes(route.Conditions)
```
davecheney: why not make mergePathPrefix return "/" in the case there are no path prefixes?

internal/contour/route.go
```diff
@@ -189,6 +195,17 @@ func (v *routeVisitor) visit(vertex dag.Vertex) {
 	}
 }
 
+func mergePathPrefixes(conditions []dag.Condition) string {
+	mergedPath := ""
```
davecheney: I'd suggest start with "/"

internal/contour/route.go
```diff
@@ -189,6 +195,17 @@ func (v *routeVisitor) visit(vertex dag.Vertex) {
...
+	for _, c := range conditions {
+		switch cond := c.(type) {
+		case *dag.PrefixCondition:
+			mergedPath = fmt.Sprintf("%s%s", mergedPath, cond.Prefix)
```
davecheney: Then use path.Join() as it will handle the cases of double //'s

internal/contour/route.go
davecheney: I cannot comment on it because GitHub, but do you need to apply the same logic lower in the SecureVirtualhost path?

internal/contour/route.go
stevesloka: In my previous impl I had the builder adding a `/` to the prefix if it wasn't defined. This way you're not required to add a `Condition` field to the route if you don't want to (e.g. you have an `Include` giving you the prefix. I could put that logic back if that makes sense to you.

The other thing I was doing was using the absence of a prefix to flag that we're doing some other type of routing (e.g. regex). If `mergePathPrefixes` now returns `/` all the time, I'm not sure how to determine if it's regex or not.

stevesloka: I'll put the quick regex check in and move along. =)

PR #1549, opened 2019-09-20
internal/featuretest: remove unused envoy helper

davecheney:
staticcheck seems to be a bit flakey, it doesn't fire in CI but does
fire locally (I blame build caching in Go 1.13). Anyhoo, this function
is no longer needed.


davecheney:
Welp. I've no idea what I did wrong here, but I do know it ain't right.

PR #1557, opened 2019-09-20
dag: add support for traffic mirroring

### LINKED ISSUE #459: ingressroute: add support for traffic mirroring
Feature request from the field, add support for services which get read only mirror traffic. Envoy supports this at the cluster level. Suggest something like this for the spec
```
spec:
  virtualhost: ...
  routes:
  - match: "/"
     services:
     - name: "red"
        port: 9000
     - name: "green"
        port: 9001
        mirror: true
```

davecheney:
Fixes #459

Add support for per route traffic mirroring. Only available in HTTPProxy

```yaml
spec:
  routes:
  - conditions:
     - prefix: "/"
  - services:
     - name: kuard
       port: 8080
     - name: kuard-mirror
       port: 8080
       mirror: true
```

### DIFF
--- a/internal/featuretests/mirrorpolicy_test.go
+++ b/internal/featuretests/mirrorpolicy_test.go
@@ -1,0 +1,91 @@
+// Copyright © 2019 VMware
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package featuretests
+
+import (
+	"testing"
+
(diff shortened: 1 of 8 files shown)


stevesloka:
What happens if a user selects multiple "mirror" upstreams?

internal/featuretests/mirrorpolicy_test.go
```diff
@@ -0,0 +1,92 @@
...
+					Name:   svc2.Name,
+					Port:   8080,
+					Mirror: true,
+				}},
+			}},
+		},
+	}
+	rh.OnAdd(p1)
+
+	// check timeout policy with malformed response timeout is propogated as infinity
```
youngnick: Nit: Comment got left behind maybe?

davecheney:
> What happens if a user selects multiple "mirror" upstreams?

excellent point. I'll add a test and some status output tests.

davecheney:
> Need to make sure to update the calculation for weighted clusters (https://github.com/projectcontour/contour/blob/master/internal/envoy/route.go#L161) so that any mirrored cluster isn't taken into account in the event there are multiple upstreams.

This should be fine because the set of clusters we pass to weighted clusters is reduced by one -- the on which is assigned to the mirror. The weightings will work out -- hopefully/

PR #1562, opened 2019-09-23
internal/dag: Fix status tests & setting status on proxies with includes

stevesloka:
- Fix commented out status tests
- Fix issue where status wasn't getting set properly on any Proxy that
has includes

### DIFF
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -480,7 +480,6 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
 		insecure.addRoute(route)
 		secure.addRoute(route)
 	}
-	sw.SetValid()
 }
 
 // httpProxyConditions converts a list of HTTPProxy conditions to dag Conditions.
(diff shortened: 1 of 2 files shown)


internal/dag/builder.go
```diff
@@ -532,7 +537,9 @@ func (b *Builder) computeRoutes(sw *ObjectStatusWriter, proxy *projcontour.HTTPP
 				return nil
 			}
 
+			vhost := sw.values["vhost"] // Copy off vhost
 			sw, commit := b.WithObject(delegate)
+			sw.WithValue("vhost", vhost)
```
davecheney: I don't think this is correct. A HTTPProxy can be included by many different roots. I that case we cannot say that its status is related to any specific ghost. Please see https://github.com/projectcontour/contour/pull/1499

internal/dag/builder.go
```diff
@@ -590,6 +597,9 @@ func (b *Builder) computeRoutes(sw *ObjectStatusWriter, proxy *projcontour.HTTPP
 		}
 		routes = append(routes, r)
 	}
+	// Mark proxy as valid
+	sw.SetValid()
```
davecheney: I don't think this is the right place for this. I believe the call should happen in the caller of this method.

internal/dag/status_test.go
```diff
@@ -1055,655 +1055,634 @@ func TestDAGIngressRouteStatus(t *testing.T) {
...
-//			dag := builder.Build()
-//			got := dag.Statuses()
-//
-//			if diff := cmp.Diff(tc.want, got); diff != "" {
-//				t.Fatal(diff)
-//			}
-//		})
-//	}
-//}
+func TestDAGHTTPProxyStatus(t *testing.T) {
```
davecheney: Thanks for restoring these, I missed them in #1555

internal/dag/status_test.go
```diff
@@ -1055,655 +1055,634 @@ func TestDAGIngressRouteStatus(t *testing.T) {
...
+					Object:      proxy17,
+					Status:      StatusInvalid,
+					Description: `fqdn "example.com" is used in multiple HTTPProxies: roots/example-com, roots/other-example`,
+					Vhost:       "example.com",
+				},
+				{name: proxy18.Name, namespace: proxy18.Namespace}: {
+					Object:      proxy18,
+					Status:      StatusInvalid,
+					Description: `fqdn "example.com" is used in multiple HTTPProxies: roots/example-com, roots/other-example`,
+					Vhost:       "example.com",
```
davecheney: See above, I don't think there should be a Vhost field here

internal/dag/builder.go
stevesloka: I'll look again to see if I can do that, but what I found was since it's called recursively, the children proxies don't get status set, only the final one. 

This is also breaking my other status check for the same reason, so will work to flush it out.

stevesloka:
@davecheney I opened #1569 to track the location of setting status properly.

Issue #1559, opened 2019-09-20
internal/dag: Look for duplicate headers in a route configuration


stevesloka:
**Describe the solution you'd like**
An HTTPProx's route can end up with many Conditions applied to it through various levels of Includes. 

Contour should verify that the resulting set of Conditions do not overlap. This is specific for headers. If any set of headers duplicate key's, then an error should be set on the Proxy which is malformed.

PR #1573, opened 2019-09-25
Add Inclusion documentation for HTTPProxy

### LINKED ISSUE #1462: docs: Update docs for httploadbalancer

### LINKED ISSUE #1476: Upgrading guide needs to be updated for 0.15 -> 1.0.0-beta1

youngnick:
Fixes #1462

Add link to HTTPProxy documentation to `upgrading.md`, fixes #1476 


docs/httpproxy.md
```diff
@@ -750,7 +750,172 @@ spec:
 
 ## HTTPProxy inclusion
 
-TODO(youngnick) document inclusion
+HTTPProxy permits the splitting of a system's configuration into separate HTTPProxy instances using **inclusion**.
+
+Inclusion, as the name implies, allows for one HTTPProxy object to be included in another, optionally with some conditions around the inclusion.
```
davecheney: suggestion

s/some conditions around the inclusion/some conditions inherited from the parent/

docs/httpproxy.md
```diff
@@ -750,7 +750,172 @@ spec:
...
+Contour reads the inclusion tree and merges the included objects into one big object internally before rendering Envoy config.
```
davecheney: s/objects/routes.

docs/httpproxy.md
```diff
@@ -750,7 +750,172 @@ spec:
...
+Importantly, the included HTTPProxy objects do not have to be in the same namespace, so this is functionally the same as the old delegation feature of the now-deprecated IngressRoute.
```
davecheney: s/old//

docs/httpproxy.md
```diff
@@ -750,7 +750,172 @@ spec:
...
+
+The "root" HTTPProxy is the only entry point for a proxied virtual host and is used as the top level configuration of a cluster's proxy resources.
```
davecheney: Maybe

Each tree of HTTPProxy starts with a root, the top level object of the configuration for a particular virtual host.

docs/httpproxy.md
```diff
@@ -750,7 +750,172 @@ spec:
...
+Each root HTTPProxy defines a `virtualhost` key, which describes properties such as the fully qualified name of the virtual host, TLS configuration, etc.
+
+HTTPProxies that will be included do not contain  virtual host information, as they will inherit the properties of the parent HTTPProxy via `conditions`.
```
davecheney: s/HTTPProxies that will be included do not contain  virtual host information/HTTPProxies included from the root _must not_ contain a `virtualhost` key.
Root objects cannot include other roots either transitively or directly.

docs/httpproxy.md
```diff
@@ -750,7 +750,172 @@ spec:
...
+This permits the owner of an HTTPProxy root to allow the inclusion of a portion of the route space inside a virtual host, and to allow that route space to be further subdivided with inclusions.
+Because the path is not necessarily used as the only key, the route space can be multi-dimensional.
+
+### Conditions and Inclusion
+
+Inclusions may be specified with conditions.
```
davecheney: Like Routes, Inclusion may specify a set of conditions.

docs/httpproxy.md
```diff
@@ -750,7 +750,172 @@ spec:
...
+These conditions are added to any conditions in the included HTTPProxy.
```
davecheney: s/conditions in the included HTTPProxy/on the routes included./

docs/httpproxy.md
```diff
@@ -750,7 +750,172 @@ spec:
...
+This process repeated for any further inclusions as well.
```
davecheney: This process is recursive.

(this could probably move higher in the section)

docs/httpproxy.md
```diff
@@ -750,7 +750,172 @@ spec:
...
+
+All [Conditions](#conditions) are supported for use with inclusion.
```
davecheney: Maybe drop this, I don't think we need to document a thing which is never false.

docs/httpproxy.md
```diff
@@ -750,7 +750,172 @@ spec:
...
+The `prefix` condition is slightly special in that, when used for inclusion, any routes will have the value of `prefix` prepended to *their* prefix value.
```
davecheney: Maybe add a sentence before this. Perhaps a paragraph.

Conditions are sets of individual condition statements, for example `prefix: /blog` is the condition that the matching request's path must start with `/blog`.
When conditions are combined through inclusion Contour merges the conditions inherited via inclusion with any conditions specified on the route.
This may result in duplicates, for example two `prefix:` conditions, or two header match conditions with the same name and value.
To resolve this Contour applies the following logic.

- `prefix:` conditions are concatenated together in the order they were applied from the root object. For example the conditions, `prefix: /api`, "prefix: /v1" becomes a single `prefix: /api/v1` conditions.
- Repeated identical `header:` conditions (the same `name`, match type, and `invert` values) are deduplicated to a single `header:` condition.

docs/httpproxy.md
```diff
@@ -750,7 +750,172 @@ spec:
...
+Header conditions are added to the conditions array.
+
+### Configuring inclusion
+
+Inclusion is a top-level part of the HTTPProxy `spec` element.
+It requires two fields, `name` and `namespace`, with the option of including a `conditions` block.
```
davecheney: nit: namespace isn't required. Maybe it is, probably worth checking.

docs/httpproxy.md
```diff
@@ -750,7 +750,172 @@ spec:
...
+metadata:
+  name: multiple-root-www
+  namespace: default
+spec:
+  virtualhost:
+    fqdn: www.bar.com
+  includes:
+  - name: main
+    namespace: default
+    conditions:
```
davecheney: the condition can be dropped here, its not necessary

docs/httpproxy.md
youngnick: It's definitely required at the moment. I don't think it should be, but wanted to document current state.

davecheney:
Ok, my mistake.

docs/httpproxy.md
```diff
@@ -750,7 +750,173 @@ spec:
...
+These conditions are added to any conditions on the routes included.
+This process is recursive.
+
+Conditions are sets of individual condition statements, for example `prefix: /blog` is the condition that the matching request's path must start with `/blog`.
+When conditions are combined through inclusion Contour merges the conditions inherited via inclusion with any conditions specified on the route.
+This may result in duplicates, for example two `prefix:` conditions, or two header match conditions with the same name and value.
+To resolve this Contour applies the following logic.
+
+- `prefix:` conditions are concatenated together in the order they were applied from the root object. For example the conditions, `prefix: /api`, `prefix: /v1` becomes a single `prefix: /api/v1` conditions.
+- Repeated identical `header:` conditions (the same `name`, match type, and `invert` values) are deduplicated to a single `header:` condition.
```
stevesloka: The `invert` language is abstracted away from users. They only see `exact` or `notexact` to set that flag. 

Headers aren't merged either. We should discuss what should happen in this case. Should there be some sort of merge? Throw an error back to the user? Or just let them all stack ending up in a potential situation where requests won't match since the route is too specific.

docs/httpproxy.md
```diff
@@ -750,7 +750,173 @@ spec:
...
+
+Inclusion is a top-level part of the HTTPProxy `spec` element.
+It requires two fields, `name` and `namespace`, with the option of including a `conditions` block.
+
+#### Within the same namespace
+
+HTTPProxies can include other HTTPProxy objects in the namespace by specifying the name of the object and its namespace in the top-level `includes` block.
+Note that `includes` is a list, and so it must use the YAML list construct.
+
+In this example, the HTTPProxy `delegation-root` has included the configuration for paths matching `/service2` from the HTTPPRoxy named `service2` in the same namespace as `delegation-root` (the `default` namespace).
```
stevesloka: I still use "delegation" in speaking about "include", but should we try and stray away? I'm not sure if it's confusing to users who know IngressRoute or not. In a sense, we're performing the same operation but in a different way. This is just me thinking out loud.

docs/httpproxy.md
```diff
@@ -750,7 +750,173 @@ spec:
...
+It's important to note that `service2` HTTPProxy has not defined a `virtualhost` property as it is NOT a root HTTPProxy.
```
stevesloka: s/service2/www/g ??

docs/httpproxy.md
```diff
@@ -750,7 +750,173 @@ spec:
...
+    - conditions:
+      - prefix: /blog # matches /service2/blog
+      services:
+        - name: blog
+          port: 80
+```
+
+#### Virtualhost aliases
+
+To present the same set of routes under multiple dns entries, for example www.example.com and example.com, including a service with a `prefix` condition of `/` can be used.
```
stevesloka: Another way to do this so that you don't have to repeat the `includes` section twice is to have each root include to a single Proxy, then that one includes a child proxy. You end up with an additional Proxy object, but changes to the includes that are passed down are in a single place.

stevesloka: Your example is 100% correct, again just me typing out loud.

docs/httpproxy.md
youngnick: For now, I think they need to just be there, and we say that tautological conditions are an error.

Identical headers will end up deduplicated though, right?

youngnick: I'll change the invert part, yes.

docs/httpproxy.md
youngnick: I should change the names, so that it's consistent, at least. Thanks.

I think that they're functionally the same, and it's probably okay in conversation, but I was trying to avoid being confusing in this reference documentation.

docs/httpproxy.md
youngnick: Yeah, too much copypasta.

docs/httpproxy.md
youngnick: I think it's okay to be a bit more verbose in this, since it's a reference guide.

docs/httpproxy.md
stevesloka: Currently, master the headers are all just smashed together, so if someone duplicates headers, the route must match all the defined headers to route properly.

There's an issue open (https://github.com/projectcontour/contour/issues/1559), which I thought we would define an error if headers were duplicated. Thoughts on that approach? I'm not sure if use-cases exist where a route should match multiple headers of the same key with different values.

docs/httpproxy.md
stevesloka: :+1:

docs/httpproxy.md
youngnick: I meant that they are deduplicated at the end when applied to Envoy - since setting a correctly-formed header is idempotent on a route (I assume). I think that *contradictory* headers should be an error, but exact matches (where the header name, value, and operator are the same)? That doesn't feel like an error to me.

youngnick: In any case, #1559 is probably the place for this discussion. Can we accept that this is the case for now, and get this one in?

stevesloka: It's fine to merge, just want to be careful that we don't explain functionality that doesn't exist or could change and we confuse folks right at the beginning.

PR #1576, opened 2019-09-25
internal/e2e: Add Header Condition tests to E2E route

stevesloka:

### DIFF
--- a/internal/e2e/rds_test.go
+++ b/internal/e2e/rds_test.go
@@ -17,8 +17,11 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"strings"
 	"testing"
 	"time"
+
+	"github.com/projectcontour/contour/internal/dag"
 
 	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
 	envoy_api_v2_route "github.com/envoyproxy/go-control-plane/envoy/api/v2/route"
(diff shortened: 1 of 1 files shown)


internal/e2e/rds_test.go
```diff
@@ -2579,6 +2582,17 @@ func websocketroute(c string) *envoy_api_v2_route.Route_Route {
 	return cl
 }
 
+func headerroute(c string) *envoy_api_v2_route.Route_Route {
+	cl := routecluster(c)
+	//cl.Route_mat
```
davecheney: nit: looks like debugging breadcrumb

stevesloka: Turns out the whole method isn't needed. =)

stevesloka:
I think there's an ordering problem with the tests and how the results are compared to the desired, sometimes they work, sometimes they fail.

davecheney:
```
                        envoy.Route(envoy.RoutePrefix("/"), routecluster("default/svc1/80/da39a3ee5e")),
                        envoy.Route(envoy.RoutePrefix("/", dag.HeaderCondition{
                                Name:      "x-header",
                                Value:     "abc",
                                MatchType: "exact",
                                Invert:    false,
                        }), routecluster("default/svc2/80/da39a3ee5e")),
```
I think the problem is the sorting of routes doesn't take headers into account. Let me see if I can fix this.

PR #1589, opened 2019-09-29
internal/dag: re-enable proxy tests

davecheney:
Re-enable some disabled httpproxy tests.

### DIFF
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -3932,7 +3932,6 @@ func TestDAGInsert(t *testing.T) {
 					want[l.Port] = l
 				}
 			}
-
 			opts := []cmp.Option{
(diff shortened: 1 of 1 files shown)


internal/dag/builder_test.go
```diff
@@ -4078,42 +4077,42 @@ func TestDAGRootNamespaces(t *testing.T) {
...
+			Routes: []projcontour.Route{{
+				Services: []projcontour.Service{{
+					Name: "kuard",
+					Port: 8080,
+				}},
+			}},
+		},
+	}
+
+	// proxy2 is like ir1, but in a different namespace
```
youngnick: Nit: 'proxy2 is like proxy1'?

davecheney: Fixed. Thank you.

PR #1601, opened 2019-10-01
Add client request timeout function

### LINKED ISSUE #1073: Contour should configure a client request timeout
In #1070 the timeout for idle connections was set to 60 seconds, but we punted on setting a request timeout because no reasonable value could be found.

This issue is to track the research on a reasonable value which should be applied for all requests. This is complicated by the following factor:

- request timeout is a property of the http connection manager attached to a listener. For https connections there is one manager per connection, but for http all vhosts share the same manager, this makes it complicated to provide a per ingress/ingressroute configuration.
- ideally we don't want to make this value configurable -- the best result will be choosing a value which is small enough to prevent sloloris style resource exhaustion, and large enough to permit gigabyte style uploads.
- possibly http_connection_manager.request_timeout is not the right parameter. We may be better served elsewhere.

youngnick:
Uses the Envoy Connection Manager request_timeout setting.

Fixes #1073.

### DIFF
--- a/internal/contour/listener.go
+++ b/internal/contour/listener.go
@@ -16,6 +16,7 @@ package contour
 import (
 	"sort"
 	"sync"
+	"time"
 
 	envoy_api_v2_auth "github.com/envoyproxy/go-control-plane/envoy/api/v2/auth"
 	envoy_api_v2_listener "github.com/envoyproxy/go-control-plane/envoy/api/v2/listener"
@@ -83,6 +84,9 @@ type ListenerVisitorConfig struct {
 	// Valid entries are the keys from internal/envoy/accesslog.go:jsonheaders
 	// Defaults to a particular set of fields.
 	AccessLogFields []string
+
+	// RequestTimeout configures the request_timeout for all Connection Managers.
+	RequestTimeout time.Duration
 }
 
 // httpAddress returns the port for the HTTP (non TLS)
@@ -173,6 +177,19 @@ func (lvc *ListenerVisitorConfig) newSecureAccessLog() []*envoy_api_v2_accesslog
 	default:
 		return envoy.FileAccessLogEnvoy(lvc.httpsAccessLog())
 	}
+}
+
+// requestTimeout sets any durations in lvc.RequestTimeout <0 to 0 so that Envoy ends up with a positive duration.
+// for the request_timeout value we are passing, there are only two valid values:
+// 0 - disabled
+// >0 duration - the timeout.
+// The value may be unset, but we always set it to 0.
+func (lvc *ListenerVisitorConfig) requestTimeout() time.Duration {
+
+	if lvc.RequestTimeout < 0 {
+		return 0
+	}
+	return lvc.RequestTimeout
 }
 
 // minProtocolVersion returns the requested minimum TLS protocol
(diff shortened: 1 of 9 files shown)


internal/e2e/lds_test.go
```diff
@@ -86,7 +87,7 @@ func TestNonTLSListener(t *testing.T) {
 			&v2.Listener{
 				Name:         "ingress_http",
 				Address:      envoy.SocketAddress("0.0.0.0", 8080),
-				FilterChains: envoy.FilterChains(envoy.HTTPConnectionManager("ingress_http", envoy.FileAccessLogEnvoy("/dev/stdout"))),
+				FilterChains: envoy.FilterChains(envoy.HTTPConnectionManager("ingress_http", envoy.FileAccessLogEnvoy("/dev/stdout"), ptypes.DurationProto(0))),
```
youngnick: I've put the zero value as simply as I could, not sure if that or having a helper of some sort is better.

internal/envoy/listener.go
```diff
@@ -66,7 +67,8 @@ func Listener(name, address string, port int, lf []*envoy_api_v2_listener.Listen
 
 // HTTPConnectionManager creates a new HTTP Connection Manager filter
 // for the supplied route and access log.
```
davecheney: Somehow this function got renamed.

internal/envoy/listener.go
```diff
@@ -66,7 +67,8 @@ func Listener(name, address string, port int, lf []*envoy_api_v2_listener.Listen
...
-func HTTPConnectionManager(routename string, accesslogger []*accesslog.AccessLog) *envoy_api_v2_listener.Filter {
+func HTTPConsnectionManager(routename string, accesslogger []*accesslog.AccessLog, requestTimeout *duration.Duration) *envoy_api_v2_listener.Filter {
```
davecheney: I'm sorry to go back and forth on this. this is just a minor complaint but it looks like there are no code paths where requestTimeout is nil so there can never be a case where the requestTimeout we pass to Envoy is nil -- its always explicitly set to a &proto.Duration{}. 

If that's the case then it is probably going to be lighter on the page to pass in a time.Duration to HTTPConnection manager. 
```
func HTTPConnectionManager(routename string, accesslogger []*accesslog.AccessLog, requestTimeout time.Duration) *envoy_api_v2_listener.Filter
```

internal/e2e/lds_test.go
davecheney: If we can define the type of duration as time.Duration, then you can just write
```
FilterChains: envoy.FilterChains(envoy.HTTPConnectionManager("ingress_http", envoy.FileAccessLogEnvoy("/dev/stdout"),0))
```

internal/envoy/listener.go
youngnick: That makes sense to me. I'll change like follows:
- time.Duration to HTTPConnectionManager.
- HTTPConnectionManager will turn that into a duration.Duration (either 0 or >0).

internal/envoy/listener.go
youngnick: FFF, how did that happen, 🤦‍♂ I'll fix.

internal/envoy/listener.go
youngnick: In the end, I put the validation logic into the requestTimeout method on the ListenerVisitorConfig, and moved HTTPConnection manager to time.Duration.

internal/contour/listener.go
```diff
@@ -175,6 +179,19 @@ func (lvc *ListenerVisitorConfig) newSecureAccessLog() []*envoy_api_v2_accesslog
...
 }
 
+// requestTimeout sets any durations in lvc.RequestTimeout <0 to 0 so that Envoy ends up with a positive duration.
+// for the request_timeout value we are passing, there are only two valid values:
+// 0 - disabled
+// >0 duration - the timeout.
+// The value may be unset, but we always set it to 0.
+func (lvc *ListenerVisitorConfig) requestTimeout() time.Duration {
+
+	if lvc.RequestTimeout <= 0 {
```
davecheney: please write like this

if lvc.RequestTimeout < 0 {
   return 0
}
return lvc.RequestTimeout
```
rationale

0 * time.Second is still zero, but its of type time.Duration. If you want to write that time.Duration(0) is clearer. 

Use < 0 rather than <= 0 because zero is not special, we only need to care about negative numbers.

youngnick: thanks, that is great feedback. Will do.

PR #1606, opened 2019-10-01
Ignore this pr fail check

ffahri:
ignore


davecheney:
Thanks but it shouldn't be necessary to experiment with travis as the travis build just runs make check. If make check passes on your machine, there's a good chance it will pass in CI.

PR #1607, opened 2019-10-01
Fix HTTPProxy include so it assumes the same namespace if not present

### LINKED ISSUE #1574: internal/dag: HTTPProxy inlclude should not require namespace
In the event that a user wants to add includes to an HTTPProxy in the same namespace, it should not be required that they specify the namespace. When the namespace is blank, then Contour should look for Proxies in the same namespace where the Include is located.

youngnick:
Fixes #1574.

### DIFF
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -2091,6 +2091,49 @@ func TestDAGInsert(t *testing.T) {
 		},
 	}
 
+	// proxy101 and proxy101a test inclusion without a specified namespace.
+	proxy101 := &projcontour.HTTPProxy{
+		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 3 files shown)


internal/dag/builder_test.go
```diff
@@ -2091,6 +2091,49 @@ func TestDAGInsert(t *testing.T) {
...
+					Port: 8080,
+				}},
+			}},
+		},
+	}
+
+	proxy101a := &projcontour.HTTPProxy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "kuarder",
+			Namespace: s1.Namespace,
```
davecheney: nit: if you say `Namespace: proxy101.Namespace` it's super clear that 101 and 101a are in the same namespace.

PR #1608, opened 2019-10-01
internal/dag: validate TLS secrets during cache insertion

### LINKED ISSUE #1065: Contour should validate TLS certificates before serving them to Envoy
#1051 showed that an invalid certificate served to Envoy will break TLS handling badly. #1054 mitigated some of the problems by checking if the secret was present, but empty. What Contour should do is validate that the certificate and key are valid by trying to load them, as envoy does, before serving them to Envoy.

I've marked this as needing a design document, but a sufficiently well thought out comment on this issue will suffice.

Some of the wrinkles in implementing this are;

- [ ] validating certificates is potentially a CPU intensive operation, when that cost is paid is important to consider.
- [ ] we need some way of reporting an invalid certificate -- ie has bytes, but the byte aren't correct. This should probably be reported on the secret object, but I'm not sure if that is possible. Failing that, if we have to report it on the ingress/ingressroute object that references the secret, that affect where validation occurs.

davecheney:
Fixes #1065

Extend the secret validation to inspect the contents of the secret
object to try to avoid sending malformed certificate data to envoy.

For this PR log the failure at the point of insertion. This should keep
the log volume low, unless some automated process is hammering the
secret with noop edits.

The result of rejecting the secret at cache insertion time is
SecureVirtualHosts or HTTPProxies that rely on certificate validation
will no longer be able to find their previously broken secrets and will
be excluded during the build. This is unlikely to be noticable by end
users as an invalid (not unsigned or expired, but garbage) certificate
would previously crash Envoy.


davecheney:
Hold up, this isn't quite ready for review. I have one too many layers of base64 decode in there by mistake. I've fixed the code but ran out of time to fix the tests tonight. I'll push a new revision in the morning.

stevesloka:
Sounds good! I was just starting to poke through it. =)

davecheney:
@stevesloka @youngnick ok, this is ready for review now

internal/dag/secret.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"crypto/x509"
+	"encoding/pem"
+	"errors"
+
+	v1 "k8s.io/api/core/v1"
+)
+
+// validateSecret returns true if the secret is interesting and well formed.
+func validateSecret(secret *v1.Secret) (bool, error) {
```
stevesloka: nit: Should the name of this method be more descriptive? Like `validSecret` or something so the caller understands that `true` is valid.

davecheney: Will do. Sadly `validSecret` is already taken by the callback we pass to `lookupSecret`.

PR #1610, opened 2019-10-01
internal/dag: Look for duplicate header conditions and mark Proxy as invalid

### LINKED ISSUE #1559: internal/dag: Look for duplicate headers in a route configuration

stevesloka:
Fixes #1559 by looking for any route that has duplicate header keys and marking them as invalid if found. The logic this includes is to only mark duplicate headers with the "exact" match type. I could see folks setting up multiple "notexact" or "notcontains" headers which should be logical match types.

### DIFF
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -541,9 +541,17 @@ func (b *Builder) computeRoutes(sw *ObjectStatusWriter, proxy *projcontour.HTTPP
 			return nil
 		}
 
+		conds := append(conditions, route.Conditions...)
+
+		// Look for duplicate exact match headers on this route
+		if !headerConditionsAreValid(conds) {
+			sw.SetInvalid("cannot specify duplicate header 'exact match' conditions in the same route")
+			return nil
+		}
+
 		r := &Route{
-			PathCondition:    pathCondition(append(conditions, route.Conditions...)),
-			HeaderConditions: headerConditions(append(conditions, route.Conditions...)),
+			PathCondition:    pathCondition(conds),
+			HeaderConditions: headerConditions(conds),
 			Websocket:        route.EnableWebsockets,
 			HTTPSUpgrade:     routeEnforceTLS(enforceTLS, route.PermitInsecure && !b.DisablePermitInsecure),
 			PrefixRewrite:    route.PrefixRewrite,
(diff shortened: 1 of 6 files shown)


internal/dag/builder.go
```diff
@@ -518,6 +518,13 @@ func (b *Builder) computeRoutes(sw *ObjectStatusWriter, proxy *projcontour.HTTPP
 			}
 
 			sw, commit := b.WithObject(delegate)
+
+			//	conds := append(conditions, include.Conditions...)
+			//if !headerConditionsAreValid(conds) {
+			//	sw.SetInvalid("cannot specify duplicate header conditions in the same include")
+			//	return nil
+			//}
+
```
davecheney: please don't commit commented out code.

internal/dag/conditions.go
```diff
@@ -70,3 +70,22 @@ func headerConditions(conds []projcontour.Condition) []HeaderCondition {
...
+	// Look for duplicate headers on conditions
+	// if found, set error condition on HTTPProxy
+	encountered := map[string]bool{}
+	for _, v := range conditions {
+		if v.Header == nil {
+			continue
+		}
+		switch {
+		case v.Header.Exact != "":
+			if encountered[v.Header.Name] {
```
davecheney: I think this needs to be a case insensitive comparison; this test case https://github.com/projectcontour/contour/blob/master/internal/dag/conditions_test.go#L196 should fail.

stevesloka:
> TL;DR I'm not sure we should throw out duplicate header matching conditions. I think there is a legitimate use case, you can easily say header "foo:" contains both "a" and a later conditions is contains "b". We should agree on the business logic before duking it out in test cases.

This code only throws out duplicate headers if they are of type `exact` match. My thought was that if you specified header `foo: a` should match exactly, then later also said that it should match `foo: b` you'd essentially create a scenario where a route won't get requests. 

Every other type of header is allowed. For example, I could say that a request does not contain "foo: a" as well as should not contain "foo: b".

davecheney:
Ahh ok. I’m sorry that was my mistake, I didn’t read the code clearly enough.

stevesloka:
Let me clean up the code a bit to make that more clear.

Issue #1457, opened 2019-09-05
internal/dag: Enable additional fields in Condition


stevesloka:
**Describe the solution you'd like**
An httplb can contain `[conditions](https://github.com/heptio/contour/blob/master/apis/projectcontour/v1alpha1/httploadbalancer.go#L44)`, as an `Include` or as part of a `Route` definition. Currently, we've implemented `pathPrefix`, but there are additional items to implement:

- [x] HeadersMatch
- [x] HeadersContain
- [x] ExactMatch: https://github.com/heptio/contour/blob/master/design/routing-design.md#exact-path-routing
- [ ] Wildcard Path (i.e. Glob routing): https://github.com/projectcontour/contour/blob/master/design/routing-design.md#path-wildcard

davecheney:
Sadly we won't be delivering `prefix:` wildcard condition for Contour 1.0. The design is close from a user facing perspective, but there are some subtle implementation details that remain unanswered and we want to proceed carefully there in case we paint outselves into a corner.

I'm moving this ticket to the backlog as a placeholder for a epic ticket which will capture the business requirements for wildcard routing.

sunjayBhatia:
removing important-soon as this is an httplb issue, not httpproxy

PR #1612, opened 2019-10-01
internal: Add wildcard path prefix to HTTPProxy

### LINKED ISSUE #1179: Wildcard path matching like nginx
We'd like the ability to use a wildcard for matching the path.  As part of the IngressRoute, we'd like to do the following

```
...
spec:
  ...
  routes:
  - match: /path/to/object/[a-zA-Z0-9_]+/logs
     service:
     - name: log-service
        port: 9000
  -match: /path/to/object/[a-zA-Z0-9_]+
    - service:
       -name: object-management-service:
         port: 9000
```
I know this could be done by putting logs at the beginning and eliminating the need for regular expression matching, but the team *really* wants to have the path to the object precede the `logs` in the path and the back-end wants a separate service for the logging functionality.

nginx Ingress Controller has support for regular expression matching: https://kubernetes.github.io/ingress-nginx/user-guide/ingress-path-matching/

**Environment:**

- Contour version: 0.12.1
```
kubectl version
Client Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.11", GitCommit:"637c7e288581ee40ab4ca210618a89a555b6e7e9", GitTreeState:"clean", BuildDate:"2018-11-26T14:38:32Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.7", GitCommit:"6f482974b76db3f1e0f5d24605a9d1d38fad9a2b", GitTreeState:"clean", BuildDate:"2019-03-25T02:41:57Z", GoVersion:"go1.10.8", Compiler:"gc", Platform:"linux/amd64"}
```
- Kubernetes installer & version: Azure AKS, Kubernetes 1.12.7
- Cloud provider or hardware configuration: Azure
- OS (e.g. from `/etc/os-release`): Centos 7

stevesloka:
Fixes #1179 
Updates #1457 

Adds a new Condition to the `prefix` condition already existing. Users can now specify a pathPrefix which includes a wildcard in the path (e.g. `/foo/*/blog`). 

This means that the prefix would match `/foo/[anyvalid]/bar*` since it's still treated like a path prefix. 

In the event this in part of an include, the same rules apply where paths are appended together. 

### DIFF
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -2134,6 +2134,28 @@ func TestDAGInsert(t *testing.T) {
 		},
 	}
 
+	// proxy102 has a wildcard path prefix
+	proxy102 := &projcontour.HTTPProxy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "example-com",
+			Namespace: s1.Namespace,
+		},
+		Spec: projcontour.HTTPProxySpec{
+			VirtualHost: &projcontour.VirtualHost{
+				Fqdn: "example.com",
+			},
+			Routes: []projcontour.Route{{
+				Conditions: []projcontour.Condition{{
+					Prefix: "/foo/*/bar",
+				}},
+				Services: []projcontour.Service{{
+					Name: s1.Name,
+					Port: 8080,
+				}},
+			}},
+		},
+	}
+
 	tests := map[string]struct {
 		objs                  []interface{}
 		disablePermitInsecure bool
@@ -4152,6 +4174,30 @@ func TestDAGInsert(t *testing.T) {
 				},
 			),
 		},
+		"insert httpproxy with wildcard route prefix": {
+			objs: []interface{}{
+				proxy102,
+				s1,
+			},
+			want: listeners(
+				&Listener{
+					Port: 80,
+					VirtualHosts: virtualhosts(
+						virtualhost("example.com",
+							routeClusterWildcard("/foo/*/bar",
+								&Cluster{
+									Upstream: &Service{
+										Name:        s1.Name,
+										Namespace:   s1.Namespace,
+										ServicePort: &s1.Spec.Ports[0],
+									},
+								},
+							),
+						),
+					),
+				},
+			),
+		},
 	}
 
 	for name, tc := range tests {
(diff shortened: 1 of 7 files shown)


stevesloka:
This PR lays the groundwork, there should be a followup to add the status logic rules. We discussed previously (as well as in the design doc), that you can only have a wildcard in the middle of the paths (e.g. /something/*/else). 

Merging conditions together produces the same prefix style match. So `/api` & `/v1/*/docs` becomes `/api/v1/*/docs*`. 

The problem I see with prefix/suffix is that it matches more than you might want. So if your path was `/api/*/users` it would match `/api` and `/foo/bar/users` which may not be the intent of the user.

I guess the design doc needs more detail? Is this still a 1.0 item?

stevesloka:
Also you cannot have multiple match types on a route if I recall correctly.

davecheney:
Thank you for working on this but I do not want to proceed with this PR until we have a design for wildcards or globbing. 

I'm very worried that we will not be able to put the genie back into the bottle so I don't want to proceed until the following have been thought through

1. Wildcard sorting; at the moment we sort routes by longest prefix. How does globbing or wildcards affect this. Eg. does `wildcard: /api/*/foo` sort before or after `prefix: /api/v2` ?
2. Can wildcards be merged and under what conditions can they be merged? It seems likely that users will expect to combine a wildcard include condition with a regular prefix condition in the included httpproxy, and vice versa. eg, what does `wildcard: /api/*` and `prefix: /foo/` merge too? 
2b. What would `wildcard: *` mean mean? Should we enforce that all wildcards start with a `/`? Should we also enforce that they end with a `/`?
3. A the moment path prefixes are treated like directory names, `prefix: /api`, plus `prefix: /v1` becomes `prefix: /api/v1`. But what if someone wanted to write `prefix: /users/d`, to capture `/users/david` and `/users/daemon` ? If they wrote this as a wildcard would they write `wildcard: /users/d*`? Is that what is intended, or should we restrict it to be more like globbing, eg `wildcard: /api/*` is actually `^/api/.+/$`.

I think there are a lot of questions to be answered and I don't want to move to code and committing to them until the implications are understood.

Thank you

pavelnikolov:
Are there any plans to work on wildcard matches? I need it for URL rewriting not for routing. I want to be able to have the prefix/replacement use regular expressions.

skriss:
Please continue conversation on #4241, thanks.

PR #1618, opened 2019-10-02
internal/dag: Set status on HTTPProxy if multiple path conditions are specified

### LINKED ISSUE #1611: internal/dag: Duplicate prefix conditions should be invalid
If a set of conditions is defined in an include or route, duplicate `prefix` conditions should mark the `Proxy` status as invalid. 

Example:
```
spec:
  routes:
  - conditions:
    - prefix: /
    - prefix: /foo
    - prefix: /bar
```

stevesloka:
Fixes #1611 
Set status on HTTPProxy if multiple path conditions are specified on a Route or Include.

Invalid Route Conditions:
```
kind: HTTPProxy
metadata:
  name: root
spec:
  virtualhost:
    fqdn: kind.local
  routes:
  - services:
    - name: app
      port: 80
    conditions:
      - prefix: /v1
      - prefix: /api
```
Invalid Include Conditions:
```
kind: HTTPProxy
metadata:
  name: root
spec:
  virtualhost:
    fqdn: kind.local
  includes:
  - name: blogsite
    namespace: teama
    conditions:
    - prefix: /v1
    - prefix: /api
  routes:
  - services:
    - name: app
      port: 80
```

### DIFF
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -517,6 +517,11 @@ func (b *Builder) computeRoutes(sw *ObjectStatusWriter, proxy *projcontour.HTTPP
 				return nil
 			}
 
+			if !pathConditionsValid(include.Conditions) {
+				sw.SetInvalid("include: cannot specify multiple path conditions in the same include")
+				return nil
+			}
+
 			sw, commit := b.WithObject(delegate)
 			routes = append(routes, b.computeRoutes(sw, delegate, append(conditions, include.Conditions...), visited, enforceTLS)...)
 			commit()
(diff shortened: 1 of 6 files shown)


internal/dag/builder.go
```diff
@@ -517,6 +517,11 @@ func (b *Builder) computeRoutes(sw *ObjectStatusWriter, proxy *projcontour.HTTPP
 				return nil
 			}
 
+			if !pathConditionsValid(include.Conditions) {
+				sw.SetInvalid("include: cannot specify multiple path conditions in the same route")
```
davecheney: s/same route/same include

PR #1629, opened 2019-10-03
docs: document our supported release policy

### LINKED ISSUE #1581: site: publish a support policy for older versions
The current, unspoken, support policy for Contour is we only support the current stable (`:latest`) version. 

For 1.0 we should make sure this policy is recorded on projectcontour.io.
For 1.0 we should revisit if this policy is appropriate and if we should extend it older versions of Contour.

davecheney:
Fixes #1581


docs/support.md
```diff
@@ -0,0 +1,16 @@
+# Supported version policy
+
+This document describes which versions of Contour are supported by the Contour team.
+
+## Stable release
+
+Only the latest stable release is supported.
+The latest stable release is identified by the [Docker tag `:latest`](../tagging.md).
```
stevesloka: We should also utilize Github's "Preview" release functionality, so `latest` docker tag matches up to github release's `latest` release.

Issue #1119, opened 2019-05-22
Contour should not 503 traffic to a down service if there is another service available in the Route


hmorikaw:
What steps did you take and what happened:

On my K8s cluster I am using Contour ingress, running on Gimbal Cluster.
2 Contour pods running separated with 10 Envoy pods and each Envoy pod combines 2 services.
Envoy pods check endpoint nodes on each service.
```
apiVersion: contour.heptio.com/v1beta1
kind: IngressRoute
metadata:
  name: foo
  namespace: foobar
spec:
  virtualhost:
    fqdn: foobar.com
  routes:
    - match: /
      services:
        - name: service0
          port: 80
          healthCheck:
            path: /healthy
            intervalSeconds: 3
            timeoutSeconds: 1
            unhealthyThresholdCount: 3
            healthyThresholdCount: 5
          strategy: WeightedLeastRequest
        - name: service1
          port: 80
          strategy: WeightedLeastRequest
          healthCheck:
            path: /healthy
            intervalSeconds: 3
            timeoutSeconds: 1
            unhealthyThresholdCount: 3
            healthyThresholdCount: 5
```

After all endpoint nodes on service1 are out of order,
each Envoy pod does not update its fowaring rule for service1 and returns only 'no healthy upstream' status.

What did you expect to happen:

If all endpoints were down on a service(service1), Envoy pods update its traffic rule to forward packets for only service0 until at least one endpoint on service1 is fine.

Anything else you would like to add:

Environment:

Contour version: v0.11.0
Kubernetes version: (use kubectl version): 1.14.0
Kubernetes installer & version: 1.14.0
Cloud provider or hardware configuration: Xeon E5-2683v4 2.10GHz / 2CPU / 512GBMEM RAM / SATA SSD 240GB x 1 / 10G Base-T*2
OS (e.g. from /etc/os-release):  Ubuntu 16.04.6 LTS
CNI : Calico v3.4

stevesloka:
Hey @ hmorikaw I'd expect that with two upstreams, the other service would take the traffic if there weren't any healthy endpoints. Let me see if I can recreate the issue.

hmorikaw:
Sorry, @stevesloka .
After  we open this issue, we can't recreate this issue.

hmorikaw:
@stevesloka 
Hi, how about this issue? Can you recreate this problem?
We still remain it.
If you can't recreate unhealthy endpoints cluster, you stop all http servers on unhealthy endpoints cluster. 
[...]

PR #1630, opened 2019-10-04
api: move HTTPProxy health check to route level

davecheney:
Updates #1119

This PR moves the HealthCheck stanza from the individual service level
to the route. This is an insurance policy if we have to move from one
cluster per service to using one service per route and multiple
clusterloadassignments. If that happens, we will not be able to support
per service configuration.

Specifically:

- Fork httpproxy/v1/healthCheck into ingressroute/v1beta/healthCheck
retaining the original feature set.
- Rename httpproxy/v1/healthCheck to healthCheckPolicy, making it
consistent with timeoutPolicy and retryPolicy.
- The underlying type is renamed to HTTPHeathCheckPolicy as it is
attached to Route.
- Healthchecks on httproxy.tcpproxy are removed. We can add them back in
rc.1 or rc.2.


davecheney:
The intention of this PR is _if_ we don't need to move to one cluster per route, we can restore the health check per backend setting as an addition to the per route health check. The former, if present, will take precedence over the latter.

stevesloka:
The original reason it was on the service was that each upstream could possibly have a different health check. For Gimbal, one service could be k8s and the other openstack. 

I know we struggled on the design originally if they would have different health checks or not in practice, so defaulted to allow them to have their own. 

TLDR: I think it's probably fine to have a single active healthcheck defined.

stevesloka:
Something to think about for this design change might be the passive health checks, if we enabled those in the future, what might this health check policy look like? (Fair enough to call YAGNI as well).

davecheney:
> The original reason it was on the service was that each upstream could possibly have a different health check. For Gimbal, one service could be k8s and the other openstack.
> 
> I know we struggled on the design originally if they would have different health checks or not in practice, so defaulted to allow them to have their own.
> 
> TLDR: I think it's probably fine to have a single active healthcheck defined.

Yeah, I'm not happy about rolling this back -- per service load balancing and health checks is very useful. 

The worry that I have is if we have to move to one cluster per route to resolve #1119 then that means having to deprecate the route.service.healthcheck field. Maybe we can table this change and look into #1119 some more. This doesn't need to happen this week, but I think we need to have a plan in place to move, or remove, the fields that would block us from moving to one cluster per route if required.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -105,23 +105,25 @@ type TLS struct {
...
-	// Indicates that during forwarding, the matched prefix (or path) should be swapped with this value
+	// Indicates that during forwarding, the matched prefix (or path) should be swapped with this value.
 	PrefixRewrite string `json:"prefixRewrite,omitempty"`
-	// The timeout policy for this route
+	// The timeout policy for this route.
 	TimeoutPolicy *TimeoutPolicy `json:"timeoutPolicy,omitempty"`
-	// The retry policy for this route
+	// The retry policy for this route.
 	RetryPolicy *RetryPolicy `json:"retryPolicy,omitempty"`
+	// The health check policy this route.
```
youngnick: Nit: 'for this route'.

PR #1637, opened 2019-10-05
Fix namespaces of tls cert delegation example

yvespp:


davecheney:
@yvespp thank you for this fix. Can you please check that we haven't made the same mistake in docs/httpproxy.md

davecheney:
@yvespp this project uses DCO please sign your commit before we can review it. Thank you.

yvespp:
docs/httpproxy.md is ok, I changed the example to use the same namespaces in both places.
Maybe the example should be removed as it is duplicate?

PR #1641, opened 2019-10-07
internal/dag: hard code secret version and kind

### LINKED ISSUE #1624: internal/dag: kind and version are not showing up in secret validation error
#1608  added logging of the details off a failing secret, but kind and version are missing

```
time="2019-10-03T06:03:31Z" level=error msg="failed to parse PEM" context=KubernetesCache kind= name=kuard namespace=default version=
```

davecheney:
Fixes #1624

For some reason v1.Secret's don't have a valid type and kind. However,
this type is known to be v1.Secret, so hard code this.


davecheney:
Yeah, its super weird that classic v1 objects don't have a reported version and kind. We need to a k8s great beard to explain why core objects don't have type and kind--my guess is because its not stored in the api server because reasons.

PR #1643, opened 2019-10-07
internal/dag: Set default CircuitBreaking defaults for both MaxConnections/MaxRequests

### LINKED ISSUE #1192: configurable default circuit breaking config

stevesloka:
Fixes #1192  by setting a default for MaxConnections as well as MaxRequests. Annotating the service will still use a user-defined value so still can be overridden. 

### DIFF
--- a/internal/contour/cluster_test.go
+++ b/internal/contour/cluster_test.go
@@ -218,6 +218,12 @@ func TestClusterVisit(t *testing.T) {
 					ConnectTimeout: protobuf.Duration(250 * time.Millisecond),
 					LbPolicy:       v2.Cluster_ROUND_ROBIN,
 					CommonLbConfig: envoy.ClusterCommonLBConfig(),
+					CircuitBreakers: &envoy_api_v2_cluster.CircuitBreakers{
+						Thresholds: []*envoy_api_v2_cluster.CircuitBreakers_Thresholds{{
+							MaxConnections: protobuf.UInt32(100000),
+							MaxRequests:    protobuf.UInt32(100000),
+						}},
+					},
 				}),
 		},
 		"single named service": {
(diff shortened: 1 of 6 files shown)


davecheney:
@stevesloka sorry, most of my comment is not relevant as I thought this was a fix for #1119. 

My question about the limit of 100,000 (sorry, I read that as 10,000) still applies. What is the justification for that number?

stevesloka:
No real backing, I asked around and couldn't find a good set of numbers which is why I originally had it in a config file. 

This is just a random start that is higher than the default, happy to change if others have opinions on the default.

davecheney:
I'd like to see some discussion about the right value for these defaults, https://github.com/projectcontour/contour/issues/1192#issuecomment-518311310 suggests that 100,000 is too low for max connections.

If the real value for these settings should be "unlimited" then I'd like to find a way to express that directly in envoy rather than an arbitrary 100,000. My concern is this number is likely to be enshrined in lore -- people will quote it in slack, others will write it in performance tuning docs, etc. Its not that the limit is 100,000 its just an arbitrary number that was chosen when the real intention was to make the default connection limit perhaps unlimited.

frankh:
perfect is the enemy of good

due to this PR being closed 5 years ago, i've run now into issues with the absurdly low default max connections of 1024

PR #1646, opened 2019-10-07
internal/dag: do not attempt to validate ca.crt

davecheney:
Updates #1164

Skip validation of ca.crt.


stevesloka:
Is there value in validating if it does exist? Just wondering if skipping all together is too much.

davecheney:
I want to do better than this, but I needed to back port something to 0.15 to make sure we didn't break people upgrading to 0.15.1. I'll do better for rc.1

Issue #1597, opened 2019-09-30
Trailing slash on prefix match stripped


davecheney:
```yams
spec:
- routes:
  - conditions:
     - prefix: "/a/"
```
Will be translated into a prefix match of `/a`, the trailing slash will be stripped.

davecheney:
Thinking about this some more I think we may have to revise how prefix merging works.

I made the classic mistake of assuming URLs were directory paths when I insisted on using `path.Join`. This mistake meant the only way to delegate `prefix: /users/a`, that is all users starting with `a` is a include condition `prefix: /users/a` and no prefix condition on the included route. If the user tried to combine `prefix: /users/a` with `prefix: /` the result would be `prefix: /users/a/` not `prefix: /users/a`.

I suggest we do the following.

1. Stop using path.Join. That was my bad, I wasn't thinking.
2. merge prefix conditions by concatenation, `prefix: /foo` and `prefix: /bar` would become `prefix: foo/bar`, but `prefix: /foo` and `prefix: bar` would become `prefix: /foobar`.
3. We probably need to do some clever cleaning of the resulting url`, prefix: /` and `prefix: /foo` should not become `prefix: //foo`. However `prefix: /` and `prefix: /foo/` should merge to `prefix: /foo/`, the trailing slash should not be stripped.

stevesloka:
What makes it difficult is if you have a path `prefix: /foo` and `prefix: /` should that become `/foo` or `/foo/`? 

The logic we have now is that if you do not specify a path condition, it defaults to "/", my thought was to make that field optional so users wouldn't have to explicitly add the "condition" field all over the place. 

I never thought of merging conditions like in your example `prefix: /foo and prefix: bar would become prefix: /foobar.` We need to be careful that we keep the delegation model working properly. 

I always thought we'd only allow full paths separated by `/`, but allowing the previous example does make it more flexible at the end given users who might need that additional complexity, but it does add to the UX problem.

davecheney:
> What makes it difficult is if you have a path prefix: /foo and prefix: / should that become /foo or /foo/?

Good question. I think under the current logic and my proposed change `prefix: /foo` and `prefix: /` would result in `prefix: /foo/`. 

> The logic we have now is that if you do not specify a path condition, it defaults to "/", my thought was to make that field optional so users wouldn't have to explicitly add the "condition" field all over the place.

[...]

Issue #1575, opened 2019-09-25
Contour should handled malformed CRDs better


stevesloka:
**Describe the solution you'd like**
If a user builds a bad / malformed CRD (i.e. yaml) and submits that to the API server, Contour should not log errors over and over when parsing the object. 

A quick example is to make the `conditions` items not a list:

```
apiVersion: projectcontour.io/v1alpha1
kind: HTTPProxy
metadata:
  name: root
spec:
  virtualhost:
    fqdn: kind.local
  routes:
  - conditions:
    - prefix: /
    services:
    - name: kuard
      port: 80
  - services:
    - name: default
      port: 80
  includes:
  - name: wwwsite
    namespace: teama
    conditions: 
     prefix: /blog  # <------------- missing a "-" (should be -prefix: /blog`
```

Example output every second:
```
W0925 10:29:41.048480   42792 reflector.go:302] pkg/mod/k8s.io/client-go@v0.0.0-20190918200256-06eb1244587a/tools/cache/reflector.go:98: watch of *v1alpha1.HTTPProxy ended with: an error on the server ("unable to decode an event from the watch stream: unable to decode watch event: v1alpha1.HTTPProxy.Spec: v1alpha1.HTTPProxySpec.Includes: []v1alpha1.Include: v1alpha1.Include.Conditions: []v1alpha1.Condition: decode slice: expect [ or n, but found {, error found in #10 byte of ...|ditions\":{\"prefix\":\"|..., bigger context ...|c7254152cabc\"},\"spec\":{\"includes\":[{\"conditions\":{\"prefix\":\"/blog\"},\"name\":\"wwwsite\",\"namespace\":\"te|...") has prevented the request from succeeding
[...]

PR #1650, opened 2019-10-08
Require prefix conditions to start with a /

### LINKED ISSUE #1628: All prefix: conditions must start with a /
Breaking out from https://github.com/projectcontour/contour/issues/1597#issuecomment-537754073

I propose that all `prefix: ` conditions, if present, must start with a `/`. Specifically

- At most one `prefix: ` condition may be present in each `include.conditions` or `route.conditions` block.
- If present the `prefix: ` condition must start with a `/`.

youngnick:
Fixes #1628
Updates #1597

### DIFF
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -535,8 +535,7 @@ func (b *Builder) computeRoutes(sw *ObjectStatusWriter, proxy *projcontour.HTTPP
 				return nil
 			}
 
-			if !pathConditionsValid(include.Conditions) {
-				sw.SetInvalid("include: cannot specify multiple path conditions in the same include")
+			if !pathConditionsValid(sw, include.Conditions, "include") {
 				return nil
 			}
 
@@ -554,8 +553,7 @@ func (b *Builder) computeRoutes(sw *ObjectStatusWriter, proxy *projcontour.HTTPP
 			continue
 		}
 
-		if !pathConditionsValid(route.Conditions) {
-			sw.SetInvalid("route: cannot specify multiple path conditions in the same route")
+		if !pathConditionsValid(sw, route.Conditions, "route") {
 			return nil
 		}
 
--- a/internal/dag/conditions.go
+++ b/internal/dag/conditions.go
@@ -14,6 +14,7 @@
 package dag
 
 import (
+	"fmt"
 	"path"
 	"strings"
 
@@ -30,13 +31,20 @@ func pathCondition(conds []projcontour.Condition) Condition {
 	}
 }
 
-func pathConditionsValid(conds []projcontour.Condition) bool {
-	found := 0
+// pathConditionsValid validates a slice of Conditions can be correctly merged.
+// It encodes the business rules about what is allowed for prefix Conditions.
+func pathConditionsValid(sw *ObjectStatusWriter, conds []projcontour.Condition, conditionsContext string) bool {
+	prefixCount := 0
 	for _, cond := range conds {
 		if cond.Prefix != "" {
-			found++
+			prefixCount++
+			if cond.Prefix[0] != '/' {
+				sw.SetInvalid(fmt.Sprintf("%s: Prefix conditions must start with /, %s was supplied", conditionsContext, cond.Prefix))
+				return false
+			}
 		}
-		if found > 1 {
+		if prefixCount > 1 {
+			sw.SetInvalid(fmt.Sprintf("%s: More than one prefix is not allowed in a condition block", conditionsContext))
 			return false
 		}
 	}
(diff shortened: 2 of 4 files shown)


internal/dag/conditions.go
```diff
@@ -30,13 +31,21 @@ func pathCondition(conds []projcontour.Condition) Condition {
 	}
 }
 
-func pathConditionsValid(conds []projcontour.Condition) bool {
+// pathConditionsValid validates a slice of Conditions.
+// It encodes the business rules about what is allowed for prefix Conditions.
+func pathConditionsValid(conds []projcontour.Condition, sw *ObjectStatusWriter, conditionsContext string) bool {
```
youngnick: I'm not a big fan of passing this conditionsContext string, but without it there is no way at all to tell which conditions block is invalid in a HTTPProxy document. Solving this properly requires solving #1575 as well, so this is here for now.

internal/dag/builder.go
```diff
@@ -517,8 +517,7 @@ func (b *Builder) computeRoutes(sw *ObjectStatusWriter, proxy *projcontour.HTTPP
 				return nil
 			}
 
-			if !pathConditionsValid(include.Conditions) {
-				sw.SetInvalid("include: cannot specify multiple path conditions in the same include")
+			if !pathConditionsValid(include.Conditions, sw, "include") {
```
davecheney: suggestion: for consistency, please move sw to the front of the parameter list.

internal/dag/conditions.go
```diff
@@ -30,13 +31,21 @@ func pathCondition(conds []projcontour.Condition) Condition {
 	}
 }
 
-func pathConditionsValid(conds []projcontour.Condition) bool {
+// pathConditionsValid validates a slice of Conditions.
```
davecheney: ... validates a slice of conditions can be correctly merged.

internal/dag/conditions.go
```diff
@@ -30,13 +31,21 @@ func pathCondition(conds []projcontour.Condition) Condition {
...
+// It encodes the business rules about what is allowed for prefix Conditions.
+func pathConditionsValid(conds []projcontour.Condition, sw *ObjectStatusWriter, conditionsContext string) bool {
 	found := 0
 	for _, cond := range conds {
 		if cond.Prefix != "" {
+			// We found a prefix, so increment the prefix counter.
```
davecheney: I don't think this commend adds a lot. Perhaps something like renaming `found` to `prefixCount` would make the code more self explanatory.

internal/dag/conditions_test.go
```diff
@@ -270,11 +271,37 @@ func TestConditionsValid(t *testing.T) {
...
 	}
 
 	for name, tc := range tests {
 		t.Run(name, func(t *testing.T) {
-			got := pathConditionsValid(tc.conditions)
+			// TODO(youngnick) This feels dirty but I don't know a better solution.
+			// We're not testing the returned info from the ObjectStatusWriter here,
+			// that's done at the status_test.go layer.
+			swblank := &ObjectStatusWriter{}
+			sw, _ := swblank.WithObject(&projcontour.HTTPProxy{
```
davecheney: yeah this is a bit unfortunate. Perhaps in the future status writer should become an interface then we can supply a dummy.

internal/dag/conditions.go
youngnick: I actually started doing that when I was rewriting into a single verify function, and forgot to bring it back when I reverted. Agreed, I'll do that now.

internal/dag/conditions_test.go
youngnick: Ah, that makes sense. Consider my mental note made, but I'll put a TODO there and make an issue.

internal/dag/builder.go
youngnick: Missed that, no problem.

internal/dag/conditions.go
```diff
@@ -30,13 +31,20 @@ func pathCondition(conds []projcontour.Condition) Condition {
...
-	found := 0
+// pathConditionsValid validates a slice of Conditions can be correctly merged.
+// It encodes the business rules about what is allowed for prefix Conditions.
+func pathConditionsValid(sw *ObjectStatusWriter, conds []projcontour.Condition, conditionsContext string) bool {
+	prefixCount := 0
 	for _, cond := range conds {
 		if cond.Prefix != "" {
-			found++
+			prefixCount++
+			if !strings.HasPrefix(cond.Prefix, "/") {
```
davecheney: suggestion: we know that cond.Prefix != "" at this point so you can just write
```
if cond.Prefix[0] != '/' {
```

davecheney: more correctly len(cond.Prefix) > 0 so its safe to access subscript 0

youngnick: of course. Neat.

PR #1656, opened 2019-10-09
Remove unused test function 'clustertimeout'.

jpeach:


davecheney:
Thanks for spotting this.

To make it easier to write release notes, please assign PR's that are ready for the review to the milestone you'd like them to be review in. This is sort of a weak signal that they're ready for review, but also makes it easy to write up the release notes.

PR #1660, opened 2019-10-09
Change prefix condition merge logic to not use path.Join()

### LINKED ISSUE #1597: Trailing slash on prefix match stripped

youngnick:
Fixes #1597.

Also clarify the documentation language around prefix condtions
to make the requirements clear.

### DIFF
--- a/docs/httpproxy.md
+++ b/docs/httpproxy.md
@@ -58,7 +58,7 @@ spec:
 **Line 6-7**: The presence of the `virtualhost` field indicates that this is a root HTTPProxy that is the top level entry point for this domain.
 The `fqdn` field specifies the fully qualified domain name that will be used to match against `Host:` HTTP headers.
 
-**Lines 8-9**: Each HTTPProxy must have one or more `routes`, each of which must have a condition to match against (e.g. `prefix: /`) and then one or more `services` which will handle the HTTP traffic.
+**Lines 8-9**: Each HTTPProxy **must** have one or more routes, each of which **must** have one or more services which will handle the HTTP traffic. In addition, each route **may** have one or more conditions to match against.
 
 **Lines 10-12**: The `services` field is an array of named Service & Port combinations that will be used for this HTTPProxy path.
 HTTP traffic will be sent directly to the Endpoints corresponding to the Service.
@@ -326,11 +326,18 @@ In this example, the permission for Contour to reference the Secret `example-com
 
 ### Conditions
 
-Each Route entry in a HTTPProxy may contain one or more conditions.
+Each Route entry in a HTTPProxy **may** contain one or more conditions.
 These conditions are combined with an AND operator on the route passed to Envoy.
 
 Conditions can be either a `prefix` or a `header` condition.
+
+#### Prefix conditions
+
 For `prefix`, this adds a path prefix.
+
+Up to one prefix condition may be present in any condition block.
+
+Prefix conditions **must** start with a `/` if they are present.
 
 #### Header conditions
 
@@ -771,7 +778,7 @@ Because the path is not necessarily used as the only key, the route space can be
 
 ### Conditions and Inclusion
 
-Like Routes, Inclusion may specify a set of [conditions](#conditions.
+Like Routes, Inclusion may specify a set of [conditions](#conditions).
 These conditions are added to any conditions on the routes included.
 This process is recursive.
 
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -2351,6 +2351,182 @@ func TestDAGInsert(t *testing.T) {
 		},
 	}
 
+	proxy104 := &projcontour.HTTPProxy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "example-com",
+			Namespace: s1.Namespace,
+		},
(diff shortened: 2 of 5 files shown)


docs/httpproxy.md
```diff
@@ -58,7 +58,7 @@ spec:
 **Line 6-7**: The presence of the `virtualhost` field indicates that this is a root HTTPProxy that is the top level entry point for this domain.
 The `fqdn` field specifies the fully qualified domain name that will be used to match against `Host:` HTTP headers.
 
-**Lines 8-9**: Each HTTPProxy must have one or more `routes`, each of which must have a condition to match against (e.g. `prefix: /`) and then one or more `services` which will handle the HTTP traffic.
+**Lines 8-9**: Each HTTPProxy **must** have one or more routes, each of which **must** have one or more services which will handle the HTTP traffic. In addition, each route **may** have one or more conditions to match against.
```
davecheney: > In addition, each route **may** have one or more conditions to match against.

I think this sentence would be better around line 330.

internal/dag/conditions.go
```diff
@@ -15,17 +15,34 @@ package dag
...
 )
 
-func pathCondition(conds []projcontour.Condition) Condition {
-	prefix := "/"
+// mergePathConditions merges the given slice of prefix Conditions into a single
+// prefix Condition.
+// pathConditionsValid guarantees that if a prefix is present, it will start with a
+// / character, so we can simply concatenate.
+func mergePathConditions(conds []projcontour.Condition) Condition {
+	prefix := ""
```
davecheney: what about
```
prefix := "/"
trailingSlash := false
for _, cond := range cones {
    cp := cond.Prefix
    if cp != "" {
          trailingSlash = cp[len(cp)-1] == '/'
    }
    prefix = path.Join(prefix, cp)
}
if trailingSlash {
   prefix += "/"
}
```

internal/dag/conditions_test.go
```diff
@@ -56,8 +56,7 @@ func TestPathCondition(t *testing.T) {
 			conditions: []projcontour.Condition{{
 				Prefix: "/a/",
 			}},
-			// TODO(dfc) issue 1597
-			want: &PrefixCondition{Prefix: "/a"},
+			want: &PrefixCondition{Prefix: "/a/"},
```
davecheney: For completeness please add a test for prefix: /a, prefix: /b/

docs/httpproxy.md
youngnick: There's a line there already:
> Each Route entry in a HTTPProxy may contain one or more conditions.

I can change the emphasis a bit there if you think that's better.

internal/dag/conditions.go
youngnick: Sure, that seems reasonable. Certainly saves on a regexp usage, which seems nice.

docs/httpproxy.md
davecheney: Cool, if that note is mentioned ~330 thats perfect.

internal/dag/conditions.go
youngnick: Actually, turns out that path.Join does not trim repeated `/` characters, so I'm going to stick with the current string concatenation implementation.

davecheney:
Leaving for @jpeach and @stevesloka

internal/dag/builder_test.go
```diff
@@ -2235,6 +2235,138 @@ func TestDAGInsert(t *testing.T) {
...
+
+	proxy105a := &projcontour.HTTPProxy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "kuarder",
+			Namespace: proxy105.Namespace,
+		},
+		Spec: projcontour.HTTPProxySpec{
+			Routes: []projcontour.Route{{
+				Conditions: []projcontour.Condition{{
+					Prefix: "/",
```
jpeach: I'm fairly sure that the ordering guarantee here is that the `/kuarder` prefix from the parent inclusion will come first, but if you added a path component (e.g. "/app") to the route prefix condition (along with a supporting comment), it would be clearer.

Maybe an additional test case?

internal/dag/conditions_test.go
```diff
@@ -56,8 +56,16 @@ func TestPathCondition(t *testing.T) {
...
+		},
+		"trailing slash on second prefix condition": {
+			conditions: []projcontour.Condition{{
+				Prefix: "/a",
+			},
+				{
+					Prefix: "/b/",
+				}},
+			want: &PrefixCondition{Prefix: "/a/b/"},
```
jpeach: Consider adding a "doubled slashes case" like:
```Go
conditions: []projcontour.Condition{{
                                  Prefix: "////",
                         }, {
                                 Prefix: "",
                         }},
```

youngnick: I've added something like this, but note that an empty string for a prefix is not possible, the conditions validation will reject them before they get to this point.

internal/dag/builder_test.go
youngnick: Added an additional test case.

Issue #899, opened 2019-02-19
prefixRewrite includes does not strip slash from prefix


davecheney:
Report from Slack, 
```
apiVersion: contour.heptio.com/v1beta1
kind: IngressRoute
metadata: 
  labels:
    app: test
  name: test
  namespace: default
spec: 
  virtualhost:
    fqdn: test.local
  routes: 
    - match: /routeapp
      prefixRewrite: "/"
      services: 
        - name: routeapp
          port: 80
```
Requests to `/routeapp/test` get rewritten as `//test` not `/test`

neeseius:
I brought this to attention. Looks like changing "/routeapp" to "/routeapp/" fixes the issue.

davecheney:
I'm not able to reproduce the problem you reported. Without a way to reproduce this, I cannot verify that I have fixed the bug.

gaelleacas:
Hi :) 

I would like to add an exemple that reproduce this issue with prefix path : 
I use  nodejs app with express routing:
  - expose my app to http://main.local/myapp 
  - A route is set to http://main.local/myapp/myroute

```
const express = require('express');
const app = express();
app.get('/', (req, res) => {
  res.setHeader("Content-Type", "text/html");
  res.status(200).send('Hello World from root');
});

app.get('/myroute', (req, res) => {
  res.setHeader('Content-Type', 'text/plain');
  res.status(200).send('Hello World from route /myroute');
});
app.listen(process.env.PORT || 8080);
```

Deploy & service: 
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: mynodeapp
  labels:
    app: mynodeapp
spec:
  replicas: 1
  selector:
    matchLabels:
      app: mynodeapp
  strategy:
    rollingUpdate:
      maxSurge: 1
      maxUnavailable: 0
    type: RollingUpdate
  minReadySeconds: 5
  template:
    metadata:
      labels:
        app: mynodeapp
    spec:
      containers:
      - name: mynodeapp
        image: docker.io/khaly/mynodeapp:v1
        imagePullPolicy: Always
---
apiVersion: v1
kind: Service
metadata:
  name: mynodeapp
  labels:
    app: mynodeap
spec:
  ports:
  - port: 80
    protocol: TCP
    targetPort: 8080
  type: ClusterIP
  selector:
    app: mynodeapp
```
Ingressroute with prefixRewrite :

```
apiVersion: contour.heptio.com/v1beta1
kind: IngressRoute
[...]

PR #1672, opened 2019-10-09
api: remove HTTPProxy.Route.PrefixRewrite

### LINKED ISSUE #1667: api,internal/dag: remove prefixRewrite
In order to properly address #899 we plan to remove the current `prefixRewrite` field from HTTPProxy. The field will remain on the deprecated IngressRoute. 

We plan to replace prefixRewrite with something better in rc.2

davecheney:
Fixes #1667
Updates #899

This PR removes route.prefixRewrite from HTTPProxy in rc.1. The plan is
to replace it with something better in rc.2. We are removing this now in
rc.1 to avoid blocking moving HTTPProxy to v1 status in rc.1.

Please follow #899 for the replacement design which is scheduled for
rc.2


davecheney:
PR is all crapped up.

PR #1683, opened 2019-10-10
all: migrate to networking/v1beta1.Ingress

### LINKED ISSUE #1633: support for networking.k8s.io/v1beta1.Ingress
It looks like contour currently only supports the deprecated `extensions/v1beta1.Ingress` API (or the CRD based API). See https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/ for more information.

We're already in progress of migrating our resources to the new `networking.k8s.io/v1beta1.Ingress` API and have some dev. clusters where the old APIs are disabled in the api-server through the `runtime-config` flag. On these clusters contour is not working properly because of the missing API:
```
E1004 05:28:35.486715       1 reflector.go:125] pkg/mod/k8s.io/client-go@v0.0.0-20190918200256-06eb1244587a/tools/cache/reflector.go:98: Failed to list *v1beta1.Ingress: the server could not find the requested resource
```

Ingress objects which are persited through `networking.k8s.io` API can also be retrieved via the old API (if not disabled 😄) so this can be considered as "feature enhancement request".

**Environment:**

- Contour version: `master`
```
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.1", GitCommit:"d647ddbd755faf07169599a625faf302ffc34458", GitTreeState:"clean", BuildDate:"2019-10-02T16:51:36Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
```
```
kubeadm version: &version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.1", GitCommit:"d647ddbd755faf07169599a625faf302ffc34458", GitTreeState:"clean", BuildDate:"2019-10-02T16:58:27Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
```
- relevant api-server flag:
```
--runtime-config=apps/v1beta1=false,apps/v1beta2=false,extensions/v1beta1/daemonsets=false,extensions/v1beta1/deployments=false,extensions/v1beta1/replicasets=false,extensions/v1beta1/networkpolicies=false,extensions/v1beta1/podsecuritypolicies=false,extensions/v1beta1/ingresses=false
```

Thanks in advance and let me know if you need further info and/or help.

davecheney:
Fixes #1633

This PR changes Contour to use the networking/v1beta1.Ingress version of
the Ingress object. Contour uses this object internally and relies on
the API server to convert the older extensions/v1beta1 form to the current
networking/v1beta1 form.  Where this translation is not possible because
the cluster is too old, Contour performs this translation internally.

The catch is when relying on this translation mechanism, registering for
*both* extensions/v1beta1 and networking/v1beta1 causes objects that are
natively in the former format to appear twice. Thus, contour can only
subscribe to one of extensions/v1beta1 or networking/v1beta1. Thus
this PR provides a new flag, --use-extensions-v1beta1-ingress to
subscribe to the older form in pre 1.13 clusters.

### DIFF
--- a/examples/contour/03-contour.yaml
+++ b/examples/contour/03-contour.yaml
@@ -37,6 +37,7 @@ spec:
       - args:
         - serve
         - --incluster
+        - --use-extensions-v1beta1-ingress
         - --xds-address=0.0.0.0
         - --xds-port=8001
         - --envoy-service-http-port=80
--- a/examples/render/contour.yaml
+++ b/examples/render/contour.yaml
@@ -327,6 +327,14 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - "networking.k8s.io"
+  resources:
+  - ingresses
+  verbs:
+  - get
+  - list
+  - watch
 - apiGroups: ["contour.heptio.com"]
   resources: ["ingressroutes", "tlscertificatedelegations"]
   verbs:
@@ -465,6 +473,7 @@ spec:
       - args:
         - serve
         - --incluster
+        - --use-extensions-v1beta1-ingress
         - --xds-address=0.0.0.0
         - --xds-port=8001
         - --envoy-service-http-port=80
--- a/internal/dag/cache.go
+++ b/internal/dag/cache.go
@@ -14,8 +14,12 @@
 package dag
 
 import (
+	"bytes"
+	"encoding/json"
+
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/api/extensions/v1beta1"
+	extensionsv1beta1 "k8s.io/api/extensions/v1beta1"
+	"k8s.io/api/networking/v1beta1"
 	"k8s.io/client-go/tools/cache"
 
 	ingressroutev1 "github.com/projectcontour/contour/apis/contour/v1beta1"
@@ -105,6 +109,16 @@ func (kc *KubernetesCache) Insert(obj interface{}) bool {
 		}
 		kc.ingresses[m] = obj
 		return true
+	case *extensionsv1beta1.Ingress:
+		ingress := new(v1beta1.Ingress)
+		if err := transposeIngress(obj, ingress); err != nil {
+			om := obj.GetObjectMeta()
+			kc.WithField("name", om.GetName()).
+				WithField("namespace", om.GetNamespace()).
+				Error(err)
+			return false
+		}
+		return kc.Insert(ingress)
 	case *ingressroutev1.IngressRoute:
 		class := ingressClass(obj)
 		if class != "" && class != kc.ingressClass() {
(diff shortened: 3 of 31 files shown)


davecheney:
Draft only for now. I've done some smoke testing on my 1.14.6 cluster with extensions/v1beta1 Ingress objects but there are a lot of other permutations to check.

davecheney:
Marking as ready for review.

I've tested this on a 1.12 cluster and found it transparently supports translating extension/v1beta1 Ingress to networking/v1beta1 so there was no need to pass `--use-extensions-v1beta1-ingress` even on this ancient cluster.

davecheney:
Ok, I think this is ready to land. I worked back through the versions of k8s I have access to

- v1.12.9-gke.15; screams like a stuck pig unless --use-extensions-v1beta1-ingress
- v1.13.10-gke.0; ditto
- v1.14.6-gke.2; fine

What I think we should do is add `--use-extensions-v1beta1-ingress` to the example deployment; that is, not make it the default, but make it the implicit default for anyone installing a cluster today. If in the rare case they are using 1.16 _and_ have disabled the extensions/v1beta1 object then they can remove the flag. 

For 1.1, we'll remove the flag from the example, thereby making networking.k8s.io/v1beta1 the default, and by the time 1.1 ships v1.14 or later will probably be the oldest version people can reasonably run -- at the moment gke is defaulting to 1.13.

examples/contour/03-contour.yaml
```diff
@@ -37,6 +37,7 @@ spec:
       - args:
         - serve
         - --incluster
+        - --use-extensions-v1beta1-ingres
```
stevesloka: missing an `s` at the end (e.g `ingress`)

examples/render/contour.yaml
```diff
@@ -465,6 +473,7 @@ spec:
       - args:
         - serve
         - --incluster
+        - --use-extensions-v1beta1-ingres
```
stevesloka: missing an s at the end (e.g ingress)

internal/dag/cache.go
```diff
@@ -105,6 +109,16 @@ func (kc *KubernetesCache) Insert(obj interface{}) bool {
...
+	case *extensionsv1beta1.Ingress:
+		ingress := new(v1beta1.Ingress)
+		if err := translateIngress(obj, ingress); err != nil {
+			om := obj.GetObjectMeta()
+			kc.WithField("name", om.GetName()).
+				WithField("namespace", om.GetNamespace()).
+				Error(err)
+			return false
+		}
+		return kc.Insert(ingress)
```
stevesloka: If the translate fails for some reason, should we log it? Also, in that case we'll insert an empty ingress. This not a stopper for this PR, just an observation.

stevesloka:
Also I'm doing a quick test on my kind cluster to validate again.

stevesloka:
LGTM!! Nice work @davecheney ! 🎉

davecheney:
This code is trying to log and then return. I’ll check the logic and double check it does what I expects it to do

internal/dag/cache.go
davecheney: If translateIngress returns and error we log it then return false rather than falling through to kc.Insert.

PR #1688, opened 2019-10-11
internal/envoy: Remove obsolete `ConfigWriter` type

jpeach:


jpeach:
> On 11 Oct 2019, at 2:36 pm, Dave Cheney <notifications@github.com> wrote:
> 
> @davecheney requested changes on this pull request.
> 
> Thanks for working on this. Rather than creating an otherwise empty file to hold the package comment, is there a existing file in the internal/envoy package that would be a natural home for the package comment?

I moved the package comment to `internal/envoy/bootstrap.go`, which seems like the next best option.

PR #1703, opened 2019-10-15
internal/contour: order the route header match list

### LINKED ISSUE #1579: internal/contour: need to assert that route objects with header conditions sort properly
https://github.com/projectcontour/contour/pull/1576#issuecomment-535402749

jpeach:
Ensure that the HeaderMatch slice on each Route is ordered. This
lets us endure that Route messages which have the same prefix and
number of header matches sort stably and will not cause unnecessary
Envoy API updates.

The previous code ordered the routes by first by the length of the
header match slice and then by the length of the prefix or regex
match. This change reverses that, but there is no net change in how
routes will be matched.

This fixes #1579.

### DIFF
--- a/internal/contour/route_test.go
+++ b/internal/contour/route_test.go
@@ -14,7 +14,6 @@
 package contour
 
 import (
-	"sort"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


jpeach:
Dammit, I just messed up this branch .. will fix :-/

jpeach:
Fixed the branch :)

internal/contour/route_test.go
```diff
@@ -1946,12 +1945,108 @@ func TestSortLongestRouteFirst(t *testing.T) {
 				Match: envoy.RoutePrefix("/"),
 			}},
 		},
+
+		// Verify that longest path sorts before longest
+		// headers. We used to sort by longest header list
+		// first, which does end up with the same net result,
+		// so this isn't strictly necessary.  However, ordering
+		// the patch first is arguably more intuitive, and
```
youngnick: Nit: s/patch/path/ ?

PR #1704, opened 2019-10-15
internal/dag: scan pem for secret key

### LINKED ISSUE #1702: TLS keys containing other pem blocks proceeding the private key are considered invalid
Created a secret whose `EC PRIVATE KEY` contains `EC PARAMETERS`.  The function `validatePrivateKey` locates the `EC PARAMETERS ` pem.Block, attempts to validate it as a private key, and rejects the secret as valid.

I expected the secret to be accepted as valid :)

The private key pem struct is something like this

```
-----BEGIN EC PARAMETERS-----
-----END EC PARAMETERS-----
-----BEGIN EC PRIVATE KEY-----
-----END EC PRIVATE KEY-----
```

I did scan, but didn't find a similar issue.

**Environment:**

- Contour version: v0.15.1 and master

mattalberts:
Walk each PEM block to identify the PRIVATE KEY

closes #1702


internal/dag/secret.go
```diff
@@ -59,9 +60,9 @@ func validateCertificate(data []byte) error {
 }
 
 func validatePrivateKey(data []byte) error {
-	key, err := decodePEM(data)
+	key, err := decodePEMPrivateKey(data)
 	if err != nil {
-		return nil
+		return err
```
mattalberts: I wasn't sure why err was not previously returned (failure to parse the PEM should be an error?)

internal/dag/secret.go
```diff
@@ -82,3 +83,21 @@ func decodePEM(in []byte) (*pem.Block, error) {
...
+		}
+		if strings.HasSuffix(block.Type, "PRIVATE KEY") {
+			found = true
+			break
+		}
+	}
+	if !found {
+		return nil, errors.New("failed to locate private key")
+	}
+	return block, nil
```
mattalberts: there is an opportunity to be ""smarter" with the outbound block and switch on type to identify the parsing method.

internal/dag/secret.go
```diff
@@ -82,3 +83,16 @@ func decodePEM(in []byte) (*pem.Block, error) {
...
 }
+
+func decodePEMPrivateKey(in []byte) (*pem.Block, error) {
+	var block *pem.Block
+	for len(in) > 0 {
+		if block, in = pem.Decode(in); block == nil {
+			return nil, errors.New("failed to parse PEM")
+		}
+		if strings.HasSuffix(block.Type, "PRIVATE KEY") {
+			return block, nil
```
mattalberts: there is an opportunity to be ""smarter" with the outbound block and switch on type to identify the appropriate parsing method.

jpeach:
I can review tomorrow, if no-one else gets there first :)

mattalberts:
Thanks!  I very much drafted this change set in the "backport to v0.15.1" mindset.   That said, the certificate parsing/validation has a similar flaw (to less detriment).  It will accept any cert whose first pem block is a valid cert.  I'll do mock up an alternate that might be better for mainline

youngnick:
I think that #1707 is the correct approach here, so I'll close this one out. Thanks for all your work @mattalberts!

PR #1707, opened 2019-10-15
internal/dag: validate each pem block

### LINKED ISSUE #1702: TLS keys containing other pem blocks proceeding the private key are considered invalid

mattalberts:
During certificate validation, walk the PEM and validate each
appropriate block.

closes #1702

### DIFF
--- a/internal/dag/secret.go
+++ b/internal/dag/secret.go
@@ -50,35 +50,63 @@ func isValidSecret(secret *v1.Secret) (bool, error) {
 }
 
 func validateCertificate(data []byte) error {
-	cert, err := decodePEM(data)
-	if err != nil {
-		return err
+	var exists bool
+	for len(data) > 0 {
+		var block *pem.Block
+		block, data = pem.Decode(data)
+		if block == nil {
+			return errors.New("failed to parse PEM block")
+		}
+		if block.Type != "CERTIFICATE" {
+			return errors.New("unexpected block type in certificate: " + block.Type)
+		}
+		if _, err := x509.ParseCertificate(block.Bytes); err != nil {
+			return err
+		}
+		exists = true
 	}
-	_, err = x509.ParseCertificate(cert.Bytes)
-	return err
+	if !exists {
+		return errors.New("failed to locate certificate")
+	}
+	return nil
 }
 
 func validatePrivateKey(data []byte) error {
-	key, err := decodePEM(data)
-	if err != nil {
+	var keys int
+	for len(data) > 0 {
+		var block *pem.Block
+		block, data = pem.Decode(data)
+		if block == nil {
+			return errors.New("failed to parse PEM block")
+		}
+		switch block.Type {
+		case "PRIVATE KEY":
+			if _, err := x509.ParsePKCS8PrivateKey(block.Bytes); err == nil {
+				return err
+			}
+			keys++
+		case "RSA PRIVATE KEY":
+			if _, err := x509.ParsePKCS1PrivateKey(block.Bytes); err == nil {
+				return err
+			}
+			keys++
+		case "EC PRIVATE KEY":
+			if _, err := x509.ParseECPrivateKey(block.Bytes); err != nil {
+				return err
+			}
+			keys++
+		case "EC PARAMETERS":
+			// ignored
+		default:
+			return errors.New("unexpected block type in private key: " + block.Type)
+		}
+	}
+	switch keys {
+	case 0:
+		return errors.New("failed to locate private key")
+	case 1:
 		return nil
+	default:
+		return errors.New("multiple private keys")
 	}
-	if _, err := x509.ParsePKCS1PrivateKey(key.Bytes); err == nil {
-		return nil
-	}
-	if _, err := x509.ParsePKCS8PrivateKey(key.Bytes); err == nil {
-		return nil
-	}
-	if _, err := x509.ParseECPrivateKey(key.Bytes); err == nil {
-		return nil
-	}
-	return errors.New("unknown private key encoding")
-}
-
-func decodePEM(in []byte) (*pem.Block, error) {
-	block, _ := pem.Decode(in)
-	if block == nil {
-		return nil, errors.New("failed to parse PEM")
-	}
-	return block, nil
 }
--- a/internal/dag/cache_test.go
+++ b/internal/dag/cache_test.go
@@ -80,6 +80,30 @@ func TestKubernetesCacheInsert(t *testing.T) {
 			},
 			want: true,
 		},
+		"insert secret referenced by ingress with multiple pem blocks": {
+			pre: []interface{}{
+				&v1beta1.Ingress{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "www",
+						Namespace: "default",
+					},
+					Spec: v1beta1.IngressSpec{
+						TLS: []v1beta1.IngressTLS{{
+							SecretName: "secret",
+						}},
+					},
+				},
+			},
+			obj: &v1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "secret",
+					Namespace: "default",
+				},
+				Type: v1.SecretTypeTLS,
+				Data: secretdata(EC_CERTIFICATE, EC_PRIVATE_KEY),
+			},
+			want: true,
+		},
 		"insert secret w/ wrong type referenced by ingress": {
 			pre: []interface{}{
 				&v1beta1.Ingress{
--- a/internal/dag/secret_test.go
+++ b/internal/dag/secret_test.go
@@ -15,9 +15,8 @@ package dag
 
 import v1 "k8s.io/api/core/v1"
 
-// sample data from https://8gwifi.org/PemParserFunctions.jsp
-
 const (
+	// sample data from https://8gwifi.org/PemParserFunctions.jsp
 	CERTIFICATE = `-----BEGIN CERTIFICATE-----
 MIIFtTCCA52gAwIBAgIJAO0cq2lJPZZJMA0GCSqGSIb3DQEBBQUAMEUxCzAJBgNV
 BAYTAkFVMRMwEQYDVQQIEwpTb21lLVN0YXRlMSEwHwYDVQQKExhJbnRlcm5ldCBX
@@ -103,6 +102,30 @@ KMkcE4BT8IZIHQ+wIMhmYLAdSQCVVv8x78jN0sZCC0fjqVuyPdYQ8sIc3OHsJZcW
 lzewFW72lfsiB/RxWZ/XwXONXeW5Quf+XwbGGboTofyzTxzsYSwn1U9Kt8iaY8zr
 z7Z5SQCSf2Js9V9lJcodYswWlxrdtoRKA/WgrvQkZhGGAePTUVoO5Lab29M8
 -----END RSA PRIVATE KEY-----`
+
+	// sample elliptical curve data generated
+	// openssl ecparam -name prime256v1 -genkey -out ec_key.pem
+	// openssl req -new -x509 -key ec_key.pem -out ec_crt.pem -days 3650
+	EC_CERTIFICATE = `-----BEGIN CERTIFICATE-----
+MIIBbjCCARQCCQCPA0hmRaqduTAKBggqhkjOPQQDAjA/MQswCQYDVQQGEwJVUzEL
+MAkGA1UECAwCQ0ExEjAQBgNVBAcMCVBhbG8gQWx0bzEPMA0GA1UECgwGVk1XYXJl
+MB4XDTE5MTAxNTAzMzkzM1oXDTI5MTAxMjAzMzkzM1owPzELMAkGA1UEBhMCVVMx
+CzAJBgNVBAgMAkNBMRIwEAYDVQQHDAlQYWxvIEFsdG8xDzANBgNVBAoMBlZNV2Fy
+ZTBZMBMGByqGSM49AgEGCCqGSM49AwEHA0IABFUOHv4hnLcopcYdjojx2j/FmFX6
+MOLVVNsNpZ4SpmcGKN2zGp0SyAQgNhY0gGojC0g+VVYrh8X3GQAXYdvIjfMwCgYI
+KoZIzj0EAwIDSAAwRQIhAJudFacSiwcRtyQ2aNYAPbDJnnwbUTXRCVRlgLysgP5G
+AiALPSbO8d0wa24Z0AU2oXocuNkDaH8qEyp2yhL5LKI3Dw==
+-----END CERTIFICATE-----
+`
+	EC_PRIVATE_KEY = `-----BEGIN EC PARAMETERS-----
+BggqhkjOPQMBBw==
+-----END EC PARAMETERS-----
+-----BEGIN EC PRIVATE KEY-----
+MHcCAQEEIOlYOKzXGQTYlKDkuM62/U84DjxEOa8T3XGYlVmycFJroAoGCCqGSM49
+AwEHoUQDQgAEVQ4e/iGctyilxh2OiPHaP8WYVfow4tVU2w2lnhKmZwYo3bManRLI
+BCA2FjSAaiMLSD5VViuHxfcZABdh28iN8w==
+-----END EC PRIVATE KEY-----
+`
 )
 
 func secretdata(cert, key string) map[string][]byte {


internal/dag/secret.go
```diff
@@ -50,35 +51,47 @@ func isValidSecret(secret *v1.Secret) (bool, error) {
...
+	var block *pem.Block
+	for len(data) > 0 {
+		block, data = pem.Decode(data)
+		if block == nil {
+			return errors.New("failed to parse PEM block")
+		}
+		if !strings.HasSuffix(block.Type, "PRIVATE KEY") {
+			continue
+		}
+		if _, err := x509.ParsePKCS1PrivateKey(block.Bytes); err == nil {
```
mattalberts: There is a question as to early exit or continue looking for more than one private key (and error'ing as a result)

internal/dag/secret.go
```diff
@@ -50,35 +51,47 @@ func isValidSecret(secret *v1.Secret) (bool, error) {
...
-		return nil
-	}
-	if _, err := x509.ParsePKCS1PrivateKey(key.Bytes); err == nil {
-		return nil
-	}
-	if _, err := x509.ParsePKCS8PrivateKey(key.Bytes); err == nil {
-		return nil
+	var exists bool
+	var block *pem.Block
+	for rest := data[:]; len(rest) > 0; {
```
davecheney: rest := data[:] is a no-op, it copies the value data into a rest, but data is itself a copy of the value passed to validateCertificate, all three point to the same underlying array albeit with their own values for Len and cap.

This code is fine, but I'm worried that a casual reader will assume that data and rest are more independent than they are; I think

for len(data)> 0 {
    block, data := per.Decode(data)
    ...
}

would probably be less open to misinterpretation.

internal/dag/secret.go
```diff
@@ -50,35 +51,47 @@ func isValidSecret(secret *v1.Secret) (bool, error) {
...
-		return nil, errors.New("failed to parse PEM")
+func validatePrivateKey(data []byte) error {
+	var block *pem.Block
+	for len(data) > 0 {
+		block, data = pem.Decode(data)
+		if block == nil {
+			return errors.New("failed to parse PEM block")
+		}
+		if !strings.HasSuffix(block.Type, "PRIVATE KEY") {
+			continue
```
davecheney: if the private key doesn't contain a private key, should that be an error?

internal/dag/secret.go
davecheney: What does envoy require. I know stacking the intermediate ca's in a the certificate is common practice but what's the use case for having more than one private key block?

internal/dag/secret.go
davecheney: Looking at the example you provided I think it would be better if rather than ignoring blocks that didn't start with `PRIVATE KEY` the logic was inverted and it excluded only blocks it was expecting, ie, `EC PARAMETERS`

internal/dag/secret.go
jpeach: IIRC it's possible to have multiple private keys and select the right key to serve to a client depending on the cipher suite negotiation and other TLS properties.

internal/dag/secret.go
mattalberts: i meant to clean that up after i was done delve'ing :) its not a nit :)

mattalberts: I meant to drop that after i was done delv'ing.  That's my bad.

internal/dag/secret.go
mattalberts: I'll try to dig through the envoy source ti find what they accept, right now I terminate the loop after locating the first key

internal/dag/secret.go
mattalberts: If the private key doesn't contain a private key, the loop will terminate on `len(data)` and return `errors.New("failed to locate private key")`.  

The way I've structured it is likely leading to confusion.  Currently, it is seek to first block that ends with `PRIVATE KEY` (covering `PRIVATE KEY`, `RSA PRIVATE KEY` and `EC PRIVATE KEY`). Validate the block bytes as a key and return either `nil` or `errors.New("unknown private key encoding")`.

I'm a little worried about inversion (though I'm not against it). I didn't want to miss a block type or have a new block type show up that is treated/validated/rejected as a key.

mattalberts:
I've been reading through the conversation, its feels like there were two sources of questions
 - what does envoy (boring ssl) expect?
 - what is the behavior of the private key validation change

In the interest of crispness, I re-worked the private key validation section to more visually mirror the behavior of the certificate section and added a restriction around duplicate private keys (by key type).

Now, I'll go back to digging in the envoy code and

internal/dag/secret.go
davecheney: Would something like this work?
```
func validatePrivateKey(data []byte) error {
        var keys int
        for len(data) > 0 {
                var block *pem.Block
                block, data = pem.Decode(data)
                if block == nil {
                        return errors.New("failed to parse PEM block")
                }
                switch block.Type {
                case "PRIVATE KEY":
                        if _, err := x509.ParsePKCS8PrivateKey(block.Bytes); err != nil {
                                return err
                        }
                        keys++
                case "RSA PRIVATE KEY":
                        if _, err := x509.ParsePKCS1PrivateKey(block.Bytes); err != nil {
                                return err
                        }
                        keys++
                case "EC PRIVATE KEY":
                        if _, err := x509.ParseECPrivateKey(block.Bytes); err != nil {
                                return err
                        }
                        keys++
                case "EC PARAMETERS":
                        // ignored
                default:
                        return errors.New("unknown private key type: " + block.Type)
                }
        }
        switch keys {
        case 0:
                return errors.New("failed to locate private key")
        case 1:
                return nil
        default:
                return errors.New("multiple private keys")
        }
}
```

mattalberts: I was trying to avoid those string compares :P . Feels like you want to add
```
type BlockType string
const (
	PrivateKey BlockType = "PRIVATE KEY"
	RSAPrivateKey BlockType = "RSA PRIVATE KEY"
	ECPrivateKey BlockType = "EC PRIVATE KEY"
	Certificate BlockType = "CERTIFICATE"
)

switch BlockType(block.Type) {
}
```

From what I can tell its close.  It appears as though, it is common to attempt parsing "PRIVATE KEY" and "RSA PRIVATE KEY" with both `ParsePKCS1PrivateKey` and `ParsePKCS8PrivateKey`

davecheney:
Could you explain a little more about your concerns about string comparison?

internal/dag/secret.go
mattalberts: What do you think about this?

```
func validatePrivateKey(data []byte) error {
	var keys int
	for len(data) > 0 {
		var block *pem.Block
		block, data = pem.Decode(data)
		if block == nil {
			return errors.New("failed to parse PEM block")
		}
		switch block.Type {
		case "PRIVATE KEY":
			if err := validatePKCSPrivateKey(block.Bytes); err != nil {
				return err
			}
			keys++
		case "RSA PRIVATE KEY":
			if err := validatePKCSPrivateKey(block.Bytes); err != nil {
				return err
			}
			keys++
		case "EC PRIVATE KEY":
			if _, err := x509.ParseECPrivateKey(block.Bytes); err != nil {
				return err
			}
			keys++
		}
	}
	switch keys {
	case 0:
		return errors.New("failed to locate private key")
	case 1:
		return nil
	default:
		return errors.New("multiple private keys")
	}
}

func validatePKCSPrivateKey(in []byte) error {
	if _, err := x509.ParsePKCS1PrivateKey(in); err == nil {
		return nil
	}
	if _, err := x509.ParsePKCS8PrivateKey(in); err == nil {
		return nil
	}
	return errors.New("unknown PKCS encoding")
}
```
 > subtly different, if it is not a block it knows how to parse/validate, it skips it.

davecheney:
The problem the validation is trying to protect against is if some data which envoy does not know how to handle reaches envoy it will in the best case simply stop accepting updates—because they contain invalid data—or in some proper cases crash. With that requirement in mind I don’t want to design for extensibility if that re-opens the possibility of causing these issues.

jpeach:
FWIW, here are the PEM constants OpenSSL has for private keys:
```
# define PEM_STRING_EVP_PKEY     "ANY PRIVATE KEY"
# define PEM_STRING_RSA          "RSA PRIVATE KEY"
# define PEM_STRING_DSA          "DSA PRIVATE KEY"
# define PEM_STRING_PKCS8        "ENCRYPTED PRIVATE KEY"
# define PEM_STRING_PKCS8INF     "PRIVATE KEY"
# define PEM_STRING_ECPRIVATEKEY "EC PRIVATE KEY"
```

There's also a few different parameter types:
```
# define PEM_STRING_DHPARAMS     "DH PARAMETERS"
# define PEM_STRING_DHXPARAMS    "X9.42 DH PARAMETERS"
# define PEM_STRING_SSL_SESSION  "SSL SESSION PARAMETERS"
# define PEM_STRING_DSAPARAMS    "DSA PARAMETERS"
# define PEM_STRING_ECPARAMETERS "EC PARAMETERS"
```

mattalberts:
@davecheney understood.  I'll add the skip for `EC PARAMETERS` back in

mattalberts:
@jpeach the only one I've seen for sure, or expect at the `EC PARAMETERS`, did we want to include any others in the switch to skip?

davecheney:
I admit a complete lack of familiarity in this area but it seems that go supports three key encodings, which we decode now, and there is a set of parameter types. Without any direct experience it seems like that set of parameter types should be bounded and relatively small so we should be able to include that set without an extended discovery phase. Is there an RFC that captures the parameter types that will be commonly encountered in the wild?

mattalberts:
(as for the strings ... i started with a switch and then felt like I was gutting/complicating code that was meant to be fairly simple)

mattalberts:
I get the feeling that "DH PARAMETERS" and "DSA PARAMETERS" do get used.
 - https://blog.cloudflare.com/keyless-ssl-the-nitty-gritty-technical-details/

I can try asking nick :)

davecheney:
And the extra bit about adding a stack of parameter types that we just skip is what will envoy do if they aren’t valid? Skipping validating them does not abdicate the responsibility of envoy stalls on them. Although the current validation is too strict I would prefer to alway err on the side of being precise about what to pass through — only that which we have validated—because the penalty for rejecting some time Contour doesn’t know how to validate is less than letting through something envoy rejects.

jpeach:
> On 17 Oct 2019, at 12:44 pm, Dave Cheney <notifications@github.com> wrote:
> 
> I admit a complete lack of familiarity in this area but it seems that go supports three key encodings, which we decode now, and there is a set of parameter types. Without any direct experience it seems like that set of parameter types should be bounded and relatively small so we should be able to include that set without an extended discovery phase. Is there an RFC that captures the parameter types that will be commonly encountered in the wild?

I think that DH and EC params are the most likely types users would need.

I'm pretty sure that DSA cipher suites are considered weak and have fallen out of use. I'm not sure what (if anything) the general "PARAMETERS" type is for. SSL Parameters are for SSL session diagnostics, and I don't know what "X9.42 DH PARAMETERS" is :)

davecheney:
I’m sorry this is coming out in fits and spurts, I’m juggling several things today. 

The goal of validation is to validate, not to whitelist things. The delta between the things we whitelist as “know that type, don’t know how to validate” and validate as in “know that type, inspect it mechanically” should be as small as possible.

mattalberts:
:) . Glad I'm not the only one (also juggling too many things and feel like i'm just a gumbly stream of sometimes words).

mattalberts:
The only one i know for sure is the `EC PARAMETERS` do not break envoy .. so .. just those?

mattalberts:
https://security.stackexchange.com/questions/29778/why-does-openssl-writes-ec-parameters-when-generating-private-key

Maybe the answer is, I delete the `EC PARAMETERS` from my cert, we close this PR, and make no change.

jpeach:
> On 17 Oct 2019, at 1:44 pm, Matthew Alberts <notifications@github.com> wrote:
> 
> https://security.stackexchange.com/questions/29778/why-does-openssl-writes-ec-parameters-when-generating-private-key
> 
> Maybe the answer is, I delete the EC PARAMETERS from my cert, we close this PR, and make no change.

I thought that EC parameters was used when you had a curve that OpenSSL didn't have built-in support for? The majority of users would be using the built-in named curves.

internal/dag/secret.go
```diff
@@ -50,35 +51,73 @@ func isValidSecret(secret *v1.Secret) (bool, error) {
...
+	var keys int
+	for len(data) > 0 {
+		var block *pem.Block
+		block, data = pem.Decode(data)
+		if block == nil {
+			return errors.New("failed to parse PEM block")
+		}
+		switch block.Type {
+		case "PRIVATE KEY":
+			if err := validatePKCSPrivateKey(block.Bytes); err != nil {
```
davecheney: Can you walk me through the logic here; if the type is private key, isn't it known that the contents will be a PKCS1 key (or vice versa for PKCS8 keys). The original logic here was taken from a rather rubbery example in the the go stdlib and, again without any concrete experience here, if the type is PRIVATE KEY, then we shoulld parse it as a PRIVATE KEY, not throw it against several decoders until one of them green lights it.

mattalberts:
I've been talking to the maintainers of [cfssl](https://github.com/cloudflare/cfssl), they don't handle any parameters.

youngnick:
I've made the classic mistake here and reviewed this from a position of 'will this work?' without considering where it fits in the larger project, sorry @mattalberts for the confusion.

I think that 'EC PARAMETERS' is by far the most likely parameters section to see on a cert, and it's reasonable (from an end user point of view) that we should support certificates with that block type.

I agree with @davecheney that passing things through without validating them should make us uncomfortable, I think in the EC PARAMETERS case, it's probably an acceptable tradeoff because most of the time they won't be used, since people don't generally make their own curves, but use one of a pretty limited set. (Thanks for that link as well @mattalberts, interesting read)

Regardless of all of this, once we all agree on the conditions, they need to be documented outside of the code. That feels like a reference-level document, that would document the requirements for keypairs that Contour will verify and allow, so it would belong in the `docs/` directory.

internal/dag/secret.go
mattalberts: Reading the RFCs and comments .. I think you're right, I may only need to do that to the RSA key.  I've been trying to piece together the correct logic (or commonly accepted logic) for parsing the private keys ([cfssl](https://github.com/cloudflare/cfssl/blob/master/helpers/derhelpers/derhelpers.go) has been my primary source of example).

internal/dag/secret.go
```diff
@@ -50,35 +51,73 @@ func isValidSecret(secret *v1.Secret) (bool, error) {
...
-	if err != nil {
-		return err
+	var exists bool
+	var block *pem.Block
+	for len(data) > 0 {
+		block, data = pem.Decode(data)
+		if block == nil {
+			return errors.New("failed to parse PEM block")
+		}
+		if !strings.HasSuffix(block.Type, "CERTIFICATE") {
```
davecheney: I'm sorry this has dragged on but the discussion in `validatePrivateKey` makes me wonder if this logic is correct. This skips checking blocks that don't end in certificate, thereby passing them to envoy without validation.

I think that two things should happen

1.  Rather than strings.HasSuffix this code should mirror validatePrivateKey and use a switch on the expected block types. 
2. What other kinds of blocks are present in public certificates? ISTM that there shouldn't be any, and the underlying requirement is to support certificates with some intermediate CAs included in the pem file (apologies for the incorrect terminology, I'm a noob).

mattalberts:
@jpeach I've been trying to answer that question all day, by way of looking at openssl and boringssl (the later being more relevant).  When using non-built-in curves, it appears that you generate the key as explicit, to encode the curve information directly into the key.

```
openssl ecparam -name brainpoolP512t1 -genkey -noout -out brainpoolP512t1-key.pem -param_enc explicit
```

https://github.com/google/boringssl/blob/master/crypto/ec_extra/ec_asn1.c#L383-L416
Digging around here, it appears parameters are extracted from the data encoded in the key. 

I've tested removal of the parameters for my certs that were originally rejected, everything still works.

davecheney:
Thanks for your continued hard work. If Contour 0.15.1 rejected a
certificate which is correctly formed Contour 0.15.2 should be fixed
to address that -- you should not have to change your certificate
generation process as those certificates are not invalid.

mattalberts:
I wish I was more versed in the world of expectations on data to expect in key pem blocks. :)

davecheney:
> I wish I was more versed in the world of expectations on data to expect in key pem blocks. :)

Me too. It was eye opening to understand that a tls.key file did not _just_ contain, a key.

mattalberts:
I'd love the extra set of eyes . The only key i've been able to generate with openssl that does leak the params into the pem (where a `-noout` would have squashed it) is for ec private keys.

mattalberts:
Okay, i'll rebase the branch and then let you guys hit merge :)

davecheney:
Thank you

PR #1712, opened 2019-10-16
Add site development Docker image.

jpeach:
Add a Dockerfile and make target that can be used for hacking on
the Contour website. This removes the need to deal with installing
Ruby and Jekyll locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,11 @@ LOCAL_BOOTSTRAP_CONFIG = localenvoyconfig.yaml
 SECURE_LOCAL_BOOTSTRAP_CONFIG = securelocalenvoyconfig.yaml
 PHONY = gencerts
 
+# The version of Jekyll is pinned in site/Gemfile.lock.
+# https://docs.netlify.com/configure-builds/common-configurations/#jekyll
+JEKYLL_IMAGE := jekyll/jekyll:3.8.5
+JEKYLL_PORT := 4000
+
 TAG_LATEST ?= false
 # Used to supply a local Envoy docker container an IP to connect to that is running
 # 'contour serve'. On MacOS this will work, but may not on other OSes. Defining
@@ -202,6 +207,11 @@ certs/envoycert.pem: certs/CAkey.pem certs/envoykey.pem
 		-out certs/envoycert.pem \
 		-days 1825 -sha256 \
 		-extfile _integration/cert-envoy.ext
+
+.PHONY: site-devel
+site-devel: ## Launch the website in a Docker container
+	docker run --publish $(JEKYLL_PORT):$(JEKYLL_PORT) -v $$(pwd)/site:/site -it $(JEKYLL_IMAGE) \
+		bash -c "cd /site && bundle install && bundle exec jekyll serve --host 0.0.0.0 --port $(JEKYLL_PORT) --livereload"
 
 help: ## Display this help
 	@echo Contour high performance Ingress controller for Kubernetes
--- a/site/README-JEKYLL.md
+++ b/site/README-JEKYLL.md
@@ -15,9 +15,8 @@ If you are running a build on Ubuntu you will need the following packages:
 * zlib1g-dev
 * nginx (or apache2)
 
-
 # Local Development
-1. Install Jekyll and plug-ins in one fell swoop. `gem install github-pages`
+1. Install [Jekyll](https://jekyllrb.com) and plug-ins in one fell swoop. `gem install github-pages`
 This mirrors the plug-ins used by GitHub Pages on your local machine including Jekyll, Sass, etc.
 2. Clone down your own fork, or clone the main repo `git clone https://github.com/projectcontour/contour` and add your own remote.
 3. `cd site`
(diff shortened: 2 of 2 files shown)


jpeach:
On Oct 16, 2019, at 11:23 AM, Dave Cheney <notifications@github.com> wrote:

﻿

@davecheney requested changes on this pull request.

I'd prefer not to add this to the list of things we deliver from this project. once its in our docker repo people will depend on it. I'm find with you maintaining this yourself but I don't want to make it an official part of the project.

I agree that it makes sense to not maintain public Docker images for this. This change never pushes the image, but to prevent accidents I could change the repository name to something other than “projectcontour”?


We can review the changes via netlify once they are on a PR if you (understandably) don't want to deal with ruby and bundler.

jonasrosland:
I agree with Steve, it can be useful to have this as a standard way for anyone who wants to contribute to the docs. And same with the Dockerfiles, if there's an already existing one that we can rely on, we should use that.

stevesloka:
I see now why the Dockerfile exists, you don't need to rebuild it each time you run if we can cache that to a local image. =)

jpeach:
On Oct 17, 2019, at 1:19 AM, Steve Sloka <notifications@github.com> wrote:

﻿

I see now why the Dockerfile exists, you don't need to rebuild it each time you run if we can cache that to a local image. =)

It didn’t occur to me that there would be a Jekyll image :) I can switch to it and see whether we can get rid of the Dockerfile.

stevesloka:
The other thing I noticed as well was the docker command used a random port to serve instead of `4000`. Not a big deal, but had to inspect the containers to find the proper port to use.

jpeach:
I updated this to remove the custom image. We now pin the right Jekyll version, which means that startup takes a bit longer but there's no risk of accidentally pushing images to the "projectcontour" repository.

WRT port 4000, that's the internal listen port, Docker will forward that to a random localhost port with the `--publish-all` option. I don't know if a way to tell Docker to used a fixed local port.

stevesloka:
The random port is "safer", but doubtful that someone would be running something over port `4000`. You could update the command to this and it will forward port 4000 from your host to 4000 in the container:
```
docker run -p 4000:4000 -v $$(pwd)/site:/site -it $(JEKYLL_IMAGE) \
		bash -c "cd /site && bundle install && bundle exec jekyll serve --host 0.0.0.0 --livereload"
```

jpeach:
@stevesloka Good point, I added the fixed forwarding to port 4000.

Makefile
```diff
@@ -197,6 +202,11 @@ certs/envoycert.pem: certs/CAkey.pem certs/envoykey.pem
 		-days 1825 -sha256 \
 		-extfile _integration/cert-envoy.ext
 
+.PHONY: site-devel
+site-devel: ## Launch the website in a Docker container
+	docker run --publish $(JEKYLL_PORT):$(JEKYLL_PORT) -v $$(pwd)/site:/site -it $(JEKYLL_IMAGE) \
```
stevesloka: The 2nd $(JEKYLL_PORT) is probably incorrect here unless we change it in the image.

stevesloka: Meaning if I choose port 5000 instead, the container will still listen on 4000

jpeach:
> On 18 Oct 2019, at 12:35 pm, Steve Sloka <notifications@github.com> wrote:
> 
> 
> In Makefile:
> 
> > @@ -197,6 +202,11 @@ certs/envoycert.pem: certs/CAkey.pem certs/envoykey.pem
>  		-days 1825 -sha256 \
>  		-extfile _integration/cert-envoy.ext
>  
> +.PHONY: site-devel
> +site-devel: ## Launch the website in a Docker container
> +	docker run --publish $(JEKYLL_PORT):$(JEKYLL_PORT) -v $$(pwd)/site:/site -it $(JEKYLL_IMAGE) \
> 
> Meaning if I choose port 5000 instead, the container will still listen on 4000

Oh I see what you mean. I updated the Makefile to pass the port to the Jekyll server.

jpeach:
@davecheney WDYT now that this isn't building a custom image?

Makefile
youngnick: Yeah, the excerpt should be `--publish $(JEKYLL_PORT):4000`, that will ensure that port 4000 on the container goes out to whatever you set in $JEKYLL_PORT.

site/README-JEKYLL.md
```diff
@@ -15,9 +15,8 @@ If you are running a build on Ubuntu you will need the following packages:
 * zlib1g-dev
 * nginx (or apache2)
 
-
 # Local Development
-1. Install Jekyll and plug-ins in one fell swoop. `gem install github-pages`
+1. Install [Jekyll](https://jekyllrb.com) and plug-ins in one fell swoop. `gem install github-pages`
```
youngnick: Feels like with this PR we should just change this to say 'use make site-devel, it will handle the magic for you'.

Makefile
jpeach: Well, you set $JEKYLL_PORT and it works. The fact that it sets the internal and host ports to the same value doesn't matter does it? FWIW, this doesn't depend on the value of the `EXPOSE` in the Docker image spec.

PR #1720, opened 2019-10-16
design: Wildcard path routing design details

stevesloka:
Updates #1179 by looking to add additional design considerations for wildcard path matching. 

### DIFF
--- a/design/routing-design.md
+++ b/design/routing-design.md
@@ -120,8 +120,58 @@ Exact Path based routing makes decisions based upon the path of the request (e.g
 
 ### Wildcard Path Routing
 
-Wildcard path based routing allows for a portion of the path to be dynamic (e.g. /path/*/foo).
-This is useful to allow for many types of request paths to be routed to the same backed without specifying them explicitly up front.
+Certain condition types, for example prefix support wildcard style matching. The specifics depend on the condition type.
+
+#### Prefix condition wildcards
+
+- Can have one or more `*` in prefix value
+- Cannot start or end with a `*` (starting with a `*` is prohibited by the general rule that prefix: must start with a `/`)
+- Prefix conditions with wildcards are merged as per the general rule for merging prefix conditions
+- Wildcard path matches cannot be defined inside an `Include` (see next section)
+
+When interpreting a prefix condition containing wildcards the following rules are applied.
+For the condition prefix: `/api/*/users` (either expressed directly, or formed via merging prefix conditions) the request's URL must meet these criteria:
+
+1. Must start with `/api/`
+2. Following `/api/` there must be one or more characters BUT NOT `/users`. These are considered to be the wildcard characters.
+Following the wildcard characters must be the string `/users`.
+3. Specifically, wildcards match one or more characters; the condition prefix: `/api/*/users` does not match the url `/api/users/foo`
+4. More specific routes have higher priority over wildcard routes (see next section)
+
+#### Proper Delegation
+Since wildcard introduces uncertainty within the exact route that will match, some additional consideration needs to be applied when Contour determines if a route has proper delegation permissions.
+
+For example if we had two includes which have similar path matches, we cannot determine who has authority:
+
+```yaml
+includes:
+- name: blogsite
+  namespace: marketing
+  conditions:
+  - prefix: /blog/*/info
+- name: infosite
+  namespace: infotech
+  conditions:
+  - prefix: /blog/tech/info
+``` 
+
+Another example of an issue if routes are defined on includes and conflict.
+The ordering that Contour places these into will result in the more specific route getting traffic (i.e. `/blog/tech/info`) before the wildcard route:
+
+```yaml
+spec:
+ routes:
+ - conditions:
+   - prefix: /blog/*/info
+   services:
+   - name: s1
+     port: 80
+ - conditions:
+   - prefix: /blog/tech/info
+   services:
+   - name: s2
+     port: 80
+``` 
 
 #### Use Cases
 
@@ -334,11 +384,8 @@ Another style of path matching is a wildcard style which allows for a portion of
 A wildcard match allows for a portion of a path to be dynamic.
 
 For example, we could define a wildcard style path: `/app2/*/foo`, but not `/app2/*`.
-Wildcard paths are be able to delegated without issue as long as the `*` is bounded by paths.
-The previous example ending with `*` has too large of a match to allow for delegation.
+If the path ends with `*` has too large of a match to allow for delegation.
 If this is encountered, Contour will set the status to be error for the corresponding HTTPProxy.
-
-*Note: The type of route match will still need to be specified (e.g. prefix, or path). The examples below show a `prefix` match, but a `path` match would work as well.*
 
 ##### Root HTTPProxy:
 
@@ -352,7 +399,7 @@ spec:
     - name: wildcard
       namespace: prefix
       conditions:
-      - prefix: /app/*/foo   
+      - prefix: /
 ```
 
 ##### Delegate HTTPProxy:
@@ -367,6 +414,8 @@ spec:
   - services:
     - name: wildcard-service
       port: 80
+    conditions:
+    - prefix: /app/*/foo
 ```
 
 ##### Requests


design/routing-design.md
```diff
@@ -123,6 +123,34 @@ Exact Path based routing makes decisions based upon the path of the request (e.g
 Wildcard path based routing allows for a portion of the path to be dynamic (e.g. /path/*/foo).
 This is useful to allow for many types of request paths to be routed to the same backed without specifying them explicitly up front.
 
+- Wildcards defined cannot be the end of the path (e.g. `/foo/*/bar` or `/*/foo` not `/foo*`)
+- Similar to path prefix matching, wildcard path routing requires the condition to start with a `/`
+- There is no change to the sorting of paths. The path `/api/*/foo` would still sort before prefix: `/api/v2`
+- Merging conditions together is allowed. If users define a prefix condition of `/foo` and include that to another Proxy which defines a wildcard path of `/*/bar`, the result is a regex expression of `/foo/*/bar`
+- If the wildcard path match is defined in a `prefix` condition, then the resulting match follows a prefix style match (Note: Exact path routing isn't yet enabled)
+- Wildcard path matches cannot be defined inside an `Include` (see next section)
```
stevesloka: I'm not sure if this is too restrictive, but it would be difficult to detect these types of conflicts in the DAG without doing a 2nd pass over it to validate. 

My problem isn't as much detecting, but understanding who is authoritative or not when the conflict arises which is why I thought of going this path first.

design/routing-design.md
```diff
@@ -123,6 +123,34 @@ Exact Path based routing makes decisions based upon the path of the request (e.g
 Wildcard path based routing allows for a portion of the path to be dynamic (e.g. /path/*/foo).
 This is useful to allow for many types of request paths to be routed to the same backed without specifying them explicitly up front.
 
+- Wildcards defined cannot be the end of the path (e.g. `/foo/*/bar` or `/*/foo` not `/foo*`)
+- Similar to path prefix matching, wildcard path routing requires the condition to start with a `/`
+- There is no change to the sorting of paths. The path `/api/*/foo` would still sort before prefix: `/api/v2`
+- Merging conditions together is allowed. If users define a prefix condition of `/foo` and include that to another Proxy which defines a wildcard path of `/*/bar`, the result is a regex expression of `/foo/*/bar`
```
davecheney: suggestion: leave the implementation details out of the business logic for merging rules. For this part of the document lets lock down the merging rules of different conditions.

design/routing-design.md
```diff
@@ -123,6 +123,34 @@ Exact Path based routing makes decisions based upon the path of the request (e.g
 Wildcard path based routing allows for a portion of the path to be dynamic (e.g. /path/*/foo).
 This is useful to allow for many types of request paths to be routed to the same backed without specifying them explicitly up front.
 
+- Wildcards defined cannot be the end of the path (e.g. `/foo/*/bar` or `/*/foo` not `/foo*`)
```
davecheney: It would be really useful for me to have a condition type defined so we can write down these rules and play with them concisely. 

I don't want to call it wildcard, I think that's a bit specific, and would come with weird rules that the path _must_ contain some glob character ie, `wildcard: /api/` ? what does this match?

One idea might be `path: /api/*/foo/` where path is defined as an _exact match_. That is the path must start with `/api/` and end with /`foo` and have _something_ in between those two, that is it would not match `/api/foo/` or `/api//foo/`

stevesloka: My thought was to not have a specific condition type. If we see a `*` in a prefix, then we build out a regex style match as a prefix match. If we implement exact later on, then we build out the regex that implements an exact match. 

I don't feel super strong about this stance, so happy to create more, it just adds to the spec and I don't feel that it buys us much. 

> don't want to call it wildcard, I think that's a bit specific
That's fine, any thoughts then on a name?

> path must contain some glob character ie, wildcard: /api/ ? what does this match?
To me that's an error. If you define a specific name for this type of matching, but don't provide the special character (i.e. `*`) then to me you entered an invalid value.

stevesloka:
> b. the business rules about how this new condition would merge with prefix:

I've defined most of the rules in the spec, what is missing? 

I added another use-case which gets users into danger, meaning, it breaks our direct delegation model. I'm not sure there are ways around this other than having contour detect for these conditions.

davecheney:
Could we take the implementation details out of the picture for the moment. Let’s say that in addition to prefix and path match that envoy supports today, it also has the exact wildcard matched we want. I know it doesn’t, but let’s just assume it does because that will let us ignore the implantation details and focus on how we surface this in HTTPProxy. 

I can see the argument for overloading prefix, that would let someone say the merge of 

prefix: /api, prefix: /v1/*/users

Would be

prefix: /api/v1/*/users

Which looks good, but what about this?

prefix: /api/, prefix: */users

Does that become

prefix: /api/*/users

Probably, but it falls foul of the rule about prefixes starting with a slash, so what users should have to write is

prefix: /api, prefix: /*/users

This is probably ok, and I’m coming around to having a wildcard character in the prefix as saying “I don’t care about this piece here”, the path had to start with /api/, then some bits I don’t care about then /users/ and then optionally some more stuff. 

If we went down the route then prefix: /api, and prefix: /api* mean the same thing. So we might have to add a rule that a prefix condition must being with a slash and must not end with a star. 

How about that?

stevesloka:
> prefix: /api/, prefix: */users

The design calls that out, prefixes must start with a slash `/`: https://github.com/projectcontour/contour/pull/1720/files#diff-
c98772964e726cdffc5dec158a5ad004R127

> If we went down the route then prefix: /api, and prefix: /api* mean the same thing. So we might have to add a rule that a prefix condition must being with a slash and must not end with a star.

I called this out as well, maybe I can make the rules more clear in the design doc: https://github.com/projectcontour/contour/pull/1720/files#diff-c98772964e726cdffc5dec158a5ad004R126

I think these rules are ok to implement, what I'm getting a bit stuck on are these called out: https://github.com/projectcontour/contour/pull/1720/files#diff-c98772964e726cdffc5dec158a5ad004R137

So if I have two routes defined and they each define these paths:

- /blog/*/info
- /blog/tech/info

Does this just become a rule that Contour needs to detect and throw an error? On a cold startup of Contour, who wins this scenario? If this is defined in the same route then that single proxy is invalid and I think that's ok. Sticking to the idea that you cannot use a wildcard in an Include, then maybe this is enough to solve.

jpeach:
I have 2 thoughts about the ordering of the path matching rules:

1. We could define a fixed class-based ordering for the path (i.e. prefix then wildcard then regex)? I think that I can rationalize this ordering in terms of specificity - a prefix is more specific than a wildcard, and regex is the most general of all.

2. Ordering should follow delegation. So conditions that are applied higher up the delegation tree should always order first. It's nor clear to me whether this property would be desirable or simply confusing :)

design/routing-design.md
davecheney: Thanks for bearing with me, you've won me over that we could add `*` to `prefix:`. What I'd like to see is the introduction rewritten to something like this

## Wildcard matching

Certain condition types, for example `prefix` support wildcard style matching. The specifics depend on the condition type.

### Prefix condition wildcards

- Can have one or more `*` in prefix value.
- Cannot start or end with a `*` (starting with a `*` is prohibited by the general rule that `prefix:` must start with a `/`)
- Prefix conditions with wildcards are merged as per the general rule for merging prefix conditions.

When interpreting a prefix condition containing wildcards the following rules are applied. For the condition `prefix: /api/*/users` (either expressed directly, or formed via merging prefix conditions) the request's URL must meet these criteria:
1. It must start  with `/api/`
2. Following `/api/`  there must be one or more characters _BUT NOT_ `/users`. These are considered to be the wildcard characters.
3. Following the wildcard characters must be the string `/users`.

Specifically, wildcards match one or more characters; the condition `prefix: /api/*/users` _does not_ match the url `/api/users/foo`.

stevesloka:
@davecheney I've updated the design with your comments. Some additional things I added that would be good to discuss:

1. I think the ordering should allow for more defined routes to take precedence, meaning if more specific requests should match before wildcards (e.g.   `/blog/tech/info` should match before `/blog/*/info`)
2. I still have the bit about not allowing wildcards in an include. I think it breaks proper delegation.

design/routing-design.md
davecheney: I agree with you that restricting wildcards to just routes is probably a good first step, it'll always be possible to lift that restriction in the future.

Issue #1723, opened 2019-10-17
Make "kubectl explain" work for HTTPProxy resources.


jpeach:
**Please describe the problem you have**

`kubectl explain` is a built-in way to provide Kubernetes resource reference material to users. It would be great to support this as part of the `HTTPProxy` documentation.

jpeach:
@youngnick I think that `kubectl explain` pull this from the OpenAPI definition for the Resource, so maybe we could get this for free as part of #1575?

jpeach:
```$ k explain httpproxy
KIND:     HTTPProxy
VERSION:  projectcontour.io/v1

DESCRIPTION:
     <empty>
```
```$ k explain httpproxy.spec
KIND:     HTTPProxy
VERSION:  projectcontour.io/v1

DESCRIPTION:
     <empty>
```

:(

youngnick:
Hmm, probably. I'll take this for now and see if that happens when I do #1414.

davecheney:
There is also #1567

youngnick:
Sadly, #1731 didn't just magically fix this one, but I think that that's due to upstream things.

Specifically:
- [Pruning](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#preserving-unknown-fields) must be enabled. This is optional for CRD version `v1beta1` (what we're currently using). We can specify it, but not with the generator I added in #1731, it doesn't support it yet. I enabled it manually in a test 1.16 cluster, but the `kubectl explain` was not the full version,  so I suspect the second condition is also not fulfilled currently.
- The schema must be [structural](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#specifying-a-structural-schema), which the ones we are publishing *should* be (thanks to the generator). I suspect that we may need to update to a newer version of the generator, but master is broken in my version. In addition, I suspect something we are doing in our current version is making the validation not structural, but I can't find what it is.

The tl;dr here for next steps is:
- We need to require at least Kubernetes 1.16, and move to CustomResourceDefinition `v1`. That will make solving this a lot easier. It's a big change though, and will require a long tail for updates.
- Until that happens the best we can do is chase `controller-tools` to add support for setting the pruning value in the spec, and to add support for ensuring the validation is structural.

There is no way any of this will be done in rc.2, so I'm moving this to backlog. I'll keep it and poke it along.

jpeach:
[...]

PR #1731, opened 2019-10-18
Add automatic rendering of CRD YAMLs

### LINKED ISSUE #1414: Add openAPIV3Schema validation for HTTPProxy
The initial build of the HTTPProxy CRD left out defining the openAPIV3Schema definitions. This was on purpose while the spec is in development.

youngnick:
CRD YAMLs not being updated will block CI, and can be done using `make render` or `make rendercrds`.

Fixes #1414

Updates #1723


examples/render/contour.yaml
```diff
@@ -211,6 +74,982 @@ data:
     #   - "x_forwarded_for"
     
 ---
+---
```
davecheney: I think the turd on line 76 comes from 02-config.yml

examples/render/contour.yaml
```diff
@@ -276,6 +1115,7 @@ spec:
   completions: 1
   backoffLimit: 1
 ---
+---
```
davecheney: I'm worried about all of these double ---'s

youngnick: I did do a check and they don't cause any errors with the kubectl yaml parsing. I figured it was better to have double `---` instead of not enough.

youngnick: I can change it if you like though.

youngnick: I've changed everything to have `---` only at the start of the file, that should remove the double ups.

apis/contour/v1beta1/ingressroute.go
```diff
@@ -73,22 +84,27 @@ type Service struct {
...
+// HealthCheck defines health checks on the upstream service.
 type HealthCheck struct {
 	// HTTP endpoint used to perform health checks on upstream service
 	Path string `json:"path"`
 	// The value of the host header in the HTTP health check request.
 	// If left empty (default value), the name "contour-envoy-healthcheck"
 	// will be used.
+	// +optional
 	Host string `json:"host,omitempty"`
 	// The interval (seconds) between health checks
```
davecheney: I think this is optional

apis/projectcontour/v1/httpproxy.go
```diff
@@ -21,32 +21,39 @@ import (
 type HTTPProxySpec struct {
 	// Virtualhost appears at most once. If it is present, the object is considered
 	// to be a "root".
+	// +optional
 	VirtualHost *VirtualHost `json:"virtualhost,omitempty"`
 	// Routes are the ingress routes. If TCPProxy is present, Routes is ignored.
 	Routes []Route `json:"routes"`
```
davecheney: Routes are optional

apis/projectcontour/v1/httpproxy.go
```diff
@@ -98,31 +111,41 @@ type TLS struct {
...
 	Passthrough bool `json:"passthrough,omitempty"`
 }
 
 // Route contains the set of routes for a virtual host.
 type Route struct {
 	// Conditions are a set of routing properties that is applied to an HTTPProxy in a namespace.
+	// +optional
 	Conditions []Condition `json:"conditions,omitempty"`
 	// Services are the services to proxy traffic.
+	// +optional
```
davecheney: services are not optional

davecheney:
Please run 
```
go mod tidy
```
and commit the results.

apis/contour/v1beta1/ingressroute.go
youngnick: I marked everything with 'omitempty' as optional, as I assumed it must be if you can omit it from the json.

davecheney: there are probably errors in our json encoding of those types as well.

youngnick:
I had to add some stuff to `tools.go` to make `go mod tidy` happy.

examples/rendercrds.sh
```diff
@@ -0,0 +1,21 @@
+#! /bin/sh
+
+# Latest semver release does not compile, gg controller-tools.
+# Override it with this for now.
+go install sigs.k8s.io/controller-tools/cmd/controller-gen
+
+mkdir -p tmp
```
jpeach: Nit: it would be more robust to use `mktemp`. Something like:
```sh
TMPDIR=`mktemp -d crd.XXXXXX`

trap "rm -rf $TMPDIR; exit" 0 1 2 15
```

apis/contour/v1beta1/ingressroute.go
```diff
@@ -73,50 +84,69 @@ type Service struct {
...
 	Strategy string `json:"strategy,omitempty"`
 	// UpstreamValidation defines how to verify the backend service's certificate
+	// +optional
 	UpstreamValidation *projcontour.UpstreamValidation `json:"validation,omitempty"`
 }
 
-// HTTPHealthCheckPolicy defines health checks on the upstream service.
+// HealthCheck defines health checks on the upstream service.
 type HealthCheck struct {
 	// HTTP endpoint used to perform health checks on upstream service
```
davecheney: path is optional

examples/rendercrds.sh
davecheney: what about $(mktemp -d) so this works cleanly if the script is ^C 1/2 way throught?

jpeach: `$()` is a bash-ism :)

examples/render/contour.yaml
```diff
@@ -209,7 +72,965 @@ data:
...
+                        description: NumRetries is maximum allowed number of retries.
+                          If not supplied, the number of retries is zero.
+                        format: int32
+                        type: integer
+                      perTryTimeout:
+                        description: PerTryTimeout specifies the timeout per retry
+                          attempt. Ignored if NumRetries is not supplied.
+                        type: string
+                    required:
+                    - count
```
davecheney: I think count should be +optional based on its comment and the code which will default to 1 if not supplied.

examples/render/contour.yaml
```diff
@@ -209,7 +72,965 @@ data:
...
+                  timeoutPolicy:
+                    description: The timeout policy for this route
+                    properties:
+                      request:
+                        description: Timeout for receiving a response from the server
+                          after processing a request from client. If not supplied
+                          the timeout duration is undefined.
+                        type: string
+                    required:
+                    - request
```
davecheney: This one is questionable, but I think for the case of IngressRoute which we don't want to change, this field should be optional even if not supplying it makes no sense.

examples/render/contour.yaml
```diff
@@ -209,7 +72,965 @@ data:
...
+                    to another IngressRoute
+                  properties:
+                    name:
+                      description: Name of the IngressRoute
+                      type: string
+                    namespace:
+                      description: Namespace of the IngressRoute. Defaults to the
+                        current namespace if not supplied.
+                      type: string
+                  required:
```
davecheney: both and name and namespace are required for IngressRoute.tcpproxy.delegate.

examples/render/contour.yaml
```diff
@@ -209,7 +72,965 @@ data:
...
+                            description: The time to wait (seconds) for a health check
+                              response
+                            format: int64
+                            type: integer
+                          unhealthyThresholdCount:
+                            description: The number of unhealthy health checks required
+                              before a host is marked unhealthy
+                            format: int32
+                            type: integer
+                        required:
```
davecheney: path should be optional.

youngnick:
It still doesn't work for me either, I'm not sure if all the pieces are in place yet. That's why I only had 'Updates' for that ticket.

examples/render/contour.yaml
```diff
@@ -209,7 +72,965 @@ data:
...
+                          response
+                        format: int64
+                        type: integer
+                      unhealthyThresholdCount:
+                        description: The number of unhealthy health checks required
+                          before a host is marked unhealthy
+                        format: int32
+                        type: integer
+                    required:
+                    - path
```
davecheney: path should be optional

examples/render/contour.yaml
youngnick: Okay, I've updated the comment to that effect as well.

examples/render/contour.yaml
youngnick: I'll fix the comment too then.

examples/render/contour.yaml
youngnick: as discussed, leaving this as required.

examples/render/contour.yaml
youngnick: turns out namespace is not required. Leaving this one how it is.

PR #1741, opened 2019-10-18
Document upstream validation on a service

### LINKED ISSUE #1415: Document the TLS certificate validation ca.crt requirement
TLSCertificateValidation uses k8s Secret objects to store the CA certificate. This convention should be documented explicitly.

stevesloka:
Fixes #1415 by adding docs for upstream validation

### DIFF
--- a/site/_docs_1_0/httpproxy.md
+++ b/site/_docs_1_0/httpproxy.md
@@ -990,6 +990,35 @@ spec:
       weight: 20
 ```
 
+## Upstream Validation
+
+When defining upstream services on a route, it's possible to configure the connection from Envoy to the backend endpoint to communicate over TLS.
+Two configuration items are required, a CA certificate and a `SubjectName` which are both used to verify the backend endpoint's identity.
+
+The CA certificate bundle for the backend service should be supplied in a Kubernetes Secret.
+The referenced Secret must be of type "Opaque" and have a data key named `ca.crt`.
+This data value must be a PEM-encoded certificate bundle.
+
+In addition to the CA certificate and the subject name, the Kubernetes service must also be annotated with a Contour specific annotation: `projectcontour.io/upstream-protocol.tls: <port>` ([see annotations section](annotations.md))
+
+_Note: This annotation is applied to the Service not the Ingress or HTTPProxy object._
+
+```yaml
+apiVersion: projectcontour.io/v1
+kind: HTTPProxy
+metadata:
+  name: blog
+  namespace: marketing
+spec:
+  routes:
+    - services:
+        - name: s2
+          port: 80
+          validation:
+            caSecret: foo-ca-cert
+            subjectName: foo.marketing
+```
+
 ## Status Reporting
 
 There are many misconfigurations that could cause an HTTPProxy or delegation to be invalid.


site/_docs_1_0/httpproxy.md
```diff
@@ -990,6 +990,31 @@ spec:
...
 ```
 
+## Upstream Validation
+
+When defining upstream services on a route, it's possible to configure the connection from Envoy to the backend endpoint to communicate over TLS.
+A CA certificate can be supplied to verify the backend endpoint contains a valid certificate.
+A `SubjectName` can also be supplied to verify the backend endpoint in addition to the CA cert.
+
+- The secret referenced must be type "Opaque" where the key of the secret is named: `ca.crt`
+- The service must be annotated with a Contour specific annotation: `projectcontour.io/upstream-protocol.tls: <port>` ([see annotations section](annotations.md))
```
youngnick: Maybe something to really make it clear that it's the referenced service and not the HTTPProxy? I know we've had queries about this before.

site/_docs_1_0/httpproxy.md
```diff
@@ -990,6 +990,31 @@ spec:
...
+metadata:
+  name: blog
+  namespace: marketing
+spec:
+  routes:
+    - services:
+        - name: s2
+          port: 80
+          validation:
+            caSecret: foo-ca-cert
```
jpeach: Is it possible to specify a secret that is in a different namespace?

site/_docs_1_0/httpproxy.md
```diff
@@ -990,6 +990,31 @@ spec:
       weight: 20
 ```
 
+## Upstream Validation
+
+When defining upstream services on a route, it's possible to configure the connection from Envoy to the backend endpoint to communicate over TLS.
+A CA certificate can be supplied to verify the backend endpoint contains a valid certificate.
+A `SubjectName` can also be supplied to verify the backend endpoint in addition to the CA cert.
+
+- The secret referenced must be type "Opaque" where the key of the secret is named: `ca.crt`
```
jpeach: Nit: I'd remove the `:` here.

site/_docs_1_0/httpproxy.md
```diff
@@ -990,6 +990,31 @@ spec:
       weight: 20
 ```
 
+## Upstream Validation
+
+When defining upstream services on a route, it's possible to configure the connection from Envoy to the backend endpoint to communicate over TLS.
```
jpeach: If both the `caSecret` and `subjectName` keys are required, we should state that explicitly.

site/_docs_1_0/httpproxy.md
```diff
@@ -990,6 +990,31 @@ spec:
...
+A CA certificate can be supplied to verify the backend endpoint contains a valid certificate.
```
jpeach: I think we can improve this by explicitly noting that CA bundles are stores in secrets. How about something like this:

```
A CA certificate bundle for the backend service should be supplied in a Kubernetes Secret.
The referenced Secret must be of type "Opaque" (this type of Secret is generated by the
[kubectl create secret generic](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#-em-secret-generic-em-) command), 
and have a data key named `ca.crt`. This data value must be a PEM-encoded certificate bundle.
```

site/_docs_1_0/httpproxy.md
stevesloka: No, the secret must be in the same namespace as the service. It would be interesting to allow this to work with cert delegation (or it might already, haven't tested that scenario out)

PR #1746, opened 2019-10-21
Add TCP Keepalives for Envoy->Contour xDS comms

### LINKED ISSUE #1744: Add TCP Keepalives to Envoy -> Contour xDS comms
By default, Envoy will not tear down half open TCP connections for xDS, and will believe that it's connected.
See https://github.com/envoyproxy/envoy/issues/6323 and https://github.com/envoyproxy/envoy/issues/5173 for more background.

We should configure tcp keepalives for the Envoy -> Contour connection to ensure that half-open connections are flushed.

A sample config from envoyproxy#5173:
```yaml
    static_resources:
      clusters:
      - name: {{ cluster_name }}_ads
        connect_timeout: { seconds: 1 }
        dns_refresh_rate: { seconds: 10 }
        type: STRICT_DNS
        ...
        upstream_connection_options:
          tcp_keepalive:
            keepalive_probes: 3
            keepalive_time: 30
            keepalive_interval: 5
```

Thanks to @ravilr for noting this one.

It's also possible that #1523 *might* be related to this.

youngnick:
Fixes #1744.

Note that this only adds this to the xDS setup in `contour bootstrap`. No other upstreams will be affected, although it sees like it's worth investigating.

### DIFF
--- a/internal/envoy/bootstrap.go
+++ b/internal/envoy/bootstrap.go
@@ -49,6 +49,13 @@ func Bootstrap(c *BootstrapConfig) *bootstrap.Bootstrap {
 						SocketAddress(c.xdsAddress(), c.xdsGRPCPort()),
 					),
 				},
+				UpstreamConnectionOptions: &api.UpstreamConnectionOptions{
+					TcpKeepalive: &envoy_api_v2_core.TcpKeepalive{
+						KeepaliveProbes:   protobuf.UInt32(3),
+						KeepaliveTime:     protobuf.UInt32(30),
+						KeepaliveInterval: protobuf.UInt32(5),
+					},
+				},
 				Http2ProtocolOptions: new(envoy_api_v2_core.Http2ProtocolOptions), // enables http2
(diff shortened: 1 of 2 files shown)


internal/envoy/bootstrap.go
```diff
@@ -49,6 +49,13 @@ func Bootstrap(c *BootstrapConfig) *bootstrap.Bootstrap {
 						SocketAddress(c.xdsAddress(), c.xdsGRPCPort()),
 					),
 				},
+				UpstreamConnectionOptions: &api.UpstreamConnectionOptions{
+					TcpKeepalive: &envoy_api_v2_core.TcpKeepalive{
+						KeepaliveProbes:   protobuf.UInt32(3),
```
davecheney: Good to see envoy continues its (in)consistent use of durations, timeouts in seconds, and various unsigned integers

davecheney:
@youngnick please assign this to rc.2 so we include it in the release notes

PR #1747, opened 2019-10-21
internal/contour: clean up holdoff timer

davecheney:
This PR addresses incorrect comments that I noticed during #1745.

Something I noticed doing this is the holdoff logic is as follows.

1. When the EventHandler is created e.last is set to time.Now, this
causes the stream of updates from the informers to be heldoff for up to
e.HoldoffDelay + e.HoldoffMaxDelay. This process may happen several
times until all updates from the API server are processed.
2. Once the cluster settles down the _next_ update from the API server
will likely be larger than e.HoldoffMaxDelay from e.last and will be
processed _immediately_.
3. If another update arrives after 2. but before another
e.HoldoffMaxDelay then a pending timer will be started and the update
processed after e.HoldoffDelay.

The long and the short of the current logic is infrequent, O(seconds),
updates will be processed immediately, not after e.HoldoffDelay. This
was surprising to me as this was not the logic that I thought that I
wrote when I last went through this function. That is not to say it is
wrong, however it does open the possibility that a large stream of
updates (k apply ...) arriving after the cluster is stable will cause at
least 2 updates with the second being at least e.HoldoffDelay since the
first.

### DIFF
--- a/internal/contour/handler.go
+++ b/internal/contour/handler.go
@@ -114,52 +114,22 @@ func (e *EventHandler) run(stop <-chan struct{}) error {
 		// yet send to the CacheHandler.
 		outstanding int
 
-		// timer holds the timer that will send on C.
+		// timer holds the timer which will expire after e.HoldoffDelay
 		timer *time.Timer
 
 		// pending is a reference to the current timer's channel.
 		pending <-chan time.Time
 	)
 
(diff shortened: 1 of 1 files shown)


davecheney:
Thank you. This code continues to frustrate me with its difficulty for testing.

PR #1752, opened 2019-10-21
internal/envoy: Escape regex strings for header type contains

### LINKED ISSUE #1751: Bug in `contains` header condition
I first started to test the new HTTPProxy CRD in the release 1.0.0 rc1 because I needed to route based on HTTP Headers.

I first started to test with [`httpproxy-multiple-headers.yaml`](https://github.com/projectcontour/contour/blob/master/site/_docs_1_0/httpproxy.md#multiple-routes) because it was close to what I wanted. It worked perfectly.

I proceed to test with the real data I had by modifying the example to route by UUID in the real header but somehow that didn't work. I was only getting 404 which was pretty strange.

I decided to look at the documentation a little bit more to find that there's not only `contains` but also `exact` and [some more](https://github.com/projectcontour/contour/blob/master/site/_docs_1_0/httpproxy.md#header-conditions). Figured it would be worth a try to use `exact` and surprisingly it worked!

To get routed correctly by UUID when using `contains`

Turns out the `contains` is implemented as a regex which would require some escaping to work with _some_ values:
https://github.com/projectcontour/contour/blob/3f664fa2b177351dc2d9c2858be50f97743c0fd4/internal/envoy/route.go#L293-L295

**Environment:**

- Contour version: 1.0.0 rc1
- Kubernetes version: (use `kubectl version`): 1.14
- Kubernetes installer & version: EKS
- Cloud provider or hardware configu

stevesloka:
Fixes #1751 by escaping characters that are special to a regex expression when type is 'contains' for a header matcher condition. 

### DIFF
--- a/internal/envoy/route.go
+++ b/internal/envoy/route.go
@@ -14,6 +14,7 @@ package envoy
 
 import (
 	"fmt"
+	"regexp"
 	"sort"
 	"time"
 
@@ -291,13 +292,22 @@ func headerMatcher(headers []dag.HeaderCondition) []*envoy_api_v2_route.HeaderMa
 		case "exact":
 			header.HeaderMatchSpecifier = &envoy_api_v2_route.HeaderMatcher_ExactMatch{ExactMatch: h.Value}
 		case "contains":
-			header.HeaderMatchSpecifier = &envoy_api_v2_route.HeaderMatcher_RegexMatch{
-				RegexMatch: fmt.Sprintf(".*%s.*", h.Value),
-			}
+			header.HeaderMatchSpecifier = containsMatch(h.Value)
 		case "present":
 			header.HeaderMatchSpecifier = &envoy_api_v2_route.HeaderMatcher_PresentMatch{PresentMatch: true}
 		}
 		envoyHeaders = append(envoyHeaders, header)
 	}
 	return envoyHeaders
+}
+
+// containsMatch returns a HeaderMatchSpecifier which will match the
+// supplied substring
+func containsMatch(s string) *envoy_api_v2_route.HeaderMatcher_RegexMatch {
+	// convert the substring s into a regular expression that matches s.
+	// note that Envoy expects the expression to match the entire string, not just the substring
+	// formed from s. see [projectcontour/contour/#1751 & envoyproxy/envoy#8283]
+	return &envoy_api_v2_route.HeaderMatcher_RegexMatch{
+		RegexMatch: fmt.Sprintf(".*%s.*", regexp.QuoteMeta(s)),
+	}
 }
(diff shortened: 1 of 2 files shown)


internal/envoy/route.go
```diff
@@ -292,7 +293,7 @@ func headerMatcher(headers []dag.HeaderCondition) []*envoy_api_v2_route.HeaderMa
 			header.HeaderMatchSpecifier = &envoy_api_v2_route.HeaderMatcher_ExactMatch{ExactMatch: h.Value}
 		case "contains":
 			header.HeaderMatchSpecifier = &envoy_api_v2_route.HeaderMatcher_RegexMatch{
-				RegexMatch: fmt.Sprintf(".*%s.*", h.Value),
+				RegexMatch: fmt.Sprintf(".*%s.*", regexp.QuoteMeta(h.Value)),
```
davecheney: Would it be simpler to write this as
```
RegexMatch: regex.QuoteMeta(h.value),
```
as the regex does not have `^` or `$` anchors so the contents of `h.Value` will match any position without the `.*` globs.

stevesloka: But for Envoy to match properly you have to match the entire string. Without the wildcards, at beginning/end I don't think it would match.

davecheney:
Hmm, that doesn’t sound right, foo is a regular expression that matches afoobaz.

stevesloka:
I agree, but not with Envoy. I opened an issue before about this because I couldn't get it working: https://github.com/envoyproxy/envoy/issues/8283#issuecomment-532801157

davecheney:
Wow. That is incredible. Thanks for bearing with me and helping me
understand the oddball way envoy does this.

jpeach:
> On 22 Oct 2019, at 10:38 am, Steve Sloka <notifications@github.com> wrote:
> 
> I agree, but not with Envoy. I opened an issue before about this because I couldn't get it working: envoyproxy/envoy#8283 (comment)

Probably worth leaving a code comment about this?

internal/envoy/route.go
davecheney: Would you do something for me. Could you add a helper
```
// containsMatch returns a HeaderMatchSpecifier which will match the
// supplied substring
func containsMatch(s string) *envoy_api_v2_route.HeaderMatcher_RegexMatch {
    // convert the substring s into a regular expression that matches s.
    // note that Envoy expects the expression to match the entire string, not just the substring
   // formed from s. see [the issue]
   return &envoy_api_v2_route.HeaderMatcher_RegexMatch{
 				RegexMatch: fmt.Sprintf(".*%s.*", regexp.QuoteMeta(h.Value)),
   }
}
```

Issue #1619, opened 2019-10-03
[proposal] move _all_ examples from GitHub to netlify


davecheney:
A perennial problem for users who start with GitHub rather than projectcontour.io is they are using our `:master` branch. This means they can get a broken or unstable expirience, eg. #1615

This sort of go worse when we move the README from recommending `:master` to the `:latest` release, moving the Quickstart yaml further from the experience of git pull && kubectl apply. #1484 

Some possible solutions I can think of:

1. Switch the default branch in git from master to `release-whatever`, this should mean the quickstart and the documentation in GitHub match up. With the write kind of url hacking this could remove the requirement to update `site/_redirects` on each release. The downside of this is it'll probably screw up every single PR pushed to the site because GitHub will base the PR on the release branch, not the master branch.
2. We move the Quickstart documentation to netlify and remove it from the README. The readme makes it clear that this is the development version and to go to projectcontour.io/quickstart for the stable version.
3. Probably in combination with 2 we move all the examples from GitHub to netlify -- move them into the `site/examples` directory. This would mean changing netlify to pull from the latest release branch, not master so the example changes are not updated until we cut a release.

stevesloka:
I'd like to see about moving more to netlify for all the docs, having everything versioned in there should fix the problem. I'd expect the master version to point to latest and all others would be tagged in the proper release.

davecheney:
I'd be down with that. Now I understand how to do redirects doing a `projectcontour.io/docs/latest` shouldn't be too difficult.

jonasrosland:
I think this would be a great move, showing users the examples front and center. Removing confusion = always 👍 

I agree with Steve as well, having versioned docs would be nice to have, it does require some minor rework of the current docs structure though.

Your solution Dave to only build the website based on the latest release should also work nicely, I think it would require a few manual steps for each release (point to the right branch in Netlify and run a manual build after release). That's a quick thing though, and we can most likely trigger something using webhooks though, to hopefully remove those manual steps.

youngnick:
I very much like the idea of versioned docs.

davecheney:
Moving to 1.0 final because of the circular dependency on a latest release version > 0.15.x

PR #1755, opened 2019-10-21
site: enable strict mode

davecheney:
Updates #1619

Enables strict mode and fix strict mode failures.


davecheney:
it's probably this change, https://github.com/projectcontour/contour/pull/1755/files#diff-0a55ba953b4a00424ce1150c8cb5e4efL48

It doesn't fail for me, but maybe Jekyll is falling back to the .git directory in the containing folder. I'll put it back.

davecheney:
@jpeach PTAL

Issue #1449, opened 2019-09-05
internal/dag: merge builder_test and status_test


davecheney:
Currently we test the DAG builder's output via `buider_test.go`, we also test the status output via `status_test.go`. Both of these tests do the same thing -- take some k8s objects, feed them to a `KubernetesCache`, pass that cache to a builder, build a dag, compare the output.

Currently adding a test for status means doing that hop/skip/jump in `status_test.go`, but if you also changed the code in `builder.go`, then you should by right be adding a test in `builder_test.go` as well.

This is double handling and means neither test suite are complete. What we should do is merge the status output checking into `builder_test.go` so that for each builder test we assert the resultant dag, and the status outputs.

PR #1761, opened 2019-10-22
all: permit TCPProxy inclusion

### LINKED ISSUE #1655: Allow TCPProxy to work with Includes in HTTPProxy
#1654 added support for `TCPProxy` at the root httpproxy level, this issue is to describe allowing this to happen from an include outside the root.

davecheney:
Fixes #1655

rc.1 shipped with a spec.tcpproxy.include key but it didn't do anything.
This PR allows a TCPProxy stanza to include the TCPProxy in another
HTTPProxy document. This mirrors the logic that was added to
IngressRoute in 0.9/0.10 (approx.)

- [x] Update generated api and crd validations
- [x] Update httpproxy 1.0 docs
- [x] Add feature test


internal/dag/builder.go
```diff
@@ -917,29 +919,89 @@ func (b *Builder) processTCPProxy(sw *ObjectStatusWriter, ir *ingressroutev1.Ing
...
 		sw, commit := sw.WithObject(dest)
-		b.processTCPProxy(sw, dest, visited, host)
+		b.processIngressRouteTCPProxy(sw, dest, visited, host)
 		commit()
 	}
 }
 
-func (b *Builder) processTCPProxyHTTPProxy(sw *ObjectStatusWriter, httpproxy *projcontour.HTTPProxy, host string) {
-	if len(httpproxy.Spec.TCPProxy.Services) > 0 {
+func (b *Builder) processHTTPProxyTCPProxy(sw *ObjectStatusWriter, httpproxy *projcontour.HTTPProxy, visited []*projcontour.HTTPProxy, host string) bool {
```
stevesloka: Maybe add a comment to the method describing what `bool` means? That's the only confusing part I noticed when reading through. Otherwise looks good. =)

davecheney: done

davecheney:
@youngnick @jpeach @stevesloka this is ready for review.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -156,9 +156,20 @@ type TCPProxy struct {
...
+}
+
+// TCPProxyInclude describes a target HTTPProxy document which contains the TCPProxy details.
+type TCPProxyInclude struct {
+	// Name of the child HTTPProxy
+	Name string `json:"name"`
+	// Namespace of the HTTPProxy to include. Defaults to the current namespace if not supplied.
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
+	// Conditions are a set of routing properties that is applied to an HTTPProxy in a namespace.
```
youngnick: Nit: Pretty sure this isn't needed any more?

site/_docs_1_0/httpproxy.md
```diff
@@ -996,6 +996,45 @@ spec:
       weight: 20
 ```
 
+### TCPProxy delegation
+
+There can be at most on TCPProxy stanza per root HTTPProxy, however that TCPProxy does not need to be defined in the root HTTPProxy object.
```
youngnick: Nit: 'at most one TCPProxy stanza'

jpeach:
I'm reviewing ...

site/_docs_1_0/httpproxy.md
davecheney: fixed

site/_docs_1_0/httpproxy.md
```diff
@@ -996,6 +996,45 @@ spec:
...
+      namespace: app
+---
+# httpproxy-child-termination.yaml
+apiVersion: projectcontour.io/v1
+kind: HTTPProxy
+metadata:
+  name: child
+  namespace: app
+spec:
+` tcpproxy:
```
jpeach: Nit: extraneous backquote.

internal/dag/builder.go
```diff
@@ -917,29 +919,94 @@ func (b *Builder) processTCPProxy(sw *ObjectStatusWriter, ir *ingressroutev1.Ing
...
-				return
+				return false
 			}
 			proxy.Clusters = append(proxy.Clusters, &Cluster{
 				Upstream:           s,
-				LoadBalancerPolicy: loadBalancerPolicy(httpproxy.Spec.TCPProxy.LoadBalancerPolicy),
+				LoadBalancerPolicy: loadBalancerPolicy(tcpproxy.LoadBalancerPolicy),
 			})
 		}
 		b.lookupSecureVirtualHost(host).TCPProxy = &proxy
```
jpeach: Nit: I think it's worth a comment there that the existing `TCPProxy` field must be nil because we constrain the tcpproxy spec to only have either services or an include (but not both).

internal/dag/builder.go
```diff
@@ -917,29 +919,94 @@ func (b *Builder) processTCPProxy(sw *ObjectStatusWriter, ir *ingressroutev1.Ing
...
+		return true
+	}
+
+	visited = append(visited, httpproxy)
+
+	if len(tcpproxy.Services) > 0 && tcpproxy.Include != nil {
+		sw.SetInvalid("tcpproxy: cannot specify services and include in the same httpproxy")
+		return false
+	}
+
```
jpeach: Should we add a check that one of services or include should be set?

internal/dag/builder_test.go
```diff
@@ -2151,6 +2151,86 @@ func TestDAGInsert(t *testing.T) {
...
+			},
+			TCPProxy: &projcontour.TCPProxy{
+				Include: &projcontour.TCPProxyInclude{
+					Name:      "foo",
+					Namespace: s1.Namespace,
+				},
+			},
+		},
+	}
+	proxy39 := &projcontour.HTTPProxy{
```
jpeach: Nit: Can we be consistent about leaving a newline between variables? (personally I prefer the newline because it works well with vim `{` and `}` navigation :) )

internal/dag/builder_test.go
```diff
@@ -2151,6 +2151,86 @@ func TestDAGInsert(t *testing.T) {
 		},
 	}
 
+	proxy37 := &projcontour.HTTPProxy{
```
jpeach: Nit: For proxy38 and proxy39, consider leaving a comment explaining why they are invalid.

internal/dag/builder_test.go
```diff
@@ -4620,6 +4700,54 @@ func TestDAGInsert(t *testing.T) {
...
 		},
+		"insert httpproxy with invalid tcpproxy": {
+			objs: []interface{}{proxy37, s1},
+			want: listeners(),
+		},
+		"insert httpproxy w/ tcpproxy w/ missing include": {
+			objs: []interface{}{proxy38, s1},
+			want: listeners(),
+		},
+		"insert httpproxy w/ tcpproxy w/ includes another root": {
```
jpeach: Nit: Add explanatory comment? I didn't connect all the dots until I read the `status_test.go` changes.

internal/dag/builder_test.go
```diff
@@ -2151,6 +2151,86 @@ func TestDAGInsert(t *testing.T) {
...
+			VirtualHost: &projcontour.VirtualHost{
+				Fqdn: "passthrough.example.com",
+				TLS: &projcontour.TLS{
+					Passthrough: true,
+				},
+			},
+			TCPProxy: &projcontour.TCPProxy{
+				Include: &projcontour.TCPProxyInclude{
+					Name:      "foo",
+					Namespace: s1.Namespace,
```
jpeach: Nit: Comment that this is including `proxy40`?

internal/dag/status_test.go
```diff
@@ -1442,6 +1442,86 @@ func TestDAGIngressRouteStatus(t *testing.T) {
...
+				},
+			},
+			TCPProxy: &projcontour.TCPProxy{
+				Services: []projcontour.Service{{
+					Name: s1.Name,
+					Port: 8080,
+				}},
+			},
+		},
+	}
```
jpeach: Nit: newline (here and above)

internal/dag/builder.go
```diff
@@ -917,29 +919,94 @@ func (b *Builder) processTCPProxy(sw *ObjectStatusWriter, ir *ingressroutev1.Ing
...
+	for _, hp := range visited {
+		path = append(path, fmt.Sprintf("%s/%s", hp.Namespace, hp.Name))
+	}
+	for _, hp := range visited {
+		if dest.Name == hp.Name && dest.Namespace == hp.Namespace {
+			path = append(path, fmt.Sprintf("%s/%s", dest.Namespace, dest.Name))
+			sw.SetInvalid(fmt.Sprintf("tcpproxy include creates a cycle: %s", strings.Join(path, " -> ")))
+			return false
+		}
+	}
```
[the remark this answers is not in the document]
davecheney: Done. #1775

internal/dag/builder.go
```diff
@@ -917,29 +919,94 @@ func (b *Builder) processTCPProxy(sw *ObjectStatusWriter, ir *ingressroutev1.Ing
...
+		return false
+	}
+
+	if dest.Spec.VirtualHost != nil {
+		sw.SetInvalid("root httpproxy cannot delegate to another root httpproxy")
+		return false
+	}
+
+	// dest is no longer an orphan
+	delete(b.orphaned, Meta{name: dest.Name, namespace: dest.Namespace})
```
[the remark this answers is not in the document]
davecheney: Good catch, done.

site/_docs_1_0/httpproxy.md
davecheney: I remember looking at that last night and thinking it was a spec of crud on my screen :)

internal/dag/builder_test.go
davecheney: I want to burn this dumpster fire tests, #1449 . After 1.0 is out, I promise.

internal/dag/builder.go
davecheney: Could you say more please, I don't think I understand.

internal/dag/builder.go
davecheney: Good catch. There is a check that one of services or include are set, but nothing happens if neither are set. This could happen directly
```
spec:
  tcpproxy:
```
or transitively
spec:
  tcpproxy:
     include:
       name: bar
       namespace: baz
---
#baz/bar
spec: 
   tcpproxy:
```
I've raised #1776 to track this.

File: internal/dag/builder.go (hunk 5 above)
Comment (jpeach) [reply internal/dag/builder.go]: Theoretically, `b.lookupSecureVirtualHost(host).TCPProxy` could already have been assigned, in which case this code overwrites it with a different proxy object. I convinced myself that this can't happen in practice because the TCP proxy can only either include or have services (not both). If it was already assigned we must be including something with services after already processing some services, which could only happen if we broke the above rule.

File: internal/dag/builder.go [hunk 14]
```diff
@@ -495,7 +495,9 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
 	}
 
 	if proxy.Spec.TCPProxy != nil && (passthrough || enforceTLS) {
```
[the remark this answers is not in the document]
davecheney: Yeah, this logic is super subtle, it got a lot more complicated when passthrough mode was enabled. I've made a note on my todo list to check if there is at least test coverage for the current behaviour and raise some issues to make this less subtle.

internal/dag/builder.go
davecheney: Thanks for clarifying (I wanted to check before I went off on a tangent). 

We have a few compensating controls that should ensure that line 959 never overwrites that value. Earlier on we check that an FQDN is unique across all IngressRoute and httpproxy records. If that check holds true then each root httpproxy should only be processed once (roots cannot include other roots) so this condition would hold true.

But I agree it is subtle, and under tested. I've raised #1777 to capture this.

PR #1763, opened 2019-10-22
internal/envoy: default the cluster connection buffer to 32K

### LINKED ISSUE #1407: Lower the per connection buffer to 64k at the listener level
Envoy assigns a 1Mb buffer to to each incoming connection. It is not clear if it consumes the whole 1Mb in one go, or that is a upper limit.

Either way, I think this should be lowered to something like 64. 

https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/lds.proto#envoy-api-field-listener-per-connection-buffer-limit-bytes

jpeach:
Introduce a `DefaultCluster()` convenience API that returns a Envoy
`v2.Cluster` struct that has only the Contour internal default
fields initialized. This gives us a single place where clusters get
default settings.

Add the 32K default for `per_connection_buffer_limit_bytes`.

Update all the tests to merge their Cluster spec into the default
Cluster struct. This makes the tests insensitive to changes in the
envoy Cluster defaults.

This fixes #1407.

### DIFF
--- a/internal/contour/cluster_test.go
+++ b/internal/contour/cluster_test.go
@@ -56,7 +56,7 @@ func TestClusterCacheContents(t *testing.T) {
 					LbPolicy:       v2.Cluster_ROUND_ROBIN,
 					CommonLbConfig: envoy.ClusterCommonLBConfig(),
 				}),
-			want: []proto.Message{
+			want: clusterslice(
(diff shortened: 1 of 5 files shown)


jpeach:
Abandoning this approach.

PR #1765, opened 2019-10-22
cmd/contour: Wait for informers to sync

### LINKED ISSUE #1280: Wait to discover all K8s resources before sending xDS responses
As far as I can tell Contour will send whatever resources it has whenever xDS DiscoveryRequests come in. This can cause xDS resources to be torn down in Envoy on a Contour restart. Contour should do a synchronous (read: eager and blocking) lookup of its CRDs, Ingresses, Services, Endpoints, Secrets, etc. on startup before responding to DiscoveryRequests.

This is causing a variety of issues depending on the resource. Examples include:
1. Missing EDS causes 503s
1. Missing RDS causes 404s
1. Missing SDS causes SSL handshake errors

/cc @bryanlatten @lrouquette

--- 

Tasks

- [x] Wait for k8s informer caches to sync
- [ ] Wait for dag to be fully built before starting gRPC server

Blocked

- [x] Leader election #1204

stevesloka:
Updates #1280 by waiting for all informer caches must sync fully before starting Contour's gRPC server.

Big shoutout to @jpeach for help with the design!

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -15,6 +15,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"os"
 	"os/signal"
@@ -22,6 +23,8 @@ import (
 	"strconv"
 	"syscall"
 	"time"
+
+	"k8s.io/client-go/tools/cache"
 
 	contourinformers "github.com/projectcontour/contour/apis/generated/informers/externalversions"
 	"github.com/projectcontour/contour/internal/contour"
@@ -170,29 +173,30 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	}
 
 	// step 4. register our resource event handler with the k8s informers.
-	coreInformers.Core().V1().Services().Informer().AddEventHandler(eh)
+	var informers []cache.SharedIndexInformer
+	informers = registerEventHandler(informers, coreInformers.Core().V1().Services().Informer(), eh)
+	informers = registerEventHandler(informers, contourInformers.Contour().V1beta1().IngressRoutes().Informer(), eh)
+	informers = registerEventHandler(informers, contourInformers.Contour().V1beta1().TLSCertificateDelegations().Informer(), eh)
+	informers = registerEventHandler(informers, contourInformers.Projectcontour().V1().HTTPProxies().Informer(), eh)
+	informers = registerEventHandler(informers, contourInformers.Projectcontour().V1().TLSCertificateDelegations().Informer(), eh)
 
 	// After K8s 1.13 the API server will automatically translate extensions/v1beta1.Ingress objects
 	// to networking/v1beta1.Ingress objects so we should only listen for one type or the other.
 	// The default behavior is to listen for networking/v1beta1.Ingress objects and let the API server
 	// transparently upgrade the extensions version for us.
 	if ctx.UseExtensionsV1beta1Ingress {
-		coreInformers.Extensions().V1beta1().Ingresses().Informer().AddEventHandler(eh)
+		informers = registerEventHandler(informers, coreInformers.Extensions().V1beta1().Ingresses().Informer(), eh)
 	} else {
-		coreInformers.Networking().V1beta1().Ingresses().Informer().AddEventHandler(eh)
+		informers = registerEventHandler(informers, coreInformers.Networking().V1beta1().Ingresses().Informer(), eh)
 	}
-	contourInformers.Contour().V1beta1().IngressRoutes().Informer().AddEventHandler(eh)
-	contourInformers.Contour().V1beta1().TLSCertificateDelegations().Informer().AddEventHandler(eh)
-	contourInformers.Projectcontour().V1().HTTPProxies().Informer().AddEventHandler(eh)
-	contourInformers.Projectcontour().V1().TLSCertificateDelegations().Informer().AddEventHandler(eh)
 
 	// Add informers for each root-ingressroute namespaces
 	for _, inf := range namespacedInformers {
-		inf.Core().V1().Secrets().Informer().AddEventHandler(eh)
+		informers = registerEventHandler(informers, inf.Core().V1().Secrets().Informer(), eh)
 	}
 	// If root-ingressroutes are not defined, then add the informer for all namespaces
 	if len(namespacedInformers) == 0 {
-		coreInformers.Core().V1().Secrets().Informer().AddEventHandler(eh)
+		informers = registerEventHandler(informers, coreInformers.Core().V1().Secrets().Informer(), eh)
 	}
 
 	// step 5. endpoints updates are handled directly by the EndpointsTranslator
@@ -200,7 +204,8 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	et := &contour.EndpointsTranslator{
 		FieldLogger: log.WithField("context", "endpointstranslator"),
 	}
-	coreInformers.Core().V1().Endpoints().Informer().AddEventHandler(et)
+
+	informers = registerEventHandler(informers, coreInformers.Core().V1().Endpoints().Informer(), eh)
 
 	// step 6. setup workgroup runner and register informers.
 	var g workgroup.Group
@@ -306,6 +311,18 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	// step 13. create grpc handler and register with workgroup.
 	g.Add(func(stop <-chan struct{}) error {
 		log := log.WithField("context", "grpc")
+
+		synced := make([]cache.InformerSynced, 0, len(informers))
+		for _, inf := range informers {
+			synced = append(synced, inf.HasSynced)
+		}
+
+		log.Printf("waiting for informer caches to sync")
+		if !cache.WaitForCacheSync(stop, synced...) {
+			return fmt.Errorf("error waiting for cache to sync")
+		}
+		log.Printf("informer caches synced")
+
 		resources := map[string]cgrpc.Resource{
 			eh.CacheHandler.ClusterCache.TypeURL():  &eh.CacheHandler.ClusterCache,
 			eh.CacheHandler.RouteCache.TypeURL():    &eh.CacheHandler.RouteCache,
@@ -354,6 +371,11 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	return g.Run()
 }
 
+func registerEventHandler(informers []cache.SharedIndexInformer, inf cache.SharedIndexInformer, eh cache.ResourceEventHandler) []cache.SharedIndexInformer {
+	inf.AddEventHandler(eh)
+	return append(informers, inf)
+}
+
 type informer interface {
 	WaitForCacheSync(stopCh <-chan struct{}) map[reflect.Type]bool
 	Start(stopCh <-chan struct{})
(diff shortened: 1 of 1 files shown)


cmd/contour/serve.go
```diff
@@ -361,10 +389,8 @@ type informer interface {
 
 func startInformer(inf informer, log logrus.FieldLogger) func(stop <-chan struct{}) error {
 	return func(stop <-chan struct{}) error {
-		log.Println("waiting for cache sync")
-		inf.WaitForCacheSync(stop)
 
-		log.Println("started")
+		log.Println("starting informer")
```
davecheney: suggestion: maybe the next line should say "stopped informer", but that might be redundant as the context is already `context=coreinformers`

cmd/contour/serve.go
```diff
@@ -303,9 +315,25 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
 		log := log.WithField("context", "grpc")
+
+		log.Printf("waiting for informer caches")
+		cache.WaitForCacheSync(stop, synced...)
+		log.Printf("informer caches synced")
+
+		// Wait for cache to process
+		timeout := 15 * time.Second // Default timeout if reached code still progresses
+		if !eh.WaitForCacheProcess(timeout) {
+			log.Printf("!! timed out waiting %v for cache to process !!", timeout)
```
davecheney: what is the goal of this line? is this a warning? should contour wait longer if the timeout is reached?

internal/contour/handler.go
```diff
@@ -267,3 +269,28 @@ func (e *EventHandler) setStatus(statuses map[dag.Meta]dag.Status) {
...
 }
+
+// WaitForCacheProcess waits for the Cache to be processed by checking the
+// value of "outstanding"
+// Return true if processed fully or false if timed out waiting
+func (e *EventHandler) WaitForCacheProcess(timeout time.Duration) bool {
+	ch := make(chan bool, 1)
+
+	go func() {
+		for e.outstanding > 0 {
```
davecheney: I think this is a data race

internal/contour/handler.go
```diff
@@ -267,3 +269,28 @@ func (e *EventHandler) setStatus(statuses map[dag.Meta]dag.Status) {
...
+			time.Sleep(100 * time.Millisecond) // ZzzzZZzzzz
+		}
+		ch <- true
```
davecheney: you don't need to send on this channel, closing it is enough to make it readable by line 291

cmd/contour/serve.go
```diff
@@ -303,9 +315,25 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+	for _, inf := range informers {
+		synced = append(synced, inf.HasSynced)
+	}
+
 	// step 13. create grpc handler and register with workgroup.
 	g.Add(func(stop <-chan struct{}) error {
 		log := log.WithField("context", "grpc")
+
+		log.Printf("waiting for informer caches")
+		cache.WaitForCacheSync(stop, synced...)
```
jpeach: If `WaitForCacheSync` returns false, it means the stop chan was signalled and we should (log and) bail

https://github.com/kubernetes/client-go/blob/master/tools/cache/shared_informer.go#L211

cmd/contour/serve.go
stevesloka: I didn't want the wait for the processing to hold forever and block contour. My thought was to have a reasonable high default (this case 15 secs). If that time is reached then move along and start the gRPC server.

internal/contour/handler.go
stevesloka: Any thoughts on how we can check for when the cache queue is done? I have a cluster with 8k services, so those all load up and cause the dag to rebuild a bunch (1600+ times). 

This all doesn't take long, so maybe this check isn't needed but wanted to make sure that the dag that got served up to any Envoy had all the objects already processed.

cmd/contour/serve.go
davecheney: But then the server starts and the cache isn't synced. Ideally we don't want to use a timeout, we want to issue a single operation then wait for it to complete.

internal/contour/handler.go
davecheney: I'm not sure polling outstanding is the right solution, even if we could solve the data race, as in a cluster with a stream of updates e.outstanding may never fall to zero, or it may fall to zero temporarily then rise.

Maybe its worth stepping back and asking the question about the operation's we want to occur before accepting gRPC requests. To my mind they are

1. Receive an initial list of updates from the api server, this is whatever client-go says it is by the "wait for cached" operation on the informer.
2. Initial at least one DAG rebuild cycle
3. Wait for that DAG cycle to be propagated through to the grpc caches.

For 1. I think this is well in hand with you change.
For 2. we have the "eh.UpdateNow" method
For 3. perhaps we need something similar to the eh.Sequence counter that is only tickled when cache updates happen. (eh.Sequence is tickled every time OnAdd/Update/Delete is called, which may not directly relate to cache updates)

cmd/contour/serve.go
jpeach: I think there are 3 basic cases to handle:
1. The cache has not been built yet
2. The cache is still being rebuilt
3. The cache has been built (at least once)

`WaitForCacheProcess` is addressing (2), though the ordering of operations in the event handler means that it can succeed before the DAG has actually been sent to the cache. IIUC, we want to guarantee that we always fall into case (3). The event handler Sequence also doesn't have the semantics we want for this.

One option to consider is an event handler method that explicitly forces a cache rebuild. You'd have to call `UpdateNow()` and then need to have a way to block until the update is done. If the sequence counter was atomic, you could sleep wait until the sequence changed. It feels a bit hack but it would work :)

cmd/contour/serve.go
stevesloka: 👍

stevesloka:
This PR only now looks to wait for the k8s cache to sync. The other bits about waiting for the cachehandler to be processed will be addressed in a future PR.

cmd/contour/serve.go
```diff
@@ -172,35 +175,45 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	// step 4. register our resource event handler with the k8s informers.
 	coreInformers.Core().V1().Services().Informer().AddEventHandler(eh)
```
davecheney: I think this is a duplicate of line 179

cmd/contour/serve.go
```diff
@@ -172,35 +175,45 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
 
 	// step 5. endpoints updates are handled directly by the EndpointsTranslator
 	// due to their high update rate and their orthogonal nature.
 	et := &contour.EndpointsTranslator{
 		FieldLogger: log.WithField("context", "endpointstranslator"),
 	}
-	coreInformers.Core().V1().Endpoints().Informer().AddEventHandler(et)
+
+	informers = append(informers, coreInformers.Core().V1().Endpoints().Informer())
+	informers[len(informers)-1].AddEventHandler(et)
```
davecheney: could this be clearer if it was written like this?
```
endpointInformer := coreInformers.Core().V1().Endpoints().Informer()
endpointInformer.AddEventhandler(et)
informers = append(informers, endpointInformer
```

cmd/contour/serve.go
```diff
@@ -172,35 +175,45 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
 
 	// Add informers for each root-ingressroute namespaces
 	for _, inf := range namespacedInformers {
-		inf.Core().V1().Secrets().Informer().AddEventHandler(eh)
+		informers = append(informers, inf.Core().V1().Secrets().Informer())
 	}
 	// If root-ingressroutes are not defined, then add the informer for all namespaces
 	if len(namespacedInformers) == 0 {
-		coreInformers.Core().V1().Secrets().Informer().AddEventHandler(eh)
+		informers = append(informers, coreInformers.Core().V1().Secrets().Informer())
```
davecheney: Maybe a small helper that took an informer and a resource handler and returned the informer would help
```
informers = append(informers, registerEventHandler(inf.Core().V1().Secrets().Informer())
...
func registerEventHandler(inf <informer type>, eh cache.ResourceEventHandler) <informer type> {
      inf.AddEventHandler(eh)
      return inf
}
```
Then every time we add an informer to the slice, we can also register an event handler at the same time

internal/contour/handler.go
```diff
@@ -62,6 +62,10 @@ type EventHandler struct {
 	// will be made once CacheHandler.OnUpdate has been called.
 	Sequence chan int
 
+	// outstanding counts the number of events received but not
```
davecheney: Please revert the changes to this file

cmd/contour/serve.go
```diff
@@ -303,9 +316,21 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+	synced := make([]cache.InformerSynced, 0, len(informers))
+	for _, inf := range informers {
+		synced = append(synced, inf.HasSynced)
+	}
+
 	// step 13. create grpc handler and register with workgroup.
 	g.Add(func(stop <-chan struct{}) error {
 		log := log.WithField("context", "grpc")
+
+		log.Printf("waiting for informer caches")
```
davecheney: waiting for informer caches to sync

cmd/contour/serve.go
```diff
@@ -303,9 +316,21 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	eh.Metrics = metrics
 	eh.CacheHandler.Metrics = metrics
 
+	synced := make([]cache.InformerSynced, 0, len(informers))
+	for _, inf := range informers {
```
davecheney: Possibly move these four lines closer to where they are used inside the step 13 function below.

stevesloka:
Comments addressed @davecheney, thanks!

cmd/contour/serve.go
```diff
@@ -170,37 +173,40 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	}
 
 	// step 4. register our resource event handler with the k8s informers.
-	coreInformers.Core().V1().Services().Informer().AddEventHandler(eh)
+	informers := []cache.SharedIndexInformer{
```
davecheney: should these be passed through `registerEventHandler`?

stevesloka: I have `informers` getting passed through now, is this in line with what you were suggesting?

cmd/contour/serve.go
```diff
@@ -306,6 +311,18 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
 		log := log.WithField("context", "grpc")
+
+		synced := make([]cache.InformerSynced, 0, len(informers))
+		for _, inf := range informers {
+			synced = append(synced, inf.HasSynced)
+		}
+
+		log.Printf("waiting for informer caches to sync")
+		if !cache.WaitForCacheSync(stop, synced...) {
+			return fmt.Errorf("error waiting for cache to sync")
```
jpeach: Nit: IIUC this isn't really an error, it's that the stop chan was signalled.

cmd/contour/serve.go
davecheney: I wasn't able to find the place where those 6 informers added to `informers` had an event handler attached to them, but I think you resolved this in d8fb549

PR #1781, opened 2019-10-23
Implement wildcard path prefix for HTTPProxy conditions

### LINKED ISSUE #1179: Wildcard path matching like nginx

### LINKED ISSUE #1457: internal/dag: Enable additional fields in Condition

stevesloka:
Fixes #1457 
Fixes #1179 

Adds wildcard path prefix support for HTTPProxy

Signed-off-by: Steve Sloka <slokas@vmare.com>

### DIFF
--- a/internal/dag/conditions.go
+++ b/internal/dag/conditions.go
@@ -34,6 +34,13 @@ func mergePathConditions(conds []projcontour.Condition) Condition {
 	re := regexp.MustCompile(`//+`)
 	prefix = re.ReplaceAllString(prefix, `/`)
 
+	// Check for wildcard path
+	if strings.Contains(prefix, "*") {
+		return &WildcardPrefixCondition{
+			Prefix: prefix,
+		}
+	}
+
 	// After the merge operation is done, if the string is still empty, then
 	// we need to set the prefix to /.
 	// Remember that this step is done AFTER all the includes have happened.
@@ -58,6 +65,32 @@ func pathConditionsValid(sw *ObjectStatusWriter, conds []projcontour.Condition, 
 			if cond.Prefix[0] != '/' {
 				sw.SetInvalid(fmt.Sprintf("%s: Prefix conditions must start with /, %s was supplied", conditionsContext, cond.Prefix))
 				return false
+			}
+			switch conditionsContext {
+			case "include":
+				if strings.Contains(cond.Prefix, "*") {
+					sw.SetInvalid(fmt.Sprintf("Cannot specify wildcard prefix conditions in an Include"))
+					return false
+				}
+			case "route":
+				if strings.HasSuffix(cond.Prefix, "*") {
+					sw.SetInvalid(fmt.Sprintf("Cannot specify trailing wilcard character, '%s' was supplied", cond.Prefix))
+					return false
+				}
+				if strings.HasPrefix(cond.Prefix, "/*") {
+					sw.SetInvalid(fmt.Sprintf("Cannot specify '/*' as leading wilcard character pattern, '%s' was supplied", cond.Prefix))
+					return false
+				}
+				if strings.Contains(cond.Prefix, "**") {
+					sw.SetInvalid(fmt.Sprintf("Cannot specify '**' character pattern in a wildcard prefix, '%s' was supplied", cond.Prefix))
+					return false
+				}
+				for _, s := range strings.Split(cond.Prefix, "/") {
+					if strings.Count(s, "*") > 1 {
+						sw.SetInvalid(fmt.Sprintf("Cannot specify multiple '*' between slashes in a wildcard prefix, '%s' was supplied", cond.Prefix))
+						return false
+					}
+				}
 			}
 		}
 		if prefixCount > 1 {
--- a/internal/dag/conditions.go
+++ b/internal/dag/conditions.go
@@ -74,11 +74,11 @@ func pathConditionsValid(sw *ObjectStatusWriter, conds []projcontour.Condition, 
 				}
 			case "route":
 				if strings.HasSuffix(cond.Prefix, "*") {
-					sw.SetInvalid(fmt.Sprintf("Cannot specify trailing wilcard character, '%s' was supplied", cond.Prefix))
+					sw.SetInvalid(fmt.Sprintf("Cannot specify trailing wildcard character, '%s' was supplied", cond.Prefix))
 					return false
 				}
 				if strings.HasPrefix(cond.Prefix, "/*") {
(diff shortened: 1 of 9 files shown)


stevesloka:
This is pretty much ready, just need to work on the filtering of routes.

internal/dag/conditions.go
```diff
@@ -59,6 +66,32 @@ func pathConditionsValid(sw *ObjectStatusWriter, conds []projcontour.Condition,
 				sw.SetInvalid(fmt.Sprintf("%s: Prefix conditions must start with /, %s was supplied", conditionsContext, cond.Prefix))
 				return false
 			}
+			switch conditionsContext {
```
jpeach: I think we should make `conditionsContext` a type. In a separate issue is OK.

internal/dag/conditions.go
```diff
@@ -59,6 +66,32 @@ func pathConditionsValid(sw *ObjectStatusWriter, conds []projcontour.Condition,
...
+				if strings.Contains(cond.Prefix, "**") {
+					sw.SetInvalid(fmt.Sprintf("Cannot specify '**' character pattern in a wildcard prefix, '%s' was supplied", cond.Prefix))
+					return false
+				}
+				for _, s := range strings.Split(cond.Prefix, "/") {
+					if strings.Count(s, "*") > 1 {
+						sw.SetInvalid(fmt.Sprintf("Cannot specify multiple '*' between slashes in a wildcard prefix, '%s' was supplied", cond.Prefix))
+						return false
+					}
+				}
```
jpeach: This might be cleaner if you split first, then apply the rules to the resulting slice.

internal/envoy/route.go
```diff
@@ -41,31 +42,47 @@ func Route(match *envoy_api_v2_route.RouteMatch, action *envoy_api_v2_route.Rout
...
+}
+
+func sanitizeRegex(prefix string) string {
+	// Since the QuoteMeta will replace the user entered "*" with a "\*",
+	// it makes that character want to match when we want to actually replace
+	// with a custom regex string
+	prefix = strings.ReplaceAll(regexp.QuoteMeta(prefix), "\\*", "*")
+
+	// Replace all "*" with the match-all regex
+	return fmt.Sprintf("%s.*", strings.ReplaceAll(prefix, "*", "[0-9a-zA-Z-]*"))
```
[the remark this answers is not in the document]
jpeach: Also, AFAIK `[0-9a-zA-Z-]` doesn't match URL path characters (e.g. % encodings). Maybe we need `[^/`]?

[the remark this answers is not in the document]
stevesloka: >A wildcard should match at least one right? i.e. [0-9a-zA-Z-]+

Updated to match, good point. 

> Why do you always add the .* to the end of the regex? Is this because envoy always wants the whole path to match? 

Because it's a `prefix` type, need to match anything after the string. Once we implement exact match, this regex would change.

> Also, AFAIK [0-9a-zA-Z-] doesn't match URL path characters (e.g. % encodings). Maybe we need [^/]?

You can have `%` in a path? I know you can in a query string which isn't part of this check.

internal/dag/conditions.go
stevesloka: Can you elaborate a bit more? Not sure I'm following you.

stevesloka:
@davecheney @jpeach I'm thinking that we need to reduce the scope of wildcards (at least for now). I'm proposing we do this:

1. Wildcard matches must only be blobs, meaning you cannot have partial matches (e.g. must match `/foo/*/bar` but not `/foo/d*/bar`)
2. Ordering will be by the longest match prioritizing non-wildcard matches
2b. Wildcard matches will sort longest first in addition to #2 previously

internal/dag/conditions.go
jpeach: If you start the validation by splitting the path into a slice of components, then you can apply all the checks component-wise after that. I think the whole thing will simplify.

internal/envoy/route.go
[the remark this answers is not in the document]
jpeach: Yep, eg. http://example.com/~jpeach, or http://example.com/james%20peach/. Maybe envoy has already %-decoded before the regex, but in that case, you'd still need it to match a space character.

jpeach:
Note that AFAICT, wildcard prefix matching will be incompatible with prefix replacement :(

Issue #1375, opened 2019-08-22
config: sane defaults for envoy


glerchundi:
I came across [this PR](https://github.com/envoyproxy/envoy/pull/8001) from @PiotrSikora and thought it would be interesting to corroborate if those 'best practices' are being followed in Contour.

davecheney:
Thank you for this. There is some very useful information here.

Scheduling for rc1 to review and see what if any changes we should take from this document.

ravilr:
Also, this https://github.com/envoyproxy/envoy/issues/6323 . Can we configure tcp keepalive options to tear down half open connections for both Contour grpc management server upstream cluster and also (optionally may be) for all K8s service upstream clusters in envoy config programmed by Contour.

youngnick:
Building a todo list for this one:

For the TCP keepalive settings, that is 100% worth doing, I've made separate issue (#1744)  for that one, as it might possibly be related to another issue.

TCP Proxies

- [ ] overload manager
- [ ] listener buffer limits to 32KiB
- [ ] cluster buffer limits to 32KiB

HTTP Proxies should additionally configure

- [ ] connection maanager use_remote_address to true (see header_sanitizing for details)
- [ ] connect_timeout 15s
- [ ] per_connection_buffer_limit_byes to 32 KiB
- [ ] HTTP/2 maximum concurrent streams limit to 100
- [ ] HTTP/2 initial stream window size limit to 64KiB
- [ ] HTTP/2 initial connection window size limit to 1MiB

YAML example:
```yaml
overload_manager:
  refresh_interval: 0.25s
  resource_monitors:
  - name: "envoy.resource_monitors.fixed_heap"
    config:
      # TODO: Tune for your system.
      max_heap_size_bytes: 2147483648 # 2 GiB
  actions:
  - name: "envoy.overload_actions.shrink_heap"
    triggers:
    - name: "envoy.resource_monitors.fixed_heap"
      threshold:
        value: 0.95
  - name: "envoy.overload_actions.stop_accepting_requests"
    triggers:
    - name: "envoy.resource_monitors.fixed_heap"
      threshold:
        value: 0.98

static_resources:
  listeners:
  - address:
      socket_address: { address: 127.0.0.1, port_value: 443 }
    listener_filters:
    - name: "envoy.listener.tls_inspector"
      typed_config: {}
    per_connection_buffer_limit_bytes: 32768 # 32 KiB
    filter_chains:
    - filter_chain_match:
[...]

PR #1782, opened 2019-10-23
tests: consolidate envoy Cluster defaults

jpeach:
Many tests need to specify default fields for envoy `Cluster` structs
so that struct equality tests can succeed. Specifying these defaults
in many places is fragile and tends to make it harder to see the
field values that actually matter to the tests.  Consolidating the
defaults into a handful of helpers reduces the overally clutter.
We still need a couple of duplicates of the helper so that we can
avoid package dependency cycles.

This updates #1375.

### DIFF
--- a/internal/envoy/cluster.go
+++ b/internal/envoy/cluster.go
@@ -33,25 +33,68 @@ import (
 // CACertificateKey stores the key for the TLS validation secret cert
 const CACertificateKey = "ca.crt"
 
+func clusterDefaults() *v2.Cluster {
+	return &v2.Cluster{
+		ConnectTimeout: protobuf.Duration(250 * time.Millisecond),
+		CommonLbConfig: ClusterCommonLBConfig(),
+		LbPolicy:       lbPolicy(""),
+	}
+}
+
 // Cluster creates new v2.Cluster from dag.Cluster.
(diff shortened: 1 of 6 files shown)


davecheney:
reviewing now ...

internal/envoy/cluster.go
```diff
@@ -33,25 +33,68 @@ import (
...
+		CommonLbConfig: ClusterCommonLBConfig(),
+		LbPolicy:       lbPolicy(""),
+	}
+}
+
 // Cluster creates new v2.Cluster from dag.Cluster.
 func Cluster(c *dag.Cluster) *v2.Cluster {
-	cl := cluster(c)
+	service := c.Upstream
+	cluster := clusterDefaults()
```
davecheney: I love the clean up that lets cluster be a local variable. 👍

internal/featuretests/envoy.go
```diff
@@ -43,18 +58,15 @@ func routeCluster(cluster string) *envoy_api_v2_route.Route_Route {
 }
 
 func cluster(name, servicename, statName string) *v2.Cluster {
-	return &v2.Cluster{
+	return DefaultCluster(&v2.Cluster{
```
davecheney: This is a really nice cleanup.

PR #1784, opened 2019-10-24
update namespace name

SDBrett:
Namespace reference in guide did not match the namespace used in the example manifest https://raw.githubusercontent.com/projectcontour/contour/release-0.15/examples/render/deployment-rbac.yaml


davecheney:
@SDBrett the docs on the website are correct, we're in the process of moving to that new namespace for the 1.0 release. I'm sorry everything is in a transitional state at the moment, it's going to be done by the first of nov. Please pardon our dust.

PR #1792, opened 2019-10-24
docs: style deprecation notices in red

jpeach:
Style deprecation notices in a red alert box so that it jumps
out and muggs the reader.


jpeach:
I'm ambivalent about the specific styling of these. Happy to drop this PR or take fixes from anyone who actually knows CSS :)

PR #1797, opened 2019-10-25
Set cluster per-connection-buffer-limit to 32KiB

### LINKED ISSUE #1408: Lowering Envoy's cluster per-connection buffer to 32kb
Envoy allocates 1mb per connection to a backend pod.

https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/cds.proto#cluster

It is not clear if it allocates it all in one go, or it is an upper limit. In either case 1Mb is a lot. I think it should be lowered to something like 64kb.

youngnick:
Fixes #1408

Shoutout to @jpeach for the cluster default helper, very nice.

### DIFF
--- a/internal/contour/cluster_test.go
+++ b/internal/contour/cluster_test.go
@@ -850,9 +850,10 @@ func serviceWithAnnotations(ns, name string, annotations map[string]string, port
 func cluster(c *v2.Cluster) *v2.Cluster {
 	// NOTE: Keep this in sync with envoy.defaultCluster().
 	defaults := &v2.Cluster{
-		ConnectTimeout: protobuf.Duration(250 * time.Millisecond),
-		CommonLbConfig: envoy.ClusterCommonLBConfig(),
-		LbPolicy:       v2.Cluster_ROUND_ROBIN,
+		ConnectTimeout:                protobuf.Duration(250 * time.Millisecond),
+		CommonLbConfig:                envoy.ClusterCommonLBConfig(),
+		LbPolicy:                      v2.Cluster_ROUND_ROBIN,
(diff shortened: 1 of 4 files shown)


davecheney:
I’m just having a quick look in the envoy sources to understand what their definition of “soft limit” entails.

youngnick:
Agreed. I'll close this PR and we can come back to it. Thanks for checking.

davecheney:
Sorry for the trouble. Envoy doesn't make it easy to figure out what is
going on.

gopakumarce:
@youngnick @davecheney I came here after a series of google searches on envoy flow control etc.. From the comments here, I understand that the cluster buffer limits are "soft" limits, maybe because it has to deal with the case of multiplexed http2 streams and listener sending pause frames back to the originator etc.. But what if its a simple http1.1 connection - will a backed up cluster cause the listener to stop reading right away ? Any clues will be appreciated

PS: I came here while trying to solve my own problems that ive asked below

https://github.com/envoyproxy/envoy/issues/17281
https://github.com/envoyproxy/envoy/issues/12542

gopakumarce:
https://github.com/envoyproxy/envoy/blob/main/source/docs/flow_control.md - this does say that for http1.1 a simple tcp-stop-reading-listener is done

gopakumarce:
never mind, I figured my issue which I was chasing. But nevertheless yes I also figured all these buffers are just soft limits like you mention in this thread. With 64K buffers defined, I can see that envoy ends up buffering around 300K worth of bytes - thats a lot, but I guess theres no option with http2 other than sending pause frames and relying on the other end receiving pause frames and stop sending data, till then the receiver of the data has to buffer up even if it cant send upstream!

Issue #1750, opened 2019-10-21
Move some resources content into guides.


jpeach:
There are a number of pages located in the "resources" section that could arguably be better kept in the "guides" section. My thought is that resources is mainly for conference videos and the like, whereas "guides" is for showing users how to achieve some useful aim. With that in mind, I think we could relocate

- [x] "IngressRoute to HTTPProxy"
- [x] "Prometheus"
- [ ] "Troubleshooting"
- [ ] "Upgrading Contour"

davecheney:
@jpeach what's left to do on this ticket?

jpeach:
I updated the description to a checklist, and I'll keep it updated as I knock off the guides.

davecheney:
@jpeach thank you!

davecheney:
@jpeach moving this to the 1.0.0 milestone so I can close rc.2 today. Please keep doing what you're doing, there's not need to get this into the rc.2 tag as the website is released from the `master` branch,

jpeach:
Removed the milestone since this is not needed for the 1.0 release.

jpeach:
I don't think we need to do any more here.

PR #1808, opened 2019-10-25
docs: update metrics documentation

jpeach:
Move the metrics documentation from the "resources" section to the
"guides" section. Add build machinery to generate metrics reference
documentation from the Contour sources.

This updates #1750.

### DIFF
--- a/hack/generate-metrics-doc.go
+++ b/hack/generate-metrics-doc.go
@@ -1,0 +1,81 @@
+// Copyright © 2019 VMware
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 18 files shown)


hack/generate-metrics-doc.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	return strings.Join(l, ", ")
+}
+
+// Generate a string name for the metric type, linking to the
+// Prometheus docs if we know there is a suitable target.
+func typeof(mf *dto.MetricFamily) string {
+	switch t := mf.GetType(); t {
+	case dto.MetricType_COUNTER:
```
davecheney: wil it work if you stack the case statements here?
```
case dto.MetricType_COUNTER, 
         dto.MetricType_SUMMARY,
         dto.MetricType_GAUGE,
         dto.MetricType_HISTOGRAM:
```

jpeach: Fixed.

jpeach:
@davecheney I added a `check-stale` makefiles target that runs as part of `check. This fails the tests if there are any uncommitted changes in directories containing generated files.

PR #1811, opened 2019-10-27
internal/dag: accept non-PEM data in PEM files

### LINKED ISSUE #1809: Dual newline character after TLS cert is considered invalid
Upgrading Contour from `0.15.0` to `0.15.3` and see the following errors in the log:
```
ERRO[0001] failed to parse PEM block  context=KubernetesCache kind= name=<redacted> namespace=<redacted> version=
```

The `ingress_https` listener is never configured in Envoy.

I expected the secret to be valid (as it was in `0.15.0`) and the https listener to be configured.

The issue stems from the fact that some of our secrets have a double newline characters as such (`$` representing the newline that would not normally be there):
```
-----BEGIN CERTIFICATE-----
<some cert>
-----END CERTIFICATE-----$
$
```

At some point in the [loop](https://github.com/projectcontour/contour/blob/v0.15.3/internal/dag/secret.go#L56), I see  that `data` is  `"Cg=="`, which is the base64 encoded value of newline. At this point, `len(data) == 1` and `block == nil`, failing the validation.

Apparently, Envoy ignores these characters (and so does openssl btw). My current fix is to change the [test](https://github.com/projectcontour/contour/blob/v0.15.3/internal/dag/secret.go#L54) to `len(data) > 1` (not sure how good/bad that is...).

Below is a full go source that reproduces the issue. The cert is the one from your [tests](https://github.com/projectcontour/contour/blob/v0.15.3/internal/dag/secret_test.go#L20-L52), with the newlines appended:

```go
package main

import (
	"crypto/x509"
	"encoding/pem"
	"errors"
	"fmt"
)

const (
	CERTIFICATE = `-----BEGIN CERTIFICATE-----
MIIFtTCCA52gAwIBAgIJAO0cq2lJPZZJMA0GCSqGSIb3DQEBBQUAMEUxCzAJBgNV
BAYTAkFVMRMwEQYDVQQIEwpTb21lLVN0YXRlMSEwHwYDVQQKExhJbnRlcm5ldCBX
aWRnaXRzIFB0eSBMdGQwHhcNMTQwMzEyMTc0NzU5WhcNMTkwMzEyMTc0NzU5WjBF
MQswCQYDVQQGEwJBVTETMBEGA1UECBMKU29tZS1TdGF0ZTEhMB8GA1UEChMYSW50
ZXJuZXQgV2lkZ2l0cyBQdHkgTHRkMIICIjANBgkqhkiG9w0BAQEFAAOCAg8AMIIC
CgKCAgEAsgzs6vN2sveHVraXV0zdoVyhWUHWNQ0xnhHTPhjt5ggHmSvrUxvUpXfK
WCP9gZo59Q7dx0ydjqBsdooXComVP4kGDjulvOHWgvcVmwTsL0bAMqmsCyyJKM6J
Wqi8E+CPTOpMBWdapUxvwaSmop8geiTtnX0aV4zGXwsz2mwdogbounQjMB/Ew7vv
8XtqwXSpnR7kM5HPfM7wb9F8MjlRuna6Nt2V7i0oUr+EEt6fIYEVZFiHTSUzDLaz
2eClJeCNdvyqaeGCCqs+LunMq3kZjO9ahtS2+1qZxfBzac/0KXRYnLa0kGQHZbw0
ecgdZC9YpqqMeTeSnJPPX4/TQt54qVLQXM3+h8xvwt3lItcJPZR0v+0yQe5QEwPL
4c5UF81jfGrYfEzmGth6KRImRMdFLF9+F7ozAgGqCLQt3eV2YMXIBYfZS9L/lO/Q
3m4MGARZXUE3jlkcfFlcbnA0uwMBSjdNUsw4zHjVwk6aG5CwYFYVHG9n5v4qCxKV
ENRinzgGRnwkNyADecvbcQ30/UOuhU5YBnfFSYrrhq/fyCbpneuxk2EouL3pk/GA
7mGzqhjPYzaaNGVZ8n+Yys0kxuP9XDOUEDkjXpa/SzeZEk9FXMlLc7Wydj/7ES4r
6SYCs4KMr+p7CjFg/a7IdepLQ3txrZecrBxoG5mBDYgCJCfLBu0CAwEAAaOBpzCB
pDAdBgNVHQ4EFgQUWQI/JOoU+RrUPUED63dMfd2JMFkwdQYDVR0jBG4wbIAUWQI/
JOoU+RrUPUED63dMfd2JMFmhSaRHMEUxCzAJBgNVBAYTAkFVMRMwEQYDVQQIEwpT
b21lLVN0YXRlMSEwHwYDVQQKExhJbnRlcm5ldCBXaWRnaXRzIFB0eSBMdGSCCQDt
HKtpST2WSTAMBgNVHRMEBTADAQH/MA0GCSqGSIb3DQEBBQUAA4ICAQBwGbAmiLHE
jubdLeMygwrV8VjlOVxV41wvRi6y1B5Bdvh71HPoOZdvuiZOogzxB22Tzon6Uv5q
8uuAy37rHLlQTOqLwdLJOu/ijMirAkh13gQWt4oZGUDikyiI4PMNo/hr6XoZWUfU
fwmcAzoEMln8HyISluTau1mtQIDgvGprU472GqC4AC7bYeED+ChCevc7Ytjl4zte
/tw8u3nqrkESYBIA2yEgyFAr1pRwJPM/T1U6Ehalp1ZTeQcAXEa7IC6ht2NlN1FC
fk2KQmrk4Z3jaSVv8GxshA354W+UEpti0o6Fv+2ozkAaQ1/xjiNwBTHtgJ1/AG1j
bDYcCFfmYmND0RFjvVu7ma+UNdKQ+t1o7ip4tHQUTEFvdqoaCLN09PcTVgvm71Lr
s8IOldiMgiCjQK3e0jwXx78tXs/msMzVI+9AR9aNzo0Y42C97ctlGu3+v07Zp+x4
6w1rg3eklJM02davNWK2EUSetn9EWsIJXU34Bj7mnI/2DFo292GVNw1kT5Bf4IvA
T74gsJLB6wacN4Ue6zPtIvrK93DABAfRUmrAWmH8+7MJolSC/rabJF3E2CeBTYqZ
R5M5azDV1CIhIeOTiPA/mq5fL1UrgVbB+IATIsUAQfuWivDyoeu96LB/QswyHAWG
8k2fPbA2QVWJpcnryesCy3qtzwbHSYbshQ==
-----END CERTIFICATE-----

`
)

func main() {
	err := validateCertificate([]byte(CERTIFICATE))
	if err != nil {
		fmt.Println("error:", err)
	}
}

func validateCertificate(data []byte) error {
	var exists bool
	for len(data) > 0 {
		var block *pem.Block
		block, data = pem.Decode(data)
		if block == nil {
			return errors.New("failed to parse PEM block")
		}
		if block.Type != "CERTIFICATE" {
			return errors.New("unexpected block type in certificate: " + block.Type)
		}
		if _, err := x509.ParseCertificate(block.Bytes); err != nil {
			return err
		}
		exists = true
	}
	if !exists {
		return errors.New("failed to locate certificate")
	}
	return nil
}
```

**Environment:**

- Contour version: v0.15.3
- Kubernetes version: (use `kubectl version`): n/a
- Kubernetes installer & version: n/a
- Cloud provider or hardware configuration: n/a
- OS (e.g. from `/etc/os-release`): n/a

jpeach:
It is valid for files to intermingle PEM and non-PEM data. The Go
`pem.Decode` API supports this, but doesn't emit any errors that
would let us distinguish between broken PEM and missing PEM. So,
verify that there is a remaining PEM header before we try to decode
it. That way, we know that any decoding failure is a actual error.

This fixes #1809.

### DIFF
--- a/internal/dag/secret_test.go
+++ b/internal/dag/secret_test.go
@@ -13,7 +13,11 @@
 
 package dag
 
-import v1 "k8s.io/api/core/v1"
+import (
+	"fmt"
+
+	v1 "k8s.io/api/core/v1"
+)
 
 const (
 	// sample data from https://8gwifi.org/PemParserFunctions.jsp
@@ -50,6 +54,24 @@ T74gsJLB6wacN4Ue6zPtIvrK93DABAfRUmrAWmH8+7MJolSC/rabJF3E2CeBTYqZ
 R5M5azDV1CIhIeOTiPA/mq5fL1UrgVbB+IATIsUAQfuWivDyoeu96LB/QswyHAWG
 8k2fPbA2QVWJpcnryesCy3qtzwbHSYbshQ==
 -----END CERTIFICATE-----`
+
+	// Sample certificate from https://tools.ietf.org/html/rfc7468#section-5.2
+	CERTIFICATE_WITH_TEXT = `
+Subject: CN=Atlantis
+Issuer: CN=Atlantis
+Validity: from 7/9/2012 3:10:38 AM UTC to 7/9/2013 3:10:37 AM UTC
+-----BEGIN CERTIFICATE-----
+MIIBmTCCAUegAwIBAgIBKjAJBgUrDgMCHQUAMBMxETAPBgNVBAMTCEF0bGFudGlz
+MB4XDTEyMDcwOTAzMTAzOFoXDTEzMDcwOTAzMTAzN1owEzERMA8GA1UEAxMIQXRs
+YW50aXMwXDANBgkqhkiG9w0BAQEFAANLADBIAkEAu+BXo+miabDIHHx+yquqzqNh
+Ryn/XtkJIIHVcYtHvIX+S1x5ErgMoHehycpoxbErZmVR4GCq1S2diNmRFZCRtQID
+AQABo4GJMIGGMAwGA1UdEwEB/wQCMAAwIAYDVR0EAQH/BBYwFDAOMAwGCisGAQQB
+gjcCARUDAgeAMB0GA1UdJQQWMBQGCCsGAQUFBwMCBggrBgEFBQcDAzA1BgNVHQEE
+LjAsgBA0jOnSSuIHYmnVryHAdywMoRUwEzERMA8GA1UEAxMIQXRsYW50aXOCASow
+CQYFKw4DAh0FAANBAKi6HRBaNEL5R0n56nvfclQNaXiDT174uf+lojzA4lhVInc0
+ILwpnZ1izL4MlI9eCSHhVQBHEp2uQdXJB+d5Byg=
+-----END CERTIFICATE-----
+` + "\t\r\n"
 
 	RSA_PRIVATE_KEY = `-----BEGIN RSA PRIVATE KEY-----
 MIIJKQIBAAKCAgEAsgzs6vN2sveHVraXV0zdoVyhWUHWNQ0xnhHTPhjt5ggHmSvr
@@ -132,5 +154,28 @@ func secretdata(cert, key string) map[string][]byte {
 	return map[string][]byte{
 		v1.TLSCertKey:       []byte(cert),
 		v1.TLSPrivateKeyKey: []byte(key),
+	}
+}
+
+// caBundleData returns a CA certificate bundle map whose value is
+// the given set of PEM certificates intermingled with some non-PEM
+// data.
+//
+// See also: https://tools.ietf.org/html/rfc7468#section-5.2
+func caBundleData(cert ...string) map[string][]byte {
+	var data string
+
+	data += "start of CA bundle\n"
+
+	for n, c := range cert {
+		data += fmt.Sprintf("certificate %d\n", n)
+		data += c
+		data += "\n"
+	}
+
+	data += "end of CA bundle\n"
+
+	return map[string][]byte{
+		"ca.crt": []byte(data),
 	}
 }
--- a/internal/dag/cache_test.go
+++ b/internal/dag/cache_test.go
@@ -56,6 +56,42 @@ func TestKubernetesCacheInsert(t *testing.T) {
 			},
 			want: true,
 		},
+		"insert CA secret w/ explanatory text": {
+			obj: &v1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "secret",
+					Namespace: "default",
+				},
+				Type: v1.SecretTypeOpaque,
+				Data: map[string][]byte{
+					"ca.crt": []byte(CERTIFICATE_WITH_TEXT),
+				},
+			},
+			want: true,
+		},
+		"insert CA bundle secret w/ non-PEM data": {
+			obj: &v1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "secret",
+					Namespace: "default",
+				},
+				Type: v1.SecretTypeOpaque,
+				Data: caBundleData(CERTIFICATE, CERTIFICATE, CERTIFICATE, CERTIFICATE),
+			},
+			want: true,
+		},
+		"insert CA bundle secret w/ non-PEM data and no certificates": {
+			obj: &v1.Secret{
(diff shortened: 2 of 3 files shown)


internal/dag/secret_test.go
```diff
@@ -51,6 +55,24 @@ R5M5azDV1CIhIeOTiPA/mq5fL1UrgVbB+IATIsUAQfuWivDyoeu96LB/QswyHAWG
...
+MB4XDTEyMDcwOTAzMTAzOFoXDTEzMDcwOTAzMTAzN1owEzERMA8GA1UEAxMIQXRs
+YW50aXMwXDANBgkqhkiG9w0BAQEFAANLADBIAkEAu+BXo+miabDIHHx+yquqzqNh
+Ryn/XtkJIIHVcYtHvIX+S1x5ErgMoHehycpoxbErZmVR4GCq1S2diNmRFZCRtQID
+AQABo4GJMIGGMAwGA1UdEwEB/wQCMAAwIAYDVR0EAQH/BBYwFDAOMAwGCisGAQQB
+gjcCARUDAgeAMB0GA1UdJQQWMBQGCCsGAQUFBwMCBggrBgEFBQcDAzA1BgNVHQEE
+LjAsgBA0jOnSSuIHYmnVryHAdywMoRUwEzERMA8GA1UEAxMIQXRsYW50aXOCASow
+CQYFKw4DAh0FAANBAKi6HRBaNEL5R0n56nvfclQNaXiDT174uf+lojzA4lhVInc0
+ILwpnZ1izL4MlI9eCSHhVQBHEp2uQdXJB+d5Byg=
+-----END CERTIFICATE-----
+`
```
davecheney: Could you add `\t\r\n` at the end of this fixture.

lrouquette:
Thx much! Any plans to backport this fix to `0.15.x`?

davecheney:
We're planning on releasing 1.0 at the end of the week which will contain
this fix.

erwbgy:
Could this issue at least be documented so that users don't upgrade to 0.1.5.3 and go to 1.0.0 instead?  This is a pretty nasty breakage which was hard to diagnose and is not mentioned in the 0.15.3 upgrade guide or the 1.0.0 bug fixes.

davecheney:
@erwbgy I’m sorry you are having issues. There is a bit to unpack from your comment so please bare with me. 

This fix was not included in 0.15.3 (to the best of my recollections), but that shouldn’t be a problem because 0.15 is not supported any more. If you have found some documentation that contradicts this I’d appreciate it if you could let me know so I can get it fixed. 

WRT to this commit, the change made contour more permissive when validating certificates, specifically when they contained white space padding. I’ll double check the release notes to see if we specifically called this out but in general this change was one of several that were intended to make the validation more useful after it first landed in 0.15. 

So we don’t loose this conversation I would greatly appreciate it if you would log a new issue that details the problems you’re having and we’ll do our best to address it. 

Thanks

Dave

erwbgy:
Thanks @davecheney.  We upgraded from 0.15.1 to 0.15.3 and some of our users started seeing their IngressRoute objects that were working before marked as invalid with TLS Secret not found or is malformed.  After some investigation we found that was because the certificates in the Secrets referenced by the IngressRoutes had trailing newlines.  I was just about to create an issue with details of how to reproduce when I found #1809 that does this already.

I really appreciate the documentation that goes into every release and try to read it carefully before each upgrade, but I found nothing to indicate 0.15.3 had this known breaking issue.  If so I would have skipped this release and done the extra work to go straight to 1.0.0.  My ask is that a note be added to 0.15.3 now to avoid anyone else getting into this situation.

davecheney:
Thank you for your reply. I think the situation is reverse. The check was not added to 0.15.3 and later removed from 1.0.0. Attempts to validate certificates occurred throughout 0.15.x and culminated in improvements in the 1.0.0 beta and rc series leading up to 1.0.0. 

Perhaps we should add that 0.15.3 has a bug where it rejects certificates with trailing whitespace, but given the current stable release is 1.0.0, folks shouldn’t be using anything older.

PR #1812, opened 2019-10-28
internal/dag: reject a TCPProxy without services or includes

### LINKED ISSUE #1776: internal/dag: httpproxy.spec.tcpproxy must have include or services[] set
At the moment this fragment is accepted by internal/dag/builder
```
# httpproxy
spec:
  tcpproxy:
```
But it shouldn't be. tcpproxies with neither include or len(services) <1 should be invalid.

jpeach:
If a `TCPProxy` object doesn't have any `Services` and does not
`Include` a delegate, then it's an empty no-op. Reject it with a
status message in this case.

This fixes #1776.


jpeach:
@davecheney Updated the tests.

PR #1814, opened 2019-10-28
site: add envoy compatibility matrix

### LINKED ISSUE #1394: Contour should document the Envoy extensions it requires
The docker.io/envoyproxy/envoy image we use in our examples ships with all the envoy extensions compiled in. Using Envoy as a reverse proxy does not require all the extensions Envoy's image ships with. We should document the extensions we need so that if users want to bring their own Envoy they can be reasonably confident that it will work.

davecheney:
Fixes #1394

Add a document recording the compatibility between Envoy and Contour
versions.

Also, document the extension that are required by Contour.

### DIFF
--- a/site/_resources/envoy.md
+++ b/site/_resources/envoy.md
@@ -1,0 +1,38 @@
+---
+title: Envoy compatibility matrix
+layout: page
+---
+
+Due to the agressive deprecation cycle of Envoy's xDS API, not all versions of Contour will work with all versions of Envoy, and vice versa.
+
+This page describes the compatibility matrix of Contour and Envoy versions.
+
+## Supported Envoy versions
+
+|              | Contour 1.0.0 | 
+| ------------ | :-----------:|
+| Envoy 1.11.0 | Not supported<sup>1</sup> |
+| Envoy 1.11.1 | Not supported<sup>2</sup> |
+| Envoy 1.11.2 |  Supported | 
+
+#### Notes
+
+1. [CVE-2019-9512, CVE-2019-9513, CVE-2019-9514, CVE-2019-9515, CVE-2019-9518][1] 
+2. [CVE-2019-15225, CVE-2019-15226][2]
+
+## Envoy extensions
+
+Contour requires the following extensions.
+If you are using the image recommended in our [example deployment](https://github.com/projectcontour/contour/blog/{{ site.github.latest_release.tag_name }}/examples/contour) no action is required.
+If you are providing your own Envoy it must be compiled with the following extensions:
+
+- `access_loggers`: `envoy.file_access_log`,`envoy.http_grpc_access_log`,`envoy.tcp_grpc_access_log`
+- `filters.http`: `envoy.buffer`,`envoy.cors`,`envoy.csrf`,`envoy.ext_authz`,`envoy.fault`,`envoy.filters.http.adaptive_concurrency`,`envoy.filters.http.dynamic_forward_proxy`,`envoy.filters.http.grpc_http1_reverse_bridge`,`envoy.filters.http.grpc_stats`,`envoy.filters.http.header_to_metadata`,`envoy.filters.http.jwt_authn`,`envoy.filters.http.original_src`,`envoy.filters.http.rbac`,`envoy.filters.http.tap`,`envoy.grpc_http1_bridge`,`envoy.grpc_json_transcoder`,`envoy.grpc_web`,`envoy.gzip`,`envoy.health_check`,`envoy.ip_tagging`,`envoy.rate_limit`,`envoy.router`,`envoy.squash`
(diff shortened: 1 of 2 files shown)


site/_resources/envoy.md
```diff
@@ -0,0 +1,38 @@
...
+|              | Contour 1.0.0 | 
+| ------------ | :-----------:|
+| Envoy 1.11.0 | Not supported<sup>1</sup> |
+| Envoy 1.11.1 | Not supported<sup>2</sup> |
+| Envoy 1.11.2 |  Supported | 
+
+#### Notes
+
+1. [CVE-2019-9512, CVE-2019-9513, CVE-2019-9514, CVE-2019-9515, CVE-2019-9518][1] 
+2. [CVE-2019-15225, CVE-2019-15226][2]
```
jpeach: I was expecting the superscripts to be clickable. How about moving these notes to the end of the doc and adding anchors so the superscripts can link to them?

site/_resources/envoy.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Envoy extensions
+
+Contour requires the following extensions.
+If you are using the image recommended in our [example deployment](https://github.com/projectcontour/contour/blog/{{ site.github.latest_release.tag_name }}/examples/contour) no action is required.
+If you are providing your own Envoy it must be compiled with the following extensions:
+
```
jpeach: This ends up with a lot of red text. It's probably impractical to link to Envoy docs for each plugin, so maybe format this as a table?

| Access Loggers |  envoy.file_access_log , envoy.http_grpc_access_log, envoy.tcp_grpc_access_log |

site/_resources/envoy.md
```diff
@@ -0,0 +1,38 @@
...
+- `access_loggers`: `envoy.file_access_log`,`envoy.http_grpc_access_log`,`envoy.tcp_grpc_access_log`
+- `filters.http`: `envoy.buffer`,`envoy.cors`,`envoy.csrf`,`envoy.ext_authz`,`envoy.fault`,`envoy.filters.http.adaptive_concurrency`,`envoy.filters.http.dynamic_forward_proxy`,`envoy.filters.http.grpc_http1_reverse_bridge`,`envoy.filters.http.grpc_stats`,`envoy.filters.http.header_to_metadata`,`envoy.filters.http.jwt_authn`,`envoy.filters.http.original_src`,`envoy.filters.http.rbac`,`envoy.filters.http.tap`,`envoy.grpc_http1_bridge`,`envoy.grpc_json_transcoder`,`envoy.grpc_web`,`envoy.gzip`,`envoy.health_check`,`envoy.ip_tagging`,`envoy.rate_limit`,`envoy.router`,`envoy.squash`
```
jpeach: Is there a way we can verify this against the code?

site/_resources/envoy.md
davecheney: Initially I had the CVE's inline, but the table looked terribly unbalanced. The goal of moving them to notes was to hide the detail from folks who just wanted the TL;DR on which version they should use.

site/_resources/envoy.md
davecheney: I thought about that, but putting it in a table was just too close to being one of those people who keeps lists of things in excel. I felt it was easier to copy and paste using a <li>

site/_resources/envoy.md
davecheney: Not at the moment. We cannot query envoy for the options it supports, and if we send envoy a message that it doesn't understand because there is no filter for it, then it will reject it an wedge the stream. I've intentionally cast the net very wide for this first version, I hope we can refine it over time.

site/_resources/envoy.md
jpeach: I agree that having them as notes make sense. I just kept trying to click on them :)

davecheney: I'm not jazzed about dropping to html for this. Perhaps in the future we can find a Jekyll footnote plugin.

jpeach: Yeh, maybe @SDBrett has some ideas?

davecheney:
> I'm fine with coming back to tweak formatting, but I do think this is reference documentation and should go in the "Documentation" menu.

Yes, this is reference documentation, but for completely perverse reasons I don't want it to be versioned in under the _documentation_ top level nav as while that _is_ reference documentation, its versioned reference documentation, whereas I see this page as being evergreen and applicable to multiple versions -- this is why I felt a table would be useful for capturing the matrix of contour <> envoy versions over time.

jpeach:
> On 29 Oct 2019, at 3:11 pm, Dave Cheney <notifications@github.com> wrote:
> 
> I'm fine with coming back to tweak formatting, but I do think this is reference documentation and should go in the "Documentation" menu.
> 
> Yes, this is reference documentation, but for completely perverse reasons I don't want it to be versioned in under the documentation top level nav as while that is reference documentation, its versioned reference documentation, whereas I see this page as being evergreen and applicable to multiple versions -- this is why I felt a table would be useful for capturing the matrix of contour <> envoy versions over time.

OK, then let's land this as is.

PR #1818, opened 2019-10-29
Update render.sh to change ImagePullPolicy to IfNotPresent

### LINKED ISSUE #1528: Review imagePullPolicy setttings for releases
Currently, the imagePullPolicy for the example deployment is set to `Always`. For releases, this is probably not necessary, since the image will not change very regularly.

This issue is to reexamine how we generate release examples, and if we should change this behaviour.

youngnick:
For examples/render/contour.yaml only.

Fixes #1528.

### DIFF
--- a/examples/render.sh
+++ b/examples/render.sh
@@ -1,3 +1,6 @@
 #! /usr/bin/env bash
 
-ls contour/*.yaml | xargs cat render/gen-warning.yaml > render/contour.yaml
+ls contour/*.yaml | \
+  xargs cat render/gen-warning.yaml | \
+  sed 's/imagePullPolicy: Always/imagePullPolicy: IfNotPresent/g' \
+  > render/contour.yaml
--- a/examples/render/contour.yaml
+++ b/examples/render/contour.yaml
@@ -1030,7 +1030,7 @@ spec:
       containers:
       - name: contour
         image: docker.io/projectcontour/contour:master
-        imagePullPolicy: Always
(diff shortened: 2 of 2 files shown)


examples/render.sh
```diff
@@ -1,3 +1,7 @@
 #! /usr/bin/env bash
 
 ls contour/*.yaml | xargs cat render/gen-warning.yaml > render/contour.yaml
+
+# MacOS' default sed requires an extension to use -i. Thanks, BSD.
+ sed -i'.bak' 's/imagePullPolicy: Always/imagePullPolicy: IfNotPresent/g' render/contour.yaml
```
davecheney: Why not do the sed inline on line 3?

youngnick: 🤦‍♂

examples/render.sh
```diff
@@ -1,3 +1,3 @@
 #! /usr/bin/env bash
 
-ls contour/*.yaml | xargs cat render/gen-warning.yaml > render/contour.yaml
+ls contour/*.yaml | xargs cat render/gen-warning.yaml | sed 's/imagePullPolicy: Always/imagePullPolicy: IfNotPresent/g' > render/contour.yaml
```
davecheney: for extra sizzle, maybe
```
ls contour/*.yaml | \
    xargs cat render/gen-warning.yaml | \
    sed 's/imagePullPolicy: Always/imagePullPolicy: IfNotPresent/g' \
    > render/contour.yaml
```

PR #1820, opened 2019-10-29
Adds a check for duplicate Conditions on a HTTPProxy include

### LINKED ISSUE #1770: Duplicate prefix conditions included to separate proxies not enforced
If I create a root proxy which delegates the same prefix to two different proxies in different namespaces, the root proxy is not set to error, but instead are `valid`.

This breaks the multi-team model since the same path is referenced in different namespaces.

```
kind: HTTPProxy
metadata:
  name: root
  namespace: projectcontour-roots
spec:
  includes:
  - conditions:
    - prefix: /blog
    name: blogsite
    namespace: projectcontour-marketing
  - conditions:
    - prefix: /blog
    name: defaultsite
    namespace: default
```

image

stevesloka:
Fixes #1770 by looking for duplicate header conditions on an Include. 

### DIFF
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -23,6 +23,7 @@ import (
 	"k8s.io/api/networking/v1beta1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 
+	"github.com/google/go-cmp/cmp"
 	ingressroutev1 "github.com/projectcontour/contour/apis/contour/v1beta1"
 	projcontour "github.com/projectcontour/contour/apis/projectcontour/v1"
 )
@@ -527,9 +528,15 @@ func (b *Builder) computeRoutes(sw *ObjectStatusWriter, proxy *projcontour.HTTPP
 
 	visited = append(visited, proxy)
 	var routes []*Route
+
+	// Check for duplicate conditions on the includes
+	if includeConditionsIdentical(proxy.Spec.Includes) {
+		sw.SetInvalid("duplicate conditions defined on an include")
+		return nil
+	}
+
 	// Loop over and process all includes
 	for _, include := range proxy.Spec.Includes {
-
 		namespace := include.Namespace
 		if namespace == "" {
 			namespace = proxy.Namespace
@@ -628,6 +635,22 @@ func (b *Builder) computeRoutes(sw *ObjectStatusWriter, proxy *projcontour.HTTPP
 	}
 	sw.SetValid()
 	return routes
+}
+
+func includeConditionsIdentical(includes []projcontour.Include) bool {
+	j := 0
+	for i := 1; i < len(includes); i++ {
+		// Now compare each include's set of conditions
+		for _, cA := range includes[i].Conditions {
+			for _, cB := range includes[j].Conditions {
+				if (cA.Prefix == cB.Prefix) && cmp.Equal(cA.Header, cB.Header) {
+					return true
+				}
+			}
+		}
+		j++
+	}
+	return false
 }
 
 // buildDAG returns a *DAG representing the current state of this builder.
--- a/internal/e2e/rds_test.go
+++ b/internal/e2e/rds_test.go
@@ -2855,6 +2855,144 @@ func TestRDSHTTPProxyRootCannotDelegateToAnotherRoot(t *testing.T) {
 	}, streamRDS(t, cc))
 }
 
+func TestRDSHTTPProxyDuplicateIncludeConditions(t *testing.T) {
+	rh, cc, done := setup(t)
+	defer done()
+
+	svc1 := &v1.Service{
+		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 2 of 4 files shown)


internal/e2e/rds_test.go
```diff
@@ -2855,6 +2855,144 @@ func TestRDSHTTPProxyRootCannotDelegateToAnotherRoot(t *testing.T) {
...
+	}
+	rh.OnAdd(proxyRoot)
+	rh.OnAdd(proxyChildA)
+	rh.OnAdd(proxyChildB)
+
+	assert.Equal(t, &v2.DiscoveryResponse{
+		VersionInfo: "2",
+		Resources: resources(t,
+			envoy.RouteConfiguration("ingress_http"),
+			envoy.RouteConfiguration("ingress_https"),
```
stevesloka: I'm a bit confused as to why the http_listener got spun up, but I see this in other tests as well.

internal/dag/builder.go
```diff
@@ -527,9 +528,16 @@ func (b *Builder) computeRoutes(sw *ObjectStatusWriter, proxy *projcontour.HTTPP
 
 	visited = append(visited, proxy)
 	var routes []*Route
-	// Loop over and process all includes
-	for _, include := range proxy.Spec.Includes {
 
+	// Check for duplicate conditions on the includes
+	includes := proxy.Spec.Includes
+	if b.includeConditionsIdentical(includes) {
```
jpeach: What happens if the same prefix condition is used on a route and an include?

internal/dag/builder.go
```diff
@@ -630,6 +638,29 @@ func (b *Builder) computeRoutes(sw *ObjectStatusWriter, proxy *projcontour.HTTPP
...
+	j := 0
+	prefixesMatch := false
+	headersMatch := false
+	for i := 1; i < len(includes); i++ {
+		// Now compare each include's set of conditions
+		for _, cA := range includes[i].Conditions {
+			for _, cB := range includes[j].Conditions {
+				// Check if the path is the same
+				if strings.TrimSuffix(cA.Prefix, "/") == strings.TrimSuffix(cB.Prefix, "/") {
+					prefixesMatch = true
```
jpeach: I thought that this would be allowed:
```
includes:
- name: foo
   conditions:
   - prefix: /foo
- name: next-foo
   conditions:
   - prefix: /foo/
```

internal/dag/builder.go
jpeach: What if we add `func (p *HttpProxySpec) AllConditions() []Condition` to get all the include and route conditions for checking?

stevesloka: Defining duplicate routes in a Spec.Match is a possibility that we should look into (I'll open an issue). My biggest concern with this PR was to solve the multi-team component where folks could define the same include across different teams (which is one of our big selling points).

stevesloka: https://github.com/projectcontour/contour/issues/1834

davecheney:
I’m really concerned about the timeline. Is it critical we land this for 1.0.0 or could it wait for a point release in a few weeks?

internal/dag/builder.go
```diff
@@ -630,6 +637,33 @@ func (b *Builder) computeRoutes(sw *ObjectStatusWriter, proxy *projcontour.HTTPP
...
 }
 
+func includeConditionsIdentical(includes []projcontour.Include) bool {
+	j := 0
+	prefixesMatch := false
+	headersMatch := false
+	for i := 1; i < len(includes); i++ {
+		// Now compare each include's set of conditions
+		for _, cA := range includes[i].Conditions {
+			for _, cB := range includes[j].Conditions {
```
jpeach: Nit: You could now simplify the inner check:
```Go
if (cA.Prefix == cB.Prefix) && cmp.Equal(cA.Header, cB.Header) {
    return true
}
```

internal/dag/builder.go
```diff
@@ -527,9 +528,15 @@ func (b *Builder) computeRoutes(sw *ObjectStatusWriter, proxy *projcontour.HTTPP
 
 	visited = append(visited, proxy)
 	var routes []*Route
+
+	// Check for duplicate conditions on the includes
+	if includeConditionsIdentical(proxy.Spec.Includes) {
+		sw.SetInvalid("duplicate conditions defined on an include is invalid")
```
jpeach: Nit: since we are setting the status to invalid, can we remove that from the reason and just say: "duplicate include conditions" or "duplicate conditions defined on an include"?

internal/dag/builder.go
stevesloka: Nice! I tend to make things simpler which makes the code read on and on, this makes it more clear. =)

internal/dag/builder.go
stevesloka: :+1:

PR #1832, opened 2019-10-30
docs: update upgrade guide for Contour 1.0

### LINKED ISSUE #1604: docs/upgrading.md: document how to upgrade from 0.15.x to 1.0.0
We need to document how to upgrade from 0.15.0 too 1.0.0. This should be a matter of summarising the 0.15 -> beta.1 -> rc.1 -> rc.2 instructions.

Hopefully the delta between rc.2 and 1.0.0 will be small enough we can just say "see rc.1 to rc.2 upgrades".

jpeach:
Update the upgrade guide for Contour 1.0. Remove all the release
candidate information and roll it into a single section that describes
changes from 0.15.3 to 1.0.0.

This fixes #1604.

### DIFF
--- a/site/_resources/upgrading.md
+++ b/site/_resources/upgrading.md
@@ -3,128 +3,102 @@ title: Upgrading Contour
 layout: page
 ---
 
+<!-- NOTE: this document should be formatted with one sentence per line to made reviewing easier. -->
+
 This document describes the changes needed to upgrade your Contour installation.
 
-## Upgrading Contour 1.0.0-rc.1 to 1.0.0-rc.2
+# Upgrading Contour 0.15.3 to 1.0.0
 
-Contour 1.0.0-rc.2 is the second release candidate for Contour 1.0.
+Contour 1.0.0 is the current stable release.
 
-### Release candidate
+<div class="alert-deprecation">
+<b>Deprecation Notice</b></br>
+<p>
+The <code>IngressRoute</code> CRD has been deprecated and will not receive further updates. Contour 1.0.0 continues to support the IngressRoute API, however we anticipate it will be removed in the future.
+<p>
+</p>
+Please see the documentation for <a href="{% link _docs_1_0/httpproxy.md %}"><code>HTTPProxy</code></a>, which is the successor to <code>IngressRoute</code>.
+You can also read the <a href="{% link _guides/ingressroute-to-httpproxy.md %}">IngressRoute to HTTPProxy upgrade</a> guide.
+</p>
+</div>
 
-Contour 0.15.3 remains the current stable release.
-The `:latest` tag will continue to point to 0.15.3 until Contour 1.0.0 is released.
+&nbsp;
+
+## Recommended Envoy version
+
+The recommended version of Envoy remains unchanged.
+Ensure the Envoy image version is `docker.io/envoyproxy/envoy:v1.11.2`.
 
 ## The easy way to upgrade
 
 If the following are true for you:
 
  * Your previous installation is in the `projectcontour` namespace.
- * You are using one of the [example]({{ site.github.repository_url }}/blob/v1.0.0-rc.2/examples/) deployments.
+ * You are using one of the [example]({{ site.github.repository_url }}/blob/v1.0.0/examples/) deployments.
  * Your cluster can take few minutes of downtime.
 
-Then the simplest way to upgrade to 1.0.0-rc.2 is to delete the `projectcontour` namespace and reapply the `examples/contour` sample manifest.
+Then the simplest way to upgrade is to delete the `projectcontour` namespace and reapply the `examples/contour` sample manifest.
 From the root directory of the repository:
-```
-kubectl delete namespace projectcontour
 
-kubectl apply -f examples/contour
+```
+$ kubectl delete namespace projectcontour
+$ kubectl apply -f examples/contour
 ```
 
+This will remove both the Envoy and Contour pods from your cluster and recreate them with the updated configuration.
 If you're using a `LoadBalancer` Service, deleting and recreating may change the public IP assigned by your cloud provider.
 You'll need to re-check where your DNS names are pointing as well, using [Get your hostname or IP address]({% link _guides/deploy-options.md %}#get_your_hostname_or_ip_address).
 
 ## The less easy way
 
-This section contains information for administrators who wish to apply the Contour 1.0.0-rc.1 to Contour 1.0.0-rc.2 changes manually.
+This section contains information for administrators who wish to manually upgrade from Contour 1.5.3 to Contour 1.0.0.
 
-### Upgrade to Contour 1.0.0-rc.2
+### Upgrade to Contour 1.0.0
 
-Change the Contour image version to `docker.io/projectcontour/contour:v1.0.0-rc.2`.
+Change the Contour image version to `docker.io/projectcontour/contour:v1.0.0`.
+
+Note that as part of sunsetting the Heptio brand, Contour Docker images have moved from `gcr.io/heptio-images` to `docker.io/projectcontour`.
 
 ### Reapply HTTPProxy and IngressRoute CRD definitions
 
-Contour 1.0.0-rc.2 ships with updated OpenAPIv3 validation schemas.
-```
-kubectl apply -f examples/contour/01-crds.yaml
-```
+Contour 1.0.0 ships with updated OpenAPIv3 validation schemas.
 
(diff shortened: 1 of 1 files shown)


site/_resources/upgrading.md
```diff
@@ -5,126 +5,122 @@ layout: page
...
 This document describes the changes needed to upgrade your Contour installation.
 
-## Upgrading Contour 1.0.0-rc.1 to 1.0.0-rc.2
+# Upgrading Contour 0.15.3 to 1.0.0
 
-Contour 1.0.0-rc.2 is the second release candidate for Contour 1.0.
+Contour 1.0.0 is the current stable release.
 
-### Release candidate
+<div class="alert-deprecation">
```
davecheney: Nit: in the future, please follow the one line per sentence pattern. It helps to be able to comment on a specific sentence and rewriting individual sentences tends to have a lower diff count using this pattern.

site/_resources/upgrading.md
```diff
@@ -5,126 +5,122 @@ layout: page
...
+<b>Deprecation Notice</b></br>
+<p>
+The <code>IngressRoute</code> CRD has been deprecated and will not receive
+further updates. Contour 1.0.0 continues to support the IngressRoute
+API, however we anticipate it will be removed in the future.
+<p>
+</p>
+Please see the documentation for <a href="/docs/1.0/httpproxy"><code>HTTPProxy</code></a>,
```
davecheney: {% link _docs_1_0/httpproxy.md %} which will break if we accidentally rename the document in the future.

site/_resources/upgrading.md
```diff
@@ -5,126 +5,122 @@ layout: page
...
+which is the successor to <code>IngressRoute</code>.
+You can also read the <a href="/guides/ingressroute-to-httpproxy">IngressRoute to HTTPProxy upgrade</a>
```
davecheney: same, please use the {% link syntax to help catch broken links.

site/_resources/upgrading.md
```diff
@@ -5,126 +5,122 @@ layout: page
...
+guide.
+</p>
+</div>
 
-Contour 0.15.3 remains the current stable release.
-The `:latest` tag will continue to point to 0.15.3 until Contour 1.0.0 is released.
+<p></p>
```
davecheney: maybe `<hr/>` or &nbsp would be better here to create a separator.

site/_resources/upgrading.md
```diff
@@ -5,126 +5,122 @@ layout: page
...
+changes in the future. See the [HTTPProxy documentation]({% link
+_docs_1_0/httpproxy.md %}) for more information.
 
-Then the simplest way to upgrade to 1.0.0-rc.1 is to delete the `projectcontour` namespace and reapply the `examples/contour` sample manifest.
-From the root directory of the repository:
 ```
-kubectl delete namespace projectcontour
-
-kubectl apply -f examples/contour
+$ kubectl apply -f examples/contour/01-crds.yaml
```
davecheney: I don't think this is sufficient when going from 0.15 to 1.0. @youngnick, thoughts?

site/_resources/upgrading.md
```diff
@@ -5,126 +5,122 @@ layout: page
...
-### projectcontour.io/v1 group version
+If you are upgrading a cluster that you previously installed a 1.0.0
+release candidate, note that Contour 1.0.0 moves the HTTPProxy CRD
+from `projectcontour.io/v1alpha1` to `projectcontour.io/v1`.  and
+will no longer recognize the former group version.
 
-Contour 1.0.0-rc.1 moves the HTTPProxy CRD from `projectcontour.io/v1alpha1` to `projectcontour.io/v1`.
-Contour will no longer recognize the former group version.
-Please edit your HTTPProxy documents to update their group verision to `projectcontour.io/v1`.
+Please edit your HTTPProxy documents to update their group verision
```
davecheney: s/verision/version

site/_resources/upgrading.md
```diff
@@ -5,126 +5,122 @@ layout: page
...
+* The per-service health check key, `healthcheck` has moved to
+  per-route and has been renamed `healthCheckPolicy`.
 
-Before:
+<table class="table table-borderless" style="border: none;">
+<tr><th>Before:</th><th>After:</th></tr>
 
-```yaml
+<tr>
+<td><pre><code class="language-yaml" data-lang="yaml">
```
davecheney: is this a copy and paste from some other text, why did you move away from the

before

after

style?

I don't mind it, but this feels like something we can do later across the whole site.

site/_resources/upgrading.md
jpeach: I updated the diff to do this, and added a comment for future editors.

site/_resources/upgrading.md
jpeach: Fixed,

site/_resources/upgrading.md
jpeach: Fixed.

site/_resources/upgrading.md
jpeach: This is forcing a bit of vertical space, otherwise the deprecation box bumps right up on the next heading and looks strange. I tried the `hr` but I think that also looks odd, since there aren't clear parts of the content to separate. `nbsp` works fine though.

site/_resources/upgrading.md
jpeach: Fixed.

site/_resources/upgrading.md
jpeach: I found it hard to compare the before and after, so I wanted to put them next to one another. You can't do that in a Markdown table, so one think led to another ...

jpeach:
@davecheney Updated, PTAL.

site/_resources/upgrading.md
jpeach: Per Slack, @youngnick thinks this is enough. I can test this if you point me in the right direction ...

youngnick: Can you check that v1alpha1 HTTPProxy is gone from `kubectl get crd`? I think it should be, as we change the definition and remove v1alpha1 from the supported versions. I can't see it in my cluster, but can't remember if I needed to do anything special for that when I was fiddling with versions.

site/_resources/upgrading.md
```diff
@@ -184,81 +161,18 @@ spec:
...
-
-As part of sunsetting the Heptio brand Docker images have moved from `gcr.io/heptio-images` to `docker.io/projectcontour`.
-
-Change the Contour image version to `docker.io/projectcontour/contour:v1.0.0-beta.1`.
-
-### Recommended Envoy version
-
-The recommended version of Envoy remains unchanged from Contour 0.15.3.
-Ensure the Envoy image version is `docker.io/envoyproxy/envoy:v1.11.2`.
+If you deploy Contour into different namespace you will need to pass `contour bootstrap --namespace=<namespace>` and update the leader election parameters in the [`contour.yaml` configuration]({% link _docs_1_0/configuration.md %})
```
youngnick: Nit: 'into a different namespace`

site/_resources/upgrading.md
youngnick: I don't like the direct HTML inclusion, but those differences are *much* easier to see. I think we should do this one, but try and find a better table solution for the rest of the documentation.

site/_resources/upgrading.md
youngnick: We will need to have instructions on removing the old CRDs when we completely remove them, but that's for later.

jpeach:
> On 31 Oct 2019, at 1:39 pm, Nick Young <notifications@github.com> wrote:
> 
> @youngnick approved this pull request.
> 

[snip]

> 
> In site/_resources/upgrading.md:
> 
> >  
> -```yaml
> +<tr>
> +<td><pre><code class="language-yaml" data-lang="yaml">
> 
> I don't like the direct HTML inclusion, but those differences are much easier to see. I think we should do this one, but try and find a better table solution for the rest of the documentation.

Yes, this is pretty unfortunate. Once you step outside basic Markdown, all bets are off and you have to HTML the whole thing. I don't think there's any real way around it.

site/_resources/upgrading.md
davecheney: Thanks for wrestling with this, let’s keep working on harmonising the formatting next week.

site/_resources/upgrading.md
jpeach: Fixed.

PR #1833, opened 2019-10-30
site: style deprecation blocks as warnings

jpeach:
Style the deprecation class as a warning (yellow) rather than an
error (red). The red is a bit too alarming.


jpeach:
I think this looks a better in yellow :)

davecheney:
Maybe the color will be an encouragement for us to remove them promptly and avoid creating them in the future. 🙂

PR #1838, opened 2019-10-31
Docs update

stevesloka:
This is a docs redesign that adds a menu to the /docs site.


site/_config.yml
```diff
@@ -44,6 +44,16 @@ defaults:
       type: "pages"
     values:
       layout: "default"
+  - scope:
+      path: docs/master
+    values:
+      version: master
+      layout: "docs"
+  - scope:
+      path: docs/v1.0.0
```
davecheney: This is really nice. Thank you.

site/_guides/ingressroute-to-httpproxy.md
```diff
@@ -140,7 +140,7 @@ spec:
 
 HTTPProxy offers additional ways to match incoming requests to routes.
 This document covers the conversion between the routing offered in IngressRoute and HTTPProxy.
-For a broader discussion of HTTPProxy routing, see the [Routing section of the HTTPProxy documentation]({% link _docs_1_0/httpproxy.md %}#routing).
+For a broader discussion of HTTPProxy routing, see the [Routing section of the HTTPProxy documentation](/docs/1.0.0/httpproxy.md#routing).
```
davecheney: Is it possible to use the link syntax? I like that it gives us some notification if we move a file and it breaks a link.

stevesloka: I've updated all of them which didn't move to /docs. I need to come up with a way to make the syntax dynamic which I think might be able to go into a new PR if that's acceptable.

SDBrett:
I'm a little concerned with the structure of the docs directory. There is a lack of consistency of the directory path a document is location. For example under deploy there is
- Deployment -> /docs/v1.0.0/deploy-options/
- Upgrade -> /resources/upgrading/

This might also be what is leading to an inconsistent experience, the side bar visible on some of the pages, but other pages do not have the side bar. Also I suspect that this will not assist with simplifying documentation management.

stevesloka:
@SDBrett the reason the sidebar disappears is that the other items are in different directories (e.g. resources or guides). Those aren't versioned the same as the other docs. Let me see if I can get them to merge, this would still allow them to live together in the menu, but not version behind the scenes.

stevesloka:
This can supersede #1851, all changes are merged together in this branch. I'd like to take @SDBrett 's comment's about the TOC menu for /guides & /resources into a new PR.

PR #1839, opened 2019-10-31
RFC: hoist DAG test fixtures.

jpeach:
This commit hoists test fixtures out of DAG tests into a commit
fixture store.

There are other approaches to this problem, so posting this as an
RFC to get consensus before I go too far down this path.

This updates #1449.

### DIFF
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -28,17 +28,21 @@ import (
 )
 
 func TestDAGInsert(t *testing.T) {
+	f := NewFixtures()
+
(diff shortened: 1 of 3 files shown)


jpeach:
Do not merge, just let me know what you think :)

youngnick:
I like that this would let us use a more descriptive name for each fixture - if we spent some time on the naming, we could avoid proxy150a, proxy150b, and so on.

jpeach:
Closing this for now. It's not clear that this improves things enough.

PR #1848, opened 2019-11-04
Add site proofing checks to travis-ci process

SDBrett:
Added site proofing process to travis-ci process.

Added script to detect if files within a directory have changes as part of a commit. Adds capability to determine only run travis jobs relevant to the files changed.

Travis matrix used to enable build tasks to be run in parallel and independently off each other.

Signed-off-by: Brett Johnson <brett@sdbrett.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,32 @@
-language: go
+matrix:
+  include:
+    - language: go
+      name: test make process
+      go:
+        - 1.13.x
+      sudo: false
+      cache:
+        directories:
+          - /home/travis/.cache/go-build
+          - /home/travis/gopath/pkg/mod
+      env:
+        - GOPROXY=https://proxy.golang.org/
+      script:
+        - make check
 
-go:
-  - 1.13.x
-
-sudo: false
-
-cache:
-  directories:
-  - /home/travis/.cache/go-build
-  - /home/travis/gopath/pkg/mod
-
-env:
-  - GOPROXY=https://proxy.golang.org/
-
-script:
-  - make check
+    - language: ruby
+      rvm: 2.6
+      name: Site Proofing
+      sudo: false # route your build to the container-based infrastructure for a faster build
+      cache: bundler
+      env:
+        - NOKOGIRI_USE_SYSTEM_LIBRARIES=true # speeds up installation of html-proofer
+        - SEARCH_DIRECTORIES="site"
+      addons:
+        apt:
+          packages:
+            - libcurl4-openssl-dev
+      install:
+        - ./hack/travis/travis-check-files-changed.sh $SEARCH_DIRECTORIES ; RETURN_CODE=$? ; if [ $RETURN_CODE -eq 137 ]; then travis_terminate 0; elif [ $RETURN_CODE -ne 0 ]; then travis_terminate $RETURN_CODE; fi
+      script:
+        - ./hack/site-proofing/cibuild
--- a/hack/travis/travis-check-files-changed.sh
+++ b/hack/travis/travis-check-files-changed.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+
+# Checks if the files changed in the last commit are contained within specifed directories.
+# This is used to determine if a job should run based on changes within a commit
+# PATHS_TO_SEARCH can be a single path "site" or multipl paths "site hack", response if for any path specified.
+# Add to a job within .travis.yml
+#      env:
+#        - SEARCH_DIRECTORIES="site"
+#      install:
+#        - ./hack/travis/travis-check-files-changed.sh $SEARCH_DIRECTORIES ; RETURN_CODE=$? ; if [ $RETURN_CODE -eq 137 ]; then travis_terminate 0; elif [ $RETURN_CODE -ne 0 ]; then travis_terminate $RETURN_CODE; fi
+
+set -e # halt script on error
+
+# 1. Make sure the paths to search are not empty
+if [ $# -eq 0 ]; then
+    usage
+    exit 1
+fi
+
+# 2. Get the latest commit
+LATEST_COMMIT=$(git rev-parse HEAD)
+
+# 3. Get the latest commit in the searched paths
+LATEST_COMMIT_IN_PATH=$(git log -1 --format=format:%H --full-diff "$@")
+
+if [ $LATEST_COMMIT != $LATEST_COMMIT_IN_PATH ]; then
+    echo "Exiting this job because code in the following paths have not changed:"
+    echo $@
+    exit 137
+fi
+
+echo "Changes detected in the following paths:"
+echo $@
+exit 0
--- a/hack/site-proofing/cibuild
+++ b/hack/site-proofing/cibuild
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+set -e # halt script on error
+
+# Builds and checks the website for broken links.
+
+readonly HERE="$(cd $(dirname $0) && pwd)"
+cd ${HERE}/../../site
+
+gem install bundler
+gem install html-proofer
+gem install jekyll -v 3.8.5
+bundle install
+bundle exec jekyll build
+htmlproofer ./_site \
+    --empty-alt-ignore \
+    --assume-extension \
+    --allow-missing-href \
+    --allow-hash-href


davecheney:
How critical is this? Jekyll runs inside netlify and they host the result.

davecheney:
Also, I want to explicitly acknowledge your enthusiasm and we’re grateful for the help, but please keep in mind our core contribution principle — talk, then code. Nobody likes to feel their waste their time, and a key to that is to get agreement on making a change before investing that effort. 

Thanks

SDBrett:
> How critical is this? Jekyll runs inside netlify and they host the result.

Low to moderate, the largest area of issue is not to the project itself, but to those contributing to the development of the projects site. Running `bundle install` will install packages as per `Gemfile.lock` and therefore placing potentially vulnerable gems on to those systems.

Good point on the comms, I'll work to raise as issues before running with an idea.

davecheney:
@SDBrett is this PR active? If not, would you please close it.

SDBrett:
It's still active, I'll rebase the branch today

SDBrett:
Performed rebase and PR is ready for review.

ciscripts/site-gem-vulnerability-test/cibuild
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -e # halt script on error
+
```
jpeach: Here, and in the other script, make this independent of the directory it is run from:

```bash
readonly HERE="$(cd $(dirname $0) && pwd)"

cd ${HERE}/../../site # probably need to change this after you move the script
```

.travis.yml
```diff
@@ -1,17 +1,42 @@
...
 
-go:
-  - 1.13.x
+    - language: ruby
+      rvm: 2.6
+      name: Site gem vulnerability check
+      sudo: false # route your build to the container-based infrastructure for a faster build
+      cache: bundler
+      before_script:
+        - chmod +x ./hack/site-gem-vulnerability-test/cibuild
```
davecheney: can you just +x the script in this commit?

SDBrett: Do you mean to change to the below?
```
script:
  - chmod +x ./hack/site-gem-vulnerability-test/cibuild
  - ./hack/site-gem-vulnerability-test/cibuild
```

SDBrett:
> How much time is this going to add to our CI builds?
>
> Can this be moved to Netlify as a post commit step instead?

All three tasks in the travis file run in parallel.

Current process as per master without my changes takes: 3 min 33 sec
Process with my changes takes: 5 min 57 sec

I am looking at options as to running site based checks only if there is a `git diff` within the /site directory. 

As for running with Netlify post commit steps, I do not know. I can have a look at that if it's an option you would like investigated further.

davecheney:
Thank you for your reply. I’m keen to avoid adding 80% to our CI runtime.

SDBrett:
> Thank you for your reply. I’m keen to avoid adding 80% to our CI runtime.

The vulnerability check is very check, under 1 minute. HTML proofer is what adds the most time.

davecheney:
What’s the reason for doing a vulnerabilty check? Doesn’t GitHub do that for us? Also, the site is built by netlify, it’s not a deliverable of the project so I’m not sure what the attack surface is for an vulnerable dependency.

SDBrett:
> What’s the reason for doing a vulnerabilty check? Doesn’t GitHub do that for us? Also, the site is built by netlify, it’s not a deliverable of the project so I’m not sure what the attack surface is for an vulnerable dependency.

The vulnerability check is primarily to assist those working on the site. This prevents them from installing vulnerable gems if they choose to install gems locally. This improves supply chain security which can prevent malicious code injection at build time; though I don't suspect that this has a high probability.


Git not 100% sure on how the scanner works with GitHub, but I know that it did not detect that the Gemfile.lock had a vulnerable version of rubyzip.

davecheney:
I’m finding it difficult  to see the value here — with the exception of myself I thought that most of y’all were using a docker workflow to do your site local builds. 

Even for myself, bundler downloads the gems into a ./vendor directory inside /site. 

What’s the threat model that you are working against?

SDBrett:
> I’m finding it difficult to see the value here — with the exception of myself I thought that most of y’all were using a docker workflow to do your site local builds. Even for myself, bundler downloads the gems into a ./vendor directory inside /site. What’s the threat model that you are working against?

Prevention of vulnerable code on developer systems or being used to build the site.
In your case you would have vulnerable gems within the ./vendor directory if they are specified within Gemfile.lock

davecheney:
But what’s the attack model? The gemfile.lock prevents newer gems — possibly with malware embedded in them— from being downloaded. So that leaves older gems which have vulnerabilities which can be triggered somehow by the Jekyll build process. How can attacker leverage Jekyll to execute malicious code in my system? I’m just not seeing this as something we should do.

SDBrett:
> But what’s the attack model? The gemfile.lock prevents newer gems — possibly with malware embedded in them— from being downloaded. So that leaves older gems which have vulnerabilities which can be triggered somehow by the Jekyll build process. How can attacker leverage Jekyll to execute malicious code in my system? I’m just not seeing this as something we should do.

Gems can be triggered outside the Jekyll build process. A gem on a local system can be executed as part of an attack, it doesn't require the build process. It's also good hygiene. 

From a site specific threat, it is to aid with prevention of malicious code into the website during the build process.

davecheney:
Thank you for working on this. I’m sorry but I’m not convinced this is a likely attack vector. I don’t want to invest ci time and engineering maintenance hours in this process when GitHub already does this albeit at a slightly lower frequency. 

Please remove this part of the pr.

davecheney:
Additionally I’d like to apologise for the back and forth here. It was my fault for not pushing for the discussion to occur on an issue under our contribution guidelines. This would have avoided wasting your time, I’m sorry, I’ll do better in the future.

SDBrett:
> Thank you for working on this. I’m sorry but I’m not convinced this is a likely attack vector. I don’t want to invest ci time and engineering maintenance hours in this process when GitHub already does this albeit at a slightly lower frequency. Please remove this part of the pr.

Will remove from the PR. 

I'll do some more testing on adding the ability to test if files have been changed and put that in it's own issue and PR. The goal being to reduce CI tasks running when not needed, for example only run html-proofer when there are changes in /site/

davecheney:
Thank you. Is there an issue for adding the HTML proofer?

SDBrett:
> Thank you. Is there an issue for adding the HTML proofer?

I couldn't find one, so have created. Issue: #1909

davecheney:
Thank you. Let’s continue the discussion on the issue about how to reduce the ci impact of this test.

site/_includes/contributors.html
```diff
@@ -18,7 +18,7 @@ <h2 class="mb-2">The {{ site.title }} Team</h2>
         <img src="{{ person.image }}" class="rounded-circle" alt="Person" />
         <div class="media-body align-self-center">
           <h6>{{ person.first_name }} {{ person.last_name }}</h6>
-        {{ person.content | markdownify }}    
+        {{ person.content | markdownify }}
```
jpeach: Can you please revert this change?

hack/travis/travis-check-files-changed.sh
```diff
@@ -0,0 +1,39 @@
+# Checks if the files changed in the last commit are contained within specifed directories.
```
jpeach: This should have a `#!/usr/bin/env bash` too.

.travis.yml
```diff
@@ -1,17 +1,36 @@
...
+      cache: bundler
+      env:
+        - NOKOGIRI_USE_SYSTEM_LIBRARIES=true # speeds up installation of html-proofer
+        - SEARCH_DIRECTORIES="site"
+      addons:
+        apt:
+          packages:
+            - libcurl4-openssl-dev
+      before_install:
+        - chmod +x ./hack/travis/travis-check-files-changed.sh
```
jpeach: You should not need to chmod these. Git preserves permissions, so chmod the script and commit them and they will still be executable.

hack/travis/travis-check-files-changed.sh
```diff
@@ -0,0 +1,39 @@
...
+#      install:
+#        - ./hack/travis/travis-check-files-changed.sh $SEARCH_DIRECTORIES ; RETURN_CODE=$? ; if [ $RETURN_CODE -eq 137 ]; then travis_terminate 0; elif [ $RETURN_CODE -ne 0 ]; then travis_terminate $RETURN_CODE; fi
+
+
+# 1. Get all the arguments of the script
+# https://unix.stackexchange.com/a/197794
+PATHS_TO_SEARCH="$*"
+
+# 2. Make sure the paths to search are not empty
+if [ -z "$PATHS_TO_SEARCH" ]; then
```
jpeach: `$#` gives you the number of arguments to the script. 
```bash
if [ $# -eq 0 ]; then
    usage
    exit 1
fi
```

hack/travis/travis-check-files-changed.sh
```diff
@@ -0,0 +1,39 @@
...
+    echo "Please provide the paths to search for."
+    echo "Example usage:"
+    echo "./exit-if-path-not-changed.sh path/to/dir1 path/to/dir2"
```
jpeach: This isn't the name of the script any more :)

hack/travis/travis-check-files-changed.sh
```diff
@@ -0,0 +1,39 @@
...
+    exit 1
+fi
+
+# 3. Get the latest commit
+LATEST_COMMIT=$(git rev-parse HEAD)
+
+# 4. Get the latest commit in the searched paths
+LATEST_COMMIT_IN_PATH=$(git log -1 --format=format:%H --full-diff $PATHS_TO_SEARCH)
```
jpeach: I think what you want here is:
```bash
LATEST_COMMIT_IN_PATH=$(git log -1 --format=format:%H --full-diff "$@")
```

https://www.gnu.org/software/bash/manual/html_node/Special-Parameters.html

hack/travis/travis-check-files-changed.sh
```diff
@@ -0,0 +1,39 @@
...
+#        - SEARCH_DIRECTORIES="site"
+#      before_install:
+#        - chmod +x ./hack/travis/travis-check-files-changed.sh
+#      install:
+#        - ./hack/travis/travis-check-files-changed.sh $SEARCH_DIRECTORIES ; RETURN_CODE=$? ; if [ $RETURN_CODE -eq 137 ]; then travis_terminate 0; elif [ $RETURN_CODE -ne 0 ]; then travis_terminate $RETURN_CODE; fi
+
+
+# 1. Get all the arguments of the script
+# https://unix.stackexchange.com/a/197794
+PATHS_TO_SEARCH="$*"
```
jpeach: Can noted below, I think you want to remove this and just use `"$@"`.

hack/travis/travis-check-files-changed.sh
jpeach: and a `set -e # halt script on error`

jpeach:
@davecheney FYI, I'm planning to work with @SDBrett to land this today. If you have concerns, ping me.

davecheney:
I would prefer we don't get into this business. Github covers this sufficiently.

jpeach:
> On Nov 29, 2019, at 8:37 AM, Dave Cheney <notifications@github.com> wrote:
> 
> ﻿I would prefer we don't get into this business. Github covers this sufficiently.

Does GitHub cover link checking with HTML proofer? Where’s that set up?

davecheney:
I’m sorry I think I misentetpreted this change from the title. If the gen scanning stuff is removed, great. 

What is the cost to ci time from link checking? Do we need to run it on every commit? Can it be done on the netlify side? Can it be done in batch?

jpeach:
> On 29 Nov 2019, at 9:09 am, Dave Cheney <notifications@github.com> wrote:
> 
> I’m sorry I think I misentetpreted this change from the title. If the gen scanning stuff is removed, great. 
> 
> What is the cost to ci time from link checking? Do we need to run it on every commit? Can it be done on the netlify side? Can it be done in batch?

* The time cost is approximately that of `make site-check`.
* The "travis-check-files-changed.sh" script is trying to ensure that this check only runs when website files change, so the cost to the majority of PRs should be zero.
* I don't know whether netlify does this or not.
* Not sure what you mean by doing it in batch?

davecheney:
On Fri, 29 Nov 2019 at 09:49, James Peach <notifications@github.com> wrote:

>
>
> > On 29 Nov 2019, at 9:09 am, Dave Cheney <notifications@github.com>
> wrote:
> >
> > I’m sorry I think I misentetpreted this change from the title. If the
> gen scanning stuff is removed, great.
> >
> > What is the cost to ci time from link checking? Do we need to run it on
> every commit? Can it be done on the netlify side? Can it be done in batch?
>
> * The time cost is approximately that of `make site-check`.
> * The "travis-check-files-changed.sh" script is trying to ensure that this
> check only runs when website files change, so the cost to the majority of
> PRs should be zero.
> * I don't know whether netlify does this or not.
> * Not sure what you mean by doing it in batch?


As in, doing this outside the CI loop.

PR #1849, opened 2019-11-04
Add namespace env var to certgen job

dhxgit:
This makes it easier to deploy examples/render/contour.yaml to a different namespace, via kustomization.yaml for example.
Without this change contour certgen will try to create a secret in the namespace projectcontour, which will not be accessable(, iff deoployed to different namespace).

The contour bootstrap initcontainer also has this namespace env, there should be no reason for certgen to not have it. :)


dhxgit:
> One thing, would you mind rebasing your changes into a single commit?

Sure thing :)

Issue #1858, opened 2019-11-05
docs: update deploy-options.md for kind (redux)


ghchinoy:
The section in /deploy-options (/site/_guides/deploy-options.md) for kind is a little different now and I'd like to add a few things to make it clearer

* change /site/_guides/deploy-options.md and /examples/kind-expose-port.yaml to reference exposing port 80 (rather than 8080), and adding 443
* add a sentence or two on how to add Contour to kind (via `kubectl apply -f examples/contour`)
* add a new file to /site/examples called kuard-httpproxy.yaml and reference this file from the /site/_guides/deploy-options.md (as https://projectcontour.io/examples/kuard-httpproxy.yaml doesn't exist)

ghchinoy:
This suggestion is to align the various ways of showing the use of Kind + Contour.
There's a great post by Steve here ([Kindly running Contour](https://projectcontour.io/kindly-running-contour/)) shows using a config yaml for Kind which exposes 80/443 (rather than 8080 / 8443). I don't want to ascribe intent, but I think having 80/443 make a quick local usage easier to use. A subsequent (also great) [blogpost on HTTPProxy](https://projectcontour.io/httpproxy-in-action/) is based upon this configuration.

Both the example kind config yaml in the repo https://github.com/projectcontour/contour/tree/master/examples/kind and the site's [Deployment Options](https://projectcontour.io/docs/v1.0.0/deploy-options/) document ( git [deploy_options.md](https://github.com/projectcontour/contour/blob/master/site/docs/v1.0.0/deploy-options.md)) show using 8080 for a Kind config. Unfortunately, the examples/walkthrough in Deployment Options for Kind is somewhat incomplete. 

1.  it's unclear how to install Contour from this page, 
2. in the HTTPProxy section, there's a 404 to https://projectcontour.io/examples/kuard-httpproxy.yaml 

For 1,  I'd suggest reiterating something like `kubectl apply -f examples/contour`
For 2, II'd suggest clarifying this by adding an HTTPProxy kuard example file and also the instructions to `kubectl apply -f site/examples/kuard-httproxy.yaml` (I'm not sure where this file should go, /site/examples looks good, but there's also an /examples)

Lastly, I'm not clear how the url forms of the files (a la https://projectcontour.io/examples/kuard-httpproxy.yaml ) get referenced, so the previous suggestions require cloning the repo to apply. I do think it's easier for users to have a publicly accessible yaml rather than having to clone the whole repo, but that's an aside.

[...]

PR #1859, opened 2019-11-05
Updates to deploy options for kind

ghchinoy:
* Modifies kind config to expose 80 (instead of 8080) and 443
* adds in a kuard-httpproxy.yaml for an example in sites/example hierarchy
* adds in some clarity to the "Test with HTTPProxy" section to reference the new kuard-httpproxy.yaml

reference #1858


stevesloka:
Hey @ghchinoy could you fix the build error? 

```
error    no new line character at the end of file  (new-line-at-end-of-file)
```

stevesloka:
Sorry, @ghchinoy I merged a big docs update and now there's a conflict. Would you mind one last pass and fixing that as well as merging your commits into a single one? Thanks!

site/examples/kuard-httpproxy.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: apps/v1
```
davecheney: Where is this file referenced from?

examples/kind/kind-expose-port.yaml
```diff
@@ -4,6 +4,10 @@ nodes:
 - role: control-plane
 - role: worker
   extraPortMappings:
-  - containerPort: 8080
-    hostPort: 8080
+  - containerPort: 80
```
davecheney: Were the original 8080 ports incorrect? As far as I understand it Contour configures the Envoy containers to listen on 8080 and 8443.

ghchinoy: No, the 8080 ports worked ok; I'd changed to 80 to align with blog post examples and other documentations.
Would be happy to change back to 8080 and add 8443 and redo docs for consistency.

site/examples/kuard-httpproxy.yaml
ghchinoy: Should be site/docs/master/deploy-options.md ... which appears to be missing from the latest squash. Oops (will correct).

davecheney: Ping. Would you please rebase and squash to one commit then reply so we know to review again. Thanks Dave

davecheney:
Thank you for working on this and I want to apologise for not having the full context here. 

Fundamentally the problem I have with this change is we spent a lot of time reducing the number of examples we include with the project to one. I’m concerned that this change increases the number of different deployment examples that we include with the product. 

Would you be able to step back a little and outline the problems with the current documentation — accepting of course that there are simple things like missing liked files. 

Ideally I’d like to resolve #1858 without adding a second deployment example.

ghchinoy:
@davecheney  this makes a lot of sense. I suggest that this PR be closed and a short discussion either in channel or on #1858 occur to clarify and align blog posts and in-repo examples.

Issue #1286, opened 2019-08-01
Contour should switch to using ADS for all configuration streams


davecheney:
This is a long term issue to track potentially moving from RDS/EDS/CDS/etc to a unified ADS implementation. 

**Note**: This is not currently in scope for Contour 1.0.

jpeach:
Related to #1178

Issue #1178, opened 2019-06-19
xDS updates are not ordered


phylake:
I don't see any evidence in Contour's source that xDS updates are ordered correctly

https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#eventual-consistency-considerations

Without ordering the update order would be subject to whichever goroutine was scheduled first. This race condition matches the behavior we see: a Contour restart doesn't *always* cause Envoy 404s.

This is an issue for us because we can't restart or upgrade Contour without causing downtime. Clients see 404s for the period of time after the `ingress_http` and `ingress_https` listeners are torn down, and before it takes the restarted Contour to respond to all the DiscoveryRequests.

We're running a split-pod Deployment. Did you test restarting Contour when it's not in the same pod as Envoy? If so, is this expected behavior?

davecheney:
Thank you for raising this issue. I'm sorry this is a problem for you. We are working to make the split envoy and contour deployment the default by the 0.15 milestone #881 .

To implement ordering we would probably have to switch to ADS. I know you've asked for this before (sorry I cannot find the issue at the moment). My answer is the same as previously, implementing ADS is not on the current roadmap.

Can you please tell me more about the sequence of events you're seeing. When you say `ingress_http` and `ingress_https` are torn down, this is not something I would expect to happen unless all ingress documents were removed.

davecheney:
For the five 👍 's I encourage you to join the next community meeting

https://projectcontour.io/community/

So we can talk about the issue in a higher bandwidth form.

I'm not saying no to ADS, I just want to understand the issues you're having. If that means I need to rescope ADS and put it on the road map, so be it, but I need your help to understand the issues you are experiencing.

phylake:
From the Envoy docs:
> It’s challenging to provide the above guarantees on sequencing to avoid traffic drop when management servers are distributed

Challenging but not impossible. Also, we're running a single Contour because of the various other protocol issues. It's handling 100+ Envoys and that could be your recommended split-pod Deployment (i.e. `replicas: 1`). The above comment is specific to distributed servers (plural), implying a singleton management server wouldn't face the same challenges.

> To implement ordering we would probably have to switch to ADS

That would probably be the best approach but some coordinated locking on the individual streams could work too. From what I see in logs LDS/CDS/RDS have no resource hints so you can count on only 1 `DiscoveryRequest` of each. EDS is variable but known by Contour so it seems feasible.

[...]

PR #1864, opened 2019-11-06
internal/envoy: enable Envoy ADS support

jpeach:
Since the Contour GRPC server already multiplexes Envoy requests
using the resource type URL, we can support ADS simply by enabling
it in the bootstrap configuration.

### DIFF
--- a/cmd/contour/bootstrap.go
+++ b/cmd/contour/bootstrap.go
@@ -37,6 +37,7 @@ func registerBootstrap(app *kingpin.Application) (*kingpin.CmdClause, *bootstrap
 	bootstrap.Flag("envoy-cert-file", "gRPC Client cert filename for Envoy to load.").Envar("ENVOY_CERT_FILE").StringVar(&ctx.config.GrpcClientCert)
 	bootstrap.Flag("envoy-key-file", "gRPC Client key filename for Envoy to load.").Envar("ENVOY_KEY_FILE").StringVar(&ctx.config.GrpcClientKey)
 	bootstrap.Flag("namespace", "The namespace the Envoy container will run in.").Envar("CONTOUR_NAMESPACE").Default("projectcontour").StringVar(&ctx.config.Namespace)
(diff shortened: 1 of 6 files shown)


cmd/contour/bootstrap.go
```diff
@@ -37,6 +37,7 @@ func registerBootstrap(app *kingpin.Application) (*kingpin.CmdClause, *bootstrap
 	bootstrap.Flag("envoy-cert-file", "gRPC Client cert filename for Envoy to load").Envar("ENVOY_CERT_FILE").StringVar(&ctx.config.GrpcClientCert)
 	bootstrap.Flag("envoy-key-file", "gRPC Client key filename for Envoy to load").Envar("ENVOY_KEY_FILE").StringVar(&ctx.config.GrpcClientKey)
 	bootstrap.Flag("namespace", "The namespace the Envoy container will run in").Envar("CONTOUR_NAMESPACE").Default("projectcontour").StringVar(&ctx.config.Namespace)
+	bootstrap.Flag("enable-ads", "Enable the Envoy Aggregated Discovery Service").BoolVar(&ctx.config.EnableADS)
```
stevesloka: Is there a reason to not just turn this on and not allow for a flag?

jpeach: Not especially, just risk mitigation. I'd be OK with removing it, or flipping the default. I'd slightly prefer the latter, since users could flip ADS off if something horrible happens :)

jpeach: Flipped this so it is enabled by default.

jpeach:
@davecheney Can you please take a look?

davecheney:
Thanks for working on this. I’d prefer to look at this after 1.1 is out.

jpeach:
Related #1178 and #1286.

jpeach:
Updated with a nonce de-synchronization fix.

jpeach:
This needs a lot more work, based on my improved understanding of ADS.

Issue #951, opened 2019-03-20
Contour should emit Certificate objects for HTTPProxy records


davecheney:
Currently cert-manager integration with Contour requires either manually creating a Certificate object (a CRD describing a Let's Encrypt managed certificate) or using [annotations][1] on an Ingress resource which triggers cert-manager's ingress-shim helper.

Rather than ingress-shim learning about HTTPProxy and staying in sync as we evolve the type, Contour should learn how to emit a Certificate object -- effectively contour will grow its own version of ingress shim for our HTTPProxy objects. 

1: https://github.com/jetstack/cert-manager/blob/master/docs/tasks/issuing-certificates/ingress-shim.rst

bgagnon:
I like this idea!

We’ve built a custom controller that watches IngressRoute objects and automatically requests a matching certificate from our PKI CertSrv server (it’s a Windows server thing).

We’ve first considered integrating with cert-manager by adding PKI CertSrv support to it and creating Certificate objects as proposed here. We punted on this design due to the number of moving pieces, but if Contour can do its part, it would feasible to switch.

On a related note, we do the same to manage DNS records for IngressRoute objects via that same controller.

Perhaps Contour could also have the option to emit DNSRecord objects for integration with external-dns.

davecheney:
Moving to the unplanned milestone to review after Contour 1.0

jpeach:
xref #2187

Given the wide variety of ways to configure and issue TLS certificates, integrating this into Contour is a fraught endeavor. Maybe a collaborating operator, but simply adding the support into cert-manager seems a lot more tractable to me.

ymmt2005:
@jpeach We develop a sidecar controller to create Certificate along with HTTPProxy.
If interested, see https://github.com/cybozu-go/contour-plus

alexgleason:
Hi all, I'm working on a hosting service with Contour. Users will bring their own domain names (in the 100s and perhaps 1000s), so it's important that I can automatically provision SSL certificates. Is the current best way to roll my own solution? Thanks. :slightly_smiling_face:

erwbgy:
[...]

Issue #950, opened 2019-03-20
Contour should support cert-managers's webhook solver


davecheney:
cert-manager is changing how certificate challenges are implemented. These are called "solvers" and represent ways of altering Ingress or DNS resources to complete the HTTP-01 or DNS-01 Let's Encrypt Challenge. The mechanism is describe in jetstack/cert-manager#1443.

Broadly a solver implements a web hook administered by the k8s api server as an api extension. This webhook is called by cert-manager to adjust the environment to support Let's Encrypt's challenge. 

Contour should grow support for supporting this webhook to allow cert-manager to adjust IngressRoute managed virtualhosts without the grottiness of an amalgam of Ingress and IngressRoute resources to complete the HTTP-01 challenge.

davecheney:
Its unlikely this will be started before the 0.15 milestone closes on the 23rd. Moving to beta.1

davecheney:
Moving to the backlog due to a lack of engineering resources and questions about the operational issues of api hooks.

PR #1867, opened 2019-11-07
Update cert-manager tutorial with HTTPProxy details

### LINKED ISSUE #1140: Add docs on how to use CertManager with HTTPProxy
Docs exist today on how to setup CertManager to create Let's Encrypt certificates with Kubernetes Ingress resources. It would be great to also demonstrate how to accomplish this with IngressRoute objects. 

We cannot use annotations with this style, however, we can create Cert Manager "Certificate" CRD's to also let CertManager help with the certificate request flow.

### LINKED ISSUE #1570: cert-manager tutorial needs updating
The cert-manager tutorial needs a solid update:

- the version of cert-manager is very old and has an issue where it will generate excessive traffic to Let's Encrypt.
- the tutorial uses `j.hept.io` which will break soon if it hasn't already.
- The docs will also need updating after #950, #951, and #1140 are done. This is just to make sure that we don't miss these bits.

youngnick:
Fixes #1570
Fixes #1140
Updates #950
Updates #951


site/_guides/cert-manager.md
```diff
@@ -120,8 +118,9 @@ There are plenty of other ways to deploy cert-manager, but they are out of scope
 
 To keep things simple, we skip cert-manager's Helm installation, and use the supplied YAML manifests:
 
-```
-$ kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v0.8.0/cert-manager.yaml
+```sh
+$ kubectl create namespace cert-manager
+$ kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v0.11.0/cert-manager.yaml
```
davecheney: Does this command create the required namespace? It doesn't include a -n cert-manager flag, so it looks like the previous command isn't needed.

site/_guides/cert-manager.md
```diff
@@ -164,10 +163,13 @@ metadata:
 spec:
   acme:
     email: user@example.com
-    http01: {}
     privateKeySecretRef:
-      name: letsencrypt-staging
+      name: letsencrypt-prod
```
davecheney: Is this change intention? The context of this file is the staging let's encrypt server

site/_guides/cert-manager.md
```diff
@@ -363,27 +368,36 @@ spec:
 The `kubernetes.io/tls-acme: "true"` annotation tells cert-manager to use the `letsencrypt-staging` cluster-wide issuer we created earlier to request a certificate from Let's Encrypt's staging servers.
 
 The certificate is issued in the name of the hosts listed in the `tls:` section, `httpbin.davecheney.com` and stored in the secret `httpbin`.
-Behind the scenes, cert-manager creates a certificate CRD to manage the lifecycle of the certificate.
+Behind the scenes, cert-manager creates a certificate CRD to manage the lifecycle of the certificate, and then a series of other CRDs to handle the challenge process.
+The process goes `certificate` -> `certificaterequest` -> `order` -> `challenge`. If something goes wrong with this process, you will need to `kubectl get` then `kubectl describe` each of these objects in order to get the full picture.
```
davecheney: This background is useful, but incomplete. How would you feel about dropping this editorial?

site/_guides/cert-manager.md
```diff
@@ -418,19 +432,23 @@ To request a properly signed certificate from the Let's Encrypt production serve
...
 spec:
   acme:
     email: user@example.com
-    http01: {}
     privateKeySecretRef:
       name: letsencrypt-prod
     server: https://acme-v02.api.letsencrypt.org/directory
+    solvers:
+    - http01:
+        ingress:
```
davecheney: Is the ingress class needed? What does it default to?

site/_guides/cert-manager.md
```diff
@@ -487,11 +505,108 @@ $ curl https://httpbin.davecheney.com/get
...
+spec:
+  virtualhost:
+    fqdn: httpbinproxy.davecheney.com
+    tls:
+      secretName: httpbin
+  routes:
+  - services:
+    - name: httpbin
+      port: 8080
+
```
davecheney: s/blank line

site/_guides/cert-manager.md
```diff
@@ -487,11 +505,108 @@ $ curl https://httpbin.davecheney.com/get
...
+Once that's done, create the dummy Ingress object:
+
+```yaml
+apiVersion: networking.k8s.io/v1beta1
+kind: Ingress
+metadata:
+  annotations:
+    cert-manager.io/cluster-issuer: letsencrypt-prod
+    ingress.kubernetes.io/force-ssl-redirect: "true"
+    kubernetes.io/ingress.allow-http: "false"
```
davecheney: I think ingress.allow-http defaults to false, we shouldn't need this annotation.

site/_guides/cert-manager.md
```diff
@@ -487,11 +505,108 @@ $ curl https://httpbin.davecheney.com/get
...
+    kubernetes.io/ingress.class: contour
```
davecheney: why is this needed?

site/_guides/cert-manager.md
```diff
@@ -487,11 +505,108 @@ $ curl https://httpbin.davecheney.com/get
...
+      - backend:
+          serviceName: httpbin
+          servicePort: 8080
+  tls:
+  - hosts:
+    - httpbinproxy.davecheney.com
+    secretName: httpbinproxy
+```
+
+Once cert-manager has done its thing, you will have a `kubernetes.io/tls` Secret present, that will contain the keypair.
```
davecheney: ... you will have a `httpbinproxy` secret.

site/_guides/cert-manager.md
```diff
@@ -487,11 +505,108 @@ $ curl https://httpbin.davecheney.com/get
...
+  "url": "https://httpbinproxy.davecheney.com/get"
+}
+```
+
+### Caveats
+
+This method is a workaround until we can deliver the changes in [#950](https://github.com/projectcontour/contour/issues/950) and [#951](https://github.com/projectcontour/contour/issues/951).
+As such, it has a few caveats.
+
+Firstly, it works because Contour will override matching FQDN/path combinations in the following order: HTTPProxy overrides IngressRoute overrides Ingress.
```
davecheney: I don't think this is correct. You can overlay httpproxy on top of n Ingress records with the same fqdn. You cannot overlay httpproxy on top of IngressRoute to the best of my understanding.

site/_guides/cert-manager.md
```diff
@@ -487,11 +505,108 @@ $ curl https://httpbin.davecheney.com/get
...
+The solver Ingress created by cert-manager has a unique FQDN/path combination so it will work, but the base one will be overridden by the HTTPProxy config.
```
davecheney: I don't think this is accurate. It might be simpler to delete this line.

site/_guides/cert-manager.md
```diff
@@ -487,11 +505,108 @@ $ curl https://httpbin.davecheney.com/get
...
+
+Secondly, the actual config of where the object sends traffic is irrelevant, *as long as it matches a route in the HTTPProxy*.
```
davecheney: Would it be easier to say -- the dummy ingress record exists only to hold the annotations ingress-shim requires.

site/_guides/cert-manager.md
```diff
@@ -487,11 +505,108 @@ $ curl https://httpbin.davecheney.com/get
...
+In this example, we're not using any advanced routing features, so this is straightforward.
```
davecheney: I don't think you need to document something that we're not doing. maybe just delete this line.

site/_guides/cert-manager.md
```diff
@@ -487,11 +505,108 @@ $ curl https://httpbin.davecheney.com/get
...
+However, if you are, you'll either need to match routes between Ingress and HTTPProxy or create a dummy path for the Ingress to match on.
```
davecheney: I think this could be said simpler by saying the ingress record exists for ingress-shim only, mixing routes across ingress and httpproxy is outside the scope of this document.

site/_guides/cert-manager.md
```diff
@@ -487,11 +505,108 @@ $ curl https://httpbin.davecheney.com/get
...
+
 ## Wrapping up
 
 Now that you've deployed your first HTTPS site using Contour and Let's Encrypt, deploying additional TLS enabled services is much simpler.
-Remember that for each HTTPS website you deploy, you create a Certificate CRD that provides the domain name and the name of the target Secret.
-After the Secret is created, you add a `tls` field to the Ingress object for your site.
+Remember that for each HTTPS website you deploy, cert-manager will  create a Certificate CRD that provides the domain name and the name of the target Secret.
+After the Secret is created, you add a `tls` field to the HTTPProxy or Ingress object for your site.
```
davecheney: The order of operations isn't this strict, contour is just not going to be able to serve the https version of the site until the secret is ready.

site/_guides/cert-manager.md
jpeach: Note the need to pass `--validate=false` here for Kubernetes < 1.15

site/_guides/cert-manager.md
```diff
@@ -487,11 +505,108 @@ $ curl https://httpbin.davecheney.com/get
...
+    - name: httpbin
+      port: 8080
+
+```
+
+This config will be marked as Invalid by Contour, since the TLS secret doesn't exist yet.
+Once that's done, create the dummy Ingress object:
+
+```yaml
+apiVersion: networking.k8s.io/v1beta1
```
jpeach: Earlier, the YAML uses `extensions/v1beta1` for Ingress. We should be consistent across the whole page and note the k8s version that require `extensions/`.

site/_guides/cert-manager.md
```diff
@@ -103,7 +101,7 @@ kuard-67ff6dd458-sfxkb     1/1       Running   0          19d
 
 Then type the DNS name you set up in the previous step into a web browser, for example `http://gke.davecheney.com/`. You should see something like:
 
-![KAURD screenshot][kuard]
+![KUARD screenshot][kuard]
```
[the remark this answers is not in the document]
youngnick: I guess, sure.

site/_guides/cert-manager.md
youngnick: I'm okay with it. I'll remove.

site/_guides/cert-manager.md
youngnick: There is no default. You must specify this for cert-manager to know the ingress class to put on the Ingress objects it creates. This is new in 0.11.

site/_guides/cert-manager.md
youngnick: Good point, will do.

site/_guides/cert-manager.md
youngnick: hmm, good point. I'll check if it is needed.

site/_guides/cert-manager.md
youngnick: Yep, that's better, thanks.

site/_guides/cert-manager.md
youngnick: Okay, I'll write it differently, thanks.

site/_guides/cert-manager.md
youngnick: Yeah, I suspect this is to allow the management of the `kubectl delete` flow.

Not deleting the namespace as part of the YAML means that you can `kubectl delete -f <file>` and you won't get stuck with the namespace not terminating because of the ValidatingWebhook config.

site/_guides/cert-manager.md
youngnick: Nope, will revert.

site/_guides/cert-manager.md
youngnick: It is not needed, which simplifies this quite a bit.

youngnick:
This is ready for rereview.

site/_guides/cert-manager.md
```diff
@@ -85,66 +83,73 @@ In your own data center, you need to arrange for traffic from a public IP addres
...
 ```bash
-$ kubectl delete -f {{ site.url }}/examples/kuard.yaml
+$ kubectl delete -f {{ site.url }}/examples/httpbin.yaml
 ```
 
 ## 2. Deploy jetstack/cert-manager
 
 **NOTE:** cert-manager is a powerful product that provides more functionality than this tutorial demonstrates.
-There are plenty of other ways to deploy cert-manager, but they are out of scope.
+There are plenty of [other ways to deploy cert-manager](4), but they are out of scope.
```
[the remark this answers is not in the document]
youngnick: 🤦‍♂ thanks

PR #1868, opened 2019-11-07
Add Status design update doc

youngnick:
Sorry for the length on this one, it's pretty bad.

There's likely to be mistakes in the YAML, it's all done by hand. TBH, I probably should have used @stevesloka's YAML generator and changed some specs, then generated.

As I've gone through and updated all the examples, I think that having all four fields: `includes`, `routes`, `allConditions` and `proxyGraph` is probably overkill in the current design. But, I'd like you all to take a look and think about the really relevant information for each.

There's still some TODO examples, but I wanted to let you all get started on reading this, so I can come back to it later.

### DIFF
--- a/design/status-design-update.md
+++ b/design/status-design-update.md
@@ -1,0 +1,1166 @@
+# HTTPProxy status redesign
+
+## Background
+
+Given its complexity, Contour's HTTPProxy CRD currently has insufficient information about what its status surfaced to the user.
+
+Currently, it only supports a `currentStatus` and `description`, which are particularly poor user experience around inclusion and conditions.
+
+In particular, the following questions about HTTPProxy are particularly problematic at the moment:
+
+- What happens when *some* parts of a HTTPProxy configuration are invalid? What if one include is invalid, and the other is not? How can the user be aware of what's happening?
+- What happens if there are problems on inclusion, particularly condition conflicts?
+- What happens if there is a problem with composable HTTPProxy use, that is, including the same HTTPProxy object from multiple parents.
+- How do users understanding the full tree of includes used to build an Envoy config from multiple HTTPProxy objects?
+
+## Design Goals
+
+### Goals Overview
+
+In rethinking Status, we need to ensure that the whatever we choose can fit all the use cases our HTTPProxy design allows for.
+
+In particular, the features that strongly affect how we design status:
+
+- Inclusion: We need to have a clear set of rules about what classes of errors show up at the *includer* level and what at the *includee* level.
+  - We also don't put limits on how big this structure should be, so we need to ensure that the status block can provide useful information for large levels of complexity.
+- Composability: Because HTTPProxies can be included under more than one root, and can also be included in multiple places under a single root, Status needs to be able to work in both those cases.
+
+We also need to consider the interaction between service admins when one side changes something that breaks something for the other side.
+
+In addition, we need to keep some rules in mind about making the status useful:
+
+- the status must be able to indicate multiple errors, not just one.
+- errors must be able to be directly attributed to the part of the config that produced them.
+  - To help with this, the status should have a way to show errors with routes and includes separately.
+  - Most portions of the status must be namespaced by at least the fqdn of the root, if not some rendering of the include path.
+- the status must be able to indicate problems that will stop the HTTPProxy working as expected, but that are not syntactic errors with the HTTPProxy construct. For example, the referenced service is invalid.
+- the status should have a way to see the object as rendered by contour (that is, with merged conditions). This allows users to inspect what output Contour produces for a given input without using debug inputs or inspecting Envoy.
+
+It's also worth noting that `status` does not have to *only* be status.
+The `status` field in a Kubernetes object is generally used for any *computed information* (as opposed to the declarative configuration in `spec`)
+
+Lastly, it's important to note that, now that HTTPProxy is GA, we can't change the existing status fields very much, if at all.
+
+### Goals
+
+Overall:
+
+- Provide a better user experience for understanding HTTPProxy usage.
+
+Required:
+
+- Detail expected personas for users
+- Provide a mechanism for detailed information about invalid inclusions, including specifying what is added to each object, includer and included.
+- Provide a mechanism for detailed information about invalid routes/services.
+- Provide a mechanism for HTTPProxy administrators to see what conditions are added at each level of inclusion.
+- Provide the information required for higher-level parsing of the status (to enable both kubectl and other UI inspection).
+- An external tool must be able to build a representation of the HTTPProxy include DAG without speaking using only the Kubernetes API.
+- Provide a set of examples that can be used both for understanding this proposal and as the basis for a test suite for status.
+
+Optional:
+
+- Add usable printer columns for kubectl use (Fixes #1567)
+
+### Non-Goals
+
+- Considering how this will be implemented at anthing but the most basic level.
+
+This is a high-level design effort, intended to think about what the ideal case would be.
+Let's understand the dream first, then work back from there.
+
+## Users/Personas
+
+### Proxy Administrator
+
+This is the person or team who controls the TLD and any associated TLS information.
+They are responsible for root HTTPProxy objects, and specifying conditions on the includes for the root HTTPProy objects.
+
+### Service Administrator
+
+This is the person or team who controls a HTTPProxy that is included into another.
+Similarly to HTTPProxy inclusion, there may be an arbitrarily long chain of Service Administrators before an actual service is reached.
+
+### Upstream tool user
+
+This is the person or team who wants to be able to visualize the overall set of HTTPProxy objects, how they interact, and get a picture of what the eventual Envoy config will look like.
+This user only wants to use the Kubernetes API to retrieve this information.
+
+Note that Proxy admins and Service admins may also be Upstream tool users.
+
+## Overall Design
+
+The overall design I have in mind is to add some further fields to the status object in each HTTPProxy, and allow for more information about the in-between states, other than 'everything is okay' and 'nothing is okay'.
+
+Each one of these fields is intended to include enough information to enable either troubleshooting of specific errors, to expose information for higher-level tooling, or both.
+
+### Extra fields
+
+There are three extra fields I believe we must add, and one I believe we should.
+The last is only optional only because I think that we could approximate its functions using the other fields.
+I think that the design is cleaner with it included, however.
+
+For each place where we add a status, I think we should keep the current naming scheme:
+
+- `currentStatus` is exactly what it says - the current status. This should have three values, `Valid`, `Invalid`, and `Warning` (or similar).
+- `description` performs the same function as the global `description` does right now, just closer to the source of the problem.
+
+So, a generic `status` stanza for each type of thing below would look like:
+
+```yaml
+status:
+  currentStatus: <Valid|Invalid|Warning>
+  description: <text>
+```
+
+#### `status.includes`
+
+This field should show the status of each include in the `spec.includes` array.
+In order to fully identify them, this means that this must include all the information in each `spec.includes` item, plus the general `status` stanza.
+
+That is, this field should include status of the includes *in this HTTPPRoxy*.
+
+See the examples below to see this in action.
+
+#### `status.routes`
+
+Similarly to `status.includes`, this should include the information from each `route` stanza in `spec.routes`, with an added `status` stanza.
+
+This field should include status of the routes in this HTTPProxy.
+
+#### `status.appliedConditions`
+
+This is where this gets trickier.
+This field is intended to show what conditions are applied when this HTTPProxy object is included from somewhere.
+
+In order to do this, the following information must be present, somehow:
+
+- full list of merged conditions. That is, all prefixes  must be fully merged, and the full set of headers must be visible.
+- a unique identifier for the condition set. In practice, this will end up being related to the inclusion path to this HTTPProxy.
+
+The first is relatively straightforward, a `conditions` array, with the merge rules having been run on it, will suffice.
+
+The second is important for higher-order tool use as well as service admin use.
+
+The exact implementation here will depend on if we do the optional field I include below.
+
+In any case, there should be an `includePath` field in the `status.appliedConditions[n]` object that ensures uniqueness.
+
+This could be rendered as a string, representing the include path, preferably including the FQDN of the root.
+
+So, something that looked like this:
+
+```yaml
+status:
+  appliedConditions:
+    - conditions:
+      - prefix: /foo/bar/baz
+      includePath: "root:www.foobar.com,foo,bar"
+```
+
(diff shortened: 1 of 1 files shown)


design/status-design-update.md
```diff
@@ -0,0 +1,1166 @@
...
+
+Given its complexity, Contour's HTTPProxy CRD currently has insufficient information about what its status surfaced to the user.
+
+Currently, it only supports a `currentStatus` and `description`, which are particularly poor user experience around inclusion and conditions.
+
+In particular, the following questions about HTTPProxy are particularly problematic at the moment:
+
+- What happens when *some* parts of a HTTPProxy configuration are invalid? What if one include is invalid, and the other is not? How can the user be aware of what's happening?
+- What happens if there are problems on inclusion, particularly condition conflicts?
+- What happens if there is a problem with composable HTTPProxy use, that is, including the same HTTPProxy object from multiple parents.
```
stevesloka: Today if a parent is `invalid` then the children become `orphaned` since they essentially have no one delegated. I bring this up because it sounds like that logic might be changing.

design/status-design-update.md
```diff
@@ -0,0 +1,1166 @@
...
+
+In rethinking Status, we need to ensure that the whatever we choose can fit all the use cases our HTTPProxy design allows for.
+
+In particular, the features that strongly affect how we design status:
+
+- Inclusion: We need to have a clear set of rules about what classes of errors show up at the *includer* level and what at the *includee* level.
+  - We also don't put limits on how big this structure should be, so we need to ensure that the status block can provide useful information for large levels of complexity.
+- Composability: Because HTTPProxies can be included under more than one root, and can also be included in multiple places under a single root, Status needs to be able to work in both those cases.
+
+We also need to consider the interaction between service admins when one side changes something that breaks something for the other side.
```
stevesloka: What does this mean? Can you clarify the example a bit?

design/status-design-update.md
```diff
@@ -0,0 +1,1166 @@
...
+
+In addition, we need to keep some rules in mind about making the status useful:
+
+- the status must be able to indicate multiple errors, not just one.
+- errors must be able to be directly attributed to the part of the config that produced them.
+  - To help with this, the status should have a way to show errors with routes and includes separately.
+  - Most portions of the status must be namespaced by at least the fqdn of the root, if not some rendering of the include path.
```
stevesloka: Namespaced as in Kubernetes namespaces? We need to be careful not to leak information to users that might not be privileged.

design/status-design-update.md
```diff
@@ -0,0 +1,1166 @@
...
+Overall:
+
+- Provide a better user experience for understanding HTTPProxy usage.
+
+Required:
+
+- Detail expected personas for users
+- Provide a mechanism for detailed information about invalid inclusions, including specifying what is added to each object, includer and included.
+- Provide a mechanism for detailed information about invalid routes/services.
+- Provide a mechanism for HTTPProxy administrators to see what conditions are added at each level of inclusion.
```
stevesloka: Can you expand this? An administrator should already have this by looking at the `Conditions` on an `Include`. I don't think that should be replicated back into status.

design/status-design-update.md
```diff
@@ -0,0 +1,1166 @@
...
+- Provide the information required for higher-level parsing of the status (to enable both kubectl and other UI inspection).
+- An external tool must be able to build a representation of the HTTPProxy include DAG without speaking using only the Kubernetes API.
```
stevesloka: I don't think we should have external tools come up with their own representation of the DAG, that should stay within Contour. Reason being, if folks build it differently, then things won't match up.

design/status-design-update.md
```diff
@@ -0,0 +1,1166 @@
...
+- Provide a set of examples that can be used both for understanding this proposal and as the basis for a test suite for status.
+
+Optional:
+
+- Add usable printer columns for kubectl use (Fixes #1567)
+
+### Non-Goals
+
+- Considering how this will be implemented at anthing but the most basic level.
```
stevesloka: It doesn't seem like a non-goal. I think the external tool building a DAG should be a non-goal.

design/status-design-update.md
```diff
@@ -0,0 +1,1166 @@
...
+
+### Extra fields
+
+There are three extra fields I believe we must add, and one I believe we should.
+The last is only optional only because I think that we could approximate its functions using the other fields.
+I think that the design is cleaner with it included, however.
+
+For each place where we add a status, I think we should keep the current naming scheme:
+
+- `currentStatus` is exactly what it says - the current status. This should have three values, `Valid`, `Invalid`, and `Warning` (or similar).
```
stevesloka: What about `orphaned` which is neither valid or invalid.

stevesloka: Do these high level items go against the previous comment about showing data from multiple roots? Meaning if one path is valid but another not?

design/status-design-update.md
```diff
@@ -0,0 +1,1166 @@
...
+            conditions:
+            - prefix: /bar
+            includes:
+              name: baz
+              namespace: default
+              conditions:
+              - prefix: /baz
+```
+
+#### Optional: `status.proxyGraph`
```
stevesloka: To me, this information isn't `Status`. Maybe something else, or I just haven't gotten my head around it enough.

design/status-design-update.md
```diff
@@ -0,0 +1,1166 @@
...
+  currentStatus: valid
+  description: valid HTTPProxy
+  includes: {}
+  routes:
+    - fqdn: bar.com
+      conditions:
+      - prefix: /
+      services:
+      - name: s2
+        port: 80
```
stevesloka: I know this was mentioned, but would be nice to have a way to represent this without duplicating over. I need to think about this some more to provide an alternative.

design/status-design-update.md
```diff
@@ -0,0 +1,1166 @@
...
+spec:
+  virtualhost:
+  includes:
+  - name: validChild
+    namespace: roots
+    conditions:
+      - prefix: /foo
+status:
+  currentStatus: invalid
+  description: Spec.VirtualHost.Fqdn must be specified
```
stevesloka: This can't happen, the only way to know what a "root" object is by the presence of the fqdn. If it doesn't have it, then it's child and would be orphaned.

jpeach:
I think that the way the `currentStatus` and `description` fields are used can be subsumed by a
[Status condition](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties).

Using the word "condition" here is a bit unfortunate, but the struct
and it's Type/Status/Reason/Message are very conventional.  We
should also follow the convention of having a slice of Conditions,
which will give is a lot of flexibility when the object is in
multiple current or pending states.

If we switch to a Condition slice, we need to consider the API
implications of dropping `currentStatus` and `description`.

Note that a Condition slice could be used to represent part of the
information you are exposing in the detailed status fields (e.g.
you could have a Condition for the state each child HTTPProxy that
a document includes, and a different Condition type with the vhost
that leads to a leaf document).

Another aspect of Conditions that occurs to me is that they strongly
imply a state machine backing the CRD. I don't have a strong sense
of what the state machine behind the HTTPProxy CRD is, but I think
it is richer than valid/invalid (e.g. a proxy can at
least be orphaned too).

In the `routes` and `proxyGraph` fields, what does the nested
`status` field refer to? I think that `status` is on a API document,
and in these cases it seems to be atached to something that doesn't
have a corresponding abstraction.

> ```
>    routes:
>    - fqdn: foo-basic.bar.com
>      conditions:
>      - prefix: /
>      services:
>        - name: s1
>          port: 80
>      status:
>        currentStatus: Valid
>        descriptions: Valid route
>      includePath: "basic:foo-basic.bar.com"
> ```

For example, this implies that a route is valid, but there is no
API object that is a route, just the HTTPProxy field. I'm not sure
that this helps to user focus on the problem, especially when this
is a rendered route that doesn't appear literally on any one
document

I'm concerned that the denormalized form is too unwieldy when you
have a lot of proxies. Maybe instead of denormalizing, we could
link back up the include path with a
[ObjectReference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#objectreference-v1-core).

Potentially, using ObjectReferences could allow standard tools to
traverse the inclusion chain (since it's used in Event and other
APIs).

What is the specification for the `includePath` string?

> - the status should have a way to see the object as rendered by
>   contour (that is, with merged conditions). This allows users to
>   inspect what output Contour produces for a given input without using
>   debug inputs or inspecting Envoy.

I think that for many users, inspecting Envoy configuration is reasonable and that's the real
source of truth when you want to know what configuration you
have.

I agree with the goal of summarizing a config somewhere along the
proxy chain.

design/status-design-update.md
youngnick: Ah, term overloading. No, not namespaced as in Kubernetes namespaces, just by the FQDN of the root HTTPProxy. I think it's okay to leak where your includers are coming from, since you should have RBAC stopping you from interacting with them.

design/status-design-update.md
youngnick: Sorry, I should have said 'all types of admins'. I meant that if you're controlling a included HTTPProxy, it is useful to see where each of the conditions came from (so you can talk to right person if there's more than one).

design/status-design-update.md
youngnick: This is not talking about the internal DAG, it's talking about the DAG produced by considering the HTTPProxy inclusions. They are similar but separate.

The DAG talked about here has HTTPProxy objects as nodes, and includes as edges.

design/status-design-update.md
youngnick: I meant of this initial proposal. Like I said just after, I'm trying to start from what users would want, and work from there, rather than starting with how it's implemented now and figuring out how to mutate it.

design/status-design-update.md
youngnick: Good point!

youngnick: That's a good question. I think that if one path is valid but another not, that's probably a case for using `Warning`.

design/status-design-update.md
youngnick: It's the status of the computed HTTPProxy tree.

Spec contains the declarative definition of what an object should look like, status contains anything computed, like Load Balancer IPs. In this case, the proxyGraph is computed by the superset of all HTTPProxy objects.

I'm not sure, to be honest, if we're better having this full-data construct, or being able to reconstruct it from the other three fields.

design/status-design-update.md
youngnick: Okay, thanks. I'll updated.

youngnick:
Summarising a Zoom discussion between James and I:
- I agree that we should look at using the Kubernetes type 'Condition' somehow, although the name collision is not great.
- I also agree that this requires substantial reworking before it can really be viable, there is both too much information and too much duplication. I'm going to leave this here while I think about it though.
- I also agree that using ObjectReferences everywhere we refer to another object is better than a more custom thing.

I'd still like it if @davecheney could take a look over this in its current form, with the fact it will end up changing quite a bit in mind.

davecheney:
Thank you for working on this. To clarify, status is not a priority for 1.1.

design/status-design-update.md
davecheney: What happens if a HTTPProxy root is included by another HTTPProxy root -- is it valid or invalid? The first path, root -> root, the parent is invalid, but is the child invalid because the parent made a boo boo?

design/status-design-update.md
```diff
@@ -0,0 +1,1166 @@
...
+For each place where we add a status, I think we should keep the current naming scheme:
+
+- `currentStatus` is exactly what it says - the current status. This should have three values, `Valid`, `Invalid`, and `Warning` (or similar).
+- `description` performs the same function as the global `description` does right now, just closer to the source of the problem.
+
+So, a generic `status` stanza for each type of thing below would look like:
+
+```yaml
+status:
+  currentStatus: <Valid|Invalid|Warning>
```
davecheney: What is the use case for this high level status? You've outlined that status is not simply a scalar property, its a product of the path of includes. Is it always going to be possible to collapse the various status paths down to a single value? I can't say for sure, but it feels likely that this isn't going to be possible 100% of the time, which means anything which relies on this summary information -- I'm thinking kubectl describe, is going to be shooting in the dark.

design/status-design-update.md
```diff
@@ -0,0 +1,1166 @@
...
+  description: <text>
```
davecheney: description is pretty useless now. It contains either the word "valid %type" or some description of the first error found processing the document -- where first is an under defined concept.

I imagine that any attempts to improve the fidelity of status are going to make it even harder to summarise than `currentStatus` above.

design/status-design-update.md
```diff
@@ -0,0 +1,1166 @@
...
+          includes:
+            name: bar
+            namespace: default
+            conditions:
+            - prefix: /bar
+            includes:
+              name: baz
+              namespace: default
+              conditions:
+              - prefix: /baz
```
davecheney: I think this is going to be the way we do it. Summarisation will lock us into a compatibility nightmare with tools that consume and attempt to rehydrate the summarised data. I think its better to be verbose and release that this information is for tools to consume.

design/status-design-update.md
davecheney: What's the use case for this information? I don't think it should be optional because we're either going to have to describe it fully in this design, then not implement it, or the opposite, underspecify it in this document and maybe implement it, sorta. I'd suggest either making it a concrete deliverable of this document or moving it out of scope.

design/status-design-update.md
```diff
@@ -0,0 +1,1166 @@
...
+```
+
+#### Optional: `status.proxyGraph`
+
+The intent of this field is to store enough breadcrumbs to allow a user (or external tool) to be able to deduce the entire HTTPProxy graph (if they have RBAC access to see the objects in all referenced namespaces).
```
davecheney: Can't we just embed that in status directly via the name/namespace of objects in the include path?

design/status-design-update.md
youngnick: In my mind, that would be a 'Warning' maybe? It's a good example though.

design/status-design-update.md
```diff
@@ -0,0 +1,1166 @@
+# HTTPProxy status redesign
+
+## Background
+
+Given its complexity, Contour's HTTPProxy CRD currently has insufficient information about what its status surfaced to the user.
+
+Currently, it only supports a `currentStatus` and `description`, which are particularly poor user experience around inclusion and conditions.
+
+In particular, the following questions about HTTPProxy are particularly problematic at the moment:
```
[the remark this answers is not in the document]
youngnick: sadokay.jpg

design/status-design-update.md
youngnick: I really only left the high level status in there because it's in the v1 spec, and so we can't take it out. I should have specified that more.

youngnick:
Closing this one as it's way out of date.

Issue #1873, opened 2019-11-08
Support prefix rewrite in HTTPProxies.


jpeach:
HTTPProxy does not have support for the Envoy prefix rewrite filter. Since that was supported in IngressRoute, we should bring it back.

jpeach:
Related to #899

Issue #69, opened 2017-11-28
Support basic host rewriting


davecheney:
Support, probably via an annotation, an Ingress object which exists to redirect www.example.com to example.com (or vice versa)

This might look like, in yaml
```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: foo
  annotations:
    kubernetes.io/ingress.host-rewrite "example.com"
spec:
  rules:
    - host: www.example.com
            backend:
              serviceName: foo // ignored
              servicePort: http // ignored
```

jevonearth:
FWIW; The ingress-nginx annotation `nginx.ingress.kubernetes.io/rewrite-target` might be a useful reference when implementing this.

davecheney:
This is quite complicated because the ingress object does not describe _a_ vhost, you can mention several in the same ingress object, so it can be unclean what the target of the rewrite is.

davecheney:
Moving to the unplanned milestone. We won’t be tackling this as part of Contour 1.0 or the next revision of ingressroute.

jpeach:
You can rewrite the Host header using [HeadersPolicy](https://projectcontour.io/docs/v1.9.0/api/#projectcontour.io/v1.HeadersPolicy).

If anyone need this for Ingress, please re-open and we can discuss support for that API specifically.

PR #1874, opened 2019-11-08
Initial design doc for HTTPProxy prefix rewrite.

jpeach:
This updates #1873.

### DIFF
--- a/design/httpproxy-prefix-rewrite.md
+++ b/design/httpproxy-prefix-rewrite.md
@@ -1,0 +1,380 @@
+# Request Rewrite For HTTPProxy
+
+**Status**: _Accepted_
+
+This document specifies a design for supporting URL prefix rewriting in the HTTPProxy CRD.
+
+## Goals
+
+Envoy currently supports [path prefix][2] and [host header][3] rewriting.
+It is likely that future Envoy releases will support [regex path](https://github.com/envoyproxy/envoy/issues/2092) rewriting.
+
+The goal of this proposal is to eventually allow the user to access Envoy URL rewriting features from the HTTPProxy CRD.
+The initial implementation target is path prefix rewriting, but the API should be specified such that it can be gracefully extended to other forms of request rewriting.
+
+## Non-Goals
+
+* Addressing rewrite limitations. Request rewriting in Envoy has
+  very limited capabilities. The [prefix_rewrite][4] action only
+  alters the prefix of a URL path. Since it does not alter response
+  headers or content, it is not possible to use this feature to
+  re-host a web application at a path prefix it is not expecting.
+
+* Controlling request routing. In some proxies (particularly NGINX),
+  rewriting the request URL can also restart the request routing
+  machinery. In this proposal, requests are always fully routed
+  before any changes are made to the URL (this is required by Envoy
+  semantics).
+
+## Background
+
+Request rewriting is ultimately specified as properties of the Envoy [RouteAction](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route.proto#envoy-api-msg-route-routeaction) object.
+The CRD semantics should avoid encouraging uses that can't be supported by Envoy.
+
+The Envoy `prefix_rewrite` configuration is coupled to prefix matching on the route.
+Each `prefix_rewrite` specification needs to be attached to a single prefix match, so each new `prefix_rewrite` needs a new prefix match.
+
+HTTPProxy documents are organized into a collection of trees, where the root is a document that has a `virtualhost` field, and the leaves are documents that have a `routes` field.
+A HTTPProxy tree could be a single document (that is both a root and a leaf), or it could be a series of documents in different namespaces, connected by the `includes` field.
+
+HTTPProxy inclusion semantics allow a single HTTPProxy CRD to be included by more than one root HTTPProxy document.
+This means that multiple routing path prefixes can end up referring to the same HTTPProxy.
+
+To reason about most kinds of request rewriting, the user needs to know the final form of the request.
+In other words, it doesn't make sense to want to rewrite a path prefix unless you have some expectations about the path prefix you will receive.
+This implies that even if the owner of a leaf HTTPProxy doesn't control the root or intermediate documents, they still have fairly concrete expectations around the full URL path to expect.
+
+Although there are a number of ways to conceptualize path prefix rewriting, path regex rewriting requires knowledge of the full request path at the time of writing the regex.
+This is because a regex rewrite can be used to re-order arbitratry portions of the URL path, and this can't be done unless the regex is defined as being applied to the complete path.
+
+## High Level Design
+
+Add a `PathRewrite` field to the HTTPProxy [Route](https://github.com/projectcontour/contour/blob/master/apis/projectcontour/v1/httpproxy.go#L125) object.
+This allows us to specify different kinds of rewrite for URL paths without also needing to specify a specific kind of request rewriting (i.e. we can add host rewrite later).
+
+Locating the `PathRewritePolicy` on the `Route` means that rewriting is controlled by the team that controls the leaf HTTPProxy document.
+This allows a useful initial implementation, but does not prevent adding `PathRewritePolicy` fields higher up the HTTPProxy tree (though allowing that would require detecting and resolving conflicts).
+
+```Go
+type Route struct
+	...
+
+	// The policy for rewriting the path of the request URL
+	// after the request has been routed to a Service.
+        //
+        // +kubebuilder:validation:Optional
+	PathRewrite *PathRewritePolicy `json:"pathRewrite,omitempty"`
+}
+
+```
+
+The `PathRewritePolicy` struct initially only supports path prefix replacement.
+Creating a policy struct allows future addition of other kinds of path rewriting.
+
+```Go
+// PathRewritePolicy specifies how a request URL path should be
+// rewritten. This rewriting takes place after a request is routed
+// and has no subsequent effects on the proxy's routing decision.
+// No HTTP headers or body content is rewritten.
+//
+// Exactly one field in this struct may be specified.
+type PathRewritePolicy struct {
+    ReplacePrefix []struct {
+        // Prefix specifies the URL path prefix to be replaced.
+        //
+	// If Prefix is specified, it must exactly match the Condition
+	// prefix that is rendered by the chain of including HTTPProxies
+	// and only that path prefix will be replaced by Replacement.
+	// This allows HTTPProxies that are included through multiple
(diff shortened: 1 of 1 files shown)


jpeach:
This is also related to #69.

design/httpproxy-prefix-rewrite.md
```diff
@@ -0,0 +1,262 @@
...
+Locating the `PathRewritePolicy` on the `Route` means that rewriting is controlled by the team that controls the leaf HTTPProxy document.
+This allows a useful initial implementation, but does not prevent adding `PathRewritePolicy` fields higher up the HTTPProxy tree (though allowing that would require detecting and resolving conflicts).
+
+```Go
+type Route struct 
+	...
+
+	// The policy for rewriting the path of the request URL
+	// after the request has been routed to a Service.
+	PathRewrite *PathRewritePolicy `json:"pathRewrite,omitempty"`
```
davecheney: For consistency with the other policies, what about calling this field `rewritePolicy`?

design/httpproxy-prefix-rewrite.md
```diff
@@ -0,0 +1,262 @@
...
+```Go
+// PathRewritePolicy specifies how a request URL path should be
+// rewritten. This rewriting takes place after a request is routed
+// and has no subsequent effects on the proxy's routing decision.
+// 
+// Exactly one field in this struct may be specified.
+type PathRewritePolicy struct {
+    ReplacePrefix []struct {
+        // +optional
+        Prefix string `json:"prefix,omitempty"`
```
davecheney: Could you add doc comments to these fields please

jpeach: For now, the fields are described in the following text. If we add this to the code, that text would move to docs comments.

design/httpproxy-prefix-rewrite.md
jpeach: If we call it rewrite policy, then that implies that it deals with all kinds of rewrites (e.g host rewrite). The way I'm thinking of dealing with that is a separate rewrite spec, so that's why I've named this to indicate we are specifically dealing with paths. We can add other rewrite kinds in sibling fields.

davecheney:
I’d like to collect all the *policy things together in a separate object as this leaves the ability to specify some policy objects at the routes level as a default for the routes that land under them. I fear if route grows boutique fields for each possible option that means we need to replicate those a level higher to implement the same UX.

jpeach:
> On 9 Nov 2019, at 1:28 am, Dave Cheney <notifications@github.com> wrote:
> 
> I’d like to collect all the *policy things together in a separate object as this leaves the ability to specify some policy objects at the routes level as a default for the routes that land under them. I fear if route grows boutique fields for each possible option that means we need to replicate those a level higher to implement the same UX. 

We we could alternatively do:

Route
    -> Rewrite
        -> PathRewrite
            -> PrefixReplace
            -> RegexReplace

or

Route
    -> Rewrite
        -> PathPrefixReplace
        -> PathRegexReplace

If we keep it as proposed, though, I don't see that it prevents us taking the first option in the future if we want to move rewrites up the tree. I am not in favor of the second flavor here, because I think that the Rewrite struct ends up with too much to do once you want to rewrite hosts, headers, schemes, etc.

jpeach:
> The reason we pulled prefixRewrite out of HTTPProxy was it didn't deal with the ambiguities around trailing slash. I would like to see this problem specifically addressed as part of this design.

IIUC the reason this happens is entirely because of the limited semantics of the Envoy prefix rewrite (it's just a dumb string substation). I think the [suggestion](https://github.com/projectcontour/contour/issues/899#issuecomment-519732842) from @stevesloka is the right way to try to address this, but we need to be a little careful since Envoy does not distinguish between the rewrite prefix and the routing prefix. It is possible to have "/foo" and "/foo/" as two separate routes. In this case, we must not generate the implicit routing prefix, and the user must specify two `PrefixReplace` entries with the separate prefixes.

design/httpproxy-prefix-rewrite.md
davecheney: I'd really appreciate it if you could summarise the use of those fields in this design document.

The reason I ask this is I want to see the use of `Prefix`. Its not clear to me what this field does as the matching of the containing route is not part of the `Prefix` field. Making this field optional makes me wonder under which circumstances it is needed. I think a doc comment right here in the code sample would help clarify these questions for me.

design/httpproxy-prefix-rewrite.md
```diff
@@ -0,0 +1,262 @@
...
+
+        Replacement string `json:"replacement"`
+    } `json:"replacePrefix,omitempty"`
+}
+```
+
+`PrefixRewrite` is a slice to allow a service to rewrite multiple path prefixes, depending on the shape of the HTTPProxy tree.
+As discussed above, a HTTPProxy leaf can be reached from multiple HTTPProxy roots via different paths.
+This implies that service owners could reasonably need to rewrite multiple routing paths to a canonical version.
+This, in turn, implies that more than one prefix may need to be rewritten.
```
davecheney: This is true, but I think this overcomplicates the semantics. As I understand them, and forgive me if I'm missing something, the semantics are, if you match this route, the matching prefix -- whatever that is -- is replied with the replacement. Yes, that is hard to use in complicated inclusion semantics, but this feature is rarely used in those circumstances, so we shouldn't need to over fit to that problem.

If there is disagreement on this part we probably need to do some user research. To the best of my understanding this feature is used to "blank out" the path prefix for applications that think they are mounted at "/"

design/httpproxy-prefix-rewrite.md
```diff
@@ -0,0 +1,262 @@
...
+
+For example, assume that a service support multiple versions of an API and is using the routing prefixes `/v1/api` and `/v2/api`.
+This can only be rewritten to `/api/v1` and `/api/v2` with two separate replacements.
+Distinguishing between these replacements requires specifying the routing prefix.
+
+* Prefix: This is the optional path prefix to match on.
```
davecheney: Is it possible to eliminate this feature from the first round of this design? It feels like it could be added later if required.

design/httpproxy-prefix-rewrite.md
```diff
@@ -0,0 +1,262 @@
...
+  If this is not specified, the replacement is applied for every routing match prefix.
+
+* Replacement: This is the string that replaces the path prefix. If a Prefix field is also specified, the replacement is only applied to the exact route prefix match.
+
+Note that the replacement prefix must be an exact match to exactly one routing prefix.
+
+### Future Extensions
+
+* Regex path rewrites can be added as a new field in the `PathRewritePolicy` struct.
```
davecheney: I'd like to leave these out of scope for the moment. As long as the current design does not make these impossible to add in the future, let's leave them out of scope.

design/httpproxy-prefix-rewrite.md
```diff
@@ -0,0 +1,262 @@
...
+kind: HTTPProxy
+metadata:
+  name: httpbin
+spec:
+  virtualhost:
+    fqdn: httpbin.jpeach.org
+  routes:
+  - conditions:
+    - prefix: /v1/
+    - prefix: /v2/
```
davecheney: This is not permitted at the moment, at most one prefix condition can be supplied per route.

design/httpproxy-prefix-rewrite.md
```diff
@@ -0,0 +1,262 @@
...
+kind: HTTPProxy
+metadata:
+  name: artifactory
+spec:
+  virtualhost:
+    fqdn: artifactory.jpeach.org
+  includes:
+  - name: artifactory-api
+    conditions:
+    - prefix: /
```
davecheney: the conditions block is a NOOP in this case.

design/httpproxy-prefix-rewrite.md
```diff
@@ -0,0 +1,262 @@
...
+    - service: artifactory-dev
+      port: 80
+      weight: 4
+    - service: artifactory-sampler
+      port: 80
+      weight: 1
+      mirror: true
+    pathRewrite:
+      replacePrefix:
+      - prefix: /v1/token/
```
davecheney: I think this is too complicated. _iff_ users have this usecase they can solve this using two different routes or httpproxy documents. If they ask for the ability to rewrite based on the resulting path we could add this later, but I don't think its necessary to replace the missing functionality that we left out of HTTPProxy.

design/httpproxy-prefix-rewrite.md
jpeach: Yes, we can ship without the `Prefix` field. In that case, the semantics will be that all prefixes will be replaced with the replacement. We should retain the enclosing struct so that we can gracefully add this later if necessary.

design/httpproxy-prefix-rewrite.md
jpeach: > Yes, that is hard to use in complicated inclusion semantics, but this feature is rarely used in those circumstances, so we shouldn't need to over fit to that problem.

My perspective on this is that inclusion is a fundamental part of the `HTTPProxy` API. So we should ensure that other parts of the API work with it in a useful, well-defined way.

> this feature is used to "blank out" the path prefix for applications that think they are mounted at "/"

There is a very limited, IMHO, use case for prefix replacement (see non-goals section). Some applications will be able to use it as you describe, but if they ever emit any links, those links are unlikely to be followable (since nothing rewrites them to match the routing prefix). For simple cases, you can do this; I think a more realistic use case is to rewrite API service urls for deployment or legacy reasons (e.g. the `/api/v1` -> `/v1/api`) example)

design/httpproxy-prefix-rewrite.md
jpeach: Right. They are considered here to demonstrate that the structure of this proposal can accommodate them in the future.

design/httpproxy-prefix-rewrite.md
jpeach: Yeh, this example did get a bit too complicated :)

However, as it stands leaf `HTTPProxy` documents can be included from multiple roots or can (I think) be included multiple times from the same root with different routing prefixes. Another perspective is to think of the `Prefix` as defending against unexpected multiple inclusion. Since the replacement has a statically known prefix, if the proxy is included again, a unexpected replacement can't happen. I don't think it's a lot of additional complexity to spec the `Prefix` field to accommodate this.

I touched on this under the heading `Multiple replacements`. Using separate `HTTPProxy` documents requires the user to copy changes to multiple documents, which increases the complexity of their deployments.

design/httpproxy-prefix-rewrite.md
jpeach: Fixed.

PR #1876, opened 2019-11-11
internal/dag: check annotation applicability

### LINKED ISSUE #1825: Cross check Contour annotations
When we validate incoming Kubernetes objects, we should cross-check annotations in the "projectcontour.io" and "contour.heptio.io" namespace to ensure that annotations are applied to the correct object kinds. We can also emit warnings (or errors) if we see unknown or deprected annotations.

jpeach:
When Contour makes use of annotations, it just checks the object
kind where the annotation ought to be. If a user has accidentally
placed the annotation on the wrong kind of object, there is no
indication that something is wrong.

This change adds some initial validation that annotations are applied
to the correct kind of object and defends against unknown annotations
in the Contour annotation namespace. If we detect a problem, we
just log a warning, but eventually we should place an indicator in
the object status so that the user has better visibility.

This fixes #1825.

### DIFF
--- a/internal/dag/annotations.go
+++ b/internal/dag/annotations.go
@@ -23,6 +23,76 @@ import (
 	"k8s.io/api/networking/v1beta1"
 )
 
+func annotationIsKnown(key string) bool {
+	// We should know about everything with a Contour prefix.
+	if strings.HasPrefix(key, "projectcontour.io/") ||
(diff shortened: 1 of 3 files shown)


jpeach:
I think the testing approach for this is a bit wonky, so would appreciate any feedback around that. Also, logging is clearly not very optimal, but I think we can do better after we land the status improvements we have been discussing.

Issue #1513, opened 2019-09-16
internal/envoy: Switch to Safe_Regex


stevesloka:
**Describe the solution you'd like**
The regex  field has been deprecated in favor of safe_regex as it is not safe for use with untrusted input in all cases.

Need to update to use [safe_regex](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route.proto) used here: https://github.com/projectcontour/contour/blob/master/internal/envoy/route.go#L175

stevesloka:
Safe_Regex is new in Envoy 1.12, so this needs to wait until Contour upgrades to that new version.

jpeach:
See #1351

Issue #1880, opened 2019-11-11
Header.Contains tests lack tests


davecheney:
Header Contains conditions that result in a regex matcher are not tested. Tests at the `internal/contour` and `internal/featuretest` level should be added.

stevesloka:
@davecheney there isn't any logic in internal/contour for regex other than sorting them.  There are tests in internal/envoy where the regex is applied. What were your thoughts for tests inside internal/contour?

Issue #1351, opened 2019-08-19
Upgrade to Envoy 1.12.2


davecheney:
Envoy 1.12 was released on 31/Oct.

TODO:

- [x] Update `example/contour` and test
- [ ] Update `Makefile` references
- [x] Update `site/_resources/envoy.md` to note that 1.1 will require envoy 1.12.
- [x] Optional, if desired, back port the 1.12 change to 1.0.x and update `site/_resources/envoy.md` to note that 1.0.x supports Envoy 1.11.2 and 1.12.

davecheney:
Moving to the unplanned milestone. We don't plan to upgrade to Envoy 1.12 for Contour 1.0 at this time.

davecheney:
The envoy xDS deprecations have been addressed. The remainder of this ticket can happen closer to the release date.

PR #1881, opened 2019-11-11
internal/envoy: switch HeaderMatch to SafeRegex

davecheney:
Updates #1351

nb. there appear to be no internal/contour, internal/e2e, or
internal/featuretest tests for header contains matching. I've raised
 #1880 to track this.

### DIFF
--- a/internal/envoy/route.go
+++ b/internal/envoy/route.go
@@ -21,6 +21,7 @@ import (
 	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
 	envoy_api_v2_core "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
 	envoy_api_v2_route "github.com/envoyproxy/go-control-plane/envoy/api/v2/route"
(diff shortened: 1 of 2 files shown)


stevesloka:
Fixes https://github.com/projectcontour/contour/issues/1513

jpeach:
IIUC regexes are exposed to users through Ingress? If so, we should document the change in regex engine? I recalled that RE2 has some slightly different syntax, but maybe I'm misremembering because I can't find it now.

davecheney:
> IIUC regexes are exposed to users through Ingress? If so, we should document the change in regex engine? I recalled that RE2 has some slightly different syntax, but maybe I'm misremembering because I can't find it now.

Yup. This misfeature of ingress is categorically the worse. Envoy offers Re2 but the Ingress spec says the regex engine is extended POSIX as defined by IEEE Std 1003.1. Luckily not many people seem to know that Ingress's path spec is a regex.

Now you're starting to understand why I really really really don't want to expose regex in any form via contour.

Issue #1852, opened 2019-11-04
Discuss site image for building


stevesloka:
**Please describe the problem you have**
Working on the projectcontour.io site (aka Jekyll), there is a makefile task to automate the site, however, each run of it requires all the Gems to be downloaded. 

The first rev of this from @jpeach had a Dockerfile which stored a cache of this essentially in an image but in a review that was taken out. 

I've been doing a bunch of work with the site and found that having that image is super handy. I'm also proposing (#1851) that we add a check to the site for links. Downloading all the Gems each time will add to our build times. 

This issue is to discuss if we want to use a default image that gets built to save us time? 

For now and my own personal needs, I built this: https://github.com/stevesloka/contour-site

jpeach:
I'm +1 on this.  Remember, building a Docker image doesn't necessarily mean publishing it to an official repository.

SDBrett:
I'm happy to work on using a Docker image and relevant documentation for building a local test site. I think that it's a good idea because managing Ruby gems can be painful to mange and this provides consistency for those working on the site.

davecheney:
`bundle exec jekyll serve --livereload` just works for me. I agree getting the right version of ruby installed is a moderate annoyance (personally I use macports and it just works). If you want to work on a docker builder for the site go for it, but please don't publish this as an image, it is not a deliverable of this project.

PR #1884, opened 2019-11-12
Enable bundler caching to decrease local site testing time

SDBrett:
Fixes: #1852 

Added --path option to bunder install commands for site-devel and site-check make targets and added path to .gitignore.

First time running make site-devel or make site-check will install gems to site/bundler/cache. Each subsequent run will use the local cache instead of downloading and installing gems. Changes to the Gemfile and Gemfile.lock will install new / updated gem versions to the cache.

This approach also means that system gems will not be used, only those in the cache location and can provide more consistent results across local testing.

Bundler cache path has been added to .gitignore and Jekyll excludes

Bundler command options: https://bundler.io/v2.0/bundle_install.html
Jekyll excludes reference: https://jekyllrb.com/docs/troubleshooting/#configuration-problems

Signed-off-by: Brett Johnson <brett@sdbrett.com>

### DIFF
--- a/site/_config.yml
+++ b/site/_config.yml
@@ -120,6 +120,12 @@ exclude:
   - Runbook.docx
   - '*.sh'
   - vendor
+  - bundler/bundle/
+  - bundler/cache/
+  - bundler/gems/
+  - bundler/ruby/
+  - node_modules
+
 redcarpet:
     extensions: ["no_intra_emphasis", "tables", "autolink", "strikethrough", "with_toc_data"]
 
--- a/.gitignore
+++ b/.gitignore
@@ -18,9 +18,10 @@ node_modules
 _site
 .jekyll
 .jekyll-metadata
+.jekyll-cache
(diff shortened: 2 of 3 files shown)


site/_config.yml
```diff
@@ -120,6 +120,12 @@ exclude:
   - Runbook.docx
   - '*.sh'
   - vendor
+  - bundler/bundle/
+  - bundler/cache/
+  - bundler/gems/
+  - bundler/ruby/
```
jpeach: - Can we just exclude `bundler/`?
- Should we also exclude `.jekyll-cache`?

SDBrett: Just excluding `bundler/` will not achieve will cause sub directories of `bundler` to be included in the build. The Jekyll link provided in the PR explains the details.

`.jekyll-cache` is new in Jekyll 4 and not present with the PR. It is ignored in the Jekyll4 PR.

PR #1910, opened 2019-11-18
internal/envoy: use common_http_protocol_options.idletimeout

davecheney:
Updates #1351

HttpConnectionManager.IdleTimeout was deprecated in 1.11.2. Switch to
HttpConnectionManager.CommonHttpProtocolOptions.IdleTimeout.

Also, there was one case where ptypes.DurationProtocol was being used
where all the other cases uses protobuf.Duration. Make that the same as
everywhere else.

### DIFF
--- a/internal/envoy/listener.go
+++ b/internal/envoy/listener.go
@@ -99,6 +99,12 @@ func HTTPConnectionManager(routename string, accesslogger []*accesslog.AccessLog
 				}, {
 					Name: wellknown.Router,
 				}},
+				CommonHttpProtocolOptions: &envoy_api_v2_core.HttpProtocolOptions{
+					// Sets the idle timeout for HTTP connections to 60 seconds.
+					// This is chosen as a rough default to stop idle connections wasting resources,
+					// without stopping slow connections from being terminated too quickly.
+					IdleTimeout: protobuf.Duration(60 * time.Second),
+				},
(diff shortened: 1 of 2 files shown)


internal/envoy/listener.go
```diff
@@ -99,6 +99,12 @@ func HTTPConnectionManager(routename string, accesslogger []*accesslog.AccessLog
 				}, {
 					Name: wellknown.Router,
 				}},
+				CommonHttpProtocolOptions: &envoy_api_v2_core.HttpProtocolOptions{
+					// Sets the idle timeout for HTTP connections to 60 seconds.
+					// This is chosen as a rough default to stop idle connections wasting resources,
+					// without stopping slow connections from being terminated too quickly.
+					IdleTimeout: protobuf.Duration(60 * time.Second),
```
stevesloka: Do we document these defaults anywhere? Might be nice to show what's different from Envoy defaults.

davecheney:
Hmm. Good question, @youngnick I think you touched this last back in 0.13/14, do you recall if we called it out anywhere beyond the release notes?

internal/envoy/listener.go
stevesloka: I opened https://github.com/projectcontour/contour/issues/1932 to track.

PR #1913, opened 2019-11-18
internal/envoy: switch to safe regex for path match

### LINKED ISSUE #1513: internal/envoy: Switch to Safe_Regex

davecheney:
Updates #1351
Fixes #1513

Switch path regex matching to safe regex. Refactor matcher.RegexMatcher
code to its own file and helper so it can be reused across
internal/envoy and internal/contour.

Also of note, internal/contour longestRouteFirst comparator has intimate
knowledge of RouteMatch_SafeRegex which is probably not ideal.

### DIFF
--- a/internal/contour/route.go
+++ b/internal/contour/route.go
@@ -255,27 +255,23 @@ func (l longestRouteFirst) Less(i, j int) bool {
 				return true
 			case -1:
 				return false
-			case 0:
+			default:
 				return longestRouteByHeaders(l[i], l[j])
 			}
-
-			panic("bad compare")
 		}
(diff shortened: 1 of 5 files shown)


stevesloka:
Also fixes: https://github.com/projectcontour/contour/issues/1513

PR #1914, opened 2019-11-18
internal/envoy: add missing copyright headers

davecheney:


davecheney:
Done, https://github.com/projectcontour/contour/issues/1917

Issue #1925, opened 2019-11-21
tls upstream validation fails with non-existent CA-secret in IngressRoute contour 1.0


movikbence:
**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
Please check the referenced issue below:
https://github.com/projectcontour/contour/issues/1924

The configuration is the same like in the referenced issue, but the result changed a bit.
Now the "flapping" disappeared from the ingressroute generation, but it still got validated even if the ca-secret is nonexistent, and I am still able to run traffic without TLS upstream validation.

**Environment:**

- Contour version: 1.0
- Kubernetes version: (use `kubectl version`):v1.15.2
- Kubernetes installer & version:  v1.14.2
- Cloud provider or hardware configuration: bare metal
- OS (e.g. from `/etc/os-release`):centOS

davecheney:
As with #1924 Contour 1.0.0 is the current supported release. Are you able to upgrade and confirm if the problem is fixed or persists?

movikbence:
I have tried with 1.0 and the problem is still in place.

davecheney:
Thank you for confirming.

Issue #1924, opened 2019-11-21
tls upstream validation fails with non-existent CA-secret in IngressRoute


movikbence:
**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
I am trying to set up end to end encryption in my environment. If the ca-secret exists in the cluster the validation seems to be working between envoy and the backend, but if I add a nonexistent ca-secret the validation fails, and I am able to run traffic to the backend.

CA-secret creation:
kubectl create secret generic ca-secret --from-file=ca-secret/ca.crt

Configuration files:

apiVersion: contour.heptio.com/v1beta1
kind: IngressRoute
metadata:
  name: e2e-encryption
spec:
  virtualhost:
    fqdn: external-hostname.example.com
    tls:
      secretName: external-tls-secret
  routes:
    - match: /
      services:
        - name: e2e-encryption-service
          port: 443
          validation:
            caSecret: **non-existent**
            subjectName: backend.example.com

Result:
kubectl get ingressroute -A
NAMESPACE   NAME                      FQDN                   TLS SECRET            FIRST ROUTE   STATUS    STATUS DESCRIPTION
default     e2e-encryption   external-hostname.example.com   external-tls-secret   /             invalid   route "/": service "e2e-encryption-service": upstreamValidation requested but secret not found or misconfigured

Although I noticed that it flaps between valid and invalid in the background, it gets generated again and again:
kubectl describe ingressroute -A
Name:         e2e-encryption
Namespace:    default
Labels:       <none>
Annotations:  kubectl.kubernetes.io/last-applied-configuration:
                {"apiVersion":"contour.heptio.com/v1beta1","kind":"IngressRoute","metadata":{"annotations":{},"name":"e2e-encryption","namespace"...
API Version:  contour.heptio.com/v1beta1
Kind:         IngressRoute
Metadata:
  Creation Timestamp:  2019-11-21T07:28:07Z
  Generation:          **1637**
  Resource Version:    23592572
  Self Link:           /apis/contour.heptio.com/v1beta1/namespaces/default/ingressroutes/e2e-encryption
  UID:                 
Spec:
  Routes:
    Match:  /
    Services:
      Name:  e2e-encryption-service
      Port:  443
      Validation:
        Ca Secret:     non-existent
        Subject Name:  backend.example.com
  Virtualhost:
    Fqdn:  external-hostname.example.com
    Tls:
      Secret Name:  external-tls-secret
Status:
  Current Status:  valid
[...]

PR #1928, opened 2019-11-21
Fix upstream validation error in IngressRoute

tsaarni:
Fixes: #1924
Fixes: #1925

Processing IngressRoute did not stop to an error when validation.caSecret
referred to a non-existing secret.  As result, IngressRoute was considered
valid and upstream validation was skipped.

Stop processing IngressRoute and indicate invalid status when upstream
validation is defined but cannot be done due to error in configuration.


davecheney:
Thank you for working on this. Please add a test in internal/dag or internal/featuretest

tsaarni:
There was actually a test case which apparently passed the unexpected behavior as expected. I have now updated it but please check that you agree on this?

PR #1936, opened 2019-11-25
internal/envoy: switch to Cluster.TransportSocket for upstream TLS

davecheney:
Updates #1351

Envoy 1.12 has deprecated the Cluster.TLSContext field, instead the
UpstreamTLSContext must be provided via a typed config attached to a
TransportSocket.

ref: https://www.envoyproxy.io/docs/envoy/v1.12.0/api-v2/api/v2/cds.proto#envoy-api-field-cluster-tls-context

### DIFF
--- a/internal/e2e/cds_test.go
+++ b/internal/e2e/cds_test.go
@@ -1017,7 +1017,9 @@ func externalnamecluster(name, servicename, statName, externalName string, port 
 
 func tlscluster(name, servicename, statsName string, ca []byte, subjectName string) *v2.Cluster {
 	c := cluster(name, servicename, statsName)
-	c.TlsContext = envoy.UpstreamTLSContext(ca, subjectName)
+	c.TransportSocket = envoy.UpstreamTLSTransportSocket(
+		envoy.UpstreamTLSContext(ca, subjectName),
+	)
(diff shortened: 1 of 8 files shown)


davecheney:
> This looks fine to me. My only concern is that maybe older versions of envoy (that we don't officially support, but that may be working in practice) would break.

Yup, this is not compatible with Enovy 1.11

> As an aside, how did you discover this was deprecated? IIUC there is a envoy build option that removes all the deprecated features; is that what you used?

I recently discovered envoy list the things they deprecate as part of the release notes

https://www.envoyproxy.io/docs/envoy/v1.12.0/intro/deprecated

jpeach:
> On 26 Nov 2019, at 10:48 am, Dave Cheney <notifications@github.com> wrote:
> 
> This looks fine to me. My only concern is that maybe older versions of envoy (that we don't officially support, but that may be working in practice) would break.
> 
> Yup, this is not compatible with Enovy 1.11

The 1.11 docs do document "transport_socket", though it wasn't clear to me whether that can be applied in the same context. I'm a bit concerned about being incompatible with 1.11, since 1.12 has been out for less than a month.

PR #1943, opened 2019-11-26
Path prefix rewrite for HTTPProxy

jpeach:
This implements prefix rewrite for HTTPProxy. Most of the semantics described in the design proposal are present, but pay particular attention to `expandPrefixMatches`.

I'm also very open to advice on improving the test coverage here. I've added feature tests and verified they have full coverage of the new code, but maybe we want some more granular tests as well.

Documentation will be coming in a subsequent PR.

This implements #1874 
This updates #1873


internal/dag/builder.go
```diff
@@ -512,6 +512,91 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
...
+// | `/foo`          | `/bar/`     | `/foosball` | X `/bar/sball` |
+// | `/foo/`         | `/bar/`     | `/foo/type` |   `/bar/type`  |
+func expandPrefixMatches(routes []*Route) []*Route {
+	prefixedRoutes := map[string][]*Route{}
+
+	expandedRoutes := []*Route{}
+
+	// First, we group the Routes by their slash-consistent prefix match condition.
+	for _, r := range routes {
+		routingPrefix := r.PathCondition.(*PrefixCondition).Prefix
```
davecheney: is this safe to do? PathCondition could be a *RegexCondition

internal/dag/builder.go
```diff
@@ -512,6 +512,91 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
...
+// match on only `/foo` or `/foo/`, then the unwanted rewrites marked
+// with X can result. This means that we need to generate separate
+// prefix matches (and replacements) for the two cases.
+//
+// | Matching Prefix | Replacement | Client Path | Rewritten Path |
+// |-----------------|-------------|-------------|----------------|
+// | `/foo`          | `/bar`      | `/foosball` |   `/barsball`  |
+// | `/foo/`         | `/bar`      | `/foo/type` | X `/bartype`   |
+// | `/foo`          | `/bar/`     | `/foosball` | X `/bar/sball` |
+// | `/foo/`         | `/bar/`     | `/foo/type` |   `/bar/type`  |
```
davecheney: Can you please add cases for

match: /foo, replace: /, client path: /foo/v1, rewritten: /v1

internal/dag/builder.go
```diff
@@ -591,6 +676,38 @@ func (b *Builder) computeRoutes(sw *ObjectStatusWriter, proxy *projcontour.HTTPP
...
+			if err := prefixReplacementsAreValid(route.GetPrefixReplacements()); err != nil {
+				sw.SetInvalid(err.Error())
+				return nil
+			}
+
+			// Note that we are guaranteed to always have a prefix
+			// condition. Even if the CRD user didn't specify a
+			// prefix condition, mergePathConditions() guarantees
+			// a prefix of '/'.
+			routingPrefix := r.PathCondition.(*PrefixCondition).Prefix
```
davecheney: this probably needs to be guarded with a type switch

internal/featuretests/kubernetes.go
```diff
@@ -124,3 +127,19 @@ func secretdata(cert, key string) map[string][]byte {
...
 	}
 }
+
+func meta(name string) metav1.ObjectMeta {
+	v := strings.SplitN(name, "/", 2)
+	switch len(v) {
+	case 1:
+		// no prefix
+		return metav1.ObjectMeta{
+			Name: v[0],
```
davecheney: I don't think this is safe to do, meta("foo") produces 
```
metav1.ObjectMeta{
  Namespace: "",
  Name: "foo",
}
```
I don't think we should let fixtures like this be created.

internal/featuretests/replaceprefix_test.go
```diff
@@ -0,0 +1,577 @@
...
+		TypeUrl: routeType,
+	})
+}
+
+func TestHTTPProxyPathPrefix(t *testing.T) {
+	subtests := []struct {
+		Name string
+		Func func(*testing.T)
+	}{
+		{Name: "Basic", Func: basic},
```
davecheney: Does this indirection add much value? What about just calling the `basic` function, `TestHTTPProxyPathPrefixBasic`?

apis/projectcontour/v1/httpproxy.go
```diff
@@ -147,6 +147,19 @@ type Route struct {
 	// The load balancing policy for this route.
 	// +optional
 	LoadBalancerPolicy *LoadBalancerPolicy `json:"loadBalancerPolicy,omitempty"`
+	// The policy for rewriting the path of the request URL
+	// after the request has been routed to a Service.
+	//
+	// +kubebuilder:validation:Optional
```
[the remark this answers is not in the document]
jpeach: Yeh, `pathRewritePolicy` is more consistent.

internal/featuretests/replaceprefix_test.go
jpeach: I think it has a sliver of value. It groups the tests better and makes it easier to run the group. I don't think there's much downside.

internal/dag/builder.go
jpeach: I convinced myself that HTTPProxy was guaranteed to have a PrefixCondition because of `mergePathConditions`. I agree that adding the check back makes it more robust though, and will be needed when HTTPProxy supports regex match.

internal/featuretests/kubernetes.go
jpeach: OK, I can just add a panic in that case ...

internal/dag/builder.go
jpeach: This table is demonstrating the envoy behavior, so in the example you give, the rewritten path  would be `//v1`.

internal/dag/builder.go
jpeach: Fixed.

apis/projectcontour/v1/httpproxy.go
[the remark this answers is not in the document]
davecheney: ta

internal/dag/builder.go
davecheney: ta

internal/dag/builder.go
davecheney: I'm sorry but we must address this problem. match: /foo, replace: /, client path: /foo/v1, must be rewritten as /v1 to resolve #899

jpeach: Yes, we do address this, see the "replaceWithSlash" test.

This table is showing "the behavior of Envoy prefix rewrite". It is not showing the behavior of how Contour configures the rewrite. In this comment, I'm trying to explain the rationale for why we need to generate new routes for the different rewrite cases.

internal/dag/policy.go
```diff
@@ -169,3 +170,22 @@ func max(a, b uint32) uint32 {
 	}
 	return b
 }
+
+func prefixReplacementsAreValid(replacements []projcontour.ReplacePrefix) error {
+	prefixes := map[string]struct{}{}
```
davecheney: why not make this 

 prefixes := map[string]bool 

Then the logs below becomes

if prefixes[r.Prefix] { ... }

davecheney:
Fixes #899 ?

internal/dag/policy.go
jpeach: Fixed.

Issue #1952, opened 2019-11-27
httpproxy's tcpproxy plus permit insecure does not require the route to be marked as permit insecure


davecheney:
Contour presents `default/service` over port 80 without the `permitInsecure` market on the route.
```
apiVersion: projectcontour.io/v1
kind: HTTPProxy
metadata:
  name: example
  namespace: default
spec:
  virtualhost:
    fqdn: tcp.example.com
    tls:
      passthrough: true
  routes:
      services:
        - name: correct
          port: 443
 tcpproxy:
    services:
    - name: correct
      port: 443
```
This is probably a bug. We should required that if routes and tcpproxy are combined, all routes must be marked `permitInsecure: true`

davecheney:
Closed in 20eb8ee, this is not a bug.

Issue #1951, opened 2019-11-27
Contour should reject httpproxy with a missing service unconditionally


davecheney:
Under some conditions Contour may be permitting the httpproxy with a missing backend service:
```
apiVersion: projectcontour.io/v1
kind: HTTPProxy
metadata:
  name: example
  namespace: default
spec:
  virtualhost:
    fqdn: tcp.example.com
    tls:
      passthrough: true
  routes:
      services:
        - name: missing
          port: 443
 tcpproxy:
    services:
    - name: correct
      port: 443
```
This may be a holdover from the days when kubectl schema validation required a dummy `routes:` key.

Issue #1945, opened 2019-11-26
Multiple applications deployed in one namespace


prateekjainaa:
We deployed one application in our k8s cluster, this application has following characteristics:

1. It handles SSL on its own.
2. It exposes multiple contexts say, /login, /dashboard etc.

We removed its Ingress object and deployed new HTTPProxy like

```
# httpproxy-tls-passthrough.yaml
apiVersion: projectcontour.io/v1
kind: HTTPProxy
metadata:
  name: example
  namespace: default
spec:
  virtualhost:
    fqdn: tcp.example.com
    tls:
      passthrough: true
  tcpproxy:
    services:
    - name: loginAndDashboard
      port: 443

```
Everything worked fine till now. We are able to login and access services from this app. Now, we deployed another application (in same namespace) with similar characteristics:

1. It handles SSL on its own.
2. It exposes multiple contexts say, /configureUser, /deleteUser etc.

So, we added its service to the list under tcpProxy

```
tcpproxy:
    services:
    - name: loginAndDashboard
      port: 443
    - name: userService
      port: 443
```

We expected that new service will also be accessible in addition to previous service. But userService was not accessible. So, we thought that it is reasonable behavior as how will contour decide which request should be forwarded to which service. So, we defined explicit paths in HTTPProxy object like (routes section),

```
# httpproxy-tls-passthrough.yaml
apiVersion: projectcontour.io/v1
kind: HTTPProxy
metadata:
  name: example
  namespace: default
spec:
  virtualhost:
    fqdn: tcp.example.com
    tls:
      passthrough: true
  routes:
    - conditions:
      - prefix: /login
      services:
        - name: loginAndDashboard
          port: 443
    - conditions:
      - prefix: /dashboard
      services:
        - name: loginAndDashboard
          port: 443
    - conditions:
      - prefix: /configureUser
      services:
        - name: userService
          port: 443
    - conditions:
      - prefix: /deleteUser
      services:
        - name: userService
          port: 443
 tcpproxy:
    services:
    - name: loginAndDashboard
      port: 443
    - name: userService
      port: 443
```

But now even the parts of previous service started to become inaccessible. Are we trying to do something wrong/not-supported here? Where can we look for errors because its status shows its a valid proxy.

davecheney:
@prateekjainaa I'm sorry you are having issues. The root cause is TCP proxying cannot be combined with routing (there is one edge case where this statement is not true, but it does not apply in your scenario).

[...]

Issue #910, opened 2019-03-01
TLS Passthrough & HTTP Redirect


256dpi:
I have a pod that accepts only HTTPS traffic on port 443. To route the traffic, I used an IngressRoute with the new TLS Passthrough option:

```yml
apiVersion: contour.heptio.com/v1beta1
kind: IngressRoute
metadata:
  name: test
spec:
  virtualhost:
    fqdn: test.example.com
    tls:
      passthrough: true
  tcpproxy:
    services:
      - name: test
        port: 443
  routes:
    - match: /
      services:
        - name: test
          port: 443
```

The TLS forwarding works great. However, contour is not accepting traffic on port 80 and therefore does not auto-redirect the request to the HTTPS port. Am I missing some configuration?

**Environment:**

- Contour version: 0.9.0
- Kubernetes version: 1.11 (kubectl: 1.13)
- Cloud provider or hardware configuration: GKE

davecheney:
Thank you for your issue. The problem is when using tcpproxy you're mixing L4 and L7 in this service. I see the argument that for your service you want to provide HTTPS from the backend service using TCP forwarding, so it makes sense that you'd want to have a 80 -> 443 redirect. But that is only in the case where your backend service is HTTPS, if it were some other service using TLS as a transport, then port 80 redirection does not make sense.

256dpi:
@davecheney  Thanks for the detailed explanation!

It would be nice if contour would support that out of the box in a HTTPS forwarding scenario.

I'm now trying to forward the HTTP traffic to my instance and perform the redirect there using this config:

```yml
apiVersion: contour.heptio.com/v1beta1
kind: IngressRoute
metadata:
  name: test
spec:
  virtualhost:
    fqdn: test.example.com
    tls:
      passthrough: true
  tcpproxy:
    services:
      - name: test
        port: 443
  routes:
    - match: /
      services:
        - name: test
          port: 80
          permitInsecure: true
```

Unfortunately, It's not working. Is this the right approach? Do I need two separate ingress route definitions? Thanks for your help!

davecheney:
@256dpi i'm sorry this is a limitation at the moment. You cannot have to ingressroute records with the same fqdn.

davecheney:
@256dpi sorry it's taken me so long to get back to this issue, this is something i'm hoping to address in beta.1 (or at least confirm that we cannot support it for Contour 1.0).

[...]

PR #1953, opened 2019-11-27
internal/featuretest: add test to assert tlspassthrough + permit insecure works

davecheney:
Updates #1945

Clean up the existing tcp proxy feature tests to assert that no
listeners are opened unexpectedly. Then add a test to assert that the
pattern of exposing a service both via tcpproxy and port 80 (permit
insecure) works

There are several bugs that have fallen out of this

- #1951 - Contour accepts a httpproxy with a missing service without
marking the httpproxy invalid.
- #1952 - httpproxy's tcpproxy plus permit insecure does not require the
service to be marked as permit insecure.
- #1954 - route should not be present on ingress_https if tcpproxy is
inuse.

Both of these should be fixed, but lets lock in the current behaviour.

### DIFF
--- a/internal/featuretests/tcpproxy_test.go
+++ b/internal/featuretests/tcpproxy_test.go
@@ -84,7 +84,7 @@ func TestTCPProxy(t *testing.T) {
 	rh.OnAdd(svc)
 	rh.OnAdd(i1)
 
-	c.Request(listenerType, "ingress_https").Equals(&v2.DiscoveryResponse{
+	c.Request(listenerType).Equals(&v2.DiscoveryResponse{
 		Resources: resources(t,
 			&v2.Listener{
 				Name:         "ingress_https",
@@ -93,8 +93,19 @@ func TestTCPProxy(t *testing.T) {
 				ListenerFilters: envoy.ListenerFilters(
 					envoy.TLSInspector(),
 				),
-			}),
+			},
+			staticListener(),
+		),
 		TypeUrl: listenerType,
+	})
+
+	// check that both ingress_http and ingress_https are empty
+	c.Request(routeType).Equals(&v2.DiscoveryResponse{
+		Resources: resources(t,
+			envoy.RouteConfiguration("ingress_http"),
+			envoy.RouteConfiguration("ingress_https"),
+		),
+		TypeUrl: routeType,
 	})
 
 	rh.OnDelete(i1)
@@ -128,7 +139,7 @@ func TestTCPProxy(t *testing.T) {
 	}
 	rh.OnAdd(hp1)
 
-	c.Request(listenerType, "ingress_https").Equals(&v2.DiscoveryResponse{
+	c.Request(listenerType).Equals(&v2.DiscoveryResponse{
 		Resources: resources(t,
 			&v2.Listener{
 				Name:         "ingress_https",
@@ -137,10 +148,20 @@ func TestTCPProxy(t *testing.T) {
 				ListenerFilters: envoy.ListenerFilters(
 					envoy.TLSInspector(),
 				),
-			}),
+			},
+			staticListener(),
+		),
 		TypeUrl: listenerType,
 	})
 
+	// check that both ingress_http and ingress_https are empty
+	c.Request(routeType).Equals(&v2.DiscoveryResponse{
+		Resources: resources(t,
+			envoy.RouteConfiguration("ingress_http"),
+			envoy.RouteConfiguration("ingress_https"),
+		),
+		TypeUrl: routeType,
+	})
 }
 
 func TestTCPProxyDelegation(t *testing.T) {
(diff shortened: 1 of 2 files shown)


internal/featuretests/tcpproxy_test.go
```diff
@@ -490,4 +553,434 @@ func TestTCPProxyTLSBackend(t *testing.T) {
...
+		),
+		TypeUrl: listenerType,
+	})
+
+	// check port 80 is open and the route is a 301 upgrade.
+	c.Request(routeType).Equals(&v2.DiscoveryResponse{
+		Resources: resources(t,
+			envoy.RouteConfiguration("ingress_http",
+				envoy.VirtualHost("kuard-tcp.example.com",
+					upgradeHTTPS(routePrefix("/")),
```
jpeach: If the listener on port 80 is meant to be routing traffic to "default/backend:80" as per the listener comment, doesn't this prevent that? ie. rather than routing requests, this route config always responds with a redirect?

internal/featuretests/tcpproxy_test.go
```diff
@@ -490,4 +553,434 @@ func TestTCPProxyTLSBackend(t *testing.T) {
...
+				ListenerFilters: envoy.ListenerFilters(
+					envoy.TLSInspector(),
+				),
+			},
+			staticListener(),
+		),
+		TypeUrl: listenerType,
+	})
+
+	// check port 80 is open and the route is a 301 upgrade.
```
jpeach: This comment seems wrong?

internal/featuretests/tcpproxy_test.go
```diff
@@ -490,4 +553,434 @@ func TestTCPProxyTLSBackend(t *testing.T) {
...
+	c.Request(routeType).Equals(&v2.DiscoveryResponse{
+		Resources: resources(t,
+			envoy.RouteConfiguration("ingress_http",
+				envoy.VirtualHost("kuard-tcp.example.com",
+					envoy.Route(routePrefix("/"),
+						routeCluster("default/backend/80/da39a3ee5e"),
+					),
+				),
```
jpeach: I think it wold be helpful to have a bit more exposition on what is happening here. IIUC envoy is terminating TLS on the 443 listener and forwarding the TCP stream, and also routing HTTP traffic from the port 80 listener. Is that right?

internal/featuretests/tcpproxy_test.go
```diff
@@ -490,4 +553,434 @@ func TestTCPProxyTLSBackend(t *testing.T) {
...
+			),
+		),
+		TypeUrl: routeType,
+	})
+}
+
+// Assert that TCPProxy + TLSPassthrough and a HTTP service can be used to expose a ingress_http
+// route on the same vhost that port ingress_https is tls passthrough + proxing.
```
jpeach: s/proxing/proxying/

above and below too

internal/featuretests/tcpproxy_test.go
```diff
@@ -490,4 +553,434 @@ func TestTCPProxyTLSBackend(t *testing.T) {
...
+					Passthrough: true,
+				},
+			},
+			Routes: []projcontour.Route{{
+				Conditions: prefixCondition("/"),
+				Services: []projcontour.Service{{
+					Name: svc.Name,
+					Port: 80,
+				}},
+			}},
```
jpeach: This seems ambiguous on an API level. Just from looking at the CRD, I wouldn't expect to be getting any HTTP traffic from the port 80 listener.

internal/featuretests/tcpproxy_test.go
```diff
@@ -490,4 +553,434 @@ func TestTCPProxyTLSBackend(t *testing.T) {
...
+				ListenerFilters: envoy.ListenerFilters(
+					envoy.TLSInspector(),
+				),
+			},
+			staticListener(),
+		),
+		TypeUrl: listenerType,
+	})
+
+	// check port 80 is open and the route is a 301 upgrade.
```
jpeach: Comment seems wrong?

internal/featuretests/tcpproxy_test.go
davecheney: my bad, this comment wasn't clear. I've rewritten it.

internal/featuretests/tcpproxy_test.go
davecheney: fixed

internal/featuretests/tcpproxy_test.go
davecheney: comment added

internal/featuretests/tcpproxy_test.go
davecheney: This is #910.

internal/featuretests/tcpproxy_test.go
davecheney: fixed

PR #1960, opened 2019-11-28
internal/dag: reject a tcpproxy httpproxy without valid tls details

### LINKED ISSUE #1958: Contour marks a httproxy w/ tcpproxy unconditionally valid if it does not have a TLS block
Contour accepts this httpproxy definition without validating if the tcpproxy service is missing because there is no tls block.
```
kind: HTTPProxy
metadata:
  name: example
  namespace: default
spec:
  virtualhost:
    fqdn: tcp.example.com
 tcpproxy:
    services:
    - name: missing
      port: 443
```
httpproxy's with a tcpproxy block cannot be valid without a spec.virtualhost.tls stanza and one of passthrough or a valid secret.

davecheney:
Fixes #1958

To be a valid HTTPProxy, if the tcpproxy stanza is provided, the
HTTPProxy must also feature a virtualhost.tls spec with either
passthrough: true, or a valid secretName.

### DIFF
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -495,9 +495,16 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
 		}
 	}
 
-	if proxy.Spec.TCPProxy != nil && (passthrough || enforceTLS) {
-		if !b.processHTTPProxyTCPProxy(sw, proxy, nil, host) {
+	if proxy.Spec.TCPProxy != nil {
+		tls := proxy.Spec.VirtualHost.TLS
+		if tls == nil {
+			sw.SetInvalid("tcpproxy: missing tls.passthrough or tls.secretName")
 			return
+		}
+		if passthrough || enforceTLS {
+			if !b.processHTTPProxyTCPProxy(sw, proxy, nil, host) {
+				return
+			}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


internal/dag/builder.go
```diff
@@ -495,10 +495,17 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
 		}
 	}
 
-	if proxy.Spec.TCPProxy != nil && (passthrough || enforceTLS) {
-		if !b.processHTTPProxyTCPProxy(sw, proxy, nil, host) {
+	if proxy.Spec.TCPProxy != nil {
```
davecheney: The logic here is pretty dense. I'll take a crack at cleaning it up after this PR lands.

internal/dag/builder.go
```diff
@@ -495,10 +495,17 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
...
+		tls := proxy.Spec.VirtualHost.TLS
+		if tls == nil {
+			sw.SetInvalid("tcpproxy: missing tls details")
```
jpeach: Nit: s/tls/TLS/

davecheney:
Thanks for the review, I’m not happy with that message at all. I wrote lower case tls, as in the name of the field, but it’s totally not clear. Maybe I should change the message to say “tcpproxy: missing tls.passthrough or tls.secretName”

PR #1964, opened 2019-11-29
internal/featuretests: add status check machinery

jpeach:
Add a Kubernetes object status cache that can be used to assert
object status in feature tests. Status can be `Equals()` (all
fields must match) or `Like()` (only specified fields need to
match). The assertion functions return a `*Contour` to that a
fluent style can be used.

Sample usage:

```Go
c.Request().Equals(
        ...
).Status().Equals(
        ...
)
```


jpeach:
@davecheney I polished this up and addressed the delete consistency. Can you please take another look?

PR #1972, opened 2019-12-02
site: add navigation class to the TOC

jpeach:
Add the "navigation" class to the in-page TOC elements so that it
can be used consistently with CSS selectors.


SDBrett:
LGTM

PR #1974, opened 2019-12-02
doc: add HTTPProxy prefix rewrite documentation

### LINKED ISSUE #1873: Support prefix rewrite in HTTPProxies.

jpeach:
This fixes #1873.

### DIFF
--- a/site/docs/master/httpproxy.md
+++ b/site/docs/master/httpproxy.md
@@ -355,7 +355,7 @@ For `header` conditions there is one required field, `name`, and five operator f
 
 - `exact` is a string, and checks that the header exactly matches the whole string. `notexact` checks that the header does *not* exactly match the whole string.
 
-#### Multiple Routes
+### Routes
 
 HTTPProxy must have at least one route or include defined.
 Paths defined are matched using prefix conditions.
@@ -715,7 +715,65 @@ spec:
           port: 80
 ```
 
-#### ExternalName
+#### Path Rewriting
+
+HTTPProxy supports rewriting the HTTP request URL path prior to delivering the request to the backend service.
+Rewriting is performed after a routing decision has been made, and never changes the request destination.
+
+The `pathRewritePolicy` field specifies how the path prefix should be rewritten.
+The `replacePrefix` rewrite policy specifies a replacement string for a HTTP request path prefix match.
+When this field is present, the path prefix that the request matched is replaced by the text specified in the `replacement` field.
+If the HTTP request path is longer than the matched prefix, the remainder of the path is unchanged.
+
+```yaml
+apiVersion: projectcontour.io/v1
+kind: HTTPProxy
+metadata:
+  name: rewrite-example
+  namespace: default
+spec:
+  virtualhost:
+    fqdn: rewrite.bar.com
+  routes:
+  - services:
+    - name: s1
+      port: 80
+    pathRewritePolicy:
+      replacePrefix:
+      - replacement: /new/prefix
+```
+
+The `replacePrefix` field accepts an array of possible replacements.
+When more than one `replacePrefix` array element is present, the `prefix` field can be used to disambiguate which replacement to apply.
+
+If no `prefix` field is present, the replacement is applied to all prefix matches made against the route.
+If a `prefix` field is present, the replacement is applied only to routes that have an exactly matching [prefix condition](#prefix-conditions).
+Specifying more than one `replacePrefix` entry is mainly useful when a HTTPProxy document is included into multiple parent documents.
+
+```yaml
+apiVersion: projectcontour.io/v1
+kind: HTTPProxy
+metadata:
+  name: rewrite-example
+  namespace: default
+spec:
+  virtualhost:
+    fqdn: rewrite.bar.com
+  routes:
+  - services:
+    - name: s1
+      port: 80
+    conditions:
+    - prefix: /v1/api
+    pathRewritePolicy:
+      replacePrefix:
+      - prefix: /v1/api
+        replacement: /app/api/v1
+      - prefix: /
+        replacement: /app
+```
+
+### ExternalName
 
 HTTPProxy supports routing traffic to service types `ExternalName`.
 Contour looks at the `spec.externalName` field of the service and configures the route to use that DNS name instead of utilizing EDS.


site/docs/master/httpproxy.md
```diff
@@ -715,7 +715,65 @@ spec:
           port: 80
 ```
 
-#### ExternalName
+#### Path Rewriting
+
+HTTPProxy supports rewriting the HTTP request URL path prior to delivering the request to the service.
```
stevesloka: nit: Should maybe not use `service`? I understand what you mean, but maybe backend application or pod or something?

site/docs/master/httpproxy.md
```diff
@@ -715,7 +715,65 @@ spec:
...
+  - services:
+    - name: s1
+      port: 80
+    conditions:
+    - prefix: /v1/api
+    pathRewritePolicy:
+      replacePrefix:
+      - prefix: "/v1/api"
+        replacement: "/app/api/v1"
+      - prefix: "/"
```
stevesloka: If the delegation model doesn't actually include these prefixes to this route, is this a status warning in the proxy?

stevesloka: In this example, the prefix is `/v1/api`, how would `/` route here? Wouldn't it need to be another route configured?

site/docs/master/httpproxy.md
```diff
@@ -715,7 +715,65 @@ spec:
...
+      port: 80
+    pathRewritePolicy:
+      replacePrefix:
+      - replacement: "/new/prefix"
+```
+
+The `replacePrefix` field accepts an array of possible replacements.
+When more than one `replacePrefix` array element is present, the `prefix` field can be used to disambiguate which replacement to apply.
+
+If no `prefix` field is present, the replacement is applied to all prefix matches made against the route.
```
stevesloka: This part is confusing to me, if my path was `/`, it would replace `/foo`, `/foo/bar`, and `/foo/bar/zed/<anything>` with the specification in the `replacement` field. 

Is this what users would want since a longer path would always get replaced with the same resulting string?

stevesloka:
I was on PTO for this feature, so I think many of my questions are around that design.

site/docs/master/httpproxy.md
```diff
@@ -355,7 +355,7 @@ For `header` conditions there is one required field, `name`, and five operator f
 
 - `exact` is a string, and checks that the header exactly matches the whole string. `notexact` checks that the header does *not* exactly match the whole string.
 
-#### Multiple Routes
+### Routes
 
 HTTPProxy must have at least one route or include defined.
```
davecheney: This is out of scope for this review, but since I saw it I don't think this is true, a httpproxy can be valid without a routes or includes section if it has a tcpproxy section.

site/docs/master/httpproxy.md
davecheney: i've always said something waffely like _backend service_

site/docs/master/httpproxy.md
```diff
@@ -715,7 +715,65 @@ spec:
...
-#### ExternalName
+#### Path Rewriting
+
+HTTPProxy supports rewriting the HTTP request URL path prior to delivering the request to the service.
+Rewriting is performed after a routing decision has been made, and never changes the request destination.
+
+The `pathRewritePolicy` field specifies how the path prefix should be rewritten.
+The `replacePrefix` rewrite policy specifies a replacement string for a HTTP request path prefix match.
+When this field is present, the path prefix that the request matched is replaced by the text specified in the `replacement` field.
+The original path before rewriting will be placed into the into the `x-envoy-original-path` header.
```
davecheney: I'm wary of documenting an internal detail of envoy. If they drop this in the future, we'll have to add it manually ourselves.

site/docs/master/httpproxy.md
```diff
@@ -715,7 +715,65 @@ spec:
...
+spec:
+  virtualhost:
+    fqdn: rewrite.bar.com
+  routes:
+  - services:
+    - name: s1
+      port: 80
+    pathRewritePolicy:
+      replacePrefix:
+      - replacement: "/new/prefix"
```
davecheney: for consistency does this need to be quoted? fqdn, name and namespace were not

site/docs/master/httpproxy.md
```diff
@@ -715,7 +715,65 @@ spec:
...
+    fqdn: rewrite.bar.com
+  routes:
+  - services:
+    - name: s1
+      port: 80
+    conditions:
+    - prefix: /v1/api
+    pathRewritePolicy:
+      replacePrefix:
+      - prefix: "/v1/api"
```
davecheney: again for consistency all strings should be quoted, or none of them should be

site/docs/master/httpproxy.md
jpeach: I documented this because `IngressRoute` documented it. Happy to drop this.

site/docs/master/httpproxy.md
jpeach: No, you don't have to consume all the replacement prefixes. When we have richer status, we could annotate unconsumed replacement prefixes with a non-error condition.

jpeach: There would have to be a parent proxy that creates a route for `/`, but there's no requirement to do that. I didn't want to make this example too complicated, so omitted other proxies.

site/docs/master/httpproxy.md
jpeach: No, only "the path prefix that the request matched is replaced". I attempted to clarify.

site/docs/master/httpproxy.md
jpeach: Fixed.

site/docs/master/httpproxy.md
jpeach: Separate issue for this?

site/docs/master/httpproxy.md
jpeach: Fixed.

site/docs/master/httpproxy.md
jpeach: Fixed.

site/docs/master/httpproxy.md
stevesloka: But if a path is delegated `/` then it will replace all paths passed since you didn't define the prefix to replace, wouldn't it?

site/docs/master/httpproxy.md
```diff
@@ -715,7 +715,65 @@ spec:
...
+  namespace: default
+spec:
+  virtualhost:
+    fqdn: rewrite.bar.com
+  routes:
+  - services:
+    - name: s1
+      port: 80
+    conditions:
+    - prefix: /v1/api
```
stevesloka: My point for the previous comment is this route defines the path `/v1/api`, how would `/` get passed to this route in the rewrite policy defined on line 772?

site/docs/master/httpproxy.md
stevesloka: Maybe open an issue so we don't forget?

site/docs/master/httpproxy.md
jpeach: `/` is the prefix that would be replaced. The net result is that it would look like you prepended to the request path.

site/docs/master/httpproxy.md
jpeach: Fixed.

site/docs/master/httpproxy.md
jpeach: You could subsequently include this in another proxy that routed `/`?

site/docs/master/httpproxy.md
jpeach: #1987

site/docs/master/httpproxy.md
jpeach: #1988

PR #1976, opened 2019-12-03
Ingress status

### LINKED ISSUE #403: Address not set on Ingress resource
Contour does not set the IP value on the Ingress resource. This value is necessary for services like [ExternalDNS](https://github.com/kubernetes-incubator/external-dns) to work properly.

### Version
The version that `https://j.hept.io/contour-deployment-norbac` produces. Atm:
docker.io/envoyproxy/envoy-alpine:v1.6.0
gcr.io/heptio-images/contour:master (pointing to `b9ede01f025`)

### Repro
1. Setup Contour in your cluster.
`kubectl apply -f kubectl apply -f https://j.hept.io/contour-deployment-norbac`
1. Deploy a simple service with Ingress rules. 
`kubectl apply -f https://gist.githubusercontent.com/nphmuller/1c980c72e2119b1482e19eb619def08f/raw/10331f4bbc18abe36a0c8c841bf70ab04a38eb92/service-with-ingress.yaml`
1. (A) `kubectl get ingress nginx`
1. (B) `kubectl get ingress nginx -o yaml`

### Expected
A. ADDRESS column should have external IP of Contour load balancer service, but is empty.
B. status should have value:
``` yaml
status:
  loadBalancer:
    ingress:
    - ip: %External IP-of-Contour-LoadBalancer-service%
```
but has value:
``` yaml
status:
  loadBalancer: {}
```

stevesloka:
Fixes #403 by updating Ingress resources with Envoy's service type=LoadBalancer status information. 

This also adds two new args to contour: 

- envoy-service-name: Envoy service name (default: `envoy`)
- envoy-service-namespace: Envoy service namespace (default: where Contour is running)

### DIFF
--- a/cmd/contour/contour.go
+++ b/cmd/contour/contour.go
@@ -99,7 +99,17 @@ func main() {
 	}
 }
 
-func newClient(kubeconfig string, inCluster bool) (*kubernetes.Clientset, *clientset.Clientset, *coordinationv1.CoordinationV1Client) {
+// Returns the current namespace in which the controller is running
+// If error returns empty string
+func getCurrentNamespace() string {
+	ns, _, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(clientcmd.NewDefaultClientConfigLoadingRules(), &clientcmd.ConfigOverrides{}).Namespace()
+	if err != nil {
+		return "projectcontour"
+	}
+	return ns
+}
+
+func newClient(kubeconfig string, inCluster bool) (kubernetes.Interface, *clientset.Clientset, *coordinationv1.CoordinationV1Client) {
 	var err error
 	var config *rest.Config
 	if kubeconfig != "" && !inCluster {
@@ -112,6 +122,7 @@ func newClient(kubeconfig string, inCluster bool) (*kubernetes.Clientset, *clien
 
 	client, err := kubernetes.NewForConfig(config)
 	check(err)
+
 	contourClient, err := clientset.NewForConfig(config)
 	check(err)
 	coordinationClient, err := coordinationv1.NewForConfig(config)
--- a/cmd/contour/leadership.go
+++ b/cmd/contour/leadership.go
@@ -27,7 +27,7 @@ import (
 
 // newLeaderElector creates a new leaderelection.LeaderElector and associated
 // channels by which to observe elections and depositions.
-func newLeaderElector(log logrus.FieldLogger, ctx *serveContext, client *kubernetes.Clientset, coordinationClient *coordinationv1.CoordinationV1Client) (*leaderelection.LeaderElector, chan struct{}, chan struct{}) {
+func newLeaderElector(log logrus.FieldLogger, ctx *serveContext, client kubernetes.Interface, coordinationClient *coordinationv1.CoordinationV1Client) (*leaderelection.LeaderElector, chan struct{}, chan struct{}) {
 
 	// leaderOK will block gRPC startup until it's closed.
 	leaderOK := make(chan struct{})
@@ -64,7 +64,7 @@ func newLeaderElector(log logrus.FieldLogger, ctx *serveContext, client *kuberne
 
 // newResourceLock creates a new resourcelock.Interface based on the Pod's name,
 // or a uuid if the name cannot be determined.
-func newResourceLock(ctx *serveContext, client *kubernetes.Clientset, coordinationClient *coordinationv1.CoordinationV1Client) resourcelock.Interface {
+func newResourceLock(ctx *serveContext, client kubernetes.Interface, coordinationClient *coordinationv1.CoordinationV1Client) resourcelock.Interface {
 	resourceLockID, found := os.LookupEnv("POD_NAME")
 	if !found {
 		resourceLockID = uuid.New().String()
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -103,6 +103,8 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 	serve.Flag("root-namespaces", "Restrict contour to searching these namespaces for root ingress routes").StringVar(&ctx.rootNamespaces)
 
 	serve.Flag("ingress-class-name", "Contour IngressClass name").StringVar(&ctx.ingressClass)
+	serve.Flag("envoy-service-name", "Defines which Envoy service name handles ingress traffic for this instance of Contour").StringVar(&ctx.envoyServiceName)
+	serve.Flag("envoy-service-namespace", "Defines which Envoy service namespace handles ingress traffic for this instance of Contour").StringVar(&ctx.envoyServiceNamespace)
 
 	serve.Flag("envoy-http-access-log", "Envoy HTTP access log").StringVar(&ctx.httpAccessLog)
 	serve.Flag("envoy-https-access-log", "Envoy HTTPS access log").StringVar(&ctx.httpsAccessLog)
@@ -124,6 +126,11 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 
 	// step 1. establish k8s client connection
 	client, contourClient, coordinationClient := newClient(ctx.Kubeconfig, ctx.InCluster)
+
+	// step 1b. get current namespace that Contour is running inside if not defined
+	if ctx.envoyServiceNamespace == "" {
+		ctx.envoyServiceNamespace = getCurrentNamespace()
+	}
 
 	// step 2. create informers
 	// note: 0 means resync timers are disabled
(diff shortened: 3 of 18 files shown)


jpeach:
@stevesloka FYI, this will have significant merge conflicts with #1975

cmd/contour/contour.go
```diff
@@ -99,7 +99,17 @@ func main() {
 	}
 }
 
-func newClient(kubeconfig string, inCluster bool) (*kubernetes.Clientset, *clientset.Clientset, *coordinationv1.CoordinationV1Client) {
+// Returns the current namespace in which the controller is running
+// If error returns empty string
```
davecheney: s/returns the empty string/returns "projectcontour"

cmd/contour/contour.go
```diff
@@ -99,7 +99,17 @@ func main() {
...
+func getCurrentNamespace() string {
+	ns, _, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(clientcmd.NewDefaultClientConfigLoadingRules(), &clientcmd.ConfigOverrides{}).Namespace()
```
davecheney: wow, that's a function name and a half :)

How does this determine where its loading the kubeconfig from? If its magic, can we use the same magic for contour's --kubeconfig flag?

cmd/contour/contour.go
```diff
@@ -112,6 +122,7 @@ func newClient(kubeconfig string, inCluster bool) (*kubernetes.Clientset, *clien
 
 	client, err := kubernetes.NewForConfig(config)
 	check(err)
+
```
davecheney: nit: maybe use \n on line 127 and onwards for consistency?

cmd/contour/leadership.go
```diff
@@ -27,7 +27,7 @@ import (
 
 // newLeaderElector creates a new leaderelection.LeaderElector and associated
 // channels by which to observe elections and depositions.
-func newLeaderElector(log logrus.FieldLogger, ctx *serveContext, client *kubernetes.Clientset, coordinationClient *coordinationv1.CoordinationV1Client) (*leaderelection.LeaderElector, chan struct{}, chan struct{}) {
+func newLeaderElector(log logrus.FieldLogger, ctx *serveContext, client kubernetes.Interface, coordinationClient *coordinationv1.CoordinationV1Client) (*leaderelection.LeaderElector, chan struct{}, chan struct{}) {
```
davecheney: This looks like an unrelated change, maybe pull it out into its own PR to make this easier to review.

cmd/contour/serve.go
```diff
@@ -125,6 +127,11 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	// step 1. establish k8s client connection
 	client, contourClient, coordinationClient := newClient(ctx.Kubeconfig, ctx.InCluster)
 
+	// step 1b. get current namespace that Contour is running inside if not defined
+	if ctx.envoyServiceNamespace == "" {
```
davecheney: setting ctx.envoyServiceName and ctx.envoyServiceNamespace are handled differently, is there a way to make them more similar?

cmd/contour/servecontext.go
```diff
@@ -164,6 +170,7 @@ func newServeContext() *serveContext {
 			Name:          "leader-elect",
 		},
 		UseExtensionsV1beta1Ingress: false,
+		envoyServiceName:            "envoy",
```
davecheney: Can `envoyServiceNamespace` default to `projectcontour`?

internal/certgen/certgen.go
```diff
@@ -75,7 +75,7 @@ func WriteSecretsYAML(outputDir, namespace string, certdata map[string][]byte) e
 
 // WriteSecretsKube writes all the keypairs out to Kube Secrets in the
 // passed Kube context.
-func WriteSecretsKube(client *kubernetes.Clientset, namespace string, certdata map[string][]byte) error {
+func WriteSecretsKube(client kubernetes.Interface, namespace string, certdata map[string][]byte) error {
```
davecheney: This looks like an unrelated change, I think this file and the one above can be pulled out into their own PR to make this simpler to review.

davecheney:
>  I think this is the right way to do this because the inputs into updating the status of k8s Ingress objects are only the service record for Envoy itself.

I'm sorry, I was not correct. k8s Ingress status depends on the Envoy pod's service record _and_ the ingress class setting of contour. With that said, I think this feature would be better implemented as stand alone as possible because it takes different inputs to the dag, and runs at a different timescale.

youngnick:
Also, I think that it's important that our design here allow for other deployment methods - nginx gets around this by allowing you to either specify the service, or just IPs or FQDNs directly. Being able to cover 'monitor this service' and 'just let me specify the IPs or FQDNs' hits all the deployment scenarios I can think of, in cloud and otherwise.

If there's a concurrency-safe struct somewhere that stores the IPs or FQDNs to go into the Ingress status stanza, then we can either poke a value in there directly at startup (for statically specified IPs), or have a goroutine pull them from an Informer.

davecheney:
> On 3 Dec 2019, at 3:15 pm, Nick Young <notifications@github.com> wrote:
> 
> Also, I think that it's important that our design here allow for other deployment methods - nginx gets around this by allowing you to either specify the service, or just IPs or FQDNs directly. Being able to cover 'monitor this service' and 'just let me specify the IPs or FQDNs' hits all the deployment scenarios I can think of, in cloud and otherwise.
> 
> If there's a concurrency-safe struct somewhere that stores the IPs or FQDNs to go into the Ingress status stanza, then we can either poke a value in there directly at startup (for statically specified IPs), or have a goroutine pull them from an Informer.

I think whatever approach we decide on this needs to run continuously; it needs to watch something which holds the IP addresses of the Service. The Service’s details can change at any time, so we need to be ready to flow that update to all relevant Ingress objects.

cmd/contour/contour.go
stevesloka: It's using client-go to determine which namespace Contour is deployed into. I was trying to avoid folks having to add extra args to deployments by automatically finding this information. 

It can still be overridden by args if the user would like. I also didn't want to read the `/var/run/secrets/kubernetes.io/serviceaccount/namespace` path in the pod which seems like  hack when we have a k8s client (via client-go).

cmd/contour/contour.go
stevesloka: Sorry I don't follow your comment. I didn't change any of this short of moving from `kubernetes.ClientSet` to `kubernetes.Interface`. The `check(err)` stops execution of the program.

cmd/contour/leadership.go
stevesloka: I wanted to add a test which actually tested that I could set the status of an Ingress object. In doing that I can use the client-go `fake` objects which gives me a fake k8s api. In order to use this I need to pass around the interface which I changed up in (https://github.com/projectcontour/contour/pull/1976/files/952c08508fffd3c71a7c0030f6a8e41511a3ba13#diff-b62d9e38d59686d41f8f41098ff436f9R112) to return an interface. That change bubbles down through all the other methods.

cmd/contour/serve.go
stevesloka: The name come from a namespace, I'm using the check for empty to determine if a user hasn't already set the flag. If they haven't then I look up what namespace Contour is running in.

cmd/contour/servecontext.go
stevesloka: It does if it cannot be looked up. I like the idea of requiring fewer flags to set on Contour, so I opted to look up the namespace which Contour is running in. If we take that out, we force users to set these flags and I can remove the lookup.

internal/certgen/certgen.go
stevesloka: See the previous comment about testing with fakes.

stevesloka:
> Also, I think that it's important that our design here allow for other deployment methods - nginx gets around this by allowing you to either specify the service, or just IPs or FQDNs directly. Being able to cover 'monitor this service' and 'just let me specify the IPs or FQDNs' hits all the deployment scenarios I can think of, in cloud and otherwise.

The implementation allows for this. Right now I just implemented the service lookup. To allow users to specify their own, we need to add additional args.

stevesloka:
> I'm sorry, I was not correct. k8s Ingress status depends on the Envoy pod's service record and the ingress class setting of contour. With that said, I think this feature would be better implemented as stand alone as possible because it takes different inputs to the dag, and runs at a different timescale.

Doing it in the dag build gave me the ingress class, service info, etc all for free. I can throw this all away and re-implement in a goroutine. Thanks for the feedback!

stevesloka:
@davecheney do you have thoughts on how we could get the events copied to this new goroutine? I put the update in the dag update because we have access to the services/ingress objects via the watch+event handler.  

However, moving the Ingress status updates outside will need to hook into the Event Handler or get a copy of the services, ingress events from the k8s watcher.

So to do this in a new routine we'd need something like this:

1. Get a new stream of updates from the informers to get access to service/ingress changes from the cluster
2. Have a new cache to store the set of ingress objects
3. Watch for changes to services, when we find one check for lb info as well as controller class info and update Ingress objects from cache

stevesloka:
@davecheney is going to pick this up, closing since he'll have a new PR.

PR #1977, opened 2019-12-03
internal/featuretests: test status in prefix tests

jpeach:


internal/featuretests/replaceprefix_test.go
```diff
@@ -110,6 +114,9 @@ func basic(t *testing.T) {
 			envoy.RouteConfiguration("ingress_https"),
 		),
 		TypeUrl: routeType,
+	}).Status(vhost).Equals(projcontour.Status{
+		CurrentStatus: dag.StatusInvalid,
```
davecheney: its kinda gross we have to couple the feature test package to the dag package just for this constant. What happens if you use the string "invalid" here. Yeah, that's duplicating information, but that might be worth it for the decoupling, currently the feature tests know nothing about the dag.

jpeach:
On Dec 3, 2019, at 6:17 PM, Dave Cheney <notifications@github.com> wrote:

﻿

@davecheney commented on this pull request.

________________________________

In internal/featuretests/replaceprefix_test.go<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fprojectcontour%2Fcontour%2Fpull%2F1977%23discussion_r353015121&data=02%7C01%7Cjpeach%40vmware.com%7C56fefe6ff8b04acedeae08d777c0e110%7Cb39138ca3cee4b4aa4d6cd83d9dd62f0%7C0%7C0%7C637109542655992396&sdata=c08YCzTeMraefTxUhdf8iSAEyzND2ItzQ8Blrl0UA0o%3D&reserved=0>:

> @@ -110,6 +114,9 @@ func basic(t *testing.T) {
                        envoy.RouteConfiguration("ingress_https"),
                ),
                TypeUrl: routeType,
+       }).Status(vhost).Equals(projcontour.Status{
+               CurrentStatus: dag.StatusInvalid,


its kinda gross we have to couple the feature test package to the dag package just for this constant. What happens if you use the string "invalid" here. Yeah, that's duplicating information, but that might be worth it for the decoupling, currently the feature tests know nothing about the dag.

Maybe we need to move these constants out of day? K8s?

internal/featuretests/replaceprefix_test.go
jpeach: @davecheney I think the right fix for this is to move those constants into the `k8s` package. WDYT?

davecheney: there are only three states "valid", "invalid" and "orphaned" all of which 

a. don't change
b. are shorter than their respective declarations

It seems simpler to just use the literal constant and avoid the import depenedency.

jpeach: Maybe we should get rid fo the constant then?

davecheney: SGTM 👍 👍

jpeach: The benefit of a constant is that you can find where it is used (and also improve type checking).

davecheney: The downside is it creates strong source coupling between packages that shouldn't know anything about each other. 1/2 a dozen of one, six of the other.

jpeach: In that case, I'd prefer to move it to the k8s package.

davecheney: Sounds good to me!

jpeach:
@davecheney I moved the status constants to the `internal/k8s` package.

PR #1979, opened 2019-12-03
Add 'omitempty' to HTTPProxy Status field.

### LINKED ISSUE #1973: HTTPProxy Status field should have `omitempty` in its JSON fieldspec
Small thing, but it's important for `ir2proxy` operations.

youngnick:
Fixes #1973

### DIFF
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -302,8 +302,10 @@ type UpstreamValidation struct {
 
 // Status reports the current state of the HTTPProxy.
 type Status struct {
-	CurrentStatus string `json:"currentStatus"`
-	Description   string `json:"description"`
+	// +optional
+	CurrentStatus string `json:"currentStatus,omitempty"`
+	// +optional
+	Description string `json:"description,omitempty"`
 }
 
 // +genclient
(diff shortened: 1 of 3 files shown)


apis/projectcontour/v1/httpproxy.go
```diff
@@ -322,7 +322,7 @@ type HTTPProxy struct {
 
 	Spec HTTPProxySpec `json:"spec"`
 	// +optional
-	Status `json:"status"`
+	Status `json:"status,omitempty"`
```
davecheney: Does this suffer from the same problem as https://github.com/projectcontour/ir2proxy/issues/8; this is a struct field marker `omitempty` -- or is the problem in https://github.com/projectcontour/ir2proxy/issues/8 some compounding factor?

youngnick: ah, 🤦‍♂ 
Yes, you're totally correct. This will need to change to
```go
Status *Status `json:"status,omitempty"`
```
Which is a much bigger change. I'll work on this today.

davecheney:
What if we mark the two fields of status omit empty?

youngnick:
That would be a smaller change, for sure. It would mean that the HTTPProxy objects will contain
```yaml
status: {}
```
stanzas when generated though.

davecheney:
Better than nothing I guess

youngnick:
After some discussion, @davecheney and I settled on adding 'omitempty' to all the parts of the Status. I think this should actually work, the creationTimestamp thing is because it has a custom Marshal method.

youngnick:
Also, I have regenerated the code and schemas.

PR #1981, opened 2019-12-04
Add tests to cover override behaviour for IngressRoute and HTTPProxy

### LINKED ISSUE #1949: HTTPProxy should overwrite IngressRoute if there is a conflict
As part of #1689, I realised that we don't have a good upgrade path for IngressRoute to HTTPProxy right now.

Currently, AFAICT, having an IngressRoute and HTTPProxy with matching FQDNs and routes will render them both invalid.

This issue covers building something so that, if an IngressRoute and HTTPProxy overlap, then the HTTPProxy will take priority.

This allows us to have an upgrade process of:
- Have IngressRoutes deployed and working
- Build matching HTTPProxy objects (#1689 is to assist here)
- Apply new HTTPProxy objects
- Verify
- Remove either HTTPProxy objects (for rollback) or IngressRoute objects (for completion).

youngnick:
Fixes #1949

Issue #1949 is about ensuring that HTTPProxy objects will override IngressRoute
ones when they match on {FQDN, path} tuple. Turns out this is already the case,
this PR adds tests to assert this behaviour.

The tests have an Ingress, IngressRoute, and HTTPProxy that are identical aside from the service where the traffic is routed to, and assert that the override behaviour holds based on which service ends up chosen.

Note that routes that are not exactly the same (on FQDN, path tuple), will just be additive, no matter where they come from. This is why the cert-manager ingress-shim thing works.

### DIFF
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -795,6 +795,18 @@ func TestDAGInsert(t *testing.T) {
 		},
 	}
 
+	i17 := &v1beta1.Ingress{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "kuard",
+			Namespace: "default",
+		},
+		Spec: v1beta1.IngressSpec{
+			Rules: []v1beta1.IngressRule{{
+				Host:             "example.com",
+				IngressRuleValue: ingressrulevalue(backend("kuard", intstr.FromInt(8080))),
+			}},
+		},
+	}
 	// s3a and b have http/2 protocol annotations
 	s3a := &v1.Service{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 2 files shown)


youngnick:
After some discussion with Dave, this is missing some tests, and a feature: The ingressroute that is superseded by the HTTPProxy should have an invalid status. I'll work more on this tomorrow.

internal/dag/builder_test.go
```diff
@@ -5479,6 +5550,57 @@ func TestDAGInsert(t *testing.T) {
 				},
 			),
 		},
+		// This tests what happens when you build with an ingressroute and a HTTPProxy
```
davecheney: suggestion: try the imperative mood

// Assert that a route in IngressRoute takes precedence over Ingress.

davecheney:
@youngnick if the tests pass, please merge this and then iterate on it. Nothing is set in stone til you cut a release of this tool.

davecheney:
pls ignore my previous reply, wrong window

internal/dag/builder_test.go
youngnick: Hmm, good idea, that's much more concise. I'll change this, and then merge.

davecheney:
@youngnick ping, this Is approved, are you ok to land it?

youngnick:
Sorry, yes, will land it now.

PR #1985, opened 2019-12-05
internal/dag: reject certificates with CN or SubjectAltName

### LINKED ISSUE #1965: Reject TLS secrets with certificate that has no CN or SANs
Envoy crashes when processing a TLS certificate that does not have SubjectAltNames or a CN field in the Subject.

Create a key, certificate and TLS Secret and then an Ingress that uses this Secret:
```
$ openssl req -new -newkey rsa:2048 -days 365 -nodes -subj="/DC=com/DC=domain/DC=my" -x509 -keyout server.key -out server.crt -sha256
$ kubectl create secret tls tls-secret-bad --cert=server.crt --key=server.key
secret/tls-secret-bad created

$ cat ingress.yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    kubernetes.io/ingress.class: contour
  name: test
spec:
  rules:
  - host: my.domain.com
    http:
      paths:
      - backend:
          serviceName: test
          servicePort: 8080
        path: /
  tls:
  - hosts:
    - my.domain.com
    secretName: tls-secret-bad

$ kubectl create -f ingress.yaml
ingress.extensions/test created

$ kubectl create deployment test --image=gcr.io/google_containers/explorer:1.0
deployment.apps/test created

$ kubectl expose deployment test --port=8080
service/test exposed
```

The envoy Pod logs show an assertion failure:
```
$ kubectl -n projectcontour logs envoy-7ms2h
...
[2019-11-29 09:32:24.547][1][info][upstream] [source/server/lds_api.cc:60] lds: add/update listener 'ingress_https'
[2019-11-29 09:32:24.547][1][info][upstream] [source/server/lds_api.cc:60] lds: add/update listener 'stats-health'
[2019-11-29 09:32:24.549][1][critical][assert] [source/extensions/transport_sockets/tls/context_impl.cc:838] assert failure: cn_index >= 0.
[2019-11-29 09:32:24.549][1][critical][backtrace] [bazel-out/k8-opt/bin/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:81] Caught Aborted, suspect faulting address 0x1
[2019-11-29 09:32:24.549][1][critical][backtrace] [bazel-out/k8-opt/bin/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:69] Backtrace (use tools/stack_decode.py to get line numbers):
[2019-11-29 09:32:24.549][1][critical][backtrace] [bazel-out/k8-opt/bin/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:75] #0: [0x7f492059e420]
[2019-11-29 09:32:24.551][1][critical][backtrace] [bazel-out/k8-opt/bin/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:73] #1: Envoy::Extensions::TransportSockets::Tls::ServerContextImpl::ServerContextImpl() [0xbadb44]
```
and stay in CrashLoopBackOff until the offending Ingress is removed.

davecheney:
Fixes #1965

During cache insertion reject certificates which lack a Subject
CommonName (CN) or SubjectAltName extension.

This PR uncovered that, well basically, all our TLS fixtures don't fit
this requirement which is extra sadface and thus this PR contains
replacement fixtures for all the previous ones.

This PR also adds a unit test for dag.isValidSecret as we are likely to
continue to add to this function as further validation edge cases are
uncovered. This necessitated a small addition to the internal/assert
comparator logic to be able to compare error values.


internal/contour/listener_test.go
```diff
@@ -1068,7 +1068,7 @@ func TestListenerVisit(t *testing.T) {
 
 func transportSocket(tlsMinProtoVersion envoy_api_v2_auth.TlsParameters_TlsProtocol, alpnprotos ...string) *envoy_api_v2_core.TransportSocket {
 	return envoy.DownstreamTLSTransportSocket(
-		envoy.DownstreamTLSContext("default/secret/28337303ac", tlsMinProtoVersion, alpnprotos...),
+		envoy.DownstreamTLSContext("default/secret/68621186db", tlsMinProtoVersion, alpnprotos...),
```
davecheney: all the hashes changed because the name of a SDS entry contains the hash of its certificate. I'm not sure if this was a bright idea in retrospect.

internal/dag/secret.go
```diff
@@ -135,6 +142,19 @@ func validateCertificate(data []byte) error {
 	return nil
 }
 
+func hasCommonName(c *x509.Certificate) bool {
+	return strings.TrimSpace(c.Subject.CommonName) != ""
+}
+
+func hasSubjectAltNames(c *x509.Certificate) bool {
```
jpeach: How about:
```Go
return len(c.DNSNames) > 0 || len(c.IPAddresses) > 0
```

internal/dag/secret.go
```diff
@@ -122,9 +123,15 @@ func validateCertificate(data []byte) error {
 		if block.Type != "CERTIFICATE" {
 			return fmt.Errorf("unexpected block type '%s'", block.Type)
 		}
-		if _, err := x509.ParseCertificate(block.Bytes); err != nil {
+		cert, err := x509.ParseCertificate(block.Bytes)
+		if err != nil {
 			return err
 		}
+
+		if !hasCommonName(cert) && !hasSubjectAltNames(cert) {
```
jpeach: IIUC, the CN is not required if you have an altname

davecheney: I think the logic I was going for was the cert needs at least one of a CN or an alt name. I think I got that logic right.

internal/dag/secret.go
davecheney: done

internal/dag/secret.go
jpeach: Erm, yes you did :)

davecheney: | !hasCommonName | !hasSubjectAltName | Result | 
| ------------------- | ---------------------| ------ |
| false                         | false                           | false |
| false                      | true | false |
| true | false | false |
| true | true | true |

I think I got it right, but writing out the truth table hurt.

PR #1995, opened 2019-12-06
api: mark all timeoutPolicy parameters optional

### LINKED ISSUE #1993: HTTPProxy is rejected by API server when timeoutPolicy.idle is empty

I wrote the following `HTTPProxy` but it was rejected by APIserver (kubectl) with the message `spec.routes.timeoutPolicy.idle in body is required`.
```yaml
kind: HTTPProxy
metadata:
  ...
spec:
  ...
  routes:
    # For static files
    - conditions:
      - prefix: /
      services:
      - name: server-https
        port: 443
      timeoutPolicy:
        response: 2m
```

In the reference document, I can omit `idle` field.
When `idle` field is omitted, it should be filled by the default value (5m).

https://projectcontour.io/docs/v1.0.0/httpproxy/#httpproxy-api-specification

I think, each field in Timeout.Policy should be optional & omitempty.

https://github.com/projectcontour/contour/blob/768510af6486f20ed032f20515a9878db07c16b0/apis/projectcontour/v1/httpproxy.go#L226

**Environment:**

- Contour version: 1.0.0
- Kubernetes version: (use `kubectl version`): 1.15.4

davecheney:
Fixes #1993

Mark timeoutPolicy.request and timeoutPolicy.idle +optional (for schema
generation) and omitempty (for JSON encoding). This permits the user to
create a timeoutPolicy without having to specify both fields. We have
test coverage in internal/envoy and internal/dag for these scenarios.

### DIFF
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -231,11 +231,13 @@ type TimeoutPolicy struct {
 
 	// Timeout for receiving a response from the server after processing a request from client.
 	// If not supplied the timeout duration is undefined.
-	Response string `json:"response"`
+	// +optional
+	Response string `json:"response,omitempty"`
 
-	// Timeout after which if there are no active requests, the connection between Envoy and the
-	// backend will be closed.
-	Idle string `json:"idle"`
+	// Timeout after which if there are no active requests for this route, the connection between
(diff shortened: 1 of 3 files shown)


jpeach:
@davecheney needs a "make render"

PR #1996, opened 2019-12-06
Move header condition tests for HTTPProxy to internal/featuretests

stevesloka:
Updates #1880 by moving header condition tests for HTTPProxy to `internal/featuretests`. 


stevesloka:
Also not sure if folks have ideas on what to test in `internal/contour` from what the issue references: https://github.com/projectcontour/contour/issues/1880#issuecomment-561345108

internal/featuretests/httpproxy.go
```diff
@@ -16,11 +16,45 @@ package featuretests
...
 
 import (
+	"strings"
+
 	projcontour "github.com/projectcontour/contour/apis/projectcontour/v1"
 )
 
-func prefixCondition(prefix string) []projcontour.Condition {
+// Returns an array of Conditions
+func prefixConditions(prefix string) []projcontour.Condition {
```
davecheney: I don't think this is accurate, the method is called prefixConditions plural, but only takes a single value.

internal/featuretests/httpproxy.go
```diff
@@ -16,11 +16,45 @@ package featuretests
...
+// Returns a single Condition
+func prefixCondition(prefix string) projcontour.Condition {
+	return projcontour.Condition{
+		Prefix: prefix,
+	}
+}
+
+func conditions(c ...projcontour.Condition) []projcontour.Condition { return c }
+
+func headerCondition(name, value, headerType string, present bool) projcontour.Condition {
```
davecheney: this helper doesn't seem to help much, you have to pass four parameters into it, some of which will be ignored depending on others. What about something like

```
headerContains(key, value string) projcontour.Condition
headerExists(key string) projcontour.Conditoin
```
and so forth

internal/featuretests/httpproxy.go
stevesloka: It's based on what it returns. I need a way to pass a set of conditions which is why the `conditions()` method was added. I can add that everywhere to resolve.

internal/featuretests/httpproxy.go
stevesloka: sure thing

internal/featuretests/httpproxy.go
davecheney: > It's based on what it returns. 

Then it should probably be something like
```
func conditions(first projcontour.Condition, rest ...projcontour.Condition) []projcontour.Condition {
   return append([]projcontour.Condition{first}, rest...)
}
```
I this would also make `prefixCondition` more annoying to use as it would become
```
Conditions: conditions(prefixCondition("/yada")),
```
Calling `prefixCondition` that was probably a mistake, but I was thinking about it from the POV that the name described what it took, not what it returned

PR #2000, opened 2019-12-06
site: Add v1.1.0 docs

stevesloka:
Updates #1968

- Updates site to add v1.1.0 docs
- Updates `master` docs to be generic so they aren't version specific

### DIFF
--- a/site/_config.yml
+++ b/site/_config.yml
@@ -54,6 +54,11 @@ defaults:
     values:
       version: v1.0.0
       layout: "docs"
+  - scope:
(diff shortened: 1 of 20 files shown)


stevesloka:
Please hold until v1.1.0 is ready to release.

davecheney:
@stevesloka thanks for working on this. I'm worried there are conflicts with other changes that landed to /master since this was taken. I think we should hold off on landing 1.1.0 docs until the day of the release -- I know this sucks, but I can't think if a different want to avoid loosing changes.

Issue #1982, opened 2019-12-04
External Name Host Rewrite


stevesloka:
# Background 
A use-case came up chatting with someone in the community that involves routing requests to external domains, so using Contour as a proxy to redirect requests to locations other than pods inside a Kubernetes cluster. 

Today Contour has support for `externalName` services. An externalName type service allows for a Kubernetes service to define a fqdn that when a request is made to it, the request is directed to that fqdn (e.g. foo.com). 

The one issue with how externalName is implemented today is that host header on the request sent to the externalName is not swapped, it includes the original host definition. 

## Example: 

- The fqdn & path, foo.com/something, is configured to route to the externalName: `extsvc`.
- `extsvc`: Is configured to route to: `bar.com`
- A request to: `foo.com/something` is then proxied to `bar.com/something`, however the host header field is still `foo.com`. 

# Proposal

Add a new field to the service struct called `rewriteHost`. When set to `true`, then the value of the externalName will be swapped in the Host header for proxied requests. This is a more straightforward approach solving a smaller use-case vs the next alternative proposal.  This allows for a host rewrite on only the services which target externalNames.

An alternative implementation would be to create a `rewriteHostPolicy` that gets attached to the VirtualHost struct allowing for a global host rewrite for that fqdn. This isn't needed for HTTPProxy since we can have multiple roots target the same child Proxy documents.

jpeach:
A few notes:

* You need to both set the right Host header, and also use the right SNI name
* I think that a "header rewrite" policy at the same level as "path rewrite" is a better approach than a specific "rewriteHostPolicy"
* For external names in particular, can we just make sure they work right without extra configuration? We should just install the correct rewrites, etc without the user needing additional config

stevesloka:
> I think that a "header rewrite" policy at the same level as "path rewrite" is a better approach than a specific "rewriteHostPolicy"

I wonder though if something like this should live higher up at the vhost level?

jpeach:
> On 5 Dec 2019, at 1:12 pm, Steve Sloka <notifications@github.com> wrote:
> 
> I think that a "header rewrite" policy at the same level as "path rewrite" is a better approach than a specific "rewriteHostPolicy"
> 
> I wonder though if something like this should live higher up at the vhost level?

I think that rewriting headers is a useful feature, but I'm not sure that we should distinguish the host header.

davecheney:
[...]

PR #2006, opened 2019-12-07
Support per-Split header manipulation.

mattmoor:
This changes adds support for adding and removing request or response headers
for each service target in a Contour HTTPProxy resource.

The structure differs from the originally proposed shape in the linked issue
for two reasons:
1. A desire to support response-header manipulation.
2. A desire to support split-specific header manipulation (needed by Knative).

The resulting shape looks like:

```yaml
spec:
  virtualhost:
    fqdn: foo.default.mattmoor.io
  routes:
  - services:
    - name: foo
      port: 80
      requestHeaders:
        add:
        - name: Foo
          value: bar
        remove:
        - Baz
      responseHeaders:
        add:
        - name: Ugh
          value: tedious
        remove:
        - Secret-Stuff
```

Fixes: https://github.com/projectcontour/contour/issues/70

### DIFF
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -152,9 +152,12 @@ type Route struct {
 	//
 	// +kubebuilder:validation:Optional
 	PathRewritePolicy *PathRewritePolicy `json:"pathRewritePolicy,omitempty"`
-	// The policy for managing headers during proxying
+	// The policy for managing request headers during proxying
 	// +kubebuilder:validation:Optional
 	RequestHeadersPolicy *HeadersPolicy `json:"requestHeadersPolicy,omitempty"`
+	// The policy for managing response headers during proxying
+	// +kubebuilder:validation:Optional
+	ResponseHeadersPolicy *HeadersPolicy `json:"responseHeadersPolicy,omitempty"`
 }
 
 func (r *Route) GetPrefixReplacements() []ReplacePrefix {
@@ -201,6 +204,12 @@ type Service struct {
 	UpstreamValidation *UpstreamValidation `json:"validation,omitempty"`
 	// If Mirror is true the Service will receive a read only mirror of the traffic for this route.
 	Mirror bool `json:"mirror,omitempty"`
+	// The policy for managing request headers during proxying
+	// +kubebuilder:validation:Optional
+	RequestHeadersPolicy *HeadersPolicy `json:"requestHeadersPolicy,omitempty"`
+	// The policy for managing response headers during proxying
+	// +kubebuilder:validation:Optional
+	ResponseHeadersPolicy *HeadersPolicy `json:"responseHeadersPolicy,omitempty"`
 }
 
 // HTTPHealthCheckPolicy defines health checks on the upstream service.
@@ -301,6 +310,9 @@ type HeadersPolicy struct {
 	// Set specifies a list of HTTP header values that will be set in the HTTP header
 	// +kubebuilder:validation:Optional
 	Set []HeaderValue `json:"set,omitempty"`
+	// Remove specifies a list of HTTP header names to remove
+	// +kubebuilder:validation:Optional
+	Remove []string `json:"remove,omitempty"`
 }
 
 // HeaderValue represents a header name/value pair
--- a/internal/contour/route.go
+++ b/internal/contour/route.go
@@ -139,10 +139,19 @@ func (v *routeVisitor) visit(vertex dag.Vertex) {
 							Action: envoy.UpgradeHTTPS(),
 						})
 					} else {
-						routes = append(routes, &envoy_api_v2_route.Route{
+						rt := &envoy_api_v2_route.Route{
 							Match:  envoy.RouteMatch(route),
 							Action: envoy.RouteRoute(route),
-						})
+						}
+						if route.RequestHeadersPolicy != nil {
+							rt.RequestHeadersToAdd = envoy.HeaderValueList(route.RequestHeadersPolicy.Set, false)
+							rt.RequestHeadersToRemove = route.RequestHeadersPolicy.Remove
+						}
+						if route.ResponseHeadersPolicy != nil {
+							rt.ResponseHeadersToAdd = envoy.HeaderValueList(route.ResponseHeadersPolicy.Set, false)
+							rt.ResponseHeadersToRemove = route.ResponseHeadersPolicy.Remove
+						}
+						routes = append(routes, rt)
 					}
 				})
 
@@ -161,10 +170,19 @@ func (v *routeVisitor) visit(vertex dag.Vertex) {
 						return
 					}
 
-					routes = append(routes, &envoy_api_v2_route.Route{
+					rt := &envoy_api_v2_route.Route{
 						Match:  envoy.RouteMatch(route),
 						Action: envoy.RouteRoute(route),
-					})
+					}
+					if route.RequestHeadersPolicy != nil {
+						rt.RequestHeadersToAdd = envoy.HeaderValueList(route.RequestHeadersPolicy.Set, false)
+						rt.RequestHeadersToRemove = route.RequestHeadersPolicy.Remove
+					}
+					if route.ResponseHeadersPolicy != nil {
+						rt.ResponseHeadersToAdd = envoy.HeaderValueList(route.ResponseHeadersPolicy.Set, false)
+						rt.ResponseHeadersToRemove = route.ResponseHeadersPolicy.Remove
+					}
+					routes = append(routes, rt)
 				})
 				if len(routes) < 1 {
 					return
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -684,14 +684,27 @@ func (b *Builder) computeRoutes(sw *ObjectStatusWriter, proxy *projcontour.HTTPP
 			return nil
 		}
 
+		reqHP, err := headersPolicy(route.RequestHeadersPolicy, true /* allow Host */)
+		if err != nil {
+			sw.SetInvalid(err.Error())
+			return nil
+		}
+
+		respHP, err := headersPolicy(route.ResponseHeadersPolicy, false /* allow Host */)
+		if err != nil {
+			sw.SetInvalid(err.Error())
+			return nil
+		}
+
 		r := &Route{
-			PathCondition:        mergePathConditions(conds),
-			HeaderConditions:     mergeHeaderConditions(conds),
-			Websocket:            route.EnableWebsockets,
-			HTTPSUpgrade:         routeEnforceTLS(enforceTLS, route.PermitInsecure && !b.DisablePermitInsecure),
-			TimeoutPolicy:        timeoutPolicy(route.TimeoutPolicy),
-			RetryPolicy:          retryPolicy(route.RetryPolicy),
-			RequestHeadersPolicy: headerPolicy(route.RequestHeadersPolicy),
+			PathCondition:         mergePathConditions(conds),
+			HeaderConditions:      mergeHeaderConditions(conds),
+			Websocket:             route.EnableWebsockets,
+			HTTPSUpgrade:          routeEnforceTLS(enforceTLS, route.PermitInsecure && !b.DisablePermitInsecure),
+			TimeoutPolicy:         timeoutPolicy(route.TimeoutPolicy),
+			RetryPolicy:           retryPolicy(route.RetryPolicy),
+			RequestHeadersPolicy:  reqHP,
+			ResponseHeadersPolicy: respHP,
 		}
 
 		if len(route.GetPrefixReplacements()) > 0 {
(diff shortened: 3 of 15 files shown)


apis/projectcontour/v1/httpproxy.go
```diff
@@ -199,6 +199,30 @@ type Service struct {
 	UpstreamValidation *UpstreamValidation `json:"validation,omitempty"`
 	// If Mirror is true the Service will receive a read only mirror of the traffic for this route.
 	Mirror bool `json:"mirror,omitempty"`
+	// RequestHeaders defines how to add or remove headers from requests routed to this service.
+	// +optional
+	RequestHeaders *HeaderAlterations `json:"requestHeaders,omitempty"`
+	// ResponseHeaders defines how to add or remove headers from responses returned from this service.
+	// +optional
+	ResponseHeaders *HeaderAlterations `json:"responseHeaders,omitempty"`
```
jpeach: This should follow the API patter established by `PathRewritePolicy`.

* The struct should be called `HeaderRewritePolicy`
* rewrites should be attached to routes, rather than services

Otherwise, this approach looks good to me

apis/projectcontour/v1/httpproxy.go
```diff
@@ -199,6 +199,30 @@ type Service struct {
...
+}
+
+// HeaderAlterations defines alterations to the headers being sent to or returned from a service.
+type HeaderAlterations struct {
+	// Add holds the header key/value pairs to add to those sent to or returned from a service.
+	// +optional
+	Add []HeaderAddition `json:"add,omitempty"`
```
jpeach: Is it possible to avoid the additional struct?
```Go
Add map[string]string `json:"add,omitempty"`
```

internal/dag/builder.go
```diff
@@ -787,6 +807,38 @@ func (b *Builder) computeRoutes(sw *ObjectStatusWriter, proxy *projcontour.HTTPP
...
+func validateHeaderAlterations(alt *projcontour.HeaderAlterations) (map[string]string, []string, error) {
+	if alt == nil {
+		return nil, nil, nil
+	}
+	add := make(map[string]string)
+
+	for _, entry := range alt.Add {
+		key := http.CanonicalHeaderKey(entry.Name)
+		_, ok := add[key]
+		if ok {
```
jpeach: nit:
```Go
if _, ok := add[key]; ok {
}
```

internal/dag/builder_test.go
```diff
@@ -6056,6 +6058,85 @@ func TestSplitSecret(t *testing.T) {
...
+		wantErr: errors.New(`Duplicate header removal: "K-Foo"`),
+	}, {
+		name: "invalid add header",
+		in: &projcontour.HeaderAlterations{
+			Add: []projcontour.HeaderAddition{{
+				Name:  "  K-Foo",
+				Value: "bar",
+			}},
+		},
+		wantErr: errors.New(`invalid add header "  K-Foo": [a valid HTTP header must consist of alphanumeric characters or '-' (e.g. 'X-Header-Name', regex used for validation is '[-A-Za-z0-9]+')]`),
```
jpeach: Is this actually the regex api machinery uses for validation? It's way too restrictive for HTTP

https://httpwg.org/specs/rfc7230.html#header.fields

internal/dag/builder.go
```diff
@@ -787,6 +807,38 @@ func (b *Builder) computeRoutes(sw *ObjectStatusWriter, proxy *projcontour.HTTPP
...
+			return nil, nil, fmt.Errorf("Duplicate header addition: %q", key)
+		}
+		if msgs := validation.IsHTTPHeaderName(key); len(msgs) != 0 {
+			return nil, nil, fmt.Errorf("invalid add header %q: %v", key, msgs)
+		}
+		add[key] = entry.Value
```
jpeach: We should make an explicit choice here whether we support envoy [% encoded header values](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#custom-request-response-headers).

internal/envoy/route.go
```diff
@@ -194,6 +193,40 @@ func UpgradeHTTPS() *envoy_api_v2_route.Route_Redirect {
 	}
 }
 
+func headerValueList(hvm map[string]string) (hvs []*envoy_api_v2_core.HeaderValueOption) {
+	for key, value := range hvm {
+		hvs = append(hvs, &envoy_api_v2_core.HeaderValueOption{
```
jpeach: If the user specified the "Host" header, should we generate a `host_rewrite`, or error out?

internal/dag/builder.go
```diff
@@ -787,6 +807,38 @@ func (b *Builder) computeRoutes(sw *ObjectStatusWriter, proxy *projcontour.HTTPP
 	return routes
 }
 
+func validateHeaderAlterations(alt *projcontour.HeaderAlterations) (map[string]string, []string, error) {
+	if alt == nil {
+		return nil, nil, nil
+	}
+	add := make(map[string]string)
+
+	for _, entry := range alt.Add {
```
jpeach: Should we handle headers with the `:` prefix explicit, or just let `validation.IsHTTPHeaderName` deal with it? I'm inclined towards the latter ...

internal/dag/builder.go
mattmoor: Interesting, I was unaware of those capabilities.  I think my personal bias would be towards having `Value` in the API hold the value of a raw string that is passed through.  This doesn't preclude adding another mutually exclusive key in the struct later for allowing these variables (e.g. `ValueWithVariables`) if the desire presents itself.

I'm happy to go either way on this, but will bias towards escaping everything for now.  thanks for catching this!

mattmoor:
@jpeach Do you want the docs issue in this repo, or is there another?  I've rebased to pick up the latest master, but separated the changes based on feedback into a second commit.

internal/envoy/route.go
jpeach: If we want to do a separate Host rewrite API, then we should disallow Host here. My preference would be to accept the Host header here and do the right thing, but then I use curl a lot :)

jpeach:
>  Do you want the docs issue in this repo, or is there another?

Please file the docs issue in this repo. The actual changes for that would go into `site/docs/master/httpproxy.md`.

internal/dag/builder.go
jpeach: In #70, there's a use case for interpolating K8s metadata to generate a linkerd header. So there's definitely good cases for interpolation, just need to talk it through and get consensus on what we want to support.

davecheney:
@mattmoor thank you for working on this. #70 is on the backlog and is not assigned to an active milestone. When were you hoping to land this? Have you talked to @michmike about priorities?

apis/projectcontour/v1/httpproxy.go
jpeach: That's a pretty interesting use case. One of the rationales for the policy structs is that we can allow them to be applied at different levels. Maybe it makes sense to allow `HeaderRewritePolicy` at both Route and Route.Service levels, though we then need to define how merge works.

internal/envoy/route.go
mattmoor: I already added logic (and tests) to exclude it in the validation logic (so a Host header should never reach here), but I can add something here too.  If we want to support host-rewriting through this API, then I'll have to dig some more about how to encoded that for Envoy, but was biasing away from scope creeping this change (and focusing on the minimum correct subset for this increment).

apis/projectcontour/v1/httpproxy.go
stevesloka: We started a `HeaderPolicy` design which deals with adding/removing/updating, etc of headers in #1982

mattmoor:
@davecheney Haven't talked to @michmike, but happy to try to connect this week.  I don't have a specific timeframe in mind. My ultimate goal is to enable writing a Contour-based Knative Ingress controller (as an alternative to Istio et al), so the sooner I can land the required pieces in a release the sooner that'll happen.  Thanks for making the connection with @michmike .

stevesloka:
@jpeach why not land the docs with the feature? Vertical slices! =)

apis/projectcontour/v1/httpproxy.go
mattmoor: re: merge, note that the docs you referred me to do talk about the ordering of the levels within Envoy.

re: #1982, sorry if I jumped the gun on this, I wasn't aware of the other issue 😅

mattmoor:
I'm happy to update docs in this PR as well, just doing my best to mold to your processes :)

jpeach:
> why not land the docs with the feature?

That's fine too. I usually try to ease a new contributor in to all the project requirements, and putting docs aside until later lets us make progress on one thing at a time.

apis/projectcontour/v1/httpproxy.go
youngnick: Yeah, to keep with the conventions, we should keep it an additional struct, agreed.

internal/envoy/route.go
youngnick: Keeping the change subset here small is good, if we are going to release this in 1.1, it's already a substantial scope increase, and we'll need to be careful about checking it and @stevesloka's change for external services for consistency.

site/docs/master/httpproxy.md
```diff
@@ -477,6 +477,42 @@ HTTPProxy weighting follows some specific rules:
...
+  name: header-manipulation
+  namespace: default
+spec:
+  virtualhost:
+    fqdn: headers.bar.com
+  routes:
+    - services:
+        - name: s1
+          port: 80
+	  requestHeaderPolicy:
```
youngnick: Looks like there's an indenting problem here... I think the red coloration means it's tabs not spaces? @mattmoor could you check? Currently, it's not aligned properly.

mattmoor: Yeah, emacs isn't smart enough about tabs vs. spaces.

internal/envoy/route.go
mattmoor: Agree on rationalizing this with @stevesloka's change, I'd like to start looking into this more and left a few comments on it.  I've added a brief design doc (following the template) to this PR (I can split it off, not sure of the process?) that tries to capture the union of the two changes.  I believe the API in this change should be what's necessary and sufficient to implement both, but leaves the Host-rewrite mechanics for @stevesloka's PR.

mattmoor:
> I like the design document, but it does collide from what I can see with Steve's work.

@youngnick For sure.  I actually wrote the design document with @stevesloka's PR in mind.  The main point of overlap between them is the API, but the mechanics diverge because these have to be expressed to envoy in fundamentally distinct ways.

I'll reach out to Mike / Steve about if/how to move this forward.

design/httpproxy-header-rewrite.md
```diff
@@ -0,0 +1,79 @@
+# Header Rewrite For HTTPProxy
```
mattmoor: This file is now defunct (in this PR) and the source of truth is: https://github.com/projectcontour/contour/pull/2066

internal/envoy/route.go
mattmoor: Ok, so @stevesloka merged the PR with the API design discussed yesterday, which should now be reflected here.  It's listed as "Draft", and I'm not sure how the lifecycle of these things progresses? 🤓

internal/contour/route.go
```diff
@@ -143,6 +143,11 @@ func (v *routeVisitor) visit(vertex dag.Vertex) {
 					routes = append(routes, &envoy_api_v2_route.Route{
 						Match:  match,
 						Action: envoy.RouteRoute(route),
+
+						RequestHeadersToAdd:     envoy.HeaderValueList(route.SetRequestHeaders, false),
```
stevesloka: nit: Should we match these names with `Set`?

internal/envoy/route.go
stevesloka: I think we're going to need to land this PR before mine. They are essentially the same, with my PR having a lot less logic. 

However, my PR does handle host rewrites as a special case, in that if you specify `Host` as the key, we set different configuration on the Envoy side.

internal/envoy/route.go
```diff
@@ -194,6 +194,44 @@ func UpgradeHTTPS() *envoy_api_v2_route.Route_Redirect {
...
+	}
+	sort.Slice(hvs, func(i, j int) bool {
+		return hvs[i].Header.Key < hvs[j].Header.Key
+	})
+	return
+}
+
+// singleSimpleCluster determines whether we can use a RouteAction_Cluster
+// or must use a RouteAction_WeighedCluster to encode additional routing data.
+func singleSimpleCluster(clusters []*dag.Cluster) bool {
```
stevesloka: Why do you need to use a weighted cluster if any of the header additions are used?

design/httpproxy-header-rewrite.md
```diff
@@ -0,0 +1,78 @@
...
+
+- Support for projecting dynamic values into header values.
+
+## Background
+
+There are a number of use cases where having this raw capability is useful, and many are called out in #70.
+
+## High-Level Design
+
+Add a new type: `HeaderRewritePolicy` that captures adding and removing headers.
```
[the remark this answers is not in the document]
mattmoor: This file's no longer in the PR, since we split it off into a separate PR.

design/httpproxy-header-rewrite.md
```diff
@@ -0,0 +1,78 @@
...
+
+```Go
+// HeaderRewritePolicy defines alterations to the headers being sent to or returned from a service.
+type HeaderRewritePolicy struct {
+	// Add holds the header key/value pairs to add to those sent to or returned from a service.
+	// +optional
+	Add []HeaderAddition `json:"add,omitempty"`
```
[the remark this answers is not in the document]
mattmoor: Should be in the latest commit.

internal/contour/route.go
mattmoor: This is Envoy's API, so it's the one thing we can't change.  The `false` parameter is what toggles append vs. set semantics.

stevesloka: Ohhhh ok you're right, sorry I overlooked that. =)

internal/envoy/route.go
mattmoor: The non-weighted version isn't expressive enough because it just takes a cluster name:
```go
type RouteAction_Cluster struct {
	Cluster string `protobuf:"bytes,1,opt,name=cluster,proto3,oneof"`
}
```

However, the `RouteAction_WeightedClusters` takes a list of clusters with additional stuff to attach.  I'd guess this is creep in the Envoy API that led to the container type being a misnomer, but it'd take some archaeology to confirm.

stevesloka: That's unfortunate that it only applied to Weighted Clusters. We had thoughts around removing weighted clusters in the future to solve different problems with healthchecks.

mattmoor:
@stevesloka I think your PR primarily dealt with pre-split (Route-level) Host rewriting, is it problematic to do it post-split (Service-level) as well?

stevesloka:
> I think your PR primarily dealt with pre-split (Route-level) Host rewriting, is it problematic to do it post-split (Service-level) as well?

@mattmoor nope, we just started mine at the Route level with the idea that we could expand it further in the future.

mattmoor:
@jpeach @stevesloka @youngnick I've rebased this on Steve's change, though for my sanity I squashed the commits before rebasing this time.

PR #2016, opened 2019-12-09
internal/dag: enforce that HTTPProxy must produce routes

### LINKED ISSUE #1988: doc: HTTPProxy must have at least one route or include defined
From #1947, the HTTPProxy routes field is documented as

> HTTPProxy must have at least one route or include defined.

This is probably not correct, since TCPProxy could be valid instead.

jpeach:
A HTTPProxy must have either a TCPProxy spec, or enough valid Routes
and Includes specs to actually emit Envouy routes. If there are no
Envoy routes, that is logically similar to having an Include that
doesn't resolve, so mark the HTTPProxy as invalid.

This fixes #1988 

### DIFF
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -514,6 +514,16 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
 		secure := b.lookupSecureVirtualHost(host)
 		addRoutes(secure, routes)
 	}
+
+	// HTTPProxy object must have either a TCPProxy spec, or
+	// some combination of Include and Route specs that result
+	// routes being generated.
+	if sw.IsValid() {
+		if proxy.Spec.TCPProxy == nil && len(routes) == 0 {
+			sw.SetInvalid("Spec does not produce any routes")
+			return
+		}
+	}
 }
 
 type vhost interface {
--- a/internal/dag/status.go
+++ b/internal/dag/status.go
@@ -94,6 +94,10 @@ func (osw *ObjectStatusWriter) SetInvalid(desc string) {
 	osw.WithValue("description", desc).WithValue("status", StatusInvalid)
 }
 
+func (osw *ObjectStatusWriter) IsValid() bool {
+	return osw.values["status"] == StatusValid
+}
+
 func (osw *ObjectStatusWriter) SetValid() {
 	switch osw.obj.(type) {
 	case *projcontour.HTTPProxy:
--- a/internal/contour/metrics_test.go
+++ b/internal/contour/metrics_test.go
@@ -890,11 +890,10 @@ func TestIngressRouteMetrics(t *testing.T) {
 			wantIR: nil,
 			wantProxy: &metrics.RouteMetric{
 				Invalid: map[metrics.Meta]int{
-					{Namespace: "roots"}: 1,
-				},
-				Valid: map[metrics.Meta]int{
+					{Namespace: "roots"}:                       1,
 					{Namespace: "roots", VHost: "example.com"}: 1,
 				},
+				Valid:    map[metrics.Meta]int{},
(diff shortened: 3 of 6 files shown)


internal/dag/builder.go
```diff
@@ -514,6 +514,16 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
 		secure := b.lookupSecureVirtualHost(host)
 		addRoutes(secure, routes)
 	}
+
+	// HTTPProxy object must have either a TCPProxy spec, or
+	// some combination of Include and Route specs that result
+	// routes being generated.
+	if sw.IsValid() {
+		if proxy.Spec.TCPProxy == nil && len(routes) == 0 {
+			sw.SetInvalid("Spec does not produce any routes")
```
davecheney: suggestion: what about something less cryptic like "HTTPProxy must have either tcpproxy or one valid route"

internal/dag/builder.go
```diff
@@ -514,6 +514,16 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
 		secure := b.lookupSecureVirtualHost(host)
 		addRoutes(secure, routes)
 	}
+
+	// HTTPProxy object must have either a TCPProxy spec, or
```
davecheney: Can this check move higher, around 507?

internal/dag/status.go
```diff
@@ -94,6 +94,10 @@ func (osw *ObjectStatusWriter) SetInvalid(desc string) {
 	osw.WithValue("description", desc).WithValue("status", StatusInvalid)
 }
 
+func (osw *ObjectStatusWriter) IsValid() bool {
```
davecheney: I'd prefer not to do this at this time. IsValid is a temporary state based on the primitive way we record status -- in the future it may become harder to determine if something is valid or not unless you know which path is being used to include it.

internal/dag/builder.go
```diff
@@ -514,6 +514,16 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
...
+	// some combination of Include and Route specs that result
+	// routes being generated.
+	if sw.IsValid() {
```
davecheney: see comment about isValid below. I would prefer that computeRoutes return a failure case to indicate that computeRoutes encountered a failure.

davecheney:
> Do you want to land this for 1.1?

Please ignore me, I see you set the milestone to 1.1

davecheney:
> Please ignore me, I see you set the milestone to 1.1

davecheney:
> Please ignore me, I see you set the milestone to 1.1

Gah, I'll get this right eventually.

If you want this for 1.1, we need an issue, not just a PR, on the 1.1 milestone please.

jpeach:
> If you want this for 1.1, we need an issue, not just a PR, on the 1.1 milestone please.

I'm ambivalent about 1.1. Happy to push this out if you prefer. The related issue is #1988

davecheney:
@jpeach I'll leave it to your discretion. /cc @michmike

internal/dag/builder.go
jpeach: Yeh, I considered that approach but the error there would be something pretty generic like `errors.New("failed to route")`. I can revisit this though.

internal/dag/status.go
jpeach: Yeh, it's ugly. I'll look into the propagating errors to remove this.

internal/dag/builder.go
jpeach: Yeh that makes sense if `computeRoutes` returns an error.

stevesloka:
@jpeach is this ready for re-review after @davecheney's comments?

Issue #2017, opened 2019-12-09
Certgen: provide parameter for setting expiry date for gRPC certificates


tsaarni:
Currently certgen generates gRPC certificates with fixed one year expiration period.  Contour does not yet support rotation of gRPC certificates. It is expected that users will handle this.

**Assumption:** Many users will miss the documentation and unknowingly set a "time bomb" which will cause an incident in 365 days.

**Proposal 1:** add a parameter to certgen for users to set the expiration date for the gRPC certificates.
**Proposal 2:** as long as there is no automated enrollment, use default expiration period that will NOT expire in any reasonable time.

Certificates are preferably rotated but due to lack of automation at this point, there is a high risk that user will not be prepared for it either.  In that case, never expiring certificate is better than expired certificate. Added security risk is negligible.

youngnick:
You make a fair point, and I think that Proposal 2 is probably better for now. I've raised #2020 to cover us writing documentation about the rotation process, as the first step towards deciding if it's automatable.

Envoy supports hot reload for lots of things, but the xDS connection is a bit tricky as it's possible to break the connection, rendering the Envoy un-updateable without a restart. So any automation of this rotation will need to be carefully tested.

jpeach:
Sorry for the inconsistent advice @tsaarni, but we now agree that Proposal 1 is preferable :)

PR #2023, opened 2019-12-10
certgen: parametrize gRPC certificate expiry

tsaarni:
Introduced new parameter `--certificate-lifetime` to set the validity duration
of gRPC certificates for the TLS connection between Contour and Envoy.

Closes: #2017

> Commit message from old proposal:

> Currently there is no automatic rotation of gRPC certificates for the TLS
> connection between Contour and Envoy.  It is safer to issue certificates
> that will never expire than to depend on all users to manage life-cycle of
> these certificates.
> 
> Closes: #2017
> 
> Signed-off-by: Tero Saarni <tero.saarni@est.tech>

### DIFF
--- a/cmd/contour/certgen.go
+++ b/cmd/contour/certgen.go
@@ -35,8 +35,8 @@ func registerCertGen(app *kingpin.Application) (*kingpin.CmdClause, *certgenConf
 	certgenApp.Flag("incluster", "use in cluster configuration.").BoolVar(&certgenConfig.InCluster)
 	certgenApp.Flag("kubeconfig", "path to kubeconfig (if not in running inside a cluster)").Default(filepath.Join(os.Getenv("HOME"), ".kube", "config")).StringVar(&certgenConfig.KubeConfig)
 	certgenApp.Flag("namespace", "Kubernetes namespace, used for Kube objects").Default("projectcontour").Envar("CONTOUR_NAMESPACE").StringVar(&certgenConfig.Namespace)
+	certgenApp.Flag("expirydays", "Duration in days how long the certificates shall be valid").Default("365").UintVar(&certgenConfig.ExpiryDays)
 	certgenApp.Arg("outputdir", "Directory to output any files to").Default("certs").StringVar(&certgenConfig.OutputDir)
-
 	return certgenApp, &certgenConfig
 }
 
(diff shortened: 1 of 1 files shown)


tsaarni:
@davecheney thanks for responding.

Security aspects are related to e.g.:
1. probability of private key being compromised: stolen or leaked e.g. due to a vulnerability
2. algorithm's security

Key compromise (1) becomes more probable when time passes.  Shorter expiration period narrows the window of opportunity of exploiting the compromised key.  However between 1 or 100 years, there is no considerable difference in that risk. Attacker is likely to use the compromised keys within days.

There is no answer to when key becomes too weak (2). For signing, RSA with 2048 bit key is minimum recommended by NIST as of today.  Changing to longer key length can be justifiable as well as switching from RSA to ECDSA.

Recommendations rely on the ability of having mechanism in place to replace the certificates.  The proposal in this PR and #2017 is based on the assumption that there is no feasible mechanism in place as of today. Users have a "time bomb" in their clusters. Average user of `contour certgen` might not be aware of the  internal certificates for xDS, they might not be prepared to do rotation, nor troubleshooting when their traffic stops working due to expired xDS gRPC certificate. 

That brings another viewpoint: the risk and consequence of loss of protection vs loss of availability. Is it reasonable that the default deployment stops working at a certain date when user misses the renewal? Can it be justified that user is better off with no traffic as the security risk has risen too high? This surely can be true for some deployments - however I believe even in those cases users first need a feasible way to rotate automatically without interrupting user plane traffic.  These users might not even care that much about contour certgen, but instead run their own CAs for service-to-service TLS.

For average user who is not prepared for yearly manual maintenance nor traffic interruption, I still think that certificates outliving ANY feasible deployment is the best option, until we can provide automatic rotation for them.

tsaarni:
I had also another option in #2017: adding command line parameter to `contour certgen` for setting the expiration period.  Would you be willing to accept that option rather than changing the default?

jpeach:
We discussed this a bit more and agreed that we should focus on the certificate rotation methodology. In the interim, we will take a patch to add a lifetime option to the certain subcommand since that can be helpful to users now as well as once the rotation is working.

tsaarni:
Thank you for responding!  

Hope it is ok to reuse this same PR, I have updated the title and content accordingly.  I have removed the change of default expiry from 1 year to 100 years (initial proposal).  The patch now contains a new parameter `--expirydays` with the default value of the currently used 365 days. 

Please review and request any changes.

tsaarni:
Thanks for your comments!  

I renamed the parameter and added a comment about future use of this flag for Duration string.

As a side note regarding Go Duration and bigger time units (such as days, months, years): there was interesting discussion [here](https://bit.ly/2T0c7Ag) that I bumped into previously.  Of course nothing prevents implementing own parser and definitions for those time units.

PR #2046, opened 2019-12-12
internal/envoy: Header policy

### LINKED ISSUE #1982: External Name Host Rewrite

stevesloka:
Fixes #1982 by adding a `headerPolicy` and allowing for the host header to be replaced during an upstream proxy via an externalName service.


stevesloka:
I need to write some docs, but worth someone having a look.

internal/dag/policy.go
```diff
@@ -34,6 +34,22 @@ func retryPolicy(rp *projcontour.RetryPolicy) *RetryPolicy {
...
 }
 
+func headerPolicy(policy *projcontour.HeaderPolicy) *HeaderPolicy {
+	if policy == nil {
+		return nil
+	}
+
+	for _, rp := range policy.ReplaceHeader {
+		// Only host rewrite is currently supported
+		if rp.Key == "Host" {
```
davecheney: Could you use https://godoc.org/strings#EqualFold here to compare the header key case insenstively?

internal/envoy/auth.go
```diff
@@ -47,13 +47,18 @@ var (
...
 // additional ALPN protocols can be provided.
-func UpstreamTLSContext(ca []byte, subjectName string, alpnProtocols ...string) *envoy_api_v2_auth.UpstreamTlsContext {
+func UpstreamTLSContext(ca []byte, subjectName string, sni string, alpnProtocols ...string) *envoy_api_v2_auth.UpstreamTlsContext {
 	context := &envoy_api_v2_auth.UpstreamTlsContext{
 		CommonTlsContext: &envoy_api_v2_auth.CommonTlsContext{
 			AlpnProtocols: alpnProtocols,
 		},
 	}
 
+	// If SNI is specified, set on the Upstream context
```
davecheney: Is this what we were asked to do? I though the request was "I want to change the host header"

stevesloka: In my testing of the feature, if I proxy to an upstream that is secure TLS, then Envoy won't proxy and returns 503 to the caller. No, we weren't directly asked for this, but @jpeach and I found that setting this allows the proper proxying.

davecheney: Ok, thanks for explaining.

internal/dag/policy.go
stevesloka: 👍

internal/envoy/auth.go
stevesloka: 👍

apis/projectcontour/v1/httpproxy.go
```diff
@@ -294,6 +297,20 @@ type LoadBalancerPolicy struct {
 	Strategy string `json:"strategy,omitempty"`
 }
 
+// HeaderPolicy defines how headers are managed during forwarding
+type HeaderPolicy struct {
+	// ReplaceHeader defines how headers will be replaced during forwarding
+	ReplaceHeader []ReplaceHeader `json:"replaceHeader,omitempty"`
```
davecheney: crd question, this is not marked +optional, does that mean that the validation will enforce that iff `headerPolicy` is present, there must be at least one `replaceHeader` entry?

internal/envoy/auth.go
```diff
@@ -47,13 +47,18 @@ var (
...
+	if len(sni) > 0 {
+		context.Sni = sni
```
davecheney: You can skip this check; if len(sni) == 0. then context.Sni = sni the same as context.Sni = ""

apis/projectcontour/v1/httpproxy.go
```diff
@@ -152,6 +152,9 @@ type Route struct {
 	//
 	// +kubebuilder:validation:Optional
 	PathRewrite *PathRewritePolicy `json:"pathRewritePolicy,omitempty"`
+	// The policy for managing headers during proxying
+	// +optional
+	HeaderPolicy *HeaderPolicy `json:"headerPolicy,omitempty"`
```
davecheney: Sorry for bringing this up late. This field needs to be called `requestheaderpolicy` so we can add `responseHeaderPolicy` later.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -294,6 +297,20 @@ type LoadBalancerPolicy struct {
 	Strategy string `json:"strategy,omitempty"`
 }
 
+// HeaderPolicy defines how headers are managed during forwarding
+type HeaderPolicy struct {
+	// ReplaceHeader defines how headers will be replaced during forwarding
+	ReplaceHeader []ReplaceHeader `json:"replaceHeader,omitempty"`
+}
+
+// ReplaceHeader defines how a header can be replaced during forwarding
```
jpeach: We need to define "replaced".

apis/projectcontour/v1/httpproxy.go
jpeach: I suggest we call this `RequestHeadersRewrite` to be consistent with `PathRewrite`.

apis/projectcontour/v1/httpproxy.go
jpeach: Should we pluralize to `ReplaceHeaders`?

internal/dag/dag.go
```diff
@@ -113,6 +113,9 @@ type Route struct {
 
 	// Mirror Policy defines the mirroring policy for this Route.
 	MirrorPolicy *MirrorPolicy
+
+	// HeaderPolicy defines how headers are managed during forwarding
+	HeaderPolicy *HeaderPolicy
```
jpeach: FYI, to mirror API changes here ...

internal/envoy/route.go
```diff
@@ -77,6 +78,13 @@ func RouteRoute(r *dag.Route) *envoy_api_v2_route.Route_Route {
 		RequestMirrorPolicy: mirrorPolicy(r),
 	}
 
+	// Check for host header policy and set if found
+	if val := hostReplaceHeader(r.HeaderPolicy); len(val) > 0 {
```
jpeach: Consider `val != ""` for consistency

apis/projectcontour/v1/httpproxy.go
davecheney: Maybe for the restricted case we're introducing now; we only allow people to replace the `Host: header` which is assumed to be present (nobody look at HTTP/1.0), thus replace means _replace unconditionally_. I'm thinking that in the future if there needs to be a specific "replace only if present, otherwise don't touch", then later we can add `replaceIfPresent`.

apis/projectcontour/v1/httpproxy.go
stevesloka: @jpeach but this isn't just a rewrite. In the future, we could add & remove headers.

apis/projectcontour/v1/httpproxy.go
stevesloka: @davecheney I found in testing that the optional doesn't really matter. I added an empty `headerPolicy`, then a `headerPolicy` with empty`replaceHeaders` and there weren't any validation errors. Let me look to see if I can get proper errors in there.

stevesloka: @jpeach I think that's a good idea. Matches what we did before with `Conditions`, etc.

jpeach:
> On 14 Dec 2019, at 1:51 am, Steve Sloka <notifications@github.com> wrote:
> 
> 
> In apis/projectcontour/v1/httpproxy.go:
> 
> > @@ -152,6 +152,9 @@ type Route struct {
>  	//
>  	// +kubebuilder:validation:Optional
>  	PathRewrite *PathRewritePolicy `json:"pathRewritePolicy,omitempty"`
> +	// The policy for managing headers during proxying
> +	// +optional
> +	HeaderPolicy *HeaderPolicy `json:"headerPolicy,omitempty"`
> 
> @jpeach but this isn't just a rewrite. In the future, we could add & remove headers.

It's a rewrite in the sense that we are rewriting the HTTP request headers. That could mean adding, removing, arbitrary value manipulations, whatever. Any changes counts as rewrite.

So I think that the policy struct should be called `HeaderRewritePolicy`. We will end up with optional instances of this policy for requests and responses. With this nomenclature, I think it makes sense for the fields to be named "additions", "removals" and "replacements".

apis/projectcontour/v1/httpproxy.go
```diff
@@ -294,6 +296,25 @@ type LoadBalancerPolicy struct {
...
 
+// RequestHeaderPolicy defines how headers are managed during forwarding
+type RequestHeaderPolicy struct {
+	// ReplaceHeaders defines how headers will be replaced during forwarding
+	// +kubebuilder:validation:Optional
+	ReplaceHeaders []ReplaceHeader `json:"replaceHeaders,omitempty"`
+}
+
+// ReplaceHeaders defines how a header can be replaced during forwarding
+type ReplaceHeader struct {
```
jpeach: Suggest that we call this "HeaderValue"

apis/projectcontour/v1/httpproxy.go
```diff
@@ -294,6 +296,25 @@ type LoadBalancerPolicy struct {
...
+	// Key represents a key of a header
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	Key string `json:"key"`
```
jpeach: Suggest that we call this "Name".

apis/projectcontour/v1/httpproxy.go
stevesloka: I like that better, thanks!

apis/projectcontour/v1/httpproxy.go
stevesloka: I don't feel super strong either way, was just going off of @davecheney's suggestion.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -294,11 +296,30 @@ type LoadBalancerPolicy struct {
...
+	// +kubebuilder:validation:MinLength=1
+	Value string `json:"value"`
+}
+
 // UpstreamValidation defines how to verify the backend service's certificate
 type UpstreamValidation struct {
 	// Name of the Kubernetes secret be used to validate the certificate presented by the backend
 	CACertificate string `json:"caSecret"`
-	// Key which is expected to be present in the 'subjectAltName' of the presented certificate
+	// Name which is expected to be present in the 'subjectAltName' of the presented certificate
```
mattmoor: SubjectName?

apis/projectcontour/v1/httpproxy.go
```diff
@@ -152,13 +152,15 @@ type Route struct {
 	//
 	// +kubebuilder:validation:Optional
 	PathRewrite *PathRewritePolicy `json:"pathRewritePolicy,omitempty"`
+	// The policy for managing headers during proxying
+	// +optional
+	RequestHeaderPolicy *RequestHeaderPolicy `json:"requestHeaderPolicy,omitempty"`
```
mattmoor: Should the type here simply be `HeaderPolicy` or `HeaderRewritePolicy`?  What about the contents is request-specific?  I could see the validation being different, but that's done separately, and could still be made context sensitive in other ways (where it matters).

site/docs/master/httpproxy.md
```diff
@@ -776,6 +776,37 @@ spec:
...
+  name: header-rewrite-example
+spec:
+  virtualhost:
+    fqdn: header.bar.com
+  routes:
+  - services:
+    - name: s1
+      port: 80
+    headerRequestPolicy:
+      replaceHeader:
```
mattmoor: qualifying these fields with `Header` feels redundant when seen in context like this.

site/docs/master/httpproxy.md
```diff
@@ -776,6 +776,37 @@ spec:
         replacement: /app
 ```
 
+### Header Policy
+
+HTTPProxy supports rewriting the `Host` header after first handling a request and before proxying to an upstream service.
+A common use-case for this is to use Contour to proxy to a resource outside the cluster referenced by an `externalName` service.
+
+The `headerRequestPolicy` supports a `replaceHeader` that currently only supports rewriting `Host` headers defined via a `key` and `value`.
```
mattmoor: note: in code this is `requestHeaderPolicy` (the first two word are transposed)

site/docs/master/httpproxy.md
```diff
@@ -776,6 +776,37 @@ spec:
...
+metadata:
+  name: header-rewrite-example
+spec:
+  virtualhost:
+    fqdn: header.bar.com
+  routes:
+  - services:
+    - name: s1
+      port: 80
+    headerRequestPolicy:
```
mattmoor: here as well.

site/docs/master/httpproxy.md
mattmoor: I think there's also at least a suggestion that `key` -> `name`.

site/docs/master/httpproxy.md
```diff
@@ -776,6 +776,37 @@ spec:
...
+      replaceHeader:
+      - name: Host
+        value: external.dev
+```
+
+#### Proxy to external resource
+
+To proxy to another resource outside the cluster (e.g. A hosted object store bucket for example), configure that external resource in a service type `externalName`.
+Then define a `headerRequestPolicy` which replaces the `Host` header with the value of the external name service defined previously.
```
mattmoor: here as well

stevesloka:
I think this is ready to go whenever we want to review/merge and should also match @mattmoor's changes.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -152,13 +152,15 @@ type Route struct {
 	//
 	// +kubebuilder:validation:Optional
 	PathRewrite *PathRewritePolicy `json:"pathRewritePolicy,omitempty"`
+	// The policy for managing headers during proxying
+	// +optional
```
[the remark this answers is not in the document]
stevesloka: There are both mixed throughout. I can use kubebuilder but should update everything else to be consistent.

internal/featuretests/headerpolicy_test.go
```diff
@@ -0,0 +1,203 @@
...
+)
+
+func TestHeaderPolicy_ReplaceHeader_HTTProxy(t *testing.T) {
+	rh, c, done := setup(t)
+	defer done()
+
+	rh.OnAdd(&v1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "svc1",
+			Namespace: "default",
```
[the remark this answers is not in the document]
stevesloka: ok, all these little methods all over to me make it all the more confusing to undertand what we're actually testing since the meat of **what** the test is hidden across a dozen helpers, but I'll follow along.

internal/featuretests/envoy.go
```diff
@@ -79,6 +79,15 @@ func routePrefix(prefix string, headers ...dag.HeaderCondition) *envoy_api_v2_ro
 	})
 }
 
+func routeHostRewrite(cluster, hostRewrite string) *envoy_api_v2_route.Route_Route {
```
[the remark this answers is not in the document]
stevesloka: This doesn't return a new Host Name but an envoy Route_Route. I matched the pattern to how we do all the other one's (e..g `routePrefix, `routeCluster`, etc)

internal/featuretests/headerpolicy_test.go
```diff
@@ -0,0 +1,203 @@
...
+		Resources: resources(t,
+			envoy.RouteConfiguration("ingress_http",
+				envoy.VirtualHost("hello.world",
+					&envoy_api_v2_route.Route{
+						Match:  routePrefix("/"),
+						Action: routeCluster("default/svc1/80/da39a3ee5e"),
+					},
+				),
+			),
+			envoy.RouteConfiguration("ingress_https"),
```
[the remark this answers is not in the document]
stevesloka: 👍

internal/envoy/route.go
```diff
@@ -116,6 +124,18 @@ func mirrorPolicy(r *dag.Route) *envoy_api_v2_route.RouteAction_RequestMirrorPol
 	}
 }
 
+func hostReplaceHeader(hp *dag.RequestHeadersPolicy) string {
```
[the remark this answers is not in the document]
stevesloka: I made this super simple for the host use-case. We can revisit with the other PR since adding all this I think is overkill for my feature.

internal/envoy/route.go
```diff
@@ -116,6 +124,18 @@ func mirrorPolicy(r *dag.Route) *envoy_api_v2_route.RouteAction_RequestMirrorPol
...
+	if hp == nil {
+		return ""
+	}
+	for _, rp := range hp.Set {
+		if strings.EqualFold(rp.Key, "Host") {
```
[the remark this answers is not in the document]
stevesloka: Sure, but this is a special use-case. I don't see the value of reformatting a specific special string that we're looking for.

stevesloka:
> Since this contains changes to the API, you will also need to run "make render-refdocs".

I think `make check` should find this for me and yell if I don't do it. Honestly, folks are going to forget (or even know) that they should run it.

stevesloka:
Also @jpeach I ran that on my branch and it outputs nothing: 

```
$ make render-refdocs                                                       
make: Nothing to be done for `render-refdocs'.
```

jpeach:
On Dec 22, 2019, at 4:18 AM, Steve Sloka <notifications@github.com> wrote:


I think make check should find this for me and yell if I don't do it. Honestly, folks are going to forget (or even know) that they should run it.

Yep it is supposed to. The makefile is checking time stamps to see whether it needs to run the generator because it takes a while. Maybe I didn’t get that right :-/

internal/envoy/route.go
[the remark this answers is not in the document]
jpeach: It's laying down the APIs and infrastructure for the general case. We can make the change later, but to me it seems that we might as well do it now since we have a good idea of where we want this code to end up.

internal/featuretests/envoy.go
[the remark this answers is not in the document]
jpeach: Sorry, I meant that the `hostRewrite` parameter would be clearer if it was called `newHostName` :)

jpeach:
@stevesloka The docs should generate when files in "apis/projectcontour/*/*.go" are updated. There's no general way to know when to do this and the generator is pretty slow, so I'm just using make dependencies, which rely on timestamps. Depending on how you rebased and merged, the timestamps might not be updated to reflect the changes in this PR (this should only happen for this specific PR because the makefile targets weren't there when you started).

If you touch something in "apis/projectcontour/*/*.go" to update the timestamps, then "make render-refdocs" should pick up the change and rebuild the API docs. 

Sorry for the hassle :)

stevesloka:
> If you touch something in "apis/projectcontour//.go" to update the timestamps, then "make render-refdocs" should pick up the change and rebuild the API docs.

Let's have a look at this, for me, when I runt the makefile task I get "Nothing to be done..", but if I run the generator by hand (e.g. `./hack/generate-refdocs.sh`) it works fine.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -294,6 +296,25 @@ type LoadBalancerPolicy struct {
...
 }
 
+// HeadersPolicy defines how headers are managed during forwarding
+type HeadersPolicy struct {
+	// Set defines how headers will be configured
+	// +kubebuilder:validation:Optional
+	Set []HeaderValue `json:"set,omitempty"`
+}
+
+// Set defines how a header can be replaced during forwarding
```
jpeach: Update this doc comment?

apis/projectcontour/v1/httpproxy.go
```diff
@@ -294,6 +296,25 @@ type LoadBalancerPolicy struct {
 	Strategy string `json:"strategy,omitempty"`
 }
 
+// HeadersPolicy defines how headers are managed during forwarding
+type HeadersPolicy struct {
+	// Set defines how headers will be configured
```
jpeach: Let's document the semantics of `Set` here? Maybe something like this ...

```
Set specifies a list of HTTP header values that will be set in the HTTP header. If the corresponding HTTP header name already exists in the HTTP message, it will be replaced, otherwise a new HTTP header will be added.
```

internal/dag/dag.go
```diff
@@ -113,6 +113,9 @@ type Route struct {
 
 	// Mirror Policy defines the mirroring policy for this Route.
 	MirrorPolicy *MirrorPolicy
+
+	// HeadersPolicy defines how headers are managed during forwarding
```
jpeach: "RequestHeadersPolicy defines ..."

internal/featuretests/headerpolicy_test.go
```diff
@@ -0,0 +1,250 @@
...
+		),
+		TypeUrl: routeType,
+	})
+
+	// Unsupported header (i.e. not `Host`)
+	proxy2 := &projcontour.HTTPProxy{
```
jpeach: Suggest that you don't make a local variable unless you end up using it. So this (and places below) could just be:
```Go
rh.OnAdd(&projcontour.HTTPProxy{
        ...
})
```

site/docs/master/httpproxy.md
```diff
@@ -776,6 +776,37 @@ spec:
...
+Finally, if the upstream service is served over TLS, annotate the external name service with: `projectcontour.io/upstream-protocol.tls: 443,https` assuming your service had a port 443 & name `https`.
+
```
jpeach: Should this section go within the "ExternalName" section below?

apis/projectcontour/v1/httpproxy.go
stevesloka: I'll update the comment to be more clear. Also, I want to be careful that we don't leak too much logic in the API spec.

PR #2047, opened 2019-12-12
internal/envoy: bump the max Envoy regex program size

### LINKED ISSUE #2043: internal/envoy: Regex "MaxProgramSize" too small
When deploying HTTPProxy resources which use `Contains` header matcher conditions, there's an error getting thrown by Envoy:

```
ERRO[0291] regex '.*Chrome.*' RE2 program size of 38 > max program size of 10. Increase configured max program size if necessary. 
```

I didn't think much of this error, but working on a different feature for Contour I found that at times all my routes would disappear from Envoy's config and I wasn't able to get them back. I found that removing the "header contains" Conditions from my Proxies, Envoy started working well again. 

Other symptoms I've seen are a lack of xDS updates between Contour<>Envoy.  At times, I was able to check Contour's RDS (e.g. `contour cli rds`) and I would see the proper config, however, it was not mirrored in Envoy's admin config. 

At the moment, I've taken out the `MaxProgramSize` from being configured (https://github.com/projectcontour/contour/blob/master/internal/envoy/regex.go#L27) and am using Envoy's default of `100`.

jpeach:
There's no obvious way to know what a reasonable regex program
size is (and that's not really what the RE2 program size number
means).  However, Envoy makes us send something here and if it's
too low, it will NACK the update and Contour will not handle that
gracefully. So, we send a large number that should be enough for
all reasonable regex cases.

This fixes #2043.

### DIFF
--- a/internal/envoy/regex.go
+++ b/internal/envoy/regex.go
@@ -18,13 +18,23 @@ import (
 	"github.com/projectcontour/contour/internal/protobuf"
 )
 
+// maxRegexProgramSize is the default value for the Envoy regex max
+// program size tunable. There's no way to really know what a good
+// value for this is, except that the RE2 maintainer thinks that 100
+// is low. As a rule of thumb, each '.*' expression costs about 15
+// units of program size. AFAIK, there's no obvious correlation
+// between regex size and execution time.
+//
+// https://github.com/envoyproxy/envoy/pull/9171#discussion_r351974033
+const maxRegexProgramSize = 1000
+
 // SafeRegexMatch retruns a matcher.RegexMatcher for the supplied regex.
 // SafeRegexMatch does not escape regex meta characters.
 func SafeRegexMatch(regex string) *matcher.RegexMatcher {
 	return &matcher.RegexMatcher{
 		EngineType: &matcher.RegexMatcher_GoogleRe2{
 			GoogleRe2: &matcher.RegexMatcher_GoogleRE2{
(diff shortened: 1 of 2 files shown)


internal/envoy/regex.go
```diff
@@ -18,13 +18,23 @@ import (
...
+const DefaultMaxRegexProgramSize = 1000
+
 // SafeRegexMatch retruns a matcher.RegexMatcher for the supplied regex.
 // SafeRegexMatch does not escape regex meta characters.
 func SafeRegexMatch(regex string) *matcher.RegexMatcher {
 	return &matcher.RegexMatcher{
 		EngineType: &matcher.RegexMatcher_GoogleRe2{
 			GoogleRe2: &matcher.RegexMatcher_GoogleRE2{
-				MaxProgramSize: protobuf.UInt32(uint32(len(regex))),
+				MaxProgramSize: protobuf.UInt32(uint32(DefaultMaxRegexProgramSize)),
```
davecheney: nit: you don't need to convert an untyped constant to a uint32, the compiler will do that for you 
```
MaxProgramSize: protobuf.UInt32(DefaultMaxRegexProgramSize)
```

Please don't call this `DefaultMaxRegexProgramSize`. It's not the default, it might be our default, but even then I don't think calling it default is accurate, `maxProgramSize` is fine.

jpeach: Thanks, fixed!

PR #2052, opened 2019-12-13
Add a script to build a kind development cluster.

jpeach:


jpeach:
>  I worry sometimes about how fragile these types of scripts become over time.

Yes I share this concern. Scripts can easily bit rot and dev set up can become unused. However, the converse of this problem is that you can easily lose hours debugging up dev environments that don't work because of some simple missing step.

> I think a good first step is getting Kustomize (or Helm)

Strongly +1 on this.

Issue #2054, opened 2019-12-13
Support traffic splitting when websockets are enabled


mattmoor:
I believe the underlying issue from https://github.com/projectcontour/contour/issues/732 has been resolved for some time.  So I believe that this block can now be removed:
https://github.com/projectcontour/contour/blob/74e0f4650f95df73d8730e48171708923f7b1540/internal/dag/builder.go#L674-L677

mattmoor:
I tested removing this and another block like it at HEAD and was able to verify that a basic websocket worked fine with the split in place.  I don't have the context about the original failure mode, but at least superficially this seems to work.

I deployed [this](https://github.com/n3wscott/sockeye) with a 50/50 split to another app using some in-progress Knative integration which unconditionally enables websockets.

cc @davecheney (in case he has more context on the original issue)

davecheney:
@mattmoor when @stevesloka look for a resolution for #732 around the Contour 1.0.0 timeframe the general agreement was web sockets + weighted clusters were still broken, at least as we were targeting envoy 1.11 at the time. I'm not sure if its been fixed in Envoy 1.12, any advice would be appreciated.

mattmoor:
What was the failure mode?  I tried last night with a 50/50 traffic split (on 1.12.2) and was able to use them, but I'm not sure if they were just half broken or totally broken.

stevesloka:
We have other issues around multicluster and health checks, but looks like the envoy issue is resolved so we can remove the internal restrictions in contour.

mattmoor:
I can try to pick this up, if folks can help guide what needs fixing.

PR #2058, opened 2019-12-16
Remove the restriction on websockets with weighted clusters.

mattmoor:
Envoy has fixed the issue related to using websockets with weighted clusters, so this drops the exclusion which kept those clusters out of the DAG.

Fixes: https://github.com/projectcontour/contour/issues/2054

cc @stevesloka


stevesloka:
This LGTM @mattmoor, just have a failing build now with tests it looks like.

mattmoor:
@stevesloka been looking at it (a missed unit test), and I hope it's fixed now.  Have to run to a meeting, but if it fails again I'll keep digging.

PR #2062, opened 2019-12-17
This makes the logic in internal/contour use envoy.Route.

mattmoor:
Unlike other utilities in `internal/envoy`, the `Route` method didn't take it's `dag.Route` equivalent, nor was it being used (outside of unit tests).  This changes the signature a bit to enable it to be used by `internal/contour` and adapts the table testing to the slightly more verbose inline struct.

This is intended to clean things up a bit in advance of trying to add more route-level configuration, which must be encoded on the envoy Route resource (e.g. route-level request headers).


internal/envoy/route.go
```diff
@@ -33,10 +33,10 @@ func Routes(routes ...*envoy_api_v2_route.Route) []*envoy_api_v2_route.Route {
 }
 
 // Route returns a *envoy_api_v2_route.Route for the supplied match and action.
-func Route(match *envoy_api_v2_route.RouteMatch, action *envoy_api_v2_route.Route_Route) *envoy_api_v2_route.Route {
+func Route(route *dag.Route) *envoy_api_v2_route.Route {
```
jpeach: I think that this change is a bit awkward because the API signature implies that for any kind of `dag.Route`, the correct `envoy_api_v2_route.Route` form will be returned, but that's not what happens. I'd prefer a change that names the API more specifically.

mattmoor: If I'm not mistaken the HTTP -> HTTPS upgrade path is the main exception to this, or are you thinking about something else?

mattmoor: Do you have a name in mind?

jpeach:
On Dec 19, 2019, at 8:12 AM, Steve Sloka <notifications@github.com> wrote:


@stevesloka approved this pull request.

This LGTM if @jpeach<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fjpeach&data=02%7C01%7Cjpeach%40vmware.com%7C406b3824019246754e3108d783ff0bf1%7Cb39138ca3cee4b4aa4d6cd83d9dd62f0%7C0%7C0%7C637123003762036224&sdata=LjBfSItHuzIpntoSWuuUSNI0tabEnUa4%2BUIgvoSFwCY%3D&reserved=0> is happy. =)

I’ll review and suggest a name tomorrow

internal/envoy/route.go
jpeach: I thought about this more and I think that for now we should just remove both `envoy.Route` and `envoy.Routes`. The latter isn't used, and I'd argue that the former doesn't really provide much value in its current form.

I'm happy to remove it now, and see whether we want to bring in back in a different form later.

mattmoor: ok, these are gone :)

Issue #2056, opened 2019-12-13
Consider the use of port name to guide protocol selection. 


mattmoor:
To support http/2 delivery today the service referenced from Contour's HTTPProxy must be annotated with something like:
```yaml
annotations:
  contour.heptio.com/upstream-protocol.h2c: "80,http"
```

This style of annotation-based configuration unfortunately requires us (Knative) to bake in knowledge of how to configure this for every networking implementation with which we integrate.

Istio [developed a convention](https://istio.io/docs/ops/configuration/traffic-management/protocol-selection/) for avoiding this that we adopted, which isn't tied to one implementation and a few others have adopted.

Essentially by naming the port a certain way, it determines what protocol should be used by default to speak to that port.  In our case, when we synthesize http/1.1 services we have:
```yaml
  ports:
  - name: http   # I am http/1.1
    port: 80
    protocol: TCP
    targetPort: 8012
```

When we synthesize http/2 services we have:
```yaml
  ports:
  - name: http2  # I am http/2
    port: 81
    protocol: TCP
    targetPort: 8013
```

At least so far, this has proven to be a relatively uncontentious middle ground, but I wanted to open this issue to start a conversation about whether adopting this (or something similar) in Contour might be feasible, and if not what other alternatives might be possible.

stevesloka:
Back when we were designing IngressRoute/v2 I thought it would be interesting to change the `name` field to the protocol. So you'd have something like this:

```
  ports:
  - http: serviceName   # I am http/1.1
    port: 80
    protocol: TCP
    targetPort: 8012
```


```
  ports:
  - http2: serviceName   # I am http/2 (This might also be h2)
    port: 80
    protocol: TCP
    targetPort: 8012
```

mattmoor:
I could also see the `Service` section of the `HTTPProxy` being used to direct the protocol semantics as well and this would allow the controller that reconciles the Knative networking abstraction => `HTTPProxy` to be what's sensitive to the convention and direct the `HTTPProxy` on when to use `http` vs. `http2`.  I biased towards this approach because I recall seeing a comment from Dave on wanting this configuration to live K8s `Service`-side.

Part of the problem here is that the networking controller that produces `HTTPProxy` _has_ to be Contour-specific / -aware, but at least today the controller that produces the K8s services that we route to encodes no real knowledge of Istio / Gloo / Ambassador / Contour / Kourier / ...

[...]

Issue #1962, opened 2019-11-29
Feature Request: Allow "protocol" Be Defined in route.services.service (Enable Upstream TLS)


ryanelian:
**Please describe the problem you have**

Continuing discussion from https://github.com/projectcontour/contour/issues/406

Currently enabling Upstream TLS requires annotations to be added to the backend `Service` manifest:

```
  annotations:
    projectcontour.io/upstream-protocol.tls: "443,https"
```

This causes some maintainability and usability issues, because the annotation and the Contour ingress object are in two different places. (Despite the annotation sole purpose is for configuring Contour ingress!)

And since one of the goals of doing custom CRD (`HTTPProxy`) instead of using default `Ingress` Kubernetes object is to reduce [*"an explosion of annotations to express missing properties of HTTP routing"*](https://github.com/projectcontour/contour/blob/master/design/ingressroute-design.md),

The `protocol` should be able to be defined in the `route.services.service` of a `HTTPProxy`:

```yaml
apiVersion: projectcontour.io/v1
kind: HTTPProxy
metadata:
  name: kubernetes-dashboard-proxy
  namespace: kubernetes-dashboard
spec:
  virtualhost:
    fqdn: dashboard.kubernetes.com
    tls:
      secretName: kubernetes-dashboard-tls
  routes:
    - conditions:
      - prefix: /
      services:
        - name: kubernetes-dashboard
          protocol: https # <--- FEATURE REQUEST!
          port: 443
```

davecheney:
Thank you for raising this issue. Moving to the backlog for prioritisation, /cc @michmike

ryanelian:
Excuse me, I am having trouble using the new `protocol: https` feature.

![image](https://user-images.githubusercontent.com/8892187/73075288-2c06c080-3eee-11ea-9f5b-a6b6d2e28c77.png)

What should I do?

```
apiVersion: projectcontour.io/v1
kind: HTTPProxy
metadata:
  name: kubernetes-dashboard-proxy
  namespace: kubernetes-dashboard
spec:
  virtualhost:
    fqdn: dashboard.cloud.xxx.com
    tls:
      secretName: kubernetes-dashboard-tls
  routes:
    - conditions:
      - prefix: /
      services:
        - name: kubernetes-dashboard
          # protocol: https
          port: 443
```

ryanelian:
OK I found the problem after reading the source code:

https://github.com/projectcontour/contour/releases/tag/v1.1.0

[...]

PR #2067, opened 2019-12-18
Adopt port-name heuristic for protocol selection.

mattmoor:
This adopts the convention outlined [here](https://istio.io/docs/ops/configuration/traffic-management/protocol-selection/) for selecting the protocol to use for speaking to particular ports of a service.  knative/serving uses a subset of this convention (`http`, `http2`) to convey the user's intent.

Fixes: https://github.com/projectcontour/contour/issues/2056


mattmoor:
If I'm not mistaken, I think #1962 is generally preferred.

PR #2069, opened 2019-12-18
docs: add generated API reference

jpeach:
Add some site scripts to generate API reference documentation for
the Contour CRDs. Unfortunately, we need to vendor the Go templates
from the documentation generator in order to add enough CSS to make
the results attractive.

Note, that if running the generator has the side-effect of updating
Go modules. These changes need to be reverted afterwards.


site/docs/master/api.md
```diff
@@ -0,0 +1,4 @@
+# Contour API Reference
+
```
jonasrosland: Include something about how this is generated, which commands are used, etc.

jpeach: Do you mean, how does a developer re-generate this?

jpeach:
> Should these be part of the HTTPProxy docs/section? Right now they are disjointed from the others.

Maybe it makes sense under "Configuration"? Once we have API docs, I'd like to discuss tearing the large HTTPProxy document into smaller task-centric documents.

stevesloka:
I 100% agree we should break apart the HTTPProxy docs, I'm thinking these API docs are just a piece of that.

site/docs/master/api.md
jpeach: There's a documented makefile target that generates these docs. If we find later that we need more info, I'll be happy to add it in a separate PR.

PR #2073, opened 2019-12-19
Implement support for specifying a service's protocol in HTTPProxy.

mattmoor:
Fixes: #1962
Fixes: #2056

This is based on the design in: https://github.com/projectcontour/contour/pull/2072, which is also included as the base commit here.

### DIFF
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -196,6 +196,10 @@ type Service struct {
 	Name string `json:"name"`
 	// Port (defined as Integer) to proxy traffic to since a service can have multiple defined.
 	Port int `json:"port"`
+	// Protocol may be used to specify (or override) the protocol used to reach this Service.
+	// Values may be tls, h2, h2c.  It ommitted protocol-selection falls back on Service annotations.
+	// +optional
+	Protocol *string `json:"protocol,omitempty"`
 	// Weight defines percentage of traffic to balance traffic
 	// +optional
 	Weight uint32 `json:"weight,omitempty"`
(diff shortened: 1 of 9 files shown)


mattmoor:
Remove the WIP since the design was merged.  I've rebased this on master.

mattmoor:
I think that the hierarchy of semantics in place now is useful.  Folks can specify L7 protocol at the Service level in a context insensitive way.  They can override this at the HTTP Proxy level in a context sensitive way.  I suspect that the field would subsume the former, and as a standard mechanism probably (eventually) dwarf the latter as the preferred mechanism. 

Do you folks have a policy on how you adopt new minimum Kubernetes versions?  In Knative, we effectively support upstream the range of versions that are supported in K8s upstream at the time we cut our release.  We are allowed to pick up features available in all supported versions of Kubernetes.  By this policy, it'd probably be 18-24 months before we can adopt this feature in Knative, but I'm sure we will start to specifying this as soon as we practically can.

stevesloka:
I like the move from annotations so to me this is a step forward. Thanks for sharing the kep @youngnick, I think it's good to watch those, but it's still a ways off from us even using it given it needs to be approved, then coded+merged, then released. Even then it will be some time before we can provide full support to users. 

I think it's worth a short chat with the team but I'm good with moving forward on this and adding support for any changes from a kep later on. 

Given that we should discuss how we're going to support or deprecate the annotations style that currently exists.

mattmoor:
The other reality is that it will take time for consumption of the new Service model to percolate through other layers (like Knative), and that some systems simply may never gain the awareness to populate this at the K8s service level.

youngnick:
I'm okay with doing this for now, and dealing with the results of this KEP once they are available.

With respect to support, we don't currently specify a supported Kubernetes version, and have customers using extremely old versions, so it will take some effort to change. I agree that at a minimum we should match the upstream supported versions, but we have a way to go to get there.

With deprecation, I think that we can possibly call the contour-specific annotations deprecated - as long as we've updated all the docs to use the new style once it's available. I think we should also make sure that @michmike is aware, and he may have some thoughts on deprecation periods. My gut feeling is that we can leave this one around for a while.

mattmoor:
@youngnick I added it to alternatives considered in the design, and rebased the PR.

mattmoor:
@stevesloka done

PR #2075, opened 2019-12-27
internal/metrics: Add healthz metrics

### LINKED ISSUE #1560: /healthz endpoint should record statistics
our /healthz endpoint returns an error to the client if the k8s ping fails but does not record that fact on contour.

I don't want to log this at the contour level as it could cause log spam if the health check is flaky.

We should add metrics around this.

- [ ] /healthz hits
- [ ] successful responses
- [ ] unsuccessful responses

IngCr3at1on:
Fixes #1560

Signed-off-by: Nathan Bass <ingcr3at1on@gmail.com>

Tests cover total and successful requests but we need to be able to pass in
a _different_ fake kubernetes.Interface for the kubernetes.Clientset to trigger
the unsuccessful flow. I don't think anyone with projectcontour wants to maintain
a separate copy of the kubernetes.Interface so at the moment that is left out.

Changes the kubernetes.Clientset expectation to the kubernetes.Inteface in
registerHealthCheck because otherwise we can't pass in a fake.Clientset, this means
the call to ServerVersion has to be made on teh Discovery object instead of directly
on the Clientset as ServerVersion is not otherwise defined on the kubernetes.Interface.

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -79,6 +79,10 @@ const (
 	DAGRebuildGauge             = "contour_dagrebuild_timestamp"
 	cacheHandlerOnUpdateSummary = "contour_cachehandler_onupdate_duration_seconds"
 	resourceEventHandlerSummary = "contour_resourceeventhandler_duration_seconds"
+
+	healthzHitsGauge         = "contour_healthz_hits"
+	healthzSuccessfulGauge   = "contour_healthz_successful"
+	healthzUnsuccessfulGauge = "contour_healthz_unsuccessful"
 )
 
 // NewMetrics creates a new set of metrics and registers them with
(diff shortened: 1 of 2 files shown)


jpeach:
> I'd be inclined to explore a single metric with labels for the response code, which could probably be implemented as a generic http.Handler. Does something like this already exist?

https://godoc.org/github.com/prometheus/client_golang/prometheus/promhttp#InstrumentHandlerCounter

https://github.com/brancz/prometheus-example-app/blob/master/main.go#L45

IngCr3at1on:
> All these Gauges should be Counters, since their values can only increment

No problem, I'm not insanely familiar with prometheus so I probably missed that type.

> Take a look at hack/generate-metrics-doc.go where we generate metrics documentation and check the comment on metrics.NewMetrics. If we register new Prometheus metrics outside this path, they won't get documentation automatically generated.

Interesting... though I do have to wonder, do we want these documented or rather just existing in the background for those that know/care?

> IIUC, kubelet already exports these metrics on the /metrics/probe endpoint. Do we still want them in the contour namespace too?
> Since hits=success+failure, do we still need to publish that metric?

Hits was listed as a requirement on #1560 (along with the metrics themselves)

> I'd be inclined to explore a single metric with labels for the response code, which could probably be implemented as a generic http.Handler. Does something like this already exist?
>
> https://godoc.org/github.com/prometheus/client_golang/prometheus/promhttp#InstrumentHandlerCounter
>
> https://github.com/brancz/prometheus-example-app/blob/master/main.go#L45

Hmm... I guess most of the above doesn't matter if this is the preferred route :joy:
I can redo this I suppose just makes me question the point of discussing the changes on the issue first :stuck_out_tongue_winking_eye:

jpeach:
> do we want these documented or rather just existing in the background for those that know/care?

Yes, all metrics should be documented. I'd prefer that this is wired up to the generator, and I think that's quite achievable.

WRT the metric formats themselves, I think that there is a strong benefit in publishing metrics that are conventional within the prometheus ecosystem. It makes it much easier for operators to consume these metrics if they follow established patterns, and there's really no benefit to doing anything bespoke here AFAICT.

Here's an approach that can work and seems reasonable to me:

1. Split out a new "func (svc *Service) Register()" function that registers the metrics and mux handlers.
2. In this function, call "(*CounterVec).GetMetricWith" to initialize the counters so they get immediately emitted by the prometheus Registry.
3. Wrap both the metrics and health handlers in with "promhttp.InstrumentHandlerCounter"
4. In `hack/generate-metrics-doc.go`, register the metrics by creating a dummy "metrics.Service".

I think that the metric subsystem name should be probably be "http", which would end up generating metric names of the form "contour_http_$ENDPOINT_requests_total" (suggested by the [naming guide](https://prometheus.io/docs/practices/naming/#metric-names)).

Probably (1) is optional, and the same approach would work fine if you just dud everything under "(*Service).Serve()". You can verify the documentation is being generated correctly my running "make metrics-docs" and looking for the new files in the "site/_metrics" directory.

youngnick:
Hey, @IngCr3at1on, are you still able to do this work, or should someone else pick it up?

IngCr3at1on:
@youngnick sorry for the delay, my intention was to do it but I haven't had a lot of time for non-work related projects lately. If someone else wants to pick it up that may not be a bad idea otherwise I'm happy to do it just not quite sure when I'll have the time...

youngnick:
Currently, it's unprioritised, so we'll leave it for now. If you can get to it, great, if not, when one of us picks it up, we'll start from here.

PR #2076, opened 2020-01-01
Pre-size the array when dumping EndpointTranslator Contents()

mattmoor:
Just something I noticed poking around.


mattmoor:
It's unlikely it is measurable, but this avoids resizing the slice as you append things by pre-allocating enough elements.  Generally when creating a slice when you know the size, it is good practice to do this.  I have updated Query to be consistent.

PR #2078, opened 2020-01-02
Add API doc generator module dependencies.

jpeach:
We don't want building the API reference generator to update the
Contour Go modules, so set the GO11MODULE environment variable
to force that build to run in GOPATH mode.


hack/generate-refdocs.sh
```diff
@@ -1,5 +1,9 @@
 #! /usr/bin/env bash
 
+# Force GOPATH mode so that we don't add the generator as a Contour
+# module dependency.
+export GO111MODULE=off
```
mattmoor: This by itself doesn't work for me because it disables Go modules for the `go build` invocation in `gendoc::build`.

Prefixing the `go get` line with `GO111MODULE=off` seems to work, although the `go get` spits an error.

mattmoor: see: https://github.com/projectcontour/contour/pull/2006/files#diff-be4c0416b9c2099f206b82ed059fca7bR11

jpeach:
On Jan 3, 2020, at 3:40 AM, Matt Moore <notifications@github.com> wrote:

This by itself doesn't work for me because it disables Go modules for the go build invocation in gendoc::build.

It probably worked for me because I had a changed build :(

jpeach:
@mattmoor I couldn't find a reasonable way to get it to build and not update the dependencies, so I added it as a tools dependency like we do with other things.

jpeach:
@youngnick I found some other problems with my attempts to avoid using modules. Can you please take another look?

jpeach:
> On 7 Jan 2020, at 7:00 am, Steve Sloka <notifications@github.com> wrote:
> 
> 
> Running this I get:
> 
> ls: cannot access 'site/_data/template/*.tpl': No such file or directory

Ack, I didn't update the Makefile :-/

jpeach:
@stevesloka Updated the makefile, PTAL

PR #2086, opened 2020-01-08
site: stop linking absent document versions

### LINKED ISSUE #2085: Versioned API docs fail the site link checker
The website link check is failing with the following errors:

```
- /site/_site/docs/master/api/index.html
  *  internally linking to /docs/v1.0.0/api/, which does not exist (line 83)
     <a class="dropdown-item" href="/docs/v1.0.0/api/">v1.0.0</a>
  *  internally linking to /docs/v1.0.1/api/, which does not exist (line 86)
     <a class="dropdown-item" href="/docs/v1.0.1/api/">v1.0.1</a>
  *  internally linking to /docs/v1.0.1/api/, which does not exist (line 284)
     <a href="/docs/v1.0.1/api/">latest version</a>
- /site/_site/docs/master/tagging/index.html
  *  External link https://projectcontour.io/docs/master/tagging/ failed: 404 No error
- /site/_site/docs/v1.0.0/tagging/index.html
  *  External link https://projectcontour.io/docs/v1.0.0/tagging/ failed: 404 No error
- /site/_site/docs/v1.0.1/tagging/index.html
  *  External link https://projectcontour.io/docs/v1.0.1/tagging/ failed: 404 No error
htmlproofer 3.13.0 | Error:  HTML-Proofer found 6 failures!
```

This happens because the version toggle on each versioned documentation page is generated to like directly to the same page in each versioned documentation directory. So if you navigate to `/docs/master/api`, the toggle will link to `/docs/v1.0.0/api` and `/docs/v1.0.1/api`. These pages don't exist (API docs are only being generated on master).

So the corollary of this is that every page in the versioned documentation must exist in every version, which seems awkward and hard to maintain.

jpeach:
If a new versioned document is added, it may not be present in all
versions of the documentation. In this case, the version selection
button should not generate a link to an absent version of the
document.

The fix is to scan the site pages before generating the link to
ensure that the target version of the document is actually present.

This fixes #2085.


jpeach:
Alternatively, we could keep the versions in the drop list, but link to '/' for the version (rather than simply omitting the droplets entry like we do here).

jpeach:
Rather than omitting the link, I forced the link to the root of the target version. This fixes the version warning message box too.

PR #2099, opened 2020-01-10
docs: suggested to delete CRDs as well when upgrading with the easy way

idealhack:
otherwise, users might see errors because of the old CRDs.

### DIFF
--- a/site/_resources/upgrading.md
+++ b/site/_resources/upgrading.md
@@ -36,7 +36,7 @@ Please see the [Envoy Release Notes][15] for information about issues fixed in E
 If the following are true for you:
 
  * Your installation is in the `projectcontour` namespace.
- * You are using one of the [example][11] deployments.
+ * You are using one of the [example][1] deployments.
  * Your cluster can take few minutes of downtime.
 
 Then the simplest way to upgrade to 1.1.0 is to delete the `projectcontour` namespace and reapply one of the example configurations.
@@ -112,7 +112,7 @@ Ensure the Envoy image version is `docker.io/envoyproxy/envoy:v1.11.2`.
 If the following are true for you:
 
  * Your previous installation is in the `projectcontour` namespace.
- * You are using one of the [example][1] deployments.
+ * You are using one of the [example][2] deployments.
  * Your cluster can take few minutes of downtime.
 
 Then the simplest way to upgrade is to delete the `projectcontour` namespace and reapply the `examples/contour` sample manifest.
@@ -416,6 +416,7 @@ $ kubectl get configmap -n heptio-contour -o yaml contour
 ```
 
 [1]: {{site.github.repository_url}}/blob/{{site.latest}}/examples
+[2]: {{site.github.repository_url}}/blob/v1.0.0/examples
 [3]: /docs/{{site.latest}}/httpproxy
 [4]: /docs/{{site.latest}}/annotations
 [5]: {{site.github.repository_url}}/issues/899
--- a/site/_resources/upgrading.md
+++ b/site/_resources/upgrading.md
@@ -54,6 +54,14 @@ You'll need to re-check where your DNS names are pointing as well, using [Get yo
 **Note:** If you deployed Contour into a different namespace than `projectcontour` with a standard example, please delete that namespace.
 Then in your editor of choice do a search and replace for `projectcontour` and replace it with your preferred name space and apply the updated manifest.
 
+**Note:** If you are deploying to a cluster where you have previously installed alpha versions of the Contour API, applying the Contour CRDs in `examples/contour` may fail with a message similar to `Invalid value: "v1alpha1": must appear in spec.versions`. In this case, you need to delete the old CRDs and apply the new ones.
+
+```
+$ kubectl delete namespace projectcontour
+$ kubectl get crd  | awk '/projectcontour.io|contour.heptio.com/{print $1}' | xargs kubectl delete crd
+$ kubectl apply -f examples/<your-desired-deployment>
+```
+
 ## The less easy way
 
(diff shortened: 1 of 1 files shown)


idealhack:
I do notice that `CustomResourceDefinition`s only exist in `examples/contour` and `examples/render`, let me know if you think there're better ways to handle this :)

jpeach:
@idealhack Can you give me more detail about why you would need to delete the CRDs? I'd expect that deploying a new Contour release should simply update the CRDs in place. Did you experience any errors with this approach?

idealhack:
For one of the older deployments, I saw errors when I apply the manifests, but after deleting the CRD, I don't see them.

I didn't keep the logs, but it seems like related to some old API versions, as I can see I searched `Invalid value: "v1alpha1": must appear in spec.versions` in my browser history.

jpeach:
I tested this and the problem seems to be upgrading across the alpha versions of the HTTPProxy API. If you apply the Contour YAML from v1.0.0-beta.1 and then the YAML from v1.0.0.0rc1, it will fail with an error similar to this:

```
...
Error from server (Invalid): error when applying patch:
...
Resource: "apiextensions.k8s.io/v1beta1, Resource=customresourcedefinitions", GroupVersionKind: "apiextensions.k8s.io/v1beta1, Kind=CustomResourceDefinition"
Name: "httpproxies.projectcontour.io", Namespace: ""
...
for: "examples/render/contour.yaml": CustomResourceDefinition.apiextensions.k8s.io "httpproxies.projectcontour.io" is invalid: [spec.version: Invalid value: "v1alpha1": must match the first version in spec.versions, status.storedVersions[0]: Invalid value: "v1alpha1": must appear in spec.versions]
``` 

The problem here is that the version of the CRD changes, but the API Server invariants are not preserved.

I was able to get the new API to apply by making the following changes to the YAML:
```
diff --git examples/render/contour.yaml examples/render/contour.yaml
index fbee85c..793988b 100644
--- examples/render/contour.yaml
+++ examples/render/contour.yaml
@@ -896,12 +896,16 @@ spec:
   - name: v1
     served: true
     storage: true
+  - name: v1alpha1
+    served: false
+    storage: false
 status:
   acceptedNames:
     kind: ""
     plural: ""
   conditions: []
-  storedVersions: []
+  storedVersions:
+  - v1alpha1
 ---
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
```

I'm not really sure what the right fix here is yet. Maybe there's a way to generate the CRD YAML so that it correctly handles upgrading versions. If not, then mentioning the problem in the documentation is fine, but we should make it clear that it is only needed when the CRD versions change.

jpeach:
@idealhack I think that you are right, and we can document this problem. I'd suggest a new paragraph describing the context:

> If your are deploying to a cluster where you have previously installed alpha versions of the
> Contour API, applying the Contour CRDs may fail with a message similar to `Invalid value:
> "v1alpha1": must appear in spec.versions`. In this case ...

This is a little picky, but in your cleanup instructions, I think that it is more robust to do: `awk '/projectcontour.io|contour.heptio.com/{print $1}'`

Thanks!

idealhack:
@jpeach thanks for your investigation and suggestions, I’ve updated the change, please take another look.

site/_resources/upgrading.md
```diff
@@ -54,6 +54,14 @@ You'll need to re-check where your DNS names are pointing as well, using [Get yo
 **Note:** If you deployed Contour into a different namespace than `projectcontour` with a standard example, please delete that namespace.
 Then in your editor of choice do a search and replace for `projectcontour` and replace it with your preferred name space and apply the updated manifest.
 
+**Note:** If you are deploying to a cluster where you have previously installed alpha versions of the Contour API, applying the Contour CRDs may fail with a message similar to `Invalid value: "v1alpha1": must appear in spec.versions`. In this case, you need to delete the old CRDs before applying the new ones.
+
+```
+$ kubectl delete namespace projectcontour
+$ kubectl get crd  | awk '/projectcontour.io|contour.heptio.com/{print $1}' | cut -d ' ' -f 1 | xargs kubectl delete crd
```
jpeach: You don't need both the awk and the cut, so you can remove the cut. FWIW the reason I suggested the awk is that it's more verbose and better documents to the reader what this pipeline is doing.

idealhack: oh, of course, thanks for catching that!

site/_resources/upgrading.md
```diff
@@ -54,6 +54,14 @@ You'll need to re-check where your DNS names are pointing as well, using [Get yo
 **Note:** If you deployed Contour into a different namespace than `projectcontour` with a standard example, please delete that namespace.
 Then in your editor of choice do a search and replace for `projectcontour` and replace it with your preferred name space and apply the updated manifest.
 
+**Note:** If you are deploying to a cluster where you have previously installed alpha versions of the Contour API, applying the Contour CRDs may fail with a message similar to `Invalid value: "v1alpha1": must appear in spec.versions`. In this case, you need to delete the old CRDs before applying the new ones.
+
+```
+$ kubectl delete namespace projectcontour
+$ kubectl get crd  | awk '/projectcontour.io|contour.heptio.com/{print $1}' | xargs kubectl delete crd
```
stevesloka: Not a big deal but I don't think this problem affects the `contour.heptio.com` crds?

It might be good to add a note that doing this will delete any CRDs existing in the cluster, so essentially, you'll have to reapply all your configuration again.

idealhack: I don’t really remember which of the four CRDs are affected, only that both `examples/contour` and `examples/render` have them (as said above), I’ll update this to be more explicit about it.

PR #2106, opened 2020-01-13
internal/featuretests: convert e2e root namespace tests to featuretests

davecheney:
Converted e2e --root-namespace tests for LDS and RDS to feature tests. In the process I discovered that the old tests asserted the negative case -- something not in the root namespace set would be excluded from RDS and LDS, but the test for the positive case missed some cases so those have been fixed as well as extending coverage across http proxy. 


internal/featuretests/rootnamespaces_test.go
```diff
@@ -0,0 +1,258 @@
...
+	c.Request(routeType).Equals(&v2.DiscoveryResponse{
+		Resources: nil,
+		TypeUrl:   routeType,
+	})
+
+	// ir1 is an ingressroute that is not in the root namespaces
+	ir1 := &ingressroutev1.IngressRoute{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "simple",
+			Namespace: svc1.Namespace,
```
jpeach: Nit: When reading tests, I find this sort of reference a but hard to follow. I always end up scrolling back up to see what it means and lose context. Maybe using local constants would achieve a similar result but improve readability? To my mind, another benefit of this approach is that it doesn't need so many temporary objects like "svc1".

PR #2112, opened 2020-01-14
internal/contour: refactor metrics test for accurate line numbers

jpeach:
When a test fails, the reported line number is in the subtest, and
you have to search for the tesst case name to find the test case
that had a problem. Refactor the test code into a local function
so we can preserve the attributed of table-driven tests but aldo
get accurate reporting of line numbers for failures.

### DIFF
--- a/internal/contour/metrics_test.go
+++ b/internal/contour/metrics_test.go
@@ -26,6 +26,43 @@ import (
 )
 
 func TestIngressRouteMetrics(t *testing.T) {
+	type testcase struct {
+		objs           []interface{}
+		wantIR         *metrics.RouteMetric
+		wantProxy      *metrics.RouteMetric
+		rootNamespaces []string
+	}
+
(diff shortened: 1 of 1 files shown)


internal/contour/metrics_test.go
```diff
@@ -26,6 +26,43 @@ import (
...
+		wantIR         *metrics.RouteMetric
+		wantProxy      *metrics.RouteMetric
+		rootNamespaces []string
+	}
+
+	run := func(t *testing.T, name string, tc testcase) {
+		t.Helper()
+
+		t.Run(name, func(t *testing.T) {
+			t.Helper()
```
stevesloka: Does this need called again after being called on line 37?

jpeach: Yep, we need to mark all the intermediate functions so that the reported line number is the place where we define and call the test case.

PR #2116, opened 2020-01-15
Update kind cluster example to expose port 80 and 443 instead of 8080

shyaamsn:
Currently examples/kind/kind-expose-port.yaml is exposing port 8080 to the host. However, by default envoy is listening on hostPort 80 and 443. So, the kind-expose-port.yaml has to be changed to expose port 80 and 443 instead of 8080

### DIFF
--- a/examples/kind/kind-expose-port.yaml
+++ b/examples/kind/kind-expose-port.yaml
@@ -4,6 +4,9 @@ nodes:
 - role: control-plane
 - role: worker
   extraPortMappings:
-  - containerPort: 8080
-    hostPort: 8080
+  - containerPort: 80
+    hostPort: 80
+    listenAddress: "0.0.0.0"
+  - containerPort: 443
+    hostPort: 443
     listenAddress: "0.0.0.0"


davecheney:
> However, by default envoy is listening on hostPort 80 and 443.

Is this true, I thought that we deployed envoy on 8080 and 8443 by default.

jpeach:
> Is this true, I thought that we deployed envoy on 8080 and 8443 by default.

https://github.com/projectcontour/contour/blob/master/examples/render/contour.yaml#L1485

jpeach:
> envoy does listen on 8080 and 8443 but we program it to expect traffic forwarded to it from whatever nat gateway or tcp proxy to be coming from 80 and 443.

IIUC the `--envoy-service-*-port` flag sets the port that the envoy listener will use when Contour brings it up. Is there more going on?

davecheney:
Its more subtle than that. Those flags have to match the configuration in the k8s service document -- they represent how remote traffic expects to land on envoy, and vice versa how envoy expects traffic to appear to it. It's related to the bs that envoy doesn't strip the port from `Host:` headers as well as knowing which port to use when redirecting, etc.

jpeach:
> Those flags have to match the configuration in the k8s service document -- they represent how remote traffic expects to land on envoy, and vice versa how envoy expects traffic to appear to it.

Right. I'm confused as to where the 8080 and 8443 ports you mentioned come into it.

At any rate, if you are OK with this change, I'll merge.

davecheney:
> Right. I'm confused as to where the 8080 and 8443 ports you mentioned come into it.

https://github.com/projectcontour/contour/blob/master/internal/contour/listener.go#L37

Issue #2118, opened 2020-01-16
Move Contour from Informers to Kubebuilder controller-runtime


youngnick:
As part of reviewing how to integrate the [service-api](https://github.com/kubernetes-sigs/service-apis) CRDs into Contour, I noticed that our code for interacting with Kubernetes is very custom, and there are now better options available.

This issue is to cover moving Contour from Informers to using the Kubebuilder [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime) library.

First up, a design proposal for us to discuss whether this is even a good idea and how it might be done.

youngnick:
I'm going to close this out, as @stevesloka has moved us to Dynamic client instead as part of #2191. We may revisit this at a later date.

PR #2120, opened 2020-01-16
design: Add design for migrating from Informers to controller-utils

youngnick:
Updates #2118

### DIFF
--- a/design/kubebuilder-migration.md
+++ b/design/kubebuilder-migration.md
@@ -1,0 +1,62 @@
+# Move Contour to use kubebuilder instead of informers
+
+Status: Draft
+
+Contour's method of interaction with Kubernetes is two years old, and the community has moved on from Informers.
+This proposal suggests that we should move to using Kubebuilder [controller-runtime][1] for interacting with Kubernetes.
+This will make extending Contour with additional CRDs as we need to more straightforward.
+
+## Goals
+
+- Migrate Contour to use Kubebuilder's [controller-runtime][1] library instead of informers.
+
+## Non Goals
+
+- Implement the new service-api types as part of this proposal
+- Use Kubebuilder for code autogeneration. We've already done equivalent work.
+
+## Background
+
+Two years ago, @davecheney built Contour to use client-go Informer code to talk to Kubernetes.
+At the time, this was a sensible decision, as there were no well-developed frameworks or common methods for interacting with Kubernetes.
+Since then, the community has done a large amount of work to simplify the construction of controllers, currently embodied in the kubebuilder [controller-runtime][1] library.
+
+This has become an issue now because we are looking at adding support for further CRDs, (the SIG-Network service-apis) and they are not shipping with Informers generated.
+Moving to the same framework as the new service-api types will allow us to easily consume them, as well as add any other new types without requiring generated Informers.
+
+In addition, moving to using controller-runtime means that we have access to frameworks for helping with:
+
+- leader election
+- defaulting values for CRDs
+- webhooks to translate between CRD versions (related to #2068)
+
+## High-Level Design
+
+In order to avoid a long-running feature branch, I propose a multi-phase approach that can be broadly summarized as:
+
+- Move things that interact with the Kubernetes apiserver to use more standard implementations (particularly Caches, but there may be others). This will involve a few changes at least.
+- Move `contour serve` to using the controller-runtime framework for its serving.
+
+In particular, the controller-runtime framework requires Controllers to have a Reconcile function that is supplied the details of an object that has changed, and the controller is expected to perform whatever actions it needs to to ensure that the new object correctly reflects the state of the world.
+It's a reconciliation loop in a function, in other words.
+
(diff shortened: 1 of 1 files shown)


design/kubebuilder-migration.md
```diff
@@ -0,0 +1,64 @@
...
+
+Status: Draft
+
+Contour's method of interaction with Kubernetes is two years old, and the community has moved on from Informers.
+This proposal suggests that we should move to using Kubebuilder [controller-runtime][1] for interacting with Kubernetes.
+This will make extending Contour with additional CRDs as we need to more straightforward.
+
+## Goals
+
+- Migrate Contour to use Kubebuilder [controller-runtime][1] instead of informers.
```
stevesloka: Are you proposing to use kubebuilder or just the controller-runtime project?

design/kubebuilder-migration.md
```diff
@@ -0,0 +1,64 @@
...
+
+### Moving towards more standard Kubernetes interactions
+
+In order to do this, we will need to move all of the functions that require access to cached Kubernetes objects to use a more standard Kubernetes cache (instead of the custom Contour one.)
+
+Once that is done, we'll need to move the code that performs changes from Add, Update and Delete, into a singular Reconcile function that figures out what has changed and does that.
+
+The detailed design section is omitted because this work will of necessity need to change as it goes.
+
+## Alternatives Considered
```
stevesloka: Another alternative is that we could use client-go dynamic client which doesn't require the generated client like we do today and I suspect it's what the controller-runtime is using.

design/kubebuilder-migration.md
youngnick: Just the controller-runtime parts, since we don't need the autogeneration scaffolding.

davecheney: It might be good to clarify that, perhaps in the non goals section.

design/kubebuilder-migration.md
```diff
@@ -0,0 +1,64 @@
...
+This proposal suggests that we should move to using Kubebuilder [controller-runtime][1] for interacting with Kubernetes.
+This will make extending Contour with additional CRDs as we need to more straightforward.
+
+## Goals
+
+- Migrate Contour to use Kubebuilder [controller-runtime][1] instead of informers.
+
+## Non Goals
+
+- Change anything after the DAG builder
```
davecheney: probably redundant, this could be moved to the high level design section with a note that we're just considering how we get notifications from the api server. period.

design/kubebuilder-migration.md
```diff
@@ -0,0 +1,64 @@
...
+- Implement the new service-api types as part of this proposal
+
+## Background
+
+Two years ago, @davecheney built Contour to use custom-wired code to talk to Kubernetes.
```
davecheney: scustom-wired/the informer framework

design/kubebuilder-migration.md
```diff
@@ -0,0 +1,64 @@
...
+At the time, this was a sensible decision, as there were no well-developed frameworks or common methods for interacting with Kubernetes.
+Since then, the community has done a large amount of work to simplify the construction of controllers, currently embodied in the kubebuilder [controller-runtime][1] library.
+
+This has become an issue now because we are looking at adding support for further CRDs, (the SIG-Network service-apis) and want to leave the door open for more if we need to.
```
davecheney: would it be better to highlight that the problem is service-api's aren't going to ship with informers?

design/kubebuilder-migration.md
```diff
@@ -0,0 +1,64 @@
...
+
+In addition, moving to using controller-runtime means that we have access to frameworks for helping with:
+
+- leader election
+- defaulting values for CRDs
+- webhooks to translate between CRD versions (related to #2068)
+
+without having to custom-build that code.
```
davecheney: delete line

design/kubebuilder-migration.md
```diff
@@ -0,0 +1,64 @@
...
+
+Lastly, this will allow us to consume the new service-api types and any other CRDs (including our own), without generating the client-go code to make Informers.
```
davecheney: I think this should be the primary point, not an ancillary benefit.

design/kubebuilder-migration.md
youngnick: That's a good point, I will check that out.

design/kubebuilder-migration.md
```diff
@@ -0,0 +1,64 @@
...
+The service-api project currently does not generate these.
+
+This is a pretty big refactor though, so we should agree it's a good idea before starting.
```
davecheney: probably not appropriate commentary.

It is probably worth pointing out in the high level design that because of our desire to avoid a long running feature branch and a with make sure the product is always shippable, the incremental approach is the way to go.

design/kubebuilder-migration.md
youngnick: Agreed, thanks.

youngnick:
I'm closing this pull request, as we've moved to the dynamic client instead as per #2191

PR #2121, opened 2020-01-16
design: add draft for ingress status loadbalancer support

davecheney:
Updates #403

Strawman design for supporting the ingress.status.loadbalancer field.

### DIFF
--- a/design/ingress-status-loadbalancer.md
+++ b/design/ingress-status-loadbalancer.md
@@ -1,0 +1,101 @@
+# Ingress Status Loadbalancer Support
+
+Status: _Draft_
+
+This proposal describes how to add support for the `ingress.status.loadBalancer` field to Contour.
+
+## Goals
+
+- Populate the status.loadbalancer field in all `networking.k8s.io/v1beta.Ingress` objects managed by Contour.
+- Contour pass the Ingress compliance tests for `networking.k8s.io/v1beta.Ingress` objects.
+
+## Non Goals
+
+- Similar functionality for IngressRoute (deprecated) and HTTPProxy are out of scope.
+
+## Background
+
+The Kubernetes Ingress object provides a field on the object's status to record address details for how access the ingresses virtualhosts from outside the cluster.
+This is known as the status loadbalancer information after the name of the fields on the Ingress Object.
+
+Details from the Kubernetes spec [are here](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#loadbalancerstatus-v1-core).
+
+## High-Level Design
+
+There are two components to this solution
+
+1. Discovering the loadbalancer information
+2. Writing the loadbalancer information to the ingress objects.
+
+The former, the discovery will need to support 
+
+1. Hard coded IP/address details (actually a list, and it can be an address or an IP)
+2. Discovering information from the service document.
+
+### Hard coded load balancer addresses
+
+While the hard coded name approach sounds simpler, and is, it will likely not be the common use case, nor will it be the default.
+This is for several reasons.
+
+- The hard coded address model implies either the cluster is running on prem, or using host networking.
+- The hard coded IP model requires the administator to know the IP addresses of the _envoy_ service (not the contour service) which may change over time. 
+
+The configuration will likely go into Contour's configuration file.
+
+These restrictions (which are administrative, not ones we've placed on the solution) make this bookkeeping expensive, prohibitively in the case of a highly dynamic cluster.
+
+### Dynamic load balancer address discovery
+
+The far more common case will be dynamic load balancer discovery.
+In this model Contour will track a service document (default to projectcontour/envoy, but configurable) and use those details as the list of IP addresses to write back to the ingress' status.
+
+The nginx controller goes to some lengths to try to dynamically figure out what it's status document as well as other side channels to discover the address if that fails.
+For this first implementation I propose we stick to the discovery methods outlined above.
+
+### Status updater
+
+The latter, writing the status to valid ingress objects, waits to be elected the leader then updates the status document on all of the ingress objects within Contour's class.
+
+As new ingress objects arrive they will need to be updated.
+
+Note: there appears to be no requirement that if Contour is shutting down we delete the status field on the ingress on the way out.
+Initially this feels wrong but consider the scenarios:
+
+- Leadership has deposed the current Contour causing it to shut down; the new leader will be responsible for updating status. Best that during this transition observers of the Ingress document do not see status disappear and reappear quickly.
+- Contour is shut down for a long period of time. Yes, the status information will be incorrect, but there is no correct value, no ingress controller is serving that vhost.
+- Contour is being replaced with something else, then the responsibility to _overwrite_ the status information falls to Contour's replacement.
+
+## Detailed Design
+
+I think the design should involve two goroutines added to cmd/contour's workgroup.
+The first will be responsible for load balancer discovery, the second will be responsible for updating status documents.
+
+### Discovery worker
+
+The discovery worker should look something like this pseudocode:
+
+1. worker starts up, blocks on leadership election.
+2. when elected leader it checks the mode.
+    * if discovery is hard coded in the config file it transmits these addresses over a channel to the other worker, then goes to sleep until deposed and/or shutdown.
+    * if discovery is dynamic it opens a watcher on the specified (defaults, config file, cli flags, in that order) service document and each time it changes sends the address details to the other worker until deposed or shutdown. 
+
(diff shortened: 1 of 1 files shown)


design/ingress-status-loadbalancer.md
```diff
@@ -0,0 +1,101 @@
...
+## High-Level Design
+
+There are two components to this solution
+
+1. Discoverying the loadbalancer information
+2. Writing the loadbalacner information to the ingress objects.
+
+The former, the discovery will need to support 
+
+1. Hard coded IP/address details (actually a list, and it can be an address or an ip)
```
youngnick: Nit: 'it can be a domain name or an ip address' is probably slightly more accurate.

design/ingress-status-loadbalancer.md
```diff
@@ -0,0 +1,101 @@
...
+2. Discoverying information from the service document.
+
+### Hard coded load balancer addresses
+
+While the hard coded approach sounds simpler, and is, it will likely not be the common use case, nor will it be the default.
+This is for several reasons.
+
+- The hard coded address model implies either the cluster is running on prem, or using host networking.
```
youngnick: This should probably be 'hard coded *name*'.

Additionally, it doesn't have to be host networking, it can be done with a HostPort and an external load balancer. The important part here is that name-based hard coding implies the use of an external load-balancer that can front the Envoy service. It might be on-prem, host networking, in AWS, or whatever. But the most common way to get a stable name is to use an external load balancer.

You could also front a pool of machines with a straight domain name, I suppose, but this would involve having Envoy listen on 80 and 443 and for those machines to have internet-reachable IPs.

design/ingress-status-loadbalancer.md
```diff
@@ -0,0 +1,101 @@
...
+These restrictions (which are administrative, not ones we've placed on the solution) make this bookkeeping expensive, prohibitivly in the case of a highly dynamic cluster.
+
+### Dynamic load balancer address discovery
+
+The far more common case will be dynamic load balancer discovery.
+In this model Contour will track a service document (default to projectcontour/envoy, but configurable) and use those details as the list of ip addresses to write back to the ingress' status.
+
+The nginx controller goes to some lengths to try to dynamically figure out what it's status document as well as other side channels to discover the address if that fails.
+For this first implementation I propose we stick to the discovery methods outlined above.
+
```
youngnick: Is the discoverer not being able to discover some option a fatal error?

My feeling is no, I think, since that's the current situation anyway.

davecheney: > Is the discoverer not being able to discover some option a fatal error?

yeah, I'm thinking no as well. maybe we log something (as long as its not too incessant) but I can easily see cases where contour's pod races with the creation of envoy's service object and that'll just generate confusion and bug reports.

design/ingress-status-loadbalancer.md
```diff
@@ -0,0 +1,101 @@
...
+While the hard coded approach sounds simpler, and is, it will likely not be the common use case, nor will it be the default.
+This is for several reasons.
+
+- The hard coded address model implies either the cluster is running on prem, or using host networking.
+- The hard coded IP model requires the administator to know the IP addresses of the _envoy_ service (not the contour service) which may change over time. 
+
+The configuration will likely go into Contour's configuration file.
+
+These restrictions (which are administrative, not ones we've placed on the solution) make this bookkeeping expensive, prohibitivly in the case of a highly dynamic cluster.
```
stevesloka: nit: prohibitively

design/ingress-status-loadbalancer.md
stevesloka: There's also a bit of delay to determine the ip/name while the lb is spun up, so we need to make sure that delay is not a failure.

design/ingress-status-loadbalancer.md
```diff
@@ -0,0 +1,101 @@
...
+### Discovery worker
+
+The discovery worker should look something like this pseudocode:
+
+1. worker starts up, blocks on leadership election
+2. when elected leader it checks the mode
+2a. if discovery is hard coded in the config file it transmits these addresses over a channel to the other worker, then goes to sleep until deposed and/or shutdown.
+2b. if discovery is dynamic it opens a watcher on the specified (defaults, config file, cli flags, in that order) service document and each time it changes sends the address details to the other worker until deposed or shutdown. 
+
+Open question: if the service document goes away, should we transmit an empty set of addresses to the status worker to indicate it should remove the status.loadbalancer details?
```
stevesloka: I think this is a reasonable way to handle this scenario. If this would happen users may not know that the service is gone, but they would when the status field is updated to empty.

design/ingress-status-loadbalancer.md
```diff
@@ -0,0 +1,101 @@
...
+
+The status worker should look something like this pseudocode:
+
+1. the worker starts up, blocks on leadership election
+2. once elected the worker blocks until it has received a set of addresses
+3. the worker starts a watch on all ingress objects.
+4. for each ingress observed, if the even is an add or update, we patch it with the latest set of addresses. For deletes, we do nothing.
+
+The open question is how to deal with the discovery worker sending a new set of addresses after the initial pass.
+It feels like we need to keep a side cache of ingress objects we have previously seen to go back and bulk update all the ingresses' seen so far.
```
stevesloka: When I did the first PR I had this logic in the builder, so we had access to the ingress caches and was helpful to have. On my PR rewrite, I was thinking of a way of making those caches generic to allow them to be consumed by more than just the builder.

design/ingress-status-loadbalancer.md
```diff
@@ -0,0 +1,101 @@
...
+Care should be taken to ensure the discovery worker can respond to the group's shutdown signal.
+
+### Status worker
+
+The status worker should look something like this pseudocode:
+
+1. the worker starts up, blocks on leadership election
+2. once elected the worker blocks until it has received a set of addresses
+3. the worker starts a watch on all ingress objects.
+4. for each ingress observed, if the even is an add or update, we patch it with the latest set of addresses. For deletes, we do nothing.
```
jpeach: s/even/event/

design/ingress-status-loadbalancer.md
jpeach: Why do you need a side-cache here? Just list the Ingresses and update as appropriate.

design/ingress-status-loadbalancer.md
```diff
@@ -0,0 +1,101 @@
...
+2b. if discovery is dynamic it opens a watcher on the specified (defaults, config file, cli flags, in that order) service document and each time it changes sends the address details to the other worker until deposed or shutdown. 
+
+Open question: if the service document goes away, should we transmit an empty set of addresses to the status worker to indicate it should remove the status.loadbalancer details?
+
+The send from the discovery worker to the status worker will be over an unbuffered channel so the discovery worker cannot send until the status worker has been elected leader and is ready to receive ip details.
+Care should be taken to ensure the discovery worker can respond to the group's shutdown signal.
+
+### Status worker
+
+The status worker should look something like this pseudocode:
```
jpeach: Is there any relationship between the load balancer status and the Ingress validity? We should call that out here.

cten:
Really excited about this, any idea when work will start? I have been putting off migrating to Nginx Ingress hoping this gets implemented. Happy to help test when work starts.

design/ingress-status-loadbalancer.md
```diff
@@ -0,0 +1,101 @@
...
+The configuration will likely go into Contour's configuration file.
+
+These restrictions (which are administrative, not ones we've placed on the solution) make this bookkeeping expensive, prohibitivly in the case of a highly dynamic cluster.
+
+### Dynamic load balancer address discovery
+
+The far more common case will be dynamic load balancer discovery.
+In this model Contour will track a service document (default to projectcontour/envoy, but configurable) and use those details as the list of ip addresses to write back to the ingress' status.
+
+The nginx controller goes to some lengths to try to dynamically figure out what it's status document as well as other side channels to discover the address if that fails.
```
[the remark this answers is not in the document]
davecheney: the nginx controller tries several ways to figure out its own IP address, https://github.com/kubernetes/ingress-nginx/blob/master/internal/ingress/status/status.go#L328

design/ingress-status-loadbalancer.md
davecheney: As a compromise, how about we log when new address details become available, ie
```
ingressStatusWriter: loadbalancer address details changed to %v, updating Ingress objects
```

youngnick: Maybe a metric as well?

PR #2128, opened 2020-01-16
internal/dag: mark ingress.class annotation valid for HTTPProxies

### LINKED ISSUE #2119: Mark "kubernetes.io/ingress.class" supported on HTTPProxy
If you set the "kubernetes.io/ingress.class" annotation on a HTTPProxy document, Contour will log a warning that it is unsupported. In fact, it is supported and works correctly. We should update the warning.

jpeach:
The "kubernetes.io/ingress.class" annotation is valid for all
ingress-like document kinds.

This fixes #2119.

### DIFF
--- a/internal/dag/annotations.go
+++ b/internal/dag/annotations.go
@@ -68,9 +68,11 @@ var annotationsByKind = map[string]map[string]struct{}{
 		"projectcontour.io/upstream-protocol.tls": {},
 	},
 	"HTTPProxy": {
+		"kubernetes.io/ingress.class":     {},
 		"projectcontour.io/ingress.class": {},
 	},
 	"IngressRoute": {
+		"kubernetes.io/ingress.class":     {},
 		"projectcontour.io/ingress.class": {},
 	},
 }


stevesloka:
Should the ingress class go in the HTTPProxy document and not an annotation?

jpeach:
> On 17 Jan 2020, at 8:54 am, Steve Sloka <notifications@github.com> wrote:
> 
> Should the ingress class go in the HTTPProxy document and not an annotation?

Possibly, as a separate change, though we can't remove the annotation support now without breaking compatibility.

davecheney:
> Should the ingress class go in the HTTPProxy document and not an annotation?

I don't have a strong opinion about this but I wanted to give some historical detail.

When we were adding ingressroute we asked the same question and at the time decided that ingress class was probably an appropriate use of an annotation, it is more or less a special kind of label. We might have also talked about adding the equivalent of the annotation as a field, but at the time that didn't go anywhere.

stevesloka:
// https://github.com/projectcontour/contour/issues/720

PR #2133, opened 2020-01-21
apis: replace unsigned integers in CRD fields with signed integers, add validation

### LINKED ISSUE #2123: DeepCopy unstructured HTTPProxy causes panic
When converting an HTTPProxy object to an unstructured object and then performing a DeepCopy on that unstructured object, a panic is experienced.

The following code can be used to replicate:

```go
package main

import (
	v1 "github.com/projectcontour/contour/apis/projectcontour/v1"
	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
	"k8s.io/apimachinery/pkg/runtime"
)

func main() {
	proxy := &v1.HTTPProxy{
		TypeMeta: metav1.TypeMeta{
			Kind:       "HTTPProxy",
			APIVersion: v1.SchemeGroupVersion.String(),
		},
		Spec: v1.HTTPProxySpec{
			Routes: []v1.Route{
				{
					Services: []v1.Service{
						{
							Weight: uint32(100),
						},
					},
				},
			},
		},
	}

	u, err := runtime.DefaultUnstructuredConverter.ToUnstructured(proxy.DeepCopyObject())
	if err != nil {
		panic(err)
	}

	unstr := &unstructured.Unstructured{Object:u}
	unstr.DeepCopy()
}
```

When run, results in:
```
panic: cannot deep copy uint64
```

For the above code to run without issue.

I believe this is occurring due to the use of `uint32` for the `weight` (and other fields) in the HTTPProxy struct. I've manually changed the `weight` field in the `Service` struct from `uint32` to a signed integer and the code above runs fine.

Works:
```go
	// Weight defines percentage of traffic to balance traffic
	// +optional
	Weight int32 `json:"weight,omitempty"`
```

Looking at the Kubernetes external API codebase, no unsigned integers are used in the public structs - https://github.com/kubernetes/api, most likely for this reason.

**Environment:**

- Contour version: v1.1.0
- Kubernetes version: (use `kubectl version`): v1.15.7

awprice:
This commit replaces the unsigned integers in the HTTPProxy and IngressRoute CRDs
with signed integers. Unsigned integers are discouraged due to API compatibility
issues. Validation rules have been added to the fields to ensure that the value
provided is always positive.

Fixes #2123

Signed-off-by: Alex Price <aprice@atlassian.com>

### DIFF
--- a/apis/contour/v1beta1/ingressroute.go
+++ b/apis/contour/v1beta1/ingressroute.go
@@ -86,7 +86,8 @@ type Service struct {
 	Port int `json:"port"`
 	// Weight defines percentage of traffic to balance traffic
 	// +optional
-	Weight uint32 `json:"weight,omitempty"`
+	// +kubebuilder:validation:Minimum=0
+	Weight int64 `json:"weight,omitempty"`
 	// HealthCheck defines optional healthchecks on the upstream service
 	// +optional
 	HealthCheck *HealthCheck `json:"healthCheck,omitempty"`
@@ -115,10 +116,12 @@ type HealthCheck struct {
 	TimeoutSeconds int64 `json:"timeoutSeconds"`
 	// The number of unhealthy health checks required before a host is marked unhealthy
 	// +optional
-	UnhealthyThresholdCount uint32 `json:"unhealthyThresholdCount"`
+	// +kubebuilder:validation:Minimum=0
+	UnhealthyThresholdCount int64 `json:"unhealthyThresholdCount"`
 	// The number of healthy health checks required before a host is marked healthy
 	// +optional
-	HealthyThresholdCount uint32 `json:"healthyThresholdCount"`
+	// +kubebuilder:validation:Minimum=0
+	HealthyThresholdCount int64 `json:"healthyThresholdCount"`
 }
 
 // Delegate allows for delegating VHosts to other IngressRoutes
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -205,7 +205,8 @@ type Service struct {
 	Protocol *string `json:"protocol,omitempty"`
 	// Weight defines percentage of traffic to balance traffic
 	// +optional
-	Weight uint32 `json:"weight,omitempty"`
+	// +kubebuilder:validation:Minimum=0
+	Weight int64 `json:"weight,omitempty"`
 	// UpstreamValidation defines how to verify the backend service's certificate
 	// +optional
 	UpstreamValidation *UpstreamValidation `json:"validation,omitempty"`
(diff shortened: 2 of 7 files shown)


apis/contour/v1beta1/ingressroute.go
```diff
@@ -86,7 +86,8 @@ type Service struct {
 	Port int `json:"port"`
 	// Weight defines percentage of traffic to balance traffic
 	// +optional
-	Weight uint32 `json:"weight,omitempty"`
+	// +kubebuilder:validation:Minimum=0
+	Weight int64 `json:"weight,omitempty"`
```
davecheney: why not just `int`?

apis/contour/v1beta1/ingressroute.go
```diff
@@ -115,10 +116,12 @@ type HealthCheck struct {
 	TimeoutSeconds int64 `json:"timeoutSeconds"`
 	// The number of unhealthy health checks required before a host is marked unhealthy
 	// +optional
-	UnhealthyThresholdCount uint32 `json:"unhealthyThresholdCount"`
+	// +kubebuilder:validation:Minimum=0
+	UnhealthyThresholdCount int64 `json:"unhealthyThresholdCount"`
```
davecheney: same as above

awprice:
Upon reading https://book.kubebuilder.io/cronjob-tutorial/api-design.html?highlight=integer#designing-an-api it also recommends not using `int` too and only use int32/int64. I felt like changing these everywhere would be out of scope for this change. 

Should I raise a second ticket for this?

awprice:
> Thanks for working on this, same comment everywhere, if we can just say `int` not `int64` throughout I'd be fine.
> 
> Also, is it just uint32 that (can't remember the thing that complained) will `uint` work?

Thanks for the quick reply! See comment https://github.com/projectcontour/contour/pull/2133#issuecomment-576505597 about usage of `int`/`uint`.

ash2k:
Hi all :)

From https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types :

> - Do not use unsigned integers, due to inconsistent support across languages and libraries. Just validate that the integer is non-negative if that's the case.
> 
> - All public integer fields MUST use the Go (u)int32 or Go (u)int64 types, not (u)int (which is ambiguous depending on target platform). Internal types may use (u)int.

ash2k:
Also see https://github.com/kubernetes/kubernetes/pull/62981 and https://github.com/kubernetes/kubernetes/pull/63522

youngnick:
@awprice if you can fix up the merge conflicts, I'll merge this one.

awprice:
@youngnick Thanks for the reminder. Conflicts fixed.

PR #2148, opened 2020-01-24
internal/contour: use internal RouteConfiguration helper

jpeach:


internal/contour/route.go
```diff
@@ -93,20 +93,10 @@ type routeVisitor struct {
 }
 
 func visitRoutes(root dag.Vertex) map[string]*v2.RouteConfiguration {
-	headers := envoy.Headers(
```
stevesloka: Why are you removing the header piece?

jpeach: The helper always adds the header.

PR #2158, opened 2020-01-26
Add kubebuilder enum validation to limit the values for service.protocol

### LINKED ISSUE #2152: api: Add enum validation for protocol field in HTTPProxy
The set of values for the field `protocol` in an HTTPProxy's service is limited to a specific set of options. We should set validation on these fields to only allow a specific set of values so users cannot configure invalid values.

_Originally posted by @davecheney in https://github.com/projectcontour/contour/issues/1962#issuecomment-578288440_

stevesloka:
Fixes #2152 by adding an enum validation to limit the values for the service.protocol field. 

```
Release Note: Users will need to reapply the crd spec to get the validation
```

### DIFF
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -197,7 +197,8 @@ type Service struct {
 	// Port (defined as Integer) to proxy traffic to since a service can have multiple defined.
 	Port int `json:"port"`
 	// Protocol may be used to specify (or override) the protocol used to reach this Service.
-	// Values may be tls, h2, h2c.  It omitted protocol-selection falls back on Service annotations.
+	// Values may be tls, h2, h2c. If omitted, protocol-selection falls back on Service annotations.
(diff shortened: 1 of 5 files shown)


apis/projectcontour/v1/httpproxy.go
```diff
@@ -197,7 +197,8 @@ type Service struct {
 	// Port (defined as Integer) to proxy traffic to since a service can have multiple defined.
 	Port int `json:"port"`
 	// Protocol may be used to specify (or override) the protocol used to reach this Service.
-	// Values may be tls, h2, h2c.  It omitted protocol-selection falls back on Service annotations.
+	// Values may be tls, h2, h2c.  If omitted protocol-selection falls back on Service annotations.
```
jpeach: Nit: comma after "omitted"

davecheney: One space after the period please.

stevesloka: :+1:

Issue #2150, opened 2020-01-24
Failed to Apply contour-certgen Job Manifest!


ryanelian:
**What steps did you take and what happened:**

`kubectl apply -f https://raw.githubusercontent.com/projectcontour/contour/v1.1.0/examples/contour/02-job-certgen.yaml`

This error appears:
```
[...]

PR #2159, opened 2020-01-27
internal/certgen: add unit test for createFile

davecheney:
Updates #2150

Some extra test coverage that fell out of investigating #2150.


davecheney:
Given that its only used for local testing, I agree, replacing it with WriteFile would be simpler (and wouldn't leak a file handle I the caller).

PR #2170, opened 2020-01-29
ci: switch to golangci-lint for linting

jpeach:
golangci-lint simplifies the CI tooling by packaging a full suite
of linters and giving us improved control over linting errors.


davecheney:
I’m on the fence about this one. One if the nice, if homely, aspects of the previous make test target was it worked the same locally as it does in ci. I’d be sad to lose that property.

davecheney:
Sorry, ignore my previous comment. I’m an idiot.

Issue #145, opened 2018-01-13
Cleanly draining envoy with DaemonSet/NLB deployment


cmaloney:
sometimes it's necessary to re-deploy newer versions of contour / envoy into the cluster, and that involves killing off the running containers and replacing them with new ones.

If I'm running with an AWS ALB or classic load balancer I could probably add http healthchecks for the envoy health check endpoint + pod lifecycle hooks to make envoy stat failing it's healthcheck so it can be cleanly removed from the nlb without losing traffic

For my deployment to get HTTP2 and minimal hops, I'm using the DaemonSet / NLB deployment of contour and envoy. The NLB does TCP forwarding to contour. Currently, it healthchecks the port of contour it's forwarding traffic to (8080 or 8443). When I want / need to replace envoy, the pod is killed, envoy stops listening, but not long enough that the AWS NLB fails health checks and redirects traffic to the other host without having some lost traffic.

I'd like to help contribute closing this gap so that envoy and contour can safely / cleanly be upgraded without loss of traffic.

Two thoughts on how to do this:
1. Figure out how to get it so the pods can be replaced with the envoy "hot restart" pieces (some form of https://www.envoyproxy.io/docs/envoy/latest/operations/hot_restarter) without dropping any traffic
2. Teach contour to start an additional listener on a configurable port which is the "healthcheck" listener. All AWS NLBs get configured to use that port for whether a given instance should be used. When contour recieves a `SIGTERM` as part of it's lifecycle hooks (https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods) it shuts down this extra listener, which should cause health checks to fail and the removal of the instance from new traffic through the AWS NLB. After a configurable number of seconds from recieving SIGTERM, contour should put envoy into its "drain down" state where it stops accepting incoming connections / drains every existing connection. After the grace period for the pod finally comes up, or if envoy is fully drained, then the pod is killed and remaining pieces die non-gracefully / have their connections cut off.

I think the second option is easier to implement and get working correctly in all cases than the first, the first I see potential issues with the envoy<->contour communication conflicting between "new" and old pods, as well as how you cleanly roll out a replacement of an existing daemon set

davecheney:
Thanks for raising this issue. There's a PR in fly for adding a healthz to envoy and contour, #135.

cmaloney:
Healthz alone wouldn't solve it though with an NLB, since there are only TCP health checks available for a TCP NLB, but healthz on a extra port could definitely work

[...]

PR #2177, opened 2020-01-29
design: Design doc for Envoy Shutdown logic 

stevesloka:
Updates #145 by adding a design doc to handle Envoy rollout across a cluster

### DIFF
--- a/design/envoy-shutdown.md
+++ b/design/envoy-shutdown.md
@@ -1,0 +1,212 @@
+# Envoy Shutdown Probes
+
+Status: _Accepted_
+
+This proposal describes how to add better support for identifying the status of Envoy connections during a restart or redeploy of Envoy.
+
+## Goals
+
+- Provide a mechanism to provide feedback on open connections & load inside an Envoy process
+- Allow for a Envoy fleet roll-out minimizing connection loss
+
+## Non Goals
+
+- Guarantee zero connection loss during a roll-out
+
+## Background
+
+The Envoy process, the data path component of Contour, at times needs to be re-deployed.
+This could be due to an upgrade, a change in configuration, or a node-failure forcing a redeployment.
+
+Contour currently implements a `preStop` hook in the container which signals to Envoy to begin draining connections.
+Once Envoy begins draining, the readiness probe on the pod triggers to fail, which in turn causes that instance of Envoy to stop accepting new connections.
+
+The main problem is that the `preStop` hook (which sends the `/healthcheck/fail` request) does not wait until Envoy has drained all connections, so when the container
+is restarted, users can receive errors.
+
+This design looks to add a new component to Contour which will allow for a way to understand if open connections exist in Envoy before sending a `SIGTERM`.
+
+## High-Level Design
+
+Implement a new sub-command to Contour named `envoy shutdown-manager` which will handle sending the healthcheck fail request 
+to Envoy and then begin polling the http/https listeners for active connections from the `/stats` endpoint available on 
+`localhost:9001`.
+
+Additionally, an optional `min-open-connections` parameter will be added which will allow users to define the minimum number of
+open connections that can be open when waiting for connections to drain.
+
+A `preStop` hook (https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/) in Kubernetes allows time for a container to do any cleanup or extra processing before getting sent a SIGTERM.
+This lifecycle hook blocks the process during this cleanup time and then returns when its ready to be shutdown.
+
+## Detailed Design
+
+- Implement new sub-command in Contour called `envoy shutdown-manager`
+- This command will expose an http endpoint over a specific port (e.g. `8090`) and path `/shutdown`. 
+- A new container will be added to the Envoy Daemonset pod which will run this new sub-command
+- This new container as well as the Envoy container will be updated to use an http preStop hook (see example below)
+- When the pod gets a request to shutdown, the `preStop` hook will send a request to `localhost:8090/shutdown` which will tell Envoy to begin draining connections as well as start polling for active connections and will block until that reaches zero or the `min-open-connections` is met
+- The `terminationGracePeriodSeconds` in the pod will be extended to a larger value (from the default 30 seconds) to allow time to drain connections. If this time limit is met, Kubernetes will SIGTERM the pod and kill it
+- A second endpoint, `/healthz`, will be made available to check for the health of this container which will be implemented in a Kubernetes liveness probe in the event the shutdown-manager exits
+
+```yaml
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  annotations:
+  labels:
+    app: envoy
+  name: envoy
+  namespace: projectcontour
+spec:
+  revisionHistoryLimit: 10
+  selector:
+    matchLabels:
+      app: envoy
+  template:
+    metadata:
+      annotations:
+        prometheus.io/path: /stats/prometheus
+        prometheus.io/port: "8002"
+        prometheus.io/scrape: "true"
+      creationTimestamp: null
+      labels:
+        app: envoy
+    spec:
+      automountServiceAccountToken: false
+      containers:
+      - command:                          # <----- New Pod
+        - /bin/shutdown-manager          
+        image: stevesloka/envoyshutdown
+        imagePullPolicy: Always
+        lifecycle:
+          preStop:                       # <----- PreStop Hook
+            httpGet:
+              path: /shutdown
+              port: 8090
+              scheme: HTTP
+         livenessProbe:                  # <------ Liveness probe  
+           httpGet:
+             path: /healthz
+             port: 8090
+           initialDelaySeconds: 3
+           periodSeconds: 10
+        name: shutdown-manager
+        terminationMessagePath: /dev/termination-log
+        terminationMessagePolicy: File
+      - args:
+        - -c
+        - /config/envoy.json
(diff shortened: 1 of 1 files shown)


design/envoy-shutdown.md
```diff
@@ -0,0 +1,51 @@
...
+Implement a new sub-command of Contour named `envoyshutdown` which will handle sending the healthcheck fail request to Envoy and then begin polling the http/https listeners for active connections from 
+the `/stats` endpoint available on `localhost:9001`.
+
+## Detailed Design
+
+- Implement new subcommand in Contour called `envoyshutdown`
+- This command will expose an http endpoint over a specific port (e.g. `8090`) and path `/shutdown`. 
+- A new container will be added to the Envoy Daemonset pod which will run this new sub-command (e.g. `contour envoyhealthcheck`)
+- This new container as well as the Envoy container will be updated to use an http preStop hook (see example below)
+- When the pod gets a request to shutdown, the `preStop` hook will send a request to `localhost:8090/shutdown` which will tell Envoy to begin draining connections as well as start polling for active connections and will block until that reaches zero
```
jpeach: This will need some configuration options. Some workloads will never drain to 0 or can take hours to drain to low numbers of connections. We should plan to have a configurable maximum drain time, and a connection count threshold.

design/envoy-shutdown.md
```diff
@@ -0,0 +1,51 @@
...
+
+Implement a new sub-command of Contour named `envoyshutdown` which will handle sending the healthcheck fail request to Envoy and then begin polling the http/https listeners for active connections from 
+the `/stats` endpoint available on `localhost:9001`.
+
+## Detailed Design
+
+- Implement new subcommand in Contour called `envoyshutdown`
+- This command will expose an http endpoint over a specific port (e.g. `8090`) and path `/shutdown`. 
+- A new container will be added to the Envoy Daemonset pod which will run this new sub-command (e.g. `contour envoyhealthcheck`)
+- This new container as well as the Envoy container will be updated to use an http preStop hook (see example below)
```
jpeach: The example is missing?

design/envoy-shutdown.md
```diff
@@ -0,0 +1,51 @@
...
+Once Envoy begins draining, the readiness probe on the pod triggers to fail, which in turn causes that instance of Envoy to stop accepting new connections.
+
+The main problem is that the `preStop` hook (which sends the `/healthcheck/fail` request) does not wait until Envoy has drained all connections, so when the container
+is restarted, users can receive errors.
+
+This design looks to add a new component to Contour which will allow for a way to understand if open connections exist in Envoy before sending a `SIGTERM`.
+
+## High-Level Design
+
+Implement a new sub-command of Contour named `envoyshutdown` which will handle sending the healthcheck fail request to Envoy and then begin polling the http/https listeners for active connections from
```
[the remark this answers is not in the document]
jpeach: Or maybe "contour envoy shutdown-manager"?

design/envoy-shutdown.md
```diff
@@ -0,0 +1,51 @@
...
+## High-Level Design
+
+Implement a new sub-command of Contour named `envoyshutdown` which will handle sending the healthcheck fail request to Envoy and then begin polling the http/https listeners for active connections from 
+the `/stats` endpoint available on `localhost:9001`.
+
+## Detailed Design
+
+- Implement new subcommand in Contour called `envoyshutdown`
+- This command will expose an http endpoint over a specific port (e.g. `8090`) and path `/shutdown`. 
+- A new container will be added to the Envoy Daemonset pod which will run this new sub-command (e.g. `contour envoyhealthcheck`)
```
jpeach: Since this is a second pod with a long-running process in it, what precautions should it take to avoid preventing pod shutdown? i.e. if envoy exits we probable want the pod to exit

design/envoy-shutdown.md
```diff
@@ -0,0 +1,51 @@
...
+This design looks to add a new component to Contour which will allow for a way to understand if open connections exist in Envoy before sending a `SIGTERM`.
+
+## High-Level Design
+
+Implement a new sub-command of Contour named `envoyshutdown` which will handle sending the healthcheck fail request to Envoy and then begin polling the http/https listeners for active connections from 
+the `/stats` endpoint available on `localhost:9001`.
+
+## Detailed Design
+
+- Implement new subcommand in Contour called `envoyshutdown`
```
youngnick: Alternate name suggestions:
- `envoyterminator`
- `envoyreaper`

if you feel like being cute. Otherwise, `envoyshutdown` is fine.

design/envoy-shutdown.md
youngnick: Example missing?

Also, does the Envoy container need a prestop hook if this one has one? I don't quite understand how that works. Does the Envoy prestop hook hit the new container as well? Or something in the Envoy container?

Either the example or some more explanation would probably help.

design/envoy-shutdown.md
```diff
@@ -0,0 +1,51 @@
...
+- This command will expose an http endpoint over a specific port (e.g. `8090`) and path `/shutdown`. 
+- A new container will be added to the Envoy Daemonset pod which will run this new sub-command (e.g. `contour envoyhealthcheck`)
+- This new container as well as the Envoy container will be updated to use an http preStop hook (see example below)
+- When the pod gets a request to shutdown, the `preStop` hook will send a request to `localhost:8090/shutdown` which will tell Envoy to begin draining connections as well as start polling for active connections and will block until that reaches zero
+- The `terminationGracePeriodSeconds` in the pod will be extended to a larger value (from the default 30 seconds) to allow time to drain connections. If this time limit is met, Kubernetes will SIGTERM the pod and kill it
+
+## Alternatives Considered
+
+- Bash scripts could be used to do this, however it's more difficult to implement and hard to test. 
+- Instead of using the http preStop hook, we could call a binary to do the checks, however, it's difficult to get this into the Envoy pod reliably (we'd have to use some shared volume)
```
jpeach: s/pod/container/

design/envoy-shutdown.md
stevesloka: I couldn't think up a good name. It's really a "wait-for-envoy-so- the-pod-doesnt-shutdown" thing. =)

design/envoy-shutdown.md
stevesloka: Ahh right, forgot to add the examples (DOH). 

Since we can't have the hook be in the Envoy container, both containers will need a hook to stay up.

design/envoy-shutdown.md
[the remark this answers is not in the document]
stevesloka: It stutters to me a bit to have that extra piece, but pretty much the same as what I had originally. The extra sub-command does allow us to do extra stuff (like maybe inspect the Envoy tables in a future feature)? 

I also toyed with the idea of a new project to run this out of, but then that adds a new build process, etc, which I don't think this really warrants.

design/envoy-shutdown.md
stevesloka: The max time can be configured by the termination grace period, if that time is met, the Kubernetes will forcibly kill everything. 

I don't see a need to add an extra timeout given this one exists currently.

jpeach: Yeh the timeout is arguable, but the connection count threshold is still needed IMHO.

stevesloka: sounds good, I added a new parameter

design/envoy-shutdown.md
stevesloka: We could add a liveliness probe in this new pod to check for its health. 

I think managing what to do if envoy would exit is out of scope for this proposal.

design/envoy-shutdown.md
stevesloka: The termination grace period applies to the pod, not an individual container within the pod.

design/envoy-shutdown.md
jpeach: My concern is that if envoy dies and the shutdown manager does not, the pod won't exit and will never be re-scheduled. Conversely, to shutdown envoy gracefully, we need the manager pod to always be available. so it probably needs a restart policy?

design/envoy-shutdown.md
youngnick: Ah, okay. And both the Envoy container and the new container use the same prestophook, calling to the /shutdown endpoint on port 8090 (which is serviced by the new container). This will effectively make the two containers coterminous (although it won't be at exactly the same time because networks handwave handwave). Nice.

Is the idea that the `/healthz` endpoint on the new container will keep track of Envoy and become unhealthy if Envoy does? Otherwise, an Envoy lockup could mean that things get stuck maybe? I'm not sure how this would work, but it can probably be an implementation detail.

stevesloka: The `/healthz` endpoint was just on the shutdown container since it's a long-running process and there were concerns that if it was down, the proper shutdown sequence wouldn't happen correctly. 

Do you think this design should account for if Envoy fails somehow? There are a number of scenarios that we'd need to look at, but I think it should be a new issue:

1. **Envoy terminates:** Kubernetes should restart
2. **Envoy stops responding**: The readiness probe existing today should take that Envoy out of service (Needs addressed since there's no automatic remedy for this). A liveliness probe could be used here but would need to be very cautious about how this gets implemented. 
3. **Envoy loses gRPC connection to Contour**: Not handled today

jpeach: I think that I agree with the behaviours @stevesloka enumerates above. Note that envoy doesn't become healthy until it has received an initial configuration from Contour, so in case (3), I would expect Envoy to keep running with its current configuration (and no changes).

Issue #2113, opened 2020-01-14
Add initial support for Ingress V2 types


youngnick:
As part of the work I'm doing with the service-api subproject of SIG-Network, we're building new CRDs which will serve as the definition of what the next version of Ingress needs.

Contour will need to support these, so this issue is to cover figuring out how to wire up the new types into Contour, and put them behind an experimental flag so that we can get this wiring in and update the types as they are worked on.

That will allow us to have a headstart on implementing them as they get closer to being finalised.

Putting the usage of them behind an experimental feature flag should insulate the rest of the project from these concerns, hopefully.

Comments on the approach welcomed.

jpeach:
> Putting the usage of them behind an experimental feature flag should insulate the rest of the project from these concerns

Do you mean a Go build tag? A configuration flag?

youngnick:
A CLI flag, sorry. No config option for now. This is mainly about ensuring that we don't need a long-running feature branch.

youngnick:
To consume the service-apis types in Contour, we first need to be able to import the Go types somehow. There are three options:
- Generate the Informers code for the new types the same way we do for Contour's types.
- Move Contour to use the dynamic client, so we don't need any generated code at all.
- Move Contour to use the same controller-runtime/kubebuilder framework that the service-apis use (#2118)

I spent some time spiking the controller-runtime/kubebuilder framework, and it probably is worth doing at some point, but it's a substantial amount of change, and would require careful design. It's probably too much work for this point in time. The PR with a design was #2120, and I have a branch in my Contour fork with the changes if we want to come back to them. (https://github.com/youngnick/contour/tree/reconcile-style)

I also had a look at generating the Informers code, but it seems that the code-generation tools do some magic around the use of `k8s.io` for package names that make using them for generating the service-apis types difficult. I've created https://github.com/kubernetes-sigs/service-apis/issues/61 in the service-apis repo, and I'll update that with my results shortly.

@stevesloka has done a great job of trying out the dynamic client in #2191, and that looks like a viable option. Steve and I would both like @davecheney to take a look before we go any further though.

Once we have a way to import the types, the next steps are:
- Create the new CLI flag (`--experimental-service-apis`) to enable use of the new alpha types
- Add watching of the new types to Contour behind that flag
- Add placeholder handling of the new types into the eventhandler. To begin with, just logging that we've seen them is okay.

Once we've made a call about the types, I'll create issues covering this work.

youngnick:
[...]

Issue #2155, opened 2020-01-24
Contour should document the minimum version of kubernetes required


davecheney:
At the moment we give users little guidance on the versions of kuberenetes contour is compatible with. 

For the most part compatibility is derived from the version of the kubernetes client (client-go) that Contour is built against. client-go itself publishes guidance on which versions of kubernetes is it compatible with. 

In addition contour may from time to time require features available in a specific kubernetes version. 

I propose that we add a new compatibility page, similar to the one we publish for contour/envoy compatibility. It should list major/minor contour version against major/minor minimum kubernetes version. 

From the version of the client used for contour 1.0 and 1.1, the minimum kubernetes version is 1.14, assuming that the aforementioned contour versions build against version 1.16 of client-go.

davecheney:
cc @michmike @youngnick

jpeach:
Duplicate of #1906

PR #2191, opened 2020-02-01
Use client-go dynamic client instead of generated

stevesloka:
Updates #2113 

This switches from using the generated client for custom CRD types to the dynamic. Conversion from `Unstructured` to types objects happens in `internal/contour/handler.go` so all other logic afterward stays consistent. 


stevesloka:
@youngnick @jpeach @davecheney I'd like to add some tests around `Unstructured` to validate that workflow, but I think this is ready for an initial look.

jpeach:
What happens to `deepcopy-gen` here? We use `DeepCopy` in `internal/k8s/status.go`, and this PR doesn't delete the `zz_generated.deepcopy.go` files, but it does delete the script that runs `deepcopy-gen`.

cmd/contour/serve.go
```diff
@@ -179,11 +196,19 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 
 	// step 4. register our resource event handler with the k8s informers.
 	var informers []cache.SharedIndexInformer
+	dynamicFactory := dynamicinformer.NewFilteredDynamicSharedInformerFactory(dynamicClient, 0, v1.NamespaceAll, nil)
+
+	dynamicClient.Resource(schema.GroupVersionResource{Group: "projectcontour.io", Version: "v1", Resource: "httpproxies"})
+	ingressRouteInformer := dynamicFactory.ForResource(schema.GroupVersionResource{Group: "contour.heptio.com", Version: "v1beta1", Resource: "ingressroutes"})
+	tlscertIRInformer := dynamicFactory.ForResource(schema.GroupVersionResource{Group: "contour.heptio.com", Version: "v1beta1", Resource: "tlscertificatedelegations"})
+	httpProxyInformer := dynamicFactory.ForResource(schema.GroupVersionResource{Group: "projectcontour.io", Version: "v1", Resource: "httpproxies"})
+	tlscertProxyInformer := dynamicFactory.ForResource(schema.GroupVersionResource{Group: "projectcontour.io", Version: "v1", Resource: "tlscertificatedelegations"})
```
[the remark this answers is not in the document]
stevesloka: Actually we have this already in the crd, "register.go", so can remove this.

cmd/contour/serve.go
```diff
@@ -396,3 +425,13 @@ func startInformer(inf informer, log logrus.FieldLogger) func(stop <-chan struct
 		return nil
 	}
 }
+
+func startDynamicInformer(inf cache.SharedIndexInformer, log logrus.FieldLogger) func(stop <-chan struct{}) error {
+	return func(stop <-chan struct{}) error {
+		log.Println("starting")
```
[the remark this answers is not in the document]
stevesloka: The informers are different types and have different ways to start. `informer` uses `Start()` and `cache.SharedIndexInformer` uses `Run()`. I'm not sure we can combine into the same method.

internal/contour/handler.go
```diff
@@ -104,6 +122,80 @@ func (e *EventHandler) Start() func(<-chan struct{}) error {
 	return e.run
 }
 
+func (e *EventHandler) ConvertUnstructured(obj interface{}) interface{} {
```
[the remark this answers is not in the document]
stevesloka: Yeah good point, I bounced around this idea while building. I'll move this out.

internal/contour/handler.go
```diff
@@ -19,17 +19,28 @@ package contour
...
 	"github.com/projectcontour/contour/internal/dag"
 	"github.com/projectcontour/contour/internal/k8s"
 	"github.com/projectcontour/contour/internal/metrics"
 	"github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// Group Version for projectcontour.io/v1
+var projectcontourv1GV = schema.GroupVersion{Group: "projectcontour.io", Version: "v1"}
+var ingressroutev1beta1GV = schema.GroupVersion{Group: "contour.heptio.com", Version: "v1beta1"}
```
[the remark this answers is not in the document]
stevesloka: Thanks for pointing this out, I moved to use this, I think it makes sense to keep this constant here with the CRD spec.

internal/contour/handler.go
```diff
@@ -104,6 +122,80 @@ func (e *EventHandler) Start() func(<-chan struct{}) error {
...
+		case "HTTPProxy":
+			proxy := &projectcontour.HTTPProxy{}
+			err := e.scheme.Convert(obj, proxy, nil)
+			if err != nil {
+				e.WithField("name", obj.GetName()).
+					WithField("namespace", obj.GetNamespace()).
+					WithField("kind", obj.GetKind()).
+					WithField("version", obj.GetAPIVersion()).
+					Error(err)
+				return nil
```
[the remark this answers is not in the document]
stevesloka: After moving to k8s package, I change this to return an error so the caller can manage.

internal/k8s/status_test.go
```diff
@@ -1,115 +0,0 @@
-// Copyright © 2019 VMware
```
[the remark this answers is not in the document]
stevesloka: They use a `fake` client from the client-go project, now that we're using a dynamic client, there isn't a matching fake client.

stevesloka:
> What happens to deepcopy-gen here? We use DeepCopy in internal/k8s/status.go, and this PR doesn't delete the zz_generated.deepcopy.go files, but it does delete the script that runs deepcopy-gen.

I added the same code the controller-runtime uses to generate the DeepCopy methods in the makefile.

internal/contour/handler_test.go
```diff
@@ -0,0 +1,279 @@
...
+
+	run(t, "irtlscertunstructured", testcase{
+		obj:  irTLSCertUnstructured,
+		want: irTLSCert1,
+	})
+
+	run(t, "proxytlscertunstructured", testcase{
+		obj:  proxyTLSCertUnstructured,
+		want: proxyTLSCert1,
+	})
```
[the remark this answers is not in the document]
stevesloka: I added an unknown type. I'm struggling to find a way to test that it errors with a known type.

cmd/contour/serve.go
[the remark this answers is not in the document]
jpeach: Ok, this part is extremely confusing.

The `informer` interface actually matches on the `coreinformers.SharedInformerFactory` type, because that type implements `(*internalinterfaces.SharedInformerFactory.Start)`. So we refer to both informers and informer factory as "informers" in the helper functions without being really clear about the differences.

So `dynamicFactory` is an object of type `*dynamic informer.DynamicSharedInformerFactory`, and it is this object that you want to `Start()` using `startInformer(). You only need to start the factory, not each informer that the factory generates.

We can improve this further (probably in a separate PR) by calling `WaitForCacheSync` on the factories rather than on each individual informer.

internal/k8s/converter.go
```diff
@@ -0,0 +1,82 @@
...
+// ConvertUnstructured converts an unstructured.Unstructured to typed struct
+// Any other type passed other than unstructured is just returned back to caller
+func (c *Converter) ConvertUnstructured(obj interface{}) (interface{}, error) {
+	if c.scheme == nil {
+		c.scheme = runtime.NewScheme()
+
+		// Setup converter to understand custom CRD types
+		_ = projectcontour.AddKnownTypes(c.scheme)
+		_ = ingressroutev1.AddKnownTypes(c.scheme)
+	}
```
jpeach: Nit: probably a bit cleaner to pull this into a `NewConverter()` API.

jpeach:
> On 7 Feb 2020, at 6:07 am, Steve Sloka <notifications@github.com> wrote:
> 
> I added an unknown type. I'm struggling to find a way to test that it errors with a known type.

What if you provide an Unstructured object of a know type where a known field is of the wrong type. For example, make "spec" an integer rather than a map/

cmd/contour/serve.go
```diff
@@ -179,11 +196,19 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 
 	// step 4. register our resource event handler with the k8s informers.
 	var informers []cache.SharedIndexInformer
+	dynamicFactory := dynamicinformer.NewFilteredDynamicSharedInformerFactory(dynamicClient, 0, v1.NamespaceAll, nil)
+
+	dynamicClient.Resource(schema.GroupVersionResource{Group: "projectcontour.io", Version: "v1", Resource: "httpproxies"})
+	ingressRouteInformer := dynamicFactory.ForResource(schema.GroupVersionResource{Group: "contour.heptio.com", Version: "v1beta1", Resource: "ingressroutes"})
```
jpeach: Nit: the `GroupVersionResource` is used in two places, so it might be worth hoisting into constants in each API package. Your call, if you think it's worth it.

go.mod
```diff
@@ -26,7 +26,7 @@ require (
 	google.golang.org/grpc v1.25.1
 	gopkg.in/alecthomas/kingpin.v2 v2.2.6
 	gopkg.in/yaml.v2 v2.2.2
-	k8s.io/api v0.0.0-20190918195907-bd6ac527cfd2
+	k8s.io/api v0.0.0-20191003000013-35e20aa79eb8
```
davecheney: Can we do this in two steps please.

1. Update the k8s client and api types, see also #2155 
2. Then do this PR.

internal/contour/handler.go
```diff
@@ -65,6 +65,9 @@ type EventHandler struct {
 	// seq is the sequence counter of the number of times
 	// an event has been received.
 	seq int
+
+	// Converter handles conversion from unstructured to a type
+	converter k8s.Converter
```
davecheney: I kinda feel that the field and the type both have missed the opportunity to obviate the comment. If k8s.Converter was called k8s.UnstructuredConverter then maybe the commend wouldn't be necessary.

internal/contour/handler.go
```diff
@@ -80,14 +83,34 @@ type opDelete struct {
 }
 
 func (e *EventHandler) OnAdd(obj interface{}) {
+	obj, err := e.converter.ConvertUnstructured(obj)
```
davecheney: Is it possible to pull the e.Error change out into its own change, this looks like a worthwhile change in and of itself.

internal/k8s/converter.go
stevesloka: > suggestion: if you don't care about the result, you doin't have to assign it to anything, ie

Yup I had that originally, but the linter complains about it if you don't handle the error. 

```
internal/k8s/converter.go:39:31: Error return value of `projectcontour.AddKnownTypes` is not checked (errcheck)
                projectcontour.AddKnownTypes(c.scheme)
                                            ^
internal/k8s/converter.go:40:31: Error return value of `ingressroutev1.AddKnownTypes` is not checked (errcheck)
                ingressroutev1.AddKnownTypes(c.scheme)
                                            ^
make: *** [Makefile:102: check-golint] Error 1

Comment (davecheney) [reply internal/k8s/converter.go]: personally I think this is incorrect, the linter needs to shut up, but do what you think is appropriate.

Comment (stevesloka) [reply internal/k8s/converter.go]: I agree with you, let me see if there's a way around this.

Comment (jpeach) [reply internal/k8s/converter.go]: FWIW, you can silence the linter in a specific case with a [comment](https://github.com/golangci/golangci-lint#nolint) `// nolint(errcheck)`. Maybe in this case, we should check the error and panic.

Comment (davecheney) [reply internal/k8s/converter.go]: yeah, you're right, if adding these types to the schema can reasonably fail then we should check the error (hint, check(err) is in scope in this package).

Comment (stevesloka) [reply internal/k8s/converter.go]: I cleaned up the calling code to remove the `error`, it was always returning `nil`, so felt just as weird to check if error.

File: internal/k8s/converter.go [hunk 13]
```diff
@@ -0,0 +1,82 @@
...
+
+	switch obj := obj.(type) {
+	case *unstructured.Unstructured:
+		switch obj.GetKind() {
+		case "HTTPProxy":
+			proxy := &projectcontour.HTTPProxy{}
+			err := c.scheme.Convert(obj, proxy, nil)
+			if err != nil {
```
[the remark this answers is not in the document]
stevesloka: good point! :+1:

internal/contour/handler.go
stevesloka: Hey, not following what you are asking? The `e.Error` is new from the converter logic.

stevesloka:
> What if you provide an Unstructured object of a know type where a known field is of the wrong type. For example, make "spec" an integer rather than a map/

@jpeach I added a case for this by swapping around the port/name in a service. Thanks for the suggestion! =)

go.mod
stevesloka: I submitted #2214 & #2213 which address these two points.

stevesloka:
I think is ready for another look for you all, I've addressed all the review comments. 

@jpeach I moved the GroupResourceVersions up to where they were already declared in the apis. Thanks for that suggestion!

@davecheney I have a quick question on your comment here if you could clarify a bit for me:  https://github.com/projectcontour/contour/pull/2191#discussion_r376835126

internal/k8s/converter.go
```diff
@@ -0,0 +1,75 @@
...
+	ingressroutev1.AddKnownTypes(uc.scheme)
+
+	return uc
+}
+
+// Convert converts an unstructured.Unstructured to typed struct
+// Any other type passed other than unstructured is just returned back to caller
+func (c *UnstructuredConverter) Convert(obj interface{}) (interface{}, error) {
+	switch obj := obj.(type) {
+	case *unstructured.Unstructured:
```
davecheney: I think this out switch can be rewritten as
```
unstructured, ok := obj.(*unstructured.Unstructured)
if !ok {
    return nil, fmt.Errorf("unsupported object type: %v", reflect.TypeOf(obj))
}
switch unstructured.GetKind() {
...
}
```
to reduce the overall indentation in this method

internal/contour/handler.go
```diff
@@ -80,14 +82,34 @@ type opDelete struct {
 }
 
 func (e *EventHandler) OnAdd(obj interface{}) {
+	obj, err := e.UnstructuredConverter.Convert(obj)
```
davecheney: its sort of a bit inconsistent that we refer to an embedded field by its fully qualified name, yet two lines below a different embedded object is called without qualification. I think it might be easier on the reader if either both were referred to by their embedded field names, or both anonymously.

internal/contour/handler.go
```diff
@@ -65,6 +65,8 @@ type EventHandler struct {
 	// seq is the sequence counter of the number of times
 	// an event has been received.
 	seq int
+
+	*k8s.UnstructuredConverter
```
davecheney: Maybe we can decouple this a little with an anonymous interface eg
```
Converter interface{
     Convert(obj interface{}) (interface{}, error)
}
```

internal/k8s/converter.go
stevesloka: This changes up the original idea of what this method did which I think is better. Before you could pass it a known type or an unstructured and it would return, but now it only converts an unstructured. This caused a check in the handler to only apply to the unstructured types.

internal/k8s/status.go
```diff
@@ -168,7 +173,8 @@ func (irs *StatusWriter) setIngressRouteStatus(existing, updated *ingressroutev1
 		return err
 	}
 
-	_, err = irs.Client.ContourV1beta1().IngressRoutes(existing.GetNamespace()).Patch(existing.GetName(), types.MergePatchType, patchBytes)
+	irGVR := schema.GroupVersionResource{Group: "contour.heptio.com", Version: "v1beta1", Resource: "ingressroutes"}
```
youngnick: Shouldn't this use `ingressroutev1.IngressRouteGVR`?

internal/k8s/status.go
```diff
@@ -190,6 +196,7 @@ func (irs *StatusWriter) setHTTPProxyStatus(existing, updated *projcontour.HTTPP
 		return err
 	}
 
-	_, err = irs.Client.ProjectcontourV1().HTTPProxies(existing.GetNamespace()).Patch(existing.GetName(), types.MergePatchType, patchBytes)
+	proxyGVR := schema.GroupVersionResource{Group: "projectcontour.io", Version: "v1", Resource: "httpproxies"}
```
youngnick: I think this should also use `projcontour.HTTPProxyGVR` here.

PR #2193, opened 2020-02-03
Improve server logging.

jpeach:
When we starts and stop a subsystem, mention the subsystem that we are
starting or stopping to log better context around what Contour is doing.

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -133,10 +133,13 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	contourInformers := contourinformers.NewSharedInformerFactory(clients.contour, 0)
 
 	// Create a set of SharedInformerFactories for each root-ingressroute namespace (if defined)
-	var namespacedInformers []coreinformers.SharedInformerFactory
+	namespacedInformers := map[string]coreinformers.SharedInformerFactory{}
+
 	for _, namespace := range ctx.ingressRouteRootNamespaces() {
-		inf := coreinformers.NewSharedInformerFactoryWithOptions(clients.core, 0, coreinformers.WithNamespace(namespace))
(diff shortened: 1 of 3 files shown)


cmd/contour/serve.go
```diff
@@ -385,8 +389,8 @@ type informer interface {
 
 func startInformer(inf informer, log logrus.FieldLogger) func(stop <-chan struct{}) error {
 	return func(stop <-chan struct{}) error {
-		log.Println("starting")
-		defer log.Println("stopped")
+		log.Println("started informer")
```
stevesloka: I don't think this is a big deal to add but is that what the `context` in the logger is meant to describe?

cmd/contour/serve.go
```diff
@@ -133,10 +133,13 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	contourInformers := contourinformers.NewSharedInformerFactory(clients.contour, 0)
 
 	// Create a set of SharedInformerFactories for each root-ingressroute namespace (if defined)
-	var namespacedInformers []coreinformers.SharedInformerFactory
+	namespacedInformers := map[string]coreinformers.SharedInformerFactory{}
+
 	for _, namespace := range ctx.ingressRouteRootNamespaces() {
-		inf := coreinformers.NewSharedInformerFactoryWithOptions(clients.core, 0, coreinformers.WithNamespace(namespace))
-		namespacedInformers = append(namespacedInformers, inf)
+		if _, ok := namespacedInformers[namespace]; !ok {
```
stevesloka: I like this as a map, it also protects from users entering the same namespace multiple times! =)

cmd/contour/serve.go
jpeach: Yes, but the context is not very obvious unless you are are reading logs with the relevant source code in front of you :)

PR #2194, opened 2020-02-03
internal/dag: improve ingress class matching logs

jpeach:
It can take a while to realize very simple errors like mismatched
ingress classes. Add a debug log that can be enabled by developers
to log this issue.


jpeach:
> Maybe we should combine ingressClass with the new method?

Good idea. I inlined and removed `(*KubernetesCache)ingressClass()`.

PR #2197, opened 2020-02-05
Add ADOPTERS.md

jonasrosland:
This will add an ADOPTERS.md file, to track and promote users of Contour and their success stories.

This is a draft PR, as we gather adopters to add into the initial document.

Signed-off-by: Jonas Rosland <jrosland@vmware.com>


youngnick:
Hey @jonasrosland, I think we should merge this empty file as soon as possible, and then it will be easier to get people to add to it...

Issue #2143, opened 2020-01-23
Add support for rotation of xDS/gRPC certificates and keys


tsaarni:
Problem statement: Contour does not provide support for certificate rotation for the control plane (xDS/gRPC) connection between Contour and Envoy.

It shall be possible to replace the certificates and keys and have them taken into use automatically. Rotation shall not cause interruptions for data plane traffic.

Tasks:

- [X] #2198 Add support for Contour to hot-reload certificates and key
- [x] #2333 Generate reloadable bootstrap configuration for Envoy
- [x] #2524 Flip default deployment over to use reloadable bootstrapping
- [x] #2229 Update deployment documentation for certificate rotation
- [x] #2538 certgen doesn’t have enough RBAC permissions to update secrets
- [x] #2537 Stop using ImagePullPolicy of Always
- [x] #2494 Certgen should generate secrets that are compatible with cert-manager

tsaarni:
This topic was discussed at [community meeting](https://hackmd.io/84Xbl4WBTpm7OBhaOAsSiw) on Jan 21th.  As agreed, renewal of certificates and keys is scoped out.

I would like to work with this issue.

Would it be OK to create two PRs:

1. Add support for Contour to hot-reload certificates and key
2. Generate bootstrap configuration according to outcome of https://github.com/envoyproxy/envoy/issues/9359

Is the scope already well understood or should separate design document be created?

davecheney:
SGTM. Thank you.

jpeach:
Tasks:

- [X] #2198 Add support for Contour to hot-reload certificates and key #2198
- [ ] #2333 Generate reloadable bootstrap configuration for Envoy
- [ ] Flip default deployment over to use reloadable bootstrapping
- [ ] Update deployment documentation for certificate rotation

jpeach:
Assigning myself since I'm shepherding @tsaarni's work on this project.

jpeach:
All the task here are complete.

Issue #1994, opened 2019-12-06
TLS certificate fixtures are duplicated many times


davecheney:
There are at least three sets of TLS certificate fixtures in the codebase. Ideally there should be one.

jpeach:
I have a partial prototype to consolidate test fixtures:
```Go
type Fixtures struct {
	secrets       map[string]*v1.Secret
	services      map[string]*v1.Service
	ingresses     map[string]*v1beta1.Ingress
	ingressRoutes map[string]*ingressroutev1.IngressRoute
	httpProxies   map[string]*projcontour.HTTPProxy
}

func (f *Fixtures) Secret(s string) *v1.Secret {
	return f.secrets[s]
}

func (f *Fixtures) Service(s string) *v1.Service {
	return f.services[s]
}

func (f *Fixtures) Ingress(s string) *v1beta1.Ingress {
	return f.ingresses[s]
}

func (f *Fixtures) IngressRoute(s string) *ingressroutev1.IngressRoute {
	return f.ingressRoutes[s]
}

func (f *Fixtures) HTTPProxy(s string) *projcontour.HTTPProxy {
	return f.httpProxies[s]
}
```

PR #2198, opened 2020-02-05
cmd/contour: hot-reload certificates and key

tsaarni:
This change adds support for certificate rotation for XDS gRPC interface
between Contour and Envoy. It is achieved by lazily loading certificates
and key every time new TLS connection is established by Envoy.  

This change addresses only the certificate rotation in Contour (server)
and similar support is needed for Envoy (client) to cover the whole use
case.


tsaarni:
This is just initial stab at #2143, all comments are more than welcome!

At this point I generated two sets of certificates for tests, but I saw #1994 and I wonder would you like to move to certificates that are generated on-the-fly by code?

stevesloka:
Thanks for this @tsaarni! I want to take a look at this soon. =)

jpeach:
I'd also like to review before merging.

tsaarni:
Here is hands-on instruction how I tested this on a deployed system:

1. Use `openssl s_client` to establish TLS connection and to observe the "Not Before" and "Not After" dates in the printed server certificate:
```bash
kubectl -n projectcontour exec \
  $(kubectl -n projectcontour get pod -l app=envoy -o jsonpath='{.items[0].metadata.name}') -- \
  openssl s_client -connect contour:8001 \
  | openssl x509 -text -noout
```
2. Re-create the server certificate and key for Contour and overwrite the old one with it:

```bash
kubectl -n projectcontour create secret tls contourcert \
  --cert=certs/contour.pem --key=certs/contour-key.pem \
  --dry-run -o yaml \
  | kubectl apply -f -
```
3. Wait for about 2 minutes for kubelet to update the secrets to the filesystem
4. Run `openssl s_client` again to observe that the  "Not Before" and "Not After" validity times were updated
5. Do steps 1-4 again to ensure that the process can be repeated multiple times
6. Observe that Envoy is still up and in good state by checking pod state and by running traffic
7. Delete Envoy pod to trigger new TLS connection from client 
```bash
kubectl -n projectcontour delete pod -l app=envoy --now
```
8. Observe that the connection was established successfully by checking Envoy pod readiness state, logs and by running traffic

In this sequence I assume that in step 2. the server certificate was re-created with the same old CA i.e. CA was not re-created.

cmd/contour/servecontext.go
```diff
@@ -232,11 +233,33 @@ func (ctx *serveContext) tlsconfig() *tls.Config {
...
+				log.Fatalf("unable to append certificate in %s to CA pool", ctx.caFile)
+			}
+
+			return &tls.Config{
+				Certificates: []tls.Certificate{cert},
+				ClientAuth:   tls.RequireAndVerifyClientCert,
+				ClientCAs:    certPool,
+				Rand:         rand.Reader,
+			}, nil
+		},
```
jpeach: I think this would be cleaner if you used an inner function to capture the common code:
```Go
func (ctx *serveContext) tlsconfig() *tls.Config {
	err := ctx.verifyTLSFlags()

	loadConfig := func() (*tls.Config, error) {
                // Load certificates, etc. Return the error, don't panic.
        }

        // Attempt to load certificates and key to catch configuration errors early.
        if _, err := loadConfig(); err != nil {
                // Fatal error here ...
        }

        return &tls.Config{
		ClientAuth: tls.RequireAndVerifyClientCert,
		Rand:       rand.Reader,
		GetConfigForClient: func(*tls.ClientHelloInfo) (*tls.Config, error) {
			return loadConfig(), nil
		},
	}
```

cmd/contour/servecontext.go
```diff
@@ -232,11 +233,33 @@ func (ctx *serveContext) tlsconfig() *tls.Config {
...
 	}
 
+	// lazily load certificates and key at TLS handshake to ensure that latest
+	// certificates are used in case they have been rotated
 	return &tls.Config{
 		Certificates: []tls.Certificate{cert},
 		ClientAuth:   tls.RequireAndVerifyClientCert,
 		ClientCAs:    certPool,
 		Rand:         rand.Reader,
+		GetConfigForClient: func(*tls.ClientHelloInfo) (*tls.Config, error) {
```
jpeach: Are we sure that we don't want to cache the `tls.Config` for the common case where the certificates have not changed?

cmd/contour/servecontext_test.go
```diff
@@ -191,9 +200,167 @@ leaderelection:
...
+			serverCredentialsDir: "testdata/2",
+			clientCredentialsDir: "testdata/1",
+			expectedServerCert:   "testdata/2/contourcert.pem",
+			expectError:          true,
+		},
+	}
+
+	// create temporary directory to store certificates and key for the server
+	configDir, err := ioutil.TempDir("", "contour-testdata-")
+	checkErr(t, err)
```
jpeach: In this case, `checkErr` should call `t.Fatalf`. It looks like all the other uses in this file should be fatal too, so we can rename it to `checkFatalErr` or something, and update it.

cmd/contour/servecontext_test.go
```diff
@@ -191,9 +200,167 @@ leaderelection:
...
+			expectedServerCert:   "testdata/1/contourcert.pem",
+			expectError:          false,
+		},
+		"rotating server credentials returns new server cert": {
+			serverCredentialsDir: "testdata/2",
+			clientCredentialsDir: "testdata/2",
+			expectedServerCert:   "testdata/2/contourcert.pem",
+			expectError:          false,
+		},
+		"fail to connect with bad client certificate": {
```
jpeach: IIUC, this fails because the client and server don't agree on the CA root certificate to expect? Let's update the description.

cmd/contour/servecontext.go
tsaarni: In theory, cache flushing could be based on `stat()` while being careful to work around how Kubernetes plays with symbolic links.  Alternatively it could be based on hash over the files, which would anyways require reading the file, so why not just return those instead without hashing?

At least from the performance point of view I cannot see reason to implement caching for this use case.   Maybe cache could support scenario where the files were deleted from filesystem but the in-memory versions would keep the system running.  This would not happen in real-life scenario though...

cmd/contour/servecontext.go
tsaarni: This is good idea and I will do it, but just one question:

I recall that we shortly mentioned at the community meeting, that even in the case when certificates are be loaded at TLS connection establishment, we would keep the panic.  

We have these alternatives:

1. If the callback function returns error, the TLS implementation will send TLS alert to Envoy with message "internal error" and reject the connection establishment.
2. If the callback function returns nil, the TLS implementation will revert back to original `tls.Config`, causing Envoy eventually to reject the connection due to expired server certificate.
3. If the callback function panics, pod will be recreated and potentially might cause crash loop, similarily as the initial call to loadConfig() would do.

I was thinking that 3) panic might be easier for user to troubleshoot from these alternatives, but of course 1) or 2) could also have error print added to them...  What do you think?

jpeach: I agree with you that (3) is the best option. Make it a fatal error if `loadConfig` fails?

cmd/contour/servecontext.go
jpeach: The minimum stat information you need is the mtime and the inode number. I expect that this isn't really that performance sensitive, but wanted to make sure that we had the conversation.

Let's link a separate issue to document the issue and consider caching the TLS context. This will give anyone experiencing problems a place to lodge feedback and ensure that we don't forget about it.

cmd/contour/servecontext.go
tsaarni: It is now fixed accordingly.  

Small note though, since `golangci-lint` was giving following error

```
cmd/contour/servecontext.go:225:37: (*serveContext).tlsconfig$1 - result 1 (error) is always nil (unparam)
        loadConfig := func() (*tls.Config, error) {
                                           ^
```
I removed `error` return value from `loadConfig()` and documented in comments that all failures inside that function will cause immediate program exit.

cmd/contour/servecontext.go
```diff
@@ -221,6 +221,7 @@ func (ctx *serveContext) tlsconfig() *tls.Config {
 	err := ctx.verifyTLSFlags()
 	check(err)
 
+	// load certificates and key to catch configuration errors early
```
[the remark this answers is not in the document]
tsaarni: Sure, I have now rewritten all my comments as full sentences.

cmd/contour/servecontext_test.go
tsaarni: I renamed `checkErr()` to `checkFatalErr()` and changed it to call `t.Fatal`.

cmd/contour/servecontext_test.go
tsaarni: Yes exactly, I updated the description.

cmd/contour/servecontext.go
tsaarni: Ok, yes let's do that!  Could you create this issue, I think you have better idea how to handle this kind of issues that we do not intend to work on, not at least now immediately :)

cmd/contour/servecontext.go
jpeach: Ah, what I meant was to return an error from `loadConfig`, then do a fatal exit in the caller.

cmd/contour/servecontext.go
jpeach: #2228

cmd/contour/servecontext.go
tsaarni: Sorry my bad! Changed it now.

tsaarni: The fatal exits were not tested previously either, but this change lowered coverage and got PR checks to fail. I can extract the inline function to test error separately, or alternatively use tricks from [here](https://stackoverflow.com/questions/26225513/how-to-test-os-exit-scenarios-in-go) to test functions that call `os.Exit()`.

PR #2201, opened 2020-02-05
site: Fix tcpproxy include to be singular

### LINKED ISSUE #2218: api: json field for TCPDelegation is plural when it should be singular
The json field for `HTTPProxy.TCPProxy.Include` is plural (i.e. `includes`) when it should be singular: https://github.com/projectcontour/contour/blob/master/apis/projectcontour/v1/httpproxy.go#L179

The docs match this (https://projectcontour.io/docs/v1.1.0/httpproxy/#tcpproxy-delegation) where it shows the singular version. 

If a user follows the examples in the docs, the delegation won't work properly since when the yaml is parsed, the field is ignored because it doesn't match the struct. 

I opened #2201 to fix the docs to at least allow users to follow along properly, however, the real fix is to set the name to be singular.

stevesloka:
Fixes #2218 

The TCPProxy `include` example was wrong. According to our spec, it should be plural (But I think this is an oversight from conversion from IngressRoute). 

### DIFF
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -180,7 +180,13 @@ type TCPProxy struct {
 	Services []Service `json:"services"`
 	// Include specifies that this tcpproxy should be delegated to another HTTPProxy.
 	// +optional
-	Include *TCPProxyInclude `json:"includes,omitempty"`
+	Include *TCPProxyInclude `json:"include,omitempty"`
+	// IncludesDeprecated allow for specific routing configuration to be appended to another HTTPProxy in another namespace.
+	//
+	// Exists due to a mistake when developing HTTPProxy and the field was marked plural
+	// when it should have been singular. This field should stay to not break backwards compatibility to v1 users.
+	// +optional
+	IncludesDeprecated *TCPProxyInclude `json:"includes,omitempty"`
 	// The health check policy for this tcp proxy
 	// +optional
 	HealthCheckPolicy *TCPHealthCheckPolicy `json:"healthCheckPolicy,omitempty"`
--- a/apis/projectcontour/v1/zz_generated.deepcopy.go
+++ b/apis/projectcontour/v1/zz_generated.deepcopy.go
@@ -486,6 +486,11 @@ func (in *TCPProxy) DeepCopyInto(out *TCPProxy) {
 		*out = new(TCPProxyInclude)
 		**out = **in
 	}
+	if in.IncludesDeprecated != nil {
+		in, out := &in.IncludesDeprecated, &out.IncludesDeprecated
+		*out = new(TCPProxyInclude)
+		**out = **in
+	}
 	if in.HealthCheckPolicy != nil {
 		in, out := &in.HealthCheckPolicy, &out.HealthCheckPolicy
 		*out = new(TCPHealthCheckPolicy)
--- a/examples/contour/01-crds.yaml
+++ b/examples/contour/01-crds.yaml
@@ -914,9 +914,26 @@ spec:
                       format: int32
                       type: integer
                   type: object
-                includes:
+                include:
                   description: Include specifies that this tcpproxy should be delegated
                     to another HTTPProxy.
+                  properties:
+                    name:
+                      description: Name of the child HTTPProxy
+                      type: string
+                    namespace:
+                      description: Namespace of the HTTPProxy to include. Defaults
+                        to the current namespace if not supplied.
+                      type: string
+                  required:
+                  - name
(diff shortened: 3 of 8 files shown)


stevesloka:
The right fix is to make it `include` since it's singular, but I think it's was a left-over from IngressRoute.

stevesloka:
@davecheney We got the json tag wrong on the CRD. @jpeach suggested we change the Struct type to match as well, thoughts? To me, it should be singular but I wanted to get your thoughts as well.

davecheney:
Thanks for cc'ing me.

Looking at `spec.tcpproxy.include` is currently a singular *TCPProxyInclude, not a slice so at first blush it doesn't seem like we should rename it to a plural unless we also change the field to be an array of *TCPProxyInclude documents. But given tcp proxy inclusion is kind of limited, I don't know what problem that solves.

Is this a bug reported in the field? Is there an issue number?

stevesloka:
@davecheney I opened https://github.com/projectcontour/contour/issues/2218 which explains the issue(s).

jpeach:
The reason this issue happened is probably because the person that wrote the examples looked at the Go field name rather than the JSON field name label. I've been burned badly by that a few times. We can mitigate by renaming the Go field to match the JSON field, which will reduce the chance of this re-occurring.

stevesloka:
> We can mitigate by renaming the Go field to match the JSON field

The problem is that is should be singular, not plural. So renaming the field to be plural just makes the problem bigger. We need to allow Contour to handle the singular version and deprecate out this plural one somehow.

davecheney:
The name of the field; `TCPProxy.Include` doesn't appear to be the issue, the problem as I understand it is the JSON tag is `includes` and the CRD validation has picked that up and enforces that the YAML field is called `includes`. I'm not sure we can change this without a compatibility break.

jpeach:
> On 11 Feb 2020, at 2:33 pm, Steve Sloka <notifications@github.com> wrote:
> 
> We can mitigate by renaming the Go field to match the JSON field
> 
> The problem is that is should be singular, not plural. So renaming the field to be plural just makes the problem bigger. We need to allow Contour to handle the singular version and deprecate out this plural one somehow.

I don't disagree that the field should be singular. All I'm saying is that in the interim it is less error-prone if the names match.

davecheney:
> I don't disagree that the field should be singular. All I'm saying is that in the interim it is less error-prone if the names match.

What are you proposing renaming, the `Include` field on `TCPProxy`? I don't think that's the problem because its not visible to anyone outside contour's codebase.

jpeach:
> On 11 Feb 2020, at 2:38 pm, Dave Cheney <notifications@github.com> wrote:
> 
> What are you proposing renaming, the Include field on TCPProxy?

Yes. When the two field names (Go and JSON) match, life is better.

> I don't think that's the problem because its not visible to anyone outside contour's codebase.

The incorrect Go field name was probably the cause of this issue.

youngnick:
We have an API guarantee problem, however. We called this API spec v1, that means we can't remove or deprecate fields. We can add things, but not remove or deprecate things without starting a v2beta1 cycle, if we really want to stick to the guarantee.

I agree that having the json tag and the Go field name not matching is confusing, but we have to keep that in mind.

davecheney:
> The incorrect Go field name was probably the cause of this issue.

But the field is called `Include` singular. It's not a slice.

jpeach:
> On 11 Feb 2020, at 2:44 pm, Dave Cheney <notifications@github.com> wrote:
> 
> The incorrect Go field name was probably the cause of this issue.
> 
> But the field is called Include singular. It's not a slice.

The API is what we send on the wire to the API server "includes".

davecheney:
> The API is what we send on the wire to the API server "includes".

davecheney:
> The API is what we send on the wire to the API server "includes".

yes, I agree, because the JSON tag on that field is incorrect, which got picked up by the generator and CRD validations.

TCPProxy.Include is a singular, there should only be at most one include at this level. I don't agree the fix is to rename `TCPProxy.Include` to `TCPProxy.Includes` so it matches the incorrect JSON tag.

stevesloka:
The right way to fix this is to allow both the `includes` as well as `include` to be processed by the CRD. 

However, by adding a second field to the struct solves the initial problem, but since we're using kubebuilder to generate the CRD, it adds this field to the spec and I don't see a way around hiding that value unless we bump the version of HTTPProxy or stop using kubebuilder to generate the CRD yaml.

stevesloka:
This now enables both the singular and plural version of `HTTPProxy.TCPProxy.Include`.

Issue #2215, opened 2020-02-10
Record and replay Kubernetes objects


jpeach:
It can be very challenging to reproduce cluster workloads in development environments. Not only is it hard to understand and characterize the workload, but if you do, it is equally hard to reproduce that locally. We can address this by recording the Kubernetes objects that Contour sees in the target cluster, then replaying them in a development cluster.

Recording observed Kubernetes object updates is relatively straight-forward. We can just write them to a file (with compression) before handling them. The main issues around recording are
- privacy (don't leak secrets)
- size (log compression and rotation)

Replay is more challenging because we may need to reconstruct any private information that we elided, and we need to attach the replay source in such a way that Contour doesn't actually connect to the API server of the current cluster.

davecheney:
I’m concerned about how this feature will metastasise over time. Can you give some background why this will be useful, ideally I’d like to find other ways to achieve the debugging requirements driving this feature.

jpeach:
> Can you give some background why this will be useful

I tried to address this in the first paragraph of the issue.

stevesloka:
Could this work? https://github.com/heptiolabs/eventrouter

davecheney:
The goal of the e2e and featuretest packages was to solve the problems of reproducing scenarios that require multiple steps. I know that these packages are verbose in their implementing, but do you think that approach is no longer viable?

On 11 Feb 2020, at 10:03, James Peach <notifications@github.com> wrote:

﻿

Can you give some background why this will be useful

I tried to address this in the first paragraph of the issue.

jpeach:
> The goal of the e2e and featuretest packages was to solve the problems of reproducing scenarios that require multiple steps. 

I think these are different use cases. When you have a cluster with a long-running workload that eventually has a problem, those aren't going to help much. they come into play once you can distal a problem down to its essence.

PR #2216, opened 2020-02-10
POC: record Kubernetes object events

jpeach:
This is a minimal POC of an event handler that records Kubernetes
object events to a log file. It's the simplest approach possible,
no compression, no object or field filtering, no configuration.

This updates #2215.


jpeach:
Dropping this PR, since there's no use case for it at the moment.

PR #2227, opened 2020-02-13
cmd/contour: Envoy Shutdown Manager

### LINKED ISSUE #145: Cleanly draining envoy with DaemonSet/NLB deployment

stevesloka:
Fixes #145 by adding a new set of commands to Contour which will watch the Envoy Prometheus endpoint to block the pod from terminating while there are open connections. 

Also adds a sample Grafana panel which shows the open connections by listener:

image


stevesloka:
I'm thinking of adding a flow diagram to the docs to explain the sequence of events and the options.

youngnick:
I think a flow diagram is an excellent idea.

jpeach:
I'm reviewing now.

cmd/contour/shutdownmanager.go
```diff
@@ -0,0 +1,132 @@
...
+		openConnections, err := getOpenConnections(prometheusURL, s.prometheusStat, s.prometheusValues)
+		if err != nil {
+			s.Error(err)
+		} else {
+			if openConnections <= s.minOpenConnections {
+				s.Infof("Found %d open connections with min number of %d connections. Shutting down...", openConnections, s.minOpenConnections)
+				return
+			}
+			s.Infof("Found %d open connections with min number of %d connections. Polling again...", openConnections, s.minOpenConnections)
+		}
```
jpeach: Suggest this log structure:
```
for {
    s.Info("polled envoy connections").
        WithField("open_connections", openConnections).
        WithField("min_connections", s.minOpenConnections)
    ....

   s.Info("allowing shutdown").
        WithField("open_connections", openConnections).
        WithField("min_connections", s.minOpenConnections)
}
```

cmd/contour/shutdownmanager.go
```diff
@@ -0,0 +1,132 @@
...
+	// Make request to Envoy Prometheus endpoint
+	resp, err := http.Get(url)
+	if err != nil {
+		return -1, fmt.Errorf("GET request for URL %q failed: %s", url, err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return -1, fmt.Errorf("GET request for URL %q returned HTTP status %s", url, resp.Status)
+	}
+
```
jpeach: Nit: getOpenConnections and getOpenConnections have identical structure but are formatted differently, do you mind aligning them?

cmd/contour/shutdownmanager.go
```diff
@@ -0,0 +1,132 @@
...
+		FieldLogger: log,
+	}
+	shutdownmgr := cmd.Command("shutdown-manager", "Start envoy shutdown-manager.")
+	shutdownmgr.Flag("check-interval", "Time to poll Envoy for open connections.").Default("5s").DurationVar(&ctx.checkInterval)
+	shutdownmgr.Flag("check-delay", "Time wait before polling Envoy for open connections.").Default("60s").DurationVar(&ctx.checkDelay)
+	shutdownmgr.Flag("min-open-connections", "Min number of open connections when polling Envoy.").Default("0").IntVar(&ctx.minOpenConnections)
+	shutdownmgr.Flag("serve-port", "Port to serve the http server on.").Default("8090").IntVar(&ctx.httpServePort)
+	shutdownmgr.Flag("prometheus-path", "The path to query Envoy's Prometheus HTTP Endpoint.").Default("/stats/prometheus").StringVar(&ctx.prometheusPath)
+	shutdownmgr.Flag("prometheus-stat", "Prometheus stat to query.").Default("envoy_http_downstream_cx_active").StringVar(&ctx.prometheusStat)
+	shutdownmgr.Flag("prometheus-values", "Prometheus values to look for in prometheus-stat.").Default("ingress_http", "ingress_https").StringsVar(&ctx.prometheusValues)
```
jpeach: Perhaps these should not be configurable. Internally, if we use the `ENVOY_HTTP_LISTENER` and `ENVOY_HTTPS_LISTENER` constants, that will let future maintainers know there is a dependency. Since users can't change the listener names that Contour uses, I don't think they can usefully change these settings either.

internal/metrics/parser.go
```diff
@@ -0,0 +1,54 @@
...
+package metrics
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/prometheus/common/expfmt"
+)
+
+// ParseOpenConnections returns the sum of open connections from a Prometheus HTTP request
```
jpeach: The name `ParseOpenConnections`, which is very specific doesn't really match the parameterization, which seems to indicate this is a general function. I'd be inclined to remove the second two arguments and hard-code them as constants. That would make the mapping from there to the envoy metrics clearer, and we wouldn't really lose much flexibility.

internal/metrics/parser.go
```diff
@@ -0,0 +1,54 @@
...
+func ParseOpenConnections(stats io.Reader, prometheusStat string, prometheusValues []string) (int, error) {
```
jpeach: `prometheusValues` is a slice of labels to match, not values.

internal/metrics/parser.go
```diff
@@ -0,0 +1,54 @@
...
+	if _, ok := metricFamilies[prometheusStat]; !ok {
+		return -1, fmt.Errorf("prometheus stat [%s] not found in request result", prometheusStat)
+	}
+
+	// Look up open connections value
+	for _, metrics := range metricFamilies[prometheusStat].Metric {
+		for _, labels := range metrics.Label {
+			for _, item := range prometheusValues {
+				if item == *labels.Value {
+					openConnections += int(*metrics.Gauge.Value)
```
jpeach: I didn't see a convenient way to make this fully safe, but `metrics.Gauge.GetValue()` is a bit safer.

go.mod
```diff
@@ -19,6 +19,7 @@ require (
 	github.com/konsorten/go-windows-terminal-sequences v1.0.2 // indirect
 	github.com/prometheus/client_golang v1.1.0
 	github.com/prometheus/client_model v0.0.0-20190812154241-14fe0d1b01d4
+	github.com/prometheus/common v0.6.0
```
jpeach: I don't know anything about prometheus releases, but should we try to get the 3 packages to all have the same version?

cmd/contour/shutdownmanager.go
```diff
@@ -0,0 +1,132 @@
...
+	w.WriteHeader(http.StatusOK)
+	_, err := w.Write([]byte("OK"))
+	if err != nil {
+		s.Error(err)
+	}
+}
+
+// handles /shutdown
+func (s *shutdownmanagerContext) shutdownHandler(w http.ResponseWriter, r *http.Request) {
+	prometheusURL := fmt.Sprintf("http://%s:%d%s", s.envoyHost, s.envoyPort, s.prometheusPath)
```
youngnick: Do you think it might be worthwhile having this write something back to the connection so you can watch it with curl maybe?

cmd/contour/shutdownmanager.go
```diff
@@ -0,0 +1,132 @@
...
+
+	return g.Run()
+}
+
+// registerShutdownManager registers the envoy shutdown sub-command and flags
+func registerShutdownManager(cmd *kingpin.CmdClause, log logrus.FieldLogger) (*kingpin.CmdClause, *shutdownmanagerContext) {
+	ctx := &shutdownmanagerContext{
+		FieldLogger: log,
+	}
+	shutdownmgr := cmd.Command("shutdown-manager", "Start envoy shutdown-manager.")
```
youngnick: In `serve.go` and `servecontext.go`, we use a `newServeContext()` function, do you think we should do this the same way instead of using the Kingpin defaults? I feel like maybe the function is a little easier place to find the defaults.

cmd/contour/shutdownmanager.go
youngnick: I missed this in my review, but I tend to agree, there's no real reason for these to be configurable by the end user, let's not expose it unless we have to.

site/docs/master/shutdown-manager.md
```diff
@@ -0,0 +1,76 @@
...
+  - Type: integer (Default 0)
+- **serve-port:** Port to serve the http server on.
+  - Type: integer (Default 8090)
+- **prometheus-path:** The path to query Envoy's Prometheus HTTP Endpoint.
+  - Type: string (Default "/stats/prometheus")
+- **prometheus-stat:** Prometheus stat to query.
+  - Type: string (Default "envoy_http_downstream_cx_active")
+- **prometheus-values:** Prometheus values to look for in prometheus-stat.
+  - Type: string array (Default ["ingress_http", "ingress_https"])
+- **envoy-host:** HTTP endpoint for Envoy's stats page.
```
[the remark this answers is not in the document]
stevesloka: I can clarify, this is the "stats" page in the "administration interface". =)

cmd/contour/shutdownmanager.go
stevesloka: Yeah, when I added this I went back & forth in my head. I landed on this because I thought maybe users might want to watch a different value than that we are looking at. But that's not a real use-case now, so will remove.

site/docs/master/shutdown-manager.md
```diff
@@ -0,0 +1,76 @@
...
+Finally, the pod's `terminationGracePeriodSeconds` is customized to extend the time in which Kubernetes will allow the pod to be in the `Terminating` state.
+If during shutdown, the connections aren't drained to the configured amount, the `terminationGracePeriodSeconds` will send a `SIGTERM` to the pod killing it.
+
+<center>
+    ![img][1]
+</center>
+
+### Shutdown Manager Config Options
+
+The shutdown manager has a set of arguments that can be passed to change how it behaves:
```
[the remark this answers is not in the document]
stevesloka: I guess my attempt at keeping it separate was to not end up in the same state that `httpproxy` is where it's a super long page. It sort of needs to be linked from the upgrade guide as well. 

I'll open an issue to track maybe an upgrade approach guide that can live alongside the upgrade guide today.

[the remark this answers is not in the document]
stevesloka: #2246

cmd/contour/shutdownmanager.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+}
+
+// handles /shutdown
+func (s *shutdownmanagerContext) shutdownHandler(w http.ResponseWriter, r *http.Request) {
+	prometheusURL := fmt.Sprintf("http://%s:%d%s", s.envoyHost, s.envoyPort, s.prometheusPath)
+	envoyAdminURL := fmt.Sprintf("http://%s:%d/healthcheck/fail", s.envoyHost, s.envoyPort)
+
+	// Send shutdown signal to Envoy to start draining connections
+	err := shutdownEnvoy(envoyAdminURL)
```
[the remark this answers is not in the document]
stevesloka: My thought was to be optimistic and keep trying. Worst case if the stats endpoint can't be found, it will still wait for the total termination grace period giving the most amount of time for connections to drain.

examples/example-workload/bombardier.job.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: batch/v1
```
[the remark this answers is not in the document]
stevesloka: I'm going to delete this file, without any context it's out of place and doesn't make any sense to the repo.

cmd/contour/shutdownmanager.go
```diff
@@ -0,0 +1,132 @@
...
+		defer config.Info("stopped")
+
+		http.HandleFunc("/healthz", config.healthzHandler)
+		http.HandleFunc("/shutdown", config.shutdownHandler)
+		log.Fatal(http.ListenAndServe(fmt.Sprintf(":%d", config.httpServePort), nil))
+
+		return nil
+	})
+
+	return g.Run()
```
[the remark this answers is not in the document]
stevesloka: Yup good point.

go.mod
stevesloka: `v0.6.0` is the latest for `Common` (https://github.com/prometheus/common/releases/tag/v0.6.0)

We could upgrade the other deps, but not sure what's different in them to move or not.

site/docs/master/shutdown-manager.md
```diff
@@ -0,0 +1,76 @@
...
+2. Start draining existing connections in Envoy by sending a `POST` request to `/healthcheck/fail` endpoint
+3. Wait for connections to drain before allowing Kubernetes to `SIGTERM` the pod
+
+## Overview
+
+Contour implements a new `envoy` sub-command which has a `shutdown-manager` who's job is to manage a single Envoy instances lifecycle for Kubernetes.
+The `shutdown-maanger` runs as a new container alongside the Envoy container in the same pod.
+It exposes two HTTP endpoints which are used for `livenessProbe` as well as to handle the Kubernetes `preStop` event hook.
+
+- **livenessProbe**: Uses to validate the shutdown manager is still running properly. If requests to `/healthz` fail, the container will be restarted.
```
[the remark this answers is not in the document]
stevesloka: ahh good catch =)

cmd/contour/shutdownmanager.go
stevesloka: It writes to the logger later on each time it polls outputting the current connections and the min configured:

```
INFO[0005] min number of open connections found, shutting down...  context=shutdown-manager min_connections=0 open_connections=0
```

cmd/contour/shutdownmanager.go
stevesloka: I have them match now per your suggestion. Good point on making them consistent.

cmd/contour/shutdownmanager.go
youngnick: I meant to the actual HTTP response, but it's probably not a good idea to be curling that endpoint anyway, so it's okay.

cmd/contour/shutdownmanager.go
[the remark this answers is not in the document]
jpeach: Yeh, I think that the stats behavior sense.

Should the shutdown retry too? If the shutdown transiently fails, then sampling the stats is futile. OTOH, if the shutdown can fail for permanent reasons (e.g. envoy is already shutting down), then we would get stuck here.

go.mod
jpeach: Yeh, there doesn't seem to be any version numbering consistency across these packages 🤷‍♂

cmd/contour/shutdownmanager.go
jpeach: To watch it with curl, the shutdown process would need to be separated from the HTTP handler, so that multiple requests could observe it. That separation would also make it robust against the shutdown handler accidentally being called multiple times, but would be more complex.

davecheney:
I think that’s technically v0.9.1 upstream, but for whatever reason go modules doesn’t want to use that version number and is reverting to the hash.

cmd/contour/shutdownmanager.go
stevesloka: I don't see a good use-case to watch with curl. To do that, you'd have to have access to the pod via kubectl since it's not exposed from the pod (i.e port-forward first, then curl). 

I would think you would tail the logs of shutdown-manager to watch what's happening.

youngnick: Yes, that's what I was trying to say earlier, this is fine for now.

cmd/contour/shutdownmanager.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+		time.Sleep(s.checkInterval)
+	}
+}
+
+// shutdownEnvoy sends a POST request to /healthcheck/fail to tell Envoy to start draining connections
+func shutdownEnvoy(url string) error {
+	resp, err := http.Post(url, "", nil)
+	if err != nil {
+		return fmt.Errorf("creating post request for url %q failed: %s", url, err)
```
jpeach: Nit: s/url/URL/, here and below.

Alternatively, dropping the "url" and just saying "request for %q failed" should be just as clear.

internal/metrics/parser.go
```diff
@@ -0,0 +1,60 @@
...
+
+	// Parse Prometheus http response
+	metricFamilies, err := parser.TextToMetricFamilies(stats)
+	if err != nil {
+		return -1, fmt.Errorf("parsing prometheus text format failed: %v", err)
+	}
+
+	// Validate stat exists in output
+	if _, ok := metricFamilies[prometheusStat]; !ok {
+		return -1, fmt.Errorf("prometheus stat [%s] not found in request result", prometheusStat)
```
jpeach: Nit: why do you use `[%s]` here but `%q` in other places?

internal/metrics/parser.go
```diff
@@ -0,0 +1,60 @@
...
+	openConnections := 0
+
+	if stats == nil {
+		return -1, fmt.Errorf("stats input was nil")
+	}
+
+	// Parse Prometheus http response
+	metricFamilies, err := parser.TextToMetricFamilies(stats)
+	if err != nil {
+		return -1, fmt.Errorf("parsing prometheus text format failed: %v", err)
```
jpeach: Nit: Should we capitalize "Prometheus" in log messages?

internal/metrics/parser.go
```diff
@@ -0,0 +1,60 @@
...
+	"fmt"
+	"io"
+
+	"github.com/prometheus/common/expfmt"
+)
+
+const prometheusStat = "envoy_http_downstream_cx_active"
+
+func prometheusLabels() []string {
+	return []string{"ingress_http", "ingress_https"}
```
jpeach: Can you use `ENVOY_HTTP_LISTENER` and `ENVOY_HTTPS_LISTENER` here?

stevesloka:
> Did we reach a resolution here? What is meant to happen to the shutdown process if we couldn't fail the healthcheck out?

@jpeach If we can't tell Envoy to start draining connections then the readiness probe will fail and the pod will wait the total number of terminationGracePeriodSeconds before dropping all the connections that Envoy.

internal/metrics/parser.go
stevesloka: The ENV vars? No, these strings match labels in the prometheus metrics.

internal/metrics/parser.go
stevesloka: I don't know that it matters. I thought you wanted everything in lower case?

stevesloka: It is a "proper noun" so normally I would capitalize.

internal/metrics/parser.go
stevesloka: no reason, what would you prefer?

internal/metrics/parser.go
jpeach: I meant [these](https://github.com/projectcontour/contour/blob/master/internal/contour/listener.go#L33). But I suppose these strings are coded in so many places that one more doesn't hurt :)

Resolve this if you want to keep this as is.

internal/metrics/parser.go
jpeach: Sorry, what I meant previously was that Contour log messages never start with a capital. So Contour wold never log "Parsing a foo".

I agree with you that we should capitalize proper nouns.

internal/metrics/parser.go
jpeach: How about `%q` for consistency?

jpeach:
> we can't tell Envoy to start draining connections then the readiness probe will fail and the pod will wait the total number of terminationGracePeriodSeconds before dropping all the connections that Envoy.

@stevesloka So in that case, are we worse off than before? That is, envoy isn't draining, but the shutdown is going to take the full termination period. If we don't retry here, we are guaranteed to consume the full grace period, right? Whereas if we retry, we might succeed and at least start the draining.

stevesloka:
> So in that case, are we worse off than before? 

Worse off than before what? I'm not following. The current preStop hook does the exact same call and has the exact same behavior.

> That is, envoy isn't draining, but the shutdown is going to take the full termination period. If we don't retry here, we are guaranteed to consume the full grace period, right? 

Yes, we'll use the entire grace period unless somehow the connections all drain by themselves.

> Whereas if we retry, we might succeed and at least start the draining.

I'd prefer to make this a new issue to follow up on. I honestly do not see this as a case that would be hit. I do think it is possible, but I'd prefer to not make this PR hold on this. Thoughts?

jpeach:
> Worse off than before what? I'm not following. The current preStop hook does the exact same call and has the exact same behavior.

The current preStop calls the fail once and then pod shutdown continues. This preStop blocks pod shutdown until the connection count converges. Previously, the preStop would never hold up pod shutdown.

> I'd prefer to make this a new issue to follow up on.

Sure, that seems fine.

internal/metrics/parser.go
stevesloka: oh I see. Hmm, possibly. As it's written now we'd get an import cycle by referencing those since the contour package already references the metrics one.

stevesloka: maybe can lift the const somewhere else. It feels bad to add more to the shutdown manager file.

stevesloka:
Retry issue: https://github.com/projectcontour/contour/issues/2262

site/docs/master/redeploy-envoy.md
```diff
@@ -0,0 +1,65 @@
...
+- **preStop**: This is used to keep the container running while waiting for Envoy to drain connections. The `/shutdown` endpoint blocks until the connections are drained.
+
+```yaml
+ - name: shutdown-manager
+   command:
+   - /bin/contour
+   args:
+     - envoy
+     - shutdown-manager
+   image: docker.io/projectcontour/contour:master
```
davecheney: should be versioned or :latest

Issue #2240, opened 2020-02-17
internal/envoy: SafeRegex maximum program size of 1000 is too conservative


davecheney:
Envoy 1.12 switch to Google's RE2 library for regex processing. RE2 includes a safety feature that requires the caller to supply, along with the input, a maximum size for the compiled DFA. This is exposed to contour via things in the xDS api which take a RE2 regex configuration. If we don't supply a value for the compile DFA size, Envoy enforces a value of 100, which is too low for most use cases.

In #2043, after failing to compute the appropriate DFA size from the input, we raised the limit to 1000. It turns out that this limit can easily be surpassed with some straight forward input strings. For example a 51 character input string caused a 39x explosion in the final DFA size of 2013. This causes Envoy to reject RDS updates which will block further updates until the offending regex is removed.

Simply put the limit does not protect Envoy from anything:

- If we derive the matching limit value from the input string, then, unless we validate the input string for badly written regex expressions, we are effectively disabling the limit.
- If we raise the limit a bit, say 10,000 we're probably just kicking the can down the road until someone tries to match with a more heroic regex.
- If we raise the limit beyond what is sensible -- some of the re2 headers suggest 8mb -- then the limit protects almost nothing -- how long will an 7.99mb DFA take to process input?

The only decision seems to be do we raise the limit a little, eg, 10,000, or a lot, ie 100,000,000.

See also https://github.com/envoyproxy/envoy/pull/9171#discussion_r351974033

jpeach:
xref #2043

sunjayBhatia:
we've set the max regex program size warning level to 1000 and the error level to 2 ^ 20 (both of which will trigger stats to be incremented)

closing

PR #2241, opened 2020-02-17
internal/envoy: raise SafeRegex limit to 1048576

davecheney:
Updates #2240

Raise the SafeRegex size limit from 1,000 to 1048576. There is no evidence
that this number is sufficient for all possible regex patterns, thus the
limit represents the "no limit" limit because it is currently not
possible for envoy to reject a regex entry in a way that Contour can
trace back to the original input.

### DIFF
--- a/internal/envoy/regex.go
+++ b/internal/envoy/regex.go
@@ -25,8 +25,9 @@ import (
 // units of program size. AFAIK, there's no obvious correlation
 // between regex size and execution time.
 //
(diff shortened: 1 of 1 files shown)


jpeach:
The unit for program size is not bytes, it's just size. So a huge limit is fine, just adjust the commit comment.

davecheney:
@jpeach done, PTAL

PR #2242, opened 2020-02-17
Add support for service-apis types

youngnick:
Updates #2113 

The support is guarded by an `--experimental-service-apis` flag, and will, at this stage, only log that the objects were seen.

The purpose here is to verify that all of our wiring is correctly installed and working.


cmd/contour/serve.go
```diff
@@ -205,6 +208,14 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	informers = registerEventHandler(informers, dynamicInformers.ForResource(projectcontour.TLSCertificateDelegationGVR).Informer(), eh)
 	informers = registerEventHandler(informers, coreInformers.Core().V1().Services().Informer(), eh)
 
+	if ctx.UseServiceAPITypes {
+		log.Info("Enabling EXPERIMETNAL SERVICE API TYPES")
```
davecheney: IF YOU'RE GOING TO YELL, PLEASE SPELL EXPERIMENTAL CORRECTLY :)

But seriously, this isn't work yelling about, no need for the shout caps.

examples/contour/01-crds.yaml
```diff
@@ -2,6 +2,8 @@
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.2.4
```
davecheney: gross, is there a way to remove this?

internal/dag/cache_test.go
```diff
@@ -1074,6 +1111,66 @@ func TestKubernetesCacheRemove(t *testing.T) {
...
+			}),
+			obj: &serviceapis.HTTPRoute{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "httproute",
+					Namespace: "default",
+				},
+			},
+			want: true,
+		},
+		"remove service-apis TcPRoute": {
```
davecheney: s/TcPRoute/TcpRoute

cmd/contour/serve.go
```diff
@@ -205,6 +208,14 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+		informers = registerEventHandler(informers, dynamicInformers.ForResource(serviceapis.GroupVersion.WithResource("gatewayclasses")).Informer(), eh)
+		informers = registerEventHandler(informers, dynamicInformers.ForResource(serviceapis.GroupVersion.WithResource("gateways")).Informer(), eh)
+		informers = registerEventHandler(informers, dynamicInformers.ForResource(serviceapis.GroupVersion.WithResource("httproutes")).Informer(), eh)
+		informers = registerEventHandler(informers, dynamicInformers.ForResource(serviceapis.GroupVersion.WithResource("tcproutes")).Informer(), eh)
+	}
```
jpeach: This will be more readable if you flip the arguments to `registerEventHandler` to take varargs of informers:
```
func registerEventHandler(informers []cache.SharedIndexInformer, eh cache.ResourceEventHandler, informersToAdd ...cache.SharedIndexInformer) []cache.SharedIndexInformer {
...
}
```

examples/contour/01-crds.yaml
youngnick: I don't think so, but I'll check.

cmd/contour/serve.go
youngnick: Sure, it was easier to pick it out in caps, but I'll change it.

internal/dag/cache_test.go
youngnick: oKaY I gUesS tHaT mAkEs sEnSe...

cmd/contour/serve.go
youngnick: Okay, that's done.

examples/contour/01-crds.yaml
youngnick: Looks like this is the desired behaviour, we've pulled it in when upgrading controller-gen, it seems. https://github.com/kubernetes-sigs/controller-tools/pull/338 and https://github.com/kubernetes-sigs/controller-tools/issues/348 have more detail.

internal/k8s/converter.go
```diff
@@ -44,7 +45,12 @@ func NewUnstructuredConverter() *UnstructuredConverter {
 	projectcontour.AddKnownTypes(uc.scheme)
 	ingressroutev1.AddKnownTypes(uc.scheme)
 
-	return uc
+	// The kubebuilder tools' contract here is different, yay.
+	if err := serviceapis.AddToScheme(uc.scheme); err != nil {
```
stevesloka: Should this be behind the experimental flag?

cmd/contour/serve.go
```diff
@@ -160,6 +163,12 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		}
 	}
 
+	// Before we can build out the event handler, we need to initialize the converter we'll
+	// use to convert from Unstructured. Thanks to kubebuilder types from service-apis, this now can
+	// return an error.
+
+	converter, err := k8s.NewUnstructuredConverter()
+	check(err)
```
stevesloka: Initialization of the service-apis failing now breaks Contour from working altogether.  Should we change the `check(err)` to just log the error? Or only use `check` if the apis are enabled?

I'm just worried that this code might somehow break users who don't buy into the experimental flag.

examples/contour/01-crds.yaml
stevesloka: I'm confused about this as well. The only tooling we'd need to consume these CRDs is possibly to create the generated `DeepCopy` types.

examples/render/contour.yaml
```diff
@@ -76,6 +76,8 @@ data:
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
+  annotations:
```
stevesloka: Should we generate an "experimental" version of the rendered? I don't think the annotations would matter to someone without the flag, but it's additional settings that someone might not need.

cmd/contour/serve.go
```diff
@@ -122,6 +124,7 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 	serve.Flag("disable-leader-election", "Disable leader election mechanism.").BoolVar(&ctx.DisableLeaderElection)
 
 	serve.Flag("use-extensions-v1beta1-ingress", "Subscribe to the deprecated extensions/v1beta1.Ingress type.").BoolVar(&ctx.UseExtensionsV1beta1Ingress)
+	serve.Flag("experimental-service-apis", "Subscribe to the new service-apis types - CAUTION, EXPERIMENTAL.").BoolVar(&ctx.UseServiceAPITypes)
```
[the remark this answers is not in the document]
youngnick: I think that we use go modules to handle the versioning, and support the current version of service APIs only. The support is going to remain experimental until at least beta, and definitely until we're confident most of the churn is done. Again, the point of this PR is to get the wiring in place to allow us to easily add to it later, not to deliver a complete implementation - which is simply not possible right now.

internal/k8s/converter.go
youngnick: Yes, it should, thanks.

cmd/contour/serve.go
youngnick: If we make the initialisation of the converter check for the flag, then it shouldn't be able to affect anyone else, thanks for catching that in `NewUnstructuredConverter()`.

examples/render/contour.yaml
youngnick: These annotation are just added by the newer version of controller-gen required for the service-apis. They can't be turned off, it's just because we updated the dependency, sadly.

davecheney: That’s kinda obnoxious but c’est la vie

examples/contour/01-crds.yaml
youngnick: In order to actually generate these CRDs, we use controller-gen, specifically:
```sh
go run sigs.k8s.io/controller-tools/cmd/controller-gen \
  crd paths=./apis/... output:dir=$TEMPDIR
```
this is part of `controller-tools`, which had to be updated from 0.2.2 to 0.2.4 as part of this change. This behaviour can't be disabled at the moment, sadly. https://github.com/kubernetes-sigs/controller-tools/pull/338 was merged before 0.2.4 was cut for controller-tools, so it got pulled in for us.

examples/render/contour.yaml
youngnick: It doesn't do anything, it's added by controller-gen only, not needed for the new code.

internal/k8s/converter.go
youngnick: So, I've added this, with a TODO to remove it once the code is no longer experimental. Adding the scheme shouldn't do anything if the experimental flag is not true, as the informers will not be registered anyway. This is a belt-and-suspenders thing for now, just in case.

cmd/contour/serve.go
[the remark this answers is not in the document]
davecheney: s/Subscribe to the new service-apis types - CAUTION, EXPERIMENTAL./Subscribe to experimental new service-apis types/

No need to shout

cmd/contour/serve.go
```diff
@@ -165,6 +168,14 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	registry.MustRegister(prometheus.NewProcessCollector(prometheus.ProcessCollectorOpts{}))
 	registry.MustRegister(prometheus.NewGoCollector())
 
+	// Before we can build out the event handler, we need to initialize the converter we'll
```
davecheney: s/out/our

cmd/contour/serve.go
```diff
@@ -210,37 +221,46 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	// wrap eventRecorder in a converter for objects from the dynamic client.
 	dynamicHandler := &k8s.DynamicClientHandler{
 		Next:      eventRecorder,
-		Converter: k8s.NewUnstructuredConverter(),
+		Converter: converter,
```
davecheney: Why not pass the ctx.UseServiceAPITypes flag into k8s.NewUnstructuredConverter ?

cmd/contour/serve.go
```diff
@@ -210,37 +221,46 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+			dynamicInformers.ForResource(serviceapis.GroupVersion.WithResource("tcproutes")).Informer())
+	}
 
 	// After K8s 1.13 the API server will automatically translate extensions/v1beta1.Ingress objects
 	// to networking/v1beta1.Ingress objects so we should only listen for one type or the other.
 	// The default behavior is to listen for networking/v1beta1.Ingress objects and let the API server
 	// transparently upgrade the extensions version for us.
 	if ctx.UseExtensionsV1beta1Ingress {
-		informers = registerEventHandler(informers, coreInformers.Extensions().V1beta1().Ingresses().Informer(), eventRecorder)
+		informers = registerEventHandler(informers, eventRecorder, coreInformers.Extensions().V1beta1().Ingresses().Informer())
```
davecheney: why are the last two arguments reversed? This smells like the api of registerEventHandler is hard to use

cmd/contour/serve.go
```diff
@@ -405,9 +425,12 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	return g.Run()
 }
 
-func registerEventHandler(informers []cache.SharedIndexInformer, inf cache.SharedIndexInformer, handler cache.ResourceEventHandler) []cache.SharedIndexInformer {
-	inf.AddEventHandler(handler)
-	return append(informers, inf)
+func registerEventHandler(informers []cache.SharedIndexInformer, handler cache.ResourceEventHandler, addinformers ...cache.SharedIndexInformer) []cache.SharedIndexInformer {
```
davecheney: I'm not sure about this, there don't seem to be many callers who are registering more than one informer.

tbh, the signature of this method just went off the deep end, I think it needs to be rethought.

cmd/contour/servecontext.go
```diff
@@ -108,6 +108,14 @@ type serveContext struct {
...
 	// the rewrite itself.
 	UseExtensionsV1beta1Ingress bool `yaml:"-"`
+
+	// Should Contour register to watch the new service-apis types?
+	// By default this value is false, meaning Contour will not do anything with any of the new
+	// types.
+	// If the value is true, Contour will register for all the service-apis types
+	// (GatewayClass, Gateway, HTTPRoute, TCPRoute, and any more as they are added)
+	// Marked EXPERIMENTAL as many places as possible.
+	UseServiceAPITypes bool `yaml:"-"`
```
davecheney: suggestion: if you call the flag `UseExperimentalServiceAPITypes` then you don't need the comments, its right there every time someone types the name of the flag.

internal/dag/cache.go
```diff
@@ -203,6 +208,38 @@ func (kc *KubernetesCache) Insert(obj interface{}) bool {
 		}
 		kc.httpproxydelegations[m] = obj
 		return true
+	case *serviceapis.GatewayClass:
+		m := toMeta(obj)
+		if kc.gatewayclasses == nil {
+			kc.gatewayclasses = make(map[Meta]*serviceapis.GatewayClass)
+		}
+		kc.Infof("Adding GatewayClass %s in namespace %s", m.name, m.namespace)
```
davecheney: are these debug logs?

internal/k8s/converter.go
```diff
@@ -94,7 +95,9 @@ type UnstructuredConverter struct {
 }
 
 // NewUnstructuredConverter returns a new UnstructuredConverter initialized
-func NewUnstructuredConverter() *UnstructuredConverter {
+// TODO(youngnick): Remove the boolean once the service-apis code is not
+// experimental.
+func NewUnstructuredConverter(addServiceApis bool) (*UnstructuredConverter, error) {
```
davecheney: is the check necessary? Why not always convert service api types? This should be safe to do, and one less dimension we need to test.

internal/dag/cache.go
youngnick: Right now, they are the only visible effect that service-apis types are being added and removed. As soon as I can add in some actual behaviour, I plan on removing this - I will add a TODO.

internal/k8s/converter.go
youngnick: After thinking about it for a bit, I agree. I'll remove. The only way that code will ever be hit is if we register the informers in `serve.go` anyway.

cmd/contour/servecontext.go
youngnick: I didn't want to have to rename it all later, but that's a better idea,yeah.

cmd/contour/serve.go
davecheney: further, why not always register the service api types, as long as they aren't hooked up the API via an informer they can't hurt anything.

cmd/contour/serve.go
youngnick: This was a request from @jpeach earlier.

davecheney: oh, my mistake. Is it possible for you to pull this change into another PR so we're not arguing about the color of the drapes.

cmd/contour/serve.go
youngnick: The contract for `AddToScheme` (the kubebuilder way of adding things to a scheme) is different to `AddKnownTypes` and can return an error, we need to be able to bubble that up, sadly. So I had to add an error value to k8s.NewUnstructureConverter.

cmd/contour/serve.go
youngnick: I think I'm going to back this change out from here, and log an issue about it, I think the problem is deeper than the readability having a lot of `informers =` lines.

cmd/contour/serve.go
youngnick: Yeah, I'll back this change out, and log an issue about registerEventHandler.

cmd/contour/serve.go
davecheney: agreed, the solution is probably to break this out into a function of its own

youngnick: Yep, PR incoming for that.

internal/dag/cache.go
```diff
@@ -203,6 +208,46 @@ func (kc *KubernetesCache) Insert(obj interface{}) bool {
...
+		kc.gatewayclasses[m] = obj
+		return true
+	case *serviceapis.Gateway:
+		m := toMeta(obj)
+		if kc.gateways == nil {
+			kc.gateways = make(map[Meta]*serviceapis.Gateway)
+		}
+		// TODO(youngnick): Remove this once service-apis actually have behavior
+		// other than being added to the cache.
+		kc.Infof("Adding Gateway %s in namespace %s", m.name, m.namespace)
```
davecheney: This looks like debugging information, is it needed?

youngnick: If we don't have this in there, there will be no indication whatsoever that the integration is actually working, since there is no coded behavior right now.

youngnick:
There is a test for k8s.NewUnstructuredConverter, at https://github.com/projectcontour/contour/pull/2242/files#diff-54306d2d669a99566d3129e44c64c3b3R45

internal/dag/cache.go
davecheney: Who is the consumer of this information? If it’s one of us, debug sounds like a better level. Also, why are the service api types handled specifically, but not the pre exisiting types?

youngnick: I will change to Debug though.

PR #2244, opened 2020-02-18
internal/contour: record EventHandler operation metrics

### LINKED ISSUE #2235: metrics: contour should report statistics on objects injected, update, deleted via the k8s api
Contour should produce prometheus metrics that track onadd/update/delete metrics for each class of object contour watches.

davecheney:
Fixes #2235 

Introduce a new handler in the EventHandler chain which emits prometheus
metrics for the various API operations and kinds observed from informers
which sink to the EventHandler chain.

Inserting the EventRecorded between the DynamicClientHandler and the
EventHandler will make this PR easier to backport to Contour versions
which don't use the dynamic client.

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -176,7 +176,7 @@ func NewMetrics(registry *prometheus.Registry) *Metrics {
 		EventHandlerOperations: prometheus.NewCounterVec(
 			prometheus.CounterOpts{
 				Name: eventHandlerOperations,
-				Help: "Total number of ResourceEventHandler operations by operation and object kind",
+				Help: "Total number of Kubernetes object changes Contour has received by operation and object kind.",
 			},
 			[]string{"op", "kind"},
 		),
@@ -225,13 +225,7 @@ func (m *Metrics) Zero() {
 	m.SetIngressRouteMetric(zeroes)
 	m.SetHTTPProxyMetric(zeroes)
 
-	ops := []string{"add", "update", "delete"}
-	kinds := []string{"Secret", "Service", "Ingress", "IngressRoute", "HTTPProxy", "TLSCertificateDelegation"}
-	for _, op := range ops {
-		for _, kind := range kinds {
-			m.EventHandlerOperations.WithLabelValues(op, kind).Add(0)
-		}
-	}
+	m.EventHandlerOperations.WithLabelValues("add", "Secret").Inc()
 
 	prometheus.NewTimer(m.CacheHandlerOnUpdateSummary).ObserveDuration()
 }
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -42,7 +42,7 @@ type Metrics struct {
 
 	dagRebuildGauge             *prometheus.GaugeVec
 	CacheHandlerOnUpdateSummary prometheus.Summary
-	EventHandlerOperationGauge  *prometheus.GaugeVec
+	EventHandlerOperations      *prometheus.CounterVec
 
 	// Keep a local cache of metrics for comparison on updates
 	ingressRouteMetricCache *RouteMetric
@@ -78,7 +78,7 @@ const (
 
 	DAGRebuildGauge             = "contour_dagrebuild_timestamp"
 	cacheHandlerOnUpdateSummary = "contour_cachehandler_onupdate_duration_seconds"
-	eventHandlerOperationGauge  = "contour_eventhandler_operation_total"
+	eventHandlerOperations      = "contour_eventhandler_operation_total"
 )
 
 // NewMetrics creates a new set of metrics and registers them with
@@ -173,10 +173,10 @@ func NewMetrics(registry *prometheus.Registry) *Metrics {
 			Help:       "Histogram for the runtime of xDS cache regeneration.",
 			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
 		}),
-		EventHandlerOperationGauge: prometheus.NewGaugeVec(
-			prometheus.GaugeOpts{
-				Name: eventHandlerOperationGauge,
-				Help: "Total number of eventHandler operations received by operation and object kind",
+		EventHandlerOperations: prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Name: eventHandlerOperations,
+				Help: "Total number of ResourceEventHandler operations by operation and object kind",
 			},
 			[]string{"op", "kind"},
 		),
@@ -200,7 +200,7 @@ func (m *Metrics) register(registry *prometheus.Registry) {
 		m.proxyOrphanedGauge,
 		m.dagRebuildGauge,
 		m.CacheHandlerOnUpdateSummary,
-		m.EventHandlerOperationGauge,
+		m.EventHandlerOperations,
 	)
 }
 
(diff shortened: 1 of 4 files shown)


davecheney:
```
% curl -s 127.0.0.1:8000/metrics |  grep eventhandler
# HELP contour_eventhandler_operation_total Total number of eventHandler operations received by operation and object kind
# TYPE contour_eventhandler_operation_total gauge
contour_eventhandler_operation_total{kind="contour.heptio.com/IngressRoutev1beta1",op="onAdd"} 2
contour_eventhandler_operation_total{kind="contour.heptio.com/TLSCertificateDelegationv1beta1",op="onAdd"} 1
contour_eventhandler_operation_total{kind="projectcontour.io/HTTPProxyv1",op="onAdd"} 1
contour_eventhandler_operation_total{kind="unknown",op="onAdd"} 76
```

internal/metrics/metrics.go
```diff
@@ -215,6 +226,10 @@ func (m *Metrics) Zero() {
 	m.SetIngressRouteMetric(zeroes)
 	m.SetHTTPProxyMetric(zeroes)
 
+	m.eventHandlerOperationGauge.WithLabelValues("onAdd", "unknown").Set(0)
+	m.eventHandlerOperationGauge.WithLabelValues("onUpdate", "unknown").Set(0)
+	m.eventHandlerOperationGauge.WithLabelValues("onDelete", "unknown").Set(0)
```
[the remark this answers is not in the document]
jpeach: IIUC, you should also zero the core Contour types here to get the documentation to include those kinds labels.

internal/metrics/metrics.go
```diff
@@ -223,6 +238,46 @@ func (m *Metrics) SetDAGLastRebuilt(ts time.Time) {
...
+// getKind returns the obj's kind, or "unknown" if the object
+// does not have a kind.
+func getKind(obj interface{}) string {
+	type typeMetaAccessor interface {
+		GroupVersionKind() schema.GroupVersionKind
+	}
+	switch obj := obj.(type) {
+	case typeMetaAccessor:
+		gvk := obj.GroupVersionKind()
+		kind := fmt.Sprintf("%s/%s%s", gvk.Group, gvk.Kind, gvk.Version)
```
jpeach: This will print `projectcontour/HTTPProxyv1`, right? Or does `Version` have some separator?. Suggest we format "$group/$version:$kind"?

davecheney: yeah, that looks like crap. What about group/version.kind ?

internal/metrics/metrics.go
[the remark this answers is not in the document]
davecheney: will do on both counts

internal/metrics/metrics.go
```diff
@@ -171,6 +174,13 @@ func NewMetrics(registry *prometheus.Registry) *Metrics {
 			Help:       "Histogram for the runtime of xDS cache regeneration.",
 			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
 		}),
+		eventHandlerOperationGauge: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
```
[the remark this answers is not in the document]
davecheney: I have several thoughts about this whole . I think for this change, following the pattern is better than being inconsistent or making a large stylistic change.

internal/contour/handler.go
```diff
@@ -79,6 +79,7 @@ type opDelete struct {
 }
 
 func (e *EventHandler) OnAdd(obj interface{}) {
+	e.Metrics.RecordOnAdd(obj)
```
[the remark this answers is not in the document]
davecheney: yup, I can do that.

internal/metrics/metrics.go
jpeach: Yeh that seems OK too. Not sure if there is a common convention here, but we will have to live with the choice until we retire this metric. Note that for core types `gvk.Group` will be empty.

internal/metrics/metrics.go
```diff
@@ -223,6 +238,46 @@ func (m *Metrics) SetDAGLastRebuilt(ts time.Time) {
...
+	m.recordOp("onDelete", obj)
+}
+
+func (m *Metrics) recordOp(op string, obj interface{}) {
+	m.eventHandlerOperationGauge.WithLabelValues(op, getKind(obj)).Inc()
+}
+
+// getKind returns the obj's kind, or "unknown" if the object
+// does not have a kind.
+func getKind(obj interface{}) string {
```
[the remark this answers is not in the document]
davecheney: I've always been super frustrated that getting the group and kind of an object is hard in k8s. In the past I've worked around this with type assertions, but in this case I don't want to limit this to type assertions.

internal/metrics/metrics.go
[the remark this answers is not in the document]
jpeach: Gauge -> counter is a semantic change. By specifying a gauge, we are telling people not to plot these as rates. But they definitely have Counter semantics.

internal/metrics/metrics.go
davecheney: I really f'ing hate the k8s client. Its inconsistencies are legion

site/_metrics/contour_eventhandler_operation_total.md
```diff
@@ -0,0 +1,7 @@
+---
+name: 'contour_eventhandler_operation_total'
+type: '[COUNTER](https://prometheus.io/docs/concepts/metric_types/#counter)'
+labels: 'kind, op, kind, op, kind, op, kind, op, kind, op, kind, op, kind, op, kind, op, kind, op, kind, op, kind, op, kind, op, kind, op, kind, op, kind, op, kind, op, kind, op, kind, op'
```
davecheney: bugger

internal/metrics/metrics.go
```diff
@@ -215,6 +225,14 @@ func (m *Metrics) Zero() {
 	m.SetIngressRouteMetric(zeroes)
 	m.SetHTTPProxyMetric(zeroes)
 
+	ops := []string{"add", "update", "delete"}
+	kinds := []string{"Secret", "Service", "Ingress", "IngressRoute", "HTTPProxy", "TLSCertificateDelegation"}
+	for _, op := range ops {
+		for _, kind := range kinds {
+			m.EventHandlerOperations.WithLabelValues(op, kind).Inc()
+		}
+	}
```
jpeach: To fix the crazy docs, you can replace all this with:
```Go
m.EventHandlerOperations.WithLabelValues("op", "kind").Inc()
```

davecheney:
ok, this is ready for another pass.

internal/metrics/metrics.go
```diff
@@ -171,6 +173,13 @@ func NewMetrics(registry *prometheus.Registry) *Metrics {
 			Help:       "Histogram for the runtime of xDS cache regeneration.",
 			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
 		}),
+		EventHandlerOperations: prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Name: eventHandlerOperations,
+				Help: "Total number of ResourceEventHandler operations by operation and object kind",
```
jpeach: Not sure that operators will know what to make of "ResourceEventHandler operations". Maybe

```
Total number of Kubernetes object changes Contour has received by operation and object kind.
```

youngnick: That's really good.

internal/metrics/metrics.go
jpeach: Actually, best to use a real label in in case this ever gets called for real :)

internal/metrics/metrics.go
davecheney: done

PR #2245, opened 2020-02-18
internal/k8s: move dynamic client translation to k8s package

davecheney:
Updates #2235

Contour 1.2.x is moving to the k8s dynamic client for deserialising
Contour's CRD objects. To make #2244 easier to backport to Contour's 1.1
branch, which does not support the dynamic client, break this logic out
into its own ResourceEventHandler wrapper which returns the current
contour.EventHandler logic to Contour 1.1 spec.

As a side effect, we only need to use the DynamicClientHandler wrapper
in cases where we know the resources are coming from the DynamicClient.
Specifically we don't need to use this path for Core.v1 and Extension
objects. Also, we don't need to use the DynamicClientHandler wrapper in
e2e/feature tests as there is no k8s API server connected to those tests
-- we construct the final k8s object as a fixture and call
contour.EventHandler.OnAdd/Update/Delete directly.

### DIFF
--- a/internal/k8s/converter.go
+++ b/internal/k8s/converter.go
@@ -15,13 +15,72 @@ package k8s
 
 import (
 	"fmt"
-	"reflect"
 
 	ingressroutev1 "github.com/projectcontour/contour/apis/contour/v1beta1"
 	projectcontour "github.com/projectcontour/contour/apis/projectcontour/v1"
+	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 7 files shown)


internal/k8s/converter.go
```diff
@@ -15,14 +15,73 @@ package k8s
...
+
+	// Next is the next handler in the chain.
+	Next cache.ResourceEventHandler
+
+	// Converter is the registered converter.
+	Converter Converter
+
+	Logger logrus.FieldLogger
+}
+
```
jpeach: Nit, add local type enforcement:

```Go
var _ cache.ResourceEventHandler = & DynamicClientHandler{}
```

davecheney: done

Issue #2223, opened 2020-02-12
Test minimal Envoy functionality before release


youngnick:
This issue covers work to test the minimum Envoy functionality in #2048 with a release candidate before the final release is cut.

This step should be performed between creating the release branch and cutting the final release in the release runbook https://projectcontour.io/resources/release-process/.

jpeach:
xref #2048

jpeach:
Probably need to spin off a separate issue for generating the deployment YAML to run envoy with extensions disabled. That might depend on doing something with kustomize, since we want to be able to deploy Contour in a standard way but just slam in some new flags to the envoy container.

xref #2088

stevesloka:
Completed manual testing for Contour release v1.2.0

jpeach:
Test scenarios from @stevesloka 

https://docs.google.com/spreadsheets/d/1Y22qy_9Q6UO1rcM24NrFq8x6sHoEpWTuNEVQq7hb42Y

stevesloka:
@jpeach what's the status on the test infra? Can we merge that and start incorporating even if it's not 100% complete. 

We also need to find a way to track changes as new features are added. For example, the ingress class work that was just added should be tracked somewhere to be added to the functional tests.

jpeach:
> @jpeach what's the status on the test infra?

The test driver works and can run declarative tests. I've got a handful of tests written for Contour.

> Can we merge that and start incorporating even if it's not 100% complete.

Yes, we should start that. I need to publish the repo and add goreleaser so that there is an artifact that we can run. I have a separate repo with the test cases, which I think would end up being merged into Contour repository.

> We also need to find a way to track changes as new features are added. For example, the ingress class work that was just added should be tracked somewhere to be added to the functional tests.

We could add a label, or file an issue for each test as we think of it?

youngnick:
I think that creating issues with a label is the way to go

skriss:
What's the action item on this issue? Should we consider this closed now that we have some integration tests in place?

youngnick:
Yes, I think that we should close this out now that we have integration tests in place. More specific items should generate new issues.

PR #2249, opened 2020-02-19
ci: verify that Contour installs

jpeach:
Add the bare minimum CI machinery to fetch a Kubernetes toolchain
and verify that Contour can be installed in the example deployment
configuration.

This updates #2223.


.travis.yml
```diff
@@ -24,6 +24,26 @@ jobs:
...
 
+  # Build the Contour container image with the local Docker. Create a kind
+  # cluster, push the image we built into it, and verify that both Contour
+  # and Envoy come up. This is an initial placeholder as we experiment with
+  # different approaches to integration testing.
+  - name: integration
+    install:
+    - mkdir -p $HOME/bin
+    - ./hack/travis/install-kubernetes-toolchain.sh
+    - make container IMAGE=docker.io/projectcontour/contour VERSION=ci
```
stevesloka: Maybe make this obvious that it's not a real container? Like contour:ci?

hack/travis/install-kubernetes-toolchain.sh
```diff
@@ -0,0 +1,44 @@
+#! /usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+readonly DESTDIR=$(cd "${DESTDIR:-${HOME}/bin}" && pwd)
```
stevesloka: `DESTDIR` is confusing to me. Maybe call it what it is like `BINDIR` or something?

.travis.yml
```diff
@@ -24,6 +24,26 @@ jobs:
...
+    - docker tag docker.io/projectcontour/contour:ci docker.io/projectcontour/contour:master
+    - docker tag docker.io/projectcontour/contour:ci docker.io/projectcontour/contour:latest
+    script:
+    - $HOME/bin/kind create cluster
+    - $HOME/bin/kind load docker-image docker.io/projectcontour/contour:master
+    - $HOME/bin/kind load docker-image docker.io/projectcontour/contour:latest
+    - $HOME/bin/kubectl apply -f examples/render/contour.yaml
```
stevesloka: You need to wait for the kind cluster to be ready before this step. There's a wait flag now in kind to do this (` --wait duration`)

hack/travis/install-kubernetes-toolchain.sh
jpeach: It's by analogy to [DESTDIR](https://www.gnu.org/prep/standards/html_node/DESTDIR.html), which was in common usage in  olden times :)

.travis.yml
jpeach: The fully qualified container name this generates is `docker.io/projectcontour/contour:ci`. Is that similar to what you were thinking of?

jpeach:
> To me, these tasks should all be runnable locally without pushing to Travis. I'd expect a way to do: make test-integration or something and spin up a kind cluster, build, test, etc all from my local machine.

For dev, there's different configurations and workflows. This PR is really a minimal baseline that we can use to do something with in CI. I'm loth to say the devs should run it, because it pollutes the build environment and is likely to be very different from anything dev will need. As in the comment, this is intended to be a step 0, but is by no means where I want to end up.

.travis.yml
stevesloka: No my thought is to make it not look like a real image. I realize the CI system won't have any creds to push images, but by using image names that couldn't match a real image, there's no risk in pulling an image by mistake because it matches an existing one.

hack/travis/install-kubernetes-toolchain.sh
stevesloka: No, I'm familiar. This is just an opinion on style.

stevesloka:
>  I'm loth to say the devs should run it, because it pollutes the build environment 

Right like I said previously I think we see differently on these things. I personally would like to be able to run everything locally and skip the commit + merge + push + wait + check + repeat cycle.  This is fine for CI things like you mentioned but to enable a local build this should be redesigned which is why I'm bringing up now.

.travis.yml
stevesloka: e.g. `make container IMAGE=contour`

hack/travis/install-kubernetes-toolchain.sh
jpeach: I can change that, but I still need to tag with the official names so I can push them into the Kind cluster.

.travis.yml
jpeach: I can change that, but I still need to tag with the official names so I can push them into the Kind cluster.

jpeach:
> This is fine for CI things like you mentioned but to enable a local build this should be redesigned which is why I'm bringing up now.

Yes, I agree. I don't expect this code to last, it's just a stopgap that does *something* on CI. It's barely testing anything :)

Maybe the next step is to add a script to deploy some proxies, and maybe the step after that we throw this away. That's fine, it's low effort and not permanent. If you would prefer, I can stuff the steps into a script now, but I didn't think that is does enough to make it worthwhile.

.travis.yml
stevesloka: Why do you need official names? Just load the image and use from the k8s objects. Shouldn't need to push to a real registry.

jpeach: I'm not modifying the example YAML, so I push the images into Kind to match the names that it uses. When we can move all that to kustomize, we can generate YAML with rewritten image names.

hack/travis/install-kubernetes-toolchain.sh
jpeach: I changed the script to take the target directory as an argument.

.travis.yml
jpeach: Added.

.travis.yml
stevesloka: Ok then. Maybe this just needs a design doc like @youngnick started previously to understand the work. This is fine, ship it.😀

PR #2261, opened 2020-02-20
Add Contour APIserver stats Grafana dashboard

### LINKED ISSUE #2248: examples: Add EventHandler metrics Grafana panel
Follow up to this metrics PR (https://github.com/projectcontour/contour/pull/2244), we should add those metrics to the example Grafana dashboard to help users visualize these metrics if they want to do that quickly with our example dashboards.

youngnick:
Fixes #2248

apiserver-operations-dashboard


youngnick:
Sadly, it's unavoidable because of the bucketing of the stats. The stat is stored as a counter, and the irate function takes samples over 1 minute to be able to get the rate. Because of that, it can't show the actual value. Maybe I should just hide the numbers...

davecheney:
> My only question is on the Y-axis for the ops/s, should the units be fractions?

yeah, I commented on this myself but its a Grafana limitation, which is annoying, but not something we can fix as a user.

PR #2276, opened 2020-02-24
cmd/contour: change registerEventHandler to new InformerSyncList type

youngnick:
Turns out the `informers` array was being added only to keep
track of whether the informer had synced yet. This PR adds
a new InformerSyncList type which stores the functions from each informer
and calls them for that first-sync check.

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -31,8 +31,6 @@ import (
 
 	"k8s.io/client-go/dynamic"
 
-	"k8s.io/client-go/tools/cache"
-
 	"github.com/projectcontour/contour/internal/contour"
 	"github.com/projectcontour/contour/internal/dag"
 	"github.com/projectcontour/contour/internal/debug"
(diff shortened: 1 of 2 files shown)


cmd/contour/serve.go
```diff
@@ -215,32 +215,35 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
-	var informers []cache.SharedIndexInformer
 
-	informers = registerEventHandler(informers, dynamicInformers.ForResource(ingressroutev1.IngressRouteGVR).Informer(), dynamicHandler)
-	informers = registerEventHandler(informers, dynamicInformers.ForResource(ingressroutev1.TLSCertificateDelegationGVR).Informer(), dynamicHandler)
-	informers = registerEventHandler(informers, dynamicInformers.ForResource(projectcontour.HTTPProxyGVR).Informer(), dynamicHandler)
-	informers = registerEventHandler(informers, dynamicInformers.ForResource(projectcontour.TLSCertificateDelegationGVR).Informer(), dynamicHandler)
-	informers = registerEventHandler(informers, coreInformers.Core().V1().Services().Informer(), eventRecorder)
+	// synclist holds a list of informer HasSynced() functions, which are used later
+	// to wait for all the informers to finish syncing for the first time.
+	synclist := &SyncList{}
```
davecheney: var syncList SyncList

davecheney: I don't think this comment really adds much, maybe called the variable `informerSyncedList` would remove the need for the comment.

cmd/contour/serve.go
```diff
@@ -346,14 +349,10 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
 
-		synced := make([]cache.InformerSynced, 0, len(informers))
-		for _, inf := range informers {
-			synced = append(synced, inf.HasSynced)
-		}
-
 		log.Printf("waiting for informer caches to sync")
-		if !cache.WaitForCacheSync(stop, synced...) {
-			return fmt.Errorf("error waiting for cache to sync")
+		err := synclist.WaitForSync(stop)
```
davecheney: might be easier to read if you use the 

if err := ... ; err != nil {
return err
}

form.

Also, check(err) should be ok to use here.

cmd/contour/syncer.go
```diff
@@ -0,0 +1,26 @@
...
+	"fmt"
+
+	"k8s.io/client-go/tools/cache"
+)
+
+type Syncer interface {
+	HasSynced() bool
+}
+
+type SyncList struct {
```
davecheney: I'm thinking that this type should be named for what it contains, not what it does. Perhaps `InformerList` or just `Infomers` or something might be better. Then you get

func (il *InformerList) Add(I Informer)

cmd/contour/syncer.go
```diff
@@ -0,0 +1,26 @@
+package main
```
davecheney: please add copywrite boilerplate

cmd/contour/syncer.go
davecheney: or maybe not, I think stand alone in this type's name is ok, but when its used in context above I find it lacking.

cmd/contour/serve.go
```diff
@@ -405,9 +404,11 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	return g.Run()
 }
 
-func registerEventHandler(informers []cache.SharedIndexInformer, inf cache.SharedIndexInformer, handler cache.ResourceEventHandler) []cache.SharedIndexInformer {
+// registerInformer is a helper to cut boilerplate from registering informers.
+// We need to add the event handler and save the informers HasSynced function for later.
+func registerInformer(inf cache.SharedIndexInformer, handler cache.ResourceEventHandler, sl *SyncList) {
```
davecheney: Hmm, I'm not sure about this function signature, now the append has been factored out I'm even less sure this function makes sense. eg.

It's called `RegisterInformer` but neither of its two lines talk about registering anything. The first add's an event handler ... not an informer. The second adds an informer to something called a SyncList. At least the previous had the, sort of verbose, `informers []cache.SharedIndexInformer` parameter.

What would happen if SyncList.Add() returned the informer it added? then you could have something like list

var informerList SyncList

informerList.Add(coreInformers.Extensions().V1beta1().Ingresses().Informer()).AddEventHandler(handler)

cmd/contour/serve.go
youngnick: I used `return err` because the rest of the function did, seems weird to use check inside an inner function to me.

cmd/contour/serve.go
youngnick: That's much better, changed.

davecheney: but it doesn't register an informer, it registers a handler.

cmd/contour/syncer.go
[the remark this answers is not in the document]
youngnick: that's fair, I'll do that.

PR #2283, opened 2020-02-24
design: Update rate limit design for HTTPProxy

stevesloka:
Updates #873 for HTTPProxy instead of IngressRoute. 

Also, moves the design from args to Contour to the configuration file. 


stevesloka:
@youngnick I think I've addressed all the comments around support and config defaults.

iyacontrol:
when the rate limiting of contour is released?

stevesloka:
Hey @iyacontrol & @bill-within, I'm going to update this design one more time then once we all agree should be able to code it up quickly (the work was done already). I'm going to see if we can get it ready for the Contour 1.5 release, but need to chat with the team on that priority.

iyacontrol:
> Hey @iyacontrol & @bill-within, I'm going to update this design one more time then once we all agree should be able to code it up quickly (the work was done already). I'm going to see if we can get it ready for the Contour 1.5 release, but need to chat with the team on that priority.

woops! Thank you very much for your reply. We urgently need the rate limiting. Thanks again for your work.

stevesloka:
The ideas around this have changed since merging the auth proposal, closing to rewrite against the new ideas.

PR #2286, opened 2020-02-25
Add simple debug logging flag

youngnick:
I don't believe that we want to add any more levels than error, info, and debug at this point.

### DIFF
--- a/cmd/contour/contour.go
+++ b/cmd/contour/contour.go
@@ -94,6 +94,9 @@ func main() {
 		// on top of any values sourced from -c's config file.
 		_, err := app.Parse(args)
 		check(err)
+		if serveCtx.Debug {
+			log.SetLevel(logrus.DebugLevel)
+		}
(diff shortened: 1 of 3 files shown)


davecheney:
I'd rather not see it added to the config file, that implies that people are running all their clusters in debug mode which mean's we haven't put enough effort into making the info level log statements useful.

jpeach:
> I'd rather not see it added to the config file, that implies that people are running all their clusters in debug mode which mean's we haven't put enough effort into making the info level log statements useful.

Since it's not enabled by default, why does it imply that? It's way harder to enable when you have to rewrite command line flags than editing the config.

davecheney:
> It's way harder to enable when you have to rewrite command line flags than editing the config.

I don't agree with this statement, the latter requires me to edit a text file, the former I just type when I want to use it.

youngnick:
I see where you are coming from @jpeach, but for now I think this is mainly for local use.

PR #2292, opened 2020-02-25
makefile: Add a 'make checkall' target

### LINKED ISSUE #2291: makefile: Run all tests in single command from makefile
Contour had a command that was `make check` and it was useful to run locally on your machine to check for failing tests, spelling errors, or linting errors.

A recent effort to optimize CI builds split apart the make targets into CI build steps which is a fantastic improvement to parallelize the build. 

However, I would still like the ability to run most of the tests locally without having to wait for CI to run them for me.

stevesloka:
Fixes #2291 

This PR adds a `make checkall` task which runs the tests, install, linting and generate tasks. It's using the aliases such that as these commands grow, the meaning behind the commands doesn't have to be translated. 

Additionally, it renames the `check-` tasks under the `lint` task to be prefixed with `lint-*` to keep consistency. 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,9 @@ export GO111MODULE=on
 .PHONY: check
 check: install check-test check-test-race ## Install and run tests
 
+.PHONY: checkall
+checkall: check lint check-generate
+
 install: ## Build and install the contour binary
 	go install -mod=readonly -v $(GO_TAGS) $(MODULE)/cmd/contour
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -151,6 +154,10 @@ generate-metrics-docs:
 	@echo Generating metrics documentation ...
 	@cd site/_metrics && rm -f *.md && go run ../../hack/generate-metrics-doc.go
 
+.PHONY: generate-check
```
jpeach: Nit: call this `check-generate` for consistency.

Makefile
```diff
@@ -151,6 +154,10 @@ generate-metrics-docs:
...
+generate-check:
+	@./hack/travis/check-uncommitted-codegen.sh
+
```
jpeach: I think this should be:
```
generate-check: generate
 	@./hack/travis/check-uncommitted-codegen.sh
```

Makefile
```diff
@@ -35,6 +35,9 @@ export GO111MODULE=on
 .PHONY: check
 check: install check-test check-test-race ## Install and run tests
 
+.PHONY: checkall
+checkall: check lint generate generate-check
```
jpeach: Are you sure you don't want to just bundle this up under `check`?

PR #2250, opened 2020-02-19
internal: Initial support for external client cert validation

tsaarni:
internal: Initial support for external client cert validation

Adds initial support for authentication of external clients (downstream) by
validating their client certificates against trusted CA certificate.

The feature is not yet exposed in the API.

Signed-off-by: Tero Saarni <tero.saarni@est.tech>

PR #2300, opened 2020-03-01
internal/dag: clean up TLS validation errors

jpeach:
When looking up and validating TLS peer validation policy, construct
the error details in the caller. This allows better error context and
later will let us re-use this code.


jpeach:
This will help error handling in #2250.

PR #2301, opened 2020-03-02
internal/dag: move extension ingress translation to a helper

davecheney:
Updates #403

Prior to this PR the automagic translation of the deprecated
extensions/ingress.v1beta1 types to networking/ingress.v1beta1 types was
handled during dag.KubernetesCache insertion.

This PR moves this code to internal/k8s as a helper, a la the
DynamicConverter. This removes the requirement to care about those
deprecated API types from things that process them, including the cache
and ingress status writer.


davecheney:
Closed by #2302

PR #2327, opened 2020-03-09
internal/dag,envoy: use constant for ca.crt

tsaarni:
Cleaning up locally defined strings when referencing to secrets with
CA certificates and replaced them with a single constant.

### DIFF
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -1345,7 +1345,7 @@ func validSecret(s *v1.Secret) bool {
 }
 
 func validCA(s *v1.Secret) bool {
-	return len(s.Data["ca.crt"]) > 0
+	return len(s.Data[CACertificateKey]) > 0
 }
 
 // routeEnforceTLS determines if the route should redirect the user to a secure TLS listener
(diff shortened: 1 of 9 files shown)


tsaarni:
This clean up was proposed by @jpeach in review comment https://github.com/projectcontour/contour/pull/2250#discussion_r386154091

internal/dag/dag.go
```diff
@@ -191,6 +191,9 @@ type UpstreamValidation struct {
 	SubjectName string
 }
 
+// CACertificateKey stores the key name for the TLS validation secret cert.
+const CACertificateKey = "ca.crt"
```
jpeach: Let's put this in `internal/dag/secret.go`, where it's closer to the subject matter.

internal/dag/dag.go
```diff
@@ -191,6 +191,9 @@ type UpstreamValidation struct {
 	SubjectName string
 }
 
+// CACertificateKey stores the key name for the TLS validation secret cert.
```
jpeach: suggest:
```
CaCertificateKey is the key name for accessing TLS CA certificate bundles in Kubernetes Secrets.
```

internal/dag/dag.go
tsaarni: Now moved.

internal/dag/dag.go
tsaarni: Fixed.

PR #2329, opened 2020-03-10
site: fix API docs rendering

jpeach:
Remove extraneous slashes from `<br>` tags so that Jekyll
doesn't end up quoting them.


jpeach:
> is a constant point of pain in this repo. Why do we need the 
here at all. Can it just be deleted?

Not sure what you are trying to say here, can you please elaborate?

Issue #2149, opened 2020-01-24
Contour should document how to use alternatives to certgen for securing traffic between contour and envoy


davecheney:
Notwithstanding @tsaarni 's efforts to make certificate rotation easier to accomplish, we should acknowledge that certgen is not going to grow beyond its current role of supporting the example deployments of Contour. To that end we should investigate and document the procedure for using alternative solutions like cert-manager or vault to maintain the certificates used by contour and envoy.

jpeach:
Ref #2143

ryanelian:
Big yes to using cert-manager replacing cert-gen.

Got hit by a deployment issue with contour-certgen job: https://github.com/projectcontour/contour/issues/2150

tsaarni:
I did a small proof of concept where I used [cert-manager](https://cert-manager.io/) instead of certgen for issuing xDS certificates.  At this point running this requires a fork of  Contour with #2333 and Envoy with https://github.com/envoyproxy/envoy/pull/10163.  

I did changes in deployment manifest by editing the final rendered version of the example manifest [contour.yaml](https://github.com/projectcontour/contour/blob/6eca7b4ccd542e09d1034cc23be49297c15a2d39/examples/render/contour.yaml) on-the-fly with a kustomize patch.  I did not create PR yet since I'm bit doubtful **how to best maintain** the manifests for these two deployment alternatives (certgen and cert-manager)?

Here is link to the PoC https://github.com/projectcontour/contour/compare/master...Nordix:cert-manager 

Summary of changes in the manifest that I think are necessary:

* removed ServiceAccount for certgen
* removed Role for certgen
* removed RoleBinding between the certgen service account and role
* removed certgen Job
* removed cacert volume mount from Envoy DaemonSet (both init container and main) since  `cacert` secret does not exist anymore. Cert-manager writes CA cert in same secret with TLS cert and key.
* changed `contour bootstrap` argument from `--envoy-cafile=/ca/cacert.pem` to `--envoy-cafile=/certs/ca.crt` in Envoy DaemonSet, since cert-manager names the CA certificate as `ca.crt`
* removed cacert volume mount from Contour Deployment
* changed `contour serve` argument from `--contour-cafile=/ca/cacert.pem` to `--contour-cafile=/certs/ca.crt `in Contour Deployment

jpeach:
xref #2088 

[...]

PR #2333, opened 2020-03-11
bootstrap: support for Envoy xDS certificate rotation

tsaarni:
bootstrap: support for Envoy xDS certificate rotation

Added new flag --resources-dir to bootstrap command for supporting path-based
SDS resources to define xDS certificate and key.  When using the flag,
following changes take place:

- Bootstrap file will contain references to SDS resource files instead of direct cert/key paths.
- SDS resource file is written into resources dir for xDS client cert and key
- SDS resource file is written into resources dir for xDS trusted CA cert

With this change there is no need to restart Envoy anymore when certificates
are rotated.  Envoy will monitor and automatically reload the certificate and
key files without causing an interruption to traffic.

Any recent version of Envoy supports the above configuration, but Envoy 1.14.1
or later is required for automatic certificate reload.

### DIFF
--- a/cmd/contour/bootstrap.go
+++ b/cmd/contour/bootstrap.go
@@ -14,56 +14,25 @@
 package main
 
 import (
-	"io"
-	"os"
-
-	"github.com/golang/protobuf/jsonpb"
 	"github.com/projectcontour/contour/internal/envoy"
 	kingpin "gopkg.in/alecthomas/kingpin.v2"
 )
 
 // registerBootstrap registers the bootstrap subcommand and flags
 // with the Application provided.
-func registerBootstrap(app *kingpin.Application) (*kingpin.CmdClause, *bootstrapContext) {
-	var ctx bootstrapContext
+func registerBootstrap(app *kingpin.Application) (*kingpin.CmdClause, *envoy.BootstrapConfig) {
+	var config envoy.BootstrapConfig
 
 	bootstrap := app.Command("bootstrap", "Generate bootstrap configuration.")
-	bootstrap.Arg("path", "Configuration file ('-' for standard output).").Required().StringVar(&ctx.path)
-	bootstrap.Flag("admin-address", "Envoy admin interface address.").StringVar(&ctx.config.AdminAddress)
-	bootstrap.Flag("admin-port", "Envoy admin interface port.").IntVar(&ctx.config.AdminPort)
-	bootstrap.Flag("xds-address", "xDS gRPC API address.").StringVar(&ctx.config.XDSAddress)
-	bootstrap.Flag("xds-port", "xDS gRPC API port.").IntVar(&ctx.config.XDSGRPCPort)
-	bootstrap.Flag("envoy-cafile", "gRPC CA Filename for Envoy to load.").Envar("ENVOY_CAFILE").StringVar(&ctx.config.GrpcCABundle)
-	bootstrap.Flag("envoy-cert-file", "gRPC Client cert filename for Envoy to load.").Envar("ENVOY_CERT_FILE").StringVar(&ctx.config.GrpcClientCert)
-	bootstrap.Flag("envoy-key-file", "gRPC Client key filename for Envoy to load.").Envar("ENVOY_KEY_FILE").StringVar(&ctx.config.GrpcClientKey)
-	bootstrap.Flag("namespace", "The namespace the Envoy container will run in.").Envar("CONTOUR_NAMESPACE").Default("projectcontour").StringVar(&ctx.config.Namespace)
-	return bootstrap, &ctx
-}
-
-type bootstrapContext struct {
-	config envoy.BootstrapConfig
-	path   string
-}
-
-// doBootstrap writes an Envoy bootstrap configuration file to the supplied path.
-func doBootstrap(ctx *bootstrapContext) {
-	var out io.Writer
-
-	switch ctx.path {
-	case "-":
-		out = os.Stdout
-	default:
-		f, err := os.Create(ctx.path)
-		check(err)
-
-		out = f
-
-		defer func() {
-			check(f.Close())
-		}()
-	}
-
-	m := &jsonpb.Marshaler{OrigName: true}
-
-	check(m.Marshal(out, envoy.Bootstrap(&ctx.config)))
+	bootstrap.Arg("path", "Configuration file ('-' for standard output).").Required().StringVar(&config.Path)
+	bootstrap.Flag("resources-dir", "Directory where configuration files will be written to.").StringVar(&config.ResourcesDir)
+	bootstrap.Flag("admin-address", "Envoy admin interface address.").StringVar(&config.AdminAddress)
+	bootstrap.Flag("admin-port", "Envoy admin interface port.").IntVar(&config.AdminPort)
+	bootstrap.Flag("xds-address", "xDS gRPC API address.").StringVar(&config.XDSAddress)
+	bootstrap.Flag("xds-port", "xDS gRPC API port.").IntVar(&config.XDSGRPCPort)
+	bootstrap.Flag("envoy-cafile", "gRPC CA Filename for Envoy to load.").Envar("ENVOY_CAFILE").StringVar(&config.GrpcCABundle)
+	bootstrap.Flag("envoy-cert-file", "gRPC Client cert filename for Envoy to load.").Envar("ENVOY_CERT_FILE").StringVar(&config.GrpcClientCert)
+	bootstrap.Flag("envoy-key-file", "gRPC Client key filename for Envoy to load.").Envar("ENVOY_KEY_FILE").StringVar(&config.GrpcClientKey)
+	bootstrap.Flag("namespace", "The namespace the Envoy container will run in.").Envar("CONTOUR_NAMESPACE").Default("projectcontour").StringVar(&config.Namespace)
+	return bootstrap, &config
 }
--- a/internal/envoy/bootstrap.go
+++ b/internal/envoy/bootstrap.go
@@ -16,7 +16,9 @@
 package envoy
 
 import (
-	"log"
+	"fmt"
+	"os"
+	"path"
 	"strconv"
 	"strings"
 	"time"
@@ -25,19 +27,126 @@ import (
 	envoy_api_v2_auth "github.com/envoyproxy/go-control-plane/envoy/api/v2/auth"
 	clusterv2 "github.com/envoyproxy/go-control-plane/envoy/api/v2/cluster"
 	envoy_api_v2_core "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
-	bootstrap "github.com/envoyproxy/go-control-plane/envoy/config/bootstrap/v2"
+	envoy_api_bootstrap "github.com/envoyproxy/go-control-plane/envoy/config/bootstrap/v2"
 	matcher "github.com/envoyproxy/go-control-plane/envoy/type/matcher"
+	"github.com/golang/protobuf/jsonpb"
+	"github.com/golang/protobuf/proto"
+	"github.com/golang/protobuf/ptypes/any"
 	"github.com/projectcontour/contour/internal/protobuf"
 )
 
-// Bootstrap creates a new v2 Bootstrap configuration.
-func Bootstrap(c *BootstrapConfig) *bootstrap.Bootstrap {
-	b := &bootstrap.Bootstrap{
-		DynamicResources: &bootstrap.Bootstrap_DynamicResources{
+// sdsResourcesSubdirectory stores the subdirectory name where SDS path resources are stored to.
+const sdsResourcesSubdirectory = "sds"
+
+// sdsTLSCertificateFile stores the path to the SDS resource with Envoy's
+// client certificate and key for XDS gRPC connection.
+const sdsTLSCertificateFile = "xds-tls-certicate.json"
+
+// sdsValidationContextFile stores the path to the SDS resource with
+// CA certificates for Envoy to use for the XDS gRPC connection.
+const sdsValidationContextFile = "xds-validation-context.json"
+
+// ConfigFiles is a map from configuration filename to configuration file content.
+type ConfigFiles map[string]proto.Message
+
+// WriteBootstrap writes bootstrap configuration to files.
+func WriteBootstrap(c *BootstrapConfig) error {
+	// Create Envoy bootstrap config and associated resource files.
+	steps, err := bootstrap(c)
+	if err != nil {
+		return err
+	}
+
+	if c.ResourcesDir != "" {
+		if err := os.MkdirAll(path.Join(c.ResourcesDir, "sds"), 0755); err != nil {
+			return err
+		}
+	}
+
+	// Write all configuration files out to filesystem.
+	for _, step := range steps {
+		if err := writeConfig(step(c)); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+type bootstrapf func(*BootstrapConfig) (string, proto.Message)
+
+// bootstrap creates a new v2 bootstrap configuration and associated resource files.
+func bootstrap(c *BootstrapConfig) ([]bootstrapf, error) {
+	steps := []bootstrapf{}
+
+	if c.GrpcClientCert == "" && c.GrpcClientKey == "" && c.GrpcCABundle == "" {
+		steps = append(steps,
+			func(*BootstrapConfig) (string, proto.Message) {
+				return c.Path, bootstrapConfig(c)
+			})
+
+		return steps, nil
+	}
+
+	// If one of the two TLS options is not empty, they all must be not empty.
+	if c.GrpcClientCert == "" || c.GrpcClientKey == "" || c.GrpcCABundle == "" {
+		return nil, fmt.Errorf("you must supply all TLS parameters - %q, %q, %q, or none of them",
+			"--envoy-cafile", "--envoy-cert-file", "--envoy-key-file")
+	}
+
+	if c.ResourcesDir == "" {
+		// For backwards compatibility, the old behavior
+		// is to use direct certificate and key file paths in
+		// bootstrap config. Envoy does not support rotation
+		// of xDS certificate files in this case.
+
+		steps = append(steps,
+			func(*BootstrapConfig) (string, proto.Message) {
+				b := bootstrapConfig(c)
+				b.StaticResources.Clusters[0].TransportSocket = UpstreamTLSTransportSocket(
+					upstreamFileTLSContext(c))
+				return c.Path, b
+			})
+
+		return steps, nil
+	}
+
+	// xDS certificate rotation is supported by Envoy by using SDS path based resource files.
+	// These files are JSON representation of the SDS protobuf messages that normally get sent over the xDS connection,
+	// but for xDS connection itself, bootstrapping is done by storing the SDS resources in a local filesystem.
+	// Envoy will monitor and reload the resource files and the certificate and key files referred from the SDS resources.
+	//
+	// Two files are written to ResourcesDir:
+	// - SDS resource for xDS client certificate and key for authenticating Envoy towards Contour.
+	// - SDS resource for trusted CA certificate for validating Contour server certificate.
+	sdsTLSCertificatePath := path.Join(c.ResourcesDir, sdsResourcesSubdirectory, sdsTLSCertificateFile)
+	sdsValidationContextPath := path.Join(c.ResourcesDir, sdsResourcesSubdirectory, sdsValidationContextFile)
+
+	steps = append(steps,
+		func(*BootstrapConfig) (string, proto.Message) {
+			return sdsTLSCertificatePath, tlsCertificateSdsSecretConfig(c)
+		},
+		func(*BootstrapConfig) (string, proto.Message) {
+			return sdsValidationContextPath, validationContextSdsSecretConfig(c)
+		},
+		func(*BootstrapConfig) (string, proto.Message) {
+			b := bootstrapConfig(c)
+			b.StaticResources.Clusters[0].TransportSocket = UpstreamTLSTransportSocket(
+				upstreamSdsTLSContext(sdsTLSCertificatePath, sdsValidationContextPath))
+			return c.Path, b
+		},
+	)
+
+	return steps, nil
+}
+
+func bootstrapConfig(c *BootstrapConfig) *envoy_api_bootstrap.Bootstrap {
+	return &envoy_api_bootstrap.Bootstrap{
+		DynamicResources: &envoy_api_bootstrap.Bootstrap_DynamicResources{
 			LdsConfig: ConfigSource("contour"),
 			CdsConfig: ConfigSource("contour"),
 		},
-		StaticResources: &bootstrap.Bootstrap_StaticResources{
+		StaticResources: &envoy_api_bootstrap.Bootstrap_StaticResources{
 			Clusters: []*api.Cluster{{
 				Name:                 "contour",
 				AltStatName:          strings.Join([]string{c.Namespace, "contour", strconv.Itoa(c.xdsGRPCPort())}, "_"),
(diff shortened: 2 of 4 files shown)


tsaarni:
I'm submitting this only to get comments and ask questions.  Sorry for failing test cases!

This PR is related to creating Envoy bootstrap configuration according to [this plan](https://docs.google.com/presentation/d/14sNKaRLqlyGbDVXcCzTH_MOT-LNmj8nJu_0IPYCptYI/edit?usp=sharing), based on the assumption I get my PR accepted by Envoy too.

Now instead of creating single file `/config/envoy.json` I need to create 3 files, e.g.

```
/config/envoy.json
/config/envoy-sds-auth-secret-tls-certicate.json
/config/envoy-sds-auth-secret-validation-context.json
```

In this PR I re-purposed the first command line argument for the target directory.  That is, instead of `contour bootstrap /config/envoy.json` one would execute `contour bootstrap /config` to generate the  config files to path `/config`

This change is backwards incompatible - bootstrap configuration generation will fail if user forgets to update the arguments of the `contour bootstrap` command in their deployment manifest.

Question: should we rather keep the old behavior by default and create another parameter for this?

Any other comments are of course welcome!

jpeach:
> Now instead of creating single file /config/envoy.json I need to create 3 files

IIRC, the files are already passed in as flags. So could you just use the files specifies by the flags?

tsaarni:
I think I did not understand how to do that.  Currently we only pass one path as a target where to write the configuration file, and named arguments to point out certificates and key.  Now when using SDS, there will be three configuration files (see [slide no 6](https://docs.google.com/presentation/d/14sNKaRLqlyGbDVXcCzTH_MOT-LNmj8nJu_0IPYCptYI/edit?usp=sharing) for illustration). These new config files contain SDS protobuf messages with JSON serialization.  

To provide paths for all three config files, maybe the command line could look something like:

```bash
contour bootstrap \
  /config/envoy.json \
  /config/envoy-sds-auth-secret-tls-certicate.json \
  /config/envoy-sds-auth-secret-validation-context.json \
  --xds-address=contour \
  --xds-port=8001 \
  --envoy-cafile=/certs/ca.crt \
  --envoy-cert-file=/certs/tls.crt \
  --envoy-key-file=/certs/tls.key
```

But here the position of unnamed arguments would become very critical and easy to mess up.  

Alternatively I could use the current arguments

```bash
contour bootstrap \
  /config/envoy.json \
  --xds-address=contour \
  --xds-port=8001 \
  --envoy-cafile=/certs/ca.crt \
  --envoy-cert-file=/certs/tls.crt \
  --envoy-key-file=/certs/tls.key
```

then get dirname of `/config/envoy.json` and write SDS config files with hardcoded filenames into that directory.  That would be backwards compatible but that does not sound too elegant either?

tsaarni:
Of course, yet another viewpoint is that when NOT using TLS, the SDS configuration files would not need to be created at all.

jpeach:
I'll play with this and review next week. I agree that we should add a flag to enable this new behaviour; not sure what that should be yet.

jpeach:
Ping @jpeach

jpeach:
@tsaarni A couple of questions

- If the envoy change to watch secrets filenames lans, do we still need to split the SDS config into separate files? i.e. is this change still needed in that case?

- I think that it's worth exploring how we do on-disk bootstrap in general. I looked at Envoy layered_config, but couldn't get it to work. If we could get this to work, then it could be a useful customization point as well as rationalize the SDS config files.

tsaarni:
> * If the envoy change to watch secrets filenames lans, do we still need to split the SDS config into separate files? i.e. is this change still needed in that case?

Yes this change is needed: this is preparation for that Envoy PR.  I have also some material linked in https://github.com/projectcontour/contour/pull/2333#issuecomment-597594709

The need to split the files comes from moving from static files to SDS resources.  And the motivation to move to SDS is that it had existing logic to hot-reload the certificates.

The number of files come from the two config parameters that refer to the SDS resources (`tls_certificate_sds_secret_configs` and `validation_context_sds_secret_config` [doc here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/cert.proto.html#envoy-v3-api-msg-extensions-transport-sockets-tls-v3-sdssecretconfig)). 

Here is a snippet of static bootstrap that refers to two `path` based SDS resources:

```json
{
  "transport_socket": {
    "name": "envoy.transport_sockets.tls",
    "typed_config": {
      "@type": "type.googleapis.com/envoy.api.v2.auth.UpstreamTlsContext",
      "common_tls_context": {
        "tls_certificate_sds_secret_configs": {
          "sds_config": {
            "path": "/path/to/envoy-sds-auth-secret-tls-certicate.yaml"
          }
        },
        "validation_context_sds_secret_config": {
          "sds_config": {
            "path": "/path/to/envoy-sds-auth-secret-validation-context.yaml"
          }
        }
      }
    }
  }
}
```

and the referred files look like this:

```json
{
  "resources": [
    {
      "@type": "type.googleapis.com/envoy.api.v2.auth.Secret",
      "tls_certificate": {
        "certificate_chain": {
          "filename": "/path/to/envoy.pem"
        },
        "private_key": {
          "filename": "/path/to/envoy-key.pem"
        }
      }
    }
  ]
}
```

```json
{
  "resources": [
    {
      "@type": "type.googleapis.com/envoy.api.v2.auth.Secret",
      "validation_context": {
        "trusted_ca": {
          "filename": "/path/to/internal-root-ca.pem"
        }
      }
    }
  ]
}
```

This was of course all existing logic within Envoy's protobuf based configuration and SDS code.
What the Envoy PR changes is that Envoy will make an inotify subscription to `path` resources and the referred `.pem` files to pick up changes in them.  Otherwise it relies to the same logic as it would have if these resources were sent via wire over xDS.

> * I think that it's worth exploring how we do on-disk bootstrap in general. I looked at Envoy layered_config, but couldn't get it to work. If we could get this to work, then it could be a useful customization point as well as rationalize the SDS config files.

According to my (very limited) understanding, Envoy's [runtime configuration](https://www.envoyproxy.io/docs/envoy/latest/configuration/operations/runtime) with `layered_runtime`i.e. *virtual file system* layering is another way to configure Envoy.  Possibly it was used *before* the protobuf based configuration was introduced? There is very little documentation or test code that I could study, but from what I can see it is not compatible with the protobuf based configuration API.  

It is not 100% clear to me what is the scope of virtual file system based config. In the [documentation](https://www.envoyproxy.io/docs/envoy/latest/configuration/operations/runtime#virtual-file-system) there is example of setting `/srv/runtime/current/envoy/health_check/min_interval`
which relates to very limited set of [runtime settings](
https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_runtime.html). It looks to me that it cannot cover the complete protobuf based config and therefore I suspect this cannot be used by Contour to provide customization hooks for Envoy config. Not 100% sure though!

tsaarni:
Support for certificate rotation is now scheduled for Envoy 1.14.0

* sds: certificate hot-reload for xDS gRPC connection https://github.com/envoyproxy/envoy/pull/10163
* doc: certificate hot-reload for xDS gRPC connection https://github.com/envoyproxy/envoy/pull/10628

jpeach:
@tsaarni Sorry it took so long for me to get back to you.

I think that the flag we need to add is `--envoy-resources-dir`. If this is set, then the bootstrap is allowed to add out-of-line resources into that directory.  This flag serves double-duty in that it enables out-of-line resources as well as specifying where they should be stored. Inside this directory, I think we can create directories for each resource class, i.e. `sds`, `cds`,  `rds`. Contour can use internal names for the resources that aren't controllable by users. 

This faked-up diff shows more about what I'm getting at here.
```
--- cmd/contour/bootstrap.go
+++ cmd/contour/bootstrap.go
@@ -28,7 +30,7 @@ func registerBootstrap(app *kingpin.Application) (*kingpin.CmdClause, *bootstrap
 	var ctx bootstrapContext
 
 	bootstrap := app.Command("bootstrap", "Generate bootstrap configuration.")
-	bootstrap.Arg("path", "Configuration file ('-' for standard output).").Required().StringVar(&ctx.path)
+	bootstrap.Arg("envoy-resources-dir", "Directory where configuration files will be written to.").StringVar(&ctx.config.ResourcesDir)
 	bootstrap.Flag("admin-address", "Envoy admin interface address.").StringVar(&ctx.config.AdminAddress)
 	bootstrap.Flag("admin-port", "Envoy admin interface port.").IntVar(&ctx.config.AdminPort)
 	bootstrap.Flag("xds-address", "xDS gRPC API address.").StringVar(&ctx.config.XDSAddress)

@@ -42,28 +44,41 @@ func registerBootstrap(app *kingpin.Application) (*kingpin.CmdClause, *bootstrap
-// doBootstrap writes an Envoy bootstrap configuration file to the supplied path.
+// doBootstrap writes an Envoy bootstrap configuration files to the supplied path.
 func doBootstrap(ctx *bootstrapContext) {
-	var out io.Writer
 
-	switch ctx.path {
-	case "-":
-		out = os.Stdout
-	default:
-		f, err := os.Create(ctx.path)
-		check(err)
+	if ctx.config.GrpcClientCert != "" || ctx.config.GrpcClientKey != "" || ctx.config.GrpcCABundle != "" {
+		// If one of the two TLS options is not empty, they all must be not empty
+		if !(ctx.config.GrpcClientCert != "" && ctx.config.GrpcClientKey != "" && ctx.config.GrpcCABundle != "") {
+			log.Fatal("You must supply all three TLS parameters - --envoy-cafile, --envoy-cert-file, --envoy-key-file, or none of them.")
+		}
 
-		out = f
+		if ctx.config.ResourcesDir != nil {
+			// TODO(jpeach): update constant names for these resource files.
+			writeConfig(path.Join(ctx.config.ResourcesDir, "sds", "xds-secrets.json"),
+				envoy.TLSCertificateSdsSecretConfig(ctx.config.GrpcClientCert, ctx.config.GrpcClientKey))
+
+			writeConfig(path.Join(ctx.config.ResourcesDir, "sds", "xds-validation.json"),
+				envoy.ValidationContextSdsSecretConfig(ctx.config.GrpcCABundle))
+		}
 
-		defer func() {
-			check(f.Close())
-		}()
 	}
 
-	m := &jsonpb.Marshaler{OrigName: true}
+	writeConfig(path.Join(ctx.config.Path, envoy.BootstrapConfigFile), envoy.Bootstrap(&ctx.config))
+}
+
+func writeConfig(filename string, config proto.Message) {
+	check(os.MkdirAll(path.Dir(filename), 0700))
 
-	check(m.Marshal(out, envoy.Bootstrap(&ctx.config)))
+	out, err := os.Create(filename)
+	check(err)
+
+	defer func() {
+		check(out.Close())
+	}()
+
+	m := &jsonpb.Marshaler{OrigName: true}
+	check(m.Marshal(out, config))
 }

--- internal/envoy/bootstrap.go
+++ internal/envoy/bootstrap.go
@@ -176,6 +264,10 @@ type BootstrapConfig struct {
 
 	// GrpcClientKey is the filename that contains a client key for secure gRPC with TLS.
 	GrpcClientKey string
+
+	// ResourcesDir is the directory where out of line Envoy resources can
+	// be placed.
+	ResourcesDir string
 }
 
 func (c *BootstrapConfig) xdsAddress() string   { return stringOrDefault(c.XDSAddress, "127.0.0.1") }
```

jpeach:
@stevesloka @youngnick PTAL at the `--envoy-resources-dir` suggestion above.

tsaarni:
I updated the PR according to `--envoy-resources-dir` suggestion.  Tests do not compile right now, but I will update them after we agree on the approach!

stevesloka:
Why do we need different files? Why inject? I'm not following the need for this. 

If the secret is rotated, we just need to know it happened and reload right?

I don't like the idea of users being able to inject their own arbitrary Envoy configuration into Contour.

tsaarni:
@stevesloka I have added feature to Envoy that can be used to get Envoy to automatically reload its xDS gRPC certificates.  Taking that feature into use requires changes to Envoy bootstrap configuration, therefore this PR is required in Contour.

I have documented the new Envoy feature here https://www.envoyproxy.io/docs/envoy/v1.14.1/configuration/security/secret#example-three-certificate-rotation-for-xds-grpc-connection

I have also depicted the solution from Contour's perspective in [this slideset](https://docs.google.com/presentation/d/14sNKaRLqlyGbDVXcCzTH_MOT-LNmj8nJu_0IPYCptYI/edit?usp=sharing) on page 6.

stevesloka:
Sure bootstrap can change, but why does it need to be injected by users? In my head, we would just change the init command to generate a different bootstrap config.

tsaarni:
This PR only changes the command to generate different bootstrap config - user has no way to inject custom configuration.  

I believe @jpeach just proposed a directory layout for the new bootstrap config that could potentially also facilitate user specified customization at later some point which was is part of this PR in any way.  He was responding to my question in this comment https://github.com/projectcontour/contour/pull/2333#issuecomment-597594709

jpeach:
> Why do we need different files? Why inject? I'm not following the need for this.

Envoy watches secrets to reload them only if SDS is being used. But the only way to use SDS is to break out the secrets resources in this context is to separate files on disk.

> If the secret is rotated, we just need to know it happened and reload right?

Yep, pretty much.

> I don't like the idea of users being able to inject their own arbitrary Envoy configuration into Contour.

That's not being proposed.

What I'm proposing is a general approach that can facilitate secrets reloading, rather than a special case that only addresses secrets.

tsaarni:
Over at the community meeting today I promised to add documentation links that explain why two new files are needed.

We use [CommonTlsContext](https://www.envoyproxy.io/docs/envoy/v1.14.1/api-v2/api/v2/auth/cert.proto#auth-commontlscontext) message to configure certificates and keys. Currently [tls_certificates](https://www.envoyproxy.io/docs/envoy/v1.14.1/api-v2/api/v2/auth/cert.proto#envoy-api-msg-auth-tlscertificate) and [validation_context](https://www.envoyproxy.io/docs/envoy/v1.14.1/api-v2/api/v2/auth/cert.proto#envoy-api-msg-auth-certificatevalidationcontext) attributes are used there.  These are of type [DataSource](https://www.envoyproxy.io/docs/envoy/v1.14.1/api-v2/api/v2/core/base.proto#envoy-api-msg-core-datasource), which can hold either inline PEM buffer or filenames that point to PEM files. 

The problem is that there is no hot-reload support when certificates and keys are configured in this way.  When investigating this, it was pointed out to me that we could use SDS path based resources instead.  The benefit here is that SDS logic in Envoy already knows how to hot-reload certificates and keys into active SSL context without disconnecting.  Envoy also knew how to load SDS resources from a file instead of loading them from xDS server. All I needed to add was a trigger to activate this behavior when the certificate and key files changed on disk. 

The complexity comes from the fact, that when using SDS resources on local path, we need to refer not only to the certificate and key files, but also to two new files: the JSON representation of the SDS resources / protobuf messages. See https://github.com/projectcontour/contour/pull/2333#issuecomment-605827761.

So the proposed change is:

The bootstrap will now contain [CommonTlsContext](https://www.envoyproxy.io/docs/envoy/v1.14.1/api-v2/api/v2/auth/cert.proto#auth-commontlscontext) message with tls_certificate_sds_secret_configs and validation_context_sds_secret_config attributes.  These refer to type [SdsSecretConfig](https://www.envoyproxy.io/docs/envoy/v1.14.1/api-v2/api/v2/auth/cert.proto#envoy-api-msg-auth-sdssecretconfig) which uses [ConfigSource](https://www.envoyproxy.io/docs/envoy/v1.14.1/api-v2/api/v2/core/config_source.proto#envoy-api-msg-core-configsource) message type (instead of [DataSource](https://www.envoyproxy.io/docs/envoy/v1.14.1/api-v2/api/v2/core/base.proto#envoy-api-msg-core-datasource) as previously).  Note that there is no "inline" alternative in the ConfigSource message, so we cannot inline the SDS resources - we need to write two new files.

There is no way for user to inject anything additional: the main bootstrap file explicitly refers to these two new files.  Furthermore it refers to the files in context of tls_certificate_sds_secret_configs and validation_context_sds_secret_config attributes, which can only refer to SDS resources like shown in https://github.com/projectcontour/contour/pull/2333#issuecomment-605827761.   Envoy would reject the configuration if user would try to replace the content of these files with some other protobuf message types.

stevesloka:
I see now, thanks for clarifying with the docs @tsaarni! The param makes sense to point to a path on the file system. I wonder if we need the parameter at all to define where the certs will live. Seems like we can default to our own, and folks could map them in as they see fit from the secret reference. 

If we need to keep the param, we should call it something that makes it specific to SDS and secrets, to me, the name `envoy-resources-dir` seems too vague for what it's really going to be used for, but names are hard. =)

tsaarni:
> Seems like we can default to our own, and folks could map them in as they see fit from the secret reference.

How could the default configuration directory layout look?  

Currently the to-be-generated bootstrap file is given as parameter now `contour bootstrap
  /config/envoy.json` but I don't know if it is based on a real use case - or could there be hardcoded layout e.g. just create all configs in `/config/`?

As discussed above, I think one aspect in whatever approach we take is backwards compatibility: users might upgrade Contour without studying all the detailed changes in example deployment manifest. If semantics of command line parameters change we might cause problems to users.

> I wonder if we need the parameter at all to define where the certs will live.

Specifically related to certificates:  the possibility to customize the certificate/key paths is important when integrating with different software that issues the certificates.  For example, there Secrets created by `contour certgen` look different than the ones created by cert-manager.  I wrote about the differences in https://github.com/projectcontour/contour/issues/2149#issuecomment-598065053 after experimenting with cert-manager.

jpeach:
> If we need to keep the param, we should call it something that makes it specific to SDS and secrets, to me, the name `envoy-resources-dir` seems too vague for what it's really going to be used for, but names are hard. =)

The reason I suggested a generic name for this was that I was concerned that we might start using the external files for more resource kinds over time. The idea behind `envoy-resources-dir` is that it can be a contour-managed hierarchy of arbitrary envoy resources. I didn't want to end up with extra flags each time we did this, so I thought that we could establish a pattern for doing this with envoy resources in a general way, then use that for SDS.

youngnick:
Okay, sorry to have not checked in here for a while.

I think the two important things to keep in mind here are:
- We need to ensure that we can rotate the certs when the secrets change. To do this, Envoy requires we do things in a certain way, which inolves multiple files. Passing a path is the cleanest way to do that.
- We need to ensure that the current behaviour - that passing a file outputs a complete config in *only* that file - is kept. Adding a new parameter is the cleanest way to do that, as the presence of the parameter is also a boolean indicating this new mode should be enabled.

I agree that keeping this with a more generic name is probably better, as it allows us to use this option for other things if we need to.

I don't think that we need to be concerned about people using this to pass arbitrary Envoy config, as Envoy is configured by the bootstrap. If people want to write their own bootstrap config, that's fine, and has always been intended as an extension point. If they want seamless certificate rotation, they'll need to do this as well though.

internal/envoy/bootstrap.go
```diff
@@ -27,11 +30,25 @@ import (
...
+// client certificate and key for XDS gRPC connection.
+const sdsTLSCertificateFile = "xds-tls-certicate.json"
+
+// sdsValidationContextFile stores the path to the SDS resource with
+// CA certificates for Envoy to use for the XDS gRPC connection.
+const sdsValidationContextFile = "xds-validation-context.json"
+
 // Bootstrap creates a new v2 Bootstrap configuration.
-func Bootstrap(c *BootstrapConfig) *bootstrap.Bootstrap {
+func Bootstrap(c *BootstrapConfig) {
```
jpeach, line 51: I think that this function ought to be broken down into internal helpers so that we can keep the existing tests working, and add coverage for the new stuff.

Here's what I would recommend:
- break out a helper that generates the Bootstrap struct
- now you can make a map of file path to content; this can be used to update the tests
- rename this function to `WriteBootstrap` and update doc comment accordingly
- all you need to do here is loop through the map and write each pair
- return an error from here if any of the output writing fails

internal/envoy/bootstrap.go
```diff
@@ -141,6 +173,81 @@ func upstreamFileTLSContext(cafile, certfile, keyfile string) *envoy_api_v2_auth
...
+						},
+					},
+				},
+			},
+		},
+	}
+	return context
+}
+
+func discoveryResponse(secret *envoy_api_v2_auth.Secret) *api.DiscoveryResponse {
```
jpeach, line 200: Let's drop this. I think that it's simple enough and more obvious to just provide the literal inline.

internal/envoy/bootstrap.go
```diff
@@ -141,6 +173,81 @@ func upstreamFileTLSContext(cafile, certfile, keyfile string) *envoy_api_v2_auth
...
+	response := &api.DiscoveryResponse{
+		Resources: []*any.Any{toAny(secret)},
+	}
+	return response
+}
+
+// tlsCertificateSdsSecretConfig creates DiscoveryResponse with file based SDS resource
+// including paths to TLS certificates and key
+func tlsCertificateSdsSecretConfig(certfile, keyfile string) *api.DiscoveryResponse {
```
jpeach, line 209: If this function, `validationContextSdsSecretConfig` and the internal `boostrapConfig` one all take a `*BootstrapConfig` argument, then you can have a consistent function signature for all of them, which will make the caller pretty clean (it can just loop through a slice of function pointers).

internal/envoy/bootstrap.go
```diff
@@ -98,12 +115,27 @@ func Bootstrap(c *BootstrapConfig) *bootstrap.Bootstrap {
...
+			writeConfig(path.Join(c.ResourcesDir, "sds", sdsTLSCertificateFile), tlsCertificateSdsSecretConfig(c.GrpcClientCert, c.GrpcClientKey))
+			writeConfig(path.Join(c.ResourcesDir, "sds", sdsValidationContextFile), validationContextSdsSecretConfig(c.GrpcCABundle))
+			tls = UpstreamTLSTransportSocket(upstreamSdsTLSContext(path.Join(c.Path, sdsTLSCertificateFile), path.Join(c.Path, sdsValidationContextFile)))
+		} else {
+			// Old behavior is to use direct certificate and key file paths in bootstrap config.
+			// Envoy does not support rotation of xDS certificate files in this case.
+			tls = UpstreamTLSTransportSocket(upstreamFileTLSContext(c.GrpcCABundle, c.GrpcClientCert, c.GrpcClientKey))
+		}
+		_ = tls
+		//b.StaticResources.Clusters[0].TransportSocket = tls
```
jpeach, line 135: TODO 🤣

internal/envoy/bootstrap.go
```diff
@@ -199,3 +312,28 @@ func intOrDefault(i, def int) int {
 	}
 	return i
 }
+
+func writeConfig(filename string, config proto.Message) {
```
jpeach, line 316: We should return the error from this, letting it bubble all the way up the stack to be handled in the main setup.

internal/envoy/bootstrap.go
```diff
@@ -27,11 +30,25 @@ import (
 	envoy_api_v2_core "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
 	bootstrap "github.com/envoyproxy/go-control-plane/envoy/config/bootstrap/v2"
 	matcher "github.com/envoyproxy/go-control-plane/envoy/type/matcher"
+	"github.com/gogo/protobuf/proto"
```
jpeach, line 33:
```suggestion
"github.com/google/protobuf/proto"
```

cmd/contour/bootstrap.go
```diff
@@ -14,56 +14,30 @@
...
-	bootstrap.Flag("envoy-key-file", "gRPC Client key filename for Envoy to load.").Envar("ENVOY_KEY_FILE").StringVar(&ctx.config.GrpcClientKey)
-	bootstrap.Flag("namespace", "The namespace the Envoy container will run in.").Envar("CONTOUR_NAMESPACE").Default("projectcontour").StringVar(&ctx.config.Namespace)
-	return bootstrap, &ctx
-}
-
-type bootstrapContext struct {
-	config envoy.BootstrapConfig
-	path   string
+	bootstrap.Arg("path", "Configuration file ('-' for standard output).").Required().StringVar(&config.Path)
+	bootstrap.Arg("envoy-resources-dir", "Directory where configuration files will be written to.").StringVar(&config.ResourcesDir)
```
jpeach, line 28: I don't know how args are supposed to work in kingpin, but I couldn't figure out the right syntax:
```bash
$ contour bootstrap --help
usage: contour bootstrap [<flags>] <path> [<envoy-resources-dir>]
...
$ contour bootstrap /tmp/contour/bootstrap.json /tmp/contour
open /tmp/contour/bootstrap.json/envoy.json: no such file or directory
```

I think that we should make this `--resources-dir` as a flag.

cmd/contour/bootstrap.go
```diff
@@ -14,56 +14,30 @@
...
-		defer func() {
-			check(f.Close())
-		}()
-	}
-
-	m := &jsonpb.Marshaler{OrigName: true}
-
-	check(m.Marshal(out, envoy.Bootstrap(&ctx.config)))
+// doBootstrap writes an Envoy bootstrap configuration files to the supplied paths.
+func doBootstrap(config *envoy.BootstrapConfig) {
```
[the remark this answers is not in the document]
tsaarni: Done!

internal/envoy/bootstrap.go
```diff
@@ -98,12 +115,27 @@ func Bootstrap(c *BootstrapConfig) *bootstrap.Bootstrap {
 		if !(c.GrpcClientCert != "" && c.GrpcClientKey != "" && c.GrpcCABundle != "") {
 			log.Fatal("You must supply all three TLS parameters - --envoy-cafile, --envoy-cert-file, --envoy-key-file, or none of them.")
 		}
-		b.StaticResources.Clusters[0].TransportSocket = UpstreamTLSTransportSocket(
-			upstreamFileTLSContext(c.GrpcCABundle, c.GrpcClientCert, c.GrpcClientKey),
-		)
+
+		var tls *envoy_api_v2_core.TransportSocket
+
```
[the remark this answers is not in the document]
tsaarni: Done!

internal/envoy/bootstrap.go
tsaarni: Done!

internal/envoy/bootstrap.go
tsaarni: Done!

internal/envoy/bootstrap.go
tsaarni: Done! :-)

internal/envoy/bootstrap.go
tsaarni: I aligned the arguments, but the return types of the three functions need to be specific types since they are used by calling code, so I could not align the return types to `proto.Message`.

There is also surrounding logic in `Bootstrap()` where these function are called from, related to how/when to call, how to use return values, so I could not figure out a way how to call them in a loop as function pointers.

Writing out the resulting `proto.Message` buffers as config files in `WriteBootstrap()` is now a loop though.

internal/envoy/bootstrap.go
tsaarni: Done!

internal/envoy/bootstrap.go
tsaarni: Oops, sorry!  Apparently now I have to deal with vscode autocorrecting and adding stuff like this, just like my phone! ;)  This should have been `github.com/golang/protobuf/proto` instead.

cmd/contour/bootstrap.go
tsaarni: Done!

internal/envoy/bootstrap.go
jpeach: I like the refactoring that you have done here. I think that if we take it a bit further, it cleans up even more

https://gist.github.com/jpeach/3a0e331181357408b544c1185037f749

LMK what you think?

tsaarni: Thanks, did this too and adjusted the test case accordingly.

PR #2334, opened 2020-03-11
doc: improve HeaderCondition API documentation

jpeach:
Make sure we accurately describe the types and semantics
of the HeaderCondition fields.

### DIFF
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -58,36 +58,37 @@ type Condition struct {
 	Header *HeaderCondition `json:"header,omitempty"`
 }
 
-// HeaderCondition specifies the header condition to match.
-// Name is required. Only one of Present or Contains must
-// be provided.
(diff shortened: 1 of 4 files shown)


apis/projectcontour/v1/httpproxy.go
```diff
@@ -58,36 +58,36 @@ type Condition struct {
...
-
-	// Name is the name of the header to match on. Name is required.
+	// Name is the name of the header to match against. Name is required.
 	// Header names are case insensitive.
 	Name string `json:"name"`
 
-	// Present is true if the Header is present in the request.
+	// Present specifies that condition is true when the named header
+	// is present, regardless of its value. Note that setting Present
+	// to false does not make the false true if the named header is absent.
```
youngnick:
```suggestion
	// to false does not make the Headercondition true if the named header is absent.
```
The "false true" part is a little tricky to parse, is this correct?

jpeach: Good catch, fixed.

PR #2339, opened 2020-03-16
cmd/contour: unify client creation

davecheney:
Updates #403

Before I can add ingress status support to cmd/contour I need to clean
up the serve method. This change unifies the creation of the various
clients to a single method and improves the naming of the various return
values -- they are infomer _factories_ not informers.

### DIFF
--- a/cmd/contour/clients.go
+++ b/cmd/contour/clients.go
@@ -1,0 +1,91 @@
+// Copyright © 2020 VMware
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"time"
+
+	"k8s.io/client-go/dynamic"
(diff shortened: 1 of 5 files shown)


cmd/contour/clients.go
```diff
@@ -0,0 +1,79 @@
...
+	"k8s.io/client-go/dynamic/dynamicinformer"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+
+	coordinationv1 "k8s.io/client-go/kubernetes/typed/coordination/v1"
+)
+
+// kubernetesClients holds the various API clients required by Contour.
+type kubernetesClients struct {
```
jpeach: Can we put this in the `internal.k8s` package along with the other generic Kubernetes helpers?

davecheney: Sure thing. Please hold

internal/k8s/clients.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// NewDynamicInformerFactory returns a new DynamicSharedInformerFactory for
+// use with any registered Kubernetes API type.
+func (c *Clients) NewDynamicInformerFactory() dynamicinformer.DynamicSharedInformerFactory {
+	return dynamicinformer.NewDynamicSharedInformerFactory(c.dynamic, resyncInterval)
+}
+
+// ClientSet returns the Kubernetes Core v1 ClientSet.
+func (c *Clients) ClientSet() *kubernetes.Clientset {
```
davecheney: These getters are unpleasant, I'll try to clean them up in a future PR.

jpeach: Yeh, I'd be inclined to just make the struct members public, but this is fine too.

PR #2342, opened 2020-03-16
internal/k8s: simplify DynamicConverter

davecheney:
Updates #403

Currently in cmd/contour.doServe we have three ResourceEventHandler
types defined;

- eventHandler, which is the internal/contour.EventHandler connected to
the dag
- eventRecorder, which wraps eventHandler to add metrics
- dynamicHandler, which wraps eventRecorder to handle dynamic objects.

These three objects are all in scope and used to register with various
informers in a fragile pattern; eventHandler needs to be in scope to
be registed with the group later in the function, eventRecorder is
registered with the k8s core types, and dynamicHandler with the CRD
types. This is all rather fragile.

This PR prepares to make the DynamicHandler the only entry point for all
k8s types by refactoring the DynamicHandler to make accepting non
dynamic types less unexpected.

### DIFF
--- a/internal/k8s/converter.go
+++ b/internal/k8s/converter.go
@@ -40,52 +40,39 @@ type DynamicClientHandler struct {
 }
 
 func (d *DynamicClientHandler) OnAdd(obj interface{}) {
-	if d.Converter.CanConvert(obj) {
-		var err error
-		obj, err = d.Converter.Convert(obj)
-		if err != nil {
-			d.Logger.Error(err)
-			return
-		}
+	obj, err := d.Converter.Convert(obj)
+	if err != nil {
+		d.Logger.Error(err)
+		return
 	}
 	d.Next.OnAdd(obj)
 }
 
 func (d *DynamicClientHandler) OnUpdate(oldObj, newObj interface{}) {
-	if d.Converter.CanConvert(oldObj) {
-		var err error
-		oldObj, err = d.Converter.Convert(oldObj)
-		if err != nil {
-			d.Logger.Error(err)
-			return
-		}
+	oldObj, err := d.Converter.Convert(oldObj)
+	if err != nil {
+		d.Logger.Error(err)
+		return
 	}
-	if d.Converter.CanConvert(newObj) {
-		var err error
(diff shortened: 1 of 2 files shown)


jpeach:
Also, in the commit message s/DyanmicConverter/DynamicConverter/

davecheney:
> Without seeing the later changes, it's hard to know how much simpler this is. After this change, the caller know makes the assumption that it knows what the converter can convert, so maybe consider a pattern like this:
> 
> ```go
> if d.Converter.CanConvert(obj) {
>   obj = d.Converter.MustConvert(obj)
> }
> ```

But `ConConvert` never asserted that it could convert the object, only that it was of a specific type. Maybe its simpler to always convert,  and return the original object on failure.

jpeach:
> But ConConvert never asserted that it could convert the object, only that it was of a specific type. Maybe its simpler to always convert, and return the original object on failure.

Yeh, arguably, it isn't fulfilling the promise of it's name :)

jpeach:
> Maybe its simpler to always convert, and return the original object on failure.

Yeh, that could work too

internal/k8s/converter.go
```diff
@@ -112,16 +99,12 @@ func NewUnstructuredConverter() (*UnstructuredConverter, error) {
...
-}
-
-// Convert converts an unstructured.Unstructured to typed struct
+// Convert converts an unstructured.Unstructured to typed struct. If obj
+// is not an unstructured.Unstructured it is returned without further processing.
 func (c *UnstructuredConverter) Convert(obj interface{}) (interface{}, error) {
 	unstructured, ok := obj.(*unstructured.Unstructured)
 	if !ok {
-		return nil, fmt.Errorf("unsupported object type: %T", obj)
+		return obj, nil
```
jpeach: If `Convert` passes through objects that it doesn't handle, should it also do so in the `default` switch case?

davecheney: I don't think so. If the object isn't an unstructured, we return it. If it _is_ an unstructured we try to decode it, and if that fails we balk.

PR #2348, opened 2020-03-17
cmd/contour: simplify leadership election setup

davecheney:
Updates #403

Ingress status updates depend on leadership election. This PR moves
leadership setup outside doServe and simplifies the interaction with
EventHandler.IsLeader.

### DIFF
--- a/cmd/contour/leadership.go
+++ b/cmd/contour/leadership.go
@@ -18,16 +18,65 @@ import (
 	"os"
 
 	"github.com/google/uuid"
+	"github.com/projectcontour/contour/internal/k8s"
+	"github.com/projectcontour/contour/internal/workgroup"
 	"github.com/sirupsen/logrus"
-	"k8s.io/client-go/kubernetes"
-	coordinationv1 "k8s.io/client-go/kubernetes/typed/coordination/v1"
 	"k8s.io/client-go/tools/leaderelection"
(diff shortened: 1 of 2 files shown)


davecheney:
@jpeach I looked at moving this logic to `internal/k8s` but the dependency on `cmd/contour.serveContext` is large and the latter type should not live in `internal/contour`.

cmd/contour/leadership.go
```diff
@@ -18,24 +18,73 @@ import (
...
-	coordinationv1 "k8s.io/client-go/kubernetes/typed/coordination/v1"
 	"k8s.io/client-go/tools/leaderelection"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
 )
 
+// setupLeadershipElection registers leadership workers with the group and returns
+// a channel which will become ready when this process becomes the leader, or, in the
+// event that leadership election is disabled, the channel will be ready immediately.
+func setupLeadershipElection(g *workgroup.Group, log logrus.FieldLogger, ctx *serveContext, clients *k8s.Clients, updateNow func()) chan struct{} {
+	if ctx.DisableLeaderElection {
```
jpeach: Suggest you pull this into a separate function `becomeElectedLeader()`, then check `ctx.DisableLeaderElection` from the caller.

Issue #2219, opened 2020-02-11
Automatically enable detected GVKs


youngnick:
This issue is to discuss whether Contour should autodetect supported GVKs, both CRD and core, at startup and watch any supported objects that are installed in a cluster.

The current option is to have specific flags for types we want to enable or disable (#1461, #1241). We will keep doing this for now, but we can consider doing a more dynamic solution.

Pros of autodetecting:
- Less config flags
- Enables cluster admins to switch off CRD functionality by not installing CRDs
- Stops the surprise factor of having a CRD or other object installed that is not watched by Contour

Cons of autodetecting:
- Doesn't allow for clearly flagging things as experimental as easily - probably warning logs would be required.

youngnick:
We would probably need to use https://godoc.org/k8s.io/client-go/discovery or similar, for future reference.

jpeach:
xref: #1241

youngnick:
After some consideration of this, I'm going ahead with it, as it will allow for people to make an ultimate "disable this configuration type" decision by removing a CRD from the cluster. This will  help with IngressRoute -> HTTPProxy migration, version changes, and so on.

stevesloka:
@youngnick I have some work for this started, did you already start? I don't want to duplicate efforts.

youngnick:
I started in that I did some refactoring in preparation - my idea was that we would build the default set of informers using the bits I added already, and then wash that set against what's actually present on the apiserver. Could also do it the other way around, of course.

PR #2352, opened 2020-03-18
Move CRD informer list generation to k8s

youngnick:
This is in preparation for adding in filtering by discovered CRD types.

Updates #2219 

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -22,12 +22,6 @@ import (
 	"syscall"
 	"time"
 
-	projectcontour "github.com/projectcontour/contour/apis/projectcontour/v1"
-
-	ingressroutev1 "github.com/projectcontour/contour/apis/contour/v1beta1"
-
-	serviceapis "sigs.k8s.io/service-apis/api/v1alpha1"
-
 	"github.com/projectcontour/contour/internal/contour"
 	"github.com/projectcontour/contour/internal/dag"
 	"github.com/projectcontour/contour/internal/debug"
@@ -206,30 +200,26 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	// using the SyncList to keep track of what to sync later.
 	var informerSyncList k8s.InformerSyncList
 
-	informerSyncList.Add(dynamicInformerFactory.ForResource(ingressroutev1.IngressRouteGVR).Informer()).AddEventHandler(dynamicHandler)
-	informerSyncList.Add(dynamicInformerFactory.ForResource(ingressroutev1.TLSCertificateDelegationGVR).Informer()).AddEventHandler(dynamicHandler)
-	informerSyncList.Add(dynamicInformerFactory.ForResource(projectcontour.HTTPProxyGVR).Informer()).AddEventHandler(dynamicHandler)
-	informerSyncList.Add(dynamicInformerFactory.ForResource(projectcontour.TLSCertificateDelegationGVR).Informer()).AddEventHandler(dynamicHandler)
+	iset := k8s.DefaultInformerSet(dynamicInformerFactory, ctx.UseExperimentalServiceAPITypes)
 
-	informerSyncList.Add(informerFactory.Core().V1().Services().Informer()).AddEventHandler(dynamicHandler)
-	informerSyncList.Add(informerFactory.Networking().V1beta1().Ingresses().Informer()).AddEventHandler(dynamicHandler)
+	// TODO(youngnick): Add in filtering the iset map by enabled apiserver types (#2219) using the discovery library.
 
-	if ctx.UseExperimentalServiceAPITypes {
-		log.Info("Enabling Experimental Service APIs types")
-		informerSyncList.Add(dynamicInformerFactory.ForResource(serviceapis.GroupVersion.WithResource("gatewayclasses")).Informer()).AddEventHandler(dynamicHandler)
-		informerSyncList.Add(dynamicInformerFactory.ForResource(serviceapis.GroupVersion.WithResource("gateways")).Informer()).AddEventHandler(dynamicHandler)
-		informerSyncList.Add(dynamicInformerFactory.ForResource(serviceapis.GroupVersion.WithResource("httproutes")).Informer()).AddEventHandler(dynamicHandler)
-		informerSyncList.Add(dynamicInformerFactory.ForResource(serviceapis.GroupVersion.WithResource("tcproutes")).Informer()).AddEventHandler(dynamicHandler)
+	for _, inf := range iset.Informers {
+		informerSyncList.RegisterInformer(inf, dynamicHandler)
 	}
+
+	// We need to register this separately for now, as the core stuff uses a non-dynamic InformerFactory.
+	informerSyncList.RegisterInformer(informerFactory.Core().V1().Services().Informer(), dynamicHandler)
+	informerSyncList.RegisterInformer(informerFactory.Networking().V1beta1().Ingresses().Informer(), dynamicHandler)
 
 	// Add informers for each root-ingressroute namespaces
 	for _, factory := range namespacedInformerFactories {
-		informerSyncList.Add(factory.Core().V1().Secrets().Informer()).AddEventHandler(dynamicHandler)
+		informerSyncList.RegisterInformer(factory.Core().V1().Secrets().Informer(), dynamicHandler)
 	}
 
 	// If root-ingressroutes are not defined, then add the informer for all namespaces
 	if len(namespacedInformerFactories) == 0 {
-		informerSyncList.Add(informerFactory.Core().V1().Secrets().Informer()).AddEventHandler(dynamicHandler)
+		informerSyncList.RegisterInformer(informerFactory.Core().V1().Secrets().Informer(), dynamicHandler)
 	}
 
 	// step 5. endpoints updates are handled directly by the EndpointsTranslator
@@ -238,7 +228,7 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		FieldLogger: log.WithField("context", "endpointstranslator"),
 	}
 
-	informerSyncList.Add(informerFactory.Core().V1().Endpoints().Informer()).AddEventHandler(et)
+	informerSyncList.RegisterInformer(informerFactory.Core().V1().Endpoints().Informer(), et)
 
 	// step 6. setup workgroup runner and register informers.
 	var g workgroup.Group
--- a/internal/k8s/informers.go
+++ b/internal/k8s/informers.go
@@ -1,0 +1,62 @@
+// Copyright © 2020 VMware
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package k8s
+
+import (
+	projectcontour "github.com/projectcontour/contour/apis/projectcontour/v1"
+
+	ingressroutev1 "github.com/projectcontour/contour/apis/contour/v1beta1"
+
+	serviceapis "sigs.k8s.io/service-apis/api/v1alpha1"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic/dynamicinformer"
+	"k8s.io/client-go/tools/cache"
+)
+
+type gvrmap map[schema.GroupVersionResource]cache.SharedIndexInformer
+
+// InformerSet stores a table of Kubernetes GVR objects to their associated informers.
+type InformerSet struct {
+	Informers gvrmap
+}
+
+// DefaultInformerSet creates a new InformerSet lookup table and populates with the default
+// GVRs that Contour will try to watch.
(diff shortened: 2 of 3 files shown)


internal/k8s/gvrautodetect.go
```diff
@@ -0,0 +1,61 @@
...
+type GVRMap map[schema.GroupVersionResource]cache.SharedIndexInformer
+type GVRInformer struct {
+	Lookup GVRMap
+}
+
+type ForResource interface {
+	ForResource(gvr schema.GroupVersionResource) informers.GenericInformer
+}
+
+func NewGVRInformerLookupDefaults(inffactory ForResource, serviceAPIs bool) GVRInformer {
```
jpeach: Can you add godoc here? Just going by the name, this should return a new `GVRInformerLookupDefaults`, so I think we can improve on that.

internal/k8s/gvrautodetect.go
```diff
@@ -0,0 +1,61 @@
...
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/tools/cache"
+)
+
+type GVRMap map[schema.GroupVersionResource]cache.SharedIndexInformer
+type GVRInformer struct {
+	Lookup GVRMap
+}
+
+type ForResource interface {
```
jpeach: What's a `ForResource`? Can we think of a name that more aptly describes that this interface does?

internal/k8s/gvrautodetect.go
```diff
@@ -0,0 +1,61 @@
...
+	ingressroutev1 "github.com/projectcontour/contour/apis/contour/v1beta1"
+
+	serviceapis "sigs.k8s.io/service-apis/api/v1alpha1"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/tools/cache"
+)
+
+type GVRMap map[schema.GroupVersionResource]cache.SharedIndexInformer
```
jpeach: Maybe don't make a public type unless clients of the API will use it directly?

internal/k8s/gvrautodetect.go
```diff
@@ -0,0 +1,61 @@
...
+type GVRInformer struct {
+	Lookup GVRMap
```
jpeach: Suggest that this type will be easier to use with a method, maybe `InformerForResource(*schema.GroupVersionResource)`

cmd/contour/serve.go
```diff
@@ -206,22 +200,17 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	// using the SyncList to keep track of what to sync later.
 	var informerSyncList k8s.InformerSyncList
 
-	informerSyncList.Add(dynamicInformerFactory.ForResource(ingressroutev1.IngressRouteGVR).Informer()).AddEventHandler(dynamicHandler)
-	informerSyncList.Add(dynamicInformerFactory.ForResource(ingressroutev1.TLSCertificateDelegationGVR).Informer()).AddEventHandler(dynamicHandler)
-	informerSyncList.Add(dynamicInformerFactory.ForResource(projectcontour.HTTPProxyGVR).Informer()).AddEventHandler(dynamicHandler)
-	informerSyncList.Add(dynamicInformerFactory.ForResource(projectcontour.TLSCertificateDelegationGVR).Informer()).AddEventHandler(dynamicHandler)
+	// Build the default GVR to Informer lookup object
```
davecheney: suggestion: if the name `k8s.NewGVRInformerLookupDefaults` isn't good enough that it needs a descriptive comment, maybe we can change the name of the function so the comment is needed.

cmd/contour/serve.go
```diff
@@ -206,22 +200,17 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+	gvri := k8s.NewGVRInformerLookupDefaults(dynamicInformerFactory, ctx.UseExperimentalServiceAPITypes)
+
+	// Use the GVR to Informer lookup map to register all the required informers.
+	for _, inf := range gvri.Lookup {
+		informerSyncList.Add(inf).AddEventHandler(dynamicHandler)
```
davecheney: maybe if informerSyncList grew a new method, `RegisterInformer` then this comment wouldn't be needed.

internal/k8s/gvrautodetect.go
youngnick: Fair, made private.

internal/k8s/gvrautodetect.go
youngnick: The aim here is for this type to grow a method that will take the output of the client-go discovery bits (basically a list of available GVRs), and filter this map so it only contains GVRs the apiserver knows about. I don't see it needing to have individual lookups in a public way for now.

cmd/contour/serve.go
youngnick: Fair enough, changed.

cmd/contour/serve.go
youngnick: Fair enough, changed `Add` to `RegisterInformer`.

internal/k8s/gvrautodetect.go
youngnick: I was trying to make this more general so we could pass other types of InformerFactory to the default setup. But I think that was premature optimisation. I've made it just take the specific DynamicSharedIndexInformerFactory type.

internal/k8s/gvrautodetect.go
youngnick: I changed the name and added godoc.

go.sum
```diff
@@ -468,6 +468,7 @@ k8s.io/apiserver v0.0.0-20190918160949-bfa5e2e684ad/go.mod h1:XPCXEwhjaFN29a8Nld
 k8s.io/client-go v0.0.0-20190918160344-1fbdaa4c8d90/go.mod h1:J69/JveO6XESwVgG53q3Uz5OSfgsv4uxpScmmyYOOlk=
 k8s.io/client-go v0.17.0 h1:8QOGvUGdqDMFrm9sD6IUFl256BcffynGoe80sxgTEDg=
 k8s.io/client-go v0.17.0/go.mod h1:TYgR6EUHs6k45hb6KWjVD6jFZvJV4gHDikv/It0xz+k=
+k8s.io/client-go v11.0.0+incompatible h1:LBbX2+lOwY9flffWlJM7f1Ct8V2SRNiMRDFeiwnJo9o=
```
davecheney: looks like a local merge turd.

internal/k8s/gvrautodetect.go
```diff
@@ -0,0 +1,61 @@
...
+// GVRInformer stores a table of Kubernetes GVR objects to their associated informers.
+type GVRInformer struct {
+	Lookup gvrmap
+}
+
+// DefaultGVRInformer creates a new GVRInformer lookup table and populates with the default
+// GVRs that Contour will try to watch.
+func DefaultGVRInformer(inffactory dynamicinformer.DynamicSharedInformerFactory, serviceAPIs bool) GVRInformer {
+
+	var gvri GVRInformer
```
davecheney: maybe move this closer to its first use around 55 and use a compact literal

gvri := GVRInformer{
   Lookup: make(gvrmap),
}

internal/k8s/gvrautodetect.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	if serviceAPIs {
+		defaultGVRs = append(defaultGVRs, serviceapis.GroupVersion.WithResource("gatewayclasses"))
+		defaultGVRs = append(defaultGVRs, serviceapis.GroupVersion.WithResource("gateways"))
+		defaultGVRs = append(defaultGVRs, serviceapis.GroupVersion.WithResource("httproutes"))
+		defaultGVRs = append(defaultGVRs, serviceapis.GroupVersion.WithResource("tcproutes"))
+	}
+
+	gvri.Lookup = make(gvrmap)
```
davecheney: lockup isn't the best name for a map, maybe a better name is informers

internal/k8s/gvrautodetect.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic/dynamicinformer"
+	"k8s.io/client-go/tools/cache"
+)
+
+type gvrmap map[schema.GroupVersionResource]cache.SharedIndexInformer
+
+// GVRInformer stores a table of Kubernetes GVR objects to their associated informers.
+type GVRInformer struct {
```
jpeach: This is an object that holds a set (in the sense of "a collection") of informers, so how about calling it `InformerSet`? Then we can cal the helper `DefaultInformerSet`, which gives the reader a very good idea of what the call sites are doing.

internal/k8s/gvrautodetect.go
```diff
@@ -0,0 +1,61 @@
+// Copyright © 2020 VMware
```
jpeach: Is the name `gvrautodetect.go` still appropriate? or should we rename it `informerset.go`, or simply `informers.go`?

cmd/contour/serve.go
```diff
@@ -206,30 +200,25 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
-		informerSyncList.Add(dynamicInformerFactory.ForResource(serviceapis.GroupVersion.WithResource("gatewayclasses")).Informer()).AddEventHandler(dynamicHandler)
-		informerSyncList.Add(dynamicInformerFactory.ForResource(serviceapis.GroupVersion.WithResource("gateways")).Informer()).AddEventHandler(dynamicHandler)
-		informerSyncList.Add(dynamicInformerFactory.ForResource(serviceapis.GroupVersion.WithResource("httproutes")).Informer()).AddEventHandler(dynamicHandler)
-		informerSyncList.Add(dynamicInformerFactory.ForResource(serviceapis.GroupVersion.WithResource("tcproutes")).Informer()).AddEventHandler(dynamicHandler)
+	for _, inf := range gvri.Informers {
+		informerSyncList.RegisterInformer(inf, dynamicHandler)
 	}
 
+	informerSyncList.RegisterInformer(informerFactory.Core().V1().Services().Informer(), dynamicHandler)
+	informerSyncList.RegisterInformer(informerFactory.Networking().V1beta1().Ingresses().Informer(), dynamicHandler)
```
jpeach: We can get rid of this code if we add the corresponding GVRs to the `InformerSet`

internal/k8s/gvrautodetect.go
```diff
@@ -0,0 +1,61 @@
...
+
+// DefaultGVRInformer creates a new GVRInformer lookup table and populates with the default
+// GVRs that Contour will try to watch.
+func DefaultGVRInformer(inffactory dynamicinformer.DynamicSharedInformerFactory, serviceAPIs bool) GVRInformer {
+
+	defaultGVRs := []schema.GroupVersionResource{
+		ingressroutev1.IngressRouteGVR,
+		ingressroutev1.TLSCertificateDelegationGVR,
+		projectcontour.HTTPProxyGVR,
+		projectcontour.TLSCertificateDelegationGVR,
```
jpeach: Add default resources:
```
// Can use v1.SchemeGroupVersion.WithResource("foo") for core resources ...
schema.GroupVersionResource{Version: "v1", Resource: "services"}
schema.GroupVersionResource{Version: "v1", Resource: "secrets"}
schema.GroupVersionResource{Version: "v1", Resource: "endpoints"}
v1beta1.SchemeGroupVersion.WithResource("ingresses")
```

internal/k8s/gvrautodetect.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	if serviceAPIs {
+		defaultGVRs = append(defaultGVRs, serviceapis.GroupVersion.WithResource("gatewayclasses"))
+		defaultGVRs = append(defaultGVRs, serviceapis.GroupVersion.WithResource("gateways"))
+		defaultGVRs = append(defaultGVRs, serviceapis.GroupVersion.WithResource("httproutes"))
+		defaultGVRs = append(defaultGVRs, serviceapis.GroupVersion.WithResource("tcproutes"))
+	}
```
jpeach: I'd be inclined to have 2 global constants, `DefaultResources` and `ServiceResources` and then pass them into the constructor based on the flag at the case site. That has the benefit of being more explicit than a bool flag.

But just a suggestion, feel free to abstain :)

youngnick: I only had this in there for this refactor - I intend to remove this as part of making the actual detection automatic. The idea is that this will have everything, then anything not present will be removed.

internal/k8s/gvrautodetect.go
youngnick: Yep, good idea.

internal/k8s/gvrautodetect.go
youngnick: Great idea, updated.

internal/k8s/gvrautodetect.go
youngnick: I've also kept pattern we have with servecontext where the constructor returns a thing with all the correct defaults set, without reference to any other constants. We can talk more about that if you want.

cmd/contour/serve.go
youngnick: We can't without further refactoring, because the core and networking stuff comes from the `informerFactory` (which is not dynamic), and `DefaultInformerSet` currently takes a `DynamicSharedInformerFactory` only. 

I'll look at refactoring this as part of the next round of changes.

jpeach: OK, sounds good. AFAIK, there should not be any problem with using a dynamic informer to watch the core types.

internal/k8s/gvrautodetect.go
jpeach: 👍

cmd/contour/serve.go
youngnick: I think we'll need some extra unstructured translation, but I'll look at that next.

PR #2355, opened 2020-03-18
cmd/contour: add IngressStatusWriter

davecheney:
Updates #403

This PR adds a worker to doServe's group which writes back load balancer
status to ingress.v1beta1 objects.

In this PR, the population of the isw.lbStatus channel is not present
but has been manually verified with the following code:

                tick := time.Tick(2 * time.Second)
                for {
                        select {
                        case <-tick:
                                isw.lbStatus <- v1.LoadBalancerStatus{
                                        Ingress: []v1.LoadBalancerIngress{{
                                                Hostname: "gke.davecheney.com",
                                        }},
                                }
                        case <-stop:
                                return nil
                        }
                }
        })

Future PRs will add workers to populate the channel.

### DIFF
--- a/cmd/contour/ingressstatus.go
+++ b/cmd/contour/ingressstatus.go
@@ -1,0 +1,94 @@
+// Copyright © 2019 VMware
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"sync"
+
+	"github.com/projectcontour/contour/internal/k8s"
+	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+)
+
+// ingressStatusWriter manages the lifetime of StatusLoadBalancerUpdaters.
+//
+// The theory of operation of the ingressStatusWriter is as follows:
+// 1. On startup the ingressStatusWriter waits to be elected leader.
+// 2. Once elected leader, the ingressStatusWriter waits to receive a
+//    v1.LoadBalancerStatus value.
+// 3. Once a v1.LoadBalancerStatus value has been received, any existing informer
+//    is stopped and a new informer started in its place.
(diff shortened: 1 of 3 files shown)


internal/k8s/ingressstatus.go
```diff
@@ -0,0 +1,61 @@
...
+	Status v1.LoadBalancerStatus
+}
+
+func (s *StatusLoadBalancerUpdater) OnAdd(obj interface{}) {
+	ing := obj.(*v1beta1.Ingress).DeepCopy()
+
+	// TODO(dfc) check ingress class
+
+	ing.Status.LoadBalancer = s.Status
+	_, err := s.Client.NetworkingV1beta1().Ingresses(ing.GetNamespace()).UpdateStatus(ing)
```
jpeach: How do you know that this is an ingress we should update? Is the ingress class guaranteed here? Are there any namespace restrictions?

cmd/contour/ingressstatus.go
```diff
@@ -0,0 +1,94 @@
...
+			// stop existing informer
+			if shutdown != nil {
+				close(shutdown)
+			}
+			stopping.Wait()
+
+			// create informer for the new LoadBalancerStatus
+			factory := isw.clients.NewInformerFactory()
+			inf := factory.Networking().V1beta1().Ingresses().Informer()
+			inf.AddEventHandler(&k8s.StatusLoadBalancerUpdater{
```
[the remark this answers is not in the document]
davecheney: Of course, I'll take that as a TODO

internal/k8s/ingressstatus.go
davecheney: Ingress class is not namespace, contour should update any ingress objects that match its ingress class.

internal/k8s/ingressstatus.go
```diff
@@ -0,0 +1,61 @@
...
+
+func (s *StatusLoadBalancerUpdater) OnUpdate(oldObj, newObj interface{}) {
+	// Ignoring OnUpdate allows us to avoid the message generated
+	// from the status update.
+
+	// TODO(dfc) handle these cases:
+	// - OnUpdate transitions from an ingress class which is out of scope
+	// to one in scope.
+	// - OnUpdate transitions from an ingress class in scope to one out
+	// of scope.
```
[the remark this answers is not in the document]
davecheney: Its all part of 403

internal/k8s/ingressstatus.go
jpeach: Where's the check for the ingress class?

davecheney: It's coming up in the next PR. At the moment this code is inactive.

jpeach: I see. I'm not sure how to review then.

PR #2373, opened 2020-03-23
Migrate Service and Ingress to client-go dynamic client

youngnick:
This is further refactoring for #2219.

Moves Service and Ingress to the client-go dynamic client, leaving only Secrets and Endpoints not handled by the new InformerSet. Secrets need namespace handling, and Endpoints need a separate handler.

### DIFF
--- a/internal/k8s/converter.go
+++ b/internal/k8s/converter.go
@@ -19,8 +19,11 @@ import (
 	ingressroutev1 "github.com/projectcontour/contour/apis/contour/v1beta1"
 	projectcontour "github.com/projectcontour/contour/apis/projectcontour/v1"
 	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/api/networking/v1beta1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/tools/cache"
 	serviceapis "sigs.k8s.io/service-apis/api/v1alpha1"
 )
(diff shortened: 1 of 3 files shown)


internal/k8s/converter.go
```diff
@@ -91,6 +93,14 @@ func NewUnstructuredConverter() (*UnstructuredConverter, error) {
 	projectcontour.AddKnownTypes(uc.scheme)
 	ingressroutev1.AddKnownTypes(uc.scheme)
 
+	// Add the core types we need
+	uc.scheme.AddKnownTypes(v1.SchemeGroupVersion, &v1.Secret{}, &v1.Service{})
```
jpeach: IIUC, `scheme.Scheme` has all the core types registered, so maybe we could use that.

Or we could just do what [client-go does](https://github.com/kubernetes/client-go/blob/master/kubernetes/scheme/register.go#L131) which would at least mean that we don't need to keep track of which core types we add.

internal/k8s/informers.go
```diff
@@ -41,6 +42,8 @@ func DefaultInformerSet(inffactory dynamicinformer.DynamicSharedInformerFactory,
 		ingressroutev1.TLSCertificateDelegationGVR,
 		projectcontour.HTTPProxyGVR,
 		projectcontour.TLSCertificateDelegationGVR,
+		schema.GroupVersionResource{Version: "v1", Resource: "services"},
```
jpeach: For consistency:
```suggestion
import 	corev1 "k8s.io/api/core/v1"

corev1.SchemeGroupVersion.WithResource("services")
```

internal/k8s/converter.go
youngnick: Ah, this was a holdover from troubleshooting another problem I had with the converter. I'll revert to `scheme.AddToScheme()`. Thanks.

PR #2374, opened 2020-03-23
Authenticate Envoy DaemonSet

cortopy:
I'm not sure if editing this file is the right thing, but I can't find the `examples/render.sh` file mentioned in the header

If applied, this commit will authenticate the pods in the Envoy DaemonSet so that a PodSecurityPolicy can be targeted specifically for the Envoy containers. The Envoy image runs as root user and a default restricted policy for the namespace would not do.

As I understand the current manifest, the pods don't need a Role to run. However, a PodSecurityPolicy for this DaemonSet is much more difficult to apply, since we'd be looking at unauthenticated pods in the namespace and that could be anything running outside RBAC.

### DIFF
--- a/examples/contour/00-common.yaml
+++ b/examples/contour/00-common.yaml
@@ -2,10 +2,16 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: projectcontour 
+  name: projectcontour
 ---
 apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: contour
+  namespace: projectcontour
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: envoy
   namespace: projectcontour
--- a/examples/contour/03-envoy.yaml
+++ b/examples/contour/03-envoy.yaml
@@ -122,7 +122,7 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
-      automountServiceAccountToken: false
+      serviceAccountName: envoy
       terminationGracePeriodSeconds: 300
       volumes:
         - name: envoy-config
--- a/examples/render/contour.yaml
+++ b/examples/render/contour.yaml
@@ -17,12 +17,18 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: projectcontour 
+  name: projectcontour
 ---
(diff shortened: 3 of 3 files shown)


jpeach:
Hi @cortopy, the files you are looking for are in the `examples/contour` directory. The final YAML agglomeration is done by `make generate` or just `./hack/generate-deployment.sh`.

An envoy service account should probably go into `examples/contour/00-common.yaml`, since that is where the contour service account is as well.

jpeach:
xref Linkerd integration instructions https://linkerd.io/2/tasks/using-ingress/#contour

Note that these specify adding the envoy service account, so we should file an issue to let the Linkerd project know once this change lands.

cortopy:
Thanks @jpeach for already submitting the correction to the script. I've updated this PR with what I believe is the changes in the right file

cortopy:
thanks @jpeach for the detailed instructions!! I've npw built it and committed the changes

In re. the email address in sign-off I would prefer not to make my email address public. I've edited the signoff to have my proper github no reply address. Would this be enough? Otherwise I could create a new email account and signoff with that one

jpeach:
> thanks @jpeach for the detailed instructions!! I've npw built it and committed the changes
> 
> In re. the email address in sign-off I would prefer not to make my email address public. I've edited the signoff to have my proper github no reply address. Would this be enough? Otherwise I could create a new email account and signoff with that one

I checked on this, and the DCO sign-off should have a real name and email address, in case someone ever needs to contact you about the change. A separate email account is fine; many people use an email address specifically for contributions.

youngnick:
Hi @cortopy, thanks for this work, but we do require a DCO signoff, which must include an email that you are reachable on (in the case there is an ownership or IP issue at a later date.) If you can't give us an email that you will check, then we will not be able to merge this PR.

We do really need this functionality, so I'm unfortunately going to need to give you a deadline for this. I'll have to close out this PR Monday morning and someone else will need to do the work if we haven't heard from you by then.

cortopy:
Hi @youngnick , thanks for letting me know about the status of this PR. Unfortunately, due to privacy concerns, I do not want to have an email address I can be contacted on listed in a public website.

I'll close this PR then

PR #2377, opened 2020-03-23
site: remove misleading comment in example deployment

jpeach:
Remove the misleading comment in the example deployment that guided
readers to a script that no longer exists. Update the generator script
to be completely self-contained.


jpeach:
xref: #2374

hack/generate-deployment.sh
```diff
@@ -6,8 +6,26 @@ set -o pipefail
 
 readonly HERE=$(cd $(dirname $0) && pwd)
 readonly REPO=$(cd ${HERE}/.. && pwd)
+readonly PROGNAME=$(basename $0)
+
+
+readonly TARGET="${REPO}/examples/render/contour.yaml"
+
+exec >$TARGET
```
youngnick: TIL about this. Neat.

PR #2379, opened 2020-03-24
internal: extract Envoy sort polices

### LINKED ISSUE #2376: Pull sorting methods into a helper package.
The fix for #2347 changes the named of the `RouteConfiguration` objects that Contour generates. This, in turn changes the sort order, which in validates the comparisons in `e2e/rds_test.go`. The basic problem here is that the tests are encoding a specific sort order, but that doesn't really matter (and tests should be insensitive to it).

So, if we hoist the sort methods into their own package, the tests can sort appropriately before comparing. Alternatively, we could switch over to `testify/assert` and use [ElementsMatch](https://godoc.org/github.com/stretchr/testify/assert#ElementsMatch) to ensure that the tests don't depend on sort order.

jpeach:
Extract the various sorting policies for Envoy types into a single
package that can be tested and conveniently consumed without causing
dependency cycles.

This fixes #2376.

### DIFF
--- a/internal/contour/cluster.go
+++ b/internal/contour/cluster.go
@@ -17,22 +17,24 @@ import (
 	"sort"
 	"sync"
 
-	envoy_api_v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
+	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
 	"github.com/envoyproxy/go-control-plane/pkg/cache"
 	"github.com/golang/protobuf/proto"
 	"github.com/projectcontour/contour/internal/dag"
 	"github.com/projectcontour/contour/internal/envoy"
+	"github.com/projectcontour/contour/internal/protobuf"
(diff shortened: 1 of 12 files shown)


jpeach:
> On Mar 25, 2020, at 2:30 AM, Steve Sloka <notifications@github.com> wrote:
> 
> Does sort need its own package? What else might go into this package in the future?

It doesn’t strictly need to go into a new package, but doing so does make it easier to consume since we have fewer worries about dependency cycles. Unlikely that we would put anything else into this package, given the naming.

stevesloka:
> Unlikely that we would put anything else into this package, given the naming.

Seems like we should rethink the package naming here, but it's merged, so I guess followup PR / discussion.

jpeach:
> On 27 Mar 2020, at 6:34 am, Steve Sloka <notifications@github.com> wrote:
> 
> Seems like we should rethink the package naming here, but it's merged, so I guess followup PR / discussion.

What was your concern about the package naming?

PR #2383, opened 2020-03-25
internal/contour: Add contour version to metrics

### LINKED ISSUE #2378: Expose Contour version in metrics
We have a number of internal dashboards that show information about our clusters. Cluster information includes CPU, memory and pod statistics as well as Kubernetes version. We would like to include the Contour version on this dashboard.

Today there is no easy way to scrape the Contour version without logging into the cluster as admin and query based on image version (dashboard and developers don't have cluster admin rights).

Request - expose the Contour version as part of the Prometheus metrics - https://projectcontour.io/guides/prometheus/#contour-metrics

pickledrick:
Fixes #2378

Signed-off-by: Peter Grant <pegrant@vmware.com>

Allows for passing version at build time and exposes build information as a metric.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,18 @@ RUN go mod download
 COPY cmd cmd
 COPY internal internal
 COPY apis apis
-RUN CGO_ENABLED=0 GOOS=linux GOFLAGS=-ldflags=-w go build -o /go/bin/contour -ldflags=-s -v github.com/projectcontour/contour/cmd/contour
+COPY Makefile Makefile
+
+ARG BUILD_BRANCH
+ARG BUILD_SHA
+ARG BUILD_VERSION
+
+RUN make install \
+	    CGO_ENABLED=0 \
+	    GOOS=linux \
+	    BUILD_VERSION=${BUILD_VERSION} \
+	    BUILD_SHA=${BUILD_SHA} \
+	    BUILD_BRANCH=${BUILD_BRANCH}
 
 FROM scratch AS final
 COPY --from=build /go/bin/contour /bin/contour
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,20 @@ VERSION ?= $(GIT_REF)
 # set outside this Makefile, as a safety valve.
 LATEST_VERSION ?= NOLATEST
 
+# Sets the current Git sha
+BUILD_SHA = $(shell git rev-parse --verify HEAD)
+# Sets the current branch
+BUILD_BRANCH = $(shell git branch --show-current)
+# Sets the current tagged git version
+BUILD_VERSION = $(VERSION)
+
+GO_BUILD_VARS = \
+	github.com/projectcontour/contour/internal/build.Version=${BUILD_VERSION} \
+	github.com/projectcontour/contour/internal/build.Sha=${BUILD_SHA} \
+	github.com/projectcontour/contour/internal/build.Branch=${BUILD_BRANCH}
+
 GO_TAGS := -tags "oidc gcp"
+GO_LDFLAGS := -s $(patsubst %,-X %, $(GO_BUILD_VARS))
 
 export GO111MODULE=on
 
@@ -39,7 +52,7 @@ check: install check-test check-test-race ## Install and run tests
 checkall: check lint check-generate
 
 install: ## Build and install the contour binary
-	go install -mod=readonly -v $(GO_TAGS) $(MODULE)/cmd/contour
+	go install -mod=readonly -v -ldflags="$(GO_LDFLAGS)" $(GO_TAGS) $(MODULE)/cmd/contour
 
 race:
 	go install -mod=readonly -v -race $(GO_TAGS) $(MODULE)/cmd/contour
@@ -48,7 +61,7 @@ download: ## Download Go modules
 	go mod download
 
 container: ## Build the Contour container image
-	docker build . -t $(IMAGE):$(VERSION)
+	docker build --build-arg BUILD_VERSION=$(BUILD_VERSION) --build-arg BUILD_BRANCH=$(BUILD_BRANCH) --build-arg BUILD_SHA=$(BUILD_SHA) . -t $(IMAGE):$(VERSION)
 
 push: ## Push the Contour container image to the Docker registry
 push: container
--- a/internal/build/version.go
+++ b/internal/build/version.go
@@ -1,0 +1,23 @@
+// Copyright © 2020 VMware
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package build
+
+// Branch allows for a queryable branch name set at build time.
+var Branch string
+
+// Sha allows for a queryable git sha set at build time.
+var Sha string
+
+// Version allows for a queryable version set at build time.
+var Version string
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -21,6 +21,7 @@ import (
 
 	"k8s.io/client-go/kubernetes"
 
+	"github.com/projectcontour/contour/internal/build"
 	"github.com/projectcontour/contour/internal/httpsvc"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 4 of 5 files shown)


pickledrick:
@jpeach I have not taken a look at metrics documentation auto-generation yet.

internal/metrics/metrics.go
```diff
@@ -230,6 +241,15 @@ func (m *Metrics) Zero() {
 	prometheus.NewTimer(m.CacheHandlerOnUpdateSummary).ObserveDuration()
 }
 
+// SetBuildInfo sets the Contour Build Info
+func (m *Metrics) SetBuildInfo(version string) {
```
pickledrick: Not particularly proud of this. If a more eloquent solution works for this I am open to it.

Dockerfile
```diff
@@ -8,7 +8,9 @@ RUN go mod download
 COPY cmd cmd
 COPY internal internal
 COPY apis apis
-RUN CGO_ENABLED=0 GOOS=linux GOFLAGS=-ldflags=-w go build -o /go/bin/contour -ldflags=-s -v github.com/projectcontour/contour/cmd/contour
+
+ARG CONTOUR_VERSION
+RUN CGO_ENABLED=0 GOOS=linux GOFLAGS=-ldflags=-w go build -ldflags="-s -X github.com/projectcontour/contour/internal/build.Version=${CONTOUR_VERSION}" -o /go/bin/contour -v github.com/projectcontour/contour/cmd/contour
```
jpeach: Conventionally, the build info is picked up in a Makefile at build time:

For example:
```
SHA := $(shell git rev-parse HEAD)
REVISION := $(shell git describe --tags --exact-match HEAD)

build: 
	@$(GO) build \
		-ldflags "-X $(REPO)/pkg/version.Revision=$(REVISION)" \
		-ldflags "-X $(REPO)/pkg/version.Sha=$(SHA)" \
		-o $(BIN) .
```

This lets you have exactly one place with the magic incantations. The variables from the makefile can be passed through the docker build, and then inside the docker file, just run "make build" (after adding the "build" target).

internal/metrics/metrics.go
jpeach: You don't need this at all. Just create the gauge in `NewMetrics`, and remove this.

internal/metrics/metrics.go
```diff
@@ -89,6 +91,13 @@ const (
 // to regenerate the metrics documentation.
 func NewMetrics(registry *prometheus.Registry) *Metrics {
 	m := Metrics{
+		buildInfoGuage: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: BuildInfoGuage,
+				Help: "Build information for Contour",
+			},
+			[]string{"version"},
```
jpeach: Check the blog post that @youngnick referenced in the issue. We should expose 'branch', 'goversion', 'revision', and 'version' (arguably goversion isn't useful, but it's best to be predictable and conventional with metrics).

internal/metrics/metrics.go
```diff
@@ -64,6 +65,7 @@ type Meta struct {
 }
 
 const (
+	BuildInfoGuage             = "contour_build_info"
```
jpeach: s/Guage/Gauge/

Here, and in other cases :)

Dockerfile
jpeach: Experiment with the git commands a bit; the version for release builds should always be a tag name, but for working trees, you probably want the short hash or something.

internal/metrics/metrics.go
pickledrick: Sure thing. Go version seems to already be instrumented through the standard go lib.

```
/ # curl -s contour.projectcontour:8000/metrics | grep go_info
# HELP go_info Information about the Go environment.
# TYPE go_info gauge
go_info{version="go1.13.8"} 1
```
I will experiment with git commands to get the desired labels.

pickledrick:
Have made some changes and this is how it's currently looking

```
/ # curl -s contour.projectcontour:8000/metrics | grep build
# HELP contour_build_info Build information for Contour
# TYPE contour_build_info gauge
contour_build_info{branch="metrics-version",revision="3ef9cbc4a72363c0cac409d030db9a1146b8ac2e",version="v1.3.0-21-g3ef9cbc4"} 1
```

Makefile
```diff
@@ -48,7 +52,7 @@ download: ## Download Go modules
 	go mod download
 
 container: ## Build the Contour container image
-	docker build . -t $(IMAGE):$(VERSION)
+	docker build --build-arg GIT_VERSION=$(VERSION) --build-arg GIT_BRANCH=$(BRANCH) --build-arg GIT_SHA=$(SHA) . -t $(IMAGE):$(VERSION)
```
jpeach: Currently `make install` links with `-s`, but the Dockerfile is intended to link with `-w -s`. It's not clear to be that this works as intended and there are upstream issues. That seem to indicate that `GOFLAGS` doesn't combind flags as intended.

Let's ensure that there is a single, canonical build through `make install`.

In the Makefile:
```

# Note we use `=` here to ensure make doesn't expand the build vars if they are already set, since git sin't in the build container.

# Sets the current Git sha
BUILD_SHA = $(shell git rev-parse --verify HEAD)
# Sets the current branch
BUILD_BRANCH = $(shell git branch --show-current)
# ...
BUILD_VERSION = $(VERSION)

GO_BUILD_VARS = \
	github.com/projectcontour/contour/internal/build.Version=${BUILD_VERSION} \
	github.com/projectcontour/contour/internal/build.Sha=${BUILD_SHA} \
	github.com/projectcontour/contour/internal/build.Branch=${BUILD_BRANCH}

GO_TAGS := -tags "oidc gcp"
GO_LDFLAGS := -s $(patsubst %,-X %, $(GO_BUILD_VARS))

install: ## Build and install the contour binary
	go install -mod=readonly -v -ldflags="$(GO_LDFLAGS)" $(GO_TAGS) $(MODULE)/cmd/contour

```

Then the docker file just calls the Makefile:

```
COPY Makefile Makefile

ARG BUILD_BRANCH
ARG BUILD_SHA
ARG BUILD_VERSION

RUN make install \
	    CGO_ENABLED=0 \
	    GOOS=linux \
	    BUILD_VERSION=${BUILD_VERSION} \
	    BUILD_SHA=${BUILD_SHA} \
	    BUILD_BRANCH=${BUILD_BRANCH}
```

Note that the Makefile and Dockerfile pass different link flags, I believe that `-s -w` is the intended combination, however.

internal/build/version.go
```diff
@@ -0,0 +1,23 @@
+// Copyright © 2019 VMware
```
jpeach: s/2019/2020/

internal/build/version.go
```diff
@@ -0,0 +1,23 @@
...
+
+package build
+
+// Branch allows for a queryable branch name set at build time
+var Branch string
+
+// Sha allows for a queryable git sha set at build time
+var Sha string
+
+// Version allows for a queryable version set at build time
```
jpeach: Can you please end these comments with fullstops?

internal/metrics/metrics.go
```diff
@@ -89,6 +92,13 @@ const (
 // to regenerate the metrics documentation.
 func NewMetrics(registry *prometheus.Registry) *Metrics {
 	m := Metrics{
+		buildInfoGauge: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: BuildInfoGauge,
+				Help: "Build information for Contour set at compile time. branch - the branch contour was built from, revision - the head git sha at the time contour was built, version - the tagged git version at the time Contour was built.",
```
jpeach: I tried a few different variations and trimmed the help down to this:
```
Build information for Contour. The labels are the branch and git SHA that Contour was built from, and the Contour version.
```

What do you think?

Makefile
```diff
@@ -28,7 +28,21 @@ VERSION ?= $(GIT_REF)
...
 LATEST_VERSION ?= NOLATEST
 
+# Sets the current Git sha
+BUILD_SHA = $(shell git rev-parse --verify HEAD)
+# Sets the current branch
+BUILD_BRANCH = $(shell git branch --show-current)
+# Sets the current tagged git version
+BUILD_VERSION = $(VERSION)
+
+
```
jpeach: Just one newline please.

jpeach:
@stevesloka Are you OK with using a separate package for the version information here?

internal/metrics/metrics.go
pickledrick: I am all for a shorter variation. Another potential one.

> Build information for Contour. Labels include the branch and git SHA that Contour was built from, and the Contour version.

jpeach: Yep, that sounds good 👍

stevesloka:
> Are you OK with using a separate package for the version information here?

I think the `build` package could be used for many things. My reference before on the `sort` package was I could hear @davecheney in my head asking me what the package was responsible for if it couldn't do a single thing and to think about that name. 

For this, I think it's fine.

PR #2386, opened 2020-03-26
Add Ingress status management

youngnick:
Updates #403.

This version of Ingress status adds the wiring to manage Ingress status, and allows the configuration of an Envoy service and namespace to watch for addresses to publish into the `status.Loadbalancer` stanza of Ingress objects.

There are still TODOs, #2387 and #2388 amongst them.

### DIFF
--- a/cmd/contour/ingressstatus.go
+++ b/cmd/contour/ingressstatus.go
@@ -1,0 +1,94 @@
+// Copyright © 2019 VMware
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"sync"
+
+	"github.com/projectcontour/contour/internal/k8s"
+	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+)
+
+// ingressStatusWriter manages the lifetime of StatusLoadBalancerUpdaters.
+//
+// The theory of operation of the ingressStatusWriter is as follows:
+// 1. On startup the ingressStatusWriter waits to be elected leader.
+// 2. Once elected leader, the ingressStatusWriter waits to receive a
+//    v1.LoadBalancerStatus value.
+// 3. Once a v1.LoadBalancerStatus value has been received, any existing informer
+//    is stopped and a new informer started in its place.
+// 4. Each informer is connected to a k8s.StatusLoadBalancerUpdater which reacts to
+//    OnAdd events for networking.k8s.io/ingress.v1beta1 objects. For each OnAdd
+//    the object is patched with the v1.LoadBalancerStatus value obtained on creation.
+//    OnUpdate and OnDelete events are ignored.If a new v1.LoadBalancerStatus value
+//    is been received, operation restarts at step 3.
+// 5. If the worker is stopped, any existing informer is stopped before the worker stops.
+type ingressStatusWriter struct {
+	log      logrus.FieldLogger
+	clients  *k8s.Clients
+	isLeader chan struct{}
+	lbStatus chan v1.LoadBalancerStatus
+}
+
+func (isw *ingressStatusWriter) Start(stop <-chan struct{}) error {
+
+	// await leadership election
+	isw.log.Info("awaiting leadership election")
+	select {
+	case <-stop:
+		// asked to stop before elected leader
+		return nil
+	case <-isw.isLeader:
+		isw.log.Info("elected leader")
+	}
+
+	var shutdown chan struct{}
+	var stopping sync.WaitGroup
+	for {
+		select {
+		case <-stop:
+			// stop existing informer and shut down
+			if shutdown != nil {
+				close(shutdown)
+			}
+			stopping.Wait()
+			return nil
+		case lbs := <-isw.lbStatus:
+			// stop existing informer
+			if shutdown != nil {
+				close(shutdown)
+			}
+			stopping.Wait()
+
+			// create informer for the new LoadBalancerStatus
+			factory := isw.clients.NewInformerFactory()
+			inf := factory.Networking().V1beta1().Ingresses().Informer()
+			log := isw.log.WithField("context", "IngressStatusLoadBalancerUpdater")
+			inf.AddEventHandler(&k8s.StatusLoadBalancerUpdater{
+				Client: isw.clients.ClientSet(),
+				Logger: log,
+				Status: lbs,
+			})
+
+			shutdown = make(chan struct{})
+			stopping.Add(1)
+			fn := startInformer(factory, log)
+			go func() {
+				defer stopping.Done()
+				fn(shutdown)
+			}()
+		}
+	}
+}
--- a/cmd/contour/ingressstatus.go
+++ b/cmd/contour/ingressstatus.go
@@ -1,4 +1,4 @@
-// Copyright © 2019 VMware
+// Copyright © 2020 VMware
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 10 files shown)


jpeach:
I'll review tomorrow.

cmd/contour/ingressstatus.go
```diff
@@ -0,0 +1,95 @@
...
+	select {
+	case <-stop:
+		// asked to stop before elected leader
+		return nil
+	case <-isw.isLeader:
+		isw.log.Info("elected leader")
+	}
+
+	var shutdown chan struct{}
+	var stopping sync.WaitGroup
```
jpeach: This wait group actually tracks whether the ingress informer is running.

internal/k8s/ingressstatus.go
```diff
@@ -0,0 +1,110 @@
...
+func (s *IngressStatusUpdater) OnDelete(obj interface{}) {
+	// we don't need to update the status on resources that
+	// have been deleted.
+}
+
+// ServiceStatusLoadBalancerWatcher implements ResourceEventHandler and
+// watches for changes to the status.loadbalancer field
+type ServiceStatusLoadBalancerWatcher struct {
+	ServiceName string
+	LBStatus    chan v1.LoadBalancerStatus
```
jpeach: We only need a send channel here, right?

internal/k8s/ingressstatus.go
```diff
@@ -0,0 +1,110 @@
...
+func (s *ServiceStatusLoadBalancerWatcher) OnAdd(obj interface{}) {
+	svc, ok := obj.(*v1.Service)
+	if !ok {
+		// not a service
+		return
+	}
+	if svc.Name != s.ServiceName {
+		return
+	}
+	s.notify(svc.Status.LoadBalancer)
```
jpeach: Nit: Is a method useful here? We could just
```Go
s.LBStatus <- svc.Status.LoadBalancer
```

jpeach: Maybe this could boil down to:
```Go
if svc, ok := obj.(*v1.Service); ok && svc.Name == s.ServiceName {
        s.LBStatus <- svc.Status.LoadBalancer
}
```

internal/k8s/ingressstatus_test.go
```diff
@@ -0,0 +1,161 @@
...
+	svc.Name = "potato"
+	sw.OnAdd(&svc)
+	_, ok = recv()
+	if ok {
+		t.Fatalf("expected no result when adding a service with a different name")
+	}
+
+	// assert adding a service with the correct name generates a notification
+	svc.Name = sw.ServiceName
+	svc.Status.LoadBalancer.Ingress = []v1.LoadBalancerIngress{{Hostname: "vmware.com"}}
```
jpeach: s/vmware.com/projectcontour.io/

Here, and below ...

cmd/contour/ingressstatus.go
```diff
@@ -0,0 +1,95 @@
...
+// 3. Once a v1.LoadBalancerStatus value has been received, any existing informer
+//    is stopped and a new informer started in its place. This ensures that all existing
+//    Ingress objects will have OnAdd events fired to the new event handler.
+// 4. Each informer is connected to a k8s.IngressStatusUpdater which reacts to
+//    OnAdd events for networking.k8s.io/ingress.v1beta1 objects. For each OnAdd
+//    the object is patched with the v1.LoadBalancerStatus value obtained on creation.
+//    OnUpdate and OnDelete events are ignored.If a new v1.LoadBalancerStatus value
+//    is been received, operation restarts at step 3.
+// 5. If the worker is stopped, any existing informer is stopped before the worker stops.
+type ingressStatusWriter struct {
```
jpeach: Nit: I think that calling this `loadBalancerStatusWriter` as an actor that writes `LoadBalancerStatuses` makes it clearer to my reading.

I went back and forth on this a bit, and I can see the perspective this name is coming from. Maybe we keep this name and think about renaming if it updates status on HTTPProxy too.

cmd/contour/ingressstatus.go
```diff
@@ -0,0 +1,95 @@
...
+			stopping.Wait()
+			return nil
+		case lbs := <-isw.lbStatus:
+			// stop existing informer
+			if shutdown != nil {
+				close(shutdown)
+			}
+			stopping.Wait()
+
+			// create informer for the new LoadBalancerStatus
```
jpeach: I think it is worth logging when then load balancer status changes. It is important, and should not be very frequent.

internal/k8s/ingressstatus.go
```diff
@@ -0,0 +1,110 @@
...
+
+// StatusLoadbalancerUpdater observes informer OnAdd events and
+// updates the ingress.status.loadBalancer field on all Ingress
+// objects that match the ingress class (if used).
+type IngressStatusUpdater struct {
+	Client clientset.Interface
+	Logger logrus.FieldLogger
+	Status v1.LoadBalancerStatus
+}
+
```
jpeach: Nit: consider asserting that this implements `cache.ResourceEventHandler`. 
```
var _ cache.ResourceEventHandler = &IngressStatusUpdater{}
```

cmd/contour/ingressstatus.go
youngnick: Yes, in my initial reading of this file it was called StatusLoadBalanceWriter or something, and it was hard to tell what it was actually acting on. The reason to move to ingressStatusWriter was to make it clear what objects were being written onto in the end.

loadBalancerStatusWriter is fine though, that covers us for HTTPProxy as well.

cmd/contour/ingressstatus.go
youngnick: Yep, makes sense.

internal/k8s/ingressstatus.go
youngnick: Leaving it bidirectional makes the testing easier - we can test what happens when pulling things out of the channel by just doing that, rather than something more complicated. I don't think being specific here is worth the additional test shenanigans.

internal/k8s/ingressstatus.go
youngnick: Agreed about the method, it's not really necessary for now, unless we end up wanting to do something more clever than just send down the channel.

For the code though, I feel like having individual clauses is more verbose, but far easier to read. For lack of a better word, I'd call it more idiomatic.

internal/k8s/ingressstatus.go
youngnick: I don't see what value this adds - if someone changes this struct to not implement that interface, then it will be a compile-time error.

youngnick:
Okay, ready for rereview.

internal/k8s/ingressstatus.go
jpeach: Fair enough. I started trying to get into the habit of doing this because it guarantees an error close to the definition and is robust against other code changes. We haven't been doing in in Contour, so not a big deal.

cmd/contour/ingressstatus.go
```diff
@@ -0,0 +1,99 @@
...
+			// Stop the existing informer.
+			if shutdown != nil {
+				close(shutdown)
+			}
+			ingressInformers.Wait()
+
+			isw.log.Info("Received a new address for status.loadBalancer")
+
+			// Create new informer for the new LoadBalancerStatus
+			factory := isw.clients.NewInformerFactory()
```
stevesloka: Should we reuse the informers created earlier? What if the informer would fail to create?

cmd/contour/servecontext.go
```diff
@@ -165,6 +173,8 @@ func newServeContext() *serveContext {
 			Name:          "leader-elect",
 		},
 		UseExperimentalServiceAPITypes: false,
+		envoyServiceName:               "envoy",
+		envoyServiceNamespace:          "projectcontour",
```
stevesloka: We should consume the downward env variable for this so we don't keep the hard coded default.

cmd/contour/ingressstatus.go
youngnick: The reason to throw away the old informer is to ensure that the new informer does a full resync, so that all the existing Ingress objects that are in scope will get updated (via OnAdd calls).

I think that if the informer failed to create, there would be a panic somewhere, because this method doesn't return an error for us to check.

cmd/contour/servecontext.go
youngnick: Good catch, I'll set that up.

PR #2389, opened 2020-03-26
Contour namespace

### LINKED ISSUE #2255: Contour should be better about running in alternate namespaces

I'd like to be able to install Contour into multiple namespaces A, B, C relatively easily, but historically this has been a game of whack-a-mole to try and find all of the places to change to run elsewhere.

The two places that come to mind that burned me are:
1. leader election [config default](https://github.com/projectcontour/contour/blob/c3646992f716dba1a3c9d0bdcbe4f2c7878e32a8/examples/contour/01-contour-config.yaml#L29) (at least at HEAD this uses the string "projectcontour")
2. the `--namespace` flag to the certgen job, which I had to dig through code to find because the flag isn't present in the examples.

Where possible, it's generally helpful to use the downward API to load smart defaults (this has helped us in Knative with namespace portability).

stevesloka:
Fixes #2255 by using the downward api to give the default namespace for configuration variables. They are still able to be overridden/configured as before but are no longer hardcoded to `projectcontour`.

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -44,7 +44,7 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 
 	// The precedence of configuration for contour serve is as follows:
 	// config file, overridden by env vars, overridden by cli flags.
-	// however, as -c is a cli flag, we don't know its valye til cli flags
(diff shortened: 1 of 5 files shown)


stevesloka:
@jpeach update message? Because of the squash? Bah, this is why I have separate commits to make it clear the changes. I guess I can just rebase them together.

PR #2394, opened 2020-03-27
internal/dag: only process ingress objects that match ingress.class 

### LINKED ISSUE #2199: When started with --ingress-class-name flag, contour still considers IngressRoutes objects without a matching ingress.class annotation

My contour deployment is started with a ` --ingress-class-name=test` flag. It however still feeds envoy configuration for some `IngressRoute` objects which don't have an annotation. 

Those objects are the one currently used by the "old" (current production deployment) contour. They don't have `*/ingress.class` annotations. 

Example (stripped from personnal infos / kubernetes runtime infos)

```
kind: IngressRoute
metadata:
  name: xxx.com
spec:
  routes:
  - match: /
    services:
    - name: xxx-some-backend
      port: xxx
  virtualhost:
    fqdn: xxx.com
    tls:
      secretName: xxxSSL
status:
  currentStatus: valid
  description: valid IngressRoute
```

Envoy config excerpt :

```
        "name": "xxx.com",
        "domains": [
         "xxx.com",
         "xxx.com:*"
        ],
        "routes": [
         {
          "match": {
           "prefix": "/"
          },
          "route": {
           "cluster": "namespace/xxx-some-backend/xxx/da39a3ee5e"
          }
         }
        ]
```
 

I'm in the process of upgrading contour from a <= 0.13.0 (I know ... :)) version and it's the opportunity to migrate `IngressRoute` objects to `HTTPProxy` at the same time. 
I'm not sure about potential side-effects with having both objects in // for the same virtualhosts/fqdn so I would like my new contour deployment to only consider `HTTPProxy` objects and ignore my current `IngressRoute` ones.

**Environment:**
- Contour version: v1.1.0, official docker image
- Kubernetes version: v1.12.10-gke.17
- OS (e.g. from `/etc/os-release`): cos

stevesloka:
Fixes #2199  

Only process ingress objects that match an ingress class when specified in Contour explicitly.

```
RELEASE NOTE: This is a change that needs to be called out in the release notes!
```

### DIFF
--- a/internal/dag/cache.go
+++ b/internal/dag/cache.go
@@ -72,28 +72,29 @@ func toMeta(obj Object) Meta {
 // belongs to the Ingress class that this cache is using.
 func (kc *KubernetesCache) matchesIngressClass(obj Object) bool {
 	objectClass := ingressClass(obj)
-	targetClass := stringOrDefault(kc.IngressClass, DEFAULT_INGRESS_CLASS)
 
 	switch objectClass {
-	// Unspecified ingress class always matches.
-	case "":
+	case kc.IngressClass:
+		// Handles kc.IngressClass == "" and kc.IngressClass == "custom".
 		return true
-		// Specifying our ingress class also matches.
-	case targetClass:
-		return true
-	// Any other ingress class fails to match.
-	default:
-		kind := k8s.KindOf(obj)
-		om := obj.GetObjectMeta()
-
-		kc.WithField("name", om.GetName()).
-			WithField("namespace", om.GetNamespace()).
-			WithField("kind", kind).
-			WithField("ingress.class", objectClass).
-			Debug("ignoring object with unmatched ingress class")
-
-		return false
+	case DEFAULT_INGRESS_CLASS:
+		// kc.IngressClass == "" implicitly matches the default too.
+		if kc.IngressClass == "" {
+			return true
+		}
 	}
+
+	// Any other ingress class fails to match.
+	kind := k8s.KindOf(obj)
+	om := obj.GetObjectMeta()
+
+	kc.WithField("name", om.GetName()).
+		WithField("namespace", om.GetNamespace()).
+		WithField("kind", kind).
+		WithField("ingress.class", objectClass).
+		Debug("ignoring object with unmatched ingress class")
+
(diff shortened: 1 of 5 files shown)


internal/dag/cache.go
```diff
@@ -74,6 +74,27 @@ func (kc *KubernetesCache) matchesIngressClass(obj Object) bool {
...
+				WithField("namespace", om.GetNamespace()).
+				WithField("kind", kind).
+				WithField("ingress.class", objectClass).
+				Debug("ignoring object with unmatched ingress class")
+
+			return false
+		}
+	}
+
+	// User defined class not specified
```
jpeach: I think that we can simplify the logic to this:

```Go
       objectClass := ingressClass(obj)

        switch objectClass {
        case kc.IngressClass:
                // Handles kc.IngressClass == "" and kc.IngressClass == "custom".
                return true
        case DEFAULT_INGRESS_CLASS:
                // kc.IngressClass == "" implicitly matches the default too.
                if kc.IngressClass == "" {
                        return true
                }
        }

        // Any other ingress class fails to match.
        kind := k8s.KindOf(obj)
        om := obj.GetObjectMeta()

        kc.WithField("name", om.GetName()).
                WithField("namespace", om.GetNamespace()).
                WithField("kind", kind).
                WithField("ingress.class", objectClass).
                Debug("ignoring object with unmatched ingress class")

        return false
```

If there's an exact match, then we have a match. The only other case we need to worry about is when Contour is running with an empty ingress class and we want to implicitly match the default.

youngnick: I agree with James here, this is a much clearer function to read, and it means we don't make the unneeded targetClass variable.

From what I can see, the available states are:
- [x] no configured ingress.class, none on object - pass
- [x] no configured ingress.class, "contour" on object - pass
- [x] no configured ingress.class, anything else on object - fail
- [x] configured ingress.class, none on object - fail
- [x] configured ingress.class, correct on object
- [x] configured ingress.class, incorrect on object

I don't think we currently do enough testing of the "no configured ingress class" cases.

stevesloka:
@youngnick @jpeach all fixed up! =)

PR #2398, opened 2020-03-30
build: omit the DWARF symbol table

jpeach:
When Dockerfile build moved across the the Makefile, the `-w` Go linker
option was dropped. This option reduces the final binary size by 10MiB
on macOS, so it's probably worth keeping.


jpeach:
xref #2383

PR #2399, opened 2020-03-30
internal/build: Add contour version as a CLI option

### LINKED ISSUE #2396: Add a "version" CLI subcommand.

The Contour build information is exposed as Prometheus metrics via #2378. We should also expose these information in a `version` subcommand so that it is easy for developers and operators to verify build artifacts.

pickledrick:
Fixes #2396

Signed-off-by: Peter Grant <pegrant@vmware.com>

Allows for a CLI option to print build and version information.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,8 +20,8 @@ TAG_LATEST ?= false
 # LOCALIP as an env var before running 'make local' will solve that.
 LOCALIP ?= $(shell ifconfig | grep inet | grep -v '::' | grep -v 127.0.0.1 | head -n1 | awk '{print $$2}')
 
-# Sets GIT_REF to a tag if it's present, otherwise the short rev.
-GIT_REF = $(shell git describe --tags || git rev-parse --short=8 --verify HEAD)
+# Sets GIT_REF to a tag if it's present, otherwise the short git sha will be used.
+GIT_REF = $(shell git describe --tags --exact-match 2>/dev/null || git rev-parse --short=8 --verify HEAD)
 VERSION ?= $(GIT_REF)
 # Used for the tag-latest action.
 # The tag-latest action will be a noop unless this is explicitly
--- a/cmd/contour/contour.go
+++ b/cmd/contour/contour.go
@@ -18,6 +18,7 @@ import (
 	"os"
 
 	"github.com/envoyproxy/go-control-plane/pkg/cache"
+	"github.com/projectcontour/contour/internal/build"
 	"github.com/sirupsen/logrus"
 	kingpin "gopkg.in/alecthomas/kingpin.v2"
 	"k8s.io/klog"
@@ -61,6 +62,7 @@ func main() {
 	sds.Arg("resources", "SDS resource filter").StringsVar(&resources)
 
 	serve, serveCtx := registerServe(app)
+	version := app.Command("version", "Build information for Contour.")
 
 	args := os.Args[1:]
 	switch kingpin.MustParse(app.Parse(args)) {
@@ -95,10 +97,13 @@ func main() {
 		}
 		log.Infof("args: %v", args)
 		check(doServe(log, serveCtx))
+	case version.FullCommand():
+		println(build.PrintBuildInfo())
 	default:
 		app.Usage(args)
 		os.Exit(2)
 	}
+
 }
 
 func check(err error) {
--- a/internal/build/version.go
+++ b/internal/build/version.go
@@ -13,6 +13,17 @@
 
 package build
 
+import (
+	"gopkg.in/yaml.v2"
+)
+
+// BuildInfo is a struct for build information.
+type BuildInfo struct {
+	Branch  string `yaml:"branch,omitempty"`
+	Sha     string `yaml:"sha,omitempty"`
+	Version string `yaml:"version,omitempty"`
+}
+
 // Branch allows for a queryable branch name set at build time.
 var Branch string
 
@@ -21,3 +32,13 @@ var Sha string
 
 // Version allows for a queryable version set at build time.
 var Version string
+
+// PrintBuildInfo prints the build information.
+func PrintBuildInfo() string {
+	buildInfo := &BuildInfo{Branch, Sha, Version}
+	out, err := yaml.Marshal(buildInfo)
+	if err != nil {
+		panic(err)
+	}
+	return string(out)
+}


pickledrick:
Sample output:

```
> $ contour version                                                                                                                                                                                                          [±version-cmd ●]
{"branch":"version-cmd","sha":"51348913a93d32b0adae970110a850c681e51d3d","version":"v1.2.0-90-g51348913"}
```

internal/build/version.go
```diff
@@ -21,3 +32,13 @@ var Sha string
 
 // Version allows for a queryable version set at build time.
 var Version string
+
+// PrintBuildInfo prints the build information.
+func PrintBuildInfo() string {
+	buildInfo := &BuildInfo{Branch, Sha, Version}
+	out, err := json.Marshal(buildInfo)
```
jpeach: Suggest that we do yaml (or manual marshaling) here?

```
$ contour version
branch: version-cmd
sha: 51348913a93d32b0adae970110a850c681e51d3d
version: v1.2.0-90-g51348913
```

There reason being that there are no other Contour commands that emit JSON, and the YAML version is easier to crack with grep, awk, etc. whereas the JSON basically needs jq.

cmd/contour/contour.go
```diff
@@ -61,6 +62,7 @@ func main() {
 	sds.Arg("resources", "SDS resource filter").StringsVar(&resources)
 
 	serve, serveCtx := registerServe(app)
+	version := app.Command("version", "Build information for Contour. Contains the branch and git SHA that Contour was built from, and the Contour version.")
```
jpeach: Shall we keep this short, just "Build information for Contour."

The difference being that you can just run it and see the output, which is not the case with the metric (you have to get it to serve, etc).

internal/build/version.go
jpeach: @youngnick LMK what you think?

pickledrick: Have updated to YAML

cmd/contour/contour.go
pickledrick: Makes sense and keeps to a similar style for flags.

pickledrick: Have updated to a shorter description.

pickledrick:
Have updated output to be YAML for readability.

```
> $ contour version                                                                                                                                                                                                          [±version-cmd ●]
branch: version-cmd
sha: 06b0d669b9531395f602e6de48081f04f50b4626
version: v1.2.0-91-g06b0d669
```

internal/build/version.go
youngnick: Yeah, YAML is good.

pickledrick:
@stevesloka given this, how about something like the following. Will have to test with make etc.

```
> $ git checkout -                                                                                                                                                                                                           [±release-1.3 ●]
Switched to branch 'version-cmd'
> $ git describe --tags --exact-match 2>/dev/null || git rev-parse --short=8 --verify HEAD                                                                                                                                   [±version-cmd ●]
ae38c692
> $ git checkout release-1.3                                                                                                                                                                                                 [±version-cmd ●]
Switched to branch 'release-1.3'
Your branch is up to date with 'upstream/release-1.3'.
> $ git describe --tags --exact-match 2>/dev/null || git rev-parse --short=8 --verify HEAD                                                                                                                                   [±release-1.3 ●]
v1.3.0
> $
```

stevesloka:
@pickledrick looks like that works. I just didn't think it was proper to have a non-release branch report a released version.

pickledrick:
> @pickledrick looks like that works. I just didn't think it was proper to have a non-release branch report a released version.

Makes sense. I have updated to this now.

Makefile
```diff
@@ -21,7 +21,7 @@ TAG_LATEST ?= false
 LOCALIP ?= $(shell ifconfig | grep inet | grep -v '::' | grep -v 127.0.0.1 | head -n1 | awk '{print $$2}')
 
 # Sets GIT_REF to a tag if it's present, otherwise the short rev.
-GIT_REF = $(shell git describe --tags || git rev-parse --short=8 --verify HEAD)
+GIT_REF = $(shell describe --tags --exact-match 2>/dev/null || git rev-parse --short=8 --verify HEAD)
```
youngnick: what happened to `git describe`? `--exact-match` seems good, but I don't think just `describe` will do anything?

jpeach:
> @pickledrick looks like that works. I just didn't think it was proper to have a non-release branch report a released version.

I think that the way to read `v1.3.0-24-gae38c692` is that this 24 commits passed the nearest tag, which is `v1.3.0-24` and the short hash is `ae38c692`. GitHub will actually serve the source tarball from this commit name, which surprised me.

https://git-scm.com/docs/git-describe#_examples

That said, I don't have too strong an opinion on what we want to show here when it's a non-release build.

stevesloka:
To me, a build from the `master` branch has no relation to v1.3.0 and is confusing. Just throwing in my 2 cents.

jpeach:
```
$ git describe --all --long
heads/master-0-g7696eb2
$ basename $(git describe --all --long)
master-0-g7696eb2
```

WDYT?

Makefile
pickledrick: my bad. Copy/Paste

pickledrick:
> master-0-g7696eb2

I like this ^^. It's descriptive if not a tagged branch, and is quickly useful if you are not looking at a combination of branch/revision information with it.

`git describe --tags --exact-match 2>/dev/null || basename $(git describe --all --long)`

jpeach:
> > master-0-g7696eb2
> 
> I like this ^^. It's descriptive if not a tagged branch, and is quickly useful if you are not looking at a combination of branch/revision information with it.
> 
> `git describe --tags --exact-match 2>/dev/null || basename $(git describe --all --long)`

Leave a long block comment explaining this :)

stevesloka:
Why did this all change? Prev it was the git tag or the git ref if not on a tag.

jpeach:
> Why did this all change? Prev it was the git tag or the git ref if not on a tag.

It didn't change, it's just that now the result is visible. The build has always done this:

```
GIT_REF = $(shell git describe --tags || git rev-parse --short=8 --verify HEAD)
VERSION ?= $(GIT_REF)
```

Which gives me this result on master:
```
$ git describe --tags || git rev-parse --short=8 --verify HEAD
v1.2.0-92-gd4fb68d
```

(which is just `git describe --tags`)

Makefile
```diff
@@ -21,7 +21,7 @@ TAG_LATEST ?= false
 LOCALIP ?= $(shell ifconfig | grep inet | grep -v '::' | grep -v 127.0.0.1 | head -n1 | awk '{print $$2}')
 
 # Sets GIT_REF to a tag if it's present, otherwise the short rev.
-GIT_REF = $(shell git describe --tags || git rev-parse --short=8 --verify HEAD)
+GIT_REF = $(shell git describe --tags --exact-match 2>/dev/null || git rev-parse --short=8 --verify HEAD)
```
jpeach, line 24: As per the main discussion, lets update this to:
```
git describe --tags --exact-match 2>/dev/null || basename $(git describe --all --long)
```

Please update the comment too.

internal/build/version.go
```diff
@@ -13,6 +13,17 @@
 
 package build
 
+import (
+	"gopkg.in/yaml.v2"
+)
+
+// BuildInfo is a struct for build information
```
jpeach, line 20: End the comment with a full stop please :)

Makefile
```diff
@@ -20,8 +20,8 @@ TAG_LATEST ?= false
 # LOCALIP as an env var before running 'make local' will solve that.
 LOCALIP ?= $(shell ifconfig | grep inet | grep -v '::' | grep -v 127.0.0.1 | head -n1 | awk '{print $$2}')
 
-# Sets GIT_REF to a tag if it's present, otherwise the short rev.
-GIT_REF = $(shell git describe --tags || git rev-parse --short=8 --verify HEAD)
+# Sets GIT_REF to a tag if it's present, otherwise the basename of a git describe is used.
+GIT_REF = $(shell git describe --tags --exact-match 2>/dev/null || basename $(shell git describe --all --long))
```
jpeach, line 24: As per @stevesloka , let's revert the `git describe` to `git rev-parse --short=8 --verify HEAD`. Since you added the `--exact-match` flag, we will hit this case now, and the `GIT_BRANCH` variable takes care of the branch field of the long description.

PR #2406, opened 2020-03-31
internal/envoy: enable merge_slashes on the httpconnection manager for all listeners

### LINKED ISSUE #2344: merge duplicate slashes in paths

We have come across an issue while testing contour.
We have found some urls which are not being routed with contour but are with nginx
For example, a route defined in the ingress object of a service:

/example/route/service/

If the client makes this call:

https://somehost.com/example/route//service/library.json

on contour this returns 404 but on nginx it resolves

If we remove the double forward slash it works:

https://somehost.com/example/route/service/library.json

Is this handling around forward slashes a known issue in contour?

In this envoy documentation https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/network/http_connection_manager/v2/http_connection_manager.proto
I can see a 'merge_slashes' configuration option. Is this something which is possible to expose in contour?

I expected contour to merge the duplicate slashes.

**Environment:**

- Contour version: 1.1
- Kubernetes version: (use `kubectl version`): 1.16

stevesloka:
Fixes #2344, by enabling `merge_slashes` in the connection manager for each listener so that requests with multiple slashes are merged and processed properly. 


stevesloka:
This needs to be flagged for an e2e test.

PR #2407, opened 2020-04-01
internal/health: separate health and metrics services

### LINKED ISSUE #2380: Allow separate listeners for metrics and healthchecks
We understand that Envoy will listen on 8002, both /ready and /stats/* endpoints get presented through here. It is /stats (and everything else under /stats/*) that we need to protect, but not /ready.

Our Contour/Envoy configuration requires /ready must be HTTP and not HTTPS due to https://github.com/projectcontour/contour/issues/1503, and it has to be unprotected so load balancers can query it.

We looked at the Prometheus.io annotations in the Envoy deployment spec, but this does not seem to do anything.

Tried to configure the Prometheus stats address/port as command line arguments to contour. Specifically, something like:

--stats-address=0.0.0.0 --stats-port=8003

Unfortunately the flag did not work. /ready and /stats are stilled tied together after bouncing Contour and Envoy, and I believe this to be hard-coded.

If we look up "statsPort", we can find it here: https://github.com/projectcontour/contour/blob/v1.2.1/cmd/contour/servecontext.go#L121

We see it used here, creating a new listener: https://github.com/projectcontour/contour/blob/v1.2.1/cmd/contour/serve.go#L184

Which is this function: https://github.com/projectcontour/contour/blob/v1.2.1/internal/contour/listener.go#L213

And finally we see where our listener payload is generated: https://github.com/projectcontour/contour/blob/v1.2.1/internal/envoy/stats.go#L27

It seems /ready and /stats are tied together, and we need these to be separated. I believe we are looking towards a feature request.

pickledrick:
Fixes #2380

Signed-off-by: Peter Grant pegrant@vmware.com

Allows for separating health/metric services.

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -16,6 +16,7 @@ package main
 import (
 	"fmt"
 	"net"
+	"net/http"
 	"os"
 	"os/signal"
 	"strconv"
@@ -26,6 +27,7 @@ import (
 	"github.com/projectcontour/contour/internal/dag"
 	"github.com/projectcontour/contour/internal/debug"
 	cgrpc "github.com/projectcontour/contour/internal/grpc"
+	"github.com/projectcontour/contour/internal/health"
 	"github.com/projectcontour/contour/internal/httpsvc"
 	"github.com/projectcontour/contour/internal/k8s"
 	"github.com/projectcontour/contour/internal/metrics"
@@ -85,8 +87,10 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 	serve.Flag("debug-http-address", "Address the debug http endpoint will bind to.").StringVar(&ctx.debugAddr)
 	serve.Flag("debug-http-port", "Port the debug http endpoint will bind to.").IntVar(&ctx.debugPort)
 
-	serve.Flag("http-address", "Address the metrics http endpoint will bind to.").StringVar(&ctx.metricsAddr)
-	serve.Flag("http-port", "Port the metrics http endpoint will bind to.").IntVar(&ctx.metricsPort)
+	serve.Flag("http-address", "Address the metrics HTTP endpoint will bind to.").StringVar(&ctx.metricsAddr)
+	serve.Flag("http-port", "Port the metrics HTTP endpoint will bind to.").IntVar(&ctx.metricsPort)
+	serve.Flag("health-address", "Address the health HTTP endpoint will bind to.").StringVar(&ctx.healthAddr)
+	serve.Flag("health-port", "Port the health HTTP endpoint will bind to.").IntVar(&ctx.healthPort)
 
 	serve.Flag("contour-cafile", "CA bundle file name for serving gRPC with TLS.").Envar("CONTOUR_CAFILE").StringVar(&ctx.caFile)
 	serve.Flag("contour-cert-file", "Contour certificate file name for serving gRPC over TLS.").Envar("CONTOUR_CERT_FILE").StringVar(&ctx.contourCert)
@@ -240,18 +244,39 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	g.Add(eventHandler.Start())
 
 	// step 8. create metrics service and register with workgroup.
-	metricsvc := metrics.Service{
-		Service: httpsvc.Service{
-			Addr:        ctx.metricsAddr,
-			Port:        ctx.metricsPort,
-			FieldLogger: log.WithField("context", "metricsvc"),
-		},
-		Client:   clients.ClientSet(),
-		Registry: registry,
+	metricsvc := httpsvc.Service{
+		Addr:        ctx.metricsAddr,
+		Port:        ctx.metricsPort,
+		FieldLogger: log.WithField("context", "metricsvc"),
+		ServeMux:    http.ServeMux{},
 	}
+
+	metricsvc.ServeMux.Handle("/metrics", metrics.Handler(registry))
+
+	if ctx.healthAddr == ctx.metricsAddr && ctx.healthPort == ctx.metricsPort {
+		h := health.Handler(clients.ClientSet())
+		metricsvc.ServeMux.Handle("/health", h)
+		metricsvc.ServeMux.Handle("/healthz", h)
+	}
+
 	g.Add(metricsvc.Start)
 
-	// step 9. create debug service and register with workgroup.
+	// step 9. create a separate health service if required.
+	if ctx.healthAddr != ctx.metricsAddr || ctx.healthPort != ctx.metricsPort {
+		healthsvc := httpsvc.Service{
+			Addr:        ctx.healthAddr,
+			Port:        ctx.healthPort,
+			FieldLogger: log.WithField("context", "healthsvc"),
+		}
+
+		h := health.Handler(clients.ClientSet())
+		healthsvc.ServeMux.Handle("/health", h)
+		healthsvc.ServeMux.Handle("/healthz", h)
+
+		g.Add(healthsvc.Start)
+	}
+
+	// step 10. create debug service and register with workgroup.
 	debugsvc := debug.Service{
 		Service: httpsvc.Service{
 			Addr:        ctx.debugAddr,
@@ -262,7 +287,7 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	}
 	g.Add(debugsvc.Start)
 
-	// step 10. register leadership election
+	// step 11. register leadership election.
 	eventHandler.IsLeader = setupLeadershipElection(&g, log, ctx, clients, eventHandler.UpdateNow)
 
 	// step 12. create grpc handler and register with workgroup.
--- a/design/listener-design.md
+++ b/design/listener-design.md
@@ -1,0 +1,22 @@
+# Listener Design
+
+**Status**: _Draft_
+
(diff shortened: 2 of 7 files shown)


pickledrick:
Hi @jpeach 

There seems to be some confusion as things currently are not named or mapped very well.

> Currently, http-address/http-port specifies where to tell Envoy
> to start its HTTP listener, so we don't need to change that. The
> stats-address/stats-port flags specify where the metrics and health
> servers listen.

`http-address/http-port` seem to only be used for Contours metrics. Envoy listeners are configured using `envoy-service-http-address/envoy-service-http-port`

My proposal is to retain the `http-address/http-port` as flags to maintain backwards compatibility. Should an end-user wish to separate health-checking from metrics. They are then able to pass `health-port` which will then create a separate listener using `<http-address>:<health-port>` to avoid port conflict.

An example of how this currently looks using defaults

```
> $ contour serve
> $ curl -s localhost:8000/metrics | grep contour_build_info                                                                                                                                                                          [±master ●]
# HELP contour_build_info Build information for Contour. Labels include the branch and git SHA that Contour was built from, and the Contour version.
# TYPE contour_build_info gauge
contour_build_info{branch="separate-ports",revision="8bacb1007a84050e6f8a15dd4a32440940bf5742",version="v1.2.0-93-g8bacb100"} 1
> $ curl localhost:8000/health                                                                                                                                                                                                        [±master ●]
OK
> $ curl localhost:8000/healthz                                                                                                                                                                                                       [±master ●]
OK
```

and then specifying a `--health-port` flag to the `serve` command

```
> $ contour serve --health-port 9999
> $
> $ curl -s localhost:8000/health
404 page not found
> $ curl -s localhost:8000/healthz
404 page not found
> $ curl -s localhost:8000/metrics | grep contour_build_info
# HELP contour_build_info Build information for Contour. Labels include the branch and git SHA that Contour was built from, and the Contour version.
# TYPE contour_build_info gauge
contour_build_info{branch="separate-ports",revision="8bacb1007a84050e6f8a15dd4a32440940bf5742",version="v1.2.0-93-g8bacb100"} 1
> $ curl -s localhost:9999/health
OK
> $ curl -s localhost:9999/healthz
OK
> $ curl -s localhost:9999/metrics | grep contour_build_info
> $ curl -I localhost:9999/metrics
HTTP/1.1 404 Not Found
Content-Type: text/plain; charset=utf-8
X-Content-Type-Options: nosniff
Date: Thu, 02 Apr 2020 00:28:31 GMT
Content-Length: 19

> $
```

> Unfortunately, there's not really a good place to document this, but I
> definitely think that it should be documented.
Agreed. I have added `site/_guides/health-checking.md` and updated `site/_guides/prometheus.md` to reflect this. Let me know if this is autogenerated and I'm not doing it right. Or if there is a better place for this.

> Finally, we should discuss whether we want to update the deployment YAML
> to configure the services separately. I'm not sure what the trade-off
> here is, but maybe we should enable this by default?
 I think not updating the deployment YAML at this stage is ok (as to not disrupt what people expect to see) and just providing a little extra doc in the guides that this is possible will work.

> Finally finally, take a look at the contributing doc and the link I
> send you on Slack about git commit messages. The former has the standard
> outline that you should try to follow for Contour.
Thanks, I will update everything on this PR and my other one after I have reviewed

jpeach:
> Hi @jpeach
> 
> There seems to be some confusion as things currently are not named or mapped very well.
> 
> > Currently, http-address/http-port specifies where to tell Envoy
> > to start its HTTP listener, so we don't need to change that. The
> > stats-address/stats-port flags specify where the metrics and health
> > servers listen.
> 
> `http-address/http-port` seem to only be used for Contours metrics. Envoy listeners are configured using `envoy-service-http-address/envoy-service-http-port`

Oh I see. Yes, I got it wrong there :-/

> My proposal is to retain the `http-address/http-port` as flags to maintain backwards compatibility. Should an end-user wish to separate health-checking from metrics. They are then able to pass `health-port` which will then create a separate listener using `<http-address>:<health-port>` to avoid port conflict.
> 
> An example of how this currently looks using defaults
> 
> ```
> > $ contour serve
> > $ curl -s localhost:8000/metrics | grep contour_build_info                                                                                                                                                                          [±master ●]
> # HELP contour_build_info Build information for Contour. Labels include the branch and git SHA that Contour was built from, and the Contour version.
> # TYPE contour_build_info gauge
> contour_build_info{branch="separate-ports",revision="8bacb1007a84050e6f8a15dd4a32440940bf5742",version="v1.2.0-93-g8bacb100"} 1
> > $ curl localhost:8000/health                                                                                                                                                                                                        [±master ●]

In this case, I was expecting that calling [ListenAndServe](https://golang.org/pkg/net/http/#ListenAndServe) would cause an error when the second server tried to bind the same address. If Go makes that work, then you are right, and we don't have to do any refactoring.

jpeach:
@pickledrick I took a better look at this, and I think that the approach we should use is to decouple metrics and health from `httpsvc.Service`. We can do this by introducing `Handler` exports for both packages. This function returns a `http.Handler` that can be directly attached to the service.

The result would look something like this:
```Go
	metricsvc := httpsvc.Service{
		Addr:        ctx.metricsAddr,
		Port:        ctx.metricsPort,
		FieldLogger: log.WithField("context", "metricsvc"),
		ServeMux:    http.ServeMux{},
	}

	metricsvc.ServeMux.Handle("/metrics", metrics.Handler(registry))

	if ctx.healthAddr == ctx.metricsAddr && ctx.healthPort == ctx.metricsPort {
		h := health.Handler(clients.ClientSet())
		metricsvc.ServeMux.Handle("/health", h)
		metricsvc.ServeMux.Handle("/healthz", h)
	}

	g.Add(metricsvc.Start)

	if ctx.healthAddr != ctx.metricsAddr || ctx.healthPort != ctx.metricsPort {
		healthsvc := httpsvc.Service{
			Addr:        ctx.healthAddr,
			Port:        ctx.healthPort,
			FieldLogger: log.WithField("context", "healthsvc"),
		}

		h := health.Handler(clients.ClientSet())
		healthsvc.ServeMux.Handle("/health", h)
		healthsvc.ServeMux.Handle("/healthz", h)

		g.Add(healthsvc.Start)
	}
```

This approach keeps the health and metrics packages separate, and also decouples them from how we set up HTTP serving. One issue that needs a bit of consideration is what log fields the HTTP service should use (since context could not be ambiguous).

Note that we should add `--health-address` for consistency with other address specifications.

pickledrick:
Thanks @jpeach 

wrt
> One issue that needs a bit of consideration is what log fields the HTTP service should use (since context could not be ambiguous).

Just clarifying if you are referring to the health service or the metrics service?

site/_guides/health-checking.md
```diff
@@ -0,0 +1,14 @@
+---
+title: Health Checking
+layout: page
+---
+
+<div id="toc" class="navigation"></div>
```
jpeach, line 6: Did you use the `make site-devel` target to preview the docs? You should not need a TOC here, since there are no header elements in the page.

site/_guides/prometheus.md
```diff
@@ -21,8 +21,11 @@ port `8002`.
 
 ## Contour Metrics
 
-Contour exposes a Prometheus-compatible `/metrics` endpoint on port `8000` with
-the following metrics:
+Contour exposes a Prometheus-compatible `/metrics` endpoint defaulted to `8000` this is configurable using the `--http-address` and `--http-port` flags for the `serve` command.
```
jpeach, line 24: "that defaults to listening on port `8000`. This can be configured by using ..."

cmd/contour/serve.go
```diff
@@ -87,6 +89,8 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 
 	serve.Flag("http-address", "Address the metrics http endpoint will bind to.").StringVar(&ctx.metricsAddr)
 	serve.Flag("http-port", "Port the metrics http endpoint will bind to.").IntVar(&ctx.metricsPort)
+	serve.Flag("health-address", "Address the health http endpoint will bind to.").StringVar(&ctx.healthAddr)
+	serve.Flag("health-port", "Port the health http endpoint will bind to.").IntVar(&ctx.healthPort)
```
jpeach, line 93: I don't know why HTTP wasn't capitalized originally, can you please capitalize it for both flag sets?

design/listener-design.md
```diff
@@ -0,0 +1,22 @@
+# Listener Design
+
+**Status**: _Draft_
+
+This document describes Contours additional http listener configuration for the purpose of separating host/port configuration for health checking and metrics services.
```
jpeach, line 5: - "Contour's"
- "HTTP"

jpeach:
> Thanks @jpeach
> 
> wrt
> 
> > One issue that needs a bit of consideration is what log fields the HTTP service should use (since context could not be ambiguous).
> 
> Just clarifying if you are referring to the health service or the metrics service?

Looking at the latest changes, I think that what you have is fine.

pickledrick:
Thanks @jpeach for your thoroughness.

site/_guides/health-checking.md
```diff
@@ -0,0 +1,12 @@
+---
+title: Health Checking
+layout: page
+---
+
+Contour exposes two health endpoints `/health` and `/healthz`. By default these paths are serviced by `0.0.0.0:8000` using the `--http-address` and `--http-port` flags. The health endpoint is configurable using the `--health-port` flag.
```
jpeach, line 6:
```
using the `--health-address` and `--health-port` flags
```

PR #2381, opened 2020-03-24
internal/contour: enforce SNI name bindings

jpeach:
reviously, Contour generates a single RouteConfiguration named
`ingress_https` that contained all the virtual hosts that were being
served over HTTPS. This meant that if the client request SNI name "A",
it could send a HTTP request for virtual host "B" and it would work.
This behaviour is generally unexpected and can violate security invariants
that we want to enforce, e.g. protecting virtual hosts with client
certificate validation.

The fix is to generate a unique route configuration for each virtual host
and match them to each unique SNI name. This means that once the SNI name
is matched, all the routes from the attached route aonfiguration belong
to the same virtual host. This enforces a one to one correspondence
between SNI names and virtual hosts.

After this, the well-known "ingress_https" route configuration no longer
exists, except for TCP proxying.

This fixes #2347.

Signed-off-by: James Peach <jpeach@vmware.com>

PR #2410, opened 2020-04-01
api: added external client certificate authentication

### LINKED ISSUE #1090: Add support for external client certificate validation (mTLS)

tsaarni:
Expose support in HTTPProxy to authenticate external clients (downstream) by
validating their client certificates against trusted CA certificate.

Closes #1090


tsaarni:
I have tested this manually end-to-end and it now seems to work correctly, thanks to #2381 :thumbsup:

I'd be interested to add integration tests when the environment is available!

youngnick:
I updated the title a little bit, because we need to be clear that we're not hitting the other external authentication use case here :).

jpeach:
I'll review this, and intend to land it for the 1.4.0 release. I'm a bit sidetracked figuring out how to write and end-to-end test for it, so the review will be a couple of days away.

tsaarni:
Even the manual test procedure is unfortunately rather complicated :-(

Just in case this could help, I wrote down my manual test cases as step-by-step instructions here https://gist.github.com/tsaarni/4bc8d5e03bc180437a1aa6392430675a.
I'm using linux and that may not work on macos as it is.

jpeach:
> Even the manual test procedure is unfortunately rather complicated :-(
> 
> Just in case this could help, I wrote down my manual test cases as step-by-step instructions here https://gist.github.com/tsaarni/4bc8d5e03bc180437a1aa6392430675a.
> I'm using linux and that may not work on macos as it is.

Thanks @tsaarni, that script is a great reference. [This](https://gist.github.com/jpeach/97ab8f44cca3b3f6c2b884e4b5524796) is my test case so far. It assumes that cert-manager is installed, creates a local CA and makes a HTTPS request using the CA from the server certificate.

kind can inject a local build of Contour, and then I can verify the client auth. I'll take some test cases from your script :)

youngnick:
Happy for you to make the call @jpeach.

tsaarni:
I've now rebased and also re-executed my [manual tests](https://gist.github.com/tsaarni/4bc8d5e03bc180437a1aa6392430675a).

apis/projectcontour/v1/httpproxy.go
```diff
@@ -120,6 +120,12 @@ type TLS struct {
 	// backing cluster.
 	// +optional
 	Passthrough bool `json:"passthrough,omitempty"`
+	// ClientValidation defines how to verify the client certificate. This setting:
+	// 1. Enables TLS client certificate validation.
+	// 2. Requires clients to present a TLS certificate (i.e. not optional validation).
+	// 3. Specifies how the client certificate will be validated.
+	// +optional
+	ClientValidation *DownstreamValidation `json:"clientValidation,omitempty"`
```
stevesloka, line 128: I wonder if there's a better word to describe what "client" is. I think it's probably best, but it gets confusing with upstream/downstream/client, etc.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -379,6 +385,15 @@ type UpstreamValidation struct {
 	SubjectName string `json:"subjectName"`
 }
 
+// DownstreamValidation defines how to verify the client certificate.
+type DownstreamValidation struct {
```
stevesloka, line 389: This is just a struct name so no big deal, but do we need to give it an identity? What if we used this to validate connections elsewhere? Maybe a more generic name that could allow for re-use later?

internal/dag/builder.go
```diff
@@ -1094,6 +1104,23 @@ func (b *Builder) lookupUpstreamValidation(uv *projcontour.UpstreamValidation, n
 	}, nil
 }
 
+func (b *Builder) lookupDownstreamValidation(vc *projcontour.DownstreamValidation, namespace string) (*PeerValidationContext, error) {
+	if vc == nil {
```
stevesloka, line 1108: This never gets called.

tsaarni: I removed the dead if branch `if vc == nil...`

apis/projectcontour/v1/httpproxy.go
tsaarni: Sometimes we have used "external" to highlight that it is the client outside of the cluster. 
What about something like this:

```diff
-       // ClientValidation defines how to verify the client certificate. This setting:
+       // ClientValidation defines how to verify the client certificate when an external client establishes TLS connection to Envoy. This setting:
```

apis/projectcontour/v1/httpproxy.go
tsaarni: Could you explain bit more what you mean by giving an identity?  I think I didn't understand it :)

There was `UpstreamValidation` before and `DownstreamValidation` mirrors that with the exception that `SubjectName` attibute is not present. Otherwise single generic structure would have worked.

jpeach: The way I was thinking about this was that as we add more API surface, we would eventually converge the two struct names into a common `PeerValidation` like the one we use internally.

tsaarni:
> Another quick question, does this apply to TCP connections as well?

I did not realize this before: the code does not make difference between the two and it does work just fine with TCP connections as well. 

Maybe there should be another unittest in `internal/featuretests/downstreamvalidation_test.go` for this case, though it basically exercises the same code paths from perspective of the downstream validation?

For what it is worth, I added tests in my [manual procedure](https://gist.github.com/tsaarni/4bc8d5e03bc180437a1aa6392430675a)
* test case: successful authentication with tcpproxy in TLS termination mode
* test case: unsuccessful authentication with tcpproxy in TLS termination mode

tsaarni:
There is corner case: `passthrough: true` with `clientValidation`

```yaml
kind: HTTPProxy
metadata:
  name: tcpproxy-passthrough-with-client-auth
spec:
  virtualhost:
    fqdn: protected-tcpproxy-passthrough.127-0-0-101.nip.io
    tls:
      passthrough: true
      clientValidation:
        caSecret: client-root-ca
  tcpproxy:
    services:
      - name: httpbin-protected
        port: 80
```

Currently this does not raise error message in `HTTPProxy` status but I think this should be rejected?

I'll see how to fix it.

tsaarni:
Yet another corner case: `permitInsecure: true` with `clientValidation`

```yaml
kind: HTTPProxy
metadata:
  name: https-with-client-auth-permitinsecure
spec:
  virtualhost:
    fqdn: protected-permitinsecure.127-0-0-101.nip.io
    tls:
      secretName: ingress
      clientValidation:
        caSecret: client-root-ca
  routes:
    - services:
        - name: httpbin-protected
          port: 80
      permitInsecure: true
```

I think this should be rejected as well?

jpeach:
> tcpproxy 

I could go either way on this. Since we are proxying, we could argue that the backend service owns the auth. But this can also be used to make additional guarantees, so it is not conflicting.

> permitInsecure

I don't see anything wrong with this. You would end up with certificate auth required on port 443 but not on 80. Seems weird, but not conflicting.

/cc @youngnick @stevesloka

tsaarni:
Regarding tcpproxy conflict:

There is already a check that covers combination `passthrough: true` and `secretName` which results in error  *tcpproxy: missing tls.passthrough or tls.secretName*.  Shouldn't I extend this check also to cover `passthrough: true` and `clientValidation` as well since that is invalid combination?

The current error message is a bit misleading though since the same error handling logic covers cases when both are missing or both are defined.  I guess it could recognize this and give error such as *tcpproxy: TLS passthrough cannot be combined with tls.secretName or tls.clientValidation*

Regarding permitInsecure:

It is of course OK for me too to allow combination of `permitInsecure` and `clientValidation` in single `HTTPProxy`.  The risk of user unintentionally exposing vhost with and without authentication is probably small.

youngnick:
@tsaarni I'm in favour of marking a HTTPProxy invalid if it contains config that doesn't do anything. So, for `clientValidation: true` and passthrough, we should mark it as invalid.

I also agree that the error message should clearly what's happening, `tcpproxy: TLS passthrough cannot be combined with tls.secretName or tls.clientValidation` is great.

tsaarni:
I've now added following:
* mark `HTTPProxy` as invalid if user configures `passthrough: true` together with `clientValidation` and 
* improved error message for case `tcpproxy: TLS passthrough cannot be combined with tls.secretName or tls.clientValidation`
* test case to `builder_test.go` for *"insert httpproxy w/ tcpproxy in tls termination mode w/ downstream verification"*.  This is when Envoy really terminates TLS on behalf of backend i.e. `passthrough: false`

jpeach:
@tsaarni From now, let's please create new issues for any more changes here.

@stevesloka Are your comments addressed? I'll need to do another review pass for the latest changes but I'll wait for your approval before merging.

apis/projectcontour/v1/httpproxy.go
stevesloka: Yeah, what I meant was @jpeach mentioned. Calling the struct something more generic so that it could be reused for other similar items later.

jpeach: @stevesloka Were you thinking that we should make that change now? If we do, that would imply to API users that all the fields currently in `UpstreamValidation` would work, or we would have to add extra validation code to display errors if those fields were used in a downstream context.

stevesloka: I never said merge it with UpstreamValidation. Just ship as is. LGTM

stevesloka: @jpeach  ^^

jpeach: 'kk thanks @stevesloka . I'll resolve this conversation and we can revisit. These names don't affect API compatibility, so we can change them later if we like.

internal/dag/builder.go
```diff
@@ -478,6 +478,12 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
 		// tls is valid if passthrough == true XOR secretName != ""
 		tlsValid = tls.Passthrough != !isBlank(tls.SecretName)
 
+		// passthrough cannot be combined with secretName or clientValidation
+		if tls.Passthrough && (!isBlank(tls.SecretName) || (tls.ClientValidation != nil)) {
```
jpeach, line 482: I think that it is enough here to check the client validation. The secret name is check for TCP proxy, though that check could arguably be tighter.

```Go
if tls.Passthrough && tls.ClientValidation != nil {
        sw.SetInvalid("Spec.VirtualHost.TLS ...")
        return
}
```

internal/dag/builder.go
```diff
@@ -489,6 +495,16 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
 			svhost := b.lookupSecureVirtualHost(host)
 			svhost.Secret = sec
 			svhost.MinProtoVersion = annotation.MinProtoVersion(proxy.Spec.VirtualHost.TLS.MinimumProtocolVersion)
+
+			// Fill in DownstreamValidation when external client validation is enabled.
+			if tls.ClientValidation != nil {
+				dv, err := b.lookupDownstreamValidation(tls.ClientValidation, proxy.Namespace)
+				if err != nil {
+					sw.SetInvalid("TLS client validation policy error: %s", err)
```
jpeach, line 503: Let's make this error consistent with other TLS errors:

```
sw.SetInvalid("Spec.VirtualHost.TLS client validation is invalid: %s", err)
```

apis/projectcontour/v1/httpproxy.go
jpeach: I think the proposed replacement is an improvements 👍

internal/dag/builder.go
tsaarni: The code around this had changed a bit so I adapted this suggestion a bit, but implemented the idea.

internal/dag/builder.go
```diff
@@ -1094,6 +1110,18 @@ func (b *Builder) lookupUpstreamValidation(uv *projcontour.UpstreamValidation, n
 	}, nil
 }
 
+func (b *Builder) lookupDownstreamValidation(vc *projcontour.DownstreamValidation, namespace string) (*PeerValidationContext, error) {
```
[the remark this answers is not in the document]
tsaarni: Rebased and changed `lookupDownstreamValidation` accordingly

tsaarni:
Sorry I missed that error, Thank You!

Issue #2388, opened 2020-03-26
Ingress status updating should respect ingress class


youngnick:
After #2386, Ingresses will have status added, but it currently does not respect the Ingress class annotation. This issue covers fixing that issue.

xref #403

PR #2412, opened 2020-04-02
internal/annotation: Refactor annotations code from internal/dag

youngnick:
Updates #2388
Update #403

Refactor annotations code from `internal/dag` into `internal/annotation` so that
code dealing with ingress status will be able to use it.

This includes moving some generic Kubernetes structs (`Object` and `Meta`) into
`k8s` from `dag`. `Meta` has also been renamed to `FullName` for greater clarity.

### DIFF
--- a/internal/k8s/objectmeta.go
+++ b/internal/k8s/objectmeta.go
@@ -1,0 +1,37 @@
+// Copyright © 2020 VMware
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package k8s
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Object is any Kubernetes object that has an ObjectMeta.
+type Object interface {
+	metav1.ObjectMetaAccessor
+}
+
+// FullName holds the name and namespace of a Kubernetes object.
+type FullName struct {
+	Name, Namespace string
+}
+
+// ToFullName returns the FullName of any given Kubernetes object.
+func ToFullName(obj Object) FullName {
+	m := obj.GetObjectMeta()
+	return FullName{
+		Name:      m.GetName(),
+		Namespace: m.GetNamespace(),
+	}
+}
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -22,6 +22,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/projectcontour/contour/internal/annotation"
 	"github.com/projectcontour/contour/internal/contour"
 	"github.com/projectcontour/contour/internal/dag"
 	"github.com/projectcontour/contour/internal/debug"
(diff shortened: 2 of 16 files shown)


internal/k8s/objectmeta.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// ToMeta returns the Meta of an Object
+func ToMeta(obj Object) Meta {
+	m := obj.GetObjectMeta()
+	return Meta{
+		Name:      m.GetName(),
+		Namespace: m.GetNamespace(),
+	}
+}
```
jpeach, line 37: I can't find a place where this copy of the `Meta` type or `ToMeta` is used. Can we remove this?

internal/k8s/objectmeta.go
```diff
@@ -0,0 +1,37 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package k8s
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Object is any Kubernetes object that has an ObjectMeta
```
jpeach, line 20: Nit: end sentence with a full stop.

internal/k8s/objectmeta.go
```diff
@@ -0,0 +1,37 @@
...
+type Object interface {
+	metav1.ObjectMetaAccessor
+}
```
jpeach, line 23: Every time I see this, I always wonder why we can't use `metav1.Object". Since you have been looking at this code, would you mind adding an explanation to this type if it became clear?

internal/k8s/annotations.go
```diff
@@ -215,24 +219,69 @@ func MinProtoVersion(version string) envoy_api_v2_auth.TlsParameters_TlsProtocol
 	}
```
jpeach: `MinProtoVersion` isn't related to Kubernetes or annotations. I think it makes more sense to put this helper into the `envoy` package, but that causes an import cycle because `envoy` depends on `dag`. I suppose it could go into `dag` in a pinch.

internal/k8s/annotations.go
```diff
@@ -23,7 +23,8 @@ import (
 	"k8s.io/api/networking/v1beta1"
 )
 
-func annotationIsKnown(key string) bool {
+// AnnotationIsKnown checks if an annotation is one Contour knows about.
+func AnnotationIsKnown(key string) bool {
```
jpeach: After seeing this patch, I'm inclined to think that it makes more sense to have the annotation helpers in an `annotation` package, so that the meaning is clearer at the call site.

For example, reading `annotations.ParseUpstreamProtocols`, it is pretty clear to the reader that this is doing something with object annotations, but reading `k8s.ParseUpstreamProtocols` doesn't give any kind of clue, and there's no naming pattern that really helps out.

So, I'd suggest that we
1. pull everything except `ParseTimeout` into an `annotations` package
1. Update the signature of all the export to take a single `metav1.ObjectMetaAccessor` parameter and call `GetAnnotations()` internally
1. Rename export to not stutter, so `k8s.ValidAnnotationForKind` would become `annotations.ValidforKind`.

As commented below, I think `ParseTimeout` is a general util, not closely related to annotations, so the `k8s` package seems OK for that.

internal/k8s/objectmeta.go
youngnick: It's used in tests a lot, I think.

jpeach: I don't see it:
```
$ git grep k8s.ToMeta
$ git grep ToMeta
internal/k8s/objectmeta.go:// ToMeta returns the Meta of an Object
internal/k8s/objectmeta.go:func ToMeta(obj Object) Meta {
```

There is a `toMeta` in internal/dag/cache.go, is that it?

youngnick: ah, for posterity - I had moved this to `k8s` in an earlier draft of this PR, but had missed it when starting from scratch. I've renamed this type to `FullName` and moved it to `k8s`.

PR #2416, opened 2020-04-06
Add ingress class filtering to ingress status updating

### LINKED ISSUE #2388: Ingress status updating should respect ingress class

youngnick:
Fixes #2388
Updates #403

Move annotations to use upstream `ObjectMetaAccessor` instead of `k8s.Object`, removes an import cycle.

Copy `KindOf` from k8s to `annotations_test.go` for use in the tests only, also prevents an import cycle.

Move `ParseTimeout` from `k8s` to `annotation` for now as well, more import cycles.

Move ingress class matching code from `dag` to `annotation`, so it's only in one place.

Add ingress class filtering to `k8s.IngressStatusUpdater` `OnAdd` and `OnUpdate`.

I spent some time looking into adding Informer tests, but that requires a lot more fakeclient usage than I am comfortable with right now.

### DIFF
--- a/internal/annotation/annotations.go
+++ b/internal/annotation/annotations.go
@@ -20,9 +20,12 @@ import (
 	"time"
 
 	envoy_api_v2_auth "github.com/envoyproxy/go-control-plane/envoy/api/v2/auth"
-	"github.com/projectcontour/contour/internal/k8s"
 	"k8s.io/api/networking/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
+
+// DEFAULT_INGRESS_CLASS is the Contour default.
+const DEFAULT_INGRESS_CLASS = "contour"
 
 // IsKnown checks if an annotation is one Contour knows about.
 func IsKnown(key string) bool {
@@ -101,7 +104,7 @@ func ValidForKind(kind string, key string) bool {
 // CompatAnnotation checks the Object for the given annotation, first with the
 // "projectcontour.io/" prefix, and then with the "contour.heptio.com/" prefix
 // if that is not found.
-func CompatAnnotation(o k8s.Object, key string) string {
+func CompatAnnotation(o metav1.ObjectMetaAccessor, key string) string {
 	a := o.GetObjectMeta().GetAnnotations()
 
 	if val, ok := a["projectcontour.io/"+key]; ok {
@@ -184,7 +187,7 @@ func NumRetries(i *v1beta1.Ingress) uint32 {
 
 // PerTryTimeout returns the duration envoy will wait per retry cycle.
 func PerTryTimeout(i *v1beta1.Ingress) time.Duration {
-	return k8s.ParseTimeout(CompatAnnotation(i, "per-try-timeout"))
+	return ParseTimeout(CompatAnnotation(i, "per-try-timeout"))
 }
 
 // IngressClass returns the first matching ingress class for the following
@@ -192,7 +195,7 @@ func PerTryTimeout(i *v1beta1.Ingress) time.Duration {
 // 1. projectcontour.io/ingress.class
 // 2. contour.heptio.com/ingress.class
 // 3. kubernetes.io/ingress.class
-func IngressClass(o k8s.Object) string {
+func IngressClass(o metav1.ObjectMetaAccessor) string {
 	a := o.GetObjectMeta().GetAnnotations()
 	if class, ok := a["projectcontour.io/ingress.class"]; ok {
 		return class
@@ -204,6 +207,25 @@ func IngressClass(o k8s.Object) string {
 		return class
 	}
 	return ""
+}
+
+// MatchesIngressClass checks that the passed object has an ingress class that matches
+// either the passed ingress-class string, or DEFAULT_INGRESS_CLASS if it's empty.
+func MatchesIngressClass(o metav1.ObjectMetaAccessor, ic string) bool {
+
+	switch IngressClass(o) {
+	case ic:
+		// Handles ic == "" and ic == "custom".
+		return true
+	case DEFAULT_INGRESS_CLASS:
+		// ic == "" implicitly matches the default too.
+		if ic == "" {
+			return true
+		}
+	}
+
+	return false
+
 }
 
 // MinProtoVersion returns the TLS protocol version specified by an ingress annotation
@@ -226,7 +248,7 @@ func MinProtoVersion(version string) envoy_api_v2_auth.TlsParameters_TlsProtocol
 // 2. contour.heptio.com/max-connections
 //
 // '0' is returned if the annotation is absent or unparseable.
-func MaxConnections(o k8s.Object) uint32 {
+func MaxConnections(o metav1.ObjectMetaAccessor) uint32 {
 	return parseUInt32(CompatAnnotation(o, "max-connections"))
 }
 
@@ -236,7 +258,7 @@ func MaxConnections(o k8s.Object) uint32 {
 // 2. contour.heptio.com/max-pending-requests
 //
 // '0' is returned if the annotation is absent or unparseable.
-func MaxPendingRequests(o k8s.Object) uint32 {
+func MaxPendingRequests(o metav1.ObjectMetaAccessor) uint32 {
 	return parseUInt32(CompatAnnotation(o, "max-pending-requests"))
 }
 
@@ -246,7 +268,7 @@ func MaxPendingRequests(o k8s.Object) uint32 {
 // 2. contour.heptio.com/max-requests
 //
 // '0' is returned if the annotation is absent or unparseable.
-func MaxRequests(o k8s.Object) uint32 {
+func MaxRequests(o metav1.ObjectMetaAccessor) uint32 {
 	return parseUInt32(CompatAnnotation(o, "max-requests"))
 }
 
@@ -256,6 +278,6 @@ func MaxRequests(o k8s.Object) uint32 {
 // 2. contour.heptio.com/max-retries
 //
 // '0' is returned if the annotation is absent or unparseable.
-func MaxRetries(o k8s.Object) uint32 {
+func MaxRetries(o metav1.ObjectMetaAccessor) uint32 {
 	return parseUInt32(CompatAnnotation(o, "max-retries"))
 }
--- a/internal/annotation/annotations_test.go
+++ b/internal/annotation/annotations_test.go
@@ -20,7 +20,6 @@ import (
 	ingressroutev1 "github.com/projectcontour/contour/apis/contour/v1beta1"
 	projectcontour "github.com/projectcontour/contour/apis/projectcontour/v1"
 	"github.com/projectcontour/contour/internal/assert"
-	"github.com/projectcontour/contour/internal/k8s"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/api/networking/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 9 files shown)


internal/k8s/ingressstatus.go
```diff
@@ -14,45 +14,94 @@
...
-	// - OnUpdate transitions from an ingress class in scope to one out
-	// of scope.
+
+	oldIng := oldObj.(*v1beta1.Ingress).DeepCopy()
+	newIng := newObj.(*v1beta1.Ingress).DeepCopy()
+
+	// It's important that we only act when we are sure that there has been a change in the
+	// annotation between the old and new versions, otherwise we could generate a status update
+	// in response to another status update.
+	if !annotation.MatchesIngressClass(newIng, s.IngressClass) {
```
jpeach, line 64: I think it would be clearer to explicitly spell our the two case that we care about here:

```
if matches(old) && !matches(new) {
    // Clear status because we stopped matching.
}

if !matches(old) && matches(new) {
    // Add status because we started matching.
}
```

I think that this is clearer because the checks are always ordered old -> new, and both conditions are fully specified, so you don't have to think about the previous one when looking at the second check.

internal/k8s/ingressstatus.go
```diff
@@ -14,45 +14,94 @@
...
+		// to clear the status.
+		if annotation.MatchesIngressClass(oldIng, s.IngressClass) {
+			// The old version did match the ingress class, this object has
+			// fallen out of scope. Reset the status info.
+			s.Logger.
+				WithField("name", newIng.GetName()).
+				WithField("namespace", newIng.GetNamespace()).
+				WithField("ingress-class", annotation.IngressClass(newIng)).
+				Debug("Updated Ingress is out of scope, clearing status info")
+			newIng.Status.LoadBalancer = v1.LoadBalancerStatus{}
```
jpeach, line 75: I wonder about this. If we are no longer responsible for this object, should we touch it at all? Maybe some other ingress controller is updating the status and we could be clearing that value? There's a a number of race conditions around this ...

internal/k8s/ingressstatus.go
```diff
@@ -14,45 +14,94 @@
...
+			Debug("Updated Ingress is in scope, updating")
+		newIng.Status.LoadBalancer = s.Status
+		_, err := s.Client.NetworkingV1beta1().Ingresses(newIng.GetNamespace()).UpdateStatus(newIng)
+		if err != nil {
+			s.Logger.
+				WithField("name", newIng.GetName()).
+				WithField("namespace", newIng.GetNamespace()).
+				WithError(err).Error("unable to update status")
+		}
+	}
```
jpeach, line 104: Given the race condition mentioned above, should we add another condition here that verifies the status is up to date if we own the ingress class? (i.e. to handle the case where someone unkindly removes it)

internal/annotation/annotations_test.go
```diff
@@ -455,15 +454,119 @@ func TestAnnotationKindValidation(t *testing.T) {
...
+	}
+
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			got := MatchesIngressClass(tc.fixture, "")
+			if tc.want[0] != got {
+				t.Fatalf("IngressClass match(%v): expected %v, got %v", tc.fixture, tc.want[0], got)
+			}
+			got = MatchesIngressClass(tc.fixture, DEFAULT_INGRESS_CLASS)
+			if tc.want[0] != got {
```
jpeach, line 533: Should this be:
```suggestion
if tc.want[1] != got {
```

internal/annotation/annotations_test.go
```diff
@@ -455,15 +454,119 @@ func TestAnnotationKindValidation(t *testing.T) {
...
+				t.Fatalf("IngressClass match(%v): expected %v, got %v", tc.fixture, tc.want[0], got)
+			}
```
jpeach, line 535: Maybe we can structure this slightly differently to match the want array more closely:
```
cases := []string{"", DEFAULT_INGRESS_CLASS}
for i := 0; i < len(cases); i++ {
    got := MatchesIngressClass(tc.fixture, cases[I])
    if tc.want[i] != got {
        ...
    }
}
```

Also, `Errorf` would be better than `Fatalf` in this case.

internal/annotation/annotations_test.go
```diff
@@ -455,15 +454,119 @@ func TestAnnotationKindValidation(t *testing.T) {
...
+					Name:      "incorrect",
+					Namespace: "default",
+					Annotations: map[string]string{
+						"contour.heptio.com/ingress.class": DEFAULT_INGRESS_CLASS,
+					},
+				},
+			},
+			want: []bool{true, true},
+		},
+	}
```
jpeach, line 524: Add a test case for no annotation?

internal/annotation/annotations.go
```diff
@@ -206,6 +209,24 @@ func IngressClass(o k8s.Object) string {
 	return ""
 }
 
+func MatchesIngressClass(o metav1.ObjectMetaAccessor, ic string) bool {
+	objectClass := IngressClass(o)
+
+	switch objectClass {
```
jpeach, line 215:
```suggestion
  	switch IngressClass(o) {
```

internal/annotation/annotations.go
```diff
@@ -184,15 +187,15 @@ func NumRetries(i *v1beta1.Ingress) uint32 {
 
 // PerTryTimeout returns the duration envoy will wait per retry cycle.
 func PerTryTimeout(i *v1beta1.Ingress) time.Duration {
```
jpeach, line 189: Do you want to change these to `metav1.ObjectMetaAccessor` as well in this PR, or is that too much?

go.sum
```diff
@@ -478,6 +478,7 @@ honnef.co/go/tools v0.0.0-20190523083050-ea95bdfd59fc/go.mod h1:rf3lG4BRIbNafJWh
 k8s.io/api v0.0.0-20190918155943-95b840bb6a1f/go.mod h1:uWuOHnjmNrtQomJrvEBg0c0HRNyQ+8KTEERVsK0PW48=
 k8s.io/api v0.17.0 h1:H9d/lw+VkZKEVIUc8F3wgiQ+FUXTTr21M87jXLU7yqM=
 k8s.io/api v0.17.0/go.mod h1:npsyOePkeP0CPwyGfXDHxvypiYMJxBWAMpQxCaJ4ZxI=
+k8s.io/api v0.18.0 h1:lwYk8Vt7rsVTwjRU6pzEsa9YNhThbmbocQlKvNBB4EQ=
```
jpeach, line 481: Is this intentional?

internal/annotation/annotations_test.go
```diff
@@ -455,15 +454,119 @@ func TestAnnotationKindValidation(t *testing.T) {
...
 
+func TestMatchIngressClass(t *testing.T) {
+
+	// This is a matrix test, we are testing the annotation parser
+	// across various annotations, with two options:
+	// ingress class is empty
+	// ingress class is not empty.
+	tests := map[string]struct {
+		fixture      metav1.ObjectMetaAccessor
+		ingressClass string
```
jpeach, line 471: `ingressClass` isn't used?

go.sum
youngnick: No, not sure how it got in there. I'll remove it.

internal/annotation/annotations.go
youngnick: That seems doable, for these specific Ingress ones. I don't want to touch `dag` yet though.

internal/annotation/annotations.go
youngnick: 🤦‍♂

internal/annotation/annotations.go
youngnick: after looking at this, it's more substantial than I thought - I'll hold off to another PR.

internal/annotation/annotations_test.go
youngnick: Great, thanks.

internal/dag/cache.go
```diff
@@ -60,30 +57,22 @@ type KubernetesCache struct {
...
-	case kc.IngressClass:
-		// Handles kc.IngressClass == "" and kc.IngressClass == "custom".
-		return true
-	case DEFAULT_INGRESS_CLASS:
-		// kc.IngressClass == "" implicitly matches the default too.
-		if kc.IngressClass == "" {
-			return true
-		}
-	}
+	if ok := annotation.MatchesIngressClass(obj, kc.IngressClass); !ok {
```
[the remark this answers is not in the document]
youngnick: Yeah, I don't know what was in my head there.

internal/k8s/ingressstatus.go
youngnick: Yeah, okay, that makes sense.

internal/k8s/ingressstatus.go
youngnick: Yeah, you're right. In the event that it falls out of our scope, it's not ours anymore, and we shouldn't touch it. So this function really should only handle things coming into our scope. I'll update it to that effect, with a comment to explain.

internal/k8s/ingressstatus.go
youngnick: A lot of the complexity here is handled by MatchIngressClass - that handles the case where there's no ingress class specified in the config, which means that Contour will match no annotation, empty annotation, or "contour". If "contour" is present, and is then removed, then it will still match, so this code is not relevant.

If there's a configured ingress class, and someone removes that annotation, then it falls out of our scope, and we should no longer touch it.

tl;dr we only need to handle the "comes into scope" case here, which is neat. Thanks for pointing that out.

internal/annotation/annotations_test.go
```diff
@@ -455,15 +454,127 @@ func TestAnnotationKindValidation(t *testing.T) {
...
+		},
+	}
+
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			cases := []string{"", DEFAULT_INGRESS_CLASS}
+			for i := 0; i < len(cases); i++ {
+				got := MatchesIngressClass(tc.fixture, cases[i])
+				if tc.want[i] != got {
+					t.Errorf("IngressClass match(%v): expected %v, got %v", tc.fixture, tc.want[i], got)
```
jpeach, line 540: We should include the configured ingress class:
```suggestion
t.Errorf("matching (%v) against ingress class %q: expected %v, got %v", tc.fixture, cases[i], tc.want[i], got)
```

internal/k8s/ingressstatus.go
jpeach: Yeh I think that makes sense.

I meant, what happens if someone removes the load balancer status that we placed there? I propose that we should put it back, in which case perhaps our policy becomes "update the load balancer if the class of the new object, and the status is stale"? Is there any use case for letting other actor update this?

internal/annotation/annotations_test.go
youngnick: Agreed, I'll do this now.

internal/k8s/ingressstatus.go
youngnick: It's actually quite hard to remove the status for Ingress, because it uses the status subresource, which you need to POST to separately. You can't do it just by editing the YAML.

The thing I'm worried about is that we could get an echo effect where we update the status with the loadbalancer, something else updates in response, then we update the loadbalancer again, leading to an update cycle.

Maybe the thing to do is check that the set status matches the one we want to configure, and only update if it doesn't.

youngnick: I've added in a comment explaining the use case and why we left it out.

PR #2421, opened 2020-04-08
cmd/contour: Add flag to specify status.loadbalancer address statically

### LINKED ISSUE #2387: Add command line flag to allow specification of Ingress.status.Loadbalancer value
Once #2386 is merged, Ingress will be able to have a Service's contact details published to the `status.Loadbalancer` stanza, similarly to the Nginx ingress controller `--publish-service` command-line argument.

This issue covers adding a manual way to specify reachability info, which will cover many other use cases for now, similarly to the `--publish-status-address` command-line option to Nginx.

The flag should take a string and publish it straight into any relevant Ingress documents. Obviously, this should also mean the service discovery is disabled.

youngnick:
Adds the `--ingress-status-address` flag to specify what address should be added to Ingress `status.loadbalancer` stanza by Contour.

Setting this flag will disable the automatic Envoy service watching for those details.

Fixes #2387
Updates #403

Further documentation to come under #403.

### DIFF
--- a/cmd/contour/ingressstatus.go
+++ b/cmd/contour/ingressstatus.go
@@ -14,6 +14,7 @@
 package main
 
 import (
+	"net"
 	"sync"
 
 	"github.com/projectcontour/contour/internal/k8s"
@@ -95,5 +96,28 @@ func (isw *loadBalancerStatusWriter) Start(stop <-chan struct{}) error {
 				}
 			}()
 		}
+	}
+}
+
+func parseStatusFlag(status string) v1.LoadBalancerStatus {
+
+	// Use the parseability by net.ParseIP as a signal, since we need
+	// to pass a string into the v1.LoadBalancerIngress anyway.
+	if ip := net.ParseIP(status); ip != nil {
+		return v1.LoadBalancerStatus{
+			Ingress: []v1.LoadBalancerIngress{
+				{
+					IP: status,
+				},
+			},
+		}
+	}
+
+	return v1.LoadBalancerStatus{
+		Ingress: []v1.LoadBalancerIngress{
+			{
+				Hostname: status,
+			},
+		},
 	}
 }
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -103,7 +103,7 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 	serve.Flag("root-namespaces", "Restrict contour to searching these namespaces for root ingress routes.").StringVar(&ctx.rootNamespaces)
 
 	serve.Flag("ingress-class-name", "Contour IngressClass name.").StringVar(&ctx.ingressClass)
-
+	serve.Flag("ingress-status-address", "Address to set in Ingress object status.").StringVar(&ctx.ingressStatusAddress)
 	serve.Flag("envoy-http-access-log", "Envoy HTTP access log.").StringVar(&ctx.httpAccessLog)
 	serve.Flag("envoy-https-access-log", "Envoy HTTPS access log.").StringVar(&ctx.httpsAccessLog)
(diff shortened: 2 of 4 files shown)


cmd/contour/ingressstatus.go
```diff
@@ -97,3 +98,26 @@ func (isw *loadBalancerStatusWriter) Start(stop <-chan struct{}) error {
 		}
 	}
 }
+
+func parseStatusFlag(status string) v1.LoadBalancerStatus {
```
jpeach, line 102: Add a simple test for the three cases here?

cmd/contour/serve.go
```diff
@@ -278,14 +278,20 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+		ssw := &k8s.ServiceStatusLoadBalancerWatcher{
+			ServiceName: ctx.envoyServiceName,
+			LBStatus:    lbsw.lbStatus,
+		}
+		factory := clients.NewInformerFactoryForNamespace(ctx.envoyServiceNamespace)
+		factory.Core().V1().Services().Informer().AddEventHandler(ssw)
+		g.Add(startInformer(factory, log.WithField("context", "serviceStatusLoadBalancerWatcher")))
+
+	} else {
+		log.Infof("will set ingress status to %v, disabling watching %v service", ctx.ingressStatusAddress, ctx.envoyServiceName)
```
jpeach, line 292: nit: "setting Ingress status to ..."

and maybe prefer %q to %v?

cmd/contour/serve.go
```diff
@@ -99,7 +99,7 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 	serve.Flag("root-namespaces", "Restrict contour to searching these namespaces for root ingress routes.").StringVar(&ctx.rootNamespaces)
 
 	serve.Flag("ingress-class-name", "Contour IngressClass name.").StringVar(&ctx.ingressClass)
-
+	serve.Flag("ingress-status-address", "Addresses Envoy deployment is reachable at from outside the cluster.").StringVar(&ctx.ingressStatusAddress)
```
jpeach, line 102: s/Addresses/Address/

but maybe consider something  "Address to set in Ingress object status."?

cmd/contour/serve.go
```diff
@@ -278,14 +278,20 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	}
 	g.Add(lbsw.Start)
 
-	// step 12. register an informer to watch envoy's service.
-	ssw := &k8s.ServiceStatusLoadBalancerWatcher{
-		ServiceName: ctx.envoyServiceName,
-		LBStatus:    lbsw.lbStatus,
+	// step 12. register an informer to watch envoy's service if we haven't been given static details.
+	if ctx.ingressStatusAddress == "" {
```
stevesloka, line 282: Do we care if the address passed in is valid? Like if I entered `steve.s.%%%` for the address. Maybe a followup issue, but just thinking out loud.

youngnick: We could check that it's either a DNS name or IP address, I suppose.

youngnick: Actually, I've thought over this more, and I don't think it's worth doing much validation here. We already check if it's an IP or a hostname, I think that's enough.

cmd/contour/ingressstatus.go
youngnick: There's only two cases? But I can add a test.

jpeach: I was thinking IPv4, IPv4, arbitrary string as the 3 cases.

youngnick: IPv6, IPv4, arbitrary string makes sense, sure.

PR #2432, opened 2020-04-10
Upgrade Envoy go-control-plane and fix related changes to the spec

### LINKED ISSUE #2429: Upgrade go-control-plane
Should keep up to date with go-control-plane as new features are moved to the v3 API.

stevesloka:
Fixes #2429

### DIFF
--- a/internal/assert/assert.go
+++ b/internal/assert/assert.go
@@ -55,6 +55,9 @@ func (a Assert) Equal(want, got interface{}) {
 }
 
 func unmarshalAny(a *any.Any) proto.Message {
+	if a == nil {
+		return nil
+	}
 	pb, err := ptypes.Empty(a)
 	if err != nil {
 		panic(err.Error())
--- a/internal/contour/cluster.go
+++ b/internal/contour/cluster.go
@@ -17,8 +17,9 @@ import (
 	"sort"
 	"sync"
 
+	resource "github.com/envoyproxy/go-control-plane/pkg/resource/v2"
+
 	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
-	"github.com/envoyproxy/go-control-plane/pkg/cache"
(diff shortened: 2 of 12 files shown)


internal/contour/cluster.go
```diff
@@ -17,8 +17,9 @@ import (
 	"sort"
 	"sync"
 
+	rsrc "github.com/envoyproxy/go-control-plane/pkg/resource/v2"
+
```
jpeach, line 21: Do we have a standard style for grouping imports?

internal/assert/assert.go
```diff
@@ -55,6 +55,9 @@ func (a Assert) Equal(want, got interface{}) {
 }
 
 func unmarshalAny(a *any.Any) proto.Message {
+	if a == nil {
+		return nil
+	}
```
jpeach, line 60: What happened here? Does this get a nil that it didn't get before?

internal/contour/cluster.go
stevesloka: I don't know if we have a standard, I copied this from the go-control-plane code.

internal/assert/assert.go
stevesloka: This one was weird, after upgrading the deps, some tests started to fail. The error was the comparison was nil when checking. 

Maybe if you could put a set of eyes on it and see if something else jumps out at you.

jpeach: I think that we just got lucky before. Any comparison involving an `any.Any` could have triggered this, since it makes `ptypes.Empty` return an error. Looks like ` RetryPolicyTypedConfig *any.Any` was added to a few types, which was probably one trigger.

internal/contour/cluster.go
stevesloka: I renamed it to `resource`. I didn't think the other alias made sense.

PR #2438, opened 2020-04-14
site: update envoy support matrix for Envoy v1.14.1

### LINKED ISSUE #2435: site: Update Envoy matrix for next release
As a follow-up, need to update the Envoy resources page once the next release goes out.

_Originally posted by @stevesloka in https://github.com/projectcontour/contour/pull/2434#issuecomment-612088531_

stevesloka:
Fixes #2435 by updating the Envoy support matrix for Envoy v1.14.1

### DIFF
--- a/site/_resources/envoy.md
+++ b/site/_resources/envoy.md
@@ -7,35 +7,30 @@ Due to the aggressive deprecation cycle of Envoy's xDS API, not all versions of 
 
 This page describes the compatibility matrix of Contour and Envoy versions.
 
-## Supported Envoy versions
+## Supported Contour/Envoy Versions
 
-| Envoy version | Contour v1.0.0<sup>5</sup>| Contour v1.0.1<sup>6</sup>| Contour v1.1.0<sup>7</sup>| Contour v1.2.0<sup>8</sup>| Contour v1.2.1            | Contour v1.3.0            |
-| ------------- | :-----------------------: | :-----------------------: | :-----------------------: | :-----------------------: | :-----------------------: | :-----------------------: |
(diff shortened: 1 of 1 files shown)


stevesloka:
Since "resources" aren't versioned, it looks weird to users to merge since Contour v1.4.0 isn't released yet. We should hold on to this PR until the release cadence.

jpeach:
> Since "resources" aren't versioned, it looks weird to users to merge since Contour v1.4.0 isn't released yet. We should hold on to this PR until the release cadence.

Oh good point. Maybe we should move this to versioned docs, since the support is very likely to change on release boundaries?

youngnick:
A couple of general things:
- There are hardly any "supported" cells in the table - I think that having the whole table there will make it increasingly harder to read. Not sure what to do here - maybe just have the "supported" cells filled out, or highlight them more? Maybe we can just say that we'll only list the last n versions here or something?

Changing it to versioned seems good to me.

stevesloka:
I don't think this table is all that valuable. Since releases are tied to Envoy versions, why do we have a separate place to hold this information? The release tells you what version to use, or at least, those release docs should be very clear on this information.

youngnick:
It's good to have the information in one place, in case you are stuck on an old version and need to check what Envoy version you should use or something.

Maybe we just need a table with two columns: "Contour Version" and "Envoy version". Have the table sorted descending by Contour version, then we just add the new one on the top when we release.

site/_resources/envoy.md
```diff
@@ -1,41 +1,36 @@
 ---
-title: Envoy Support Matrix
+title: Envoy cveport Matrix
```
jpeach, line 2: "Support" :)

youngnick: That is a spicy typo though!

stevesloka: image

youngnick:
Merging this now, as I'm going to cut Contour 1.4 today.

PR #2440, opened 2020-04-15
site: add code highlighting

jpeach:
Enable rouge as the code higherlighter. Add CSS for the default pygments
theme, which has reasonable contrast against a grey background.


jpeach:
> LGTM, neat. I'd say one of the others more familiar with Jekyll should also review though.

@SDBrett ??

PR #2442, opened 2020-04-15
use sni for clusters when requestHeaderPolicy is set

### LINKED ISSUE #2437: ExternalName proxying uses external service name for SNI
If a user wants to proxy to an external resource using Kubernetes service type `externalName`, and "tls" is set on the service, the SNI used during the proxy is the value of the `externalName`. 

If, for example, a user wants to proxy to a resource using externalName, but the `host` header doesn't match, the wrong SNI is used.

What this example needs is to use the value of the `requestsHeadersPolicy` rewrite as the SNI value not the externalName. 

Here's where the change needs to be done: https://github.com/projectcontour/contour/blob/master/internal/envoy/cluster.go#L83

stevesloka:
Fixes #2437 by setting the SNI for the request with the following rules:

1. If the requestHeadersPolicy.set.Host is set on the service
2. If the requestHeadersPolicy.set.Host is set on the route
3. If an externalName is defined 

//cc @jeremyrickard

### DIFF
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -858,6 +858,7 @@ func (b *Builder) computeRoutes(sw *ObjectStatusWriter, proxy *projcontour.HTTPP
 				RequestHeadersPolicy:  reqHP,
 				ResponseHeadersPolicy: respHP,
 				Protocol:              protocol,
+				SNI:                   determineSNI(r.RequestHeadersPolicy, reqHP, s),
 			}
 			if service.Mirror && r.MirrorPolicy != nil {
 				sw.SetInvalid("only one service per route may be nominated as mirror")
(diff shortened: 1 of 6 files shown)


jeremyrickard:
Thank you @stevesloka! It was nice to review this as I was walking through the code base to determine how to implement this. Thanks for the quick turn around!

internal/dag/dag.go
```diff
@@ -409,6 +409,9 @@ type Cluster struct {
 
 	// ResponseHeadersPolicy defines how headers are managed during forwarding
 	ResponseHeadersPolicy *HeadersPolicy
+
+	// SNI describes how the SNI is set on a Cluster
+	SNI string
```
youngnick, line 414: I think this could use a little explanation of how the field works.

What happens when it's set?

What happens when it's the empty string?

It seems from the rest of the code that if it's not set having the field be present does nothing, it would be good to have that contract explicitly stated here.

A link to the Envoy docs or something would maybe do as well.

internal/dag/builder.go
```diff
@@ -880,6 +881,26 @@ func (b *Builder) computeRoutes(sw *ObjectStatusWriter, proxy *projcontour.HTTPP
 	return routes
 }
 
+// determineSNI decides what the SNI should be on the request
```
youngnick, line 884: Some explanation of what happens where there should be no SNI on the request would be good.

internal/dag/dag.go
youngnick: That's great, thanks.

Issue #2045, opened 2019-12-12
Generate config file documentation


jpeach:
To ensure that the Contour configuration file documentation is always up to date, we should generate it from code comments in the Contour source.

jpeach:
Related to #2111

jpeach:
xref #2621 #3023

xaleeks:
Agreed.  Ideally it comes from some config system or CI/CD pipeline so we have full audit capabilities, know who made what changes and can tie this back to the specific release with a little bit of effort.  If user deploys an older version of Contour, they can also grab the older config file that goes with it.  This is not critical for us though, but it’d be a nice bonus

PR #2445, opened 2020-04-16
doc: document the Contour configuration file

jpeach:
Document the Contour configuration file.

This updates #2045.

### DIFF
--- a/site/docs/master/configuration.md
+++ b/site/docs/master/configuration.md
@@ -1,9 +1,55 @@
-# Contour Configuration
+# Contour Configuration Reference
 
-A configuration file can be passed to the `contour serve` command which specified additional properties that Contour should use when starting up.
-This file is passed to Contour via a ConfigMap which is mounted as a volume to the Contour pod.
+## Configuration File
 
-Following is an example ConfigMap with configuration file included: 
+A configuration file can be passed to the `--config-path` argument of the `contour serve` command to specify additional configuration to Contour.
+In most deployments, this file is passed to Contour via a ConfigMap which is mounted as a volume to the Contour pod.
+
+The Contour configuration file is optional.
+In its absence, Contour will operate with reasonable defaults.
+Where Contour settings can also be specified with command-line flags, the command-line value takes precedence over the configuration file.
+
+| Field Name | Type | Default | Description |
+|------------|------|---------|-------------|
+| accesslog-format | string | `envoy` | This key sets the global [access log format][2] for Envoy. Valid options are `envoy` or `json`. |
+| debug | boolean | `false` | Enables debug logging. |
+| disablePermitInsecure | boolean | `false` | If this field is true, Contour will ignore `PermitInsecure` field in HTTPProxy documents. |
+| incluster | boolean | `false` | This field specifies that Contour is running in a Kubernetes cluster and should use the in-cluster client access configuration.  |
+| json-fields | string array | [fields][5]| This is the list the field names to include in the JSON [access log format][2]. |
+| kubeconfig | string | `$HOME/.kube/config` | Path to a Kubernetes [kubeconfig file][3] for when Contour is executed outside a cluster. |
+| leaderelection | leaderelection | | The [leader election configuration](#leader-election-configuration). |
+| request-timeout | [duration][4] | `0s` | This field specifies the default request timeout as a Go duration string. Zero means there is no timeout. |
+| tls | TLS | | The default [TLS configuration](#tls-configuration). |
+{: class="table thead-dark table-bordered"}
+
+### TLS Configuration
+
+The TLS configuration block can be used to configure default values for how
+Contour should provision TLS hosts.
+
+| Field Name | Type| Default  | Description |
+|------------|-----|----------|-------------|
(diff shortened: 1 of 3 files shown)


site/docs/master/configuration.md
```diff
@@ -1,9 +1,54 @@
 # Contour Configuration
 
+## Configuration File
```
stevesloka, line 3: nit: Maybe just change the title to be `Contour Configuration File`

site/docs/master/configuration.md
```diff
@@ -1,9 +1,54 @@
...
+In its absence, Contour will operate with reasonable defaults.
+
+| Field Name | Type | Description |
+|------------|------|-------------|
+| accesslog-format | string | This key sets the global [access log format][2] for Envoy. Valid options are `envoy` or `json`. |
+| debug | boolean | Enables debug logging. |
+| disablePermitInsecure | boolean | If this field is true, Contour will ignore `PermitInsecure` field in HTTPProxy documents. |
+| incluster | boolean | This field specifies that Contour is running in a Kubernetes cluster and should use the in-cluster client access configuration.  |
+| json-fields | string array | This is list the field names to include in the JSON [access log format][2]. |
+| kubeconfig | string | Path to a Kubernetes [kubeconfig file][3]. This should not be needed in the vast majority of deployments. |
```
stevesloka, line 18: nit: Maybe change the vast majority of deployments line to say what it's used for? Like, only used when running outside of the cluster?

site/docs/master/configuration.md
```diff
@@ -1,9 +1,54 @@
...
+
 A configuration file can be passed to the `contour serve` command which specified additional properties that Contour should use when starting up.
 This file is passed to Contour via a ConfigMap which is mounted as a volume to the Contour pod.
 
-Following is an example ConfigMap with configuration file included: 
+The Contour configuration file is optional.
+In its absence, Contour will operate with reasonable defaults.
+
+| Field Name | Type | Description |
+|------------|------|-------------|
```
stevesloka, line 12: Should we add a "defaults" column to explain what the defaults are for these values?

site/docs/master/configuration.md
```diff
@@ -1,9 +1,54 @@
 # Contour Configuration
 
+## Configuration File
+
```
stevesloka, line 4: Should we also explain the hierarchy of what values override which? Meaning if I pass say `incluster` in the config file but also set the arg to `contour serve`, what happens?

jpeach:
> We should also copy this to the v1.3.0 release notes.

Since 1.4 is coming out next week, I'd rather land this for master and then let the release cut it over. Do you think it's important to add this in 1.3 considering how long we have lived without it?

site/docs/master/configuration.md
jpeach: Good point, I'll add that.

site/docs/master/configuration.md
jpeach: Since this page is all for Contour, I felt that it was clear. I'm planning to add similar sections for command line flags and environment variables. Do you think that titling all of these "Contour Foo" would get too repetitive?

jpeach: Next push, I'm going to rename this page to "Contour Configuration Reference", WDYT?

site/docs/master/configuration.md
jpeach: I'll add this to see what you think

jpeach: Adding defaults looks OK to me.

jpeach:
@stevesloka Updated.

site/docs/master/configuration.md
```diff
@@ -1,9 +1,55 @@
...
+{: class="table thead-dark table-bordered"}
+
+### TLS Configuration
+
+The TLS configuration block can be used to configure default values for how
+Contour should provision TLS hosts.
+
+| Field Name | Type| Default  | Description |
+|------------|-----|----------|-------------|
+| minimum-protocol-version| string | `""` | This field specifies the minimum TLS protocol version that is allowed. Valid options are `1.3` and `1.3`. Any other value defaults to TLS 1.1. |
```
youngnick, line 32:
```suggestion
| minimum-protocol-version| string | `""` | This field specifies the minimum TLS protocol version that is allowed. Valid options are `1.2` and `1.3`. Any other value defaults to TLS 1.1. |
```

stevesloka:
> Since 1.4 is coming out next week, I'd rather land this for master and then let the release cut it over. Do you think it's important to add this in 1.3 considering how long we have lived without it?

I just think it's a good habit to be consistent. In reality, is it a big deal? Not really, but why not? =)

jpeach:
> On Apr 18, 2020, at 1:43 AM, Steve Sloka <notifications@github.com> wrote:
> 
> I just think it's a good habit to be consistent. In reality, is it a big deal? Not really, but why not? =)

I thought that this was the consistent path :)

Isn’t the methodology that we make changes on the master copy, not previous ones?

youngnick:
If it's in the versioned docs, then I agree, we make changes in master and they come out in the release. If we don't want that behaviour, then it shouldn't be in the versioned docs.

PR #2448, opened 2020-04-19
upgrade client-go to v0.18.2

### LINKED ISSUE #2447: Upgrade client-go to v1.18.2
Need to upgrade to client-go v0.18.2 to get the new Ingress v1 types.

//cc #2139

stevesloka:
Fixes #2447 by upgrading client-go to v0.18.2. Also upgrades controller-runtime since client-go upgrade broke the yaml generation.


youngnick:
Technically, yes. Client-go has n-2 support, so this means that we are only supported back to 1.16.

In practice, things will probably work just fine because most objects we use are now GA.

PR #2458, opened 2020-04-22
Add docs search to the site

jonasrosland:
This will add documentation search to the website.

Signed-off-by: jonasrosland <jrosland@vmware.com>


stevesloka:
I wonder if we could make the search bar line up with the release version drop-down? 

image

Some other things we can work on later, but noticed that the search can have duplicates:

image

jonasrosland:
I've changed the search bar to line up with the version, like this:
image

PR #2467, opened 2020-04-24
doc: update HeadersPolicy documentation

### LINKED ISSUE #2462: Header rewrite docs claim it only works for the Host header
From https://projectcontour.io/docs/master/httpproxy/#header-policy

> The requestHeadersPolicy supports a list of Set options that currently only supports rewriting Host headers defined via a name and value.

Pretty sure that we support setting arbitrary headers.

jpeach:
Update the API reference for `HeadersPolicy` to clarfy the semantics of
setting headers and to mention the special treatment of the `Host` header.

Rewrite the primary HTTPProxy guide to document that general header
rewriting is available in the `HeadersPolicy` fields.

This fixes #2462.

### DIFF
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -357,12 +357,16 @@ type LoadBalancerPolicy struct {
 	Strategy string `json:"strategy,omitempty"`
 }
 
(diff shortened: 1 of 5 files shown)


jpeach:
I'm also thinking that we should fold the section titled "Proxy to external resource" in to the extenalname section. LMK.

Issue #2088, opened 2020-01-08
Move the deployment YAML to kustomize


jpeach:
The current Contour deployment YAML builds from the fragments in `examples/contour` and doesn't give any help to operators that need to modify the deployment in any way.

If we move this across to a kustomize configuration, we have the opportunity to be able to modify the YAML for different deployment targets (e.g. AWS, GCP, Kind), as well as generate the same consolidated YAML file that we document for the Contour quickstart.  Other people that deploy Contour may be able to consume the kustomize configuration if we structure it correctly.

Pros of kustomize:
* Gives some deployment flexibility we don't have today
* Built in to kubectl

Cons on kustomize:
* Documentation leaves a something to be desired
* Limited configuration flexibility (e.g. hard or impossible to remove components)

Proof of concept: https://github.com/jpeach/contour/commit/1c575c772e9fd747fba72ae41ab99bdae7a01864

Related to #1190, #2050

jpeach:
One problem the just bit me is that `kustomize` and `kubectl kustomize` releases are not aligned. This means that it is possible to use syntax that `kustomize` will accept but `kubectl` won't. I'm currently in this situation with fieldRef expressions and I have:

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.0", GitCommit:"70132b0f130acc0bed193d9ba59dd186f0e634cf", GitTreeState:"clean", BuildDate:"2019-12-13T11:52:32Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.0-gke.20", GitCommit:"d324c1db214acfc1ff3d543767f33feab3f4dcaa", GitTreeState:"clean", BuildDate:"2019-11-26T20:51:21Z", GoVersion:"go1.12.11b4", Compiler:"gc", Platform:"linux/amd64"}
```

```
$ kustomize version
{Version:3.5.3 GitCommit:5ba90fe5ef1dc4599e359edd41d1d0e6373b247d BuildDate:2019-12-18T03:07:49+00:00 GoOs:darwin GoArch:amd64}
```

stevesloka:
My first thought was to default to using the version of kustomize in kubectl, but there's no way to know what version of kubectl a user has. 

I think we just need to pick a version of kustomize and document it.

jpeach:
[...]

Issue #1190, opened 2019-06-21
Contour deployment docs / tooling


stevesloka:
Contour has an `examples` folder in the root of the repo which shows different ways to deploy Contour in some various scenarios. Depending on requirements these might be deploying Envoy with `hostNetworking` for performance, or as a `daemonset` to horizontally scale the network data plane. 

We've also had the community ask for other deployment options other than just yaml (as exists today in the examples directory). Those could be helm, kustomize, or any other style of tooling to assist with deploying Contour to a cluster. 

We should also create some documentation on the various examples and discuss why one version is preferable over another and the trade-offs included.

- [ ] Create deployment docs
- [ ] Create helm chart or other tooling

victornoel:
Here is a use case for an officially supported helm chart.

To operate my kubernetes clusters, I use a combination of helmfile and helm to store the whole cluster configuration in a git repository. In this context, I usually use upstream charts for the dependencies in order to reduce the surface of resources to maintain internally.

Without an officially provided chart, I would create one myself exactly for my needs, but it would most certainly not contain the best practices to deploy contour and also I wouldn't be very confident to know if I am deploying it correctly. This especially as time goes by: maybe practices change and I miss the information and stay with an incorrect way of deploying contour.

So I feel like the contour team is the best person to maintain an up to date chart ;)

victornoel:
I was thinking about this, and an alternative to provide an helm chart would be to provide a self-sufficient contour operator that have CRDs to describe the envoy deployments on top of the `IngressRoute`.

In this case, if the operator has zero configuration because everything happens via declarative resources, it could make sense to deploy it by hand without the need for an helm chart (or other packaging solution).

In the end, the use case is always to be able to easily describe and configure the deployments as an ops and to be able to store it in git.

Issue #855, opened 2019-01-22
Internal and External ingress with 2 contour deployments


so0k:
**Objective**

I need to configure internal and external ingress controllers on AWS with NLB 

Internal Ingress is served by an internal NLB (only accessible from within the VPC)
External Ingress is served by an external NLB (publicly accessible)

**What steps did you take and what happened:**

I'm using `deployment/ds-hostnet-split` as a guideline.

[full final manifest](https://gist.github.com/so0k/2cb93c405cf05fcb8ecb73fa665efbd0)

- I'm using the same rbac / crb / sa for both contour deployments
- I have 2 Contour deployments:
    -  name & label:`contour-external` for deploy / svc (clusterIP) manifest
        the arguments:
        ```
        - serve
        - --incluster
        - --xds-address
        - 0.0.0.0
        - --xds-port
        - $(CONTOUR_EXTERNAL_SERVICE_PORT)
        - --envoy-http-port
        - "80"   #<<<<<
        - --envoy-https-port
        - "443" #<<<<<
        - --ingress-class-name=envoy-external  
        ````
    -  name & label: `contour-internal` for deploy / svc (clusterIP) manifest
        the arguments:
        ```
        - serve
        - --incluster
        - --xds-address
        - 0.0.0.0
        - --xds-port
        - $(CONTOUR_INTERNAL_SERVICE_PORT)
        - --envoy-http-port
        - "8080" #<<<<<
        - --envoy-https-port
        - "8443" #<<<<<
        - --ingress-class-name=envoy-internal  
        ````
- I have 2 Envoy daemonsets with `hostNetwork: true`:
    -  name & label: `envoy-external` for ds / svc (LoadBalancer:NLB) manifest
        environment:
        ```
        - name: NODE_NAME
          valueFrom:
            fieldRef:
              fieldPath: spec.nodeName
        ```
        args:
        ```
        - -c
        - /config/contour.yaml
        - --service-cluster
        - envoy-external-cluster
        - --service-node
        - $(NODE_NAME)
        ````
        bootstrap:
        ```
        - --xds-address
        - $(CONTOUR_EXTERNAL_SERVICE_HOST)
        - --xds-port
        - $(CONTOUR_EXTERNAL_SERVICE_PORT)
        - --stats-port
        - "8002"
        - --admin-port
        - "9001"
        ````
         HealthCheck uses `8002` 
         preStop uses `9001` 
[...]

PR #2474, opened 2020-04-27
deployment: move Contour deployment to Kustomize

jpeach:
Move the example deployment to Kustomize. This requires the `kustomize`
tool, since the version of Kustomize vendored in `kubectl apply -k`
is too old to support.

The YAML documents in the example deployment are broken into 4 components
located in `config/components` - types, contour, envoy and certgen. These
are all included in the default deployments, but operators have the
option of creating deployments that dont't include all the components.
The `types-v1` component contains the Contour CRDs suitable for Kubernetes
1.16 or later.

Deployments to various Kubernetes infrastructure are in the `deployment`
directory. The base deployment pulls in all the components and sets the
namespace to `projectcontour`. The `kind` deployment updates the Envoy
Daemonset to use a `NodePort` service, and the `aws` deployment enables
TCP load balancing with PROXY protocol support. No special options are
needed for `gke` as far as I know, but it is included for completeness.

The traditional quickstart YAML is now located at `config/quickstary.yaml`
and is just a rendering of the base deployment. The netlify redirect can't
be updated until after a release because it points to a release branch.

This updates #855, #1190, #2088, #2544.


jpeach:
@stevesloka @youngnick Posting this PR as a draft to emphasize that I'd like us all to discuss and be in agreement with the approach. I'm proposing this as a basis for a reference install that would be recommended, rather than an example. I think that this can be sustainable as a reference install because users can select the base components they need and pull them directly into kustomize, or some other YAML-centric pipeline.

jpeach:
Travis failed because kustomize isn't in that build environment
```
$ make generate
253Updating generated CRD deep-copy API code ...
254Generating example deployment files ...
255./hack/generate-deployment.sh: line 30: kustomize: command not found
256Makefile:183: recipe for target 'generate-deployment' failed
257make: *** [generate-deployment] Error 127
258The command "make generate" exited with 2.
2590.02s$ ./hack/travis/check-uncommitted-codegen.sh
260Uncommitted changes in generated sources:
261 M config/quickstart.yaml
262The command "./hack/travis/check-uncommitted-codegen.sh" exited with 1.
```

stevesloka:
I think this is probably a good step forward. We should look back and see, however, who would-be users of this. We've had many more requests for Helm charts off the top of my head. 

Is there a kustomize requirement to break things up into `components`, `types,` etc? I worry that it makes the deployment seem too complex now that it's spread apart. Also, I personally use the `examples/contour` directory daily to deploy Contour locally. I don't care to switch up my workflow, but I worry it's difficult to grok the model for someone new. 

Maybe the customize workflow could spit out the current `examples/contour` directory for simplicity?

>The kind deployment updates the Envoy Daemonset to use a NodePort service

Why change this? I would expose ports 80/443 over locahost and you could remove this config all together. 

>  aws deployment enables TCP load balancing with PROXY protocol support

We've always left this in the default, it's only consumed by the AWS provider so you can remove another set of config.

jpeach:
> I think this is probably a good step forward. We should look back and see, however, who would-be users of this. We've had many more requests for Helm charts off the top of my head.

I think/hope that this is complementary to helm and other templating approaches. With kustomize we can generate that parts of the YAML that a helm chart needs; we don't need to constrain ourselves to using kustomize to poke changes into the YAML. helm or yet would consume YAML that comes out of kustomize, or can reference the component YAML directly.

> Is there a kustomize requirement to break things up into `components`, `types,` etc? I worry that it makes the deployment seem too complex now that it's spread apart.

There's no requirement to do this, but I think that it actually makes the deployment easier to understand when we provide clear components and boundaries. There are a few different ways we can carve up the monolithic YAML, but we have existing use cases for all the components proposed here:

- for installing contour multiple times (e.g. knative), you need to install the types once, then contour/envoy/certgen multiple times into different namespaces
- if you have your own certificates, then you would omit certgen
- if you have your own envoy, then you would omit envoy

>  Also, I personally use the `examples/contour` directory daily to deploy Contour locally. I don't care to switch up my workflow, but I worry it's difficult to grok the model for someone new.

I've preserved that by generating `config/quickstart.yaml` from the base deployment directory. We will have to flip the quickstart redirect after the next release.
 
> Maybe the customize workflow could spit out the current `examples/contour` directory for simplicity?

One aspect of this work is to move away from examples, to fully supported deployment approaches. I agree with you that a single-file YAML deployment is useful to support, but I'd rather move that out of examples.

> > The kind deployment updates the Envoy Daemonset to use a NodePort service
> 
> Why change this? I would expose ports 80/443 over locahost and you could remove this config all together.

I don't remember why I had this, I thought that I got it from you? There's not much difference AFAICT, though NodePort seems slightly preferable for kind (and it's easy to toggle here).

LoadBalancer:
```
projectcontour   envoy        LoadBalancer   10.96.35.182   <pending>     80:32341/TCP,443:31583/TCP   117s
```

NodePort:
```
projectcontour   envoy        NodePort    10.96.35.182   <none>        80:32341/TCP,443:31583/TCP   49s
```

> > aws deployment enables TCP load balancing with PROXY protocol support
> 
> We've always left this in the default, it's only consumed by the AWS provider so you can remove another set of config.

Right, but IIUC for the annotation to work correctly you also need to enable PROXY protocol in Contour. Keeping all the AWS tweaks in the same deployment directory makes them easier to discover and means that non-AWS deployments don't get unnecessary configuration.

One of the reasons I've added the `deployments` directory is to capture the little tweaks that make deploying to a target a high quality experience. Even if nothing is needed (e.g. GKE), then the directory serves as documentation for that. In the past, with only a single YAML file, it hasn't been possible to tweak things just right, but I'm hoping that with this approach we can do better.

stevesloka:
It all just is too much for me, we went from a simple 5 yaml files to directories and files and things to run on top of other things to run.

stevesloka:
@youngnick what are your thoughts on this?

youngnick:
I think that having *something* like this is unavoidable, especially as we add more configurability. The more configurability we add, the more we need to add documentation for it, and the best type of documentation is that which includes a 'to install this run this command'.

I think that, in practice, most places that run an app like Contour in production end up with  a pipeline that does something _like_ what this Kustomize does, whether it's Kustomize, Helm, some Go Template rig, Ansible, or whatever. Breaking the YAML apart into more consumable components makes that much easier, and makes upgrading easier too.

So, I guess I'm okay with this in principle, but I'd like to ask the community about the choice of tool. Do our users care if it's kustomize, or Helm? If not, then we should go with this, since most of the hard work is already done.

@jpeach, will this make the integration testing life any easier? How does it interact there? I'd prefer an install update that has a path forward for integration testing, and ease of use for local dev too. I'd be okay with doc updates and maybe Make targets for this.

That said, I wouldn't be willing to merge this PR without either included install documentation updates, or a close-behind followup PR that updated all the install and example documentation with how to use the new regime.

jpeach:
> So, I guess I'm okay with this in principle, but I'd like to ask the community about the choice of tool. Do our users care if it's kustomize, or Helm? If not, then we should go with this, since most of the hard work is already done.

There will be a wide variety of tools for the foreseeable future. I think that constrains us somewhat because we need to publish something that lots of toolchains can support. Since kustomize is pretty minimal and basically generates YAML, I think that it is relatively easy for other systems to consume. I definitely would like to support helm as well, built on top of this work.

> @jpeach, will this make the integration testing life any easier? How does it interact there? I'd prefer an install update that has a path forward for integration testing, and ease of use for local dev too. I'd be okay with doc updates and maybe Make targets for this.

We can use this to support dev environments, though in that case, we can afford to be more opinionated about tooling. I have a [kind-devel](https://github.com/jpeach/kustomize/tree/master/contour/configurations/kind-devel) configuration that consumes the kustomize pieces to deploy a debug config where you run a local contour.

As far as integration testing goes, this starts us on the path of being able to standardize deployment configurations for integration testing, which is useful. 
> That said, I wouldn't be willing to merge this PR without either included install documentation updates, or a close-behind followup PR that updated all the install and example documentation with how to use the new regime.

Yes, agreed. The component structure here gives us a corresponding structure to document the install. We probably want to publish the YAML as versioned URLs, which needs some figuring out. The patch here preserves the existing published quickstart YAML, the only difference there being that it no longer contains the AWS annotations (those are only in the AWS deployment target).

We have a few options for choosing what to publish and document, any combination of:
- the contour kustomize bases
- the kustomize deployment targets
- a single global quickstart YAML
- a single YAML per deployment target

bgagnon:
I don't think we're representative of the majority of users, but I can provide an insight on how we deploy Contour on the shared platform in my organisation:
* we maintain [Jsonnet](https://jsonnet.org/) recipes for the `Deployment`, `DaemonSet`, `Service` and `ConfigMap` using [ksonnet-lib](https://github.com/ksonnet/ksonnet-lib)
* we fetch the `examples/contour` directory from this repo via [jsonnet-bundler](https://github.com/jsonnet-bundler/jsonnet-bundler)
* we convert the upstream `examples/contour/01-crds.yaml` and `examples/contour/02-rbac.yaml` file streams to individual JSON files using `yq`
* we import these JSON files from Jsonnet and inject additional `metadata.labels` on each object
* we also override the `namespace` value for each subject in the `ClusterRoleBinding` objects
* we throw in `PodMonitor` and `PrometheusRule` objects to collect metrics from Contour and Envoy
* we spit out all the manifests to JSON files

In the end we obtain the following tree, which is passed to the deployment tool (eg. `kubectl apply`):

```
.
├── contour
│   ├── contour-configmap.json
│   ├── contour-deployment.json
│   ├── contour-podmonitor.json
│   ├── contour-prometheusrule.json
│   ├── contour-service.json
│   ├── crd
│   │   ├── httpproxies.projectcontour.io-customresourcedefinition.json
│   │   ├── ingressroutes.contour.heptio.com-customresourcedefinition.json
│   │   ├── tlscertificatedelegations.contour.heptio.com-customresourcedefinition.json
│   │   └── tlscertificatedelegations.projectcontour.io-customresourcedefinition.json
│   └── rbac
│       ├── contour-clusterrole.json
│       ├── contour-clusterrolebinding.json
│       ├── contour-leaderelection-role.json
│       ├── contour-leaderelection-rolebinding.json
│       └── contour-serviceaccount.json
├── envoy
│   ├── envoy-bootstrap-configmap.json
│   ├── envoy-daemonset.json
│   ├── envoy-podmonitor.json
│   ├── envoy-prometheusrule.json
│   ├── envoy-service.json
│   └── rbac
│       ├── envoy-clusterrole.json
│       ├── envoy-rolebinding.json
│       └── envoy-serviceaccount.json
```

In the end we only have a partial dependency on the upstream manifests (the subject of this PR). We wanted to nail down the pieces that relate directly to the Contour implementation:
* `CustomResourceDefinitions`
* `ClusterRole` permissions

We explored the idea of leveraging more of the upstream manifests (`Deployment`, `Service`, etc.), but the number of overrides that would have been needed would have lead to an overlay almost as big as the manifests themselves. Plus, it's not really less fragile.

We have the following (relatively minor) pain points with our setup:

* vanilla Jsonnet [does not support YAML natively](https://github.com/google/jsonnet/issues/460), so we need some pre-processing 
* YAML file streams are not supported by many YAML parsing tools (hence `yq`)
* when upgrading contour, we have to see how `examples` shifted around from the prior release
* when upgrading contour, we have to read the release notes carefully to catch CLI and config changes, and sometimes diff the upstream YAML manifests to understand the changes we need to make to our Jsonnet generators

Regarding this PR, I don't see much that would help us. The separation into multiple themed YAML files in the previous releases was already a big help.

One (selfish) request would be to keep the canonical CRD and ClusterRole objects in individual JSON files (not YAML file streams) upstream, so that we can read them directly. Since YAML is a superset of JSON, it shouldn't break kustomize or any other tool. Maybe those files shouldn't be in `examples`, but somewhere in the codebase instead.

To broaden the discussion, have you guys considered adopting a model similar to [Velero's `install` command](https://velero.io/docs/master/customize-installation/#generate-yaml-only)? The Go code outputs the k8s manifests directly from the CLI or can talk to k8s natively. This is a nice way to provide CLI flags/options that influence the output. For us, that may include simple tweaks like opting for JSON output and changing the namespace.

lmickh:
I would suggest using jsonnet or kustomize as the base so to speak and generate the `example/contour` and any future helm charts from those.  It is far easier to composite in my config layer for stuff like multi-tenant setup.

I've got the same pain point of either taking a reference yaml and figuring out what changed between releases then backporting into jsonnet or kustomize.  If a project only provides helm charts, I typically end up just using `helm template` to get yaml that I can composite other things on top of.

From my experience it is way easier to go from compositing (jsonnet/kustomize) -> templates (helm) than the other way around.

bgagnon:
> From my experience it is way easier to go from compositing (jsonnet/kustomize) -> templates (helm) than the other way around.

Do you have any concrete examples of generating Helm charts from other sources? I'm curious to hear what the strategy is to inject Go template placeholders in the JSON/YAML output.

jpeach:
> > From my experience it is way easier to go from compositing (jsonnet/kustomize) -> templates (helm) than the other way around.
> 
> Do you have any concrete examples of generating Helm charts from other sources? I'm curious to hear what the strategy is to inject Go template placeholders in the JSON/YAML output.

That's a great question. Looking around I see a number of examples of applying kustomize to helm charts, but not the inverse. Probably one way to approach it is to use kustomize to dump the yaml with only very light templating and then rebase the chart on top of that. I don't know if that would be a sustainable approach, but it probably depends on how flexible the chart needs to be and how heavily templated it gets.

/cc @pickledrick

lmickh:
> > Do you have any concrete examples of generating Helm charts from other sources? I'm curious to hear what the strategy is to inject Go template placeholders in the JSON/YAML output.
> 
> That's a great question. Looking around I see a number of examples of applying kustomize to helm charts, but not the inverse. Probably one way to approach it is to use kustomize to dump the yaml with only very light templating and then rebase the chart on top of that. 

I'll see if I can dig one up, but I don't have an open source one readily available off the top of my head. But the gist of it is as mentioned.  Start with a reference output and then basically replace default yaml lines with the go template snippets.  It is the same process you would go through to write a helm chart from an example manifest in the first place.  Replace `name: contour` with `name: {{ .Chart.Name }}`. Unless the values your are templating out change frequently, it doesn't require a lot of continued work.

By comparison taking a chart that had template logic to handle a multi-tenant setup would be very hard to reduce back to a kustomize base with internal/external overlays.  At best it would probably be  charts of chart with different values files.

If you got a chart that is trying to tie together a large number of resources/sub-charts and bake in logic then I would imagine it becomes harder.  But at that point why not just move to an operator instead?  I don't really think that would apply to this project though.  Even the multi-tenant setup should not be that complicated.

tsaarni:
I checked this PR from the perspective of #2149, that is, replacing `certgen` with something else.
  
Having `certgen` as a separate component makes it easy to leave everything related away. User needs to write a new version of "base" to do that, which can be ok, though it is a bit more complicated than it ideally should be.

Users may need to do some other minor tweaks. For example, typically the certificate and key files in Secrets would be named differently from environment to another like discussed in  https://github.com/projectcontour/contour/issues/2149#issuecomment-598065053.
This impacts the command line parameter for `contour serve` and `contour bootstrap`.
Another example could be the forthcoming fallback certificate command line parameter.
These are something that might be more suited for a templating system like helm, where the `values.yaml` acts as the user facing API towards the templates.

At my company my colleagues use helm and in practice they need its full templating capability (sometimes it even falls short). That said, I personally like the simplicity of kustomize a lot more and dislike the complexity of helm charts.

stevesloka:
I think if we can keep the current `examples/contour` structure as-is, but use Kustomize to generate that's fine. This way we get the flexibility of Kustomize but keep current users + simplicity as an option.

jpeach:
> I think if we can keep the current `examples/contour` structure as-is, but use Kustomize to generate that's fine. This way we get the flexibility of Kustomize but keep current users + simplicity as an option.

I don't think that's really feasible. The example isn't well structured to be able to let us break it apart into the underlying components. I think that we need to start with components, then build up a deployment from that.

jkirklan:
This is an issue for us..  The helm chart in the helm github is using the v1 API.  We need supported mechanisms for installing including kustomize, helm v3, etc.  We need to avoid helm v2 due to multiple issues.

jpeach:
Closing since the underlying YAML has changed and this branch needs to be rebuilt from scratch.

PR #2477, opened 2020-04-28
internal: implement fallback certificate

### LINKED ISSUE #1503: Support Virtual Host Based Routing for Non-SNI HTTPS Clients
We need to do virtual host based routing, so that the TLS request is routed to the appropriate service, based on the hostname specified in the HTTPS’s “HOST” header. As HOST Header is encrypted during TLS handshake, it can’t be used for virtual host based routing unless client sends HTTPS request specifying hostname using the SNI or the request is first decrypted using a default TLS certificate.
Here’s the problem, we need to support clients which may not be using SNI. So, when a HTTPS request is coming, Contour needs to first decrypt the request using a default TLS certificate and then based on the HOST header, route it to appropriate service.  As of now, Contour only provides certificate at virtual host level in an IngressRoute and there is no way to define a default TLS certificate in Contour . Many Ingress controllers allow you to specify the default TLS cert but Contour doesn’t support it. The solution provided in #410 i.e. certificate delegation can't be used in this case.

stevesloka:
Fixes #1503 by adding a `FallbackCertificate` so that clients who do not offer SNI will still match a filterchain in Envoy. Envoy will present a default/fallback certificate to the client. 

Note: One side-effect of this is when you have 2 fqdn's in a cluster, one's enabled for fallback and the other isn't, requests will only route without SNI for the enabled cluster, however, if you request a cert for the non-enabled fqdn, the fallback cert will still be presented. The normal result is an error from Envoy: `curl: (35) LibreSSL SSL_connect: SSL_ERROR_SYSCALL in connection to fdqn:443` 

Docs to come in a new PR. 

---
This change is [Reviewable](https://reviewable.io/reviews/projectcontour/contour/2477)

### DIFF
--- a/cmd/contour/servecontext.go
+++ b/cmd/contour/servecontext.go
@@ -25,6 +25,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/projectcontour/contour/internal/k8s"
+
 	"github.com/projectcontour/contour/internal/contour"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
@@ -172,6 +174,38 @@ func newServeContext() *serveContext {
 // TLSConfig holds configuration file TLS configuration details.
 type TLSConfig struct {
 	MinimumProtocolVersion string `yaml:"minimum-protocol-version"`
+
+	// FallbackCertificate defines the namespace/name of the Kubernetes secret to
+	// use as fallback when a non-SNI request is received.
+	FallbackCertificate FallbackCertificate `yaml:"fallback-certificate,omitempty"`
+}
+
+// FallbackCertificate defines the namespace/name of the Kubernetes secret to
+// use as fallback when a non-SNI request is received.
+type FallbackCertificate struct {
+	Name      string `yaml:"name"`
+	Namespace string `yaml:"namespace"`
+}
+
+func (ctx *serveContext) fallbackCertificate() (*k8s.FullName, error) {
+	if len(strings.TrimSpace(ctx.TLSConfig.FallbackCertificate.Name)) == 0 && len(strings.TrimSpace(ctx.TLSConfig.FallbackCertificate.Namespace)) == 0 {
+		return nil, nil
+	}
+
+	// Validate namespace is defined
+	if len(strings.TrimSpace(ctx.TLSConfig.FallbackCertificate.Namespace)) == 0 {
+		return nil, errors.New("namespace must be defined")
+	}
+
+	// Validate name is defined
+	if len(strings.TrimSpace(ctx.TLSConfig.FallbackCertificate.Name)) == 0 {
+		return nil, errors.New("name must be defined")
+	}
+
+	return &k8s.FullName{
+		Name:      ctx.TLSConfig.FallbackCertificate.Name,
+		Namespace: ctx.TLSConfig.FallbackCertificate.Namespace,
+	}, nil
 }
 
 // LeaderElectionConfig holds the config bits for leader election inside the
@@ -272,6 +306,7 @@ func (ctx *serveContext) verifyTLSFlags() error {
 	if !(ctx.caFile != "" && ctx.contourCert != "" && ctx.contourKey != "") {
 		return errors.New("you must supply all three TLS parameters - --contour-cafile, --contour-cert-file, --contour-key-file, or none of them")
 	}
+
 	return nil
 }
 
--- a/examples/contour/01-crds.yaml
+++ b/examples/contour/01-crds.yaml
@@ -1135,6 +1135,11 @@ spec:
                       required:
                       - caSecret
                       type: object
+                    enableFallbackCertificate:
+                      description: EnableFallbackCertificate defines if the vhost
+                        should allow a default certificate to be applied which handles
+                        all requests which don't match the SNI defined in this vhost.
+                      type: boolean
                     minimumProtocolVersion:
                       description: Minimum TLS version this vhost should negotiate
                       type: string
--- a/internal/contour/listener.go
+++ b/internal/contour/listener.go
@@ -33,6 +33,7 @@ import (
 
 const (
 	ENVOY_HTTP_LISTENER            = "ingress_http"
+	ENVOY_FALLBACK_ROUTECONFIG     = "ingress_fallbackcert"
 	ENVOY_HTTPS_LISTENER           = "ingress_https"
 	DEFAULT_HTTP_ACCESS_LOG        = "/dev/stdout"
 	DEFAULT_HTTP_LISTENER_ADDRESS  = "0.0.0.0"
@@ -294,8 +295,8 @@ func visitListeners(root dag.Vertex, lvc *ListenerVisitorConfig) map[string]*v2.
 
 	lv.visit(root)
 
-	// Add a listener if there are vhosts bound to http.
 	if lv.http {
+		// Add a listener if there are vhosts bound to http.
 		cm := envoy.HTTPConnectionManagerBuilder().
 			RouteConfigName(ENVOY_HTTP_LISTENER).
 			MetricsPrefix(ENVOY_HTTP_LISTENER).
@@ -310,7 +311,6 @@ func visitListeners(root dag.Vertex, lvc *ListenerVisitorConfig) map[string]*v2.
 			proxyProtocol(lvc.UseProxyProto),
 			cm,
 		)
-
 	}
 
 	// Remove the https listener if there are no vhosts bound to it.
@@ -402,6 +402,32 @@ func (v *listenerVisitor) visit(vertex dag.Vertex) {
 
 		v.listeners[ENVOY_HTTPS_LISTENER].FilterChains = append(v.listeners[ENVOY_HTTPS_LISTENER].FilterChains,
 			envoy.FilterChainTLS(vh.VirtualHost.Name, downstreamTLS, filters))
+
+		// If this VirtualHost has enabled the fallback certificate then set a default FilterChain which will allow
+		// routes with this vhost to accept non SNI TLS requests
+		if vh.FallbackCertificate != nil && !envoy.ContainsFallbackFilterChain(v.listeners[ENVOY_HTTPS_LISTENER].FilterChains) {
+
+			// Construct the downstreamTLSContext passing the configured fallbackCertificate. The TLS minProtocolVersion will use
+			// the value defined in the Contour Configuration file if defined.
+			downstreamTLS = envoy.DownstreamTLSContext(
+				vh.FallbackCertificate,
+				v.ListenerVisitorConfig.minProtoVersion(),
+				vh.DownstreamValidation,
+				alpnProtos...)
+
+			// Default filter chain
+			filters = envoy.Filters(
+				envoy.HTTPConnectionManagerBuilder().
+					RouteConfigName(ENVOY_FALLBACK_ROUTECONFIG).
+					MetricsPrefix(ENVOY_HTTPS_LISTENER).
+					AccessLoggers(v.ListenerVisitorConfig.newSecureAccessLog()).
+					RequestTimeout(v.ListenerVisitorConfig.requestTimeout()).
+					Get(),
+			)
+
+			v.listeners[ENVOY_HTTPS_LISTENER].FilterChains = append(v.listeners[ENVOY_HTTPS_LISTENER].FilterChains,
+				envoy.FilterChainTLSFallback(downstreamTLS, filters))
+		}
 
 	default:
 		// recurse
--- a/internal/contour/listener_test.go
+++ b/internal/contour/listener_test.go
@@ -17,12 +17,15 @@ import (
 	"path"
 	"testing"
 
+	"github.com/projectcontour/contour/internal/k8s"
+
 	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
 	envoy_api_v2_auth "github.com/envoyproxy/go-control-plane/envoy/api/v2/auth"
 	envoy_api_v2_core "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
 	envoy_api_v2_listener "github.com/envoyproxy/go-control-plane/envoy/api/v2/listener"
 	"github.com/golang/protobuf/proto"
 	ingressroutev1 "github.com/projectcontour/contour/apis/contour/v1beta1"
+	projcontour "github.com/projectcontour/contour/apis/projectcontour/v1"
 	"github.com/projectcontour/contour/internal/assert"
 	"github.com/projectcontour/contour/internal/dag"
 	"github.com/projectcontour/contour/internal/envoy"
@@ -132,10 +135,17 @@ func TestListenerVisit(t *testing.T) {
 			Get()
 	}
 
+	fallbackCertFilter := envoy.HTTPConnectionManagerBuilder().
+		MetricsPrefix(ENVOY_HTTPS_LISTENER).
+		RouteConfigName(ENVOY_FALLBACK_ROUTECONFIG).
+		AccessLoggers(envoy.FileAccessLogEnvoy(DEFAULT_HTTP_ACCESS_LOG)).
+		Get()
+
 	tests := map[string]struct {
 		ListenerVisitorConfig
-		objs []interface{}
-		want map[string]*v2.Listener
+		fallbackCertificate *k8s.FullName
+		objs                []interface{}
+		want                map[string]*v2.Listener
 	}{
 		"nothing": {
 			objs: nil,
@@ -273,7 +283,7 @@ func TestListenerVisit(t *testing.T) {
 					FilterChainMatch: &envoy_api_v2_listener.FilterChainMatch{
 						ServerNames: []string{"whatever.example.com"},
 					},
-					TransportSocket: transportSocket(envoy_api_v2_auth.TlsParameters_TLSv1_1, "h2", "http/1.1"),
+					TransportSocket: transportSocket("secret", envoy_api_v2_auth.TlsParameters_TLSv1_1, "h2", "http/1.1"),
 					Filters:         envoy.Filters(httpsFilterFor("whatever.example.com")),
 				}},
 			}),
@@ -360,13 +370,13 @@ func TestListenerVisit(t *testing.T) {
 					FilterChainMatch: &envoy_api_v2_listener.FilterChainMatch{
 						ServerNames: []string{"sortedfirst.example.com"},
 					},
-					TransportSocket: transportSocket(envoy_api_v2_auth.TlsParameters_TLSv1_1, "h2", "http/1.1"),
+					TransportSocket: transportSocket("secret", envoy_api_v2_auth.TlsParameters_TLSv1_1, "h2", "http/1.1"),
 					Filters:         envoy.Filters(httpsFilterFor("sortedfirst.example.com")),
 				}, {
 					FilterChainMatch: &envoy_api_v2_listener.FilterChainMatch{
 						ServerNames: []string{"sortedsecond.example.com"},
 					},
-					TransportSocket: transportSocket(envoy_api_v2_auth.TlsParameters_TLSv1_1, "h2", "http/1.1"),
+					TransportSocket: transportSocket("secret", envoy_api_v2_auth.TlsParameters_TLSv1_1, "h2", "http/1.1"),
 					Filters:         envoy.Filters(httpsFilterFor("sortedsecond.example.com")),
 				}},
 			}),
@@ -482,7 +492,7 @@ func TestListenerVisit(t *testing.T) {
 					FilterChainMatch: &envoy_api_v2_listener.FilterChainMatch{
 						ServerNames: []string{"www.example.com"},
 					},
-					TransportSocket: transportSocket(envoy_api_v2_auth.TlsParameters_TLSv1_1, "h2", "http/1.1"),
+					TransportSocket: transportSocket("secret", envoy_api_v2_auth.TlsParameters_TLSv1_1, "h2", "http/1.1"),
 					Filters:         envoy.Filters(httpsFilterFor("www.example.com")),
 				}},
 				ListenerFilters: envoy.ListenerFilters(
@@ -563,7 +573,7 @@ func TestListenerVisit(t *testing.T) {
 					FilterChainMatch: &envoy_api_v2_listener.FilterChainMatch{
 						ServerNames: []string{"www.example.com"},
 					},
-					TransportSocket: transportSocket(envoy_api_v2_auth.TlsParameters_TLSv1_1, "h2", "http/1.1"),
+					TransportSocket: transportSocket("secret", envoy_api_v2_auth.TlsParameters_TLSv1_1, "h2", "http/1.1"),
 					Filters:         envoy.Filters(httpsFilterFor("www.example.com")),
 				}},
 				ListenerFilters: envoy.ListenerFilters(
@@ -637,7 +647,7 @@ func TestListenerVisit(t *testing.T) {
 					FilterChainMatch: &envoy_api_v2_listener.FilterChainMatch{
 						ServerNames: []string{"whatever.example.com"},
 					},
-					TransportSocket: transportSocket(envoy_api_v2_auth.TlsParameters_TLSv1_1, "h2", "http/1.1"),
+					TransportSocket: transportSocket("secret", envoy_api_v2_auth.TlsParameters_TLSv1_1, "h2", "http/1.1"),
 					Filters:         envoy.Filters(httpsFilterFor("whatever.example.com")),
 				}},
 			}),
@@ -709,7 +719,7 @@ func TestListenerVisit(t *testing.T) {
 					FilterChainMatch: &envoy_api_v2_listener.FilterChainMatch{
 						ServerNames: []string{"whatever.example.com"},
 					},
-					TransportSocket: transportSocket(envoy_api_v2_auth.TlsParameters_TLSv1_1, "h2", "http/1.1"),
+					TransportSocket: transportSocket("secret", envoy_api_v2_auth.TlsParameters_TLSv1_1, "h2", "http/1.1"),
 					Filters:         envoy.Filters(httpsFilterFor("whatever.example.com")),
 				}},
 			}),
@@ -778,7 +788,7 @@ func TestListenerVisit(t *testing.T) {
 					FilterChainMatch: &envoy_api_v2_listener.FilterChainMatch{
 						ServerNames: []string{"whatever.example.com"},
 					},
-					TransportSocket: transportSocket(envoy_api_v2_auth.TlsParameters_TLSv1_1, "h2", "http/1.1"),
+					TransportSocket: transportSocket("secret", envoy_api_v2_auth.TlsParameters_TLSv1_1, "h2", "http/1.1"),
 					Filters: envoy.Filters(envoy.HTTPConnectionManagerBuilder().
 						MetricsPrefix(ENVOY_HTTPS_LISTENER).
 						RouteConfigName(path.Join("https", "whatever.example.com")).
@@ -847,7 +857,7 @@ func TestListenerVisit(t *testing.T) {
 					FilterChainMatch: &envoy_api_v2_listener.FilterChainMatch{
 						ServerNames: []string{"whatever.example.com"},
 					},
-					TransportSocket: transportSocket(envoy_api_v2_auth.TlsParameters_TLSv1_3, "h2", "http/1.1"),
+					TransportSocket: transportSocket("secret", envoy_api_v2_auth.TlsParameters_TLSv1_3, "h2", "http/1.1"),
 					Filters:         envoy.Filters(httpsFilterFor("whatever.example.com")),
 				}},
 				ListenerFilters: envoy.ListenerFilters(
@@ -918,7 +928,7 @@ func TestListenerVisit(t *testing.T) {
 					FilterChainMatch: &envoy_api_v2_listener.FilterChainMatch{
 						ServerNames: []string{"whatever.example.com"},
 					},
-					TransportSocket: transportSocket(envoy_api_v2_auth.TlsParameters_TLSv1_3, "h2", "http/1.1"), // note, cannot downgrade from the configured version
+					TransportSocket: transportSocket("secret", envoy_api_v2_auth.TlsParameters_TLSv1_3, "h2", "http/1.1"), // note, cannot downgrade from the configured version
 					Filters:         envoy.Filters(httpsFilterFor("whatever.example.com")),
 				}},
 				ListenerFilters: envoy.ListenerFilters(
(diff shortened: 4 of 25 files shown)


internal/dag/builder.go
```diff
@@ -518,6 +521,18 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
 			svhost.Secret = sec
 			svhost.MinProtoVersion = annotation.MinProtoVersion(proxy.Spec.VirtualHost.TLS.MinimumProtocolVersion)
 
+			if proxy.Spec.VirtualHost.TLS.EnableFallbackCertificate && !isBlank(b.FallbackCertificate) {
+				secretName := splitSecret(b.FallbackCertificate, "projectcontour")
+				sec, err = b.lookupSecret(secretName, validSecret)
+				if err != nil {
+					sw.SetInvalid("FallbackCertificate requested, but Certificate Secret %q is invalid or missing", b.FallbackCertificate)
```
jpeach, line 528: This status message should follow the conventions of the other messages, so something like:
```Go
sw.SetInvalid("Spec.Virtualhost.TLS fallback certificate Secret %q is invalid: %s", b.FallbackCertificate, err)
```

jpeach, line 528: Here, and in the case above, we should put the resolved secret name (i.e. the result of `splitSecret`) in the status message, since that tells the user exactly which object had the problem.

internal/dag/builder.go
```diff
@@ -518,6 +521,18 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
 			svhost.Secret = sec
 			svhost.MinProtoVersion = annotation.MinProtoVersion(proxy.Spec.VirtualHost.TLS.MinimumProtocolVersion)
 
+			if proxy.Spec.VirtualHost.TLS.EnableFallbackCertificate && !isBlank(b.FallbackCertificate) {
+				secretName := splitSecret(b.FallbackCertificate, "projectcontour")
```
jpeach, line 525: What should the default namespace for this fallback secret be? I don't think we should hard-code it here, but if it isn't in a namespace we are already watching, then we won't be able to look it up.

Maybe we should *require* the namespace field on the fallback secret name, to ensure that only one is ever used? It's not clear to me what the right default namespace should be here.

We should also check the secret name flag on startup and ensure that it is consistent with the root namespaces handling of Secrets?

internal/dag/builder.go
```diff
@@ -518,6 +521,18 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
...
+				sec, err = b.lookupSecret(secretName, validSecret)
+				if err != nil {
+					sw.SetInvalid("FallbackCertificate requested, but Certificate Secret %q is invalid or missing", b.FallbackCertificate)
+					return
+				} else {
+					svhost.FallbackCertificate = sec
+					svhost.EnableFallbackCertificate = true
+				}
+			}
+
```
jpeach, line 535: I don't think the fallback certificate is compatible with client certificate auth, since when we have multiple routing tables behind a single filter chain, we can't enforce which routes get client auth and which don't. If we solve that problem, we could lift the restriction in the future.

internal/dag/dag.go
```diff
@@ -34,6 +34,8 @@ type DAG struct {
 
 	// status computed while building this dag.
 	statuses map[k8s.FullName]Status
+
+	FallbackCertificate *Secret
```
jpeach, line 38: Is this used anywhere?

internal/dag/dag.go
```diff
@@ -223,6 +225,10 @@ type VirtualHost struct {
 	Name string
 
 	routes map[string]*Route
+
+	// EnableFallbackCertificate defines if the vhost should allow a default certificate to
+	// be applied which handles all requests which don't match the SNI defined in this vhost
+	EnableFallbackCertificate bool
```
jpeach, line 231: As per comment below, perhaps we can remove this field?

internal/sorter/sorter.go
```diff
@@ -176,10 +176,15 @@ type filterChainSorter []*envoy_api_v2_listener.FilterChain
...
-	return s[i].FilterChainMatch.ServerNames[0] < s[j].FilterChainMatch.ServerNames[0]
+
+	// (SAS) Default
+	if len(s[i].FilterChainMatch.ServerNames) > 0 && len(s[j].FilterChainMatch.ServerNames) > 0 {
+		// The ServerNames field will only ever have a single entry
+		// in our FilterChain config, so it's okay to only sort
+		// on the first slice entry.
+		return s[i].FilterChainMatch.ServerNames[0] < s[j].FilterChainMatch.ServerNames[0]
+	}
+	return false
```
jpeach, line 187: I think that this needs to sort filter chains with no ServerNames to the end? We should be able to add a test for that case too.

internal/contour/listener.go
```diff
@@ -276,7 +277,7 @@ type listenerVisitor struct {
 	*ListenerVisitorConfig
 
 	listeners map[string]*v2.Listener
-	http      bool // at least one dag.VirtualHost encountered
+	http      bool
```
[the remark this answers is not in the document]
stevesloka: that's a bad re-refactor (if that's a thing). I had removed the `http` bit and made it something else once.

internal/dag/builder.go
```diff
@@ -518,6 +521,18 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
 			svhost.Secret = sec
 			svhost.MinProtoVersion = annotation.MinProtoVersion(proxy.Spec.VirtualHost.TLS.MinimumProtocolVersion)
 
+			if proxy.Spec.VirtualHost.TLS.EnableFallbackCertificate && !isBlank(b.FallbackCertificate) {
```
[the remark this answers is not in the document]
stevesloka: This would emit a warning for each dag rebuild. We actually don't have any logging in the builder, so we'd need to introduce that which is fine, but we'll spam the logs with this warning.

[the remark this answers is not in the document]
youngnick: If a proxy sets `EnableFallbackCertificate` but the fallback cert isn't configured, that should be an invalid proxy, and we should use the status update to communicate that.

internal/dag/builder.go
youngnick: I think we can do it with a string that *may* include a namespace name `<namespace>/secretname` or `secretname` would both be valid. Agreed that the secret name flag should be checked on startup if we do this, to ensure that the namespace is one we are watching for Secrets. We may also need to add it if we aren't watching already.

internal/dag/builder.go
stevesloka: yup makes sense. I'll change it so you can't have both together in the same Root.

internal/dag/builder.go
stevesloka: I was changing this around a bit today (just haven't pushed). My intent was to make the namespace/name what was required. By doing that I can check that 1. users followed the right pattern, and 2. remove the need for a default namespace later on. 

If we allow just a `name`, then it would have to be in the context of where Contour is deployed and we could use the downward API logic that already exists to use as the default.

jpeach: What should the namespace be if it secret name isn't fully qualified?

jpeach: > I was changing this around a bit today (just haven't pushed). My intent was to make the namespace/name what was required. By doing that I can check that 1. users followed the right pattern, and 2. remove the need for a default namespace later on.

Yup, I agree that requiring a fully qualified name makes sense here.

youngnick: I'm okay with requiring a namespace, it would be a little complicated to explain how it interacted with the downward API bits.

stevesloka: Right now I have it as a `namespace/name` and Contour checks for that on startup. This is passed in via an arg on `contour serve` and in the config file. 

One way to maybe make this more clear would be to remove the ability to set the command like arg, and only use the config file. The file could extend another struct to have `fallbackCertificate` and then ask for the name and namespace in separate fields. 

Example:

```yaml
tls:
    minimum-protocol-version: "1.1"
    fallbackCertificate:
      name: "secret"
      namespace: "userns"
```

youngnick: I like this. I'd like to move away from adding cmdline flags, and start pushing people towards using the config file only.

internal/dag/dag.go
```diff
@@ -261,6 +267,9 @@ type SecureVirtualHost struct {
 	// The cert and key for this host.
 	Secret *Secret
 
+	// FallbackCertificate
+	FallbackCertificate *Secret
```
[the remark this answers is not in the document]
stevesloka: That was the problem I was mentioning on Slack the other day. There isn't a good way to inject this into the listener config which is why I drove it from the dag rebuild.

internal/dag/builder.go
```diff
@@ -50,6 +50,8 @@ type Builder struct {
 
 	orphaned map[k8s.FullName]bool
 
+	FallbackCertificate string
```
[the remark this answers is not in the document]
stevesloka: Since the split out to namespace/name via the config file, I made this a struct.

internal/dag/builder.go
stevesloka: The reason I left it out is that it leaks the namespace/name of the secret used.  Since it's a wildcard, I didn't know if we wanted to keep that from users.

internal/envoy/listener.go
```diff
@@ -300,6 +300,21 @@ func FilterChainTLS(domain string, downstream *envoy_api_v2_auth.DownstreamTlsCo
 	return fc
 }
 
+// FilterChainTLSDomain returns a TLS enabled envoy_api_v2_listener.FilterChain
+func FilterChainTLSNoDomain(downstream *envoy_api_v2_auth.DownstreamTlsContext, filters []*envoy_api_v2_listener.Filter) *envoy_api_v2_listener.FilterChain {
```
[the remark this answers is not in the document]
stevesloka: Doesn't matter to me, doing that reverses what it does today, but might be more clear.

internal/dag/dag.go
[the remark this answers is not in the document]
stevesloka: yup we can do that

internal/contour/listener.go
```diff
@@ -403,6 +403,36 @@ func (v *listenerVisitor) visit(vertex dag.Vertex) {
...
+						RouteConfigName(ENVOY_FALLBACKCERT).
+						MetricsPrefix(ENVOY_HTTPS_LISTENER).
+						AccessLoggers(v.ListenerVisitorConfig.newSecureAccessLog()).
+						RequestTimeout(v.ListenerVisitorConfig.requestTimeout()).
+						Get(),
+				)
+
+				v.listeners[ENVOY_HTTPS_LISTENER].FilterChains = append(v.listeners[ENVOY_HTTPS_LISTENER].FilterChains,
+					envoy.FilterChainTLSNoDomain(downstreamTLS, filters))
+			}
```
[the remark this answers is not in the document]
stevesloka: I added a check to see if the filterchain is already added.

internal/featuretests/fallbackcert_test.go
```diff
@@ -0,0 +1,164 @@
...
+			},
+			Routes: []projcontour.Route{{
+				Services: []projcontour.Service{{
+					Name: s1.Name,
+					Port: 80,
+				}},
+			}},
+		},
+	}
+	rh.OnAdd(proxy1)
```
jpeach, line 90: Can you add a test that adds multiple fallback proxies?

cmd/contour/servecontext.go
```diff
@@ -172,6 +172,41 @@ func newServeContext() *serveContext {
...
+	// to use as fallback when a non-SNI request is received
+	Name string `yaml:"name"`
+
+	// Namespace defines the namespace of the Kubernetes secret containing certificate
+	// to use as fallback when a non-SNI request is received
+	Namespace string `yaml:"namespace"`
+}
+
+func (ctx *serveContext) fallbackCertificate() (*FallbackCertificate, error) {
+	if len(ctx.TLSConfig.FallbackCertificate.Name) == 0 && len(ctx.TLSConfig.FallbackCertificate.Namespace) == 0 {
```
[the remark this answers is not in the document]
stevesloka: The downside of re-using `k8s.FullName` is that it has a default namespace of `default` which doesn't work with fallback well since I don't want a default namespace.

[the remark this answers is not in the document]
stevesloka: Also in servecontext.go this is looking at the struct type `main.FallbackCertificate` since it has the yaml parsing fields on it. I don't think we should bake those into `k8s.FullName`.

cmd/contour/servecontext.go
```diff
@@ -172,6 +172,41 @@ func newServeContext() *serveContext {
 // TLSConfig holds configuration file TLS configuration details.
 type TLSConfig struct {
 	MinimumProtocolVersion string `yaml:"minimum-protocol-version"`
+
+	// FallbackCertificate defines the namespace/name of the Kubernetes secret to
+	// use as fallback when a non-SNI request is received
```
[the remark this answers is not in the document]
stevesloka: I really dislike periods at the ends of comments but will keep for consistency.

internal/contour/listener.go
```diff
@@ -403,8 +403,44 @@ func (v *listenerVisitor) visit(vertex dag.Vertex) {
...
+				envoy.HTTPConnectionManagerBuilder().
+					RouteConfigName(ENVOY_FALLBACK_CERTIFICATE).
+					MetricsPrefix(ENVOY_HTTPS_LISTENER).
+					AccessLoggers(v.ListenerVisitorConfig.newSecureAccessLog()).
+					RequestTimeout(v.ListenerVisitorConfig.requestTimeout()).
+					Get(),
+			)
+
+			v.listeners[ENVOY_HTTPS_LISTENER].FilterChains = append(v.listeners[ENVOY_HTTPS_LISTENER].FilterChains,
+				envoy.FilterChainTLSNoDomain(downstreamTLS, filters))
```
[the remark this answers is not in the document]
stevesloka: yup, the filter changes will place the servernames above these without.

internal/contour/listener.go
```diff
@@ -403,8 +403,44 @@ func (v *listenerVisitor) visit(vertex dag.Vertex) {
 		v.listeners[ENVOY_HTTPS_LISTENER].FilterChains = append(v.listeners[ENVOY_HTTPS_LISTENER].FilterChains,
 			envoy.FilterChainTLS(vh.VirtualHost.Name, downstreamTLS, filters))
 
+		// If this VirtualHost has enabled the fallback certificate then set a default FilterChain which will allow
+		// routes with this vhost to accept non SNI TLS requests
+		if vh.FallbackCertificate != nil && !containsFallbackFilterChain("fallback-certificate", v.listeners[ENVOY_HTTPS_LISTENER].FilterChains) {
```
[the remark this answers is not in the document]
stevesloka: Yup we could do that. My struggle was to push the cert into the `ListenerVistitorConfig`. We don't need the actual cert, just the name that matches SDS, but the cert is needed to generate the name.

internal/contour/listener.go
```diff
@@ -403,8 +403,44 @@ func (v *listenerVisitor) visit(vertex dag.Vertex) {
...
+
+			// Choose the higher of the configured or requested TLS version.
+			vers := max(v.ListenerVisitorConfig.minProtoVersion(), vh.MinProtoVersion)
```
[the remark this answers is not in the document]
stevesloka: yeah good point. I guess we could wire up a `minimumProtocolVersion` to the `fallbackCertificate` spec in the config file.

internal/dag/builder.go
```diff
@@ -518,6 +521,25 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
...
+			if b.FallbackCertificate == nil {
+				if proxy.Spec.VirtualHost.TLS.EnableFallbackCertificate {
+					sw.SetInvalid("Spec.Virtualhost.TLS.FallbackCertificate requested but Certificate Secret not configured in Contour configuration file")
+					return
+				}
+			} else {
+				if proxy.Spec.VirtualHost.TLS.EnableFallbackCertificate {
+					fmt.Println(b.FallbackCertificate)
+					sec, err = b.lookupSecret(k8s.FullName{Name: b.FallbackCertificate.Name, Namespace: b.FallbackCertificate.Namespace}, validSecret)
+					if err != nil {
```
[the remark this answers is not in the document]
stevesloka: Sorry not following what you are asking here.

cmd/contour/servecontext.go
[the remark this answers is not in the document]
jpeach: https://godoc.org/gopkg.in/yaml.v2#Marshal

> Struct fields are only marshalled if they are exported (have an upper case first letter), and are marshalled using the field name lowercased as the default key.

So the yaml package will do what you want without any need for adding yaml tags.

I don't think that the default namespace will be an issue with `k8s.FullName` since you are always requiring a namespace to be specified here. `FullName` only prints "default" when the default (empty) namespace is actually used.

`FullName` is an open struct so you can do the same checks on it that you are doing now. Duplicating the type then means that we need to manually add consistent `String()` formatting, in addition to the overhead of having two types that do the same thing.

internal/dag/builder.go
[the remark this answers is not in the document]
jpeach: Sorry, I'm asking whether we need to check certificate delegation here?

cmd/contour/servecontext.go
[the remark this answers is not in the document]
jpeach: Oh, that's interesting! I have always added them, and usually ask for it in code reviews. I'm regarding the comment as a full sentence, so it seems to me that comments should always be capitalized and punctuated. Why do you dislike adding the period?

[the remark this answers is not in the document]
stevesloka: They just don't feel like sentences to me. Nothing super strong, just feels werid, but no big deal. =)

internal/dag/builder.go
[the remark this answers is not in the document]
stevesloka: Ohh ok interesting. Do you think certificate delegation matters?

cmd/contour/servecontext.go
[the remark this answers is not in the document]
stevesloka: What made me pause with that struct is that if you don't set a namespace, but call `String()` it uses `default` as it's default value. So if I added an `IsEmpty()` which looked for both name/namespace were empty, then if you called `string()` it would print out a namespace. 

In either case, if we add mintlsprotocol then we'll need a new struct to hold that.

internal/dag/builder.go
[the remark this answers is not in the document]
jpeach: > Ohh ok interesting. Do you think certificate delegation matters?

I'm not sure, I can probably go either way. Doing certificate delegation here makes it consistent and easier to explain. However, we could also consider the default cert as a special case that implies delegation as a practical matter.

If we require delegation, then we could always raise that restriction later.

internal/contour/listener.go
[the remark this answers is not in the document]
jpeach: What if we just use the version from the ListenerVisitorConfig?

internal/dag/builder.go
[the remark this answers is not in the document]
youngnick: I think a default certificate should be implicitly delegated everywhere.

examples/contour/01-crds.yaml
```diff
@@ -328,6 +328,11 @@ spec:
                       required:
                       - caSecret
                       type: object
+                    enableFallbackCertificate:
```
ravilr, line 331: is this v1beta1.ingressRoutes changing expected? only httpProxy cr right?

internal/contour/listener.go
```diff
@@ -403,6 +403,33 @@ func (v *listenerVisitor) visit(vertex dag.Vertex) {
...
 
+		// If this VirtualHost has enabled the fallback certificate then set a default FilterChain which will allow
+		// routes with this vhost to accept non SNI TLS requests
+		if vh.FallbackCertificate != nil && !envoy.ContainsFallbackFilterChain(v.listeners[ENVOY_HTTPS_LISTENER].FilterChains) {
+
+			// Choose the higher of the configured or requested TLS version.
+			vers := max(v.ListenerVisitorConfig.minProtoVersion(), vh.MinProtoVersion)
+
+			downstreamTLS = envoy.DownstreamTLSContext(
+				vh.FallbackCertificate,
```
ravilr, line 414: this needs to be added to secret cache here https://github.com/projectcontour/contour/blob/ea65fa53a7d3feae99713bfb253a04b1671630b8/internal/contour/secret.go#L88
for it to be available through SDS cache ?

internal/dag/builder.go
```diff
@@ -518,6 +521,21 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
 			svhost.Secret = sec
 			svhost.MinProtoVersion = annotation.MinProtoVersion(proxy.Spec.VirtualHost.TLS.MinimumProtocolVersion)
 
+			// If FallbackCertificate is enabled, but no cert passed, set error
+			if proxy.Spec.VirtualHost.TLS.EnableFallbackCertificate {
```
ravilr, line 525: Is it the expectation that virtual host which opted for fallback certificate in HttpProxy has to set proxy.Spec.VirtualHost.TLS.secretName also ?

stevesloka: Currently, the design is you still need to specify the secretname for each vhost to enable the fallbackCertificate. With `TLSCertificateDelegation`, you're able to make a "default" certificate available to specific, or all namespaces, so that functionality which exists in other controllers can still be utilized here with Contour. 

The main goal of the fallback certificate is to handle the use-case where requests are received that do not have SNI on them.

internal/contour/listener.go
stevesloka: the secret should already be in the cache since it's in Kubernetes.

internal/dag/builder.go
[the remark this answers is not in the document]
stevesloka: I updated the design to add the cert delegation piece.

examples/contour/01-crds.yaml
stevesloka: oh no, that's not expected! Must be something with the generators.

youngnick: Some of the IngressRoute stuff uses structs shared from HTTPProxy; is this the case here?

stevesloka: I opened https://github.com/projectcontour/contour/pull/2511 to resolve.

stevesloka:
@jpeach @youngnick I think this is good for a first step. The remaining bits are to do the checks on if the delegated certificate is set correctly & docs. I think the review might be easier if we can merge this then work on that in a separate PR.

examples/contour/01-crds.yaml
jpeach: 👍

internal/contour/listener.go
```diff
@@ -33,6 +33,7 @@ import (
 
 const (
 	ENVOY_HTTP_LISTENER            = "ingress_http"
+	ENVOY_FALLBACKCERT             = "ingress_fallbackcert"
```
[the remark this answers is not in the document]
jpeach: Can we please name this `ENVOY_FALLBACK_ROUTE_CONFIG`? Any reader seeing the constant name `ENVOY_FALLBACK_CERTIFICATE` will think that this refers to a certificate not to a route config.

jpeach:
@stevesloka I tried using `reviewable.io` for this review and I didn't realize how it would come out on this side of the review. Sorry about that; I'll be happy to get with you and clean this up :(

I think we are in good shape here, all the above is just various nits, and we can do the last remaining functional stuff in followup PRs. If you merge, could you please update the commit message to follow the contributing conventions (i.e. the format you've used for the PR), otherwise I can just edit it in the squash UI.

stevesloka:
@jpeach rebase & updated. Reveiwable is going to take some time to get used to if we continue to use that going forward.

stevesloka:
@jpeach I reverted the `k8s.FullName` bit in `servecontext` since it wouldn't parse from the yaml file properly.

internal/contour/listener_test.go
```diff
@@ -111,6 +114,21 @@ func TestListenerCacheQuery(t *testing.T) {
...
+			}),
+			query: []string{ENVOY_FALLBACK_CERTIFICATE},
+			want: []proto.Message{
+				&v2.Listener{
+					Name:         ENVOY_FALLBACK_CERTIFICATE,
+					Address:      envoy.SocketAddress("0.0.0.0", 8080),
+					FilterChains: envoy.FilterChains(envoy.HTTPConnectionManager(ENVOY_FALLBACK_CERTIFICATE, envoy.FileAccessLogEnvoy(DEFAULT_HTTP_ACCESS_LOG), 0)),
+				},
+			},
+		},
```
jpeach, line 131: We should remove this case since it doesn't add any extra coverage.

internal/contour/listener.go
```diff
@@ -33,6 +33,7 @@ import (
 
 const (
 	ENVOY_HTTP_LISTENER            = "ingress_http"
+	ENVOY_FALLBACK_CERTIFICATE     = "ingress_fallbackcert"
```
jpeach, line 36: Rename this `ENVOY_FALLBACK_ROUTE_CONFIG` since it is the name of a RouteConfg, not name of a certificate.

cmd/contour/servecontext.go
```diff
@@ -172,6 +174,38 @@ func newServeContext() *serveContext {
 // TLSConfig holds configuration file TLS configuration details.
 type TLSConfig struct {
 	MinimumProtocolVersion string `yaml:"minimum-protocol-version"`
+
+	// FallbackCertificate defines the namespace/name of the Kubernetes secret to
+	// use as fallback when a non-SNI request is received.
+	FallbackCertificate `yaml:"fallback-certificate,omitempty"`
```
jpeach, line 180:
```suggestion
	FallbackCertificate FallbackCertificate  `yaml:"fallback-certificate,omitempty"`
```

We should name the field here, otherwise `Name` and `Namespace` are accessible as fields on a `TLSConfig` variable, which is confusing.

internal/dag/builder.go
```diff
@@ -518,6 +521,27 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
...
+			// Check if FallbackCertificate && ClientValidation are both enabled in the same vhost
+			if proxy.Spec.VirtualHost.TLS.EnableFallbackCertificate && tls.ClientValidation != nil {
+				sw.SetInvalid("Spec.Virtualhost.TLS.FallbackCertificate & Spec.Virtualhost.TLS.ClientValidation are incompatible together")
+				return
+			}
+
+			// If FallbackCertificate is enabled, but no cert passed, set error
+			if proxy.Spec.VirtualHost.TLS.EnableFallbackCertificate {
+				if b.FallbackCertificate == nil {
+					sw.SetInvalid("Spec.Virtualhost.TLS.FallbackCertificate requested but the fallback Certificate Secret is not configured in Contour configuration file")
```
jpeach, line 533:
```suggestion
					sw.SetInvalid("Spec.Virtualhost.TLS enabled fallback but the fallback certificate Secret is not configured in Contour configuration file")
```

for consistency with the status below

internal/dag/builder.go
```diff
@@ -518,6 +521,27 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
 			svhost.Secret = sec
 			svhost.MinProtoVersion = annotation.MinProtoVersion(proxy.Spec.VirtualHost.TLS.MinimumProtocolVersion)
 
+			// Check if FallbackCertificate && ClientValidation are both enabled in the same vhost
+			if proxy.Spec.VirtualHost.TLS.EnableFallbackCertificate && tls.ClientValidation != nil {
+				sw.SetInvalid("Spec.Virtualhost.TLS.FallbackCertificate & Spec.Virtualhost.TLS.ClientValidation are incompatible together")
```
jpeach, line 526:
```suggestion
				sw.SetInvalid("Spec.Virtualhost.TLS.EnableFallbackCertificate & Spec.Virtualhost.TLS.ClientValidation are incompatible together")
```

internal/dag/builder.go
```diff
@@ -518,6 +521,27 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
 			svhost.Secret = sec
 			svhost.MinProtoVersion = annotation.MinProtoVersion(proxy.Spec.VirtualHost.TLS.MinimumProtocolVersion)
 
+			// Check if FallbackCertificate && ClientValidation are both enabled in the same vhost
+			if proxy.Spec.VirtualHost.TLS.EnableFallbackCertificate && tls.ClientValidation != nil {
```
jpeach, line 525:
```suggestion
			if tls.EnableFallbackCertificate && tls.ClientValidation != nil {
```

internal/dag/builder.go
```diff
@@ -518,6 +521,27 @@ func (b *Builder) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
...
+				sw.SetInvalid("Spec.Virtualhost.TLS.FallbackCertificate & Spec.Virtualhost.TLS.ClientValidation are incompatible together")
+				return
+			}
+
+			// If FallbackCertificate is enabled, but no cert passed, set error
+			if proxy.Spec.VirtualHost.TLS.EnableFallbackCertificate {
```
jpeach, line 531:
```suggestion
			if tls.EnableFallbackCertificate {
```

internal/contour/listener.go
stevesloka: They should all be configs then, no?

e.g. `ENVOY_HTTP_LISTENER_CONFIG`?

internal/dag/builder.go
stevesloka: I changed to this to be consistent with the requests earlier: `Spec.Virtualhost.TLS fallback & client validation are incompatible together`

jpeach: Nice.

internal/contour/listener.go
jpeach: `ENVOY_HTTP_LISTENER` is the name of a listener object so that makes sense to my reading. Maybe `ENVOY_FALLBACK_ROUTECONFIG` because it is the name of a RouteConfig?

internal/dag/builder.go
jpeach: 👍

internal/contour/listener.go
stevesloka: But `ENVOY_HTTP_LISTENER` is also a route config same as the `FALLBACK`, it just so happens that that name is shared with the listener as well. 

**Example:**

```
"dynamic_route_configs": [
    {
     "version_info": "6",
     "route_config": {
      "@type": "type.googleapis.com/envoy.api.v2.RouteConfiguration",
      "name": "ingress_http",  <--------------  route config name
      "virtual_hosts": [
```

We should probably call it something like `ENVOY_TLSFALLBACK_ROUTECONFIG` to denote that it's a TLS fallback. It's also possible we'll have more in the future.

jpeach:
> On May 15, 2020, at 11:34 PM, Steve Sloka <notifications@github.com> wrote:
> 
> ENVOY_TLSFALLBACK_ROUTECONFIG

Yeh I like this name a lot!

Issue #2478, opened 2020-04-28
[Doc Fix]: AWS annotation for enabling proxy-protocol doesn't work


savithruml:
**What steps did you take and what happened:**
AWS EKS service annotation to enable proxy-protocol on NLB doesn't take any action. This has been confirmed with AWS support & here's an additional issue to back it up https://github.com/kubernetes/kubernetes/issues/57250.

**What did you expect to happen:**
The annotation `service.beta.kubernetes.io/aws-load-balancer-proxy-protocol: '*'` should enable proxy-protocol on NLB but it doesn't. Hence the documentation must be fixed on Contour's page which discusses the scenario https://projectcontour.io/guides/proxy-proto/

**Anything else you would like to add:**
Need to remove the content of this section [Enable PROXY protocol on your service in AWS](https://github.com/projectcontour/contour/blob/master/site/_guides/proxy-proto.md#enable-proxy-protocol-on-your-service-in-aws) & instead link it to [AWS documentation to enable proxy-protocol](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-target-groups.html#proxy-protocol) to manually enable/disable through AWS console.

**Environment:**

- Cloud provider or hardware configuration: AWS EKS

stevesloka:
Thanks for reporting this @savithruml! Would you like to contribute a PR to update? Happy to help and walk through this if you have questions.

savithruml:
> Thanks for reporting this @savithruml! Would you like to contribute a PR to update? Happy to help and walk through this if you have questions.

Sure @stevesloka, sent out a PR.

PR #2480, opened 2020-04-29
docs: Update content for proxy protocol support on AWS ELBs

savithruml:
Signed-off-by: savithruml <slokanath@salesforce.com>
Fixes: https://github.com/projectcontour/contour/issues/2478

### DIFF
--- a/site/_guides/proxy-proto.md
+++ b/site/_guides/proxy-proto.md
@@ -20,21 +20,9 @@ Your services should see the addresses in the `X-Forwarded-For` or `X-Envoy-Exte
 
 ## Enable PROXY protocol on your service in AWS
 
-To instruct EC2 to place the ELB into `tcp`+`PROXY` mode, add the following annotations to the `contour` Service:
+To instruct EC2 to place the ELB into `tcp`+`PROXY` mode, follow the steps mentioned in this [AWS documentation][2]
 
-```
-apiVersion: v1
-kind: Service
-metadata:
-  annotations:
-      service.beta.kubernetes.io/aws-load-balancer-backend-protocol: tcp
-      service.beta.kubernetes.io/aws-load-balancer-proxy-protocol: '*'
-    name: contour
-    namespace: projectcontour
-spec:
-  type: LoadBalancer
-...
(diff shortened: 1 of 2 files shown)


site/_guides/proxy-proto.md
```diff
@@ -20,21 +20,9 @@ Your services should see the addresses in the `X-Forwarded-For` or `X-Envoy-Exte
 
 ## Enable PROXY protocol on your service in AWS
 
-To instruct EC2 to place the ELB into `tcp`+`PROXY` mode, add the following annotations to the `contour` Service:
+To instruct EC2 to place the ELB into `tcp`+`PROXY` mode, follow the steps mentioned in this [AWS documentation][2]
```
youngnick, line 23: I haven't looked at this in a while, but doesn't this still work with the old ("classic") ELBs? Just not with NLBs? If that's the case, then this documentation should reflect that.

Just want to make sure that we're not leaving people who are using other configurations behind here.

youngnick: If you don't have time to check if it works with ELBs, that's fine, could you update so that it's clear this is talking about NLBs?

savithruml: @youngnick fixed it. Thanks!

youngnick:
The change looks great @savithruml, but we do need your commits signed with a [DCO signoff](https://github.com/projectcontour/contour/blob/master/CONTRIBUTING.md#dco-sign-off). Note that this needs to be an email address you're actually reachable on, noreply addresses aren't acceptable. You'll need to update each commit message with the DCO signoff as well.

Sorry not to have pointed this out before.

savithruml:
> The change looks great @savithruml, but we do need your commits signed with a [DCO signoff](https://github.com/projectcontour/contour/blob/master/CONTRIBUTING.md#dco-sign-off). Note that this needs to be an email address you're actually reachable on, noreply addresses aren't acceptable. You'll need to update each commit message with the DCO signoff as well.
> 
> Sorry not to have pointed this out before.

Thanks & done @youngnick

PR #2482, opened 2020-04-29
Pass `ingress.class` through to IngressStatusUpdater

### LINKED ISSUE #2481: Ingress class name is not passed to the ingress status updater correctly
Didn't use the default namespace (projectcontour), instead it was shortened to "contour".

When using non-default namespace, saw that all ingresses were being skipped by IngressStatusUpdater ie.
`level=debug msg="unmatched ingress class, skip status update" context=IngressStatusUpdater ingress-class=contour-intern
al name=dcp-es-internal namespace=elasticsearch`

Ingress Status field to be updated.

Can reference slack chat to get more background
https://kubernetes.slack.com/archives/C8XRH2R4J/p1588123185243300

**Environment:**

- Contour version: 1.4.0
- Kubernetes version: (use `kubectl version`): 1.17
- Kubernetes installer & version: kops 1.16.7
- OS (e.g. from `/etc/os-release`): debian ?

youngnick:
Fixes #2481.

This is a missed bit of wiring to make the Ingress status updating work in the case
that you have supplied an ingress class that is not "contour".

### DIFF
--- a/cmd/contour/ingressstatus.go
+++ b/cmd/contour/ingressstatus.go
@@ -38,10 +38,11 @@ import (
 //    is been received, operation restarts at step 3.
 // 5. If the worker is stopped, any existing informer is stopped before the worker stops.
 type loadBalancerStatusWriter struct {
-	log      logrus.FieldLogger
-	clients  *k8s.Clients
-	isLeader chan struct{}
-	lbStatus chan v1.LoadBalancerStatus
+	log          logrus.FieldLogger
+	clients      *k8s.Clients
+	isLeader     chan struct{}
+	lbStatus     chan v1.LoadBalancerStatus
+	ingressClass string
 }
 
 func (isw *loadBalancerStatusWriter) Start(stop <-chan struct{}) error {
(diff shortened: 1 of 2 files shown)


youngnick:
Can't test this without testing how command line arguments are propagated throughout `cmd/contour`, which we don't have a mechanism for yet. Guess this is another integration test in the making.

cmd/contour/ingressstatus.go
```diff
@@ -81,9 +82,10 @@ func (isw *loadBalancerStatusWriter) Start(stop <-chan struct{}) error {
 			inf := factory.Networking().V1beta1().Ingresses().Informer()
 			log := isw.log.WithField("context", "IngressStatusUpdater")
 			inf.AddEventHandler(&k8s.IngressStatusUpdater{
-				Client: isw.clients.ClientSet(),
-				Logger: log,
-				Status: lbs,
+				Client:       isw.clients.ClientSet(),
```
jpeach, line 85: Add another log field above on line 83 to capture the ingress class we are matching against (can't use the `ingress-class` field name since that's already in use.

youngnick: Good point, thanks.

stevesloka:
Should we release a v1.4.1 @youngnick?

Issue #1493, opened 2019-09-12
Requests can be misrouted due to HTTP/2 Connection Coalescing under certain scenarios


alexbrand:
tl;dr: Browsers reuse connections when using HTTP/2. **With certain IngressRoute configurations (Envoy configurations)**, requests can be routed to the wrong backend or produce puzzling 404s.

We believe the root cause to be a bug in Envoy that has an open issue here: https://github.com/envoyproxy/envoy/issues/6767

## What we observed

We noticed that requests destined for an application (let's call it `foo`) were being routed to another application (let's call it `bar`). 

Interestingly, the misrouting would only happen if we accessed `bar` before trying to reach `foo`. That is, if we quit the browser, opened it again, and accessed `foo` without first accessing `bar`, we would get routed to `foo` as expected. Once we accessed `bar`, all future requests to `foo` would get misrouted to `bar`.

In addition, we noticed that the misrouting only happened when using Chrome as the client. Requests produced by Firefox and cURL were being routed properly.

## The deployment configuration

- The environment has a wildcard DNS that resolves to the IP addresses of ingress nodes (e.g. `*.platform.us-east.example.com`)
- Application `foo` is exposed via `foo.platform.us-east.example.com` using a regular IngressRoute with TLS enabled. The certificate used for TLS is a wildcard cert for `*.platform.us-east.example.com`.
- Application `bar` is exposed using a `tcpproxy` IngressRoute with TLS passthrough enabled. The certificate used for TLS is a wildcard cert for `*.platform.us-east.example.com`.

## Our understanding of the issue

Based on https://github.com/envoyproxy/envoy/issues/6767#issuecomment-488811660, we believe the problem is caused due to the connection coalescing that happens in HTTP/2.

Once the browser opens the connection to `bar.platform.us-east.example.com`, it latches that connection to the Envoy listener/filter pair that is responsible for serving `bar`. Future requests to other applications on the same wildcard DNS entry (e.g. `foo.platform.us-east.example.com`) are sent over the same connection to the same Envoy listener/filter pair. 

Depending on which application is accessed first, we see different behavior:

### App exposed over TCP Proxy with TLS passthrough is accessed first

1. Open fresh browser process
2. Browse to `https://bar.platform.us-east.example.com`
3. Browser opens a connection to Envoy's `ingress_https` listener on the `envoy.tcp_proxy` filter.
4. Request routed successfully
5. Browse to `https://foo.platform.us-east.example.com`
[...]

PR #2483, opened 2020-04-29
internal: filter misdirected TLS requests

jpeach:
TLS routes are specialized to a unique virtual hostname. However, if
wildcard certificates are being used, browsers will aggressively coalesce
and reuse server connections even when the full origin hostname doesn't
match. This results on 404 responses because each TLS virtual host only
has routes for one host.

We can avoid this behaviour bleeding out to users by generating a 421
Misdirected Request response if the request authority doesn't match
the FQDN of the virtual host. In this case, the browser is supposed
to understand that the request wasn't processed and re-send it on a
new connection.

### DIFF
--- a/internal/contour/listener.go
+++ b/internal/contour/listener.go
@@ -1,4 +1,4 @@
-// Copyright © 2019 VMware
+// Copyright © 2020 VMware
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -298,6 +298,7 @@ func visitListeners(root dag.Vertex, lvc *ListenerVisitorConfig) map[string]*v2.
 	if lv.http {
 		// Add a listener if there are vhosts bound to http.
 		cm := envoy.HTTPConnectionManagerBuilder().
+			DefaultFilters().
 			RouteConfigName(ENVOY_HTTP_LISTENER).
 			MetricsPrefix(ENVOY_HTTP_LISTENER).
 			AccessLoggers(lvc.newInsecureAccessLog()).
@@ -366,6 +367,8 @@ func (v *listenerVisitor) visit(vertex dag.Vertex) {
 			// coded into monitoring dashboards.
 			filters = envoy.Filters(
 				envoy.HTTPConnectionManagerBuilder().
+					AddFilter(envoy.FilterMisdirectedRequests(vh.VirtualHost.Name)).
+					DefaultFilters().
 					RouteConfigName(path.Join("https", vh.VirtualHost.Name)).
 					MetricsPrefix(ENVOY_HTTPS_LISTENER).
 					AccessLoggers(v.ListenerVisitorConfig.newSecureAccessLog()).
@@ -403,10 +406,12 @@ func (v *listenerVisitor) visit(vertex dag.Vertex) {
 		v.listeners[ENVOY_HTTPS_LISTENER].FilterChains = append(v.listeners[ENVOY_HTTPS_LISTENER].FilterChains,
 			envoy.FilterChainTLS(vh.VirtualHost.Name, downstreamTLS, filters))
 
-		// If this VirtualHost has enabled the fallback certificate then set a default FilterChain which will allow
-		// routes with this vhost to accept non SNI TLS requests
+		// If this VirtualHost has enabled the fallback certificate then set a default
+		// FilterChain which will allow routes with this vhost to accept non-SNI TLS requests.
+		// Note that we don't add the misdirected requests filter on this chain because at this
+		// point we don't actually know the full set of server names that will be bound to the
+		// filter chain through the ENVOY_FALLBACK_ROUTECONFIG route configuration.
 		if vh.FallbackCertificate != nil && !envoy.ContainsFallbackFilterChain(v.listeners[ENVOY_HTTPS_LISTENER].FilterChains) {
-
 			// Construct the downstreamTLSContext passing the configured fallbackCertificate. The TLS minProtocolVersion will use
 			// the value defined in the Contour Configuration file if defined.
 			downstreamTLS = envoy.DownstreamTLSContext(
--- a/_integration/testsuite/httpproxy/004-https-sni-enforcement.yaml
+++ b/_integration/testsuite/httpproxy/004-https-sni-enforcement.yaml
@@ -117,7 +117,7 @@ fatal_proxy_is_not_present[msg] {
   msg := sprintf("HTTPProxy for %q is not present", [ fqdn ])
 }
 
---- 
+---
 
 Name := "echo-one"
 
@@ -243,7 +243,7 @@ fatal_proxy_is_not_present[msg] {
   msg := sprintf("HTTPProxy for %q is not present", [ fqdn ])
 }
 
---- 
+---
 
 Name := "echo-two"
 
@@ -309,9 +309,10 @@ error_path_routing_mismatch[msg] {
 ---
 
 import data.contour.http.client
+import data.contour.http.response
 
 # Ensure that sending a request to "echo-one" with the SNI from "echo-two"
-# generates a 404.
+# generates a 4xx response status.
 
 Response := client.Get({
   "url": sprintf("https://%s/https-sni-enforcement/%d", [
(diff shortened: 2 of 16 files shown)


jpeach:
When using the wrong host, envoy responds with a 421:
```
$ /usr/bin/curl -v -k -H "Host: foo.jpeach.org" --resolve echo-no-auth.projectcontour.io:443:127.0.0.1 https://echo-no-auth.projectcontour.io/get
* Added echo-no-auth.projectcontour.io:443:127.0.0.1 to DNS cache
* Hostname echo-no-auth.projectcontour.io was found in DNS cache
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to echo-no-auth.projectcontour.io (127.0.0.1) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/ssl/cert.pem
  CApath: none
* TLSv1.2 (OUT), TLS handshake, Client hello (1):
* TLSv1.2 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-CHACHA20-POLY1305
* ALPN, server accepted to use h2
* Server certificate:
*  subject: O=cert-manager
*  start date: Apr 29 06:37:40 2020 GMT
*  expire date: Jul 28 06:37:40 2020 GMT
*  issuer: O=cert-manager; OU=io + OU=projectcontour + OU=testsuite; CN=issuer
*  SSL certificate verify result: unable to get local issuer certificate (20), continuing anyway.
* Using HTTP2, server supports multi-use
* Connection state changed (HTTP/2 confirmed)
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* Using Stream ID: 1 (easy handle 0x7fd16f00d600)
> GET /get HTTP/2
> Host: foo.jpeach.org
> User-Agent: curl/7.64.1
> Accept: */*
>
* Connection state changed (MAX_CONCURRENT_STREAMS == 2147483647)!
< HTTP/2 421
< content-length: 0
< date: Wed, 29 Apr 2020 07:08:14 GMT
< server: envoy
<
* Connection #0 to host echo-no-auth.projectcontour.io left intact
* Closing connection 0
```

When using the correct host, routing works right:
```$ /usr/bin/curl -v -k --resolve echo-no-auth.projectcontour.io:443:127.0.0.1 https://echo-no-auth.projectcontour.io/get
* Added echo-no-auth.projectcontour.io:443:127.0.0.1 to DNS cache
* Hostname echo-no-auth.projectcontour.io was found in DNS cache
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to echo-no-auth.projectcontour.io (127.0.0.1) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/ssl/cert.pem
  CApath: none
* TLSv1.2 (OUT), TLS handshake, Client hello (1):
* TLSv1.2 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-CHACHA20-POLY1305
* ALPN, server accepted to use h2
* Server certificate:
*  subject: O=cert-manager
*  start date: Apr 29 06:37:40 2020 GMT
*  expire date: Jul 28 06:37:40 2020 GMT
*  issuer: O=cert-manager; OU=io + OU=projectcontour + OU=testsuite; CN=issuer
*  SSL certificate verify result: unable to get local issuer certificate (20), continuing anyway.
* Using HTTP2, server supports multi-use
* Connection state changed (HTTP/2 confirmed)
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* Using Stream ID: 1 (easy handle 0x7febc980b400)
> GET /get HTTP/2
> Host: echo-no-auth.projectcontour.io
> User-Agent: curl/7.64.1
> Accept: */*
>
* Connection state changed (MAX_CONCURRENT_STREAMS == 2147483647)!
< HTTP/2 200
< content-type: application/json
< date: Wed, 29 Apr 2020 07:08:20 GMT
< content-length: 435
< x-envoy-upstream-service-time: 1
< server: envoy
<
* Connection #0 to host echo-no-auth.projectcontour.io left intact
{"TestId":"ingress-conformance-echo","Path":"/get","Host":"echo-no-auth.projectcontour.io","Method":"GET","Proto":"HTTP/1.1","Headers":{"Accept":["*/*"],"Content-Length":["0"],"User-Agent":["curl/7.64.1"],"X-Envoy-Expected-Rq-Timeout-Ms":["15000"],"X-Envoy-Internal":["true"],"X-Forwarded-For":["172.17.0.1"],"X-Forwarded-Proto":["https"],"X-Request-Id":["0a813dec-dc78-4e4e-b1f2-44c15d41608f"],"X-Request-Start":["t=1588144100.581"]}}* Closing connection 0
```

jpeach:
@stevesloka @youngnick LMK what you think about this approach. As far I know, this is the only way to fix #1493 .

stevesloka:
The right place is to fix upstream in Envoy, but this is a good workaround for now. This has existing for some time, nice work @jpeach!

youngnick:
If it's the only way for now, I'm okay with it. I think we need to be clear that this doesn't mean we'll be adding support for arbitrary Lua filters (yet). I would like to see this removed in favour of a better upstream solution if it's implemented.

jpeach:
> If it's the only way for now, I'm okay with it. I think we need to be clear that this doesn't mean we'll be adding support for arbitrary Lua filters (yet). I would like to see this removed in favour of a better upstream solution if it's implemented.

I don't see much upstream activity on this. If we want something upstream we probably need to make a proposal and drive it.

youngnick:
I can't do C++, so I can't do the coding, but I agree.

stevesloka:
I just chatted with another Contour user who ran into this. They are going to make a build and test it out to verify (in their environment) that this solves, but I think we should merge this if the tests go well and then release a v1.14.1 to include this and the ingress.class status info fixes.

primeroz:
In my use case this seems to fix the issue , even though i am confused by the regex in https://github.com/projectcontour/contour/pull/2483/files#diff-98b92a08d0022a6c73eceeb2e1d99a43R3128 which i would expect to be a _negation_ ... but i obviously am not understanding this properly since it works :)

my setup

* wilcard `*.example.com` for HTTPProxy on https://podinfo-tls.example.com
* specific cert `vault.example.com` for a TLS passthrough tcp service

with 1.4.0 
* Open Incognito Chromium
* Hit _https://podinfo-tls.example.com_ works
* Hit , in the same incognito session, _https://vault.example.com_ fail with 404

Using a build of this branch
* Open Incognito Chromium
* Hit _https://podinfo-tls.example.com_ works
* Hit , in the same incognito session, _https://vault.example.com_ works

jpeach:
@youngnick @stevesloka Could you please review? The diff is pretty big unfortunately, due to testing impact. The main functional change is around the new FilterMisdirectedRequests API.

internal/contour/listener.go
```diff
@@ -403,10 +406,12 @@ func (v *listenerVisitor) visit(vertex dag.Vertex) {
 		v.listeners[ENVOY_HTTPS_LISTENER].FilterChains = append(v.listeners[ENVOY_HTTPS_LISTENER].FilterChains,
 			envoy.FilterChainTLS(vh.VirtualHost.Name, downstreamTLS, filters))
 
-		// If this VirtualHost has enabled the fallback certificate then set a default FilterChain which will allow
-		// routes with this vhost to accept non SNI TLS requests
+		// If this VirtualHost has enabled the fallback certificate then set a default
+		// FilterChain which will allow routes with this vhost to accept non-SNI TLS requests.
+		// Note that we don't add the misdirected requests filter on this chain because at this
```
youngnick, line 411: Presumably this means that if you're a client old enough not to support SNI, then you're a client that's not going to coalesce requests onto a single TLS session?

jpeach: The misdirected requests filter works by matching the hostname we get against the hostname we expect. In the fallback case, there are a set of hostnames that could be expected (all the vhosts that set the fallback enabled). At this point, however, we don't know  all those hostnames so we can't make the right filter chain.

So I think we could implement this; we would need to update the Lua code to match against a set of expected hostnames, and we would need to delay creating the fallback filter until we have seen all the vhosts (so we know the full set that enabled fallback).

ashirvadgupta:
@jpeach In my setup, I am changing the envoy https port from 443 to 8443. Its trying to match the host portal.example.com but its getting hosts as portal.example.com:8443. This is not handled in this change.

PR #2492, opened 2020-05-03
Add flag to restrict Contour to a single namespace

josephglanville:
The `--single-namespace` flag can be enabled to restrict Contour to searching and modifying resources in only the namespace it's deployed in.

This is especially useful in environment where it's desirable to run multiple instances of Contour, for instance one per namespace, each with their own Envoy deployments without conflicting with each other.

### DIFF
--- a/cmd/contour/ingressstatus.go
+++ b/cmd/contour/ingressstatus.go
@@ -38,11 +38,12 @@ import (
 //    is been received, operation restarts at step 3.
 // 5. If the worker is stopped, any existing informer is stopped before the worker stops.
 type loadBalancerStatusWriter struct {
-	log          logrus.FieldLogger
-	clients      *k8s.Clients
-	isLeader     chan struct{}
-	lbStatus     chan v1.LoadBalancerStatus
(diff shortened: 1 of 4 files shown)


youngnick:
Thanks for this PR @josephglanville. I see what the PR is intended to do, but generally, we prefer to discuss on an issue before making code changes. Could you please open an issue describing the use case you're trying to hit?

In this case, the thing I'd like to address is that Contour is mostly designed to hit the use case of a multitenant cluster where people want to control access to the namespace it's installed in (I'll write `projectcontour` namespace for brevity), and then allow app developers to deploy their Ingress or HTTPProxy objects alongside their app.

Using the model you have here, in order for to get that flow, you will need to grant RBAC on those objects inside the `projectcontour` namespace, and have that namespace be the only place where ingress objects are stored.

In any case, the best place to discuss this is on an issue. Thanks for your contribution, looking forward to reading more.

josephglanville:
Will re-visit later when plan for multiple namespaces is worked out.

Issue #2495, opened 2020-05-04
design: Add a pattern for adding extensible status information to HTTPProxy


youngnick:
This issue covers the work to add more information to HTTPProxy's `status` stanza. 

Features we need from HTTPProxy's status stanza:
- [x] Ability to see the way to reach the HTTPProxy (the equivalent to the Ingress status.loadBalancer stanza). #2371 covers the work to implement this.
- [x] Ability to set multiple statuses on a HTTPProxy. Although the guidance for their design is somewhat muddy at the moment (see kubernetes/community#4521), Conditions are the API-standard way to do this.
- [ ] A way for HTTPProxy resources to understand what routing conditions are delegated to them from parent HTTPProxies, and possibly a way for parent HTTPProxies to see what is visible underneath them.

Blocks #2325, through this blocks #370, #432, and #1691, and #399 
Blocks #2021 
Blocks #2371 
Blocks #1598
Blocks #2141  

Previous work was in #1868, although that was before I knew about Conditions.

~~We may also need to consider changing the `status` stanza to be an actual subresource, although that may be a large enough change that it should be a new document version~~ We did need to update to a subresource to make handling the status updates easier, and to allow separate RBAC of them.

bgagnon:
This would be a much welcome addition for us:
* we auto-provision DNS from HTTPProxy, but we're lacking a way to communicate the status of that back to the user
* we also auto-provision TLS certificates -- Contour's status field says _valid_ when the Secret exists and _invalid_ before that, but there's not a lot of nuance

Our solution for now is to emit detailed events (success/error) attached to the `HTTPProxy` for the above. If the status field can reflect the Envoy configuration status, plus some extensible fields, then we'd have a comprehensive status.

The only thing missing for a full CD pipeline would be for `kubectl` to grow the ability to _wait_ on these status values (https://github.com/kubernetes/kubernetes/issues/83094).

youngnick:
That's a good point @bgagnon - As I start experimenting with adding Conditions to HTTPProxy, I'll make sure I try out the support `kubectl wait` has for Conditions already.

In addition, after talking to @jpeach and @stevesloka today, I'm going to go ahead with #2371, and add the `loadBalancer` stanza to HTTPProxy. This will work in the same way that Ingress status updating works today - either pulled from the Envoy service, or passed directly to Contour.

stevesloka:
[...]

PR #2496, opened 2020-05-04
Add status subresource to HTTPProxy

youngnick:
This commit enables the status subresource for HTTPProxy.
This means that:
- normal HTTPProxy changes that include updates to the `status` stanza will have the status updates dropped
- updates to `status` can only be made using requests to the status subresource.

Also changes the HTTPProxy status updating to use the status subresource,
and adds RBAC for Contour to be able to update it.

Having the status subresource enabled is best-practice for CRDs and will be a foundation for further status work.

It also means that HTTPProxy objects no longer need to include the `status` stanza to be valid.

Updates #2495

### DIFF
--- a/internal/k8s/status.go
+++ b/internal/k8s/status.go
@@ -112,7 +112,8 @@ func (c *StatusCacher) SetStatus(status, desc string, obj interface{}) error {
 
 // StatusWriter updates the object's Status field.
 type StatusWriter struct {
-	Client dynamic.Interface
+	Client    dynamic.Interface
+	Converter Converter
 }
 
 // GetStatus is not implemented for StatusWriter.
@@ -141,7 +142,7 @@ func (irs *StatusWriter) SetStatus(status, desc string, existing interface{}) er
 				CurrentStatus: status,
 				Description:   desc,
 			}
-			return irs.setHTTPProxyStatus(exist, updated)
+			return irs.setHTTPProxyStatus(updated)
 		}
 	}
 	return nil
@@ -176,24 +177,14 @@ func (irs *StatusWriter) setIngressRouteStatus(existing, updated *ingressroutev1
 	return err
 }
 
-func (irs *StatusWriter) setHTTPProxyStatus(existing, updated *projcontour.HTTPProxy) error {
-	existingBytes, err := json.Marshal(existing)
+func (irs *StatusWriter) setHTTPProxyStatus(updated *projcontour.HTTPProxy) error {
+
+	usUpdated, err := irs.Converter.ToUnstructured(updated)
 	if err != nil {
-		return err
-	}
-	// Need to set the resource version of the updated endpoints to the resource
-	// version of the current service. Otherwise, the resulting patch does not
-	// have a resource version, and the server complains.
-	updated.ResourceVersion = existing.ResourceVersion
-	updatedBytes, err := json.Marshal(updated)
-	if err != nil {
-		return err
-	}
-	patchBytes, err := jsonpatch.CreateMergePatch(existingBytes, updatedBytes)
-	if err != nil {
-		return err
+		return fmt.Errorf("unable to convert status update to HTTPProxy: %s", err)
 	}
 
-	_, err = irs.Client.Resource(projcontour.HTTPProxyGVR).Namespace(existing.GetNamespace()).Patch(context.TODO(), existing.GetName(), types.MergePatchType, patchBytes, metav1.PatchOptions{})
+	_, err = irs.Client.Resource(projcontour.HTTPProxyGVR).Namespace(updated.GetNamespace()).
+		UpdateStatus(context.TODO(), usUpdated, metav1.UpdateOptions{})
 	return err
 }
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -423,6 +423,7 @@ type Status struct {
 // +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.currentStatus",description="The current status of the HTTPProxy"
 // +kubebuilder:printcolumn:name="Status Description",type="string",JSONPath=".status.description",description="Description of the current status"
 // +kubebuilder:resource:scope=Namespaced,path=httpproxies,shortName=proxy;proxies,singular=httpproxy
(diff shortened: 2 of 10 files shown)


youngnick:
Adding as a draft PR so that we can talk about if this is a good idea; I'm pretty sure that it is, and although it's kind of a GA guarantee violation, it should be worthwhile as we do more with status.

Importantly, it doesn't change any behaviour about *retrieving* the status, only about *setting* it, which only Contour should be doing.

I've updated the one place we update HTTPProxy objects' status right now, but anyone else consuming HTTPProxy objects and updating their status would also need to change to using a status update instead of patching the object.

stevesloka:
This also limits us to Kubernetes v1.16 which is where `subresources` went GA (https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#status-subresource).

youngnick:
Ah, thanks for that Steve. Subresources are available in the previous CRD versions though, I think. I will check that out.

youngnick:
The `status` subresource on CRDs has been beta since at least 1.14, so I'm confident we are okay to include it.

youngnick:
After talking briefly to @jpeach and @stevesloka about this today, I'm going to promote this to an actual PR - I really do think we need to do this.

internal/k8s/status.go
```diff
@@ -177,23 +180,27 @@ func (irs *StatusWriter) setIngressRouteStatus(existing, updated *ingressroutev1
...
 	if err != nil {
 		return err
 	}
-	patchBytes, err := jsonpatch.CreateMergePatch(existingBytes, updatedBytes)
+
+	_, err = irs.Client.Resource(projcontour.HTTPProxyGVR).Namespace(existing.GetNamespace()).UpdateStatus(context.TODO(), usUpdated, metav1.UpdateOptions{})
+	return err
+}
+
+func toUnstructured(obj runtime.Object, gvk schema.GroupVersionKind) (*unstructured.Unstructured, error) {
```
jpeach, line 193: Shall we move this to the `internal/k8s` package as `k8s.toUnstructured(runtime.Object)*u.U`? `k8s.ToFullName` already established the API pattern.

internal/k8s/status.go
```diff
@@ -177,23 +180,27 @@ func (irs *StatusWriter) setIngressRouteStatus(existing, updated *ingressroutev1
...
+	objMsi, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	_, err = irs.Client.Resource(projcontour.HTTPProxyGVR).Namespace(existing.GetNamespace()).Patch(context.TODO(), existing.GetName(), types.MergePatchType, patchBytes, metav1.PatchOptions{})
-	return err
+	us := &unstructured.Unstructured{}
+	us.SetUnstructuredContent(objMsi)
```
jpeach, line 202: I think that the default scheme can do this conversion directly without needing to pass in the GVK, because we register all our types with it:
```
        u := &unstructured.Unstructured{}

        if err := scheme.Scheme.Convert(obj, u, nil); err != nil {
                log.Fatalf("conversion to unstructured failed: %s", err)
        }

        return u
```

Might need to check into the overlap with `k8s.UnstructuredConverter`.

internal/k8s/status.go
```diff
@@ -177,23 +180,27 @@ func (irs *StatusWriter) setIngressRouteStatus(existing, updated *ingressroutev1
...
-	existingBytes, err := json.Marshal(existing)
-	if err != nil {
-		return err
-	}
-	// Need to set the resource version of the updated endpoints to the resource
-	// version of the current service. Otherwise, the resulting patch does not
-	// have a resource version, and the server complains.
-	updated.ResourceVersion = existing.ResourceVersion
-	updatedBytes, err := json.Marshal(updated)
+
```
jpeach, line 183: We used to need `existing` and `updated` to generate the JSON patch, but now we should only need to update the status resource with the `updated` proxy object.

internal/k8s/status.go
```diff
@@ -177,23 +180,27 @@ func (irs *StatusWriter) setIngressRouteStatus(existing, updated *ingressroutev1
...
+	usUpdated, err := toUnstructured(updated, projcontour.SchemeGroupVersion.WithKind("HTTPProxy"))
 	if err != nil {
 		return err
 	}
-	patchBytes, err := jsonpatch.CreateMergePatch(existingBytes, updatedBytes)
+
+	_, err = irs.Client.Resource(projcontour.HTTPProxyGVR).Namespace(existing.GetNamespace()).UpdateStatus(context.TODO(), usUpdated, metav1.UpdateOptions{})
```
jpeach, line 189: Nit: this is a pretty long line, can we fold it a bit? :)

internal/k8s/status.go
youngnick: Yes, I'll need it in the next PR anyway, thanks.

internal/k8s/status.go
youngnick: The UnstructuredConverter currently only converts *from* unstructured. Makes sense for it to do both ways though. I'll refactor accordingly, that way I'll move the logic where it should be and make it more available.

internal/k8s/status.go
```diff
@@ -176,24 +177,14 @@ func (irs *StatusWriter) setIngressRouteStatus(existing, updated *ingressroutev1
...
-	if err != nil {
-		return err
-	}
-	patchBytes, err := jsonpatch.CreateMergePatch(existingBytes, updatedBytes)
+func (irs *StatusWriter) setHTTPProxyStatus(updated *projcontour.HTTPProxy) error {
+
+	usUpdated, err := irs.Converter.ToUnstructured(updated)
 	if err != nil {
-		return err
+		return fmt.Errorf("Unable to convert status update to HTTPPRoxy: %s", err)
```
jpeach, line 184: Nit, drop the initial caps

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

jpeach: s/HTTPPRoxy/HTTPProxy/

youngnick: One of these days I'll get that right, sigh. On both counts.

Issue #2486, opened 2020-04-30
Ability to vendor contour configuration files


shashwathi:
## Problem
[Knative](https://github.com/knative/) project is starting to switch individual projects to use Go modules instead of Godep. [`net-contour`](https://github.com/knative/net-contour) project provides configuration yaml files in [examples folder](https://github.com/projectcontour/contour/tree/master/examples/contour) for generating contour configurations.
In Godep world the whole repo is vendored in and [knative scripts](https://github.com/knative/net-contour/blob/master/hack/update-deps.sh#L112) would refer to the vendored example files. Folders with only non Go files are not vendored by Gomod. 

Possible solutions : One option is to add a dummy go file so we can vendor in `examples/contour` folder.

We are open to other ideas or thoughts? Thanks 

cc @andrew-su

stevesloka:
//cc #2254

stevesloka:
I think adding a dummy go file in that directory is fine, the only caution we need to make is how to maintain going forward. There's some work to break apart the current `examples` folder into pieces that would break this (or require more dummy files). 

We just need to decide on what we're going to maintain long-term and how.

jpeach:
xref #2474

youngnick:
Yes, I think that this should wait until we decide what we're doing with #2474 - knative will have a bunch of work to update their installs if we do that as well, so let's talk there.

mattmoor:
Ack that this might not be the forever plan (or even the 1.5+ plan), but I wonder if you would be receptive to landing `dummy.go` on `release-1.4` to unblock our `go mod` migration?  🙏

The idea of trying to land `go mod` while simultaneously pulling in kustomize scares me a little, so I'm mostly looking for ways we can take it in pieces with the least disruption to what you all have planned.

stevesloka:
Closed via #2503

mattmoor:
🤔 I didn't close this with my PR because we still need a story going forward.  If that's tracked elsewhere (#2474?) then 👍

youngnick:
I think part of the story with kustomize is for us to understand how knative is consuming Contour and make sure that whatever we do has a way forward for you too.

PR #2503, opened 2020-05-06
[RELEASE 1.4] Optimistically add dummy.go to release-1.4

mattmoor:
While kustomize is explored at HEAD, this adds dummy.go to the 1.4 release branch to unblock consuming the example configurations through `go mod` vendoring.

Related: https://github.com/projectcontour/contour/issues/2486

### DIFF
--- a/examples/contour/dummy.go
+++ b/examples/contour/dummy.go
@@ -1,0 +1,16 @@
+// Copyright © 2020 VMware
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package contour contains a dummy Go file that enables the Contour
+// configuration files to be pulled in via go mod vendoring.
+package contour
--- a/examples/contour/dummy.go
+++ b/examples/contour/dummy.go
@@ -1,3 +1,5 @@
+// +build vendormagic
+
 // Copyright © 2020 VMware
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.


stevesloka:
@mattmoor one thing we might do for now is also put this on `master`.  If/when we redo the deployment files then we can remove it at that time. I'm thinking whatever we start using, we'll still have this style output available for users.

mattmoor:
@stevesloka Sure, happy to do that now or later, just let me know what you prefer.  Early indications are that this unblocks us on the Contour front :)

jpeach:
Why doesn't knative just copy the yaml into their repo (i.e. vendor it manually)?

mattmoor:
We actually do, but we do it thru vendoring what you all check in and then mutating it into our tree so that we have a single consistent dependency declaration on Contour (thru `go mod`).

mattmoor:
this used to be a lot easier with `dep` because it exposed pruning options :(

jpeach:
> Why doesn't knative just copy the yaml into their repo (i.e. vendor it manually)?

OTOH, _if_ we move the YAML to kustomize (or anything else), the same problems with Go vending would still occur. We would have to add a dummy file to every yaml directory.

mattmoor:
@jpeach It may only be needed alongside the base templates, then folks can use kustomize to render what they want using their own customized overrides.  I'm waving my hands a bit as I haven't looked through the proposal, but that's what I suspect we'd do... 🤔

jpeach:
> this used to be a lot easier with `dep` because it exposed pruning options :(

Maybe https://github.com/goware/modvendor helps?

mattmoor:
FWIW I'd really rather avoid introducing a new tool (I get that this sucks, but I'll hold my colorful `go mod` commentary 😉).  This pattern is becoming more standard as the "solution" for pulling in tool and non-Go dependencies.

Here's an example from the K8s code-generators: https://github.com/kubernetes/code-generator/blob/7baaea6316a654efa04671651b591fa6d87069b5/tools.go#L19-L21

This example actually does double-duty, it turns the root of the repo into a package so that `generate-groups.sh` gets vendored and acts as a single thing to link against (via a similar file in the downstream repo) so that each of the code-generators need not be enumerated.

jpeach:
> FWIW I'd really rather avoid introducing a new tool (I get that this sucks, but I'll hold my colorful `go mod` commentary 😉). 

Yeh I get that. Nick and Steve are cool with this, so I'm cool with it too.

mattmoor:
👍 I added the build tags, so LMK if you want me to change anything else.  Also happy to continue discussing options for 1.5+, just LMK when/where!

PR #2507, opened 2020-05-08
site: remove `kind/documentation` label

jpeach:
There's no obvious distinction between `kind/documentation` and
`area/documentation`. Since `area` scopes the part of the project
an issue applies to, that seems lke the better choice, so move all
the `kind/documentation` issues to `area/documentation`.

### DIFF
--- a/site/_data/github-labels.yaml
+++ b/site/_data/github-labels.yaml
@@ -57,6 +57,7 @@ default:
       addedBy: label
       previously:
         - name: documentation
+        - name: kind/documentation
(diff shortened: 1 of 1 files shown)


stevesloka:
Could we ever have a `kind/documentation`  with a different `area`  than documentation?

jpeach:
> Could we ever have a `kind/documentation` with a different `area` than documentation?

Yeh a documentation issue about HTTPProxy could be for `area/httpproxy`. An issue could have multiple area labels.

PR #2508, opened 2020-05-08
doc: improve ambiguous upstream TLS example

### LINKED ISSUE #2457: HTTPProxy Upstream TLS doc - sample YAML doesn't match the scenario explained.

Documentation at:
https://github.com/projectcontour/contour/blob/master/site/docs/master/httpproxy.md#upstream-tls

Describes:
"""
In this example, the upstream service is named https and uses port 443
"""

The sample YAML does not reflect this statement, and can be misleading to the reader.

I'd like to see both the samples for HTTPProxy, and also the sample for the upstream service with the appropriate annotations, for an e2e example.

**Environment:**

- Contour version:

jpeach:
The text that attempts to describe using annotations to enable upstream
TLS doesn't match the example properly, and the example isn't very clear.

Update both the HTTPProxy and the annotations documentation for clarity,
and add a better example that demonstrates how to apply the protocol
annotation.

This fixes #2457.

### DIFF
--- a/site/docs/master/httpproxy.md
+++ b/site/docs/master/httpproxy.md
@@ -254,31 +254,50 @@ The TLS **Minimum Protocol Version** a vhost should negotiate can be specified b
 
 #### Upstream TLS
 
-A HTTPProxy can proxy to an upstream TLS connection by first annotating the upstream Kubernetes service with: `projectcontour.io/upstream-protocol.tls: "443,https"`.
-This annotation tells Contour which port should be used for the TLS connection.
-In this example, the upstream service is named `https` and uses port `443`.
-Additionally, it is possible for Envoy to verify the backend service's certificate.
-The service of an HTTPProxy can optionally specify a `validation` struct which has a mandatory `caSecret` key as well as an mandatory `subjectName`.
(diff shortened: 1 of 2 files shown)


jpeach:
This is a local improvement, but we really need to rethink the structure of the large HTTPProxy document :(

site/docs/master/httpproxy.md
```diff
@@ -254,31 +254,48 @@ The TLS **Minimum Protocol Version** a vhost should negotiate can be specified b
 
 #### Upstream TLS
 
-A HTTPProxy can proxy to an upstream TLS connection by first annotating the upstream Kubernetes service with: `projectcontour.io/upstream-protocol.tls: "443,https"`.
+A HTTPProxy can proxy to an upstream TLS connection by first annotating the upstream Kubernetes Service with the `projectcontour.io/upstream-protocol.tls` annotation.
```
stevesloka, line 257: We should call out here (and probably the annotations sections maybe), that you can alternatively use the `Spec.Routes.Services.Protocol` to specify the same information and not use the annotation.

jpeach: Yep, that's a good idea. Fixed.

jpeach:
@stevesloka Could you please do another review pass?

PR #2512, opened 2020-05-11
site: tune codecov settings

jpeach:
Add a default codecov settings file, and make the `codecov/project`
status informational, since it otherwise makes CI look broken.

### DIFF
--- a/codecov.yaml
+++ b/codecov.yaml
@@ -1,0 +1,26 @@
+codecov:
+  require_ci_to_pass: yes
+
+coverage:
+  precision: 2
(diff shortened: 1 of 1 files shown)


jpeach:
@stevesloka @youngnick Pretty sure that this is the same codecov config that we get with the defaults. The change is that the `codecov/project` entry in the CI checks won't ever be marked as failed (the `codecov/patch` entry could still be), so PRs won't seem to get penalized when they reduce the overall coverage (we don't generally worry about that in reviews).

PTAL

Issue #2487, opened 2020-04-30
Ability to generate contour clients


shashwathi:
## Feature

Ability to generate clients for contour API. 

In `net-contour` knative project we are maintaining a [github patch](https://github.com/knative/net-contour/blob/master/hack/contour.patch) to overcome this problem as a workaround. 

cc @andrew-su

stevesloka:
We might be able to create a `contour-api` project and have it auto-gen this client in automatically. This would give folks a full client to use in projects.

bgagnon:
We're using the Contour API via [kubernetes-sigs/controller-runtime](https://github.com/kubernetes-sigs/controller-runtime) and have been mostly fine:

```go
	ingressroutev1b1.AddKnownTypes(mgr.GetScheme())
	projectcontour.AddKnownTypes(mgr.GetScheme())
```

...

```go
	ctrl.NewControllerManagedBy(mgr).
		For(&projectcontour.HTTPProxy{}).
		Owns(&v1.Secret{}).
		Complete(...)
```

But the idea of an official Contour API package is appealing. In the last upgrade we had to switch to `AddKnownTypes()` which was an unexpected "breakage".

stevesloka:
Yeah, we did a poor job of polling the community on that change, so I'm sorry for that. 😞

jpeach:
@stevesloka Any reason we can't take the knative patch? Or perhaps we should double-check what kubebuilder would generate and ensure we have the expected API there ...

youngnick:
We removed the generated Informer types earlier because we weren't using them, and we didn't realise anyone else was. I don't see any issue with putting the generation back if people are using them, as long as we call out that the types should be consumed by version tag (which go modules does anyway).

youngnick:
Splitting the objects out to a new repo is doable but requires a bit of coordination.

jpeach:
> We removed the generated Informer types earlier because we weren't using them, and we didn't realise anyone else was. I don't see any issue with putting the generation back if people are using them, as long as we call out that the types should be consumed by version tag (which go modules does anyway).

It's not clear to me that we need to generate the client (anyone who needs it can generate it from the API package). Rather, it should be easy to use the api package with unstructured types. The Knative patch looks to be targeted at the latter case.

youngnick:
[...]

PR #2514, opened 2020-05-11
api: add kubebuilder-compatible type registration exports

jpeach:
Restore the kubebuilder-compatible API type registration exports for
IngressRoute and HTTPProxy. Keep exporting AddKnownTypes() for projects
that already started consuming this symmbol, but note that it is not
should no longer be used.

This updates #2487.

### DIFF
--- a/apis/contour/v1beta1/register.go
+++ b/apis/contour/v1beta1/register.go
@@ -1,4 +1,4 @@
-// Copyright © 2019 VMware
+// Copyright © 2020
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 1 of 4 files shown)


apis/projectcontour/v1/register.go
```diff
@@ -33,12 +33,30 @@ func Resource(resource string) schema.GroupResource {
...
 	metav1.AddToGroupVersion(scheme, SchemeGroupVersion)
+	return nil
 }
+
+// The following declarations are kubebuilder-compatible and will be expected
+// by third parties who import the Contour API types.
+
+var (
+	// GroupVersion is group version used to register these objects
+	GroupVersion = schema.GroupVersion{Group: GroupName, Version: "v1"}
```
stevesloka, line 55: We already have this field above with a different name, maybe merge them? (https://github.com/projectcontour/contour/blob/751a0ebc939d14ce16bd31469d86082896c3cb09/apis/projectcontour/v1/register.go#L27)

apis/contour/v1beta1/register.go
```diff
@@ -34,12 +34,30 @@ func Resource(resource string) schema.GroupResource {
...
 	metav1.AddToGroupVersion(scheme, SchemeGroupVersion)
+	return nil
 }
+
+// The following declarations are kubebuilder-compatible and will be expected
+// by third parties who import the Contour API types.
+
+var (
+	// GroupVersion is group version used to register these objects
+	GroupVersion = schema.GroupVersion{Group: GroupName, Version: "v1beta1"}
```
stevesloka, line 56: Same as previous comment, duplicated field.

apis/projectcontour/v1/register.go
jpeach: Yeh, I just got paranoid about breaking someone that imports it? I added a comment that is is for compatibility, but we can remove it if we think that's a low risk change.

youngnick: We need to keep both names - that's the whole point of this exercise, sadly. I think James' solution for now is the best we're going to get.

internal/k8s/converter.go
```diff
@@ -93,8 +93,8 @@ func NewUnstructuredConverter() (*UnstructuredConverter, error) {
 	}
 
 	// Setup converter to understand custom CRD types
-	projectcontour.AddKnownTypes(uc.scheme)
-	ingressroutev1.AddKnownTypes(uc.scheme)
+	projectcontour.AddToScheme(uc.scheme) // nolint:errcheck
```
youngnick, line 96: Should probably just add the error checking here, we do it below for everything else that uses `AddToScheme`

internal/k8s/status_test.go
```diff
@@ -47,7 +47,7 @@ func TestSetIngressRouteStatus(t *testing.T) {
 			t.Helper()
 			var gotPatchBytes []byte
 			s := runtime.NewScheme()
-			ingressroutev1.AddKnownTypes(s)
+			ingressroutev1.AddToScheme(s) // nolint:errcheck
```
youngnick, line 50: I think we should just add the error checking here as well.

internal/k8s/converter.go
jpeach: 'kk, will do 👍

jpeach: Fixed.

internal/k8s/status_test.go
jpeach: Fixed.

apis/projectcontour/v1/register.go
jpeach: 'kk, leaving this for now.

apis/contour/v1beta1/register.go
jpeach: Leaving this for compatibility for now.

PR #2519, opened 2020-05-12
internal/featuretests: add simple HTTPProxy fixtures

jpeach:
Add some simple helpers to generate HTTPProxy test fixtures. This
helps reduce some boilerplate in setting up test cases.


jpeach:
@stevesloka @youngnick Let me know what you think of this approach. We can remove a bit more boilerplate by adding more fluent helpers and extending to other types.

jpeach:
> I think this is fine. We might also need a `.WithNamespace()` or something to allow different namespaces, but obviously can come when needed.

`.WithName` accepts `namespace/name` syntax to set the namespace, but you are right, we can add a new namespace setter too if it's convenient.

jpeach:
FWIW the Travis checks passed but the GitHub status wasn't updated.

PR #2520, opened 2020-05-13
internal: consolidate duplicate toAny functions

### LINKED ISSUE #2493: Consolidate toAny helper functions
We have 4 extremely similar `toAny` functions to generate `any.Any` protobufs:

```
internal/e2e/e2e.go:func toAny(t *testing.T, pb proto.Message) *any.Any {
internal/envoy/listener.go:func toAny(pb proto.Message) *any.Any {
internal/featuretests/featuretests.go:func toAny(t *testing.T, pb proto.Message) *any.Any {
internal/grpc/xds.go:func toAny(typeURL string, values []proto.Message) ([]*any.Any, error) {
```

We should consolidate these into the `internal/protobuf` package to improve discovery, documentation and testing.

jpeach:
Create a `protobuf.MustMarshalAny` wrapper and use it to replace
local copies of the `toAny` helper function.

This fixes #2493.

### DIFF
--- a/internal/protobuf/helpers.go
+++ b/internal/protobuf/helpers.go
@@ -19,6 +19,8 @@ import (
 	"time"
 
 	"github.com/golang/protobuf/proto"
+	"github.com/golang/protobuf/ptypes"
+	"github.com/golang/protobuf/ptypes/any"
 	"github.com/golang/protobuf/ptypes/duration"
 	"github.com/golang/protobuf/ptypes/wrappers"
 )
@@ -61,4 +63,15 @@ func AsMessages(messages interface{}) []proto.Message {
 	}
 
 	return protos
+}
+
+// MustMarshalAny marshals a protobug into an any.Any type, panicing
+// if that operation fails.
+func MustMarshalAny(pb proto.Message) *any.Any {
+	a, err := ptypes.MarshalAny(pb)
+	if err != nil {
+		panic(err.Error())
+	}
+
+	return a
 }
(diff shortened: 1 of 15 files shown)


internal/protobuf/helpers.go
```diff
@@ -62,3 +64,14 @@ func AsMessages(messages interface{}) []proto.Message {
 
 	return protos
 }
+
+// MustMarshalAny marshals a protobug into an any.Any type, panicing
+// if that operation fails.
+func MustMarshalAny(pb proto.Message) *any.Any {
+	a, err := ptypes.MarshalAny(pb)
+	if err != nil {
+		panic(err.Error())
```
stevesloka, line 73: Contour will crash if this fails? Should we return the error and let the caller decide what action to take?

youngnick: I agree that in general we shouldn't panic, but here, if the function returns an error as well, it can't be used as an inline helper, and will require caller-location error handling. So you can't just use it inline in structs any more to help with brevity.

It's actually the same behaviour as the thing it's replacing. I think it's probably okay in this case because we control the generation of the objects that are being marshalled; if anything does go wrong, things are in a bad enough state that Contour should stop working.

jpeach: Yup, there is no change in behaviour, but the `Must*` naming convention makes it clearer at the callsite that it might panic.

Previously on slack, I've argued that the way Contour uses this helper is safe because we never directly marshall anything the user gives us. If we internally generate a protobuf marshaling error, we are very broken :)

Issue #2371, opened 2020-03-23
Set load balancer address on HTTPProxy objects


jpeach:
**Please describe the problem you have**

The Ingress status exposes the load balancer IP address so that ingresses can automatically be registered in DNS, or directly accessed. We should add the same capabilities to HTTPProxy.

Updated todo from @youngnick:

- [x] Add a StatusUpdateHandler that takes status updates for Ingress objects from `k8s.StatusAddressUpdater` (renamed from IngressStatusUpdater), and applies them to the apiserver if they're an actual change. This is done with an interface (`k8s.StatusUpdater`) with a caching implementation for testing, and a writing implementation for actual use. This is PR #2530. 
- [x] Add support for HTTPProxy to `k8s.StatusAddressUpdater`. This will produce some raciness similar to #2522, which should be mitigated by the next step. This is PR #2551 
- [x] ~~Move the existing DAG rebuild status updates to use StatusUpdateHandler, including adding support for IngressRoute and HTTPProxy `valid` and `desc` fields. This may involve a small update to IngressRoute to enable the `status` subresource, so that the same code paths can be used for everything.~~ Moved this work to #2560.

jpeach:
xref #403

jpeach:
Moving to the current work stream to match the status of #403

stevesloka:
We need to come up with at least a high-level design doc on this first before implementing anything different in Status I would recommend.

youngnick:
I think that the loadBalancer stanza is standard enough we could just add it to HTTPProxy wholesale, and leave the other conditions work until later.

This would give HTTPProxy the same functionality as Ingress for a minimal investment.

youngnick:
After discussing today, I'm going to go ahead with just bringing the `loadBalancer` stanza from Ingress and Service into `status` on HTTPProxy, and updating it in the same way as we do for Ingress today.

youngnick:
So, turns out that doing the HTTPProxy status the same way as we do for Ingress doesn't work, as there are then two places updating the status, and neither of them will retry if the other has updated a status while they were working.

We *could* address this with `RetryOnConflict` from client-go, but I think there's an opportunity here to refactor to make the further changes coming from #2495 easier to implement.

[...]

PR #2530, opened 2020-05-18
Add StatusUpdateHandler to centralize status updates

youngnick:
Updates #2371

This is step 1 in a three phase plan, see https://github.com/projectcontour/contour/issues/2371#issuecomment-629890435 for the full version.

Also add StatusUpdater as an interface, to allow testing of the IngressStatusUpdater Informer.

Rename `k8s.IngressStatusUpdater` to `k8s.StatusAddressUpdater`, in preparation for it being used for
HTTPProxy objects as well.


internal/k8s/statusaddress.go
```diff
@@ -14,82 +14,82 @@
...
-// IngressStatusUpdater observes informer OnAdd events and
+// StatusAddressUpdater observes informer OnAdd and OnUpdate events and
 // updates the ingress.status.loadBalancer field on all Ingress
 // objects that match the ingress class (if used).
-type IngressStatusUpdater struct {
-	Client       clientset.Interface
-	Logger       logrus.FieldLogger
-	Status       v1.LoadBalancerStatus
-	IngressClass string
+type StatusAddressUpdater struct {
```
stevesloka, line 26: Looking forward this field is going to be used for more than just status. We should take `Address` out of the name to allow it to update more than just the address.

internal/k8s/statusaddress.go
```diff
@@ -14,82 +14,82 @@
...
+func (s *StatusAddressUpdater) OnAdd(obj interface{}) {
 
 	ing := obj.(*v1beta1.Ingress).DeepCopy()
+	ing.SetGroupVersionKind(v1beta1.SchemeGroupVersion.WithKind("ingress"))
 	if !annotation.MatchesIngressClass(ing, s.IngressClass) {
 		s.Logger.
 			WithField("name", ing.GetName()).
 			WithField("namespace", ing.GetNamespace()).
 			WithField("ingress-class", annotation.IngressClass(ing)).
+			WithField("defined-ingress-class", s.IngressClass).
```
stevesloka, line 42: nit: Maybe call this `annotation-ingress-class` or something? It's a good idea to show the difference between the configured and what the object has, but still not clear to me what "defined" is vs "ingress-class".

internal/k8s/statusaddress.go
```diff
@@ -14,82 +14,82 @@
...
+	IngressClass  string
+	StatusUpdater StatusUpdater
 }
 
-func (s *IngressStatusUpdater) OnAdd(obj interface{}) {
+func (s *StatusAddressUpdater) OnAdd(obj interface{}) {
 
 	ing := obj.(*v1beta1.Ingress).DeepCopy()
+	ing.SetGroupVersionKind(v1beta1.SchemeGroupVersion.WithKind("ingress"))
 	if !annotation.MatchesIngressClass(ing, s.IngressClass) {
```
stevesloka, line 37: Should do this check first to fail fast in the event this is an object we don't care about.

internal/k8s/statusaddress_test.go
```diff
@@ -0,0 +1,355 @@
...
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/projectcontour/contour/internal/assert"
+	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/api/networking/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestServiceStatusLoadBalancerWatcherOnAdd(t *testing.T) {
```
stevesloka, line 27: Should make this table style tests like in `TestStatusAddressUpdater_OnUpdate`.

internal/k8s/statusaddress_test.go
```diff
@@ -0,0 +1,355 @@
...
+
+}
+
+func simpleIngressGenerator(name, ingressClass string, lbstatus v1.LoadBalancerStatus) *v1beta1.Ingress {
+	return &v1beta1.Ingress{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: name,
+			Annotations: map[string]string{
+				"kubernetes.io/ingress.class": ingressClass,
```
stevesloka, line 347: Should we also test using `projectcontour.io/ingress.class`? Might be better as a feature test for that one.

internal/k8s/statusaddress.go
youngnick: This one probably won't be used for anything else, it's only designed to be used to update `status.loadBalancer` stanzas in things. Other status updates will come from the Builder, replacing the `StatusClient` interface there.

internal/k8s/statusaddress.go
youngnick: how about `configured-ingress-class`?

internal/k8s/statusaddress_test.go
youngnick: The `ServiceStatusLoadBalancerWatcher` tests are actually old, I've moved them into a better-named file, so they are showing as new. I've kept the same format for the `ServiceAddressUpdater` tests, so all the tests in the file were consistent. This is also the style of tests that @jpeach moved the `StatusClient` tests to, I think.

internal/k8s/statusaddress_test.go
youngnick: We can test using either, that's actually testing the code in `annotation.MatchesIngressClass`, which already has code for `projectcontour.io/ingress.class`. A feature test is a good idea though.,

internal/k8s/statusaddress.go
youngnick: We can't do the check first, because `MatchesIngressClass` requires typecasting from `interface{}` to a type that fulfills the `ObjectMeta` interface, in this case `*v1beta1.Ingress`. We'll have to keep this pattern even after I add HTTPProxy in here (which is most likely the only other type that will be added in here).

internal/k8s/statusaddress_test.go
youngnick: Looks like I was wrong on that one, but @stevesloka is fixing this in #2536, thanks Steve!

cmd/contour/serve.go
```diff
@@ -316,13 +316,29 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+		Log:           log.WithField("context", "StatusUpdateWriter"),
+		Clients:       clients,
+		UpdateChannel: updateChannel,
+		LeaderElected: eventHandler.IsLeader,
+		Converter:     converter,
+	}
+
+	suw := k8s.StatusUpdateWriter{
+		UpdateChannel: updateChannel,
+	}
```
jpeach, line 330: We can remove the need to manage a channel in the caller by adding a helper to the handler:
```Go
func (*k8sStatusUpdateHandler) Writer() k8s.StatusUpdater
```

The update handler can create the channel implicitly and hand one end to the writer, but the caller doesn't need to know there are channels involved and you don't need to expose StatusUpdateWriter as a type.

internal/k8s/statusupdater.go
```diff
@@ -0,0 +1,186 @@
...
+	FullName FullName
+	GVR      schema.GroupVersionResource
+	Mutator  StatusUpdateMutator
+}
+
+// StatusUpdateMutator is an interface to hold mutator functions for status updates.
+type StatusUpdateMutator interface {
+	Mutate(obj interface{}) interface{}
+}
+
```
jpeach, line 38: Suggest add a function adaptor
```Go
type StatusMutatorFunc func(interface{})interface{}

func (m StatusMutatorFunc) Mutate(old interface{}) interface{} {
  if m == nil {
    return nil
  }

  return m(old)
}
```

internal/k8s/statusaddress.go
```diff
@@ -14,86 +14,96 @@
...
-				WithError(err).Error("unable to update status")
+		m := mutator{
+			lbstatus: s.LBStatus,
 		}
+
+		s.StatusUpdater.SetStatus(
+			newIng.GetName(),
+			newIng.GetNamespace(),
+			v1beta1.SchemeGroupVersion.WithResource("ingresses"),
+			m,
```
jpeach, line 79: Rather than a local struct type, you can adapt a function in place:

```Go
k8s.StatusMutatorFunc(func (obj interface{}) interface {
  ing := obj.(*v1beta1.Ingress).DeepCopy()
  ing.Status.LoadBalancer = s.LBStatus
  return ing
})
```

internal/k8s/statusupdater.go
```diff
@@ -0,0 +1,186 @@
...
+	"github.com/projectcontour/contour/internal/assert"
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// StatusUpdate contains an all the information needed to change an object's status to perform a specific update.
+// Send down a channel to the goroutine that actually writes the changes back.
+type StatusUpdate struct {
+	FullName FullName
```
jpeach, line 29: Suggest just `Name` for this field?

internal/k8s/statusupdater.go
```diff
@@ -0,0 +1,186 @@
...
+		case <-stop:
+			return nil
+		case <-sw.LeaderElected:
+			sw.Log.Info("elected leader")
+			sw.IsLeader = true
+			// disable this case
+			sw.LeaderElected = nil
+		case upd := <-sw.UpdateChannel:
+			if !sw.IsLeader {
+				sw.Log.WithField("fullname", upd.FullName).Debug("not leader, not applying update")
```
jpeach, line 64: In other places we use separate `name` and `namespace` fields. Not sure what is more useful, but we should be consistent here and below.

internal/k8s/statusupdater.go
```diff
@@ -0,0 +1,186 @@
...
+				continue
+			}
+
+			sw.Log.WithField("fullname", upd.FullName).Debug("Received a status update to write.")
+			uObj, err := sw.Clients.DynamicClient().Resource(upd.GVR).Namespace(upd.FullName.Namespace).Get(context.TODO(), upd.FullName.Name, metav1.GetOptions{})
+			if err != nil {
+				sw.Log.WithError(err).WithField("fullname", upd.FullName).WithField("groupversionresource", upd.GVR).Error("unable to retrieve object for updating")
```
jpeach, line 71: Suggest "resource" rather than "groupversionresource"

internal/k8s/statusupdater.go
```diff
@@ -0,0 +1,186 @@
...
+
+		}
+
+	}
+
+}
+
+// StatusUpdater describes an interface to send status updates somewhere.
+type StatusUpdater interface {
+	SetStatus(name, namespace string, gvr schema.GroupVersionResource, mutator StatusUpdateMutator)
```
jpeach, line 108: Suggest that the natural operation for a status updater is `Update`.

jpeach: Did you consider using a `FullName` for the first argument?

internal/assert/status.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"github.com/google/go-cmp/cmp"
+	"k8s.io/api/networking/v1beta1"
+)
+
+// StatusEqual checks that two objects of supported Kubernetes types
+// have equivalent Status structs.
+// Currently supports:
+// networking.k8s.io/ingress/v1beta1
+func StatusEqual(objA, objB interface{}) bool {
```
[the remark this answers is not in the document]
youngnick: Done

internal/k8s/statusupdater.go
```diff
@@ -0,0 +1,186 @@
...
+
+			obj, err := sw.Converter.FromUnstructured(uObj)
+			if err != nil {
+				sw.Log.WithError(err).WithField("fullname", upd.FullName).Error("unable to convert from unstructured")
+				continue
+			}
+
+			newObj := upd.Mutator.Mutate(obj)
+
+			if assert.StatusEqual(obj, newObj) {
```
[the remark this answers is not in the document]
youngnick: Agreed.

internal/k8s/statusupdater.go
```diff
@@ -0,0 +1,186 @@
...
+// StatusUpdate contains an all the information needed to change an object's status to perform a specific update.
+// Send down a channel to the goroutine that actually writes the changes back.
+type StatusUpdate struct {
+	FullName FullName
+	GVR      schema.GroupVersionResource
+	Mutator  StatusUpdateMutator
+}
+
+// StatusUpdateMutator is an interface to hold mutator functions for status updates.
+type StatusUpdateMutator interface {
```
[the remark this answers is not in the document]
youngnick: You are technically correct, the best kind of correct. :)

internal/k8s/statusaddress.go
youngnick: Agreed, but I'll still need a local function so that there's only one place where that mutation function is kept.

internal/k8s/statusupdater.go
youngnick: I'm leaving it as FullName, so that when you're retrieving the name you get `x.FullName.Name` and `x.FullName.Namespace`, rather than `x.Name.Name` and `x.Name.Namespace`.

internal/k8s/statusupdater.go
youngnick: Moved everything to `name` and `namespace` fields.

internal/k8s/statusupdater.go
youngnick: Renaming to Update and using a `FullName` are both great ideas, thanks.

youngnick: Hmm, `Ingress` objects don't have an accessor that gives you the `FullName`, I don't think that changing to `FullName` here adds much except the need to make a new `FullName{}` struct whenever we call `Update`.

PR #2534, opened 2020-05-18
update integration tests to use non-standard ports by default

### LINKED ISSUE #2533: _integration: kind cluster should not require standard ports for integration tests
The integration tests currently utilize standard ports (e.g. `80/443`) for tests. This conflicts with other environments running on the same machine. 

The integration tests should be run on non-standard ports so Contour integration tests should run on the same machine.

stevesloka:
Fixes #2533


stevesloka:
yup I like that much better @jpeach!. Do you want to land this or update to just use that style?

jpeach:
> yup I like that much better @jpeach!. Do you want to land this or update to just use that style?

Let's land this as is, then update in a followup.

PR #2535, opened 2020-05-18
docs: add fallback certificate docs

stevesloka:
Updates #1503 by adding fallback certificate documentation.

### DIFF
--- a/site/docs/master/httpproxy.md
+++ b/site/docs/master/httpproxy.md
@@ -5,7 +5,7 @@
 The [Ingress][1] object was added to Kubernetes in version 1.1 to describe properties of a cluster-wide reverse HTTP proxy.
 Since that time, the Ingress object has not progressed beyond the beta stage, and its stagnation inspired an [explosion of annotations][2] to express missing properties of HTTP routing.
 
-The goal of the `HTTPProxy` (previously `IngressRoute`) Custom Resource Definition (CRD) is to expand upon the functionality of the Ingress API to allow for a richer user experience as well addressing the limitations of the latter's use in multi tenent environments.
+The goal of the HTTPProxy (previously `IngressRoute`) Custom Resource Definition (CRD) is to expand upon the functionality of the Ingress API to allow for a richer user experience as well addressing the limitations of the latter's use in multi tenent environments.
 
 ## Key HTTPProxy Benefits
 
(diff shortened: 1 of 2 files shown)


site/docs/master/httpproxy.md
```diff
@@ -252,6 +252,54 @@ The TLS **Minimum Protocol Version** a vhost should negotiate can be specified b
...
 
+##### Fallback Certificate
+
+Contour provides virtual host based routing, so that any TLS request is routed to the appropriate service based on the hostname specified in the request's HOST header.
+As HOST Header is encrypted during TLS handshake, it can’t be used for virtual host based routing unless client sends HTTPS request specifying hostname using the SNI or the request is first decrypted using a default TLS certificate.
+Some users need to support clients which may not be using SNI.
+When an HTTPS request is received, Envoy needs to first decrypt the request using a default TLS certificate and then based on the HOST header, route it to appropriate service.
+Contour provides a way to define a default or fallback TLS certificate for an TLS request which does not provide SNI in the request.
+
+_Note: The minimum tls protocol for any fallback request is defined by the `minimumProtocolVersion` set in the Contour configuration file._
```
jpeach, line 263: "minimum TLS protocol version"

site/docs/master/httpproxy.md
```diff
@@ -252,6 +252,54 @@ The TLS **Minimum Protocol Version** a vhost should negotiate can be specified b
 - 1.2
 - 1.1 (Default)
 
+##### Fallback Certificate
+
+Contour provides virtual host based routing, so that any TLS request is routed to the appropriate service based on the hostname specified in the request's HOST header.
```
jpeach, line 257: "based on both the server name requested by the TLS client and the HOST header in the HTTP request".

site/docs/master/httpproxy.md
```diff
@@ -252,6 +252,54 @@ The TLS **Minimum Protocol Version** a vhost should negotiate can be specified b
...
+As HOST Header is encrypted during TLS handshake, it can’t be used for virtual host based routing unless client sends HTTPS request specifying hostname using the SNI or the request is first decrypted using a default TLS certificate.
+Some users need to support clients which may not be using SNI.
```
jpeach, line 259: I think these 2 sentence could be clarified. The issue here is that legacy TLS client don't send the server name, so we don't know how to select the right certificate, therefore a fallback certificate is needed for these clients (i.e. what you say in the last sentence)

site/docs/master/httpproxy.md
```diff
@@ -252,6 +252,54 @@ The TLS **Minimum Protocol Version** a vhost should negotiate can be specified b
...
+When an HTTPS request is received, Envoy needs to first decrypt the request using a default TLS certificate and then based on the HOST header, route it to appropriate service.
+Contour provides a way to define a default or fallback TLS certificate for an TLS request which does not provide SNI in the request.
+
+_Note: The minimum tls protocol for any fallback request is defined by the `minimumProtocolVersion` set in the Contour configuration file._
+
```
jpeach, line 264: Also mention that this conflicts with TLS client authentication.

site/docs/master/httpproxy.md
```diff
@@ -252,6 +252,54 @@ The TLS **Minimum Protocol Version** a vhost should negotiate can be specified b
...
+###### Configuration
+
+First define the `namespace/name` in the [Contour configuration file][10] of a Kubernetes secret which will be used as the fallback certificate.
+Next configure `TLSCertificateDelegation` to specific namespaces or all (e.g. `*`) which should be allowed to enable the fallback certificate.
+Any `HTTPProxy` which enables fallback certificate delegation must have the fallback certificate delegated to the namespace in which the `HTTPProxy` object resides.
```
jpeach, line 269: Suggest that we swap these sentences around. Then it will read something like
- first define the secret (I assume you need to roll the Contour deployment?)
- any proxy that enables fallback must have certificate delegation
- therefore, configure a TLSCertificate delegation to the proxies that need it

site/docs/master/httpproxy.md
```diff
@@ -252,6 +252,54 @@ The TLS **Minimum Protocol Version** a vhost should negotiate can be specified b
...
+Finally, for each root `HTTPProxy`, set the `Spec.TLS.enableFallbackCertificate` parameter to allow that fqdn to opt-in to the fallback certificate routing.
```
jpeach, line 270: Rather than "fqdn" here, suggest that we use "to allow that `HTTPProxy`" or even "allow that domain"

stevesloka: Was just trying to match the language in the document. We call it a `fqdn` not `domain`, so was thinking that using the same terminology would be less confusing to users.

site/docs/master/httpproxy.md
```diff
@@ -252,6 +252,53 @@ The TLS **Minimum Protocol Version** a vhost should negotiate can be specified b
 - 1.2
 - 1.1 (Default)
 
+##### Fallback Certificate
+
+Contour provides virtual host based routing, so that any TLS request is routed to the appropriate service based on both the server name requested by the TLS client and the HOST header in the HTTP request. 
+
+As HOST Header is encrypted during TLS handshake, it can’t be used for virtual host based routing unless client sends HTTPS request specifying hostname using the SNI or the request is first decrypted using a default TLS certificate.
+Some TLS clients do not sent the server name, so Envoy does not know how to select the right certificate, therefore a fallback certificate is needed for these clients.
```
jpeach, line 260:
```suggestion
Some legacy TLS clients do not send the server name, so Envoy does not know how to select the right certificate. A fallback certificate is needed for these clients.
```

site/docs/master/httpproxy.md
jpeach: Ah I see it used above in the "TLS" section. I think the change you made reads well though.

site/docs/master/httpproxy.md
stevesloka: Why does it have to be "legacy" clients?

jpeach: Because SNI has been standardized for at least a decade. SNI is our preferred path because we are virtual hosting and don't support the usual workarounds (certificates per address, etc).

stevesloka: Right but I don't know why we need to categorize it as such. Built 20 years ago or last week as "cloud-native", if the client doesn't support SNI, then this is the path they need to follow. 

I'll add "legacy" so that we can get this closed for the release next week.

PR #2542, opened 2020-05-19
cmd/contour: support comma-separated ingress-status-address

### LINKED ISSUE #2540: --ingress-status-address supports only a single address
See #2539.

I tried using `--ingress-status-address` as a workaround, with a list of comma-separated IP addresses. The result was an error message

    time="2020-05-19T05:54:53Z" level=error msg="unable to update status" context=IngressStatusUpdater error="Ingress.extensions \"nginx-test\" is invalid: status.loadBalancer.ingress[0].hostname: Invalid value: \"192.168.1.17,192.168.1.18,192.168.1.19,192.168.1.20\": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')" name=nginx-test namespace=ajh

I expected all of the addresses to appear on my Ingress.

It's true that I could specify a DNS name that points to all of those IP addresses. It's also true that the result of doing that, if I want to use External DNS (and who doesn't?), is that I'll end up with a CNAME record. I don't want CNAME records. So I'd like to set multiple IP addresses.

Or, actually, for Contour to do what `ingress-nginx` does, give or take, and give the IP addresses of all of the nodes on which the `envoy` Pods are running (see #2539), then I wouldn't need to use this option in the first place.

**Environment:**

- Contour version: e1961860
- Kubernetes version: (use `kubectl version`): Client Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.6-beta.0", GitCommit:"e7f962ba86f4ce7033828210ca3556393c377bcc", GitTreeState:"clean", BuildDate:"2020-01-15T08:26:26Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.5", GitCommit:"e0fccafd69541e3750d460ba0f9743b90336f24f", GitTreeState:"clean", BuildDate:"2020-04-16T11:35:47Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/arm64"}
- Kubernetes installer & version: kubeadm version: &version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.5", GitCommit:"e0fccafd69541e3750d460ba0f9743b90336f24f", GitTreeState:"clean", BuildDate:"2020-04-16T11:41:38Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/arm64"}
- Cloud provider or hardware configuration: 4 x Raspberry Pi 4 (4GB)
- OS (e.g. from `/etc/os-release`): Ubuntu 20.04 LTS

al45tair:
When using a cluster without a LoadBalancer, it's common to configure
Envoy to use host networking and to run on some number of nodes (maybe
all of them, using a DaemonSet).

In that case, Contour won't pick up the IP addresses automatically as
there's no LoadBalancer; one approach to working around this is to
supply the --ingress-status-address argument, but this only supported
a single address.

This changeset adds support for multiple addresses, separated by
commas.

Fixes #2540

Signed-off-by: Alastair Houghton <alastair@alastairs-place.net>

### DIFF
--- a/cmd/contour/ingressstatus.go
+++ b/cmd/contour/ingressstatus.go
@@ -15,6 +15,7 @@ package main
 
 import (
 	"net"
+	"strings"
 	"sync"
 
 	"github.com/projectcontour/contour/internal/k8s"
@@ -108,24 +109,30 @@ func (isw *loadBalancerStatusWriter) Start(stop <-chan struct{}) error {
 
 func parseStatusFlag(status string) v1.LoadBalancerStatus {
 
-	// Use the parseability by net.ParseIP as a signal, since we need
-	// to pass a string into the v1.LoadBalancerIngress anyway.
-	if ip := net.ParseIP(status); ip != nil {
-		return v1.LoadBalancerStatus{
-			Ingress: []v1.LoadBalancerIngress{
-				{
-					IP: status,
-				},
-			},
+	// Support ','-separated lists.
+	ingresses := []v1.LoadBalancerIngress{}
+
+	for _, item := range strings.Split(status, ",") {
+		item = strings.TrimSpace(item)
+		if len(item) == 0 {
+			continue
+		}
+
+		// Use the parseability by net.ParseIP as a signal, since we need
+		// to pass a string into the v1.LoadBalancerIngress anyway.
+		if ip := net.ParseIP(item); ip != nil {
+			ingresses = append(ingresses, v1.LoadBalancerIngress{
+				IP: item,
+			})
+		} else {
+			ingresses = append(ingresses, v1.LoadBalancerIngress{
+				Hostname: item,
+			})
 		}
 	}
 
 	return v1.LoadBalancerStatus{
-		Ingress: []v1.LoadBalancerIngress{
-			{
-				Hostname: status,
-			},
-		},
+		Ingress: ingresses,
 	}
 }
 
--- a/cmd/contour/ingressstatus_test.go
+++ b/cmd/contour/ingressstatus_test.go
@@ -59,6 +59,94 @@ func Test_parseStatusFlag(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:   "WhitespacePadded",
+			status: "  anarbitrarystring      ",
+			want: v1.LoadBalancerStatus{
+				Ingress: []v1.LoadBalancerIngress{
+					{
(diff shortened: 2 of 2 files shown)


cmd/contour/ingressstatus.go
```diff
@@ -108,24 +109,30 @@ func (isw *loadBalancerStatusWriter) Start(stop <-chan struct{}) error {
...
-	// Use the parseability by net.ParseIP as a signal, since we need
-	// to pass a string into the v1.LoadBalancerIngress anyway.
-	if ip := net.ParseIP(status); ip != nil {
-		return v1.LoadBalancerStatus{
-			Ingress: []v1.LoadBalancerIngress{
-				{
-					IP: status,
-				},
-			},
+	// Support ','-separated lists
```
jpeach, line 112: Nit. Could you please end comments with a full stop?

cmd/contour/ingressstatus.go
```diff
@@ -108,24 +109,30 @@ func (isw *loadBalancerStatusWriter) Start(stop <-chan struct{}) error {
...
+	items := strings.Split(status, ",")
+	ingresses := make([]v1.LoadBalancerIngress, len(items))
+	count := 0
+
+	for _, item := range items {
```
jpeach, line 117: I think that we can drop the preallocation and simplify. Then `append` the items and don't worry about tracking the count.

```suggestion
	ingresses := []v1.LoadBalancerIngress{}
	
	for _, item := range strings.Split(status, ",") {
```

cmd/contour/ingressstatus.go
al45tair: Done.

cmd/contour/ingressstatus.go
al45tair: I'd automatically written it that way to avoid (theoretical) extra memory allocations. But of course, the chances are that the number of items is small anyway, in which case it hardly matters and you're right, it seems simpler not to.

Done.

PR #2545, opened 2020-05-21
test: switch to IfNotPresent image pull policy

### LINKED ISSUE #2537: Stop using ImagePullPolicy of Always

Spent 4 hours trying to figure out why pushing Contour images into a local kind cluster didn't work properly.

The underlying problem is that we always specify an `ImagePullPolicy` of `Always` for everything, which makes contained fetch upstream images to replace the ones we pushed, and makes it ignore existing images that we places in the cluster.

Unless there is a special case (e.g. certgen job), we should default to `IfNotPresent` to preserve the bandwidth on users clusters.

jpeach:
When deploying release builds, Contour images are tagged with the release
version and are intended to be immutable. This means we can generally use
the `IfNotPresent` image pull policy, except for cthe certgen job, which
needs to use `Always` because it pulls the mutable `:latest` image tag.

This fixes #2537.

### DIFF
--- a/_integration/testsuite/make-kind-cluster.sh
+++ b/_integration/testsuite/make-kind-cluster.sh
@@ -27,6 +27,12 @@ readonly WAITTIME=${WAITTIME:-5m}
 readonly HERE=$(cd $(dirname $0) && pwd)
 readonly REPO=$(cd ${HERE}/../.. && pwd)
 
+# List of tags to apply to the image built from the working directory.
+# The "working" tag is applied to unambigiously reference the working
+# image, since "master" and "latest" could also come from the Docker
+# registry.
+readonly TAGS="master latest working"
+
 kind::cluster::exists() {
     ${KIND} get clusters | grep -q "$1"
 }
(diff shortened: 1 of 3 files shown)


jpeach:
@youngnick I found that a tooling change I made in #2184 was forcing the quickstart pull policy to always be `IfNotPresent`. I've reverted that so that the certgen job gets the right policy.

jpeach:
> I'm on board with using `IfNotPresent` for the versioned specs, however, I think for `master`, it shouldn't change. I know folks shouldn't use master, but with this change, they'll never pull the latest version of `master`.

Yeh you have a good point for that use case, but I also don't think that we can support all these use cases at the same time. Currently, we sed the image pull policy for the release YAML but get it wrong for certgen. We have shipped that mistake for a couple of releases now. It's difficult to switch the image pull policy correctly without either tooling or manual changes.

For testing in Kind, I really need the policy to *not* be "Always" because I never want to used DockerHub issues; I always want to use the version that is pushed into kind.

jpeach:
@stevesloka I removed the changes to the example deployments, PTAL

danehans:
xref: https://github.com/projectcontour/contour-operator/pull/33 has been updated to use `IfNotPresent`.

PR #2547, opened 2020-05-21
certgen: add cert-manager certificate compatibility

### LINKED ISSUE #2494: Certgen should generate secrets that are compatible with cert-manager
As noted by @tsaarni in https://github.com/projectcontour/contour/issues/2149, the secret that certgen generates isn't interchangeable with how cert-manager generates secrets.

Maybe we can update certgen to be compatible with cert-manager, which would make it easier for sites to switch.

jpeach:
This change adds new options to certgen that will allow it to overwrite
existing Kubernetes Secrets, and to generate Secrets in a format that
is compatible with cert-manager (we call this "compact" format).

The `--overwrite` flag can be used to let certgen update existing secrets,
and this will enable both the transition to the new format and certificate
rotation on a per-release granularity.

The `--secrets-format` flag switches the Secrets output format from the
current format to the compatible "compact" format. Once this is enabled
in the deployment YAML, operators will easily be able to substitute
cert-manager certificates for certgen certificates.

Finally, the bootstrap command is updated to ensure that the specified
certificate files are actually present. This prevents Envoy starting up
and rejecting the static configuration, which is unrecoverable.

This fixes #2494.


jpeach:
@tsaarni This PR adds all the machinery for xDS certificate rotation and transitioning to xDS Secrets that are compatible with cert-manager. Would be great if you cold do a review pass.

internal/certgen/certgen.go
```diff
@@ -39,107 +55,107 @@ func writePEM(outputDir, filename string, data []byte) error {
...
 }
 
-func writeCACertKube(client *kubernetes.Clientset, namespace string, cert []byte) error {
-	secret := newCertOnlySecret("cacert", namespace, "cacert.pem", cert)
-	if _, err := client.CoreV1().Secrets(namespace).Create(context.TODO(), secret, metav1.CreateOptions{}); err != nil {
-		if k8serrors.IsAlreadyExists(err) {
-			fmt.Print("secret/cacert already exists\n")
-			return nil
+// WriteSecretsKube writes all the keypairs out to Kubernetes Secrets in the
+// compact format which is compatible with Secrets generated by cert-manager.
```
stevesloka, line 100: If cert-manager is being used, why have this in the mix? Shouldn't cert-manager handle all the cert generation?

cmd/contour/bootstrap.go
```diff
@@ -21,7 +21,9 @@ import (
 // registerBootstrap registers the bootstrap subcommand and flags
 // with the Application provided.
 func registerBootstrap(app *kingpin.Application) (*kingpin.CmdClause, *envoy.BootstrapConfig) {
-	var config envoy.BootstrapConfig
+	config := envoy.BootstrapConfig{
+		CheckFilePaths: true,
```
stevesloka, line 25: This is always true right? Why have the option?

internal/certgen/certgen.go
youngnick: It's so that you can start with `certgen` and then change your mind and move to cert-manager, or vice versa.

jpeach: Yup; for general interoperability without having to fork and own the deployment YAML.

cmd/contour/bootstrap.go
jpeach: So that testing doesn't require the actual files. I could flip the default around though?

internal/certgen/certgen.go
tsaarni: Sorry, I now become bit confused about this topic in general :)  So I'd would like to ask you if this reasoning holds:

* The default example deployment should start to use `contour certgen` "compat" option - The motivation is to allow converting any new Contour deployment to cert-manager easily.
* The "legacy" option is still default behavior for `contour certgen` - The motivation is to allow people to upgrade their existing deployments by just changing the version tag.

On the other hand, all existing deployments already have their legacy secrets generated previously.  So now my confusion is: what is the scenario when new version of `contour certgen` needs to generate **old** secret?

Isn't the backwards compatibility problem rather coming from the changes needed in volume mounts and command line parameters (i.e. the last four bullets in https://github.com/projectcontour/contour/issues/2149#issuecomment-598065053)?

I guess this question could potentially be quite fundamental for this PR, sorry for being late on this!!!

internal/certgen/k8s.go
```diff
@@ -14,51 +14,42 @@
...
-			corev1.TLSCertKey:       certPEM,
-			corev1.TLSPrivateKeyKey: keyPEM,
-		},
+		Data: data,
 	}
 }
 
-func newCertOnlySecret(secretname, namespace, certfilename string, certPEM []byte) *corev1.Secret {
-
+func newCertOnlySecret(name string, namespace string, data map[string][]byte) *corev1.Secret {
```
tsaarni, line 39: Now that both `newTLSSecret()` and `newCertOnlySecret()` both take map of data as parameter, the only difference left seems to be the secret `Type`.  Maybe consider combining these into single function and also pass secret type as a parameter?

jpeach:
@stevesloka @youngnick With the "compact" secrets format, should we still create the `cacert` secret with a single `ca.crt` data element?

internal/certgen/certgen.go
jpeach: The main motivation for adding the new flags is so that we can make the configuration change separately (and revert it if I screw it up). After this PR, the default `certgen` behavior will be unchanged, so I can then make a separate PR to enable the new options along with the deployment changes. This way, the repo moves from one consistent state to another, and we can safely stop at any time.

internal/certgen/k8s.go
jpeach: Fixed.

internal/certgen/certgen.go
```diff
@@ -20,16 +20,32 @@ import (
...
+	ContourPrivateKeyKey  = "contourkey.pem"
+	EnvoyCertificateKey   = "envoycert.pem"
+	EnvoyPrivateKeyKey    = "envoykey.pem"
+)
+
+type OverwritePolicy int
+
+const (
+	Overwrite   OverwritePolicy = iota
+	NoOverwrite OverwritePolicy = iota
```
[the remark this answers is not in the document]
jpeach: Fixed.

youngnick:
> @stevesloka @youngnick With the "compact" secrets format, should we still create the `cacert` secret with a single `ca.crt` data element?

If that will help with migration, yes, otherwise, no.

jpeach:
> > @stevesloka @youngnick With the "compact" secrets format, should we still create the `cacert` secret with a single `ca.crt` data element?
> 
> If that will help with migration, yes, otherwise, no.

Pretty sure it's not needed for migration.

cmd/contour/bootstrap.go
tsaarni: I always get confused about negative flags since my mind does not easily wrap around something like `dont = false`, especially when it might later get used in ways like `if dont == false { how_did_we_end_up_here? }` :-)

jpeach: Renamed to `SkipFilePathCheck`

PR #2550, opened 2020-05-25
test: add URL construction helpers

jpeach:
The integration testsuite uses passed-in parameters to construct a
URL that can be used with the target load balancer. Add a `url` helper
package to use those parameters to construct URL strings so that the
tests themselves can be eadier to read and less fiddly.


_integration/testsuite/policies/contour-client.rego
```diff
@@ -14,45 +14,35 @@
...
 # send requests through.
 target_http_port = port {
-  port := data.test.params.proxy.http_port
-}
-
-# target_https_port returns the TLS port for the proxy that tests should
-# send requests through.
-target_https_port = "9143" {
-  not data.test.params.proxy.https_port
+  trace(sprintf("^^^^ port=%s", [data.test.params.proxy.http_port]))
```
stevesloka, line 28: debug?

_integration/testsuite/policies/contour-url.rego
```diff
@@ -0,0 +1,35 @@
...
+
+package contour.http.client.url
+
+import data.contour.http.client
+
+# Generate a HTTP URL from the `target_addr` and `target_http_port`
+# parameters.  The path argument specifies the full URL path and must
+# include the leading '/'.
+http(path) = url {
+    url := sprintf("http://%s:%d%s", [
```
stevesloka, line 23: Should we have a check that the path begins with a `/`?

_integration/testsuite/policies/contour-client.rego
jpeach: Oops, good catch!

_integration/testsuite/policies/contour-url.rego
jpeach: Yeh, I got a bit lazy here, but you are right that we can do a bit better here. I'll look into adding the leading '/' only if it is necessary.

stevesloka: I guess it's not dynamic, so it shouldn't really come up. ¯\_(ツ)_/¯

_integration/testsuite/policies/contour-client.rego
jpeach: Fixed.

_integration/testsuite/policies/contour-url.rego
jpeach: Fixed.

PR #2553, opened 2020-05-27
examples/contour: switch to reloadable xDS certificates

### LINKED ISSUE #2524: Update deployment to use rotatable bootstrap config

Update the default deployment YAML to generate an Envoy bootstrap that is able to rotate xDS certificates (as per #2333)

jpeach:
Switch the certgen xDS certificate generation over to using certificates
that are compatible with certificate-manager. This requires the certgen
job to run again, which means we have to give is a unique name. Using
a version-locked name will let it run on each upgrade, which will have
the beneficial side-effect of rotating the xDS certificates.

The envoy and contour pods will restart because the secrets mounts change,
but they will restart at release time anyway because the container image
will change.

After this change, re-running certgen to rotate the xDS certificates
will not require restarting and pods.

This fixes #2524.
This updates #2143.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ jobs:
     - $HOME/bin/kind create cluster --wait 2m
     - $HOME/bin/kind load docker-image docker.io/projectcontour/contour:master
     - $HOME/bin/kind load docker-image docker.io/projectcontour/contour:latest
-    - $HOME/bin/kubectl apply -f examples/render/contour.yaml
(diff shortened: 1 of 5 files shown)


jpeach:
Tested locally with kind.

1. Install Contour using https://projectcontour.io/quickstart/v1.4.0/contour.yaml
2. Upgrade to working copy using `_integration/testsuite/make-release-cluster.sh`
3. Ensure pods come back and can serve ingresses

For (2), you need to hack the script to not exit when the test cluster already exists.

PR #2569, opened 2020-06-05
internal/dag: remove Envoy header name quoting

### LINKED ISSUE #2516: Add support to set Header value from another header
I have some Third Party LoadBalancers, on which i don't have any Control, setting the `Real IP` header to a custom header like `x-real-client-ip` rather than the usual `x-forwarded-for` 

In my current `haproxy ingress` , which i am trying to replace with contour , i simply have a frontend snippet like 

```
acl acl_real_client_ip_header_exists req.hdr(X-Real-Client-IP) -m found
http-request set-header X-Forwarded-For %[req.hdr(X-Real-Client-IP)] if acl_real_client_ip_header_exists
http-request set-header X-Forwarded-For %[src] if !acl_real_client_ip_header_exists
```

* If the `header X-Real-Client-IP` exist then set `x-forwarded-for` to the value of `x-real-client-ip`
* else set `x-forwarded-for` to the `src ip` 

Ideally i would expect something like 
```
  routes:
    - services:
      - name: podinfo
        port: 8443
        responseHeadersPolicy:
          set:
              - name: x-forwarded-for
                valueFromHeader: x-real-client-ip
                fallback: ignore
```

To copy the value if present and `ignore` if not present ( and hopefully still get the standard `x-forwarded-for` behavior from contour

jpeach:
Remove the '%' quoting on Envoy header names so that operators can
configure Envoy header variables.

This fixes #2516.

### DIFF
--- a/_integration/testsuite/httpproxy/011-header-envoy-variables.yaml
+++ b/_integration/testsuite/httpproxy/011-header-envoy-variables.yaml
@@ -1,0 +1,152 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: ingress-conformance-echo
+$apply:
+  fixture:
+    as: echo
+
+---
+
+apiVersion: v1
+kind: Service
+metadata:
+  name: ingress-conformance-echo
+$apply:
+  fixture:
+    as: echo
+
+---
+
+apiVersion: projectcontour.io/v1
+kind: HTTPProxy
+metadata:
+   name: echo
+spec:
+  virtualhost:
+    fqdn: echo.projectcontour.io
+  routes:
+  - services:
+    - name: echo
+      port: 80
+    requestHeadersPolicy:
+      set:
+      - name: Test-User-Agent
+        value: '%REQ(User-Agent)%'
+      remove:
+      - Test-Removed-Header
+    responseHeadersPolicy:
+      set:
+      - name: Test-Response-Protocol
+        value: '%PROTOCOL%'
+      remove:
+      - X-Envoy-Upstream-Service-Time
+
+---
+
+import data.contour.resources
+
+fatal_proxy_is_not_valid[msg] {
+  name := "echo"
+  proxy := resources.get("httpproxies", name)
+  status := object.get(proxy, "status", {})
+
+  object.get(status, "currentStatus", "") != "valid"
+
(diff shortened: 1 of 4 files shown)


jpeach:
@youngnick This is a draft because it's possible to have Envoy reject the config update (e.g. if you use a variable name that isn't known to Envoy). I'll un-draft if we have consensus that we are OK with that.

jpeach:
xref https://github.com/projectcontour/contour/issues/1176

youngnick:
If the update is rejected, do we end up in the state where Envoy will not accept config until a restart?

jpeach:
> If the update is rejected, do we end up in the state where Envoy will not accept config until a restart?

Yes.

youngnick:
I guess the full questions I have are:
- If someone configures this and gets it wrong, how do they know? Currently, it seems that the only way they will know is when their Envoys stop accepting updates. 
- How could we build something to expose the errors to the user without needing to parse all the header values ourselves? I don't see how it's possible, frankly.

My position here is that, if Contour accepts the config, it's Contour's job to surface any problems with the config. With this PR, I don't see how that's possible. That's why we did not allow this level of configurability to begin with.

I appreciate that the functionality is useful, but I think that introducing a feature that may break all the Envoys for all users of Contour because one HTTPProxy has a typo in a header rewrite field is too risky.

jpeach:
> I guess the full questions I have are:
> 
> * If someone configures this and gets it wrong, how do they know? Currently, it seems that the only way they will know is when their Envoys stop accepting updates.

Right. There's no way to attribute the error to any specific proxy document.

> * How could we build something to expose the errors to the user without needing to parse all the header values ourselves? I don't see how it's possible, frankly.

Right again. The canonical list of header values is in Envoy source code, and it's not readily extractible.

> My position here is that, if Contour accepts the config, it's Contour's job to surface any problems with the config. With this PR, I don't see how that's possible. That's why we did not allow this level of configurability to begin with.
> 
> I appreciate that the functionality is useful, but I think that introducing a feature that may break all the Envoys for all users of Contour because one HTTPProxy has a typo in a header rewrite field is too risky.

Yep, it's unfortunate but I agree that it makes sense, particularly from a multi-team perspective.

PR #2583, opened 2020-06-15
cmd/contour: exclusively use dynamic Informers

jpeach:
This change standardizes Contour to always use dynamic Informers. This
makes the startup code path a little simpler, and lets us consolidate
on a single method of handling object updates. Update the downstream
handlers to fully support unstructured objects.

### DIFF
--- a/internal/k8s/converter.go
+++ b/internal/k8s/converter.go
@@ -15,12 +15,10 @@ package k8s
 
 import (
 	"context"
-	"fmt"
 
 	ingressroutev1 "github.com/projectcontour/contour/apis/contour/v1beta1"
 	projectcontour "github.com/projectcontour/contour/apis/projectcontour/v1"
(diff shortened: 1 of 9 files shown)


internal/k8s/converter.go
```diff
@@ -122,68 +108,22 @@ func NewUnstructuredConverter() (*UnstructuredConverter, error) {
...
-		err := c.scheme.Convert(obj, hr, nil)
-		return hr, err
-	case "TcpRoute":
-		tr := &serviceapis.TcpRoute{}
-		err := c.scheme.Convert(obj, tr, nil)
-		return tr, err
-	default:
-		return nil, fmt.Errorf("unsupported object type: %T", obj)
+
+	newObj, err := c.scheme.New(u.GetObjectKind().GroupVersionKind())
```
youngnick, line 116: Neat, this is much better.

Issue #2579, opened 2020-06-12
[security] Upgrade to Envoy 1.14.2


pims:
Due to CVE https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-11080 upgrading to Envoy 1.14.2 is recommended.

https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.14.2

Happy to submit a PR if you’d like.

jpeach:
Yup, there will be  new security release soon. @pims if you would like to be the release manager for this, I'd be glad to shepherd that!

pims:
@jpeach I’d love to!

jpeach:
> On 13 Jun 2020, at 12:27 pm, Tim Bart <notifications@github.com> wrote:
> 
> 
> @jpeach I’d love to!

Cool, ping me on Kubernetes slack.

jpeach:
Steps we need to do to make a 1.5.1 release:

- Update the envoy image version on master
- Update the envoy image version on the release-1.5 branch
- Update hack/release/prepare-release.go to optionally take an OLDVERS NEWVERS arguments so that we can cut the 1.5.0 docs across to 1.5.1

Then we can:

- cut the 1.5.1 tag
- push the 1.5.1 Docker image
- create and update GitHub release notes

Then, various docs updates (on master):
- https://github.com/projectcontour/contour/blob/master/site/_resources/envoy.md
- https://github.com/projectcontour/contour/blob/master/site/_resources/upgrading.md

The release process is documented [here](https://projectcontour.io/resources/release-process/), so if you find any inaccuracies we can correct them as we go.

pims:
`1.5.1` is out: https://github.com/projectcontour/contour/releases/tag/v1.5.1

PR #2585, opened 2020-06-15
Update prepare-release to accept old version as doc source

pims:
Example use:

```
prepare-release v1.5.0 v1.5.1 

Verifying repository state ...
Cloning versioned documentation ...
modified:   site/_config.yml
modified:   site/_data/toc-mapping.yml
new: site/_data/v1-5-1-toc.yml
new: site/docs/v1.5.1/
```

### DIFF
--- a/hack/release/prepare-release.go
+++ b/hack/release/prepare-release.go
@@ -138,13 +138,19 @@ values:
 }
 
 func main() {
-	if len(os.Args) != 2 {
-		fmt.Printf("Usage: %s NEWVERS\n",
-			path.Base(os.Args[0]))
+	oldVers := "master"
+	newVers := ""
+
+	switch len(os.Args) {
+	case 2:
+		newVers = os.Args[1]
+	case 3:
+		oldVers = os.Args[1]
+		newVers = os.Args[2]
+	default:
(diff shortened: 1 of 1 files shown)


pims:
This updates #2579

hack/release/prepare-release.go
```diff
@@ -138,13 +138,17 @@ values:
...
+		fmt.Printf("Usage: %s NEWVERS [OLDVERS]\n",
 			path.Base(os.Args[0]))
 		os.Exit(1)
 	}
 
 	newVers := os.Args[1]
+	oldVers := "master"
+	if len(os.Args) == 3 {
+		oldVers = os.Args[2]
+	}
```
jpeach, line 151: Let's switch the order of OLDVERS and NEWVERS, so that it is the same as for `make-release-tag.sh`.

Maybe something like this:
```Go

oldVers := master
newVers := ""

switch len(os.Args) {
case 2:
        newVers := os.Args[1]
case 3:
        oldVers := os.Args[1]
        newVers := os.Args[2]
default:
        fmt.Printf("Usage: %s NEWVERS | OLDVERS NEWVERS\n", ...)
}
```

Issue #2572, opened 2020-06-09
Remove IngressRoute handling code


youngnick:
This issue covers removing IngressRoute handling from Contour.

I'll try and break the removal up into a series of PRs for ease of review.

To do:
- [x] Remove watches on IngressRoute objects, so that the code paths can't be exercised.
- [x] Remove code from `internal/contour`
- [x] Remove code from `internal/dag`
- [x] Remove code from `internal/k8s`

Epic for this is #1789.

youngnick:
Hmm, I made a mistake, this is closed now by #2593.

PR #2590, opened 2020-06-16
internal/dag: Remove IngressRoute Code

youngnick:
Update `internal/e2e` and `internal/featuretests` as well.

Updates #2572.


youngnick:
I am now also annoyed by the tests enough to want to come back and tidy them up; but I don't think this is the time to do it. Thanks for the review.

PR #2593, opened 2020-06-17
Remove remaining IngressRoute code

youngnick:
Updates #2572.

I've left type switches with only one (HTTPProxy) type intact for now, as we will need
to add the Service-APIs types later, and some stuff will be moved or removed as
part of the status updates.

Also remove all the code from `internal/`.
Remove the CRDs from code generation and YAML generation.


examples/README.md
```diff
@@ -18,7 +18,7 @@ Single file renderings of other examples suitable for `kubectl apply`ing via a U
 
 ## `example-workload`
 
-IngressRoute examples under the `example-workload/ingressroute` directory. See the [README](./example-workload/ingressroute/README.md) for more details on each example.
+HTTPPROXY examples under the `example-workload/httpproxy` directory. See the [README](./example-workload/httpproxy/README.md) for more details on each example.
```
jpeach, line 21: s/HTTPPROXY/HTTPProxy/

hack/generate-crd-deepcopy.sh
```diff
@@ -41,4 +41,4 @@ boilerplate > "${HEADER}"
 exec go run sigs.k8s.io/controller-tools/cmd/controller-gen \
     object \
     "object:headerFile=${HEADER}" \
-    paths=./apis/...
+    paths=./apis/projectcontour/...
```
jpeach, line 44: Is this change actually needed?

hack/generate-crd-yaml.sh
```diff
@@ -16,6 +16,6 @@ cd $REPO
 # so, output them to separate files, then concatenate those files.
 # That should give a stable sort.
 go run sigs.k8s.io/controller-tools/cmd/controller-gen \
-  crd paths=./apis/... output:dir=$TEMPDIR
+  crd paths=./apis/projectcontour/... output:dir=$TEMPDIR
```
jpeach, line 19: Same as above ... if it's not necessary, we should leave this as is.

hack/generate-crd-deepcopy.sh
youngnick: We could maybe leave this one, it just updates the deepcopy things.

hack/generate-crd-yaml.sh
youngnick: This one ensures that the example install no longer includes the `contour.heptio.com` CRDs.

youngnick: I will make sure that I update it back when we do remove the actual CRDs.

PR #2594, opened 2020-06-17
internal/k8s: remove redundant coordination client

skriss:

`Clients` currently instantiates a Kubernetes clientset plus a `CoordinationV1` client. Since the former already contains a `CoordinationV1` client, the latter is redundant, so this PR removes it.


skriss:
cc @stevesloka

PR #2600, opened 2020-06-17
internal/dag: allow routes with overlapping header conditions

### LINKED ISSUE #2597: HTTPProxy: some routes are ignored when they differ only on header condition operator or value
Define an HTTPProxy with several routes that all have the same prefix and all have a condition on the same header, but different operators or values (some examples below).

Observe that Envoy is only configured with the last route; the other ones are dropped.

```yaml
# two routes that both have a 'contains' condition on the
# same header, with two different values.
kind: HTTPProxy
metadata:
  name: dropped-routes-1
spec:
  virtualhost:
    fqdn: stevek.com
  routes:
    - conditions:
        - header:
            name: x-stevek
            contains: foo
      services:
        - name: svc1
          port: 80
    - conditions:
        - header:
            name: x-stevek
            contains: bar
      services:
        - name: svc2
          port: 80
---
# two routes that both have a condition on the
# same header -- one using 'contains' and one
# using 'notcontains'.
kind: HTTPProxy
metadata:
  name: dropped-routes
spec:
  virtualhost:
    fqdn: stevek.com
  routes:
    - conditions:
        - header:
            name: x-stevek
            contains: foo
      services:
        - name: svc1
          port: 80
    - conditions:
        - header:
            name: x-stevek
            notcontains: foo
      services:
        - name: svc2
          port: 80
```

All of the routes to be configured with Envoy.

skriss:
closes #2597 

Fixes a bug where an HTTPProxy with multiple routes with the same prefixes
and conditions on the same headers, differing only on the condition operators
or values, would result in only the last route being configured in Envoy and
all others being dropped.

### DIFF
--- a/internal/dag/dag.go
+++ b/internal/dag/dag.go
@@ -17,6 +17,7 @@ package dag
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 
@@ -79,7 +80,14 @@ type HeaderCondition struct {
 }
 
 func (hc *HeaderCondition) String() string {
-	return "header: " + hc.Name
+	details := strings.Join([]string{
+		"name=" + hc.Name,
+		"value=" + hc.Value,
+		"matchtype=", hc.MatchType,
+		"invert=", strconv.FormatBool(hc.Invert),
+	}, "&")
+
+	return "header: " + details
 }
 
 // Route defines the properties of a route to a Cluster.
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -1762,6 +1762,106 @@ func TestDAGInsert(t *testing.T) {
 		},
 	}
 
+	// proxy2d is a proxy with two routes that have the same prefix and a Contains header
+	// condition on the same header, differing only in the value of the condition.
+	proxy2d := &projcontour.HTTPProxy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "example-com",
+			Namespace: "default",
+		},
+		Spec: projcontour.HTTPProxySpec{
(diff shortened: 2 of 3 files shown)


internal/dag/dag.go
```diff
@@ -79,7 +80,14 @@ type HeaderCondition struct {
 }
 
 func (hc *HeaderCondition) String() string {
-	return "header: " + hc.Name
+	details := strings.Join([]string{
```
chintan8saaras, line 83: Maybe there are cases where different routes are required. So maybe this is the way to go.

However an incoming request (method:GET, Prefix:/) will match both -

```
(HeaderName: ":method", MatchType: "Exact", HeaderValue: "GET")
(HeaderName: ":method", MatchType: "Contains", HeaderValue: "GET")

```
Allowing routes differing on MatchType **may** create duplicates.

stevesloka: I think ideally we could detect if routes created won't match and set the status on the route, just not sure how difficult determining all those scenarios are. We'd have to match on duplicate header types of different types as @chintan8saaras noted.

skriss: Seems like there are a couple of related improvements we could make here:

1. within a single route, determine when one condition is a proper subset of another (e.g. "exact: foo" and "contains: foo" on the same header), and collapse them into just a single condition (in this case, just "exact: foo").
2. when two routes can be combined similarly to the above, collapse them into a single route
3. when there are conflicting header conditions for a given route (e.g. both "contains: foo" and "notcontains: foo"), mark the proxy as invalid. This is already done for conflicting "exact" conditions, but we could strengthen this validation.

PR #2604, opened 2020-06-18
replace current xds server with go-control-plane impl

### LINKED ISSUE #2134: Replace GRPC server with envoy's go-control-plane
I looked ad [go-control-plane](https://github.com/envoyproxy/go-control-plane) a bit and we ought to be able to use it to replace our custom xDS code. The interfaces are a bit different, but we should be able to bind the DAG in without a lot of trouble. This likely gives us ADS support for free.

Related #1286

stevesloka:
Fixes #2134 by implementing the Envoy [go-control-plane](https://github.com/envoyproxy/go-control-plane) to replace Contour's current custom xDS gRPC server. 

The change utilizes snapshots as a way to represent a versioned point in time representation of the xDS resources (RDS, CDS, EDS, LDS, SDS). When the dag is rebuilt or an endpoint changes, a new snapshot is created with the updated caches for each xDS resource type. 


stevesloka:
quick update, I got ADS working, but need to fix https://github.com/projectcontour/contour/issues/2680 first before this can move forward.

stevesloka:
This has a couple major type things  that I've encountered making it difficult. 

1. Many of our tests assert that specific items in xDS are empty. Things like, "there shouldn't be any listeners" at a specific point in the feature test, however, I've found that the xDS server hangs with this invalid config. An example is the root-namespaces feature test. It will run with the validations I have commented out (https://github.com/stevesloka/contour/blob/eebb6e546d640bfb0261708abb26157d3073a8ae/internal/featuretests/rootnamespaces_test.go#L31). 
2. The second bit that is an issue is the cache is not sorted when the xDS client requests. So each run of the test may pass or fail depending if there are multiple values in the list or just good luck with the resulting response (https://github.com/envoyproxy/go-control-plane/issues/218). 

The first issue isn't a difficult work around, just need to clean up all the bits for the tests. The second I'm not sure how to solve without changing the way we compare equality. 

@youngnick @jpeach @skriss any thoughts? We should chat about this on the next community meeting.

jpeach:
> This has a couple major type things that I've encountered making it difficult.
> 
> 1. Many of our tests assert that specific items in xDS are empty. Things like, "there shouldn't be any listeners" at a specific point in the feature test, however, I've found that the xDS server hangs with this invalid config. You'll see this in the tests where the unit tests literally just run for 10m before giving up. An example is the root-namespaces feature test. It will run with the validations I have commented out (https://github.com/stevesloka/contour/blob/eebb6e546d640bfb0261708abb26157d3073a8ae/internal/featuretests/rootnamespaces_test.go#L31).
> 2. The second bit that is an issue is the cache is not sorted when the xDS client requests. So each run of the test may pass or fail depending if there are multiple values in the list or just good luck with the resulting response ([envoyproxy/go-control-plane#218](https://github.com/envoyproxy/go-control-plane/issues/218)).
> 
> The first issue isn't a difficult work around, just need to clean up all the bits for the tests. The second I'm not sure how to solve without changing the way we compare equality.
> 
> @youngnick @jpeach @skriss any thoughts? We should chat about this on the next community meeting.

`testify` assertions are better about comparing collections for equality, though they are worse about comparing protobufs IME. Maybe that's fixable.

Otherwise, I'd investigate whether it's feasible to slip in some test framework shim to do the sorting.

youngnick:
Good to see this is getting closer!

I do have one request though - I think that this will be much easier to review if there's some explanation of how the snapshot handler fits in to how Contour used to work, and what it changes, either in a comment in the Github UI, or in docstring or comment in the code. You've gone on a long journey to understand all of this @stevesloka, bring us along with you.

stevesloka:
>  I think that this will be much easier to review if there's some explanation of how the snapshot handler fits in to how Contour used to work

For sure @youngnick, the initial attempt was to get it working, will come back now and clean everything up.

stevesloka:
I'm going to update the description to explain the flow a bit. Also, I want to run the full integration tests against this, but this is ready for a first glance.

stevesloka:
I need to also double check the contour cli commands, I don't think they will work just yet.

cmd/contour/serve.go
```diff
@@ -197,14 +203,28 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 
 	listenerConfig.DefaultHTTPVersions = defaultHTTPVersions
 
+	ch := &contour.CacheHandler{
+		ListenerVisitorConfig: contour.ListenerVisitorConfig{
```
skriss, line 207: you'll want to use the `listenerConfig` constructed above

cmd/contour/serve.go
```diff
@@ -172,6 +173,11 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		return err
 	}
 
+	// snapshotCache is used to store the state of what all xDS services should
+	// contain at any given point in time.
+	snapshotCache := cache.NewSnapshotCache(false, cache.IDHash{},
+		log.WithField("context", "XDSServer"))
+
```
skriss, line 180: nit: move this down to closer to where it's first used

internal/grpc/server.go
```diff
@@ -17,107 +17,40 @@ package grpc
...
-	discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v2"
-	loadstats "github.com/envoyproxy/go-control-plane/envoy/service/load_stats/v2"
-	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
 )
 
 // NewAPI returns a *grpc.Server which responds to the Envoy v2 xDS gRPC API.
-func NewAPI(log logrus.FieldLogger, resources map[string]Resource, registry *prometheus.Registry, opts ...grpc.ServerOption) *grpc.Server {
+func NewAPI(registry *prometheus.Registry, snapshotCache cache.SnapshotCache, opts ...grpc.ServerOption) *grpc.Server {
 	s := &grpcServer{
```
skriss, line 31: Seems like `s` and the `grpcServer` struct def can be removed since they're not being used, and the metrics can be handled separately.

internal/contour/cachehandler.go
```diff
@@ -47,6 +52,8 @@ func (ch *CacheHandler) OnChange(dag *dag.DAG) {
 	ch.updateRoutes(dag)
 	ch.updateClusters(dag)
 
+	ch.SnapshotHandler.UpdateSnapshot()
```
skriss, line 55: are the `internal/contour` caches even necessary anymore? It seems like in the `updateSecrets`, etc. methods here, we could just return the resources directly, and then pass them to `UpdateSnapshot`. I might be missing something about how the contour caches are used, though, so take my comment with a grain of salt :)

stevesloka: No I don't think they are, we could just pick them out of the dag rebuild. Endpoints currently are the only caches left that would need to stay since they are managed per change.

stevesloka:
Integration tests pass: 

```
./_integration/testsuite/httpproxy/001-required-field-validation.yaml     PASSED
./_integration/testsuite/httpproxy/002-header-condition-match.yaml        PASSED
./_integration/testsuite/httpproxy/003-path-condition-match.yaml          PASSED
./_integration/testsuite/httpproxy/004-https-sni-enforcement.yaml         PASSED
./_integration/testsuite/httpproxy/005-pod-restart.yaml                   PASSED
./_integration/testsuite/httpproxy/006-merge-slash.yaml                   PASSED
./_integration/testsuite/httpproxy/007-client-cert-auth.yaml              PASSED
./_integration/testsuite/httpproxy/008-tcproute-https-termination.yaml    PASSED
./_integration/testsuite/httpproxy/009-https-misdirected-request.yaml     PASSED
./_integration/testsuite/httpproxy/010-include-prefix-condition.yaml      PASSED
./_integration/testsuite/httpproxy/011-retry-policy-validation.yaml       PASSED
```

stevesloka:
Ok I got Contour cli commands to work. =)

internal/contour/listener_test.go
```diff
@@ -37,7 +37,7 @@ import (
 func TestListenerCacheContents(t *testing.T) {
 	tests := map[string]struct {
 		contents map[string]*v2.Listener
-		want     []proto.Message
+		want     []cachetypes.Resource
```
jpeach, line 40: If we make a `xds` package that just contained type aliases this would read much more clearly as `xds.Resource`.

cmd/contour/serve.go
```diff
@@ -275,6 +278,28 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		FieldLogger: log.WithField("context", "endpointstranslator"),
 	}
 
+	// snapshotCache is used to store the state of what all xDS services should
+	// contain at any given point in time.
+	snapshotCache := cache.NewSnapshotCache(false, cache.IDHash{},
+		log.WithField("context", "XDSServer"))
```
jpeach, line 284: I think it's better to use a specialized node ID hasher to avoid taking a dependency on the Envoy configuration. For example:
```Go

package xds

type ConstantHash string

func (c ConstantHash) ID(*envoy_config_core_v3.Node) string {
         return string(c)
}

func (c ConstantHash) String() string {
         return string(c)
}

var _ cache.NodeHash = ConstantHash("")
var DefaultHash = ConstantHash("contour") 

```

Usage looks like:
```Go
s := cache.NewSnapshotCache(false, xds.DefaultHash, log)
...
s.SetSnapshot(xds.DefaultHash.String(), snapshot)
```

internal/contour/snapshot.go
```diff
@@ -0,0 +1,55 @@
...
+//
+// It does not take into account a specific cache changing.
+// When called, all xDS resources are updated with a new version.
+// Envoy sees this as a noop, but could be improved in future refactorings.
+func (s *SnapshotHandler) UpdateSnapshot() {
+
+	// Increment the snapshot version
+	s.snapshotVersion++
+
+	// Create an xds snapshot with all xDS resources.
```
jpeach, line 38: Nit: s/xds/xDS/

cmd/contour/serve.go
```diff
@@ -275,6 +278,28 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+		CacheHandler:        ch,
+		EndpointsTranslator: et,
+		SnapshotCache:       snapshotCache,
+	}
+
+	// Pass the snapshotHandler off to the CacheHandler & EndpointsTranslator
+	// so they are able to inform the snapshotHandler that the xDS resources
+	// have changed and should be updated in Envoy.
+	ch.SnapshotHandler = snapshotHandler
+	et.SnapshotHandler = snapshotHandler
```
jpeach, line 298: It would be good to try to avoid the circular reference between the cache handler and the snapshots.

Probably what I would try to explore is whether we can take the snapshot in `(*EventHandler) updateDAG()`, and making `(*CacheHandler) OnChange` and free function that creates and returns a new snapshot.

internal/contour/snapshot.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/envoyproxy/go-control-plane/pkg/cache/types"
+
+	"github.com/envoyproxy/go-control-plane/pkg/cache/v2"
+)
+
+// SnapshotHandler implements xds snapshot cache
+type SnapshotHandler struct {
```
jpeach, line 14: Maybe a more evocative name would be `Snapshotter`, since it's not clear what this is a handler for or of. Probably `CacheHandler` is also not a great name any more for the same reason :)

cmd/contour/serve.go
```diff
@@ -275,6 +278,28 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+
+	// Trigger an initial snapshot for static resources
```
jpeach, line 300: Can you explain this in more detail? What static resources are getting snapshotted? Don't we want to wait until the informer caches sync before sending anything to an Envoy?

cmd/contour/serve.go
jpeach, line 284: Nit: we use `xDS` everywhere else in logging.

internal/contour/endpointstranslator.go
```diff
@@ -182,12 +167,12 @@ func (e *EndpointsTranslator) recomputeClusterLoadAssignment(oldep, newep *v1.En
 		}
 	}
 
+	e.SnapshotHandler.UpdateSnapshot()
```
jpeach, line 170: Does this change the frequency at which we update Envoy? IIUC every endpoint change will do a full snapshot; is that what we want here?

jpeach:
> I'm going to update the description to explain the flow a bit. 

Yes, this would be great. Good to include in the commit message too.

internal/contour/snapshot.go
```diff
@@ -0,0 +1,55 @@
+package contour
+
+import (
+	"fmt"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/envoyproxy/go-control-plane/pkg/cache/types"
+
+	"github.com/envoyproxy/go-control-plane/pkg/cache/v2"
```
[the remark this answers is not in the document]
stevesloka: Not intentional, PR is still in draft, so still cleaning up these bits.

cmd/contour/serve.go
stevesloka: There are static resources in listeners that need to get set for probes and other bits to come up. If you don't do this, then you'd have to wait for some resource to show up and trigger a dag rebuild to get these pushed into Envoy.

internal/contour/endpointstranslator.go
stevesloka: That's what happens today, any endpoint change in Kubernetes updates EDS with that new endpoint. I have another issue that looks to reduce this churn since every endpoint ends up in EDS today which is overkill for the amount of endpoints that actually need to be in EDS. 

The alternative is to put updates to EDS in the same loop as the dag rebuild, but I don't think that's a good idea as you'll be delaying updates to EDS which is why it's been kept separate, we want changes to update asap in that cache.

cmd/contour/serve.go
stevesloka: Neat I see how you'd use this now. I wasn't ever setting the hash in the `cache.NewSnapshotCache(...)`. I've got this wired in and looks like it's working!

cmd/contour/serve.go
stevesloka: Yup I've been working on that from @skriss's suggestion earlier. Will have a new push soon.

stevesloka: ok I've done some refactoring on this.

internal/contour/cachehandler.go
stevesloka: Thanks for the suggestion @skriss, I've done some refactoring on this.

internal/contour/cluster.go
```diff
@@ -15,66 +15,28 @@ package contour
...
-	values map[string]*v2.Cluster
-	Cond
-}
-
-// Update replaces the contents of the cache with the supplied map.
-func (c *ClusterCache) Update(v map[string]*v2.Cluster) {
-	c.mu.Lock()
-	defer c.mu.Unlock()
+// ClusterTranslator manages the contents of the gRPC CDS translator.
+type ClusterTranslator struct{}
```
skriss, line 28: any particular reason to keep this as a struct? Seems like it could just be a package function since there's no state in the struct anymore.

internal/contour/cluster.go
```diff
@@ -15,66 +15,28 @@ package contour
...
-	c.values = v
-	c.Cond.Notify()
-}
-
-// Contents returns a copy of the cache's contents.
-func (c *ClusterCache) Contents() []proto.Message {
-	c.mu.Lock()
-	defer c.mu.Unlock()
+// Contents returns an array of CDS resources.
+func (c *ClusterTranslator) Contents(clusters map[string]*v2.Cluster) []xds.Resource {
```
skriss, line 31: `Contents` probably isn't the best name anymore

internal/contour/snapshot.go
```diff
@@ -0,0 +1,130 @@
...
+var _ cache.NodeHash = ConstantHash("")
+var DefaultHash = ConstantHash("contour")
+
+// UpdateSnapshot is called when any cache changes and
+// Envoy should be updated with a new configuration.
+//
+// It does not take into account a specific cache changing.
+// When called, all xDS resources are updated with a new version.
+// Envoy sees this as a noop, but could be improved in future refactorings.
+func (s *SnapshotHandler) UpdateSnapshot(xDSResources map[xds.ResponseType][]xds.Resource) {
```
skriss, line 73: I believe that since both the CacheHandler and the EndpointsTranslator could call this concurrently, we need some synchronization around it.

internal/contour/cachehandler.go
```diff
@@ -19,19 +19,25 @@ package contour
...
 
+	xds "github.com/envoyproxy/go-control-plane/pkg/cache/types"
 	"github.com/projectcontour/contour/internal/dag"
 	"github.com/projectcontour/contour/internal/metrics"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
 )
 
-// CacheHandler manages the state of xDS caches.
+// CacheHandler manages the state of xDS translators.
```
skriss, line 29: I wonder if it would make sense to combine this struct with the SnapshotHandler. It's not doing a whole lot anymore, seems like it may be an unnecessary layer of indirection. The EventHandler could directly call into the SnapshotHandler if we got rid of this one.

internal/contour/cachehandler.go
```diff
@@ -42,30 +48,28 @@ func (ch *CacheHandler) OnChange(dag *dag.DAG) {
...
+	})
 
 	ch.SetDAGLastRebuilt(time.Now())
 }
 
-func (ch *CacheHandler) updateSecrets(root dag.Visitable) {
-	secrets := visitSecrets(root)
-	ch.SecretCache.Update(secrets)
+func (ch *CacheHandler) updateSecrets(root dag.Visitable) []xds.Resource {
+	return ch.SecretTranslator.Contents(visitSecrets(root))
```
skriss, line 62: `visitSecrets` (and other similar functions) currently returns a map, but that seems unnecessary now. If we changed them to just return a slice, then we could also drop the code in the translators that extract a slice of map values.

internal/contour/cachehandler.go
stevesloka: Yup that's a good point. I was taking it at baby steps, but like you said, it doesn't need to stay that way. Let me refactor a bit more.

internal/contour/endpointstranslator.go
jpeach: > That's what happens today, any endpoint change in Kubernetes updates EDS with that new endpoint. I have another issue that looks to reduce this churn since every endpoint ends up in EDS today which is overkill for the amount of endpoints that actually need to be in EDS.

Yeh, but with this change, a new endpoint increments the version of every other resource too, which IIUC means they would get sent to Envoy.

stevesloka: I can give endpoints their own version separate from the other services. Is that what you are asking for? 

The way it is now if Envoy receives the same version, it ignores the change.

jpeach: I'm asking about the implications and effects of this change. IIUC, we handle endpoints specially due to the rate of change. If we stop doing what, what happens?

internal/contour/snapshot.go
```diff
@@ -0,0 +1,135 @@
...
+
+func (c ConstantHash) ID(*envoy_api_v2_core.Node) string {
+	return string(c)
+}
+
+func (c ConstantHash) String() string {
+	return string(c)
+}
+
+var _ cache.NodeHash = ConstantHash("")
```
skriss, line 67: what's this (L67) for?

internal/contour/cluster.go
```diff
@@ -15,77 +15,37 @@ package contour
...
-func (c *ClusterCache) Contents() []proto.Message {
-	c.mu.Lock()
-	defer c.mu.Unlock()
+// translateClusters returns an array of CDS resources.
+func translateClusters(clusters map[string]*v2.Cluster) []xds.Resource {
 	var values []*v2.Cluster
-	for _, v := range c.values {
+	for _, v := range clusters {
 		values = append(values, v)
 	}
```
skriss, line 32: We could avoid having to do this iteration to convert from a map to a slice, if we change `*clusterVisitor.visit(...)` to produce a slice instead of a map. There's not actually any reason to produce the map anymore, since we never need to do a lookup by key on the end result (previously we did though). It'd be slightly more efficient if we switched to just a slice, but probably doesn't matter too much in practice. Something to keep in mind.

The implementation of `*clusterVisitor.visit(...)` _would_ still need to keep a local map of cluster names it's seen already, to avoid putting dupes in the slice it's producing, but that could be just a local var separate from the slice being produced.

cmd/contour/serve.go
jpeach: We ought to keep the previous semantics of not pushing and xDS resources until after the informer caches sync (preferably IMHO until after the first DAG rebuild), so that we don't remove working config and then replace it

stevesloka: Yeah that's a good point. We really need to wait until the first dag build though like you said, the caches being sync'd doesn't mean that the dag is built and ready. 

I can look to add that in here or a followup PR.

internal/contour/endpointstranslator.go
stevesloka: Are you suggesting that we handle endpoints differently now with go-control-plane? Previously we updated per endpoint and this is what the PR does now. 

The alternative is to tie endpoint updates into the dag rebuild time (or some other backoff) which would delay updates to Envoy and not what we want since those endpoints can change and if we don't update Envoy, users will get errors.

internal/contour/cluster.go
```diff
@@ -15,77 +15,37 @@ package contour
...
-	c.values = v
-	c.Cond.Notify()
-}
-
-// Contents returns a copy of the cache's contents.
-func (c *ClusterCache) Contents() []proto.Message {
-	c.mu.Lock()
-	defer c.mu.Unlock()
+// translateClusters returns an array of CDS resources.
+func translateClusters(clusters map[string]*v2.Cluster) []xds.Resource {
```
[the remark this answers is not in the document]
stevesloka: nope that's possible. I left it as a helper  in the tests, but can move this off to another place in the tests.

internal/contour/snapshot.go
stevesloka: This is a suggestion from @jpeach which allows the service-node name to be anything.

jpeach: Assigning to `_` does a compile-time check that `ConstantHash` implements the given interface. It's useful as documentation and type enforcement.

stevesloka:
This has gotten super stale with a bunch of the other refactoring going on, going to come back after we get all that other work done and re-implement.

PR #2606, opened 2020-06-18
hack/release: fix make-release-tag.sh script

jpeach:
Fix the make-release-tag.sh script to support both BSD and GNU variants
of sed. Support running it from a release branch, where the image version
substitution we need to make is different. Update the test for detecting
the git remote, which was backwards.

### DIFF
--- a/hack/release/make-release-tag.sh
+++ b/hack/release/make-release-tag.sh
@@ -10,7 +10,7 @@ readonly OLDVERS="$1"
 readonly NEWVERS="$2"
 
 if [ -z "$OLDVERS" ] || [ -z "$NEWVERS" ]; then
-    printf "Usage: %s OLDVERS NEWVERS\n" $PROGNAME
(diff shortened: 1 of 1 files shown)


hack/release/make-release-tag.sh
```diff
@@ -20,26 +20,33 @@ set -o pipefail
...
-        sed -i "-es|docker.io/projectcontour/contour:master|$IMG|" "$f"
-        ;;
-    *)
-        printf "Unsupported system '%s'" "$(uname -s)"
-        exit 2
-        ;;
-    esac
+for example in examples/contour/03-envoy.yaml examples/contour/03-contour.yaml ; do
+    # The version might be master or OLDVERS depending on whether we are
+    # tagging from the release brancah or from master.
```
stevesloka, line 45: nit: s/brancah/branch/g

Issue #2602, opened 2020-06-18
Envoy pod stuck due to incomplete certificate


shyaamsn:
**What steps did you take and what happened:**
Deployed contour v1.5.0 and envoy v1.14.1 on a kind cluster with cert-manager providing certificates for contour and envoy. Envoy container never becomes ready. The logs point to certificate error: Filesystem config update rejected: Failed to load incomplete certificate from /certs/tls.crt, /certs/tls.key. 

I believe cert-manager takes sometime to update the secret with all the certificates. Envoy starts and fails to load certificates. 

**What did you expect to happen:**
Envoy container should be restarted again so that it can load the certificates again

**Anything else you would like to add:**
This happens about 50% of the times in my kind cluster.

The contour example yamls dont have any liveness probe configured for envoy container. Can adding a HTTP liveness probe to /ready on port 8002 help with restarting the envoy container in such cases?

Any help on how to resolve the issue is appreciated.

**Environment:**

- Contour version: v1.5.0
- Kubernetes version: (use `kubectl version`): 1.17.2
- Kubernetes installer & version: kind version 0.8.1
- Cloud provider or hardware configuration:
- OS (e.g. from `/etc/os-release`):

youngnick:
Adding a liveness probe is not straightforward, because Envoy will not open any ports until it reaches Contour for its configuration. Because of this, if we have a naive liveness port, Envoy will not be alive for a little while, so it will be constantly restarted.

To change this, someone will need to investigate options. Things I can think off from the top of my head:
- Adding a (long-ish) timeout or a lot of retries to a liveness probe. This would mean that, if Envoy doesn't start, you have to wait that period for the liveness probe to fail.
- Adding a static configuration for some port where its only purpose is to indicate that Envoy is running, not that it's configured ( which `/ready` on 8002 does already).
- Additionally, in Kubernetes 1.16, startup probes were added, which are designed for this purpose.

To be clear, I think this is definitely a problem we should look at - particularly as people start to use the newer ability of Contour and Envoy to use a rotated certificate for securing gRPC as you're doing here. I can't look into it right now myself though, so I'll just mark this as accepted for now.

jpeach:
FWIW `contour bootstrap` already checks the certificate files are present, but not the contents.

PR #2607, opened 2020-06-18
Check for empty envoy TLS cert file during bootstrap

shyaamsn:
Currently during envoy bootstrap, TLS cert file is checked for existence. Adding a check for empty file too since cert-manager generates secret with empty certificates first and then updates it.

Short term fix for: https://github.com/projectcontour/contour/issues/2602 . Long term fix is to come up with a liveness probe for envoy which can restart envoy if it fails to start.

Signed-off-by: Shyaam Nagarajan <nagarajans@vmware.com>

### DIFF
--- a/internal/envoy/bootstrap.go
+++ b/internal/envoy/bootstrap.go
@@ -100,8 +100,12 @@ func bootstrap(c *BootstrapConfig) ([]bootstrapf, error) {
 			// but there is no way to detect and fix that. If
 			// we check and fail here, that is visible in the
 			// Pod lifecycle and therefore fixable.
-			if _, err := os.Stat(f); err != nil {
+			fi, err := os.Stat(f)
+			if err != nil {
 				return nil, err
+			}
+			if fi.Size() == 0 {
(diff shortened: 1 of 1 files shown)


internal/envoy/bootstrap.go
```diff
@@ -100,9 +100,13 @@ func bootstrap(c *BootstrapConfig) ([]bootstrapf, error) {
 			// but there is no way to detect and fix that. If
 			// we check and fail here, that is visible in the
 			// Pod lifecycle and therefore fixable.
-			if _, err := os.Stat(f); err != nil {
+			fi, err := os.Stat(f)
+			if err != nil {
 				return nil, err
 			}
+			if fi.Size() == 0 {
+				return nil, fmt.Errorf("TLS file %s is empty", f)
```
jpeach, line 108:
```suggestion
				return nil, fmt.Errorf("%q is empty", f)
```

shyaamsn: Thanks. Updated

PR #2617, opened 2020-06-19
internal/dag: add more header condition validations

skriss:
Follow-up from https://github.com/projectcontour/contour/pull/2600#discussion_r442916936 (specifically item 3).

Detect additional invalid header condition scenarios and write them
to the proxy's status, specifically for:
- routes with both Exact and NotExact conditions with the same
value for the same header
- routes with both Contains and NotContains conditions with the
same value for the same header


skriss:
fixed typo

PR #2620, opened 2020-06-22
docs: correct errors in timeout policy section

### LINKED ISSUE #2544: HTTPProxy timeoutPolicy.response of "0s" (as mentioned in docs) is NOT infinite
With contour 1.4.0, I've just attempted to set a timeoutPolicy.response, and "0s" doesn't seem to work.

I expected the value "0s" to mean an infinite timeout, as the docs describe:
> `timeoutPolicy.response` This field can be any positive time period or “infinity”. The time period of 0s will also be treated as infinity. This timeout covers the time from the end of the client request to the end of the upstream response. By default, Envoy has a 15 second value for this timeout. More information can be found in Envoy’s documentation.
https://projectcontour.io/docs/v1.4.0/httpproxy/

First with `"infinity"` (as worked in the old Ingress [annotation](https://projectcontour.io/docs/v1.4.0/annotations/)):
```
      timeoutPolicy:
        response: "infinity"
        idle: "355s"
```

That gives a timeout of zero in Envoy's `/config_dump`:
```
          "route": {
           "cluster": "default/myapp-service/5000/da33a3ff5e",
           "timeout": "0s",
           "idle_timeout": "355s"
          }
```

But then with `"0s"` (as described in the `timeoutPolicy` documentation):
```
      timeoutPolicy:
        response: "0s"
        idle: "355s"
```

That gives a default setting (no timeout field) in Envoy's `/config_dump`:
```
          "route": {
           "cluster": "default/myapp-service/5000/da33a3ff5e",
           "idle_timeout": "355s"
          }
```

So I have a workaround (setting timeoutPolicy.response to "infinity"), but either there's a bug or the docs are out of date.  😺

skriss:

closes #2544

### DIFF
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -297,17 +297,19 @@ type TCPHealthCheckPolicy struct {
 // TimeoutPolicy durations are expressed in the Go [Duration format](https://godoc.org/time#ParseDuration).
 // Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
 // The string "infinity" is also a valid input and specifies no timeout.
+// A value of "0s" will be treated as if the field were not set, i.e. by using Envoy's default behavior.
 //
 // Example input values: "300ms", "5s", "1m".
 type TimeoutPolicy struct {
 	// Timeout for receiving a response from the server after processing a request from client.
-	// If not supplied, the timeout duration is undefined.
+	// If not supplied, Envoy's default value of 15s applies.
 	// +optional
 	Response string `json:"response,omitempty"`
 
 	// Timeout after which, if there are no active requests for this route, the connection between
 	// Envoy and the backend or Envoy and the external client will be closed.
-	// If not specified, there is no per-route idle timeout.
+	// If not specified, there is no per-route idle timeout, though a connection manager-wide
+	// stream_idle_timeout default of 5m still applies.
 	// +optional
 	Idle string `json:"idle,omitempty"`
 }
--- a/site/docs/master/httpproxy.md
+++ b/site/docs/master/httpproxy.md
@@ -673,20 +673,20 @@ In this example, requests to `timeout.bar.com/` will have a response timeout pol
 This refers to the time that spans between the point at which complete client request has been processed by the proxy, and when the response from the server has been completely processed.
 
 - `timeoutPolicy.response` This field can be any positive time period or "infinity".
-The time period of **0s** will also be treated as infinity.
 This timeout covers the time from the *end of the client request* to the *end of the upstream response*.
 By default, Envoy has a 15 second value for this timeout.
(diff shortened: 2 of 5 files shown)


site/docs/master/httpproxy.md
```diff
@@ -673,14 +673,15 @@ In this example, requests to `timeout.bar.com/` will have a response timeout pol
 This refers to the time that spans between the point at which complete client request has been processed by the proxy, and when the response from the server has been completely processed.
 
 - `timeoutPolicy.response` This field can be any positive time period or "infinity".
-The time period of **0s** will also be treated as infinity.
 This timeout covers the time from the *end of the client request* to the *end of the upstream response*.
 By default, Envoy has a 15 second value for this timeout.
 More information can be found in [Envoy's documentation][4].
+Note that a value of **0s** will be treated as if the field were not set, i.e. by using Envoy's default behavior.
```
jpeach, line 679: Can you please also update the API documentation in `apis/projectcontour/v1/httpproxy.go`, which is the authoritative reference for the HTTPProxy spec?

site/docs/master/httpproxy.md
```diff
@@ -1416,3 +1417,4 @@ Some examples of invalid configurations that Contour provides statuses for:
  [9]: {% link docs/master/annotations.md %}
  [10]: /docs/{{site.latest}}/api/#projectcontour.io/v1.Service
  [11]: configuration.md#fallback-certificate
+ [12]: https://www.envoyproxy.io/docs/envoy/v1.11.2/api-v2/api/v2/route/route.proto.html#envoy-api-field-route-routeaction-idle-timeout
```
jpeach, line 1420: It probably makes sense to use a versioned link here, but we should link to the latest version `1.14.2`.

site/docs/master/httpproxy.md
skriss: added the same note to the API docs

site/docs/master/httpproxy.md
skriss: updated all the links on the page; I also realized there was already a link for the idle timeout (`[6]`) so de-duped.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -297,6 +297,7 @@ type TCPHealthCheckPolicy struct {
 // TimeoutPolicy durations are expressed in the Go [Duration format](https://godoc.org/time#ParseDuration).
 // Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
 // The string "infinity" is also a valid input and specifies no timeout.
+// A value of "0s" will be treated as if the field were not set, i.e. by using Envoy's default behavior.
```
jpeach, line 300: There's 2 comments on the struct fields about the timeout ... are they still accurate?

skriss: I added some more clarity there around default behavior.

Issue #2619, opened 2020-06-22
Add support to disable HTTP2.


primeroz:
**Please describe the problem you have**

HTTP2 Coalescing , when using a wildcard, can lead to problems in SNI based routing.\
This was addressed in https://github.com/projectcontour/contour/issues/1493 by responding with the **http 421** code pushing on the client the responsability to re-establish a connection using the correct tls name for sni routing.

_unfortunately_ Safari ( or apparently webkit ... or actually the CFNetwork  ) has a [bug](https://bugs.webkit.org/show_bug.cgi?id=192926) and so it just hangs when it receive a 421 back.\
This have been known for years but no fix exists for it yet.

As discussed with @jpeach the only solution , for people in my situation where we can't stop using wildcard certificates, is to to disable HTTP2 and revert to the always great HTTP 1.1. 

This is currently not possible in Contour though

maplain:
setting `default-http-versions` to only `HTTP/1.1` solves the problem in my case: https://projectcontour.io/docs/main/configuration/#configuration-file

youngnick:
Following up on issues, it seems like @maplain's fix may cover your use case, @primeroz. Is there anything else you would like done on this issue?

primeroz:
It sounds good to me, feel free to close this issue and thanks!

PR #2622, opened 2020-06-23
cmd/contour: support defaulting HTTP protocol versions

jpeach:
Add Contour config file for defaulting the HTTP protocol version
accepted by the proxy listener. This principally applies to HTTPS
listeners, where HTTP/1.1 and HTTP/2 can be accepted on the same
listener.

If no configuration is specified, all HTTP protocol versions are
automatically detected and supported. Otherwise, HTTP/1.1 and HTTP/2
can be individually toggled. There is no support for per-resource
configuration in this change, but it can easily be supported if
necessary.

The config file syntax looks like this:

```
default-http-versions:
- http/1.1
- http/2
```

This updates #2619.


jpeach:
Marked as draft. @youngnick LMK what you think of this approach.

jpeach:
@youngnick added docs to configuration file page and fax

cmd/contour/servecontext.go
```diff
@@ -131,6 +132,14 @@ type serveContext struct {
...
 	// Name of the envoy service to inspect for Ingress status details.
 	EnvoyServiceName string `yaml:"envoy-service-name,omitempty"`
+
+	// DefaultHTTPVersions defines the default set of HTTPS
+	// versions the proxy should accept. HTTP versions are
+	// strings of the form "HTTP/xx". Supported versions are
+	// "HTTP/1.1" and "HTTP/2".
+	//
+	// If this field not specified, all supported versions are accepted.
+	DefaultHTTPVersions []string `yaml:"default-http-versions"`
```
stevesloka, line 142: "Default" seems like the wrong term here, it's more like "Enabled" since this is overriding the "supported" versions.

cmd/contour/servecontext.go
```diff
@@ -131,6 +132,14 @@ type serveContext struct {
 
 	// Name of the envoy service to inspect for Ingress status details.
 	EnvoyServiceName string `yaml:"envoy-service-name,omitempty"`
+
+	// DefaultHTTPVersions defines the default set of HTTPS
+	// versions the proxy should accept. HTTP versions are
+	// strings of the form "HTTP/xx". Supported versions are
+	// "HTTP/1.1" and "HTTP/2".
```
stevesloka, line 139: Do we need to include `HTTP` in the value? The parameter name specifies it can only be HTTP.

site/_resources/faq.md
```diff
@@ -27,4 +27,21 @@ The HTTPProxy custom resource is an attempt to solve these issues with an API th
 More importantly, the HTTPProxy CRD is designed with inclusion in mind, a feature that enables administrators to configure top-level ingress settings (for example, which virtual hosts are available to each team), while delegating the lower-level configuration (for example, the mapping between paths and backend services) to each development team.
 More information about the HTTPProxy API can be found [in the HTTPProxy documentation][1].
 
+# Q: When I load my site in Safari, it shows me an empty page. Developer tools show that the HTTP response was 421. Why does this happen?
```
stevesloka, line 30: Should we add the version tested of Safari? Presuming eventually in the future the issue gets resolved?

internal/contour/listener.go
```diff
@@ -198,8 +203,8 @@ func (lvc *ListenerVisitorConfig) requestTimeout() time.Duration {
 // minProtocolVersion returns the requested minimum TLS protocol
 // version or envoy_api_v2_auth.TlsParameters_TLSv1_1 if not configured {
 func (lvc *ListenerVisitorConfig) minProtoVersion() envoy_api_v2_auth.TlsParameters_TlsProtocol {
```
stevesloka, line 205: nit: You renamed `MinimumProtocolVersion` --> `MinimumTLSVersion` but the method is still called `minProtoVersion()`. Change to `minTLSVersion()` to match?

internal/contour/listener.go
```diff
@@ -75,8 +75,13 @@ type ListenerVisitorConfig struct {
 	// If not set, defaults to false.
 	UseProxyProto bool
 
-	// MinimumProtocolVersion defines the min tls protocol version to be used
-	MinimumProtocolVersion envoy_api_v2_auth.TlsParameters_TlsProtocol
+	// MinimumTLSVersion defines the minimum TLS protocol version the proxy should accept.
+	MinimumTLSVersion envoy_api_v2_auth.TlsParameters_TlsProtocol
+
+	// DefaultHTTPVersions defines the default set of HTTPS
```
stevesloka, line 81: This field applies to both TLS and non-TLS listeners, right? The comment is misleading that it only applied to TLS ones.

site/_resources/faq.md
jpeach: I can't be sure about the versions, so I'd prefer to update this if we ever find out there is a version that fixes the problem.

cmd/contour/servecontext.go
jpeach: The intention is to set the default HTTP versions for all the proxy objects. Later we can add per-proxy versions to the `VirtualHost` spec to override these defaults. This also gives us the space to add something like "enabled versions", which would restrict what can become the defaults.

cmd/contour/servecontext.go
jpeach: It's not strictly necessary, but IMHO it reads better and makes the config more obvious :)

internal/contour/listener.go
jpeach: Yeh it's applied to both, but since we don't support h2c, it's really a moot point. I'll attempt to clarify this comment.

internal/contour/listener.go
jpeach: Fixed.

internal/contour/listener.go
jpeach: Updated the comment.

PR #2626, opened 2020-06-25
Design Proposal: Control the conditions of a `RetryPolicy`

KevinSnyderCodes:


design/retryon-design.md
```diff
@@ -0,0 +1,147 @@
...
+        5xx: false
+      	connectFailure: true
+      	gatewayError: false
+      	...
+```
+
+Arguments for this approach:
+
+- Contour can control the options that it will allow users to populate Envoy's `retry_on` field with
+- It ensures that only valid values are allowed in Envoy's `retry_on` field
```
kevincantu, line 132: What happens when an invalid value is given to Envoy?  Does it do something silently which we should indeed try to avoid?

(If so, we might still prefer to check the validity in a small addition to the design, above, but it seems worth knowing and I'm relatively new to Envoy so I don't know.)

kevincantu:
My two cents: this all looks good!

design/retryon-design.md
jpeach: In general, if you give an invalid value to Envoy, it will NACK the update, and you will run into #1176.

youngnick: I think that the validity of the retryOn strings should absolutely be checked against a known list, not just passed through to Envoy. It can be difficult to determine if the config has been applied correctly to Envoy, it's much better to be confident Contour has passed a correct config.

stevesloka: +1 to what @youngnick said, Contour should always strive to produce clean config before passing off to envoy.

KevinSnyderCodes: Agree about the validation before sending to Envoy. I'll update the design proposal to solve this with `kubebuilder` validation markers. I still think it makes sense to use a `[]string` for `RetryOn` as using a unique boolean field for each possible value seems too verbose.

stevesloka:
I think @youngnick has some good questions around sane "defaults" that an operator could set. Meaning, should we allow this to also be configured at a vhost level or even at the inclusion level where any proxy that gets included from that point down the chain inherits this configuration.

We do that in places today in the Contour configuration file level, but it's worth a discussion if that would be a logical place to set or at the vhost level to set the default (if even required).

KevinSnyderCodes:
To me it's out of scope for this particular design proposal, but certainly there's a good discussion to be had there. I know I'll be enabling retries on all my `HTTPProxy` manifests on `connect-failure` to mitigate 503s during rollouts, so asking "should there be a top-level default across all proxies" seems reasonable.

But if we'd like to expand scope and include it here, I'm open to it.

stevesloka:
It might be nice @KevinSnyderCodes to discuss what could be done, this design doesn't have to match 1:1 for a single feature. We can design a larger bit, then implement smaller portions as needed. 

I'll leave to @youngnick on if he wants to include that in this design.

KevinSnyderCodes:
@youngnick I've added considerations around global retry policy and retry policies at scale (the tens or hundreds of retry policies you mentioned)

youngnick:
Thanks for adding that @KevinSnyderCodes, that's just what I was looking for. The last thing I'd like in the proposal is something to cover what we will do if people want a Contour-wide default later (that is, add something to the config file to configure the default). Implementing that is absolutely not part of the first phase of this. Once there's something in there about this, I'm all good.

KevinSnyderCodes:
Thank you @youngnick 🙂 

What are the next steps? Can I get started on implementation or do we need to merge this first, or something else?

youngnick:
Merged, feel free to get started @KevinSnyderCodes!

Issue #2225, opened 2020-02-13
Exposing more Envoy configuration knobs


pims:
**Please describe the problem you have**
[A clear, concise, description of the problem you are facing. What is the _problem_ that feature X would solve for you?]

When deploying Contour/Envoy in a large cluster with heterogenous workloads (gRPC, websockets, h2, etc.), some Envoy settings become quite important to get right.

Here's a subset of settings we have configured through a custom patch:

```golang
{
  DrainTimeout:      safetime(customCtx.DrainTimeout),
  IdleTimeout:       safetime(customCtx.IdleTimeout), 
  RequestTimeout:    safetime(customCtx.RequestTimeout),
  StreamIdleTimeout: safetime(customCtx.StreamIdleTimeout),

  PerConnectionBufferLimitBytes: customCtx.PerConnectionBufferLimitBytes,
  MaxConcurrentStreams:          customCtx.MaxConcurrentStreams,
  ...
}
```

Most of these are part of the `HTTPConnectionManager` filter configuration.
We have been using these settings with great success for several months in production now, and would be interested in contributing these changes upstream.

My questions to the Contour team: **would this fall into the scope of what you envision for Contour?**

If you believe that this is out of scope for this project, would you be open to minor code changes that would make it easier for people to patch.

A concrete example being the addition of `requestTimeout` in commit https://github.com/projectcontour/contour/commit/d19748209ae047c1f21134ed49a401822090963f

If this could be wrapped in some kind of `type HTTPConnectionManagerConfiguration struct{}`, adding unsupported fields would no longer require changing the function signature, which is used extensively through the comprehensive test suite.

Examples of features we have added over the years:
* tracing
* timeouts
* limits for annotations (maxRetries, maxTimeout, maxPendingRequests…)
* ext-authz

Worth noting, we were able to retain all of Contour's simplicity through custom defaults while giving us the ability to configure Envoy to our needs.

wdyt?

jpeach:
xref #1902 
xref #1375

stevesloka:
https://github.com/projectcontour/contour/issues/1192

davecheney:
Thank you for raising this issue. I want to give a bit of background.

The general policy of this project has been to avoid knobs and tuneable for their own sake where possible. We've done this in the past by adjusting our defaults, and only in the cases where there are no generally applicable defaults exposed those tuneable to contour users. 

[...]

PR #2632, opened 2020-06-26
add configurable idle timeout

skriss:
updates #2225 

I started with just one timeout setting chosen at random; I'll add the others if the approach looks OK.

I can see how adding additional tuneables one-by-one is not ideal (+268 LOC to add a single config parameter). I'm still paging in the context to see if there are other ideas/options for how to do this in a more general way.

### DIFF
--- a/cmd/contour/servecontext.go
+++ b/cmd/contour/servecontext.go
@@ -115,6 +115,10 @@ type serveContext struct {
 	// LeaderElectionConfig can be set in the config file.
 	LeaderElectionConfig `yaml:"leaderelection,omitempty"`
 
+	// TimeoutConfig holds various configurable timeouts that can
+	// be set in the config file.
+	TimeoutConfig `yaml:"timeouts,omitempty"`
+
 	// RequestTimeout sets the client request timeout globally for Contour.
 	RequestTimeout time.Duration `yaml:"request-timeout,omitempty"`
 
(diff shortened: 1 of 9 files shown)


cmd/contour/servecontext.go
```diff
@@ -227,6 +236,14 @@ type LeaderElectionConfig struct {
 	Name          string        `yaml:"configmap-name,omitempty"`
 }
 
+// TimeoutConfig holds various configurable proxy timeout values.
+type TimeoutConfig struct {
+	// IdleTimeout defines how long the proxy should wait while there
+	// are no active requests before terminating an HTTP connection. Set
+	// to 0 to disable the timeout.
+	IdleTimeout time.Duration `yaml:"idle-timeout,omitempty"`
```
skriss, line 244: @youngnick @jpeach @stevesloka before merging this, I'm considering renaming this field and related code to `ConnectionIdleTimeout` to more explicitly differentiate from `StreamIdleTimeout` which I'm adding next. What do you all prefer? `IdleTimeout` and `StreamIdleTimeout`, or `ConnectionIdleTimeout` and `StreamIdleTimeout`?

jpeach: Envoy has separate stream and connection timeouts, so we should name this for whichever Envoy timeout it refers to. [This](https://www.envoyproxy.io/docs/envoy/latest/faq/configuration/timeouts) helps a lot.

skriss: This one is the connection idle timeout, but envoy just calls it "idle_timeout".

skriss: I'll leave this as-is so I can move ahead, but can always circle back if desired.

stevesloka: I think `ConnectionIdleTimeout` & `StreamIdleTimeout` like you suggested is good. It helps to better clarify the context a bit.

skriss: OK @stevesloka I updated the name, mind taking a quick look before I merge?

jpeach: > This one is the connection idle timeout, but envoy just calls it "idle_timeout".

Right. The doc I linked to explains that this is an alias to the stream timeout.

PR #2638, opened 2020-06-29
Enable TCP keepalive on listener sockets.

### LINKED ISSUE #2633: Enable TCP keepalive on listener sockets
Contour should configure TCP keepalives for Envoy listener sockets so that half-open connections with clients (downstream in Envoy speak) are flushed. This helps Envoy work better with overlay networks which drop long running idle TCP connections. TCP keepalives are already enabled for server (upstream) and xDS connections so this would complete the set.

I have a pull request that implements this by adding SocketOptions if you're open to this change.

erwbgy:
Fixes #2633  


internal/envoy/tcp_keepalive.go
```diff
@@ -0,0 +1,63 @@
...
+
+package envoy
+
+import (
+	"syscall"
+
+	envoy_api_v2_core "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
+)
+
+func TcpKeepaliveSocketOptions() []*envoy_api_v2_core.SocketOption {
```
jpeach, line 21: Conventionally, this should be called `TCPKeepaliveSocketOptions()`. 

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

internal/envoy/tcp_keepalive.go
```diff
@@ -0,0 +1,63 @@
...
+
+	// Note: TCP_KEEPIDLE + (TCP_KEEPINTVL * TCP_KEEPCNT) must be greater than
+	// the grpc.KeepaliveParams time + timeout (currently 60 + 20 = 80 seconds)
+	// otherwise TestGRPC/StreamClusters fails
```
jpeach, line 25: Nit: Could you please consistently end code comments with a period? Here, and below.

internal/envoy/tcp_keepalive.go
erwbgy: Fixed.

internal/envoy/tcp_keepalive.go
erwbgy: Fixed.

erwbgy:
@jpeach @youngnick  I saw some discussion on Slack about putting this behind a feature flag because these socket options don't work on Windows. Another way would be to check runtime.GOOS and only set the socket options on Linux, which should be a simple update to this pull request.  Thoughts on that?

erwbgy:
@pims Did you get a chance to test if this works on MacOS?

pims:
@erwbgy I should be able to test this weekend. I’ll update the PR.

I think build flags might also be a good way to gate this without a dedicated contour flag.

pims:
Hmm, so this doesn't build on MacOS:
```
internal/envoy/tcp_keepalive.go:40:17: undefined: syscall.TCP_KEEPIDLE
internal/envoy/tcp_keepalive.go:48:17: undefined: syscall.TCP_KEEPINTVL
internal/envoy/tcp_keepalive.go:58:17: undefined: syscall.TCP_KEEPCNT
```

if we rename `tcp_keepalive.go` to `tcp_keepalive_linux.go` and add the `//+build linux` directive
as well as create a `tcp_keepalive.go` with the following:

```go
package envoy

import (
	envoy_api_v2_core "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
)

func TCPKeepaliveSocketOptions() []*envoy_api_v2_core.SocketOption {
    // should we return nil instead?
    return []*envoy_api_v2_core.SocketOption{}
}
```

it builds correctly. I still need to test it though.

erwbgy:
Thanks @pims, learnt something new today as I hadn't had the need for build constraints before.

jpeach:
> Hmm, so this doesn't build on MacOS:
> 
> ```
> internal/envoy/tcp_keepalive.go:40:17: undefined: syscall.TCP_KEEPIDLE
> internal/envoy/tcp_keepalive.go:48:17: undefined: syscall.TCP_KEEPINTVL
> internal/envoy/tcp_keepalive.go:58:17: undefined: syscall.TCP_KEEPCNT
> ```
> 
> if we rename `tcp_keepalive.go` to `tcp_keepalive_linux.go` and add the `//+build linux` directive
> as well as create a `tcp_keepalive.go` with the following:
> 
> ```go
> package envoy
> 
> import (
> 	envoy_api_v2_core "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
> )
> 
> func TCPKeepaliveSocketOptions() []*envoy_api_v2_core.SocketOption {
>     // should we return nil instead?
>     return []*envoy_api_v2_core.SocketOption{}
> }
> ```
> 
> it builds correctly. I still need to test it though.

Sorry, I should have picked up this issue during review :(

What we should do is carry our own constants for the socket options. If we conditionally compile the whole of the keep alive options file, then Envoy would get programmed differently depending on what platform Contour is running on, which is not the result that we want.

We could use +build to define the socket option number, but since we will need to hard-code them for non-linux platforms, I'm not sure there is much benefit in that. Basically what we need is a set of constants per target Envoy platform.

pims:
🤷  `syscall.TCP_KEEPINTVL` is available on `// +build aix freebsd linux netbsd` which I'd guess would cover 99.9% of use cases of Contour/k8s + Envoy.
My main concern was the ability to build/run contour on MacOS for development purposes.

pims:
```
#define	TCP_KEEPINTVL		0x101	/* interval between keepalives */
#define	TCP_KEEPCNT		0x102	/* number of keepalives before close */
```

from https://opensource.apple.com/source/xnu/xnu-2422.115.4/bsd/netinet/tcp.h

Couldn't find references to `TCP_KEEPIDLE` on MacOS but I could have missed it.

erwbgy:
Looks like we can get these values from https://github.com/felixge/tcpkeepalive/.

erwbgy:
Added constants for the TCP keepalive options, defaulting to the syscall values for Linux and BSD with MacOS (Darwin) being different.  Apparently MacOS calls the TCP_KEEPIDLE socket option TCP_KEEPALIVE.

The GOOS values not covered are aix, android, illumos, js, plan9, solaris, windows. Do we need to be concerned about those?

jpeach:
> Added constants for the TCP keepalive options, defaulting to the syscall values for Linux and BSD with MacOS (Darwin) being different. Apparently MacOS calls the TCP_KEEPIDLE socket option TCP_KEEPALIVE.

Remember that the socket options are being used to program Envoy, so we need to send the right options for the platform that Envoy is running on, not the ones for the platform that Contour is running on. There's no in-band way to detect the [Envoy platform](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/core/base.proto#core-node) AFAIK, so the only option is to assume that it is running on Linux (which would be true in the majority of cases).

erwbgy:
Is it not valid to assume that Contour and Envoy would run on the same platform?

Do we need to support any platforms other than Linux?  For development or runtime or both?

jpeach:
> Is it not valid to assume that Contour and Envoy would run on the same platform?

Not always. The main case is developer usage, where you may run a working copy of Contour on macOS and Envoy in a kind cluster.

> Do we need to support any platforms other than Linux? For development or runtime or both?

For Contour, we should support macOS as a developer convenience. I don't know of any use case for supporting Envoy on non-Linux.

erwbgy:
Thanks @jpeach I think I've got it now.  We only support Envoy on Linux so always configure Linux TCP keep-alive socket options regardless of the platform that Contour is running on.

internal/envoy/tcp_keepalive.go
```diff
@@ -0,0 +1,71 @@
...
+
+// We only support Envoy on Linux so always configure Linux TCP keep-alive
+// socket options regardless of the platform that Contour is running on.
+const (
+	TCP_KEEPIDLE  = 0x4 // Linux syscall.TCP_KEEPIDLE
+	TCP_KEEPINTVL = 0x5 // Linux syscall.TCP_KEEPINTVL
+	TCP_KEEPCNT   = 0x6 // Linux syscall.TCP_KEEPCNT
+)
+
+func TCPKeepaliveSocketOptions() []*envoy_api_v2_core.SocketOption {
```
stevesloka, line 29: @erwbgy we should document this (could be this PR or a new one) to make the defaults here known. Also, we should think about adding these to the config file to allow users to define their own timeouts.

erwbgy: Indeed - added a suggestion about documenting this change to #2633 and would like to do docs and configuration in separate PRs, if that's okay.

erwbgy: Added documentation for Envoy TCP Keepalives.

site/_resources/envoy.md
```diff
@@ -53,3 +53,14 @@ If you are providing your own Envoy it must be compiled with the following exten
...
+## Envoy TCP keepalives
+
+Contour configures TCP keepalives for Envoy listener sockets so that half-open connections with clients (downstream in Envoy speak) are flushed which helps Envoy work better with networks that drop long running idle TCP connections.  Once a TCP connection has been idle for 45 seconds a keepalive packet is sent every 5 seconds and if 9 packets receive no response then the connection will be terminated by Envoy.
+
+TCP keepalives are also enabled for connections from Envoy to servers (upstream in Envoy terminology). Once a connection has been idle for 30 seconds a keepalive packet is sent every 5 seconds and if 3 packets receive no response then the connection will be terminated by Envoy.
+
+Envoy is only supported on Linux so Contour always configures Linux TCP keep-alive socket options regardless of the platform that Contour is running on.
+
+Note that TCP keepalives are handled entirely at the network layer and require no application configuration.  They are not the same as HTTP Keep-Alive or application level heartbeats.
+
```
jpeach, line 66: Unfortunately, this doc is the Envoy support matrix, so it's not the right place to document operational or configuration concerns. Can you please remove this for now?

erwbgy: Ok, done.

erwbgy:
Rebased and squashed the commits with a single commit message.

PR #2642, opened 2020-06-30
Add design for HTTPProxy conditions

youngnick:
Updates #2495.

This is the intial design for this, comments and initial feedback welcome.


design/httpproxy-status-conditions.md
```diff
@@ -0,0 +1,279 @@
...
+    services:
+    - name: service-does-not-exist
+      port: 80
+  virtualhost:
+    fqdn: foo2.bar.com
+    tls:
+      secretName: testsecret-does-not-exist
+status:
+  conditions:
+  - type: Valid
```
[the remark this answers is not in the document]
youngnick: `Valid` with `status: true` can't have errors. I haven't explained this clearly enough, thanks.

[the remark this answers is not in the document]
youngnick: I think I've updated this to make it clearer now.

design/httpproxy-status-conditions.md
```diff
@@ -0,0 +1,279 @@
...
+- To add a place in the CRDs for the Conditions to live.
+
+## Non-goals
+Add support for some method of exposing the inclusion graph in the HTTPProxy’s status. That is a separate issue (#xxx)
+
+## High-level design
+
+We're going to do the following things:
+- add a `conditions` section to HTTPProxy `status`. We're doing this to allow for additional extensibility in the standard way.
+- Add a `Valid` condition as the standard HTTPProxy condition. We've chosen a `Valid` condition here rather than a more traditional `Ready` because `Ready` implies that we're sure that the Envoy config has been accepted - which we don't have the mechanisms to ensure yet. We may add a `Ready` condition at a later date.
```
jpeach, line 31: Suggest that `Validated` is a better term than `Valid`, since if `Valid` condition has errors, then it's not really valid? Or would errors only be present when the status is `false`?

design/httpproxy-status-conditions.md
```diff
@@ -0,0 +1,279 @@
...
+- The `Valid` condition's `Reason` field and the `validity` field will have the same value, as will the `Valid` condition's message field and the `description` field. That is, we will keep the functionality of the `validity` and `description` fields, but also duplicate it in the `Valid` Condition.
```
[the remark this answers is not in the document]
youngnick: There are a few ideas I'm working towards here:
- keep the existing fields with their existing behavior. This is unavoidable, because of v1 API guarantees.
- Add a Condition so that people can use things that expect Conditions (like `kubectl wait --for condition=`), and to have a place to put other conditions in the future.
- Make it so that there are multiple ways for errors to be sourced if there are multiple errors, on the assumption that a HTTPProxy user shouldn't need lots of passes to fix their object if it's broken. This is a current problem with the single reason setup.
- Try to keep the polarity of the top-level conditions consistent. Having a `Validated` condtion with normal-true polarity and `ValidationError` and `ValidationWarning` conditions with abnormal-true polarities seems also not ideal.

The last two things are the reasons why I thought that sub-conditions would be better.

I'll grant that changing the condition schema is a bigger deal than I initially thought though.

design/httpproxy-status-conditions.md
```diff
@@ -0,0 +1,279 @@
...
+The `conditions` stanza also allows for external controllers to add information to the HTTPProxy (in particular, this allows `external-dns` to possibly add a `DNSProvisioned` condition or similar in the future).
+
+Yes, the overloading of the term `conditions` between `status` and `spec` is unfortunate, but in order to be able to use standard tooling, we are stuck with the YAML key name for `status`.
+
+## Detailed design
+
+We will add the following generic type as a stand-in until the `metav1.Condition` type is available upstream.
+
+```go
+type UpstreamCondition struct {
```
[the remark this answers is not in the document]
youngnick: Agreed, I hadn't thought of that.

design/httpproxy-status-conditions.md
youngnick: Errors should only be present when the status is `false`. I go into more detail in the detailed design section.

youngnick:
This is ready for rereview.

design/httpproxy-status-conditions.md
```diff
@@ -0,0 +1,305 @@
...
+
+## Non-goals
+Add support for some method of exposing the inclusion graph in the HTTPProxy’s status. That is a separate issue (#xxx)
+
+## High-level design
+
+We're going to do the following things:
+- add a `conditions` section to HTTPProxy `status`. We're doing this to allow for additional extensibility in the standard way.
+- Add a `Valid` condition as the standard HTTPProxy condition. We've chosen a `Valid` condition here rather than a more traditional `Ready` because `Ready` implies that we're sure that the Envoy config has been accepted - which we don't have the mechanisms to ensure yet. We may add a `Ready` condition at a later date.
+- The `Valid` condition's `Reason` field and the `validity` field will have the same value, as will the `Valid` condition's message field and the `description` field. That is, we will keep the functionality of the `validity` and `description` fields, but also duplicate it in the `Valid` Condition.
```
jpeach, line 32: Did you mean `currentStatus` rather than`validity`? We should consider adding a string enum (with validation) for this to make it clear what the possible values are.

design/httpproxy-status-conditions.md
```diff
@@ -0,0 +1,305 @@
...
+
+We will create the following generic type as a stand-in until the `metav1.Condition` type is available upstream.
+
+```go
+type Condition struct {
+	// Type of condition in CamelCase or in foo.example.com/CamelCase.
+	// Many .condition.type values are consistent across resources like Available, but because arbitrary conditions can be
+	// useful (see .node.status.conditions), the ability to deconflict is important.
+	// +required
+	Type string `json:"type" protobuf:"bytes,1,opt,name=type"`
```
jpeach, line 50: Recommend that (here and below), we use a string enum for the Type and Reason fields. This is transparent on the wire, but gives is a structure to hang documentation from.

design/httpproxy-status-conditions.md
youngnick: Yes, sorry.

design/httpproxy-status-conditions.md
youngnick: It does mean that other people can't add additional Types into the `DetailedCondition` list though, which I would argue is desirable. We can document the Contour-supported values here, but making it an enum breaks extensibility - that's the reason for Condtions in the first place.

jpeach: IIRC the argument for other controllers manipulating conditions is that they can (relatively blindly) add their own conditions. If you need to change a condition that another controller creates, then you need to understand its semantics.

youngnick: Yeah, but if the `type` field is an enum, don't they then need to be able to add something to the Contour repo to be able to add their own type value?

design/httpproxy-status-conditions.md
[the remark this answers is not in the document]
stevesloka: This example still shows `type: Valid` with errors below it. I know it's just an example but want to clarify that its a case that shouldn't happen.

design/httpproxy-status-conditions.md
```diff
@@ -0,0 +1,305 @@
...
+      port: 80
+    - name: service-has-no-endpoints
+  virtualhost:
+    fqdn: foo2.bar.com
+    tls:
+      secretName: testsecret
+status:
+  conditions:
+  - type: Valid
+    status: true
```
stevesloka, line 191: warnings with `status: true` can't happen right?

youngnick: The warning is the `ServiceError`, not the `Valid` condition. This one means "The config is partially, valid, with some problems"

design/httpproxy-status-conditions.md
jpeach: If you use a Go string type, there's no wire enforcement, so it would be fine AFAIK. It would only be a constraint if you added a kubebuilder validation comment.

Issue #2459, opened 2020-04-23
External auth background and goals.


jpeach:
This issue captures background context and research for adding Contour support for service authentication. It's not the final design, though preliminary design directions are included. The goal of this issue is to capture feedback about the direction.

## Goals

1. Support multiple external authentication servers.

Since Contour targets soft multi-tenancy for proxies, we should expect
that different services have different authentication needs. This means
that Contour must support multiple authentication servers.

2. Support multiple external authentication types.

The Contour external authentication mechanism should be general enough to
support multiple kinds of authentication. At minimum, we should support
HTTP basic auth and OAuth2 (via OIDC).

3. Support externalName authentication servers.

For some kinds of authentication, the authentication server can run in
the same cluster. However, running an auth server has a certain amount
of overhead cost and may have specific security operational requirements,
so Contour should support using an out-of-cluster service as an external
authentication server.

4. Interoperate with existing authentication servers.

Writing a external authentication server can be non-trivial. Contour
should facilitate users who wish to use existing authentication
servers. Contour should not invent a bespoke protocol to attach
authentication servers.

5. Support GRPC and HTTP authentication servers.

Since Contour will not invent its own server attachment protocol,
exposing the Envoy protocol makes the most sense. Both the GRPC and
HTTP attachment protocols should be exposed since (1) other Envoy-based
Ingress controllers expose both and (2) there are authentication servers
that depend on both protocols.

6. Support TLS sessions with authentication servers.

Contour should always use TLS to send request to authentication
servers. Mutual TLS is not required, though it seems reasonable
for sites to want that capability.

7. Re-use existing HTTPProxy configuration as much as possible.

In API that describes how to bind an external authentication service,
Contour should leverage the existing configuration stanzas and concepts
in HTTPProxy as much as possible (e.g. Service, HeadersPolicy).

8. Disable authentication on specific paths.

Many applications have internal HTTP paths that don't require
authentication (e.g. static assets). There needs to be a way to disable
the authentication requirement for specific paths within a virtual host.

## Non-Goals

1. Internal or plugin authentication mechanisms.

We should not support any non-external authentication in Contour since
this would divert resources (development, testing and documentation)
from the external server path, which is the primary use case.  This means
that the Contour service itself should not satisfy authentication requests
or load plugins that do.

2. Insecure authentication server attachment.

Contour does not need to support attaching authentication servers over
HTTP. It can assume HTTPS.

3. Support Ingress and IngressRoute documents.

[...]

Issue #2325, opened 2020-03-05
Failures in External services configured in envoy should be visible to the application developer


youngnick:
We have a few requests currently to allow the configuration of external services of various kinds in Envoy: #432, #1691, and #370.

The thing that all of these have in common is that with our current Contour design, there is no way for Contour to do anything other than pass the config to Envoy.

This means that, for example, there is no way for someone configuring external auth for a particular HTTPProxy to know if that external auth is working correctly without having access to the external auth service.

Another example is that there is no way for someone using a service with a rate-limiter to know if their service has tripped the rate limit.

When something does go wrong with one of these services, there needs to be a way for someone using them indirectly to know where the problem is.

I think that this problem requires two things:

## Contour should be able to health check clusters in Envoy

All the external services must be configured as a cluster in Envoy. So, if Contour has a way to check the state of a cluster in Envoy (whether it has healthy endpoints, or some other information about it), then we have the information we need to pass to the application developer.

## Contour should be able to expose external service health info

Contour should be able to expose external service health info in the relevant place, whether that is a `status` field on an object like a HTTPProxy, a log line in Contour, a metric, or some combination of the above.

Obviously the first is a requirement for the second.

I'm not sure of the best way to check Envoy clusters from Contour, whether it's some gRPC thing, checking the stats by fetching them, or something else.

This issue is to cover:
- if this is a good idea
- doing the two steps if it is.

jpeach:
In principle, I think that exposing envoy's information about clusters is a good idea. It's useful for regular services and also for special infrastructure services.

jpeach:
Note to self:

General envoy cluster metrics that could be used to support this feature:

Name | Type | Desc
-- | -- | --
membership_healthy | Gauge | Current cluster healthy total (inclusive of both health checking and outlier detection)
membership_degraded | Gauge | Current cluster degraded total
membership_total | Gauge | Current cluster membership total
upstream_cx_none_healthy | Counter | Total times connection not established due to no healthy hosts

upstream_cx_none_healthy is pretty interesting if we can use it to create a level-based signal. Otherwise need some more research around the membership metrics.

youngnick:
[...]

Issue #1922, opened 2019-11-19
Forward auth support


jieyu:
**Please describe the problem you have**

We're looking for some equivalent functionality of forward auth in Traefik
https://docs.traefik.io/middlewares/forwardauth/
https://github.com/thomseddon/traefik-forward-auth

Is there any plan to support something similar to that? This is for the case when the user app does not have built-in support for authentication/authorization, and would like to have a general way to do that in the proxy.

stevesloka:
Hey @jieyu, this is currently an issue on the backlog (https://github.com/projectcontour/contour/issues/432). But it is something that I really would like to get added. We're going to be doing some planning for the next release (after 1.1), so we will have more details after 1.1 ships.

jpeach:
Relating #432, #433 , #986 and #1014 which are all touching on aspects of authentication support.

jpeach:
Related #68

MrSaints:
I've implemented a Contour / Envoy compatible `ext_authz` service to act as a forward authN middleware: https://github.com/MrSaints/forward-ext-authz-service

It is based on the same code, and flow used by Traefik's forward authN middleware. I've currently tested it with Pomerium.

youngnick:
That looks really neat, thanks @MrSaints!

PR #2643, opened 2020-06-30
design: add initial external authorization proposal

jpeach:
This updates #432.
This updates #2459.

### DIFF
--- a/design/external-authorization-design.md
+++ b/design/external-authorization-design.md
@@ -1,0 +1,710 @@
+# External Authorization Support
+
+Status: Approved
+
+## Abstract
+This document describes a design for performing request authorization for virtual hosts hosted by Contour.
+
+## Background
+
+- [#432][5] Provide support for integrating with an external auth service.
+- [#2459][3] External auth background and goals.
+- [#2325][9] Failures in External services configured in envoy should be visible to the application developer.
+
+## Goals
+- Allow operators to integrate existing authorization services with Contour.
+- Allow arbitrary types of authentication and authorization to be supported.
+- Decouple Contour from external authorization, so it can evolve at an independent rate.
+- Integrate cleanly with the HTTPProxy API.
+
+## Non Goals
+
+- Support for multiple external authorization mechanisms (i.e. the protocol the proxy uses to communicate with the authorization server).
+  Various proxies have different mechanisms for integrating external services, but it is impractical to support them all.
+  This means that an authorization servers written to integrate with NGINX or Traefik will not work without modification.
+- Abstracting the Envoy authorization mechanism.
+  The scope of abstracting the Envoy authorization mechanism is too large to be tractable.
+  Abstracting the protocol would also work against the goal of being able to integrate existing authorization servers.
+- Implement specific authorization protocols.
+  This design does not address implementing specific methods of authorizing client requests (e.g. OAuth2, SAML, LDAP, NTLM).
+  These specific authorization protocols can be implemented in external authorization servers (see discussion below).
+- Support for Ingress. Supporting Ingress resources is out of scope an not addressed in this design.
+
+## High-Level Design
+A new `ExtensionService` CRD adds a way to represent and track an authorization service.
+This CRD is relatively generic, so that it can be re-used for Envoy rate limiting and logging services.
+The core of the `ExtensionService` CRD is subset of the `projectcontour.v1.HTTPProxy` `Service` specification.
+Re-using the `Service` type allows the operator to specify configuration in familiar and consistent terms, especially TLS configuration.
+
+Note that only the Envoy [GRPC authorization protocol][2] will be supported.
+The GRPC protocol is a superset of the HTTP protocol and requires less configuration.
+The drawback of only supporting the GRPC protocol is that many existing Envoy authorization servers only support the HTTP protocol.
+Until Contour adds support for the v3 Envoy API, only the version 2 of the GRPC authorization protocol can be configured.
+
+Operators can bind an authorization service to a `HTTPProxy` using a new field in the `VirtualHost` struct.
+This field configures which `ExtensionService` resource to use, and allows the operator to set initial authorization policy.
+Authorization policy can also be set on a `Route`, so that application owners can pass metadata and disable authorization on specific routes.
+
+## Detailed Design
+
+### ExtensionService Changes
+
+The `ExtensionService` CRD has API version `projectcontour.io/v1alpha1`.
+
+There are a number of benefits to creating a CRD to represent a supporting service:
+
+- The CRD directly generates an Envoy [Cluster][4] that Contour can use for any purpose.
+  This means that with a single new API, Contour can add support for authorization, rate limiting and logging support services.
+- A CRD gives Contour a way to communicate the operational status of the support service, which was a [desired][9] [goal][7].
+- A CRD allows the team operating the support service and the team operating Contour to collaborate more loosely.
+
+```Go
+// SupportProtocolVersion is the version of the GRPC protocol used
+// to request support services.
+type SupportProtocolVersion string
+
+// SupportProtocolVersion2 requests the "v2" support protocol version.
+const SupportProtocolVersion2 SupportProtocolVersion = "v2"
+
+type ExtensionService struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   ExtensionServiceSpec   `json:"spec,omitempty"`
+	Status ExtensionServiceStatus `json:"status,omitempty"`
+}
+
+// ExtensionServiceStatus should follow the pattern being established in [#2642][10].
+// This field will updated by Contour.
+type ExtensionServiceStatus struct {
+	Conditions []Condition `json:"conditions"`
+}
+
+type ExtensionServiceSpec struct {
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:Required
+	Services []contourv1.Service `json:"services"`
+
+	// The load balancing policy for sending requests to the service.
+	// +optional
+	LoadBalancerPolicy *contourv1.LoadBalancerPolicy `json:"loadBalancerPolicy,omitempty"`
+
+	// +optional
+	TimeoutPolicy *contourv1.TimeoutPolicy `json:"timeoutPolicy,omitempty"`
+
+	// This field sets the version of the GRPC protocol that Envoy uses to
+	// send requests to the support service. Since Contour always uses the
+	// v2 Envoy API, this is currently fixed at "v2". However, other
+	// protocol options will be available in future.
+	//
+	// +kubebuilder:validation:Enum=v2
+	ProtocolVersion SupportProtocolVersion `json:"protocolVersion,omitempty"`
+}
+```
+
+The `Conditions` field in `ExtensionServicesStatus` follows the standardized Contour structure (proposed in #2642).
+This type should be shared with other Contour API status types.
+
+This document does not yet define how Contour monitors the support service.
+However, the presence of the `Conditions` field allows Contour to expose the status of the support service fairly directly.
+At minimum, Contour should observe the underlying Kubernetes Service and expose whether there are sufficient healthy Endpoints.
+This information, analogous to Deployments observing and exposing Pod status, can be published as a `Ready` condition.
+
+Each `ExtensionService` CRD generates a unique upstream Envoy [Cluster][4], which will emit standard Envoy metrics.
+Note that the Envoy cluster name can be non-obvious, so exposing it in status may be helpful.
+
+If the `Service` refers to a Kubernetes `ExternalName`, Contour should program Envoy to send the traffic to the external destination.
+
+The `ExtensionService` CRD re-uses the `Service` type from the `projectcontour.io/v1` API.
+However, the setting following fields can generate a validation errors:
+
+- `Protocol` may only be set to `h2` or `h2c` (the default should be `h2`).
+
+Note that the `Service` type does not include a field for a namespace name.
+This constrains `ExtensionService` resources to be in the same namespace as the `Service` they expose,
+ensuring that whoever creates the `ExtensionService` also has authority over the `Service` backing it.
+
+### HTTPProxy Changes
+
+```Go
+// ExtensionReference names a Contour extension resource (ExtensionService
+// by default).
+type ExtensionExtensionReference struct {
+	// API version of the referent.
+        // If this field has no value, Contour will use a default of "projectcontour.io/v1alpha1".
+	// +optional
+	APIVersion string `json:"apiVersion,omitempty" protobuf:"bytes,5,opt,name=apiVersion"`
+	// Kind of the referent.
+	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds
+        // If this field has no value, Contour will use a default of "ExtensionService".
+	// +optional
+	Kind string `json:"kind,omitempty" protobuf:"bytes,1,opt,name=kind"`
+	// Namespace of the referent.
+	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/
+        // If this field has no value, Contour will use the namespace of the enclosing HTTPProxy resource.
+	// +optional
+	Namespace string `json:"namespace,omitempty" protobuf:"bytes,2,opt,name=namespace"`
+	// Name of the referent.
+	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+	// +required
+	Name string `json:"name,omitempty" protobuf:"bytes,3,opt,name=name"`
+}
+
+type AuthorizationPolicy struct {
+	// When true, this field disables client request authentication
+	// for the scope of the policy.
+	//
+	// +optional
+	Disabled bool `json:"disabled,omitempty"`
+
+	// Context is a set of key/value literal strings that are sent to the
+	// authentication server in the check request. If a context
+	// is provided at an enclosing scope, the entries are merged
+	// such that the inner scope overrides matching keys from the
+	// outer scope.
+	//
+	// +optional
+	Context map[string]string `json:"context,omitempty"`
+}
+
+type AuthorizationExtension struct {
+        // ExtensionRef names the authorization service.
+	// +required
+	ExtensionRef ExtensionReference `json:"extensionRef"`
+
+        // AuthPolicy is the default authorization policy applied to client requests.
+	// +optional
+	AuthPolicy *AuthorizationPolicy `json:"authPolicy,omitempty"`
+
+        // If FailOpen is true, the client request is forwarded to the upstream service
+        // even if the authorization server fails to respond. This field should not be
+        // set in most cases. It is intended for use only while migrating applications
+        // from internal authorization to Contour external authorization.
+        //
+        // +optional
+        FailOpen bool `json:"failOpen,omitempty"`
+}
+
+// VirtualHost appears at most once. If it is present, the object is considered
+// to be a "root".
+type VirtualHost struct {
+    ...
+
+	// This field configures an extension service to perform
+	// authorization for this virtual host. Authorization can
+	// only be configured on virtual hosts that have TLS enabled.
+	// If the TLS configuration requires client certificate
+	///validation, the client certificate is always included in the
+	// authentication check request.
+	//
+	// +optional
+	Authorization *AuthorizationExtension `json:"authorization,omitempty"`
+}
+
+// Route contains the set of routes for a virtual host.
+type Route struct {
+    ...
+
+        // AuthPolicy updates the authorization policy for client
+	// requests that match this route.
+	// +optional
+	AuthPolicy *AuthorizationPolicy `json:"authPolicy,omitempty"`
+}
+```
+
+An operator configures authorization on a root `HTTPProxy` by setting the `AuthorizationExtension` field.
+Setting this field without also setting the `TLS` field is an error.
+
+The `AuthorizationExtension` field carries and optional default authorization policy.
+This policy sets the default policy for each subsequent `Route`.
+If `authPolicy.Disabled` is true, included `Routes` will not require authorization unless they configure their own authorization policy with `Disabled` set to false.
+The `authPolicy.Context` entries from the root `HTTPProxy` are merged with the context entries on each `Route`.
+Where the two policies have context keys in common, the key from the root `HTTPProxy` is overwritten.
+The merged `Context` entries are send to the authorization service as part of the check request.
+
+Contour should require authorization to be disabled on all routes that set `PermitInsecure` to true (see [Compatibility](#compatibility)).
+
+Contour sets the following Envoy external authorization configurations that operators cannot configure:
+
+- TLS client certificates are always sent if they are present
+- The client request body is never sent
+
+The `FailOpen` field in the `AuthorizationExtension` struct is present to make it easier to migrate applications.
+Currently, applications must perform authorization internally.
+Operators make experience bugs or availability issues migrating to external authorization, so the `FailOpen` field
+provides a way to gracefully migrate applications while the external authorization service may still be unreliable.
+
+See the [Compatibility](#compatibility) section for information on how authorization interacts with other `HTTPProxy` features.
+To summarize:
+1. Authorization is only available on HTTPS
+1. Routes that are marked `PermitInsecure` must also disable authorization
+1. Authorization is not compatible with using the fallback TLS certificate
+
+### Authorization Flows
+
+Authorizing requests with an external server involves a number of
+collaborating parties, so it can be helpful so understand how requests
+flow through the various services.
+
+- **Client:** A HTTP client.
+- **Proxy:** An Envoy proxy server.
+- **ExtAuth:** A external authorization server that implements the Envoy GRPC protocol.
+- **AuthProvider:** A server that can authenticate requests.
+  This could be an OIDC server, a server that can do LDAP authentication, or HTTP Basic auth.
+  It could be implemented in the same process as the ExtAuth service.
+- **Origin:** A HTTP origin server that responds to authenticated requests.
+
+#### Verifying an authorized request
+
+This flow describes what happens if the client already has already authenticated and obtained some sort of authorization token.
+
+```
++---------+                  +-------+                    +---------+             +---------------+ +---------+
+| Client  |                  | Proxy |                    | ExtAuth |             | AuthProvider  | | Origin  |
++---------+                  +-------+                    +---------+             +---------------+ +---------+
+     |                           |                             |                          |              |
+     | Send HTTP request         |                             |                          |              |
+     |-------------------------->|                             |                          |              |
+     |                           |                             |                          |              |
+     |                           | Authorize HTTP request      |                          |              |
+     |                           |---------------------------->|                          |              |
+     |                           |                             | -----------------\       |              |
+     |                           |                             |-| Verify request |       |              |
+     |                           |                             | |----------------|       |              |
+     |                           |                             |                          |              |
+     |                           |                      200 OK |                          |              |
+     |                           |<----------------------------|                          |              |
+     |         ----------------\ |                             |                          |              |
+     |         | Inject        |-|                             |                          |              |
+     |         | authorization | |                             |                          |              |
+     |         | metadata      | |                             |                          |              |
+     |         |---------------| | Forward HTTP request        |                          |              |
+     |                           |---------------------------------------------------------------------->|
+     |                           |                             |                          |              |
+     |                           |                             |                 Respond to HTTP request |
+     |                           |<----------------------------------------------------------------------|
+     |                           |                             |                          |              |
+     |     Forward HTTP response |                             |                          |              |
+     |<--------------------------|                             |                          |              |
+     |                           |                             |                          |              |
+```
+
+#### Signing in to a service
+
+This flow describes what happens if the client doesn't already have any authentication information and needs to sign in to some authentication service.
+
+Note that in the sign in flow, the ExtAuth server may need to generate a 302 redirection to send the client back to the authentication provider to obtain the proper authorization tokens.
+Once that happens, the client has to resend the original request and it will enter the verification flow above.
+
+```
++---------+                   +-------+                    +---------+      +---------------+ +---------+
+| Client  |                   | Proxy |                    | ExtAuth |      | AuthProvider  | | Origin  |
++---------+                   +-------+                    +---------+      +---------------+ +---------+
+     |                            |                             |                   |              |
+     | Send HTTP request          |                             |                   |              |
+     |--------------------------->|                             |                   |              |
+     |                            |                             |                   |              |
+     |                            | Authorize HTTP request      |                   |              |
+     |                            |---------------------------->|                   |              |
+     |                            |                             |                   |              |
+     |                            |                             | Authorize         |              |
+     |                            |                             |------------------>|              |
+     |                            |                             |                   |              |
+     |                            |                             |         Challenge |              |
+     |                            |                             |<------------------|              |
+     |                            |                             |                   |              |
+     |                            |                302 Redirect |                   |              |
+     |                            |<----------------------------|                   |              |
+     |                            | ----------------\           |                   |              |
+     |                            |-| Generate      |           |                   |              |
+     |                            | | authorization |           |                   |              |
+     |                            | | redirection   |           |                   |              |
+     |     Authorization redirect | |---------------|           |                   |              |
+     |<---------------------------|                             |                   |              |
+     |                            |                             |                   |              |
+     | Sign in                    |                             |                   |              |
+     |----------------------------------------------------------------------------->|              |
+     |                            |                             |                   |              |
+     |                            |                          Authorization response |              |
+     |<-----------------------------------------------------------------------------|              |
+     |                            |                             |                   |              |
+     | Resend HTTP request        |                             |                   |              |
+     |--------------------------->|                             |                   |              |
+     |                            |                             |                   |              |
+```
+
+## Alternatives Considered
+
+1. Contour could install itself as the authorization server.
+   This could remove some of the limitations of the Envoy configuration structure at the cost of more complexity in Contour.
+1. Integrate external authorization directly into `HTTPProxy`.
+   This increases the complexity of the `HTTPProxy` structure and makes it difficult to re-use the same authorization service acrtoss multiple proxies.
+   A separate CRD gives better opportunities to expose useful operational status.
+   Integrating specific authorization parameters into `HTTPProxy` prevents operators implementing their own authorization flows.
+
+## Security Considerations
+
+- Most HTTP authorization protocols depend on HTTPS to keep headers and URL query parameters private.
+  Supporting only authorization on HTTPS, Contour reduces the changes of misconfiguring this.
+- HTTPS sessions between Envoy and the authorization server are required.
+  TLS validation can be configured if necessary (should be recommended),
+- Authorization services can run in separate Kubernetes namespaces with limited privilege.
+
+## Compatibility
+
+Contour can currently support v2 of the Envoy GRPC authorization protocol.
+When Contour moves to v3 of the Envoy API, it can support both v2 and v3 authorization protocols.
+
+The `ExtensionService` CRD supports GRPC authorization support services, but not HTTP ones.
+
+### HTTPS Only
+
+As discussed in [#2459][3], the Envoy external authorization HTTP filter is configured on the HTTP Connection Manager.
+This means that the scope of how Contour can configure authorization depends on the structure of the HTTP configuration.
+For HTTP, there is only one HTTP Connection Manager that contains all the virtual hosts.
+Configuring a single authorization service for all the virtual hosts might be acceptable for some organizations, but it doesn't really fix the Contour multi-tenancy model.
+On the other hand, Contour configures HTTPS with a separate HTTP Connection Manager for each virtual host, so different authorization servers can naturally be attached to each HTTPS virtual host.
+
+The result of this is that authorization servers are not supported on HTTP.
+While it is convenient from an implementation perspective, this policy is also consistent with Contour's security-first posture.
+Note that using the TLS fallback certificate (for non-SNI clients) has the same HTTP Connection Manager properties as HTTP, so authentication servers also cannot be configured on virtual hosts that share the fallback certificate.
+
+### PermitInsecure
+
+If a `Route` has the `PermitInsecure` field set to true, Contour will program it into both the secure and insecure Envoy listeners.
+This means that the same `Route` will be authenticated differently depending on how the client accesses it.
+In this case, it may be impossible for the upstream service to reliably detect whether the request is authorized.
+To resolve this, Contour should require authorization to be disabled on all routes that set `PermitInsecure` to true.
+
+### Fallback Certificates
+
+The fallback certificate feature is used to support TLS access by clients that do not support the SNI extension.
+This is implemented by a default TLS inspection matcher that catches session requests that do not match an existing SNI name.
+This matcher contains a HTTP Connection Manager that routes requests to all the virtual hosts that request fallback certificate support.
+Because the Envoy `ext_authz` filter applies to a whole HTTP Connection Manager, it is incompatible with this structure.
+There is no way to express that different virtual hosts can have different `ext_authz` configuration while using a single HTTP Connection Manager.
+
+## Implementation
+
+The Contour implementation provides a framework, but isn't useful until authorization services are available for the protocols that Contour operators need.
+
+The Contour implementation is quite large, but can be separated into smaller stages:
+
+- Support service CRD implementation
+- Support service monitoring and status
+- `HTTPProxy` integration
+- Integration tests
+- Documentation
+
+Since the scope of authorization is quite large, documentation should be written as a new top level section.
+If the documentation is included in the main `httpproxy.md` page, it will be difficult to organize and consume.
+
+### Contour Authserver
+
+The Contour project should host a simple server that can be used for testing and for simple authorization use cases.
+
+The `contour-authserver` server implements two authentication backends.
+First, a no-op test server that authorizes every request.
+Second, a [HTTP basic authentication][1] server that authenticates request using htpasswd files stored as Kubernetes secrets.
+
+### OIDC Authentication
+
+Most enterprise users will want some combination of OAUTH2, SAML or OIDC.
+Writing an Envoy authorization proxy for these protocols is considerably more complex than for HTTP basic authentication, and will need to be staffed if they are to be supported.
+There are existing implementations of all these protocols, though to my knowledge none are exposed through the Envoy GRPC.
+It should be possible to write an authorization proxy that integrates with specific implementations (e.g. [Dex](https://github.com/dexidp/dex)).
+
+One open issue for OIDC is whether Contour ought to program the Envoy [JWT][6] filter.
+Configuring this filter would substantially increase the required `HTTPProxy` API, and it's likely that the authorization server would still need to be invoked.
+This issue should be revisited after gaining some implementation experience.
+
+## Sample Configuration
+
+First, add a cert-manager issuer so that we can easily self-sign TLS
+certificates.
+
+```yaml
+apiVersion: cert-manager.io/v1alpha3
+kind: ClusterIssuer
+metadata:
+  name: selfsigned
+spec:
+  selfSigned: {}
+```
+
+Next, deploy the service that needs to be authorized, and request a TLS
+certificate for it.
+
+```yaml
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: ingress-conformance-echo
+
+---
+
+apiVersion: v1
+kind: Service
+metadata:
+  name: ingress-conformance-echo
+
+---
+
+apiVersion: cert-manager.io/v1alpha2
+kind: Certificate
+metadata:
+  name: echo
+spec:
+  dnsNames:
+  - echo.projectcontour.io
+  secretName: echo
+  issuerRef:
+    name: selfsigned
+    kind: ClusterIssuer
+
+```
+
+Next, deploy the authorization service.
+This is a test service that will just authorize all requests.
+The authorization service, it's TLS secret and service are all placed into the `contour-auth` namespace.
+
+```yaml
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: contour-auth
+
+---
+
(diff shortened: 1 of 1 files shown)


design/external-authorization-design.md
```diff
@@ -0,0 +1,453 @@
...
+This document describes a design for performing request authorization for virtual hosts hosted by Contour.
+
+## Background
+
+- [#432][5] Provide support for integrating with an external auth service.
+- [#2459][3] External auth background and goals.
+
+## Goals
+- Allow operators to integrate existing authorization services with Contour.
+- Support arbitrary types of authentication and authorization.
```
stevesloka, line 15: We should spell out the the specific types this proposal incorporates.

design/external-authorization-design.md
```diff
@@ -0,0 +1,453 @@
...
+- Decouple Contour from external authorization, so it can evolve at an independent rate.
+- Integrate cleanly with the HTTPProxy API.
+
+## Non Goals
+
+- Support for multiple external authorization mechanisms.
```
stevesloka, line 21: Can you clarify these two bits? I'm not sure what is not going to be a goal.

design/external-authorization-design.md
```diff
@@ -0,0 +1,453 @@
...
+The GRPC protocol is a superset of the HTTP protocol and requires less configuration.
+The drawback of only supporting the GRPC protocol is that many existing Envoy authorization servers only support the HTTP protocol.
+Until Contour adds support for the v3 Envoy API, only the version 2 of the GRPC authorization protocol can be configured.
+
+The authorization service is bound to a `HTTPProxy` with a new field in the `VirtualHost` struct.
+This field configures which `SupportService` resource to use, and allows the operator to set initial authorization policy.
+Authorization policy can also be set on a `Route`, so that application owners can pass metadata and disable authorization on specific routes.
+
+As discussed in [#2459][3], the Envoy external authorization HTTP filter is configured on the HTTP Connection Manager.
+This means that the scope of how Contour can configure authorizationi depends on the structure of the HTTP configuration.
```
stevesloka, line 44: s/authorizationi/authorization

design/external-authorization-design.md
```diff
@@ -0,0 +1,453 @@
...
+This type should be shared with other Contour API status types.
+
+This document does not yet define how Contour monitors the support service.
+However, the presence of the `Conditions` field allows Contour to expose the status of the support service fairly directly.
+At minimum, Contour should monitor the underlying Kubernetes service and expose whether there are sufficient healthy endpoints.
+
+Each `SupportService` CRD generates a unique upstream Envoy [Cluster][#4], which will emit standard Envoy metrics.
+Note that the Envoy cluster name can be non-obvious, so exposing it in status may be helpful.
+
+The `SupportService` CRD re-uses the `Service` type from the `projectcontour.io/v1` API.
```
stevesloka, line 107: I don't think I'm keen on this idea of re-use. If we're going to build a new CRD, let's do that and make the user experience simple.

design/external-authorization-design.md
```diff
@@ -0,0 +1,453 @@
...
+	ProtocolVersion SupportProtocolVersion `json:"protocolVersion,omitempty"`
+}
+```
+
+The `Conditions` field in `SupportServicesStatus` follows the standardized upstream structure.
+This type should be shared with other Contour API status types.
+
+This document does not yet define how Contour monitors the support service.
+However, the presence of the `Conditions` field allows Contour to expose the status of the support service fairly directly.
+At minimum, Contour should monitor the underlying Kubernetes service and expose whether there are sufficient healthy endpoints.
```
stevesloka, line 102: Ideally we encourage folks to setup readiness probes and rely on k8s to do the lifting, Contour is just going to record which of those pass the probe. My point here is that Contour shouldn't be doing active checks against the external service.

design/external-authorization-design.md
```diff
@@ -0,0 +1,453 @@
...
+	// send requests to the support service. Since Contour always uses the
+	// v2 Envoy API, this is currently fixed at "v2". However, other
+	// protocol options will be available in future.
+	//
+	// +kubebuilder:validation:Enum=v2
+	ProtocolVersion SupportProtocolVersion `json:"protocolVersion,omitempty"`
+}
+```
+
+The `Conditions` field in `SupportServicesStatus` follows the standardized upstream structure.
```
stevesloka, line 97: Link to upstream reference?

design/external-authorization-design.md
```diff
@@ -0,0 +1,453 @@
...
+	AuthPolicy *AuthorizationPolicy `json:"authPolicy,omitempty"`
+}
+```
+
+An operator configures authorization on a root `HTTPProxy` by setting the `AuthorizationServer` field.
+Setting this field without also setting the `TLS` field is an error.
+
+The `AuthorizationServer` field carries and optional default authorization policy.
+This policy sets the default policy for each subsequent `Route`.
+If `authPolicy.Disabled` is true, included `Routes` will not require authorization unless they configure their own authorization policy with `Disabled` set to false.
```
stevesloka, line 200: How does `permitInsecure` play into this?

design/external-authorization-design.md
```diff
@@ -0,0 +1,453 @@
...
+- Support service CRD implementation
+- Support service monitoring and status
+- `HTTPProxy` integration
+- Integration tests
+- Documentation
+
+Since the scope of authorization is quite large, documentation should be written as a new top level section.
+If the documentation is included in the main `httpproxy.md` page, it will be difficult to organize and consume.
+
+### Contour Authserver
```
stevesloka, line 250: I think we should just re-use existing things instead of building a new one.

design/external-authorization-design.md
```diff
@@ -0,0 +1,453 @@
...
+
+- Support for multiple external authorization mechanisms.
+  Various proxies have different mechanisms for integrating external services, but it is impractical to support them all.
+
+- Abstracting Envoy authorization protocols.
+  The scope of abstracting the Envoy authorization protocol is too large to be tractable.
+  Abstracting the protocol would also work against the goal of being able to integrate existing authorization servers.
+
+## High-Level Design
+A new CRD `SupportService` adds a way to represent and track an authorization service.
```
stevesloka, line 29: I think we should just build out a proper CRD for Auth and not try to create a generic one that is useful for many purposes. I feel in the long run, we're going to get into a spot where it's super difficult to setup since it's so generic.

stevesloka:
Another thought I had was adding auth to `Conditions` somehow, so you could allow operators to tell a child set of Proxies should have specific auth, but then not allow them to configure from there. I think some operators might not want each route to be disabled for example.

jpeach:
> Another thought I had was adding auth to `Conditions` somehow, so you could allow operators to tell a child set of Proxies should have specific auth, but then not allow them to configure from there. I think some operators might not want each route to be disabled for example.

I'm not really following this, can you explain this a bit more?

design/external-authorization-design.md
jpeach: I don't think it's tractable to build that. There are an unbounded set of different authorization mechanisms and trying to get them all into a single CRD will be a mess. It has the other disadvantage that it's not extensible by operators, so that only Contour maintainers can implement new authorization types.

design/external-authorization-design.md
jpeach: I agree. However, we need something to test, which is this. If there's an existing Envoy authserver that covers this case, we would support that.

design/external-authorization-design.md
jpeach: Previously, when dependencies on other services have been proposed, having Contour make active availability checks against those has been suggested. Here, I'm not saying we should actually do that, but it's a possibility that we can leave open.

What I'm suggesting here is consistent with your suggestion I think. Rather than operators checking the state of the underlying service pods, Contour can do that on their behalf (since it already knows about endpoints), and roll that information up to the support service status.

design/external-authorization-design.md
jpeach: This is about creating the mechanism that will let anyone add specific types.

design/external-authorization-design.md
jpeach: What I meant by this is that different proxies have different protocol mechanisms for incorporating external authorization servers. We aren't going to support multiple mechanisms (i.e. you can't use your external auth server from nginx unmodified).

design/external-authorization-design.md
jpeach: In re-using the types, my goal was to keep it simple, since users should already be familiar with the service struct from using the HTTPProxy.

jpeach:
> * How does the actual auth flow work?

There's some relevant sequence diagrams in https://github.com/projectcontour/contour/issues/2459. I can reproduce them here for context.

>  You mentioned OIDC, but not how we might configure that?

You would create a `SupportService` CRD that references a service that implements OIDC. That will need to be configured with all the necessary OIDC parameters. There's still a bunch of work needed to explore and document how to do this.

> * You mentioned not allowing external auth servers, so this proposal requires all auth to exist inside the k8s cluster where Contour is running?

If you used an ExternalName Service to back the SupportService CRD, then you could have an auth server outside the cluster. Not sure if that will need special code to support, or whether we would get it for free from Contour's existing Service support.

design/external-authorization-design.md
jpeach: That's a good point, I'll check on that.

jpeach:
xref #1922

design/external-authorization-design.md
```diff
@@ -0,0 +1,453 @@
...
+- Most HTTP authorization protocols depend on HTTPS to keep headers and URL query parameters private.
+  Supporting only authorization on HTTPS, Contour reduces the changes of misconfiguring this.
+- HTTPS sessions between Envoy and the authorization server are required.
+  TLS validation can be configured if necessary (should be recommended),
+- Authorization services can run in separate Kubernetes namespaces with limited privilege.
+
+## Compatibility
+
+Contour can support v2 of the Envoy GRPC authorization protocol.
+When Contour moves to v3 of the Envoy API, it ccan support both v2 and v3 authorization protocols.
```
pims, line 231: nit: `s/ccan/can`

design/external-authorization-design.md
```diff
@@ -0,0 +1,453 @@
...
+
+The `SupportService` CRD supports GRPC authorization support services, but not HTTP ones.
+
+## Implementation
+
+The Contour implementation provides a framework, but isn't useful until authorization services are exist for the protocols that Contour operators need.
```
pims, line 237: nit: `s/are//`

design/external-authorization-design.md
jpeach: I rephrased to reference #2642.

design/external-authorization-design.md
jpeach: If a route can be insecure, then I don't see a way for the upstream service to guarantee that it can know whether a request has been authorized or not. I updated to specify that insecure routes must disable authorization.

design/external-authorization-design.md
jpeach: Fixed!

design/external-authorization-design.md
jpeach: Fixed!

jpeach:
FYI, @pims convinced me that being able to toggle the fail-open or fail-closed behavior is needed when migrating from internal application auth to external auth, so I'm planning to add that flag to the API.

jpeach:
@jpeach Note to self ... consider what happens to headers injected by the external auth server on routes that have auth disabled. A client request could still present those headers, causing an application to become confused (a request could accidentally be performed as an authorized identity).

So perhaps we need to add API to ensure that any injected auth headers are scrubbed from the incoming client request. Envoy has the concept of internal headers, which aligns fairly closely, but that (and the API for removing headers) requires listing specific headers (would be more robust to pattern match IMHO). The Envoy concept on internal requests is also a bit hard to understand and it's not obvious that it would be the right choice in all deployments.

design/external-authorization-design.md
stevesloka: Sorry still not 100% following you on this thought. Could you rewrite the nongoal in a way that better explains it? Right now it reads as you aren't going to support the auth provider that I want to bring which I know is not the case.

design/external-authorization-design.md
stevesloka: We got burned by this before when re-using. It's an internal type, I'd rather bite the work to create a 5 line struct vs all the additional logic to make sure users don't reuse this field incorrectly due to the different context.

design/external-authorization-design.md
stevesloka: > What I'm suggesting here is consistent with your suggestion I think.

This was not my intent, my thought was to take Contour out of the business of doing these checks. They are external systems that Contour doesn't control, we should use the primitives provided by Kubernetes to implement.

design/external-authorization-design.md
jpeach: OK, I can copy a matching Service type across, leaving out the unnecessary fields. I guess it will need to have a unique name, since this CRD will eventually migrate to the same package.

design/external-authorization-design.md
jpeach: > > What I'm suggesting here is consistent with your suggestion I think.
> 
> This was not my intent, my thought was to take Contour out of the business of doing these checks. They are external systems that Contour doesn't control, we should use the primitives provided by Kubernetes to implement.

Yes, I agree with you :)

what I'm trying to say is that "monitor the underlying Kubernetes service" doesn't have to mean "actively health check"; it can mean "see whether there are any endpoints alive", or whatever we agree makes sense. The goal here is that we can take advantage of the CRD status to make the status of the underlying objects more obvious, just like deployments do for pods.

https://github.com/projectcontour/contour/issues/2325

design/external-authorization-design.md
jpeach: > Support for multiple external authorization mechanisms (i.e. the protocol the proxy uses to communicate with the authorization server).

The mechanism is the protocol by which the proxy communicates with the authorization server.

> Various proxies have different mechanisms for integrating external services, but it is impractical to support them all.

Nginx, envoy, traefik, etc all use different external authorization mechanisms (ie. communicating with the authorization server follows different protocols or conventions). What I'm getting at is that it is not a goal for an authorization server written for nginx to work with Contour.

design/external-authorization-design.md
```diff
@@ -0,0 +1,669 @@
...
+```Go
+// SupportServiceReference names a SupportService resource.
+type SupportServiceReference struct {
+	// API version of the referent.
+	// +required
+	APIVersion string `json:"apiVersion,omitempty" protobuf:"bytes,5,opt,name=apiVersion"`
+	// Kind of the referent.
+	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds
+	// +required
+	Kind string `json:"kind,omitempty" protobuf:"bytes,1,opt,name=kind"`
```
youngnick, line 139: Should we make this optional with the behavior being that empty is equivalent to `SupportService`? I know you're basing this off `TypedLocalObjectReference`, but I think it's worth considering a slightly different behavior in this case.

design/external-authorization-design.md
```diff
@@ -0,0 +1,669 @@
...
+type AuthorizationPolicy struct {
+	// When true, this field disables client request authentication
+	// for the scope of the policy.
+	//
+	// +optional
+	Disabled bool `json:"disabled,omitempty"`
+
+	// Context is a set of key/value pairs that are sent to the
+	// authentication server in the check request. If a context
+	// is provided at an enclosing scope, the entries are merges
```
youngnick, line 159: Nit:
```suggestion
	// is provided at an enclosing scope, the entries are merged
```

design/external-authorization-design.md
```diff
@@ -0,0 +1,669 @@
...
+}
+
+type AuthorizationPolicy struct {
+	// When true, this field disables client request authentication
+	// for the scope of the policy.
+	//
+	// +optional
+	Disabled bool `json:"disabled,omitempty"`
+
+	// Context is a set of key/value pairs that are sent to the
```
youngnick, line 157: Do the values need to allow some sort of `%` interpolation or something? That might be worth mentioning if so.

design/external-authorization-design.md
youngnick: I think it's clearer to have a similar type that doesn't have restrictions about not setting fields that are present - better to just not have fields that aren't relevant.

design/external-authorization-design.md
youngnick: Maybe something like `Contour should observe the status of the service, directly or indirectly` would make this clearer.

design/external-authorization-design.md
```diff
@@ -0,0 +1,669 @@
...
+  Various proxies have different mechanisms for integrating external services, but it is impractical to support them all.
+
+- Abstracting the Envoy authorization mechanism.
+  The scope of abstracting the Envoy authorization mechanism is too large to be tractable.
+  Abstracting the protocol would also work against the goal of being able to integrate existing authorization servers.
+
+## High-Level Design
+A new `SupportService` CRD adds a way to represent and track an authorization service.
+This CRD is relatively generic, so that it can be re-used for Envoy rate limiting and logging services.
+The core of the `SupportService` CRD is a `projectcontour.v1.HTTPProxy` `Service` specification.
```
youngnick, line 32: This part will need updating if we make a new struct, I would propose something like:
```suggestion
The core of the `SupportService` CRD is a subset of the `projectcontour.v1.HTTPProxy` `Service` specification.
```

design/external-authorization-design.md
```diff
@@ -0,0 +1,669 @@
...
+
+## SupportServiceReference kind field
+
+Contour doesn't need to support arbitrary types of support service resources.
+The design here includes a `Kind` field because it is modeled after `TypedLocalObjectReference`.
+
+## SupportService status
+
+One of the original requirements for adding dependencies on other services was being able to expose useful operational status.
+The `SupportService` status field is where this information would be exposed, but this design does not address what status should be exposed.
```
youngnick, line 648: I'd say at a minimum there should be a `Ready` condition (in this case indicating that the SupportService is ready to service requests). Probably, we'll want some fields eventually, but `Ready` allows for using the "wait for condition" flow easily.

design/external-authorization-design.md
jpeach: There's no interpolation, just arbitrary key/value strings. As we get more implementation experience with authorization servers, some sort of interpolation feature could turn out to be useful.

design/external-authorization-design.md
jpeach: I looked at this, and we would also have to duplicate the `UpstreamValidation`, `HeadersPolicy` and `LoadBalancerPolicy`. That's possible, but it's also pretty intrusive; it means that we will have to do substantial refactoring to be able to apply helpers like `determineSNI()`, `getProtocol()`, etc. However, it's also possible that this refactoring will end up with cleaner internal APIs (maybe create a `model` package for types that compose the DAG?).

We also need to keep in mind that this API will graduate to `projectcontour/v1` and make sure that what we do now will make sense in terms of that destination. Would we want to deduplicate the types at that point, or keep the more specific types?

design/external-authorization-design.md
jpeach: If we default the kind, then users would do:
```yaml
    authorization:
      serviceRef:
        name: htpasswd
        namespace: projectcontour-auth
```

Maybe we could remove or default `apiVersion` as well, depending on how we manage the versioning of this type. I'm also open to just removing these fields altogether.

youngnick: I think it's better to default them than remove them, that gives us the chance to use other kinds later.

design/external-authorization-design.md
youngnick: Really? That much duplication? We can't inline the policies? I only meant that it's better not to use something that has fields that you shouldn't set - that's a pretty confusing UX.

I also hadn't though about methods and type-passing, that will make copying the type a significant refactor. I'm not so enamoured of the idea with that much extra work added.

design/external-authorization-design.md
youngnick: Awesome, I'd prefer no interpolation, that is a whole can of worms. I meant more if it was expected as part of this type of functionality.

design/external-authorization-design.md
youngnick: This can be dealt with later though; not blocking.

design/external-authorization-design.md
stevesloka: I guess the ask here was to make the doc more clear to remove this confusion.

design/external-authorization-design.md
stevesloka: I posted in slack, but why do you need UpstreamValidation, HeadersPolicy and LoadBalancerPolicy? We're just connecting Envoy to another service. That piece is confusing to me which is why I suggested not to resue the service struct.

jpeach: Envoy should connect to the service over TLS, so you need the `UpstreamValidationPolicy` to express how to check the server certificate. If the service is an externalName service, you might need to rewrite the host header, so you'd need `HeadersPolicy`. There could be other operational reasons to add the headers, e.g. assigning Linkerd headers. The service probably has multiple pods, so `LoadBalancerPolicy` is probably a useful control, otherwise we would just hard-code a load balancing policy.

design/external-authorization-design.md
```diff
@@ -0,0 +1,670 @@
...
+}
+
+// Route contains the set of routes for a virtual host.
+type Route struct {
+    ...
+
+    // AuthPolicy updates the authorization policy for client
+	// requests that match this route.
+	// +optional
+	AuthPolicy *AuthorizationPolicy `json:"authPolicy,omitempty"`
```
stevesloka, line 215: Is there a way that operators can not allow users to change this? I'm thinking that an operator wants to have auth on for a set of routes, what stops a user from circumventing that and disabling?

design/external-authorization-design.md
```diff
@@ -0,0 +1,670 @@
...
+
+	// This field configures a support service to perform
+	// authorization for this virtual host. Authorization can
+	// only be configured on virtual hosts that have TLS enabled.
+	// If the TLS configuration requires client certificate
+	///validation, the client certificate is always included in the
+	// authentication check request.
+	//
+	// +optional
+	Authorization *AuthorizationServer `json:"authorization,omitempty"`
```
stevesloka, line 205: Should this be part of the `TLS` stuct?

design/external-authorization-design.md
stevesloka: Those are the bits I don't understand. 

- How would an external name service fit? The goal of this CRD is to connect Envoy to the Auth service over gRPC. Are you saying folks would deploy the Auth server outside the cluster and Envoy would reach out over the external name service? 
- Kubernetes doesn't provide a way to set load balancer policies, is that really required for this implementation?
- Is external name the only reason to have a Headers Policy? This is another field I don't understand how it would be applied in this situation and be useful. 

It seems like the only reason to  have the new crd is to allow for setting conditions, but I have questions about how that will work.

design/external-authorization-design.md
```diff
@@ -0,0 +1,670 @@
...
+type SupportService struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   SupportServiceSpec   `json:"spec,omitempty"`
+	Status SupportServiceStatus `json:"status,omitempty"`
+}
+
+type SupportServiceStatus struct {
+	Conditions []Condition `json:"conditions"`
```
stevesloka, line 84: How do users write to this field? If I build my own auth server to handle my auth mechanism, who manages this field? It's outside the scope of Contour since we may not own the implementation and can't understand what values to write.

design/external-authorization-design.md
stevesloka, line 215: Can I have an AuthorizationPolicy in an include? I'm thinking of setting a bunch of routes that should be insecure, you could include them and add this to the includes conditions without having to set them on each child route.

design/external-authorization-design.md
```diff
@@ -0,0 +1,670 @@
...
+	// +optional
+	Context map[string]string `json:"context,omitempty"`
+}
+
+type AuthorizationServer struct {
+	// +required
+	ServiceRef SupportServiceReference `json:"serviceRef"`
+
+	// +optional
+	AuthPolicy *AuthorizationPolicy `json:"authPolicy,omitempty"`
```
stevesloka, line 173: Should we not allow `Disabled` at the root level? This essentially would allow users to configure auth, then disable it at the root and child routes would have to enable by setting `Disabled=false` on their route?

design/external-authorization-design.md
```diff
@@ -0,0 +1,670 @@
...
+	// authentication server in the check request. If a context
+	// is provided at an enclosing scope, the entries are merged
+	// such that the inner scope overrides matching keys from the
+	// outer scope.
+	//
+	// +optional
+	Context map[string]string `json:"context,omitempty"`
+}
+
+type AuthorizationServer struct {
```
stevesloka, line 168: It feels like these settings (AuthPolicy, Timeout, Failopen) should be part of the `SupportService` CRD. That's where the contract to the upstream service is defined / configured. 

If many vhosts utilize the same auth server, you'd have to configure these for each vhost and keep in sync.

design/external-authorization-design.md
```diff
@@ -0,0 +1,670 @@
...
+     |--------------------------->|                             |                   |              |
+     |                            |                             |                   |              |
+```
+
+## Alternatives Considered
+
+1. Contour could install itself as the authorization server.
+   This could remove some of the limitations of the Envoy configuration structure at the cost of more complexity in Contour.
+1. Integrate external authorization directly into `HTTPProxy`.
+   This increases the complexity of the `HTTPProxy` structure and makes it difficult to re-use the same authorization service acrtoss multiple proxies.
```
stevesloka, line 343: nit: s/acrtoss/across

stevesloka:
Also don't want to loose context from a Slack conversation about renaming s/SupportService/ExtensionService: https://kubernetes.slack.com/archives/C8XRH2R4J/p1594707008255100

design/external-authorization-design.md
stevesloka: Was your thought on headers policy to map to the different response types like `OkHttpResponse`?   https://www.envoyproxy.io/docs/envoy/v1.14.3/api-v2/service/auth/v2/external_auth.proto#service-auth-v2-deniedhttpresponse

design/external-authorization-design.md
```diff
@@ -0,0 +1,670 @@
...
+This CRD is relatively generic, so that it can be re-used for Envoy rate limiting and logging services.
+The core of the `SupportService` CRD is subset of the `projectcontour.v1.HTTPProxy` `Service` specification.
+Re-using the `Service` type allows the operator to specify configuration in familiar and consistent terms, especially TLS configuration.
+
+Note that only the Envoy [GRPC authorization protocol][2] will be supported.
+The GRPC protocol is a superset of the HTTP protocol and requires less configuration.
+The drawback of only supporting the GRPC protocol is that many existing Envoy authorization servers only support the HTTP protocol.
+Until Contour adds support for the v3 Envoy API, only the version 2 of the GRPC authorization protocol can be configured.
+
+Operators can bind an authorization service to a `HTTPProxy` using a new field in the `VirtualHost` struct.
```
skriss, line 40: if I'm understanding correctly, this implies that all routes for a given vhost must use the same auth service (or disable it). Is that workable for users? Are there cases where users would want to use a different auth service for different subsets of the routes under a vhost?

design/external-authorization-design.md
```diff
@@ -0,0 +1,670 @@
...
+	ProtocolVersion SupportProtocolVersion `json:"protocolVersion,omitempty"`
+}
+```
+
+The `Conditions` field in `SupportServicesStatus` follows the standardized Contour structure (proposed in #2642).
+This type should be shared with other Contour API status types.
+
+This document does not yet define how Contour monitors the support service.
+However, the presence of the `Conditions` field allows Contour to expose the status of the support service fairly directly.
+At minimum, Contour should observe the underlying Kubernetes Service and expose whether there are sufficient healthy Endpoints.
```
skriss, line 110: how is "sufficient healthy Endpoints" defined/determined?

design/external-authorization-design.md
skriss: Given that this typed is named `SupportServiceReference`, the `Kind` field seems like it'd be either redundant or contradictory, unless the intent is to allow specializations of SupportService (e.g. AuthSupportService, RateLimitingSupportService)?

design/external-authorization-design.md
```diff
@@ -0,0 +1,670 @@
...
+## Compatibility
+
+Contour can currently support v2 of the Envoy GRPC authorization protocol.
+When Contour moves to v3 of the Envoy API, it can support both v2 and v3 authorization protocols.
+
+The `SupportService` CRD supports GRPC authorization support services, but not HTTP ones.
+
+## Implementation
+
+The Contour implementation provides a framework, but isn't useful until authorization services are available for the protocols that Contour operators need.
```
skriss, line 364: Are we (Contour maintainers) realistically going to be building & maintaining all of the auth services? Or are there other obvious parties who would be interested in collaborating on these or building new ones?

I'm just wondering, if practically we're going to be building and maintaining all of them, if that might change how we'd approach the framework design.

design/external-authorization-design.md
```diff
@@ -0,0 +1,670 @@
...
+- [#2459][3] External auth background and goals.
+- [#2325][9] Failures in External services configured in envoy should be visible to the application developer.
+
+## Goals
+- Allow operators to integrate existing authorization services with Contour.
+- Allow arbitrary types of authentication and authorization to be supported.
+- Decouple Contour from external authorization, so it can evolve at an independent rate.
+- Integrate cleanly with the HTTPProxy API.
+
+## Non Goals
```
skriss, line 20: I think it'd be useful to explicitly call out somewhere in Goals or Non-Goals that this design proposes the framework, but does not propose auth server implementations for any specific auth protocols, and that those will also need to be implemented before a user can actually make use of external auth.

design/external-authorization-design.md
jpeach: > Those are the bits I don't understand.
> 
> * How would an external name service fit? The goal of this CRD is to connect Envoy to the Auth service over gRPC. Are you saying folks would deploy the Auth server outside the cluster and Envoy would reach out over the external name service?

Yes, this would work.

> * Kubernetes doesn't provide a way to set load balancer policies, is that really required for this implementation?

Required? Maybe not. But it's obviously useful and easy to implement.

> * Is external name the only reason to have a Headers Policy? This is another field I don't understand how it would be applied in this situation and be useful.

I gave 2 possible examples above. Probably creative operators can think of more uses for this.

> It seems like the only reason to have the new crd is to allow for setting conditions, but I have questions about how that will work.

OK?

> Was your thought on headers policy to map to the different response types like OkHttpResponse?

No, that's part of the GRPC payload; there's no way to intercept that.

design/external-authorization-design.md
jpeach: It's possible someone would like more flexibility than that, but I've yet to hear from that person. There's no way to specify this at a finer granularity other than funneling everything through an intermediate proxy that we configure.

design/external-authorization-design.md
jpeach: Probably greater than 0. I expect this to evolve over time.

design/external-authorization-design.md
jpeach: I don't know the answers to that. However, it is a requirement that third parties be able to add their own authorization.

design/external-authorization-design.md
jpeach: This topic is discussed later in the doc, but I suppose we could add it here too.

design/external-authorization-design.md
jpeach: You'd have to use an admission controller (gatekeeper or suchlike) to control this. Generally, I expect that the app owner is responsible for the app and best placed to tweak authorization.

design/external-authorization-design.md
jpeach: It's not TLS configuration, so I didn't put it in that struct.

design/external-authorization-design.md
jpeach: It's a Contour CRD, so Contour updates the status.

design/external-authorization-design.md
jpeach: Does anything else in the HTTPProxy API behaves that way? If there's an established pattern we can try to follow it here, but I don't think that this should be too different from what we already have.

design/external-authorization-design.md
jpeach: I think that adding the special case makes it harder for users to reason about. The use case you present here seems like a reasonable one to me; or at least reasonably possible.

design/external-authorization-design.md
jpeach: It can't be part of the CRD. Imagine you have 50 services and 1 auth service. You want to gradually migrate the 50 services over to the centralized auth. This means that you move the services over a few at a time, flipping the fail open policy as you go. This means that these parameters, and the others in this struct are specific to the client.

design/external-authorization-design.md
jpeach: We could add a `AuthorizationPolicy` to the include and I can see the usefulness of that. But it's more complexity, so I'd rather wait and address it when there's demand. I'm expecting that `AuthorizationPolicy` won't be tweaked that much in practice.

design/external-authorization-design.md
jpeach: We should build whatever is needed to make OIDC+OAuth2 work really well.

jpeach:
I'm pretty sure I addressed all the comments, but I'll leave this until next week to give a change for additional feedback. I added a new "Compatibility" section to make it more obvious how auth will interact with existing features.

design/external-authorization-design.md
stevesloka: I think that use-case is only relevant once. The backend auth I doubt would have different timeouts per vhost. If that was the case you'd have different CRDs to represent the different backends. But given many use the same backend auth, you now need to sync those settings manually across many vhosts.

design/external-authorization-design.md
stevesloka: To be clear from another conversation had via Zoom, this statement isn't true. The intent is that both Contour & 3rd party services should update this field.

jpeach: > To be clear from another conversation had via Zoom, this statement isn't true. The intent is that both Contour & 3rd party services should update this field.

Personally, I would be surprised if that happened 🤷

design/external-authorization-design.md
stevesloka: So this is the same exact use-case of `permitInsecure`. @youngnick I know you had thoughts around this that you don't like the arg today to disable this feature, but I think it needs thought out. 

If admission controller is the answer, we need to scope that work since many folks will want to enforce this from the start.

stevesloka:
> This CRD is relatively generic, so that it can be re-used for Envoy rate limiting and logging services.

If the goal of this CRD is to re-use for other bits other than just auth. How does Contour understand that context? For instance, if we use this for rate limiting and features of the spec aren't valid to be set, how could that be implemented?

design/external-authorization-design.md
jpeach: > So this is the same exact use-case of `permitInsecure`. @youngnick I know you had thoughts around this that you don't like the arg today to disable this feature, but I think it needs thought out.
> 
> If admission controller is the answer, we need to scope that work since many folks will want to enforce this from the start.

I'm happy to have this discussion, but I don't think it's practical to have separate admin control over every field that can be specified in a HTTPProxy resource. General purpose admission controllers exist for this reason, and it's hard for me to imagine that doing the equivalent in Contour is really worth it.

design/external-authorization-design.md
```diff
@@ -0,0 +1,715 @@
...
+
+// ExtensionServiceStatus should follow the pattern being established in [#2642][10].
+// This field will updated by Contour.
+type ExtensionServiceStatus struct {
+	Conditions []Condition `json:"conditions"`
+}
+
+type ExtensionServiceSpec struct {
+	// +required
+	Service contourv1.Service `json:"service"`
```
stevesloka, line 84: Just thinking out loud, should `Service` be an array? If you did that like HTTPProxy does, you would be able to shift traffic to new instances of an auth backend via weights.

jpeach:
> > This CRD is relatively generic, so that it can be re-used for Envoy rate limiting and logging services.
> 
> If the goal of this CRD is to re-use for other bits other than just auth. How does Contour understand that context?

The context can be understood from where the CRD is referenced. In this design, Contour can do the right thing because the CRD is referenced from an authorization policy. In logging and rate limiting, there would be a similar context in the `HTTPProxy` resource.

>  For instance, if we use this for rate limiting and features of the spec aren't valid to be set, how could that be implemented?

Hard to answer this hypothetical, but since the extension service CRD is basically scoped to generate an Envoy cluster, that's sufficient for the rate limiting and logging cases that we know about. Maybe there could be conflicting use cases in the future, but I expect we could address them by revving the CRD, adding another CRD, or in some other way.

design/external-authorization-design.md
stevesloka: Just something to think about. By doing this, you'd have an easy way to not need the route disable bit. You could put the insecure routes in an HTTPProxy chain, then the secure in another.

design/external-authorization-design.md
```diff
@@ -0,0 +1,715 @@
...
+- [#2459][3] External auth background and goals.
+- [#2325][9] Failures in External services configured in envoy should be visible to the application developer.
+
+## Goals
+- Allow operators to integrate existing authorization services with Contour.
+- Allow arbitrary types of authentication and authorization to be supported.
+- Decouple Contour from external authorization, so it can evolve at an independent rate.
+- Integrate cleanly with the HTTPProxy API.
+
+## Non Goals
```
stevesloka, line 20: Also supporting v1/v1beta1.Ingress is not a goal?

stevesloka:
> The context can be understood from where the CRD is referenced. In this design, Contour can do the right thing because the CRD is referenced from an authorization policy. In logging and rate limiting, there would be a similar context in the HTTPProxy resource.

Ahh ok that could work looking at where it's referenced from. =)

design/external-authorization-design.md
jpeach: Yeh, arguably there is a use case for timeout being on the CRD.

design/external-authorization-design.md
jpeach: Yeh, I did consider that, but you can already handle this use case (to some extent) with services and labels. If this was a slice, then weights and mirroring could be implemented, which would make it the same as `HTTPProxy`.

design/external-authorization-design.md
jpeach: Yeh, just a matter of whether we want to do this up front in what will already be a big change.

design/external-authorization-design.md
jpeach: Yep, so far it is not a goal. In general, when adding features, we haven't also added them to Ingress, so not sure if we want to do that here 🤷

youngnick: I think that we will add these features to HTTPProxy first, and consider if we want to add them to Ingress later. The only way we could do that would be with a forest of annotations, which I am reluctant to do without a good reason.

design/external-authorization-design.md
youngnick: To be clear, @stevesloka, I don't like that we have to disable a permit globally, that's all. I agree that sometimes, cluster admins are going to want to say "you must have auth on, you can't turn it off", but that conflicts with the goal of being able to selectively disable some routes, which is more useful to start with, I think.

To put this another way, I think that being able to expose *some* parts of an app without auth is more immediately useful than having a way to disable that functionality.

youngnick: I agree that I think we leave this one and see what our users want. We're not ruling this out by not doing it now.

stevesloka: I just think that admins are going to want to put a limitation on this field. It's a big hole in security if teams can purposely or accidentally disable security per route and will want a way to lock this field down to specific roles in the cluster. Just my 2 cents.

design/external-authorization-design.md
stevesloka: I'm not suggesting that we add, just to call it out in the design so when someone comes and reads the design doc, it's clear. That was one of the original goals of these designs. It was to plan out how or what a feature looked like, even if parts were far off in the future. 

The second goal was to have these features planned out so that more users could contribute with a plan already in place.

design/external-authorization-design.md
stevesloka: > already handle this use case (to some extent) with services and labels

This was the reason HTTPProxy has it as a slice since labels & selectors aren't a great way to effectively implement a rollout.

jpeach:
* added ingress to non-goals
* moved timeout to ExtensionService CRD
* made ExtensionService services field a slice

design/external-authorization-design.md
```diff
@@ -0,0 +1,710 @@
...
+	// +optional
+	TimeoutPolicy *contourv1.TimeoutPolicy `json:"timeoutPolicy,omitempty"`
+
+	// This field sets the version of the GRPC protocol that Envoy uses to
+	// send requests to the support service. Since Contour always uses the
+	// v2 Envoy API, this is currently fixed at "v2". However, other
+	// protocol options will be available in future.
+	//
+	// +kubebuilder:validation:Enum=v2
+	ProtocolVersion SupportProtocolVersion `json:"protocolVersion,omitempty"`
```
tsaarni, line 101: Question: is there timeplan for moving to Envoy API v3?  [Deprecation](https://www.envoyproxy.io/docs/envoy/latest/api/api_supported_versions#api-supported-versions) seems to be scheduled for end of the year.  

Comment: The `transport_api_version` parameters can only be set in [v3](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto#extensions-filters-http-ext-authz-v3-extauthz).  So user could only set the `protocolVersion` field after Contour has moved to v3.  I wonder if development of external authorization could be synchronized with the API version move so that v2 support could be skipped completely, and `protocolVersion` field in the spec could be removed?  If the Envoy v2 deprecation period holds, one could set `protocolVersion` to `v2` for only couple of months anyways.

design/external-authorization-design.md
```diff
@@ -0,0 +1,710 @@
...
+type AuthorizationExtension struct {
+        // ExtensionRef names the authorization service.
+	// +required
+	ExtensionRef ExtensionReference `json:"extensionRef"`
+
+        // AuthPolicy is the default authorization policy applied to client requests.
+	// +optional
+	AuthPolicy *AuthorizationPolicy `json:"authPolicy,omitempty"`
+
+        // If FailOpen is true, the client request is forwarded to the upstream service
```
tsaarni, line 179: Is this controlling the `status_on_error` on Envoy API ([link](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto#extensions-filters-http-ext-authz-v3-extauthz))? 
Maybe document the default behavior here to clarify that HTTP 403 Forbidden will be returned by default if authorization server fails to respond.

design/external-authorization-design.md
```diff
@@ -0,0 +1,710 @@
...
+Contour.
+
+```yaml
+apiVersion: projectcontour.io/v1alpha1
+kind: ExtensionService
+metadata:
+  name: testserver
+  namespace: contour-auth
+spec:
+  service:
```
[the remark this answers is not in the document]
jpeach: Hah, services became a slice at the last minute and I forgot to update the sample YAML.

design/external-authorization-design.md
```diff
@@ -0,0 +1,710 @@
...
+   This increases the complexity of the `HTTPProxy` structure and makes it difficult to re-use the same authorization service acrtoss multiple proxies.
+   A separate CRD gives better opportunities to expose useful operational status.
+   Integrating specific authorization parameters into `HTTPProxy` prevents operators implementing their own authorization flows.
+
+## Security Considerations
+
+- Most HTTP authorization protocols depend on HTTPS to keep headers and URL query parameters private.
+  Supporting only authorization on HTTPS, Contour reduces the changes of misconfiguring this.
+- HTTPS sessions between Envoy and the authorization server are required.
+  TLS validation can be configured if necessary (should be recommended),
```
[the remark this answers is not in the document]
jpeach: Yeh that should be doable eventually.

design/external-authorization-design.md
```diff
@@ -0,0 +1,710 @@
...
+}
+
+type AuthorizationPolicy struct {
+	// When true, this field disables client request authentication
+	// for the scope of the policy.
+	//
+	// +optional
+	Disabled bool `json:"disabled,omitempty"`
+
+	// Context is a set of key/value literal strings that are sent to the
```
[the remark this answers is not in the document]
jpeach: Yeh this is context extensions, will make it clear.

design/external-authorization-design.md
jpeach: > Question: is there timeplan for moving to Envoy API v3?  [Deprecation](https://www.envoyproxy.io/docs/envoy/latest/api/api_supported_versions#api-supported-versions) seems to be scheduled for end of the year.  

There is an issue but it's not being worked on yet.

> 
> Comment: The `transport_api_version` parameters can only be set in [v3](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto#extensions-filters-http-ext-authz-v3-extauthz).  So user could only set the `protocolVersion` field after Contour has moved to v3.  I wonder if development of external authorization could be synchronized with the API version move so that v2 support could be skipped completely, and `protocolVersion` field in the spec could be removed?  If the Envoy v2 deprecation period holds, one could set `protocolVersion` to `v2` for only couple of months anyways.

In principle at least they are independent versions. I am pretty sure I asked htuch and he said they are likely to keep it, but I suppose that could always change. At any rate I'd prefer to not block on v3 API support.

design/external-authorization-design.md
jpeach: Yup, definitely worth make that clear in the docs.

glerchundi:
Thinking out loud but as this design doc i ntroduces the concept of
ExtensionService, have you considered the idea of using WASM based
extension services instead of just those based on gRPC?

It’s problably premature to take even into consideration but I would like
to put it on the table to see if the new spec definitions can absorb them
in the future.

jpeach:
> Thinking out loud but as this design doc i ntroduces the concept of ExtensionService, have you considered the idea of using WASM based extension services instead of just those based on gRPC? It’s problably premature to take even into consideration but I would like to put it on the table to see if the new spec definitions can absorb them in the future.

In principle the extensionRef field in the `HTTPProxy` resource could reference something that is not an `ExtensionService` CRD.

Issue #2369, opened 2020-03-20
Add retriable status codes to HTTPProxy retryPolicy


KevinSnyderCodes:
**Please describe the problem you have**

During the deployment of a service (i.e. a change to a `Deployment` manifest), there is a non-zero amount of time during which Contour registers/deregisters new/old pods with Envoy. For services that receive hundreds of requests per second, this results in a noticeable amount of 5xx responses.

At present `HTTPProxy` offers a `retryPolicy` that will retry all 5xx status codes. We would like to configure the status codes that should be retried (e.g. only `503`) to mitigate the 5xx responses during deployments without retrying all 5xx status codes.

Envoy offers a `connect-failure` option for their [x-envoy-retry-on](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#x-envoy-retry-on) header which would solve our problem. However, this is specific to Envoy and a list of retriable status codes seems less opinionated and a better fit for Contour's design.

Relevant Slack discussion:
https://kubernetes.slack.com/archives/C8XRH2R4J/p1582153240440600
https://kubernetes.slack.com/archives/C8XRH2R4J/p1584729842024600

stevesloka:
We'd need to think about how to implement this. My initial thought was to expand the `retryPolicy` to add an additional field to accept the status codes to retry on, however, in doing so, you'd leave out the ability to define separate policies, meaning you could define a policy when a `503` is found like in your example, but couldn't define for any other 5XX code.

kevincantu:
This is seems like a feature regression from Ingress to HTTPProxy, and the the lack of this ability is potentially blocking my upgrade to a newer version of Contour.  (I'm getting a lot of 504s where we previously had a config to just retry around 'em.)

So I've done a bit more digging.

Ingress configs used to support [annotations](https://projectcontour.io/docs/v1.4.0/annotations/#contour-specific-ingress-annotations) like these:
```
    projectcontour.io/num-retries: "5"
    projectcontour.io/request-timeout: "infinity"
    projectcontour.io/retry-on: "gateway-error"
```

To address @stevesloka's comment, Envoy [supports](https://www.envoyproxy.io/docs/envoy/v1.11.1/configuration/http_filters/router_filter.html#config-http-filters-router-x-envoy-retry-on) a comma-separated list of retry-on policy codes (and Contour's tests [included them](https://github.com/projectcontour/contour/blob/master/internal/featuretests/retrypolicy_test.go#L55)):
[...]

PR #2646, opened 2020-07-01
Add support for retry on, retriable status codes

KevinSnyderCodes:
Design: https://github.com/projectcontour/contour/blob/master/design/retryon-design.md
Design PR: https://github.com/projectcontour/contour/pull/2626
Issue: https://github.com/projectcontour/contour/issues/2369


apis/projectcontour/v1/httpproxy.go
```diff
@@ -314,6 +314,10 @@ type TimeoutPolicy struct {
 	Idle string `json:"idle,omitempty"`
 }
 
+// RetryOn is a string type alias with validation to ensure that the value is valid.
+// +kubebuilder:validation:Enum=5xx;gateway-error;reset;connect-failure;retriable-4xx;refused-stream;retriable-status-codes;retriable-headers
```
KevinSnyderCodes, line 318: Not sure what the right way to test this validation is.

youngnick: Probably verification by hand would be sufficient (that is, run `make generate`, apply to a cluster, then try add add things that don't match this list).

apis/projectcontour/v1/httpproxy.go
```diff
@@ -324,6 +328,12 @@ type RetryPolicy struct {
 	// PerTryTimeout specifies the timeout per retry attempt.
 	// Ignored if NumRetries is not supplied.
 	PerTryTimeout string `json:"perTryTimeout,omitempty"`
+	// RetryOn specifies the conditions on which to retry a request.
+	// +optional
+	RetryOn []RetryOn `json:"retryOn"`
```
youngnick, line 333: This should have `omitempty`, I think.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -324,6 +328,12 @@ type RetryPolicy struct {
...
+	// RetriableStatusCodes specifies the HTTP status codes that should be retried.
+	// +optional
+	RetriableStatusCodes []uint32
```
youngnick, line 336: This should also have `omitempty`, and a json tag.

youngnick:
Also, to get the CI to pass, you will need to run `make generate` and commit the result - the YAMLs are out of date.

apis/projectcontour/v1/httpproxy.go
jpeach: You can also test validation with integration tests, e.g. https://github.com/projectcontour/contour/blob/master/_integration/testsuite/httpproxy/001-required-field-validation.yaml

apis/projectcontour/v1/httpproxy.go
```diff
@@ -324,6 +328,12 @@ type RetryPolicy struct {
 	// PerTryTimeout specifies the timeout per retry attempt.
 	// Ignored if NumRetries is not supplied.
 	PerTryTimeout string `json:"perTryTimeout,omitempty"`
+	// RetryOn specifies the conditions on which to retry a request.
+	// +optional
```
jpeach, line 332: This is generated into API reference documentation, so we should enumerate and document the values (the generator doesn't look at the kubebuilder annotations). This comment an be markdown, so perhaps the balance is to add a list to enumerate the options, then link to Envoy docs for detailed semantics.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -324,6 +328,12 @@ type RetryPolicy struct {
...
+	RetryOn []RetryOn `json:"retryOn,omitempty"`
+	// RetriableStatusCodes specifies the HTTP status codes that should be retried.
+	// +optional
```
jpeach, line 335: This should document when the field is used (i.e. only when you set `RetryOn` to `retriable-status-codes`.

internal/dag/policy.go
```diff
@@ -25,15 +26,28 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation"
 )
 
+func retryOn(ro []projcontour.RetryOn) string {
```
jpeach, line 29: Nit: could you add a comment mentioning that the CRD validation ensures that only valid retry strings are present in this slice?

internal/dag/policy.go
```diff
@@ -25,15 +26,28 @@ import (
...
 	if rp == nil {
 		return nil
 	}
 	perTryTimeout, _ := time.ParseDuration(rp.PerTryTimeout)
 	return &RetryPolicy{
-		RetryOn:       "5xx",
-		NumRetries:    max(1, uint32(rp.NumRetries)),
-		PerTryTimeout: perTryTimeout,
+		RetryOn:              retryOn(rp.RetryOn),
+		RetriableStatusCodes: rp.RetriableStatusCodes,
```
jpeach, line 48: What does Envoy do if you send it status codes that are out of range. e.g. 9000?

apis/projectcontour/v1/httpproxy.go
```diff
@@ -314,6 +314,10 @@ type TimeoutPolicy struct {
 	Idle string `json:"idle,omitempty"`
 }
 
+// RetryOn is a string type alias with validation to ensure that the value is valid.
+// +kubebuilder:validation:Enum="5xx";gateway-error;reset;connect-failure;retriable-4xx;refused-stream;retriable-status-codes;retriable-headers
```
jpeach, line 318: The Envoy protobuf docs say that this accepts both `x-envoy-retry-on` and `x-envoy-retry-grpc-on` values. Should we add the `x-envoy-retry-grpc-on` values to the enum here as well?

internal/dag/policy.go
KevinSnyderCodes: Looks like Envoy will accept any `uint32` as a retriable status code:
https://github.com/envoyproxy/envoy/blob/master/source/common/router/config_impl.cc#L99-L101
And if the response status code matches any retriable status code, regardless of its validity, it will retry it:
https://github.com/envoyproxy/envoy/blob/master/source/common/router/retry_state_impl.cc#L307-L313

If you'd like I can verify this behavior by spinning up Envoy and sending it a payload to see how it behaves.

apis/projectcontour/v1/httpproxy.go
KevinSnyderCodes: Looking at Envoy's code, it appears that retrying gRPC behaves independently of any other configuration:
https://github.com/envoyproxy/envoy/blob/master/source/common/router/config_impl.cc#L74-L75
https://github.com/envoyproxy/envoy/blob/master/source/common/router/retry_state_impl.cc#L323-L342

So yes, this should be harmless to add.

KevinSnyderCodes:
I was unable to get the integration tests working, but I was able to apply an invalid `HTTPProxy` and got an error message:

```yaml
kind: HTTPProxy
metadata:
  name: nginx-deployment
spec:
  virtualhost:
    fqdn: nginx.foo.com
  routes:
    - conditions:
      - prefix: /
      services:
        - name: nginx-deployment
          port: 80
      retryPolicy:
        retryOn:
          - foobar
```

```
The HTTPProxy "nginx-deployment" is invalid: spec.routes.retryPolicy.retryOn: Unsupported value: "foobar": supported values: "5xx", "gateway-error", "reset", "connect-failure", "retriable-4xx", "refused-stream", "retriable-status-codes", "retriable-headers", "cancelled", "deadline-exceeded", "internal", "resource-exhausted", "unavailable"
```

_integration/testsuite/httpproxy/011-retry-policy-validation.yaml
```diff
@@ -0,0 +1,39 @@
+# Copyright 2020 VMware, Inc.
```
skriss, line 1: we've recently switched all existing copyright headers to `// Copyright Project Contour Authors`, please update this one.

kevincantu:
Yay!

PR #2668, opened 2020-07-08
internal/k8s: add check to verify resources exist in cluster

### LINKED ISSUE #2219: Automatically enable detected GVKs

stevesloka:
Add a discovery client to look up server groups and resources which
allows for a check to be made that a resource type exists before
starting any informers against.

Fixes #2219


internal/k8s/syncer.go
```diff
@@ -16,24 +16,36 @@ package k8s
...
+			errors = append(errors, fmt.Errorf("GroupVersion: %s Kind: %s not found in ServerGroupsAndResources", r.GroupVersion().String(), r.Resource))
+			continue
+		}
 		informer := f.ForResource(r).Informer()
 		informer.AddEventHandler(handler)
 
 		sl.syncers = append(sl.syncers, informer.HasSynced)
 	}
+
+	return errors
```
jpeach, line 48: Using errors to communicate non-error status doesn't feel that clean to me ... I had a suggestion for a different approach above. I agree with you that we we should log when we filter out a resource though, and that makes this harder to manage.

cmd/contour/serve.go
```diff
@@ -234,24 +234,39 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 
 	// step 4. register our resource event handler with the k8s informers,
 	// using the SyncList to keep track of what to sync later.
-	var informerSyncList k8s.InformerSyncList
+	informerSyncList := k8s.InformerSyncList{
+		Clients: clients,
+	}
```
jpeach, line 239: I think that adding the errors from `InformOnResources` makes this setup a fair bit harder to read, and it's already pretty complicated.

What do you think about an approach like this:
```

resources, err := clients.ServerGroupsAndResources()
// error check ...

informerSyncList := k8s.InformerSyncList {
    ServerResources: resources,
    UnsupportedResources: []metav1.GroupVersionResource{},
}

// Then create informers, but the sync list will internally filter and track unsupported resources.
informerSyncList.InformOnResources(clusterInformerFactory, dynamicHandler, k8s.ServiceAPIResources()...)

for _, r := range informerSyncList.UnsupportedResources {
    // Assuming the GVR formats into a field properly ...
    log.WithField("resource", r).Info("skipping watch on unsupported resource")
}
```

jpeach: IIUC, the main use case for checking the server resources is to automatically enable the Ingress v1 APIs. In that case, maybe a pattern more like this could work:

```Go
serverResources, err := clients.ServerGroupsAndResources()

if serverResources.Supports(k8s.IngressV1Resources()) {
    // Do the stuff ...
}
```

In this approach `Supports()` would return true if all the GVRs in the list were present on the server.

cmd/contour/serve.go
```diff
@@ -234,24 +234,39 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
 
-	informerSyncList.InformOnResources(clusterInformerFactory, dynamicHandler, k8s.DefaultResources()...)
+	apiResourceList, err := clients.ServerGroupsAndResources()
+	if err != nil {
+		log.WithField("context", "ServerGroupsAndResources").Errorf("could not get list of API Resources: %v", err)
+		return err
+	}
```
jpeach, line 245: Since we only pull the API resources once, do you think should document somewhere that Contour will need to be restarted after installing CRDs?

internal/k8s/clients.go
```diff
@@ -84,3 +95,26 @@ func (c *Clients) ClientSet() *kubernetes.Clientset {
...
+}
+
+// ResourceKindExists returns true if an APIResource.Kind exists in the cluster.
+// ex: groupVersion=projectcontour.io/v1 kind=httpproxies
+func (c *Clients) ResourceKindExists(groupVersion, kind string, apiResourceList []*metav1.APIResourceList) (bool, error) {
+
+	for _, r := range apiResourceList {
+		if strings.EqualFold(r.GroupVersion, groupVersion) {
+			for _, list := range r.APIResources {
+				if strings.EqualFold(list.Name, kind) {
```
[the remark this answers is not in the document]
stevesloka: `Kind` is the singular version, `Name` is the plural. When we setup the informers we look for the plural which is why I used `Name`. I agree it seems like the wrong field but matches properly.

internal/k8s/clients.go
```diff
@@ -84,3 +95,26 @@ func (c *Clients) ClientSet() *kubernetes.Clientset {
...
+	_, resources, err := c.discovery.ServerGroupsAndResources()
+	return resources, err
+}
+
+// ResourceKindExists returns true if an APIResource.Kind exists in the cluster.
+// ex: groupVersion=projectcontour.io/v1 kind=httpproxies
+func (c *Clients) ResourceKindExists(groupVersion, kind string, apiResourceList []*metav1.APIResourceList) (bool, error) {
+
+	for _, r := range apiResourceList {
+		if strings.EqualFold(r.GroupVersion, groupVersion) {
```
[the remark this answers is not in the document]
stevesloka: Not fully, just wanted to make sure we took that issue out of the equation since we're doing a string comparison.

internal/k8s/clients.go
```diff
@@ -84,3 +95,26 @@ func (c *Clients) ClientSet() *kubernetes.Clientset {
...
+// ServerGroupsAndResources returns the set of API Resource object that exist in the
+// Kubernetes cluster
+func (c *Clients) ServerGroupsAndResources() ([]*metav1.APIResourceList, error) {
+	_, resources, err := c.discovery.ServerGroupsAndResources()
+	return resources, err
+}
+
+// ResourceKindExists returns true if an APIResource.Kind exists in the cluster.
+// ex: groupVersion=projectcontour.io/v1 kind=httpproxies
+func (c *Clients) ResourceKindExists(groupVersion, kind string, apiResourceList []*metav1.APIResourceList) (bool, error) {
```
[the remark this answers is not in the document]
stevesloka: Ahh good point! I did a last minute refactor and moved the `err` logic to a different spot.

internal/k8s/syncer.go
stevesloka: Yeah I didn't want to plumb in all the extra meta data to properly log and thought the caller should handle the logging. I'll look at your other suggestion.

cmd/contour/serve.go
stevesloka: Returning error here should cause Contour to crash and be restarted. But yes I agree docs in any case should be good to add as well since there possibly could be a race condition from the time you apply the CRDs and the time the API server is ready to serve them.

stevesloka:
@jpeach I just refactored this a bit. I need to do another internal pass, but I've got to run for a bit so thought maybe it would help to push to show the newer direction. 

I ended up just applying this to the serviceapis bits, I think maybe adding checks to the other apis isn't all that necessary.

internal/k8s/clients.go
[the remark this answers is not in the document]
jpeach: IIUC, `Name` is the resource name and `Kind` is the kind name, and while the resource name is conventionally just the pluralized kind, that's not enforced. I've always found this pretty confusing so I try to be struct about terminology now :)

Now you've plumbed the GVR through the API, that removes any possible ambiguity 👍

jpeach:
> @jpeach I just refactored this a bit. I need to do another internal pass, but I've got to run for a bit so thought maybe it would help to push to show the newer direction.

Yeh, this approach looks promising. I think that if you make `map[schema.GroupVersionResource]struct{}` a type, it will clean up nicely.

> I ended up just applying this to the serviceapis bits, I think maybe adding checks to the other apis isn't all that necessary.

That makes sense to me. I think that it will be hard for Contour to gracefully handle it when core resources are missing :)

stevesloka:
@jpeach this is ready for another look. Thanks!

internal/k8s/clients.go
```diff
@@ -84,3 +103,31 @@ func (c *Clients) ClientSet() *kubernetes.Clientset {
...
+// serverResources returns the list of all the resources supported
+// by the API server. Note that this method guarantees to populate the
+// Group and Version fields in the result.
+func (c *Clients) serverResources() (*APIResources, error) {
+	_, apiList, err := c.discovery.ServerGroupsAndResources()
+	if err != nil {
+		return nil, err
+	}
+
+	gvrs, err := discovery.GroupVersionResources(apiList)
```
jpeach, line 121: Good find!

internal/k8s/clients.go
```diff
@@ -84,3 +103,31 @@ func (c *Clients) ClientSet() *kubernetes.Clientset {
 func (c *Clients) DynamicClient() dynamic.Interface {
 	return c.dynamic
 }
+
+// APIResources holds a map of API, GroupVersionResources, that exist in the Kubernetes cluster
+type APIResources struct {
```
jpeach, line 108: I think that this types makes sense.

Looking at `DefaultResources()`, `ServiceAPIResources()`, etc, we could (in a followup PR) refactor those to return `APIResources`, then we have a consistent type for all this stuff.

cmd/contour/serve.go
```diff
@@ -241,8 +241,14 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
 	if ctx.UseExperimentalServiceAPITypes {
-		informerSyncList.InformOnResources(clusterInformerFactory,
-			dynamicHandler, k8s.ServiceAPIResources()...)
+		for _, s := range k8s.ServiceAPIResources() {
+			// Check if the resource exists in the API server before setting up the informer.
+			if !clients.ResourceExists(s) {
+				log.WithField("InformOnResources", "ExperimentalServiceAPITypes").Warnf("resource %v not found in api server", s.String())
+				continue
+			}
+			informerSyncList.InformOnResources(clusterInformerFactory, dynamicHandler, s)
```
jpeach, line 250: Does it make sense to try to keep going if some resources are missing? I think that it would be better (or at least more predictable) to make it all or nothing.

I suggest something like the following:
```Go
apiResources, err := clients.ServerResources()
if err != nil {
        log.Fatal(err) // Not being able to query API server resources is totally broken.
}

if ctx.UseExperimentalServiceAPITypes) {
        // ".Contains()" would return true if all the arguments are present in the set of resources.
        if apiResources.Contains(k8s.ServiceAPIResources()...) {
                informerSyncList.InformOnResources(...)
        } else {
                log.Warnf("APIs not present)
        }
}
```

This approach lets the caller determine how to handle any errors (I think we mostly would want to fail hard on this). It also avoids any awkward questions about revalidating the cache copy in the client since we can also push that onto the caller.

WDYT?

stevesloka: Sure thing, I pushed an update.

jpeach: I approved, but I still think that we should avoid caching the resources in the client.

PR #2674, opened 2020-07-09
internal/dag: improve validation status for invalid Secrets

### LINKED ISSUE #2616: Misleading Proxy status for secrets type opaque
If a user creates a secret type `opaque` and references that via a Proxy, the resulting status will say the proxy is invalid because the secret does not exist when it actually does. 

The problem is Contour assumes any tls secret to be type `kubernetes.io/tls`, so when it finds a secret type `Opaque` it assumes it should only contain a `ca.crt` (https://github.com/projectcontour/contour/blob/master/internal/dag/secret.go#L57). When it doesn't, then the secret is not added to the internal k8s cache and further lookups cannot find it which results in the current error. 

This is misleading to users since the secret does exist, just not in the format that Contour would like.

jpeach:
Secrets validation is performed before a Secrets is accepted into
the cache. This means that if you make an error in a Secret, Contour
typically describes that as "secret not found", and you have to dig
in logs to figure out the real problem.

This change accepts otherwise invalid Secrets into the cache, but
tracks them as invalid. When another resource uses an invalid Secret,
the validation error can be regurgitated and presented in status.

This fixes #2616.

### DIFF
--- a/internal/annotation/annotations.go
+++ b/internal/annotation/annotations.go
@@ -14,6 +14,7 @@
 package annotation
 
 import (
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
@@ -276,4 +277,12 @@ func MaxRequests(o metav1.ObjectMetaAccessor) uint32 {
 // '0' is returned if the annotation is absent or unparseable.
 func MaxRetries(o metav1.ObjectMetaAccessor) uint32 {
 	return parseUInt32(CompatAnnotation(o, "max-retries"))
+}
+
+func Tombstone(o metav1.ObjectMetaAccessor) error {
+	if s := CompatAnnotation(o, "tombstone"); s != "" {
+		return errors.New(s)
+	}
+
+	return nil
 }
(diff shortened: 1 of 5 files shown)


jpeach:
I think the tombstoning is a bit weird :-/

/cc @stevesloka

stevesloka:
> I think the tombstoning is a bit weird :-/

Yup I agree. Looking at a similar problem previously I tried creating a new struct to hold the object and some status (maybe valid, invalid, etc). I think I ended up with just re-evaluating the `isValid()` bit each time it was used which seemed safer in the end.

jpeach:
> > I think the tombstoning is a bit weird :-/
> 
> Yup I agree. Looking at a similar problem previously I tried creating a new struct to hold the object and some status (maybe valid, invalid, etc). I think I ended up with just re-evaluating the `isValid()` bit each time it was used which seemed safer in the end.

Even though the tombstoning is weird, I think that just storing the the extra secrets is worse
1. Caching these secrets unnecessarily increases attack surface
1. Caching costs us in memory
1. Revalidating costs us in performance (reprising PEM all the time)

youngnick:
@jpeach is this going to need substantial reworking as we abstract away various cache things? Should this be closed out?

jpeach:
> @jpeach is this going to need substantial reworking as we abstract away various cache things? Should this be closed out?

If we move to controller-runtime caching we won't be able to take this approach.

PR #2675, opened 2020-07-09
cmd/contour: add config file option to set drain timeout

skriss:

Updates #2225 

I'm open to alternate names if anyone thinks there's a better option here than reusing the Envoy name.


skriss:
@stevesloka this is ready for review when you get a chance, thanks!

cmd/contour/servecontext.go
```diff
@@ -188,6 +188,9 @@ func newServeContext() *serveContext {
 
 			// This is the Envoy default.
 			StreamIdleTimeout: 5 * time.Minute,
+
+			// This is the Envoy default.
+			DrainTimeout: 5 * time.Second,
```
stevesloka, line 193: This is a separate issue, but I just thought about Contour re-using the defaults in Envoy. We should almost not set the same value here in Contour. My worry is that the default changes upstream (for whatever reason) and we still use a different value.

skriss: Yeah, it's a valid concern. We could make these fields pointers, so if there's no Contour-specified default and the user doesn't explicitly specify them in the config file, they're nil, in which case we'd not configure them in Envoy and let the Envoy defaults apply. We do still need to retain 0 as a valid config value since that usually disables the timeout in question completely.

skriss: I've been able to address this through some incoming work (#2698 and another upcoming PR), so we no longer need to copy the Envoy defaults - if we want to use them, we just don't configure a value. We only set a default here if it's a Contour-specific default

PR #2676, opened 2020-07-09
cmd/contour: improve docs for timeout settings

skriss:

updates #2225

### DIFF
--- a/cmd/contour/servecontext.go
+++ b/cmd/contour/servecontext.go
@@ -244,19 +244,28 @@ type LeaderElectionConfig struct {
 
 // TimeoutConfig holds various configurable proxy timeout values.
 type TimeoutConfig struct {
-	// ConnectionIdleTimeout defines how long the proxy should wait while there
-	// are no active requests before terminating an HTTP connection. Set
-	// to 0 to disable the timeout.
+	// ConnectionIdleTimeout defines how long the proxy should wait while there are
+	// no active requests (for HTTP/1.1) or streams (for HTTP/2) before terminating
+	// an HTTP connection. Set to 0 to disable the timeout.
+	//
+	// See https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/core/protocol.proto#envoy-api-field-core-httpprotocoloptions-idle-timeout
+	// for more information.
 	ConnectionIdleTimeout time.Duration `yaml:"connection-idle-timeout,omitempty"`
 
-	// StreamIdleTimeout defines how long the proxy should wait while there
-	// is no stream activity before terminating a stream. Set to 0 to
-	// disable the timeout.
+	// StreamIdleTimeout defines how long the proxy should wait while there is no
+	// request activity (for HTTP/1.1) or stream activity (for HTTP/2) before
+	// terminating the HTTP request or stream. Set to 0 to disable the timeout.
+	//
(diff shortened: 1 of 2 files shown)


cmd/contour/servecontext.go
```diff
@@ -241,14 +241,14 @@ type LeaderElectionConfig struct {
 
 // TimeoutConfig holds various configurable proxy timeout values.
 type TimeoutConfig struct {
-	// ConnectionIdleTimeout defines how long the proxy should wait while there
-	// are no active requests before terminating an HTTP connection. Set
-	// to 0 to disable the timeout.
+	// ConnectionIdleTimeout defines how long the proxy should wait while there are
```
stevesloka, line 244: Is there anything you could add to help describe the difference between Connection & Stream?

site/docs/master/configuration.md
```diff
@@ -75,8 +75,8 @@ The timeout configuration block can be used to configure various timeouts for th
 
 | Field Name | Type| Default  | Description |
 |------------|-----|----------|-------------|
-| connection-idle-timeout| [duration][4] | `60s` | This field defines how long the proxy should wait while there are no active requests before terminating an HTTP connection. Set to 0 to disable the timeout. |
-| stream-idle-timeout| [duration][4] | `5m` | This field defines how long the proxy should wait while there is no stream activity before terminating a stream. Set to 0 to disable the timeout. |
+| connection-idle-timeout| [duration][4] | `60s` | This field defines how long the proxy should wait while there are no active requests (for HTTP/1.1) or streams (for HTTP/2) before terminating an HTTP connection. Set to 0 to disable the timeout. |
```
stevesloka, line 78: Maybe even link back to the Envoy docs?

Another set of questions I've heard is what do these parameters apply to, http, tcp, websockets, etc. Would be super nice to clarify that as well.

stevesloka:
Also, do you want to add https://github.com/projectcontour/contour/pull/2675 to this PR before merging?

site/docs/master/configuration.md
skriss: I added links, and I think the descriptions are now all pretty clear on these being HTTP timeouts, but let me know if not.

We could consider prefixing the actual field names with `http-` to be super-clear if that's something others think makes sense. Since HTTP is the main use case, I opted not to include it on these fields.

cmd/contour/servecontext.go
skriss: I'm not sure I could describe it succinctly, but I could probably find a link to add that explains the concepts in more detail -- maybe this? https://developers.google.com/web/fundamentals/performance/http2#streams_messages_and_frames

skriss: @stevesloka do you want me to add external links to the HTTP/2 spec (https://tools.ietf.org/html/rfc7540#section-2.2) to define these terms? I don't want to clutter with too many links now that I've added the Envoy ones, but let me know what you think here.

stevesloka: I guess I was thinking of ways to describe without going out to links to find out. Mostly for the user-facing docs this applies the most to. 

But for instance ConnectionIdleTimeout is the time for `no active requests` vs StreamIdleTimeout is `no request activity`. It's difficult to explain, but that was my thought. It is good for now probably to get it all merged in.

skriss: @stevesloka I'm just worried that to effectively define these terms and differentiate between them for someone with no prior knowledge of the protocols, it's going to take a couple of paragraphs, which I'm not sure is desirable here. WDYT?

youngnick: I think for now it's better to get *an* improvement merged, and then iterate on it afterwards.

I think it's fine to have links out to other documentation as well.

PR #2677, opened 2020-07-09
internal/dag: return precise error for failed service lookup

### LINKED ISSUE #2641: Improve service lookup error message.

When `(*Builder)lookupService()` fails, the proxy is updated with a generic error status. However, the failure can be because the service isn't present, or because the wanted port could not be found. Emitting a more specific message can distinguish between these cases and make debugging quicker and easier.

skriss:
When a service with the given name+port can't be found during a DAG
build, returns an error indicating whether the service wasn't present
or the port couldn't be found.

closes #2641

### DIFF
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -82,8 +82,9 @@ func (b *Builder) reset() {
 	b.statuses = make(map[k8s.FullName]Status, len(b.statuses))
 }
 
-// lookupService returns a Service that matches the Meta and Port of the Kubernetes' Service.
-func (b *Builder) lookupService(m k8s.FullName, port intstr.IntOrString) *Service {
+// lookupService returns a Service that matches the Meta and Port of the Kubernetes' Service,
+// or an error if the service or port can't be located.
+func (b *Builder) lookupService(m k8s.FullName, port intstr.IntOrString) (*Service, error) {
 	lookup := func() *Service {
 		if port.Type != intstr.Int {
 			// can't handle, give up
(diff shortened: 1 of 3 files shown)


internal/dag/builder.go
```diff
@@ -99,22 +100,22 @@ func (b *Builder) lookupService(m k8s.FullName, port intstr.IntOrString) *Servic
 
 	s := lookup()
 	if s != nil {
-		return s
+		return s, nil
 	}
 	svc, ok := b.Source.services[m]
 	if !ok {
-		return nil
+		return nil, errors.New("service not found")
```
jpeach, line 107: This will result in some stutter in the error message, like "service foo error: service not found". That's not so bad, but maybe it's ok to just return "not found" here? Alternatively, we could move the quoted service name to this error?

internal/dag/builder.go
```diff
@@ -989,9 +989,9 @@ func (b *Builder) processHTTPProxyTCPProxy(sw *ObjectStatusWriter, httpproxy *pr
 		var proxy TCPProxy
 		for _, service := range httpproxy.Spec.TCPProxy.Services {
 			m := k8s.FullName{Name: service.Name, Namespace: httpproxy.Namespace}
-			s := b.lookupService(m, intstr.FromInt(service.Port))
-			if s == nil {
-				sw.SetInvalid("tcpproxy: service %s/%s/%d: not found", httpproxy.Namespace, service.Name, service.Port)
+			s, err := b.lookupService(m, intstr.FromInt(service.Port))
+			if err != nil {
+				sw.SetInvalid("tcpproxy: service %s/%s/%d: %s", httpproxy.Namespace, service.Name, service.Port, err)
```
jpeach, line 994: Can we take the opportunity to align the status error formatting? Current best practice seems to be starting with the field name and quoting any resource names

So maybe it could be something like this:
```Go
sw.SetInvalid("Spec.TCPProxy unresolved service reference: %s", err)
```

And something equivalent for the other lookup

skriss:
@jpeach made some updates that I think cover your comments, please take a look!

internal/dag/builder.go
```diff
@@ -99,22 +100,22 @@ func (b *Builder) lookupService(m k8s.FullName, port intstr.IntOrString) *Servic
 
 	s := lookup()
 	if s != nil {
-		return s
+		return s, nil
 	}
 	svc, ok := b.Source.services[m]
 	if !ok {
-		return nil
+		return nil, fmt.Errorf("service %q not found", m.Name)
```
jpeach, line 107:
```suggestion
		return nil, fmt.Errorf("service %q not found", m)
```

Let's print the fully qualified name to remove any possible ambiguity.

internal/dag/builder.go
```diff
@@ -99,22 +100,22 @@ func (b *Builder) lookupService(m k8s.FullName, port intstr.IntOrString) *Servic
...
 		case int(p.Port) == port.IntValue():
-			return b.addService(svc, p)
+			return b.addService(svc, p), nil
 		case port.String() == p.Name:
-			return b.addService(svc, p)
+			return b.addService(svc, p), nil
 		}
 	}
-	return nil
+	return nil, fmt.Errorf("port %q on service %q not matched", port.String(), m.Name)
```
jpeach, line 118:
```suggestion
	return nil, fmt.Errorf("port %q on service %q not matched", port.String(), m)
```

PR #2684, opened 2020-07-14
tcpproxy: revert 'services' to be optional again

### LINKED ISSUE #2681: tcpproxy: services should be optional to support clean delegation
We tried to upgrade from 1.1 to 1.6 following the upgrading tips on the webpage
with no success. We already know where the problem is but I wanted to double
check if it's something done on purpose or if it's a bug.

One of our `HTTPProxy` definitions, which was working before the upgrade, is
erroring out with the following message:

```
The HTTPProxy "tunnel" is invalid: spec.tcpproxy.services: Required value
```

The problem seems to be introduced in 1076f5a156bbb0a9e804f63296b94faa147bbe7f
(👋 @davecheney as he was the one who made the change) where `services` was made
a required property with a minimum items of 1, even if the idea of that
definition was to be delegated into another namespace.

IMHO, this is somehow inconsistent with how you designed the `routes` because
they can colive with `includes` as both are optional. As the `include` for the
`tcpproxy` belongs to the entity, the `services` should be made optional and add
a validity check programatically to allow one (`services`) or the other
(`include`) (in case there isn't a conditional solution with `kubebuilder`
tags).

WDYT?

**Environment:**

- Contour version: Upgrading 1.1 -> 1.6
- Kubernetes version: (use `kubectl version`): 1.15

glerchundi:
Fixes #2681


glerchundi:
Thanks for the reviews 🙏

PR #2686, opened 2020-07-14
internal/featuretests: add tests for timeouts

skriss:

@stevesloka is this what you had in mind?


skriss:
Looking at the failing tests

PR #2689, opened 2020-07-15
apis/projectcontour: rename the HTTPProxy Status type

jpeach:
Rename the `Status` struct to `HTTPProxyStatus`. This means that the
type name won't conflict when we bring new CRDs into the `v1` namespace,
and makes the nomenclature consistent with established CRD patterns.

This is a no-op for API users, since the field name doesn't change,
but is an incompatible source change for projects using the Go types.


stevesloka:
@jpeach just need to commit the generated files back to pass CI.

PR #2695, opened 2020-07-17
Rename HTTPProxy match Conditions to MatchConditions for disambiguation

youngnick:
This is preparation to add a copy of the upstream Condition type to
the projectcontour.io/v1 api group.

Preparing to add Conditions (for status)  made me realize that we needed to
change the terminology everywhere in Contour (except the `conditions` yaml tag),
or else we could get very confused about Conditions (for routing and including)
versus Conditions (for status).


apis/projectcontour/v1/httpproxy.go
```diff
@@ -42,27 +42,33 @@ type Include struct {
 	// Namespace of the HTTPProxy to include. Defaults to the current namespace if not supplied.
 	// +optional
 	Namespace string `json:"namespace,omitempty"`
-	// Conditions are a set of routing properties that is applied to an HTTPProxy in a namespace.
-	// +optional
-	Conditions []Condition `json:"conditions,omitempty"`
+	// MatchConditions are a set of rules that are applied to included HTTPProxies
```
jpeach, line 45: End sentence with a `.`

apis/projectcontour/v1/httpproxy.go
```diff
@@ -139,9 +145,13 @@ type TLS struct {
...
 type Route struct {
-	// Conditions are a set of routing properties that is applied to an HTTPProxy in a namespace.
+	// MatchConditions are a set of rules that are applied to a Route.
+	// When applied, they are merged using AND, with one exception:
+	// There can be only one Prefix MatchCondition per MatchConditions slice.
+	// More than one Prefix, or contradictory MatchConditions, will make the
+	// route invalid.
 	// +optional
-	Conditions []Condition `json:"conditions,omitempty"`
+	MatchConditions []MatchCondition `json:"conditions,omitempty"`
```
jpeach, line 154: Should keep this field name as `Conditions`.

internal/dag/conditions.go
```diff
@@ -22,11 +22,11 @@ import (
...
 )
 
-// mergePathConditions merges the given slice of prefix Conditions into a single
+// mergePathMatchConditions merges the given slice of prefix MatchConditions into a single
 // prefix Condition.
-// pathConditionsValid guarantees that if a prefix is present, it will start with a
+// pathMatchConditionsValid guarantees that if a prefix is present, it will start with a
 // / character, so we can simply concatenate.
-func mergePathConditions(conds []projcontour.Condition) Condition {
+func mergePathMatchConditions(conds []projcontour.MatchCondition) MatchCondition {
```
stevesloka, line 29: Should this have been called `mergePrefixConditions` from the start? So now it would be `mergePrefixMatchConditions`?

Just a little confusing if we do add true "path" matching later on.

youngnick: Probably, yeah.

youngnick:
About the match prefix on conditions for include - since we use the same struct in both places, we're kind of stuck. I think it's better to make it clear that if you add support for a condition in one place, you add it in the other as well.

internal/dag/conditions.go
youngnick: Actually, I looked into this, and I think it's better to leave it the same, and extend this function with additional path matching merge logic if we add non-prefix matching later.

youngnick: If we're going to do that, I'd rather it be in a separate PR, because we'll need to rename a bunch of other stuff from Path to Prefix as well.

PR #2701, opened 2020-07-20
hack: Update  make-release-tag script to update certgen-job

stevesloka:
Updates the `hack/make-release-tag.sh` script to change the job name to match
the release being created.

### DIFF
--- a/hack/release/make-release-tag.sh
+++ b/hack/release/make-release-tag.sh
@@ -38,14 +38,21 @@ run::sed() {
     esac
 }
 
-# NOTE(jpeach): this will go away or change once we move to kustomize
-# since at that point the versioned image name will appear exactly once.
-for example in examples/contour/03-envoy.yaml examples/contour/03-contour.yaml ; do
+# Update the image tags in the Contour & Envoy manifests to the new version.
+for example in examples/contour/03-envoy.yaml examples/contour/03-contour.yaml examples/contour/02-job-certgen.yaml ; do
     # The version might be master or OLDVERS depending on whether we are
     # tagging from the release branch or from master.
     run::sed \
         "-es|docker.io/projectcontour/contour:master|$IMG|" \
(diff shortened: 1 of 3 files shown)


hack/release/make-release-tag.sh
```diff
@@ -49,6 +48,13 @@ for example in examples/contour/03-envoy.yaml examples/contour/03-contour.yaml ;
         "$example"
 done
 
+# Update the certgen job name to the new version.
+for example in examples/contour/02-job-certgen.yaml ; do
+    run::sed \
+        "-es|contour-certgen-master|contour-certgen-$NEWVERS|" \
```
skriss, line 54: I'm thinking we should also look for `$OLDVER`, like https://github.com/projectcontour/contour/pull/2701/files#diff-69f6aa2773b15b9355c06df3155fd0afR47, for consistency - probably relevant for patch releases being created off an existing release branch?

stevesloka: I guess I don't understand why we have $OLDVER in the script. If you made a patch from say v1.5.0 --> v1.5.1, you'd still want $NEWVER (e.g. v1.5.1) as the tag right?

skriss: right, but by then the release-1.5 branch would already have `v1.5.0` in its YAML, so you'd need to sed on that, rather than `master` - that's my understanding anyway.

stevesloka: Ahh ok that makes sense. Let me check on my script, it won't work like that.

skriss: I think you just want:

```
run::sed \
        "-es|contour-certgen-master|contour-certgen-$NEWVERS|" \
        "-es|contour-certgen-$OLDVERS|contour-certgen-$NEWVERS|" \
        "$example"
```

stevesloka: ok updated @skriss

PR #2708, opened 2020-07-21
site: clarify cert-manager guide

### LINKED ISSUE #2692: "Making cert-manager work with HTTPProxy" docs talk about unknown ingress-shim
The Making cert-manager work with HTTPProxy section of the [cert-manager guide](https://projectcontour.io/guides/cert-manager) mentions an `ingress-shim` but it is not clear what `ingress-shim` is. 

> The dummy Ingress record exists only to hold the annotations ingress-shim requires. Because it does not include the ingress-class: contour annotation, Contour will not see it and so the configuration of the Ingress does not matter, except that it is valid enough for ingress-shim to use.

skriss:
closes #2692 

Edits the cert-manager guide to remove references to ingress-shim,
which is an internal component of cert-manager, knowledge of which
is unnecessary for a Contour user. Also adds a link to the relevant
cert-manager docs.

### DIFF
--- a/site/_guides/cert-manager.md
+++ b/site/_guides/cert-manager.md
@@ -603,8 +603,9 @@ $ curl https://httpbinproxy.davecheney.com/get
 
 This method is a workaround until we can deliver the changes in [#950][10] and [#951][11].
 
-The dummy Ingress record exists only to hold the annotations `ingress-shim` requires.
-Because it does not include the `ingress-class: contour` annotation, Contour will not see it and so the configuration of the Ingress does not matter, except that it is valid enough for ingress-shim to use.
+The dummy Ingress record exists only to hold the annotations that cert-manager requires in order to kick off the certification creation process.
+Because it does not include the `ingress-class: contour` annotation, Contour will not see it and so the configuration of the Ingress does not matter, except that it is valid enough for cert-manager to use.
+See the [cert-manager docs][12] for more information.
 
 ## Wrapping up
 
@@ -657,3 +658,4 @@ $ curl -v http://httpbin.davecheney.com/get
 [9]: {% link /img/cert-manager/httpbin.png %}
 [10]: {{site.github.repository_url}}/issues/950
 [11]: {{site.github.repository_url}}/issues/951
+[12]: https://cert-manager.io/docs/usage/ingress/


site/_guides/cert-manager.md
```diff
@@ -603,8 +603,9 @@ $ curl https://httpbinproxy.davecheney.com/get
 
 This method is a workaround until we can deliver the changes in [#950][10] and [#951][11].
 
-The dummy Ingress record exists only to hold the annotations `ingress-shim` requires.
-Because it does not include the `ingress-class: contour` annotation, Contour will not see it and so the configuration of the Ingress does not matter, except that it is valid enough for ingress-shim to use.
+The dummy Ingress record exists only to hold the annotations that cert-manager requires in order to kick off the certification creation process.
+Because it does not include the `ingress-class: contour` annotation, Contour will not see it and so the configuration of the Ingress does not matter, except that it is valid enough for cert-manager to use.
```
stevesloka, line 607: > Because it does not include the `ingress-class: contour` annotation, Contour will not see it...

I know this isn't part of this change directly, but I don't understand this statement that Contour won't see the resource. Contour will "see" this resource and use it to make the HTTP01 challenge work to generate the certificate unless a specific ingress class is configured for Contour.

Also the annotation referenced earlier is: `kubernetes.io/ingress.class: contour` which I think is what the `ingress-class: contour` statement is referring to.

skriss: OK - I can try to dig into what's wrong with that part of the sentence too but I'll need to spend some more time understanding this whole process first. Do you want to accept this fix as-is, or hold it until the other thing is addressed too?

stevesloka: This is fine as-is probably. I still feel the Cert way is much simpler than this dummy ingress resource. So maybe you can take a look at this method (using ingress) and the other way (using Certificate CRD) and help decide what we should have in this guide, possibly both or two guides.

skriss: Yeah I can definitely do that, will probably defer for a bit until I wrap up all the timeout-related changes. I thought this issue was just an easy fix, but nope :)

stevesloka: I create #2709 to follow this change.

Issue #2697, opened 2020-07-17
Reconcile timeout syntax between config file and HTTPProxy API


skriss:
There are some inconsistencies in the syntax for timeouts between the `contour serve` config file and the HTTPProxy API.

In the HTTPProxy API:
- `infinity` is a valid timeout value and results in a `0s` being explicitly passed to Envoy (which Envoy generally interprets as disabling the timeout in question)
- `0` is interpreted as if the timeout were not specified, and results in no value being passed to Envoy, so Envoy's default behavior applies
- an unparseable string is treated the same as "infinity"

In the config file:
- `infinity` is *not* a valid timeout value since we're YAML parsing into `time.Duration` fields.
- `0s` is interpreted as "disable this timeout", so *usually* results in `0s` being explicitly passed to Envoy to disable the timeout (`max_connection_duration` is an exception to this, though, since it needs to be omitted to be disabled).
- if the field is not specified, the Contour defaults apply, which may or may not be the same as the Envoy defaults
    - there is no true fallback to Envoy default behavior; however, per parent bullet, for some timeout fields Contour explicitly passes the current Envoy default as the timeout value if it wasn't specified in the config file (stream_idle, drain).
- an unparseable string is treated as if the field were not specified

I think for UX consistency, we should adopt the current HTTPProxy syntax in both places, so:
- change the config file timeout fields to strings
- allow `infinity`, which means "disable this timeout"
- not specifying the field at all falls back to either a Contour default (e.g. connection_idle), or the Envoy default behavior (others)
- `0` would also be interpreted as if the field were not specified
- an unparseable string is treated the same as "infinity"
    - I have a weak preference for changing this behavior, so it uses the Contour/Envoy default instead, but we can do that separately if we want to

----

`drain_timeout` is weird, because `infinity` doesn't make sense for disabling it -- rather, setting it to 0 makes sense, since it's really a grace period, not a timeout (specifically, it's the time between sending a "warning" GOAWAY frame and sending a final GOAWAY frame).

Perhaps our magic string should be "disabled" instead of "infinity"?

youngnick:
I think that adding `disabled` is probably good, it's certainly a clearer way to indicate that the timeout is disabled. I love the idea of making things consistent everywhere though, and the PR.

jpeach:
> change the config file timeout fields to strings

👍 

> allow infinity, which means "disable this timeout"

[...]

Issue #2690, opened 2020-07-15
move serveContext.RequestTimeout into TimeoutConfig and drop old field


skriss:
Spun out from #2225:

> Once this work is complete, we will move the current `request-timeout` value in the configuration file into this stanza, and add `request-timeout-max` and `request-timeout-min` optional parameters that specify the bounds in which HTTPProxy or other objects may configure the timeout. This will be the pattern from now for all timeouts that have Contour-tunable defaults with a HTTPProxy override. This will be done under a separate issue.

skriss:
Note that per https://github.com/projectcontour/contour/issues/2697#issuecomment-662772957, the new field will use the standard timeout syntax, which is slightly different from the existing syntax for RequestTimeout. Make sure to call this out in docs.

jpeach:
Did anyone investigate using an admission controller (e.g. [gatekeeper](https://github.com/open-policy-agent/gatekeeper)) to enforce this policy? If we add policy controls for every field, configuration will get rapidly out of hand. It's also way better to give policy errors before a resource enters the cluster if possible.

skriss:
> Did anyone investigate using an admission controller (e.g. gatekeeper) to enforce this policy? If we add policy controls for every field, configuration will get rapidly out of hand. It's also way better to give policy errors before a resource enters the cluster if possible.

I did not; I was going off of @youngnick's spec. I can play around with gatekeeper if we're interested.

stevesloka:
I think you probably still need the validation checks in Contour even with an admission controller. Folks would have to buy into that configuration and it sort of feels like client-side validation where you'd still want the check on the backend.

skriss:
There are also concerns that an operator who adds new limits has no built-in way of ensuring that they're not going to break existing proxies. It sounds like we need to take a step back here and think more holistically about the design for adding these types of controls. I'm going to close out the associated PR until we get to a consensus on the approach.

youngnick:
Let's talk about where we go from here and come back around about how we do min/max settings. I agree that my initial ideas about this are not sustainable; we need a more general solution for policy control over settings.

skriss:
The min/max limits will be implemented via optional Gatekeeper policies.

Leaving this open to track removing the old config file field after a few releases.

skriss:
I'd like to remove the deprecated field in 1.10 - 3 minor releases after it was deprecated. That sound reasonable to everyone?

youngnick:
Yes, that's quite reasonable.

PR #2721, opened 2020-07-22
use internal/timeout for config file timeouts

skriss:
Changes all timeouts within the TimeoutConfig block in the config
file to use internal/timeout.Setting, to be consistent with timeouts
specified in HTTPProxy.

updates #2697 

Couple notes for reviewers:
- I did not update the existing `RequestTimeout` config field; that's slated to move into TimeoutConfig per #2690 and the new field will use this pattern.
- there are still some open discussion threads in #2697 on some details of the timeout syntax, but I think those can be addressed in subsequent PRs once we reach consensus.


cmd/contour/serve.go
```diff
@@ -184,10 +185,10 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		AccessLogFields:       ctx.AccessLogFields,
 		MinimumTLSVersion:     annotation.MinTLSVersion(ctx.TLSConfig.MinimumProtocolVersion),
 		RequestTimeout:        ctx.RequestTimeout,
-		ConnectionIdleTimeout: ctx.ConnectionIdleTimeout,
-		StreamIdleTimeout:     ctx.StreamIdleTimeout,
-		MaxConnectionDuration: ctx.MaxConnectionDuration,
-		DrainTimeout:          ctx.DrainTimeout,
+		ConnectionIdleTimeout: timeout.Parse(ctx.ConnectionIdleTimeout),
```
stevesloka, line 188: What happens if someone passes an invalid config option, it's disabled?

skriss: Currently yes, it would be disabled, which is consistent with how timeouts specified in HTTPProxy are treated today. But there's discussion in #2697 about bubbling these up as errors instead, which I think makes more sense.

stevesloka: Fantastic! :tada:

PR #2727, opened 2020-07-24
internal/envoy: use a consistent SNI server name for H2

jpeach:
When the SNI server name is overridden by rewriting the `Host` header,
Contour was updating the outbound SNI name for the `tls` protocol but
still using the external name value for the `h2` protocol. Since `h2`
implies TLS, Contour should do the same thing in both cases.

### DIFF
--- a/internal/featuretests/externalname_test.go
+++ b/internal/featuretests/externalname_test.go
@@ -17,6 +17,7 @@ import (
 	"testing"
 
 	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
+	envoy_api_v2_core "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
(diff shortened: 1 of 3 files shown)


internal/featuretests/externalname_test.go
```diff
@@ -27,6 +28,10 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+func stringPtr(s string) *string {
```
stevesloka, line 31: I've used this in the past for string pointers: https://github.com/kubernetes/utils/blob/6999998975a717e7f5fc2a7476497044cb111854/pointer/pointer.go#L74

jpeach: Nice!

PR #2732, opened 2020-07-25
travis: Run integration tests on each build

stevesloka:
Adds a `makefile` task which spins up a kind cluster, loads in a Contour build and runs all the integrations tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -311,6 +311,11 @@ site-check: ## Test the site's links
 	docker run --rm -v $$(pwd)/site:/site -it $(JEKYLL_IMAGE) \
 		bash -c "cd /site && bundle install --path bundler/cache && bundle exec jekyll build && htmlproofer --assume-extension /site/_site"
 
+integration: ## Run integration tests against a real k8s cluster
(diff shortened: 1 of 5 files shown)


stevesloka:
I'm having flakey runs with the testing. It seems like sometimes the tests will succeed and sometimes they will fail. I want to try and make these a priority in Contour as we grow the feature set out which is why flushing out these nits I think is helpful in the long run.

Makefile
```diff
@@ -310,6 +310,11 @@ site-check: ## Test the site's links
 	docker run --rm -v $$(pwd)/site:/site -it $(JEKYLL_IMAGE) \
 		bash -c "cd /site && bundle install --path bundler/cache && bundle exec jekyll build && htmlproofer --assume-extension /site/_site"
 
+integration: ## Run integration tests against a real k8s cluster
+	./_integration/testsuite/make-kind-cluster.sh
+	./_integration/testsuite/run-test-case.sh ./_integration/testsuite/httpproxy/*
```
jpeach, line 315: Running `httpproxy/*.yaml` would be slightly safer.

PR #2734, opened 2020-07-27
internal/contour: ensure the http.Router filter is present

### LINKED ISSUE #2733: Curl request hangs when SNI is omitted (httpproxy fallback certificate mode)
@jpeach and I have been actively investigating this. We succeeded in reliably recreating it and James has already developed a fix. This issue is for tracking purposes so we can document when the fix is merged. Thanks James!

This is similar, but separate from the previously filed https://github.com/projectcontour/contour/issues/2720

While the correct fallback certificate is presented, the request seems to hang and never succeeds. 

Request to succeed when SNI is omitted. `curl` examples below
```
# succeeds (SNI included) 
curl -v -k https://fallback.instakube.io:8843/healthy

# hangs and eventually times out after presenting fallback cert (SNI omitted)
curl -v -k -H "Host:fallback.instakube.io" https://127.0.0.1:8843/healthy
```

Contour version: master
Kubernetes version: v1.18.6
Kubernetes installer & version: `kind` via https://github.com/projectcontour/contour/blob/master/_integration/testsuite/make-kind-cluster.sh

jpeach:
Ensure that the `http.Router` filter is present whenever we build a
HTTP Connection Manager. The `Router` filter is responsible for actually
forwarding HTTP requests, so omitting it leads to configurations that
are syntactically valid but otherwise inoperable.

This fixes #2733.

### DIFF
--- a/_integration/testsuite/run-test-case.sh
+++ b/_integration/testsuite/run-test-case.sh
@@ -33,5 +33,5 @@ exec "$TESTER" run \
     --param proxy.address="$ADDRESS" \
     --param proxy.http_port="$HTTP_PORT" \
     --param proxy.https_port="$HTTPS_PORT" \
-    --watch pods,secrets \
+    --watch pods,secrets,configmaps \
     "$@"
--- a/internal/envoy/listener.go
+++ b/internal/envoy/listener.go
@@ -220,11 +220,34 @@ func (b *httpConnectionManagerBuilder) AddFilter(f *http.HttpFilter) *httpConnec
 	return b
 }
 
+// Validate runs builtin validation rules against the current builder state.
+func (b *httpConnectionManagerBuilder) Validate() error {
+	filterNames := map[string]struct{}{}
+
+	for _, f := range b.filters {
+		filterNames[f.Name] = struct{}{}
+	}
+
+	// If there's no router filter, requests won't be forwarded.
+	if _, ok := filterNames[wellknown.Router]; !ok {
+		return fmt.Errorf("missing required filter %q", wellknown.Router)
+	}
+
+	return nil
+}
+
 // Get returns a new http.HttpConnectionManager filter, constructed
 // from the builder settings.
 //
 // See https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/network/http_connection_manager/v2/http_connection_manager.proto.html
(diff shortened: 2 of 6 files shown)


_integration/testsuite/run-test-case.sh
```diff
@@ -33,5 +33,5 @@ exec "$TESTER" run \
     --param proxy.address="$ADDRESS" \
     --param proxy.http_port="$HTTP_PORT" \
     --param proxy.https_port="$HTTPS_PORT" \
-    --watch pods,secrets \
+    --watch pods,secrets,configmaps \
```
stevesloka, line 36: Looks like this might be a change when you were testing an integration test?

internal/envoy/listener.go
```diff
@@ -220,11 +220,34 @@ func (b *httpConnectionManagerBuilder) AddFilter(f *http.HttpFilter) *httpConnec
...
+// Validate runs builtin validation rules against the current builder state.
+func (b *httpConnectionManagerBuilder) Validate() error {
+	filterNames := map[string]struct{}{}
+
+	for _, f := range b.filters {
+		filterNames[f.Name] = struct{}{}
+	}
+
+	// If there's no router filter, requests won't be forwarded.
+	if _, ok := filterNames[wellknown.Router]; !ok {
```
stevesloka, line 232: I understand your thought here to add this, but seems like a better place for unit tests to cover. There are probably other items that we should be checking on as well.

_integration/testsuite/run-test-case.sh
jpeach: Yep, though I added it to this PR on purpose. I can remove it.

internal/envoy/listener.go
jpeach: > I understand your thought here to add this, but seems like a better place for unit tests to cover. 

This is by way of defense in depth. I think it's hard for unit tests to cover that every call site does the right thing.

> There are probably other items that we should be checking on as well.

Yep, we can add those here in the future. If you know of things we should check, I can add some extras now.

jpeach:
> Just a small nit on the run-test-case.sh script.
> 
> Also, I think we should collapse the HTTPConnectionManagerBuilder into less pieces as a future change, but that's a different PR.

Yeh, I expect we will see more changes to this API, e.g. the `ext_authz` filter needs to go in between the current list of default filters.

PR #2741, opened 2020-07-28
add response-timeout-min/max settings to config file

### LINKED ISSUE #2690: move serveContext.RequestTimeout into TimeoutConfig and drop old field

skriss:
Adds optional response timeout min/max settings to the config
file. If specified, these fields limit the allowed per-route
response timeouts on HTTPProxy and Ingress.

closes #2690


cmd/contour/serve.go
```diff
@@ -199,6 +199,11 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 
 	listenerConfig.DefaultHTTPVersions = defaultHTTPVersions
 
+	responseTimeoutRange, err := timeout.ParseRange(ctx.ResponseTimeoutMin, ctx.ResponseTimeoutMax)
+	if err != nil {
+		return fmt.Errorf("failed to parse request timeout min/max: %w", err)
```
stevesloka, line 204: Should we use the logger here to print out the error message? I know in other parts of serve.go we do the same so at least this is consistent.

internal/timeout/timeout.go
```diff
@@ -83,3 +94,90 @@ func Parse(timeout string) Setting {
...
+	maxSet bool
+}
+
+// NewAllowedRange creates an AllowedRange with the given min and
+// max allowed durations.
+func NewAllowedRange(min, max time.Duration) AllowedRange {
+	return AllowedRange{
+		min:    min,
+		max:    max,
+		maxSet: true,
```
stevesloka, line 116: If `max` is not configured, the essentially it becomes `math.MaxInt64`? Is there an Envoy default if not set? 

Just curious since it sort of contradicts the comment above (https://github.com/projectcontour/contour/pull/2741/files#diff-555ffae2a78c3e333d329e38f9e252aeR258-R259).

cmd/contour/serve.go
skriss: I *think* we want this one to be a fatal error, which is why I'm returning it rather than just logging it. I was following the pattern from just above for parsing the default HTTP version from the config file. Alternately, we could warn and resort to no limits.

skriss:
closing this out for now as there are some fundamental concerns about the approach.

PR #2745, opened 2020-07-29
internal/contour: simplify the endpoints translator

jpeach:
Breaking EndpointsTranslator into two parts doesn't provide any benefits
and we can make the code easier to follow by removing all the indirection.

### DIFF
--- a/internal/contour/endpointstranslator.go
+++ b/internal/contour/endpointstranslator.go
@@ -33,14 +33,17 @@ import (
 // A EndpointsTranslator translates Kubernetes Endpoints objects into Envoy
 // ClusterLoadAssignment objects.
 type EndpointsTranslator struct {
+	Cond
 	logrus.FieldLogger
-	clusterLoadAssignmentCache
+
+	mu      sync.Mutex
+	entries map[string]*v2.ClusterLoadAssignment
 }
 
 func (e *EndpointsTranslator) OnAdd(obj interface{}) {
 	switch obj := obj.(type) {
 	case *v1.Endpoints:
-		e.addEndpoints(obj)
+		recomputeClusterLoadAssignment(e, nil, obj)
 	default:
 		e.Errorf("OnAdd unexpected type %T: %#v", obj, obj)
 	}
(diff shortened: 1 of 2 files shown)


internal/contour/endpointstranslator.go
```diff
@@ -34,14 +34,17 @@ import (
 // A EndpointsTranslator translates Kubernetes Endpoints objects into Envoy
 // ClusterLoadAssignment objects.
 type EndpointsTranslator struct {
+	Cond
 	logrus.FieldLogger
-	clusterLoadAssignmentCache
+
+	lock    sync.Mutex
```
stevesloka, line 40:
```suggestion
	mu     sync.Mutex
```

To be consistent with Contour, let's call this `mu`.

go.mod
```diff
@@ -28,6 +28,7 @@ require (
 	k8s.io/client-go v0.18.2
 	k8s.io/gengo v0.0.0-20191120174120-e74f70b9b27e // indirect
 	k8s.io/klog v1.0.0
+	k8s.io/utils v0.0.0-20200324210504-a9aa75ae1b89
```
stevesloka, line 31: I don't see this used in the PR? Bad commit?

skriss: I think this was left out from https://github.com/projectcontour/contour/pull/2727 - I noticed this yesterday. I guess we don't have any CI checks to ensure go.mod is up-to-date.

stevesloka: I'm surprised it didn't yell on the build since it's marked as readonly. In any case then this is fine since it's missing.

jpeach: Yeh, `make check` yelled at me about it :(

jpeach: The same change is in #2746, so I'll merge that and rebase this on top.

jpeach:
@stevesloka @skriss PTAL. I also restored the original name of `recomputeClusterLoadAssignment` but made it a free function to clarify the `EndpointsTranslator` locking rules.

internal/contour/endpointstranslator.go
```diff
@@ -97,27 +118,33 @@ func (e *EndpointsTranslator) Query(names []string) []proto.Message {
...
+	defer e.mu.Unlock()
+
+	delete(e.entries, name)
+	e.Notify(name)
 }
 
 // recomputeClusterLoadAssignment recomputes the EDS cache taking into account old and new endpoints.
-func (e *EndpointsTranslator) recomputeClusterLoadAssignment(oldep, newep *v1.Endpoints) {
-	// skip computation if either old and new services or endpoints are equal (thus also handling nil)
+func recomputeClusterLoadAssignment(e *EndpointsTranslator, oldep, newep *v1.Endpoints) {
```
stevesloka, line 145: Any reason to not make this part of `EndpointsTranslator` when you wouldn't need to pass in the ref each time to this method?

jpeach: This is to emphasize that only public methods of `EndpointsTranslator` are used. This makes it clearer (IMHO) that no locking is required here. When this is a member function, the reader has to think about whether it directly accesses internal fields (which would be legal but not conventional) or should hold locks.

PR #2747, opened 2020-07-29
add github actions pr workflow

### LINKED ISSUE #2200: Move off Travis to a different CI system (GitHub Actions or CircleCI)
To simplify our setup, and have more consistency with other OSS projects in the same team, it's desirable to move away from Travis. There have been good results with GitHub Actions, and it's GA now, so that is likely the best option. CircleCI also has more functionality here.

Moving to GitHub actions will bring the maintainers and community a more consolidated "single source of truth" for introspecting build results over time. The Contour maintainers are already using Actions for the release process, so this move would bring these different stages of our build process together.

stevesloka:
Adds Github Actions workflow to mirror what is done with TravisCI for each push or pull request.

Fixes #2200


examples/kind/kind-expose-port.yaml
```diff
@@ -2,7 +2,6 @@ kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
 - role: control-plane
-- role: worker
```
jpeach, line 5: Is this change intentional?

stevesloka: I was looking for ways to reduce resource usage in the integration tests, so my thought was taking out this worker would help and I think it has.

jpeach: This isn't used though, did you mean `_integration/testsuite/kind-expose-port.yaml`?

PR #2751, opened 2020-07-30
cmd/contour: refactor shutdown-manager to initiate shutdown using an Exec command

stevesloka:
This changes the preStop lifecycle hook to utilize a command instead of an http request. The /shutdown http endpoint now only returns when the pod is ready to be terminated, and no longer initiates the Envoy shutdown sequence.
    
The Envoy container will block on its preStop hook until the /shutdown endpoint returns. That endpoint, implemented in the shutdown-manager, will utilize a file messaging pattern to know when it is safe for Envoy to terminate by checking for the existence of a file in the filesystem.

### DIFF
--- a/cmd/contour/shutdownmanager.go
+++ b/cmd/contour/shutdownmanager.go
@@ -18,6 +18,7 @@ import (
 	"io"
 	"log"
 	"net/http"
+	"os"
 	"time"
 
 	"github.com/projectcontour/contour/internal/contour"
@@ -34,6 +35,9 @@ const (
 	healthcheckFailURL = "http://localhost:9001/healthcheck/fail"
 	prometheusStat     = "envoy_http_downstream_cx_active"
 )
+
+// File path used in the /shutdown endpoint.
+const shutdownReadyFile = "/ok"
 
 func prometheusLabels() []string {
 	return []string{contour.ENVOY_HTTP_LISTENER, contour.ENVOY_HTTPS_LISTENER}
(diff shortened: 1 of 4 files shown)


cmd/contour/shutdownmanager.go
```diff
@@ -66,21 +70,38 @@ func newShutdownManagerContext() *shutdownmanagerContext {
...
 	}
 }
 
-// shutdownHandler handles the /shutdown endpoint which should be called from a pod preStop hook,
-// where it will block pod shutdown until envoy is able to drain connections to below the min-open threshold
-func (s *shutdownmanagerContext) shutdownHandler(w http.ResponseWriter, r *http.Request) {
+// shutdownReadyHandler handles the /shutdown endpoint which is used by Envoy to determine if it can terminate.
+// Once enough connections have drained based upon configuration, a file will be written to "/ok" in
+// the shutdown manager's file system. Any HTTP request to /shutdown will use the existence of this
+// file to understand if it is safe to terminate.
```
skriss, line 84: It's worth mentioning that we're using a file here because the pre-stop exec hook runs in a different process than the `/shutdown` handler so we need a means to communicate between them.

cmd/contour/shutdownmanager.go
```diff
@@ -66,21 +70,38 @@ func newShutdownManagerContext() *shutdownmanagerContext {
...
+func (s *shutdownmanagerContext) shutdownReadyHandler(w http.ResponseWriter, r *http.Request) {
+	for {
+		if _, err := os.Stat(shutdownReadyFile); err == nil {
```
skriss, line 87: It'd be nice to log any non-nil, non-IsNotFound errors (I guess we could log the not-found errors as well, but they're expected until the connections have drained).

cmd/contour/shutdownmanager.go
stevesloka: Updated the comment.

cmd/contour/shutdownmanager.go
stevesloka: I added and `else` to check for errors and log them.

PR #2755, opened 2020-08-03
testsuite: add a fallback certificate test

jpeach:
Add a simple integration test to ensure that using the fallback
certificate works as expected. This requires that Contour is deployed
with a valid fallback certificate. For integration testing, we can use
cert-manager to pre-configure that, but this approach won't work well
for large numbers of configuration permutations, so we should revisit
it relatively soon.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -313,6 +313,7 @@ site-check: ## Test the site's links
 
 integration: ## Run integration tests against a real k8s cluster
 	./_integration/testsuite/make-kind-cluster.sh
+	./_integration/testsuite/install-fallback-certificate.sh
 	./_integration/testsuite/run-test-case.sh ./_integration/testsuite/httpproxy/*.yaml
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -313,6 +313,7 @@ site-check: ## Test the site's links
 
 integration: ## Run integration tests against a real k8s cluster
 	./_integration/testsuite/make-kind-cluster.sh
+	./_integration/testsuite/install-fallback-certificate.sh
```
stevesloka, line 316: Can the test apply these bits? This might get out of hand as we add more test logic.

jpeach: Not really. I discussed this on slack .. the problem is that we have no way to roll back the change that we make to the contour deployment. It's theoretically possible to do that, but also pretty involved.

jpeach: But yes, trying to test different Contour install configs will rapidly get out of hand :-(

PR #2765, opened 2020-08-04
Parse yaml

### LINKED ISSUE #2623: Enable strict YAML parsing for the configuration file.

Added the following to the Contour config file:
```yaml
    default-http-vertsions:
     - "HTTP/1.1"
```

Contour started up and I had to debug why the config didn't seem to work.

Any kind of error.

xref #2621

tthebst:
Enables strict configuration YAML parsing.

Fixes #2623

Signed-off-by: Tim Gretler "gretler.tim@gmail.com"

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -76,6 +76,7 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 		}
 		defer f.Close()
 		dec := yaml.NewDecoder(f)
+		dec.SetStrict(true)
 		parsed = true
 		return dec.Decode(&ctx)
 	}
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -78,7 +78,10 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 		dec := yaml.NewDecoder(f)
 		dec.SetStrict(true)
 		parsed = true
-		return dec.Decode(&ctx)
+		if err := dec.Decode(&ctx); err != nil {
+			return fmt.Errorf("failed to parse contour configuration: %w", err)
+		}
+		return nil
 	}
 
 	serve.Flag("config-path", "Path to base configuration.").Short('c').Action(parseConfig).ExistingFileVar(&configFile)


youngnick:
Thanks for this PR @tthebst! Could you do the DCO signoff again on the initial PR? It looks like the DCO bot is not recognising what you have there.

```
git commit --amend --signoff
```
Should maybe do the trick?

cmd/contour/serve.go
```diff
@@ -76,6 +76,7 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 		}
 		defer f.Close()
 		dec := yaml.NewDecoder(f)
+		dec.SetStrict(true)
```
jpeach, line 79: I think that we should decorate the YAML error, since it doesn't give much context:
```
$ contour serve -c /tmp/t
contour: error: yaml: unmarshal errors:
  line 2: field configs not found in type main.serveContext, try --help
```

Suggest:
```Go
if err := dec.Decode(&ctx); err != nil {
        return fmt.Errorf("failed to parse configuration file %q: %w", configFile, err)
}

return nil
```

tthebst: I agree with decorating the error.

I just tested your suggestion.  This is the output. 

```
contour: error: failed to parse configuration file "/config/contour.yaml": yaml: unmarshal errors:
  line 54: field defaulst-http-versions not found in type main.serveContext, try --help
```

`configFile` might be a little misleading if the user used a configMap to configure contour. This could be enough. 
```
if err := dec.Decode(&ctx); err != nil {
        return fmt.Errorf("failed to parse contour configuration: %w", err)
}
return nil
```

If you want you could also be more verbose.

```
if err := dec.Decode(&ctx); err != nil {
        return fmt.Errorf("failed to parse contour configuration: %w Valid values can be found here: https://projectcontour.io/docs/%s/configuration/ ", err,version)
}
return nil
```

jpeach: I think the second version is fine with "failed to parse contour configuration" is fine, if you prefer that. Thanks Tim 👍

tthebst: I will add it. btw sorry for the mess here, I made a git mistake while trying to signoff an old commit.

PR #2766, opened 2020-08-05
internal/dag: improve status errors for invalid TLS

### LINKED ISSUE #2273: httpproxy: spec.virtualhost.tls.secretName is optional, or is it?
```
type TLS struct {
        // required, the name of a secret in the current namespace.
        SecretName string `json:"secretName,omitempty"`
```
The comment says spec.virtualhost.tls.secretName is required, but its marked as optional by dint of the `omitempty` json tag. 

Decide if secretName is required, and if is, adjust validations, if not adjust the comment.

### LINKED ISSUE #2501: Misleading status message with TLS passthrough and secret
If you create a HTTPProxy document that contains both the `tls.passthrough` and `tls.secretName`fields, you get a really misleading error status:

```
$ k get proxies
NAME   FQDN                     TLS SECRET   STATUS    STATUS DESCRIPTION
echo   echo.projectcontour.io   echo-one     invalid   tcpproxy: missing tls.passthrough or tls.secretName
```

This is because:
```Go
        tlsValid = tls.Passthrough != !isBlank(tls.SecretName)

        if proxy.Spec.TCPProxy != nil {
                if !tlsValid {
                        sw.SetInvalid("tcpproxy: missing tls.passthrough or tls.secretName")
                        return
                }
     
        }
```

skriss:
When validating an HTTPProxy's TLS configuration, separately check
for cases where both secretName and passthrough are specified, and
for cases where neither secretName nor passthrough are specified.
Sets a corresponding HTTPProxy status message for both cases. Also
improves the status message for when TCPProxy is enabled, but TLS
is not enabled.

Fixes #2501 
Fixes #2273

### DIFF
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -115,17 +115,19 @@ type VirtualHost struct {
 }
 
 // TLS describes tls properties. The SNI names that will be matched on
-// are described in fqdn, the tls.secretName secret must contain a
-// matching certificate unless tls.passthrough is set to true.
+// are described in the HTTPProxy's Spec.VirtualHost.Fqdn field.
 type TLS struct {
-	// required, the name of a secret in the current namespace
+	// SecretName is the name of a TLS secret in the current namespace.
+	// Either SecretName or Passthrough must be specified, but not both.
+	// If specified, the named secret must contain a matching certificate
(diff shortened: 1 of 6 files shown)


jpeach:
Does this also fix https://github.com/projectcontour/contour/issues/2273 ?

youngnick:
Seems like it does, yes.

PR #2781, opened 2020-08-11
internal: encapsulate the Contour xDS server

### LINKED ISSUE #2775: Introduce DAG model consumer abstractions

We have a few different features coming up that affect the way we consume resources that Contour models in the DAG:

* Envoy API v3 #1898
* Endpoint slices #2696
* service clusters for auth #2769 
* go-control-plane #2134

Most of these features will need to be enabled or disables in configuration. This issue addresses creating the right internal abstractions for adding the features above without creating a ton of churn through Contour internals.

jpeach:
Add a `xds` package that exports a basic interface that xDS servers
can implement. Move the existing Contour xDS server into this package,
and add a factory function to create an instance of it.

After this change, different xDS server implementations can be selected
at startup.

This fixes #2775.


jpeach:
@stevesloka @skriss After this change, it should be straight forward to enable go-control-plane as the xDS server at runtime. The setup in `serve.go` will probably be a bit ugly though :-(

internal/xds/server.go
```diff
@@ -0,0 +1,57 @@
+package xds
```
skriss, line 1: needs a copyright header

Issue #2786, opened 2020-08-12
testing: protobuf testing improvements


youngnick:
As part of testing for #2134 in #2604, @stevesloka identified that we need a better way to compare protobufs. The current `internal/assert` package has a number of workarounds using `go-cmp/cmp` in order to be able to test Envoy protobufs, particularly in the `featuretests` package, but as we move towards #2134 and `go-control-plane`, we need some more flexibility.

This is to investigate using other options like [testify](https://github.com/stretchr/testify)'s `assert` or [`protocmp`](https://godoc.org/google.golang.org/protobuf/testing/protocmp) for comparing protobufs.

youngnick:
Okay, reviewing this one, I think I'm going to call it done.

PR #2787, opened 2020-08-12
internal/assert: Better protobuf diffs by aliasing testify/assert

youngnick:
This change replaces the current `github.com/google/go-cmp/cmp` based internal
`assert.Equal` with an aliased version of testify `assert.Equal`. If we need
other assertions, we'll need to add them one by one. This allows better output
for non-protobuf tests. However, it does mean that now Error message text is
being checked - there were about four places I needed to fix as a result.

Unfortunately, testify does not have support for unmarshaling protobufs, so
I needed to move to a new `assert.EqualProto` function that uses `protocmp`
to compare the two protobufs.

This also exposed another issue that the old `assert.Equal` was working around:
the Envoy DiscoveryResponse proto has version and nonce fields that can change in
ways that make tests flaky. However, it turns out that all the tests in `featuretests`
that needed the workaround only change the `Resources` field. So I've updated the
featuretests infrastructure to have the `Equals` method on the Response struct
for testing only compare that `Resources` field. This should probably be replaced
with `EqualResources` and `NoResources` for the normal and empty case respectively,
but this PR is big enough for now.

Updates #2786 for now, until a second PR that will change the names as above.


internal/assert/assert.go
```diff
@@ -39,14 +40,22 @@ func Equal(t *testing.T, want, got interface{}, opts ...cmp.Option) {
 
 // Equal will call t.Fatal if want and got are not equal.
 func (a Assert) Equal(want, got interface{}, opts ...cmp.Option) {
+	tassert.Equal(a.t, want, got, "Two values should be equal.")
```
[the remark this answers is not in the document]
youngnick: Okay, that makes this a much bigger change, but makes sense.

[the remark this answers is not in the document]
jpeach: > Okay, that makes this a much bigger change, but makes sense.

You don't have to update any call sites (the `cmp` options are never used).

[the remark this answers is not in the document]
youngnick: Actually, I suppose we can skip filling out the msgAndArgs part, so it's not so bad.

internal/assert/assert.go
```diff
@@ -17,54 +17,28 @@ package assert
 import (
 	"testing"
 
-	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
-	"github.com/golang/protobuf/proto"
-	"github.com/golang/protobuf/ptypes"
-	"github.com/golang/protobuf/ptypes/any"
 	"github.com/google/go-cmp/cmp"
-	"github.com/google/go-cmp/cmp/cmpopts"
+	tassert "github.com/stretchr/testify/require"
```
jpeach, line 21: This should be `testify/assert` so that `testing.Error` is used.

internal/assert/assert.go
```diff
@@ -17,54 +17,28 @@ package assert
...
+// EqualProto will test that want == got for protobufs, call t.Error if it does not,
+// and return a bool to indicate the result. This mimics the behavior of the testify `assert`
+// functions.
+func EqualProto(t *testing.T, want, got interface{}, msgAndArgs ...interface{}) bool {
 	t.Helper()
-	Assert{t}.Equal(want, got, opts...)
-}
 
-// Equal will call t.Fatal if want and got are not equal.
-func (a Assert) Equal(want, got interface{}, opts ...cmp.Option) {
```
skriss, line 41: I know there are a couple places where we were previously passing in `opts`, and I don't see them being removed. Looks like since `tassert.Equal` also takes a variadic arg of type `interface{}` those occurrences won't cause compile errors, but it'd still be good to clean them up since they are no longer relevant. I'd start by looking for `cmp.AllowUnexported` (e.g. https://github.com/projectcontour/contour/blob/master/internal/dag/policy_test.go#L170).

youngnick: Good catch, thanks!

youngnick: I made a temporary copy of `Equal` without the variadic params, then fixed up all those tests. Testify `assert` doesn't need the AllowUnexported thing, so all the tests work. I broke a few to check. :)

youngnick:
Okay, this is good for rereview now.

cmd/contour/serve_test.go
```diff
@@ -64,8 +63,8 @@ func TestGetRequestTimeout(t *testing.T) {
 	for _, tc := range tests {
 		log := logrus.New()
 		log.Out = ioutil.Discard
-		opt := cmp.AllowUnexported(timeout.Setting{})
+		//opt := cmp.AllowUnexported(timeout.Setting{})
```
jpeach, line 66: Just delete this?

internal/assert/assert.go
```diff
@@ -17,54 +17,28 @@ package assert
...
+)
 
-// Equal will test that want == got, and call t.Fatal if it does not.
-// Notably, for errors, they are equal if they are both nil, or are both non-nil.
-// No value information is checked for errors.
-func Equal(t *testing.T, want, got interface{}, opts ...cmp.Option) {
+// EqualProto will test that want == got for protobufs, call t.Error if it does not,
+// and return a bool to indicate the result. This mimics the behavior of the testify `assert`
+// functions.
+func EqualProto(t *testing.T, want, got interface{}, msgAndArgs ...interface{}) bool {
```
jpeach, line 34: Nit: drop the `msgAndArgs` argument since it's not used here.

cmd/contour/serve_test.go
youngnick: Thanks, oops.

PR #2795, opened 2020-08-16
internal/featuretests: add v1.Service fixture support

jpeach:
Add fixture support for generating `v1.Service` resources with a reduced
amount of boilerplate.

### DIFF
--- a/internal/fixture/service.go
+++ b/internal/fixture/service.go
@@ -1,0 +1,51 @@
+package fixture
+
+import (
+	v1 "k8s.io/api/core/v1"
+)
+
+type ServiceBuilder v1.Service
+
+// NewService creates a new ServiceBuilder with the given resource name.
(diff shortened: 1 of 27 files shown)


internal/fixture/service.go
```diff
@@ -0,0 +1,51 @@
...
+	s := &ServiceBuilder{
+		ObjectMeta: ObjectMeta(name),
+		Spec:       v1.ServiceSpec{},
+	}
+
+	return s
+}
+
+// Annotate adds the given values as metadata annotations.
+func (b *ServiceBuilder) Annotate(k string, v string) *ServiceBuilder {
```
youngnick, line 20: nit: why `b` here and `s` elsewhere?

jpeach: fixed.

jpeach:
AFAICT, test failures are flakes. Merging.

Issue #2589, opened 2020-06-16
Invalid annotation log messages include a hard-coded resource version 


jpeach:
In `(*KubernetesCache) Insert()` we see the following:

```Go
				kc.WithField("name", om.GetName()).
					WithField("namespace", om.GetNamespace()).
					WithField("kind", kind).
					WithField("version", "v1"). <<< 
					WithField("annotation", key).
					Error("ignoring invalid or unsupported annotation")
```

I think that the value of the "version" field should the API group/version.

ffahri:
Hey i want to change this, is it okey to use object.GetResourceVersion() ? @jpeach

ffahri:
I was confused with the resourceVersion sorry, we can get apiVersion with using same as 
k8s.kindOf  -> k8s.apiVersionOf func, and retrieving the version from that function ? 
I thought maybe we can use unstructured type for this but it says 
```
WARNING: This object has accessors for the v1 standard metadata. You *MUST NOT* use this type if you are dealing with objects that are not in the server meta v1 schema.
```
https://godoc.org/github.com/kubernetes/apimachinery/pkg/apis/meta/v1/unstructured

So another option can be a jsonpath of the $object.apiVersion what is your thoughts of this ?

jpeach:
> I was confused with the resourceVersion sorry, we can get apiVersion with using same as
> k8s.kindOf -> k8s.apiVersionOf func, and retrieving the version from that function ?

Yeh we will need something like this, since it's difficult to guarantee that the resource meta is populated in all cases.

> I thought maybe we can use unstructured type for this but it says
> 
> ```
> WARNING: This object has accessors for the v1 standard metadata. You *MUST NOT* use this type if you are dealing with objects that are not in the server meta v1 schema.
> ```

This is fine, because every resource we deal with does use the meta v1 schema.

ffahri:
Hey @jpeach i tried out with different thing and succeeded but 
```
	return obj.(runtime.Object).GetObjectKind().GroupVersionKind().GroupVersion().String()

---
//test case objects
{"test.projectcontour.io/v1", &unstructured.Unstructured{
			Object: map[string]interface{}{
				"apiVersion": "test.projectcontour.io/v1",
				"kind":       "Foo",
			}}
},
{"apps/v1", &v1.Pod{
			TypeMeta: metav1.TypeMeta{
				APIVersion: "apps/v1",
			}}
},

```

however you said 
[...]

PR #2796, opened 2020-08-16
#2589 - add groupversion for logging

ffahri:
Signed-off-by: Fahri Yardımcı

### DIFF
--- a/internal/k8s/kind_test.go
+++ b/internal/k8s/kind_test.go
@@ -17,7 +17,7 @@ func TestKindOf(t *testing.T) {
 		{"Secret", &v1.Secret{}},
 		{"Service", &v1.Service{}},
 		{"Endpoints", &v1.Endpoints{}},
-		{"", &v1.Pod{}},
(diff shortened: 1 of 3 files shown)


ffahri:
+ also not sure about the class 
should i create version.go or keep using kind.go

ffahri:
Trying to scheme.Scheme.ObjectKinds function as a generic approach

internal/k8s/kind_test.go
```diff
@@ -17,7 +17,7 @@ func TestKindOf(t *testing.T) {
 		{"Secret", &v1.Secret{}},
 		{"Service", &v1.Service{}},
 		{"Endpoints", &v1.Endpoints{}},
-		{"", &v1.Pod{}},
+		{"Pod", &v1.Pod{}},
```
ffahri, line 20: Since we can get Pod Kind using schema.ObjectKind, changed this

ffahri:
no prob :+1:

Issue #2226, opened 2020-02-13
internal/dag: Refactor builder


stevesloka:
The first step to build the DAG is to convert resources into internal Contour types which are done in [`internal/dag/builder.go`](https://github.com/projectcontour/contour/blob/master/internal/dag/builder.go). 

The builder processes the following types:

- Ingress/v1beta1 
- IngressRoute/v1beta1
- TLSCertificateDelegation/v1beta1
- HTTPProxy/v1
- TLSCertificateDelegation/v1

The builder has grown to be a single file handling all these types and I think could win by being split apart to make it more clear, especially as this list of types grows and the Contour API's progress.

jpeach:
I think that this is closely related to https://github.com/projectcontour/contour/issues/2683.

If we define a builder abstraction, we can have independent builders for all the APIs that we want to support (ingress, httproute, service apis). However that needs the support of a generic Kubernetes resource cache (i.e. #2683). All the builders end up creating objects in the DAG, then you can have a DAG visitor interface like the one in #2753 to generate envoy resources in whatever way you want. This approach would result in a set of loosely coupled components that collaborate via the DAG.

skriss:
Now that #2847 has landed I'm starting to look at having the `Processors` (probably to be renamed builders eventually) operate directly on the DAG, rather than a Builder. Looking at what a good API for that looks like.

PR #2802, opened 2020-08-18
internal/dag: Refactor some logic from builder into KubernetesCache

stevesloka:
Moves some common components from the builder into the KubernetesCache
so that it can be reused from other components. This is the first of a few PRs in refactoring
the builder out. 

Updates #2226

### DIFF
--- a/internal/dag/cache.go
+++ b/internal/dag/cache.go
@@ -14,6 +14,7 @@
 package dag
 
 import (
+	"fmt"
 	"sync"
 
 	projectcontour "github.com/projectcontour/contour/apis/projectcontour/v1"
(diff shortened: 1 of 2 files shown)


internal/dag/cache.go
```diff
@@ -405,3 +406,22 @@ func (kc *KubernetesCache) secretTriggersRebuild(secret *v1.Secret) bool {
 
 	return false
 }
+
+// lookupSecret returns a Secret if present or nil if the underlying kubernetes
```
jpeach, line 410: nit: `LookupSecret`.

internal/dag/cache.go
```diff
@@ -405,3 +406,22 @@ func (kc *KubernetesCache) secretTriggersRebuild(secret *v1.Secret) bool {
...
+// secret fails validation or is missing.
+func (kc *KubernetesCache) LookupSecret(m types.NamespacedName, validate func(*v1.Secret) error) (*Secret, error) {
```
jpeach, line 412: Nit: the `m` argument should probably be `n` for "name"?

Issue #2769, opened 2020-08-06
Introduce ServiceCluster concept.


jpeach:
**Please describe the problem you have**

The Contour DAG model needs a new concept to be able to represent the concept of an Envoy `ClusterLoadAssignment` that is backed by multiple (weighted) Kubernetes services.

The following is pretty rough, but I'm thinking of something like this to represent the load assignment:
```Go
type WeightedService struct {
       Weight      uint32
       ServiceName types.NamespacedName
       ServicePort v1.ServicePort
}

// ServiceCluster capture the set of Kubernetes Services that will
// compose the endpoints for a Envoy cluster.
type ServiceCluster struct {
       Services []WeightedService
}
```

Then we could rename the `dag.Service` to something more specific (maybe `RouteService`).

Next, the  Kubernetes Service field in `RouteService` would be replaced by a `ServiceCluster` with 1 element, and accessors and invariants added as appropriate.

jpeach:
This blocks https://github.com/projectcontour/contour/issues/2713

jpeach:
Fixed in #2805

PR #2805, opened 2020-08-18
internal: rebuild endpoints based on the DAG services

jpeach:
The EndpointsTranslator generates Envoy ClusterLoadAssignment resources
in a fixed configuration. Each service/port pair generates a separate
ClusterLoadAssignment, that is associated with an Envoy Cluster solely
by its matching name.

This change introduces a new ServiceCluster concept that represents a
single set of endpoints, weighted across multiple Kubernetes Services.
This allows the upcoming ext_auth support to weight traffic, even though
(by definition), it still uses a single Envoy Cluster. The weighted
balancing is done within the cluster.

The DAG Service object is updated to use a ServiceCluster rather than
referencing a Service directly. This is exposed to the EndpointsTranslator,
which is now a DAG observer. The EndpointsTranslator rebuilds the
ClusterLoadAssigments globally on a DAG rebuild, and incrementally
on an Endpoints update.

The only semantic change here is that Contour will no longer generate
Envoy ClusterLoadAssignment resources for all Kubernetes Services. They
will only be generated for Services that are referenced by an Ingress
or HTTPProxy resource.

This updates #2769.

### DIFF
--- a/internal/contour/endpointstranslator.go
+++ b/internal/contour/endpointstranslator.go
@@ -14,6 +14,7 @@
 package contour
 
 import (
+	"fmt"
 	"sort"
 	"sync"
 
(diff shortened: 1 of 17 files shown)


internal/dag/dag.go
```diff
@@ -371,10 +372,7 @@ func (t *TCPProxy) Visit(f func(Vertex)) {
 
 // Service represents a single Kubernetes' Service's Port.
 type Service struct {
-	Name      string
-	Namespace string
-
-	ServicePort v1.ServicePort
+	Service WeightedService
```
stevesloka, line 375: Do we need `WeightedService` here? Could just use `Service` and remove the stutter of `service.service.xxx`.

internal/xds/util.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// ClusterLoadAssignmentName generates the name used for an EDS
+// ClusterLoadAssignment, given a fully qualified Service name and
+// port. This name is a contract between the producer of a cluster
+// (i.e. the EDS service) and the consumer of a cluster (most likely
+// a HTTP Route Action).
+func ClusterLoadAssignmentName(service types.NamespacedName, portName string) string {
```
stevesloka, line 27: Should this be in the `internal/envoy` package?

internal/contour/endpointstranslator.go
```diff
@@ -26,30 +29,307 @@ import (
...
+		if s.ClusterName == "" {
+			return errors.New("missing .ClusterName field")
+		}
+
+		if len(s.Services) == 0 {
+			return errors.New("empty .Services field")
+		}
+
+		for i, w := range s.Services {
+			if w.Weight < 1 {
```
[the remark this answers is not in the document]
jpeach: Allowing weight of 0 would disable traffic to the service. Seems like an unnecessary footgun. What was your use case for a weight of 0?

internal/contour/endpointstranslator.go
```diff
@@ -26,30 +29,307 @@ import (
...
+
+	e.Infof("started DAG rebuild")
+	defer e.Infof("finished DAG rebuild")
+
+	var visitor func(dag.Vertex)
+	visitor = func(vertex dag.Vertex) {
+		if svc, ok := vertex.(*dag.ServiceCluster); ok {
+
+			if _, ok := names[svc.ClusterName]; ok {
+				e.Errorf("dropping service cluster with duplicate name %q", svc.ClusterName)
```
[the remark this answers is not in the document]
jpeach: Maybe this should panic? It's a programming error if Contour generates duplicate ClusterNames. Perhaps in a separate PR, we can add a DAG validation pass that enforces these kinds of invariant?

internal/xds/util.go
jpeach: It was previously in the envoy package, but now that it is used in the `dag` package, that resulted in import loops that I needed to break.

internal/dag/dag.go
jpeach: Originally, I wanted to use a full ServiceCluster, but that ended up being a bit too much refactoring for now.

internal/contour/endpointstranslator.go
```diff
@@ -26,30 +28,295 @@ import (
...
+					&LocalityEndpoints{
+						LbEndpoints:         lb,
+						LoadBalancingWeight: protobuf.UInt32(w.Weight),
+					},
+				)
+			}
+		}
+
+		assignments[cla.ClusterName] = &cla
+		log.Printf("currently %d clusters", len(assignments))
```
stevesloka, line 141: This should be a debug level and use the logger to print properly. 

image

internal/contour/endpointstranslator.go
```diff
@@ -26,30 +28,295 @@ import (
...
 
+// Merge combines the given entries with the existing entries in the
+// EndpointsTranslatore. If the same key exists in both maps, an existing entry
+// is replaced.
+func (e *EndpointsTranslator) Merge(entries map[string]*v2.ClusterLoadAssignment) {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+
+	for k, v := range entries {
+		e.Infof("added LoadAssignment name=%q as %s", k, v)
```
stevesloka, line 262: These should be debug level.

internal/contour/endpointstranslator.go
```diff
@@ -111,107 +388,17 @@ func (e *EndpointsTranslator) Query(names []string) []proto.Message {
 	for _, n := range names {
 		v, ok := e.entries[n]
 		if !ok {
+			e.Infof("no cache entry for %q", n)
 			v = &v2.ClusterLoadAssignment{
 				ClusterName: n,
 			}
 		}
 		values = append(values, v)
+		e.Infof("query response for %q: %s", n, v)
```
stevesloka, line 397: debug level

stevesloka:
I'm running a test of this now, need to check what the metrics end up looking like since now you can't get metrics per cluster? Is there a way to expose each locality? Just looking forward to what changes for users with the dashboard stuff. 

Also, does the Contour CLI also still work? I'm guessing it's fine since EDS/CDS streaming didn't change, something else that just came up thinking through this.

jpeach:
> I'm running a test of this now, need to check what the metrics end up looking like since now you can't get metrics per cluster?

Since the name of the load assignment for `dag.Service` objects doesn't change, there should be no impact to metrics. Can we work together to verify that?

> Is there a way to expose each locality?

I don't think so; the localities have weights but not names. Not clear to me what envoy does with those metrics.

> Just looking forward to what changes for users with the dashboard stuff.
> 
> Also, does the Contour CLI also still work? I'm guessing it's fine since EDS/CDS streaming didn't change, something else that just came up thinking through this.

I can check this.

internal/dag/dag.go
stevesloka: Is there a reason to not just remove the extra struct? I think it would read much better.

internal/xds/util.go
stevesloka: oh I see now, makes sense. Unfortunate though this is getting all mixed around.

stevesloka:
So I understand a bit more now playing around with this @jpeach. The localities are implemented but not used just yet.  I thought this was changing how routes were created by using a single cluster with both sets of endpoints from two different services. 

The only real big question left is if we want to tie this into the dag rebuild or not.  We didn't in the past since endpoints update much more frequently than ingress resources which is why it wasn't part of that. But I think the rebuild happens fast enough that it might not matter? That's the only large thing left, curious your thoughts @skriss & @youngnick.

jpeach:
> So I understand a bit more now playing around with this @jpeach. The localities are implemented but not used just yet. I thought this was changing how routes were created by using a single cluster with both sets of endpoints from two different services.

The HTTPProxy API depends on how clusters are currently constructed. We can't move the weighting down into the load assignment without breaking things. This change should be a no-op semantically.

> The only real big question left is if we want to tie this into the dag rebuild or not.

In the next patch, external services will use service clusters directly and we will need to pull them out of the DAG. That's the reason why this rewrite was needed. Only the DAG can tell us how multiple Kubernetes services can be bundled into a single load assignment.

>  We didn't in the past since endpoints update much more frequently than ingress resources which is why it wasn't part of that. But I think the rebuild happens fast enough that it might not matter? That's the only large thing left, curious your thoughts @skriss & @youngnick.

Yeh, we really ought to have metrics around this. I thought about adding something here, but really we might as well add metrics for updates to all the xDS resource types.

internal/dag/dag.go
jpeach: Keeping the struct improves readability and discoverability. It's clear to the reader that a WeighedService is being used, whereas if the fields were embedded directly it would be nard to find.

stevesloka: Sure but that only makes sense if you're going to have another type of service. If you have a service with no weights, the same "weighted" struct is still used, so I don't think it's any clearer. 

If wanting to call out the specifics of the weighted service then it should be renamed to `Weighted` or something so you'd call it as `Service.Weighted` since calling it `Service.Service` doesn't make it any clearer in my eye. 

I'd just remove the extra struct, but ¯\_(ツ)_/¯

jpeach: I can rename to `Weighted`.

internal/contour/endpointstranslator.go
```diff
@@ -26,30 +27,289 @@ import (
...
+	c.endpoints[name] = ep.DeepCopy()
+
+	// If any service clusters include this endpoint, mark them
+	// all as stale.
+	if affected := c.services[name]; len(affected) > 0 {
+		c.stale = append(c.stale, affected...)
+	}
+}
+
+// DeleteEndpoint deletes ep from the cache, or replaces it if it
```
youngnick, line 201: nit: I don't think the endpoint can be replaced if it's deleted?

internal/contour/endpointstranslator_test.go
```diff
@@ -418,21 +582,24 @@ func TestEndpointsTranslatorScaleToZeroEndpoints(t *testing.T) {
...
-		envoy.ClusterLoadAssignment("default/simple", envoy.SocketAddress("192.168.183.24", 8080)),
+		&v2.ClusterLoadAssignment{
+			ClusterName: "default/simple",
+			Endpoints:   envoy.WeightedEndpoints(1, envoy.SocketAddress("192.168.183.24", 8080)),
+		},
 	}
-	got := et.Contents()
 
-	assert.Equal(t, want, got)
+	protobuf.ExpectEqual(t, want, et.Contents())
```
youngnick, line 591: I don't think using `assert` or `ExpectEqual` here is correct - if there is a failure here, there's no point running the rest of the test. I think it should be `RequireEqual`.

internal/contour/endpointstranslator_test.go
```diff
@@ -418,21 +582,24 @@ func TestEndpointsTranslatorScaleToZeroEndpoints(t *testing.T) {
...
 
 	// Assert endpoints are removed
-	want = nil
-	got = et.Contents()
+	want = []proto.Message{
+		&v2.ClusterLoadAssignment{ClusterName: "default/simple"},
+	}
 
-	assert.Equal(t, want, got)
+	protobuf.ExpectEqual(t, want, et.Contents())
```
youngnick, line 602: Should probably be `RequireEqual` as well, although it won't change anything.

internal/dag/dag.go
youngnick: I agree with renaming to Weighted.

youngnick (approved):
LGTM with a couple of small things.

internal/dag/dag.go
jpeach: Renamed.

PR #2806, opened 2020-08-18
internal/dag: Refactor out DelegationPermitted into the KubernetesCache

stevesloka:
Moves some common components from the builder into the KubernetesCache
so that it can be reused from other components. This is the first of a few PRs in refactoring
the builder out.

Updates #2226

### DIFF
--- a/internal/dag/cache.go
+++ b/internal/dag/cache.go
@@ -425,3 +425,38 @@ func (kc *KubernetesCache) LookupSecret(name types.NamespacedName, validate func
 
 	return s, nil
 }
+
+// DelegationPermitted returns true if the referenced secret has been delegated
+// to the namespace where the ingress object is located.
+func (kc *KubernetesCache) DelegationPermitted(secret types.NamespacedName, targetNamespace string) bool {
+	contains := func(haystack []string, needle string) bool {
+		if len(haystack) == 1 && haystack[0] == "*" {
+			return true
+		}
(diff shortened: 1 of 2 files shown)


internal/dag/cache.go
```diff
@@ -425,3 +425,36 @@ func (kc *KubernetesCache) LookupSecret(name types.NamespacedName, validate func
 
 	return s, nil
 }
+
+func (kc *KubernetesCache) DelegationPermitted(secret types.NamespacedName, to string) bool {
```
jpeach, line 429: Nit: Perhaps `secretName` and `targetNamespace` would be clearer variable names. `secret` works ok, but it took me a few minutes to understand what `to` meant.

jpeach (approved):
Nice improvement.

internal/dag/cache.go
```diff
@@ -425,3 +425,36 @@ func (kc *KubernetesCache) LookupSecret(name types.NamespacedName, validate func
...
+	contains := func(haystack []string, needle string) bool {
```
skriss, line 430: haystack/needle 😂

internal/dag/cache.go
stevesloka: ok cool I renamed. These were just lifted and shifted from the existing code, so wasn't changing much, but internal names won't be a problem.

PR #2807, opened 2020-08-18
internal/dag: Move dag.Builder data into a new struct

stevesloka:
Break out data components from the dag.Builder into its own struct
to seperate the builder cache data apart.

Updates #2226

### DIFF
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -20,16 +20,15 @@ import (
 	"strconv"
 	"strings"
 
-	v1 "k8s.io/api/core/v1"
-	"k8s.io/api/networking/v1beta1"
-	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/intstr"
-
 	"github.com/google/go-cmp/cmp"
 	projcontour "github.com/projectcontour/contour/apis/projectcontour/v1"
 	"github.com/projectcontour/contour/internal/annotation"
 	"github.com/projectcontour/contour/internal/k8s"
 	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/api/networking/v1beta1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 type RouteServiceName struct {
@@ -45,27 +44,23 @@ type Builder struct {
 	// from which to build a DAG.
 	Source KubernetesCache
 
+	// data holds the DAG information after a builder
+	// processes through the Kubernetes cache which is
+	// used to build a new DAG.
+	data BuildContext
+
 	// DisablePermitInsecure disables the use of the
 	// permitInsecure field in HTTPProxy.
 	DisablePermitInsecure bool
 
-	services map[RouteServiceName]*Service
-	secrets  map[types.NamespacedName]*Secret
-
-	virtualhosts       map[string]*VirtualHost
-	securevirtualhosts map[string]*SecureVirtualHost
-
-	orphaned map[types.NamespacedName]bool
-
 	FallbackCertificate *types.NamespacedName
 
-	StatusWriter
 	logrus.FieldLogger
 }
 
 // Build builds a new DAG.
 func (b *Builder) Build() *DAG {
-	b.reset()
+	b.data.reset()
 
 	// setup secure vhosts if there is a matching secret
 	// we do this first so that the set of active secure vhosts is stable
@@ -79,18 +74,6 @@ func (b *Builder) Build() *DAG {
 	return b.buildDAG()
 }
 
-// reset (re)inialises the internal state of the builder.
-func (b *Builder) reset() {
-	b.services = make(map[RouteServiceName]*Service, len(b.services))
-	b.secrets = make(map[types.NamespacedName]*Secret, len(b.secrets))
-	b.orphaned = make(map[types.NamespacedName]bool, len(b.orphaned))
-
-	b.virtualhosts = make(map[string]*VirtualHost)
-	b.securevirtualhosts = make(map[string]*SecureVirtualHost)
-
-	b.statuses = make(map[types.NamespacedName]Status, len(b.statuses))
-}
-
 // lookupService returns a Service that matches the Meta and Port of the Kubernetes' Service,
 // or an error if the service or port can't be located.
 func (b *Builder) lookupService(m types.NamespacedName, port intstr.IntOrString) (*Service, error) {
@@ -99,7 +82,7 @@ func (b *Builder) lookupService(m types.NamespacedName, port intstr.IntOrString)
 			// can't handle, give up
 			return nil
 		}
-		return b.services[RouteServiceName{
+		return b.data.services[RouteServiceName{
 			Name:      m.Name,
 			Namespace: m.Namespace,
 			Port:      int32(port.IntValue()),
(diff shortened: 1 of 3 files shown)


internal/dag/builder.go
```diff
@@ -38,18 +37,15 @@ type Builder struct {
 	// from which to build a DAG.
 	Source KubernetesCache
 
+	// data holds the dag information after a builder
+	// processes through the Kubernetes cache which is
+	// used to build a new DAG.
+	data BuilderData
```
stevesloka, line 43: I'm not in love with this name, but `cache` seemed incorrect as well. I'm looking to use this for the builder refactor so each implementation can be in charge of populating this struct.

jpeach: Maybe `BuildContext`, or `BuilderState`?

internal/dag/builderdata.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/projectcontour/contour/internal/annotation"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// BuilderData holds the information process from ingress
+// resources and is used to build out a new DAG.
+type BuilderData struct {
+	services map[servicemeta]*Service
+	secrets  map[types.NamespacedName]*Secret
```
jpeach, line 30: AFAICT `secrets` is written but never read. Can we remove it?

jpeach (commented):
I'm not sure about this approach. If it's an intermediate step, then I don't see the harm, but I would expect that in the end each Builder would end up keeping its own state that is relevant ti its build process (i.e. it would embed a builder-specific data object).

stevesloka:
> I'm not sure about this approach. If it's an intermediate step, then I don't see the harm, but I would expect that in the end each Builder would end up keeping its own state that is relevant ti its build process (i.e. it would embed a builder-specific data object).

This is worth figuring out now. =) 

My thought was that each "processor" (i.e. Ingress, HTTPProxy, Service-APIs, etc) would implement their own logic to fill out info in the `BuilderData` struct. That's what's needed to build out the DAG, so the interface they need to implement returns a `BuilderData` struct at the end of it's processing. Probably some errors as well should be thought to work in. 

The initial bits was that I was trying to apply these changes slowly. By refactoring out into this struct, next I can introduce the interface and start to pick apart the other builder types. 

I was also keeping the same "serial" logic that @youngnick described on the community call today whereas HTTPProxy take precedence over Ingress objects, etc. So if each "processer" returns a `BuilderData` struct, then the `builder.go` package needs to call each implemented processor and then merge together the results like it does today. 

That is unless we want to change that logic which I don't think we're ready to do just yet (at least on the `main` branch).

jpeach:
> > I'm not sure about this approach. If it's an intermediate step, then I don't see the harm, but I would expect that in the end each Builder would end up keeping its own state that is relevant ti its build process (i.e. it would embed a builder-specific data object).
> 
> This is worth figuring out now. =)

👍 

> My thought was that each "processor" (i.e. Ingress, HTTPProxy, Service-APIs, etc) would implement their own logic to fill out info in the `BuilderData` struct. That's what's needed to build out the DAG, so the interface they need to implement returns a `BuilderData` struct at the end of it's processing. Probably some errors as well should be thought to work in.
> 
> The initial bits was that I was trying to apply these changes slowly. By refactoring out into this struct, next I can introduce the interface and start to pick apart the other builder types.
> 
> I was also keeping the same "serial" logic that @youngnick described on the community call today whereas HTTPProxy take precedence over Ingress objects, etc. So if each "processer" returns a `BuilderData` struct, then the `builder.go` package needs to call each implemented processor and then merge together the results like it does today.

One way to approach this would be to run each processor in parallel, have each one emit a DAG and then merge. That's probably a bit more involved that we want to deal with now though :)

In the system that you describe here, it sounds something like this:

```
d dag.DAG

for _, p := range processors {
    p.Build(&d)
}
```

This preserves the ordering of processors, and assumes that conflicts can be detected by probing the DAG (which I think is reasonable?). Most of the state in the `BuilderData` is local to the build pass, except the conflict detection. Conflict detection can be supported with a collection of DAG helpers like:

```Go
package dag

// Probably a bit too complicated ¯\_(ツ)_/¯

type Matcher func(*Vertex)bool

func MatchVirtualHost(name string) Matcher {
        ...
}

func Find(...) interface{} {
}
```

stevesloka:
Yup I think we're on the same page. My thought right now wasn't to have each processor return a DAG, but to have it return a `BuilderData` struct. Again this matches what we're doing today with how the builder is setup. 

```
type BuildProcessor interface {
	Build(cache *KubernetesCache) BuilderData
}
```

That makes builder look something like this;
```
func (b *Builder) Build() *DAG {

bd BuilderData

for _, p := range processors {
    builderData := p.Build(&k8sCache)

    // merge the data with the existing builder set
   bd.Merge(builderData)
}

return b.buildDAG()
}
```

skriss:
Couple general questions:

In the target state, is `Merge(...)` basically a matter of taking virtualhosts/securevirtualhosts from all of the different processors, and attaching them to the HTTP/HTTPS listener? Or is it more complex than that?

What kinds of conflicts will we have to worry about across processors? Are we just looking for duplicate vhosts/securevhosts being returned from different processors (i.e. `foo.bar.com` is described by both `HTTPProxy` and `Ingress`) or are there other conflicts that could occur as well?

stevesloka:
> In the target state, is Merge(...) basically a matter of taking virtualhosts/securevirtualhosts from all of the different processors, and attaching them to the HTTP/HTTPS listener? Or is it more complex than that?

Yeah that's what it does now. I was trying to not change too much to get this enabled, then look to refactor more out later to minimize the impact. 

> What kinds of conflicts will we have to worry about across processors? Are we just looking for duplicate vhosts/securevhosts being returned from different processors (i.e. foo.bar.com is described by both HTTPProxy and Ingress) or are there other conflicts that could occur as well?

I'm pretty sure the "merge" logic is that HTTPProxy takes precedence over an v1beta1.Ingress object. We should determine how to handle this in the future, I think some of the logic can get a bit weird. It might just simpler to only accept one at a time, but today having v1beta1.Ingress allows us to do cert-manager & external-dns integrations easily until those get full support. But we'll run into the same issues with service-apis as they are so new. 

@jpeach & I had a Slack convo about this a bit and not using the `BuilderData` as the return type but rather a `dag.DAG`. I'm going to investigate that a bit further to see how that might look.

skriss:
> @jpeach & I had a Slack convo about this a bit and not using the BuilderData as the return type but rather a dag.DAG. I'm going to investigate that a bit further to see how that might look.

Cool, just catching up on that. Sounds like you were already discussing along these lines, but I'm wondering if they (BuilderData and DAG) can be merged into a single data type, since it's really two different views on the same information. One API for builders, a different API for consumers, but a common underlying data structure.

jpeach:
> I'm pretty sure the "merge" logic is that HTTPProxy takes precedence over an v1beta1.Ingress object. We should determine how to handle this in the future, I think some of the logic can get a bit weird. It might just simpler to only accept one at a time, but today having v1beta1.Ingress allows us to do cert-manager & external-dns integrations easily until those get full support. But we'll run into the same issues with service-apis as they are so new.

For the merge to be able to prioritize conflicts, the build processor would need to have an explicit priority field, which might be a bit awkward.

jpeach:
> > @jpeach & I had a Slack convo about this a bit and not using the BuilderData as the return type but rather a dag.DAG. I'm going to investigate that a bit further to see how that might look.
> 
> Cool, just catching up on that. Sounds like you were already discussing along these lines, but I'm wondering if they (BuilderData and DAG) can be merged into a single data type, since it's really two different views on the same information. One API for builders, a different API for consumers, but a common underlying data structure.

I think that in the end, we can have everything operate directly on the DAG.

youngnick:
I think keeping the serial aspect is important, because some objects need other objects to be present before they can be meaningfully processed for the DAG. We may end up having to run multiple times if we tried to do things in parallel...

stevesloka:
Going to close this out since the PRs from @skriss were merged for the builder changes. We can revisit this refactoring later on.

PR #2808, opened 2020-08-18
#2589 - add groupversion for logging

ffahri:
Signed-off-by: Fahri Yardımcı 

Tested on minikube v1.12.3  - Kubernetes v1.17.11 on Docker 19.03.8

logs:
``` 
time="2020-08-18T20:50:34Z" level=info msg="started HTTP server" address="127.0.0.1:6060" context=debugsvc
time="2020-08-18T20:50:34Z" level=error msg="ignoring invalid or unsupported annotation" annotation=projectcontour.io/wololo context=KubernetesCache kind=Ingress name=kuard namespace=default version=networking.k8s.io/v1beta1
time="2020-08-18T20:50:34Z" level=error msg="ignoring invalid or unsupported annotation" annotation=projectcontour.io/wololo context=KubernetesCache kind=HTTPProxy name=basic namespace=default version=projectcontour.io/v1
```

### DIFF
--- a/internal/k8s/kind.go
+++ b/internal/k8s/kind.go
@@ -18,6 +18,8 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/api/networking/v1beta1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
 )
 
 // KindOf returns the kind string for the given Kubernetes object.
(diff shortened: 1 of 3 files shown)


internal/k8s/kind.go
```diff
@@ -26,22 +25,30 @@ import (
...
-		return "Ingress"
-	case *projectcontour.HTTPProxy:
-		return "HTTPProxy"
-	case *projectcontour.TLSCertificateDelegation:
-		return "TLSCertificateDelegation"
-	case *unstructured.Unstructured:
-		return obj.GetKind()
-	default:
+	sch := scheme.Scheme
+	_ = projectcontour.AddToScheme(sch)
```
ffahri, line 29: this will help us add current contour types however I'm not sure how performant this approach is ? We may pass the schema from upper but does it create a lot of issue?

ffahri: This help us to not include each object type for contour (especially v1- v1beta1 etc in future maybe)

ffahri:
got a lot of test errors in action will check them

internal/k8s/kind.go
ffahri: I changed to addToSchema since it messed up concurrency in tests. However im still asking this question :smile: 

with this approach we dont need to add each type to our switch case (especially everytime Contour supports another type)

youngnick: Perhaps we could provide an 'all types registered' schema as an export of the `k8s` package - this would be useful in a number of places, and would be a good place to add more types as we go.

jpeach: Yep, +1 to exposing a universal Contour scheme from the `k8s` package.

internal/k8s/kind.go
```diff
@@ -26,22 +28,56 @@ import (
...
+	}
+	return ""
+}
+
+// VersionOf returns the GroupVersion string for the given Kubernetes object.
+//
+func VersionOf(obj interface{}) string {
+	//If err not nil we have the GVK and we can use it
+	//Else we have an error and we're going to use switch-case method as failover
+	//Some of the objects ex: HTTPProxy will get NotRegisteredErrForType errors so we have to use Else block
```
jpeach, line 63: Nit, could you please format this as a single paragraph:

```suggestion
	// If err is not nil we have the GVK and we can use it. Otherwise
	// we have an error and we're going to use switch-case method
	// as failover.  Some of the objects, e.g. HTTPProxy, will get
	// NotRegisteredErrForType errors so we have to the type assertion.

```

jpeach (approved):
Looks great, pending nits.

If you could update the commit message, following the [template](https://github.com/projectcontour/contour/blob/main/CONTRIBUTING.md#commit-message-template, that would be great, but I can also fix it when I merge the PR 👍

internal/k8s/kind.go
ffahri: done, also added the commit template

internal/k8s/kind.go
ffahri: So which one should i go with it , exposing universal schema or create an issue for it :question:

youngnick: Let's get your change in, if you could log an issue for adding the schema, that would be great!

youngnick (approved):
LGTM, thanks!

jpeach (approved):
Thanks!

internal/k8s/kind.go
ffahri: https://github.com/projectcontour/contour/issues/2818
created a issue for universal Contour types

PR #2791, opened 2020-08-13
[WIP] add Gatekeeper guide and example configs

skriss:
Signed-off-by: Steve Kriss <krisss@vmware.com>

Still a WIP, but putting it up to get any early feedback on general structure.

Some things I still need to do:

- [ ] add more example constraints
- [ ] some more testing of constraints
- [ ] test the guide again
- [ ] any other spots in docs to add refs to the Gatekeeper config?

Issue #2742, opened 2020-07-28
Some admission control support for validating local issues


bgagnon:
For user errors than can be validated without resolving other nodes in the DAG, admission control would provide a better user experience than the current asynchronous system of flagging the proxy object as "invalid".

Some validation points require late binding and couldn't be covered here:
* TLS secret existence or validity of contents
* upstream service existence
* inclusions and delegations
* etc.

A quick list of validation points not covered by the JSON schema but still local to a single object:

* malformed FQDN (too long, misplaced dots, etc.)
* out of range timeout values (#2690, #2728)
* malformed or illogical routes/services stanzas

cc @stevesloka

bgagnon:
Also, I think a DAG-aware admission controller is possible if Contour itself is serving the admission controller webhook, but I'm not sure it's a good idea to force DAG nodes to be created in a topological order to avoid admission errors. This isn't true of normal Kubernetes objects in general (eg. `Service` and `Pod` can be created in any order and eventually find each other).

jpeach:
How far can you get with https://github.com/open-policy-agent/gatekeeper ?

youngnick:
I think that admission control of some sort is inevitable. It's so much better a user experience to be made aware of invalid objects as soon as possible.

The details we need to figure out are:

## What admission control do we use?

As @jpeach says, many use cases could be hit using Gatekeeper, and that would also give advanced users the flexibility of customising the admission control to their liking. 

A DAG-aware admission controller would need to run as part of Contour, I agree, in a similar way to what kubebuilder controllers now do. I think this is much less necessary, given that we (well, I) will implement Conditions support into status, which should at least provide a way to surface the same information.

## When do we do this?

Again, the advantage of using a general admisson controller like Gatekeeper (and providing sample policies) is that we will be able to deliver some value sooner.

Building something into Contour will take longer.

This is a great discussion for us to have one community meeting! @bgagnon, if you could make a community meeting, you can give us a shout, and we will add an item to the agenda.

jpeach:
> A DAG-aware admission controller would need to run as part of Contour, I agree, in a similar way to what kubebuilder controllers now do. I think this is much less necessary, given that we (well, I) will implement Conditions support into status, which should at least provide a way to surface the same information.

IIUC, you can't really do admission control against more than one resource, so there's no real benefit to being DAG aware.

stevesloka:
> so there's no real benefit to being DAG aware

[...]

Issue #2032, opened 2019-12-10
site proof job checks the external url not the pr url


davecheney:
https://travis-ci.com/davecheney/contour/jobs/265781824?utm_medium=notification&utm_source=email

The site proof job checks the external url, which means files which reference other files added in a PR will fail the proof job.

jpeach:
```
- ./_site/docs/v1.0.1/README/index.html
546  *  External link https://projectcontour.io/docs/v1.0.1/README/ failed: 404 No error
547- ./_site/docs/v1.0.1/annotations/index.html
548  *  External link https://projectcontour.io/docs/v1.0.1/annotations/ failed: 404 No error
549- ./_site/docs/v1.0.1/architecture/index.html
550  *  External link https://projectcontour.io/docs/v1.0.1/architecture/ failed: 404 No error
551- ./_site/docs/v1.0.1/configuration/index.html
552  *  External link https://projectcontour.io/docs/v1.0.1/configuration/ failed: 404 No error
```

Hmm, maybe there's a way to know what the netlify URL for the PR build will be?

skriss:
With the migration to Hugo we dropped the siteproof job, closing this out.

PR #2814, opened 2020-08-19
add Gatekeeper guide and sample configs

skriss:

replaces #2791 
updates #2742 

To do:
- [x] some more testing of constraints
- [x] test the guide again
- [ ] add more example constraints (to be done in separate PRs)
- [ ] any other spots in docs to add refs to the Gatekeeper config?

### DIFF
--- a/examples/gatekeeper/01-sync-config.yml
+++ b/examples/gatekeeper/01-sync-config.yml
@@ -1,0 +1,17 @@
+---
+# This config tells Gatekeeper to keep all HTTPProxy
+# resources in its internal cache so pre-existing
+# resources can be referenced in Constraints. See
+# https://github.com/open-policy-agent/gatekeeper#replicating-data
+# for more info.
+apiVersion: config.gatekeeper.sh/v1alpha1
+kind: Config
+metadata:
+  name: config
+  namespace: "gatekeeper-system"
+spec:
+  sync:
+    syncOnly:
+      - group: "projectcontour.io"
+        version: "v1"
+        kind: "HTTPProxy"
--- a/site/_guides/gatekeeper.md
+++ b/site/_guides/gatekeeper.md
@@ -1,0 +1,457 @@
+---
+title: Using Gatekeeper as a validating admission controller with Contour
+layout: page
+---
+
+This tutorial demonstrates how to use [Gatekeeper](https://github.com/open-policy-agent/gatekeeper) as a validating admission controller for Contour.
+
+Gatekeeper is a project that enables users to define flexible policies for Kubernetes resources using [Open Policy Agent (OPA)](https://www.openpolicyagent.org/) that are enforced when those resources are created/updated via the Kubernetes API.
+
+The benefits of using Gatekeeper with Contour are:
+- Immediate feedback for the user when they try to create an `HTTPProxy` with an invalid spec. Instead of having to check the `HTTPProxy`'s status after creation for a possible error message, the create is rejected and the user is immediately provided with a reason for the rejection.
+- User-defined policies for `HTTPProxy` specs. For example, the Contour admin can define policies to enforce maximum limits on timeouts and retries, disallow certain FQDNs, etc.
+
+## Prerequisites
+
+- A Kubernetes cluster with a minimum version of 1.14 (to enable webhook timeouts for Gatekeeper).
+- Cluster-admin permissions
+
+## Deploy Contour
+
+Run:
+
+```bash
+$ kubectl apply -f {{ site.url }}/quickstart/contour.yaml
+```
+
+This creates a `projectcontour` namespace and sets up Contour as a deployment and Envoy as a daemonset, with communication between them secured by mutual TLS.
+
+Check the status of the Contour pods with this command:
+
+```bash
+$ kubectl -n projectcontour get pods -l app=contour
+NAME                           READY   STATUS      RESTARTS   AGE
+contour-8596d6dbd7-9nrg2       1/1     Running     0          32m
+contour-8596d6dbd7-mmtc8       1/1     Running     0          32m
+```
+
+If installation was successful, all pods should reach `Running` status shortly.
+
+## Deploy Gatekeeper
+
+The following instructions are summarized from the [Gatekeeper documentation](https://github.com/open-policy-agent/gatekeeper#installation-instructions).
+If you already have Gatekeeper running in your cluster, you can skip this section.
+
+Run:
+
+```bash
+$ kubectl apply -f https://raw.githubusercontent.com/open-policy-agent/gatekeeper/master/deploy/gatekeeper.yaml
+```
+
+This creates a `gatekeeper-system` namespace and sets up the Gatekeeper controller manager and audit deployments using the latest Gatekeeper release.
+
+Check the status of the Gatekeeper pods with this command:
+
+```bash
+$ kubectl -n gatekeeper-system get pods
+NAME                                             READY   STATUS    RESTARTS   AGE
+gatekeeper-audit-67dfc46db6-kjcmc                1/1     Running   0          40m
+gatekeeper-controller-manager-7cbc758844-64hhn   1/1     Running   0          40m
+gatekeeper-controller-manager-7cbc758844-c4dkd   1/1     Running   0          40m
+gatekeeper-controller-manager-7cbc758844-xv9jn   1/1     Running   0          40m
+```
+
+If installation was successful, all pods should reach `Running` status shortly.
+
+## Configure Gatekeeper
+
+### Background
+
+Gatekeeper uses the [OPA Constraint Framework](https://github.com/open-policy-agent/frameworks/tree/master/constraint) to define and enforce policies.
+This framework has two key types: `ConstraintTemplate` and `Constraint`.
+A `ConstraintTemplate` defines a reusable OPA policy, along with the parameters that can be passed to it when it is instantiated.
+When a `ConstraintTemplate` is created, Gatekeeper automatically creates a custom resource definition (CRD) to represent it in the cluster.
+
+A `Constraint` is an instantiation of a `ConstraintTemplate`, which tells Gatekeeper to apply it to specific Kubernetes resource types (e.g. `HTTPProxy`) and provides any relevant parameter values.
+A `Constraint` is defined as an instance of the CRD representing the associated `ConstraintTemplate`.
+
+We'll now look at some examples to make these concepts concrete.
+
+### Configure resource caching
+
+First, Gatekeeper needs to be configured to store all `HTTPProxy` resources in its internal cache, so that existing `HTTPProxy` resources can be referenced within constraint template policies.
+This is essential for being able to define constraints that look across all `HTTPProxies` -- for example, to verify FQDN uniqueness.
+
+Create a file called `config.yml` containing the following YAML:
+
+```yaml
+apiVersion: config.gatekeeper.sh/v1alpha1
+kind: Config
+metadata:
+  name: config
+  namespace: "gatekeeper-system"
+spec:
+  sync:
+    syncOnly:
+      - group: "projectcontour.io"
+        version: "v1"
+        kind: "HTTPProxy"
+```
+
+Apply it to the cluster:
+
+```bash
+$ kubectl apply -f config.yml
+```
+
+Note that if you already had Gatekeeper running in your cluster, you may already have the `Config` resource defined.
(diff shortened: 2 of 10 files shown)


skriss:
Siteproof job failing due to #2032 (links to pages being added in the current PR fail because siteproof checks the external URL).

site/_guides/gatekeeper.md
```diff
@@ -0,0 +1,406 @@
...
+
+A `Constraint` is an instantiation of a `ConstraintTemplate`, which tells Gatekeeper to apply it to specific Kubernetes resource types (e.g. `HTTPProxy`) and provides any relevant parameter values.
+A `Constraint` is defined as an instance of the CRD representing the associated `ConstraintTemplate`.
+
+We'll now look at some examples to make these concepts concrete.
+
+### Configure resource caching
+
+First, Gatekeeper needs to be configured to store all `HTTPProxy` resources in its internal cache, so that existing `HTTPProxy` resources can be referenced within constraint template policies.
+This is essential for being able to define constraints that look across all `HTTPProxies` -- for example, to verify FQDN uniqueness.
```
bgagnon, line 82: This is really interesting! I had assumed that Gatekeeper could only see the resource being admitted, plus some contextual data needed for ACLs. This idea of caching resources opens up a whole world of possibilities.

site/_guides/gatekeeper.md
```diff
@@ -0,0 +1,406 @@
...
+        plural: HTTPProxyUniqueFQDNs
+        singular: HTTPProxyUniqueFQDN
+  targets:
+    - target: admission.k8s.gatekeeper.sh
+      rego: |
+        package httpproxy.uniquefqdn
+
+        violation[{"msg": msg, "other": sprintf("%v/%v", [other.metadata.namespace, other.metadata.name])}] {
+          got := input.review.object.spec.virtualhost.fqdn
+          other := data.inventory.namespace[_]["projectcontour.io/v1"]["HTTPProxy"][_]
```
bgagnon, line 134: Since Contour is especially good in multi-tenant situations, I'd be interested to see more mentions of namespaces in the Gatekeeper/Contour presentation.

* are constraints namespace-specific by default; can they be made cluster wide
* how constraints at the cluster level interact with namespace-level ones
* what is the order of precedence
* etc.

site/_guides/gatekeeper.md
```diff
@@ -0,0 +1,406 @@
...
+httpproxyretrycountrange.constraints.gatekeeper.sh   2020-08-13T16:12:10Z
+```
+
+Now, create a file called `retry-count-range-constraint.yml` containing the following YAML:
+
+```yaml
+apiVersion: constraints.gatekeeper.sh/v1beta1
+kind: HTTPProxyRetryCountRange
+metadata:
+  name: httpproxy-retry-count-range
```
bgagnon, line 293: I think the example should specify `metadata.namespace` field explicitly and a comment below should make the scoping of the constraint extra clear. In other words, if I create this `HTTPProxyRetryCountRange` in namespace `foo`, does it only apply to `HTTPProxy` objects in that namespace? Does it depend on the content of the policy?

site/_guides/gatekeeper.md
skriss: That's great feedback. AFAIK, both `ConstraintTemplates` and `Constraints` must be cluster-scoped, and there's not any other way in Gatekeeper to specify that constraints only apply to specific namespaces. I'll double-check on that to make sure I'm correct but I believe that's the case.

However, you *can* add logic to the rego to effectively make certain constraints namespace-specific -- you'd add a `namespace` parameter to the `ConstraintTemplate`, and then in the rego, you'd add logic to only check `HTTPProxy` objects in that namespace (if specified).

I'll look at getting an example or two of this set up and covering it in the guide. Thanks again for the review!

skriss:
Just pushed a new commit with some edits to demonstrate namespace-scoped Constraints, but I'll re-review and test in the morning.

youngnick (approved):
LGTM, assuming others are okay.

examples/gatekeeper/04-policy-constraint-templates.yml
```diff
@@ -0,0 +1,112 @@
...
+          input.review.object.spec.routes[_].timeoutPolicy[input.parameters.field] == "infinity" 
+        } else = true {
+          # otherwise, compare durations
+          got := input.review.object.spec.routes[_].timeoutPolicy[input.parameters.field]
+          time.parse_duration_ns(got) < time.parse_duration_ns(input.parameters.min)
+        }
+
+        violation[{"msg": msg}] {
+          above_max
+          msg := sprintf("timeout policy field '%s' must have value less than/equal to %v", [input.parameters.field, input.parameters.max])
```
jpeach, line 106: I think that this would make the validation more readable:

```
msg := sprintf("%q timeout must have value less than or equal to %q", [input.parameters.field, input.parameters.max])
```

examples/gatekeeper/04-policy-constraint-templates.yml
```diff
@@ -0,0 +1,112 @@
...
+      rego: |
+        package httpproxy.timeoutrange
+
+        above_max = true {
+          # special-case for when field value is "infinity" -- max must be unspecified
+          input.review.object.spec.routes[_].timeoutPolicy[input.parameters.field] == "infinity"
+          input.parameters.max
+        } else = true {
+          # otherwise, compare durations 
+          got := input.review.object.spec.routes[_].timeoutPolicy[input.parameters.field]
```
jpeach, line 91: Suggest you hoist this up to package scope so you doesn't need to be repeated in all the rules.

examples/gatekeeper/04-policy-constraint-templates.yml
```diff
@@ -0,0 +1,112 @@
...
+        }
+---
+# httpproxytimeoutrange is a ConstraintTemplate that enforces
+# min/max allowed timeout values for a specified field within
+# a route's timeoutPolicy on an HTTPProxy. Both min and max are
+# optional. 
+apiVersion: templates.gatekeeper.sh/v1beta1
+kind: ConstraintTemplate
+metadata:
+  name: httpproxytimeoutrange
```
jpeach, line 59: hyphenate?

examples/gatekeeper/04-policy-constraint-templates.yml
```diff
@@ -0,0 +1,112 @@
...
+        kind: HTTPProxyTimeoutRange
+        listKind: HTTPProxyTimeoutRangeList
+        plural: HTTPProxyTimeoutRanges
+        singular: HTTPProxyTimeoutRange
+      scope: Namespaced
+      validation:
+        openAPIV3Schema:
+          properties:
+            field:
+              type: string
```
jpeach, line 75: Is it syntactically possible to apply an enum constraint on this field to enforce getting the correct field name? Could be a separate PR.

examples/gatekeeper/04-policy-constraint-templates.yml
```diff
@@ -0,0 +1,112 @@
...
+
+          got > input.parameters.max
+          msg := sprintf("retry policy count must be less than/equal to %v", [input.parameters.max])
+        }
+
+        violation[{"msg": msg}] {
+          got := input.review.object.spec.routes[_].retryPolicy.count
+
+          got < input.parameters.min
+          msg := sprintf("retry policy count must be greater than/equal to %v", [input.parameters.min])
```
jpeach, line 49: Suggest:
```
msg := sprintf("retry count must be greater than or equal to %q", [input.parameters.min])
```

examples/gatekeeper/04-policy-constraint-templates.yml
```diff
@@ -0,0 +1,112 @@
...
+
+        violation[{"msg": msg}] {
+          got := input.review.object.spec.routes[_].retryPolicy.count
+
+          got > input.parameters.max
+          msg := sprintf("retry policy count must be less than/equal to %v", [input.parameters.max])
+        }
+
+        violation[{"msg": msg}] {
+          got := input.review.object.spec.routes[_].retryPolicy.count
```
jpeach, line 46: Suggest hoisting this out of the rule?

site/_guides/gatekeeper.md
```diff
@@ -0,0 +1,450 @@
...
+        Message:             retry policy count must be less than/equal to 4
+        Name:                demo-retries
+        Namespace:           default
+```
+
+However, our `HTTPProxy` remains in the cluster and can continue to route requests, and the user can remediate the proxy to bring it inline with the policy on their own timeline.
+
+## Next steps
+
+Contour has a [growing library](https://github.com/projectcontour/contour/tree/master/examples/gatekeeper) of Gatekeeper constraint templates and constraints, for both **validations** and **policies**.
```
jpeach, line 439: I really like the idea of the project hosting a growing collection of templates and constraints. What I would suggest to make this more manageable is to separate each YAML object into a separate file. I think that your classification of validations and policies makes sense, so we can continue it, maybe like this:

```
002-validation-template-Foo-Name.yaml
002-validation-template-Bar-Name.yaml
003-validation-constraint-Foo.yaml
003-validation-constraint-Bar.yaml
004-policy-template-A.yaml
004-policy-template-B.yaml
```

This preserves the classification, but makes it a bit easier to pick and choose.

jpeach (approved):
Overall, I think that this looks great and should be pretty helpful for people who are interested in admission control polices.

I made a few suggestions, but there's no need to block the PR on them.

examples/gatekeeper/01-sync-config.yml
```diff
@@ -0,0 +1,17 @@
+---
+# This config tells Gatekeeper to keep all HTTPProxy
+# resources in its internal cache so pre-existing
+# resources can be referenced in Constraints. See
+# https://github.com/open-policy-agent/gatekeeper#replicating-data
+# for more info.
+apiVersion: config.gatekeeper.sh/v1alpha1
+kind: Config
+metadata:
+  name: config
```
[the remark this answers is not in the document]
skriss: Unfortunately `Config` is a singleton resource for gatekeeper - see https://github.com/open-policy-agent/gatekeeper/blob/master/pkg/controller/config/config_controller.go#L193-L196 - and the one instance of it has to be named `config`.

examples/gatekeeper/02-validation-constraint-templates.yml
```diff
@@ -0,0 +1,44 @@
...
+
+---
+# httpproxyuniquefqdn is a ConstraintTemplate that requires
+# an HTTPProxy's spec.virtualhost.fqdn to not already be in 
+# use by another HTTPProxy. This ConstraintTemplate requires 
+# HTTPProxy to be in the Gatekeeper cache.
+apiVersion: templates.gatekeeper.sh/v1beta1
+kind: ConstraintTemplate
+metadata:
+  name: httpproxyuniquefqdn
```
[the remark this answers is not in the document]
skriss: It *could*, but Gatekeeper requires that the name of the `ConstraintTemplate` be the lowercase equivalent of the `Kind` of the CRD that it generates. So if we hyphenate the names, then the `Kind` names have to be hyphenated as well, which is very nonstandard for Kubernetes kinds.

examples/gatekeeper/01-sync-config.yml
```diff
@@ -0,0 +1,17 @@
...
+  namespace: "gatekeeper-system"
```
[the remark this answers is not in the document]
skriss: See https://github.com/projectcontour/contour/pull/2814#discussion_r477420409

examples/gatekeeper/02-validation-constraint-templates.yml
```diff
@@ -0,0 +1,44 @@
...
+  labels:
+    app: contour
```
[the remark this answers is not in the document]
skriss: yep!

examples/gatekeeper/04-policy-constraint-templates.yml
skriss: I think it should be possible, I'll look into it

skriss:
Addressed most comments, will run through one more round of testing before being ready to merge.

examples/gatekeeper/04-policy-constraint-templates.yml
skriss: Done (had to construct a set since this uses iteration, rather than just a scalar variable)

site/_guides/gatekeeper.md
skriss: I ended up creating two subdirectories, one for `validations` and one for `policies`, and then split things out to a resource per file, per your suggestion.

PR #2820, opened 2020-08-20
cmd/contour: add drain-delay arg to the envoy shutdown command

### LINKED ISSUE #2819: Add an optional delay before draining envoy connection during shutdown
When the envoy pods are deployed as headless, during its shutdown, we have to wait the kubernetes endpoints update before draining the connections. This is usually done by adding a sleep command to the preStop hook. As contour container image is made from scratch and sleep command is not available, this is not possible.
So, it would be very useful if the 'contour envoy shutdown' command could have a drain-delay parameter. The default behavior would stay the same: no delay.

laurovenancio:
Fixes #2819
Depends on #2817

Signed-off-by: Lauro Ramos Venancio <lauro.venancio@inloco.com.br>

### DIFF
--- a/cmd/contour/shutdownmanager.go
+++ b/cmd/contour/shutdownmanager.go
@@ -57,6 +57,9 @@ type shutdownContext struct {
 	// checkDelay defines time to wait before polling Envoy for open connections
 	checkDelay time.Duration
 
+	// drainDelay defines time to wait before draining Envoy connections
+	drainDelay time.Duration
+
 	// minOpenConnections defines the minimum amount of connections
 	// that can be open when polling for active connections in Envoy
 	minOpenConnections int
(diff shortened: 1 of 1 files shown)


cmd/contour/shutdownmanager.go
```diff
@@ -239,6 +245,7 @@ func registerShutdown(cmd *kingpin.CmdClause, log logrus.FieldLogger) (*kingpin.
 	shutdown := cmd.Command("shutdown", "Initiate an shutdown sequence which configures Envoy to begin draining connections.")
 	shutdown.Flag("check-interval", "Time to poll Envoy for open connections.").DurationVar(&ctx.checkInterval)
 	shutdown.Flag("check-delay", "Time wait before polling Envoy for open connections.").Default("60s").DurationVar(&ctx.checkDelay)
+	shutdown.Flag("drain-delay", "Time wait before draining Envoy connections.").Default("0s").DurationVar(&ctx.drainDelay)
```
skriss, line 248: nit: "Time to wait..." (it'd be great if you could update the line just above as well since it has the same typo)

youngnick (approved):
LGTM, nice work.

PR #2823, opened 2020-08-23
internal/featuretests: remove check helper

jpeach:
The featuretests `check` helper is the same as the testify `NoError`
helper. Now that we have a testify dependency, we can just use that.

### DIFF
--- a/internal/featuretests/featuretests.go
+++ b/internal/featuretests/featuretests.go
@@ -39,6 +39,7 @@ import (
 	"github.com/projectcontour/contour/internal/workgroup"
 	"github.com/projectcontour/contour/internal/xds"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


youngnick (changes requested):
We're already using the internal `assert`, and I think I'm going to change the imported functions there to alias the `require` versions instead - this has always been how we do testing, and it's probably better not to change the contract. Once I do that, you can use the aliased `assert.NoError` from the internal package, and be using a single assert package everywhere.

jpeach:
> We're already using the internal `assert`, and I think I'm going to change the imported functions there to alias the `require` versions instead - this has always been how we do testing, and it's probably better not to change the contract. Once I do that, you can use the aliased `assert.NoError` from the internal package, and be using a single assert package everywhere.

As discussed on slack, we can just use `testify` directly and start removing the `internal/assert` package.

youngnick (approved):
LGTM

Issue #2743, opened 2020-07-28
SNI integration test is flaky


skriss:
See https://travis-ci.com/github/projectcontour/contour/jobs/365814535

Some relevant output:
```
# reading document from ./_integration/testsuite/httpproxy/002-header-condition-match.yaml

...

ok 46 - running object update check
# Fatal - internal/check/grnqoUyHmVtJ:5: eval_conflict_error: complete rules must not produce multiple outputs
not ok 47 - running Rego check lines 160-172
  ---
  - Message: 'internal/check/grnqoUyHmVtJ:5: eval_conflict_error: complete rules must
      not produce multiple outputs'
    Severity: Fatal
    Timestamp: "2020-07-28T18:28:52.766611607Z"
  
  ...
# skipping
ok 48 - deleting test objects
1..48

...

# reading document from ./_integration/testsuite/httpproxy/003-path-condition-match.yaml

... 

ok 31 - running Rego check lines 98-106
# Error - raised predicate "error_proxy_is_not_valid"
# Error - HTTP proxy for 'conditions.projectcontour.io' is not valid
# Error - status: invalid
# Error - desc: fqdn "conditions.projectcontour.io" is used in multiple HTTPProxies: default/header-conditions, default/path-conditions
not ok 32 - running Rego check lines 108-121
  ---
  - Message: |-
      raised predicate "error_proxy_is_not_valid"
      HTTP proxy for 'conditions.projectcontour.io' is not valid
      status: invalid
      desc: fqdn "conditions.projectcontour.io" is used in multiple HTTPProxies: default/header-conditions, default/path-conditions
    Severity: Error
    Timestamp: "0001-01-01T00:00:00Z"
  
  ...
# Error - internal/check/RaYbFFqqeGji:31: eval_builtin_error: http.send: Get "http://127.0.0.1:9080/": read tcp 127.0.0.1:45072->127.0.0.1:9080: read: connection reset by peer
# Error - internal/check/RaYbFFqqeGji:31: eval_builtin_error: http.send: Get "http://127.0.0.1:9080/": read tcp 127.0.0.1:45076->127.0.0.1:9080: read: connection reset by peer
# Error - internal/check/RaYbFFqqeGji:31: eval_builtin_error: http.send: Get "http://127.0.0.1:9080/": read tcp 127.0.0.1:45080->127.0.0.1:9080: read: connection reset by peer
not ok 33 - running Rego check lines 123-182
  ---
[...]

PR #2825, opened 2020-08-24
testsuite: switch merge-slash test away from httpbin

jpeach:
The `httpbin` container takes a long time to start up, and there's no good
way to figure out how long it should take or when to start the timeout
from. Replace this with the `ingress-conformance-echo` container which
has much better startup time.

This updates #2743.


youngnick (approved):
LGTM

PR #2827, opened 2020-08-24
cmd/contour: Replace check function by log.Fatal()

### LINKED ISSUE #2811: Replace check with Fatal logs

There are quite a few places in Contour that call a local `check` helper that prints to standard error and exits. This approach loses error context and makes it difficult for aggregated log collectors to recognize that this is a fatal error message. We should replace these with fatal level logs.

ShaileshSurya:
There are quite a few places in Contour that call a
local check helper that prints to standard error and exits.
This approach loses error context and makes it difficult for
aggregated log collectors to recognize that this is a fatal error message

Fixes #2811

Signed-off-by: Shailesh Suryawanshi <ss.shailesh28@gmail.com>

### DIFF
--- a/cmd/contour/certgen.go
+++ b/cmd/contour/certgen.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/projectcontour/contour/internal/certgen"
 	"github.com/projectcontour/contour/internal/k8s"
+	"github.com/sirupsen/logrus"
 	kingpin "gopkg.in/alecthomas/kingpin.v2"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes"
@@ -123,7 +124,7 @@ func GenerateCerts(certConfig *certgenConfig) (map[string][]byte, error) {
 }
 
 // OutputCerts outputs the certs in certs as directed by config.
-func OutputCerts(config *certgenConfig, kubeclient *kubernetes.Clientset, certs map[string][]byte) {
+func OutputCerts(config *certgenConfig, kubeclient *kubernetes.Clientset, certs map[string][]byte) error {
 	secrets := []*corev1.Secret{}
 	force := certgen.NoOverwrite
 	if config.Overwrite {
@@ -137,34 +138,46 @@ func OutputCerts(config *certgenConfig, kubeclient *kubernetes.Clientset, certs 
 		case "compact":
 			secrets = certgen.AsSecrets(config.Namespace, certs)
 		default:
-			check(fmt.Errorf("unsupported Secrets format %q", config.Format))
+			return fmt.Errorf("unsupported Secrets format %q", config.Format)
 		}
 	}
 
 	if config.OutputPEM {
 		fmt.Printf("Writing certificates to PEM files in %s/\n", config.OutputDir)
-		check(certgen.WriteCertsPEM(config.OutputDir, certs, force))
+		if err := certgen.WriteCertsPEM(config.OutputDir, certs, force); err != nil {
+			return fmt.Errorf("failed to write certificates to %q: %w", config.OutputDir, err)
+		}
 	}
 
 	if config.OutputYAML {
 		fmt.Printf("Writing %q format Secrets to YAML files in %s/\n", config.Format, config.OutputDir)
-		check(certgen.WriteSecretsYAML(config.OutputDir, secrets, force))
+		if err := certgen.WriteSecretsYAML(config.OutputDir, secrets, force); err != nil {
+			return fmt.Errorf("failed to write Secrets to %q: %w", config.OutputDir, err)
+		}
 	}
 
 	if config.OutputKube {
 		fmt.Printf("Writing %q format Secrets to namespace %q\n", config.Format, config.Namespace)
-		check(certgen.WriteSecretsKube(kubeclient, secrets, force))
+		if err := certgen.WriteSecretsKube(kubeclient, secrets, force); err != nil {
+			return fmt.Errorf("failed to write certificates to %q: %w", config.Namespace, err)
+		}
 	}
+	return nil
 }
 
-func doCertgen(config *certgenConfig) {
+func doCertgen(config *certgenConfig, log logrus.FieldLogger) {
 	generatedCerts, err := GenerateCerts(config)
-	check(err)
+	if err != nil {
+		log.WithError(err).Fatal("failed to generate certificates")
+	}
 
 	clients, err := k8s.NewClients(config.KubeConfig, config.InCluster)
 	if err != nil {
-		check(fmt.Errorf("failed to create Kubernetes client: %w", err))
+		log.WithError(err).Fatalf("failed to create Kubernetes client")
 	}
 
-	OutputCerts(config, clients.ClientSet(), generatedCerts)
+	if oerr := OutputCerts(config, clients.ClientSet(), generatedCerts); oerr != nil {
+		log.WithError(oerr).Fatalf("failed output certificates")
+	}
+
 }
--- a/cmd/contour/contour.go
+++ b/cmd/contour/contour.go
@@ -14,7 +14,6 @@
 package main
 
 import (
-	"fmt"
 	"os"
 
 	resource "github.com/envoyproxy/go-control-plane/pkg/resource/v2"
@@ -76,9 +75,11 @@ func main() {
 	case sdmShutdown.FullCommand():
 		sdmShutdownCtx.shutdownHandler()
 	case bootstrap.FullCommand():
-		check(envoy.WriteBootstrap(bootstrapCtx))
+		if err := envoy.WriteBootstrap(bootstrapCtx); err != nil {
+			log.WithError(err).Fatal("failed to write bootstrap configuration")
+		}
 	case certgenApp.FullCommand():
-		doCertgen(certgenConfig)
+		doCertgen(certgenConfig, log)
 	case cds.FullCommand():
 		stream := client.ClusterStream()
 		watchstream(stream, resource.ClusterType, resources)
@@ -97,13 +98,15 @@ func main() {
 	case serve.FullCommand():
 		// parse args a second time so cli flags are applied
 		// on top of any values sourced from -c's config file.
-		_, err := app.Parse(args)
-		check(err)
+		kingpin.MustParse(app.Parse(args))
+
 		if serveCtx.Debug {
 			log.SetLevel(logrus.DebugLevel)
 		}
 		log.Infof("args: %v", args)
-		check(doServe(log, serveCtx))
+		if err := doServe(log, serveCtx); err != nil {
+			log.WithError(err).Fatal("failed to serve contour")
+		}
 	case version.FullCommand():
 		println(build.PrintBuildInfo())
 	default:
@@ -111,11 +114,4 @@ func main() {
 		os.Exit(2)
 	}
 
-}
-
-func check(err error) {
-	if err != nil {
-		fmt.Fprintln(os.Stderr, err)
-		os.Exit(1)
-	}
 }
--- a/cmd/contour/leadership.go
+++ b/cmd/contour/leadership.go
@@ -77,14 +77,14 @@ func setupLeadershipElection(g *workgroup.Group, log logrus.FieldLogger, ctx *se
 // newLeaderElector creates a new leaderelection.LeaderElector and associated
 // channels by which to observe elections and depositions.
 func newLeaderElector(log logrus.FieldLogger, ctx *serveContext, clients *k8s.Clients) (*leaderelection.LeaderElector, chan struct{}, chan struct{}) {
-
+	log = log.WithField("context", "leaderelection")
 	// leaderOK will block gRPC startup until it's closed.
 	leaderOK := make(chan struct{})
 	// deposed is closed by the leader election callback when
 	// we are deposed as leader so that we can clean up.
 	deposed := make(chan struct{})
 
-	rl := newResourceLock(ctx, clients)
+	rl := newResourceLock(ctx, clients, log)
 
 	// Make the leader elector, ready to be used in the Workgroup.
 	le, err := leaderelection.NewLeaderElector(leaderelection.LeaderElectionConfig{
@@ -107,13 +107,15 @@ func newLeaderElector(log logrus.FieldLogger, ctx *serveContext, clients *k8s.Cl
 			},
 		},
 	})
-	check(err)
+	if err != nil {
+		log.WithError(err).Fatal("failed to create leader elector")
+	}
 	return le, leaderOK, deposed
 }
 
 // newResourceLock creates a new resourcelock.Interface based on the Pod's name,
 // or a uuid if the name cannot be determined.
-func newResourceLock(ctx *serveContext, clients *k8s.Clients) resourcelock.Interface {
+func newResourceLock(ctx *serveContext, clients *k8s.Clients, log logrus.FieldLogger) resourcelock.Interface {
 	resourceLockID, found := os.LookupEnv("POD_NAME")
 	if !found {
 		resourceLockID = uuid.New().String()
@@ -133,6 +135,12 @@ func newResourceLock(ctx *serveContext, clients *k8s.Clients) resourcelock.Inter
 			Identity: resourceLockID,
 		},
 	)
-	check(err)
+	if err != nil {
+		log.WithError(err).
+			WithField("name", ctx.LeaderElectionConfig.Name).
+			WithField("namespace", ctx.LeaderElectionConfig.Namespace).
+			WithField("identity", resourceLockID).
+			Fatal("failed to create new resource lock")
+	}
 	return rl
 }
--- a/cmd/contour/servecontext.go
+++ b/cmd/contour/servecontext.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/projectcontour/contour/internal/contour"
 	"github.com/projectcontour/contour/internal/envoy"
+	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/keepalive"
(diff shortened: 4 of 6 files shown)


ShaileshSurya:
@jpeach Please review this PR as well.

cmd/contour/contour.go
```diff
@@ -97,13 +99,16 @@ func main() {
 	case serve.FullCommand():
 		// parse args a second time so cli flags are applied
 		// on top of any values sourced from -c's config file.
-		_, err := app.Parse(args)
-		check(err)
+		if _, err := app.Parse(args); err != nil {
```
youngnick, line 102: This one should probably use `kingpin.MustParse(app.Parse(args))` instead of a full `if err != nil` construct.

That will have Kingpin exit if the parsing fails.

cmd/contour/certgen.go
```diff
@@ -21,6 +21,7 @@ import (
 
 	"github.com/projectcontour/contour/internal/certgen"
 	"github.com/projectcontour/contour/internal/k8s"
+	log "github.com/sirupsen/logrus"
```
youngnick, line 24: In general, with logrus, we try to pass a `logrus.FieldLogger` down into functions, then add extra fields to the logger in each function. This allows a user to easily see which part of the code is failing.

In this case, I'd recommend:
- creating a new logger in `doCertgen` (just `log := logrus.StandardLogger()` is fine).
- passing that logger into `GenerateCerts` and `OutputCerts`
- adding `context` field that indicates which function the error was in (`generatecerts` or `outputcerts`). This can be done by having the log call be `log.WithField("context", "VALUE").Fatal(` instead of just `log.Fatal`.

This also means you don't need to give `logrus` an import name.

cmd/contour/leadership.go
```diff
@@ -133,6 +136,8 @@ func newResourceLock(ctx *serveContext, clients *k8s.Clients) resourcelock.Inter
 			Identity: resourceLockID,
 		},
 	)
-	check(err)
+	if err != nil {
```
youngnick, line 139: Could you please change `newResourceLock` to take a `log logrus.FieldLogger` param? Then we no longer need to import `logrus` twice.

cmd/contour/servecontext.go
```diff
@@ -320,7 +321,9 @@ func (ctx *serveContext) grpcOptions() []grpc.ServerOption {
 // for tls communication, tlsconfig returns nil.
 func (ctx *serveContext) tlsconfig() *tls.Config {
```
youngnick, line 322: I think maybe changing `tlsconfig` to take a `log logrus.FieldLogger` param is better here.

youngnick (changes requested):
A few changes around cleaning up the way we handle logs, since we're removing the `check()` calls. Let's clean up what we touch when we touch it!

cmd/contour/servecontext.go
ShaileshSurya: done

cmd/contour/certgen.go
ShaileshSurya: Fixed this. Added the standard logger.

cmd/contour/contour.go
youngnick: oops, thanks. You're right. Leave this one as is.

youngnick (approved):
LGTM, nice work.

jpeach (changes requested):
Placeholding until I can review.

cmd/contour/leadership.go
```diff
@@ -133,6 +135,8 @@ func newResourceLock(ctx *serveContext, clients *k8s.Clients) resourcelock.Inter
...
+		log.WithField("context", "newResourceLock").Fatalf("err: %v", err)
```
jpeach, line 139:
```Go
log.WithError(err).
    WithField("name", ctx.LeaderElectionConfig.Name).
    WithField("namespace", ctx.LeaderElectionConfig.Namespace).
    WithField("identity", resourceLockID).
    Fatal("failed to create new resource lock")
```

Note that I've added the conventional `name` and `namespace` fields because this is refereeing to a Kubernetes resource, and an `identity` key to assist debug triage.

cmd/contour/servecontext.go
```diff
@@ -309,7 +310,8 @@ func (ctx *serveContext) grpcOptions() []grpc.ServerOption {
 		}),
 	}
 	if !ctx.PermitInsecureGRPC {
-		tlsconfig := ctx.tlsconfig()
+		log := logrus.StandardLogger()
```
jpeach, line 313: I think that this is another case where we should return the error and have the caller do the logging. So let's make this:
```Go
func (ctx *serveContext) grpcOptions() ([]grpc.ServerOption, error) {
...
}
```

Then the caller can log at an appropriate level.

cmd/contour/servecontext.go
```diff
@@ -349,15 +353,18 @@ func (ctx *serveContext) tlsconfig() *tls.Config {
...
 
 	return &tls.Config{
 		ClientAuth: tls.RequireAndVerifyClientCert,
 		Rand:       rand.Reader,
 		GetConfigForClient: func(*tls.ClientHelloInfo) (*tls.Config, error) {
 			config, err := loadConfig()
-			check(err)
+			if err != nil {
+				log.Fatalf("err: %v", err)
+			}
```
jpeach, line 367: I think that we we can probably just drop this check and pass the error up to the caller.

jpeach (changes requested):

cmd/contour/contour.go
```diff
@@ -97,13 +99,16 @@ func main() {
...
+			log.Fatalf("err: %v", err)
```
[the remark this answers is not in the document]
jpeach: I see that this issue was discussed in a previous comment. The reason to continue to use `MustParse` here is that it is the same error (i.e. invalid flag parsing), so we should handle it the same way.

cmd/contour/servecontext.go
ShaileshSurya: This function  `grpcOptions` do not have any error scenario. So I am skipping to resolve this review comment.

cmd/contour/servecontext.go
ShaileshSurya: Done

cmd/contour/leadership.go
ShaileshSurya: Done

cmd/contour/leadership.go
```diff
@@ -107,13 +107,15 @@ func newLeaderElector(log logrus.FieldLogger, ctx *serveContext, clients *k8s.Cl
 			},
 		},
 	})
-	check(err)
+	if err != nil {
+		log.WithField("context", "NewLeaderElector").Fatalf("err: %v", err)
```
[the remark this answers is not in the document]
ShaileshSurya: Done

cmd/contour/certgen.go
```diff
@@ -137,34 +138,44 @@ func OutputCerts(config *certgenConfig, kubeclient *kubernetes.Clientset, certs
...
-		check(certgen.WriteSecretsKube(kubeclient, secrets, force))
+		if err := certgen.WriteSecretsKube(kubeclient, secrets, force); err != nil {
+			log.WithField("context", "OutputKube").Fatalf("err : %v", err)
+		}
 	}
 }
 
 func doCertgen(config *certgenConfig) {
+
+	log := logrus.StandardLogger()
```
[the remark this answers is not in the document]
ShaileshSurya: Done

cmd/contour/certgen.go
```diff
@@ -137,34 +138,44 @@ func OutputCerts(config *certgenConfig, kubeclient *kubernetes.Clientset, certs
...
 	generatedCerts, err := GenerateCerts(config)
-	check(err)
+	if err != nil {
+		log.WithField("context", "GenerateCerts").Fatalf("err : %v", err)
```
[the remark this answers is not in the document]
ShaileshSurya: Done

cmd/contour/certgen.go
```diff
@@ -137,34 +138,44 @@ func OutputCerts(config *certgenConfig, kubeclient *kubernetes.Clientset, certs
...
+	}
 
 	clients, err := k8s.NewClients(config.KubeConfig, config.InCluster)
 	if err != nil {
-		check(fmt.Errorf("failed to create Kubernetes client: %w", err))
+		log.WithField("context", "kubernetes-NewClients").Fatalf("failed to create Kubernetes client: %v", err)
 	}
 
-	OutputCerts(config, clients.ClientSet(), generatedCerts)
+	OutputCerts(config, clients.ClientSet(), generatedCerts, log)
```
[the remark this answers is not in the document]
ShaileshSurya: Done

cmd/contour/contour.go
```diff
@@ -76,7 +76,9 @@ func main() {
 	case sdmShutdown.FullCommand():
 		sdmShutdownCtx.shutdownHandler()
 	case bootstrap.FullCommand():
-		check(envoy.WriteBootstrap(bootstrapCtx))
+		if err := envoy.WriteBootstrap(bootstrapCtx); err != nil {
+			log.Fatalf("err :%v", err)
```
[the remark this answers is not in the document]
ShaileshSurya: Done

cmd/contour/contour.go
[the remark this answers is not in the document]
ShaileshSurya: Done

cmd/contour/contour.go
```diff
@@ -97,13 +99,16 @@ func main() {
...
+		if _, err := app.Parse(args); err != nil {
+			log.Fatalf("err: %v", err)
+		}
 		if serveCtx.Debug {
 			log.SetLevel(logrus.DebugLevel)
 		}
 		log.Infof("args: %v", args)
-		check(doServe(log, serveCtx))
+		if err := doServe(log, serveCtx); err != nil {
+			log.Fatalf("err: %v", err)
```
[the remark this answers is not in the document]
ShaileshSurya: Done.

cmd/contour/certgen.go
```diff
@@ -137,34 +138,47 @@ func OutputCerts(config *certgenConfig, kubeclient *kubernetes.Clientset, certs
...
+		if err := certgen.WriteCertsPEM(config.OutputDir, certs, force); err != nil {
+			return fmt.Errorf("Writing certificates to PEM files in %s/\n", config.OutputDir)
+		}
 	}
 
 	if config.OutputYAML {
 		fmt.Printf("Writing %q format Secrets to YAML files in %s/\n", config.Format, config.OutputDir)
-		check(certgen.WriteSecretsYAML(config.OutputDir, secrets, force))
+		if err := certgen.WriteSecretsYAML(config.OutputDir, secrets, force); err != nil {
+			return fmt.Errorf("failer to write %q format Secrets to YAML files in %s/\n", config.Format, config.OutputDir)
```
jpeach, line 155: suggest:

```Go
fmt.Errorf("failed to write Secrets to %q: %w", config.OutputDir, err)
```

cmd/contour/certgen.go
```diff
@@ -137,34 +138,47 @@ func OutputCerts(config *certgenConfig, kubeclient *kubernetes.Clientset, certs
...
+		}
 	}
 
 	if config.OutputKube {
 		fmt.Printf("Writing %q format Secrets to namespace %q\n", config.Format, config.Namespace)
-		check(certgen.WriteSecretsKube(kubeclient, secrets, force))
+		if err := certgen.WriteSecretsKube(kubeclient, secrets, force); err != nil {
+			return fmt.Errorf("failed to write  %q format Secrets to namespace %q\n", config.Format, config.Namespace)
```
jpeach, line 162: suggest:

```Go
fmt.Errorf("failed to write certificates to %q: %w", config.Namespace, err)
```

cmd/contour/certgen.go
```diff
@@ -137,34 +138,47 @@ func OutputCerts(config *certgenConfig, kubeclient *kubernetes.Clientset, certs
...
+	return nil
 }
 
-func doCertgen(config *certgenConfig) {
+func doCertgen(config *certgenConfig, log logrus.FieldLogger) {
+
 	generatedCerts, err := GenerateCerts(config)
-	check(err)
+	if err != nil {
+		log.WithError(err).Fatal("failed to generate certs")
```
jpeach, line 172: Nit: we should avoid abbreviations in user-visible messages

```Go
log.WithError(err).Fatal("failed to generate certificates")
```

cmd/contour/certgen.go
```diff
@@ -137,34 +138,47 @@ func OutputCerts(config *certgenConfig, kubeclient *kubernetes.Clientset, certs
...
+	}
 
 	clients, err := k8s.NewClients(config.KubeConfig, config.InCluster)
 	if err != nil {
-		check(fmt.Errorf("failed to create Kubernetes client: %w", err))
+		log.WithError(err).Fatalf("failed to create Kubernetes client: %v", err)
+	}
+
+	if oerr := OutputCerts(config, clients.ClientSet(), generatedCerts, log); oerr != nil {
+		log.WithError(err).Fatalf("failed output certs: %v", err)
```
jpeach, line 181: `WithError` already logs the error, so you can just do:
```Go
log.WithError(err).Fatal("failed to write certificates")
```

cmd/contour/certgen.go
```diff
@@ -137,34 +138,47 @@ func OutputCerts(config *certgenConfig, kubeclient *kubernetes.Clientset, certs
...
 	generatedCerts, err := GenerateCerts(config)
-	check(err)
+	if err != nil {
+		log.WithError(err).Fatal("failed to generate certs")
+	}
 
 	clients, err := k8s.NewClients(config.KubeConfig, config.InCluster)
 	if err != nil {
-		check(fmt.Errorf("failed to create Kubernetes client: %w", err))
+		log.WithError(err).Fatalf("failed to create Kubernetes client: %v", err)
```
jpeach, line 177: `WithError` already logs the error, so you can just do:
```Go
log.WithError(err).Fatal("failed to create Kubernetes client")
```

cmd/contour/servecontext_test.go
```diff
@@ -338,7 +339,8 @@ func TestServeContextCertificateHandling(t *testing.T) {
 	checkFatalErr(t, err)
 
 	// Start a dummy server.
-	opts := ctx.grpcOptions()
+	log := logrus.StandardLogger()
+	opts := ctx.grpcOptions(log)
```
jpeach, line 343: Since this is a test, you can use the test logger to capture any logs:
```Go
opts := ctx.grpcOptions(fixture.NewTestLogger(t))
```

cmd/contour/certgen.go
```diff
@@ -137,34 +138,47 @@ func OutputCerts(config *certgenConfig, kubeclient *kubernetes.Clientset, certs
...
+	}
+
+	if oerr := OutputCerts(config, clients.ClientSet(), generatedCerts, log); oerr != nil {
```
jpeach, line 180: Rename `oerr` to `err`. Note that this causes a bug because the log message below uses `err` :)

jpeach (changes requested):
This is pretty close now. @ShaileshSurya can you please rebase and squash your commits?

cmd/contour/certgen.go
ShaileshSurya: Replaced the error with oerr.

cmd/contour/certgen.go
```diff
@@ -137,34 +138,47 @@ func OutputCerts(config *certgenConfig, kubeclient *kubernetes.Clientset, certs
...
-			check(fmt.Errorf("unsupported Secrets format %q", config.Format))
+			return fmt.Errorf("unsupported Secrets format %q", config.Format)
 		}
 	}
 
 	if config.OutputPEM {
 		fmt.Printf("Writing certificates to PEM files in %s/\n", config.OutputDir)
-		check(certgen.WriteCertsPEM(config.OutputDir, certs, force))
+		if err := certgen.WriteCertsPEM(config.OutputDir, certs, force); err != nil {
+			return fmt.Errorf("Writing certificates to PEM files in %s/\n", config.OutputDir)
```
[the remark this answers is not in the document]
ShaileshSurya: Done.

cmd/contour/certgen.go
```diff
@@ -137,34 +138,47 @@ func OutputCerts(config *certgenConfig, kubeclient *kubernetes.Clientset, certs
...
+		if err := certgen.WriteSecretsKube(kubeclient, secrets, force); err != nil {
+			return fmt.Errorf("failed to write  %q format Secrets to namespace %q\n", config.Format, config.Namespace)
+		}
 	}
+	return nil
 }
 
-func doCertgen(config *certgenConfig) {
+func doCertgen(config *certgenConfig, log logrus.FieldLogger) {
+
```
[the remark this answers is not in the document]
ShaileshSurya: Done

cmd/contour/certgen.go
ShaileshSurya: Done.

cmd/contour/certgen.go
ShaileshSurya: Done.

cmd/contour/certgen.go
ShaileshSurya: Done.

cmd/contour/certgen.go
ShaileshSurya: Done.

cmd/contour/certgen.go
ShaileshSurya: Done.

cmd/contour/servecontext_test.go
ShaileshSurya: Done.

ShaileshSurya:
> This is pretty close now. @ShaileshSurya can you please rebase and squash your commits?

Done.

jpeach (approved):
LGTM

jpeach:
Merged bb14b25. Thanks a lot @ShaileshSurya 🎉

PR #2828, opened 2020-08-24
CI: run apt-get update at start of siteproof job

skriss:

This should fix the persistent error we've been getting with the siteproof CI job.


skriss:
replaced by #2829

PR #2829, opened 2020-08-24
CI: run site proof as its own job only if site/ has changed

### LINKED ISSUE #2031: site proof job did not run when site/ changed
https://github.com/projectcontour/contour/pull/2025 which changed site/ did not trigger the site proof job.

https://travis-ci.org/projectcontour/contour/jobs/623359710?utm_medium=notification&utm_source=github_status

### LINKED ISSUE #2772: CI: Site proof job is running on every PR
After the switch to GitHub actions, the site proof job is running on every PR. It really only needs to run when something in the site changes. There's a script to do this check, but it's not working as intended right now.

skriss:
Moves the site proof CI job into its own GitHub workflow, which
enables it to be run *only* if the contents of site/ have changed.

Fixes #2031
Fixes #2772 


skriss:
We'll need to remove the old `siteproof` check as required from the branch protection rules. I think we probably want to leave the new job as not required as well, since it won't always run (and it can be flaky).

youngnick (approved):
LGTM

jpeach (approved):
Nice.

Issue #2683, opened 2020-07-14
Adopt the controller-runtime cache.


jpeach:
**Please describe the problem you have**

controller-runtime has a [caching informer](https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg/cache#Cache) implementation that we could adopt without without necessarily adopting the rest of the framework. Since we are now doing more work with informers, and the cache has more types in it, this could be useful.

jpeach:
I'm thinking that I should tackle this before informing on `ExtensionService` resources for auth. In my PoC branch, it is a bit  repetitive to add the new type into the cache.

However, this will be a fair chunk of refactoring
- controller-runtime doesn't use namespaces informer factories, so root namespaces needs a refactor
- the clients of the KubernetesCache can't reach its internals any more, so we need to refactor to add an API
- figuring out rebuild triggers will be an issue, maybe FieldIndexers can help with that

/cc @stevesloka

PR #2831, opened 2020-08-25
internal/k8s: adopt the RESTMapper interface

jpeach:
The RESTMapper interface wraps up most of the policy around dealing
with detecting API resources and versions. It also has convenience API
for converting between Kind and Resource, which is useful since the
controller-runtime cache deals in kinds, and Contour deals in resources.

This updates #2683.


jpeach:
Draft at least until I figure out the right version of controller-runtime.

jpeach:
> Draft at least until I figure out the right version of controller-runtime.

Updated to the latest controller-runtime release, but let's still hold off merging this until after Contour 1.8 releases.

youngnick (approved):
LGTM

examples/contour/01-crds.yaml
```diff
@@ -41,67 +41,46 @@ spec:
       description: HTTPProxy is an Ingress CRD specification.
       properties:
         apiVersion:
-          description: 'APIVersion defines the versioned schema of this representation
-            of an object. Servers should convert recognized schemas to the latest
-            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+          description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
```
stevesloka, line 44: Why did all the crd specs change?

jpeach:
> This looks fine on the top, but what is it solving?

It's a baby step on the path towards controller-runtime. The controller-runtime cache API wants kinds but we always use resources. The RESTMapper interface has convenience functions to do the conversion, which will be useful when we plug it in.

> The code looks the same pretty much just a different client. Also a small nit on why all the CRD descriptions changed.

Yeh, it just seems to be string literal formatting. Probably because the Kubernetes dependencies got a minor version bump?

stevesloka:
> Probably because the Kubernetes dependencies got a minor version bump?

:shakes_fist: at k8s deps. =)

PR #2832, opened 2020-08-25
internal/dag: upstream server cert validation for HTTP/2

### LINKED ISSUE #2637: Upstream peer validation is only available for the "tls" protocol.
In `computeRoutes`, the upstream validation stanza is only checked if the upstream protocol is "tls". It should also be available for all protocols that imply TLS (i.e. `h2`).

tsaarni:
Fixes #2637.


youngnick (approved):
LGTM

PR #2837, opened 2020-08-27
internal: improve Ingress class matching log messages

jpeach:
Make the two debug log messages for Ingress class matching emit
the same field.


youngnick (approved):
LGTM, thanks.

PR #2840, opened 2020-08-27
cmd/contour: add xds-server-type arg to config file

stevesloka:
Adds a config file option to define the XDS server to use when serving Envoy's gRPC XDS.

### DIFF
--- a/examples/contour/01-contour-config.yaml
+++ b/examples/contour/01-contour-config.yaml
@@ -6,6 +6,8 @@ metadata:
   namespace: projectcontour
 data:
   contour.yaml: |
+    # determine which XDS Server implementation to utilize in Contour.
+    # xds-server-type: contour
     # should contour expect to be running inside a k8s cluster
     # incluster: true
     #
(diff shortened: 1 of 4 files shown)


examples/contour/01-contour-config.yaml
```diff
@@ -6,6 +6,9 @@ metadata:
   namespace: projectcontour
 data:
   contour.yaml: |
+    # determine which XDS Server implementation to utilize in Contour.
+    # Options are: "contour" or "gocontrolplane"
```
youngnick, line 10: Nit: Just a note here that "gocontrolplane" is TBD?

youngnick (approved):
LGTM, with a small nit. Up to you.

examples/contour/01-contour-config.yaml
stevesloka: I'll take out since it's not implemented yet, but the next one will probably implement. We'll need names, if that's `contour` & `gcp` or `gocontrolplane` or something like `builtin` & `upstream`. Or `stable` & `experimental`? I dunno, but that can be for the next one. =)

youngnick: `gcp` is out because `Google Compute Platform` or whatever it actually stands for. ;)

PR #2841, opened 2020-08-27
Revert "cmd/contour: add xds-server-type arg to config file"

stevesloka:
Reverts projectcontour/contour#2840


jpeach:
I don't think we need to revert this. @youngnick is just going to branch the release from an earlier commit.

PR #2842, opened 2020-08-27
docs/tls: Add a note for case-sensitive cert

### LINKED ISSUE #2793: Document SNI case sensitivity in Ingress and HTTPProxy FQDNs
Envoy maintains case sensitivity for FQDNs. This should be reflected in Contour's documentation for configuration.

see: [envoyproxy/envoy/6199](https://github.com/envoyproxy/envoy/issues/6199)

tong101:
Currently SNI matching during TLS handshake is case-sensitive in Envoy. Add a section in TLS guide doc to document this.

Fixes #2793

### DIFF
--- a/site/_guides/tls.md
+++ b/site/_guides/tls.md
@@ -17,6 +17,13 @@ Enabling TLS support requires Contour version 0.3 or later. You must also add an
 
 If you deploy behind an AWS Elastic Load Balancer, see [EC2 ELB PROXY protocol support][2] for special instructions.
 
+## TLS SNI name matching
+Envoy SNI name matching during TLS handshake is case-sensitive.
+For example, for a cert with common name foo.bar.com, requests to Foo.bar.com would not match.
+Similarly, for cert with wildcard name \*.bar.com, only requests to lower case name will match.
+Here is the [known issue][4] reported on Envoy.
+
 [1]: {{site.github.repository_url}}/blob/{{site.github.latest_release.tag_name}}/examples/contour/03-contour.yaml/#L45
 [2]: {% link _guides/proxy-proto.md %}
 [3]: {% link _guides/cert-manager.md %}
+[4]: https://github.com/envoyproxy/envoy/issues/6199


jpeach (changes requested):
It's a bit more subtle than this change implies. The problem is that TLS SNI name matching is case-sensitive, not the domain name itself. There's no wildcard dependency IIRC.

Maybe this should be a FAQ?

tong101:
@jpeach You are totally right. By domain name, I meant the SNI name during TLS handshake, not the domain name on HTTP. It maybe a little confusing. How about adding a section in [tls guide](https://github.com/projectcontour/contour/blob/main/site/_guides/tls.md)?

jpeach:
> @jpeach You are totally right. By domain name, I meant the SNI name during TLS handshake, not the domain name on HTTP. It maybe a little confusing. How about adding a section in [tls guide](https://github.com/projectcontour/contour/blob/main/site/_guides/tls.md)?

I think that's a good start. I don't see that page linked from the live site, so I expect there's some problem but I forget how this works so it needs a bit of digging.

tong101:
Here is the [link to the site](https://projectcontour.io/guides/tls/).

jpeach:
> Here is the [link to the site](https://projectcontour.io/guides/tls/).

Oh, it just doesn't appear in the "Guides" index on the Documentation header. Nice :-/

site/_guides/tls.md
```diff
@@ -17,6 +17,10 @@ Enabling TLS support requires Contour version 0.3 or later. You must also add an
 
 If you deploy behind an AWS Elastic Load Balancer, see [EC2 ELB PROXY protocol support][2] for special instructions.
 
+## TLS SNI name matching
+Envoy SNI name matching during TLS handshake is case-sensitive. For cert with common name foo.bar.com, requests to Foo.bar.com would like. Similarly, for cert with wildcard name \*.bar.com, only requests to lower case name are allowed. Here is the [known issue][4] reported on Envoy.
```
youngnick, line 21: It's better to split long lines like this up into one-sentence-per-line.

```suggestion
Envoy SNI name matching during TLS handshake is case-sensitive.
For example, for a cert with common name foo.bar.com, requests to Foo.bar.com would not match.
Similarly, for a cert with wildcard name \*.bar.com, only requests to the lower case name will match.
Here is the [known issue][4] reported on Envoy.
```

youngnick (changes requested):
Just a small wording change from me.

site/_guides/tls.md
tong101: Good point.

youngnick (approved):
LGTM

jpeach (approved):
Thanks @tong101

PR #2845, opened 2020-08-28
internal: Implement Envoy go-control-plane xDS Server

stevesloka:
Adds the Envoy go-control-plane xDS Server as an option instead
of the Contour built-in version configurable by the config file.

Note: This doesn't actually serve up any resources, just implements the gRPC xDS server. Future PR will add that support.

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -24,6 +24,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/envoyproxy/go-control-plane/pkg/cache/v2"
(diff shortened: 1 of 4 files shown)


jpeach:
This updates #2134 right? Is this intended to land before #2850?

jpeach:
> This updates #2134 right? Is this intended to land before #2850?

The split in functionality between this and #2850 isn't that clear to me. This one has more snapshot cache code, but the PR description implies the opposite?

stevesloka:
I was trying to make smaller commits to make it easier to review, but this never got in before I kept moving along. I'll just close this and use the other since it supersedes this one.

PR #2847, opened 2020-08-28
internal/dag: extract DAG processors from the main builder

skriss:
Extracts the logic specific to processing HTTPProxy, Ingress
APIs into their own processors that are invoked by the DAG
builder. Also extracts a listener processor to add HTTP/HTTPS
listeners after the other processors have run.

updates #2226 

Heavily inspired by all the work @stevesloka has done to date, and previous conversations between the maintainers.

I think there's still some more refactoring that can be done here, so this isn't necessarily final state, but it does hit the goal of being able to separate out processors for different APIs into their own types and turn them on/off at runtime, and I don't want this PR to get too large.

A few areas I have in mind for followup:
- clean up the interfaces between the `Builder`, the `Processors`, the `KubernetesCache`. Specifically, I'd like the `Processors` not to depend on a full `Builder`, but a subset of it - something like a `BuildContext` (or we could move to using the `DAG` here). Also, it'd be nice for the `Processors` to have a read-only view on the `KubernetesCache`.
- look at possibly moving the `Processors` into their own package(s)
- decide if we want to have `Processors` operate directly on a `DAG` rather than using a `Builder`/`BuildContext`


internal/dag/builder.go
```diff
@@ -38,53 +32,63 @@ type RouteServiceName struct {
...
 
+// Processor uses a DAG builder to construct part
+// of a DAG.
+type Processor interface {
+	// SetBuilder provides the Processor with a Builder
+	// to use.
+	SetBuilder(builder *Builder)
+
+	// Run executes the processor.
+	Run()
```
jpeach, line 43: Can we pass the Builder into `Run` and remove `SetBuilder`?

internal/dag/builder.go
```diff
@@ -38,53 +32,63 @@ type RouteServiceName struct {
...
-
-	FallbackCertificate *types.NamespacedName
+	listeners          []*Listener
 
 	StatusWriter
 	logrus.FieldLogger
 }
 
-// Build builds a new DAG.
+// Build builds and returns a new DAG.
```
jpeach, line 65: Add a note about processors? Especially we should note the API guarantee that processors are run in order (maybe add a test for that?).

internal/dag/builder.go
```diff
@@ -38,53 +32,63 @@ type RouteServiceName struct {
...
+	for _, p := range b.Processors {
+		p.SetBuilder(b)
+		p.Run()
+	}
 
-	b.computeIngresses()
+	var dag DAG
 
-	b.computeHTTPProxies()
+	// By definition, only listeners can be roots.
```
jpeach, line 76: This is not true for ext_auth clusters, which will be roots, so we need to refactor towards a model that will let the processors take care of which objects are roots (e.g. rewrite the DAG to lift hosts into listeners).

jpeach (changes requested):
I think that this is a useful step towards breaking down the monolith. The main improvement I'd like is to remove the `SetBuilder` method.

> Also, it'd be nice for the Processors to have a read-only view on the KubernetesCache.

Agreed; This is closely related to #2683, where I'd like to be able to shuffle the controller-runtime cache as the `Source`.

> look at possibly moving the Processors into their own package(s)

I'd support this so that we can enforce clean interface usage.

> decide if we want to have Processors operate directly on a DAG rather than using a Builder/BuildContext

I'm pretty sure it's possible to just have processors make passes over the DAG and not have a separate build context. That would be my preferred solution.

internal/dag/builder.go
skriss: Yep, I already ended up doing this in a subsequent WIP commit anyway. I'll make that change for this PR.

internal/dag/builder.go
skriss: I removed the comment. Can do the necessary refactoring in a separate PR.

skriss:
> I'm pretty sure it's possible to just have processors make passes over the DAG and not have a separate build context. That would be my preferred solution.

Would you move the `services`, `virtualhosts`, `securevirtualhosts` maps into the `DAG` struct along with the related methods that enable the `Processors` to easily check for/record existence of those objects? Or did you have a different approach in mind for how the `Processors` would interact with the `DAG`?

I think there are some nice things about keeping the Builder/BuildContext and the final DAG separate - you can have a different API for builders of the DAG vs. consumers of the DAG, and also you know that a DAG object is always a "final" representation rather in the process of being built.

internal/dag/httpproxy_processor.go
```diff
@@ -0,0 +1,763 @@
...
+	// TLS secret to use by default when SNI is not set on a
+	// request.
+	FallbackCertificate *types.NamespacedName
+}
+
+// Run translates HTTPProxies into DAG objects and
+// adds them to the DAG builder.
+func (p *HTTPProxyProcessor) Run(builder *Builder) {
+	p.builder = builder
+	p.orphaned = make(map[types.NamespacedName]bool, len(p.orphaned))
```
jpeach, line 51: To be safe, add:
```Go
defer func(){
    p.builder = nil
    p.orphaned = nil
}()
```

internal/dag/builder_test.go
```diff
@@ -6704,6 +6716,32 @@ func TestValidateHeaderAlteration(t *testing.T) {
...
+	assert.Equal(t, []string{"foo", "bar", "baz", "abc", "def"}, got)
+}
+
+type pluggableProcessor struct {
+	runFunc func(builder *Builder)
+}
+
+func (p *pluggableProcessor) Run(builder *Builder) {
+	p.runFunc(builder)
+}
```
jpeach, line 6743: I've generally found it's pretty useful to add a function adaptor at the API layer, eg.
```Go
type BuildProcessorFunc func(*Builder)

func (b BuildProcessorFunc) Build(builder *Builder) {
    if b != nil {
        b(builder)
    }
}
```

Maybe add this in a subsequent PR?

jpeach (approved):
Looks great; a few optional suggestions.

jpeach:
> > I'm pretty sure it's possible to just have processors make passes over the DAG and not have a separate build context. That would be my preferred solution.
> 
> Would you move the `services`, `virtualhosts`, `securevirtualhosts` maps that are currently in the `Builder` into the `DAG` struct along with the related methods that enable the `Processors` to easily check for/record existence of those objects? Or did you have a different approach in mind for how the `Processors` would interact with the `DAG`?

I think that some go into the DAG and some get subsumed by the resource cache. 

> I think there are some nice things about keeping the Builder/BuildContext and the final DAG separate - you can have a different API for builders of the DAG vs. consumers of the DAG, and also you know that a DAG object is always a "final" representation rather than in the process of being built.

I think that the DAG-only approach allows builders to be more independent. You can drop in a builder that manipulates the DAG without having to make corresponding changes to a build context. I suspect that in the medium term, this makes the API more robust.

youngnick (approved):
LGTM too.

Issue #2338, opened 2020-03-15
Securing connection between Envoy and upstream service: client authentication


tsaarni:
It is possible to use TLS for the connection between Envoy and upstream server, but currently only the server is authenticated. After external client validation #2250 is introduced, it would make sense to add client authentication also to the internal TLS connection (highlighted with red dashed box in the figure).

Use case: only authorized clients shall be able to access resources of the upstream server. Requests from unauthorized clients running within the same cluster shall be rejected.

Unauthorized clients are other applications that share the same cluster. This could potentially include vulnerable software that could be remotely exploited to access the server.

![client-auth](https://user-images.githubusercontent.com/85464/76843389-63d21a80-6844-11ea-92db-f5396669c43f.png)

**Edit 2020-03-17**: updated the picture to highlight that envoy should get own client certificate which is different from the certificate of the external client.

jpeach:
Envoy can't use the client certificate from the authorized client to connect to the upstream server because it doesn't have the private key. If you need to provide the client certificate to the upstream server, then probably TCP proxy (terminating TLS at the upstream) is the right approach.

There may be some merit to issuing client certificates to each Envoy proxy and programming it to use them to establish upstream connections, but that would require the upstream service to opt in to the scheme, and I suspect that few would do that.

tsaarni:
Using external client certificate is of course impossible, but in my case we do issue certificates internally within the cluster to secure cluster-internal communication. This follows the same principle as securing the xDS TLS connection between Contour and Envoy.

Similarly as the upstream services can be configured with server certificates (which Envoy verifies currently) in many cases they can be also configured with CA certificate to validate Envoy's certificate. 

Some existing applications have that opt-in feature in their configuration. When it comes to services that users are developing themselves, in my experience it is *very common*  for the REST/web frameworks to provide opt-in support to configure client certificate validation.

jpeach:
OK, confirming this is the second case, where each envoy proxy uses its own client certificate to connect to the backend.

codeyq:
+1 on this feature request. Securing out-cluster and in-cluster with different set of mTLS key certs.

jpeach:
I think this is reasonably do-able with the recent changes to certgen and bootstrap, though it is a bit trickier if you want to have a unique client certificate for each envoy (which would be ideal). I think that we could have the bootstrapper lay down a envoy client certificate that would automatically be configured and be presented on request.  The tricky part is that certgen doesn't know how many envoy instances need certificates.

[...]

PR #2848, opened 2020-08-28
design: backend TLS with client certificate

tsaarni:
This proposal describes how to implement support for client certificate for Envoy for the TLS connection between Envoy and backend services.

This updates #2338

### DIFF
--- a/design/tls-backend-client-authentication.md
+++ b/design/tls-backend-client-authentication.md
@@ -1,0 +1,62 @@
+# Backend TLS: client authentication
+
+Status: Draft
+
+This proposal describes how to implement support for client certificate for Envoy for the TLS connection between Envoy and backend services.
+
+## Goals
+
+- Envoy shall be able authenticate itself to the backend
+- Backend service shall be able to request clients to present their client certificate
+
+## Non Goals
+
+- Configure several client certificates for Envoy(s) e.g. different client certificate per each backend or different client certificate for each Envoy.
+- Specify mechanism how the CA certificate for validating Envoy's client certificate is distributed to the backend developers.
+
+## Background
+
+It is already possible to use TLS for the connection between Envoy and the backend service.
+However only the backend is authenticated and the backend is not able to authenticate Envoy.
+Since the backend has no means to differentiate between clients, any client within the cluster network is able to access backend's resources.
+This could potentially include vulnerable software which might be running within the cluster and remotely exploited to access the backend.
+
+After external client certificate validation was introduced to Contour 1.4.0, it makes sense to allow configuring client certificates for the TLS connection between Envoy and backends.
+This completes securing the end-to-end connection all the way from the external client to the backend.
+
+Securing the connection between Envoy and backend can have similar importance as securing the xDS TLS connection between Contour and Envoy.
+
+## High-Level Design
+
+Configuring the client certificate shall follow the same configuration pattern as `fallback-certificate`.
+The client certificate and key are stored in Secret of type `tls` in PEM format.
+The client certificate file `tls.crt` shall contain the client certificate as the first PEM block, followed by optional chain of CA certificates, up to but not including the root CA certificate.
+New configuration option is added to the Contour configuration file in the optional `tls.envoy-client-certificate` location.
+The value of the option refers to a Secret.
+The reference contains the namespace and the Secret name e.g. `namespace/name`.
+
+Certificate delegation is not needed since client certificate is configured within configuration file at deployment level, not in `HTTPProxy` resources on the application level.
+
+## Detailed Design
+
+New field is added to `dag.Cluster` that holds the client certificate for Envoy.
+Note that since only single client certificate is supported, the secret is the same for all instances of type `dag.Cluster`.
+When `CommonTlsContext` is constructed for the upstream connection, the client certificate is added using `SdsSecretConfig`, that is, the secret is streamed over SDS.
+Updates in `SecretCache` triggers secrets to be streamed to Envoy, which will reload the updated certificates and keys.
+
+When client certificate is defined in Contour configuration, it can be added to all Envoy clusters regardless if it will be used or not.
+Envoy will send the client certificate only when the backend requests the client to present its certificate during the TLS handshake.
+
+When [external authorization](external-authorization-design.md) is implemented in the future, the gRPC connection to the external authorization server shall also be protected by client authentication by the same principle laid out in this document.
+
+## Alternatives Considered
+
+Alternatively, client certificates could be configured in `HTTPProxy.spec.routes[].services[]` similarly as server validation is currently configured in `HTTPProxy.spec.routes[].services[].validation`.
+This allows different client certificate to be used towards each backend, and it allows teams in multitenant cluster to manage both client and server certificates by themselves.
+However, single client certificate is sufficient currently and allowing override per `HTTPProxy` can be considered as future extension.
+
+## Security Considerations
+
+Envoys can be considered as single "distributed" TLS client, all equally authorized to connect to the upstream services.
+Therefore single client certificate is enough for the use case.
+Separate certificates could be needed e.g. if upstream needs to tell each Envoy apart but this is not valid use case currently.


tsaarni:
Adding the picture from the referred issue here to explain the use case also visually:

76843389-63d21a80-6844-11ea-92db-f5396669c43f

Here in this picture the logical end-to-end connection is split into two TLS connections. The cluster external green leg is already protected by mutual TLS authentication but due to missing client certificate the red leg is not.  Unauthorized client in red is able to connect the upstream server as long as this is true.

I had a few questions in my mind and I'd appreciate to hear your opinions:

**Q1:** Configure client cert as a global command line parameters vs HTTPProxy? 

Since from Envoy's perspective the client certificate is a parameter in the upstream cluster definition, each upstream could be configured with a different client certificate.  From that point of view, the parameters could be part of `HTTPProxy.spec.routes[].services[]` instead of being a global parameter.

**Q2:** Would the multitenancy perspective drive towards global vs per service client certificate?

From security perspective I consider all Envoys being single "distributed" TLS client, all equally authorized to connect to the upstream services.  Therefore single client certificate is enough for the use case. Separate certificates could be needed e.g. if upstream needs to tell each Envoy apart but I don't see that as valid use case. 

But what about multitenancy perspective?  I don't have experience in using Contour in multitenant way myself so I'd appreciate to hear how you see this from workflow / management perspective. For the TLS server validation, each team is able to manage the CA certificate with `HTTPProxy.spec.routes[].services[].validation` structure.  For symmetry: should client side be managed by individual teams as well?

- If yes, team manages both client and server certs.
- If not, team receives the CA certificate for validating Envoy's client certificate from the cluster admin.

**Q3:** Cluster internal TLS and certificate delegation?

Delegation is for wildcard certificates in cluster external TLS, but would this feature add another use case for delegation: If a client certificate would be configured per HTTPProxy, would users still want to refer to a single globally configured secret?

youngnick:
Thanks for this PR @tsaarni!

My initial read on your questions is:

I think that in general, the most used pattern will probably be that the Contour Envoys constitute a 'service' and should have a single identity most of the time.

To that end, a global setting is the place to start, with the ability to set a client certificate and a single CA. This should hit the more obvious use-cases, even for multitenancy. If users still need greater configurability, we can come back around and see if we grant the ability to override the default on a per-HTTPProxy basis.

However, I'm going to check the service-apis for how we will need to do this there, and see if there's anything we can crib back.

design/tls-backend-client-authentication.md
```diff
@@ -0,0 +1,42 @@
...
+
+After external client certificate validation was introduced to Contour 1.4.0, it makes sense to allow configuring client certificates for the TLS connection between Envoy and backends.
+This completes securing the end-to-end connection all the way from the external client to the backend.
+
+Securing the connection between Envoy and backend can have similar importance as securing the xDS TLS connection between Contour and Envoy.
+
+## High-Level Design
+
+The client certificate and key is stored in Secret of type `tls`.
+New optional argument is added to `contour serve` named `--envoy-client-certificate` which refers to a Secret.
```
jpeach, line 31: Not sure that we need a flag. Shall we start with config-only and see how that goes?

design/tls-backend-client-authentication.md
```diff
@@ -0,0 +1,42 @@
...
+The reference contains the namespace and the Secret name e.g. `namespace/name`.
+The same configuration is added to the Contour configuration file in the optional `tls.envoy-client-certificate` location.
+
+## Detailed Design
+
+New field is added to `dag.Cluster` that holds the client certificate for Envoy.
+Note that since only single client certificate is supported, the secret is the same for all instances of type `dag.Cluster`.
+When `CommonTlsContext` is constructed for the upstream connection, the client certificate is added using `SdsSecretConfig`, that is, the secret is streamed over SDS.
+
+When client certificate is defined in Contour configuration, it can be added to all Envoy clusters regardless if it will be used or not.
```
jpeach, line 41: When ExtensionClusters land, we should have the same feature for this too. Just need to mention that here, and make sure we have an issue for it when you scope the work.

design/tls-backend-client-authentication.md
```diff
@@ -0,0 +1,42 @@
...
+After external client certificate validation was introduced to Contour 1.4.0, it makes sense to allow configuring client certificates for the TLS connection between Envoy and backends.
+This completes securing the end-to-end connection all the way from the external client to the backend.
+
+Securing the connection between Envoy and backend can have similar importance as securing the xDS TLS connection between Contour and Envoy.
+
+## High-Level Design
+
+The client certificate and key is stored in Secret of type `tls`.
+New optional argument is added to `contour serve` named `--envoy-client-certificate` which refers to a Secret.
+The reference contains the namespace and the Secret name e.g. `namespace/name`.
```
jpeach, line 32: I think it's OK to not need a certificate delegation here since the certificate is being used by the infrastructure not by an application. Just add a note if you agree with this reasoning.

jpeach (changes requested):
This looks great - cleanly scoped, solid use case and nicely written. I think that it would be good for posterity if you merge some of the FAQ items you discussed in the issue into the main doc.

design/tls-backend-client-authentication.md
```diff
@@ -0,0 +1,42 @@
...
+The same configuration is added to the Contour configuration file in the optional `tls.envoy-client-certificate` location.
```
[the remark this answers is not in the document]
tsaarni: Added a note on following the same pattern.

design/tls-backend-client-authentication.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Detailed Design
+
+New field is added to `dag.Cluster` that holds the client certificate for Envoy.
+Note that since only single client certificate is supported, the secret is the same for all instances of type `dag.Cluster`.
```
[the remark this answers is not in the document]
tsaarni: I added chapter "Alternatives Considered" with per `HTTPProxy` ideas.

design/tls-backend-client-authentication.md
```diff
@@ -0,0 +1,42 @@
...
+When `CommonTlsContext` is constructed for the upstream connection, the client certificate is added using `SdsSecretConfig`, that is, the secret is streamed over SDS.
```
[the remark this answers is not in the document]
tsaarni: Do you mean CA bundle that the backend administrators/developers need for validating Envoy's client cert?

I think it would be outside of scope of this document. Since Contour or Envoy would not use it by itself, it would be up to the users to decide.  It could be well-known secret (assuming that backend admins have read access to to it) or it could be any out-of-band mechanism, like email.

I added mention about this in "Non Goals" chapter.

design/tls-backend-client-authentication.md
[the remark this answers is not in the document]
tsaarni: Yes, I added information on "Detailed Design".  This is how I assume xDS streaming works in Contour, and I hope this is not too far from truth, since I did not try this myself yet!

design/tls-backend-client-authentication.md
tsaarni: I removed the flag and just left the config.

design/tls-backend-client-authentication.md
tsaarni: I was not able to find any details on ExtensionClusters, is there anything out yet?

design/tls-backend-client-authentication.md
tsaarni: I did not write anything about this one yet, since wouldn't certificate delegation become relevant only if user is allowed to configure client certs in `HTTPProxy`?  Delegation concept does not apply when there is single global parameter and no way for teams in multitenant cluster to impact that?

tsaarni:
> I think that it would be good for posterity if you merge some of the FAQ items you discussed in the issue into the main doc.

I added "Security Considerations" chapter with content from the discussions.

design/tls-backend-client-authentication.md
jpeach: https://github.com/projectcontour/contour/blob/main/design/external-authorization-design.md

design/tls-backend-client-authentication.md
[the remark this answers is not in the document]
jpeach: The reason we need the CA is so the TLS peer can chain the client certificate back to a trusted root. So I think that we should add the CA bundle if is it present in the secret, consistent with how we use the `ca.crt` field in other TLS secrets. I think that we can start without supporting CA bundles in a separate secret.

jpeach:
> **Q1:** Configure client cert as a global command line parameters vs HTTPProxy?

There's 3 possible options AFAICT
1. Certificate per Contour deployment
2. Certificate per Envoy proxy pod
3. Certificate per HTTPProxy service

Each of these options has slightly different properties. I'm inclined to view this as an infrastructure certificate, which means starting at (1) and maybe progressing to (2) in the future. Would be interested if there are use cases for (3).

> **Q2:** Would the multitenancy perspective drive towards global vs per service client certificate?

A Contour deployment is 1-1 with a fleet of Envoy proxies, so each Contour deployment would be for a tenant. I agree that a separate certificate for each deployment is OK.

We should check whether the Contour TLS certificate validation allows SPIFFE certificates (looks to me like it won't accept them).

> **Q3:** Cluster internal TLS and certificate delegation?

Delegation is where an application is granted access to secrets owned by a cluster operator. Since these secrets are never used by an application, I don't think delegation comes into play.

design/tls-backend-client-authentication.md
jpeach: Yep, just capture that reasoning in the doc.

tsaarni:
> There's 3 possible options AFAICT
> 
> 1. Certificate per Contour deployment
> 2. Certificate per Envoy proxy pod
> 3. Certificate per HTTPProxy service

When multitenancy is done by having isolated deployments (1) we of course have per deployment configuration (so in a sense it was wrong for me to say "global configuration").

I was also thinking (2) but it would require Contour to tell Envoys apart from each other.  Also, since Contour pushes all configuration to all Envoys, the instances would need to be equally authorized to connect to any backend.  Backend X should either allow connections from all Envoys within Contour deployment or none.  Last, the number of Envoys can change, which makes it challenging to manage the number of client certificates.  Therefore I was thinking it is not a valid use case - all Envoys are seen as single distributed TLS client with a single identity.

Multitenancy can be bit overloaded word.  When reading how it was used within Contour, I understood that it is used to refer to a cluster with single Contour deployment where several devops teams manage their own routes by `HTTPProxy` inclusions?  In this scenario, I imagine tenants would be within single organization, so this is not necessarily the most typical use of the word.  Following that thought, I was thinking that the teams might want to manage their certificates, including the client certificate at the Envoy side, and that this would be scenario (3): define per `HTTPProxy`. This would be symmetrical how they already now manage the CA certificates for validating server certificate for the backends they set up.

> We should check whether the Contour TLS certificate validation allows SPIFFE certificates (looks to me like it won't accept them).

Yeah, I agree. SPIFFE certs do not have common name and `validateCertificate()` will fail.

design/tls-backend-client-authentication.md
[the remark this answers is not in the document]
tsaarni: Ok got it!  I think the convention is to add chain certificates as additional PEM blocks to `tls.crt` which I now added to the design doc.  I believe `kubectl create secret tls` has not allowed add separate `ca.crt`bundle file for chain certs.

design/tls-backend-client-authentication.md
tsaarni: I've added note with reference to external auth doc in the "detailed design" chapter.

design/tls-backend-client-authentication.md
tsaarni: Done.

jpeach:
> > We should check whether the Contour TLS certificate validation allows SPIFFE certificates (looks to me like it won't accept them).
> 
> Yeah, I agree. SPIFFE certs do not have common name and `validateCertificate()` will fail.

xref https://github.com/projectcontour/contour/issues/2883

design/tls-backend-client-authentication.md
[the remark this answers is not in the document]
jpeach: cert-manager secrets put the CA bundle in `ca.crt` so it's a pattern that's used in the wild. Not sure Contour deliberately supports placing the bundle in the `tls.crt` yep but it probably works too.

jpeach (approved):
This looks great to me.

jpeach:
@youngnick LMK if you have any further feedback. If not, I'll land this on Monday.

design/tls-backend-client-authentication.md
```diff
@@ -0,0 +1,62 @@
...
+This could potentially include vulnerable software which might be running within the cluster and remotely exploited to access the backend.
+
+After external client certificate validation was introduced to Contour 1.4.0, it makes sense to allow configuring client certificates for the TLS connection between Envoy and backends.
+This completes securing the end-to-end connection all the way from the external client to the backend.
+
+Securing the connection between Envoy and backend can have similar importance as securing the xDS TLS connection between Contour and Envoy.
+
+## High-Level Design
+
+Configurating the client certificate shall follow the same configuration pattern as `fallback-certificate`.
```
youngnick, line 31: Nit: 
```suggestion
Configuring the client certificate shall follow the same configuration pattern as `fallback-certificate`.
```

youngnick (approved):
This LGTM, with one small nit. Thanks again @tsaarni, great work!

design/tls-backend-client-authentication.md
tsaarni: Fixed!

tsaarni:
xref #2910

PR #2849, opened 2020-08-28
site: Fix typo in configuration reference

zianke:
Fix a typo in `default-http-versions` by moving description to the same row.
[Link](https://projectcontour.io/docs/main/configuration/) to the current Contour Configuration Reference.

Signed-off-by: Zian Ke <zian.ke@zian.ke>


zianke:
@skriss Updated for v1.8.0 doc.

PR #2850, opened 2020-08-28
internal: Wire up new snapshots to DAG rebuilds

stevesloka:
Adds the snapshotHandler to react to DAG rebuilds so that new
snapshots can be created.

Follows #2845 
Updates #2134 

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -24,6 +24,8 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/envoyproxy/go-control-plane/pkg/cache/v2"
+	"github.com/envoyproxy/go-control-plane/pkg/server/v2"
 	projectcontourv1alpha1 "github.com/projectcontour/contour/apis/projectcontour/v1alpha1"
 	"github.com/projectcontour/contour/internal/annotation"
 	"github.com/projectcontour/contour/internal/contour"
@@ -38,6 +40,7 @@ import (
 	"github.com/projectcontour/contour/internal/xds"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
 	"gopkg.in/alecthomas/kingpin.v2"
 	"gopkg.in/yaml.v2"
 	corev1 "k8s.io/api/core/v1"
@@ -216,11 +219,19 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		endpointHandler,
 	}
 
+	// snapshotCache is used to store the state of what all xDS services should
+	// contain at any given point in time.
+	snapshotCache := cache.NewSnapshotCache(false, xds.DefaultHash,
+		log.WithField("context", "xDS"))
+
+	// snapshotHandler is used to produce new snapshots when the internal state changes for any xDS resource.
+	snapshotHandler := contour.NewSnapshotHandler(snapshotCache, resources, log.WithField("context", "snapshotHandler"))
+
 	// Build the core Kubernetes event handler.
 	eventHandler := &contour.EventHandler{
 		HoldoffDelay:    100 * time.Millisecond,
 		HoldoffMaxDelay: 500 * time.Millisecond,
-		Observer:        dag.ComposeObservers(contour.ObserversOf(resources)...),
+		Observer:        dag.ComposeObservers(append(contour.ObserversOf(resources), snapshotHandler)...),
 		Builder: dag.Builder{
 			FieldLogger: log.WithField("context", "builder"),
 			Source: dag.KubernetesCache{
@@ -423,42 +434,48 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		}
 		log.Printf("informer caches synced")
 
+		var grpcServer *grpc.Server
+
 		switch ctx.XDSServerType {
 		case "contour":
-			rpcServer := xds.RegisterServer(
+			grpcServer = xds.RegisterServer(
 				xds.NewContourServer(log, contour.ResourcesOf(resources)...),
 				registry,
 				ctx.grpcOptions()...)
-
-			addr := net.JoinHostPort(ctx.xdsAddr, strconv.Itoa(ctx.xdsPort))
-			l, err := net.Listen("tcp", addr)
-			if err != nil {
-				return err
-			}
-
-			log = log.WithField("address", addr)
-			if ctx.PermitInsecureGRPC {
-				log = log.WithField("insecure", true)
-			}
-
-			log.Info("started xDS server")
-			defer log.Info("stopped xDS server")
-
-			go func() {
-				<-stop
-
-				// We don't use GracefulStop here because envoy
-				// has long-lived hanging xDS requests. There's no
-				// mechanism to make those pending requests fail,
-				// so we forcibly terminate the TCP sessions.
-				rpcServer.Stop()
-			}()
-
-			return rpcServer.Serve(l)
+		case "envoy":
+			grpcServer = xds.RegisterServer(
+				server.NewServer(context.Background(), snapshotCache, nil),
+				registry,
+				ctx.grpcOptions()...)
 		default:
 			log.Fatalf("invalid xdsServerType %q configured", ctx.XDSServerType)
 		}
-		return nil
+
+		addr := net.JoinHostPort(ctx.xdsAddr, strconv.Itoa(ctx.xdsPort))
+		l, err := net.Listen("tcp", addr)
+		if err != nil {
+			return err
+		}
+
+		log = log.WithField("address", addr)
+		if ctx.PermitInsecureGRPC {
+			log = log.WithField("insecure", true)
+		}
+
+		log.Info("started xDS server")
+		defer log.Info("stopped xDS server")
+
+		go func() {
+			<-stop
+
+			// We don't use GracefulStop here because envoy
+			// has long-lived hanging xDS requests. There's no
+			// mechanism to make those pending requests fail,
+			// so we forcibly terminate the TCP sessions.
+			grpcServer.Stop()
+		}()
+
+		return grpcServer.Serve(l)
 	})
 
 	// Set up SIGTERM handler for graceful shutdown.
--- a/internal/contour/snapshot.go
+++ b/internal/contour/snapshot.go
@@ -1,0 +1,46 @@
+// Copyright Project Contour Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package contour
+
+import (
+	"github.com/envoyproxy/go-control-plane/pkg/cache/v2"
+	"github.com/projectcontour/contour/internal/dag"
+	"github.com/sirupsen/logrus"
+)
+
+// SnapshotHandler implements the xDS snapshot cache
+type SnapshotHandler struct {
+	// resources holds the cache of xDS contents.
+	resources []ResourceCache
+
+	// snapshotCache is a snapshot-based cache that maintains a single versioned
+	// snapshot of responses for xDS resources that Contour manages
+	snapshotCache cache.SnapshotCache
+
+	logrus.FieldLogger
+}
+
+// NewSnapshotHandler returns an instance of SnapShotHandler
+func NewSnapshotHandler(c cache.SnapshotCache, resources []ResourceCache, logger logrus.FieldLogger) *SnapshotHandler {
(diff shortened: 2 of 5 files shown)


zianke:
Thanks @skriss. I guess you're commenting on #2849. Sure I'll fix it.

internal/contour/snapshot.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/projectcontour/contour/internal/dag"
+	"github.com/sirupsen/logrus"
+)
+
+// SnapshotHandler implements the xDS snapshot cache
+type SnapshotHandler struct {
+
+	// Observer allows for an event to propagate when
+	// the DAG is rebuilt and a new snapshot is needed.
+	dag.Observer
```
jpeach, line 28: Why do we embed the interface rather than just implementing it?

internal/xds/server.go
```diff
@@ -14,12 +14,17 @@
 package xds
 
 import (
+	"context"
+
+	api "github.com/envoyproxy/go-control-plane/envoy/api/v2"
```
jpeach, line 19: How about keeping this import as `api` and removing the other duplicates?

cmd/contour/serve.go
```diff
@@ -418,8 +428,8 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 
 		switch ctx.XDSServerType {
 		case "contour":
-			rpcServer := xds.RegisterServer(
-				xds.NewContourServer(log, contour.ResourcesOf(resources)...),
+			grpcServer := contourxds.RegisterServer(
+				contourxds.NewContourServer(log, contour.ResourcesOf(resources)...),
```
jpeach, line 432: Suggest that we hoist a `*grpc.Server`  variable out of the switch and then use the same listen code for both kinds of server, rather than duplicating it.

cmd/contour/serve.go
```diff
@@ -24,6 +24,8 @@ import (
 	"syscall"
 	"time"
 
+	xds "github.com/envoyproxy/go-control-plane/pkg/cache/types"
```
jpeach, line 27: If you add an alias for `ResourceType` to the existing `xds` package, then you can avoid the import conflict here and in other places. I'd recommend aliasing it as `ResourceType`, which indicates its usage better.

internal/contour/resources.go
```diff
@@ -17,29 +17,30 @@
...
 type ResourceCache interface {
 	dag.Observer
-	xds.Resource
+	contourxds.Resource
 }
 
 // ResourcesOf transliterates a slice of ResourceCache into a slice of xds.Resource.
-func ResourcesOf(in []ResourceCache) []xds.Resource {
-	out := make([]xds.Resource, len(in))
+func ResourcesOf(in map[xds.ResponseType]ResourceCache) []contourxds.Resource {
```
jpeach, line 34: I don't think you'll need to do this if you can make the mapping internal to the snapshot observer.

internal/contour/snapshot.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	xds "github.com/envoyproxy/go-control-plane/pkg/cache/types"
+	"github.com/envoyproxy/go-control-plane/pkg/cache/v2"
+	"github.com/projectcontour/contour/internal/dag"
+	"github.com/sirupsen/logrus"
+)
+
+// SnapshotHandler implements the xDS snapshot cache
+type SnapshotHandler struct {
```
jpeach, line 24: I'm not sure that "handler" is the right naming here. The "Handler" name is used for types that deal with Kubernetes events directly, but this is a DAG observer which is a different role.

jpeach (changes requested):

jpeach:
Can you please link the PR and commit to #2134? thanks :)

cmd/contour/serve.go
stevesloka: 👍

cmd/contour/serve.go
stevesloka: Something like this? 

```
type ResourceType = xds.ResponseType
```

How do I get to the enumeration types? 

This is the struct:

```
// ResponseType enumeration of supported response types
type ResponseType int

const (
	Endpoint ResponseType = iota
	Cluster
	Route
	Listener
	Secret
	Runtime
	UnknownType // token to count the total number of supported types
)
```

internal/contour/resources.go
stevesloka: I was wanting to re-use the ResourceCache, but need to know the types later on of which is the listener, which is endpoints, etc. 

Can you explain a bit more on your thought?

internal/contour/snapshot.go
stevesloka: It "handles' snapshots? I dunno, please suggest I think the name is fine. =)

internal/xds/server.go
stevesloka: Was just copy pasting from envoy example code, works to shorten the imports.

internal/contour/resources.go
jpeach: What I was thinking was using `ResourceCache.TypeURL()` to internally map to the control-plane `ResponseType`. The type URLs are global unique, so you can just map internally and keep the implementation details that need `ResponseType` private.

internal/contour/snapshot.go
jpeach: > It "handles' snapshots? I dunno, please suggest I think the name is fine. =)

ROFL, it's just that it's a handler in a different sense than the other types that we name "*Handler".  I will get by with "handler" if we can't think of a evocative alternative :)

internal/contour/resources.go
stevesloka: I can do that and remove the map, but I'd rather have a quick lookup to the resources than have to loop through them all and figure out what type is which. 

What is the downside of switching the type to a map?

If we ditch the map I'd end up with something like this in the snapshot constructor which to me feels bad, I'd rather just lookup the index of the map:

```
	for _, r := range resources {
		switch r.TypeURL() {
		case resource.ClusterType:
			clusterCache = r
		case resource.RouteType:
			routeCache = r
		case resource.ListenerType:
			listenerCache = r
		case resource.SecretType:
			secretCache = r
		case resource.EndpointType:
			endpointsCache = r
		}
	}
```

internal/xds/server.go
```diff
@@ -68,3 +69,39 @@ func RegisterServer(srv Server, registry *prometheus.Registry, opts ...grpc.Serv
 
 	return g
 }
+
+// RegisterEnvoyServer registers the given xDS protocol Server with the gRPC
+// runtime. If registry is non-nil gRPC server metrics will be automatically
+// configured and enabled.
+func RegisterEnvoyServer(registry *prometheus.Registry, sc cache.SnapshotCache, opts ...grpc.ServerOption) *grpc.Server {
```
skriss, line 76: It looks like you can probably de-dupe a lot of the implementation of this function with `RegisterServer` - seems like the only difference is that this one is instantiating the `xdsServer` directly rather than having it be passed in. You could either extract a helper that's used internally within this package, or you could instantiate the go-control-plane xds server in `serve.go` directly like the contour one, and then I don't think you even need `RegisterEnvoyServer`, you could just reuse `RegisterServer`.

internal/contour/snapshot.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// resources holds a map of ResourceCaches which
+	// relate back to the xDS caches used to populate
+	// new snapshots.
+	resources map[xds.ResponseType]ResourceCache
```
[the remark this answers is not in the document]
stevesloka: yes, the next step is to actually push through snapshots, but the observer on the dag is the key to initiate a new snapshot from the ResourceCaches.

internal/xds/server.go
stevesloka: sweet, I was able to do that and remove the extra RegisterEnvoyServer method.

internal/contour/resources.go
jpeach: Both the integer types and the typeURLs are well-known constants with a permanently fixed relationship, so you can preinitialize a constant map:
```Go
m := map[string]xds.ResponseType {
resource.EndpointType: types.Endpoint,
...
}

snapshot[m[r.TypeURL()]] = resourcesOf(r.Contents())
```

The benefit of this approach is that knowledge of the go-control-plane snapshot types can be encapsulated by the one component that really needs to know. The rest of Contour can be oblivious.

stevesloka: It's still a map though, so why are strings better than types? I'd prefer the types since they are easier to work with unless I'm not following.

jpeach: What I'm trying to avoid is the need for the go-control-plane types to propagate over the whole codebase. Only the snapshot handler needs them and I'm suggesting that we can localize their use internally to that.

stevesloka: The only way to hide from the other bits is to remove the map as described in https://github.com/projectcontour/contour/pull/2850#discussion_r480177395. 

I still prefer the type as implemented keeping the map.

stevesloka: To keep this moving I did what you said @jpeach, I don't want this to block hold up the implementation.

internal/contour/snapshot.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+// SnapshotHandler implements the xDS snapshot cache
+type SnapshotHandler struct {
+	clusterCache   contourxds.Resource
+	listenerCache  contourxds.Resource
+	secretCache    contourxds.Resource
+	endpointsCache contourxds.Resource
+	routeCache     contourxds.Resource
```
skriss, line 30: You could still store all these caches in a map keyed on ResourceType if you wanted, but I'm not exactly sure how they'll eventually be used. Easy to change later though if you do want to do that, so not a big deal.

internal/contour/resources.go
```diff
@@ -18,20 +18,20 @@ package contour
 
 import (
 	"github.com/projectcontour/contour/internal/dag"
-	"github.com/projectcontour/contour/internal/xds"
+	contourxds "github.com/projectcontour/contour/internal/xds"
```
jpeach, line 21: Nit: looks like we don't need to change this file at all?

internal/contour/server_test.go
```diff
@@ -25,7 +25,7 @@ import (
 	resource "github.com/envoyproxy/go-control-plane/pkg/resource/v2"
 	"github.com/projectcontour/contour/internal/dag"
 	"github.com/projectcontour/contour/internal/fixture"
-	"github.com/projectcontour/contour/internal/xds"
+	contourxds "github.com/projectcontour/contour/internal/xds"
```
jpeach, line 28: Nit: looks like we don't need to change this file at all?

internal/featuretests/featuretests.go
```diff
@@ -36,7 +36,7 @@ import (
 	"github.com/projectcontour/contour/internal/protobuf"
 	"github.com/projectcontour/contour/internal/sorter"
 	"github.com/projectcontour/contour/internal/workgroup"
-	"github.com/projectcontour/contour/internal/xds"
+	contourxds "github.com/projectcontour/contour/internal/xds"
```
jpeach, line 39: Nit: looks like we don't need to change this file at all

internal/xds/util.go
```diff
@@ -16,9 +16,12 @@ package xds
 import (
 	"strings"
 
+	xds "github.com/envoyproxy/go-control-plane/pkg/cache/types"
 	"k8s.io/apimachinery/pkg/types"
 )
 
+type ResourceType = xds.ResponseType
```
jpeach, line 23: Nit: now that the resource types are encapsulated by the snapshot handler, I don't think you need this?

internal/contour/snapshot.go
```diff
@@ -0,0 +1,79 @@
...
+
+	return &SnapshotHandler{
+		snapshotCache:  c,
+		clusterCache:   clusterCache,
+		secretCache:    secretCache,
+		listenerCache:  listenerCache,
+		endpointsCache: endpointsCache,
+		routeCache:     routeCache,
+		FieldLogger:    logger,
+	}
```
jpeach, line 73: Nit:

What I had in mind was something along the lines of the pseudocode below. We can treat the resources generically, and control how we forward the DAG changes.


```Go
type SnapshotHandler struct {
    snapshot types.Snapshot
    resources []ResourceCache
}

func NewSnapshotHandler(c cache.SnapshotCache, resources []ResourceCache, logger logrus.FieldLogger) *SnapshotHandler {
}

func (s *SnapshotHandler) OnChange(root *dag.DAG) {
    // Forward the DAG change to each resource cache.
    for _, r := range s.resources {
	r.OnChange(root)
    }

    for _, r := range s.resources {
	contents := r.Contents()
	responseType := responseTypeOf(r.TypeURL())
	version := nextVersionForType()

	s.snapshot.Resources[responseType] = cache.NewResources(version, responseSlice(contents)
    }
}
```

jpeach (approved):
Just a few nits around cleaning up the import alias. I think that we can still simplify the way the snapshot handler works, but we could do that as we land the implementation too.

internal/contour/snapshot.go
stevesloka: I prefer that approach and not have to map each type in a for loop, but since we need that it's nice to have each cache addressable directly. If that doesn't pan out can refactor later, but I think this will work out nice.

PR #2852, opened 2020-08-30
site: Add clickable anchor links to the website

### LINKED ISSUE #2469: Add clickable anchor links to the website
Most of the headings on the website have anchor link targets, but it's awkward to actually get a link to them. Add [anchor.js](https://github.com/bryanbraun/anchorjs) or something equivalent so there is clickable page element that can be used to obtain a link to an anchored element.

zianke:
Use [anchor.js](https://github.com/bryanbraun/anchorjs) to put an anchored element by the webpage headers, which can be used to obtain a link. It currently uses the default style (displayed on hover) and adds anchors to `h2`, `h3`, `h4`, `h5`, `h6` elements, which looks good to me, though there are other [options](https://www.bryanbraun.com/anchorjs/#options).

Fixes #2469.

A preview with `make site-devel`:
anchor-preview

Signed-off-by: Zian Ke <zian.ke@zian.ke>

### DIFF
--- a/site/_includes/footer.html
+++ b/site/_includes/footer.html
@@ -48,6 +48,8 @@
 <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.12.9/umd/popper.min.js" integrity="sha384-ApNbgh9B+Y1QKtv3Rn7W3mgPxhU9K/ScQsAP7hUibX39j7fakFPskvXusvfa0b4Q"
     crossorigin="anonymous"></script>
 <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.1.3/js/bootstrap.min.js" crossorigin="anonymous"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/anchor-js/4.2.2/anchor.min.js" integrity="sha512-I7w3ZdSFzw5j3jU3ZkNikBNeIrl3i+hEuEdwNmqUJvwNcaBUNcijnP2gd9DtGlgVYDplfjGoD8vTNsID+lCjqg=="
+    crossorigin="anonymous"></script>
 
 <script src="/js/jquery.matchHeight.js?{{site.time | date: '%s%N'}}"></script>
 <script src="/js/scripts.js?{{site.time | date: '%s%N'}}"></script>
@@ -62,4 +64,9 @@ $(document).ready(function() {
         listType: 'ul',
     });
 });
+</script>
+
+<!-- https://github.com/bryanbraun/anchorjs -->
+<script>
+    anchors.add();
 </script>
--- a/site/_resources/faq.md
+++ b/site/_resources/faq.md
@@ -3,7 +3,7 @@ title: Frequently Asked Questions
 layout: page
 ---
 
-# Q: What's the relationship between Contour and Istio? 
+## Q: What's the relationship between Contour and Istio? 
 
 Both projects use Envoy under the covers as a "data plane".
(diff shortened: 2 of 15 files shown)


site/_includes/footer.html
```diff
@@ -48,6 +48,7 @@ <h5>{{ site.footer.title }}</h5>
 <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.12.9/umd/popper.min.js" integrity="sha384-ApNbgh9B+Y1QKtv3Rn7W3mgPxhU9K/ScQsAP7hUibX39j7fakFPskvXusvfa0b4Q"
     crossorigin="anonymous"></script>
 <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.1.3/js/bootstrap.min.js" crossorigin="anonymous"></script>
+<script src="https://cdn.jsdelivr.net/npm/anchor-js/anchor.min.js"></script>
```
jpeach, line 51: Let's use `cdnjs` here so that we don't have to hit a new domain for this. You can also add the `integrity` and `crossorigin` tags

jpeach (changes requested):
Nice!

One thing I would suggest is to remove the underline from the anchor links. This would be CSS something like:
```css
a.anchorjs-link {
  text-decoration: none;
}
```

This would go into the CSS in `site/_scss`

site/_includes/footer.html
zianke: Yeah makes sense. Updated.

zianke:
Thanks. I've added it to `site/_scss/site/objects/_anchor.scss`. Also did a few code formatting to make the .scss style look more consistent.

youngnick (changes requested):
Looks awesome, but when I tested with the test deploy (which, sadly, is only accessible to maintainers atm), I noticed that the FAQ page doesn't get the links? That's the one that I've needed links for the most!

https://deploy-preview-2852--projectcontour.netlify.app/resources/faq/

is the offending page, you can see that other pages in the 'resources' section work fine though:
https://deploy-preview-2852--projectcontour.netlify.app/resources/support/
https://deploy-preview-2852--projectcontour.netlify.app/resources/kubernetes/

@zianke, you should be able to use that deploy preview to check your PR's rendering too, it will get updated when your PR is pushed.

zianke:
@youngnick Thanks for letting me know about this issue. That's because the FAQ questions use `h1` elements, while anchor.js by default applies only to `h2`~`h6` (though this can be configured to apply to `h1` as well). I think the reason for this default behavior is that the main titles, e.g., "Contour Support Policy" on [support](https://deploy-preview-2852--projectcontour.netlify.app/resources/support/), usually use `h1`, to which we perhaps won't want to add an anchor. Would you prefer to change the FAQ questions to `h2` as a workaround?

youngnick:
I think changing the affected pages to H2 is much better, I'm okay with including that in here.

zianke:
Sounds good. I've reduced the header sizes by one level for those two affected pages. They should look good now:
https://deploy-preview-2852--projectcontour.netlify.app/resources/faq/
https://deploy-preview-2852--projectcontour.netlify.app/resources/upgrading/

Besides, I removed the notes by @jpeach regarding `toc: h1,h2` as we are no longer using `h1`, and the default config `toc: h2,h3` should just be fine. (Actually we're not enabling toc on the two pages though :)

jpeach (approved):

youngnick (approved):

Issue #2851, opened 2020-08-29
Couple nits on helper method


mattmoor:
This comment looks outdated:
https://github.com/projectcontour/contour/blob/7e3f5a380e1a5fe2239613159825b729141f206d/internal/k8s/helpers.go#L22-L26

I would also advise against the direct use of `cmp.Equal` (or `cmp.Diff`) in non-test code because it panics as a failure mode.  We do use it in knative, but only through a wrapper that guards against panics.  Regardless, I believe the appropriate K8s library for this case is `k8s.io/apimachinery/pkg/api/equality`, which provides a method `equality.Semantic.DeepEqual`.

PR #2853, opened 2020-08-30
Switch from cmp.Equal to equality.Semantic.DeepEqual

mattmoor:
Google's `cmp` library is great, but it's intended for test code, so it can panic as a failure mode.  Also, the K8s equality package seems purpose built for the use case here.

Fixes: https://github.com/projectcontour/contour/issues/2851


jpeach (approved):
LGTM, thanks @mattmoor

PR #2861, opened 2020-08-31
internal: Add helpers to return []xds.Resource proto's for snapshots xDS Server

stevesloka:
This preps the local Contour caches for returning arrays of type xds.Resource
so they can be called when creating new snapshots for the go-control-plane
xDS implementation.

Updates #2134

### DIFF
--- a/internal/contour/cluster.go
+++ b/internal/contour/cluster.go
@@ -17,9 +17,9 @@ import (
 	"sort"
 	"sync"
 
-	resource "github.com/envoyproxy/go-control-plane/pkg/resource/v2"
-
 	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
+	envoy_api_v2_xds "github.com/envoyproxy/go-control-plane/pkg/cache/types"
+	resource "github.com/envoyproxy/go-control-plane/pkg/resource/v2"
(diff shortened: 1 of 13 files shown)


jpeach (changes requested):
The go-control-plane Resource interface is just the `proto.Message` interface. So is there a reason why we can't add a helper to do the casting, and just call that in the consumer that needs the `Resource` type?

```
func resourcesOf(m []proto.Message) []xds.Resource {
...
}
```

stevesloka:
That's essentially what this is doing instead if casting twice. This converts the type to the desired proto depending on the consumer who needs a specific type. 

One we switch over to go control plane fully, the old methods can be deleted.

jpeach:
> That's essentially what this is doing instead if casting twice. This converts the type to the desired proto depending on the consumer who needs a specific type.

But it's the same object and it would convert automatically if Go casting applied to slices. Just seems like a lot of churn for a change that could be isolated to the single place that needs it.

> One we switch over to go control plane fully, the old methods can be deleted.

I'm not sure why we would need to do that since, Resources and Messages are the same thing, but it we do, then wouldn't we just change the return type of `Contents()`?

Issue #2141, opened 2020-01-23
Define how HTTPProxy validity interacts with delegation


jpeach:
HTTPProxy delegation can interact with validity in unexpected ways.

1. If proxy A delegates to proxy B, and proxy B is not present, then proxy A is marked invalid.
2. If proxy A delegates to B, but B is invalid for some reason, then B is marked invalid. A is considered valid, but lacks any routes that would have been imported by B.

If we consider a proxy with no routes (or TCP proxying) invalid as per #1988, then that can have the effect of bubbling errors up the delegation chain until the root proxy becomes invalid. Consider the case of A includes B, which includes C and D. If C and D are absent, then B will be invalid and A will subsequently have no routes. Similarly, if C and D both have errors, B will ve valid but have no routes, which means that A will also be valid and have no routes.

So we can see that defining a proxy with no routes as invalid has the undesirable property of propagating errors all the way up the delegation chain. This is because Contour's internal concept of valid/invalid doesn't align well with actual cases of partial or debatable validity.

Somewhat orthogonally to the concept of validity, we can consider what the corresponding Envoy configuration is for a proxy. Currently, it is reasonable to expect that if Contour marks a proxy as valid, then it will provision the corresponding Envoy resources. If a proxy has no routes, that's a problem that should be exposed through the API, However, if this state is tightly coupled to Envoy provisioning, as we described above, this could end up removing the Envoy vhost and even the corresponding listener. The implication of this is that propagating errors at the leaves of a delegation chain can cascade. what we really want is for Contour to limit the blast radius of configuration errors.

The status concepts described in #1868 can be helpful here, because applying conditions to the proxy status gives us a way to expose partial validity. If we can expose partial validity, then we can decouple that from the decision of whether to expose proxy resources into Envoy.

To resolve this issue, we should
- [ ] Document the principles for HTTPProxy delegation and error handling
- [ ] Document how HTTPProxy ought to expose Envoy resources in the presence of errors
- [ ] Evaluate moving to status conditions in order to make the documentation true

PR #2862, opened 2020-08-31
internal/dag: set HTTPProxy with no routes/includes/tcpproxy invalid

### LINKED ISSUE #2835: an HTTPProxy with no routes, includes, or tcpproxy is marked as valid
Create the following `HTTPProxy`:
```yaml
kind: HTTPProxy
metadata:
  name: empty
spec:
  virtualhost:
    fqdn: foo.bar.com
```

See that the proxy is marked as valid:
```bash
NAME    FQDN          TLS SECRET   STATUS   STATUS DESCRIPTION
empty   foo.bar.com                valid    valid HTTPProxy
```

The `HTTPProxy` should be marked invalid since it's "empty" and likely is a result of user error.

We should also add either a kubebuilder validation or a Gatekeeper validation for this to give the user immediate feedback.

**Environment:**

- Contour version: `v1.7.0`

skriss:
Fixes #2835 

This was what I had in mind for #2835. It does not attempt to solve the whole issue around inclusion described in #2141. It simply ensures that one of `spec.routes`, `spec.includes`, or `spec.tcpproxy` is defined/non-empty for each proxy.

I think this has value separate from addressing #2141, since it is a useful validation for folks who aren't using inclusion, and it shouldn't have any impact on inclusion chains, but let me know if I'm missing something.

Will likely rebase this on #2847.

### DIFF
--- a/examples/gatekeeper/validations/01-template-non-empty.yml
+++ b/examples/gatekeeper/validations/01-template-non-empty.yml
@@ -1,0 +1,32 @@
+# httpproxynonempty is a ConstraintTemplate that requires
+# an HTTPProxy to have at least one route, include, or a
+# tcp proxy defined.
+apiVersion: templates.gatekeeper.sh/v1beta1
+kind: ConstraintTemplate
+metadata:
+  name: httpproxynonempty
+  labels:
+    app: contour
+spec:
+  crd:
+    spec:
+      names:
+        kind: HTTPProxyNonEmpty
+        listKind: HTTPProxyNonEmptyList
+        plural: HTTPProxyNonEmptys
+        singular: HTTPProxyNonEmpty
+  targets:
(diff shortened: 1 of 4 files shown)


jpeach (commented):
This looks pretty reasonable to me. Note that the gatekeeper check isn't equivalent to the Contour check (which only tests root proxies).

youngnick (approved):
Yes, this LGTM.

skriss:
> Note that the gatekeeper check isn't equivalent to the Contour check (which only tests root proxies).

Good point, I overlooked that. Thinking about whether I should change the Gatekeeper config, or the Contour code...

skriss:
rebased & modified Gatekeeper validation to only apply to root proxies.

jpeach (approved):

youngnick (approved):
LGTM

Issue #2857, opened 2020-08-31
Performance: Contour suffers 2-3x slowdown programming Envoys under heavy churn


mattmoor:
**What steps did you take and what happened:**

tl;dr Run Knative's "scale-100" test.

I've been using Knative's Prow to run this for me.  
1. `ko publish ./cmd/contour`
2. Replace references to `docker.io/projectcontour/contour:*` with digest in [here](https://github.com/knative/serving/blob/master/third_party/contour-latest/contour.yaml)
3. Open a PR, this will automatically run `pull-knative-serving-contour-latest`

To access logs...
1. Click the "Details" on the `pull-knative-serving-contour-latest` check.
2. At the top of the next page click on "Artifacts".
3. Download `build-log.txt`

I generally scan for `scale-100` and snip everything above that (`scale-10` typically runs first), then you can `grep time-to-200 {build-log.txt} | sort`

**What did you expect to happen:**

See the timings below, but I'd love to see Contour somewhere closer in latency to where it is with status disabled 🤩 

**Anything else you would like to add:**

With 1.4 I see (just the tail latencies of 100 ksvcs):

```
    scale_test.go:38: "time-to-200" took 2m13.922577766s
    scale_test.go:38: "time-to-200" took 2m15.546537895s
    scale_test.go:38: "time-to-200" took 2m17.604262618s
    scale_test.go:38: "time-to-200" took 2m2.548284264s
    scale_test.go:38: "time-to-200" took 2m24.920357586s
    scale_test.go:38: "time-to-200" took 2m35.001542654s
    scale_test.go:38: "time-to-200" took 2m4.284511208s
    scale_test.go:38: "time-to-200" took 2m5.082870703s
    scale_test.go:38: "time-to-200" took 2m6.411782863s
```

With Contour 1.5+ these climb **over 5 minutes**.

With Contour 1.8 and status updates manually disabled (these lines commented):

```
    scale_test.go:38: "time-to-200" took 45.933159712s
    scale_test.go:38: "time-to-200" took 46.045489619s
    scale_test.go:38: "time-to-200" took 48.399353288s
    scale_test.go:38: "time-to-200" took 48.469338024s
    scale_test.go:38: "time-to-200" took 48.661313485s
    scale_test.go:38: "time-to-200" took 48.662251973s
    scale_test.go:38: "time-to-200" took 50.3964567s
    scale_test.go:38: "time-to-200" took 52.196902407s
[...]

PR #2864, opened 2020-09-01
Second attempt at fixing knative-sandbox/net-contour#226

mattmoor:
Replace the dynamic client Get (uncached) with a read from
the informer's lister cache.  This code path is evaluated
for every HTTP Proxy on each DAG rebuild, and the call to
the API Server on each results in a thundering herd that
triggers the built-in rate limiting in k8s.io/client-go.

With this change, the only API Server calls on DAG rebuilds
happen only when our cached copy of the HTTP Proxy shows
that we need to perform a status update.

Updates #2857

### DIFF
--- a/internal/k8s/statusupdater.go
+++ b/internal/k8s/statusupdater.go
@@ -50,12 +50,13 @@ func (m StatusMutatorFunc) Mutate(old interface{}) interface{} {
 
 // StatusUpdateHandler holds the details required to actually write an Update back to the referenced object.
 type StatusUpdateHandler struct {
-	Log           logrus.FieldLogger
-	Clients       *Clients
-	UpdateChannel chan StatusUpdate
-	LeaderElected chan struct{}
-	IsLeader      bool
-	Converter     *UnstructuredConverter
+	Log             logrus.FieldLogger
+	Clients         *Clients
+	UpdateChannel   chan StatusUpdate
+	LeaderElected   chan struct{}
+	IsLeader        bool
+	Converter       *UnstructuredConverter
+	InformerFactory InformerFactory
 }
 
 // Start runs the goroutine to perform status writes.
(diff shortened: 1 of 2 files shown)


mattmoor:
I think this gets us the vast majority of the speedup to be had.  I believe that this is actually faster than 1.4, where 1.5-1.8 are currently quite a bit slower under high churn.

mattmoor:
Pushed an update with a detailed commit message now that we know this is a keeper 😉

mattmoor:
I copied the new commit message into the PR description as well.

mattmoor:
Rebased

internal/k8s/statusupdater.go
```diff
@@ -82,9 +83,10 @@ func (suh *StatusUpdateHandler) Start(stop <-chan struct{}) error {
 			suh.Log.WithField("name", upd.NamespacedName.Name).
 				WithField("namespace", upd.NamespacedName.Namespace).
 				Debug("received a status update")
-			uObj, err := suh.Clients.DynamicClient().
-				Resource(upd.Resource).
-				Namespace(upd.NamespacedName.Namespace).Get(context.TODO(), upd.NamespacedName.Name, metav1.GetOptions{})
+
+			// Fetch the lister cache for the informer associated with this resource.
+			lister := suh.InformerFactory.ForResource(upd.Resource).Lister()
+			uObj, err := lister.ByNamespace(upd.NamespacedName.Namespace).Get(upd.NamespacedName.Name)
```
skriss, line 89: We're not explicitly waiting for cache sync here before making the Get call; I'm not sure yet if that's actually an issue or not, as it's possible that this code can't actually execute until other sections that *do* wait for cache sync have run. Looking into it some more.

skriss: I guess it's probably OK since this is all getting triggered off an event handler for the HTTPProxy informer anyway.

mattmoor: Based on slack threads, my understanding is that the only way things would land in the DAG is via informer events.

So while I think you are right (based on Contour's architecture) that the informer cache may not have fully synchronized, I'd suppose that the resource in question would be present in the cache (since informer events announcing its presence in the cache are what triggered its presence in the DAG).

If the cache is stale, then I'd expect an optimistic concurrency failure on the UpdateStatus.  If that proves to be a problem, then we can wrap this fetch/mutate/update in a retry loop that uses the client on subsequent retries.

skriss: Yep, that sounds right/reasonable to me!

mattmoor: It's notable that one of the ways more conventional controllers handle this is to decouple the pre-sync informer events from the post-sync reconciliation using the workqueue (cc @youngnick ).  Basically:

1. Set up informer events that dump into workqueue
2. Wait for sync
3. Continuously drain workqueue

The direct attachment of the event and the processing makes waiting for informer syncs before leverage the cache a trickier proposition.

I bring it up because @youngnick and I were chatting about workqueue usage last night, and this felt like a relevant use case in the broader ecosystem.

jpeach (approved):
LGTM.

@youngnick can you please weak the commit subject when you merge?

mattmoor:
@jpeach I think they should match.

youngnick (approved):
LGTM

youngnick:
Using squash and merge puts the PR number into the commit message anyway, so I changed it a little bit.

PR #2866, opened 2020-09-02
internal/k8s: Change StatusUpdaterHandler channel to buffered

youngnick:
Updates #2857.

This should ensure that the DAG has some headroom before
status updates start blocking the DAG rebuilds, even in
reasonably busy clusters.


jpeach (approved):
LGTM

Issue #2713, opened 2020-07-21
Auth: Create Envoy clusters from ExtensionServices


jpeach:
**Please describe the problem you have**

Refactor the services code so that it can be substantially shared between Routes and ExtensionServices. Add DDAG support for generating Envoy clusters that from ExtensionServices.

xref #432
xref #2643

jpeach:
The `ExtensionService` type has a slice of Contour Services (not the same as Kubernetes Services) so that an operator can weight requests between multiple Kubernetes Services, giving them more flexibility to split traffic between Kuberenetes Services that implement the same GRPC endpoints. This is the same flexibility that `HTTPProxy` provides on for routes.

Internally, the Contour EndpointsTranslator implements a one-to-one mapping between Kubernetes Service endpoints and Envoy ClusterLoadAssignments (i.e. endpoint discovery service). The EndpointsTranslator operates independently of Contour ingress reconciliation. Ingresses (including HTTPProxies) consume the output of the EndpointsTranslator because internally, Contour generates a stable and predictable ClusterLoadAssignment from a Kubernetes Service name.

Note well that this means that the ClusterLoadAssignment generation is completely independent of the rest of the ingress reconciliation. Everything works in the end because the ClusterLoadAssignment names are consistent and globally unique, but there is no explicit coordination here.

A HTTPProxy Service generates a one or more DAG Services. A DAG Service generates both and Envoy Cluster (for the cluster discovery service) and a RouteAction (for the route configuration discovery service). Since the DAG Service represents Envoy configuration at multiple configuration layers, some fields are valid at only one of the layers.

So, to summarize:
* A Contour Route balances across multiple Contour Services (at the HTTP layer)
* A DAG Service maps to one Kubernetes Service
* A DAG Service generates one Envoy Cluster
* Each Envoy Cluster references one EDS ClusterLoadAssignment
* Each ClusterLoadAssignment contains endpoints from exactly one Kubernetes Service

The Envoy `ext_authz` filter configuration accepts a Envoy Cluster name as the target for its requests. We have to program the Envoy Cluster for `ext_authz` filter from the `ExtensionService` API (since that's why it was designed). We don't have a HTTP Connection Manager, so there is no involvement from the route configuration discovery service. This means that to balance requests over multiple Kubernetes Services, we have to do so within an Envoy Cluster. This is possible, but it doesn't fit naturally within the Contour Service API, so we need to reconfigure the ExtensionService API.

```Go
[...]

PR #2870, opened 2020-09-02
internal/dag: fix adding a ServiceCluster entry

jpeach:
`(*ServiceCluster).AddWeightedService` didn't copy the service name
properly, leading to a failure to match the service endpoints. Copy
the correct fields and add tests for `ServiceCluster` type.

This updates #2713.


internal/dag/dag.go
```diff
@@ -541,7 +541,7 @@ func (s *ServiceCluster) AddService(name types.NamespacedName, port v1.ServicePo
 func (s *ServiceCluster) AddWeightedService(weight uint32, name types.NamespacedName, port v1.ServicePort) {
 	w := WeightedService{
 		Weight:           weight,
-		ServiceName:      name.Namespace,
```
youngnick, line 544: :oof:

youngnick (approved):
LGTM, ouch.

PR #2871, opened 2020-09-02
internal: make ServiceCluster weights consistent

jpeach:
Make ServiceCluster weights consistent with HTTPProxy Route weights. If
no weights are specified, distribute the load equally, otherwise take
the weights as specified by the users. In the latter case, 0 is a valid
weight and removes all traffic from that destination.

Hoist the 2 helpers for marshaling protobuf uint32 wrappers into the
`protobuf` package since they are part of an API family we now use one
of them in the endpoints handler.

This updates #2713.


youngnick (approved):
LGTM

PR #2881, opened 2020-09-03
[WIP] move builder caches into DAG

skriss:

Updates #2226 

This is a WIP for moving the DAG builder caches into the `DAG` struct directly. So far, I've only moved two of them, just to illustrate what I was thinking.

If this looks reasonable, then I can move `services` and `listeners` over too (`services` is a little tricky since there are related `Builder` methods that access both the `KubernetesCache` and the `services` map, but I'll sort it out).

This could get us to something like the following, where the `Builder` is just orchestrating calls to the `Processors` and the `Processors` no longer depend on the `Builder`:

```go
// Processor constructs part of a DAG.
type Processor interface {
	// Run executes the processor with the given DAG.
	Run(dag *DAG)
}

// Build builds and returns a new DAG by running the
// configured DAG processors, in order.
func (b *Builder) Build() *DAG {
	dag := &DAG{
		virtualhosts:       make(map[string]*VirtualHost),
		securevirtualhosts: make(map[string]*SecureVirtualHost),
                ...
	}

	for _, p := range b.Processors {
		p.Run(dag)
	}
        
        return dag
}
```

The `Processors` would probably be instantiated with a `KubernetesCache`, a `StatusWriter` (if needed), and a `logrus.FieldLogger` so those wouldn't have to come from the `Builder`.

Let me know what you think!

### DIFF
--- a/internal/dag/builder.go
+++ b/internal/dag/builder.go
@@ -36,7 +36,15 @@ type RouteServiceName struct {
 // of a DAG.
 type Processor interface {
 	// Run executes the processor with the given Builder.
-	Run(builder *Builder)
+	Run(dag *DAG, builder *Builder)
+}
+
+type ProcessorFunc func(dag *DAG, builder *Builder)
+
+func (pf ProcessorFunc) Run(dag *DAG, builder *Builder) {
+	if pf != nil {
+		pf(dag, builder)
+	}
 }
 
 // Builder builds a DAG.
@@ -49,10 +57,8 @@ type Builder struct {
 	// use to build the DAG.
 	Processors []Processor
 
-	services           map[RouteServiceName]*Service
-	virtualhosts       map[string]*VirtualHost
-	securevirtualhosts map[string]*SecureVirtualHost
-	listeners          []*Listener
+	services  map[RouteServiceName]*Service
+	listeners []*Listener
 
 	StatusWriter
 	logrus.FieldLogger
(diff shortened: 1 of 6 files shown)


jpeach:
I guess that the way I had imagined this was a set of free functions that visited the DAG to do their work, though visiting the DAG requires dynamic type comparisons, so it's probably a lot slower. One role of thumb is being able to add new types to the DAG without the DAG really knowing about it. For ExtensionService, for example, you need a way to add a new root; for Listener, you need to find the virtual host types, add new roots, and remove existing roots.

That said, the additional structure in this PR is an improvement.

stevesloka:
This is what I had in mind @skriss with eventually pulling out the `processors` bit into a new package forcing a good API in the dag package to work against.

skriss:
> I guess that the way I had imagined this was a set of free functions that visited the DAG to do their work, though visiting the DAG requires dynamic type comparisons, so it's probably a lot slower. One role of thumb is being able to add new types to the DAG without the DAG really knowing about it. For ExtensionService, for example, you need a way to add a new root; for Listener, you need to find the virtual host types, add new roots, and remove existing roots.

@jpeach I get what you mean, though it does mean a whole lot more DAG-walking (i.e. every time you want to do a lookup or list), and I do worry a bit about efficiency at scale. Also, even if the `DAG` struct itself doesn't need to be changed to add a new type, you'd potentially still need to write new DAG-walking code for the new type. 

On the flip side, it does make the API simpler for processors since it doesn't require them to correctly add all relevant objects to the build caches - they just need to add roots, and then the visitors will be able to find whatever they need to from the DAG.

So I'm open to exploring that approach, but given our discussion last week about timeboxing this, I'd like explicit agreement from the other maintainers that it's worth it before going down that road (cc @youngnick @stevesloka).

Regardless, I do think we should expose the `roots` collection to processors so they can manipulate them, rather than having that hardcoded in the builder.

youngnick:
Consider this my explicit agreement the approach you just discussed sounds good to me @skriss 😀. I think the DAG-walking is probably okay, as there are other scaling bottlenecks to worry about first around the apiserver and actually programming Envoy.

I also think exposing the `roots` collection is a good idea.

skriss:
closing this out in favor of #2900

PR #2882, opened 2020-09-03
makefile: Update checkall task to vendor go modules before running

stevesloka:
The `lint` task fails when a previous vendor directory is out of date resulting in having to run `go mod vendor` manually, then run the `checkall` task again. This has the vendor step happen automatically to solve that problem. 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ export GO111MODULE=on
 check: install check-test check-test-race ## Install and run tests
 
 .PHONY: checkall
-checkall: check lint check-generate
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -82,7 +82,11 @@ race:
 	go install -mod=readonly -v -race $(GO_TAGS) $(MODULE)/cmd/contour
 
 download: ## Download Go modules
-	go mod download
+	go mod downloaddownload:
```
jpeach, line 85: typo?

skriss (approved):

PR #2884, opened 2020-09-04
internal/contour: Create new snapshot when dag is rebuilt

stevesloka:
When the dag is rebuilt, the OnChange() event is fired. When using
the go-control-plane xDS server, this then creates a new snapshot
for Envoy to consume and updates its configuration.

I'm going to update the featuretests next with this code once this is merged. 

Updates #2134


jpeach (changes requested):
This looks nice and clean!

internal/contour/snapshot.go
```diff
@@ -40,7 +68,70 @@ func NewSnapshotHandler(c cache.SnapshotCache, resources []ResourceCache, logger
...
+
+	// Generate new snapshot version.
+	snapshotVersion := s.newSnapshotVersion()
+
+	// Create an snapshot with all xDS resources.
+	snapshot := cache.Snapshot{}
+
+	// Loop through all the xDS resources and pass them to the snapshot
+	for _, r := range s.resources {
+		snapshot.Resources[responseTypeOf(r.TypeURL())] = cache.NewResources(snapshotVersion, asResources(r.Contents()))
```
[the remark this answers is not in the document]
stevesloka: How do you know it changed? At this point all we know is that the dag was rebuilt, so need to assume all the resources are different. We'd need to add a version to each resource type and then increment that if the content cache was updated. 

I think long term we should look at testing out ADS and reducing the gRCP load on Contour. Using that means we need a single version for the snapshot.

internal/contour/snapshot.go
```diff
@@ -28,10 +37,13 @@ type SnapshotHandler struct {
 	// snapshot of responses for xDS resources that Contour manages
 	snapshotCache cache.SnapshotCache
 
+	// SnapshotVersion holds the current version of the snapshot
```
skriss, line 40: nit: `snapshotVersion` ...

internal/contour/snapshot.go
```diff
@@ -40,7 +52,70 @@ func NewSnapshotHandler(c cache.SnapshotCache, resources []ResourceCache, logger
...
+
+	// Generate new snapshot version.
+	snapshotVersion := s.newSnapshotVersion()
+
+	// Create an snapshot with all xDS resources.
+	snapshot := cache.Snapshot{}
+
+	// Loop through all the xDS resources and pass them to the snapshot
+	for _, r := range s.resources {
+		snapshot.Resources[responseTypeOf(r.TypeURL())] = cache.NewResources(snapshotVersion, asResources(r.Contents()))
```
skriss, line 66: Threw me for a second that `snapshot.Resources` is an array, not a map or even a slice, and that the response types being used to index are ints. Might be worth a comment somewhere.

internal/contour/snapshot.go
```diff
@@ -40,7 +52,70 @@ func NewSnapshotHandler(c cache.SnapshotCache, resources []ResourceCache, logger
...
+
+	for i := range protos {
+		protos[i] = v.Index(i).Interface().(envoy_xds.Resource)
+	}
+
+	return protos
+}
+
+// responseTypeOf converts a typeURL to an envoy_xds.ResponseType.
+func responseTypeOf(typeURL string) envoy_xds.ResponseType {
```
skriss, line 107: it'd probably be slightly more efficient to use a map from `typeURL` to `ResponseType` here, rather than having to execute the switch every time. May not make any material difference in practice, though.

skriss (changes requested):
couple minor comments, but overall looks good

internal/contour/snapshot.go
stevesloka: Yeah I'll do that, it was the initial thought to make Resources a map from the start, but I'll do that work here.

stevesloka:
So endpoints don't update WITH the dag, they update AFTER the dag, I misunderstood that. We should just update the endpointstranslator to happen inline with the dag rebuild since we'll push cluster updates before even looking at endpoint updates given we want to still tie the endpoints change with a dag rebuild. 

Doing this then give the snapshot handler a simple way to get at the endpoint bits when they change.

internal/contour/snapshot.go
```diff
@@ -40,7 +52,70 @@ func NewSnapshotHandler(c cache.SnapshotCache, resources []ResourceCache, logger
 	}
 }
 
-// OnChange is called when the DAG is rebuilt
-// and a new snapshot is needed.
+// OnChange is called when the DAG is rebuilt and a new snapshot is needed.
 func (s *SnapshotHandler) OnChange(root *dag.DAG) {
+
```
jpeach, line 57: You can also make the ordering robust by forwarding the DAG change to the resources here (also remove them from the builder in serve.go). This guarantees that all the resources are updated before the snapshot is made.

```Go
for _, r := range s.resources {
    r.OnChange(root)
}
```

internal/contour/snapshot.go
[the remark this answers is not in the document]
jpeach: > How do you know it changed? At this point all we know is that the dag was rebuilt, so need to assume all the resources are different. 

I was just thinking about something simple like comparing the contents to the last contents you have of that version.

You don't need a per-resource version, you can just update the resource to the latest global version (the version is opaque so it's OK to skip).

stevesloka:
I ended up adding a new Notifier such that when any endpoints change the snapshothandler also gets an update.

skriss (approved):
LGTM, though someone else should approve as well.

internal/contour/endpointstranslator.go
```diff
@@ -244,6 +236,9 @@ func NewEndpointsTranslator(log logrus.FieldLogger) EndpointsInterface {
 // A EndpointsTranslator translates Kubernetes Endpoints objects into Envoy
 // ClusterLoadAssignment resources.
 type EndpointsTranslator struct {
+	// Observer notifies when the endpoints cache has been updated.
+	Observer Observer
```
jpeach, line 240: Why do we need a new observer interface? `Cond` already provides a way to get notifications on a change. You could add an object that does the channel wait on that and calls a function.

internal/contour/snapshot.go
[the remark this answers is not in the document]
stevesloka: But we'd need to keep a cache of the old version's contents to compare against the new or like I said keep a per-resource version and just compare against that. So if CDS last was version 10, and it's still version 10, then no changes have happened. 

Are you asking to only update the bits that change? If we want to move to ADS then we need to send all the caches together so that work is not required (assuming I understand you ask).

internal/contour/endpointstranslator.go
stevesloka: Right but I'd need to add this to the workgroup and have this be long running. I think the Observer interface is much cleaner to implement & consume.

internal/contour/snapshot.go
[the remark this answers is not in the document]
jpeach: > But we'd need to keep a cache of the old version's contents to compare against the new or like I said keep a per-resource version and just compare against that. So if CDS last was version 10, and it's still version 10, then no 
changes have happened.

You have the old contents in the snapshot already.

> Are you asking to only update the bits that change?

Yep, exactly.  Then we don't ask Envoy to do any work unless there are actual config changes. WE don't have to do it in this PR; maybe we can leave a comment and plan to experiment in a future issue.

>  If we want to move to ADS then we need to send all the caches together so that work is not required (assuming I understand you ask).

I don't think I'm following this. ADS is just using the same GRPC endpoint  for all the Envoy resource types. You still get a DiscoveryRequest for each resource type and AFAIK can version the types independently.

stevesloka:
@youngnick thoughts on this? I think everything is addressed in comments short of some resource versioning with James. I think it's fine as-is and if needed, can follow up with additional PRs later on.

youngnick (approved):
LGTM, although I do worry about the number of xDS updates we'll generate if endpoints are updating a lot.

stevesloka:
>  I do worry about the number of xDS updates we'll generate if endpoints are updating a lot.

Is this any different than today or are you referencing that all the caches are updated? Endpoints updates are delayed a bit since new services.endpoints are tied to the dag rebuild. I'm happy to change things but want to understand since it keeps coming up.

youngnick:
Yeah, I meant the caches being updated. I agree that it's almost certainly causing problems as it is, but I'm just worried about ensuring that we keep it to only necessary changes; I don't want those Envoy memory issues caused by no-op updates to come back.

stevesloka:
Thanks for clarifying @youngnick. I'll get that fixed up on the next one.

PR #2887, opened 2020-09-07
internal/dag: Change status mechanisms to be ready for Conditions

youngnick:
This commit updates a number of pieces in `status.go` to be ready for adding conditions.
In general, these are to make the names clearer and to keep space for us to add further
support for other objects later as well.

`StatusUpdate` is renamed to `ProxyStatusUpdate` as it's only used for HTTPProxy.
`ObjectStatusWriter` is renamed to `ProxyStatusWriter`, for the same reason.

`StatusWriter`'s status cache has been renamed to `proxyStatuses` from `statuses`, in
case we want to use it for storing other object updates later.

Short values have been updated to reflect the new struct names (`osw` changed to `psw`, etc.)

`SetInvalid` has been updated to `AddInvalidCondition`, with a new `conditionType`
argument, that is currently only used to add some extra text into the `description`
field of the HTTPProxy. This will mean that further PRs should not need to touch the
HTTPProxy Processor.

This is currently a WIP however, the tests have not been updated yet. Since we
moved to `assert.Equal` for errors, the errors messages must now match for
tests to pass.

### DIFF
--- a/internal/dag/status.go
+++ b/internal/dag/status.go
@@ -21,8 +21,8 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
-// Status contains the status for an HTTPProxy (valid / invalid / orphan, etc)
-type Status struct {
+// ProxyStatusUpdate contains the status for an HTTPProxy (valid / invalid / orphan, etc)
+type ProxyStatusUpdate struct {
 	Object      k8s.Object
 	Status      string
 	Description string
@@ -30,24 +30,24 @@ type Status struct {
 }
 
 type StatusWriter struct {
-	statuses map[types.NamespacedName]Status
+	proxyStatuses map[types.NamespacedName]ProxyStatusUpdate
 }
 
-type ObjectStatusWriter struct {
+type ProxyStatusWriter struct {
 	sw     *StatusWriter
 	obj    k8s.Object
 	values map[string]string
 }
 
-// WithObject returns an ObjectStatusWriter that can be used to set the state of
+// WithProxy returns an ProxyStatusWriter that can be used to set the state of
 // the object. The state can be set as many times as necessary. The state of the
-// object can be made permanent by calling the commit function returned from WithObject.
-// The caller should pass the ObjectStatusWriter to functions interested in writing status,
+// object can be made permanent by calling the commit function returned from WithProxy.
+// The caller should pass the ProxyStatusWriter to functions interested in writing status,
 // but keep the commit function for itself. The commit function should be either called
 // via a defer, or directly if statuses are being set in a loop (as defers will not fire
 // until the end of the function).
-func (sw *StatusWriter) WithObject(obj k8s.Object) (_ *ObjectStatusWriter, commit func()) {
-	osw := &ObjectStatusWriter{
+func (sw *StatusWriter) WithProxy(obj k8s.Object) (_ *ProxyStatusWriter, commit func()) {
+	osw := &ProxyStatusWriter{
 		sw:     sw,
 		obj:    obj,
 		values: make(map[string]string),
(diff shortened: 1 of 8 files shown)


youngnick:
Reviewers, I'd like some feedback on the approach here. I'm aiming to get the call-site change in here, mostly, so that changes to the processor are kept to a minimum in the future. I don't see that the call-site needs to supply any more info than what we're doing here, given that we're passing the object itself as well.

Calculating `observedGeneration` etc can be done by the actual `commit()` function after we add in actual Condition support.

internal/dag/status.go
```diff
@@ -21,33 +21,33 @@ import (
...
-// object can be made permanent by calling the commit function returned from WithObject.
-// The caller should pass the ObjectStatusWriter to functions interested in writing status,
+// object can be made permanent by calling the commit function returned from WithProxy.
+// The caller should pass the ProxyStatusWriter to functions interested in writing status,
 // but keep the commit function for itself. The commit function should be either called
 // via a defer, or directly if statuses are being set in a loop (as defers will not fire
 // until the end of the function).
-func (sw *StatusWriter) WithObject(obj k8s.Object) (_ *ObjectStatusWriter, commit func()) {
-	osw := &ObjectStatusWriter{
+func (sw *StatusWriter) WithProxy(obj k8s.Object) (_ *ProxyStatusWriter, commit func()) {
```
jpeach, line 49: If this is only for HTTPProxy objects now, then pass a `*HTTPProxy` argument?

jpeach, line 49: Remove the named returns?

internal/dag/status.go
```diff
@@ -67,49 +67,50 @@ func (sw *StatusWriter) commit(osw *ObjectStatusWriter) {
...
-	return osw
+
+func (psw *ProxyStatusWriter) WithValue(key, val string) *ProxyStatusWriter {
+	psw.values[key] = val
+	return psw
 }
 
-func (osw *ObjectStatusWriter) SetInvalid(format string, args ...interface{}) {
-	osw.WithValue("description", fmt.Sprintf(format, args...)).WithValue("status", k8s.StatusInvalid)
+func (psw *ProxyStatusWriter) AddInvalidCondition(conditionType, formattedReason string, args ...interface{}) {
```
jpeach, line 86: Need some comment or example to explain what this will eventually do with a `DetailedCondition`

jpeach:
I'm unsure about this approach. Did you consider trying to introduce a parallel Condition-oriented API, then migrate `StatusWriter` onto that? It's not really clear to me how this change extends to Conditions; maybe just need some more explanation?

youngnick:
I did consider a separate API, but whatever I did, I needed to update the call sites for the errors to include a Reason, as Reason and Message are the two critical pieces of information we need from where the errors are actually generated inside the DAG.

So, the intent is that:
- the processors can add any number of conditions that describe how a HTTPProxy is invalid.
- we should pass the minimum amount of information to be able to do this. The Proxy object, a Reason, and Message are enough to populate any Condition (or subtype thereof).
- This mechanism should be reasonably easily extensible to other types of objects.
- these conditions will be implemented as sub-conditions under a `Valid` condition, but this is an implementation detail.

I *could* implement this by creating a new API first, having the code be present and not active, and then coming back and doing this PR afterwards, true. But I'm still going to need to do either this or something similar to update the call sites anyway. So I thought, better to update the call sites first, and then the later PRs will be easier to review because less change.

I can do it the other way around if you really want.

internal/dag/status.go
youngnick: It will eventually be used to construct a SubCondition under the `Valid` DetailedCondition.

youngnick:
I talked to @jpeach some more on Slack about this, this approach needs quite a bit of retooling. Closing.

PR #2890, opened 2020-09-08
internal: add HTTProxy CORS support

### LINKED ISSUE #437: Feature request: CORS support

aberasarte:
Closes #437


apis/projectcontour/v1/httpproxy.go
```diff
@@ -146,6 +149,26 @@ type TLS struct {
 	EnableFallbackCertificate bool `json:"enableFallbackCertificate,omitempty"`
 }
 
+// CorsPolicy allows setting de CORS policy
```
skriss, line 152: nit: s/de/the

apis/projectcontour/v1/httpproxy.go
```diff
@@ -146,6 +149,26 @@ type TLS struct {
...
+	AllowCredentials bool `json:"allowCredentials"`
+	// AllowOrigin specifies the origins that will be allowed to do CORS requests.
+	AllowOrigin []string `json:"allowOrigin"`
+	// AllowMethods specifies the content for the *access-control-allow-methods* header.
+	AllowMethods []string `json:"allowMethods"`
+	// AllowHeaders specifies the content for the *access-control-allow-headers* header.
+	//  +optional
+	AllowHeaders []string `json:"allowHeaders"`
+	//  +optional
+	// ExposeHeaders Specifies the content for the *access-control-expose-headers* header.
```
skriss, line 165: let's move the `// +optional` tag below the godoc for the field

apis/projectcontour/v1/httpproxy.go
```diff
@@ -146,6 +149,26 @@ type TLS struct {
...
+	ExposeHeaders []string `json:"exposeHeaders"`
+	//  +optional
+	// MaxAge specifies the content for the *access-control-max-age* header.
```
skriss, line 168: let's move the `// +optional` tag below the godoc for the field

apis/projectcontour/v1/httpproxy.go
```diff
@@ -146,6 +149,26 @@ type TLS struct {
 	EnableFallbackCertificate bool `json:"enableFallbackCertificate,omitempty"`
 }
 
+// CorsPolicy allows setting de CORS policy
+type CorsPolicy struct {
```
skriss, line 153: I think that for any fields that are optional, we want to add `omitempty` to the JSON tag.

internal/dag/httpproxy_processor.go
```diff
@@ -700,6 +709,27 @@ func determineSNI(routeRequestHeaders *HeadersPolicy, clusterRequestHeaders *Hea
...
+	if policy == nil {
+		return nil, nil
+	}
+	maxAge := timeout.DefaultSetting()
+	if perTryMaxAge, err := time.ParseDuration(policy.MaxAge); err == nil {
+		maxAge = timeout.DurationSetting(perTryMaxAge)
+		if maxAge.Duration().Seconds() < 0 {
+			return nil, fmt.Errorf("invalid max age value: %q", policy.MaxAge)
+		}
+	}
```
skriss, line 722: Do you want to support disabling this setting? Per https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age, the value `-1` disables the timeout.

If so, the standard for other Contour timeouts is that the input string `infinity` disables the timeout. The `timeout.Parse` function already handles parsing this.

internal/envoy/route.go
```diff
@@ -294,6 +298,43 @@ func RouteConfiguration(name string, virtualhosts ...*envoy_api_v2_route.Virtual
...
+			Value: cp.AllowCredentials,
+		},
+		AllowHeaders:  strings.Join(cp.AllowHeaders, ","),
+		AllowMethods:  strings.Join(cp.AllowMethods, ","),
+		ExposeHeaders: strings.Join(cp.ExposeHeaders, ","),
+	}
+
+	if !cp.MaxAge.UseDefault() {
+		rcp.MaxAge = fmt.Sprintf("%.0f", cp.MaxAge.Duration().Seconds())
+	}
```
skriss, line 325: If you want to support disabling MaxAge (see other comment), then you'd also want to check for `cp.MaxAge.IsDisabled()` here, and if true, pass a value of `-1` to Envoy.

skriss (changes requested):
Thanks for the PR @aberasarte! I added a few initial comments.

internal/featuretests/corspolicy_test.go
```diff
@@ -0,0 +1,341 @@
...
+				}},
+			}},
+		}),
+	)
+
+	c.Request(routeType).Equals(&v2.DiscoveryResponse{
+		Resources: resources(t,
+			envoy.RouteConfiguration("ingress_http")),
+		TypeUrl: routeType,
+	})
```
jpeach, line 339: You can check additional status here too:
```Go
c.Request(routeType).Equals(
    ...
).Status(obj).Like(...)
```

It's a bit inconvenient to have to pass an object here though.

jpeach (changes requested):
This looks like it's in pretty good shape to me. I have a few comments. I'll be out of office for a couple of weeks, so happy to defer to @skriss on reviewing subsequent iterations.

Thanks @aberasarte :)

internal/envoy/route.go
```diff
@@ -269,7 +272,7 @@ func weightedClusters(clusters []*dag.Cluster) *envoy_api_v2_route.WeightedClust
 }
 
 // VirtualHost creates a new route.VirtualHost.
-func VirtualHost(hostname string, routes ...*envoy_api_v2_route.Route) *envoy_api_v2_route.VirtualHost {
+func VirtualHost(hostname string, corspolicy *envoy_api_v2_route.CorsPolicy, routes ...*envoy_api_v2_route.Route) *envoy_api_v2_route.VirtualHost {
```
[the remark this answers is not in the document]
aberasarte: I like it. I'll change it if @skriss agrees.

internal/envoy/route.go
```diff
@@ -294,6 +298,43 @@ func RouteConfiguration(name string, virtualhosts ...*envoy_api_v2_route.Virtual
...
+
+	if !cp.MaxAge.UseDefault() {
+		rcp.MaxAge = fmt.Sprintf("%.0f", cp.MaxAge.Duration().Seconds())
+	}
+
+	rcp.AllowOriginStringMatch = []*matcher.StringMatcher{}
+	for _, ao := range cp.AllowOrigin {
+		rcp.AllowOriginStringMatch = append(rcp.AllowOriginStringMatch, &matcher.StringMatcher{
+			MatchPattern: &matcher.StringMatcher_Exact{
+				Exact: ao,
```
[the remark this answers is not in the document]
aberasarte: Yep, definitely.

internal/envoy/route.go
```diff
@@ -294,6 +298,43 @@ func RouteConfiguration(name string, virtualhosts ...*envoy_api_v2_route.Virtual
...
+		AllowMethods:  strings.Join(cp.AllowMethods, ","),
+		ExposeHeaders: strings.Join(cp.ExposeHeaders, ","),
+	}
+
+	if !cp.MaxAge.UseDefault() {
+		rcp.MaxAge = fmt.Sprintf("%.0f", cp.MaxAge.Duration().Seconds())
+	}
+
+	rcp.AllowOriginStringMatch = []*matcher.StringMatcher{}
+	for _, ao := range cp.AllowOrigin {
```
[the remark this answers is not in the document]
aberasarte: I think that Envoy always makes an exception for `*` so it should work even though we use the exact match. https://github.com/envoyproxy/envoy/blob/d6e2fd0185ca620745479da2c43c0564eeaf35c5/source/extensions/filters/http/cors/cors_filter.cc#L142.

internal/envoy/route.go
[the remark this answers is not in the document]
skriss: Sounds reasonable to me

internal/envoy/route.go
```diff
@@ -294,6 +298,43 @@ func RouteConfiguration(name string, virtualhosts ...*envoy_api_v2_route.Virtual
...
+		return nil
+	}
+	rcp := &envoy_api_v2_route.CorsPolicy{
+		EnabledSpecifier: &envoy_api_v2_route.CorsPolicy_FilterEnabled{
+			FilterEnabled: &envoy_api_v2_core.RuntimeFractionalPercent{
+				DefaultValue: &envoy_type.FractionalPercent{
+					Numerator:   100,
+					Denominator: envoy_type.FractionalPercent_HUNDRED,
+				},
+			},
```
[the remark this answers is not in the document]
aberasarte: I think that we should keep using `filter_enabled` as `enabled` only applies to routes and according to the docs, [it is deprecated:](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route_components.proto.html?highlight=allow_origin#envoy-api-field-route-corspolicy-enabled)

>enabled
(BoolValue) Specifies if the CORS filter is enabled. Defaults to true. Only effective on route.
Attention
This field is deprecated. Set the filter_enabled field instead.

internal/envoy/listener.go
```diff
@@ -207,6 +207,9 @@ func (b *httpConnectionManagerBuilder) DefaultFilters() *httpConnectionManagerBu
 		&http.HttpFilter{
 			Name: wellknown.GRPCWeb,
 		},
+		&http.HttpFilter{
+			Name: wellknown.CORS,
+		},
```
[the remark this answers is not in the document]
aberasarte: After testing it, I think that if we don't set the `EnabledSpecifier` property of `&envoy_api_v2_route.CorsPolicy`, the CORS filter will be enabled by default. When CORS configuration is omitted, we are not setting any CORS policy to the VirtualHost therefore, the filter is disabled.

internal/envoy/route.go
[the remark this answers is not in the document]
aberasarte: I've ended up leaving the `EnabledSpecified` property unset as the CORS filter is enabled by default (see my comment above).

apis/projectcontour/v1/httpproxy.go
```diff
@@ -146,6 +149,26 @@ type TLS struct {
...
+	// Specifies whether the resource allows credentials.
+	//  +optional
+	AllowCredentials bool `json:"allowCredentials"`
+	// AllowOrigin specifies the origins that will be allowed to do CORS requests.
+	AllowOrigin []string `json:"allowOrigin"`
+	// AllowMethods specifies the content for the *access-control-allow-methods* header.
+	AllowMethods []string `json:"allowMethods"`
+	// AllowHeaders specifies the content for the *access-control-allow-headers* header.
+	//  +optional
+	AllowHeaders []string `json:"allowHeaders"`
```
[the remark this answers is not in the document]
aberasarte: We can't apply this validation pattern for the `AllowOrigin` header as URLs like `http://example.com` are valid values. @skriss should I apply it for the rest of the headers and leave `AllowOrigin` as a `[]string`?

internal/dag/httpproxy_processor.go
aberasarte: > For practical purposes, a max-age of -1 and 0 seem like they would be treated similarly?

Both 0 and -1 mean disable caching for the preflight request but I decided to support just the 0 value. I found tricky to handle the -1  taking into account that the durations are expressed in the Go duration format. If we wanted to support -1 how would we do that? -1s disables caching but -1h is an invalid value? I'd rather allow negative values and assume that all of them disable the cache.

internal/envoy/route.go
aberasarte: Now I'm doing that but for the `0` value.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -146,6 +149,26 @@ type TLS struct {
...
+	//  +optional
+	// ExposeHeaders Specifies the content for the *access-control-expose-headers* header.
+	ExposeHeaders []string `json:"exposeHeaders"`
+	//  +optional
+	// MaxAge specifies the content for the *access-control-max-age* header.
+	MaxAge string `json:"maxAge"`
```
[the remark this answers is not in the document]
aberasarte: done!

internal/dag/dag.go
```diff
@@ -270,6 +286,8 @@ type VirtualHost struct {
 	// Name is the fully qualified domain name of a network host,
 	// as defined by RFC 3986.
 	Name string
+	// CorsPolicy is the cross-origin policy to apply to the VirtualHost.
```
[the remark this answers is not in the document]
aberasarte: done!

internal/featuretests/corspolicy_test.go
aberasarte: done!

aberasarte:
I think that all the comments have been addressed so it should be ready for a second review, PTAL @skriss . If everything looks good, I'll rebase the `main` branch bringing the latest changes.

skriss:
@aberasarte I'll take another look through this afternoon; thanks for the updates!

internal/envoy/route.go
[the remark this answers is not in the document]
skriss: It's worth a comment here about that since it's non-obvious.

skriss (changes requested):
@aberasarte this is looking pretty good, I just had a few more comments. Feel free to rebase when you're ready.

I am taking one more look at the `MaxAge` field just to make sure we have reasonable/consistent API semantics, I'll add another comment on that shortly.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -146,6 +149,34 @@ type TLS struct {
...
+	// ExposeHeaders Specifies the content for the *access-control-expose-headers* header.
+	//  +optional
+	ExposeHeaders []CORSHeaderValue `json:"exposeHeaders,omitempty"`
+	// MaxAge indicates for how long the results of a preflight request can be cached.
+	// MaxAge durations are expressed in the Go [Duration format](https://godoc.org/time#ParseDuration).
+	// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
+	// Only possitive values are allowed while 0 disables the cache requiring a preflight OPTIONS
+	// check for all cross-origin requests.
+	//  +optional
+	MaxAge string `json:"maxAge,omitempty"`
```
skriss, line 177: Normally, users have 3 different options for fields like this:

1. Specify a positive duration - Contour configures that specific value in Envoy
2. Don't specify a value (or specify 0) - Contour **does not** configure any value for that field in Envoy, which tells Envoy to use its default behavior
3. Specify `infinity` - Contour (usually) explicitly passes a 0 to Envoy for that field, which tells Envoy to disable the timeout in question.

It'd be best if we could have similar semantics for this field. The challenge here seems to be that the word "infinity" doesn't really make sense to mean "disable caching".

I can think of a few different options:
- empty string means "don't configure the field", explicit "0" means "disable caching"
- empty string or 0 means "don't configure the field", "-1" or any negative duration string means "disable caching"
- empty string or 0 means "don't configure the field", the string "disabled" means "disable caching"

Personally I think I prefer the third of these (i.e. support "disabled" as a magic string), but I'd like to get @stevesloka  and @youngnick's opinions here.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -146,6 +149,34 @@ type TLS struct {
 	EnableFallbackCertificate bool `json:"enableFallbackCertificate,omitempty"`
 }
 
+// CORSHeaderValue specifies the value of the string headers returned by a cross-domain request.
+// +kubebuilder:validation:Pattern=^[a-zA-Z0-9!#$%&'*+.^_\x60|~-]+$
```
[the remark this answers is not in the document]
aberasarte: 🤦

apis/projectcontour/v1/httpproxy.go
```diff
@@ -146,6 +149,34 @@ type TLS struct {
...
+
+// CORSPolicy allows setting the CORS policy
+type CORSPolicy struct {
+	// Specifies whether the resource allows credentials.
+	//  +optional
+	AllowCredentials bool `json:"allowCredentials,omitempty"`
+	// AllowOrigin specifies the origins that will be allowed to do CORS requests.
+	AllowOrigin []string `json:"allowOrigin"`
+	// AllowMethods specifies the content for the *access-control-allow-methods* header.
+	AllowMethods []CORSHeaderValue `json:"allowMethods"`
```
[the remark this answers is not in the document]
aberasarte: Yes, I meant to make them required, so I'll add the kubebuilder validations.

apis/projectcontour/v1/httpproxy.go
aberasarte: I prefer the first option because in most cache systems 0 means "disable caching" and using that value for "don't configure the field" could be misleading.

youngnick: That's a fair point @aberasarte, but I'd also like to keep it a bit consistent with the rest of Contour. WDYT about "0" or "disabled" means "disable caching"? If that's a no, I'm okay with just "0".

aberasarte: > WDYT about "0" or "disabled" means "disable caching"?

That sounds good to me but then I would need a way to identify the "don't configure the field" option. `UseDefault`  would work [but at the moment 0 is treated as UseDefault](https://github.com/projectcontour/contour/blob/main/internal/timeout/timeout.go#L37) and that's not what we want here. 

I could create a different package (`maxage`? `expiration`?) for handling this kind of values. It would be similar to https://github.com/projectcontour/contour/blob/main/internal/timeout/timeout.go but it would allow the "disabled" value instead of "infinity" and it wouldn't treat `0` as `UseDefault`. WDYT?

jpeach: This seems pretty clear to me. The field is called `MaxAge`, so a maximum age of 0 would disable caching, and a maximum age of infinity would cache forever. To distinguish between the default and 0 as an explictly specified age, you can use `MaxAge *string`. It also seems pretty clear that a value of "" would have a different result than a value of "0s".

aberasarte: > The field is called MaxAge, so a maximum age of 0 would disable caching, and a maximum age of infinity would cache forever.

I wouldn't allow `infinity` as it's potentially dangerous. Browsers even have an imposed limit for this value. [From the spec](https://fetch.spec.whatwg.org/#cors-preflight-fetch):

> If max-age is greater than an imposed limit on max-age, then set max-age to the imposed limit.

I think that we should encourage people to use lower values.

> To distinguish between the default and 0 as an explictly specified age, you can use MaxAge *string. It also seems pretty clear that a value of "" would have a different result than a value of "0s".

I totally agree. The problem is that we turn that string into a Golang duration, so we need a way for differentiating whether that value has been set. At the moment https://github.com/projectcontour/contour/blob/main/internal/timeout/timeout.go is used for the same purpose but for other properties. I was trying to use the same logic for `MaxAge` but if we want to allow the "disabled" value without breaking the current behavior, it would probably be better to create a different utility package.

skriss: I'm satisfied with "0s" explicitly specified means "disabled", "" (no value specified) means "use default". We'll need custom parse logic instead of the standard `timeout.Parse` to properly handle those cases, but we can still use `timeout.Setting` values internally to represent it.

youngnick: Yes, that seems fine to me.

aberasarte: >I'm satisfied with "0" explicitly specified means "disabled", "" (no value specified) means "use default". We'll need custom parse logic instead of the standard timeout.Parse to properly handle those cases, but we can still use timeout.Setting values internally to represent it. I don't think we need to use a string pointer either, we can just check whether the field was (a) empty or (b) a non-empty representation of "0".

Perfect, then we need to change the logic [of the UseDefault() method](https://github.com/projectcontour/contour/blob/main/internal/timeout/timeout.go#L37) as well. Instead of checking `s.val == 0` for deciding whether a value has been set, I could add a new property to the `Setting` struct for that. Something in the lines of:

```
type Setting struct {
	val        time.Duration
	disabled   bool
	usedefault bool
}
```

And check that property in the `UseDefault` method:

```
func (s Setting) UseDefault() bool {
	return !s.disabled && s.usedefault
}
```

The current `Parse` function would set that property to `true` if the value is `0`, so nothing should break. I would add a new parsing function (`ParseMaxeAge`?) that would be similar to `Parse` but this one would set `usedefault` to `true` only if the received string is empty `""`.

Would that work for you?

skriss: Yep, that sounds reasonable to me.

youngnick:
@aberasarte, this seems pretty close! If you can get this merged in the next 7 days, we should be able to get it into Contour 1.9. Seems like it just needs a rebase and the MaxAge fix? @skriss is that the only todo left?

skriss:
I had a few other comments that still need to be addressed, but it was nothing major. Agree that we should be able to get this into 1.9!

apis/projectcontour/v1/httpproxy.go
aberasarte: I inverted the logic in order to make the change backwards compatible. The new property is called `valset`.

aberasarte:
All comments addressed and `main` rebased. PTAL.

apis/projectcontour/v1/httpproxy.go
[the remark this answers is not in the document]
skriss: Looks like this still needs to be addressed.

internal/timeout/timeout.go
```diff
@@ -82,5 +86,35 @@ func Parse(timeout string) (Setting, error) {
...
+// in cache related settings. An example of this is the MaxAge field of the
+// CORS policy:
+//	- an empty string means "use the default".
+//	- the "disabled" string means "disable cache", so it is turned into a 0 duration.
+//	- a valid Go duration string is used as the specific timeout value.
+//	- any other input means "use the default".
+func ParseMaxAge(timeout string) (Setting, error) {
+	if timeout == "disabled" {
+		return DisabledSetting(), nil
+	}
```
skriss, line 106: I think consensus was we don't to support this magic string, since 0 explicitly specified means "disable cache". Please update the godoc to match as well.

skriss (changes requested):
Just one minor thing, plus it looks like one more rebase is needed (sorry, we've been merging a lot of big changes lately).

internal/timeout/timeout.go
aberasarte: I thought that supporting both was preferred 😅 . I'll change it.

skriss: sorry for the back and forth!

apis/projectcontour/v1/httpproxy.go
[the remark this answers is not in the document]
aberasarte: done!

internal/timeout/timeout.go
aberasarte: done and main rebased.

skriss (approved):
No further comments, LGTM! @stevesloka or @youngnick it'd be great if you could take a look as well.

I have filed https://github.com/projectcontour/contour/issues/2943 as a followup issue to add some more documentation on this new feature, it'd be great if we could get this in by mid-next week in order to have it in the upcoming 1.9 release.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -241,6 +244,37 @@ type TLS struct {
...
+	// AllowHeaders specifies the content for the *access-control-allow-headers* header.
+	//  +optional
+	AllowHeaders []CORSHeaderValue `json:"allowHeaders,omitempty"`
+	// ExposeHeaders Specifies the content for the *access-control-expose-headers* header.
+	//  +optional
+	ExposeHeaders []CORSHeaderValue `json:"exposeHeaders,omitempty"`
+	// MaxAge indicates for how long the results of a preflight request can be cached.
+	// MaxAge durations are expressed in the Go [Duration format](https://godoc.org/time#ParseDuration).
+	// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
+	// Only possitive values are allowed while 0 disables the cache requiring a preflight OPTIONS
```
stevesloka, line 272: nit: /possitive/positive/

stevesloka (approved):
Just a small nit, but otherwise lgtm! Thanks for the work here @aberasarte!! 🎉

youngnick:
Sorry @aberasarte, one more rebase and that nit, and I'll merge this for you. Yay!

apis/projectcontour/v1/httpproxy.go
aberasarte: done!

aberasarte:
It should be ready for merge now 🤞 .

skriss:
integration tests flaking on SNI. Merging.

glerchundi:
I cannot believe this have been finally merged. Titanic work @aberasarte, thanks for pushing this forward and for being very responsive throughout all the design & impl process. Also thanks to all the reviewers for the time you dedicated here. 👏🏻👏🏻👏🏻

PR #2894, opened 2020-09-09
internal/envoy: Set the dns lookup family on externalName type clusters

### LINKED ISSUE #2873: proxy to externalName service returns 503
When configuring a route to an `externalName` service, requests don't get fulfilled properly, but return 503 errors. My first thought is that Envoy is not proxying when its DNS lookup returns an ipv6 address for the external resource. 

Below are some examples which recreate the issue. One thing to note is if you use my domain (e.g. `stevesloka.com`), it results in an ipv4 address and the proxy works just fine. My cluster is running in DigitalOcean proxying to a Node directly fronted with Contour. 

HTTPProxy:
```
kind: HTTPProxy
metadata:
  name: my-ingress
spec:
  virtualhost:
    fqdn: baremetal.slokalabs.io
    # tls:
    #   secretName: steveslokadev-tls
  routes:
    - conditions:
        - prefix: /
      services:
        - name: external-svc
          port: 443
          protocol: tls
      requestHeadersPolicy:
        set:
        - name: Host
          value: google.com
    - conditions:
        - prefix: /potato
      services:
        - name: rootapp
          port: 80
```

Service:
```
apiVersion: v1
kind: Service
metadata:
  name: external-svc
  annotations:
    projectcontour.io/upstream-protocol.tls: "external-svc,443"
spec:
  type: ExternalName
  externalName: google.com
  ports:
    - name: external-svc
      port: 443
      targetPort: 443
```

Logs from Envoy:

```
[2020-09-02 20:19:12.544][1][debug][upstream] [source/common/upstream/upstream_impl.cc:278] transport socket match, socket default selected for host with address [2607:f8b0:4004:807::200e]:443
[2020-09-02 20:19:12.545][1][debug][upstream] [source/common/upstream/strict_dns_cluster.cc:167] DNS refresh rate reset for google.com, refresh rate 5000 ms
[2020-09-02 20:19:12.593][1][debug][upstream] [source/common/upstream/logical_dns_cluster.cc:100] starting async DNS resolution for 127.0.0.1
[2020-09-02 20:19:12.593][1][debug][upstream] [source/common/upstream/logical_dns_cluster.cc:108] async DNS resolution complete for 127.0.0.1
[2020-09-02 20:19:12.593][1][debug][upstream] [source/common/upstream/logical_dns_cluster.cc:153] DNS refresh rate reset for 127.0.0.1, refresh rate 5000 ms
[2020-09-02 20:19:12.653][1][debug][main] [source/server/server.cc:189] flushing stats
[2020-09-02 20:19:13.225][1][debug][upstream] [source/common/upstream/upstream_impl.cc:278] transport socket match, socket default selected for host with address 10.100.207.214:8001
[2020-09-02 20:19:13.225][1][debug][upstream] [source/common/upstream/strict_dns_cluster.cc:167] DNS refresh rate reset for contour, refresh rate 5000 ms
[2020-09-02 20:19:16.055][17][debug][conn_handler] [source/server/connection_handler_impl.cc:422] [C19] new connection
[2020-09-02 20:19:16.055][12][debug][conn_handler] [source/server/connection_handler_impl.cc:422] [C20] new connection
[2020-09-02 20:19:16.055][17][debug][http] [source/common/http/conn_manager_impl.cc:261] [C19] new stream
[2020-09-02 20:19:16.056][17][debug][http] [source/common/http/conn_manager_impl.cc:808] [C19][S8889813871901120571] request headers complete (end_stream=true):
':authority', 'baremetal.slokalabs.io'
':path', '/'
':method', 'GET'
'connection', 'keep-alive'
'cache-control', 'max-age=0'
'upgrade-insecure-requests', '1'
'user-agent', 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/84.0.4147.135 Safari/537.36'
'accept', 'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9'
'accept-encoding', 'gzip, deflate'
'accept-language', 'en-US,en;q=0.9'
'cookie', '_ga=GA1.2.702597985.1599075532; _gid=GA1.2.763115516.1599075532'

[2020-09-02 20:19:16.056][17][debug][http] [source/common/http/conn_manager_impl.cc:1377] [C19][S8889813871901120571] request end stream
[2020-09-02 20:19:16.056][17][debug][router] [source/common/router/router.cc:426] [C19][S8889813871901120571] cluster 'projectcontour-roots/external-svc/443/da39a3ee5e' match for URL '/'
[2020-09-02 20:19:16.056][17][debug][router] [source/common/router/router.cc:583] [C19][S8889813871901120571] router decoding headers:
':authority', 'google.com'
':path', '/'
':method', 'GET'
':scheme', 'https'
'cache-control', 'max-age=0'
'upgrade-insecure-requests', '1'
'user-agent', 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/84.0.4147.135 Safari/537.36'
'accept', 'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9'
'accept-encoding', 'gzip, deflate'
'accept-language', 'en-US,en;q=0.9'
'cookie', '_ga=GA1.2.702597985.1599075532; _gid=GA1.2.763115516.1599075532'
'x-forwarded-for', '10.52.131.186'
'x-forwarded-proto', 'http'
'x-envoy-internal', 'true'
'x-request-id', '87cf37cf-e332-4c21-8c4a-e42967919e7f'
'x-envoy-expected-rq-timeout-ms', '15000'
'x-request-start', 't=1599077956.056'

[2020-09-02 20:19:16.056][17][debug][pool] [source/common/http/conn_pool_base.cc:71] queueing request due to no available connections
[2020-09-02 20:19:16.056][17][debug][pool] [source/common/conn_pool/conn_pool_base.cc:53] creating a new connection
[2020-09-02 20:19:16.056][17][debug][client] [source/common/http/codec_client.cc:35] [C21] connecting
[2020-09-02 20:19:16.056][17][debug][connection] [source/common/network/connection_impl.cc:753] [C21] connecting to [2607:f8b0:4004:807::200e]:443
[2020-09-02 20:19:16.056][17][debug][connection] [source/common/network/connection_impl.cc:773] [C21] immediate connection error: 101
[2020-09-02 20:19:16.056][17][debug][connection] [source/common/network/connection_impl.cc:517] [C21] raising immediate error
[2020-09-02 20:19:16.056][17][debug][connection] [source/common/network/connection_impl.cc:208] [C21] closing socket: 0
[2020-09-02 20:19:16.056][17][debug][client] [source/common/http/codec_client.cc:92] [C21] disconnect. resetting 0 pending requests
[2020-09-02 20:19:16.056][17][debug][pool] [source/common/conn_pool/conn_pool_base.cc:255] [C21] client disconnected, failure reason: 
[2020-09-02 20:19:16.056][17][debug][router] [source/common/router/router.cc:1022] [C19][S8889813871901120571] upstream reset: reset reason connection failure
[2020-09-02 20:19:16.056][17][debug][http] [source/common/http/conn_manager_impl.cc:1520] [C19][S8889813871901120571] Sending local reply with details upstream_reset_before_response_started{connection failure}
[2020-09-02 20:19:16.056][17][debug][http] [source/common/http/conn_manager_impl.cc:1777] [C19][S8889813871901120571] encoding headers via codec (end_stream=false):
':status', '503'
'content-type', 'text/plain'
'vary', 'Accept-Encoding'
'content-encoding', 'gzip'
'date', 'Wed, 02 Sep 2020 20:19:16 GMT'
'server', 'envoy'

[2020-09-02 20:19:16.218][17][debug][http] [source/common/http/conn_manager_impl.cc:261] [C19] new stream
[2020-09-02 20:19:16.218][17][debug][http] [source/common/http/conn_manager_impl.cc:808] [C19][S10594451152967010302] request headers complete (end_stream=true):
':authority', 'baremetal.slokalabs.io'
':path', '/favicon.ico'
':method', 'GET'
'connection', 'keep-alive'
'pragma', 'no-cache'
'cache-control', 'no-cache'
'user-agent', 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/84.0.4147.135 Safari/537.36'
'accept', 'image/webp,image/apng,image/*,*/*;q=0.8'
'referer', 'http://baremetal.slokalabs.io/'
'accept-encoding', 'gzip, deflate'
'accept-language', 'en-US,en;q=0.9'
'cookie', '_ga=GA1.2.702597985.1599075532; _gid=GA1.2.763115516.1599075532'

[2020-09-02 20:19:16.218][17][debug][http] [source/common/http/conn_manager_impl.cc:1377] [C19][S10594451152967010302] request end stream
[2020-09-02 20:19:16.218][17][debug][router] [source/common/router/router.cc:426] [C19][S10594451152967010302] cluster 'projectcontour-roots/external-svc/443/da39a3ee5e' match for URL '/favicon.ico'
[2020-09-02 20:19:16.218][17][debug][router] [source/common/router/router.cc:583] [C19][S10594451152967010302] router decoding headers:
':authority', 'google.com'
':path', '/favicon.ico'
':method', 'GET'
':scheme', 'https'
'pragma', 'no-cache'
'cache-control', 'no-cache'
'user-agent', 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/84.0.4147.135 Safari/537.36'
'accept', 'image/webp,image/apng,image/*,*/*;q=0.8'
'referer', 'http://baremetal.slokalabs.io/'
'accept-encoding', 'gzip, deflate'
'accept-language', 'en-US,en;q=0.9'
'cookie', '_ga=GA1.2.702597985.1599075532; _gid=GA1.2.763115516.1599075532'
'x-forwarded-for', '10.52.131.186'
'x-forwarded-proto', 'http'
'x-envoy-internal', 'true'
'x-request-id', '77a6dd34-cd0b-43e3-9174-70d9842e53b2'
'x-envoy-expected-rq-timeout-ms', '15000'
'x-request-start', 't=1599077956.218'

[2020-09-02 20:19:16.218][17][debug][pool] [source/common/http/conn_pool_base.cc:71] queueing request due to no available connections
[2020-09-02 20:19:16.218][17][debug][pool] [source/common/conn_pool/conn_pool_base.cc:53] creating a new connection
[2020-09-02 20:19:16.218][17][debug][client] [source/common/http/codec_client.cc:35] [C22] connecting
[2020-09-02 20:19:16.218][17][debug][connection] [source/common/network/connection_impl.cc:753] [C22] connecting to [2607:f8b0:4004:807::200e]:443
[2020-09-02 20:19:16.218][17][debug][connection] [source/common/network/connection_impl.cc:773] [C22] immediate connection error: 101
[2020-09-02 20:19:16.219][17][debug][connection] [source/common/network/connection_impl.cc:517] [C22] raising immediate error
[2020-09-02 20:19:16.219][17][debug][connection] [source/common/network/connection_impl.cc:208] [C22] closing socket: 0
[2020-09-02 20:19:16.219][17][debug][client] [source/common/http/codec_client.cc:92] [C22] disconnect. resetting 0 pending requests
[2020-09-02 20:19:16.219][17][debug][pool] [source/common/conn_pool/conn_pool_base.cc:255] [C22] client disconnected, failure reason: 
[2020-09-02 20:19:16.219][17][debug][router] [source/common/router/router.cc:1022] [C19][S10594451152967010302] upstream reset: reset reason connection failure
[2020-09-02 20:19:16.219][17][debug][http] [source/common/http/conn_manager_impl.cc:1520] [C19][S10594451152967010302] Sending local reply with details upstream_reset_before_response_started{connection failure}
[2020-09-02 20:19:16.219][17][debug][http] [source/common/http/conn_manager_impl.cc:1777] [C19][S10594451152967010302] encoding headers via codec (end_stream=false):
':status', '503'
'content-type', 'text/plain'
'vary', 'Accept-Encoding'
'content-encoding', 'gzip'
'date', 'Wed, 02 Sep 2020 20:19:16 GMT'
'server', 'envoy'

[2020-09-02 20:19:17.545][1][debug][upstream] [source/common/upstream/upstream_impl.cc:278] transport socket match, socket default selected for host with address [2607:f8b0:4004:807::200e]:443
```

//cc @brunopadz

stevesloka:
Adds a config file option `Cluster.DnsLookupFamily` which allows users to define what dns lookup family is used for any
externalName type cluster. This ensures that lookups to external resources are resolved correctly.

Fixes #2873

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -296,6 +296,11 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	// register observer for endpoints updates.
 	endpointHandler.Observer = contour.ComposeObservers(snapshotHandler)
 
+	dnsLookupFamily, err := ParseDNSLookupFamily(ctx.DNSLookupFamily)
+	if err != nil {
+		return fmt.Errorf("failed to configure configuration file parameter cluster.dns-lookup-family: %w", err)
+	}
+
 	// Build the core Kubernetes event handler.
 	eventHandler := &contour.EventHandler{
 		HoldoffDelay:    100 * time.Millisecond,
@@ -317,6 +322,7 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 				&dag.HTTPProxyProcessor{
 					DisablePermitInsecure: ctx.DisablePermitInsecure,
 					FallbackCertificate:   fallbackCert,
+					DNSLookupFamily:       dnsLookupFamily,
 				},
 				&dag.ListenerProcessor{},
 			},
--- a/cmd/contour/servecontext.go
+++ b/cmd/contour/servecontext.go
@@ -147,6 +147,10 @@ type serveContext struct {
 	//
 	// If this field not specified, all supported versions are accepted.
 	DefaultHTTPVersions []string `yaml:"default-http-versions"`
+
+	// ClusterConfig holds various configurable Envoy cluster values that can
+	// be set in the config file.
+	ClusterConfig `yaml:"cluster,omitempty"`
 }
 
 // newServeContext returns a serveContext initialized to defaults.
@@ -191,6 +195,9 @@ func newServeContext() *serveContext {
 			xdsAddr:       "127.0.0.1",
 			xdsPort:       8001,
 			XDSServerType: "contour",
+		},
+		ClusterConfig: ClusterConfig{
+			DNSLookupFamily: "auto",
 		},
 	}
 }
@@ -299,6 +306,22 @@ type TimeoutConfig struct {
 	ConnectionShutdownGracePeriod string `yaml:"connection-shutdown-grace-period,omitempty"`
 }
 
+// ClusterConfig holds various configurable cluster values.
+type ClusterConfig struct {
+	// DNSLookupFamily defines how external names are looked up
+	// When configured as V4, the DNS resolver will only perform a lookup
+	// for addresses in the IPv4 family. If V6 is configured, the DNS resolver
+	// will only perform a lookup for addresses in the IPv6 family.
+	// If AUTO is configured, the DNS resolver will first perform a lookup
+	// for addresses in the IPv6 family and fallback to a lookup for addresses
+	// in the IPv4 family.
+	// Note: This only applies to externalName clusters.
+	//
+	// See https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/cluster.proto#enum-cluster-dnslookupfamily
+	// for more information.
+	DNSLookupFamily string `yaml:"dns-lookup-family"`
+}
+
 // grpcOptions returns a slice of grpc.ServerOptions.
 // if ctx.PermitInsecureGRPC is false, the option set will
 // include TLS configuration.
@@ -403,6 +426,20 @@ func (ctx *serveContext) proxyRootNamespaces() []string {
 		ns = append(ns, strings.TrimSpace(s))
 	}
 	return ns
+}
+
+// ParseDNSLookupFamily parses the dnsLookupFamily returning the value
+// configured or the default of "auto" if invalid.
+func ParseDNSLookupFamily(value string) (string, error) {
+
+	dlf := strings.ToLower(value)
+	switch dlf {
+	case "auto", "v4", "v6":
+		return dlf, nil
+	case "":
+		return "auto", nil
+	}
+	return "auto", fmt.Errorf("invalid dns lookup family %q configured, defaulting to \"auto\"", dlf)
 }
 
 // parseDefaultHTTPVersions parses a list of supported HTTP versions
--- a/cmd/contour/servecontext_test.go
+++ b/cmd/contour/servecontext_test.go
@@ -18,6 +18,7 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"net"
 	"os"
@@ -510,6 +511,49 @@ func TestParseHTTPVersions(t *testing.T) {
 
 			assert.Equal(t, testcase.parseError, err)
 			assert.Equal(t, testcase.parseVersions, vers)
+		})
+	}
+}
+
+func TestParseDNSLookupFamily(t *testing.T) {
+	cases := map[string]struct {
+		input         string
+		expectedError error
+		expected      string
+	}{
+		"empty": {
+			input:         "",
+			expectedError: nil,
+			expected:      "auto",
+		},
+		"auto": {
+			input:         "auto",
+			expectedError: nil,
+			expected:      "auto",
+		},
+		"v4": {
+			input:         "v4",
+			expectedError: nil,
+			expected:      "v4",
+		},
+		"v6": {
+			input:         "v6",
+			expectedError: nil,
+			expected:      "v6",
+		},
+		"invalid": {
+			input:         "abcdefg",
+			expectedError: fmt.Errorf("invalid dns lookup family \"abcdefg\" configured, defaulting to \"auto\""),
+			expected:      "auto",
+		},
+	}
+
+	for name, testcase := range cases {
+		testcase := testcase
+		t.Run(name, func(t *testing.T) {
+			got, err := ParseDNSLookupFamily(testcase.input)
+			assert.Equal(t, testcase.expectedError, err)
+			assert.Equal(t, testcase.expected, got)
 		})
 	}
 }
--- a/site/docs/main/api-reference.html
+++ b/site/docs/main/api-reference.html
@@ -1,12 +1,425 @@
 <p>Packages:</p>
 <ul>
 <li>
-<a href="#projectcontour.io%2fv1">projectcontour.io/v1</a>
-</li>
-<li>
 <a href="#projectcontour.io%2fv1alpha1">projectcontour.io/v1alpha1</a>
 </li>
+<li>
+<a href="#projectcontour.io%2fv1">projectcontour.io/v1</a>
+</li>
 </ul>
+<h2 id="projectcontour.io/v1alpha1">projectcontour.io/v1alpha1</h2>
+<p>
+<p>Package v1alpha1 contains API Schema definitions for the projectcontour.io v1alpha1 API group</p>
+</p>
+Resource Types:
+<ul><li>
+<a href="#projectcontour.io/v1alpha1.ExtensionService">ExtensionService</a>
+</li></ul>
+<h3 id="projectcontour.io/v1alpha1.ExtensionService">ExtensionService
+</h3>
+<p>
+<p>ExtensionService is the schema for the Contour extension services API.
+An ExtensionService resource binds a network service to the Contour
+API so that Contour API features can be implemented by collaborating
+components.</p>
+</p>
+<table class="table table-striped table-borderless" style="border:none">
+<thead class="border-bottom">
+<tr>
+<th>Field</th>
+<th>Description</th>
+</tr>
+</thead>
+<tbody class="border-top">
+<tr>
+<td>
+<code>apiVersion</code>
+<br>
+string</td>
+<td>
+<code>
+projectcontour.io/v1alpha1
+</code>
+</td>
+</tr>
+<tr>
+<td>
+<code>kind</code>
+<br>
+string
+</td>
+<td><code>ExtensionService</code></td>
+</tr>
+<tr>
+<td style="white-space:nowrap">
+<code>metadata</code>
+<br>
+<em>
+<a href="https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#objectmeta-v1-meta">
+Kubernetes meta/v1.ObjectMeta
+</a>
+</em>
+</td>
+<td>
+Refer to the Kubernetes API documentation for the fields of the
+<code>metadata</code> field.
+</td>
+</tr>
+<tr>
+<td style="white-space:nowrap">
+<code>spec</code>
+<br>
+<em>
+<a href="#projectcontour.io/v1alpha1.ExtensionServiceSpec">
+ExtensionServiceSpec
+</a>
+</em>
+</td>
+<td>
+<br>
+<br>
+<table style="border:none">
+<tr>
+<td style="white-space:nowrap">
+<code>services</code>
+<br>
+<em>
+<a href="#projectcontour.io/v1alpha1.ExtensionServiceTarget">
+[]ExtensionServiceTarget
+</a>
+</em>
+</td>
+<td>
+<p>Services specifies the set of Kubernetes Service resources that
+receive GRPC extension API requests.
+If no weights are specified for any of the entries in
+this array, traffic will be spread evenly across all the
+services.
+Otherwise, traffic is balanced proportionally to the
+Weight field in each entry.</p>
+</td>
+</tr>
+<tr>
+<td style="white-space:nowrap">
+<code>validation</code>
+<br>
+<em>
+<a href="#projectcontour.io/v1.UpstreamValidation">
+UpstreamValidation
+</a>
+</em>
+</td>
+<td>
+<em>(Optional)</em>
+<p>UpstreamValidation defines how to verify the backend service&rsquo;s certificate</p>
+</td>
+</tr>
+<tr>
+<td style="white-space:nowrap">
+<code>protocol</code>
+<br>
+<em>
+string
+</em>
+</td>
+<td>
+<em>(Optional)</em>
+<p>Protocol may be used to specify (or override) the protocol used to reach this Service.
+Values may be tls, h2, h2c. If omitted, protocol-selection falls back on Service annotations.</p>
+</td>
+</tr>
+<tr>
+<td style="white-space:nowrap">
+<code>loadBalancerPolicy</code>
+<br>
+<em>
+<a href="#projectcontour.io/v1.LoadBalancerPolicy">
+LoadBalancerPolicy
+</a>
+</em>
+</td>
+<td>
+<em>(Optional)</em>
+<p>The policy for load balancing GRPC service requests. Note
+that the <code>Cookie</code> load balancing strategy cannot be used here.</p>
+</td>
+</tr>
+<tr>
+<td style="white-space:nowrap">
+<code>timeoutPolicy</code>
+<br>
+<em>
+<a href="#projectcontour.io/v1.TimeoutPolicy">
+TimeoutPolicy
+</a>
+</em>
+</td>
+<td>
+<em>(Optional)</em>
+<p>The timeout policy for requests to the services.</p>
+</td>
+</tr>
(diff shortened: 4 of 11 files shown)


jpeach:
@stevesloka Can you give more detail about why v4-only is the right approach here? What was the underlying problem?

skriss:
Based on https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/cluster.proto#envoy-api-enum-cluster-dnslookupfamily, it sounds to me like this *should* be working properly (i.e. fallback to v4 resolution if v6 fails), so perhaps an Envoy bug?

I guess my only question re: this change is, are there any scenarios where specifying v4-only would cause issues for users by not supporting v6?

jpeach:
> @jpeach the issue describes the problem in more detail.
> 
> The TLDR is: When configuring a route to an externalName service, requests don't get fulfilled properly, but return 503 errors. When Envoy does a DNS lookup to some external resources, a ipv6 address is returned for the external resource and proxying fails with a 503 error. Setting this field ensures an IPv4 is returned so Envoy proxies correctly. I tested this in a GKE cluster as well as my home lab, both showed the same issue and was also resolved with the same fix.
> 
> Chatting with @moderation in an Envoy issue ([envoyproxy/envoy#13037 (comment)](https://github.com/envoyproxy/envoy/issues/13037#issuecomment-689777236)), he verified that v3 resolved the issue which we haven't yet moved to, or by setting the dns lookup family might also help.

Thanks for linking the Envoy issue. It's not clear to me from the issue whether the problem was that Envoy couldn't connect to the v6 target for some reason, or whether it hit an internal bug first. Possibly, the bug is that it does't fall back to v4 on a connection failure?

moderation:
To clarify I don't think changing to v3 fixed the issue, it was the addition of `dns_lookup_family: V4_ONLY`. Not sure hard coding this to V4 makes sense. Although rare I've heard of V6 only environments. Seems like something that should be configurable based on the environment where Contour is running. I think there could be quite a few permutation here too. A DNS that resolves V6 and V4 addresses but the network is V4. A dual stack network but with DNS that only resolves V4 etc.

stevesloka:
I had some more discussion in the upstream Envoy issue and think this isn't 100% an Envoy issue. The Envoy setting `dns_lookup_family` determines how Envoy will lookup dns entries. The default which Contour uses today is `AUTO`, meaning Envoy will try to find an ipv6 address first, then fallback to ipv4 addresses. 

Since some sites are returning ipv6 addresses, it's then up to the network capabilities of the cluster's network to allow the routing to work. I think this is why we're seeing the errors in the issue referenced. 

To fully resolve I think we should add a config setting which will allow users to determine which setting to pass to envoy, be that the ipv4, ipv6, or auto (which should remain the default). 

@youngnick @skriss thoughts?

youngnick:
I think that's the right way to do this, thanks @stevesloka.

stevesloka:
Ok this is updated with a new config file option. I named it "cluster", but not sure if that makes sense or not. This new setting defines what is used for the `dns_lookup_family` on any outgoing request to an externalName type cluster.

cmd/contour/serve.go
```diff
@@ -231,6 +231,11 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	// register observer for endpoints updates.
 	endpointHandler.Observer = contour.ComposeObservers(snapshotHandler)
 
+	dnsLookupFamily, err := ParseDNSLookupFamily(ctx.DNSLookupFamily)
+	if err != nil {
+		log.WithField("context", "dns-lookup-family").Infof("error parsing config parameter dns-lookup-config %q", ctx.DNSLookupFamily)
```
skriss, line 236: in other places within this function we return parse errors (e.g. timeouts, default HTTP versions) rather than just logging them - I think we should probably do the same here for consistency.

skriss, line 236: s/dns-lookup-config/dns-lookup-family

internal/envoy/cluster_test.go
```diff
@@ -147,13 +147,15 @@ func TestCluster(t *testing.T) {
 		},
 		"externalName service": {
 			cluster: &dag.Cluster{
-				Upstream: service(s2),
+				Upstream:        service(s2),
+				DNSLookupFamily: "v4",
```
skriss, line 151: it'd be nice to add test cases here for the other options

cmd/contour/servecontext_test.go
```diff
@@ -508,3 +509,46 @@ func TestParseHTTPVersions(t *testing.T) {
 		})
 	}
 }
+
+func TestParseDNSLookupFamily(t *testing.T) {
+	cases := map[string]struct {
+		input        string
+		expetedError error
```
skriss, line 516: nit: s/expetedError/expectedError

cmd/contour/servecontext.go
```diff
@@ -398,6 +421,24 @@ func (ctx *serveContext) proxyRootNamespaces() []string {
...
+	switch dlf {
+	case "auto":
+		fallthrough
+	case "v4":
+		fallthrough
+	case "v6":
+		return dlf, nil
+	case "":
+		return "auto", nil
+	}
```
skriss, line 438: Let's simplify this to get rid of `fallthrough`:

```go
switch dlf {
case "auto", "v4", "v6":
    return dlf, nil
case "":
    return "auto", nil
}
```

cmd/contour/servecontext.go
```diff
@@ -153,6 +153,10 @@ type serveContext struct {
 	//
 	// If this field not specified, all supported versions are accepted.
 	DefaultHTTPVersions []string `yaml:"default-http-versions"`
+
+	// ClusterConfig holds various configurable Envoy cluster values that can
+	// be set in the config file.
+	ClusterConfig `yaml:"cluster,omitempty"`
```
skriss, line 159: I don't really have a better suggestion for naming here. I guess this could just be a top-level field in the config file too. 🤷‍♂️

skriss (changes requested):
Approach looks reasonable to me -- it seems unlikely that this would need to be toggled on a per-route basis so a global config file setting makes sense.

Need to update `configuration.md` as well.

cmd/contour/servecontext.go
stevesloka: Yup that's where I started, but if we ever needed to add another "cluster" type config it's nice to have this section for it.

cmd/contour/serve.go
```diff
@@ -251,6 +251,11 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	// register observer for endpoints updates.
 	endpointHandler.Observer = contour.ComposeObservers(snapshotHandler)
 
+	dnsLookupFamily, err := ParseDNSLookupFamily(ctx.DNSLookupFamily)
+	if err != nil {
+		log.WithField("context", "ConfigurationFile").WithField("name", "cluster.dns-lookup-family").Errorf("error parsing parameter dns-lookup-family: %s", err)
```
skriss, line 256: I'm still thinking that returning the error rather than just logging it is more consistent with how we handle parse errors on other config file fields - failures to parse cause `contour serve` to exit. Above all I just prefer consistency.

skriss:
Still need to update `site/docs/main/configuration.md` too.

cmd/contour/serve.go
stevesloka: It now returns an error instead of logging.

stevesloka:
@skriss added configuration.md as well. =)

skriss:
Changes look good; you've got some merge conflicts now.

site/docs/main/api-reference.html
```diff
@@ -1,33 +1,27 @@
 <p>Packages:</p>
 <ul>
 <li>
-<a href="#projectcontour.io%2fv1">projectcontour.io/v1</a>
+<a href="#projectcontour.io%2fv1alpha1">projectcontour.io/v1alpha1</a>
```
stevesloka, line 4: I don't know why this file was changed. Maybe another PR didn't update properly?

skriss: Hmm, maybe an unstable sort within the doc generator? Looks like it switched the order of v1 and v1alpha1

skriss: When I check out your branch and re-run the generate, it undoes this change..

skriss: I think I spotted the issue in the API docs generator, TLDR it's using a map somewhere along the line which potentially throws away the sort

skriss (approved):
LGTM.

It'd be nice to figure out how to resolve the API ref docs sorting issue separately now that we have >1 group, otherwise we're gonna get a lot of noise from this going fwd. We can try getting a patch upstream (looks like there might be an open PR with a fix that we could make some noise on), else we may need to fork the tool.

jpeach:
Are we sure that this makes sense as a global config option? To me, it seems likely that you may need to set this differently for different targets, which argues for an annotation. I could see the utility of setting a default in addition to that though.

youngnick:
I think this is most likely to be an install-wide problem, caused by the shenanigans around AAAA and A records when you've got V4 and V6 around, so that's why I supported the global config first. I think that if we need more configurability, we can add it later, with this as the default.

jpeach:
> I think this is most likely to be an install-wide problem, caused by the shenanigans around AAAA and A records when you've got V4 and V6 around, so that's why I supported the global config first. I think that if we need more configurability, we can add it later, with this as the default.

It's an external service, managed by an entirely separate org. That separate org is the one publishing DNS and both the org and the I remember work paths are factors in determining whether IPv6 connectivity works. It's extremely likely that only one out of many external services would exhibit similar problems. At least spin off an issue for the per-service config.

What was the root cause in the problem that triggered this PR? Was it a local install problem or a external service problem?

youngnick:
You're right,  it was an external service problem.

stevesloka:
The problem I ran into is the local cluster wouldn't route ipv6. The external service returned an ipv6 address first since that's what we configured envoy to do when using the default of "auto". 

That's said, this is a local cluster problem. Nothing to do with the external one. I could see a different issue to add seperate configs for each service type, but not needed today.

youngnick:
So, I've thought about this some more, and I'm still supportive of this config being a global config, for two reasons:
- it fixes the ExternalService issue as it's supposed to, when the ExternalService referenced has a V6 record and the Envoy's can't route there.
- it can also be used to improve DNS lookup performance in clusters that only do V4 - it's pretty common for a large percentage of DNS queries in a V4-only cluster to be spurious AAAA ones, this is a neat knob to help with reducing that a bit.

site/docs/main/api-reference.html
skriss: Maybe just remove this file from the PR since you didn't actually change anything in the API?

PR #2895, opened 2020-09-10
Update linter configuration.

jpeach:
Ensure that `golint` and `gofmt -s` are run as part of the lint
configuration.

### DIFF
--- a/cmd/contour/shutdownmanager.go
+++ b/cmd/contour/shutdownmanager.go
@@ -99,15 +99,15 @@ func (s *shutdownmanagerContext) healthzHandler(w http.ResponseWriter, r *http.R
 // the kubelet calls the shutdown command is different than the HTTP request from Envoy to /shutdown
 func (s *shutdownmanagerContext) shutdownReadyHandler(w http.ResponseWriter, r *http.Request) {
 	for {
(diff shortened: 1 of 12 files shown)


youngnick (approved):
LGTM, good to have those cleared up.

cmd/contour/shutdownmanager.go
```diff
@@ -99,15 +99,15 @@ func (s *shutdownmanagerContext) healthzHandler(w http.ResponseWriter, r *http.R
...
+		if err != nil {
 			http.StatusText(http.StatusOK)
 			if _, err := w.Write([]byte("OK")); err != nil {
 				s.WithField("context", "shutdownReadyHandler").Error(err)
 			}
 			return
-		} else {
-			s.WithField("context", "shutdownReadyHandler").Errorf("error checking for file: %v", err)
 		}
+		s.WithField("context", "shutdownReadyHandler").Errorf("error checking for file: %v", err)
```
skriss, line 110: this looks backwards now - it should respond when the `os.Stat` call does **not** return an error

bgagnon: Fixed in #2899

PR #2898, opened 2020-09-10
internal/dag: Update how headers are processed for special 'Host' header

### LINKED ISSUE #2872: requestHeadersPolicy does not get applied to a service
A `requestHeadersPolicy` should be able to be specified for a service (https://github.com/projectcontour/contour/blob/main/apis/projectcontour/v1/httpproxy.go#L261) but doesn't get applied properly when configured in an HTTPProxy resource. 

If configured in a route, then it does get configured properly.

stevesloka:
The 'Host" header value is special when applied to a
requestHeadersPolicy. When this is set it configures
the hostRewrite field in envoy to match the value of the
requestHeadersPolicy value field.

However, setting this value is only valid when applied to
a route in Envoy, it cannot be applied to a cluster, so
any user setting the "host" header field on a requestHeadersPolicy
on a Spec.Route.Service is invalid.

Fixes #2872

### DIFF
--- a/internal/dag/policy.go
+++ b/internal/dag/policy.go
@@ -64,7 +64,12 @@ func retryPolicy(rp *projcontour.RetryPolicy) *RetryPolicy {
 	}
 }
 
-func headersPolicy(policy *projcontour.HeadersPolicy, allowHostRewrite bool) (*HeadersPolicy, error) {
+func headersPolicyService(policy *projcontour.HeadersPolicy) (*HeadersPolicy, error) {
+	return headersPolicyRoute(policy, false)
+
+}
+
+func headersPolicyRoute(policy *projcontour.HeadersPolicy, allowHostRewrite bool) (*HeadersPolicy, error) {
 	if policy == nil {
 		return nil, nil
 	}
--- a/internal/dag/status_test.go
+++ b/internal/dag/status_test.go
@@ -1709,6 +1709,84 @@ func TestDAGStatus(t *testing.T) {
 		},
 	}
 
+	invalidRequestHeadersPolicyService := &projcontour.HTTPProxy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "invalidRHPService",
+			Namespace: serviceKuard.Namespace,
+		},
+		Spec: projcontour.HTTPProxySpec{
+			VirtualHost: &projcontour.VirtualHost{
+				Fqdn: "example.com",
+			},
+			Routes: []projcontour.Route{{
+				Services: []projcontour.Service{
+					{
+						Name: serviceKuard.Name,
+						Port: 8080,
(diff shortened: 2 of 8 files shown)


internal/dag/policy.go
```diff
@@ -78,7 +83,7 @@ func headersPolicy(policy *projcontour.HeadersPolicy, allowHostRewrite bool) (*H
 		}
 		if key == "Host" {
 			if !allowHostRewrite {
-				return nil, fmt.Errorf("rewriting %q header is not supported", key)
+				return nil, fmt.Errorf("rewriting %q header is only supported on a Spec.Route", key)
```
skriss, line 86: This error msg would be misleading in the case where a user tried to configure a host rewrite on a route's response headers.

I would probably return a more generic error from this function (something closer to what it was before), then augment the error message at the call sites to add more context about why it wasn't allowed.

skriss (changes requested):

jpeach (commented):
C an we update the API reference to document this?

internal/dag/policy.go
stevesloka: Last push rewrites the error messages so they are more specific.

stevesloka:
> Can we update the API reference to document this?

Updated. =)

youngnick (approved):
LGTM

internal/dag/status_test.go
```diff
@@ -2366,6 +2444,50 @@ func TestDAGStatus(t *testing.T) {
...
+			objs: []interface{}{invalidResponseHeadersPolicyService, serviceKuard},
+			want: map[types.NamespacedName]Status{
+				{Name: invalidResponseHeadersPolicyService.Name, Namespace: invalidResponseHeadersPolicyService.Namespace}: {
+					Object:      invalidResponseHeadersPolicyService,
+					Status:      "invalid",
+					Description: "rewriting \"Host\" header is not supported on response headers",
+					Vhost:       invalidResponseHeadersPolicyService.Spec.VirtualHost.Fqdn,
+				},
+			},
+		},
```
skriss, line 2479: these 2 look identical - I'd drop the second since the name for the first is better.

internal/dag/status_test.go
```diff
@@ -2366,6 +2444,50 @@ func TestDAGStatus(t *testing.T) {
...
+		"responseHeadersPolicy invalid on Route": {
```
skriss, line 2480: I'd add "Host header invalid" to the name of the test

skriss (changes requested):
couple minor tweaks to the tests, otherwise LGTM

internal/dag/status_test.go
stevesloka: I'm such a bad copy/paster at 11pm at night. =(

skriss (approved):
LGTM

PR #2899, opened 2020-09-10
cmd/shutdownmanager: /shutdown handler fixes

### LINKED ISSUE #2896: Shutdown manager "error checking for file" log messages shouldn't be at ERROR level
After upgrading to Contour 0.8.1, I was surprised to find these entries in the `shutdown-manager` container logs:

```
envoy-m2trt shutdown-manager time="2020-09-10T02:11:58Z" level=error msg="error checking for file: stat /ok: no such file or directory" context=shutdownReadyHandler
envoy-m2trt shutdown-manager time="2020-09-10T02:11:59Z" level=error msg="error checking for file: stat /ok: no such file or directory" context=shutdownReadyHandler
envoy-m2trt shutdown-manager time="2020-09-10T02:12:00Z" level=error msg="error checking for file: stat /ok: no such file or directory" context=shutdownReadyHandler
envoy-m2trt shutdown-manager time="2020-09-10T02:12:01Z" level=error msg="error checking for file: stat /ok: no such file or directory" context=shutdownReadyHandler
```

My initial thought was that something was not configured correctly. But further investigation leads me to believe this is the normal behavior for the shutdown manager.

It polls a file continuously until it exists. Since it expects **not** to find it, I would argue errors from `os.Stat()` that relate to the file not existing are not to be logged at `level=error`, but rather `level=info`.

A minor detail for sure, but since we cannot see the pre stop hook logs from the Envoy side, it's best if normal things are reported as normal events from the `shutdown-manager` side.

Also, a final logging message with a confirmation that the file now exists would be nice.

### LINKED ISSUE #2897: ShutdownManager's /shutdown loop does not exit when client cancels request
The `/shutdown` handler in the `shutdown-manager` container has an infinite loop. If the HTTP client aborts the request and closes the connection, the goroutine performing that infinite loop will continue indefinitely (unless the the `/ok` file shows up).

I think the handler should use the `CloseNotifier` channel to exit if the client gives up:
https://golang.org/pkg/net/http/#CloseNotifier

This can be combined with the 1 second sleep.

This issue was discovered while investigating #2896 

**Environment:**
- Contour version: v1.8.1
- Kubernetes version: 1.17.1

bgagnon:
Fixes #2896 , fixes #2897 

- shutdownReady filepath ("/ok") is configurable
- os.Stat() interval is configurable
- detect client disconnection and stop check loop
- don't use error logging level when os.Stat() tells us the file does not yet exist
- add test coverage for success and client disconnect cases

### DIFF
--- a/cmd/contour/shutdownmanager.go
+++ b/cmd/contour/shutdownmanager.go
@@ -111,16 +111,16 @@ func (s *shutdownmanagerContext) shutdownReadyHandler(w http.ResponseWriter, r *
 	ctx := r.Context()
 	for {
 		_, err := os.Stat(s.shutdownReadyFile)
-		if err == nil {
+		if os.IsNotExist(err) {
+			l.Infof("file %s does not exist; checking again in %v", s.shutdownReadyFile,
+				s.shutdownReadyCheckInterval)
+		} else if err == nil {
 			l.Infof("detected file %s; sending HTTP response", s.shutdownReadyFile)
 			http.StatusText(http.StatusOK)
 			if _, err := w.Write([]byte("OK")); err != nil {
 				l.Error(err)
 			}
 			return
-		} else if os.IsNotExist(err) {
-			l.Infof("file %s does not yet exist; checking again in %v", s.shutdownReadyFile,
-				s.shutdownReadyCheckInterval)
 		} else {
 			l.Errorf("error checking for file: %v", err)
 		}
--- a/cmd/contour/shutdownmanager.go
+++ b/cmd/contour/shutdownmanager.go
@@ -36,8 +36,10 @@ const (
 	prometheusStat     = "envoy_http_downstream_cx_active"
 )
 
-// File path used in the /shutdown endpoint.
+// shutdownReadyFile is the default file path used in the /shutdown endpoint.
 const shutdownReadyFile = "/ok"
+
+// shutdownReadyFile is the default polling interval for the file used in the /shutdown endpoint.
 const shutdownReadyCheckInterval = time.Second * 1
 
 func prometheusLabels() []string {
@@ -46,8 +48,10 @@ func prometheusLabels() []string {
 
 type shutdownmanagerContext struct {
 	// httpServePort defines what port the shutdown-manager listens on
-	httpServePort              int
-	shutdownReadyFile          string
+	httpServePort int
+	// shutdownReadyFile is the default file path used in the /shutdown endpoint
+	shutdownReadyFile string
+	// shutdownReadyCheckInterval is the polling interval for the file used in the /shutdown endpoint
 	shutdownReadyCheckInterval time.Duration
 
 	logrus.FieldLogger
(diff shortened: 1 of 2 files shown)


cmd/contour/shutdownmanager.go
```diff
@@ -98,17 +101,44 @@ func (s *shutdownmanagerContext) healthzHandler(w http.ResponseWriter, r *http.R
...
+		signalFilepath = s.shutdownReadyFile
+	} else {
+		signalFilepath = shutdownReadyFile
+	}
+	var signalFileCheckDuration time.Duration
+	if s.shutdownReadyCheckInterval != 0 {
+		signalFileCheckDuration = s.shutdownReadyCheckInterval
+	} else {
+		signalFileCheckDuration = shutdownReadyCheckInterval
+	}
```
bgagnon, line 117: Not too happy with the location of this logic... maybe the defaults assignment should move to `newShutdownManagerContext()`.

cmd/contour/shutdownmanager.go
```diff
@@ -38,14 +38,17 @@ const (
 
 // File path used in the /shutdown endpoint.
 const shutdownReadyFile = "/ok"
+const shutdownReadyCheckInterval = time.Second * 1
```
skriss, line 41: nit: since we have a comment on the previous constant, it'd be nice to add one for this new const as well.

cmd/contour/shutdownmanager.go
skriss: Yeah, I'd probably assign the default values to these fields in `newShutdownManagerContext()`, then you can just override them as needed.

cmd/contour/shutdownmanager.go
```diff
@@ -98,17 +101,44 @@ func (s *shutdownmanagerContext) healthzHandler(w http.ResponseWriter, r *http.R
...
+		signalFileCheckDuration = shutdownReadyCheckInterval
+	}
+
+	l := s.WithField("context", "shutdownReadyHandler")
+	ctx := r.Context()
 	for {
-		_, err := os.Stat(shutdownReadyFile)
-		if err != nil {
+		_, err := os.Stat(signalFilepath)
+		if err == nil {
```
skriss, line 123: Not a huge deal, but I might reorder this to handle the error cases first, which is more common in go (I'd guess that not writing it that way in the first place contributed to the regression here) - so something like:

```go
if err != nil {
    if os.IsNotExist(err) {
        ...
    } else {
       ...
    }

    // wait & continue
}

// no error: ready for shutdown, write response
```

skriss (changes requested):
thanks for the PR @bgagnon! I had a couple comments.

cmd/contour/shutdownmanager.go
bgagnon: Done!

cmd/contour/shutdownmanager.go
bgagnon: Done.

cmd/contour/shutdownmanager.go
bgagnon: Not exactly what you suggested, but I reordered the if cases.

youngnick (approved):
LGTM, great change, thanks @bgagnon

skriss (approved):
LGTM, thanks @bgagnon

PR #2900, opened 2020-09-10
[WIP] replace DAG builder caches with visitors

skriss:
Updates #2881 
Updates #2226 

Summary of changes:
- get rid of DAG builder caches; reads are replaced by DAG visitors, and writes are replaced by writing directly to the DAG's roots
- StatusWriter is moved to the DAG object (@youngnick will presumably replace this with the conditions work)
- Processors no longer directly depend on a DAG builder

I probably still have some comments to update and unit tests to write, but wanted to solicit feedback on this revised approach first.

After this, the main thing preventing the processors from being moved to their own package is the lack of an exported interface for the `KubernetesCache`.


youngnick:
I like this approach, it looks good. I'll have pretty substantial changes to the status stuff in my PR, so this will definitely need to go in before my stuff. I'll concentrate on getting #2886 landed.

skriss:
> I like this approach, it looks good. I'll have pretty substantial changes to the status stuff in my PR, so this will definitely need to go in before my stuff. I'll concentrate on getting #2886 landed.

:+1: I'll get this updated locally on top of #2886 today so it's ready to merge assuming we're happy with it.

skriss:
rebased, ready for a full review from @youngnick and @stevesloka

skriss:
I marked this as closing #2226 as I think it accomplishes what was described in that issue. Any further refactoring can be tracked in new issues.

internal/dag/accessors.go
```diff
@@ -0,0 +1,241 @@
...
+
+// RouteServiceName identifies a service used in a route.
+type RouteServiceName struct {
+	Name      string
+	Namespace string
+	Port      int32
+}
+
+// GetServices returns all services in the DAG.
+func GetServices(dag *DAG) map[RouteServiceName]*Service {
```
stevesloka, line 35: Could this be a method off of the `dag` type so that you wouldn't need to pass in the dag each time?

stevesloka (approved):
/lgtm

Just a couple thoughts around what should be member methods or not.

internal/dag/accessors.go
skriss: Definitely could, yes. I guess the one benefit of keeping them as standalone functions is that they could move into a different package along with the processors. But, in the current state, they could definitely be methods. Happy to change them if you think that makes sense.

youngnick (approved):
LGTM. Feels like maybe it's better to keep things that access a DAG as methods _on_ the DAG to me, but as @skriss said, that would preclude moving them out with the processors.

skriss:
I'll make them methods; it's easy to change down the road if we want to.

stevesloka:
Nice work @skriss! =)

Issue #2728, opened 2020-07-24
improve validation around user-specified timeout values


skriss:
Currently, if an unparseable timeout value is specified in either the config file or in a `TimeoutPolicy` field in an `HTTPProxy` route, the parse error is swallowed and the timeout is disabled/set to infinity.

There are a few things we can do to improve this:
- for the HTTPProxy fields, add kubebuilder validations (e.g. `+kubebuilder:validation:Pattern`)
- for HTTPProxy, if there's an unparseable timeout value, set a validation error or warning in the proxy's status/conditions.
- for the config file, if there's an unparseable timeout value, return the error so either `contour serve` can exit, or we can log a warning/error.

Also, consider what to do with unparseable timeout annotations on `Ingress` resources.

ref. #2697

stevesloka:
// cc auth impl: https://github.com/projectcontour/contour/pull/2886#discussion_r486606226

PR #2905, opened 2020-09-11
internal/timeout: return parse errors instead of defaulting to infinity

skriss:
Returns errors when parsing timeout strings so they can be reported to
the user, either as invalid HTTPProxies, or in Contour logs. Previously,
such errors were swallowed and the timeout in question was disabled/set
to infinity.

Updates #2728

### DIFF
--- a/internal/dag/httpproxy_processor.go
+++ b/internal/dag/httpproxy_processor.go
@@ -224,8 +224,14 @@ func (p *HTTPProxyProcessor) computeHTTPProxy(proxy *projcontour.HTTPProxy) {
 				}
 
 				svhost.AuthorizationService = ext
-				svhost.AuthorizationResponseTimeout = timeout.Parse(auth.ResponseTimeout)
 				svhost.AuthorizationFailOpen = auth.FailOpen
+
+				timeout, err := timeout.Parse(auth.ResponseTimeout)
+				if err != nil {
+					sw.SetInvalid("Spec.Virtualhost.Authorization.ResponseTimeout string %q cannot be parsed: %s", auth.ResponseTimeout, err)
+					return
+				}
+				svhost.AuthorizationResponseTimeout = timeout
 			}
 		}
 	}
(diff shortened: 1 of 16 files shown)


jpeach:
Quick comments:
- bumping the Go version should be in a separate PR
- we should enforce the timeout syntax with a validation regex in the CRD
- once we enforce validation, do we still need the parse errors?

skriss:
> Quick comments:
> 
> * bumping the Go version should be in a separate PR

See #2906 :) 
> * we should enforce the timeout syntax with a validation regex in the CRD

Yep, just haven't gotten to that part yet
> * once we enforce validation, do we still need the parse errors?

It's still an issue for (a) `contour serve` config file, and (b) Ingress annotations, so I think yes.

internal/featuretests/timeoutpolicy_test.go
```diff
@@ -170,17 +170,10 @@ func TestTimeoutPolicyRequestTimeout(t *testing.T) {
 	}
 	rh.OnAdd(p1)
 
-	// check timeout policy with malformed response timeout is propogated as infinity
+	// check timeout policy with malformed response timeout is not propogated
```
youngnick, line 173: Nit:
```suggestion
	// check timeout policy with malformed response timeout is not propagated
```

internal/dag/policy.go
```diff
@@ -123,25 +124,32 @@ func escapeHeaderValue(value string) string {
 }
 
 // ingressRetryPolicy builds a RetryPolicy from ingress annotations.
-func ingressRetryPolicy(ingress *v1beta1.Ingress) *RetryPolicy {
+func ingressRetryPolicy(ingress *v1beta1.Ingress, log logrus.FieldLogger) *RetryPolicy {
```
youngnick, line 127: Feels like this is wiring the logger way down into the code, but I don't see any other way to do it.

internal/featuretests/timeoutpolicy_test.go
```diff
@@ -281,17 +274,10 @@ func TestTimeoutPolicyIdleTimeout(t *testing.T) {
...
```
youngnick, line 277: Nit:
```suggestion
	// check timeout policy with malformed response timeout is not propagated
```

youngnick (commented):
I like the approach, looks good.

internal/dag/policy.go
skriss: Yeah, I didn't love doing this either. The alternative I considered was having this function return a partial policy along with an error and letting the caller log the error, but this pattern seemed more obvious to me.

skriss: I guess we could make these methods on the ingress processor, and then they could access the logger as a field of the processor. 🤷‍♂️

skriss:
rebased for auth work

stevesloka (approved):
/lgtm 🎉

skriss:
I'll have to rebase this post #2900 as well

internal/dag/httpproxy_processor.go
```diff
@@ -225,8 +225,14 @@
 				}
 
 				svhost.AuthorizationService = ext
-				svhost.AuthorizationResponseTimeout = timeout.Parse(auth.ResponseTimeout)
 				svhost.AuthorizationFailOpen = auth.FailOpen
+
+				timeout, err := timeout.Parse(auth.ResponseTimeout)
```
youngnick, line 230: Nit: Maybe a test for this one? No problems if not, because I'm going to have to refactor all the status tests anyway.

youngnick (approved):
LGTM, with one small testing nit.

internal/dag/httpproxy_processor.go
skriss: ah yeah, I think I breezed over that during a rebase. Added one.

PR #2909, opened 2020-09-14
examples/contour: change certgen job to use main image tag

### LINKED ISSUE #2834: release examples: Make certgen job use tagged image
The cert-gen job has always used an image tag of `:latest` to get around the issue of applying the same manifest on top of an existing installation, however, tagged releases use a unique name for each certgen job for that release so there's no need to use latest for releases and can cause issues for users in air-gapped environments.  

We should remove the `:latest` tag and use a specific version. 

ref: https://github.com/projectcontour/contour/blob/v1.7.0/examples/render/contour.yaml#L1466

skriss:
Changes the certgen job to use the "main" image tag instead of "latest".
On release, Contour's release tooling will replace this with a version-
specific tag. Since the certgen job also now gets a unique name for every
release, this change will not cause issues related to job specs being
immutable for users who are upgrading and reapplying the YAML.

Fixes #2834.


youngnick (approved):
LGTM

stevesloka (approved):
/lgtm

Issue #2885, opened 2020-09-04
Add support for forwarding of client certificate material to pod (xfcc in Envoy)


moderation:
https://github.com/projectcontour/contour/issues/1090 references this functionality but the issue was closed with merged support for client cert validation at Contour only.

NGINX implements this functionality with the annotation `nginx.ingress.kubernetes.io/auth-tls-pass-certificate-to-upstream`

We have teams using this annotation with NGINX to perform mTLS with the certificate material upstream at the pod in a k8s cluster.

The xfcc functionality in Envoy is documented at https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-forwarded-client-cert

Following a community call I chatted with @tsaarni who went as far as providing a rough draft of the Contour code required and tested it:

```diff
--git a/internal/envoy/listener.go b/internal/envoy/listener.go
index 4a2f8111..c987536b 100644
--- a/internal/envoy/listener.go
+++ b/internal/envoy/listener.go
@@ -275,6 +275,11 @@ func (b *httpConnectionManagerBuilder) Get() *envoy_api_v2_listener.Filter {
                RequestTimeout:    envoyTimeout(b.requestTimeout),
                StreamIdleTimeout: envoyTimeout(b.streamIdleTimeout),
                DrainTimeout:      envoyTimeout(b.connectionShutdownGracePeriod),
+
+               ForwardClientCertDetails: http.HttpConnectionManager_SANITIZE_SET,
+               SetCurrentClientCertDetails: &http.HttpConnectionManager_SetCurrentClientCertDetails{
+                       Cert: true,
+               },
        }        // Max connection duration is infinite/disabled by default in Envoy, so if the timeout setting
```
However it probably doesn't make sense to send the cert material all the time as it can get quite large and there are bandwidth and message size implications. Ideally this would be exposed as either a HTTPProxy config option or enabled with a Service annotation.

youngnick:
Just to add to @jpeach's labels, I think this is a reasonable request, that will just require some product thought about where we put it in terms of config.

jpeach:
For anyone who would like to tackle this, here's the API I would recommend:
```
--- apis/projectcontour/v1/httpproxy.go
+++ apis/projectcontour/v1/httpproxy.go
@@ -242,6 +242,8 @@ type TLS struct {
        // EnableFallbackCertificate defines if the vhost should allow a default certificate to
        // be applied which handles all requests which don't match the SNI defined in this vhost.
[...]

PR #2910, opened 2020-09-15
Implement client authentication for backend TLS

### LINKED ISSUE #2338: Securing connection between Envoy and upstream service: client authentication

tsaarni:
* Added new parameter to configuration file to configure a name of a secret
  containing client certificate and private key.

* Configure Envoy to authenticate itself towards upstream cluster with
  the client certificate when upstream connection uses TLS.

Closes #2338 


tsaarni:
I've documented a manual test procedure here https://gist.github.com/tsaarni/245584e8b4e40b9aa5a9924c7d13eadf to validate following on a target environment:

* successful connection with valid client certificate
* rotate client certificate
* unsuccessful connection with invalid client certificate
* unsuccessful connection without client certificate

To run the procedure fully requires Kind on linux (I think...  the wireshark bits to be able to see TLS handshake)

cmd/contour/servecontext.go
```diff
@@ -205,37 +205,49 @@ type TLSConfig struct {
...
+
+	// ClientCertificate defines the namespace/name of Kubernetes secret containing client
+	// certificate andprivate key to be used when establishing TLS connection to upstream cluster.
+	ClientCertificate NamespacedName `yaml:"envoy-client-certificate,omitempty"`
 }
 
-// FallbackCertificate defines the namespace/name of the Kubernetes secret to
-// use as fallback when a non-SNI request is received.
-type FallbackCertificate struct {
+// NamespacedName defines the namespace/name of the Kubernetes resource referred
```
youngnick, line 215: I wasn't sure at first why you weren't using `types.NamespacedName` here, but now I see it's used for the yaml value and for validation. Maybe something in this comment explaining that?

youngnick (approved):
LGTM with one small suggestion. I'll add @jpeach as well though for his review.

cmd/contour/servecontext.go
tsaarni: I was thinking the same when renaming/reusing `FallbackCertificate` type for the new certificate:  Wouldn't Kubernetes `types.NamespacedName` have worked as well? The Kubernetes type does not have YAML tags, but isn't the default "untagged" behavior the same as the tags here? Or would there be some differences?

I was then reasoning that `FallbackCertificate` was defined likely because we did not want to depend on external type for our configuration file. In theory `types.NamespacedName` could change at any time, e.g. new field could be introduced which would automatically change parsing of our configuration file as well.  

@stevesloka: I wonder if this was the motivation for defining `type FallbackCertificate` and not using K8s `types.NamespacedName`? If yes, I'll write that down in the comment.

internal/contour/secret.go
```diff
@@ -107,6 +107,10 @@ func (v *secretVisitor) visit(vertex dag.Vertex) {
 		if svh.FallbackCertificate != nil {
 			v.addSecret(svh.FallbackCertificate)
 		}
+	case *dag.Cluster:
+		if svh.ClientCertificate != nil {
```
skriss, line 111: nit: let's rename `svh` now that it's not only for a `SecureVirtualHost` -- `obj` or something similarly generic.

cmd/contour/serve.go
```diff
@@ -247,11 +258,13 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 					FieldLogger: log.WithField("context", "IngressProcessor"),
 				},
```
skriss, line 259: should we be supporting this for `Ingress` too?

skriss (commented):
this looks great to me @tsaarni, just a couple things.

skriss:
Unfortunately you'll need a rebase as well, there are some merge conflicts.

cmd/contour/servecontext.go
tsaarni: I added a comment about the type being used for YAML parsing with the above assumption, hope the assumption was correct!

internal/contour/secret.go
tsaarni: Fixed!

cmd/contour/serve.go
tsaarni: I missed mentioning this in the design doc, but I've understood that there is no intention to support backend TLS for `Ingress`.  So I believe upstream TLS code is missing completely there.  This is mentioned in older design doc [design/tls-backend-verification.md](https://github.com/projectcontour/contour/blob/main/design/tls-backend-verification.md).

tsaarni:
Thanks for reviews so far! 

I've rebased the PR.

Note: I've also added draft of integration test which was not previously included in the PR, to automate the manual test procedure.  The tests will be skipped by default, and I can gladly also remove them for now if you wish!  The test depends on echoserver with my own changes to ingress-controller-conformance echoserver which add TLS support. I sent a PR but it might not work out at the end, so it may be that another approach should be taken at the end.

jpeach:
xref https://github.com/kubernetes-sigs/ingress-controller-conformance/pull/71

jpeach (changes requested):
I haven't finished reviewing yet, but here's the first set of feedback.

cmd/contour/serve.go
tsaarni: Yes sorry, it is the server authentication that is missing: TLS as itself is supported using annotation `projectcontour.io/upstream-protocol.tls` on service like you say, but server authentication can be only expressed with `HTTPProxy` `spec.routes.services[].validation`. 

We can still add client authentication regardless of not authenticating the server, though it is unusual to have it this way.

tsaarni (commented):
I would like to double check and confirm with you that we agree following is wanted behavior: 

When Envoy client certificate is configured, but the Secret does not exist, the clusters will not be created at all.  I believe it mimics how FallbackCertificate was implemented.   Is this your expectation here too? This makes the configuration error very visible, and error can be seen in all `HTTPProxy` statuses but it will be quite easy to break everything by removing just a single Secret.  

Alternative could be not configure client certificate to the upstream clusters while Secret is missing (or when it is broken).

cmd/contour/serve.go
tsaarni: I've now added support for client auth for `Ingress` too.

internal/dag/httpproxy_processor.go
```diff
@@ -495,6 +499,15 @@ func (p *HTTPProxyProcessor) computeRoutes(
...
 
+			var clientCertSecret *Secret
+			if p.ClientCertificate != nil {
+				clientCertSecret, err = p.source.LookupSecret(*p.ClientCertificate, validSecret)
+				if err != nil {
+					sw.SetInvalid("tls.envoy-client-certificate Secret %q is invalid: %s", p.ClientCertificate, err)
+					return nil
+				}
+			}
+
```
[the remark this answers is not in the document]
tsaarni: I've added test cases for `IngressProcessor` and `ExtensionServiceProcessor` and also added tests where Secret is removed.

site/docs/main/httpproxy.md
```diff
@@ -349,6 +349,8 @@ spec:
 
 ```
 
+Contour can be configured with a `namespace/name` in the [Contour configuration file][11] of a Kubernetes secret which Envoy will present as client certificate when establishing TLS connection to the backend.
+
```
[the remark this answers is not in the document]
tsaarni: I've added new section and explained these.

cmd/contour/servecontext_test.go
```diff
@@ -290,6 +290,72 @@ func TestFallbackCertificateParams(t *testing.T) {
...
+		},
+	}
+
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			got, err := tc.ctx.envoyClientCertificate()
+
+			if diff := cmp.Diff(tc.want, got); diff != "" {
+				t.Fatal(diff)
+			}
```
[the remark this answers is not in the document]
tsaarni: Fixed.

internal/dag/extension_processor.go
```diff
@@ -27,13 +27,21 @@ import (
 
 type ExtensionServiceProcessor struct {
 	logrus.FieldLogger
+
+	source *KubernetesCache
```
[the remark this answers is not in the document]
tsaarni: It is not needed, my bad!

internal/dag/extension_processor.go
```diff
@@ -27,13 +27,21 @@ import (
...
 var _ Processor = &ExtensionServiceProcessor{}
 
-func (p *ExtensionServiceProcessor) Run(dag *DAG, cache *KubernetesCache) {
-	for _, e := range cache.extensions {
-		if ext := p.buildExtensionService(cache, e); ext != nil {
+func (p *ExtensionServiceProcessor) Run(dag *DAG, source *KubernetesCache) {
+	p.source = source
+
+	for _, e := range source.extensions {
+		if ext := p.buildExtensionService(source, e); ext != nil {
```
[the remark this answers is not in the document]
tsaarni: Fixed.

site/docs/main/configuration.md
```diff
@@ -54,6 +55,15 @@ Contour should provision TLS hosts.
 {: class="table thead-dark table-bordered"}
 <br>
 
+### Client Certificate
```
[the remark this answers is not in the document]
tsaarni: Fixed.

site/docs/main/configuration.md
```diff
@@ -54,6 +55,15 @@ Contour should provision TLS hosts.
...
+
+| Field Name | Type| Default  | Description |
+|------------|-----|----------|-------------|
+| name       | string | `""` | This field specifies the name of the Kubernetes secret to use as the client certificate and private key when establishing TLS connection to the backend service. |
+| namespace  | string | `""` | This field specifies the namespace of the Kubernetes secret to use as the client certificate and private key when establishing TLS connection to the backend service. |
```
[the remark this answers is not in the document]
tsaarni: Fixed.

youngnick:
I feel like it's okay for this to behave the same as `fallbackCertificate`, personally. @jpeach @skriss @stevesloka ?

jpeach:
> I feel like it's okay for this to behave the same as `fallbackCertificate`, personally. @jpeach @skriss @stevesloka ?

Yeh, I think that this is a problem to handle in Contour in a more general way ...

_integration/testsuite/httpproxy/013-backend-tls.yaml
```diff
@@ -0,0 +1,262 @@
...
+# delete envoy certificate to force cert-manager to rotate it
+apiVersion: v1
+kind: Secret
+metadata:
+  name: envoy-client-cert
+  namespace: 013-backend-tls
+$apply: delete
+
+---
+
```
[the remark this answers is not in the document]
jpeach: Each fragment gets its own timeout, so the test ends up being more deterministic if you test what you expect in each stage in a separate fragment. Even with this, I think it's possible for the check to spuriously succeed since deleted objects can be visible through the API for a short time.

jpeach (approved):
Nice work @tsaarni !

jpeach:
@skriss LMK if you want to make another review pass. If you are OK, I'll merge this tomorrow.

tsaarni:
> Nice work @tsaarni !

Thanks, and thanks for great review comments!

BTW: I still did a small change & force push, which only contains typo fix for the doc and integration test improvement for certificate rotation.

skriss (approved):
LGTM!

jpeach:
@tsaarni can you please rebase?

tsaarni:
PR rebased!

moderation:
@tsaarni does the PR close https://github.com/projectcontour/contour/issues/2885? I don't _think_ so but wanted to check.

tsaarni:
> @tsaarni does the PR close #2885? I don't _think_ so but wanted to check.

No, this one was unrelated to #2885.

PR #2912, opened 2020-09-16
docs: Add Contour deprecation policy

### LINKED ISSUE #2749: Deprecate and then drop support for contour.heptio.com/ annotation prefix
We should eventually drop support for annotations prefixed with `contour.heptio.com/` since they've been replaced with `projectcontour.io/` for quite awhile. This could be a 2.0 thing, or we could put out a deprecation notice and then drop support a few releases later.

### LINKED ISSUE #2836: Add deprecation policy statement for `projectcontour.io` CRDs
Among the maintainers, we've always agreed that the `projectcontour.io` CRDs would abide by the Kubernetes deprecation policy's principles, but we've never written that down. 

This issue is to create a policy document so that we have it in writing, officially.

youngnick:
Fixes #2836 

### DIFF
--- a/site/_resources/deprecation-policy.md
+++ b/site/_resources/deprecation-policy.md
@@ -1,0 +1,63 @@
+---
+title: Contour Deprecation Policy
+layout: page
+---
+
+Contour publishes a few APIs, the most notable being the `projectcontour.io` api group of Kubernetes objects and their associated helper code, the command line for `contour`, and the Contour configuration file.
+Each of these have deprecation policies, with all of them being based on the [Kubernetes API deprecation policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/).
+
+
+
+## `projectcontour.io` API groups (aka Contour CRDs)
+
+Our Kubernetes CRDs use the Kubernetes API deprecation conventions, including a similar deprecation timeline.
+
+- There are three levels of stability, Alpha, Beta and GA (in increasing order).
+- We may skip the Beta level when we are reasonably confident the schema is good.
+
+We use the same rules as Kubernetes:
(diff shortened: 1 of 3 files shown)


site/_resources/deprecation-policy.md
```diff
@@ -0,0 +1,50 @@
+---
+title: Contour Deprecation Policy
+layout: page
+---
+
+Contour published a few APIs, the most notable being the `projectcontour.io` api group of Kubernetes objects, the command line for `contour`, and the Contour configuration file.
```
stevesloka, line 6: nit: s/published/publishes

stevesloka (changes requested):
Also, how does code in here (https://github.com/projectcontour/contour/blob/main/apis/projectcontour/v1/helpers.go), for example, get deprecated / changed? It's not really part of an API version, but since public, changes can be breaking for users.

site/_resources/deprecation-policy.md
youngnick: Fixed.

youngnick:
@stevesloka thanks for that. I've added some statements around the helper code.

stevesloka (approved):
The helpers in the API sections still feel like we're going to need to version stuff all the time when those change and they should't be there (at least the one's with business logic), but otherwise lgtm.

skriss:
> The helpers in the API sections still feel like we're going to need to version stuff all the time when those change and they should't be there (at least the one's with business logic), but otherwise lgtm.

Yeah, I hadn't really considered that when those helpers were added, but now that you bring it up, it'll certainly make me less inclined to add things in there so I don't have to maintain compatibility :)

skriss (approved):

PR #2913, opened 2020-09-16
apis: add kubebuilder regex validation for timeout strings

### LINKED ISSUE #2728: improve validation around user-specified timeout values

skriss:
Adds kubebuilder regex validations for all timeout string API
fields.

Closes #2728 

### DIFF
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -153,6 +153,7 @@ type AuthorizationServer struct {
 	// The string "infinity" is also a valid input and specifies no timeout.
 	//
 	// +optional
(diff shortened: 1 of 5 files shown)


stevesloka (approved):
/lgtm

This would be great to have functional tests for this to cover as well.

skriss:
> This would be great to have functional tests for this to cover as well.

Yep, good idea. Still need to wrap my brain around how to add new test cases there as I haven't done it yet.

youngnick (approved):
LGTM, nice work.

skriss:
@stevesloka I'll file a followup issue to look at adding an integration test for this.

skriss:
ignoring SNI integration test flake.

PR #2916, opened 2020-09-16
examples: generate v1 CustomResourceDefinitions

### LINKED ISSUE #1723: Make "kubectl explain" work for HTTPProxy resources.

### LINKED ISSUE #1978: HTTPProxy CRD should be structural and have pruning enabled
As noted in https://github.com/projectcontour/contour/issues/1723#issuecomment-544307006, our CRDs are not structural, and can't have pruning enabled at the moment.

My latest investigation into this suggests that pruning requires (for CustomResourceDefinitions v1beta1):
- structural definitions (this should be taken care of by the tooling added in #1723)
- At least Kubernetes v1.15

This is complicated by the fact that we'll need to move to CustomResourceDefinitions v1 at some point, which requires *only* structural schemas, and enforces pruning by default. v1beta1 is now deprecated, as of v1.16. However, changing this will mean that Contour will require v1.16 at a minimum.

So, this issue is to record these details for now, until we are ready to mandate at least either v1.15 or v1.16 of Kubernetes.

### LINKED ISSUE #2527: Incorrectly nesting routes under virtualhost is considered valid
If _routes_ is incorrectly nested under _virtualhost_ the HTTPProxy object is still considered valid, but all calls just return 404 Not Found:
```
$ cat bad-explorer-httpproxy.yaml
kind: HTTPProxy
metadata:
  annotations:
    heptio.contour.com/ingress.class: contour
  name: explorer
  namespace: burdik
spec:
  virtualhost:
    fqdn: explorer.my.domain
    routes:
    - services:
      - name: explorer
        port: 8080
$ kubectl create -f bad-explorer-httpproxy.yaml
httpproxy.projectcontour.io/explorer created
$ kubectl get httpproxy
NAME                    FQDN                                  TLS SECRET     STATUS   STATUS DESCRIPTION
explorer                explorer.my.domain                                   valid    valid HTTPProxy
$ curl -I explorer.my.domain
HTTP/1.1 404 Not Found
date: Fri, 15 May 2020 16:55:28 GMT
server: envoy
transfer-encoding: chunked

```

I expected that the HTTPProxy status would be invalid because _routes_ is not allowed under _virtualhost_.

With the indentation fixed the call works fine:
```
$ cat explorer-httpproxy.yaml
kind: HTTPProxy
metadata:
  annotations:
    heptio.contour.com/ingress.class: contour
  name: explorer
  namespace: burdik
spec:
  virtualhost:
    fqdn: explorer.my.domain
  routes:
  - services:
    - name: explorer
      port: 8080
$ kubectl replace -f explorer-httpproxy.yaml
httpproxy.projectcontour.io/explorer replaced
$ curl -I explorer.my.domain
HTTP/1.1 200 OK
date: Fri, 15 May 2020 16:56:46 GMT
content-length: 446
content-type: text/html; charset=utf-8
x-envoy-upstream-service-time: 0
server: envoy

```
**Environment:**

- Contour version: 1.4.0
- Kubernetes version: v1.16.9
- Kubernetes installer & version: kubeadm v1.16.9
- Cloud provider or hardware configuration: on-prem
- OS (e.g. from `/etc/os-release`): RHEL 8.1

### LINKED ISSUE #2678: Generate Contour CRDs with V1 CustomResourceDefinition.
Generating CRDs as  v1 CustomResourceDefinition resources enables better CRD validation and defaulting, as well as the ability to use `kubectl explain`.

This requires Kubernetes 1.16 or later. We can do this as the default YAML, but we can also generate different sets of YAML (as demonstrated in #2474).

This issue tracks the decision for adding this, and we can also use it to collect other issues that are blocked on the CRD support.

### LINKED ISSUE #2903: CRD "Strict mode" to validate schema on create and edit

During the creation of new HTTPProxy resources, values, such as timeoutPolicy, can be incorrectly indented and the creation will be accepted without any errors or warnings.  This can lead to long periods of troubleshooting believing the timeoutPolicy has been set correctly rather than being ignored.  Could the CRD be edited to include a "strict mode" to [validate the structure](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema) allowing for warnings to be presented for incorrect structure or prevent the object being created at all?  This would greatly improve confidence in production deployments where it is important to validate the configuration is as expected.

skriss:
Updates the Contour CustomResourceDefinition YAML files
to contain v1 resources instead of v1beta1.

Closes #2678
Closes #1723
Closes #1978
Closes #2903
Closes #2527

**Release note**: 
- we need to be very clear that Kubernetes 1.16 is now the hard minimum supported version since the sample YAML contains v1 CRDs.

- For upgrade instructions, for folks who don't want to reapply the full YAML, they can either reapply just the CRD YAML, or they can run the following:

```
kubectl patch crd httpproxies.projectcontour.io --type json -p '[{"op": "replace", "path": "/spec/preserveUnknownFields", value: false}]'
kubectl patch crd tlscertificatedelegations.projectcontour.io --type json -p '[{"op": "replace", "path": "/spec/preserveUnknownFields", value: false}]'
kubectl patch crd extensionservices.projectcontour.io --type json -p '[{"op": "replace", "path": "/spec/preserveUnknownFields", value: false}]'
```

- Upgrading users will also need to update RBAC (the ClusterRole) since we added `list` permissions on CRDs

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Its goal is to expand upon the functionality of the Ingress API to allow for a r
 
 ## Prerequisites
 
-Contour is tested with Kubernetes clusters running version 1.10 and later, but should work with earlier versions where Custom Resource Definitions are supported (Kubernetes 1.7+).
+Contour requires Kubernetes version 1.16 or later, for [v1 Custom Resource Definition support](https://kubernetes.io/blog/2019/09/18/kubernetes-1-16-release-announcement/#custom-resources-reach-general-availability).
 
 RBAC must be enabled on your cluster.
 
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/envoyproxy/go-control-plane/pkg/cache/v2"
 	"github.com/envoyproxy/go-control-plane/pkg/server/v2"
+	contourv1 "github.com/projectcontour/contour/apis/projectcontour/v1"
 	"github.com/projectcontour/contour/internal/annotation"
 	"github.com/projectcontour/contour/internal/contour"
 	"github.com/projectcontour/contour/internal/dag"
@@ -43,10 +44,17 @@ import (
 	"gopkg.in/alecthomas/kingpin.v2"
 	"gopkg.in/yaml.v2"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
 )
 
 // Add RBAC policy to support leader election.
 // +kubebuilder:rbac:groups="",resources=configmaps,verbs=create;get;update
+
+// Add RBAC policy to support getting CRDs.
+// +kubebuilder:rbac:groups="apiextensions.k8s.io",resources=customresourcedefinitions,verbs=list
 
 // registerServe registers the serve subcommand and flags
 // with the Application provided.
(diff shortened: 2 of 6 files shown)


youngnick:
Don't forget the other types as well - `TLSCertificateDelegation` and `ExtensionService` will also need similar `kubectl patch` commands.

youngnick:
Also, great that the actual change is quite straightforward.

stevesloka (commented):
Should we have a version check against the users cluster to verify it's 1.16+ or just let Contour fail in a bad way?

skriss:
> Should we have a version check against the users cluster to verify it's 1.16+ or just let Contour fail in a bad way?

If they're running pre-1.16, then they'll get the following error when they apply the Contour YAML:
```
unable to recognize "examples/render/contour.yaml": no matches for kind "CustomResourceDefinition" in version "apiextensions.k8s.io/v1"
```

I'm not sure where else we could put a check, given that our install method is applying raw YAML.

skriss:
I added some logic to our CRD generation script to add `preserveUnknownFields: false` to our CRD specs, which means users don't need to run a manual step to patch/remove this field.

stevesloka:
For the version check idea I was just thinking of someone who upgraded, but didn't apply the yaml properly (i.e. just updated the image). Not a huge thing to look at, but was trying to think of ways to gate people into better happy paths.

skriss:
Ah. So in that scenario, Contour will run fine; you just won't get `kubectl explain` or unknown field pruning.

We could add code to check for that scenario in Contour and log a warning -- would just need to give the Contour service account read access to CRDs. It'd be easy to implement.

skriss:
Marking this ready for review. @stevesloka @youngnick let me know if you think it's worth adding the code to check for the `preserveUnknownFields: true` field in Contour (note it requires an RBAC addition, which is really my only hesitation).

youngnick:
I think it's okay for Contour to have read access (only) to CRD definitions, that allows it to check things like versions quite nicely. That feels like good UX to me, that if things aren't going to work, we at least warn you at startup. Good point @stevesloka!

youngnick (approved):
LGTM, nice work on the awk bits to make the `preserveUnknownFields` stuff just work.

stevesloka (approved):
/lgtm

skriss:
Added code to check for non-upgraded Contour CRDs.

skriss:
RBAC seems not working, don't merge yet!

youngnick (approved):
Yep, that's a definite LGTM with the extra warning, great.

skriss:
@stevesloka I'll leave this open until EOD in case you want to take a look at the warning code.

stevesloka:
Ship it! =)

PR #2917, opened 2020-09-16
wip: implementing versioned snapshots

stevesloka:


stevesloka:
There's a logic error in this with endpoints somewhere. If I scale an app to zero, the requests fail like it should since there aren't any endpoints. Scaling back up to a higher number still has the requests fail, so the endpoints in EDS aren't getting populated properly. 

(Just adding this so I don't forget)

PR #2919, opened 2020-09-17
config file: Move xds-server-type into a ServerConfig struct

stevesloka:
Moves the xds-server-type field into a ServerConfig struct to allow
for other future server parameters to be added to the config file
under the same section.

Updates: #2134
Fixes: #2923

### DIFF
--- a/cmd/contour/servecontext.go
+++ b/cmd/contour/servecontext.go
@@ -48,13 +48,7 @@ type serveContext struct {
 	InCluster  bool   `yaml:"incluster,omitempty"`
 	Kubeconfig string `yaml:"kubeconfig,omitempty"`
 
-	// contour's xds service parameters
-	xdsAddr string
-	xdsPort int
-	// Defines the XDSServer to use for `contour serve`
-	// Defaults to "contour"
-	XDSServerType                   string `yaml:"xds-server-type,omitempty"`
-	caFile, contourCert, contourKey string
+	ServerConfig `yaml:"server,omitempty"`
 
 	// contour's debug handler parameters
 	debugAddr string
@@ -160,9 +154,6 @@ func newServeContext() *serveContext {
 	// Set defaults for parameters which are then overridden via flags, ENV, or ConfigFile
 	return &serveContext{
 		Kubeconfig:            filepath.Join(os.Getenv("HOME"), ".kube", "config"),
-		xdsAddr:               "127.0.0.1",
-		xdsPort:               8001,
-		XDSServerType:         "contour",
 		statsAddr:             "0.0.0.0",
 		statsPort:             8002,
 		debugAddr:             "127.0.0.1",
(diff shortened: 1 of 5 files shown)


cmd/contour/servecontext.go
```diff
@@ -208,6 +209,12 @@ type TLSConfig struct {
 	FallbackCertificate FallbackCertificate `yaml:"fallback-certificate,omitempty"`
 }
 
+type ServerConfig struct {
+	// Defines the XDSServer to use for `contour serve`
+	// Defaults to "contour"
+	XDSServerType string `yaml:"xds-server-type,omitempty"`
```
youngnick, line 215: nit: I slightly prefer the camelcase versions of config key names so, `xdsServerType`, over the hyphenated one., but I didn't say anything last time, so I'm okay with leaving it be.

youngnick (approved):
LGTM

cmd/contour/servecontext.go
stevesloka: I was just following how much of everything else is setup which is the hyphenated one.

cmd/contour/servecontext.go
```diff
@@ -49,13 +49,12 @@ type serveContext struct {
 	Kubeconfig string `yaml:"kubeconfig,omitempty"`
 
 	// contour's xds service parameters
-	xdsAddr string
-	xdsPort int
-	// Defines the XDSServer to use for `contour serve`
-	// Defaults to "contour"
-	XDSServerType                   string `yaml:"xds-server-type,omitempty"`
+	xdsAddr                         string
+	xdsPort                         int
```
skriss, line 53: does it make sense for these (and/or any others) to move as well? need to keep in mind compat issues if moving any exported fields

examples/contour/01-contour-config.yaml
```diff
@@ -6,8 +6,11 @@ metadata:
   namespace: projectcontour
 data:
   contour.yaml: |
-    # determine which XDS Server implementation to utilize in Contour.
-    # xds-server-type: contour
+    #
+    # server:
```
skriss, line 10: need to document this in `site/docs/main/configuration.md` as well

skriss (changes requested):

examples/contour/01-contour-config.yaml
stevesloka: just pushed that. =)

cmd/contour/servecontext.go
stevesloka: Yup I think they should, just a matter of doing that and starting up the deprecation cycle. I can make a new issue to track this work maybe rather than doing in this PR?

skriss: sure thing. Any nonexported fields should be able to just move (I think), since they're not settable via config file.

stevesloka: Nope you're right, I was overthinking all this deprecation stuff. I'll move them in now.

skriss (approved):
LGTM!

Issue #1898, opened 2019-11-13
Add support for Envoy xDS v3.


jpeach:
A year after Envoy ships xDS v3, they plan to remove support for v2. It's not entirely clear to me what that entails, but it seems likely we will need to add and qualify v3 support and maybe support v2 at the same time.

See envoyproxy/envoy#6271

jpeach:
Related to #2132

jpeach:
xref https://github.com/envoyproxy/envoy/issues/10943

bgagnon:
According to https://www.envoyproxy.io/docs/envoy/latest/api/api_supported_versions, the V2 API will stop being supported at the end of _this year_.

skriss:
We should move this into Prioritized -- we're aware of the EOY deprecation deadline and are working towards adding v3 support.

stevesloka:
I'm going to poke around a bit to see how we might address this migration.

skriss:
@stevesloka think we can consider this closed now?

stevesloka:
Yup I think so! 🎉

PR #2921, opened 2020-09-17
internal: Refactor Envoy into v2 package

stevesloka:
Refactors the envoy package into a v2 package for Envoy API V2 types.
Common items are kept in the envoy package for use by other versions.

This doesn't change anything functionally, just refactors envoy into a v2 package. 

Updates: #1898


stevesloka:
I'm going to do a 2nd pass to clean up imports and names, etc.

skriss:
@stevesloka and I talked live about this today, but recording some thoughts for others. I'm not thrilled about having to copy-paste all of this stuff (one copy for v2, another likely very similar copy for v3) but I don't have a better idea that will clearly work well at the moment. This'll definitely add to our maintenance burden while we carry both versions, since we'll potentially have to make changes in both, add tests in both.

That said, the specifics of this PR look fine to me if this is the path we're going down. I do think that once we support multiple versions, it may give us some better ideas about ways to make adding new versions in the future easier.

Interested in thoughts from other maintainers here.

youngnick:
Yes, I also don't see a better way to achieve having simultaneous support for both versions for a while, but also agree that if anything is to show us better ways it's trying it out.

Also, the annoyance factor of having to make changes in two places seems like it will make us want to get rid of this situation as soon as we can once we're confident v3 is working.

internal/envoy/listener.go
```diff
@@ -14,27 +14,9 @@
...
-	"time"
 
-	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
-	envoy_api_v2_auth "github.com/envoyproxy/go-control-plane/envoy/api/v2/auth"
-	envoy_api_v2_core "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
-	envoy_api_v2_listener "github.com/envoyproxy/go-control-plane/envoy/api/v2/listener"
-	accesslog "github.com/envoyproxy/go-control-plane/envoy/config/filter/accesslog/v2"
-	envoy_config_filter_http_ext_authz_v2 "github.com/envoyproxy/go-control-plane/envoy/config/filter/http/ext_authz/v2"
-	lua "github.com/envoyproxy/go-control-plane/envoy/config/filter/http/lua/v2"
 	http "github.com/envoyproxy/go-control-plane/envoy/config/filter/network/http_connection_manager/v2"
```
youngnick, line 19: Does this mean that the HTTP stuff still in here is also xDS v2?

youngnick (approved):
LGTM, with the already-discussed caveats.

internal/envoy/v2/secret.go
```diff
@@ -0,0 +1,44 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v2
+
+import (
+	// nolint:gosec
```
skriss, line 17: don't think this is needed, see prev comment re: `crypto/sha1`

skriss (changes requested):
Handful of nits, otherwise LGTM

internal/envoy/bootstrap.go
```diff
@@ -10,323 +10,29 @@
...
 	"github.com/golang/protobuf/jsonpb"
 	"github.com/golang/protobuf/proto"
-	"github.com/golang/protobuf/ptypes/any"
-	"github.com/projectcontour/contour/internal/protobuf"
 )
 
-// sdsResourcesSubdirectory stores the subdirectory name where SDS path resources are stored to.
-const sdsResourcesSubdirectory = "sds"
+// SdsResourcesSubdirectory stores the subdirectory name where SDS path resources are stored to.
+const SdsResourcesSubdirectory = "sds"
```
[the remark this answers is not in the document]
youngnick: In general we should do initialisms capitalised, I think, and regard places where we don't do that as tech debt.

internal/envoy/listener.go
stevesloka: yup good catch, I moved it all into the v2 package.

skriss (approved):
:shipit:

stevesloka:
I was able to get the import prefixes to work. So the standard we discussed is `envoyv2` and `contourv2` (for the next PR).

Issue #2479, opened 2020-04-29
Replace wellknown filter names with typed config


jpeach:
The go-control-plane constants for well-known filters use the deprecated filter names. Contour should switch over to using typed config for these in case some future Envoy version removes the deprecated names.

xref: https://github.com/envoyproxy/go-control-plane/issues/293

stevesloka:
I just ran into this as well, I can take and update it.

jpeach:
> I just ran into this as well, I can take and update it.

Cool! Also see https://envoyproxy.slack.com/archives/CEFDKQ3RQ/p1588129018117000

stevesloka:
Example logs:

```
[2020-04-29 22:07:24.797][1][warning][misc] [source/common/protobuf/utility.cc:198] Using deprecated option 'envoy.config.filter.network.http_connection_manager.v2.HttpFilter.config' from file http_connection_manager.proto. Th
is configuration will be removed from Envoy soon. Please see https://www.envoyproxy.io/docs/envoy/latest/intro/deprecated for details.
[2020-04-29 22:07:24.798][1][warning][misc] [bazel-out/k8-opt/bin/source/extensions/common/_virtual_includes/utility_lib/extensions/common/utility.h:65] Using deprecated extension name 'envoy.gzip' for 'envoy.filters.http.gzip
'. This name will be removed from Envoy soon. Please see https://www.envoyproxy.io/docs/envoy/latest/intro/deprecated for details.
[2020-04-29 22:07:24.798][1][warning][misc] [bazel-out/k8-opt/bin/source/extensions/common/_virtual_includes/utility_lib/extensions/common/utility.h:65] Using deprecated extension name 'envoy.grpc_web' for 'envoy.filters.http.
grpc_web'. This name will be removed from Envoy soon. Please see https://www.envoyproxy.io/docs/envoy/latest/intro/deprecated for details.
[2020-04-29 22:07:24.798][1][warning][misc] [bazel-out/k8-opt/bin/source/extensions/common/_virtual_includes/utility_lib/extensions/common/utility.h:65] Using deprecated extension name 'envoy.router' for 'envoy.filters.http.ro
uter'. This name will be removed from Envoy soon. Please see https://www.envoyproxy.io/docs/envoy/latest/intro/deprecated for details.
[...]

PR #2926, opened 2020-09-18
internal/envoy: Update GZip to valid wellknown name

stevesloka:
Updates: #2479

### DIFF
--- a/internal/envoy/listener.go
+++ b/internal/envoy/listener.go
@@ -40,10 +40,11 @@ import (
 type HTTPVersionType = http.HttpConnectionManager_CodecType
 
 const (
-	HTTPVersionAuto HTTPVersionType = http.HttpConnectionManager_AUTO
-	HTTPVersion1    HTTPVersionType = http.HttpConnectionManager_HTTP1
-	HTTPVersion2    HTTPVersionType = http.HttpConnectionManager_HTTP2
-	HTTPVersion3    HTTPVersionType = http.HttpConnectionManager_HTTP3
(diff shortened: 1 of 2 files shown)


jpeach:
IIUC, the name can be anything (at least with current Envoy versions); just specifying the typed config is enough envoy do do the right thing.

youngnick:
I don't understand what's happening here. @stevesloka, is this PR just to clear the messages from the Envoy logs? It seems like the better long-term fix is to move the filter config to the typed configs, and to stop using the names altogether?

stevesloka:
@youngnick yes the goal of this is to update the names as the current one's are deprecated and will stop working at some point. 

I'd like to move to the typedconfig, but can't seem to get that to work (with xDS v2 at least): https://github.com/envoyproxy/go-control-plane/pull/354#issuecomment-696801996

For now this bumps it down the road to avoid compatibility issues until we get it sorted out properly.

skriss:
Per [this Slack thread](https://envoyproxy.slack.com/archives/CEFDKQ3RQ/p1588129018117000) (linked in #2479) it seems like the typed-config thing is only supported in v3.

skriss:
@stevesloka I'm seeing this error in the Contour log when I test this change:
```
ERRO[0011] Error adding/updating listener(s) ingress_http: Compressor filter doesn't have compressor_library defined  code=13 connection=6 context=xds node_id=envoy-2s99g node_version=v1.15.0 response_nonce=1 version_info=2
```

stevesloka:
Closing this since the typed config isn't working with v2 xds types. Will come back when implementing v3 and I think it will work then avoiding this issue in the future.

PR #2928, opened 2020-09-20
internal/contour: Refactor v2 contour files into v2 package

stevesloka:
Refactors the contour package into a v2 package for Envoy API V2 types.
Common items are kept in the contour package for use by other versions.

This doesn't change anything functionally, just refactors contour into a v2 package.

Updates: #1898

Blocked waiting on #2921 to merge.


skriss:
ready for a rebase

stevesloka:
all rebased @skriss

cmd/contour/serve.go
```diff
@@ -29,6 +29,7 @@ import (
 	contourv1 "github.com/projectcontour/contour/apis/projectcontour/v1"
 	"github.com/projectcontour/contour/internal/annotation"
 	"github.com/projectcontour/contour/internal/contour"
+	contourv2 "github.com/projectcontour/contour/internal/contour/v2"
```
skriss, line 32: it's a little unfortunate that `contourv1` refers to the API package, while `contourv2` refers to the internal package. We could use `contourv1api` or similar for the former. Not blocking.

skriss (approved):
LGTM. I do think we could now consider moving `internal/contour/v2` into a better-named package (e.g. `internal/xdscache/v2` or something) now that they're separated out from the other things in `internal/contour`, but I'm fine following up on that later.

cmd/contour/serve.go
stevesloka: Yup I agree. =)

stevesloka: I opened #2938.

skriss:
I also opened https://github.com/projectcontour/contour/issues/2939 for the other naming-related comment.

PR #2930, opened 2020-09-21
internal/featuretests: Refactor featuretests into v2 package

stevesloka:
Refactors the featuretests package into a v2 package for Envoy API V2 types.

Common items are kept in the feature package for use by other versions.
    
Updates: #1898

Blocked on #2928

### DIFF
--- a/internal/featuretests/kubernetes.go
+++ b/internal/featuretests/kubernetes.go
@@ -22,7 +22,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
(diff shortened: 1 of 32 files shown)


skriss:
rebase time!

stevesloka:
I tried to make this as simple as possible so not much code changes. There are some bits that we can reuse in the `envoy.go` file, but I didn't want to confuse the PR too much. My thought was to follow up with another pulling out the common bits that will be needed between v2 & v3, but again, trying to keep this large PR as simple as possible.

skriss (approved):
this is straightforward, LGTM.

PR #2937, opened 2020-09-23
deps: Upgrade client-go to v1.18.9

### LINKED ISSUE #2936: Contour lose connection to Kubernetes API Server
A question came up in the last Contour Office Hours as to what would happen if Contour lost connection to the API server. In the past, client-go would reconnect and that was my answer, however, it seems there's a bug in client-go < v.18.9 where contour will just crash if connection is lost. 

ref: https://github.com/kubernetes/kubernetes/issues/91073

stevesloka:
Fixes #2936


skriss (approved):
LGTM pending CI

PR #2940, opened 2020-09-24
Make status test readable

youngnick:
This PR attempts to make the status tests more readable by putting the
fixtures next to the test cases they are used in as much as possible.

- Extracted the common Service and Secret fixtures (and the certificates) to `internal/fixtures`.
- Names are also changed. No more `proxy47a`!
- I also found some duplicate tests (probably from when we converted IngressRoute tests to HTTPProxy ones).

Blocks #2495 until it's merged; I don't want to attempt refactoring all these tests until it's a bit
easier to see what everything's used for.


internal/dag/status_test.go
```diff
@@ -29,147 +29,75 @@ import (
...
-		},
-		Spec: v1.ServiceSpec{
-			Ports: []v1.ServicePort{{
-				Name:     "http",
-				Protocol: "TCP",
-				Port:     8080,
-			}},
-		},
-	}
+	tests := make(map[string]testcase)
```
jpeach, line 39: I don't think that this goes quite far enough - there is still 2500 lines of test cases and fixtures here. Could you apply the pattern of a local `run` function rather than putting the test cases in a table? It's a very similar result, except that it make it way easier to find the test case that failed.


```Go
run := func(desc string, tc testcase) {
        r.Run(desc, func(t * testing.T) {
        ...
        }
}

run("root proxy does not specify FQDN", testcase{
        ...
})
```

jpeach (changes requested):
The main things here are that we should establish a consistent naming convention for fixture globals, and use a local `run` helper in the status tests. I haven't reviewed the test coverage in detail; the coverage percentage is unchanged, which is great, but we may want to ensure that we didn't drop important test cases.

internal/dag/status_test.go
```diff
@@ -2517,7 +2517,7 @@ func TestDAGStatus(t *testing.T) {
 			}: {
 				Object:      invalidResponseTimeout,
 				Status:      "invalid",
-				Description: "route.timeoutPolicy failed to parse: error parsing response timeout: unable to parse timeout string \"invalid-val\": time: invalid duration invalid-val",
+				Description: "route.timeoutPolicy failed to parse: error parsing response timeout: unable to parse timeout string \"invalid-val\": time: invalid duration \"invalid-val\"",
```
skriss, line 2520: Yep, I ran into this when I wrote these tests. Hazards of testing on precise error messages, I guess.

skriss (changes requested):
couple nits so far; will look through status_test.go in some more detail.

skriss (approved):
Other than the nits I'm happy with this, it's definitely an improvement in readability. I agree with James' comment that we may want to do a detailed tick and tie to ensure we didn't inadvertently drop any important test cases, but I haven't done that.

internal/dag/status_test.go
youngnick: Yeah, that's a great idea, I'll do that.

youngnick:
I only knowingly removed two test cases, that I checked carefully to ensure that were duplicates. I'll use the "show coverage" thing on `main` and here and check if I've dropped anything.

I'll also swap over to the run style rather than table style; that's not a huge change now the hard part (moving everything) is done.

youngnick:
I've gone over all of the places in `httpproxy_processor.go` that are not covered in this branch and `main`, and confirmed that they match. I've also got a list which will be handy of what we're not covering.

internal/dag/status_test.go
youngnick: It changed between Go 1.14 and Go 1.15, sigh.

internal/dag/status_test.go
youngnick: Done.

youngnick:
Only test failure is the SNI flake, I'm going to go ahead and merge this so I can bring it in to the other status work.

PR #2947, opened 2020-09-28
cmd/contour: deprecate TLS versions for gRPC XDS

### LINKED ISSUE #2944: TLS 1.1 and below should be removed from contour GRPC connection options.
Currently the TLS version of the communication between Envoy and Contour is currently defaulting to 1.2, but on the Contour side TLS 1.0 and 1.1 is allowed.

As TLS 1.1 and below is being deprecated these connection options shall be removed.

tsaarni:
Set the minimum TLS version to TLSv1.2 for gRPC XDS interface, removing
support for TLSv1.0 and TLSv1.1, which are now deprecated.

Fixes #2944

### DIFF
--- a/cmd/contour/servecontext.go
+++ b/cmd/contour/servecontext.go
@@ -385,6 +385,7 @@ func (ctx *serveContext) tlsconfig(log logrus.FieldLogger) *tls.Config {
 			ClientAuth:   tls.RequireAndVerifyClientCert,
 			ClientCAs:    certPool,
 			Rand:         rand.Reader,
+			MinVersion:   tls.VersionTLS12,
 		}, nil
 	}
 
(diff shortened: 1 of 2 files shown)


tsaarni:
Some considerations:

* The TLS version supported by the gRPC XDS server has not been controlled by Contour config file `tls.minimum-protocol-version` option. This proposed change therefore simply sets the minimum TLS version to TLSv1.2 and does not introduce configurability.
* Envoy uses TLSv1.2 to connect to Contour. See description of `tls_minimum_protocol_version` and `tls_maximum_protocol_version` in the documentation [here](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/auth/common.proto.html?highlight=tlsparameters).
* It may be possible that some use proxy between Contour and Envoy to mutate XDS configuration. In that case, this change may have impact for them.

jpeach (approved):
LGTM

PR #2948, opened 2020-09-28
testsuite: update SNI test resource checks

jpeach:
Adopt the standard resources policies to check whether HTTPProxy resources
were created and were marked valid.

This updates #2743.


_integration/testsuite/httpproxy/004-https-sni-enforcement.yaml
```diff
@@ -231,42 +214,25 @@ spec:
...
 
 fatal_proxy_is_not_valid[msg] {
-  proxy := data.resources.httpproxies[Name]
-  fqdn := proxy.spec.virtualhost.fqdn
-  status := proxy.status
+  status := resources.status("httpproxies", "echo-two")
+  validity := object.get(status, "currentStatus", "not-present")
 
-  not proxy_is_valid
+  not validity == "valid"
```
stevesloka, line 232: So this expects the validity to "not" be `valid`?

skriss (approved):
Looks OK to me

_integration/testsuite/httpproxy/004-https-sni-enforcement.yaml
jpeach: yep, it's an error in that case.

stevesloka: Maybe change `validity` to `resultStatus` or something such that it reads: `not resultStatus == "valid"` or check on the reverse such that it's `invalid`.

stevesloka (changes requested):

_integration/testsuite/httpproxy/004-https-sni-enforcement.yaml
jpeach: I renamed it to `resultStatus`.

jpeach:
Failure is:
```
not ok 25 - running Rego check lines ./_integration/testsuite/httpproxy/004-https-sni-enforcement.yaml:116-129
  ---
  - Message: |
      raised predicate "fatal_proxy_is_not_valid"
      HTTP proxy "echo-one" status is "not-present"
      {}
    Severity: Fatal
    Timestamp: "0001-01-01T00:00:00Z"
```

This would happen if Contour has never updated the status of the proxy.

stevesloka (approved):
/lgtm

PR #2949, opened 2020-09-29
Introduce xdscache package to handle xds cache resources

### LINKED ISSUE #2939: internal: consider renaming `internal/contour/v2` to something more descriptive
The XDS caches have now been extracted to their own package, `internal/contour/v2`. Consider renaming this package to something more descriptive, for example `internal/xdscache/v2`.

stevesloka:
Fixes #2939


cmd/contour/serve.go
```diff
@@ -292,21 +292,21 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+	endpointHandler.Observer = xdscache.ComposeObservers(snapshotHandler)
 
 	dnsLookupFamily, err := ParseDNSLookupFamily(ctx.DNSLookupFamily)
 	if err != nil {
 		return fmt.Errorf("failed to configure configuration file parameter cluster.dns-lookup-family: %w", err)
 	}
 
 	// Build the core Kubernetes event handler.
-	eventHandler := &contour.EventHandler{
+	eventHandler := &xdscache.EventHandler{
```
skriss, line 306: I'd probably leave the `EventHandler` type in the `internal/contour` package, not sure it makes sense under `xdscache`.

cmd/contour/serve.go
```diff
@@ -339,7 +339,7 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	// Wrap eventHandler in a converter for objects from the dynamic client.
 	// and an EventRecorder which tracks API server events.
 	dynamicHandler := &k8s.DynamicClientHandler{
-		Next: &contour.EventRecorder{
+		Next: &xdscache.EventRecorder{
```
skriss, line 342: `EventRecorder` seems like another one that should probably stay where it is.

cmd/contour/serve.go
```diff
@@ -444,7 +444,7 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 
 	// Once we have the leadership detection channel, we can
 	// push DAG rebuild metrics onto the observer stack.
-	eventHandler.Observer = &contour.RebuildMetricsObserver{
+	eventHandler.Observer = &xdscache.RebuildMetricsObserver{
```
skriss, line 447: `RebuildMetricsObserver` - keep in `internal/contour`?

skriss (changes requested):
Just looking at this again, mostly LGTM, but there are a few types that I don't think really belong under an `internal/xdscache` package - I think it'd make sense to leave them in `internal/contour` for now and we can decide if they should move somewhere else separately.

skriss (approved):
LGTM

skriss:
Probably should update the commit msg/PR title since the whole package isn't being renamed.

PR #2954, opened 2020-09-30
internal: publish ExtensionService status conditions

### LINKED ISSUE #2874: Auth: Set ExtensionService status Conditions

There's currently no internal API to set status conditions from the context of a build processor. This means that processing an ExtensionService has no way to update the status on the ExtensionService resource to publish validation or consistency errors.

Once there is an internal API, we can do this.

Blocked by #2495

jpeach:
This fixes #2874.


jpeach:
@youngnick This is what I'm thinking for the ExtensionService adoption. Once you are ready, I'll need to access the `status.Cache` correctly, and also update all the sub conditions to use a vocabulary consistent with HTTPProxy.

jpeach:
This needs a new approach due to underlying code changes.

PR #2958, opened 2020-09-30
docs: add CORS policy related documentation

### LINKED ISSUE #2943: add CORS documentation to HTTPProxy reference
We need to add documentation on CORS support to the HTTPProxy reference (https://github.com/projectcontour/contour/blob/main/site/docs/main/httpproxy.md).

It probably makes sense to add a section under https://github.com/projectcontour/contour/blob/main/site/docs/main/httpproxy.md#virtual-host-configuration, similar to the existing section there on TLS.

Followup from #2890

aberasarte:
Closes #2943

### DIFF
--- a/site/docs/main/httpproxy.md
+++ b/site/docs/main/httpproxy.md
@@ -403,6 +403,74 @@ spec:
 In this example, the permission for Contour to reference the Secret `example-com-wildcard` in the `admin` namespace has been delegated to HTTPProxy objects in the `example-com` namespace.
 Also, the permission for Contour to reference the Secret `another-com-wildcard` from all namespaces has been delegated to all HTTPProxy objects in the cluster.
 
+#### CORS policy
+
+A CORS (Cross-origin resource sharing) policy can be set for a HTTPProxy in order to allow cross-domain requests for trusted sources. If a policy is set, it will be applied to all the routes of the virtual host.
+
+Contour allows configuring the headers involved in cross-domain requests. In this example, cross-domain requests will be allowed for any domain (note the `*` value).
+
+```yaml
+apiVersion: projectcontour.io/v1
+kind: HTTPProxy
+spec:
+  virtualhost:
+    fqdn: www.example.com
+    corsPolicy:
+        allowCredentials: true
+        allowOrigin: 
+          - "*" # allows any origin
+        allowMethods:
+          - GET
+          - POST
+          - OPTIONS
+        allowHeaders: 
+          - authorization
+          - cache-control
+        exposeHeaders: 
+          - Content-Length
+          - Content-Range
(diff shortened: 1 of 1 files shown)


aberasarte:
I added a CORS policy block to the HTTPProxy reference with some basic information. I'm not sure whether a more thorough explanation of CORS and the involved headers is needed here.

site/docs/main/httpproxy.md
```diff
@@ -403,6 +403,74 @@ spec:
...
+      maxAge: "10m"
+  routes:
+    - conditions:
+      - prefix: /
+      services:
+        - name: s1
+          port: 80
+```
+
+`MaxAge` durations are expressed in the Go [duration format](https://godoc.org/time#ParseDuration). Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h". Only positive values are allowed and 0 disables the cache requiring a preflight OPTIONS	check for all cross-origin requests.
```
skriss, line 472: nit: there's a tab between "OPTIONS" and "check"

skriss (approved):
Thanks for writing this up @aberasarte! It looks reasonable to me, just one minor nit.

site/docs/main/httpproxy.md
```diff
@@ -403,6 +403,74 @@ spec:
...
+
+In the following example, cross-domain requests are restricted to https://client.example.com only.
+
+```yaml
+apiVersion: projectcontour.io/v1
+kind: HTTPProxy
+spec:
+  virtualhost:
+    fqdn: www.example.com
+  corsPolicy:
```
stevesloka, line 449: This should be indented one more to be inside the `virtualhost` struct.

site/docs/main/httpproxy.md
```diff
@@ -403,6 +403,74 @@ spec:
...
+
+Contour allows configuring the headers involved in cross-domain requests. In this example, cross-domain requests will be allowed for any domain (note the `*` value).
+
+```yaml
+apiVersion: projectcontour.io/v1
+kind: HTTPProxy
+spec:
+  virtualhost:
+    fqdn: www.example.com
+  corsPolicy:
```
stevesloka, line 418: This should be indented one more to be inside the `virtualhost` struct.

stevesloka (changes requested):
Just two small nits in the example yaml files.

site/docs/main/httpproxy.md
skriss: good catch!

stevesloka:
Also @aberasarte you need to sign the commit to make the DCO robots happy. =)

skriss (approved):
LGTM, thanks for making the updates @aberasarte.

PR #2962, opened 2020-10-01
Add code for DAG status to emit Conditions

youngnick:
Here's where I'm up to with the Conditions work.

So far, I'm at:
- DAG now emits conditions, `status_test.go` passes for everything.
- `internal/dag/metrics.go` has been updated with a new method of calculation, it turned out we were counting some things as root HTTPProxies even if they weren't valid for some reason, so I had to change the tests.
- featuretests all pass.


internal/status/cache.go
```diff
@@ -87,3 +106,45 @@ func (c Cache) getProxyStatusUpdates() []k8s.StatusUpdate {
...
+	validConds := make(map[types.NamespacedName]projcontour.DetailedCondition)
+
+	for fullname, pu := range c.proxyUpdates {
+		validConds[fullname] = *pu.Conditions[ValidCondition]
+	}
+
+	return validConds
+}
+
+type Metric struct {
```
jpeach, line 120: Metrics doesn't seem like the right name for this?

internal/k8s/statusupdater.go
```diff
@@ -152,21 +153,46 @@ func (suh *StatusUpdateHandler) Writer() StatusUpdater {
 // StatusUpdater describes an interface to send status updates somewhere.
 type StatusUpdater interface {
 	Update(name, namespace string, gvr schema.GroupVersionResource, mutator StatusMutator)
+	SendStatusUpdate(su StatusUpdate)
```
jpeach, line 156: We don't need to add an interface method here. If you want the convenience, then we can add a free function:

```Go
func SendStatusUpdate(u StatusUpdater, s StatusUpdate) {
        u.Update(s.Name.Name, s.Name.Namespace, s.Resource, s.Mutator)
}
```

internal/featuretests/featuretests.go
```diff
@@ -85,11 +86,11 @@ func setup(t *testing.T, opts ...interface{}) (cache.ResourceEventHandler, *Cont
 
 	rand.Seed(time.Now().Unix())
 
-	statusCache := &k8s.StatusCacher{}
-
+	statusCache := status.NewCache()
+	statusUpdateCacher := &k8s.StatusUpdateCacher{}
```
jpeach, line 90: Once you flip this, you can delete the `StatusClient` interface and the `StatusCache` implementation.

Maybe `StatusUpdateCacher` can take the `StatusCache` name or move into the `fixtures` package.

internal/status/cache.go
```diff
@@ -61,6 +74,12 @@ func (c Cache) commitProxy(pu *ProxyUpdate) {
 		return
 	}
 
+	validCond := pu.ConditionFor(ValidCondition)
+	if validCond.Reason == "" {
+		validCond.Reason = "Valid"
+		validCond.Message = "Valid HTTPProxy"
```
jpeach, line 80: This depends on behavior that might not always be true. I'd suggest that we default these when the condition is created.

internal/status/cache.go
```diff
@@ -87,3 +106,45 @@ func (c Cache) getProxyStatusUpdates() []k8s.StatusUpdate {
...
+		metric := Metric{
+			Vhost:     pu.Vhost,
+			Namespace: name.Namespace,
+		}
+		validCond := pu.ConditionFor(ValidCondition)
+		switch validCond.Status {
+		case projcontour.ConditionTrue:
+			metric.Status = ProxyStatusValid
+		case projcontour.ConditionFalse:
+			_, ok := validCond.GetError("orphaned")
```
jpeach, line 138: The error type name is `"Orphaned"`. Best to add constants for this.

internal/status/cache.go
```diff
@@ -87,3 +106,45 @@ func (c Cache) getProxyStatusUpdates() []k8s.StatusUpdate {
...
+	return validConds
+}
+
+type Metric struct {
+	Vhost     string
+	Status    ProxyStatus
+	Namespace string
+}
+
+func (c Cache) GetProxyStatusMetrics() []Metric {
```
jpeach, line 126: I would prefer to avoid having this function. The caller should be able to implement this in terms of `GetProxyValidConditions()`.

internal/status/cache.go
```diff
@@ -87,3 +106,45 @@ func (c Cache) getProxyStatusUpdates() []k8s.StatusUpdate {
 	return psu
 
 }
+
+func (c Cache) GetProxyValidConditions() map[types.NamespacedName]projcontour.DetailedCondition {
```
jpeach, line 110: This is only used by `(*DAG) Statuses()`, which is only used in a test. Can we remove it (and the caller)?

internal/contour/metrics_test.go
```diff
@@ -413,9 +413,7 @@ func TestHTTPProxyMetrics(t *testing.T) {
 			},
 			Valid:    map[metrics.Meta]int{},
 			Orphaned: map[metrics.Meta]int{},
-			Root: map[metrics.Meta]int{
-				{Namespace: "finance"}: 1,
-			},
+			Root:     map[metrics.Meta]int{},
```
jpeach, line 416: What's going on here, is the test wrong?

internal/dag/dag.go
```diff
@@ -79,10 +80,16 @@ func (d *DAG) Visit(fn func(Vertex)) {
 	}
 }
 
-// Statuses returns a slice of Status objects associated with
-// the computation of this DAG.
-func (d *DAG) Statuses() map[types.NamespacedName]Status {
-	return d.statuses
+// GetTestStatuses returns a slice of Status objects associated with
+// the computation of this DAG, for testing status output.
```
jpeach, line 84: Is the function defined to only emit HTTPProxy conditions? Can we add a TODO and an issue to remove it?

internal/dag/httpproxy_processor.go
```diff
@@ -219,8 +224,7 @@ func (p *HTTPProxyProcessor) computeHTTPProxy(proxy *contour_api_v1.HTTPProxy) {
 				ref := defaultExtensionRef(auth.ExtensionServiceRef)
 
 				if ref.APIVersion != contour_api_v1alpha1.GroupVersion.String() {
-					sw.SetInvalid("Spec.Virtualhost.Authorization.ServiceRef specifies an unsupported resource version %q",
-						auth.ExtensionServiceRef.APIVersion)
+					validCond.AddErrorf("AuthError", "AuthBadResourceVersion", "Spec.Virtualhost.Authorization.extensionRef specifies an unsupported resource version %q", auth.ExtensionServiceRef.APIVersion)
```
jpeach, line 227:
```suggestion
					validCond.AddErrorf("AuthError", "AuthBadResourceVersion",
					        "Spec.Virtualhost.Authorization.extensionRef specifies an unsupported resource version %q",
					        auth.ExtensionServiceRef.APIVersion)
```

Break super long lines.

internal/featuretests/replaceprefix_test.go
```diff
@@ -153,10 +150,7 @@ func basic(t *testing.T) {
 			envoy_v2.RouteConfiguration("ingress_http"),
 		),
 		TypeUrl: routeType,
-	}).Status(vhost).Equals(contour_api_v1.HTTPProxyStatus{
-		CurrentStatus: k8s.StatusInvalid,
-		Description:   "duplicate replacement prefix '/foo'",
-	})
+	}).Status(vhost).HasError("PrefixReplaceError", "PrefixReplaceError", "duplicate replacement prefix '/foo'")
```
jpeach, line 153: If we have `PrefixReplaceError` as the type, then we can have separate reasons for the different error cases (ambiguity, duplicates).

internal/fixture/detailedcondition.go
```diff
@@ -0,0 +1,91 @@
...
+package fixture
+
+import (
+	v1 "github.com/projectcontour/contour/apis/projectcontour/v1"
+)
+
+// DetailedConditionBuilder is a builder object to make creating HTTPProxy fixtures more succinct.
+type DetailedConditionBuilder v1.DetailedCondition
+
+// NewProxy creates a new ProxyBuilder with the specified object name.
```
jpeach, line 23:
```suggestion
// NewValidCondition creates a new DetailedConditionBuilder.
```

internal/k8s/statuscache.go
```diff
@@ -0,0 +1,134 @@
...
+	case *contour_api_v1.HTTPProxy:
+		objectKey := suc.objKey(o.Name, o.Namespace, contour_api_v1.HTTPProxyGVR)
+		cachedObj, ok := suc.objectCache[objectKey]
+		if ok {
+			switch c := cachedObj.(type) {
+			case *contour_api_v1.HTTPProxy:
+				return &c.Status, nil
+			}
+		}
+		fmt.Printf("%#v\n", suc.objectCache)
```
jpeach, line 112: Delete debug line.

internal/status/cache.go
```diff
@@ -18,12 +18,22 @@ package status
...
 )
 
+type ProxyStatus string
+
+const (
+	ProxyStatusValid    ProxyStatus = "valid"
+	ProxyStatusInvalid  ProxyStatus = "invalid"
+	ProxyStatusOrphaned ProxyStatus = "orphaned"
+
+	OrphanedConditionType string = "Orphaned"
```
jpeach, line 34: This should be of type `ConditionType`.

internal/status/cache.go
```diff
@@ -61,6 +76,18 @@ func (c Cache) commitProxy(pu *ProxyUpdate) {
 		return
 	}
 
+	_, ok := c.proxyUpdates[pu.Fullname]
+	if ok {
+		// When we're committing, if we already have a Valid Condition with an error, and we're trying to
+		// set the object back to Valid, skip the commit, as we've visited too far down.
+		// If this is removed, the status reporting for when a parent delegates to a child that delegates to itself
+		// will not work. Yes, I know, problems everywhere. I'm sorry.
```
jpeach, line 84: ref an issue to clean this up.

internal/status/proxystatus.go
```diff
@@ -91,16 +96,18 @@ func (pu *ProxyUpdate) Mutate(obj interface{}) interface{} {
 	case projectcontour.ConditionTrue:
 		// TODO(youngnick): bring the k8s.StatusValid constants in here?
 		proxy.Status.CurrentStatus = k8s.StatusValid
-		proxy.Status.Description = validCond.Reason + ": " + validCond.Message
+		proxy.Status.Description = validCond.Message
 	case projectcontour.ConditionFalse:
-		orphanCond, orphaned := validCond.GetError(k8s.StatusOrphaned)
+		orphanCond, orphaned := validCond.GetError(OrphanedConditionType)
```
jpeach, line 101:
```Go
if orphanCond, ok := validCond.GetError(OrphanedConditionType); ok {
 			proxy.Status.CurrentStatus = k8s.StatusOrphaned
  			proxy.Status.Description = orphanCond.Message
  			break
}
```

skriss (changes requested):
Handful of things so far, still looking..

internal/status/cache.go
```diff
@@ -87,3 +114,16 @@ func (c Cache) getProxyStatusUpdates() []k8s.StatusUpdate {
 	return psu
 
 }
+
+// GetProxyUpdates gets the underlying ProxyUpdate objects
+// from the cache, used by various things (`internal/contour/metrics.go` and `internal/dag/status_test.go`)
+// to retrieve info they need.
+// TODO(youngnick): This could conceivably be replaced with a Walk pattern.
+func (c Cache) GetProxyUpdates() []*ProxyUpdate {
```
jpeach, line 122: ref an issue to clean this up

cmd/contour/serve.go
```diff
@@ -486,11 +486,9 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	}
 	g.Add(sh.Start)
 
-	// Now we have the statusUpdateWriter, we can create the StatusWriter, which will take the
+	// Now we have the statusUpdateWriter, we can create the event handler's StatusUpdater, which will take the
```
skriss, line 489: nit: s/statusUpdateWriter/statusUpdateHandler

Related, a few lines above this the log context is still "StatusUpdateWriter", should now be "StatusUpdateHandler".

internal/contour/metrics.go
```diff
@@ -113,14 +110,18 @@ func calculateRouteMetric(statuses map[types.NamespacedName]dag.Status) metrics.
...
+	switch validCond.Status {
+	case contour_api_v1.ConditionTrue:
+		metricValid[metrics.Meta{VHost: u.Vhost, Namespace: u.Fullname.Namespace}]++
+	case contour_api_v1.ConditionFalse:
+		_, ok := validCond.GetError(status.OrphanedConditionType)
+		if ok {
+			metricOrphaned[metrics.Meta{Namespace: u.Fullname.Namespace}]++
+			break
+		}
+		metricInvalid[metrics.Meta{VHost: u.Vhost, Namespace: u.Fullname.Namespace}]++
```
jpeach, line 124:
```suggestion
		if _, ok := validCond.GetError(status.OrphanedConditionType); ok {
  			metricOrphaned[metrics.Meta{Namespace: u.Fullname.Namespace}]++	
  		} else {
  		        metricInvalid[metrics.Meta{VHost: u.Vhost, Namespace: u.Fullname.Namespace}]++
  		}
```

internal/status/cache.go
```diff
@@ -18,12 +18,22 @@ package status
...
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 )
 
+type ProxyStatus string
+
+const (
+	ProxyStatusValid    ProxyStatus = "valid"
+	ProxyStatusInvalid  ProxyStatus = "invalid"
+	ProxyStatusOrphaned ProxyStatus = "orphaned"
```
skriss, line 32: these constants are also defined in internal/k8s/status.go, can we consolidate to one set?

skriss (changes requested):

internal/status/proxystatus.go
```diff
@@ -51,7 +49,14 @@ func (pu *ProxyUpdate) ConditionFor(cond ConditionType) *projectcontour.Detailed
 	if !ok {
 		newDc := &projectcontour.DetailedCondition{}
 		newDc.Type = string(cond)
-
+		newDc.ObservedGeneration = pu.Generation
+		if newDc.IsPositivePolarity() {
+			newDc.Status = projectcontour.ConditionTrue
+			newDc.Reason = "Valid"
+			newDc.Message = "Valid HTTPProxy"
```
jpeach, line 56: `IsPositivePolarity` treats the condition generically, but then the reason and message assume that the condition was `ValidCondition`. So just be explicit that this expects a `ValidCondition`.

internal/k8s/statuscache.go
```diff
@@ -0,0 +1,134 @@
...
+// OnAdd adds an object to the status cache.
+func (suc *StatusUpdateCacher) OnAdd(obj interface{}) {
+	if suc.objectCache == nil {
+		suc.objectCache = make(map[string]interface{})
+	}
+
+	switch o := obj.(type) {
+	case *contour_api_v1.HTTPProxy:
+		suc.objectCache[suc.objKey(o.Name, o.Namespace, contour_api_v1.HTTPProxyGVR)] = obj
+		// fmt.Printf("Saved %s to cache, cache has %d entries\n", suc.objKey(o.Name, o.Namespace, contour_api_v1.HTTPProxyGVR), len(suc.objectCache))
```
stevesloka, line 61: nit: Remove comment.

internal/k8s/statuscache.go
```diff
@@ -0,0 +1,134 @@
...
+	default:
+		panic(fmt.Sprintf("status caching not supported for object type %T", obj))
```
stevesloka, line 63: I still don't like the panic when an invalid type is encountered, not sure how others feel about but, but to me, doesn't seem like case where Contour should be terminated.

stevesloka: If it needs to stay, should update the message to understand a bit more about where this was encountered.

jpeach: This is an assertion that programmers obeyed the invariants, so panic is OK by me.

internal/dag/httpproxy_processor.go
```diff
@@ -90,16 +91,19 @@ func (p *HTTPProxyProcessor) Run(dag *DAG, source *KubernetesCache) {
 	for meta := range p.orphaned {
 		proxy, ok := p.source.httpproxies[meta]
 		if ok {
-			sw, commit := p.dag.WithObject(proxy)
-			sw.WithValue("status", k8s.StatusOrphaned).
-				WithValue("description", "this HTTPProxy is not part of a delegation chain from a root HTTPProxy")
+			pa, commit := p.dag.StatusCache.ProxyAccessor(proxy)
+			pa.ConditionFor(status.ValidCondition).AddError(status.OrphanedConditionType,
```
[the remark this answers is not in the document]
youngnick: There's only one type of Condition, called `Valid`, that can be either True or False. `Valid`, `true` = `valid`, `Valid`, `false` = `invalid`.

internal/dag/httpproxy_processor.go
```diff
@@ -90,16 +91,19 @@ func (p *HTTPProxyProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+				"this HTTPProxy is not part of a delegation chain from a root HTTPProxy")
 			commit()
 		}
 	}
 }
 
 func (p *HTTPProxyProcessor) computeHTTPProxy(proxy *contour_api_v1.HTTPProxy) {
-	sw, commit := p.dag.WithObject(proxy)
+	pa, commit := p.dag.StatusCache.ProxyAccessor(proxy)
+	validCond := pa.ConditionFor(status.ValidCondition)
```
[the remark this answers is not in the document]
youngnick: See my comment above, the condition is always _called_ `Valid`, but is either true or false when valid or invalid respectively.

internal/k8s/statuscache.go
youngnick: Yeah, the panic will only happen if a developer has passed the wrong type of thing, so this  should never happen. This is belt-and-suspenders territory.

internal/dag/dag.go
youngnick: I've renamed it. We won't be able to remove it completely without completely refactoring the status tests, as they need the name of the object and the detailed object. That will need a lot of extra testing machinery to sit between the `status.Cache` and the status tests.

#2967 opened to track.

internal/status/cache.go
youngnick: I'll create an issue, but I don't see how it's fixable, as it's an artifact of:
- Conditions are designed to allow multiple errors to be added, so we removed the behavior that only the first `commit()` makes it into the cache.
- If you allow multiple commits, in this case, because the child is recursed into twice during `computeRoutes`, the error is added correctly once, and then it's overwritten by a `Valid` condition.
- So, we're going to have to do something to check that we don't flip a condition back to `Valid`: `true` if there's already an error on it.

youngnick: #2968 is the related issue.

internal/dag/status_test.go
```diff
@@ -114,18 +118,21 @@ func TestDAGStatus(t *testing.T) {
 		},
 	}
 
-	run("valid proxy", testcase{
+	run(t, "valid proxy", testcase{
```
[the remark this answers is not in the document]
youngnick: I'm not sure what you mean? It's below the proxy where it's used, like the rest.

internal/contour/metrics_test.go
youngnick: It was, but I hadn't looked deeply enough into why; I needed to change the order of some things in `httpproxy_processor` for the information to flow out correctly.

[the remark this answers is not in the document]
youngnick: It's because the object itself wasn't available; I had to change the order of things at the top of `computeRoutes` in `httpproxy_processor.go` to enable the vhost info to propagate out correctly. 

The remaining changes in `metrics_test.go` are because we actually emit more info now about the vhost. In addition, there was one time when we were actually counting things incorrectly.

internal/status/cache.go
youngnick: Done.

internal/dag/httpproxy_processor.go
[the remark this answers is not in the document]
stevesloka: That honestly makes no sense to me. What's the point of true or false then? Just get rid of that and call it what it is.

internal/dag/status_test.go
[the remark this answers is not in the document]
stevesloka: I guess that's my fault for getting this far and not understanding/speaking up. But honestly, to me its way confusing to have something be "valid", but contain errors. It's like saying my soup is "done" but it still has 30mins to cook.

internal/status/cache.go
jpeach: It sounds to me like this is a problem with the API design but we can address in #2968.

> I'll create an issue, but I don't see how it's fixable, as it's an artifact of:
> 
> * Conditions are designed to allow multiple errors to be added, so we removed the behavior that only the first `commit()` makes it into the cache.

Need to consider what commit really means. If we committed a change then we want to make a change on top of that commit, the API should support that rather than hack it up at the end.

> * If you allow multiple commits, in this case, because the child is recursed into twice during `computeRoutes`, the error is added correctly once, and then it's overwritten by a `Valid` condition.

I would have thought that a child proxy was either valid or not; I'm surprised that it depends on the context. If it is dependent on context, then I expect this could also happen across proxies.

> * So, we're going to have to do something to check that we don't flip a condition back to `Valid`: `true` if there's already an error on it.

Yup, sounds like it. I think that the problem here is that `ProxyAccessor` isn't cache-aware. Each time the cache is accessed, the latest accessor should get the most recent condition state. We need to be careful about reentrancy too (either disallow that or have some merge strategy when it happens).

internal/dag/httpproxy_processor.go
```diff
@@ -127,6 +120,13 @@ func (p *HTTPProxyProcessor) computeHTTPProxy(proxy *contour_api_v1.HTTPProxy) {
 	}
 	pa.Vhost = host
 
+	// ensure root httpproxy lives in allowed namespace
```
jpeach, line 123: Add a comment explaining why the ordering of these checks is important?

jpeach (approved):

internal/dag/httpproxy_processor.go
youngnick: Good idea, thanks.

PR #2963, opened 2020-10-01
Add whitespace linting.

jpeach:

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -11,6 +11,7 @@ linters:
   - misspell
   - unconvert
   - unparam
+  - wsl
 
 # TODO(jpeach): enable these later:
 #  - gocyclo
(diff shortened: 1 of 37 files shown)


jpeach:
This starts taking longer than I expected, so PTAL.

stevesloka:
I don't feel super strong about spaces like this, but I do agree with whatever we decide automating it with linters.

skriss:
The rules look reasonble to me (I'm generally a fan of more whitespace to separate logical sections in the code). 

I will ask the question though, of whether we think precise consistency here is important/valuable enough to make this a blocking CI check? While I like consistency, adhering to these standards will add overhead to every PR, particularly for newer contributors who are less familiar with the precise rules, and I don't know if the benefit of standardized whitespace is worth that cost. I'm willing to try it out and see how it goes, as long as we can revisit later if we think it's too burdensome.

youngnick:
I think if we're going to add this, we should definitely do the other PR that puts the linting errors in as comments - this will make this much faster for contributors to fix things up.

jpeach:
Closing this out for now. It's more whitespace changes than I'm really enthusiastic about updating :)

Issue #2968, opened 2020-10-02
internal/status: recursive commits must not overwrite an error condition with one with no error


youngnick:
During the HTTPProxy processor step, we compute which HTTPProxies are valid by recursively calling `computeRoutes()`. In the case that a child HTTPProxy includes itself, producing a cycle, we recurse into the object twice, once for the initial include from its parent, and then one for its self-include. We pick up the cycle on the final pass, and add an error condition. However, the *containing* copy of the child proxy is valid, aside from the include being bad, so unless we check for the changing of a condition from errors to no errors during the `commit()` function, this check will be broken.

This is because:
- Conditions are designed to allow multiple errors to be added, so we removed the behavior that only the first commit() makes it into the cache.
- If you allow multiple commits, in this case, because the child is recursed into twice during computeRoutes, the error is added correctly once, and then it's overwritten by a Valid condition.
- So, we're going to have to do something to check that we don't flip a condition back to Valid: true if there's already an error on it.

When I discovered this issue, I put a fix in - checking on commit if the transition is from "condition has some errors" to " condition has no errors", and not committing if that's the case. This is the only time, I think, when this sequence of events can be hit.

PR #2975, opened 2020-10-02
internal: update ExtensionService status conditions

### LINKED ISSUE #2874: Auth: Set ExtensionService status Conditions

jpeach:
Add status cache support for `ExtensionService` resources. This is built
on a more general cache interface, which ought to be extensible to more
types, but that whill need to wait for additional refactoring passes.

Adopt the status cache in the `ExtensionService` processor so that
conditions are correctly updates for `ExtensionService` resources.
The processor doesn't fail on errors, which allows multiple errors
to be published in the same condition.

This fixes #2874.

### DIFF
--- a/apis/projectcontour/v1alpha1/helpers.go
+++ b/apis/projectcontour/v1alpha1/helpers.go
@@ -1,0 +1,28 @@
+// Copyright Project Contour Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1alpha1
+
+import contour_api_v1 "github.com/projectcontour/contour/apis/projectcontour/v1"
+
+// GetConditionFor returns the a pointer to the condition for a given type,
+// or nil if there are none currently present.
+func (status *ExtensionServiceStatus) GetConditionFor(condType string) *contour_api_v1.DetailedCondition {
+	for i, cond := range status.Conditions {
+		if cond.Type == condType {
+			return &status.Conditions[i]
+		}
+	}
+
+	return nil
+}
--- a/internal/dag/extension_processor.go
+++ b/internal/dag/extension_processor.go
@@ -17,10 +17,13 @@ import (
 	"path"
 	"strings"
 
-	"github.com/projectcontour/contour/apis/projectcontour/v1alpha1"
+	contour_api_v1 "github.com/projectcontour/contour/apis/projectcontour/v1"
+	contour_api_v1alpha1 "github.com/projectcontour/contour/apis/projectcontour/v1alpha1"
 	"github.com/projectcontour/contour/internal/k8s"
+	"github.com/projectcontour/contour/internal/status"
 	"github.com/projectcontour/contour/internal/xds"
 	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
@@ -28,8 +31,9 @@ import (
 type ExtensionServiceProcessor struct {
 	logrus.FieldLogger
 
-	// ClientCertificate is the optional identifier of the TLS secret containing client certificate and
-	// private key to be used when establishing TLS connection to upstream cluster.
+	// ClientCertificate is the optional identifier of the TLS
+	// secret containing client certificate and private key to be
+	// used when establishing TLS connection to upstream cluster.
 	ClientCertificate *types.NamespacedName
 }
 
@@ -37,9 +41,22 @@ var _ Processor = &ExtensionServiceProcessor{}
 
 func (p *ExtensionServiceProcessor) Run(dag *DAG, cache *KubernetesCache) {
 	for _, e := range cache.extensions {
-		if ext := p.buildExtensionService(cache, e); ext != nil {
-			dag.AddRoot(ext)
+		extStatus, commit := status.ExtensionAccessor(&dag.StatusCache, e)
+		validCondition := extStatus.ConditionFor(status.ValidCondition)
+
+		if ext := p.buildExtensionService(cache, e, validCondition); ext != nil {
+			if len(validCondition.Errors) == 0 {
+				dag.AddRoot(ext)
+			}
 		}
+
+		if len(validCondition.Errors) == 0 {
+			validCondition.Status = contour_api_v1.ConditionTrue
+			validCondition.Reason = "Valid"
+			validCondition.Message = "Valid ExtensionService"
+		}
+
+		commit()
 	}
 }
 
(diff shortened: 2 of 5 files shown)


apis/projectcontour/v1alpha1/helpers.go
```diff
@@ -0,0 +1,28 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1alpha1
+
+import contourv1 "github.com/projectcontour/contour/apis/projectcontour/v1"
```
skriss, line 16: nit: alias as `contour_api_v1`

internal/status/cache.go
```diff
@@ -38,22 +38,50 @@ const (
...
 // Cache holds status updates from the DAG back towards Kubernetes.
 // It holds a per-Kind cache, and is intended to be accessed with a
 // KindAccessor.
 type Cache struct {
 	proxyUpdates map[types.NamespacedName]*ProxyUpdate
+
+	entries map[string]map[types.NamespacedName]CacheEntry
+}
+
+func (c *Cache) Acquire(obj k8s.Object) CacheEntry {
```
skriss, line 59: `Acquire` and `Release` imply to me some kind of exclusive lock/lease semantics, which isn't the case here - maybe just `Get` and `Put` would be more accurate.

internal/status/cache.go
```diff
@@ -38,22 +38,50 @@ const (
...
+	kind := k8s.KindOf(obj)
```
skriss, line 60: `k8s.KindOf` apparently has a hardcoded map of type to kind, and `ExtensionService` is missing from it (so it's always returning the empty string).

internal/status/cache.go
```diff
@@ -38,22 +38,50 @@ const (
...
+	ConditionFor(ConditionType) *contour_api_v1.DetailedCondition
+}
+
 // Cache holds status updates from the DAG back towards Kubernetes.
 // It holds a per-Kind cache, and is intended to be accessed with a
 // KindAccessor.
 type Cache struct {
 	proxyUpdates map[types.NamespacedName]*ProxyUpdate
+
+	entries map[string]map[types.NamespacedName]CacheEntry
```
skriss, line 56: it'd be helpful to reiterate in a comment here that the outer map is keyed on Kind.

skriss (changes requested):
This approach LGTM, had a handful of comments.

internal/status/cache.go
jpeach: These names emphasize the you are getting (possibly shared) access to an entry. Later on, we may need to deal with HTTPProxy using this in a reentrant pattern. "get" and "put" (at least to me), imply getting copies.

internal/status/cache.go
jpeach: fixed

internal/status/cache.go
jpeach: https://github.com/projectcontour/contour/blob/main/internal/k8s/kind.go#L47-L48

Also update the test to check this.

skriss: guess i'm either blind or was looking at something stale. 🤷‍♂️

internal/status/cache.go
stevesloka: +1 on Aquire & Release implying a lock/lease semantic. Contour doesn't use this terminology anywhere and I think it does make it less clear. The user shouldn't need to understand if the entry is shared or not and that logic should be hidden away in the implementation of the methods.

internal/dag/extension_processor.go
```diff
@@ -148,21 +160,16 @@ func (p *ExtensionServiceProcessor) buildExtensionService(
 
 		svc, port, err := cache.LookupService(svcName, intstr.FromInt(target.Port))
 		if err != nil {
-			// TODO(jpeach): Add status condition, #2874.
-			p.WithError(err).
-				WithField("name", svcName).
-				WithField("port", target.Port).
-				Error("failed to look up service")
-			return nil
+			validCondition.AddErrorf("ServiceError", "ServiceUnresolvedReference",
```
stevesloka, line 163: Do we have any plans to make these categories lookups? Meaning `ServiceError` & `ServiceUnresolvedReference` should have some sort of central place along with docs that explain their meaning.

internal/status/extensionstatus.go
```diff
@@ -0,0 +1,111 @@
...
+	TransitionTime v1.Time
+}
+
+var _ CacheEntry = &ExtensionCacheEntry{}
+
+func (e *ExtensionCacheEntry) AsStatusUpdate() k8s.StatusUpdate {
+	m := k8s.StatusMutatorFunc(func(obj interface{}) interface{} {
+		o, ok := obj.(*contour_api_v1alpha1.ExtensionService)
+		if !ok {
+			panic(fmt.Sprintf("unsupported %T object %q in status mutator", obj, e.Name))
```
stevesloka, line 70: boo panic().

internal/status/extensionstatus.go
```diff
@@ -0,0 +1,111 @@
...
+	})
+
+	return k8s.StatusUpdate{
+		NamespacedName: e.Name,
+		Resource:       contour_api_v1alpha1.ExtensionServiceGVR,
+		Mutator:        m,
+	}
+}
+
+func ExtensionAccessor(c *Cache, ext *contour_api_v1alpha1.ExtensionService) (*ExtensionCacheEntry, func()) {
```
stevesloka, line 98: nit: Comment for what this does.

internal/status/extensionstatus.go
```diff
@@ -0,0 +1,111 @@
...
+	entry := c.Acquire(ext)
+	if entry == nil {
+		entry = &ExtensionCacheEntry{
+			Name:           k8s.NamespacedNameOf(ext),
+			Generation:     ext.GetGeneration(),
+			TransitionTime: v1.NewTime(time.Now()),
+		}
+	}
+
+	return entry.(*ExtensionCacheEntry), func() {
```
stevesloka, line 108: Should you check the type here an panic if not an `ExtensionCacheEntry`?

jpeach: This is an implicit panic because it is an unchecked cast.

internal/dag/extension_processor.go
jpeach: Yeh we should do this. I did this change quickly because I was told it was needed for the release. I'll file a separate issue to do add contents which goods can be attached to. Not sure we need to expose these names as user docs.

internal/dag/extension_processor.go
```diff
@@ -110,30 +126,26 @@ func (p *ExtensionServiceProcessor) buildExtensionService(
...
+			// future.
+			//
+			// TODO(jpeach): expose SNI in the API, https://github.com/projectcontour/contour/issues/2893.
+			extension.SNI = uv.SubjectName
 		}
 
 		if extension.Protocol != "h2" {
-			// TODO(jpeach): Add status condition, #2874.
-			p.WithError(err).Errorf(
+			validCondition.AddErrorf("SpecError", "InconsistentProtocol",
```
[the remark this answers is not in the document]
jpeach: Yeh, this will make the "valid" condition false.

internal/status/cache.go
jpeach: It's for https://github.com/projectcontour/contour/issues/2968

internal/dag/extension_processor.go
jpeach: I filed https://github.com/projectcontour/contour/issues/2988

internal/status/cache.go
jpeach: > +1 on Aquire & Release implying a lock/lease semantic. Contour doesn't use this terminology anywhere and I think it does make it less clear. The user shouldn't need to understand if the entry is shared or not and that logic should be hidden away in the implementation of the methods.

Something has to know that the cache entry can be shared. Either the cache needs to arbitrate access, the entry needs to arbitrate, or the caller needs to deal with it. In the specific case of `ExtensionService` we know that the caller can do that, and we also know that there is no reentrancy.

I'm cautiously optimistic that we can move HTTPProxy to use the generic cache, but that's a bigger change that needs more time and care.

I renamed the methods in the interest of moving this review forward.

skriss (approved):
LGTM :shipit:

internal/dag/extension_processor.go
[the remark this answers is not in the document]
stevesloka: I understand what is does, just the name is confusing.

stevesloka (approved):
I'll keep my preference to not call it `validCondition` and add errors along with not adding `panics()` to go code when types aren't supported. 

/lgtm

PR #2976, opened 2020-10-02
add golangci lint GH action

### LINKED ISSUE #2972: Use the Github Action for Go linting

Run the linter from the Github action for better PR integration

https://github.com/golangci/golangci-lint-action

bonus points if we can add this to https://github.com/projectcontour/.github/ in such a way that every repo gets it automatically.

Glyphack:
use [golangci-lint](https://github.com/marketplace/actions/run-golangci-lint) action to lint new code changes on PRs.

Fixes #2972

Signed-off-by: Shayegan Hooshyari  <sh.hooshyari@gmail.com>


jpeach:
@Glyphack Could you add a lint error to this PR to see how the GH action displays it?

Glyphack:
@jpeach I added a unused function here is how it looks when action fails:
image

jpeach:
> @jpeach I added a unused function here is how it looks when action fails:
> image
> The implementation of commenting on PRs is not available yet. see: [golangci/golangci-lint-action#5](https://github.com/golangci/golangci-lint-action/issues/5)

@stevesloka I think this is a worthwhile improvement, WDYT?

.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,14 @@
...
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v2
+        with:
+          version: v1.29
```
stevesloka, line 13: Any way we can keep the versions in sync with the `Makefile`? (https://github.com/projectcontour/contour/blob/main/hack/golangci-lint#L15)

stevesloka (changes requested):
Also will need to remove the steps which run this in the PR GH Action currently: https://github.com/projectcontour/contour/blob/main/.github/workflows/prbuild.yaml#L48-L58

.github/workflows/golangci-lint.yaml
Glyphack: I did not quite understand what you mean. Do you mean is there any way to sync these two versions or syncing it now from v1.29 to v.1.31.0?

jpeach: I can't think of a great way to have only a single place with the version. We might just have to live with it in multiple places like we do with the Go version :(

stevesloka: There's a version in the hacks/golanglint-ci script as well as now in this GH Action PR. So we need to be diligent in keeping the two in in sync. This is the benefit of just leaving the Makefile doing the work and not having a separate action.

skriss:
Can someone catch me up on https://github.com/projectcontour/.github/pull/3 and this PR? Reading the backscroll, the idea of the `.github` repo PR was that it would then be available to all org repos, so in theory we shouldn't need this PR. Is that not the case? (I'm not familiar with how the org-level action sharing works).

Glyphack:
> Can someone catch me up on [projectcontour/.github#3](https://github.com/projectcontour/.github/pull/3) and this PR? Reading the backscroll, the idea of the `.github` repo PR was that it would then be available to all org repos, so in theory we shouldn't need this PR. Is that not the case? (I'm not familiar with how the org-level action sharing works).

Me and @jpeach thought that placing workflows under `.github` directory will be used as default actions, like `CONTRIBUTING.md ` and `CODE_OF_CONDUCT.md`. But now I searched for it and found this [link](https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/sharing-workflows-with-your-organization#creating-a-workflow-template) simply it means having actions under `workflow-templates` directory will allow you to copy a workflow between org repositories.
I don't know having it under `.github` directory have same effect or not. If it's not working(your can check whether it's available in workflow templates or not) I can send another PR and fix it.

jpeach:
> > Can someone catch me up on [projectcontour/.github#3](https://github.com/projectcontour/.github/pull/3) and this PR? Reading the backscroll, the idea of the `.github` repo PR was that it would then be available to all org repos, so in theory we shouldn't need this PR. Is that not the case? (I'm not familiar with how the org-level action sharing works).

> I don't know having it under `.github` directory have same effect or not. If it's not working(your can check whether it's available in workflow templates or not) I can send another PR and fix it.

Yeh, I thought that `.github` worked for workflows, but it turns out that knative has custom automation that copies the workflows around. So that was a bust :(

Glyphack:
> > > Can someone catch me up on [projectcontour/.github#3](https://github.com/projectcontour/.github/pull/3) and this PR? Reading the backscroll, the idea of the `.github` repo PR was that it would then be available to all org repos, so in theory we shouldn't need this PR. Is that not the case? (I'm not familiar with how the org-level action sharing works).
> 
> > I don't know having it under `.github` directory have same effect or not. If it's not working(your can check whether it's available in workflow templates or not) I can send another PR and fix it.
> 
> Yeh, I thought that `.github` worked for workflows, but it turns out that knative has custom automation that copies the workflows around. So that was a bust :(

We can still move it into `workflow-templates` and It works according to github, right?

jpeach:
> @jpeach We can still move it into `workflow-templates` and It works according to github, right?

IIUC, that copies the workflow to each repo, and doesn't update when the template does. I'd be happy if I was wrong about that though :)

.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,14 @@
+name: golangci-lint
```
skriss, line 1: is there any specific reason to break this out into its own workflow, rather than adding it as a job to the existing prbuild.yaml workflow file (in place of the existing `lint` job)? If no specific reason, my pref would be to keep it in one workflow file for simplicity.

skriss (commented):
Thanks for the PR @Glyphack!

I'm personally fine with or without this PR, I don't really have a preference either way for if the linting runs as an action or not. I do find the integrated comments from the codecov action annoying and wish we could toggle them off, but this one probably won't be as spammy.

See one specific comment.

jpeach:
> I do find the integrated comments from the codecov action annoying and wish we could toggle them off

I also find the codecov annotations annoying when I'm reviewing. There's a toggle to turn them off in the top right of the source diff view.

skriss:
> > I do find the integrated comments from the codecov action annoying and wish we could toggle them off
> 
> I also find the codecov annotations annoying when I'm reviewing. There's a toggle to turn them off in the top right of the source diff view.

Where is this toggle?! I still can't find it.

.github/workflows/golangci-lint.yaml
Glyphack: First I thought this should run on push to master too. Then decided to keep it for PRs, And yes I move it to `prbuild.yaml`

skriss (approved):
Looks reasonable to me!

youngnick (approved):
LGTM

PR #2980, opened 2020-10-04
Remove custom issue templates.

jpeach:
Remove the custom issue templates so that this repository can use the
standard project-wide ones from the `.github` repository. This ensures
that updates to those templates automatically propagate here.


stevesloka (commented):
I think these issue templates need some love and having them per repo might be nice to keep. 

For instance, the questions we ask about the `contour` repo will be different than those we ask for about say the authserver repo.

jpeach:
> I think these issue templates need some love and having them per repo might be nice to keep.

If you want to keep these ones, that's OK by me.

PR #2983, opened 2020-10-05
internal/dag: rename `validHTTPProxies` for clarity

jpeach:
The `(*HTTPProxyProcessor)validHTTPProxies()` method only returns
non-conflicing HTTPProxy roots. Rename it for clarity, and copy-edit
the comments a bit.

### DIFF
--- a/internal/dag/httpproxy_processor.go
+++ b/internal/dag/httpproxy_processor.go
@@ -84,7 +84,7 @@ func (p *HTTPProxyProcessor) Run(dag *DAG, source *KubernetesCache) {
 		p.orphaned = nil
 	}()
 
-	for _, proxy := range p.validHTTPProxies() {
+	for _, proxy := range p.listRootProxies() {
 		p.computeHTTPProxy(proxy)
 	}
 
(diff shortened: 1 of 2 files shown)


stevesloka (commented):
I think the current name `validHTTPProxies` is more clear. List implies that you get all of the proxies back, but valid is clear that some logic is being applied in the resulting set. I don't think this change is needed.

jpeach:
> I think the current name `validHTTPProxies` is more clear. List implies that you get all of the proxies back, but valid is clear that some logic is being applied in the resulting set. I don't think this change is needed. 

List doesn't need to imply that it is exhaustive, at least it doesn't to my reading. The most important thing about this method is that it returns root proxies. It make no guarantees whether those are actually valid (just unique FQDN).

internal/dag/httpproxy_processor.go
```diff
@@ -686,33 +686,36 @@ func (p *HTTPProxyProcessor) processHTTPProxyTCPProxy(validCond *contour_api_v1.
...
+// have their status updated with the corresponding error.
+func (p *HTTPProxyProcessor) listRootProxies() []*contour_api_v1.HTTPProxy {
+	// Ensure that a given FQDN is only referenced in a single HTTPProxy resource.
 	var valid []*contour_api_v1.HTTPProxy
-	fqdnHTTPProxies := make(map[string][]*contour_api_v1.HTTPProxy)
+	roots := make(map[string][]*contour_api_v1.HTTPProxy)
+
 	for _, proxy := range p.source.httpproxies {
 		if proxy.Spec.VirtualHost == nil {
 			valid = append(valid, proxy)
```
skriss, line 701: This adds non-root proxies to the return list too which contradicts the new name/comments.

skriss (changes requested):
WDYT about rearranging this a bit, so that this function is only responsible for computing the FQDN->[]HTTPProxy map, and then the validation step (does the HTTPProxy share an FQDN) can go inside `computeHTTPProxy` with all the other validations? I think that also makes this function easier to name :)

internal/dag/httpproxy_processor.go
jpeach: Ah I missed that; good point.

Issue #2984, opened 2020-10-05
use nonroot vs. nobody user


dprotaso:
related: https://github.com/knative-sandbox/net-contour/pull/292

`nobody` doesn't have a valid `/home` directory - so it is recommended to use `nonroot`

jpeach:
Thanks for reporting @dprotaso. This seems to be a problem introduced in knative by their image building process. I'll check back later for new information, but I think that any fix needs to be made in the knative tooling.

dprotaso:
gonna close this out since contour doesn't use distroless

PR #2985, opened 2020-10-05
use nonroot user vs nobody

dprotaso:
nobody does not have a valid home directory

Fixes: https://github.com/projectcontour/contour/issues/2984

### DIFF
--- a/examples/contour/02-job-certgen.yaml
+++ b/examples/contour/02-job-certgen.yaml
@@ -66,8 +66,8 @@ spec:
       serviceAccountName: contour-certgen
       securityContext:
         runAsNonRoot: true
-        runAsUser: 65534
-        runAsGroup: 65534
+        runAsUser: 65532
+        runAsGroup: 65532
   parallelism: 1
   completions: 1
   backoffLimit: 1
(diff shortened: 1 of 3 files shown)


mattmoor:
This is a function of us rebuilding contour with distroless.

Contour runs on scratch (which doesn't even have `/etc/passwd`), so I'm not sure this is really an upstream problem.

ofc I wouldn't object to an upstream fix, but I don't think there's anything technically wrong with what contour is currently doing.

dprotaso:
interesting I'm surprised `65534` in the examples work with scratch

jpeach:
> interesting I'm surprised `65534` in the examples works with scratch - when on k8s - i guess there's no validation

AFAICT, accessing the home directory is a distress thing?

dprotaso:
> AFAICT, accessing the home directory is a distroless thing?

No - distroless doesn't do anything special that requires a home directory.

The nobody user doesn't have a home directory by design - this seems to be [a linux convention](https://unix.stackexchange.com/questions/186568/what-is-nobody-user-and-group).

I'll close this out if you aren't using distroless as your base

PR #2986, opened 2020-10-05
internal/k8s: Ignore LastTransitionTime for HTTPProxy objects status equal check

### LINKED ISSUE #2978: Remove Description printer column
Since merging #2962 we don't need the `Description` field in the printer columns since now it's either "valid" or "invalid". The description is redundant and doesn't offer any more information and we can remove the description. 

image

### LINKED ISSUE #2979: Status updates are delayed with 100 proxies
Since #2962 was merged, I've been testing Contour for load and perf. With 100 services & 100 HTTPProxies in my cluster, status updates to proxies take anywhere from 15-30s. 

I see the same results on my local baremetal cluster as well as on a AWS EKS cluster.

**Repo steps:**
1. Create 100 services
2. Create 100 proxies
3. Wait for status to update & settle (will take a few minutes)
4. Edit a proxy and set the service to something invalid
5. The update will take 15-30 seconds to show on the object

stevesloka:
Change how the comparison of the status of the object of HTTPProxy objects by
ignoring the LastTransitionTime which is always updates on each DAG rebuild
regardless if the status of object changed or not.

Not ignoring this causes each status to be updated each time since the objects
are always different for each DAG rebuild.

Fixes #2979

### DIFF
--- a/internal/k8s/helpers.go
+++ b/internal/k8s/helpers.go
@@ -14,27 +14,39 @@
 package k8s
 
 import (
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	contour_api_v1 "github.com/projectcontour/contour/apis/projectcontour/v1"
 	"k8s.io/api/networking/v1beta1"
-	"k8s.io/apimachinery/pkg/api/equality"
 )
 
 // IsStatusEqual checks that two objects of supported Kubernetes types
 // have equivalent Status structs.
+//
 // Currently supports:
 // networking.k8s.io/ingress/v1beta1
+// projectcontour.io/v1
 func IsStatusEqual(objA, objB interface{}) bool {
 
 	switch a := objA.(type) {
 	case *v1beta1.Ingress:
 		switch b := objB.(type) {
 		case *v1beta1.Ingress:
-			return equality.Semantic.DeepEqual(a.Status, b.Status)
+			if cmp.Equal(a.Status, b.Status) {
+				return true
+			}
 		}
 	case *contour_api_v1.HTTPProxy:
 		switch b := objB.(type) {
 		case *contour_api_v1.HTTPProxy:
-			return equality.Semantic.DeepEqual(a.Status, b.Status)
+			// Compare the status of the object ignoring the LastTransitionTime which is always
+			// updated on each DAG rebuild regardless if the status of object changed or not.
+			// Not ignoring this causes each status to be updated each time since the objects
+			// are always different for each DAG rebuild (Issue #2979).
+			if cmp.Equal(a.Status, b.Status,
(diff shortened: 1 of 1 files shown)


jpeach:
FYI this tags the wrong issue

skriss (approved):
LGTM. I don't know if there was a particular reason for using the `equality` package to do the equality compare previously, but we use `cmp.Equal` in the Contour handler for doing equality of API objects, and the tests pass, so I don't see an issue with changing it.

jpeach (commented):
This reverts the switch to Kubernetes semantic equality from https://github.com/projectcontour/contour/pull/2853

I think that a better fix might be to add the `v1.DetailedCondition` type to the equality funcs that we use for this comparison. That retains the special handling that api-machinery has for existing Kubernetes types.

```Go
var equalities = conversion.Equalities{equality.Semantic}

equalities.AddFunc(func(a, b contourv1.DetailedCondition) bool{
        ...
})
```

jpeach:
> This reverts the switch to Kubernetes semantic equality from #2853
> 
> I think that a better fix might be to add the `v1.DetailedCondition` type to the equality funcs that we use for this comparison. That retains the special handling that api-machinery has for existing Kubernetes types.
> 
> ```go
> var equalities = conversion.Equalities{equality.Semantic}
> 
> equalities.AddFunc(func(a, b contourv1.DetailedCondition) bool{
>         ...
> })
> ```

That said, I don't see any practical issues with landing this and following up.

stevesloka:
But will adding the type still skip that field? That's the goal here, and why we went with cmp which is used in other places.

jpeach:
> But will adding the type still skip that field? That's the goal here, and why we went with cmp which is used in other places.

Looking at the way the interfaces work, you would add your own comparator that skips or flattens the relevant fields (e.g. the time comparison is flattened to UTC). `cmp` doesn't do the same thing as semantic equality AFAICT; trade-offs in both directions.

mattmoor:
FWIW in Knative we define a `VolatileTime` type, which we exclude from semantic equality, which we use exclusively (I think) for `LastTransitionTime`.

mattmoor:
We define our common condition type here: https://github.com/knative/pkg/blob/21cb070a8b050ec23100078e89df11d67444350c/apis/condition_types.go#L76

skriss:
We also have https://github.com/projectcontour/contour/blob/main/internal/contour/handler.go#L179-L183 which would need a custom comparer if we want to migrate that to `equality.Semantic.DeepEqual`.

@stevesloka do you want to merge this as-is for now (@jpeach was OK with that too) and do a follow-up to fully switch compare libraries?

stevesloka:
@skriss let's just land this as-is. I think we should chat about how to best approach the solution. I like Knavtive's approach that @mattmoor linked, it removes the extra need to understand that this field is different for other comparisons, but should chat about this a bit before making that change.

PR #2987, opened 2020-10-05
internal/contour: don't trigger DAG rebuild on ManagedFields change

skriss:
Ignores ManagedFields when checking if an object update should trigger
a DAG rebuild. Previously, ManagedFields would get updated when an
object's status changed, resulting in an extraneous DAG update.

Reviewers: I don't think this *needs* to go in 1.9 if we want to hold it back, since it doesn't seem to cause material perf issues.


stevesloka (approved):
/lgtm

PR #2990, opened 2020-10-05
internal: consistently use the controller-runtime informer cache

### LINKED ISSUE #2904: Sometimes the ingress status address is not updated
I created a new ingress and the Ingress Status address was not updated.
There is an error message in the contour log:

time="2020-09-11T20:40:57Z" level=error msg="unable to retrieve object for updating" context=StatusUpdateWriter error="ingresses.networking.k8s.io \"redacted\" not found" name=redacted namespace=redacted resource="networking.k8s.io/v1beta1, Resource=ingresses"

The status address should have been updated.

This might be related to PR #2864  

**Environment:**

- Contour version: 1.8.1
Server Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.13-eks-2ba888", GitCommit:"2ba888155c7f8093a1bc06e3336333fbdb27b3da", GitTreeState:"clean", BuildDate:"2020-07-17T18:48:53Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}

jpeach:
Prior to this change, the Kubernetes client API exposed methods to
create a informer factories as well as namespaced informer factories.
This encouraged other systems to use multiple factories and multiple
informers for each factory. The result of this is that objects could be
cached multiple times in the different informer factory caches, and that
the informer states could be subtly out of sync.

This change removes the internal APIs that expose informer factories.
There is now just one API to create an informer, and it is backed by the
controller-runtime cache API. This ensures that all the informers used
across Contour share a common cache and therefore a consistent view of
Kubernetes state.

This updates #2683.
This fixes #2904.


jpeach:
Need to test with knative integration tests tomorrow.

cmd/contour/serve.go
```diff
@@ -374,50 +368,80 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
 
+	// Inform on service-apis types if they are present.
 	if ctx.UseExperimentalServiceAPITypes {
-		// Check if the resource exists in the API server before setting up the informer.
-		if !clients.ResourcesExist(k8s.ServiceAPIResources()...) {
-			log.WithField("InformOnResources", "ExperimentalServiceAPITypes").Warnf("resources %v not found in api server", k8s.ServiceAPIResources())
-		} else {
-			informerSyncList.InformOnResources(clusterInformerFactory, dynamicHandler, k8s.ServiceAPIResources()...)
+		for _, r := range k8s.ServiceAPIResources() {
+			if !clients.ResourcesExist(k8s.ServiceAPIResources()...) {
```
skriss, line 384:
```suggestion
			if !clients.ResourcesExist(r) {
```

cmd/contour/serve.go
```diff
@@ -374,50 +368,80 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
-	g.Add(startInformer(clusterInformerFactory, log.WithField("context", "contourinformers")))
 
-	for ns, factory := range namespacedInformerFactories {
-		g.Add(startInformer(factory, log.WithField("context", "corenamespacedinformers").WithField("namespace", ns)))
-	}
+	// Register a task to start all the informers.
+	g.Add(func(stop <-chan struct{}) error {
+		log := log.WithField("context", "informers")
+
+		log.Info("started informers")
```
skriss, line 435: should either be "starting", or move to after they've actually been started

cmd/contour/ingressstatus.go
```diff
@@ -14,32 +14,35 @@
...
 )
 
-// loadBalancerStatusWriter manages the lifetime of IngressStatusUpdaters.
+// loadBalancerStatusWriter orchestrates LoadBalancer address status
+// updates for HTTPProxy and Infress objects.
 //
 // The theory of operation of the loadBalancerStatusWriter is as follows:
+//
 // 1. On startup the loadBalancerStatusWriter waits to be elected leader.
 // 2. Once elected leader, the loadBalancerStatusWriter waits to receive a
```
skriss, line 35: it'd be helpful to also document that once elected leader, it creates a `StatusAddressUpdater` which will handle writing LBStatus for future Ingress/HTTPProxy objects.

cmd/contour/serve.go
```diff
@@ -215,22 +213,18 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	}
 
 	if rootNamespaces := ctx.proxyRootNamespaces(); len(rootNamespaces) > 0 {
+		informerNamespaces = append(informerNamespaces, rootNamespaces...)
+
 		// Add the FallbackCertificateNamespace to the root-namespaces if not already
```
skriss, line 218:
```suggestion
		// Add the FallbackCertificateNamespace to informerNamespaces if not already
```

skriss (changes requested):
Handful of comments but this looks pretty good and sets us up for using the controller-runtime cache in place of our hand-rolled one. I do wonder if there are additional places where we need to wait for the cache to sync during startup (AFAIK right now we only do it before starting the xds server), but i'll need to look some more to see where that might be.

cmd/contour/ingressstatus.go
jpeach: I added this info, though in a slightly different place.

jpeach:
Seeing this error a couple of times when I run the Knative integration tests:
```
time="2020-10-07T08:39:31Z" level=error msg="unable to update status" context=StatusUpdateHandler error="Operation cannot be fulfilled on httpproxies.projectcontour.io \"ingress-conformance-0-dispatch-ede9da76e278bca44bfaa9dee298cf80\": StorageError: invalid object, Code: 4, Key: /registry/projectcontour.io/httpproxies/serving-tests/ingress-conformance-0-dispatch-ede9da76e278bca44bfaa9dee298cf80, ResourceVersion: 0, AdditionalErrorMsg: Precondition failed: UID in precondition: c53843a2-7db7-4343-97fa-c1dd1dd47380, UID in object meta: " name=ingress-conformance-0-dispatch-ede9da76e278bca44bfaa9dee298cf80 namespace=serving-tests
```

jpeach:
@skriss Could you please take another look. I ran into Go module hell (which is why I had to update the indirect `go-logr/logr` dependency) with knative, so going to have to table that stuff for a while. I think I've addressed all your feedback, and I added conflict retries to take care of the conflict error I saw logged.

cmd/contour/ingressstatus.go
```diff
@@ -62,62 +65,76 @@ func (isw *loadBalancerStatusWriter) Start(stop <-chan struct{}) error {
...
+			}
+
+			return log
+		}(),
+		IngressClass:  isw.ingressClass,
+		StatusUpdater: isw.statusUpdater,
+		Converter:     isw.Converter,
+	}
+
+	// Create new informer for the new LoadBalancerStatus. The
```
skriss, line 83: nit: this comment is slightly stale, could be clarified since we're no longer creating a new informer every time the LB address changes.

internal/k8s/status.go
```diff
@@ -60,19 +62,71 @@ func (m StatusMutatorFunc) Mutate(old interface{}) interface{} {
...
+	if err != nil {
+		suh.Log.WithError(err).
+			WithField("name", upd.NamespacedName.Name).
+			WithField("namespace", upd.NamespacedName.Namespace).
+			WithField("kind", gvk).
+			Error("failed to allocate template object")
+		return
+	}
+
+	if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
skriss, line 94: this looks reasonable to me. Were the conflicts due to Contour writing both conditions updates and LB address updates ~concurrently?

cmd/contour/serve.go
```diff
@@ -374,50 +369,80 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+		if err := clients.StartInformers(stop); err != nil {
+			log.WithError(err).Error("failed to start informers")
+		}
+
+		<-stop
+		return nil
+	})
 
 	// Register our event handler with the workgroup.
 	g.Add(eventHandler.Start())
```
skriss, line 448: I *think* that we should probably be waiting for the informer caches to sync in here before starting to process events to avoid any race conditions with the caches populating, but I'd have to experiment a bit to confirm if needed/not problematic.

cmd/contour/serve.go
```diff
@@ -503,7 +527,10 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	g.Add(lbsw.Start)
```
skriss, line 527: similar to the main contour event handler, I *think* that we should probably be waiting for the informer caches to sync in here before starting to process events to avoid any race conditions with the caches populating, but I'd have to experiment a bit to confirm if needed/not problematic.

skriss (changes requested):
changes look pretty good to me. I added thoughts on where we might want to wait for caches to sync before processing events, but as I noted, I'd have to experiment to confirm if that's actually needed, and doesn't break anything.

cmd/contour/ingressstatus.go
jpeach: The comment doesn't say that it is creating a new informer every time?

internal/k8s/status.go
jpeach: I wasn't able to debug it. Rebasing onto the Kubernetes 1.19 Go client broke knative with module dependencies :(

cmd/contour/serve.go
jpeach: Can you elaborate why you think this is needed? This change keeps the current semantics of not starting the xDS server until the informer cache is present. That's not awesome, but it's not different than what we had.

cmd/contour/serve.go
jpeach: ???

cmd/contour/ingressstatus.go
skriss: it says "for the new LoadBalancerStatus", which came from the old code where when a new LB address was received, a new informer would be created. Can we just say "Create informers" or something similar that doesn't reference the old code structure?

jpeach: Ah I see ... fixed.

cmd/contour/serve.go
jpeach: https://github.com/projectcontour/contour/issues/1280

skriss: Typically in Kubernetes controllers, you would wait for the informer caches to sync (via WaitForCacheSync) before starting to process events from those informers. I think the reason for this is that if you don't, then you may get not-found errors if you try to look up other items from the informer cache as part of your event handler logic.

One scenario I'm thinking of is in the load balancer address update code, which handles an event for the load balancer service, then tries to look up all ingresses/proxies from the cache so it can write addresses to them. if the ingress/informer caches haven't fully synced yet, then we could get an incomplete list back and not update all of them as expected. 

Practically it may or may not be an issue for us, I'm not sure. If we're not seeing issues, then we can defer/ignore.

jpeach: > One scenario I'm thinking of is in the load balancer address update code, which handles an event for the load balancer service, then tries to look up all ingresses/proxies from the cache so it can write addresses to them. if the ingress/informer caches haven't fully synced yet, then we could get an incomplete list back and not update all of them as expected.

I don't think this is a problem. The List is for updating objects that won't get passed to `OnAdd`. If the cache isn't synced when the List happens, then the `OnAdd` would work.

stevesloka: The current code does wait for the caches to be synced, but doesn't wait for the first dag rebuild to happen before starting the xds server. Need to make sure that behavior is still kept. Contour needs to wait for the caches to be synced before starting things up. The 2nd bit is still an open issue that needs solved.

cmd/contour/serve.go
```diff
@@ -511,24 +538,31 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 			Converter: converter,
 			Logger:    log.WithField("context", "serviceStatusLoadBalancerWatcher"),
 		}
-		factory := clients.NewInformerFactoryForNamespace(ctx.EnvoyServiceNamespace)
-		informerSyncList.InformOnResources(factory, dynamicServiceHandler, k8s.ServicesResources()...)
 
-		g.Add(startInformer(factory, log.WithField("context", "serviceStatusLoadBalancerWatcher")))
+		for _, r := range k8s.ServicesResources() {
```
stevesloka, line 542: Might help for readability to move this into a method to clean up the serve method.

cmd/contour/serve.go
```diff
@@ -372,50 +367,80 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+			inf.AddEventHandler(dynamicHandler)
+		} else {
+			inf.AddEventHandler(k8s.NewNamespaceFilter(informerNamespaces, dynamicHandler))
+		}
 	}
 
-	informerSyncList.InformOnResources(clusterInformerFactory,
-		&k8s.DynamicClientHandler{
+	// Inform on endpoints.
+	for _, r := range k8s.EndpointsResources() {
```
stevesloka, line 411: Might help for readability to move this into a method to clean up the serve method.

cmd/contour/serve.go
```diff
@@ -372,50 +367,80 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
 		}
 	}
 
-	// TODO(youngnick): Move this logic out to internal/k8s/informers.go somehow.
-	// Add informers for each root namespace
-	for _, factory := range namespacedInformerFactories {
-		informerSyncList.InformOnResources(factory, dynamicHandler, k8s.SecretsResources()...)
-	}
+	// Inform on secrets, filtering by root namespaces.
+	for _, r := range k8s.SecretsResources() {
```
stevesloka, line 397: Might help for readability to move this into a method to clean up the serve method.

cmd/contour/serve.go
```diff
@@ -372,50 +367,80 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+	}
 
+	// Inform on service-apis types if they are present.
 	if ctx.UseExperimentalServiceAPITypes {
-		// Check if the resource exists in the API server before setting up the informer.
-		if !clients.ResourcesExist(k8s.ServiceAPIResources()...) {
-			log.WithField("InformOnResources", "ExperimentalServiceAPITypes").Warnf("resources %v not found in api server", k8s.ServiceAPIResources())
-		} else {
-			informerSyncList.InformOnResources(clusterInformerFactory, dynamicHandler, k8s.ServiceAPIResources()...)
+		for _, r := range k8s.ServiceAPIResources() {
```
stevesloka, line 382: Might help for readability to move this into a method to clean up the serve method.

cmd/contour/serve.go
```diff
@@ -511,24 +538,31 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
 	}
 
 	g.Add(func(stop <-chan struct{}) error {
 		log := log.WithField("context", "xds")
 
 		log.Printf("waiting for informer caches to sync")
-		if err := informerSyncList.WaitForSync(stop); err != nil {
-			return err
+		if !clients.WaitForCacheSync(stop) {
+			return errors.New("informer cache failed to synx")
```
stevesloka, line 565: s/synx/sync

cmd/contour/serve.go
```diff
@@ -372,50 +367,80 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+	for _, r := range k8s.SecretsResources() {
+		inf, err := clients.InformerForResource(r)
+		if err != nil {
+			log.WithError(err).WithField("resource", r).Fatal("failed to create informer")
+		}
 
-	// If root namespaces are not defined, then add the informer for all namespaces
-	if len(namespacedInformerFactories) == 0 {
-		informerSyncList.InformOnResources(clusterInformerFactory, dynamicHandler, k8s.SecretsResources()...)
+		if len(informerNamespaces) == 0 {
```
stevesloka, line 403: Add the comment back describing this if statement.

cmd/contour/serve.go
```diff
@@ -501,7 +525,10 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	g.Add(lbsw.Start)
 
 	// Register an informer to watch envoy's service if we haven't been given static details.
-	if ctx.IngressStatusAddress == "" {
+	if ctx.IngressStatusAddress != "" {
+		log.WithField("loadbalancer-address", ctx.IngressStatusAddress).Info("Using supplied information for Ingress status")
```
stevesloka, line 529: s/loadbalancer-address/ingress-status-address to keep consistent with the config name.

stevesloka (changes requested):

cmd/contour/serve.go
jpeach: This field was present in the original code (the log line must moved). We should keep this field name because it correlates with the log message in `serve.go`.

cmd/contour/serve.go
jpeach: Fixed.

cmd/contour/serve.go
stevesloka: I don't agree, Contour should be consistent with its logs and matching the same name against what it's logging is what we should strive to do.

jpeach: This is consistent. You need to think about logs from the perspective of someone trying to triage a problem in production. That person doesn't know what the internal names are in the code, but can correlate context tags if we use them consistently.

When we use different tag names, we make life harder for operators. If we apply your suggestion, then there are different context tags for setting the address in config vs. applying the address to an object. That makes is unnecessarily hard for operators to understand what Contour is doing with the address.

At any rate, this code has not changed in this PR.

This PR:
```
cmd/contour/ingressstatus.go:                   isw.log.WithField("loadbalancer-address", lbAddress(lbs)).
cmd/contour/serve.go:           log.WithField("loadbalancer-address", ctx.IngressStatusAddress).Info("Using supplied information for Ingress status")
```

main:
```
cmd/contour/ingressstatus.go:                   isw.log.WithField("loadbalancer-address", lbAddress(lbs)).Info("received a new address for status.loadBalancer")
cmd/contour/serve.go:           log.WithField("loadbalancer-address", ctx.IngressStatusAddress).Info("Using supplied information for Ingress status")
```

cmd/contour/serve.go
jpeach: IIUC, we all agree that this PR doesn't change the current Contour semantics (even though those semantics leave something to be desired)? If so, then we can resolve this thread.

cmd/contour/serve.go
jpeach: Yeh, that's a different issue though.

skriss: filed #3020

cmd/contour/serve.go
skriss: filed #3020

cmd/contour/serve.go
stevesloka: Right, to be consistent we should use the name that operators use to configure the address. When specifying an address they set `ingress-status-address`, not `loadbalancer-address` which is what is log is referencing.

stevesloka: If you don't want to change the the log line @jpeach I'll do a follow up PR to make it consistent.

stevesloka (approved):

cmd/contour/serve.go
jpeach: This patch doesn't change the existing code. I think that if we want to make that change, it would be best to have a separate issue.

Issue #2818, opened 2020-08-20
Create universal schema for Contour types


ffahri:
Currently logging needs for the object types, kinds etc.

``` 
switch obj := obj.(type) {
		case *projectcontour.HTTPProxy, *projectcontour.TLSCertificateDelegation:
``` 

With the https://github.com/projectcontour/contour/pull/2808 PR schema method is added. However we need to add all Contour schemes into that.

```
scheme.Scheme.ObjectKinds(obj.(runtime.Object))
```

Create a universal schema under /k8s that way we don't need to manage objects manually.

jpeach:
2 suggestions for this:
1. Add `func k8s.NewDefaultScheme()*runtime.Scheme` that returns a new, fully populated scheme
2. Add `var DefaultScheme *runtime.Scheme` as a global that is set in a package `init()`.

Probably either of these approaches is fine. I expect that a default global scheme will be useful for the helper APIs though.

PR #2992, opened 2020-10-06
internal/k8s: add a Contour scheme constructor

jpeach:
Add a constructor to generate a scheme that includes all the
Contour API types.

This updates #2818.


jpeach:
This is needed for https://github.com/projectcontour/contour/pull/2990, because the controller-runtime informers want a scheme to do the conversion internally.

skriss (approved):

PR #2996, opened 2020-10-06
internal/featuretests: fix no-op SDS test

### LINKED ISSUE #2782: tests: TestSDSShouldNotIncrementVersionNumberForUnrelatedSecret is a no-op
Currently, the test TestSDSShouldNotIncrementVersionNumberForUnrelatedSecret is a no-op, as it checks the value of the `VersionInfo` field of an envoy `DiscoveryResponse` struct, which is explicitly ignored by `internal/assert.go:assert.Equal()`.

Issue covers either rewriting the test to actually test something, or removing it.

skriss:
Fixes test TestSDSShouldNotIncrementVersionNumberForUnrelatedSecret
that was a no-op because the fields it was trying to validate were
not being compared by the equality function being used.

Closes #2782


stevesloka (approved):
/lgtm

PR #3001, opened 2020-10-06
Update quickstart redirect to v1.9.0

stevesloka:


jpeach:
Why did the API reference change too?

skriss:
Yeah let's remove that file from this PR (@jpeach FYI there's an iteration over a map somewhere in the API ref docs generator, so the output does not guarantee ordering).

skriss (approved):

PR #3005, opened 2020-10-07
cmd/contour: remove deprecated request-timeout field from config file

### LINKED ISSUE #2690: move serveContext.RequestTimeout into TimeoutConfig and drop old field

skriss:
Removes the request-timeout field from the root of the config file. This
field has been replaced by timeouts.request-timeout, and the original
field has been deprectated for several releases now.

Closes #2690.


stevesloka (approved):
/lgtm

Issue #2777, opened 2020-08-07
Deprecate & remove TLS 1.1


dprotaso:
Reading the docs https://projectcontour.io/docs/v1.7.0/httpproxy/ - TLS section it says

> The TLS Minimum Protocol Version a vhost should negotiate can be specified by setting the spec.virtualhost.tls.minimumProtocolVersion:
> 1.3
> 1.2
> *1.1 (Default)*

I thought TLS 1.1 was deprecated/insecure so I’m surprised it’s the default

More details: https://hacks.mozilla.org/2020/02/its-the-boot-for-tls-1-0-and-tls-1-1/

I guess it would make sense to immediately bump the default to 1.2 and to have a path to remove 1.1

dprotaso:
This is a handy tool - https://ssl-config.mozilla.org/

jpeach:
xref https://github.com/projectcontour/contour/issues/2401

I don't think we should remove this (without a sufficiently log deprecation time), but we should add configuration and flip the defaults.

youngnick:
The 1.1 minimum version is the default for compatibility reasons, currently. I agree that we should change the default to 1.2, and allow 1.1 for a while, with a log indicating that you should stop using it.

This comes under the heading of "people use the proxy as a fixup point", sadly.

dprotaso:
> 
> 
> xref #2401
> 
> I don't think we should remove this (without a sufficiently log deprecation time), but we should add configuration and flip the defaults.


yeah I agree.

youngnick:
Okay, so todo here:
- [ ] Change the default for TLS version to 1.2, with a release note mentioning the current guidance, and noting that we consider support for TLS 1.1 in Contour deprecated and risky, and that it will be removed eventually.
- [ ] Add a startup warning log that using 1.1 is not a good idea and you should use 1.2 if you can.

Wait to see how much community comment there is after this, and set the removal period for TLS 1.1 based on that.

youngnick:
Also, we should note that you can use `MinimumProtocolVersion` to stop the use of 1.1 entirely without it being disabled in Contour itself. Possibly we should default this to 1.2 at some point, as a "removal" step, rather than totally removing the functionality.

mattmoor:
@youngnick does setting that change the defaulting to use that minimum version, or would changing that cause previously valid things to be rejected?

youngnick:
[...]

PR #3007, opened 2020-10-07
internal: update HTTPProxy & Ingress default min TLS version to 1.2

skriss:
Changes the default minimum TLS protocol version from 1.1 to 1.2 for
HTTPProxies and Ingresses that don't specify it, since 1.1 is now
end-of-life. 1.1 can still be used, but it must be explicitly configured.

Updates #2777

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -179,14 +179,6 @@ func validateCRDs(dynamicClient dynamic.Interface, log logrus.FieldLogger) {
 
 // doServe runs the contour serve subcommand.
 func doServe(log logrus.FieldLogger, ctx *serveContext) error {
-	// log a warning if the Contour config file doesn't already disable TLS 1.1 and 1.0.
-	if annotation.MinTLSVersion(ctx.TLSConfig.MinimumProtocolVersion) <= envoy_api_v2_auth.TlsParameters_TLSv1_1 {
-		log.Warn("In Contour 1.10, TLS 1.1 will be disabled by default. HTTPProxies with no Spec.VirtualHost.TLS.MinimumProtocolVersion" +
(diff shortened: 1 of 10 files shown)


cmd/contour/serve.go
```diff
@@ -267,6 +259,9 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		return fmt.Errorf("error parsing request timeout: %w", err)
 	}
 
+	// TLS 1.1 is still allowed, but it's not the default for HTTPProxies/Ingresses.
+	minimumTLSVersion := annotation.MinTLSVersion(ctx.TLSConfig.MinimumProtocolVersion, envoy_api_v2_auth.TlsParameters_TLSv1_1)
```
stevesloka, line 263: Should this set the min version to TLSV1_2?

stevesloka (changes requested):

cmd/contour/serve.go
skriss: We/I have gone back and forth on this a bit in https://github.com/projectcontour/contour/issues/2777, but I was currently planning to leave this value at 1.1 for at least another release before changing its default, to give users some more time to address this situation if needed. Leaving this setting at 1.1 allows proxies that explicitly specify 1.1 to continue working as-is.

stevesloka: Reading this I thought it was setting 1.1 to the default. Where does 1.2 become the default? Somewhere else in the builder?

skriss: Yeah it's in the processors (e.g. https://github.com/projectcontour/contour/pull/3007/files#diff-203dc8cde1099ac3666b0b37414f448aR184-R185).

This setting acts as a global minimum, so if any proxy or ingress has a min TLS version lower than this value, then it will be (silently) upgraded to this global minimum when programming Envoy. I can add some more comments to explain that behavior.

stevesloka (approved):
/lgtm

PR #3012, opened 2020-10-08
cmd/contour: log warning for HTTPProxies/Ingresses using TLS 1.1

skriss:
Logs warnings on server startup for HTTPProxies and Ingresses using
a TLS minimum protocol version of 1.1 without also having the config
file's tls.minimum-protocol-version set to 1.1, since the default
value for the config file field will change to 1.2 in an upcoming
release.

Updates #2777.


stevesloka (approved):
/lgtm

PR #3016, opened 2020-10-09
internal/xds: Refactor xDS Server into a v2 package

stevesloka:
Refactors the xDS Server into a v2 package allowing for
a v3 server to be created seperate from the v2 instance.

Updates #1898

Signed-off-by: Steve Sloka <slokas@vmware.com>

PR #3017, opened 2020-10-09
internal/xds: xDS v3 server implementation

stevesloka:
Follows #3016 by adding an xDS v3 server.  Follow-up PR will introduce a config option to specify the xDS version wanting to serve. 

Updates #1898

### DIFF
--- a/internal/xds/resource.go
+++ b/internal/xds/resource.go
@@ -13,7 +13,11 @@
 
 package xds
 
-import "github.com/golang/protobuf/proto"
+import (
+	"sync/atomic"
+
+	"github.com/golang/protobuf/proto"
+)
 
 // Resource represents a source of proto.Messages that can be registered
 // for interest.
@@ -29,4 +33,11 @@ type Resource interface {
 
 	// TypeURL returns the typeURL of messages returned from Values.
 	TypeURL() string
+}
+
+// Counter holds an atomically incrementing counter.
+type Counter uint64
+
+func (c *Counter) Next() uint64 {
+	return atomic.AddUint64((*uint64)(c), 1)
 }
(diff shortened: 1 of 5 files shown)


jpeach (commented):
Do we really need to duplicate all the code here? The only think that is really different is the types of the discovery request and response. Can we take an approach where we only have to duplicate or abstract the guts of `(*contourServer)stream()`?

stevesloka:
I pulled out some more bits that were duplicated. Please take a look @jpeach, @skriss.

skriss:
It seems like in order to reuse much more code here, we'd need to define some interfaces and wrapper structs to adapt the v2 & v3 discovery request and response structs to common interfaces that the `stream` method could then interact with.

Given that we're looking to replace this all with the go-control-plane xds server though, I'm not sure it's really worth spending that effort.

internal/xds/resource.go
```diff
@@ -30,3 +34,10 @@ type Resource interface {
 	// TypeURL returns the typeURL of messages returned from Values.
 	TypeURL() string
 }
+
+// counter holds an atomically incrementing counter.
```
skriss, line 38:
```suggestion
// Counter holds an atomically incrementing counter.
```

internal/xds/v2/contour.go
```diff
@@ -174,6 +165,7 @@ func (s *contourServer) stream(st grpcStream) error {
 			return done(log, ctx.Err())
 		}
 	}
+
```
skriss, line 168: nit: rm blank line

skriss (approved):
Couple nits but overall I'm fine with this as-is, given that (1) we're going to drop v2 support pretty soon since it will be removed from Envoy early next year; (2) we're in the process of moving exclusively to the go-control-plane impl; and (3) upstream Envoy will not be revving a major version of the API every year going fwd.

PR #3018, opened 2020-10-09
xDS v3 config file option

stevesloka:
Adds a `server.xds-server-version` configuration item which allows users to define which version of xDS to use when serving xDS resources to Envoy. 

Updates #1898


cmd/contour/serve.go
```diff
@@ -677,7 +695,7 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 			log = log.WithField("insecure", true)
 		}
 
-		log.Info("started xDS server")
+		log.Infof("started xDS server type: %q version: %q", ctx.XDSServerType, xDSServerVersion)
```
skriss, line 698: Maybe just add the server type and version as structured fields? e.g. 

```
log.WithField("xdsServerType", ctx.XDSServerType).WithField("xdsVersion", xDSServerVersion).Info("started xDS server")
```

skriss (changes requested):
Need to add to `site/docs/main/configuration.md` as well.

cmd/contour/serve.go
stevesloka: I modeled after the PR to refactor config out to the `pkg` package @skriss.

skriss: I still prefer structured log fields but 🤷‍♂️

skriss (approved):
LGTM

youngnick (commented):
LGTM

PR #3019, opened 2020-10-11
Implements #2789 - Debug Container

### LINKED ISSUE #2789: Provide debug variants of contour container images
As already mentioned (https://github.com/projectcontour/contour/issues/2050#issuecomment-672570646) it would be nice, if you build debug variants of your container images containing debug tools like busybox or another minimal linux distro. This would help people troubleshoot (eg. look at the configs in the empty dir, curl the envoy healthcheck API, etc.)

For inspiration, look at the guys behind fluent-bit:
* Normal image: https://github.com/fluent/fluent-bit/blob/master/Dockerfile
* Debug image: https://github.com/fluent/fluent-bit/blob/master/Dockerfile.debug 

**Update:**
We only need to talk about the contour container images. Envoy images seems to be ubuntu based.

leowmjw:
Fixes #2789 

First cut implementation; any suggestions to make it better?

Where do you recommend I should put the docs for this?  It appears in the "make help" output already.

Any other common tools are needed when debugging contour? 

Any other common scenarios usually encountered that I should add to the examples?

### DIFF
--- a/Dockerfile.debug
+++ b/Dockerfile.debug
@@ -1,0 +1,42 @@
+FROM golang:1.15.2 AS build
+WORKDIR /contour
+
+ENV GOPROXY=https://proxy.golang.org
+COPY go.mod go.sum /contour/
+RUN go mod download
+
+COPY cmd cmd
+COPY internal internal
+COPY apis apis
+COPY Makefile Makefile
+
+ARG BUILD_BRANCH
+ARG BUILD_SHA
+ARG BUILD_VERSION
+
+RUN make install \
+	    CGO_ENABLED=0 \
+	    GOOS=linux \
+	    BUILD_VERSION=${BUILD_VERSION} \
+	    BUILD_SHA=${BUILD_SHA} \
+	    BUILD_BRANCH=${BUILD_BRANCH}
+
+FROM debian:stable-slim
+RUN apt-get update && \
+    apt-get install --no-install-recommends -y \
+      curl \
+      iproute2 \
+      iputils-ping \
+      dnsutils \
+      netcat \
+      tcpdump \
+      tcptraceroute traceroute \
+      net-tools \
+      libc6-dbg gdb \
+      elvis-tiny \
+      lsof \
+      procps \
+      busybox && \
+    alias vim=elvis-tiny && \
+    rm -rf /var/lib/apt/lists/*
+COPY --from=build /go/bin/contour /bin/contour
--- a/Makefile
+++ b/Makefile
@@ -97,6 +97,16 @@ container: ## Build the Contour container image
 		$(shell pwd) \
 		--tag $(IMAGE):$(VERSION)
 
+debug-container: ## Build the Contour debug container image
+	docker build \
+		-f Dockerfile.debug \
+		--build-arg "BUILD_VERSION=$(BUILD_VERSION)" \
+		--build-arg "BUILD_BRANCH=$(BUILD_BRANCH)" \
+		--build-arg "BUILD_SHA=$(BUILD_SHA)" \
+		$(DOCKER_BUILD_LABELS) \
+		$(shell pwd) \
+		--tag $(IMAGE)-debug:$(VERSION)
+
 push: ## Push the Contour container image to the Docker registry
 push: container
 	docker push $(IMAGE):$(VERSION)


stevesloka:
Thanks for the PR @leowmjw! Can you help describe a bit about what this PR does and why we need it? Are you looking for ways to debug your cluster networking? I'm just wondering if this is the right place for that sort of tool.

leowmjw:
@stevesloka This addresses #2789 which was asking for a debug tools with contour. Done as part of Hacktoberfest. 
Am a beginner in contour so going the issues with docs to follow; will tackle the ArgoCD one next.

If this is not required, feel free to close.

jpeach:
Is Debian stable the best choice for debug tooling? Would something faster-moving make more sense (personally I'm most familiar with Fedora)?

leowmjw:
@jpeach I think the debugging tools do not need to be bleeding edge; I choose the slimmest debian that I am familiar with.  I can definitely change the base image to Fedora if that is the preferred; which base image do you suggest?

stevesloka:
+1 to a smaller image

I prefer Debian over Fedora, but we'll all never agree. =)

jpeach:
> +1 to a smaller image
> 
> I prefer Debian over Fedora, but we'll all never agree. =)

Is it worth using the same base as envoy? FWIW, I can live with whatever you choose :)

youngnick (changes requested):
Thanks for this PR @leowmjw! I like the overall approach, and don't mind debian as the distro. I don't really think it matters too much, as long as there *is* a distro.

We're got some docs restructuring queued up, but for now if you could write up a small guide and put it in `/site/_guides`, I think that will at least ensure that we have some instructions captured, ready to move around when we do that.

stevesloka:
Envoy uses `ubuntu` as it's base: https://github.com/envoyproxy/envoy/blob/master/ci/Dockerfile-envoy#L1-L2

I think I'm fine with landing this unless folks think they can enable the alpha features in ephemeral containers which ultimately I think would be the best solution (https://kubernetes.io/docs/tasks/debug-application-cluster/debug-running-pod/#ephemeral-container). 

In any case as long as the deploy scripts can be updated to build this, I don't think it's a big deal to ship as long as the docs clearly state this is only for debugging. =)

Dockerfile.debug
```diff
@@ -0,0 +1,42 @@
+FROM golang:1.15.2 AS build
+WORKDIR /contour
+
+ENV GOPROXY=https://proxy.golang.org
+COPY go.mod go.sum /contour/
+RUN go mod download
+
+COPY cmd cmd
+COPY internal internal
+COPY apis apis
```
skriss, line 10: need to add the `pkg` dir as well now - see recent commit https://github.com/projectcontour/contour/commit/6668c5913cf73a32293edac813419771df5aaf53#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557

Dockerfile.debug
```diff
@@ -0,0 +1,42 @@
+FROM golang:1.15.2 AS build
```
skriss, line 1: Since we now have multi-arch support for our container image, it'd be nice to support that here as well - see recent commit https://github.com/projectcontour/contour/commit/baa89ee5b856b4ba9603caf7fa346afc6eef5b7f#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557

Makefile
```diff
@@ -97,6 +97,16 @@ container: ## Build the Contour container image
 		$(shell pwd) \
 		--tag $(IMAGE):$(VERSION)
 
+debug-container: ## Build the Contour debug container image
```
skriss, line 100: I assume we'd want to actually push these images to Docker Hub? If so, need to add `make` target and GitHub Actions code for that.

skriss (changes requested):
I'm fine with this if it's useful for users to have these debug variants. A few updates needed!

jpeach:
> Envoy uses `ubuntu` as it's base: https://github.com/envoyproxy/envoy/blob/master/ci/Dockerfile-envoy#L1-L2

I think ubuntu is a better choice - consistent with Envoy, updated more frequently than Debian stable, wider user base

youngnick (changes requested):
Agree with @skriss suggestions, and let's change the base to `ubuntu`. Sorry about the back-and-forth @leowmjw.

skriss:
@leowmjw are you still interested in working on this?

skriss:
Closing this out as stale, feel free to reopen or open a new PR if you want to pick this up again!

PR #3023, opened 2020-10-13
pkg/config: export a package for config parameters

### LINKED ISSUE #2621: Separate the Contour config file code into a new package

The Contour config file definition is embedded in `serveContext`, which contains both public definitions for the config file, internal definitions used by command-line flags, and other internals. It's confusing trying to separate out how the various fields are used and likely to become more so as we expand the use of the config file.

Let's break this struct and create an `internal/config` package to formalize the configuration file and bring all the defaults into one place. This should make it easier to understand the config flow and easier to add new configurations.

jpeach:
Move the configuration file handling into a separate package. This
cleanly separates the configuration file from the contour internal
server details, making it clearer which struct fields are internal
and which are configuration (and therefore have compatibility concerns).

The configuration separation allows the configuration to be validated in
one place prior to setting up the Contour controller. This change modifies
the server start up to do that. This guarantees that the configuration
is which lets us remove subsequent error handling.

The new configuration package is exported so that other projects in the 
Contour ecosystem can consume it. This means that the API here should
be considered (reasonably, but not absolutely) stable.

This fixes #2621.


jpeach:
@stevesloka @skriss I'm pushing out this draft early to give you an idea of what I'd like to do with the config file. There is a new exported package `pkg/config` with nomenclature oriented around `xxxParameters`. The basic operations are `Defaults()`, and `Parse()` and there's some initial validation support so that by the time the `contour serve` command starts up we know the config is valid.

LMK.

stevesloka (commented):
There's a ton more code changed here than that's related to just exporting the config. Is there a different branch pulled in or something?

jpeach:
> There's a ton more code changed here than that's related to just exporting the config. Is there a different branch pulled in or something?

Yep, it's currently on top on another PR because both make changes to `serve.go` and I didn't want to deal with merge conflicts. Just the second commit is relevant.

jpeach:
@stevesloka @skriss This is ready for review. I expect we will need a couple of passes :)

youngnick (approved):
LGTM

PR #3026, opened 2020-10-13
internal/dag: default the authorization response timeout

### LINKED ISSUE #3025: Auth: ExtensionService timeout policy doesn't do anything

The `ExtensionService` timeout policy field isn't wired up to anything. Since this CRD generates an Envoy cluster, I don't see a way to implement the idle timeout (that's only on the HTTP route, not the cluster AFAICT), but it should be possible to use the response timeout to default the authorization response timeout.

jpeach:
If the authorization spec on the virtual host doesn't set the response
timeout, default it from the corresponding `ExtensionService`.

This fixes #3025.

### DIFF
--- a/internal/dag/httpproxy_processor.go
+++ b/internal/dag/httpproxy_processor.go
@@ -265,11 +265,16 @@ func (p *HTTPProxyProcessor) computeHTTPProxy(proxy *contour_api_v1.HTTPProxy) {
 
 				timeout, err := timeout.Parse(auth.ResponseTimeout)
 				if err != nil {
-					validCond.AddErrorf("AuthError", "AuthReponseTimeoutInvalid",
+					validCond.AddErrorf("AuthError", "AuthResponseTimeoutInvalid",
 						"Spec.Virtualhost.Authorization.ResponseTimeout is invalid: %s", err)
 					return
 				}
-				svhost.AuthorizationResponseTimeout = timeout
+
+				if timeout.UseDefault() {
+					svhost.AuthorizationResponseTimeout = ext.TimeoutPolicy.ResponseTimeout
+				} else {
+					svhost.AuthorizationResponseTimeout = timeout
+				}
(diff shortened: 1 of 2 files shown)


skriss (approved):
LGTM

internal/dag/httpproxy_processor.go
```diff
@@ -265,11 +265,16 @@ func (p *HTTPProxyProcessor) computeHTTPProxy(proxy *contour_api_v1.HTTPProxy) {
...
 				timeout, err := timeout.Parse(auth.ResponseTimeout)
 				if err != nil {
-					validCond.AddErrorf("AuthError", "AuthReponseTimeoutInvalid",
+					validCond.AddErrorf("AuthError", "AuthResponseTimeoutInvalid",
 						"Spec.Virtualhost.Authorization.ResponseTimeout is invalid: %s", err)
 					return
 				}
-				svhost.AuthorizationResponseTimeout = timeout
+
+				if timeout.UseDefault() {
```
stevesloka, line 273: Maybe collapse this into the pattern we've used before which is UseDefaultOrValue? Wondering if that might read better. Alternatively, a good place is to set the default when the extensionservice is created and have the default be there.

jpeach: Sorry I didn't see this comment when I merged :-(

Maybe something like:
```Go
func (s Setting) DurationOr(t time.Duration) time.Duration {
    if s.UseDefault() {
        return t
    }

    return s.val
}

```

One consideration is that the natural way to do this for a `timeout.Setting` is probably to specify a backup `time.Duration`, but here we have a backup `timeout.Setting`.

Maybe we could add something like:
```Go
package timeout

func FirstOf(settings ...Setting) Setting {
    for _, s := range settings {
        if !s.UseDefault() {
            return s
        }
    }

    return DefaultSetting()
}
```

PR #3028, opened 2020-10-14
site: add `lifecycle/needs-triage` label

### LINKED ISSUE #2981: Add a lifecycle/needs-triage issue label

I recently discovered that Github issue templates can add a default set of labels. This opens the possibility of automatically adding a `lifecycle/needs-triage` label to all new issues. This label would (IMHO) make it easier to keep track of which issues need to be triaged.

jpeach:
Add a `lifecycle/needs-triage` label and apply it on all new
issues opened with the issue templates.

This fixes #2981.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -2,7 +2,7 @@
 name: Bug report
 about: Tell us about a problem you are experiencing
 title: ''
-labels: kind/bug
+labels: kind/bug, lifecycle/needs-triage
 assignees: ''
 
 ---
--- a/.github/ISSUE_TEMPLATE/feature-enhancement-request.md
+++ b/.github/ISSUE_TEMPLATE/feature-enhancement-request.md
@@ -2,7 +2,7 @@
 name: Feature enhancement request
 about: Suggest an idea for this project
 title: ''
-labels: kind/feature
+labels: kind/feature, lifecycle/needs-triage
 assignees: ''
 
 ---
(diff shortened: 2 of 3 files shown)


stevesloka (approved):
/lgtm 

The interesting bit is I don't know if the labels are always getting added automagically. I've created some issues recently that don't have the `kind/feature` label added. /shrug

skriss (approved):
LGTM

skriss:
> The interesting bit is I don't know if the labels are always getting added automagically. I've created some issues recently that don't have the `kind/feature` label added. /shrug

If you go directly to https://github.com/projectcontour/contour/issues/new, no template is used so no labels are added, but if you [choose a template](https://github.com/projectcontour/contour/issues/new/choose) that seems to add the labels properly. Maybe that explains what you're seeing?

stevesloka:
Yup possibly, I'll have to pay more attention next time I open an issue. 

As an aside thinking about this more (I know it's already merged), should we remove the "bug" or "feature" label from the default until after it's been triaged? Seems like that output should be a result of the triage process.

jpeach:
> If you go directly to https://github.com/projectcontour/contour/issues/new, no template is used so no labels are added

Good find!

> As an aside thinking about this more (I know it's already merged), should we remove the "bug" or "feature" label from the default until after it's been triaged? Seems like that output should be a result of the triage process.

I think that when we do initial triage, we can modify labels however we think makes sense for the issue. Applying them up front isn't meant to be carved in stone; it's just an initial hint (and hopefully labor-saving). As we get more regular at triage, I'd expect that our process and expectations will solidify as we get experience with what works for us and what metadata we need. We can definitely revisit this, and should plan to make improvements.

PR #3030, opened 2020-10-14
.github/workflows: automatically tag new release images as 'latest'

### LINKED ISSUE #1844: Build and tag release action does not tag latest
Now that we're fully cut over to GitHub Actions for image building and the docker hub for storage we should fix the build action so it tags latest.

skriss:
Adds logic to release tag builds to detect if the tag being built
is the highest semantic version that's not a pre-release, and if
so, sets TAG_LATEST to true so the image will be tagged as 'latest'
in addition to the version-specific tag.

Closes #1844

### DIFF
--- a/hack/actions/build-and-push-release-images.sh
+++ b/hack/actions/build-and-push-release-images.sh
@@ -1,0 +1,57 @@
+#! /usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# This scripts sets TAG_LATEST based on whether the tag currently being
+# built is the highest semantic version tag that's not a pre-release
+# (alpha, beta, rc), and calls 'make push'.
+#
+# This script is intended to be run only for tag builds and will no-op
+# if GITHUB_REF is not in the format "refs/tags/<tag-name>".
+# Ref. https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables
+
+# Check if the current build is for a tag and exit early if not.
+REF_TYPE=$(echo "$GITHUB_REF" | cut -d / -f 2)
+if [[ "$REF_TYPE" != "tags" ]]; then
+    echo "REF_TYPE \"$REF_TYPE\" is not a tag, exiting"
+    exit 0
+fi
+
+# Get the current tag name.
+CURRENT_TAG=$(echo "$GITHUB_REF" | cut -d / -f 3)
+if [[ -z "$CURRENT_TAG" ]]; then
+    echo "Error getting current tag name from GITHUB_REF \"$GITHUB_REF\""
+    exit 1
+fi
+
+# Fetch all tags so we can check if the current tag
+# is the highest semver.
+git fetch --tags
+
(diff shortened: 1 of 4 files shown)


hack/actions/build-and-push-release-images.sh
```diff
@@ -0,0 +1,56 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# This scripts sets TAG_LATEST based on whether the tag currently being
+# built is the highest semantic version tag that's not a pre-release
+# (alpha, beta, rc), and calls 'make push'.
+#
+# This script is intended to be run only for tag builds and will no-op
+# if GITHUB_REF is not in the format "refs/tags/<tag-name>".
```
jpeach, line 12: Not sure how stable the link is, but it would be handy to link to the docs for GITHUB_REF

https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables

hack/actions/build-and-push-release-images.sh
```diff
@@ -0,0 +1,56 @@
...
+
+# Check if the current build is for a tag and exit early if not.
+REF_TYPE=$(echo "$GITHUB_REF" | cut -d / -f 2)
+if [[ "$REF_TYPE" != "tags" ]]; then
+    echo "REF_TYPE $REF_TYPE is not a tag, exiting."
```
jpeach, line 17:
```suggestion
    echo "REF_TYPE \"$REF_TYPE\" is not a tag, exiting"
```

hack/actions/build-and-push-release-images.sh
```diff
@@ -0,0 +1,56 @@
...
+    exit 0
+fi
+
+# Get the current tag name.
+CURRENT_TAG=$(echo "$GITHUB_REF" | cut -d / -f 3)
+if [[ -z "$CURRENT_TAG" ]]; then
+    echo "Error getting current tag name from GITHUB_REF $GITHUB_REF."
```
jpeach, line 24:
```suggestion
    echo "Error getting current tag name from GITHUB_REF \"$GITHUB_REF\""
```

jpeach (approved):
Nice!

stevesloka (commented):
Only thing from history that I know we ran into with automating this work was when we had other tags that got pushed for alpha, beta, etc. Pushing updates to those shouldn't move the `latest` tag. But I think you're sorting on that so maybe not an issue?

skriss:
> Only thing from history that I know we ran into with automating this work was when we had other tags that got pushed for alpha, beta, etc. Pushing updates to those shouldn't move the latest tag. But I think you're sorting on that so maybe not an issue?

Yep, tags that include alpha/beta/rc are being excluded from consideration as "latest" (FWIW we used a similar script in velero so it's been pretty-well tested).

This also handles the scenario where we release 1.9.0, then subsequently release 1.8.1. 1.9.0 would retain the `latest` tag, since it's the newest semver, even though it's not the most recent (by date) tag.

stevesloka (approved):
/lgtm

PR #3035, opened 2020-10-16
testsuite: use official ingress-conformance-echo image

### LINKED ISSUE #3034: Use official ingress-controller-conformance echo server

Currently the integration tests use `agervais/ingress-conformance-echo`, but we should switch to the official images.

xref https://github.com/kubernetes-sigs/ingress-controller-conformance/issues/11

jpeach:
Switch the integration test suite over to use the official
ingress-conformance-echo image. The echo server no longer emits the
value of the `TEST_ID` environment variable, but we can switch to using
`SERVICE_NAME` for the same purpose. Adding some policy helpers makes
this more consistent across test scripts.

This fixes #3034.

### DIFF
--- a/_integration/testsuite/fixtures/ingress-conformance-echo.yaml
+++ b/_integration/testsuite/fixtures/ingress-conformance-echo.yaml
@@ -12,6 +12,9 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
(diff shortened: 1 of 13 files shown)


youngnick (approved):
LGTM

Issue #3032, opened 2020-10-14
Support custom JSON fields for Envoy access logs


mike1808:
We want to be able to configure Contour to have custom JSON access log fields, such as `custom_field: %REQ(OUR-CUSTOM-HEADER)%`. 

We are thinking that this can be achieved by having a new property with the JSON fields mapping like for [DefaultField](https://godoc.org/github.com/projectcontour/contour/internal/envoy#DefaultFields) in the Contour config. For example:

```yaml
accesslog-format: json
json-fields:
  - "@timestamp"
  - custom_field
extra-json-fields:
  custom_field:  "%REQ(X-CUSTOM-HEADER)%",
```

We [@cloudfoundry/cf-for-k8s-networking](https://github.com/orgs/cloudfoundry/teams/cf-for-k8s-networking) are happy to contribute this feature to upstream.

xcrezd:
+1

jpeach:
Historically, new fields were added to the internal list, see #1734. I think that customizing the fields is a reasonable need, however IIUC there's always been a concern that directly exposing the envoy config is too risky because an invalid config would break the envoy config, and we don't have good ways to surface the error.

Maybe we can address this without directly exposing the envoy config, or maybe it's time to just accept the risk 🤷

mike1808:
@jpeach do you think restricting fields to only take custom headers names will increase the chances of accepting this feature? For example:

```json
json-fields:
  - "@timestamp"
  - custom_field
extra-json-fields-headers:
  custom_field:  "X-CUSTOM-HEADER"
```
 
So it won't be the exact Envoy config, and will help Contour to keep this API in future if Envoy decided to change it.

xcrezd:
This is the only thing that stops us to move from nginx-controller to contour

jpeach:
xref #1507

jpeach:
When we send an unsupported log token, the listener resource is rejected:
```
ERRO[0004] Error adding/updating listener(s) ingress_http: Not supported field in StreamInfo: JPEACH  code=13 connection=4 context=xds node_id=envoy-f2lvb node_version=v1.16.0 response_nonce=1 version_info=1
```

Neither Contour nor go-control-plane recover gracefully from a rejected update, so we want to be careful to only push valid configuration.

https://github.com/envoyproxy/envoy/issues/13620

jpeach:
> @jpeach do you think restricting fields to only take custom headers names will increase the chances of accepting this feature?

[...]

PR #3042, opened 2020-10-19
Add envoy-json-logging-custom-fields-design doc

mike1808:
Co-authored-by: Leah Hanson <lhanson@pivotal.io>

In continuation of #3032 we're making a design doc to discuss possible solutions for our problem.


design/envoy-json-logging-custom-fields-design.md
```diff
@@ -0,0 +1,146 @@
...
+
+* Should all the fields with obvious translations consistently either require parameterization or not?
+  - In the current implementation, `method` becomes `"method": "%REQ(:METHOD)%"`.
+  - When the new implementation adds a way to achieve the JSON config `"response_duration": "%RESPONSE_DURATION%"`, should the user write:
+    * `response_duration`, or
+    * `response_durationn=%RESPONSE_DURATION%`
+  - This is related to [#2523](https://github.com/projectcontour/contour/issues/2523)
+* Is one of the alternative approaches preferable?
+   - [User-defined Translation Table](#user-defined-translation-table) seems easier to implement because it was partially implemented in #3033 and because the configuration is more structured and easier to work with as there no need for additional string parsing.
+   - [Map Syntax](#map-syntax) is not backward-compatible, but does avoid string parsing.
```
jpeach, line 143: Map syntax seems clean, but we would need to introduce it in a new config field, then deprecate the old field after a number of releases. Would still need to define what happens to the built-in field mappings.

design/envoy-json-logging-custom-fields-design.md
```diff
@@ -0,0 +1,146 @@
...
+Both original and the alternative translation table high-level designs don't introduce breaking changes, but the map syntax alternative would be a breaking change.
+The current implementation skips unknown fields so any new custom fields will be ignored.
+
+## Open Issues
+
+* Should all the fields with obvious translations consistently either require parameterization or not?
+  - In the current implementation, `method` becomes `"method": "%REQ(:METHOD)%"`.
+  - When the new implementation adds a way to achieve the JSON config `"response_duration": "%RESPONSE_DURATION%"`, should the user write:
+    * `response_duration`, or
+    * `response_durationn=%RESPONSE_DURATION%`
```
jpeach, line 139: The built-in mappings still need to be supported. I would not want to force people using the existing field mappings to manually remake their configuration.

design/envoy-json-logging-custom-fields-design.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Edge Cases
+
+#### Field name is repeated
+
+If the field name is repeated the last entry in the configuration takes precedes and overwrites the previous value.
+
+#### Unknown field name in old format
+
+If the field name is not defined in the default fields transition map and the user doesn't specify the Envoy field format, then Contour will raise a validation error.
```
jpeach, line 61: See https://github.com/projectcontour/contour/pull/3023 for config file validation.

jpeach (approved):
This looks good to me. I'll give other maintainers a chance to review before I merge though.

youngnick (approved):
This proposal LGTM, I think it's a good balance between adding extra functionality and backward compatibility. Thanks for writing it up @mike1808 and @astrieanna!

design/envoy-json-logging-custom-fields-design.md
astrieanna: We agree that built-in mappings need to be supported. The question here is whether fields that could be added to those mappings should be.
* `method` will still work.
* should `response_duration` work?

If `response_duration` works, then the interface is consistent, but the implication is that the translation table will be used forever (despite users having a way to write `response_duration` without the translation table).
If `response_duration=%RESPONSE_DURATION%` is the only option for users, then there is inconsistency between the original set that are in the translations and the new ones that require more spelling out.

design/envoy-json-logging-custom-fields-design.md
astrieanna: Yea, since the map syntax would require deprecating the prior configuration field anyway, the new field would not have to support the built-in field mappings. In our example in this doc, we assumed that empty values would imply that the key should be looked up in the built-in field mappings, but I'm not sure whether that's desirable or not.

## In favor of a built-in translation table:
* Users of Contour only have to read Contour's docs (vs. also referencing Envoy's config syntax).
* Users get reasonable, standardized field names because that's what's setup by the field mappings.

## In favor of writing all key-value pairs out:
* Users of Contour can read the Envoy docs to decide what config to write. They don't need to know Contour's specific translation of special keys.

youngnick: I like the ease of use of having the translation table, personally. I think that having the two types seems like an okay tradeoff for that ease of use (for now at least).

design/envoy-json-logging-custom-fields-design.md
youngnick: The original intent was that we add simple translations as required to the translation table; we figured that people could just request anything they needed added. With that in mind, I'm okay with adding whatever simple translations to the translation table (like `response_duration`, where it's a straight mapping between field name and Envoy template string), as part of this implementation.

Obviously, this will be frozen at this point in time, but perhaps a comment saying that adding things to the standard translations is fine would make it clearer that we're open to adding more?

design/envoy-json-logging-custom-fields-design.md
```diff
@@ -0,0 +1,144 @@
...
+The Envoy field command will be validated to prevent Envoy misconfiguration.
+The validation will be performed using regular expressions.
+These expressions will be similar to [one used in Envoy](https://github.com/envoyproxy/envoy/blob/4d77fc802c3bc1c517e66c54e9c9507ed7ae8d9b/source/common/formatter/substitution_formatter.cc#L291), but with the specification of exact commands.
+If the validation is failed an error will be raised and printed in the logs of Contour during startup.
+
+### Edge Cases
+
+#### Field name is repeated
+
+If the field name is repeated the last entry in the configuration takes precedes and overwrites the previous value.
```
stevesloka, line 57: Wonder if this should raise a validation error instead of overwriting?

skriss (approved):
LGTM!

design/envoy-json-logging-custom-fields-design.md
jpeach: A validation error could be OK if a field name is repeated in a single `json-fields` array. My original intention with suggesting overwriting was that operators should be able to use this mechanism to change the definition of builtin log fields (ie. overwriting the builtin field mapping).

jpeach:
Last call for comments.

design/envoy-json-logging-custom-fields-design.md
```diff
@@ -0,0 +1,144 @@
+# Allow Envoy to output arbitrary fields for JSON logs
+
+Status: Draft
```
youngnick, line 3: If there are no further comments, @mike1808, could you change this to 'Accepted' before we merge?

mike1808: Done @youngnick

youngnick:
Okay, I'm calling this one done. Thanks @mike1808 and @astrieanna!

Issue #2070, opened 2019-12-19
Break HTTPProxy document into separate pages.


jpeach:
The HTTPProxy document is really long and it is hard to find the section that describe how to achieve specific tasks (especially since we lack good indexing and search). Now that we have generated API reference docs in #2069, let's use that as the primary reference page, and break the existing documentation into a set of task-centric or feature-centric pages. For example, "how to make a virtual host", "how to use the include mechanism", "route conditions", etc.

skriss:
Yeah, and in the TOC it/they should be called something other than "HTTPProxy **Reference**", since there's also an "API Reference" page which is a true reference.

jpeach:
Proposed breakdown of pages for the `Configuration` topic. Numbered items will become a page, items marked by a "#" indicate current headings on the `HTTPProxy.md` page whose content will be migrated.

```
1. HTTPProxy

# HTTPProxy Reference
## Key HTTPProxy Benefits
## Ingress to HTTPProxy
## Interacting with HTTPProxies
## HTTPProxy API Specification
## Status Reporting

2. HTTPProxy virtual hosts 

### Virtual Host Configuration
#### Fully Qualified Domain Name
#### Virtualhost aliases
### Restricted root namespaces

3. HTTPProxy Inclusion and Delegation

## HTTPProxy inclusion
### Conditions and Inclusion
### Configuring inclusion
#### Within the same namespace
#### Across namespaces
### Orphaned HTTPProxy children

4. TLS Termination

#### TLS
##### Fallback Certificate
###### Configuration
#### Permit Insecure
## Client Certificate Validation

## TCP Proxying
### TLS Termination at the edge
### TLS passthrough to the backend service
### TCPProxy delegation

5. Upstream TLS

#### Upstream TLS
##### Envoy Client Certificate
## Upstream Validation
##### Error conditions (for upstream validation)

6. Request Routing

### Conditions
#### Prefix conditions
#### Header conditions
### Routes
#### Multiple Upstreams
#### Upstream Weighting
#### Request and Response Header Policies
#### Traffic mirroring
#### Response Timeout
#### Load Balancing Strategy
#### Session Affinity
##### Limitations (of session affinity)

7. External Service Routing

### ExternalName
#### Proxy to external resource

8. Request rewriting

#### Path Rewriting
### Header Policy

9. CORS

#### CORS policy

10. Websockets

#### WebSocket Support

11. Backend Service Health Checking

#### TCP Proxy health checking
#### Per route health checking

[...]

PR #3049, opened 2020-10-21
doc: reclassify Contour configuration as deployment

jpeach:
Move the Contour configuration reference under the "Deployment" heading,
since it is more properly related to that. This leaves the "Configuration"
heading more strictly aligned with configuring the Contour API.

This updates #2070.


skriss (approved):
LGTM

PR #3050, opened 2020-10-21
testsuite: add Contour installation scripts

jpeach:
Separate the Contour installation from the script that builds a kind
cluster for integration testing. This makes it easier to install a
released version of Contour and test upgrading.


youngnick (approved):
LGTM

PR #3055, opened 2020-10-22
site: change navbar element type

jpeach:
The leftmost navbar headings are `h1` elements, which means that table
of contents generators would notice them. However, for the structure of
the documentation, they should be ignored and we shoild only have to
care about the `h` elelements in the main page.

To fix this, we can use a `p` element styled like a heading. This also
fixes the inline per-page TOC which can otherwise erroneously display
the nav headings.


youngnick (approved):
LGTM

PR #3057, opened 2020-10-22
doc: break the HTTPProxy documention into smaller pages

### LINKED ISSUE #2070: Break HTTPProxy document into separate pages.

jpeach:
Break the single large HTTPProxy document into multiple smaller pages,
each focused on a single task or aspect of the Contour API. This makes
it easier for users to search through the documentation, and it also
makes it easier for contributors to find a place to add new information.

This change contains the minimal necessary editing of the original text
that was taken from the larger document.

This fixes #2070.


youngnick (approved):
LGTM, much easier to navigate.

PR #3059, opened 2020-10-23
Add support for custom Envoy JSON fields

### LINKED ISSUE #1507: JSON logging should report invalid headers
The current implementation of JSON logging silently discards invalid headers. Invalid headers should be surfaced to the user, probably at startup.

(see `internal/contour/accesslog.go` to get started).

### LINKED ISSUE #3032: Support custom JSON fields for Envoy access logs

mike1808:
Add support for custom Envoy JSON fields via `=` in the fields name. The specify custom field name and its Envoy format string use `=` in the field name. The syntax is the following: `field_name=Envoy format string`, where Envoy format string can have any command except `DYNAMIC_METADATA` and `FILTER_STATE` commands. 

Example of custom field:

```yaml
json-fields:
  - @timestamp
  - customer_id=%REQ(X-CUSTOMER-ID)%
```

Fixes #3032, Fixes #1507

cc @KauzClay

### DIFF
--- a/pkg/config/parameters.go
+++ b/pkg/config/parameters.go
@@ -19,6 +19,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"time"
 
@@ -88,6 +89,90 @@ func (a AccessLogType) Validate() error {
 
 const EnvoyAccessLog AccessLogType = "envoy"
 const JSONAccessLog AccessLogType = "json"
+
+type AccessLogFields []string
+
+func (a AccessLogFields) Validate() error {
+	// Capture Groups:
+	// Given string "the start time is %START_TIME(%s):3% wow!"
+	//
+	//   0. Whole match "%START_TIME(%s):3%"
+	//   1. Full operator: "START_TIME(%s):3%"
+	//   2. Operator Name: "START_TIME"
+	//   3. Arguments: "(%s)"
+	//   4. Truncation length: ":3"
+	re := regexp.MustCompile(`%(([A-Z_]+)(\([^)]+\)(:[0-9]+)?)?%)?`)
+
+	for key, val := range a.AsFieldMap() {
+		if val == "" {
+			return fmt.Errorf("invalid JSON log field name %s", key)
+		}
+
+		if jsonFields[key] == val {
+			continue
+		}
+
+		// FindAllStringSubmatch will always return a slice with matches where every slice is a slice
+		// of submatches with length of 5 (number of capture groups + 1).
+		tokens := re.FindAllStringSubmatch(val, -1)
+		if len(tokens) == 0 {
+			continue
+		}
+
+		for _, f := range tokens {
+			op := f[2]
+			if op == "" {
+				return fmt.Errorf("invalid JSON field: %s, invalid Envoy format: %s", val, f)
+			}
+
+			_, okSimple := envoySimpleOperators[op]
+			_, okComplex := envoyComplexOperators[op]
+			if !okSimple && !okComplex {
+				return fmt.Errorf("invalid JSON field: %s, invalid Envoy format: %s, invalid Envoy operator: %s", val, f, op)
+			}
+
+			if (op == "REQ" || op == "RESP" || op == "TRAILER") && f[3] == "" {
+				return fmt.Errorf("invalid JSON field: %s, invalid Envoy format: %s, arguments required for operator: %s", val, f, op)
+			}
+
+			// START_TIME cannot not have truncation length.
+			if op == "START_TIME" && f[4] != "" {
+				return fmt.Errorf("invalid JSON field: %s, invalid Envoy format: %s, operator %s cannot have truncation length", val, f, op)
+			}
+		}
+	}
+
+	return nil
+}
+
+func (a AccessLogFields) AsFieldMap() map[string]string {
+	fieldMap := map[string]string{}
+
+	for _, val := range a {
+		parts := strings.SplitN(val, "=", 2)
+
+		if len(parts) == 1 {
+			operator, foundInFieldMapping := jsonFields[val]
+			_, isSimpleOperator := envoySimpleOperators[strings.ToUpper(val)]
+
+			if isSimpleOperator && !foundInFieldMapping {
+				// Operator name is known to be simple, upcase and wrap it in percents.
+				fieldMap[val] = fmt.Sprintf("%%%s%%", strings.ToUpper(val))
+			} else if foundInFieldMapping {
+				// Operator name has a known mapping, store the result of the mapping.
+				fieldMap[val] = operator
+			} else {
+				// Operator name not found, save as emptystring and let validation catch it later.
+				fieldMap[val] = ""
+			}
+		} else {
+			// Value is a full key:value pair, store it as is.
+			fieldMap[parts[0]] = parts[1]
+		}
+	}
+
+	return fieldMap
+}
 
 // HTTPVersionType is the name of a supported HTTP version.
 type HTTPVersionType string
@@ -288,7 +373,7 @@ type Parameters struct {
 
 	// AccessLogFields sets the fields that JSON logging will
 	// output when AccessLogFormat is json.
-	AccessLogFields []string `yaml:"json-fields,omitempty"`
+	AccessLogFields AccessLogFields `yaml:"json-fields,omitempty"`
 
 	// TLS contains TLS policy parameters.
 	TLS TLSParameters `yaml:"tls,omitempty"`
@@ -341,10 +426,8 @@ func (p *Parameters) Validate() error {
 		return err
 	}
 
-	for _, f := range p.AccessLogFields {
-		if _, ok := JSONFields[f]; !ok {
-			return fmt.Errorf("invalid JSON log field name %s", f)
-		}
+	if err := p.AccessLogFields.Validate(); err != nil {
+		return err
 	}
 
 	// Check TLS secret names.
--- a/pkg/config/parameters_test.go
+++ b/pkg/config/parameters_test.go
@@ -126,6 +126,48 @@ func TestValidateAccessLogType(t *testing.T) {
 	assert.NoError(t, JSONAccessLog.Validate())
 }
 
+func TestValidateAccessLogFields(t *testing.T) {
+	errorCases := [][]string{
+		{"dog", "cat"},
+		{"req"},
+		{"resp"},
+		{"trailer"},
+		{"@timestamp", "dog"},
+		{"@timestamp", "content-id=%REQ=dog%"},
+		{"@timestamp", "content-id=%dog(%"},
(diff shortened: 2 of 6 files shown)


pkg/config/parameters_test.go
```diff
@@ -126,6 +126,24 @@ func TestValidateAccessLogType(t *testing.T) {
 	assert.NoError(t, JSONAccessLog.Validate())
 }
 
+func TestValidateAccessLogFields(t *testing.T) {
+	assert.Error(t, AccessLogFields([]string{"dog", "cat"}).Validate())
```
jpeach, line 130: Should we accept something like`[]string{"dog=pug", "cat=black"}`?

pkg/config/parameters.go
```diff
@@ -89,6 +90,88 @@ func (a AccessLogType) Validate() error {
 const EnvoyAccessLog AccessLogType = "envoy"
 const JSONAccessLog AccessLogType = "json"
 
+type AccessLogFields []string
```
jpeach, line 93: Why don't you make this a `map[string]string` so that it can be fully denormalized in the config? i.e, after validation, break the config down into key/value pairs so that the caller in `internal/envoy/v2/accesslog.go` doesn't need to do it.

jpeach (changes requested):
Hi @mike1808,

Overally, I think this looks pretty good.

1. We should document this in `./site/_site/docs/main/configuration.md`. It's
OK to do this in a separate PR if you prefer.

2. I think that it would be clearer to split the validation regex to handle the
simple case (e.g. `%HOSTNAME%`) and complex case (e.g. `%REQ(foo?bar):baz%`)
separately.  This would make the regex much more obvious and the code easier to
follow. The `REQ/RESP` regex only needs to validate the capture from the simple
case (i.e. `REQ(foo?bar):baz`)

3. Once you do (2), there's an opportunity to have a single definition of the
simple expressions. If you define `EnvoyOperators` to just be the simple
operators, then you can derive the JSON field name from that, and add those
mappings to `JSONFields` in an `init` function. The `EnvoyOperators` map should
go into `pkg/config/accesslog.go`.

pkg/config/parameters_test.go
```diff
@@ -126,6 +126,24 @@ func TestValidateAccessLogType(t *testing.T) {
...
+	assert.Error(t, AccessLogFields([]string{"@timestamp", "content-id=%DOG%"}).Validate())
+	assert.Error(t, AccessLogFields([]string{"@timestamp", "duration=my durations % are %DURATION%.0 and %REQ(:METHOD)%"}).Validate())
+
+	assert.NoError(t, AccessLogFields([]string{"@timestamp", "method"}).Validate())
+	assert.NoError(t, AccessLogFields([]string{"@timestamp", "response_duration"}).Validate())
+	assert.NoError(t, AccessLogFields([]string{"@timestamp", "duration=%DURATION%.0"}).Validate())
+	assert.NoError(t, AccessLogFields([]string{"@timestamp", "duration=My duration=%DURATION%.0"}).Validate())
+	assert.NoError(t, AccessLogFields([]string{"@timestamp", "duration=%START_TIME(%s.%6f)%"}).Validate())
+	assert.NoError(t, AccessLogFields([]string{"@timestamp", "content-id=%REQ(X-CONTENT-ID)%"}).Validate())
+	assert.NoError(t, AccessLogFields([]string{"@timestamp", "duration=my durations are %DURATION%.0 and method is %REQ(:METHOD)%"}).Validate())
```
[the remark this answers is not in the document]
mike1808: Yes, it does.

pkg/config/parameters.go
mike1808: We were thinking about that, but because this type is used for a [YAML config parser](https://github.com/projectcontour/contour/pull/3059/files/c007f5c3b7e6ad3edb1c617a57f945ded2e9490f#diff-a324876cd0bc5df1659faac7631005ed9d83b9445c6720142db3c8d2d3e83e56R374) we cannot change it as it will break the current API. One option is to save it to a different field and then copy to this fields as a `map[string]string`

pkg/config/parameters_test.go
mike1808: @jpeach in theory Envoy supports that and it will put the static value `dog: field` in access log.  I think, if we're supporting other string operators (static strings) we should also accept this.

mike1808:
@jpeach 

1. We'll add the docs in the separate PR.
2. I don't think that's possible because Envoy can have multiple operators in one field. E.g. it can have a simple operator like `%HOSTNAME%` and parameterized one like `%REQ()%`. like `req from %HOSTNAME% with %REQ(X-CONTNENT-ID)%`
3. Because the option (2) is not possible we cannot do that. But we can move `EnvoyOperators` to `pkg/config/accesslog.go`

mike1808:
@jpeach we updated the PR and try to addressed all review comments. One thing we're not sure is our approach for keeping parsed access log fields in `Parameters` structure.

cc @xanderstrike

pkg/config/parameters_test.go
```diff
@@ -126,6 +126,24 @@ func TestValidateAccessLogType(t *testing.T) {
 	assert.NoError(t, JSONAccessLog.Validate())
 }
 
+func TestValidateAccessLogFields(t *testing.T) {
+	assert.Error(t, AccessLogFields([]string{"dog", "cat"}).Validate())
+	assert.Error(t, AccessLogFields([]string{"@timestamp", "dog"}).Validate())
+	assert.Error(t, AccessLogFields([]string{"@timestamp", "content-id=%REQ=dog%"}).Validate())
+	assert.Error(t, AccessLogFields([]string{"@timestamp", "content-id=%dog(%"}).Validate())
+	assert.Error(t, AccessLogFields([]string{"@timestamp", "content-id=%REQ()%"}).Validate())
```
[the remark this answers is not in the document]
mike1808: Done

pkg/config/parameters_test.go
```diff
@@ -126,6 +126,24 @@ func TestValidateAccessLogType(t *testing.T) {
 	assert.NoError(t, JSONAccessLog.Validate())
 }
 
+func TestValidateAccessLogFields(t *testing.T) {
+	assert.Error(t, AccessLogFields([]string{"dog", "cat"}).Validate())
+	assert.Error(t, AccessLogFields([]string{"@timestamp", "dog"}).Validate())
+	assert.Error(t, AccessLogFields([]string{"@timestamp", "content-id=%REQ=dog%"}).Validate())
```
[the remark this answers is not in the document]
mike1808: Fixed

pkg/config/parameters.go
jpeach: > We were thinking about that, but because this type is used for a [YAML config parser](https://github.com/projectcontour/contour/pull/3059/files/c007f5c3b7e6ad3edb1c617a57f945ded2e9490f#diff-a324876cd0bc5df1659faac7631005ed9d83b9445c6720142db3c8d2d3e83e56R374) we cannot change it as it will break the current API.

Oh good point, I had blanked that out. WDYT about adding a method to to the conversion:
```Go
func (a *AccessLogFields) AsFieldMap() map[string]string {
...
}
```

mike1808: Done

mike1808:
@jpeach we renamed `JSONFields` to `jsonFields` and encapsulated all the translation logic in `config` package.

youngnick (approved):
LGTM, nice work. I'm looking forward to seeing the docs update as well, I think that's key for making this feature more usable. We should wait for @jpeach to approve as well though.

pkg/config/parameters.go
```diff
@@ -89,6 +90,82 @@ func (a AccessLogType) Validate() error {
...
+	for _, val := range a {
+		parts := strings.SplitN(val, "=", 2)
+		if len(parts) == 1 {
+			operator := ""
+			operator, ok := jsonFields[val]
+			if _, isSimple := envoySimpleOperators[strings.ToUpper(val)]; isSimple && !ok {
+				operator = fmt.Sprintf("%%%s%%", strings.ToUpper(val))
+			}
+
+			fieldMap[val] = operator
```
jpeach, line 160: I think we are missing a case here. There's 3 things that can happen
1. The field name exists in `jsonFields`
2. The field name is a case-folding match to a `envoySimpleOperators` entry
3. Neither of the above.

In case (3), we would basically do `fieldMap[val] = ""`. I think that this is OK, because we subsequently validate the mapping. Let's comment this clearly for future readers!

pkg/config/parameters.go
```diff
@@ -89,6 +90,82 @@ func (a AccessLogType) Validate() error {
...
+			}
+
+			if (op == "REQ" || op == "RESP" || op == "TRAILER") && f[3] == "" {
+				return fmt.Errorf("invalid JSON field: %s, invalid Envoy format: %s, arguments required for operator: %s", val, f, op)
+			}
+
+			if op == "START_TIME" && f[4] != "" {
+				return fmt.Errorf("invalid JSON field: %s, invalid Envoy format: %s, arguments required for operator: %s", val, f, op)
+			}
+			// START_TIME cannot not have truncation length
```
jpeach, line 141: Nit: can you please make comments are full sentences, including a period (here and above).

jpeach (approved):
Thanks @mike1808, this looks great. Could you just take care of the couple of nits, then we can merge.

pkg/config/parameters.go
```diff
@@ -89,6 +90,82 @@ func (a AccessLogType) Validate() error {
...
+	return nil
+}
+
+func (a AccessLogFields) AsFieldMap() map[string]string {
+	fieldMap := map[string]string{}
+
+	for _, val := range a {
+		parts := strings.SplitN(val, "=", 2)
+		if len(parts) == 1 {
+			operator := ""
```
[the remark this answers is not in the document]
XanderStrike: Good catch, addressed

pkg/config/parameters.go
XanderStrike: This is good feedback thank you. I've reorganized the cases and took another look at the variable names to hopefully clarify what's going on, and of course added some comments to make it extra clear.

XanderStrike:
Latest comments addressed and fixed up into the commit, rebased against `projectcontour:main`. Ready for review again.

jpeach (approved):

Issue #3058, opened 2020-10-23
Incorrect Port Name in Contour Deployment


danehans:
**What steps did you take and what happened:**
The example Contour [deployment](https://github.com/projectcontour/contour/blob/main/examples/contour/03-contour.yaml) names the metrics port as `debug`:

```
kind: Deployment
metadata:
  labels:
    app: contour
  name: contour
  namespace: projectcontour
...
  template:
    metadata:
      annotations:
        prometheus.io/scrape: "true"
        prometheus.io/port: "8000"
...
- containerPort: 8000
    name: debug
    protocol: TCP
...
```

Contour logs confirm that the metrics service listens on `8000` and the debug service listens on 6060:
```
time="2020-10-23T18:38:20Z" level=info msg="started HTTP server" address="0.0.0.0:8000" context=metricsvc
time="2020-10-23T18:38:20Z" level=info msg="started HTTP server" address="127.0.0.1:6060" context=debugsvc
```
Contour's metric service is listening on 
**What did you expect to happen:**
The metrics service to listen on `8000`

**Anything else you would like to add:**
xref: https://github.com/projectcontour/contour-operator/issues/46

**Environment:**

- Contour version: v1.9

/cc @Miciah @jpeach @stevesloka

youngnick:
Agreed, this should be changed to `metrics`.

PR #3061, opened 2020-10-26
Updates Contour Example Container Ports

danehans:
This PR fixes the `containerPort` reference of the Contour example `Deployment` manifest to correctly reference metrics and debug TCP port numbers.

Fixes https://github.com/projectcontour/contour/issues/3058

/asssign @youngnick @jpeach @stevesloka 
/cc @Miciah 
/xref https://github.com/projectcontour/contour-operator/issues/73


danehans:
I see the ` codegen` job failed due to:

```
Uncommitted changes in generated sources:
 M examples/render/contour.yaml
Error: Process completed with exit code 1.
```

When updating the example manifests, do I need to run a make target or something else to update the rendered manifests?

skriss:
Yep, `make generate` will do it.

youngnick (approved):
LGTM, thanks for this.

examples/contour/03-contour.yaml
```diff
@@ -55,6 +55,9 @@ spec:
           name: xds
           protocol: TCP
         - containerPort: 8000
+          name: metrics
+          protocol: TCP
+        - containerPort: 6060
```
stevesloka, line 60: I don't see a need to add the debug port to the deployment. By default it only listens on `127.0.0.1:6060` you can't get to it unless you port-forward to the pod.

stevesloka (changes requested):

examples/contour/03-contour.yaml
danehans: @stevesloka `containerPort` is informational only. Container network ports will only be exposed by the containerized app. I including the debug port here so users understand the app is listening on 6060  (even though it's local host). I can remove if you prefer.

stevesloka: I would remove since it feels weird to me to have a debug port on a production system, but chatting with some others it's fine to leave.

stevesloka (approved):

PR #3067, opened 2020-10-26
Remove Tim and add Tong on Contour website

tong101:
Remove Tim Hinderliter and add Tong Liu to Contour website. Just
remove the contributor and keep Tim's image in case it is still
referenced somewhere else.

Change-Id: I949196067dfa5e0e7dac3466a2e395fc173fad31
Signed-off-by: Tong Liu <tongl@vmware.com>


youngnick (approved):
LGTM

PR #3068, opened 2020-10-27
internal/envoy: Add xDS v3 bootstrap config file

stevesloka:
Adds new config file option to generate a v3 xDS boostrap configuration file.
To generate the v3 version a new arg is passed to the bootstrap config file (`--xds-server-version`)
which defaults to v2, but passing v3 will generate a v3 config.

Updates #1898

### DIFF
--- a/internal/envoy/bootstrap.go
+++ b/internal/envoy/bootstrap.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/golang/protobuf/jsonpb"
 	"github.com/golang/protobuf/proto"
+	"github.com/projectcontour/contour/pkg/config"
 )
 
 // SDSResourcesSubdirectory stores the subdirectory name where SDS path resources are stored to.
@@ -34,7 +35,7 @@ const SDSTLSCertificateFile = "xds-tls-certificate.json"
 // CA certificates for Envoy to use for the XDS gRPC connection.
 const SDSValidationContextFile = "xds-validation-context.json"
 
-// BootstrapConfig holds configuration values for a v2.Bootstrap.
+// BootstrapConfig holds configuration values for a Bootstrap configuration.
 type BootstrapConfig struct {
 	// AdminAccessLogPath is the path to write the access log for the administration server.
 	// Defaults to /dev/null.
@@ -55,6 +56,10 @@ type BootstrapConfig struct {
 	// XDSGRPCPort is the management server port that provides the v2 gRPC API.
 	// Defaults to 8001.
 	XDSGRPCPort int
+
+	// XDSResourceVersion defines the XDS Server Version to use.
+	// Defaults to "v2"
+	XDSResourceVersion config.ResourceVersion
 
 	// Namespace is the namespace where Contour is running
 	Namespace string
--- a/internal/envoy/v3/endpoint.go
+++ b/internal/envoy/v3/endpoint.go
@@ -1,0 +1,43 @@
+// Copyright Project Contour Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v3
+
+import (
+	envoy_core_v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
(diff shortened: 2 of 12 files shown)


internal/envoy/bootstrap.go
```diff
@@ -17,24 +17,26 @@
...
+	"github.com/projectcontour/contour/pkg/config"
 )
 
 // SDSResourcesSubdirectory stores the subdirectory name where SDS path resources are stored to.
 const SDSResourcesSubdirectory = "sds"
 
 // SDSTLSCertificateFile stores the path to the SDS resource with Envoy's
 // client certificate and key for XDS gRPC connection.
-const SDSTLSCertificateFile = "xds-tls-certicate.json"
+const SDSTLSCertificateFile = "xds-tls-certificate.json"
```
skriss, line 33: didn't look deeply into how this is used, but will changing this possibly break existing installs that expect the old typo'd name?

internal/envoy/bootstrap.go
```diff
@@ -56,6 +58,10 @@ type BootstrapConfig struct {
 	// Defaults to 8001.
 	XDSGRPCPort int
 
+	// Defines the XDS Server Version to use.
```
skriss, line 61:
```suggestion
	// XDSServerVersion defines the XDS Server Version to use.
```

internal/envoy/bootstrap.go
```diff
@@ -92,6 +98,18 @@ func (c *BootstrapConfig) GetAdminAccessLogPath() string {
 	return stringOrDefault(c.AdminAccessLogPath, "/dev/null")
 }
 
+// GetXdsVersion returns the version if specified or an error if invalid value.
+// Defaults to "v2".
+func (c *BootstrapConfig) GetXdsVersion() (config.ServerVersion, error) {
+	switch c.XDSServerVersion {
```
skriss, line 104: should be able to use pkg/config/ServerVersion.Validate() here to avoid duplicating the validation

internal/envoy/v3/listener.go
```diff
@@ -0,0 +1,47 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v3
+
+import (
+	envoy_v3_core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+)
+
+// SocketAddress creates a new TCP envoy_api_v2_core.Address.
```
skriss, line 20:
```suggestion
// SocketAddress creates a new TCP envoy_v3_core.Address.
```

skriss (changes requested):

internal/envoy/bootstrap.go
stevesloka: I don't think so, but can also revert since it's not really related. Envoy should look at the directory not the file names.

internal/envoy/bootstrap.go
stevesloka: ahh right, good call. =)

internal/envoy/bootstrap.go
stevesloka: @skriss I rolled this back

jpeach: It should not break any existing installs. If the bootstrap version is updated, it will write and consume the new name. There's no case where an old bootstrap version would write the old name and a new version would try to consume the new name.

internal/envoy/v3/endpoint.go
```diff
@@ -0,0 +1,43 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v3
+
+import (
+	envoy_v3_core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	envoy_v3_endpoint "github.com/envoyproxy/go-control-plane/envoy/config/endpoint/v3"
```
jpeach, line 18: IIRC, the contributing guide suggests this should be:
```suggestion
	envoy_endpoint_v3 "github.com/envoyproxy/go-control-plane/envoy/config/endpoint/v3"
```

stevesloka: 👍

internal/envoy/bootstrap.go
```diff
@@ -56,6 +58,10 @@ type BootstrapConfig struct {
 	// Defaults to 8001.
 	XDSGRPCPort int
 
+	// XDSServerVersion defines the XDS Server Version to use.
+	// Defaults to "v2"
+	XDSServerVersion config.ServerVersion
```
skriss, line 63: it'd be good to rename this one to XDSResourceVersion to match the flag.

skriss (approved):
One more rename but otherwise LGTM.

youngnick (approved):
LGTM with the same nit as @skriss.

internal/envoy/bootstrap.go
stevesloka: 👍 

I'll rename `ServerVersion` in config.Parameters to this as well.

PR #3070, opened 2020-10-27
[WIP] rate limiting background doc

skriss:

Putting this up as a draft PR to solicit feedback. Content subject to change, particularly around possible solution paths. Please comment!

### DIFF
--- a/design/ratelimit-background.md
+++ b/design/ratelimit-background.md
@@ -1,0 +1,379 @@
+# Rate Limiting Support
+
+[Rate limiting support](https://github.com/projectcontour/contour/issues/370) is a common feature request for Contour. An initial [design document](https://github.com/projectcontour/contour/blob/main/design/ratelimit-design.md) was proposed and accepted in early 2019, and an [implementation](https://github.com/projectcontour/contour/pull/873) was drafted, but it was not merged due to competing priorities and became stale. There was also a more recent [design update PR](https://github.com/projectcontour/contour/pull/2283) that was not merged.
+
+The purpose of this document is to solicit further input on the topic from interested parties, in order to have enough information to successfully design and implement the feature in Contour. It presents background information on Envoy's rate limiting support, a survey of other ingress controllers' rate limiting support, and some possible approaches for Contour to take.
+
+## Rate Limiting Goals
+
+Rate limiting is a means of protecting backend services against unwanted traffic. This can be useful for a variety of different scenarios:
+
+- Protecting against denial-of-service (DoS) attacks by malicious actors
+- Protecting against DoS incidents due to bugs in client applications/services
+- Enforcing usage quotas for different classes of clients, e.g. free vs. paid tiers
+- Controlling resource consumption/cost
+
+## Rate Limiting in Envoy
+
+Envoy supports both [local](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/other_features/local_rate_limiting) and [global](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/other_features/global_rate_limiting#arch-overview-global-rate-limit) rate limiting, at both the network and HTTP layers:
+
+| Type | Network (L4) | HTTP (L7) |
+| ---- | ------------ | --------- |
+| Local | [envoy.filters.network.local_ratelimit](https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/local_rate_limit_filter#config-network-filters-local-rate-limit) | [envoy.filters.http.local_ratelimit](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/local_rate_limit_filter#config-http-filters-local-rate-limit) |
+| Global | [envoy.filters.network.ratelimit](https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/rate_limit_filter#config-network-filters-rate-limit) | [envoy.filters.http.ratelimit](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/rate_limit_filter#config-http-filters-rate-limit) |
+
+This document focuses on the HTTP filters, as it's assumed that users will want L7 control over rate limiting.
+
+### Local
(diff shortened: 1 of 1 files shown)


design/ratelimit-background.md
```diff
@@ -0,0 +1,219 @@
...
+This is a more "batteries included" approach that simplifies the UX for folks who don't want to plug their own RLS, but comes at a cost of reduced flexibility.
+
+#### Lyft RLS operator
+
+- As an alternative to the previous option, an operator for the Lyft RLS could be written that defines a `RateLimitPolicy` CRD and translates it into the Lyft config file format. This would be independent of Contour.
+- Contour users would either use the `RateLimitPolicy` API directly to define their rate limits, or Contour could expose its own API (as new `HTTPProxy`fields or a new CRD) that would be translated into the operator's `RateLimitPolicy` API.
+
+## Questions
+
+- Are folks interested in local rate limiting, global rate limiting, or both?
```
jpeach, line 216: I don't see a good use case for local rate limiting. The goal of rate limiting is usually to protect a backend service for availability or cost reasons. Local rate limiting would require someone to distribute a global rate limit across the envoy fleet, then assume that clients are balanced evenly.

As I write this, I think that Contour potentially has enough information to do that. You'd need to build and expose the envoy fleet information internally, then dynamically adjust the rate limits.

stevesloka:
Thanks for this @skriss, this is a great start!

We should add a section that outlines some of the implementation specifics that @moderation brought up about where/how to deploy the rate-limiting service along with its different benefits, drawbacks, and added complexity.

design/ratelimit-background.md
skriss: I agree that it's pretty limited in terms of what it can provide. BUT, it's also far easier to configure and operate. I put it out there to see if users are interested in it or not. 🤷‍♂️

jpeach: Right. One thing that this doc doesn't address is the purpose of rate limiting. What capabilities are we trying to give to operators and why? That would help to put the different options in context.

tsaarni:
I believe the network level filters were left out on purpose, but maybe they could be mentioned, possibly together with a summary of the available rate limiting filters:

| filter | description |
| --- | --- |
| [extensions.filters.http.ratelimit.v3.RateLimit](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ratelimit/v3/rate_limit.proto#extensions-filters-http-ratelimit-v3-ratelimit) | Global HTTP rate limit filter |
| [extensions.filters.http.local_ratelimit.v3.LocalRateLimit](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/local_ratelimit/v3/local_rate_limit.proto#extensions-filters-http-local-ratelimit-v3-localratelimit) | Local HTTP rate limit filter |
| [extensions.filters.network.ratelimit.v3.RateLimit](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/ratelimit/v3/rate_limit.proto#envoy-v3-api-msg-extensions-filters-network-ratelimit-v3-ratelimit) | Global network rate limit filter |
| [extensions.filters.network.local_ratelimit.v3.LocalRateLimit](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/local_ratelimit/v3/local_rate_limit.proto#extensions-filters-network-local-ratelimit-v3-localratelimit) | Local network rate limit filter |

In my case, there is interest for rate limiting for availability reasons, for protecting from denial of service attack. Running global rate limiting server raises concerns due to overhead / footprint / latency, and high availability of that solutions itself. It might be overkill in my use case. So for that reason, local rate limiting seems attractive if it could still provide meaningful protection.

youngnick (commented):
This is a great document, nice work @skriss!

Reading it, I think that it's critical we get some answers to the questions section, so that we can let our community guide some of the design decisions here.

My gut feeling is that some level of local rate-limiting, at a very coarse-grained level, seems useful pretty much everywhere, as long as the limits are also tweakable, as it will help to stop denial-of-service on any the service itself and any ExtensionServices used by the deployment, be they rate-limiting, auth, or something else.

erwbgy:
Thanks for writing this up.  

My 2c.  We're most interested in the "Protecting against DoS incidents due to bugs in client applications/services" scenario as we've had some instances where a bad release of one application has affected others using the same shared ingress proxy and going forward malicious DoS will be a concern.

Local rate limiting will be helpful in avoiding a single Envoy from being overloaded so we'd like that supported but global rps rate limiting by service and client IP is what we really need.  One important feature that I don't see mentioned is the ability to set a default rate limit that is applied if one is not explicitly covered.  Experience has shown that this helps prevent issues rather than reactively adding limits after they have occurred.  Like with other fields we'd also need to be able to configure maximum values, but this could be handled using a validating webhook if required.

Integrating tightly with the Envoy reference RLS makes a lot of sense as it appears to be well maintained and will be compatible with Envoy going forward.  There does need to be some way for users to configure rate limits in their own Namespaces, otherwise every admin is going to have to figure out a way to configure the RLS.  

My expectation is that it will be easier to have a separate RateLimitPolicy CRD and then have an HTTPProxy ratelimit field that refers to it (like tls/secretName).  This should allow the same RateLimitPolicy objects to be used by multiple routes and make it easy to configure a default policy when no specific one is configured by using an object in projectcontour, for example.  It will also make validation easier. When creating RateLimitPolicy objects it will be really important to know what the current rates are so documentation about what metrics to look at will be required.

stevesloka:
> My expectation is that it will be easier to have a separate RateLimitPolicy CRD and then have an HTTPProxy ratelimit field that refers to it (like tls/secretName). This should allow the same RateLimitPolicy objects to be used by multiple routes and make it easy to configure a default policy when no specific one is configured by using an object in projectcontour, for example. It will also make validation easier. When creating RateLimitPolicy objects it will be really important to know what the current rates are so documentation about what metrics to look at will be required.

This is an interesting take @erwbgy! I like the idea of creating "rate limit policies", then applying them to routes. 

We have been struggling to come up with a good integration UX between Contour<>RateLimitService. One idea was similar to yours which would be to define the rate limits inside an HTTPProxy, then have Contour write out a CRD which represented that ratelimit. Another controller could then pick up that RatelimitCRD and configure the external rate limit service. 

But your idea I think would work as well if the controller watching the ratelimitPolicies could configure the rate limit service.  Thanks for the feedback!

PR #3071, opened 2020-10-27
WiP internal: program 503 routes for misconfigured services

### LINKED ISSUE #3039: Ingress path missbehaving when upstream service is not found.
Installed an ingress resource with:
/ path served by service1
/second served by service2 

[backend1.txt](https://github.com/projectcontour/contour/files/5403230/backend1.txt)
[backend2.txt](https://github.com/projectcontour/contour/files/5403231/backend2.txt)
[ingress.txt](https://github.com/projectcontour/contour/files/5403234/ingress.txt)

Both service1 and service2 is installed.
So far everything is fine
/second prefix is routed to service2, everything else routed to service1.

I remove service2 from the cluster.
Now everything, including /second prefix is served by **service1**

The same setup should result in 503 to everything that is with prefix /second

I will include ingress, and backend manifests.
Also config dump when service2 is present and when it is not.

[backend_missing.log](https://github.com/projectcontour/contour/files/5403232/backend_missing.log)
[backend_present.log](https://github.com/projectcontour/contour/files/5403233/backend_present.log)

**Environment:**

- Contour version: 1.7.0
- Kubernetes version: (use `kubectl version`): v1.15.3
- OS (e.g. from `/etc/os-release`): suse

tsaarni:
Program Envoy with routes that return 503 for misconfigured services.
This avoids unnecessarily taking vhost down or leaking requests into
wrong backend when Ingress has several routes and one of those is
misconfigured.

Fixes #3039

### DIFF
--- a/internal/contour/metrics_test.go
+++ b/internal/contour/metrics_test.go
@@ -462,11 +462,9 @@ func TestHTTPProxyMetrics(t *testing.T) {
 		wantIR: nil,
 		wantProxy: &metrics.RouteMetric{
 			Invalid: map[metrics.Meta]int{
-				{Namespace: "roots"}: 1,
+				{Namespace: "roots"}: 2,
 			},
-			Valid: map[metrics.Meta]int{
-				{Namespace: "roots", VHost: "example.com"}: 1,
-			},
+			Valid:    map[metrics.Meta]int{},
 			Orphaned: map[metrics.Meta]int{},
 			Root: map[metrics.Meta]int{
 				{Namespace: "roots", VHost: "example.com"}: 1,
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -3057,7 +3057,14 @@ func TestDAGInsert(t *testing.T) {
 			objs: []interface{}{
 				i1,
 			},
-			want: listeners(),
+			want: listeners(
+				&Listener{
+					Port: 80,
+					VirtualHosts: virtualhosts(
+						virtualhost("*", prefixroute("/", nil)), // 503 Service Unvailable
+					),
+				},
+			),
 		},
 		"insert ingress w/ default backend": {
 			objs: []interface{}{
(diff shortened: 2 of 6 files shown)


tsaarni:
In this draft I've used Envoy Direct response for routes that should return 503. Would this be reasonable?
I tried creating empty clusters instead (=cluster with no upstream) but that caused too much trouble. For example, the cluster names are derived from services and several places were needing if-statement to cover the special case of service not existing. 

Tests are yet to be fixed.  Previously e.g. listener was not configured at all when ingress was misconfigured with non-existing service. Now there will a listener and a route with Direct response.

I'd like to double-check also about equivalent `HTTPProxy` case.  Currently, if one of the routes for virtualhost is misconfigured with a missing service, the whole HTTPProxy is taken down instead of having that specific path returning 503.  Is this intentional?

tsaarni:
I haven't looked at the tests, but just as an experiment I added similar branch to `HTTPProxy` to produce empty route when service reference is misconfigured. That seems to behave similarly as it did for `Ingress`, though I suppose there are more scenarios in case of `HTTPProxy` that could be considered as non-fatal errors.  What do you think about this in general? 

The status still gets set "invalid" and the detailed errors are shown in `status.conditions`, but the vhost still responds and routes that point to existing service work.

jpeach:
+1 I think that this is an improvement. You can add a warning to HTTPProxy status too.

skriss:
> In this draft I've used Envoy Direct response for routes that should return 503. Would this be reasonable?

Seems reasonable to me, I agree this is a better/more obvious UX 👍

youngnick:
Thanks for this @tsaarni. I think this is an improvement for Ingress, but for HTTPProxy would like to see the end state be that we have the same behavior, and there are warning conditions on the HTTPProxy. Having the proxy be invalid and still serve traffic seems confusing to me, I think? I'd probably rather have it be valid with warnings for these kind of non-fatal errors, although that's on me for not doing it right the first time.

tsaarni:
Update from my side: I've been working with something else that came up, but I still intend to come back to this PR shortly. Sorry for being so quiet!

sunjayBhatia:
Heo @tsaarni are you planning on actively working on this? Just wondering so we can clean up some of the open issues in the project board

tsaarni:
I halted working with this partly due to waiting for #3145 to give guidance.  Now that the design doc is parked, I think this PR should be closed as well - so doing that for now.

Issue #3047, opened 2020-10-20
xDS v2-->v3 Upgrade Plan


stevesloka:
As Contour moves to the xDS v3 version API we need to think about how users upgrade from v2 to v3 in production systems.

Upgrading to v3 requires a new bootstrap config on the Envoy instances. After Contour is upgraded and only serves v3, there is a period of time where any Envoy that still runs the v2 API won't receive updates. 

## How to solve?
- Have two instances of Contour, one with v2 and one with v3 such that as old Envoy's are recycled into new, then they can still receive updates. 
- Have Contour serve up two instances of xDS version from the same contour pod, but running on different ports? 

// cc #1898

jpeach:
> * Have Contour serve up two instances of xDS version from the same contour pod, but running on different ports?

Pretty sure you can have v2 and v3 running in the same gRPC server

moderation:
Related to the move to Envoy 1.16.0 in https://github.com/projectcontour/contour/pull/3029 will by more v2 warnings as per https://www.envoyproxy.io/docs/envoy/v1.16.0/version_history/current
` config: additional warnings have been added for the use of v2 APIs. These appear as log messages and are also captured in the deprecated_feature_use counter after server initialization.`

jpeach:
Accepted. Holding for feedback.

stevesloka:
https://www.envoyproxy.io/docs/envoy/latest/faq/api/control_plane_version_support

stevesloka:
@moderation thanks for the link to the logs, I do see those now in Envoy v1.16.0. =)

stevesloka:
I played around with this a bit today and think @jpeach's suggestion is good to run both versions of the server at the same time.

I haven't been able to get different transport / resource api versions nor get the `envoy.reloadable_features.enable_type_url_downgrade_and_upgrade` to work. 

**Solution:**
Currently Contour serves a v2 endpoint which Envoy clients connect to since `transport_api_version` & `resource_api_version` aren't currently defined in the contour boostrap config file. We will then add the v3 endpoints to the same v2 server currently running. New Envoy clients can be upgraded with v3 boostrap config files referencing `transport_api_version: V3` & `resource_api_version: V3` and will connect to the v3 endpoints and get their configuration where older Envoy clients whom still reference v2 will still get their configuration. 

[...]

PR #3073, opened 2020-10-27
config: Change xds-server-version to be plural and serve multiple xDS versions

stevesloka:
Changes the xds-server-version config item to be plural so that users can specify a set
of xDS versions Contour will serve. Defaults to V2 only for now.

Updates #1898
Updates #3047 

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -48,7 +48,6 @@ import (
 	"github.com/projectcontour/contour/pkg/config"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
-	"google.golang.org/grpc"
 	"gopkg.in/alecthomas/kingpin.v2"
 	corev1 "k8s.io/api/core/v1"
 	networking_api_v1beta1 "k8s.io/api/networking/v1beta1"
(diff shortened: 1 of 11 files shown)


pkg/config/parameters.go
```diff
@@ -333,8 +333,13 @@ func (p *Parameters) Validate() error {
 		return err
 	}
 
-	if err := p.Server.XDSServerVersion.Validate(); err != nil {
-		return err
+	if len(p.Server.XDSServerVersions) == 0 {
+		return fmt.Errorf("at least one xds server version must be specified")
+	}
+	for _, v := range p.Server.XDSServerVersions {
```
skriss, line 339: it'd probably be good to de-dupe the list too, so `[v2, v2]` becomes just `[v2]`.

skriss (approved):
This looks OK to me pending the one comment.

This field will default to `[v2]` now, then `[v2, v3]` when we support both, and then just `[v3]` once we drop `v2` support - right?

internal/xds/server.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
+	"github.com/prometheus/client_golang/prometheus"
+	"google.golang.org/grpc"
+)
+
+// If registry is non-nil gRPC server metrics will be automatically
+// configured and enabled.
+func NewServer(registry *prometheus.Registry, opts ...grpc.ServerOption) *grpc.Server {
```
jpeach, line 24: nit: we could make the `registry` argument compulsory and just use `grpc.NewServer` when there are no metrics. Not a big deal either way though.

jpeach (approved):
LGTM, just a few nits. Nice work!

stevesloka:
> This field will default to [v2] now, then [v2, v3] when we support both, and then just [v3] once we drop v2 support - right?

Yes, once we fully make v3 the default, we'll swap the default values, then will need to deprecate out the v2 version.

cmd/contour/serve.go
```diff
@@ -652,32 +652,30 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
 		case config.EnvoyServerType:
-			switch ctx.Config.Server.XDSServerVersion {
-			case config.XDSv3:
-				log.Fatalf("xDS server type %q not yet implemented for server type %q!", ctx.Config.Server.XDSServerVersion, ctx.Config.Server.XDSServerType)
-			default:
-				grpcServer = contour_xds_v2.RegisterServer(
-					envoy_server_v2.NewServer(context.Background(), snapshotCache, nil),
-					registry,
-					ctx.grpcOptions(log)...)
+			grpcServer = xds.NewServer(registry, ctx.grpcOptions(log)...)
```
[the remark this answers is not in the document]
stevesloka: ahh good call!

cmd/contour/serve.go
```diff
@@ -652,32 +652,30 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+				case config.XDSv2:
+					contour_xds_v2.RegisterServer(envoy_server_v2.NewServer(context.Background(), snapshotCache, nil), grpcServer)
+				case config.XDSv3:
+					log.Fatalf("xDS server type %q not yet implemented for server type %q!", ver, ctx.Config.Server.XDSServerType)
+				}
 			}
 		default:
-			// XXX(jpeach) can't happen due to config validation.
-			log.Fatalf("invalid xdsServerType %q configured", ctx.Config.Server.XDSServerType)
+			grpcServer = xds.NewServer(registry, ctx.grpcOptions(log)...)
```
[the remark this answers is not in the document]
stevesloka: ok I can add back. We should already get the error message earlier that that value is bad and panic, but agree it makes sense to be explicit.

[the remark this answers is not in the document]
stevesloka: @jpeach I'm going to add the same `default` condition for the version as well to keep that consistent.

internal/xds/server.go
stevesloka: Yeah I think that makes sense. I have an issue covering that I think that we could do that work there.

stevesloka:
@jpeach all updated. Also please let me know if you agree with the extra default cases for the version:

-https://github.com/stevesloka/contour/blob/bb60880a172875024a276914e7a5a83c8e331285/cmd/contour/serve.go#L662
- https://github.com/stevesloka/contour/blob/bb60880a172875024a276914e7a5a83c8e331285/cmd/contour/serve.go#L675

cmd/contour/serve.go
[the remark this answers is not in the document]
jpeach: Sounds good!

cmd/contour/serve.go
```diff
@@ -649,31 +648,32 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
 		case config.EnvoyServerType:
-			switch ctx.Config.Server.XDSServerVersion {
-			case config.XDSv3:
-				log.Fatalf("xDS server type %q not yet implemented for server type %q!", ctx.Config.Server.XDSServerVersion, ctx.Config.Server.XDSServerType)
-			default:
-				grpcServer = contour_xds_v2.RegisterServer(
-					envoy_server_v2.NewServer(context.Background(), snapshotCache, nil),
-					registry,
-					ctx.grpcOptions(log)...)
+			for _, ver := range ctx.Config.Server.XDSServerVersions {
```
jpeach, line 655: Quick drive-by suggestion ... it might be cleaner to make this more like `xds.NewContourServer(vers, log, other params)` and let the package deal with the versioning internally.

internal/featuretests/v2/featuretests.go
```diff
@@ -130,9 +131,8 @@ func setup(t *testing.T, opts ...interface{}) (cache.ResourceEventHandler, *Cont
 	l, err := net.Listen("tcp", "127.0.0.1:0")
 	require.NoError(t, err)
 
-	srv := contour_xds_v2.RegisterServer(
-		contour_xds_v2.NewContourServer(log, xdscache.ResourcesOf(resources)...),
-		r /* Prometheus registry */)
+	srv := xds.NewServer(r /* Prometheus registry */)
```
skriss, line 134: we should probably rename `r` and drop the comment :)

cmd/contour/serve.go
stevesloka: I had that originally, but ran into import cycle errors. 😢

pkg/config/parameters.go
stevesloka: I added a bit about de-duping in the validate method. Please take a look @skriss

skriss (approved):
LGTM

PR #3074, opened 2020-10-28
internal/xds: convert v3 protobufs automatically

jpeach:

### DIFF
--- a/internal/xds/util.go
+++ b/internal/xds/util.go
@@ -14,8 +14,32 @@
 package xds
 
 import (
+	"fmt"
 	"strings"
 
+	envoy_api_v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
+	envoy_api_auth_v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2/auth"
+	envoy_api_core_v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
+	envoy_config_accesslog_v2 "github.com/envoyproxy/go-control-plane/envoy/config/accesslog/v2"
+	envoy_config_cluster_v3 "github.com/envoyproxy/go-control-plane/envoy/config/cluster/v3"
+	envoy_config_endpoint_v3 "github.com/envoyproxy/go-control-plane/envoy/config/endpoint/v3"
+	envoy_config_filter_http_ext_authz_v2 "github.com/envoyproxy/go-control-plane/envoy/config/filter/http/ext_authz/v2"
+	envoy_config_filter_http_lua_v2 "github.com/envoyproxy/go-control-plane/envoy/config/filter/http/lua/v2"
+	envoy_config_filter_network_http_connection_manager_v2 "github.com/envoyproxy/go-control-plane/envoy/config/filter/network/http_connection_manager/v2"
+	envoy_config_filter_network_tcp_proxy_v2 "github.com/envoyproxy/go-control-plane/envoy/config/filter/network/tcp_proxy/v2"
+	envoy_config_listener_v3 "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
+	envoy_config_route_v3 "github.com/envoyproxy/go-control-plane/envoy/config/route/v3"
+	envoy_extensions_access_loggers_file_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/access_loggers/file/v3"
+	envoy_extensions_filters_http_ext_authz_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/ext_authz/v3"
+	envoy_extensions_filters_http_lua_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/lua/v3"
+	envoy_extensions_filters_network_http_connection_manager_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/http_connection_manager/v3"
+	envoy_extensions_filters_network_tcp_proxy_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/tcp_proxy/v3"
+	envoy_extensions_transport_sockets_tls_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/transport_sockets/tls/v3"
+	"github.com/golang/protobuf/ptypes"
+	"github.com/golang/protobuf/ptypes/any"
+	"github.com/projectcontour/contour/internal/protobuf"
+
+	"github.com/golang/protobuf/proto"
 	"k8s.io/apimachinery/pkg/types"
 )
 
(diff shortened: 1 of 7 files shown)


stevesloka:
Thanks for this @jpeach! This is a neat approach to moving form one version to the next. Taking some time to think through this a bit more and how it might integrate I'm leaning towards our original plan of duplicating code into a `v3` package. 

- One goal was to allow users to decide if they want to run a v2 or v3 cluster, so doing this approach won't let us serve v2 once we start migrating over to v3 unless, like you mentioned, we'd need to convert from v3 back to v2. 
- There is a deprecated field with that I'm aware of already (`DrainConnectionsOnHostRemoval` has been removed from v3), so we'd have a special case there to convert. I'm not sure if there are more or not. 

I'm still open to discussing this moving forward since this is a pretty big decision. What is everyone else's thoughts?

internal/xds/util.go
```diff
@@ -38,3 +60,157 @@ func ClusterLoadAssignmentName(service types.NamespacedName, portName string) st
...
+		}
+
+		for _, f := range v3.HttpFilters {
+			if c := f.GetTypedConfig(); c != nil {
+				f.ConfigType = &envoy_extensions_filters_network_http_connection_manager_v3.HttpFilter_TypedConfig{
+					TypedConfig: Fixup(c).(*any.Any),
+				}
+			}
+		}
+
```
skriss, line 163: I think you need something like this in here for the access log config, it was still coming through as v2 without:

```
for i, a := range v3.AccessLog {
        v3.AccessLog[i].ConfigType = &envoy_config_accesslog_v3.AccessLog_TypedConfig{
	        TypedConfig: Fixup(a.GetTypedConfig()).(*any.Any),
	}
}
```

jpeach:
I verified that migrating the protobufs through JSON serialization drops the migrated fields:

```
2020/10/30 10:47:01 original *envoy_api_v2.Cluster:
{"name":"default/httpbin/80/f7b4732cd5","altStatName":"default_httpbin_80","type":"EDS","edsClusterConfig":{"edsConfig":{"apiConfigSource":{"apiType":"GRPC","grpcServices":[{"envoyGrpc":{"clusterName":"contour"}}]}},"serviceName":"default/httpbin/http"},"connectTimeout":"0.250s","healthChecks":[{"timeout":"2s","interval":"10s","unhealthyThreshold":3,"healthyThreshold":2,"httpHealthCheck":{"host":"contour-envoy-healthcheck","path":"/status/200"}}],
"commonLbConfig":{"healthyPanicThreshold":{}},"drainConnectionsOnHostRemoval":true}
```
```
2020/10/30 10:47:01 converted *envoy_config_cluster_v3.Cluster:
{"name":"default/httpbin/80/f7b4732cd5","altStatName":"default_httpbin_80","type":"EDS","edsClusterConfig":{"edsConfig":{"apiConfigSource":{"apiType":"GRPC","transportApiVersion":"V3","grpcServices":[{"envoyGrpc":{"clusterName":"contour"}}]},"resourceApiVersion":"V3"},"serviceName":"default/httpbin/http"},"connectTimeout":"0.250s","healthChecks":[{"timeout":"2s","interval":"10s","unhealthyThreshold":3,"healthyThreshold":2,"httpHealthCheck":{"host":"contour-envoy-healthcheck","path":"/status/200"}}],
"commonLbConfig":{"healthyPanicThreshold":{}}}
```

Envoy has internal code that migrates field in the version converter. Possibly if we just rewrite the typeURL the internal envoy code will do the right thing. Experimenting ...

jpeach:
> Envoy has internal code that migrates field in the version converter. Possibly if we just rewrite the typeURL the internal envoy code will do the right thing. Experimenting ...

Confirmed that this works:

```
"cluster": {
      "@type": "type.googleapis.com/envoy.config.cluster.v3.Cluster",
      "name": "default/httpbin/80/f7b4732cd5",
      "type": "EDS",
      "eds_cluster_config": {
       "eds_config": {
        "api_config_source": {
         "api_type": "GRPC",
         "grpc_services": [
          {
           "envoy_grpc": {
            "cluster_name": "contour"
           }
          }
         ]
        }
       },
       "service_name": "default/httpbin/http"
      },
      "connect_timeout": "0.250s",
      "health_checks": [
       {
        "timeout": "2s",
        "interval": "10s",
        "unhealthy_threshold": 3,
        "healthy_threshold": 2,
        "http_health_check": {
         "host": "contour-envoy-healthcheck",
         "path": "/status/200"
        }
       }
      ],
      "common_lb_config": {
       "healthy_panic_threshold": {}
      },
      "alt_stat_name": "default_httpbin_80",
      "ignore_health_on_host_removal": true
     },
```

internal/xds/util.go
jpeach: fixed in HCM typeswitch

internal/xds/util.go
```diff
@@ -38,3 +61,158 @@ func ClusterLoadAssignmentName(service types.NamespacedName, portName string) st
 
 	return strings.Join(name, "/")
 }
+
+// TypeMapping maps xDS type URLs from v2 to v3.
+var TypeMapping map[string]string
+
+func init() {
```
youngnick, line 68: Nit: Why do this one on `init()` rather than statically? Maybe might be worth a comment as well for future us.

youngnick (approved):
LGTM with one small comment; I think @stevesloka needs to approve too though, as I'm not familiar enough with the protos to be 100%.

internal/xds/util.go
jpeach: Originally I had something like `var TypeMapping = build_table()`. I switched to `init`, but I don't think there's much practical difference. Happy to take whichever approach is preferred :)

internal/xds/util.go
```diff
@@ -38,3 +63,161 @@ func ClusterLoadAssignmentName(service types.NamespacedName, portName string) st
...
+
+	case *envoy_api_auth_v2.UpstreamTlsContext:
+		for _, s := range msg.GetCommonTlsContext().TlsCertificateSdsSecretConfigs {
+			rewriteConfigSource(s.GetSdsConfig())
+		}
+
+		return msg
+
+	default:
+		log.Printf("missing conversion for %T, good luck", msg)
```
stevesloka, line 220: Add the context where this is logging and remove `good luck`, it's not needed.

stevesloka (changes requested):
Testing through this on my local cluster I keep getting log messages like this:

```
2020/10/30 10:42:12 missing conversion for *envoy_config_accesslog_v2.FileAccessLog, good luck
2020/10/30 10:42:12 missing conversion for *envoy_config_filter_http_lua_v2.Lua, good luck
2020/10/30 10:42:12 missing conversion for *envoy_config_filter_http_ext_authz_v2.ExtAuthz, good luck
2020/10/30 10:42:12 missing conversion for *envoy_config_accesslog_v2.FileAccessLog, good luck
2020/10/30 10:42:12 missing conversion for *envoy_config_filter_http_ext_authz_v2.ExtAuthzPerRoute, good luck
2020/10/30 10:42:12 missing conversion for *envoy_extensions_filters_network_http_connection_manager_v3.HttpConnectionManager, good luck
```

internal/xds/util.go
skriss: If we're gonna log anything here (not sure we need to once this is finalized), we should plumb a logrus logger through

internal/xds/v3/contour.go
```diff
@@ -114,15 +119,17 @@ func (s *contourServer) stream(st grpcStream) error {
...
 		}
 
-		// from the request we derive the resource to stream which have
+		// From the request we derive the resource to stream which have
 		// been registered according to the typeURL.
-		r, ok := s.resources[req.TypeUrl]
+		r, ok := s.resources[req.GetTypeUrl()]
 		if !ok {
-			return done(log, fmt.Errorf("no resource registered for typeURL %q", req.TypeUrl))
+			if !ok {
```
skriss, line 126: remove extraneous `if !ok {`

internal/xds/util.go
jpeach: I'll leave this as it for now; we will remove the code in Contour 1.11 when we are fully over to v3.

skriss (approved):
No other comments; LGTM assuming testing's OK (has seemed fine in my limited smoke-tests).

PR #3077, opened 2020-10-29
Site: Document support for custom Envoy JSON Fields

XanderStrike:
follow-up for https://github.com/projectcontour/contour/pull/3059

### DIFF
--- a/site/_guides/structured-logs.md
+++ b/site/_guides/structured-logs.md
@@ -9,26 +9,45 @@ This document describes how to configure structured logging for Envoy via Contou
 
 Contour allows you to choose from a set of JSON fields that will be expanded into Envoy templates and sent to Envoy.
 There is a default set of fields if you enable JSON logging, and you may customize which fields you log.
-Custom fields are not currently possible, however, we welcome PRs on the field list.
 
-The canonical location for the current field list is at [JSONFields][1].
-The default list of fields is available at [DefaultFields][2]
+The list of available fields are discoverable in the following objects:
+- [JSONFields][1] are fields that have built in mappings to commonly used envoy
+  operators.
+- [envoySimpleOperators][2] are the names of simple envoy operators that don't
+  require arguments, they are case-insensitive when configured.
+- [envoyComplexOperators][3] are the names of complex envoy operators that require
+  arguments.
+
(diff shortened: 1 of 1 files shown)


site/_guides/structured-logs.md
```diff
@@ -60,7 +78,8 @@ json-fields:
   - "x_forwarded_for"
 ```
 
-[1]: https://godoc.org/github.com/projectcontour/contour/internal/envoy#JSONFields
-[2]: https://godoc.org/github.com/projectcontour/contour/internal/envoy#DefaultFields
-[3]: {{site.github.repository_url}}/issues/1507
-[4]: {{site.github.repository_url}}/blob/{{site.github.latest_release.tag_name}}/examples/contour/01-contour-config.yaml
+[1]: https://github.com/projectcontour/contour/blob/main/pkg/config/accesslog.go#L33-L45
```
youngnick, line 81: I'd prefer these to point to `godoc.org` with the `#ObjectName` on the end of the link - this seems a bit fragile, and would need updating whenever we change `accesslog.go`.

youngnick (changes requested):
Mostly LGTM, but I'd really like the reference links to go to `godoc.org`, as I think that will be more stable than line numbers in Github.

site/_guides/structured-logs.md
XanderStrike: Sorry, I should've explained this decision in the PR comment, totally slipped my mind.

As a result of some feedback on the previous PR where we implemented this feature, we stopped exporting JSONFields and didn't export envoy{Simple,Complex}Operators either, as they weren't required outside the package. As a result, they don't appear on what I believe to be the [relevant godoc page](https://godoc.org/github.com/projectcontour/contour/pkg/config#pkg-variables). I figured that the possibility of the link drifting off by a few lines over time was more desirable than no links, or the dead links that were there before.

If there's another page I can point to, or some godoc configuration that will allow it to expose these variables, I'm happy to make that change.

youngnick: Ah, right, that makes total sense. Line numbers will have to do then.

youngnick (approved):
LGTM

jpeach (approved):

PR #3078, opened 2020-10-29
cmd/contour: Remove xds-server-version config parameter

stevesloka:
Remove the xds-server-version from the Contour configuration file. The original
intent was to allow users to specify the xDS resource version Contour would serve
to Envoy clients, but is now no longer needed.

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -652,32 +652,10 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 
 		switch ctx.Config.Server.XDSServerType {
 		case config.EnvoyServerType:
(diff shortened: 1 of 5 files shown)


cmd/contour/serve.go
```diff
@@ -652,32 +652,10 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 
 		switch ctx.Config.Server.XDSServerType {
```
youngnick, line 653: Non-blocking question: Is this the switch between Contour and go-control-plane? Sorry, I've missed some stuff here in the last little while.

youngnick (approved):
LGTM with one small question.

jpeach (approved):

PR #3079, opened 2020-10-29
cmd/contour: fix bootstrap xDS version default

jpeach:
Set "v2" as the xDS bootstrap default (for now) so that the bootstrapper
doesn't fail if the flag isn't given. Fix the integration test script
to populate the right image tags.


youngnick (approved):
LGTM

PR #3082, opened 2020-10-30
testsuite: add an integration test for health checking

jpeach:


youngnick (approved):
LGTM

PR #3085, opened 2020-10-30
site: Add guide on upgrading from xDS v2-->v3

### LINKED ISSUE #3047: xDS v2-->v3 Upgrade Plan

stevesloka:
Add a small guide in the docs to discuss how xDS Resource API versions are implemented
and how to upgrade from v2 to v3.

Note: This should be referenced in the upgrade guide as well once we get to the release. 

Fixes #3047

### DIFF
--- a/site/_guides/xds-migration.md
+++ b/site/_guides/xds-migration.md
@@ -1,0 +1,41 @@
+---
+title: Migrating from xDS v2 --> v3
+layout: page
+---
+
+This guide shows you how to migrate in-place instances of Envoy running xDS v2 to v3. 
+
+## Summary
+
+Envoy communicates with Contour over gRPC which allows for dynamic communication for Envoy configuration updates.
+Until Contour v1.10, this gRPC xDS communication utilized the v2 xDS for transport as well as resource version.
+
+In the beginning of Q1 20201, the Envoy community is [deprecating][0] the v2 xDS API in favor of the v3 which has been stable since Q1 2020.
+Contour offers support for the v3 xDS API in Contour v1.10.0. 
+
+Practically for users, this change has no effect on how Contour configures Envoy to route ingress traffic inside a Kubernetes cluster, however
+it's important to upgrade to this new version immediately since newer versions of Envoy won't support the `v2` api. 
+
+## Background
+
+Envoy gets configured with a boostrap configuration file which Contour provides via an `initContainer` on the Envoy daemonset.
+This file configures the dynamic xDS resources, Listener Discovery Service (LDS) and Cluster Discovery Service (CDS), to point to Contour's xDS gRPC server endpoint.
+
+The boostrap configuration file has two settings in the LDS/CDS entries which tell Contour what Resource & Transport version they would like to use. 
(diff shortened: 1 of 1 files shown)


site/_guides/xds-migration.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Background
+
+Envoy gets configured with a boostrap configuration file which Contour provides via an `initContainer` on the Envoy daemonset.
+This file configures the dynamic xDS resources, Listener Discovery Service (LDS) and Cluster Discovery Service (CDS), to point to Contour's xDS gRPC server endpoint.
+
+The boostrap configuration file has two settings in the LDS/CDS entries which tell Contour what Resource & Transport version they would like to use. 
+In Contour v1.10.0, there's a new flag, `--xds-resource-version`, on the `contour boostrap` command which is used in the `initContainer` that allows users to specify the xDS resource version.
+
+Setting this flag to `v3` will configure Envoy to request the `v3` xDS Resource API version and will eventually become the default in future Contour versions.
```
jpeach, line 27: Probably clearer to say it will be the default in 1.11.

jpeach (approved):
Just one nit, otherwise LGTM

site/_guides/xds-migration.md
```diff
@@ -0,0 +1,40 @@
...
+This file configures the dynamic xDS resources, Listener Discovery Service (LDS) and Cluster Discovery Service (CDS), to point to Contour's xDS gRPC server endpoint.
+
+The boostrap configuration file has two settings in the LDS/CDS entries which tell Contour what Resource & Transport version they would like to use. 
+In Contour v1.10.0, there's a new flag, `--xds-resource-version`, on the `contour boostrap` command which is used in the `initContainer` that allows users to specify the xDS resource version.
+
+Setting this flag to `v3` will configure Envoy to request the `v3` xDS Resource API version and will eventually become the default in future Contour versions.    
+
+## In-Place Upgrade
+
+When users have an existing Contour installation and wish to upgrade without dropping connections, Contour v1.10.0 will support this upgrade path.
```
youngnick, line 31: By "this upgrade path", do you mean changing the bootstrap? I think this wording is a little unclear. Maybe something like this?
```suggestion
When users have an existing Contour installation and wish to upgrade without dropping connections, they need only to change the Envoy Daemonset or deployment to include `--xds-resource-version=v3`. The usual rollout process will handle draining connections for you.
```

Or something. Just to make it absolutely clear, in case people skip straight to the section they care about.

youngnick (approved):
LGTM with one wording suggestion.

PR #3088, opened 2020-11-03
examples: Make xds-resource-version default to v2

stevesloka:
Support for xDS v3 has been added, however, Contour shouldn't default to that version
until v1.11.0. This gives users a chance to upgrade during the v1.10.0 cycle at their own pace.

### DIFF
--- a/examples/contour/03-envoy.yaml
+++ b/examples/contour/03-envoy.yaml
@@ -99,7 +99,7 @@ spec:
         - /config/envoy.json
         - --xds-address=contour
         - --xds-port=8001
-        - --xds-resource-version=v3
+        - --xds-resource-version=v2
         - --resources-dir=/config/resources
         - --envoy-cafile=/certs/ca.crt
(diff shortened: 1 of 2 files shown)


stevesloka:
Maybe this is the right move. My original thought was that users should first upgrade to Contour v1.10.0, the move their Envoy's over to xds-resource-version=v3. If they don't do that, however, then come v1.11.0 v2 will be gone. 

Either way, I think the docs/upgrade guide/release notes need to be super clear so this is known to users.

jpeach:
> Maybe this isn't the right move to switch back to v2. My original thought was that users should first upgrade to Contour v1.10.0, the move their Envoy's over to xds-resource-version=v3. If they don't do that, however, then come v1.11.0 v2 will be gone.

The idea was to switch to v3 in 1.10 while still keeping the v2 code. This gives sites the ability to switch back if it goes south. As you point out, from 1.11, it will be v3 only, so there is a 1 release window where you can fall back.

PR #3089, opened 2020-11-03
_integration: Nit: Update comment on how images are loaded into Kind …

stevesloka:
…cluster for intergration testing.

### DIFF
--- a/_integration/testsuite/install-contour-working.sh
+++ b/_integration/testsuite/install-contour-working.sh
@@ -64,11 +64,10 @@ for t in $TAGS ; do
     kind::cluster::load docker.io/projectcontour/contour:$t
 done
 
-# Install Contour.
+# Install Contour
 #
-# NOTE(jpeach): The certgen job uses the ":latest" tag with the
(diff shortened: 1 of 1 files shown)


_integration/testsuite/install-contour-working.sh
```diff
@@ -64,11 +64,10 @@ for t in $TAGS ; do
     kind::cluster::load docker.io/projectcontour/contour:$t
 done
 
-# Install Contour.
+# Install Contour
 #
-# NOTE(jpeach): The certgen job uses the ":latest" tag with the
-# "Latest" pull policy, which forces the kubelet to re-fetch from
-# DockerHub, which is why we have to whack the image pull policy.
+# Manifests use the "Latest" image pull policy, which forces the kubelet to re-fetch from
```
skriss, line 69:
```suggestion
# Manifests use the "Always" image pull policy, which forces the kubelet to re-fetch from
```

stevesloka: good catch! =)

skriss (approved):

PR #3092, opened 2020-11-03
internal/xdscache: only update ClusterLoadAssignment cache on changes

### LINKED ISSUE #3053: Only push ClusterLoadAssignments updates if they have changed

As reported on slack, after upgrading to 1.9 Contour is spamming the following log message:
```
level=info msg=stream_wait connection=1469 context=xds node_id=envoy-rklgl node_version=v1.15.1 resource_names="[default/<service_name>]" response_nonce=51581 type_url=type.googleapis.com/envoy.api.v2.ClusterLoadAssignment version_info=51581
```

This really should be a debug level log, but more serious is the rate of updates this indicates.

Most likely cause is the notification that is sent when the load assignment cache is rebuilt on each DAG run. The assumption
was that (1) The DAG notifications are less frequent than endpoint changes and (2) envoy would filter no-op updates with low cost.

We probably should revisit this and only notify if a DAG rebuild caused the cache to change.

skriss:
Only updates the ClusterLoadAssignment xDS cache and notifies via the
condition variable if the contents have changed during recalculation.

Closes #3053.

### DIFF
--- a/internal/xdscache/v2/endpointstranslator.go
+++ b/internal/xdscache/v2/endpointstranslator.go
@@ -298,11 +298,38 @@ func (e *EndpointsTranslator) OnChange(d *dag.DAG) {
 	// set the entries rather than merging them.
 	entries := e.cache.Recalculate()
 
-	e.mu.Lock()
-	e.entries = entries
-	e.mu.Unlock()
+	// Only update and notify if entries has changed.
+	if !equal(e.entries, entries) {
+		e.Debug("cluster load assignments changed, notifying waiters")
 
-	e.Notify()
+		e.mu.Lock()
+		e.entries = entries
+		e.mu.Unlock()
+
+		e.Notify()
+	} else {
+		e.Debug("cluster load assignments did not change")
+	}
+}
+
+// equal returns true if a and b are the same length, have the same set
+// of keys, and have proto-equivalent values for each key, or false otherwise.
+func equal(a, b map[string]*envoy_api_v2.ClusterLoadAssignment) bool {
+	if len(a) != len(b) {
+		return false
+	}
+
+	for k := range a {
+		if _, ok := b[k]; !ok {
+			return false
+		}
+
+		if !proto.Equal(a[k], b[k]) {
+			return false
+		}
+	}
+
+	return true
 }
 
 func (e *EndpointsTranslator) OnAdd(obj interface{}) {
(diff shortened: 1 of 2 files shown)


skriss:
This looked OK in my basic testing, but would appreciate any extra 👀  !

internal/xdscache/v2/endpointstranslator.go
```diff
@@ -298,11 +298,38 @@ func (e *EndpointsTranslator) OnChange(d *dag.DAG) {
...
+
+		e.Notify()
+	} else {
+		e.Debug("cluster load assignments did not change")
+	}
+}
+
+// equal returns true if a and b are the same length, have the same set
+// of keys, and have proto-equivalent values for each key, or false otherwise.
+func equal(a, b map[string]*envoy_api_v2.ClusterLoadAssignment) bool {
```
stevesloka, line 317: I was doing this with `equality.Semantic.DeepEqual` on my branch, but it wasn't working. I can test this out to see how it behaves.

internal/xdscache/v2/endpointstranslator.go
```diff
@@ -298,11 +298,38 @@ func (e *EndpointsTranslator) OnChange(d *dag.DAG) {
...
+	if len(a) != len(b) {
+		return false
+	}
+
+	for k := range a {
+		if _, ok := b[k]; !ok {
```
stevesloka, line 323: Does this assume ordering is the same each time?

internal/xdscache/v2/endpointstranslator.go
skriss: Yeah, I observed the same with that or `reflect.DeepEqual`, I think there are some private fields in there that don't match. Doing an item-by-item comparison using `proto.Equal` fixes it, though.

internal/xdscache/v2/endpointstranslator.go
skriss: Nope, no order assumptions. We're just checking each key from a to ensure it exists in b, and then comparing the values.

stevesloka (changes requested):
Could you maybe add some tests to validate the `equal` method?

internal/xdscache/v2/endpointstranslator.go
stevesloka: Nice!

internal/xdscache/v2/endpointstranslator.go
stevesloka: 🎉

skriss:
> Could you maybe add some tests to validate the `equal` method?

good call, added.

jpeach (approved):

PR #3093, opened 2020-11-03
internal: Remove max_program_size from regex

### LINKED ISSUE #2933: RegexMatcher Deprecated, needs upgraded
Seeing this deprecated log message in Envoy due to Contour's use of regex's. We should look to upgrade to resolve this so we don't end up with a breaking change in a newer version of Envoy. 

 ```
Deprecated field: type envoy.type.matcher.RegexMatcher.GoogleRE2 Using deprecated option 'envoy.type.matcher.RegexMatcher.GoogleRE2.max_program_size' from file regex.proto. This configuration will be removed from Envoy soon. Please see https://www.envoyproxy.io/docs/envoy/latest/version_history/version_history for details. If continued use of this field is absolutely necessary, see https://www.envoyproxy.io/docs/envoy/latest/configuration/operations/runtime#using-runtime-overrides-for-deprecated-features for how to apply a temporary and highly discouraged override.
```

### PR DESCRIPTION
**Pull request by stevesloka**
The Regex `max_program_size` is deprecated from Envoy and the control plane should
do this validation now. Removes the max_program_size field and adds a regex compile
check to the Ingress resource builder.

Fixes #2933

### DIFF
--- a/internal/dag/conditions.go
+++ b/internal/dag/conditions.go
@@ -179,3 +179,10 @@ func headerMatchConditionsValid(conditions []contour_api_v1.MatchCondition) erro
 
 	return nil
 }
+
+// ValidateRegex returns an error if the supplied
+// RE2 regex syntax is invalid.
+func ValidateRegex(regex string) error {
+	_, err := regexp.Compile(regex)
+	return err
+}
--- a/internal/dag/ingress_processor.go
+++ b/internal/dag/ingress_processor.go
@@ -140,7 +140,15 @@ func (p *IngressProcessor) computeIngressRule(ing *v1beta1.Ingress, rule v1beta1
 			continue
 		}
 
-		r := route(ing, path, s, clientCertSecret, p.FieldLogger)
+		r, err := route(ing, path, s, clientCertSecret, p.FieldLogger)
+		if err != nil {
+			p.WithError(err).
+				WithField("name", ing.GetName()).
+				WithField("namespace", ing.GetNamespace()).
+				WithField("regex", path).
+				Errorf("path regex is not valid")
+			return
+		}
 
 		// should we create port 80 routes for this ingress
 		if annotation.TLSRequired(ing) || annotation.HTTPAllowed(ing) {
@@ -158,7 +166,7 @@ func (p *IngressProcessor) computeIngressRule(ing *v1beta1.Ingress, rule v1beta1
 }
 
 // route builds a dag.Route for the supplied Ingress.
-func route(ingress *v1beta1.Ingress, path string, service *Service, clientCertSecret *Secret, log logrus.FieldLogger) *Route {
+func route(ingress *v1beta1.Ingress, path string, service *Service, clientCertSecret *Secret, log logrus.FieldLogger) (*Route, error) {
 	log = log.WithFields(logrus.Fields{
 		"name":      ingress.Name,
 		"namespace": ingress.Namespace,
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: internal/envoy/v2/regex_test.go [hunk 1]
```diff
@@ -29,19 +29,15 @@ func TestSafeRegexMatch(t *testing.T) {
 			regex: "",
 			want: &matcher.RegexMatcher{
 				EngineType: &matcher.RegexMatcher_GoogleRe2{
-					GoogleRe2: &matcher.RegexMatcher_GoogleRE2{
-						MaxProgramSize: protobuf.UInt32(maxRegexProgramSize),
-					},
+					GoogleRe2: &matcher.RegexMatcher_GoogleRE2{},
```
stevesloka, line 32: This whole file could be deleted probably, these tests don't really test anything.

internal/dag/conditions.go
```diff
@@ -179,3 +179,10 @@ func headerMatchConditionsValid(conditions []contour_api_v1.MatchCondition) erro
 
 	return nil
 }
+
+// ValidateRegex returns an error if the supplied
+// regex is invalid.
+func ValidateRegex(regex string) error {
```
jpeach, line 185: Nit: Best to add a comment that this is specifically validating RE2 regex syntax because that I what we are programming into Envoy (though Go isn't *identical*l apparently?).

internal/dag/ingress_processor.go
```diff
@@ -140,7 +140,15 @@ func (p *IngressProcessor) computeIngressRule(ing *v1beta1.Ingress, rule v1beta1
...
 		}
 
-		r := route(ing, path, s, clientCertSecret, p.FieldLogger)
+		r, err := route(ing, path, s, clientCertSecret, p.FieldLogger)
+		if err != nil {
+			p.WithError(err).
+				WithField("name", ing.GetName()).
+				WithField("namespace", ing.GetNamespace()).
+				WithField("regex", path).
+				Errorf("path regex is not valid: %s", err)
```
jpeach, line 149:
```suggestion
			p.WithError(err).
  				WithField("name", ing.GetName()).
  				WithField("namespace", ing.GetNamespace()).
  				WithField("regex", path).
  				Error("invalid path regular expression")
```

since the error is already logged by `WithError()`

internal/dag/ingress_processor.go
```diff
@@ -177,13 +185,18 @@ func route(ingress *v1beta1.Ingress, path string, service *Service, clientCertSe
 	}
 
 	if strings.ContainsAny(path, "^+*[]%") {
-		// path smells like a regex
+		// validate the regex
+		err := ValidateRegex(path)
+		if err != nil {
+			return nil, err
+		}
```
jpeach, line 192:
```suggestion
		if err := ValidateRegex(path); err != nil {
  			return nil, err
  		}
```

nit :)

internal/featuretests/v2/route_test.go
```diff
@@ -1509,6 +1509,48 @@ func TestRoutePrefixRouteRegex(t *testing.T) {
 		TypeUrl: routeType,
 		Nonce:   "1",
 	})
+
+	// add default/kuard to translator.
```
jpeach, line 1513: Stale comment?

jpeach (approved):
Looks good, with jus a few nits.

internal/dag/ingress_processor.go
stevesloka: 👍

PR #3094, opened 2020-11-04
testsuite: standardize HTTP checks

jpeach:
Add a new `expect` policy module to make assertions on HTTP responses
from the conformance echo server. This lets us reduce the boilerplate
needed to check HTTP responses in test documents.


stevesloka (approved):
/lgtm

PR #3095, opened 2020-11-04
internal: fix issue with duplicate DAG services & clusters

skriss:
Fixes an issue where services were being added as DAG roots, even
though they're always children of a route, so by definition not
roots. Also fixes a bug in the endpoints translator that did not
properly track already-visited service clusters to avoid processing
them twice.

### DIFF
--- a/internal/xdscache/v2/endpointstranslator.go
+++ b/internal/xdscache/v2/endpointstranslator.go
@@ -272,11 +272,11 @@ func (e *EndpointsTranslator) OnChange(d *dag.DAG) {
 			if err := svc.Validate(); err != nil {
 				e.WithError(err).Errorf("dropping invalid service cluster %q", svc.ClusterName)
 			} else if _, ok := names[svc.ClusterName]; ok {
-				e.Errorf("dropping service cluster with duplicate name %q", svc.ClusterName)
+				e.Debugf("dropping service cluster with duplicate name %q", svc.ClusterName)
 			} else {
-
 				e.Debugf("added ServiceCluster %q from DAG", svc.ClusterName)
 				clusters = append(clusters, svc.DeepCopy())
+				names[svc.ClusterName] = true
 			}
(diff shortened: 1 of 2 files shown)


internal/xdscache/v2/endpointstranslator.go
```diff
@@ -272,11 +272,11 @@ func (e *EndpointsTranslator) OnChange(d *dag.DAG) {
 			if err := svc.Validate(); err != nil {
 				e.WithError(err).Errorf("dropping invalid service cluster %q", svc.ClusterName)
 			} else if _, ok := names[svc.ClusterName]; ok {
-				e.Errorf("dropping service cluster with duplicate name %q", svc.ClusterName)
+				e.Debugf("dropping service cluster with duplicate name %q", svc.ClusterName)
```
jpeach, line 275: Maybe `Warnf` here, since we don't really expect there to be duplicate service clusters? If the names collide, we don't know whether that's the same thing multiple times or different things with the name name.

jpeach (approved):
Just a nit about the log level.

internal/xdscache/v2/endpointstranslator.go
skriss: I guess that's fine. It is unfortunate that it will generate Warn logs for valid situations, though, e.g. when two routes use the same service, so the service gets visited twice.

jpeach: Yeh, maybe debug is better then

skriss: I put it at debug; if anyone else has strong feelings LMK.

skriss:
> Shall we drop the code that adds Services as dag roots in this PR too?

Made this change as well.

jpeach (approved):

youngnick (approved):
LGTM

PR #3096, opened 2020-11-05
internal/xds: clone xDS protobufs before rewriting them

jpeach:
Each xDS resource caches the protoufs that it generates from DAG updates
and hands them out to the xDS consumers via a pointer reference. When
we rewrite these to xDS v3, we were modifying the cached values, which
would cause errors if we had to rewrite them again.

This updates #1898.

### DIFF
--- a/internal/xds/v3/contour.go
+++ b/internal/xds/v3/contour.go
@@ -151,7 +151,10 @@ func (s *contourServer) stream(st grpcStream) error {
 
 			// Rewrite the embedded message types to v3.
 			for _, r := range resources {
-				xds.Rewrite(r)
+				// Note that the resource cache
+				// gave us a pointer to its internal
+				// object, so we should rewrite a copy.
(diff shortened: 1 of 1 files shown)


youngnick (approved):
LGTM

internal/xds/v3/contour.go
```diff
@@ -151,7 +151,10 @@ func (s *contourServer) stream(st grpcStream) error {
 
 			// Rewrite the embedded message types to v3.
 			for _, r := range resources {
-				xds.Rewrite(r)
+				// Note that the resource cache
+				// gave us a pointer to its internal
+				// object, so we should rewrite a copy.
+				xds.Rewrite(proto.Clone(r))
```
skriss, line 157: don't we need to *do* something with the result of this rewrite? E.g. collect them and iterate over them in L160? Seems like as-is we're rewriting and throwing away the rewritten result.

PR #3104, opened 2020-11-06
site/docs: update links to point to current docs version

skriss:
Changes several versioned docs links from using site.latest to
page.version, so that the links remain within the same docs
version as the source page.


skriss:
turns out this needs to be part of the 1.10 docs rollout due to some other page moves that I hadn't noticed.

PR #3109, opened 2020-11-06
fix broken troubleshooting links

skriss:

We don't have a main Troubleshooting page anymore, so just link to the docs home page. I think this is fine for now; we could also add back a landing page for Troubleshooting but then we'd have to keep it up to date with an index.


stevesloka (approved):
/lgtm

PR #3098, opened 2020-11-06
Fix bunch of typos

invidian:
Found using Codespell with the following command:

codespell -S .git,*.png,*.woff,*.ttf,*.jpg,*.ico -L \
keypair,keypairs,nd,suh,od,rouge,als,wit

Signed-off-by: Mateusz Gozdek <mgozdekof@gmail.com>

PR #3110, opened 2020-11-06
Replace misspell with codespell

invidian:
**This PR includes commits from PR #3098**

As discussed in #3098, creating a PR which replaces `misspell` with `codespell`. IMO `misspell` can stay enabled in `golangci-lint`, as it may still catch some typos which won't be covered by `codespell`.

### DIFF
--- a/.codespell.ignorewords
+++ b/.codespell.ignorewords
@@ -1,0 +1,7 @@
+keypair
+keypairs
+nd
+suh
+od
+als
+wit
--- a/.codespell.skip
+++ b/.codespell.skip
@@ -1,0 +1,8 @@
+.git
+*.png
+*.woff
+*.ttf
+*.jpg
+*.ico
+./site/Gemfile.lock
+./site/_config.yml
--- a/.github/workflows/prbuild.yaml
+++ b/.github/workflows/prbuild.yaml
@@ -54,6 +54,19 @@ jobs:
         with:
           version: v1.31
           only-new-issues: true
+  codespell:
+    name: Codespell
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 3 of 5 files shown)


skriss (approved):

.codespell.ignorewords
```diff
@@ -0,0 +1,8 @@
+keypair
+keypairs
+nd
+suh
+od
+rouge
```
youngnick, line 6: Nit: Why are we ignoring this one? It's a common misspelling of `rogue`, and I don't think we are referring to the color rouge anywhere...

youngnick (approved):
LGTM with one nit.

.codespell.ignorewords
invidian: Because of:
```console
$ git grep rouge
.codespell.ignorewords:rouge
site/Gemfile.lock:      rouge (~> 3.0)
site/Gemfile.lock:    rouge (3.12.0)
site/_config.yml:highlighter: rouge
site/_config.yml:  syntax_highlighter: rouge
```

Codespell tries to correct those as following:
```
./site/Gemfile.lock:49: rouge ==> rogue
./site/Gemfile.lock:100: rouge ==> rogue
./site/_config.yml:14: rouge ==> rogue
./site/_config.yml:17: rouge ==> rogue
```

I can ignore entire `site/Gemfile.lock` and `site/_config.yml` files instead if you prefer.

stevesloka: > I can ignore entire site/Gemfile.lock and site/_config.yml files instead if you prefer.

I think that's a good move overall to ignore those config type files. 👍

invidian: Adapted the PR then.

stevesloka (approved):
/lgtm thanks @invidian!! 🎉

PR #3111, opened 2020-11-09
internal: Drop support for v2 xDS resource version

stevesloka:
Drop support for xDS v3 resource version by changing the default
boostrap resource version to v3 and removing the v2 xDS resource server.

Updates #1898


skriss:
We should update https://github.com/projectcontour/contour/blob/main/cmd/contour/bootstrap.go#L37 and the associated documentation to default to `v3` as well.

youngnick (commented):
+5, -1395, impressive!

LGTM.

stevesloka:
@skriss I updated the default bootstrap value, thanks, I missed that. Also I removed the flag from the initContainer since it's not needed anymore.

cmd/contour/bootstrap.go
```diff
@@ -34,6 +34,6 @@ func registerBootstrap(app *kingpin.Application) (*kingpin.CmdClause, *envoy.Boo
 	bootstrap.Flag("envoy-cert-file", "Client certificate filename for Envoy secure xDS gRPC communication.").Envar("ENVOY_CERT_FILE").StringVar(&config.GrpcClientCert)
 	bootstrap.Flag("envoy-key-file", "Client key filename for Envoy secure xDS gRPC communication.").Envar("ENVOY_KEY_FILE").StringVar(&config.GrpcClientKey)
 	bootstrap.Flag("namespace", "The namespace the Envoy container will run in.").Envar("CONTOUR_NAMESPACE").Default("projectcontour").StringVar(&config.Namespace)
-	bootstrap.Flag("xds-resource-version", "The versions of the xDS resources to request from Contour.").Default("v2").StringVar((*string)(&config.XDSResourceVersion))
+	bootstrap.Flag("xds-resource-version", "The versions of the xDS resources to request from Contour.").Default("v3").StringVar((*string)(&config.XDSResourceVersion))
```
skriss, line 37: so I guess we're now ignoring this flag and just automatically serving v3 only, right? I guess that's fine, maybe worth a callout in docs or help text though.

skriss (approved):
Need to update the docs for the flag too: https://github.com/projectcontour/contour/blob/main/site/docs/main/configuration.md#bootstrap-config-file

Otherwise LGTM.

cmd/contour/bootstrap.go
stevesloka: Yeah essentially the flag does nothing now until we need to support say v4.

jpeach: I'm way late to this party, but I agree that keeping the flag makes sense. Do you think that emitting an error for `--xds-resource-version=v2` is worth consideration?

PR #3112, opened 2020-11-09
internal/xdscache: change default global TLS minimum to 1.2

### LINKED ISSUE #3010: change config file's tls.minimum-protocol-version default value to 1.2
Updates #2777 

This change will globally disallow TLS 1.1 by default, even for proxies/ingresses that explicitly request it as a minimum protocol version. If the user has a true need for TLS 1.1, they will need to explicitly set this config file field to 1.1.

skriss:
Updates the default global minimum TLS version to 1.2 from 1.1.
This forces *all* HTTPProxies and Ingresses to use at least 1.2.

Closes #3010.

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -22,7 +22,6 @@ import (
 	"os"
 	"os/signal"
 	"strconv"
-	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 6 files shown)


stevesloka (approved):
/lgtm

youngnick (approved):
LGTM

PR #3114, opened 2020-11-10
internal/envoy: Enable xDS v3 Clusters

stevesloka:
Enables v3 xDS Clusters as well as supporting structs.

Updates #1898

### DIFF
--- a/internal/envoy/v3/auth.go
+++ b/internal/envoy/v3/auth.go
@@ -1,0 +1,110 @@
+// Copyright Project Contour Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


skriss:
Is it possible going forward to make a first commit that copies files from v2->v3 as-is, and then a second commit that does all the fixup to make it work for v3? I think it'll make it much easier to review since we can just look at diffs on the second commit.

skriss:
I assume you're holding off on wiring this up to the v3 xDS servers for another PR?

stevesloka:
> Is it possible going forward to make a first commit that copies files from v2->v3 as-is, and then a second commit that does all the fixup to make it work for v3? I think it'll make it much easier to review since we can just look at diffs on the second commit.

Yup I can do that now. The one bit is there are a few things partially moved due to needing them in v3 bootstrap.

stevesloka:
> The one bit is there are a few things partially moved due to needing them in v3 bootstrap.

My thought was to

1. Get the conflicting bits merged into v3
2. copy/paste other files from v2 --> v3
3. Bump up the v3 copied files into v3 version
4. Upgrade feature tests

Probably need to to do this at some smallish rate to make it easier?

PR #3120, opened 2020-11-12
internal: use string representation of TLS version

skriss:

### DIFF
--- a/internal/dag/dag.go
+++ b/internal/dag/dag.go
@@ -22,7 +22,6 @@ import (
 	"strings"
 	"time"
 
-	envoy_api_v2_auth "github.com/envoyproxy/go-control-plane/envoy/api/v2/auth"
 	contour_api_v1 "github.com/projectcontour/contour/apis/projectcontour/v1"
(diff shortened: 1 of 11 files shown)


internal/dag/dag.go
```diff
@@ -361,7 +360,7 @@ type SecureVirtualHost struct {
 	VirtualHost
 
 	// TLS minimum protocol version. Defaults to envoy_api_v2_auth.TlsParameters_TLS_AUTO
-	MinTLSVersion envoy_api_v2_auth.TlsParameters_TlsProtocol
+	MinTLSVersion string
```
stevesloka, line 363: I think this is fine and matches other types. The only thing we could think about is making this an enum then using that later on such that we wouldn't need to keep parsing strings, but could always be a later refactor.

skriss:
@stevesloka depending on which PR merges first, one of us will need to copy `internal/envoy/v2/tls.go` to v3 as well.

stevesloka (approved):
/lgtm

PR #3123, opened 2020-11-12
switch from v2 to v3 xDS caches

skriss:


skriss:
Rebased on #3124, so ignore that commit.

stevesloka (approved):
/lgtm

PR #3127, opened 2020-11-13
internal/envoy: moved to typed configs for httpfilters

### LINKED ISSUE #2479: Replace wellknown filter names with typed config

stevesloka:
Using some HTTP filter names from the go-control-plane wellknown package give warnings
due to being deprecated from Envoy.

This updates the names to use TypedConfigs which are matched differently in Envoy
and won't have the same issues with deprecation.

Fixes #2479

### DIFF
--- a/internal/envoy/v3/listener.go
+++ b/internal/envoy/v3/listener.go
@@ -23,6 +23,7 @@ import (
 	accesslog "github.com/envoyproxy/go-control-plane/envoy/config/accesslog/v3"
 	envoy_core_v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	envoy_listener_v3 "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
+	envoy_compressor_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/compressor/v3"
 	envoy_config_filter_http_ext_authz_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/ext_authz/v3"
 	lua "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/lua/v3"
 	http "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/http_connection_manager/v3"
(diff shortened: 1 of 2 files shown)


internal/envoy/v3/listener.go
```diff
@@ -202,18 +204,44 @@ func (b *httpConnectionManagerBuilder) ConnectionShutdownGracePeriod(timeout tim
 }
 
 func (b *httpConnectionManagerBuilder) DefaultFilters() *httpConnectionManagerBuilder {
+
 	b.filters = append(b.filters,
 		&http.HttpFilter{
 			Name: wellknown.Gzip,
```
skriss, line 210: Are the wellknown names no longer required? If not, we could remove them, or at least add a comment that they're not important.

internal/envoy/v3/listener.go
```diff
@@ -202,18 +204,44 @@ func (b *httpConnectionManagerBuilder) ConnectionShutdownGracePeriod(timeout tim
...
+			ConfigType: &http.HttpFilter_TypedConfig{
+				TypedConfig: protobuf.MustMarshalAny(&envoy_compressor_v3.Compressor{
+					CompressorLibrary: &envoy_core_v3.TypedExtensionConfig{
+						Name: "gzip",
+						TypedConfig: &any.Any{
+							TypeUrl: "type.googleapis.com/envoy.extensions.compression.gzip.compressor.v3.Gzip",
```
skriss, line 216: is there anywhere we can source the type URLs as constants from?

internal/envoy/v3/listener.go
stevesloka: Chatting with folks in Envoy slack, the names aren't supposed to be required I don't think, but do fail a proto constraint if you don't include. I can add a comment.

skriss:
A `go mod tidy` removes the misspell dependency entirely from `go.mod` for me

internal/envoy/v3/listener.go
stevesloka: Hmm not sure if go-control-plane has some, I could extract into constants in this file though.

internal/envoy/v3/listener.go
skriss: Just re-read the issue; I'm curious if it still works if you put an arbitrary name in there that's not the well-known name.

stevesloka: The name is not relevant now that the typed configs are used, however, there are checks like this (https://github.com/projectcontour/contour/blob/9fb3eba838f628d05b1280d899eb57fedbb3ef31/internal/envoy/v3/listener.go#L282) which check for the filter and panic if not there. 

I personally dislike this check since it causes Contour to exit and are better verified with unit tests.

stevesloka: @skriss I took out the wellknown names and now do a check on the typeurl.

skriss (approved):
LGTM pending CI

PR #3134, opened 2020-11-17
apis: Add detail subcondition type to DetailedCondition

stevesloka:
Adds a details subcodition array to the DetailedCondition allowing
for more information to be exposed to users via an HTTPProxy
Conditions struct. Things like fqdn, included paths or headers
can be written to child proxies allowing users to understand
what has been included to them from a parent resource.

Updates #2495


stevesloka:
Need thoughts on naming from comments in https://github.com/projectcontour/contour/issues/2495#issuecomment-728342415.

stevesloka:
<Hold for more discussion on implementation api>

PR #3135, opened 2020-11-17
pkg/certs: Move certgen into a public package so other projects can consume

### LINKED ISSUE #3130: Extract certgen logic into public package
Thinking about how the Operator is looking to integrate Contour (https://github.com/projectcontour/contour-operator/issues/34), we should extract the optional certgen logic which is currently used to generate self-signed certs into a public package such that the operator project can utilize without having to run binaries. This would be similar to how the config options were exposed.

stevesloka:
Some projects like the Contour Operator, need ways to create certificates for
Contour deployments. Currently, this is done via a Kubernetes Job which executes
the Contour command `certgen` which generates self-signed certificates used for
communication between Contour<>Envoy.

This moves the generation logic into a public package (e.g. pkg/certs) which
allows projects to consume this same code to generate certificates by other
means other than the Contour `certgen` command.

Fixes #3130


stevesloka:
@danehans @Miciah 

// ref https://github.com/projectcontour/contour-operator/issues/34

pkg/certs/certgen.go
```diff
@@ -0,0 +1,138 @@
...
+	"k8s.io/client-go/kubernetes"
+)
+
+// CertgenConfig holds the configuration for the certificate generation process.
+type CertgenConfig struct {
+
+	// KubeConfig is the path to the Kubeconfig file if we're not running in a cluster
+	KubeConfig string
+
+	// Incluster means that we should assume we are running in a Kubernetes cluster and work accordingly.
```
skriss, line 31:
```suggestion
	// InCluster means that we should assume we are running in a Kubernetes cluster and work accordingly.
```

pkg/certs/certgen.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// Overwrite allows certgen to overwrite any existing files or Kubernetes Secrets.
+	Overwrite bool
+
+	// Format specifies how to format the Kubernetes Secrets (must be "legacy" or "compat").
+	Format string
+}
+
+// GenerateCerts performs the actual cert generation steps and then returns the certs for the output function.
+func GenerateCerts(certConfig *CertgenConfig) (map[string][]byte, error) {
```
skriss, line 60: any reason not to combine `GenerateCerts` and `OutputCerts` into a single exported function, since they need to be called one after the other? It'd make for a slightly cleaner public API IMHO.

stevesloka: Yeah I'll do that, less API exposed is better. =)

stevesloka: In looking into this a bit more, it seems like we should have 3 public methods outlined below. The confusing part about how it's built today is that if you request a cert in `yaml` or `pem` format, you still need to pass a kubeclient which isn't used at all in those two situations. 

1. Kubernetes Secret (`--kube`), `GenerateKubeCertificates(format string, kubeclient *kubernetes.Clientset, overwrite bool)`
2. Yaml (`--yaml`), `GenerateYamlCertificates(format, outputDirectory string, force bool)`
3. PEM (`--pem`), `GeneratePemCertificates(format, outputDirectory string, force bool)`

_Note: We could combine the yaml & pem into a single one, but I'd probably keep separate? /shrug_

This means then that the `doCertgen()` inside `cmd/contour` will look at the flags passed and call each public method appropriately, but external callers have a clean API to use to code against.

jpeach: I propose just one API:
```Go

type CertificateSet struct {
    Contour []byte
    Envoy []byte
    CA []byte
}

func GenerateCertificates(options ...) (CertificateSet, error)
```

The caller can deal with emitting the certificates in the way they need (i.e. contour-operator probably can't directly use any of the specific APIs above). We probably need the options so the caller can do things like set unique SAN entries.

stevesloka: The operator can use the apis above as-is. 

The work to serialize them to pem or yaml or secrets is a benefit of using a library as this is intended so the caller doesn't need to reinvent that work.

stevesloka: There could be further customizations to add name, etc, etc, but could come as well as needed.

jpeach: > The operator can use the apis above as-is.

Are you sure? The operator doesn't need `GenerateYamlCertificates` or `GenerateYamlCertificates`. It may need `GenerateKubeCertificates`, but (1) depending on the Kubernetes framework it uses, it may not have easy access to the right client struct (2) it may need to specify the name or namespace of the secret (3) it may need to set owner references or other object meta information (4) it may combine the certificates into secrets in way this function does not anticipate.

stevesloka: Requirements / needs will change over time, 💯 .

youngnick: I like the sound of @jpeach's suggestion, it seems like a nice, clean API. I think it's probably better to have separate Output functions for each type though, so that we can ensure that things consuming this package output their certs in the right way (ie files have the correct permissions etc, certificates are in the right formats, and so on). Callers don't *have* to use the Output functions, but supplying them makes keeping things in sync easier.

stevesloka:
Ok folks, I just did a big refactor to this, please take a look.

pkg/certs/certgen.go
```diff
@@ -151,3 +244,17 @@ func serviceNames(service, namespace string) []string {
 		fmt.Sprintf("%s.%s.svc.cluster.local", service, namespace),
```
stevesloka, line 244: These assume a cluster dns of `cluster.local`, should we add more config for that?

pkg/certs/certgen.go
```diff
@@ -0,0 +1,272 @@
...
+	DefaultCertificateLifetime = 365
+
+	// DefaultNamespace where Contour is deployed. This value is added
+	// to the certificates Subject Alt Names.
+	DefaultNamespace = "projectcontour"
+
+	// DefaultDNSName holds the Kubernetes local dns suffix name
+	// specific to the cluster where Contour is deployed and is used when
+	// configuring Subject Alt Names on the certificates.
+	DefaultDNSName = "cluster.local"
```
sunjayBhatia, line 51: Worth adding a flag to make this configurable? Or should that maybe go in a separate issue/PR

stevesloka: I think that's a good separate issue. I was trying to make this super generic since it's public and harder to change after committed.

pkg/certs/certgen.go
```diff
@@ -0,0 +1,272 @@
...
+
+	// keySize sets the RSA key size to 2048 bits. This is minimum recommended size
+	// for RSA keys.
+	keySize = 2048
+)
+
+// Configuration holds config parameters used for
+// generating certificates
```
skriss, line 59:
```suggestion
// generating certificates.
```

sunjayBhatia (approved):
LGTM, looks like a solid move over of the existing code and kept to the pared down interface, with some configurability

pkg/certs/certgen.go
```diff
@@ -0,0 +1,272 @@
...
+	ContourServiceName string
+
+	// EnvoyServiceName holds the name of the Envoy service name.
+	EnvoyServiceName string
+}
+
+// Certificates contains a set of Certificates as []byte each holding
+// the CA Cert along with with Contour & Envoy Certs.
+type Certificates struct {
+	CACertificateKey      []byte
```
skriss, line 84: should this just be `CACertificate`?

pkg/certs/certgen.go
```diff
@@ -0,0 +1,272 @@
...
+	ContourCertificateKey []byte
```
skriss, line 85: should this just be `ContourCertificate`?

pkg/certs/certgen.go
```diff
@@ -0,0 +1,272 @@
...
+	ContourPrivateKey     []byte
+	EnvoyCertificateKey   []byte
```
skriss, line 87: should this just be `EnvoyCertificate`?

skriss (changes requested):
It'd be great to get @danehans to take a look at this too and validate that the public API will work for the operator's needs.

pkg/certs/certgen.go
```diff
@@ -0,0 +1,272 @@
...
+		stringOrDefault(config.EnvoyServiceName, DefaultEnvoyServiceName),
+		stringOrDefault(config.Namespace, DefaultNamespace),
+		stringOrDefault(config.DNSName, DefaultDNSName),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Certificates{
+		CACertificateKey:      caCertPEM,
```
jpeach, line 130: `caCertPEM` is a certificate, not a private key?

skriss (approved):
LGTM assuming @danehans confirms this will meet the Operator's needs.

youngnick (approved):
LGTM, nice work @stevesloka

PR #3136, opened 2020-11-17
cmd/contour: Enable retry for shutdown-manager initiating Envoy healthcheck fail

### LINKED ISSUE #2262: cmd/contour: Retry Envoy healthcheck fail
In the Envoy ShutdownManager, the first thing that happens is we tell Envoy to fail its healthchecks which causes the container to fail its readiness probe and start to drain connections.

It's possible that the initial call will fail and Envoy never starts to drain connections. If this occurs then the pod will wait it's total grace period seconds before terminating. 

A best-effort could be applied to retry the healthcheck fail call if it doesn't succeed on the first attempt. 

_Originally posted by @jpeach in https://github.com/projectcontour/contour/pull/2227#issuecomment-588566883_

stevesloka:
The call in the shutdown-manager to intitiate the Envoy healthcheck fail HTTP request
did not retry in the event that the call fails. Adds the client-go retry.OnError method
which takes a backoff object allowing for a retry with backoff in the event the request
fails.

Fixes #2262


skriss (approved):
LGTM

youngnick (approved):
LGTM

skriss:
needs rebase + CI to pass

PR #3137, opened 2020-11-17
Increase the max width of the container to 1340px

jonasrosland:
This will improve readability and remove a lot of whitespace on the site.

Signed-off-by: jonasrosland <jrosland@vmware.com>


skriss (approved):
definitely an improvement, thanks @jonasrosland!

PR #3125, opened 2020-11-13
Add a Helm option to the Getting Started guide.

georgegoh:
For users who require customization of their Contour deployment, we point
them to the Bitnami Helm Chart.

With reference to #2050 - we are pointing users towards a Bitnami hosted helm chart,
as the charts and containers are maintained, hardened and tested by the Bitnami team(see
the following links for hardening):

- https://docs.bitnami.com/tutorials/bitnami-best-practices-hardening-containers/
- https://docs.bitnami.com/tutorials/bitnami-best-practices-hardening-charts/

@youngnick @jpeach 

Signed-off-by: George Goh <gohge@vmware.com>

PR #3138, opened 2020-11-17
Adds Operator Installation to Quickstart

danehans:
Adds contour-operator to Contour's quick-start workflow and guide.

Requires: https://github.com/projectcontour/contour-operator/pull/120

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -26,10 +26,46 @@
   status = 302
 
 # Redirect versioned quickstarts so that they can easily be referenced by
-# operators or for upgrade testing.
+# users or for upgrade testing.
 #
 # kubectl apply https://projectcontour.io/quickstart/v1.4.0/contour.yaml
 [[redirects]]
   from = "/quickstart/*/contour.yaml"
   to = "https://raw.githubusercontent.com/projectcontour/contour/:splat/examples/render/contour.yaml"
   status = 302
+
+# Redirect /quickstart/operator.yaml to the deployment that matches :latest.
+#
+# kubectl apply https://projectcontour.io/quickstart/operator.yaml
+[[redirects]]
+from = "/quickstart/operator.yaml"
+# TODO [danehans]: Update main -> v1.11.0 when operator v1alpha1 is tagged.
+to = "https://raw.githubusercontent.com/projectcontour/contour-operator/main/examples/operator/operator.yaml"
+status = 302
+
+# Redirect versioned quickstarts so that they can easily be referenced by
+# users or for upgrade testing.
+#
+# kubectl apply https://projectcontour.io/quickstart/v1.11.0/operator.yaml
+[[redirects]]
+from = "/quickstart/*/operator.yaml"
+to = "https://raw.githubusercontent.com/projectcontour/contour-operator/:splat/examples/operator/operator.yaml"
+status = 302
+
+# Redirect /quickstart/contour-custom-resource.yaml to the Contour custom resource that matches :latest.
+#
+# kubectl apply https://projectcontour.io/quickstart/contour-custom-resource.yaml
+[[redirects]]
+from = "/quickstart/contour-custom-resource.yaml"
+# TODO [danehans]: Update main -> v1.11.0 when operator v1alpha1 is tagged.
+to = "https://raw.githubusercontent.com/projectcontour/contour-operator/main/examples/contour/contour.yaml"
+status = 302
+
+# Redirect versioned quickstarts so that they can easily be referenced by
+# users or for upgrade testing.
+#
+# kubectl apply https://projectcontour.io/quickstart/v1.11.0/contour-custom-resource.yaml
+[[redirects]]
+from = "/quickstart/*/contour-custom-resource.yaml"
+to = "https://raw.githubusercontent.com/projectcontour/contour-operator/:splat/examples/contour/contour.yaml"
+status = 302
--- a/site/getting-started.md
+++ b/site/getting-started.md
@@ -17,15 +17,41 @@ Before you start you will need:
 
 ## Add Contour to your cluster
 
-Run:
+### Option 1
+
+Deploy Contour using the rendered manifest:
 
 ```bash
 $ kubectl apply -f {{ site.url }}/quickstart/contour.yaml
 ```
 
+### Option 2
+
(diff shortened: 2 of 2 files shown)


danehans:
The codegen job will fail with `Failed to get the operator YAML document from https://raw.githubusercontent.com/projectcontour/contour-operator/main/examples/operator/operator.yaml.` until https://github.com/projectcontour/contour-operator/pull/120 merges.

stevesloka (changes requested):
The updates to the docs on how to deploy Contour with the operator are great, but I think the scripts to generate the operator manifests should live in the operator repo.

danehans:
@stevesloka https://github.com/projectcontour/contour-operator/pull/120 will ensure the rendered operator manifest are always up-to-date in projectcontour/contour-operator. To support the user experience presented in this PR (equivalent to the quickstart `kubectl apply -f https://projectcontour.io/quickstart/contour.yaml`), I would expect the rendered operator and Contour CR manifests must reside in `projectcontour/contour/examples/rendered`. If so, I would expect some automation, as presented in tis PR, to ensure these manifests are kept up-to-date in projectcontour/contour. Can `https://projectcontour.io/quickstart/operator.yaml` and `https://projectcontour.io/quickstart/operator-cr.yaml` point to a URL in projectcontour/contour-operator containing the rendered manifests? This would be the preferred method so we don't need to keep the rendered manifests up-to-date in multiple repos.

youngnick:
Sorry @danehans, I'm not sure what you're asking for in that last comment?

I think it's fine to point `https://projectcontour.io/quickstart/operator.yaml` and `https://projectcontour.io/quickstart/operator-cr.yaml` to the locations in the operator repo, we just need to change the `netlify.toml` file in the repo root.

Doing that implies that:
- we cut releases at the same time for both (we've agreed this already)
- we'll need to figure out how the operator URLs that the quickstart points to get changed for its release.

youngnick:
Needs a rebase please.

danehans:
@youngnick thanks for the pointer in https://github.com/projectcontour/contour/pull/3138#issuecomment-732475783. Commit `2e093bc` implements your guidance, PTAL.

> we'll need to figure out how the operator URLs that the quickstart points to get changed for its release.

The URLs will be updated to reference the release branch as part of https://github.com/projectcontour/contour-operator/issues/134. For example:

```
[[redirects]]
from = "/quickstart/operator.yaml"
to = "https://raw.githubusercontent.com/projectcontour/contour-operator/main/examples/operator/operator.yaml"
status = 302
```
will be updated to:
```
[[redirects]]
from = "/quickstart/operator.yaml"
to = "https://raw.githubusercontent.com/projectcontour/contour-operator/release-1.11/examples/operator/operator.yaml"
status = 302
```

The `operator.yaml` rendered example manifest is managed by the make [example](https://github.com/projectcontour/contour-operator/blob/main/Makefile#L112-L115) target. `contour.yaml` is static, but will be managed the same way in the future.

jpeach:
What does the name `operator-cr` mean? Can we make it full words so it's obvious?

youngnick:
I think maybe `operator-crd` may be clearer, since that YAML just contains the CRD definitions. `operator-customresourcedefinitions` seems to defeat the point of a short link. 😛

netlify.toml
```diff
@@ -26,10 +26,46 @@
...
+[[redirects]]
+from = "/quickstart/*/operator.yaml"
+to = "https://raw.githubusercontent.com/projectcontour/contour-operator/:splat/examples/operator/operator.yaml"
+status = 302
+
+# Redirect /quickstart/operator-cr.yaml to the deployment that matches :latest.
+#
+# kubectl apply https://projectcontour.io/quickstart/operator-cr.yaml
+[[redirects]]
+from = "/quickstart/operator-cr.yaml"
```
youngnick, line 59: I think that `operator-crd.yaml` is a bit clearer for the cost of one character.

netlify.toml
```diff
@@ -26,10 +26,46 @@
...
+# TODO [danehans]: Update main -> release-1.11 when operator v1alpha1 is tagged.
```
youngnick, line 60: I thought we agreed that the tags would be the same across repos? So the operator release tags should be v1.11.0, the same as Contour?

site/getting-started.md
```diff
@@ -17,15 +17,39 @@ Before you start you will need:
...
-Run:
+### Option 1
+
+Deploy Contour using the rendered manifest:
 
 ```bash
 $ kubectl apply -f {{ site.url }}/quickstart/contour.yaml
 ```
 
+### Option 2
```
youngnick, line 28: Is the operator still marked experimental/not-GA? If so, we probably should call it out here, until we are ready to call the operator GA.

site/getting-started.md
```diff
@@ -17,15 +17,39 @@ Before you start you will need:
...
+
```
youngnick, line 29: Also note that we have guidance for using Helm incoming as well in #3125, thanks to @georgegoh.

youngnick (changes requested):
LGTM aside from a few small things.

netlify.toml
danehans: You're right, I was using the release branch instead of the release tag in the comment. I'll push a new commit with the comment corrected.

danehans:
@youngnick commit `bfc1531` resolves your feedback, PTAL.

danehans:
> I think maybe operator-crd may be clearer, since that YAML just contains the CRD definitions.

@youngnick the `Contour` CRD is installed as part of "/quickstart/operator.yaml" and not "/quickstart/operator-cr.yaml", see [here](https://github.com/projectcontour/contour-operator/blob/main/examples/operator/operator.yaml#L8-L134) for details. "/quickstart/operator-cr.yaml" is an instance of a `Contour` custom resource. Commit `bfc1531` updates "/quickstart/operator-cr.yaml" to "/quickstart/contour-custom-resource.yaml".

youngnick (approved):
LGTM

youngnick:
Thanks for the explanation as well @danehans, should have read more, sigh.

youngnick:
The site-proof check is showing as failing, but this is the recurring issue with it and external links. The listed link works fine.

Issue #2019, opened 2019-12-09
internal/builder: One bad route should not break other routes


stevesloka:
I have an HTTPProxy which I'll share below. One route in the proxy is invalid since the port in the HTTPProxy route does not match the service it's configured to. Contour shows the proper status on that proxy, however, since one route is bad the entire proxy is thrown out causing all other routes to fail. 

Since this route is part of my "root" proxy, it took down my entire site. =)

Root Proxy:
```
apiVersion: projectcontour.io/v1
kind: HTTPProxy
metadata:
  name: root
  namespace: projectcontour-roots
spec:
  includes:
  - conditions:
    - prefix: /blog
    - header:
        name: User-Agent
        notcontains: Chrome
    - header:
        name: User-Agent
        notcontains: Firefox
    name: blogsite
    namespace: projectcontour-marketing
  routes:
  - conditions:
    - prefix: /get
    services:
    - name: external
      port: 80  # <---- Invalid port!
  - conditions:
    - prefix: /secure
    services:
    - name: secureapp-default
      port: 80
  - conditions:
    - prefix: /secure
    - header:
        contains: Chrome
        name: User-Agent
    services:
    - name: secureapp
      port: 80
  virtualhost:
    fqdn: containersteve.com
    tls:
      secretName: containersteve-secret
status:
  currentStatus: invalid
  description: Service [external:80] is invalid or missing
```

```
$ kubectl get proxy -A                                                                                                                                                                                                  
NAMESPACE                  NAME       FQDN                 TLS SECRET              STATUS    STATUS DESCRIPTION
projectcontour-marketing   blogsite                                                valid     valid HTTPProxy
projectcontour-marketing   infosite                                                valid     valid HTTPProxy
projectcontour-roots       root       containersteve.com   containersteve-secret   invalid   Service [external:80] is invalid or missing
```

davecheney:
/cc @jpeach @youngnick 

I don't have a position on this; we've gone back and forth on what constitutes invalid deliberately and accidentally a few times in the past.

My ask is:

- If the decision is to continue on invalid route, what is the status of the object? Is it in invalid? Likely not as its (partially) working. If so, how do we communicate this list of _warnings_ to users?
- Consistency; if a httpproxy is not invalid because once of its routes is missing a service, what other sets of things are we consider to be non fatal errors? This list should be enumerated up front before any code is cut.

davecheney:
removing the bug tag, this is working as intended -- although the intention is the point of debate.

[...]

Issue #2206, opened 2020-02-07
The "includes" function in HTTPProxy causes bad life-cycle management


uablrek:
## The "includes" function in HTTPProxy causes bad life-cycle management

The "includes" can be used to delegate route definitions to sub-objects;

```
apiVersion: projectcontour.io/v1
kind: HTTPProxy
metadata:
  name: kahttp-default
spec:
  virtualhost:
    fqdn: kahttp.com
    tls:
      secretName: contour-secret
  includes:
    - name: kahttp-default
    - name: kahttp-admin
```

In a larger site where route definitions are updates regularly this can be used to allow applications to provide their own route definition and add them to the virtual host when loaded and remove them when un-loaded.

**The problem is that the "top" object containing the vhost and the "includes" array must be updated for these operations.**

This update operation of the top object requires coordination between otherwise independent applications. To update the top object manually is not feasible in a larger sites.

### Use Case

As an operator I want to be able to deploy a applications that adds a route definitions to a vhost with a simple install, e.g. with "helm". When an application is removed it's route definition shall be removed automatically.

As an application owner I want to use "secure backend" to protect my traffic.

### Comparison with the Ingress object

K8s allows "Ingress" objects to specify the same vhost in multiple instances;

```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: kahttp-default
spec:
  tls:
  - hosts:
    - kahttp.com
    secretName: contour-secret
  rules:
  - host: kahttp.com
    http:
      paths:
      - path: /
        backend:
          serviceName: kahttp-ipv4
          servicePort: 80
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: kahttp-admin
spec:
  tls:
  - hosts:
    - kahttp.com
    secretName: contour-secret
  rules:
  - host: kahttp.com
    http:
      paths:
      - path: /admin
        backend:
          serviceName: kahttp-admin
          servicePort: 80
```

Contour handles this nicely. An "admin" application can be deployed (and removed) independently and the route to "/admin" is updated automatically. The vhost can set with a value/parameter in a helm install.

But Contour supports backend encryption only for `HTTPProxy`.

### Proposal

Allow to specify the relation in the sub-objects, example;

```
apiVersion: projectcontour.io/v1
kind: HTTPProxy
metadata:
  name: kahttp-admin
spec:
  virtualhost:
    from:
	- name: kahttp-default
	  namespace: default
  routes:
    - conditions:
[...]

Issue #2189, opened 2020-01-31
First-write-wins conflict resolution for root hostnames


rosenhouse:
## problem statement
As a cluster operator, I want to enable namespace admins to "claim" a root hostname for themselves in a fully self-service way, without my involvement.  I also want to prevent them from creating conflicting claims.  Instead, I'd prefer that the first `HTTPProxy` to "claim" a particular hostname "wins" and no others may be created for that root hostname.

[Earlier slack convo about this](https://kubernetes.slack.com/archives/C8XRH2R4J/p1580365808224600)

## current behavior
- If one `HTTPProxy` has a root fqdn of `inventory.example.com` and then someone writes a different `HTTPProxy` for the same fqdn, then it breaks both:
  ```
   fqdn "inventory.example.com" is used in multiple HTTPProxies: mobile-team/inventory, warehouse-team/inventory'
  ```
- If an operator specifies `--root-namespaces` to restrict the creation of root `HTTPProxy` resources, then namespace admins can't set up their own routes anymore.  That prevents the desired "self-service" workflow

## possible solution
Admission controller?

stevesloka:
OPA might also be a way to solve this. If something restricts only allowing a single root to exist within the cluster, there wouldn't be any changes to make to Contour.

rosenhouse:
Wow, good pointer!  That is, in fact, an example in their documentation: https://www.openpolicyagent.org/docs/v0.12.2/kubernetes-admission-control/#6-modify-the-policy-and-exercise-the-changes
which points at
https://github.com/FairwindsOps/charts/blob/79632a01090ecc65e838df8ab2835bb2179a3ce4/incubator/opa/policies/ingress-conflicts.rego

Would it make sense then for Contour to document an equivalent policy for the HTTPProxy resource?

rosenhouse:
Ok, so after looking at this a bit, and a pointer from @cppforlife, I'm concerned that this usage of the OPA Admission Controller (as documented in my previous comment) would suffer from a race-condition, in which two conflicting ingress resources that are submitted simultaneously could both be accepted.

The OPA maintainers seem to acknowledge this risk in [a brief slack chat we had](https://openpolicyagent.slack.com/archives/C1H19LW4F/p1580952581157900?thread_ts=1580951074.156600&cid=C1H19LW4F).  I opened a [PR to the OPA docs](https://github.com/open-policy-agent/opa/pull/2064) to call this out.

I don't know how likely this issue is to show up in practice.

[...]

PR #3145, opened 2020-11-20
design: Design for HTTPProxy Conflict Resolution

stevesloka:
Updates: #2019 #3039 #3071


design/httpproxy-conflict.md
```diff
@@ -0,0 +1,82 @@
...
+Contour will set the error or warning condition when an error is encountered, but do its best to still serve valid configurations.
+
+- If an invalid route is encountered, Contour will return a `404` HTTP status code to the requester
+- If an invalid service is encountered,Contour will return a`503` HTTP status code to the requester
+
+It's difficult to find errors and block them from breaking valid configurations due to how Contour currently processes objects.
+Contour takes a passive look to the cluster by reacting to changes after they have changed. 
+Due to this, Contour cannot block changes to resources before they are committed like an `Admission Controller` would be able to do (See alternatives: `HTTPProxyInstance`)
+
+## Detailed Design
```
tsaarni, line 36: Should the usability aspect be discussed here, I've had following concerns in my mind:

* When the `HTTPProxy` handles traffic for the (valid) part of the configuration, the configuration errors that were previously very visible will become warnings and might be more difficult to notice.
* Previously the user got detailed error message (though only for the first error). After turning invalid routes/services to warnings, the final error message will be something general like "all routes are invalid". The details must be looked from warnings.  This might seem especially strange for the user when there was just single route in the `HTTPProxy`.

design/httpproxy-conflict.md
```diff
@@ -0,0 +1,82 @@
+# HTTPProxy Conflict Resolution Design
+
+Status: Draft
+
+## Abstract
+
+Configuration errors in HTTPProxy resources can result in objects being marked as invalid and conversely will not serve traffic.
```
skriss, line 7:
```suggestion
Configuration errors in HTTPProxy resources can result in objects being marked as invalid and consequently will not serve traffic.
```

design/httpproxy-conflict.md
```diff
@@ -0,0 +1,82 @@
...
+Due to this, Contour cannot block changes to resources before they are committed like an `Admission Controller` would be able to do (See alternatives: `HTTPProxyInstance`)
+
+## Detailed Design
+
+### Invalid Include
+Contour utilizes an inclusion model to allow for proper route/header delegation across namespaces and teams within a Kubernetes cluster.
+If an include does not have a matching HTTPProxy.Name or HTTPProxy.Namespace, then Contour will set an error in the Conditions of the object.
+Any requests matching the `Conditions` on the `Spec.Include` will return a `404` to the requester.
+
+If the include was valid previously and is now no longer valid, the routes will no longer be valid which the include previously enabled.
```
skriss, line 43:
```suggestion
If the include was valid previously and is now no longer valid, the routes which the include previously enabled will no longer be valid.
```

skriss, line 43: So any routes previously defined on the included proxy would now return a `404`, right?

design/httpproxy-conflict.md
```diff
@@ -0,0 +1,82 @@
...
+
+It's difficult to find errors and block them from breaking valid configurations due to how Contour currently processes objects.
+Contour takes a passive look to the cluster by reacting to changes after they have changed. 
+Due to this, Contour cannot block changes to resources before they are committed like an `Admission Controller` would be able to do (See alternatives: `HTTPProxyInstance`)
+
+## Detailed Design
+
+### Invalid Include
+Contour utilizes an inclusion model to allow for proper route/header delegation across namespaces and teams within a Kubernetes cluster.
+If an include does not have a matching HTTPProxy.Name or HTTPProxy.Namespace, then Contour will set an error in the Conditions of the object.
```
skriss, line 40: If we report an error in the conditions, then will the HTTPProxy be marked as invalid, i.e. the `Valid` condition would be set to `false`? To me, that's a little misleading, since parts of it are valid and being programmed with Envoy. It seems like we would need to move to a non-binary model of validity in the status.

skriss (changes requested):
I definitely like the idea of limiting the blast radius of config errors, and using 503's for missing services and 404's for invalid route configs seems to make sense.

design/httpproxy-conflict.md
```diff
@@ -0,0 +1,82 @@
...
+When the first error or warning is detected, these fields are updated and in the status of the HTTPProxy object, however if there are more errors, they are not all shown to the user, only the first.
+There are ongoing efforts to change this logic to inform of all the errors or warnings dealing with a specific object.
+
+Contour currently invalidates the entire resource if a configuration error is encountered which can cause downtime for specific routes. 
+
+## Goals
+- Serve valid traffic configurations if portions of the spec are invalid
+
+## Non Goals
+- Surface a complete set errors or warnings for a resource (this is a different issue)
```
youngnick, line 24: I think it's important to ensure that we surface which parts are invalid though.

design/httpproxy-conflict.md
youngnick: I agree that the usability should be discussed here - I think that we'll need to make it easier for people to get all the conditions for an object somehow. Some experiments I did earlier didn't show any easy way to summarise a list of things into a field using the jsonPath tools that the CRD markup allows, but I'd love to be proved wrong on this one.

design/httpproxy-conflict.md
youngnick: Yes, the current behavior is that there is one Condition (`Valid`), that is `status: true` as long as there are *zero* errors in the `errors` list. What this change seems to imply to me is that missing routes or missing services would become `warnings`, and serve status codes instead, *without* impacting the validity of the HTTPProxy. Note that it's not possible to mark a condition as partially true - the only valid values for `status` are `true`, `false`, and `unknown` (which is set by convention to mean "not reconciled by owning controller yet".)

We *can* extend the meaning of the top-level validity field to summarise things in a different way - it currently may be `valid`, `invalid`, or `orphaned`. We could add another summary entry to *that* field that would mean `valid, but with problems`. That does muddy up the principle that the Conditions are the place to look, but may be an acceptable compromise while the tooling for Conditions is still not as good as the tooling for showing a single field in the `kubectl get` output.

design/httpproxy-conflict.md
youngnick: I agree that this design document should be very specific and exhaustive about these cases.

design/httpproxy-conflict.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Invalid Include
+Contour utilizes an inclusion model to allow for proper route/header delegation across namespaces and teams within a Kubernetes cluster.
+If an include does not have a matching HTTPProxy.Name or HTTPProxy.Namespace, then Contour will set an error in the Conditions of the object.
+Any requests matching the `Conditions` on the `Spec.Include` will return a `404` to the requester.
+
+If the include was valid previously and is now no longer valid, the routes will no longer be valid which the include previously enabled.
+
+The logic described above should only apply if the object has a proper delegation chain to the defined `Conditions` on the include. 
+If an object creates an `Include` to values that it does not have permission to, then it shouldn't create `404` responses.
```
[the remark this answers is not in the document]
youngnick: I agree that "access denied" should be treated the same as "object not found" in general terms, preferably with a status message somewhere that tells you it can't be found because access denied.

It would be possible to use this feature for object enumeration, by trying various names and seeing which ones are "not found" and which are "access denied", but that seems like an okay price to pay - I don't think that knowing the names of objects in a namespace you don't have access to is an unacceptable risk.

design/httpproxy-conflict.md
```diff
@@ -0,0 +1,82 @@
...
+For the case where only a single service is referenced from a `Spec.Route` of an HTTPProxy, Contour will return a `503` status code and also set an error in the Conditions of the object.
+
+#### Multiple Services
+For the case where multiple services are referenced from a `Spec.Route` of an HTTPProxy, Contour will set the invalid service weight to be `zero` so that no traffic is routed to it and will also set an error in the Conditions of the object.
+
+## Alternatives Considered
+
+### Admission Controller
+An admission controller could be utilized to do these checks before the object is written to API.
+This would require a way to look at a DAG to verify status information and would require further thought into how to implement but would provide the best user experience.
```
youngnick, line 70: An Admission Controller will require the Admission controller to contain a full copy of the DAG, and run the same DAG process on admission, I think.

youngnick (changes requested):
I think overall this looks good, but in this one, the devil is in the details, and I think the design needs some more specificity about how classes of error cases will be handled.

design/httpproxy-conflict.md
stevesloka: We can try and summarize the bits into a single line, but I thought the idea was folks would need to now dig into the object to figure out what's wrong, or we build some sort of plugin/octant plugin, etc to assist.

design/httpproxy-conflict.md
stevesloka: It feels to me that either it's working or it not, meaning, there's not a case where if Contour finds an issue, you're going to ignore it and move on. 

It might be too strict to define the entire HTTPProxy object as "valid, invalid, or orphaned". But it does give signal to the user that something is wrong and it needs to be investigated.

design/httpproxy-conflict.md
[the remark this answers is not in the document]
stevesloka: I removed this, I was pulling in IngressRoute logic for some reason. Anyone who can create an Include will only be able to do so at any level they have access to, do there's no way for someone to request a path they don't have access to.

design/httpproxy-conflict.md
stevesloka: Contour could also expose an endpoint that is called and asked, "hey is this valid" or something as well to keep the logic in one binary.

stevesloka:
I did another pass on updating this, mostly by adding tables for what Contour should do in various places. The big thing I was trying to figure out is when something is an `Error` vs a `Warning`. Errors change the response to the requestor, warnings do not.

design/httpproxy-conflict.md
```diff
@@ -0,0 +1,110 @@
...
+#### Service Summary
+
+| Category           | Issue            | Request Response       | Conditions |
+| ------------------ | ---------------- | ---------------------- | ---------- |
+| Spec.Route.Service | Missing Service  | HTTP 503               | Error      |
+| Spec.Route.Service | Invalid Protocol | HTTP 503               | Error      |
+| Spec.Route.Service | Port Mismatch    | HTTP 503               | Error      |
+| Spec.Route.Service.RequestHeadersPolicy | Invalid Header  | Header Not Configured | Warning    |
+| Spec.Route.Service.ResponseHeadersPolicy | Invalid Header  | Header Not Configured | Warning    |
+| Spec.Route.Service.UpstreamValidation | Invalid CACertificate | HTTP 503 | Error    |
```
jpeach, line 90: I think 502 (bad gateway) is slightly more apt here.

jpeach (approved):
To my reading this looks like it's in good shape for implementation. I'm approving (for me) but I'll hold off merging while discussions are continuing.

stevesloka:
I'm stuck on this now thinking about it over the night. It seems bad to have users making requests to an application, be exposed to bad configuration inside the cluster. 

I'm going to pause on this for the moment and think through an admission controller. I think that's a cleaner approach vs reengineering the internals of Contour (i.e. the DAG).

skriss:
> I'm stuck on this now thinking about it over the night. It seems bad to have users making requests to an application, be exposed to bad configuration inside the cluster.
> 
> I'm going to pause on this for the moment and think through an admission controller. I think that's a cleaner approach vs reengineering the internals of Contour (i.e. the DAG).

I think returning clear errors to the user is an improvement over (a) removing the entire proxy's Envoy config or (b) having the traffic potentially end up at an incorrect destination (i.e. to a route with a shorter prefix). So I still think this work is valuable, even if ultimately we want to version config or have an admission controller that's DAG-aware. Just my 2c.

stevesloka:
Yup I agree with you @skriss, just felt weird to throw errors like this back to users. Let me clean this up again and at least fix this issue, like you said, can add the other optimizations later without issue.

design/httpproxy-conflict.md
skriss: It'd be nice to be able to differentiate between "the proxy was not configured at all in Envoy" vs. "the proxy was partially configured in Envoy, with these exceptions" vs. "the proxy was fully configured in Envoy".

jpeach:
> > I'm stuck on this now thinking about it over the night. It seems bad to have users making requests to an application, be exposed to bad configuration inside the cluster.
> > I'm going to pause on this for the moment and think through an admission controller. I think that's a cleaner approach vs reengineering the internals of Contour (i.e. the DAG).
> 
> I think returning clear errors to the user is an improvement over (a) removing the entire proxy's Envoy config or (b) having the traffic potentially end up at an incorrect destination (i.e. to a route with a shorter prefix). So I still think this work is valuable, even if ultimately we want to version config or have an admission controller that's DAG-aware. Just my 2c.

In the admission controller approach, how can you deal with linked changes? For example, renaming a child proxy would require a linked change to the parent. If admission control enforces validation on includes you would need to duplicate the child proxy to the new name (as an orphan), then change the parent. That kind of workflow is pretty awkward.

design/httpproxy-conflict.md
```diff
@@ -0,0 +1,110 @@
...
+- Surface a complete set errors or warnings for a resource (this is a different issue)
+
+## High-Level Design
+Contour will set the error or warning condition when an error is encountered, but do its best to still serve valid configurations.
+
+It's difficult to find errors and block them from breaking valid configurations due to how Contour currently processes objects.
+Contour takes a passive look to the cluster by reacting to changes after they have changed. 
+Due to this, Contour cannot block changes to resources before they are committed like an `Admission Controller` would be able to do (See alternatives: `HTTPProxyInstance`)
+
+In general, Contour will set an object to be status `Error` if the request response to the user is changed from what is configured.
```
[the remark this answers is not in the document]
stevesloka: thanks updated with your suggestion as well as a little bit more cleanup.

design/httpproxy-conflict.md
```diff
@@ -0,0 +1,110 @@
...
+Any requests matching the `Conditions` on the `Spec.Include` will return a `404` to the requester.
+
+If the include was valid previously and is now no longer valid, the routes which the include previously enabled will no longer be valid.
+All requests to the path specified in the `spec.conditions.prefix` will return a `404` HTTP status code. 
+
+_Note: This is a use-case of an alternative model of implementing an `HTTPProxyInstance` resource which gives users a "last known good configuration" that can be reverted back to or an `Admission Controller`._
+
+#### Include Summary
+
+| Category           | Issue               | Request Response                               | Conditions |
```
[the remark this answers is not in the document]
stevesloka: The response a client would get from the result of a request. 

e.g. GET /blog ---> responds with 502, etc

design/httpproxy-conflict.md
sunjayBhatia: Maybe enough to mention a specific other issue URL here?

design/httpproxy-conflict.md
```diff
@@ -0,0 +1,110 @@
...
+| ------------------ | ------------------- | ---------------------------------------------- | ---------- |
+| Spec.Route.Conditions | Invalid `prefix` | HTTP 404                                       | Error      |
+| Spec.Route.Conditions | Invalid `header` | HTTP 404                                       | Error      |
+| Spec.Route.TimeoutPolicy | Invalid       | Request routes without timeoutpolicy set       | Warning      |
+| Spec.Route.RetryPolicy | Invalid       | Request routes without retrypolicy set       | Warning      |
+| Spec.Route.HTTPHealthCheckPolicy | Invalid       | Request routes without healthcheckpolicy set       | Warning      |
+| Spec.Route.LoadBalancerPolicy | Invalid       | Request routes without loadbalancerpolicy set       | Warning      |
+| Spec.Route.PathRewritePolicy | Invalid       | HTTP 404       | Error      |
```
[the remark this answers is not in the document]
stevesloka: I updated to 502.

design/httpproxy-conflict.md
stevesloka: 👍

design/httpproxy-conflict.md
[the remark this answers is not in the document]
sunjayBhatia: Not too bad to figure out but a small nit would maybe be to clarify some personas at the top of this doc to be clear who will look for/see certain types of errors, user -> someone sending requests to apps that contour configures routes etc. for and expected to see HTTP error codes, developer -> expected to be monitoring status/conditions on resources they create AND respond to user complaints about HTTP error codes (so they kinda have to know both sides)

maybe persona definitions are something to add as a section at the top of the design template?

design/httpproxy-conflict.md
```diff
@@ -0,0 +1,110 @@
...
+Status: Draft
+
+## Abstract
+
+Configuration errors in HTTPProxy resources can result in objects being marked as invalid and consequently will not serve traffic.
+These errors can come from invalid route configuration, missing service references, or invalid includes to other resources. 
+
+## Background
+HTTPProxy is designed as a multi-team aware Ingress that reduces annotation sprawl by providing a sensible home for configuration items.
+Additionally, HTTPProxy offers a feedback loop to users via a `Status` field on the object.
```
sunjayBhatia, line 12: nit: related to my other comment about personas, here `users` is a bit overloaded?

design/httpproxy-conflict.md
stevesloka: > I think it's important to ensure that we surface which parts are invalid though.
Yup 100%, that won't go away.

The point of this statement was to say that this change isn't going to do what I think https://github.com/projectcontour/contour/issues/2495 was intending to do. 

Today we show a single error with HTTPProxy and bubble that back up. Conditions were added which allows for multiple errors/warnings to be set on the object. 

My original thought was that I didn't want to tackle that work in this design, but possibly it might need to since it will require a change to the objects are processed.

youngnick (changes requested):
Overall this looks really good, and is really close. I think that it needs two things though, to help tighten it up and ensure that there's no confusion.

Firstly, I think that there needs to be some explanation at the beginning about who a user is in this context. I think that it's important to note that the person who will see the errors and warnings is the user of the **HTTPProxy**, not the user of the **application**. The user of the **application** will see HTTP error codes only. Which is fine, but we need to be clear here, so that we can be building something that does exactly what we expect.

Secondly, we need to be clear that when we say "error condition" and "warning condition", we mean Error and Warning SubConditions on the main `Valid` Condition, not adding new DetailedConditoins at the top level. Also, a quick refresher on the behavior of the top-level Valid condition is relevant. (that is, `Valid` is `status: true` as long as there's zero `Errors`, or else `status: false`.)

The detailed design  in this proposal is excellent, thanks for the great work @stevesloka !

design/httpproxy-conflict.md
```diff
@@ -0,0 +1,110 @@
+# HTTPProxy Conflict Resolution Design
+
+Status: Draft
+
+## Abstract
+
+Configuration errors in HTTPProxy resources can result in objects being marked as invalid and consequently will not serve traffic.
```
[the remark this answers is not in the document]
youngnick: One additional thing I'd like here to make this a better record: What are you suggesting doing about it in this document? Just a short summary, it's an abstract. Something like "To fix this, Contour will treat errors and warnings differently and, where possible, serve HTTP error codes for partially valid configurations".

design/httpproxy-conflict.md
youngnick: Nit: Going by the new Personas section, this should be updated to ResourceOwner? Or, if you don't want to do that until you've defined it, maybe "HTTPProxy users" or HTTPProxy Owners would be more correct.

design/httpproxy-conflict.md
```diff
@@ -0,0 +1,110 @@
...
+Additionally, HTTPProxy offers a feedback loop to users via a `Status` field on the object.
+Contour has further extended this to add `Conditions` which allows for a set of `Errors` and `Warnings` which inform users of errors in the HTTPProxy resource.
+
+When the first error or warning is detected, these fields are updated and in the status of the HTTPProxy object, however if there are more errors, they are not all shown to the user, only the first.
+There are ongoing efforts to change this logic to inform of all the errors or warnings dealing with a specific object.
+
+Contour currently invalidates the entire resource if a configuration error is encountered which can cause downtime for specific routes. 
+
+## Goals
+- Serve valid traffic configurations if portions of the spec are invalid
```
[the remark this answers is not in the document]
youngnick: I'd also add "Define and distinguish between fatal and non-fatal errors" or something, as to implement this, we're going to have to go deep on the distinction between Errors and Warnings.

design/httpproxy-conflict.md
```diff
@@ -0,0 +1,118 @@
...
+- Change how Contour processes objects today from informers set against the Kubernetes API
+
+## High-Level Design
+Contour will set the error or warning condition for the when an error is encountered, but do its best to still serve valid configurations.
+
+It's difficult to find errors and block them from breaking valid configurations since Contour currently processes objects after they are committed to the API server.
+Due to this, Contour cannot block changes to resources before they are committed like an `Admission Controller` would be able to do (See alternatives: `HTTPProxyInstance`)
+If Contour handles a configuration error by programming Envoy to return an error response instead of proxying the request through the route, the error will be published as an `Error` status.
+
+In general, Contour will set an object to be status `Error` if the request response to the `User` is changed from what is configured.
```
youngnick, line 40: I initially liked this way of describing it, but I think it's better to be much more specific about the exact behavior that an error indicates, versus a warning. That is, that an `Error` indicates that the object will not be processed by Contour - no part of it will end up as config. This is the behavior we have now. A Warning indicates that the object has a problem, but it's not significant enough for Contour not to be able to accept the object, so at least parts of the HTTPProxy should end up in Envoy.

This is not one hundred percent correct - I think there are edge cases, but I think it's closer to the main intent of what we're trying to do.

Given that, I'd suggest moving something like this to much earlier in the document, probably we should draw the distinction as part of the Background section. Something like this:

"When processing HTTPProxy objects, it turns out some problems should not stop the rest of the object from being processed. To distinguish between fatal-for-processing and not, we propose calling the former `Errors`, and the latter `Warnings`."

Then we've got the description of what an error and a warning are in this context out of the way early, and we can refer to them throughout this document, confident we're all on the same page.

design/httpproxy-conflict.md
youngnick: Doing a hypothetical DAG run is an interesting idea, but I don't know how well it would work - I don't think that the DAG is uncoupled enough from the rest of Contour at the moment for this to fly.

design/httpproxy-conflict.md
```diff
@@ -0,0 +1,118 @@
+# HTTPProxy Conflict Resolution Design
+
+Status: Draft
```
youngnick, line 3: Just make this `Status: Approved` for now. We won't come back and update this later.

youngnick (changes requested):
This is really close, I think that it's *really* important to ensure that there's no confusion about what we mean by "Error" and "Warning" though.

design/httpproxy-conflict.md
stevesloka: > Error indicates that the object will not be processed by Contour

This isn't my intent and get's to what @skriss was asking for. I'll update the doc with my thoughts, but an "Error indicates that **parts** of the ojbect  will not be processed by Contour, but other parts still may be". The goal of this design is to still allow portions of the object to be configured while others aren't. This means the object could have an "Error" status, but still serve traffic (which does sound weird).

design/httpproxy-conflict.md
```diff
@@ -0,0 +1,122 @@
...
+- Define and distinguish between fatal and non-fatal errors
+
+## Non Goals
+- Change how Contour processes objects today from informers set against the Kubernetes API
+
+## High-Level Design
+Contour will set the error or warning condition when a problem is encountered, but do its best to still serve valid configurations.
+
+In general, Contour will set an object to be status `Error` if the request response to the `User` is changed from what is configured.
+Contour will set a `Warning` when the object has an issue, but the response is not modified.
```
sunjayBhatia, line 45:
```suggestion
Contour will set a `Warning` when the object has an issue, but the response a `User` sees when accessing the application for which the resource is configured is not modified.
```

sunjayBhatia (approved):
left a tiny suggestion on wording but otherwise LGTM

design/httpproxy-conflict.md
stevesloka: added suggested text

design/httpproxy-conflict.md
```diff
@@ -0,0 +1,125 @@
...
+- Serve valid traffic configurations if portions of the spec are invalid
+- Define and distinguish between fatal and non-fatal errors
+
+## Non Goals
+- Change how Contour processes objects today from informers set against the Kubernetes API
+
+## High-Level Design
+Contour will set the error or warning condition when a problem is encountered, but do its best to still serve valid configurations.
+
+In general, Contour will set an object to be status `Error` if the request response to the `User` is changed from what is configured.
```
skriss, line 47: L28 and L47 seem to give different definitions for `Error` - L28 says it means "fatal-for-processing", i.e. no part of the proxy will be programmed with Envoy, whereas L47 says it means the response to the user is changed. Can we clear that up? Do *some* errors mean no part of the proxy is programmed with Envoy, whereas *other* errors can indicate partial programming? Or should the latter always correspond with a `Warning`? I think we need to be super-clear about this for users' sake.

youngnick: I agree that we should be super clear. My original idea about this was closer to the latter - any entry in the `Errors` slice in the `DetailedCondition` implies a fatal-for-processing, unrecoverable error that means that no config from that HTTPProxy will end up in Envoy.

When a HTTPProxy with an `Error` is included in another, the includ**ing** HTTPProxy must have a `Warning` at least, unless that would mean that there are no valid Routes, in which case the includ**ing** HTTPProxy would also have an `Error`.

stevesloka: > unrecoverable error that means that no config from that HTTPProxy will end up in Envoy

This is 100% NOT my goal in this design. One HTTPProxy can have errors (say an invalid service), but still serve routes to other proxies.  

I removed line 28 and moved line 47 it up to where line 28 was.

Any entry in the Error slice tells that the response is changed (e.g. 503, etc), that's it.

youngnick: So how do we tell the difference then between "you made a big mistake in your HTTPProxy" and "Oh, turns out your service doesn't exist"? I *thought* that was the distinction between `Error` and `Warning`?

youngnick: To put this another way: I thought that `Errors` basically mapped to the old, non-condition `Valid` field. I thought that entries in the `Errors` field would be the same as the thing being `Invalid`.

stevesloka: The goal of this design is to make sure users applications serve traffic. 

For example if a root proxy resource has 2 paths: `/foo` & `/bar` and the service for `/foo` is missing, today both paths are invalid and won't serve traffic. This design aims to allow `/bar` to still serve traffic even if the service for `/foo` is missing. 

Since the service is missing, this design changes how Contour handles this scenario, which means that any requests to `/foo` will return an HTTP 503 and requests to `/bar` will still route properly. 

Since the request is changed from it's original configuration (meaning the service is missing and now returns a 503), that's an Error. Any request to the path `/foo` should route to a service, but doesn't, they get a 503 response. 

A different scenario where the configured path is not changed, is a warning, which says, "hey we tried to configure this, something's wrong, but your user will still get to the place configured" (at best). 

See: https://github.com/projectcontour/contour/pull/3145/files#diff-dc07f1241f06c945856ac6f7fc155659c36976138500362c6955ccdf02adbc70R34-R36

youngnick: I understand that, so do you mean that you feel there's no distinction between "hey, most of this config worked" and "you messed up your FQDN, nothing did"?

Because right now, both of those things are Errors. I feel like the latter is a more severe class of Error, that we should treat differently.

I agree that we should be setting things to 503 when the service exists, and that that's why we started this. I get the rule about needing to tell the user that things are different to what they asked for, but I think there's a difference in degree between "one of your endpoints is busted" and "none of this config will work at all". If we use `Error` to represent the former, that leaves us with nothing for the latter.

Also, what about when it's a combination of headers that we can't satisfy? That is, the headers produce an empty set? Is that a warning or an error? Feels like a warning to me, but I'm not sure if I'm right here.

stevesloka:
I'm putting this design back in the drawer for now, can pick it back up later.

design/httpproxy-conflict.md
stevesloka: > I understand that, so do you mean that you feel there's no distinction between "hey, most of this config worked" and "you messed up your FQDN, nothing did"?

What do you suggest then? I feel like calling both of those "Errors" is fine. Are you proposing we add a "SevereErrors" set to the conditions? 

> Also, what about when it's a combination of headers that we can't satisfy? That is, the headers produce an empty set? Is that a warning or an error? Feels like a warning to me, but I'm not sure if I'm right here.

Warning

stevesloka:
Hey, @youngnick I forgot about @tsaarni's work on the other PR. I feel like this is super close and would like to unblock his work which is why I re-opened.

design/httpproxy-conflict.md
stevesloka: Some of this is what @skriss brought up about informing the user that the config was rejected or only served a partial config. Short of explaining in the error messages, I don't think we have a good way to explain that to users.

stevesloka:
@skriss did you have any more comments on this?

skriss:
> @skriss did you have any more comments on this?

I'll take another look today.

design/httpproxy-conflict.md
skriss: Resolving this thread b/c we've continued discussion on this topic in a separate thread.

design/httpproxy-conflict.md
skriss: We talked about this at length in the community office hours today. Here's where I think we ended up:

- both "fatal errors" (e.g. an invalid FQDN) and "non-fatal errors" (e.g. a route references a non-existent service) should be reported as `Errors` under the `Valid` Condition.
- in the case of "non-fatal errors", Contour should continue to process the `HTTPProxy` and attempt to program as much of it as possible with Envoy. This is particularly beneficial for when an `HTTPProxy` already exists and is valid, and has a bad edit made to it (e.g. to a single route). We don't want to stop serving traffic to other unaffected routes if we can avoid it.
- the result of the above two points is that an `HTTPProxy` with a `Valid: false` condition _may_ be partially programmed in Envoy, or may not be programmed at all in Envoy
- **it's not essential to differentiate between those cases**; either way, the user will be told something is wrong with the HTTPProxy via the `Valid: false` condition, and they will want to remediate it.

@youngnick, originally I was on the other side of the last bullet, i.e. I thought we needed to differentiate between partial programming and no programming, but I no longer think it's essential since either way, the user is told about the problem and will want to fix it. I *do* think that in the long run, we probably want to look at not just having a single, binary `Valid` condition, but instead having more top-level conditions that, in combination, could describe these shades of gray better. However, (a) I don't think we need to do that now, and (b) that's probably considered a breaking change that requires a v2 anyway.

Thoughts?

design/httpproxy-conflict.md
```diff
@@ -0,0 +1,123 @@
...
+
+Contour currently invalidates the entire resource if a configuration error is encountered which can cause downtime for specific routes or even the entire virtual host.
+
+When processing HTTPProxy objects, some problems should not stop the rest of the object from being processed.
+Contour will set an object to be status `Error` if the request response to the `User` is changed from what is configured.
+Contour will set a `Warning` when the object has an issue, but the response a `User` sees when accessing the application for which the resource is configured is not modified.
+
+## Goals
+- Serve valid traffic configurations if portions of the spec are invalid
+- Define and distinguish between fatal and non-fatal errors
```
skriss, line 40: Since it's a goal to define these, I think it'd be great to add a short **Definitions** section that very precisely defines each of these, something like (feel free to adjust):

## Definitions
### Fatal Error
An error encountered while processing an HTTPProxy spec where:
- Contour stops processing the rest of the spec
- Contour does not program any part of the HTTPProxy in Envoy

For example, an invalid FQDN.

### Non-Fatal Error
An error encountered while processing an HTTPProxy spec where:
- Contour continues processing the rest of the spec
- Contour *does* program the unaffected part of the HTTPProxy in Envoy
- Users may get a different response for a given route than what was configured (e.g. a 502 instead of the desired upstream service)

For example, a route to a nonexistent service (results in a 502 being programmed)

### Warning
An error encountered while processing an HTTPProxy spec where:
- Contour continues processing the rest of the spec
- Contour *does* program the unaffected part of the HTTPProxy in Envoy
- Users are routed to the correct destinations, but some aspects of the request handling may be different than desired (e.g. missing timeout or retry settings)

For example, invalid retry settings on a route (the route is still programmed, but without retry settings)

skriss (commented):
A couple minor clarifications, but I think I'm now broadly in agreement with this design.

The way I'm thinking about this now is - currently, all HTTPProxy spec errors are treated as fatal in Contour; let's start to relax some of those into non-fatal errors and warnings. 

I think for implementation, we can definitely start with just one or a small handful of these errors that we'd like to relax, rather than trying to do them all at once, since we might learn something that causes us to adjust our approach.

design/httpproxy-conflict.md
```diff
@@ -0,0 +1,123 @@
...
+Contour has further extended this to add `Conditions` which allows for a set of `Errors` and `Warnings` which inform users of errors in the HTTPProxy resource.
+
+It's difficult to find errors and block them from breaking valid configurations since Contour currently processes objects after they are committed to the API server.
+Due to this, Contour cannot block changes to resources before they are committed like an `Admission Controller` would be able to do (See alternatives: `HTTPProxyInstance`).
+When processing HTTPProxy objects, some problems should not stop the rest of the object from being processed.
+To distinguish between fatal-for-processing and not, we propose calling the former `Errors`, and the latter `Warnings`.
+
+When the first error or warning is detected, these fields are updated and in the status of the HTTPProxy object, however if there are more errors, only the first is shown to the user.
+There are ongoing efforts to change this logic to inform of all the errors or warnings dealing with a specific object.
+
```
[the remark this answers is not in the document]
stevesloka: 👍

youngnick (changes requested):
So what we've really ended up with here is Error (Fatal Error), Warning (Non-Fatal Error), and Info (Warning).

I think the distinction is useful, but we have no way to represent this currently in the Conditions. We currently only give you one `Valid` condition, that can be either `status: true` or `status: false`. Currently, it's also defined in its spec that **any entries in the `Errors` slice** mean that it is `status: false` (or Invalid). Conditions behavior is all around the `type` and `status` fields, and if we want to be broadly compatible and ready for our DetailedCondition to be able to be duck-typed back into a Condition, we need to ensure we're behaving in the expected way.

If we're going to go ahead with breaking out Fatal errors, non-fatal errors, and Warnings, then this design **is not complete** without an update to the Conditions schema for HTTPProxy to take that into account. 

Either you'll need to add a field to the `DetailedCondition`, so you have three levels, or you need to redesign how the `status` field of the DetailedCondition interacts with the `errors` and `warnings` slices.

stevesloka:
The idea was to treat errors like errors regardless if it's fatal or
nonfatal, still an error.

youngnick:
So, if that means that the meaning of the `Valid` condition has changed, then we'll need to update that. To me, the fact that you can't tell if an error will mean that you'll end up with Envoy config or not is a bad user experience.

The intended use of Conditions is that you should be able to use the `status` of the condition to find out something quickly. Right now, if the `Valid` Condition has `status: false`, that means that maybe your config has shown up in Envoy or maybe it hasn't. It could be that we really shouldn't be doing the DetailedCondition thing, and we should just add top-level conditions for all the possible errors instead. But, as a user, how many dereferences is it going to take you to figure out what's going on? If you are getting a 503 for a service, how will you know what caused it?

I understand what we're trying to do, but **if we build it before answering these questions**, we will probably have to throw away more work that we are spending by discussing this. That's what design documents are **for**.

youngnick:
To put this another way: the `Valid` condition is currently built to mean the same thing as the `validity` field on the original spec. If we implement this, it won't work like that any more, because things that were fatal will no longer be fatal, but will still be marked as `Errors`. 

So we have to define what it **does** mean, as part of this document, so that we can make sure we're not missing anything.

skriss:
My thinking was that `Valid: true` means "100% valid", and `Valid: false` means "not 100% valid". If you're in a "Valid: false" state you should go fix your errors, whether your proxy got partially programmed or not. Your action as a user is the same in either case: remediate your errors to get to `Valid: true`. That's why I decided it wasn't so important to distinguish between partial programming and no programming, so we can record both fatal and non-fatal errors as just errors.

As I said in https://github.com/projectcontour/contour/pull/3145#discussion_r570535939, I do think ultimately we should move away from the single top-level binary `Valid` condition, but I don't think we have to do that to make progress here.

I'd suggest that we try to resolve this on a call since there's still some contention here, I think it'll be a lot more efficient & effective than continuing to go back and forth on GitHub.

youngnick:
Yeah, I agree, I think that a call is the way to go here.

youngnick:
I was reading over some other issues today, and I realised that I hadn't been explicit about why I have been being picky about what behavior we are defining here. I've got lost in the weeds, and not explained myself well. I apologise for that.

This PR is a design primarily to introduce the possibility of having invalid routes still serve an error code. In order to be able to do this, we need a way to tell the user that the config they are getting is not the config they asked for.

A closely related problem is that the current behavior means that, if a HTTPProxy that is included in another HTTPProxy, and the includ**ed** HTTPProxy has an error, it is rendered invalid and dropped from processing. An additional purpose of this design is to allow us to not drop the whole config, preventing a small error from propagating up an include graph. If we can get this right, then we can look at adding some way to select a set of HTTPProxies rather than a single one, as requested in #2206. This may also help with soemthing like #2189, although the gateway-apis implementation could also be used for both of those use-cases.

So, with that background out of the way, I think that I've been overly cautious here. The proposal as it stands when I write this allows a user to tell that what they configured could not be satisfied somehow, and points them to where to look. What I think is missing is a way to know what bits have made it to Envoy and what haven't. If you have a severe enough error, then the whole HTTPProxy may not produce any Envoy config, but there's no way to know without knowing which errors mean that and which don't.

After leaving it to background process in my brain though, I think we'd probably be better served to have a separate top-level condition or something that conveys that information, if it's required. It's much quicker to explain that the Condition `Valid` being `false` means that there's something wrong, and you need to look into it, than to make things more complicated with extra layers. If users tell us that they would like a quicker way to know exactly what's wrong, we can figure out what to do then.

youngnick (approved):

stevesloka:
Thanks for the explanation @youngnick!

Issue #1598, opened 2019-09-30
User Experience with HTTPProxy


stevesloka:
## Background:

A key feature that Contour provided with `IngressRoute` and now `HTTPProxy` is a way to safely enable ingress with teams spread across namespaces. 

A likely scenario is that `root HTTPProxies` are managed by Administrators in a namespace that users do not have access. Users also will have `Includes` defined to their HTTPProxies which delegate path prefixes and headers across namespaces (i.e Conditions). 

Additionally, users working in their team namespace need to self-manage their proxies so that they can route traffic to applications (i.e. services). 

## User Story

As a user working in a namespace within a single Kubernetes cluster, I want to be able to self-manage my HTTPProxy resources and understand what Includes have been added to my proxies from other parent/root HTTPProxy resources.

## Problem

Users have no visibility as to what conditions are delegated to them from parent/root HTTPProxy resources. This makes understanding what Conditions are delegated to them very difficult.  Users also need to understand exactly what names are used in the parent HTTPProxy to properly link them together and enable the delegation.

stevesloka:
https://github.com/projectcontour/contour/issues/2495#issuecomment-728562486

xaleeks:
Kind of an old issue so not sure if this is still on your mind @stevesloka.  How do you envision the child leveraging this info about the parent  I’m just thinking about it as this subdomain with the relevant resources have been assigned to him, so does this facilitate privilege escalation in anyway?  

I kind of imagined it the other way around actually, when a child resource is modified in some meaningful way, Contour would append some kind of warning to the status of the parent/root object

stevesloka:
https://github.com/projectcontour/contour/pull/3146

youngnick:
I don't think this is a privilege escalation risk @xaleeks. This is more about telling a HTTPProxy owner where their route is included into, and with what conditions, so that you have a way to know what different requests will look like. This is important because you can have a HTTPProxy included into multiple parents, and there's currently no way to know that's happening.

In particular, this is important because of something like this:
- parent proxy for `www.foo.com` includes a proxy from `marketing` namespace, proxy name is `blog`. It includes that resource with a prefix Condition of `/blog`.
- So all requests coming in to that service will come in with their path set to `http://www.foo.com/blog`. 
- If the service backing the `blog` HTTPProxy is not expecting requests at that path (say, it's expecting them at `/`), then you'll get 404s.
[...]

PR #3146, opened 2020-11-20
design: HTTPProxy Status Metadata design

stevesloka:
Updates: #1598

### DIFF
--- a/design/httpproxy-status-meta.md
+++ b/design/httpproxy-status-meta.md
@@ -11,6 +11,13 @@ In addition, with this information, additional tooling can be developed against 
 ## Goals
 - Expose extra metadata information to HTTPProxy resources
 
+## Use-Cases
+- Allow owners of child HTTPProxy resource to understand what has been included to them without having access to a parent HTTPProxy
+
+## Future Use-Cases
+- If authentication is enabled
+- VHost has TLS
+
 ## Detailed Design
 Extend the HTTPProxy.Status struct to include additional fields to support exposing metadata information:
 
@@ -32,6 +39,7 @@ type HTTPProxyMetadata struct {
 // HTTPProxyParent describes an HTTPProxy who included
 // any MatchConditions or Fqdn to a child HTTPProxy.
 type HTTPProxyParent struct {
+  ObjectReference `json:
   // Name of HTTPProxy object
   Name string  `json:"name,omitempty"`
 
--- a/design/httpproxy-status-meta.md
+++ b/design/httpproxy-status-meta.md
@@ -1,0 +1,52 @@
+# HTTPProxy Status Metadata
+
+Status: Draft
+
+## Abstract
+Since the addition of `Includes` for HTTPProxy, it can be difficult for owners of child HTTPProxy resources to understand what has been included to them without having access to a parent HTTPProxy.
+ 
+Users should have a way to understand this extra metadata about the ingress resources. 
+In addition, with this information, additional tooling can be developed against HTTPProxy resources to drive dashboards (i.e. Octant) and other `kubectl plugins`.
+
+## Goals
+- Expose extra metadata information to HTTPProxy resources
+
+## Detailed Design
+Extend the HTTPProxy.Status struct to include additional fields to support exposing metadata information:
+
+```
+type HTTPProxyStatus struct {
+  ...
+  <existing fields>
+  ...
+  
+  Metadata HTTPProxyMetadata `json:"metadata,omitempty"`
+}
+
+type HTTPProxyMetadata struct {
+  // ParentsRef shows one level parent of any HTTPProxy resources 
+  // which have included this resource.
+  Includes []HTTPProxyParent `json:"includes,omitempty"`
+}
+
+// HTTPProxyParent describes an HTTPProxy who included
+// any MatchConditions or Fqdn to a child HTTPProxy.
+type HTTPProxyParent struct {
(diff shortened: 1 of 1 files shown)


design/httpproxy-status-meta.md
```diff
@@ -0,0 +1,52 @@
...
+Status: Draft
+
+## Abstract
+Since the addition of `Includes` for HTTPProxy, it can be difficult for owners of child HTTPProxy resources to understand what has been included to them without having access to a parent HTTPProxy.
+ 
+Users should have a way to understand this extra metadata about the ingress resources. 
+In addition, with this information, additional tooling can be developed against HTTPProxy resources to drive dashboards (i.e. Octant) and other `kubectl plugins`.
+
+## Goals
+- Expose extra metadata information to HTTPProxy resources
```
jpeach, line 12: I think it would be helpful to discuss use cases for different kinds of metadata. For example, exposing the vhost is helpful for debugging; I'd also suggest exposing whether authentication is enabled, and whether the vhost is TLS.

design/httpproxy-status-meta.md
```diff
@@ -0,0 +1,52 @@
...
+  <existing fields>
+  ...
+  
+  Metadata HTTPProxyMetadata `json:"metadata,omitempty"`
+}
+
+type HTTPProxyMetadata struct {
+  // ParentsRef shows one level parent of any HTTPProxy resources 
+  // which have included this resource.
+  Includes []HTTPProxyParent `json:"includes,omitempty"`
```
jpeach, line 29: Why only one level here? Are you expecting that tooling would walk the chain of references to build up the inclusion tree? If the user can't access all the parents, would that limit the ability of the user to see the full context?

stevesloka: I was thinking giving users the full chain wouldn't be good since you're exposing the entire cluster. Plus I think it could get out of hand depending on how many includes there might be. Currently, Contour only reasons about having a single include to a child, so my thought was to mirror that and have the child 1. know where they get their includes from 2. Understand the full path of includes that have been passed.

design/httpproxy-status-meta.md
stevesloka: I have the vhost in the design currently, but could add a section for use-cases. 

Yeah auth is probably a good addon to understand which routes have it enabled and which are disabled, etc.

stevesloka:
I'm going to think about this design a bit. I think the go structs need a bit of love to better be able to describe the status information but want to still get feedback from folks. =)

design/httpproxy-status-meta.md
youngnick: Yes, I'd like some more detail about the use cases you're trying to hit. Why those ones first? What other ones can you think of? Why not those this time? How are you going to allow for growth in the future?

youngnick (changes requested):
Edit: I'm sorry, I didn't catch that this is still WIP. I think what I said below still applies though.

What you need to be doing with this design is not just laying out the structs you want to add, but explaining *why* this is a good idea (which I think it is), and thinking about how this might be extended in the future.

This document should be consumable to a person with some understanding of Contour, but none about status, so that they walk away understanding what this work is about.

Right now I can see *what* you want to do, but not *why*, or how this will impact Contour in the future.

design/httpproxy-status-meta.md
youngnick: Your comment there Steve is a great example of explanation that should be in the high-level design section, outlining what the goals for your design are and how you expect to achieve them.

youngnick: I think that the structs are fine in general, they meet the goal you've put there. I worry that the UX of actually tracing things might take a lot of steps, but I agree that exposing the full include chain is probably too much information.

design/httpproxy-status-meta.md
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// HTTPProxyParent describes an HTTPProxy who included
+// any MatchConditions or Fqdn to a child HTTPProxy.
+type HTTPProxyParent struct {
```
youngnick, line 34: Perhaps in here instead of just `name` and `namespace` we should use one of the `ObjectReference` fields? Maybe that would make consuming this information with tooling easier.

stevesloka: So just tuck this into another struct like `GatewayReference`? (https://kubernetes-sigs.github.io/service-apis/spec/)

design/httpproxy-status-meta.md
```diff
@@ -0,0 +1,60 @@
...
+  // Name of HTTPProxy object
+  Name string  `json:"name,omitempty"`
+
+  // Namespace of HTTPProxy Object
+  Namespace string `json:"namespace,omitempty"`
+
+  // Fqdn of HTTPProxy Object
+  Fqdn string `json:"fqdn,omitempty"`
+
+  // IncludedConditions show the set of MatchConditions available
```
youngnick, line 52: What happens when a HTTPProxy is included by more than one parent?

youngnick (changes requested):
The struct design is great, but this needs more information about what we're trying to achieve with the design, and how we can extend it in the future.

Additionally, it needs to handle cases of a HTTPProxy being included by multiple parents (which is definitely allowed in the spec).

skriss:
Moving to 1.14 milestone, this isn't actively being worked on at the moment.

Issue #2988, opened 2020-10-05
Add documented constants for condition status types and reasons


jpeach:
**Please describe the problem you have**

All the condition status types and reasons should have well-known constants and corresponding godoc.

PR #3148, opened 2020-11-20
apis: Add ErrorType constants to DetailedConditions

stevesloka:
Adds ErrorType enum to be used for the conditionType passed
when setting Error status on HTTPProxy resources.

Updates: #2988

### DIFF
--- a/apis/projectcontour/v1/detailedconditions.go
+++ b/apis/projectcontour/v1/detailedconditions.go
@@ -113,4 +113,57 @@ type DetailedCondition struct {
 	Warnings []SubCondition `json:"warnings,omitempty"`
 }
 
-const ValidConditionType string = "Valid"
+const (
+	// ConditionTypeAuthError describes an error condition related to Auth.
+	ConditionTypeAuthError = "AuthError"
+
+	// ConditionTypeCORSError describes an error condition related to CORS.
+	ConditionTypeCORSError = "CORSError"
+
+	// ConditionTypeIncludeError describes an error condition with
+	// inclusion of another HTTPProxy resource.
+	ConditionTypeIncludeError = "IncludeError"
+
+	// ConditionTypeOrphanedError describes an error condition
+	// with an HTTPProxy resource which is not part of a delegation chain.
+	ConditionTypeOrphanedError = "Orphaned"
+
+	// ConditionTypePrefixReplaceError describes an error condition with
+	// an HTTPProxy path prefix replacement issue.
+	ConditionTypePrefixReplaceError = "PrefixReplaceError"
+
+	// ConditionTypeRootNamespaceError describes an error condition
+	// with an HTTPProxy resource created in non-root namespace.
+	ConditionTypeRootNamespaceError = "RootNamespaceError"
+
+	// ConditionTypeRouteError describes an error condition that
+	// relates to Routes within an HTTPProxy.
+	ConditionTypeRouteError = "RouteError"
+
+	// ConditionTypeServiceError describes an error condition that
+	// relates to a Service error within an HTTPProxy.
+	ConditionTypeServiceError = "ServiceError"
+
+	// ConditionTypeSpecError describes an error condition that
+	// relates to the Spec of an HTTPProxy resource.
+	ConditionTypeSpecError = "SpecError"
+
+	// ConditionTypeTCPProxyIncludeError describes an error condition
+	// with inclusion of another HTTPProxy TCP Proxy resource.
+	ConditionTypeTCPProxyIncludeError = "TCPProxyIncludeError"
+
+	// ConditionTypeTCPProxyError describes an error condition relating
+	// to a TCP Proxy HTTPProxy resource.
+	ConditionTypeTCPProxyError = "TCPProxyError"
+
+	// ConditionTypeTLSError describes an error condition relating
+	// to TLS configuration.
+	ConditionTypeTLSError = "TLSError"
+
+	// ConditionTypeValid describes an valid condition.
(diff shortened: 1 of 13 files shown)


apis/projectcontour/v1/detailedconditions.go
```diff
@@ -114,3 +114,48 @@ type DetailedCondition struct {
...
+	ServiceError
+	TCPProxyIncludeError
+	VirtualHostError
+	TLSError
+	RootNamespaceError
+	SpecError
+	OrphanedError
+)
+
+var errorConditions = []string{
```
stevesloka, line 145: A follow-up would be to add docs to this, would be neat to autogenerate those types.

youngnick (changes requested):
Could you please separate out the import rename? I agree, let's make the import names consistent here with everything else.

In general, I don't like using an enum in a public API like this, because adding a value to the enum can be a breaking change. I think it's better to use a string alias type, with constants in that type for the specified values. This is the same pattern used for the core Kubernetes objects.

stevesloka:
I liked the enum because it forces you to add to the list and document, with a string it doesn't. You can use any type you want if it gets past code review.

youngnick:
I think it's a mistake to try and enforce all good engineering practices in code. We can just agree that we need to document all the types and reasons we use, and do our best to make sure that happens. I don't think it's practical to do that programmatically, so there's always going to be an element of us doing the right thing.

stevesloka:
> I think it's a mistake to try and enforce all good engineering practices in code.

??? 

>  We can just agree that we need to document all the types and reasons we use and do our best to make sure that happens.

I'll update the PR next week when back from PTO.

youngnick:
The point I was (poorly) trying to make is that code only gets you so far, at some point people have to agree to do their best to have good engineering practices.

apis/projectcontour/v1/detailedconditions.go
```diff
@@ -113,4 +113,20 @@ type DetailedCondition struct {
 	Warnings []SubCondition `json:"warnings,omitempty"`
 }
 
-const ValidConditionType string = "Valid"
+const (
+	ValidConditionType string = "Valid"
+
+	IncludeError         = "IncludeError"
```
stevesloka, line 119: I can't make these a type since that breaks the API being the helpers aren't in `internal` packages so this was left as strings. 

Should we name these with `ConditionType` appended? For example, s/IncludeError/IncludeErrorConditionType/ ?

skriss: I might do `ConditionTypeIncludeError` - helps group them, particularly for code completion.

skriss: Alternately we could add a method per condition type - e.g. `.AddIncludeError(...)`, `.AddAuthError(...)`, etc. Perhaps makes it less likely that a developer will inadvertently pass an invalid condition type?

apis/projectcontour/v1/detailedconditions.go
```diff
@@ -113,4 +113,20 @@ type DetailedCondition struct {
 	Warnings []SubCondition `json:"warnings,omitempty"`
 }
 
-const ValidConditionType string = "Valid"
+const (
+	ValidConditionType string = "Valid"
```
skriss, line 117: may as well make this declaration consistent with the other ones, i.e. omit `string`

apis/projectcontour/v1/detailedconditions.go
stevesloka: I think maybe let's just do the strings for now, could revisit the helpers. The thing about the helpers is they get out of hand quickly, like how the `HTTPConnectionManagerBuilder` is today inside `envoy/listener.go`

apis/projectcontour/v1/detailedconditions.go
skriss, line 117: since these are part of the API, let's add godoc per constant.

stevesloka: added

stevesloka: removed

apis/projectcontour/v1/detailedconditions.go
```diff
@@ -113,4 +113,57 @@ type DetailedCondition struct {
 	Warnings []SubCondition `json:"warnings,omitempty"`
 }
 
-const ValidConditionType string = "Valid"
+const (
+	// ConditionTypeAuthError describes an error condition error related to Auth.
```
skriss, line 117:
```suggestion
	// ConditionTypeAuthError describes an error condition related to Auth.
```

skriss: looks like there are a few instances of this typo

skriss (approved):
one nit, otherwise LGTM

apis/projectcontour/v1/detailedconditions.go
```diff
@@ -113,4 +113,57 @@ type DetailedCondition struct {
...
+	// ConditionTypeTCPProxyError describes an error condition relating
+	// to a TCP Proxy HTTPProxy resource.
+	ConditionTypeTCPProxyError = "TCPProxyError"
+
+	// ConditionTypeTLSError describes an error condition relating
+	// to TLS configuration.
+	ConditionTypeTLSError = "TLSError"
+
+	// ConditionTypeValid describes an valid condition.
+	ConditionTypeValid = "Valid"
```
skriss, line 164: sigh. I guess technically changing the name of this is a breaking change too. We can either leave this one as it was, or support both forms of the constant for it, and mark the old one as deprecated.

PR #3152, opened 2020-11-24
Fix integration test failure

youngnick:
The auth integration test was finding a problem; that in v3
the HTTP router filter must be the last in the list. This was not the
case when external auth was enabled.

This commit updates the `AddFilter` logic to always move the Router
filter to the end of the filter chain, whenever any filter is added,
and updates the tests accordingly.

It is also technically possible, though a programmer error, to add
a second Router filter to the filter chain. Added a panic in this case,
as there is no way it's possible using the `AddFilter` method. This
catches a programmer error and fails fast and hard to tell that person
not to do that, and a test to lock in the behavior.

### DIFF
--- a/internal/envoy/v3/listener.go
+++ b/internal/envoy/v3/listener.go
@@ -14,6 +14,7 @@
 package v3
 
 import (
+	"errors"
 	"fmt"
 	"log"
 	"sort"
@@ -26,11 +27,13 @@ import (
 	envoy_compressor_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/compressor/v3"
 	envoy_config_filter_http_ext_authz_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/ext_authz/v3"
 	lua "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/lua/v3"
+	envoy_extensions_filters_http_router_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/router/v3"
 	http "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/http_connection_manager/v3"
 	tcp "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/tcp_proxy/v3"
 	envoy_tls_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/transport_sockets/tls/v3"
 	envoy_type "github.com/envoyproxy/go-control-plane/envoy/type/v3"
 	"github.com/envoyproxy/go-control-plane/pkg/wellknown"
+	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/any"
 	"github.com/projectcontour/contour/internal/dag"
 	"github.com/projectcontour/contour/internal/envoy"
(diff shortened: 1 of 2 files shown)


internal/envoy/v3/listener.go
```diff
@@ -257,23 +257,43 @@ func (b *httpConnectionManagerBuilder) DefaultFilters() *httpConnectionManagerBu
...
+
+	if len(b.filters) == 1 {
+		return b
+	}
+
+	lastIndex := len(b.filters) - 1
+	routerIndex := -1
+	for i, filter := range b.filters {
+		if filter.GetTypedConfig().TypeUrl == HTTPFilterRouter {
+			routerIndex = i
```
jpeach, line 280: nit: `break`

internal/envoy/v3/listener.go
```diff
@@ -257,23 +257,43 @@ func (b *httpConnectionManagerBuilder) DefaultFilters() *httpConnectionManagerBu
...
 		}
 	}
 
-	b.filters = append(b.filters, f)
+	// We can't add more than one router entry, and there should be no way to do it.
+	// If this happens, it has to be programmer error, so we panic to tell them
+	// it needs to be fixed.
+	if routerIndex != -1 && f.GetTypedConfig().TypeUrl == HTTPFilterRouter {
```
jpeach, line 287: We can leave this as it is, but maybe a cleaner way to test the type is to use  [`ptypes.Is()`](https://godoc.org/github.com/golang/protobuf/ptypes#Is), .e.g.

```Go
if ptypes.Is(f.GetTypedConfig(), &Router{}) {
        // Do the thing ....
}
```

jpeach: Or we also have `protobuf.AnyMessageTypeOf()` to generate a type URL from a message struct.

internal/envoy/v3/listener.go
```diff
@@ -257,23 +257,43 @@ func (b *httpConnectionManagerBuilder) DefaultFilters() *httpConnectionManagerBu
...
+		panic("Can't add more than one router to a filter chain")
+	}
+
+	if routerIndex != lastIndex {
+		// Move the router to the end of the filters array
```
jpeach, line 292: Nit: end sentences with a full stop :)

internal/envoy/v3/listener.go
```diff
@@ -296,6 +316,17 @@ func (b *httpConnectionManagerBuilder) Validate() error {
...
+	// More specifically, the last filter must be a terminating filter. The only one
+	// of these used by Contour is the router filter, which is set as the one
+	// with typeUrl `type.googleapis.com/envoy.extensions.filters.http.router.v3.Router`.
+	if len(b.filters) > 0 {
+		lastIndex := len(b.filters) - 1
+		if b.filters[lastIndex].GetTypedConfig().TypeUrl != HTTPFilterRouter {
+			return fmt.Errorf("The last filter is not of type %s", HTTPFilterRouter)
+		}
+	}
+
```
jpeach, line 329: Not that we want to guarantee the router is always present and always last, can we replace this whole function with:

```Go
// It's not OK for the filters to be empty.
if len(b.filters) == 0 {
            return errors.New("filter list is empty")
}

lastIndex := len(b.filters) - 1
if protobuf.AnyMessageTypeOf(b.filters[lastIndex].GetTypedConfig()) != HTTPFilterRouter {
        // ptypes.Is is probably faster ...
        return fmt.Errorf("last filter is not of type %q", HTTPFilterRouter)
}
```

Error strings [should not be capitalized](https://github.com/golang/go/wiki/CodeReviewComments#error-strings)

internal/envoy/v3/listener_test.go
```diff
@@ -1005,15 +1004,186 @@ func TestTCPProxy(t *testing.T) {
...
+			TypedConfig: &any.Any{
+				TypeUrl: "foo",
+			},
+		},
+	})
+	assert.Errorf(t, badBuilder.Validate(), "Adding a filter after the Router filter should fail")
+}
+
+func TestAddFilter(t *testing.T) {
+
```
jpeach, line 1037: Nit: omit this newline

internal/envoy/v3/listener_test.go
```diff
@@ -1005,15 +1004,186 @@ func TestTCPProxy(t *testing.T) {
...
+		"Add a single router filter to empty builder": {
+			builder: HTTPConnectionManagerBuilder(),
+			add: &http.HttpFilter{
+				Name: "router",
+				ConfigType: &http.HttpFilter_TypedConfig{
+					TypedConfig: &any.Any{
+						TypeUrl: HTTPFilterRouter,
+					},
+				},
+			},
```
jpeach, line 1057: Seems like we could make a helper:
```Go
func TypedConfigOf(msg proto.Message) *TypedConfig {
        return http.HttpFilter_TypedConfig {
                TypedConfig: protobuf.MustMarshalAny(msg),
        }
}
```

then 
```Go
&http.HttpFilter{Name: "grpcweb", ConfigType: TypedConfigOf(&GrpcWeb{}))
```

internal/envoy/v3/listener.go
stevesloka: Maybe this should be broken out into `AddFilter()` and `SortFilter()` since the logic is different if it's an HTTP set or TCP set.

internal/envoy/v3/listener_test.go
youngnick: This meant we had to swap to using a protobuf comparer for the tests, which I feel is a bigger change than this warrants. I'll leave it for now.

internal/envoy/v3/listener.go
youngnick: Swapped to `ptypes.Is()`.

jpeach (approved):

PR #3159, opened 2020-11-30
apis/projectcontour: replace kubeconditions.go with apimachinery ref

skriss:
Signed-off-by: Steve Kriss <krisss@vmware.com>

PR #3161, opened 2020-11-30
Revert "apis/projectcontour: replace kubeconditions.go with apimachinery ref"

skriss:
Reverts projectcontour/contour#3159

Reverting this for now since it removed some publicly exported types from the API package which was a breaking API change for consumers of the Go package.


skriss:
The lint check is going to fail since this file has a non-conformant copyright header. Can add DCO when squashing for merge (the commit was created via GitHub GUI so doesn't have it).

stevesloka (approved):
/lgtm

PR #3167, opened 2020-12-01
golangci-lint: ignore kubeconditions.go file

skriss:
Ignores apis/projectcontour/v1/kubeconditions.go when linting because
this file is copy-pasted from k8s.io/apimachinery and does not conform
to Contour's linting rules -- in particular, it has a different
copyright header.

This will clear up the persistent lint failure on `main`.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,5 +1,7 @@
 run:
   timeout: 10m
+  skip-files:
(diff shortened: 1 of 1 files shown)


stevesloka (approved):
/lgtm

PR #3173, opened 2020-12-03
internal/xds/v3: Add common logging for envoy and contour xDS servers

### LINKED ISSUE #3037: The "envoy" xDS server doesn't log xDS errors

I hacked up contour so it would configure envoy with a bogus access log token.

Envoy should reject the xDS update, and contour should log that.

Envoy log:
```
[2020-10-18 23:48:27.955][1][warning][config] [source/common/config/grpc_subscription_impl.cc:107] gRPC config for type.googleapis.com/envoy.api.v2.Listener rejected: Error adding/updating listener(s) ingress_http: Not supported field in StreamInfo: JPEACH
```

Contour log:
```
INFO[0000] informer caches synced                        context=xds
INFO[0000] started xDS server                            address="0.0.0.0:8001" context=xds insecure=true
INFO[0000] performing delayed update                     context=contourEventHandler last_update=129.259019ms outstanding=6
```

Contour should at least be logging the error on this code path.

sunjayBhatia:
- logs in the same pattern on discovery requests
- ensure rejected update errors are logged in "envoy" xDS server

Fixes #3037


internal/xds/v3/contour.go
```diff
@@ -121,9 +111,6 @@ func (s *contourServer) stream(st grpcStream) error {
 			return done(log, fmt.Errorf("no resource registered for typeURL %q", req.GetTypeUrl()))
```
sunjayBhatia, line 111: As of this revision, the error log from the `done` func when we reject a request with no type URL will not have the other log fields on it set by the logs in `logDiscoveryRequestDetails`, whether this is an issue or not is an open question

sunjayBhatia:
Maybe still to do/open questions about testing
- Add integration testing ensuring envoy and contour xDS server logging match?
- Add higher level testing to ensure server callbacks are wired up in envoy xDS server mode?

cmd/contour/serve.go
```diff
@@ -553,7 +553,7 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		case config.EnvoyServerType:
 			v3cache := contour_xds_v3.NewSnapshotCache(false, log)
 			snapshotHandler.AddSnapshotter(v3cache)
-			contour_xds_v3.RegisterServer(envoy_server_v3.NewServer(context.Background(), v3cache, nil), grpcServer)
+			contour_xds_v3.RegisterServer(envoy_server_v3.NewServer(context.Background(), v3cache, contour_xds_v3.NewCallbacks(log.Logger)), grpcServer)
```
sunjayBhatia, line 556: Probably need to add higher level tests to make sure this is covered

internal/xds/v3/callbacks.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"fmt"
+
+	envoy_service_discovery_v3 "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
+	envoy_server_v3 "github.com/envoyproxy/go-control-plane/pkg/server/v3"
+	"github.com/sirupsen/logrus"
+)
+
+func NewCallbacks(log *logrus.Logger) envoy_server_v3.Callbacks {
```
stevesloka, line 24: nit: Need a comment describing this function.

internal/xds/v3/callbacks.go
```diff
@@ -0,0 +1,47 @@
...
+		StreamRequestFunc: func(streamID int64, req *envoy_service_discovery_v3.DiscoveryRequest) error {
+			logDiscoveryRequestDetails(log, req)
+			return nil
+		},
+	}
+}
+
+func logDiscoveryRequestDetails(l *logrus.Logger, req *envoy_service_discovery_v3.DiscoveryRequest) {
+	log := l.WithField("version_info", req.VersionInfo).WithField("response_nonce", req.ResponseNonce)
+	if req.Node != nil {
```
stevesloka, line 35: It would be helpful to log the node's name as well which today should correlate to the pod name.

sunjayBhatia: Does `node_id` capture what youre looking for? I think that is set as the pod name at least in a basic/example deployment by the bootstrapper https://github.com/projectcontour/contour/blob/e4ce0d18261c1c036674f9530ae2b0b08d4f4859/examples/contour/03-envoy.yaml#L52

```
time="2020-12-03T21:00:19Z" level=info msg="handling v3 xDS resource request" node_id=envoy-v2g4n node_cluster=projectcontour node_version=v1.16.1 resource_names="[default/kuard]" response_nonce=18 type_url=type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment version_info=18
```

(I added `node_cluster` just now but not sure how useful that is?)

internal/xds/v3/callbacks.go
jpeach: Quick drive-by follow up ... we should name this something more specific too :)

sunjayBhatia: `NewEnvoyXDSCallbacks` (maybe stick `Server` somewhere in there)?

jpeach: This specifically gives you logging, so maybe `RequestLoggger`, `RequestLoggingCallbacks`?

sunjayBhatia:
Having a little trouble finding a good place to put a test that ensures the "wiring" of the Envoy xDS server callbacks and the Contour xDS server request handling without duplication/abstraction+injection for the sake of testing, but it feels like valuable coverage that is missing (e.g. testing the Contour xDS server `serve` method to make sure it logs correctly which would be duplication or adding an injected `callbacks` field to check the logging callback was called)

What is the project's philosophy on this sort of thing? Usually I would just pick something (erring on the side of more test coverage) but since I'm new to the project I thought I would ask since this is a relatively small issue/change since it involves only logging

cmd/contour/serve.go
sunjayBhatia: N/A for now since this is just does logging

internal/xds/v3/callbacks_test.go
```diff
@@ -0,0 +1,126 @@
...
+		},
+	}
+
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			logDiscoveryRequestDetails(log, tc.discoveryReq)
+			var expectedLogEntry *logrus.Entry
+			for _, le := range logHook.AllEntries() {
+				if le.Message == tc.expectedLogMsg {
+					expectedLogEntry = le
```
skriss, line 105: nit, I think you can break the loop at this point if you've found the expected entry

internal/xds/v3/contour.go
skriss: IMHO it'd be better not to lose those fields, for debuggability. 

You could extract a helper that returns a FieldLogger for a request with all the appropriate fields set, then explicitly call it here and also inside `logDiscoveryRequestDetails`. That helper would end up getting called twice for each request for the contour server, but that should be fine since the log fields will just get set again, not duplicated.

skriss (commented):
@sunjayBhatia this looks good! I had a couple small comments.

internal/xds/v3/callbacks.go
skriss: I think `node_id` should cover it, unless @stevesloka had something else in mind?

internal/xds/v3/contour.go
sunjayBhatia: fixed by returning the `*logrus.Entry` so we can use it after populated

skriss: OK, that works

skriss (approved):
LGTM

skriss:
@stevesloka PTAL 🙏

stevesloka (approved):
/lgtm 🎉

PR #3177, opened 2020-12-07
internal/dag: fix debug printout for client cert secret

tsaarni:
This change corrects the debug message when mutual authentication is
misconfigured with a missing secret.  When Ingress was being processed,
the error was printed with empty field secret="<nil>" while the intention
was to print the secret name: secret="namespace/secretname".


skriss (approved):
LGTM, thanks @tsaarni!

stevesloka (approved):
/lgtm

Issue #399, opened 2018-05-17
Tracing Configuration


bobbytables:
It would be nice to expose the bootstrap config a bit more with things like tracing and stats configs. We could even offer something like a file path that gets merged into the bootstrap YAML on boot.

For context, we're building our own control plane for service -> service mesh for Envoy. We're including stats and Zipkin traces (via the jaeger endpoint). We lose the trace context between Contour and our sidecar'd envoys because we can't configure Contour to include trace information.

bobbytables:
Roger that, I'll probably end up doing the mounted configmap then. Thanks!

davecheney:
That’s the approach I’d recommend. If you check out the heptio/gimbal repo, they’re probably already doing that.

mattalberts:
@bobbytables @davecheney 
I loved this suggestion (replacing contour bootstrap with a ConfigMap mounted volume); it's fantastically simple.  I'm also trying to enable tracing (span injection) at the ingress controller; here's the results of my experiment.

TL;DR
  * configuring envoy opentracing manually (without bootstrap) is easy
  * envoy.http_connection_manager still needs a "tracing" definition to emit headers
  * I have a patch! 😄 (if you're interested)

### Contour with OpenTracing
First, deploy contour without bootstrap and mount an updated envoy configuration from a ConfigMap.

```yaml
---
apiVersion: v1
kind: ConfigMap
metadata:
  labels:
    app: contour
    role: ingress
  name: contour-config
  namespace: ingress-system
data:
  contour.yaml: |
    dynamic_resources:
      lds_config:
        api_config_source:
          api_type: GRPC
          cluster_names: [contour]
          grpc_services:
          - envoy_grpc:
              cluster_name: contour
      cds_config:
        api_config_source:
          api_type: GRPC
          cluster_names: [contour]
          grpc_services:
          - envoy_grpc:
              cluster_name: contour
    static_resources:
      clusters:
      - name: contour
        connect_timeout: { seconds: 5 }
        type: STRICT_DNS
        hosts:
          - socket_address:
              address: 127.0.0.1
              port_value: 8001
        lb_policy: ROUND_ROBIN
        http2_protocol_options: {}
        circuit_breakers:
          thresholds:
            - priority: high
              max_connections: 100000
              max_pending_requests: 100000
              max_requests: 60000000
              max_retries: 50
            - priority: default
              max_connections: 100000
              max_pending_requests: 100000
              max_requests: 60000000
              max_retries: 50
[...]

PR #3178, opened 2020-12-07
design: add rate limiting design doc

skriss:
Updates #370.


design/ratelimit-design-new.md
```diff
@@ -0,0 +1,180 @@
...
+```yaml
+rateLimitService:
+  # extensionService contains the namespace/name of the RLS extension service.
+  extensionService: projectcontour/ratelimit
+  # domain is passed to the RLS for all rate limit requests.
+  domain: contour
+  # denyOnFailure defines whether to fail open or closed. If true, then if the RLS
+  # cannot be reached or does not return a valid rate limit decision within the
+  # specified timeout, the client will receive a 429 response for their request.
+  denyOnFailure: true
```
jpeach, line 142: Suggest calling this `FailOpen` to match the AuthorizationServer struct.

design/ratelimit-design-new.md
```diff
@@ -0,0 +1,180 @@
...
+  services:
+    - name: ratelimit
+      port: 8081
+  timeoutPolicy:
+    response: 50ms
+```
+
+Second, a new block will be added to the Contour config file to identify the RLS extension service and provide some additional parameters:
+```yaml
+rateLimitService:
```
jpeach, line 134: Can envoy support a separate service per vhost? Does that config make sense?

design/ratelimit-design-new.md
```diff
@@ -0,0 +1,180 @@
...
+spec:
+  protocol: h2
+  services:
+    - name: ratelimit
+      port: 8081
+  timeoutPolicy:
+    response: 50ms
+```
+
+Second, a new block will be added to the Contour config file to identify the RLS extension service and provide some additional parameters:
```
stevesloka, line 132: 1. This assumes a single cluster will only have one RLS? (Maybe good to call out in the goals). I'm thinking possibly one team may want their own implementation of a global RLS different than another. 
2. It's unfortunate to need another config place. It seems like the ExtensionService should be where you define the domain, denyOnFailure, etc. Having an `RateLimitExtensionService` and a `AuthorizationExtensionService` would solve this, otherwise we've got to have the extra setup in the config file, or tons of validation in the ExtensionService. 
2b. What is the ExtensionService buying us here? I think the original intent was to allow a place to add status conditions, but I'm still unclear as to how that happens.

design/ratelimit-design-new.md
```diff
@@ -0,0 +1,180 @@
...
+```yaml
+rateLimitService:
+  # extensionService contains the namespace/name of the RLS extension service.
+  extensionService: projectcontour/ratelimit
+  # domain is passed to the RLS for all rate limit requests.
+  domain: contour
```
stevesloka, line 138: In my original design I had this hard coded, but could see this being implemented by teams possibly so that things can be defined separately in the external rate limit service. For instance, I think if you setup a remote address descriptor, there isn't an easy way to tell which part of the path is getting that configuration?

design/ratelimit-design-new.md
```diff
@@ -0,0 +1,180 @@
...
+metadata:
+  namespace: projectcontour
+  name: ratelimit
+spec:
+  protocol: h2
+  services:
+    - name: ratelimit
+      port: 8081
+  timeoutPolicy:
+    response: 50ms
```
stevesloka, line 129: There's also a `timeout` property of the rate_limit service (https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ratelimit/v3/rate_limit.proto) , should we expose that as well?

design/ratelimit-design-new.md
```diff
@@ -0,0 +1,180 @@
...
+- The external RLS returns a response indicating whether the client request should be rate-limited or not.
+- If the request should be rate limited, a `429 Too Many Requests` response is returned to the client along with the `x-envoy-ratelimited` header.
+- If the request should not be rate limited, it is routed to the appropriate upstream cluster and proceeds as normal.
+
+## Alternatives Considered
+
+### Tight integration with Lyft RLS
+Lyft has open-sourced a [reference external RLS implementation](https://github.com/envoyproxy/ratelimit/).
+We considered tightly integrating with this specific RLS implementation, to provide a more streamlined UX for configuring rate limits.
+A CRD could be used -- either `HTTPProxy` or a stand-alone one -- to define actual rate limits, and a controller could convert the CRDs into configuration in the Lyft RLS format.
```
stevesloka, line 162: This CRD can still be a generic extension point. Similar to how auth is implemented today in Contour, users would write up rate limits which would turn into CRDs that a controller would need to watch for and convert into specific rate limit configuration depending on the implementation. This would match how we integrate with cert-manager today.

stevesloka: Yeah noted on line 164.

stevesloka (commented):
It would be neat to have some concrete examples of some HTTPProxy resources to get a better idea of what it might look like. For instance, limiting on a key, limiting on a header, limiting on remote IP, etc. 

Also, how are the other fields in the spec configured (https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ratelimit/v3/rate_limit.proto)?

Can we rate limit on gRPC services?

design/ratelimit-design-new.md
skriss: It does not seem to be able to, unfortunately. I'll triple-check, but my attempt at specifying filter config per-vhost did not work.

design/ratelimit-design-new.md
skriss: > 1. This assumes a single cluster will only have one RLS? (Maybe good to call out in the goals). I'm thinking possibly one team may want their own implementation of a global RLS different than another.

The challenge is that (AFAICT) you can only have one RLS per HTTP connection manager, and all non-TLS virtual hosts share the same HTTP connection manager. External auth ran into the same constraint, and opted to support external auth *only* for TLS vhosts, so that you could have a separate auth service per vhost. We *could* make that same choice for rate limiting - allow global rate limiting only for TLS vhosts - but that seems a little strange to me. Definitely worth some discussion.

> 2. It's unfortunate to need another config place. It seems like the ExtensionService should be where you define the domain, denyOnFailure, etc. Having an `RateLimitExtensionService` and a `AuthorizationExtensionService` would solve this, otherwise we've got to have the extra setup in the config file, or tons of validation in the ExtensionService.

Yeah - *if* we could configure an RLS per vhost, then these things would go into a struct as part of HTTPProxy, like AuthorizationServer today. See previous comment, though.

>    2b. What is the ExtensionService buying us here? I think the original intent was to allow a place to add status conditions, but I'm still unclear as to how that happens.

Basically gets the programming of the Envoy cluster for the RLS for free. Other than that, there's not currently any infrastructure for doing health checks, etc.

design/ratelimit-design-new.md
```diff
@@ -0,0 +1,180 @@
...
+```yaml
+rateLimitPolicy:
+  # local defines local rate limiting properties for the virtual host or route.
+  local:
+    # tokenBucket defines the properties of the local rate limiting token bucket.
+    # ref. https://www.envoyproxy.io/docs/envoy/latest/api-v3/type/v3/token_bucket.proto#envoy-v3-api-msg-type-v3-tokenbucket
+    tokenBucket:
+      maxTokens: 100
+      tokensPerFill: 10
+      fillInterval: 1s
```
jpeach, line 75: I'm not that comfortable exposing token bucket parameters directly; that's not how people who want to rate limit think about their problem. Could we re-cast this in terms of requests per second, burst, etc?

design/ratelimit-design-new.md
```diff
@@ -0,0 +1,180 @@
...
+This option was discarded (for now) because we have potential users who are interested in using other rate limiting services, so just providing an integration with the Lyft implementation is not sufficient. 
+Future work could still be done to enable rate limits to be defined via CRD (as part of `HTTPProxy`, or stand-alone) and automatically configured with 1+ underlying RLS implementations.
+
+### Contour as external RLS
+Contour could function as an external RLS itself.
+This would put Contour in the data path for requests.
+It would simplify deployment and configuration for the user, at a cost of significant additional complexity for Contour.
+
+## Compatibility
+Rate limiting (both local and global) will be an optional, opt-in feature for Contour users.
```
jpeach, line 172: If you looked at how other proxies or CRDs express rate limits, maybe make some comparison notes here.

design/ratelimit-design-new.md
stevesloka: Just reading the comments on the struct it wasn't clear to me.

design/ratelimit-design-new.md
stevesloka: > The challenge is that (AFAICT) you can only have one RLS per HTTP connection manager
That makes sense, should just call it out in the doc so that is clear to readers. I don't think we should add a requirement that only TLS'd vhosts will get this service, for auth it makes sense to keep the security boundaries tight, but for rate limiting I don't think it's a big deal other than possibly someone could fake the requests to get around rate limiting? 

> Yeah - if we could configure an RLS per vhost, then these things would go into a struct as part of HTTPProxy, like AuthorizationServer today. See previous comment, though.

You could move some of the configuration items to the vhost, like domain, etc. Also might be a place to have global ratelimiting configuration per vhost.

design/ratelimit-design-new.md
skriss: Yeah, we can do that. It does mean that there's not a clear mapping to the Envoy config anymore, so anyone who wants to dig into how it translates has to do some mental gymnastics, but I guess that's Contour's MO.

skriss: I think the main challenge here is to decide on a fill interval - we could arbitrarily set it to `1s` or some other fixed value. qps and burst don't really allow a user to express what they want there. I suppose, if we let the user specify both the "requests per time period" (e.g. 100) and the "time period duration" (e.g. 1 hour), then we could set fill interval equal to the "time period duration".

skriss: TBQH I find token buckets a lot easier to reason about, but maybe that's just me :)

design/ratelimit-design-new.md
skriss: I don't see any reason not to expose it, but we can default it to `contour`.

design/ratelimit-design-new.md
jpeach: >TBQH I find token buckets a lot easier to reason about, but maybe that's just me :)

That surprises me. I would have thought that people think about rate limiting as rates (ie. events per unit time). 

* The Gloo API is requests per time unit https://docs.solo.io/gloo-edge/latest/guides/security/rate_limiting/crds/
* The Ambassador API is rate per unit https://www.getambassador.io/docs/latest/howtos/advanced-rate-limiting/
* ingress nginx allows request/sec and request/min https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#rate-limiting

design/ratelimit-design-new.md
```diff
@@ -0,0 +1,397 @@
...
+For global rate limiting, the user defines the descriptors to be generated and sent to the external RLS.
+Descriptors contain entries including things like: the client IP, the value of a particular header, the destination cluster, etc.
+The external RLS makes the rate limit decision based on the descriptors, and returns either a 200 or a 429 to Envoy.
+The operator of the external RLS must configure it with actual rate limits for different descriptors.
+Each RLS implementation may have its own configuration format and mechanism for defining rate limits for descriptors, so Contour cannot provide a generic API for defining these.
+
+
+For global rate limiting, the details of the external RLS are configured via an `ExtensionService` and in the Contour config file.
+Note that the main reason this is defined globally for Contour in the config file, rather than per-virtual host, is that there's a rate limit filter per HTTP connection manager, and all non-TLS virtual hosts share a single HTTP connection manager, so they necessarily must share the same rate limit filter config.
+If no global RLS is defined, no global rate limiting will occur.
```
jpeach, line 59: Perhaps what would be useful here is a paragraph positioning global and local rate limiting. From one point of view you could see rate limiting as an abstract API that could be backed by either a global or a local rate limiter. However, I think the implication here is that local and global are solving different problems. Making that explicit (briefly) here would help clarify the API decisions and later (more detailed) help users approach the topic with the right mental model.

design/ratelimit-design-new.md
youngnick: I'm increasingly thinking that we should bring some per-ExtensionService properties into the ExtensionService,  like we already do with timeouts and load balancing policy. However, each type of ExtensionService almost always has extra parameters that belong to the service (in this case, they're currently listed as in the configuration file, and in the tracing case, they would probably need to be part of the vhost configuration currently), and each type of thing has a different set of things that need to be configured.

In this case,  it's these two fields here: `domain` and `failOpen`, and when we do tracing (ie #399 ) there will be other service-specific things, that will be the same for every reference of that service. If we moved to doing this, the config file would just be for the ExtensionServiceRef rather than for config details, which would allow those config details to be flipped without Contour redeploys. Here I'm thinking of flipping between `failOpen` states - that seems like a change that you would want to be able to back out quickly in the event of a problem.

If we could add some sort of (hah) extension point into ExtensionService, that would allow you to supply a set of fields or something, we could move this config into the relevant ExtensionService (where I feel like it kind of belongs, since it's really a property of that service).

It does seem like the second time we use ExtensionService is a good time to think about its general properties. However, I'd be okay with putting this as an open question, to be determined during implementation because it requires some experimentation.

design/ratelimit-design-new.md
```diff
@@ -0,0 +1,397 @@
...
+- First, Envoy applies local rate limiting to incoming requests.
+- If there are no tokens left in the relevant token bucket(s), a `429 Too Many Requests` response is returned.
+- If there was no local rate limiting defined, or there is a token available, Envoy proceeds to global rate limiting.
+- 1+ descriptors, each of which is an ordered list of key/value pairs, are generated for the request based on the global rate limit configuration.
+- The descriptors are sent to the external RLS via gRPC.
+- The external RLS makes a rate limit decision for the request, based on the descriptors it receives.
+- The external RLS returns a response indicating whether the client request should be rate-limited or not.
+- If the request should be rate limited, a `429 Too Many Requests` response is returned to the client along with the `x-envoy-ratelimited` header.
+- If the request should not be rate limited, it is routed to the appropriate upstream cluster and proceeds as normal.
+
```
youngnick, line 166: I'd like to see some thoughts about what and how we communicate to the different personas, both cluster admin and application owner.

In the case of the application owner, it would be nice to be able to tell if I've hit the rate limit, or at least that the rate limit is configured and working correctly for my service.

In the case of the cluster operator, it seems like it's useful to be able to find the services that are consuming the rate limit, ideally with some information about the passed descriptors.

Basically, where is the status surfaced, and what can we make available? I suspect that getting information out of the RLS is not possible currently, but at least specifying what a HTTPProxy user should expect to see if there is a ratelimit seems useful.

In general, I'd like to see a little more info about failure cases. If the RLS fails, how do we know? Does Contour know at all?

It's okay to put those as open questions if we have to, but I feel like a extension design isn't complete without some detail about the UX of consuming it.

We probably will need to see if there's some metrics you can pull from Envoy about this, and include some information about that in the guide page as well.

youngnick (changes requested):
Overall, the design looks great.

I think the big question that's outstanding is about ExtensionService, and what it is as we start using it in more places. I feel like it's a good place for two things:
- Status information about the configured service (that is, that it's Accepted/configured correctly at least, and one day, maybe, if it's Ready).
- Service-specific configuration, like the `failOpen` policy - this allows this to be quickly flipped (at the speed of changing a Kube object rather than the speed of redeploying Contour).

design/ratelimit-design-new.md
jpeach: > In this case, it's these two fields here: domain and failOpen, and when we do tracing (ie #399 ) there will be other service-specific things, that will be the same for every reference of that service.

Without checking how Envoy configures this, I'd suggest that these are properties of the vhost, not properties of the rate limiting service. Just like auth, there is probably some transition where vhosts move from their own rate limiting to the platform one and would twiddle these kinds of knobs.

youngnick: Fair point, I should probably have checked. I agree that for things that are per-vhost, the config should live there, but for things that are properties of the service itself, they should end up on the ExtensionService object.

design/ratelimit-design-new.md
stevesloka: > If we could add some sort of (hah) extension point into ExtensionService, that would allow you to supply a set of fields or something, we could move this config into the relevant ExtensionService (where I feel like it kind of belongs, since it's really a property of that service).

This was my original suggestion when first building out ExensionService, don't make it generic, but make it specific to the need you are configuring. So you'd have an `ExtensionServiceAuth` and `ExtensionServiceRateLimit`, `ExtensionServiceOauth2`, etc, etc. This way you can be clear about fields that need configured and not rely on conditions to say, "actually user, that field isn't meant for this context so don't set it". 

There's a trade-off in complexity here. The overhead of having more CRDs vs a nicer user experience.

design/ratelimit-design-new.md
stevesloka: > In general, I'd like to see a little more info about failure cases. If the RLS fails, how do we know? Does Contour know at all?

I'm still confused about how we push status info into the ExtensionService when the external service is a black box (this applies to auth as well). Is the operator implementing the external service supposed to write status back to the ExtensionService? 

I see two stats for RateLimiting that might be useful to understand if the rate limit was hit, otherwise you might be able to scrape the metrics for specific HTTP status codes to understand? 
```
upstream_rq_retry_backoff_exponential | Counter | Total retries using the exponential backoff strategy
upstream_rq_retry_backoff_ratelimited | Counter | Total retries using the ratelimited backoff strategy
```

skriss: > > In this case, it's these two fields here: domain and failOpen, and when we do tracing (ie #399 ) there will be other service-specific things, that will be the same for every reference of that service.
> 
> Without checking how Envoy configures this, I'd suggest that these are properties of the vhost, not properties of the rate limiting service. Just like auth, there is probably some transition where vhosts move from their own rate limiting to the platform one and would twiddle these kinds of knobs.

The challenge here is that, if we want to support rate limiting for non-TLS vhosts, then the filter can't be per-vhost, since all non-TLS vhosts share the same HTTP connection mananager. That's why it's not at the vhost level.

If we decide not to support non-TLS vhosts, or can come up with some alternate way of getting around the single HCM/filter constraint, then these params could move under the vhost.

I'm open to ideas here!

skriss:
> I'm increasingly thinking that we should bring some per-ExtensionService properties into the ExtensionService, like we already do with timeouts and load balancing policy. However, each type of ExtensionService almost always has extra parameters that belong to the service (in this case, they're currently listed as in the configuration file, and in the tracing case, they would probably need to be part of the vhost configuration currently), and each type of thing has a different set of things that need to be configured.
> 
> In this case, it's these two fields here: domain and failOpen, and when we do tracing (ie #399 ) there will be other service-specific things, that will be the same for every reference of that service. If we moved to doing this, the config file would just be for the ExtensionServiceRef rather than for config details, which would allow those config details to be flipped without Contour redeploys. Here I'm thinking of flipping between failOpen states - that seems like a change that you would want to be able to back out quickly in the event of a problem.
> 
> If we could add some sort of (hah) extension point into ExtensionService, that would allow you to supply a set of fields or something, we could move this config into the relevant ExtensionService (where I feel like it kind of belongs, since it's really a property of that service).
> 
> It does seem like the second time we use ExtensionService is a good time to think about its general properties. However, I'd be okay with putting this as an open question, to be determined during implementation because it requires some experimentation.

Yep, we could add an opaque `parameters` field, or have some kind of specialization of the CRD (`AuthExtensionService`, `RateLimitExtensionService`, etc.). The former is probably much simpler in practice. I agree it'd be nice for these things not to have to go in a config file, and if they can't go into vhost config for reasons discussed in other threads, then into the ExtensionService itself would make sense.

design/ratelimit-design-new.md
skriss: As far as status, metrics are emitted by Envoy with rate limit info:

(Global) https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/rate_limit_filter#statistics
(Local) https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/local_rate_limit_filter#statistics

For global, looking at those should give a picture of if the service is configured correctly, as well as if requests are being limited. Do we want Contour to be scraping these and using them to populate info in the API somewhere?

skriss: Added some details on the stats that Envoy exposes. Given that we don't scrape Envoy metrics to pull into Contour anywhere else, I'm not going to propose doing that now; the user can pull those into a Grafana dashobard from Envoy directly to observe the runtime behavior. We can pursue this separately if we want, though we'd probably want to start with more fundamental metrics around traffic serving, rather than doing it just for rate limiting, IMHO.

design/ratelimit-design-new.md
```diff
@@ -0,0 +1,475 @@
...
+  namespace: projectcontour
+spec:
+  virtualhost:
+    fqdn: local.projectcontour.io
+    tls:
+      secretName: local-tls
+    # rateLimitServer optionally defines a non-default RLS to use
+    # for this HTTPProxy. This field can only be specified for root
+    # TLS-enabled HTTPProxies.
+    rateLimitServer:
```
sunjayBhatia, line 165: Sorry if this is digging backward to the ext-authz extension but is this decoupled from the `rateLimitPolicy` for any specific reason? I noticed it is decoupled there too. What advantages are there in having the extension ref etc. a separate top level field in the vhost config from the policy? (why not have them all under a common `ratelimit` struct?)

sunjayBhatia: Is it so it's easily overridable by the global one in the contour bootstrap config?

skriss: `rateLimitPolicy` can be defined at either the vhost level or at the route level, whereas `rateLimitServer` can only apply for a vhost. That said, we *could* have a parent `rateLimit` struct for vhost that contains both the `rateLimitServer` and `rateLimitPolicy` just for organization, if we wanted. Though, it would add even more nesting to the YAML.

skriss:
Updated this based on yesterday's discussion, around not currently supporting an RLS per TLS vhost.

Let me know if there's any further feedback on the overall design. I'd like to get local rate limiting in for 1.12, and global for 1.13.

design/ratelimit-design.md
```diff
@@ -1,175 +1,501 @@
-# Envoy Rate Limiting Support
+# Rate Limiting Support
 
-The rate limit service configuration specifies the global rate limit service Envoy should talk to when it needs to make global rate limit decisions.
-If no rate limit service is configured, a “null” service will be used which will always return OK if called.
+Status: Draft
```
youngnick, line 3: Until we make a further policy change about design documents, let's just do this as Status: Accepted.

youngnick (approved):
This is a seriously excellent design document, can't wait to see the implementation. Nice work @skriss!

(one extremely small thing that is non-blocking).

skriss:
Looks like the HTTP local rate limit filter just merged descriptor support into Envoy `main`:
- https://github.com/envoyproxy/envoy/pull/14588
- https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/local_rate_limit_filter#config-http-filters-local-rate-limit-descriptors

Need to consider if/how this will fit into the Contour design.

design/ratelimit-design.md
skriss: https://github.com/projectcontour/contour/pull/3178/commits/d9beb4da80a3814a026d6e6f578d28f2aac3a2ef

design/ratelimit-design.md
```diff
@@ -1,175 +1,501 @@
...
-      unit: <see below: required>
-      requests_per_unit: <see below: required>
-    descriptors: (optional block)
-      - ... (nested repetition of above)
-```
 
-#### Rate limit definition
+The `HTTPProxy` processor would ensure that the `rateLimitServer` field could only be specified for root TLS-enabled `HTTPProxies`.
+
+This alternative may be pursued in the future, as we gather more information on how users want to configure this feature.
```
sunjayBhatia, line 283: Some reasons why we went against this could be that we don't anticipate (until users tell us so) that there will be the kind of many-to-many virtualhost/route to ratelimit server relationship that you might get with other extension services and that starting with this structure gives us the ability to be flexible without adding too much complexity up front.

sunjayBhatia, line 283: Some reasons why we went against this could be that we don't anticipate (until users tell us so) that there will be the kind of many-to-many virtualhost/route to ratelimit server relationship that you might get with other extension services and that starting with this structure gives us the ability to be flexible without adding too much complexity up front.

sunjayBhatia (approved):
Looks good!

design/ratelimit-design.md
skriss: :+1: edited.

skriss:
@stevesloka let me know if you've got anything else here.

design/ratelimit-design.md
```diff
@@ -1,175 +1,502 @@
...
+    # burst defines how many additional requests above the baseline requests
+    # are allowed in a short period of time. This, along with "requests", 
+    # programs the "max_tokens" field on the Envoy local rate limit filter.
+    # See ref. https://www.envoyproxy.io/docs/envoy/latest/api-v3/type/v3/token_bucket.proto#envoy-v3-api-msg-type-v3-tokenbucket.
+    burst: 20
+  # global defines global rate limiting properties for the virtual host or route.
+  global:
+    # rateLimitDescriptors defines the descriptors to be generated and sent to the
+    # external Rate Limiting Service (RLS) for a rate limit decision.
+    rateLimitDescriptors:
```
stevesloka, line 96: Should `rateLimitDescriptors` just be named `descriptors` since we're already inside a `rateLimitPolicy`?

skriss: +1 from me, will update.

design/ratelimit-design.md
```diff
@@ -1,175 +1,502 @@
...
+This API maps closely to Envoy's rate limiting API.
 
-The rate limit service is a Go/gRPC service designed to enable generic rate limit scenarios from different types of applications. Applications request a rate limit decision based on a domain and a set of descriptors. The service reads the configuration from disk via runtime, composes a cache key, and talks to the Redis cache. A decision is then returned to the caller.
+For local rate limiting, the rate limit itself is defined inline in the `RateLimitPolicy`, as token bucket properties.
+Since local rate limits apply *per Envoy*, each pod in the Envoy daemon set will have a token bucket with these properties.
 
-New arguments to Contour:
+For global rate limiting, per Envoy's API, the `RateLimitPolicy` only defines a list of descriptors to send to an external RLS for a given request.
+All [descriptor entries defined by Envoy](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-ratelimit-action) will be supported except `metadata` and `dynamic_metadata`.
```
stevesloka, line 128: nit: Maybe define what these all would be in the API? For example `requestHeaders` can also have `skip_if_absent` which isn't in the examples. 

Also, I assume we're not going to support `SourceCluster` & `DestinationCluster` which might be new since this design doc was drafted.

skriss: OK I'll spec out the details for all of them.

stevesloka: ok to merge as-is, can do this work during implementation phase.

stevesloka (approved):

PR #3189, opened 2020-12-09
build: enable PIE build mode

jpeach:
Add a new make variable to capture common build flags and enable PIE
build mode by default. Reset the build mode for cases where we enable
race detection, since that is incompatible with PIE mode.


jpeach:
Adding PIE mode implicitly forces dynamic linking, which causes contour to fail to start because there's no dynamic loader or libc in the container.

https://github.com/golang/go/issues/40711#issuecomment-672761768

stevesloka:
Does adding the `linkmode` fix the problem?

jpeach:
> Does adding the `linkmode` fix the problem?

Not in local testing (and the external linker flags are linux-specific, which adds complexity).

youngnick (approved):
~~LGTM~~, dang, Nick, read the comments first.

youngnick (changes requested):
Oops and boo to needing dynamic linking.

jpeach:
> Oops and boo to needing dynamic linking.

yeh, this is still broken :-(

jpeach:
AFAICT, building in PIE mode also requires enabling CGO. Statically linking CGO can cause some APIs (e.g. DNS resolving) to attempt to open shared libraries. I'm not convinced that the complexity here is worth it, since we deploy into an "scratch" container that doesn't have libc or any nsswitch configuration.

https://github.com/golang/go/issues/30986

jpeach:
And, tests fail because of the CGO issue:

```
=== RUN   TestServeContextCertificateHandling/rotating_server_credentials_returns_new_server_cert
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0x63 pc=0x7ff306584a64]
runtime stack:
runtime.throw(0x16f434f, 0x2a)
	/usr/lib/golang/src/runtime/panic.go:1116 +0x74
runtime.sigpanic()
	/usr/lib/golang/src/runtime/signal_unix.go:726 +0x4ac

goroutine 37 [syscall]:
runtime.cgocall(0x15f3240, 0xc0005055c0, 0xc000366320)
	/usr/lib/golang/src/runtime/cgocall.go:133 +0x5d fp=0xc000505590 sp=0xc000505558 pc=0x40511d
net._C2func_getaddrinfo(0xc000958d30, 0x0, 0xc0001b6c00, 0xc000366320, 0x0, 0x0, 0x0)
	_cgo_gotypes.go:94 +0x57 fp=0xc0005055c0 sp=0xc000505590 pc=0x7cbb37
net.cgoLookupIPCNAME.func1(0xc000958d30, 0xa, 0xa, 0xc0001b6c00, 0xc000366320, 0x27eddc0, 0xc0005056a0, 0x7cf074)
	/usr/lib/golang/src/net/cgo_unix.go:161 +0xc5 fp=0xc000505608 sp=0xc0005055c0 pc=0x7d1905
net.cgoLookupIPCNAME(0x16baf6d, 0x3, 0x16c7eec, 0x9, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/usr/lib/golang/src/net/cgo_unix.go:161 +0x16d fp=0xc000505718 sp=0xc000505608 pc=0x7cd06d
net.cgoIPLookup(0xc00032c6c0, 0x16baf6d, 0x3, 0x16c7eec, 0x9)
	/usr/lib/golang/src/net/cgo_unix.go:218 +0x69 fp=0xc0005057b8 sp=0xc000505718 pc=0x7cd7a9
runtime.goexit()
	/usr/lib/golang/src/runtime/asm_amd64.s:1374 +0x1 fp=0xc0005057c0 sp=0xc0005057b8 pc=0x471981
created by net.cgoLookupIP
	/usr/lib/golang/src/net/cgo_unix.go:228 +0xc9
```

This failure doesn't happen if we enable CGO but don't enable PIE.

jpeach:
@youngnick Recommend we don't pursue this.

youngnick:
Agreed, while PIE is a security win, I think it's offset in this case by the increased attack surface needed to move to a non-scratch container.

To say it another way, I don't think the gain is worth the effort, and think we should mark this as not worth it. We'll just need to record that somewhere visible (alongside the results of the security audit somehow). Open to thoughts on that.

jpeach:
Closing this NTBF

PR #3191, opened 2020-12-09
internal/dag: trigger rebuild for configured secrets

### LINKED ISSUE #3190: Client certificate missing will halt all traffic indefinitely
Created a client certificate of type TLS for Envoy to use as Client cert towards the upstream.
Installed Contour 1.10, set the mentioned client certificate in configmap.

Create HTTPProxy object, set up mTLS with the upstream.
Check the proxy is valid, traffic is fine.
kubectl delete <Client cert for envoy>
Check: Proxy invalid (missing secret error message, understandable it is not there), no traffic (of any kind)
Kubectl create <Client cert for envoy> (the same data actually) 
Check: Proxy still invalid (missing secret error message still), no traffic (of any kind)

Other than this certificate rotation works, if secret is overwritten and not removed.

This situation is kept until something triggers reevaluation e.g: removal or addition of httpproxy, removal or addition of service

I would expect things to go back to normal automatically when a missing certificate is reintroduced.

Normal rotation works.

The same happens if the secret is not deleted but not there when Contour 1.10 is installed.
Contour installed with secret missing.
httpproxy added 
(no traffic)
secret added 
(still no traffic)
httpproxy recreated
(everything is fine)

**Environment:**

- Contour version:  1.10
- Kubernetes version: (use `kubectl version`): 1.15.2
- Kubernetes installer & version: 1.16.2

tsaarni:
This change adds secret references from the configuration file to the list of
secrets that will trigger DAG rebuild.  Previously only secrets referred by
HTTPProxy and Ingress resources were considered.  The result was that secrets
referred by fallback-certificate and envoy-client-certificate config file entries 
were not picked up correctly if they were created after HTTPProxy or Ingress.

Fixes #3190 

### DIFF
--- a/internal/dag/cache.go
+++ b/internal/dag/cache.go
@@ -43,6 +43,9 @@ type KubernetesCache struct {
 	// If not set, defaults to DEFAULT_INGRESS_CLASS.
 	IngressClass string
 
+	// Secrets that are referred from the configuration file.
+	ConfiguredSecretRefs []*types.NamespacedName
+
 	ingresses            map[types.NamespacedName]*v1beta1.Ingress
 	httpproxies          map[types.NamespacedName]*contour_api_v1.HTTPProxy
 	secrets              map[types.NamespacedName]*v1.Secret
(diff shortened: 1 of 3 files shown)


internal/dag/cache.go
```diff
@@ -407,6 +410,12 @@ func (kc *KubernetesCache) secretTriggersRebuild(secret *v1.Secret) bool {
 		}
 	}
 
+	for _, s := range kc.ConfiguredSecretRefs {
```
stevesloka, line 413: nit: Maybe add a comment as to why this is added?

tsaarni: I added small comment and also updated the description of `secretTriggersRebuild()` to reflect this change. Were you thinking about longer description?

skriss (approved):
LGTM, thanks @tsaarni!

stevesloka (approved):
/lgtm

stevesloka:
@tsaarni can you rebase? It looks like I can click the update branch button workflow.

tsaarni:
@stevesloka sure! The PR is now rebased.

PR #3195, opened 2020-12-10
Add ingress conformance check

sunjayBhatia:
Runs ingress conformance with sonobuoy as per example here: https://github.com/kubernetes-sigs/ingress-controller-conformance/tree/master/examples/kind


examples/contour/03-contour.yaml
```diff
@@ -40,6 +40,7 @@ spec:
         - --incluster
         - --xds-address=0.0.0.0
         - --xds-port=8001
+        - --ingress-status-address=envoy.projectcontour # TODO: inject this only when running ingress conformance
```
youngnick, line 43: This TODO seems critical 😄

youngnick (commented):
This looks great, just need that TODO sorted. I'll try and give this a spin tomorrow.

_integration/testsuite/run-ingress-conformance.sh
```diff
@@ -0,0 +1,52 @@
...
+    --kube-conformance-image=${INGRESS_CONFORMANCE_IMAGE} \
+    --plugin e2e \
+    --plugin-env e2e.INGRESS_CLASS=${INGRESS_CLASS} \
+    --plugin-env e2e.WAIT_FOR_STATUS_TIMEOUT=${WAIT_FOR_STATUS_TIMEOUT:-5m} \
+    --plugin-env e2e.TEST_TIMEOUT=${TEST_TIMEOUT:-20m} \
+    --wait
+
+${SONOBUOY} retrieve
+
+# TODO: publish/display results properly
```
sunjayBhatia, line 48: Any opinions on reporting the results etc.?

Currently this script just cats them out and exits with error if there are failures

stevesloka: That's a good start I I think to just cat them out and fail if an issue is encountered.

youngnick (commented):
This seems like a win to me - it's certainly useful to be able to run the conformance tests on demand. What else do we need to do here?

sunjayBhatia:
If this looks ok to everyone I can comment out/remove the CI check and let this be mergeable

stevesloka:
What's covered here vs the current integration tests. It would be nice to settle on a single way to run these tests along will contributing back that work to the upstream set of tests.

sunjayBhatia:
> What's covered here vs the current integration tests. It would be nice to settle on a single way to run these tests along will contributing back that work to the upstream set of tests.

Looks like in our existing integration tests there are some common path prefix stuff that may become duplicated with the Ingress conformance tests, other than that I don't think there is much content overlap (maybe a tiny bit with TLS+hostname verification)? See https://github.com/kubernetes-sigs/ingress-controller-conformance/tree/master/features, Ingress tests verify default backends, host name matching, Ingress class name, load balancing, and path matching rules.

Also looks like https://github.com/kubernetes-sigs/service-apis/pull/516 is taking a completely different approach with a simpler Go test framework, so we potentially have 3 different high level test frameworks/strategies unfortunately. Theoretically based on the comments on that PR Ingress conformance and Service APIs conformance could converge

So I guess do we want to merge as-is and refactor or archive for now and refactor later once things have settled down? (Contour integration tests to match upstream + contribute to upstream, in no particular order)

stevesloka:
Yeah I think it's fine to merge and use them in parallel if it's adding value.

.github/workflows/prbuild.yaml
```diff
@@ -17,6 +17,18 @@ jobs:
       - name: integration tests
         run: |
           make integration
+  # TODO: re-enable once Ingress v1 support is complete
```
stevesloka, line 20: So just to make sure I understand, this adds the tests but doesn't execute unless you do it manually? Is there an easy way to comment out the tests that will always fail / or are not yet supported?

sunjayBhatia: > So just to make sure I understand, this adds the tests but doesn't execute unless you do it manually?

yep, there is an `--e2e-focus` flag available for `sonobuoy`, testing that out

sunjayBhatia: so far it doesn't look like that flag is respected, also looking through the Ingress conformance code I don't see anywhere it grabs the `E2E_FOCUS|SKIP` variables, doesn't have a flag for that either

from the `sonobuoy run` help

```
--e2e-focus string                                         Specify the E2E_FOCUS flag to the conformance tests. Overrides --mode. (default "\\[Conformance\\]")
--e2e-skip string                                          Specify the E2E_SKIP flag to the conformance tests. Overrides --mode. (default "\\[Disruptive\\]|NoExecuteTaintManager")
```

sunjayBhatia: Also there is only one actual `go` test named `TestSuite` so using go filtering (`-test.run` flag) doesn't work

sunjayBhatia: Using `cucumber` scenario tags works, but that logic isn't piped through the e2e script/docker image etc., would have to do some more plumbing

Added https://github.com/kubernetes-sigs/ingress-controller-conformance/issues/86

youngnick (approved):
LGTM on the basis that having a way to run the current conformance tests seems useful regardless of if it's blocking in CI. I agree that we need to standardise on our integration and end-to-end testing, but I think it's okay to merge for now and sort this out later.

sunjayBhatia:
cc @skriss @stevesloka when you get a change ready for a look

skriss (approved):
LGTM, this'll be useful as we work through full Ingress v1 support and hopefully we can get to green soon.

Issue #3200, opened 2020-12-11
Remove all uses of Envoy xDS v2 configs


sunjayBhatia:
We still have references to Envoy `v2` xDS APIs in various places in the code, from types we are using, links to Envoy documentation, and some APIs.

From a quick `go list` and grep:

```
$ go list -f '{{ join .Imports "\n" }}' ./cmd/contour | grep -E 'envoy.*v2'
github.com/envoyproxy/go-control-plane/envoy/api/v2
github.com/envoyproxy/go-control-plane/pkg/resource/v2
```

```
$ grep -r envoyproxy\.io apis pkg cmd internal | grep v2
apis/projectcontour/v1/httpproxy.go:// [v2](https://www.envoyproxy.io/docs/envoy/latest/api-v2/service/auth/v2/external_auth.proto)
pkg/config/parameters.go:       // See https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/network/http_connection_manager/v2/http_connection_manager.proto#envoy-api-field-config-filter-network-http-connection-manager-v2-httpconnectionmanager-request-timeout
pkg/config/parameters.go:       // See https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/core/protocol.proto#envoy-api-field-core-httpprotocoloptions-idle-timeout
pkg/config/parameters.go:       // See https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/network/http_connection_manager/v2/http_connection_manager.proto#envoy-api-field-config-filter-network-http-connection-manager-v2-httpconnectionmanager-stream-idle-timeout
pkg/config/parameters.go:       // See https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/core/protocol.proto#envoy-api-field-core-httpprotocoloptions-max-connection-duration
pkg/config/parameters.go:       // See https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/network/http_connection_manager/v2/http_connection_manager.proto#envoy-api-field-config-filter-network-http-connection-manager-v2-httpconnectionmanager-drain-timeout
pkg/config/parameters.go:       // See https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/cluster.proto#enum-cluster-dnslookupfamily
internal/dag/dag.go:    // See https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/cds.proto#envoy-api-enum-cluster-lbpolicy
[...]

PR #3202, opened 2020-12-14
cmd/contour: update CLI commands to use xDS v3

skriss:
Updates the CLI commands to use xDS v3. Also fixes a panic
when no user agent build version is present.

Updates #3200.

### DIFF
--- a/internal/xds/v3/callbacks.go
+++ b/internal/xds/v3/callbacks.go
@@ -40,7 +40,11 @@ func NewRequestLoggingCallbacks(log logrus.FieldLogger) envoy_server_v3.Callback
 func logDiscoveryRequestDetails(l logrus.FieldLogger, req *envoy_service_discovery_v3.DiscoveryRequest) *logrus.Entry {
 	log := l.WithField("version_info", req.VersionInfo).WithField("response_nonce", req.ResponseNonce)
 	if req.Node != nil {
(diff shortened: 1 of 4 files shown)


internal/xds/v3/callbacks.go
```diff
@@ -40,7 +40,11 @@ func NewRequestLoggingCallbacks(log logrus.FieldLogger) envoy_server_v3.Callback
 func logDiscoveryRequestDetails(l logrus.FieldLogger, req *envoy_service_discovery_v3.DiscoveryRequest) *logrus.Entry {
 	log := l.WithField("version_info", req.VersionInfo).WithField("response_nonce", req.ResponseNonce)
 	if req.Node != nil {
-		log = log.WithField("node_id", req.Node.Id).WithField("node_version", fmt.Sprintf("v%d.%d.%d", req.Node.GetUserAgentBuildVersion().Version.MajorNumber, req.Node.GetUserAgentBuildVersion().Version.MinorNumber, req.Node.GetUserAgentBuildVersion().Version.Patch))
+		log = log.WithField("node_id", req.Node.Id)
+
+		if bv := req.Node.GetUserAgentBuildVersion(); bv != nil && bv.Version != nil {
```
sunjayBhatia, line 45: Good catch! Since this can panic, might be worth adding a test case to cover this

sunjayBhatia (commented):
```
go list -f '{{ join .Imports "\n" }}' ./... | grep -E 'envoy.*v2'
```

Doesn't return any v2 packages 👍 

Just one comment about possibly adding test coverage

internal/xds/v3/callbacks.go
skriss: yep good call, will add

sunjayBhatia (approved):

PR #3203, opened 2020-12-14
change remaining v2 xDS references to v3

### LINKED ISSUE #3200: Remove all uses of Envoy xDS v2 configs

skriss:
Updates comments, documentation, etc. to exclusively reference
xDS v3.

Closes #3200.

### DIFF
--- a/apis/projectcontour/v1alpha1/extensionservice.go
+++ b/apis/projectcontour/v1alpha1/extensionservice.go
@@ -20,11 +20,19 @@ import (
 
 // ExtensionProtocolVersion is the version of the GRPC protocol used
 // to access extension services. The only version currently supported
-// is "v2".
+// is "v3".
 type ExtensionProtocolVersion string
 
-// SupportProtocolVersion2 requests the "v2" support protocol version.
-const SupportProtocolVersion2 ExtensionProtocolVersion = "v2"
+const (
+	// SupportProtocolVersion2 requests the "v2" support protocol version.
+	//
+	// Deprecated: this protocol version is no longer supported and the
+	// constant is retained for backwards compatibility only.
+	SupportProtocolVersion2 ExtensionProtocolVersion = "v2"
+
+	// SupportProtocolVersion3 requests the "v3" support protocol version.
+	SupportProtocolVersion3 ExtensionProtocolVersion = "v3"
+)
 
 // ExtensionServiceTarget defines an Kubernetes Service to target with
 // extension service traffic.
@@ -88,11 +96,11 @@ type ExtensionServiceSpec struct {
 
 	// This field sets the version of the GRPC protocol that Envoy uses to
 	// send requests to the extension service. Since Contour always uses the
-	// v2 Envoy API, this is currently fixed at "v2". However, other
+	// v3 Envoy API, this is currently fixed at "v3". However, other
 	// protocol options will be available in future.
 	//
 	// +optional
-	// +kubebuilder:validation:Enum=v2
+	// +kubebuilder:validation:Enum=v3
 	ProtocolVersion ExtensionProtocolVersion `json:"protocolVersion,omitempty"`
 }
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -148,8 +148,8 @@ Examples:
 ```
 contour_api_v1 "github.com/projectcontour/contour/apis/projectcontour/v1"
 contour_api_v1alpha1 "github.com/projectcontour/contour/apis/projectcontour/v1alpha1"
-envoy_v2 "github.com/projectcontour/contour/internal/envoy/v2"
-xdscache_v2 "github.com/projectcontour/contour/internal/xdscache/v2"
+envoy_v3 "github.com/projectcontour/contour/internal/envoy/v3"
+xdscache_v3 "github.com/projectcontour/contour/internal/xdscache/v3"
 ```   
  
 ### Pre commit CI
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -129,12 +129,8 @@ type ExtensionServiceReference struct {
 }
 
 // AuthorizationServer configures an external server to authenticate
-// client requests. The external server must implement the Envoy
-// external authorization GRPC protocol. Currently, the
-// [v2](https://www.envoyproxy.io/docs/envoy/latest/api-v2/service/auth/v2/external_auth.proto)
-// protocol is always used, but authorization server authors should implement
-// the v3 protocol as well in the expectation that it will be supported
-// in future.
+// client requests. The external server must implement the v3 Envoy
+// external authorization GRPC protocol (https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/auth/v3/external_auth.proto).
 type AuthorizationServer struct {
 	// ExtensionServiceRef specifies the extension resource that will authorize client requests.
 	//
--- a/examples/contour/01-crds.yaml
+++ b/examples/contour/01-crds.yaml
@@ -49,9 +49,9 @@ spec:
                 - h2c
                 type: string
               protocolVersion:
-                description: This field sets the version of the GRPC protocol that Envoy uses to send requests to the extension service. Since Contour always uses the v2 Envoy API, this is currently fixed at "v2". However, other protocol options will be available in future.
+                description: This field sets the version of the GRPC protocol that Envoy uses to send requests to the extension service. Since Contour always uses the v3 Envoy API, this is currently fixed at "v3". However, other protocol options will be available in future.
                 enum:
-                - v2
+                - v3
                 type: string
               services:
                 description: Services specifies the set of Kubernetes Service resources that receive GRPC extension API requests. If no weights are specified for any of the entries in this array, traffic will be spread evenly across all the services. Otherwise, traffic is balanced proportionally to the Weight field in each entry.
--- a/examples/render/contour.yaml
+++ b/examples/render/contour.yaml
@@ -170,9 +170,9 @@ spec:
                 - h2c
                 type: string
               protocolVersion:
-                description: This field sets the version of the GRPC protocol that Envoy uses to send requests to the extension service. Since Contour always uses the v2 Envoy API, this is currently fixed at "v2". However, other protocol options will be available in future.
+                description: This field sets the version of the GRPC protocol that Envoy uses to send requests to the extension service. Since Contour always uses the v3 Envoy API, this is currently fixed at "v3". However, other protocol options will be available in future.
                 enum:
-                - v2
+                - v3
                 type: string
               services:
                 description: Services specifies the set of Kubernetes Service resources that receive GRPC extension API requests. If no weights are specified for any of the entries in this array, traffic will be spread evenly across all the services. Otherwise, traffic is balanced proportionally to the Weight field in each entry.
--- a/internal/dag/dag.go
+++ b/internal/dag/dag.go
@@ -359,7 +359,7 @@ func (v *VirtualHost) Valid() bool {
 type SecureVirtualHost struct {
 	VirtualHost
 
-	// TLS minimum protocol version. Defaults to envoy_api_v2_auth.TlsParameters_TLS_AUTO
+	// TLS minimum protocol version. Defaults to envoy_tls_v3.TlsParameters_TLS_AUTO
 	MinTLSVersion string
 
 	// The cert and key for this host.
@@ -508,7 +508,7 @@ type Cluster struct {
 	UpstreamValidation *PeerValidationContext
 
 	// The load balancer type to use when picking a host in the cluster.
-	// See https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/cds.proto#envoy-api-enum-cluster-lbpolicy
+	// See https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#enum-config-cluster-v3-cluster-lbpolicy
 	LoadBalancerPolicy string
 
 	// Cluster http health check policy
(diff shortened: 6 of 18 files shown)


skriss:
I *think* this got all the remaining v2 refs but still double-checking.

apis/projectcontour/v1alpha1/extensionservice.go
```diff
@@ -20,11 +20,11 @@ import (
...
 // ExtensionProtocolVersion is the version of the GRPC protocol used
 // to access extension services. The only version currently supported
-// is "v2".
+// is "v3".
 type ExtensionProtocolVersion string
 
-// SupportProtocolVersion2 requests the "v2" support protocol version.
-const SupportProtocolVersion2 ExtensionProtocolVersion = "v2"
+// SupportProtocolVersion3 requests the "v3" support protocol version.
+const SupportProtocolVersion3 ExtensionProtocolVersion = "v3"
```
skriss, line 27: I assumed that since this is in the `v1alpha` API, we can make a breaking change here by renaming it/changing the default value.

sunjayBhatia:
Not sure if you want to cover it in this change, but this is an odd repeated test string that could be worth changing (don't think it needs to reference Envoy at all) since we're doing this:

```
$ grep -r -E 'envoy.*v2' internal/
internal//featuretests/v3/replaceprefix_test.go:					{Replacement: "/api/envoy_api_v2"},
internal//featuretests/v3/replaceprefix_test.go:					{Prefix: "/api", Replacement: "/api/envoy_api_v2"},
internal//featuretests/v3/replaceprefix_test.go:						Action: withPrefixRewrite(routeCluster("default/kuard/8080/da39a3ee5e"), "/api/envoy_api_v2/"),
internal//featuretests/v3/replaceprefix_test.go:						Action: withPrefixRewrite(routeCluster("default/kuard/8080/da39a3ee5e"), "/api/envoy_api_v2"),
internal//featuretests/v3/replaceprefix_test.go:					{Prefix: "/foo", Replacement: "/api/envoy_api_v2"},
internal//featuretests/v3/replaceprefix_test.go:				Conditions: matchconditions(prefixMatchCondition("/envoy_api_v2")),
internal//featuretests/v3/replaceprefix_test.go:						{Prefix: "/envoy_api_v2", Replacement: "/api/envoy_api_v2"},
internal//featuretests/v3/replaceprefix_test.go:						Match:  routePrefix("/envoy_api_v2/"),
internal//featuretests/v3/replaceprefix_test.go:						Action: withPrefixRewrite(routeCluster("default/kuard/8080/da39a3ee5e"), "/api/envoy_api_v2/"),
internal//featuretests/v3/replaceprefix_test.go:						Match:  routePrefix("/envoy_api_v2"),
internal//featuretests/v3/replaceprefix_test.go:						Action: withPrefixRewrite(routeCluster("default/kuard/8080/da39a3ee5e"), "/api/envoy_api_v2"),
internal//featuretests/v3/replaceprefix_test.go:						Match:  routePrefix("/envoy_api_v2"),
internal//featuretests/v3/replaceprefix_test.go:						{Prefix: "/envoy_api_v2/container-sandbox", Replacement: "/artifactory/api/docker/container-sandbox/envoy_api_v2"},
internal//featuretests/v3/replaceprefix_test.go:						{Prefix: "/envoy_api_v2/container-release", Replacement: "/artifactory/api/docker/container-release/envoy_api_v2"},
internal//featuretests/v3/replaceprefix_test.go:						{Prefix: "/envoy_api_v2/container-external", Replacement: "/artifactory/api/docker/container-external/envoy_api_v2"},
internal//featuretests/v3/replaceprefix_test.go:						{Prefix: "/envoy_api_v2/container-public", Replacement: "/artifactory/api/docker/container-public/envoy_api_v2"},
internal//featuretests/v3/replaceprefix_test.go:				{Name: "routes", Conditions: matchconditions(prefixMatchCondition("/envoy_api_v2/container-sandbox"))},
internal//featuretests/v3/replaceprefix_test.go:				{Name: "routes", Conditions: matchconditions(prefixMatchCondition("/envoy_api_v2/container-release"))},
internal//featuretests/v3/replaceprefix_test.go:				{Name: "routes", Conditions: matchconditions(prefixMatchCondition("/envoy_api_v2/container-external"))},
internal//featuretests/v3/replaceprefix_test.go:				{Name: "routes", Conditions: matchconditions(prefixMatchCondition("/envoy_api_v2/container-public"))},
internal//featuretests/v3/replaceprefix_test.go:						Match: routePrefix("/envoy_api_v2/container-sandbox/"),
internal//featuretests/v3/replaceprefix_test.go:							"/artifactory/api/docker/container-sandbox/envoy_api_v2/"),
internal//featuretests/v3/replaceprefix_test.go:						Match: routePrefix("/envoy_api_v2/container-sandbox"),
internal//featuretests/v3/replaceprefix_test.go:							"/artifactory/api/docker/container-sandbox/envoy_api_v2"),
internal//featuretests/v3/replaceprefix_test.go:						Match: routePrefix("/envoy_api_v2/container-release/"),
internal//featuretests/v3/replaceprefix_test.go:							"/artifactory/api/docker/container-release/envoy_api_v2/"),
internal//featuretests/v3/replaceprefix_test.go:						Match: routePrefix("/envoy_api_v2/container-release"),
internal//featuretests/v3/replaceprefix_test.go:							"/artifactory/api/docker/container-release/envoy_api_v2"),
internal//featuretests/v3/replaceprefix_test.go:						Match: routePrefix("/envoy_api_v2/container-public/"),
internal//featuretests/v3/replaceprefix_test.go:							"/artifactory/api/docker/container-public/envoy_api_v2/"),
internal//featuretests/v3/replaceprefix_test.go:						Match: routePrefix("/envoy_api_v2/container-public"),
internal//featuretests/v3/replaceprefix_test.go:							"/artifactory/api/docker/container-public/envoy_api_v2"),
internal//featuretests/v3/replaceprefix_test.go:						Match: routePrefix("/envoy_api_v2/container-external/"),
internal//featuretests/v3/replaceprefix_test.go:							"/artifactory/api/docker/container-external/envoy_api_v2/"),
internal//featuretests/v3/replaceprefix_test.go:						Match: routePrefix("/envoy_api_v2/container-external"),
internal//featuretests/v3/replaceprefix_test.go:							"/artifactory/api/docker/container-external/envoy_api_v2"),
```

skriss:
Yeah, I saw that and ignored it since it was just test data, but may as well get rid of it too. Will update.

skriss:
OK updated that one too. Looks like that weird string was an inadvertent copy-paste error from renaming some imports in prep for the v3 migration anyway, so good to get rid of :)

sunjayBhatia (approved):
I dont see any more references to Envoy xDS v2 APIs, checked with:

```
grep -r -E 'envoy.*v2' _integration apis cmd docs examples hack internal pkg site/docs/main
```

LGTM, assuming the Ext-Authz/v1alpha1 version API is cool to change (https://github.com/projectcontour/contour-authserver implements v2 and v3 already which is 👍 )

apis/projectcontour/v1alpha1/extensionservice.go
jpeach: My preference would be to no make a breaking change unless it's really required. Now we are in the v3 API, we can simply send the version in the proto message.

apis/projectcontour/v1alpha1/extensionservice.go
```diff
@@ -88,11 +88,11 @@ type ExtensionServiceSpec struct {
 
 	// This field sets the version of the GRPC protocol that Envoy uses to
 	// send requests to the extension service. Since Contour always uses the
-	// v2 Envoy API, this is currently fixed at "v2". However, other
+	// v3 Envoy API, this is currently fixed at "v3". However, other
 	// protocol options will be available in future.
 	//
 	// +optional
-	// +kubebuilder:validation:Enum=v2
+	// +kubebuilder:validation:Enum=v3
```
jpeach, line 95:
```suggestion
	// +kubebuilder:validation:Enum=v2,v3
```

Either v2 or v3 is OK since it's whatever the service happened to implement. Might be out of the control of the operator.

skriss: Is v2 still going to be a valid option when Envoy drops xDS v2 support? I was assuming it would drop v2 support for the ext auth API as well, so we wouldn't want to allow it to be configured.

skriss: @jpeach @youngnick @stevesloka I need some further input on this, so we can get this PR in before 1.11.

stevesloka: I think it's fine to only support v3. v2 will be dropped in a month, also, the field isn't even implemented in the code, it's only in the API, so even if you configure it, it's not going to change any logic.

stevesloka: We should just drop the field. How far off is v4 and the api is alpha, so do we even need to carry it forward?

sunjayBhatia: According to https://docs.google.com/document/d/1afQ9wthJxofwOMCAle8qF-ckEfNWUuWtxbmdhxji0sI/edit# (still in review/not sure the status of the proposal) it seems like major version bumps/deprecations aren't going to be yearly anymore, `v4alpha` changes have started (https://github.com/envoyproxy/envoy/tree/7d348d1b65d3c5ae58d0259e397fbd3df8686f11/api/envoy/admin/v4alpha) but I would guess there is time

+1 to drop the field and bring it back for v3->v4+ if needed

apis/projectcontour/v1alpha1/extensionservice.go
skriss: I changed this to retain the v2 constant, but as of now this code is still statically programming Envoy to v3.

apis/projectcontour/v1alpha1/extensionservice.go
youngnick: This API is alpha, so we can just drop the field. I agree that we should do that - I think we need to consider where we put settings that are for a particular type of ExtensionService (ie one that needs gRPC, or one that connects to a tracing service and needs properties set.) I don't think top-level fields are right for that, so we probably will need to remove this anyway. Let's just remove it now, and come back and add things if we need to later, in a more sustainable way.

skriss: I *believe* that we'd need to bump this API to `v1alpha2` in order to drop a field, per [rule 1](https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-parts-of-the-api).

I can look at doing that, but maybe a simpler solution for now is to leave it as I currently have it in this PR - the field still exists, but it's not being used, and Envoy is being programmed to use `v3`. Then when we're ready to make a rev to this API, we can drop the field if we still think that makes sense?

stevesloka: I don't think a version rev at this point is needed. My only thing is that it's not implemented so what are we breaking? =)

We can just move on from here, but need to find a place to put this type of discussion for the future without TODO's all over the place that never get used. Should we have a spot in the wiki? Or an issue?

apis/projectcontour/v1alpha1/extensionservice.go
```diff
@@ -20,11 +20,19 @@ import (
 
 // ExtensionProtocolVersion is the version of the GRPC protocol used
 // to access extension services. The only version currently supported
-// is "v2".
+// is "v3".
```
stevesloka, line 23: nit: Should this now say v2 is supported as well?

stevesloka (approved):
/lgtm pending the other nit comment.

apis/projectcontour/v1alpha1/extensionservice.go
stevesloka: I read this wrong, it's correct.

PR #3204, opened 2020-12-14
cmd/contour: Log correct error on failed tls.Config creation

sunjayBhatia:
Should log `lerr` since that was the error from loading the config


skriss (approved):
LGTM

youngnick (approved):
LGTM< nice.

PR #3208, opened 2020-12-16
internal/certgen: Ensure certgen output files mode

sunjayBhatia:
- Write PEM and YAML files with 0600 mode to ensure r/w permissions are
  restricted
- Delete existing files when overwrite mode is specified to ensure
  correct mode is set and a previous insecure mode is not persisted
- Add file mode test


youngnick (approved):
LGTM

skriss (approved):
LGTM!

PR #3211, opened 2020-12-16
examples: update Envoy to 1.16.2

skriss:

The compatibility matrix will need to be updated when 1.10.1 and 1.11 are cut.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ SRCDIRS := ./cmd ./internal ./apis
 LOCAL_BOOTSTRAP_CONFIG = localenvoyconfig.yaml
 SECURE_LOCAL_BOOTSTRAP_CONFIG = securelocalenvoyconfig.yaml
 PHONY = gencerts
-ENVOY_IMAGE = docker.io/envoyproxy/envoy:v1.16.1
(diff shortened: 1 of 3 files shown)


danehans:
@skriss should "Note: These extensions are tested against Envoy v1.16.1." in https://projectcontour.io/resources/envoy/ get updated to reference v1.16.2?

danehans:
xref: https://github.com/projectcontour/contour-operator/pull/165

skriss:
> @skriss should "Note: These extensions are tested against Envoy v1.16.1." in https://projectcontour.io/resources/envoy/ get updated to reference v1.16.2?

Well, @stevesloka just did this testing, and it was against 1.16.1, so not sure we just want to update the text if we haven't actually retested. @stevesloka thoughts? In any case, we can make that change (if needed) as part of updating the compat matrix for the release.

youngnick (approved):

stevesloka:
Yup my testing was technically on 1.16.1, but I'd expect it to work fine with 1.16.2.

stevesloka (approved):
/lgtm

stevesloka:
I just built a new image: `stevesloka/envoy:v1.16.2.contour`

PR #3213, opened 2020-12-17
design: Service APIs implementation

youngnick:
This is a WIP of the Service APIs implementation. It's been slow going, as I've been trying to check as I went that I've got the Service APIs intent correct, and that the way we'll implement is relatively consistent with Contour's current state.

Interested mostly in feedback on the high-level parts, did I miss anything?

There's quite a bit more detail required, particular for HTTPRoute and TLSRoute.

Very interested to hear about any other areas you all think I should cover as part of this design.

### DIFF
--- a/design/service-apis-implementation.md
+++ b/design/service-apis-implementation.md
@@ -184,10 +184,12 @@ This behavior is similar to the HTTPProxy processor's `validHTTPProxies()` metho
 
 #### Listener merging rules
 
+TODO: The basic principle is to allow for people to specify lots of names across a set of Gateways, or within one Gateway.
 
+Listeners that match on ProtocolType and PortNumber can be merged, using Hostname as a discriminator.
 
-
-
+TODO: There are some edge cases to check into here because TLS config can be specified both here and in the Route.
+Also need to check for if you can mix-and-match between Routes and here, using here as a fallback?
 
 ### HTTPRoute
 
@@ -213,13 +215,12 @@ The same applies to changes in CRDs, YAML examples, and so on.
 Ideally the changes should be made in sequence so that the work required to implement this design can be done incrementally, possibly in parallel.
 
 ## Alternatives Considered
-If there are alternative high level or detailed designs that were not pursued they should be called out here with a brief explanation of why they were not pursued.
+TODO: This will need further explanation of what the solution would look like if we didn't make one GatewayClass == One Contour.
+
+TODO: Some discussion of why layer 7 only.
 
 ## Security Considerations
-If this proposal has an impact to the security of the product, its users, or data stored or transmitted via the product, they must be addressed here.
-
-## Compatibility
-A discussion of any compatibility issues that need to be considered
+TODO: I can't think of any yet.
 
 ## Implementation
 A description of the implementation, timelines, and any resources that have agreed to contribute.
--- a/design/service-apis-implementation.md
+++ b/design/service-apis-implementation.md
@@ -42,44 +42,163 @@ If Project Contour (the organisation) does add support for TCP and UDP forwardin
 
 ## High-Level Design
 
-### Contour watches GatewayClass objects
-Looking for entries with `spec.controller` equal to `projectcontour.io/contour` by default.
+### Service APIs implementation surface area
+
+Contour's support for the Service APIs will include support for the HTTPRoute and TLSRoute objects only, not any other type of Route - this includes no TCPRoute or UDPRoute support.
+Contour is a layer 7 ingress controller, and the layer 4 load balancing implies by TCPRoute and UDPRoute is out of scope for this tool (that is, `projectcontour/contour`).
+Project Contour (the organisation) may investigate a Service-APIs based layer 4 solution in the future, but that effort will not be in `projectcontour/contour`.
+
+### Contour Gateway model
+
+Contour considers a GatewayClass and associated set of Gateways to describe a single Envoy deployment, deployed and managed by something outside of itself.
+As such, Contour will error out if it is asked to watch more than one GatewayClass.
+Inside its GatewayClass, Contour will merge Listeners within a single Gateway, or across Gateways in the GatewayClass. (See the detailed design for the exact rules Contour will use for this.)
+
+### Interrelated watches
+
+The Service APIs are a set of interrelated Kubernetes objects, where a change in one object can mean that the scope of objects Contour is interested in will change.
+Because of this, Contour will watch all events associated with GatewayClass, Gateway, HTTPRoute, TLSRoute, and BackendPolicy objects, and filter the objects it takes action on internally, rather than using a filtered watch (as those are costly to set up and tear down).
+
+### Combining Service APIs with other configuration
+
+When it calculates the Envoy configuration using the DAG, Contour layers different types of configuration in order.
+So, currently, Ingress is overwritten by HTTPProxy, so if an Ingress and a HTTPProxy specify the exact same route, the HTTPProxy will win.
+
+Once we have the Service-APIs available as well, we have to choose the order.
+In this design, I suggest having the order be Ingress is overwritten by HTTPProxy, is overwritten by the service-apis.
+I could see reversing HTTPProxy and the Service APIs here, but I think this acknowledges that the Service APIs are really an evolution of the ideas in HTTPProxy, and will probably end up being the community standard.
+
+### Status management
+
+Each object in the Service APIs set has its own status definition. In general, Contour will update the status of any object that comes into scope with its details, ensuring that the `observedGeneration` field is set correctly.
+When objects fall out of scope, any status set by Contour will not be removed.
+It's expected that things that check the status will also check the `observedGeneration` to check if the status information is up-to-date with the current object generation.
+
+### Per-object design notes
+
+#### GatewayClass
+
+Contour will look for entries with `spec.controller` equal to `projectcontour.io/contour` by default.
 Contour can change this value to any `projectcontour.io/<value>` with a config setting (similar to how `ingress.class` works today.)
 Importantly, this means that changing this values requires a Contour restart; this means that the GatewayClass watcher can have this value baked in at runtime.
+
+If Contour finds more than one matching GatewayClass, this is a conflict, and Contour will use the oldest GatewayClass (by creation timestamp).
+All GatewayClasses will be updated with status conditions indicating if they are in use.
+
 The output of this watcher is a list of GatewayClass name/namespace details.
 
-### Contour watches Gateway objects
-Looking for entries with known GatewayClass name/namespace details, from the previous watch.
+#### Gateway
+Contour will watch for entries with known GatewayClass name/namespace details, from the previous watch.
 
-Listeners inside Gateways will be merged where possible, as will listeners across Gateways.
-Conflicts within a Gateway will result in the relevant Listeners both being rejected.
-Conflicts between Gateways will result in the oldest Gateway (by creation timestamp) being kept.
-The rejected Listener will have its status updated with the reason and the name of the conflicting Gateway.
+Contour is not able to specify the address that the Envoy deployment will listen on, and so having any entries in Gateway's `spec.addresses` stanza will result in Contour marking those addresses as invalid in the Gateway's status.
+Contour expects the `spec.addresses` field to be empty.
+
+For Listeners:
+- Listeners inside Gateways will be merged where possible, as will listeners across Gateways.
+- Conflicts within a Gateway will result in the relevant Listeners both being rejected.
+- Conflicts between Gateways will result in the oldest Gateway (by creation timestamp) being kept.
+- The rejected Listener will have its status updated with the reason and the name of the conflicting Gateway.
+- Listeners are considered conflicting if all the fields out of `hostname`, `port`, `protocol`, and `tls` match.
+- Further merging rules are specified in the detailed design below.
 
 The output of this watcher is both Envoy configuration, and a list of kind/name/namespace details for Routes to watch.
 
 ### Contour watches HTTPRoutes
-HTTPRoutes refereenced in Gateways will be watched, and associated routes configured.
+HTTPRoutes matching a label selector in the Gateway `spec.listeners[].routes` will be watched, and associated routes configured.
+HTTPRoutes referencing Gateways that fit into the Contour GatewayClass will be watched, and associated routes configured.
 Errors or conflicts here will render that section of the config invalid.
 Other valid sections will still be passed to Envoy.
+For each invalid section, Contour will update status information with the section and the reason.
 
 The output of this watcher is Envoy configuration.
 
 ### Contour watches TLSRoutes
-TLSRoutes referenced in Gateways will be watched and associated routes configured in Contour.
+TLSRoutes matching a label selector in the Gateway `spec.listeners[].routes` will be watched, and associated routes configured.
+TLSRoutes referencing Gateways that fit into the Contour GatewayClass will be watched, and associated routes configured.
+
 Errors or conflicts here will render that section of the config invalid.
 Other valid sections will still be passed to Envoy.
 
 The output of this watcher is Envoy configuration.
 
-### Interrelated watches
-
-TODO: Design how the information flow will work. What watches? How does the GatewayClass pass object names to the Gateway, for example?
-
-In general, because of the dynamic, interrelated nature of these resources, Contour will need to watch all events associated with all objects, and filter internally, rather than using a filtered watch (as those are costly to set up and tear down).
-
-
 ## Detailed Design
+
+### Configuration
+Contour will have an entry added to the config file for the GatewayClass string, `gatewayClassID`.
+By default, this is set to `contour`.
+This parameter is concatenated to `projectcontour.io/` to produce the string that Contour will check the `spec.controller` field of GatewayClass entries for.
+So, by default, Contour will check for `projectcontour.io/contour`.
+
+
+### Code changes
+Contour already has support for importing the Service APIs objects into its Kubernetes cache.
+However, for some types, we need to be able to keep some details - this design suggests making those details properties of the cache, as `IngressClass` is currently.
+
+For ingestion, the general pattern currently is that the EventHandler in `internal/contour/handler.go` handles all objects, and calls out to the KubernetesCache from `internal/dag/cache.go`.
+The current pattern across most of the objects is:
+- check if the object is in scope
+- add it to the cache
+- indicate with a return value if the add should result in a DAG (and consequently an Envoy config) rebuild.
+
+This pattern is applicable to all the Service APIs objects, with some differences for GatewayClass.
+
+There are already internal fields in the cache to hold GatewayClass and some other service-apis objects.
+We will use this scaffolding to build the actual functionality on top.
+
+### GatewayClass
+
+When processing GatewayClass, Contour will use the configuration value `gatewayClassID` to build a controller URL of the form `projectcontour.io/<gatewayClassID>`.
+By default, this value will be `projectcontour.io/contour`.
+
+At startup, if this value is specified, Contour will fetch all GatewayClass objects and check the `spec.controller` field of each.
+Contour will consider the oldest GatewayClass with a matching `spec.controller` field to be the canonical GatewayClass for that instance of Contour, and update its status.
+Other GatewayClasses than the oldest will also have their status updated.
+The status update will be adding a condition to the GatewayClass `spec.conditions` field with a type of `Accepted`, where the actually in use GatewayClass will be `Status: true` and any other GatewayClasses will be `status: false`.
+
+Inside Contour, the KubernetesCache will have a `GatewayClassDetails` field added, holding a struct that specifies the name of the active GatewayClass that Gateways may reference, and the creation timestamp for that object.
+If this field is empty, then no GatewayClasses match, there is no active GatewayClass, and no Gateways will be checked.
+
+Note that Contour's status update machinery should ensure that only status updates that are changes are actually applied.
+
+A DAG rebuild should only be kicked off if the GatewayClass value in the KubernetesCache changed.
+
+### Gateway
+
+When ingesting Gateways, Contour will import all Gateways into the KubernetesCache.
+
+#### Validity
+
+Before processing Gateways, Contour will run a quick check across all gateways to look for conflicting or out-of-scope Listeners.
+
+Contour will use the `GatewayClassDetails` field in the Kubernetes cache to check each object's `spec.gatewayClassName` field against the GatewayClass name.
+If they do not match, the Gateway will be skipped from processing.
+A debug-level log line should be output here to indicate the skip.
+
+Listener **conflict** is defined as two listeners sharing 
+Listeners that cannot be merged *within a Gateway* are invalid and will be rejected, and their status updated accordingly.
+
+When Listeners cannot be merged *across Gateways* the **oldest** Gateway's Listeners (by creation timestamp) will be accepted, and other Listeners will be rejected.
+An accepted Listener will also have a condition placd on it to indicate the Gateway that contains Listeners in conflict with it.
+
+This behavior is similar to the HTTPProxy processor's `validHTTPProxies()` method.
+
+#### Listener merging rules
+
+
+
+
+
+
+### HTTPRoute
+
+### TLSRoute
+
+
+
+### Other concerns
+
+TODO: What do we do about updating status information? Probably just rely on whatever provisions Envoy to put the reachability info in the Gateway Status.
+
 
 For each object, I need to specify:
 How the watch will work
--- a/design/service-apis-implementation.md
+++ b/design/service-apis-implementation.md
@@ -20,7 +20,7 @@ The cluster operator and application developer are basically the same as Contour
 In terms of the APIs themsleves, the Service APIs have 3 primary API resources (taken from the service-apis docs site):
 
 - GatewayClass defines a set of gateways with a common configuration and behavior.
-- Gateway requests a point where traffic can be translated to Services within the cluster.
+- Gateway requests a point where traffic can be translated to Services within the cluster, using an internal Listener construct.
 - Routes describe how traffic coming via the Gateway maps to the Services.
 
 In Contour, we've previously solved a lot of the same problems with HTTPProxy (and IngressRoute before it).
@@ -88,33 +88,40 @@ All GatewayClasses will be updated with status conditions indicating if they are
 The output of this watcher is a list of GatewayClass name/namespace details.
 
 #### Gateway
-Contour will watch for entries with known GatewayClass name/namespace details, from the previous watch.
+Contour will watch all Gateway objects, and filter for entries with known GatewayClass name/namespace details, from the previous watch.
 
 Contour is not able to specify the address that the Envoy deployment will listen on, and so having any entries in Gateway's `spec.addresses` stanza will result in Contour marking those addresses as invalid in the Gateway's status.
 Contour expects the `spec.addresses` field to be empty.
 
+For Contour, the key Gateway section are the Listeners. These define how an implementation should listen for traffic described by Routes.
+
 For Listeners:
-- Listeners inside Gateways will be merged where possible, as will listeners across Gateways.
+- Listeners inside Gateways will be merged where possible, as will listeners across Gateways, in order to end up with two broad groups of listeners, secure and insecure.
 - Conflicts within a Gateway will result in the relevant Listeners both being rejected.
 - Conflicts between Gateways will result in the oldest Gateway (by creation timestamp) being kept.
 - The rejected Listener will have its status updated with the reason and the name of the conflicting Gateway.
-- Listeners are considered conflicting if all the fields out of `hostname`, `port`, `protocol`, and `tls` match.
+- Listeners are considered mergable if all the fields out of `hostname`, `port`, `protocol`, and `tls` match.
 - Further merging rules are specified in the detailed design below.
 
 The output of this watcher is both Envoy configuration, and a list of kind/name/namespace details for Routes to watch.
 
-### Contour watches HTTPRoutes
-HTTPRoutes matching a label selector in the Gateway `spec.listeners[].routes` will be watched, and associated routes configured.
-HTTPRoutes referencing Gateways that fit into the Contour GatewayClass will be watched, and associated routes configured.
+### HTTPRoute
+Contour will watch all HTTPRoute objects, and filter for entries matching a label selector in a valid Gateway `spec.listeners[].routes`,
+and will configure the associated routes.
+
+Contour will only ever update the `status` of HTTPRoute objects.
+
 Errors or conflicts here will render that section of the config invalid.
 Other valid sections will still be passed to Envoy.
 For each invalid section, Contour will update status information with the section and the reason.
 
 The output of this watcher is Envoy configuration.
 
-### Contour watches TLSRoutes
-TLSRoutes matching a label selector in the Gateway `spec.listeners[].routes` will be watched, and associated routes configured.
-TLSRoutes referencing Gateways that fit into the Contour GatewayClass will be watched, and associated routes configured.
+### TLSRoute
+Contour will watch all TLSRoute objects, and filter for entries matching a label selector in a valid Gateway `spec.listeners[].routes`,
+and will configure the associated routes.
+
+Contour will only ever update the `status` of TLSRoute objects.
 
 Errors or conflicts here will render that section of the config invalid.
 Other valid sections will still be passed to Envoy.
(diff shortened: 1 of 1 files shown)


design/service-apis-implementation.md
```diff
@@ -0,0 +1,229 @@
...
+- Routes describe how traffic coming via the Gateway maps to the Services.
+
+In Contour, we've previously solved a lot of the same problems with HTTPProxy (and IngressRoute before it).
+That functionality can be described in the Service APIs by HTTPRoutes, HTTPSRoutes and TLSRoutes, as they describe Layer 7 ingress, as Contour does today.
+
+Other types of routes include TCPRoutes and UDPRoutes, which are intended for Layer 4 load balancers. Implementations may also define their own Route objects, at the cost of interoperability.
+
+In terms of its stated goals, Contour is aiming at being an ingress controller - that is, a Layer 7 proxy with some api gateway functions.
+Currently, Contour provides "TCP Proxying" that allows the forwarding of TLS streams based on the SNI information, which is precisely what the Service APIs TLSRoute object is for.
+If Project Contour (the organisation) does add support for TCP and UDP forwarding, it will not be in the `projectcontour/contour` repo, but will be a separate repo.
```
stevesloka, line 33: I'm still confused on the practically of this, you're proposing that we recreate all the same logic to just watch for these two types (e.g. TCP / UDP)?

design/service-apis-implementation.md
```diff
@@ -0,0 +1,229 @@
...
+The output of this watcher is a list of GatewayClass name/namespace details.
+
+#### Gateway
+Contour will watch for entries with known GatewayClass name/namespace details, from the previous watch.
+
+Contour is not able to specify the address that the Envoy deployment will listen on, and so having any entries in Gateway's `spec.addresses` stanza will result in Contour marking those addresses as invalid in the Gateway's status.
+Contour expects the `spec.addresses` field to be empty.
+
+For Listeners:
+- Listeners inside Gateways will be merged where possible, as will listeners across Gateways.
```
stevesloka, line 97: Contour only supports two listeners today (e.g. http/https). Does this statement assume that you are permitted to create more dynamically?

design/service-apis-implementation.md
```diff
@@ -0,0 +1,229 @@
...
+- Conflicts within a Gateway will result in the relevant Listeners both being rejected.
+- Conflicts between Gateways will result in the oldest Gateway (by creation timestamp) being kept.
+- The rejected Listener will have its status updated with the reason and the name of the conflicting Gateway.
+- Listeners are considered conflicting if all the fields out of `hostname`, `port`, `protocol`, and `tls` match.
```
stevesloka, line 101: Could TLS listeners also be in conflict if they reference different certificates?

stevesloka (commented):
What also do we do with things like:
- web sockets
- timeouts
- wildcard domain names (from api spec notes)
- regex headers
- exact path match type (also needed in ingress v1 support)

danehans:
xref: https://github.com/projectcontour/contour-operator/pull/171

design/service-apis-implementation.md
```diff
@@ -0,0 +1,229 @@
...
+
+### Service APIs implementation surface area
+
+Contour's support for the Service APIs will include support for the HTTPRoute and TLSRoute objects only, not any other type of Route - this includes no TCPRoute or UDPRoute support.
+Contour is a layer 7 ingress controller, and the layer 4 load balancing implies by TCPRoute and UDPRoute is out of scope for this tool (that is, `projectcontour/contour`).
+Project Contour (the organisation) may investigate a Service-APIs based layer 4 solution in the future, but that effort will not be in `projectcontour/contour`.
+
+### Contour Gateway model
+
+Contour considers a GatewayClass and associated set of Gateways to describe a single Envoy deployment, deployed and managed by something outside of itself.
```
skriss, line 53: can you remind me of the scenario where a Contour user would have more than one `Gateway`?

design/service-apis-implementation.md
```diff
@@ -0,0 +1,229 @@
...
+
+### Per-object design notes
+
+#### GatewayClass
+
+Contour will look for entries with `spec.controller` equal to `projectcontour.io/contour` by default.
+Contour can change this value to any `projectcontour.io/<value>` with a config setting (similar to how `ingress.class` works today.)
+Importantly, this means that changing this values requires a Contour restart; this means that the GatewayClass watcher can have this value baked in at runtime.
+
+If Contour finds more than one matching GatewayClass, this is a conflict, and Contour will use the oldest GatewayClass (by creation timestamp).
```
skriss, line 85: it looks like the "oldest object wins" strategy is used in a few places - are there any alternatives? why did you choose this strategy?

sunjayBhatia: seems to be some mention of oldest-wins here as well: https://kubernetes-sigs.github.io/service-apis/httproute/#tls (notes section)

design/service-apis-implementation.md
sunjayBhatia: relevant: https://github.com/projectcontour/contour/pull/3213/files#r552799600

design/service-apis-implementation.md
```diff
@@ -0,0 +1,229 @@
...
+The cluster operator and application developer are basically the same as Contour's Cluster Administrator and Application Developer personas, which will be important for this design.
+
+In terms of the APIs themsleves, the Service APIs have 3 primary API resources (taken from the service-apis docs site):
+
+- GatewayClass defines a set of gateways with a common configuration and behavior.
+- Gateway requests a point where traffic can be translated to Services within the cluster.
+- Routes describe how traffic coming via the Gateway maps to the Services.
+
+In Contour, we've previously solved a lot of the same problems with HTTPProxy (and IngressRoute before it).
+That functionality can be described in the Service APIs by HTTPRoutes, HTTPSRoutes and TLSRoutes, as they describe Layer 7 ingress, as Contour does today.
```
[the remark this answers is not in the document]
sunjayBhatia: doesn't look like it: https://kubernetes-sigs.github.io/service-apis/httproute/#spec (TLS optional)

design/service-apis-implementation.md
skriss: just found https://kubernetes-sigs.github.io/service-apis/guidelines/#conflicts as well which explicitly mentions this as a principle

design/service-apis-implementation.md
sunjayBhatia: Looks like we have significant conflicts between the operator proposal and this one: https://github.com/projectcontour/contour-operator/pull/171/files#diff-1cbda5ba0844ea1d442d220ece0cd2f53066104e33808dffb0b7d8df6c2ff5d8R106

sunjayBhatia: cc @danehans

sunjayBhatia: https://github.com/projectcontour/contour-operator/pull/171/files#diff-1cbda5ba0844ea1d442d220ece0cd2f53066104e33808dffb0b7d8df6c2ff5d8R114

design/service-apis-implementation.md
```diff
@@ -0,0 +1,229 @@
...
+
+Each object in the Service APIs set has its own status definition. In general, Contour will update the status of any object that comes into scope with its details, ensuring that the `observedGeneration` field is set correctly.
+When objects fall out of scope, any status set by Contour will not be removed.
+It's expected that things that check the status will also check the `observedGeneration` to check if the status information is up-to-date with the current object generation.
+
+### Per-object design notes
+
+#### GatewayClass
+
+Contour will look for entries with `spec.controller` equal to `projectcontour.io/contour` by default.
```
danehans, line 81: @youngnick if contour operator is responsible for managing `GatewayClass`, then I would expect the value to be `projectcontour.io/contour-operator`. See https://github.com/projectcontour/contour-operator/pull/171 for additional details.

design/service-apis-implementation.md
```diff
@@ -0,0 +1,229 @@
...
+
+If Contour finds more than one matching GatewayClass, this is a conflict, and Contour will use the oldest GatewayClass (by creation timestamp).
+All GatewayClasses will be updated with status conditions indicating if they are in use.
+
+The output of this watcher is a list of GatewayClass name/namespace details.
+
+#### Gateway
+Contour will watch for entries with known GatewayClass name/namespace details, from the previous watch.
+
+Contour is not able to specify the address that the Envoy deployment will listen on, and so having any entries in Gateway's `spec.addresses` stanza will result in Contour marking those addresses as invalid in the Gateway's status.
```
danehans, line 93: The contour-operator will attempt to set these addresses and surface a status condition accordingly. See https://github.com/projectcontour/contour-operator/pull/171/files#r552860708 for add'l background.

design/service-apis-implementation.md
```diff
@@ -0,0 +1,229 @@
...
+
+A detailed design describing how the changes to the product should be made.
+
+The names of types, fields, interfaces, and methods should be agreed on here, not debated in code review.
+The same applies to changes in CRDs, YAML examples, and so on.
+
+Ideally the changes should be made in sequence so that the work required to implement this design can be done incrementally, possibly in parallel.
+
+## Alternatives Considered
+TODO: This will need further explanation of what the solution would look like if we didn't make one GatewayClass == One Contour.
```
danehans, line 218: The contour-operator [design doc](https://github.com/projectcontour/contour-operator/pull/171/files) more closely aligns with this alternative approach. An instance of Contour (i.e. Contour Deployment, Envoy DaemonSet, etc.) equals one Gateway. GatewayClass is simply a resource that points to an implementation-specific config object, i.e. `ContourConfig`. Contour will only be concerned with xRoute resources and using `xroute.spec.gateways[]` to know if it's responsible for handling the route.

design/service-apis-implementation.md
youngnick: I'd guess that as part of this, we'd spin some code out into libraries, probably.

design/service-apis-implementation.md
youngnick: The insecure/secure split definitely needs some thinking as a result of this. The initial  thought is that we would replicate the current behavior, which would mean accepting only one http and one https listener, but this will be tricky to do. We may need to do something different with automatic http->https redirect for the Service APIs.

design/service-apis-implementation.md
```diff
@@ -0,0 +1,229 @@
+# Service-APIs implementation design
+
+Status: Accepted
```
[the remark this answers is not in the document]
youngnick: Well, I figured that if we merge it, it's accepted. Otherwise, this will get closed. Traditionally, we've just not updated this field, and everything is a draft, which is also not ideal. :)

design/service-apis-implementation.md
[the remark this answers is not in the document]
youngnick: 🤦 Yeah, that should just be `HTTPRoutes` and `TLSRoutes`, for terminated and passthrough traffic respectively.

design/service-apis-implementation.md
youngnick: Yes, this is why I picked this, I think that oldest object wins is about the best we can hope for here.

design/service-apis-implementation.md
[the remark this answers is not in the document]
skriss: We could just drop the field, and assume anything that's been merged is accepted; anything still a PR is draft. Not a big deal though.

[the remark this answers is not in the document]
stevesloka: The original idea was to submit ideas here without them being "accepted" yet, to gather thoughts, etc. We've been waiting to commit the design until it's accepted in a way, so removing the field might better follow how we're implementing features.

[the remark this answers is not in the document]
youngnick: Yeah, removing the field may make more sense, if it's merged, it's accepted. If we change things like that, we'll need to:
- update the template and remove the field
- update CONTRIBUTING.md/ other policy docs
- update the old PRs to remove the field.

design/service-apis-implementation.md
youngnick: I've opened #3263 with a new design for this, and will update this document accordingly.

design/service-apis-implementation.md
youngnick: As discussed in our maintainer meeting, Contour will move to being configured to check a *specific*, named by namespace and name Gateway only. This is because of this line. I'll update this to make it clearer.

design/service-apis-implementation.md
youngnick: As discussed in our maintainer meeting, the controller default will be `projectcontour.io/contour`, which will be overridden by the operator to `projectcontour.io/contour-operator`. This is because the Operator is one way to install Contour and Envoy and an associated Gateway, but not the only one.

I'll clarify the properties that Contour expects a Gateway to have.

youngnick:
In the latest round of changes, I've updated a few things:
- change the model that Contour only looks at a single Gateway (since it's part of the implementation of that specific Gateway).
- Contour is configured with the full name (name and namespace) of the Gateway.
- Added references to #3263 and updated the design assuming it's going ahead. If not, this is still doable, just will require extra rules for Listener merging.

design/service-apis-implementation.md
```diff
@@ -0,0 +1,253 @@
...
+Contour does not watch a GatewayClass, as its deployment model expects a Contour to be part of the implementation for a *particular* Gateway.
+
+### Combining Service APIs with other configuration
+
+When it calculates the Envoy configuration using the DAG, Contour layers different types of configuration in order.
+So, currently, Ingress is overwritten by HTTPProxy, so if an Ingress and a HTTPProxy specify the exact same route, the HTTPProxy will win.
+
+Once we have the Service-APIs available as well, we have to choose the order.
+In this design, I suggest having the order be Ingress is overwritten by HTTPProxy, is overwritten by the service-apis.
+I could see reversing HTTPProxy and the Service APIs here, but I think this acknowledges that the Service APIs are really an evolution of the ideas in HTTPProxy, and will probably end up being the community standard.
```
skriss, line 71: I'm OK with either order, as long as we document it.

skriss, line 71: Is it a goal to eventually sunset `HTTPProxy` and have service APIs be the only API for Contour? If not yet a goal, do we think it might be some day? Just curious how others are thinking about this, and whether we expect to support multiple APIs indefinitely or if we hope to converge on one.

design/service-apis-implementation.md
```diff
@@ -0,0 +1,253 @@
...
+- Listeners are considered mergable if all the fields out of `hostname`, `port`, and `protocol` match, with some additional rules around TLS.
+- Further merging rules are specified in the detailed design below.
+- Listeners that refer to any other Route than HTTPRoute or TLSRoute will be ignored, and a condition placed on the corresponding `status.listeners[]` object saying that it was ignored because those objects are not supported.
+
+The output of this watcher is both Envoy configuration, and a list of kind/name/namespace details for Routes to watch.
+
+Contour will only update the `status` of Gateway objects.
+
+### HTTPRoute
+Contour will watch all HTTPRoute objects, and filter for entries matching a label selector in a valid Gateway `spec.listeners[].routes`,
```
skriss, line 109:
```suggestion
Contour will watch all HTTPRoute objects, and filter for entries matching a label selector in its configured Gateway's `spec.listeners[].routes`,
```

design/service-apis-implementation.md
```diff
@@ -0,0 +1,253 @@
...
+Contour will only ever update the `status` of HTTPRoute objects.
+
+Errors or conflicts here will render that section of the config invalid.
+Other valid sections will still be passed to Envoy.
+For each invalid section, Contour will update status information with the section and the reason.
+
+The output of this watcher is Envoy configuration.
+
+### TLSRoute
+Contour will watch all TLSRoute objects, and filter for entries matching a label selector in a valid Gateway `spec.listeners[].routes`,
```
skriss, line 124:
```suggestion
Contour will watch all TLSRoute objects, and filter for entries matching a label selector in its configured Gateway's`spec.listeners[].routes`,
```

design/service-apis-implementation.md
```diff
@@ -0,0 +1,253 @@
...
+In Contour, we've previously solved a lot of the same problems with HTTPProxy (and IngressRoute before it).
+That functionality can be described in the Service APIs by HTTPRoutes and TLSRoutes, as they describe Layer 7 ingress, as Contour does today.
+
+Other types of routes include TCPRoutes and UDPRoutes, which are intended for Layer 4 load balancers. Implementations may also define their own Route objects, at the cost of interoperability.
+
+In terms of its stated goals, Contour is aiming at being an ingress controller - that is, a Layer 7 proxy with some api gateway functions.
+Currently, Contour provides "TCP Proxying" that allows the forwarding of TLS streams based on the SNI information, which is precisely what the Service APIs TLSRoute object is for.
+If Project Contour (the organisation) does add support for TCP and UDP forwarding, it will not be in the `projectcontour/contour` repo, but will be a separate repo.
+
+## Goals
```
skriss, line 35: is it a goal for Contour's service-apis support to reach ~feature parity with `HTTPProxy`? If so, how will we implement features that aren't explicitly part of the service-apis spec? And if not, what set of features _do_ we expect to implement?

Another way of asking this is, are we going to use annotations to implement features that aren't directly a part of service-apis, like we do with Ingress?

design/service-apis-implementation.md
youngnick: I'd like to think that, if the Service APIs had full feature parity with HTTPProxy, and our users were willing to move, we would remove the CRD.

I don't think either of those is likely though, so what I'm trying to do is behave as though that might be the case someday. Practically, I don't think it will be, but I'm not convinced of the benefit of having our own CRD if the above two conditions are fulfilled.

design/service-apis-implementation.md
youngnick: I'll put something in the background to summarise this, but:
- I think the goal should absolutely be that eventually service-apis has feature parity with HTTPProxy
- the upstream does not have support for many features right now
- we should implement what is implementable right now
- we should collaborate with upstream to work out what features are part of the core, extension, and implementation specific sections of service-apis.

The service-apis are specifically built with extension points so you don't have to use annotations. That's what the reference points that allow you to point to any object, but default to a specific kind, are for.

youngnick:
To answer @stevesloka's question about additional features, I've also put some more in the documentation about what parts of the API we will target, and how we will build support.

In short, we'll build whatever is in v1alpha1 in the initial cut, then work with the upstream to define the other features (timeouts, websockets, regex headers etc) in the upstream, and support that.

design/service-apis-implementation.md
youngnick: I think I've covered this in a recent update, please tell me if it's clearer now.

design/service-apis-implementation.md
youngnick: I think we cleared this up in a maintainer's meeting, now that the Contour installation (= Contour deployment and Envoy DaemonSet and associated bits) is 1-to-1 with a named Gateway.

design/service-apis-implementation.md
```diff
@@ -0,0 +1,258 @@
...
+
+In terms of its stated goals, Contour is aiming at being an ingress controller - that is, a Layer 7 proxy with some api gateway functions.
+Currently, Contour provides "TCP Proxying" that allows the forwarding of TLS streams based on the SNI information, which is precisely what the Service APIs TLSRoute object is for.
+If Project Contour (the organisation) does add support for TCP and UDP forwarding, it will not be in the `projectcontour/contour` repo, but will be a separate repo.
+
+This design is intended to cover the initial `v1alpha1` release of the Service APIs. We will aim to implement the core featureset of the APIs at that point.
+We will then work with the upstream community on features that Contour and HTTPProxy currently support, but the Service APIs do not, and how best to represent those features in the APIs.
+So, there are some features that are currently gaps for Contour (wildcard domain names, and exact path matching seem like big ones),
+and some that Contour supports that the Service APIs do not (webscokets, configurable timeouts, header replacement, external auth, rate limiting, and so on).
```
stevesloka, line 38: nit: s/webscokets/websockets/

design/service-apis-implementation.md
```diff
@@ -0,0 +1,258 @@
...
+Configuration of TLSRoutes  will be subject to the rules around the `RouteGateways` field.
+
+Contour will only ever update the `status` of TLSRoute objects.
+
+Errors or conflicts here will render that section of the config invalid.
+Other valid sections will still be passed to Envoy.
+
+The output of this watcher is Envoy configuration.
+
+## Detailed Design
```
skriss, line 140: do we want to have one DAG processor for all service-apis objects? One per object type (e.g. `GatewayProcessor`, `HTTPRouteProcessor`, etc)? Something else?

skriss (commented):
I don't have any other significant feedback here, I think this points us in the right direction. I'd be OK working out the remaining details of processing objects in code, especially given the alpha status of the API.

stevesloka:
I don't have any additional large comments. There are a lot of behavior details in the API spec which we'll also need to reference during implementation. 

There are still quite a few TODO's throughout the doc that will need another review pass from folks if they are expaned upon. 

Thanks for this work @youngnick!

design/service-apis-implementation.md
```diff
@@ -0,0 +1,258 @@
...
+### Service APIs implementation surface area
+
+Contour's support for the Service APIs will include support for the HTTPRoute and TLSRoute objects only, not any other type of Route - this includes no TCPRoute or UDPRoute support.
+Contour is a layer 7 ingress controller, and the layer 4 load balancing implies by TCPRoute and UDPRoute is out of scope for this tool (that is, `projectcontour/contour`).
+Project Contour (the organisation) may investigate a Service-APIs based layer 4 solution in the future, but that effort will not be in `projectcontour/contour`.
+
+### Contour Gateway model
+
+Contour considers a Gateway to describe a single Envoy deployment, deployed and managed by something outside of itself.
+Contour expects to be supplied with the full name (name and namespace) of a single Gateway, which it will watch and update the status of.
```
stevesloka, line 59: Do you see this as new flags to Contour or placed into the config file?

youngnick: I think config file for this one, I'll update.

design/service-apis-implementation.md
youngnick: I'm not sure. I think that splitting the processors up would be a bit different to how we've done things before, but that's not necessarily bad. It's probably worth the implementor giving this a go.

youngnick:
I've updated this document a bit today, but I still have a few TODOs. I've removed the WIP as this is much closer now, but left it as draft as it will still need some more review passes.

youngnick:
Okay, I've marked this ready for review, I've finished up all the outstanding TODOs, I think. Can I get some more eyes across this, then we can get it merged.

design/service-apis-implementation.md
```diff
@@ -0,0 +1,326 @@
...
+
+### Gateway
+
+When ingesting Gateways, Contour will import the configured named Gateway into its cache, and watch it for spec changes. Spec changes will trigger a DAG run.
+
+#### Listener Ports
+
+Contour allows the configuration of arbitrary ports, **as long as those ports are exposed via the Envoy service**.
+The port that is configured for the listener is the port as it appears at the Service.
+That is, the Envoy Service `port` number is what the Listener must specify, regardless of if there is a `targetPort` set on the Service.
```
sunjayBhatia, line 213: Maybe a comment somewhere making it explicit that we will start watching the Envoy service would be good to add. Also it forces users (cluster operators?) to deploy their Envoys in a certain way. Practically speaking its unlikely someone had their Envoys deployed much differently than we have in the examples or the operator, but it is technically a new constraint I guess? Maybe not that serious

design/service-apis-implementation.md
```diff
@@ -0,0 +1,326 @@
...
+As part of that, Contour is requested to look at one and only one Gateway, which contradicts the "watches multiple Gateways" model.
+
+## Implementation
+
+There is currently experimental support for *reading* the Service-APIs objects in Contour, we will add full support in something like this order:
+- Add support for configuring Gateway to the config file
+- Deprecate `--expermental-service-apis` command-line flag, so that it does nothing but log a referral to the new config item.
+- Implement the Service-APIs processor as a single DAG processor.
+- Release a version with Service-APIs support still marked experimental.
+- Remove the experimental tag once we're happy with the support.
```
sunjayBhatia, line 326: Might be worth adding some notes about how before this stage we will try to evaluate our implementation and work with upstream to work out issues, conformance testing, etc.

sunjayBhatia (commented):
Overall looks reasonable to me, theres lots of details to get right but with incremental PRs and a good focus in PR review on test coverage we should be good to refer to this design for help

design/service-apis-implementation.md
youngnick: I've updated this quite a bit to reflect the fact that we're not going to allow arbitrary ports for now. There's definitely a problem that will need to be solved with some experimentation during implementation here.

design/service-apis-implementation.md
```diff
@@ -0,0 +1,342 @@
...
+This allows the Gateway to configure a default TLS certificate.
+Note that this feature has fiddly interactions with the various places in which a hostname may be specified;
+this will require careful test case design.
+
+The output of this watcher is both Envoy configuration, and a list of kind/name/namespace details for Routes to watch.
+
+Contour will only update the `status` of Gateway objects.
+
+### HTTPRoute
+Contour will watch all HTTPRoute objects, and filter for entries matching a label selector in its configured Gateway's `spec.listeners[].routes`,
```
tsaarni, line 129: Did I understand it correctly that `HTTPRoute` and `TLSRoute` resources are first selected from Gateway's own namespace or from namespaces listed in Gateway's `spec.listeners[].routes.namespaces` (if given), and after that optional label selector `spec.listeners[].routes.selector` is applied (if given).  With the current wording this two level process might be bit unclear. 
 
Since the selection logic is more `Gateway` topic, it could be moved from HTTPRoute and TLSRoute chapters to Gateway chapter, but it is readable as it is today as well.

By the way, the new "inclusion" logic seems to solve #2206.

design/service-apis-implementation.md
```diff
@@ -0,0 +1,342 @@
...
+Contour's support for the Service APIs will include support for the HTTPRoute and TLSRoute objects only, not any other type of Route - this includes no TCPRoute or UDPRoute support.
+Contour is a layer 7 ingress controller, and the layer 4 load balancing implies by TCPRoute and UDPRoute is out of scope for this tool (that is, `projectcontour/contour`).
+Project Contour (the organisation) may investigate a Service-APIs based layer 4 solution in the future, but that effort will not be in `projectcontour/contour`.
+
+### Contour Gateway model
+
+Contour considers a Gateway to describe a single Envoy deployment, deployed and managed by something outside of itself.
+Contour expects to be supplied with the full name (name and namespace) of a single Gateway, which it will watch and update the status of.
+This will be a configuration file parameter.
+When a Gateway is not supplied, Contour will not do Service-APIs processing, or in other words, the existing `--experimental-service-apis` flag will be obsoleted and removed).
```
stevesloka, line 67: +1 on removing the flag. 

Should we default the name/namespace of a gateway (i.e. contour/projectcontour) so given a generic deployment, someone could use that as a default and not require a config file change?

Contour can still check if the service-apis CRDs exist in the cluster and use that as a key for setting up informers.  I just merged https://github.com/projectcontour/contour/pull/3283 which acts this way, but is different than this design doc, but @skriss asked which is a good point.

design/service-apis-implementation.md
```diff
@@ -0,0 +1,342 @@
...
+Inside Contour will merge Listeners within its Gateway. (See the detailed design for the exact rules Contour will use for this.)
+
+### Interrelated watches
+
+The Service APIs are a set of interrelated Kubernetes objects, where a change in one object can mean that the scope of objects Contour is interested in will change.
+Because of this, Contour will watch all events associated with its named Gateway, HTTPRoute, TLSRoute, and BackendPolicy objects, and filter the objects it takes action on internally, rather than using a filtered watch (as those are costly to set up and tear down).
+
+Contour does not watch a GatewayClass, as its deployment model expects a Contour to be part of the implementation for a *particular* Gateway.
```
stevesloka, line 75: If we did watch the class, we could allow a single gateway in a cluster to be configured by a single contour without any configuration file changes. But when a second Gateway for the same class controller was created you'd need to specify.

Would just be nice to allow users to configure Gateways without having to edit the configmap + restart contour instances.

design/service-apis-implementation.md
```diff
@@ -0,0 +1,342 @@
...
+When a HTTPRoute specifies a hostname or slice of hostnames, those hostnames must match the hostnames in the Gateway.
+Note that more specific precise matches at the Hostname level may match less specific wildcard matches at the Gateway level.
+This allows the Gateway to define a default TLS certificate, which may only be overridden
+
+The HTTPRoute also has a facility to supply additional TLS Config using the `tls` stanza (the RouteTLSConfig field).
+The most important part here is that the field is only used if the `AllowRouteOverride` field is set in the referencing Gateway resource.
+
+Contour will only ever update the `status` of HTTPRoute objects.
+
+Errors or conflicts here will render that section of the config invalid.
```
stevesloka, line 142: For a missing `serviceName` or `LocalObjectReference`, the docs say: 
```
If the referent cannot be found, the route must be dropped from the Gateway. The controller should raise the “ResolvedRefs” condition on the Gateway with the “DegradedRoutes” reason. The gateway status for this route should be updated with a condition that describes the error more specifically.
```

Does that mean the entire `HTTPRoute` is invalid or just the specific `HTTPRoute.HTTPRouteRule` is invalid?

sunjayBhatia: Yeah I wasn't sure about that either (and how it would play with the proposal you had around keeping as much routability as possible even in partially applied resources), I think we chatted about this in office hours or something but I'm not sure there was a conclusion

design/service-apis-implementation.md
skriss: :+1:

design/service-apis-implementation.md
```diff
@@ -0,0 +1,342 @@
...
+Contour does not watch a GatewayClass, as its deployment model expects a Contour to be part of the implementation for a *particular* Gateway.
+
+### Combining Service APIs with other configuration
+
+When it calculates the Envoy configuration using the DAG, Contour layers different types of configuration in order.
+So, currently, Ingress is overwritten by HTTPProxy, so if an Ingress and a HTTPProxy specify the exact same route, the HTTPProxy will win.
+
+Once we have the Service-APIs available as well, we have to choose the order.
+In this design, I suggest having the order be Ingress is overwritten by HTTPProxy, is overwritten by the service-apis.
+I could see reversing HTTPProxy and the Service APIs here, but I think this acknowledges that the Service APIs are really an evolution of the ideas in HTTPProxy, and will probably end up being the community standard.
```
skriss, line 84: Some aspects of this are not entirely clear to me, e.g. what if there's a service-apis Listener with a wildcard domain name that overlaps with an HTTPProxy with a specific domain name, or what if some other service-apis Listener settings conflict with Contour's listener settings -- but I'm also not sure it's that important right now, and we can figure it out during implementation.

skriss (approved):
no further comments/questions from me, I think we can sort out remaining details during implementation. 🎉

design/service-apis-implementation.md
youngnick: So, the flow if you want to use the service-apis is this:
- install the objects in the cluster
- create a Gateway for Contour to reference
- Configure Contour to use that

If you don't do all those three things, then no service-apis.

That does make sense, although it means that if you install the objects and don't change the default, we should end up logging something saying "hey, you asked me to look for a Gateway called `projectcontour/contour`, but it doesn't exist". Not a huge deal.

design/service-apis-implementation.md
youngnick: I think that it's better to be explicit for now. We can revisit this later if the configuration proves too complicated. This way we can avoid a more complicated state like we had for the ingress.class annotation, where the behavior if you don't specify a value is not straightforward.

design/service-apis-implementation.md
youngnick: You're right, I think we need to implement this per the spec for now, and raise this upstream as a concern.

design/service-apis-implementation.md
youngnick: I think the wildcard domain name thing will be sorted out by implementing wildcard domains everywhere - we should do that before starting with this, since we need to do it for Ingress v1 anyway. I think the limited form of wildcard FQDN defined in the service-apis is probably implementable in HTTPProxy as well.

I think the LIstener settings details should be sorted out during implementation, but I think that the service-apis spec should probably override the Contour config one, since it's more specific (I think). I don't think it matters really, as long as we document it.

design/service-apis-implementation.md
youngnick: Yes, you're correct @tsaarni. I'll try and update this to be closer to the spec.

Also, yes, I know that #2206 is partially solved by HTTPRoutes, although they currently cannot be included in themselves (so you can only do the equivalent of one layer of HTTPProxy includes). There was a vision of doing that initially for the service-apis, but I'm not sure where it's at, and it would add significant complexity to processing of the HTTPRoute objects.

design/service-apis-implementation.md
stevesloka: Right now I have a default programmed, but we could remove that, just thought it might be nice to have some way of allowing someone to create a Gateway without needing to reconfigure / restart Contour. 

This came up in a review with @skriss and thought it was a good point that what I PR'd conflicted with this design and thought we should clear up. =)

design/service-apis-implementation.md
stevesloka: Chatting in Slack the thought is that only that HTTPRoute.Rule is invalidated so I think the design is correct unless we want to specifically call it out better: https://kubernetes.slack.com/archives/CR0H13KGA/p1612463856030500?thread_ts=1612459180.030400&cid=CR0H13KGA

stevesloka (approved):
I think this is good enough to start implementation and dig into what comes out of that when we hit those bits. 🎉

design/service-apis-implementation.md
youngnick: Thanks for asking this question in the upstream Slack @stevesloka. I agree with Harry's answer there - we will implement this as "conflicts will invalidate that *rule*" not "conflicts with invalidate the whole Route" for now, and see if we can push that upstream.

youngnick:
Looks like the new ExternalName integration test is failing on `main`, I'm going to merge this design, and take a look.

PR #3216, opened 2020-12-17
[release-1.10] Update Contour Docker image to v1.10.1.

skriss:

### DIFF
--- a/examples/contour/02-job-certgen.yaml
+++ b/examples/contour/02-job-certgen.yaml
@@ -36,7 +36,7 @@ rules:
 apiVersion: batch/v1
 kind: Job
 metadata:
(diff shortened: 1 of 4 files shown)


sunjayBhatia:
@skriss I guess this has to be merged before manually making a tag and running the image push so the v1.10.1 image exists?

skriss:
Yeah, I think so...I think this change may usually be pushed directly to the branch, rather than PR'ed.

skriss:
I guess the integration tests won't pass because of the tags, so we'll just have to ignore that check on this PR.

stevesloka (approved):
/lgtm

sunjayBhatia (approved):

PR #3224, opened 2020-12-17
site: rewrite release instructions

skriss:

This is a rough draft of revised documentation for the release process. The intent is to make it easier to execute manually and to serve as a better blueprint for automation. I specifically did a few things:

- ensure every step is included, and have bash snippets where appropriate
- separated out minor vs. patch releases so the reader doesn't need to decide if each step is needed or not.
- separate out a general overview of what needs to be done, from the actionable steps to make it happen

Please give me a 👍  or 👎  as to whether this format is an improvement over the current https://projectcontour.io/resources/release-process/, as well as any specific feedback. If folks like it, I'll bring it to completion; if not, we can try something else.

Hopefully it goes without saying that ultimately we should automate as much as we can, but improved documentation is a good and easy first step.

### DIFF
--- a/site/_resources/release-process.md
+++ b/site/_resources/release-process.md
@@ -3,136 +3,268 @@ title: Contour Release Process
 layout: page
 ---
 
-This page documents the process for releasing a new version of Contour or [Contour Operator][1].
+- [Minor release process](#minor-release-process)
+- [Patch release process](#patch-release-process)
 
-The release types are as follows. All are tagged from the same release branch.
+# Minor Release Process
 
-- Alpha releases.
-- Beta releases.
-- RC (Release Candidate) releases.
-- Final releases.
-- Patch releases.
+## Overview
 
-For minor releases, generally we will not do an Alpha/Beta/RC/Final flow, we will jump straight to cutting the final from main.
+A minor release is e.g. `v1.11.0`.
 
-## Alpha and beta releases
+A minor release requires:
+- website updates
+- a release branch to be created
+- YAML to be customized
+- a release tag to be created
+- an Operator release
+- a GitHub release with release notes
+- public communication
+- cleanup
 
-The steps for an alpha or beta release are
 
-- Tag the head of main with the relevant release tag (in this case `alpha.1`), and push
+## Process
 
-```sh
-$ git tag -a v0.15.0-alpha.1 -m 'contour 1.2.0 alpha 1'
-$ git push --tags
+### Set environment variables
+
+Set environment variables for use in subsequent steps:
+
+```bash
+export CONTOUR_RELEASE_VERSION=v1.11.0
+export CONTOUR_RELEASE_VERSION_MAJOR=1
+export CONTOUR_RELEASE_VERSION_MINOR=11
+
+export CONTOUR_UPSTREAM_REMOTE_NAME=upstream
+export CONTOUR_OPERATOR_UPSTREAM_REMOTE_NAME=upstream
 ```
 
-Once the tag is present on main, Github Actions will build the tag and push it to Docker Hub for you.
-Then, you are done until we are ready to branch for an rc or final release.
+### Update the website with release-specific information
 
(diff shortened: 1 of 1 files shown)


site/_resources/release-process-new.md
```diff
@@ -0,0 +1,252 @@
...
+1. Push the tag to github.com/projectcontour/contour:
+
+    ```bash
+    # if your remote corresponding to github.com/projectcontour/contour is not "upstream", modify accordingly.
+    git push upstream ${VERSION}
+    ```
+
+### Release the operator
+
+TODO
```
skriss, line 232: Need to think about (a) Contour patch releases; and (b) Operator patch releases

sunjayBhatia:
👍🏽 from me, nice to have v explicit instructions, should make it an easier task to automate as well

stevesloka (commented):
👍  from me. I think we can even combine the operator & contour release instructions together since they are releasing at the same time, but not a big deal. Nice work!

stevesloka:
Something else that hit my mind that we might look at adding to this as another section is a "ready for release" checklist. Things like a blog post, all PRs ready in corresponding repos, release notes, etc.

site/_resources/release-process.md
```diff
@@ -3,136 +3,269 @@ title: Contour Release Process
...
+    
+    ```bash
+    go run ./hack/release/prepare-release.go $CONTOUR_PREVIOUS_VERSION $CONTOUR_RELEASE_VERSION
+    ```
+
+1. Add the new release to the compatibility matrix (`/site/_resources/compatibility-matrix.md`).
+1. Document upgrade instructions for the new release (`/site/_resources/upgrading.md`).
+1. Commit all changes, push the branch, and PR it into `main`.
+
+TODO do we want to cherrypick docs into the release branch??
```
skriss, line 208: open question - do we care to to cherrypick the new versioned docs into the release branch? The current instructions don't have us do this. I don't think it really matters since the site is always generated from `main`.

skriss:
this should be complete & ready for review now.

site/_resources/release-process.md
```diff
@@ -3,136 +3,269 @@ title: Contour Release Process
...
+1. Update `netlify.toml` to redirect quickstart links to the new release branch.
+1. Commit all changes, push the branch, and PR it into `main`.
+
+### Do the Github release and write release notes
+
+Now you have a tag pushed to Github, go to the release tab on github, select the tag and write up your release notes.
+
+_Note: Filter on the Github label "release note" and Github milestone which should include any PRs which should be called out in the release notes._ 
+
+### Toot your horn
```
sunjayBhatia, line 136: Should we mention the blog post for the release here or is that action item captured elsewhere?

site/_resources/release-process.md
sunjayBhatia: Given we would likely never change the site/take a site PR for a previous release branch its probably ok to leave out IMO

site/_resources/release-process.md
skriss: Yep good catch, I'll add that.

site/_resources/release-process.md
stevesloka: Could we reverse the steps and do the docs first, then create the release branch?

stevesloka (approved):
/lgtm

site/_resources/release-process.md
skriss: This is for a patch release so we already have a release branch. We could definitely do docs first, in `main`, and then cherrypick them into the release branch, if we care to have them in there.

stevesloka: ohh I see, I don't think it's a big deal just for docs.

skriss: cool, SGTM. we can always decide to do it down the road if we want for some reason.

sunjayBhatia (approved):

PR #3231, opened 2020-12-21
Do a case insensitive match when checking for duplicate FQDNs

### LINKED ISSUE #3230: RouteConfiguration rejected due to duplicate entry
Creating two HTTPProxy objects with FQDNs that only differ by case is allowed and causes Envoy to reject all further route configuration.

Duplicate FQDNs should be rejected, regardless of case.

Example - creating these two HTTPProxy objects:
```
$ cat explorer-httpproxy-routes.yaml
kind: HTTPProxy
metadata:
  name: explorer
spec:
  virtualhost:
    fqdn: explorer.mydomain.com
    tls:
      secretName: explorer-tls
  routes:
    - services:
      - name: explorer
        port: 8080
 
$ cat explorer-httpproxy-routes-up.yaml
kind: HTTPProxy
metadata:
  name: explorer-up
spec:
  virtualhost:
    fqdn: EXPLORER.mydomain.com
    tls:
      secretName: explorer-tls
  routes:
    - services:
      - name: explorer
        port: 8080
 ```
Results in Envoy rejecting all further route configuration with an error like:
```
envoy-vdpsb envoy-485m7envoy-wpvz4envoy-t82l7envoy-v4krhenvoy  envoy    [2020-12-21 13:12:00.024][1][warning]
[config] [source /common/config/grpc_subscription_impl.cc:107] gRPC config for 
type.googleapis.com/envoy.config.route.v3.RouteConfiguration rejected: Only unique values for domains are permitted.
Duplicate entry of domain explorer.mydomain.com in route ingress_http
```

**Environment:**

- Contour version: 1.10.1
- Kubernetes version: (use `kubectl version`): 1.18.11
- Kubernetes installer & version: kubeadm 1.18
- Cloud provider or hardware configuration: on-prem
- OS (e.g. from `/etc/os-release`): RHEL 8.2

erwbgy:
Do a case insensitive match when checking for duplicate FQDNs to avoid RouteConfiguration failures.

Fixes #3230


youngnick (approved):
LGTM, thanks for this!

youngnick:
We'll need one more review on this one, we recently changed our review settings to require 2 reviews. But this will definitely be in 1.12 (due January 2021).

sunjayBhatia (approved):
Looks got to me, only additional change I could see would be some detail in an error message but given people should know virtual host names should be case insensitive it also seems unnecessary

jpeach:
> We'll need one more review on this one, we recently changed our review settings to require 2 reviews. But this will definitely be in 1.12 (due January 2021).

Maybe just verify whether we should lowercase the name before sending it to envoy?

sunjayBhatia:
General question, are there any higher level tests we should add for this? Worth adding an integration test?

youngnick:
Envoy only performs lower-case host name matches itself anyway, so we are not gaining anything by case insensitivity. See envoyproxy/envoy#2412 for the fix there.

youngnick:
Seems like there is also currently case sensitivity for SNI hostnames, although it will be changed at some point, see envoyproxy/envoy#6199 for more.

This definitely will require a release note and some explanation, since in the event that people are relying on the non-RFC-compliant case sensitivity we currently have, they need to know.

However, given we have a reasonably important bug here (#3230), we need to have a solution, which this is. I don't feel that we should be tying ourselves in knots to maintain compatibility when that compatibility is against the relevant RFCs anyway. I won't merge this until the new year, to give people a chance to see this PR and comment if it will affect them.

erwbgy:
One thing to note in the release notes is that configuring an FQDN like EXPLORER.domain.com still works after this fix, so won't break any users with uppercase in their URL hostnames. They just won't be able to configure multiple FQDNs that differ only by case.

The multiple case-insensitive FQDNs only works with tcpproxy today and fails when routes are used, so that is the only use-case affected by this change. On the other hand it is trivially easy to accidentally (or on purpose) break a multi-tenant Contour setup, though unlikely (one of our users did it twice).

youngnick:
I munged the commit message a bit, @erwbgy.

PR #3235, opened 2020-12-31
internal/envoy/v3/bootstrap: Make configurable about keepalive settings.

### LINKED ISSUE #2652: Add configuration support to tune or disable listener Keepalives.

Followup from #2638 #2633 

If the Envoy proxy that Contour is programming isn't running on Linux, it will NACK the listener update, and Contour won't know how to recover from that. In this case, an operator would need a way to disable setting Linux-specific socket options.

There may be use cases for tweaking the Listener keep alive tunables. We can collect implementation experience here.

### LINKED ISSUE #3214: Get upstream_connection_options configurable to do workaround for the flaky behavior of envoy
I deployed a Golang HTTP client that makes HTTP requests periodically, about once every 10 seconds,  to L7 endpoints which are exposed with envoy. This is done to check if the endpoints are alive as end-to-end testing.

While running it, I experienced the same error as reported in the [issue](https://github.com/envoyproxy/envoy/issues/8639) about envoy.
[This](https://github.com/envoyproxy/envoy/issues/8639#issuecomment-551245943) comment says that the error has something to do with keepalive, and [this](https://github.com/envoyproxy/envoy/issues/8639#issuecomment-694303504) comment says that this issue is actually solved by configuring the `tcp_keepalive` under `upstream_connection_options`.

I think it's better to get the `upstream_connection_options` fields configurable because some developers seem to have faced this issue.
Any thoughts on this?

binoue:
This PR is an Implementation for allowing users to tune TCP keep alive settings.
I think this feature is already argued in #2652 and would help for #3214.

Fixes #2652
Fixes #3214

Signed-off-by: binoue <banji-inoue@cybozu.co.jp>

### DIFF
--- a/cmd/contour/bootstrap.go
+++ b/cmd/contour/bootstrap.go
@@ -35,5 +35,8 @@ func registerBootstrap(app *kingpin.Application) (*kingpin.CmdClause, *envoy.Boo
 	bootstrap.Flag("envoy-key-file", "Client key filename for Envoy secure xDS gRPC communication.").Envar("ENVOY_KEY_FILE").StringVar(&config.GrpcClientKey)
 	bootstrap.Flag("namespace", "The namespace the Envoy container will run in.").Envar("CONTOUR_NAMESPACE").Default("projectcontour").StringVar(&config.Namespace)
 	bootstrap.Flag("xds-resource-version", "The versions of the xDS resources to request from Contour.").Default("v3").StringVar((*string)(&config.XDSResourceVersion))
+	bootstrap.Flag("tcp-keepalive-probe", "Maximum number of keepalive probes to send without response.").Uint32Var(&config.KeepaliveProbe)
+	bootstrap.Flag("tcp-keepalive-time", "The number of seconds a connection needs to be idle before keep-alive probes start being sent.").Uint32Var(&config.KeepaliveTime)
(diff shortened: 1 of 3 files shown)


binoue:
@youngnick 

Hi! As #2652 is not progressing, I tried to make TCP keep-alive settings of Envoy configurable.
Would you look at this PR because I am not sure this implementation is acceptable for Contour?

sunjayBhatia:
This change is currently only changing the keepalive settings between Envoy and Contour (the change to `UpstreamConnectionOptions` here is only for the static `contour` xDS cluster) so I don't think it solves the linked issues.

binoue:
@sunjayBhatia 

> This change is currently only changing the keepalive settings between Envoy and Contour (the change to `UpstreamConnectionOptions` here is only for the static `contour` xDS cluster) so I don't think it solves the linked issues.

Thank you for your comment!
As I was misunderstanding, I will fix this PR

stevesloka:
Hey @binoue just wanted to touch base again on this PR, were looks like there were a couple more things to clean up first. Let us know when you're ready for another review! =)

binoue:
@stevesloka 

Thank you for your comment! 
However, I could not take enough time to implement this feature for a while, so let me close this PR

PR #3236, opened 2021-01-05
Support specific Envoy dynamic request headers

### LINKED ISSUE #3234: Support specific dynamic request headers
From https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#custom-request-response-headers:

> Envoy supports adding dynamic values to request and response headers. The percent symbol (%) is used to delimit variable names...
> Supported variable names are:
> %DOWNSTREAM_REMOTE_ADDRESS%
> Remote address of the downstream connection. If the address is an IP address it includes both address and port.
> ...

I would like to use these dynamic values to set request headers - for example:
```
spec:
  routes:
  - services:
    - name: myservice
      port: 80
      requestHeadersPolicy:
        set:
        - name: X-Envoy-Hostname
          value: '%HOSTNAME%'
```

but currently the % is escaped so the result is a literal %HOSTNAME% rather than the expected dynamic value.

There have also been other requests like #2516 to set header values based on other header values.

There are legitimate concerns about sending invalid data to Envoy (see #1176 and #2569) as invalid custom headers cause route configuration failures - for example testing using a patched contour that allows %UNKNOWN% through:
```
      requestHeadersPolicy:
        set:
        - name: X-Envoy-unknown
          value: '%UNKNOWN%'  
```
results in:
```
envoy-wpvz4 envoy [2020-12-27 14:37:57.945][1][warning][config] [source/common/config/grpc_subscription_impl.cc:107] 
gRPC config for type.googleapis.com/envoy.config.route.v3.RouteConfiguration rejected: field 'UNKNOWN' not supported
 as custom header
```

The proposal is to only allow a specific set of known good Envoy dynamic headers to pass through unescaped and continue escaping the % for anything else:
* %DOWNSTREAM_REMOTE_ADDRESS%
* %DOWNSTREAM_REMOTE_ADDRESS_WITHOUT_PORT%
* %DOWNSTREAM_LOCAL_ADDRESS%
* %DOWNSTREAM_LOCAL_ADDRESS_WITHOUT_PORT%
* %DOWNSTREAM_LOCAL_PORT%
* %DOWNSTREAM_LOCAL_URI_SAN%
* %DOWNSTREAM_PEER_URI_SAN%
* %DOWNSTREAM_LOCAL_SUBJECT%
* %DOWNSTREAM_PEER_SUBJECT%
* %DOWNSTREAM_PEER_ISSUER%
* %DOWNSTREAM_TLS_SESSION_ID%
* %DOWNSTREAM_TLS_CIPHER%
* %DOWNSTREAM_TLS_VERSION%
* %DOWNSTREAM_PEER_FINGERPRINT_256%
* %DOWNSTREAM_PEER_FINGERPRINT_1%
* %DOWNSTREAM_PEER_SERIAL%
* %DOWNSTREAM_PEER_CERT%
* %DOWNSTREAM_PEER_CERT_V_START%
* %DOWNSTREAM_PEER_CERT_V_END%
* %HOSTNAME%
* %PROTOCOL%
* %REQ(header-name)%
* %START_TIME%
* %UPSTREAM_REMOTE_ADDRESS%
* %RESPONSE_FLAGS%
* %RESPONSE_CODE_DETAILS%

Notes:
* None of these variables cause route failures when passed through unescaped
* Envoy passes variables that can't be expanded through unchanged or skips them entirely - for example:
  * %UPSTREAM_REMOTE_ADDRESS% as a request header remains as %UPSTREAM_REMOTE_ADDRESS% because as noted in the Envoy docs: "The upstream remote address cannot be added to request headers as the upstream host has not been selected when custom request headers are generated."
  * %DOWNSTREAM_TLS_VERSION% is skipped if TLS is not in use
* Envoy ignores REQ headers that refer to an non-existent header - for example %REQ(Host)% works as expected but %REQ(Missing-Header)% is skipped
* The %DYNAMIC_METADATA([“namespace”, “key”, …])% and %PER_REQUEST_STATE(reverse.dns.data.name)% headers were explicitly excluded as this adds significant complexity and it is not clear these can be used safely

I think this provides useful functionality while still ensuring that bad config is not sent to Envoy.

Would you be open to a pull request that implements this?

erwbgy:
internal/dag: support specific Envoy dynamic request headers

Envoy supports adding dynamic values to request and response headers like %HOSTNAME% but currently the % is escaped over concerns about sending invalid data to Envoy which will be rejected and cause failures.  These changes allow a specific set of known good Envoy dynamic headers to pass through unescaped and continue escaping the % for anything else.

Fixes #3234 


internal/dag/status_test.go
```diff
@@ -804,6 +804,24 @@ func TestDAGStatus(t *testing.T) {
 		},
 	}
 
+	proxyValidReuseCaseExampleCom := &contour_api_v1.HTTPProxy{
```
sunjayBhatia, line 807: Should this test be part of a separate PR?

erwbgy: This is part of https://github.com/projectcontour/contour/pull/3231 which has been approved but not been merged yet.  Looks like because that was on the master branch it gets included here so I'll try to remember not to do that next time.

internal/dag/httpproxy_processor.go
```diff
@@ -705,7 +705,8 @@ func (p *HTTPProxyProcessor) validHTTPProxies() []*contour_api_v1.HTTPProxy {
 			valid = append(valid, proxy)
 			continue
 		}
-		fqdnHTTPProxies[proxy.Spec.VirtualHost.Fqdn] = append(fqdnHTTPProxies[proxy.Spec.VirtualHost.Fqdn], proxy)
+		fqdn := strings.ToLower(proxy.Spec.VirtualHost.Fqdn)
```
sunjayBhatia, line 708: I think you'll have to do a rebase and remove the commits referencing the case-insensitive fqdn code

internal/dag/status_test.go
sunjayBhatia: I think you'll have to do a rebase and remove the commits referencing the case-insensitive fqdn code (same as above comment)

sunjayBhatia (changes requested):
Looks good to me minus the rebase changes requested 👍🏽

internal/dag/status_test.go
erwbgy: Thanks @sunjayBhatia, that did the trick.

sunjayBhatia (approved):

internal/dag/policy_test.go
```diff
@@ -295,3 +295,116 @@ func TestLoadBalancerPolicy(t *testing.T) {
...
+		"simple percentage escape": {
+			hp: &contour_api_v1.HeadersPolicy{
+				Set: []contour_api_v1.HeaderValue{{
+					Name:  "X-App-Weight",
+					Value: "100%",
+				}},
+			},
+			want: HeadersPolicy{
+				Set: map[string]string{
+					"X-App-Weight": "100%%",
```
stevesloka, line 327: Should this be escaped like this? Meaning, adding the double `%%`?

internal/dag/policy.go
```diff
@@ -125,7 +126,43 @@ func headersPolicyRoute(policy *contour_api_v1.HeadersPolicy, allowHostRewrite b
 func escapeHeaderValue(value string) string {
 	// Envoy supports %-encoded variables, so literal %'s in the header's value must be escaped.  See:
 	// https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#custom-request-response-headers
-	return strings.Replace(value, "%", "%%", -1)
+	// Only allow a specific set of known good Envoy dynamic headers to pass through unescaped
+	if !strings.Contains(value, "%") {
```
stevesloka, line 130: Should this check that it begins / ends with the `%`?

sunjayBhatia: i think the header value can be a combination of variables and static content e.g. `foo: %HOSTNAME% bar`

internal/dag/policy_test.go
sunjayBhatia: i think this is correct b/c contour will send the escaped value to Envoy, since we want a literal `%` to show up in the resulting header value we have to escape it (https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#custom-request-response-headers)

internal/dag/policy.go
erwbgy: This just avoids all of the processing below if there is no % in the header value.

There are tests to ensure that %'s that are not part of a dynamic header are properly escaped, including with multiple dynamic headers and other non-dynamic content.

internal/dag/policy_test.go
erwbgy: Yes, that's right. The user specifies the header value as "100%", Contour sends it to Envoy as "100%%" and it ends up being sent as "100%".

Without the escaping Envoy would throw a route configuration error as anything with a single % is considered to be a dynamic header and 100% is not valid.

stevesloka: gotcha, makes sense, thanks for clarifying for me @sunjayBhatia & @erwbgy

stevesloka (approved):
/lgtm 🎉 

Just need some docs to follow up and spell out how this all worked. Thanks @erwbgy!!

sunjayBhatia:
@stevesloka worth a release note tag?

stevesloka:
Yup! Release note added!

skriss:
> /lgtm 🎉
> 
> Just need some docs to follow up and spell out how this all worked. Thanks @erwbgy!!

Looks like https://github.com/projectcontour/contour/blob/main/site/docs/main/config/request-rewriting.md#header-rewriting would be the spot to add docs. @erwbgy if you can get those added to this PR we should be good to merge.

skriss (approved):
LGTM, thanks @erwbgy - I'll let @stevesloka take a last look before merging.

stevesloka (approved):
/lgtm

PR #3240, opened 2021-01-05
add CODEOWNERS file to make Github reviews work properly

stevesloka:
Currently we have Github configured to auto-request reviews from the @maintainers group, however, it's not working. I think this is because the PR's are opened against the `main` branch which is protected. Adding a `CODEOWNERS` file will allow the review to work properly: https://docs.github.com/en/free-pro-team@latest/github/creating-cloning-and-archiving-repositories/about-code-owners

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,5 @@
+# These owners will be the default owners for everything in
+# the repo. Unless a later match takes precedence,
+# @maintainers will be requested for review when someone
+# opens a pull request.
+*       @projectcontour/maintainers


sunjayBhatia (approved):

sunjayBhatia:
might have to be `@org/team-name` like `@envoyproxy/api-shepherds`: https://github.com/envoyproxy/envoy/blob/7fa6aa591231afe18d53200e15eeaa919712a3e6/CODEOWNERS#L6

stevesloka:
@sunjayBhatia the docs weren't super clear there, but I think you are right, updated. =)

sunjayBhatia:
> @sunjayBhatia the docs weren't super clear there, but I think you are right, updated. =)

and it might have to have a file path or glob before the owner designation? https://github.com/dotnet/samples/blob/1bb5dc93e2ce9edafddbee8c9dfd25cd31dafdf1/.github/CODEOWNERS#L15

stevesloka:
>and it might have to have a file path or glob before the owner designation?

ha yup you are right, I glossed over that as a comment.

sunjayBhatia (approved):

skriss (approved):
LGTM

PR #3245, opened 2021-01-11
Bump to Envoy 1.17.0

sunjayBhatia:
Merge after https://github.com/projectcontour/contour/pull/3250


sunjayBhatia:
Just a draft right now b/c Dockerhub is still missing the Envoy release image

sunjayBhatia:
Hmm, not getting this error running the integration tests locally

youngnick (changes requested):
Image changes are fine, but we'll need to update the support page as well. Are there any API changes?

sunjayBhatia:
> Image changes are fine, but we'll need to update the support page as well. Are there any API changes?
> 
> Looks like the integration tests have failed because Envoy didn't start, seems like we may have other things to fix.

```
$ grep -r v1.16.2 .
./site/_resources/compatibility-matrix.md:[12]: https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.16.2
./site/_resources/upgrading.md:All users should ensure the Envoy image version is `docker.io/envoyproxy/envoy:v1.16.2`.
./site/_resources/upgrading.md:[27]: https://www.envoyproxy.io/docs/envoy/v1.16.2/version_history/current
```

These are the refs left for 1.16.2, I thought these would be updated when we cut a release

As far as API changes, didn't see anything significant: https://www.envoyproxy.io/docs/envoy/v1.17.0/version_history/current

Some interesting notes:
- `listener: injection of the TLS inspector has been disabled by default. This feature is controlled by the runtime guard envoy.reloadable_features.disable_tls_inspector_injection.`
- `tls: removed RSA key transport and SHA-1 cipher suites from the client-side defaults.`
- Some possibly relevant notes in the deprecations section: https://www.envoyproxy.io/docs/envoy/v1.17.0/version_history/current#deprecated

youngnick:
By the support matrix I meant the (newly renamed) [compatibility matrix](https://projectcontour.io/resources/compatibility-matrix/), and yes, we'll need to do it as part of the release, you're right.

sunjayBhatia:
> Hmm, not getting this error running the integration tests locally

jk I must have changed branches in the middle of running the tests

Some deprecated fields are barfing, will need a code change

stevesloka:
I was testing the externalName issue and ran into this on my cluster upgrading to Envoy v1.17.0:

```
[2021-01-12 03:01:30.390][1][info][main] [source/server/server.cc:782] exiting
Proto constraint validation failed (UpstreamTlsContextValidationError.CommonTlsContext: ["embedded message failed validation"] | caused by CommonTlsContextValidationError.TlsCertificateSdsSecretConfigs[i]: ["embedded message failed validation"] | caused by SdsSecretConfigValidationError.Name: ["value length must be at least " '\x01' " runes"]): common_tls_context {
  tls_certificate_sds_secret_configs {
    sds_config {
      path: "/config/resources/sds/xds-tls-certificate.json"
    }
  }
  validation_context_sds_secret_config {
    sds_config {
      path: "/config/resources/sds/xds-validation-context.json"
    }
  }
}
```

skriss (approved):
LGTM and integration tests looks OK. We'll probably all want to do some more poking before release to ensure there are no other breaks in features not covered by the integration tests.

stevesloka:
@sunjayBhatia does this work backwards with an older version of Envoy? I'm thinking of upgrade docs and also what might happen to the operator upgrades that we've been discussing (@danehans).

sunjayBhatia:
> @sunjayBhatia does this work backwards with an older version of Envoy? I'm thinking of upgrade docs and also what might happen to the operator upgrades that we've been discussing (@danehans).

I think it would since it’s only config validation but I’ll check to make sure

sunjayBhatia:
> @sunjayBhatia does this work backwards with an older version of Envoy? I'm thinking of upgrade docs and also what might happen to the operator upgrades that we've been discussing (@danehans).

See https://github.com/projectcontour/contour/pull/3250 which should validate the config changes are safe to apply, keeping them in a separate PR so theyre validated by CI independently

skriss:
> @sunjayBhatia does this work backwards with an older version of Envoy? I'm thinking of upgrade docs and also what might happen to the operator upgrades that we've been discussing (@danehans).

@stevesloka good point, I guess we'll want to have users first upgrade the Contour deployment, then upgrade the Envoy daemonset, if they want to not drop traffic?

sunjayBhatia:
> > @sunjayBhatia does this work backwards with an older version of Envoy? I'm thinking of upgrade docs and also what might happen to the operator upgrades that we've been discussing (@danehans).
> 
> @stevesloka good point, I guess we'll want to have users first upgrade the Contour deployment, then upgrade the Envoy daemonset, if they want to not drop traffic?

Could we consider doing a two-step release to support Envoy 1.17.0 so we don't need people to care? Release 1.12.0 with the changes to configs, then another release to bump Envoy

Regardless, that upgrade order seems like it would generally correct, unless we start using an Envoy feature that is completely new, then the opposite would be needed

stevesloka:
@sunjayBhatia so the newer config won't work with older versions of Envoy? Say v1.16.2?

stevesloka:
So looks like it will work with v1.16.x, but yeah we'd need to upgrade Contour first before upgrading Envoy's.

stevesloka (approved):
/lgtm

sunjayBhatia:
cc @danehans for operator concerns

skriss:
this should be OK to merge now as nick's comments have been addressed.

PR #3251, opened 2021-01-12
add local rate limiting support

### LINKED ISSUE #3255: add local rate limiting support
child issue of #370

- [x] write code
- [x] add feature tests
- [x] add integration tests
- [x] write docs

skriss:
Adds support for Envoy's local HTTP rate limiting filter,
which allows requests to be rate-limited on a per-Envoy
basis, at the virtual host and route levels.

Closes #3255.


apis/projectcontour/v1/httpproxy.go
```diff
@@ -320,6 +323,48 @@ type Route struct {
...
+	Local *LocalRateLimitPolicy `json:"local,omitempty"`
+}
+
+// LocalRateLimitPolicy defines local rate limiting parameters.
+type LocalRateLimitPolicy struct {
+	// Requests defines how many requests per unit of time should
+	// be allowed before rate limiting occurs.
+	// +required
+	// +kubebuilder:validation:Minimum=1
+	Requests uint32 `json:"requests"`
```
stevesloka, line 345: Would it make sense to combine `Requests` & `Unit` into a duration field so you could specify it once like: `1m`?

go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/ahmetb/gen-crd-api-reference-docs v0.1.5
 	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751 // indirect
 	github.com/alecthomas/units v0.0.0-20190717042225-c3de453c63f4 // indirect
-	github.com/envoyproxy/go-control-plane v0.9.7
+	github.com/envoyproxy/go-control-plane v0.9.8
```
stevesloka, line 9: Maybe spin out the go-control-plane upgrade into it's own PR?

internal/dag/policy.go
```diff
@@ -311,3 +311,41 @@ func prefixReplacementsAreValid(replacements []contour_api_v1.ReplacePrefix) (st
...
+	var fillInterval time.Duration
+	switch in.Local.Unit {
+	case "second":
+		fillInterval = time.Second
+	case "minute":
+		fillInterval = time.Minute
+	case "hour":
+		fillInterval = time.Hour
+	default:
+		// TODO should not be possible due to kubebuilder validation
```
stevesloka, line 331: maybe just debug log and return in the event this does get hit?

internal/envoy/v3/ratelimit.go
```diff
@@ -0,0 +1,61 @@
...
+
+// LocalRateLimitConfig returns a config for the HTTP local rate
+// limit filter.
+func LocalRateLimitConfig(config *dag.LocalRateLimitPolicy) *any.Any {
+	if config == nil {
+		return nil
+	}
+
+	c := &envoy_config_filter_http_local_ratelimit_v3.LocalRateLimit{
+		// TODO it'd be nice for this to be unique per vhost/route
```
stevesloka, line 33: yup this would be super cool to be able to get stats per vshost/route.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -320,6 +323,48 @@ type Route struct {
...
+	// Burst defines the number of requests above the requests per
+	// unit that should be allowed within a short period of time.
+	// +optional
+	Burst uint32 `json:"burst,omitempty"`
+
+	// ResponseStatusCode is the HTTP status code to use for responses
+	// to rate-limited requests. If not specified, the Envoy default of
+	// 429 (Too Many Requests) is used.
+	// +optional
+	ResponseStatusCode uint32 `json:"responseStatusCode,omitempty"`
```
stevesloka, line 362: Should we validate this is a 400x or 500x code? It would be weird to maybe return a 200 status from here, but if a user did it up to them I guess. I don't want to add too much restriction here, however.

go.mod
skriss: https://github.com/projectcontour/contour/pull/3259

internal/envoy/v3/ratelimit.go
skriss: :+1: gotta figure out a naming scheme. Do you know off the top of your head if we do this anywhere else (vhost/route-specific stat prefix)?

apis/projectcontour/v1/httpproxy.go
skriss: It could work, but it'd be a little weird because the relationship is "n <requests> / 1 <unit>", so if you merged it into a duration string, you'd write e.g. `30s` to mean "30 requests per second". Personally I'd rather keep it as-is. We could rename `Requests` to `RequestsPerUnit` or `Rate` or something else.

internal/dag/policy.go
skriss: Yeah, we could do that and return an empty rate limit policy, or return an error, record it as a warning in conditions, and not configure a rate limit policy. I guess the latter is probably best since it surfaces the issue to a user, just annoying to have to add an `error` return parameter to helpers like this :/

apis/projectcontour/v1/httpproxy.go
skriss: yep, actually Envoy ignores it if it's <400, so we should enforce that too (https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/local_ratelimit/v3/local_rate_limit.proto#extensions-filters-http-local-ratelimit-v3-localratelimit)

apis/projectcontour/v1/httpproxy.go
stevesloka: nah your design is correct, I wasn't thinking about it correctly.

internal/envoy/v3/ratelimit.go
stevesloka: We have cluster & listener stats and just append the cluster name or vhost name to the stat. 

In the following example, the cluster is `projectcontour-marketing_info_80` which is `projectcontour-marketing` is the namespace, `info` is the svc name, and `80` is the listener it's tied to. 
```
envoy_cluster_upstream_cx_length_ms_bucket{envoy_cluster_name="projectcontour-marketing_info_80",le="50"}
```

So maybe we need something that's `vhost_route_%something%`. The tricky part is a route is the sum of the conditions which could also include a header.

apis/projectcontour/v1/httpproxy.go
stevesloka: nice!

apis/projectcontour/v1/httpproxy.go
stevesloka: At one point we talked about (or maybe it was just in my head), of coming up with all the error logic to figure out what's wrong with stuff, then can return there before you get too deep into the logic, but yeah it gets messy quickly.

apis/projectcontour/v1/httpproxy.go
skriss: Added validation in https://github.com/projectcontour/contour/pull/3251/commits/e0b1f94af0bfe2a133982518947dbda46a6dca46

internal/envoy/v3/ratelimit.go
skriss: :+1: i'll see what i can figure out

internal/dag/policy.go
skriss: I ended up returning an error and reporting it as an error condition on the proxy if this happens. Though unlikely, it probably still makes sense to have the check and the user-visible error.

internal/envoy/v3/ratelimit.go
skriss: yeah, so if we want per-route stats, it'll have to be some big long ugly string that incorporates the prefix and any header matches. I think it could work, but it's a bit ugly.

Alternately, we could just roll everything up to the vhost, so only have a unique stats prefix per vhost. You lose some of the granular information, but it's maybe more easily parsed.

skriss: I implemented a per-vhost prefix in https://github.com/projectcontour/contour/pull/3251/commits/021a0c664c4570a3f50963287031413214ece78e, so all routes in a given vhost will roll up to the vhost metrics. If we think having per-route metrics is important, I can work out a string representation of a route.

skriss:
Marking this ready for review but we should get #3178 merged before this.

_integration/testsuite/httpproxy/018-local-rate-limiting.yaml
```diff
@@ -0,0 +1,200 @@
...
+      local:
+        requests: 1
+        unit: hour
+  routes:
+  - services:
+    - name: echo
+      port: 80
+---
+
+# Wait for the service to have endpoints before trying to make
```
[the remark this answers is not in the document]
skriss: yeah, this works much better than the original non-deterministic impl I had where I just hoped that we'd get a 200 in less than 20 attempts.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -320,6 +323,51 @@ type Route struct {
...
+	// Many Requests) is used.
+	// +optional
+	// +kubebuilder:validation:Minimum=400
+	// +kubebuilder:validation:Maximum=599
+	ResponseStatusCode uint32 `json:"responseStatusCode,omitempty"`
+
+	// ResponseHeadersToAdd is an optional list of response headers to
+	// set when a request is rate-limited.
+	// +optional
+	ResponseHeadersToAdd []HeaderValue `json:"responseHeadersToAdd,omitempty"`
```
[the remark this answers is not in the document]
skriss: Good question. I guess without having a use case for needing them here, I'd opt to start by not supporting them, knowing that we could add support later if needed.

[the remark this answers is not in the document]
sunjayBhatia: I guess in addition to documentation we would have to escape % characters so if someone does pass a variable or that character Envoy doesn’t reject the update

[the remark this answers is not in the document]
skriss: Yeah, true, so maybe just as easy to support the dynamic vals. I'll look at adding this.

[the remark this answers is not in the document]
skriss: Added better header validation + tests in https://github.com/projectcontour/contour/pull/3251/commits/332ffad3db9e925a0c90af074a126b3febdbb402

internal/dag/policy_test.go
```diff
@@ -404,7 +404,130 @@ func TestHeadersPolicy(t *testing.T) {
...
+			want: &RateLimitPolicy{
+				Local: &LocalRateLimitPolicy{
+					MaxTokens:          10,
+					TokensPerFill:      10,
+					FillInterval:       time.Minute,
+					ResponseStatusCode: 431,
+				},
+			},
+		},
+		"custom response headers to add": {
```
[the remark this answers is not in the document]
skriss: Added better header validation + tests in 332ffad

internal/dag/policy.go
```diff
@@ -311,3 +311,52 @@ func prefixReplacementsAreValid(replacements []contour_api_v1.ReplacePrefix) (st
...
+		fillInterval = time.Second
+	case "minute":
+		fillInterval = time.Minute
+	case "hour":
+		fillInterval = time.Hour
+	default:
+		return nil, fmt.Errorf("invalid unit %q in local rate limit policy", in.Local.Unit)
+	}
+
+	rp := &RateLimitPolicy{
```
stevesloka, line 336: Might need one more additional validation that the configured fill_interval is more than `50ms`. 

ref: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/local_ratelimit/v3/local_rate_limit.proto#envoy-v3-api-field-extensions-filters-network-local-ratelimit-v3-localratelimit-token-bucket

skriss: We should be OK since users can't configure `fill_interval` directly, it's inferred based on the `unit`, which has a smallest allowed value of `second`.

stevesloka: Nice!

stevesloka (approved):
/lgtm :tada: Nice work @skriss!!

sunjayBhatia (approved):

skriss:
🎉 

FWIW the [siteproof failure](https://github.com/projectcontour/contour/pull/3251/checks?check_run_id=1736158025) is spurious, it doesn't handle new docs pages correctly.

PR #3254, opened 2021-01-13
Add integration test to verify Host header rewrite policy

sunjayBhatia:
Regression test to ensure host header rewrite policy is correct and
rewrites with literal value, regression introduced in xDS v2 -> v3
switch.

Merge after https://github.com/projectcontour/contour/pull/3252


sunjayBhatia:
Added integration test should fail before https://github.com/projectcontour/contour/pull/3252 is merged into the branch

sunjayBhatia:
See check run that fails as intended: https://github.com/projectcontour/contour/pull/3254/checks?check_run_id=1696330990

stevesloka (approved):
/lgtm but feels weird to merge with it breaking into `main`, so maybe can merge #3252 first, then watch this one go green?

sunjayBhatia:
> /lgtm but feels weird to merge with it breaking into `main`, so maybe can merge #3252 first, then watch this one go green?

Yep, I would merge this after the one that fixes the regression so `main` stays green

skriss (approved):
LGTM!

skriss:
@sunjayBhatia do you mind bumping the integration test number to 017 to avoid conflicting with https://github.com/projectcontour/contour/pull/3236? (I don't think it actually causes any problems to have dupes, but may as well avoid it since we have the numbers).

skriss:
well I thought #3236 would merge before this, but we're waiting on docs for that one. I'll merge this one now but our integration test numbering will be missing 016 for a bit.

PR #3256, opened 2021-01-13
Sets Envoy Volume Mounts to ReadOnly

danehans:
Sets the volume mounts for the Envoy container to  `readOnly: true`. By default, volume mounts are R/W. Since no process in the Envoy container should need to write to these volumes, it's more secure to mount them as read only.


stevesloka (approved):
/lgtm

skriss (approved):
LGTM

Issue #3086, opened 2020-11-01
TCP Layer 4 Routing


JasonMorgan:
I'd like the ability to forward a particular port to a particular service without needing HTTPS. An example being, I run a CI tool called concourse and it gives me the ability to launch ssh sessions to ci jobs. In order to broker the ssh session the cli tool opens a connection to the concourse web app over port 2222. Using an NGINX ingress I accomplish this by creating a config map that instructs the ingress controller to forward everything it gets on 2222 to my concourse web service, `tcp: {2222: "tools/ci-concourse-web:2222"}`. As of right now this doesn't seem possible with Contour.

stevesloka:
I see this as a new CRD, something like `TCPProxy` where you can define the L4 bits, IP Address, ports, etc. I think the starting bits can be limited and just document that (i.e. we're not going to deal with IP address management, etc), but overall I think this can be helpful for the community while still be limited compared to other projects that are full "L4" style proxies.  

We just discussed this a bit in #2809 (https://github.com/projectcontour/contour/issues/2809#issuecomment-714911351) as well.

youngnick:
The problem with defining things like IP address is that it requires the Envoy deployment to either have that IP address present on the Linux host, or have certain Linux features enabled so that it can add any address, even if it's not routable.

One of the reasons I've been reluctant to just go ahead with this is that there is a long list of things that you have to achieve with your Envoy deployment before you can request an arbitrary IP address and port (which I see as fundamental for this sort of load balancing). Contour has no way to determine how Envoy will react to sending configs that involve L4 details like IP address without just sending it, and if the config is rejected, Envoy will stop receiving updates from Contour until restarted (this is #1176).

I think that the approach nginx seems to take (you can forward any port on whatever IP address(es) Envoy is listening on) is probably more achievable, but will require a new CRD, as you say.

stevesloka:
Yeah I was starting too detailed and working out. I think an initial implementation could be to do how we handle listeners today which is to listen on `0.0.0.0` over ports `80` & `443` (configurable). 

My suggestion is to allow L4 load balancers to be defined the same way, listening on ports that are defined in the CRD over `0.0.0.0`. A user could add those ports to the Envoy service to expose them from an external load balancer if that was their environment. 

This leaves the other bits like assigning IPs, etc, etc, etc to another controller or implementation based upon the environment, but is simple enough to give users a bit more functionality to fill in any gaps today.

spines83:
[...]

Issue #2922, opened 2020-09-18
Exposing TCP with FQDN for mongodb


binnythomas-1989:
Hello Guys,

Kubernetes: 1.17.5
Contour- contour:v1.8.1

Im aware that we can expose TCP port on Contour using the below link
https://github.com/projectcontour/contour/blob/main/site/docs/main/httpproxy.md#tls-termination-at-the-edge

Is it possible say to expose the port 27017 which is a default mongo db port by the haproxy resource on fqdn.

On implementation of the link, im able to access the port as follows

mongo --host virtualhost:443 -tls

I would like to access it with Virtualhost:27017

youngnick:
The way this can be done, is that you specify that all secure (HTTPS or TLS) traffic is served out of port 27017 for that Contour installation. Then, any requests to the set of Envoys Contour controls will need to arrive at 27017 instead of 443. You can't run 443 TLS and 27017 tls at the same time in Contour right now.

To do that, you need to pass the _Contour_ deployment the command-line flag `--envoy-service-https-port=27017`. In the example deployment, this is set to 443.

To say this another way, Contour is capable of listening on two ports, an insecure one (80 in the example deployment) and a secure one (443 in the example deployment). Only the secure port will do SNI routing, and you can change it to whatever you want. That means, however, that you can no longer use that same Contour install for HTTPS serving.

binnythomas-1989:
Since the contour runs as a Daemonset could we just use it for one tcp scenario?

So how I worked around it is like this.
```
xxx macos-signing-ansible (master) $ kubectl get svc -n projectcontour
NAME      TYPE           CLUSTER-IP      EXTERNAL-IP     PORT(S)                                      AGE
contour   ClusterIP      172.26.45.185   <none>          8001/TCP                                     4d11h
envoy     LoadBalancer   172.26.35.226   xx.xxx.xx.xxx   80:31032/TCP,443:31458/TCP,27017:30194/TCP   4d11h
xx macos-signing-ansible (master) $ 
````
I have created another listener  to point 27017 to 443. In this case would I be able to do tcp and 443?

youngnick:
Contour runs as a Deployment in the example YAMLs, Envoy runs as a Daemonset. You can absolutely use a single Contour-Envoy pair for the 27017 use case, but yeah, that does not seem ideal.

[...]

PR #3263, opened 2021-01-19
design: Add design for multiple Envoy listeners

youngnick:
As part of implementing Service APIs, it's necessary to reconsider Contour's insecure to secure redirect functionality.
This document suggests keeping the functionality, while allowing additional listening ports to be defined.

Updates #1300
Updates #2922
Updates #3086

Blocks #3213 until we decide one way or another.

### DIFF
--- a/design/multiple-listeners-design.md
+++ b/design/multiple-listeners-design.md
@@ -39,8 +39,9 @@ These ports must also have a place to go to on your backend's Service.
 
 Currently, Contour has command line flags to specify the ports that the Envoy containers should listen on.
 There are only two ports you can specify, secure and insecure, which default to `8443` and `8080` respectively.
-It's assumed that as part of the installation, you have a Service translating from port `80` to `8080` and `443` to `8443`.
-This is part of the example YAMLs.
+It's assumed that as part of the installation, your ports are translated from port `80` to `8080` and `443` to `8443`.
+In the example YAMLs, this is done with a `hostPort` on the Envoy daemonset, and the `Service` refers to `80` and `443`, not the translated ports.
+The important part for most Contour installations is the ports that are defined on the Envoy Service.
 
 If you have a Service in the packet forwarding path, then any extra ports you ask Envoy to serve on *must* be exposed via that Service for traffic to reach them.
 
@@ -63,7 +64,7 @@ These ports are referred to as "externally visible ports" throughout this docume
 with "externally visible secure port" meaning "whatever port an external user will go to to get to the secure port",
 and similarly for the "externally visible insecure port".
 
-Importantly, specifying a port for Envoy's secure or insecure listening port will be a fatal error for Contour.
+Importantly, pecifying a port for Envoy's secure or insecure listening port that does not match the a port on the Service will be a fatal error for Contour.
 This makes a simple misconfiguration that would result in no traffic flowing much more obvious.
 
 If users require, at a later date, Contour may add a feature that disables the service lookup, and assumes that the externally visible ports are the specified ones (that is, that there is no port translation to the Envoy processes).
@@ -88,6 +89,9 @@ The rules for how this `port` field interacts with the rest of Contour would the
     - TLS details set - port is the external version of the secure or insecure ports - treat as though no port was set. Redirection will be created.
     - no TLS details set - any port other than the insecure or secure ports, allow. A new listener will be created if one doesn't exist.
 
+A `port` setting, if set, will be included in the uniqueness check that currently only checks the FQDN.
+Two HTTPProxies that both contain the same FQDN and port will both be rendered invalid (which mirrors the current behavior when only checking FQDN).
+
 Contour will watch the `envoy` Service, and check that the supplied port matches one port on that Service, and apply slightly different logic if it's also one of the ports specified as secure or insecure in the config.
 
 ### Service APIs processing
@@ -98,8 +102,7 @@ This design will be included in the Service APIs design document.
 Pending agreement on the high-level design
 ## Alternatives Considered
 ### Don't change anything
-We can keep the same "only two ports" requirements going forward, but that will not allow us to solve the issues:
-[TCP Layer 4 Routing](https://github.com/projectcontour/contour/issues/3086)
+We can keep the same "only two ports" requirements going forward, but that will not allow us to solve this issue:
 [Exposing TCP with FQDN for mongodb](https://github.com/projectcontour/contour/issues/2922)
 
 ## Compatibility
--- a/design/multiple-listeners-design.md
+++ b/design/multiple-listeners-design.md
@@ -18,10 +18,12 @@ This feature is achieved by having Contour only serve two ports - a secure and a
 
 However, when designing Contour's support for the Service APIs, this feature appears to conflict part of the Service APIs contract; namely that you can create an arbitrary number of listeners, on any port, even for the subset of the Service APIs that Contour is targeting.
 
-This document outlines a design for adding support for arbitrary port listeners that attempts to allow marry to ease-of-use of the automatic redirect to the configurability of being able to add arbitrary extra listening ports to Envoy's configuration.
+This document outlines a design for adding support for **additional** arbitrary port listeners that attempts to allow marry to ease-of-use of the automatic redirect to the configurability of being able to add arbitrary extra listening ports to Envoy's configuration.
 
 It should be noted here that using this feature, combined with a TCPProxy with no TLS configured, will allow a very bsaic facsimile of a Layer 4 load balancer to be implemented by Contour.
-Contour's main purupose is still a Layer 7 Ingress Controller, not a Layer 4 load balancer.
+Contour's main purpose is still a Layer 7 Ingress Controller, not a Layer 4 load balancer.
+Contour will still require a HTTP port and a HTTPS port to be configured, this proposal allows for **additional** ports to be added.
+These ports must also have a place to go to on your backend's Service.
 
 ## Goals
 - Allow adding multiple (an arbitrary number) of extra listeners to Contour's Envoy fleet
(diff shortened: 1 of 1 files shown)


youngnick:
I've opened this one because I think that the Gateway object in Service-APIs implies that we should allow the creation of extra Listeners. We also have outstanding requests for similar functionality.

I'm going to say again though, that although this allows you to forward some traffic on an arbitrary port at Layer 4, I don't believe this is "Adding layer 4 load balancing to Contour". It's missing large swaths of functionality for that (most importantly, the ability to choose your serving IP), and is heavily dependent on your Envoy being configured correctly (trying to bind to ports under 1024 will fail if your container is not running, and so on).

design/multiple-listeners-design.md
```diff
@@ -0,0 +1,87 @@
...
+The rules for how this `port` field interacts with the rest of Contour would be this:
+- no port set
+    - TLS details set - default secure port, redirect will be created for you
+    - no TLS details set - default insecure port, same `permitInsecure` behavior.
+- port set
+    - TLS details set - any port other than the insecure or secure ports, allow. No redirection created for you.
+    - no TLS details set - any port other than the insecure or secure ports, allow.
+    - use the insecure or the secure port - fatal error, HTTPProxy will not be processed.
+
+The documents for the `port` field must clearly state that it is used for specifying a non-default port *only*. Don't use it for port 80 and 443.
```
jpeach, line 67: What's the rationale for this? Note that it can create new error cases (proxy is OK, then contour config changed and proxy is now broken)

youngnick: The idea was just to make it easier to write the handling for the "supplied a port" case.

I'll make this clearer, but the port that is specified in the HTTPProxy *must* be the port on the **`Service`**, not the port on the Pod. This is one of the things I need to address in the detailed design - Contour needs to know what port to put in URLs, and we currently don't distinguish that from "what port to tell Envoy to bind to".

In the example YAML, we tell Envoy to bind to `8080` and `8443` (to avoid requiring the container to be run as root), but assume that those will be translated to `80` and `443` respectively by the Service. We just haven't documented this requirement really.

Part of changing this is that we need to make the requirements explicit somehow.

design/multiple-listeners-design.md
```diff
@@ -0,0 +1,111 @@
+# Automatic redirection with arbitrary listeners
+
+Status: Accepted
```
youngnick, line 3: As we've done for other current designs, setting this to `Accepted` because it won't be merged unless it is accepted.

design/multiple-listeners-design.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Creating a redirection
+For the redirect to be able to work currently, the example YAML and the Operator currently have the Envoy listen on port `8443` and `8080` respectively *inside its Pod*, but this is translated out to `80` and `443` by the `Service` or `Type: Loadbalancer` included there.
+This means that the redirect generated can redirect from `http://somedomanin.com` to `https://somedomain.com` and have everything work.
+
+As part of this change, Contour will include the port number on the HTTPS redirect, if the externally visible secure port is anything other than `443`.
+
+### HTTPProxy processing
+
+For HTTPProxy resources, Contour will add an optional `port` field to the `virtualhost` YAML stanza, which matches the externally visible port for the Envoys.
```
youngnick, line 79: I'm open to discussion on the name here, as it's the translated port. Not sure if that means we should give it a different name, or if we should just be *very* clear in the documentation what port we are talking about.

design/multiple-listeners-design.md
```diff
@@ -0,0 +1,111 @@
...
+Specifying a port that does not match at least one `port` on the Envoy `Service` will result a fatal error for HTTPProxy processing, and an Error condition on the HTTPProxy.
+
+The rules for how this `port` field interacts with the rest of Contour would then be this:
+- no port set
+    - TLS details set - default secure port, redirect will be created for you
+    - no TLS details set - default insecure port, same `permitInsecure` behavior.
+- port set
+    - TLS details set - any port other than the external version of the secure or insecure ports, allow. A new listener will be created if it doesn't exist. No redirection created for you.
+    - TLS details set - port is the external version of the secure or insecure ports - treat as though no port was set. Redirection will be created.
+    - no TLS details set - any port other than the insecure or secure ports, allow. A new listener will be created if one doesn't exist.
```
sunjayBhatia, line 89: Should we add some detail around here or the detailed design section about what will happen if a port is set but it conflicts with an existing listener?

e.g.
- existing HTTPProxy that asks to listen on port 9000 with no TLS details set
  -  gets Envoy listener + filter chain with plaintext transport socket
- new HTTPProxy (different vhost) that asks to listen on port 9000 with TLS details set
  - TLS transport socket would conflict with plaintext transport socket
 
With Service APIs this configuration would be "not mergeable", seems the `port` in addition to vhost could be a uniqueness constraint to add onto HTTPProxy resources if I'm understanding correctly?

youngnick: Yes, that's a great point, thanks Sunjay. Currently any two HTTPProxy objects that match on vhost name render *both* proxies invalid. We will need to extend that to ports as well, there are some cases around what to do with mixed port/no port configurations that match on vhosts as well. That is definitely something for the detailed design section.

design/multiple-listeners-design.md
```diff
@@ -0,0 +1,111 @@
...
+- UDP Layer 4 load balancing support
+
+## High-Level Design
+
+### Extra problems to consider
+
+Currently, Contour has command line flags to specify the ports that the Envoy containers should listen on.
+There are only two ports you can specify, secure and insecure, which default to `8443` and `8080` respectively.
+It's assumed that as part of the installation, you have a Service translating from port `80` to `8080` and `443` to `8443`.
+This is part of the example YAMLs.
```
stevesloka, line 43: nit: The examples use hostPorts and change the defaults to 80/443: https://github.com/projectcontour/contour/blob/main/examples/contour/02-service-envoy.yaml#L18-L23

design/multiple-listeners-design.md
```diff
@@ -0,0 +1,111 @@
...
+
+If you do something else (like running a separate set of machines for Envoy by themselves), then currently you *must* have the secure port set to `443` and the insecure to `80`.
+Otherwise generated URLs will have no port, which is not correct if you are running on nonstandard ports.
+
+### Configuring the secure and insecure ports
+
+Currently, Contour allows the configuration of the secure and insecure listening ports for Envoy using the `--envoy-service-https-port` and the `--envoy-service-http-port` parameters respectively.
+(It's fair to say that these parameters are confusingly named).
+
+These will also be exposed in the configuration file as `secureListenerPort` and `insecureListenerPort` respectively.
```
stevesloka, line 55: It feels like the service-api spec should be able to define this and not rely on the implementation.

design/multiple-listeners-design.md
```diff
@@ -0,0 +1,111 @@
...
+
+The Contour Operator may then add support to its `Contour` CRD for tweaking this value.
+
+At startup, and throughout its lifetime, Contour will watch the Envoy service and check that the ports specified are exposed, and, if they are translated, what they are translated *to*.
+Contour will use the externally visible port as the basis for creating redirections and matching ports for listeners.
+These ports are referred to as "externally visible ports" throughout this document,
+with "externally visible secure port" meaning "whatever port an external user will go to to get to the secure port",
+and similarly for the "externally visible insecure port".
+
+Importantly, specifying a port for Envoy's secure or insecure listening port will be a fatal error for Contour.
```
stevesloka, line 66: What does this statement mean? A few lines ago this was specified with a config file option, now it's invalid?

design/multiple-listeners-design.md
stevesloka, line 55: Could also use a convention that the service port named `http` will be the insecure port, the port named `https` will be the secure. This removes the need for the extra config options.

design/multiple-listeners-design.md
stevesloka: Maybe just not add to HTTPProxy at first revisit the extra overhead later?

design/multiple-listeners-design.md
```diff
@@ -0,0 +1,111 @@
...
+### Service APIs processing
+
+This design will be included in the Service APIs design document.
+
+## Detailed Design
+Pending agreement on the high-level design
+## Alternatives Considered
+### Don't change anything
+We can keep the same "only two ports" requirements going forward, but that will not allow us to solve the issues:
+[TCP Layer 4 Routing](https://github.com/projectcontour/contour/issues/3086)
```
stevesloka, line 102: This design doesn't address this issue. #3086 wants a non-TLS TCP path into the cluster. Adding additional ports will not solve this since it will still require them to have TLS termination.

danehans:
xref: https://github.com/projectcontour/contour-operator/pull/151 (operator network publishing API)

design/multiple-listeners-design.md
youngnick: Thanks, I'll update. The important thing here is that we can't rely on that, so the port that's visible in the Service is the key.

design/multiple-listeners-design.md
youngnick: I don't disagree, but don't want to block this on waiting for upstream.

youngnick: In the service APIs though, the ports don't have names. So it would need to be "the first port with `protocol: HTTP`" or something. I thought it was clearer to put the behavior in Contour, which allows us to backport it to HTTPProxy.

design/multiple-listeners-design.md
youngnick: ah, this is a mistake. It should read "specifying a port for Envoy's secure or insecure listening port that does not match the Service will be a fatal error for Contour."

design/multiple-listeners-design.md
youngnick: It seemed better to me that if we're going to add the functionality, we should add it into HTTPProxy first, while we do the rest of the Service APIs design, then it's easier to just consume it.

design/multiple-listeners-design.md
youngnick: ah, you're right. I'll remove.

design/multiple-listeners-design.md
youngnick: I've added a small note in here about the uniqueness constraint, but I think further detail is for the detailed design section.

design/multiple-listeners-design.md
stevesloka: The k8s service has names and the design says we're going to watch them, so given that you can match up the port name in the service to a port on a listener.

design/multiple-listeners-design.md
```diff
@@ -0,0 +1,114 @@
...
+
+The Contour Operator may then add support to its `Contour` CRD for tweaking this value.
+
+At startup, and throughout its lifetime, Contour will watch the Envoy service and check that the ports specified are exposed, and, if they are translated, what they are translated *to*.
+Contour will use the externally visible port as the basis for creating redirections and matching ports for listeners.
+These ports are referred to as "externally visible ports" throughout this document,
+with "externally visible secure port" meaning "whatever port an external user will go to to get to the secure port",
+and similarly for the "externally visible insecure port".
+
+Importantly, pecifying a port for Envoy's secure or insecure listening port that does not match the a port on the Service will be a fatal error for Contour.
```
stevesloka, line 67: nit: s/pecifying/specifying/

design/multiple-listeners-design.md
```diff
@@ -0,0 +1,114 @@
...
+
+For HTTPProxy resources, Contour will add an optional `port` field to the `virtualhost` YAML stanza, which matches the externally visible port for the Envoys.
+Specifying a port that does not match at least one `port` on the Envoy `Service` will result a fatal error for HTTPProxy processing, and an Error condition on the HTTPProxy.
+
+The rules for how this `port` field interacts with the rest of Contour would then be this:
+- no port set
+    - TLS details set - default secure port, redirect will be created for you
+    - no TLS details set - default insecure port, same `permitInsecure` behavior.
+- port set
+    - TLS details set - any port other than the external version of the secure or insecure ports, allow. A new listener will be created if it doesn't exist. No redirection created for you.
```
skriss, line 88: So this means you *only* get an automatic HTTP->HTTPS redirect if you are using the default secure listener port, right? Just making sure I'm following.

skriss (commented):
I'm fine with this proposal, though I think I'd also be fine saying that for our initial service-apis implementation, we only support two listeners (one http, one https) just like Contour works today, and any additional ones would be errors; then we could tackle this later. IOW, I'm not seeing that this *has* to be a prereq for an initial service-apis implementation, especially given the alpha state of the API.

design/multiple-listeners-design.md
skriss: Related to this, you could have conflicts across different APIs, e.g. an `HTTPProxy` conflicts with a `Gateway`.

design/multiple-listeners-design.md
youngnick: Yes, that's right.

youngnick:
I'm going to leave this open for now, it will be easy to rebase later, and I think we should look at this again soon (once we have some basic Service-APIs support).

stevesloka:
@cten this was the listeners design doc if you had anything to add. =)

nak3:
The additional arbitrary port listeners makes it possible to select one envoy by 2 private (`ClusterIP`) and public (`LoadBalancer`) services.
Knative currently deploys two envoy for private and private access, but we would like to use only one envoy with 2 services like this https://gist.github.com/danehans/c19b56fcef6219c6bf1af419fc8ead4e#file-gistfile1-txt (Thanks @danehans).
So, we want this feature not only for Gateway API support but also even now :smile:

nak3:
I think this issue is still not solved so is it alright to re-open?

youngnick:
We can leave it open; I suspect we will need it sooner than I thought.

moizrassiwala:
Is this available yet?

youngnick:
No, this design has not been finished. As I said earlier, I think we still need it.

tsaarni:
> Is this feature dead?

@cocodee Did you check this https://github.com/projectcontour/contour/releases/tag/v1.26.0

PR #3265, opened 2021-01-19
drop all TLS 1.1 references

### LINKED ISSUE #3246: docs: Update TLS references for TLS 1.1 support drop
TLS 1.1 used to be the Contour default version of TLS, the was deprecated and removed as part of #2777. The last remaining work there is for the documentation to be updated.

The behavior suggested in #2777 originally was that we would *allow* the configuration of TLS 1.1, both as a minimum version (as specified in https://projectcontour.io/docs/v1.11.0/config/tls-termination/), and in the HTTPProxy spec ( the `minimumProtocolVersion` field).

The new behavior everywhere is that, if TLS 1.1 is specified, it will be silently upgraded to TLS 1.2. This *shouldn't* be a problem because almost everything support TLS 1.2 now.

So, we need to update the docs to specify that using TLS 1.1 anywhere will be silently upgraded to TLS 1.2.

We already supply warnings for HTTPProxy objects that are using TLS 1.1, so I think that ensuring the docs match the current behavior is sufficient.

xref #2777

skriss:
Support for TLS 1.1 has been officially dropped from Contour.
This PR updates all code, documentation and examples to indicate
that 1.2 is now the minimum/default.

Closes #3246.
Updates #2777.
    
### DIFF
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -216,7 +216,9 @@ type TLS struct {
 	// If specified, the named secret must contain a matching certificate
 	// for the virtual host's FQDN.
 	SecretName string `json:"secretName,omitempty"`
-	// Minimum TLS version this vhost should negotiate
+	// MinimumProtocolVersion is the minimum TLS version this vhost should
+	// negotiate. Valid options are `1.2` (default) and `1.3`. Any other value
+	// defaults to TLS 1.2.
 	// +optional
 	MinimumProtocolVersion string `json:"minimumProtocolVersion,omitempty"`
 	// Passthrough defines whether the encrypted TLS handshake will be
--- a/examples/contour/01-contour-config.yaml
+++ b/examples/contour/01-contour-config.yaml
@@ -21,7 +21,7 @@ data:
     disablePermitInsecure: false
     tls:
     # minimum TLS version that Contour will negotiate
-    # minimum-protocol-version: "1.1"
+    # minimum-protocol-version: "1.2"
     # Defines the Kubernetes name/namespace matching a secret to use
     # as the fallback certificate when requests which don't match the
     # SNI defined for a vhost.
--- a/examples/contour/01-crds.yaml
+++ b/examples/contour/01-crds.yaml
@@ -923,7 +923,7 @@ spec:
                         description: EnableFallbackCertificate defines if the vhost should allow a default certificate to be applied which handles all requests which don't match the SNI defined in this vhost.
                         type: boolean
                       minimumProtocolVersion:
-                        description: Minimum TLS version this vhost should negotiate
+                        description: MinimumProtocolVersion is the minimum TLS version this vhost should negotiate. Valid options are `1.2` (default) and `1.3`. Any other value defaults to TLS 1.2.
                         type: string
                       passthrough:
                         description: Passthrough defines whether the encrypted TLS handshake will be passed through to the backing cluster. Either Passthrough or SecretName must be specified, but not both.
--- a/examples/render/contour.yaml
+++ b/examples/render/contour.yaml
@@ -55,7 +55,7 @@ data:
     disablePermitInsecure: false
     tls:
     # minimum TLS version that Contour will negotiate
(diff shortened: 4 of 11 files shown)


sunjayBhatia (commented):
Not documentation, but want to get rid of these in this PR too just for cleanup/so we don't have any `1.1` references left to confuse us?
- https://github.com/projectcontour/contour/blob/78c434fcd9aa8e08f12ee5def7c0e215c0c805c5/internal/envoy/v3/tls.go#L22
- https://github.com/projectcontour/contour/blob/78c434fcd9aa8e08f12ee5def7c0e215c0c805c5/internal/annotation/annotations.go#L205
- https://github.com/projectcontour/contour/blob/78c434fcd9aa8e08f12ee5def7c0e215c0c805c5/internal/envoy/v3/socket_test.go#L69-L73
- https://github.com/projectcontour/contour/blob/78c434fcd9aa8e08f12ee5def7c0e215c0c805c5/internal/envoy/v3/listener_test.go#L199

This one maybe is good to keep b/c the test that refers to it ensures upgrade from 1.1 -> 1.2?
- https://github.com/projectcontour/contour/blob/78c434fcd9aa8e08f12ee5def7c0e215c0c805c5/internal/featuretests/v3/listeners_test.go#L1029

skriss:
> Not documentation, but want to get rid of these in this PR too just for cleanup/so we don't have any `1.1` references left to confuse us?
> 
> * https://github.com/projectcontour/contour/blob/78c434fcd9aa8e08f12ee5def7c0e215c0c805c5/internal/envoy/v3/tls.go#L22
> * https://github.com/projectcontour/contour/blob/78c434fcd9aa8e08f12ee5def7c0e215c0c805c5/internal/annotation/annotations.go#L205
> * https://github.com/projectcontour/contour/blob/78c434fcd9aa8e08f12ee5def7c0e215c0c805c5/internal/envoy/v3/socket_test.go#L69-L73
> * https://github.com/projectcontour/contour/blob/78c434fcd9aa8e08f12ee5def7c0e215c0c805c5/internal/envoy/v3/listener_test.go#L199
> 
> This one maybe is good to keep b/c the test that refers to it ensures upgrade from 1.1 -> 1.2?
> 
> * https://github.com/projectcontour/contour/blob/78c434fcd9aa8e08f12ee5def7c0e215c0c805c5/internal/featuretests/v3/listeners_test.go#L1029

Yep good call. I'll adjust the commit message a bit.

sunjayBhatia (approved):

apis/projectcontour/v1/httpproxy.go
```diff
@@ -216,7 +216,9 @@ type TLS struct {
 	// If specified, the named secret must contain a matching certificate
 	// for the virtual host's FQDN.
 	SecretName string `json:"secretName,omitempty"`
-	// Minimum TLS version this vhost should negotiate
+	// MinimumProtocolVersion is the minimum TLS version this vhost should
+	// negotiate. Valid options are `1.2` (default) and `1.3`. Any other value
+	// defaults to TLS 1.2.
```
stevesloka, line 221: > Any other value defaults to TLS 1.2.

Do we get a warning or anything if this case is hit?

skriss: Nope. We can change that behavior if desired; I just left it the way it already was but not sure if there was a reason for it or not.

youngnick: This is documenting the current behavior, I don't believe there's any reason for people to run the massively insecure TLS 1.1, if there is, they can come and ask.

youngnick (approved):
LGTM, I added the release note labels for this one.

apis/projectcontour/v1/httpproxy.go
stevesloka: @youngnick no my point is if someone sets an invalid option, say they make the value `99.99`, Contour will change that to `1.2`. I'm cool with it in this PR since that's how it works, but there's no indication to a user, that they 1. set an invalid value and 2. Contour saw that and changed it to the default.

stevesloka (approved):

Issue #2146, opened 2020-01-23
Ingress V1: support IngressClass


jpeach:
Ingress V1 changes the IngressClass string to a document.

https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20190125-ingress-api-group.md#ingress-class

> Promoting the annotation as it is currently defined as an opaque string is the most direct path but precludes any future enhancements to the concept.
>
> An alternative is to create a new resource IngressClass to take its place. This resource will serve a couple of purposes:
> 
> Define the set of valid classes available to the user. Gives operators control over allowed classes.
> Allow us to evolve the API to express concepts such a levels of service associated with a given Ingress controller.

https://github.com/cmluciano/kubernetes/pull/3

stevesloka:
We need to add a flag I think to buy into this appropriately before making this standard. 

I'm proposing an `--enable-ingress-v1` feature flag that puts Contour into a mode that enables the v1 features and changes the default behavior of how Ingress Classes are handled (e.g. annotation vs spec.class).

Today Contour matches on the following rules:
-  If no ingress.class is specified on Contour args
    - Matches ingress.class of `contour` or no annotation
- If ingress.class is specified on Contour args
  - Matches only ingress.class equal to the one specified in args, any other object is ignored if not matching or the annotation is not supplied

The new Ingress v1 adds a `ingressclassname` field on the Ingress spec which references an IngressClass object. 

The docs say that the annotation should take precedence over the spec.ingressclassname. If spec.ingressclassname does not match an IngressClass object then no match, however, we can have an `IngressClass` object that is the "default" by using the annotation:  `ingressclass.kubernetes.io/is-default-class: "true"`.

What we need is a way to know when we're going to enforce the ingressv1 behavior by requiring ingressclasses.

stevesloka:
The other way to think about this is if not using annotations, we could just require a default ingressclass to be created. I think that makes sense, but I want to chat about it further first.

stevesloka:
Here are the rules I've come up with which I think covers all the cases we could run into:

If `--ingress-class-name` is **not** set:
- Matches any ingress object without any annotation
- Matches any ingress object with annotation the default ingress.class of `contour`
- Matches any ingress object, if a `v1beta1.IngressClass` exists with the name of `contour` and the ingress object has the `spec.ingressclassname: contour`
[...]

Issue #2135, opened 2020-01-22
Ingress V1: path matching mode


jpeach:
Support Ingress V1 path changes:

https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20190125-ingress-api-group.md#path-as-a-prefix

> Paths proposal

> Explicitly state the match mode of the path.
> Support the existing implementation-specific behavior.
> Support a portable prefix match and future expansion of behavior.
> Add a field ingress.spec.rules.http.paths.pathType to indicate the desired interpretation of the meaning of the path

jpeach:
https://github.com/cmluciano/kubernetes/pull/2

sunjayBhatia:
So we already support regex path matching on Ingress resources but the current implementation is such that the path match type is ignored if you put something in the path field that "looks like" a regex.

Seems like this regex path match deal is better suited for just the `ImplementationSpecific` path match type and should be documented as such, however that would be a breaking change.

So moving forward how does this sound:
- If you provide a path that "looks like" a regex, you will get a regex match in practice, no matter what you set on the path type
  - Another option would be to honor the path type set and only do a regex match if you set the type to implementation specific
  - We could also do the "looks like regex" check if you don't specify a path type
- If you do not specify a path match type, you get a prefix match
- If you specify a prefix path match type, you get a prefix match
- If you specify an exact path match type, you get an exact match
- If you specify an implementation specific path match type, you get a regex match

youngnick:
We could also put a changelog note and change it, although that would also not be ideal. I anticipate that the Ingress code is going to be around a *long* time though, so we should make it as tidy as posisble.

If we don't want to do that, @sunjayBhatia's solution above seems like the only way.

sunjayBhatia:
Another weird caveat that the spec wants: https://github.com/kubernetes-sigs/ingress-controller-conformance/blob/79bd068cbb31d77c2a060f61eea795f9791a3d48/features/path_rules.feature#L162

Prefix matches are not string prefix matches, but path label matches

See also https://kubernetes.io/docs/concepts/services-networking/ingress/#examples

```

Kind | Path(s) | Request path(s) | Matches?
...
Prefix | /aaa/bbb | /aaa/bbbxyz | No, does not match string prefix
```

youngnick:
We specifically had to stop doing prefix matches using Go's Path support because of weird interactions with inclusion for HTTPProxy, but if we're going to meet the spec, maybe the Ingress validation code can just treat them as path labels.

[...]

PR #3266, opened 2021-01-20
internal: Add support for watching ingress/v1 resources

stevesloka:
Add support for watching networking.k8s.io/v1 Ingress resources and add/remove
from internal Cache later allowing these objects to be implemented.

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -373,6 +373,20 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		inf.AddEventHandler(&dynamicHandler)
 	}
 
+	// If Ingress v1 resource exist, then add informers to watch, otherwise
+	// add Ingress v1beta1 informers.
+	if clients.ResourcesExist(k8s.IngressV1Resources()...) {
+		for _, r := range k8s.IngressV1Resources() {
+			if err := informOnResource(clients, r, &dynamicHandler); err != nil {
+				log.WithError(err).WithField("resource", r).Fatal("failed to create informer")
+			}
+		}
+	} else {
+		if err := informOnResource(clients, k8s.IngressV1Beta1Resource(), &dynamicHandler); err != nil {
+			log.WithError(err).WithField("resource", k8s.IngressV1Beta1Resource()).Fatal("failed to create informer")
+		}
+	}
+
 	// Inform on service-apis types if they are present.
 	if ctx.UseExperimentalServiceAPITypes {
 		for _, r := range k8s.ServiceAPIResources() {
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -23,6 +23,7 @@ import (
 	"github.com/projectcontour/contour/internal/timeout"
 	"github.com/stretchr/testify/assert"
 	v1 "k8s.io/api/core/v1"
+	networking_v1 "k8s.io/api/networking/v1"
 	"k8s.io/api/networking/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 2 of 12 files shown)


stevesloka:
Please hold on this for a second, I'm going to add IngressClass as well which is needed along with ingress/v1.

internal/k8s/informers.go
```diff
@@ -38,6 +39,7 @@ func DefaultResources() []schema.GroupVersionResource {
 		contour_api_v1alpha1.ExtensionServiceGVR,
 		corev1.SchemeGroupVersion.WithResource("services"),
 		v1beta1.SchemeGroupVersion.WithResource("ingresses"),
+		networking_v1.SchemeGroupVersion.WithResource("ingresses"),
```
skriss, line 42: `v1` didn't exist until 1.19, so this will effectively bump our min supported Kube version to 1.19, won't it? https://github.com/projectcontour/contour/blob/main/cmd/contour/serve.go#L368 will `Fatal` if it doesn't exist.

Also, since all `Ingresses` will simultaneously be represented as `v1beta1` and `v1` Ingresses (assuming a 1.19-1.22 cluster that has both API versions), how do we avoid processing them twice?

stevesloka: I thought about the version not being there working on IngressClasses, we can add the logic to check if the resource exists like we did for service-apis. I think we can use that to also check if v1 resources exist and do logic on that. I have new commit here in a second to show that.

skriss: Yeah I think we'll need to add the existence check, will probably have to do the same eventually for `v1beta1` too since it'll be dropped in 1.22, if we want to support both 1.18 and 1.22 simultaneously.

stevesloka:
Integration tests are failing because of RBAC. =)

cmd/contour/serve.go
```diff
@@ -371,6 +371,27 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+	// add Ingress v1beta1 informers.
+	if clients.ResourcesExist(k8s.IngressV1Resources()...) {
+		for _, r := range k8s.IngressV1Resources() {
+			inf, err := clients.InformerForResource(r)
+			if err != nil {
+				log.WithError(err).WithField("resource", r).Fatal("failed to create informer")
+			}
+
+			inf.AddEventHandler(&dynamicHandler)
+		}
```
skriss, line 384: nit: could use `informOnResource(...)` here.

cmd/contour/serve.go
```diff
@@ -371,6 +371,27 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+
+	} else {
+		inf, err := clients.InformerForResource(k8s.IngressV1Beta1Resource())
+		if err != nil {
+			log.WithError(err).WithField("resource", k8s.IngressV1Beta1Resource()).Fatal("failed to create informer")
+		}
+
+		inf.AddEventHandler(&dynamicHandler)
```
skriss, line 392: nit: could use `informOnResource(...)` here.

internal/dag/cache.go
```diff
@@ -46,7 +47,9 @@ type KubernetesCache struct {
 	// Secrets that are referred from the configuration file.
 	ConfiguredSecretRefs []*types.NamespacedName
 
-	ingresses            map[types.NamespacedName]*v1beta1.Ingress
+	ingressesv1beta1     map[types.NamespacedName]*v1beta1.Ingress
+	ingressesv1          map[types.NamespacedName]*networking_v1.Ingress
+	ingressclasses       map[string]*networking_v1.IngressClass
```
skriss, line 52: do we need to cache `IngressClass`?

internal/dag/cache.go
```diff
@@ -288,7 +310,7 @@ func (kc *KubernetesCache) remove(obj interface{}) bool {
 // serviceTriggersRebuild returns true if this service is referenced
 // by an Ingress or HTTPProxy in this cache.
 func (kc *KubernetesCache) serviceTriggersRebuild(service *v1.Service) bool {
-	for _, ingress := range kc.ingresses {
+	for _, ingress := range kc.ingressesv1beta1 {
```
skriss, line 313: I think we're going to need to check both `kc.ingresses` and `kc.ingressesv1beta1`, since either one of them could be populated based on whether `v1` exists in the cluster or not.

internal/dag/cache.go
```diff
@@ -358,7 +380,7 @@ func (kc *KubernetesCache) secretTriggersRebuild(secret *v1.Secret) bool {
 		}
 	}
 
-	for _, ingress := range kc.ingresses {
+	for _, ingress := range kc.ingressesv1beta1 {
```
skriss, line 383: I think we're going to need to check both `kc.ingresses` and `kc.ingressesv1beta1`, since either one of them could be populated based on whether `v1` exists in the cluster or not.

internal/dag/ingress_processor.go
```diff
@@ -58,7 +58,7 @@ func (p *IngressProcessor) Run(dag *DAG, source *KubernetesCache) {
 // computeSecureVirtualhosts populates tls parameters of
 // secure virtual hosts.
 func (p *IngressProcessor) computeSecureVirtualhosts() {
-	for _, ing := range p.source.ingresses {
+	for _, ing := range p.source.ingressesv1beta1 {
```
skriss, line 61: I think we're going to need to check both `kc.ingresses` and `kc.ingressesv1beta1`, since either one of them could be populated based on whether `v1` exists in the cluster or not.

internal/dag/ingress_processor.go
```diff
@@ -95,7 +95,7 @@ func (p *IngressProcessor) computeSecureVirtualhosts() {
 
 func (p *IngressProcessor) computeIngresses() {
 	// deconstruct each ingress into routes and virtualhost entries
-	for _, ing := range p.source.ingresses {
+	for _, ing := range p.source.ingressesv1beta1 {
```
skriss, line 98: I think we're going to need to check both `kc.ingresses` and `kc.ingressesv1beta1`, since either one of them could be populated based on whether `v1` exists in the cluster or not.

stevesloka:
Ok turns out testing against a kind v1.16.4 cluster (image: `kindest/node:v1.16.4@sha256:b91a2c2317a000f3a783489dfb755064177dbc3a0b2f4147d50f04825d016f55`) doesn't let Envoy connect to Contour, but using a v1.16.9 image (`kindest/node:v1.16.9@sha256:7175872357bc85847ec4b1aba46ed1d12fa054c83ac7a8a11f5c268957fd5765`) does. 

I was down a rabbit hole chasing the wrong issue. =)

internal/dag/cache.go
youngnick: I think it's better to, I don't think there'll be that many objects, and it means we don't have to restart informers if ingressclasses change.

youngnick (commented):
This is looking good! I like explicitly translating v1beta1 Ingress objects into v1 objects internally, that makes things easier.

stevesloka:
@skriss I updated with your comments in the initial review. The method now is I have a converter which turns all v1beta1 ingress objects into v1 ingress types. From that point, all the processing is the same.

internal/dag/cache.go
```diff
@@ -64,7 +66,8 @@ type KubernetesCache struct {
 
 // init creates the internal cache storage. It is called implicitly from the public API.
 func (kc *KubernetesCache) init() {
-	kc.ingresses = make(map[types.NamespacedName]*v1beta1.Ingress)
+	kc.ingresses = make(map[types.NamespacedName]*networking_v1.Ingress)
+	kc.ingressclasses = make(map[string]*networking_v1.IngressClass)
```
skriss, line 70: it's still not clear to me that we need to cache `IngressClass`, but we can always remove it later if we don't need it.

internal/dag/ingress_processor.go
```diff
@@ -165,7 +175,7 @@ func (p *IngressProcessor) computeIngressRule(ing *v1beta1.Ingress, rule v1beta1
 }
 
 // route builds a dag.Route for the supplied Ingress.
-func route(ingress *v1beta1.Ingress, path string, service *Service, clientCertSecret *Secret, log logrus.FieldLogger) (*Route, error) {
+func route(ingress *networking_v1.Ingress, path string, service *Service, clientCertSecret *Secret, log logrus.FieldLogger) (*Route, error) {
```
skriss, line 178: Wondering what we should do with the [pathType](https://kubernetes.io/docs/concepts/services-networking/ingress/#path-types) field for now. It looks like the main issue would be if a user has explicitly specified `pathType: Exact`, Contour would currently still treat it as a prefix. We have https://github.com/projectcontour/contour/issues/2135 to implement that properly; perhaps until then we should return an error if a `pathType: Exact` is encountered?

internal/dag/cache.go
```diff
@@ -200,6 +214,82 @@ func (kc *KubernetesCache) Insert(obj interface{}) bool {
 	return false
 }
 
+func toV1Ingress(obj *v1beta1.Ingress) *networking_v1.Ingress {
```
skriss, line 217: I'll do a more detailed look at this but I need more ☕ first

skriss (commented):
@stevesloka I think this is looking pretty good. Looks like it'll address a few of the sub-issues listed in https://github.com/projectcontour/contour/issues/2135#issue-553296018, and then we'll likely have to follow up on a few more.

internal/dag/cache.go
```diff
@@ -200,6 +214,82 @@ func (kc *KubernetesCache) Insert(obj interface{}) bool {
...
+							Port: serviceBackendPort(p.Backend.ServicePort),
+						},
+					},
+				})
+			}
+		}
+
+		rule := networking_v1.IngressRule{
+			Host: r.Host,
+		}
```
skriss, line 249: you could move this above L233, allowing the subsequent `if r.HTTP != nil` to be combined with the one on L233.

internal/dag/cache.go
```diff
@@ -200,6 +214,82 @@ func (kc *KubernetesCache) Insert(obj interface{}) bool {
...
+		if r.HTTP != nil {
```
skriss, line 250: you can move the body of this `if` into the body of the `if` that starts on L233

internal/dag/cache.go
```diff
@@ -200,6 +214,82 @@ func (kc *KubernetesCache) Insert(obj interface{}) bool {
...
+		})
+	}
+
+	for _, r := range obj.Spec.Rules {
+
+		if r.HTTP != nil {
+			for _, p := range r.HTTP.Paths {
+				paths = append(paths, networking_v1.HTTPIngressPath{
+					Path: p.Path,
+					Backend: networking_v1.IngressBackend{
```
skriss, line 237: should assign `PathType` from the v1beta1 obj as well, though I guess this can wait until we implement #2135

skriss (changes requested):
couple more comments on the converter function

internal/dag/cache.go
```diff
@@ -200,6 +214,82 @@ func (kc *KubernetesCache) Insert(obj interface{}) bool {
...
+			Service: &networking_v1.IngressServiceBackend{
+				Name: obj.Spec.Backend.ServiceName,
+				Port: serviceBackendPort(obj.Spec.Backend.ServicePort),
+			},
+		}
+	}
+
+	converted := &networking_v1.Ingress{
+		ObjectMeta: obj.ObjectMeta,
+		Spec: networking_v1.IngressSpec{
```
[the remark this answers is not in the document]
stevesloka: We can do that, but wasn't trying to implement it all in this PR.

internal/dag/cache.go
stevesloka: 👍

internal/dag/cache.go
stevesloka: 👍

internal/dag/cache.go
stevesloka: setup pathtype just to make the converter complete

internal/dag/ingress_processor.go
skriss: per discussion, Contour is already ignoring this field and treating everything as either a regex or a prefix, so it's OK to continue doing that until we implement #2135.

internal/dag/cache.go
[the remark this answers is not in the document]
skriss: OK to implement in #2146

internal/dag/builder_test.go
```diff
@@ -104,154 +105,185 @@ func TestDAGInsert(t *testing.T) {
 		},
 	}
 
-	i1 := &v1beta1.Ingress{
+	i1V1 := &networking_v1.Ingress{
```
skriss, line 108: this file's diff is pretty impossible to review - can you describe the changes? I think you said you were creating a second copy of all the Ingress tests, that used a `v1` Ingress instead of a `v1beta1`, so we'd have tests for both the `v1beta1` and the `v1` representations.

stevesloka: Yeah it's unfortunate that the diff is so hard to read, but I copied all the v1beta1 ingress resources into networking_v1 ingress resources and corresponding tests so that we could test the v1beta1 resources along with the networking_v1 resources.

In the next set of PRs, we should expand these tests with the new features in v1 (e.g. ingressclass, path type, etc)

skriss (commented):
This looks pretty good to me. Do you want to hold it until after we release 1.12, so we can get a few extra test cycles in on it (along with the additional items for full v1 support)?

stevesloka:
> This looks pretty good to me. Do you want to hold it until after we release 1.12, so we can get a few extra test cycles in on it (along with the additional items for full v1 support)?

Yup we can hold, this PR doesn't add anything to the release at the moment.

internal/dag/cache.go
```diff
@@ -230,6 +336,15 @@ func (kc *KubernetesCache) remove(obj interface{}) bool {
 		_, ok := kc.ingresses[m]
 		delete(kc.ingresses, m)
 		return ok
+	case *networking_v1.Ingress:
```
sunjayBhatia, line 339: Could we also add coverage on the removal of the newly watched resources in `TestKubernetesCacheRemove`?

sunjayBhatia (changes requested):

sunjayBhatia:
> This looks pretty good to me. Do you want to hold it until after we release 1.12, so we can get a few extra test cycles in on it (along with the additional items for full v1 support)?

+1 just a small comment

internal/dag/cache.go
stevesloka: yup, added a test

youngnick (approved):
This LGTM, but agree that we should hold until after 1.12 releases.

skriss:
ready for a rebase & merge

internal/dag/cache.go
skriss: It doesn't look like the `IngressClass` cache is actually used anywhere. We can follow up and remove it if it's not needed for any of the v1 functionality.

skriss:
Needs a `make generate`.

stevesloka:
> Needs a make generate.

Yup! Just did that and pushed, should work now.

sunjayBhatia (approved):
~~I might be missing where the logic for this is, but do we have coverage for when an IngressClass object is removed? Should we be removing all the Ingress objects that reference that IngressClass from the cache etc.?~~

I was initially going to say the above but read through https://github.com/projectcontour/contour/issues/2146 before clicking submit. Given we agreed to add all Ingress objects to the cache and only add them to the DAG if they apply, this LGTM

stevesloka:
I'm trying to make smaller commits for this stuff so it's easier to read. Now that we have the plumbing to get the objects, we can now operate against them. Thanks for the reviews @skriss & @sunjayBhatia! 🎉

stevesloka:
@skriss are you good with the converter bits? There were changes requested so wanted to double-check.

skriss (approved):
LGTM!

Issue #3099, opened 2020-11-06
Support for header hash based load balancer policy


sonawaneshivraj:
We have a requirement to load balance gRPC streams by performing a consistent hash over a header value. 

Clients send messages (gRPC) to multiple instances of a service, the message contains a well-known header value (entity_id), the messages for a given entity_id needs to be processed by the same service instance.

With Envoy, this can be achieved via [header hash policy](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-routeaction-hashpolicy-header). Is there a way to achieve similar behaviour with Contour?

I don't see this as a currently supported [LoadBalancerPolicy](https://projectcontour.io/docs/v1.9.0/api/#projectcontour.io/v1.LoadBalancerPolicy), is this on the roadmap?

I understand this might be a very specific request and I am happy to give out more information if required.

jpeach:
xref #3044

youngnick:
Sorry about the delay in getting back to you @sonawaneshivraj, as @jpeach mentioned, I'd encourage you to take a look at #3044 to see if that level of flexibility would provide what you need (I suspect it would).

In an ideal world, how would you want to configure this?

sonawaneshivraj:
@youngnick, not sure #3044 covers our requirement. I believe #3044 talks in general about a  richer set of load-balancing functionality and the ability to set additional rules. It does not specifically mention the support for header hash-based load balancing. 

I will get back with the details of how we foresee the configuration for this.

youngnick:
If we implement #3044, then as you can see in the examples, that means that you would be able to specify a header hash policy, amongst the rest. In terms of configuration, I don't mean that I need a fully specced out YAML, I meant - what fields would you like to see? Presumably you would need a field to configure which header to use, but would you want anything else?

youngnick:
After having a conversation about this, I think we came up with the following requirements:
- the gRPC client will be supply the required header as part of the initial request (this means that we don't need Envoy to generate a header value)
- if requests come in without that initial request, then the request will be randomly assigned to a backend.

I *believe* we should be able to do this with a relatively simple hash_policy, so it's probably okay to add to our set of shorcuts (similarly to how we have `Cookiee` load balancing now, which is a shortcut for a certain hash_policy).

[...]

Issue #3044, opened 2020-10-20
Support list of loadBalancerPolicy for HTTPProxy


wildermesser:
Currently, I can use only one policy from:
 - RoundRobin
 - WeightedLeastRequest
 - Random
 - Cookie
But envoy supports a list of parameters for hash_policy. And the cookie is only one of them. With cookie `loadBalancerPolicy`, I have this envoy's config:
```
                    "route": {      
                      "cluster": "cluster_name",
                      "hash_policy": [
                        {     
                          "cookie": {                                         
                            "name": "X-Contour-Session-Affinity",
                            "ttl": "0s",
                            "path": "/"
                          }
                        }
                      ]                                                       
                    }
```
When a request does not contain this cookie, it will be routed to a random backend server. But I want to set additional rules like:
```
                    "route": {      
                      "cluster": "cluster_name",
                      "hash_policy": [
                        {     
                          "cookie": {                                         
                            "name": "X-Contour-Session-Affinity",
                            "ttl": "0s",
                            "path": "/"
                          }
                        },
                        "terminal": "true",
                        {     
                          "header": {                                         
                            "header_name": "User-Agent",
                          }
                         },
                          {
                          "connection_properties": {                                         
                            "source_ip": "true",
                          }
                        }
                      ]                                                       
                    }
```
It would be cool when Contour can support this rich load balancing functionality.
I understand that my case may be very specific. What is the simplest way to change runtime envoy configuration which will not changes when changing pod's list?
Should I use something like https://github.com/fullstorydev/grpcurl for changing the configuration of envoy instance on the fly? Or contour has some kind of configuration template?

jpeach:
Assign to Nick to review.

youngnick:
Thanks for logging this issue @wildermesser. I think there are a few questions in here, I'll do my best to answer as I can.

For your example, it looks like the more complex functionality you want to use allows specifying what the hash that goes into the cookie will look like - in this case, based on the user-agent and source IP. I can see how that would be useful, I hadn't taken the time to fully parse how this hash_policy works before - I thought you could only have one at a time. It does seem like it would be useful to be able to do things like this, and I can see we also have #2965 which describes a similar request. I have absolutely no idea what the way to configure this would look like though, this would almost certainly need a full design document, to figure out how to fit it into HTTPProxy. We're also constrained by the fact that HTTPProxy is a GA API, so we can't change the meaning of existing fields in incompatible ways.

[...]

PR #3268, opened 2021-01-22
design: Header Hash based (and generic extension for other methods of) load balancing

sunjayBhatia:
Updates: https://github.com/projectcontour/contour/issues/3099
Updates: https://github.com/projectcontour/contour/issues/3044


design/loadbalancer-hash-policy-design.md
```diff
@@ -0,0 +1,185 @@
...
+	Strategy string `json:"strategy,omitempty"`
+
+	RequestHashPolicies []RequestHashPolicy `json:"requestHashPolicies,omitempty"`
+}
+```
+
+This field will be a list of `RequestHashPolicy` objects with each element holding configuration for an individual hash policy.
+Each list element will allow users configuring a `Route` on a `HTTPProxy` to specify a separate request attribute for Envoy to hash in order to make an upstream load balancing decision.
+`RequestHashPolicy` contains a field `RequestAttribute` specifying which type of request attribute it is targeting.
+If `RequestAttribute` is empty or an unknown value, this hash policy entry will be ignored.
```
skriss, line 60: what are the valid values? Just `Header` for now?

design/loadbalancer-hash-policy-design.md
```diff
@@ -0,0 +1,185 @@
...
+### Handling Invalid Strategy Choice
+As per the documentation of the [`LoadBalancerPolicy` field on the `ExtensionService` object](https://projectcontour.io/docs/v1.11.0/config/api/#projectcontour.io/v1alpha1.ExtensionService), the `Cookie` load balancing strategy is invalid for use with the `ExtensionService`.
+Currently, that restriction does not seem to be enforced (or on the `TCPRoute` object either, for which the `Cookie` strategy is not compatible).
+This design proposes that we start to enforce these strategy restrictions and also restrict the new `RequestHash` strategy to only valid on `Route` objects.
+If specified on `ExtensionService` or `TCPRoute` objects, it will be overridden to the default `RoundRobin` strategy and a warning will be logged.
+Overriding rather than throwing an error is a deliberate attempt to continue to allow the configured `HTTPProxy` or configured extension to function, albeit with basic load balancing enabled.
+
+## Alternatives Considered
+
+### Per-Hash Policy Structs
```
skriss, line 117: I would definitely consider choosing this alternative over using a `map[string]string`. I'm doing something similar for global rate limiting, where there are a number of different ways of generating a rate limit descriptor, and each one has its own set of configuration, and I am planning to have a strongly-typed config struct for each. So choosing this alternative would be consistent with the rate limiting approach. Also, I worry about if there's ever a case where a hash policy requires nested config that doesn't fit nicely into a `map[string]string`. And, like you said, the possible set of these is bounded, it shouldn't be too much of a maintenance burden to add new structs as needed.

design/loadbalancer-hash-policy-design.md
```diff
@@ -0,0 +1,185 @@
...
+
+## Goals
+- Contour should offer the ability to load balance requests based on HTTP user specified request headers
+- The design should offer the ability to configure multiple hash policies (of which specific new types may be implemented separately from this design)
+
+## Non Goals
+- Provide advanced load balancing configuration for `ExtensionService` or `TCPRoute`
+- Provide users direct access to Envoy configuration fields/structures
+
+## High-Level Design
```
skriss, line 37: Based on background, I assume the plan is to support the `RingHash` load balancer? Do we want to support `Maglev` as well, and if so, how would we do that?

skriss (commented):
This looks great! Couple questions/comments.

design/loadbalancer-hash-policy-design.md
```diff
@@ -0,0 +1,185 @@
...
+      - requestAttribute: Header
+        hashOptions:
+          headerName: X-Some-Header
+        terminal: true
+      - requestAttribute: Header
+        hashOptions:
+          headerName: User-Agent
+```
+
+The value of the `X-Some-Header` header will be hashed by Envoy if present and used to make a load balancing decision.
```
xaleeks, line 104: How do I guarantee consistency in sending traffic to the same pod based on the uniqueness of a pair or tuple of headers?  Do I just list the headers one by one, do I need to put terminal:true for the last one?  I don't understand how terminal works

design/loadbalancer-hash-policy-design.md
```diff
@@ -0,0 +1,185 @@
...
+If the header is present, Envoy will not move on to attempting to hash the `User-Agent` header value.
+Consistent values in this header would lead to Envoy routing to the same backend service instance.
+
```
xaleeks, line 107: also, I guess there's no way to specify a set of values to match on right, since this is on envoy side.  As in 'abc' and 'def' for a specified header both should hash to the same pod.

design/loadbalancer-hash-policy-design.md
youngnick: `Maglev` is a drop-in replacement for `RingHash` as far as the Envoy config is concerned. So I think we can kick that can down the road.

design/loadbalancer-hash-policy-design.md
youngnick: The last entry has and implied `terminal: true`. `terminal` is used as a short-circuit for further processing if matched.

design/loadbalancer-hash-policy-design.md
youngnick: Based on my understanding of the `hash_policy` docs, that's right.

design/loadbalancer-hash-policy-design.md
youngnick: The nice property of the `map[string]string` is that it allows us to implement the more general "list of hash policies" setup (that is, #3044 ), by adding more options for the `RequestAttribute` field, which all require their own set of options. As you say, there is a bounded set, but I don't think there's also a lot of configuration options for each hash policy.

It does have the disadvantage that we can't do input validation of the fields using the CRD validation bits though, which is a nice user experience.

I think there's enough tradeoffs either way that I don't really have a strong opinion which we pick.

youngnick (approved):
Great design @sunjayBhatia, looks excellent. I think the only real outstanding question is if we can all agree that a `map[string]string` will do for parameter storage. I can see doing either, so I'm happy to leave it to people with stronger opinions to make the call.

design/loadbalancer-hash-policy-design.md
sunjayBhatia: just `Header` for now, could be expanded to `Cookie` etc. from here: https://www.envoyproxy.io/docs/envoy/v1.17.0/api-v3/config/route/v3/route_components.proto#config-route-v3-routeaction-hashpolicy

design/loadbalancer-hash-policy-design.md
sunjayBhatia: I chose the `map[string]string` mostly because of recent chats we've had and misgivings around proliferating too many types, structs, etc., but would probably choose the struct option without that, will switch it around to that

design/loadbalancer-hash-policy-design.md
sunjayBhatia: Yeah, could be another "strategy" to use MagLev, instead of Ring Hash, could lead us to change the name of the strategy proposed here or just add a new one in the future if people ask for MagLev specifically

design/loadbalancer-hash-policy-design.md
sunjayBhatia: heres an example from the Envoy docs: https://www.envoyproxy.io/docs/envoy/v1.17.0/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routeaction-hashpolicy-terminal

I can also link this in the design doc

design/loadbalancer-hash-policy-design.md
sunjayBhatia: There is this option: https://www.envoyproxy.io/docs/envoy/v1.17.0/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routeaction-hashpolicy-header-regex-rewrite

But no one has asked for that yet, though I think that would do what youre asking: you would write a regex that matches two header values and the replacement value would be the same for both, would hash to the same thing etc.

design/loadbalancer-hash-policy-design.md
skriss: Let's see what @stevesloka thinks. I think either one's workable, though I do lean towards the structs.

design/loadbalancer-hash-policy-design.md
sunjayBhatia: Added clarification in https://github.com/projectcontour/contour/pull/3268/commits/72a81ca93c84d002f309ba4b65d6e1188e40c4cb

design/loadbalancer-hash-policy-design.md
```diff
@@ -0,0 +1,188 @@
...
+Consistent values in this header would lead to Envoy routing to the same backend service instance.
+If the header is present, Envoy will *not* move on to attempting to hash the `User-Agent` header value.
+
+### Handling Invalid Strategy Choice
+As per the documentation of the [`LoadBalancerPolicy` field on the `ExtensionService` object](https://projectcontour.io/docs/v1.11.0/config/api/#projectcontour.io/v1alpha1.ExtensionService), the `Cookie` load balancing strategy is invalid for use with the `ExtensionService`.
+Currently, that restriction does not seem to be enforced (or on the `TCPRoute` object either, for which the `Cookie` strategy is not compatible).
+This design proposes that we start to enforce these strategy restrictions and also restrict the new `RequestHash` strategy to only valid on `Route` objects.
+If specified on `ExtensionService` or `TCPRoute` objects, it will be overridden to the default `RoundRobin` strategy and a warning will be logged.
+Overriding rather than throwing an error is a deliberate attempt to continue to allow the configured `HTTPProxy` or configured extension to function, albeit with basic load balancing enabled.
```
stevesloka, line 116: Set a warning on the object telling the user this is happening?

design/loadbalancer-hash-policy-design.md
stevesloka: I typically like structs since it's more clear sometimes as to what's going on and can validate better.  If the data lends itself to a struct then 👍  from me.

design/loadbalancer-hash-policy-design.md
sunjayBhatia: whoops I thought i added a sentence about that! will add

sunjayBhatia: WDYT about adding a warning on the object if there are other lower level configuration errors as well? Realized I forgot to add that as well

design/loadbalancer-hash-policy-design.md
sunjayBhatia: See https://github.com/projectcontour/contour/pull/3268/commits/4852bc08ee0b567204dba03a7a24a6f7ccd9861e

design/loadbalancer-hash-policy-design.md
sunjayBhatia: See https://github.com/projectcontour/contour/pull/3268/commits/4852bc08ee0b567204dba03a7a24a6f7ccd9861e

design/loadbalancer-hash-policy-design.md
skriss: Yeah I don't think we need to add it now. Should be fine to add `Maglev` as a strategy in the future.

skriss (approved):
LGTM, nice work @sunjayBhatia!

youngnick (approved):
Approving again, I really like how you've included both what we chose for the parameters and why we chose it, excellent. That means we can come back here later, when we've all forgotten the context. :)

design/loadbalancer-hash-policy-design.md
xaleeks: perfect, thanks guys

design/loadbalancer-hash-policy-design.md
xaleeks: Thanks.  yea I was thinking sometimes you only have a referential identity, not a canonical one, as in I hold 'abc' and 'edf' to be equivalent and that's all that matters.  I guess you can always engineer a regex to encapsulate those two things

Issue #3258, opened 2021-01-14
Enable setting headers on all routes with dynamic service values


erwbgy:
Using Linkerd with Contour requires setting the `l5d-dst-override` header on every request to the destination service hostname and port - for example: `kuard.default.svc.cluster.local:80`.

I agree with @jpeach's comment on #2089 that Contour shouldn't have special support for Linkerd headers - like Ambassador's _add_linkerd_headers_ option - but I think it should be possible to support it more generically including @stevesloka's idea of adding default headers.

My proposal is to enable request and response headers specified in the config file to be set on all routes with dynamic values for namespace, service name and service port:

1) Add support for %CONTOUR_NAMESPACE%, %CONTOUR_SERVICE_NAME% and %CONTOUR_SERVICE_PORT% dynamic variables and have these expanded like Envoy dynamic variables are in #3234.  (The CONTOUR_ prefix is used so that there is no possibility of a clash with a future Envoy dynamic variable.)

This would allow a header to be set in an HTTPProxy object:
```
      requestHeadersPolicy:
        set:
        - name: X-Envoy-Hostname
          value: "%HOSTNAME%"
        - name: l5d-dst-override
          value: "%CONTOUR_SERVICE_NAME%.%CONTOUR_NAMESPACE%.svc.cluster.local:%CONTOUR_SERVICE_PORT%"
```
If any of these variables are used where they can't be expanded then they are just passed through literally. For example if %CONTOUR_SERVICE_NAME% is used at the route level then it is not expanded and becomes %%CONTOUR_SERVICE_NAME%% when passed to Envoy. This avoids any Envoy configuration errors due to bad dynamic variables.

2) Allow default request and response headers added to every route to be set in the Contour config file:
```
headers:
  set-request-headers:
  - "X-Envoy-Hostname=%HOSTNAME%"
  - "l5d-dst-override=%CONTOUR_SERVICE_NAME%.%CONTOUR_NAMESPACE%.svc.cluster.local:%CONTOUR_SERVICE_PORT%"
  set-response-headers:
  - "X-Envoy-Response-Flags=%RESPONSE_FLAGS%"
```
This is the same as if the above requestHeadersPolicy was added to every route service. These headers would need to be set at the service level so that the service name and port are available.

With these changes the Linkerd header requirement is met but without anything specific for Linkerd, enabling similar requirements to be met in future.

[...]

PR #3269, opened 2021-01-23
Dynamic service headers

erwbgy:
First part of #3258

Add support for %CONTOUR_NAMESPACE%, %CONTOUR_SERVICE_NAME% and
%CONTOUR_SERVICE_PORT% dynamic variables and have these expanded like
Envoy dynamic variables are in #3234. (The CONTOUR_ prefix is used so
that there is no possibility of a clash with a future Envoy dynamic
variable.)

If any of these variables are used where they can't be expanded then
they are just passed through literally. For example if
%CONTOUR_SERVICE_NAME% is used at the route level then it is not
expanded and becomes %%CONTOUR_SERVICE_NAME%% when passed to Envoy. This
avoids any Envoy configuration errors due to bad dynamic variables.


internal/dag/httpproxy_processor.go
```diff
@@ -405,6 +406,10 @@ func (p *HTTPProxyProcessor) computeRoutes(
 		delete(p.orphaned, types.NamespacedName{Name: includedProxy.Name, Namespace: includedProxy.Namespace})
 	}
 
+	dynamicHeaders := map[string]string{
+		"CONTOUR_NAMESPACE": proxy.Namespace,
```
youngnick, line 410: Seems like this is the only dynamic header added so far? I can't see any sign of the other two, `%CONTOUR_SERVICE_NAME%`, and `%CONTOUR_SERVICE_PORTR%`?

youngnick (changes requested):
The overall shape of this looks great, I think it's just missing:
- support for the extra two fields mentioned in the issue and the description
- testing that the dynamic headers are added correctly. I'd like to see the tests in `builder_test.go` and `policy_test.go` filled out a bit.
- Ideally, we'd have an integration test of some sort that covers testing the user-visible behavior, but that could conceivably be follow-up work.

internal/dag/httpproxy_processor.go
erwbgy: They're set on lines 567 and 568 below where the service name and port are available.

sunjayBhatia (commented):
> testing that the dynamic headers are added correctly. I'd like to see the tests in builder_test.go and policy_test.go filled out a bit.
 +1

skriss:
Let's make sure we get docs updates for these new variables as well.

internal/dag/httpproxy_processor.go
youngnick: Ah, so they are, thanks.

youngnick:
@erwbgy, thanks again for all your work on this. We're due to cut a release before the end of this week, and this is really close. Any chance you could hit the outstanding TODOs?
- expanded testing in `builder_test.go` and `policy_test.go`
- documentation of the available fields in the header replacement HTTPProxy documentation?

erwbgy:
Apologies, I should have added the tests earlier - and they found a bug.  Some documentation changes added as well.

erwbgy:
Add integration test.

youngnick (approved):
This looks great now, thanks @erwbgy.

sunjayBhatia (approved):

skriss:
Any objections to getting this one into 1.12?

sunjayBhatia:
> Any objections to getting this one into 1.12?

I would think it is good to go 👍🏽 looks well tested and a small surface area so not super risky

erwbgy:
@skriss I see the [code](https://github.com/projectcontour/contour/blob/release-1.12/internal/dag/httpproxy_processor.go#L409) changes in the 1.12.0 release but it looks like the docs did not get updated?  I see the documentation changes in [main](https://github.com/projectcontour/contour/blob/main/site/docs/main/config/request-rewriting.md) but not [1.12.0](https://github.com/projectcontour/contour/blob/main/site/docs/v1.12.0/config/request-rewriting.md).

skriss:
Ah, shoot, I probably got the 1.12 docs ready before your PR merged and forgot to update them. I'll get a fix in, thanks for pointing this out.

skriss:
See https://github.com/projectcontour/contour/pull/3341.

Issue #3227, opened 2020-12-18
Linter errors not showing


stevesloka:
Looking at the go reportcard, there are a bunch of linter errors that aren't showing up when the lint task is run, we should fix that and push our score up: https://goreportcard.com/report/github.com/projectcontour/contour#golint

seemiller:
I've been working with Contour on a work-related project and am interested in contributing more to the Kubernetes OSS community. This looks like a good first issue that I am willing to take on. Could this issue be assigned to me?

skriss:
@seemiller that'd be great! Let us know if you need any input.

PR #3271, opened 2021-01-24
contour: Linter errors not showing #3227

seemiller:
There are a number of linter errors that golangci-lint is excluding by default.
By including EXC0002, linter errors of the following form should be shown.

warning: exported ... should have comment or be unexported

Updates #3227

Signed-off-by: Nicholas Seemiller <seemiller@gmail.com>

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -59,3 +59,9 @@ issues:
   - linters:
     - golint
     text: "don't use ALL_CAPS in Go names; use CamelCase"
+
+  # Independently from option `exclude` we use default exclude patterns,
+  # it can be disabled by this option. To list all
+  # excluded by default patterns execute `golangci-lint run --help`.
+  # Default value for this option is true.
+  exclude-use-default: false


seemiller:
It appears that by default, golangci-lint excludes a number of common, "annoying" linter warnings. It is suppose to be possible to include, or re-enable some of these checks. I've done that with one such check, the EXC0002. However, I am still seeing quite a discrepancy between GolangCI-Lint and golint. For example, running `golint ./...` in the Contour repo results in around 140 warnings of this type. This is consistent with what is reported from the Go Report Card. Running `golangci-lint run | grep golint` in the Contour directory reports only around 50 warnings. Clearly, GolangCI-Lint is still suppressing warnings somehow.

I'll continue research why this difference exists.

jpeach:
> It appears that by default, golangci-lint excludes a number of common, "annoying" linter warnings. It is suppose to be possible to include, or re-enable some of these checks. I've done that with one such check, the EXC0002. However, I am still seeing quite a discrepancy between GolangCI-Lint and golint. For example, running `golint ./...` in the Contour repo results in around 140 warnings of this type. This is consistent with what is reported from the Go Report Card. Running `golangci-lint run | grep golint` in the Contour directory reports only around 50 warnings. Clearly, GolangCI-Lint is still suppressing warnings somehow.
> 
> I'll continue research why this difference exists.

See the `--exclude-use-default` flag in golangci-lint. Not sure we want the comment-related ones; Contour hasn't been consistent about godoc comments and backfilling is probably a lot of typing.

seemiller:
@jpeach thanks for the suggestion. I had seen that option, but I guess I was focused on the prevalent warning.

I reverted the change I made to include the `comment on exported...` warnings. I then added the config, `exclude-use-default: false`.

So the reason that I focused on the `comment on exported...` warnings is that that type of warning are the majority of warnings reported in the report card. Of the 178 warnings in the report card, 152 of them are of the `comment on exported...` type. If enabling and reporting these would create too much noise and too much typing to backfill, I then ask, is this mostly a non-issue?

Here is a summary of the 26 linter warnings remaining:
- 10 - comment on exported ... should be of the form
- 9 - don't use underscores in Go names
- 7 - package comment should be of the form ...

I'm happy to take a stab at fixing these 26 warnings. Just don't think that this is the proper issue to perform that work under.

Also, another issue is that the Make `lint` target does not seem to work -
```shell
make lint
Running Go linter ...
Running YAML linter ...
Unable to run codespell. Please check installation instructions:
	https://github.com/codespell-project/codespell#installation
make: *** [Makefile:158: lint-codespell] Error 69
```

If you agree that is another issue, I'm happy to create that issue and look into it as well.

jpeach:
...
> Here is a summary of the 26 linter warnings remaining:
> 
> * 10 - comment on exported ... should be of the form
> * 9 - don't use underscores in Go names
> * 7 - package comment should be of the form ...
> 
> I'm happy to take a stab at fixing these 26 warnings. Just don't think that this is the proper issue to perform that work under.

I'm very happy for you to tackle these warnings :) 

The approach I recommend is to keep this PR open, and post a separate PR to fix up the linter warnings that you find. Once the lint runs cleanly with the new config, then we can land this.

> Also, another issue is that the Make `lint` target does not seem to work -
> 
> ```shell
> make lint
> Running Go linter ...
> Running YAML linter ...
> Unable to run codespell. Please check installation instructions:
> 	https://github.com/codespell-project/codespell#installation
> make: *** [Makefile:158: lint-codespell] Error 69
> ```
> 
> If you agree that is another issue, I'm happy to create that issue and look into it as well.

Yep I agree with this. All the linters should work locally.

skriss:
@seemiller did you want to continue with this PR?

seemiller:
@skriss I would like to continue, but to be honest, I'm not going to be able to look at it in the next few weeks (you know, life and all that). I certainly would not object if someone wanted to pick it up in the meantime, or de-prioritize it until I can devote some more time.

skriss:
@seemiller no problem. I'll bump it into the next milestone for now!

sunjayBhatia:
Moving to 1.15.0

youngnick:
@seemiller, this one got closed by the bot, if you want to pick it back up, please comment and we'll reopen, or you could log a new one (since it's probably pretty out of date now anyway).

Issue #2287, opened 2020-02-25
Implement service-apis behavior


youngnick:
Once the behavior of the service-apis objects is decided, this issue will cover actually implementing that behavior.

youngnick:
Blocked on service-apis actually defining the behaviour.

danehans:
[v1alpha1](https://docs.google.com/document/d/1BJoxZTbh2bkYQnqVOETm6_7q4hjwGop2pTZDc__5olI/edit) of service apis is scheduled to be released this month (August 2020). As discussed during this week's community call, it would be helpful to begin planning for the implementation. @jpeach suggested that work such as https://github.com/projectcontour/contour/issues/2683 be completed prior to the service apis implementation. Should this issue be used as the meta issue for describing a) the work that needs to be completed prior to implementing service apis and b) the actual service api's implementation? Should a separate issue be created for a? In general, I would like to enumerate the work needed to implement service-apis, including anything that's needed prior to the actual implementation so myself and others can begin helping with the service apis implementation.

/cc @Miciah @stevesloka

youngnick:
Thanks @danehans, I agree that this issue should hold:
- a list of issues that block doing some implementation of the service apis behavior
- a list of issues that need to be implemented to actually deliver the behavior as at v1alpha1.

We should figure those out here, then I will add them to the initial description as a checklist so we get a count.

jpeach:
Blocked on https://github.com/projectcontour/contour/issues/2226, which will allow new builders to be added at runtime.

youngnick:
- [ ] Decision about *Class behavior (IngressClass and GatewayClass)
- [ ] Work to add builder interface (@stevesloka)
- [ ] Work to add interface between DAG and xDS caches (@jpeach)

youngnick:
Okay, for the service-apis objects to be supported in Contour, we will need:
- [ ] GatewayClass watching support. Note that as decided in #2809, only one GatewayClass is supported per Contour deployment. The best way to make this work is TBD.
- [ ] Gateway watching support. Note that Contour itself will not create Gateways, it will watch for Gateways in its GatewayClass, and build a configuration to send valid Listeners to Envoy, writing back information about Listener status to the relevant Gateway.
- [ ] HTTPRoute watching support. Contour should take HTTPRoutes referenced from Listeners, and add config to Envoy to expose the routes.
[...]

PR #3275, opened 2021-01-25
internal/dag: Filter GatewayClasses to projectcontour.io domain

stevesloka:
Filters the GatewayClasses discovered to only cache those that have the Spec.Controller domain of projectcontour.io.

Updates #2287

### DIFF
--- a/internal/dag/cache.go
+++ b/internal/dag/cache.go
@@ -16,6 +16,7 @@ package dag
 import (
 	"errors"
 	"fmt"
+	"strings"
 	"sync"
 
 	contour_api_v1 "github.com/projectcontour/contour/apis/projectcontour/v1"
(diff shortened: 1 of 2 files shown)


youngnick:
Is there any reason to watch GatewayClass at all any more? Contour doesn't interact with them at all, since we will pass it a named Gateway to check.

stevesloka:
But what if you don't pass a named gateway. Given a single GatewayClass with a selected Gateway we can filter on the class and verify it's intended for contour. 

Couldn't someone create a class named "external" but for another controller and set that to have contour use, but then you'd get conflicting routes. 

I can take it out, but seems like we should at least validate that we're processing against a class that is intended for contour.

stevesloka:
Closing since the team doesn't see a need for this.

PR #3278, opened 2021-01-26
internal/dag: Initial implementation for service-apis

stevesloka:
Super rough initial implementation for service-apis.

Needs a lot of love, but at least has started to work while still behind the feature flag in Contour. Using this sample yaml I can get things to route properly:

```
kind: Gateway
metadata:
  name: contour-gateway
spec:
  gatewayClassName: contour-class
  listeners:
    - protocol: HTTP
      port: 80
      routes:
        kind: HTTPRoute
        selector:
          matchLabels:
            app: filter
        namespaces:
          from: "All"
---
kind: HTTPRoute
metadata:
  name: http-filter-1
  namespace: projectcontour-roots 
  labels:
    app: filter
spec:
  hostnames:
    - local.projectcontour.io
  rules:
    - matches:
      - path:
          type: Prefix
          value: /
      forwardTo:
      - serviceName: rootapp
        port: 80
```

Updates #2287

### DIFF
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -18,6 +18,8 @@ import (
 	"testing"
 	"time"
 
+	serviceapis "sigs.k8s.io/service-apis/apis/v1alpha1"
+
 	contour_api_v1 "github.com/projectcontour/contour/apis/projectcontour/v1"
 	"github.com/projectcontour/contour/internal/fixture"
 	"github.com/projectcontour/contour/internal/timeout"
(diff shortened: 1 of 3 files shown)


stevesloka:
Ok this is a bit better with tests, but I think a reasonable start. It doesn't have any status information which I think would be a good next step to get integrated, but want to limit the changes since it's turning into a large PR.

internal/dag/serviceapis_processor.go
```diff
@@ -26,5 +35,126 @@ func (p *ServiceAPIsProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+		r.PathMatchCondition = &PrefixMatchCondition{Prefix: prefix}
+		routes = append(routes, r)
+	}
+
+	return routes
+}
+
+// httpRoutes returns a slice of *serviceapis.HTTPRoute objects.
+// invalid HTTPRoute objects are excluded from the slice and their status
+// updated accordingly.
```
danehans, line 130: s/Invalid/invalid/

I don't see where status is being updated. If the status update occurs elsewhere, I suggest removing "and their status updated accordingly."

internal/dag/serviceapis_processor.go
```diff
@@ -26,5 +35,126 @@ func (p *ServiceAPIsProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+func (p *ServiceAPIsProcessor) validHTTPRoutes() []*serviceapis.HTTPRoute {
+	var valid []*serviceapis.HTTPRoute
+
+	// Validate this HTTPRoute references an existing Gateway when the "allow" type
+	// on the HTTPRoute is set to "FromList".
+	for _, route := range p.source.httproutes {
+		switch route.Spec.Gateways.Allow {
+		case serviceapis.GatewayAllowFromList:
```
danehans, line 138: Should case statements be added for "All" and "SameNamespace" allow types? I would expect a route to be precluded from `valid` if `allow: All` and no gateways exist in any namespace. I would also expect a route to be precluded from `valid` if `allow: SameNamespace` and no gateways exist in the route's namespace.

internal/dag/serviceapis_processor.go
```diff
@@ -26,5 +35,126 @@ func (p *ServiceAPIsProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+		for _, vhost := range hosts {
+			vhost := p.dag.EnsureVirtualHost(vhost)
+			for _, route := range routes {
+				vhost.addRoute(route)
+			}
+		}
+	}
+}
+
+// route builds a []*dag.Route for the supplied HTTPRoute.
```
danehans, line 105: I'm confused by the above godoc. A string and Service slice is being provided, not an HTTPRoute.

internal/dag/serviceapis_processor.go
```diff
@@ -26,5 +35,126 @@ func (p *ServiceAPIsProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+
+func (p *ServiceAPIsProcessor) computeHTTPRoute(route *serviceapis.HTTPRoute) {
+
+	// Determine the hosts on the route, if no hosts
+	// are defined, then set to "*".
+	var hosts []string
+	if len(route.Spec.Hostnames) == 0 {
+		hosts = append(hosts, "*")
+	} else {
+		for _, host := range route.Spec.Hostnames {
```
danehans, line 57: I suggest validating each hostname. The first phase of hostname validation can be:
- Ensure the hostname is not an IP address. xref: https://golang.org/pkg/net/#ParseIP
- Ensure the hostname does not contain a port. xref: https://golang.org/pkg/net/#SplitHostPort

As a second phase, the `hostname` spec states "Hostname is the fully qualified domain name of a network host, as defined by RFC 3986."  We should be able to use https://golang.org/pkg/net/?m=all#isDomainName (verify that it supports wildcard prefixes or augment to support wildcards).

internal/dag/serviceapis_processor.go
```diff
@@ -26,5 +35,126 @@ func (p *ServiceAPIsProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+		for _, forward := range rule.ForwardTo {
+			// Verify the service is valid
+			if forward.ServiceName == nil {
+				p.Error("ServiceName must not specified and is currently only implemented!")
+				break
+			}
+			meta := types.NamespacedName{Name: *forward.ServiceName, Namespace: route.Namespace}
+
+			// TODO: Refactor EnsureService to take an int32 so conversion to intstr is not needed.
+			service, err := p.dag.EnsureService(meta, intstr.FromInt(int(forward.Port)), p.source)
```
danehans, line 86: Note that if https://github.com/kubernetes-sigs/service-apis/pull/538 merges, port will be optional. With this PR, when port is unspecified, the implementation should forward the request with the port unmodified. Take a look and comment if this is an issue.

internal/dag/serviceapis_processor.go
```diff
@@ -26,5 +35,126 @@ func (p *ServiceAPIsProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+		routes = append(routes, r)
+	}
+
+	return routes
+}
+
+// httpRoutes returns a slice of *serviceapis.HTTPRoute objects.
+// invalid HTTPRoute objects are excluded from the slice and their status
+// updated accordingly.
+func (p *ServiceAPIsProcessor) validHTTPRoutes() []*serviceapis.HTTPRoute {
```
skriss, line 131: I think limiting the routes we process to only those that map to this Contour's gateway is probably essential functionality for an MVP here - we have the config option in the config file, so should be able to wire that through.

skriss (changes requested):
Handful of initial comments. This looks like a good start on MVP functionality. In general, I'd say let's just be explicit about which subset of functionality we're supporting as we go, and logging errors/not processing config that we can't support yet.

internal/dag/serviceapis_processor.go
```diff
@@ -26,5 +35,126 @@ func (p *ServiceAPIsProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+		// Paths are a set of paths that are
+		var pathPrefixes []string
+		var services []*Service
+
+		for _, match := range rule.Matches {
+			switch match.Path.Type {
+			case serviceapis.PathMatchPrefix:
+				pathPrefixes = append(pathPrefixes, stringOrDefault(match.Path.Value, "/"))
+			default:
+				p.Error("NOT IMPLEMENTED: Only PathMatchPrefix is currently implemented.")
```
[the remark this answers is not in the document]
stevesloka: There's an `exact` match as well which Contour doesn't have support for yet. I was trying to not get backed into the current HTTPProxy issue of one `error` makes the entire resource invalid.

internal/dag/serviceapis_processor.go
```diff
@@ -26,5 +35,126 @@ func (p *ServiceAPIsProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+			if forward.ServiceName == nil {
+				p.Error("ServiceName must not specified and is currently only implemented!")
+				break
+			}
+			meta := types.NamespacedName{Name: *forward.ServiceName, Namespace: route.Namespace}
+
+			// TODO: Refactor EnsureService to take an int32 so conversion to intstr is not needed.
+			service, err := p.dag.EnsureService(meta, intstr.FromInt(int(forward.Port)), p.source)
+			if err != nil {
+				// TODO: Raise `ResolvedRefs` condition on Gateway with `DegradedRoutes` reason.
```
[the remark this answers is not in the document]
stevesloka: 👍

internal/dag/serviceapis_processor.go
stevesloka: I'll update the comment, currently this PR doesn't have status implemented, just trying to get a baseline set of code in opening up more features.

internal/dag/serviceapis_processor.go
stevesloka: Let me take this logic out all together, then reference resolution in a new issue? I think that would be best if you're ok with that.  (#3351)

internal/dag/serviceapis_processor.go
stevesloka: Updated the godoc string. I think originally I had this take a set of HTTPRoutes but changed to just prefixes/services.

internal/dag/serviceapis_processor.go
stevesloka: #3352

internal/dag/serviceapis_processor.go
stevesloka: #3353

internal/dag/serviceapis_processor.go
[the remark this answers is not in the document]
stevesloka: These are places where we need to raise status condition errors. I think we should try and not throw away the entire resource for things that are either not implemented yet or are in error (ref #3145).

internal/dag/serviceapis_processor.go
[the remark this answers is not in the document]
stevesloka: 👍

internal/dag/serviceapis_processor.go
```diff
@@ -26,5 +35,126 @@ func (p *ServiceAPIsProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+			vhost := p.dag.EnsureVirtualHost(vhost)
+			for _, route := range routes {
+				vhost.addRoute(route)
+			}
+		}
+	}
+}
+
+// route builds a []*dag.Route for the supplied HTTPRoute.
+func (p *ServiceAPIsProcessor) route(pathPrefixes []string, services []*Service) []*Route {
```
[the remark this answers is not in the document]
stevesloka: 👍

internal/dag/serviceapis_processor.go
```diff
@@ -26,5 +35,126 @@ func (p *ServiceAPIsProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+		p.source = nil
+	}()
+
+	for _, route := range p.validHTTPRoutes() {
+		p.computeHTTPRoute(route)
+	}
+}
+
+func (p *ServiceAPIsProcessor) computeHTTPRoute(route *serviceapis.HTTPRoute) {
+
```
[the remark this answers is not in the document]
stevesloka: 👍

internal/dag/serviceapis_processor.go
stevesloka: I opened https://github.com/projectcontour/contour/issues/3351 to track this. Do you think it's ok to merge this as-is, then follow up with that? I do understand that right now we'll process ALL gateways. =)

stevesloka: https://github.com/projectcontour/contour/pull/3278#discussion_r574809218

internal/dag/serviceapis_processor.go
```diff
@@ -26,5 +35,98 @@ func (p *ServiceAPIsProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+		hosts = append(hosts, "*")
+	} else {
+		for _, host := range route.Spec.Hostnames {
+			hosts = append(hosts, string(host))
+		}
+	}
+
+	for _, rule := range route.Spec.Rules {
+
+		// Paths are a set of paths that are
```
youngnick, line 69: Comment needs finishing up?

internal/dag/builder_test.go
```diff
@@ -28,8 +30,375 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/pointer"
 )
 
+func TestDAGInsertServiceAPIs(t *testing.T) {
+
+	svc1 := &v1.Service{
```
youngnick, line 38: I'd recommend calling these kuardSvc and blogSvc respectively, that saves one lookup when you're writing more tests.

youngnick (approved):
Couple of small things from me, but this looks okay as a first pass. Definitely plenty of work to do still though.

internal/dag/serviceapis_processor.go
```diff
@@ -13,9 +13,18 @@
...
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	serviceapis "sigs.k8s.io/service-apis/apis/v1alpha1"
+)
+
 // ServiceAPIsProcessor translates Service API types into DAG
 // objects and adds them to the DAG.
 type ServiceAPIsProcessor struct {
+	logrus.FieldLogger
```
jmprusi, line 26: All the usages of the FieldLogger (p.Error,p.Errorf...) panic, as this has seems to not be initialized (nil).

stevesloka: 👍 I added this to the init of the processor.

internal/dag/builder_test.go
```diff
@@ -28,8 +30,375 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/pointer"
 )
 
+func TestDAGInsertServiceAPIs(t *testing.T) {
```
sunjayBhatia, line 36: Worth adding tests (or maybe a tracking issue for adding tests) around the interactions between different resources (HTTPProxy, Ingress, Gateway APIs etc.)

internal/dag/serviceapis_processor.go
```diff
@@ -26,5 +35,97 @@ func (p *ServiceAPIsProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+				pathPrefixes = append(pathPrefixes, stringOrDefault(match.Path.Value, "/"))
+			default:
+				p.Error("NOT IMPLEMENTED: Only PathMatchPrefix is currently implemented.")
+			}
+		}
+
+		for _, forward := range rule.ForwardTo {
+			// Verify the service is valid
+			if forward.ServiceName == nil {
+				p.Error("ServiceName must not specified and is currently only implemented!")
```
sunjayBhatia, line 84: nit: Should this error be reworded?

sunjayBhatia (approved):
Looks good to me!

On testing, left a comment about adding a new issue to track the test coverage we will need

Also, since the conformance suite is not ready, should we add some basic integration tests  that we can throw away later once we can start contributing upstream? Worth adding an issue for that?

internal/dag/builder_test.go
stevesloka: 👍  https://github.com/projectcontour/contour/issues/3367

internal/dag/serviceapis_processor.go
stevesloka: haha yup! 👍

PR #3283, opened 2021-01-28
pkg/config: Add Gateway reference parameters to config file

stevesloka:
Adds a gateway reference name/namespace parameter to the config file so that when using service-apis, users are able to define which Gateway this instance of Contour should configure.

Updates #3213

### DIFF
--- a/examples/contour/01-contour-config.yaml
+++ b/examples/contour/01-contour-config.yaml
@@ -11,6 +11,11 @@ data:
     #   determine which XDS Server implementation to utilize in Contour.
     #   xds-server-type: contour
     #
+    # Specify the service-apis Gateway Contour should watch.
+    # gateway:
+    #   name: contour
+    #   namespace: projectcontour
+    #
     # should contour expect to be running inside a k8s cluster
     # incluster: true
     #
--- a/pkg/config/parameters.go
+++ b/pkg/config/parameters.go
@@ -42,6 +42,27 @@ func (s ServerType) Validate() error {
 	}
 }
 
+// Validate the GatewayConfig.
+// Name & Namespace must be specified.
+func (g GatewayParameters) Validate() error {
+
+	var errorString string
+	if len(g.Name) == 0 {
+		errorString = "name required"
+	}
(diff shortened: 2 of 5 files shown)


examples/contour/01-contour-config.yaml
```diff
@@ -11,6 +11,11 @@ data:
     #   determine which XDS Server implementation to utilize in Contour.
     #   xds-server-type: contour
     #
+    # specify the service-apis Gateway Contour should configure
```
sunjayBhatia, line 14: I think it makes sense to be at top level since this will be guarded behind the service APIs flag anyways right?

pkg/config/parameters.go
```diff
@@ -421,6 +453,10 @@ func (p *Parameters) Validate() error {
 		return err
 	}
 
+	if err := p.GatewayConfig.Validate(); err != nil {
```
sunjayBhatia, line 456: Should we only do validation of these bits if the service APIs CLI flag is set?

sunjayBhatia (commented):
left my 2 cents on the config file structure

examples/contour/01-contour-config.yaml
youngnick: This is the design I have in the design document - I did it this way so that this being non-nil can replace the boolean flag. I don't think there's a need for the flag when setting a value performs the same function.

youngnick: Nit:
```suggestion
    # Specify the service-apis Gateway Contour should watch.
```

stevesloka: Yeah this was just setting up the groundwork to allow the Gateway to be configured here. Once this is in we can start to utilize it in the code without breaking anything else.

pkg/config/parameters.go
stevesloka: That flag will go away, so the idea here is to validate what the user inputed. Both are required according to Nick's design, but will default if not specified.

sunjayBhatia (approved):

PR #3290, opened 2021-01-29
internal/featuretests: use IsValid/IsInvalid helpers

skriss:
Adds an IsInvalid() helper to match the existing IsValid() helper,
and uses them in all the feature tests to be more concise.

Also adds comments about the internal/featuretests event handler
being blocking.


sunjayBhatia (approved):
LGTM nice cleanup

stevesloka (approved):

Issue #2880, opened 2020-09-03
Make the TLS cipher suite configurable


jpeach:
**Please describe the problem you have**

It's common for sites to need to configure the TLS cipher suite. This should be globally configurable, even better if we can allow it to be configures per host.

xref #2878
xref #2401

sunjayBhatia:
WIP branch: https://github.com/sunjayBhatia/contour/tree/configurable-cipher-suites

youngnick:
For this @sunjayBhatia, how are you planning on having the cipher suites configurable? Could you put a summary in here? That way, we can get any bikeshedding about config file formatting, etc out of the way before you get too much coding done.

sunjayBhatia:
Basically following how the minimum TLS version was added:

- global option in config file, if not set/empty will use default list we already use
- Configurable per ingress with an annotation
- Configurable per httpproxy with a field under TLS
- Ingress/httpproxy setting overrides global config setting
- Open question about validation: should we allow invalid ciphers or ciphers that are insecure? If not, how do we fail, ignore, overwrite?

youngnick:
I think that the default list is our best effort. If you're taking the time to specify a cipher list, then you get what you ask for, and it's on you if it's insecure.

For cipher validity, though, we probably need to have an allow-list for valid cipher names, so we don't run afoul of the NACK problem (#1176) if we pass an invalid cipher name to Envoy, but also so that we can surface typos back to the user quickly, instead of waiting for the config to go to Envoy.

sunjayBhatia:
Working through this, and noticing that the Min TLS version is not *really* validated, if it's not `1.1`, `1.2`, or `1.3` we set it to `1.2` silently and don't seem to error or give any feedback to the user.

We could keep this and do the same with TLS ciphers (if you provide a set of invalid ciphers, silently set the default set), just wondering if this is intended. Also of course, we could and maybe should give more feedback via logs or `contour` failing to start (at least if there are mistakes in the config file).

sunjayBhatia:
Basically got most of the pieces done, but the above seems like something to iron out before polishing the change

youngnick:
I thought we had changed the TLS version to at least log that the setting was unusable. Part of me wants to make that fail fast and hard - there are only three valid options currently, if you don't supply one of those, then fail Contour startup. That's a pretty big behavior change though, and needs its own ticket.

[...]

PR #3292, opened 2021-01-29
Add global TLS cipher suites configurability

sunjayBhatia:
Ciphers are validated against the allowed ciphers for Envoy. Contour
will exit on startup if any invalid ciphers are present in the config
file. If no ciphers are provided, Contour will use the defaults that
exist now.

Fixes https://github.com/projectcontour/contour/issues/2880


sunjayBhatia:
TODO: add docs

internal/xdscache/v3/listener.go
```diff
@@ -80,6 +80,8 @@ type ListenerConfig struct {
 	// MinimumTLSVersion defines the minimum TLS protocol version the proxy should accept.
 	MinimumTLSVersion string
 
+	CipherSuites []string
```
skriss, line 83: nit: add godoc for consistency.

pkg/config/parameters.go
```diff
@@ -225,6 +225,26 @@ type TLSParameters struct {
 	// to be used when establishing TLS connection to upstream
 	// cluster.
 	ClientCertificate NamespacedName `yaml:"envoy-client-certificate,omitempty"`
+
+	CipherSuites TLSCiphers `yaml:"cipher-suites,omitempty"`
```
skriss, line 229: needs godoc

skriss (changes requested):
This is looking good, just a couple nits.

I guess the one limitation worth calling out here is the issue around FIPS/custom ciphers potentially not being customizable (per https://github.com/projectcontour/contour/issues/2880#issuecomment-758011309), but we're no worse off in those scenarios with this PR so I think we can kick that can down the road.

sunjayBhatia:
> This is looking good, just a couple nits.
> 
> I guess the one limitation worth calling out here is the issue around FIPS/custom ciphers potentially not being customizable (per [#2880 (comment)](https://github.com/projectcontour/contour/issues/2880#issuecomment-758011309)), but we're no worse off in those scenarios with this PR so I think we can kick that can down the road.

Tried to address it briefly in the docs page: https://github.com/projectcontour/contour/pull/3292/files#diff-f24f2ba56c35093431024eed678979b39efcfd720fc3ef24ccccb705a4ad66e8

For FIPS-allowed ciphers actually I think I'm missing some ciphers to validate against since I just did a map element check

https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto#envoy-v3-api-field-extensions-transport-sockets-tls-v3-tlsparameters-cipher-suites

non-FIPS
```
[ECDHE-ECDSA-AES128-GCM-SHA256|ECDHE-ECDSA-CHACHA20-POLY1305]
[ECDHE-RSA-AES128-GCM-SHA256|ECDHE-RSA-CHACHA20-POLY1305]
...
```

vs.  FIPS
```
ECDHE-ECDSA-AES128-GCM-SHA256
ECDHE-RSA-AES128-GCM-SHA256
```

internal/xdscache/v3/listener.go
```diff
@@ -80,6 +80,10 @@ type ListenerConfig struct {
 	// MinimumTLSVersion defines the minimum TLS protocol version the proxy should accept.
 	MinimumTLSVersion string
 
+	// Ciphersuites defines the ciphers Envoy TLS listeners will accept when
```
skriss, line 83:
```suggestion
	// CipherSuites defines the ciphers Envoy TLS listeners will accept when
```

youngnick (approved):
This LGTM. I'd like to get @moderation to check this one out though, to make sure that the documentation around the default and supported ciphers is clear.

We also need to make sure that we check each time we upgrade Envoy to ensure that we match the available ciphers. Maybe we need an "Envoy upgrade" checklist about things to look for?

sunjayBhatia:
Getting some failed CI runs with `authorization required` errors on Docker operations :(

pkg/config/ciphersuites.go
```diff
@@ -0,0 +1,81 @@
...
+	"ECDHE-RSA-AES256-GCM-SHA384",
+	"ECDHE-ECDSA-AES256-SHA",
+	"ECDHE-RSA-AES256-SHA",
+	//"AES256-GCM-SHA384",
+	//"AES256-SHA",
+})
+
+// ValidTLSCiphers contains the list of TLS ciphers that Envoy supports
+// See: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto#extensions-transport-sockets-tls-v3-tlsparameters
+// Note: This list is a superset of what is valid for stock Envoy builds and those using BoringSSL FIPS.
```
skriss, line 65: what happens if a user configures a FIPS cipher suite as allowed, but they're not using a FIPS Envoy build? It sounds like from your previous comment that Envoy would just ignore it? If so, I think that's probably fine; it's a reasonable tradeoff for enabling this to work for FIPS environments too.

sunjayBhatia: The Envoy docs on the allowed ciphers are a little misleading, the literal strings of the two ciphers that "differ" for the FIPS build `ECDHE-ECDSA-AES128-GCM-SHA256` and `ECDHE-RSA-AES128-GCM-SHA256` are accepted by a plain Envoy and not ignored even though the docs make it look like Envoy wouldn't accept that string (though of course if you see `[ECDHE-ECDSA-AES128-GCM-SHA256|ECDHE-ECDSA-CHACHA20-POLY1305]` it makes sense).

skriss: Ah reading https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#Cipher-suite-configuration helps explain that notation, I wasn't aware of what it meant.

skriss:
```
If configuring a server, one may also configure equal-preference groups to partially respect the client's preferences when SSL_OP_CIPHER_SERVER_PREFERENCE is enabled. Ciphers in an equal-preference group have equal priority and use the client order. This may be used to enforce that AEADs are preferred but select AES-GCM vs. ChaCha20-Poly1305 based on client preferences. An equal-preference is specified with square brackets, combining multiple selectors separated by |. For example:

[TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256|TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256]

Once an equal-preference group is used, future directives must be opcode-less. Inside an equal-preference group, spaces are not allowed.
```

sunjayBhatia: yep exactly so we allow one to specify the preference groups *or* the standalone cipher that is FIPS compatible that Envoy will accept

pkg/config/ciphersuites.go
```diff
@@ -0,0 +1,81 @@
...
+var ValidTLSCiphers = map[string]struct{}{
```
stevesloka, line 66: Would it make any sense to move this valid list outside of a go file into a config somewhere such that someone could update the ciphers without a Contour release?

skriss (changes requested):
one small suggestion, but otherwise LGTM!

pkg/config/ciphersuites.go
sunjayBhatia: Like an additional config file to pass to contour? Or an additional entry in the contour config file (that would default to the existing list if not set)?

internal/annotation/annotations.go
```diff
@@ -244,3 +245,25 @@ func MaxRequests(o metav1.Object) uint32 {
...
+				uniqueCiphers[cipher] = true
+				validatedCiphers = append(validatedCiphers, cipher)
+			}
+		}
+	}
+	if len(validatedCiphers) == 0 {
+		return defaultCiphers
+	}
+	return validatedCiphers
+}
```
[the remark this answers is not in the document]
sunjayBhatia: ah yeah I copied the `MinTLSVersion` annotation placement and did neglect to mention that this function would be used in a subsequent PR

If you think it makes more sense to move it out, i can change that! I didn't put it there initially so we didn't add and remove something from `pkg` etc.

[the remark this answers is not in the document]
sunjayBhatia: we decided to drop the support for this in HTTPProxy/Ingress so this function will be moved! (most likely to `pkg/config` as you suggest @skriss )

pkg/config/ciphersuites.go
youngnick: I don't think that's a good idea, we should be maintaining this list as "all the ciphers Envoy will support", and updating it when we update Envoy.

Then, people may choose if they allow them via the allowed ciphers setting. Otherwise we end up with two ways to do the same thing (limit the cipher set). I think that changing either the valid cipher set or the default one are both significant enough actions to require a version bump of Contour.

sunjayBhatia:
TODO: don't bother validating ciphers if TLS version is `1.3`

sunjayBhatia:
In office hours today and the maintainer meeting yesterday, it was brought up that directly configuring ciphers is often dangerous/not desirable/requires some more in depth knowledge/can be fiddly etc. for operators and rather a security "profile" that may include different sets of cipher suites would be more ideal in order to support different use cases (dealing with legacy systems, dealing with up to date cloud native systems, etc.). Inspiration for that comes from the Mozilla server side TLS configuration guide: https://wiki.mozilla.org/Security/Server_Side_TLS

tldr of the Mozilla profiles/Contour implications:
- `Modern compatibility`:
  - Forces TLS 1.3
  - Ciphers irrelevant, covered by TLS 1.3
  - *We can support this already in Contour since we support setting the minimum TLS version to 1.3*
  - There doesn't seem to be much utility in supporting this exact thing
- `Intermediate compatibility (recommended)`:
  - TLS 1.2, 1.3
  - CIphers are equivalent to the default we set today in Contour (minus the ones we recently wanted to disable `ECDHE-ECDSA-AES128-SHA` and `ECDHE-RSA-AES128-SHA` and a few that Envoy doesn't support)
  - *This would be supported if a user does not set any ciphers when we allow configurability without any changes the current PR*
  - By default Contour should be in this "mode" so if you don't customize TLS parameters you already get this
- `Old backward compatibility`:
  - TLS 1.0 - 1.3
  - Ciphers for TLS 1.2 are a wider range than Envoy supports (and also includes the couple above we wanted to disable recently)
  - *We already restrict to TLS 1.2+ so a lot of the utility of this profile is redundant*
  - *This could be implemented with the current PR by a user selecting the whole set of Envoy supported ciphers (including the ones Contour has deemed to be insecure)*
  - We could build the "profile" to support this but I'm not sure how useful that is since we're restricted to TLS 1.2 anyway?
    - Adding a profile would mean adding a new field to the `TLSParameters` struct to describe which profile is selected which is not difficult, but I'm not sure we would even have many options to provide?
  - We would probably also want to support a `Custom` profile as well, which would be covered by the existing changes in this PR.

Overall my evaluation is we stick with just allowing ciphers to be configurable as is, but WDYT of this assessment? Is there anything I'm missing here? @projectcontour/maintainers @moderation @tsaarni

skriss:
> Overall my evaluation is we stick with just allowing ciphers to be configurable as is, but WDYT of this assessment? Is there anything I'm missing here? @projectcontour/maintainers @moderation @tsaarni

That sounds reasonable to me. There's not a ton of room for users to shoot themselves in the foot as-is since the allowable set of ciphers is small to begin with. We could always follow up with more enhancements here later if/when it makes sense.

Definitely interested to hear users' thoughts as well, though.

sunjayBhatia:
Hold on this please, actually gonna undo the bit that skips validation when TLS 1.3 is the min version as that can be overridden in an Ingress or HTTPProxy

tsaarni:
I also agree that sticking with the cipher lists is good. 

The classification from Mozilla wiki is of course good server setup guideline, but I think in practice "profile" might just make dealing with ciphers even more difficult than it already is.  All tools, such as ssllabs.com directly list ciphers, so I would rather like to keep using the same "language" and not hide them under abstraction.

skriss (approved):
LGTM!

youngnick:
To make this explicit: I agree with @tsaarni here, I think that configuring cipher lists is the way to go. If we want to provide syntactic sugar for this later, we can do that, but this is the base feature to build on.

youngnick:
@sunjayBhatia could you sort out this merge conflict, then we can merge this one.

skriss:
@sunjayBhatia ready to merge here?

PR #3293, opened 2021-01-30
internal/envoy: Set XffNumTrustedHops to keep header proto intact

stevesloka:
If a user has an external load balancer that terminates TLS, the X-Forwarded-Proto
header gets overwritten unless the downstream connection is trusted.

This adds a config file option to set the number of trusted hops which will
allow the headers to be instact already set from downstream.

Fixes https://github.com/projectcontour/contour/issues/3294


stevesloka:
This still needs docs & feature tests.

pkg/config/parameters_test.go
```diff
@@ -86,7 +86,7 @@ default-http-versions: []
 cluster:
   dns-lookup-family: auto
 `
-	assert.Equal(t, strings.TrimSpace(string(data)), strings.TrimSpace(expected))
+	assert.Equal(t, strings.TrimSpace(expected), strings.TrimSpace(string(data)))
```
stevesloka, line 89: Debugging tests I noticed this was swapped.

youngnick (approved):
LGTM

sunjayBhatia (commented):
LGTM pending an addition I’m guessing you’re going to add here https://github.com/projectcontour/contour/blob/main/site/docs/main/configuration.md

stevesloka:
@sunjayBhatia yup added docs and a feature test to cover.

site/docs/main/configuration.md
```diff
@@ -106,6 +107,16 @@ The cluster configuration block can be used to configure various parameters for
 {: class="table thead-dark table-bordered"}
 <br>
 
+### Network Configuration
```
sunjayBhatia, line 110: Do we have any issues we can earmark so we make sure to add configuration to this object?

sunjayBhatia (approved):
LGTM, just a tiny discussion starter comment

site/docs/main/configuration.md
stevesloka: There's an issue linked for the max hops, are you referring to the `Network Configuration` section?

sunjayBhatia: Yeah, just wondering if we already had other things to add in the `network` hash we already know about

stevesloka: Yeah at the moment not sure, things that the chunked length maybe should have gone into there. We can remove the `network` parent, but was trying to group stuff together in a way.

youngnick: I think having the `network` top-level item is great, we should definitely keep that, and look out for things to put in there.

stevesloka: 👍

PR #3295, opened 2021-02-01
remove usages of vendor/ directory

### LINKED ISSUE #3276: Remove `-mod=vendor`, `go mod vendor` from scripts
Contour has a number of places where it runs `go mod vendor` or `export GOFLAGS=-mod=vendor`. I *think* these are obsolete at this point and we can just remove them. 

My specific issue is that I'd like to avoid having to run `go mod vendor` before running `make generate` locally.

The specific instances are shown here:
https://github.com/projectcontour/contour/search?l=Shell&q=vendor&type=code
https://github.com/projectcontour/contour/search?l=Makefile&q=vendor&type=code
https://github.com/projectcontour/contour/search?l=YAML&q=vendor&type=code

Is anyone aware of a reason why we need to keep these, vs. letting `go` automatically download modules as needed?

skriss:
Removes unneeded setting of -mod=vendor from code-gen scripts
and calls to `go mod vendor` to remove requirement for a local
vendor/ directory.

Closes #3276.

### DIFF
--- a/.github/workflows/prbuild.yaml
+++ b/.github/workflows/prbuild.yaml
@@ -38,7 +38,6 @@ jobs:
           go-version: '1.15.6'
       - name: test
         run: |
-          go mod vendor
           make install
           make check-coverage
       - name: codeCoverage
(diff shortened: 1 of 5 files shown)


sunjayBhatia (approved):
nice!

stevesloka (approved):

PR #3297, opened 2021-02-01
Ignore tools.ietf.org in site proof check

sunjayBhatia:
Returns frequent false positive timeouts

See https://github.com/projectcontour/contour/pull/3292/checks?check_run_id=1809735846, checking for https://tools.ietf.org/html/rfc7540#section-9.1.1

Also syncs the example contour configmap with the actual file to ensure the siteproof check reruns in CI


skriss (approved):

youngnick (approved):
LGTM

PR #3288, opened 2021-01-29
contour: Linter warning fixes (#3287)

seemiller:
Fixes warnings of the type:
    - comment on exported ... should be of the form
    - don't use underscores in Go names
    - package comment should be of the form ...

Fixes #3287

Signed-off-by: Nicholas Seemiller <seemiller@gmail.com>

PR #3298, opened 2021-02-01
add global rate limiting support

### LINKED ISSUE #370: [feature] Add rate limiting support

skriss:
Adds support for configuring an external Rate Limit
Service (RLS) that is used for making global rate
limit decisions. Adds support to HTTPProxy for
global rate limit policies, that generate descriptors
for requests to be sent to the RLS for rate-limit
decisions.

Closes #370.


_integration/testsuite/install-ratelimit-service.sh
```diff
@@ -0,0 +1,90 @@
...
+  name: contour
+  namespace: projectcontour
+data:
+  contour.yaml: |
+    rateLimitService:
+      extensionService: projectcontour/ratelimit
+      domain: contour
+      failOpen: false  
+    # The options below are copied from
+    # install-fallback-certificate.sh.
```
skriss, line 74: TODO issue to enable reloading different Contour config files for different test cases

skriss:
@stevesloka @youngnick @sunjayBhatia I'd appreciate an initial review on this. I am still working on documentation.

skriss:
Ran through a full test of the guide, fixed a couple small things.

internal/xdscache/v3/listener.go
```diff
@@ -477,23 +497,26 @@ func (v *listenerVisitor) visit(vertex dag.Vertex) {
...
+				RequestTimeout(v.ListenerConfig.RequestTimeout).
+				ConnectionIdleTimeout(v.ListenerConfig.ConnectionIdleTimeout).
+				StreamIdleTimeout(v.ListenerConfig.StreamIdleTimeout).
+				DelayedCloseTimeout(v.ListenerConfig.DelayedCloseTimeout).
+				MaxConnectionDuration(v.ListenerConfig.MaxConnectionDuration).
+				ConnectionShutdownGracePeriod(v.ListenerConfig.ConnectionShutdownGracePeriod).
+				AllowChunkedLength(v.ListenerConfig.AllowChunkedLength).
+				NumTrustedHops(v.ListenerConfig.XffNumTrustedHops)
+
+			if rlc := v.RateLimitConfig; rlc != nil {
```
sunjayBhatia, line 514: nit: Could this be collapsed so we don't need to add the if (here and above)?

e.g. `envoy_v3.GlobalRateLimitFilter` could take just a `RateLimitConfig` and do the `nil` check there? and `builder.AddFilter` does nothing if it gets `nil`?

sunjayBhatia (commented):
Looks good overall! Lots of changes but you're right mostly generated etc.

Just a couple small comments

cmd/contour/serve.go
```diff
@@ -294,6 +296,36 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		XffNumTrustedHops:             ctx.Config.Network.XffNumTrustedHops,
 	}
 
+	if ctx.Config.RateLimitService.ExtensionService != "" {
+		namespacedName := k8s.NamespacedNameFrom(ctx.Config.RateLimitService.ExtensionService)
```
sunjayBhatia, line 300: Would it make sense to pull this sort of logic (fetching a service/checking for existence/getting some spec params) into a common helper? I'm thinking this logic could possibly be duplicated elsewhere?

site/docs/main/config/api-reference.html
```diff
@@ -9,7 +9,7 @@
 </ul>
 <h2 id="projectcontour.io/v1">projectcontour.io/v1</h2>
 <p>
-<p>This package holds the specification for the projectcontour.io Custom Resource Definitions (CRDs).</p>
+<p>Package v1 holds the specification for the projectcontour.io Custom Resource Definitions (CRDs).</p>
```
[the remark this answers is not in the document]
skriss: I was wondering about this too, just looked into it -- looks like our script to check for uncommitted codegen changes doesn't check the API reference docs: https://github.com/projectcontour/contour/blob/main/hack/actions/check-uncommitted-codegen.sh#L10-L15.

The docs change came from https://github.com/projectcontour/contour/pull/3288.

I'll file an issue to fix the check.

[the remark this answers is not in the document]
skriss: See #3345 #3346, I'll pull this change out of this PR.

internal/xdscache/v3/listener.go
skriss: Yeah good idea, that'd clean things up a bit; though as-is that would create an import cycle so need to re-jigger things a bit.

pkg/config/parameters.go
```diff
@@ -470,6 +470,24 @@ type Parameters struct {
...
+type RateLimitService struct {
+	// ExtensionService identifies the extension service defining the RLS,
+	// formatted as <namespace>/<name>.
+	ExtensionService string `yaml:"extensionService,omitempty"`
+
+	// Domain is passed to the Rate Limit Service.
+	Domain string `yaml:"domain,omitempty"`
+
+	// FailOpen defines whether to allow requests to proceed when the
+	// Rate Limit Service fails to respond properly.
```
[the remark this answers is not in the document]
skriss: :+1: also reminds me I forgot to add to `site/docs/main/configuration.md` and the example config file, will do that as well.

cmd/contour/serve.go
skriss: The existing extension service code gets its `ExtensionService` via the DAG cache, so it's a different code path / not really duplicated with this logic. Maybe we can keep an eye on this for refactoring when we add our next extension service impl.

cmd/contour/serve.go
```diff
@@ -294,6 +296,36 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+		client := clients.DynamicClient().Resource(contour_api_v1alpha1.ExtensionServiceGVR).Namespace(namespacedName.Namespace)
+
+		// ensure the specified ExtensionService exists
+		res, err := client.Get(context.Background(), namespacedName.Name, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("error getting rate limit extension service %s: %v", namespacedName, err)
```
stevesloka, line 306: Should this be a fatal error or not configure rate limiting until the extension service exists?

examples/contour/01-contour-config.yaml
```diff
@@ -99,3 +99,7 @@ data:
     #   Configure the number of additional ingress proxy hops from the
     #   right side of the x-forwarded-for HTTP header to trust.
     #   num-trusted-hops: 0
+    # rateLimitService:
+    #   extensionService: projectcontour/ratelimit
```
stevesloka, line 103: We have comments for other bits in this config explaining what they mean, maybe match that a bit?

examples/ratelimit/04-contour-config.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
```
stevesloka, line 1: Not sure if we should include this or just reference what to setup. I guess it's a nice copy/paste way to follow a guide so probably good.

site/_guides/global-rate-limiting.md
```diff
@@ -0,0 +1,427 @@
...
+## Prerequisites
+
+This guide assumes that you have:
+
+- A local KinD cluster created using [the Contour guide][2].
+- Contour installed and running in the cluster using the [quick start][3].
+
+## Deploy an RLS
+
+For this guide, we'll deploy the [Envoy rate limit service][4] as our RLS.
```
stevesloka, line 26: Maybe just give a couple more sentences on how the RLS works just so it's clear what is happening? (i.e. The RLS uses a Redis cache internally. Blah blah?)

site/docs/main/config/rate-limiting.md
```diff
@@ -128,3 +148,118 @@ spec:
...
+  services:
+    - name: ratelimit
+      port: 8081
+```
+
+Now add a reference to it in the Contour config file:
+```yaml
+rateLimitService:
+  extensionService: projectcontour/ratelimit
+  domain: contour
```
stevesloka, line 181: Maybe describe these fields a bit? Why would I change the `domain`? What does `failOpen` mean, etc.

stevesloka: Just noticed it's defined later on, but maybe redefine or link?

stevesloka (commented):
Overall looks great to me @skriss! Just a couple questions / thoughts.

cmd/contour/serve.go
youngnick, line 306: If this is not fatal, then we need instead to do this similarly to the Envoy service status watcher - a separate goroutine that looks for the defined ExtensionService and sets up the config once it exists and is working.

I'd say that's quite a bit of extra work for minimal gain. If you're configuring the feature, you should have access to restart Contour, and you should be able to ensure everything is set up correctly.

youngnick (approved):
LGTM, nice work.

cmd/contour/serve.go
skriss: Agree it's not ideal to require the extension service to exist at startup. I think we could change the flow somewhat to resolve the extension service ref during the DAG build instead. If it's OK with you, I'll file a follow-up issue to look at that?

examples/contour/01-contour-config.yaml
skriss: added

site/_guides/global-rate-limiting.md
skriss: copied an excerpt from the Envoy RLS README to give some more color.

site/docs/main/config/rate-limiting.md
skriss: Added comments.

cmd/contour/serve.go
skriss: See https://github.com/projectcontour/contour/pull/3298#discussion_r577753164, I think we could make some improvements here but I'm not considering it blocking.

skriss:
I believe I've addressed everyone's review comments now, please take another look!

sunjayBhatia (approved):

cmd/contour/serve.go
stevesloka: yup sure thing!

stevesloka (approved):

skriss:
😅

Issue #3299, opened 2021-02-01
Feature request: moratorium on cipher deprecations


moderation:
Contour 1.11.0 deprecated some older TLS ciphers and that change didn't make it into the release notes - https://github.com/projectcontour/contour/releases/tag/v1.11.0. The specific PR is https://github.com/projectcontour/contour/pull/3154 which removed `ECDHE-RSA-AES128-SHA` and `ECDHE-RSA-AES256-SHA` ciphers.

In our enterprise environment we have Contour configured to only allow >= TLS 1.2. However the cipher removal broke:
- IE11 on Windows 7 virtual desktops ability to negotiate TLS 1.2 with Contour
- Windows Server 2012 R2 .NET or IE11 clients ability to negotiate TLS 1.2 with Contour

Request:
- ensure any breaking changes regarding TLS ciphers, version etc. make the release notes
- revert https://github.com/projectcontour/contour/pull/3237 from release 1.12.0 until a full investigation of the enterprise platforms that the PR may effect
- introduce a moratorium on cipher deprecation's until https://github.com/projectcontour/contour/issues/2880 lands

/cc @youngnick as per Slack chat

youngnick:
Thanks for this @moderation. In this case, I agree we need to roll back #3237 for 1.12.0, while we work on finishing out #2880. @sunjayBhatia already has #3292 which gets started on that.

I'd like some input from the other maintainers here though? @stevesloka @skriss @xaleeks @danehans ?

tsaarni:
> revert #3237 from release 1.12.0 until a full investigation of the enterprise platforms that the PR may effect

Quick comment about the effects:

#3237 impacts the same platforms as #3154 since these ciphers should go together as a pair.  

The difference between the two sets of ciphers is based on which type of keys are used in the server certificate: if server certificate with RSA key is used, that will allow the server to negotiate RSA variants of the ciphers, and server certificate with EC key will allow server to negotiate the EC variants of the ciphers:

| cipher with RSA server certificate | cipher with EC server certificate |
| - | - |
| ECDHE-RSA-AES128-SHA | ECDHE-ECDSA-AES128-SHA |
| ECDHE-RSA-AES256-SHA | ECDHE-ECDSA-AES256-SHA |

youngnick:
We reverted both PRs as part of the 1.12 release, we will bring them back in once #3292 is done, and look towards finishing up #2880 completely in the 1.13 timeframe.

sunjayBhatia:
[...]

PR #3237, opened 2021-01-05
internal/envoy: disabled rest of the CBC mode ciphers

tsaarni:
Disabled ECDHE-ECDSA-AES128-SHA and ECDHE-ECDSA-AES256-SHA which are marked
as weak ciphers by SSL Labs and RFC 7540.

Signed-off-by: Tero Saarni <tero.saarni@est.tech>

PR #3300, opened 2021-02-02
Revert "internal/envoy: disabled rest of the CBC mode ciphers (#3237)"

youngnick:
This reverts commit 88805b8af935900b734e8857a20fe0f320fb8946.

Sorry about this, @tsaarni, we'll replay this once we get a solution for #2880, since we have evidence other cipher deprecations have impacted at least one user in a big way (that's #3299).

Updates #3299


youngnick:
Getting this one sorted out so we can merge asap if we all agree, so we can not hold up 1.12.0 any longer.

tsaarni:
No problem!  Though it would be great to do #2880 soon since my organization needs to be compliant with [RFC7540 TLS 1.2 Cipher Suite Black List](https://tools.ietf.org/html/rfc7540#page-83) which lists these ciphers as having inadequate security.

I also wonder a bit about the asymmetry.  There is an equal set of ciphers flagged insecure for people that have RSA vs Elliptic Curve Cryptography based server certificates. These ciphers should go together. But I assume the idea is that the accident already happened for RSA versions and that is not wanted to be reverted? 

| cipher with RSA server certificate | cipher with EC server certificate |
| - | - |
| ~ECDHE-RSA-AES128-SHA~ | ~ECDHE-ECDSA-AES128-SHA~ |
| ECDHE-RSA-AES256-SHA | ECDHE-ECDSA-AES256-SHA |

sunjayBhatia:
@skriss @stevesloka per @tsaarni’s comments should we revert the ciphers for RSA certificates as well in this PR (I think we have edit access so we could push a commit) or just make a second

skriss:
> @skriss @stevesloka per @tsaarni’s comments should we revert the ciphers for RSA certificates as well in this PR (I think we have edit access so we could push a commit) or just make a second

I think we may as well revert the RSA ciphers too; indifferent between this PR or a second.

sunjayBhatia:
> > @skriss @stevesloka per @tsaarni’s comments should we revert the ciphers for RSA certificates as well in this PR (I think we have edit access so we could push a commit) or just make a second
> 
> I think we may as well revert the RSA ciphers too; indifferent between this PR or a second.

https://github.com/projectcontour/contour/pull/3301

sunjayBhatia (approved):

skriss (approved):

sunjayBhatia:
https://github.com/projectcontour/contour/pull/3304 will re-apply what was reverted here

PR #3301, opened 2021-02-02
Revert "Disable ECDHE-RSA-AES128-SHA and ECDHE-RSA-AES256-SHA ciphers (#3154)"

sunjayBhatia:
This reverts commit 2ccbaee653011f2247124073e79a2b3029708df3.

RSA ciphers version of to https://github.com/projectcontour/contour/pull/3300

Updates https://github.com/projectcontour/contour/issues/3299

PR #3154, opened 2020-11-25
internal/envoy: Disable ECDHE-RSA-AES128-SHA and ECDHE-RSA-AES256-SHA ciphers

bascht:
This change is basically the 2020 version of 18d23790 and will solve #2401, since those CBC ciphers get flagged by recent versions of Qualys SSL Labs or [TestSSL](https://testssl.sh/).

Refs also #2880 and #2777 but is way less complex.

Closes #2401.

PR #3304, opened 2021-02-02
Re-disable insecure AES128/256 ciphers by default

sunjayBhatia:
Re-applies #3237 and #3154
    
Blocked on #3300 and #3301 and should be merged for 1.13.0 release


youngnick (commented):
This LGTM, but I'd like to see #3292 go in first, just so we can be sure we can fix this later.

skriss:
LGTM. Are we going to do this for 1.13, or hold off for another release? (I think we talked about this but I can't recall the decision)

sunjayBhatia:
> LGTM. Are we going to do this for 1.13, or hold off for another release? (I think we talked about this but I can't recall the decision)

There was this comment which I think implied we should do it in 1.13: https://github.com/projectcontour/contour/issues/3299#issuecomment-774785433

skriss (approved):

youngnick (approved):
This LGTM, but we really need to figure out how we're going to document which ciphers are the default, and which are available in Envoy.

sunjayBhatia:
> This LGTM, but we really need to figure out how we're going to document which ciphers are the default, and which are available in Envoy.

Right now this section https://projectcontour.io/docs/main/configuration/#tls-configuration has a link to https://pkg.go.dev/github.com/projectcontour/contour/pkg/config#pkg-constants which will get updated to show the ciphers we allow when we cut a release (should actually be fixed to https://pkg.go.dev/github.com/projectcontour/contour/pkg/config#pkg-variables). It also links to https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto#envoy-v3-api-field-extensions-transport-sockets-tls-v3-tlsparameters-cipher-suites.

We could document these in-line as well and/or add another configuration/guide page for "Advanced TLS Configuration" that describes our cipher default selection and how to override it.

youngnick:
That sounds fine to me. I'll merge this one now.

PR #3260, opened 2021-01-14
internal: update HTTP/2 protocol options config

skriss:
Replaces the deprecated cluster Http2ProtocolOptions field
with TypedExtensionProtocolOptions.

Signed-off-by: Steve Kriss <krisss@vmware.com>

Fixes the lint errors that came up on `main` after upgrading go-control-plane to v0.9.8 (these didn't show up on the PR lint run, that's a separate issue).

https://github.com/projectcontour/contour/runs/1703439466

ref. https://www.envoyproxy.io/docs/envoy/v1.17.0/api-v3/extensions/upstreams/http/v3/http_protocol_options.proto#envoy-v3-api-msg-extensions-upstreams-http-v3-httpprotocoloptions

PR #3305, opened 2021-02-02
Revert "internal: update HTTP/2 protocol options config (#3260)"

skriss:
Reverts commit 78c434fcd9aa8e08f12ee5def7c0e215c0c805c5, which made
a Contour change that was not backwards-compatible with Envoy 1.16.
Also ignores lint errors which were the original motivation for #3260.


sunjayBhatia (approved):

youngnick (approved):
LGTM

PR #3310, opened 2021-02-03
site: Add contour serve flags documentation

### LINKED ISSUE #2615: expand configuration docs to cover all CLI commands, flags and env vars
Spun out from discussion in #2403.

Ideally, `site/docs/master/configuration.md` would cover all CLI flags, config file fields, and environment variables for all `contour` commands.

stevesloka:
Note: I didn't update the descriptions of the commands (I think some could be explained more/better).

Fixes #2615


site/docs/main/configuration.md
```diff
@@ -1,9 +1,55 @@
...
+- [Serve Flags](#serve-flags)
 - [Configuration File](#configuration-file)
 - [Environment Variables](#environment-variables)
 - [Bootstrap Config File](#bootstrap-config-file)
 
+## Serve Flags
+
+The `contour serve` command is the main command which is used to watch for Kubernetes resource and process them into Envoy configuration which is then streamed to any Envoy via its xDS gRPC connection.
+There are a number of flags that can be passed to this command which further configures how Contour operates. 
+Many of these flags are mirrored in the [Contour Configuration File](#configuration-file).
```
sunjayBhatia, line 12: Worth adding a note about the precedence order of CLI flags vs. the config file?

stevesloka: yes, good call!

stevesloka: I added a `Configuration` section at the top to explain the ordering.

sunjayBhatia (approved):

site/docs/main/configuration.md
```diff
@@ -1,9 +1,60 @@
...
+
+| Flag Name         | Description        |
+|-------------------|--------------------|
+| --config-path       | Path to base configuration |
+| --incluster         | Use in cluster configuration |
+| --kubeconfig=KUBECONFIG |    Path to kubeconfig (if not in running inside a cluster) |
+| --xds-address=XDS-ADDRESS | xDS gRPC API address |
```
youngnick, line 24: The all-caps here kind of suggests that these could be environment variables. Can I suggest a format like this?
```suggestion
| `--xds-address=<ipaddr>` | xDS gRPC API address |
```
where the `<ipaddr>` could be any type like `<int>` or `<string>`.

Also note the backticks, I think that having the flags fixed-width will help them stand out more.

site/docs/main/configuration.md
```diff
@@ -1,9 +1,60 @@
...
+| --xds-port=XDS-PORT       | xDS gRPC API port |
+| --stats-address=STATS-ADDRESS | Envoy /stats interface address |
+| --stats-port=STATS-PORT  |  Envoy /stats interface port |
+| --debug-http-address=DEBUG-HTTP-ADDRESS | Address the debug http endpoint will bind to. |
+| --debug-http-port=DEBUG-HTTP-PORT  | Port the debug http endpoint will bind to |
+| --http-address=HTTP-ADDRESS  | Address the metrics HTTP endpoint will bind to |
+| --http-port=HTTP-PORT  |    Port the metrics HTTP endpoint will bind to. |
+| --health-address=HEALTH-ADDRESS |   Address the health HTTP endpoint will bind to |
+| --health-port=HEALTH-PORT | Port the health HTTP endpoint will bind to |
+| --contour-cafile=CONTOUR-CAFILE | CA bundle file name for serving gRPC with TLS |
```
youngnick, line 34: If we change the format as I suggested for other things, we should reflect that these can be environment variables:
```suggestion
| --contour-cafile=<string|CONTOUR_CAFILE> | CA bundle file name for serving gRPC with TLS |
```
Or we could add another column "Environment variable" also.

youngnick (approved):
This is great. I had a couple of small change requests, but they shouldn't block this.

stevesloka:
@youngnick I updated with your comments, also made the serve.Flags match. Let me know if this is what you had in mind.

youngnick:
Looks like you missed signing a commit @stevesloka, want to fix that one up?

skriss:
needs a rebase

skriss (approved):

PR #3314, opened 2021-02-04
envoy listener exact balancer

iyacontrol:
envoy issues: [4602](https://github.com/envoyproxy/envoy/issues/4602)
envoy docs: [docs](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/listener.proto#envoy-api-msg-listener-connectionbalanceconfig)

When long keepalive connections, the listener should use the exact connection balancer. Otherwise, the cores of cpus are not balanced.


stevesloka:
@sunjayBhatia have you worked with this Envoy setting?

sunjayBhatia:
Not directly, no. I believe this feature is used for balancing connections on worker threads within Envoy so it might make some sense to expose for an environment that has many long lived connections since I guess you have some prior contextual knowledge the kernel does not that help in scheduling. From reading the issue this was useful to help solve some specific tail latency issues.

@iyacontrol I think before accepting this change we would want some more context on why, what this solves for the submitter, what are the symptoms, reproduction maybe, etc. as well as testing, docs, etc.

youngnick:
Yes, thanks for this PR, but in general, we like an issue to document the problem before accepting a PR to fix that problem.

In this case, I'd like to know what the tradeoffs are. Should we just be using the exact balancer all the time? In general, we prefer not to add a configuration setting unless there is a strong need.

iyacontrol:
thanks @stevesloka @sunjayBhatia @youngnick !

1: Add the parameter settings of envoy's listener, which can meet the needs of more people in more scenarios.
2: [Envoy threading model](https://blog.envoyproxy.io/envoy-threading-model-a8d44b922310) . The existing architecture will not work well if Envoy is deployed in a scenario in which there are very few connections that require substantial resources to handle. This is because there is no guarantee that connections will be evenly spread between workers. This can be solved by implementing worker connection balancing in which a worker is capable of forwarding a connection to another worker for handling.

We use grpc extensively in our company. I have been using the exact balancer in our production environment for a month，the effect is very good.

iyacontrol:
> Not directly, no. I believe this feature is used for balancing connections on worker threads within Envoy so it might make some sense to expose for an environment that has many long lived connections since I guess you have some prior contextual knowledge the kernel does not that help in scheduling. From reading the issue this was useful to help solve some specific tail latency issues.
> 
> @iyacontrol I think before accepting this change we would want some more context on why, what this solves for the submitter, what are the symptoms, reproduction maybe, etc. as well as testing, docs, etc.

i will  add some docs.

iyacontrol:
#3331

youngnick:
Thanks for this PR @iyacontrol, and thanks for being patient while I clarified what we're doing here.

I agree that we should add the configurability, and it should be a config-file item, but I'm not sure `listeners` is the right place to put it. I'll do a proper review and raise this there though.

pkg/config/parameters.go
```diff
@@ -383,6 +383,15 @@ type NetworkParameters struct {
 	XffNumTrustedHops uint32 `yaml:"num-trusted-hops"`
 }
 
+// ListenerParameters hold various configurable listener values.
+type ListenerParameters struct {
```
youngnick, line 387: Feels like these are really upstream settings, not listener settings. How about we call it something like `upstream`, or `upstreamConfig` or similar? @skriss @stevesloka @sunjayBhatia, any thoughts?

youngnick (changes requested):
I think we need to make sure we have the right name for the config item, once we put it in, it's very difficult to change.

Also, any chance we could get a test in `internal/xdscache` somewhere?

pkg/config/parameters.go
sunjayBhatia: I think the balancing is done on which Envoy worker thread will accept and do work on a connection, I would say tying "upstream" into the naming makes this seem like an upstream load balancing parameter which it isn't

sunjayBhatia: Maybe remove "Use" and have the name be something to the effect of "evenly distribute accepted connections between Envoy workers"

iyacontrol: let us read the envoy docs. there is some params:

 ```
 "name": "...",
  "address": "{...}",
  "filter_chains": [],
  "use_original_dst": "{...}",
  "per_connection_buffer_limit_bytes": "{...}",
  "metadata": "{...}",
  "drain_type": "...",
  "listener_filters": [],
  "listener_filters_timeout": "{...}",
  "continue_on_listener_filters_timeout": "...",
  "transparent": "{...}",
  "freebind": "{...}",
  "socket_options": [],
  "tcp_fast_open_queue_length": "{...}",
  "traffic_direction": "...",
  "udp_listener_config": "{...}",
  "api_listener": "{...}",
  "connection_balance_config": "{...}",
  "reuse_port": "...",
  "access_log": []
}
```

Exact-balance is one of the parameters, I think contour may be added to support other parameter settings in the future.

So i think  the listenterParams is appropriate

Comment (youngnick) [reply pkg/config/parameters.go:427]: That makes sense, I guess we're adding `listener` then.

I think we should call the inner parameter `connectionBalancer` and have the only valid value be `exact`. Other values should be ignored, with an error logged. Putting the wrong value should not be fatal.

Yes, I know, this is a more complicated way to do what's a boolean anyway, but this way it's more descriptive, shorter, and allows for more flexibility in the future.

So, the stanza in YAML should look like:
```yaml
listener:
  connectionBalancer: exact
```
What do you think? @sunjayBhatia @iyacontrol?

pkg/config/parameters.go
```diff
@@ -423,6 +423,15 @@ type NetworkParameters struct {
...
+	// UseExactConnectionBalancer. If specified, the listener will use the exact connection balancer
+
+	// See https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/listener.proto#envoy-api-msg-listener-connectionbalanceconfig
+	// for more information.
+	UseExactConnectionBalancer bool `yaml:"use-exact-connection-balancer"`
```
youngnick, line 432: As I commented earlier, I think this should be called `ConnectionBalancer`, and be a string where the only valid value is `exact`. The YAML tag should be `connectionBalancer`.

youngnick (changes requested):
I agree that `listener` is the right top-level name, but in that case, I'd like to see the actual field name a little more clear.

pkg/config/parameters.go
iyacontrol: thanks @youngnick .

i agree with you.

pkg/config/parameters.go
```diff
@@ -423,6 +423,15 @@ type NetworkParameters struct {
 	XffNumTrustedHops uint32 `yaml:"num-trusted-hops"`
 }
 
+// ListenerParameters hold various configurable listener values.
+type ListenerParameters struct {
+	// ConnectionBalancer. If the value is exact, the listener will use the exact connection balancer
+
+	// See https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/listener.proto#envoy-api-msg-listener-connectionbalanceconfig
+	// for more information.
+	ConnectionBalancer string `yaml:"connection-balancer"`
```
youngnick, line 432: Nit: I generally prefer k8s-style `camelCase` YAML names, but I see we have plenty of hyphenated ones in our config filie. So this is a small nit, I'll leave it up to other maintainers to make a call.

youngnick (approved):
LGTM aside from one small nonblocking nit.

pkg/config/parameters.go
```diff
@@ -423,6 +423,15 @@ type NetworkParameters struct {
 	XffNumTrustedHops uint32 `yaml:"num-trusted-hops"`
 }
 
+// ListenerParameters hold various configurable listener values.
+type ListenerParameters struct {
+	// ConnectionBalancer. If the value is exact, the listener will use the exact connection balancer
+
```
sunjayBhatia, line 429: nit: Remove this extra line

cmd/contour/serve.go
```diff
@@ -272,6 +272,12 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		return fmt.Errorf("error parsing request timeout: %w", err)
 	}
 
+	// connection balancer
+	if ok := ctx.Config.Listener.ConnectionBalancer == "exact" || ctx.Config.Listener.ConnectionBalancer == ""; !ok {
+		log.Warnf("invalid connection balancer value %s. the connection balancer only support exact for now.", ctx.Config.Listener.ConnectionBalancer)
```
sunjayBhatia, line 277:
```suggestion
		log.Warnf("Invalid listener connection balancer value %q. Only "exact" connection balancing is supported for now.", ctx.Config.Listener.ConnectionBalancer)
```

sunjayBhatia (changes requested):
Lookd good overall, just a few minor nits and we probably still need some documentation added here: https://github.com/projectcontour/contour/blob/1afeff96f27ec0dfe75746194dbf08da66a1182d/site/docs/main/configuration.md#configuration-file (a listener entry at the top level of the config description table and a sub section describing the configurable listener fields)

cmd/contour/serve.go
iyacontrol: done it.

iyacontrol:
> Lookd good overall, just a few minor nits and we probably still need some documentation added here: https://github.com/projectcontour/contour/blob/1afeff96f27ec0dfe75746194dbf08da66a1182d/site/docs/main/configuration.md#configuration-file (a listener entry at the top level of the config description table and a sub section describing the configurable listener fields)

hi @sunjayBhatia  i will add docs.

pkg/config/parameters.go
iyacontrol: done it

iyacontrol:
hi @sunjayBhatia can you resolve the conflicts ？

sunjayBhatia:
> hi @sunjayBhatia can you resolve the conflicts ？

You’ll have to merge the main branch with your PR branch and fix the merge conflicts

iyacontrol:
> fix the merge conflicts

done it. hi @sunjayBhatia can you approval the pr.

sunjayBhatia:
> > Lookd good overall, just a few minor nits and we probably still need some documentation added here: https://github.com/projectcontour/contour/blob/1afeff96f27ec0dfe75746194dbf08da66a1182d/site/docs/main/configuration.md#configuration-file (a listener entry at the top level of the config description table and a sub section describing the configurable listener fields)
> 
> hi @sunjayBhatia i will add docs after pr is merged.

I think we would usually like to keep the docs and code change together as one PR for consistency, also to ensure if we cut a release we have features documented already

iyacontrol:
> > > Lookd good overall, just a few minor nits and we probably still need some documentation added here: https://github.com/projectcontour/contour/blob/1afeff96f27ec0dfe75746194dbf08da66a1182d/site/docs/main/configuration.md#configuration-file (a listener entry at the top level of the config description table and a sub section describing the configurable listener fields)
> > 
> > 
> > hi @sunjayBhatia i will add docs after pr is merged.
> 
> I think we would usually like to keep the docs and code change together as one PR for consistency, also to ensure if we cut a release we have features documented already

the docs has been added.

site/docs/main/configuration.md
```diff
@@ -172,6 +173,16 @@ The network configuration block can be used to configure various parameters netw
 {: class="table thead-dark table-bordered"}
 <br>
 
+### Listener Configuration
+
+The listener configuration block can be used to configure various parameters for Envoy listener.
+
+| Field Name | Type| Default  | Description |
+|------------|-----|----------|-------------|
+| connection-balancer | string | `""` | This field specifies the listener connection balancer. If the value is `exact`, the listener will use the exact connection balancer. See [the Envoy documentation][14] for more information. |
```
youngnick, line 182:
```suggestion
| connection-balancer | string | `""` | This field specifies the listener connection balancer. If the value is `exact`, the listener will use the exact connection balancer to balance connections between threads in a single Envoy process. See [the Envoy documentation][14] for more information. |
```

Small suggestion to make it clearer that this is Envoy-internal, and not for load-balancing to upstreams/backends.

youngnick (approved):

site/docs/main/configuration.md
iyacontrol: done it. hi @youngnick @sunjayBhatia  maybe this pr can be merged. Any other questions？

sunjayBhatia (approved):

youngnick:
This is all good to go, @iyacontrol, except that the commits need a DCO signoff (a `git commit -s`), then we can get this one merged. If you can do this in the next twenty four hours, we can get this into Contour 1.13.

iyacontrol:
> This is all good to go, @iyacontrol, except that the commits need a DCO signoff (a `git commit -s`), then we can get this one merged. If you can do this in the next twenty four hours, we can get this into Contour 1.13.

Hi @youngnick  All checks have passed.

PR #3316, opened 2021-02-04
internal/envoy: Expose Envoy's delayed_close_timeout

### LINKED ISSUE #3285: Expose Envoy's 'delayed_close_timeout' in the Contour configuration

It is documented that envoy's "delayed_close_timeout" can close connections to the client even while there is data remaining to be written (https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/network/http_connection_manager/v2/http_connection_manager.proto#config-filter-network-http-connection-manager-v2-httpconnectionmanager).  It seems that this can happen to clients that set the "Connection: close" header and are slow to read the response. This can be replicated simply by returning a large file from an nginx container and using curl's `--limit-rate` parameter to slow down the transfer to the client. 

Here's an example setup:
```
---
kind: HTTPProxy
metadata:
  name: nginx-test-httpproxy
spec:
  routes:
  - conditions:
    - prefix: /
    timeoutPolicy:
      response: 4800s
      idle: 4800s
    services:
      - name: nginx-test-svc
        port: 80
    permitInsecure: true
  virtualhost:
    fqdn: nginx-test.instakube.io
---
apiVersion: v1
kind: Service
metadata:
  labels:
    app: nginx-test
  name: nginx-test-svc
spec:
  ports:
  - port: 80
    protocol: TCP
    targetPort: 80
  selector:
    app: nginx-test
  sessionAffinity: None
  type: ClusterIP
---
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: nginx-test
  name: nginx-test
spec:
  replicas: 1
  selector:
    matchLabels:
      app: nginx-test
  template:
    metadata:
      labels:
        app: nginx-test
    spec:
      containers:
      - image: nginx:latest
        name: nginx-test
        volumeMounts:
        - mountPath: /usr/share/nginx/html
          name: html
      volumes:
      - name: html
        hostPath:
          path: /html
```

And here's a curl command which will trigger the problem
```
curl -H "Connection: close" --limit-rate 1800K http://nginx-test.instakube.io/verylargefile.txt -vvv >/dev/null
```

This will result in the following error:

```
curl: (18) transfer closed with 10411 bytes remaining to read
``` 

When that request completes the envoy debug logs show:

```
[2021-01-18 23:16:51.079][19][debug][connection] [source/common/network/connection_impl.cc:112] [C243] closing data_to_write=859667 type=2
[2021-01-18 23:16:51.079][19][debug][connection] [source/common/network/connection_impl_base.cc:30] [C243] setting delayed close timer with timeout 1000 ms
[2021-01-18 23:16:52.043][19][debug][connection] [source/common/network/connection_impl_base.cc:44] [C243] triggered delayed close
[2021-01-18 23:16:52.043][19][debug][connection] [source/common/network/connection_impl.cc:208] [C243] closing socket: 1
[2021-01-18 23:16:52.043][19][debug][conn_handler] [source/server/connection_handler_impl.cc:111] [C243] adding to cleanup list
```

However when the "Connection: close" header is omitted from the request the envoy debug logs look like:

```
[2021-01-18 23:31:46.692][16][debug][connection] [source/common/network/connection_impl.cc:584] [C16] remote close
[2021-01-18 23:31:46.692][16][debug][connection] [source/common/network/connection_impl.cc:208] [C16] closing socket: 0
[2021-01-18 23:31:46.693][16][debug][conn_handler] [source/server/connection_handler_impl.cc:111] [C16] adding to cleanup list
 ```

Those messages about the delayed close seem to hint that envoy's delayed_close_timeout might be involved in the premature closing of the connection.  I have experimented a bit with it and found that either setting `delayed_close_timeout` to '0' or increasing it prevents this early connection closing problem. 

Envoy has a default of 1s (1000ms) for the delayed_close_timeout and since it seems like that default might not be ideal in all circumstances I think it would be beneficial to expose it in the contour configuration, similar to the way stream_idle_timeout, max_connection_duration etc. can be configured. 

There is some good discussion around this timeout in https://github.com/envoyproxy/envoy/issues/6616 and it possibly makes the case (although by no means strongly) for setting it to zero. I don't know what the universally correct value for this is so I'm inclined to request that it be configurable.

Since I've already experimented with configuring this I'm happy to submit a PR with the implementation.

xtreme-jesse-malone:
There are situations where Envoy's "delayed_close_timeout"
can close connections to the client even while there is data
remaining to be written. This can happen when a client sets the
"Connection: close" header and is slow to read the response.
Users who encounter this may need to adjust the 'delayed_close_timeout'.

This commit will expose delayed_close_timeout, making it
configurable via the global contour configuration.

Fixes #3285

Signed-off-by: Jesse Malone <jmalone@vmware.com>


xtreme-jesse-malone:
I've realized there's a need for additional tests in `xdscache/v3/listener_test.go`. I will add those and update this.

xtreme-jesse-malone:
Integration tests are failing but it looks like a known problem. (https://github.com/projectcontour/contour/issues/3320)

youngnick (approved):
LGTM, thanks very much @xtreme-jesse-malone, this is great work, particularly the documentation of the setting and why you shouldn't change it unless you need it.

skriss (approved):
LGTM, thanks @xtreme-jesse-malone!

PR #3317, opened 2021-02-04
site: Update WeightedLeastRequestDescription in docs

stevesloka:
The 'WeightedLeastRequest' load balancer strategy in the docs outlined some text that it wasn't useful outside of just testing. This is due to copying from the Envoy docs (v1.5) when this algorithm was new and has evolved much since that time, we just never updated the docs to match.

This came up on the Community call today (@moderation). 

v1.17 docs: https://www.envoyproxy.io/docs/envoy/v1.17.0/intro/arch_overview/upstream/load_balancing/load_balancers#weighted-least-request
Old docs: https://www.envoyproxy.io/docs/envoy/v1.6.0/intro/arch_overview/load_balancing.html#weighted-least-request

### DIFF
--- a/site/docs/main/config/request-routing.md
+++ b/site/docs/main/config/request-routing.md
@@ -223,7 +223,7 @@ Each route can have a load balancing strategy applied to determine which of its 
 The following list are the options available to choose from:
 
 - `RoundRobin`: Each healthy upstream Endpoint is selected in round robin order (Default strategy if none selected).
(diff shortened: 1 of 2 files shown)


sunjayBhatia (approved):
LGTM, should we bother to backport the change to more release versions? Given main is what is rendered by default I would say maybe not?

stevesloka:
> LGTM, should we bother to backport the change to more release versions? Given main is what is rendered by default I would say maybe not?

The latest release is what's rendered as default (e.g. v1.12.0) which is why I did that plus `main`.

skriss (approved):
LGTM, I think main + v1.12 is sufficient.

skriss:
sigh, looks like my external name integration test is failing now, but that's definitely not related to this PR, so I'll merge as-is.

moderation:
Awesome. Thanks for the update

PR #3321, opened 2021-02-04
internal: update HTTP/2 protocol options config

### LINKED ISSUE #3308: Replace deprecated Http2ProtocolOptions with TypedExtensionProtocolOptions
Revert https://github.com/projectcontour/contour/pull/3305, which itself reverted https://github.com/projectcontour/contour/pull/3260, to replace this deprecated Envoy config option.

Now that Contour 1.12 has shipped with Envoy 1.17, it should be safe to include this change in Contour 1.13 since upgrading users would already be on Envoy 1.17 so the Contour change is backwards-compatible with a 1.12 installation.

skriss:
Replaces the deprecated cluster Http2ProtocolOptions field
with TypedExtensionProtocolOptions.

Closes #3308.
Updates #3260.
Updates #3305.


skriss:
This should be safe to do in 1.13 because we only support upgrading from the previous version, 1.12, which uses Envoy 1.17, which supports this new config field, so this Contour change should be backwards-compatible. Please check me on this :)

sunjayBhatia (approved):

youngnick (approved):
LGTM, and yes, I think that we're safe to move to this now everyone should be on Envoy 1.17. Maybe a release-note label to make sure we mention this in the next release?

PR #3339, opened 2021-02-09
_integration: improve local rate limit test robustness

skriss:
Makes the local rate limit integration test less flaky by
first creating an HTTPProxy without a rate limit, and waiting
until it serves 200's before adding a rate limit and counting
allowed requests. This replaces the previous method of waiting
for the upstream service to have endpoints, which was less
reliable.

### DIFF
--- a/_integration/testsuite/httpproxy/019-local-rate-limiting.yaml
+++ b/_integration/testsuite/httpproxy/019-local-rate-limiting.yaml
@@ -37,7 +37,48 @@ $apply:
 
 ---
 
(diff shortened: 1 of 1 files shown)


sunjayBhatia (approved):
LGTM, sounds like a reasonable approach to make sure apps are up etc. before making requests we expect to count against the rate limit

skriss:
A different integration test (006) flaked the first time, but passed the second time. The local rate limiting test passed on both runs and has passed reliably for me locally.

youngnick (approved):

PR #3343, opened 2021-02-10
Add HTTPProxy FQDN Validation

### LINKED ISSUE #3166: HTTPProxy FQDN Validation

During the creation of new HTTPProxy resources, FQDNs are not validated resulting that invalid urls such as `my-domain-${somevariable}.domain.tld` are passed to envoy but the HTTPProxy is not created. Could the CRD definition be edited to include DNS1123 validation similar to [apimachinery](https://github.com/kubernetes/apimachinery/blob/2456ebdaba229616fab2161a615148884b46644b/pkg/util/validation/validation.go#L78)  to check that the FQDN does not container invalid characters?

abhide:
Fixes #3166

### DIFF
--- a/internal/dag/httpproxy_processor.go
+++ b/internal/dag/httpproxy_processor.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/validation"
 )
 
 // defaultExtensionRef populates the unset fields in ref with default values.
@@ -102,6 +103,28 @@ func (p *HTTPProxyProcessor) Run(dag *DAG, source *KubernetesCache) {
 	}
 }
 
+// Credits to https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L95
+// Had to rewrite it here as we don't need FieldPath
+func (p *HTTPProxyProcessor) isValidFQDN(name string) []string {
+	allErrs := []string{}
+	if strings.HasSuffix(name, ".") {
+		name = name[:len(name)-1]
+	}
+	if errs := validation.IsDNS1123Subdomain(name); len(errs) > 0 {
+		allErrs = append(allErrs, errs...)
+	}
+	if len(strings.Split(name, ".")) < 2 {
+		allErrs = append(allErrs, "domain should have at least 2 segments separated by dots")
+	}
+
+	for _, label := range strings.Split(name, ".") {
+		if errs := validation.IsDNS1123Label(label); len(errs) > 0 {
+			allErrs = append(allErrs, errs...)
+		}
+	}
+	return allErrs
+}
+
 func (p *HTTPProxyProcessor) computeHTTPProxy(proxy *contour_api_v1.HTTPProxy) {
 	pa, commit := p.dag.StatusCache.ProxyAccessor(proxy)
 	validCond := pa.ConditionFor(status.ValidCondition)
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -6,23 +6,23 @@ require (
...
-	github.com/stretchr/testify v1.5.1
+	github.com/stretchr/testify v1.6.1
 	google.golang.org/genproto v0.0.0-20200526211855-cb27e3aa2013
 	google.golang.org/grpc v1.27.0
 	google.golang.org/protobuf v1.25.0
 	gopkg.in/alecthomas/kingpin.v2 v2.2.6
 	gopkg.in/yaml.v2 v2.3.0
 	k8s.io/api v0.19.2
-	k8s.io/apimachinery v0.19.2
+	k8s.io/apimachinery v0.20.2
```
abhide, line 23: Will separate ticket and send out a separate PR for go mod dependencies update

abhide:
@skriss / @stevesloka Is this the correct place for adding the validation? Seems like there are lot of test failures with this validation.

internal/dag/status_test.go
```diff
@@ -2369,4 +2369,36 @@ func TestDAGStatus(t *testing.T) {
...
+					Port: 8080,
+				}},
+			}},
+		},
+	}
+
+	run(t, "invalid FQDN", testcase{
+		objs: []interface{}{invalidFQDN, fixture.ServiceRootsHome},
+		want: map[types.NamespacedName]contour_api_v1.DetailedCondition{
+			{Name: invalidFQDN.Name, Namespace: invalidFQDN.Namespace}: fixture.NewValidCondition().
```
abhide, line 2398: intentional! want it to fail. Will fix as part of removing WIP

internal/dag/httpproxy_processor.go
```diff
@@ -120,6 +121,13 @@ func (p *HTTPProxyProcessor) computeHTTPProxy(proxy *contour_api_v1.HTTPProxy) {
 			"Spec.VirtualHost.Fqdn must be specified")
 		return
 	}
+
+	// Check if FQDN is conformant according to RFC1123
+	if errs := validation.IsDNS1123Label(host); len(errs) != 0 {
```
youngnick, line 126: The reason you're getting a lot of test failures here is that a DNS label is a single *component* of a hostname.

So, in `subdomain.foo.com`, `subdomain`, `foo`, and `com` are all labels. A period (`.`) is not permitted in a DNS label.

I'd suggest using `IsFullyQualifiedDomainName` instead, however, if it fails, you also need to check for a single DNS label (which we accept as valid).

So, something like:
```suggestion
	if errs := validation.IsFullyQualifiedDomainName(host); len(errs) != 0 {
	  if labelErr := validation.IsDNS1123Label; labelErr != nil {
	    // add the error
```
I think something like that should help clear up your errors, and then you just need to make sure that you're testing both some fully qualified names (like `subdomain.foo.com`, `foo.com`), and a single label (like `foo`).

youngnick (commented):
I think you've run afoul of RFC-speak here, specifically about the difference between a *label* and a *hostname*.

internal/dag/httpproxy_processor.go
abhide: Thanks @youngnick . FIxed in  latest patch.

internal/dag/status_test.go
abhide: fixed

internal/dag/httpproxy_processor.go
```diff
@@ -102,6 +103,28 @@ func (p *HTTPProxyProcessor) Run(dag *DAG, source *KubernetesCache) {
 	}
 }
 
+// Credits to https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L95
```
abhide, line 106: Is there a better way to use the original one from apimachinery?

internal/dag/httpproxy_processor.go
```diff
@@ -102,6 +103,28 @@ func (p *HTTPProxyProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+// Had to rewrite it here as we don't need FieldPath
+func (p *HTTPProxyProcessor) isValidFQDN(name string) []string {
```
youngnick, line 108: This looks great! Unfortunately, because of Kube things, we actually need to allow a single DNS label as an option as well. I'm sorry not to have made that clear earlier. So, the check about having at least 2 segments can't be added, as it's an API change.

So, this function should check:
Is this a valid DNS subdomain (using the validation check)?
Are all the labels valid (one is a valid number of labels)?
If not, it's not valid.

For this new function, I'd also like to see unit tests that check each if statement adds the correct errors.

youngnick (changes requested):
This is really close @abhide! Just a couple of things about the expected behavior, and a little more testing, and it will be done.

PR #3345, opened 2021-02-10
hack: check API ref docs for uncommitted codegen

### LINKED ISSUE #3344: CI doesn't fail when API reference docs haven't been updated
PR #3288 updated some API godoc but didn't run `make codegen` to update the API reference docs. However, this wasn't flagged as an error by CI that checks for uncommitted codegen.

It looks like the [script that checks for changes](https://github.com/projectcontour/contour/blob/main/hack/actions/check-uncommitted-codegen.sh#L10-L15) isn't looking at the API reference docs. We should be able to directly add the file to the list.

skriss:
Adds the API reference docs file to the list of targets
to check for uncommitted codegen.

Closes #3344.

### DIFF
--- a/hack/actions/check-uncommitted-codegen.sh
+++ b/hack/actions/check-uncommitted-codegen.sh
@@ -7,16 +7,17 @@ set -o pipefail
 readonly HERE=$(cd $(dirname $0) && pwd)
 readonly REPO=$(cd ${HERE}/../.. && pwd)
 
-declare -r -a DIRS=(
+declare -r -a TARGETS=(
 	${REPO}/apis
 	${REPO}/site/_metrics
 	${REPO}/examples/render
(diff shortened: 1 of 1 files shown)


skriss:
This now fails as expected (https://github.com/projectcontour/contour/pull/3345/checks?check_run_id=1873055918) since the API ref docs are currently out-of-date.

Once https://github.com/projectcontour/contour/pull/3346 merges, this should pass.

sunjayBhatia (approved):
As far as I can tell, this should now capture the output of all the `make generate*` targets

skriss:
Passing now that the API docs have been updated: https://github.com/projectcontour/contour/pull/3345/checks?check_run_id=1874593720

youngnick (approved):
LGTM, nice catch.

Issue #2878, opened 2020-09-03
Does Contour support FIPS-allowed encryption? 


javsalgar:
**Please describe the problem you have**

Some applications like Elasticsearch have a FIPS mode that forces the application to use FIPS-compliant encryption mechanisms (https://www.elastic.co/guide/en/elasticsearch/reference/current/fips-140-compliance.html). I didn't see here (https://projectcontour.io/guides/tls/) if it is possible to allow something similar to Contour. Looking a bit further it seems that Envoy does allow it (https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/security/ssl#fips-140-2), could you confirm if that would be enough or it is necessary to enable anything else on the Contour side.

jpeach:
If you bring a FIPS-compliant Envoy image to your deployment, then everything should work fine. The only potential issue that I can think of is whether the cipher suite that Contour configures is allowed in FIPS (I suspect that it should be OK).

sunjayBhatia:
See https://github.com/projectcontour/contour/pull/3347

youngnick:
I'm not sure (I'm no FIPS expert), but it looks like in order to be able to have a FIPS-compliant Contour install, you need three things:
- A FIPS compliant Envoy. Looking around, it seems likely that if you're doing FIPS, you'll need to do this yourself. Envoy itself doesn't publish a FIPS-compliant container (see envoyproxy/envoy#8107), nor can you use `getenvoy` to do it (see tetratelabs/getenvoy-package#20).
- Contour must be compiled using a FIPS-compliant Go version. Currently we don't provide this, but it looks like we may be able to provide an alternate Docker file (and possibly image) by using the `dev.boringcrypto` Go branch, which has a Docker build image availabel at `goboring/golang`.
- Contour must be able to be configured to only use FIPS-compliant TLS ciphers. This is enabled by #3292.

To wrap this up, we will probably need a guide in the website guides section for what to do to end up with a FIPS-compliant Contour, and possibly have something in the Operator eventually.

sunjayBhatia:
I can write up a guide to start with, depending on how that works we can assess if we want to provide more resources (scripts, docker images, etc.)

sunjayBhatia:
Or @youngnick if you want go for it since you self assigned!

youngnick:
No, I'm happy for you to take this one @sunjayBhatia!

sunjayBhatia:
[...]

PR #3347, opened 2021-02-10
internal/envoy: Disable CHACHA20 ciphers

sunjayBhatia:
Related to an ask that Contour is FIPS compliant by default. Envoy built with
BoringSSL-FIPS removes the CHACHA20 ciphers by default.

Fixes https://github.com/projectcontour/contour/issues/2878

Note: Should probably merge after https://github.com/projectcontour/contour/pull/3292


sunjayBhatia:
Open question, should we do this in 1.13 since we plan for https://github.com/projectcontour/contour/pull/3292 to be done then too? Or wait for 1.14 to give people some time to adapt?

xaleeks:
I vote we announce this in the 1.13 and perform the deprecation on the following release, so folks have some time to react.  We should release this concurrent to or after the FIPS configurability capability is released so people can customize the specific list of ciphers they need.

jpeach:
There's a good use case for an easy FIPS-mode, but I don't think that FIPS is an end in itself or necessarily a good default in general. AFAIK CHACHA20 cipher suites are valid, reasonable and secure choices. 

https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility

youngnick:
Yes, I agree that FIPS should not be Contour's default. We should make it *possible* to do FIPS, and document what you have to do, but I don't think that FIPS should be the default.

I agree that we shouldn't turn these ciphers off by default. When you are doing FIPS, you're going to be specifying a cipher set anyway, so it's better to just document what that will be, and make sure it works.

xaleeks:
FIPS is too much for default for OSS I agree.  So how do we provide a mechanism that allows FIPS to be kept off in OSS but the default in TKG, without having to maintain two code bases.  So can we leverage the configurable cipher suites capability to not only provide granularity at the cipher suite level but also just turn FIPS on / off as a profile.  

I'm kinda just throwing ideas out at this point.  Beware that the FIPS requirement is a long term thing for all future releases within TKG.  So maintaining two Contours would be very heavy

youngnick:
In terms of actual FIPS compliance, it seems to me like there are two requirements:
- Cipher suite must only have FIPS ciphers (basically, something like this PR)
- Go crypto libraries must only use FIPS-certified versions. We can actually do this with a separte Dockerfile using `FROM goboring/golang:<version>` instead of the current Dockerfile. So we may be able to make this relatively easy.

Then, we do not have to maintain two codebases anywhere. But we should talk more about this on #2878.

Thanks for this work, @sunjayBhatia, but I think we should close this PR out.

sunjayBhatia:
Sounds good to me I mostly wanted to start the conversation and capture opinions etc. I figured there would be some reticence to pulling more ciphers out etc.! I think your comments on this and the linked FIPS issue are spot on

PR #3358, opened 2021-02-12
Apply correct image pull policy when running ingress conformance

sunjayBhatia:
Running the conformance tests should work exactly the integration tests do: a locally built contour image is loaded in the kind cluster to test against. Without the change to image pull policy we end up always running tests against the `main` image pulled from docker hub.


youngnick:
When we run the conformance test, how does the image pulling work? Do we generate a new image and push it into the kind cluster (tagged `main`)? If so, I think this change makes sense. Just want to check for posterity. :)

sunjayBhatia:
> When we run the conformance test, how does the image pulling work? Do we generate a new image and push it into the kind cluster (tagged `main`)? If so, I think this change makes sense. Just want to check for posterity. :)

Yep, running the conformance tests should work exactly like that which is how the integration tests work. Without the change to image pull policy we end up pulling from docker hub.

stevesloka (approved):
Looks fine to me, just hard to test since these aren't enabled at the moment.

youngnick (approved):
This makes sense now, thanks @sunjayBhatia!

Issue #3054, opened 2020-10-22
Add metrics for DAG changes


jpeach:
**Please describe the problem you have**

Add counter metrics to capture the rate of DAG change; we should count the number of times the DAG is rebuilt (and also the number of times the DAG actually changed if we can figure that out)

youngnick:
I thought we were doing this already - maybe we have some stats about the DAG rebuild _time_?

PR #3361, opened 2021-02-14
Add Counter to record DAG Rebuilds

abhide:
Prometheus Counter to record total number of times DAG got rebuilt
`contour_dagrebuild_total`

Fixes: #3054


internal/metrics/metrics.go
```diff
@@ -127,6 +129,12 @@ func NewMetrics(registry *prometheus.Registry) *Metrics {
 			},
 			[]string{},
 		),
+		dagRebuildTotal: prometheus.NewCounter(
+			prometheus.CounterOpts{
+				Name: DAGRebuildTotal,
+				Help: "Total number of times DAG is rebuilt",
```
youngnick, line 135:
```suggestion
				Help: "Total number of times DAG has been rebuilt since startup",
```

site/_metrics/contour_dagrebuild_total.md
```diff
@@ -0,0 +1,7 @@
+---
+name: 'contour_dagrebuild_total'
+type: '[COUNTER](https://prometheus.io/docs/concepts/metric_types/#counter)'
+labels: ''
+---
+
+Total number of times DAG is rebuilt
```
youngnick, line 7:
```suggestion
Total number of times DAG has been rebuilt since startup
```

youngnick (approved):
LGTM, nice work.

internal/metrics/metrics.go
abhide: Fixed.

site/_metrics/contour_dagrebuild_total.md
abhide: Fixed

internal/metrics/metrics.go
```diff
@@ -185,6 +187,12 @@ func NewMetrics(registry *prometheus.Registry) *Metrics {
 			},
 			[]string{},
 		),
+		dagRebuildTotal: prometheus.NewCounter(
+			prometheus.CounterOpts{
+				Name: DAGRebuildTotal,
+				Help: "Total number of times DAG has been rebuilt since startup",
```
stevesloka, line 193: Do you think we should add a startup timestamp as well? This would be unique to the instance of the Contour that started.

youngnick: I think that's a pretty good idea, but we could do it in a followup.

sunjayBhatia (approved):
LGTM

I think adding the contour startup time would be useful, though also may be captured well enough with the additional metadata around pod ids that get recycled with restarts. Could be done here or in a second PR

PR #3370, opened 2021-02-17
dep: Add dep to new gateway-api reference.

stevesloka:
Service APIs have been renamed to Gateway API. This adds the new release version to Gatway API to allow refactoring to the new name.

Updates #3349

Signed-off-by: Steve Sloka <slokas@vmware.com>

PR #3374, opened 2021-02-18
Update the controller-runtime version to allow GatewayAPI dep to be bumped

stevesloka:
Updates #3370


youngnick (approved):

cmd/contour/serve.go
```diff
@@ -187,6 +187,10 @@ func validateCRDs(dynamicClient dynamic.Interface, log logrus.FieldLogger) {
 
 // doServe runs the contour serve subcommand.
 func doServe(log logrus.FieldLogger, ctx *serveContext) error {
+	ectx := context.Background()
+	serveContext, cancel := context.WithCancel(ectx)
```
sunjayBhatia, line 191: maybe consider renaming `serveContext` since it sorta conflicts with the other `ctx *serveContext` parameter (which is also a bit misleading since I don't think it implements `context.Context`)

sunjayBhatia: We could also rename the `serveContext` type and `ctx` param since we have actual `context.Context`s involved here.

Another option would be to make `serveContext` implement `context.Context` but that might be overkill

cmd/contour/serve.go
```diff
@@ -187,6 +187,10 @@ func validateCRDs(dynamicClient dynamic.Interface, log logrus.FieldLogger) {
...
+	defer cancel()
```
skriss, line 192: I don't think you need to manage cancellation explicitly here, since the work group should already do that for us; I think we just need to switch a few of the tasks to use `g.AddContext(...)` instead of `g.Add(...)`

cmd/contour/serve.go
```diff
@@ -448,7 +452,7 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		log.Info("starting informers")
 		defer log.Println("stopped informers")
 
-		if err := clients.StartInformers(stop); err != nil {
+		if err := clients.StartInformers(serveContext); err != nil {
```
skriss, line 455: If you switch to `g.AddContext(...)` for this task, then the workgroup will pass in a context, and will manage cancellation of that context at the appropriate time.

cmd/contour/serve.go
```diff
@@ -578,7 +582,7 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		log := log.WithField("context", "xds")
 
 		log.Printf("waiting for informer caches to sync")
-		if !clients.WaitForCacheSync(stop) {
+		if !clients.WaitForCacheSync(serveContext) {
```
skriss, line 585: Switch to `g.AddContext(...)` for this task too so the workgroup can manage cancellation of the context. It looks like this task may require adding an `error` return parameter to the function that `g.AddContext(...)` accepts, too.

cmd/contour/serve.go
```diff
@@ -631,14 +635,15 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		select {
 		case sig := <-c:
 			log.WithField("context", "sigterm-handler").WithField("signal", sig).Info("shutting down")
+			cancel()
```
skriss, line 638: If you make the above changes, then you shouldn't need to explicitly cancel here since the workgroup will handle it for you when this function returns.

internal/k8s/clients.go
```diff
@@ -95,15 +95,15 @@ func (c *Clients) InformerForResource(gvr schema.GroupVersionResource) (Informer
...
 
-func (c *Clients) StartInformers(stopChan <-chan struct{}) error {
-	return c.cache.Start(stopChan)
+func (c *Clients) StartInformers(ctx context.Context) error {
+	return c.cache.Start(ctx)
 }
 
-func (c *Clients) WaitForCacheSync(stopChan <-chan struct{}) bool {
+func (c *Clients) WaitForCacheSync(ctx context.Context) bool {
 	// Note that in later controller-runtime releases, the API
```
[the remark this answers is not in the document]
stevesloka: Yup! I did in one branch, but forgot.

cmd/contour/serve.go
```diff
@@ -187,6 +187,10 @@ func validateCRDs(dynamicClient dynamic.Interface, log logrus.FieldLogger) {
 
 // doServe runs the contour serve subcommand.
 func doServe(log logrus.FieldLogger, ctx *serveContext) error {
+	ectx := context.Background()
```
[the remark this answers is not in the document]
stevesloka: 👍

skriss (changes requested):

cmd/contour/serve.go
stevesloka: yeah it's unfortunate that it has `context` in the name which does make it confusing.

cmd/contour/serve.go
stevesloka: Workgroup needs a way to get the sigterm and shutdown which is why I had this calling `cancel()` when it received that signal. Let me see if I can put that into the workgroup and remove the sigterm bits from serve.go all together.

skriss: You should be able to leave the sigterm handling as-is; when it receives a sigterm, that function returns, which then causes the workgroup to signal all the other tasks to stop. But that requires all the other tasks to get their cancellation from the workgroup too.

skriss:
see https://github.com/skriss/contour/commit/b4bd19304f04a062b7c2d269ce85bca9cdf3bf07 for what I was thinking (can't guarantee I don't have a bug)

stevesloka:
Closing in favor of #3375

Issue #3323, opened 2021-02-05
AAAA querys in IPv4 mode between envoy and contour


movikbence:
I am using contour as an IPV4 loadbalancer service and I noticed that in every 5 seconds a dns query is sent by envoy.
That is clear that the 5 sec query is happening due to the Strict DNS option in envoy bootstrap config.
"When using strict DNS service discovery, Envoy will continuously and asynchronously resolve the specified DNS targets."

I don't understand why are the querys are being sent on IPv6.

On slack there is a conversation started yesterday, here is the link for it:
https://kubernetes.slack.com/archives/C8XRH2R4J/p1612456585116300

**Environment:**

- Contour version: 1.11
- Kubernetes version: (use `kubectl version`): v1.15.2
- Kubernetes installer & version: v1.16.2
- Cloud provider or hardware configuration: bare metal
- OS (e.g. from `/etc/os-release`): suse

sunjayBhatia:
As mentioned in Slack, its likely we could prevent this by ensuring the Envoy -> Contour cluster service (and probably other extension gRPC services) lookups happen over IPv4 only if desired via a config parameter etc.

movikbence:
Hi

Has the investigation started on this issue?
Thanks in advance!

youngnick:
Hi @movikbence, I'm sorry but it hasn't yet. Thanks for the ping, I will make sure we get someone onto it.

movikbence:
Thank you for the update @youngnick !

youngnick:
So, the AAAA query is being sent because that's Envoy's default. There is a parameter for each cluster - `dns_lookup_family`, that may be set to `AUTO`, `V4_ONLY`, or `V6_only`, with `AUTO` being the default.

In the `AUTO` case, Envoy will look for IPv6 addresses first with AAAA queries, then try IPv4 with A queries.

In order to fix this, we would need to add a way to specify the DNS family for the `contour` cluster in the bootstrap config output by `contour bootstrap`. Which doesn't seem like too big a change, to be honest.

abhide:
@youngnick makes sense. Looked through `contour bootstrap` code. 
Posted a PR: https://github.com/projectcontour/contour/pull/3376 .

sunjayBhatia:
See https://github.com/projectcontour/contour/pull/3376#pullrequestreview-594262150 for some comments on what we should test to ensure this is fully finished

youngnick:
[...]

PR #3376, opened 2021-02-19
Allow user to set DNS lookup family as part of bootstrap config

abhide:
Updates: #3323

### DIFF
--- a/cmd/contour/bootstrap.go
+++ b/cmd/contour/bootstrap.go
@@ -35,5 +35,6 @@ func registerBootstrap(app *kingpin.Application) (*kingpin.CmdClause, *envoy.Boo
 	bootstrap.Flag("envoy-key-file", "Client key filename for Envoy secure xDS gRPC communication.").Envar("ENVOY_KEY_FILE").StringVar(&config.GrpcClientKey)
 	bootstrap.Flag("namespace", "The namespace the Envoy container will run in.").Envar("CONTOUR_NAMESPACE").Default("projectcontour").StringVar(&config.Namespace)
 	bootstrap.Flag("xds-resource-version", "The versions of the xDS resources to request from Contour.").Default("v3").StringVar((*string)(&config.XDSResourceVersion))
+	bootstrap.Flag("dns-lookup-family", "Defines what DNS Resolution Policy to use for Envoy -> Contour cluster name lookup. Either v4, v6 or auto.").StringVar(&config.DNSLookupFamily)
 	return bootstrap, &config
 }
--- a/internal/envoy/bootstrap.go
+++ b/internal/envoy/bootstrap.go
@@ -84,6 +84,10 @@ type BootstrapConfig struct {
 	// referenced in the configuration actually exist. This option is for
 	// testing only.
 	SkipFilePathCheck bool
+
+	// DNSLookupFamily specifies DNS Resolution Policy to use for Envoy -> Contour cluster name lookup.
+	// Either v4, v6 or auto.
+	DNSLookupFamily string
 }
 
 func (c *BootstrapConfig) GetXdsAddress() string { return stringOrDefault(c.XDSAddress, "127.0.0.1") }
@@ -95,7 +99,9 @@ func (c *BootstrapConfig) GetAdminPort() int { return intOrDefault(c.AdminPort, 
 func (c *BootstrapConfig) GetAdminAccessLogPath() string {
 	return stringOrDefault(c.AdminAccessLogPath, "/dev/null")
 }
-
+func (c *BootstrapConfig) GetDNSLookupFamily() string {
+	return stringOrDefault(c.DNSLookupFamily, "auto")
+}
 func stringOrDefault(s, def string) string {
 	if s == "" {
 		return def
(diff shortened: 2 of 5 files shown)


internal/envoy/bootstrap.go
```diff
@@ -84,6 +84,9 @@ type BootstrapConfig struct {
 	// referenced in the configuration actually exist. This option is for
 	// testing only.
 	SkipFilePathCheck bool
+
+	// DNSLookupFamily
```
abhide, line 88: oops missed this out. will fix in next patch

cmd/contour/bootstrap.go
```diff
@@ -35,5 +35,6 @@ func registerBootstrap(app *kingpin.Application) (*kingpin.CmdClause, *envoy.Boo
 	bootstrap.Flag("envoy-key-file", "Client key filename for Envoy secure xDS gRPC communication.").Envar("ENVOY_KEY_FILE").StringVar(&config.GrpcClientKey)
 	bootstrap.Flag("namespace", "The namespace the Envoy container will run in.").Envar("CONTOUR_NAMESPACE").Default("projectcontour").StringVar(&config.Namespace)
 	bootstrap.Flag("xds-resource-version", "The versions of the xDS resources to request from Contour.").Default("v3").StringVar((*string)(&config.XDSResourceVersion))
+	bootstrap.Flag("dns-lookup-family", "DNS Resolution Policy. Either v4, v6 or auto").StringVar(&config.DNSLookupFamily)
```
abhide, line 38: naming is hard! :)

sunjayBhatia, line 38: This addition also needs some docs here: https://projectcontour.io/docs/v1.12.0/configuration/#flags (source file https://github.com/projectcontour/contour/blob/d31d2b6fe22097e44ab55924e4707927c5084807/site/docs/main/configuration.md#flags)

sunjayBhatia (changes requested):
I think to fully fix https://github.com/projectcontour/contour/issues/3323 we need to allow configurability of the DNS lookup family on `ExtensionService` clusters as well. Right now I *believe* we don't set anything on those clusters so they would do ipv6/ipv4 lookups as well. We might want to make that configurable per `ExtensionService` or take this global default, not sure! Could be done in another PR though

As final validation for this we should deploy a Contour + Envoys and configure some `Ingresses`, and `HTTPProxies` with `ExtensionServices` and monitor the k8s DNS logs.

cmd/contour/bootstrap.go
abhide: Thanks @sunjayBhatia . Fixed this in next patchset.

abhide:
> I think to fully fix #3323 we need to allow configurability of the DNS lookup family on `ExtensionService` clusters as well. Right now I _believe_ we don't set anything on those clusters so they would do ipv6/ipv4 lookups as well. We might want to make that configurable per `ExtensionService` or take this global default, not sure! Could be done in another PR though
> 
> As final validation for this we should deploy a Contour + Envoys and configure some `Ingresses`, and `HTTPProxies` with `ExtensionServices` and monitor the k8s DNS logs.

@sunjayBhatia Thanks for giving further direction here. Currently, seems like we set `dns-lookup-family` for Kubernetes `externalName` service via Contour. Yes, we need to extend it for clusters. Will look into that next. Would be good if we can merge this PR if things look good and address the remaining items in a next PR. Will keep #3323 open till its correctly fixed.

cmd/contour/bootstrap.go
```diff
@@ -35,5 +35,6 @@ func registerBootstrap(app *kingpin.Application) (*kingpin.CmdClause, *envoy.Boo
 	bootstrap.Flag("envoy-key-file", "Client key filename for Envoy secure xDS gRPC communication.").Envar("ENVOY_KEY_FILE").StringVar(&config.GrpcClientKey)
 	bootstrap.Flag("namespace", "The namespace the Envoy container will run in.").Envar("CONTOUR_NAMESPACE").Default("projectcontour").StringVar(&config.Namespace)
 	bootstrap.Flag("xds-resource-version", "The versions of the xDS resources to request from Contour.").Default("v3").StringVar((*string)(&config.XDSResourceVersion))
+	bootstrap.Flag("dns-lookup-family", "DNS Resolution Policy for Envoy -> Contour cluster lookup. Either v4, v6 or auto").StringVar(&config.DNSLookupFamily)
```
youngnick, line 38:
```suggestion
	bootstrap.Flag("dns-lookup-family", "Defines what DNS Resolution Policy to use for Envoy -> Contour cluster name lookup. Either v4, v6 or auto").StringVar(&config.DNSLookupFamily)
```

youngnick (commented):
One small wording suggestion, but other than that, LGTM.

cmd/contour/bootstrap.go
abhide: Fixed

internal/envoy/bootstrap.go
sunjayBhatia, line 88: Should we add a descriptive comment here? At least to make the linter etc. happy (though it is a bit redundant and an internal struct field)

sunjayBhatia (approved):

skriss (approved):
LGTM

internal/envoy/bootstrap.go
youngnick: Yes, we absolutely should have a more descriptive comment here. Better to have too much documentation than not enough.

abhide: done

PR #3378, opened 2021-02-19
integration: Add simple integration test for GatewayAPIs

stevesloka:
Adds a simple integration test based upon the httpproxy/003-path-condition-match.yaml test that already
exists for HTTPProxy resources, but aligns to GatewayAPI.

Updates #2287


sunjayBhatia (commented):
Looks good to me, good idea to split out the different directory for gateway APIs, probably should do that for some Ingress v1 tests as well that maybe don't get pushed upstream to the conformance tests

sunjayBhatia (approved):

youngnick (approved):

PR #3380, opened 2021-02-19
internal/dag: Only process Gateways that match configuration

stevesloka:
Removes the map of Gateway objects to only store a single since that's all Contour can be configured to use. Then filters on storing the Gateway that matches its configuration. 

Created off of #3370 so that should merge first before this.


sunjayBhatia:
Moving to 1.14.0 as this is still a draft

stevesloka:
@sunjayBhatia I'd like to get this into 1.13. It was draft since I was stacking PRs.

internal/dag/cache.go
```diff
@@ -100,6 +102,27 @@ func (kc *KubernetesCache) matchesIngressClass(obj metav1.Object) bool {
 
 }
 
+// matchesGateway returns true if the given Kubernetes object
+// belongs to the Gateway that this cache is using.
+func (kc *KubernetesCache) matchesGateway(obj metav1.Object) bool {
+
+	fmt.Println("--- obj: ", k8s.NamespacedNameOf(obj))
+	fmt.Println("--- gateway: ", kc.Gateway)
```
skriss, line 110: can drop these now :)

internal/dag/cache.go
```diff
@@ -100,6 +102,27 @@ func (kc *KubernetesCache) matchesIngressClass(obj metav1.Object) bool {
 
 }
 
+// matchesGateway returns true if the given Kubernetes object
+// belongs to the Gateway that this cache is using.
+func (kc *KubernetesCache) matchesGateway(obj metav1.Object) bool {
```
skriss, line 107: I'd change this function to take a `*gatewayapi_v1alpha1.Gateway` as a parameter. Or, if the intent is to have this work for `HTTPRoutes`/etc. in the future, then I'd probably add a type switch within the implementation, since comparing NS+Name is only valid for `Gateways`.

internal/dag/cache.go
```diff
@@ -354,18 +374,15 @@ func (kc *KubernetesCache) remove(obj interface{}) bool {
...
 		return ok
 	case *gatewayapi_v1alpha1.Gateway:
-		m := k8s.NamespacedNameOf(obj)
-		_, ok := kc.gateways[m]
-		// TODO(youngnick): Remove this once gateway-api actually have behavior
-		// other than being removed from the cache.
-		kc.WithField("experimental", "gateway-api").WithField("name", m.Name).WithField("namespace", m.Namespace).Debug("Removing Gateway")
-		delete(kc.gateways, m)
-		return ok
+		if obj.Name == kc.gateway.Name && obj.Namespace == kc.gateway.Namespace {
```
skriss, line 377: any reason not to use the helper here?

skriss (changes requested):
couple comments.

Also, just confirming, but this isn't intended to limit the `HTTPRoutes` that get processed, right? That's coming in a subsequent PR?

skriss:
Moved back to 1.13, shouldn't be an issue.

internal/dag/cache.go
stevesloka: bah I'm so bad at leaving those in :headbang:

internal/dag/cache.go
stevesloka: Yup the intent was just to filter out for a specific Gateway. I'll update.

internal/dag/cache.go
stevesloka: Which helper are you referring to?

stevesloka: The helper I added, haha. Yeah I updated makes total sense. =)

skriss (approved):
LGTM

sunjayBhatia (approved):

Issue #2138, opened 2020-01-22
Ingress V1: hostname wildcard support


jpeach:
Support Ingress V1 hostname wildcards

https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20190125-ingress-api-group.md#hostname-wildcards

> Add support for a single wildcard * as the first label in the hostname.

This may have knock-on effects on HTTPS, as a wildcard hostname probably needs a wildcard TLS certificate. Alternatively, you could support routing several non-wildcard TLS certificates to a single wildcard hostname.

mspinassi:
+1

sunjayBhatia:
Updated KEP link: https://github.com/kubernetes/enhancements/tree/c14615cbf22c760fd0a6cbb4e962c66260c98463/keps/sig-network/1453-ingress-api#hostname-wildcards

sunjayBhatia:
Using ingress conformance tests as integration tests, however we have a general issue with the conformance tests:

In https://github.com/projectcontour/contour/issues/1697 and https://github.com/projectcontour/contour/pull/1714 we stopped accepting `Opaque` type secrets that had certain keys (basically we reject those that mirror TLS secrets). The ingress conformance tests currently generate an `Opaque` secret so we don't set up any TLS listeners for `Ingresses` in the tests/the tests fail etc (see https://github.com/kubernetes-sigs/ingress-controller-conformance/blob/79bd068cbb31d77c2a060f61eea795f9791a3d48/test/kubernetes/kubernetes.go#L226-L231).

I made a PR to change the secrets generated by the tests to a TLS certificate (https://github.com/kubernetes-sigs/ingress-controller-conformance/pull/88) but should we reconsider allowing `Opaque` certs? Has there been any clamor for this?

sunjayBhatia:
Also it seems this test: https://github.com/kubernetes-sigs/ingress-controller-conformance/blob/79bd068cbb31d77c2a060f61eea795f9791a3d48/features/host_rules.feature#L78-L82 is failing when wildcards are enabled (all other wildcard tests pass)

```
                    {
                        "keyword": "When ",
                        "name": "I send a \"GET\" request to \"http://baz.bar.foo.com\"",
                        "line": 81,
                        "match": {
                            "location": "feature.go:112"
                        },
                        "result": {
                            "status": "passed",
                            "duration": 9144200
                        }
                    },
                    {
                        "keyword": "Then ",
                        "name": "the response status-code must be 404",
[...]

Issue #2139, opened 2020-01-22
Support Ingress V1 


jpeach:
Kubernetes SIG Network is proposing a number of changes to the Ingress spec in order to graduate it to V1. Since Ingress is a standard common feature, we should evaluate whether Contour should support the V1 changes.

- [x] Ingress V1: path matching mode #2135
- [x] Ingress V1: support spec.backend field rename #2136
- [x] Ingress V1: support for service backends #2137
- [x] Ingress V1: hostname wildcard support #2138
- [x] Ingress V1: support IngressClass #2146
- [x] Detailed documentation on Contour Ingress v1 support: https://github.com/projectcontour/contour/issues/3469

https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20190125-ingress-api-group.md

youngnick:
This issue is no longer blocked, and needs a design document. As it's related to the service-apis design, I'll get started on a design document for it as well.

youngnick:
Sorry, I was unclear here. This work will need its own design document, separate to the service-apis one, but it will be similar, in that an IngressClass describes a Contour deployment and its associated Envoys - Contour will ignore the `parameters` field of the IngressClass, and will only use the oldest IngressClass that has a matching `controller` field. See the WIP PR #3213 in the GatewayClass section for inspiration.

sunjayBhatia:
Not sure if the last step needs to be done as part of this or if we can call it pretty much done, all the ingress conformance tests are passing but flaky in CI/locally

skriss:
Personally I'd call this one done and just leave the conformance issue open separately.

sunjayBhatia:
Removed the ingress conformance tests item, closing this

PR #3381, opened 2021-02-19
Add support for wildcard hosts on Ingresses

sunjayBhatia:
- Wildcard virtualhosts now allowed
- Envoy's virtualhost matching does not match on a single DNS label like
we want to implement
- For wildcard host routes, we add a header match rule
to match on the ":authority" header that is more strict and will only
match one DNS label wildcard
- Adds an integration test to validate behavior with a wildcard
certificate
- Passes ingress conformance tests for wildcards
- Adds the "regex" header match strategy (only internal for now)

Fixes https://github.com/projectcontour/contour/issues/2138
Updates https://github.com/projectcontour/contour/issues/2139


sunjayBhatia:
How to run Ingress conformance tests:
- Check out this branch of the tests: https://github.com/sunjayBhatia/ingress-controller-conformance/tree/focus-host-rules-tests
  - Focuses the host rules tests (some non-wildcard ones are a bit flaky so beware) and uses a k8s TLS secret instead of an Opaque one
- Run `make build-image`
- Load the built image into your kind cluster
- Add `--image-pull-policy=IfNotPresent` to the `sonobuoy` args in `_integration/testsuite/run-ingress-conformance.sh` and set `INGRESS_CONFORMANCE_IMAGE` to the name of the image built above
- run `make check-ingress-conformance`
- Results are printed to the screen and a tarball in your working directory with the results should show the wildcard tests pass

sunjayBhatia:
Current caveats/TODOs:
- Docs
- Mismatched SNI and `:authority` header is not detected (integration test failure should show this)
  - We don't have the ability to get the requested server name in Lua at the moment so the Lua script for checking the above is disabled for wildcard hosts
  - If we didn't disable the Lua script, no HTTPS requests would work going to a wildcard host (checking for equality between `bar.foo.com` and `*.foo.com` cannot work!)
- Currently to use a wildcard route with TLS you need a cert that is intended to exactly match the hostname, e.g. as in the integration/feature tests:

```
apiVersion: networking.k8s.io/v1
kind: Ingress
...
spec:
  - hosts:
      - "*.foo.com"
    secretName: wildcard
  rules:
  - host: *.foo.com
     ...
```

- Still have to make a decision on if allowing a secret for a specific hostname to be applied on a wildcard Ingress rule is allowed, for example for the below, should `bar` be applied to `*.foo.com`, should it be an error, ignored?

```
apiVersion: networking.k8s.io/v1
kind: Ingress
...
spec:
  - hosts:
      - bar.foo.com
    secretName: bar
  rules:
  - host: "*.foo.com"
     ...
```

- Something that could be a separate PR: certs for wildcard hosts allowed to be applied to Ingress rules that match the wildcard, e.g in the below, `wildcard` would be applied to the `bar.foo.com` rule

```
apiVersion: networking.k8s.io/v1
kind: Ingress
...
spec:
  - hosts:
      - "*.foo.com"
    secretName: wildcard
  rules:
  - host: bar.foo.com
     ...
```

sunjayBhatia:
Going to mark this as ready just so it notifies people to review to get some feedback

internal/dag/ingress_processor.go
```diff
@@ -209,6 +215,24 @@ func route(ingress *networking_v1.Ingress, path string, service *Service, client
 	}
 
 	r.PathMatchCondition = &PrefixMatchCondition{Prefix: path}
+
+	// If we have a wildcard match, add a header match regex rule to match the
+	// hostname so we can be sure to only match one DNS label. This is required
+	// as Envoy's virtualhost hostname wildcard matching can match multiple
+	// labels. This match ignores a port in the hostname in case it is present.
+	if strings.HasPrefix(host, "*.") {
```
stevesloka, line 223: Should we validate the only a single `*` was added? Meaning if someone entered: `*.foo.*.com` it should get thrown out?

sunjayBhatia: `Ingress` objects seem to already have validation set up so I skipped doing that in Contour:

The following gets an error when applied:

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: test
spec:
  rules:
    - host: "*.projectcontour.*.io"
      http:
        paths:
        - pathType: Prefix
          path: "/"
          backend:
            service:
              name: echo
              port:
                number: 80
```

Error:

```
% kubectl apply -f /tmp/ingress.yaml
The Ingress "test" is invalid: spec.rules[0].host: Invalid value: "*.projectcontour.*.io": a wildcard DNS-1123 subdomain must start with '*.', followed by a valid DNS subdomain, which must consist of lower case alphanumeric characters, '-' or '.' and end with an alphanumeric character (e.g. '*.example.com', regex used for validation is '\*\.[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

internal/dag/builder_test.go
```diff
@@ -1752,7 +1745,9 @@ func TestDAGInsert(t *testing.T) {
 					},
 				},
 			}, {
-				Host: "*",
+				// Allow wildcard as first label.
+				// K8s will only allow hostnames with wildcards of this form.
+				Host: "*.example.com",
```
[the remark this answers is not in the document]
sunjayBhatia: Do you mean `*` as a valid hostname in `v1beta1` resources? Not quite sure which line you mean

`*` as a hostname is rejected:

```
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: test
spec:
  rules:
    - host: "*"
      http:
        paths:
        - pathType: Prefix
          path: "/"
          backend:
            serviceName: echo
            servicePort: 80
```

Error:

```
% kubectl apply -f /tmp/ingress.yaml
Warning: networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
The Ingress "test" is invalid: spec.rules[0].host: Invalid value: "*": a wildcard DNS-1123 subdomain must start with '*.', followed by a valid DNS subdomain, which must consist of lower case alphanumeric characters, '-' or '.' and end with an alphanumeric character (e.g. '*.example.com', regex used for validation is '\*\.[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

sunjayBhatia:
See https://github.com/envoyproxy/envoy/issues/15142

internal/dag/builder_test.go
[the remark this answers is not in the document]
stevesloka: I was just looking at the diff from the old test and it did have a `*` for a hostname, but was deleted so was just making sure that test is invalid or if it should still be valid for the v1beta1 tests.

internal/dag/ingress_processor.go
stevesloka: 👍

internal/dag/builder_test.go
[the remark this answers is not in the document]
sunjayBhatia: Yeah the same test was deleted for `v1` as well since you can't get a lone `*` that far

sunjayBhatia:
Still TODO: Add testing and fix for combining a catchall virtualhost (default backend or no supplied hostname in a rule) with a wildcard route b/c as of right now they are both using `*` for virtualhost domain which is invalid

youngnick:
> Still TODO: Add testing and fix for combining a catchall virtualhost (default backend or no supplied hostname in a rule) with a wildcard route b/c as of right now they are both using `*` for virtualhost domain which is invalid
I don't get this one. Are you saying the current catchall vhost is implemented using a `*` domain when we send the config to Envoy? Isn't a wildcard route more specific than that, because only the first label can be a wildcard, and there has to be more than one label in a wildcard hostname?

> Right now this change attempts to do the following:
> * for HTTP and HTTPS wildcard virtualhosts, match on `*` and only match routes if the `:authority` header matches something >   * this is because we allow you to supply `bar.foo.com:<port>` as a hostname that should match `bar.foo.com` so to match this for wildcards we have to jump through some hoops
Is that part of the Ingress spec? Really? Doesn't feel like we should be doing that.

> ...
>   * Could become a problem in Gateway APIs as we can have `*` as a hostname match and allow TLS
I didn't think that `*` by itself was a valid hostname, I thought you needed at least two labels to be able to do a wildcard.

sunjayBhatia:
> Is that part of the Ingress spec? Really? Doesn't feel like we should be doing that.

We have this: https://github.com/projectcontour/contour/blob/711001c50b3ff1a7e80ac61d4a76ccbe073dc236/internal/envoy/v3/route.go#L263-L266

Which was introduced b/c of https://github.com/projectcontour/contour/commit/8b16fd83e9747016910eed753ec42171ca799fa3

This is a contour-specific thing but presumably we would want wildcard routes to ignore the port if present just as on a precise hostname.

If we try to do a route match for `*.foo.com` we would also generate `*.foo.com:*` which Envoy accepts but it doesn't actually work to do pre AND post wildcard matching (the docs say either-or and empirically this is true as well, this test case fails: https://github.com/projectcontour/contour/pull/3381/files#diff-0261dd69a192ff188fa76df82a9df9b5c003c1c92fff658d47e5151f2894e857R126).

To implement ignoring a supplied port in the host, I've made it so wildcard routes match `*` and then we use the header match rules to match the backend (with a regex) for the wildcard as specified by the user. Not the greatest as we then end up making Envoy do a lot more work.

Another solution would be to strip the port from the `:authority` header in the early Lua filter so the router doesn't even need to think about matching the port with a post-wildcard and the matched domains could just be a single entry, making precise and wildcard hostname matches simpler. Downsides of this are more Lua, maybe the port could be relevant to some application so stripping it is no good?

sunjayBhatia:
> I didn't think that * by itself was a valid hostname, I thought you needed at least two labels to be able to do a wildcard.

GatewayAPI Listener.hostname says it can be empty or `*`: https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.Listener

sunjayBhatia:
This requires https://github.com/projectcontour/contour/pull/3458 as it makes life a lot simpler!

sunjayBhatia:
This still relies on https://github.com/envoyproxy/envoy/pull/15215 in an Envoy release, waiting for Envoy 1.18

sunjayBhatia:
Now that we have https://github.com/projectcontour/contour/pull/3458 it removes all the logic we would have needed around polluting the `*` wildcard host used for default backends or rules without a hostname so those comments can be disregarded

sunjayBhatia:
Still validating ingress conformance tests again since they are flaky

sunjayBhatia:
> Still validating ingress conformance tests again since they are flaky
> 
> Getting a 503 on https://github.com/kubernetes-sigs/ingress-controller-conformance/blob/79bd068cbb31d77c2a060f61eea795f9791a3d48/features/host_rules.feature#L70-L76 when running the tests

Creating the same Ingress and accompanying services etc. as the test manually everything works, no 503s

sunjayBhatia:
> > Still validating ingress conformance tests again since they are flaky
> > Getting a 503 on https://github.com/kubernetes-sigs/ingress-controller-conformance/blob/79bd068cbb31d77c2a060f61eea795f9791a3d48/features/host_rules.feature#L70-L76 when running the tests
> 
> Creating the same Ingress and accompanying services etc. as the test manually everything works, no 503s, gonna have to tail Envoy logs to see the reason for the 503 if it can be reproduced

Not able to repro again fortunately, we should note that running conformance tests in kind are flaky right now, might need some work when we enable them for real

sunjayBhatia:
**Note:** With this change we will need to update Envoy *first* before Contour b/c we are using a net new Envoy Lua API

skriss (commented):
this looks pretty good to me, pending a new Envoy release.

sunjayBhatia:
TODO: Add sorting of regex header match type, realized while working on path matches that this needs to be done as well

internal/featuretests/kubernetes.go
```diff
@@ -17,12 +17,24 @@ package featuretests
...
 import (
 	v1 "k8s.io/api/core/v1"
+	networking_v1 "k8s.io/api/networking/v1"
 	"k8s.io/api/networking/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-func Backend(svc *v1.Service) *v1beta1.IngressBackend {
+func IngressBackend(svc *v1.Service) *networking_v1.IngressBackend {
```
youngnick, line 26: Can we have a TODO here to move everything over to v1 from v1beta1, since v1beta1 will be removed soon?

youngnick (approved):
LGTM with one small followup.

sunjayBhatia:
Going to try to fix siteproof in a separate PR

sunjayBhatia:
See https://github.com/projectcontour/contour/pull/3596 for siteproof fixes

sunjayBhatia:
This should be ready for review again

Note if we merge this for 1.15 we will need to mention in upgrade instructions that the Envoy deployment has to be updated first before Contour

skriss (approved):
LGTM

PR #3382, opened 2021-02-22
ci: Add envoyproxy.io to ignore list for site-proof check

stevesloka:
The site proof check which runs on each run checks for invalid links, however, the links
to envoyproxy.io seem to fail the site check. This adds the envoyproxy.io site to the ignore
list which will skip this site from getting checked.

### DIFF
--- a/hack/site-proofing/cibuild
+++ b/hack/site-proofing/cibuild
@@ -18,4 +18,4 @@ htmlproofer ./_site \
     --allow-hash-href \
     --http-status-ignore 429 \
     --typhoeus-config '{"headers": {"User-Agent": "Mozilla/5.0 (Linux x86_64; rv:84.0) Gecko/20100101 Firefox/84.0"}}' \
-    --url-ignore "/www.haproxy.org/,/tools.ietf.org/" # gives false positive timeouts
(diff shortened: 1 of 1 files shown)


abhide (approved):

youngnick (approved):

sunjayBhatia (approved):
This is a tricky one b/c Envoy doc links would be bad to not catch if they're legitimately broken

LGTM though maybe we should file an issue to try to fiddle with the siteproof params so we don't skip as many sites or at least can craft config that lets us still check Envoy

skriss:
> This is a tricky one b/c Envoy doc links would be bad to not catch if they're legitimately broken
> 
> LGTM though maybe we should file an issue to try to fiddle with the siteproof params so we don't skip as many sites or at least can craft config that lets us still check Envoy

Agreed. Also, we should keep in mind we may be moving to Hugo in the near future, so perhaps our tooling will change.

sunjayBhatia:
> > This is a tricky one b/c Envoy doc links would be bad to not catch if they're legitimately broken
> > LGTM though maybe we should file an issue to try to fiddle with the siteproof params so we don't skip as many sites or at least can craft config that lets us still check Envoy
> 
> Agreed. Also, we should keep in mind we may be moving to Hugo in the near future, so perhaps our tooling will change.

https://github.com/projectcontour/contour/issues/3384

skriss:
Merging as-is, no risk of conflict.

PR #3386, opened 2021-02-22
Bump golang to 1.16.0

sunjayBhatia:
Release notes: https://golang.org/doc/go1.16


skriss:
We have a reference to the major/minor go version in `CONTRIBUTING.md` too that can be updated (or maybe removed so we don't have to keep bumping it?)

sunjayBhatia:
> We have a reference to the major/minor go version in `CONTRIBUTING.md` too that can be updated (or maybe removed so we don't have to keep bumping it?)

Good catch! Also seems like we could reword some of that doc and maybe remove some of the `GOPATH` stuff but that can be a separate issue

stevesloka (approved):

abhide (approved):

skriss (approved):

PR #3389, opened 2021-02-23
Makefile: Check if integration-tester is installed before executing tests

abhide:
integration-tests require integration-tester tool to be installed.
If its not installed, the execution fails after doing all the test setup.
Instead check if required tool is installed right at start.


stevesloka (approved):
/lgtm

sunjayBhatia (approved):

Issue #3264, opened 2021-01-19
envoy container in CrashLoopBackOff : error initializing configuration


vinzo99:
Hi

we are deploying Contour (v1.11.0), with Envoy (v1.16.2) as a DaemonSet, using the following yaml templates :
https://github.com/projectcontour/contour/blob/release-1.11/examples/render/contour.yaml

We only applied minor changes to fit our configuration (such as pointing to our local images repository, adding privileges for RBAC etc).

When firing up the helm installation, the Envoy pod fails with CrashLoopBackOff, with the following error in the envoy container :

```
[2021-01-19 14:41:26.115][1][info][main] [source/server/server.cc:305] initializing epoch 0 (base id=0, hot restart version=11.104)
[2021-01-19 14:41:26.115][1][info][main] [source/server/server.cc:307] statically linked extensions:
[2021-01-19 14:41:26.115][1][info][main] [source/server/server.cc:309]   envoy.thrift_proxy.filters: envoy.filters.thrift.rate_limit, envoy.filters.thrift.router
[2021-01-19 14:41:26.115][1][info][main] [source/server/server.cc:309]   envoy.tracers: envoy.dynamic.ot, envoy.lightstep, envoy.tracers.datadog, envoy.tracers.dynamic_ot, envoy.tracers.lightstep, envoy.tracers.opencensus, envoy.tracers.xray, envoy.tracers.zipkin, envoy.zipkin
[2021-01-19 14:41:26.115][1][info][main] [source/server/server.cc:309]   envoy.retry_priorities: envoy.retry_priorities.previous_priorities
[2021-01-19 14:41:26.115][1][info][main] [source/server/server.cc:309]   envoy.internal_redirect_predicates: envoy.internal_redirect_predicates.allow_listed_routes, envoy.internal_redirect_predicates.previous_routes, envoy.internal_redirect_predicates.safe_cross_scheme
[2021-01-19 14:41:26.115][1][info][main] [source/server/server.cc:309]   envoy.thrift_proxy.protocols: auto, binary, binary/non-strict, compact, twitter
[2021-01-19 14:41:26.115][1][info][main] [source/server/server.cc:309]   envoy.resource_monitors: envoy.resource_monitors.fixed_heap, envoy.resource_monitors.injected_resource
[...]

PR #3390, opened 2021-02-23
Change config/resources directory permissions 075 -> 077

abhide:
Fixes: #3264

### DIFF
--- a/internal/envoy/v3/bootstrap.go
+++ b/internal/envoy/v3/bootstrap.go
@@ -46,7 +46,11 @@ func WriteBootstrap(c *envoy.BootstrapConfig) error {
 	}
 
 	if c.ResourcesDir != "" {
-		if err := os.MkdirAll(path.Join(c.ResourcesDir, "sds"), 0750); err != nil {
+		// Setting permissions to 0777 explicitly.
+		// Refer Issue: https://github.com/projectcontour/contour/issues/3264
+		// The secrets in this directory are "pointers" to actual secrets
+		// mounted from Kubernetes secrets; that means the actual secrets aren't 0777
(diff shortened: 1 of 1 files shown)


internal/envoy/v3/bootstrap.go
```diff
@@ -46,7 +46,7 @@ func WriteBootstrap(c *envoy.BootstrapConfig) error {
 	}
 
 	if c.ResourcesDir != "" {
-		if err := os.MkdirAll(path.Join(c.ResourcesDir, "sds"), 0750); err != nil {
+		if err := os.MkdirAll(path.Join(c.ResourcesDir, "sds"), 0777); err != nil {
```
youngnick, line 49: I'd like a comment here linking back to the issue and explaining why we are doing this, having any directory containing secrets as `0777` is very unusual, we need to explain that in this case it's okay because these are actually pointers to secrets that are mounted in via Secret volume mounts (which have their own separate set of controls).

That way, when this is queried in a future security audit, we'll all be able to remember why we did it. :)

youngnick (changes requested):
The change itself looks great, I think we just need a bit more documentation around why we're doing it.

youngnick:
@abhide, are you still working on this one?

abhide:
> @abhide, are you still working on this one?

@youngnick yes, I will post an update to fix your review comment later today.

internal/envoy/v3/bootstrap.go
abhide: Pushed fix. PTAL.

sunjayBhatia (approved):

youngnick (approved):

PR #3394, opened 2021-02-23
Adds Gateway API User Guide

danehans:
Adds a guide for using Gateway API with contour/contour-operator. Note that `netlify.toml` was updated for v1.13.0 versioned redirects .

xref: https://github.com/projectcontour/contour-operator/issues/106

### DIFF
--- a/site/_guides/gateway-api.md
+++ b/site/_guides/gateway-api.md
@@ -1,0 +1,126 @@
+---
+title: Using Gateway API with Contour
+layout: page
+---
+
+## Introduction
+
+[Gateway API][1] is an open source project managed by the Kubernetes SIG-NETWORK community. The project's goal is to
+evolve service networking APIs within the Kubernetes ecosystem. Gateway API consists of multiple resources that provide
+user interfaces to expose Kubernetes applications- Services, Ingress, and more.
+
+### Background
+
+Gateway API targets three personas:
+
+- __Platform Provider__: The Platform Provider is responsible for the overall environment that the cluster runs in, i.e.
+  the cloud provider. The Platform Provider will interact with GatewayClass and Contour resources.
+- __Platform Operator__: The Platform Operator is responsible for overall cluster administration. They manage policies,
+  network access, application permissions and will interact with the Gateway resource.
+- __Service Operator__: The Service Operator is responsible for defining application configuration and service
+  composition. They will interact with xRoute resources and other typical Kubernetes resources.
+
+Gateway API contains three primary resources:
+
+- __GatewayClass__: Defines a set of gateways with a common configuration and behavior.
+- __Gateway__: Requests a point where traffic can be translated to a Service within the cluster.
+- __xRoutes__: Describes how traffic coming via the Gateway maps to the Services.
+
+Resources are meant to align with personas. For example, a platform operator will create a Gateway, so a developer can
+expose an HTTP application using an HTTPRoute resource.
+
+### Prerequisites
+The following prerequisites must be met before using Gateway API with Contour:
+
+- A working [Kubernetes][2] cluster. Refer to the [compatibility matrix][3] for cluster version requirements.
+- The [kubectl][4] command-line tool, installed and configured to access your cluster.
+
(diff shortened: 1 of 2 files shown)


site/_guides/gateway-api.md
```diff
@@ -0,0 +1,118 @@
...
+Create the Gateway API resources:
+```shell
+$ kubectl apply -f https://projectcontour.io/quickstart/gateway.yaml
+```
+This command creates:
+
+- Namespace `projectcontour` to run the Gateway and child resources, i.e. Envoy DaemonSet.
+- A Contour custom resource named `contour-gateway-sample` in the operator's namespace. This resource exposes
+  infrastructure-specific configuration and is referenced by the GatewayClass.
+- A GatewayClass named `sample-gatewayclass` that abstracts the infrastructure-specific configuration from Gateways.
```
stevesloka, line 65: Maybe we should tie the name of the class to Contour in some way? No big deal either way just might help identify it better.

site/_guides/gateway-api.md
```diff
@@ -0,0 +1,118 @@
...
+Run the test application:
+```shell
+$ kubectl apply -f https://projectcontour.io/quickstart/kuard.yaml
+```
+This command creates:
+
+- A Deployment named `kuard` in namespace `projectcontour` to run kuard as the test application.
+- A Service named `kuard` in namespace `projectcontour` to expose the kuard application on TCP port 80.
+- An HTTPRoute named `kuard` in namespace `projectcontour` to route requests for "*.projectcontour.io" to the kuard
+  service.
```
skriss, line 80: Also not a big deal, but consider installing the workload in a namespace other than `projectcontour` since it usually wouldn't go in there.

skriss (approved):
LGTM as-is, one minor comment for consideration.

site/_guides/gateway-api.md
danehans: @skriss "SameNamespace" is the [default](https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha1/shared_types.go#L65) for a route. I wanted to stick with defaults as much as possible to keep things simple.

youngnick (approved):
I think this is great for a first start. We definitely should look at tidying up some of the comments later, but I think is fine for a first cut.

site/_guides/gateway-api.md
```diff
@@ -0,0 +1,118 @@
...
+NAME                         READY   STATUS    RESTARTS   AGE
+pod/kuard-798585497b-78x6x   1/1     Running   0          21s
+pod/kuard-798585497b-7gktg   1/1     Running   0          21s
+pod/kuard-798585497b-zw42m   1/1     Running   0          21s
+
+NAME            TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)   AGE
+service/kuard   ClusterIP   172.30.168.168   <none>        80/TCP    21s
+
+NAME                                  HOSTNAMES
+httproute.networking.x-k8s.io/kuard   ["*.projectcontour.io"]
```
sunjayBhatia, line 94: Can we actually maybe modify the example to not use a wildcard and rather a specific hostname? Since the actual implications of a wildcard hostname are somewhat nuanced and advertising that we support this already is a bit misleading

sunjayBhatia (changes requested):

site/_guides/gateway-api.md
youngnick: That is a good catch, I agree.

danehans: @sunjayBhatia https://github.com/projectcontour/contour-operator/pull/220 is required for me to update the hostname.

sunjayBhatia: @danehans merged!

danehans:
Commit `15883b0` resolves https://github.com/projectcontour/contour/pull/3394#discussion_r581467844 and adds an option for deploying using a nodeport service instead of a loadbalancer service.

sunjayBhatia (approved):

Issue #3401, opened 2021-02-24
Implement the RouteBindingSelector


stevesloka:
A RouteBindingSelector defines a schema for associating routes with the Gateway. If Namespaces and Selector are defined, only routes matching both selectors are associated with the Gateway.

skriss:
@stevesloka was this closed by #3415?

stevesloka:
Not fully, there's still a namespaces bit that needs added.

PR #3402, opened 2021-02-24
internal/dag: Watch namespaces from k8s api

stevesloka:
The RouteBindingSelector in Gateway API has an option to only process routes which exist
in a namespace based upon a selector defined in the Gateway.

When "Selector" is chosen then only Routes in namespaces selected by the selector may be used by this Gateway (https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.RouteSelectType).

_Added release note action label since this updates RBAC permissions, but only needed if Gateway APIs are present._ 

Updates #3401

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -477,9 +477,8 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		if !clients.ResourcesExist(r) {
 			log.WithField("resource", r).Warn("resource type not present on API server")
 			continue
-		} else {
-			foundGatewayAPI = true
 		}
+		foundGatewayAPI = true
 
 		if err := informOnResource(clients, r, &dynamicHandler); err != nil {
 			log.WithError(err).WithField("resource", r).Fatal("failed to create informer")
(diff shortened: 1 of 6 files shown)


stevesloka:
I think informer caches would make this much simpler, but thought we could add that later. This would let us use internal label queries and not loop over the internal caches, but think it's a good spot for a tech debt update if someone wanted to pick it up. 

// https://github.com/projectcontour/contour/issues/2683

cmd/contour/serve.go
```diff
@@ -471,17 +471,28 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	if ctx.UseExperimentalServiceAPITypes {
 		log.Warn("DEPRECATED: The flag '--experimental-service-apis' is deprecated and should not be used. Please configure the gateway.name & gateway.namespace in the configuration file to specify which Gateway Contour will be watching.")
 	}
+
+	foundGatewayAPI := false
 	for _, r := range k8s.GatewayAPIResources() {
 		if !clients.ResourcesExist(r) {
 			log.WithField("resource", r).Warn("resource type not present on API server")
 			continue
+		} else {
```
skriss, line 480: nit: can drop the `else` since the `if` block ends with a `continue`.

skriss (changes requested):
one tiny nit, otherwise looks good.

youngnick (approved):

cmd/contour/serve.go
stevesloka: removed! =)

skriss (approved):

PR #3407, opened 2021-02-24
Generated v1.13.0 Release Docs

danehans:
Adds generated docs for v1.13.0 release.


skriss (approved):
LGTM. Looks like a bunch of whitespace changes on the YAML but I don't think that should cause any issue.

youngnick (approved):
LGTM, the site-proof failure is another instance of that not doing what we wanted.

PR #3412, opened 2021-02-25
cmd/contour: pass pointers to StatusAddressUpdater

### LINKED ISSUE #3411: seeing "unsupported type received" context=StatusAddressUpdater in logs
While debugging another issue, I noticed the following in the logs:

`time="2021-02-25T18:15:45Z" level=debug msg="unsupported type received" context=StatusAddressUpdater`

This is logged from https://github.com/projectcontour/contour/blob/main/internal/k8s/statusaddress.go#L91.

This code expects pointers to Ingress or HTTPProxy (see https://github.com/projectcontour/contour/blob/main/internal/k8s/statusaddress.go#L79-L88).

However, in [this code](https://github.com/projectcontour/contour/blob/main/cmd/contour/ingressstatus.go#L116-L130), we're passing non-pointers to the `OnAdd` method.

I'd expect the impact of this is that if the external IP of the envoy service changes, it may not be immediately reflected in HTTPProxies/Ingresses, though they should eventually get updated the next time the Ingresses/HTTPProxies change.

^^ I was able to confirm this; when the `envoy` service's load balancer IP changes, it's not reflected immediately in the HTTPProxy.

skriss:
Fixes an issue where non-pointers were being passed to
the StatusAddressUpdater when the load balancer address
changed, which was resulting in HTTPProxies/Ingresses
not immediately getting updated with the new address.

Fixes #3411.

### DIFF
--- a/cmd/contour/ingressstatus.go
+++ b/cmd/contour/ingressstatus.go
@@ -116,16 +116,16 @@ func (isw *loadBalancerStatusWriter) Start(stop <-chan struct{}) error {
 			if err := isw.clients.Cache().List(context.Background(), &ingressList); err != nil {
 				isw.log.WithError(err).WithField("kind", "Ingress").Error("failed to list objects")
 			} else {
-				for _, i := range ingressList.Items {
-					u.OnAdd(i)
+				for i := range ingressList.Items {
+					u.OnAdd(&ingressList.Items[i])
 				}
 			}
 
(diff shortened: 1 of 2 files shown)


skriss:
I've confirmed manually that with v1.13.0, changes to the LB IP are not immediately reflected in an `HTTPProxy`, whereas with this patch, they are.

stevesloka (approved):

abhide (approved):

sunjayBhatia (approved):

PR #3415, opened 2021-02-26
internal/dag: Filter gateway-api routes against the Gateway RouteBindingSelector

stevesloka:
Filters the HTTPRoutes to make sure they match the RouteBindingSelector for a Gateway.

Updates #3401


_integration/testsuite/gatewayapi/001-path-condition-match.yaml
```diff
@@ -76,6 +76,7 @@ apiVersion: networking.x-k8s.io/v1alpha1
 kind: GatewayClass
 metadata:
   name: contour-class
+  namespace: projectcontour
```
youngnick, line 79: I thought that GatewayClass resources are still cluster-scoped? If that's the case then this will be discarded, I think?

internal/dag/gatewayapi_processor.go
```diff
@@ -41,8 +42,42 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
 		p.source = nil
 	}()
 
+	// Gateway must defined for resources to be processed.
```
youngnick, line 45:
```suggestion
	// Gateway must be defined for resources to be processed.
```

internal/selector/selector.go
```diff
@@ -0,0 +1,89 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package selector
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// MatchesSelector returns true if the selector set matches the values in the labels.
+func MatchesLabelSelector(selector metav1.LabelSelector, matchLabels map[string]string) bool {
```
youngnick, line 19: This feels like a little bit of a stutter, although it's technically correct. What do you think about `selector.MatchesLabels` instead?

internal/dag/builder_test.go
```diff
@@ -83,6 +83,11 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 						MatchLabels: map[string]string{
 							"app": "contour",
 						},
+						MatchExpressions: []metav1.LabelSelectorRequirement{{
```
youngnick, line 86: Does this mean we're not testing the "no selectors" case explicitly?

youngnick (changes requested):
A few small things, this is looking really good though.

_integration/testsuite/gatewayapi/001-path-condition-match.yaml
stevesloka: Yup it is and is ignored, but it's moving back to namespace scoped. I'll take out to match the v0.2.0 release we're tagged against.

internal/selector/selector.go
stevesloka: I had it like that first time in coding, but the label selector can be a set of labels or expressions so that's why I moved it back to have "selector" in the name.

internal/dag/builder_test.go
stevesloka: Yup we're not, I'll add a test for that. =)

internal/selector/selector.go
youngnick: What about calling the package `match` and this function `LabelSelector`? Or possibly `compare` or `comparer`? Then we have a place to put other comparison operators, and we're not stuttering.

youngnick (changes requested):
The `selector.SomethingSelector` name pattern really feels wrong to me. I've offered some alternatives, Interested to know what others think.

stevesloka:
Yeah I think changing the package works, I'll do that.

stevesloka:
@youngnick I renamed that to `match`.

internal/dag/gatewayapi_processor.go
```diff
@@ -41,8 +42,42 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+		// If Namespaces and Selector are defined, only routes matching both selectors are associated with the Gateway.
+
+		// ## RouteBindingSelector ##
+		//
+		// Selector specifies a set of route labels used for selecting routes to associate
+		// with the Gateway. If this Selector is defined, only routes matching the Selector
+		// are associated with the Gateway. An empty Selector matches all routes.
+		for _, listener := range p.source.gateway.Spec.Listeners {
+
+			if len(listener.Routes.Selector.MatchLabels) > 0 || len(listener.Routes.Selector.MatchExpressions) > 0 {
```
sunjayBhatia, line 66: Could this be moved into `match.LabelSelector`? Or should the empty case not generically match all resources?

sunjayBhatia, line 66: Could this be moved into `match.LabelSelector`? Or should the empty case not generically match all resources?

sunjayBhatia: Seems like it already does? https://github.com/projectcontour/contour/pull/3415/files#diff-cc1654d51a4731d486b648439e46b022045d9bc596e7e5f0da32cf33e2ded3eaR26

```
if len(selector.MatchExpressions) == 0 {
	matchesExpression = true
} ...
```

 and the same for matching labels

internal/match/selector.go
```diff
@@ -0,0 +1,89 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package match
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// LabelSelector returns true if the selector set matches the values in the labels.
+func LabelSelector(selector metav1.LabelSelector, matchLabels map[string]string) bool {
```
skriss, line 19: I think you should be able to use https://pkg.go.dev/k8s.io/apimachinery@v0.20.4/pkg/apis/meta/v1#LabelSelectorAsSelector and then https://pkg.go.dev/k8s.io/apimachinery/pkg/labels here to get the selector logic for free.

stevesloka: oh nice! I was thinking it was already implemented but couldn't find it. Let me see how that would work.

skriss: we used it in Velero, let me look for a link

skriss: https://github.com/vmware-tanzu/velero/blob/0246a32ad0897053fbc444d86a0bec65762bc714/pkg/restore/restore.go#L160-L169
https://github.com/vmware-tanzu/velero/blob/0246a32ad0897053fbc444d86a0bec65762bc714/pkg/restore/restore.go#L769

skriss: also, sorry i'm late here with the review :/

stevesloka: nah no worries! Felt weird re-writing all that when it should have just been there (I should have asked).

internal/dag/gatewayapi_processor.go
stevesloka: The Selector is not a pointer so we can't check against that. The API docs say that if it's not specified, then it matches all. 

I could refactor this out into a method to make it cleaner though.

stevesloka: @sunjayBhatia I refactored this up a bit and I think reads much better.

internal/dag/builder_test.go
```diff
@@ -83,29 +83,91 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
...
 			objs: []interface{}{
-				gateway,
 				kuardService,
 				&gatewayapi_v1alpha1.HTTPRoute{
 					ObjectMeta: metav1.ObjectMeta{
 						Name:      "basic",
 						Namespace: "default",
 						Labels: map[string]string{
-							"app": "contour",
+							"app":  "contour",
```
skriss, line 127: let's add a test case where the route's labels don't match the gateway's selector and verify it doesn't get included

internal/dag/gatewayapi_processor.go
```diff
@@ -41,9 +43,59 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+			if matches {
+				// Empty Selector matches all routes.
+				validRoutes = append(validRoutes, route)
+			}
+		}
+
+		// Process all the routes that match this Gateway.
+		for _, validRoute := range validRoutes {
+			p.computeHTTPRoute(validRoute)
+		}
```
skriss, line 80: should this go after L81?

internal/dag/gatewayapi_processor.go
```diff
@@ -41,9 +43,59 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+		// are associated with the Gateway. An empty Selector matches all routes.
+		for _, listener := range p.source.gateway.Spec.Listeners {
+
+			matches, err := matchRoutes(listener.Routes.Selector, route.Labels)
+			if err != nil {
+				p.Errorf("error validating routes against Listener.Routes.Selector: %s", err)
+			}
+			if matches {
+				// Empty Selector matches all routes.
+				validRoutes = append(validRoutes, route)
```
skriss, line 73: can we break out of the listeners loop, once we've found that the route matches at least one?

internal/dag/gatewayapi_processor.go
```diff
@@ -41,9 +43,59 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+
+		// Process all the routes that match this Gateway.
+		for _, validRoute := range validRoutes {
+			p.computeHTTPRoute(validRoute)
+		}
+	}
+}
+
+// matchRoutes returns true if the selector matches the labels on the object or is not defined.
+func matchRoutes(selector metav1.LabelSelector, objLabels map[string]string) (bool, error) {
```
skriss, line 85: I'd call this `selectorMatches` or `matchLabels` or `matchesRoute` -- it's only checking a single object, and it's not really route-specific.

skriss (changes requested):

internal/dag/gatewayapi_processor.go
stevesloka: I like `selectorMatches`.

internal/dag/gatewayapi_processor.go
stevesloka: Yeah for now I think that's fine.

internal/dag/gatewayapi_processor.go
stevesloka: 👍

internal/dag/builder_test.go
stevesloka: 👍

skriss (approved):

sunjayBhatia (approved):

PR #3419, opened 2021-02-28
Document how to use a side-car authorization server

### LINKED ISSUE #3373: Client Authorization over localhost with Envoy sidecar
Currently configuring [client authorization](https://projectcontour.io/docs/v1.12.0/config/client-authorization/) requires using an ExtensionService object with a reference to a Service that resolves to Pod endpoints for an authorization server that implements the Envoy External Authorisation API.  

But, instead of running the authorization server as a separate set of Pods I would like to run it as a side-car with each Envoy Pod ([opa-envoy](https://github.com/open-policy-agent/opa-envoy-plugin)) accessed over localhost as this is much easier to scale and secure.

Perhaps ExtensionService could support a host and port to call as an alternative to a Service?

erwbgy:
Document how to connect to an authorization as a side-car alongside each Envoy Pod using an ExternalName Service for localhost.

Fixes #3373.


youngnick (approved):
LGTM, thanks.

skriss (approved):

PR #3422, opened 2021-03-01
internal/envoy/v3: Strip possible port from :authority header

sunjayBhatia:
This ensures we don't have to have routes match on a wildcard suffix.
Makes internal Contour logic simpler as we don't have to think about
a possible port in route matches any longer and would make adding
support for wildcard domains a little easier.

Downsides:
- More Lua
- Strips port from header that some application may care about?

Could also apply this change *only* for wildcard routes but it seemed
good to be consistent.


sunjayBhatia:
Context around this PR is here: https://github.com/projectcontour/contour/pull/3381#issuecomment-788086911

sunjayBhatia:
Realized this isn’t complete, the Alia filter is only added for TLS routes

sunjayBhatia:
Closing as this can be replaced by using this feature of the HTTP Connection Manager: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto.html#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-strip-any-host-port

PR #3426, opened 2021-03-01
CI: run integration tests against all supported Kubernetes versions

### LINKED ISSUE #3396: testing: run integration tests against all supported Kubernetes versions
Per https://projectcontour.io/resources/compatibility-matrix/, each version of Contour is considered compatible with three minor Kubernetes versions. However, our integration tests tend to run only against the latest version (based on whatever image the version of kind that we're using pulls by default).

We should add runs of the tests against the other two versions as well. Personally, I'd be OK with this starting as a nightly job, as I don't know that we *need* to run against all versions for every PR. Open to other inputs on this though.

skriss:
Runs integration tests against the three supported minor versions
of Kubernetes.

Closes #3396.

### DIFF
--- a/.github/workflows/prbuild.yaml
+++ b/.github/workflows/prbuild.yaml
@@ -8,6 +8,22 @@ env:
 jobs:
   integration:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # use stable kubernetes_version values since they're included
+        # in the name of the GitHub Actions job, and we don't want to
+        # have to update branch protection rules every time we change
+        # a Kubernetes version number.
+        kubernetes_version: ["kubernetes:latest", "kubernetes:n-1", "kubernetes:n-2"]
+        # include defines an additional variable (the specific node
+        # image to use) for each kubernetes_version value.
+        include:
+          - kubernetes_version: "kubernetes:latest"
+            node_image: "docker.io/kindest/node:v1.20.2"
+          - kubernetes_version: "kubernetes:n-1"
+            node_image: "docker.io/kindest/node:v1.19.7"
+          - kubernetes_version: "kubernetes:n-2"
+            node_image: "docker.io/kindest/node:v1.18.15"
     steps:
       - uses: actions/checkout@v2
       - name: add deps to path
@@ -15,6 +31,8 @@ jobs:
           ./hack/actions/install-kubernetes-toolchain.sh $GITHUB_WORKSPACE/bin
           echo "$GITHUB_WORKSPACE/bin" >> $GITHUB_PATH
       - name: integration tests
+        env:
+          NODEIMAGE: ${{ matrix.node_image }}
         run: |
           make integration
   # TODO: re-enable once Ingress v1 support is complete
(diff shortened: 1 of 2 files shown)


skriss:
This is the quick and dirty version of this. On the plus side, integration tests are run against each minor version in parallel, so it won't slow down CI for PRs. On the downside, the versions to test against are hardcoded and will have to be periodically updated. @projectcontour/maintainers WDYT? Good enough for now?

stevesloka:
Looks fine to me! I don't know of a good way around the hard coding unless we somehow could tie it to the support matrix so an update there updated this, but for now it's a one every three month change.

skriss:
> Looks fine to me! I don't know of a good way around the hard coding unless we somehow could tie it to the support matrix so an update there updated this, but for now it's a one every three month change.

Yeah, I agree, it seemed not worth the effort to try to automate. Can always revisit if we need to.

youngnick (approved):
LGTM, I think automating this more at the moment is probably premature. Nice one.

skriss:
will require a branch protection rules update when merged to account for the three new required job names and plain old `integration` being dropped.

sunjayBhatia (approved):

.github/workflows/prbuild.yaml
```diff
@@ -6,8 +6,36 @@ on: [push, pull_request]
 env:
   GOPROXY: https://proxy.golang.org/
 jobs:
-  integration:
+  integration-latest:
     runs-on: ubuntu-latest
+    env:
+      NODEIMAGE: docker.io/kindest/node:v1.20.2
+    steps:
```
sunjayBhatia, line 13: Oh just occurred to me could we use YAML anchors/references to DRY this out?

skriss: did some googling, AFAICT GH Actions does not support YAML anchors: https://github.community/t/support-for-yaml-anchors/16128

sunjayBhatia: ah thats no good :(

stevesloka (approved):

.github/workflows/prbuild.yaml
sunjayBhatia: maybe we could use this? https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

sunjayBhatia: could make it so we only have 1 "job" still but different steps within that

skriss: Switched to a matrix and it runs well. The only annoying thing is that now in branch protection rules, the required job names include the specific node image value, so we'll have to change those every time we change a Kube version we run against. Looking to see if there's a way around that..

sunjayBhatia: oh that is very annoying 👎🏽

.github/workflows/prbuild.yaml
```diff
@@ -8,13 +8,31 @@ env:
 jobs:
   integration:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # use stable kubernetes_version values since they're included
```
sunjayBhatia, line 13: pretty slick, nice!

skriss: :+1: I think this is the best of both worlds, thanks for pointing me in this direction. Now if only the jobs will run...

Issue #3432, opened 2021-03-03
grpc example


surajchhetry:
**grpc service and HttpProxy example**

sunjayBhatia:
Could you add some more detail to this issue and a more specific title to describe what this is about? So we have some historic record of the motivation behind this issue and can point people to it in the future if needed

sunjayBhatia:
closing in favor of #1894

PR #3433, opened 2021-03-03
 grpc httpProxy example added

surajchhetry:
this is an example of how HttpProxy can be configured to access GRPC backed service #3432

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -29,3 +29,7 @@ If you have your own Grafana and Prometheus deployment already, the supplied [Co
 ## `kind`, `root-rbac`
 
 Both of these examples are fragments used in other documentation ([deploy-options](https://projectcontour.io/docs/main/deploy-options))
+
+## `grpc`
+
+GRPC configuration with HTTPProxy.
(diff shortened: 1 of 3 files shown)


sunjayBhatia (changes requested):
Could you provide some more detail in the title and PR description about this change?

As I understand it, this is an example to show exactly how a GRPC service can be configured which was missing from our examples.

Also some comments in the example YAML about relevant portions would be great (why the `h2c` protocol is specifically chosen and when you might want to not use it, etc.)

surajchhetry:
@sunjayBhatia updated with the requested change.

stevesloka:
Looks good to me @surajchhetry! Could you look at the failures for the `codegen` job? Looks like a couple small spacing issues.

youngnick (approved):
I think this is a good start on further gRPC documentation, but eventually I'd like to see a better guide that covered more of the options (like using the service annotation vs the `protocol` field, `h2c` vs `h2` and so on).

Thanks @surajchhetry!

skriss:
@surajchhetry are you able to address the feedback in https://github.com/projectcontour/contour/pull/3433#issuecomment-793030837 so we can merge this?

PR #3163, opened 2020-11-30
_integration: temporarily disable flaky SNI test

skriss:
Disables 004-https-sni-enforcement until the cause for the flaky
behavior can be found and fixed.

Signed-off-by: Steve Kriss <krisss@vmware.com>

PR #3435, opened 2021-03-03
_integration: re-enable SNI test

skriss:
The SNI integration test was disabled in #3163
due to unexplained sporadic failures. The failures
have not been reproducible in recent testing. It's
possible that a change made to Contour since #3163
has eliminated the source of the test flake, but
also possible that the test is still flaky and just
did not reproduce during testing.


stevesloka (approved):
/shrug ship it? =)

sunjayBhatia (approved):

skriss:
> /shrug ship it? =)

Yep let's see what happens! I've stared at it on a few different occasions and don't see any issues, nor can I repro.

PR #3445, opened 2021-03-04
internal/envoy: use generic key descriptor's key

### LINKED ISSUE #3443: Contour ignores key field of genericKey for global rateLimitPolicy
When applying a rateLimitPolicy on an HTTPProxy using a `genericKey` entry, the `key` field of the `genericKey` is ignored and the default value (`generic_key`) is passed on to the RLS.

Example configuration (using a kuard deployment as example and the envoy ratelimit service).

*HTTPProxy*
```
kind: HTTPProxy
metadata:
  name: kuard
  labels:
    app: kuard
spec:
  virtualhost:
    fqdn: local.projectcontour.io
    rateLimitPolicy:
      global:
        descriptors:
          - entries:
            - genericKey:
                key: foo
                value: bar
  routes:
  - services:
    - name: kuard
      port: 80
```

*Envoy ratelimit config*
```
domain: contour
descriptors:
  - key: foo
    value: bar
    rate_limit:
      unit: minute
      requests_per_unit: 100
```

*Envoy ratelimit debug logs*
```
level=debug msg="starting cache lookup"
level=debug msg="returning normal response"
level=debug msg="got descriptor: (generic_key=bar)"
level=debug msg="starting get limit lookup"
level=debug msg="looking up key: generic_key_bar"
level=debug msg="looking up key: generic_key"
level=debug msg="descriptor does not match any limit, no limits applied"
```

Contour should set the genericKey key value properly and pass that value on to the RLS.
The RLS should get the following descriptor:
```
descriptor: (foo=bar)
```
instead of
```
descriptor: (generic_key=bar)
```

This happens for a RL policy defined on the virtualhost as well for a RL policy defined at the route.

**Environment:**

- Contour version: v1.13.1
- Kubernetes version: (use `kubectl version`): v1.19.4+k3s1
- Kubernetes installer & version: k3d
- Cloud provider or hardware configuration: k3d local test cluster
- OS (e.g. from `/etc/os-release`): /

skriss:
Fixes a bug where a generic key descriptor's key
was not being configured in the Envoy descriptor
if it was specified, resulting in Envoy always
generating the default key of "generic_key".

Fixes #3443.

### DIFF
--- a/_integration/testsuite/httpproxy/020-global-rate-limiting.yaml
+++ b/_integration/testsuite/httpproxy/020-global-rate-limiting.yaml
@@ -161,6 +161,9 @@ spec:
         descriptors:
           - entries:
               - genericKey:
+                  # specify a custom key here as an added test
+                  # that this is configured with Envoy if specified.
+                  key: route_limit_key
                   value: routelimit
   - conditions:
       - prefix: /unlimited
--- a/_integration/testsuite/install-ratelimit-service.sh
+++ b/_integration/testsuite/install-ratelimit-service.sh
@@ -46,7 +46,7 @@ data:
         rate_limit:
           unit: hour
           requests_per_unit: 1
-      - key: generic_key
+      - key: route_limit_key
         value: routelimit
         rate_limit:
           unit: hour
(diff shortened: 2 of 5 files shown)


sunjayBhatia (approved):

sunjayBhatia:
Is there any more integration testing we can add for this? Change one of these? https://github.com/projectcontour/contour/blob/13e48574b9c06b7c80642bcf942b34d7302de3ab/_integration/testsuite/install-ratelimit-service.sh#L43

skriss:
Yep, definitely can. There are about 5 different places where this *could* be tested. I'll add a few more test cases in some of those other layers.

stevesloka (approved):

PR #3448, opened 2021-03-05
internal/featuretests: use helper for default HTTP listener

skriss:
Replaces the many repeated definitions of the default HTTP listener
with calls to defaultHTTPListener(). In places where a non-default
HTTP listener is needed, uses the default as a starting point and
applies modifications to it. This reduces copy-pasted boilerplate.


sunjayBhatia (approved):
As far as I can tell looks like a correct 1-1 replacement, nice!

stevesloka (approved):

PR #3450, opened 2021-03-05
cmd/contour: only add Gateway API processor if API exists

### LINKED ISSUE #3449: gateway processor logs an error in a cluster that doesn't have Gateway API installed
Contour should continue to behave as it does today for users that don't have Gateway API installed in their cluster.

When running Contour in such a cluster, the following error-level log is generated by the Gateway API processor:

```
ERRO[0017] Gateway is not defined!                       context=GatewayAPIProcessor
```

There are several ways this could be resolved, but optimally it'd be nice to not even add the Gateway API processor to the DAG builder if the Gateway API isn't present in the cluster. That'd require reordering a few things in the server setup, but it's doable.

skriss:
Only adds the Gateway API DAG processor to the DAG
builder if the Gateway API resources exist in the
cluster.

Closes #3449.

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -407,6 +407,37 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	// register observer for endpoints updates.
 	endpointHandler.Observer = contour.ComposeObservers(snapshotHandler)
 
+	// Get the appropriate DAG processors.
+	dagProcessors := []dag.Processor{
+		&dag.IngressProcessor{
+			FieldLogger:       log.WithField("context", "IngressProcessor"),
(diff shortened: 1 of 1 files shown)


sunjayBhatia (approved):
LGTM

Sad we don't have direct coverage of `cmd/contour/...` since theres increasing logic, esp. in `serve.go`

skriss:
> LGTM
> 
> Sad we don't have direct coverage of `cmd/contour/...` since theres increasing logic, esp. in `serve.go`

Yeah, a couple of these recent PRs led me to start looking if we could do some refactoring of `serve.go` and also add test coverage to it.

stevesloka (approved):

PR #3456, opened 2021-03-08
internal: Allow status on HTTPRoutes to be set

### LINKED ISSUE #3436: Set status on HTTPRoute resources
Need to set the "admitted" status on a route along with the matching Gateway. Also, errors are bubbled up here into the `conditions` set. 

> Gateways is a list of Gateways that are associated with the route, and the status of the route with respect to each Gateway. When a Gateway selects this route, the controller that manages the Gateway must add an entry to this list when the controller first sees the route and should update the entry as appropriate when the route is modified.

A maximum of 100 Gateways will be represented in this list. If this list is full, there may be additional Gateways using this Route that are not included in the list. An empty list means the route has not been admitted by any Gateway.

Conditions describes the status of the route with respect to the Gateway. The “Admitted” condition must always be specified by controllers to indicate whether the route has been admitted or rejected by the Gateway, and why. Note that the route’s availability is also subject to the Gateway’s own status conditions and listener status.

//ref: https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.RouteGatewayStatus

stevesloka:
- Update the status package to process GatewayAPI HTTPRoutes.
- Update the dag.GatewayAPIProcessor to set status.

Fixes #3436

### DIFF
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -65,14 +65,12 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
 
 		// Validate the Group on the selector is a supported type.
 		if listener.Routes.Group != "" && listener.Routes.Group != gatewayapi_v1alpha1.GroupName {
-			// TODO: Set the “ResolvedRefs” condition to false for this listener with the “InvalidRoutesRef” reason.
 			p.Errorf("Listener.Routes.Group %q is not supported.", listener.Routes.Group)
 			continue
 		}
 
 		// Validate the Kind on the selector is a supported type.
 		if listener.Routes.Kind != KindHTTPRoute {
-			// TODO: Set the “ResolvedRefs” condition to false for this listener with the “InvalidRoutesRef” reason.
 			p.Errorf("Listener.Routes.Kind %q is not supported.", listener.Routes.Kind)
 			continue
 		}
@@ -173,7 +171,7 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
 
 	// Validate TLS Configuration
 	if route.Spec.TLS != nil {
-		routeAccessor.ConditionFor(status.NotImplemented, "NotImplemented", "HTTPRoute.Spec.TLS: Not yet implemented.")
+		routeAccessor.AddCondition(status.ConditionNotImplemented, metav1.ConditionTrue, status.ReasonNotImplemented, "HTTPRoute.Spec.TLS: Not yet implemented.")
 	}
 
 	// Determine the hosts on the route, if no hosts
@@ -215,13 +213,12 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
 			case gatewayapi_v1alpha1.PathMatchPrefix:
 				pathPrefixes = append(pathPrefixes, stringOrDefault(match.Path.Value, "/"))
 			default:
-				routeAccessor.ConditionFor(status.Invalid, "PathMatchType", "HTTPRoute.Spec.Rules.PathMatch: Only Prefix match type is supported.")
+				routeAccessor.AddCondition(status.ConditionInvalid, metav1.ConditionTrue, status.ReasonPathMatchType, "HTTPRoute.Spec.Rules.PathMatch: Only Prefix match type is supported.")
 			}
 		}
 
 		if len(rule.ForwardTo) == 0 {
-			// TODO: Raise `ResolvedRefs` condition on Gateway with `DegradedRoutes` reason (#3455).
-			routeAccessor.ConditionFor(status.Invalid, "ForwardTo", "At least one Spec.Rules.ForwardTo must be specified.")
+			routeAccessor.AddCondition(status.ConditionInvalid, metav1.ConditionTrue, status.ReasonForwardTo, "At least one Spec.Rules.ForwardTo must be specified.")
 			continue
 		}
 
@@ -229,13 +226,13 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
 		for _, forward := range rule.ForwardTo {
 			// Verify the service is valid
 			if forward.ServiceName == nil {
-				routeAccessor.ConditionFor(status.Invalid, "ForwardTo", "Spec.Rules.ForwardTo.ServiceName must be specified.")
+				routeAccessor.AddCondition(status.ConditionInvalid, metav1.ConditionTrue, status.ReasonForwardTo, "Spec.Rules.ForwardTo.ServiceName must be specified.")
 				continue
 			}
 
 			// TODO: Do not require port to be present (#3352).
 			if forward.Port == nil {
-				routeAccessor.ConditionFor(status.Invalid, "ForwardTo", "Spec.Rules.ForwardTo.ServicePort must be specified.")
+				routeAccessor.AddCondition(status.ConditionInvalid, metav1.ConditionTrue, status.ReasonForwardTo, "Spec.Rules.ForwardTo.ServicePort must be specified.")
 				continue
 			}
 
@@ -244,15 +241,14 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
 			// TODO: Refactor EnsureService to take an int32 so conversion to intstr is not needed.
 			service, err := p.dag.EnsureService(meta, intstr.FromInt(int(*forward.Port)), p.source)
 			if err != nil {
-				// TODO: Raise `ResolvedRefs` condition on Gateway with `DegradedRoutes` reason (#3455).
-				routeAccessor.ConditionFor(status.Invalid, "ForwardTo", fmt.Sprintf("Service %q does not exist in namespace %q", meta.Name, meta.Namespace))
+				routeAccessor.AddCondition(status.ConditionInvalid, metav1.ConditionTrue, status.ReasonForwardTo, fmt.Sprintf("Service %q does not exist in namespace %q", meta.Name, meta.Namespace))
 				continue
 			}
 			services = append(services, service)
 		}
 
 		if len(services) == 0 {
-			routeAccessor.ConditionFor(status.Invalid, "ForwardTo", "All Spec.Rules.ForwardTos are invalid.")
+			routeAccessor.AddCondition(status.ConditionInvalid, metav1.ConditionTrue, status.ReasonForwardTo, "All Spec.Rules.ForwardTos are invalid.")
 			continue
 		}
 
@@ -269,9 +265,9 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
 	// condition accordingly.
 	switch len(routeAccessor.Conditions) {
 	case 0:
-		routeAccessor.ConditionFor(status.AdmittedConditionValid, "Valid", "Valid HTTPRoute")
+		routeAccessor.AddCondition(gatewayapi_v1alpha1.ConditionRouteAdmitted, metav1.ConditionTrue, status.ReasonValid, "Valid HTTPRoute")
 	default:
-		routeAccessor.ConditionFor(status.AdmittedConditionInvalid, "ErrorsExist", "Errors Found, check other Conditions for details.")
+		routeAccessor.AddCondition(gatewayapi_v1alpha1.ConditionRouteAdmitted, metav1.ConditionFalse, status.ReasonErrorsExist, "Errors Found, check other Conditions for details.")
 	}
 }
 
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -218,7 +218,7 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
 		}
 
 		if len(rule.ForwardTo) == 0 {
-			routeAccessor.AddCondition(status.ConditionResolvedRefs, metav1.ConditionTrue, status.ReasonDegradedRoutes, "At least one Spec.Rules.ForwardTo must be specified.")
+			routeAccessor.AddCondition(status.ConditionResolvedRefs, metav1.ConditionTrue, status.ReasonDegraded, "At least one Spec.Rules.ForwardTo must be specified.")
 			continue
 		}
 
@@ -226,13 +226,13 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
 		for _, forward := range rule.ForwardTo {
 			// Verify the service is valid
 			if forward.ServiceName == nil {
-				routeAccessor.AddCondition(status.ConditionResolvedRefs, metav1.ConditionTrue, status.ReasonDegradedRoutes, "Spec.Rules.ForwardTo.ServiceName must be specified.")
+				routeAccessor.AddCondition(status.ConditionResolvedRefs, metav1.ConditionTrue, status.ReasonDegraded, "Spec.Rules.ForwardTo.ServiceName must be specified.")
 				continue
 			}
 
 			// TODO: Do not require port to be present (#3352).
 			if forward.Port == nil {
-				routeAccessor.AddCondition(status.ConditionResolvedRefs, metav1.ConditionTrue, status.ReasonDegradedRoutes, "Spec.Rules.ForwardTo.ServicePort must be specified.")
+				routeAccessor.AddCondition(status.ConditionResolvedRefs, metav1.ConditionTrue, status.ReasonDegraded, "Spec.Rules.ForwardTo.ServicePort must be specified.")
 				continue
 			}
 
@@ -241,14 +241,14 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
 			// TODO: Refactor EnsureService to take an int32 so conversion to intstr is not needed.
 			service, err := p.dag.EnsureService(meta, intstr.FromInt(int(*forward.Port)), p.source)
 			if err != nil {
-				routeAccessor.AddCondition(status.ConditionResolvedRefs, metav1.ConditionTrue, status.ReasonDegradedRoutes, fmt.Sprintf("Service %q does not exist in namespace %q", meta.Name, meta.Namespace))
+				routeAccessor.AddCondition(status.ConditionResolvedRefs, metav1.ConditionTrue, status.ReasonDegraded, fmt.Sprintf("Service %q does not exist in namespace %q", meta.Name, meta.Namespace))
 				continue
 			}
 			services = append(services, service)
 		}
 
 		if len(services) == 0 {
-			routeAccessor.AddCondition(status.ConditionResolvedRefs, metav1.ConditionTrue, status.ReasonDegradedRoutes, "All Spec.Rules.ForwardTos are invalid.")
+			routeAccessor.AddCondition(status.ConditionResolvedRefs, metav1.ConditionTrue, status.ReasonDegraded, "All Spec.Rules.ForwardTos are invalid.")
 			continue
 		}
 
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -218,7 +218,7 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
 		}
 
 		if len(rule.ForwardTo) == 0 {
-			routeAccessor.AddCondition(status.ConditionResolvedRefs, metav1.ConditionTrue, status.ReasonDegraded, "At least one Spec.Rules.ForwardTo must be specified.")
+			routeAccessor.AddCondition(status.ConditionResolvedRefs, metav1.ConditionFalse, status.ReasonDegraded, "At least one Spec.Rules.ForwardTo must be specified.")
 			continue
 		}
 
@@ -226,13 +226,13 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
 		for _, forward := range rule.ForwardTo {
 			// Verify the service is valid
 			if forward.ServiceName == nil {
-				routeAccessor.AddCondition(status.ConditionResolvedRefs, metav1.ConditionTrue, status.ReasonDegraded, "Spec.Rules.ForwardTo.ServiceName must be specified.")
+				routeAccessor.AddCondition(status.ConditionResolvedRefs, metav1.ConditionFalse, status.ReasonDegraded, "Spec.Rules.ForwardTo.ServiceName must be specified.")
 				continue
 			}
 
 			// TODO: Do not require port to be present (#3352).
 			if forward.Port == nil {
-				routeAccessor.AddCondition(status.ConditionResolvedRefs, metav1.ConditionTrue, status.ReasonDegraded, "Spec.Rules.ForwardTo.ServicePort must be specified.")
+				routeAccessor.AddCondition(status.ConditionResolvedRefs, metav1.ConditionFalse, status.ReasonDegraded, "Spec.Rules.ForwardTo.ServicePort must be specified.")
 				continue
 			}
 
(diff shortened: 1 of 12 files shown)


internal/status/cache.go
```diff
@@ -16,27 +16,41 @@
...
 
 // NewCache creates a new Cache for holding status updates.
-func NewCache() Cache {
+func NewCache(gateway types.NamespacedName) Cache {
 	return Cache{
-		proxyUpdates: make(map[types.NamespacedName]*ProxyUpdate),
-		entries:      make(map[string]map[types.NamespacedName]CacheEntry),
+		proxyUpdates:     make(map[types.NamespacedName]*ProxyUpdate),
+		gatewayRef:       gateway,
+		httpRouteUpdates: make(map[types.NamespacedName]*HTTPRouteUpdate),
```
skriss, line 52: is it possible to use the generic `entries` structure to store the HTTPRoute data, like with `ExtensionService`? If I remember correctly, I think there was a desire to move to using that. See also https://github.com/projectcontour/contour/blob/main/internal/status/extensionstatus.go#L108 and https://github.com/projectcontour/contour/blob/main/internal/dag/extension_processor.go#L44

stevesloka: The problem I ran into was that the `CacheEntry` interface has a `ConditionFor` method returning a `DetailedCondition` type which is specific to HTTPProxy and GatewayAPI uses a `metav1.Condition` type.

skriss: Ah, yeah that makes sense. I'll take another look with that in mind.

stevesloka: We could potentially refactor this into something similar to how it works with ExtensionService since we'll also need to set status on TLSRotues & Gateways.

skriss:
has a conflict to resolve now.

stevesloka:
All rebased @skriss, there are probably more spots to set status now that more features have been added in the last 2 weeks. I can look to add them, but won't change the overall code base.

skriss:
:+1: I'll try to get through this this afternoon, sorry for the delay

stevesloka:
> sorry for the delay

No worries! 🌮

internal/dag/gatewayapi_processor.go
```diff
@@ -149,10 +150,12 @@ func selectorMatches(selector metav1.LabelSelector, objLabels map[string]string)
 }
 
 func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRoute) {
+	routeAccessor, commit := p.dag.StatusCache.HTTPRouteAccessor(route)
+	defer commit()
 
 	// Validate TLS Configuration
 	if route.Spec.TLS != nil {
-		p.Error("NOT IMPLEMENTED: The 'RouteTLSConfig' is not yet implemented.")
+		routeAccessor.ConditionFor(status.NotImplemented, "NotImplemented", "HTTPRoute.Spec.TLS: Not yet implemented.")
```
sunjayBhatia, line 158: Maybe worth calling out somewhere that we are adding additional conditions with details about why a Route is invalid, I forget if the original design doc mentioned this specifically? Or if it was assumed the reason/description on the "Admitted" reason would have all this info?

sunjayBhatia: Another point for not using status as message passing is that it could be relatively unwieldy it seems, maybe keying off of reasons is good enough, but if you need more detail than that you would want some structured description (JSON probably?)

internal/dag/status_test.go
```diff
@@ -2441,3 +2448,417 @@ func TestDAGStatus(t *testing.T) {
...
+						FieldLogger: fixture.NewTestLogger(t),
+					},
+					&ListenerProcessor{},
+				},
+			}
+			for _, o := range tc.objs {
+				builder.Source.Insert(o)
+			}
+			dag := builder.Build()
+			t.Logf("%#v\n", dag.StatusCache)
```
sunjayBhatia, line 2499: nit: Do we want to remove this or keep it?

internal/dag/gatewayapi_processor.go
```diff
@@ -194,44 +197,44 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
 			case gatewayapi_v1alpha1.PathMatchPrefix:
 				pathPrefixes = append(pathPrefixes, stringOrDefault(match.Path.Value, "/"))
 			default:
-				p.Error("NOT IMPLEMENTED: Only PathMatchPrefix is currently implemented.")
+				routeAccessor.ConditionFor(status.Invalid, "PathMatchType", "HTTPRoute.Spec.Rules.PathMatch: Only Prefix match type is supported.")
```
sunjayBhatia, line 200: Do we want to pull the reasons into constants as well or since these are temporary we're ok with leaving them as is?

internal/dag/gatewayapi_processor.go
```diff
@@ -194,44 +197,44 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
...
+			continue
+		}
+
 		// Validate the ForwardTos.
-		var forwardTos []gatewayapi_v1alpha1.HTTPRouteForwardTo
 		for _, forward := range rule.ForwardTo {
 			// Verify the service is valid
 			if forward.ServiceName == nil {
-				p.Error("ServiceName must be specified and is currently only type implemented!")
+				routeAccessor.ConditionFor(status.Invalid, "ForwardTo", "Spec.Rules.ForwardTo.ServiceName must be specified.")
```
sunjayBhatia, line 214: This reason seems like it would be good to have as a constant?

internal/status/httproutestatus.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func (c *Cache) commitHTTPRoute(pu *HTTPRouteUpdate) {
+	if len(pu.Conditions) == 0 {
+		return
+	}
+	c.httpRouteUpdates[pu.Fullname] = pu
+}
+
+func (pu *HTTPRouteUpdate) Mutate(obj interface{}) interface{} {
```
sunjayBhatia, line 79: tiniest of nits: seems like `pu` as a variable name is used in various places and might have been copied from the proxy update type, not a big deal to leave in

internal/dag/gatewayapi_processor.go
sunjayBhatia: Same logic as the conditions I guess, constraining what could be passed in with a type

sunjayBhatia (commented):
Overall looks good and straightforward to me! Just some comments on some specifics but generally LGTM

internal/dag/gatewayapi_processor.go
skriss: Echoing @sunjayBhatia's q on the conditions, "Admitted" is the only type that Gateway API documents, so should we only use that? I'm not sure, I guess it depends on who is consuming the info.

internal/dag/gatewayapi_processor.go
```diff
@@ -243,6 +246,15 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
...
 	}
+
+	// Determine if any errors exist in conditions and set the "Admitted"
+	// condition accordingly.
+	switch len(routeAccessor.Conditions) {
+	case 0:
+		routeAccessor.ConditionFor(status.AdmittedConditionValid, "Valid", "Valid HTTPRoute")
+	default:
+		routeAccessor.ConditionFor(status.AdmittedConditionInvalid, "ErrorsExist", "Errors Found, check other Conditions for details.")
+	}
```
skriss, line 257: Ah, I see now that you're setting the `Admitted` condition here.

internal/status/cache.go
```diff
@@ -16,27 +16,41 @@
...
-	ProxyStatusOrphaned ProxyStatus = "orphaned"
-)
+// ValidCondition is the ConditionType for Valid.
+const ValidCondition ConditionType = "Valid"
+
+// AdmittedConditionValid is the valid ConditionType for Admitted.
+const AdmittedConditionValid ConditionType = "AdmittedValid"
+
+// AdmittedConditionInValid is the invalid ConditionType for Admitted.
+const AdmittedConditionInvalid ConditionType = "AdmittedInvalid"
```
skriss, line 39: this was slightly confusing because these aren't actually condition types; they correspond to the different states ("True" or "False" for the `Admitted` condition type). Thinking about suggestions here.

internal/status/httproutestatus.go
```diff
@@ -0,0 +1,143 @@
...
+// limitations under the License.
+
+package status
+
+import (
+	"fmt"
+	"time"
+
+	contour_api_v1 "github.com/projectcontour/contour/apis/projectcontour/v1"
+	v1 "github.com/projectcontour/contour/apis/projectcontour/v1"
```
skriss, line 21: nit: de-dupe with above import

internal/status/httproutestatus.go
```diff
@@ -0,0 +1,143 @@
...
+
+type HTTPRouteUpdate struct {
+	Fullname           types.NamespacedName
+	Conditions         []metav1.Condition
+	ExistingConditions []metav1.Condition
+	GatewayRef         types.NamespacedName
+}
+
+// ConditionFor returns a v1.Condition for a given ConditionType.
+func (pu *HTTPRouteUpdate) ConditionFor(cond ConditionType, reason string, message string) metav1.Condition {
```
skriss, line 36: WDYT about having this function also take a `status` as an arg, so instead of having the "AdmittedValid" and "AdmittedInvalid" psuedo-condition types, you can just explicitly pass "Admitted", "True" or "Admitted", "False" from the call sites?

internal/status/cache.go
skriss: Yeah, I'm not thrilled that we now have 3 slightly different ways of doing status updates for 3 different types, so let's definitely look for opportunities to consolidate. I'd need to look in some more detail at all of them to have suggestions though.

skriss (commented):
overall looks good, handful of comments

internal/dag/gatewayapi_processor.go
youngnick: So, it's okay for implementations to add their own Conditions, but they should roll up into the `Admitted` condition. I think that Steve has handled that here in another way, but I think it would be best to be clear about what we're doing, and make sure that the status gets rolled up into the `Admitted` condition at the end (we can use the `commit` function to do that).

I think that `ConditionFor` is not a good name for this one, as it's really doing an `AddCondition` to the proxy update.

internal/dag/gatewayapi_processor.go
youngnick: In general, I agree that we should make a set of defined Reasons for any condition we're using. As you say, because this is temporary though, we can probably not worry for now. Depends on how we update `ConditionFor`.

internal/dag/gatewayapi_processor.go
youngnick: Yeah, this is the rollup I was talking about. I think it's definitely worth some comments/explanations that this is what we are doing, as it's hard to see the flow as it stands. I also think that we need to try to use the upstream Reasons wherever we can, as if we don't we will diverge from the standards. We can, of course, push updates to the upstream reasons if we need to.

internal/dag/status_test.go
youngnick: I vote remove.

youngnick (changes requested):
This is looking pretty good, although I have two broad things I think need fixing:
- It's vital that we get the ObservedGeneration behavior right.
- If we're going to generate extra Conditions per GatewayRef, we have to document what they are going to be, and provide a type with constants for the `type` and `reason` fields, and document what they're for when we use them, just like the upstream gateway-apis code does.

I think that having extra conditions is great, as long as we make sure we're doing the Condition contract as expected by the upstream code.

internal/k8s/kind.go
```diff
@@ -48,6 +49,8 @@ func KindOf(obj interface{}) string {
 			return "ExtensionService"
 		case *unstructured.Unstructured:
 			return obj.GetKind()
+		case *gatewayapi_v1alpha1.HTTPRoute:
```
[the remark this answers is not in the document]
stevesloka: 👍

internal/dag/gatewayapi_processor.go
stevesloka: If we only use `Admitted` then we can only set a single status on the resource? I was under the impression that since it's a set, we'd want to list out all the issues with a route and not just a single.

internal/dag/status_test.go
stevesloka: removed.

internal/status/httproutestatus.go
stevesloka: 👍

internal/status/httproutestatus.go
stevesloka: 👍

stevesloka:
I think I've addressed everyone's comments, can you all take a look again? Thanks!

internal/dag/gatewayapi_processor.go
```diff
@@ -212,44 +213,42 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
 			case gatewayapi_v1alpha1.PathMatchPrefix:
 				pathPrefixes = append(pathPrefixes, stringOrDefault(match.Path.Value, "/"))
 			default:
-				p.Error("NOT IMPLEMENTED: Only PathMatchPrefix is currently implemented.")
+				routeAccessor.AddCondition(status.ConditionInvalid, metav1.ConditionTrue, status.ReasonPathMatchType, "HTTPRoute.Spec.Rules.PathMatch: Only Prefix match type is supported.")
```
skriss, line 216: I think `status.ConditionNotImplemented` would make more sense here

internal/dag/gatewayapi_processor.go
```diff
@@ -261,6 +260,15 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
...
 		}
 	}
+
+	// Determine if any errors exist in conditions and set the "Admitted"
+	// condition accordingly.
+	switch len(routeAccessor.Conditions) {
+	case 0:
+		routeAccessor.AddCondition(gatewayapi_v1alpha1.ConditionRouteAdmitted, metav1.ConditionTrue, status.ReasonValid, "Valid HTTPRoute")
+	default:
+		routeAccessor.AddCondition(gatewayapi_v1alpha1.ConditionRouteAdmitted, metav1.ConditionFalse, status.ReasonErrorsExist, "Errors Found, check other Conditions for details.")
```
skriss, line 270:
```suggestion
		routeAccessor.AddCondition(gatewayapi_v1alpha1.ConditionRouteAdmitted, metav1.ConditionFalse, status.ReasonErrorsExist, "Errors found, check other Conditions for details.")
```

internal/status/cache.go
skriss: Probably worth filing a follow-up issue to look at cleaning some of this up.

internal/status/httproutestatus.go
```diff
@@ -0,0 +1,151 @@
...
+type RouteReasonType string
+
+const ReasonNotImplemented RouteReasonType = "NotImplemented"
+const ReasonPathMatchType RouteReasonType = "PathMatchType"
+const ReasonForwardTo RouteReasonType = "ForwardTo"
+const ReasonValid RouteReasonType = "Valid"
+const ReasonErrorsExist RouteReasonType = "ErrorsExist"
+
+type HTTPRouteUpdate struct {
+	Fullname           types.NamespacedName
```
skriss, line 39:
```suggestion
	FullName           types.NamespacedName
```

internal/status/httproutestatus.go
```diff
@@ -0,0 +1,151 @@
...
+
+package status
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/projectcontour/contour/internal/k8s"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
skriss, line 22: nit: double import

internal/status/httproutestatus.go
```diff
@@ -0,0 +1,151 @@
...
+	for _, cond := range routeUpdate.Conditions {
+
+		// Look through the newly calculated conditions, if they already exist
+		// on the object or if our obervation is stale, then leave them be,
+		// otherwise add them to the list to be written back to the API.
+		for _, existingCond := range routeUpdate.ExistingConditions {
+			if (cond.Status == existingCond.Status &&
+				cond.Reason == existingCond.Reason &&
+				cond.Message == existingCond.Message &&
+				cond.Type == existingCond.Type) ||
```
skriss, line 108: Should this be an `&&` instead of an `||`? If not, them I'm not quite understanding the logic here.

internal/dag/gatewayapi_processor.go
stevesloka: 👍

internal/dag/gatewayapi_processor.go
stevesloka: 👍

internal/status/httproutestatus.go
stevesloka: 👍

internal/status/httproutestatus.go
stevesloka: 👍

internal/status/httproutestatus.go
stevesloka: My logic was that if the object is the same, skip it. Or if the generation is out of date, skip it.

internal/status/httproutestatus.go
```diff
@@ -0,0 +1,150 @@
...
+				existingCond.ObservedGeneration > cond.ObservedGeneration {
+
+				cond.ObservedGeneration = existingCond.ObservedGeneration
+				cond.LastTransitionTime = existingCond.LastTransitionTime
+			}
+		}
+
+		conditionsToWrite = append(conditionsToWrite, cond)
```
skriss, line 115: OK, I took a look at the `HTTPProxy` equivalent here, and I think there are a couple things:

- at the top of the loop, we should *always* set `cond.ObservedGeneration = routeUpdate.Generation` and `cond.LastTransitionTime = routeUpdate.LastTransitionTime`
- the inner loop should be looking for an existing Condition on the HTTPRoute that's got the same `Type`, but a newer observed generation. If we find one, that means `cond` is stale, so we should keep the existing one instead. If we don't find a newer existing one, then we should use `cond`.

So I think this should look more like (I wrote this in GitHub so it's definitely not 100% right):

```go
for _, cond := range routeUpdate.Conditions {
        // set the Condition's observed generation based on 
        // the generation of the HTTPRoute we looked at.
        cond.ObservedGeneration = routeUpdate.Generation
        cond.LastTransitionTime = routeUpdate.TransitionTime

		// is there a newer Condition on the HTTPRoute matching
		// this condition's type? If so, our observation is stale,
		// so don't write it, keep the newer one instead.
		var newerConditionExists bool
		for _, existingCond := routeUpdate.ExistingConditions {
			if existingCond.Type != cond.Type {
				continue
			}
			
			if existingCond.ObservedGeneration > cond.ObservedGeneration {
				conditionsToWrite = append(conditionsToWrite, existingCond)
				newerConditionExists = true
				break
			}
		}

		// if we didn't find a newer version of the Condition on the
		// HTTPRoute, then write the one we computed.
		if !newerConditionExists {
			conditionsToWrite = append(conditionsToWrite, cond)
		}
	}
```

internal/status/httproutestatus.go
```diff
@@ -0,0 +1,150 @@
...
+	FullName           types.NamespacedName
+	Conditions         []metav1.Condition
+	ExistingConditions []metav1.Condition
+	GatewayRef         types.NamespacedName
+	Generation         int64
+	TransitionTime     metav1.Time
+}
+
+// AddCondition returns a metav1.Condition for a given ConditionType.
+func (routeUpdate *HTTPRouteUpdate) AddCondition(cond gatewayapi_v1alpha1.RouteConditionType, status metav1.ConditionStatus, reason RouteReasonType, message string) metav1.Condition {
```
youngnick, line 47: This method really should do some checking by `cond` to see if there's an existing one, since conditions are supposed to be a List that's keyed of the Type. Not a blocker on this one, maybe a TODO though?

youngnick (commented):
This mostly LGTM, I think that Steve's suggestion is great.

internal/status/httproutestatus.go
stevesloka: Ohh you can't have multiple conditions of the same type? What if you had two different missing services in a ForwardTo? Or on two different paths in the same HTTPRoute?

youngnick: You'll have to concatenate the details into the Message, but yeah, the Conditions is a ListTypeMap, meaning it's a map that's implemented as a list.

youngnick (approved):
LGTM

internal/dag/status_test.go
```diff
@@ -2441,3 +2448,418 @@ func TestDAGStatus(t *testing.T) {
...
+					},
+				},
+				Spec: gatewayapi_v1alpha1.HTTPRouteSpec{
+					Hostnames: []gatewayapi_v1alpha1.Hostname{
+						"test.projectcontour.io",
+					},
+					Rules: []gatewayapi_v1alpha1.HTTPRouteRule{{
+						Matches: []gatewayapi_v1alpha1.HTTPRouteMatch{{
+							Path: gatewayapi_v1alpha1.HTTPPathMatch{
+								Type:  "Exact", // <---- exact type not yet supported
```
sunjayBhatia, line 2657: We'll have to be careful around the order of merging this PR and  https://github.com/projectcontour/contour/pull/3478

internal/status/httproutestatus_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"testing"
+
+	projectcontour "github.com/projectcontour/contour/apis/projectcontour/v1"
+	"github.com/projectcontour/contour/internal/k8s"
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	gatewayapi_v1alpha1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
+)
+
+func TestHTTPRouteAddCondition(t *testing.T) {
```
sunjayBhatia, line 26: Worth adding more testing around conditions not being updated if the existing conditions are more recent than the attempted update?

internal/status/httproutestatus.go
skriss: I wonder what happens if you *do* add multiple instances of the same Type, since the current code can definitely do that.

skriss (commented):
mostly LGTM, couple comments on true vs. false conditions

internal/dag/gatewayapi_processor.go
```diff
@@ -212,44 +213,42 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
 			case gatewayapi_v1alpha1.PathMatchPrefix:
 				pathPrefixes = append(pathPrefixes, stringOrDefault(match.Path.Value, "/"))
 			default:
-				p.Error("NOT IMPLEMENTED: Only PathMatchPrefix is currently implemented.")
+				routeAccessor.AddCondition(status.ConditionNotImplemented, metav1.ConditionTrue, status.ReasonPathMatchType, "HTTPRoute.Spec.Rules.PathMatch: Only Prefix match type is supported.")
 			}
 		}
 
+		if len(rule.ForwardTo) == 0 {
+			routeAccessor.AddCondition(status.ConditionResolvedRefs, metav1.ConditionTrue, status.ReasonDegraded, "At least one Spec.Rules.ForwardTo must be specified.")
```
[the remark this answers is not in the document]
stevesloka: Arg this is so confusing. In my head "False" would mean that it's NOT in that condition (but maybe it was). 🤯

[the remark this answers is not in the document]
skriss: It probably would've been easier if the upstream condition was "UnresolvedRefs", not "ResolvedRefs".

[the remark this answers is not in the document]
stevesloka: `confusing: true`

internal/status/httproutestatus.go
stevesloka: Right now you get multiple conditions since I thought you wouldn't want to concat but you'd want to have the details. 

But like it seems with this API, my thought of how it should work is completely opposite. I'll have to come up with some concat thing to fix this since I guess I'm not following the spec in the current impl.

stevesloka: Here's an example of that: https://github.com/projectcontour/contour/pull/3456/files#diff-72b53646a46f97f3db8eabaf70aa271e571f0a3e51c73a1ba593abd85d53bfebR2761-R2776

skriss: Is this OK to merge as-is for 1.14? The current code can definitely potentially add multiple conditions of the same type, and i'm not sure what the Kube behavior is in that scenario - does it accept all of them? IMHO, given Gateway support is all alpha/WIP, I think we're probably OK as long as it doesn't cause Contour to crash or anything like that.

internal/status/httproutestatus_test.go
skriss: I think we definitely should, though personally I'd be OK merging this PR as-is and adding more testing in subsequent PRs, since this area is still WIP and this is all alpha anyway.

skriss (approved):
I think this is probably OK to merge now, had a couple comments/questions on open threads though.

sunjayBhatia:
> > Is this OK to merge as-is for 1.14?
> 
> I think it's fine to merge and gives us a set of status information which is helpful. I agree that there's lots of room for improvement. =)

Could you add issues and/or TODOs for the bits that have open questions/discussion threads so we don't lose them?

internal/status/httproutestatus_test.go
sunjayBhatia: https://github.com/projectcontour/contour/issues/3533

internal/status/httproutestatus.go
skriss: added #3534 to track this

PR #3457, opened 2021-03-08
cmd/contour: add EnableXRateLimitHeaders config file setting

### LINKED ISSUE #3431: global rate limiting: allow configuring enable_x_ratelimit_headers
Just checking, can our rate limiting feature querying capabilities like checking number of requests left and the time period in effect for current rate limits until they are reset.  

Seems Envoy provide some of these [APIs](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ratelimit/v3/rate_limit.proto#envoy-v3-api-msg-extensions-filters-http-ratelimit-v3-ratelimit) in v3  

(extensions.filters.http.ratelimit.v3.RateLimit.XRateLimitHeadersRFCVersion) Defines the standard version to use for X-RateLimit headers emitted by the filter:

X-RateLimit-Limit - indicates the request-quota associated to the client in the current time-window followed by the description of the quota policy. The values are returned by the rate limiting service in current_limit field. Example: 10, 10;w=1;name=”per-ip”, 1000;w=3600.
X-RateLimit-Remaining - indicates the remaining requests in the current time-window. The values are returned by the rate limiting service in limit_remaining field.
X-RateLimit-Reset - indicates the number of seconds until reset of the current time-window. The values are returned by the rate limiting service in duration_until_reset field.

skriss:
Adds the EnableXRateLimitHeaders field to the config file's
RateLimitService block. When set to true, adds the X-RateLimit
headers to responses that required checking the RLS.

Closes #3431.

### DIFF
--- a/examples/contour/01-contour-config.yaml
+++ b/examples/contour/01-contour-config.yaml
@@ -120,6 +120,12 @@ data:
     #   service fails to respond with a valid rate limit decision within
     #   the timeout defined on the extension service.
     #   failOpen: false
+    #   Defines whether to include the X-RateLimit headers X-RateLimit-Limit,
+    #   X-RateLimit-Remaining, and X-RateLimit-Reset (as defined by the IETF
+    #   Internet-Draft linked below), on responses to clients when the Rate
+    #   Limit Service is consulted for a request.
+    #   ref. https://tools.ietf.org/id/draft-polli-ratelimit-headers-03.html
+    #   enableXRateLimitHeaders: false
     #
     # Global Policy settings.
     # policy:
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -350,10 +350,11 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		}
 
 		listenerConfig.RateLimitConfig = &xdscache_v3.RateLimitConfig{
-			ExtensionService: namespacedName,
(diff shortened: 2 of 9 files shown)


skriss:
I assumed that folks agreed with https://github.com/projectcontour/contour/issues/3431#issuecomment-791494276, but I'd still appreciate input one way or another on that. Easy enough to change things if we want to switch to a non-boolean config option.

sunjayBhatia (approved):
LGTM assuming the boolean option is cool with everyone

stevesloka (approved):

examples/contour/01-contour-config.yaml
```diff
@@ -120,6 +120,12 @@ data:
     #   service fails to respond with a valid rate limit decision within
     #   the timeout defined on the extension service.
     #   failOpen: false
+    # Defines whether to include the X-RateLimit headers X-RateLimit-Limit,
```
stevesloka, line 123: Just one silly nit to align the help text with the variable like the other options.

stevesloka (approved):

examples/contour/01-contour-config.yaml
skriss: fixed!

skriss:
sounds like [using the bool](https://github.com/projectcontour/contour/issues/3431#issuecomment-791494276) works for everyone, but I'll leave this open for one more day in case @youngnick or @xaleeks have further input.

PR #3458, opened 2021-03-08
internal/envoy/v3: Strip port from hostname

sunjayBhatia:
- Strips host port using the strip_any_host_port field on HTTP connection manager
- All filter/router processing will be done internally without the port number attached
- Ensures <HOST>:<PORT_NUM> is accepted and treated the same as the plain <HOST>
- Removes Lua logic to check port number as it is no longer needed, the filter will never see the port on the :authority header
- Removes <HOST>:* wildcard domain match in router, the router will never see the port on the domain to match
- Note that this will strip the port from the Host/:authority header sent to upstreams
- Bumps go-control-plane as latest tagged release is not new enough and does not include the field used
- Will be useful in implementing wildcard host matching, makes internal logic simpler

See: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto.html#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-strip-any-host-port

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.15
 
 require (
 	github.com/ahmetb/gen-crd-api-reference-docs v0.3.0
-	github.com/envoyproxy/go-control-plane v0.9.8
+	github.com/envoyproxy/go-control-plane v0.9.9-0.20210111201334-f1f47757da33
 	github.com/golang/protobuf v1.4.3
 	github.com/google/go-cmp v0.5.2
 	github.com/google/uuid v1.1.2
--- a/internal/envoy/v3/listener.go
+++ b/internal/envoy/v3/listener.go
@@ -389,6 +389,14 @@ func (b *httpConnectionManagerBuilder) Get() *envoy_listener_v3.Filter {
 		UseRemoteAddress: protobuf.Bool(true),
 		NormalizePath:    protobuf.Bool(true),
 
+		// We can ignore any port number supplied in the Host/:authority header
+		// before processing by filters or routing.
+		// Note that the port a listener is bound to will already be selected
+		// and that the port is stripped from the header sent upstream as well.
+		StripPortMode: &http.HttpConnectionManager_StripAnyHostPort{
+			StripAnyHostPort: true,
+		},
+
 		// issue #1487 pass through X-Request-Id if provided.
 		PreserveExternalRequestId: true,
 		MergeSlashes:              true,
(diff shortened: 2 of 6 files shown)


internal/envoy/v3/listener.go
```diff
@@ -389,6 +389,14 @@ func (b *httpConnectionManagerBuilder) Get() *envoy_listener_v3.Filter {
 		UseRemoteAddress: protobuf.Bool(true),
 		NormalizePath:    protobuf.Bool(true),
 
+		// We can ignore any port number supplied in the Host/:authority header
+		// before processing by filters or routing.
+		// Note that the port a listener is bound to will already be selected
+		// and that the port is stripped from the header sent upstream as well.
+		StripPortMode: &http.HttpConnectionManager_StripAnyHostPort{
```
sunjayBhatia, line 396: Note: I did not mention the redirect/multiple listener thing b/c I don't think it should be relevant.
- The hierarchy of the configs is Listener -> Filter chain -> Route
- When we build the redirect at the Route level, we know in contour what the "partner" secure port of the plaintext HTTP port should be and can set the redirect to the port with this field: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-redirectaction-port-redirect
- The redirect will result in a whole new request to a different Listener so stripping the port at the HTTP Connection Manager filter level should be irrelevant

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.15
 
 require (
 	github.com/ahmetb/gen-crd-api-reference-docs v0.3.0
-	github.com/envoyproxy/go-control-plane v0.9.8
+	github.com/envoyproxy/go-control-plane v0.9.9-0.20210111201334-f1f47757da33
```
sunjayBhatia, line 7: This commit is the one that mirrored the protos from the 1.17.0 release

skriss (approved):
Looks ok to me, but someone more familiar with the Lua should definitely look too

youngnick (approved):
LGTM, we'll revisit when we do multiple listeners later.

sunjayBhatia:
Release Note: Ensure Envoy strips port from `:authority`/`Host` header. Simplifies internal routing logic. Note that host in header sent to upstream will have port stripped.

PR #3459, opened 2021-03-09
internal: Add support for exact path match type

stevesloka:
Adds support for exact match type which will be used by IngressV1 & GatewayAPI.


skriss (approved):
LGTM

sunjayBhatia (approved):

PR #3464, opened 2021-03-10
design: Update gateway-api design doc

stevesloka:


youngnick:
Hold #3464 and #3465 until we have more consensus that this is the direction we want to go with.

stevesloka:
I still feel the non-operator mode is important and something that a set of folks is going to want, but my changes to this design don't reflect the last maintainer's meeting. However, this does need updating to reflect the new model which isn't outlined in the current design.

PR #3465, opened 2021-03-10
site: Update gateway-api guide

stevesloka:
Updates the guide to show how to run sample gateway-apis resources without the operator.

### DIFF
--- a/examples/gatewayapi/01-common.yaml
+++ b/examples/gatewayapi/01-common.yaml
@@ -1,0 +1,47 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: projectcontour-roots
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: rootapp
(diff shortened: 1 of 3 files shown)


youngnick:
Hold #3464 and #3465 until we have more consensus that this is the direction we want to go with.

stevesloka:
I still think option is relevant (non-operator). I'll write up a controller in my own time to update status on the Gateway to allow this implementation.

skriss:
I do think Contour should support this mode, and I think we should talk about it in terms of "static" vs. "dynamic" provisioning of the Gateway. You can deploy Contour+Envoy manually to statically provision, or you can use the operator/Gateway controller to dynamically provision. There are different tradeoffs for each.

youngnick:
The static vs dynamic distinction makes sense to me, but I don't understand the point of having a controller that will "reconcile" Gateways, but not update the Contour and Envoy deployment when changes are made. In my book, that is not reconciliation.

PR #3466, opened 2021-03-10
internal/dag: Add namespace selector check for Gateways

### LINKED ISSUE #3401: Implement the RouteBindingSelector

stevesloka:
Filter the HTTPRoutes that match against the namespace field on any struct
looking for if they match the "Same", "All", or "Selector" types.

Fixes #3401

### DIFF
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -42,7 +42,22 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 	kuardService := &v1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "kuard",
-			Namespace: "default",
+			Namespace: "projectcontour",
+		},
+		Spec: v1.ServiceSpec{
+			Ports: []v1.ServicePort{{
+				Name:       "http",
+				Protocol:   "TCP",
+				Port:       8080,
+				TargetPort: intstr.FromInt(8080),
+			}},
+		},
+	}
+
+	kuardServiceCustomNs := &v1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "kuard",
+			Namespace: "custom",
 		},
 		Spec: v1.ServiceSpec{
 			Ports: []v1.ServicePort{{
@@ -57,7 +72,7 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 	blogService := &v1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "blogsvc",
-			Namespace: "default",
+			Namespace: "projectcontour",
 		},
 		Spec: v1.ServiceSpec{
 			Ports: []v1.ServicePort{{
@@ -71,14 +86,17 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 
 	gatewayWithSelector := &gatewayapi_v1alpha1.Gateway{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "gatewayWithSelector",
-			Namespace: "default",
+			Name:      "contour",
+			Namespace: "projectcontour",
 		},
 		Spec: gatewayapi_v1alpha1.GatewaySpec{
 			Listeners: []gatewayapi_v1alpha1.Listener{{
 				Port:     80,
 				Protocol: "HTTP",
 				Routes: gatewayapi_v1alpha1.RouteBindingSelector{
+					Namespaces: gatewayapi_v1alpha1.RouteNamespaces{
+						From: gatewayapi_v1alpha1.RouteSelectSame,
+					},
 					Selector: metav1.LabelSelector{
 						MatchLabels: map[string]string{
 							"app": "contour",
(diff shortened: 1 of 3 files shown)


internal/dag/builder_test.go
```diff
@@ -71,14 +86,17 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
...
+			Name:      "contour",
+			Namespace: "projectcontour",
 		},
 		Spec: gatewayapi_v1alpha1.GatewaySpec{
 			Listeners: []gatewayapi_v1alpha1.Listener{{
 				Port:     80,
 				Protocol: "HTTP",
 				Routes: gatewayapi_v1alpha1.RouteBindingSelector{
+					Namespaces: gatewayapi_v1alpha1.RouteNamespaces{
+						From: "Same",
```
skriss, line 98: I'd prefer to use the constants here and in the other gateway defs: https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha1/gateway_types.go#L448-L455

internal/dag/gatewayapi_processor.go
```diff
@@ -83,6 +91,43 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+		return true, nil
+	case gatewayapi_v1alpha1.RouteSelectSame:
+		return p.source.Gateway.Namespace == route.Namespace, nil
+	case gatewayapi_v1alpha1.RouteSelectSelector:
+		if len(namespaces.Selector.MatchLabels) == 0 || len(namespaces.Selector.MatchExpressions) == 0 {
+			return false, fmt.Errorf("RouteNamespaces selector must be specified when `RouteSelectType=Selector`")
+		}
+
+		// Look up the HTTPRoute's namespace in the list of cached namespaces.
+		if ns := p.source.namespaces[types.NamespacedName{Name: route.Namespace, Namespace: "default"}]; ns != nil {
```
skriss, line 115: The fact that the key here has a `Namespace: "default"` is confusing because `Namespaces` are cluster-scoped. I'm seeing that it's due to `k8s.NamespacedNameOf(...)` automatically setting a namespace of "default" if it's not provided. I'm thinking we should change the `namespaces` map in the DAG cache to be keyed just on a string (the name), rather than a `NamespacedName`, and not use `NamespacedNameOf` when getting/setting. This looks like the pattern that `IngressClass` uses, which is also cluster-scoped.

internal/dag/builder_test.go
stevesloka: 👍

internal/dag/gatewayapi_processor.go
stevesloka: Yup good call! It didn't feel right when I did that, but this makes more sense. I opened https://github.com/projectcontour/contour/pull/3473 to cover.

youngnick: There is some discussion upstream about making both GatewayClass and IngressClass either namespace-scoped, or have namespace-scoped variants.

skriss:
ready for a rebase since #3473 has merged

internal/dag/gatewayapi_processor.go
```diff
@@ -64,11 +66,17 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+
+			selMatches, err := selectorMatches(listener.Routes.Selector, route.Labels)
 			if err != nil {
 				p.Errorf("error validating routes against Listener.Routes.Selector: %s", err)
 			}
-			if matches {
+
+			if selMatches && nsMatches {
 				// Empty Selector matches all routes.
 				validRoutes = append(validRoutes, route)
```
skriss, line 81: nit: `matchingRoutes` would be a more accurate name for this slice than `validRoutes`, I think.

skriss (changes requested):
one more nit, I think LGTM otherwise (pending rebase) - will check once more when it's updated.

internal/dag/gatewayapi_processor.go
stevesloka: 👍

stevesloka:
@skriss all rebased and fixed up, please take a look when you get a second.

skriss (approved):
LGTM

skriss:
now has a conflict to resolve

internal/dag/gatewayapi_processor.go
```diff
@@ -68,25 +69,68 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
-				validRoutes = append(validRoutes, route)
+				matchingRoutes = append(matchingRoutes, route)
 				break
 			}
 		}
 	}
 
 	// Process all the routes that match this Gateway.
-	for _, validRoute := range validRoutes {
+	for _, validRoute := range matchingRoutes {
```
youngnick, line 91: nit: maybe `matchedRoute` instead of `validRoute`?

youngnick (approved):
LGTM with one tiny nit. Don't block this on that though.

Issue #3461, opened 2021-03-09
Ingress: Should default backend be merged with or overridden by Ingress rules


sunjayBhatia:
Ingress resources have the ability to configure set a default backend: https://kubernetes.io/docs/concepts/services-networking/ingress/#default-backend

You can also choose to not configure a host on an Ingress rule and use a prefix path of `/` to match all requests: https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-rules

These two concepts technically conflict and need proper resolution logic

Open questions:
- Does the default backend "win"?
- Does the ingress rule "win"? - *seems like this should be the correct answer to me*
- If multiple Ingresses are present, which configuration wins? - currently the Ingress that is processed last "wins"
- Should these backends/rules be merged?

sunjayBhatia:
See https://github.com/projectcontour/contour/pull/3468

youngnick:
So, I think the general rules should be:
- when there's no conflict, merge.
- when there is a conflict, more specific beats less specific. So a specifically configured item would beat a default backend.
- When two things that are equally specific match, the referred services should be load balanced across (this is the nginx-ingress behavior, and I don't think it's great, but it's either this or drop them both.)

sunjayBhatia:
> So, I think the general rules should be:
> 
> * when there's no conflict, merge.
> * when there is a conflict, more specific beats less specific. So a specifically configured item would beat a default backend.
> * When two things that are equally specific match, the referred services should be load balanced across (this is the nginx-ingress behavior, and I don't think it's great, but it's either this or drop them both.)

Playing around with this a bit, its easy to get the simple parts of some of these rules working in isolation but when you combine them and the various possible other fields that are set on an Ingress and route within that, and then put HTTPProxy resources on top (and then eventually HTTPRoutes) it becomes much more complicated (which is also seemingly why Gateway API has some concept of oldest-wins as well).

e.g. How do we treat two Ingresses like below:

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: foo1
  namespace: bar
  annotations:
    ingress.kubernetes.io/force-ssl-redirect: true
spec:
  rules:
  - http:
      paths:
      - path: "/foo"
         pathType: Prefix
         backend:
         ...
```

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: foo2
  namespace: bar
  annotations:
    projectcontour.io/websocket-routes: "/foo"
[...]

PR #3468, opened 2021-03-10
Ingress: Default backend overridden by Ingress rule and add some additional tests

sunjayBhatia:
- Overlaying Ingresses in reverse order
- A default backend is overriden by a no-host prefix '/' rule
  - default backend is prepended to list of Ingress Rules to make this
    work

Updates https://github.com/projectcontour/contour/issues/3461


skriss (approved):
LGTM, this logic makes sense to me & seems consistent with the spec

youngnick (approved):

PR #3471, opened 2021-03-11
Ingress: Implement Exact and ImplementationSpecific path matches, Prefix match updated to match spec

### LINKED ISSUE #2135: Ingress V1: path matching mode

sunjayBhatia:
Since we already had support for regex path matching without checking
the type of the path match, we now have the following rules to ensure backwards compatibility:

- If you specify a prefix path match type, you get a prefix match
  - prefix matches only match path segments, not a string prefix
  - trailing slashes ignored from prefix match string
- If you specify an exact path match type, you get an exact match
- If you specify an implementation specific path match type you get path matching behavior that Contour already supports
  - If it looks like a regex we give a regex match
  - otherwise a string prefix match like Contour already does
  - ImplementationSpecific match is the default if not set when an Ingress
  resource is created so this should give people the default they are
  already using

Fixes #2135

Signed-off-by: Sunjay Bhatia <sunjayb@vmware.com>


sunjayBhatia:
Still WIP, need to add docs and to make sure Ingress Conformance tests are covered: https://github.com/kubernetes-sigs/ingress-controller-conformance/blob/master/features/path_rules.feature

sunjayBhatia:
See https://github.com/projectcontour/contour/issues/2135#issuecomment-796866113 for a discussion on whether we should let "looks like regex" paths override the path type

sunjayBhatia:
Ingress Conformance tests do not pass with this unfortunately, though the config dump from Envoy looks good

sunjayBhatia:
Created an Ingress and backing services etc. manually since the conformance tests are flaky running locally

Failing tests:
- https://github.com/kubernetes-sigs/ingress-controller-conformance/blob/79bd068cbb31d77c2a060f61eea795f9791a3d48/features/path_rules.feature#L161-L165
  - prefix matches are not "string prefix" matches but label prefix matches, Envoy is doing string prefix matching
- https://github.com/kubernetes-sigs/ingress-controller-conformance/blob/79bd068cbb31d77c2a060f61eea795f9791a3d48/features/path_rules.feature#L174-L179
  - when a prefix and exact match are both present, the prefix is winning when the exact should
- https://github.com/kubernetes-sigs/ingress-controller-conformance/blob/79bd068cbb31d77c2a060f61eea795f9791a3d48/features/path_rules.feature#L181-L186
  - prefix path matches are expected to strip any trailing slash in the match specifier

internal/dag/ingress_processor.go
```diff
@@ -208,7 +215,20 @@ func route(ingress *networking_v1.Ingress, path string, service *Service, client
...
 	}
 
-	r.PathMatchCondition = &PrefixMatchCondition{Prefix: path}
+	switch pathType {
+	case networking_v1.PathTypePrefix:
+		if path == "/" {
+			r.PathMatchCondition = &PrefixMatchCondition{Prefix: path}
+		} else {
+			prefixRegex := regexp.QuoteMeta(strings.TrimRight(path, "/")) + prefixPathEnsureSegmentOnlyMatchRegex
+			r.PathMatchCondition = &RegexMatchCondition{Regex: prefixRegex}
```
sunjayBhatia, line 224: A refactor I am considering: this type of path match would still be `PrefixMatchCondition` but the type would add an field bool option as to whether we do a string prefix match or path segment match only. We would then move this logic to a lower level when we convert to the Envoy config type instead, keeping the DAG simpler. We would do the regex conversion if the segment match is desired.

I didn't make this the default behavior b/c we may not want the segment match only for HTTPProxy etc. but if we do, it makes this simpler still, and the above refactor is even better.

sunjayBhatia:
Still need to add docs, do some more ingress conformance validation, etc. but I have some implementation detail questions so figured I would start a review

internal/dag/builder_test.go
```diff
@@ -858,8 +859,9 @@ func TestDAGInsert(t *testing.T) {
 						Paths: []networking_v1.HTTPIngressPath{{
 							Backend: *backendv1("kuard", intstr.FromString("http")),
 						}, {
-							Path:    "/kuarder",
-							Backend: *backendv1("kuarder", intstr.FromInt(8080)),
+							Path:     "/kuarder",
+							PathType: (*networking_v1.PathType)(pointer.StringPtr("Exact")),
```
sunjayBhatia, line 863: The `PathType` field is a `*PathType` and the `PathType` type is an alias for `string`

This is correct and what VSCode gives me as a completion but it is ugly, maybe worth a test helper function

internal/dag/ingress_processor.go
sunjayBhatia: Also this refactor makes it so the tests for the DAG don't care how the prefix is implemented which is always good

internal/sorter/sorter.go
```diff
@@ -119,6 +119,24 @@ func (s routeSorter) Less(i, j int) bool {
 		case *envoy_route_v3.RouteMatch_Prefix:
 			return true
 		}
+	case *envoy_route_v3.RouteMatch_Path:
```
sunjayBhatia, line 122: Now that prefix matches have become regexes, could this have some weird unintended ordering/precedence consequences? Especially since we add a few characters in the prefix regex, and we sort regex matches on length? Haven't really thought about it yet

stevesloka:
@sunjayBhatia is this ready for review?

sunjayBhatia:
> @sunjayBhatia is this ready for review?

Yeah it could use a review and discussion on some of the threads I started (if you like the proposed refactors etc.)

internal/dag/builder_test.go
```diff
@@ -5160,7 +5262,7 @@ func TestDAGInsert(t *testing.T) {
 					VirtualHosts: virtualhosts(
 						securevirtualhost("b.example.com", sec1,
 							prefixroute("/", service(s1)),
-							prefixroute("/kuarder", service(s2)),
+							exactroute("/kuarder", service(s2)),
```
sunjayBhatia, line 5265: FYI: Made these sorts of changes so we wouldn't have to assert on regex rules all over the place

internal/dag/ingress_processor.go
sunjayBhatia: For any reviewers this is the main thread to look at ^^

Summary: Prefix path matches in Ingress are not "string prefix" matches but path segment matches (prefix of `/foo` matches `/foo/bar` but not `/foobar`)
- Do we want to make this the case for other resources as well?
- The comments above describe a possible refactor from this current implementation

sunjayBhatia:
Also for reviewers, do the following rules look legit:

> * If you provide a path that "looks like" a regex, you will get a regex match, no matter what you set on the path type
> * If you specify a prefix path match type, you get a prefix match
>   
>   * In reality this is a regex match that matches on a prefix that is not a string prefix but a path segment prefix
> * If you specify an exact path match type, you get an exact match
> * If you specify an implementation specific path match type, you get a regex match

stevesloka:
> In reality this is a regex match that matches on a prefix that is not a string prefix but a path segment prefix

So this isn't using the Envoy Prefix match type?  (https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-routematch)

internal/dag/ingress_processor.go
sunjayBhatia: from an in-person chat ImplementationSpecific should be like our current rules:
- move the "looks like regex" handling to this type
- if not a regex, you get a "string prefix" match like we support today

internal/sorter/sorter.go
sunjayBhatia: Potential solution:
- Prefix "regex" matches stay as a Prefix match type with a flag telling that they are not string prefix
- Sort routes before converting to Envoy types

sunjayBhatia:
Ok to get the path match type sorting completely right I'm going to pull it into a new PR since it and the test cases are a non-trivial amount of lines to add to this and it changes some patterns

sunjayBhatia:
> Ok to get the path match type sorting completely right I'm going to pull it into a new PR since it and the test cases are a non-trivial amount of lines to add to this and it changes some patterns

See https://github.com/projectcontour/contour/pull/3504

sunjayBhatia:
Working on merging this with `main` now and cleaning it up

sunjayBhatia:
Ok, need to run Ingress conformance again but should be good to review

sunjayBhatia:
Ingress conformance is all good for the path rules checks as of the latest push

skriss (approved):
this LGTM, nice work @sunjayBhatia! 

Adding the `release-note` label since this is definitely worth a callout.

stevesloka (approved):
Awesome work here @sunjayBhatia!! 🎉 🎉

stevesloka:
@sunjayBhatia ready to merge if you want to clean up the merge commit message a bit. =)

skriss:
@sunjayBhatia dunno if you saw this: https://github.com/projectcontour/contour/runs/2203268540

failing on osx only? 🤷‍♂️

sunjayBhatia:
> @sunjayBhatia dunno if you saw this: https://github.com/projectcontour/contour/runs/2203268540
> 
> failing on osx only? 🤷‍♂️ but i'm seeing it on my recent PR too: https://github.com/projectcontour/contour/pull/3519/checks?check_run_id=2203690235
> 
> It's definitely a flake, it passed on a re-run on my PR. It looks like there might be some maps in there that are being iterated over, that is often a culprit here since order is not guaranteed.

Ah shoot I think I might have forgotten to sort route segment prefix matches before string prefix matches

sunjayBhatia:
@skriss Made a PR here: https://github.com/projectcontour/contour/pull/3521 I think that should fix the issue

internal/envoy/v3/route.go
```diff
@@ -67,11 +71,23 @@ func RouteMatch(route *dag.Route) *envoy_route_v3.RouteMatch {
...
-		return &envoy_route_v3.RouteMatch{
-			PathSpecifier: &envoy_route_v3.RouteMatch_Prefix{
-				Prefix: c.Prefix,
-			},
-			Headers: headerMatcher(route.HeaderMatchConditions),
+		switch c.PrefixMatchType {
+		case dag.PrefixMatchSegment:
+			return &envoy_route_v3.RouteMatch{
+				PathSpecifier: &envoy_route_v3.RouteMatch_SafeRegex{
+					SafeRegex: SafeRegexMatch(regexp.QuoteMeta(c.Prefix) + prefixPathMatchSegmentRegex),
```
shashankram, line 78: Per the `Note` in https://kubernetes.io/docs/concepts/services-networking/ingress/#path-types, a pathPrefix `/foo/bar` must match `/foo/bar/baz` but not `/foo/barbaz`, where the last path segment in the request is a subtring of the specified pathPrefix.

The existing code will result in a regex `/foo/bar((\/).*)?` which will match `/foo/barbaz` when it shouldn't.

sunjayBhatia: The regex could be simplified a bit to remove the inner capture group and discard the captures since they are not actually used, but I believe it does match the path segment prefix correctly

sunjayBhatia: Also this regex is validated against the ingress conformance tests

shashankram: In the example `^/foo/bar((\/).*)?$` you shared, who adds the end of line match `$`? Is this implicit in Envoy?

sunjayBhatia: Yes, this is implicit in Envoy, see the Envoy docs link above

```
If specified, the route is a regular expression rule meaning that the regex must match the :path header once the query string is removed. The entire path (without the query string) must match the regex. The rule will not match if only a subsequence of the :path header matches the regex.
```

shashankram: The envoy doc is a bit cryptic in this regard. I would assume one must set the EOL matcher `$`. In my experience, setting the `...$` matcher explicitly does the job fine.

Thanks for clarifying!

Issue #3438, opened 2021-03-03
Support TLS on Gateway


stevesloka:
A Gateway can define a `GatewayTLSConfig` which allows the TLS bits to defined on the Gateway and not required on the route. 

//ref: https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.GatewayTLSConfig

PR #3472, opened 2021-03-11
internal: Add support for TLS on Gateway

stevesloka:
Gateways can now supply certificateRefs which allow the Gateway & it's selected HTTPRoutes to
terminate TLS at the Envoy listener.

Updates #3438

### DIFF
--- a/_integration/testsuite/gatewayapi/004-tls-gateway.yaml
+++ b/_integration/testsuite/gatewayapi/004-tls-gateway.yaml
@@ -1,0 +1,237 @@
+# Copyright Project Contour Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.  You may obtain
+# a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+import data.contour.resources
+
+# Ensure that cert-manager is installed.
+# Version check the certificates resource.
+
+Group := "cert-manager.io"
+Version := "v1"
+
+have_certmanager_version {
+  v := resources.versions["certificates"]
+  v[_].Group == Group
+  v[_].Version == Version
+}
+
+skip[msg] {
+  not resources.is_supported("certificates")
+  msg := "cert-manager is not installed"
+}
+
+skip[msg] {
+  not have_certmanager_version
+
+  avail := resources.versions["certificates"]
+
+  msg := concat("\n", [
+  sprintf("cert-manager version %s/%s is not installed", [Group, Version]),
+  "available versions:",
+  yaml.marshal(avail)
+  ])
+}
+
+---
+
+apiVersion: cert-manager.io/v1
+kind: ClusterIssuer
+metadata:
+  name: selfsigned
+spec:
+  selfSigned: {}
+
+---
+
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: tlscert
+  namespace: projectcontour
+spec:
+  dnsNames:
+  - secure.projectcontour.io
+  secretName: tlscert
+  issuerRef:
+    name: selfsigned
+    kind: ClusterIssuer
+
+---
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: ingress-conformance-echo
+$apply:
+  fixture:
+    as: echo-slash-default
+
+---
+
+apiVersion: v1
+kind: Service
+metadata:
+  name: ingress-conformance-echo
+$apply:
+  fixture:
+    as: echo-slash-default
+
+---
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: ingress-conformance-echo
+$apply:
+  fixture:
+    as: echo-slash-secure
+
+---
+
+apiVersion: v1
+kind: Service
+metadata:
+  name: ingress-conformance-echo
+$apply:
+  fixture:
+    as: echo-slash-secure
+
+---
+
+apiVersion: networking.x-k8s.io/v1alpha1
+kind: GatewayClass
+metadata:
+  name: contour-class
+spec:
+  controller: projectcontour.io/ingress-controller
+
+---
+
+apiVersion: networking.x-k8s.io/v1alpha1
+kind: Gateway
+metadata:
+  name: contour
+  namespace: projectcontour
+spec:
+  gatewayClassName: contour-class
+  listeners:
+    - protocol: HTTP
+      port: 80
+      routes:
+        kind: HTTPRoute
+        namespaces:
+          from: "All"
+        selector:
+          matchLabels:
+            type: insecure
+    - protocol: HTTPS
+      port: 443
+      tls:
+        certificateRef:
+          kind: Secret
+          group: core
+          name: tlscert
+      routes:
+        kind: HTTPRoute
+        namespaces:
+          from: "All"
+        selector:
+          matchLabels:
+            type: secure
+---
+
+apiVersion: networking.x-k8s.io/v1alpha1
+kind: HTTPRoute
+metadata:
+  name: http-filter-1
+  labels:
+    type: insecure
+spec:
+  hostnames:
+    - secure.projectcontour.io
+  rules:
+    - matches:
+      - path:
+          type: Prefix
+          value: /
+      forwardTo:
+      - serviceName: echo-slash-default
+        port: 80
+
+---
+
+apiVersion: networking.x-k8s.io/v1alpha1
+kind: HTTPRoute
+metadata:
+  name: http-filter-2
+  labels:
+    type: secure
+spec:
+  hostnames:
+    - secure.projectcontour.io
+  rules:
+    - matches:
+      - path:
+          type: Prefix
+          value: /
+      forwardTo:
+      - serviceName: echo-slash-secure
+        port: 80
+
+---
+
+import data.contour.http.client
+import data.contour.http.client.url
+import data.contour.http.expect
+
+# Ensure http (insecure) request returns 200 status code.
+Response := client.Get({
+  "url": url.http("/"),
+    "headers": {
+      "Host": "secure.projectcontour.io",
+      "User-Agent": client.ua("insecure"),
+  },
+})
+
+check_for_status_code [msg] {
+  msg := expect.response_status_is(Response, 200)
+}
+
+check_for_service_routing [msg] {
+  msg := expect.response_service_is(Response, "echo-slash-default")
+}
+
+---
+
+import data.contour.http.client
+import data.contour.http.client.url
+import data.contour.http.expect
+
+# Ensure https (secure) request returns 200 status code.
+Response := client.Get({
+  "url": url.https("/"),
+  "headers": {
+    "Host": "secure.projectcontour.io",
+    "User-Agent": client.ua("secure"),
+  },
+  "tls_insecure_skip_verify": true,
+})
+
+check_for_status_code [msg] {
+  msg := expect.response_status_is(Response, 200)
+}
+
+check_for_service_routing [msg] {
+  msg := expect.response_service_is(Response, "echo-slash-secure")
+}
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -133,6 +133,183 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 		},
 	}
 
+	sec1 := &v1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "tlscert",
+			Namespace: "projectcontour",
+		},
+		Type: v1.SecretTypeTLS,
+		Data: secretdata(fixture.CERTIFICATE, fixture.RSA_PRIVATE_KEY),
+	}
+
+	gatewayWithOnlyTLS := &gatewayapi_v1alpha1.Gateway{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "contour",
+			Namespace: "projectcontour",
+		},
+		Spec: gatewayapi_v1alpha1.GatewaySpec{
+			Listeners: []gatewayapi_v1alpha1.Listener{{
+				Port:     443,
+				Protocol: gatewayapi_v1alpha1.HTTPSProtocolType,
+				TLS: &gatewayapi_v1alpha1.GatewayTLSConfig{
+					CertificateRef: &gatewayapi_v1alpha1.LocalObjectReference{
+						Group: "core",
+						Kind:  "Secret",
+						Name:  sec1.Name,
+					},
+				},
+				Routes: gatewayapi_v1alpha1.RouteBindingSelector{
+					Kind: KindHTTPRoute,
+					Namespaces: gatewayapi_v1alpha1.RouteNamespaces{
+						From: gatewayapi_v1alpha1.RouteSelectAll,
+					},
+				},
+			}},
+		},
+	}
+
+	gatewayWithTLSandHTTP := &gatewayapi_v1alpha1.Gateway{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "contour",
+			Namespace: "projectcontour",
+		},
+		Spec: gatewayapi_v1alpha1.GatewaySpec{
+			Listeners: []gatewayapi_v1alpha1.Listener{{
+				Port:     80,
+				Protocol: gatewayapi_v1alpha1.HTTPProtocolType,
+				Routes: gatewayapi_v1alpha1.RouteBindingSelector{
+					Kind: KindHTTPRoute,
+					Namespaces: gatewayapi_v1alpha1.RouteNamespaces{
+						From: gatewayapi_v1alpha1.RouteSelectAll,
+					},
+				},
+			}, {
+				Port:     443,
+				Protocol: gatewayapi_v1alpha1.HTTPSProtocolType,
+				TLS: &gatewayapi_v1alpha1.GatewayTLSConfig{
+					CertificateRef: &gatewayapi_v1alpha1.LocalObjectReference{
+						Group: "core",
+						Kind:  "Secret",
+						Name:  sec1.Name,
+					},
+				},
+				Routes: gatewayapi_v1alpha1.RouteBindingSelector{
+					Kind: KindHTTPRoute,
+					Namespaces: gatewayapi_v1alpha1.RouteNamespaces{
+						From: gatewayapi_v1alpha1.RouteSelectAll,
+					},
+				},
+			}},
+		},
+	}
+
+	gatewaywithtlsDifferentselectors := &gatewayapi_v1alpha1.Gateway{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "contour",
+			Namespace: "projectcontour",
+		},
+		Spec: gatewayapi_v1alpha1.GatewaySpec{
+			Listeners: []gatewayapi_v1alpha1.Listener{{
+				Port:     80,
+				Protocol: gatewayapi_v1alpha1.HTTPProtocolType,
+				Routes: gatewayapi_v1alpha1.RouteBindingSelector{
+					Kind: KindHTTPRoute,
+					Namespaces: gatewayapi_v1alpha1.RouteNamespaces{
+						From: gatewayapi_v1alpha1.RouteSelectAll,
+					},
+					Selector: metav1.LabelSelector{
+						MatchExpressions: []metav1.LabelSelectorRequirement{{
+							Key:      "protocol",
+							Operator: "In",
+							Values:   []string{"http"},
+						}},
+					},
+				},
+			}, {
+				Port:     443,
+				Protocol: gatewayapi_v1alpha1.HTTPSProtocolType,
+				TLS: &gatewayapi_v1alpha1.GatewayTLSConfig{
+					CertificateRef: &gatewayapi_v1alpha1.LocalObjectReference{
+						Group: "core",
+						Kind:  "Secret",
+						Name:  sec1.Name,
+					},
+				},
+				Routes: gatewayapi_v1alpha1.RouteBindingSelector{
+					Kind: KindHTTPRoute,
+					Namespaces: gatewayapi_v1alpha1.RouteNamespaces{
+						From: gatewayapi_v1alpha1.RouteSelectAll,
+					},
+					Selector: metav1.LabelSelector{
+						MatchExpressions: []metav1.LabelSelectorRequirement{{
+							Key:      "protocol",
+							Operator: "In",
+							Values:   []string{"https"},
+						}},
+					},
+				},
+			}},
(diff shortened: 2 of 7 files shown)


stevesloka:
Hmm these integration tests have been a pain for some reason. I'll dig into those better to see why the one I added keeps failing.

stevesloka:
Ok, this should all work now. A follow-up will need to refactor the tls logic away. It's modeled how Contour handles TLS today, but not every listener could select the same HTTPRoutes, so we may need to process those routes differently.

stevesloka:
> A follow-up will need to refactor the tls logic away. It's modeled how Contour handles TLS today, but not every listener could select the same HTTPRoutes, so we may need to process those routes differently.

Should just do this now and have it be done right from the beginning which is why I put it back into draft status.

stevesloka:
Just realized I need to update with status info.

stevesloka:
Ok all fixed up I think we're good to review. There isn't any status updates since right now only processing for the Gateway certs. // cc @danehans

internal/dag/gatewayapi_processor.go
```diff
@@ -61,6 +63,34 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
 
 	for _, listener := range p.source.gateway.Spec.Listeners {
 
+		var listenerSecret *Secret
+		var err error
+
+		// Check for TLS on the Gateway.
+		if tls := listener.TLS; tls != nil {
+			// Validate the CertificateRef is configured.
```
danehans, line 71: you should also check for the listener protocol. The protocol != "HTTP". Check out similar validation that I planned for the operator:

https://github.com/projectcontour/contour-operator/pull/237/files#diff-feefa601b9dc749f80c19b5769691942730947a8a8226c6ff15767916e58980cR144-R171

Also, [the validation pkg](https://github.com/kubernetes-sigs/gateway-api/pull/597) has support upstream... just working through the details. I think it makes sense for Contour to have a validation pkg that includes these checks. When the upstream validation is ready, the contour validation pkg can dump all the local validation code except any implementation-specific validation. Thoughts?

internal/dag/gatewayapi_processor.go
```diff
@@ -61,6 +63,34 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+			if tls.CertificateRef == nil {
+				p.Error(contour_api_v1.ConditionTypeTLSError, "SecretNotConfigured",
+					"Spec.VirtualHost.TLS Secret is not configured")
```
danehans, line 74: [field.Path](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation/field#Path) and [field.ErrorList](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation/field#ErrorList) is what I'm using upstream and provides helpful error context. [networking api validation](https://github.com/kubernetes/kubernetes/blob/release-1.19/pkg/apis/networking/validation/validation.go) is a good reference.

internal/dag/gatewayapi_processor.go
```diff
@@ -61,6 +63,34 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+
+			// Validate a v1.Secret is referenced which can be kind: secret & group: core.
+			// ref: https://github.com/kubernetes-sigs/gateway-api/pull/562
+			if strings.ToLower(tls.CertificateRef.Kind) != "secret" && strings.ToLower(tls.CertificateRef.Group) != "v1" {
+				p.Error(contour_api_v1.ConditionTypeTLSError, "InvalidReference",
+					"Spec.VirtualHost.TLS Secret must be type v1.Secret")
+				continue
+			}
+
+			listenerSecret, err = p.source.LookupSecret(types.NamespacedName{Name: tls.CertificateRef.Name, Namespace: p.source.gateway.Namespace}, validSecret)
```
danehans, line 86: Might not be a bad idea to expand the cert validation. You should be able to copy/paste [this code](https://github.com/projectcontour/contour-operator/pull/237/files#diff-feefa601b9dc749f80c19b5769691942730947a8a8226c6ff15767916e58980cR242-R271) I planned for the operator.

internal/dag/gatewayapi_processor.go
```diff
@@ -106,11 +136,44 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
 		// Process all the routes that match this Gateway.
 		for _, matchingRoute := range matchingRoutes {
-			p.computeHTTPRoute(matchingRoute)
+			hosts := p.computeHosts(matchingRoute)
+			p.computeHTTPRoute(matchingRoute, hosts, listenerSecret)
 		}
 	}
 }
 
+func (p *GatewayAPIProcessor) computeHosts(route *gatewayapi_v1alpha1.HTTPRoute) []string {
```
danehans, line 145: I expect https://github.com/kubernetes-sigs/gateway-api/pull/597 to merge upstream. I suggest aligning as much as possible to ease the transition to upstream.

internal/featuretests/v3/featuretests.go
```diff
@@ -110,6 +112,10 @@ func setup(t *testing.T, opts ...interface{}) (cache.ResourceEventHandler, *Cont
 		Builder: dag.Builder{
 			Source: dag.KubernetesCache{
 				FieldLogger: log,
+				ConfiguredGateway: types.NamespacedName{
+					Name:      "contour",
+					Namespace: "projectcontour",
```
danehans, line 117: I thought the ns/name comes from the contour cfgmap?

internal/dag/builder_test.go
```diff
@@ -132,6 +132,190 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
...
+				},
+			}},
+		},
+	}
+
+	gatewayWithTLS := &gatewayapi_v1alpha1.Gateway{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "contour",
+			Namespace: "projectcontour",
+		},
```
danehans, line 174: Do we have any tests that cover a gateway using the non-default namespace?

internal/dag/gatewayapi_processor.go
stevesloka: Yup sounds good! I added support to check for HTTPS protocol since that's what we're supporting in this PR and opened another issue to track the other bits: https://github.com/projectcontour/contour/issues/3540

internal/dag/gatewayapi_processor.go
stevesloka: Right now in Contour, certs are validated before they are put into the cache (https://github.com/projectcontour/contour/blob/main/internal/dag/secret.go#L34:6). This logic here in the processor is just checking to see if we've cached a matching secret to use or not. 

If we want we could expand the validation logic in a separate PR in the cache bits.

internal/dag/gatewayapi_processor.go
stevesloka: Yeah for sure, when that merges we can update and remove this code if it's not required any longer.

internal/dag/builder_test.go
stevesloka: There aren't really any default namespaces any longer. At the moment, you're required to define which gateway name/namespace to use. We hardcode that here: https://github.com/projectcontour/contour/blob/main/internal/dag/builder_test.go#L1081-L1083

We do test resources in different namespaces, but right now use `contour/projectcontour` for the configured one.

internal/dag/gatewayapi_processor.go
```diff
@@ -61,6 +63,41 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+					"Spec.VirtualHost.TLS Secret must be type v1.Secret")
+				continue
+			}
+
+			listenerSecret, err = p.source.LookupSecret(types.NamespacedName{Name: tls.CertificateRef.Name, Namespace: p.source.gateway.Namespace}, validSecret)
+			if err != nil {
+				p.Errorf(contour_api_v1.ConditionTypeTLSError, "SecretNotValid",
+					"Spec.VirtualHost.TLS Secret %q is invalid: %s", tls.CertificateRef.Name, err)
+				continue
+			}
```
skriss, line 98: might be helpful to move this to a helper method for validating TLS/getting the TLS secret

internal/dag/gatewayapi_processor.go
```diff
@@ -61,6 +63,41 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+			// Validate the correct protocol is specified.
+			if listener.Protocol != gatewayapi_v1alpha1.HTTPSProtocolType {
+				p.Error(contour_api_v1.ConditionTypeTLSError, "InvalidProtocol",
+					"Spec.VirtualHost.Protocol %q is not valid.")
+				continue
+			}
+
+			// Validate a v1.Secret is referenced which can be kind: secret & group: core.
+			// ref: https://github.com/kubernetes-sigs/gateway-api/pull/562
+			if strings.ToLower(tls.CertificateRef.Kind) != "secret" && strings.ToLower(tls.CertificateRef.Group) != "v1" {
```
skriss, line 87: Should be an `OR`, otherwise e.g. `v1/configmap` would be allowed.
```suggestion
			if strings.ToLower(tls.CertificateRef.Kind) != "secret" || strings.ToLower(tls.CertificateRef.Group) != "v1" {
```

Might actually be easiest to just define a positive-polarity `isSecretRef` func and then check the negation of that:
```
func isSecretRef(certificateRef) bool {
        return certificateRef.Group == "v1" && certificateRef.Kind == "secret"
}
...
if !isSecretRef(tls.CertificateRef) {
        ....
}
```

internal/dag/gatewayapi_processor.go
```diff
@@ -106,9 +143,42 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+		}
+	}
+}
+
+func (p *GatewayAPIProcessor) computeHosts(route *gatewayapi_v1alpha1.HTTPRoute) []string {
+	// Determine the hosts on the route, if no hosts
+	// are defined, then set to "*".
+	var hosts []string
+	if len(route.Spec.Hostnames) == 0 {
+		hosts = append(hosts, "*")
```
skriss, line 157: can return early here and then drop the `else`

internal/dag/gatewayapi_processor.go
```diff
@@ -106,9 +143,42 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+	} else {
+		for _, host := range route.Spec.Hostnames {
+
+			hostname := string(host)
+			if isIP := net.ParseIP(hostname) != nil; isIP {
+				p.Errorf("hostname %q must be a DNS name, not an IP address", hostname)
```
skriss, line 163: I assume these should eventually become conditions, any reason not to do that now?

skriss (commented):
couple of small comments, will look at tests next.

internal/dag/gatewayapi_processor.go
stevesloka: > Should be an OR, otherwise e.g. v1/configmap would be allowed.

Wouldn't the `OR` allow the configmap to work? I'm confused on that bit.

internal/dag/gatewayapi_processor.go
stevesloka: I'll update, the conditions merged before this so just got lost.

internal/dag/gatewayapi_processor.go
skriss: The current code says that to be an invalid ref, it has to both (a) not be a `Secret`; **and** (b) not be in `v1`. Since `v1/ConfigMap` doesn't meet condition (b), it wouldn't be considered invalid. But really, if it meets either (a) **or** (b) it should be invalid.

I think just doing the `isSecretRef` helper, and then checking the negation of it, is easiest :)

_integration/testsuite/gatewayapi/002-tls-gateway.yaml
```diff
@@ -0,0 +1,195 @@
...
+          port: 80
+
+---
+
+import data.contour.http.client
+  import data.contour.http.client.url
+  import data.contour.http.expect
+
+  cases := [
+    [ "/", "echo-slash-default" ],
```
sunjayBhatia, line 153: Seems like this is just one case so far so the overhead of the test written for multiple cases is not needed?

Are there plans for adding more test cases to this test?

internal/dag/builder_test.go
```diff
@@ -132,6 +132,190 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
...
+					},
+				},
+				Routes: gatewayapi_v1alpha1.RouteBindingSelector{
+					Kind: KindHTTPRoute,
+					Namespaces: gatewayapi_v1alpha1.RouteNamespaces{
+						From: gatewayapi_v1alpha1.RouteSelectAll,
+					},
+					Selector: metav1.LabelSelector{
+						MatchExpressions: []metav1.LabelSelectorRequirement{{
+							Key:      "protocol",
```
sunjayBhatia, line 202: Given the `gatewaywithtlsDifferentselectors` gateway, did we want to change this to match the HTTP listener selector or have another test case that makes sure when they have the same selector things work?

internal/dag/builder_test.go
```diff
@@ -1113,6 +1246,105 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
...
+		},
+		"insert basic single route, single hostname, gateway with missing TLS certificate": {
+			gateway: gatewayWithOnlyTLS,
+			objs: []interface{}{
+				kuardService,
+				genericHTTPRoute,
+			},
+			want: listeners(),
+		},
+		"insert basic single route, single hostname, gateway with invalid TLS certificate": {
```
sunjayBhatia, line 1280: this looks identical to the previous test case, seems like it should have an actual invalid secret being inserted?

internal/dag/gatewayapi_processor.go
```diff
@@ -106,11 +118,82 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
-			p.computeHTTPRoute(matchingRoute)
+			p.computeHTTPRoute(matchingRoute, listenerSecret)
 		}
 	}
 }
 
+func (p *GatewayAPIProcessor) validGatewayTLS(listener gatewayapi_v1alpha1.Listener) *Secret {
+
+	// Validate the CertificateRef is configured.
+	if listener.TLS.CertificateRef == nil {
```
sunjayBhatia, line 129: Worth adding status tests for the various status conditions that might be set for errors?

internal/dag/gatewayapi_processor.go
```diff
@@ -106,11 +118,82 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+		return nil
+	}
+	return listenerSecret
+}
+
+func isSecretRef(certificateRef *gatewayapi_v1alpha1.LocalObjectReference) bool {
+	return strings.ToLower(certificateRef.Kind) == "secret" && strings.ToLower(certificateRef.Group) == "core"
+}
+
+func (p *GatewayAPIProcessor) computeHosts(route *gatewayapi_v1alpha1.HTTPRoute) ([]string, []error) {
```
sunjayBhatia, line 163: Do we have tests for all the codepaths in here? Esp. the wildcards etc.

internal/dag/gatewayapi_processor.go
```diff
@@ -165,42 +248,24 @@ func selectorMatches(selector metav1.LabelSelector, objLabels map[string]string)
...
+func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRoute, listenerSecret *Secret) {
 	routeAccessor, commit := p.dag.StatusCache.HTTPRouteAccessor(route)
 	defer commit()
 
-	// Validate TLS Configuration
-	if route.Spec.TLS != nil {
-		routeAccessor.AddCondition(status.ConditionNotImplemented, metav1.ConditionTrue, status.ReasonNotImplemented, "HTTPRoute.Spec.TLS: Not yet implemented.")
+	hosts, errs := p.computeHosts(route)
+	for _, err := range errs {
+		routeAccessor.AddCondition(status.ConditionResolvedRefs, metav1.ConditionFalse, status.ReasonDegraded, err.Error())
```
sunjayBhatia, line 257: Worth adding status tests for this and the other status being set in this helper?

internal/featuretests/v3/gatewayapi_test.go
```diff
@@ -0,0 +1,160 @@
...
+						Action: routeCluster("default/svc2/80/da39a3ee5e"),
+					}, &envoy_route_v3.Route{
+						Match:  routePrefix("/"),
+						Action: routeCluster("default/svc1/80/da39a3ee5e"),
+					},
+				),
+			),
+		),
+		TypeUrl: routeType,
+	})
```
sunjayBhatia, line 154: worth checking the secrets returned include the configured secret?

internal/dag/gatewayapi_processor.go
stevesloka: These are Gateway statuses which were going to be handled by the Operator. Since then we've discussed this happening inside Contour, but that is happening in a separate branch.

sunjayBhatia: ah right, forgot about that

internal/dag/gatewayapi_processor.go
stevesloka: added tests

internal/dag/gatewayapi_processor.go
stevesloka: yup good call, added some more.

internal/featuretests/v3/gatewayapi_test.go
stevesloka: 👍

internal/featuretests/v3/featuretests.go
stevesloka: We don't have any good tests that cover configmap --> code (like a feature test). I know @skriss is working on some of those integration test rework which I think will allow us to better validate the settings from the configmap onwards.

internal/dag/gatewayapi_processor_test.go
```diff
@@ -0,0 +1,185 @@
...
+	gatewayapi_v1alpha1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
+)
+
+func TestComputeHosts(t *testing.T) {
+	tests := map[string]struct {
+		route     *gatewayapi_v1alpha1.HTTPRoute
+		want      []string
+		wantError []error
+	}{
+		"single host": {
```
sunjayBhatia, line 33: Maybe add a test what has a hostname that is a single DNS label? Just to make sure `validation.IsDNS1123Subdomain` accepts that

stevesloka: 👍

internal/dag/gatewayapi_processor.go
```diff
@@ -113,11 +125,81 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+			p.computeHTTPRoute(matchingRoute, listenerSecret)
 		}
 	}
 }
 
+func (p *GatewayAPIProcessor) validGatewayTLS(listener gatewayapi_v1alpha1.Listener) *Secret {
+
+	// Validate the CertificateRef is configured.
+	if listener.TLS.CertificateRef == nil {
+		p.Error(contour_api_v1.ConditionTypeTLSError, "SecretNotConfigured",
```
skriss, line 137: This and the other log lines in this function confused me for a sec because they _look_ like `Conditions`, but they're actually logging -- is this format for the logging desired, or would it make sense to switch to more standard `.Errorf(...)` formatting?

skriss (commented):
Just one nit, otherwise I think this looks pretty good; there are a few comments still outstanding I believe on the tests.

internal/dag/builder_test.go
stevesloka: 👍

internal/dag/builder_test.go
stevesloka: I took out the extra selector from the gateway `gatewayWithTLSandHTTP` since we're not testing those in that scenario.

_integration/testsuite/gatewayapi/002-tls-gateway.yaml
stevesloka: Nope right now no plans, I can clean up to remove.

internal/dag/gatewayapi_processor.go
stevesloka: I cleaned up, I think what happened was I was writing status to the Gateway, then turned out that Contour wasn't going to do that, so it got left like this.

stevesloka:
@skriss @sunjayBhatia I think this should all be fixed up. =)

_integration/testsuite/gatewayapi/004-tls-gateway.yaml
```diff
@@ -0,0 +1,217 @@
...
+  msg := expect.response_status_is(Response, 200)
+}
+
+check_for_service_routing [msg] {
+  msg := expect.response_service_is(Response, "echo-slash-default")
+}
+#
+#---
+#
+#import data.contour.http.client
```
sunjayBhatia, line 170: leaving this in for more test cases later?

sunjayBhatia (approved):

_integration/testsuite/gatewayapi/004-tls-gateway.yaml
stevesloka: 🤭

skriss:
got a merge conflict to resolve

skriss:
Pretty sure this looked good to me, but I'll take one more pass in the morning or @youngnick can feel free to merge if it looks good to him.

youngnick (approved):
A small note about making sure we come back and fix the Conditions handling for multiple errors (we shouldn't end up with multiple Conditions of the same Type), but I think that's a convention rather than enforced by anything. So we have time to fix it.

Aside from that, looks great to me.

youngnick:
I'd rather leave this to @skriss to merge.

internal/dag/gatewayapi_processor.go
```diff
@@ -172,42 +248,24 @@ func selectorMatches(selector metav1.LabelSelector, objLabels map[string]string)
...
+func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRoute, listenerSecret *Secret) {
 	routeAccessor, commit := p.dag.StatusCache.HTTPRouteAccessor(route)
 	defer commit()
 
-	// Validate TLS Configuration
-	if route.Spec.TLS != nil {
-		routeAccessor.AddCondition(status.ConditionNotImplemented, metav1.ConditionTrue, status.ReasonNotImplemented, "HTTPRoute.Spec.TLS: Not yet implemented.")
+	hosts, errs := p.computeHosts(route)
+	for _, err := range errs {
+		routeAccessor.AddCondition(status.ConditionResolvedRefs, metav1.ConditionFalse, status.ReasonDegraded, err.Error())
```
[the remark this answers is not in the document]
stevesloka: https://github.com/projectcontour/contour/issues/3534

skriss (changes requested):
sorry, just noticed one minor test issue

skriss (approved):
LGTM

PR #3473, opened 2021-03-11
internal/dag: Change key for namespaces cache to be string

stevesloka:
Namespaces don't have a "namespace", so keying off of them is incorrect, so this
changes the namespace key to be scoped to only the name.

Updates #3466


youngnick (approved):
LGTM, this makes sense regardless of what else happens.

sunjayBhatia (approved):

Issue #3353, opened 2021-02-11
Validation for hostnames


stevesloka:
Need to validate each hostname The first phase of hostname validation can be:

- Ensure the hostname is not an IP address. xref: https://golang.org/pkg/net/#ParseIP
- Ensure the hostname does not contain a port. xref: https://golang.org/pkg/net/#SplitHostPort

As a second phase, the hostname spec states "Hostname is the fully qualified domain name of a network host, as defined by RFC 3986." We should be able to use https://golang.org/pkg/net/?m=all#isDomainName (verify that it supports wildcard prefixes or augment to support wildcards).

> Hostnames defines a set of hostname that should match against the HTTP Host header to select a HTTPRoute to process the request. Hostname is the fully qualified domain name of a network host, as defined by RFC 3986. Note the following deviations from the “host” part of the URI as defined in the RFC:

    IPs are not allowed.
    The : delimiter is not respected because ports are not allowed.

Incoming requests are matched against the hostnames before the HTTPRoute rules. If no hostname is specified, traffic is routed based on the HTTPRouteRules.

Hostname can be “precise” which is a domain name without the terminating dot of a network host (e.g. “foo.example.com”) or “wildcard”, which is a domain name prefixed with a single wildcard label (e.g. “.example.com”). The wildcard character ‘’ must appear by itself as the first DNS label and matches only a single label. You cannot have a wildcard label by itself (e.g. Host == “”). Requests will be matched against the Host field in the following order: 1. If Host is precise, the request matches this rule if the http host header is equal to Host. 2. If Host is a wildcard, then the request matches this rule if the http host header is to equal to the suffix (removing the first label) of the wildcard rule.

// ref: https://kubernetes-sigs.github.io/service-apis/spec/#networking.x-k8s.io/v1alpha1.HTTPRoute
// ref: https://github.com/projectcontour/contour/pull/3278#discussion_r574180711

sunjayBhatia:
Do we have to do this ourselves? Does the `HTTPRoute` CRD already have some admission/validation logic like `Ingress` does for hostnames?

danehans:
> Do we have to do this ourselves? Does the HTTPRoute CRD already have some admission/validation logic like Ingress does for hostnames?

@sunjayBhatia the API validation of this field is pretty basic, xref: https://github.com/kubernetes-sigs/gateway-api/blob/v0.2.0/apis/v1alpha1/shared_types.go#L218-L219. Implementations will need to perform additional validation to meet the spec.

PR #3477, opened 2021-03-13
Add validation for hostnames in HTTPRoute

nak3:
This patch adds validaiton for HTTPRoute.

The requirement is same with
`ingresses.networking.k8s.io.spec.rules.host` (please refer to it by
`$ kubectl explain ingresses.networking.k8s.io.spec.rules.host`) so
this patch just adds the same code from [here](https://github.com/kubernetes/kubernetes/blob/3b5d809eedcc60e1e8fe63dace1f453d6a27b949/pkg/apis/networking/validation/validation.go#L346-L360).

Fix https://github.com/projectcontour/contour/issues/3353

/cc @stevesloka @sunjayBhatia @danehans

### DIFF
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -14,11 +14,15 @@
 package dag
 
 import (
+	"net"
+	"strings"
+
 	"github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/validation"
 	gatewayapi_v1alpha1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
 )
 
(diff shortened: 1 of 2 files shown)


stevesloka (approved):

sunjayBhatia (commented):
Should we leave a TODO to set some status on errors?

If we are going with the whatever plans to do Gateway validation in the operator will we need to duplicate this logic there?

Also could maybe add a test or two for some more edge cases, like wildcard domains with the `*` in different places, a single DNS label for the hostname (like we might have for a k8s Service), etc.

danehans:
Since the same validation needs to be performed for Gateway, we should do either:
1. Create a contour/pkg/validation pkg that includes the logic in this PR.
2. If the operator will responsible for reconciling a Gateway (which includes the same hostname field), this logic should be in repo that contains shared packages.

xref the same validation implemented by the operator for Gateway hostnames: https://github.com/projectcontour/contour-operator/pull/226

danehans:
/lgtm

I'll submit a follow-on PR to move this logic into a validation pkg.

danehans (approved):

internal/dag/gatewayapi_processor.go
```diff
@@ -114,6 +118,22 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
...
+				continue
+			}
+			if strings.Contains(hostname, "*") {
+				if errs := validation.IsWildcardDNS1123Subdomain(hostname); errs != nil {
+					p.Errorf("invalid hostname %q: %v", hostname, errs)
+					continue
+				}
+			} else {
+				if errs := validation.IsDNS1123Subdomain(hostname); errs != nil {
+					p.Errorf("invalid listener hostname %q", hostname, errs)
```
jpeach, line 133:
```suggestion
p.Errorf("invalid listener hostname %q: %v", hostname, errs)
```

PR #3478, opened 2021-03-14
Implement "Exact" HTTPPathMatch

nak3:
This patch adds "Exact" HTTPPathMatch.

/cc @stevesloka @youngnick

### DIFF
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -205,15 +205,17 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
 
 	for _, rule := range route.Spec.Rules {
 
-		var pathPrefixes []string
+		var pathMatchConditions []MatchCondition
 		var services []*Service
 
 		for _, match := range rule.Matches {
 			switch match.Path.Type {
 			case gatewayapi_v1alpha1.PathMatchPrefix:
-				pathPrefixes = append(pathPrefixes, stringOrDefault(match.Path.Value, "/"))
+				pathMatchConditions = append(pathMatchConditions, &PrefixMatchCondition{Prefix: stringOrDefault(match.Path.Value, "/")})
+			case gatewayapi_v1alpha1.PathMatchExact:
+				pathMatchConditions = append(pathMatchConditions, &ExactMatchCondition{Path: stringOrDefault(match.Path.Value, "/")})
 			default:
-				routeAccessor.AddCondition(status.ConditionNotImplemented, metav1.ConditionTrue, status.ReasonPathMatchType, "HTTPRoute.Spec.Rules.PathMatch: Only Prefix match type is supported.")
+				routeAccessor.AddCondition(status.ConditionNotImplemented, metav1.ConditionTrue, status.ReasonPathMatchType, "HTTPRoute.Spec.Rules.PathMatch: Only Prefix match type and Exact match type are supported.")
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


internal/dag/gatewayapi_processor.go
```diff
@@ -120,13 +120,15 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
 
 	for _, rule := range route.Spec.Rules {
 
-		var pathPrefixes []string
+		var pathMatchConditaions []MatchCondition
```
stevesloka, line 123: nit: s/pathMatchConditaions/pathMatchConditions/g

internal/dag/gatewayapi_processor.go
```diff
@@ -120,13 +120,15 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
...
 
 		for _, match := range rule.Matches {
 			switch match.Path.Type {
 			case gatewayapi_v1alpha1.PathMatchPrefix:
-				pathPrefixes = append(pathPrefixes, stringOrDefault(match.Path.Value, "/"))
+				pathMatchConditaions = append(pathMatchConditaions, &PrefixMatchCondition{Prefix: stringOrDefault(match.Path.Value, "/")})
+			case gatewayapi_v1alpha1.PathMatchExact:
+				pathMatchConditaions = append(pathMatchConditaions, &ExactMatchCondition{Path: stringOrDefault(match.Path.Value, "/")})
 			default:
 				p.Error("NOT IMPLEMENTED: Only PathMatchPrefix is currently implemented.")
```
stevesloka, line 133: Can we update this error message?

stevesloka (changes requested):

sunjayBhatia (commented):
See also: https://github.com/projectcontour/contour/pull/3471 which does the same for ingress

there are some pieces in there that overlap with this change and are needed here, namely the sorting and other test changes

We should probably also add some integration testing to make sure this works end to end

youngnick:
@nak3, do you think you could check over #3471 and see if anything needs updating here?

nak3:
Umm... It seems sorting are moved to https://github.com/projectcontour/contour/pull/3504. The sort dag is already called by gateway APIs builder or this PR needs to be changed for the sort dag? @sunjayBhatia

sunjayBhatia:
> Umm... It seems sorting are moved to #3504. The sort dag is already called by gateway APIs builder or this PR needs to be changed for the sort dag? @sunjayBhatia

Ah yeah forgot to update here when I pulled the sorting PR out as its own thing. You shouldn't need to change anything to accommodate sorting route matches in this PR, #3504 should take care of it and it was already merged.

nak3:
Thank you!
Then I think this PR is ready now. cc @stevesloka @youngnick 
For e2e test, I think I should add it to contour-operator.

stevesloka (approved):

nak3:
Rebased to solve confliction.

sunjayBhatia:
> Rebased to solve confliction.
Looks like a CI failure to check out

nak3:
Sorry I missed it. Update it now. Thank you!

sunjayBhatia (approved):

skriss (commented):
Thanks for this @nak3, the code LGTM. It would be great to also add some test cases for exact matching to `_integration/testsuite/gatewayapi/001-patch-condition-match.yaml`.

nak3:
Sure, added integration test. Thank you!

Issue #3453, opened 2021-03-08
cmd/contour: refactor and add test coverage to serve.go


skriss:
Over time, https://github.com/projectcontour/contour/blob/main/cmd/contour/serve.go has grown to be fairly complex. Currently there's no test coverage for it, and most of the logic is in one giant helper function, `doServe(...)`.

We should try to refactor this to break it down into some smaller logical pieces, and also add test coverage (doing this as part of refactoring may be easier than trying to test the entire `doServe` helper right now). 

Tentatively throwing this in the 1.15 milestone but that can be open to discussion.

PR #3481, opened 2021-03-15
pkg/config: Make gatewayconfig optional

### LINKED ISSUE #3479: Make gateway config optional
Currently with GatewayAPI Contour can be configured to watch a single gateway (default: `contour/projectcontour`). 

The problem I've run into is that we have two places to configure listeners, we need to decide which parameter takes precedence if they differ.  

I propose we make the gateway config in the configuration file optional (which what @skriss said in the beginning 😁 ). If Contour finds a Gateway configured then that takes precedence and ignores the listener config flags that might be passed. If not defined then Contour does what it does today since gateway api is not in the mix.  

Contour listener config:
- `--envoy-service-http-address`
- `--envoy-service-https-address`
- `--envoy-service-http-port`
- `--envoy-service-https-port`
 
Gateway listener config:
```
kind: Gateway
metadata:
  name: prod-web
spec:
  gatewayClassName: acme-lb
  listeners:  
  - protocol: HTTP
    port: 80 # <----- port
  - protocol: HTTPS
    port: 443 # <----- port
```

stevesloka:
Takes away the default gatewayapi gateway configuration and makes it optional. This helps determine
which listener configuration values should be used for Contour.

This does change behavior in the config file, however, it's against an alpha api without any usage so to me this is fine to change. 

Fixes #3479

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -397,14 +397,24 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 			ResponseHeadersPolicy: &responseHeadersPolicy,
 		},
 	}
-	if clients.ResourcesExist(k8s.GatewayAPIResources()...) {
-		log.Debug("Gateway API found, adding Gateway API processor")
-		dagProcessors = append(dagProcessors, &dag.GatewayAPIProcessor{
-			FieldLogger: log.WithField("context", "GatewayAPIProcessor"),
-		})
-	} else {
-		log.Debug("Gateway API not found, not adding Gateway API processor")
+
+	// Log that we're using the fallback certificate if configured.
+	if fallbackCert != nil {
+		log.WithField("context", "fallback-certificate").Infof("enabled fallback certificate with secret: %q", fallbackCert)
 	}
+
+	if clientCert != nil {
+		log.WithField("context", "envoy-client-certificate").Infof("enabled client certificate with secret: %q", clientCert)
+	}
+
+	if clients.ResourcesExist(k8s.GatewayAPIResources()...) {
+		if ctx.Config.GatewayConfig != nil {
+			dagProcessors = append(dagProcessors, &dag.GatewayAPIProcessor{
+				FieldLogger: log.WithField("context", "GatewayAPIProcessor"),
+			})
+		}
+	}
+
 	// The listener processor has to go last since it looks at
 	// the output of the other processors.
 	dagProcessors = append(dagProcessors, &dag.ListenerProcessor{})
@@ -416,12 +426,8 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		Observer:        dag.ComposeObservers(append(xdscache.ObserversOf(resources), snapshotHandler)...),
 		Builder: dag.Builder{
 			Source: dag.KubernetesCache{
-				RootNamespaces: ctx.proxyRootNamespaces(),
-				IngressClass:   ctx.ingressClass,
-				Gateway: types.NamespacedName{
-					Name:      ctx.Config.GatewayConfig.Name,
-					Namespace: ctx.Config.GatewayConfig.Namespace,
-				},
+				RootNamespaces:       ctx.proxyRootNamespaces(),
+				IngressClass:         ctx.ingressClass,
 				ConfiguredSecretRefs: configuredSecretRefs,
 				FieldLogger:          log.WithField("context", "KubernetesCache"),
 			},
@@ -430,13 +436,11 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		FieldLogger: log.WithField("context", "contourEventHandler"),
 	}
 
-	// Log that we're using the fallback certificate if configured.
-	if fallbackCert != nil {
-		log.WithField("context", "fallback-certificate").Infof("enabled fallback certificate with secret: %q", fallbackCert)
-	}
-
-	if clientCert != nil {
-		log.WithField("context", "envoy-client-certificate").Infof("enabled client certificate with secret: %q", clientCert)
+	if ctx.Config.GatewayConfig != nil {
+		eventHandler.Builder.Source.ConfiguredGateway = types.NamespacedName{
+			Name:      ctx.Config.GatewayConfig.Name,
+			Namespace: ctx.Config.GatewayConfig.Namespace,
+		}
 	}
 
 	// Wrap eventHandler in a converter for objects from the dynamic client.
@@ -479,23 +483,20 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		log.Warn("DEPRECATED: The flag '--experimental-service-apis' is deprecated and should not be used. Please configure the gateway.name & gateway.namespace in the configuration file to specify which Gateway Contour will be watching.")
 	}
 
-	foundGatewayAPI := false
-	for _, r := range k8s.GatewayAPIResources() {
-		if !clients.ResourcesExist(r) {
-			log.WithField("resource", r).Warn("resource type not present on API server")
-			continue
-		}
-		foundGatewayAPI = true
-
-		if err := informOnResource(clients, r, &dynamicHandler); err != nil {
-			log.WithError(err).WithField("resource", r).Fatal("failed to create informer")
-		}
-	}
-
-	// Only watch namespaces if Gateway API is found.
-	if foundGatewayAPI {
-		if err := informOnResource(clients, k8s.NamespacesResource(), &dynamicHandler); err != nil {
-			log.WithError(err).WithField("resource", k8s.NamespacesResource()).Fatal("failed to create informer")
+	// Only inform on GatewayAPI resources if Gateway API is found.
+	if ctx.Config.GatewayConfig != nil {
+		if clients.ResourcesExist(k8s.GatewayAPIResources()...) {
+			for _, r := range k8s.GatewayAPIResources() {
+				if err := informOnResource(clients, r, &dynamicHandler); err != nil {
+					log.WithError(err).WithField("resource", r).Fatal("failed to create informer")
+				}
+			}
+			// Inform on Namespaces.
+			if err := informOnResource(clients, k8s.NamespacesResource(), &dynamicHandler); err != nil {
+				log.WithError(err).WithField("resource", k8s.NamespacesResource()).Fatal("failed to create informer")
+			}
+		} else {
+			log.Fatalf("GatewayAPI Gateway configured but APIs not installed in cluster.")
 		}
 	}
 
--- a/pkg/config/parameters.go
+++ b/pkg/config/parameters.go
@@ -44,10 +44,17 @@ func (s ServerType) Validate() error {
 }
 
 // Validate the GatewayConfig.
-// Name & Namespace must be specified.
-func (g GatewayParameters) Validate() error {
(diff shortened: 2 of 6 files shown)


_integration/testsuite/install-contour-working.sh
```diff
@@ -35,42 +35,53 @@ readonly REPO=$(cd ${HERE}/../.. && pwd)
...
+#for t in $TAGS ; do
+#    docker tag \
+#        docker.io/projectcontour/contour:"v$$" \
+#        docker.io/projectcontour/contour:$t
+#done
+#
+## Push the Contour build image into the cluster.
+#for t in $TAGS ; do
+#    kind::cluster::load docker.io/projectcontour/contour:$t
+#done
```
skriss, line 65: uncomment

_integration/testsuite/install-contour-working.sh
```diff
@@ -35,42 +35,53 @@ readonly REPO=$(cd ${HERE}/../.. && pwd)
...
-        docker.io/projectcontour/contour:$t
-done
+${KUBECTL} apply -f ${REPO}/examples/contour/00-common.yaml
 
-# Push the Contour build image into the cluster.
-for t in $TAGS ; do
-    kind::cluster::load docker.io/projectcontour/contour:$t
-done
+# Update the config file to set a GatewayAPI Gateway.
+${KUBECTL} apply -f <(sed 's/    # gateway:\n    #   name: contour\n    #   namespace: projectcontour/    gateway:\n           name: contour\n           namespace: projectcontour\n/' < ${REPO}/examples/contour/01-contour-config.yaml)
```
skriss, line 73: we've got several spots now that are customizing the config file for int tests (fallback cert, global rate limiting, and this); I'm thinking we should just maintain a static config file YAML to use for int tests. I'll file an issue for that.

skriss: Actually, I wouldn't expect this change here to stick, since subsequently both the install-fallback-certificate and install-ratelimit-service scripts are updating the configmap with different contents that don't include this. Though, the integration tests seem to be passing. Will look at it some more..

cmd/contour/serve.go
```diff
@@ -396,13 +396,13 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 			ResponseHeadersPolicy: &responseHeadersPolicy,
 		},
 	}
-	if clients.ResourcesExist(k8s.GatewayAPIResources()...) {
+	if ctx.Config.GatewayConfig != nil {
```
skriss, line 399: What do we want to do if a gateway is configured, but the gateway APIs aren't installed in the cluster? Right now I believe we'll log some warnings about not finding the resources but allow Contour to start up (ignoring the gateway config). Might be a better UX to exit, to make it obvious that something is wrong.

cmd/contour/serve.go
```diff
@@ -396,13 +396,13 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
 		log.Debug("Gateway API found, adding Gateway API processor")
 		dagProcessors = append(dagProcessors, &dag.GatewayAPIProcessor{
 			FieldLogger: log.WithField("context", "GatewayAPIProcessor"),
 		})
 	} else {
-		log.Debug("Gateway API not found, not adding Gateway API processor")
+		log.Debug("Gateway API, Gateway not configured, not adding Gateway API processor")
```
skriss, line 405: I *think* that if a Gateway is not configured, then we also shouldn't set up the informers (https://github.com/projectcontour/contour/blob/main/cmd/contour/serve.go#L481-L499)

_integration/testsuite/install-contour-working.sh
skriss: Ah, I think it's due to https://github.com/projectcontour/contour/pull/3481/files#r594627493 being commented out, so the tests are actually running against `main` right now?

stevesloka: Oh I didn't realized the ratelimit stuff was changing the configmap (makes sense though). If you don't mind, I may just update the ratelimit bits to sed the changes like I did for the gateway stuff to keep this moving, then can discuss like you mentioned a better/different approach.

skriss: I'm fine with whatever works for now. I do think a static, integration test-specific file will be easiest to maintain and least convoluted in the long term but I can do that as a follow-up.

skriss: https://github.com/projectcontour/contour/issues/3482

cmd/contour/serve.go
stevesloka: I redid that logic to incorporate your suggestion. 

- If the APIs are not installed & gateway is not configured, all is well
- If the APIs are not installed & gateway IS configured, fatal
- If the APIs are installed & gateway is not configured, all is well

cmd/contour/serve.go
stevesloka: 👍

pkg/config/parameters.go
```diff
@@ -59,7 +66,7 @@ func (g GatewayParameters) Validate() error {
 	}
 
 	if len(errorString) > 0 {
-		return fmt.Errorf("invalid Gateway parameters specified: %s", errorString)
+		return fmt.Errorf("invalid ConfiguredGateway parameters specified: %s", errorString)
```
sunjayBhatia, line 69: Should this and the related changes in `pkg/config/` be reverted? Maybe some other find/replaces that could be reviewed and reverted where it makes sense?

cmd/contour/serve.go
```diff
@@ -475,26 +471,35 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 
 	// Inform on gateway-api types if they are present.
 	if ctx.UseExperimentalServiceAPITypes {
-		log.Warn("DEPRECATED: The flag '--experimental-service-apis' is deprecated and should not be used. Please configure the gateway.name & gateway.namespace in the configuration file to specify which Gateway Contour will be watching.")
+		log.Warn("DEPRECATED: The flag '--experimental-service-apis' is deprecated and should not be used. Please configure the gateway.name & gateway.namespace in the configuration file to specify which ConfiguredGateway Contour will be watching.")
```
sunjayBhatia, line 474: Another example where we might want to revert back to plain `Gateway`

pkg/config/parameters.go
stevesloka: bah my goland refactor tool doesn't always do the right thing. I'll swap those back.

_integration/testsuite/install-contour-working.sh
skriss: this section of code still needs to be uncommented, unless I'm missing something

cmd/contour/serve.go
```diff
@@ -396,14 +396,7 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
 		},
 	}
-	if clients.ResourcesExist(k8s.GatewayAPIResources()...) {
-		log.Debug("Gateway API found, adding Gateway API processor")
-		dagProcessors = append(dagProcessors, &dag.GatewayAPIProcessor{
-			FieldLogger: log.WithField("context", "GatewayAPIProcessor"),
-		})
-	} else {
-		log.Debug("Gateway API not found, not adding Gateway API processor")
-	}
```
skriss: I think this section of code needs to stay where it is, since it needs to happen before adding the listener processor to the list (see comment just below) and then adding all the processors to the builder. I'd just add a check for the gateway being defined in addition to the gateway API resources existing.

stevesloka: yup just found that and pushed a change to move that back to where it needs to be.

_integration/testsuite/install-contour-working.sh
stevesloka: yeah I'm not doin so hot today. It's uncommented now.

_integration/testsuite/install-contour-working.sh
stevesloka: So the integration tests worked? I ran them locally as well. /shrug

cmd/contour/serve.go
```diff
@@ -478,23 +482,25 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
-		}
-	}
+	// Only inform on GatewayAPI resources if Gateway API is found.
+	if ctx.Config.GatewayConfig != nil {
+		if clients.ResourcesExist(k8s.GatewayAPIResources()...) {
+			for _, r := range k8s.GatewayAPIResources() {
+				if !clients.ResourcesExist(r) {
+					log.WithField("resource", r).Warn("resource type not present on API server")
+					continue
+				}
```
skriss, line 492: this is now redundant with L487 that checks that all the resources exist. I'd probably remove this one.

stevesloka: 👍

_integration/testsuite/install-contour-working.sh
skriss: Hmm, I don't understand that, the fallback & ratelimit install scripts should overwrite the config map, removing the gateway.

skriss: Looks like the issue was that the certgen job was still set to an imagePullPolicy of Always, so when it ran it re-pulled the `main` tag from docker hub. Subsequently, when the Contour pods got created (or re-created, since the install-fallback and install-ratelimit scripts both restart the Contour deployment), they used the local `main` tag of the image, which was now the docker hub one, **not** the one built for this PR.

skriss: OK so now the integration test is failing, as expected.

cmd/contour/serve.go
```diff
@@ -397,14 +397,24 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+
+	// Log that we're using the fallback certificate if configured.
+	if fallbackCert != nil {
+		log.WithField("context", "fallback-certificate").Infof("enabled fallback certificate with secret: %q", fallbackCert)
+	}
+
+	if clientCert != nil {
+		log.WithField("context", "envoy-client-certificate").Infof("enabled client certificate with secret: %q", clientCert)
+	}
+
```
youngnick, line 409: This seems like a win, but not really related to this change?

youngnick (approved):
LGTM, this seems like a good way to do it. One small nit about extra changes...

skriss:
Looks like you still have a failing unit test, after a rebase need to change `Gateway` -> `ConfiguredGateway` since it was renamed in this PR: `internal/dag/cache_test.go:1354:4: unknown field 'Gateway' in struct literal of type KubernetesCache`

cmd/contour/serve.go
stevesloka: This wasn't added, just moved around.

stevesloka: We really need to refactor this whole section up. I think @skriss has an open issue for that. There are assumed ordering patterns that need to be kept in place making it difficult to manage.

skriss: Yeah, I would definitely like to spend some time on this soon, this whole thing keeps getting more complicated and potentially error-prone. See #3453 for tracking.

skriss:
:tada: units passing, now just need #3488 to help with int tests!

stevesloka:
W00! I rebased and tests pass again @skriss. =)

skriss (approved):
LGTM

Issue #1691, opened 2019-10-11
Support Access Log Service (ALS)


bgagnon:
Structured logging for the Envoy access logs (ie. JSON access logs) was requested in #624 and implemented in #1511.  Envoy supports a more advanced and flexible access logging option: an [Access Log Service (ALS)](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/observability/access_logging.html).

With this activated, Envoy uses gRPC streams to pass rich and strongly typed protobufs with all details to a sink. This sink is free to do whatever it pleases with the access logs.

Use cases valuable to us:
* produce new time series of metrics (gauges, histograms, counters) based on values observed in access logs
* use a custom JSON format not supported by Envoy (this is a limitation of #1511)
* post-process the log messages to enrich/simplify them (ex: user agent parsing shenanigans where every browser claims to be Chrome, Mozilla, Firefox or all of them)
* enrichment that requires Kubernetes API access (ex: identify the upstream pod and retrieve its metadata such as labels or namespace)

We've implemented this as a proof of concept with the following strategy:
* custom bootstrap on top of Contour's bootstrap that adds the ALS upstream cluster at `127.0.0.1`  (note: this cannot be done via CDS, we've tried -- https://github.com/envoyproxy/envoy/issues/3660)
* ALS receiver running in a sidecar of Envoy -- scales nicely with the DaemonSet
* interceptor that modifies Contour's xDS responses to inject ALS options before they reach Envoy

If Contour supported this, fewer hacks would be needed. I think the minimum would be:
- a new CLI flag that activates this and allows the user to specific the ALS sink address/port/service
- a small modification to LDS/RDS responses when that mode is activated

I don't think Contour needs to provide anything related to _implementing_ an ALS receiver, though we'd be happy to contribute this somewhere if there is interest.

This may be too complex for the scope of Contour and too niche of a feature, but I thought I'd file an issue regardless following @youngnick's recommendation.

stevesloka:
Sounds like this could be implemented like our discussions with the RateLimiting service. Contour configures the endpoint and then the receiving end is responsible for implementing it's side.

davecheney:
ping @m2 for backlog prioritisation.

youngnick:
Now that we've added the `ExtensionService` CRD, it should be much easier to have a way to configure this - we could conceivably have a config file item that asks for Envoy logs to be sent to a given ExtensionService, with failback to stdout if not available.

[...]

PR #3485, opened 2021-03-16
design: add access log (ALS) design doc

abhide:
Updates: #1691

### DIFF
--- a/design/als-design.md
+++ b/design/als-design.md
@@ -1,0 +1,67 @@
+# AccessLog Service (ALS) Support
+
+Status: Draft
+
+## Abstract
+
+This document proposed a design for supporting Envoy's AccessLog Service capabilities in Contour.
+
+## Background
+
+Envoy has the ability to send rich access logs to access logging sinks.
+With AccessLog Service (ALS) enabled, Envoy can send access log messages using gRPC streams to a gRPC access log service. 
+
+This document focuses on configuring access log sinks and enabling Envoy to send access log messages via Contour.
+
+## Goals
+- Enable users to specify ALS configuration
+- Enable users to send HTTP access logs for specific listeners 
+
+## Non Goals
+- TCP Access Log Support
+- As part of first version, communicate between Envoy and ALS sinks will be insecure
+- As part of first version, users can only specify a single ALS sink.
+
+## High-Level Design
+Contour will add support for Envoy's HTTP Access Log Service.
+ALS information will be specified as a [static cluster for Envoy](https://github.com/envoyproxy/envoy/blob/main/source/common/grpc/async_client_manager_impl.cc#L44).
+Contour controller will add details about ALS sink to `access_log` field for each HTTPConnectionManager filter.
+
+## Detailed Design
+To enable Envoy to send access logs to a sink; a static cluster needs to be set for Envoy that will have details about ALS sink.
+There are two ways to add this configuration:
+- ExtensionService
+- Contour Configuration file
+We have chosen to use ExtensionService to enable users to configure ALS sink details.
+
+### ExtensionService
+Contour administrator/operator can add an ExtensionService with ALS configuration details.
+As part of `contour bootstrap`; the ExtensionService will be used to configure a static ALS cluster for Envoy.
+
+An annotation `projectcontour.io/als-admitted: true` needs to be added to the particular ExtensionService.
+This annotations hints `contour bootstrap` to use this ExtensionService for setting the ALS cluster. 
+`contour bootstrap` will populate `Conditions` field to reflect if ALS sink could be correctly configured for Envoy.
+
+Contour controller to will read ALS details from ExtensionService and Conditions field of ExtensionService.
+If there are no configuration errors for ALS; then Contour will populate `access_log` field for each HTTP filter.
+Required changes will be made to Kubernetes RBAC that would enable `contour bootstrap` to read from `ExtensionService`.
+For example:
+```yaml
+apiVersion: projectcontour.io/v1alpha1
+kind: ExtensionService
+metadata:
+  namespace: projectcontour
+  name: accesslog-service
+  annotations:
+    projectcontour.io/als-admitted: true
+spec:
+  protocol: h2
+  services:
+    - name: accesslog-service
+      port: 8080
+```
+### Configuration File
+New command line args will be added to `contour bootstrap` for passing ALS details. `contour bootstrap` will use this ALS details to configure a static ALS sink for Envoy. `contour bootstrap` will write out the ALS details to a Kubernetes Configmap. This needs to be done to ensure that users don't have to specify the same information 2 times; (once for `contour bootstrap` and then for `contour serve`).
+`contour serve` can then read ALS information for this generated configmap and use this information to populate `access_log` field for each HTTP filter.
+
+Required changes will be made to Kubernetes RBAC to enable required access for `contour bootstrap` and `contour serve`.


design/als-design.md
```diff
@@ -0,0 +1,68 @@
...
+  annotations:
+    projectcontour.io/als: true
+spec:
+  protocol: h2
+  services:
+    - name: accesslog-service
+      port: 8080
+```
+### Configuration File
+`contour bootstrap` and `contour serve` commmands can read from a configuration file which is mounted via a Kubernetes Configmap.
```
sunjayBhatia, line 56: I don't think we currently are able to pass a config file to `contour bootstrap` so everything is currently done with CLI flags and arguments.

design/als-design.md
```diff
@@ -0,0 +1,68 @@
...
+There are two ways to add this configuration:
+- ExtensionService
+- Contour Configuration file
+
+### ExtensionService
+Contour administrator/operator can add an ExtensionService with ALS configuration details.
+As part of `contour bootstrap`; the ExtensionService used to configure a static ALS cluster for Envoy.
+Contour controller will set `access_log` field for each HTTP filter.
+
+An annotation `projectcontour.io/als: true` needs to be added to the particular ExtensionService.
```
sunjayBhatia, line 38: This annotation pattern is an interesting one and not one we have used already in Contour

It has the downside of making `boostrap` and `serve` look through all `ExtensionService` resources on startup, but makes it so we don't have to pass the same config in two places since we need to tell both `serve` and `bootstrap` about the ALS service.

design/als-design.md
```diff
@@ -0,0 +1,68 @@
...
+```
+### Configuration File
+`contour bootstrap` and `contour serve` commmands can read from a configuration file which is mounted via a Kubernetes Configmap.
+ALS configuration can we specified as part of the configuration file.
+`contour bootstrap` will then use this information to add a static ALS cluster for Envoy and Contour can then use this information to populate `access_log` field for each HTTP filter.
+The only caveat is `contour bootstrap` and `contour serve` read from separate configuration files.
+
+
+
+## Open Issues:
```
sunjayBhatia, line 63: Some discussion about what happens when someone isn't using `contour bootstrap` to set up their Envoy bootstrap config would be useful. Also some discussion about what happens with the case where the bootstrap does not have the ALS cluster, since the Contour controller sort of has to assume that it is set up correctly.

sunjayBhatia, line 63: We could try to push for Access Logging clusters not having to be static, though from previous Envoy issues and discussion it seems the maintainers are not super into that (which makes sense)

sunjayBhatia (commented):
Some initial comments

design/als-design.md
```diff
@@ -0,0 +1,68 @@
...
+## Goals
+- Enable users to specify ALS configuration
+- Enable users to send HTTP access logs for specific listeners 
+
+## Non Goals
+- TCP Access Log Support
+
+## High-Level Design
+Contour will add support for Envoy's HTTP Access Log Service.
+ALS information will be specified as a static cluster for Envoy and Contour controller will set `access_log` field for each HTTPConnectionManager filter.
```
skriss, line 25: It looks like Envoy requires the ALS cluster to be static, not a CDS cluster, is that right? If so, a brief explanation of that as a constraint (and some links) would be helpful since that guides the design.

sunjayBhatia: +1

design/als-design.md
jpeach: Yeh, the idea of using an annotation is OK in theory, but in practice I suggest that the name of the annotation should be much more specific (ie. it obviously means "use this as a default log sink"), and it somehow needs to be tied to a specific envoy deployment (there can be more than one in the cluster).

Perhaps this use case can be configured more directly in the Contour flags, or in the operator?

design/als-design.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Detailed Design
+To enable Envoy to send access logs to a sink; first a static cluster needs to be set for Envoy.
+There are two ways to add this configuration:
+- ExtensionService
+- Contour Configuration file
+
+### ExtensionService
+Contour administrator/operator can add an ExtensionService with ALS configuration details.
+As part of `contour bootstrap`; the ExtensionService used to configure a static ALS cluster for Envoy.
```
stevesloka, line 35: To do this, the bootstrap command will need a k8s client to first query the api for the ExtensionService. 

To me, setting args (or a config file) is much more simpler for this use-case and the extra indirection of adding an ExtensionService doesn't help but make it more confusing.

stevesloka (changes requested):
Just a small comment on not utilizing an ExtensionService for this implementation. Curious about what others also think.

design/als-design.md
youngnick: The good thing about the ExtensionService is that it allows users to tune the timeouts, which will almost certainly be useful if you are using a sidecar.

I think it will also end up really confusing if we end up with a bunch of different ways to configure extra services in Envoy. We are using ExtensionService for auth and rate limiting already, and will probably need to for tracing. Although it's an extra step, it feels to me like it fits better to have the ExtensionService in there.

design/als-design.md
youngnick: The point about multiple Envoy deployments is a very good one, perhaps the annotation could be something like:
`projectcontour.io/envoy-log-sink: <string>" where string is a value that's defaulted, but can be tuned individually on the `contour serve` and `contour bootsrap` commands?

The whole problem is that the bootstrap and the contour deployment need to have the same info configured somehow, so we either need to make a way for them to coordinate, or configure things twice, which seems very prone to failure.

design/als-design.md
abhide: fixed.

design/als-design.md
abhide: https://github.com/envoyproxy/envoy/issues/3660

abhide (commented):
Thank you for the feedback and review comments. 
I have incorporated the feedback on the review and from the community meeting and addressed the review comments.

design/als-design.md
```diff
@@ -0,0 +1,67 @@
...
+There are two ways to add this configuration:
+- ExtensionService
+- Contour Configuration file
+We have chosen to use ExtensionService to enable users to configure ALS sink details.
+
+### ExtensionService
+Contour administrator/operator can add an ExtensionService with ALS configuration details.
+As part of `contour bootstrap`; the ExtensionService will be used to configure a static ALS cluster for Envoy.
+
+An annotation `projectcontour.io/als-admitted: true` needs to be added to the particular ExtensionService.
```
youngnick, line 41: Doing ExtensionService like this does mean that there can only ever be one ALS per cluster, and we will need to build a totally different mechanism to do more than one, which seems not ideal.

On the other hand, I suspect that most people doing this are probably going to want to send to a single central log sink anyway, or to localhost. @bgagnon is currently using ALS through some clever tricks, maybe he can check out this proposal.

youngnick (changes requested):
I'm a bit worried that we're painting ourselves into a corner with this one. I'd like to summon @bgagnon to check it out and see what he thinks, as he's actually using ALS already.

design/als-design.md
stevesloka: What if we ditch the `contour bootstrap` initContainer and have the Leader in Contour manage this configmap? This would mean that Contour needs to write a configmap RBAC permission in a namespace, but would allow for a single config + setup.

bgagnon:
> I'm a bit worried that we're painting ourselves into a corner with this one. I'd like to summon @bgagnon to check it out and see what he thinks, as he's actually using ALS already.

Unfortunately, I have since changed jobs and no longer using Contour or Envoy in my day job. That being said, I remain a fan, and can still have a look at the design proposal!

abhide:
> > I'm a bit worried that we're painting ourselves into a corner with this one. I'd like to summon @bgagnon to check it out and see what he thinks, as he's actually using ALS already.
> 
> Unfortunately, I have since changed jobs and no longer using Contour or Envoy in my day job. That being said, I remain a fan, and can still have a look at the design proposal!

Thanks @bgagnon. Looking forward to your feedback

design/als-design.md
skriss: 👍  to leaving the door open for multiple ALS in the future.

stevesloka:
I think your plan with the annotations would work @abhide, but just curious if there's a way to make the setup easier for now until we're sure we want multiple ALS instances to be configured. 

What if we setup an ALS cluster by default in the bootstrap config? Then the static cluster defined in the bootstrap config chooses an EDS cluster to fill the endpoints of the static cluster. 

Then in Contour, you'd configure which ExtensionService (or just Contour config) which maps to the ALS service and Contour would fill in the endpoints of the static cluster already defined in the bootstrap config. Contour would translate the endpoints of the configured cluster in Kubernetes to the static EDS endpoint. 

I'm not sure what happens in Envoy if it tries to send traffic to an ALS cluster with no endpoints (perf issue?). 

This also might need to follow @youngnick's suggestion of this being the "default" for Contour such that it allows more instances to be added later on.

youngnick:
I understand the suggestion here @stevesloka, but having an ALS static cluster created by default in every Contour installation feels like we're working hard to avoid something that we really should do anyway (and that's have a way to configure the bootstrap that's not just command line flags). This feels like a good forcing function for us to do something to fix that up.

The sort of thing we're talking about using an annotation for here is exactly what they're designed for, to allow you to find a specific set of things without knowing their name beforehand.

However, both of the options in this proposal as it stands have a problem that I hadn't raised enough before. That is that both of these require the Envoy pod to have direct access to the Kubernetes apiserver, because the bootstrap will need access, it runs in the Envoy pod, and there is no way to different security contexts within a Pod. This is a significant expansion in the Envoy Pod's security footprint, that does worry me more now I've thought about it.

Our options there are either that we accept that security expansion and implement one of these, or move to a convention-based approach like @stevesloka suggested. I don't like having that we'll be doing such a different flow to anything else, and I'm not sure how we would extend this to allow multiple ALS sinks, but both of those seem like less of a problem than the security footprint increase.

abhide:
@stevesloka 
I agree with @youngnick's comment; that always configuring` a static ALS cluster by default doesn't seem like the right thing to do in this case.
Yes, we would need access to Kubernetes APIServer from the envoy pod as part of bootstrap. As far as RBAC is concerned, allowing access to ExtensionService objects might be a one-level better as compared to granting access to ConfigMaps. 
Reason being ExtesnionService is a Contour CRD whereas Configmap is a more general Kubernetes object that would be used by other services as well. 
Also, ExtensionService has a Status field, that can be used to communicate whether ALS was configured successfully as part of bootstrap and whether it can be used by `contour serve` to configure ALS entries for HTTP filters.

design/als-design.md
abhide: Yes, I would suggest we add 2 annotations that would be specific to ALS
`projectcontour.io/als-admitted: true` -> Indicates whether to configure this ALS
`projectcontour.io/envoy-log-sink: <name>` -> ALS name or default.

abhide:
@youngnick @stevesloka @skriss Seems like we are in consensus about usage of ExtensionService and exposing the ability to configure multiple ALS sinks (for future extensibility)? 
If yes, would like to know the next steps to get  a :+1: on this design so that implementation work could begin.

stevesloka:
I'm always a fan for simpler than extra setup/config. If we're not sure folks would even want multiple ALS instances, then to me, a simpler approach is to have one setting and see what the community says about their needs, then change it and make it more complex later on. Then don't need to deal with the init requiring access to the API server, etc, annotations, etc. 

I think your approach would work @abhide, thanks for putting this together, but I'll leave this up to what other folks think since it seems my initial implementation ideas are different. 

@skriss @sunjayBhatia do you have some time to look over this design doc + comments today?

sunjayBhatia (commented):
Some issues that need some more detail written down and worked out in the design I think would be around some error/misuse flows etc. and what happens there. Especially calling out what happens when someone changes something on the ExtensionService, wants to switch between ALS servers, restarts the Envoy daemonset, isn't using `contour bootstrap` to manage their Envoy bootstrap and makes a mistake specifying the ALS cluster, deletes an ALS ExtensionService etc. With all of our other ExtensionService uses we do not have the need to configure a static cluster in the bootstrap config so most of these issues are completely handled within the Contour controller only. However since we now have multiple components interacting, its a little more complicated.

It would also be good to summarize some of the questions and answers/proposed alternatives in discussion threads in this PR in the design as an "Alternatives Considered"

sunjayBhatia (commented):
From discussion in office hours today, there is at least one more option to add to the possible implementations:

As we seem to be moving forward with a plan to enable Contour to manage the Envoy data-plane (see https://docs.google.com/document/d/1kOrGqHXULYO6YLxGCOtVyjz_7mfcW7Yi1WXCpQz6_Qo/edit#heading=h.awh08ldngzyk) it may be that we have a future where Contour itself writes out the bootstrap config, which would really simplify things. However, if Contour is not managing the Envoy bootstrap/deployment, we still have an issue.

skriss:
A few comments:

- I don't think we *need* to support multiple ALS sinks initially, I just didn't want to make design choices that would prevent us from doing that down the road if we find out users want it. I'm fine with supporting at most one ALS per Contour/Envoy initially.
- We talked some at office hours today about the possibility of rearchitecting how Contour generates the Envoy bootstrap config, so that instead of generating it from an initContainer in the Envoy daemonset, we have the main `contour serve` command generate it as a ConfigMap when Contour starts up, and then all the Envoys would mount the ConfigMap as a volume. The benefits of this are that (a) the `envoy` service account doesn't need Kubernetes API access; and (b) the ALS config could now go into the Contour config file, so we wouldn't have to use the new annotation pattern to identify the ALS. This approach seems to solve both of these issues nicely, but it is a fairly large change to the startup sequence for Contour+Envoy, and we'd have to think about how to support users who want to bring their own bootstrap, etc.
- Another alternative idea to consider (sorry if we already discussed this): but couldn't the `bootstrap` init container mount the **Contour** config file as a volume, and read ALS config info from there? That would eliminate the need for the annotation; we could just identify the ALS ExtensionService in the config file, similarly to how the rate limit service is configured today, and then both `contour bootstrap` and `contour serve` would read it from there. This shouldn't require the `envoy` service account to have any new RBAC permissions.

youngnick:
> A few comments:
> 
> * I don't think we _need_ to support multiple ALS sinks initially, I just didn't want to make design choices that would prevent us from doing that down the road if we find out users want it. I'm fine with supporting at most one ALS per Contour/Envoy initially.

That was my feeling as well. It felt to me like the "add a static cluster by default" would be hard to move away from, this feels less hard to change our mind.

> * We talked some at office hours today about the possibility of rearchitecting how Contour generates the Envoy bootstrap config, so that instead of generating it from an initContainer in the Envoy daemonset, we have the main `contour serve` command generate it as a ConfigMap when Contour starts up, and then all the Envoys would mount the ConfigMap as a volume. The benefits of this are that (a) the `envoy` service account doesn't need Kubernetes API access; and (b) the ALS config could now go into the Contour config file, so we wouldn't have to use the new annotation pattern to identify the ALS. This approach seems to solve both of these issues nicely, but it is a fairly large change to the startup sequence for Contour+Envoy, and we'd have to think about how to support users who want to bring their own bootstrap, etc.

This idea seems neat, and does away with some of these problems, while also allowing us to do more clever bootstrap things in the future. Supporting users who want to do their own bootstrap seems straightforward, we can have a boolean to turn off managing the configmap, and/or have users change their Envoy deployment to not use the generated/managed ConfigMap.
Overall, a neat idea, that plays well in the case that Contour ends up managing the Envoys as well. +1 from me.

> * Another alternative idea to consider (sorry if we already discussed this): but couldn't the `bootstrap` init container mount the **Contour** config file as a volume, and read ALS config info from there? That would eliminate the need for the annotation; we could just identify the ALS ExtensionService in the config file, similarly to how the rate limit service is configured today, and then both `contour bootstrap` and `contour serve` would read it from there. This shouldn't require the `envoy` service account to have any new RBAC permissions since it's a volume mount, not an API read (this seems to be true from my testing but we should definitely confirm). I guess `envoy` would still need to be able to read `ExtensionServices`, though, unless we also did away with using an ExtensionService at all.

Yes, @abhide and I talked about this one, and the change in the size of the Envoy RBAC still seems like the sticking point for me. Right now, Envoy has *no* Kubernetes RBAC, or even the token mounted in, so there's zero privilege escalation risk. I'd prefer to keep this if we can.

youngnick:
I should add, I think that Sunjay's thoughts about adding more information about the flows for various error cases is very useful, as is some more background in the "Alternatives considered" area.

youngnick:
@abhide, what's your status here? No problem if you're intending to come back to it, but if not, one of us can probably finish this up for you.

abhide:
> @abhide, what's your status here? No problem if you're intending to come back to it, but if not, one of us can probably finish this up for you.

Hey @youngnick I will fix the review comments and push another PR.
Sorry, had been busy lately. Will look into this in a next few days.

youngnick:
Great, thanks @abhide!

PR #3486, opened 2021-03-16
makefile: Remove --new from make lint-golint task

stevesloka:
When running make checkall locally on my machine, I wouldn't get any errors, but running in CI I would.
If I take out the `--new` flag to the golangci-lint binary I get the matching errors found in CI runs.

My typical workflow is to run a `make checkall` to find any issues before waiting for CI to run, but this
wasn't working.


skriss:
Weird, since our CI run has `only-new-issues: true`: https://github.com/projectcontour/contour/blob/main/.github/workflows/prbuild.yaml#L84.

Ideally we'd be at a point where we could drop that too, but I'm guessing there are a bunch of pre-existing issues we'd have to address or add to an ignore list?

youngnick:
I would guess this is about what constitutes "new" given the Actions build environment vs a laptop.

youngnick (approved):

skriss:
I tried turning off `only-new-issues` on the CI run too: https://github.com/projectcontour/contour/compare/main...skriss:exp-lint (`false` is the default)

It passes: https://github.com/skriss/contour/runs/2131769867?check_suite_focus=true

So should we go ahead and change that too, since we're currently clean?

skriss:
@stevesloka mind adding that change to this PR?

stevesloka:
Should be there now @skriss

skriss (approved):
LGTM pending CI (and 🤞  that this also passes on `main`!)

PR #3488, opened 2021-03-16
integration: Combine integration test setup into single file

### LINKED ISSUE #3482: integration tests should have a single YAML file defining the Contour config map
Today, the process for creating the Contour config file ConfigMap is:

1. Apply `examples/contour/01-contour-config.yaml` as a baseline (https://github.com/projectcontour/contour/blob/main/_integration/testsuite/install-contour-working.sh#L71-L73)
2. Apply a customized version of it to include a fallback certificate, then restart Contour to pick up the change (https://github.com/projectcontour/contour/blob/main/_integration/testsuite/install-fallback-certificate.sh#L69-L92)
3. Repeat step 2, but for configuring a rate limit service (note that this one **also** contains the fallback-cert config so as not to drop it) (https://github.com/projectcontour/contour/blob/main/_integration/testsuite/install-ratelimit-service.sh#L70-L100).

Since ultimately, a single ConfigMap is produced and used for all integration tests, I think this would be easier if we just defined a static `contour-config.yaml` for the integration tests that contains all the necessary settings, and apply it once instead of `examples/contour/01-contour-config.yaml`.

Longer-term, it'd be even better if we could use a different config map per integration test case, but that's a separate issue.

stevesloka:
Fixes #3482


stevesloka:
@skriss not sure what you were working on, but I hacked this up.

_integration/testsuite/install-contour-working.sh
```diff
@@ -64,13 +64,129 @@ for t in $TAGS ; do
     kind::cluster::load docker.io/projectcontour/contour:$t
 done
 
+
 # Install Contour
-#
+
+${KUBECTL} apply -f ${REPO}/examples/contour/00-common.yaml
+
+# Update the config file to set a GatewayAPI Gateway.
```
skriss, line 72: can delete this comment since it's not relevant here.

_integration/testsuite/install-contour-working.sh
```diff
@@ -64,13 +64,129 @@ for t in $TAGS ; do
...
+  contour.yaml: |
+    gateway:
+      name: contour
+      namespace: projectcontour
+    rateLimitService:
+      extensionService: projectcontour/ratelimit
+      domain: contour
+      failOpen: false
+    # The options below are copied from
+    # install-fallback-certificate.sh.
```
skriss, line 101: can delete this comment since it's no longer relevant.

_integration/testsuite/install-contour-working.sh
```diff
@@ -64,13 +64,129 @@ for t in $TAGS ; do
...
+    tls:
+      fallback-certificate:
+        name: fallback-cert
+        namespace: projectcontour
+    accesslog-format: envoy
+    disablePermitInsecure: false
```
skriss, line 107: I believe these two settings are both defaults, so can be removed, but double-check me there.

_integration/testsuite/install-contour-working.sh
```diff
@@ -64,13 +64,129 @@ for t in $TAGS ; do
...
 # Manifests use the "Always" image pull policy, which forces the kubelet to re-fetch from
 # DockerHub, which is why we have to update policy to `IfNotPresent`.
-for y in ${REPO}/examples/contour/*.yaml ; do
-  ${KUBECTL} apply -f <(sed 's/imagePullPolicy: Always/imagePullPolicy: IfNotPresent/g' < "$y")
-done
+${KUBECTL} apply -f <(sed 's/imagePullPolicy: Always/imagePullPolicy: IfNotPresent/g' < ${REPO}/examples/contour/02-job-certgen.yaml)
+${KUBECTL} apply -f <(sed 's/imagePullPolicy: Always/imagePullPolicy: IfNotPresent/g' < ${REPO}/examples/contour/03-contour.yaml)
+${KUBECTL} apply -f <(sed 's/imagePullPolicy: Always/imagePullPolicy: IfNotPresent/g' < ${REPO}/examples/contour/03-envoy.yaml)
+
+${KUBECTL} apply -f - <<EOF
```
skriss, line 85: maybe worth a comment noting that the Contour pod won't schedule until this ConfigMap is created, since it mounts it as a volume, which is why it's fine to create the ConfigMap *after* creating the Contour deployment.

_integration/testsuite/install-contour-working.sh
```diff
@@ -64,13 +64,129 @@ for t in $TAGS ; do
...
+    tls:
+      fallback-certificate:
+        name: fallback-cert
+        namespace: projectcontour
+    accesslog-format: envoy
+    disablePermitInsecure: false
+EOF
+
+
+# Install Certs
```
skriss, line 111:
```suggestion
# Install fallback cert
```

skriss (commented):
This looks reasonable, just a handful of minor comments.

_integration/testsuite/install-contour-working.sh
stevesloka: yup, they are the same.

skriss (approved):
LGTM!

sunjayBhatia (approved):
LGTM, nice to have this in one spot so installing everything doesn't take as long and is sure to actually update everything before continuing

PR #3499, opened 2021-03-18
Adds Validation Package

danehans:
Refactors https://github.com/projectcontour/contour/pull/3477 into a validation pkg that can be used by https://github.com/projectcontour/contour-operator/pull/226 and others.

### DIFF
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -14,15 +14,12 @@
 package dag
 
 import (
-	"net"
-	"strings"
-
+	"github.com/projectcontour/contour/pkg/validation"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


skriss:
One option for these validation helpers would be to put them in a new module/repo under projectcontour, with a v0 version, so that we don't have to have compatibility guarantees for now. As soon as they're publicly exported in `projectcontour/contour`, we're not support to change the API.

I'd like to see some discussion of this vs other approaches for these shared libraries.

danehans:
Closing in favor of adding validation upstream, xref: https://github.com/kubernetes-sigs/gateway-api/issues/487

danehans:
xref https://github.com/kubernetes-sigs/gateway-api/pull/597 to add validation pkg upstream.

PR #3501, opened 2021-03-18
cmd/contour: Added customised adminPort for the shutdown process

### LINKED ISSUE #3490: shutdown manager uses hardcoded admin port
we use customised admin port for our envoy pod through [--admin-port](https://github.com/projectcontour/contour/blob/64db9081049afb968b24be94541cadf074fd75ad/cmd/contour/bootstrap.go#L30) but it seems shutdown manager uses [ hardcoded port 9001](https://github.com/projectcontour/contour/blob/main/cmd/contour/shutdownmanager.go#L33) which causes problem during graceful shutdown.

shutdown manager should also take the same admin port or at least configurable.

**Environment:**

- Contour version: all version seems to be impacted including latest v1.13.1
- Kubernetes version: v1.19.7
- OS: Flatcar OS

alessandroargentieri:
To pass adminPort (got from the flag --admin-port) to shutdownManager,
we need to enrich the shutdownContext struct with this new attribute.
Then we are able to use it in the shutdownHandler() to shutdown Envoy.

Fixes #3490

Signed-off-by: alessandroargentieri <alexmawashi87@gmail.com>

### DIFF
--- a/cmd/contour/contour.go
+++ b/cmd/contour/contour.go
@@ -34,10 +34,11 @@ func main() {
 	envoyCmd := app.Command("envoy", "Sub-command for envoy actions.")
 	sdm, shutdownManagerCtx := registerShutdownManager(envoyCmd, log)
 
+	bootstrap, bootstrapCtx := registerBootstrap(app)
+
 	// Add a "shutdown" command which initiates an Envoy shutdown sequence.
 	sdmShutdown, sdmShutdownCtx := registerShutdown(envoyCmd, log)
 
-	bootstrap, bootstrapCtx := registerBootstrap(app)
 	certgenApp, certgenConfig := registerCertGen(app)
 
 	cli := app.Command("cli", "A CLI client for the Contour Kubernetes ingress controller.")
--- a/cmd/contour/shutdownmanager.go
+++ b/cmd/contour/shutdownmanager.go
@@ -30,9 +30,9 @@ import (
 )
 
 const (
-	prometheusURL      = "http://localhost:9001/stats/prometheus"
-	healthcheckFailURL = "http://localhost:9001/healthcheck/fail"
-	prometheusStat     = "envoy_http_downstream_cx_active"
+	prometheusURLFormat      = "http://localhost:%d/stats/prometheus"
+	healthcheckFailURLFormat = "http://localhost:%d/healthcheck/fail"
+	prometheusStat           = "envoy_http_downstream_cx_active"
 )
 
 // shutdownReadyFile is the default file path used in the /shutdown endpoint.
@@ -70,6 +70,9 @@ type shutdownContext struct {
 	// that can be open when polling for active connections in Envoy
 	minOpenConnections int
 
+	// adminPort defines the port for our envoy pod, being configurable through --admin-port flag
+	adminPort int
+
 	logrus.FieldLogger
 }
 
@@ -83,12 +86,12 @@ func newShutdownManagerContext() *shutdownmanagerContext {
 }
 
 func newShutdownContext() *shutdownContext {
-	// Set defaults for parameters which are then overridden via flags, ENV, or ConfigFile
 	return &shutdownContext{
 		checkInterval:      5 * time.Second,
 		checkDelay:         60 * time.Second,
 		drainDelay:         0,
 		minOpenConnections: 0,
+		adminPort:          9001,
 	}
 }
 
@@ -142,7 +145,7 @@ func (s *shutdownContext) shutdownHandler() {
 	// Send shutdown signal to Envoy to start draining connections
 	s.Infof("failing envoy healthchecks")
 
-	// Retry any failures to shutdownEnvoy() in a Backoff time window
+	// Retry any failures to shutdownEnvoy(s.adminPort) in a Backoff time window
 	// doing 4 total attempts, multiplying the Duration by the Factor
 	// for each iteration.
 	err := retry.OnError(wait.Backoff{
@@ -155,7 +158,7 @@ func (s *shutdownContext) shutdownHandler() {
 		return true
 	}, func() error {
 		s.Infof("attempting to shutdown")
-		return shutdownEnvoy()
+		return shutdownEnvoy(s.adminPort)
 	})
 	if err != nil {
 		// May be conflict if max retries were hit, or may be something unrelated
@@ -167,7 +170,7 @@ func (s *shutdownContext) shutdownHandler() {
 	time.Sleep(s.checkDelay)
 
 	for {
-		openConnections, err := getOpenConnections()
+		openConnections, err := getOpenConnections(s.adminPort)
 		if err != nil {
 			s.Error(err)
 		} else {
@@ -193,7 +196,9 @@ func (s *shutdownContext) shutdownHandler() {
 }
 
 // shutdownEnvoy sends a POST request to /healthcheck/fail to tell Envoy to start draining connections
-func shutdownEnvoy() error {
+func shutdownEnvoy(adminPort int) error {
+	healthcheckFailURL := fmt.Sprintf(healthcheckFailURLFormat, adminPort)
+	/* #nosec */
 	resp, err := http.Post(healthcheckFailURL, "", nil)
 	if err != nil {
 		return fmt.Errorf("creating healthcheck fail POST request failed: %s", err)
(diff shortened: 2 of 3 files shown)


cmd/contour/contour.go
```diff
@@ -34,10 +34,11 @@ func main() {
 	envoyCmd := app.Command("envoy", "Sub-command for envoy actions.")
 	sdm, shutdownManagerCtx := registerShutdownManager(envoyCmd, log)
 
+	bootstrap, bootstrapCtx := registerBootstrap(app)
+
 	// Add a "shutdown" command which initiates an Envoy shutdown sequence.
-	sdmShutdown, sdmShutdownCtx := registerShutdown(envoyCmd, log)
+	sdmShutdown, sdmShutdownCtx := registerShutdown(envoyCmd, bootstrapCtx.AdminPort, log)
```
youngnick, line 40: I see what you've tried to do here, but I don't think that what this is doing is what you want.

Currently, this will bring in the default port from the bootstrap context, but it won't get overridden by a parameter to the `contour shutdown`  command, so you won't actually have a way to change it.

cmd/contour/shutdownmanager.go
```diff
@@ -82,13 +85,17 @@ func newShutdownManagerContext() *shutdownmanagerContext {
 	}
 }
 
-func newShutdownContext() *shutdownContext {
+func newShutdownContext(adminPort int) *shutdownContext {
```
youngnick, line 88: I'd call this `defaultAdminPort`, as ideally it should be setting the default, that can be overridden by a flag in the outer `registerShutdown` function.

cmd/contour/shutdownmanager.go
```diff
@@ -277,8 +288,8 @@ func registerShutdownManager(cmd *kingpin.CmdClause, log logrus.FieldLogger) (*k
 }
 
 // registerShutdown registers the envoy shutdown sub-command and flags
-func registerShutdown(cmd *kingpin.CmdClause, log logrus.FieldLogger) (*kingpin.CmdClause, *shutdownContext) {
-	ctx := newShutdownContext()
+func registerShutdown(cmd *kingpin.CmdClause, adminPort int, log logrus.FieldLogger) (*kingpin.CmdClause, *shutdownContext) {
```
youngnick, line 291: As with `newShutdownContext`, I'd rename this to `defaultAdminPort`.

cmd/contour/shutdownmanager.go
```diff
@@ -277,8 +288,8 @@ func registerShutdownManager(cmd *kingpin.CmdClause, log logrus.FieldLogger) (*k
...
+	ctx := newShutdownContext(adminPort)
 	ctx.FieldLogger = log.WithField("context", "shutdown")
 
 	shutdown := cmd.Command("shutdown", "Initiate an shutdown sequence which configures Envoy to begin draining connections.")
```
youngnick, line 295: You'll also need to duplicate the `admin-port` flag from `registerBootstrap` here, so that `contour shutdown` will pick up the same flag.

```suggestion
	shutdown := cmd.Command("shutdown", "Initiate an shutdown sequence which configures Envoy to begin draining connections.")
	shutdown.Flag("admin-port", "Envoy admin interface port.").IntVar(&ctx.adminPort)
```

youngnick (changes requested):
Nice work @alessandroargentieri, this is really close! Just need to make sure that the `contour shutdown` command has the argument available as well.

cmd/contour/contour.go
alessandroargentierimilkman: Hi @youngnick I hope to have correctly addressed all the requested changes now. Thanks!

youngnick (approved):
LGTM

cmd/contour/contour.go
stevesloka, line 40: The `bootstrapCtx.AdminPort` arg should come out and need to use the flag defined in the shutdown manager command: https://github.com/projectcontour/contour/pull/3501/files#diff-ab19309345ae701385f9f0a7315157a1a76ad909f3dacef2e8ad9fb973a4232fR296

cmd/contour/shutdownmanager.go
```diff
@@ -277,11 +288,12 @@ func registerShutdownManager(cmd *kingpin.CmdClause, log logrus.FieldLogger) (*k
 }
 
 // registerShutdown registers the envoy shutdown sub-command and flags
-func registerShutdown(cmd *kingpin.CmdClause, log logrus.FieldLogger) (*kingpin.CmdClause, *shutdownContext) {
-	ctx := newShutdownContext()
+func registerShutdown(cmd *kingpin.CmdClause, defaultAdminPort int, log logrus.FieldLogger) (*kingpin.CmdClause, *shutdownContext) {
+	ctx := newShutdownContext(defaultAdminPort)
```
stevesloka, line 292: This should pass `ctx.adminPort` and not `defaultAdminPort` which should be removed.

cmd/contour/shutdownmanager.go
```diff
@@ -277,11 +288,12 @@ func registerShutdownManager(cmd *kingpin.CmdClause, log logrus.FieldLogger) (*k
...
 	ctx.FieldLogger = log.WithField("context", "shutdown")
 
 	shutdown := cmd.Command("shutdown", "Initiate an shutdown sequence which configures Envoy to begin draining connections.")
+	shutdown.Flag("admin-port", "Envoy admin interface port.").IntVar(&ctx.adminPort)
```
stevesloka, line 296:
```suggestion
	shutdown.Flag("admin-port", "Envoy admin interface port.").Default("9001").IntVar(&ctx.adminPort)
```

Should be able to default the value here which removes the need for the check on https://github.com/projectcontour/contour/pull/3501/files#diff-ab19309345ae701385f9f0a7315157a1a76ad909f3dacef2e8ad9fb973a4232fR90

stevesloka (changes requested):
This is looking good @alessandroargentieri! Just a couple comments. 

Also, could you update the log message in `doShutdownManger` to log the port that this being used for Envoy's Admin? This might be helpful in the future.

cmd/contour/contour.go
alessandroargentierimilkman: Hi @stevesloka I'm not getting what you mean. That param `bootstrapCtx.AdminPort` is exactly the flag (admin-port) specified in the shutdown manager command. So what is the change?

cmd/contour/shutdownmanager.go
alessandroargentierimilkman: @stevesloka same thing here: the defaultAdminPort param passed is exactly the ctx.AdminPort which is passed to the registerShutdown function. So, what's the thing to be changed here?

alessandroargentierimilkman: @stevesloka same thing here: the defaultAdminPort param passed is exactly the ctx.AdminPort which is passed to the registerShutdown function. So, what's the thing to be changed here?

cmd/contour/contour.go
alessandroargentierimilkman: The fact is, that if I remove that param, practically I revert all my changes, leaving the code the way it was before and leaving the original issue unresolved.

alessandroargentieri: Hi, @stevesloka thanks for the kind suggestions. I've applied all the changes you told me. Hope it's good

cmd/contour/shutdownmanager.go
```diff
@@ -136,13 +139,16 @@ func (s *shutdownmanagerContext) shutdownReadyHandler(w http.ResponseWriter, r *
 // shutdownHandler is called from a pod preStop hook, where it will block pod shutdown
 // until envoy is able to drain connections to below the min-open threshold.
 func (s *shutdownContext) shutdownHandler() {
+
+	fmt.Println("---admin-port: ", s.adminPort)
```
stevesloka, line 143: nit: Remove the log.

stevesloka (changes requested):
LGTM @alessandroargentieri with just some small nit with an extra log line. Thanks for getting this added! 🎉

cmd/contour/shutdownmanager.go
alessandroargentieri: Hi @stevesloka, I removed the log and for that minimal change the codecoverage tool blocks my PR. With the log line (1 line more), the PR is allowed. I don't understand why. What would you suggest to get the work done? I'm not sure how to test those functions in which I made the little adds, because I didn't find any test on those too.

stevesloka (approved):
/lgtm Thanks for getting this done @alessandroargentieri! No worries about the code coverage tool, this is all good. =)

PR #3502, opened 2021-03-19
Updates Gateway API Design Doc

danehans:
Updates the Gateway API design doc based on the operator managing Gateway objects.

xref: https://docs.google.com/document/d/1fq_zSozkPeImD1SoBvxQ6_mXmtqfGCn_77UQgFn0mAY/edit?usp=sharing

### DIFF
--- a/design/gateway-apis-implementation.md
+++ b/design/gateway-apis-implementation.md
@@ -45,11 +45,13 @@ There should never be annotations required for functionality on Gateway API obje
 
 
 ## Goals
-- Define a data model and implementation for Contour's Gateway API support, covering the v1alpha1 version of the Gateway API.
-- Layer 7 support only, which means HTTPRoutes, TLSRoutes only.
+- Define a data model and implementation for Contour's Gateway API support, covering Gateway API [v0.2.0][https://github.com/kubernetes-sigs/gateway-api/releases/tag/v0.2.0].
+- Layer 7 support only, which means Contour is only responsible for reconciling HTTPRoute and TLSRoute types. It is expected
+that Contour will need to watch other objects, i.e. Gateway, to properaly surface HTTPRoute/TLSRoute status.
 
 ## Non Goals
 - No TCPRoute or UDPRoute support, that is, no support for arbitrary TCP/UDP forwarding.
+- Manage infrastructure-level Gateway API resources, i.e. Gateway.
 
 
 ## High-Level Design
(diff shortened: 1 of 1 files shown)


design/gateway-apis-implementation.md
```diff
@@ -97,24 +120,31 @@ It's expected that things that check the status will also check the `observedGen
...
+A separate Gateway controller is responsible for managing Gateway objects. Contour will read and watch its configured Gateway for
+the purposes of managing HTTPRoute and TLSRoute objects. Contour is not able to specify the address that the Envoy deployment
+will listen on, and so it will ignore any entries in the Gateway's `spec.addresses` field. For Contour, the key Gateway
+section are the Listeners. These define how an implementation should listen for traffic described by Routes.
 
 For Listeners:
-- Listeners inside Gateways will be merged where possible.
+- Listeners inside Gateways will be merged where possible. Note that only two listeners are supported, HTTP and HTTPS.
 - Listeners that specify a port that the Contour controller considers the default secure port will have a HTTP->HTTPS redirect created for them.
 - Listeners that specify a port that is not exposed via the Envoy service will be rejected.
```
skriss, line 131: is there a corresponding Condition for this?

design/gateway-apis-implementation.md
```diff
@@ -97,24 +120,31 @@ It's expected that things that check the status will also check the `observedGen
...
-- Conflicts within a Gateway will result in the relevant Listeners both being rejected (as there is no way to determine which one was first).
+- Conflicts within a Gateway will result in the both Listeners being rejected (as there is no way to determine which one was first).
```
skriss, line 132: Is there a Condition for this?

design/gateway-apis-implementation.md
```diff
@@ -97,24 +120,31 @@ It's expected that things that check the status will also check the `observedGen
...
+  The "ResolvedRefs=False" Listener condition and "InvalidCertificateRef" reason will be surfaced by
+  the gateway controller.
+- Listeners that specify the same port are ignored. The "Conflicted=True" Listener condition and "ProtocolConflict"
+  reason will be surfaced by the gateway controller.
+- A Listener will be ignored if it specifies a protocol of "HTTPS" or "TLS" and mode "Terminate", but does not specify a certificate reference.
+  The "ResolvedRefs=False" Listener condition and "InvalidCertificateRef" reason will be surfaced by the gateway controller.
+- A Listener will be ignored if it specifies an "HTTPS" or "TLS" protocol and a certificate reference that is not a Secret.
+  The "ResolvedRefs=False" Listener condition and "InvalidCertificateRef" reason will be surfaced by the gateway controller.
+- A Listener will be ignored if it specifies "Selector" as a namespace selector type but does not specify matchExpression
+  or matchLabels. The "ResolvedRefs=False" Listener condition and "InvalidRoutesRef" reason will be surfaced by the gateway controller.
```
skriss, line 147: Is this an exhaustive list of all the Conditions + Reasons which should cause Contour to *not* try to program a Listener/its routes? And otherwise, it *should* try to program that listener/routes?

design/gateway-apis-implementation.md
```diff
@@ -264,16 +294,7 @@ Precedence rules in the spec must also be followed.
...
-This slice contains `ListenerStatus` entries that are expected to be keyed by `port`.
-If more than one Listener shares the same port, the `ListenerStatus` reports the combined status.
-
-As the API currently stands, Contour must combine the statuses for all Listeners that share a `port`, including Listeners that are being merged.
-If this proves awkward, we may need to ask for some upstream changes.
-
-Whenever Contour looks at a Listener, it will add an `Admitted` Condition.
-In the event that a ListenerStatus has everything okay (all mergeable Listeners pass validation), the `Admitted` condition will be `status: true`.
-If there are any errors, the `Admitted` Condition will be `status: false`, and the Reason will tell you more about why.
+The gateway controller is responsible for managing Gateway status.
```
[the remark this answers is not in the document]
danehans: @skriss I'm trying to generalize the design to use "gateway controller" since it appears that the maintainer team has not achieved consensus on where the Gateway controller should live. A potential approach to unblock us, is to support multiple [controller](https://github.com/kubernetes-sigs/gateway-api/blob/v0.2.0/apis/v1alpha1/gatewayclass_types.go#L49-L63) domain/paths. 

__Option 1- Operator Managed:__ The operator will own a Gateway of class "projectcontour.io/contour-operator" and be responsible for managing the Gateway, including status. Contour will be responsible reading Gateway status to manage HTTPRoutes and TLSRoutes, including status. 

__Option 2- Contour Managed:__ The operator will __not__ manage a Gateway of class "projectcontour.io/contour". Instead, Contour will be responsible for managing the Gateway, including status. The GatewayClass should either not reference an object or reference an object other than the Contour custom resource... maybe a ConfigMap. With this approach, users are expected to manage Contour using standard Kubernetes manifests. Contour will need to ensure the following for a Gateway:
- If [addresses](https://github.com/kubernetes-sigs/gateway-api/blob/v0.2.0/apis/v1alpha1/gateway_types.go#L115-L132) are specified, ensure they match Envoy service addresses and surface the appropriate Gateway status condition if not.
- Ensure each [listener port/protocol](https://github.com/kubernetes-sigs/gateway-api/blob/v0.2.0/apis/v1alpha1/gateway_types.go#L163-L184) matches the port name/protocol in the Envoy service and surface the appropriate Gateway/Listener status condition if not.

Supporting both approaches will be a short-term approach that allows us to figure out the model that works best for Contour. Thoughts? cc: @stevesloka @sunjayBhatia @youngnick

design/gateway-apis-implementation.md
```diff
@@ -124,7 +154,7 @@ this will require careful test case design.
 
 The output of this watcher is both Envoy configuration, and a list of kind/name/namespace details for Routes to watch.
 
-Contour will only update the `status` of Gateway objects.
+Contour will only update the `status` of HTTPRoute and TLSRoute objects.
```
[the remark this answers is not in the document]
danehans: Yes, [RouteGatewayStatus](https://github.com/kubernetes-sigs/gateway-api/blob/v0.2.0/apis/v1alpha1/shared_types.go#L178-L195) should always be managed by Contour using condition type "Admitted". When "Admitted=False", Contour should handle the conflict according to the [guidelines](https://gateway-api.sigs.k8s.io/guidelines/). The gateway controller will watch HTTPRoutes/TLSRoutes and update Gateway/Listener status accordingly.

design/gateway-apis-implementation.md
danehans: The update intends to cover the different Listener condition types and reasons. Let me know if I missed anything.

danehans:
@skriss commit `775c201` addresses the above resolved feedback.

danehans:
@sunjayBhatia commit `f43c572` updates the PR based on your feedback above, PTAL.

design/gateway-apis-implementation.md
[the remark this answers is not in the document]
youngnick: While I agree that having something that doesn't require installing two things would be preferable, we need to not be pausing development on Gateway support while we figure out what that looks like. What we have in this current design is deliverable with what we currently have, and I would rather have something working first before spending a lot of engineering effort refactoring.

youngnick (approved):
This LGTM, as a place to get what we have right now working. I agree that we should discuss further options, but any of the options that I've seen discussed so far are multiple-month engineering efforts, and think having something working at all is better than nothing.

stevesloka:
> This LGTM, as a place to get what we have right now working. I agree that we should discuss further options, but any of the options that I've seen discussed so far are multiple-month engineering efforts, and think having something working at all is better than nothing.

I disagree, now is the time to figure these things out while the api is alpha and there are no users. If @danehans, @skriss, & @sunjayBhatia feel the same the that's fine, but if we don't do it now, it won't happen ever.

youngnick:
I agree that this should not be merged unless and until we have the whole team in agreement that this is what we are going to do for the long term.

danehans:
Setting as "Draft" as we work through [the revised implementation plan](https://docs.google.com/document/d/1kOrGqHXULYO6YLxGCOtVyjz_7mfcW7Yi1WXCpQz6_Qo/edit?usp=sharing).

PR #3504, opened 2021-03-19
internal/xdscache/v3: Sort dag Routes before converting to Envoy Routes

sunjayBhatia:
- This change is currently a no-op but in the future when we add more
path match types it will become important
- Extracted as its own change since it is relatively large and will
complicate reviewing other PRs
- When we add path segment prefix matching (in addition to the string
prefix matching we do today) for Ingress v1 this will become important
  - This new type of Prefix matching has to be implemented as a regex
  but should still sort after user requested regex matches
- Also adds sorting of Exact path matches which should come before all
other path match types
- Adds constants for the various header match types allowed
- Some cons: this change means we end up iterating the routes on a vhost twice each

### DIFF
--- a/internal/xdscache/v3/route.go
+++ b/internal/xdscache/v3/route.go
@@ -138,27 +138,27 @@ func (v *routeVisitor) onVirtualHost(vh *dag.VirtualHost) {
 				Match:  envoy_v3.RouteMatch(route),
 				Action: envoy_v3.UpgradeHTTPS(),
 			}
-		} else {
-			rt := &envoy_route_v3.Route{
-				Match:  envoy_v3.RouteMatch(route),
-				Action: envoy_v3.RouteRoute(route),
-			}
-			if route.RequestHeadersPolicy != nil {
-				rt.RequestHeadersToAdd = envoy_v3.HeaderValueList(route.RequestHeadersPolicy.Set, false)
-				rt.RequestHeadersToRemove = route.RequestHeadersPolicy.Remove
-			}
-			if route.ResponseHeadersPolicy != nil {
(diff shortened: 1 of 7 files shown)


sunjayBhatia:
TODO: add test cases and logic for when match is inverted

This was covered before with deterministic sorting b/c we were doing proto string versions of the object comparisons

internal/xdscache/v3/route.go
```diff
@@ -303,12 +318,12 @@ func (v *routeVisitor) visit(vertex dag.Vertex) {
...
 
 // sortRoutes sorts the given Route slice in place. Routes are ordered
-// first by longest prefix (or regex), then by the length of the
-// HeaderMatch slice (if any). The HeaderMatch slice is also ordered
-// by the matching header name.
-func sortRoutes(routes []*envoy_route_v3.Route) {
+// first by path match type, path match value via string comparison and
+// then by the length of the HeaderMatch slice (if any). The HeaderMatch
+// slice is also ordered by the matching header name.
+func sortRoutes(routes []*dag.Route) {
```
skriss, line 324: not sure where the best place is, but it'd be good to add a comment about *why* we sort the DAG routes here vs. the Envoy ones. That may make more sense with the subsequent PR, though.

skriss (commented):
Otherwise I think this looks good!

This reminded me that I would like to do some refactoring in the route xds cache to remove some duplication, but that's separate from this PR and I'll file another issue for it.

stevesloka (approved):

internal/xdscache/v3/route.go
sunjayBhatia: Added a comment, LMK what you think!

skriss: :+1: I think that's helpful

skriss (approved):

PR #3505, opened 2021-03-21
Add default gatewayconfig

nak3:
https://github.com/projectcontour/contour/pull/3481 makes gatewayconfig optional so users always need to configure gatewayconfig. 
And so Gateway APIs stops working out of the box.

So this patch adds the default gatewayconfig.

### DIFF
--- a/pkg/config/parameters.go
+++ b/pkg/config/parameters.go
@@ -672,6 +672,10 @@ func Defaults() Parameters {
 		Listener: ListenerParameters{
 			ConnectionBalancer: "",
 		},
+		GatewayConfig: &GatewayParameters{
+			Name:      "contour",
+			Namespace: "projectcontour",
+		},
 	}
 }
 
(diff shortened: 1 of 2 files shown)


nak3:
@stevesloka Please feel free to close if adding the default was wrong approach. 
@danehans FYI: contour-opeartor's CI keep failing due to this issue https://github.com/projectcontour/contour-operator/actions/workflows/prbuild.yaml

danehans:
@nak3 https://github.com/projectcontour/contour-operator/pull/248 is a PR that's been waiting to get merged which. includes this fix and also a fix for regression introduced by https://github.com/projectcontour/contour-operator/commit/1410e5952edb1526664319051955a376330657bd

stevesloka:
We moved this config to be optional to have a key to know if GatewayAPI is installed. We need a way to know if we're going to use the listeners defined in the Gateway or the listeners configured as `contour serve --flags`. 

It will be up to the user/operator to configure Contour with this configuration in the config file to be that hook along with having the gateway-api crds installed in the cluster.

youngnick:
Yes, thanks for this PR @nak3, but I think we'll have to decline this time. The operator's support for configuring this in projectcontour/contour-operator#248 should cover what you need.

nak3:
Sure, no problem at all. Let's close this PR. Thank you for the comments.

PR #3506, opened 2021-03-22
internal: Allow listeners to be created dynamically if provided otherwise defaulted

stevesloka:
Since the GatewayAPI Gateway needs to be the driver of the ports/addresses that get exposed, we need a way to push that through to Envoy. At the same time, if there are multiple listeners, need a way to provision that to Envoy. 

This sets up the DAG to accept multiple listeners vs the current static `insecure` & `secure`.

Updates #2287

### DIFF
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -92,7 +92,7 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 		Spec: gatewayapi_v1alpha1.GatewaySpec{
 			Listeners: []gatewayapi_v1alpha1.Listener{{
 				Port:     80,
-				Protocol: "HTTP",
+				Protocol: gatewayapi_v1alpha1.HTTPProtocolType,
 				Routes: gatewayapi_v1alpha1.RouteBindingSelector{
 					Kind: KindHTTPRoute,
 					Namespaces: gatewayapi_v1alpha1.RouteNamespaces{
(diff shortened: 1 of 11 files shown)


youngnick (changes requested):
This PR seems fine, but I don't understand where it's going. Why do things in this way? It seems like we probably want to be able to change the names of the created listeners somehow for Gateway APIs support, but I'd like to see an explanation of all that in the commit log and PR description so that once this is merged, we see from the history where things are at and where they are going.

As it stands, all I can say when reviewing this one is "yes, that looks like it will allow dynamic creation of listeners with an arbitrary string name", but I have no context on how that plays into the larger Gateway APIs effort.

stevesloka:
If the GatewayAPI Gateway needs to be the driver of the ports/addresses that get exposed, we need a way to push that through to Envoy. At the same time, if there are multiple listeners, need a way to provision that to Envoy. 

This sets up the DAG to accept multiple listeners vs the current static `insecure` & `secure`.

youngnick (approved):
Perfect, that's exactly what I needed, thanks.

internal/dag/builder_test.go
```diff
@@ -188,7 +188,7 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 		Spec: gatewayapi_v1alpha1.GatewaySpec{
 			Listeners: []gatewayapi_v1alpha1.Listener{{
 				Port:     80,
-				Protocol: "HTTP",
+				Protocol: "ingress_http",
```
sunjayBhatia, line 191: Is this supposed to change? Seems like it should stay `HTTP`? Could we use a `const` from the gateway API package? https://github.com/kubernetes-sigs/gateway-api/blob/1caea70dce2571124a7a4887d32076ff8aa166e2/apis/v1alpha1/gateway_types.go#L255

stevesloka: I can switch to those, I think this a find/replace error.

internal/dag/builder_test.go
```diff
@@ -217,7 +217,7 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
...
```
sunjayBhatia, line 220: Couple more in this section

I guess we're not yet validating Listener.Protocol which would help catch this in unit tests

sunjayBhatia (approved):
LGTM, just one more find/replace to fix

internal/dag/builder_test.go
stevesloka: This would normally be caught by the api validation. But since we're skipping that it's missed. Could break this out into helper functions, but until we actually parse this field, it's not validated.

PR #3508, opened 2021-03-23
Test PR to check if bad fmt fails CI

sunjayBhatia:


sunjayBhatia:
Ok it fails as I would have expected, closing

Issue #2696, opened 2020-07-17
EndpointSlice Support


andrewsykim:
EndpointSlice is now enabled by default in Kubernetes and offers some solid performance benefits for clusters with large Endpoint sets. It might be worthwhile to support consuming EndpointSlice, at least as optional.

youngnick:
Thanks for raising this @andrewsykim, totally agree that we should support consuming EndpointSlice in some fashion.

zianke:
Hi Team. Agree that Contour should support EndpointSlice in the future. I just discussed with @stevesloka and would like to start working on an experimental implementation. Will send you an update once it's done.

stevesloka:
@zianke has a branch which implements endpoint slices (https://github.com/projectcontour/contour/compare/master...zianke:endpointslice) which is fantastic! 🎉  

Before pushing up a PR to implement to better help shape the PR, when would we want to add support for EndpointSlices? It's still beta in Kubernetes v1.17, so we. can't fully move to it. Should probably add a feature flag to enable which would disable the current Endpoints implementation. 

After we get this merged we'd also need to build out some E2E tests which mock up the behavior.

youngnick:
Great work on getting something done so quickly @zianke!

However, given that we're working on the Endpoint translator at the moment, this may be something that will need refactoring if we change anything about EndpointTranslator or ClusterLoadAssignment.

I'd also like to do our best not to add feature flags for enabling Kinds - in particular, EndpointSlice is designed so that if it's available, you can use either it or Endpoints, relatively transparently. I'd prefer to see detection of enabled Kinds used rather than a feature flag.

dprotaso:
Hey following up what's the status of EndpointSlices? and does Contour currently support [External Service Routing](https://projectcontour.io/docs/v1.18.0/config/external-service-routing/#external-service-routing) with [EndpointSlices with the CNAME address type](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#address-types)?

youngnick:
Thinking more about this, I think that we should confirm what K8s version EndpointSlice is GA in, and if that's less than our minimum supported (which it almost certainly is), we should move to using EndpointSlices exclusively (since Kube will ensure that changes are propagated as necessary back to Endpoints).

[...]

PR #3510, opened 2021-03-23
migrate to controller-runtime controllers for k8s watch events

stevesloka:
This replaces the internal informers and moves to the controller-runtime controllers
setting up Contour to migrate over to the controller-runtime caches.

Updates #2683


youngnick:
Thanks for doing this Steve. It looks much more straightforward than I feared.

How does this fit into the overall "migrate everything to controller-runtime" plan? What's left to do after this?

stevesloka:
> How does this fit into the overall "migrate everything to controller-runtime" plan? What's left to do after this?

This essentially just replaces the informers with controller-runtime. To use the runtime caches, we'd need to totally swap out `cache.go` and figure out how to do the new filtering bits since there's a lot of logic there now that would need to be done somewhere else in some way. That's the more difficult part and would need a bunch of thought as to how to approach that problem.

danehans:
@stevesloka needs a rebase.

cmd/contour/serve.go
```diff
@@ -419,6 +424,12 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	// the output of the other processors.
 	dagProcessors = append(dagProcessors, &dag.ListenerProcessor{})
 
+	// Setup a Manager
+	mgr, err := manager.New(controller_config.GetConfigOrDie(), manager.Options{})
+	if err != nil {
+		log.Fatal(err, "unable to set up overall controller manager")
```
danehans, line 430: I suggest dropping"overall", maybe something like:
"unable to create contour manager" or "unable to create contour server manager".

internal/dag/cache.go
```diff
@@ -36,6 +36,7 @@ import (
 // A KubernetesCache holds Kubernetes objects and associated configuration and produces
 // DAG values.
 type KubernetesCache struct {
+
```
danehans, line 39: Is this extra space needed?

internal/k8s/cache/backendpolicy.go
```diff
@@ -0,0 +1,75 @@
...
+	if err := c.Watch(&source.Kind{Type: &gatewayapi_v1alpha1.BackendPolicy{}}, &handler.EnqueueRequestForObject{}); err != nil {
+		return nil, err
+	}
+	return c, nil
+}
+
+func (r *backendPolicyReconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+
+	// Fetch the BackendPolicy from the cache.
+	gateway := &gatewayapi_v1alpha1.BackendPolicy{}
```
danehans, line 58: I find this var name misleading, maybe something like `policy` or `be`.

internal/k8s/cache/endpoint.go
```diff
@@ -0,0 +1,75 @@
...
+	if err := c.Watch(&source.Kind{Type: &corev1.Endpoints{}}, &handler.EnqueueRequestForObject{}); err != nil {
+		return nil, err
+	}
+	return c, nil
+}
+
+func (r *endpointReconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+
+	// Fetch the Service from the cache.
+	svc := &corev1.Endpoints{}
```
danehans, line 58: I find this var name misleading. Should this be `ep` or `endpoint`?

internal/k8s/cache/gateway.go
```diff
@@ -0,0 +1,75 @@
...
+	r := &gatewayReconciler{
+		client:       mgr.GetClient(),
+		eventHandler: eventHandler,
+		FieldLogger:  log,
+	}
+	c, err := controller.New("gateway-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return nil, err
+	}
+	if err := c.Watch(&source.Kind{Type: &gatewayapi_v1alpha1.Gateway{}}, &handler.EnqueueRequestForObject{}); err != nil {
```
danehans, line 49: Should gateway resources be filtered using a predicate? For example, only enqueue gateways that match Contour's [gateway config](https://github.com/projectcontour/contour/blob/main/examples/contour/01-contour-config.yaml#L14-L17) and reference a GatewayClass owned by contour or contour-operator?

internal/k8s/cache/httproute.go
```diff
@@ -0,0 +1,75 @@
...
+	r := &httpRouteReconciler{
+		client:       mgr.GetClient(),
+		eventHandler: eventHandler,
+		FieldLogger:  log,
+	}
+	c, err := controller.New("httproute-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return nil, err
+	}
+	if err := c.Watch(&source.Kind{Type: &gatewayapi_v1alpha1.HTTPRoute{}}, &handler.EnqueueRequestForObject{}); err != nil {
```
danehans, line 49: Similar to https://github.com/projectcontour/contour/pull/3510/files#r605974519, should routes be filtered based on the gateway config they specify matching Contour's gateway config?

xref: https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.RouteGateways

internal/k8s/cache/namespace.go
```diff
@@ -0,0 +1,75 @@
...
+	if err := c.Watch(&source.Kind{Type: &core_v1.Namespace{}}, &handler.EnqueueRequestForObject{}); err != nil {
+		return nil, err
+	}
+	return c, nil
+}
+
+func (r *namespaceReconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+
+	// Fetch the Namespace from the cache.
+	ingress := &core_v1.Namespace{}
```
danehans, line 58: `ns` var name?

internal/k8s/cache/tlsroute.go
```diff
@@ -0,0 +1,75 @@
...
+	if err := c.Watch(&source.Kind{Type: &gatewayapi_v1alpha1.TLSRoute{}}, &handler.EnqueueRequestForObject{}); err != nil {
+		return nil, err
+	}
+	return c, nil
+}
+
+func (r *tlsRouteReconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+
+	// Fetch the TLSRoute from the cache.
+	gateway := &gatewayapi_v1alpha1.TLSRoute{}
```
danehans, line 58: var name `tls`, `tr` or `route`.

danehans (changes requested):
I provided a few bits of feedback throughout.

cmd/contour/serve.go
stevesloka: I dropped "overall".

internal/dag/cache.go
stevesloka: Nope, just reads better with it personally. I'll take out.

internal/k8s/cache/backendpolicy.go
stevesloka: I'm the best at copy/paste. =)

internal/k8s/cache/endpoint.go
stevesloka: 👍

internal/k8s/cache/namespace.go
stevesloka: 👍

internal/k8s/cache/tlsroute.go
stevesloka: 👍

cmd/contour/serve.go
```diff
@@ -375,36 +386,64 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+		Logger: log.WithField("context", "dynamicHandler"),
 	}
 
-	// Inform on DefaultResources.
-	for _, r := range k8s.DefaultResources() {
-		inf, err := clients.InformerForResource(r)
-		if err != nil {
-			log.WithError(err).WithField("resource", r).Fatal("failed to create informer")
-		}
+	err = contour_api_v1.AddToScheme(mgr.GetScheme())
```
youngnick, line 397: It's a bit disappointing to end up moving back to doing everything individually, rather than being able to use a loop, I'd really like us to be making `doServe` smaller.

Maybe this is a good candidate for pulling out into a helper function? Feels like we could eliminate a bunch of boilerplate here.

internal/k8s/cache/endpoint.go
```diff
@@ -0,0 +1,75 @@
...
+
+type endpointReconciler struct {
+	client       client.Client
+	eventHandler cache.ResourceEventHandler
+	logrus.FieldLogger
+}
+
+// NewEndpointController creates the endpoint controller from mgr. The controller will be pre-configured
+// to watch for Endpoint objects across all namespaces.
+func NewEndpointController(mgr manager.Manager, eventHandler cache.ResourceEventHandler, log logrus.FieldLogger) (controller.Controller, error) {
```
youngnick, line 39: Tangentially related: We're also going to have to figure out what to do about EndpointSlice at some point, since Endpoints will go away (I think) in the future.

internal/k8s/cache/gateway.go
youngnick: I'd say "and/or" since the GatewayClass is not required, but aside from that I agree.

internal/k8s/converter.go
```diff
@@ -37,34 +37,14 @@ type DynamicClientHandler struct {
 }
 
 func (d *DynamicClientHandler) OnAdd(obj interface{}) {
-	obj, err := d.Converter.FromUnstructured(obj)
```
youngnick, line 40: Is it guaranteed that we'll never get Unstructured? If so, neat.

youngnick (approved):
I'm not a fan of this much boilerplate, but it seems like it's a necessary step on getting us closer to using controller-runtime fully, including caches.

A small non-blocking suggestion about keeping `doServe` small as well.

Thanks @stevesloka for getting this done.

cmd/contour/serve.go
stevesloka: There actually is already a function to do this, so I swapped in the scheme that comes from the controller-runtime and can remove this. =)

internal/k8s/cache/endpoint.go
stevesloka: Yup I agree: https://github.com/projectcontour/contour/issues/2696

internal/k8s/converter.go
stevesloka: Yeah we shouldn't since the controller-runtime is doing all of this for us now.

cmd/contour/serve.go
```diff
@@ -375,36 +387,64 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+		Logger: log.WithField("context", "dynamicHandler"),
 	}
 
-	// Inform on DefaultResources.
-	for _, r := range k8s.DefaultResources() {
-		inf, err := clients.InformerForResource(r)
-		if err != nil {
-			log.WithError(err).WithField("resource", r).Fatal("failed to create informer")
-		}
+	//err = contour_api_v1.AddToScheme(mgr.GetScheme())
```
sunjayBhatia, line 398: Some extra content here?

sunjayBhatia (commented):
I think the cmd/contour/ingressstatus.go file also sets up some informers so it should probably be updated as well to use the same pattern

I guess I would echo the sentiments about trying to deduplicate some of the repeated bits in the controller constructors/reconcile methods, if I need to make a change it would be great to do it one place rather than many since its pretty much the same code

maybe a constructor that takes a type and for any slight differences between the controllers for different types there could be an optional mutator function or something? not sure how much would be different between different controllers other than different predicates/filters

cmd/contour/serve.go
```diff
@@ -231,7 +243,7 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 
 	// Before we can build the event handler, we need to initialize the converter we'll
 	// use to convert from Unstructured.
-	converter, err := k8s.NewUnstructuredConverter()
+	converter, err := k8s.NewUnstructuredConverter(mgr.GetScheme())
```
danehans, line 246: Does it make more sense to create the scheme with the gvk's that we care about and pass the scheme in as a `manager.Option`?

skriss (commented):
@stevesloka could you provide some additional context on what the end state for this looks like? I'm having a little trouble giving feedback on approach since I'm not sure where we're trying to get to. My understanding is that we're already using some controller-runtime libraries for the informers (see https://github.com/projectcontour/contour/blob/main/internal/k8s/clients.go#L36 and https://github.com/projectcontour/contour/blob/main/internal/k8s/clients.go#L95). What is the role of these reconcilers going to be in the target state? Is that going to change how the DAG is processed?

stevesloka:
The goal of this change was to allow us to manage separate controllers for different objects, mainly the gatewayapi work. We have cleaner hooks now to work on changes to gatewayclasses, gateways, etc which may not be DAG relevant (i.e. a new gateway might cause an envoy service to be created, etc). 

There has been discussion of removing the internal k8s cache and moving to a controller-runtime cache, but that needs some more thoughts as to how to approach since we filter on the k8s cache so much today.

cmd/contour/serve.go
stevesloka: Yeah that's sort of what was happening here: https://github.com/projectcontour/contour/blob/4885db6354ec9cbb92ef183e2c02e71c9aee9cbd/internal/k8s/scheme.go#L27

cmd/contour/serve.go
stevesloka: 👍

stevesloka:
> I guess I would echo the sentiments about trying to deduplicate some of the repeated bits in the controller constructors/reconcile methods, if I need to make a change it would be great to do it one place rather than many since its pretty much the same code

@sunjayBhatia yup I agree that there are duplicated bits at the moment, but I think when @danehans starts to add some of the "operator" logic that might make more sense. We've overloaded the cahce.go file to do so much, splitting out more I think would be helpful. 

Things we could do would be to move the secret validation out, the logic to know if a service is referenced from an ingress object, etc. That way we'd have a cleaner approach to look at when making changes and the k8s cache turns into just a cache, not a cache w/logic.

skriss:
> The goal of this change was to allow us to manage separate controllers for different objects, mainly the gatewayapi work. We have cleaner hooks now to work on changes to gatewayclasses, gateways, etc which may not be DAG relevant (i.e. a new gateway might cause an envoy service to be created, etc).

So is this primarily useful only if we decide to move the gatewayclass/gateway controllers into Contour? If so, I'm wondering if it'd make sense to keep this in the experimental branch that you/@danehans are going to be working on for now until we make a decision on path forward there. Otherwise, this seems like a fair amount of overhead to add to Contour if all the controllers are doing is passing events through to the event handler since the informers are capable of doing that today. Maybe we can spend some time talking through this one in the maintainers meeting.

stevesloka:
Aside from GatewayAPI work, this allows us to further clean up cache.go file to not do so much, splitting out into different places which would also be helpful.

We do need to discuss if we do want to move to controller-runtime, doing that would be best done in steps vs a drastic change. 

I think fine to keep experimental until we decide the above steps.

youngnick:
To reiterate what I said late last week: I think that we should go ahead with this regardless of what else we end up doing. It doesn't add much value by itself right now, but it sets us up well for further work to move us more towards the controller-runtime `Reconcile` pattern, rather than the `OnAdd`, `OnUpdate`, `OnDelete` Informer pattern we have now.

danehans:
@stevesloka this PR is not hitting https://github.com/kubernetes-sigs/controller-runtime/issues/1473 due to the gateway api bump:
```
$ make checkall
go install -mod=readonly -v -ldflags="-s -w -X github.com/projectcontour/contour/internal/build.Version=81b067b7 -X github.com/projectcontour/contour/internal/build.Sha=81b067b74e0c0252455656665e9ffd03832c0255 -X github.com/projectcontour/contour/internal/build.Branch=gw_api_support " -tags "oidc gcp osusergo netgo" github.com/projectcontour/contour/cmd/contour
sigs.k8s.io/controller-runtime/pkg/metrics
# sigs.k8s.io/controller-runtime/pkg/metrics
../../../../pkg/mod/sigs.k8s.io/controller-runtime@v0.8.3/pkg/metrics/client_go_adapter.go:134:3: cannot use &latencyAdapter literal (type *latencyAdapter) as type metrics.LatencyMetric in field value:
	*latencyAdapter does not implement metrics.LatencyMetric (wrong type for Observe method)
		have Observe(string, url.URL, time.Duration)
		want Observe(context.Context, string, url.URL, time.Duration)
../../../../pkg/mod/sigs.k8s.io/controller-runtime@v0.8.3/pkg/metrics/client_go_adapter.go:135:3: cannot use &resultAdapter literal (type *resultAdapter) as type metrics.ResultMetric in field value:
	*resultAdapter does not implement metrics.ResultMetric (wrong type for Increment method)
		have Increment(string, string, string)
		want Increment(context.Context, string, string, string)
make: *** [install] Error 2
```

@sunjayBhatia bumped controller-runtime to `sigs.k8s.io/controller-runtime v0.9.0-beta.0` for the operator, but we may not want to do the same for Contour.

stevesloka:
@danehans yup I don't see a need to bump the dep of controller-runtime just now, we can hang with the current as-is for now until resolved upstream.

danehans:
> @danehans yup I don't see a need to bump the dep of controller-runtime just now, we can hang with the current as-is for now until resolved upstream.

@stevesloka the current version (v0.8.3) fails to compile, see https://github.com/projectcontour/contour/pull/3510#issuecomment-832132678.

stevesloka:
@danehans ahh let me look, I didn't rebase :grin:

stevesloka:
So  given we've got controller-runtime dep issues, just chatted with @danehans and have this approach if everyone is ok with. 

1.  Update this PR to implement the `v0.9.0-beta.1` release (which is latest)
2. Implement the controller-runtime to only work off of GatewayAPI resources

This will reduce the impact of any users having issues with the beta release of controller-runtime giving us time to move to that once a stable version is out. At that point we can move more of the informer bits over to controller-runtime as we see fit.

danehans:
@stevesloka merge conflict due to https://github.com/projectcontour/contour/pull/3638

stevesloka:
rebased @danehans 🚀

danehans:
@stevesloka the updated PR no longer starts the controller-runtime manager. See [this](https://github.com/danehans/contour/commit/ed4be046ad6b3223dbb2a88ad169c3e34282cfd7) for the fix.

youngnick (approved):
This still LGTM, I'd like to get this one in so we can unblock further Gateway API work.

cmd/contour/serve.go
```diff
@@ -415,10 +424,43 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+			// Setup a Manager
+			mgr, err := manager.New(controller_config.GetConfigOrDie(), manager.Options{})
+			if err != nil {
+				log.Fatal(err, "unable to set up controller manager")
+			}
+
+			err = gatewayapi_v1alpha1.AddToScheme(mgr.GetScheme())
+			if err != nil {
+				log.Error(err, "unable to add GatewayAPI to scheme.")
+				os.Exit(1)
```
skriss, line 437:
```suggestion
				log.WithError(err).Fatal("unable to add GatewayAPI to scheme.")
```

cmd/contour/serve.go
```diff
@@ -415,10 +424,43 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 					log.WithError(err).WithField("resource", r).Fatal("failed to create informer")
 				}
 			}
+
+			// Setup a Manager
+			mgr, err := manager.New(controller_config.GetConfigOrDie(), manager.Options{})
+			if err != nil {
+				log.Fatal(err, "unable to set up controller manager")
```
skriss, line 431:
```suggestion
				log.WithError(err).Fatal("unable to set up controller manager")
```

internal/k8s/cache/gateway.go
```diff
@@ -0,0 +1,75 @@
+// Copyright Project Contour Authors
```
skriss, line 1: I think I'd put these into an `internal/reconciler` or `internal/controller` package, I think that reflects what they do better.

internal/k8s/cache/gateway.go
```diff
@@ -0,0 +1,75 @@
...
+
+func (r *gatewayReconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+
+	// Fetch the Gateway from the cache.
+	gateway := &gatewayapi_v1alpha1.Gateway{}
+	err := r.client.Get(ctx, request.NamespacedName, gateway)
+	if errors.IsNotFound(err) {
+		r.Error(nil, "Could not find Gateway %q in Namespace %q", request.Name, request.Namespace)
+		return reconcile.Result{}, nil
+	}
```
skriss, line 63: should this case result in an `OnDelete` event?

internal/k8s/cache/gateway.go
```diff
@@ -0,0 +1,75 @@
...
+
+	// Check if object is deleted.
+	if !gateway.ObjectMeta.DeletionTimestamp.IsZero() {
+		r.eventHandler.OnDelete(gateway)
+		return reconcile.Result{}, nil
+	}
+
+	// Pass the new changed object off to the eventHandler.
+	r.eventHandler.OnAdd(gateway)
```
skriss, line 72: I guess `OnAdd` makes sense here since, regardless of whether this was actually a create or an update, this will have the effect of overwriting the object in the Contour cache. Maybe a brief comment to explain why `OnAdd` is fine whether it's an add or an update would be helpful.

skriss (changes requested):

internal/k8s/cache/gateway.go
stevesloka: Nope it shouldn't, this handles that the event fired but can't be found in the internal cache. I think once we get the finalizers bit it will mean more since the object can't be removed until those are dropped.

skriss: OK, yeah I agree on the finalizers. My assumption was that if an object (without finalizers) was deleted, then `Reconcile` would get triggered but since the object was deleted, it would potentially not be in the cache (i.e. this case). In that scenario, we'd want to make sure we removed the item from the Contour cache. But if everything's got finalizers, then the `deletionTimestamp` case should handle it.

stevesloka: I can add it to delete just to make sure, but would want to extend the delete method in the cache.go to check if the object exists first in the event that it got removed, but then this got called again for some reason.

skriss: Just looked in the kubebuilder book (ref. https://book.kubebuilder.io/reference/using-finalizers.html):

> The key point to note is that a finalizer causes “delete” on the object to become an “update” to set deletion timestamp. Presence of deletion timestamp on the object indicates that it is being deleted. **Otherwise, without finalizers, a delete shows up as a reconcile where the object is missing from the cache.**

skriss:
Looks like you hit #3660 in the E2E run (fixed by #3662). We can ignore since it's not a required check yet.

internal/k8s/cache/gateway.go
stevesloka: ...or maybe should remove the timestamp checks since we don't have finalizers?

internal/reconciler/gateway.go
```diff
@@ -0,0 +1,69 @@
...
+	return c, nil
+}
+
+func (r *gatewayReconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+
+	// Fetch the Gateway from the cache.
+	gateway := &gatewayapi_v1alpha1.Gateway{}
+	err := r.client.Get(ctx, request.NamespacedName, gateway)
+	if errors.IsNotFound(err) {
+		r.eventHandler.OnDelete(gateway)
```
skriss, line 61: I *think* you'll need to explicitly set `gateway`'s name and namespace from `request` here, since if the `Get` failed, it likely won't be populated with anything.

stevesloka: +1 Let me write some tests around this too, should have been there from the start to find these things.

internal/reconciler/gateway.go
```diff
@@ -0,0 +1,69 @@
...
+
+type gatewayReconciler struct {
+	client       client.Client
+	eventHandler cache.ResourceEventHandler
+	logrus.FieldLogger
+}
+
+// NewGatewayController creates the gateway controller from mgr. The controller will be pre-configured
+// to watch for Gateway objects across all namespaces.
+func NewGatewayController(mgr manager.Manager, eventHandler cache.ResourceEventHandler, log logrus.FieldLogger) (controller.Controller, error) {
```
sunjayBhatia, line 39: looks like 90% of code between the controller files is the same, seems like a future refactor could be to have a controller factory that creates a resource specific controller given a type and a few other fields so we don't have to duplicate?

do we foresee needing different logic for different resources?

stevesloka: Yup I think we will for sure need different logic. We'll need to manage finalizers as well as Envoy instances, etc.

internal/reconciler/gateway.go
stevesloka: I think #3659 should bring in a bunch of the test infra. I don't want to recreate that here if that's cool with folks.

cmd/contour/serve.go
```diff
@@ -399,6 +405,9 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		}
 	}
 
+	// Set up workgroup runner and register informers.
+	var g workgroup.Group
+
 	// Only inform on GatewayAPI resources if Gateway API is found.
 	if ctx.Config.GatewayConfig != nil {
 		if clients.ResourcesExist(k8s.GatewayAPIResources()...) {
```
skriss, line 413: we're still setting up the informers for Gateway API resources too, is that intended?

skriss (approved):
LGTM pending CI.

PR #3511, opened 2021-03-23
internal/dag: Only process Gateway.Listener.Selector.Kind of HTTPRoute

stevesloka:
Only process currently supported types of HTTPRoute on the Gateway.Listener.Selector.Kind field.

Updates #3401

### DIFF
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -74,9 +74,9 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
 		for _, listener := range p.source.gateway.Spec.Listeners {
 
 			// Validate the Group on the selector is a supported type.
-			if listener.Routes.Group != "" && listener.Routes.Group != "networking.x-k8s.io" {
+			if listener.Routes.Group != "" && listener.Routes.Group != gatewayapi_v1alpha1.GroupName {
 				// TODO: Set the “ResolvedRefs” condition to false for this listener with the “InvalidRoutesRef” reason.
-				p.Errorf("Listener.Routes.Group %q is not supported.", listener.Routes.Kind)
+				p.Errorf("Listener.Routes.Group %q is not supported.", listener.Routes.Group)
 				continue
 			}
 
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -27,6 +27,10 @@ import (
 	gatewayapi_v1alpha1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
 )
 
+const (
+	KindHTTPRoute = "HTTPRoute"
+)
+
 // GatewayAPIProcessor translates Gateway API types into DAG
 // objects and adds them to the DAG.
 type GatewayAPIProcessor struct {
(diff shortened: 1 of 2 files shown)


internal/dag/gatewayapi_processor.go
```diff
@@ -69,20 +73,27 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
 		// are associated with the Gateway. An empty Selector matches all routes.
 		for _, listener := range p.source.gateway.Spec.Listeners {
 
-			nsMatches, err := p.namespaceMatches(listener.Routes.Namespaces, route)
-			if err != nil {
-				p.Errorf("error validating namespaces against Listener.Routes.Namespaces: %s", err)
-			}
+			// Validate the Kind on the selector is a supported type.
+			switch listener.Routes.Kind {
```
skriss, line 77: just looking at the docs (https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.RouteBindingSelector), we should probably check `listener.Routes.Group` as well - sounds like empty string or `networking.x-k8s.io` would be valid here, anything else would not be valid.

stevesloka: Yup sounds good! 👍

internal/dag/gatewayapi_processor.go
```diff
@@ -69,6 +73,19 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
 		// are associated with the Gateway. An empty Selector matches all routes.
 		for _, listener := range p.source.gateway.Spec.Listeners {
 
+			// Validate the Group on the selector is a supported type.
+			if listener.Routes.Group != "" && listener.Routes.Group != "networking.x-k8s.io" {
+				// TODO: Set the “ResolvedRefs” condition to false for this listener with the “InvalidRoutesRef” reason.
+				p.Errorf("Listener.Routes.Group %q is not supported.", listener.Routes.Kind)
```
skriss, line 79:
```suggestion
				p.Errorf("Listener.Routes.Group %q is not supported.", listener.Routes.Group)
```

internal/dag/gatewayapi_processor.go
```diff
@@ -69,6 +73,19 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
 		// are associated with the Gateway. An empty Selector matches all routes.
 		for _, listener := range p.source.gateway.Spec.Listeners {
 
+			// Validate the Group on the selector is a supported type.
+			if listener.Routes.Group != "" && listener.Routes.Group != "networking.x-k8s.io" {
```
skriss, line 77:
```suggestion
			if listener.Routes.Group != "" && listener.Routes.Group != gatewayapi_v1alpha1.GroupName {
```

skriss (commented):
two nits, otherwise LGTM

stevesloka:
I updated your two comments @skriss and also added a test for the group.

skriss (approved):

sunjayBhatia (commented):
Would it make sense to reverse the processing order and maybe chuck out the invalid listeners first before iterating through routes? Just so we don't have to re-examine an invalid listener when we process each route?

Otherwise, looks correct, just maybe a perf change

stevesloka:
> Would it make sense to reverse the processing order and maybe chuck out the invalid listeners first before iterating through routes? Just so we don't have to re-examine an invalid listener when we process each route?

Nope that's a great comment @sunjayBhatia! I'll do that! Also, we need to make sure that we're only processing routes that match the listener (which I don't think is happening today). 

One weird edge case that I've thought up is that since the selectors on a Listener don't need to match across HTTP & HTTPS listeners, then different routes can be selected by different ones.

sunjayBhatia (approved):

PR #3512, opened 2021-03-23
cmd/contour: extract getDAGBuilder helper function

skriss:
Extracts all logic in doServe() related to constructing
the dag.Builder into a helper function.

Updates #3453.

### DIFF
--- a/cmd/contour/serve_test.go
+++ b/cmd/contour/serve_test.go
@@ -1,0 +1,108 @@
+// Copyright Project Contour Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"testing"
+
+	"github.com/projectcontour/contour/internal/dag"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func TestGetDAGBuilder(t *testing.T) {
+	commonAssertions := func(t *testing.T, builder *dag.Builder) {
+		t.Helper()
+
+		// note that these first two assertions will not hold when a gateway
+		// is configured, but we don't currently have test cases that cover
+		// that so it's OK to keep them in the "common" assertions for now.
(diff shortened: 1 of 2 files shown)


skriss:
@projectcontour/maintainers I'm not 100% sure yet where this refactoring is going to end up, but as a starting point I'm trying to extract some atomic chunks of logic from the `doServe()` function into their own helpers, and to start to add some tests. Feel free to provide feedback on if you think there's a better approach.

youngnick (approved):
LGTM, anything we can do to make that functions smaller is good in my book.

sunjayBhatia (approved):
LGTM, nice start to getting more test coverage

Maybe something to consider for the future would be mutator methods on `dag.Builder`?
So we could do something like `dag.NewBuilder().WithFoo(...)`
Not sure if we like that pattern and have used it in Contour or want to just use a more structured `New` function instead so we can test in the `dag.Builder` tests?
Something needs to ensure `serve.go` is correctly wiring things up and this is a good start

cmd/contour/serve_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+		got := getDAGBuilder(newServeContext(), nil, clientCert, fallbackCert, logrus.StandardLogger())
+
+		assert.Len(t, got.Processors, 4)
+		assert.ElementsMatch(t, got.Source.ConfiguredSecretRefs, []*types.NamespacedName{clientCert, fallbackCert})
+		assert.Equal(t, types.NamespacedName{}, got.Source.ConfiguredGateway)
+		assert.IsType(t, &dag.ListenerProcessor{}, got.Processors[len(got.Processors)-1])
+	})
+
+	// TODO request/response headers
```
stevesloka, line 66: Are these tests still in progress or just futures @skriss ?

skriss: I didn't want to go too nuts before getting some 👀  on the change; let me see if I can cover a few more of these now, and I'll add an issue link in the TODO for anything I'm not gonna cover.

stevesloka: awesome, sounds good, just wanted to double check. =)

skriss: Heh I actually just found a bug in upstream Contour by adding some more test cases here, PR incoming

skriss: OK, added a couple more cases and added an issue link for the TODO.

skriss: @stevesloka LMK if you had any other comments, otherwise this should be good to merge.

sunjayBhatia:
Coverage numbers going 🆙  😍

skriss:
Eh, guess we can hold this one for 1.15 since it's just a refactor and that'll give it a little more time to bake in dev testing.

stevesloka (approved):
/lgtm

stevesloka:
@skriss looks like you need a rebase.

skriss:
> @skriss looks like you need a rebase.

Thanks, done!

> @skriss looks like you need a rebase.

PR #3517, opened 2021-03-25
pkg/certs: fix flaky test

### LINKED ISSUE #3516: certgen_test.go test flake
https://github.com/projectcontour/contour/runs/2194721987?check_suite_focus=true

I've seen this test flake a number of times. It looks like for validating custom cert lifetimes we're doing some "current time" manipulation (https://github.com/projectcontour/contour/blob/main/pkg/certs/certgen_test.go#L50-L53) and there's probably a race condition in there when the test is running slowly.

skriss:
Updates the custom lifetime cert test to use
a lifetime longer than the default, and to
verify the certs as of an hour before expiration
rather than 5 seconds, to avoid a race condition
when test execution is slow.

Closes #3516.

### DIFF
--- a/pkg/certs/certgen_test.go
+++ b/pkg/certs/certgen_test.go
@@ -47,9 +47,11 @@ func TestGenerateCerts(t *testing.T) {
 				t.Errorf("wanted no error, got error %q", err)
 			}
 
+			// If using a custom lifetime, validate the certs
+			// as of an hour before the intended expiration.
 			currentTime := time.Now()
 			if tc.config.Lifetime != 0 {
-				currentTime = time.Now().Add(24 * time.Duration(tc.config.Lifetime) * time.Hour).Add(-5 * time.Second)
+				currentTime = currentTime.Add(24 * time.Hour * time.Duration(tc.config.Lifetime)).Add(-time.Hour)
 			}
 
 			roots := x509.NewCertPool()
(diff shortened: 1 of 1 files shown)


pkg/certs/certgen_test.go
```diff
@@ -47,9 +47,11 @@ func TestGenerateCerts(t *testing.T) {
 				t.Errorf("wanted no error, got error %q", err)
 			}
 
+			// If using a custom lifetime, validate the certs
+			// as of an hour before the intended expiration.
 			currentTime := time.Now()
 			if tc.config.Lifetime != 0 {
-				currentTime = time.Now().Add(24 * time.Duration(tc.config.Lifetime) * time.Hour).Add(-5 * time.Second)
+				currentTime = time.Now().Add(24 * time.Hour * time.Duration(tc.config.Lifetime)).Add(-time.Hour)
```
sunjayBhatia, line 54:
```suggestion
				currentTime = currentTime.Add(24 * time.Hour * time.Duration(tc.config.Lifetime)).Add(-time.Hour)
```

nit: Not sure if this makes a real difference, at least it makes it so we don't fetch the now twice?

sunjayBhatia (approved):
LGTM! One tiny comment, not sure if it even really matters

pkg/certs/certgen_test.go
skriss: doesn't hurt, done

youngnick (approved):
LGTM, nice pickup.

stevesloka (approved):

PR #3519, opened 2021-03-26
cmd/contour: properly configure request/response headers policy

skriss:
Fixes two issues with configuring request/response headers policy.
The first issue was that the user's configured request headers to
set were inadvertently also being used as the response headers to
set. The second issue was that some harmless empty strings were
being added to the lists of request/response headers to remove.


skriss:
Per https://github.com/projectcontour/contour/pull/3512#discussion_r602442658 I'll have a test case covering this in that PR.

youngnick (approved):
Oof, that's a trap for new players. LGTM.

sunjayBhatia (approved):

PR #3520, opened 2021-03-26
Ingress: Filter Ingress resources based on Spec.IngressClassName

sunjayBhatia:
Rules are as follows:
- If Contour is not configured with an ingress class name
  - Cache IngressClass named "contour" if exists (not actually used)
  - Accept Ingress objects with no class name in annotation or spec
  - Accept Ingress objects with class name "contour" in annotation
  - Accept Ingress objects with no class name in annotation and "contour" in spec
  - If accepted, update the Address in the status of the Ingress resource
- If Contour is configured with an ingress class name (e.g. "foo")
  - Cache IngressClass named "foo" if exists (not actually used)
  - Accept Ingress objects with class name "foo" in annotation
  - Accept Ingress objects with no class name in annotation and "foo" in spec
  - If accepted, update the Address in the status of the Ingress resource
- If an IngressClass is annotated as "default" the admission controller should add that class name to the spec of any Ingresses created without a class
- If multiple IngressClasses are annotated as "default" admission controller does not allow new Ingresses to be created unless they specify an ingress class in the spec

Fixes https://github.com/projectcontour/contour/issues/2146


sunjayBhatia:
Need to do a conformance check on this one but should be good to review

youngnick (approved):
LGTM, nice work @sunjayBhatia.

internal/dag/cache.go
```diff
@@ -84,24 +84,89 @@ func (kc *KubernetesCache) init() {
...
 
-// matchesIngressClass returns true if the given Kubernetes object
-// belongs to the Ingress class that this cache is using.
-func (kc *KubernetesCache) matchesIngressClass(obj metav1.Object) bool {
+// matchesIngressClass returns true if the given IngressClass
+// is the one this cache is using.
+func (kc *KubernetesCache) matchesIngressClass(obj *networking_v1.IngressClass) bool {
+	// If no ingress class name set, we allow an ingress class that is named
+	// with the default Contour accepted name.
+	if kc.IngressClassName == "" && obj.Name == "contour" {
```
sunjayBhatia, line 92: I wanted to pull the default ingress class name out of the `annotation` package but wasn't quite sure where to put it tbh

sunjayBhatia:
Right now we are not taking ingress class name into account when setting Ingress address in status, this is still needed to get conformance

internal/dag/cache.go
```diff
@@ -84,24 +84,89 @@ func (kc *KubernetesCache) init() {
...
+	case *networking_v1.Ingress:
+		specClass = pointer.StringPtrDerefOr(obj.Spec.IngressClassName, "")
+	}
+
+	// If annotation is set or annotation is not set and spec is not set, we
+	// will try to match the annotation.
+	checkAnnotation := annotationClass != "" || specClass == ""
+
+	// If the annotation matches, return true.
+	if checkAnnotation && annotation.MatchesIngressClass(obj, kc.IngressClassName) {
```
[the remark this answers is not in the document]
sunjayBhatia: Yeah, we are not using the content of any `IngressClass` resource at all today. According to this it is an "optional" way to provide config which could be interpreted in a few ways (currently we have interpreted it such that a resource does not have to exist with the name referenced by an `Ingress`). We added read access to the resource but could totally walk it back if we want since we're not actually using it, though users might be annoyed if we have to eventually add RBAC to `IngressClass` back again at some point.

If we start making the `IngressClass` resource have to exist it may be an unexpected change for users who would have to go create before they upgrade or their `Ingress` routes will be dropped. We could require this now, if/when we actually look at the content of an `IngressClass`, or say we will never look at the content and never require it (maybe we could only require it you're using the spec field and not the annotation to specify the class? Seems like this might create more annoying edge cases and unexpected behavior, idk).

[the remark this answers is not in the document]
skriss: I guess it wouldn't hurt to leave it as-is for now, and we can always add the existence requirement down the road, maybe once ingress v1beta1 is dropped?

internal/dag/cache.go
```diff
@@ -84,24 +84,89 @@ func (kc *KubernetesCache) init() {
...
+	switch obj := obj.(type) {
+	case *v1beta1.Ingress:
+		specClass = pointer.StringPtrDerefOr(obj.Spec.IngressClassName, "")
+	case *networking_v1.Ingress:
+		specClass = pointer.StringPtrDerefOr(obj.Spec.IngressClassName, "")
+	}
+
+	// If annotation is set or annotation is not set and spec is not set, we
+	// will try to match the annotation.
+	checkAnnotation := annotationClass != "" || specClass == ""
```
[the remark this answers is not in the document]
sunjayBhatia: I think I was trying to only do the annotation match check once for the multiple cases where it is relevant and I felt it collapsed some logic together. I think also I tried to fall through to return `false` and log the debug message in one place and that made this a bit more complicated. Another way to do this would be the following, though it would require some duplication and checking whether we had a match or not to log the debug message:

Doing things another way it might look like this:
- if annotation is non-empty, check if it matches
  - return result of annotation helper
- if spec field is non-empty check if it matches
  - return result of checking class
- return true if no class configured

internal/dag/cache.go
sunjayBhatia: I'm gonna have to pull some of the `IngressClass` matching logic out into something most likely, since I realized the `Ingress` status needs the same logic around status address updates

internal/dag/cache.go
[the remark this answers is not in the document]
skriss: Yeah to my eye that's a little easier to read since each case is handled directly. I see what you mean about the logging. You could even move the impl of getting the spec ingress class to a helper if you wanted (since we have one for the annotation, it keeps things at the same level of abstraction) but not a big deal.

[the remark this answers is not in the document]
sunjayBhatia: >  move the impl of getting the spec ingress class to a helper if you wanted

Yeah gonna have to do that since the status updater bits will need it as well

internal/dag/cache.go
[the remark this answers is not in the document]
youngnick: Even when it's only Ingress v1, the IngressClass is still optional, that's the API. I did not compute that until very recently either.

stevesloka:
> If Contour is configured with an ingress class name (e.g. "foo")
>  - Accept Ingress objects with no class name in annotation and "foo" in spec

I thought that if you configured an ingress class in Contour, then you had to match that class. If any resource was not defined with a class, then it was ignored.

skriss (commented):
this LGTM, just wasn't sure if you were going to do the Ingress address updating changes as part of this PR too or not

sunjayBhatia:
> this LGTM, just wasn't sure if you were going to do the Ingress address updating changes as part of this PR too or not

Ah yeah, replied in Slack as well but yeah I am almost done with the status address updating bit too

sunjayBhatia:
Confirmed we don't have to inform on /update Ingress v1beta1 explicitly, fetching v1 resources and updating them works just fine

Also, ingress conformance ingress class test passes with the status updater filtering

sunjayBhatia:
Hmm might have broken something about the n-2 integration tests

sunjayBhatia:
Failure in k8s 1.18:
```
time="2021-03-31T00:46:42Z" level=fatal msg="failed to create informer" context=loadBalancerStatusWriter error="no matches for networking.k8s.io/v1, Resource=ingresses" resource="networking.k8s.io/v1, Resource=ingresses"
```

sunjayBhatia:
> Failure in k8s 1.18:
> 
> ```
> time="2021-03-31T00:46:42Z" level=fatal msg="failed to create informer" context=loadBalancerStatusWriter error="no matches for networking.k8s.io/v1, Resource=ingresses" resource="networking.k8s.io/v1, Resource=ingresses"
> ```

should be fixed, only informing on Ingress v1 if it is present

internal/validation/ingress/ingress.go
```diff
@@ -0,0 +1,57 @@
...
+	switch o := obj.(type) {
+	case *networking_v1.Ingress:
+		specClass = pointer.StringPtrDerefOr(o.Spec.IngressClassName, "")
+	case *v1beta1.Ingress:
+		specClass = pointer.StringPtrDerefOr(o.Spec.IngressClassName, "")
+	}
+
+	// If annotation is set, check if it matches.
+	if annotationClass != "" {
+		return annotation.MatchesIngressClass(obj, ingressClassName)
```
skriss, line 43: WDYT about moving `annotation.MatchesIngressClass` into this package? I think it'd be nice to have all the ingress class matching logic in one place. I *think* it could even be unexported, and just let `MatchesIngressClassName` be the sole entrypoint for this package (but possible I'm overlooking something). 

I'm also very fine with doing this later, if we want to do it but it's gonna cause a bunch of churn right now while we're trying to get the release out.

skriss (approved):
LGTM, with one suggestion.

internal/validation/ingress/ingress.go
sunjayBhatia: Yeah, I considered moving that but I think it is also used to check the annotation on HTTPProxy resources as well since we support setting a class in an annotation on those too.

I'll add an issue to clean this up b/c I think it could def be structured better

skriss: Sounds good. Yeah it is used for `HTTPProxy` as well, but `MatchesIngressClassName` takes a `metav1.Object` which could be either an `Ingress` or an `HTTPProxy`, and we could type-switch as needed. Anyway we can discuss some more on the issue.

sunjayBhatia: https://github.com/projectcontour/contour/issues/3532

sunjayBhatia: Please comment and/or edit the issue if you think of some more ideas/context

PR #3521, opened 2021-03-26
internal/sorter: Sort Route prefix segment matches before string

sunjayBhatia:
Segment matches are more specific than string prefixes so they should
sort first.


skriss (approved):
LGTM!

youngnick (approved):
LGTM, nice work.

PR #3523, opened 2021-03-28
chore: remove description column of httpproxy

kushthedude:
Signed-off-by: kushthedude <kushthedude@gmail.com>

fixes https://github.com/projectcontour/contour/issues/2978


youngnick (approved):
LGTM, thanks @kushthedude.

stevesloka:
Is there a way to show all the errors in the field instead of the current static message as James suggested (
https://github.com/projectcontour/contour/issues/2978#issuecomment-703987557)? Would be nice to have a way to know what's wrong without having to describe each resource and dig into those details.

youngnick:
There's some shortcuts we can do (you can pick out the Reason or the Message of a single Condition, which would let us pick out the Reason and Message for the Valid Condition), but there doesn't seem to be a good way to concatenate lots of things in the custom column definition, so I think removing it for now is probably better.

stevesloka:
> but there doesn't seem to be a good way to concatenate lots of things in the custom column definition

Could we just do that work when updating the status? We're setting the value of that field as a static string, why not just update when processing?

youngnick:
Yes, we absolutely could do that. That's what I started in the initial Condition design, but I don't think it made it in in the end. In that case, I guess we would keep the column, as a rollup of all the Conditions.

kushthedude:
So should this PR be closed ? 
/cc @youngnick

youngnick:
Sorry @kushthedude, I think it's probably better, yeah. Let's see if we can fix up the field rather than removing it.

PR #3537, opened 2021-04-01
internal/dag: Fix GatewayAPI Namespace Selector All test

stevesloka:
Fixes the "insert basic single route, single hostname, gateway All namespace selector" test which
should return valid listeners given a HTTPRoute in any namespace.

The test was invalid since the service namespace in the forwardTo did not match the route's namespace.

Fixes #3535


skriss (approved):
good catch, LGTM!

sunjayBhatia (approved):

Issue #3356, opened 2021-02-11
Implement "Exact" HTTPHeaderMatch


stevesloka:
For an HTTPRoute, need to implement the "Exact" `HTTPHeaderMatch` on a `HTTPRouteRule.HTTPRouteMatch`.

// ref: https://kubernetes-sigs.github.io/gateway-api/spec/#networking.x-k8s.io/v1alpha1.HTTPHeaderMatch

youngnick:
Updated the link in the OP.

This is core support, and so must be part of our initial HTTPRoute implementation.

PR #3538, opened 2021-04-03
Implement Exact HTTPHeaderMatch

nak3:
This patch adds Exact HTTPHeaderMatch support.

Fixes https://github.com/projectcontour/contour/issues/3356

/cc @stevesloka @youngnick @sunjayBhatia

### DIFF
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -1113,6 +1113,170 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 				},
 			),
 		},
+		"insert basic single route with single header match and path match": {
+			gateway: gatewayWithSelector,
+			objs: []interface{}{
+				kuardService,
+				&gatewayapi_v1alpha1.HTTPRoute{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "basic",
+						Namespace: "projectcontour",
+						Labels: map[string]string{
+							"app":  "contour",
+							"type": "controller",
+						},
+					},
+					Spec: gatewayapi_v1alpha1.HTTPRouteSpec{
+						Hostnames: []gatewayapi_v1alpha1.Hostname{
+							"test.projectcontour.io",
+						},
+						Rules: []gatewayapi_v1alpha1.HTTPRouteRule{{
+							Matches: []gatewayapi_v1alpha1.HTTPRouteMatch{{
+								Path: gatewayapi_v1alpha1.HTTPPathMatch{
+									Type:  "Prefix",
+									Value: "/",
+								},
+								Headers: &gatewayapi_v1alpha1.HTTPHeaderMatch{
+									Type:   "Exact",
+									Values: map[string]string{"foo": "bar"},
+								},
+							}},
+							ForwardTo: []gatewayapi_v1alpha1.HTTPRouteForwardTo{{
+								ServiceName: pointer.StringPtr("kuard"),
+								Port:        gatewayPort(8080),
+							}},
+						}},
+					},
+				},
+			},
+			want: listeners(
+				&Listener{
+					Port: 80,
+					VirtualHosts: virtualhosts(virtualhost("test.projectcontour.io",
+						&Route{
+							PathMatchCondition: prefixString("/"),
+							HeaderMatchConditions: []HeaderMatchCondition{
+								{Name: "foo", Value: "bar", MatchType: "exact", Invert: false},
+							},
+							Clusters: clusters(service(kuardService)),
+						}),
+					),
+				},
+			),
+		},
+		"insert two routes with single header match, path match and header match": {
+			gateway: gatewayWithSelector,
+			objs: []interface{}{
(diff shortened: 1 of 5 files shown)


internal/dag/builder_test.go
```diff
@@ -1071,6 +1071,124 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
...
+				},
+			},
+			want: listeners(
+				&Listener{
+					Port: 80,
+					VirtualHosts: virtualhosts(virtualhost("test.projectcontour.io",
+						&Route{
+							PathMatchCondition: prefixString("/tech"),
+							HeaderMatchConditions: []HeaderMatchCondition{
+								{Name: "foo", Value: "bar", MatchType: "exact", Invert: false},
```
sunjayBhatia, line 1177: Is this correct? From the documentation it seems like there should be three separate routes generated for this test case, one that matches `/tech`, one that matches `/blog` and one that matches the header `foo` with value `bar` combined with the path `/`

internal/dag/builder_test.go
```diff
@@ -1071,6 +1071,124 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
...
+				},
+			},
+			want: listeners(
+				&Listener{
+					Port: 80,
+					VirtualHosts: virtualhosts(virtualhost("test.projectcontour.io", &Route{
+						PathMatchCondition: prefixString("/"),
+						HeaderMatchConditions: []HeaderMatchCondition{
+							{Name: "foo", Value: "bar", MatchType: "exact", Invert: false},
+						},
```
sunjayBhatia, line 1118: Is this correct? From the documentation it seems this test case should have generated two separate routes, one that matches the header `/foo` against value `bar` *and* the path `/` and another route that matches only the path `/`

sunjayBhatia: To get the output this test is currently written with you would need just this:

```
						Rules: []gatewayapi_v1alpha1.HTTPRouteRule{{
							Matches: []gatewayapi_v1alpha1.HTTPRouteMatch{{
								Headers: &gatewayapi_v1alpha1.HTTPHeaderMatch{
									Type:   "Exact",
									Values: map[string]string{"foo": "bar"},
								},
							}},
							ForwardTo: ...
```

Since matches are only ANDed together within a `HTTPRouteMatch` list element, otherwise they are treated as separate route matches

See the description here: https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.HTTPRouteRule

```
For example, take the following matches configuration:

matches:
- path:
value: "/foo"
headers:
values:
version: "2"
- path:
value: "/v2/foo"

For a request to match against this rule, a request should satisfy EITHER of the two conditions:
- path prefixed with /foo AND contains the header version: "2"
- path prefix of /v2/foo
```

sunjayBhatia (changes requested):

internal/dag/builder_test.go
nak3: You are correct. It was wrong. Thank you! Updated.

internal/dag/builder_test.go
nak3: Updated this as weell.

internal/dag/gatewayapi_processor.go
```diff
@@ -205,16 +211,29 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
 
 	for _, rule := range route.Spec.Rules {
 
-		var pathPrefixes []string
+		matchconditions := []*matchConditions{}
 		var services []*Service
 
 		for _, match := range rule.Matches {
+			mc := &matchConditions{}
 			switch match.Path.Type {
```
sunjayBhatia, line 219: probably need a `nil` check on the `Path` field as well, since it can be omitted

probably should add a test that adds a header match but no path match (which should result in a default `/` prefix match) to ensure this is covered

internal/dag/gatewayapi_processor.go
```diff
@@ -41,6 +41,12 @@ type GatewayAPIProcessor struct {
 	source *KubernetesCache
 }
 
+// matchConditions holds match rules.
+type matchConditions struct {
+	pathPrefixes         string
```
sunjayBhatia, line 46: nit: rename to the singular `pathPrefix`
```suggestion
	pathPrefix           string
```

internal/dag/gatewayapi_processor.go
nak3: Added `match.Path.Value == "" || match.Path.Type == ""` check since `match.Path` was changed to pointer recently by https://github.com/kubernetes-sigs/gateway-api/commit/9d63656df8cc9e67da60d4fe3f6289aad22e72d3 (not included in current gateway-api release) so we need to change it to nilt check when gateway-api lib is updated.

sunjayBhatia (approved):

stevesloka (approved):

PR #3539, opened 2021-04-05
Implement HTTPRouteFilterRequestHeaderModifier for HTTPRouteFilter

nak3:
This patch implements HTTPRouteFilterRequestHeaderModifier for HTTPRouteFilter.

[HTTPRouteFilter](https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.HTTPRouteFilter) is defined in `HTTPRouteForwardTo` and `HTTPRouteRule`. This patch
adds `HTTPRouteFilterRequestHeaderModifier` for both fields.

/cc @stevesloka @youngnick @sunjayBhatia

### DIFF
--- a/_integration/testsuite/gatewayapi/006-rewritehost.yaml
+++ b/_integration/testsuite/gatewayapi/006-rewritehost.yaml
@@ -1,0 +1,102 @@
+# Copyright Project Contour Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.  You may obtain
+# a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: ingress-conformance-echo
+$apply:
+  fixture:
+    as: echo-header-filter
+
+---
+
+apiVersion: v1
+kind: Service
+metadata:
+  name: ingress-conformance-echo
+$apply:
+  fixture:
+    as: echo-header-filter
+
+---
+
+apiVersion: networking.x-k8s.io/v1alpha1
+kind: Gateway
+metadata:
+  name: contour
+  namespace: projectcontour
+spec:
+  gatewayClassName: contour-class
+  listeners:
+    - protocol: HTTP
+      port: 80
+      routes:
+        kind: HTTPRoute
+        namespaces:
+          from: All
+        selector:
+          matchLabels:
+            app: filter
+
+---
+
+apiVersion: networking.x-k8s.io/v1alpha1
+kind: HTTPRoute
+metadata:
+  name: http-filter-1
+  labels:
+    app: filter
+spec:
+  hostnames:
+    - hostrewrite.projectcontour.io
+  rules:
+    - matches:
+      - path:
+          type: Prefix
+          value: /
+      filters:
+      - type: RequestHeaderModifier
+        requestHeaderModifier:
+          add:
+            Host: rewritten.com
+      forwardTo:
+      - serviceName: echo-header-filter
+        port: 80
+---
+
+import data.contour.http.client
+import data.contour.http.client.url
+import data.contour.http.expect
+import data.contour.http.response
+import data.builtin.result
+
+Response := client.Get({
+  "url": url.http("/"),
+  "headers": {
+    "Host": "hostrewrite.projectcontour.io",
+    "User-Agent": client.ua("host-rewrite-test"),
+  },
+})
+
+check_for_status_code [msg] {
+  msg := expect.response_status_is(Response, 200)
+}
+
+check_for_host_header [msg] {
+  response_body := response.body(Response)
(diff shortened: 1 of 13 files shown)


stevesloka (changes requested):
Overall this looks great @nak3, just thinking maybe we should try and serve up routes even if some headers are invalid vs failing quickly.

nak3:
Thank you @stevesloka ! Updated.
Although I updated to stop invalidate the entire route by one invalid header, other existing functions like [headersPolicyRoute](https://github.com/projectcontour/contour/blob/c55d96a6419238a6b3574cd130caf29b9b7d0fff/internal/dag/policy.go#L136) and [headersPolicyService](https://github.com/projectcontour/contour/blob/c55d96a6419238a6b3574cd130caf29b9b7d0fff/internal/dag/policy.go#L93) seems to invalidate all the entire route/service by one invalid header. Is that alright?

internal/dag/gatewayapi_processor.go
```diff
@@ -246,15 +247,53 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
...
 				continue
 			}
-			services = append(services, service)
+
+			var headerPolicy *HeadersPolicy
+			for _, filter := range forward.Filters {
+				switch filter.Type {
+				case gatewayapi_v1alpha1.HTTPRouteFilterRequestHeaderModifier:
+					if filter.RequestHeaderModifier.Add != nil {
+						routeAccessor.AddCondition(status.ConditionNotImplemented, metav1.ConditionTrue, status.ReasonHTTPRouteFilterType,
```
stevesloka, line 256: Sorry I missed this on the first pass. We should be able to handle "add". Internally it's just a "set", if the header key doesn't exist then Envoy should add the header.

internal/dag/policy.go
```diff
@@ -188,6 +190,54 @@ func headersPolicyRoute(policy *contour_api_v1.HeadersPolicy, allowHostRewrite b
 	}, nil
 }
 
+// headersPolicyGatewayAPI builds a *HeaderPolicy for the supplied HTTPRequestHeaderFilter.
+// TODO: Take care about the order of operators once https://github.com/kubernetes-sigs/gateway-api/issues/480 was solved.
+func headersPolicyGatewayAPI(hf *gatewayapi_v1alpha1.HTTPRequestHeaderFilter) (*HeadersPolicy, error) {
```
stevesloka, line 195: I think we should refactor the headersPolicy bits that exist today into something more generic such that we don't duplicate this logic. We could do that work on a separate PR, or refactor after this one.

stevesloka (changes requested):
Sorry a few more things to think about @nak3.

stevesloka:
//ref https://github.com/projectcontour/contour/issues/3559

internal/dag/gatewayapi_processor.go
nak3: Umm... As per gateway-api spec, `add` appends the header even with the same key name.

ref: https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.HTTPRequestHeaderFilter

```
Input: GET /foo HTTP/1.1 my-header: foo
Config: add: {“my-header”: “bar”}
Output: GET /foo HTTP/1.1 my-header: foo my-header: bar
```

So we can handle the `add` operation when the header key does not exist, but we cannot handle it when the header key already exist as current `HeadersPolicy.Set` is `map[string]string`. I think the support of `add` operation can include the refactoring of headersPolicy.

internal/dag/policy.go
nak3: I think refactor after this PR by separate PR is better. I will open an issue ticket for it.

nak3: Opened https://github.com/projectcontour/contour/issues/3562

internal/dag/gatewayapi_processor.go
stevesloka: If the header already exists in the request Envoy will handle the adding bits. We have a setting that handles that: https://github.com/projectcontour/contour/blob/46d5ba9df6f6634143a534f8d0edd0d497cad2dc/internal/envoy/v3/route.go#L247

internal/dag/policy.go
stevesloka: 👍

_integration/testsuite/gatewayapi/006-rewritehost.yaml
```diff
@@ -0,0 +1,102 @@
...
+  rules:
+    - matches:
+      - path:
+          type: Prefix
+          value: /
+      filters:
+      - type: RequestHeaderModifier
+        requestHeaderModifier:
+          add:
+            Host: rewritten.com
```
nak3, line 73: Rewrite Host works here. But it does not work per-service. HTTPProxy also has the issue so I opened https://github.com/projectcontour/contour/issues/3565.

internal/dag/gatewayapi_processor.go
nak3: I see! Updated. Than you!

stevesloka (approved):
/lgtm Thanks @nak3!!

sunjayBhatia (commented):
Maybe some unit tests to ensure invalid headers etc. are rejected, headers are canonicalized, and duplicates are handled appropriately with status updates etc.?

Otherwise LGTM

nak3:
@sunjayBhatia Sure, added invalid header test cases. Thank you.

sunjayBhatia (approved):

nak3:
(Rebased to fix confliction.)

Issue #3541, opened 2021-04-05
Upgrade instructions for upgrades to 1.14.0 miss out a vital step causing downtime


yoitsro:
**What steps did you take and what happened:**
I tried to upgrade my deployment to 1.14.0 following the steps mentioned here: https://projectcontour.io/resources/upgrading/#the-less-easy-way

All my endpoints became unresponsive.

**What did you expect to happen:**
I expected all my endpoints to stay up and responsive.

**Anything else you would like to add:**
This is due to a port change as part of this PR: https://github.com/projectcontour/contour/pull/3393

**Environment:**

- Contour version: 1.14.0
- Kubernetes version: (use `kubectl version`): `Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-16T23:35:15Z", GoVersion:"go1.14.2", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.15-gke.1102", GitCommit:"2e625a55a420ded12d04226c63c4cf415efc9cdb", GitTreeState:"clean", BuildDate:"2021-02-23T19:24:16Z", GoVersion:"go1.13.15b4", Compiler:"gc", Platform:"linux/amd64"}` 
- Kubernetes installer & version: GKE
- Cloud provider or hardware configuration: GKE, 2 nodes
- OS (e.g. from `/etc/os-release`): GKE cos_containerd

sunjayBhatia:
Ah apologies for that, looks like the PR in question never got a release note label so was missed in the release cutting process

sunjayBhatia:
Updated release notes: https://github.com/projectcontour/contour/releases/tag/v1.14.0

sunjayBhatia:
See https://github.com/projectcontour/contour/pull/3543

Issue #3223, opened 2020-12-17
rationalize upgrading instructions


skriss:
https://projectcontour.io/resources/upgrading/ is a pain to update manually with each release, doesn't cover things like the operator or Helm, and probably can be improved to be more user-friendly. Let's review it and see how we can both simplify and improve it, for both users and maintainers.

skriss:
What exactly are we trying to communicate in https://projectcontour.io/resources/upgrading/#the-less-easy-way ? Is this intended to describe how to upgrade with no/minimal downtime? If so, we should state that, and maybe explain *how* it minimizes downtime.

It doesn't re-apply things like RBAC; we should probably include that too.

stevesloka:
Yeah the "less-easy-way" is more of an "in-place" upgrade.

skriss:
If you want to minimize downtime, it's usually (though maybe not always) best practice to upgrade the Contour deployment before upgrading the Envoy daemonset, right? I guess the assumption there is that it's more likely that the new Contour is backwards-compatible with the old Envoy, than it is that the old Contour is forwards-compatible with the new Envoy (since Envoy tends to have more breaking changes than Contour).

This makes me think maybe we should have some automated upgrade tests, if we care about no/min-downtime upgrades.

youngnick:
It's called "the less easy way" just because it's a holdover from when the preferred way was to nuke the namespace. We should absolutely update this and have some testing around upgrades, that will be very helpful for the operator as well.

skriss:
Couple thoughts:
- perhaps renaming from "the easy way" and "the less easy way" to "reinstall" and "in-place upgrade" (or similar) would be more descriptive
- we should also add the operator
- I think we should add a sentence or two explaining that with the "in-place upgrade" instructions, we try to minimize downtime, but we don't currently guarantee no downtime.

I'll take a stab at making some updates here during this release cycle.

sunjayBhatia:
Suggestion from an other issue around the upgrade instructions (cc @yoitsro): https://github.com/projectcontour/contour/pull/3543#discussion_r607769057

PR #3543, opened 2021-04-05
Include Envoy service upgrade instructions

sunjayBhatia:
Fixes: https://github.com/projectcontour/contour/issues/3541

### DIFF
--- a/site/_resources/upgrading.md
+++ b/site/_resources/upgrading.md
@@ -71,6 +71,12 @@ $ kubectl apply -f examples/contour/02-job-certgen.yaml
 
 If your version of Contour is older than v1.13, please upgrade to v1.13 first, then upgrade to v1.14.0.
 
+This release includes an update to the Envoy service ports. Upgrade your Envoy service with the following:
+
+```bash
+$ kubectl apply -f examples/contour/02-service-envoy.yaml
+```
+
 Upgrade your Contour deployment:
 
 ```bash


site/_resources/upgrading.md
```diff
@@ -71,6 +71,12 @@ $ kubectl apply -f examples/contour/02-job-certgen.yaml
 
 If your version of Contour is older than v1.13, please upgrade to v1.13 first, then upgrade to v1.14.0.
 
+This release includes an update to the Envoy service ports. Upgrade your Envoy service with the following:
+
+```bash
+$ kubectl apply -f examples/contour/02-service-envoy.yaml
+```
+
```
skriss, line 79: I guess unfortunately there's not a great way to do this without a little bit of downtime. I'm not sure if/how we should communicate that. We don't make any guarantees right now about no-downtime upgrades so probably fine as-is.

sunjayBhatia: Yeah, this at least fixes the opened issue with some instructions so things aren't broken indefinitely, but there would still be some downtime

skriss (approved):

youngnick (approved):
Thanks for getting this sorted @sunjayBhatia. Another good lesson in how a seemingly-trivial change can be a big deal, and we have to be better about marking things as "release-note".

site/_resources/upgrading.md
yoitsro: Based on [line 41](https://github.com/projectcontour/contour/pull/3543/files#diff-a439f57ffcafffc9e8a4507a4fbb29e95d61abe4e2832b633f09994eaca3462fR41), where the easy way suggests there'll be downtime, the less easy way (which this change addresses) is still suggestive that there'll be no downtime.

sunjayBhatia: Would it make sense to pull this into a separate issue? Seems like we need some thinking around setting expectations for these upgrade instructions, as ultimately these are "example" manifests and not actually hardened upgrade instructions

skriss: xref #3223, yeah I'd say let's get this omission merged and revisit upgrades more broadly.

PR #3549, opened 2021-04-06
internal/xdscache: extract route cache helper

### LINKED ISSUE #3507: internal/xdscache/v3/route.go: remove duplication
This file has a bunch of duplication in the visitor code across the sections that deal with insecure vhosts, secure vhosts, and fallback vhosts. #3409 was an example of a bug that was partly caused by this duplication, because a change was made in one place but not duplicated in the other spots.

 It looks like some helpers could be extracted to go from DAG routes to Envoy routes, and from DAG vhosts to Envoy vhosts. Not sure yet if they should go into internal/envoy/v3, or stay in the internal/xdscache/v3 package.

skriss:
Extracts a helper function, toEnvoyVirtualHost, to reduce
duplication across the route cache code for virtual hosts,
secure virtual hosts, and fallback virtual hosts.

Closes #3507.

### DIFF
--- a/internal/xdscache/v3/route.go
+++ b/internal/xdscache/v3/route.go
@@ -129,6 +129,10 @@ func (v *routeVisitor) onVirtualHost(vh *dag.VirtualHost) {
 		routes = append(routes, route)
 	})
 
+	if len(routes) == 0 {
+		return
+	}
+
 	toEnvoyRoute := func(route *dag.Route) *envoy_route_v3.Route {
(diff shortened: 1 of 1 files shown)


sunjayBhatia (approved):
LGTM, +1 for DRYing this code out, had feedback cycles in some changes where having to go back to redo the same thing was a pain/mistake prone

stevesloka (approved):

PR #3550, opened 2021-04-06
internal/dag: always apply default headers

mattatcha:
Fixes a case where default header policies would not be applied unless headers were set in the HTTPProxy


skriss (approved):
good catch @mattatcha, LGTM

sunjayBhatia (approved):

PR #3555, opened 2021-04-08
internal/dag: Return an HTTP503 if the service on an invalid ForwardTo

### LINKED ISSUE #3553: Route.ForwardTo missing service should return HTTP 503
The ForwardTo defines the backend(s) where matching requests should be sent on an HTTPRoute. If unspecified, the rule performs no forwarding. If unspecified and no filters are specified that would result in a response being sent, a 503 error code is returned.

//ref: https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.HTTPRouteRule

stevesloka:
If the service on an HTTPRoute.Spec.Rule.ForwardTo is invalid or missing, an HTTP503 should be returned.

Fixes #3553

### DIFF
--- a/_integration/testsuite/gatewayapi/003-invalid-forwardto.yaml
+++ b/_integration/testsuite/gatewayapi/003-invalid-forwardto.yaml
@@ -83,7 +83,7 @@ spec:
           type: Prefix
           value: /invalidport
       forwardTo:
-      - serviceName: invalid
+      - serviceName: echo-slash-default
     - matches:
         - path:
             type: Prefix
--- a/_integration/testsuite/gatewayapi/003-invalid-forwardto.yaml
+++ b/_integration/testsuite/gatewayapi/003-invalid-forwardto.yaml
@@ -1,0 +1,233 @@
+# Copyright Project Contour Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.  You may obtain
+# a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: ingress-conformance-echo
+$apply:
+  fixture:
+    as: echo-slash-prefix
+
+---
+
+apiVersion: v1
+kind: Service
+metadata:
+  name: ingress-conformance-echo
+$apply:
+  fixture:
+    as: echo-slash-prefix
+
+---
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: ingress-conformance-echo
+$apply:
+  fixture:
+    as: echo-slash-noprefix
+
+---
+
+apiVersion: v1
+kind: Service
+metadata:
+  name: ingress-conformance-echo
+$apply:
+  fixture:
+    as: echo-slash-noprefix
+
+---
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: ingress-conformance-echo
+$apply:
+  fixture:
+    as: echo-slash-default
+
+---
+
+apiVersion: v1
+kind: Service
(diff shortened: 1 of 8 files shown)


internal/dag/builder_test.go
```diff
@@ -983,7 +983,14 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 					},
 				},
 			},
-			want: listeners(),
+			want: listeners(
```
sunjayBhatia, line 986: nit: maybe update the comment before this test case to reflect this change

internal/dag/gatewayapi_processor.go
```diff
@@ -249,15 +250,18 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
...
 		routes := p.routes(pathMatchConditions, services)
 		for _, vhost := range hosts {
 			vhost := p.dag.EnsureVirtualHost(ListenerName{Name: vhost, ListenerName: "ingress_http"})
 			for _, route := range routes {
+				if len(services) == 0 {
+					// Configure a direct response HTTP status code of 503 so the
+					// route still matches the configured conditions since the
+					// service is missing or invalid.
+					route.DirectResponse = &DirectResponse{
+						StatusCode: 503,
```
sunjayBhatia, line 262: nit: maybe use a constant e.g. `http.StatusServiceUnavailable`

_integration/testsuite/gatewayapi/003-invalid-forwardto.yaml
```diff
@@ -0,0 +1,233 @@
...
+            value: /invalidservicename
+      forwardTo:
+        - serviceName: ""
+          port: 80
+    - matches:
+      - path:
+          type: Prefix
+          value: /
+      forwardTo:
+      - serviceName: echo-slash-default
```
sunjayBhatia, line 159: I think all the deployments/services besides this one can be deleted since we only need one valid one

internal/dag/gatewayapi_processor.go
stevesloka: good call! =)

_integration/testsuite/gatewayapi/003-invalid-forwardto.yaml
stevesloka: 👍

sunjayBhatia (approved):

_integration/testsuite/gatewayapi/003-invalid-forwardto.yaml
```diff
@@ -0,0 +1,173 @@
...
+          value: /invalidref
+      forwardTo:
+      - serviceName: invalid
+        port: 80
+    - matches:
+      - path:
+          type: Prefix
+          value: /invalidport
+      forwardTo:
+      - serviceName: invalid
```
skriss, line 86: maybe use a valid service name here, to ensure this is hitting the "invalid port" case

internal/xdscache/v3/route.go
```diff
@@ -143,6 +143,14 @@ func (v *routeVisitor) onVirtualHost(vh *dag.VirtualHost) {
 				Action: envoy_v3.UpgradeHTTPS(),
 			}
 		}
+
+		if route.DirectResponse != nil {
+			return &envoy_route_v3.Route{
+				Match:  envoy_v3.RouteMatch(route),
+				Action: envoy_v3.RouteDirectResponse(route.DirectResponse),
+			}
+		}
```
skriss, line 152: Just wondering if we should still apply request/response headers, rate limit policies to the route even if it's a direct response. I guess it's kind of a moot point right now since those things aren't supported yet for Gateway, and we can revisit if/when we support direct response for HTTPProxy.

skriss (changes requested):
Looks great @stevesloka, just one test nit. Looking forward to being able to use this in HTTPProxy too.

_integration/testsuite/gatewayapi/003-invalid-forwardto.yaml
stevesloka: Yup good call.

internal/xdscache/v3/route.go
stevesloka: I think it's a great point that we should keep in mind when Gateway gets that support and also to add to HTTPProxy when we do that.

skriss (approved):
LGTM

PR #3569, opened 2021-04-12
internal/dag: Add weight processing to HTTPRoute.ForwardTos

### LINKED ISSUE #3560: Support weight on HTTPRoute.ForwardTo
Weight specifies the proportion of HTTP requests forwarded to the backend referenced by the ServiceName or BackendRef field. This is computed as weight/(sum of all weights in this ForwardTo list). For non-zero values, there may be some epsilon from the exact proportion defined here depending on the precision an implementation supports. Weight is not a percentage and the sum of weights does not need to equal 100.

If only one backend is specified and it has a weight greater than 0, 100% of the traffic is forwarded to that backend. If weight is set to 0, no traffic should be forwarded for this entry. If unspecified, weight defaults to 1.

// ref: https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.HTTPRouteForwardTo

stevesloka:
Adds weight processing to the HTTPRoute.ForwardTos. If not specified,
defaults to the API default of "1". If set to zero, and all other services
are also set to zero or are invalid, an HTTP 503 response is returned.

Fixes #3560

### DIFF
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -55,6 +55,36 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 		},
 	}
 
+	kuardService2 := &v1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "kuard2",
+			Namespace: "projectcontour",
+		},
+		Spec: v1.ServiceSpec{
+			Ports: []v1.ServicePort{{
+				Name:       "http",
+				Protocol:   "TCP",
+				Port:       8080,
+				TargetPort: intstr.FromInt(8080),
+			}},
+		},
+	}
+
+	kuardService3 := &v1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "kuard3",
+			Namespace: "projectcontour",
+		},
+		Spec: v1.ServiceSpec{
+			Ports: []v1.ServicePort{{
+				Name:       "http",
+				Protocol:   "TCP",
+				Port:       8080,
+				TargetPort: intstr.FromInt(8080),
+			}},
+		},
+	}
+
 	kuardServiceCustomNs := &v1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "kuard",
@@ -267,16 +297,8 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 				"test.projectcontour.io",
 			},
 			Rules: []gatewayapi_v1alpha1.HTTPRouteRule{{
-				Matches: []gatewayapi_v1alpha1.HTTPRouteMatch{{
-					Path: gatewayapi_v1alpha1.HTTPPathMatch{
-						Type:  "Prefix",
-						Value: "/",
-					},
-				}},
-				ForwardTo: []gatewayapi_v1alpha1.HTTPRouteForwardTo{{
-					ServiceName: pointer.StringPtr("kuard"),
-					Port:        gatewayPort(8080),
-				}},
+				Matches:   httpRouteMatch(gatewayapi_v1alpha1.PathMatchPrefix, "/"),
+				ForwardTo: httpRouteForwardTo("kuard", 8080, 1),
 			}},
 		},
 	}
@@ -296,16 +318,8 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 				"test.projectcontour.io",
 			},
 			Rules: []gatewayapi_v1alpha1.HTTPRouteRule{{
-				Matches: []gatewayapi_v1alpha1.HTTPRouteMatch{{
-					Path: gatewayapi_v1alpha1.HTTPPathMatch{
-						Type:  "Prefix",
-						Value: "/",
-					},
-				}},
-				ForwardTo: []gatewayapi_v1alpha1.HTTPRouteForwardTo{{
-					ServiceName: pointer.StringPtr("blogsvc"),
-					Port:        gatewayPort(80),
-				}},
+				Matches:   httpRouteMatch(gatewayapi_v1alpha1.PathMatchPrefix, "/"),
+				ForwardTo: httpRouteForwardTo("blogsvc", 80, 1),
 			}},
 		},
 	}
@@ -447,7 +461,7 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 				&Listener{
 					Port: 80,
 					VirtualHosts: virtualhosts(
-						virtualhost("test.projectcontour.io", prefixroute("/", service(kuardService))),
+						virtualhost("test.projectcontour.io", prefixrouteHTTPRoute("/", service(kuardService))),
 					),
 				},
 			),
@@ -463,7 +477,7 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 				&Listener{
 					Port: 80,
 					VirtualHosts: virtualhosts(
-						virtualhost("test.projectcontour.io", prefixroute("/", service(kuardService))),
+						virtualhost("test.projectcontour.io", prefixrouteHTTPRoute("/", service(kuardService))),
 					),
 				},
 			),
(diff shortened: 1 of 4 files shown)


stevesloka:
Note, didn't add any feature tests. Those are started in #3472, so to avoid all the duplication, would like to merge that first, then can add more tests.

sunjayBhatia (approved):
LGTM, looks straightforward

At least with the current integration test framework, seems like it's not the most straightforward to write up a good test that wont be flaky

internal/dag/builder_test.go
```diff
@@ -1107,6 +1139,193 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
...
+									ServiceNamespace: kuardService3.Namespace,
+									ServicePort:      kuardService3.Spec.Ports[0],
+								},
+							},
+						)),
+					),
+				},
+			),
+		},
+		"one service weight zero w/weights for other forwardTos": {
```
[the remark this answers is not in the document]
stevesloka: It's an easy way to stage changes to routes. So you can configure the route with a new service of weight 0, then add the service and update the weight. 

For Contour, it's nice because the clusters/endpoints will be configured in Envoy, then to activate the traffic can flow faster without needing configuration changes to push over to Envoy.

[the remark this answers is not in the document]
sunjayBhatia: 👍🏽 that makes sense, switching weights to switch over to a new backend

stevesloka:
Ok this is all rebased, I'm going to clean up some of the feature tests a bit since we've got enough to test now. Plus this is currently difficult to test with the current e2e integration setup.

internal/dag/gatewayapi_processor.go
```diff
@@ -309,6 +309,7 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
 		var clusters []*Cluster
 
 		// Validate the ForwardTos.
+		totalWeight := int32(0)
```
danehans, line 312: From gateway api spec:

"If only one backend is specified and it has a weight greater than 0, 100% of the traffic is forwarded to that backend."

@stevesloka can you help me understand how this is implemented?

internal/dag/gatewayapi_processor.go
```diff
@@ -332,6 +333,10 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
 				continue
 			}
 
+			// Apply the weight of the forwardTo.
+			service.Weighted.Weight = uint32(forward.Weight)
```
danehans, line 337: From gateway api spec:

"If unspecified, weight defaults to 1."

Do a nil check, and if so, set to 1?

danehans:
@stevesloka from the gateway api spec:

"Weight specifies the proportion of HTTP requests forwarded to the backend referenced by the ServiceName or BackendRef field. This is computed as weight/(sum of all weights in this ForwardTo list)."

Can you help me understand how this is implemented?

> Ok this is all rebased, I'm going to clean up some of the feature tests a bit since we've got enough to test now. Plus this is currently difficult to test with the current e2e integration setup.

Will an e2e be created that pumps requests through a pair of weighted backend services and verifies the traffic split is as expected?

internal/dag/gatewayapi_processor.go
stevesloka: If no weight is supplied on the forwardTo, then it's given a default of `1` and that backend will get all the traffic. 

What this is doing is looking at all the combined weights of all the forwardTos and looking if they are all configured to zero, if that's true then we should return an HTTP 503 since there won't be any clusters to proxy to.

internal/dag/gatewayapi_processor.go
stevesloka: We're still based off of v0.2.0 where weight isn't nil, but has since changed in `master` branch. I don't understand why we have this as a nil with a default. To me, if it's not defined, you get `1` as the default so is it really going to be nil? 

We'll need to update this when we move to work off of master which should come next (https://github.com/projectcontour/contour/issues/3578).

stevesloka:
> "Weight specifies the proportion of HTTP requests forwarded to the backend referenced by the ServiceName or BackendRef field. This is computed as weight/(sum of all weights in this ForwardTo list)."
> Can you help me understand how this is implemented?

This is done via Envoy. If a weight is configured, we pass that to the route with a weighted cluster setup (https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route_components.proto#envoy-api-msg-route-weightedcluster). Envoy will then load balance between the clusters with the weights defined. Contour will add up all the weights and set the `totalWeight` field. 

> Will an e2e be created that pumps requests through a pair of weighted backend services and verifies the traffic split is as expected?

I'm writing some feature tests now to validate the proper Envoy config is passed (e.g. your previous question). E2E tests is going to be difficult to track since weights aren't 100% split evenly, meaning, if you setup 2 services both with a weight of `1` (even distribution) and curl that, you won't get alternating responses each time.

stevesloka:
@danehans I think I found a bug with the refactor from the headers bit. This test I'm adding is super helpful, let me fix this up and repush a change.

stevesloka:
@danehans ok all fixed up now. 

This feels like a mess though, we really need to refactor out the `Weighted` cluster part of `Cluster` (https://github.com/projectcontour/contour/issues/3593). It's making this super confusing to test.

sunjayBhatia (approved):
re approving the featuretests changes

internal/dag/builder_test.go
```diff
@@ -276,6 +306,7 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 				ForwardTo: []gatewayapi_v1alpha1.HTTPRouteForwardTo{{
 					ServiceName: pointer.StringPtr("kuard"),
 					Port:        gatewayPort(8080),
+					Weight:      1,
```
skriss, line 309: nit, looks like you could use `httpRouteForwardTo` here to simplify

internal/dag/builder_test.go
```diff
@@ -1713,6 +1739,193 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
...
+								}, {
+									ServiceName: pointer.StringPtr("kuard2"),
+									Port:        gatewayPort(8080),
+									Weight:      10,
+								},
+								{
+									ServiceName: pointer.StringPtr("kuard3"),
+									Port:        gatewayPort(8080),
+									Weight:      15,
+								},
```
skriss, line 1774: use `httpRouteForwardTo` here?

internal/dag/builder_test.go
```diff
@@ -1713,6 +1739,193 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
...
+								}, {
+									ServiceName: pointer.StringPtr("kuard2"),
+									Port:        gatewayPort(8080),
+									Weight:      0,
+								},
+								{
+									ServiceName: pointer.StringPtr("kuard3"),
+									Port:        gatewayPort(8080),
+									Weight:      15,
+								},
```
skriss, line 1846: use `httpRouteForwardTo` here?

internal/dag/builder_test.go
```diff
@@ -1713,6 +1739,193 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
...
+						},
+					},
+					Spec: gatewayapi_v1alpha1.HTTPRouteSpec{
+						Rules: []gatewayapi_v1alpha1.HTTPRouteRule{{
+							Matches: httpRouteMatch(gatewayapi_v1alpha1.PathMatchPrefix, "/"),
+							ForwardTo: []gatewayapi_v1alpha1.HTTPRouteForwardTo{{
+								ServiceName: pointer.StringPtr("kuard"),
+								Port:        gatewayPort(8080),
+								Weight:      0,
+							}},
```
skriss, line 1908: use `httpRouteForwardTo`?

skriss (changes requested):
Couple test nits otherwise LGTM

stevesloka:
@skriss I think I've got this all fixed up with all the helpers.

skriss (approved):
nice, LGTM!

Issue #3564, opened 2021-04-09
ClusterDiscoveryType on Clusters with an IP as address should be Static


sunjayBhatia:
When an IP address is configured as the address of a cluster, it seems we don't need to configure anything more complicated than Static Cluster Discovery: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/service_discovery#static

Currently this is an issue with setting the Admin address to an ipv6 address (e.g. `::1`) since we use Logical DNS discovery and run into https://github.com/envoyproxy/envoy/issues/10489. Configuring `127.0.0.1` happens to "just work" as mentioned in the issue because of Envoys implementation of URL parsing (the error Envoy throws in URL parsing is around finding `:` characters, which is different in ipv6 addresses of course).

Seems like the Envoy bug should be fixed anyway, but this would make things simpler

sunjayBhatia:
Partially completed by https://github.com/projectcontour/contour/pull/3572, we could consider this done since practically only the admin address is likely to be configured with an IP address, but for completeness we should apply this to other upstream clusters as well (`ExternalName` services could be configured as an IP address potentially so this is a direct place this comes into play)

sunjayBhatia:
closing since we're basically disabling ExternalName services and noone has asked for it

static DNS is being used in bootstrap clusters with IP addresses

Issue #3284, opened 2021-01-28
ipv6 validation


xaleeks:
Have been getting lots of pings from Telco&Cloud customers who would like to run Contour on ipv6 environments, this means both Single Stack and Dual Stack k8s clusters.  We need to validate Contour deployments on both Tanzu k8s products as well on DIY cloud platforms.  Before identifying the exact set of stipulations for 'Contour support ipv6', the very basic scenario we have in mind is:

We assume the entire k8s cluster is running on ipv6 (assuming worker nodes running ipv6, pods have ipv6 connectivity with each other & to the internet, an ipv6 capable CNI, an ipv6 capable external Loadbalancer), Contour should run as usual and brings traffic into pods running on ipv6 addresses.  And the same functionality for pods with dual ipv4 / ipv6 addresses.

timeframe: v1.13 or v1.14 before May

youngnick:
I think that all the conditions I can think of for this to work are:
- Contour must be reachable from Envoy, currently we supply the DNS name `contour` for Envoy to connect back to Contour. In an IPv6 world, this will be a AAAA lookup, so the DNS must support AAAA for service records (since the `contour` name is the name of the Service).
- Contour must be able to publish the endpoints of the pods to route to to Envoy.
- Envoy must be reachable on its serving address (aka the "outside" address). So you'll either need a `Service` of `type: LoadBalancer` (what we do now), or some equivalent functionality that works with IPv6.
- Envoy must be able to route the traffic to the running pods. So the dynamic IPv6 routing must be working in the cluster.

I think that any cluster that's actually working in IPv6 only mode should probably work. I don't know about dual-stack, it's possible we may need to do some work with Endpoints, I haven't looked at dual-stack enough to be sure.

In any case, the steps from here seem to be the same:
- Test Contour on an IPv6-only cluster.
- Test Contour on a dual-stack cluster.

Ideally we figure out a way to run the current set of integration tests in both those cases.

sunjayBhatia:
What do we think about running some validation (the integration test suite probably) against an ipv6 cluster in CI and writing a guide on what all the flags to flip are to support a cluster with ipv6 only?

Initially seems like some bits to do around Cluster discovery (see https://github.com/projectcontour/contour/issues/3564) and how we parse ipv6 addresses consistently, there are some rough edges where some flags need to take `::` but others need `[::]` etc.

xaleeks:
[...]

PR #3572, opened 2021-04-13
internal/envoy/v3: Bootstrap clusters with IPs configured as STATIC

sunjayBhatia:
Strict/Logical DNS clusters fail to parse ipv6 IPs and according to the
Envoy documentation should not use an IP
See: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/service_discovery#arch-overview-service-discovery-types
See: https://github.com/envoyproxy/envoy/issues/10489

Limited to bootstrap clusters for now

Also cleans up whitespace in tests

Updates: #3564
Updates: #3284


skriss (approved):
LGTM

stevesloka (approved):

PR #3579, opened 2021-04-14
internal/httpsvc: Join host/port with net helper

sunjayBhatia:
Ensures we take an ipv6 address w/o brackets which matches the existing flags

Updates #3284


youngnick (approved):
Awesome that this is such a small fix.

stevesloka (approved):

PR #3581, opened 2021-04-14
Update k8s deps for 1.21.0 release

sunjayBhatia:
No kind image yet so no update to tested k8s versions yet


stevesloka:
Could you also update the support matrix bits as well?

sunjayBhatia:
I'll add an Issue for the 1.15.0 release, since we're not actually testing against 1.21.0 yet (no kind image) I would hold off on changing the matrix

sunjayBhatia:
https://github.com/projectcontour/contour/issues/3583

youngnick:
I think we should add a 'main` line to the support matrix, that makes keeping everything up to date much easier. In addition, we probably need to remove really old versions.

sunjayBhatia:
> I think we should add a 'main` line to the support matrix, that makes keeping everything up to date much easier. In addition, we probably need to remove really old versions.

https://github.com/projectcontour/contour/pull/3605

youngnick (approved):

Issue #3467, opened 2021-03-10
Ingress: Support multiple secrets for a given host


sunjayBhatia:
The [Ingress spec](https://pkg.go.dev/k8s.io/api/networking/v1#IngressSpec) supports configuring multiple TLS secrets which each can be applied to [multiple hostnames](https://pkg.go.dev/k8s.io/api/networking/v1#IngressTLS).

There is no admission check for uniqueness of secret names or hostnames within the various lists of hosts.

So, something like below is valid:

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: test1
spec:
  tls:
  - secretName: foo
    hosts:
    - foo.bar.com
    - bar.com
    - foo.bar.com
  - secretName: bar
    hosts:
    - foo.bar.com
    - bar.com
    - foo.bar.com
  - secretName: bar
    hosts:
    - bar.com
...
```

Currently, we create ensure a secure virtualhost in the DAG exists for each hostname on a secret. However, since we only allow one secret per, the last one in the list "wins"

We should fully support the spec and allow a list of secrets to be configured and then set on the TLS Transport socket in Envoy config (see [this repeated field](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/tls.proto.html#envoy-v3-api-field-extensions-transport-sockets-tls-v3-commontlscontext-tls-certificate-sds-secret-configs))

abhide:
@sunjayBhatia would like to work on fixing this one. Can I assign this to myself ?

xaleeks:
Can we tag this v1.15? @sunjayBhatia

sunjayBhatia:
Possibly, not sure if @abhide has had time to look into it yet

abhide:
@sunjayBhatia I haven't started working on this yet. Will look into this weekend.

abhide:
@sunjayBhatia Seems like we need to change https://github.com/projectcontour/contour/blob/main/internal/dag/dag.go#L486 to be  a slice of Secret? and fix things accordingly.
Just checking if I am on the correct path here?

sunjayBhatia:
Yep sounds right to me, might be useful to get started with a `internal/featuretests` test and work from there 👍🏽

abhide:
> Yep sounds right to me, might be useful to get started with a `internal/featuretests` test and work from there 👍🏽 

Thanks @sunjayBhatia

PR #3592, opened 2021-04-19
Ingress: Support multiple secrets for a given host

abhide:
Fixes: #3467

### DIFF
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -1483,7 +1483,7 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 								ListenerName: "ingress_https",
 								routes:       routes(prefixrouteHTTPRoute("/", service(kuardService))),
 							},
-							Secrets: []*Secret{secret(sec1)},
+							Secrets: map[string]*Secret{sec1.Name: {Object: sec1}},
 						},
 					),
 				},
@@ -1530,7 +1530,7 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 								ListenerName: "ingress_https",
 								routes:       routes(prefixrouteHTTPRoute("/", service(blogService))),
 							},
-							Secrets: []*Secret{secret(sec1)},
+							Secrets: map[string]*Secret{sec1.Name: {Object: sec1}},
 						},
 					),
 				},
@@ -1716,7 +1716,7 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 								ListenerName: "ingress_https",
 								routes:       routes(prefixrouteHTTPRoute("/", service(blogService))),
 							},
-							Secrets: []*Secret{secret(sec1)},
+							Secrets: map[string]*Secret{sec1.Name: {Object: sec1}},
 						},
 					),
 				},
@@ -5517,9 +5517,6 @@ func TestDAGInsert(t *testing.T) {
 				&Listener{
 					Port: 80,
 					VirtualHosts: virtualhosts(
-<<<<<<< HEAD
-						virtualhost("*", prefixroute("/", service(s1))),
-=======
 						virtualhost("foo.com", routeUpgrade("/", service(s1))),
 					),
 				}, &Listener{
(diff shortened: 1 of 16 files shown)


sunjayBhatia (changes requested):
I think one `featuretest` (or unit test, but something that shows the full Envoy config generated would be nice) that would be good to have would be creating an ingress with multiple TLS secrets that have the same hostname they should be applied for:

the equivalent of this:
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: wildcard-ingress
spec:
  tls:
  - hosts:
    - "foo.com"
    secretName: foo1
  - hosts:
    - "foo.com"
    secretName: foo2
  rules:
  - host: "foo.com"
...
```

abhide:
Thanks @sunjayBhatia . Will add that test. 
Trying to figure out the current test failure.

internal/dag/builder_test.go
```diff
@@ -1114,7 +1114,7 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 								ListenerName: "ingress_https",
 								routes:       routes(prefixroute("/", service(kuardService))),
 							},
-							Secret: secret(sec1),
+							Secrets: []*Secret{secret(sec1)},
```
sunjayBhatia, line 1117: refactor suggestion: turn this to use a new `secrets()`  helper that takes a variadic list of secrets (the k8s type) and turns it into a list

signature like this: `func secrets(secrets ...*v1.Secret) []*Secret`

should be able to keep the `secret()` helper as well in case it is useful

internal/dag/gatewayapi_processor.go
```diff
@@ -379,7 +379,7 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
 
 				if listenerSecret != nil {
 					svhost := p.dag.EnsureSecureVirtualHost(ListenerName{Name: host, ListenerName: "ingress_https"})
-					svhost.Secret = listenerSecret
+					svhost.Secrets = append(svhost.Secrets, listenerSecret)
```
sunjayBhatia, line 382: The gateway API `featuretests` are failing b/c with multiple HTTPRoute objects or a single one with multiple path matches, this will be called multiple times for the same secret

see https://github.com/projectcontour/contour/blob/efecfc2b8a56a9744b589dedf1e8c844b3660776/internal/dag/gatewayapi_processor.go#L367-L383

sunjayBhatia (commented):
Seems like we need to think a bit about how we implement multiple secrets in the DAG so we don't append duplicates and so we properly define how we want to handle interactions between different resources attempting to create routes for the same secure host

On duplicates:
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: wildcard-ingress
spec:
  tls:
  - hosts:
    - "*.projectcontour.io"
    secretName: wildcard
  - hosts:
    - "*.projectcontour.io"
    secretName: wildcard
```

The above is a config that is accepted by k8s. With the change as-is we would add the wildcard secret twice, which seems suboptimal (Maybe the collection of secrets should be implemented as a map, "secretnamespace/secretname" -> secret).

On interaction between different resources:

Before this change, Ingress resources are processed first, can be overriden by HTTPProxies, which can then be overriden by HTTPRoutes. If we intend to keep this behavior, we probably need some tests showing that and maybe need to not append secrets if both an Ingress and HTTPProxy/HTTPRoute specify them for a hostname.

Similarly, between Ingresses, whichever is processed last "wins" with TLS secret configuration. This behavior might need to change? Not sure what the other @projectcontour/maintainers think about this topic

youngnick:
I think that both behaviors here are locked in (last Ingress in wins, HTTPProxy overrides Ingress TLS), since they've been in place for some time. The HTTPProxy overriding Ingress rule is very useful to allow migrations from Ingress to HTTPProxy - you can drop the HTTPProxy in, and back out by deleting it if something doesn't work.

I think there are some tests around Ingress being overridden by HTTPProxy already, but I agree we should test how the TLS handling works between the two.

PR #3599, opened 2021-04-20
global rate limiting: support header value matches

### LINKED ISSUE #3595: global rate limiting: support envoy HeaderValueMatch action
Envoy has a [HeaderValueMatch rate limiting action](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-ratelimit-action-headervaluematch). This supports matching on header values, including not-exists, prefix, suffix, etc. matching.

Contour should support this rate limiting action. Note that Contour already exposes a [HeaderMatchCondition](https://github.com/projectcontour/contour/blob/main/apis/projectcontour/v1/httpproxy.go#L69-L102) for the purposes of routing; we should be able to reuse this in the API.

ref. https://kubernetes.slack.com/archives/C8XRH2R4J/p1618493972125100

skriss:
Adds support for Envoy's HeaderValueMatch rate limit action
which generates descriptor entries based on match criteria
for the request's headers.

Closes #3595.


skriss:
Note, support for `notpresent` header matching to follow in a separate PR (https://github.com/skriss/contour/tree/header-match-notpresent)

stevesloka (approved):

apis/projectcontour/v1/httpproxy.go
```diff
@@ -411,6 +411,12 @@ type RateLimitDescriptorEntry struct {
 	// +optional
 	RequestHeader *RequestHeaderDescriptor `json:"requestHeader,omitempty"`
 
+	// RequestHeaderValueMatch defines a descriptor entry that's populated
+	// if the request's headers match a set of 1+ match criteria. The
+	// descriptor key is "header_match", and the descriptor value is static.
+	// +optional
+	RequestHeaderValueMatch *RequestHeaderValueMatchDescriptor `json:"requestHeaderValueMatch,omitempty"`
```
sunjayBhatia, line 418: We probably can't remove it but this sorta overlaps with the existing `RequestHeader` a bit since the `HeaderMatchCondition` type has a `Present` field. Difference I guess is that this one will set the descriptor key and value to be static. Might be good to call out in some docs

sunjayBhatia (approved):

apis/projectcontour/v1/httpproxy.go
skriss: Yeah, it's similar but I think different enough to be its own thing (as it is in Envoy). I'll do a follow-up PR to add some more docs on each descriptor entry type to the rate limiting page; the API godoc covers it somewhat but some additional material would be helpful too.

skriss: #3604

PR #3602, opened 2021-04-20
header match conditions: add support for notpresent

skriss:
Adds support for the notpresent header match type.

### DIFF
--- a/internal/dag/policy_test.go
+++ b/internal/dag/policy_test.go
@@ -761,12 +761,12 @@ func TestRateLimitPolicy(t *testing.T) {
 									RequestHeaderValueMatch: &contour_api_v1.RequestHeaderValueMatchDescriptor{
 										Headers: []contour_api_v1.HeaderMatchCondition{
 											{
-												Name:    "X-Header",
-												Present: true,
+												Name:       "X-Header",
+												NotPresent: true,
 											},
 										},
 										ExpectMatch: true,
-										Value:       "header-is-present",
+										Value:       "header-is-not-present",
 									},
 								},
 							},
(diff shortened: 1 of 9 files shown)


skriss:
Updated docs since this impacts header conditions for routing too.

sunjayBhatia (approved):

skriss:
rebased.

skriss:
[Site proof failure](https://github.com/projectcontour/contour/pull/3602/checks?check_run_id=2411437174) can be ignored, that link works fine.

internal/dag/policy_test.go
```diff
@@ -761,12 +761,12 @@ func TestRateLimitPolicy(t *testing.T) {
 									RequestHeaderValueMatch: &contour_api_v1.RequestHeaderValueMatchDescriptor{
 										Headers: []contour_api_v1.HeaderMatchCondition{
 											{
-												Name:    "X-Header",
-												Present: true,
+												Name:       "X-Header",
```
stevesloka, line 764: Do we still have the `Present` test?

skriss: It's covered here: https://github.com/projectcontour/contour/pull/3602/files/9cc3aaddcc7e9f819f13955d042608e3f00e8697#diff-ae9441e6df4b29887feba6f76574a3d376a09c00edde988ad16e1609548ad940L660-L665

skriss: Not directly in this test but the function that processes all of those match types is `headerMatchConditions` (called at https://github.com/projectcontour/contour/blob/9cc3aaddcc7e9f819f13955d042608e3f00e8697/internal/dag/policy.go#L560) and the test for that function covers it: https://github.com/projectcontour/contour/blob/main/internal/dag/conditions_test.go#L116-L127

skriss: Just to follow up on this, IMHO this test doesn't need to cover all the permutations of header match types, but let me know if you disagree @stevesloka. Definitely want to get this in for 1.15 so let me know if you have any other comments!

stevesloka: I just asked because the "header-is-present" was removed and swapped for "header-is-not-present".

stevesloka (approved):

PR #3603, opened 2021-04-21
 internal: bump gatewayapi dep to latest

### LINKED ISSUE #3578: Update GatewayAPI Reference to use SHA
We're currently pinned against the v0.2.0 version of the GatewayAPI release, but to make the breaking changes easier we should swap out to use the latest github sha. 

At the next Contour release we need to update this back to v0.3.0 as described in #3577

stevesloka:
Updates the gateway-api dep to "latest" to keep on pace with breaking  changes in the upcoming v0.3.0 release.
    
Fixes #3578


stevesloka:
I tried to keep things into separate commits to make review easier, but the pointers, oh my. =)

skriss:
Looks like this also covers the k8s 1.21 bump that's also in https://github.com/projectcontour/contour/pull/3581 - can either merge https://github.com/projectcontour/contour/pull/3581 or close it

stevesloka:
We can merge #3581 first no problem. Let me go build out a 1.21 kind image to test with in the meantime until an official image is out.

youngnick (commented):
This is looking great, and really illustrates why it's better to do this *now*, rather than later.

skriss:
needs a rebase now

stevesloka:
I ended up squashing my original commits since I had to keep doing the entire rebase again. 😢

internal/dag/gatewayapi_processor.go
```diff
@@ -104,14 +108,22 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
 
-			nsMatches, err := p.namespaceMatches(listener.Routes.Namespaces, route)
-			if err != nil {
-				p.Errorf("error validating namespaces against Listener.Routes.Namespaces: %s", err)
+			var err error
+			nsMatches := true
+			selMatches := true
+
+			if listener.Routes.Namespaces != nil {
+				nsMatches, err = p.namespaceMatches(*listener.Routes.Namespaces, route)
```
sunjayBhatia, line 116: would it be a little cleaner to have `namespaceMatches` and `selectorMatches` take a pointer and just return true if the passed in namespace/selector is nil?

internal/dag/gatewayapi_processor.go
```diff
@@ -269,32 +285,34 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
...
-		matchconditions := []*matchConditions{}
+		var matchconditions []*matchConditions
 
 		for _, match := range rule.Matches {
 			mc := &matchConditions{}
-			// TODO: Replace this with nil check when gateway-api includes this fix -
-			// https://github.com/kubernetes-sigs/gateway-api/commit/9d63656df8cc9e67da60d4fe3f6289aad22e72d3
-			if match.Path.Value == "" || match.Path.Type == "" {
-				match.Path = gatewayapi_v1alpha1.HTTPPathMatch{Type: gatewayapi_v1alpha1.PathMatchPrefix, Value: "/"}
+			if match.Path == nil || match.Path.Value == nil || match.Path.Type == nil {
```
sunjayBhatia, line 292: Ah, I think this may have been wrong before and this is a bit wrong still?

In the case you have specified a path value that is nonempty/non-nil and no type, i think your path would be overwritten with `/`?

sunjayBhatia: should this be `match.Path == nil || (match.Path.Value == nil && match.Path.Type == nil)`

or maybe separate out the cases, check if `match.Path` is `nil` and fill it in, and then fill in a `nil` value or type if the whole match wasn't `nil`

internal/dag/gatewayapi_processor.go
```diff
@@ -269,32 +285,34 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
...
+				match.Path = &gatewayapi_v1alpha1.HTTPPathMatch{Type: pathMatchTypePtr(gatewayapi_v1alpha1.PathMatchPrefix), Value: pointer.StringPtr("/")}
 			}
-			switch match.Path.Type {
-			case gatewayapi_v1alpha1.PathMatchPrefix:
-				mc.pathMatchConditions = append(mc.pathMatchConditions, &PrefixMatchCondition{Prefix: stringOrDefault(match.Path.Value, "/")})
-			case gatewayapi_v1alpha1.PathMatchExact:
-				mc.pathMatchConditions = append(mc.pathMatchConditions, &ExactMatchCondition{Path: stringOrDefault(match.Path.Value, "/")})
-			default:
-				routeAccessor.AddCondition(status.ConditionNotImplemented, metav1.ConditionTrue, status.ReasonPathMatchType, "HTTPRoute.Spec.Rules.PathMatch: Only Prefix match type and Exact match type are supported.")
+			if match.Path.Type != nil {
```
sunjayBhatia, line 295: related to previous comments about path match value and type, could this maybe use the same `stringOrDefaultFromPtr` idea instead of a nil check? the next line would be `switch typeOrDefaultFromPtr(match.Path.Type, gatewayapi_v1alpha1.PathMatchPrefix)`

would make it so we don't have to do as many nil checks above or here

internal/dag/gatewayapi_processor.go
```diff
@@ -433,3 +457,18 @@ func (p *GatewayAPIProcessor) cluster(headerPolicy *HeadersPolicy, service *Serv
...
+
+func pathMatchTypePtr(pmt gatewayapi_v1alpha1.PathMatchType) *gatewayapi_v1alpha1.PathMatchType {
+	return &pmt
+}
+
+func headerMatchTypePtr(hmt gatewayapi_v1alpha1.HeaderMatchType) *gatewayapi_v1alpha1.HeaderMatchType {
+	return &hmt
+}
+
+func stringOrDefaultFromPtr(s *string, def string) string {
```
sunjayBhatia, line 469: I think this can be replaced by using `pointer.StringDeref` from the k8s util package

internal/dag/gatewayapi_processor.go
```diff
@@ -269,32 +285,34 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
...
+					routeAccessor.AddCondition(status.ConditionNotImplemented, metav1.ConditionTrue, status.ReasonPathMatchType, "HTTPRoute.Spec.Rules.PathMatch: Only Prefix match type and Exact match type are supported.")
+				}
 			}
 
 			if match.Headers != nil {
-				switch match.Headers.Type {
-				case gatewayapi_v1alpha1.HeaderMatchExact:
-					for k, v := range match.Headers.Values {
-						mc.headerMatchCondition = append(mc.headerMatchCondition, HeaderMatchCondition{MatchType: HeaderMatchTypeExact, Name: k, Value: v})
+				if match.Headers.Type != nil {
```
sunjayBhatia, line 307: ah maybe this was missing before, what do we do if header match type is not supplied? should we default to something?

internal/dag/gatewayapi_processor.go
stevesloka: 👍

internal/dag/gatewayapi_processor.go
stevesloka: If there's no header match then there isn't a default, only if nothing is defined do we get the default prefix match on "/".

internal/dag/gatewayapi_processor.go
stevesloka: 👍

internal/dag/gatewayapi_processor.go
stevesloka: I refactored stuff a bit which I think makes the main logic read better, PTAL. =)

internal/dag/gatewayapi_processor.go
```diff
@@ -269,33 +285,23 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
...
-		matchconditions := []*matchConditions{}
+		var matchconditions []*matchConditions
 
 		for _, match := range rule.Matches {
 			mc := &matchConditions{}
-			// TODO: Replace this with nil check when gateway-api includes this fix -
-			// https://github.com/kubernetes-sigs/gateway-api/commit/9d63656df8cc9e67da60d4fe3f6289aad22e72d3
-			if match.Path.Value == "" || match.Path.Type == "" {
-				match.Path = gatewayapi_v1alpha1.HTTPPathMatch{Type: gatewayapi_v1alpha1.PathMatchPrefix, Value: "/"}
+			if match.Path == nil && match.Headers == nil {
```
sunjayBhatia, line 292: nice short circuit, +1

internal/dag/gatewayapi_processor.go
```diff
@@ -269,33 +285,23 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
...
+				// No match conditions are defined, so default to 'PrefixMatch' with a value of '/'.
+				match.Path = &gatewayapi_v1alpha1.HTTPPathMatch{Type: pathMatchTypePtr(gatewayapi_v1alpha1.PathMatchPrefix), Value: pointer.StringPtr("/")}
```
sunjayBhatia, line 294: should this modify `mc` insteadof `match`?

internal/dag/gatewayapi_processor.go
```diff
@@ -405,6 +417,47 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
...
+		return nil
+	}
+
+	if match.Type == nil {
+		// If path match type is not defined, default to 'PrefixMatch'.
+		mc.pathMatchConditions = append(mc.pathMatchConditions, &PrefixMatchCondition{Prefix: pointer.StringDeref(match.Value, "/")})
+	} else {
+		switch *match.Type {
+		case gatewayapi_v1alpha1.PathMatchPrefix:
+			mc.pathMatchConditions = append(mc.pathMatchConditions, &PrefixMatchCondition{Prefix: pointer.StringDeref(match.Value, "/")})
```
sunjayBhatia, line 433: nit: could pull out the `pointer.StringDeref(match.Value, "/")` above so it is only done once

internal/dag/gatewayapi_processor.go
sunjayBhatia: I assume we have tests already for this case

internal/dag/gatewayapi_processor.go
```diff
@@ -405,6 +417,47 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha1.HTTPRo
...
+	}
+	return nil
+}
+
+func headerMatchCondition(mc *matchConditions, match *gatewayapi_v1alpha1.HTTPHeaderMatch) error {
+	if match == nil {
+		return nil
+	}
+
+	if match.Type != nil {
```
sunjayBhatia, line 448: Looks like this is being defaulted to Exact: https://github.com/kubernetes-sigs/gateway-api/blob/cbbdc4c1ba2b0468f5412477c0de2f02d4089abf/apis/v1alpha1/httproute_types.go#L315

👍🏽

internal/dag/gatewayapi_processor.go
stevesloka: We can actually remove this whole bit since now we're checking for nil in the other helper bits.

internal/dag/gatewayapi_processor.go
stevesloka: Gotcha, I updated. We're finding some good bugs in this PR. =)

skriss (approved):
LGTM, all the pointer stuff is definitely annoying but 🤷‍♂️

sunjayBhatia (approved):

youngnick (approved):

stevesloka:
Thanks for all the reviews @skriss @sunjayBhatia @youngnick! 🎉  I'll merge this now and refresh once the GatewayAPI release is out.

PR #3605, opened 2021-04-21
Remove old versions from matrix and add main row

sunjayBhatia:
So we can track main, make adding a new release easier
- When we bump Envoy or k8s supported versions we can bump this

Chopped off table up to previous year, we can limit this more if we think we should


skriss (approved):
LGTM, yeah idk exactly where to draw the line on the old versions but this seems like a good start.

youngnick (approved):
Yep, this looks great to me.

PR #3611, opened 2021-04-22
Downstream TLS: Allow Skip Verify of client cert

### LINKED ISSUE #3253: Support a TLS Client Authentication policy: request client certificate but not validate it or request client certificate and proceed even though validation fails
Hi @youngnick @jpeach 

**Problem Statement**
We import client certificates into the secret `(HTTPProxy: spec->tls->clientValidation->caSecret)` so that the envoy can validate the client certificate against the caSecret.  But because of the limit of k8s secret, we can only import around limited client certificates (around 500-1000 certificates) into the caSecret. Currently, we have more than 5k hosts with self-signed certificates that need to connect to the ingress entrypoint. Because of some reasons, we can't use a CA and issue a ca-signed certificate for each host. We can only use a self-signed certificate for each host.

**Solution**
We could use external auth for client certificate validation. The external auth server gets the client certificate from envoy and could check it against as many secrets as it needs to.

**Feature Request**
for contour v1.11.0, if we don’t specify the `clientValidation` part in HTTPProxy spec, the envoy will not request the clients certificate. If we give the `clientValidation` part in HTTPProxy spec, then the envoy will request the client certificate and validate it against the caSecret.

To implement the above solution, we need that envoy can request the client certificate but not validate it or request the client certificate and proceed even though validation fails so that the envoy can pass the client certificate to external auth server.

Thanks!

sunjayBhatia:
Enables users to configure Envoy to not verify client certs.
This can be used in conjunction with External Authorization as
Envoy will still require client certs are supplied and will pass
them along to an ExtAuthz server if configured.
Contour does not provision Envoy with a CA for client CA validation
if downstream validation is skipped.

Also adds a Gatekeeper template and constraint policy for operators
who want to limit HTTPProxies that try to use this setting in case
it is desired to restrict it.

Fixes #3253


stevesloka (commented):
Overall this looks good to me, just worried that the dependency on gatekeeper will be valid for all users since this introduces a field which could create an insecure HTTPProxy.

sunjayBhatia:
> Overall this looks good to me, just worried that the dependency on gatekeeper will be valid for all users since this introduces a field which could create an insecure HTTPProxy.

Ah interesting, do you think a better path would be to add a Contour config file field so an operator could disable skip verify (require cert verification to undo the double negative) globally instead

stevesloka:
> Ah interesting, do you think a better path would be to add a Contour config file field so an operator could disable skip verify (require cert verification to undo the double negative) globally instead

I don't know what users are ok with. Just ends up being a large ask for folks when upgrading. The only other path (to get feedback but still implement), is to add a "features" flag to enable globally, then use what you have in this PR. Then after a while can remove that feature flag when folks are sure of the changes that are included. 

Not sure if I'm over thinking this though.

sunjayBhatia:
> > Ah interesting, do you think a better path would be to add a Contour config file field so an operator could disable skip verify (require cert verification to undo the double negative) globally instead
> 
> I don't know what users are ok with. Just ends up being a large ask for folks when upgrading. The only other path (to get feedback but still implement), is to add a "features" flag to enable globally, then use what you have in this PR. Then after a while can remove that feature flag when folks are sure of the changes that are included.
> 
> Not sure if I'm over thinking this though.

right now I don't think there is anything in Contour that ensures all HTTPProxies have downstream TLS validation set so this new field is similar to that when in use, except Envoy will require a client cert is sent as well

if operators already have a mechanism to restrict HTTPProxies that do not have downstream TLS settings this would fall into the same category

youngnick:
I think that, in this case, having Gatekeeper as a requirement is okay because users have to specify "make this insecure by bypassing certs", rather than it being insecure by default. All we can do is make sure we've warned people in the field docs.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -785,9 +785,16 @@ type UpstreamValidation struct {
...
 	// The client certificate must validate against the certificates in the bundle.
-	// +kubebuilder:validation:Required
+	// +optional
 	// +kubebuilder:validation:MinLength=1
 	CACertificate string `json:"caSecret"`
+
+	// SkipClientCertValidation disables downstream client certificate
+	// validation. Defaults to false. When external authorization is in use
+	// and is set to true, the downstream certificate is still passed to the
+	// external authorization server.
```
youngnick, line 795: I think there should be something here about how this is intended to be used with external auth, and that turning this on means that your client certs add no security value in themselves.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -241,8 +241,8 @@ type TLS struct {
 	// This setting:
 	//
 	// 1. Enables TLS client certificate validation.
-	// 2. Requires clients to present a TLS certificate (i.e. not optional validation).
-	// 3. Specifies how the client certificate will be validated.
+	// 2. Specifies how the client certificate will be validated (i.e.
+	//    validation required or optional).
```
youngnick, line 245: I think we need a note here saying that if you make the client certificate validation optional, you should be using external auth as well, or there's no point enabling it.

youngnick (approved):
Overall LGTM, but I think some doc changes to make the security risks of using this a bit clearer would be good.

skriss:
Per https://github.com/projectcontour/contour/issues/3253#issuecomment-760310714 do we need to set `include_peer_certificate: true` for the ext auth filter?

sunjayBhatia:
> Per [#3253 (comment)](https://github.com/projectcontour/contour/issues/3253#issuecomment-760310714) do we need to set `include_peer_certificate: true` for the ext auth filter?

We're currently setting on the ext auth filter always which was nice: https://github.com/projectcontour/contour/blob/4fbf6550553ddd99ace7ffaf4d6cf612bdb2f0c6/internal/envoy/v3/listener.go#L632

skriss:
Ah, :+1:

skriss (commented):
I think this looks good, it's a valid point that a user can already just not configure client validation and be "insecure" so to me it doesn't seem like we're adding risk with this new field.

skriss (approved):
Approving but it'd be good to make sure @stevesloka is OK with this too before merge

PR #3612, opened 2021-04-22
internal/dag: Filter on Gateway.Listener.Protocol for supported types

### LINKED ISSUE #3540: Filter on Gateway.Listener.Protocol
Need to filter on supported protocol types for any Gateway.Listener.Protocol. Supported types should be:

- HTTP
- HTTPS
- TLS

stevesloka:
Filters the Protocol on gateway listeners for only valid types (e.g. HTTP, HTTPS, TLS).

Fixes #3540

### DIFF
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -1276,39 +1276,6 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 				},
 			),
 		},
-		//"Only the Spec.Listener.Protocol: HTTPS should be valid for a TLS Listener Gateway": {
-		//	gateway: &gatewayapi_v1alpha1.Gateway{
-		//		ObjectMeta: metav1.ObjectMeta{
-		//			Name:      "contour",
-		//			Namespace: "projectcontour",
-		//		},
-		//		Spec: gatewayapi_v1alpha1.GatewaySpec{
-		//			Listeners: []gatewayapi_v1alpha1.Listener{{
-		//				Port:     443,
-		//				Protocol: gatewayapi_v1alpha1.HTTPProtocolType,
-		//				TLS: &gatewayapi_v1alpha1.GatewayTLSConfig{
-		//					CertificateRef: &gatewayapi_v1alpha1.LocalObjectReference{
-		//						Group: "core",
-		//						Kind:  "Secret",
-		//						Name:  sec1.Name,
-		//					},
-		//				},
-		//				Routes: gatewayapi_v1alpha1.RouteBindingSelector{
-		//					Kind: KindHTTPRoute,
-		//					Namespaces: gatewayapi_v1alpha1.RouteNamespaces{
-		//						From: gatewayapi_v1alpha1.RouteSelectAll,
-		//					},
-		//				},
-		//			}},
-		//		},
-		//	},
-		//	objs: []interface{}{
-		//		sec1,
-		//		blogService,
-		//		httpRouteProtocolHTTPS,
-		//	},
-		//	want: listeners(),
-		//},
 		"TLS Listener Gateway CertificateRef must be type core.Secret": {
 			gateway: &gatewayapi_v1alpha1.Gateway{
 				ObjectMeta: metav1.ObjectMeta{
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -1159,7 +1159,7 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 				},
 			),
 		},
-		"insert basic single route, single hostname, gateway with TLS": {
+		"insert basic single route, single hostname, gateway with TLS, HTTP protocol is ignored": {
 			gateway: &gatewayapi_v1alpha1.Gateway{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      "contour",
@@ -1168,7 +1168,7 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 				Spec: gatewayapi_v1alpha1.GatewaySpec{
 					Listeners: []gatewayapi_v1alpha1.Listener{{
 						Port:     443,
-						Protocol: gatewayapi_v1alpha1.HTTPSProtocolType,
+						Protocol: gatewayapi_v1alpha1.HTTPProtocolType,
 						TLS: &gatewayapi_v1alpha1.GatewayTLSConfig{
 							CertificateRef: &gatewayapi_v1alpha1.LocalObjectReference{
 								Group: "core",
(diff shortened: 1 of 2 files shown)


internal/dag/builder_test.go
```diff
@@ -1160,6 +1160,39 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
...
 		"insert basic single route, single hostname, gateway with TLS": {
+			gateway: &gatewayapi_v1alpha1.Gateway{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "contour",
+					Namespace: "projectcontour",
+				},
+				Spec: gatewayapi_v1alpha1.GatewaySpec{
+					Listeners: []gatewayapi_v1alpha1.Listener{{
+						Port:     443,
+						Protocol: gatewayapi_v1alpha1.HTTPProtocolType, // <--- invalid protocol, must be "HTTPS"
```
sunjayBhatia, line 1171: is this comment removable/movable to the below test case?

internal/dag/gatewayapi_processor.go
```diff
@@ -71,8 +71,23 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+		if listener.Protocol != gatewayapi_v1alpha1.HTTPProtocolType &&
+			listener.Protocol != gatewayapi_v1alpha1.HTTPSProtocolType &&
+			listener.Protocol != gatewayapi_v1alpha1.TLSProtocolType {
+			p.Errorf("Listener.Protocol %q is not supported.", listener.Protocol)
+			continue
+		}
+
 		// Check for TLS on the Gateway.
 		if listener.TLS != nil {
+			if listener.Protocol != gatewayapi_v1alpha1.HTTPSProtocolType &&
```
sunjayBhatia, line 84: only point of contention for this PR might be if this is the right choice

https://github.com/kubernetes-sigs/gateway-api/blob/e3483c5bdf32f2468ca3c17cf29a988808c132fd/apis/v1alpha1/gateway_types.go#L187-L189 says "ignored otherwise" which could be interpreted as the opposite of this

internal/dag/builder_test.go
sunjayBhatia: or maybe the test case names got mixed up? looking at the full file it looks like this test case is actually `"insert basic single route, single hostname, gateway with TLS, invalid listener protocol"`

internal/dag/gatewayapi_processor.go
skriss: Yeah that does sound like we should still process routes for the listener.

stevesloka: 👍

internal/dag/builder_test.go
stevesloka: I'll remove the comment since now the protocol is ignored in relation to if tls is defined or not.

internal/dag/builder_test.go
```diff
@@ -1160,6 +1160,39 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
...
+				},
+			},
+			objs: []interface{}{
+				sec1,
+				kuardService,
+				genericHTTPRoute,
+			},
+			want: listeners(),
+		},
+		"insert basic single route, single hostname, gateway with TLS, invalid listener protocol": {
```
skriss, line 1195: Per sunjay's comment seems like this test should still be called "insert basic single route, single hostname, gateway with TLS" since nothing changed with it and it uses a valid protocol, and then the one added just above should be ""insert basic single route, single hostname, gateway with TLS, invalid listener protocol" and you probably want to use a non-HTTP/HTTPS/TLS protocol value to test the protocol validation

internal/dag/gatewayapi_processor.go
```diff
@@ -71,6 +71,14 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
 		var listenerSecret *Secret
 
+		// Validate the Kind on the selector is a supported type.
+		if listener.Protocol != gatewayapi_v1alpha1.HTTPProtocolType &&
+			listener.Protocol != gatewayapi_v1alpha1.HTTPSProtocolType &&
+			listener.Protocol != gatewayapi_v1alpha1.TLSProtocolType {
+			p.Errorf("Listener.Protocol %q is not supported.", listener.Protocol)
+			continue
+		}
+
```
skriss, line 81: based on the API docs Sunjay linked, it sounds like we should validate that if the protocol is HTTPS/TLS, that `listener.TLS` is non-nil (but if the protocol is something else, we just ignore `listener.TLS`).

skriss (changes requested):

internal/dag/gatewayapi_processor.go
stevesloka: Ok I'll do that and open an issue upstream. This protocol specification doesn't seem clear to me. If you specify on an HTTP protocol, just ignore, but no error if you don't specify on an HTTPS protocol? Right now this just uses if it's there and ignores if not.

stevesloka: The other unclear bit is if we specify TLS on the HTTPRoute (which isn't yet added to Contour), what happens then? 

// ref: https://gateway-api.sigs.k8s.io/references/spec/#networking.x-k8s.io/v1alpha1.RouteTLSConfig

skriss: Sounds like in that case `listener.TLS` still needs to be specified with a mode of `Passthrough`

skriss (approved):
LGTM pending CI.

internal/dag/builder_test.go
```diff
@@ -1159,6 +1159,73 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
...
 		},
+		"insert basic single route, single hostname, gateway with TLS, HTTP protocol is ignored": {
+			gateway: &gatewayapi_v1alpha1.Gateway{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "contour",
+					Namespace: "projectcontour",
+				},
+				Spec: gatewayapi_v1alpha1.GatewaySpec{
+					Listeners: []gatewayapi_v1alpha1.Listener{{
+						Port:     443,
```
sunjayBhatia, line 1170: just noticed this, we're asking for a HTTP listener on port 443 and we end up getting one on port 80

sunjayBhatia: probably b/c of this https://github.com/projectcontour/contour/blob/9e7b4e6aad08be068e5566b757f8f1a76a6a80ba/internal/dag/gatewayapi_processor.go#L403-L406 but it is a bit counterintuitive

maybe so it is not confusing just change the port in the test to 80 for now and we can figure out what to do with the whole mismatched/one port thing separately

stevesloka: The ports aren't used right now from Gateway listeners.

stevesloka: #3616

sunjayBhatia: ah gotcha, that makes sense

this test may fail when we resolve the issue you created, but sounds good for now, HTTP protocol always gets you port 80

sunjayBhatia (approved):

PR #3618, opened 2021-04-26
internal/dag: Process HTTPRoute.Spec.Gateways

### LINKED ISSUE #3615: Gateway API: HTTPRoute spec.gateways is not cared
HTTPRoute has spec.gateways field:

https://github.com/kubernetes-sigs/gateway-api/blob/01b86b06e26c435996b71a90fa35122616f4f731/apis/v1alpha1/httproute_types.go#L57

but Contour does not care the field.

It is not urgent issue to me as Gateway has `spec.listners.routes.selector` which can map to HTTPRoutes and Gateway.

stevesloka:
Implement support for HTTPRoutes.Spec.Gateways to allow an HTTPRoute to define how it
binds to a Gateway via "All", "SameNamespace", or "FromList".

Fixes #3615

### DIFF
--- a/_integration/testsuite/gatewayapi/007-httproute-gatewayallowtype.yaml
+++ b/_integration/testsuite/gatewayapi/007-httproute-gatewayallowtype.yaml
@@ -176,7 +176,7 @@ check_for_status_code [msg] {
 }
 
 check_for_service_routing [msg] {
-  msg := expect.response_service_is(Response, "echo-slash-default")
+  msg := expect.response_service_is(Response, "echo-slash-blue")
 }
 
 ---
--- a/_integration/testsuite/gatewayapi/007-httproute-gatewayallowtype.yaml
+++ b/_integration/testsuite/gatewayapi/007-httproute-gatewayallowtype.yaml
@@ -161,7 +161,7 @@ import data.contour.http.client
 import data.contour.http.client.url
 import data.contour.http.expect
 
-# Ensure / request returns 200 status code since
+# Ensure /blue request returns 200 status code since
 # it matches the gatewayRef.
 Response := client.Get({
   "url": url.http("/blue"),
@@ -208,7 +208,7 @@ import data.contour.http.client
 import data.contour.http.client.url
 import data.contour.http.expect
 
-# Ensure / request returns 404 status code since
+# Ensure /green request returns 404 status code since
 # the HTTPRoute does not match the gatewayRef.
 Response := client.Get({
   "url": url.http("/green"),
--- a/_integration/testsuite/gatewayapi/007-httproute-gatewayallowtype.yaml
+++ b/_integration/testsuite/gatewayapi/007-httproute-gatewayallowtype.yaml
@@ -1,0 +1,247 @@
+# Copyright Project Contour Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.  You may obtain
+# a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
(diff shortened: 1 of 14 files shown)


stevesloka:
bah these integration tests, will look into why they are failing in CI. Works on my machine. ™️

youngnick (commented):
This is lookin great, thanks Steve.

Do we need to mention anything about the default value of `spec.gateways`? I can't remember how it plays into these interactions.

stevesloka:
> Do we need to mention anything about the default value of spec.gateways? I can't remember how it plays into these interactions

The default value is `SameNamespace`, so users would need to modify their HTTPRoutes with `All` or `FromList` to work in namespaces outside the one in which the Gateway resides.

_integration/testsuite/gatewayapi/007-httproute-gatewayallowtype.yaml
```diff
@@ -0,0 +1,268 @@
...
+      - serviceName: echo-slash-blue
+        port: 80
+
+---
+
+import data.contour.http.client
+import data.contour.http.client.url
+import data.contour.http.expect
+
+# Ensure / request returns 200 status code since
```
skriss, line 164:
```suggestion
# Ensure /blue request returns 200 status code since
```

skriss: or just remove the path from the comment

_integration/testsuite/gatewayapi/007-httproute-gatewayallowtype.yaml
```diff
@@ -0,0 +1,268 @@
...
+        - serviceName: echo-slash-green
+          port: 80
+
+---
+
+import data.contour.http.client
+import data.contour.http.client.url
+import data.contour.http.expect
+
+# Ensure / request returns 404 status code since
```
skriss, line 211:
```suggestion
# Ensure /green request returns 404 status code since
```

Or just remove the path from the comment

skriss (changes requested):
couple nits but otherwise looks good

_integration/testsuite/gatewayapi/007-httproute-gatewayallowtype.yaml
```diff
@@ -0,0 +1,268 @@
...
+kind: Service
+metadata:
+  name: ingress-conformance-echo
+$apply:
+  fixture:
+    as: echo-slash-green
+
+---
+
+apiVersion: apps/v1
```
sunjayBhatia, line 56: probably not needed since we're moving to the go framework, but maybe we shouldve made the fixture able to set a namespace as well as name of the resource

internal/dag/gatewayapi_processor.go
```diff
@@ -114,7 +114,9 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
 				p.Errorf("error validating routes against Listener.Routes.Selector: %s", err)
 			}
 
-			if selMatches && nsMatches {
+			// If all the match criteria for this HTTPRoute match the Gateway, then add
+			// the route to the set of matchingRoutes.
+			if selMatches && nsMatches && p.gatewayMatches(route) {
```
sunjayBhatia, line 119: the `RouteGateways` type mentions:

```
RouteGateways defines which Gateways will be able to use a route. If this field results in preventing the selection of a Route by a Gateway, an “Admitted” condition with a status of false must be set for the Gateway on that Route.
```

Seems like we should set status of "Admitted=false" on Routes that are selected by a Gateway but the Route does not "allow" the Gateway to select it

See https://gateway-api.sigs.k8s.io/references/spec/#networking.x-k8s.io/v1alpha1.RouteGateways

stevesloka: Hmm good catch. We'll need to see if they are defined now to know if they matched by default vs matching the selectors. Let me write up some more tests.

stevesloka: Ok just pushed a new commit @sunjayBhatia. =)

youngnick (approved):
LGTM, nice.

skriss (approved):

sunjayBhatia (approved):

PR #3620, opened 2021-04-28
internal: Refactor HTTPRoute status into a generic form

stevesloka:
Refactors the HTTPRoute status logic to be generic xRoute status which will
allow us to set status on more than just HTTPRoutes.

Signed-off-by: Steve Sloka <slokas@vmware.com>

Issue #3440, opened 2021-03-03
Support TLSRoute


stevesloka:
> The TLSRoute resource is similar to TCPRoute, but can be configured to match against TLS-specific metadata. This allows more flexibility in matching streams for a given TLS listener.

> If you need to forward traffic to a single target for a TLS listener, you could choose to use a TCPRoute with a TLS listener.

//ref: https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.TLSRoute

xaleeks:
Is this part of the v1.15 in any way? We can always push it out to next release if needed. tagging it first

youngnick:
I'd rate this as nice-to-have for 1.15, rather than must-have.

sunjayBhatia:
Seems like this is going to move to 1.16?

youngnick:
Agreed.

youngnick:
Some support implemented in 1.16, but completion will need to be in 1.17.

youngnick:
Hmm, this is partially implemented, so I'll move it to 1.18.

skriss:
@stevesloka I can take this over if that works for you, looks like still several TODOs

skriss:
I'll see if there's an opportunity for some reuse between `HTTPRoute` and `TLSRoute` processing as well since there's a lot of overlap in functionality.

stevesloka:
Go for it! =)

skriss:
Task list is complete so I'm closing this out, can open new issues if we find any other gaps.

PR #3627, opened 2021-04-30
internal: Adds support for TLSRoute

stevesloka:
Add support for TLSRoute to enable Passthrough TCP Proxying to pods via SNI.

Note: This is currently rebased from https://github.com/projectcontour/contour/pull/3620 which makes the Status bits more generic. Merging that first will make this PR smaller. 

Updates #3440

### DIFF
--- a/_integration/testsuite/gatewayapi/008-tlsroute.yaml
+++ b/_integration/testsuite/gatewayapi/008-tlsroute.yaml
@@ -1,0 +1,187 @@
+# Copyright Project Contour Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.  You may obtain
+# a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+  import data.contour.resources
+
+  # Ensure that cert-manager is installed.
+  # Version check the certificates resource.
+
+  Group := "cert-manager.io"
+  Version := "v1"
+
+  have_certmanager_version {
+  v := resources.versions["certificates"]
+  v[_].Group == Group
+  v[_].Version == Version
+}
+
+  skip[msg] {
+  not resources.is_supported("certificates")
+  msg := "cert-manager is not installed"
+}
+
+  skip[msg] {
+  not have_certmanager_version
+
+  avail := resources.versions["certificates"]
+
+  msg := concat("\n", [
+  sprintf("cert-manager version %s/%s is not installed", [Group, Version]),
+  "available versions:",
+  yaml.marshal(avail)
+  ])
+}
+
+---
+
+apiVersion: cert-manager.io/v1
+kind: ClusterIssuer
+metadata:
+  name: selfsigned
+spec:
+  selfSigned: {}
+
+---
+
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: backend-server-cert
+spec:
+  dnsNames:
+    - tcp.projectcontour.io
+  secretName: backend-server-cert
+  issuerRef:
+    name: selfsigned
+    kind: ClusterIssuer
+
+---
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: ingress-conformance-echo-tls
+$apply:
+  fixture:
+    as: echo-slash-blue
+
+---
+
+apiVersion: v1
+kind: Service
+metadata:
+  name: ingress-conformance-echo-tls
+$apply:
+  fixture:
+    as: echo-slash-blue
+
+---
+
+apiVersion: networking.x-k8s.io/v1alpha1
+kind: GatewayClass
+metadata:
+  name: contour-class
+spec:
+  controller: projectcontour.io/ingress-controller
+
+---
+
+apiVersion: networking.x-k8s.io/v1alpha1
+kind: Gateway
+metadata:
+  name: contour
+  namespace: projectcontour
+spec:
+  gatewayClassName: contour-class
+  listeners:
+    - protocol: TLS
+      port: 443
+      routes:
+        kind: TLSRoute
+        namespaces:
+          from: "All"
+---
+
+apiVersion: networking.x-k8s.io/v1alpha1
+kind: TLSRoute
+metadata:
+  name: tlsroute1
+spec:
+  rules:
+  - matches:
+    - snis:
+      - tcp.projectcontour.io
+    forwardTo:
+    - serviceName: echo-slash-blue
+      port: 443
+---
+
+import data.contour.http.client
+import data.contour.http.client.url
+import data.contour.http.expect
+
+# Ensure / request returns 200 status code.
+Response := client.Get({
+  "url": url.https("/"),
+    "headers": {
+      "Host": "tcp.projectcontour.io",
+      "User-Agent": client.ua("tlsroute"),
+  },
+  "tls_insecure_skip_verify": true,
+})
+
+check_for_status_code [msg] {
+  msg := expect.response_status_is(Response, 200)
+}
+
+check_for_service_routing [msg] {
+  msg := expect.response_service_is(Response, "echo-slash-blue")
+}
+
+---
+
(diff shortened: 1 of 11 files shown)


internal/dag/gatewayapi_processor.go
```diff
@@ -150,15 +152,47 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+			// RouteBindingSelector defines a schema for associating routes with the Gateway.
+			// If Namespaces and Selector are defined, only routes matching both selectors are associated with the Gateway.
+
+			// ## RouteBindingSelector ##
+			//
+			// Selector specifies a set of route labels used for selecting routes to associate
+			// with the Gateway. If this Selector is defined, only routes matching the Selector
+			// are associated with the Gateway. An empty Selector matches all routes.
+
+			nsMatches, err := p.namespaceMatches(listener.Routes.Namespaces, route.Namespace)
```
sunjayBhatia, line 171: should we add some route matching/namespace matching tests for tlsroutes as well?

internal/featuretests/v3/tlsroute_test.go
```diff
@@ -0,0 +1,170 @@
...
+			Rules: []gatewayapi_v1alpha1.TLSRouteRule{{
+				ForwardTo: []gatewayapi_v1alpha1.RouteForwardTo{{
+					ServiceName: pointer.StringPtr("correct-backend"),
+					Port:        gatewayPort(80),
+				}},
+			}},
+		},
+	}
+
+	rh.OnUpdate(route1, route2)
```
sunjayBhatia, line 138: maybe add a test for one rule with no SNI and another with an SNI?

_integration/testsuite/gatewayapi/008-tlsroute.yaml
```diff
@@ -0,0 +1,187 @@
...
+
+---
+
+apiVersion: networking.x-k8s.io/v1alpha1
+kind: TLSRoute
+metadata:
+  name: tlsroute1
+spec:
+  rules:
+    - forwardTo:
```
sunjayBhatia, line 161: how does it work if we have both no SNI an SNI together? I think Envoy will choose the most specific so they should be able to coexist

https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener_components.proto#config-listener-v3-filterchainmatch

```
For criteria that allow ranges or wildcards, the most specific value in any of the configured filter chains that matches the incoming connection is going to be used (e.g. for SNI www.example.com the most specific match would be www.example.com, then *.example.com, then *.com, then any filter chain without server_names requirements).
```

internal/featuretests/v3/tlsroute_test.go
```diff
@@ -0,0 +1,170 @@
...
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/pointer"
+	gatewayapi_v1alpha1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
+)
+
+func TestTLSRoute(t *testing.T) {
+	rh, c, done := setup(t)
+	defer done()
+
+	//s1 := &v1.Secret{
```
sunjayBhatia, line 35: the docs mention that a secret is required on TLSRoutes, seems like the integration test and this one have omitted it and not made the TLS secret required?

sunjayBhatia: is this because this just enables passthrough TLS and not terminating and re-encrypting?

sunjayBhatia: what are the rules that we/upstream decided on for supporting TLS passthrough/termination and how they relate to TLSRoute/TCPRoute?

internal/dag/gatewayapi_processor.go
```diff
@@ -315,11 +354,106 @@ func selectorMatches(selector *metav1.LabelSelector, objLabels map[string]string
...
+			hosts = []string{"*"}
+		}
+
+		if len(rule.ForwardTo) == 0 {
+			routeAccessor.AddCondition(status.ConditionResolvedRefs, metav1.ConditionFalse, status.ReasonDegraded, "At least one Spec.Rules.ForwardTo must be specified.")
+			continue
+		}
+
+		var proxy TCPProxy
+		for _, forward := range rule.ForwardTo {
```
sunjayBhatia, line 404: maybe `ForwardTo` validation could be abstracted to a common helper?

stevesloka: 👍

internal/featuretests/v3/tlsroute_test.go
stevesloka: There were some discussions upstream (https://github.com/kubernetes-sigs/gateway-api/pull/633#discussion_r622203260) and essentially from what I understand, `TLSRoute` is a passthrough TCPProxy against SNI. 

If you want to terminate then route, you'd have to use TCPRoute. 

@youngnick said we can't support TCPRoute so this will be a point that differs from HTTPProxy. 

- HTTPProxy.TCPProxy (passthrough: true) == TLSRoute
- HTTPProxy.TCPProxy (passthrough: false) == TCPRoute

_integration/testsuite/gatewayapi/008-tlsroute.yaml
stevesloka: This is a great point, there probably should be some validation higher up across all the listeners to check that only one wildcard (e.g. '*') can exist for the same port. 

Additionally, we need to validate that the same domain isn't used for both TLSRoute as well as an HTTPRoute.

internal/featuretests/v3/tlsroute_test.go
stevesloka: 👍

internal/dag/gatewayapi_processor.go
stevesloka: I opened https://github.com/projectcontour/contour/pull/3684 to cover tests. Was trying to reduce the review sizes and was already there (also found a bug). =)

sunjayBhatia (approved):
I think this is all done? I think all my previous comments are resolved

internal/dag/gatewayapi_processor.go
```diff
@@ -72,12 +73,13 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
-		var matchingRoutes []*gatewayapi_v1alpha1.HTTPRoute
+		var matchingHTTPRoutes []*gatewayapi_v1alpha1.HTTPRoute
+		var matchingTLSRoutes []*gatewayapi_v1alpha1.TLSRoute
 		var listenerSecret *Secret
 
 		// Validate the Kind on the selector is a supported type.
 		switch listener.Protocol {
-		case gatewayapi_v1alpha1.HTTPSProtocolType, gatewayapi_v1alpha1.TLSProtocolType:
+		case gatewayapi_v1alpha1.HTTPSProtocolType:
 			// Validate that if protocol is type HTTPS or TLS that TLS is defined.
```
skriss, line 83:
```suggestion
			// Validate that if protocol is type HTTPS that TLS is defined.
```

internal/dag/gatewayapi_processor.go
```diff
@@ -150,15 +152,47 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+			if err != nil {
+				p.Errorf("error validating namespaces against Listener.Routes.Namespaces: %s", err)
+			}
+
+			selMatches, err := selectorMatches(listener.Routes.Selector, route.Labels)
+			if err != nil {
+				p.Errorf("error validating routes against Listener.Routes.Selector: %s", err)
+			}
+
+			if selMatches && nsMatches {
```
skriss, line 181: Is [this block of logic](https://github.com/projectcontour/contour/blob/e3cf805af778fe42ec39681dbf71b783b091007e/internal/dag/gatewayapi_processor.go#L138-L157) that we have for `HTTPRoutes` not needed for `TLSRoutes`?

stevesloka: Yup it is to satisfy the API. I think I just didn't add it yet for this PR. Is it easier to merge this with out (assuming the rest is ok), then follow up with a new PR to add that with corresponding tests?

internal/dag/gatewayapi_processor.go
stevesloka: 👍

internal/dag/gatewayapi_processor.go
skriss: OK with me

skriss (approved):
LGTM pending CI

Issue #3621, opened 2021-04-29
Migrate E2E tests to Go


skriss:
- [x] get basic framework defined and running in CI
- [x] implement HTTPProxy tests in Go
- [x] implement Gateway API tests in Go
- [x] implement Ingress tests in Go
- [x] turn off `_integration` tests (#3737)
- [ ] remove numbering from tests (leaving in for now for ease of correlating to the `_integration` tests) (#3838)

Future:
- improvements to how test data (`HTTPProxies`, `Gateways`, `HTTPRoutes`, etc) are defined (YAML? Builders? etc)
- overlap with `featuretests` - can we reuse test cases, helper libs, etc?
- have test framework generate unique test name & namespace & pass an abstraction to tests that "knows" to create stuff in the right place (https://github.com/projectcontour/contour/pull/3622#discussion_r623232987, https://github.com/projectcontour/contour/pull/3622#discussion_r623234704)
- make setup faster (create cluster, build Contour image, load images into cluster, deploy Contour to cluster)
- ~~get HTTPProxy test 013-backend-tls migrated & running (requires an alternate Contour config file) (#3729)~~
- ~~look at Ginkgo (in-progress https://github.com/projectcontour/contour/pull/3638)~~
- ~~improvements to how subtests are registered to be run~~ (covered by move to Ginkgo)
- ~~support different Contour configurations for different sets of tests~~ (covered by running Contour out-of-cluster)

skriss:
@stevesloka @sunjayBhatia @youngnick I'm planning to turn the `_integration` tests off around the end of this month/release cycle, let me know if you have any objections.

youngnick:
As far as I can see, we have coverage over the same things, so that sounds great.

sunjayBhatia:
Relevant issue as well, to track the backend-tls test that we dont have a go e2e for yet: https://github.com/projectcontour/contour/issues/3729

skriss:
We're at a point where we can remove the numbering from the test file names and namespace names, I'll get a PR ready for that so we can close this issue out. cc @sunjayBhatia

PR #3630, opened 2021-04-30
test/e2e: add Gateway API tests 001..005

skriss:
Re-implements Gateway API tests 001..005 from
_integration/testsuite/gatewayapi as Go tests.

Updates #3621.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -59,3 +59,6 @@ issues:
   - linters:
     - golint
     text: "don't use ALL_CAPS in Go names; use CamelCase"
+  - path: test/e2e
+    linters:
+      - bodyclose
--- a/test/e2e/gateway/003_invalid_forward_to_test.go
+++ b/test/e2e/gateway/003_invalid_forward_to_test.go
@@ -1,0 +1,167 @@
+// Copyright Project Contour Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build e2e
+
+package gateway
+
+import (
+	"testing"
+
+	"github.com/projectcontour/contour/test/e2e"
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	gatewayv1alpha1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
+)
+
+func testInvalidForwardTo(t *testing.T, fx *e2e.Framework) {
+	namespace := "gateway-003-invalid-forward-to"
+
+	fx.CreateNamespace(namespace)
+	defer fx.DeleteNamespace(namespace)
+
+	fx.Fixtures.Echo.Deploy(namespace, "echo-slash-default")
+
+	route := &gatewayv1alpha1.HTTPRoute{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: namespace,
+			Name:      "http-filter-1",
+			Labels:    map[string]string{"app": "filter"},
+		},
+		Spec: gatewayv1alpha1.HTTPRouteSpec{
+			Hostnames: []gatewayv1alpha1.Hostname{"invalidforwardto.projectcontour.io"},
+			Gateways: &gatewayv1alpha1.RouteGateways{
+				Allow: gatewayAllowTypePtr(gatewayv1alpha1.GatewayAllowAll),
+			},
+			Rules: []gatewayv1alpha1.HTTPRouteRule{
+				{
+					Matches: []gatewayv1alpha1.HTTPRouteMatch{
+						{
+							Path: &gatewayv1alpha1.HTTPPathMatch{
+								Type:  pathMatchTypePtr(gatewayv1alpha1.PathMatchPrefix),
+								Value: stringPtr("/invalidref"),
+							},
+						},
+					},
+					ForwardTo: []gatewayv1alpha1.HTTPRouteForwardTo{
+						{
+							ServiceName: stringPtr("invalid"),
+							Port:        portNumPtr(80),
+						},
+					},
+				},
+
+				{
+					Matches: []gatewayv1alpha1.HTTPRouteMatch{
+						{
+							Path: &gatewayv1alpha1.HTTPPathMatch{
+								Type:  pathMatchTypePtr(gatewayv1alpha1.PathMatchPrefix),
+								Value: stringPtr("/invalidport"),
+							},
+						},
+					},
+					ForwardTo: []gatewayv1alpha1.HTTPRouteForwardTo{
+						{
+							ServiceName: stringPtr("echo-slash-default"),
+						},
+					},
+				},
+
+				{
+					Matches: []gatewayv1alpha1.HTTPRouteMatch{
+						{
+							Path: &gatewayv1alpha1.HTTPPathMatch{
+								Type:  pathMatchTypePtr(gatewayv1alpha1.PathMatchPrefix),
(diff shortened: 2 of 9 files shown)


.golangci.yml
```diff
@@ -59,3 +59,6 @@ issues:
   - linters:
     - golint
     text: "don't use ALL_CAPS in Go names; use CamelCase"
+  - path: test/e2e
+    linters:
+      - bodyclose
```
skriss, line 64: I kept getting a bunch of false positives from this linter so I just turned it off for `test/e2e` but can revisit and explicitly ignore each occurrence if folks aren't comfortably with this. I don't think it's a big deal since it's just test code so any negative impact of not closing HTTP response bodies is limited to CI.

test/e2e/gateway/gateway_test.go
```diff
@@ -0,0 +1,192 @@
...
+	gateway *gatewayv1alpha1.Gateway
+	tests   map[string]func(t *testing.T, f *e2e.Framework)
+}
+
+// getTests defines the test groups that make up the Gateway test suite.
+// Each group shares a single Gateway definition.
+func getTests() []*testGroup {
+	var testGroups []*testGroup
+
+	testGroups = append(testGroups, &testGroup{
```
sunjayBhatia, line 39: shameless plug: this is where Ginkgo `BeforeEach` blocks in a `Describe` or `Context` would shine, so the gateway that needs to be used with a group of tests is close to the tests themselves and grouped together

Ginkgo does still end up with long test files though which is a pain, but that could be worked around with the structure we have here with helper methods that implement the tests themselves I think

test/e2e/gateway/003_invalid_forward_to_test.go
```diff
@@ -0,0 +1,167 @@
...
+			path:           "/invalidport",
+			expectResponse: 503,
+		},
+		{
+			path:           "/invalidservicename",
+			expectResponse: 503,
+		},
+	}
+
+	for _, tc := range cases {
```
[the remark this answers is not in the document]
skriss: yeah it is fairly common - did you have a particular approach in mind? there is definitely variation for each test - different request/TLS options, different expected status codes, different assertions on the response body, so I'm not sure off the top of my head what would make sense to do here.

test/e2e/gateway/003_invalid_forward_to_test.go
```diff
@@ -0,0 +1,167 @@
...
+						},
+					},
+				},
+			},
+		},
+	}
+	// The HTTPRoute won't be fully valid, so just wait for it to have any Gateway
+	// status.
+	fx.CreateHTTPRouteAndWaitFor(route, func(route *gatewayv1alpha1.HTTPRoute) bool {
+		return len(route.Status.Gateways) > 0
```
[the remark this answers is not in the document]
skriss: Yeah that makes sense, let me figure out what the status *should* be.

test/e2e/gateway/gateway_test.go
skriss: :+1: I'll spend some time looking at moving into Ginkgo so we can decide if we want to go that route. Also need to get Contour config file updating implemented, since ideally the Gateway tests would use a Gateway outside of the `projectcontour` namespace.

test/e2e/gateway/003_invalid_forward_to_test.go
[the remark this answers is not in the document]
sunjayBhatia: for full coverage of all the repeated cases im thinking we would maybe need a test case type that has full configurability over the requests being sent for each case, and fields for all the attributes that we would check (status code, path, service, etc.), maybe pointers/empty so if theyre not specified we don't bother to assert on those

mainly trying to abstract the repetition of https://github.com/projectcontour/contour/blob/5ac5e76f0f878e6ab12951c853f683565b2de5ff/test/e2e/gateway/003_invalid_forward_to_test.go#L148-L166 out but its not that much toil it seems, could be nice to be able to write the test cases and just write `checkTestCases(foo)` (name pending haha)

i don't think its critical and something we can refactor later if we need

for the more involved tests that require more specific setup, yeah maybe not worth it to force a generic helper to fit

test/e2e/gateway/gateway_test.go
skriss: Here's a branch that converts the Gateway tests to use Ginkgo (haven't bothered to convert subtests to use Gomega instead of testify):

https://github.com/skriss/contour/commit/a68d44134c37781b55791d0ca08459d7f1c7b6f7
gateway_test.go: https://github.com/skriss/contour/blob/a68d44134c37781b55791d0ca08459d7f1c7b6f7/test/e2e/gateway/gateway_test.go

sunjayBhatia: yeah i think IMO thats not bad and looks maybe easy to transition to without that much headache? i didnt do the work though so would have to ask you how you felt about it 

we could even pull the `It` blocks out into the individual test functions as well, and if needed probably even implement them as contexts themselves e.g.

```
	Describe("Insecure (Non-TLS) Gateway", func() {
		var gateway *gatewayv1alpha1.Gateway

		BeforeEach(func() {
...

		someTestName()
```

and 

```
func someTestName() {
    It("foo ...
```

or

```
func someTestName() {
    Context("foo ...
        BeforeEach(func() { ...
```

sunjayBhatia: yeah i think IMO thats not bad and looks maybe easy to transition to without that much headache? i didnt do the work though so would have to ask you how you felt about it 

we could even pull the `It` blocks out into the individual test functions as well, and if needed probably even implement them as contexts themselves e.g.

```
	Describe("Insecure (Non-TLS) Gateway", func() {
		var gateway *gatewayv1alpha1.Gateway

		BeforeEach(func() {
...

		someTestName()
```

and 

```
func someTestName() {
    It("foo ...
```

or

```
func someTestName() {
    Context("foo ...
        BeforeEach(func() { ...
```

skriss: Yeah it's easy to switch to. I'm happy to move ahead with that approach if folks like using Ginkgo. I do need to figure out running the spec in parallel - haven't done that before - but it seems doable. The other question would be if we want to switch the subtests to use gomega instead of testify? We don't have to, we can mix and match, I think it's really just up to our preferences.

sunjayBhatia: we could even eventually do:

```
// some repeated test setup helper
func withTestNamespace(ns string, body func()) {
    Context("with namespace: " + ns, func() {
        BeforeEach(func() {
            // create ns
        ...
        AfterEach(func() {
            // delete ns
        ...
        // call inner test body
        body()
```

and

```
func someTestName() {
    withTestNamespace("some-test-namespace", func() {
        Context("foo ...
            BeforeEach(func() { ...
            It("does something", func() { ...
```

test/e2e/gateway/003_invalid_forward_to_test.go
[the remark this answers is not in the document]
skriss: Addressed in https://github.com/projectcontour/contour/pull/3630/commits/be25f9df841a495f5aec963d7c5c1d938ff05ec4

test/e2e/gateway/gateway_test.go
sunjayBhatia: if we invoke tests with the ginkgo CLI there should be a flag to control running in parallel and choosing the number of test nodes, i forget if it directly translates to a go test flag or not but shouldnt be too hard to work out

skriss: WDYT about merging this as-is so we have the basic Gateway structure set up so any new/updated tests can be written there, then I'll get everything migrated over to Ginkgo in the next PR, then I'll continue with test case migrations?

sunjayBhatia: yeah that sounds good to me, don't want to derail your progress on this and other PRs

we might see some other patterns in PRs that follow that we want to have solutions for too 👍🏽

stevesloka (approved):
/lgtm 

Happy to get feature parity with the current, then can look at improvements.

stevesloka:
Will need to add a test for tlsRoute which is added in https://github.com/projectcontour/contour/pull/3627 (but not yet merged). Just wanted to track the change. =)

skriss:
@sunjayBhatia if you're OK with this as-is let's get it merged so we can continue moving fwd with migration. Follow-ups noted.

sunjayBhatia (approved):

PR #3635, opened 2021-05-04
internal/k8s: Use klog.SetLogger

### LINKED ISSUE #3043: Switch to log.SetLogger to intercept Kubernetes logs

klogv2 has the https://godoc.org/github.com/kubernetes/klog#SetLogger API, which should let us drop the vaguely sketchy output stream we use to capture  core Kubernetes log output.

sunjayBhatia:
Instead of our homegrown writer/pipe solution

Uses bombsimon/logrusr to wrap a logrus logger to implement the
logr interface.

Wraps that again to provide logging of file location using the call
depth interface that logr now provides.

Fixes #3043


stevesloka (commented):
Hey, @sunjayBhatia I tested this out by passing 10 to the log level, but I don't see any logs being streamed to the terminal. 

```
contour serve --kubernetes-debug=10 
```

sunjayBhatia:
> Hey, @sunjayBhatia I tested this out by passing 10 to the log level, but I don't see any logs being streamed to the terminal.
> 
> ```
> contour serve --kubernetes-debug=10 
> ```

huh, guess i need to add some test coverage! thanks for the check 👍🏽

internal/k8s/log_test.go
```diff
@@ -118,19 +135,26 @@ func TestMultipleLogWriterOptions(t *testing.T) {
...
+				if verbosityLevel <= logLevel {
 					assert.True(t, enabled)
+					klog.V(klog.Level(verbosityLevel)).Info("something")
+					klog.Flush()
+					assert.Eventually(t, func() bool { return len(logHook.AllEntries()) == 1 }, time.Second*1, time.Millisecond*10)
 				} else {
 					assert.False(t, enabled)
+					klog.V(klog.Level(verbosityLevel)).Info("something")
+					klog.Flush()
+					assert.Never(t, func() bool { return len(logHook.AllEntries()) > 0 }, time.Second*1, time.Millisecond*10)
```
sunjayBhatia, line 155: downside of this test change is that it takes the full 1s for each non-logging log level and verbosity combination to make sure no log lines come through

if that is unacceptable we could remove this check (makes the package take ~45s to run)

youngnick: Is 1s really necessary? Maybe we could start out with 500ms and increase if we need to?

youngnick (commented):
Love the idea in principle, 45sec to test the package seems a bit long though.

internal/k8s/log_test.go
sunjayBhatia: ha, yeah was able to get it down to 10ms, we'll see if it works in CI, it worked locally

youngnick (approved):
LGTM, nice work.

stevesloka:
boo conflicts @sunjayBhatia, can you rebase? =)

skriss:
Looks like ready to go on a rebase

skriss:
@sunjayBhatia good to merge here?

PR #3638, opened 2021-05-05
test/e2e: migrate to Ginkgo for running e2e's

skriss:
Introduces Ginkgo for running the E2E tests. Currently runs
subtests sequentially; parallelism to be re-introduced in
a subsequent PR.

Updates #3621.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,8 @@ require (
 	github.com/google/uuid v1.1.2
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
 	github.com/jetstack/cert-manager v1.3.0
+	github.com/onsi/ginkgo v1.16.1
+	github.com/onsi/gomega v1.10.2
 	github.com/prometheus/client_golang v1.9.0
 	github.com/prometheus/client_model v0.2.0
 	github.com/prometheus/common v0.15.0
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,6 @@ require (
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
 	github.com/jetstack/cert-manager v1.3.0
 	github.com/onsi/ginkgo v1.16.1
-	github.com/onsi/gomega v1.10.2
 	github.com/prometheus/client_golang v1.9.0
 	github.com/prometheus/client_model v0.2.0
 	github.com/prometheus/common v0.15.0
--- a/test/e2e/gateway/gateway_test.go
+++ b/test/e2e/gateway/gateway_test.go
@@ -19,134 +19,150 @@ import (
 	"context"
 	"testing"
 
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
 	"github.com/projectcontour/contour/test/e2e"
-	"github.com/stretchr/testify/require"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	gatewayv1alpha1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
 )
 
-type testGroup struct {
-	name    string
-	gateway *gatewayv1alpha1.Gateway
-	tests   map[string]func(t *testing.T, f *e2e.Framework)
-}
-
-// getTests defines the test groups that make up the Gateway test suite.
-// Each group shares a single Gateway definition.
-func getTests() []*testGroup {
-	var testGroups []*testGroup
-
-	testGroups = append(testGroups, &testGroup{
-		name: "default gateway",
-		tests: map[string]func(t *testing.T, f *e2e.Framework){
-			"001-path-condition-match":               testGatewayPathConditionMatch,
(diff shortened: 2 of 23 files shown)


skriss:
This is really just the bare minimum to run in Ginkgo and gives us something to iterate on. @sunjayBhatia had some additional thoughts on how to structure in https://github.com/projectcontour/contour/pull/3630#discussion_r625148868 which I think we can address in subsequent PRs.

skriss:
Gateway tests are running really slow in kubernetes n-2, not sure why

stevesloka:
@skriss I'm not sure why that error would show on 1.19 and not other versions. Could we be ratelimited or out of resources? Let me do a run locally on that version and see if it's much different.

skriss:
> @skriss I'm not sure why that error would show on 1.19 and not other versions. Could we be ratelimited or out of resources? Let me do a run locally on that version and see if it's much different.

Yeah, I'm not sure either. It seems like a pretty specific error related to validation on that `controller` field that *should* be optional. I tried it locally after seeing the issue in CI and saw the same behavior. 🤷‍♂️

skriss:
Spun off #3641, I don't think it's a blocker for this PR though.

test/e2e/gateway/gateway_test.go
```diff
@@ -19,134 +19,150 @@ import (
...
+
+	BeforeEach(func() {
+		f = e2e.NewFramework(GinkgoT())
+	})
+
+	Describe("Insecure (Non-TLS) Gateway", func() {
+		var gateway *gatewayv1alpha1.Gateway
+
+		BeforeEach(func() {
+			gateway = &gatewayv1alpha1.Gateway{
```
skriss, line 45: It doesn't really make sense to have this in a `BeforeEach`, I should just need to create the Gateway once for the subsequent set of specs. Deleting it and re-creating it for each spec isn't a problem, it's just extraneous operations. I guess my alternative is to have each unique Gateway config be its own test suite (file), and then use `BeforeSuite` in each to set up the gateway?

sunjayBhatia: we could also use `Context` or `Describe` and segment unique gateways in that, maybe the before/aftereach each does an "ensure" instead of create/delete?

test/e2e/gateway/gateway_test.go
```diff
@@ -19,134 +19,150 @@ import (
...
+								},
 							},
 						},
 					},
 				},
-			},
-		},
+			}
+
+			Expect(f.Client.Create(context.TODO(), gateway)).To(Succeed())
```
sunjayBhatia, line 72: if we're not sold on using ginkgo yet, I think we could still use testify for these, just maybe pass the `testing.T` argument as `GinkgoT()`?

test/e2e/httpproxy/httpproxy_test.go
```diff
@@ -18,28 +18,53 @@ package httpproxy
 import (
 	"testing"
 
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
```
sunjayBhatia, line 22: possibly removable if we can get rid of `Expect` usage etc.

go.mod
```diff
@@ -10,6 +10,8 @@ require (
 	github.com/google/uuid v1.1.2
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
 	github.com/jetstack/cert-manager v1.3.0
+	github.com/onsi/ginkgo v1.16.1
+	github.com/onsi/gomega v1.10.2
```
sunjayBhatia, line 14: might be possible to remove gomega given some of the other changes

test/e2e/gateway/gateway_test.go
sunjayBhatia: ha, didnt really read the code before this comment, disregard, thats what you already have ^

test/e2e/httpproxy/008_tcproute_https_termination_test.go
```diff
@@ -29,7 +28,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func testTCPRouteHTTPSTermination(t *testing.T, fx *e2e.Framework) {
+func testTCPRouteHTTPSTermination(fx *e2e.Framework) {
+	t := fx.T()
```
sunjayBhatia, line 32: maybe we could do some embedding of test function tricks to make this cleaner? can do in another PR tbh

test/e2e/gateway/gateway_test.go
skriss: Yeah, I think it's fine for now but potentially something to refine. I'll add a code comment about it.

skriss:
I removed gomega and commented about the extraneous Gateway creates/deletes.

sunjayBhatia (approved):

stevesloka (approved):

skriss:
Updated to move the Ingress tests to ginkgo as well, 🤞 for CI

skriss:
all green!

PR #3640, opened 2021-05-05
internal/dag: Implement Gateway.Listener.Hostname when processing HTTPRoutes

### LINKED ISSUE #3561: Take the Gateway.Listener.hostname into account when processing routes
```
Hostname specifies the virtual hostname to match for protocol types that define this concept. When unspecified, “”, or *, all hostnames are matched. This field can be omitted for protocols that don’t require hostname based matching.

Hostname is the fully qualified domain name of a network host, as defined by RFC 3986. Note the following deviations from the “host” part of the URI as defined in the RFC:

IP literals are not allowed.
The : delimiter is not respected because ports are not allowed.
Hostname can be “precise” which is a domain name without the terminating dot of a network host (e.g. “foo.example.com”) or “wildcard”, which is a domain name prefixed with a single wildcard label (e.g. *.example.com). The wildcard character * must appear by itself as the first DNS label and matches only a single label.
```

//ref: https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.Listener

stevesloka:
Take into account gateway.listeners.hostname into account when processing HTTPRoutes
which applies to all HTTPRoutes selected for that listener.

Fixes #3561

### DIFF
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -143,7 +143,10 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 		},
 	}
 
-	gatewayWithTLSRouteSelector := &gatewayapi_v1alpha1.Gateway{
+	hostname := gatewayapi_v1alpha1.Hostname("gateway.projectcontour.io")
+	wildcardHostname := gatewayapi_v1alpha1.Hostname("*.projectcontour.io")
+
+	gatewayWithHostname := &gatewayapi_v1alpha1.Gateway{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "contour",
 			Namespace: "projectcontour",
@@ -151,9 +154,40 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 		Spec: gatewayapi_v1alpha1.GatewaySpec{
 			Listeners: []gatewayapi_v1alpha1.Listener{{
 				Port:     80,
+				Hostname: &hostname,
 				Protocol: gatewayapi_v1alpha1.HTTPProtocolType,
 				Routes: gatewayapi_v1alpha1.RouteBindingSelector{
-					Kind: KindTLSRoute,
+					Kind: KindHTTPRoute,
+					Namespaces: &gatewayapi_v1alpha1.RouteNamespaces{
+						From: routeSelectTypePtr(gatewayapi_v1alpha1.RouteSelectSame),
+					},
+					Selector: &metav1.LabelSelector{
+						MatchLabels: map[string]string{
+							"app": "contour",
+						},
+						MatchExpressions: []metav1.LabelSelectorRequirement{{
+							Key:      "type",
+							Operator: "In",
+							Values:   []string{"controller"},
+						}},
+					},
+				},
+			}},
+		},
+	}
+
+	gatewayWithWildcardHostname := &gatewayapi_v1alpha1.Gateway{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "contour",
+			Namespace: "projectcontour",
+		},
+		Spec: gatewayapi_v1alpha1.GatewaySpec{
+			Listeners: []gatewayapi_v1alpha1.Listener{{
+				Port:     80,
+				Hostname: &wildcardHostname,
+				Protocol: gatewayapi_v1alpha1.HTTPProtocolType,
+				Routes: gatewayapi_v1alpha1.RouteBindingSelector{
+					Kind: KindHTTPRoute,
 					Namespaces: &gatewayapi_v1alpha1.RouteNamespaces{
 						From: routeSelectTypePtr(gatewayapi_v1alpha1.RouteSelectSame),
 					},
@@ -204,6 +238,35 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 					Kind: KindHTTPRoute,
 					Namespaces: &gatewayapi_v1alpha1.RouteNamespaces{
 						From: routeSelectTypePtr(gatewayapi_v1alpha1.RouteSelectSame),
+					},
+				},
+			}},
+		},
+	}
+
+	gatewayWithTLSRouteSelector := &gatewayapi_v1alpha1.Gateway{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "contour",
+			Namespace: "projectcontour",
+		},
+		Spec: gatewayapi_v1alpha1.GatewaySpec{
+			Listeners: []gatewayapi_v1alpha1.Listener{{
+				Port:     80,
+				Protocol: gatewayapi_v1alpha1.HTTPProtocolType,
+				Routes: gatewayapi_v1alpha1.RouteBindingSelector{
+					Kind: KindTLSRoute,
+					Namespaces: &gatewayapi_v1alpha1.RouteNamespaces{
+						From: routeSelectTypePtr(gatewayapi_v1alpha1.RouteSelectSame),
+					},
+					Selector: &metav1.LabelSelector{
+						MatchLabels: map[string]string{
+							"app": "contour",
+						},
+						MatchExpressions: []metav1.LabelSelectorRequirement{{
+							Key:      "type",
+							Operator: "In",
+							Values:   []string{"controller"},
+						}},
 					},
 				},
 			}},
@@ -2797,6 +2860,77 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 				},
 			},
 			want: listeners(),
+		},
+		"insert gateway listener with host": {
+			gateway: gatewayWithHostname,
+			objs: []interface{}{
+				kuardService,
+				&gatewayapi_v1alpha1.HTTPRoute{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "basic",
+						Namespace: "projectcontour",
+						Labels: map[string]string{
+							"app":  "contour",
+							"type": "controller",
+						},
+					},
+					Spec: gatewayapi_v1alpha1.HTTPRouteSpec{
+						Gateways: &gatewayapi_v1alpha1.RouteGateways{
+							Allow: gatewayAllowTypePtr(gatewayapi_v1alpha1.GatewayAllowSameNamespace),
+						},
+						Rules: []gatewayapi_v1alpha1.HTTPRouteRule{{
+							Matches:   httpRouteMatch(gatewayapi_v1alpha1.PathMatchExact, "/blog"),
+							ForwardTo: httpRouteForwardTo("kuard", 8080, 1),
+						}},
+					},
+				},
+			},
+			want: listeners(
+				&Listener{
+					Port: 80,
+					VirtualHosts: virtualhosts(
+						virtualhost("gateway.projectcontour.io",
+							exactrouteHTTPRoute("/blog", service(kuardService))),
+					),
+				},
+			),
+		},
+		"insert gateway listener with host, httproute with host": {
+			gateway: gatewayWithWildcardHostname,
+			objs: []interface{}{
+				kuardService,
+				&gatewayapi_v1alpha1.HTTPRoute{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "basic",
+						Namespace: "projectcontour",
+						Labels: map[string]string{
+							"app":  "contour",
+							"type": "controller",
+						},
+					},
+					Spec: gatewayapi_v1alpha1.HTTPRouteSpec{
+						Gateways: &gatewayapi_v1alpha1.RouteGateways{
+							Allow: gatewayAllowTypePtr(gatewayapi_v1alpha1.GatewayAllowSameNamespace),
+						},
+						Hostnames: []gatewayapi_v1alpha1.Hostname{
+							"http.projectcontour.io",
+						},
+						Rules: []gatewayapi_v1alpha1.HTTPRouteRule{{
+							Matches:   httpRouteMatch(gatewayapi_v1alpha1.PathMatchExact, "/blog"),
+							ForwardTo: httpRouteForwardTo("kuard", 8080, 1),
+						}},
+					},
+				},
+			},
+			want: listeners(
+				&Listener{
+					Port: 80,
+					VirtualHosts: virtualhosts(
+						virtualhost("http.projectcontour.io",
+							exactrouteHTTPRoute("/blog", service(kuardService))),
+					),
+				},
+			),
 		},
 	}
 
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -318,7 +318,10 @@ func (p *GatewayAPIProcessor) computeHosts(hostnames []gatewayapi_v1alpha1.Hostn
 }
 
 func removeFirstDNSLabel(input string) string {
-	return input[strings.IndexAny(input, ".")+1:]
+	if strings.Contains(input, ".") {
+		return input[strings.IndexAny(input, "."):]
+	}
+	return input
 }
 
 func validHostName(hostname string) error {
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -276,12 +276,6 @@ func (p *GatewayAPIProcessor) computeHosts(hostnames []gatewayapi_v1alpha1.Hostn
 
 		hostname := string(host)
 
-		//// A "*" hostname matches anything.
-		//if hostname == "*" {
-		//	hosts[hostname] = struct{}{}
-		//	continue
-		//}
-
 		// Validate the hostname.
 		if err := validHostName(hostname); err != nil {
 			errors = append(errors, err)
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -289,7 +289,6 @@ func (p *GatewayAPIProcessor) computeHosts(hostnames []gatewayapi_v1alpha1.Hostn
 		}
 
 		if listenerHostname != nil {
-			const wildcardString string = "[a-z0-9]([-a-z0-9]*[a-z0-9])*."
 			lhn := string(*listenerHostname)
 
 			// A "*" hostname matches anything.
@@ -303,38 +302,34 @@ func (p *GatewayAPIProcessor) computeHosts(hostnames []gatewayapi_v1alpha1.Hostn
 				continue
 			} else if strings.Contains(lhn, "*") {
 
-				// Build a regex to match the hostname against the wildcard listener hostname.
-				wildcardDNS1123SubdomainRegexp := regexp.MustCompile("^" + strings.Replace(lhn, "*", wildcardString, 1) + "$")
-
-				// Validate the wildcard gateway listener hostname matches the hostnames hostname.
-				if !wildcardDNS1123SubdomainRegexp.MatchString(hostname) {
-					errors = append(errors, fmt.Errorf("gateway hostname %q does not match hostnames hostname %q", lhn, hostname))
+				if removeFirstDNSLabel(lhn) != removeFirstDNSLabel(hostname) {
+					errors = append(errors, fmt.Errorf("gateway hostname %q does not match route hostname %q", lhn, hostname))
 					continue
 				}
-			} else if strings.Contains(hostname, "*") {
+				//} else if strings.Contains(hostname, "*") {
 
-				// Build a regex to match the hostname against the wildcard hostnames hostname.
-				wildcardDNS1123SubdomainRegexp := regexp.MustCompile("^" + strings.Replace(hostname, "*", wildcardString, 1) + "$")
-
-				// Validate the wildcard hostnames hostname matches gateway listener hostname.
-				if !wildcardDNS1123SubdomainRegexp.MatchString(lhn) {
-					errors = append(errors, fmt.Errorf("hostnames hostname %q does not match gateway hostname %q", hostname, lhn))
-					continue
-				}
+				//if removeFirstDNSLabel(lhn) != removeFirstDNSLabel(hostname) {
+				//	errors = append(errors, fmt.Errorf("hostnames hostname %q does not match gateway hostname %q", hostname, lhn))
+				//	continue
+				//}
 
 				// The Listener hostname is not a wildcard, but the route is a wildcard.
 				// Since the gateway is more specific, use that to match the route against.
-				hosts[lhn] = struct{}{}
-				continue
+				//hosts[lhn] = struct{}{}
+				//continue
 			} else {
 				// Validate the gateway listener hostname matches the hostnames hostname.
-				errors = append(errors, fmt.Errorf("gateway hostname %q does not match hostnames hostname %q", lhn, hostname))
+				errors = append(errors, fmt.Errorf("gateway hostname %q does not match route hostname %q", lhn, hostname))
 				continue
 			}
 		}
 		hosts[hostname] = struct{}{}
 	}
 	return hosts, errors
+}
+
+func removeFirstDNSLabel(input string) string {
+	return input[strings.IndexAny(input, ".")+1:]
 }
 
 func validHostName(hostname string) error {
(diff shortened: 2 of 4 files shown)


internal/dag/builder_test.go
```diff
@@ -2266,6 +2298,79 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
...
+							"http.projectcontour.io",
+						},
+						Rules: []gatewayapi_v1alpha1.HTTPRouteRule{{
+							Matches:   httpRouteMatch(gatewayapi_v1alpha1.PathMatchExact, "/blog"),
+							ForwardTo: httpRouteForwardTo("kuard", 8080, 1),
+						}},
+					},
+				},
+			},
+			want: listeners(
```
sunjayBhatia, line 2362: is there any documentation upstream about how we handle the route and listener specifying a hostname?

stevesloka: I assumed they were combined together which is what this PR does. I'm asking upstream to see if that's correct.

stevesloka: So my assumption is wrong, going to fix the docs upstream and rework this now. 

TLDR is that the listener `hostname` defines what the routes can use. So for a listener hostname of `*.test.com`, you could have a route with `foo.test.com`, but not `foo.steve.com`.

stevesloka:
Ok I think I have this all fixed up now for how it should work. Going to PR a docs change upstream now to try and have a reference to validate against.

internal/dag/gatewayapi_processor.go
```diff
@@ -195,39 +196,111 @@ func isSecretRef(certificateRef *gatewayapi_v1alpha1.LocalObjectReference) bool
...
+			} else if strings.Contains(hostname, "*") {
+
+				// Build a regex to match the hostname against the wildcard route hostname.
+				wildcardDNS1123SubdomainRegexp := regexp.MustCompile("^" + strings.Replace(hostname, "*", wildcardString, 1) + "$")
+
+				// Validate the wildcard route hostname matches gateway listener hostname.
+				if !wildcardDNS1123SubdomainRegexp.MatchString(lhn) {
+					errors = append(errors, fmt.Errorf("route hostname %q does not match gateway hostname %q", hostname, lhn))
+					continue
+				}
```
skriss, line 272: can both the listener and route hostname be a wildcard, e.g. `*.foo.com` for both? If so then before doing the wildcard matching we should check for string equality between the listener and route host and add it if so.

internal/dag/gatewayapi_processor.go
```diff
@@ -195,39 +196,111 @@ func isSecretRef(certificateRef *gatewayapi_v1alpha1.LocalObjectReference) bool
 	return strings.ToLower(certificateRef.Kind) == "secret" && strings.ToLower(certificateRef.Group) == "core"
 }
 
-func (p *GatewayAPIProcessor) computeHosts(route *gatewayapi_v1alpha1.HTTPRoute) ([]string, []error) {
-	// Determine the hosts on the route, if no hosts
-	// are defined, then set to "*".
+func (p *GatewayAPIProcessor) computeHosts(route *gatewayapi_v1alpha1.HTTPRoute, listenerHostname *gatewayapi_v1alpha1.Hostname) ([]string, []error) {
```
skriss, line 199: A comment explaining the logic of this method would be helpful, including the details around how the route and listener hostnames relate.

skriss (commented):
code looks pretty good, I had one question on the logic. Looking at tests next.

internal/dag/gatewayapi_processor_test.go
```diff
@@ -26,11 +26,13 @@ import (
 
 func TestComputeHosts(t *testing.T) {
```
skriss, line 27: I think you could change `computeHosts` to not take a full `HTTPRoute`, just the hostnames string slice. It'd make the test setup easier here.

internal/dag/gatewayapi_processor_test.go
```diff
@@ -184,9 +193,199 @@ func TestComputeHosts(t *testing.T) {
...
+				},
+				Spec: gatewayapi_v1alpha1.HTTPRouteSpec{
+					Hostnames: []gatewayapi_v1alpha1.Hostname{
+						"http.example.com",
+					},
+				},
+			},
+			want:      nil,
+			wantError: []error{fmt.Errorf("gateway hostname \"*.projectcontour.io\" does not match route hostname \"http.example.com\"")},
+		},
```
skriss, line 312: it'd be nice to add some cases to verify that e.g. `too.many.labels.projectcontour.io` doesn't match `*.projectcontour.io` (in both directions, i.e. where the listener has the wildcard and where the route has the wildcard) to get better coverage around the regex.

internal/dag/gatewayapi_processor_test.go
```diff
@@ -184,9 +193,199 @@ func TestComputeHosts(t *testing.T) {
...
+				},
+				Spec: gatewayapi_v1alpha1.HTTPRouteSpec{
+					Hostnames: []gatewayapi_v1alpha1.Hostname{
+						"http.projectcontour.io",
+					},
+				},
+			},
+			want:      []string{"http.projectcontour.io"},
+			wantError: nil,
+		},
```
skriss, line 292: per previous comment, let's add a test case for when both the listener and route have e.g. `*.projectcontour.io` (not exactly sure what expected behavior is but let's define it via test).

internal/dag/gatewayapi_processor_test.go
sunjayBhatia: yeah good point, probably need an integration test as well for wildcards if we're intending on supporting it already

internal/dag/gatewayapi_processor.go
stevesloka: 👍

internal/dag/gatewayapi_processor.go
stevesloka: Yup nice catch, I fixed that up and added a test for it.

internal/dag/gatewayapi_processor_test.go
stevesloka: 👍

internal/dag/gatewayapi_processor_test.go
stevesloka: I added some tests, I'll make some additional integration tests in a future PR.

stevesloka:
Thanks for the comments @skriss @sunjayBhatia, I've got this updated.

internal/dag/accessors.go
```diff
@@ -24,7 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-// RouteServiceName identifies a service used in a route.
+// RouteServiceName identifies a service used in a hostnames.
```
skriss, line 27: probably an inadvertent find/replace?

internal/dag/conditions.go
```diff
@@ -39,7 +39,7 @@ func mergePathMatchConditions(conds []contour_api_v1.MatchCondition) MatchCondit
 	// we need to set the prefix to /.
 	// Remember that this step is done AFTER all the includes have happened.
 	// Setting this to / allows us to pass this prefix to Envoy, as there must
-	// be at least one path, prefix, or regex set on each Envoy route.
+	// be at least one path, prefix, or regex set on each Envoy hostnames.
```
skriss, line 42: this file's changes also look like inadvertent find/replace

internal/dag/dag.go
```diff
@@ -187,13 +187,13 @@ func (hc *HeaderMatchCondition) String() string {
 }
 
 // DirectResponse allows for a specific HTTP status code
-// to be the response to a route request vs routing to
+// to be the response to a hostnames request vs routing to
```
skriss, line 190: this file too

internal/dag/gatewayapi_processor.go
```diff
@@ -119,7 +120,7 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
 
 			// ## RouteBindingSelector ##
 			//
-			// Selector specifies a set of route labels used for selecting routes to associate
+			// Selector specifies a set of hostnames labels used for selecting routes to associate
```
skriss, line 123: looks like there are a bunch more find/replace errors in this and other file, I'm not gonna comment on them all, hopefully you can go through and revert all the erroneous ones.

internal/dag/gatewayapi_processor.go
```diff
@@ -195,39 +196,121 @@ func isSecretRef(certificateRef *gatewayapi_v1alpha1.LocalObjectReference) bool
...
+				wildcardDNS1123SubdomainRegexp := regexp.MustCompile("^" + strings.Replace(hostname, "*", wildcardString, 1) + "$")
+
+				// Validate the wildcard hostnames hostname matches gateway listener hostname.
+				if !wildcardDNS1123SubdomainRegexp.MatchString(lhn) {
+					errors = append(errors, fmt.Errorf("hostnames hostname %q does not match gateway hostname %q", hostname, lhn))
+					continue
+				}
+			} else {
+				// Validate the gateway listener hostname matches the hostnames hostname.
+				if lhn != hostname {
```
skriss, line 285: I think this check for non-equality is redundant/can be removed since we already checked for equality on https://github.com/projectcontour/contour/pull/3640/files#diff-2dfb3932d84ad36571b20cb8d66d9f0b623fdbbef49a0b41eff48eaa7fa22a05R258

internal/dag/gatewayapi_processor.go
```diff
@@ -195,39 +196,121 @@ func isSecretRef(certificateRef *gatewayapi_v1alpha1.LocalObjectReference) bool
...
-				errors = append(errors, fmt.Errorf("invalid listener hostname %q: %v", hostname, errs))
+
+			// If the listener.hostname matches then no need to
+			// do any other validation.
+			if lhn == hostname {
+				hosts = append(hosts, hostname)
 				continue
 			}
+
+			if strings.Contains(lhn, "*") {
```
skriss, line 263: nit: I would switch this to an `else if` joined with the above `if` since they all go together, for clarity, or could do a switch.

skriss (changes requested):
Looks pretty good, just a couple minor style nits and some over-eager find/replaces to revert.

internal/dag/gatewayapi_processor.go
stevesloka: They all have `continues` in them which showed they exit, but can update.

internal/dag/gatewayapi_processor.go
stevesloka: We need this to cover this test case where the hostnames don't match without wildcards: https://github.com/projectcontour/contour/blob/d0ed425ce3a3d6950e989e3ae58963fea5e2bfae/internal/dag/gatewayapi_processor_test.go#L119

stevesloka:
@skriss sorry about those renames, Goland's refactor+rename bit me again.

internal/dag/gatewayapi_processor.go
skriss: We need the `else`, but since we already checked for `if lhn == hostname` above, then we know if we got here that `lhn != hostname`. It's not a big deal, can leave.

skriss (approved):
LGTM

internal/dag/gatewayapi_processor.go
stevesloka: Ohh I see now, I thought you meant to remove the `if` statement. I updated. =)

skriss (approved):

internal/dag/gatewayapi_processor_test.go
```diff
@@ -17,175 +17,190 @@ import (
...
+		"listener host & hostnames host do not match wildcard host": {
+			listenerHost: listenerHostname("*.projectcontour.io"),
+			hostnames: []gatewayapi_v1alpha1.Hostname{
+				"http.example.com",
 			},
+			want:      nil,
+			wantError: []error{fmt.Errorf("gateway hostname \"*.projectcontour.io\" does not match hostnames hostname \"http.example.com\"")},
+		},
+		"listener host & wildcard hostnames host match": {
```
sunjayBhatia, line 151: is this allowed? this seems like the listener is more specific than the route, which could be not allowed?

internal/dag/gatewayapi_processor_test.go
```diff
@@ -17,175 +17,190 @@ import (
...
+		},
+		"listener host wildcard & wildcard hostnames host match": {
+			listenerHost: listenerHostname("*.projectcontour.io"),
+			hostnames: []gatewayapi_v1alpha1.Hostname{
+				"*.projectcontour.io",
+			},
+			want:      []string{"*.projectcontour.io"},
+			wantError: nil,
+		},
+		"listener host & wildcard all hostnames host match": {
```
sunjayBhatia, line 167: similar to the above, seems like this is a more specific gateway hostname than the route hostname

internal/dag/gatewayapi_processor.go
```diff
@@ -195,39 +196,115 @@ func isSecretRef(certificateRef *gatewayapi_v1alpha1.LocalObjectReference) bool
...
+	}
+
+	for _, host := range hostnames {
 
 		hostname := string(host)
-		if isIP := net.ParseIP(hostname) != nil; isIP {
-			errors = append(errors, fmt.Errorf("hostname %q must be a DNS name, not an IP address", hostname))
+
+		// A "*" hostname matches anything.
+		if hostname == "*" {
```
sunjayBhatia, line 235: a `*` should only be able to come from line https://github.com/projectcontour/contour/pull/3640/files#diff-2dfb3932d84ad36571b20cb8d66d9f0b623fdbbef49a0b41eff48eaa7fa22a05R211 above, but could the order of this processing mean that we now allow `*` in a hostname on a route? (which doesn't seem like it should be allowed, per the docs)

internal/dag/gatewayapi_processor_test.go
```diff
@@ -17,175 +17,190 @@ import (
...
+			listenerHost: listenerHostname("http.projectcontour.io"),
+			hostnames: []gatewayapi_v1alpha1.Hostname{
+				"*",
```
sunjayBhatia, line 170: i think this is an illegal hostname on a route:

```
The wildcard character * must appear by itself as the first DNS label and matches only a single label. You cannot have a wildcard label by itself (e.g. Host == *).
```

see https://gateway-api.sigs.k8s.io/references/spec/#networking.x-k8s.io/v1alpha1.HTTPRoute hostname field

sunjayBhatia (commented):
also my guess is that an integration test will show we need something similar to https://github.com/projectcontour/contour/blob/dee66310416af9e0477705916421a7e9a7219f99/internal/dag/ingress_processor.go#L233-L248

to make sure wildcard matching works as expected

internal/dag/gatewayapi_processor.go
stevesloka: You're correct based on the docs. We can set a `*` for the hostname and not setting is the same thing, but agree should update.

internal/dag/gatewayapi_processor_test.go
stevesloka: I guess this test would be the same as (https://github.com/projectcontour/contour/blob/d0ed425ce3a3d6950e989e3ae58963fea5e2bfae/internal/dag/gatewayapi_processor_test.go#L183) which essentially has no hostname defined on the Route, but the Gateway does, so it should end up only matching the Gateway.

internal/dag/gatewayapi_processor_test.go
stevesloka: I'm not 100% sure, but the test is wrong. If the Gateway defines `hostname.projectcontour.io` and the route defines `*.projectcontour.io`, then I think `hostname.projectcontour.io` should end up being the match. Does that make sense?

stevesloka:
Need to also add some tests thinking about this some more if the route defines multiple hostnames or the set of routes end up with more than one hostname.

stevesloka:
I think this is ready for a final look @sunjayBhatia. I added the wildcard tests/changes to https://github.com/projectcontour/contour/pull/3701

internal/dag/gatewayapi_processor_test.go
sunjayBhatia: yeah i think if we do allow this, the route programmed should be for the more specific one

I guess the question is can a route have a wildcard on a gateway/listener with a specific match? `*.projectcontour.io` is a superset of `http.projectcontour.io` so it is a bit odd to me this could be valid? idk

sunjayBhatia: cc @projectcontour/maintainers if there are any other opinions on this

internal/dag/gatewayapi_processor_test.go
```diff
@@ -27,166 +27,251 @@ import (
...
+			listenerHost: listenerHostname("http.projectcontour.io"),
+			hostnames: []gatewayapi_v1alpha1.Hostname{
+				"*.projectcontour.io",
+			},
+			want: map[string]struct{}{
+				"http.projectcontour.io": {},
+			},
+			wantError: nil,
+		},
+		"listener host & wildcard hostname and matching hostname match": {
```
sunjayBhatia, line 196: similar to the above comment on this one

internal/dag/gatewayapi_processor_test.go
```diff
@@ -27,166 +27,251 @@ import (
...
+			want:      map[string]struct{}{},
+			wantError: []error{fmt.Errorf("hostnames hostname \"*.projectcontour.io\" does not match gateway hostname \"too.many.labels.projectcontour.io\"")},
+		},
+		"listener wildcard host doesn't match hostnames with many labels host": {
+			listenerHost: listenerHostname("*.projectcontour.io"),
+			hostnames: []gatewayapi_v1alpha1.Hostname{
+				"too.many.labels.projectcontour.io",
+			},
+			want:      map[string]struct{}{},
+			wantError: []error{fmt.Errorf("gateway hostname \"*.projectcontour.io\" does not match hostnames hostname \"too.many.labels.projectcontour.io\"")},
```
sunjayBhatia, line 274: maybe add a test or two where a route tries to define a hostname `*` with the listener also allowing `*`

internal/dag/gatewayapi_processor_test.go
stevesloka: My understanding is the listener.hostname defines the rules, so if it specifies a wildcard, and the route's hostname matches, then that route hostname is allowed to serve.

internal/dag/gatewayapi_processor_test.go
stevesloka: ok added a test with `*` on listener with `*.projectcontour.io` on route.

stevesloka:
I ended up squashing stuff since the rebase was horrible, sorry @sunjayBhatia

internal/dag/gatewayapi_processor_test.go
danehans: @stevesloka IMO a wildcard hostname should not be allowed for a route bound to a gateway listener with a more precise hostname. The API does not specify this use case, so I suggest updating https://github.com/kubernetes-sigs/gateway-api/issues/654 to get input from the community.

internal/dag/builder_test.go
```diff
@@ -2515,6 +2547,77 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
...
+				&Listener{
+					Port: 80,
+					VirtualHosts: virtualhosts(
+						virtualhost("*.projectcontour.io",
+							exactrouteHTTPRoute("/blog", service(kuardService))),
+					),
```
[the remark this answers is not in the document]
sunjayBhatia: ok looks like this is resolved since we made the route use a precise hostname match with the wildcard gateway

internal/dag/gatewayapi_processor.go
```diff
@@ -230,24 +230,98 @@ func isSecretRef(certificateRef *gatewayapi_v1alpha1.LocalObjectReference) bool
...
+				continue
+			} else if lhn == hostname {
+				// If the listener.hostname matches then no need to
+				// do any other validation.
+				hosts[hostname] = struct{}{}
+				continue
+			} else if strings.Contains(lhn, "*") {
+
+				// Build a regex to match the hostname against the wildcard listener hostname.
+				wildcardDNS1123SubdomainRegexp := regexp.MustCompile("^" + strings.Replace(lhn, "*", wildcardString, 1) + "$")
```
sunjayBhatia, line 299: nit: I think this could be simplified to not need a regex match? If the listener hostname is a wildcard (`*.foo.com`) just chop the first DNS label from that and the route hostname and those suffixes must be equal

not super critical but makes it harder to make a mistake in case we get our regex wrong etc.

internal/dag/gatewayapi_processor.go
```diff
@@ -230,24 +230,98 @@ func isSecretRef(certificateRef *gatewayapi_v1alpha1.LocalObjectReference) bool
...
+
+		// Validate the hostname.
 		if err := validHostName(hostname); err != nil {
 			errors = append(errors, err)
 			continue
 		}
-		hosts = append(hosts, string(host))
+
+		if listenerHostname != nil {
+			const wildcardString string = "[a-z0-9]([-a-z0-9]*[a-z0-9])*."
```
sunjayBhatia, line 284: just noticed, do we have to escape the `.` at the end b/c we want a literal `.`?

internal/dag/gatewayapi_processor_test.go
sunjayBhatia: commented on the upstream issue, not sure what we want to do for now, restrict to something less permissive and expand or open it up now and restrict later

internal/dag/gatewayapi_processor_test.go
sunjayBhatia: sorry to be a pain but I think a test where the HTTPRoute tries to use `*` as a hostname should be added too, since that is invalid

I don't remember if somewhere else in the code already restricts us from using that but if so would be good to have an explicit test somewhere, here or in the builder or something?

internal/dag/gatewayapi_processor.go
stevesloka: nope I like it!

internal/dag/gatewayapi_processor.go
stevesloka: took out all the regex stuff with your suggestion.

stevesloka:
> I don't remember if somewhere else in the code already restricts us from using that but if so would be good to have an explicit test somewhere, here or in the builder or something?

I added a test to be explicit, but it's checked here: https://github.com/projectcontour/contour/blob/af1856edc34c7423b05796d33d0f8063727225f5/internal/dag/gatewayapi_processor.go#L278

internal/dag/gatewayapi_processor.go
```diff
@@ -233,28 +233,86 @@ func isSecretRef(certificateRef *gatewayapi_v1alpha1.LocalObjectReference) bool
...
+				continue
+			}
+		}
+		hosts[hostname] = struct{}{}
 	}
 	return hosts, errors
 }
 
+func removeFirstDNSLabel(input string) string {
+	return input[strings.IndexAny(input, ".")+1:]
```
sunjayBhatia, line 313: i think the one thing to be careful about with this is this case:

```
package main

import (
	"fmt"
	"strings"
)

func main() {
	input := "*.foo"
	fmt.Println(input[strings.IndexAny(input, ".")+1:])
	input = "foo"
	fmt.Println(input[strings.IndexAny(input, ".")+1:])
}
```

these both end up being `foo`

so to be safe, just return the whole string if it has no `.` and keep the `.` if it does?

stevesloka: nope makes sense!

internal/dag/gatewayapi_processor.go
```diff
@@ -233,28 +233,89 @@ func isSecretRef(certificateRef *gatewayapi_v1alpha1.LocalObjectReference) bool
...
+				return hosts, nil
+			}
+		}
+	}
+
 	for _, host := range hostnames {
 
 		hostname := string(host)
+
+		//// A "*" hostname matches anything.
```
sunjayBhatia, line 271: sorry last thing 😅  probably can delete

sunjayBhatia (approved):

PR #3642, opened 2021-05-05
cmd/contour: Remove deprecated flag 'experimental-service-apis'

stevesloka:


sunjayBhatia (approved):

skriss:
https://github.com/projectcontour/contour/blob/main/site/docs/main/configuration.md has a reference to it too which should be removed

stevesloka:
good call on the site reference, updated @skriss!

Issue #3628, opened 2021-04-30
Remove all usage of v1beta1 Ingress


sunjayBhatia:
Now that we haved moved our support matrix to k8s 1.19+, we can solely rely on Ingress v1 resources in informers etc. and can remove references to Ingress v1beta1 internally and in tests.

This issue is to track the changes which will touch a lot of files so would be good to break up to make reviews easier.

I believe possibly the one place to leave an Ingress v1beta1 resource would be in an integration test just to make sure everything works properly with the v1 informers, internal processing, etc. in case Ingress v1beta1 resources are still in use. This test can be removed once we are on k8s versions that remove Ingress v1beta1.

sunjayBhatia:
This is complete, closing

PR #3645, opened 2021-05-06
internal/dag: Remove Ingress v1beta1

sunjayBhatia:
- Removes v1beta1 tests
- Removes v1beta1 Ingress insertion/deletion from cache
  - Note: This will mean k8s pre-1.19 will stop working with Ingress
    v1beta1 since Ingress v1 does not exist yet in those versions

Updates: #3628


sunjayBhatia:
Converted to draft b/c this package refactor relies on others

sunjayBhatia:
added release note labels becasue this will make Contour unusable in versions of k8s that do not have ingress v1, as we're no longer falling back to v1beta1

internal/dag/builder_test.go
```diff
@@ -3080,777 +3079,575 @@ func TestDAGInsert(t *testing.T) {
 		},
 	}
 
-	i1 := &v1beta1.Ingress{
+	// s3a and b have http/2 protocol annotations
```
skriss, line 3082: oof, this file's diff is pretty hard to review, but tests are passing so 🤞

skriss (approved):
LGTM

PR #3650, opened 2021-05-07
interna/k8s: Remove v1beta1 usage

sunjayBhatia:
- Removes informer for v1beta1
- Removes status updater/informer for v1beta1
- Ensures k8s versions older than 1.19 are incompatible
  as we are relying on Ingress v1 existing

Updates: #3628

### DIFF
--- a/internal/k8s/kind.go
+++ b/internal/k8s/kind.go
@@ -70,6 +70,8 @@ func VersionOf(obj interface{}) string {
 		switch obj := obj.(type) {
 		case *v1.Secret, *v1.Service, *v1.Endpoints:
 			return v1.SchemeGroupVersion.String()
+		case *networking_v1.Ingress:
+			return networking_v1.SchemeGroupVersion.String()
 		case *contour_api_v1.HTTPProxy, *contour_api_v1.TLSCertificateDelegation:
 			return contour_api_v1.GroupVersion.String()
 		case *v1alpha1.ExtensionService:
--- a/internal/k8s/kind.go
+++ b/internal/k8s/kind.go
@@ -18,7 +18,6 @@ import (
 	"github.com/projectcontour/contour/apis/projectcontour/v1alpha1"
 	v1 "k8s.io/api/core/v1"
 	networking_v1 "k8s.io/api/networking/v1"
-	"k8s.io/api/networking/v1beta1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes/scheme"
@@ -40,8 +39,6 @@ func KindOf(obj interface{}) string {
 			return "Service"
 		case *v1.Endpoints:
 			return "Endpoints"
-		case *v1beta1.Ingress:
-			return "Ingress"
 		case *networking_v1.Ingress:
 			return "Ingress"
 		case *contour_api_v1.HTTPProxy:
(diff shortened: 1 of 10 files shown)


sunjayBhatia:
added release note labels becasue this will make Contour unusable in versions of k8s that do not have ingress v1, as we're no longer falling back to v1beta1

stevesloka (commented):
Should this wait until our oldest supported version is 1.22? Meaning, you can't use v1beta1 any more?

Should we also mark this as deprecated so that there's a heads up this is going away in Contour?

internal/k8s/kind.go
```diff
@@ -73,8 +70,6 @@ func VersionOf(obj interface{}) string {
 		switch obj := obj.(type) {
 		case *v1.Secret, *v1.Service, *v1.Endpoints:
 			return v1.SchemeGroupVersion.String()
-		case *v1beta1.Ingress:
```
skriss, line 76: this function doesn't have a case for Ingress v1, I assume it should (haven't looked at where it's used)

internal/k8s/kind_test.go
```diff
@@ -60,7 +58,6 @@ func TestVersionOf(t *testing.T) {
 		{"v1", &v1.Secret{}},
 		{"v1", &v1.Service{}},
 		{"v1", &v1.Endpoints{}},
-		{"networking.k8s.io/v1beta1", &v1beta1.Ingress{}},
```
skriss, line 63: add case for v1 Ingress per previous comment

sunjayBhatia:
> Should this wait until our oldest supported version is 1.22? Meaning, you can't use v1beta1 any more?
> 
> Should we also mark this as deprecated so that there's a heads up this is going away in Contour?

Our current support matrix is 1.19+ so we're only ever informing on and acting on Ingress v1 objects (b/c we check for v1 Ingress existing and use that informer, otherwise we use v1beta1), v1beta1 created resources are translated by the API server to v1 when we grab them, so you can still use v1beta1 now.

I think the explicit v1beta1 code is now effectively dead code, unless we want Contour 1.16+ to still be usable on k8s 1.18 (even though our compatibility matrix doesn't cover it explicitly).

skriss (changes requested):
So my understanding is that this draws a hard line under 1.19 as our minimum version, but you can still use `v1beta1` Ingress resources in 1.19+ with Contour since the API server automagically converts them to `v1` so we'll pick them up via our `v1` informers. Once 1.22 rolls around, though, `v1beta1` is completely gone.

Personally I think this is OK given that our support matrix says 1.19 is the minimum, and users on older Kube versions can always continue using older Contour versions if they like. We did something similar when we updated to v1 CRDs, and drew a hard line at Kube 1.16 as the minimum version that could run Contour OOTB.

stevesloka:
👍  you're both right, sorry, forgot about the auto-conversion stuff. =)

internal/k8s/kind.go
sunjayBhatia: ah good call, missed that one 👍🏽

sunjayBhatia: seems a good few of these cases aren't actually hit with the tests, the coverage report shows

i don't know the history on this code but maybe this could be shortened 🤔

skriss: Yeah, I don't really know either. Looks like the first line in the function might be the one that's actually hit to get the `Kind` for these objects. In any case, update looks good.

skriss (approved):

stevesloka (approved):

Issue #2509, opened 2020-05-08
docs: Terminate TLS at AWS NLB


stevesloka:
Need to update docs to show how to use an AWS NLB and terminate TLS there, then proxy traffic to Contour running behind the NLB.

_Originally posted by @stevesloka in https://github.com/projectcontour/contour/issues/2441#issuecomment-625853628_

geoffcline:
Hi from AWS 👋 I'm taking a look

youngnick:
I'll assign this to you for now @geoffcline, please ping us with any updates!

xaleeks:
Seems this is done, going to close it out

PR #3655, opened 2021-05-10
add guide for AWS NLB TLS termination

geoffcline:
projectcontour/contour: add guide for AWS NLB TLS termination

Add a new page to the guide section. The page describes the advantages of configuring an AWS NLB to handle TLS termination, and the steps required. 

Updates #2509 

Signed-off-by: Geoffrey Cline <gcline@amazon.com>


stevesloka (commented):
Overall looks great to me @geoffcline! One quick question though, is the Route53 bit a hard requirement? Can I just CNAME my domain to the NLB dns name?

geoffcline:
that's a good question @stevesloka -- the domain needs to be validated before generating the certificate in ACM. I'll add a note clarifying that the requirement is the verification process in ACM **or** the domain registered with Route 53. 

Thanks! -Geoffrey

geoffcline:
updated guidance on DNS services!

it looks like the tests are failing because the new URL of the added page (https://projectcontour.io/guides/deploy-aws-tls-nlb/) is currently returning a 404 from the live server? am I understanding that right? any guidance?

stevesloka:
> the domain needs to be validated before generating the certificate in ACM. 

Ahh that's right, forgot about that part! =)

> it looks like the tests are failing because the new URL of the added page (https://projectcontour.io/guides/deploy-aws-tls-nlb/) is currently returning a 404 from the live server

Yup, that will forever fail until the url exists. We have an open issue somewhere to fix this and test against the preview url but just haven't yet implemented that yet.

stevesloka (approved):
/lgtm Thanks for adding this @geoffcline!! 🎉

youngnick (approved):
This looks great @geoffcline, thanks very much!

Issue #3617, opened 2021-04-26
Move away from Gateway Configuration in Contour Config File


stevesloka:
Currently, Contour requires a name/namespace of a Gateway to be [configured](https://github.com/projectcontour/contour/blob/main/examples/contour/01-contour-config.yaml#L15-L17) in the Contour configuration file. Today if this isn't configured and a user wants to use GatewayAPI, then they must set the value in the file, then restart Contour. 

We could move to use the `controller` field of the `GatewayClass` or a specific field in the `parametersRef`. This would allow the Gateway to be specified easily without needing a configuration file change / management.

youngnick:
I think using the `controller` field in the GatewayClass is a better idea, and allows for more dynamic configuration. It's also the most Kube-y way to do it.

stevesloka:
What would you use to specify the name/namespace? Something like: `projectcontour.io/name/namespace`? So for a gateway in the `projectcontour` namespace named `contour`, you'd have a GatewayClass like this:

```
kind: GatewayClass
apiVersion: networking.x-k8s.io/v1alpha1
metadata:
  name: contour-class
spec: 
  controller: projectcontour.io/contour/projectcontour
```

youngnick:
I assumed that we'd just do like what we do with IngressClass, you specify some value, and the controller uses that, probably with a `projectcontour.io/` prefix. Doing things that way means that we need to handle what happens if more than one GatewayClass has the same `spec.controller` field, we've talked about that before.

But, this does highlight a similar problem to the IngressClass field/annotation: What happens if two Contour installations set the same IngressClass or GatewayClass `spec.controller`? I *think* that they'll probably fight right now?

For the simple internal-external multiple-Contours case, I think that having the namespace in there and not just some string (that is, `internal` and `external`) feels like overkill.

I'm not sure though, it feels like this bit could also use some upstream discussion maybe?

stevesloka:
What I want to do here though is remove the need to configure Contour with something. So in the example of having an internal/external instances, you would configure the GatewayClass  to say, the Contour controller named `internal` in the namespace `ingress-internal` is who will service my routes for internal requests, then another GatewayClass would say, the Contour controller named `external` in the namespace `ingress-external` would service requests for external resources.

When deploying Contour, there's nothing special to configure, everything is driven from the GatewayClass and Contour doesn't need restarted for this to work.

youngnick:
[...]

PR #3659, opened 2021-05-10
Implements GatewayClass Controller

danehans:
Updates Gateway API support to use controller-runtime added by https://github.com/projectcontour/contour/pull/3510:

- `cmd/contour/serve.go`: Adds `NewGatewayClassController()` to create a new gatewayclass controller.
- `examples/contour/02-role-contour.yaml` and `examples/render/contour.yaml`: Updates Contour's RBAC to support reconciling gatewayclasses and managing the Envoy service.
- `internal/dag/cache.go`: Adds gatewayclasses that contain the contour controller string to the kubernetes cache.
- `internal/equality/equality.go`: Adds an equality pkg used to compare current/desired state of resources, conditions, etc..
- `internal/controller/gatewayclass.go`: Adds gatewayclass  controller and reconciler functionality.
- `internal/k8s/informers.go`: Updates kubebuilder RBAC tags and the list of Gateway API resources to include gatewayclasses.
- `internal/status/conditions.go`: Adds support for managing gateway and gatewayclass status conditions.
- `internal/status/gatewayclass.go`: Adds support for managing gatewayclass status.
- `internal/validation/gatewayclass.go`: Adds support for validating gatewayclass resources.
- `pkg/config/parameters.go`: Adds `GatewayClassController` identifier used by Contour's gatewayclass controller to determine if a GatewayClass object should be managed.

Requires https://github.com/projectcontour/contour/pull/3510

TODOs:
- https://github.com/projectcontour/contour/issues/3705
- https://github.com/projectcontour/contour/issues/3700
- https://github.com/projectcontour/contour/issues/3617 (requires gateway controller implementation)
- https://github.com/projectcontour/contour/issues/3706


danehans:
xref https://github.com/projectcontour/contour/pull/3664 that updates deployment manifests/tooling.

stevesloka (changes requested):
This is a quick first pass, needs a rebase now that I just merged #3510. Also, would be super nice to pick this out into smaller PRs, one for the controller bits, one for status, one for validation, etc. 

Also seems like many of the bits aren't tested (the Codecov is complaining all over the place which is making the review difficult).

internal/k8s/cache/gateway.go
```diff
@@ -0,0 +1,337 @@
...
+	if err := c.Watch(&source.Kind{Type: &gatewayv1a1.Gateway{}}, r.enqueueRequestForOwnedGateway()); err != nil {
+		return nil, err
+	}
+	// TODO: Add a watch for gatewayclasses owned by contour to keep gateway status updated.
+	return c, nil
+}
+
+// enqueueRequestForOwnedGateway returns an event handler that maps events to
+// Gateway objects that reference a GatewayClass owned by Contour.
+func (r *gatewayReconciler) enqueueRequestForOwnedGateway() handler.EventHandler {
```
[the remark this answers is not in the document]
danehans: @stevesloka `enqueueRequestForOwnedGateway()` filters events before they are given to the controller's EventHandler. This follows controller-runtime best practices for event filtering. Maybe the dag filter you reference should be removed in favor of this watch predicate, thoughts?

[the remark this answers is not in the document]
stevesloka: It's not a problem to move the logic from where it's happening, just need to pick one and remove the other. Also, need to make sure we're filtering against the configured name/namespace of the gateway.

internal/validation/gateway.go
```diff
@@ -0,0 +1,115 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/validation"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gateway_v1alpha1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
+)
+
+// Gateway returns an error if gw is an invalid Gateway.
+func Gateway(ctx context.Context, cli client.Client, gw *gateway_v1alpha1.Gateway) error {
```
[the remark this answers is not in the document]
stevesloka: We need to think about where the status is going to be set for Gateways. There is some up-front validation that can be applied, but what parts do routes have an affect on the status of the gateway? 

Additionally, if we don't pass the configured gateway off to the gateway_processor, then all the other routes will be invalidated and taken out of Envoy config. That might be ok since if the gateway is invalid, then the routes probably should't get traffic, but want to understand their relationship.

internal/status/conditions_test.go
```diff
@@ -15,20 +15,23 @@ package status
...
 import (
 	"testing"
+	"time"
 
 	projectcontour "github.com/projectcontour/contour/apis/projectcontour/v1"
 	"github.com/projectcontour/contour/internal/k8s"
 	"github.com/stretchr/testify/assert"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	gatewayapi_v1alpha1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
```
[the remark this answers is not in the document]
danehans: I was following upstream conventions, i.e. `metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"` as `gatewayapi_v1alpha1` seems overly verbose. I'll update my imports to use `gatewayapi_v1alpha1` and we can address the verbosity in the future.

internal/dag/gatewayapi_processor.go
```diff
@@ -52,7 +52,7 @@ type matchConditions struct {
 	headerMatchCondition []HeaderMatchCondition
 }
 
-// Run translates Service APIs into DAG objects and
+// Run translates Gateway API objects into DAG objects and
 // adds them to the DAG.
 func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
```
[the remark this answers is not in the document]
danehans: > The gateway didn't have to exist, only be configured in Contour. 

@stevesloka we must have a bug because I see the error when gateway is configured but the gateway does not exist:
```
$ go run ./cmd/contour serve --disable-leader-election --insecure --xds-address=0.0.0.0 --config-path=./config/contour.yaml
INFO[0000] args: [serve --disable-leader-election --insecure --xds-address=0.0.0.0 --config-path=./config/contour.yaml] 
INFO[0000] Leader election disabled                     
...
ERRO[0000] Gateway is not defined!                       context=GatewayAPIProcessor
```

The error goes away when I create the projectcontour/contour gateway.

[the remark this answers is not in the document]
stevesloka: The message probably needs updated, but essentially you're trying to process resources without the gateway existing, so the processor throws that error. So configuring which gateway to use is step #1, then actually having the resource in the cluster is step #2.

danehans:
@stevesloka to reduce the size of this PR, I only implemented the gatewayclass controller in commit `330e981`. The PR description has been updated accordingly. I resolved your other feedback above in this commit as well. I will implement the gateway controller in a follow-on PR.

internal/controller/gatewayclass.go
```diff
@@ -0,0 +1,128 @@
...
+			r.log.WithField("name", gc.Name).Info("queueing gatewayclass")
+			return []reconcile.Request{
+				{
+					NamespacedName: types.NamespacedName{
+						Name: gc.Name,
+					},
+				},
+			}
+		}
+		r.log.WithField("name", gc.Name).Info("gatewayclass not owned by contour, bypassing reconciliation")
```
stevesloka, line 84: Suggest maybe adding the `config.ContourGatewayClass` string to the error to tell folks what it required to match?

internal/status/conditions.go
```diff
@@ -38,6 +39,14 @@ const ReasonValid RouteReasonType = "Valid"
 const ReasonErrorsExist RouteReasonType = "ErrorsExist"
 const ReasonGatewayAllowMismatch RouteReasonType = "GatewayAllowMismatch"
 
+type GatewayClassReasonType string
+
+const reasonValidGatewayClass = "Valid"
+const reasonInvalidGatewayClass = "Invalid"
+
+// clock is used to set lastTransitionTime on status conditions.
+var clock utilclock.Clock = utilclock.RealClock{}
```
stevesloka, line 48: Ahh we missed this on Conditions! =)

internal/status/gatewayclass.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+// SyncGatewayClass computes the current status of gc and updates status upon
+// any changes since last sync.
+func SyncGatewayClass(ctx context.Context, cli client.Client, gc *gateway_v1alpha1.GatewayClass, valid bool) error {
+	updated := gc.DeepCopy()
+
+	updated.Status.Conditions = mergeConditions(updated.Status.Conditions, computeGatewayClassAdmittedCondition(valid))
+
+	if equality.GatewayClassStatusChanged(gc.Status, updated.Status) {
```
stevesloka, line 32: This is diverting from how we manage conditions today by creating a ConditionsAccessor (https://github.com/projectcontour/contour/blob/main/internal/status/conditions.go#L74). I think this is ok, but need to make sure we don't duplicate logic.

stevesloka: Also it doesn't make sense, but what if two controllers start updating status for the Class? Can a class be shared across controllers for multiple gateways?

internal/validation/gatewayclass.go
```diff
@@ -0,0 +1,48 @@
...
+
+package validation
+
+import (
+	"fmt"
+
+	retryable "github.com/projectcontour/contour/internal/retryableerror"
+	"github.com/projectcontour/contour/pkg/config"
+
+	gatewayv1a1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
```
stevesloka, line 24: nit: s/gatewayv1a1/gateway_v1alpha1

internal/validation/gatewayclass_test.go
```diff
@@ -0,0 +1,73 @@
...
+// limitations under the License.
+
+package validation
+
+import (
+	"testing"
+
+	"github.com/projectcontour/contour/pkg/config"
+
+	gatewayv1a1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
```
stevesloka, line 21: nit: s/gatewayv1a1/gateway_v1alpha1

pkg/config/parameters.go
```diff
@@ -32,6 +32,7 @@ type ServerType string
 
 const ContourServerType ServerType = "contour"
 const EnvoyServerType ServerType = "envoy"
+const ContourGatewayClass = "projectcontour.io/contour"
```
stevesloka, line 35: I wonder if we can use the class to know if we should dynamically provision Envoy or not? This isn't part of this review per-say just thinking about how to best manage that.

stevesloka (commented):
Overall looks good to me @danehans! Just a few comments/questions/thoughts. 

The smaller bits make it much easier to review, thanks for doing that. =)

examples/gateway/00-crds.yaml
```diff
@@ -0,0 +1,205 @@
+apiVersion: apiextensions.k8s.io/v1
```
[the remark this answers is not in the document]
danehans: @stevesloka I think it makes sense add the upstream CRDs here so we can provide a similar users experience as `kubectl apply -f https://projectcontour.io/quickstart/contour.yaml`. I automated the process in https://github.com/projectcontour/contour/pull/3676.

internal/status/conditions.go
danehans: Ack. I updated the existing time pkg references so this pkg does not need both "time" and "k8s.io/apimachinery/pkg/util/clock" imports.

internal/status/gatewayclass.go
danehans: @stevesloka hmm... I wonder if this is a gap in the gc status spec. Maybe gc status should be similar to route status, where conditions are specified per gateway. Otherwise we need to perform a dance when multiple contours in gateway-mode are running in the cluster. cc: @youngnick @jpeach

stevesloka: I can open an upstream issue in the meantime.

danehans: xref upstream discussion: https://kubernetes.slack.com/archives/CR0H13KGA/p1620933776270200

pkg/config/parameters.go
danehans: In an effort to minimize the size of this PR, I'm adding Envoy support in a future PR. IMO the gc.controller string should simply differentiate implementations from one another. From my understanding, that's the intent of the field. By default, Contour should manage the Envoy infra when Gateway CRDs/CRs exist and the gc controller is "projectcontour.io/contour". If a user does not want Contour to manage the Envoy infra, we can add a config knob to the resource referenced by the gc paramRef. I prefer that we avoid adding this knob until someone makes a strong case for it.

stevesloka: > I prefer that we avoid adding this knob until someone makes a strong case for it.

I've got a strong case for it, which was part of the reason moving this into Contour. I think we absolutely need to have a solution for folks who want a managed Envoy and for those who don't. 

The controller name is just a string, just thinking of using that to tell us which controller should manage stuff, meaning should Contour provision or not. It can go in the spec somewhere as well, just trying to reduce the number of fields that need configuration.

danehans:
@stevesloka commit `3258aad` resolves your feedback above.

internal/status/gatewayclass.go
danehans: @stevesloka commit `61b743d` introduces a gateway config param to enable/disable the gatewayclass controller. I think this is a simple and explicit approach that we can iterate on if needed. I was considering using gateway ns/name ownership labels but that's a bit tricky and has it's own disadvantages.

pkg/config/parameters.go
stevesloka: The other thing to think about is how do we determine if a GatewayClass should be operator managed or not? Right now it will act on all instances, but do all users want that? The class could then be `projectcontour.io/operator` telling us that it's an operator managed instance vs a self-managed.

internal/status/gatewayclass.go
stevesloka: What does this config option do? Why would I want to disable the GatewayClass controller?

pkg/config/parameters.go
danehans: @stevesloka IMO the proper way to provide the "--enableEnvoyMgt" config knob is through `gc.spec.parametersRef`. At first, the referent can be the contour.yaml configmap that includes the knob. I think it's beneficial in the future to transition the configmap to a structured object similar to what's being done with the operator.

youngnick: I think that we should avoid magic strings in the `gatewayClass.spec.controller` field as much as possible. That string should only be human-meaningful, not parsed to determine program options like which controller should use a particular GatewayClass. Configuring things with a field, either in a configmap or an explicit structure, is much more sustainable.

danehans: > Configuring things with a field, either in a configmap or an explicit structure, is much more sustainable

@youngnick my thinking is to not make the gatewayclass controller identifier, i.e. `gc.spec.controller`, configurable. Instead, the ID is based on scheme "<API_GROUP>/<NAMESPACE>/contour". I don't understand the benefit of exposing this to a user.

pkg/config/parameters.go
```diff
@@ -491,6 +493,10 @@ type Parameters struct {
 	// Server contains parameters for the xDS server.
 	Server ServerParameters `yaml:"server,omitempty"`
 
+	// GatewayClassController is the identifier used by Contour's gatewayclass
+	// controller to determine if a GatewayClass object should be managed.
+	GatewayClassController string `yaml:"gatewayclass-controller,omitempty"`
```
stevesloka, line 498: Doesn't the gatewayclass reference give you this value automatically? We've got to have another set of config items now in addition to the gateway to configure?

pkg/config/parameters_test.go
```diff
@@ -49,6 +49,7 @@ debug: false
 kubeconfig: TestParseDefaults/.kube/config
 server:
   xds-server-type: contour
+gatewayclass-controller: projectcontour.io/projectcontour/contour
```
stevesloka, line 52: Does this need to be so verbose? Can just `projectcontour.io/contour` be sufficient?

internal/validation/gatewayclass.go
```diff
@@ -0,0 +1,37 @@
...
+// GatewayClass returns nil if gc is a valid GatewayClass, otherwise an error.
+func GatewayClass(gc *gatewayapi_v1alpha1.GatewayClass) error {
+	return parameterRef(gc)
+}
+
+// parameterRef returns nil if parametersRef of gc is valid, otherwise an error.
+func parameterRef(gc *gatewayapi_v1alpha1.GatewayClass) error {
+	// ParametersRef is optional. Default config should be used when nil.
+	if gc.Spec.ParametersRef != nil {
+		return fmt.Errorf("invalid parametersRef; field must be unspecified")
```
stevesloka, line 33: Not sure I follow why this should be unspecified? Is this where we're going to have all the configuration bits?

internal/status/conditions_test.go
```diff
@@ -49,3 +53,183 @@ func TestHTTPRouteAddCondition(t *testing.T) {
...
+	}
+
+	// Simulate the passage of time between original condition creation
+	// and update processing
+	fakeClock.SetTime(later)
+
+	for _, tc := range testCases {
+		actual := mergeConditions(tc.current, tc.updates...)
+		if conditionChanged(tc.expected[0], actual[0]) {
+			t.Errorf("expected:\n%v\nactual:\n%v", tc.expected, actual)
```
stevesloka, line 232: We've used testify `assert.Equal(t, want, got)` in other places, might simplify the error message.

internal/status/conditions_test.go
```diff
@@ -49,3 +53,183 @@ func TestHTTPRouteAddCondition(t *testing.T) {
...
+			b: metav1.Condition{
+				Type:   string(gatewayapi_v1alpha1.GatewayClassConditionStatusAdmitted),
+				Status: metav1.ConditionFalse,
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		if actual := conditionChanged(tc.a, tc.b); actual != tc.expected {
+			t.Fatalf("%q: expected %v, got %v", tc.name, tc.expected, actual)
```
stevesloka, line 164: We've used testify assert.Equal(t, want, got) in other places, might simplify the error message.

internal/equality/equality.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
+	gatewayapi_v1alpha1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
+)
+
+// GatewayClassStatusChanged checks if current and expected match and if not,
+// returns true.
+func GatewayClassStatusChanged(current, expected gatewayapi_v1alpha1.GatewayClassStatus) bool {
+	return !apiequality.Semantic.DeepEqual(current.Conditions, expected.Conditions)
```
stevesloka, line 24: We have `conditionChanged` below, could this use that to verify if something changed? Seems like we now have 2 different ways of doing comparisons.

stevesloka (changes requested):
Some more comments/questions and still a few open from last review. Thanks @danehans!

pkg/config/parameters.go
danehans: This tells the controller how to construct the `gatewayclass.spec.controller` string used for reconciling gatewayclasses. If a cluster has multiple Contours, the gatewayclass-controller for each needs to differentiate its self. For example:
```
kind: GatewayClass
metadata:
  name: public
spec:
  controller: projectcontour.io/public/contour
---
kind: Gateway
metadata:
  name: contour
  namespace: public
spec:
  gatewayClassName: public
<SNIP>
---
kind: GatewayClass
metadata:
  name: private
spec:
  controller: projectcontour.io/private/contour
---
kind: Gateway
metadata:
  name: contour
  namespace: private
spec:
  gatewayClassName: private
<SNIP>
---

We can remove the API group from the string to remove the verbosity. Note that this is internal config and not exposed to users.

File: pkg/config/parameters_test.go (hunk 13 above)
Comment (danehans) [reply pkg/config/parameters_test.go:52]: We can remove the API group from the string to remove the verbosity.

File: internal/validation/gatewayclass.go (hunk 14 above)
Comment (danehans) [reply internal/validation/gatewayclass.go:33]: This is unspecified for the time being to make this PR as small as possible. We will either introduce a config CRD or use the contour configmap to plumb config options to gateways through `gatewayclass.spec.parametersRef`. Until then, a gateway, will be created with default setting ([xref](https://github.com/projectcontour/contour/blob/main/examples/contour/02-service-envoy.yaml)).

File: pkg/config/parameters.go (hunk 12 above)
Comment (stevesloka) [reply pkg/config/parameters.go:498]: > This is internal config and not exposed to users.

It's in the public `config` package and specifies yaml parsers, so from the PR it seems like users are supposed to configure this field. 

Should we just look at the configured gateway.spec.controller and verify the class matches that? If not then skip, otherwise process? Seems like we just need a `gateway.spec.parametersRef` and skip the whole gatewayclass object all together.

File: internal/equality/equality.go (hunk 17 above)
Comment (danehans) [reply internal/equality/equality.go:24]: I created the equality package since we'll have to perform current/desired equality comparisons for different objects, fields, etc. ([xref](https://github.com/projectcontour/contour-operator/blob/main/internal/equality/equality.go)). `conditionChanged` is meant to compare 2 conditions, whereas `GatewayClassStatusChanged` is meant to compare 2 condition slices. I'll update the PR to add a private func in the status pkg to compare the condition slice of current/desired gatewayclasses.

File: pkg/config/parameters.go (hunk 10 above)
Comment (stevesloka) [reply pkg/config/parameters.go]: I was just following the docs:

> Controller is a domain/path string that indicates the controller that is managing Gateways of this class.

So we're defining which controller is the one in charge. We're introducing three different controller who are responsible. Is it the operator? Is it contour? Is it the cluster admin? Etc, etc.

File: internal/validation/gateway.go (hunk 2 above)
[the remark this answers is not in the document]
Comment (danehans) [reply internal/validation/gateway.go]: @stevesloka I'm going to consider this comment resolved since the scope of this PR was updated to focus only on the gatewayclass controller.

File: internal/dag/gatewayapi_processor.go (hunk 4 above)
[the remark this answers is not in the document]
Comment (danehans) [reply internal/dag/gatewayapi_processor.go]: > you're trying to process resources without the gateway existing

Understood, this is why the `p.source.gateway == nil` check needs to be removed from the gatewayapi processor. This is now a moot point for this PR since the scope strictly focuses on the gatewayclass controller.

File: internal/k8s/cache/gateway.go (hunk 1 above)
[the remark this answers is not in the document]
Comment (danehans) [reply internal/k8s/cache/gateway.go]: @stevesloka ack. I think it makes sense to filter as close to the source as possible, so this PR sticks with the predicate approach and does not filter gatewayclasses in the dag pkg.

File: internal/k8s/cache/gateway.go [hunk 18]
```diff
@@ -0,0 +1,337 @@
...
+		// TODO: Add method to remove gateway sub-resources and finalizer.
+		return reconcile.Result{}, nil
+	}
+
+	// Pass the new changed object off to the eventHandler.
+	r.eventHandler.OnAdd(gw)
+
+	// Check if the gateway is valid.
+	valid := true
+	if err := validation.Gateway(ctx, r.client, gw); err != nil {
```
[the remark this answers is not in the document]
danehans: @stevesloka I consider this comment resolved since this PR has been updated to strictly focus on the gatewayclass controller.

internal/k8s/cache/gateway.go
```diff
@@ -0,0 +1,337 @@
...
+			for i, p := range ports {
+				if p.Name == strings.ToLower(string(gatewayv1a1.HTTPProtocolType)) {
+					svc.Spec.Ports[i].NodePort = int32(30080)
+				}
+			}
+		}
+		if l.Protocol == gatewayv1a1.HTTPSProtocolType {
+			for i, p := range ports {
+				if p.Name == strings.ToLower(string(gatewayv1a1.HTTPSProtocolType)) {
+					svc.Spec.Ports[i].NodePort = int32(30443)
```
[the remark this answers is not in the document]
danehans: @stevesloka I consider this comment resolved since this PR has been updated to strictly focus on the gatewayclass controller.

internal/status/gatewayclass.go
danehans: @stevesloka I initially thought that the 1st Contour would have the gc controller enabled and subsequent Contours would disable the gc controller so only 1 Contour manages gc's (since the controller string is the same for all Contours). I updated the approach in this PR so each instance of Contour will run the gc controller with a unique `gc.spec.controller` that is based on the namespace Contour is running in. This will ensure multiple instances of Contour do not fight over managing a gc.

pkg/config/parameters.go
danehans: @stevesloka IMO the upstream docs need add'l details. I'm trying to define a `gc.spec.controller` naming schema for the gatewayclass contrller implemented in Contour. We have to consider that multiple Contour's may be running in the cluster, each with a gc controller. Therefore, `gc.spec.controller` should take the form of "CONTOUR_NAMESPACE/contour". I was originally thinking "CONTOUR_API_GROUP/CONTOUR_NAMESPACE/contour" but I reduced the verbosity based on https://github.com/projectcontour/contour/pull/3659#discussion_r633747716.

> We're introducing three different controller who are responsible. Is it the operator? Is it contour? Is it the cluster admin?

The number is based on the number of Contours in the cluster. Each instance of Contour will have its own `gc.spec.controller`. The operator will continue to support its existing controller string until we can deprecate Gateway API support in the operator.

internal/status/gatewayclass.go
stevesloka: Feels like Contour should just not support gatewayclass. I don't think it buys us anything. Couldn't we just map to the config file or something? It's going to be a lot of work to have leader election across controllers & test/validate that.

stevesloka: Let's have a chat about this, feels like we need to better map out how these pieces are going to fit together.

pkg/config/parameters.go
stevesloka: So is the goal @danehans to try and make the gatewayclass unique? So regardless of what the string looks like, we only have a problem if there are two gateways which reference the same gatewayclass (regardless of the controller)? Is that correct?

cmd/contour/serve.go
```diff
@@ -418,12 +418,17 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 				log.WithError(err).Fatal("unable to set up controller manager")
 			}
 
-			// Add the GatetwayAPI Scheme.
+			// Add the Gatetway API Scheme.
```
sunjayBhatia, line 421: nit: old spelling mistake would be nice to fix

examples/gateway/00-crds.yaml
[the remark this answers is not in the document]
sunjayBhatia: Commented on linked issue, would be good to get some discussion as to why this is desirable there

internal/controller/gatewayclass.go
```diff
@@ -0,0 +1,130 @@
+// Copyright Project Contour Authors
```
sunjayBhatia, line 1: I think there is starting to be enough logic in these controllers we need a way to ensure some more direct test coverage, or have some more detailed integration tests that we can be confident all the logic here is covered

internal/dag/cache.go
```diff
@@ -224,6 +225,9 @@ func (kc *KubernetesCache) Insert(obj interface{}) bool {
 	case *contour_api_v1.TLSCertificateDelegation:
 		kc.tlscertificatedelegations[k8s.NamespacedNameOf(obj)] = obj
 		return true
+	case *gatewayapi_v1alpha1.GatewayClass:
```
sunjayBhatia, line 228: I think we need some cases in cache_test.go for this files changes

sunjayBhatia (commented):
Some initial review comments

sunjayBhatia (commented):
Stepping back a bit from the changes around a GatewayClass controller reference, it seems we need to define what we're solving with that and what other issues exist once we start reconciling GatewayClasses

- An installation of Contour needs to know what GatewayClass to reconcile/admit
  - And by extension, which Gateways it should admit
  - Making the controller name configurable on a Contour instance make sense for this
  - However we also have a configurable Gateway name field, which probably makes sense to remove so we don't have an unwieldy UX and two fields to configure
- What happens when there are multiple GatewayClasses with the matching controller name Contour is configured to reconcile?
  - Assuming you can create multiple separate GatewayClasses with different names but the same controller name field, what happens? Right now it looks like this change does not do any filtering so if you were to add multiple we would process both and admit both

Just quickly thinking out loud, this is how I could see things working:
- contour config changes:
  - we remove support for the gateway name specification in contour
  - we add some config that tells a contour instance what its "controller string" to look for is  (like in this PR)
- user flow:
  - contour controller installed, with a specific controller string to look for on gatewayclasses
    - cluster operator creates a gatewayclass with controller name set to contour's, contour takes note and sets status
    - *error case*: cluster operator creates a second gatewayclass with controller name set to contour's, contour takes note and says it is not admitted, since it is younger than the first gatewayclass
      - this could change in the future if we want to refactor contour to support multiple gatewayclasses somehow, though seems improbable
    - *Below points are maybe out of scope of this PR, but related*
    - cluster operator creates a gateway referencing admitted gatewayclass above, contour takes note, starts creating data plane, sets status, etc.
    - *error case*: cluster operator creates a second gateway referencing admitted gatewayclass above, contour takes note, sets status saying it is not admitted since it is younger than first gateway
      - this could change if we want to merge the gateways, or support a single contour running multiple data planes etc.
    - app devs/operators create routes as expected, contour reconciles, etc.

cmd/contour/serve.go
danehans: I wanted to keep the PR as focused as possible, but I'll go ahead and fix spelling in my upcoming commit.

danehans:
> Commented on linked issue, would be good to get some discussion as to why this is desirable there

@sunjayBhatia I provided a response to your comment in the linked PR.

pkg/config/parameters.go
danehans: @stevesloka that is correct. When Contour supports managing multiple Envoy data-planes, we can drop the existing Gateway ns/name from the Contour config. For every gateway that references the expected gatewayclass name, the matching Contour will deploy the envoy infra (daemonset, service, etc.).

pkg/config/parameters.go
danehans: @stevesloka OK, I'm still trying to connect the dots with the Contour code. I'm trying to generate the class name for Contour based on the namespace that Contour is running, i.e. <CONTOUR_NAMESPACE>/contour or even better <CONTOUR_NAMESPACE>/<CONTOUR_DEPLOYMENT_NAME>.

internal/controller/gatewayclass.go
danehans: @sunjayBhatia the event handler is making things difficult to mock the controller's `Reconcile()` method. Looking at `cmd/contour/serve.go` a lot needs to happen to mock the `k8s.DynamicClientHandler`. Maybe it's better to use e2e for testing. Thoughts? cc: @stevesloka

internal/dag/cache.go
danehans: ack

internal/controller/gatewayclass.go
stevesloka: I can pick up the testing bit next, was going to add some since now we're adding logic into these controllers. Happy to take the initial framework bits for that.

danehans:
> Making the controller name configurable on a Contour instance make sense for this

@sunjayBhatia as I mentioned to @stevesloka, I don't think the value Contour uses for `gc.spec.controller` should be configurable. I suggest we use a naming scheme such as `CONTOUR_NAMESPACE/contour` or `CONTOUR_NAMESPACE/CONTOUR_DEPLOYMENT_NAME`. Each Contour should generate a unique `gc.spec.controller` to identify the gc that it's responsible for managing.

> And by extension, which Gateways it should admit
>
>However we also have a configurable Gateway name field, which probably makes sense to remove so we don't have an unwieldy UX and two fields to configure

In our refactored model, Contour represents a gc and envoy represents a gw. Until Contour supports managing multiple data-planes, i.e. gw's, Contour will support a 1:1 of gc-to-gw. When Contour does support managing multiple data-planes, we can remove [the gateway config](https://github.com/projectcontour/contour/blob/main/pkg/config/parameters.go#L291-L296) and Contour will watch all gw's and filter based on the `gw.spec.gatewayClassName` that matches its `gc.spec.controller: CONTOUR_NAMESPACE/contour`.  Another option is to remove the gateway config now and only allow the first gateway that matches `gc.spec.controller: CONTOUR_NAMESPACE/contour`. All subsequent gateways will be denied until Contour supports multiple data-planes.

> What happens when there are multiple GatewayClasses with the matching controller name Contour is configured to reconcile?

This PR does not prevent this scenario (trying to keep the PR small). I like your idea of using `creationTimestamp`. I'll add a TODO to address this in a follow-on PR.

> we remove support for the gateway name specification in contour

If we remove, then Contour should only manage the 1st gateway of a matching gc. Otherwise, we should keep this config until Contour supports managing multiple data-planes. Either way, this issue will be addressed in a separate PR that focuses on the gw controller.

> contour controller installed, with a specific controller string to look for on gatewayclasses

Instead of a config knob (I think users may misconfigure and cause problems), the string should be unique per Contour using a defined naming scheme `gc.spec.controller: CONTOUR_NAMESPACE/contour`, `gc.spec.controller: CONTOUR_NAMESPACE/UID`, etc..

> Below points are maybe out of scope of this PR, but related

Yes, I plan to address these steps in a follow-on PR.

youngnick:
We discussed this in our maintainer's meeting today. Here's what I remember:
- The spec says that the `controller` field should be a domain name plus a path (effectively a HTTP URL without the scheme on the front). So we should be using `projectcontour.io/whatever` as a base.
- The main requirement is that it's easy to ensure there is a unique combination expected for each Contour. Maybe we could have the `projectcontour.io/namespace/name` as a default, with "whatever is specified by the ingress-class flag" as an override? The thing I like about keeping the ingress-class flag alive is that it gives people already using it a very clear upgrade path.

danehans:
@youngnick @stevesloka @skriss @sunjayBhatia commit `1bbbcbb` implements the feedback from the maintainer's meeting yesterday.

cmd/contour/serve.go
```diff
@@ -155,6 +155,9 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 
 	serve.Flag("debug", "Enable debug logging.").Short('d').BoolVar(&ctx.Config.Debug)
 	serve.Flag("kubernetes-debug", "Enable Kubernetes client debug logging with log level.").PlaceHolder("<log level>").UintVar(&ctx.KubernetesDebug)
+
+	serve.Flag("gateway-class-name", "Contour GatewayClass name.").PlaceHolder("<api_group>/<namespace>/contour").StringVar(&ctx.gatewayClassName)
```
sunjayBhatia, line 159: maybe this should be `gateway-controller-name` or `gatewayclass-controller-name` or even `gatewayapi-controller-name`

currently this reads to me like the name of a `GatewayClass` resource

internal/controller/gatewayclass.go
```diff
@@ -0,0 +1,127 @@
...
+			r.log.WithField("name", request.Name).Info("failed to find gatewayclass")
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object, so requeue the request.
+		return reconcile.Result{}, fmt.Errorf("failed to get gatewayclass %q: %w", request.Name, err)
+	}
+
+	// Check if object is marked for deletion.
+	if !gc.ObjectMeta.DeletionTimestamp.IsZero() {
+		r.eventHandler.OnDelete(gc)
```
stevesloka, line 108: A future issue we should discuss what to do when the class is deleted. Should we drop all configurations from Envoy? Throw an error? This would be super descructive. =)

internal/controller/gatewayclass.go
```diff
@@ -0,0 +1,127 @@
...
+		return reconcile.Result{}, nil
+	}
+
+	// Pass the new changed object off to the eventHandler.
+	r.eventHandler.OnAdd(gc)
+
+	// The gatewayclass is safe to process, so check if it's valid.
+	errs := validation.ValidateGatewayClass(ctx, r.client, gc)
```
stevesloka, line 116: Should we add to the event handler on line 113 after the Valid Check?

internal/validation/gatewayclass.go
```diff
@@ -0,0 +1,71 @@
...
+
+	return errs
+}
+
+// validateGatewayClassObjMeta validates whether required fields of metadata are set according
+// to the Gateway API specification.
+func validateGatewayClassObjMeta(ctx context.Context, cli client.Client, meta *metav1.ObjectMeta, path *field.Path) field.ErrorList {
+	errs := apivalidation.ValidateObjectMeta(meta, false, ValidateGatewayClassName, path.Child("name"))
+
+	classes := &gatewayapi_v1alpha1.GatewayClassList{}
```
stevesloka, line 51: Why do we need to check for other gatewayclasses? I'd expect "Validate" would just validate the one that we care about and we'd ignore all others. In this example now, we'd only care about classes that match `projectcontour.io/projectcontour/contour`? If there are more classes, then we'd need to validate against that controller but I don't see where the controller is pulled in?

internal/dag/cache.go
```diff
@@ -59,6 +59,7 @@ type KubernetesCache struct {
 	tlscertificatedelegations map[types.NamespacedName]*contour_api_v1.TLSCertificateDelegation
 	services                  map[types.NamespacedName]*v1.Service
 	namespaces                map[string]*v1.Namespace
+	gatewayclass              *gatewayapi_v1alpha1.GatewayClass
```
stevesloka, line 62: Might not need to add this to the cache, but for now it's fine until we determine we don't need it here.

internal/controller/gatewayclass.go
```diff
@@ -0,0 +1,127 @@
...
+	// Check if object is marked for deletion.
+	if !gc.ObjectMeta.DeletionTimestamp.IsZero() {
+		r.eventHandler.OnDelete(gc)
+		return reconcile.Result{}, nil
+	}
+
+	// Pass the new changed object off to the eventHandler.
+	r.eventHandler.OnAdd(gc)
+
+	// The gatewayclass is safe to process, so check if it's valid.
```
sunjayBhatia, line 115: should this be before the `OnAdd` call?

cmd/contour/serve.go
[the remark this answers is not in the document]
stevesloka: If we do find we need this then it should come from the config file, not a flag. We've been adding all those bits to the config file that needs configuring.

sunjayBhatia:
> One question is where does the magic string of the controller name get filtered? I couldn't find that bit which I think is important for this or a future PR depending on your order. Thanks!

I think this comes in here: https://github.com/projectcontour/contour/blob/1bbbcbb5fadfaa2ae3cc1181ccaf053b860690e5/internal/controller/gatewayclass.go#L61

https://github.com/projectcontour/contour/blob/1bbbcbb5fadfaa2ae3cc1181ccaf053b860690e5/internal/controller/gatewayclass.go#L68-L90

danehans:
> One question is where does the magic string of the controller name get filtered?

@stevesloka the string is passed from the serveContext in cmd/serve.go to `NewGatewayClassController()`, which in turn passes the string to the `enqueueRequestForGatewayClass()` predicate.

internal/controller/gatewayclass.go
danehans: xref: https://github.com/projectcontour/contour/issues/3700

internal/controller/gatewayclass.go
danehans: ack

stevesloka: The other thought as well related to #3700 is if the class was valid previously, then becomes invalid, what should happen? Doesn't need to be handled here per say, but just thinking out loud. =)

internal/validation/gatewayclass.go
danehans: @stevesloka a gatewayclass is invalid if one is already admitted with the same spec.controller. I did update the func as the one you reviewed did not check the controller and status conditions of existing gatewayclasses.

internal/dag/cache.go
danehans: Yeah, I did not understand the use for it, but I thought it would be best to follow existing patterns.

cmd/contour/serve.go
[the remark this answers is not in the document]
danehans: > From the last meeting, was this intended to be convention?

@stevesloka from my understanding `<api_group>/<namespace>/contour` was the agreed upon convention.

> The GatewayClass name would match how Contour is deployed? 

Yes, the namespace that Contour is deployed.

> Also should replace <api_group> with projectcontour.io since that's what's expected.

I used <api_group> since this portion of the string is derived from the API group. I'll update to the group string.

> If we do find we need this then it should come from the config file, not a flag.

Per feedback from the meeting, I was following how the ingress class name is configured. I'll update to use the config file.

danehans: Per https://github.com/projectcontour/contour/pull/3659#discussion_r637055343, I moved this to the config file and renamed to `ControllerName`.

internal/controller/gatewayclass.go
danehans: ack

stevesloka (approved):
/lgtm 🎉 

I think we can follow up with more but would be nice to get this in and expand from there. Thanks, @danehans!!

internal/validation/gatewayclass_test.go
```diff
@@ -0,0 +1,165 @@
...
+			tc.mutateOld(oldMutated)
+
+			if err := cl.Create(context.TODO(), oldMutated); err != nil {
+				t.Fatalf("failed to create gatewayclass %q: %v", oldMutated.Name, err)
+			}
+		}
+
+		actual := ValidateGatewayClass(ctx, cl, newMutated, newMutated.Spec.Controller)
+		if actual != nil && tc.expect {
+			t.Fatalf("%q: expected %#v, got %v", tc.name, tc.expect, actual)
```
sunjayBhatia, line 159: could use `t.Run` to run these different cases and use the name of the tests instead of adding here

internal/validation/gatewayclass_test.go
```diff
@@ -0,0 +1,165 @@
...
+			oldMutated := old.DeepCopy()
+			tc.mutateOld(oldMutated)
+
+			if err := cl.Create(context.TODO(), oldMutated); err != nil {
+				t.Fatalf("failed to create gatewayclass %q: %v", oldMutated.Name, err)
+			}
+		}
+
+		actual := ValidateGatewayClass(ctx, cl, newMutated, newMutated.Spec.Controller)
+		if actual != nil && tc.expect {
```
sunjayBhatia, line 158:
```suggestion
		if tc.expect {
		    assert.Nil(t, actual, "descriptive message if needed")
```

internal/validation/gatewayclass_test.go
```diff
@@ -0,0 +1,165 @@
...
+			t.Fatalf("%q: expected %#v, got %v", tc.name, tc.expect, actual)
+		}
+		if actual == nil && !tc.expect {
```
sunjayBhatia, line 161: similar suggestion to the above around using `assert.NotNil`

do we want to assert on the contents of the errors?

internal/validation/gatewayclass.go
```diff
@@ -0,0 +1,81 @@
...
+	var errs field.ErrorList
+
+	classes := &gatewayapi_v1alpha1.GatewayClassList{}
+	if err := cli.List(ctx, classes); err != nil {
+		errs = append(errs, field.InternalError(path, fmt.Errorf("failed to list gatewayclasses: %v", err)))
+		return errs
+	}
+
+	// Consider the gatewayclass invalid if another is admitted with the same controller.
+	for _, item := range classes.Items {
```
sunjayBhatia, line 64: if we were to create multiple gatewayclasses with the same controller name before Contour is running, I'm not sure if there is a defined behavior to which one is admitted/valid, it depends on the one that is reconciled first, which may not be the oldest etc. right?

if we don't already have a follow up issue for adding logic and tests around that would be good to add

sunjayBhatia (commented):
I think before we move too far forward with more of the controller pattern, it would be good to get some unit/functional testing and some more testing on the order of the `featuretests` package or something

by moving to the controller/reconcile pattern we've sort of skirted around the coverage the `featuretests` package provides it seems and should rectify that

having tests in this style makes it really easy to show a more end to end flow and what should happen when certain resources are added and to do positive and negative cases

also would be good to collect/list (in the PR description maybe) and have issues for all the follow ups we intend to do since this PR is big already but we want to address some things and not block this too much

danehans:
@sunjayBhatia I created an issue regarding the controller testing, added a link to the PR description and assigned @stevesloka since offered to take on the task in a follow-on PR ([xref](https://github.com/projectcontour/contour/pull/3659#discussion_r635268432)).

danehans:
@sunjayBhatia thanks for the review. Commit `12aad68` resolves your feedback above.

danehans:
@stevesloka commit `ab6445d` makes the `controllerName` optional so existing Gateway API users can upgrade from a previous version without failing config validation.

sunjayBhatia (approved):

stevesloka:
Nice work @danehans! 🎉

PR #3662, opened 2021-05-10
test/e2e: add retry with backoff to all updates

### LINKED ISSUE #3660: test/e2e test flake
I've seen a couple instances of a test flake with the health check test:

https://github.com/projectcontour/contour/pull/3658/checks?check_run_id=2549283488#step:4:1007

We should either (a) move to `Patch` instead of `Update`, or (b) add retry with backoff for updates.

skriss:
Closes #3660.

### DIFF
--- a/test/e2e/httpproxy/015_http_health_checks_test.go
+++ b/test/e2e/httpproxy/015_http_health_checks_test.go
@@ -22,6 +22,7 @@ import (
 	"github.com/projectcontour/contour/test/e2e"
 	"github.com/stretchr/testify/require"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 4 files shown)


test/e2e/httpproxy/015_http_health_checks_test.go
```diff
@@ -64,11 +65,17 @@ func testHTTPHealthChecks(fx *e2e.Framework) {
...
+		if err := fx.Client.Get(context.TODO(), client.ObjectKeyFromObject(p), p); err != nil {
+			return err
+		}
+
+		p.Spec.Routes[0].HealthCheckPolicy = &contourv1.HTTPHealthCheckPolicy{
+			Path: "/status/418",
+		}
+
+		return fx.Client.Update(context.TODO(), p)
+	}))
```
skriss, line 78: Since this logic is ~repeated (get object, apply changes, update object), we could extract this to a test framework helper method. I was undecided on whether it was worth it or not, since you'd have to use some type-assertions or closure tricks to make it generic for types beyond just `HTTPProxy`.

youngnick: The status updating code had a similar problem, and I used a Mutator function there to handle it, I'm not sure how good I feel about it, looking back. Although this is repetitive, this does have the advantage of clarity. I agree we should leave this for now, and see if we should dry it up later.

youngnick (approved):
LGTM, I think that doing retry with backoff is a good habit anyway, leaving aside concerns about repetition.

skriss:
cc @danehans we may want to do something similar in the operator e2e's to try to avoid some of the flakes we've been seeing.

sunjayBhatia (approved):
sounds good to me, can DRY this up later if we need to

PR #3663, opened 2021-05-10
internal/status: Refactor status tests to allow for FullName to be validated

stevesloka:
Following up to #3636, refactors the status tests to allow for the the FullName
to be validated along with the Conditions so that if there are two different
objects, each will have a set of conditions to be asserted against.


skriss (approved):
LGTM

PR #3667, opened 2021-05-11
test/e2e: add HTTPProxy tests 001, 011

skriss:
Updates #3621.

### DIFF
--- a/test/e2e/httpproxy/001_required_field_validation_test.go
+++ b/test/e2e/httpproxy/001_required_field_validation_test.go
@@ -1,0 +1,150 @@
+// Copyright Project Contour Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build e2e
+
+package httpproxy
+
+import (
+	"context"
+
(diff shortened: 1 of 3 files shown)


test/e2e/httpproxy/001_required_field_validation_test.go
```diff
@@ -0,0 +1,150 @@
...
+				},
+			},
+		},
+	}
+
+	err := fx.Client.Create(context.TODO(), missingConditionHeaderName)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "spec.routes.conditions.header.name: Required value")
+
+	// This HTTPProxy is expressed as an Unstructured because the JSON
```
stevesloka, line 74: 😱

skriss:
Looks like another test flake with the pod restart test where the original pod doesn't delete within `60s`. I'll bump the timeout for that since I've seen it a couple other times.

test/e2e/httpproxy/001_required_field_validation_test.go
skriss: Yeah this isn't the greatest, but necessary right now. If we tweak the kubebuilder validations (e.g. apply a minLength of 1) and/or add `omitempty` JSON tags, then this can go away, but I wanted to take a second to think about those changes before making them since they're in the API package.

skriss:
@sunjayBhatia ping for a review when you get a chance, thanks!

test/e2e/httpproxy/001_required_field_validation_test.go
sunjayBhatia: yeah, probably fine to leave as is and file an issue for the api change

sunjayBhatia (approved):
LGTM, just one comment

test/e2e/httpproxy/001_required_field_validation_test.go
skriss: Good call I'll do that

skriss: https://github.com/projectcontour/contour/issues/3674

PR #3670, opened 2021-05-11
Draft: A bit more ginkgo-ey style tests and helper wrappers

sunjayBhatia:
While messing around with the e2e test suite I started gravitating to doing thinks in a bit more ginkgo-ey style

Summary/some highlights:
- each file can have a top level `Describe` to register the tests in that file which is a bit more idiomatic ginkgo IMO
- added a framework wrapper for running a test in a unique namespace
- added a gateway package helper for running a block of tests with a HTTP Gateway
- framework can be a global test package variable to share across the package
- converted a few tests over so we can see if we like the patterns

### DIFF
--- a/test/e2e/framework.go
+++ b/test/e2e/framework.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/apimachinery/pkg/util/wait"
 	kubescheme "k8s.io/client-go/kubernetes/scheme"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -66,7 +67,8 @@ type Framework struct {
 	t ginkgo.GinkgoTInterface
 }
 
-func NewFramework(t ginkgo.GinkgoTInterface) *Framework {
+func NewFramework() *Framework {
+	t := ginkgo.GinkgoT()
 	scheme := runtime.NewScheme()
 	require.NoError(t, kubescheme.AddToScheme(scheme))
 	require.NoError(t, contourv1.AddToScheme(scheme))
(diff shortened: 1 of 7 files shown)


test/e2e/framework.go
```diff
@@ -133,6 +135,25 @@ func NewFramework(t ginkgo.GinkgoTInterface) *Framework {
 	}
 }
 
+// NamespacedTest generates a unique namespace name and sets up
+// a Before/AfterEach to create and delete the namespace before/after the
+// test body. Intended to be used to wrap a single test.
+func (f *Framework) NamespacedTest(body func(string)) func() {
```
skriss, line 141: I definitely like the wrapper, but I think I prefer non-random names, makes it a little easier if debugging tests locally to find what you're looking for, especially if we're running things concurrently.

test/e2e/gateway/005_request_header_modifier_test.go
```diff
@@ -18,241 +18,238 @@ package gateway
...
-						{
-							Path: &gatewayv1alpha1.HTTPPathMatch{
-								Type:  pathMatchTypePtr(gatewayv1alpha1.PathMatchPrefix),
-								Value: stringPtr("/filter"),
-							},
-						},
+var _ = Describe("005-gateway-request-header-modifier", func() {
+	withHTTPGateway(func() {
+		Context("header modification on HTTPRoute rule ForwardTo", f.NamespacedTest(func(namespace string) {
+			It("works", func() {
```
skriss, line 32: One of the things I don't like about Ginkgo is how much nesting it seems to lead to, just a whole lot of whitespace. WDYT about moving the bodies of the specs into their own functions that can be called from within the `It(...)`s? I think the downside is you have to explicitly pass any closure variables.

test/e2e/gateway/005_request_header_modifier_test.go
```diff
@@ -18,241 +18,238 @@ package gateway
...
-				{
-					Matches: []gatewayv1alpha1.HTTPRouteMatch{
-						{
-							Path: &gatewayv1alpha1.HTTPPathMatch{
-								Type:  pathMatchTypePtr(gatewayv1alpha1.PathMatchPrefix),
-								Value: stringPtr("/filter"),
-							},
-						},
+var _ = Describe("005-gateway-request-header-modifier", func() {
+	withHTTPGateway(func() {
```
skriss, line 30: makes sense

test/e2e/framework.go
skriss: Also some tests use >1 namespace since they're testing cross-namespace things, so those could either have custom handling or you could allow `NamespacedTest` to set up/tear down multiple namespaces

test/e2e/gateway/gateway_test.go
```diff
@@ -26,18 +26,14 @@ import (
 	gatewayv1alpha1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
 )
 
+var f = e2e.NewFramework()
```
skriss, line 29: makes sense

skriss (commented):
thanks for working on this @sunjayBhatia! Had a few comments

test/e2e/gateway/005_request_header_modifier_test.go
sunjayBhatia: yeah, the indenting gets especially bad with big go structs for the resources we create in k8s

sunjayBhatia:
closing as this is mostly just a reference, ideas/changes should be taken into https://github.com/projectcontour/contour/pull/3803

PR #3676, opened 2021-05-13
Adds Make Target to Sync Gateway API CRDs

danehans:
This PR allows for the same [getting started](https://projectcontour.io/getting-started/#option-1-quickstart) user experience when using Contour with Gateway API. For example:
`kubectl apply -f https://projectcontour.io/quickstart/contour-gateway.yaml`

- Adds a make target to keep Gateway API CRDs in sync with the consumed go module version.
- Updates kubectl version to v1.21.1 (required for "-o" flag and aligns toolchain with [compatibility matrix](https://projectcontour.io/resources/compatibility-matrix/)).

xref: https://github.com/projectcontour/contour/pull/3659


stevesloka:
I'm not sure if we want to sync these in and maintain a copy of them in the Contour repo. Feels like we should reference from upstream links since we're not modifying the CRDs in any way. We need to think about versions as well going forward and how to manage those as the GatewayAPI's evolve. 

I'm curious what @skriss & @sunjayBhatia think. If we're going this route, then the CI runs and other places need updated to remove the upstream reference.

sunjayBhatia:
yeah, what is the advantage of keeping the CRDs in this repo vs just using the version we support from upstream?

I don't know if you install contour that you also *need* to install the gateway api crds, unless that is something we want as an explicit goal to get people to use them or something

sunjayBhatia:
Can you expand on why this is desirable @danehans? Seems like more surface area for us to keep updated and in sync

danehans:
> yeah, what is the advantage of keeping the CRDs in this repo vs just using the version we support from upstream?

@sunjayBhatia it allows for the same [getting started](https://projectcontour.io/getting-started/#option-1-quickstart) user experience when using Contour with Gateway API. For example:
`kubectl apply -f https://projectcontour.io/quickstart/contour-gateway.yaml`

Otherwise it's a 2-step process, install the supported version of the Gateway API CRDs and then install the example Contour manifests.

Optionally, we can tailor the script to be more selective regarding what CRDs to install. For example, do not install TCPRoute and UDPRoute since they're unsupported by Contour.

> I don't know if you install contour (via example manifests etc.) that you also need to install the gateway api crds, unless that is something we want as an explicit goal to get people to use them or something

If the Gateway API CRDs are not installed, then the Gateway API-related controllers will not be enabled.

With or without this PR, I suggest having 2 different example manifests. The current example that installs Contour and a 2nd example that installs Contour configured for Gateway API and the necessary instances of Gateway API resources (and optionally Gateway API CRDs). The difference between the two will be minimal at first, but the gap will widen over time.

Makefile
```diff
@@ -212,9 +214,15 @@ generate-deployment:
...
 generate-crd-yaml:
-	@echo "Generating CRD YAML documents..."
+	@echo "Generating Contour CRD YAML documents..."
 	@./hack/generate-crd-yaml.sh
 
+.PHONY: generate-gateway-crd-yaml
+generate-gateway-crd-yaml:
+	@echo "Generating Gateway API CRD YAML documents..."
+	@kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd?ref=${GATEWAY_API_VERSION}" \
+-o examples/gateway/00-crds.yaml
```
sunjayBhatia, line 224: nit: if possible indent this so it falls indented under the `kubectl` command

Makefile
```diff
@@ -193,7 +194,8 @@ lint-flags:
 
 .PHONY: generate
 generate: ## Re-generate generated code and documentation
-generate: generate-rbac generate-crd-deepcopy generate-crd-yaml generate-deployment generate-api-docs generate-metrics-docs generate-uml
+generate: generate-rbac generate-crd-deepcopy generate-crd-yaml generate-deployment generate-api-docs \
+generate-metrics-docs generate-uml generate-gateway-crd-yaml
```
sunjayBhatia, line 198: nit: would be good to indent this or keep it on one line

sunjayBhatia (changes requested):
i think the justification of why to do is reasonable and could make sense as a goal, but would be good to have in the PR description etc. up front so we can come to consensus on whether the goal this has in mind is agreed on

change seems reasonable, just a couple comments

also would be good to add to https://github.com/projectcontour/contour/blob/96efa022e0544283838f89e1890e81633f513ecf/hack/actions/check-uncommitted-codegen.sh#L10-L16 to ensure any changes to the gateway examples cause a CI failure

youngnick:
I think this one needs a discussion in a maintainer's meeting or the Community meeting.

danehans:
@youngnick we discussed this topic during the community meeting and the feedback was to have a similar user experience as the current getting started guide, so we need to sync the Gateway API CRDs. We can discuss this further since this PR is not required for v1.16.0.

danehans:
@sunjayBhatia commit `8fa1b99` adds the gateway example directory to `contour/hack/actions/check-uncommitted-codegen.sh` and resolves your other feedback, PTAL.

sunjayBhatia (commented):
looks like a CI failure still https://github.com/projectcontour/contour/pull/3676/checks?check_run_id=2738741101

weird, since that same command worked fine for me

Makefile
```diff
@@ -8,6 +8,7 @@ LOCAL_BOOTSTRAP_CONFIG = localenvoyconfig.yaml
 SECURE_LOCAL_BOOTSTRAP_CONFIG = securelocalenvoyconfig.yaml
 PHONY = gencerts
 ENVOY_IMAGE = docker.io/envoyproxy/envoy:v1.18.3
+GATEWAY_API_VERSION = $(grep "sigs.k8s.io/gateway-api" go.mod | awk '{print $2}')
```
sunjayBhatia, line 11: this isn't actually working, I get an empty string when i make the kubectl command below echo

```
kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd?ref=" -o examples/gateway/00-crds.yaml
```

sunjayBhatia, line 11:
```suggestion
GATEWAY_API_VERSION = $(shell grep "sigs.k8s.io/gateway-api" go.mod | awk '{print $$2}')
```

sunjayBhatia (changes requested):

Makefile
sunjayBhatia: turns out we were pulling the latest `main` or whatever CRD yaml as well, we'll need to regenerated the CRDs as well before this will pass the check

sunjayBhatia:
@danehans tried to run `kubectl kustomize <arg> -o <outfile>` and it fails with kubectl 1.20.4 which we are currently installing

going to need to bump the kubectl version here: https://github.com/projectcontour/contour/blob/e342a71f3c4156b1332bc57231703a3f8dab0cbe/hack/actions/install-kubernetes-toolchain.sh#L8 I have no versions with my later version

danehans:
@sunjayBhatia thanks for the pointer. I updated the PR based on your comments.

sunjayBhatia:
> @sunjayBhatia thanks for the pointer. I updated the PR based on your comments.

hmm something weird going on, why did it fail again the same way....

sunjayBhatia:
> > @sunjayBhatia thanks for the pointer. I updated the PR based on your comments.
> 
> hmm something weird going on, why did it fail again the same way....

oh, this CI job doesnt do the k8s tools install in the pipeline here: https://github.com/projectcontour/contour/blob/e342a71f3c4156b1332bc57231703a3f8dab0cbe/.github/workflows/prbuild.yaml#L204-L224

another addition needed in that job: https://github.com/projectcontour/contour/blob/e342a71f3c4156b1332bc57231703a3f8dab0cbe/.github/workflows/prbuild.yaml#L113-L116

sunjayBhatia:
could also just use redirection and not the flag tbh, so the kubectl version doesnt matter 🤷🏽

sunjayBhatia (approved):

danehans:
> could also just use redirection and not the flag tbh, so the kubectl version doesnt matter 🤷🏽

@sunjayBhatia Thanks again for the review and guidance. I thought it made sense to update kubectl to match the latest k8s version of the [compatability matrix](https://projectcontour.io/resources/compatibility-matrix/).

youngnick (approved):

PR #3678, opened 2021-05-13
Add initial upgrade test suite

sunjayBhatia:
Adds an upgrade test suite (only 1 test for now) that runs against a cluster with a contour installed, checks routability, upgrades contour, then checks routability again.

Unmarshals current deployment yaml to update the contour deployment.

Tests in this change can be run with:

```
CONTOUR_UPGRADE_FROM_VERSION=v1.15.0 CONTOUR_UPGRADE_TO_IMAGE=projectcontour/contour:<some-image tag> make upgrade
```

Make target installs kind cluster, installs "from" contour version, loads image into cluster, and runs tests


test/e2e/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,342 @@
...
+	contourClusterRole        *rbac_v1.ClusterRole
+	contourService            *v1.Service
+	envoyService              *v1.Service
+	contourDeployment         *apps_v1.Deployment
+	envoyDaemonSet            *apps_v1.DaemonSet
+}
+
+// Unmarshals resources from rendered Contour manifest in order and updates
+// each.
+func updateContourDeploymentResources() contourDeploymentResources {
```
youngnick, line 178: This is really neat, seems like it will need to be updated if we add any bits to the example YAML though? If so, Not a dealbreaker, since the job will fail then, but maybe worth a comment?

youngnick (approved):
LGTM aside from a small doc request. Nice work.

.github/workflows/prbuild.yaml
```diff
@@ -106,6 +106,54 @@ jobs:
...
+            ${{ runner.os }}-${{ github.job }}-go-
+      - name: add deps to path
+        run: |
+          ./hack/actions/install-kubernetes-toolchain.sh $GITHUB_WORKSPACE/bin
+          echo "$GITHUB_WORKSPACE/bin" >> $GITHUB_PATH
+      - name: upgrade tests
+        env:
+          NODEIMAGE: ${{ matrix.node_image }}
+          LOAD_PREBUILT_IMAGE: "true"
+          CONTOUR_UPGRADE_FROM_VERSION: v1.15.1
```
stevesloka, line 153: Should add this to the release notes somewhere so we don't forget to change it.

stevesloka (commented):
Just one small comment but looks neat to me. Would also be nice to have a multi-node k8s cluster, but that again can be a future iteration. =)

.github/workflows/prbuild.yaml
sunjayBhatia: hmm yeah good point, i'll add it to the release-process.md doc to start?

skriss: documenting for now sounds good. we might be able to compute this -- one way off the top of my head is to curl `https://github.com/projectcontour/contour/releases/latest`, follow the redirect, and grab the version from there

skriss: Actually you already have code to compute it from git tags, do we even need to pass it explicitly here?

sunjayBhatia: ah yeah i forgot about that, it wasn't working in CI b/c of the way a PR branch is cloned but I can try to get it to work

skriss (approved):
LGTM, great to get this in!

test/e2e/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,344 @@
...
+
+	const appHost = "upgrade-echo.test.com"
+
+	BeforeEach(func() {
+		namespace = "contour-upgrade-test"
+		f.CreateNamespace(namespace)
+
+		By("deploying an app")
+		f.Fixtures.Echo.Deploy(namespace, "echo")
+		i := &networking_v1.Ingress{
```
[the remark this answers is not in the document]
sunjayBhatia: no reason was just quicker to write up when i did it, but yeah we should probably have some SLO with all the different resources we support

thinking in the future we have a check to ensure N seconds after a resource is created, its route should be configured, some number of 9s percent of the time

.github/workflows/prbuild.yaml
sunjayBhatia: should be fixed now

test/e2e/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,344 @@
...
+		f.DeleteNamespace(namespace)
+	})
+
+	Specify("applications remain routable after the upgrade", func() {
+		resources := updateContourDeploymentResources()
+
+		By("waiting for contour deployment to be updated")
+		require.Eventually(f.T(), func() bool {
+			require.NoError(f.T(), f.Client.Get(context.TODO(), client.ObjectKeyFromObject(resources.contourDeployment), resources.contourDeployment))
+			return resources.contourDeployment.Status.Replicas == *resources.contourDeployment.Spec.Replicas
```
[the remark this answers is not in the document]
sunjayBhatia: yeah, CI is passing too fast, no way it has updated the deployment in 2s :(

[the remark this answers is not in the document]
sunjayBhatia: should hopefully be better now, checking contour pods are updated with the container image and are in the ready state

skriss (approved):
nice, the update looks good, LGTM!

Issue #3681, opened 2021-05-14
CI pipeline/job improvements


sunjayBhatia:
Related to https://github.com/projectcontour/contour/issues/3679 there are various improvements we may be able to make in order to have more confidence in our CI jobs/pipelines, speed them up, and follow best practices

- [x] Use dependencies between jobs more often to pipeline operations to reduce wasted effort/time to get results
  - e.g. integration job will depend on an image build job, if that fails, don't bother with expensive integration tests
- [ ] Build Contour image for e2e/integration/push to registry etc. *once* so we are testing and shipping the same artifact
  - Also should reduce the time CI takes, each run does not have to rebuild a container image
- [x] Use github actions build caches to speed up compilation
  - See Steve K's WIP here: https://github.com/skriss/contour/commit/cc2bc154a3c0468c718df40f499136de4e256bf7

skriss:
@sunjayBhatia I think this is all done so we can close this out, agree?

sunjayBhatia:
yeah, i think maybe one improvement could still be to block image push on CI/push the tested image but that can be pulled into its own issue

skriss:
> yeah, i think maybe one improvement could still be to block image push on CI/push the tested image but that can be pulled into its own issue

Makes sense, I misread that second bullet the first time. Opened #4317 to specifically cover that.

PR #3683, opened 2021-05-14
Build Contour image before e2e/integration jobs

sunjayBhatia:
- Image is build once and loaded into kind clusters in each case
- Modifies multiarch build target to optionally push image
- Now outputs an OCI tar archive if push not specified
- OCI image is uploaded from build step, used in test steps later
- Still retains ability to build image as before when invoking make e2e
  etc.
- Next step will be to make image push jobs (on main and tags) use the
  output of the image build job and those jobs should be blocked on
  tests passing
- Eventually if we end up with a test suite that runs the contour binary
  itself (no container image or unit test etc.) we can extract the binary
  from the build image and run that, so we are testing against the
  artifact we are actually shipping

Updates: #3681


_integration/testsuite/install-contour-working.sh
```diff
@@ -39,30 +41,44 @@ kind::cluster::exists() {
...
+else
+    # Build the current version of Contour.
+    VERSION="v$$"
+    make -C ${REPO} container IMAGE=docker.io/projectcontour/contour VERSION=$VERSION
+
+    for t in $TAGS ; do
+        docker tag \
+            docker.io/projectcontour/contour:"v$$" \
+            docker.io/projectcontour/contour:$t
+    done
```
skriss, line 74: this seems unnecessary now since we're updating the YAML to point to the specific version tag. Shouldn't hurt anything so not a big deal either way.

_integration/testsuite/install-contour-working.sh
```diff
@@ -39,30 +41,44 @@ kind::cluster::exists() {
...
+
+    # Push the Contour build image into the cluster.
+    kind::cluster::load::docker docker.io/projectcontour/contour:${VERSION}
+    for t in $TAGS ; do
+        kind::cluster::load::docker docker.io/projectcontour/contour:$t
+    done
```
skriss, line 80: ditto for this

skriss (approved):
LGTM; one optional comment.

youngnick (approved):
This looks great, really nice work @sunjayBhatia!

PR #3688, opened 2021-05-18
internal/status: set a Controller on route status updates

### LINKED ISSUE #3641: HTTPRoute status updates are failing in Kubernetes 1.19.7
Seeing this error in the Contour log when it's trying to update `HTTPRoute` status on 1.19.7:
```
time="2021-05-05T17:02:01Z" level=error msg="unable to update status" context=StatusUpdateHandler error="HTTPRoute.networking.x-k8s.io \"http-filter-1\" is invalid: status.gateways.gatewayRef.controller: Invalid value: \"null\": status.gateways.gatewayRef.controller in body must be of type string: \"null\"" name=http-filter-1 namespace=gateway-005-request-header-modifier-rule
```

It's marked as optional, so not sure why this would be erroring, esp. only on 1.19.x, makes me think maybe there's an upstream issue. @stevesloka any ideas here?

_Originally posted by @skriss in https://github.com/projectcontour/contour/pull/3638#issuecomment-832835980_

skriss:
Closes #3641.

### DIFF
--- a/internal/status/conditions.go
+++ b/internal/status/conditions.go
@@ -19,6 +19,7 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/pointer"
 	gatewayapi_v1alpha1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
 )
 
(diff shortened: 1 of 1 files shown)


stevesloka (approved):

skriss:
Definitely fixed the issue given that E2E's are running faster which means that `HTTPRoute` status updates are going through.

sunjayBhatia (approved):

PR #3691, opened 2021-05-18
.github/workflows: set stale PR close time to 30 days

skriss:
The current settings are to mark PRs stale after 14 days
of no activity, and then to close after another 90 days
of no activity. This lowers the time to close after being
marked stale to 30 days, so stale PRs don't hang around
for so long.


skriss:
If you all prefer keeping 90 days, that's OK; I just thought 30 was plenty long and cleaned up stale stuff sooner.

sunjayBhatia (approved):
I’m cool with 30days 👌🏽

stevesloka (approved):

PR #3699, opened 2021-05-20
CI: Speed up Go builds in CI by using Github caching

sunjayBhatia:
- Cache for each job that invokes Go
- Namespaced by OS to be safe
- Each job needs its own cache so they don't clobber each other; otherwise jobs that finish first take over the cache, e.g. the `codegen` job would own the cache for all Linux jobs and the cache would be useless for the `test-linux` job
- Speeds up `test-linux/osx` jobs significantly (the former down to ~1 minute)

### DIFF
--- a/.github/workflows/prbuild.yaml
+++ b/.github/workflows/prbuild.yaml
@@ -50,6 +50,16 @@ jobs:
         with:
           name: image
           path: image
+      - uses: actions/cache@v2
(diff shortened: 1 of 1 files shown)


skriss (approved):
awesome, LGTM!

youngnick (approved):
LGTM, really nice.

PR #3701, opened 2021-05-21
gateway wildcard hostname e2e test

stevesloka:

### DIFF
--- a/test/e2e/gateway/009_tls_wildcard_host_test.go
+++ b/test/e2e/gateway/009_tls_wildcard_host_test.go
@@ -1,0 +1,116 @@
+// Copyright Project Contour Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build e2e
+
+package gateway
+
+import (
+	"crypto/tls"
+
+	gatewayv1alpha1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
+
+	"github.com/projectcontour/contour/test/e2e"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func testTLSWildcardHost(fx *e2e.Framework) {
+	t := fx.T()
+	namespace := "009-gateway-tls-wildcard-host"
+	hostSuffix := "wildcardhost.gateway.projectcontour.io"
+
+	fx.CreateNamespace(namespace)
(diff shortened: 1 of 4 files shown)


test/e2e/gateway/009_tls_wildcard_host_test.go
```diff
@@ -0,0 +1,121 @@
...
+func testTLSWildcardHost(fx *e2e.Framework) {
+	t := fx.T()
+	namespace := "009-gateway-tls-wildcard-host"
+	hostSuffix := "wildcardhost.gateway.projectcontour.io"
+
+	fx.CreateNamespace(namespace)
+	defer fx.DeleteNamespace(namespace)
+
+	fx.Fixtures.Echo.Deploy(namespace, "echo")
+	fx.Certs.CreateSelfSignedCert(namespace, "echo-one-cert", "echo-one-cert", "*."+hostSuffix)
```
skriss, line 37: Does this cert need to be configured on the Gateway? If so, see https://github.com/projectcontour/contour/blob/5e45ddc6a35b95810d5d3e86f3dd91d83f10ccf0/test/e2e/gateway/gateway_test.go#L144-L150 and https://github.com/projectcontour/contour/blob/5e45ddc6a35b95810d5d3e86f3dd91d83f10ccf0/test/e2e/gateway/gateway_test.go#L166 for how that's done. You'd need a new [Describe](https://github.com/projectcontour/contour/blob/5e45ddc6a35b95810d5d3e86f3dd91d83f10ccf0/test/e2e/gateway/gateway_test.go#L113) block for the Gateway with the wildcard cert

test/e2e/gateway/009_tls_wildcard_host_test.go
```diff
@@ -0,0 +1,121 @@
...
+		t.Logf("Making request with hostname=%s, sni=%s", tc.hostname, tc.sni)
+
+		res, ok := fx.HTTP.SecureRequestUntil(&e2e.HTTPSRequestOpts{
+			Host: tc.hostname,
+			TLSConfigOpts: []func(*tls.Config){
+				e2e.OptSetSNI(tc.sni),
+			},
+			Condition: e2e.HasStatusCode(tc.wantStatus),
+		})
+		require.Truef(t, ok, "expected %d response code, got %d", tc.wantStatus, res.StatusCode)
```
sunjayBhatia, line 119: panic might be resolved by the cert/tls being set up on the gateway, not sure if we should do a "require" `res` to be non-nil so we can see the error w/o panics

sunjayBhatia (commented):
LGTM generally once the gateway/panic is fixed 👍🏽 

curious to see if the `a.random3` test passes, my guess is we will need the same header match hack as for Ingress

stevesloka:
The a.random3 test indeed fails @sunjayBhatia. =)

I was using this PR to run the tests since they don't work on my m1 mac. Need to go look at your wildcard hostname hack to fix this up.

sunjayBhatia (approved):
LGTM

I wonder if a future refactor could be done to abstract the header match needed for the wildcard host more, i didnt do it at first for Ingress so we could see if doing it again for another resource made things unwieldy

- maybe at the very least a helper function that generates the regex match rule so we don't have to repeat much code?
- make `dag.Route` wildcard aware or only add the header match rule on conversion to Envoy type?
  - would have to take care to consider sorting of routes if we change things

PR #3707, opened 2021-05-25
internal: add optional TLS support for contour metrics endpoint

### LINKED ISSUE #3594: Provide the ability to protect the metrics endpoints with TLS
The metrics endpoints in Envoy and Contour are currently exposed as unauthenticated HTTP.  This could have some security related consequences:

1. An attacker might find some sensitive information that they can use for their advantage. For example, Envoy `/stats` endpoint can be used to enumerate all upstream services in the cluster.
2. In theory an attacker could masquerade the metrics endpoint(s) and inject fake data to monitoring systems, in order to e.g. hide an ongoing attack, confuse the system to autoscale up/down etc...

It would be nice to protect the metrics endpoints with TLS, using mutual authentication (client+server certs).

tsaarni:
**Description**

This change adds an option to protect Contour and Envoy metrics endpoint with TLS, using server and alternatively also client authentication.

Motivation: An attacker might find some sensitive information from metrics that they can use for their advantage, or attacker could masquerade the metrics endpoint and inject fake data to monitoring systems.

Fixes #3594

**Proposed design (for comments)**

New fields are added to the configuration to set certificates as local file paths, not as secret references. 

New "static secrets" concept is introduced to `SecretCache` for configuring the TLS credentials for Envoy metrics endpoint. It is intended to be similar to static listeners in `ListenerCache`. Static secret is not connected to any Kubernetes resource and as result it does not get updated (similarly as static listeners do not get updated). It can be used for file path based SDS secrets, and Envoy itself will pick up changes from filesystem (via inotify). The reason for proposing "static secrets" for this use case is to avoid changing DAG processing. Currently the Kubernetes events are closely related to DAG, which is built around the virtualhosts that user defines. As far as I can see, handling of "extra secrets", which are not related to virtualhosts, is not easy to fit into this model.

Envoy stats and readiness endpoints have been served by single listener so far. When metrics endpoint is served over HTTPS this is not the case anymore. Two separate listeners are created.

Regarding Contour metrics, Contour already had the capability to define separate metrics and readiness ports, or combining them into single port. This proposal will add new constraint: if TLS is configured for Contour metrics, two separate ports must be configured for metrics and readiness.

**Remaining tasks**

- [X] Wait for CRD PR https://github.com/projectcontour/contour/pull/4048 to be merged and align this PR with that
- [X] Add documentation


tsaarni:
I would have couple questions on my mind, where I'd appreciate some guidance:

**(1) How should user configure optional TLS support?**

Currently user configures the address:port for metrics service by `contour serve` flags `--http-address` and `--http-port`. Should certificates be configured also with command line flags or with config file?

**(2) Should /healthz endpoint be also TLS protected?**

Currently the metrics and health services are automatically combined into single server, unless user explicitly provides [different address:port ](https://github.com/projectcontour/contour/blob/1d9007fe9bddf2c7bff934dd3fe4d97fc4992d90/cmd/contour/serve.go#L520) for the two services. By default [they are the same](https://github.com/projectcontour/contour/blob/1d9007fe9bddf2c7bff934dd3fe4d97fc4992d90/cmd/contour/servecontext.go#L95).

Should it then be possible to have health service in the HTTPS server?  

While Kubernetes `livenessProbe` can be HTTPS there is no security benefit in it. What is worse, is that the probe would stop working if user configures also client certificate validation for this combined endpoint - there is no way to configure client certificate for Kubernetes.  So from that point of view, I would lean into thinking that when user wants to protect metrics endpoint with TLS, it should be configured as a separate HTTPS server: `/metrics` should not be possible to combine with `/healthz`.

Thanks in advance!

youngnick:
Hi @tsaarni, thanks for this PR! I like the direction this is heading, and thanks for asking for clarification here.

> **(1) How should user configure optional TLS support?**
> 
> Currently user configures the address:port for metrics service by `contour serve` flags `--http-address` and `--http-port`. Should certificates be configured also with command line flags or with config file?

If we're going to do TLS protection for metrics, all those TLS details should be passed in the config file, by filename. Obviously, we'd want to reuse a similar setup to what you built earlier with cert rotation (which I can see you've pulled in here).

 
> **(2) Should Contour /healthz endpoint be also TLS protected?**
>
> [cut a great explanation]

I agree with your assessment here. I don't think that protecting `/healthz` gains very much, and would add a lot of complexity.
 
> 
> **(3) Should Envoy /ready endpoint be also TLS protected?**
 
I think that, for the same reason, the `/ready` endpoint doesn't add any value to being TLS-protrected. To change this, though, would mean that specifying TLS details for stats would imply splitting these two things up (which will also require substantial changes to the shutdown-manager and the Envoy Daemonset YAMLs).

It seems like we're a bit stuck here. We've got a couple of options I can see, and they're both not great.

- We allow users to configure TLS on the `/stats` endpoint. This forces the splitting of `/stats` and `/healthz` (for Contour) or `/ready` (for Envoy), which will mean that the Daemonset YAML will need to be changed so that the `/stats` scrape config matches reality. This does have the advantage that it keeps the stats hidden, and we can add client certificates, and all that will break if there's an error there is stats. However, it has the disadvantage that the shutdown-manager (which polls the `/stats` endpoint) will need the credentials to connect, which means those credentials will need to be mounted in the Envoy pod. So the client and serving certs will have to be visible to the same Pod. That does not seem ideal.
- We allow users to configure TLS on the `/stats` endpoint, but have that also affect the `/healthz` and `/ready` endpoints. This means that either the Kubelet will need to be configured with TLS details (including a client cert if configured), or Kubernetes health checks will need to be removed. The former seems like it is not supported (kubernetes/kubernetes#64866 was closed without a fix), and the latter seems like the operability loss would outweigh the security win.

What do you think @tsaarni ? Other thoughts @projectcontour/maintainers ?

tsaarni:
Hi @youngnick 

> This forces the splitting of /stats and /healthz (for Contour) or /ready (for Envoy), which will mean that the Daemonset YAML will need to be changed so that the /stats scrape config matches reality.

This makes most sense to me right now, to force user to provide separate ports when using TLS - implying also split for Envoy `stats-health` listener into separate `stats`and `health` listeners which is not an option currently.

Regarding DaemonSet changes: would it make sense to skip the example YAML for this feature? That is, keep the current manifests and only document how to enable TLS for metrics, including explaining the YAML changes?  User would anyways need to do more work for this feature, since Prometheus must be also configured with client cert and key and CA cert.

> However, it has the disadvantage that the shutdown-manager (which polls the `/stats` endpoint) will need the credentials to connect

Isn't `/stats` endpoint available in two different ports:

1. `localhost:9001`: this is the envoy admin port
1. `0.0.0.0:8002`: this is the `stats-health` listener that routes to `service-stats` cluster which proxies for (1)

TLS could be added only to (2).  Shutdown-manager seems to use (1) and should be able to access the stats without credentials.

youngnick:
Thanks for the correction @tsaarni, you're right about the shutdown-manager.

After some time thinking about this, I think that having the behavior be "if you supply TLS for `/stats`, it has to run on a different port" is the best idea. @projectcontour/maintainers any thoughts from anyone else?

tsaarni:
Reopening this since I'd like to come back to this PR.  I think there is one more design question where I'd appreciate your guidance.  Sorry for having so many questions, for such a simple feature there is surprisingly many obstacles :astonished: 

Currently stats configuration is split in half:  during bootstrap [the stats cluster](https://github.com/projectcontour/contour/blob/f9f263e96668437a4f1c33ab7aa7d2be9aeb3d69/internal/envoy/v3/bootstrap.go#L197-L209) static resource is created.  However corresponding stats [listener](https://github.com/projectcontour/contour/blob/f9f263e96668437a4f1c33ab7aa7d2be9aeb3d69/internal/xdscache/v3/listener.go#L292) is configured at runtime as a dynamic resource via XDS.  I have not found if there is important reason for this curious split in responsibilities, but the implication is that key/certificate configuration for the listener must be also dynamically configured.  

This brings additional complication: stats listener *is not* part of normal DAG processing, but instead it is [treated separately in listener cache](https://github.com/projectcontour/contour/blob/f9f263e96668437a4f1c33ab7aa7d2be9aeb3d69/internal/xdscache/v3/listener.go#L295).  As result, I believe the listener won't be getting secret cache updates correctly **[1]**.  

I have some options:

1) Provision certificates/keys dynamically via XDS as well and hook stats listener to DAG rebuild somehow
2) Refer to static file based SDS resources from dynamic stats listener and modify the bootstrap to generate the static file based SDS resources

As third option, I guess Envoy stats listener could have been changed to static resource, removing the split configuration we now have, but that would mean breaking command line parameter compatibility: `contour serve --stats-address=<ipaddr> --stats-port=<port>` parameters would need to be removed and reintroduced in `contour bootstrap` instead.


**[1]** Contour implements SDS partially, only for TLS credentials only (keys/certificates).  CA certificates are inlined as byte buffer in listener config (see associated [TODO](https://github.com/projectcontour/contour/blob/f9f263e96668437a4f1c33ab7aa7d2be9aeb3d69/internal/envoy/v3/auth.go#L76)).  According to my understanding, it means that all listeners must be re-configured when CA secret is updated, to update the inlined certificates.

sunjayBhatia:
I *think* setting up the listener dynamically is maybe not documented as such but ensures the Envoy pods become healthy only once they receive a config update from Contour, since the default deployment uses the stats endpoint in the container healthcheck: https://github.com/projectcontour/contour/blob/e55cd3cd9ae78b1d3fd71cb8a17789f085c05a73/examples/contour/03-envoy.yaml#L79-L84

tsaarni:
That would make a lot of sense, thanks @sunjayBhatia for clarifying that! 

I've started to wonder if as 3rd option, I should try to add SDS support for validation context / CA certificates. I think that could solve my problem with getting secrets updating even while stats listener remains outside of DAG rebuild.

tsaarni:
I've been experimenting a little to get better understanding how this could be added e.g. regarding secrets provisioning for Envoy metrics endpoint. This appeared to be bit more complicated than I first thought and this could have been a topic for a separate design doc, but on the other hand, I would not have been able to propose that much without experimenting first.  So instead, I now wrote a proposal in the top of this PR. See *Proposed design (for comments)* above. 

@projectcontour/maintainers I would appreciate if you can take a look at that and give any comments.

tsaarni:
@projectcontour/maintainers Reviews appreciated!  Please, check also the short introduction chapter in the description of this PR.

internal/httpsvc/http.go
```diff
@@ -45,12 +54,29 @@ func (svc *Service) Start(stop <-chan struct{}) (err error) {
...
+		tlsConfig, err = svc.tlsConfig()
+		if err != nil {
+			return err
+		}
+	}
+
+	// If one of the TLS parameters are defined, at least server certificate
+	// and key must be defined.
+	if (svc.Cert != "" || svc.Key != "" || svc.CABundle != "") &&
+		!(svc.Cert != "" && svc.Key != "") {
```
sunjayBhatia, line 69: could simplify
```suggestion
		(svc.Cert == "" || svc.Key == "") {
```

sunjayBhatia:
from @youngnick we had this as well:

> After some time thinking about this, I think that having the behavior be "if you supply TLS for `/stats`, it has to run on a different port" is the best idea. @projectcontour/maintainers any thoughts from anyone else?

which ties in with what you mention in the proposal above about ports etc.

I don't have a super strong opinion but I've always thought using the same port for disparate things that could be different listeners was a bit confusing, wouldn't be opposed to making them separate when you supply TLS configuration

other than that, I think this might just need an end-to-end test to show how it all hooks up, and would be a good opportunity to start enabling some more testing of these sorts of admin/operational features

tsaarni:
>>After some time thinking about this, I think that having the behavior be "if you supply TLS for /stats, it has to run on a different port" is the best idea. @projectcontour/maintainers any thoughts from anyone else?
>
>which ties in with what you mention in the proposal above about ports etc.
> 
>I don't have a super strong opinion but I've always thought using the same port for disparate things that could be different >listeners was a bit confusing, wouldn't be opposed to making them separate when you supply TLS configuration

Ready and metrics listener is now split into two listeners on Envoy side, i.e. separate port for HTTP vs HTTPS.


> other than that, I think this might just need an end-to-end test to show how it all hooks up, and would be a good opportunity to start enabling some more testing of these sorts of admin/operational features

I added a test case.  I had some difficulties in getting port-forward to work for some reason. I left a big comment in test code for that.  It seemed to me that I needed to bend e2e framework a bit, to adapt it into new listeners.  I did not want to touch too much though, since I don't really know the backing reason why the http helpers had fixed URLs matching the existing listeners and therefore making hard to re-use for new ones.

sunjayBhatia (approved):

youngnick (approved):

PR #3711, opened 2021-05-25
fixup generate task after move to hugo

stevesloka:


Makefile
```diff
@@ -198,7 +198,7 @@ lint-flags:
 
 .PHONY: generate
 generate: ## Re-generate generated code and documentation
-generate: generate-rbac generate-crd-deepcopy generate-crd-yaml generate-deployment generate-api-docs generate-metrics-docs generate-uml
```
youngnick, line 201: After this change, how do we update the metrics docs? They're still present on the "Prometheus metrics" page, but it seems like now we won't be able to manage them programmatically? If the plan is to fix them up later, that's fine, but a TODO here or an issue link would be great.

youngnick (commented):
LGTM in general, but would like to know what the plan is for metrics documentation.

sunjayBhatia:
Just submitted a couple PRs to fix other targets/scripts that are part of the `codegen` CI job:
- https://github.com/projectcontour/contour/pull/3716
- https://github.com/projectcontour/contour/pull/3715

site/_data/refdocs/config.json
```diff
@@ -0,0 +1,28 @@
+{
```
sunjayBhatia, line 1: should this content be added? the hugo site format doesnt have a `_data` dir so this seems entirely new

Makefile
stevesloka: Why do we need to dynamically generate that page? It rarely changes and is just overhead to me. Additionally, the bits to generate the docs needed to be added by hand.

site/_data/refdocs/config.json
stevesloka: It's used for the generator to build the site. The _data dir doesn't get published over to the hugo site from my tests.

sunjayBhatia: from the generate-api-docs.sh script looks like these templates shouldnt have been deleted, maybe we can modify that script to look for them somewhere else to keep with the hugo structure?

Makefile
sunjayBhatia: removing it also means we don't run promlint on our metrics, which could mean we add something not ideal as well

stevesloka: https://github.com/projectcontour/contour/issues/3717

sunjayBhatia (approved):

danehans (approved):

PR #3721, opened 2021-05-26
site: fix links to point to version-specific content

skriss:
Updates the "main" docs links to point to the "main" GitHub
branch instead of the latest release version. Since the "main"
docs are cloned for each new versioned set of docs, this enables
the versioned docs to point to the appropriate GitHub location.

### DIFF
--- a/site/content/docs/main/_index.md
+++ b/site/content/docs/main/_index.md
@@ -1,7 +1,6 @@
 ---
 cascade:
   layout: docs
-  gh: https://github.com/projectcontour/contour/tree/release-1.15.1
   version: main
 ---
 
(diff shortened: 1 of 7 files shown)


site/content/docs/main/_index.md
```diff
@@ -1,7 +1,6 @@
 ---
 cascade:
   layout: docs
-  gh: https://github.com/projectcontour/contour/tree/release-1.15.1
```
skriss, line 4: AFAICT this isn't used anywhere (this isn't actually a valid link anyway), so I removed it so it doesn't have to be updated as part of `prepare-release.go`. LMK if you're aware of a need for this!

sunjayBhatia (approved):
makes sense to me!

stevesloka (approved):

PR #3723, opened 2021-05-26
hack/release: update prepare-release.go for Hugo

### LINKED ISSUE #3720: site: update hack/release/prepare-release.go for move to Hugo
`hack/release/prepare-release.go`, which generates a new set of versioned docs, needs to be updated to handle the new Hugo site structure.

skriss:
Closes #3720.

### DIFF
--- a/hack/release/prepare-release.go
+++ b/hack/release/prepare-release.go
@@ -95,28 +95,21 @@ func updateConfigForSite(filePath string, vers string) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


skriss:
Still doing some testing but this is looking pretty good (also requires #3721 to fully work).

sunjayBhatia (approved):
just tried it out, as far as I can tell looks right!

stevesloka (approved):

PR #3724, opened 2021-05-26
internal/dag: Filter route Kind by listener

### LINKED ISSUE #3529: Gateway API: Listener Route binding selector kind must match Routes compatible with Listener protocol
The `kind` field of the route binding selector has the following description:

```
Kind is the kind of the route resource to select.

Kind MUST correspond to kinds of routes that are compatible with the application protocol specified in the Listener’s Protocol field.

If an implementation does not support or recognize this resource type, it SHOULD set the “ResolvedRefs” condition to false for this listener with the “InvalidRoutesRef” reason.
```

The `protocol` field is defined as:

```
Protocol specifies the network protocol this listener expects to receive. The GatewayClass MUST apply the Hostname match appropriately for each protocol:

- For the “TLS” protocol, the Hostname match MUST be applied to the SNI server name offered by the client.
- For the “HTTP” protocol, the Hostname match MUST be applied to the host portion of the effective request URI or the :authority pseudo-header
- For the “HTTPS” protocol, the Hostname match MUST be applied at both the TLS and HTTP protocol layers.
```

- [x] Process Kind
- [ ] Set status on Gateway

See: https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.Listener

stevesloka:
Filters the route kind by listener so that only the specified route type
will be processed by a listener.

Updates #3529

### DIFF
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -143,6 +143,35 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 		},
 	}
 
+	gatewayWithTLSRouteSelector := &gatewayapi_v1alpha1.Gateway{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "contour",
+			Namespace: "projectcontour",
+		},
+		Spec: gatewayapi_v1alpha1.GatewaySpec{
+			Listeners: []gatewayapi_v1alpha1.Listener{{
+				Port:     80,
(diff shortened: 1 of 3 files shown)


danehans (approved):

sunjayBhatia (approved):
LGTM, this change means we are properly only looking at TLSRoutes or HTTPRoutes for each Gateway Listener's respective kind

PR #3727, opened 2021-05-26
Speed up CodeQL job

sunjayBhatia:
- Use go action cache
- Select only Go as language
- Remove HEAD^2 checkout, prevents warning in codeql init step

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -17,38 +17,28 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@v2
-      with:
-        # We must fetch at least the immediate parents so that if this is
-        # a pull request then we can checkout the head.
-        fetch-depth: 2
 
-    # If this run was triggered by a pull request event, then checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -22,33 +22,27 @@ jobs:
...
+          ${{ runner.os }}-${{ github.job }}-go-
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
       uses: github/codeql-action/init@v1
       # Override language selection by uncommenting this and choosing your languages
-      # with:
-      #   languages: go, javascript, csharp, python, cpp, java
+      with:
+        languages: go
```
sunjayBhatia, line 41: I'm assuming we only care about this, previously it was also scanning javascript from our site, not sure if we want to add that back

skriss (approved):
LGTM

.github/workflows/codeql-analysis.yml
youngnick: I think this is fine for now.

youngnick (approved):
LGTM, nice work.

PR #3728, opened 2021-05-26
move cert-contour.ext, cert-envoy.ext to certs/

skriss:
Moves the two .ext files for cert generation from _integration/
to certs/ since they're not used in _integration and that dir will
be deleted soon.

Updates #3621.


sunjayBhatia (approved):

stevesloka (approved):
Do we even need this anymore? I've never used it, to be honest.

skriss:
Yeah, not sure - there are docs and make targets for generating certs manually that use these files, but I'm not sure if folks are using them. cc @youngnick for thoughts.

skriss:
Per maintainers discussion we can remove all of the make targets/etc for manual cert generation as they're no longer relevant. I'll update this PR to do that.

youngnick:
Thanks @skriss, there's no reason for the external cert generation stuff to be in here any more.

skriss:
Looked at this some more and the make targets are/were used for generating test data for some unit tests, so maybe worth keeping them around. I'm going to leave that as-is for now and merge this; I'll log a follow-up issue to look at if we want to remove everything.

PR #3732, opened 2021-05-26
add gateway.controllername to config docs

stevesloka:
Followup to #3659 adds missing config page info for new controllerName config option. 

### DIFF
--- a/site/content/docs/main/configuration.md
+++ b/site/content/docs/main/configuration.md
@@ -180,6 +180,7 @@ The gateway configuration block is used to configure which gateway-api Gateway C
 
 | Field Name | Type| Default  | Description |
 |------------|-----|----------|-------------|
+| controllerName | string |  | Gateway Class controller name (i.e. projectcontour.io/projectcontour/contour).  |
 | name | string | contour | This field specifies the name of a Gateway.  |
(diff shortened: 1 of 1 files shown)


site/content/docs/main/configuration.md
```diff
@@ -180,6 +180,7 @@ The gateway configuration block is used to configure which gateway-api Gateway C
 
 | Field Name | Type| Default  | Description |
 |------------|-----|----------|-------------|
+| controllerName | string |  | Gateway Class controller name (i.e. projectcontour.io/projectcontour/contour).  |
```
sunjayBhatia, line 183:
```suggestion
| controllerName | string | `projectcontour.io/<contour namespace>/contour` | Specifies the GatewayClass controller name to reconcile. |
```

maybe add this as the default? If i remember right that's what the default ends up being, maybe good enough to make the namespace `projectcontour`

stevesloka: There isn't a default anymore. The controllerName was made a pointer right now to allow some time to not break any existing users.

sunjayBhatia: ah ok, disregard then

sunjayBhatia (approved):

PR #3735, opened 2021-05-27
internal/dag: Implement GatewayAllow for TLSRoutes

### LINKED ISSUE #3695: Implement GatewayAllow in TLSRoute
Is [this block of logic](https://github.com/projectcontour/contour/blob/e3cf805af778fe42ec39681dbf71b783b091007e/internal/dag/gatewayapi_processor.go#L138-L157) that we have for `HTTPRoutes` not needed for `TLSRoutes`?

_Originally posted by @skriss in https://github.com/projectcontour/contour/pull/3627#discussion_r634681222_

stevesloka:
Fixes #3695

### DIFF
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -172,6 +172,25 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 		},
 	}
 
+	gatewayTLSRouteNoSelector := &gatewayapi_v1alpha1.Gateway{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "contour",
+			Namespace: "projectcontour",
+		},
+		Spec: gatewayapi_v1alpha1.GatewaySpec{
+			Listeners: []gatewayapi_v1alpha1.Listener{{
+				Port:     80,
+				Protocol: gatewayapi_v1alpha1.HTTPProtocolType,
+				Routes: gatewayapi_v1alpha1.RouteBindingSelector{
+					Kind: KindTLSRoute,
+					Namespaces: &gatewayapi_v1alpha1.RouteNamespaces{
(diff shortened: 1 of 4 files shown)


stevesloka:
I'm going to follow up with some integration tests which would be nice to route through the API server properly.

internal/dag/builder_test.go
```diff
@@ -707,7 +707,7 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 					},
 					Spec: gatewayapi_v1alpha1.HTTPRouteSpec{
 						Gateways: &gatewayapi_v1alpha1.RouteGateways{
-							Allow: gatewayAllowTypePtr(gatewayapi_v1alpha1.GatewayAllowAll),
+							Allow: gatewayAllowTypePtr(gatewayapi_v1alpha1.GatewayAllowSameNamespace),
```
stevesloka, line 710: This was wrong previously and should have been "SameNamespace". The result it still correct, but just fixing up in this PR.

stevesloka: Also I think we could DRY this tests up, there's a lot of duplication. 😢

stevesloka:
My logic is broken somewhere causing the TLSRoute to be kicked out. Digging into that now.

skriss (approved):
LGTM

stevesloka:
I didn't update the _integration tests, probably need fixes. Do we care at this point since we're trashing those?

sunjayBhatia:
> I didn't update the _integration tests, probably need fixes. Do we care at this point since we're trashing those?

idk if the _integration removal will make it in before 1.16 so just to keep the pipeline green, maybe fix or remove the test assuming it is mirrored into e2e

sunjayBhatia (commented):
LGTM pending CI (removal of _integration test or adding missing allow all selector)

sunjayBhatia (approved):

Issue #3743, opened 2021-05-28
test/e2e/upgrade: Upgrade tests should work on release branches, tags, and main


sunjayBhatia:
Currently the upgrade tests fail to deploy the correct *from* version on release branches, see: https://github.com/projectcontour/contour/runs/2688468015?check_suite_focus=true

This is because we try to figure out the previous version to upgrade from here: https://github.com/projectcontour/contour/blob/9f665c80e86d46efb61494d0aec3ec4bdc1a4eb8/Makefile#L13

On a release branch, this will not be correct, as it finds the most recent tag globally (e.g. on the 1.16 release it tries to upgrade 1.16 to 1.16 which is a no-op, it should be upgrading from 1.15.x)

Things to fix here:
- whatever branch we are on, we need to upgrade from the most recent release for that branch
  - so that main and non-tagged commits on release branches can properly run the upgrade tests
  - complication on main is that there are no tags on the branch itself, vs. release branches which always have a tag
- if we are on a tag commit, upgrade from the tag that immediately precedes it

PR #3744, opened 2021-05-28
upgrade tests: Correct "from" version for different git refs

sunjayBhatia:
- On a tag, use previous tag
- On a release branch, find previous tag on that branch
- On main or other refs, just use globally most recent tag

Also moves another script form hack into the test/scripts dir

Output from various branches:

branch main: `v1.16.0`

tag v1.16.0: `v1.15.1`

branch release-1.16 with an empty commit on top: `v1.16.0`

branch release-1.16 with no commit: `v1.15.1`

*Would be good to cherry-pick this back to the release-1.16 branch to keep CI there green*

Fixes: #3743


skriss (approved):
LGTM

stevesloka (approved):

sunjayBhatia:
so just realizing, my only fear with this is the CI run on a PR to the release branch might fail 🤔

sunjayBhatia:
see https://github.com/projectcontour/contour/pull/3747

sunjayBhatia:
hmm, also within a release branch, this will only upgrade from the previous patch release, which might be ok but maybe we should test upgrading from the previous minor release?

sunjayBhatia:
going to merge, our upgrade matrix will be something to continue to think about

PR #3748, opened 2021-05-28
handle tls route gateway.listener modes

stevesloka:
internal/dag: handle tls route gateway.listener modes
    
 Fixup issue where the listener protocol wasn't checked for type=tls.
 Also, adds logic to validate the TLS.Mode is valid.
    

### DIFF
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -78,7 +78,7 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
 		var matchingTLSRoutes []*gatewayapi_v1alpha1.TLSRoute
 		var listenerSecret *Secret
 
-		// Validate the Kind on the selector is a supported type.
+		// Validate the Protocol on the selector is a supported type.
 		switch listener.Protocol {
 		case gatewayapi_v1alpha1.HTTPSProtocolType:
 			// Validate that if protocol is type HTTPS, that TLS is defined.
@@ -93,7 +93,27 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
 				// routes to be bound to this listener since it can't serve TLS traffic.
 				continue
 			}
-		case gatewayapi_v1alpha1.HTTPProtocolType, gatewayapi_v1alpha1.TLSProtocolType:
+		case gatewayapi_v1alpha1.TLSProtocolType:
+
+			if tlsConfig := listener.TLS; tlsConfig != nil {
+				if tlsConfig.Mode != nil {
+					switch *tlsConfig.Mode {
+					case gatewayapi_v1alpha1.TLSModeTerminate:
+						// Check for TLS on the Gateway.
+						if listenerSecret = p.validGatewayTLS(listener); listenerSecret == nil {
+							// If TLS was configured on the Listener, but it's invalid, don't allow any
+							// routes to be bound to this listener since it can't serve TLS traffic.
+							continue
+						}
+					case gatewayapi_v1alpha1.TLSModePassthrough:
+						if listener.TLS.CertificateRef != nil {
+							p.Errorf("Listener.TLS cannot be defined when TLS Mode is %q.", tlsConfig.Mode)
+							continue
+						}
+					}
+				}
+			}
+		case gatewayapi_v1alpha1.HTTPProtocolType:
 			break
 		default:
 			p.Errorf("Listener.Protocol %q is not supported.", listener.Protocol)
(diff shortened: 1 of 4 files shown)


test/e2e/gateway/gateway_test.go
```diff
@@ -212,7 +212,50 @@ var _ = Describe("Gateway API", func() {
 		})
 
 		It("008-tlsroute", func() {
-			testTLSRoute(f)
+			testTLSRoute(f, "gateway-008-tlsroute")
+		})
+	})
+
+	Describe("TLSRoute Gateway: Mode:Terminate", func() {
```
sunjayBhatia, line 219: should this be passthrough? given https://github.com/projectcontour/contour/pull/3748/files#diff-d6a872e95d486ae59e2bb8b2aaaecf28211105b1c0ec03d65ce18da7a7837290R237

test/e2e/gateway/008_tlsroute_test.go
```diff
@@ -27,16 +27,16 @@ import (
 	gatewayv1alpha1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
 )
 
-func testTLSRoute(fx *e2e.Framework) {
+func testTLSRoute(fx *e2e.Framework, namespace string) {
```
sunjayBhatia, line 30: we could make this `testTLSRoutePassthrough` and have another that is `testTLSRouteTerminate` for the other case maybe? run a plain HTTP server behind contour, connect on HTTPS and our request should go to that service

test/e2e/gateway/gateway_test.go
stevesloka: yup!

test/e2e/gateway/008_tlsroute_test.go
stevesloka: Yeah we can do that, was just trying to minimize the number of duplicate resources, but I can pull out the TLS route and reference from both since these tests are really testing the gateway configuration, not the TLSRoute configuration, but that will need to change as we add more.

stevesloka: I ended up taking out the terminate test. We don't support it right now and also don't set status. 

This is happening because when the TLSRoute is selected by an invalid Gateway, the route is not processed (since the Gateway is bad), but need to set an "Admitted: false" for the selected routes probably, but not yet implemented.

internal/dag/gatewayapi_processor.go
```diff
@@ -168,6 +188,12 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
 				// with the Gateway. If this Selector is defined, only routes matching the Selector
 				// are associated with the Gateway. An empty Selector matches all routes.
 
+				// Validate the listener protocol is type=TLS.
+				if listener.Protocol != gatewayapi_v1alpha1.TLSProtocolType {
+					p.Errorf("invalid listener protocol %q for Kind: TLSRoute", listener.Protocol)
+					continue
+				}
+
```
skriss, line 196: I'd check this before entering the loop over `p.source.tlsroutes` since it doesn't require any info from each individual route and doing it once per route will result in repeated errors.

skriss (changes requested):

test/e2e/gateway/008_tlsroute_test.go
sunjayBhatia: > I ended up taking out the terminate test. We don't support it right now and also don't set status.

Could you clarify this? Looks like we're implementing some of it here: https://github.com/projectcontour/contour/blob/d5d9e55050d85c0793ff6daf8e3726bd7c3ba8bd/internal/dag/gatewayapi_processor.go#L101

are we just doing validation for now and still doing passthrough for all TLSRoutes?

sunjayBhatia: i see we're still not attaching the TLS secret to the dag secure vhost etc. https://github.com/projectcontour/contour/blob/d5d9e55050d85c0793ff6daf8e3726bd7c3ba8bd/internal/dag/gatewayapi_processor.go#L447

sunjayBhatia: should we explicitly say Terminate mode is not allowed for now and enable it in the PR that provides support for that?

stevesloka: Yeah this doesn't add Terminate mode, just some more plumbing to make it useful. It was my understanding originally chatting with @youngnick that TLSRoute **only** did passthrough, but that's not the case, so this is gradually adding more to enable Terminate mode in the next PR.

internal/dag/gatewayapi_processor.go
stevesloka: 👍

sunjayBhatia (approved):
looks good to me, modulo Steve K's comments of course

internal/dag/gatewayapi_processor.go
```diff
@@ -168,6 +188,12 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
 				// with the Gateway. If this Selector is defined, only routes matching the Selector
 				// are associated with the Gateway. An empty Selector matches all routes.
 
+				// Validate the listener protocol is type=TLS.
+				if listener.Protocol != gatewayapi_v1alpha1.TLSProtocolType {
```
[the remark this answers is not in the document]
stevesloka: yup added! =)

PR #3750, opened 2021-06-01
Refactor Makefile and update local dev options

youngnick:
This is a WIP PR to update the local dev options. I'm aiming to have
a make target that will:
- spin you up a kind cluster
- set you up with Envoys running in there, trying to connect to a Contour
  running on the host.
- with the SSL pulled out for you.

First up though, is a Makefile refactor to make it a bit more composable.

This is a WIP PR because it's quicker to use Actions for testing than my
local machine. :)

This commit splits up many of the make targets into sub-steps,
that we can reuse, instead of having the same script called in
multiple places.

I've also removed the old local-dev shortcuts, I think we can do
better.

The intent here is to make it so that we can use the pieces
more easily, eventually aiming at making local dev and bug
reproduction quicker and easier.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -176,18 +176,18 @@ lint-codespell: CODESPELL_SKIP := $(shell cat .codespell.skip | tr \\n ',')
 lint-codespell:
 	@./hack/codespell.sh --skip $(CODESPELL_SKIP) --ignore-words .codespell.ignorewords --check-filenames --check-hidden -q2
 
-.PHONY: check-golint
+.PHONY: lint-golint
 lint-golint:
 	@echo Running Go linter ...
 	@./hack/golangci-lint run --build-tags=e2e
 
-.PHONY: check-yamllint
+.PHONY: lint-yamllint
 lint-yamllint:
 	@echo Running YAML linter ...
 	@./hack/yamllint examples/ site/content/examples/
 
 # Check that CLI flags are formatted consistently. We are checking
-# for calls to Kingping Flags() and Command() APIs where the 2nd
+# for calls to Kingpin Flags() and Command() APIs where the 2nd
 # argument (the help text) either doesn't start with a capital letter
 # or doesn't end with a period. "xDS" and "gRPC" are exceptions to
 # the first rule.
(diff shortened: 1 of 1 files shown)


sunjayBhatia (commented):
the Makefile refactor looks good to me, pulled the branch down and messed with it 👍🏽 

are you intending to add the local dev commands to this PR as well?

Makefile
```diff
@@ -331,28 +313,51 @@ site-devel: ## Launch the website
 site-check: ## Test the site's links
 	# TODO: Clean up to use htmltest
 
-.PHONY: e2e
-e2e: ## Run E2E tests against a real k8s cluster
+
+# Tools for testing and troubleshooting
+
+.PHONY: make-kind-cluster
+make-kind-cluster: ## Make a kind cluster with standard ports forwarded
```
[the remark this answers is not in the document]
youngnick: haha, that is a great suggestion. I'll update.

youngnick:
Yes, I'm working on the local dev commands here too, but maybe I should just push this in first. I'll tidy it up and do that, make the PRs easier to review.

youngnick:
This should be good for actual review now. I'll build the local-dev stuff from here.

sunjayBhatia (approved):

Makefile
```diff
@@ -338,28 +320,51 @@ site-devel: ## Launch the website
...
+.PHONY: cleanup-kind
+cleanup-kind:
 	./test/scripts/cleanup.sh
 
-.PHONY: upgrade
-upgrade: ## Run upgrade tests against a real k8s cluster
-	./test/scripts/make-kind-cluster.sh
-	./test/scripts/install-contour-release.sh $(CONTOUR_UPGRADE_FROM_VERSION)
+.PHONY: load-contour-image-kind
+load-contour-image-kind:
```
stevesloka, line 350: Can probably remove this since it's required to have the image exported first to disk at a specific place.

stevesloka: Or if you want to keep, add a comment stating what needs set to properly use?

stevesloka (changes requested):

Makefile
```diff
@@ -46,6 +46,8 @@ BUILD_GOPROXY ?= https://proxy.golang.org
 # LOCALIP as an env var before running 'make local' will solve that.
 LOCALIP ?= $(shell ifconfig | grep inet | grep -v '::' | grep -v 127.0.0.1 | head -n1 | awk '{print $$2}')
 
+DOCKERIP ?= $(shell docker network inspect kind | jq -r ".[0].IPAM.Config[0].Gateway")
```
stevesloka, line 49: This isn't used anywhere in this changeset?

stevesloka:
Added label `release-note` due to the deprecation comment.

> This method of certificate generation is DEPRECATED and will be removed soon.

Makefile
youngnick: ah, thanks, I'll pull it out. It will be in the next one instead.

Makefile
youngnick: I was just moving things around, but I'll have a look and see if we can use Make properly here, and have a rule that will rebuild the contour image if stuff has changed, which can then be a prereq for this rule.

youngnick: Agreed that if I keep, I'll add a comment which tells you what this step expects.

skriss (commented):
LGTM pending open comments

Makefile
youngnick: I've updated with an explanatory comment and a `make help` line.

youngnick:
Okay, actually resolved the pending comments this time, thanks everyone. Should be good for final review now.

skriss (approved):
LGTM

Issue #3749, opened 2021-05-29
The Contour v1.16.0 requires Kubernetes >= v1.19


takama:
**What steps did you take and what happened:**
Install the Contor v1.16.0 via helm charts

contour-7c9f75d965-d27tk   0/1     CrashLoopBackOff   1          20s

time="2021-05-28T20:23:23Z" level=**fatal** msg="failed to create informer" error="no matches for networking.k8s.io/v1, Resource=ingresses" resource="networking.k8s.io/v1, Resource=ingresses"

**Anything else you would like to add:**
The ingress `apiVersion: networking.k8s.io/v1`started from the version of Kubernetes [v1.19 ](https://kubernetes.io/docs/concepts/services-networking/ingress/#the-ingress-resource)
To be compatible with Kubernetes <= v1.18 the ingress API version should be `networking.k8s.io/v1beta1`.

**What did you expect to happen:**
Expectedly, it should work with the last stable GKE v1.18 if it was not declared incompatible with Kubernetes <= v1.18

**Environment:**

- Contour version: v1.16.0
- Kubernetes version: (use `kubectl version`): v1.18.17-gke.700
- Kubernetes installer & version: GKE 1.18.17-gke.700
- Cloud provider or hardware configuration: GKE
- OS (e.g. from `/etc/os-release`): COS containerd

youngnick:
Thanks for logging this issue @takama. In general, Contour's Kubernetes version support tracks the supported Kubernetes versions. As you can see from our (compatibility matrix){https://projectcontour.io/resources/compatibility-matrix/], we haven't officially supported Kubernetes 1.18 since Contour 1.14.

However, I one hundred percent acknowledge that we have done a bad job of publicizing both this fact, and the Ingress v1beta1 removal. I will be adding some bits to our standard changelog template:
- We will call out the currently supported Kubernetes versions, and note if that changes.
- We will call out object deprecations like this as a first-level changelog item.

Thanks as well for the note about GKE stable only being 1.18, but since that's out of upstream support, it's out of our support too.

takama:
Thank you @youngnick  for your response
> As you can see from our (compatibility matrix){https://projectcontour.io/resources/compatibility-matrix/], we haven't officially supported Kubernetes 1.18 since Contour 1.14.

Ok, then, please change the message on the page above and [here](https://github.com/projectcontour/contour) that the minimum compatible Kubernetes version now is v1.19. (not v 1.16)
[...]

PR #3751, opened 2021-06-01
Add k8s min version clarification

sunjayBhatia:
- We dropped support for k8s 1.18
- Contour 1.16.0 stopped falling back to Ingress v1beta1 resources

Updates: #3749

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Its goal is to expand upon the functionality of the Ingress API to allow for a r
 
 ## Prerequisites
 
-Contour requires Kubernetes version 1.16 or later, for [v1 Custom Resource Definition support](https://kubernetes.io/blog/2019/09/18/kubernetes-1-16-release-announcement/#custom-resources-reach-general-availability).
+Contour requires Kubernetes version 1.19 or later, for [stable Ingress v1 support](https://kubernetes.io/docs/concepts/services-networking/ingress/).
 
 RBAC must be enabled on your cluster.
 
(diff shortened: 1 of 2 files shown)


stevesloka (commented):
I still think we should revert the changes and allow 1.16+ clusters to function even if it's not "supported + tested".

We have lots of notes all over that page saying that other combinations might work since the API integration surface area is so small. If this is going to become a hard min, then those bits of docs should also be reworded. 

To me, we should stick with what's "supported", but not block users if their environment is still behind (e.g. GKE being 1.18 for stable version). I thought we'd have a quick maintainer discussion on this, but that hasn't happened. =)

skriss (approved):
LGTM

sunjayBhatia:
> I still think we should revert the changes and allow 1.16+ clusters to function even if it's not "supported + tested".
> 
> We have lots of notes all over that page saying that other combinations might work since the API integration surface area is so small. If this is going to become a hard min, then those bits of docs should also be reworded.
> 
> To me, we should stick with what's "supported", but not block users if their environment is still behind (e.g. GKE being 1.18 for stable version). I thought we'd have a quick maintainer discussion on this, but that hasn't happened. =)

I feel a bit weird saying things "might work" and not testing that they do or don't so I would be down to remove that text, especially since what is there now does have a hard minimum of 1.16 already

but yeah we should have a chat about this with everyone

stevesloka:
> but yeah we should have a chat about this with everyone

I think it's just me. So I'll bow out and go with ya'll.

stevesloka (approved):

youngnick (approved):

skriss:
@sunjayBhatia looks like still some test flakiness on os x, I think we can just merge this one now and track that down separately.

skriss:
#3763 for the flake

PR #3753, opened 2021-06-01
Contour binary version and image version labels can be overridden

sunjayBhatia:
Output of "contour version" can be controlled separately from image tag
if needed.


sunjayBhatia:
Maybe we need to do some cleanup on the variables etc. we use in the makefile to namespace them more/make them more intuitive? I just added onto the existing patterns, but could see this be improved

youngnick (approved):
LGTM, I think that namespacing these is a good idea, let's do that after we finish cleaning up a bit (I think there is more to remove, like the certificate generation stuff).

skriss (approved):

Issue #3712, opened 2021-05-25
Docs: Cleanup after Hugo migration


jonasrosland:
There are a few instances within the docs which still have artifacts left over from the Jekyll to Hugo migration. These instances have all been fixed in v1.15.1 and main, but still remain in previous versions.

This will be a tracking issue for the fixes, so if you want to just fix part of the list please do so and reference this issue.

The list below shows how the artifacts impact the v1.15.0 docs, and if you fix them please also fix the previous versions impacted so we can mark them as complete.

- [x] https://projectcontour.io/docs/v1.15.0/architecture/ has a .center-image tag (twice) (#3755)
- [x] https://projectcontour.io/docs/v1.15.0/config/fundamentals has tag {% highlight yaml linenos %} | {% endhighlight %} as well as md link at the bottom: [4]: {% link docs/{{page.version}}/config/api.md %} (#3756)
- [x] https://projectcontour.io/docs/v1.15.0/config/virtual-hosts/ has link examples directory which 404s (#3758, #3757)
- [x] https://projectcontour.io/docs/v1.15.0/config/inclusion-delegation/ has link conditions which 404s (#3765)
- [x] https://projectcontour.io/docs/v1.15.0/config/tls-termination/ link to Contour configuration file 404s
- [x] https://projectcontour.io/docs/v1.15.0/config/upstream-tls/ has markdown links at the bottom (#3767)
- [x] https://projectcontour.io/docs/v1.15.0/config/request-routing/ under Header Conditions section, missing bullet arrow for notpresent, and on the same page: retryPolicy.count & retryPolicy.perTryTimeout has extra indentation
- [x] https://projectcontour.io/docs/v1.15.0/config/client-authorization/ under section Scoping Authorization Policy Settings link to authorization policy 404s, and same page: extra md link at bottom (#3768)
- [x] https://projectcontour.io/docs/v1.15.0/config/annotations/ under first section has missing link [HTTPProxy API][15], and markdown links at the bottom (#3777)
- [x] https://projectcontour.io/docs/v1.15.0/deploy-options/ example manifest link 404s, and a bunch of broken links and markdown links at the bottom (#3778)
- [x] https://projectcontour.io/docs/v1.15.0/configuration/#configuration-file tables has {: class="table thead-dark table-bordered"} inline, and an extraneous `<br>`, and on the same page link to `example manifests` 404s (#3780)
[...]

PR #3755, opened 2021-06-01
site: Fix architecture page with extra center tags after upgrade to Hugo

stevesloka:
Updates #3712


sunjayBhatia (approved):
preview on the site looks good 👍🏽

PR #3766, opened 2021-06-02
site: Fix links in tls-termination docs for change to Hugo

stevesloka:
Updates #3712

### DIFF
--- a/site/content/docs/main/config/tls-termination.md
+++ b/site/content/docs/main/config/tls-termination.md
@@ -246,5 +246,5 @@ spec:
       weight: 20
 ```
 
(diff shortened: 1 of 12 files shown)


sunjayBhatia (commented):
link change looks good, but maybe we should link this to the config file reference rather than an example?

stevesloka:
@sunjayBhatia I updated the link to point to the configuration docs page for the fallback certificate.

sunjayBhatia (approved):

skriss (changes requested):
should make the same update for the `main` version of the page so it sticks going fwd

stevesloka:
+1 @skriss

skriss (approved):
:+1: LGTM!

Issue #3705, opened 2021-05-24
Add Controller Runtime Reconcile Tests


danehans:
**Please describe the problem you have**
https://github.com/projectcontour/contour/pull/3510 added controller-runtime controllers but no tests for these controllers. Tests should be added to test controller functionality.

PR #3773, opened 2021-06-04
Adds Controller Runtime Test Suite

danehans:
Adds a test suite to test controller-runtime controllers, starting with the gatewayclass controller.

Fixes https://github.com/projectcontour/contour/issues/3705

### DIFF
--- a/internal/controller/gatewayclass.go
+++ b/internal/controller/gatewayclass.go
@@ -116,7 +116,9 @@ func (r *gatewayClassReconciler) Reconcile(ctx context.Context, request reconcil
 	}
 
 	// Pass the new changed object off to the eventHandler.
-	r.eventHandler.OnAdd(gc)
+	if r.eventHandler != nil {
+		r.eventHandler.OnAdd(gc)
+	}
 
 	if err := status.SyncGatewayClass(ctx, r.client, gc, errs); err != nil {
 		return reconcile.Result{}, fmt.Errorf("failed to sync gatewayclass %q status: %w", gc.Name, err)
--- a/internal/controller/gatewayclass_test.go
+++ b/internal/controller/gatewayclass_test.go
@@ -87,12 +87,12 @@ var _ = Describe("GatewayClass Controller", func() {
 			// Create
 			Expect(cl.Create(context.Background(), waiting)).Should(Succeed())
 
-			By("Expecting waiting reason")
+			By("Expecting not admitted status condition")
 			Eventually(func() bool {
 				gc := &gatewayv1alpha1.GatewayClass{}
 				_ = cl.Get(context.Background(), key, gc)
-				return isWaiting(gc)
-			}, timeout, interval).Should(BeTrue())
+				return isAdmitted(gc)
+			}, timeout, interval).Should(BeFalse())
 
 			// Delete
 			By("Expecting to delete successfully")
--- a/internal/controller/gatewayclass_test.go
+++ b/internal/controller/gatewayclass_test.go
@@ -1,0 +1,229 @@
+// Copyright Project Contour Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package controller
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	gatewayv1alpha1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
+)
+
+var _ = Describe("GatewayClass Controller", func() {
+	BeforeEach(func() {
+		// Add any setup steps that needs to be executed before each test
+	})
+
+	AfterEach(func() {
+		// Add any teardown steps that needs to be executed after each test
+	})
+
+	Context("Managed GatewayClass", func() {
+		It("Should surface admitted status", func() {
+
+			gen, err := generateRandomString(10, charset)
+			Expect(err).NotTo(HaveOccurred())
+
+			key := types.NamespacedName{Name: "test-gatewayclass-" + gen}
+
+			admitted := &gatewayv1alpha1.GatewayClass{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      key.Name,
+					Namespace: key.Namespace,
+				},
+				Spec: gatewayv1alpha1.GatewayClassSpec{Controller: gcController},
+			}
+
+			// Create
+			Expect(cl.Create(context.Background(), admitted)).Should(Succeed())
+
+			By("Expecting admitted status")
+			Eventually(func() bool {
+				gc := &gatewayv1alpha1.GatewayClass{}
+				_ = cl.Get(context.Background(), key, gc)
+				return isAdmitted(gc)
+			}, timeout, interval).Should(BeTrue())
+
+			// Delete
+			By("Expecting to delete successfully")
+			Eventually(func() error {
+				gc := &gatewayv1alpha1.GatewayClass{}
+				_ = cl.Get(context.Background(), key, gc)
+				return cl.Delete(context.Background(), gc)
+			}, timeout, interval).Should(Succeed())
+
+			By("Expecting to delete finish")
+			Eventually(func() error {
+				gc := &gatewayv1alpha1.GatewayClass{}
+				return cl.Get(context.Background(), key, gc)
+			}, timeout, interval).ShouldNot(Succeed())
+		})
+	})
+	Context("Unmanaged GatewayClass", func() {
+		It("Should surface waiting status", func() {
+
+			// Test a GatewayClass that should not be managed by Contour.
+			gen, err := generateRandomString(10, charset)
+			Expect(err).NotTo(HaveOccurred())
+
+			key := types.NamespacedName{Name: "test-gatewayclass-" + gen}
+			waiting := &gatewayv1alpha1.GatewayClass{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      key.Name,
+					Namespace: key.Namespace,
+				},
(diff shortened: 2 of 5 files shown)


stevesloka (commented):
I played with this last week working on https://github.com/projectcontour/contour/issues/3705. This will need to run the setup bits to add the etcd/apiserver logic for the tests to run. 

I personally use Goland for development, what I didn't like, was that after adding this, I couldn't use the built-in tools anymore, but had to use the makefile task. 

Might be nice to split these out into an "integration" test along with pulling out the logic of the controllers to let that be easily unit testable, but then allow these to be a "faster" set of e2e tests.

danehans:
> I played with this last week working on #3705. This will need to run the setup bits to add the etcd/apiserver logic for the tests to run.

@stevesloka I update the PR to use kubebuilder v2.x since v3 removes the test tools, i.e. etcd. However, I'm still having issues with the CI job to setup kubebuilder. We have to set the `KUBEBUILDER_ASSETS` env var due to https://github.com/kubernetes-sigs/kubebuilder/issues/326. The same workflow passes locally, but not in CI. When you have a moment, PTAL and let me know if you have any troubleshooting suggestions.

> Might be nice to split these out into an "integration" test along with pulling out the logic of the controllers to let that be easily unit testable, but then allow these to be a "faster" set of e2e tests.

The approach follows the [pattern](https://book.kubebuilder.io/cronjob-tutorial/writing-tests.html) for kubebuilder-based controllers.

sunjayBhatia:
@danehans seems like this is part of the issue: https://github.com/projectcontour/contour/runs/2804949179?check_suite_focus=true#step:6:8

the env var isn't being expanded the way it is being set up in the pipeline

internal/controller/suite_test.go
```diff
@@ -0,0 +1,154 @@
...
+func (r *gatewayClassReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&gatewayv1alpha1.GatewayClass{}).
+		Complete(r)
+}
+
+// generateRandomString returns a securely generated random string with the provided
+// length. It will return an error if the system's secure random number generator
+// fails to function correctly, in which case the caller should not continue.
+func generateRandomString(length int, charset string) (string, error) {
```
sunjayBhatia, line 118: might be able to replace with a k8s util helper: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/rand#String

internal/controller/gatewayclass_test.go
```diff
@@ -0,0 +1,229 @@
...
+						Name:  "baz",
+					},
+				},
+			}
+
+			// Create
+			Expect(cl.Create(context.Background(), unsupported)).Should(Succeed())
+
+			By("Expecting not admitted status")
+			Eventually(func() bool {
```
sunjayBhatia, line 222: Should this be a `Consistently` instead? since the default status is admitted==false

sunjayBhatia: similar for other places we're creating a gatewayclass and expecting it to not be admitted

danehans: @sunjayBhatia the reason differs from the [default](https://github.com/kubernetes-sigs/gateway-api/blob/v0.3.0/apis/v1alpha1/gatewayclass_types.go#L44) of "Waiting".

internal/controller/suite_test.go
danehans: ack

internal/controller/gatewayclass_test.go
sunjayBhatia: doesn't look like `isAdmitted` checks the reason, only looks like it is checking status right now

danehans: @sunjayBhatia I played around with `Consistently` and had no luck. I found this [ref](https://medium.com/@william.la.martin/ginkgotchas-yeh-also-gomega-13e39185ec96) that states results of a returned func should not be passed to `Consistently`.

> doesn't look like isAdmitted checks the reason, only looks like it is checking status right now

I added `isWaiting` that checks for the "Waiting" status reason.

danehans:
Commit `353d0e6` removed the `isWaiting()` as the test case was flaking.

internal/controller/gatewayclass.go
```diff
@@ -116,7 +116,9 @@ func (r *gatewayClassReconciler) Reconcile(ctx context.Context, request reconcil
 	}
 
 	// Pass the new changed object off to the eventHandler.
-	r.eventHandler.OnAdd(gc)
+	if r.eventHandler != nil {
```
danehans, line 119: The condition was added to support testing the reconciliation.

stevesloka:
I downgraded to v2.3.1 and still get these errors. Any reason to not move to v3 now and not have to do it later?  

```
 $ make check                                                                       [12:03:25]
go install -mod=readonly -v -ldflags="-s -w -X github.com/projectcontour/contour/internal/build.Version=736c0ea0 -X github.com/projectcontour/contour/internal/build.Sha=736c0ea04f32da3ae7579e1c4a3ca767dc7879ff -X github.com/projectcontour/contour/internal/build.Branch=danehans-controller_test_suite " -tags "oidc gcp osusergo netgo" github.com/projectcontour/contour/cmd/contour
go test -tags "oidc gcp osusergo netgo" -cover -mod=readonly github.com/projectcontour/contour/...
?   	github.com/projectcontour/contour/apis/projectcontour	[no test files]
ok  	github.com/projectcontour/contour/apis/projectcontour/v1	(cached)	coverage: 5.0% of statements
?   	github.com/projectcontour/contour/apis/projectcontour/v1alpha1	[no test files]
ok  	github.com/projectcontour/contour/cmd/contour	(cached)	coverage: 19.1% of statements
?   	github.com/projectcontour/contour/hack	[no test files]
ok  	github.com/projectcontour/contour/internal/annotation	(cached)	coverage: 82.1% of statements
?   	github.com/projectcontour/contour/internal/build	[no test files]
?   	github.com/projectcontour/contour/internal/certgen	[no test files]
ok  	github.com/projectcontour/contour/internal/contour	(cached)	coverage: 37.6% of statements
Running Suite: Runtime Controller Suite
=======================================
Random Seed: 1623945814
Will run 4 of 4 specs

STEP: Bootstrapping the test environment
time="2021-06-17T12:03:34-04:00" level=info msg="starting control plane" api server flags="[]" logger=controller-runtime.object-cache.eventhandler.EnqueueRequestForObject.eventhandler.EnqueueRequestForOwner.metrics.predicate.eventFilters.source.EventHandler.source.healthz.admission.certwatcher.webhook.builder.client.config.test-env.conversion-webhook
time="2021-06-17T12:03:55-04:00" level=error msg="unable to start the controlplane" error="timeout waiting for process kube-apiserver to start" logger=controller-runtime.object-cache.eventhandler.EnqueueRequestForObject.eventhandler.EnqueueRequestForOwner.metrics.predicate.eventFilters.source.EventHandler.source.healthz.admission.certwatcher.webhook.builder.client.config.test-env.conversion-webhook tries=0
time="2021-06-17T12:04:15-04:00" level=error msg="unable to start the controlplane" error="timeout waiting for process kube-apiserver to start" logger=controller-runtime.object-cache.eventhandler.EnqueueRequestForObject.eventhandler.EnqueueRequestForOwner.metrics.predicate.eventFilters.source.EventHandler.source.healthz.admission.certwatcher.webhook.builder.client.config.test-env.conversion-webhook tries=1
Timeout [60.001 seconds]
[BeforeSuite] BeforeSuite 
/Users/slokas/gdev/projectcontour/contour/internal/controller/suite_test.go:53

  Timed out

  /Users/slokas/gdev/projectcontour/contour/internal/controller/suite_test.go:53
------------------------------


Ran 4 of 0 Specs in 60.069 seconds
FAIL! -- 0 Passed | 4 Failed | 0 Pending | 0 Skipped
You're using deprecated Ginkgo functionality:
=============================================
Ginkgo 2.0 is under active development and will introduce (a small number of) breaking changes.
To learn more, view the migration guide at https://github.com/onsi/ginkgo/blob/v2/docs/MIGRATING_TO_V2.md
To comment, chime in at https://github.com/onsi/ginkgo/issues/711

  You are passing a Done channel to a test node to test asynchronous behavior.  This is deprecated in Ginkgo V2.  Your test will run synchronously and the timeout will be ignored.
  Learn more at: https://github.com/onsi/ginkgo/blob/v2/docs/MIGRATING_TO_V2.md#removed-async-testing
    /Users/slokas/gdev/projectcontour/contour/internal/controller/suite_test.go:53
  You are using a custom reporter.  Support for custom reporters will likely be removed in V2.  Most users were using them to generate junit or teamcity reports and this functionality will be merged into the core reporter.  In addition, Ginkgo 2.0 will support emitting a JSON-formatted report that users can then manipulate to generate custom reports.

  If this change will be impactful to you please leave a comment on https://github.com/onsi/ginkgo/issues/711
  Learn more at: https://github.com/onsi/ginkgo/blob/v2/docs/MIGRATING_TO_V2.md#removed-custom-reporters

To silence deprecations that can be silenced set the following environment variable:
  ACK_GINKGO_DEPRECATIONS=1.16.2

--- FAIL: TestAPIs (60.07s)
FAIL

```

danehans:
xref https://github.com/projectcontour/contour/issues/3832 for the use of v1.19.2 of controller-runtime test tools, i.e. kube-apiserver.

danehans:
`upgrade (kubernetes:n-1)` fails with:
```
./test/scripts/install-contour-release.sh v1.16.0
error: unable to read URL "https://projectcontour.io/quickstart/v1.16.0/contour.yaml", server reported 502 Bad Gateway, status code=502
```

This appears to be a flake, rerunning CI jobs.

danehans:
> I downgraded to v2.3.1 and still get these errors

You may still be using your previously installed version of the testenv tools. Run `./hack/actions/install-kubernetes-toolchain.sh $DEST_DIR` to make sure you're using the correct version.

> Any reason to not move to v3 now and not have to do it later?

See https://github.com/projectcontour/contour/issues/3832#issuecomment-867876921.

youngnick:
Just reminding everyone that Friday is the last day to get @danehans' changes on this one, so it would be good to try and get everything sorted if we can.

stevesloka (commented):
I think this all looks good to me. I was able to get the tests to run once the tooling was setup properly on my local machine. We may need a new Makefile task to help automate this since we're getting lots of different deps to build. Or have a backup dockerfile which gets around all of that for new folks. 

/lgtm pending @sunjayBhatia's comments.

sunjayBhatia (approved):
generally yeah looks good, I have a few local changes I would like to make but can do that in a follow up PR, just to clean up some of the Ginkgo patterns etc. and get some more coverage

We should also probably mock the event handler so we can ensure it actually gets called from the controller, that could be done in a follow up

Branch/PR here with some changes on top of this: https://github.com/projectcontour/contour/pull/3844

skriss:
Looks like we're good to go here, merging.

PR #3778, opened 2021-06-07
site: Fix links on deploy-options pages

stevesloka:
Updates #3712

### DIFF
--- a/site/content/docs/v1.11.0/deploy-options.md
+++ b/site/content/docs/v1.11.0/deploy-options.md
@@ -209,14 +209,14 @@ $ kubectl delete ns contour-operator
 ```
 
 [1]: #running-without-a-kubernetes-loadbalancer
-[2]: {{site.github.repository_url}}/tree/{{page.version}}/examples/contour/README.md
+[2]: {{< param github_url>}}/tree/{{< param version >}}/examples/contour
 [3]: #host-networking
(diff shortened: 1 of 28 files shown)


skriss (approved):

site/content/docs/v1.11.0/deploy-options.md
```diff
@@ -209,14 +209,14 @@ $ kubectl delete ns contour-operator
 ```
 
 [1]: #running-without-a-kubernetes-loadbalancer
-[2]: {{site.github.repository_url}}/tree/{{page.version}}/examples/contour/README.md
+[2]: {{< param github_url>}}/tree/{{< param version >}}/examples/contour
 [3]: #host-networking
-[4]: {% link _guides/proxy-proto.md %}
+[4]: /guides/proxy-proto.md
```
sunjayBhatia, line 214: not this PRs remit potentially but I think this page has the wrong URL used for the Proxy protocol hyperlink, should be 4 not 9

sunjayBhatia (approved):
link corrections look good, one mistake in the page that was from before this change, ok with me if we fix later or in this one

site/content/docs/v1.11.0/deploy-options.md
stevesloka: nah good catch, I'll just update here to save the review spam.

PR #3781, opened 2021-06-07
site: Fixup the upgrade guide

stevesloka:
I'd like to revamp this guide, it's sort of a mess, but for now, this makes the guide work with all the links properly. 

Updates #3712


youngnick (approved):
Agreed that needs needs a rework, but that the links should work in the meantime.

skriss (approved):
Looks fine for now, when we clean this page up we should revisit exactly which docs version each link is pointing to (version-specific links should be pinned to a version, not using `latest` or `main`).

PR #3785, opened 2021-06-08
site: bulk replacement of Jekyll templates

skriss:
Does the following replacements:

```
{{site.github.repository_url}}              {{< param github_url >}}
{{site.footer_social_links.Slack.url}}      {{< param slack_url >}}
{{site.latest}}                             {{< param latest_release_tag_name >}}
{{site.github.latest_release.tag_name}}     {{< param latest_release_tag_name >}}
{{page.version}}                            {{< param version >}} (only for versions 1.9.0+)
```

Separately, going to file an issue to de-dupe `latest_release_tag_name` and `docs_latest` params, I think in practice they're always the same. I think we'll have to update our automation for this as well.

### DIFF
--- a/site/content/docs/main/config/ingress.md
+++ b/site/content/docs/main/config/ingress.md
@@ -77,4 +77,4 @@ If not provided, Contour will use the address of the Envoy service using the pas
 [5]: https://kubernetes.io/docs/concepts/services-networking/ingress/#hostname-wildcards
 [6]: https://kubernetes.io/docs/concepts/services-networking/ingress/#path-types
 [7]: https://kubernetes.io/docs/concepts/services-networking/ingress/#tls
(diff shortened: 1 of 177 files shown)


skriss:
Still need to test this with the site preview.

skriss:
Looks like `{{< param version >}}` doesn't work for the older site versions that aren't in https://github.com/projectcontour/contour/blob/main/site/config.yaml#L39-L52

skriss:
A quick spot-check looks good here

stevesloka (changes requested):
All the `{{page.version}}` need to be replaced. Should be able to use `{{< param version >}}`

stevesloka (approved):
lgtm with the known issue with page_version.

sunjayBhatia (approved):

PR #3787, opened 2021-06-08
site: update page version template

skriss:
Follow-up to #3785.


skriss:
cherry-picked this into #3785.

PR #3797, opened 2021-06-10
design: Update GatewayAPI design doc

stevesloka:
Updates the GatewayAPI design doc for a optional, Contour managed Envoy controller.


design/gateway-apis-implementation.md
```diff
@@ -62,69 +61,77 @@ Project Contour (the organisation) may investigate a Gateway API based layer 4 s
...
+When a Gateway is not supplied, Contour will not do Gateway API processing.
+Contour will merge Listeners within its Gateway. (See the detailed design for the exact rules Contour will use for this.)
+
+### Dynamic Envoy Configuration
+
+Typically, the fleet of Envoy's that connect to Contour to get their configuration information has been managed manually by the cluster operator.
+Given a more dynamic configuration API, it's convenient to have the Envoy fleet configuration in Kubernetes be managed by a controller.
+This allows cluster operator to simply deploy a Contour deployment, leaving the details about how Envoy is provisioned in the cluster as well as exposed outside of the cluster to a configuration file or CRD.
+
+This behavior is meant to be optional and layered, such that users who want a fully managed environment can deploy the Contour Operator which manages instances of Contour which in turn manages instances of Envoy.
```
sunjayBhatia, line 76: nice 👍🏽

design/gateway-apis-implementation.md
```diff
@@ -62,69 +61,77 @@ Project Contour (the organisation) may investigate a Gateway API based layer 4 s
 
 ### Contour Gateway model
 
-Contour considers a Gateway to describe a single Envoy deployment, deployed and managed by something outside of itself.
-Contour expects to be supplied with the full name (name and namespace) of a single Gateway, which it will watch and update the status of.
-This will be a configuration file parameter.
-When a Gateway is not supplied, Contour will not do Gateway API processing, or in other words, the existing `--experimental-service-apis` flag will be obsoleted and removed).
-Inside Contour will merge Listeners within its Gateway. (See the detailed design for the exact rules Contour will use for this.)
+Contour considers a Gateway to describe a single Envoy deployment.
+Contour expects to be supplied with the full name and namespace of a single Gateway as well as the GatewayClass controller name.
```
sunjayBhatia, line 65: I forget the conversation on this, are we still going to require the gateway name/namespace or are we phasing that out for just the controller name? I vaguely remember something about how we choose which Gateway to admit following the conflict resolution guidelines in the spec etc. but I don't remember where that got written down

design/gateway-apis-implementation.md
```diff
@@ -62,69 +61,77 @@ Project Contour (the organisation) may investigate a Gateway API based layer 4 s
...
 
 #### GatewayClass
 
-Contour does not watch or interact with GatewayClass in any way.
+Contour watches for GatewayClass resources which can, optionally, define how an instance of Contour should be configured.
+A GatewayClass can reference a configuration resource (Configmap or CRD) which describes how the fleet of Envoys should be provisioned in the cluster along with how they are exposed outside of the cluster.
+
+If the GatewayClass doesn't reference any configuration resources, then a standard set of configuration is utilized much like the Contour QuickStart today.
+
+Contour will need to be configured with the controller name matching the GatewayClass's `Spec.Controller` field.
```
sunjayBhatia, line 110: maybe add a note about the convention/format of controller name/default?

sunjayBhatia (commented):
just a few quickfire comments so far

maybe something to add in an implementation section could be how we're also using this as an opportunity to transition to use controller runtime? might be too much of a detail, idk

design/gateway-apis-implementation.md
danehans: > I forget the conversation on this, are we still going to require the gateway name/namespace or are we phasing that out for just the controller name?

It was my understanding that the gateway ns/name config will be removed and Contour will reconcile all gateway's that reference a gatewayclass matching its `controllerName`.

design/gateway-apis-implementation.md
```diff
@@ -62,69 +61,77 @@ Project Contour (the organisation) may investigate a Gateway API based layer 4 s
...
+In general, Contour will update the status of any object that comes into scope with its details, ensuring that the `observedGeneration` field is set correctly.
 When objects fall out of scope, any status set by Contour will not be removed.
 It's expected that things that check the status will also check the `observedGeneration` to check if the status information is up-to-date with the current object generation.
 
 ### Per-object design notes
 
 #### GatewayClass
 
-Contour does not watch or interact with GatewayClass in any way.
+Contour watches for GatewayClass resources which can, optionally, define how an instance of Contour should be configured.
```
danehans, line 105: We may want to consider supporting a gatewayclass referencing the Contour ConfigMap for params until a config CRD is established.

design/gateway-apis-implementation.md
```diff
@@ -62,69 +61,77 @@ Project Contour (the organisation) may investigate a Gateway API based layer 4 s
...
+A GatewayClass can reference a configuration resource (Configmap or CRD) which describes how the fleet of Envoys should be provisioned in the cluster along with how they are exposed outside of the cluster.
```
danehans, line 106: We may want to consider supporting a gatewayclass referencing the Contour ConfigMap for params until a config CRD is established.

design/gateway-apis-implementation.md
```diff
@@ -62,69 +61,77 @@ Project Contour (the organisation) may investigate a Gateway API based layer 4 s
...
 
-Contour considers a Gateway to describe a single Envoy deployment, deployed and managed by something outside of itself.
-Contour expects to be supplied with the full name (name and namespace) of a single Gateway, which it will watch and update the status of.
-This will be a configuration file parameter.
-When a Gateway is not supplied, Contour will not do Gateway API processing, or in other words, the existing `--experimental-service-apis` flag will be obsoleted and removed).
-Inside Contour will merge Listeners within its Gateway. (See the detailed design for the exact rules Contour will use for this.)
+Contour considers a Gateway to describe a single Envoy deployment.
+Contour expects to be supplied with the full name and namespace of a single Gateway as well as the GatewayClass controller name.
+Contour will update the status of both the Gateway & GatewayClass.
+When a Gateway is not supplied, Contour will not do Gateway API processing.
```
youngnick, line 67: If we are moving to pass a `controllerName` instead of a Gateway name/namespace, we'll need to update this too:
```suggestion
When a GatewayClass `controllerName` is not configured, Contour will not do Gateway API processing.
```

design/gateway-apis-implementation.md
youngnick, line 105: If we move to only taking GatewayClass, we'll need to remove `optionally` here.

design/gateway-apis-implementation.md
youngnick: Agreed that the default should be documented here.

youngnick (changes requested):
This looks great, but we need to agree on what the destination is for configuring Gateway API in Contour.

Personally, I think we should plan to move to configuring only the `controllerName` string in Contour, and having Contour find the correct Gateway from that. We should deprecate the current config and migrate over a version or two, with both accepted for a while though.

If we do decide that is the plan, I've noted some changes that we need, and it also means we will need to have clear rules around GatewayClass conflict resolution, which should be documented here.

I don't think that having both GatewayClass and specifying Gateway by name/namespace is a good idea, to my knowledege we are the only implementation doing that, and we should avoid non-standard behavior as much as possible.

design/gateway-apis-implementation.md
stevesloka: But there's no way to specify a single controller in the gatewayclass. What happens when multiple gateways match the same controller? 

```
$ kubectl get gatewayclass,gateway -A                                                                                 
NAME                                                  CONTROLLER                             AGE
gatewayclass.networking.x-k8s.io/contour-class        projectcontour.io/ingress-controller   2m47s
gatewayclass.networking.x-k8s.io/contour-class-dupe   projectcontour.io/ingress-controller   2m47s

NAMESPACE        NAME                                       CLASS                AGE
projectcontour   gateway.networking.x-k8s.io/contour        contour-class        2m24s
projectcontour   gateway.networking.x-k8s.io/contour-dupe   contour-class-dupe   20s
```

stevesloka:
> Personally, I think we should plan to move to configuring only the controllerName string in Contour, and having Contour find the correct Gateway from that. We should deprecate the current config and migrate over a version or two, with both accepted for a while though.

This was my discussion here (https://github.com/projectcontour/contour/issues/3617). Have your opinions changed on the approach?

design/gateway-apis-implementation.md
stevesloka: I guess the suggestion is that the controllername is a convention matching the contour namespace/name. We can do that, just don't think folks are going to get this right.

design/gateway-apis-implementation.md
stevesloka: yup its left open right now: `A GatewayClass can reference a configuration resource (Configmap or CRD)`

design/gateway-apis-implementation.md
stevesloka: The way @danehans designed it in the operator is you don't need to specify a paramsRef resource which would give you a set of generic defaults.

youngnick:
> > Personally, I think we should plan to move to configuring only the controllerName string in Contour, and having Contour find the correct Gateway from that. We should deprecate the current config and migrate over a version or two, with both accepted for a while though.
> 
> This was my discussion here (#3617). Have your opinions changed on the approach?

Rereading that issue, I think that I've always been in favor of using the controllerName field as intended, rather than specifying the Gateway name and class. It does work out a little weird because Contour is so bound to a single Gateway at the moment, but the more we add features like Envoy management, the less that is the case.

If we do move towards controllerName being the key discriminator, then we need to document the conflict rules for GatewayClasses, as you said. I suspect that something along the lines of "oldest matching GatewayClass object wins" would probably be the least likely to break, and would be consistent with the broader Gateway API philosophy on conflicts. (Note that we should probably also raise this upstream, I don't think it's well defined there either).

At the end of the day, I *am* in favour of using a controllerName string, just because it forces our model closer to the plain upstream model.

design/gateway-apis-implementation.md
stevesloka: There is no default from what I understand. The examples do generate a field that has a "default", however, it's not applied automatically from code.

stevesloka: @danehans do you foresee a default in Contour?

stevesloka:
@sunjayBhatia @danehans @youngnick design updated with comments.

design/gateway-apis-implementation.md
```diff
@@ -163,23 +169,26 @@ The output of this watcher is Envoy configuration.
 ## Detailed Design
 
 ### Configuration
-Contour will have an entry added to the config file for the Gateway it should watch.
+Contour will have an entry added to the config file for the `ControllerName` it should watch.
+The `ControllerName` takes the format of `projectcontour.io/<namespace>/contour` which requires the `<namespace>` param to be specified matching the namespace where Contour is deployed.
```
stevesloka, line 173: Since we're going to add Class logic to pick the oldest? The format of this probably doesn't matter, but should follow the upstream convention. 

Should also look to make sure the upstream admission controller also enforces the correct pattern.

design/gateway-apis-implementation.md
```diff
@@ -62,69 +61,77 @@ Project Contour (the organisation) may investigate a Gateway API based layer 4 s
 
 ### Contour Gateway model
 
-Contour considers a Gateway to describe a single Envoy deployment, deployed and managed by something outside of itself.
-Contour expects to be supplied with the full name (name and namespace) of a single Gateway, which it will watch and update the status of.
-This will be a configuration file parameter.
-When a Gateway is not supplied, Contour will not do Gateway API processing, or in other words, the existing `--experimental-service-apis` flag will be obsoleted and removed).
-Inside Contour will merge Listeners within its Gateway. (See the detailed design for the exact rules Contour will use for this.)
+Contour considers a Gateway to describe a single Envoy deployment.
+Contour expects to be supplied with a `ControllerName` which is required to match a `GatewayClass`.
```
skriss, line 65: Contour's expectation is that there's only one `GatewayClass` for a given `ControllerName`, and only one `Gateway` for that `GatewayClass`, right? Might be helpful to state that explicitly here since that's different than what the API itself allows.

design/gateway-apis-implementation.md
```diff
@@ -62,69 +61,77 @@ Project Contour (the organisation) may investigate a Gateway API based layer 4 s
...
-Contour does not watch or interact with GatewayClass in any way.
+Contour watches for GatewayClass resources which can, optionally, define how an instance of Contour should be configured.
+A GatewayClass can reference a configuration resource (Configmap or CRD) which describes how the fleet of Envoys should be provisioned in the cluster along with how they are exposed outside of the cluster.
+
+If the GatewayClass doesn't reference any configuration resources, then a standard set of configuration is utilized much like the Contour QuickStart today.
+
+Contour will need to be configured with the controller name matching the GatewayClass's `Spec.Controller` field.
 
 #### Gateway
 Contour will watch its configured Gateway object.
```
skriss, line 113: Since Contour's not configured with a `Gateway` anymore I'd spell this out a little more - i.e. Contour watches the `Gateway` that uses the `GatewayClass` that has the configured `ControllerName`.

design/gateway-apis-implementation.md
```diff
@@ -62,69 +61,77 @@ Project Contour (the organisation) may investigate a Gateway API based layer 4 s
...
+
+### Dynamic Envoy Configuration
+
+Typically, the fleet of Envoy's that connect to Contour to get their configuration information has been managed manually by the cluster operator.
+Given a more dynamic configuration API, it's convenient to have the Envoy fleet configuration in Kubernetes be managed by a controller.
+This allows cluster operator to simply deploy a Contour deployment, leaving the details about how Envoy is provisioned in the cluster as well as exposed outside of the cluster to a configuration file or CRD.
+
+This behavior is meant to be optional and layered, such that users who want a fully managed environment can deploy the Contour Operator which manages instances of Contour which in turn manages instances of Envoy.
+However, other users may not desire that level of controller driven infrastructure.
+This design leaves users with their choice of deployment models.
```
skriss, line 78: I think it'd be helpful to have a detailed design section that goes into this more since this is a big new feature. How does a user opt into getting the Envoy fleet automatically managed? How do they opt out? What are the configuration knobs/where are they specified? etc.

design/gateway-apis-implementation.md
```diff
@@ -51,7 +51,6 @@ There should never be annotations required for functionality on Gateway API obje
 ## Non Goals
 - No TCPRoute or UDPRoute support, that is, no support for arbitrary TCP/UDP forwarding.
```
shashankram, line 52: Is there a plan on supporting this in the future?

shashankram (commented):
Thanks @stevesloka for pointing me to this design doc. I am interested in leveraging Contour as an ingress gateway using the Gateway API. This doc answered many questions I had, and I have a few additional ones that I hope you can provide answers for:

1. Would it be possible to document the supported deployment models and what Contour supports at the moment (ex. how to configure a single contour ingress gateway to proxy traffic to backends residing in different namespaces). The [existing demo](https://projectcontour.io/guides/gateway-api/) is simplified to have both the gateway and backend in the same namespace.
2. Will the listener ports specified in the Gateway resource influence the ports over which the gateway can be accessed from a client outside the cluster. In other words, if the gateway is deployed as a load balancer service in k8s, do the listener ports specified in the Gateway resource have any bearing on the load balancer service's ports? 
3. For TLS/mTLS proxying to backends using the Gateway API, will Contour support specifying the [BackendPolicy resource] (https://gateway-api.sigs.k8s.io/references/spec/#networking.x-k8s.io/v1alpha1.BackendTLSConfig) to validate backend certs, and also allow [specifying the client cert](https://projectcontour.io/docs/main/config/upstream-tls/) to use in case the backend connection needs to be mTLS?

youngnick:
@shashankram, I think the answer to most of your questions generally is "yes, but we haven't got there yet". To answer your points specifically:

1. I agree that this is a worthy goal, but as we are still in the process of figuring out how all of the pieces fit together, it's not really achievable yet.
2. Today, I think not. However, the idea of having Contour manage the Envoy deployment is to allow exactly this feature.
3. Contour has support today for supplying backend TLS details via other methods, we will support BackendPolicy or whatever is required to do that in the future. We just haven't got there yet.

The exception to the "yes, but not yet" rule is that Contour is not currently planning to support the TCPRoute and UDPRoute types, as we are focussed on being a Layer 7 ingress controller, not a layer 4 load balancer. Cluster networking means that doing this sort of routing using an in-cluster proxy has unfavourable interactions with some TCP and UDP properties that take a while to explain, but boil down to "it probably won't work the way you think it will".

design/gateway-apis-implementation.md
youngnick: Agreed, we should say that Contour will only use a single Gateway and GatewayClass, and say that the rules for choosing a winner in the event there's more than one are specified in this document.

design/gateway-apis-implementation.md
```diff
@@ -62,69 +61,77 @@ Project Contour (the organisation) may investigate a Gateway API based layer 4 s
...
-Inside Contour will merge Listeners within its Gateway. (See the detailed design for the exact rules Contour will use for this.)
+Contour considers a Gateway to describe a single Envoy deployment.
+Contour expects to be supplied with a `ControllerName` which is required to match a `GatewayClass`.
+Contour will update the status of both the Gateway & GatewayClass.
+When a `ControllerName` is not supplied, Contour will not do Gateway API processing.
+Contour will merge Listeners within its Gateway. (See the detailed design for the exact rules Contour will use for this.)
+
+### Dynamic Envoy Configuration
+
+Typically, the fleet of Envoy's that connect to Contour to get their configuration information has been managed manually by the cluster operator.
```
youngnick, line 72: Nit:
```suggestion
Typically, the fleet of Envoys that connect to Contour to get their configuration information has been managed manually by the cluster operator.
```

design/gateway-apis-implementation.md
youngnick: Ah right, I read that wrong. Maybe something like this?
```suggestion
Contour watches for GatewayClass resources which can optionally refer to a configuration object that will then be used for Contour configuration.
```

design/gateway-apis-implementation.md
youngnick: Since we're using the presence of a value in this field to determine whether to switch on the Gateway API processors, it makes sense there's no default specified here.

design/gateway-apis-implementation.md
youngnick: Agreed, needs a caveat about confict resolution rules too.

design/gateway-apis-implementation.md
youngnick: We can probably say something like "The `ControllerName` should take the format of`, since upstream _recommends_ that controllers namespace this, but does not _require_ it.

design/gateway-apis-implementation.md
```diff
@@ -163,23 +169,26 @@ The output of this watcher is Envoy configuration.
...
 ```
 
-The current flag `--experimental-service-apis` will be removed.
-The boolean it represents will be deduced from the following rules:
-- the gateway is configured in the config file.
-- the referenced gateway exists.
+Contour will look for a Gateway + GatewayClass that matches the `ControllerName` field specified in the configuration file.
+If it does not find a matching Gateway/GatewayClass, then it will begin processing GatewayAPI resources once they are available. 
+
+Should two different Gateway or GatewayClasses be found matching the same `ControllerName`, then Contour will utilize the oldest created object and set a status on the other(s) which describes that object as being rejected.
```
youngnick, line 186: This should specify what Condition will be emitted, what Reason should be set, and so on.

design/gateway-apis-implementation.md
```diff
@@ -163,23 +169,26 @@ The output of this watcher is Envoy configuration.
...
+  controllerName: 
+  name: gatewayname  # <--- deprecate
+  namespace: gatewaynamespace  # <-- deprecate
 ```
 
-The current flag `--experimental-service-apis` will be removed.
-The boolean it represents will be deduced from the following rules:
-- the gateway is configured in the config file.
-- the referenced gateway exists.
+Contour will look for a Gateway + GatewayClass that matches the `ControllerName` field specified in the configuration file.
```
youngnick, line 183: This probably needs to be broken up to be more specific, so that we say that Contour will find a single valid GatewayClass, and then any Gateways that reference the GatewayClass.

Similarly, below should say that if Contour doesn't find any matching GatewayClasses, then no Gateways will be selected until there is one.

youngnick (changes requested):
This is looking super close, I think we need a bit more specificity in some of the details - ideally we shouldn't need to decide status details and so on as part of implementation.

shashankram:
> @shashankram, I think the answer to most of your questions generally is "yes, but we haven't got there yet". To answer your points specifically:
> 
> 1. I agree that this is a worthy goal, but as we are still in the process of figuring out how all of the pieces fit together, it's not really achievable yet.
> 2. Today, I think not. However, the idea of having Contour manage the Envoy deployment is to allow exactly this feature.
> 3. Contour has support today for supplying backend TLS details via other methods, we will support BackendPolicy or whatever is required to do that in the future. We just haven't got there yet.
> 
> The exception to the "yes, but not yet" rule is that Contour is not currently planning to support the TCPRoute and UDPRoute types, as we are focussed on being a Layer 7 ingress controller, not a layer 4 load balancer. Cluster networking means that doing this sort of routing using an in-cluster proxy has unfavourable interactions with some TCP and UDP properties that take a while to explain, but boil down to "it probably won't work the way you think it will".

Thanks @youngnick for answering all my questions and providing clarifications where necessary. Is there a rough timeline for this feature to be delivered end-to-end? I understand this is a huge change, so curious if the project had something specific in mind so potential users can plan accordingly.

youngnick:
Thanks @shashankram. Right now, we only have a timeframe for when the Gateway API will move to its second alpha. I don't really see this feature being really ready for full production use until the upstream is at least beta (since that implies a much stronger level of API guarantee). I think that *that* timeframe is probably longer than a few months, but less than a year. As soon as the beta becomes available, we will transition Contour to it as fast as we can. In terms of GA for the API, again, very difficult to say. As before, I can confidently say that when the API goes GA, Contour will follow as fast as we are able.

Sorry I can't give you a better answer.

design/gateway-apis-implementation.md
```diff
@@ -27,22 +27,22 @@ In terms of the APIs themselves, the Gateway API has 3 primary API resources (ta
...
 Currently, Contour provides "TCP Proxying" that allows the forwarding of TLS streams based on the SNI information, which is precisely what the Gateway API TLSRoute object is for.
 If Project Contour (the organisation) does add support for TCP and UDP forwarding, it will not be in the `projectcontour/contour` repo, but will be a separate repo.
 
-This design is intended to cover the initial `v1alpha1` release of the Gateway API. We will aim to implement the core featureset of the APIs at that point.
+This design is intended to cover the initial alpha releases of the Gateway API. 
+We will aim to implement the core featureset of the APIs at that point.
 We will then work with the upstream community on features that Contour and HTTPProxy currently support, but the Gateway API do not, and how best to represent those features in the APIs.
-So, there are some features that are currently gaps for Contour (wildcard domain names, and exact path matching seem like big ones),
+So, there are some features that are currently gaps for Contour (wildcard domain names, and exact path matching, for example),
```
sunjayBhatia, line 40: nit: Aren't both of these supported now?

design/gateway-apis-implementation.md
sunjayBhatia: should we instead document the convention of how the controller name string is constructed? I remember someone mentioning we should do `projectcontour.io/<namespace-of-contour-instance>/<some-name>`

sunjayBhatia: ah i see a comment about this below, nvm

design/gateway-apis-implementation.md
stevesloka: yup!

design/gateway-apis-implementation.md
stevesloka: @xaleeks @youngnick ^^

design/gateway-apis-implementation.md
stevesloka: 👍

design/gateway-apis-implementation.md
stevesloka: 👍

design/gateway-apis-implementation.md
stevesloka: I actually just took that specific bit out, it doesn't need to match a convention, only be unique.

design/gateway-apis-implementation.md
stevesloka: There's a note about conflict resolution right after that paragraph.

design/gateway-apis-implementation.md
stevesloka: Yup it states that right below that sentence:
```
If it does not find a matching Gateway/GatewayClass, then it will begin processing GatewayAPI resources once they are available. 
```

design/gateway-apis-implementation.md
youngnick: Uh, this one isn't changed?

design/gateway-apis-implementation.md
youngnick: oh, 🤦 . Thanks.

youngnick (commented):
Some comments on small things in the document, I'll do a separate review for the managed Envoy bits.

design/gateway-apis-implementation.md
```diff
@@ -163,37 +169,82 @@ The output of this watcher is Envoy configuration.
...
+gateway:
+  controllerName: projectcontour.io/ingress-controller
+```
+
+_NOTE: Typically Envoy is deployed as a Daemonset, but not required (could also be a Deployment), so the term `fleet` is used to identify the set of Kubernetes managed pods which represent an Envoy instance._
+
+##### managed: true
+
+If envoy.managed is true, then Contour will automatically create a Envoy fleet/service.
+Additionally, if a `Gateway.ControllerName` is set then the ports that are assigned to the Envoy service will match the Gateway discovered based upon the `ControllerName`.
```
youngnick, line 210: We need to specify what happens in the case that `envoy.managed` is `true` and `gateway.controllerName` is not specified in this section too. Seems likely that we would do the same as below, and the `--envoy-service-*` flags would set the ports.

design/gateway-apis-implementation.md
```diff
@@ -163,37 +169,82 @@ The output of this watcher is Envoy configuration.
...
+  managed: true
+gateway:
+  controllerName: projectcontour.io/ingress-controller
+```
+
+_NOTE: Typically Envoy is deployed as a Daemonset, but not required (could also be a Deployment), so the term `fleet` is used to identify the set of Kubernetes managed pods which represent an Envoy instance._
+
+##### managed: true
+
+If envoy.managed is true, then Contour will automatically create a Envoy fleet/service.
```
youngnick, line 209: Given the discussion of Loadbalancer Services below, I'd suggest something like this:
```suggestion
If envoy.managed is true, then Contour will automatically create a Envoy fleet and a load balancer to get traffic to it. See below for a discussion on options for how Contour will do that.
```

design/gateway-apis-implementation.md
```diff
@@ -163,37 +169,82 @@ The output of this watcher is Envoy configuration.
...
+Additionally, if a `Gateway.ControllerName` is set then the ports that are assigned to the Envoy service will match the Gateway discovered based upon the `ControllerName`.
+
+Additional RBAC permissions will be required to the namespace where Contour is deployed to allow it to dynamically manage Daemonsets/Deployments/Services, etc.
+
+##### managed: false
+
+If envoy.managed is false, then Contour won't manage the Envoy fleet/service.
+It's up to the cluster operator to deploy and managed these resources.
```
youngnick, line 217: We need some specification here of what the cluster operator will need to do. We do this already in the example deployments, but we need to be specific about things like:
- External ports of the load balancer must match the ports Contour expects (otherwise insecure-secure redirects won't work).
- The external ports should almost always be 80 and 443.
- and so on.

Also, we should specify what happens when `managed` is `false` and cluster admins *do* specify Gateway API details.

youngnick (commented):
I've put some initial thoughts on the managed Envoy design below, so that we have some things to discuss in person. I think that this one is involved enough, has enough moving parts, and is a big enough change to our product direction, that we need to talk through it in a meeting rather than just over Github. But I wanted to give a chance for you all to read my thoughts before the meeting.

Please don't respond to these comments yet, let's talk in the team meeting more about them.

youngnick:
So, to reiterate what else I would like to see in this design doc under the `Managed Envoy` section, I think that we need to cover three variables:
- Using Operator or not
- Using Managed Envoy or not
- Using Gateway APIs or not

That's eight combinations, unfortunately, but I think we could do the first two variables in one set of four, assuming users are using Gateway APIs, then have a separate section that talks about the differences when  users are *not* using the Gateway APIs (since I don't think those differences will be that big).

So, there should be four sections:
- Operator, managed Envoy
- No Operator, managed Envoy
- Operator, no managed Envoy
- No Operator, no managed Envoy (this is basically what we have now).

In each of those four combinations, we need to know:
- How Envoy is provisioned
- How traffic gets to Envoy (if there's an upstream load balancer, how is it configured?)
- How Contour is provisioned
- How Contour is configured

I anticipate the most complex section to be the "how traffic gets to Envoy", because that's where we have to care about getting Loadbalancer Services (or equivalent Gateway) provisioned, which means we need to know about things like what cloud provider, what options for that cloud provider, and so on.

I think that the suggested option of having a Gateway->LoadBalancer Service bridge controller is actually pretty useful for the whole community, but it's probably a big enough thing to be deserving of its own design document. This document should at least sketch out what that controller would do, and more importantly, how the various options would be passed through a Gateway to a LB Service. ("How do you ask for an NLB instead of an ELB in AWS" is a good example case).

There's a lot of work here to flesh out all of that detail, so I'm definitely supportive of Aarthi's idea of splitting up the scope, it's also a good way for us all to be able to make sure we understand things in the same way.

But, this is @stevesloka's PR, so Steve, over to you. We're all ready to do whatever we need to to help get this done.

stevesloka:
@skriss @sunjayBhatia @youngnick I moved the gDoc back into this document.

design/gateway-apis-implementation.md
```diff
@@ -1,204 +1,420 @@
...
+* GatewayClass
+
+#### Finalizer workflows:
+
+* If the Envoy service is deleted what happens?
+* If the Envoy Daemonset is deleted what happens?
+* If the Gateway is deleted what happens?
+* If the GatewayClass is deleted what happens?
+
+### Exposing Envoy
```
sunjayBhatia, line 275: had a thought about this and what the Gateway listener ports "mean" and how we could service multiple scenarios of deployments

how much do users *need to customize* the ports Envoy listens on in the container/how much do we have to think about how the Envoy pods are routable?

If we choose a certain direction, are we locking people who use a managed Envoy to do a specific thing, which may be incompatible with how they have their cluster set up (routable pods or not, CNI, other constraints)?

if the ports on a Gateway listener represent the ports that a user outside a cluster will access apps via and we choose Envoy internal ports for you/make you configure them outside of the Gateway, that seems unwieldy potentially?

also, if we change how we treat those ports (are they on the LB service or just what envoy listens on) halfway through implementation/migration that will probably not be a great experience

thinking about this a bit and seeing all the places we have port configurations etc. in our example deployments (that people right now have the ability to customize to make work) makes me think we need to start earlier with the "gateways everywhere" approach for managed Envoy, and reconcile 2 different types of Gateways, one for configuring how Envoy listens, and another for how Envoy is exposed
- a migration step here could be to keep the Envoy LoadBalancer service and associated implementation logic a bit separate from the Gateway stuff, and make that for users using separate configuration
- another approach would be to just include it in example YAMLs and leave it up to the user how they want traffic into Envoy

sunjayBhatia: this was discussed a bit in the doc, but thought I would bring it up again, I dont know if we fully discussed the various migration steps to the end state of things and how all these ports interact when youre not using Gateway API though

sunjayBhatia: just worried we get locked into one deployment model or cluster configuration and find it hard to adapt later b/c theres no space in the API for adding more information to a Gateway for example

youngnick: I think it's not stated very clearly, but the assumption is that the Gateway's ports are the ports that the traffic should be sent to from the outside of the Gateway, and that doing other stuff here is implementation detail. I don't think that having the Gateway ports represent the actual container ports Envoy is running on is a good idea without a really good reason.

I think that for this bit, we should start with the easy case, and build outward.

The ports declared on Contour's Gateway are the externally-visible ports that traffic should be sent to via a client. When we are managing the Envoy inbound traffic using a Service or second Gateway object, those should be the requested ports, and we should use a port-translation mechanism like `targetPort` in Service to translate to a high port on the Envoy deployment. As we add more complicated Service requests in, we need to keep this in mind, that the *outer* ports of the Service need to be the same as the ones on the Contour Gateway.

People can customise those ports right now in our example YAMLs, and if they want to do that, managed Envoy is not ready for them yet. I think this is okay to start with. Then, people can tell us what they need, and we can help make that work.

The fact we have to build that is an unavoidable cost of adding the Envoy management functionality in to Contour, sadly.

It's really important to get this right because otherwise we'll break the insecure->secure redirects as well.

youngnick: tl;dr I think we should go ahead with just LB Service to start with, and consider the container ports and mappings an implementation detail.

youngnick (approved):
@stevesloka @sunjayBhatia @skriss, I would like to see this merged very soon please, as we're going to have to make changes to the implemenation again for v1alpha2. I don't think we should update this again, but we should get this overall approach approved and merged ASAP, since we are already working on parts of it.

I don't have any further changes to request personally, I think this is fine how it is.

design/gateway-apis-implementation.md
```diff
@@ -1,204 +1,420 @@
...
+The details defined in the Contour configuration CRD will be used to pre-populate annotations that will be stored in a Kubernetes configmap and referenced from the GatewayClass.Spec.ParametersRef.
+The Gateway Stub controller will be utilized to configure a new Kubernetes Service with the annotations defined in the configmap on the GatewayClass and send traffic to Envoy.
+Once implementers catch up, specific implementations will be utilized and the stub controller can be deprecated.
+
+New fields will be added to the Contour configuration file which will allow users to define the GVK of a resource used in the GatewayClass along with the ControllerName to be used.
+The Operator will still auto-create the GatewayClass/Gateway which will allow the operator to still maintain the proper listeners matching the Gateway Contour is processing, but allow for custom implementations without the need for code changes to Contour.
+
+![alt_text](images/gatewayapi-external-gateway.jpg)
+
+### WorkFlows
```
skriss, line 293: One thing that's not described here is upgrades. Presumably the flow is - upgrade the Contour deployment, which then notices that its corresponding Envoy is out of date, and updates it? Also, do we care about downgrades / if so, do they work the same way?

If so, worth noting that this now enforces that Contour always gets upgraded before Envoy (which has been my preferred approach to upgrades, but now it's a hard requirement). That means new versions of Contour shouldn't start using new Envoy features on the critical path until they've been out in Envoy for at least a release.

design/gateway-apis-implementation.md
```diff
@@ -1,204 +1,420 @@
...
+
+The operator is responsible for managing instances of Contour as well as any additional configuration items that are related to running Contour or Envoy.
+These pieces will be configured in the Contour Configuration CRD.
+
+### Configmap to CRD
+
+Currently, Contour gets its configuration from two different places, one is the configuration file represented as a Kubernetes configmap.
+The other are flags which are passed to Contour.
+
+This design will migrate the Contour configuration configmap to a full custom resource definition which will allow for better validation, easier processing, as well as represent a full Contour configuration including how Envoy is to be provisioned if the user desires.
```
skriss, line 173: Are we planning to continue supporting the config file as well until Contour 2.0? Otherwise seems like this would constitute a breaking change.

design/gateway-apis-implementation.md
```diff
@@ -1,204 +1,420 @@
...
+
+#### Benefits
+
+* Eliminates the need to translate from a CRD to a configmap (Like the Operator does today)
+* Allows for a place to surface information about configuration errors - the CRD status, in addition to the Contour log files
+* Allows the Operator workflow to match a non-operator workflow  (i.e. you start with a Contour configuration CRD)
```
skriss, line 179: What exactly do you mean by this? Are you implying that [the Operator's Contour CRD](https://github.com/projectcontour/contour-operator/blob/main/api/v1alpha1/contour_types.go#L38-L50) goes away and is replaced by the new Contour config CRD? If so, I'm not clear where an operator user would configure things like the namespace to install Contour in, or the # of replicas in the Contour deployment, etc. But if that's not what you meant, disregard.

skriss (commented):
I'm broadly in sync with what's proposed here, still have a few questions though.

design/gateway-apis-implementation.md
stevesloka: I think this would only support the crd. So we'd deprecate support for managed bits using a configmap.

design/gateway-apis-implementation.md
stevesloka: Feels like a missing bit in the gateway design since an ingress controller doesn't really manage the ports, the external gateway does. 

An ingress gateway us just a cluster service at this point.

stevesloka: I'll update the design. Can start with type loabalancer, but Im going to add a spot to reference a gateway instead. We need to do that, seems silly to not enable gateways when implementing gateways. 🙃

design/gateway-apis-implementation.md
stevesloka: Num of replicas, etc is a good point. Originally I had wanted to have a different card for the operator vs contour, but didn't get much support for the idea. 

We could do something like that but make the config crd a reference or something? Worth a discussion on. 

Or those fields are just ignored if not using the operator which feels bad as well.

design/gateway-apis-implementation.md
youngnick: I agree, I think the stance here is:
- Contour proper continues supporting the config file until we release a 2.0 (which may be after October), but it's deprecated.
- The fully-managed, operator flow requires the CRD before then.

design/gateway-apis-implementation.md
youngnick: I think we're probably going to end up with two CRDs: the operator Contour/ContourDeployment CRD that describes how Contour should be deployed. This should reference a ContourConfig CRD, or reuse the same structs as the ContourConfig CRD, which the controller will watch and handle.

youngnick: Actually, now I've thought about it a little more, I think that using the ContourConfig Spec struct in the Operator's ContourDeployment object is the way to go, so you would end up with something like:
```yaml
kind: ContourDeployment
spec:
  deployment: <replicas and other Contour controller deployment settings>
  config:
    <include ContourConfigSpec struct here>
```

Doing this means that the Operator can simply take whatever is in the `config` stanza, and turn it into the Spec for a ContourConfig object, and that updating ContourConfig automatically updates the ContourDeployment's fields.

skriss: Yeah, seems to me like we'll need to have 2 CRDs as well. Will think a little more about embedding the Spec struct vs. using a reference but we can definitely make one of those work.

stevesloka: I'd prefer to not have the reference, having to manage 100 different objects is way too complex for what we're trying to do. That was the idea behind having HTTPProxy be able to work as a single resource, or many. 

To make it simpler, we could put all the CRDs in the Contour repo, and the operator could just reference those CRDs, so it's just a consumer of the `ContourDeployment` (or whatever name it gets).

design/gateway-apis-implementation.md
stevesloka: Yup I agree with what you mention @skriss. I can update the design to call that out specifically. 

I think there are also ways that we can support new Envoy features, but pick apart the Envoy version connecting. We'd need to dig into that more, but would be a neat think to at least think about for future work. I know that @sunjayBhatia had some ideas around this as well.

design/gateway-apis-implementation.md
sunjayBhatia: would be good to enumerate what the operator only information is like mentioned above (# of replicas etc.) and see what we need as fields and what we can achieve by convention (eg. namespace of created Contour could be the same the Config CRD is in etc.)

design/gateway-apis-implementation.md
```diff
@@ -1,204 +1,420 @@
...
+
+**Assertion**:_ A Contour Configuration CRD is required for Contour to function in all scenarios._
+
+**Assertion**:_ The GatewayClass.Spec.ParamsRef will not be defined since all the work to define how a Contour/Envoy is configured is defined in the Contour configuration CRD._
+
+#### Operator with GatewayAPI (Envoy is managed by convention)
+
+In this mode, the Operator is managing instances of Contour. Each instance of Contour manages its own Envoy fleet by convention (since the Operator is in use). Contour will expose the Envoy fleet by managing a Gateway/GatewayClass/TCPRoute to get traffic to the Envoy instances. The Contour configuration CRD will define how that traffic will get routed or if another Gateway should be managing the traffic (i.e. custom).
+
+1. Deploy Operator to Cluster
```
sunjayBhatia, line 309: nit: I think some of the markdown rendering on the lists in this and following sections might need a look just b/c of indenting etc. to make sure everything is how you expect it

sunjayBhatia (approved):
just a couple open threads rn that others have started but the proposed changes and discussion topics open look good to me! i'll put down my approval with some nits and assuming other comments get resolved

design/gateway-apis-implementation.md
youngnick: Agreed on that last point, @sunjayBhatia, and agreed as well that putting all the CRDs in one place makes sense (probably the Contour repo, but at some point it may be worth extracting them out into a separate repo).

youngnick: The reason I suggested embedding also is because of things like Deployment and ReplicaSet already using that strategy.

design/gateway-apis-implementation.md
stevesloka: 👍

design/gateway-apis-implementation.md
stevesloka: Added a note about upgrades calling this out.

design/gateway-apis-implementation.md
stevesloka: Yeah I'll clean that up. I used a plugin to move from gDocs to markdown and it didn't do a great job in some places.

design/gateway-apis-implementation.md
```diff
@@ -280,12 +280,10 @@ Contour’s implementation of the Operator has taken a similar approach where ba
 
 Initially Contour can do the same work to move that same logic and expose Envoy via annotations on the Service Type=LoadBalancer.
 
-The future looking way to approach this is to utilize GatewayAPI fully and instead of exposing a service with type=LoadBalancer, Contour can create a GatewayClass, Gateway which selects a TCPRoute used to send traffic to the Envoy fleet.
-The details defined in the Contour configuration CRD will be used to pre-populate annotations that will be stored in a Kubernetes configmap and referenced from the GatewayClass.Spec.ParametersRef.
-The Gateway Stub controller will be utilized to configure a new Kubernetes Service with the annotations defined in the configmap on the GatewayClass and send traffic to Envoy.
-Once implementers catch up, specific implementations will be utilized and the stub controller can be deprecated.
+Additionally, Contour will support utilizing GatewayAPI itself fully and instead of exposing a service with type=LoadBalancer, Contour can create a GatewayClass, Gateway which selects a TCPRoute used to send traffic to the Envoy fleet.
```
youngnick, line 283: Nit: "Eventually" is probably closer to what we mean here, since we're not going to do this right away, I think.

design/gateway-apis-implementation.md
```diff
@@ -535,14 +540,4 @@ Features that won’t be tackled with this design, but could be extended in futu
...
-Pasted from Slack, proposed:
-
-* First thing we will do is merge in Steve S and Sunjay's fixes to the operator/Contour interaction for 1.18 release next week.
-* After 1.18 we will open for discussion that we move to quarterly releases starting end October. If the community agrees, we will release between 1.18 and October when the following phases are done.
-* Next, we will work on the config file to CRD change - this has operator changes as well. Release when this one is done.
-* Next, we will bring the Service management code from the operator into Contour. Release when this is done.
-* Then, we evaluate if we change Service management code to Gateway management code before October or wait until Gateway API is beta.
-
-This implementation plan is tricky because it’s interacting with the changing of the release cadence.
+=
```
youngnick, line 543: Nit: Not required?

youngnick (approved):
Yeah, this looks great, thanks. Just a small suggestion about the language to make it clear that Gateway config and Gateway -> LB Service shim is in the future, not right away.

design/gateway-apis-implementation.md
stevesloka: Yup, it's just down to timing I think. I'd like to have the config option in the configuration file. If that's set then Contour doesn't manage the envoy service of type=LoadBalancer, just a regular ClusterIP service.

skriss (approved):
:shipit:

PR #3801, opened 2021-06-11
internal/dag: Implement TLSRoute mode:terminate

stevesloka:
Implements support for GatewayAPI TLSRoute mode: terminate which terminates TLS
at the Gateway.

Updates #3440


internal/dag/gatewayapi_processor.go
```diff
@@ -112,6 +112,14 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
 						}
 					}
 				}
+			} else {
```
sunjayBhatia, line 115: so this case now means a TLSRoute with no TLS field set defaults to terminate, though terminate without a secret doesn't really work? so can this just be an error now if you don't set TLS at all?

stevesloka: /shrug The spec says that "Terminate" is the default.

stevesloka: - https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha2/gateway_types.go#L321
- https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha2/gateway_types.go#L329-L330

sunjayBhatia: is this intended to be the code for if the mode is nil? rather than the whole TLS block is nil?

sounds like the listener should be rejected if the TLS field is not set: https://github.com/kubernetes-sigs/gateway-api/blob/17a98f6f5aaa1e7fa607c1fc974efeb915b3337c/apis/v1alpha2/gateway_types.go#L184-L197

stevesloka: Essentially it is since the cert validation then fails on L118. Part of the problem with our unit tests is that we don't have the kubebuilder defaults applied. So we can create a test that doesn't specify the cert, but would get the mode applied by the api server and wouldn't pass the unit test.

internal/dag/gatewayapi_processor.go
```diff
@@ -112,6 +112,14 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+				// Config is nil so validate the default which is mode: terminate
+				// Check for TLS on the Gateway.
+				if listenerSecret = p.validGatewayTLS(listener); listenerSecret == nil {
+					// If TLS was configured on the Listener, but it's invalid, don't allow any
+					// routes to be bound to this listener since it can't serve TLS traffic.
+					continue
+				}
```
skriss, line 122: Related to @sunjayBhatia's comment, I'd do this a little differently for clarity, something like:

```go
case gatewayapi_v1alpha1.TLSProtocolType:
    if listener.TLS == nil {
        // log error and continue since .TLS is required if protocol is "TLS"
    }

    // if we get here we know listener.TLS is non-nil
    // proceed to checking tlsConfig.Mode
    if listener.TLS.Mode != nil {
        // ...
    }
```

So basically put the nil-check at the top, and no need to call into `p.validGatewayTLS(...)` if we know `listener.TLS` is nil

skriss: Should be similar to how the [`HTTPSProtocolType` block](https://github.com/projectcontour/contour/blob/436f5dcedf980d60fe16ba0bbda6ba2701059690/internal/dag/gatewayapi_processor.go#L84-L95) is structured

stevesloka: 👍

internal/dag/gatewayapi_processor.go
stevesloka: Ok I rereead the spec and at the top level it does say "p.Errorf("Listener.TLS is required when protocol is %q.", listener.Protocol)", so that make sense. I've updated.

internal/dag/gatewayapi_processor.go
```diff
@@ -95,30 +95,26 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+				case gatewayapi_v1alpha1.TLSModeTerminate:
+					// Check for TLS on the Gateway.
+					if listenerSecret = p.validGatewayTLS(listener); listenerSecret == nil {
+						// If TLS was configured on the Listener, but it's invalid, don't allow any
+						// routes to be bound to this listener since it can't serve TLS traffic.
+						continue
+					}
+				case gatewayapi_v1alpha1.TLSModePassthrough:
+					if listener.TLS.CertificateRef != nil {
+						p.Errorf("Listener.TLS cannot be defined when TLS Mode is %q.", *listener.TLS.Mode)
```
skriss, line 115: "Listener.TLS cannot..." -> "Listener.TLS.CertificateRef cannot..."

skriss (commented):
one tiny nit, looking through tests next.

internal/dag/gatewayapi_processor.go
stevesloka: 👍

test/e2e/gateway/008_tlsroute_test.go
```diff
@@ -100,3 +100,71 @@ func testTLSRoutePassthrough(fx *e2e.Framework, namespace string) {
...
+
+	// Ensure request routes to echo.
+	res, ok := fx.HTTP.SecureRequestUntil(&e2e.HTTPSRequestOpts{
+		Host:      "tlsroute.gatewayapi.projectcontour.io",
+		Condition: e2e.HasStatusCode(200),
+	})
+	assert.Truef(t, ok, "expected 200 response code, got %d", res.StatusCode)
+	assert.Equal(t, "echo", fx.GetEchoResponseBody(res.Body).Service)
+
+	require.NoError(t, retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
skriss, line 144: a quick comment indicating what's happening here would be helpful (looks like the SNI match is being dropped and the routing is expected to still work).

skriss (commented):
One more small comment on the tests, otherwise looks good. Some of the tests are pretty tough to review but I skimmed and it looks reasonable/reasonable coverage.

test/e2e/gateway/008_tlsroute_test.go
stevesloka: I wonder if it's still working or the SNI is added automatically. I can update with the SNI bit (now that I see it's in the test suite) since that's the expected behavior.

stevesloka: ok updated with SNI

skriss: Sorry, I was talking about https://github.com/projectcontour/contour/blob/f53e69113238ab1a6714a7ba507f575ee66f1fca/test/e2e/gateway/008_tlsroute_test.go#L144-L161 (GH doesn't always make it clear what line you commented on) -- that updates the TLSRouteRule to *not* have the `Matches` block (https://github.com/projectcontour/contour/blob/f53e69113238ab1a6714a7ba507f575ee66f1fca/test/e2e/gateway/008_tlsroute_test.go#L122-L126), only to have the `ForwardTo`. So seems like that's verifying that if you don't explicitly specify a match on SNI, then any SNI will still route traffic there.

stevesloka: Ahh ok I'm following now. I updated some comments. Also added a negative test to validate a non-matching SNI doesn't produce a response.

skriss: :+1: nice

skriss (approved):
no further comments, LGTM pending CI.

skriss:
This one needs a rebase now that the E2E changes have gone in

stevesloka:
all rebased @skriss @sunjayBhatia, please take a look!

skriss (approved):
LGTM

sunjayBhatia (approved):

skriss:
gonna merge, can fix the nit separately

PR #3803, opened 2021-06-11
test/e2e: tests use Contour running locally

sunjayBhatia:
- Adds unmarshaling etc. for ratelimit resources
- Adds convenience method for deploying resources needed for local
    contour
- e2e Framework can be requested in incluster or not format
  - will expect correct variables to be set
- namespaced test helper can be used to automatically manage namespace
    for an individual test
- all e2e tests updated to use namespaced test helper and other wrappers
    to make tests easier to duplicate with similar config
  - tests modified to be contained in a namespace as needed
  - gateway tests modified to use gateway in unique namespace for test

Updates: #3772

Issue #3545, opened 2021-04-06
Contour Managed Envoy


stevesloka:
As part of the GatewayAPI implementation we've been discussing how to integrate Contour<>Envoy when utilizing GatewayAPI.Gateways. I have a small discussion document outlining a bit more background here (https://docs.google.com/document/d/1kOrGqHXULYO6YLxGCOtVyjz_7mfcW7Yi1WXCpQz6_Qo/edit#heading=h.s3afnqyzdsta). 

The goal of this issue is to create a branch in Contour and investigate how this approach might or might not work moving forward with the GatewayAPI work 

Tasks:

- Configuration CRD for GatewayClass
  - networkPublishing: Define how Envoy is exposed from the cluster
  - gatewayClassRef: 
- Controller to manage Envoy instances
- Set Gateway Status

xaleeks:
Thanks for this @stevesloka .  This spike being worked on right now in the  v1.15 right?  Feel free to untag it if not the case.

danehans:
When Contour manages the envoy infra, the operator should be updated to remove envoy infra management.

cc: @Miciah @candita

sunjayBhatia:
Closing as we have the Gateway provisioner which manages Envoy dynamically

PR #3805, opened 2021-06-13
Implements Gateway Controller

danehans:
Adds an initial Gateway controller implementation.
- `cmd/contour/serve.go`: Updates `NewGatewayController()` to accept a gatewayclass controller string. `ControllerName` is now a required param, so the nil check is removed.
- `examples/*`: Updates examples for Gateway support. Note that gateway/gatewayclass RBAC added the 'update' verb to support adding/removing the finalizer.
- `internal/controller/gateway.go`: Adds a controller to reconcile Gateway objects owned by Contour. The controller validates, finalizes the referenced gatewayclass (if needed), and updates the status of managed `Gateway` objects.
- `internal/gateway/*`: Adds functions for managing gateway objects.
- `internal/k8s/informers.go`: Updates RBAC for managing status of Gateway objects.
- `internal/slice/*`: Introduces a slice package for adding/removing a string from a []string.
- `internal/status/*`: Adds support for managing Gateway conditions.
- `internal/validation/*`: Adds initial support for validating Gateway objects.
- `internal/validation/gatewayclass_test.go` Changes the scope of existing variables to support Gateway tests.
- `pkg/config/parameters_test.go` Updates `controllerName` from optional to required.
- `test/scripts/install-contour-working.sh`: Adds the controller name of the test gatewayclass to the contour config.


internal/slice/slice.go
```diff
@@ -0,0 +1,42 @@
...
+	if len(newSlice) == 0 {
+		// Sanitize for unit tests so we don't need to distinguish empty array
+		// and nil.
+		newSlice = nil
+	}
+	return newSlice
+}
+
+// ContainsString checks if a given slice of strings contains the provided string.
+func ContainsString(slice []string, s string) bool {
```
sunjayBhatia, line 35: worth using the existing utils from here?

internal/slice/slice_test.go
```diff
@@ -0,0 +1,67 @@
...
+// limitations under the License.
+
+package slice
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestRemoveString(t *testing.T) {
```
sunjayBhatia, line 21: if we keep the `slice` package would be good to have a test for the other utility as well

internal/validation/gateway.go
```diff
@@ -0,0 +1,66 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gatewayapi_v1alpha1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
+)
+
+// ValidateGateway validates gw according to the Gateway API specification.
+// For additional details of the Gateway spec, refer to:
+//   https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.Gateway
+func ValidateGateway(ctx context.Context, cli client.Client, gw *gatewayapi_v1alpha1.Gateway) field.ErrorList {
```
sunjayBhatia, line 30: we should probably take stock of what is validated here and what is validated in the dag processor and consolidate

internal/slice/slice.go
danehans: That pkg is not meant to be a library and can not be imported, [xref](https://github.com/kubernetes/kubernetes/issues/90358).

internal/slice/slice_test.go
danehans: ack

internal/validation/gateway.go
danehans: @sunjayBhatia the plan is to move the dag processor validation to the [validation](https://github.com/projectcontour/contour/tree/v1.16.0/internal/validation) package and then migrate to upstream. To make this PR reviewable, I made the gateway validation as small as possible. I created https://github.com/projectcontour/contour/issues/3807 to detail migrating the validation logic out of the dag processor.

danehans:
https://github.com/projectcontour/contour/issues/3812 created to track the need for gateway controller integration tests.

internal/controller/gateway.go
```diff
@@ -31,46 +37,201 @@ import (
...
+		}
 		return reconcile.Result{}, nil
 	}
 
-	// Pass the new changed object off to the eventHandler.
-	r.eventHandler.OnAdd(gateway)
+	// The gateway is safe to process, so check if it's valid.
+	errs := validation.ValidateGateway(ctx, r.client, gw)
+	if errs != nil {
+		r.log.WithField("name", request.Name).WithField("namespace", request.Namespace).
```
youngnick, line 151: We should also be setting the `Scheduled` Condition to `status: false` here, with a `NotReconciled` Reason.

internal/controller/gateway.go
```diff
@@ -31,46 +37,201 @@ import (
...
+			Error("invalid gateway: ", int_errors.ParseFieldErrors(errs))
+	} else {
+		// TODO: Finalize the gateway when managed infrastructure is added.
```
youngnick, line 154: We should set `Scheduled` to `status: true` at this point, with a Reason of `Scheduled`, I think? Reference is https://github.com/kubernetes-sigs/gateway-api/blob/16876c7362d58cae211e5599a6ef939ed1b859e9/apis/v1alpha2/gateway_types.go#L574-L614

internal/controller/gateway.go
```diff
@@ -31,46 +37,201 @@ import (
...
+					r.referencedClass.Name, err)
+			}
+		}
+		// Pass the gateway off to the eventHandler.
+		r.eventHandler.OnAdd(gw)
+
+		// TODO: Ensure the gateway by creating manage infrastructure, i.e. the Envoy service.
+	}
+
+	if err := status.SyncGateway(ctx, r.client, gw, errs); err != nil {
```
youngnick, line 179: Is this status Sync pattern common for controller-runtime controllers? In any case, we need to ensure that the `Scheduled` Condition is populated correctly (it signals that Contour has accepted the Gateway for processing and, eventually, started the process of creating infrastructure). We should add `Ready` with `status: false` when we add `scheduled`.

Then, `Ready` can be set to `status: true` when the infrastructure is provisioned.

I think we may need to change `SyncGateway()` to update `Scheduled` first maybe?

internal/status/gateway.go
```diff
@@ -0,0 +1,43 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gateway_v1alpha1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
+)
+
+// SyncGateway computes the current status of gw and updates status based on
+// any changes since last sync.
+func SyncGateway(ctx context.Context, cli client.Client, gw *gateway_v1alpha1.Gateway, errs field.ErrorList) error {
+	updated := gw.DeepCopy()
+
+	// Gateway's contain a default status condition that must be removed when reconciled by a controller.
```
youngnick, line 31: I'm pretty sure this is incorrect, it should be set to `status: true` when reconciled by a controller.

youngnick (changes requested):
This is mostly looking pretty good, but I think that we have to use the `Scheduled` Condition, not just remove it. `Scheduled` means that the resource is valid, has been accepted for processing, and infrastructure is being created. `Ready` means that infrastructure is created and ready to route traffic.

Obviously, until we are handling Envoy management, we can't guarantee `Ready`, but we should still be updating `Scheduled`. I guess the upstream docs could use some improvement to make this flow clearer?

internal/controller/gateway.go
danehans: The Gateway has been reconciled, but no infra is being provisioned since it's invalid. I will set `Scheduled=false` and use an `InvalidGateway` reason with the message indicating why the Gateway is invalid.

internal/validation/gateway.go
```diff
@@ -0,0 +1,89 @@
...
+	errs = append(errs, validateGatewaySpec(ctx, cli, gw, field.NewPath("spec"))...)
+
+	if len(errs) > 0 {
+		return errs
+	}
+	return nil
+}
+
+// validateGatewayNamespace validates whether other gateways exist in the namespace of gw.
+func validateGatewayNamespace(ctx context.Context, cli client.Client, gw *gatewayapi_v1alpha1.Gateway, path *field.Path) field.ErrorList {
```
stevesloka, line 48: Right now all validation is done in the DAG builder processor. How do we manage validation here on top of validation there? There are a bunch of bits that could be pulled out and pushed upstream, however, other things, like specific configuration items can't have generic validation and need to set status.

internal/controller/gateway.go
danehans: Currently `status.SyncGateway()` sets `Ready=true`, but I'm updating that to `Scheduled=true` since the controller is not provisioning Envoy infra yet. I have also updated `ensureClassFinalizer()` to `addClassFinalizer()` so the func does not update the object. The object's status will now be sync'ed before being updated on the API server.

stevesloka (commented):
How do we manage setting status from the result of the DAG build vs the controller? Seems like we need to move the status setting to the DAG processor like it is today, and leave the gateway controller to manage the Envoy instances (if desired).

internal/controller/gateway.go
danehans: > Is this status Sync pattern common for controller-runtime controllers?

It's a pattern I have seen with other controllers, essentially reconcile the object and calculate status before updating the object on the API server.

> I think we may need to change SyncGateway() to update Scheduled first maybe?

Ack. I was still mentally operating in the operator world where the infra is provisioned. I have updated `computeGatewayReadyCondition()` to `computeGatewayScheduledCondition()` to manage the "Scheduled" condition type. When Envoy infra management is added, we can re-introduce `computeGatewayReadyCondition()` to calculate the "Ready" condition type.

danehans:
> How do we manage setting status from the result of the DAG build vs the controller? Seems like we need to move the status setting to the DAG processor like it is today, and leave the gateway controller to manage the Envoy instances (if desired).

@stevesloka I thought we discussed moving away from the DAG and start utilizing the runtime controllers cache? @skriss did you mention that you were going to explore this in more detail?

internal/status/gateway.go
danehans: Ack

stevesloka:
>  I thought we discussed moving away from the DAG and start utilizing the runtime controllers cache?

The current DAG build is critical to how Contour works. Moving away from that is a core fundamental change to Contour. It's doable but needs a ton of work + overhaul to the core internals.

internal/validation/gateway.go
danehans: IMO validation is one of the first steps that a controller should perform on the object it reconciles. If utilizing the controller-runtime caches is the long-term direction, I think it's best to not add any more functionality to the DAG. As-is, the DAG will continue to perform validation, status mgt, etc. for xRoutes and Gateway listeners. The controller-runtime controllers will do the same for the objects they reconcile. When the xRoute controllers are implemented, we migrate the validation, status mgt, etc. from the Gateway API DAG processor.

stevesloka: I wouldn't go down that road right now, I'd let the DAG builder validate its stuff and keep everything together.

To me, this new controller should just manage Envoy instances, everything else should stay consistent as-is.

stevesloka: But don't let me distract, if other's don't agree then go with that. =)

internal/controller/gateway.go
stevesloka: How does it work if not provisioning infra? Need to keep that use-case in mind.

youngnick: If not provisioning infra, `Scheduled` is roughly equivalent to the `Valid` condition we use on HTTPProxy, it indicates that whether the object is syntactically valid. Should this be more than one Condition? Possibly, but I wasn't able to make that argument convincingly enough upstream.

internal/validation/gateway.go
youngnick: I think this is a good chance to start pulling functionality out of the DAG, and decomposing it a bit. I think of Contour as having three stages: Ingestion of objects into the DAG, the DAG run, and then export to Envoy. I think that it's okay to have some validation at the ingestion stage, as long as we keep in mind what that means for the DAG. It's just a data model, and if we can split complexity out of it, it will be easier to code and maintain.

youngnick:
I agree that the DAG needs to stay, but I think that filtering objects before they go in is pretty reasonable (which is what pre-DAG validation logic will do).

internal/validation/gateway.go
stevesloka: We'll need to either process a full dag to find some of the validation errors. One simple one is a missing/invalid service reference should set status on the Gateway. Probably for now we'll need to have the Gateway Controller get feedback from the result of a dag run and see if errors exist that it cares about to set proper status.

stevesloka:
> prefiltering

We also need to be careful to make sure if a cluster is running and something causes the Gateway to become invalid for one specific reason (say an invalid service), that we still keep the Gateway in the Contour cache so that we don't break all other routes that would be valid.

internal/status/cache.go
```diff
@@ -35,10 +35,11 @@ const ValidCondition ConditionType = "Valid"
 // NewCache creates a new Cache for holding status updates.
 func NewCache(gateway types.NamespacedName) Cache {
 	return Cache{
-		proxyUpdates: make(map[types.NamespacedName]*ProxyUpdate),
-		gatewayRef:   gateway,
-		routeUpdates: make(map[types.NamespacedName]*ConditionsUpdate),
-		entries:      make(map[string]map[types.NamespacedName]CacheEntry),
+		proxyUpdates:   make(map[types.NamespacedName]*ProxyUpdate),
+		gatewayRef:     gateway,
+		gatewayUpdates: make(map[types.NamespacedName]*GatewayConditionsUpdate),
```
stevesloka, line 40: Shouldn't need this, I think we can use the ConditionsUpdate as-is.

internal/status/cache.go
```diff
@@ -53,8 +54,9 @@ type CacheEntry interface {
 type Cache struct {
 	proxyUpdates map[types.NamespacedName]*ProxyUpdate
 
-	gatewayRef   types.NamespacedName
-	routeUpdates map[types.NamespacedName]*ConditionsUpdate
+	gatewayRef     types.NamespacedName
+	gatewayUpdates map[types.NamespacedName]*GatewayConditionsUpdate
```
stevesloka, line 58: remove this since it should already work.

internal/status/cache.go
```diff
@@ -53,8 +54,9 @@ type CacheEntry interface {
...
+	routeUpdates   map[types.NamespacedName]*RouteConditionsUpdate
```
stevesloka, line 59: I'd rename this back to `ConditionsUpdate` and name the var the same. Originally it was HTTPRoute, then we added TLSRoute, so it just became `Route`, but could change to plain conditions since that's what its processing.

stevesloka (changes requested):
Here's a first pass, does seem like you're still working on it @danehans. =)

cmd/contour/serve.go
```diff
@@ -425,16 +425,15 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 			}
 
 			// Create and register the gatewayclass controller with the manager.
-			if ctx.Config.GatewayConfig.ControllerName != nil {
-				gcController := *ctx.Config.GatewayConfig.ControllerName
-				if _, err := controller.NewGatewayClassController(mgr, &dynamicHandler,
-					log.WithField("context", "gatewayclass-controller"), gcController); err != nil {
-					log.WithError(err).Fatal("failed to create gatewayclass-controller")
-				}
+			gcController := *ctx.Config.GatewayConfig.ControllerName
```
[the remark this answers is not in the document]
danehans: The code is only executed if `ctx.Config.GatewayConfig != nil`. I thought I updated `ControllerName` to be a required field of a `GatewayConfig`, but I don't see it in the PR. I'll update now.

examples/contour/02-role-contour.yaml
```diff
@@ -98,12 +96,23 @@ rules:
...
   - update
+- apiGroups:
+  - networking.x-k8s.io
+  resources:
+  - gatewayclasses
+  - gateways
+  verbs:
+  - get
+  - list
+  - update
```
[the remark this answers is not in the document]
danehans: The update verb is required since the controller will apply a finalizer to gatewayclass and gateway objects.

examples/gateway/02-gateway.yaml
```diff
@@ -0,0 +1,16 @@
+---
+kind: Gateway
+apiVersion: networking.x-k8s.io/v1alpha1
+metadata:
+  name: contour
+  namespace: projectcontour
+spec:
+  gatewayClassName: example
```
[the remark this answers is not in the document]
danehans: `gatewayClassName` is a reference to a `gatewayclass.metadata.name`. The current example gatewayclass is named "example".  The `spec.controller` of the referenced gatewayclass must match what contour expects as described by the contour config `gateway.controllerName` parameter (currently `projectcontour.io/projectcontour/contour`).

examples/contour/02-role-contour.yaml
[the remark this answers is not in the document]
stevesloka: Ahh that's right, forgot about the finalizer.

examples/gateway/02-gateway.yaml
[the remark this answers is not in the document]
stevesloka: Ohh it is ClassName, I was thinking all along that was ControllerName. Why then do we care about the controller name field? Can we just reference the class by name?

internal/controller/gateway.go
```diff
@@ -30,47 +32,269 @@ import (
 	gatewayapi_v1alpha1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
 )
 
+const gatewayFinalizer = "gateway.projectcontour.io"
```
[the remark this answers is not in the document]
danehans: I typically don't see finalizers used to express object ownership. Either of [these approaches](https://sdk.operatorframework.io/docs/building-operators/ansible/reference/retroactively-owned-resources/) are used to solve ownership references. 

xref the operator's transition from labels to ownerRef: https://github.com/projectcontour/contour-operator/issues/383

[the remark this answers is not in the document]
danehans: > I see that the delete finalizer has a reference to this:https://github.com/projectcontour/contour/pull/3805/files#diff-2e576c5279d63e220a680a1643f6cef655d451271714eed9405d4f0203d4704aR226

The code you reference, `othersRefClass()`, ensures the controller only removes the finalizer from the referenced gatewayclass if no other gateways reference it.

internal/dag/gatewayapi_processor.go
```diff
@@ -46,6 +47,10 @@ type GatewayAPIProcessor struct {
 
 	dag    *DAG
```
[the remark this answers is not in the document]
danehans: @stevesloka not sure what your comment means, can you elaborate?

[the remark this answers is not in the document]
stevesloka: huh? Must be this silly M1 Mac rendering fonts incorrectly. It showed `day *DAY` on my screen. 🤯

internal/controller/gateway.go
```diff
@@ -30,47 +32,269 @@ import (
...
+		}
+		// Error reading the object, so requeue the request.
+		return reconcile.Result{}, fmt.Errorf("failed to get gateway %s/%s: %w", request.Namespace, request.Name, err)
+	}
+
+	// Check if object is deleted.
+	if !gw.ObjectMeta.DeletionTimestamp.IsZero() {
+		r.log.WithField("name", request.Name).WithField("namespace", request.Namespace).Info("zero timestamp")
+
+		if err := r.ensureGatewayDeleted(gw); err != nil {
```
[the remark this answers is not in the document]
stevesloka: This must be my M1 mac rendering fonts weirdly. Looks ok on an intel box.

internal/controller/gateway.go
```diff
@@ -30,47 +32,269 @@ import (
...
+	// TODO: Ensure the gateway by creating manage infrastructure, i.e. the Envoy service.
+
+	// Pass the gateway off to the eventHandler.
+	r.eventHandler.OnAdd(gw)
 
 	return reconcile.Result{}, nil
 }
+
+// isFinalized returns true if gw is finalized.
+func (r *gatewayReconciler) isFinalized(gw *gatewayapi_v1alpha1.Gateway) bool {
```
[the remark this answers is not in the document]
stevesloka: This must be my M1 mac rendering fonts weirdly. Looks ok on an intel box.

internal/dag/gatewayapi_processor.go
```diff
@@ -59,36 +64,115 @@ type matchConditions struct {
...
 		return
 	}
 
-	for _, listener := range p.source.gateway.Spec.Listeners {
+	// Validate the Gateway
+	errs := p.validateGateway()
+
+	// Process all the HTTPRoutes that match this Gateway.
+	listeners := p.source.gateway.Spec.Listeners
+	for i, listener := range listeners {
```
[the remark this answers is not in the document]
danehans: @stevesloka thanks for helping me understand this issue. What are your thoughts on GatewayAPIProcessor containing a field that represents valid listener information? GatewayAPIProcessor will validate a Gateway, including listeners, set this field with valid listener information, and only process valid listeners. For example:
```
type GatewayAPIProcessor struct {
	logrus.FieldLogger

	dag    *DAG
	source *KubernetesCache

	gatewayListeners map[types.NamespacedName][]gatewayListener
}

type gatewayListener struct {
	matchingHTTPRoutes []*gatewayapi_v1alpha1.HTTPRoute
	matchingTLSRoutes  []*gatewayapi_v1alpha1.TLSRoute
	secret             *Secret
}

File: internal/dag/gatewayapi_processor.go [hunk 20]
```diff
@@ -59,36 +64,115 @@ type matchConditions struct {
...
+	gcName := gw.Spec.GatewayClassName
+	gc := p.source.gatewayclass
+	if gc == nil {
+		errs = append(errs, field.InternalError(path.Child("gatewayClassName"),
+			fmt.Errorf("referenced gatewayclass %q doesn't exist in DAG", gcName)))
+		// return early since additional validation checks require the gatewayclass.
+		return errs
+	}
+
+	if gcName != gc.Name {
```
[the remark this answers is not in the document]
danehans: Ack.

internal/dag/gatewayapi_processor.go
```diff
@@ -59,36 +64,115 @@ type matchConditions struct {
...
+		errs = append(errs, field.InternalError(path.Child("gatewayClassName"),
+			fmt.Errorf("referenced gatewayclass %q differs from DAG", gcName)))
+		// return early since additional validation checks require the gatewayclass.
+		return errs
+	}
+
+	// See if the referenced gatewayclass is admitted.
+	gcAdmitted := false
```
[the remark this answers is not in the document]
danehans: @stevesloka the gateway controller does not check whether the referenced gc is admitted or not, just that it contains the controller string that matches Contour's config. We can implement this check in the event handler, but then we have no way to reflect this condition in status and will be required to look at logs to troubleshoot. Thoughts?

internal/dag/gatewayapi_processor.go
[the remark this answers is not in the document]
stevesloka: Yeah I think that could work. Or just add a gateway status bit to the existing code and replace the "fmt.Error" lines with status conditions back to the gateway. So as we process listeners and find issues, you set the status on them.

internal/status/cache.go
danehans: @stevesloka the existing `ConditionsUpdate()` takes a `RouteGatewayStatus` which differs from a `GatewayStatus` type. The new `GatewayConditionsAccessor()` is called by the GatewayAPIProcessor to calculate the gateway status conditions.

internal/status/cache.go
danehans: Same response as https://github.com/projectcontour/contour/pull/3805/files#r655899998.

internal/status/cache.go
danehans: @stevesloka we need to have separate types for updating route and gateway conditions. With that said, keeping the `ConditionsUpdate` name makes it difficult to differentiate between a gateway and a route condition update.

examples/gateway/02-gateway.yaml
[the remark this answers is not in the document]
danehans: We care about the controller field so Contour knows which gatewayclasses to process. If a gateway references a gatewayclass by name, the gateway controller will get that gatewayclass and check that the controller field matches what's configured. If it does not match the configured controllerName, then the gateway should be considered invalid.

internal/dag/gatewayapi_processor.go
```diff
@@ -445,9 +529,24 @@ func selectorMatches(selector *metav1.LabelSelector, objLabels map[string]string
...
+		gwAccessor.AddCondition(gatewayapi_v1alpha1.GatewayConditionScheduled, metav1.ConditionTrue, status.ReasonValidGateway, "Valid Gateway")
+	default:
+		gwAccessor.AddCondition(gatewayapi_v1alpha1.GatewayConditionScheduled, metav1.ConditionFalse, status.ReasonInvalidGateway, errors.ParseFieldErrors(errs))
+	}
+}
+
 func (p *GatewayAPIProcessor) computeTLSRoute(route *gatewayapi_v1alpha1.TLSRoute) {
 
-	routeAccessor, commit := p.dag.StatusCache.ConditionsAccessor(k8s.NamespacedNameOf(route), route.Generation, status.ResourceTLSRoute, route.Status.Gateways)
+	routeAccessor, commit := p.dag.StatusCache.RouteConditionsAccessor(k8s.NamespacedNameOf(route), route.Generation, status.ResourceTLSRoute, route.Status.Gateways)
```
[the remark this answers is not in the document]
danehans: This method takes a `gateways []gatewayapi_v1alpha1.RouteGatewayStatus` type as a parameter which differs from a `GatewayStatus` type used to manage Gateway conditions.

internal/dag/gatewayapi_processor.go
```diff
@@ -59,36 +64,115 @@ type matchConditions struct {
...
 
 	// reset the processor when we're done
 	defer func() {
 		p.dag = nil
 		p.source = nil
 	}()
 
-	// Gateway must be defined for resources to be processed.
 	if p.source.gateway == nil {
-		p.Error("Gateway is not defined!")
```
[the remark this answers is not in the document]
danehans: Ack. I changed error -> info.

internal/controller/gateway.go
```diff
@@ -30,47 +32,269 @@ import (
...
+			r.referencedClass.Name, err)
+	}
+	if !others {
+		if err := r.EnsureClassFinalizerRemoved(); err != nil {
+			return fmt.Errorf("failed to remove gatewayclass %s finalizer: %w", r.referencedClass.Name, err)
+		}
+	}
+
+	// Remove the finalizer from the gateway.
+	if err := r.EnsureFinalizerRemoved(gw); err != nil {
```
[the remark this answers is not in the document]
danehans: I'm not sure if we want to remove the finalizer in this case. It might be safer to keep the finalizer applied so a user is required to investigate further. Thoughts?

internal/controller/gateway.go
```diff
@@ -30,47 +32,269 @@ import (
...
 	return c, nil
 }
 
+// enqueueRequestForOwnedGateway returns an event handler that maps events to
+// Gateway objects that reference a GatewayClass owned by Contour.
+func (r *gatewayReconciler) enqueueRequestForOwnedGateway() handler.EventHandler {
+	return handler.EnqueueRequestsFromMapFunc(func(a client.Object) []reconcile.Request {
+		gw, ok := a.(*gatewayapi_v1alpha1.Gateway)
+		if !ok {
+			r.log.WithField("name", a.GetName()).WithField("namespace", a.GetNamespace()).
```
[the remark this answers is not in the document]
danehans: L75 logs at info level.

internal/status/cache.go
youngnick: I think the Route status and Gateway status are different enough that we need separate handlers for them (because Route status is effectively namespaced by the Gateway that the status is relevant to).

internal/controller/gateway.go
```diff
@@ -30,47 +32,269 @@ import (
...
+			Info("gateway not owned by contour, bypassing reconciliation.")
+		return []reconcile.Request{}
+	})
+}
+
+// classForGateway returns an nil, error if the gatewayclass referenced by gw does not exist
+// or nil, nil if gw is not owned by Contour. Otherwise, the referenced gatewayclass is returned.
+func (r *gatewayReconciler) classForGateway(gw *gatewayapi_v1alpha1.Gateway) (*gatewayapi_v1alpha1.GatewayClass, error) {
+	gc := &gatewayapi_v1alpha1.GatewayClass{}
+	if err := r.client.Get(r.ctx, types.NamespacedName{Name: gw.Spec.GatewayClassName}, gc); err != nil {
```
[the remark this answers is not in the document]
danehans: >What if there are two classes with the same controller name?

I believe we discussed this in https://github.com/projectcontour/contour/pull/3659. Contour starts up with the `gateway.className` config. Each instance of contour should use a unique `gateway.className`, otherwise they can fight over gatewayclasses with the same `spec.controller`. If we want to set gateway status for this use case, then we need to remove the `classForGateway()` check from the event handler so the gateway is processed and status can be set.

internal/controller/gateway.go
```diff
@@ -30,47 +32,269 @@ import (
...
+	// If needed, finalize the referenced gatewayclass.
+	if !r.isClassFinalized() {
+		// Before doing anything with the gateway, ensure the reference gatewayclass is finalized.
+		if err := r.ensureClassFinalizer(); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to finalize gatewayclass %s: %w",
+				r.referencedClass.Name, err)
+		}
+	}
+
+	// TODO: Ensure the gateway by creating manage infrastructure, i.e. the Envoy service.
```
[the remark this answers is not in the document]
danehans: Ack
xref: https://github.com/projectcontour/contour/issues/3545

internal/status/cache.go
stevesloka: The only difference is that gateway has a `listenerStatus`. We can separate, just trying to keep duplication to a minimum as a `GatewayStatus.Conditions` is exactly the same as routes.

stevesloka: Maybe we can re-use the `Conditions` struct to eliminate the code duplication. My only worry is that we change something with a route status and that doesn't roll through to Gateway for the core bits that all use `meta/v1.Condition`.

stevesloka: ...but can change it up later if it comes up.

examples/gateway/02-gateway.yaml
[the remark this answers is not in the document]
stevesloka: We just need to understand if Contour will ever process more than one Gateway described in https://github.com/projectcontour/contour/issues/3825. 

If not, then to me it makes more sense to only require the ClassName to be configured in the config file since now we have a specific class to match against.

internal/controller/gateway.go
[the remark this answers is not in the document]
stevesloka: I don't have strong feelings, either way, just hate when clusters get stuck trying to remove stuff when finalizers are in play. We've had lots of similar trouble with cert-manager in the same way.

internal/dag/builder_test.go
```diff
@@ -114,12 +114,31 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 		},
 	}
 
+	gc := &gatewayapi_v1alpha1.GatewayClass{
```
stevesloka, line 117: Do we have any tests validating when the class doesn't match?

internal/dag/gatewayapi_processor.go
```diff
@@ -66,11 +70,34 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
 		p.source = nil
 	}()
 
-	// Gateway must be defined for resources to be processed.
+	// Gateway and GAtewayCLass must be defined for resources to be processed.
```
stevesloka, line 73: nit: s/GAtewayCLass/GatewayClass

(Unless my mac is rendering wrong again)
image

internal/dag/gatewayapi_processor.go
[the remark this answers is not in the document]
stevesloka: Sweet. Should we update the message while we're here? Something like the gateway is not found in the cluster since it would need to be defined already to get to this point.

internal/dag/gatewayapi_processor.go
```diff
@@ -225,14 +252,15 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
 
 		// Process all the HTTPRoutes that match this Gateway.
 		for _, matchingRoute := range matchingHTTPRoutes {
-			p.computeHTTPRoute(matchingRoute, listenerSecret, listener.Hostname)
+			p.computeHTTPRoute(matchingRoute, listenerSecret, listener.Hostname, errs)
```
stevesloka, line 255: I see you added `errs` as an arg to `computeHTTPRoute` and `computeTLSRoute` but those aren't used. Should we change this to be just a boolean to simply like `validGateway` or something? 

If not would be nice to describe what `errs` are a bit maybe in the name to make it clear what they mean and where they come from down in the methods.

internal/featuretests/v3/gatewayapi_test.go
```diff
@@ -30,40 +30,60 @@ import (
...
-					From: routeSelectTypePtr(gatewayapi_v1alpha1.RouteSelectAll),
+var (
+	gc = &gatewayapi_v1alpha1.GatewayClass{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "contour",
+		},
+		Spec: gatewayapi_v1alpha1.GatewayClassSpec{
+			Controller: "projectcontour.io/contour",
+		},
+		Status: gatewayapi_v1alpha1.GatewayClassStatus{
```
stevesloka, line 41: Should we leave status off of the object so that we can verify that functionality works as expected?

internal/status/conditions.go
```diff
@@ -210,6 +216,123 @@ func (c *Cache) getGatewayConditions(gatewayStatus []gatewayapi_v1alpha1.RouteGa
 	return map[gatewayapi_v1alpha1.RouteConditionType]metav1.Condition{}
 }
 
+type GatewayConditionsUpdate struct {
```
stevesloka, line 219: If this is staying, then maybe move it to its own go file and rename this file to `routeconditions.go`?

examples/gateway/02-gateway.yaml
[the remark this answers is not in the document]
danehans: > We just need to understand if Contour will ever process more than one Gateway described in #3825.

This is a must-have to support knative's use case. I see multi-gateway support as one of the next big follow-on's to this PR.

internal/dag/gatewayapi_processor.go
danehans: I'll update the PR to pass a bool that indicated gateway validity to the compute methods.

internal/status/conditions.go
danehans: Commit `4bd950a` splits route/gc/gateway conditions into separate files.

internal/featuretests/v3/gatewayapi_test.go
danehans: I tried removing status and setting "Scheduled=false" (default) but the test fails. This works when I run the e2e tests and when I test against a running cluster. Can you tell if I need to change something in `setup()`?

internal/dag/gatewayapi_processor.go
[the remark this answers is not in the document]
youngnick: Maybe the error should say something like "No gateway pointing to the GatewayClass whatever could be found"?

internal/featuretests/v3/gatewayapi_test.go
youngnick: As long as we have a way to check that Gateways that point to not-admitted GatewayClasses don't work, then I think this is okay.

internal/status/cache.go
youngnick: The base Gateway status is actually pretty simple, but the Listener status is not. It's where all the Route stuff shows up, currently, so I think that having a way to distinguish between those is good.

youngnick: I think that we should get this working and come back for code duplication later.

youngnick (approved):
Some small things, but overall I think this is okay. I'll leave to @stevesloka to give the final approval.

danehans:
> So, this needs to be here because we don't update status on admission?

@youngnick that is the gatewayclass referenced by the gateway. If a gateway references a gatewayclass that is not admitted, then the gateway is considered invalid and the appropriate status condition is surfaced.

danehans:
`e2e (kubernetes:n-2` failure due to:

```
object is being deleted: gateways.networking.x-k8s.io "contour" already exists
```

e2e tests should consider appending a UID to gateway names for each test to avoid this issue. Gateways/GatewayClasses will use finalizers so deletion may take time, especially as Contour manages Envoy infra.

stevesloka (approved):
/lgtm to get this in and unblock a bunch of other issues once we get CI passing.

skriss:
> `e2e (kubernetes:n-2` failure due to:
> 
> ```
> object is being deleted: gateways.networking.x-k8s.io "contour" already exists
> ```
> 
> e2e tests should consider appending a UID to gateway names for each test to avoid this issue. Gateways/GatewayClasses will use finalizers so deletion may take time, especially as Contour manages Envoy infra.

The plan is to move to having each test create a Gateway in its own namespace, to avoid test pollution like this. Requires #3803 to be able to easily update Contour config to point to the right gateway.

youngnick:
Only failing CI is the codecov one, I think this is good to merge @danehans.

skriss:
Looks like we've got some merge conflicts; @stevesloka were you going to take this one over?

stevesloka:
Yup I can figure this one out @skriss.

youngnick:
@stevesloka if you can clear the merge conflicts and CI is still passing, then I think this one is good to go for v1.17?

stevesloka:
@youngnick nope there's an issue with finalizers. I'd like to pick this apart into pieces so we can make sure to test everything and find the flake. Personally, I'd delay the release, but need to chat about that. Otherwise would land in 1.18.

stevesloka:
Implemented in #3858, closing and will follow the new PR for updates.

Issue #3593, opened 2021-04-19
Refactor DAG cluster to remove 'Weighted'


stevesloka:
When processing through DAG objects, a dag.Service has a `Weighted` struct which stores the service information along with the weight (if configured). 

That then translates into a `dag.Cluster` which has both a `Weighted` and a `Weight`. It's not clear which one needs to be set and makes no sense to have duplicate weight fields in the same struct.

We should refactor back out the `Weighted` struct to remove the duplication and also the stutter. 

```
// Upstream is the backend Kubernetes service traffic arriving
// at this Cluster will be forwarded too.
Upstream *Service

// The relative weight of this Cluster compared to its siblings.
Weight uint32
```

PR #3824, opened 2021-06-18
internal: program 502/503 routes for misconfigured services and includes

tsaarni:
This change implements

* Handle specific configuration errors as non-fatal, so that traffic will be served for the valid parts of the HTTPProxy.
* Program 502/503 direct responses for the faulty parts.

See design/httpproxy-conflict.md for more details, but note that this change addresses only missing includes and missing service references, while the design document covers wider concept of misconfigured proxies.

### DIFF
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -22,6 +22,7 @@ import (
 
 	contour_api_v1 "github.com/projectcontour/contour/apis/projectcontour/v1"
 	"github.com/projectcontour/contour/internal/fixture"
+	"github.com/projectcontour/contour/internal/status"
 	"github.com/projectcontour/contour/internal/timeout"
 	"github.com/stretchr/testify/assert"
 	v1 "k8s.io/api/core/v1"
@@ -6380,13 +6381,8 @@ func TestDAGInsert(t *testing.T) {
 		},
 		Spec: contour_api_v1.HTTPProxySpec{
 			Routes: []contour_api_v1.Route{{
-				Conditions: []contour_api_v1.MatchCondition{{
-					Prefix: "/v1",
-				}, {
-					Prefix: "/api",
-				}},
 				Services: []contour_api_v1.Service{{
-					Name: s1.Name,
+					Name: s12.Name,
 					Port: 8080,
 				}},
 			}},
@@ -6946,8 +6942,8 @@ func TestDAGInsert(t *testing.T) {
 					Prefix: "/",
 				}},
 				Services: []contour_api_v1.Service{{
-					Name: "kuard",
-					Port: 8080,
+					Name: "nginx",
+					Port: 80,
 				}},
 				ResponseHeadersPolicy: &contour_api_v1.HeadersPolicy{
 					Set: []contour_api_v1.HeaderValue{{
@@ -6973,8 +6969,8 @@ func TestDAGInsert(t *testing.T) {
 					Prefix: "/",
 				}},
 				Services: []contour_api_v1.Service{{
-					Name: "kuard",
-					Port: 8080,
+					Name: "nginx",
+					Port: 80,
 					ResponseHeadersPolicy: &contour_api_v1.HeadersPolicy{
 						Set: []contour_api_v1.HeaderValue{{
 							Name:  "Host",
@@ -7640,7 +7636,14 @@ func TestDAGInsert(t *testing.T) {
 			objs: []interface{}{
 				proxyMultipleBackends, s2,
 			},
-			want: listeners(),
+			want: listeners(
+				&Listener{
+					Port: 80,
+					VirtualHosts: virtualhosts(
+						virtualhost("example.com", prefixroute("/", service(s2))),
+					),
+				},
+			),
 		},
 		"insert httpproxy referencing two backends": {
 			objs: []interface{}{
@@ -8290,7 +8293,19 @@ func TestDAGInsert(t *testing.T) {
 					},
 				},
 			},
-			want: nil, // no listener created
+			want: listeners(
+				&Listener{
+					Port: 80,
+					VirtualHosts: virtualhosts(
+						virtualhost("example.com", &Route{
+							PathMatchCondition: prefixString("/finance"),
+							DirectResponse: &DirectResponse{
+								StatusCode: http.StatusBadGateway,
+							},
+						}),
+					),
+				},
+			),
 		},
 		"insert httproxy w/ conditions": {
 			objs: []interface{}{
@@ -9196,9 +9211,17 @@ func TestDAGInsert(t *testing.T) {
 		},
 		"insert httpproxy with multiple prefix conditions on include": {
 			objs: []interface{}{
-				proxy103, proxy103a, s1,
+				proxy103, proxy103a, s1, s12,
 			},
-			want: listeners(),
+			want: listeners(
+				&Listener{
+					Port: 80,
+					VirtualHosts: virtualhosts(
+						// route on root proxy is served, includes is ignored since condition is invalid
+						virtualhost("example.com", prefixroute("/", service(s1))),
+					),
+				},
+			),
 		},
 		"insert httpproxy duplicate conditions on include": {
 			objs: []interface{}{
@@ -10327,6 +10350,7 @@ func TestDAGInsert(t *testing.T) {
 				},
 			),
 		},
+		"multiple services with weight and missing service reference": {},
 	}
 
 	for name, tc := range tests {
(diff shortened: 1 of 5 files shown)


tsaarni:
I would like to propose that we could proceed with a partial implementation of [httpproxy-conflict.md](https://github.com/projectcontour/contour/blob/main/design/httpproxy-conflict.md). This is relatively small change, since it addresses only the low-hanging fruits of the whole conflict resolution design. It could be seen as first step towards implementing it. And this would be immediately useful at least for us: we have deployment scenarios where we can loose complete virtualhost due to having invalid references when the application is deployed partially.

internal/dag/httpproxy_processor.go
```diff
@@ -400,22 +401,34 @@ func (p *HTTPProxyProcessor) computeRoutes(
...
+		}
+
 		includedProxy, ok := p.source.httpproxies[types.NamespacedName{Name: include.Name, Namespace: namespace}]
 		if !ok {
 			validCond.AddErrorf(contour_api_v1.ConditionTypeIncludeError, "IncludeNotFound",
 				"include %s/%s not found", namespace, include.Name)
-			return nil
+
+			// Set 502 response when include was not found but route prefix was valid.
+			if len(include.Conditions) > 0 {
```
stevesloka, line 416: We should probably require some set of conditions on this, but that's a different issue. =)

internal/dag/httpproxy_processor.go
```diff
@@ -640,6 +653,11 @@ func (p *HTTPProxyProcessor) computeRoutes(
 				r.Clusters = append(r.Clusters, c)
 			}
 		}
+		if len(r.Clusters) == 0 {
```
stevesloka, line 656: We should baby step this, but the next one would be to handle multiple services in the same route and setting the invalid one to a weight of zero.

internal/dag/httpproxy_processor.go
```diff
@@ -640,6 +653,11 @@ func (p *HTTPProxyProcessor) computeRoutes(
...
+			r.DirectResponse = &DirectResponse{
```
stevesloka, line 657: Might be helpful now that this is used in multiple places to centralize the logic so that if it changes, we'll have a single place to update.

internal/dag/status_test.go
```diff
@@ -568,8 +568,8 @@ func TestDAGStatus(t *testing.T) {
 			}},
 			Routes: []contour_api_v1.Route{{
 				Services: []contour_api_v1.Service{{
-					Name: "green",
-					Port: 80,
+					Name: "kuard",
```
stevesloka, line 571: Why change this?

internal/dag/builder_test.go
```diff
@@ -8452,7 +8471,14 @@ func TestDAGInsert(t *testing.T) {
 			objs: []interface{}{
 				proxy103, proxy103a, s1,
 			},
-			want: listeners(),
+			want: listeners(
```
stevesloka, line 8474: This should still result in an error right? You can't specify multiple prefixes on a route. Additionally, the one's specified in the proxyies are for `/v1` and `/api`, so did `/` match?

internal/dag/builder_test.go
```diff
@@ -6945,7 +6945,14 @@ func TestDAGInsert(t *testing.T) {
 			objs: []interface{}{
 				proxyMultipleBackends, s2,
 			},
-			want: listeners(),
+			want: listeners(
```
stevesloka, line 6948: Why does this now return listeners?

internal/dag/builder_test.go
```diff
@@ -6271,8 +6271,8 @@ func TestDAGInsert(t *testing.T) {
 					Prefix: "/",
 				}},
 				Services: []contour_api_v1.Service{{
-					Name: "kuard",
-					Port: 8080,
+					Name: "nginx",
```
stevesloka, line 6274: Why swap out for `nginx`?

stevesloka (changes requested):
Quick once over @tsaarni, but looking good I think!

internal/dag/status_test.go
tsaarni: I think the test case has always been wrong:  According to comment, the test case is supposed to test include cycle, but besides that, it has another error as well: it referred to non-existing service "green".  Now that non-existing service reference is not fatal anymore, I noticed this

internal/dag/builder_test.go
tsaarni: I think this has also been wrong all the time: The proxy is used in test _"insert proxy with response header policy - route - host header"_ [here](https://github.com/projectcontour/contour/blob/0a94c00f2bf748e1afee914c8ce5fe06b0707a60/internal/dag/builder_test.go#L8913) and it is expected to return no listeners. However, the reason why it returned no listeners previously was because the proxy has invalid service reference. The test includes service `s9` which is `nginx`, not `kuard`. Now that invalid service reference will result in listener, I catched this error.

internal/dag/builder_test.go
tsaarni: I think this is according to [design doc chapter:](https://github.com/projectcontour/contour/blob/main/design/httpproxy-conflict.md#multiple-services) 

> For the case where multiple services are referenced from a Spec.Route of an HTTPProxy and has an Error, Contour will set the invalid service weight to be zero so that no traffic is routed to the service in error and will also set an error in the Conditions of the object.

That is, since one (of the two) service references is still valid, traffic should be routed to the valid service.

internal/dag/builder_test.go
tsaarni: True, something fishy here, sorry for missing this! Will come back shortly, after going through this with a debugger.

tsaarni: Didn't need debugger but it seems some of the test cases get bit tricky. Even if I just gone through this one I still got confused a few days after :)  In this case we have two proxies:  root proxy `proxy103` and child proxy `proxy103a`. Note that the [root proxy](https://github.com/projectcontour/contour/blob/0a94c00f2bf748e1afee914c8ce5fe06b0707a60/internal/dag/builder_test.go#L5674) has both `includes` **and also** `routes` section:

* Include section is invalid because there is two prefix conditions.  Therefore child `proxy103a` is not considered at all.
* Route section to `s1` is valid. It has no prefix condition i.e. prefix is implied `/`

Since the [route section](https://github.com/projectcontour/contour/blob/0a94c00f2bf748e1afee914c8ce5fe06b0707a60/internal/dag/builder_test.go#L5692) is valid we got listener for routing everything to `s1`. 

So I think from this perspective it was correct after all.  Maybe the test could be made bit easier to understand, for example: 

* Maybe errors in child proxy could be removed. They serve no meaning for the result of this test but may confuse reader.
* Don't use exactly same `/v1` and `/api` prefixes in both [root proxy](https://github.com/projectcontour/contour/blob/0a94c00f2bf748e1afee914c8ce5fe06b0707a60/internal/dag/builder_test.go#L5674) and the [child proxy](https://github.com/projectcontour/contour/blob/0a94c00f2bf748e1afee914c8ce5fe06b0707a60/internal/dag/builder_test.go#L5701).  If we did not have error due to multiple includes, full prefix could be doubled, like `/v1/v1` which might unnecessarily confuse the reader.

tsaarni: If I change the test case by removing the double prefix conditions and I leave `/api` to root proxy, `/v1` to child proxy, then I get listener with

* `/` routed to `s1`
* `/api/v1` with 503 direct response

reader maybe would have expected route `/api/v1` to be valid and direct to `s1` as well, but now the problem is that child proxy is create in namespace `teama`, while services `s1` is in `default`. So maybe two more suggestions to make the test easier to understand:

* Have two different services, so route in root and route in child proxy do not refer to same service
* Create the second service in `teama` namespace and have child refer to that

internal/dag/httpproxy_processor.go
tsaarni: I may have misunderstood so please correct me! I wonder wouldn't the logic around direct responses always be different, making it hard to centralize? Now currently there are two direct responses by this PR:

* Here in this spot we have processed whole `routes`-section without failures, including missing service which is now passed through, but we still have no clusters to route to -> instantiate `StatusServiceUnavailable`
* Another case is in when `includes`-section is processed, we did not find target of include, but we have known condition -> instantiate `StatusBadGateway`

So will the logic always depend on the "context"?  When we cover more error scenarios from the design doc, wouldn't that logic also need to be sprinkled around each error handling branch, not leaving any logic to be centralized?

internal/dag/httpproxy_processor.go
tsaarni: Here my thinking was that I move the `pathMatchConditionsValid()` call up, so that when we recognize missing include, I know that prefix is valid -> program `502` for the prefix.

But I now realized this is bit more complex:  `pathMatchConditionsValid()`  does not guarantee there is prefix. Should we program `502` also for header conditions if someone used them in `includes[].conditions[]`? I suppose we should?

The code that processes includes has always validated only prefix by calling `pathMatchConditionsValid(include.Conditions)`, there is no call to `headerMatchConditionsValid(include.Conditions)`.
If I can leave prefix field empty and add `includes[].conditions[].headers[]` instead, which gets merged in child proxy route `conditions`.  Lets say for example:

* root has header condition `x-foo` equals `bar`
* child has prefix condition `/api`

both are then combined when processing child proxy's route conditions. In this case traffic is forwarded only if prefix is `/api` and request has `x-foo` header with value `bar`. I guess this is intentional?

tsaarni:
> Quick once over @tsaarni, but looking good I think!

Thanks @stevesloka for the first round!  And sorry for long responses from my side. I'm bit overwhelmed by all the nitty-gritty details I bump into and how to best explain/ask about them :)

internal/dag/httpproxy_processor.go
tsaarni: I wonder, will the invalid service be programmed at all, i.e. does it exist from Envoy perspective?

The simplest case seemed to work already now: I had two services for a route, one of those missing (and of course missing service cannot be programmed to Envoy).

internal/dag/httpproxy_processor.go
stevesloka: >  In this case traffic is forwarded only if prefix is /api and request has x-foo header with value bar. I guess this is intentional?

Yup that's how it's designed, the resulting set of conditions define the match so adding in different headers could let the request route to different places.

internal/dag/httpproxy_processor.go
stevesloka: Yes, there wouldn't be a cluster programming in envoy matching the missing service, but I believe our route logic would still try and program it. We could change that logic to not program a the route with a missing service or set the weight to zero. 

Either way we'll need a way to understand that the service is invalid somehow when building out the route. But this can be a different PR. This one handling the case for a single service, then another PR to cover multiple services cases.

internal/dag/httpproxy_processor.go
stevesloka: I was more suggesting that the `&DirectResponse` bit might be hidden inside a method somewhere and the specifics could be passed into that method. This way if the struct changed in Envoy, we'd update it in one place not multiple.

internal/dag/status_test.go
stevesloka: 👍

internal/dag/builder_test.go
tsaarni: I've now updated existing test case "insert httpproxy with multiple prefix conditions on include" just for bit more clarity.

internal/dag/httpproxy_processor.go
tsaarni: Fixed.

internal/dag/httpproxy_processor.go
tsaarni: I did not manage to find where in the route logic missing services should be deal. Though I have some difficulties understanding the difference between cluster weight (`Cluster.Weight`) vs service weight `Cluster.Upstream.Weighted.Weight`, so I might be missing some key piece here...

Meanwhile, I added test case "root proxy refers to three services with weights, one is missing", where the missing service just results in one less cluster.  The weights of existing clusters are kept like they were in the spec.

internal/dag/httpproxy_processor.go
tsaarni: Now both prefix and header conditions are considered when programming 502 response for include.

There was an existing test case `TestDAGStatus/duplicate_include_condition_headers` that had header (intentionally) invalid header conditions for includes. Because header conditions for includes were not previously validated at root proxy, they were propagated to child and child got blamed for the error.  Now the error condition is set for root proxy where the invalid conditions are and the test case is fixed as well.

internal/dag/httpproxy_processor.go
stevesloka: > Though I have some difficulties understanding the difference between cluster weight (Cluster.Weight) vs service weight Cluster.Upstream.Weighted.Weight, so I might be missing some key piece here...

No I agree that's confusing and never got cleaned up: https://github.com/projectcontour/contour/issues/3593

stevesloka (commented):
All in all, this looks great to me @tsaarni!! We should also try and set some sort of Conditions on the HTTPProxy when this happens to inform the user that something has gone wrong. So in the case where a service is missing, we need to still serve traffic and return the 502 like this PR adds, but also inform the user that it's happening so they have the ability to resolve. 

We could split that out into a seperate PR if needed since it's going to be a bunch more work to spin up that code. But would be needed before shipping a release.

stevesloka:
The only other thing I just thought of would be to add a Contour specific header on the response when a 502 or something is returned. This would just be to help identify that Contour programmed the 502 vs the backend service responding.

youngnick:
I think both of those are great ideas @stevesloka.

tsaarni:
Sorry for slow response, it is summer vacation period here 🌴 

> We should also try and set some sort of Conditions on the HTTPProxy when this happens to inform the user that something has gone wrong. 

I thought the errors were reflected quite well in status conditions previously, and that allowing traffic for valid parts of `HTTPProxy` would not necessarily change anything regarding the error messages. For better visibility, I've now added `wantStatus` field in `internal/dag/builder_test.go:TestHTTPProxyConficts()`. It shows what user sees in various misconfiguration scenarios.  Could you check those and give an example on what kind of status conditions you were thinking?

> The only other thing I just thought of would be to add a Contour specific header on the response when a 502 or something is returned. This would just be to help identify that Contour programmed the 502 vs the backend service responding.

What would be a good header name and how should we set the value?  Just as initial proposal, I've now added `X-Contour-Status: Invalid configuration` to the direct response. It seemed all direct responses are for configuration errors, so I just added a fixed value for now.

tsaarni:
PTAL when time allows / vacations are over.

stevesloka:
Sorry for the delay on this @tsaarni, let me take another look over this and see if we can get this merged. =)

stevesloka (approved):
/lgtm at a first start! Thanks for all the TODOs in line for other issues. We should make sure to capture those. I'll put together an epic and link off issues to follow up with.

internal/featuretests/v3/tcpproxy_test.go
```diff
@@ -97,7 +107,20 @@ func TestTCPProxy(t *testing.T) {
 	// check that ingress_http is empty
```
sunjayBhatia, line 107: nit: this comment might not be relevant anymore

internal/featuretests/v3/tcpproxy_test.go
```diff
@@ -254,7 +288,20 @@ func TestTCPProxyTLSPassthrough(t *testing.T) {
...
```
sunjayBhatia, line 288: same for this one

internal/xdscache/v3/route.go
```diff
@@ -148,6 +149,14 @@ func (v *routeVisitor) onVirtualHost(vh *dag.VirtualHost) {
 			return &envoy_route_v3.Route{
 				Match:  envoy_v3.RouteMatch(route),
 				Action: envoy_v3.RouteDirectResponse(route.DirectResponse),
+				// Add custom header to the direct response to help user identify that
```
sunjayBhatia, line 152: I think this is missing in the `onSecureVirtualHost` version of this as well

tsaarni:
I wonder how to proceed with this PR, is there be something I could still address?

internal/xdscache/v3/route.go
```diff
@@ -197,6 +206,14 @@ func (v *routeVisitor) onSecureVirtualHost(svh *dag.SecureVirtualHost) {
...
```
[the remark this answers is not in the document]
stevesloka: This came up on office hours last week, another user wanted this feature as well. 

I know I brought up adding the response header, but I think we should take that out. Something like a metric might be a better use case to identify when this is happening.

youngnick:
Reading over this, looks like we are currently at:
- approach looks good overall
- we should probably remove the extra header being added for now, and add this in in a more considered way.
- Instead, there should be a metric incremented when a DAG run means that there will be a 502/503 direct response created. In that case, I think that the metric should include some labels about the object that produced the config. That way the Contour owner has a way to track who's generating direct responses if they wish. Not sure if we do any metrics like this at the moment, this could possibly also be left for a followup PR, since we have HTTPProxy status already updating.

I haven't had the bandwidth to cover all the stuff that @stevesloka and @sunjayBhatia have here, so I'd like to hear from them here too.

Thanks for the poke @tsaarni, it would be really nice to get this in.

tsaarni:
I've now removed the `X-Contour-Status` response header from direct responses.  

I'm looking into how to add this information on metrics.  It seems to me it may require some new mechanism. I see that currently most metrics from are originating from `HTTPProxyProcessor` and they are carried via `ProxyUpdate.Conditions`.  I have not yet figured out how to hook new metrics in DAG processing - I assume it would need to be a place that would be aware of both vhosts and routes (for the direct response status code info) in order to add relevant labels to the metrics.

sunjayBhatia:
> I've now removed the `X-Contour-Status` response header from direct responses.
> 
> I'm looking into how to add this information on metrics. It seems to me it may require some new mechanism. I see that currently most metrics from are originating from `HTTPProxyProcessor` and they are carried via `ProxyUpdate.Conditions`. I have not yet figured out how to hook new metrics in DAG processing - I assume it would need to be a place that would be aware of both vhosts and routes (for the direct response status code info) in order to add relevant labels to the metrics.

personally I would think looking at the existing `contour_httpproxy_invalid` metric would be enough to alert an owner of an HTTPProxy that their configuration is invalid, rather than adding an additional metric for now

sunjayBhatia (approved):

tsaarni:
> from the code looks like we are setting condition Valid=false with errors, so maybe this is enough?

Yes exactly, we will get that. We lack direct feedback telling user it was Envoy who generated 502/503, not the upstream service. But I wonder how metric with a number of direct responses (per virtualhost) programmed to Envoy would help? I would think the current metrics are enough. Though, I feel I might have missed how this new metrics could work :)

youngnick (approved):
After thinking about this one for a bit, I think that the current conditions are enough to start with here. Let's see if we can get some feedback from users on what's useful for them, if the condition is not enough.

PR #3827, opened 2021-06-21
pkg/config: Mark Gateway.Name & Gateway.Namespace as deprecated

stevesloka:
The fields in the Gateway configuration, Name & Namespace, are deprecated and will be removed in Contour v1.18.

Updates #3617

### DIFF
--- a/site/content/docs/main/configuration.md
+++ b/site/content/docs/main/configuration.md
@@ -180,8 +180,11 @@ The gateway configuration block is used to configure which gateway-api Gateway C
 | Field Name | Type| Default  | Description |
 |------------|-----|----------|-------------|
 | controllerName | string |  | Gateway Class controller name (i.e. projectcontour.io/projectcontour/contour).  |
-| name | string | contour | This field specifies the name of a Gateway.  |
-| namespace | string | projectcontour | This field specifies the namespace of a Gateway.  |
+| name | string | contour | DEPRECATED: This field specifies the name of a Gateway.  |
+| namespace (Deprecated) | string | projectcontour | DEPRECATED: This field specifies the namespace of a Gateway.  |
+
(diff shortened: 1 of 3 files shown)


site/content/docs/main/configuration.md
```diff
@@ -180,8 +180,11 @@ The gateway configuration block is used to configure which gateway-api Gateway C
 | Field Name | Type| Default  | Description |
 |------------|-----|----------|-------------|
 | controllerName | string |  | Gateway Class controller name (i.e. projectcontour.io/projectcontour/contour).  |
-| name | string | contour | This field specifies the name of a Gateway.  |
-| namespace | string | projectcontour | This field specifies the namespace of a Gateway.  |
+| name | string | contour | DEPRECATED: This field specifies the name of a Gateway.  |
```
skriss, line 183: For consistency:
```suggestion
| name (Deprecated) | string | contour | DEPRECATED: This field specifies the name of a Gateway.  |
```

skriss (commented):
One nit, otherwise looks reasonable and inline with our [config file deprecation policy](https://projectcontour.io/resources/deprecation-policy/)

skriss (approved):

youngnick (approved):
Should we hold this until we merge #3797 though? Since we haven't actually technically agreed until that happens?

stevesloka:
>  Since we haven't actually technically agreed until that happens?

Yeah maybe? It's already there: https://github.com/projectcontour/contour/blob/main/pkg/config/parameters.go#L293-L296

youngnick:
I think we're agreed that we will do this, we're just cleaning up the doc now, so it's probably fine.

danehans:
cc: @nak3

PR #3834, opened 2021-06-24
Bump protobuf and fix lint issues

sunjayBhatia:
Found while bumping gomega in another PR, protobuf can be updated and
some deprecated helpers can be removed.

### DIFF
--- a/internal/protobuf/helpers.go
+++ b/internal/protobuf/helpers.go
@@ -21,10 +21,10 @@ import (
 
 	"github.com/golang/protobuf/jsonpb"
 	"github.com/golang/protobuf/proto"
-	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/any"
 	"github.com/golang/protobuf/ptypes/duration"
 	"github.com/golang/protobuf/ptypes/wrappers"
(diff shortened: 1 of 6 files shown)


internal/protobuf/helpers.go
```diff
@@ -87,10 +87,10 @@ func AsMessages(messages interface{}) []proto.Message {
 	return protos
 }
 
-// MustMarshalAny marshals a protobug into an any.Any type, panicking
```
skriss, line 90: 😂

skriss (approved):
LGTM pending CI

internal/xds/v3/contour_test.go
```diff
@@ -81,7 +82,7 @@ func TestXDSHandlerStream(t *testing.T) {
 					}, nil
 				},
 			},
-			want: fmt.Errorf("proto: Marshal called with nil"),
+			want: protoimpl.X.NewError("invalid nil source message"),
```
sunjayBhatia, line 85: could also change the test to assert on error strings, so we don't have to fish out how exactly the error is constructed

skriss (approved):

youngnick (approved):

Issue #3729, opened 2021-05-26
test/e2e: Convert 013-backend-tls.yaml to Go e2e test


sunjayBhatia:
See: https://github.com/projectcontour/contour/blob/02d46eecf1ebea8da4228cca49936ee164c6ac9d/_integration/testsuite/httpproxy/013-backend-tls.yaml

This test is currently skipped in CI b/c it requires a specific config to be set in the Contour config file: https://github.com/projectcontour/contour/blob/02d46eecf1ebea8da4228cca49936ee164c6ac9d/_integration/testsuite/httpproxy/013-backend-tls.yaml#L34-L48

This test and others like it that we may want to write in the future will likely be the impetus for running e2e tests against a Contour installation that can be restarted as part of the tests

PR #3846, opened 2021-06-28
Add backend tls e2e tests

sunjayBhatia:
Ingress and HTTPProxy backend TLS tests

Fixes: #3729

### DIFF
--- a/test/e2e/httpproxy/backend_tls_test.go
+++ b/test/e2e/httpproxy/backend_tls_test.go
@@ -1,0 +1,141 @@
+// Copyright Project Contour Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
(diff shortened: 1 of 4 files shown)


test/e2e/httpproxy/backend_tls_test.go
```diff
@@ -0,0 +1,141 @@
...
+		// Delete client cert so it is rotated.
+		require.NoError(f.T(), f.Client.Delete(context.TODO(), clientSecret))
+		// Make sure it is deleted.
+		require.Eventually(f.T(), func() bool {
+			return errors.IsNotFound(f.Client.Get(context.TODO(), clientSecretKey, clientSecret))
+		}, time.Second*10, time.Millisecond*50)
+		// Make sure it comes back.
+		require.Eventually(f.T(), func() bool {
+			return f.Client.Get(context.TODO(), clientSecretKey, clientSecret) == nil
+		}, time.Second*10, time.Millisecond*50)
```
skriss, line 121: I think there's a race condition here where if it's deleted and recreated in between iterations of the first check for deletion, this will fail. Not sure if it's a real issue in practice or not. One alternative would be to capture the UID before deletion, and then wait until a get succeeds and returns an object with a different UID meaning it was deleted + recreated.

skriss (changes requested):

test/e2e/httpproxy/backend_tls_test.go
sunjayBhatia: ah yeah good point, thats why the old test added an annotation and waited for it to be gone, I'll look into changing this around to prevent races

skriss (approved):
:+1: LGTM when green!

stevesloka (approved):

Issue #3576, opened 2021-04-14
Add option to remove query strings from access log


tsaarni:
**Feature request**

Envoy will log the complete request path, including the query string like `?supersecret=password` in this example

```
[2021-04-14T16:36:00.361Z] "GET /foo?supersecret=password HTTP/1.1" 200 - 0 463 6 3 "-" "HTTPie/1.0.3" "837aa8dc-344f-4faa-b7d5-c9cce1028519" "localhost:8080" "127.0.0.1:8081"
```

It would be great to provide an option to remove query string from the access log, since it may contain sensitive information introducing a potential [vulnerability](https://owasp.org/www-community/vulnerabilities/Information_exposure_through_query_strings_in_url). If the problem cannot be fixed in the proxied application itself, the vulnerability could be mitigated by sanitazing the access logs.

**Potential options**

1) Contour could add HTTP filter called [Header-To-Metadata](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/header_to_metadata/v3/header_to_metadata.proto) and use it to remove query string from :PATH header and set the result in dynamic metadata, see https://github.com/envoyproxy/envoy/pull/15711#issuecomment-810511085.  Then Contour could provide the user an option to change the [default access log format string](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage#default-format-string), where they would replace `%REQ(X-ENVOY-ORIGINAL-PATH?:PATH)%` (including the query string) with `%DYNAMIC_METADATA(projectcontour.io:path)%` (query string sanitized)

2) Proceed with the PR https://github.com/envoyproxy/envoy/pull/15711 (or something similar) which intends to introduce a new extension command e.g. `REQ_WITHOUT_QUERY` for the purpose of sanitizing the headers. Next step would be for Contour to enable this extension in the access log configuration within HTTP Connection Manager. Then provide the user an option to change the default access log format string where they would replace `%REQ(X-ENVOY-ORIGINAL-PATH?:PATH)%` with `%REQ_WITHOUT_QUERY(X-ENVOY-ORIGINAL-PATH?:PATH)%`

Option (1) is possible to implement with existing versions of Envoy, while (2) would introduce a version dependency.

youngnick:
Thanks for this issue @tsaarni. I'm leaning towards option 2, maybe with a boolean to strip query strings from the access logs? The only other option I see is full configurability of the access log format, which, while worthwhile, would need some more work.

[...]

Issue #3829, opened 2021-06-22
make the guides version-specific


skriss:
Right now, the [Contour guides](https://projectcontour.io/guides/) are not version-specific, whereas all of our other docs are. I think it would make sense to version them, so that as things change in Contour, the guides can remain accurate for past versions while still being updated to account for new features/changes.

youngnick:
Yeah, this makes sense to me.

PR #3849, opened 2021-06-29
site: Added versioned access logging document

tsaarni:
Added information about 

* customizing text based access logs
* supported access log formatter extensions (there is only one so far)

This documentation is related to #3694 and #3576

### DIFF
--- a/site/content/docs/main/config/access-logging.md
+++ b/site/content/docs/main/config/access-logging.md
@@ -1,0 +1,126 @@
+# Access Logging
+
+## Overview
+
+Contour allows you to control Envoy's access logging.
+By default, HTTP and HTTPS access logs are written to `/dev/stdout` by the Envoy containers and look like following:
+
+```
+[2021-04-14T16:36:00.361Z] "GET /foo HTTP/1.1" 200 - 0 463 6 3 "-" "HTTPie/1.0.3" "837aa8dc-344f-4faa-b7d5-c9cce1028519" "localhost:8080" "127.0.0.1:8081"
+```
+
+The detailed description of each field can be found in [Envoy access logging documentation][7].
+
+
+## Customizing Access Log Destination and Formats
+
+You can change the destination file where the access log is written by using Contour [command line parameters][1] `--envoy-http-access-log` and `--envoy-https-access-log`.
+
+The access log can take two different formats, both can be customized
+
+* Text based access logs, like shown in the example above.
+* Structured JSON logging.
+
+### Text Based Access Logging
+
+Ensure that you have selected `envoy` as the access log format.
+Note that this is the default format if the parameters are not given.
+
+- Add `--accesslog-format=envoy` to your Contour startup line, or
+- Add `accesslog-format: envoy` to your configuration file.
+
+Customize the access log format by defining `accesslog-format-string` in your configuration file.
+
+```yaml
+accesslog-format-string: "[%START_TIME%] \"%REQ(:METHOD)% %REQ(X-ENVOY-ORIGINAL-PATH?:PATH)% %PROTOCOL%\" %RESPONSE_CODE% %RESPONSE_FLAGS% %BYTES_RECEIVED% %BYTES_SENT% %DURATION% %RESP(X-ENVOY-UPSTREAM-SERVICE-TIME)% \"%REQ(X-FORWARDED-FOR)%\" \"%REQ(USER-AGENT)%\" \"%REQ(X-REQUEST-ID)%\" \"%REQ(:AUTHORITY)%\" \"%UPSTREAM_HOST%\"\n"
+```
+After restarting Contour and successful validation of the configuration, the new format will take effect in a short while.
+
+Refer to [Envoy access logging documentation][7] for the description of the command operators, and note that the format string needs to end in a linefeed `\n`.
+
+### Structured JSON Logging
+
+Contour allows you to choose from a set of JSON fields that will be expanded into Envoy templates and sent to Envoy.
+There is a default set of fields if you enable JSON logging, and you may customize which fields you log.
+
+The list of available fields are discoverable in the following objects:
+- [JSONFields][2] are fields that have built in mappings to commonly used envoy operators.
+- [envoySimpleOperators][3] are the names of simple envoy operators that don't require arguments, they are case-insensitive when configured.
+- [envoyComplexOperators][4] are the names of complex envoy operators that require arguments.
+
+The default list of fields is available at [DefaultFields][5].
+
+#### Enabling the Feature
+
+To enable the feature you have two options:
+
+- Add `--accesslog-format=json` to your Contour startup line.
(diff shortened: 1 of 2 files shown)


tsaarni:
xref #3829

site/content/docs/main/config/access-logging.md
```diff
@@ -0,0 +1,125 @@
+# Access logging
+
+## Overview
+
+Contour allows you to control Envoy's access logging.
+HTTP and HTTPS access logs are written to `/dev/stdout` by the Envoy containers.
+The access log looks like following:
```
skriss, line 7:
```suggestion
By default, HTTP and HTTPS access logs are written to `/dev/stdout` by the Envoy containers and look like the following:
```

site/content/docs/main/config/access-logging.md
```diff
@@ -0,0 +1,125 @@
...
+
+- Add `--accesslog-format=envoy` to your Contour startup line, or
+- Add `accesslog-format: envoy` to your configuration file.
+
+Customize the access log format by defining `accesslog-format-string` in your configuration file:
+
+```yaml
+accesslog-format-string: "[%START_TIME%] \"%REQ(:METHOD)% %REQ(X-ENVOY-ORIGINAL-PATH?:PATH)% %PROTOCOL%\" %RESPONSE_CODE% %RESPONSE_FLAGS% %BYTES_RECEIVED% %BYTES_SENT% %DURATION% %RESP(X-ENVOY-UPSTREAM-SERVICE-TIME)% \"%REQ(X-FORWARDED-FOR)%\" \"%REQ(USER-AGENT)%\" \"%REQ(X-REQUEST-ID)%\" \"%REQ(:AUTHORITY)%\" \"%UPSTREAM_HOST%\"\n"
+```
+
```
skriss, line 39: I'd add a note here that the format string needs to end in a `\n`

site/data/docs/main-toc.yml
```diff
@@ -39,6 +39,8 @@ toc:
         url: /config/annotations
       - page: API Reference
         url: /config/api
+      - page: Access logging
+        url: /config/access-logging
```
skriss, line 43: I'd move this up to just above "Annotations Reference" to leave the two reference pages at the bottom of this TOC section. Separately (not needed for this PR) we might want to move them into their own "Reference" section.

skriss (changes requested):
Looks pretty good to me @tsaarni, just a couple small comments. It looks like this effectively replaces the existing guide, so should we remove it?

site/content/docs/main/config/access-logging.md
```diff
@@ -0,0 +1,125 @@
...
+  - "uber_trace_id"
+  - "upstream_cluster"
+  - "upstream_host"
+  - "upstream_local_address"
+  - "upstream_service_time"
+  - "user_agent"
+  - "x_forwarded_for"
+```
+
+## Using Access Log Formatter Extensions
```
youngnick, line 109: As with my comment on #3694, I don't think that we should include this until it's included in the Envoy version that Contour supports.

youngnick (changes requested):
Similarly to #3694, I think that we should either remove references to REQ_WITHOUT_QUERY until Envoy supports it, or hold this PR until then.

tsaarni:
> It looks like this effectively replaces the existing guide, so should we remove it?

How should we cover the old versions with regards #3829 while moving information from guides?

Before removing the original document from site level guides, should the information be duplicated to old versions as well for clarity? And in that case, I guess we should also check what information applies to which version, e.g. JSON logs beginning from Contour 1.10.0.

tsaarni:
> Similarly to #3694, I think that we should either remove references to REQ_WITHOUT_QUERY until Envoy supports it, or hold this PR until then.

Lets hold until Envoy 1.19.0 is released!

site/content/docs/main/config/access-logging.md
tsaarni: I've added note about `\n`.  I also linked the page about Envoy command operator explanations here in this chapter.

site/content/docs/main/config/access-logging.md
tsaarni: Fixed!

site/data/docs/main-toc.yml
tsaarni: Fixed!

tsaarni:
I noticed that in the JSON chapter, the links to Contour code (for the list of command operators in `envoySimpleOperators` and `envoyComplexOperators`) are bit out-of-sync.  If we wan't to keep linking to the code, maybe we should fix the git hash in links to a specific version or just link to the file, without deep linking to line numbers?

Also, I guess user would be interested in reading the Envoy docs what the command operators do. We could have a separate chapter that links to Envoy doc as explanation on what the operators do and then say that from available command operators, Contour supports X,Y and Z.  

The list of command operators is a bit moving target and even now Envoy has some command operators introduced in latest released Envoy that are not known by the current parser code. Maybe these should be added?

skriss:
> I noticed that in the JSON chapter, the links to Contour code (for the list of command operators in `envoySimpleOperators` and `envoyComplexOperators`) are bit out-of-sync. If we wan't to keep linking to the code, maybe we should fix the git hash in links to a specific version or just link to the file, without deep linking to line numbers?

Yeah, I think either of these could work; I'd probably lean towards just linking to the file and describing which part to look at.

> Also, I guess user would be interested in reading the Envoy docs what the command operators do. We could have a separate chapter that links to Envoy doc as explanation on what the operators do and then say that from available command operators, Contour supports X,Y and Z.

Yeah, I think this is a good idea.

> The list of command operators is a bit moving target and even now Envoy has some command operators introduced in latest released Envoy that are not known by the current parser code. Maybe these should be added?

:+1: also probably a good idea, can file a separate issue to add the missing one.

skriss:
> > It looks like this effectively replaces the existing guide, so should we remove it?
> 
> How should we cover the old versions with regards #3829 while moving information from guides?
> 
> Before removing the original document from site level guides, should the information be duplicated to old versions as well for clarity? And in that case, I guess we should also check what information applies to which version, e.g. JSON logs beginning from Contour 1.10.0.

These are all good points. I'm fine with adding this new doc for now without removing the guide, and then we can resolve these questions as part of #3829.

youngnick (approved):
Since the support for REQ_WITHOUT_QUERY is now in, I'm good with this going in.

skriss:
@tsaarni I'll leave this open for now in case you can address the recent comments, but they can also be addressed in a follow-up if needed, we'll definitely get this in for the upcoming release.

sunjayBhatia (commented):
content LGTM, pending the content that may be added based on other reviews

skriss:
I filed #3920 for follow-ups, merging this as-is. Thanks @tsaarni!

skriss (approved):

PR #3851, opened 2021-06-30
test/e2e: Start of incluster test suite

sunjayBhatia:
This test suite should contain tests for things that we have to test
with Contour running in an actual k8s cluster.

This initial commit adds some tests that make sure our Contour
ClusterRole is set up with RBAC for projectcontour.io and Ingress
resources.

Next steps in this track of work are to add more coverage to ensure RBAC
is correct for resources in other API groups and to test other
codepaths/default configuration we rely on in our default deployment
etc.

Updates: https://github.com/projectcontour/contour/issues/3854


sunjayBhatia:
Oops forgot to fix up the CI bits

sunjayBhatia:
hmm, flake in the gateway tests

sunjayBhatia:
opened https://github.com/projectcontour/contour/pull/3855 for the gateway test flake

Makefile
```diff
@@ -343,11 +343,12 @@ install-contour-release: | setup-kind-cluster ## Install the release version of
 	./test/scripts/install-contour-release.sh $(CONTOUR_UPGRADE_FROM_VERSION)
 
 .PHONY: e2e
-e2e: | setup-kind-cluster run-e2e cleanup-kind ## Run E2E tests against a real k8s cluster
+e2e: | setup-kind-cluster load-contour-image-kind run-e2e cleanup-kind ## Run E2E tests against a real k8s cluster
```
stevesloka, line 346: I think for this to work, we'll also need to build the image locally, or have different make tasks to specify which (e.g. `make e2e` & `make e2e-incluster` or something). 

I got the following testing locally: 
```
./test/scripts/kind-load-contour-image.sh
ls: /Users/slokas/gdev/projectcontour/contour/image/contour-*.tar: No such file or directory
ERROR: stat : no such file or directory
make: *** [load-contour-image-kind] Error 1
```

stevesloka (changes requested):

Makefile
sunjayBhatia: yeah this might need some changes, i want to selectively build and load the image if you don't have it locally, or maybe you can opt out of something

CI should use the common image built in the pipeline `build-image` step, but locally probably a different story

stevesloka: Maybe not default the `CONTOUR_E2E_IMAGE` variable? Then if not present it's an out of cluster build? Or just make it explicit if you want incluster or out of cluster.

PR #3852, opened 2021-06-30
internal/xdscache: ignore irrelevant Endpoints

### LINKED ISSUE #3782: Unrelated endpoint changes trigger ClusterLoadAssignment change and push
I used the 1.15 contour example.yaml to reproduce the issue, the only changed thing is adding --debug to the contour serve process, and changing "--log-level debug" for envoy.

I installed 2 services to the cluster both tied with a separate deployment with one pod.
- relevant-service
- other-service
and created the simplest httpproxy I could think of sending traffic of example.com to the relevant-service.

So at this point the setup is simple and contour in theory only cares about the relevant-service, traffic is fine.
Next I follow the logs of one contour pod with **kubectl logs -f** and at the same time I do
**kubectl edit deployment other-deployment**, where I change the replica count to 2. 

This creates a new pod, and a new endpoint on a service that in theory contour does not have anything to do with.

To my surprise I see the following in the contour logs:
time="2021-06-08T08:17:11Z" level=debug msg="handling v3 xDS resource request" connection=12 context=xds node_id=envoy-z8wxd node_version=v1.18.3 resource_names="[**default/relevant-service/http]**" response_nonce=18 type_url=type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment version_info=18
time="2021-06-08T08:17:11Z" level=debug msg="handling v3 xDS resource request" connection=9 context=xds node_id=envoy-48brv node_version=v1.18.3 resource_names="[**default/relevant-service/http**]" response_nonce=18 type_url=type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment version_info=18
time="2021-06-08T08:17:11Z" level=debug msg="handling v3 xDS resource request" connection=10 context=xds node_id=envoy-r495t node_version=v1.18.3 resource_names="[**default/relevant-service/http**]" response_nonce=18 type_url=type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment version_info=18

Contour triggers and sends the endpoint list to all connected envoy pods.

This on a busy cluster does happen a lot of times, as this is triggered by any endpoint in any namespace especially when most pods do have failing readiness probes.
This multiplies by the dinamyc clusters in contour so for every backend service, and also multiplies by the connected envoys.
This consumes network traffic, and resource usage in both contour and envoy.

I would expect no endpoint push triggered by unrelated endpoint changes.

**Environment:**

- Contour version: 1.15.1
- Kubernetes version: (use `kubectl version`): 1.21

skriss:
In the EndpointsTranslator, do not recompute ClusterLoadAssignments
or notify cache observers if the Endpoints being added/updated/deleted
is not used by a ServiceCluster in the DAG.

Closes #3782.

### DIFF
--- a/internal/xdscache/v3/endpointstranslator.go
+++ b/internal/xdscache/v3/endpointstranslator.go
@@ -347,7 +347,6 @@ func (e *EndpointsTranslator) OnAdd(obj interface{}) {
 	switch obj := obj.(type) {
 	case *v1.Endpoints:
 		if !e.cache.UpdateEndpoint(obj) {
-			e.WithField("endpoint", k8s.NamespacedNameOf(obj)).Debug("Endpoint is not in use by a ServiceCluster, not recalculating ClusterLoadAssignments")
 			return
 		}
 
@@ -385,7 +384,6 @@ func (e *EndpointsTranslator) OnUpdate(oldObj, newObj interface{}) {
 		}
 
 		if !e.cache.UpdateEndpoint(newObj) {
-			e.WithField("endpoint", k8s.NamespacedNameOf(newObj)).Debug("Endpoint is not in use by a ServiceCluster, not recalculating ClusterLoadAssignments")
 			return
 		}
 
@@ -404,7 +402,6 @@ func (e *EndpointsTranslator) OnDelete(obj interface{}) {
 	switch obj := obj.(type) {
 	case *v1.Endpoints:
 		if !e.cache.DeleteEndpoint(obj) {
-			e.WithField("endpoint", k8s.NamespacedNameOf(obj)).Debug("Endpoint is not in use by a ServiceCluster, not recalculating ClusterLoadAssignments")
 			return
 		}
 
--- a/internal/xdscache/v3/endpointstranslator.go
+++ b/internal/xdscache/v3/endpointstranslator.go
@@ -188,8 +188,9 @@ func (c *EndpointsCache) SetClusters(clusters []*dag.ServiceCluster) error {
 
 // UpdateEndpoint adds ep to the cache, or replaces it if it is
 // already cached. Any ServiceClusters that are backed by a Service
-// that ep belongs become stale.
-func (c *EndpointsCache) UpdateEndpoint(ep *v1.Endpoints) {
+// that ep belongs become stale. Returns a boolean indicating whether
+// any ServiceClusters use ep or not.
+func (c *EndpointsCache) UpdateEndpoint(ep *v1.Endpoints) bool {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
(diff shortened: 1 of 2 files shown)


skriss:
I think this should solve this issue but I'd prefer to hold this to 1.18 to give it some more bake time.

internal/xdscache/v3/endpointstranslator.go
```diff
@@ -338,7 +346,12 @@ func equal(a, b map[string]*envoy_endpoint_v3.ClusterLoadAssignment) bool {
 func (e *EndpointsTranslator) OnAdd(obj interface{}) {
 	switch obj := obj.(type) {
 	case *v1.Endpoints:
-		e.cache.UpdateEndpoint(obj)
+		if !e.cache.UpdateEndpoint(obj) {
+			e.WithField("endpoint", k8s.NamespacedNameOf(obj)).Debug("Endpoint is not in use by a ServiceCluster, not recalculating ClusterLoadAssignments")
```
skriss, line 350: this might be too noisy to ultimately leave in the log, but useful for testing purposes.

skriss:
Well, since the 1.17 release is gonna be a few days later than originally planned, if other folks are comfortable with this then I think we can get it into 1.17

skriss:
I've manually tested this by creating, deleting, and scaling up/down Deployments, some of which are used by an HTTPProxy and some of which are not, and confirming that changing ones that are used result in updates, while changing ones that are not used do not. So unless anyone else sees any weird corner cases, I think this is in good shape.

sunjayBhatia (approved):
Yeah did a similar manual test and looked at logs from Envoy/Contour etc. looks like things are working with this change! Code changes and tests look good to me from a quick-ish read

The debug logs are super useful but yeah could be noisy and maybe have a perf impact in a cluster with a lot of services and a lot of endpoint churn? maybe not something to prematurely optimize but maybe something to keep an eye on

internal/xdscache/v3/endpointstranslator.go
youngnick: Yeah, does seem like it could get very busy indeed in a busy cluster, but it definitely seems useful to be able to have this for a bit at least.

Maybe we need to consider a metric instead?

youngnick (approved):
LGTM, this is nice.

I think that we need to seriously consider emitting some metrics for these debug logs rather than just logging, as our logs are getting very spammy, even by debug log standards. That can be addressed in a future issue though.

internal/xdscache/v3/endpointstranslator.go
skriss: I dropped the negative case log where the Endpoints is not relevant, since that's probably the noisiest and we can infer if there's no log that it was not relevant. I left the positive case log in at debug level.

Issue #3800, opened 2021-06-11
HTTPProxy with tcpproxy and routes results in 301 moved permanently


tsaarni:
The description of `HTTPProxy.spec.routes` says

> If TCPProxy is present, Routes is ignored.

But this is not fully true currently.  The code processes routes regardless, and it can be even observed externally.  If I define `HTTPProxy` with `tcpproxy` and `routes`

```yaml
apiVersion: projectcontour.io/v1
kind: HTTPProxy
metadata:
  name: protected
spec:
  virtualhost:
    fqdn: protected.127-0-0-101.nip.io
    tls:
      secretName: ingress
  routes:
    - services:
        - name: echoserver-1
          port: 80
  tcpproxy:
    services:
      - name: echoserver-2
        port: 80
```

and I make request to `http://protected.127-0-0-101.nip.io/` the result is `HTTP/1.1 301 Moved Permanently`
with `location: https://protected.127-0-0-101.nip.io/` header set in the response.  On the other hand, if I remove the `routes` from the spec, then the result is `HTTP/1.1 404 Not Found`  (or if this was the only `HTTPProxy`, then HTTP listener is removed completely).

I looked up the code and found that [computeHTTPProxy()](https://github.com/projectcontour/contour/blob/caedf100b83d37b83650a4322f62da6d115a2843/internal/dag/httpproxy_processor.go#L304) does continue to proces `routes`, even after `tcpproxy` was recognized - it does not ignore routes.   The resulting routes will not be added to secure virtualhost, but they are added to insecure virtualhost, which will result in the HTTPS upgrade as shown above.  The redirect is of course not working.

The above behavior seems unintentional to me. Couldn't we just stop processing routes completely in this case? On the other hand, there are some signs in [tcpproxy_test.go](https://github.com/projectcontour/contour/blob/caedf100b83d37b83650a4322f62da6d115a2843/internal/featuretests/v3/tcpproxy_test.go#L346) that seem to suggest otherwise, though I did not fully understand that test. So maybe there is something I missed.

tsaarni:
I bumped in the above since I wanted to come back to implement the easiest parts of [httpproxy-conflict.md](https://github.com/projectcontour/contour/blob/main/design/httpproxy-conflict.md) (missing include, missing service). While looking impacts to tests, I've found this issue.

skriss:
Would be good to get @youngnick or @stevesloka's input here, I'm not 100% sure if this is intentional or not.

[...]

PR #3853, opened 2021-06-30
WIP: internal/dag: routes in tcpproxy should result in error

tsaarni:
`HTTPProxy` spec says *"If TCPProxy is present, Routes is ignored"* but this has not been true.  Routes are processed and HTTPS upgrade is added for given routes in HTTP listener (and only HTTP listener).  In this case request for such route will get `301 Moved Permanently` response.

If routes field does not exist, then HTTP listener may be left unconfigured unless it is configured due to some other `HTTPProxy`.  User will get either connection refused or `404 Not Found` if connecting to insecure port and making request to virtualhost with `Spec.TCPProxy`.

The current behavior also has meant that user has needed to define routes (against documented behavior) if they need HTTPS upgrade for their tcpproxy.

This change is the first phase aiming to conform with the document.

This change will:

* Raise error condition if `Spec.Routes` are given (but continue processing them)
* Raise error condition if `Spec.Includes` are given (but continue processing them)
* Ensure that there will be also HTTP listener for the tcpproxy with a catch-all route `/` that will unconditionally trigger the HTTPS upgrade.

The last bullet implies that all tcpproxies will get `301 Moved Permanently` in the insecure listener. This may be undesirable if tcpproxy is handling some other protocol on top of TLS than HTTP, or even for HTTP use case if user wants to avoid the surface area of an insecure listener in Envoy.

Updates #3800

### DIFF
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -7919,6 +7919,7 @@ func TestDAGInsert(t *testing.T) {
 				cert1, proxy19, s1, sec1,
 			},
 			want: listeners(
+				httpsRedirectListener("example.com"),
 				&Listener{
 					Port: 443,
 					VirtualHosts: virtualhosts(
@@ -8024,6 +8025,7 @@ func TestDAGInsert(t *testing.T) {
 		"insert httpproxy w/ tcpproxy w/ includes another root": {
 			objs: []interface{}{proxy38, proxy39, s1},
 			want: listeners(
+				httpsRedirectListener("www.example.com"),
 				&Listener{
 					Port: 443,
 					VirtualHosts: virtualhosts(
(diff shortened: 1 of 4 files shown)


tsaarni:
Submitting this PR as draft / WiP for now, so that it exists but I'm bit unsure about this: there were several test cases that specifically test `routes` in combination of `tcpproxy` so it seems pretty intentional, even though it is against the documented behavior. 

When following issues that sometimes were linked in the test cases, it seemed the need for routes was always just about getting HTTPS upgrade for tcpproxy, which gives some confidence this change could be done, unless users rely on fancier routes for their HTTPS upgrade than the catch-all `/` route that this change adds automatically.

youngnick:
Hmm, I think there's still discussion we need to have here, so I'll reopen this one.

PR #3869, opened 2021-07-06
hack: Change ginkgo install var to use CI

stevesloka:
The install-kubernetes-toolchain.sh installs a few dependencies in order to
build, test, and lint Contour code. The env var to install gingkgo was using
OSType=Linux which breaks on a local linux machine.

This changes that variable to use "CI" which is setup by Github Actions to let the
script know if it's running in Github Actions or not.

### DIFF
--- a/hack/actions/install-kubernetes-toolchain.sh
+++ b/hack/actions/install-kubernetes-toolchain.sh
@@ -10,6 +10,10 @@ readonly KIND_VERS="v0.11.1"
 readonly SONOBUOY_VERS="0.19.0"
 readonly KUBEBUILDER_VERS="3.1.0"
 
+# The CI env variable is set by Github Actions when running jobs.
+# This is used to know how to configure specific GithubAction jobs.
+readonly CI="false"
+
 # Envtest Binaries Manager is required for newer versions. See the following for details:
 # https://github.com/projectcontour/contour/issues/3832
 readonly KUBEBUILDER_TOOLS_VERS="1.19.2"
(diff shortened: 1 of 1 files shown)


sunjayBhatia (approved):
LGTM, might steal this idea to use later in the CI bits of https://github.com/projectcontour/contour/pull/3851 as well

sunjayBhatia (changes requested):
looks like CI might still not be happy about this, seeing that ginkgo is missing

stevesloka:
Ahh turns out `CI` was false when the docs said otherwise. I did find a var named `GTIHUB_ACTIONS` which I just switched this to verify it works.

hack/actions/install-kubernetes-toolchain.sh
```diff
@@ -10,6 +10,10 @@ readonly KIND_VERS="v0.11.1"
 readonly SONOBUOY_VERS="0.19.0"
 readonly KUBEBUILDER_VERS="3.1.0"
 
+# The CI env variable is set by Github Actions when running jobs.
```
sunjayBhatia, line 13: this section seems removable?

sunjayBhatia (changes requested):

skriss (commented):
LGTM once @sunjayBhatia's last comment is addressed

hack/actions/install-kubernetes-toolchain.sh
stevesloka: I think I had that since I was running this script locally to install some deps. But we should approach that a different way I think.

sunjayBhatia (approved):

skriss (approved):

PR #3876, opened 2021-07-08
internal/validation: allow GatewayClass.Spec.ParametersRef to be specified

### LINKED ISSUE #3874: GatewayClass parametersRef being not-nil should mean the value is ignored, not the GatewayClass rejected

When a GatewayClass is found using the controllerName parameter, currently if the `spec.parametersRef` field is non-nil the GatewayClass is rejected for processing by Contour. This is because Contour cannot use the parametersRef field for anything. 

However, this breaks the Operator's Gateway support, because the Operator sets the parametersRef field to the Contour object the Operator uses to configure the install that is created when a Gateway of the GatewayClass is created. 

I get it, this is complex, I'll explain more. However, the tl;dr is that we should be ignoring `paramtersRef`, not rejecting based on it.

When you create a Gateway, the Gateway has to reference an owning GatewayClass. Here's a sample Gateway:

```yaml
kind: Gateway
metadata:
  name: contour
  namespace: contour-test
spec:
  gatewayClassName: sample-gatewayclass
  listeners:
    - protocol: HTTP
      port: 80
      routes:
        kind: HTTPRoute
        selector:
          matchLabels:
            app: kuard
```

The `gatewayClassName` field refers to a GatewayClass called `sample-gatewayclass`. Now, as far as Contour is concerned, this is just a bucket that holds Gateways (currently), and is only relevant because Contour needs to find the `spec.controller` field that is part of the GatewayClass and ensure it matches the one configured in Contour's config file via the `gateway.controllerName` parameter (newly introduced in v1.17.0).

Here is a sample GatewayClass the Operator would use to create a Contour installation to implement that Gateway:
```yaml
kind: GatewayClass
metadata:
  name: sample-gatewayclass
spec:
  controller: projectcontour.io/contour-operator
  parametersRef:
    group: operator.projectcontour.io
    kind: Contour
    scope: Namespace
    name: contour-gateway-test
    namespace: contour-operator
```
When creating a Contour, the Operator will (after projectcontour/contour-operator#400 is fixed) pass a `gateway.controllerName` field to the Contour install it creates of `projectcontour.io/contour-operator`.

However, for the Operator to be *able* to create that Contour install, it expects to have the `parametersRef` field set to the Projectcontour CRD `Contour`, which enables someone to configure how the Operator will configure the Contour installation it creates.

Currently, if that `parametersRef` field is anything other than unset or `nil`, then Contour will refuse to use the GatewayClass, and so the Operator's Gateway support will fail.

However, there's no need for Contour to reject the GatewayClass if that field is set, it can just silently ignore it. Once we do that, the Gateway support should work.

The Operator to be able to create a Contour install that references a GatewayClass and Gateway and have everything work.

**Environment:**

- Contour version: v1.17.0

stevesloka:
The field parametersRef on a GatewayClass should be able to be defined.

Fixes #3874

### DIFF
--- a/internal/controller/gatewayclass_test.go
+++ b/internal/controller/gatewayclass_test.go
@@ -26,6 +26,7 @@ import (
 )
 
 var _ = Describe("GatewayClass Controller", func() {
+
 	Context("Managed GatewayClass", func() {
 		It("Should surface admitted status", func() {
 			key := types.NamespacedName{Name: "test-gatewayclass-" + rand.String(10)}
@@ -174,11 +175,11 @@ var _ = Describe("GatewayClass Controller", func() {
 		})
 	})
 
-	Context("Managed GatewayClass", func() {
-		It("With parameterRefs should not be admitted", func() {
+	Context("GatewayClass with parametersRef", func() {
+		It("With parametersRefs should be admitted", func() {
 			key := types.NamespacedName{Name: "test-gatewayclass-" + rand.String(10)}
 
-			unsupported := &gatewayv1alpha1.GatewayClass{
+			admitted := &gatewayv1alpha1.GatewayClass{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      key.Name,
 					Namespace: key.Namespace,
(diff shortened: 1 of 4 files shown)


sunjayBhatia (approved):

youngnick (approved):
LGTM, thanks.

skriss:
adjusted the PR title to be accurate -- would be good to have the squashed commit message match as well when merged.

skriss (approved):

youngnick:
Putting `do-not-merge` on this one until we've had a chance to agree on the end plan (although I think this one is pretty uncontroversial anyway).

sunjayBhatia:
I think `do not merge` can be removed and along with https://github.com/projectcontour/contour-operator/pull/401 should represent Operator+Contour not conflicting for now

PR #3878, opened 2021-07-08
notify slack channel if failures on main builds

stevesloka:

### DIFF
--- a/.github/workflows/build_main.yaml
+++ b/.github/workflows/build_main.yaml
@@ -5,6 +5,8 @@ on:
     branches:
       - main
 
+env:
+  SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/build_main.yaml
+++ b/.github/workflows/build_main.yaml
@@ -31,4 +31,4 @@ jobs:
          status: ${{ job.status }}
          steps: ${{ toJson(steps) }}
          channel: '#contour-ci-notifications'
-       if: ${{ failure() && github.head_ref == 'refs/heads/main' }}
+       if: failure()
--- a/.github/workflows/build_main.yaml
+++ b/.github/workflows/build_main.yaml
@@ -26,3 +26,9 @@ jobs:
         echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
         make multiarch-build
         docker logout
+     - uses: act10ns/slack@v1
+       with:
+         status: ${{ job.status }}
+         steps: ${{ toJson(steps) }}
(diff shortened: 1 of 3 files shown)


.github/workflows/build_main.yaml
```diff
@@ -26,3 +26,9 @@ jobs:
         echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
         make multiarch-build
         docker logout
+     - uses: act10ns/slack@v1
+       with:
+         status: ${{ job.status }}
+         steps: ${{ toJson(steps) }}
+         channel: '#contour-ci-notifications'
+       if: ${{ failure() && github.head_ref == 'refs/heads/main' }}
```
sunjayBhatia, line 34: nit: might be ok to remove the branch check for this one since its always on main

sunjayBhatia (approved):

.github/workflows/prbuild.yaml
```diff
@@ -24,6 +24,12 @@ jobs:
       with:
         name: image
         path: image/contour-*.tar
+    - uses: act10ns/slack@v1
+      with:
+        status: ${{ job.status }}
+        steps: ${{ toJson(steps) }}
+        channel: '#contour-ci-notifications'
+      if: ${{ failure() && github.head_ref == 'refs/heads/main' }}
```
sunjayBhatia, line 32: turns out `github.ref` is the thing to use for `main` builds (worked on operator)

sunjayBhatia (changes requested):

.github/workflows/prbuild.yaml
sunjayBhatia: docs: https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context

.github/workflows/build_main.yaml
stevesloka: yeah copy/paste just for consistency. I can take out since it's really not needed.

.github/workflows/prbuild.yaml
stevesloka: 👍

youngnick (approved):
This seems great, although we should look at moving this to a different Slack instance at some point. It's more important we have some signal first though.

sunjayBhatia (approved):

skriss:
Do we need to explicitly set `SLACK_WEBHOOK_URL` in each of these  like in https://github.com/projectcontour/contour-operator/blob/main/.github/workflows/prbuild.yaml#L13?

stevesloka:
Yup I think you're right @skriss, I added the env var to all the jobs.

skriss (approved):
cool, I think this looks good to go!

skriss:
Got that OS X flake again

Issue #3865, opened 2021-07-06
Add internal/controller/gateway.go tests


stevesloka:
Need to validate the internal/controller/gateway.go logic with appropriate tests. These were missing from #3858.

skriss:
Ended up going with E2E's to test the GatewayClass/Gateway controller logic. This can be closed out now.

PR #3881, opened 2021-07-09
Deprecate gateway name/namespace from config file + internal packages

### LINKED ISSUE #3862: Deprecate gateway name/namespace from config file + internal packages
The Gateway Name + Namespace have been deprecated in favor of the GatewayClass.ControllerName field. When #3858 merges, ControllerName is actually used to look up the Gateway, so these fields can be fully deprecated and removed.

stevesloka:
Fixes #3862

### DIFF
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -402,7 +402,11 @@ func (p *GatewayAPIProcessor) namespaceMatches(namespaces *gatewayapi_v1alpha1.R
 	case gatewayapi_v1alpha1.RouteSelectAll:
 		return true, nil
 	case gatewayapi_v1alpha1.RouteSelectSame:
-		return p.source.ConfiguredGateway.Namespace == namespace, nil
+		gatewayNSName := types.NamespacedName{}
+		if p.source.gateway != nil {
+			gatewayNSName = k8s.NamespacedNameOf(p.source.gateway)
+		}
+		return gatewayNSName.Namespace == namespace, nil
 	case gatewayapi_v1alpha1.RouteSelectSelector:
 		if len(namespaces.Selector.MatchLabels) == 0 && len(namespaces.Selector.MatchExpressions) == 0 {
 			return false, fmt.Errorf("RouteNamespaces selector must be specified when `RouteSelectType=Selector`")
@@ -434,17 +438,22 @@ func (p *GatewayAPIProcessor) gatewayMatches(routeGateways *gatewayapi_v1alpha1.
 		return true
 	}
 
+	gatewayNSName := types.NamespacedName{}
+	if p.source.gateway != nil {
+		gatewayNSName = k8s.NamespacedNameOf(p.source.gateway)
+	}
+
 	switch *routeGateways.Allow {
 	case gatewayapi_v1alpha1.GatewayAllowAll:
 		return true
 	case gatewayapi_v1alpha1.GatewayAllowFromList:
 		for _, gateway := range routeGateways.GatewayRefs {
-			if gateway.Name == p.source.ConfiguredGateway.Name && gateway.Namespace == p.source.ConfiguredGateway.Namespace {
+			if gateway.Name == gatewayNSName.Name && gateway.Namespace == gatewayNSName.Namespace {
 				return true
 			}
 		}
 	case gatewayapi_v1alpha1.GatewayAllowSameNamespace:
-		return p.source.ConfiguredGateway.Namespace == namespace
+		return gatewayNSName.Namespace == namespace
 	}
 	return false
 }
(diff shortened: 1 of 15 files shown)


stevesloka:
Need to add tests in https://github.com/projectcontour/contour/issues/3865 to validate the right Gateway is selected.

internal/dag/gatewayapi_processor.go
```diff
@@ -398,7 +398,11 @@ func (p *GatewayAPIProcessor) namespaceMatches(namespaces *gatewayapi_v1alpha1.R
 	case gatewayapi_v1alpha1.RouteSelectAll:
 		return true, nil
 	case gatewayapi_v1alpha1.RouteSelectSame:
-		return p.source.ConfiguredGateway.Namespace == namespace, nil
+		gatewayNSName := types.NamespacedName{}
+		if p.source.gateway != nil {
```
sunjayBhatia, line 402: maybe we should just bail out of the processor early and log if there is no gateway present so we don't have to keep doing these checks, it is invalid to generate config if there is no gateway anyways

sunjayBhatia (changes requested):

sunjayBhatia:
This change should also have an accompanying Operator change to remove the bits that set a Gateway name/namespace in the the Contour ConfigMap

internal/dag/gatewayapi_processor.go
skriss: Looks like we're already doing this (https://github.com/projectcontour/contour/blob/abeefb3f941951858b8b9f11a8775d896aba7757/internal/dag/gatewayapi_processor.go#L73-L81) so yeah we can drop these redundant nil checks

sunjayBhatia:
Same as the linked operator issue:

Now that I think about it, might be required to move this to 1.19 so we don't re-break the operator/contour

right now it works ok having multiple contours b/c we're still passing the gateway name/namespace so a contour reconciles a single gateway

skriss:
> Same as the linked operator issue:
> 
> Now that I think about it, might be required to move this to 1.19 so we don't re-break the operator/contour
> 
> right now it works ok having multiple contours b/c we're still passing the gateway name/namespace so a contour reconciles a single gateway

Yep seems like it'd be better to hold this until 1.19 unless there's a strong need to get it into 1.18.

pkg/config/parameters.go
```diff
@@ -51,19 +51,6 @@ func (g *GatewayParameters) Validate() error {
 		return nil
 	}
 
-	if len(g.Name) == 0 && len(g.Namespace) == 0 && len(g.ControllerName) == 0 {
-		return nil
-	}
-
-	if len(g.Name) == 0 {
```
mike1808, line 58: this should have been removed when the deprecation notice was added

stevesloka: We ended up going slower than I anticipated on this unfortunately. This will be cleaned up in Contour v1.19.

internal/dag/gatewayapi_processor.go
stevesloka: yup good call, I dropped this one in the method and kept the global one.

sunjayBhatia:
made a draft here in the operator https://github.com/projectcontour/contour-operator/pull/422

removes the gatewayclass/gateway controllers from there so should make the operator compatible with this change

internal/dag/gatewayapi_processor.go
```diff
@@ -402,7 +402,9 @@ func (p *GatewayAPIProcessor) namespaceMatches(namespaces *gatewayapi_v1alpha1.R
 	case gatewayapi_v1alpha1.RouteSelectAll:
 		return true, nil
 	case gatewayapi_v1alpha1.RouteSelectSame:
-		return p.source.ConfiguredGateway.Namespace == namespace, nil
+		gatewayNSName := types.NamespacedName{}
```
youngnick, line 405: I guess the linter is saying that this should be a `var` declaration instead?
```suggestion
		var gatewayNSName types.NamespacedName
```

internal/dag/gatewayapi_processor.go
```diff
@@ -434,17 +436,22 @@ func (p *GatewayAPIProcessor) gatewayMatches(routeGateways *gatewayapi_v1alpha1.
 		return true
 	}
 
+	gatewayNSName := types.NamespacedName{}
```
youngnick, line 439: Probably better to have this one a `var` assignement as well, I guess.
```suggestion
	var gatewayNSName types.NamespacedName
```

internal/k8s/objectmeta.go
```diff
@@ -22,6 +22,7 @@ import (
 
 // NamespacedNameOf returns the NamespacedName of any given Kubernetes object.
 func NamespacedNameOf(obj metav1.Object) types.NamespacedName {
+
```
youngnick, line 25: Nit: ?

youngnick (approved):
LGTM with a few small nits.

sunjayBhatia (approved):
LGTM, CI/linter not happy but thats an easy fix

Should fast-follow with operator PRs to keep CI healthy once this is merged

PR #3884, opened 2021-07-12
internal/controller: Add Gateway controller tests

### LINKED ISSUE #3867: Add status tests for Gateway
Add status tests for Gateway conditions that are set from the GatewayAPI Processor.

_Originally posted by @sunjayBhatia in https://github.com/projectcontour/contour/pull/3858#discussion_r663194086_

stevesloka:
Adds initial framewwork for testing Gateway controllers. Since the
status of a Gateway is udpated after a DAG is built, we need to
spin up all the dag creation infrastructure.

Fixes #3867


stevesloka:
These have almost turned into full e2e tests FWIW. 🤷

sunjayBhatia (commented):
Maybe a different approach for these tests would be to stub out the event handler completely and just check it is called in the correct scenarios, rather than check status is set since that is the job of a different component than the controller right now

pivot to just testing that reconciliation steps happen or not for gateways with certain configuration so we reduce duplication and have tests covering just what is in this package

stevesloka:
> Maybe a different approach for these tests would be to stub out the event handler completely and just check it is called in the correct scenarios, rather than check status is set since that is the job of a different component than the controller right now

Yup I agree. This would mean dropping the stub controller-runtime tests which I'd be ok with. @skriss thoughts?

skriss:
> > Maybe a different approach for these tests would be to stub out the event handler completely and just check it is called in the correct scenarios, rather than check status is set since that is the job of a different component than the controller right now
> 
> Yup I agree. This would mean dropping the stub controller-runtime tests which I'd be ok with. @skriss thoughts?

Yeah I'd be in favor of that. I think these would otherwise end up being pretty redundant with other types of tests that we already have in Contour.

PR #3888, opened 2021-07-14
deps: update controller-runtime to v0.9.2

skriss:
Updates controller-runtime to v0.9.2 from v0.9.0-beta.1.


cmd/contour/serve.go
```diff
@@ -46,6 +46,7 @@ import (
 	xdscache_v3 "github.com/projectcontour/contour/internal/xdscache/v3"
 	"github.com/projectcontour/contour/pkg/config"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/collectors"
```
youngnick, line 49: Did updating controller-runtime also update prometheus versions?

youngnick (approved):
LGTM, seems weird the main update is in Prometheus deps though.

cmd/contour/serve.go
skriss: It did, yeah

stevesloka (approved):

cmd/contour/serve.go
skriss: FWIW, the code change is because the functions in the `prometheus` package are marked deprecated and generated a linter error, however the `collector` implementations just call back into the `prometheus` package for now, so no behavior change: https://github.com/prometheus/client_golang/blob/master/prometheus/collectors/process_collector.go#L49-L55

Issue #3892, opened 2021-07-15
E2E test flakes


skriss:
I've seen a few sporadic E2E test failures recently. Just now:
https://github.com/projectcontour/contour/runs/3080942529?check_suite_focus=true
```
• Failure in Spec Teardown (AfterEach) [6.341 seconds]
Gateway API
/home/runner/work/contour/contour/test/e2e/gateway/gateway_test.go:50
  HTTPRoute: Insecure (Non-TLS) Gateway [AfterEach]
  /home/runner/work/contour/contour/test/e2e/gateway/gateway_test.go:120
    with namespace: 007-gateway-allow-type
    /home/runner/work/contour/contour/test/e2e/framework.go:206
      with gateway 007-gateway-allow-type/http, controllerName: projectcontour.io/ingress-controller-1981
      /home/runner/work/contour/contour/test/e2e/gateway/gateway_test.go:66
        allowtype on route is respected
        /home/runner/work/contour/contour/test/e2e/gateway/007_gateway_allow_type_test.go:30

        Timed out after 1.000s.
        Expected process to exit.  It did not.

        /home/runner/work/contour/contour/test/e2e/deployment.go:511
```

I've seen another one related to the TLS Gateway when set to terminate, and a request that has a non-matching SNI that's supposed to not route, but apparently does (but again, it's a flake).

The recent ones have all been in Gateway tests, not sure if that's a coincidence or not.

skriss:
Just got another of those on main:

```
• Failure in Spec Teardown (AfterEach) [19.200 seconds]
Gateway API
/home/runner/work/contour/contour/test/e2e/gateway/gateway_test.go:50
  TLSRoute Gateway: Mode: Terminate [AfterEach]
  /home/runner/work/contour/contour/test/e2e/gateway/gateway_test.go:281
    with namespace: gateway-tlsroute-mode-terminate
    /home/runner/work/contour/contour/test/e2e/framework.go:206
      with gateway gateway-tlsroute-mode-terminate/tls-passthrough, controllerName: projectcontour.io/ingress-controller-4062
      /home/runner/work/contour/contour/test/e2e/gateway/gateway_test.go:66
        with TLS secret gateway-tlsroute-mode-terminate/tlscert for hostname tlsroute.gatewayapi.projectcontour.io
        /home/runner/work/contour/contour/test/e2e/gateway/gateway_test.go:313
          TLS requests terminate via SNI at Envoy and then are routed to a service
[...]

PR #3893, opened 2021-07-15
test/e2e: increase timeout for Contour process exiting

skriss:
Increases the timeout for waiting for the Contour process
to exit from 1s (default) to 1m, to avoid test flakes.

Updates #3892.

### DIFF
--- a/test/e2e/deployment.go
+++ b/test/e2e/deployment.go
@@ -508,6 +508,8 @@ func (d *Deployment) StartLocalContour(config *config.Parameters, additionalArgs
 }
 
 func (d *Deployment) StopLocalContour(contourCmd *gexec.Session, configFile string) error {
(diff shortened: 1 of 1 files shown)


youngnick (approved):
Agreed, better to have tests take a little longer than flakes. I guess we need to keep an eye on the overall runtime though?

skriss:
> Agreed, better to have tests take a little longer than flakes. I guess we need to keep an eye on the overall runtime though?

Yeah, for sure we should look at improving the runtime. In practice I don't think this particular step takes more than a second or two (it usually succeeds with the existing 1 second timeout). It seems like right now the biggest target for improvement is the namespace deletions in between test suites, I'll see if I can dig into that today.

stevesloka (approved):

PR #3902, opened 2021-07-20
HTTPProxy: add Spec.IngressClassName

### LINKED ISSUE #2566: HTTPProxy lacking ingressClassName field
**Problem Description** 

As of Kubernetes 1.18 the Ingress resource have been updated with specific field 
`ingressClassName`, which deprecates the previously used annotation `kubernetes.io/ingress.class`, 
see: [Kubernetes Blog](https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/).
The change comes with a new IngressClass resource as well.

HTTPproxy resource still relies on the annotation for selecting an ingress class.
From a user perspective this makes it complicated using both Ingress and HTTPproxy as it is hard to know which one to use.
Also, the HTTPproxy relying on an annotation deprecated in Kubernetes in itself causes a risk of confusion regarding HTTPproxy resource relevance.

**Suggestion**

HTTPproxy object should implement similar IngressClassName field as found in Ingress resource.
This would align the usage of both resources and will make sure HTTPproxy evolves together with Ingress.

skriss:
Adds a `Spec.IngressClassName` field to `HTTPProxy` following
the same rules/precedence relative to annotations as for `Ingress`.

Closes #2566.

For reviewers: the first commit consolidates all existing ingress class related logic into a dedicated `internal/ingressclass` package. The second commit adds the new field to `HTTPProxy` and makes all relevant changes for it.


skriss:
TODO docs, looking to see where it would make sense to cover

sunjayBhatia (approved):
nice refactor 👍🏽 I think this looks all correct to me

skriss:
Added some more words to the godoc and annotations.md; the latter isn't ideal since this is *not* an annotation, but I think it made the most sense given the current structure. If/when we actually retire the annotation, we can do some restructuring.

youngnick (approved):
LGTM, making the ingressclass package is a great idea, nice work.

PR #3904, opened 2021-07-21
internal/controller: switch to using event predicates for filtering

skriss:
Switches from using EnqueueRequestsFromMapFunc to EnqueueRequestForObject
along with a filtering predicate to control which events trigger reconcile
requests. This is a cleanup only, functionality should not change.

IMHO this is more idiomatic/logical for what we're doing here and is easier to read.


sunjayBhatia (approved):
+1 from me, makes things a bit cleaner

stevesloka (approved):

PR #3916, opened 2021-07-23
Add pkg to coverage check

sunjayBhatia:
Should be considered when checking coverage


skriss (approved):

youngnick (approved):
Good catch, we all missed it before.

Issue #3905, opened 2021-07-21
deleting a Gateway doesn't remove it from the DAG, still allows HTTPRoutes to be programmed against it


skriss:
**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

Create Gateway, confirm it's admitted
Delete Gateway, confirm it's gone from the API
Create HTTPRoute using the deleted Gateway
Get route programmed/accessible over HTTP

**What did you expect to happen:**
When the Gateway is deleted from the API, it should be removed from the DAG and Routes should no longer be able to bind to it

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

[This code section](https://github.com/projectcontour/contour/blob/main/internal/controller/gateway.go#L120-L124) calls the `OnDelete` event, however I believe since we dropped the finalizers this is no longer valid -- instead, when a Gateway is deleted, it triggers [this code branch](https://github.com/projectcontour/contour/blob/main/internal/controller/gateway.go#L111-L115) since it's immediately removed from the API. So we should move calling the OnDelete event handler in to that branch.

Note that if/when we re-add finalizers, then any finalizer logic would have to go within a conditional that checks for a non-zero deletion timestamp.

The same goes for GatewayClass since the controller code is very similar.

youngnick:
Oof, great catch, thanks @skriss.

skriss:
I'll try to get this done for 1.18, should be an easy fix 🤞

skriss:
xref #3526

skriss:
This was closed by #3923.

PR #3923, opened 2021-07-27
internal/controller: fix Gateway/GatewayClass deletion bug

skriss:
Fixes a bug where when Gateways/GatewayClasses were deleted, they
were not being removed from the DAG.

Updates #3905.

### DIFF
--- a/internal/controller/gateway.go
+++ b/internal/controller/gateway.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


skriss:
I've manually verified that deleting a GatewayClass or Gateway now removes them from the DAG and removes any associated listeners/routes from Envoy. We don't really have a good structure set up right now for testing this, so I didn't add any tests, but let me know if you think it's important. It'd probably be easiest as an E2E that directly handles Gateway creation/deletion.

sunjayBhatia (approved):
if we got [this pr](https://github.com/projectcontour/contour/pull/3884) in we could probably add some tests

this kind of thing i think would be good to test like i mentioned [here](https://github.com/projectcontour/contour/pull/3884#pullrequestreview-710827332) rather than the structure of that pr and the other controller tests at the moment (so we dont have to repeat a lot of test coverage)

skriss:
> if we got [this pr](https://github.com/projectcontour/contour/pull/3884) in we could probably add some tests
> 
> this kind of thing i think would be good to test like i mentioned [here](https://github.com/projectcontour/contour/pull/3884#pullrequestreview-710827332) rather than the structure of that pr and the other controller tests at the moment (so we dont have to repeat a lot of test coverage)

Yeah that sounds good to me

youngnick (approved):

PR #3928, opened 2021-07-28
Documentation updates for 1.18 release

sunjayBhatia:


site/content/resources/upgrading.md
```diff
@@ -9,9 +9,68 @@ This document describes the changes needed to upgrade your Contour installation.
...
+```bash
+$ kubectl delete namespace projectcontour
+$ kubectl apply -f {{< param base_url >}}/quickstart/v1.18.0/contour.yaml
+```
+
+This will remove the Envoy and Contour pods from your cluster and recreate them with the updated configuration.
+If you're using a `LoadBalancer` Service, (which most of the examples do) deleting and recreating may change the public IP assigned by your cloud provider.
+You'll need to re-check where your DNS names are pointing as well, using [Get your hostname or IP address][12].
+
+### The less easy way
```
skriss, line 41: there was at least one CRD change as well to add `HTTPProxy.Spec.IngressClassName` so should reapply those too.

site/content/resources/compatibility-matrix.md
```diff
@@ -11,6 +11,7 @@ These combinations of versions are specifically tested in CI and supported by th
 | Contour Version | Envoy Version        | Kubernetes Versions | Operator Version |
 | --------------- | :------------------- | ------------------- | ---------------- |
 | main            | [1.19.0][14]         | 1.21, 1.20, 1.19    | [main][50]       |
+| 1.18.1          | [1.19.0][14]         | 1.21, 1.20, 1.19    | [1.18.0][61]       |
```
skriss, line 14: could edit L41 to drop "It will be fixed in the 1.18 release." now

site/content/resources/upgrading.md
sunjayBhatia: good catch!

site/content/resources/upgrading.md
```diff
@@ -9,9 +9,68 @@ This document describes the changes needed to upgrade your Contour installation.
 
 <div id="toc" class="navigation"></div>
 
-## Upgrading Contour 1.17.0 to 1.17.1
+## Upgrading Contour 1.17.1 to 1.18.0
```
stevesloka, line 12: Should we have a big call out about the externalName thing just so folks are aware that upgrading will disable all those unless the config option is set?

sunjayBhatia: good idea, I'll add a link to the CVE page

stevesloka (approved):

PR #3929, opened 2021-07-28
cmd/contour: Add admin-address flag to shutdown-manager

stevesloka:
The shutdown-manager had no way of specifying the ip address of the Envoy Admin interface.
That address can be customized via the admin-address flag on the bootstrap command.

### DIFF
--- a/cmd/contour/shutdownmanager.go
+++ b/cmd/contour/shutdownmanager.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"net"
 	"net/http"
 	"os"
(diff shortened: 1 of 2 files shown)


stevesloka:
Going to run some smoke tests with a big cluster now to validate further.

stevesloka:
Going to hold on this while investigating the socket approach.

sunjayBhatia (commented):
just flushing a review on this to clear my queue, i think this is probably closable since the socket approach should work well

stevesloka:
Closing this since the work in #3934 supersedes this work.

PR #3933, opened 2021-07-29
internal/envoy: fix TCPProxy weights bug

skriss:
Fixes a bug where when a TCPProxy had more than one
service defined, any service with a zero weight was
updated to have a weight of 1. The correct behavior
is that if all service have no weights, then they
should be evenly weighted, but if some have non-zero
weights and others have zero weights, they stay as-is.

### DIFF
--- a/internal/envoy/v3/listener.go
+++ b/internal/envoy/v3/listener.go
@@ -479,16 +479,23 @@ func TCPProxy(statPrefix string, proxy *dag.TCPProxy, accesslogger []*accesslog.
 		}
 	default:
 		var clusters []*tcp.TcpProxy_WeightedCluster_ClusterWeight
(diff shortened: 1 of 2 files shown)


skriss:
Note that the desired behavior that I describe in the commit message is not explicitly documented anywhere for `TCPProxy`, however it is for `HTTPProxy` (and my assumption is that the behavior should be the same between them): https://projectcontour.io/docs/v1.18.0/config/request-routing/#upstream-weighting

skriss:
Also FYI, this came up while working on supporting weighted forward to's for `TLSRoute` (https://github.com/projectcontour/contour/issues/3440).

skriss:
Looks like this is not quite right, closing for now.

PR #3934, opened 2021-07-29
use unix socket for Envoy admin webpage

stevesloka:
- Moves the Envoy admin interface to use a unix socket
- Creates new listener from Contour to expose read only information
  from Envoy admin page over port 9001
- Updates shutdown-manager to use new unix socket to begin Envoy
  draining procedure

### DIFF
--- a/cmd/contour/bootstrap.go
+++ b/cmd/contour/bootstrap.go
@@ -26,7 +26,7 @@ func registerBootstrap(app *kingpin.Application) (*kingpin.CmdClause, *envoy.Boo
 	bootstrap := app.Command("bootstrap", "Generate bootstrap configuration.")
 	bootstrap.Arg("path", "Configuration file ('-' for standard output).").Required().StringVar(&config.Path)
 	bootstrap.Flag("resources-dir", "Directory where configuration files will be written to.").StringVar(&config.ResourcesDir)
-	bootstrap.Flag("admin-address", "Envoy admin interface address.").StringVar(&config.AdminAddress)
+	bootstrap.Flag("admin-address", "Envoy admin interface address.").Default("/admin/admin.sock").StringVar(&config.AdminAddress)
 	bootstrap.Flag("admin-port", "Envoy admin interface port.").IntVar(&config.AdminPort)
 	bootstrap.Flag("xds-address", "xDS gRPC API address.").StringVar(&config.XDSAddress)
 	bootstrap.Flag("xds-port", "xDS gRPC API port.").IntVar(&config.XDSGRPCPort)
--- a/cmd/contour/bootstrap.go
+++ b/cmd/contour/bootstrap.go
@@ -27,7 +27,7 @@ func registerBootstrap(app *kingpin.Application) (*kingpin.CmdClause, *envoy.Boo
 	bootstrap.Arg("path", "Configuration file ('-' for standard output).").Required().StringVar(&config.Path)
 	bootstrap.Flag("resources-dir", "Directory where configuration files will be written to.").StringVar(&config.ResourcesDir)
 	bootstrap.Flag("admin-address", "Envoy admin interface address.").Default("/admin/admin.sock").StringVar(&config.AdminAddress)
-	bootstrap.Flag("admin-port", "DEPRECATED: Envoy admin interface port.").Hidden().IntVar(&config.AdminPort)
+	bootstrap.Flag("admin-port", "DEPRECATED: Envoy admin interface port.").IntVar(&config.AdminPort)
 	bootstrap.Flag("xds-address", "xDS gRPC API address.").StringVar(&config.XDSAddress)
 	bootstrap.Flag("xds-port", "xDS gRPC API port.").IntVar(&config.XDSGRPCPort)
 	bootstrap.Flag("envoy-cafile", "CA Filename for Envoy secure xDS gRPC communication.").Envar("ENVOY_CAFILE").StringVar(&config.GrpcCABundle)
--- a/cmd/contour/bootstrap.go
+++ b/cmd/contour/bootstrap.go
@@ -27,7 +27,7 @@ func registerBootstrap(app *kingpin.Application) (*kingpin.CmdClause, *envoy.Boo
 	bootstrap.Arg("path", "Configuration file ('-' for standard output).").Required().StringVar(&config.Path)
 	bootstrap.Flag("resources-dir", "Directory where configuration files will be written to.").StringVar(&config.ResourcesDir)
 	bootstrap.Flag("admin-address", "Envoy admin interface address.").Default("/admin/admin.sock").StringVar(&config.AdminAddress)
-	bootstrap.Flag("admin-port", "Envoy admin interface port.").IntVar(&config.AdminPort)
+	bootstrap.Flag("admin-port", "DEPRECATED: Envoy admin interface port.").Hidden().IntVar(&config.AdminPort)
 	bootstrap.Flag("xds-address", "xDS gRPC API address.").StringVar(&config.XDSAddress)
 	bootstrap.Flag("xds-port", "xDS gRPC API port.").IntVar(&config.XDSGRPCPort)
 	bootstrap.Flag("envoy-cafile", "CA Filename for Envoy secure xDS gRPC communication.").Envar("ENVOY_CAFILE").StringVar(&config.GrpcCABundle)
--- a/cmd/contour/bootstrap.go
+++ b/cmd/contour/bootstrap.go
@@ -26,7 +26,7 @@ func registerBootstrap(app *kingpin.Application) (*kingpin.CmdClause, *envoy.Boo
 	bootstrap := app.Command("bootstrap", "Generate bootstrap configuration.")
 	bootstrap.Arg("path", "Configuration file ('-' for standard output).").Required().StringVar(&config.Path)
 	bootstrap.Flag("resources-dir", "Directory where configuration files will be written to.").StringVar(&config.ResourcesDir)
-	bootstrap.Flag("admin-address", "Envoy admin interface address.").Default("/admin/admin.sock").StringVar(&config.AdminAddress)
+	bootstrap.Flag("admin-address", "Path to Envoy admin unix domain socket.").Default("/admin/admin.sock").StringVar(&config.AdminAddress)
 	bootstrap.Flag("admin-port", "DEPRECATED: Envoy admin interface port.").IntVar(&config.AdminPort)
 	bootstrap.Flag("xds-address", "xDS gRPC API address.").StringVar(&config.XDSAddress)
 	bootstrap.Flag("xds-port", "xDS gRPC API port.").IntVar(&config.XDSGRPCPort)
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -354,7 +354,7 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		// The enableDebugListener boolean is defaulted to true which matches Contour's current behavior.
 		// Future iterations can allow for this to be customized by the user if they desire to disable the
 		// Envoy admin interface entirely.
-		xdscache_v3.NewListenerCache(listenerConfig, ctx.statsAddr, ctx.statsPort, ctx.Config.Network.EnvoyAdminPort, true),
+		xdscache_v3.NewListenerCache(listenerConfig, ctx.statsAddr, ctx.statsPort, ctx.Config.Network.EnvoyAdminPort),
 		&xdscache_v3.SecretCache{},
 		&xdscache_v3.RouteCache{},
 		&xdscache_v3.ClusterCache{},
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -351,9 +351,6 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	endpointHandler := xdscache_v3.NewEndpointsTranslator(log.WithField("context", "endpointstranslator"))
 
 	resources := []xdscache.ResourceCache{
-		// The enableDebugListener boolean is defaulted to true which matches Contour's current behavior.
-		// Future iterations can allow for this to be customized by the user if they desire to disable the
-		// Envoy admin interface entirely.
 		xdscache_v3.NewListenerCache(listenerConfig, ctx.statsAddr, ctx.statsPort, ctx.Config.Network.EnvoyAdminPort),
 		&xdscache_v3.SecretCache{},
 		&xdscache_v3.RouteCache{},
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -351,7 +351,10 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	endpointHandler := xdscache_v3.NewEndpointsTranslator(log.WithField("context", "endpointstranslator"))
 
 	resources := []xdscache.ResourceCache{
-		xdscache_v3.NewListenerCache(listenerConfig, ctx.statsAddr, ctx.statsPort),
+		// The enableDebugListener boolean is defaulted to true which matches Contour's current behavior.
+		// Future iterations can allow for this to be customized by the user if they desire to disable the
+		// Envoy admin interface entirely.
+		xdscache_v3.NewListenerCache(listenerConfig, ctx.statsAddr, ctx.statsPort, true),
 		&xdscache_v3.SecretCache{},
 		&xdscache_v3.RouteCache{},
 		&xdscache_v3.ClusterCache{},
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -354,7 +354,7 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 		// The enableDebugListener boolean is defaulted to true which matches Contour's current behavior.
 		// Future iterations can allow for this to be customized by the user if they desire to disable the
 		// Envoy admin interface entirely.
-		xdscache_v3.NewListenerCache(listenerConfig, ctx.statsAddr, ctx.statsPort, true),
+		xdscache_v3.NewListenerCache(listenerConfig, ctx.statsAddr, ctx.statsPort, ctx.Config.Network.EnvoyAdminPort, true),
 		&xdscache_v3.SecretCache{},
 		&xdscache_v3.RouteCache{},
 		&xdscache_v3.ClusterCache{},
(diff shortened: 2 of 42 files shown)


stevesloka:
I think this is ready for a first look. The next bit would be to update all the docs which requires more drawing to fix up the diagram, etc.

examples/contour/03-envoy.yaml
```diff
@@ -117,6 +122,8 @@ spec:
         - name: envoycert
           mountPath: /certs
           readOnly: true
+        - name: envoy-admin
```
sunjayBhatia, line 125: does the bootstrap container need this mounted? I think the bootstrap command just generates the config with the requested address and doesnt talk on the socket to envoy since it runs before the envoy container exists

internal/envoy/bootstrap.go
```diff
@@ -102,6 +105,20 @@ func (c *BootstrapConfig) GetAdminAccessLogPath() string {
...
 }
+
+// ValidAdminAddress checks if the address supplied is
+// "localhost" or an IP address. Only a Unix Socket
+// is supported for this address to mitigate security.
+func ValidAdminAddress(address string) error {
+	// Value of "localhost" is invalid.
+	if address == "localhost" ||
+		net.ParseIP(address) != nil ||
+		!strings.Contains(address, "/") {
```
sunjayBhatia, line 116: edge case, but someone could conceivably want to use a relative file path with just the file name here (e.g. just `foo.sock`), which could be valid

maybe we should make the validation be an absolute file path to prevent this and make sure people are exact (e.g. check for `/` at the start of the string)? or relax the validation a bit to not require a slash?

could also make it check for https://pkg.go.dev/os#PathSeparator so its not another step to cross-platform convert if we want to in the future

internal/envoy/bootstrap.go
```diff
@@ -102,6 +105,20 @@ func (c *BootstrapConfig) GetAdminAccessLogPath() string {
...
+		return fmt.Errorf("invalid value %q, must be a unix socket name", address)
```
sunjayBhatia, line 117: maybe worth separating the error out for the localhost and ip address cases to say explicitly they are not allowed

internal/envoy/bootstrap.go
```diff
@@ -102,6 +105,20 @@ func (c *BootstrapConfig) GetAdminAccessLogPath() string {
 func (c *BootstrapConfig) GetDNSLookupFamily() string {
 	return stringOrDefault(c.DNSLookupFamily, "auto")
 }
+
+// ValidAdminAddress checks if the address supplied is
+// "localhost" or an IP address. Only a Unix Socket
+// is supported for this address to mitigate security.
+func ValidAdminAddress(address string) error {
```
sunjayBhatia, line 112: this could also go in the `internal/validation` package as well

internal/envoy/v3/listener.go
```diff
@@ -524,6 +524,17 @@ func SocketAddress(address string, port int) *envoy_core_v3.Address {
 			},
 		}
 	}
+	// Check if the address is a socket.
+	if strings.Contains(address, "/") {
```
sunjayBhatia, line 528: could use https://pkg.go.dev/os#PathSeparator here though this could get interesting if we had a socket address that the contour controller (not bootstrap) was programming somehow for a non-unix host

test/e2e/deployment.go
```diff
@@ -443,27 +444,42 @@ func (d *Deployment) EnsureResourcesForLocalContour() error {
 	}
 
 	// Add bootstrap ConfigMap as volume on Envoy pods (also removes cert volume).
```
sunjayBhatia, line 446: nit: add comment on the admin volume as well

test/e2e/infra/admin_test.go
```diff
@@ -0,0 +1,52 @@
...
+package infra
+
+import (
+	. "github.com/onsi/ginkgo"
+	"github.com/projectcontour/contour/test/e2e"
+	"github.com/stretchr/testify/require"
+)
+
+func testAdminInterface() {
+	Specify("requests to admin listener are served", func() {
```
sunjayBhatia, line 25: maybe add a test to make sure the endpoints that can modify envoy server settings are not available/we get connection errors/404s etc.

internal/xdscache/v3/listener.go
```diff
@@ -288,14 +288,24 @@ type ListenerCache struct {
 }
 
 // NewListenerCache returns an instance of a ListenerCache
-func NewListenerCache(config ListenerConfig, address string, port int) *ListenerCache {
+func NewListenerCache(config ListenerConfig, address string, port int, enableDebugListener bool) *ListenerCache {
 	stats := envoy_v3.StatsListener(address, port)
-	return &ListenerCache{
+	admin := envoy_v3.AdminListener("127.0.0.1", 9001)
```
sunjayBhatia, line 293: the admin port is configurable currently and I think that flag has kinda been dropped in a sense?

should we deprecate the flag? wire it through to make the listen port here for the read-only admin page configurable?

cmd/contour/shutdownmanager.go
```diff
@@ -30,8 +32,8 @@ import (
 )
 
 const (
-	prometheusURLFormat      = "http://localhost:%d/stats/prometheus"
-	healthcheckFailURLFormat = "http://localhost:%d/healthcheck/fail"
+	prometheusURLFormat      = "http://unix:%d/stats/prometheus"
```
sunjayBhatia, line 35: can we remove the `%d` from these? shouldnt need a port anymore

sunjayBhatia (commented):
left some comments, still gotta try it out locally but looks good generally

we'll probably need some documentation on how this is going to work and what new knobs/limitations we've added here

nice addition to the e2e tests!

tsaarni:
Regarding troubleshooting experience (ref [discussion](https://kubernetes.slack.com/archives/C8XRH2R4J/p1627571349100200?thread_ts=1627491065.080800&cid=C8XRH2R4J)) I myself depend on port-forward + curl config_dump as well. Could the troubleshooting use case be covered by new sub-command on contour instead of having option to enable additional admin listener? Something like `contour envoy config-dump`.

internal/envoy/bootstrap.go
stevesloka: Yeah you'd have to map the path of that to where Envoy is running to make that work. Maybe this should change to just the path and not the name? Just trying to keep this super simple for folks. 

I'll take it out and not require the `/`.

examples/contour/03-envoy.yaml
stevesloka: nope, that's a mistake.

internal/envoy/bootstrap.go
stevesloka: Yup we could, not sure how useful it would be in other packages.

internal/xdscache/v3/listener.go
stevesloka: We could reuse the port flag, but maybe that's confusing? 

I could reuse it for now, introduce a new flag and deprecate the old.

stevesloka: Ahh I see, I'm mixing up the bits. This would be a new flag on Contour to specify. Yeah let's do that, introduce a new config file item I guess and deprecate the bootstrap port.

youngnick (approved):
This LGTM, but I think it's worth calling out that this change has implications for running Envoy on Windows. I think it's more important to have a secure Linux deployment for now, so it's fine, but something for @sunjayBhatiaand @xaleeks to have a chat about maybe?

sunjayBhatia:
> This LGTM, but I think it's worth calling out that this change has implications for running Envoy on Windows. I think it's more important to have a secure Linux deployment for now, so it's fine, but something for @sunjayBhatiaand @xaleeks to have a chat about maybe?

windows does now support UDS but we would have to see if it works in this context specifically with envoy: https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/

i can see permissions in an emptydir volume possibly being a hangup etc. and would need to test things out

stevesloka:
> Regarding troubleshooting experience (ref discussion) I myself depend on port-forward + curl config_dump as well. Could the troubleshooting use case be covered by new sub-command on contour instead of having option to enable additional admin listener? Something like contour envoy config-dump.

@tsaarni ended up adding a new listener which exposes the read-only endpoints from the admin webpage over 127.0.0.1:9001 so you can still port-forward and get to those stats, but not expose the dangerous bits.

tsaarni:
> @tsaarni ended up adding a new listener which exposes the read-only endpoints from the admin webpage over 127.0.0.1:9001 so you can still port-forward and get to those stats, but not expose the dangerous bits.

@stevesloka This is surely better but wouldn't it still leave Envoy vulnerable for [information exposure](https://cwe.mitre.org/data/definitions/200.html) when `enableExternalNameService` is set `true`?

stevesloka:
> @stevesloka This is surely better but wouldn't it still leave Envoy vulnerable for information exposure when enableExternalNameService is set true?

Yup it would indeed do that if you enabled external name services. You'd still need to find a domain that points to localhost, etc, etc. 

I'm also adding a config item to Contour to remove the readonly endpoint. So for folks who want to enable externalname, they can disable the entire admin interface.

tsaarni:
> Maybe I can just have it so that if the port is set to `0`, then it's disabled rather than adding another config item.

Since the parameter is new in contour config, in theory not setting `admin-port` at all could mean disable admin listener. But then of course it would be a change in the current behavior - it was on by default, even without explicitly setting parameters in bootstrap command line.

test/e2e/infra/admin_test.go
stevesloka: 👍

test/e2e/deployment.go
stevesloka: 👍

cmd/contour/shutdownmanager.go
stevesloka: 👍

stevesloka:
That's a good point @tsaarni about it defaulting to on. Maybe we should keep that for now and if you want to disable, you have to explicitly set that.

We could change that behavior in the future if we'd like. I'm going to write up a docs PR next to explain all of this.

stevesloka:
@youngnick @sunjayBhatia I made some more changes, probably enough for a 2nd look. I'm going to write up a docs PR now to accompany, but don't want to distract away from this too much.

skriss (changes requested):
I think this change makes sense, had a few comments inline. I guess we also need to confirm that we're comfortable making this change in a minor release, although it could be considered a breaking change, in the interest of security.

internal/envoy/bootstrap.go
```diff
@@ -44,6 +46,7 @@ type BootstrapConfig struct {
 	// Defaults to 127.0.0.1.
 	AdminAddress string
```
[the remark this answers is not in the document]
stevesloka: 👍

cmd/contour/bootstrap.go
```diff
@@ -26,8 +26,8 @@ func registerBootstrap(app *kingpin.Application) (*kingpin.CmdClause, *envoy.Boo
 	bootstrap := app.Command("bootstrap", "Generate bootstrap configuration.")
 	bootstrap.Arg("path", "Configuration file ('-' for standard output).").Required().StringVar(&config.Path)
 	bootstrap.Flag("resources-dir", "Directory where configuration files will be written to.").StringVar(&config.ResourcesDir)
-	bootstrap.Flag("admin-address", "Envoy admin interface address.").StringVar(&config.AdminAddress)
-	bootstrap.Flag("admin-port", "Envoy admin interface port.").IntVar(&config.AdminPort)
+	bootstrap.Flag("admin-address", "Envoy admin interface address.").Default("/admin/admin.sock").StringVar(&config.AdminAddress)
```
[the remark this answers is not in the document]
stevesloka: 👍

cmd/contour/serve.go
```diff
@@ -351,7 +351,10 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	endpointHandler := xdscache_v3.NewEndpointsTranslator(log.WithField("context", "endpointstranslator"))
 
 	resources := []xdscache.ResourceCache{
-		xdscache_v3.NewListenerCache(listenerConfig, ctx.statsAddr, ctx.statsPort),
+		// The enableDebugListener boolean is defaulted to true which matches Contour's current behavior.
+		// Future iterations can allow for this to be customized by the user if they desire to disable the
+		// Envoy admin interface entirely.
```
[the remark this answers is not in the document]
stevesloka: Yup good catch. I initially had that flag in there but took it out to reduce the amount of configuration items.

internal/envoy/v3/stats.go
```diff
@@ -80,3 +146,20 @@ func StatsListener(address string, port int) *envoy_listener_v3.Listener {
 		SocketOptions: TCPKeepaliveSocketOptions(),
 	}
 }
+
+func serviceStatsRoute(prefix string) *envoy_route_v3.Route {
```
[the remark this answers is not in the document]
stevesloka: yup I agree. That's what it got called initially. Not sure if we want to split off this admin stuff into a new one? But that seems like a waste. Also not sure if folks have the `service-stats` in any dashboards, etc.

pkg/config/parameters.go
```diff
@@ -525,7 +525,10 @@ type NetworkParameters struct {
 	//
 	// See https://www.envoyproxy.io/docs/envoy/v1.17.0/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto?highlight=xff_num_trusted_hops
 	// for more information.
-	XffNumTrustedHops uint32 `yaml:"num-trusted-hops"`
+	XffNumTrustedHops uint32 `yaml:"num-trusted-hops,omitempty"`
+
+	// Configure the port used to access the Envoy Admin interface.
```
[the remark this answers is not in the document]
stevesloka: This will default to "on" if you don't specify. The only way to turn it off is to make the port  "0". This matches what we do today, but offers up a way for folks to be "more secure" if they'd like. 

I'm working on explaining those bits in the the docs PR which is turning into a bunch of work. =)

internal/envoy/bootstrap.go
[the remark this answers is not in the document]
youngnick: Since we're blocking anything other than file paths, we should document that the only AdminAddress we support is a file path, not an IP address.

internal/envoy/bootstrap.go
```diff
@@ -102,6 +104,18 @@ func (c *BootstrapConfig) GetAdminAccessLogPath() string {
...
+
+// ValidAdminAddress checks if the address supplied is
+// "localhost" or an IP address. Only a Unix Socket
+// is supported for this address to mitigate security.
+func ValidAdminAddress(address string) error {
+	// Value of "localhost" is invalid.
+	if address == "localhost" || net.ParseIP(address) != nil {
+		return fmt.Errorf("invalid value %q, cannot be `localhost` or an ip address", address)
+	}
+	return nil
```
youngnick, line 116: Should we check that the unix socket exists, or at least that the path is a valid one?

internal/xdscache/v3/listener.go
```diff
@@ -288,14 +288,24 @@ type ListenerCache struct {
 }
 
 // NewListenerCache returns an instance of a ListenerCache
-func NewListenerCache(config ListenerConfig, address string, port int) *ListenerCache {
-	stats := envoy_v3.StatsListener(address, port)
-	return &ListenerCache{
+func NewListenerCache(config ListenerConfig, statsAddress string, statsPort, adminPort int) *ListenerCache {
+	stats := envoy_v3.StatsListener(statsAddress, statsPort)
+	admin := envoy_v3.AdminListener("127.0.0.1", adminPort)
```
youngnick, line 293: Should we set this to the socket name to make sure there's no confusion, or is this only used if we're exposing the admin back on a network interface?

youngnick (commented):
Only some small questions left from me, this is looking really great.

cmd/contour/shutdownmanager.go
```diff
@@ -290,6 +309,7 @@ func registerShutdown(cmd *kingpin.CmdClause, log logrus.FieldLogger) (*kingpin.
 
 	shutdown := cmd.Command("shutdown", "Initiate an shutdown sequence which configures Envoy to begin draining connections.")
 	shutdown.Flag("admin-port", "Envoy admin interface port.").IntVar(&ctx.adminPort)
```
sunjayBhatia, line 311: seems like this flag should be noted as deprecated?

sunjayBhatia (commented):
largely looks good! interested in the comment(s) around "service-stats" name etc. and not sure as well if it is ok to change that etc.

internal/xdscache/v3/listener.go
stevesloka: This is used to create a listener which exposes the read-only bits from the admin page.

internal/envoy/bootstrap.go
stevesloka: I'm not sure of the timing of that type of check. This is the bootstrap code which executes in the initContainer. Then the pod would startup after that. Need to double check if the volumes are all ready in init phase or not, I don't remember off the top of my head.

cmd/contour/shutdownmanager.go
stevesloka: marked as deprecated!

internal/envoy/v3/stats.go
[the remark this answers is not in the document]
stevesloka: I think to be fair we need to do a proper deprecation of this field. I really doubt this would be a breaking change for most, but not 100% sure. Ok to make a followup PR?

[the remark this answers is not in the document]
stevesloka: //cc @sunjayBhatia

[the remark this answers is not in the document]
skriss: Yep 👍

internal/envoy/bootstrap.go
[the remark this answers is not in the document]
skriss: This should still be changed from "AdminAddress is the TCP address..." to "AdminAddress is the Unix domain socket path..."

skriss (changes requested):
Just one last nit then I think this LGTM

stevesloka:
> This should still be changed from "AdminAddress is the TCP address..." to "AdminAddress is the Unix domain socket path..."

bah, didn't push, sorry. fixed up!

internal/envoy/v3/stats.go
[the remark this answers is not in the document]
sunjayBhatia: sounds good to me, should err on the side of being careful 👍🏽

sunjayBhatia (approved):

skriss (approved):
no further comments, LGTM!

PR #3935, opened 2021-07-29
internal/envoy: fix TCPProxy weights bug

skriss:
Fixes a bug where when a TCPProxy had more than one
service defined, any service with a zero weight was
updated to have a weight of 1. The correct behavior
is that if all service have no weights, then they
should be evenly weighted, but if some have non-zero
weights and others have zero weights, they stay as-is.

Notes for reviewers:
- The desired behavior that I describe in the commit message is not explicitly documented anywhere for TCPProxy, however it is for HTTPProxy (and my assumption is that the behavior should be the same between them): https://projectcontour.io/docs/v1.18.0/config/request-routing/#upstream-weighting
- This came up while working on supporting weighted forward to's for TLSRoute (#3440).

### DIFF
--- a/internal/envoy/v3/listener.go
+++ b/internal/envoy/v3/listener.go
@@ -460,51 +460,71 @@ func TCPProxy(statPrefix string, proxy *dag.TCPProxy, accesslogger []*accesslog.
 	// The value of two and a half hours for reasons documented at
 	// https://github.com/projectcontour/contour/issues/1074
 	// Set to 9001 because now it's OVER NINE THOUSAND.
-	idleTimeout := protobuf.Duration(9001 * time.Second)
+	tcpProxy := &tcp.TcpProxy{
+		StatPrefix:  statPrefix,
+		AccessLog:   accesslogger,
+		IdleTimeout: protobuf.Duration(9001 * time.Second),
+	}
 
-	switch len(proxy.Clusters) {
+	// Get the total weight of all clusters for the proxy.
+	var totalWeight uint32
+	for _, c := range proxy.Clusters {
+		totalWeight += c.Weight
+	}
+
+	// If some clusters have non-zero weights, then drop
+	// any others that have zero weights since they shouldn't
+	// get traffic (note that TCPProxy weighted clusters can't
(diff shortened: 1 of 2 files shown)


stevesloka (approved):

internal/envoy/v3/listener.go
```diff
@@ -460,52 +460,72 @@ func TCPProxy(statPrefix string, proxy *dag.TCPProxy, accesslogger []*accesslog.
...
+	// If some clusters have non-zero weights, then drop
+	// any others that have zero weights since they shouldn't
+	// get traffic (note that TCPProxy weighted clusters can't
+	// have zero weights, unlike HTTP route weighted clusters,
+	// so we can't include them with a zero weight). Also note
+	// that if all clusters have zero weights, then we keep them
+	// all and evenly weight them below.
+	var keepClusters []*dag.Cluster
+	for _, c := range proxy.Clusters {
+		if totalWeight > 0 && c.Weight == 0 {
```
sunjayBhatia, line 484: just a thought how to do this in one pass (havent tested it tho)

```
var totalWeight uint32
var weightedClusters []*dag.Cluster
for _, c := range proxy.Clusters {
  if c.Weight != 0 {
    weightedClusters = append(weightedClusters, c)
  }
  totalWeight += c.Weight
}

var actualClusters []*dag.Cluster
if totalWeight == 0 {
  actualClusters = proxy.Clusters
} else {
  actualClusters = weightedClusters
}

switch len(actualClusters) {
...
```

could rename `weightedClusters` to `clustersWithWeight` too

sunjayBhatia (approved):
just one thought about how to maybe do the processing with one loop, otherwise lgtm! nice fix

internal/envoy/v3/listener.go
skriss: :+1: I prioritized readability over perf here, but made an update here along the lines of ^^^ that I think is still easy to read/follow.

sunjayBhatia: nice! yeah good adaptation

PR #3938, opened 2021-07-30
internal/dag: support TLSRoute weighted ForwardTos

skriss:
Add support for weighted ForwardTos in TLSRoute rules.
Per spec, weights default to 1 if unspecified, and weights
of 0 mean no traffic is forwarded to the backend.

Updates #3440.

### DIFF
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -537,6 +537,8 @@ func (p *GatewayAPIProcessor) computeTLSRoute(route *gatewayapi_v1alpha1.TLSRout
 		}
 
 		var proxy TCPProxy
+		var totalWeight uint32
+
 		for _, forward := range rule.ForwardTo {
 
 			service, err := p.validateForwardTo(forward.ServiceName, forward.Port, route.Namespace)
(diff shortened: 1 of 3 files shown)


internal/dag/gatewayapi_processor.go
```diff
@@ -545,14 +547,28 @@ func (p *GatewayAPIProcessor) computeTLSRoute(route *gatewayapi_v1alpha1.TLSRout
...
+				Weight:   routeWeight,
 			})
 		}
 
-		if len(proxy.Clusters) == 0 {
-			// No valid clusters so the route should get rejected.
+		// No valid clusters or all forwardTos have a weight of 0
+		// so the route should get rejected.
+		if len(proxy.Clusters) == 0 || totalWeight == 0 {
 			continue
```
skriss, line 572: TODO should log and/or set a condition here

sunjayBhatia (approved):
looks good, that comment about status/logging makes sense, from what i saw in the spec, it doesnt say what to do when all weights are explicitly 0, so maybe worth a PR upstream to clarify

youngnick (approved):
LGTM, nice work.

youngnick:
Agree that a PR upstream to clarify docs may be worthwhile.

PR #3939, opened 2021-07-30
internal/dag: improve TLSRoute selection test coverage

skriss:
Adds test coverage for the matching of TLSRoutes with Gateways.

Updates #3440.

For reviewers: I replaced some of the existing test cases that covered some of this functionality just because it was easier to start fresh and build out the matrix. I did not cover every possible permutation of Gateway + TLSRoute binding settings as the matrix is pretty big, but this should cover each setting individually.

The diff is not very useful, here's the section I added: https://github.com/projectcontour/contour/blob/e9d26ee418e14e410e3010d2c4771554b7c62a6a/internal/dag/builder_test.go#L1169-L1605

### DIFF
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -305,6 +305,81 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 		},
 	}
 
+	gatewayTLSRouteSameNamespace := &gatewayapi_v1alpha1.Gateway{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "contour",
+			Namespace: "projectcontour",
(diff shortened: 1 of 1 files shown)


stevesloka:
Looks like a conflict needs resolved when you're back @skriss.

skriss:
> Looks like a conflict needs resolved when you're back @skriss.

Fixed!

stevesloka (approved):

youngnick (approved):
This looks fine, sadly will need substantial refactoring with the Gateway - Route binding changes in v1alpha2. Let's get this in for now, and use it as a base for those changes.

Issue #3835, opened 2021-06-24
Accessibility Assessment


johnnycase:
Please run an accessiblity assessment on the project contour docs site and create updates on this issue with any changes needed. This can be done using the Lighthouse tool that is built into the dev tools of chromium browsers.

jonasrosland:
Hi @johnnycase , just wanted to follow up and see what we can do to help you with this? Happy to talk on Slack as well.

johnnycase:
Hey @jonasrosland I do have a question about the scope of the work. Do we need to check accessibility on the whole project contour site or only on the docs portion of the site?

johnnycase:
Accessibility Issues to work:

- Page: https://projectcontour.io/docs/main/
- Link text does not have a sufficient contrast ratio
- Use color #007db1 to fix the color contrast

pnbrown:
It would be good for us to tick off the boxes above that haven't yet been ticked as well as running another accessibility test on the site.

PR #3957, opened 2021-08-21
Contrast change - accessibility

johnnycase:
Updates #3835 
Signed-off-by: johnnycase <johneverettcase@gmail.com>

Changing the color for grey and blue in the variables so the contrast meets accessibility standards.

### DIFF
--- a/site/themes/contour/assets/scss/_variables.scss
+++ b/site/themes/contour/assets/scss/_variables.scss
@@ -1,10 +1,10 @@
 $white: #ffffff;
-$blue: #0095D3;
+$blue: #007db1;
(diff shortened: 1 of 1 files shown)


youngnick:
This LGTM, but I'd like to hear more from other people and check that we're not violating some branding agreement or something by changing colors. @jonasrosland, anything we need to think about there?

sunjayBhatia (approved):
@johnnycase showed us in office hours, looked good to me and in the preview

@jonasrosland mentioned it was ok to change colors

youngnick (approved):

Issue #3956, opened 2021-08-20
controller-runtime controllers and the rest of Contour code use two different caches


skriss:
The controller-runtime controllers (for Gateway API) use one `cache.Cache`, while the rest of Contour code uses a `cache.Cache` from `k8s.Clients` (see https://github.com/projectcontour/contour/blob/main/internal/k8s/clients.go#L36). This is both memory-inefficient, and can cause subtle race conditions due to the two caches not being 100% in sync. I've seen this manifest when a controller-runtime controller is processing an event for a new object (meaning it's in the controllers' cache), it goes to update the status of that object, and the StatusUpdater can't find that object in its cache (which is the `k8s.Clients` one) so it can't update it.

Ideally we could use one cache instance in both places. It's not immediately obvious how to wire things up correctly for this, so will require some investigation.

skriss:
Here's a log snippet from a WIP branch for #3706 demonstrating the race condition upon creating a new `GatewayClass`:

```
INFO[0951] enqueueing gatewayclass                       context=gatewayclass-controller name=example-2
INFO[0951] reconciling gatewayclass                      context=gatewayclass-controller name=example-2
ERRO[0951] unable to update status                       context=StatusUpdateHandler error="GatewayClass.networking.x-k8s.io \"example-2\" not found" name=example-2 namespace=
```

youngnick:
Oof, guess we all missed this one. Agreed that this should all use the same cache, definiely.

stevesloka:
Good catch! The original idea was to move all of them over to controller runtime, but we ended up with just the Gateway bits. That might be an easier path forward to just move to that. Just a thought while thinking out loud. Would mean moving the other bits to use the same cache then.

skriss:
Was able to get everything using the one instance of the cache and have not seen the race condition since; need to clean it up though.

Issue #3864, opened 2021-07-06
Gateway: Add validation to pick a single gateway from ControllerName


stevesloka:
For Gateway, the problem is that you need to enqueue it to set the status on it. So I think the logic for this is to make sure that we only select a single gateway and set errors on those that may match but don't fit the criteria (meaning oldest wins?)

_Originally posted by @stevesloka in https://github.com/projectcontour/contour/pull/3858#r664574230_

PR #3960, opened 2021-08-23
internal/controller: admit the oldest controlled GatewayClass

### LINKED ISSUE #3706: Improve Handling Multiple GatewayClasses
https://github.com/projectcontour/contour/pull/3659 adds gatewayclass support. The PR handles multiple gatewayclasses by admitting the first one that is reconciled. We should have well defined behavior for handling multiple gatewayclases, i.e. admit the oldest.

xref: https://github.com/projectcontour/contour/pull/3659#discussion_r638178777

skriss:
When more than one gateway class exists with a matching
controller name, admit the oldest one (tie-breaking by
alphabetical order) by setting the "Admitted" condition
to true on it and setting the "Admitted" condition to
false on all other gateway classes with a matching
controller name.

Closes #3706.


skriss:
I'll have another PR or two for the `Gateway` side of things, as well as for #3956.

Still need to decide the best way/ways to test this; since we were talking in #3884 about changing the approach to testing these controllers I haven't fixed the existing tests yet.

skriss:
@stevesloka question, do we still need to add the `GatewayClass` to the DAG? We have a cache field for it (https://github.com/projectcontour/contour/blob/main/internal/dag/cache.go#L59), and we check that it's populated (https://github.com/projectcontour/contour/blob/main/internal/dag/gatewayapi_processor.go#L83-L86), but we don't actually use it for anything. My thinking is that we can drop it from the DAG, set its status directly in the controller like this PR is doing, and then only set a `Gateway` in the DAG if it's using an Admitted=true GatewayClass (which I will have a subsequent PR for). Does that sound right to you?

stevesloka:
@skriss at this point I don't think we need the Class in the dag. Before we were using the spec on it for configuration details, but that's not the case in the new design.

skriss:
OK, I got the existing tests passing + added one more case for the new logic so I'll mark this ready for review.

internal/controller/gatewayclass.go
```diff
@@ -96,10 +103,27 @@ func (r *gatewayClassReconciler) hasMatchingController(obj client.Object) bool {
...
+		}
+
+		controlledClasses.add(&gc)
+	}
+
+	// no controlled gatewayclasses, trigger a delete
+	if controlledClasses.len() == 0 {
 		r.log.WithField("name", request.Name).Info("failed to find gatewayclass")
 
 		r.eventHandler.OnDelete(&gatewayapi_v1alpha1.GatewayClass{
```
skriss, line 129: eventually we can drop passing the GC to the DAG, we really just need to update the status, but leaving it for now until I get a PR up for https://github.com/projectcontour/contour/issues/3864

internal/controller/gatewayclass.go
```diff
@@ -108,24 +132,96 @@ func (r *gatewayClassReconciler) Reconcile(ctx context.Context, request reconcil
...
+				}),
+			})
+		} else {
+			// this branch makes testing easier by not going through the StatusUpdater.
+			copy := status.SetGatewayClassAdmitted(context.Background(), r.client, gc.DeepCopy(), false)
+
+			if err := r.client.Status().Update(context.Background(), copy); err != nil {
+				return reconcile.Result{}, fmt.Errorf("error updating status of gateway class %s: %v", copy.Name, err)
+			}
+		}
```
skriss, line 159: this is slightly hacky but setting up the StatusUpdater in an envtest setting is kind of a pain.

internal/controller/gatewayclass.go
```diff
@@ -108,24 +132,96 @@ func (r *gatewayClassReconciler) Reconcile(ctx context.Context, request reconcil
...
+	} else {
+		// this branch makes testing easier by not going through the StatusUpdater.
+		copy := status.SetGatewayClassAdmitted(context.Background(), r.client, controlledClasses.admittedClass().DeepCopy(), true)
+		if err := r.client.Status().Update(context.Background(), copy); err != nil {
+			return reconcile.Result{}, fmt.Errorf("error updating status of gateway class %s: %v", copy.Name, err)
+		}
 	}
-	r.log.WithField("name", gc.Name).Info("synced gatewayclass status")
+
+	r.eventHandler.OnAdd(controlledClasses.admittedClass())
```
skriss, line 184: eventually we can drop passing the GC to the DAG, we really just need to update the status, but leaving it for now until I get a PR up for #3864

PR #3974, opened 2021-08-25
Cherrypick #3834

stevesloka:
Move to Envoy Admin over unix socket to mitigate security issues with external name services.


stevesloka:
The e2e's failed for me locally, but not sure if a local dep thing or not, waiting to see what CI does.

PR #3979, opened 2021-08-25
(release-1.14) CI: use locally built image for integration tests

skriss:
Updates the example manifests to use the "working"
image tag when running integration tests, so the
locally built image is used rather than a versioned
image from the registry.


sunjayBhatia (approved):

skriss:
Passed everything relevant, will need a power-merge.

stevesloka (approved):

PR #3988, opened 2021-08-26
doc: program 502/503 routes for misconfigured services and includes

tsaarni:
Documentation changes related to #3824.

### DIFF
--- a/site/content/docs/main/config/fundamentals.md
+++ b/site/content/docs/main/config/fundamentals.md
@@ -116,8 +116,8 @@ httpproxy "basic" deleted
 ## Status Reporting
 
 There are many misconfigurations that could cause an HTTPProxy or delegation to be invalid.
-To aid users in resolving these issues, Contour updates a `status` field in all HTTPProxy objects.
-In the current specification, invalid HTTPProxy are ignored by Contour and will not be used in the ingress routing configuration.
+Contour will make its best effort to process even partially valid configuration and allow traffic to be served for the valid parts.
+To aid users in resolving any issues, Contour updates a `status` field in all HTTPProxy objects.
 
 If an HTTPProxy object is valid, it will have a status property that looks like this:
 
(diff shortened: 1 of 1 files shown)


site/content/docs/main/config/fundamentals.md
```diff
@@ -151,6 +151,9 @@ Some examples of invalid configurations that Contour provides statuses for:
 - Multiple header conditions of type "exact match" with the same header key.
 - Contradictory header conditions on a route, e.g. a "contains" and "notcontains" condition for the same header and value.
 
+Invalid configuration is ignored and will be not used in the ingress routing configuration.
```
sunjayBhatia, line 154: I would say enumerating what invalid configurations are "ignored" and 50x responses programmed for would be good

maybe a table that we could expand on in the future, or a couple quick sentences here

youngnick: I agree that we should at least have some examples of invalid config and how this works in those cases. Maybe just describing the scenarios covered by test cases is enough for now.

I don't think that having an exhaustive list is very maintainable long-term, but we should at least give people an idea of what happens and where to look for more info in this bit.

youngnick (commented):
This is looking great! Just needs a little more info, maybe some examples.

I think that regardless of any of the ConfigCRD work, we can't cut 1.19 until this is done in addition to #3824.

tsaarni:
Thanks for your comments!

I did not end up with a table of cases, but just a bullet list instead, since we basically just have these two categories covered so far: missing service and missing include.  The test cases had some more variations so it looks like there is more, but I think it basically boiled down to these two.

I added an example for missing service, showing what kind of information the user will find in the status conditions to help their troubleshooting.  I could add one for missing include too, but the approach is anyways the same with a different error message in conditions.

site/content/docs/main/config/fundamentals.md
```diff
@@ -151,6 +151,40 @@ Some examples of invalid configurations that Contour provides statuses for:
...
+        - name: valid-service
+          port: 80
+    - conditions:
+      - prefix: /subpage
+      services:
+        - name: service-that-does-not-exist
+          port: 80
+```
+
+The status of the `HTTPProxy` will be `failed` and the conditions will have detailed error message: `Spec.Routes unresolved service reference: service "default/service-that-does-not-exist" not found`.
```
sunjayBhatia, line 185: "failed" => "Invalid=false"?

sunjayBhatia:
> Thanks for your comments!
> 
> I did not end up with a table of cases, but just a bullet list instead, since we basically just have these two categories covered so far: missing service and missing include. The test cases had some more variations so it looks like there is more, but I think it basically boiled down to these two.
> 
> I added an example for missing service, showing what kind of information the user will find in the status conditions to help their troubleshooting. I could add one for missing include too, but the approach is anyways the same with a different error message in conditions.

thanks for that update, looks good 👍🏽 just one nit from me

site/content/docs/main/config/fundamentals.md
tsaarni: Done!

sunjayBhatia (approved):

skriss (approved):

PR #3992, opened 2021-08-26
use the oldest Gateway for the controlled+admitted GatewayClass

### LINKED ISSUE #3812: Adds Gateway Controller Integration Tests
https://github.com/projectcontour/contour/pull/3773 adds support for testing controller-runtime controllers, starting with the gatewayclass controller. https://github.com/projectcontour/contour/pull/3805 adds a controller-runtime controller to reconcile gateway objects. Tests should be added to exercise the gateway controller.

### LINKED ISSUE #3825: Support Multiple Gateways
~~Currently, the operator supports multiple gateways that translate to multiple Contour/Envoy deployments. This is required for Knative. In order to deprecate the gatewayclass/gateway controllers in the operator, Contour must remove `gateway.namespace` and `gateway.name` from the config and support multiple gateways. Each gateway should equal a Contour/Envoy deployment or one contour managing multiple envoy fleets.~~

~~xref: https://github.com/projectcontour/contour/pull/3805~~

Contour is/has changed to only take a Gateway Controller Name to choose what GatewayClass/Gateways to reconcile

Currently a Contour instance+Envoy DaemonSet will serve routes attached to a single Gateway.

Contour needs to appropriately handle multiple Gateways:
- [ ] Initially (for correctness), when multiple Gateways that reference the same GatewayClass exist, only the oldest should reconciled
  - Removes possible conflicts, follows API conflict resolution guidelines
  - Can be implemented ASAP, no dependencies on other features
- [ ] A further feature: when multiple Gateways that reference the same GatewayClass exist, they can be "merged" 
  - This needs more discussion
  - "merged" could mean that compatible listeners between Gateways are merged
  - Could mean that we support more dynamic listener configuration
  - Possibly needs to be done *after* Contour starts managing Envoy since you probably will need to change pod spec
  - Need to take into account endpoint publishing policy

  
cc: @nak3 @Miciah

### LINKED ISSUE #3863: Gateway check for GatewayClass Admitted Status
Need to resync the Gateway when a GatewayClass status changes. 

For instance, if a Gateway is created, and the Class is not yet Admitted, we'll skip and wait for that status. But when the Class is finally admitted, then we should resync for Gateways to check for any that are valid and Admit those, but there won't be an object update so we'll need to query the API for Gateways.

_Originally posted by @stevesloka in https://github.com/projectcontour/contour/pull/3858#r664571792_

### LINKED ISSUE #3864: Gateway: Add validation to pick a single gateway from ControllerName

skriss:
Finds the oldest Gateway for the controlled and admitted
GatewayClass and passes that one to the DAG. Sets a
condition of "Scheduled: false" on all other GatewayClass
for that class.

Closes #3812.
Closes #3825.
Closes #3863.
Closes #3864.


skriss:
OK, got some E2E's written for this, ready for review.

sunjayBhatia (approved):
looks good, cant think of any additional tests we need, nice work!

stevesloka (approved):
Looks good!

I wonder if we will, however, end up with a timing issue where a GatewayClass & Gateway are created at the same time. The Class is not yet admitted so the Gateway is rejected, but then never gets processed. Different issue though, just thought about it when looking through this.

skriss:
> Looks good!
> 
> I wonder if we will, however, end up with a timing issue where a GatewayClass & Gateway are created at the same time. The Class is not yet admitted so the Gateway is rejected, but then never gets processed. Different issue though, just thought about it when looking through this.

So [this part here](https://github.com/projectcontour/contour/blob/4d8333baa12c1c35adb8a34ebeab80bd38cada6e/internal/controller/gateway.go#L78-L86) causes Gateways to be reconciled any time their GatewayClass changes, so in your scenario, once the GC changes to admitted, then the Gateways should get processed and admitted/scheduled as appropriate. It's a similar issue to when an older GC gets deleted and a newer one takes effect, and a Gateway for the newer one needs to then be admitted (but nothing about the Gateway itself has changed).

PR #3993, opened 2021-08-26
Add Cookie hash parameters to RequestHash load balancing strategy

sunjayBhatia:
To address https://github.com/projectcontour/contour/issues/3044

Users will be able to supply header or cookie load balancing policies

Cookie load balancing can be migrated from the existing to the list of policies
- We can deprecate the existing `Cookie` strategy if needed
- Offer a backwards compatible "default session affinity" option that matches what the `Cookie` strategy provides with the `X-Contour-Session-Affinity` cookie name
- Allow users to use a custom cookie name
- Allow users to opt-in to the "passive" cookie load balancing if they choose
- Don't allow setting a TTL, we will set a TTLS of 0 like we do with the existing `Cookie` strategy, felt easier to give the option of "passive" or not

In the future for generated cookies we might be able to also allow some cookie modification directly in the new cookie hash policy struct to add secure, httponly, etc. attributes


apis/projectcontour/v1/httpproxy.go
```diff
@@ -725,6 +725,36 @@ type HeaderHashOptions struct {
...
+type CookieHashOptions struct {
+	// UseDefaultSessionAffinity
+	UseDefaultSessionAffinity bool `json:"useDefaultSessionAffinity,omitempty"`
+
+	// CookieName is the name of the HTTP Cookie that will be used to
+	// calculate the hash key. If the cookie specified is not present on a
+	// request and the Passive field is true, no hash will be produced.
+	// If the cookie specified is not present on a request and the Passive
+	// field is false, a hash will be produced and sent in the Set-Cookie
+	// header of the response.
```
youngnick, line 739: Maybe a short statement here about what happens when the cookie is set? Something like "If the cookie specified is present, then that value will be the value used to generate a hash to a backend; as long as no backends change, the traffic will flow consistently to a single backend."

apis/projectcontour/v1/httpproxy.go
```diff
@@ -725,6 +725,36 @@ type HeaderHashOptions struct {
...
+
+	// Passive configures if Envoy should generate and return a hash when
+	// the specified cookie is not present on a request. By default when
+	// Passive is false, Envoy will be configured to generate a hash for
+	// a request without the specified cookie and return it in the
+	// Set-Cookie header in the response. Setting Passive to true ensures
+	// responses to requests without the specified cookie do not contain
+	// a hash that can be used for session-affinity. Setting Passive to true
+	// can be used to implement session-affinity with application generated
+	// cookies.
```
youngnick, line 753: Two suggestions here:
- I think that making the field name a bit more descriptive would be helpful - `GeneratedCookie` indicates the behavior a little better, but could probably still be improved upon. What I think we should aim for herre is a field name that means you can get most of the way to understanding what the field is for without reading the docstring.
- I'd also call out the important detail that if you do set this to true, it's not guaranteed that traffic will return to the app instance that generated the session token.

youngnick (commented):
This is looking great, I like the approach, just a couple of small suggestions to clarify some details.

PR #4004, opened 2021-08-30
remove BackendPolicy informers, caching and RBAC

skriss:
The upcoming v1alpha2 release of Gateway API drops
BackendPolicy. Contour currently sets up an informer
for it but does not do any further processing based
on it. As such, this PR fully removes all code for it.

### DIFF
--- a/examples/contour/02-role-contour.yaml
+++ b/examples/contour/02-role-contour.yaml
@@ -82,7 +82,6 @@ rules:
 - apiGroups:
   - networking.x-k8s.io
   resources:
-  - backendpolicies
   - gatewayclasses
   - gateways
   - httproutes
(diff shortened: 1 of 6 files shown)


sunjayBhatia (approved):

sunjayBhatia (commented):
one thing to add would maybe be to remove the CRD from being added to the gateway example? though that might be annoying and have minimal value for the effort

skriss:
> one thing to add would maybe be to remove the CRD from being added to the gateway example? though that might be annoying and have minimal value for the effort

Yeah, I thought about it but we'll update to the `v1alpha2` YAML soon anyway which will drop that CRD, and I figured keeping the CRD definition for now isn't hurting anything and costs nothing. Let me know if you feel strongly, otherwise I won't do anything further here.

youngnick (approved):
LGTM, nice work.

PR #4011, opened 2021-09-02
Add threat model and security posture documentation

youngnick:
This PR adds a document outlining what I think is a description of Contour's current threat model and security posture, which is required for CII Silver, but is a good idea on its own.

After making changes to SECURITY.md earlier, I realised during this process that Github's UI does not follow symlinks, so I've put this threat model into that file, with a reference to the reporting process document. I've duplicated that content into a page on the site as well, which is not ideal, but I couldn't think of a way to do this without opening the whole site up to inclusion (which seems worse).

I've also backported the `Security` section to 1.18.1, so the initial load of the `Documentation` tab on the site has it available (currently you have to switch to `main` for it to be viewable.) Obviously this will be fixed when we release 1.19, since we base the docs on what's in the `main` directory, but this fixes us for right now.

Comments and discussion welcomed, this is by no means the final word on the subject.


tsaarni (commented):
Here are some additional thoughts that came to my mind while reading

### Internal threats 

Contour control plane exposes few cluster internal endpoints as well (xDS server, and metrics as well could be counted in here). These could be interesting attack surface for a malicious internal actor who already has access to the cluster, either by being a valid user (application developer), or via vulnerabilities in some other components running in the cluster. 

### Assets

What are the assets that Contour has access to, that are also in interest of an attacker? 

Obviously the most sensitive are the Kubernetes secrets, but Contour also has information about all service endpoints which attacker would be interested in when planning their route to the next target in the system. Contour in part controls who gets access to these assets via xDS when it shares the information when configuring Envoys.

Contour creates the routing tables for Envoy which may provide interesting opportunities for an attacker if they could impact the routing decision (like already referred in the doc: Confused deputy).  As circulated widely, it is not only `Service` of type `ExternalName`, but also `Endpoint` (and potentially `EndpointSlice` at some point, if implemented). Those may provide means for an attacker to impact the flow of the traffic in malicious ways. In case of `Endpoint` and `EndpointSlice` it is not within control of Contour if they are misused, like already pointed out in the doc, but I think it is still relevant information. These can be abused.

Contour is in critical path when it comes to setting up authentication and authorization for the backend services, yet another interesting opportunity if that could be bypassed.

Contour has its own logs, and it decides how HTTP access logging is set up in Envoy. If attacker could prevent access logging, it could be useful for them.

### Multi-team ingress controller

Single instance of Contour control plane can serve individual teams working on a shared cluster. In this use case, different teams are not expected to have access to each others resources. Contour could be seen as multi-tenant service, which is always tricky thing to get right.

Contour instance itself has no concept of tenancy - it obviously cannot associate any kind of "tenant id" to each resource, to ensure that they are kept separate from each other in all possible cases.  To keep teams separate, we rely on RBAC. This model of course originates from Kubernetes, not much that Contour could do differently, but when this model is "leaking", it can create interesting opportunities for an internal attacker also from our point of view.

Just as an another confused deputy scenario: assume there is a malicious user in a team, confined in a namespace by  RBAC. They might learn about interesting attack targets in the cluster via Envoy metrics (exposure controlled by Contour). The target might be protected by Contour External Authorization. Internal TLS (Envoy->backend) or Kubernetes network policy might be set up, so that attacker cannot directly gain access to the target via cluster internal network, but (depending on the RBAC settings) they might have the capability to create `Endpoints` in their own namespace to expose the target, bypassing authentication and authorization.

### Other thoughts

I wonder how Envoy should play part in this doc. There surely are hardening options that Contour sets in Envoy configuration (or in theory could fail to set). This is relevant in the "big picture" security of ingress but how is the document scoped regarding that?

youngnick:
Thanks for that detailed comment @tsaarni!. I've made a substantial update to the doc, if you could take another look and see what you think, I'd be very grateful.

xaleeks:
Thanks for the great writeup Nick.  This is a fantastic start, and I can think of some other sections to add over time which I’ll list here.  I don’t see any issues with going ahead and merging this and then iterating off it over time.  Great comments made by Tero as well.  

Other things worth considering adding to the doc :

A section on attacker motivations.  For ex., attackers can target bringing down ingress to cause widespread network outage or deny serviceability of particular applications, or attempt to alter http/https requests made to services.  They can also silently hijack requests to glean critical information.  

A section on blast radius and recovery that focuses on the various exploits, attacker’s gained level of access, and steps to recovery. For ex., what happens when particular resources are compromised (services, endpoints, HTTPProxy resources, Ingress resources, ExtensionService resources etc).  What remediation steps such as shutting down Envoy / Contour, reinstalling, deleting or reapplying resources, regenerating new keys etc should users take in each scenario.  How can the attacker be thwarted mid attack and the system be alerted that an intrusion is ongoing.  What forensics can then be taken afterwards to prevent a similar attack from happening.

What happens if the relevant Kubernetes secrets or Kubernetes master nodes/etcd are compromised.  The damage may very well be moot or no worse than if a contour resource is compromised but it would be good to have these documented somewhere (not sure if this should be public, just that it should exist for the maintainer)

What compensating mechanisms exist in the project today that demonstrates our focus on creating a secure ingress controller.  This could be prolific use of TLS for securing communication both internal (btw Contour and Envoy) and external (to clients, to upstream services), and specific cryptographic protocols employed that are NIST compliant.  

Finally, it may be good for us to start cataloging detailed list of access, tokens, and credentials used in Contour in table format, either in this doc or in a separate doc.  For each of each, we should document the 

- Type 
- Encryption standards
- How is it used
- Who supplies this
- Who can update this
- How is it stored in Contour
- How is forgery prevented or detected
- What happens if it’s compromised

tsaarni (commented):
Thanks for the update @youngnick, looking good!

Few more comments, again as separate reply since inline comments do not work for this PR.

> since access to that namespace equals access to all the secrets Contour can see (which is, by default, all the secrets relevant to ingress config in the cluster).

This is not strictly the case. Contour has access to any secret in the cluster, it listens for any secret update and copies of secrets are cached in memory:

Following is with freshly installed quickstart deployment 

```console
$  echo "Hello secret world!" > secret
$ kubectl create secret generic hello --from-file=secret
secret/hello created
$ sudo gcore -a $(pgrep -f "contour serve" | head -1)  # pick one of two contour instances, dump its memory to a file
Saved corefile core.2057282
$ strings core.2057282 | grep Hello
Hello secret world!
Hello secret world!
Hello secret world!
Hello secret world!
```

Although it would be nice to avoid this, I believe this is expected behavior from an controller. It makes it very interesting target for an attacker.

> As seen in CVE-2021-25740, it's possible to manipulate Endpoints and EndpointSlices to access services in other clusters. 

Did you mean "in other namespaces of the cluster"?

stevesloka:
- nit: Line 21:  Remove `thye`
- Line 36: I think this is what @tsaarni was suggesting, but access to the namespace isn't really access to the secrets. Those are baked into Contour/Envoy memory. The comment suggests that the secrets are written to that namespace. 
- Line 46: I think the `ReferencePolicy` in GatewayAPI is too loose since it exposes all types in the target namespaces whereas TLSCertificateDelegation exposes a single secret. 
- Line 50: I think HTTPProxy does this for you. You can specify auth at the root and the children cannot bypass that.
- Line 63: This talks about vulnerabilities and not external access to Envoy
- Line 73: Could someone abuse the configmap that contour consumes? We parse that with a yaml parser.

stevesloka (changes requested):
nit: Line 21: Remove thye
Line 36: I think this is what @tsaarni was suggesting, but access to the namespace isn't really access to the secrets. Those are baked into Contour/Envoy memory. The comment suggests that the secrets are written to that namespace.
Line 46: I think the ReferencePolicy in GatewayAPI is too loose since it exposes all types in the target namespaces whereas TLSCertificateDelegation exposes a single secret.
Line 50: I think HTTPProxy does this for you. You can specify auth at the root and the children cannot bypass that.
Line 63: This talks about vulnerabilities and not external access to Envoy
Line 73: Could someone abuse the configmap that contour consumes? We parse that with a yaml parser.

youngnick:
I think that the commenting doesn't work because the file used to be a symlink, and so the diff doesn't register, sorry everyone, and thanks for the comments anyway.

I've just pushed an update that should fix @tsaarni and @stevesloka's comments. I'm having a think about what to add out of @xaleeks' excellent suggestions, I don't want this file to be an exhaustive list of all the things that could happen, it's intended to be a summary of what we think about in this area. An exhaustive list is going to be exhausting to keep up to date. 😃

youngnick:
In terms of what @xaleeks has suggested:

> A section on attacker motivations. 

This is a good idea, I'll bring something like this in.

> A section on blast radius and recovery that focuses on the various exploits, attacker’s gained level of access, and steps to recovery. 

I think this is good but should be a separate document, as it will change more frequently than this one.

> What happens if the relevant Kubernetes secrets or Kubernetes master nodes/etcd are compromised. 
In this case, the whole cluster is lost, the game is over, I don't think it's worth considering here, really. It's the same if an attacker gets cluster-admin, then you better hope you have good logging and auditing.

> What compensating mechanisms exist in the project today that demonstrates our focus on creating a secure ingress controller. This could be prolific use of TLS for securing communication both internal (btw Contour and Envoy) and external (to clients, to upstream services), and specific cryptographic protocols employed that are NIST compliant.
I can update some of the wording to make this clearer, but this is the point of the document. 😄 

> Finally, it may be good for us to start cataloging detailed list of access, tokens, and credentials used in Contour in table format, either in this doc or in a separate doc. 

Definitely a separate doc for this one. I'll put something like this at the top, but my intended audience here is:
- people who want to know more about what we consider important in terms of security
- security researchers looking for places to hit.

Practically, there *is* a risk in being this open about our security - this information makes it easier and lower-effort to start attacking Contour. My feeling is that the benefits of making it easier for security researchers to check Contour out without needing to read the whole code base outweigh the risks of giving bad actors easy access to the same information (which you could figure out by reading the code, issues, and PRs anyway).

tsaarni:
I've re-read the doc, looks good to me :+1:

youngnick:
Okay, I've just pushed one more round of changes, I think this is pretty close to done.

stevesloka:
Should the markdown file just point to the site page? If not then you've got to always keep the two in sync.

site/content/resources/security-threat-model.md
```diff
@@ -0,0 +1,101 @@
...
+
+We anticipate that the most likely attacks are created by the relatively untrusted application developer users, whether they are malicious or not. We expect the most likely attacks to be:
+- Confused deputy attacks - since Contour is trusted to build config and send to Envoy, that access can be misused to produce insecure Envoy configurations. [ExternalName Services can be used to gain access to Envoy's admin interface](https://github.com/projectcontour/contour/security/advisories/GHSA-5ph6-qq5x-7jwc) was an example of this attack in action, and was specifically dealt with by disallowing ExternalName services by default, and by removing the Envoy admin interface from use across any network, even localhost.
+- Insecure or conflicting configurations produced my manipulation of Kubernetes objects used for configuration.
+
+Our general method of mitigating both of these styles of attack is to be proscriptive about what configurations Contour will accept. Obviously, in cases like the ExternalName issue above, it's possible for a syntatically and allowed configuration to produce an insecure Envoy config, and this is therefore a primary focus of our thread model.
```
sunjayBhatia, line 36: "threat model" instead of "thread" at the end of this sentence here

sunjayBhatia: same for the `SECURITY.md` mirrored file as well

sunjayBhatia:
> Should the markdown file just point to the site page? If not then you've got to always keep the two in sync.

+1 if we can consolidate

xaleeks:
> Practically, there is a risk in being this open about our security - this information makes it easier and lower-effort to start attacking Contour. My feeling is that the benefits of making it easier for security researchers to check Contour out without needing to read the whole code base outweigh the risks of giving bad actors easy access to the same information (which you could figure out by reading the code, issues, and PRs anyway).

You're right, this should be in a private doc accessible to Contour security team and members of CNCF only

youngnick:
The reason I didn't consolidate is that Github's UI draws you to the SECUIRTY.md file, and I didn't want that to be just a pointer. I only noticed that after the previous time, when I did pull everything into the site file and change the SECURITY.md file to a symlink. In that case, Github doesn't let you just click the SECURITY.md file and view the contents. Maybe instead, I'll just have SECURITY.md link to the various site pages.

> You're right, this should be in a private doc accessible to Contour security team and members of CNCF only

I strongly disagree, this information is already accessible just by reading the code, being open about our security model shows our confidence in what we've built. We're open source, security by obscurity is a non-starter.

youngnick:
Okay, I think this is ready to go. Could I get reviews again please? @stevesloka @skriss @sunjayBhatia ?

youngnick:
Oh, and @tsaarni too, a review would be very welcomed.

site/content/resources/security-threat-model.md
sunjayBhatia: this nit still outstanding

sunjayBhatia (commented):
i think in the maintainer meeting we chatted about changing the SECURITY.md file to point to relevant website pages rather than contain the same content

tsaarni (approved):
LGTM!

youngnick:
> i think in the maintainer meeting we chatted about changing the SECURITY.md file to point to relevant website pages rather than contain the same content

You're totally right @sunjayBhatia, I've done that now.

site/content/resources/security-threat-model.md
sunjayBhatia: @youngnick I'm going to approve and merge and then fix this little one

sunjayBhatia (approved):

PR #4013, opened 2021-09-02
apis: ContourConfig CRD implementation

stevesloka:
Initial implementation of a ContourConfiguration CRD. Still needs work, but just shows an initial wip.


stevesloka:
Might look at breaking apart the configuration.go file, it seems just too big to manage. =(

apis/projectcontour/v1alpha1/accesslog.go
```diff
@@ -0,0 +1,251 @@
...
+type AccessLogType string
+
+func (a AccessLogType) Validate() error {
+	switch a {
+	case EnvoyAccessLog, JSONAccessLog:
+		return nil
+	default:
+		return fmt.Errorf("invalid access log format %q", a)
+	}
+}
```
skriss, line 135: Would it be worth moving these validation helpers into an internal package so that we didn't have to deal with API compatibility guarantees? I'd sort of prefer that all else being equal.

Related, wondering if there's a way to reuse implementations across here and `pkg/config` so we don't have to maintain both. I know the `pkg/config` ones will go away eventually, but probably not until 2.0 since they're in a public package. Perhaps we could internally convert the config file structs into the CRD structs (or vice versa) and then just have one validation implementation. Would have to poke around a little more to see exactly what that would look like.

skriss (commented):
Working on a detailed field-level review, just a couple initial topics for discussion

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -22,6 +25,396 @@ import (
...
 type ContourConfigurationSpec struct {
+	// XDSServer contains parameters for the xDS server.
+	// +optional
+	XDSServer ServerParameters `json:"xdsServer,omitempty"`
+
+	// IngressOptions contains parameters for ingress options.
+	// +optional
+	IngressOptions Ingress `json:"ingress,omitempty"`
+
+	EnvoyConfig Envoy `json:"envoy,omitempty"`
```
skriss, line 36: needs godoc

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -22,6 +25,396 @@ import (
...
+	// +optional
+	EnableExternalNameService bool `json:"enableExternalNameService,omitempty"`
+
+	// RateLimitService optionally holds properties of the Rate Limit Service
+	// to be used for global rate limiting.
+	RateLimitService RateLimitService `json:"rateLimitService,omitempty"`
+
+	// Policy specifies default policy applied if not overridden by the user
+	Policy PolicyParameters `json:"policy,omitempty"`
+}
```
skriss, line 62: There's a mix of "[area]Config", "[area]Parameters", "[area]Options", and just "[area]" in the struct and field names, it'd be nice to be consistent. Perhaps we can just go with e.g. `LeaderElection`, `Gateway`, etc. for field names (since they're already within a `ContourConfigurationSpec` struct so repeating `Config` is probably redundant)? Struct names could either match the field names, or be e.g. `LeaderElectionConfig`, `GatewayConfig`, etc.

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -22,6 +25,396 @@ import (
...
+
+// PolicyParameters holds default policy used if not explicitly set by the user
+type PolicyParameters struct {
+	// RequestHeadersPolicy defines the request headers set/removed on all routes
+	RequestHeadersPolicy HeadersPolicy `json:"requestHeaders,omitempty"`
+
+	// ResponseHeadersPolicy defines the response headers set/removed on all routes
+	ResponseHeadersPolicy HeadersPolicy `json:"responseHeaders,omitempty"`
+}
```
skriss, line 389: I'm thinking these fields could move into `HTTPProxyConfig` since (a) `Policy` is a pretty vague name and (b) these only get applied to `HTTPProxy` config anyway. I guess though if we ever wanted to apply these to Ingress/Gateway routes, that wouldn't make sense. Thoughts?

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -22,6 +25,396 @@ import (
...
+type Metrics struct {
+	// Defines the Contour metrics address interface.
+	//  +optional
+	// +kubebuilder:validation:MinLength=1
+	Address string `json:"address,omitempty"`
+
+	// Defines the Contour metrics port.
+	// Defaults to 8000.
+	//  +optional
+	Port int `json:"port,omitempty"`
```
skriss, line 112: can set a kubebuilder default here

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -22,6 +25,396 @@ import (
...
+	// +optional
+	Logging EnvoyLogging `json:"logging,omitempty"`
+
+	// DefaultHTTPVersions defines the default set of HTTPS
+	// versions the proxy should accept. HTTP versions are
+	// strings of the form "HTTP/xx". Supported versions are
+	// "HTTP/1.1" and "HTTP/2".
+	//
+	// If this field not specified, all supported versions are accepted.
+	DefaultHTTPVersions []HTTPVersionType `json:"defaultHTTPVersions"`
```
skriss, line 152: add `omitempty` tag

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -22,6 +25,396 @@ import (
...
+	// are encountered.
+	// See: https://github.com/projectcontour/contour/issues/3221
+	//  +optional
+	DisableAllowChunkedLength bool `json:"disableAllowChunkedLength,omitempty"`
+
+	// ConnectionBalancer. If the value is exact, the listener will use the exact connection balancer
+	// See https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/listener.proto#envoy-api-msg-listener-connectionbalanceconfig
+	// for more information.
+	//  +optional
+	ConnectionBalancer string `json:"connection-balancer"`
```
skriss, line 186: `omitempty`

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -22,6 +25,396 @@ import (
...
+	// +optional
+	GatewayConfig *GatewayParameters `json:"gateway,omitempty"`
+
+	// HTTPProxy defines parameters on HTTPProxy.
+	// +optional
+	HTTPProxy *HTTPProxyConfig `json:"httpproxy,omitempty"`
+
+	// LeaderElection contains leader election parameters.
+	// +optional
+	LeaderElection LeaderElectionParameters `json:"leaderelection,omitempty"`
```
skriss, line 49: Using this field as an example - currently, we have an [internal set of defaults](https://github.com/projectcontour/contour/blob/main/pkg/config/parameters.go#L722-L728) that are applied if the user does not specify values, but with the ConfigMap those defaults weren't really visible to the user anywhere other than documentation. Now that we're using a CRD, I'm thinking it would make sense to configure those as kubebuilder defaults, so that the default values are actually reflected in the CRD spec when a user creates a `ContourConfiguration`. WDYT? If that makes sense, then I guess we should go through all the defaults in the `Defaults()` func and set them up as kubebuilder defaults here.

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -22,6 +25,396 @@ import (
...
+	// +kubebuilder:validation:MinLength=1
+	Address string `json:"address"`
+
+	// Defines an Envoy listener Port.
+	Port int `json:"port"`
+
+	// AccessLog defines where Envoy logs are outputted for this listener.
+	// +optional
+	// +kubebuilder:default=/dev/stdout
+	AccessLog string `json:"access_log,omitempty"`
```
skriss, line 223:
```suggestion
	AccessLog string `json:"accessLog,omitempty"`
```

skriss (changes requested):
Few more comments, will need to do another pass or two though

apis/projectcontour/v1alpha1/contourconfig.go
skriss: Heh OK just saw https://github.com/projectcontour/contour/pull/4018 so probably doesn't make sense to move this into `HTTPProxy`. If we can think of a more descriptive name than `Policy`, great, otherwise, not a big deal.

apis/projectcontour/v1alpha1/accesslog.go
stevesloka: yeah I agree and brought that up before but the team at the time felt we should have them in the public API. 

Some of the issue with the `pkg/config` resuse is that the names not being camel-cases and other groupings. 

For validation, we may need to convert Contour internally to use the CRD configuration spec and deprecate out the `ServeContext`, having a converter migrate the `ServeContext` into a `ContourConfiguration`.

apis/projectcontour/v1alpha1/contourconfig.go
sunjayBhatia: +1 for me, that sounds reasonable if it works well with kubebuilder etc.

apis/projectcontour/v1alpha1/accesslog.go
youngnick: I pushed for these to be in here so that anyone who consumes HTTPProxy or any Contour CRDs would have the same validation logic available to them, which makes actually reusing them in other cases more likely to be consistent.

I think that we should absolutely be using only one representation internally inside Contour, and ideally it should be these new structs (since this is where we are heading). But I'm not sure if we do that now or later. In either case, we're going to end up doing work that will end up thrown away translating from one thing to another, I think that the decision is more around how much do we do before 1.19 in the initial release vs later as refactoring.

apis/projectcontour/v1alpha1/contourconfig.go
youngnick: 💯 percent agree that we should use CRD defaulting and move away from the `Defaults()` pattern wherever possible. We should then get documentation of the config format and its defaults for close to free with the already-existing reference generation that parses the types.

apis/projectcontour/v1alpha1/accesslog.go
youngnick: In this specific case, using a kubebuilder validation is actually heaps better, since that will mean that the object can't be applied unless the validation passes. Kubebuilder has enums available already too.

apis/projectcontour/v1alpha1/server.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+}
+
+// ServerParameters holds the config for the Contour xDS server.
+type ServerParameters struct {
+	// Defines the XDSServer to use for `contour serve`.
+	// Defaults to "contour"
+	// +kubebuilder:default=contour
+	// +kubebuilder:validation:Enum=contour;envoy
+	Type XDSServerType `json:"type,omitempty"`
```
youngnick, line 42: This field doesn't need `omitempty` because it's required and an enum.

apis/projectcontour/v1alpha1/server.go
```diff
@@ -0,0 +1,72 @@
...
+	// Defaults to false.
+	//  +optional
+	// +kubebuilder:default=false
+	Insecure bool `json:"insecure,omitempty"`
+
+	// TLS holds TLS file config details.
+	//  +optional
+	TLS *TLS `json:"tls,omitempty"`
+
+	// Health defines the endpoints Contour will serve to enable health checks.
```
youngnick, line 67: Needs a `+optional` tag, as does `Metrics` below.

apis/projectcontour/v1alpha1/server.go
```diff
@@ -0,0 +1,72 @@
...
+// limitations under the License.
+
+package v1alpha1
+
+import (
+	"fmt"
+)
+
+// XDSServerType is the type of xDS server implementation.
+type XDSServerType string
```
youngnick, line 21: Kubebuilder validation will let us throw away `Validate()` on this type.

apis/projectcontour/v1alpha1/tlsciphers.go
```diff
@@ -0,0 +1,101 @@
...
+
+package v1alpha1
+
+import (
+	"fmt"
+	"strings"
+)
+
+// TLSCiphers holds a list of TLS ciphers
+type TLSCiphers []string
```
youngnick, line 22: I'm not sure about the maximum lengths for kubebuilder validations, but if we can fit the ciphersuites into kubebuilder tags, this whole file can just be the typedefs with kubebuilder tags instead.

apis/projectcontour/v1alpha1/server.go
```diff
@@ -0,0 +1,72 @@
...
+	// +kubebuilder:default=contour
+	// +kubebuilder:validation:Enum=contour;envoy
+	Type XDSServerType `json:"type,omitempty"`
+
+	// Defines the xDS gRPC API address which Contour will serve.
+	// Defaults to "0.0.0.0".
+	//  +optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:default="0.0.0.0"
+	Address string `json:"address,omitempty"`
```
youngnick, line 49: In fields like this where it's optional but has a default, `omitempty` will never do anything because the field can never be empty. It will always have at least the default value. So I guess I think we should take it out.

apis/projectcontour/v1alpha1/server.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// Allow serving the xDS gRPC API without TLS.
+	// Defaults to false.
+	//  +optional
+	// +kubebuilder:default=false
+	Insecure bool `json:"insecure,omitempty"`
+
+	// TLS holds TLS file config details.
+	//  +optional
+	TLS *TLS `json:"tls,omitempty"`
```
youngnick, line 65: This is only FYI, but this is the correct way to do optional struct fields.

They have to have:
- `+optional`
- be a pointer
- `omitempty`

For everything to work properly. I've got burnt by this a few times before, so just thought I would note it.

youngnick (changes requested):
This is a great start, but I think that, since we are building a CRD here, we should lean hard on Kubebuilder/apiserver validation and try to get rid of the `Validate()` methods.

That means that invalid config CRDs won't be accepted by the apiserver.

apis/projectcontour/v1alpha1/accesslog.go
stevesloka: Just need to be careful about some of that. IngressClass is a great example. If it's not defined, Contour looks at ALL ingress objects, if it's defaulted to `Contour` then it won't process any resource without a class defined.

apis/projectcontour/v1alpha1/contourconfig.go
stevesloka: To do that we'll need to make different structs for each type as Envoy's metrics are different defaults than Contour's.

apis/projectcontour/v1alpha1/contourconfig.go
stevesloka: Yup agree and doing that in many places. Just need to be aware of defaulting the wrong things (https://github.com/projectcontour/contour/pull/4013#discussion_r705359031)

apis/projectcontour/v1alpha1/accesslog.go
skriss: > I think that we should absolutely be using only one representation internally inside Contour, and ideally it should be these new structs (since this is where we are heading). But I'm not sure if we do that now or later. In either case, we're going to end up doing work that will end up thrown away translating from one thing to another, I think that the decision is more around how much do we do before 1.19 in the initial release vs later as refactoring.

I agree, we can definitely revisit this after the initial release. Ultimately we might decide it's not worth it given that we'll throw the old stuff away.

apis/projectcontour/v1alpha1/contourconfig.go
skriss: Ah. I think you can also set a default at the place where the struct is used as a field - see e.g. https://github.com/projectcontour/contour-operator/blob/main/api/v1alpha1/contour_types.go#L73-L74

examples/contour/01-contour-config-crd.yaml
```diff
@@ -0,0 +1,108 @@
+apiVersion: projectcontour.io/v1alpha1
```
stevesloka, line 1: I have this just to play around with applying various configs + testing defaults. We'll need to clean this up and come up with a valid "default".

apis/projectcontour/v1alpha1/tlsciphers.go
stevesloka: I think I got 'em in there, API server didn't care much. =)

examples/contour/01-contour-config-crd.yaml
stevesloka: aaaand I took it out. Was causing all kinds of issues. =)

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -22,6 +22,520 @@ import (
...
+// TLS holds TLS file config details.
+type TLS struct {
+	// CA filename.
+	CAFile string `json:"caFile"`
+
+	// Client certificate filename.
+	CertFile string `json:"certFile"`
+
+	// Client key filename.
+	KeyFile string `json:"keyFile"`
```
skriss, line 124: Should there be defaults here? E.g. https://github.com/projectcontour/contour/blob/main/examples/contour/03-contour.yaml#L43-L45?

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -22,6 +22,520 @@ import (
...
+	AccessLogFormatString *string `json:"accessLogFormatString,omitempty"`
+
+	// AccessLogFields sets the fields that JSON logging will
+	// output when AccessLogFormat is json.
+	// +optional
+	AccessLogFields AccessLogFields `json:"jsonFields,omitempty"`
+}
+
+// TimeoutParameters holds various configurable proxy timeout values.
+type TimeoutParameters struct {
```
skriss, line 352: I did figure out a validation for timeout strings that you could use here: https://github.com/projectcontour/contour/blob/main/apis/projectcontour/v1/httpproxy.go#L622

Perhaps worth defining a timeout string type alias too, to consolidate the validation instead of copy-pasting it everywhere, but we can do that separately.

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -22,6 +22,520 @@ import (
...
+
+	// EnableExternalNameService allows processing of ExternalNameServices
+	// Defaults to disabled for security reasons.
+	// +optional
+	// +kubebuilder:default:false
+	EnableExternalNameService bool `json:"enableExternalNameService,omitempty"`
+
+	// RateLimitService optionally holds properties of the Rate Limit Service
+	// to be used for global rate limiting.
+	RateLimitService RateLimitServiceConfig `json:"rateLimitService,omitempty"`
```
skriss, line 65: this one can be a pointer since it's entirely optional

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -22,6 +22,520 @@ import (
...
+
+	// Policy specifies default policy applied if not overridden by the user
+	Policy PolicyConfig `json:"policy,omitempty"`
```
skriss, line 68: can be a pointer since entirely optional

apis/projectcontour/v1alpha1/contourconfig.go
stevesloka: I don't know how to "turn them off" which is why I didn't provide, meaning, if you want to run `--insecure`, but don't specify the fields, you still get values.

youngnick: I spent a few hours playing with various kubebuilder tags, and confirmed that it works this way.

If you provide defaults inside a struct that has `+optional` and `omitempty`, then those fields will only be used if you specify the struct at all - that is, if it's not empty. But, that does mean that there's no way to default them usefully, using only the kubebuilder defaults, in this `--insecure` case.

So, here, I think we should set the defaults to what Contour *will* default them to, even if it's not used (that is, the values used in the example deployment).

youngnick (commented):
This is looking even better. I've been finding how validation, `+optional`, and `omitempty`, and default values intersect very confusing though, so I spent some time today playing around. I think we should discuss this in a meeting soon.

apis/projectcontour/v1alpha1/contourconfig.go
stevesloka: Just to clarify @youngnick, even if we set `insecure: true` you'd still get defaults for the keys (ca, cert, key) even though they aren't used?

youngnick: The defaults won't get used unless at least the struct is non-nil. Which would mean that, in YAML form, it's unspecified, which I would guess would be the case in this one. So, if we set defaults *inside* the struct, they won't be used as long as we don't set a default at the outer level.

youngnick:
I think we need to talk about this one in the maintainer's meeting. I'll add it to the agenda for early next week.

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -21,7 +21,519 @@ import (
...
...e: "/certs/ca.crt", certFile: "/certs/tls.crt", keyFile: "/certs/tls.key", insecure: false}}, debug: {logLevel: "info", kubernetesLogLevel: 0}, health: {address: "0.0.0.0", port: 8000}, envoy: {listener: {useProxyProtocol: false, disableAllowChunkedLength: false, connectionBalancer: "", tls: { minimumProtocolVersion: "1.2", cipherSuites: "[ECDHE-ECDSA-AES128-GCM-SHA256|ECDHE-ECDSA-CHACHA20-POLY1305]";"[ECDHE-RSA-AES128-GCM-SHA256|ECDHE-RSA-CHACHA20-POLY1305]";"ECDHE-ECDSA-AES256-GCM-SHA384";"ECDHE-RSA-AES256-GCM-SHA384" }}, service: {name: "envoy", namespace: "projectcontour"}, http: {address: "0.0.0.0", port: 8080, accessLog: "/dev/stdout"}, https: {address: "0.0.0.0", port: 8443, accessLog: "/dev/stdout"}, metrics: {address: "0.0.0.0", port: 8002}, logging: { accessLogFormat: "envoy"}, defaultHTTPVersions: "http/1.1";"http/2", cluster: {dnsLookupFamily: "auto"}, network: { adminPort: 9001}}, httpproxy: {disablePermitInsecure: false}, leaderElection: {leaseDuration: "15s", renewDeadline: "10s", retryPeriod: "2s", disableLeaderElection: false, configmap: {name: "leader-elect", namespace: "projectcontour"}}, enableExternalNameService: false, metrics: {address: "0.0.0.0", port: 8000}
```
skriss, line 24: One option to consider here would be leaving off this aggregated default for the `spec` field itself. Since `spec` is a required field, the UX would be that a user would create a `ContourConfig` like the following:

```yaml
kind: ContourConfiguration
metadata:
  name: contour-config
  namespace: default
spec: {}
```

Then they'll still get all the defaults for the subfields of `spec`:

```yaml
kind: ContourConfiguration
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"projectcontour.io/v1alpha1","kind":"ContourConfiguration","metadata":{"annotations":{},"name":"contour-config","namespace":"default"},"spec":{}}
  creationTimestamp: "2021-09-17T16:18:54Z"
  generation: 1
  name: contour-config
  namespace: default
  resourceVersion: "175456"
  uid: 907940dd-39c7-4fe5-ae14-001061dcb0aa
spec:
  debug:
    kubernetesLogLevel: 0
    logLevel: info
  enableExternalNameService: false
  envoy:
    cluster:
      dnsLookupFamily: auto
    defaultHTTPVersions:
    - http/1.1
    - http/2
    http:
      accessLog: /dev/stdout
      address: 0.0.0.0
      port: 8080
    https:
      accessLog: /dev/stdout
      address: 0.0.0.0
      port: 8443
    listener:
      connectionBalancer: ""
      disableAllowChunkedLength: false
      tls:
        cipherSuites:
        - '[ECDHE-ECDSA-AES128-GCM-SHA256|ECDHE-ECDSA-CHACHA20-POLY1305]'
        - '[ECDHE-RSA-AES128-GCM-SHA256|ECDHE-RSA-CHACHA20-POLY1305]'
        - ECDHE-ECDSA-AES256-GCM-SHA384
        - ECDHE-RSA-AES256-GCM-SHA384
        minimumProtocolVersion: "1.2"
      useProxyProtocol: false
    logging:
      accessLogFormat: envoy
    metrics:
      address: 0.0.0.0
      port: 8002
    network:
      adminPort: 9001
    service:
      name: envoy
      namespace: projectcontour
  health:
    address: 0.0.0.0
    port: 8000
  httpproxy:
    disablePermitInsecure: false
  leaderElection:
    configmap:
      name: leader-elect
      namespace: projectcontour
    disableLeaderElection: false
    leaseDuration: 15s
    renewDeadline: 10s
    retryPeriod: 2s
  metrics:
    address: 0.0.0.0
    port: 8000
  xdsServer:
    address: 0.0.0.0
    port: 8001
    tls:
      caFile: /certs/ca.crt
      certFile: /certs/tls.crt
      insecure: false
      keyFile: /certs/tls.key
    type: contour
```

The benefit is that we don't have to keep defaults in sync in multiple places.

stevesloka: That's how this should work with my last commit. If you specify an empty `spec: {}` you get a full config, otherwise you can override the pieces you'd like and it will still default the rest. 

For example, here's how you'd disable tls communication between contour<>envoy:
```
kind: ContourConfiguration
metadata:
  name: contour
  namespace: projectcontour
spec:
  xdsServer:
    type: contour
    address: 0.0.0.0
    port: 8001
    tls:
      insecure: true
```

```
$ kubectl get contourconfig -o yaml -n projectcontour contour                                                                                                                  

kind: ContourConfiguration
metadata:
  name: contour
  namespace: projectcontour
spec:
  debug:
    kubernetesLogLevel: 0
    logLevel: info
  enableExternalNameService: false
  envoy:
    cluster:
      dnsLookupFamily: auto
    defaultHTTPVersions:
    - http/1.1
    - http/2
    http:
      accessLog: /dev/stdout
      address: 0.0.0.0
      port: 8080
    https:
      accessLog: /dev/stdout
      address: 0.0.0.0
      port: 8443
    listener:
      connectionBalancer: ""
      disableAllowChunkedLength: false
      tls:
        cipherSuites:
        - '[ECDHE-ECDSA-AES128-GCM-SHA256|ECDHE-ECDSA-CHACHA20-POLY1305]'
        - '[ECDHE-RSA-AES128-GCM-SHA256|ECDHE-RSA-CHACHA20-POLY1305]'
        - ECDHE-ECDSA-AES256-GCM-SHA384
        - ECDHE-RSA-AES256-GCM-SHA384
        minimumProtocolVersion: "1.2"
      useProxyProtocol: false
    logging:
      accessLogFormat: envoy
    metrics:
      address: 0.0.0.0
      port: 8002
    network:
      adminPort: 9001
    service:
      name: envoy
      namespace: projectcontour
  health:
    address: 0.0.0.0
    port: 8000
  httpproxy:
    disablePermitInsecure: false
  leaderElection:
    configmap:
      name: leader-elect
      namespace: projectcontour
    disableLeaderElection: false
    leaseDuration: 15s
    renewDeadline: 10s
    retryPeriod: 2s
  metrics:
    address: 0.0.0.0
    port: 8000
  xdsServer:
    address: 0.0.0.0
    port: 8001
    tls:                                # <----- tls bits not defaulted
      insecure: true
    type: contour
```

skriss: Yeah, I'm just suggesting that we can remove L24 here so we don't have to maintain it/keep it in sync with the lower-level defaults going fwd; even without it, we'll still get all the defaulting when the user specifies `spec: {}`.

stevesloka: Ahh I see, removed that. =)

skriss (commented):
I think this looks like it's in pretty good shape. I tested it out and was able to create a `ContourConfig` and get all the defaults; override some of my commonly used fields; and see the validation at work. 

I did spot a handful of places where it looks like we could copy the defaults that are applied at the top level down to lower levels of nesting, which would improve UX when a user wants to override just a single nested field, but I don't think it's blocking and would rather get this merged to unblock the rest of this work. One example of this, just to make it concrete, would be for the `EnvoyConfig.Listener` field. A default is applied for that within the kubebuilder default on the top-level `Envoy` field (see [here](https://github.com/projectcontour/contour/blob/7e9366df26a336709d967f9b1bae46eb9add67ce/apis/projectcontour/v1alpha1/contourconfig.go#L49), but then it's not applied on the [`Listener` field itself within the Envoy struct](https://github.com/projectcontour/contour/blob/7e9366df26a336709d967f9b1bae46eb9add67ce/apis/projectcontour/v1alpha1/contourconfig.go#L180-L181). So, if a user overrides anything within the `Envoy` block, they now have to explicitly specify all of the `Listener` options, even if they just want the defaults there.

Again, I don't think it needs to be blocking for this PR. I'll add an issue with all the spots where I think we could do something similar that we can follow up on separately.

sunjayBhatia:
> Again, I don't think it needs to be blocking for this PR. I'll add an issue with all the spots where I think we could do something similar that we can follow up on separately.

yeah I think since this is generally WIP and not a released thing yet, we could merge as-is and have some smaller PRs where we can address some of the details, would make moving forward faster since we can just make a PR to address a specific concern rather than making Steve S go back and change things how we describe

sunjayBhatia (approved):

youngnick (approved):
Agreed, let's get this in and make separate PRs to change the defaults.

skriss (approved):

PR #4014, opened 2021-09-03
docs: GatewayConfig updates

skriss:
Updates GatewayConfig docs to drop namespace
and name going forward, and update older
versions' docs to name the correct version
in which they are removed.

### DIFF
--- a/site/content/docs/v1.17.0/configuration.md
+++ b/site/content/docs/v1.17.0/configuration.md
@@ -183,7 +183,7 @@ The gateway configuration block is used to configure which gateway-api Gateway C
 | name (Deprecated) | string | contour | DEPRECATED: This field specifies the name of a Gateway.  |
 | namespace (Deprecated) | string | projectcontour | DEPRECATED: This field specifies the namespace of a Gateway.  |
 
-_NOTE: The fields `name` and `namespace` have been deprecated and will be removed in Contour v1.18.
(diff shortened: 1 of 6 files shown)


sunjayBhatia (approved):

site/content/docs/v1.17.0/configuration.md
```diff
@@ -183,7 +183,7 @@ The gateway configuration block is used to configure which gateway-api Gateway C
 | name (Deprecated) | string | contour | DEPRECATED: This field specifies the name of a Gateway.  |
 | namespace (Deprecated) | string | projectcontour | DEPRECATED: This field specifies the namespace of a Gateway.  |
 
-_NOTE: The fields `name` and `namespace` have been deprecated and will be removed in Contour v1.18.
+_NOTE: The fields `name` and `namespace` have been deprecated and will be removed in Contour v1.19.
```
stevesloka, line 186: nit: Should this read `has been removed in v1.19` since that release will be out?

skriss: I'm indifferent but if we're gonna word it that way should we wait until 1.19 is actually out?

stevesloka: I guess does the line even need to be here anymore? If it's deprecated and gone, can just delete.

skriss: WFM

youngnick (approved):

PR #4020, opened 2021-09-08
Pare down docs version in site dropdown

sunjayBhatia:
Include main plus last 5 minor releases in site dropdown (include patch releases)

Docs pages are retained to ensure links to old docs are still valid

### DIFF
--- a/changelogs/unreleased/4020-sunjayBhatia-docs.md
+++ b/changelogs/unreleased/4020-sunjayBhatia-docs.md
@@ -1,0 +1,1 @@
+Pare down docs versions available in site dropdown.


youngnick (approved):

stevesloka (commented):
This is probably ok to remove, the only good part about leaving it is that any links to older versions will still work. 

//cc @jonasrosland

jonasrosland:
As a preservationist it hurts me to delete docs like this :)
The dropdown menu should definitely be pared down to only show more recent versions, but I'm not sure what the harm is in keeping the files in place?

sunjayBhatia:
> As a preservationist it hurts me to delete docs like this :)
> The dropdown menu should definitely be pared down to only show more recent versions, but I'm not sure what the harm is in keeping the files in place?

Fine with closing this if this is overkill, just felt like this old content was probably not being used and adding to clutter in the repo

youngnick:
I approved because I kind of don't think that the really old docs are very relevant any more, but I don't see any issue with leaving them there if people feel strongly.

PR #4023, opened 2021-09-09
Design for HTTP cookie rewriting in Contour

sunjayBhatia:


design/cookie-rewrite-design.md
```diff
@@ -0,0 +1,196 @@
...
+	Name string `json:"name"`
+
+	// +optional
+	PathRewrite *CookiePathRewrite `json:"pathRewrite,omitempty"`
+
+	// +optional
+	DomainRewrite *CookieDomainRewrite `json:"domainRewrite,omitempty"`
+
+	// +optional
+	Secure *bool `json:"secure,omitempty"`
```
youngnick, line 94: Is there any way we can set defaults for these, or is that too much of a behavior change? I guess we could also do that dynamically later based on if the route is a HTTPS one or not.

youngnick (approved):
LGTM with one small question about defaults.

design/cookie-rewrite-design.md
sunjayBhatia: the intent of making all these pointers was to not try to suggest any defaults since I'm not sure we can know valid values in kubebuilder validation/defaulting (besides simple regex validation on names/paths/domains)

i think we could do something as you suggest to set the `Secure` attribute automatically if the rewrite policy is attached to an HTTPS route, but that would have to be in the processing stage

sunjayBhatia: I don't know if we can *always* set the `Secure` attribute for an HTTPS route since that might be an incorrect assumption, conversely we can probably assume that we *never* set `Secure` for plain HTTP routes however

design/cookie-rewrite-design.md
```diff
@@ -0,0 +1,196 @@
...
+`HTTPProxy` will have new fields to configure cookie rewrite policy.
+Cookie attributes related to infrastructure and security considerations will be customizable.
+Other fields, cookie names, and values will not be made customizable.
+The fields we will offer as customizable are as follows:
+- Path
+- Domain
+- Secure
+- SameSite
+
+Initially, the added field for cookie rewriting will be added to the `Route` and `Service` elements of the `HTTPProxy`.
```
stevesloka, line 51: What's the use-case of having the `Service` element be able to customize this? Feels like a route type configuration, but curious if there's a use-case I'm not thinking of.

sunjayBhatia: I just added it here to match the header request/response policies, but for all of these could be that one of the services (but not all) you forward traffic to returns some header/cookie values you want to modify. having it on the route only might mean you can't have customization per backend service, which is why i assume we added the per-service header customization options?

stevesloka: gotcha, that makes sense!

stevesloka (approved):

design/cookie-rewrite-design.md
```diff
@@ -0,0 +1,196 @@
...
+
+type CookieRewritePolicy struct {
+	// Name is the name of the cookie for which attributes will be rewritten.
+	Name string `json:"name"`
+
+	// +optional
+	PathRewrite *CookiePathRewrite `json:"pathRewrite,omitempty"`
+
+	// +optional
+	DomainRewrite *CookieDomainRewrite `json:"domainRewrite,omitempty"`
```
skriss, line 91: Are these two implemented as structs rather than just string pointers because there may be additional fields added within them over time?

skriss (approved):
Sorry I'm late to the party here, but this looks great - just one question.

design/cookie-rewrite-design.md
sunjayBhatia: yes! in case we want to do regex replace, or some sort of conditional rewrite

Issue #3574, opened 2021-04-14
Should globally configurable TLS options should be set on upstream TLS contexts?


sunjayBhatia:
Global TLS configuration here currently only applies to Envoy listeners: https://projectcontour.io/docs/v1.14.0/configuration/#tls-configuration

This means that settings like minimum TLS version and TLS ciphers are not applied to Envoy -> Upstream cluster communication or Envoy -> Extension Service cluster communication.

We do not seem to offer any configurability of these TLS connections.

Is there any reason we have not made these options configurable on the aforementioned connections?

Should we make the existing global configuration also apply to upstreams etc.?

youngnick:
I agree that we should allow configurability of this, and also agree that starting with making the existing global config also apply to upstreams is probably okay.

The only risk there is that we need to ensure that the certificates generated by `contour certgen` will work with TLS 1.3, and make sure that existing installations that have certs that don't won't break if we suddenly start enforcing that the Contour <-> Envoy connection must be TLS 1.3.

digibank-rons:
Hi,

So just to get it straight, currently there is no way of setting cipher suite for TLS connection?

Thanks

sunjayBhatia:
> Hi,
> 
> So just to get it straight, currently there is no way of setting cipher suite for TLS connection?
> 
> Thanks

You can for the connections between downstream clients and Envoy, however this does not apply to connections between Envoy and upstream clusters

xaleeks:
And to satisfy FIPS Compliance, I think Min TLS and TLS ciphers should apply to Envoy <> upstream services as well I would think. I’m tagging this v1.16 first, feel free to adjust it based on your priorities @sunjayBhatia

sunjayBhatia:
@xaleeks Envoy -> upstreams with a FIPS build of Envoy will already send FIPS compatible ciphers without configurability

Does this apply to FIPS really? Or does it more apply if you want to override the defaults that Envoy will choose? (to allow non-FIPS etc. while you are transitioning to FIPS or something or for legacy)

sunjayBhatia:
haven't had a chance to work on this yet and 1.16 release is v soon, should probably move this to the next milestone

sunjayBhatia:
[...]

PR #4024, opened 2021-09-10
Add option to set minProtocolVersion and ciphersuites for upstream co…

abhide:
…nnections

### DIFF
--- a/internal/dag/extension_processor.go
+++ b/internal/dag/extension_processor.go
@@ -35,6 +35,14 @@ type ExtensionServiceProcessor struct {
 	// secret containing client certificate and private key to be
 	// used when establishing TLS connection to upstream cluster.
 	ClientCertificate *types.NamespacedName
+
+	// MinimumTLSVersion is optional minimumTLSVersion to be used for client
+	// connections to upstream ExtensionCluster
+	MinimumTLSVersion string
+
+	// CipherSuites is optional cipher suites to be used for client connections
+	// to upstream ExtensionCluster
+	CipherSuites []string
 }
 
 var _ Processor = &ExtensionServiceProcessor{}
@@ -108,6 +116,11 @@ func (p *ExtensionServiceProcessor) buildExtensionService(
 		TimeoutPolicy:      tp,
 		SNI:                "",
 		ClientCertificate:  clientCertSecret,
+	}
+
+	if len(p.MinimumTLSVersion) != 0 {
+		extension.MinimumProtocolVersion = p.MinimumTLSVersion
+		extension.CipherSuites = p.CipherSuites
 	}
 
 	lbPolicy := loadBalancerPolicy(ext.Spec.LoadBalancerPolicy)
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -52,6 +52,14 @@ type GatewayAPIProcessor struct {
 	// This is normally disabled for security reasons.
 	// See https://github.com/projectcontour/contour/security/advisories/GHSA-5ph6-qq5x-7jwc for details.
 	EnableExternalNameService bool
+
+	// MinimumTLSVersion is optional minimumTLSVersion to be used for client
+	// connections to upstream ExtensionCluster
+	MinimumTLSVersion string
+
+	// CipherSuites is optional cipher suites to be used for client connections
+	// to upstream ExtensionCluster
+	CipherSuites []string
 }
 
 // matchConditions holds match rules.
@@ -847,6 +855,16 @@ func (p *GatewayAPIProcessor) routes(matchConditions []*matchConditions, headerP
 
 // cluster builds a *dag.Cluster for the supplied set of headerPolicy and service.
 func (p *GatewayAPIProcessor) cluster(headerPolicy *HeadersPolicy, service *Service, weight uint32) *Cluster {
+	if len(p.MinimumTLSVersion) != 0 {
+		return &Cluster{
+			Upstream:               service,
+			Weight:                 weight,
+			Protocol:               service.Protocol,
+			RequestHeadersPolicy:   headerPolicy,
+			MinimumProtocolVersion: p.MinimumTLSVersion,
+			CipherSuites:           p.CipherSuites,
+		}
+	}
 	return &Cluster{
 		Upstream:             service,
 		Weight:               weight,
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -855,22 +855,17 @@ func (p *GatewayAPIProcessor) routes(matchConditions []*matchConditions, headerP
 
 // cluster builds a *dag.Cluster for the supplied set of headerPolicy and service.
 func (p *GatewayAPIProcessor) cluster(headerPolicy *HeadersPolicy, service *Service, weight uint32) *Cluster {
-	if len(p.MinimumTLSVersion) != 0 {
-		return &Cluster{
-			Upstream:               service,
-			Weight:                 weight,
-			Protocol:               service.Protocol,
-			RequestHeadersPolicy:   headerPolicy,
-			MinimumProtocolVersion: p.MinimumTLSVersion,
-			CipherSuites:           p.CipherSuites,
-		}
-	}
-	return &Cluster{
+	cluster := &Cluster{
 		Upstream:             service,
 		Weight:               weight,
 		Protocol:             service.Protocol,
 		RequestHeadersPolicy: headerPolicy,
 	}
+	if len(p.MinimumTLSVersion) != 0 {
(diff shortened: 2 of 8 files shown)


abhide:
@sunjayBhatia Created a draft PR to address https://github.com/projectcontour/contour/issues/3574. 
Could you please take a look and let me know if I am on the correct path.

skriss:
@abhide this looks to me like it's headed in the right direction. You will also need to pass these config options to [the various DAG processors](https://github.com/projectcontour/contour/blob/main/cmd/contour/serve.go#L711-L733) so they can set these fields on the DAG Clusters/ExtensionClusters.

abhide:
> @abhide this looks to me like it's headed in the right direction. You will also need to pass these config options to [the various DAG processors](https://github.com/projectcontour/contour/blob/main/cmd/contour/serve.go#L711-L733) so they can set these fields on the DAG Clusters/ExtensionClusters.

Thanks @skriss. Updated the PR with changes to DAG processors. 

I have left out TCPProxy; I don't think these apply to TCPProxy, correct?

abhide:
Will add unit tests once these changes look good

internal/dag/gatewayapi_processor.go
```diff
@@ -847,6 +855,16 @@ func (p *GatewayAPIProcessor) routes(matchConditions []*matchConditions, headerP
 
 // cluster builds a *dag.Cluster for the supplied set of headerPolicy and service.
 func (p *GatewayAPIProcessor) cluster(headerPolicy *HeadersPolicy, service *Service, weight uint32) *Cluster {
+	if len(p.MinimumTLSVersion) != 0 {
```
sunjayBhatia, line 858: maybe refactor to something a little less repetetive here (and other places it may apply):

```
cluster := &Cluster{
  Upstream: service
  ...
}

// if minimum tls version and cipher suites set, then set them on `cluster`

return cluster
```

would make it easier if we need to add a new future field, don't have to add it in 2 places

internal/dag/httpproxy_processor.go
```diff
@@ -660,6 +668,11 @@ func (p *HTTPProxyProcessor) computeRoutes(
 				DNSLookupFamily:       string(p.DNSLookupFamily),
 				ClientCertificate:     clientCertSecret,
 			}
+
+			if len(p.MinimumTLSVersion) != 0 {
```
sunjayBhatia, line 672: here and other places, what if `CipherSuites` is set but not `MinimumTLSVersion`? seems like we would not get the ciphers set properly?

sunjayBhatia: seems like we would need separate checks for those two things being set

pkg/config/parameters.go
```diff
@@ -321,6 +321,10 @@ type TLSParameters struct {
 	// by advanced users. Note that these will be ignored when TLS 1.3 is in
 	// use.
 	CipherSuites TLSCiphers `yaml:"cipher-suites,omitempty"`
+
+	// EnableUpstreamTLS will apply MinimumProtocolVersion and CipherSuites on every
+	// Envoy -> UpstreamCluster communication and Envoy -> ExternalServiceCluster communication
+	EnableUpstreamTLS bool `yaml:"enable-upstream-tls"`
```
sunjayBhatia, line 327: is this the API we want @projectcontour/maintainers? this means you *must* set ciphers if you want to propagate these global tls settings to upstream connections, rather than being able to just set min tls version or ciphers and have that propagate

also could do with a little bit more specific naming of this field, maybe something that conveys more that the TLS settings in this section are going to be applied to upstream connections

sunjayBhatia: might want to change the godoc on the CipherSuites field too since that mentions Envoy listeners speciifcally, and now we want to apply this to upstream connections

youngnick: Yes, I think that users should just be able to enable upstream TLS and then have the upstream use the same values as whatever the downstream is configured to, by default. If people want to customise, they can.

internal/dag/gatewayapi_processor.go
abhide: Fixed

internal/dag/httpproxy_processor.go
youngnick: Yes, I think @sunjayBhatia is right here, users should be able to set the minimum TLS version and ciphersuites separately.

abhide: Ok. I will make the change accordingly. Would it make sense to add 2 separate flags to enable this feature in that case?
EnableUpstreamTLSVersion and EnableUpstreamTLSCipherSuites?

internal/dag/extension_processor.go
```diff
@@ -110,6 +118,11 @@ func (p *ExtensionServiceProcessor) buildExtensionService(
 		ClientCertificate:  clientCertSecret,
 	}
 
+	if len(p.MinimumTLSVersion) != 0 {
+		extension.MinimumProtocolVersion = p.MinimumTLSVersion
+		extension.CipherSuites = p.CipherSuites
+	}
```
skriss, line 124: I think per @sunjayBhatia's previous comment we should handle these independently, so:

```go
if len(p.MinimumTLSVersion) != 0 {
	extension.MinimumTLSVersion = p.MinimumTLSVersion
}
if len(extension.CipherSuites) != 0 {
	extension.CipherSuites = p.CipherSuites
}
```

internal/dag/gatewayapi_processor.go
```diff
@@ -847,12 +855,17 @@ func (p *GatewayAPIProcessor) routes(matchConditions []*matchConditions, headerP
...
+	cluster := &Cluster{
 		Upstream:             service,
 		Weight:               weight,
 		Protocol:             service.Protocol,
 		RequestHeadersPolicy: headerPolicy,
 	}
+	if len(p.MinimumTLSVersion) != 0 {
+		cluster.MinimumProtocolVersion = p.MinimumTLSVersion
+		cluster.CipherSuites = p.CipherSuites
+	}
```
skriss, line 867: ditto re: assigning each one independently based on whether it's configured on the `GatewayAPIProcessor`.

internal/dag/httpproxy_processor.go
skriss: I don't think we need two separate flags; we should just check each variable independently in the code.

internal/envoy/v3/auth.go
```diff
@@ -60,6 +60,21 @@ func UpstreamTLSContext(peerValidationContext *dag.PeerValidationContext, sni st
...
+func UpstreamTLSContextWithMinProtocolVersion(
+	peerValidationContext *dag.PeerValidationContext, sni string, clientSecret *dag.Secret, minProtocolVersion string, cipherSuites []string,
+	alpnProtocols ...string) *envoy_v3_tls.UpstreamTlsContext {
+	context := UpstreamTLSContext(peerValidationContext, sni, clientSecret, alpnProtocols...)
+	context.CommonTlsContext.TlsParams = &envoy_v3_tls.TlsParameters{
+		TlsMinimumProtocolVersion: ParseTLSVersion(minProtocolVersion),
+		CipherSuites:              cipherSuites,
+	}
+	return context
+}
```
skriss, line 76: Instead of defining a new function to handle the additional parameters, I might define an `UpstreamTLSContextOptions` struct that holds all of the possible parameters, e.g.:

```go
type UpstreamTLSContextOptions struct {
	PeerValidationContext *dag.PeerValidationContext
	SNI                   string
	ClientSecret          *dag.Secret
	ALPNProtocols         []string
	MinimumTLSVersion     string
	CipherSuites          []string
}
```

And then refactor the existing function to look like:

```go
func UpstreamTLSContext(opts UpstreamTLSContextOptions) *envoy_v3_tls.UpstreamTlsContext {
	...
}
```

Where you can check for the existence of each setting in the Options struct. That way it's just one function that can handle the presence/absence of any of the options.

It's not used in *too* many places so the refactor shouldn't be too bad.

This can be non-blocking, it's easy enough to refactor later too.

pkg/config/parameters.go
tsaarni: Although it would result in less streamlined API, I wonder if it would be safer to have full configurability. There might be different needs for downstream vs upstream. Maybe one could have strict TLS 1.3 only policy for cluster-internal traffic but less restrictive towards external clients (although the current API without upstream controls, is basically the opposite). Or for practical reasons some weaker TLS 1.2 ciphers might need to be enabled for external interface but are unnecessary for internal.

youngnick:
@abhide, this needs a rebase, but looks like it might not be too far from done aside from that?

abhide:
> @abhide, this needs a rebase, but looks like it might not be too far from done aside from that?

@youngnick yes, I will rebase this and fix review comments.

PR #4025, opened 2021-09-13
Add path based logging (#3994)

### LINKED ISSUE #3994: Support path-based logging for controller
As a cluster operator, I would like the ability to configure path-based logging for the Contour controller via a configuration flag.

poidag:
Adds an optional flag to configure file-based logging for the controller.

Closes #3994.

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -136,6 +136,7 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 	serve.Flag("contour-cafile", "CA bundle file name for serving gRPC with TLS.").Envar("CONTOUR_CAFILE").StringVar(&ctx.caFile)
 	serve.Flag("contour-cert-file", "Contour certificate file name for serving gRPC over TLS.").PlaceHolder("/path/to/file").Envar("CONTOUR_CERT_FILE").StringVar(&ctx.contourCert)
 	serve.Flag("contour-key-file", "Contour key file name for serving gRPC over TLS.").PlaceHolder("/path/to/file").Envar("CONTOUR_KEY_FILE").StringVar(&ctx.contourKey)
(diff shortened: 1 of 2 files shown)


cmd/contour/serve.go
```diff
@@ -136,6 +136,7 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 	serve.Flag("contour-cafile", "CA bundle file name for serving gRPC with TLS.").Envar("CONTOUR_CAFILE").StringVar(&ctx.caFile)
 	serve.Flag("contour-cert-file", "Contour certificate file name for serving gRPC over TLS.").PlaceHolder("/path/to/file").Envar("CONTOUR_CERT_FILE").StringVar(&ctx.contourCert)
 	serve.Flag("contour-key-file", "Contour key file name for serving gRPC over TLS.").PlaceHolder("/path/to/file").Envar("CONTOUR_KEY_FILE").StringVar(&ctx.contourKey)
+	serve.Flag("contour-log-file", "Optional Logging file path for the controller.").PlaceHolder("/path/to/file").Envar("CONTOUR_LOG_FILE").StringVar(&ctx.Config.ContourLogFile)
```
stevesloka, line 139: We're looking to deprecate these config flags, so I'd remove this and just utilize the config file.

cmd/contour/serve.go
```diff
@@ -195,6 +196,15 @@ func validateCRDs(dynamicClient dynamic.Interface, log logrus.FieldLogger) {
 
 // doServe runs the contour serve subcommand.
 func doServe(log logrus.FieldLogger, ctx *serveContext) error {
+	// Set log target.
+	if ctx.Config.ContourLogFile != "" {
+		contourLogFile, err := os.OpenFile(ctx.Config.ContourLogFile, os.O_WRONLY|os.O_CREATE, 0755)
+		if err != nil {
+			fmt.Println(err)
```
stevesloka, line 203: Contour should probably fail at this point if the file can't be created.

stevesloka (changes requested):
Just a couple small nits @poidag, but otherwise looks good!

Also, would be great to add some docs to the config page around this parameter: https://projectcontour.io/docs/main/configuration/

PR #4027, opened 2021-09-13
internal/dag: set TLSRoute condition when no backend weights

### LINKED ISSUE #3946: Set a condition if a TLSRoute is invalid
If there are no valid clusters or all forwardTos have a weight of 0, the route should get rejected. Need to set a condition on the route stating this happened so users can be made aware. 

_Originally posted by @skriss in https://github.com/projectcontour/contour/pull/3938#discussion_r680026979_

skriss:
Sets a "ValidForwardTos: false" condition on TLSRoutes when
no backend has a non-zero weight.

Fixes #3946.


internal/status/routeconditions.go
```diff
@@ -28,6 +28,7 @@ const ResourceTLSRoute = "tlsroutes"
 
 const ConditionNotImplemented gatewayapi_v1alpha1.RouteConditionType = "NotImplemented"
 const ConditionResolvedRefs gatewayapi_v1alpha1.RouteConditionType = "ResolvedRefs"
+const ConditionValidForwardTos gatewayapi_v1alpha1.RouteConditionType = "ValidForwardTos"
```
skriss, line 31: Note that this condition type and reason are not defined by Gateway API. Feel free to suggest alternatives if preferred.

sunjayBhatia: seems reasonable to me to add our own conditiontype/reason, don't see any upstream

seems like this reasonably captures what we're trying to convey

sunjayBhatia (approved):

internal/status/routeconditions.go
youngnick: Yeah, I don't think that the ResolvedRefs condition really matches this either, so we should do our own for now. Off the top of my head, I can't think of a better name though. I will also make a note to raise this upstream and see if we can put something into upstream.

youngnick (approved):
This LGTM, as I said, I will see about pushing something upstream for this as well.

internal/status/routeconditions.go
```diff
@@ -39,6 +40,7 @@ const ReasonDegraded RouteReasonType = "Degraded"
 const ReasonValid RouteReasonType = "Valid"
 const ReasonErrorsExist RouteReasonType = "ErrorsExist"
 const ReasonGatewayAllowMismatch RouteReasonType = "GatewayAllowMismatch"
+const ReasonNoWeightedForwardTos = "NoWeightedForwardTos"
```
stevesloka, line 43: nit: Should we name this `ZeroWeightedForwardTos` to say that they are explicitly zero and just not unset?

internal/status/routeconditions.go
stevesloka: We still need to set status on the Gateway with a resolved refs, but that can be a different issue.

stevesloka: We should also somehow build out some docs on this to explain how Contour will work since Gateway doesn't have any docs on how this works.

skriss:
I'll hold this until after gateway v1alpha2 so we don't have to rename things (ForwardTo -> BackendRef).

PR #4028, opened 2021-09-13
site: clarify docs on how idle timeout parameters work

### LINKED ISSUE #3972: Clarification needed for the idle timeout parameters
There are various idle timeout parameters that can be configured by the user.  The documentation could be clarified a bit regarding the purpose and effect of each one:

[`HTTPProxy.spec.routes.timeoutPolicy.idle`](https://projectcontour.io/docs/main/config/api/#projectcontour.io/v1.TimeoutPolicy)

> Timeout after which, if there are no active requests for this route, the connection between Envoy and the backend or Envoy and the external client will be closed. If not specified, there is no per-route idle timeout, though a connection manager-wide stream_idle_timeout default of 5m still applies.

I read this as if upstream HTTP connection would be closed where there is no new requests in that connection during configured period. In reality, it applies to single HTTP/1 request-response, or HTTP/2 stream. Timeout will not trigger when HTTP/1 connection is idling between two requests ([envoy doc](https://www.envoyproxy.io/docs/envoy/latest/faq/configuration/timeouts#route-timeouts)).

[`timeouts.connection-idle-timeout`](https://projectcontour.io/docs/main/configuration/#timeout-configuration)

> This field defines how long the proxy should wait while there are no active requests (for HTTP/1.1) or streams (for HTTP/2) before terminating an HTTP connection. Must be a valid Go duration string, or infinity to disable the timeout entirely. See the Envoy documentation for more information.

It could be clarified that it applies only to downstream connection ([envoy doc](https://www.envoyproxy.io/docs/envoy/latest/faq/configuration/timeouts#connection-timeouts)). Corresponding upstream connection idle timeout is not exposed currently, but Envoy has built-in default of 1h, xref #3313.

[`timeouts.stream-idle-timeout`](https://projectcontour.io/docs/main/configuration/#timeout-configuration)

> This field defines how long the proxy should wait while there is no request activity (for HTTP/1.1) or stream activity (for HTTP/2) before terminating the HTTP request or stream. Must be a valid Go duration string, or infinity to disable the timeout entirely. See the Envoy documentation for more information.

It could be clarified that "_request activity (for HTTP/1.1)_" means only the duration of a single request-response, and not the HTTP/1 connection ([envoy doc](https://www.envoyproxy.io/docs/envoy/latest/faq/configuration/timeouts#stream-timeouts)).

tsaarni:
Fixes #3972


youngnick (approved):
Thanks for this @tsaarni, the level of subtlety involved in describing exactly what the timeouts cover is an eternal headache.

PR #4037, opened 2021-09-17
Deprecate the static cluster `service-stats` 

### LINKED ISSUE #3959: Deprecate service-stats cluster for Envoy Admin
The envoy admin socket uses a static cluster named `service-stats`, this can be confusing since it's not really stats related. 

We should deprecate Envoy from using this and move to a clearer name. 

_Originally posted by @stevesloka in https://github.com/projectcontour/contour/pull/3934#discussion_r694048372_

stevesloka:
Rename to `envoy-admin` which better represents the backend cluster.

Fixes #3959

### DIFF
--- a/cmd/contour/shutdownmanager_test.go
+++ b/cmd/contour/shutdownmanager_test.go
@@ -193,14 +193,14 @@ func TestParseOpenConnections(t *testing.T) {
 
 // nolint:revive
 const (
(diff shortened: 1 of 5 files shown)


sunjayBhatia (approved):
doesn't look like the `service-stats` cluster is documented in the site explicitly so looks like you got all the mentions! (should we have something? maybe a separate issue)

will def need a release note etc. to make sure users are aware

youngnick (approved):

PR #4038, opened 2021-09-17
internal/dag: Add additional validation for fqdn

### LINKED ISSUE #3166: HTTPProxy FQDN Validation

stevesloka:
Adds additional validation for FQDN in HTTPProxy resources.

Fixes #3166

### DIFF
--- a/internal/dag/httpproxy_processor.go
+++ b/internal/dag/httpproxy_processor.go
@@ -15,10 +15,13 @@ package dag
 
 import (
 	"fmt"
+	"net"
 	"net/http"
 	"sort"
 	"strconv"
 	"strings"
+
+	"k8s.io/apimachinery/pkg/util/validation"
 
 	contour_api_v1 "github.com/projectcontour/contour/apis/projectcontour/v1"
 	contour_api_v1alpha1 "github.com/projectcontour/contour/apis/projectcontour/v1alpha1"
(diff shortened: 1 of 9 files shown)


internal/dag/httpproxy_processor.go
```diff
@@ -132,6 +135,17 @@ func (p *HTTPProxyProcessor) computeHTTPProxy(proxy *contour_api_v1.HTTPProxy) {
 			"Spec.VirtualHost.Fqdn must be specified")
 		return
 	}
+	if strings.Contains(host, "*") {
```
youngnick, line 138: Does the CRD validation allow the `*` character? If not, why check here?

youngnick (approved):
LGTM, with one question about if we need the extra validation once we have the CRD valdiation.

internal/dag/httpproxy_processor.go
stevesloka: I didn't add that one but the one below. I removed both assuming users update the CRDs properly. If not then these checks won't happen.

sunjayBhatia (approved):

PR #4041, opened 2021-09-17
makefile: Remove certgen targets

### LINKED ISSUE #3736: possibly remove manual certgen code
ref. https://github.com/projectcontour/contour/pull/3728#pullrequestreview-669498159

There are [Makefile targets](https://github.com/projectcontour/contour/blob/main/Makefile#L260-L312) for manually generating certs for Contour<->Envoy communication. These are unlikely to be used anymore and could be removed.

However, they're referenced as the generator for test data in https://github.com/projectcontour/contour/blob/main/cmd/contour/servecontext_test.go#L123-L125, so possibly worth keeping around for the purposes of being able to re-generate test data there.

Additionally, `certs/cert-contour.ext` and `certs/cert-envoy.ext` are referenced in https://github.com/projectcontour/contour/blob/main/site/content/docs/main/grpc-tls-howto.md. The references could be replaced with inline `cat <<EOF >>certs/cert-contour.ext ...` commands to produce the needed files as part of that guide (if we want to keep that guide).

stevesloka:
Removes unused certgen targets from the Makefile
which are no longer used.

Fixes #3736

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,6 @@ IMAGE := $(REGISTRY)/$(PROJECT)
 SRCDIRS := ./cmd ./internal ./apis
 LOCAL_BOOTSTRAP_CONFIG = localenvoyconfig.yaml
 SECURE_LOCAL_BOOTSTRAP_CONFIG = securelocalenvoyconfig.yaml
-PHONY = gencerts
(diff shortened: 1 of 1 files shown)


youngnick (approved):
LGTM, thanks for tidying this up.

sunjayBhatia (approved):

PR #4048, opened 2021-09-22
cmd/contour: Use ContourConfiguration for all config

stevesloka:
Convert the ServeContext to a ContourConfiguration and use
that for all config items which will allow for the ServeContext
to be deprecated and the optional ContourConfig CRD to be
specified.

### DIFF
--- a/.github/workflows/prbuild.yaml
+++ b/.github/workflows/prbuild.yaml
@@ -31,7 +31,7 @@ jobs:
         steps: ${{ toJson(steps) }}
         channel: '#contour-ci-notifications'
       if: ${{ failure() && github.ref == 'refs/heads/main' }}
-  e2e-contourconfiguration:
+  e2e:
     runs-on: ubuntu-latest
     # TODO uncomment the below once we're using the image
     # or its binary for testing.
@@ -44,7 +44,7 @@ jobs:
         # a Kubernetes version number.
         kubernetes_version: ["kubernetes:latest", "kubernetes:n-1", "kubernetes:n-2"]
         # run tests using the configuration crd as well as without
-        use_config_crd: [true, false]
+        config_type: ["ConfigmapConfiguration", "ContourConfiguration"]
         # include defines an additional variable (the specific node
         # image to use) for each kubernetes_version value.
         include:
@@ -54,6 +54,11 @@ jobs:
             node_image: "docker.io/kindest/node:v1.21.2@sha256:9d07ff05e4afefbba983fac311807b3c17a5f36e7061f6cb7e2ba756255b2be4"
           - kubernetes_version: "kubernetes:n-2"
             node_image: "docker.io/kindest/node:v1.20.7@sha256:cbeaf907fc78ac97ce7b625e4bf0de16e3ea725daf6b04f930bd14c67c671ff9"
+          - config_type: "ConfigmapConfiguration"
+            use_config_crd: "false"
+          - config_type: "ContourConfiguration"
+            use_config_crd: "false"
+
     steps:
       - uses: actions/checkout@v2
       # TODO uncomment the below once we're using the image
--- a/.github/workflows/prbuild.yaml
+++ b/.github/workflows/prbuild.yaml
@@ -31,7 +31,7 @@ jobs:
         steps: ${{ toJson(steps) }}
         channel: '#contour-ci-notifications'
       if: ${{ failure() && github.ref == 'refs/heads/main' }}
-  e2e:
+  e2e-configmap:
     runs-on: ubuntu-latest
     # TODO uncomment the below once we're using the image
     # or its binary for testing.
@@ -82,6 +82,66 @@ jobs:
         env:
           NODEIMAGE: ${{ matrix.node_image }}
           LOAD_PREBUILT_IMAGE: "true"
+        run: |
+          make e2e
+      - uses: act10ns/slack@v1
+        with:
+          status: ${{ job.status }}
+          steps: ${{ toJson(steps) }}
+          channel: '#contour-ci-notifications'
+        if: ${{ failure() && github.ref == 'refs/heads/main' }}
+  e2e-contourconfiguration:
+    runs-on: ubuntu-latest
+    # TODO uncomment the below once we're using the image
+    # or its binary for testing.
+    # needs: [build-image]
+    strategy:
+      matrix:
+        # use stable kubernetes_version values since they're included
+        # in the name of the GitHub Actions job, and we don't want to
+        # have to update branch protection rules every time we change
+        # a Kubernetes version number.
+        kubernetes_version: ["kubernetes:latest", "kubernetes:n-1", "kubernetes:n-2"]
+        # include defines an additional variable (the specific node
+        # image to use) for each kubernetes_version value.
+        include:
+          - kubernetes_version: "kubernetes:latest"
+            node_image: "docker.io/kindest/node:v1.22.0@sha256:f97edf7f7ed53c57762b24f90a34fad101386c5bd4d93baeb45449557148c717"
+          - kubernetes_version: "kubernetes:n-1"
+            node_image: "docker.io/kindest/node:v1.21.2@sha256:9d07ff05e4afefbba983fac311807b3c17a5f36e7061f6cb7e2ba756255b2be4"
+          - kubernetes_version: "kubernetes:n-2"
+            node_image: "docker.io/kindest/node:v1.20.7@sha256:cbeaf907fc78ac97ce7b625e4bf0de16e3ea725daf6b04f930bd14c67c671ff9"
+    steps:
+      - uses: actions/checkout@v2
+      # TODO uncomment the below once we're using the image
+      # or its binary for testing.
+      # - name: Download image
+      #   uses: actions/download-artifact@v2
+      #   with:
+      #     name: image
+      #     path: image
+      - uses: actions/cache@v2
+        with:
+          # * Module download cache
+          # * Build cache (Linux)
+          path: |
+            ~/go/pkg/mod
+            ~/.cache/go-build
+          key: ${{ runner.os }}-${{ github.job }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-${{ github.job }}-go-
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.17.0'
+      - name: add deps to path
+        run: |
+          ./hack/actions/install-kubernetes-toolchain.sh $GITHUB_WORKSPACE/bin
+          echo "$GITHUB_WORKSPACE/bin" >> $GITHUB_PATH
+      - name: e2e tests
+        env:
+          NODEIMAGE: ${{ matrix.node_image }}
+          LOAD_PREBUILT_IMAGE: "true"
+          USE_CONTOUR_CONFIGURATION_CRD: "true"
         run: |
           make e2e
       - uses: act10ns/slack@v1
--- a/.github/workflows/prbuild.yaml
+++ b/.github/workflows/prbuild.yaml
@@ -57,7 +57,7 @@ jobs:
           - config_type: "ConfigmapConfiguration"
             use_config_crd: "false"
           - config_type: "ContourConfiguration"
-            use_config_crd: "false"
+            use_config_crd: "true"
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/prbuild.yaml
+++ b/.github/workflows/prbuild.yaml
@@ -31,65 +31,6 @@ jobs:
         steps: ${{ toJson(steps) }}
         channel: '#contour-ci-notifications'
       if: ${{ failure() && github.ref == 'refs/heads/main' }}
-  e2e-configmap:
-    runs-on: ubuntu-latest
-    # TODO uncomment the below once we're using the image
-    # or its binary for testing.
-    # needs: [build-image]
-    strategy:
-      matrix:
-        # use stable kubernetes_version values since they're included
-        # in the name of the GitHub Actions job, and we don't want to
-        # have to update branch protection rules every time we change
-        # a Kubernetes version number.
-        kubernetes_version: ["kubernetes:latest", "kubernetes:n-1", "kubernetes:n-2"]
-        # include defines an additional variable (the specific node
-        # image to use) for each kubernetes_version value.
-        include:
-          - kubernetes_version: "kubernetes:latest"
-            node_image: "docker.io/kindest/node:v1.22.0@sha256:f97edf7f7ed53c57762b24f90a34fad101386c5bd4d93baeb45449557148c717"
-          - kubernetes_version: "kubernetes:n-1"
-            node_image: "docker.io/kindest/node:v1.21.2@sha256:9d07ff05e4afefbba983fac311807b3c17a5f36e7061f6cb7e2ba756255b2be4"
-          - kubernetes_version: "kubernetes:n-2"
-            node_image: "docker.io/kindest/node:v1.20.7@sha256:cbeaf907fc78ac97ce7b625e4bf0de16e3ea725daf6b04f930bd14c67c671ff9"
-    steps:
-      - uses: actions/checkout@v2
-      # TODO uncomment the below once we're using the image
-      # or its binary for testing.
-      # - name: Download image
-      #   uses: actions/download-artifact@v2
-      #   with:
-      #     name: image
-      #     path: image
-      - uses: actions/cache@v2
-        with:
-          # * Module download cache
-          # * Build cache (Linux)
-          path: |
-            ~/go/pkg/mod
-            ~/.cache/go-build
-          key: ${{ runner.os }}-${{ github.job }}-go-${{ hashFiles('**/go.sum') }}
-          restore-keys: |
-            ${{ runner.os }}-${{ github.job }}-go-
-      - uses: actions/setup-go@v2
-        with:
-          go-version: '1.17.0'
-      - name: add deps to path
-        run: |
-          ./hack/actions/install-kubernetes-toolchain.sh $GITHUB_WORKSPACE/bin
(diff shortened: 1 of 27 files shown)


stevesloka:
I think there's a bunch more cleanup that we can do to this, but didn't want to change too much around.

cmd/contour/serve.go
```diff
@@ -821,3 +879,265 @@ func informOnResource(clients *k8s.Clients, gvr schema.GroupVersionResource, han
 	inf.AddEventHandler(handler)
 	return nil
 }
+
+func convertServeContext(ctx *serveContext) contour_api_v1alpha1.ContourConfigurationSpec {
```
stevesloka, line 883: I'm going to add some tests to validate this convert method a bit.

skriss (changes requested):

cmd/contour/serve.go
```diff
@@ -713,36 +762,45 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 	return g.Run(context.Background())
 }
 
-func getDAGBuilder(ctx *serveContext, clients *k8s.Clients, clientCert, fallbackCert *types.NamespacedName, log logrus.FieldLogger) dag.Builder {
+func getDAGBuilder(ingressClassName string, rootNamespaces []string, gatewayAPIConfigured bool, disablePermitInsecure bool, enableExternalNameService bool,
```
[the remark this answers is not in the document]
stevesloka: yeah for sure, I can do that. I was trying the least bit of change at first, but agree.

cmd/contour/serve.go
```diff
@@ -296,79 +302,106 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+			if err != nil {
+				return fmt.Errorf("error parsing connection shutdown grace period: %w", err)
+			}
+		}
+		if contourConfiguration.Envoy.Timeouts.RequestTimeout != nil {
+			requestTimeout, err = timeout.Parse(*contourConfiguration.Envoy.Timeouts.RequestTimeout)
+			if err != nil {
+				return fmt.Errorf("error parsing request timeout: %w", err)
+			}
+		}
```
tsaarni, line 348: This is more general comment than just for this block alone: Would it make sense to try move code from `doServe()` since it starts to get really long?  
For example, if there was listener builder that takes `ContourConfiguration`, maybe the above code could be moved there from `doServe()`. Similarly maybe dag builder could take `ContourConfiguration` to offload parsing some parameters from `doServe()` to the builder.

stevesloka: Yes for sure! I was trying to convert and keep the existing logic in place for easier reviewing. I can take some new commits from here and clean some of this up if it doesn't get too difficult to review.

cmd/contour/leadership.go
```diff
@@ -97,12 +99,16 @@ func newLeaderElector(
 
 	rl := newResourceLock(log, conf, clients)
 
+	leaseDuration, _ := time.ParseDuration(conf.LeaseDuration) // TODO;
```
sunjayBhatia, line 102: missing error handling?

cmd/contour/serve.go
```diff
@@ -200,196 +240,125 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
-			if err != nil {
-				return fmt.Errorf("error parsing rate limit extension service %s response timeout: %v", namespacedName, err)
-			}
-		}
+	cipherSuites := []string{}
+	for _, cs := range contourConfiguration.Envoy.Listener.TLS.CipherSuites {
+		cipherSuites = append(cipherSuites, string(cs))
+	}
+
+	listenerConfig := xdscache_v3.NewListenerConfig(
```
youngnick, line 283: Note for later; this feels like a good candidate for just passing in the Config object or something.

cmd/contour/serve.go
```diff
@@ -557,87 +441,160 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
...
+			// Do nothing. The group is shutting down.
+		}
+		return nil
+	})
+
+	// GO!
+	return s.group.Run(context.Background())
+}
+
+func (s *Serve) setupRateLimitService(contourConfiguration contour_api_v1alpha1.ContourConfigurationSpec, listenerConfig *xdscache_v3.ListenerConfig) error {
```
youngnick, line 509: I like these `setup`-style functions, good way to pull this out of `doServe`.

cmd/contour/serve.go
```diff
@@ -778,3 +835,57 @@ func informOnResource(clients *k8s.Clients, gvr schema.GroupVersionResource, han
 	inf.AddEventHandler(handler)
 	return nil
 }
+
+// commandOperatorRegexp parses the command operators used in Envoy access log configuration
```
youngnick, line 839: Was this moved in from the public API? Seems like it might be a good candidate to extract out to its own thing later.

cmd/contour/servecontext.go
```diff
@@ -155,28 +161,31 @@ func (ctx *serveContext) grpcOptions(log logrus.FieldLogger) []grpc.ServerOption
 
 // tlsconfig returns a new *tls.Config. If the context is not properly configured
```
youngnick, line 162: Nit: Small doc update here since it's not passed a context any more.

internal/xdscache/v3/listener.go
```diff
@@ -156,6 +159,122 @@ type RateLimitConfig struct {
 	EnableXRateLimitHeaders bool
 }
 
+func NewListenerConfig(
```
youngnick, line 162: I commented this at the calling site, but seeing the function definition really drives home that this one really should take the config object and extract the fields itself. Not blocking for this one, but should be refactored at some point.

youngnick (commented):
This is looking pretty good to me, comments are minor doc update or style things that could be refactored later. I didn't want to approve yet until this is not draft though. Looking good Steve!

cmd/contour/serve.go
```diff
@@ -161,32 +163,70 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 	return serve, ctx
 }
 
-// doServe runs the contour serve subcommand.
-func doServe(log logrus.FieldLogger, ctx *serveContext) error {
+type Serve struct {
```
skriss, line 166: I'd opt for `Server` over `Serve` here for the struct name.

cmd/contour/serve.go
```diff
@@ -161,32 +163,70 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
...
+		mgr:      mgr,
+		registry: registry,
+	}, nil
+}
+
+// doServe runs the contour serve subcommand.
+func (s *Serve) doServe() error {
+
+	// Convert the Contour ServeContext into a ContourConfiguration
+	contourConfiguration := s.ctx.convertToContourConfigurationSpec()
```
skriss, line 218: Only need to do this if `s.ctx.contourConfigurationName` is empty, right?

skriss: Related, on L237 we're declaring a new `contourConfiguration` var inside the if statement and populating it with the CRD contents, but we should be populating the var declared here on L218.

stevesloka: the initial point of this was to only covert the servecontext to the crd spec. The next one was to then actually implement and use it, but can do that all here in this PR if you want.

stevesloka: I just updated to use it the config if passed in, just another line of code.

skriss: Ah, well if it's only one more line to use the CRD when present may as well :)

cmd/contour/serve.go
```diff
@@ -161,235 +163,202 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
...
+type Server struct {
+	group    workgroup.Group
+	log      logrus.FieldLogger
+	ctx      *serveContext
+	clients  *k8s.Clients
+	mgr      manager.Manager
+	registry *prometheus.Registry
+}
+
+func NewServe(log logrus.FieldLogger, ctx *serveContext) (*Server, error) {
```
sunjayBhatia, line 175: tiny nit: `NewServer` 😅

cmd/contour/serve.go
```diff
@@ -161,235 +163,202 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
...
+
+	// Set up workgroup runner.
+	var group workgroup.Group
+
 	// Establish k8s core & dynamic client connections.
 	clients, err := k8s.NewClients(ctx.Config.Kubeconfig, ctx.Config.InCluster)
 	if err != nil {
-		return fmt.Errorf("failed to create Kubernetes clients: %w", err)
+		return nil, fmt.Errorf("failed to create Kubernetes clients: %w", err)
```
sunjayBhatia, line 183: nit on what `NewServe` returns, seems like we have a mix of `log.Fatal` and returning an error, should we standardize on one?

cmd/contour/serve.go
```diff
@@ -161,235 +163,202 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 	return serve, ctx
 }
 
-// doServe runs the contour serve subcommand.
-func doServe(log logrus.FieldLogger, ctx *serveContext) error {
+type Server struct {
```
sunjayBhatia, line 166: +1 for pulling this out into an object so we can refactor into its own package potentially at some point for test coverage etc.

stevesloka:
I still need to address the review comments, but I also now want to write some e2e tests to cover the config crd since now that's added to this PR.

cmd/contour/serve.go
```diff
@@ -547,15 +594,15 @@ func doServe(log logrus.FieldLogger, ctx *serveContext) error {
 
 	// Create metrics service and register with workgroup.
 	metricsvc := httpsvc.Service{
-		Addr:        ctx.metricsAddr,
-		Port:        ctx.metricsPort,
+		Addr:        contourConfiguration.Metrics.Address,
```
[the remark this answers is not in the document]
stevesloka: yup! You're right, I'll update.

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -535,6 +523,21 @@ type NamespacedName struct {
 	Namespace string `json:"namespace"`
 }
 
+func (n *NamespacedName) NamespacedNameOf() *types.NamespacedName {
```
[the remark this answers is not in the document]
stevesloka: The upstream structs don't have the proper yaml or json tags for parsing.

cmd/contour/serve.go
stevesloka: Yeah this is from parameters.go. This needs to live somewhere better than in serve.go, but needed a way to validate the accesslog fields.

internal/xdscache/v3/listener.go
stevesloka: yeah I agree the params aren't great, but I don't think passing the full ContourConfig.Spec is good either. A future place to refactor for sure. =)

cmd/contour/serve.go
stevesloka: good catch! Copy pasted that all together. Updated to make it consistent.

.github/workflows/prbuild.yaml
```diff
@@ -90,6 +90,66 @@ jobs:
           steps: ${{ toJson(steps) }}
           channel: '#contour-ci-notifications'
         if: ${{ failure() && github.ref == 'refs/heads/main' }}
+  e2e-contourconfiguration:
```
skriss, line 93: You should be able to edit the matrix in the existing `e2e` job to have it run both with and without the CRD -- see e.g. https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-running-with-multiple-operating-systems (ignore the operating system bit, but it shows having two different dimensions for the matrix). Will avoid having to duplicate the entire job config. So something like:

```
matrix:
    kubernetes_version: ["kubernetes:latest", "kubernetes:n-1", "kubernetes:n-2"]
    use_config_crd: [true, false]
...
```

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -518,6 +503,9 @@ type PolicyConfig struct {
 	// ResponseHeadersPolicy defines the response headers set/removed on all routes
 	// +optional
 	ResponseHeadersPolicy *HeadersPolicy `json:"responseHeaders,omitempty"`
+
+	// +optional
+	ApplyToIngress bool `json:"applyToIngress"`
```
skriss, line 508: needs some godoc

test/e2e/deployment.go
```diff
@@ -555,40 +559,117 @@ func (d *Deployment) DeleteResourcesForLocalContour() error {
...
+func (d *Deployment) StartLocalContour(config *config.Parameters, contourConfiguration *contour_api_v1alpha1.ContourConfiguration, additionalArgs ...string) (*gexec.Session, string, error) {
+
+	var content []byte
+	var configReferenceName string
+	var contourServeArgs []string
+	var err error
+
+	// Look for the ENV variable to tell if this test run should use
+	// the ContourConfiguration file or the ContourConfiguration CRD.
+	if _, useContourConfiguration := os.LookupEnv("USE_CONTOUR_CONFIGURATION_CRD"); useContourConfiguration {
```
skriss, line 571: If you change the GH Actions matrix, probably will end up being easier to check for a value of `true` here rather than just existence of the var.

test/e2e/deployment.go
```diff
@@ -555,40 +559,117 @@ func (d *Deployment) DeleteResourcesForLocalContour() error {
...
-		"serve",
-		"--xds-address=0.0.0.0",
-		"--xds-port=" + d.localContourPort,
-		"--insecure",
-		"--kubeconfig=" + d.kubeConfig,
-		"--config-path=" + configFile.Name(),
-		"--disable-leader-election",
-	}, additionalArgs...)
+	// Set the CONTOUR_NAMESPACE env variable so Contour can lookup the Configuration CRD in the proper namespace.
+	//os.Setenv("CONTOUR_NAMESPACE", "projectcontour")
```
skriss, line 632: not needed?

test/e2e/fixtures.go
```diff
@@ -303,3 +305,76 @@ func (e *EchoSecure) Deploy(ns, name string) func() {
 		require.NoError(e.t, e.client.Delete(context.TODO(), deployment))
 	}
 }
+
+var ContourConfigration = &contour_api_v1alpha1.ContourConfiguration{
```
skriss, line 309:
```suggestion
var ContourConfiguration = &contour_api_v1alpha1.ContourConfiguration{
```

test/e2e/deployment.go
stevesloka: I ended up adding this as a default in serve.go, so no not needed.

test/e2e/deployment.go
stevesloka: yup good point

apis/projectcontour/v1alpha1/contourconfig.go
stevesloka: 👍

.github/workflows/prbuild.yaml
```diff
@@ -52,6 +54,11 @@ jobs:
             node_image: "docker.io/kindest/node:v1.21.2@sha256:9d07ff05e4afefbba983fac311807b3c17a5f36e7061f6cb7e2ba756255b2be4"
           - kubernetes_version: "kubernetes:n-2"
             node_image: "docker.io/kindest/node:v1.20.7@sha256:cbeaf907fc78ac97ce7b625e4bf0de16e3ea725daf6b04f930bd14c67c671ff9"
+          - config_type: "ConfigmapConfiguration"
+            use_config_crd: "false"
+          - config_type: "ContourConfiguration"
+            use_config_crd: "false"
```
sunjayBhatia, line 60:
```suggestion
            use_config_crd: "true"
```

stevesloka: 🤭

test/e2e/fixtures.go
```diff
@@ -303,3 +305,76 @@ func (e *EchoSecure) Deploy(ns, name string) func() {
 		require.NoError(e.t, e.client.Delete(context.TODO(), deployment))
 	}
 }
+
+var ContourConfiguration = &contour_api_v1alpha1.ContourConfiguration{
```
skriss, line 309: I'd prefer this to be a function that returns a ContourConfig, can help avoid issues where it's inadvertently modified by one test and pollutes other tests. I know you have DeepCopy's where it's used now, just makes it a little less error-prone going fwd.

test/e2e/deployment.go
```diff
@@ -555,40 +559,115 @@ func (d *Deployment) DeleteResourcesForLocalContour() error {
...
 	}
-	return session, configFile.Name(), nil
+	return session, configReferenceName, nil
 }
 
 func (d *Deployment) StopLocalContour(contourCmd *gexec.Session, configFile string) error {
+
+	// Look for the ENV variable to tell if this test run should use
+	// the ContourConfiguration file or the ContourConfiguration CRD.
+	if _, useContourConfiguration := os.LookupEnv("USE_CONTOUR_CONFIGURATION_CRD"); useContourConfiguration {
```
skriss, line 642: Also need to check for "true" here

test/e2e/fixtures.go
skriss: Don't need to block on this, can do it later too.

stevesloka: Yup np.

test/e2e/deployment.go
stevesloka: 👍

youngnick (approved):
This is looking good to me. Space for future work, but this seems good enough to get us going for 1.19 to me.

skriss (approved):

Issue #3584, opened 2021-04-16
Update CI base images for E2E tests


stevesloka:
Our current kind based testing infra pulls a bunch of images each time the e2e tests are run. In some cases the images are very large (500mb+). To speed up the time to pull/load into the cluster, we could pre-load these images into a custom kind image that allows to be the base image. Additionally we could pick smaller test images that aren't so large to help reduce build/test time. 

- [ ] Reduce image sizes for tests
- [ ] CI job/script to build base image & push to a registry

skriss:
It could also be worth considering maintaining our own test "echo server" binary that has the functionality we need for testing purposes. Right now we depend on `ingress-conformance-echo` (a custom build of it from `tsaarni`) and `httpbin` for the combo of functionality we need.

youngnick:
I agree that we should either pick one upstream thing to use for all our tests, or make our own instead and do that. But it should be the one thing across everything regardless.

tsaarni:
I think the "official" build of `ingress-conformance-echo` is at https://gcr.io/k8s-staging-ingressconformance/echoserver, latest still being `gcr.io/k8s-staging-ingressconformance/echoserver:v20201006-42d00bd`.  This one includes the TLS stuff, which was the reason for my custom build back then.

In my own personal dev tests I've completely gotten rid of `httpbin` and use `ingress-conformance-echo` exclusively.  It is much lighter and still contains everything I need.

skriss:
> In my own personal dev tests I've completely gotten rid of `httpbin` and use `ingress-conformance-echo` exclusively. It is much lighter and still contains everything I need.

Yep, we use `httpbin` only in one test -- the health checks test -- and only because it has functionality where if you request a path like `/status/418` it will return a `418` response code. It would be really nice to not have to use it anymore since it's so big.

tsaarni:
> Yep, we use `httpbin` only in one test -- the health checks test -- and only because it has functionality where if you request a path like `/status/418` it will return a `418` response code. It would be really nice to not have to use it anymore since it's so big.

I guess the conformance test suite does not go into status codes of upstream services currently but I wonder if it could be justified that `ingress-conformance-echo` would have a `/status/` endpoint for testing controllers, something simple like [this](https://github.com/tsaarni/ingress-controller-conformance/commit/adb4adecf6b6c70fb6cbb30177a6e6db74e04c38).

skriss:
[...]

PR #4051, opened 2021-09-23
test/e2e: replace httpbin with ingress echoserver

skriss:
Removes usages of httpbin and replaces it with
the ingress echoserver now that the latter supports
returning specified status codes.

Updates #3584.


stevesloka (approved):
w00t! Works on my M1 mac now! 🎉

tsaarni (approved):
Drop-in replacement, nice :-)

skriss:
Credit goes to @tsaarni here for getting us the feature we needed in the ingress-conformance-echo image 💯

Issue #3142, opened 2020-11-20
Contour "X-Contour-Session-Affinity" cookie is missing "secure" and "SameSite" settings


ryanelian:
**What steps did you take and what happened:**

Enable Contour sticky session / affinity for a website:

```yaml
spec:
  routes:
    loadBalancerPolicy:
      strategy: Cookie
```

Then open the said website in Firefox / scan the website using some security audit software.

Firefox will warn you that they will soon **reject** the cookie:

```
Cookie “X-Contour-Session-Affinity” will be soon rejected because it has the “SameSite” attribute set to “None” or an invalid value, without the “secure” attribute. To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite
```

A security audit gives us this message:

```
The 'secure' flag is not set on this cookie. There is no Cookie Prefix on this cookie. This is not a SameSite Cookie. 
```

This is a somewhat important problem because it may cause developers using Contour to fail security audit with no way of fixing it other than changing their ingress / reverse proxy.

**What did you expect to happen:**

Cookie is not flagged as insecure: set `SameSite` flag to `Lax` or `Strict`

Set cookie as secure if accessing the site via HTTPS / TLS enabled.

**Anything else you would like to add:**

Firefox version: 84.0b2
Sample security header scanner: https://securityheaders.com

**Environment:**

- Contour version: 1.8.1
- Kubernetes version: (use `kubectl version`): v1.18.9
- Kubernetes installer & version: kubeadm
- Cloud provider or hardware configuration: UpCloud 2core 4GB RAM
- OS (e.g. from `/etc/os-release`): Ubuntu 18.04.5 LTS

youngnick:
Thanks for this issue @ryanelian. We've got a number of issues around the Contour cookie generation open, #2965 seems very similar to yours. Thanks for the additional information from Firefox though, that's very helpful. I'll speak to the team and get this prioritised.

youngnick:
Just looking over this again and wanted to ask what you would expect to see, @ryanelian?

It seems like the `Secure` property should be set if accessing the site via HTTPS, but I think one of us will need to spend some time parsing how `SameSite` and `Secure` interact between themselves and with CORS?

youngnick:
We've also got #3044 open as a request to add a more direct ability to configure the Envoy `hash_policy`, which seems like it would allow this to be solved, but it seems to me like it might be overkill.

[...]

Issue #2965, opened 2020-10-01
httpproxy cookie attributes


lienerieksta:
It would be very helpful if there was option to set additional attributes for httpproxy session affinity cookie. 

We have a stateful application that requires the cookie to have "SameSite=None" and "Secure" attributes set for them to be returned by the client.

I believe this feature would complement the very welcome addition of CORS support in contour.

youngnick:
Thanks for this issue @lienerieksta, sorry about the delay in getting back to you. We have another request open (#3142) that's similar, and I wanted to ask what you would like for this one.

Would you want `SameSite=None` and `Secure` unconditionally added when using the `loadBalancerPolicy: cookie` setting? Or would you prefer more configurability?

lienerieksta:
Thank you for you reply!

I think it needs to be more configurable. Adding `Secure` unconditionally would break the option to use `loadBalancerPolicy: cookie` without HTTPS. 

Most browsers nowadays seem to gravitate to `SameSite=Lax` as default behaviour and therefore that seems the most reasonable default, if one needs to be set. 

The `Secure` attribute could be added transparently when `SameSite=None` is set, or just as another configuration field independent of SameSite. 

From configuration perspective I imagined it something along the lines of:
```yaml
spec:
  routes:
    loadBalancerPolicy:
      strategy: Cookie
      attributes:
        sameSite: Lax/None/Strict
        secure: True/False
```

sunjayBhatia:
This is something that would require an Envoy change no? Contour currently doesn't have control over the cookie attributes as per: https://www.envoyproxy.io/docs/envoy/v1.17.0/api-v3/config/route/v3/route_components.proto#config-route-v3-routeaction-hashpolicy-cookie

youngnick:
Ah, you're right @sunjayBhatia. If we implement #3044, can you use the hash_policy list to get these attributes?

Regardless, it's probably worth opening an upstream issue to see if we can get these settings configurable.

sunjayBhatia:
> Ah, you're right @sunjayBhatia. If we implement #3044, can you use the hash_policy list to get these attributes?
> 
> Regardless, it's probably worth opening an upstream issue to see if we can get these settings configurable.

Currently Envoy doesn't expose them but if it did and we do #3044 we would get the opportunity to configure them per named cookie

[...]

PR #4056, opened 2021-09-27
Cookie manipulation with Lua per Route/Cluster

sunjayBhatia:
See design: https://github.com/projectcontour/contour/pull/4023

Includes:
- change to API, Cookie rewrite per Route and Service on HTTPProxy
- can rewrite Path, Domain, Secure, SameSite
- kubebuilder validation ensures invalid Name, Path, Domain, etc. cannot
have invalid characters
- rewrite rules de-duplicated, cannot have multiple on Route or Service
rewriting the same cookie
- Lua placeholder filter added to each HTTP Connection Manager so we can
use the per-Route/per-Cluster overrides
- Lua generated per Envoy Route and Cluster
- We find Set-Cookie headers in envoy_on_response and look through the
cookies to match them by name to the rewrite rules
- Parse existing attributes into Lua table and rewrite with rules from
user configuration
- e2e tests ensure kubebuilder validations are correct, cookies can be
rewritten when returned from app, when returned from manipulated
headers, cookie load balancing, rewrite rules on route, rewrite rules on
service, rewrite rules for TLS vhosts
- e2e tests use a different echo server that allows us to customize
returned headers to send back a Set-Cookie header

Fixes: #3142
Fixes: #2965


internal/envoy/v3/route.go
```diff
@@ -402,3 +411,123 @@ func containsMatch(s string) *envoy_route_v3.HeaderMatcher_SafeRegexMatch {
...
+			response_handle:headers():add("set-cookie", v)
+		end
+	end
+end
+	`
+
+	t := new(bytes.Buffer)
+	if err := template.Must(template.New("code").Parse(codeTemplate)).Execute(t, policies); err != nil {
+		// If template execution fails, return empty filter.
+		return nil
```
sunjayBhatia, line 520: don't super love this resolution, not sure how to test the effects tbh

`policies` should already be validated so not sure how this could fail

test/e2e/httpproxy/cookie_rewrite_test.go
```diff
@@ -0,0 +1,827 @@
...
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{"app.kubernetes.io/name": name},
+				},
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{
+						{
+							Name:  "echo",
+							Image: "docker.io/ealen/echo-server:0.5.1",
```
sunjayBhatia, line 757: will need to use this alternative echo server, or get custom features into the ingress-conformance one

might want to push this into `fixtures.go` so it gets loaded into the test cluster before tests run

sunjayBhatia:
For reviewers: Something I was stuck on for a while was trying to distribute the Lua scripts differently. Currently this change is implemented such that we add an empty global Lua placeholder that can be overridden by per-Route and per-Cluster Lua scripts. The per-Route and per-Cluster scripts do include a lot of the same content in the `rewrite_cookie` function and some of the `Set-Cookie` header finding and rewriting. Ideally I would want to load that part from a common Lua script but that doesn't seem to be possible unless we package that in the Envoy container filesystem or some other manner. My next try is to use per-Route/Cluster metadata and have a common cookie rewriting script that pulls that in.

internal/envoy/v3/route.go
```diff
@@ -402,3 +411,123 @@ func containsMatch(s string) *envoy_route_v3.HeaderMatcher_SafeRegexMatch {
...
+			local attr_end = string.find(original, ";", iter)
+			local new_iter = attr_end
+			if attr_end == nil then
+				-- Set to 0 since we have to subtract below if we did find a ;
+				attr_end = 0
+				new_iter = original_len
+			end
+			local attr_value = string.sub(original, iter + 1, attr_end - 1)
+			-- Strip whitespace from front
+			attr_value = string.gsub(attr_value, "^%s*(.-)$", "%1")
```
tsaarni, line 474: Looking at [RFC6265](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1), it seems to me that attribute values must be always separated exactly by single space

```
set-cookie-string = cookie-pair *( ";" SP cookie-av )
```

but maybe there are misbehaving servers that add multiple spaces or none and stripping whitespace fixes that?

internal/dag/policy.go
```diff
@@ -304,6 +305,53 @@ func escapeHeaderValue(value string, dynamicHeaders map[string]string) string {
...
+			domain = pointer.String(p.DomainRewrite.Value)
+		}
+		secure := uint(0)
+		if p.Secure != nil {
+			policiesSet++
+			secure++
+			if *p.Secure {
+				secure++
+			}
+		}
```
tsaarni, line 334: Integer `secure` "security level" might be bit difficult to understand without having a comment close by. It was well explained in `dag.go` 
```
// Using an uint since pointer to boolean gets dereferenced in golang
// text templates so we have no way of distinguishing if unset or set to false.
// 0 means unset, 1 means false, 2 means true
```

internal/envoy/v3/route.go
```diff
@@ -402,3 +411,123 @@ func containsMatch(s string) *envoy_route_v3.HeaderMatcher_SafeRegexMatch {
...
+		{{if $p.Path}}attributes["Path"] = "Path={{$p.Path}}"{{end}}
+		{{if $p.Domain}}attributes["Domain"] = "Domain={{$p.Domain}}"{{end}}
+		{{if $p.SameSite}}attributes["SameSite"] = "SameSite={{$p.SameSite}}"{{end}}
+		{{if eq $p.Secure 1}}attributes["Secure"] = nil{{end}}
+		{{if eq $p.Secure 2}}attributes["Secure"] = "Secure"{{end}}
+	end
+	rewrite_table["{{$p.Name}}"] = cookie_{{$i}}_attribute_rewrite
+	{{end}}
+
+	function rewrite_cookie(original)
```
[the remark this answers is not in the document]
sunjayBhatia: Yeah that approach was what I wanted to do but the ability to use a script/module from the global filters wasn’t working for me

this will be refactored though, putting the rewrite data into per route filter data and using a common script to extract that and do rewrites

internal/dag/policy.go
sunjayBhatia: Might be able to get rid of this since I’m refactoring a bit, but will comment more here if it stays

internal/envoy/v3/route.go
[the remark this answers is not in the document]
tsaarni: I see you already mentioned this in the reviewer notes. Sorry for going through this too hastily!
(I recently started reviewing in vscode "github pull requests and issues" extension, instead of github.com and I'm bit disoriented where to find existing comments :-))

sunjayBhatia:
> > For reviewers: Something I was stuck on for a while was trying to distribute the Lua scripts differently. Currently this change is implemented such that we add an empty global Lua placeholder that can be overridden by per-Route and per-Cluster Lua scripts. The per-Route and per-Cluster scripts do include a lot of the same content in the `rewrite_cookie` function and some of the `Set-Cookie` header finding and rewriting. Ideally I would want to load that part from a common Lua script but that doesn't seem to be possible unless we package that in the Envoy container filesystem or some other manner. My next try is to use per-Route/Cluster metadata and have a common cookie rewriting script that pulls that in.
> 
> using that now, so far so good

might not work actually, works fine on per-route rewrites but for per route+per-service rewrites there isn't a place to put filter metadata

you can add per-cluster metadata but that is for the cluster as a whole and devising a system by which we add metadata for all rewrites to a cluster and then differentiate based on the route match seems like a pain, since we have to have a 2 level matching

sunjayBhatia:
> might not work actually, works fine on per-route rewrites but for per route+per-service rewrites there isn't a place to put filter metadata
> 
> you can add per-cluster metadata but that is for the cluster as a whole and devising a system by which we add metadata for all rewrites to a cluster and then differentiate based on the route match seems like a pain, since we have to have a 2 level matching

so something that would work is the following:
- global cookie rewriting Lua script/Lua filter in each http filter chain
- when building a route, add metadata for each cookie rewrite rule
- also add metadata for each service/cluster that has a cookie rewrite rule
- would look something like:
```
          "match": {
           "prefix": "/rewrite-some"
          },
          "route": {
           "cluster": "app-cookie-rewrite/echo/80/da39a3ee5e"
          },
          "metadata": {
           "filter_metadata": {
            "envoy.filters.http.lua": {
             "projectcontour.cookie_rewrite.service_policies": {
              "app-cookie-rewrite/echo/80/da39a3ee5e": {
               ...
               }
             },
             "projectcontour.cookie_rewrite.route_policies": {
              "rewrite-some": {
               "Secure": 0,
               "Domain": "rs.com"
              }
             }
            }
           }
          },
          "typed_per_filter_config": {
           "envoy.filters.http.lua": {
            "@type": "type.googleapis.com/envoy.extensions.filters.http.lua.v3.LuaPerRoute",
            "name": "cookierewrite.lua"
           }
          }
         },
```
- each route would reference cookie rewrite script as needed
- when building a cluster with cookie rewrite, we use Lua to add dynamic metadata on the request, to tell us what upstream cluster this is for
- e.g.
```
function envoy_on_request(request_handle)
	request_handle:streamInfo():dynamicMetadata():set("envoy.filters.http.lua", "upstream.cluster", "%s")
end
```
- global cookie rewrite script would grab route cookie rewrite rules and cluster cookie rewrite rules if present and merge

I have most of this working in a spike except the somewhat tedious bits of merging route-level and cluster-level rules.

before I do it and polish it, in the spirit of time-boxing this refactor, what do yall think @projectcontour/maintainers ? If we eventually want to move to a native implementation, this might be unnecessary, but currently we could have a lot of repeated copies of a v slightly different Lua script being sent over the wire, which is unfortunate

internal/envoy/v3/route.go
```diff
@@ -402,3 +411,123 @@ func containsMatch(s string) *envoy_route_v3.HeaderMatcher_SafeRegexMatch {
 		SafeRegexMatch: SafeRegexMatch(regex),
 	}
 }
+
+func CookieRewriteConfig(policies []dag.CookieRewritePolicy) *any.Any {
+	codeTemplate := `
```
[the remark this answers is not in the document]
sunjayBhatia: yeah thats a nice idea, I think we could do some refactors to enable unit testing and linting and embed a lua file rather than inline string

might take that as a future action item if we keep this Lua long term (and i don't get it implemented in Envoy natively) and so we could update the other place we have a Lua script

internal/envoy/v3/route.go
sunjayBhatia: yeah, might make sense to be stricter and reject rewriting illegally structured set-cookie headers or assume valid content and act accordingly (though the latter could be odd/cause issues), this was an attempt to be a little lenient which might be wrong haha, might need more tests to see if we can break what I've written in practice as well

youngnick:
About to start a detailed review, but wrt your last comment @sunjayBhatia, the global script sounds fine to me if you have it mostly working already.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -345,11 +345,63 @@ type Route struct {
...
 	// The policy for rate limiting on the route.
 	// +optional
 	RateLimitPolicy *RateLimitPolicy `json:"rateLimitPolicy,omitempty"`
 }
 
+type CookieRewritePolicy struct {
+	// Name is the name of the cookie for which attributes will be rewritten.
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=4096
+	// +kubebuilder:validation:Pattern=`^[^()<>@,;:\\"\/[\]?={} \t\x7f\x00\x01\x02\x03\x04\x05\x06\x07\x08\x09\x0a\x0b\x0c\x0d\x0e\x0f\x10\x11\x12\x13\x14\x15\x16\x17\x18\x19\x1a\x1b\x1c\x1d\x1e\x1f]+$`
```
youngnick, line 360: Nit/small question: Why do an exclusive regex here and not only include the characters we want? Non-blocking.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -567,6 +619,9 @@ type Service struct {
 	// Rewriting the 'Host' header is not supported.
 	// +optional
 	ResponseHeadersPolicy *HeadersPolicy `json:"responseHeadersPolicy,omitempty"`
+	// The policies for rewriting Set-Cookie header attributes.
+	// +optional
+	CookieRewritePolicies []CookieRewritePolicy `json:"cookieRewritePolicies,omitempty"`
```
youngnick, line 624: Will Envoy end up processing these in the order specified? A small comment here to ensure that's clear would be useful.

internal/dag/policy.go
youngnick: +1 on comment if it stays.

internal/envoy/v3/route.go
youngnick: Yeah, I don't think this is easily testable without a big refactor.

youngnick (approved):
This all looks great to me, with a few small non-blocking questions.

internal/envoy/v3/listener.go
```diff
@@ -277,6 +277,14 @@ func (b *httpConnectionManagerBuilder) DefaultFilters() *httpConnectionManagerBu
...
 			},
 		},
+		&http.HttpFilter{
+			Name: "envoy.filters.http.lua",
+			ConfigType: &http.HttpFilter_TypedConfig{
+				TypedConfig: protobuf.MustMarshalAny(&lua.Lua{
+					InlineCode: "-- Placeholder for per-Route or per-Cluster overrides.",
+				}),
+			},
+		},
```
skriss, line 287: Are there any issues/conflicts with this + the lua filter added by `FilterMisdirectedRequests`?

sunjayBhatia: i *think* this test should prove it doesnt: https://github.com/projectcontour/contour/blob/d344821cf1e02eb1414432c2a81afd95348df3ce/test/e2e/httpproxy/cookie_rewrite_test.go#L632-L640

apis/projectcontour/v1/httpproxy.go
sunjayBhatia: the inclusive set would be quite a bit longer, of 127 ASCII characters we need to exclude ~50

sunjayBhatia:
ok, with respect to minimizing Lua proliferation, was not able to get the full thing I wanted, there are complications on setting dynamic metadata and using Lua scripts at different levels of config (per-Route vs. global vs. per-routeAction)

apis/projectcontour/v1/httpproxy.go
sunjayBhatia: multiple policies on the same cookie are not allowed, so i think order should not matter?

apis/projectcontour/v1/httpproxy.go
```diff
@@ -345,11 +345,63 @@ type Route struct {
...
+	// Secure enables rewriting the Set-Cookie Secure element.
+	// If not set, Secure attribute will not be rewritten.
+	// +optional
+	Secure *bool `json:"secure,omitempty"`
+
+	// SameSite enables rewriting the Set-Cookie SameSite element.
+	// If not set, SameSite attribute will not be rewritten.
+	// +optional
+	// +kubebuilder:validation:Enum=Strict;Lax;None
+	SameSite *string `json:"sameSite,omitempty"`
```
sunjayBhatia, line 382: are we ok with this being a string pointer and not a custom type? I dont remember if we have standardized on a custom type for enums or not

apis/projectcontour/v1/httpproxy.go
```diff
@@ -345,11 +345,63 @@ type Route struct {
...
+
+	// DomainRewrite enables rewriting the Set-Cookie Domain element.
+	// If not set, Domain will not be rewritten.
+	// +optional
+	DomainRewrite *CookieDomainRewrite `json:"domainRewrite,omitempty"`
+
+	// Secure enables rewriting the Set-Cookie Secure element.
+	// If not set, Secure attribute will not be rewritten.
+	// +optional
+	Secure *bool `json:"secure,omitempty"`
```
sunjayBhatia, line 376: ok with this being a bool pointer and not a custom type?

youngnick:
I approved earlier, and that hasn't changed. I only had one tiny doc nit, but it is non-blocking.

sunjayBhatia:
unrelated osx flake

PR #4061, opened 2021-09-29
internal/dag: don't walk DAG to look for service

### LINKED ISSUE #4058: Performance Regression v1.8.1 => v1.18.1
### Problem

A Knative user [recently reported](https://knative.slack.com/archives/C93E33SN8/p1632319040135700) that when their cluster had around 1000 Knative Services (~4k HTTPProxy) deploying new Services would take O(minutes).  They mentioned earlier versions did not see such delays.

### Context

First Knative has a two contour instances - one for external ingress and the other only for cluster local traffic. Readiness is defined by our controllers probing the envoy proxies for all the desired FQDN and waiting to receive a 200 OK.

We create 4 HTTPProxies for every Knative Service - ie. https://gist.github.com/dprotaso/13561f973e44d119d66326d2cdc68173

This is because we add 4 FQDN 
- (external) foo-test.default.example.com
- (internal) foo-test.default
- (internal) foo-test.default.svc
- (internal) foo-test.default.svc.cluster.local

### Test Setup

I wanted to compare the performance of Contour v1.8.1 against v1.18.1 to rule out any regressions there

#### Cluster Setup
```sh
#!/bin/bash -e

function create_cluster() {
  local CLUSTER_NAME=dave-knative
  local CLUSTER_ZONE=us-east1-d
  local CLUSTER_VERSION=${1:-latest}
  local CLUSTER_NODES=10

  shift 2>&1 > /dev/null

  gcloud container clusters create $CLUSTER_NAME \
    --cluster-version $CLUSTER_VERSION \
    --enable-autoscaling \
    --machine-type=e2-standard-4 \
    --addons=NodeLocalDNS \
    --min-nodes=$CLUSTER_NODES \
    --max-nodes=$CLUSTER_NODES \
    --num-nodes=$CLUSTER_NODES \
    --zone=$CLUSTER_ZONE \
    --scopes=cloud-platform \
    "$@"

  kubectl create clusterrolebinding cluster-admin-binding \
    --clusterrole=cluster-admin \
    --user=$(gcloud config get-value core/account)
}

function install_serving() {
  kubectl apply -f https://github.com/knative/serving/releases/download/v0.26.0/serving-crds.yaml
  kubectl apply -f https://github.com/knative/serving/releases/download/v0.26.0/serving-core.yaml
  kubectl apply -f https://github.com/knative-sandbox/net-contour/releases/download/v0.26.0/contour.yaml
  kubectl apply -f https://github.com/knative-sandbox/net-contour/releases/download/v0.26.0/net-contour.yaml
  kubectl patch configmap/config-network \
    --namespace knative-serving \
    --type merge \
    --patch '{"data":{"ingress.class":"contour.ingress.networking.knative.dev"}}'
}

create_cluster
install_serving
```

Note: To test Contour v1.8.1 the above script was altered to install the following contour version

```
kubectl apply -f https://github.com/knative-sandbox/net-contour/releases/download/v0.18.2/contour.yaml
```

Additional note: When testing on GKE bump the resource quota so that you can create about `3000` K8s Services - default quota is 1500 which will stall the tests

#### Running the tests

We just simply create 1000 Knative services serially

```bash
#!/bin/bash -e

for i in $(seq 1 1000); do
  kn service create foo-$i --image=gcr.io/knative-samples/autoscale-go:0.1 || true
done
```

### Gathering the data

Here we capture a CSV output which includes the Service's creation time and the `Ready` Time.

```
kubectl get ksvc -ojsonpath='{range .items[*]}{.metadata.name},{.metadata.creationTimestamp},{.status.conditions[?(@.type=="Ready")].lastTransitionTime}{"\n"}{end}' | pbcopy
```

### Comparing the results

I plotted the 'Time to Ready' to visualize the performance regression

#### Contour v1.8.1

Screen Shot 2021-09-28 at 9 59 19 AM

#### Contour v1.18.1

Screen Shot 2021-09-28 at 9 59 14 AM

**Environment:**

- Kubernetes version: (use `kubectl version`): v1.20.10-gke.1600

skriss:
Removes code that walks the DAG looking for
a matching existing Service when adding a
new Service. This code is very inefficient
when the DAG is large and creates a
performance bottleneck. Some extra memory
may end up being used if duplicate Services
are created, but it's unlikely to be a
material issue in practice.

Closes #4058.

### DIFF
--- a/internal/dag/accessors.go
+++ b/internal/dag/accessors.go
@@ -18,35 +18,10 @@ import (
 	"strconv"
 
 	"github.com/projectcontour/contour/internal/annotation"
-	"github.com/projectcontour/contour/internal/k8s"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
-
-// RouteServiceName identifies a service used in a route.
-type RouteServiceName struct {
-	Name      string
-	Namespace string
-	Port      int32
-}
-
-// GetServices returns all services in the DAG.
-func (dag *DAG) GetServices() map[RouteServiceName]*Service {
-	getter := serviceGetter(map[RouteServiceName]*Service{})
-	dag.Visit(getter.visit)
-	return getter
-}
-
-// GetService returns the service in the DAG that matches the provided
-// namespace, name and port, or nil if no matching service is found.
-func (dag *DAG) GetService(meta types.NamespacedName, port int32) *Service {
-	return dag.GetServices()[RouteServiceName{
-		Name:      meta.Name,
-		Namespace: meta.Namespace,
-		Port:      port,
-	}]
-}
 
 // EnsureService looks for a Kubernetes service in the cache matching the provided
 // namespace, name and port, and returns a DAG service for it. If a matching service
@@ -62,11 +37,14 @@ func (dag *DAG) EnsureService(meta types.NamespacedName, port intstr.IntOrString
 		return nil, err
 	}
 
-	if dagSvc := dag.GetService(k8s.NamespacedNameOf(svc), svcPort.Port); dagSvc != nil {
-		return dagSvc, nil
-	}
-
-	dagSvc := &Service{
+	// There's no need to walk the DAG to look for a matching
+	// existing Service here. They're terminal nodes in the DAG
+	// so nothing is getting attached to them, and when used
+	// to generate an Envoy cluster any copy of this info will
+	// do. Doing a DAG walk to look for them is also very
+	// inefficient and can cause performance isuses at scale
+	// (see https://github.com/projectcontour/contour/issues/4058).
+	return &Service{
 		Weighted: WeightedService{
 			ServiceName:      svc.Name,
 			ServiceNamespace: svc.Namespace,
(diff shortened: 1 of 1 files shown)


skriss:
Here's the graph I got running 1.18.1 + this patch:

image

Which looks much better than the 1.18.1 graph in https://github.com/projectcontour/contour/issues/4058#issue-1009797429 (and even looks better than the baseline 1.8.1 graph).

stevesloka (commented):
I think this is fine and a good catch on the perf issue! The only risk is that a service is looked up that isn't in the DAG (meaning it's not referenced from anything), but that really shouldn't happen or cause an issue if it would happen.

sunjayBhatia (approved):
LGTM, can't think of anything that this would cause other than as you mention small added memory footprint in the case you have say tons of services being referenced in a new httpproxy etc. that are already in the dag, would have another copy in memory, but something to optimize later if needed, we need some sort of benchmark in addition to creating 1000s of services/proxies that checks memory footprint

youngnick (approved):
LGTM, this is a great optimization.

dprotaso:
Awesome - is it low risk to cherry pick these changes?

skriss:
> Awesome - is it low risk to cherry pick these changes?

Yeah, it should be low-risk -- this code hasn't really changed at all since 1.9. Are there particular versions you'd be looking for patch releases on? We don't currently have any patch releases planned, and as a policy we generally only patch the latest minor release, but if you let us know what your needs are we can have the conversation.

dprotaso:
Knative currently ships Contour v1.14.x (0.23, 0.24, 0.25) and v1.18.x (0.26)

I think we're ok with a minor bump on the v1.18.x branch

PR #4063, opened 2021-09-29
move Envoy from daemonset to deployment

stevesloka:
This is a WIP playground on what it might look like to move Envoy from a daemonset to a Deployment. 

Motivation around this is the inability to have Envoy pod react to a `kubectl drain <node>` or a scale up/down of nodes since daemonsets are not applied in those events. 

```
 If there are DaemonSet-managed pods, drain will not proceed without –ignore-daemonsets, and regardless it will not delete any DaemonSet-managed pods, because those pods would be immediately replaced by the DaemonSet controller, which ignores unschedulable markings.
```

### DIFF
--- a/examples/contour/03-envoy.yaml
+++ b/examples/contour/03-envoy.yaml
@@ -1,16 +1,19 @@
 ---
 apiVersion: apps/v1
-kind: DaemonSet
+kind: Deployment
 metadata:
   labels:
     app: envoy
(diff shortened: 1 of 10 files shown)


stevesloka:
The upgrade test is failing since it's upgrading from a daemonset to a deployment which isn't a supported model the way it's currently configured.

PR #4064, opened 2021-09-30
[WIP] Changelog plus CI check for changelogs.

youngnick:
This is currently WIP, because checking Actions stuff locally is impossible, sigh.

Not ready for review yet.


hack/actions/check-changefile-exists.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	changelogFile, err := os.Stat(fmt.Sprintf("./changelogs/unreleased/%d-%s-%s.md",
+		pr, *prDetails.User.Login, category))
+
+	if os.IsNotExist(err) {
+		log.Fatalf(`
+Thanks for your PR, and thanks for labelling it with a release-note.
+For a PR to be accepted to Contour, it must have a file named
+changelogs/unreleased/%d-%s-%s.md with a description of the change.`,
```
tsaarni, line 122: Just some general questions and thoughts about the changelog files:

I'm guessing that category is part of the changelog filename to make it easier to collect the entries under certain heading in the release notes? And the reason for user name in filename is for giving shoutout for the contributor in the release notes?

In theory this info could be also collected at the time of release from the github PRs. Category can possibly change during development of the PR, and I assume user may change the label even after the check was already run. It might also happen more often than not, that the used template file was wrong, causing e.g. formatting problems with wrong heading levels.

There probably is many projects using automatic changelogs, but I've seen similar process in Vault where the changelog entries used format like this https://github.com/hashicorp/vault/ including some metadata in the document itself, collected into sections like this https://github.com/hashicorp/vault/blob/main/CHANGELOG.md. Though that seems bit more limited, allowing only one-liners and not longer descriptions / formatting.

hack/actions/check-changefile-exists.go
```diff
@@ -0,0 +1,136 @@
...
+	tc := oauth2.NewClient(ctx, ts)
+
+	client := github.NewClient(tc)
+
+	prDetails, _, err := client.PullRequests.Get(ctx, "projectcontour", "contour", pr)
+	if err != nil {
+		log.Fatalf("Couldn't get PR details: %s", err)
+	}
+
+	// No labels. This is most likely what people will see at first, so this message
```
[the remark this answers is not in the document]
youngnick: That's what I was trying to do here - if the initial set of checks fails, then there's no point running the e2es, upgrade tests etc, because they probably won't pass anyway. The initial checks are changefile, linter, codegen, and codespell. These take about a minute to run, so people can get fast feedback on the initial things. Once those are passing, then the longer checks will run.

I definitely tried to make the errors in this script say all the steps needed, but I'll do another pass to check that it's identifying the first thing to fix.

hack/actions/check-changefile-exists.go
youngnick: Yes, right now we have very detailed changelogs, and that's something I definitely wanted to keep, so I wanted to have a mechanism to be able to have a reasonably large amount of explanation for any particular change.

I lifted a lot of this design from Velero and Envoy, as it seems to be working reasonably well for them.

I agree that the downside is that, best case, you have to do a second ccommit with the changelog for your PR once you know the number, but I think the benefits we will get will be worth it. In particular, I would like for it not to take a full day of writing release notes before we can release, while hopefully keeping the changelogs to the standard we've been delivering.

hack/actions/check-changefile-exists.go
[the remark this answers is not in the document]
tsaarni: Yes absolutely, optimizing on those tests is the BIG win!   

I was only thinking about these very minor steps with the new changelog process, which now do:

1. check if any labels exists in issue -> if not, fail and point out general requirements
2. check if any of the labels can be parsed `release-note/<some string>` -> if not, fail and point out general requirements
3. check if file `./changelogs/unreleased/%d-%s-%s.md` exists -> if not, fail and point the exact name of missing file

in theory, it could do bit more work in advance and pinpoint e.g. "you do not seem to have github label `release-note/<category>` attach to the PR and changelog file `./changelogs/unreleased/%d-%s-%s.md` seems to be missing" at one go.  But this is just minor thing of course!

hack/actions/check-changefile-exists.go
tsaarni: Yeah, having to do the second commit is a minor nuisance for the benefit of not requiring a full work-day for release notes!

As an alternative to putting "metadata" like `category` and `username` in filename, there could be release tooling that fetches that info on-demand when writing release notes.  Username of course does not change during the life-cycle of the PR, but category might change. 

But same as in my other comment: this is just a minor thing that came to my mind, thinking what could go wrong in practice :)

hack/actions/check-changefile-exists.go
```diff
@@ -0,0 +1,136 @@
...
+
+			// Otherwise, extract the category.
+			labelSplit := strings.Split(name, "/")
+			if len(labelSplit) > 1 {
+				category = labelSplit[1]
+			}
+		}
+	}
+
+	if category == "" {
```
tsaarni, line 98: One more question came to my mind: When PR is missing the correct category label, user would add it in github, but that would not re-trigger this check? Would is require a dummy commit to fix that error or are there other ways for the user to re-trigger CI run?

.github/workflows/prbuild.yaml
```diff
@@ -7,7 +7,86 @@ env:
   GOPROXY: https://proxy.golang.org/
   SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
 jobs:
+  check-changelog:
```
sunjayBhatia, line 10: should we skip this whole job on `main` commits?

changelogs/unreleased/major-sample.md
```diff
@@ -0,0 +1,11 @@
+## Sample Major Change
```
stevesloka, line 1: Maybe for these also provide a real-world example? Pick out a feature we've done and write it up with actual implementation? Sometimes is helpful.

hack/actions/check-changefile-exists.go
```diff
@@ -0,0 +1,136 @@
...
+
+	// No labels. This is most likely what people will see at first, so this message
+	// is as friendly as I could make it.
+	if len(prDetails.Labels) == 0 {
+		log.Fatal(`
+Thanks for your PR.
+For a PR to be accepted to Contour, it must have:
+- at least one release-note label set
+- a file named changelogs/unreleased/PR#-author-category,
+	where category matches the relase-note/category label you apply`)
```
stevesloka, line 74: Maybe link to the docs on how this process works so folks can find out more information?

hack/actions/check-changefile-exists.go
```diff
@@ -0,0 +1,136 @@
...
+	client := github.NewClient(tc)
+
+	prDetails, _, err := client.PullRequests.Get(ctx, "projectcontour", "contour", pr)
+	if err != nil {
+		log.Fatalf("Couldn't get PR details: %s", err)
+	}
+
+	// No labels. This is most likely what people will see at first, so this message
+	// is as friendly as I could make it.
+	if len(prDetails.Labels) == 0 {
```
stevesloka, line 68: Could we layer on something like this (https://github.com/marketplace/actions/require-labels) to get faster feedback without having to drill into the logs?

stevesloka (commented):
Just couple of small things to chat about. Also, don't forget to add 1.18.2 which just came out last week.

.github/workflows/prbuild.yaml
youngnick: Yes, we should.

hack/actions/check-changefile-exists.go
youngnick: That's a great idea, thanks. I'll bring that in as well.

hack/actions/check-changefile-exists.go
youngnick: I think I've found a way to fix this, actually, but it will mean that the full CI suite will run on every relabel. C'est la vie.

.github/workflows/prbuild.yaml
youngnick: We actually have a separate build workflow for `main`, so I think we should just skip this workflow altogether for `main`. I've added this in.

hack/actions/check-changefile-exists.go
youngnick: Made some further changes and added this in.

hack/actions/check-changefile-exists.go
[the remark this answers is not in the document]
youngnick: I've added the Action that @stevesloka suggested as a prerequisite for the changefile check, so hopefully the other cases should not come up.

hack/actions/check-changefile-exists.go
youngnick: I've just confirmed this works - the CI will now rerun on label/unlabel events.

youngnick:
This is ready for rereview now, I think. I've updated the description.

hack/actions/check-changefile-exists.go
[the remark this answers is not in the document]
tsaarni: Cool! This now of course made label checks in `check-changelog-exists.go` redundant and they could be even removed...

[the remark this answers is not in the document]
youngnick: I figured I would leave them in as extra precautions.

tsaarni (approved):
lgtm!

changelogs/unreleased/4064-youngnick-infra.md
```diff
@@ -0,0 +1,9 @@
+## New Changelog process
```
sunjayBhatia, line 1: nit
```suggestion
### New Changelog process
```

sunjayBhatia (approved):

.github/workflows/prbuild.yaml
```diff
@@ -1,13 +1,110 @@
 name: Build and Test Pull Request
 
 # Trigger the workflow on push or pull request
-on: [push, pull_request]
+on: 
+  push:
+    branches-ignore:
+      - "main"
```
skriss, line 7: Why disable this workflow for pushes to `main`? That means we're only running tests against PRs, not against what's in `main`.

.github/workflows/prbuild.yaml
```diff
@@ -1,13 +1,110 @@
...
+    name: Check for changelog file
+    needs:
+      - check-label
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - run: go run ./hack/actions/check-changefile-exists.go
+      env:
+        PR_NUMBER: ${{ github.event.number }}
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
skriss, line 36: Related to previous comment, I'm thinking it might be easier to move these jobs into a separate workflow that can be triggered just for PRs, on all the label/unlabel/etc events, and leave `prbuild.yaml` more or less as-is.

youngnick:
Once this is merged, I'll update the branch protection rules to make the label checks required as well.

PR #4065, opened 2021-09-30
cmd/contour: Update to TLS 1.3 for xDS between Contour and Envoy

### LINKED ISSUE #3518: Configurable option for TLS version between envoy and contour
Hi

I asked a question on slack, if it is possible to set the minimum version of TLS between envoy and contour. I got that answer that in the moment it is not. I think it would be nice to have that.

Thank you for taking into consideration.

Br,
Bence

tsaarni:
This change sets TLSv1.3 as the minimum TLS protocol version for Contour xDS server.

Envoys that still run with old bootstrap config during upgrade, will fail to connect to new Contour if they happen to connect to a server that was upgraded first. This is because Envoy explicitly refuses to connect to TLSv1.3 servers. It defaults to maximum version TLSv1.2 (see [doc here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto.html#extensions-transport-sockets-tls-v3-tlsparameters)). After the upgrade proceeds and Envoys have been updated with the latest Contour init container, the bootstrap config will be set to allow TLSv1.3 and connecting to xDS server will proceed again normally.

Fixes #3518

Open issues:

- [ ] ~~Alternative implementation: To avoid the above upgrade scenario, we could make xDS server TLS version configurable and let it remain min TLSv1.2 by default. This allows new Contour to serve Envoys with old bootstrap config during upgrade. When all Envoys have been upgraded, the administrator would manually need to change the minimum version to TLSv1.3.~~
- [X] Documentation: there is no cipher compatibility issues to document in this case: there is no configurability for TLSv1.3 implementations at this point.  But since Envoy defaults to max TLSv1.2, the above upgrade scenario would require docs to draw user's attention to the upgrade. Exactly what is to be documented depends on chosen solution.


tsaarni:
What we could do, is to set the max version to TLSv1.3 in the bootstrap. If that had been done already much earlier, Envoys running in deployed environments today would be able to connect to TLSv1.3 servers and upgrade problem would have been avoided. 

Of course, setting max version today, would not help the transition in near future.

youngnick:
After #4081 merging, this should now be mergable in 1.20.

sunjayBhatia (approved):

tsaarni:
I've added major release note on this to draw the attention of users. Please take a look and comment!

changelogs/unreleased/4065-tsaarni-major.md
```diff
@@ -0,0 +1,9 @@
+## xDS management connection between Contour and Envoy set to TLSv1.3
+
+The minimum accepted TLS version for Contour xDS server is changed from TLSv1.2 to TLSv1.3.
+Previously in Contour 1.19, the maximum accepted TLS version for Envoy xDS client was increased to TLSv1.3 which allows it to connect to Contour xDS server using TLSv1.3.
+
+If upgrading from version **prior Contour 1.19**, the old Envoys will be unable to connect to new Contour until also Envoys are upgraded.
```
youngnick, line 6: Minor nit:
```suggestion
If upgrading from a version **prior to Contour 1.19**, the old Envoys will be unable to connect to new Contour until also Envoys are upgraded.
```

youngnick (approved):
LGTM, great changelog message as well.

changelogs/unreleased/4065-tsaarni-major.md
tsaarni: Done!

sunjayBhatia (approved):

Issue #3996, opened 2021-08-27
Fix generate-gateway-deployment.sh and contour-gateway.yaml in older release branches


skriss:
xref https://github.com/projectcontour/contour/pull/3995

This script is inadvertently uncommenting all lines starting with `namespace:` from the Contour config file in an attempt to uncomment *just* the `gateway:` config stanza. Ideally we can fix the script and the output in past release branches to keep those branches healthy. Note that this is not an issue on `main` anymore since we dropped namespace/name config for Gateway.

skriss:
I'm going to close this out since the release branches that this affects are now pretty old.

PR #4066, opened 2021-10-01
[release-1.18] internal/dag: don't walk DAG to look for service (#4061)

### LINKED ISSUE #4058: Performance Regression v1.8.1 => v1.18.1

skriss:
Removes code that walks the DAG looking for
a matching existing Service when adding a
new Service. This code is very inefficient
when the DAG is large and creates a
performance bottleneck. Some extra memory
may end up being used if duplicate Services
are created, but it's unlikely to be a
material issue in practice.

Closes #4058.

### DIFF
--- a/internal/dag/accessors.go
+++ b/internal/dag/accessors.go
@@ -18,35 +18,10 @@ import (
 	"strconv"
 
 	"github.com/projectcontour/contour/internal/annotation"
-	"github.com/projectcontour/contour/internal/k8s"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
-
-// RouteServiceName identifies a service used in a route.
-type RouteServiceName struct {
-	Name      string
-	Namespace string
-	Port      int32
-}
-
-// GetServices returns all services in the DAG.
-func (dag *DAG) GetServices() map[RouteServiceName]*Service {
-	getter := serviceGetter(map[RouteServiceName]*Service{})
-	dag.Visit(getter.visit)
-	return getter
-}
-
-// GetService returns the service in the DAG that matches the provided
-// namespace, name and port, or nil if no matching service is found.
-func (dag *DAG) GetService(meta types.NamespacedName, port int32) *Service {
-	return dag.GetServices()[RouteServiceName{
-		Name:      meta.Name,
-		Namespace: meta.Namespace,
-		Port:      port,
-	}]
-}
 
 // EnsureService looks for a Kubernetes service in the cache matching the provided
 // namespace, name and port, and returns a DAG service for it. If a matching service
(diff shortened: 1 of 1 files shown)


skriss:
Codegen is failing due to https://github.com/projectcontour/contour/issues/3996, should be ignored since the script is broken.

stevesloka:
@skriss codegen failed?

skriss:
> @skriss codegen failed?

See https://github.com/projectcontour/contour/pull/4066#issuecomment-932310765

stevesloka:
Oh right, forgot about that!

stevesloka (approved):

sunjayBhatia (approved):

PR #4071, opened 2021-10-04
Fixup `go get` from kubernetes toolchain install script since go get is deprecated

stevesloka:
`go get` is deprecated in go v1.17:

```
go get: installing executables with 'go get' in module mode is deprecated.
        To adjust and download dependencies of the current module, use 'go get -d'.
        To install using requirements of the current module, use 'go install'.
        To install ignoring the current module, use 'go install' with a version,
        like 'go install example.com/cmd@latest'.
        For more information, see https://golang.org/doc/go-get-install-deprecation
        or run 'go help get' or 'go help install'.
```


youngnick (approved):
Thanks @stevesloka!

PR #4074, opened 2021-10-05
ContourConfiguration status

stevesloka:
Follow up to #4048 to wire up a way to set status on the ContourConfiguration object. Future refactoring will enable the ability to add more validations and setting status accordingly. This only sets a "valid" status on the object relying on Kubebuilder annotations for all current validations.

### DIFF
--- a/cmd/contour/contour.go
+++ b/cmd/contour/contour.go
@@ -14,15 +14,23 @@
 package main
 
 import (
+	"context"
+	"fmt"
 	"os"
 
 	resource_v3 "github.com/envoyproxy/go-control-plane/pkg/resource/v3"
+	contour_api_v1alpha1 "github.com/projectcontour/contour/apis/projectcontour/v1alpha1"
 	"github.com/projectcontour/contour/internal/build"
 	"github.com/projectcontour/contour/internal/envoy"
 	envoy_v3 "github.com/projectcontour/contour/internal/envoy/v3"
 	"github.com/projectcontour/contour/internal/k8s"
+	"github.com/projectcontour/contour/internal/workgroup"
 	"github.com/sirupsen/logrus"
 	kingpin "gopkg.in/alecthomas/kingpin.v2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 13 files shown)


youngnick:
The first commit looks good; are you planning to only have a way to set status to `True` to start with? That's fine, just want to understand the schedule.

cmd/contour/contour.go
```diff
@@ -120,7 +129,67 @@ func main() {
...
+				log.WithError(err).Fatalf("error converting contour configuration %s", namespacedName)
+			}
+
+			// Set valid status on the ContourConfiguration object
+			contourConfig.Status = contour_api_v1alpha1.ContourConfigurationStatus{
+				Conditions: []contour_api_v1.DetailedCondition{{
+					Condition: contour_api_v1.Condition{
+						Type:               "Valid",
+						Status:             "True",
+						ObservedGeneration: 0,
```
youngnick, line 171: This should be set to whatever `metadata.generation` is on the `contourConfig` object. Not a huge deal, but that's what the `observedGeneration` says: I based this Condition on this `metadata.generation` of the object.

cmd/contour/contour.go
```diff
@@ -126,7 +135,61 @@ func main() {
...
+						LastTransitionTime: metav1.NewTime(time.Now()),
+						Reason:             "ValidContourConfiguration",
+						Message:            "Valid ContourConfiguration",
+					},
+					Errors:   nil,
+					Warnings: nil,
+				}},
+			}
+
+			if err := serve.sh.SetContourConfigurationStatus(&contourConfig); err != nil {
```
skriss, line 181: This new method doesn't take into account leader election, so all Contour replicas are going to try to write status which could be problematic. Guessing there's an order of operations problem preventing using the existing `serve.sh.Writer().Send()` method since the StatusUpdateHandler hasn't been started yet?

stevesloka: Yeah that's a good point. I started down that road but then thought getting to the group.Start() bit was overkill since this will just write then quit. Let me see about doing another refactor to things to allow for this better.

stevesloka:
> as i understand it, before, the leaderelection bits happened more async with the server starting

Yup that's right.

PR #4075, opened 2021-10-05
Bump Envoy to v1.20.1

sunjayBhatia:
See release notes:
https://www.envoyproxy.io/docs/envoy/v1.20.0/version_history/current

Also bumps go-control-plane to latest

Deprecation of existing headermatch usage will need to be handled in the release after this is
merged. The "string match" header matching method is preferred but was
added in 1.20.1 and it was not present in 1.19.x, we will need to
release a version with Envoy 1.20.x compatibility and the subsequent
release will need to move to the "string match" method (rather than the
individual regex/exact/etc. methods.

### DIFF
--- a/internal/xds/v3/snapshotter.go
+++ b/internal/xds/v3/snapshotter.go
@@ -36,17 +36,11 @@ type snapshotter struct {
 	envoy_cache_v3.SnapshotCache
 }
 
-func (s *snapshotter) Generate(version string, resources map[envoy_types.ResponseType][]envoy_types.Resource) error {
+func (s *snapshotter) Generate(version string, resources map[envoy_resource_v3.Type][]envoy_types.Resource) error {
 	// Create a snapshot with all xDS resources.
 	snapshot, err := envoy_cache_v3.NewSnapshot(
 		version,
-		map[envoy_resource_v3.Type][]envoy_types.Resource{
-			envoy_resource_v3.EndpointType: resources[envoy_types.Endpoint],
-			envoy_resource_v3.ClusterType:  resources[envoy_types.Cluster],
-			envoy_resource_v3.RouteType:    resources[envoy_types.Route],
-			envoy_resource_v3.ListenerType: resources[envoy_types.Listener],
-			envoy_resource_v3.SecretType:   resources[envoy_types.Secret],
-		},
+		resources,
 	)
 	if err != nil {
 		return err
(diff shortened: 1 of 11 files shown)


sunjayBhatia:
Not sure if we want to merge for 1.19 or not cc @projectcontour/maintainers

internal/xds/v3/snapshotter.go
```diff
@@ -37,16 +38,19 @@ type snapshotter struct {
...
-		nil,
-		resources[envoy_types.Secret],
-		nil,
+		map[envoy_resource_v3.Type][]envoy_types.Resource{
+			envoy_resource_v3.EndpointType: resources[envoy_types.Endpoint],
+			envoy_resource_v3.ClusterType:  resources[envoy_types.Cluster],
+			envoy_resource_v3.RouteType:    resources[envoy_types.Route],
+			envoy_resource_v3.ListenerType: resources[envoy_types.Listener],
+			envoy_resource_v3.SecretType:   resources[envoy_types.Secret],
+		},
```
skriss, line 49: Looks like we could probably update this method signature to take in a `map[envoy_resource_v3.Type][]envoy_types.Resource`, and update the caller to pass that in. Then could use `resources` directly here.

sunjayBhatia: ah yeah let me rework that, theres some room for improvement and de-duplication

skriss (approved):
:+1: I don't see any reason not to ship it in Contour 1.19.

sunjayBhatia:
added `do not merge` since this is a bit of a curious one, tests seem to be failing in CI but not locally "on my machine"

stevesloka:
I tried running on my machine and seems like some tests are timing out. 🤔

sunjayBhatia:
> > I tried running on my machine and seems like some tests are timing out. 🤔
> 
> yeah latest CI runs ~140s, my local machine ~975 and CI ~882, some issues here for sure

just switching back to the older Envoy, I get ~140s locally, drastic change from the 1.20.0 image

sunjayBhatia:
See https://github.com/envoyproxy/envoy/issues/18616

sunjayBhatia:
testing with image `docker.io/envoyproxy/envoy-dev:6f0ecc5ae16f3c39937dd13a3bc0327177a0e754` (and removed http client timeouts in tests) which has the fix from https://github.com/envoyproxy/envoy/pull/18686

wrowe:
@sunjayBhatia could you fix the PR title? Thx!

youngnick (approved):

skriss (approved):
🎉

PR #4080, opened 2021-10-06
test/e2e: Nil check before assertion on succeeded request

sunjayBhatia:
Log message uses status code from response which might be nil


skriss (approved):
LGTM, had been meaning to get to this for awhile so thanks for taking care of it!

PR #4086, opened 2021-10-08
Add metadata name to complete the minimum runnable example

TheApeMachine:
Adding the name provides a runnable example for the docs, as without it there is an error that `the resource name may not be empty`.


TheApeMachine:
@youngnick You are very welcome. Must say I like Contour so far. Wasn't entirely painless to get working (though I am only slowly getting to intemediate when it comes to understanding networking in Kubernetes), but didn't take me that long either. The main thing was to discover HTTPProxy over Ingress, and this errors section of HTTPProxy as it showed the reasons related to CORS and WebSockets why Grafana was throwing an error about not being able to connect to it's live stream :) Reading the manual helped :p

PR #4098, opened 2021-10-12
only run label check workflow for PRs

skriss:
The label check workflow is only relevant
for PRs. This disables it for all push
builds, not just for pushes to "main", to
avoid failures on other branches.


skriss:
Yep, based on https://github.com/projectcontour/contour/blob/76c898fcd245438e49ab93e283ce4b8f2096499d/.github/workflows/label_check.yaml#L5-L6 (just tested it).

Issue #4072, opened 2021-10-04
consider dropping Visitor pattern for querying DAG info


skriss:
We should consider dropping the Visitor pattern that's used to extract information from Contour's internal model ("the DAG"), and instead have the internal model provide methods for getting information out of it. This would result in simpler code and better runtime performance. 

The Visitor pattern is useful when you want to separate the algorithm that operates on an object structure from the object structure itself. This is useful for extensibility, i.e. when an object structure is provided as part of a library, or when the object structure itself changes rapidly or is not known ahead of time. These are not the case in Contour - the user of the DAG is Contour code, and the DAG structure is well-defined and changes infrequently. So we can easily define methods on the DAG struct that understand the structure of the internal model and simply and efficiently extract it for consumers. IMHO this is more straightforward code that's easier to work with.

On the performance front, using the Visitor pattern ends up being a pretty inefficient way to get information out of the internal model, particularly with large sets of configuration, because the *entire* DAG is visited to get any subset of information out if it. This is because the Visitor pattern makes no assumptions about the structure of the DAG, but instead assumes that any type (e.g. a Listener) could be located at any place in the object structure. If we directly defined methods for getting e.g. Clusters or VirtualHosts out of the model, those methods would "know" where to find that information, and get it in the most efficient way, rather than needlessly visiting irrelevant parts of the object structure.

Interested in others' thoughts here.

PR #4099, opened 2021-10-12
drop Visitor pattern for getting info out of DAG

skriss:
This is roughly what I had in mind for #4072. 

With the knative perf test, when there are ~1000 knative services already in the cluster (~4000 HTTPProxies), this drops the time to process an additional one from ~18 seconds (Contour 1.18.2) down to ~10 seconds.

Still needs a bit of cleanup but the general approach is in place.


internal/dag/accessors.go
```diff
@@ -108,116 +109,193 @@ func externalName(svc *v1.Service) string {
...
+	if svh := d.GetSecureVirtualHost(hostname); svh != nil {
 		return svh
 	}
 
 	svh := &SecureVirtualHost{
 		VirtualHost: VirtualHost{
-			Name:         ln.Name,
-			ListenerName: ln.ListenerName,
+			Name:         hostname,
+			ListenerName: "ingress_https",
```
sunjayBhatia, line 129: nit: maybe we can use a constant here? there is one in the `internal/xdscache/v3` package but not sure that makes sense to use that

internal/dag/gatewayapi_processor.go
```diff
@@ -540,7 +540,7 @@ func (p *GatewayAPIProcessor) computeTLSRoute(route *gatewayapi_v1alpha2.TLSRout
 		}
 
 		for host := range hosts {
-			secure := p.dag.EnsureSecureVirtualHost(ListenerName{Name: host, ListenerName: "ingress_https"})
```
sunjayBhatia, line 543: i vaguely remember this whole ListenerName thing was to enable more than just the 2 basic listeners

makes sense maybe to go back to this and could use a bigger refactor in the future once we support that? wdyt?

sunjayBhatia:
I'm on board with this change, def a fan of `the user of the DAG is Contour code, and the DAG structure is well-defined and changes infrequently. So we can easily define methods on the DAG struct that understand the structure of the internal model and simply and efficiently extract it for consumers` from the linked issue

internal/dag/accessors.go
skriss: yeah absolutely. I was gonna do a follow-up PR to move the ListenerName onto the Listener struct, so I'll add/use a const as part of that. Need to double-check on the many listeners thing though to make sure I'm not undoing past work here.

internal/dag/gatewayapi_processor.go
skriss: Let me look at this some more, but one thought here is that I'd like to push more of the Listener logic into the DAG -- right now they're mostly built out in the xdscache logic itself, which feels off and inconsitent from other resource types. Perhaps as part of that refactor we can make sure that the DAG properly supports >2 listeners and that the xdscache can consume them.

skriss: Yeah so I'm thinking that if we want to retain support for many Listeners, we should properly build that out in the DAG - model out all of the Listeners, attach vhosts to the relevant listener(s) there, and then this ListenerName field on the VirtualHost is probably not needed, we'll just traverse the association from Listener->VirtualHosts.

youngnick: I agree with that last comment, @skriss. I think that we definitely should be aiming to be able to do multiple listeners, so that in the future, we can allow some extra ports and have the managed Envoy stuff ensure the wiring.

internal/dag/listener_processor.go
```diff
@@ -32,74 +32,52 @@ func (p *ListenerProcessor) Run(dag *DAG, _ *KubernetesCache) {
...
-		VirtualHosts: virtualhosts,
+		VirtualHosts: vhosts,
 	}
 
-	dag.AddRoot(http)
+	dag.Listeners = append(dag.Listeners, http)
 }
 
 // buildHTTPSListener builds a *dag.Listener for the vhosts bound to port 443.
 // The list of virtual hosts will attached to the listener will be sorted
```
youngnick, line 59: Nit: You didn't do this, but any chance while you are here:
```suggestion
// The list of virtual hosts attached to the listener will be sorted
```
I *think* that's a more accurate description?

youngnick (commented):
The approach is looking great to me, feels like I would sum it up as that it's changing the DAG from being a generic DAG model to being more domain-specific. It's probably more correct to just call it a data model now, rather than a DAG, I guess.

I'd be happy for this to go in as-is, but wasn't sure if @skriss had more changes to put in here, so not approving yet.

skriss:
I think this can get merged as-is (pending any further comments) and I can follow up with subsequent PRs for:
- moving ListenerName to be a property of Listener in the DAG
- renaming "DAG" to "Model" or similar
- more Listener-related refactoring to move logic from the xdscache into the DAG processor, and fully supporting multiple listeners there

sunjayBhatia (approved):

stevesloka (approved):

stevesloka:
Rhu-Rho @skriss, conflicts!

skriss:
Got 'em resolved, just gonna do a little extra testing before merging since there were a couple big conflicts to deal with.

PR #4100, opened 2021-10-12
Use envoy xds server type in daily e2e tests

sunjayBhatia:
Run e2e tests daily against envoy xds server type that uses
go-control-plane.

Once we are confident it is stable, we can flip the default to be the
envoy go-control-plane variant and run the daily tests with the contour
variant until we drop support for it.


sunjayBhatia:
bumping the bootstrap connect timeout seems to have helped, it's hardcoded at 5s for now, maybe should be configurable

youngnick:
This sounds like another reason we need to fix up the initialization process and have a way to ensure the ordering for some goroutines?

sunjayBhatia:
recent refactors have made https://github.com/projectcontour/contour/pull/4100#issuecomment-951107288 obsolete, the event handler is now notified of leader election when it is enabled and disabled through the leadership notifier runnable that triggers it to run an update

skriss (approved):
LGTM.

One other idea here would be to set up a nightly run of the full E2E suite using the "envoy" control plane, to get regular full coverage on it without blowing up our E2E matrix for PRs.

sunjayBhatia:
> LGTM.
> 
> One other idea here would be to set up a nightly run of the full E2E suite using the "envoy" control plane, to get regular full coverage on it without blowing up our E2E matrix for PRs.

yeah i like that better actually, going to update to do that instead of the smoke tests

skriss (approved):
nice, LGTM.

I'm now thinking maybe we should re-jigger https://github.com/projectcontour/contour/pull/4126 to also run its tests in this daily run, instead of running a small subset for every PR, for similar reasons as for this one. Can ponder/tackle that separately.

stevesloka (approved):

PR #4103, opened 2021-10-12
changelogs: Clean up for the v1.19.0 release

stevesloka:


skriss:
Just realized there's nothing about the new `ContourConfiguration` CRD, we should probably mention that first :)

Issue #4097, opened 2021-10-12
Multiple ingress class values


erwbgy:
**Allow the contour --ingress-class-name value to be a comma-separated list of classes to match against**

We run multiple Contour instances in each of our clusters for different groups with each instance having a different ingress class value.  In less busy clusters this results in wasted resources with many instances largely idle so we'd like to be able to share a single Contour instance but have it watch objects for multiple ingress classes.  This way users can continue to use their own ingress classes and not have to reconfigure when deploying to different clusters but we avoid running multiple instances when we don't need to.

For example:
```
  --ingress-class-name=contour,team1,team2
```
would process HTTPProxy or Ingress objects that have an ingress.class annotation or ingressClassName field with any of these values.

If you're open to this change it is a relatively small change to the matches function in internal/ingressclass/ingressclass.go (excluding tests) and I can create a pull request.

youngnick:
It seems like this is because users want to reuse the same Ingress object in multiple clusters, that may have different Contour setups installed. So you want to have the users be able to use the same, unmodified Ingress object wherever, and have the ingress-class magically hook up to the right Contour depending on where you're running.

Having a way to have Contour listen to multiple ingress-classes makes sense, I guess. I wonder how this would work if you are using the IngressClass object as well though...

tl;dr I can't see any issues off the top of my head, but it would be nice to have a consistent story for this across different ways of configuring ingress-class. I'll also keep this in mind for Gateway and GatewayClass for the future.

PR #4109, opened 2021-10-13
Multiple ingress classes

erwbgy:
Closes https://github.com/projectcontour/contour/issues/4097 Allow the contour --ingress-class-name value to be a comma-separated list of classes to match against.


stevesloka (dismissed):
- Just one small nit on the release note
- Maybe we should add some validation of the ingressClassName field now to make sure it is a comma separated list? I guess worst case it validates against the wrong string, but might be difficult to figure that out? 
- Would be neat to have some tests to show say two ingress objects, two httpproxy that match multiple classes.

youngnick:
@erwbgy, have you had a chance to check out the e2e tests?

erwbgy:
> @erwbgy, have you had a chance to check out the e2e tests?

Sorry I haven't and will be unlikely to be able to do so before Friday at the earliest.  If anyone else can see what the issues are and fix, please do!

test/e2e/httpproxy/multiple_ingress_classes_test.go
```diff
@@ -0,0 +1,121 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func testMultipleIngressClassesField(namespace string) {
+	Specify("multiple ingress class names with a single Contour instance (ingressClassName field)", func() {
+		t := f.T()
+
+		f.Fixtures.Echo.Deploy(namespace, "ingress-conformance-echo")
+
+		for _, class := range []string{"", "contour", "team1"} {
```
skriss, line 33: An empty ingress class for the HTTPProxy wouldn't be valid here since Contour has been explicitly configured to process specific ingress classes.

test/e2e/httpproxy/multiple_ingress_classes_test.go
```diff
@@ -0,0 +1,121 @@
...
+	})
+}
+
+func testMultipleIngressClassesAnnotation(namespace string) {
+	Specify("multiple ingress class names with a single Contour instance (annotation)", func() {
+		t := f.T()
+
+		f.Fixtures.Echo.Deploy(namespace, "ingress-conformance-echo")
+
+		for _, class := range []string{"", "contour", "team1"} {
```
skriss, line 80: ditto re: empty ingress class

test/e2e/ingress/ingress_class_test.go
```diff
@@ -0,0 +1,76 @@
...
+		name := class + "ingress"
+
+		f.Fixtures.Echo.Deploy(namespace, "echo")
+
+		i := &networkingv1.Ingress{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: namespace,
+				Name:      name,
+			},
+			Spec: networkingv1.IngressSpec{
```
skriss, line 41: Need to specify an ingress class name here for the same reason as with HTTPProxy.

test/e2e/httpproxy/multiple_ingress_classes_test.go
erwbgy: Thanks @skriss ! Such a simple fix but I couldn't see it.

test/e2e/ingress/ingress_class_test.go
erwbgy: Fixed, thanks.

test/e2e/httpproxy/multiple_ingress_classes_test.go
erwbgy: Fixed, thanks.

changelogs/unreleased/4109-erwbgy-small.md
```diff
@@ -0,0 +1 @@
+Allow the contour --ingress-class-name value to be a comma-separated list of classes to match against.
```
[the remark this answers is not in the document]
erwbgy: Fixed e2e tests with help from @skriss

youngnick:
This is looking pretty good to me, we'll need one of the other maintainers to take a look.

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -138,9 +138,9 @@ type TLS struct {
 
 // IngressConfig defines ingress specific config items.
 type IngressConfig struct {
-	// Ingress Class Name Contour should use.
+	// Ingress Class Names Contour should use.
 	// +optional
-	ClassName *string `json:"className,omitempty"`
+	ClassName []string `json:"className,omitempty"`
```
skriss, line 143: May as well update the names while the API is alpha:
```suggestion
	ClassNames []string `json:"classNames,omitempty"`
```

changelogs/unreleased/4109-erwbgy-small.md
```diff
@@ -0,0 +1 @@
+Allow the contour --ingress-class-name value to be a comma-separated list of classes to match against.  Contour will process Ingress and HTTPProxy objects with any of the specified ingress classes.
```
skriss, line 1: Even though the ContourConfiguration CRD is alpha, I think it's worth calling out in the release note that the name (if you accept my previous comment) and type of the ingress class field are changing in a non-backwards-compatible way.

internal/ingressclass/ingressclass.go
```diff
@@ -54,6 +56,11 @@ func matches(objIngressClass, contourIngressClass string) bool {
 		return objIngressClass == "" || objIngressClass == DefaultClassName
 	}
 
-	// Otherwise, the object's ingress class must match Contour's.
-	return objIngressClass == contourIngressClass
+	// Otherwise, the object's ingress class must match one of Contour's.
+	for _, contourIngressClassEntry := range strings.Split(contourIngressClass, ",") {
```
skriss, line 60: let's add a test case to https://github.com/projectcontour/contour/blob/main/internal/ingressclass/ingressclass_test.go#L26 and https://github.com/projectcontour/contour/blob/main/internal/ingressclass/ingressclass_test.go#L111 for a comma-separated `contourIngressClass`.

internal/dag/cache.go
```diff
@@ -94,8 +95,10 @@ func (kc *KubernetesCache) matchesIngressClass(obj *networking_v1.IngressClass)
 	}
 	// Otherwise, the name of the ingress class must match what has been
 	// configured.
-	if obj.Name == kc.IngressClassName {
-		return true
+	for _, contourIngressClassEntry := range strings.Split(kc.IngressClassName, ",") {
```
skriss, line 98: I don't think we're actually using the DAG cache's `ingressclass` field (https://github.com/projectcontour/contour/blob/81cbcabd93b5051e6c46ae0f122089a4fb8af926/internal/dag/cache.go#L54), which this function `matchesIngressClass` provides a filter for. Since Contour can now process >1 ingress class, I'd rather drop the field entirely, and this function, if it's truly not being used; otherwise, it would make sense that it'd need to be converted to a map or slice to support multiple ingress classes.

skriss (changes requested):
Thanks for fixing those tests @erwbgy. Had a couple of additional comments

skriss:
@erwbgy just checking in to see if you'll be able to update this PR to resolve the conflicts and address the few comments, would be great to get this in before the next release!

changelogs/unreleased/4109-erwbgy-small.md
erwbgy: Updated release note.

internal/dag/cache.go
erwbgy: I don't feel confident making DAG changes beyond simple ones like this.  Could this cleanup be deferred to a different MR?

erwbgy:
@skriss @stevesloka I think I have addressed all of your comments.

Looks like the flake mentioned in https://github.com/projectcontour/contour/issues/4305 caused the e2e failure.

test/e2e/httpproxy/httpproxy_test.go
```diff
@@ -428,3 +456,20 @@ descriptors:
...
+	}
+
+	if len(proxy.Status.Conditions) == 0 {
+		return false
+	}
+
+	cond := proxy.Status.GetConditionFor("Valid")
+	return cond.Status == "True"
+
+}
```
skriss, line 475: this function has moved into test/e2e/framework.go (`HTTPProxyValid`), so let's drop this copy and use that one instead.

internal/dag/cache.go
skriss: Yeah, we can do that, I'll create an issue for this.

skriss (commented):
Thanks for the updates @erwbgy, just one more small thing from me.

skriss:
> Looks like the flake mentioned in https://github.com/projectcontour/contour/issues/4305 caused the e2e failure.

Yep, looks like it, passed on a re-run.

test/e2e/httpproxy/httpproxy_test.go
skriss: I can do this along with #4327 as a follow-up since it's just cleanup.

skriss (approved):
LGTM, thanks @erwbgy!

PR #4123, opened 2021-10-20
Gateway API: add request redirect filter support

### LINKED ISSUE #4087: Gateway API: implement HTTPRequestRedirectFilter
ref. https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.HTTPRequestRedirectFilter

This is core in Gateway API v1alpha2.

skriss:
Adds support for the "RequestRedirect" HTTPRoute
filter type at the rule level.

Closes #4087.

### DIFF
--- a/internal/envoy/v3/route.go
+++ b/internal/envoy/v3/route.go
@@ -122,9 +122,11 @@ func RouteDirectResponse(response *dag.DirectResponse) *envoy_route_v3.Route_Dir
 // client.
 func RouteRedirect(redirect *dag.Redirect) *envoy_route_v3.Route_Redirect {
 	r := &envoy_route_v3.Route_Redirect{
-		Redirect: &envoy_route_v3.RedirectAction{
-			HostRedirect: redirect.Hostname,
-		},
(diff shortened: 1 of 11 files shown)


internal/envoy/v3/route.go
```diff
@@ -117,6 +117,39 @@ func RouteDirectResponse(response *dag.DirectResponse) *envoy_route_v3.Route_Dir
 	}
 }
 
+// RouteRedirect creates a *envoy_route_v3.Route_Redirect for the
+// redirect specified. This allows a redirect to be returned to the
+// client.
+func RouteRedirect(redirect *dag.Redirect) *envoy_route_v3.Route_Redirect {
```
sunjayBhatia, line 123: nit: do we usually add units for these small helpers? could be nice to avoid future mistakes in case of refactor

sunjayBhatia (approved):

internal/envoy/v3/route.go
skriss: :+1: added.

PR #4126, opened 2021-10-20
Add deployment example

stevesloka:
- Adds an example for running Envoy as a deployment vs a daemonset
- Updates the e2e tests to read an env var to determine if the deployment version should be run
- Enables running e2e tests for Envoy deployment on each PR build (😱)


stevesloka:
I thought about just having the deployment version run hourly or something to pick up issues vs running on each PR. That might reduce the churn time for each iteration.

examples/contour/03-envoy-deployment.yaml
```diff
@@ -0,0 +1,154 @@
+---
```
skriss, line 1: Only downside to putting this into `examples/contour` is that now `kubectl apply -f examples/contour` (which I use somewhat frequently) will create *both* a DaemonSet and a Deployment for Envoy. But I guess that's preferable to creating a full second copy of the directory.

skriss (commented):
:+1: to having this as an example. It'd also be good to update `examples/contour/README.md`, as well as some stuff in the docs about deployment options (e.g. https://projectcontour.io/docs/v1.19.1/deploy-options/, maybe some other pages too)

.github/workflows/prbuild.yaml
```diff
@@ -121,6 +121,8 @@ jobs:
         kubernetes_version: ["kubernetes:latest", "kubernetes:n-1", "kubernetes:n-2"]
         # run tests using the configuration crd as well as without
         config_type: ["ConfigmapConfiguration", "ContourConfiguration"]
+        # run tests against the daemonset or deployment model.
+        envoy_deployment_type: ["Daemonset", "Deployment"]
```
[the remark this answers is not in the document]
youngnick: That sounds like a good idea, I think we are risking running up against limits with expanding the test matrix this much - plus, we're going to be adding performance testing in soon as well.

youngnick (approved):
LGTM, I agree with @skriss about the risk of burning all our Actions time if we have a few PRs though.

.github/workflows/prbuild.yaml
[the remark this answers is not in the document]
stevesloka: Yeah that was my comment earlier: 

> I thought about just having the deployment version run hourly or something to pick up issues vs running on each PR. That might reduce the churn time for each iteration.

What I don't want is to have an issue with the configuration and to find out by a user that can't deploy using that model.

examples/contour/03-envoy-deployment.yaml
stevesloka: Yeah I do the same. =)

I had the deployment.yaml file outside of that directory but had troubles getting it worked in. Maybe I can try that path again to keep things separate.

.github/workflows/prbuild.yaml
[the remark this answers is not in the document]
youngnick: Ah sorry, I missed that comment. Having it be that the "main" tests (currently the Daemonset ones, but conceivably we could flip them if we wanted to) are the ones that get run on every PR and the other ones run on a timed basis make sense to me.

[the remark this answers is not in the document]
skriss: I'm OK with that approach too, odds are there won't be too many Deployment-specific issues that slip through.

examples/contour/03-envoy-deployment.yaml
```diff
@@ -0,0 +1,154 @@
...
+      terminationGracePeriodSeconds: 300
+      volumes:
+        - name: envoy-admin
+          emptyDir: {}
+        - name: envoy-config
+          emptyDir: {}
+        - name: envoycert
+          secret:
+            secretName: envoycert
+      restartPolicy: Always
```
r0bj, line 154: In comparison to DaemonSet variant, there is securityContext missing: https://github.com/projectcontour/contour/blob/main/examples/contour/03-envoy.yaml#L142-L145

stevesloka: :+1: Thanks for catching this @r0bj!

examples/contour/03-envoy-deployment.yaml
stevesloka: I moved it out into it's own dir so should keep the old way intact.

.github/workflows/prbuild.yaml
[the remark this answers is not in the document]
stevesloka: I copied the infra tests quickly. We can add some more or expand to run the same tests different ways. I did have some of that coded in, but ended up running the ginkgo's twice. For now I think this is a good smoke test.

stevesloka:
@skriss @sunjayBhatia I refactored this a bit to add some "deployment" tests. I'd like to write up some docs, but can follow that up with a new PR.

test/e2e/httpproxy/httpproxy_test.go
```diff
@@ -427,23 +427,6 @@ descriptors:
 	})
 })
 
-// httpProxyValid returns true if the proxy has a .status.currentStatus
```
sunjayBhatia, line 430: seems like some commits from a merge or rebase snuck in here?

sunjayBhatia: ah wait maybe nvm, i think i did the same thing in one of my PRs that I'm mixing up: https://github.com/projectcontour/contour/pull/4170/files#diff-85ebf4a94b7ed92e12d4ccea5e41be34ceeaf112df1ebba49de2fec7c297b780R461

.github/workflows/prbuild.yaml
```diff
@@ -1,7 +1,7 @@
 name: Build and Test Pull Request
 
 # Trigger the workflow on push or pull request
-on: 
+on:
```
sunjayBhatia, line 4: nit: seems like this is all whitespace changes we could maybe remove most of them?

test/e2e/deployment.go
```diff
@@ -486,8 +540,56 @@ func (d *Deployment) EnsureResourcesForLocalContour() error {
 		return err
 	}
 
+	if d.EnvoyDeploymentMode == DaemonsetMode {
```
sunjayBhatia, line 543: could this be a mutator func that takes a pointer to `d.EnvoyDaemonSet.Spec.Template` or `d.EnvoyDeployment.Spec.Template` and modifies it as needed? just so we don't have to duplicate this

test/e2e/deployment/deployment_test.go
```diff
@@ -0,0 +1,155 @@
...
+)
+
+var (
+	f = e2e.NewFramework(false, e2e.DeploymentMode)
+
+	// Functions called after suite to clean up resources.
+	cleanup []func()
+)
+
+func TestInfra(t *testing.T) {
```
sunjayBhatia, line 40: nit: naming on this

test/e2e/deployment/path_condition_match_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+import (
+	. "github.com/onsi/ginkgo"
+	contourv1 "github.com/projectcontour/contour/apis/projectcontour/v1"
+	"github.com/projectcontour/contour/test/e2e"
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func testPathConditionMatch(namespace string) {
```
sunjayBhatia, line 27: could we use exported test functions maybe instead of duplicating tests, so we can just call them from different suites?

stevesloka: Do you know a way to import across test packages?

sunjayBhatia: I think our e2e test packages aren't `_test` named so could work like normal?

stevesloka: I don't think the `_test.go` files are importable.

sunjayBhatia: ah, the filenames could change as well, or we could pull out a helper package maybe

sunjayBhatia:
ah @stevesloka just merged https://github.com/projectcontour/contour/pull/4170 and realized there will be merge conflicts now, my apologies

stevesloka:
No worries @sunjayBhatia, all rebased!

test/e2e/deployment/path_condition_match_test.go
skriss: If we wanted to get this in as-is, I'd be willing to take the follow-up to do some de-duplication here.

skriss (commented):
This looks pretty good, one small merge conflict to resolve. I do think de-duping the tests would be good, but maybe we can do that in a follow-up for the sake of getting this in.

test/e2e/framework.go
```diff
@@ -458,6 +460,22 @@ func UsingContourConfigCRD() bool {
...
+		got, err := strconv.ParseBool(useEnvoyDeployment)
+		if err != nil {
+			fmt.Printf("error trying to parse \"USE_ENVOY_DEPLOYMENT\" env var, defaulting to use Daemonset: %v\n", err)
+			return true
+		}
+		return !got
+	}
+	return true
+}
+
```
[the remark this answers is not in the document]
stevesloka: yup I think so, good catch.

stevesloka:
> If we wanted to get this in as-is, I'd be willing to take the follow-up to do some de-duplication here.

@skriss @sunjayBhatia I think that's a good plan, merge this in and we can refactor everything around maybe. What do you think?

skriss:
> > If we wanted to get this in as-is, I'd be willing to take the follow-up to do some de-duplication here.
> 
> @skriss @sunjayBhatia I think that's a good plan, merge this in and we can refactor everything around maybe. What do you think?

It works for me. I was thinking that maybe another path here would be to run the full suite of E2E's against a Deployment as part of the daily run, so we could make a decision on that and then make some subsequent revisions here.

skriss (approved):

stevesloka:
I'm inclined to just merge this, then I'll add it to the daily build to help validate. Just want to check in with you @sunjayBhatia if that makes sense.

sunjayBhatia:
> I'm inclined to just merge this, then I'll add it to the daily build to help validate. Just want to check in with you @sunjayBhatia if that makes sense.

sounds good to me 👍🏽 good to get changes in and then we can refactor the CI as needed

PR #4138, opened 2021-10-25
Gateway API: implement ReferencePolicy

### LINKED ISSUE #4118: Gateway API: implement ReferencePolicy
https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.ReferencePolicy

skriss:
Implements ReferencePolicy, which enables certain
cross-namespace references to be allowed. This
currently applies only to Route->Service refs.

Closes #4118.

### DIFF
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -21,7 +21,6 @@ import (
 	"strings"
 
 	"github.com/projectcontour/contour/internal/errors"
-	"github.com/projectcontour/contour/internal/gatewayapi"
 	"github.com/projectcontour/contour/internal/k8s"
 	"github.com/projectcontour/contour/internal/status"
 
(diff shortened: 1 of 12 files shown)


internal/dag/gatewayapi_processor.go
```diff
@@ -745,13 +746,58 @@ func (p *GatewayAPIProcessor) validateBackendRef(backendRef gatewayapi_v1alpha2.
...
+			}
+			if !fromAllowed {
+				continue
+			}
+
+			// "To" must contain an entry matching the Service that
+			// is being referenced.
+			var toAllowed bool
+			for _, to := range referencePolicy.Spec.To {
+				if (to.Group == "" || to.Group == "core") && to.Kind == "Service" && (to.Name == nil || to.Name == gatewayapi.ObjectNamePtr("") || to.Name == &backendRef.Name) {
```
sunjayBhatia, line 777: nit: only test cases I think might be missing are:
- when the reference policy has a name in the `To` section
- when there is a reference policy that would possible be valid but not in the right namespace

(just to get these pesky warnings about line coverage out of here as well)

sunjayBhatia (approved):

internal/dag/gatewayapi_processor.go
skriss: good call, got a little lazy here, also caught a pointer bug writing the additional test cases

skriss: gotta make some more fixes to that last commit still.

skriss: OK should be thoroughly covered now, thanks for calling this out.

PR #4143, opened 2021-10-28
Gateway API: set TCPRoute and UDPRoute status

### LINKED ISSUE #3441: Set Status on TCPRoute
It's been stated that `projectcontour/contour` won't support `TCPRoute` in gateway-api. Currently, Contour doesn't watch for those types of resources, but thinking maybe we should and set an invalid (or `Rejected`) status on those objects in the event someone does setup a TCPRoute with a Contour Gateway and expects something to happen, but nothing ever does. 

//ref: https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.TCPRoute

### LINKED ISSUE #3442: Set Status on UDPRoute
It's been stated that projectcontour/contour won't support UDPRoute in gateway-api. Currently, Contour doesn't watch for those types of resources, but thinking maybe we should and set an invalid (or Rejected) status on those objects in the event someone does setup a UDPRoute with a Contour Gateway and expects something to happen, but nothing ever does.

//ref: https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.UDPRoute

skriss:
Contour does not support Gateway API's
TCPRoute or UDPRoute. This PR sets a
condition of "Accepted: false" on all
TCPRoutes & UDPRoutes referencing a
Contour Gateway.

Closes #3441.
Closes #3442.

### DIFF
--- a/internal/controller/helpers.go
+++ b/internal/controller/helpers.go
@@ -1,0 +1,154 @@
+// Copyright Project Contour Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package controller
+
+import (
+	"context"
+
+	"github.com/projectcontour/contour/internal/gatewayapi"
+	"github.com/projectcontour/contour/internal/k8s"
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gatewayapi_v1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+)
+
+// setRouteCondition sets a condition on a Gateway API route
+// for a given gateway & controller. It performs an upsert,
+// updating the existing condition (if there has been a change),
+// otherwise adding a new condition.
+func setRouteCondition(
+	routeParentStatuses []gatewayapi_v1alpha2.RouteParentStatus,
+	condition metav1.Condition,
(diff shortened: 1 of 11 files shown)


youngnick:
Yeah, I think that upstream direction is suffering from overload of MAY - in this case it's pointing out that it's not always possible to guarantee that you'll see everything that you don't support. I agree that it could use clarification to say that if you have Routes attached to a Gateway whose controller doesn't support them, that controller must update the status to tell you so. I'll work on an upstream change for that.

internal/controller/helpers.go
```diff
@@ -0,0 +1,154 @@
...
+// specified name and namespace.
+func isRefToGateway(ref gatewayapi_v1alpha2.ParentRef, gateway types.NamespacedName) bool {
+	return ref.Group != nil && *ref.Group == gatewayapi_v1alpha2.GroupName &&
+		ref.Kind != nil && *ref.Kind == "Gateway" &&
+		ref.Namespace != nil && *ref.Namespace == gatewayapi_v1alpha2.Namespace(gateway.Namespace) &&
+		string(ref.Name) == gateway.Name
+}
+
+// referencesContourGateway returns whether a given list of
+// ParentRefs references a Gateway controlled by this Contour.
```
youngnick, line 89: Nit:
```suggestion
// referencesContourGateway returns the first Gateway from the
// list of provided ParentRefs that matches a Contour GatewayClass.
```

internal/controller/helpers.go
```diff
@@ -0,0 +1,154 @@
...
+func referencesContourGateway(
+	parentRefs []gatewayapi_v1alpha2.ParentRef,
+	routeNamespace string,
+	controllerName gatewayapi_v1alpha2.GatewayController,
+	kubeClient client.Client,
+	log logrus.FieldLogger,
+) (bool, *gatewayapi_v1alpha2.Gateway) {
```
youngnick, line 96: Nit: Feels like it would be more idiomatic to have this the other way round, so that we can use it like the other two value functions that return an `ok` variable?
```suggestion
) (b*gatewayapi_v1alpha2.Gateway, bool) {
```
would allow the caller to do:
```go
gw, ok := referencesContourGateway(...)
if !ok {
}
```

youngnick (approved):
This LGTM with a couple of small doc and style nits. Those are definitely non-blocking though.

internal/controller/helpers.go
skriss: :+1: done

internal/controller/helpers.go
skriss: :+1: done

Issue #4022, opened 2021-09-09
Add support for with_request_body


mfridman:
After setting up envoy, contour and implementing an authorization server we noticed this setup does not work for SAML-based auth flows.

Specifically, the request body is not sent to the authorization server. We'd like to keep the AuthProvider in the same process as the ExtAuth service.

There was an issue filed and implemented to support per path filtering:
https://github.com/envoyproxy/envoy/issues/13285
https://github.com/envoyproxy/envoy/pull/13389

The documentation here:
https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto.html?highlight=with_request_body#extensions-filters-http-ext-authz-v3-extauthz

There is a configuration example here:
https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/ext_authz_filter.html?highlight=with_request_body#configuration-examples

> One of the features of this filter is to send HTTP request body to the configured gRPC authorization server as part of the check request.

^ This is what we'd like to accomplish.

Is it possible to add support in contour to enable passing the request body to the authorization server? 

We rely on contour to manage the envoy configuration and would like to avoid setting filters in envoy directly through static files.

youngnick:
This seems like a reasonable ask, thanks for the issue @mfridman.

Do you think that a boolean would be okay to start with, or would you need to tune the additional parameters (`max_request_bytes`, `allow_partial_message`, `pack_as_bytes`) as well? I'd prefer to start with a boolean and default these unless there's a strong need to do otherwise?

mfridman:
Hi @youngnick, thank you for consideration. Yes, a boolean would work.

Assuming there is no limitation on the body size, we don't need those additional parameters. I mention this because SAML-based responses for ACS endpoints can be quite large, ~10KB.

mfridman:
If this is something trivial for the maintainers to add that'd be awesome, otherwise I could take a look at submitting a PR. 

I presume this is a common task (syncing _some_ envoy options to be exposed via contour) .. maybe there is an existing PR you could point me at? If its a fairly trivial chunk of work I could submit a PR.

abhide:
@youngnick I would like to work on adding support for this. Would that be fine? 
Seems like we can add options to 
`type AuthorizationServer struct` under `apis/projectcontour/v1/httpproxy.go`

abhide:
If we add a boolean flag to what would be defaults for 
```code
      with_request_body:
[...]

PR #4148, opened 2021-11-02
Add support for with_request_body to ExtAuthorizationServer

abhide:
ExtAuthzFilter supports adding buffer settings that enable
the filter to buffer the client request body and
send it within the authorization request.

This change adds support to set these buffer settings for HTTPProxy

Fixes: #4022


internal/dag/dag.go
```diff
@@ -585,6 +585,18 @@ type SecureVirtualHost struct {
 	// only reason to set this to `true` is when you are migrating
 	// from internal to external authorization.
 	AuthorizationFailOpen bool
+
+	// AuthorizationBufferSettingsEnabled sets whether buffer settings are set for buffering request data sent to AuthorizationServer
+	AuthorizationBufferSettingsEnabled bool
+
+	// AuthorizationBufferSettingsMaxRequestBytes sets the maximum size of message body ExtAuthz filter will hold in-memory.
+	AuthorizationBufferSettingsMaxRequestBytes uint32
```
abhide, line 593: Sorry, I am terrible at picking names! Naming is hard! 

@skriss please suggest! :smiley:

abhide:
@skriss / @sunjayBhatia Please take a look. 
If the change looks good, I will add unit tests.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -171,6 +171,25 @@ type AuthorizationServer struct {
...
+	// BufferSettings specifies configuration for buffering request data sent to AuthorizationServer
+	// +optional
+	BufferSettings *AuthorizationServerBufferSettings `json:"bufferSettings,omitEmpty"`
+}
+
+// AuthorizationServerBufferSettings enables ExtAuthz filter to buffer client request data and send it as part of authorization request
+type AuthorizationServerBufferSettings struct {
+	// MaxRequestBytes sets the maximum size of message body ExtAuthz filter will hold in-memory.
+	// +optional
+	MaxRequestBytes uint32 `json:"maxRequestBytes,omitempty"`
```
abhide, line 184: I will have to add validation to ensure this is greater than 0. 
Similar to validation in protobuf: https://github.com/envoyproxy/envoy/blob/dbb1f50ef340e565c5a495a17aac0afff5d9a4f9/api/envoy/extensions/filters/http/ext_authz/v3/ext_authz.proto#L158

Should I add it to internal/dag/httpproxy_processor.go ?

internal/dag/dag.go
```diff
@@ -585,6 +585,18 @@ type SecureVirtualHost struct {
 	// only reason to set this to `true` is when you are migrating
 	// from internal to external authorization.
 	AuthorizationFailOpen bool
+
+	// AuthorizationBufferSettingsEnabled sets whether buffer settings are set for buffering request data sent to AuthorizationServer
+	AuthorizationBufferSettingsEnabled bool
```
sunjayBhatia, line 590: maybe wrap these fields in a struct and use a pointer to that struct in this object, can remove this boolean field with that

internal/xdscache/v3/listener.go
```diff
@@ -598,11 +598,23 @@ func (v *listenerVisitor) visit(vertex dag.Vertex) {
 			var authFilter *http.HttpFilter
 
 			if vh.AuthorizationService != nil {
-				authFilter = envoy_v3.FilterExternalAuthz(
-					vh.AuthorizationService.Name,
-					vh.AuthorizationFailOpen,
-					vh.AuthorizationResponseTimeout,
-				)
+				if vh.AuthorizationBufferSettingsEnabled {
```
sunjayBhatia, line 601: if we use a struct pointer to wrap the buffer settings, we could remove the `if` and pass the buffer settings in always and just do a nil check in `FilterExternalAuthz` on whether or not to add buffer settings

apis/projectcontour/v1/httpproxy.go
sunjayBhatia: we can use some `kubebuilder` validation here too to set defaults and minimum values

internal/dag/dag.go
abhide: Done

internal/dag/dag.go
abhide: Moving it to a separate struct makes naming easier and obvious.

internal/xdscache/v3/listener.go
abhide: done

apis/projectcontour/v1/httpproxy.go
abhide: Thanks @sunjayBhatia . Added kubebuilder validation and set minimum to be 1

abhide:
@sunjayBhatia Thanks for your review comments. Addressed your comments. 

Will be great if you could please take another look. If this looks good, I will add unit tests.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -171,6 +171,26 @@ type AuthorizationServer struct {
...
+	// BufferSettings specifies configuration for buffering request data sent to AuthorizationServer
+	// +optional
+	BufferSettings *AuthorizationServerBufferSettings `json:"bufferSettings,omitempty"`
+}
+
+// AuthorizationServerBufferSettings enables ExtAuthz filter to buffer client request data and send it as part of authorization request
+type AuthorizationServerBufferSettings struct {
+	// MaxRequestBytes sets the maximum size of message body ExtAuthz filter will hold in-memory.
+	// +optional
+	// ++kubebuilder:validation:Minimum=1
```
sunjayBhatia, line 184:
```suggestion
	// +kubebuilder:validation:Minimum=1
```

will need a regenerate of the crds as well

internal/envoy/v3/listener.go
```diff
@@ -668,6 +666,21 @@ func FilterExternalAuthz(authzClusterName string, failOpen bool, timeout timeout
 		// `transport_api_version` from ExtensionServiceSpec ProtocolVersion.
 		TransportApiVersion: envoy_core_v3.ApiVersion_V3,
 	}
+}
+
+// FilterExternalAuthz returns an `ext_authz` filter configured with the
+// requested parameters.
+func FilterExternalAuthz(authzClusterName string, failOpen bool, timeout timeout.Setting, bufferSettings *dag.AuthorizationServerBufferSettings) *http.HttpFilter {
+	authConfig := getExtAuthzConfig(authzClusterName, failOpen, timeout)
```
sunjayBhatia, line 674: do we need this extra helper?

internal/dag/httpproxy_processor.go
```diff
@@ -297,6 +297,14 @@ func (p *HTTPProxyProcessor) computeHTTPProxy(proxy *contour_api_v1.HTTPProxy) {
 				} else {
 					svhost.AuthorizationResponseTimeout = timeout
 				}
+
+				if auth.BufferSettings != nil {
+					svhost.AuthorizationServerBufferSettings = &AuthorizationServerBufferSettings{
+						MaxRequestBytes:     auth.BufferSettings.MaxRequestBytes,
```
sunjayBhatia, line 303: unfortunately we can't just pass this straight through, in the case that someone makes an httpproxy that looks like:

```
  virtualhost:
    authorization:
      authPolicy:
        context:
          hostname: externalauth.projectcontour.io
      bufferSettings: {}
```

we get `0` for this value since it is unset, and Envoy returns the following xDS error:

```
time="2021-11-15T16:46:46Z" level=error msg="Error adding/updating listener(s) ingress_https: Proto constraint validation failed (ExtAuthzValidationError.WithRequestBody: embedded message failed validation | caused by BufferSettingsValidationError.MaxRequestBytes: value must be greater than 0): grpc_service {\n  envoy_grpc {\n    cluster_name: \"extension/httpproxy-external-auth/testserver\"\n  }\n  timeout {\n    nanos: 500000000\n  }\n}\nwith_request_body {\n}\nclear_route_cache: true\nstatus_on_error {\n  code: Forbidden\n}\ninclude_peer_certificate: true\ntransport_api_version: V3\n\n" code=13 connection=2 context=xds node_id=envoy-k7l9l node_version=v1.19.1 response_nonce=4 version_info=3
```

we may have to set a default in kubebuilder and/or validate the field is not in fact `0` in here and set a default size

sunjayBhatia: When the field is unsetin YAML it will still come through as 0, in the example above you can see that if someone uses {} for the buffersettings

youngnick: Yes, I think the way to think this through is: what happens when someone supplies the absolute minimum of configuration? Does the combination of just-the-required-fields and zero values for optional fields work?

internal/dag/dag.go
```diff
@@ -532,6 +532,22 @@ type SecureVirtualHost struct {
...
 	AuthorizationFailOpen bool
+
+	// BufferSettings specifies configuration for buffering request data sent to AuthorizationServer
+	AuthorizationServerBufferSettings *AuthorizationServerBufferSettings
+}
+
+// AuthorizationServerBufferSettings enables ExtAuthz filter to buffer client
+// request data and send it as part of authorization request
+type AuthorizationServerBufferSettings struct {
+	// MaxRequestBytes sets the maximum size of message body ExtAuthz filter will hold in-memory.
```
youngnick, line 543: I think a little explanation of this value would be really helpful for people sizing it. What happens when this is too small? (Sounds like the body will not be sent unless you set `AllowPartialMessage` to `true`?) What happens when it's too big (I would guess that you'll use a lot of memory).

youngnick (commented):
Sorry to be slow on review here, I think the implementation is mostly fine, I just have some queries around the settings. I'd encourage you to walk through how setting up this config will feel to a user. How will they know what values they should pick for the optional fields? If the leave the optional fields blank, what will happen? And so on.

abhide:
Thanks @youngnick for your review comments. Will address those.

internal/dag/httpproxy_processor.go
abhide: @sunjayBhatia Thanks for the clarification. Added` validation check. 

The other 2 fields are boolean; so we just need to take care of validating and setting default for maxBytes.

apis/projectcontour/v1/httpproxy.go
skriss: I think we want a *minimum* of 1, and a *default* of e.g. 1024:

```
// +kubebuilder:default=1024

```

apis/projectcontour/v1/httpproxy.go
```diff
@@ -171,6 +171,26 @@ type AuthorizationServer struct {
 	//
 	// +optional
 	FailOpen bool `json:"failOpen,omitempty"`
+
+	// BufferSettings specifies configuration for buffering request data sent to AuthorizationServer
+	// +optional
+	BufferSettings *AuthorizationServerBufferSettings `json:"bufferSettings,omitempty"`
```
skriss, line 177: interested in what others think, but I'd prefer a field name more like `IncludeRequestBody` which I think is more obvious to users, e.g.:

```yaml
authorizationServer:
  includeRequestBody:
    maxRequestBytes: 1024
    allowPartialMessage: true
...
```

internal/dag/dag.go
skriss: Perhaps a brief explanation that doesn't require jumping to Envoy docs, plus a link for more details.

internal/dag/dag.go
```diff
@@ -532,6 +532,22 @@ type SecureVirtualHost struct {
 	// only reason to set this to `true` is when you are migrating
 	// from internal to external authorization.
 	AuthorizationFailOpen bool
+
+	// BufferSettings specifies configuration for buffering request data sent to AuthorizationServer
+	AuthorizationServerBufferSettings *AuthorizationServerBufferSettings
```
skriss, line 537: If we rename the API field per my previous comment, then a similar rename here would make sense for consistency.

skriss (changes requested):
This looks pretty good @abhide! Couple small comments.

skriss:
You'll also need to add a changelog file, `changelogs/unreleased/4148-abhide-small.md`, that describes the change from a user perspective -- see others for examples of formatting/content.

internal/dag/httpproxy_processor.go
```diff
@@ -297,6 +297,18 @@ func (p *HTTPProxyProcessor) computeHTTPProxy(proxy *contour_api_v1.HTTPProxy) {
 				} else {
 					svhost.AuthorizationResponseTimeout = timeout
 				}
+
+				if auth.BufferSettings != nil {
+					var maxRequestBytes uint32 = 1024
```
sunjayBhatia, line 302: note that this logic adds a default in the processing code, in addition to a default we may add in the kubebuilder defaults, just have to keep in mind in case we change the default

apis/projectcontour/v1/httpproxy.go
youngnick: I think that's much clearer in YAML form, great idea.

internal/dag/httpproxy_processor.go
youngnick: Yes, if we're adding a default, it should be in the kubebuilder settings, and if we do have a default in processing, it should be in a constant or something somewhere, so there's an obvious place to change it.

apis/projectcontour/v1/httpproxy.go
abhide: yes.. renaming makes sense. 
Done

internal/dag/dag.go
abhide: Done

apis/projectcontour/v1/httpproxy.go
abhide: Thanks! Fixed

internal/dag/httpproxy_processor.go
abhide: Added a const to this file. 
I didn't see a common place to add constants.

internal/dag/dag.go
abhide: done.

abhide:
Thanks for reviewing @skriss . 

Addressed review comments. Please take a look and let me know. 
I have also added changelog file as suggested.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -171,6 +171,27 @@ type AuthorizationServer struct {
 	//
 	// +optional
 	FailOpen bool `json:"failOpen,omitempty"`
+
+	// BufferSettings specifies configuration for buffering request data sent to AuthorizationServer
```
skriss, line 175:
```suggestion
	// WithRequestBody specifies configuration for sending the client request's body to AuthorizationServer
```

changelogs/unreleased/4148-abhide-small.md
```diff
@@ -0,0 +1 @@
+Add support for with_request_body to ExtAuthorizationServer. See [External Authorization Configuration Overview](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto#envoy-v3-api-msg-extensions-filters-http-ext-authz-v3-extauthz) for more information.
```
skriss, line 1:
```suggestion
Adds a new field, `withRequestBody`, within HTTPProxy's `authorizationServer` to allow client request bodies to be sent to the external authorization server. See [External Authorization Configuration Overview](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto#envoy-v3-api-msg-extensions-filters-http-ext-authz-v3-extauthz) for more information.
```

skriss (changes requested):
Thanks for the updates @abhide, just two more edits and this looks good to me

apis/projectcontour/v1/httpproxy.go
abhide: done

changelogs/unreleased/4148-abhide-small.md
abhide: done

abhide:
Thanks @skriss. Fixed your latest comments.

skriss (approved):
LGTM, thanks @abhide!

sunjayBhatia (approved):

PR #4149, opened 2021-11-02
internal/dag: move ListenerName to Listener struct

skriss:
Moves the ListenerName field from VirtualHost
to Listener, to better reflect the data model.
This field is now populated by the listener
processor.

### DIFF
--- a/internal/dag/listener_processor.go
+++ b/internal/dag/listener_processor.go
@@ -15,6 +15,12 @@ package dag
 
 import "sort"
 
+// nolint:revive
+const (
+	HTTP_LISTENER_NAME  = "ingress_http"
+	HTTPS_LISTENER_NAME = "ingress_https"
+)
+
 // ListenerProcessor adds an HTTP and an HTTPS listener to
 // the DAG if there are virtual hosts and secure virtual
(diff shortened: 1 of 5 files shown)


internal/dag/listener_processor.go
```diff
@@ -15,6 +15,12 @@ package dag
 
 import "sort"
 
+// nolint:revive
+const (
+	HTTP_LISTENER_NAME  = "ingress_http"
+	HTTPS_LISTENER_NAME = "ingress_https"
+)
```
skriss, line 22: I wasn't quite sure if it made sense to have these + the constants currently defined in internal/xdscache/v3 be consolidated (meaning the DAG listener names always equal the Envoy listener names), so for now I left them separate. Open to input here.

youngnick: I think this is fine for now. As @stevesloka said in another comment, if we do multiple listeners, all of this is going to need to change anyway.

youngnick (approved):
This LGTM, I think it gives us a place to build from for multiple listeners later.

stevesloka (approved):

PR #4157, opened 2021-11-04
WIP: Moving to controller-runtime leader election and runnables

sunjayBhatia:
- adds flags instead of contour config crd fields for leader election
  - logic behind this is that we would exit anyway if we want to change
leader election config, not much use in it being dynamically
configurable
  - also just removes from v1alpha1 ContourConfigurationCRD, but that
might be not allowed, so might need to do v1alpha2
- Gets rid of our leader election wiring to use controller-runtime
manager leader election
- Transitions to use configmaps/leases multilock so we can transition to
solely leases in the future
- wraps controller-runtime controllers to ensure they are run even if
not leader
- e2e tests (and manual test) works to reconcile gateways if you don't
specify --disable-leader-election flag

Next step is to turn all other components into controller runtime
runnables *and* add more testing of wiring bits and controllers


youngnick:
Only did a very quick look, but I really like this approach. I think it will go quite a way to solving some of our process-startup issues.

cmd/contour/serve.go
```diff
@@ -193,9 +199,26 @@ func NewServer(log logrus.FieldLogger, ctx *serveContext) (*Server, error) {
 	}
 
 	// Instantiate a controller-runtime manager.
-	mgr, err := manager.New(restConfig, manager.Options{
+	options := manager.Options{
 		Scheme: scheme,
-	})
+	}
+	if !ctx.DisableLeaderElection {
```
stevesloka, line 205: nit: Just a thought, maybe handle the positive case and log that leader election is disabled? Then the else can set the leader election config to reduce the stutter with not-disable-leader-election?

```suggestion
	if ctx.DisableLeaderElection {
	 <log disabled>
	 options.LeaderElection = false
	} else {
	 options.LeaderElection = true
	 ...	
	}
```

skriss (commented):
👍  this looks promising! As you say, we need to think a little more about the change to flags for configuring leader election, as well as the configmap -> lease migration path to make sure that users can upgrade smoothly, but I think they both make sense to do if we can.

cmd/contour/serve.go
sunjayBhatia: ah yeah logging leader election is disabled would be good, the leader election code in controller runtime does some logging so we could rely on that or also add a log for the enabled case too if we want

cmd/contour/serve.go
```diff
@@ -361,6 +386,7 @@ func (s *Server) doServe() error {
 			fallbackCert:              fallbackCert,
 		}),
 		FieldLogger: s.log.WithField("context", "contourEventHandler"),
+		IsLeader:    s.mgr.Elected(),
```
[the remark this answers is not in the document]
sunjayBhatia: should do!

https://github.com/kubernetes-sigs/controller-runtime/blob/6c7c1d73fc5ed965c2962d6f4552b15f9841ce86/pkg/manager/manager.go#L63-L66

sunjayBhatia:
> 👍 this looks promising! As you say, we need to think a little more about the change to flags for configuring leader election, as well as the configmap -> lease migration path to make sure that users can upgrade smoothly, but I think they both make sense to do if we can.

I misread the controller-runtime docs, i guess you can use `configmapleases` as a multilock on configmaps and leases as a built in feature! we could do this now or in future releases

sunjayBhatia:
superceded by https://github.com/projectcontour/contour/pull/4202, closing

PR #4159, opened 2021-11-05
Fix examples and install-contour-working script

sunjayBhatia:
Useful for local dev and manual testing

### DIFF
--- a/examples/gateway/kuard/kuard.yaml
+++ b/examples/gateway/kuard/kuard.yaml
@@ -44,6 +44,10 @@ metadata:
   labels:
     app: kuard
 spec:
+  parentRefs:
+  - group: gateway.networking.k8s.io
+    kind: Gateway
+    name: contour
   hostnames:
   - "local.projectcontour.io"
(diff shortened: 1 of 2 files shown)


stevesloka (commented):
The HTTPRoute also needs the `spec.parentRefs` as well.

examples/gateway/kuard/kuard.yaml
```diff
@@ -37,20 +37,21 @@ spec:
   type: ClusterIP
 ---
 kind: HTTPRoute
-apiVersion: networking.x-k8s.io/v1alpha1
+apiVersion: gateway.networking.k8s.io/v1alpha2
```
skriss, line 40: I have this change in https://github.com/projectcontour/contour/pull/4122 but had been holding off until closer to release so as to not break the existing guide against our latest released version. Maybe create a `-va1lpha2` version for now?

skriss (approved):
LGTM

youngnick (approved):

Issue #4124, opened 2021-10-20
Gateway API: implement ListenerStatus for Gateways


skriss:
ref. https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.GatewayStatus

We need to populate the `gateway.status.listeners` field with supported kinds, # of attached routes, listener-specific conditions, etc.

skriss:
Working on this now.

youngnick:
Will this include changes to the operator to pull the code there out?

skriss:
I'll double-check but I'm pretty sure there's nothing in the operator that sets any gateway status anymore (and I don't think it had ever gotten to setting listener status specifically).

youngnick:
Awesome, thanks. I think it's particularly important that when we are ready to cut 1.20-rc1, we have a working flow for the Operator to do Gateway API.

skriss:
ref. 3529

PR #4160, opened 2021-11-05
internal/dag: set Gateway Listener status

skriss:
Sets status for Gateway Listeners, specifically
the "attachedRoutes" and "supportedKinds" fields.

Updates #4124.

### DIFF
--- a/internal/status/gatewaystatus.go
+++ b/internal/status/gatewaystatus.go
@@ -29,7 +29,9 @@ const ReasonInvalidGateway = "Invalid"
 
 const MessageValidGateway = "Valid Gateway"
 
-type GatewayConditionsUpdate struct {
+// GatewayStatusUpdate represents an atomic update to a
+// Gateway's status.
+type GatewayStatusUpdate struct {
 	FullName           types.NamespacedName
 	Conditions         map[gatewayapi_v1alpha2.GatewayConditionType]metav1.Condition
 	ExistingConditions map[gatewayapi_v1alpha2.GatewayConditionType]metav1.Condition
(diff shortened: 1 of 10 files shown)


internal/status/gatewaystatus.go
```diff
@@ -93,7 +108,15 @@ func getGatewayConditions(gs *gatewayapi_v1alpha2.GatewayStatus) map[gatewayapi_
 	return conditions
 }
 
-func (gatewayUpdate *GatewayConditionsUpdate) Mutate(obj client.Object) client.Object {
+func (gatewayUpdate *GatewayStatusUpdate) Mutate(obj client.Object) client.Object {
+	o, ok := obj.(*gatewayapi_v1alpha2.Gateway)
+	if !ok {
+		panic(fmt.Sprintf("Unsupported %T object %s/%s in GatewayStatusUpdate status mutator",
```
stevesloka, line 114: I commented when this was first added, but I still dislike the `panic(..)` here. I don't feel that status failing to update should cause Contour to hard crash. :man_shrugging:

stevesloka (approved):

internal/status/gatewaystatus.go
skriss: I have no objection to switching the Mutate funcs to return an error instead of panicking. In reality this code branch really _shouldn't_ ever get hit in a released version of Contour, it would require there to be a bug in the status code that would cause other tests to fail too, but never say never.

skriss: Or we can wait until Go 1.18 and use generics :trollface:

youngnick: Yeah, the reason I put the panic in there is because there's already a few layers of defense against unsupported objects getting into a mutator. It should never happen, and in the unlikely event that it does, we want it to break tests or something.

youngnick (approved):

Issue #4163, opened 2021-11-05
deadlock on shutdown


skriss:
https://github.com/projectcontour/contour/runs/4113809841?check_suite_focus=true#step:6:790
https://github.com/projectcontour/contour/runs/4120027004?check_suite_focus=true#step:6:685

Hypothesis is that the sigterm handler hasn't actually started by the time Ginkgo sends a sigterm to shut down Contour, so it's never handled and the test times out waiting for Contour to shut down.

tsaarni:
I also bumped into this previously. It seemed to handle the signal

```
    time="2021-11-05T05:31:00Z" level=info msg="shutting down" context=sigterm-handler signal=terminated
    time="2021-11-05T05:31:00Z" level=error msg="terminated HTTP server with error" context=metricsvc error="http: Server closed"
    time="2021-11-05T05:31:00Z" level=info msg="stopped event handler" context=contourEventHandler
    time="2021-11-05T05:31:00Z" level=error msg="terminated HTTP server with error" context=debugsvc error="http: Server closed"
```

but still it took the full 60 seconds after that...

tsaarni:
There is difference in successful and unsuccessful shutdown:

When shutdown is successful, xDS server goroutine has reached this far (see successful log [here](https://github.com/projectcontour/contour/runs/4120027004?check_suite_focus=true#step:6:142)) https://github.com/projectcontour/contour/blob/cb3ec9aee1ea74ac30b8823ab0b3b03588dfb3d8/cmd/contour/serve.go#L571

In case of unsuccessful shutdown it seems like that goroutine is stuck here https://github.com/projectcontour/contour/blob/cb3ec9aee1ea74ac30b8823ab0b3b03588dfb3d8/cmd/contour/serve.go#L567-L568

Why would that get stuck?

If that is the case, then the `WaitGroup` is stuck forever and shutdown does not proceed, since one of the goroutines never calls `Done()` https://github.com/projectcontour/contour/blob/cb3ec9aee1ea74ac30b8823ab0b3b03588dfb3d8/internal/workgroup/group.go#L92-L99

skriss:
:+1: agreed, that looks suspicious. Haven't been able to spot the bug there yet.

One other suspect thing I'm wondering about: https://github.com/projectcontour/contour/blob/main/cmd/contour/serve.go#L250-L253

[...]

PR #4164, opened 2021-11-08
test/e2e: simplify API validation tests

skriss:
Moves E2E tests that verify kubebuilder validation
annotations to their own Describe block which
does not run a Contour instance, since it's not
needed for these tests.

Updates #4163.


sunjayBhatia (approved):
this test could be moved into the new block as well: https://github.com/projectcontour/contour/blob/cb3ec9aee1ea74ac30b8823ab0b3b03588dfb3d8/test/e2e/httpproxy/httpproxy_test.go#L337

stevesloka (approved):

Issue #3889, opened 2021-07-14
Validation of Upstream TLS and Client Certificate with kubernetes.io/tls Secret


alex1989hu:
**What steps did you take and what happened:**
Updated already existing and valid `HTTPProxy` with `validation` object to try out feature `Validation of Upstream TLS` . 

The referenced Secret (`caSecret: backend-certificate`) was issued by `cert-manager.io` instance running in Kubernetes cluster. 

The Secret Type is `kubernetes.io/tls` and placed into the same Namespace where `HTTPProxy` is also configured .

The Secret exists:
```sh
kubectl -n dashboard get secrets backend-certificate
NAME                  TYPE                DATA   AGE
backend-certificate   kubernetes.io/tls   3      32h
# See content below
Data
====
ca.crt:   2390 bytes
tls.crt:  1976 bytes
tls.key:  1675 bytes
```

The `HTTProxy` is invalid:
```sh
 kubectl -n dashboard get httpproxies.projectcontour.io dashboard-proxy
NAME              FQDN                         TLS SECRET                              STATUS    STATUS DESCRIPTION
dashboard-proxy   dashboard.health.acme   projectcontour/wildcard-health-cert   invalid   At least one error present, see Errors for details
```

`HTTPProxy` become invalid, here is the error message:
> Service [dashboard-backend-svc:8443] TLS upstream validation policy error: invalid CA Secret "dashboard/backend-certificate": Secret not found

**What did you expect to happen:**
Got a valid `HTTPProxy` due to the fact that the referenced Secret contains `ca.crt`, not doing type checking: `kubernetes.io/tls` vs `Opaque`.

It is not clear why the referenced Secret must be of type `Opaque`.

Reference:
- [Upstream Validation](https://projectcontour.io/docs/v1.17.0/config/upstream-tls/)
- [Client Certificate Validation](https://projectcontour.io/docs/v1.17.0/config/tls-termination/#client-certificate-validation)

**Anything else you would like to add:**
I found some old issues, it is still ambiguous for me why we need `Opaque` Secret:  https://github.com/projectcontour/contour/pull/1714

**Environment:**

- Contour version:
> 1.17.0
- Kubernetes version: (use `kubectl version`):
[...]

Issue #3496, opened 2021-03-18
TLS Error


sanketshirode:
I'm facing issue when configuring certificate in httpproxy. Below are the details of the error
    Errors:
      Message:             Spec.VirtualHost.TLS Secret "vc-staging-cert" is invalid: Secret not found
      Reason:              SecretNotValid
      Status:              True
      Type:                TLSError
      
   But the secret exists in the same namespace as proxy , also I tried tls delegation but same error. Please check below screenshots for reference:
   
![image](https://user-images.githubusercontent.com/30955166/111574505-a0777f80-87d2-11eb-9f51-0d867edf3fa1.png)
![image](https://user-images.githubusercontent.com/30955166/111574568-bbe28a80-87d2-11eb-99c7-42d9eb8d557c.png)
![image](https://user-images.githubusercontent.com/30955166/111575956-825f4e80-87d5-11eb-8793-b1803a7d44bb.png)

stevesloka:
Hey @sanketshirode, it's possible that there is a validation error happening and the "secret not found" is misleading. 

Would you be able to tail your Contour logs to see if there's something that pops up? Contour looks for the following in a secret:

- TLSCertKey: `tls.crt`
- TLSPrivateKeyKey: `tls.key`
- CommonName
- SubjectAltNames

If any of those are invalid the secret doesn't get added to Contour's cache, so later on when your HTTPProxy references the secert, it's "not found" since it's not in the cache.

sanketshirode:
Hey @stevesloka thanks a lot for responding. Please take a look at my secret given here 
![image](https://user-images.githubusercontent.com/30955166/111731702-0418b000-889a-11eb-802d-73af47a3516e.png)

Nothing I could find in the contour logs, which can relate to this error. Please let me know anything else I could check ?

sanketshirode:
Hey @stevesloka you were right,  now I can see the logs complaining about missing CN or Sub Alt Name but when I checked both are present in cert. Please check the attached screen shot

![image](https://user-images.githubusercontent.com/30955166/111797629-5506c380-88ef-11eb-9b08-629d2418113d.png)
![image](https://user-images.githubusercontent.com/30955166/111797753-723b9200-88ef-11eb-8f14-6a92337861db.png)

And this is the log complaining about cert
 
![image](https://user-images.githubusercontent.com/30955166/111797973-a2833080-88ef-11eb-85a4-8780075fa291.png)

sanketshirode:
[...]

Issue #2372, opened 2020-03-23
Certificate validation fails when certificate in chain is missing CN


nowylie:
**What steps did you take and what happened:**

Contour failed to validate our full chain certificate when the last certificate in the chain was missing a CN. The error in the Contour logs was:
```
time="2020-03-20T06:39:20Z" level=error msg="invalid TLS certificate: certificate has no common name or subject alt name" context=KubernetesCache kind=Secret name=... namespace=... version=v1
```

Specifically, Contour doesn't appear to like the root certificate in the **GoDaddy Certificate Bundles - G2 With Cross to G1, includes Root** (`gd_bundle-g2-g1.crt`) bundle found [here](https://ssl-ccp.godaddy.com/repository).

The result was that no HTTPS endpoint was configured for the specified ingress.

**What did you expect to happen:**

An HTTPS endpoint for the Ingress.

**Anything else you would like to add:**

*N/A*

**Environment:**

- Contour version: `1.2.1`
- Kubernetes version: (use `kubectl version`): 
   ```
   Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"archive", BuildDate:"2020-02-29T16:37:45Z", GoVersion:"go1.14", Compiler:"gc", Platform:"linux/amd64"}
   Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.8", GitCommit:"ec6eb119b81be488b030e849b9e64fda4caaf33c", GitTreeState:"clean", BuildDate:"2020-03-12T20:52:22Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
   ```
- Kubernetes installer & version:
   ```
   kubeadm version: &version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.8", GitCommit:"ec6eb119b81be488b030e849b9e64fda4caaf33c", GitTreeState:"clean", BuildDate:"2020-03-12T20:57:57Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
   ```
- Cloud provider or hardware configuration: `baremetal`
- OS (e.g. from `/etc/os-release`): `Debian GNU/Linux 9.9 (stretch)`

stevesloka:
Contour has some validation to make sure that referenced certs are valid. Does your generated cert have a CN or SubjectAltName?

[...]

PR #4165, opened 2021-11-08
Update Certificate Validation with more clarity and tests

youngnick:
This commit updates the Secret validation framework to address bugs
around CA certificates, certificates with no CN, and various combinations
of the above.

Also updates documentation to make the rules for the types of Secrets
clearer.

Updates #2372
Updates #3889
Updates #3496

Future work is to add some of these checks into the e2e tests, and ensure
that all these configs produce valid Envoy config, with no NACKs.


site/content/docs/main/config/tls-termination.md
```diff
@@ -13,7 +13,17 @@ Contour also follows a "secure first" approach.
...
+- be a Secret of type `kubernetes.io/tls`. This means that it must contain keys named `tls.crt` and `tls.key` that contain the certificate and private key to use for TLS, in PEM format.
+
+The TLS secret may also:
+- add any additional certificates required for validation into the `tls.crt` PEM bundle. If this is the case, the serving certificate must be the first certificate in the bundle.
+- contain a `ca.crt` key that contains a certificate bundle to use for CA validation.
+
+We recommend using only one of these options.
+
+Practically, this means that, if you have been issued a certificate bundle from a CA that contains the certificate plus intermediate certificates, you can put the contents of that bundle into `tls.crt`, and everything will work.
+If you have a separate file with an intermediate certificate chain, you may put those details into a `ca.crt` key, and everything will work.
```
tsaarni, line 26: Maybe consider reformulating, something along the lines of: 

```suggestion
- add any chain CA certificates required for validation into the `tls.crt` PEM bundle. If this is the case, the serving certificate must be the first certificate in the bundle and the intermediate CA certificates must be appended in issuing order.
```

I think everything else could be removed, because AFAIK `ca.crt` is not used currently and I would not change that. I believe it is correct behavior. It is pretty popular convention to have chain certs appended and not as separate CA bundle. Envoy also expects the server certificate and its chain in field `CertificateChain` in the SDS secret ([doc link](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto.html?highlight=tls#extensions-transport-sockets-tls-v3-tlscertificate)) and currently we just map `tls.crt` contents into `PrivateKey` and `CertificateChain` fields. We do not additionally append the contents of `ca.crt` to CertificateChain.

tsaarni (changes requested):
The actual change looks good to me but I have few comments on the documentation part.

youngnick:
Thanks for the detailed comments @tsaarni, you're totally right.

site/content/docs/main/config/tls-termination.md
```diff
@@ -13,7 +13,12 @@ Contour also follows a "secure first" approach.
 When TLS is enabled for a virtual host, any request to the insecure port is redirected to the secure interface with a 301 redirect.
 Specific routes can be configured to override this behavior and handle insecure requests by enabling the `spec.routes.permitInsecure` parameter on a Route.
 
-The TLS secret must contain keys named tls.crt and tls.key that contain the certificate and private key to use for TLS, e.g.:
+The TLS secret must:
+- be a Secret of type `kubernetes.io/tls`. This means that it must contain keys named `tls.crt` and `tls.key` that contain the certificate and private key to use for TLS, in PEM format.
+
+The TLS secret may also:
+- add any additional certificates required for validation into the `tls.crt` PEM bundle. If this is the case, the serving certificate must be the first certificate in the bundle.
```
tsaarni, line 20: Shouldn't the first bullet be removed? The additional certs in `tls.crt` should be **only** the chain CA certs which are there to support the peer when it is validating of our serving cert. Sending other certs would be redundant.

youngnick: Yep, I missed this, thanks.

tsaarni (approved):
LGTM

sunjayBhatia (approved):

PR #4167, opened 2021-11-09
Fix typo

lewislbr:
The quickstart has three steps:

1. Set up a kubernetes environment
2. Install a Contour service
3. Install a kuard workload

On step 1,  there is the option to skip to "Install a Contour service", which is the step 2, but is referred to as 3.


youngnick (approved):
Thanks for picking this up @lewislbr!

Issue #3854, opened 2021-06-30
test/e2e: Incluster test suite


sunjayBhatia:
Now that the e2e tests no longer run Contour in a k8s cluster, we have actually lost a bit of coverage. We need to add an in-cluster test suite to make sure that Contour and our example deployment work actually in a cluster.

This suite should be focused on validating features of Contour and/or the example deployment that are only coverable in an actual k8s cluster, e.g. validating RBAC

- [ ] Initial suite setup: https://github.com/projectcontour/contour/pull/3851
- [ ]

PR #4170, opened 2021-11-10
test/e2e: Add incluster test suite

sunjayBhatia:
Adds test for incluster behavior we should be verifying:
- simple smoke test
- rbac tests
- leader election test
  - to ensure when we make changes to leader election, we have proper
coverage and don't break anything

Future tests for behavior that rely on in-cluster setup can be added as
well, e.g. tests around Contour<->Envoy connections, shutdown manager,
etc.

Updates: https://github.com/projectcontour/contour/issues/3854


test/e2e/incluster/leaderelection_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+	. "github.com/onsi/ginkgo"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func testLeaderElection(namespace string) {
+	// This test is solely a check on the fact that we have set up leader
```
sunjayBhatia, line 35: for reviewers, if you have any thoughts about how to do this, couldn't figure out a good way to observe which instance sets status on a resource

test/e2e/incluster/rbac_test.go
```diff
@@ -0,0 +1,185 @@
...
+	contourv1 "github.com/projectcontour/contour/apis/projectcontour/v1"
+	contourv1alpha1 "github.com/projectcontour/contour/apis/projectcontour/v1alpha1"
+	"github.com/projectcontour/contour/test/e2e"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	networkingv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
```
sunjayBhatia, line 34: for reviewers: this was a cheap way to do this via behavioral testing, but maybe an even cheaper way would be to assert on the content of the rbac rules applied?

still could add gateway to this, but wanted to wait until i heard people's thoughts on this method

Makefile
```diff
@@ -281,11 +281,12 @@ install-contour-release: | setup-kind-cluster ## Install the release version of
 	./test/scripts/install-contour-release.sh $(CONTOUR_UPGRADE_FROM_VERSION)
 
 .PHONY: e2e
-e2e: | setup-kind-cluster run-e2e cleanup-kind ## Run E2E tests against a real k8s cluster
+e2e: | setup-kind-cluster load-contour-image-kind run-e2e cleanup-kind ## Run E2E tests against a real k8s cluster
```
stevesloka, line 284: This breaks running the tests locally. Could this maybe check for the condition and default to not running the incluster? Or maybe we can buy into the in-cluster only when running in GH actions?

stevesloka (changes requested):

Makefile
sunjayBhatia: ah yeah good call, I'll work on that so you can still just run `make e2e` and things Just Work ™️

test/e2e/incluster/rbac_test.go
```diff
@@ -0,0 +1,185 @@
...
+			}
+			return e2e.HTTPProxyValid(p)
+		}, time.Second*5, time.Millisecond*20)
+
+		// Check request to app works.
+		res, ok := f.HTTP.SecureRequestUntil(&e2e.HTTPSRequestOpts{
+			Host: "rbac-test.projectcontour.io",
+			TLSConfigOpts: []func(*tls.Config){
+				e2e.OptSetSNI("rbac-test.projectcontour.io"),
+			},
```
skriss, line 101: no need to explicitly specify this since SNI matches the Host (see https://github.com/projectcontour/contour/blob/main/test/e2e/http.go#L173-L174)

test/e2e/incluster/rbac_test.go
skriss: I like the style you have here since it's less likely to end up as just a copy-paste of whatever the defined RBAC is.

test/e2e/incluster/leaderelection_test.go
skriss: No good ideas of the top of my head, I thought maybe the managed fields metadata would help but looks like it just sets the manager to "contour", so fine with this for now.

skriss: Was just reading https://kubernetes.io/docs/reference/using-api/server-side-apply/#managers, looks like _maybe_ if we explicitly set the user agent based on the pod name in [here](https://github.com/projectcontour/contour/blob/011fe4023796769e8762587f7e7136c432549720/internal/k8s/clients.go#L41) then the manager would show up as that, would have to do some experimenting though. Wouldn't consider it a blocker for this PR.

test/e2e/incluster/rbac_test.go
sunjayBhatia: ah yes, copy paste!

test/e2e/incluster/leaderelection_test.go
sunjayBhatia: hm yeah confirmed you would have to have an update to the httpproxy or ingress resource you want to observe a different manager on, unless we make some significant changes to the status updater bits to make them rewrite status on all resources when leader election changes

not too bad, but if we're doing it just for a test, probably not worth it, not sure if theres other significant value we get

skriss (approved):
👍

stevesloka (approved):
This is great @sunjayBhatia! In a future PR, maybe we could have a make task which doesn't run the image building and skips the incluster suite for faster iterations when running locally.

sunjayBhatia:
> This is great @sunjayBhatia! In a future PR, maybe we could have a make task which doesn't run the image building and skips the incluster suite for faster iterations when running locally.

yeah that would make iteration quicker for sure, also just had a thought we might want to expose the `skip-package` flag from ginkgo as well so you could do `SKIP=incluster make run-e2e` or something similar

PR #4171, opened 2021-11-10
Use flags for leader election setup

sunjayBhatia:
Pulled out of #4157

- leader election setup is now done via flags
- mention config file usage is deprecated
- remove from contour CRD

Could also wait on https://github.com/projectcontour/contour/pull/4170 to ensure no breakage


apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -59,11 +59,6 @@ type ContourConfigurationSpec struct {
 	// +kubebuilder:default={disablePermitInsecure: false}
 	HTTPProxy HTTPProxyConfig `json:"httpproxy"`
 
-	// LeaderElection contains leader election parameters.
-	// +optional
-	// +kubebuilder:default={leaseDuration: "15s", renewDeadline: "10s", retryPeriod: "2s", disableLeaderElection: false, configmap: {name: "leader-elect", namespace: "projectcontour"}}
-	LeaderElection LeaderElectionConfig `json:"leaderElection"`
```
sunjayBhatia, line 65: note: this is technically a breaking change on this v1alpha1 CRD, do we need to worry about it since its isn't widely circulated/used likely?

cmd/contour/serve.go
```diff
@@ -121,6 +121,13 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 	serve.Flag("incluster", "Use in cluster configuration.").BoolVar(&ctx.Config.InCluster)
 	serve.Flag("kubeconfig", "Path to kubeconfig (if not in running inside a cluster).").PlaceHolder("/path/to/file").StringVar(&ctx.Config.Kubeconfig)
 
+	serve.Flag("disable-leader-election", "Disable leader election mechanism.").BoolVar(&ctx.DisableLeaderElection)
+	serve.Flag("leader-election-lease-duration", "The duration of the leadership lease.").Default("15s").DurationVar(&ctx.Config.LeaderElection.LeaseDuration)
+	serve.Flag("leader-election-renew-deadline", "The duration leader will retry refreshing leadership before giving up.").Default("10s").DurationVar(&ctx.Config.LeaderElection.RenewDeadline)
+	serve.Flag("leader-election-retry-period", "The interval which Contour will attempt to acquire leadership lease.").Default("2s").DurationVar(&ctx.Config.LeaderElection.RetryPeriod)
+	serve.Flag("leader-election-resource-name", "The name of the resource (ConfigMap) leader election will lease.").Default("leader-elect").StringVar(&ctx.Config.LeaderElection.Name)
```
sunjayBhatia, line 128: note this is more generic as "resource-name/namespace" as we can move away from configmap based leader election and use leases, but this covers the current state and the transition period

apis/projectcontour/v1alpha1/contourconfig.go
skriss: Our [documented policy](https://projectcontour.io/resources/deprecation-policy/) is somewhat unclear here: 

The Alpha row says "Behavior can change any time, Fields can be removed at any time", as compared to the Beta row, which says "Behavior can change any time, fields won’t be removed without a version bump (ie v1alpha1 to v1alpha2)". That implies we could drop a field from the alpha group without revving the API version. However, the policy also indicates we're consistent with the Kubernetes policy, which AFAIK does not allow dropping fields without revving the version, even for alpha APIs.

Pragmatically, I feel like this should be OK, but we should also update/clarify our policy if this is something we're comfortable with.

sunjayBhatia: yeah agreed, some of the statements in that doc are at odds, I'll make a PR to update it a bit

skriss: Maybe we can explicitly say that we follow the policy for beta/GA APIs, but for alpha, we're more lenient in that we allow dropping fields without revving the version? I feel like that'd be a reasonably policy.

sunjayBhatia: yep, basically what i was gonna change, something to the effect of "we follow k8s policies, with a caveat for alpha"

sunjayBhatia: https://github.com/projectcontour/contour/pull/4173

youngnick: I agree on this one, alpha should mean "we can change anything at any time".

youngnick (approved):
This LGTM, I think this is a clean way to cut the know about reconciling the Config CRD.

skriss (approved):
LGTM

PR #4173, opened 2021-11-10
Updates to deprecation policy for alpha APIs

sunjayBhatia:
Makes it more lenient since we had conflicting language here, some
saying behavior/fields could change at any time, other saying we are in
sync with k8s policies.


skriss (approved):
:+1: this is sensible to me, gives users good guarantees on the more stable APIs while not creating a huge amount of overhead for us to iterate on alpha APIs.

stevesloka (approved):

Issue #4168, opened 2021-11-10
release v1.18.3 with the RetryPolicy fix


xaleeks:
We need to backport / cherrypick the RetryPolicy fix (https://github.com/projectcontour/contour/pull/4135) that we implemented in the latest v1.19.1 to the 1.18.x as well.  There are users that are still actively using v1.18 because they are not ready to upgrade their k8s version yet.

xaleeks:
assigning to @youngnick first

youngnick:
Okay, I will sort out a 1.18.3 release, containing just this fix.

youngnick:
Release 1.18.3 is out, for both Contour and the Operator.

PR #4174, opened 2021-11-11
[Release 1.18] Cherry-pick: Allow retry policy, num retries to be zero

youngnick:
Cherry-pick PR #4117 to release-1.18, in preparation for a 1.18.3 release.

Updates #4168

Original commit message below.

---
Since the current field for NumRetries is not a pointer, Contour
doesn't have a good way to understand if the field is unspecified,
or set to zero.

To keep the v1 API contract, Contour will allow this field to be set
to -1 which means the numRetries should be zero. If unspecified or set
to zero, then the Envoy default of 1 is used.

### DIFF
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -638,9 +638,12 @@ type RetryOn string
 // RetryPolicy defines the attributes associated with retrying policy.
 type RetryPolicy struct {
 	// NumRetries is maximum allowed number of retries.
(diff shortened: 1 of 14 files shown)


sunjayBhatia (approved):

skriss (approved):

stevesloka (changes requested):
Needs a code regen.

youngnick:
Merging this with admin powers so I can get the release cut.

Issue #3679, opened 2021-05-13
test/e2e: Upgrade testing epic


sunjayBhatia:
This issue is a top-level/epic for capturing todos and progress on establishing an upgrade test suite for contour (roughly in order)

- [x] Build contour image once in CI, share between various e2e jobs
  - Should speed up e2e jobs and ensure we are actually testing against the same artifact
- [x] Initial upgrade test suite: #3678
- [x] Fix upgrade tests for all branches: #3743
- [ ] Upgrade tests should ensure uptime of an app during upgrade
  - Currently just tests routability before/after upgrade
  - [x] Will require multiple nodes in kind/cluster/replicas of app: https://github.com/projectcontour/contour/pull/4176
  - Will require defining some SLO that we need to fulfill during an upgrade (X% of requests succeed with 200 etc.)
- [ ] Upgrade tests should ensure creation of routes during upgrade works
  - Will require defining some SLO that we need to fulfill during an upgrade (creating an Ingress/HTTPProxy etc. works and route is enabled within X time etc.)
- [ ] Upgrade tests should deploy "upgrade from" version rather than relying on cluster running it already to support multiple upgrade scenarios
  - To reduce test time, we should lower the shutdown-manager drain/wait timeouts so that the Envoy pods can be deleted/recreated more quickly

PR #4176, opened 2021-11-11
test/e2e/upgrade: Use multi-node cluster for upgrade testing

sunjayBhatia:
Now we are testing routability *during* an upgrade, we need to have a
more HA deployment of Envoy to test against. This change ensures we have
2 nodes in our kind cluster so the Envoy daemonset is not completely
unavailable during an upgrade.

Kind setup script optionally installs MetalLB in the kind cluster to provide load balancing between nodes

Updates: #3679

We can potentially use this setup for the regular e2e tests, this is just a fix for the upgrade tests currently


sunjayBhatia:
hm, (hopefully) separate flake in the e2e tests that use port forwarding

sunjayBhatia:
> hm, (hopefully) separate flake in the e2e tests that use port forwarding

maybe not a flake, possible influenced by change to `WaitForEnvoyDaemonSetUpdated`

test/scripts/make-kind-cluster.sh
```diff
@@ -77,6 +82,35 @@ do
     kind::cluster::load "$image"
 done
 
+if [[ "${MULTINODE_CLUSTER}" == "true" ]]; then
+    # Install metallb.
+    ${KUBECTL} apply -f https://raw.githubusercontent.com/metallb/metallb/master/manifests/namespace.yaml
+    if ! kubectl get secret -n metallb-system memberlist; then
+        ${KUBECTL} create secret generic -n metallb-system memberlist --from-literal=secretkey="$(openssl rand -base64 128)"
+    fi
+    ${KUBECTL} apply -f https://raw.githubusercontent.com/metallb/metallb/master/manifests/metallb.yaml
```
skriss, line 91: probably makes sense to pin a tagged version of metallb, WDYT?

skriss (changes requested):
Looks reasonable to me, 🤞 that the multi-node cluster doesn't bring GH Actions to its knees. Might be a good idea to try re-running it a bunch of times before we merge, to see how the reliability is.

test/scripts/make-kind-cluster.sh
sunjayBhatia: good point!

stevesloka (approved):

skriss (approved):

PR #4179, opened 2021-11-15
Release 1.18.3 docs

youngnick:
Updates #4168


site/config.yaml
```diff
@@ -37,6 +37,7 @@ params:
   docs_versioning: true
   docs_versions:
     - main
+    - v1.18.3
```
skriss, line 40: Looks like the version drop-down renders things per the order in this list, so let's move this down to below v1.19.0

skriss (changes requested):

site/content/resources/upgrading.md
```diff
@@ -141,6 +141,66 @@ If your version of Contour is older than v1.18.2, please upgrade to v1.18.2 firs
     $ kubectl apply -f examples/contour/03-envoy.yaml
     ```
 
+## Upgrading Contour 1.18.2 to 1.18.3
```
sunjayBhatia, line 144: should we update the upgrade instructions after this as well? (currently 1.18.2 to 1.19.0)

site/data/docs/toc-mapping.yml
```diff
@@ -41,3 +41,4 @@ v1.18.1: v1-18-1-toc
 v1.18.2: v1-18-2-toc
 v1.19.0: v1-19-0-toc
 v1.19.1: v1-19-1-toc
+v1.18.3: v1-18-3-toc
```
sunjayBhatia, line 44: nit: move this after 1.18.2?

site/content/resources/upgrading.md
youngnick: Yeah, I'll update. But this upgrade doc is increasingly terrifying to update, we may need to look at splitting it up somehow.

site/data/docs/toc-mapping.yml
youngnick: Thanks, that makes more sense.

youngnick:
@skriss @sunjayBhatia updated, PTAL.

skriss (approved):

PR #4185, opened 2021-11-16
internal/dag: set Listener condition for invalid route kinds

### LINKED ISSUE #3529: Gateway API: Listener Route binding selector kind must match Routes compatible with Listener protocol

skriss:
Sets a Listener condition of "ResolvedRefs: false" with
a reason of "InvalidRouteKinds" when a Listener has an
invalid allowed route in its spec.

Closes #3529.
Updates #4124.

### DIFF
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -146,7 +146,7 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
 		}
 
 		// Get a list of the route kinds that the listener accepts.
-		listenerRouteKinds := p.getListenerRouteKinds(listener)
+		listenerRouteKinds := p.getListenerRouteKinds(listener, gwAccessor)
 		gwAccessor.SetListenerSupportedKinds(string(listener.Name), listenerRouteKinds)
 
 		attachedRoutes := 0
@@ -205,7 +205,7 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
 
 // getListenerRouteKinds gets a list of the valid route kinds that
 // the listener accepts.
-func (p *GatewayAPIProcessor) getListenerRouteKinds(listener gatewayapi_v1alpha2.Listener) []gatewayapi_v1alpha2.Kind {
+func (p *GatewayAPIProcessor) getListenerRouteKinds(listener gatewayapi_v1alpha2.Listener, gwAccessor *status.GatewayStatusUpdate) []gatewayapi_v1alpha2.Kind {
 	// None specified on the listener: return the default based on
 	// the listener's protocol.
 	if len(listener.AllowedRoutes.Kinds) == 0 {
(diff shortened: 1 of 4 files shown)


internal/dag/status_test.go
```diff
@@ -3738,6 +3738,156 @@ func TestGatewayAPIHTTPRouteDAGStatus(t *testing.T) {
...
+			},
+		},
+		wantGatewayStatusUpdate: []*status.GatewayStatusUpdate{{
+			FullName: types.NamespacedName{Namespace: "projectcontour", Name: "contour"},
+			Conditions: map[gatewayapi_v1alpha2.GatewayConditionType]metav1.Condition{
+				gatewayapi_v1alpha2.GatewayConditionReady: {
+					Type:    string(gatewayapi_v1alpha2.GatewayConditionReady),
+					Status:  contour_api_v1.ConditionTrue,
+					Reason:  status.ReasonValidGateway,
+					Message: "Valid Gateway",
```
stevesloka, line 3772: Should the gateway still be `valid` if the listeners are bad?

internal/dag/gatewayapi_processor.go
```diff
@@ -221,24 +221,35 @@ func (p *GatewayAPIProcessor) getListenerRouteKinds(listener gatewayapi_v1alpha2
...
-		}
-		if *routeKind.Group != gatewayapi_v1alpha2.GroupName {
-			p.Errorf("Listener.AllowedRoutes.Group %q not supported", *routeKind.Group)
+		if routeKind.Group != nil && *routeKind.Group != gatewayapi_v1alpha2.GroupName {
+			gwAccessor.AddListenerCondition(
+				string(listener.Name),
+				gatewayapi_v1alpha2.ListenerConditionResolvedRefs,
+				metav1.ConditionFalse,
+				gatewayapi_v1alpha2.ListenerReasonInvalidRouteKinds,
+				fmt.Sprintf("Group %q is not supported", *routeKind.Group),
```
stevesloka, line 231: nit: Maybe add the group that is supported (since it's only one?).

internal/dag/gatewayapi_processor.go
```diff
@@ -221,24 +221,35 @@ func (p *GatewayAPIProcessor) getListenerRouteKinds(listener gatewayapi_v1alpha2
...
+			)
 			continue
 		}
 		if routeKind.Kind != gatewayapi_v1alpha2.Kind(KindHTTPRoute) && routeKind.Kind != gatewayapi_v1alpha2.Kind(KindTLSRoute) {
```
stevesloka, line 235: nit: Could get rid of the `gatewayapi_v1alpha2.Kind()` and just compare directly. 

e.g.
```
if routeKind.Kind != KindHTTPRoute && routeKind.Kind != KindTLSRoute {
}
``

stevesloka (changes requested):
What about TCPRoute? If you specify that as a valid type shouldn't that get thrown out as unsupported? 

Also the spec says that `when unspecified or empty, the kinds of Routes selected are determined using the Listener protocol.`, so if a listener set to `TCP` or `UDP` then you'd have to set status as well?

internal/dag/status_test.go
skriss: Yeah good point, looks like it should be set to "Ready: false" with reason "ListenersNotValid" if at least one Listener is not valid.

skriss:
> What about TCPRoute? If you specify that as a valid type shouldn't that get thrown out as unsupported?

Yep, that should be covered by https://github.com/projectcontour/contour/blob/c41e3c5553016874c2fc5796ce7c39af57768612/internal/dag/gatewayapi_processor.go#L235-L244

> Also the spec says that `when unspecified or empty, the kinds of Routes selected are determined using the Listener protocol.`, so if a listener set to `TCP` or `UDP` then you'd have to set status as well?

Looks like for this case, we should set "Detached: true" with reason "UnsupportedProtocol" (per https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.ProtocolType).

skriss:
@stevesloka thanks for the comments, they've also made me consider that we should probably always be setting the listeners' Ready and maybe Detached conditions to either true or false, based on what other conditions were set for each listener, similar to how we do for the Gateway's Ready condition. Let me look at adding that here. It might also just be easier to set all of the relevant listener conditions in one PR; I was originally going to do each of them in separate PRs but might be simpler to do in one go.

skriss:
I'm gonna close this PR out and open a revamped one that sets additional listener conditions, along with the summary Listener and Gateway conditions. Stay tuned..

PR #4189, opened 2021-11-19
internal/httpsvc: set TLSv1.3 for optional TLS for contour metrics

tsaarni:
This change was accidentally left out from #3707.  It updates minimum TLS version to TLSv1.3 for Contour metrics endpoint when optional TLS support is used to protect the metrics. Corresponding feature for Envoy metrics was already set to TLSv1.3 by the previous change. Support for metrics over TLS has not been released yet, so this change is backwards compatible.

Added test cases for httpsvc.

Updates #3594


sunjayBhatia (approved):
nice! only nit i would add the commands to generate the testdata somewhere but looks great

skriss (approved):

tsaarni:
> nice! only nit i would add the commands to generate the testdata somewhere but looks great

I created the certs with my own command line tool https://github.com/tsaarni/certyaml which I've been using for couple of years.
OpenSSL is a pain and `certs` target was removed from the Makefile at some point.  I can write a comment how to use `certyaml` to create the test data.

But it occurred to me having this test data would not be necessary if there was easy API for generating test certs, so I added this https://github.com/tsaarni/certyaml#go-api. What do you think, would it make sense / would it be feasible to add this kind of dependency and use it in the tests?

There are couple of places in unittest that relies on certs on git repo, those would be primary candidates to remove.  In theory, even cert-manager could be replaced in e2e as well.

sunjayBhatia:
> > nice! only nit i would add the commands to generate the testdata somewhere but looks great
> 
> I created the certs with my own command line tool https://github.com/tsaarni/certyaml which I've been using for couple of years. OpenSSL is a pain and `certs` target was removed from the Makefile at some point. I can write a comment how to use `certyaml` to create the test data.
> 
> But it occurred to me: having test data in repo would not be necessary if there was easy API for generating test certs "on the fly". So I added API to my tool https://github.com/tsaarni/certyaml#go-api. What do you think, would it make sense / would it be feasible to add this kind of dependency and use it in the tests?
> 
> There are couple of places in unittest that relies on certs on git repo, those would be primary candidates to remove. In theory, even cert-manager could be replaced in e2e as well.

yeah i think that could be a nice change, having certs generated on the fly can def reduce future toil etc.

tsaarni:
> yeah i think that could be a nice change, having certs generated on the fly can def reduce future toil etc. (doesn't need to be a blocker for this if you want to pull it out separately)

Great! I will create new ticket and submit PR for that and merge this TLS version fix as it is for now.

PR #4194, opened 2021-11-23
tests: generate test certificates and keys on the fly

### LINKED ISSUE #4192: Remove cert testdata from remote and create them on the fly
Couple of TLS test cases depend on certificates/private keys that are stored as testdata 

- `cmd/contour/testdata`
- `internal/httpsvc/testdata`

It would be better if the certificates were generated during execution, instead of stored in the repo (xref discussion at https://github.com/projectcontour/contour/pull/4189#issuecomment-975431963). Reasons include:

- Having the code generating certs within the test case expresses the intention better, especially if test exercises exceptional case like expired certificate.
- Security tools might give alerts for credentials stored in git repo.
- Certificates might cause failure of the test suite if they are considered outside of their validity time e.g. due to system clock being set wrong, or in future if test certs were created with too short expiration time.

tsaarni:
Remove certificates and keys from testdata directories and prefer generating
them on demand.

Fixes #4192


tsaarni:
This proposal uses a new library I created https://github.com/tsaarni/certyaml. If needed, that could be used in e2e tests as well when needing very specific certificates for the tests e.g. already expired certs, or just for smaller footprint.

youngnick:
The idea makes sense, and I like the approach in general. I'll give the code a proper review in a bit.

youngnick (approved):
Yep, this looks great.

I feel like I should mention the (small) risk of picking up the dependency on something that only one maintainer owns though. It's not a blocker,  and I don't think any problems there are likely, but it is something to note. 

Sorry @tsaarni, not making any comment about you personally, I'd just like to have talked about the event that you can't continue as a maintainer at some point.

sunjayBhatia (approved):

tsaarni:
> I feel like I should mention the (small) risk of picking up the dependency on something that only one maintainer owns though. It's not a blocker, and I don't think any problems there are likely, but it is something to note.
> 
> Sorry @tsaarni, not making any comment about you personally, I'd just like to have talked about the event that you can't continue as a maintainer at some point.

I agree.  I have of course also bumped into several projects myself that practically have no maintainer anymore.  I was bit hesitant if I should propose additional dependency, and I would not have proposed if it wasn't for the test code, where the impact would be likely less severe.

youngnick:
I think as long as we all acknowledge the risk, we can manage it, and as you say, it's in the test code as well, so the impact is less.

PR #4195, opened 2021-11-23
move route DAG->Envoy logic into internal/envoy/v3

skriss:
Moves logic for converting DAG objects to
Envoy objects from the route xds cache to
the internal/envoy/v3 package.

### DIFF
--- a/internal/envoy/v3/route.go
+++ b/internal/envoy/v3/route.go
@@ -457,7 +457,7 @@ func RouteConfiguration(name string, virtualhosts ...*envoy_route_v3.VirtualHost
 	}
 }
 
-// corsPolicy returns a *envoy_route_v3.corsPolicy
+// corsPolicy returns a *envoy_route_v3.CorsPolicy
 func corsPolicy(cp *dag.CORSPolicy) *envoy_route_v3.CorsPolicy {
 	if cp == nil {
 		return nil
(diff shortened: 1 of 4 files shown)


internal/envoy/v3/route.go
```diff
@@ -34,6 +34,96 @@ import (
...
+
+	if vh.RateLimitPolicy != nil && vh.RateLimitPolicy.Global != nil {
+		evh.RateLimits = GlobalRateLimits(vh.RateLimitPolicy.Global.Descriptors)
+	}
+
+	return evh
+}
+
+// Route converts a DAG route to an Envoy route.
+func Route(dagRoute *dag.Route, vhostName string, secure bool, authService *dag.ExtensionCluster) *envoy_route_v3.Route {
```
sunjayBhatia, line 64: nit: could this be unexported possibly? maybe name to `buildRoute` or something similar, just to disambiguate between this and the other `Route*` helpers (particularly `RouteRoute`)

sunjayBhatia (approved):

internal/envoy/v3/route.go
skriss: :+1: yeah a few of the helpers can be unexported now, and yeah naming is tough, but `buildRoute` seems sensible. Will make some updates.

stevesloka (approved):

Issue #4191, opened 2021-11-22
single ingress path regex error breaks all other ingress objects processing


r0bj:
**What steps did you take and what happened:**
If ingress path regex is broken (e.g. too long) it breaks all other ingress objects processing. It means that a single ingress object containing broken regex can break entire ingress controller. During that time all other ingress object are not updated, ingress changes are not reflected in envoy.

Example:
ingress:
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: broken-contour
spec:
  rules:
  - host: test.example.com
    http:
      paths:
      - backend:
          service:
            name: test-service
            port:
              number: 80
        path: /{corporate-pages:about|careers|contact|mediakit|press|partnerships|privacy-policy|do-not-sell-my-info}
        pathType: Prefix
```

Contour error message is generated:
```
time="2021-11-22T20:39:22Z" level=error msg="regex '/\\{corporate-pages:about\\|careers\\|contact\\|mediakit\\|press\\|partnerships\\|privacy-policy\\|do-not-sell-my-info\\}((\\/).*)?' RE2 program size of 121 > max program size of 100 set for the error level threshold. Increase configured max program size if necessary." code=13 connection=561 context=xds node_id=envoy-3-vxb7p node_version=v1.19.1 response_nonce=1 version_info=7
```

**What did you expect to happen:**
Even if regex is broken in single ingress object contour should be able to still process all other ingresses.

**Anything else you would like to add:**
It would probably be useful to have some metric that could indicate that contour ingress object processing is broken. Otherwise it's difficult to detect such case when not looking at contour logs.

**Environment:**

- Contour version: v1.19.1
- Kubernetes version: (use `kubectl version`):
```
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.2", GitCommit:"092fbfbf53427de67cac1e9fa54aaa09a28371d7", GitTreeState:"clean", BuildDate:"2021-06-16T12:52:14Z", GoVersion:"go1.16.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.2", GitCommit:"092fbfbf53427de67cac1e9fa54aaa09a28371d7", GitTreeState:"clean", BuildDate:"2021-06-16T12:53:14Z", GoVersion:"go1.16.5", Compiler:"gc", Platform:"linux/amd64"}
```

[...]

PR #4197, opened 2021-11-24
Simplify regex path match to reduce program size

sunjayBhatia:
Ingress/HTTPRoute prefix matches now have an anchor to ensure long
prefixes don't increase the program size. Should ensure they are not
rejected by Envoy.

In addition, add an anchor to regex path matches since they at least
have to start with a literal /

Manually ran ingress conformance tests and besides flakes, all passed 👍🏽 

Updates: #4191


youngnick (approved):

skriss (approved):
LGTM (relying on those ingress conformance tests to validate the regex itself).

sunjayBhatia:
> LGTM (relying on those ingress conformance tests to validate the regex itself).

yeah, need to get those running in CI but theyre pretty flaky as-is

PR #4201, opened 2021-11-30
implement redirect route policy

### LINKED ISSUE #1627: Redirect: adding a redirect (301 or 302) option in ingressroute / httpproxy
The ingressroute / httpproxy should be able to support a match and then redirect with either a 301 or 302 to an http or https destination

Something like:
```
spec:
  routes:
  - match: /
    services:
    - name: awesome-service-01
      port: 1000
   - match: /redirectpath
    redirect:
      method: 301
      destination: https://externalawesomewebsite.com
  virtualhost:
    fqdn: awesomehost.tld
```

The use case would be that for example if you have a website pointed to the ingress you can now also do redirects in contour. This would prevent you from having to spin up another container inside for the redirect (nginx for example).

**Environment:**

- Contour version: v1.0.0-beta.1
- Kubernetes version: (use `kubectl version`): v1.14.6-gke.1
- Kubernetes installer & version: N/A (Google)
- OS (e.g. from `/etc/os-release`): GKE

stevesloka:
add a HTTPRequestRedirectPolicy to the HTTPProxy spec which allows for HTTPProxy.Routes to specify a requestRedirectPolicy. This policy will allow a redirect to be configured for a specific set of Conditions within a single route.

Fixes #1627


apis/projectcontour/v1/httpproxy.go
```diff
@@ -353,6 +353,43 @@ type Route struct {
 	// The policy for rate limiting on the route.
 	// +optional
 	RateLimitPolicy *RateLimitPolicy `json:"rateLimitPolicy,omitempty"`
+
+	// RequestRedirectPolicy defines an HTTP redirection.
+	// +optional
+	RequestRedirectPolicy *HTTPRequestRedirectPolicy `json:"requestRedirectPolicy,omitempty"`
```
skriss, line 359: The only issue I see with the API is that `services` is still required to have at least one entry, even if you just want a redirect. We could make `services` optional, and then check in the DAG build that either `services` or `requestRedirectPolicy` is specified. That does have the downside of slower feedback to the user though (at DAG build time rather than at HTTPProxy creation time). I _think_ I'd probably still make that change, as having to put dummy data into `services` when you want a redirect feels clumsy, but interested in what others think.

youngnick: Agreed, having to specify a service when the traffic won't get there feels a bit weird.

Making a required field optional is a compatible API change (because old objects will continue to work), so we can conceivably make `services` optional without an API version increment. So I guess I'm saying I'm a +1 to making `services` optional.

stevesloka: That's a good point @skriss! Looks like it's just a kubebuilder default that we could change to enable not requiring services to be present.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -353,6 +352,43 @@ type Route struct {
...
+	// header in the response.
+	// When empty, the scheme of the request is used.
+	// +optional
+	// +kubebuilder:validation:Enum=http;https
+	Scheme *string `json:"scheme,omitempty"`
+
+	// Hostname is the hostname to be used in the value of the `Location`
+	// header in the response.
+	// When empty, the hostname of the request is used.
+	// +optional
```
youngnick, line 373: Feels like it's probably worth mentioning that this should be a precise hostname, no wildcards allowed (good job logging that upstream issue in Gateway API as well, thanks).

changelogs/unreleased/4201-stevesloka-minor.md
```diff
@@ -0,0 +1,5 @@
+### HTTP Request Redirect Policy
```
youngnick, line 1: As part of making `service` optional, this needs an update.

Something like:

### HTTPProxy.Route.Service field now optional

As part of implementing Request Redirect Policy, HTTPProxy.Route.Service has been made optional. Either the Service or HTTPRequestRedirectPolicy fields must be set under the HTTPProxy.Route for the HTTPProxy to be valid. Not having at least one of these fields set will mean that the Route is rejected, and the appropriate Condition set on the proxy.

youngnick (changes requested):
Just a couple of small things, but this is looking really close to me.

internal/featuretests/v3/redirectroutepolicy_test.go
```diff
@@ -0,0 +1,80 @@
...
+					&envoy_route_v3.Route{
+						Match: routePrefix("/"),
+						Action: &envoy_route_v3.Route_Redirect{
+							Redirect: &envoy_route_v3.RedirectAction{
+								HostRedirect: "envoyproxy.io",
+								PortRedirect: 443,
+								SchemeRewriteSpecifier: &envoy_route_v3.RedirectAction_SchemeRedirect{
+									SchemeRedirect: "https",
+								},
+								ResponseCode: 0,
```
sunjayBhatia, line 70: should be 301 here?

sunjayBhatia: ah nvm this is an enum type, maybe worth using a constant for readability (just bc seeing 0 threw me off)

skriss (approved):

sunjayBhatia (approved):

PR #4202, opened 2021-11-30
Convert to controller runtime runnables

sunjayBhatia:
- controller runtime manages controllers and other runnables (previously
goroutines added to the server workgroup)
- leader election now uses a combination of configmap and lease objects,
in a future release we can move to just leases
- controllers and other components are currently not dependent on leader election
- other components (status writers) are set to be dependent on leader
election
- adds to the leaderelection e2e tests to check on leaderelection
configmap in addition to new lease resource and events


sunjayBhatia:
overrides https://github.com/projectcontour/contour/pull/4157

internal/controller/controller_test.go
```diff
@@ -0,0 +1,65 @@
...
+		},
+		"tlsroute controller": func(mockManager *mocks.Manager) error {
+			return controller.RegisterTLSRouteController(fixture.NewTestLogger(t), mockManager, nil)
+		},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			mockManager := &mocks.Manager{}
+
```
sunjayBhatia, line 49: curious what people think about these tests, they ensure the controller is registered and that it does not require leader election, using a generated mock manager (see mock package and `go generate` command above)

internal/contour/handler_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+package contour
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestEventHandlerNotRequireLeaderElection(t *testing.T) {
```
sunjayBhatia, line 22: curious what people think of these tests for the various components, settled on this after spiking through another option which i wonder what people think of

- new object as thin wrapper around controller-runtime manager
- mainly for testing, can test by mocking controller-runtime manager (see controller tests for example)
- instead of `manager.Add(Runnable)` method, have something like `manager.Add(Runnable, requireLeaderElection bool)`

decided against this for a few reasons, mainly bc it ended up being a v thin wrapper (need to expose a lot of things from the base controller-runtime manager) and puts more logic in the `main` package, rather than closer to the components themselves which should have the knowledge about whether they need leader election or not

examples/contour/02-role-contour.yaml
```diff
@@ -13,6 +13,7 @@ rules:
   - ""
   resources:
   - configmaps
+  - events
```
skriss, line 16: Is this for the controller-runtime manager to write events?

hack/actions/install-kubernetes-toolchain.sh
```diff
@@ -56,6 +56,9 @@ if [[ ${GITHUB_ACTIONS} == "true" && ${OS} == "linux" ]]; then
   mv /home/runner/go/bin/ginkgo ${DESTDIR}/ginkgo
 fi
 
+# Install mockery for generating go mocks
+go install github.com/vektra/mockery/v2@v2.9.4
+
```
skriss, line 61: I believe this can be moved into `tools.go` and managed as a normal go module dependency

internal/controller/controller.go
```diff
@@ -0,0 +1,25 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package controller
+
+import "sigs.k8s.io/controller-runtime/pkg/controller"
+
+// TODO: could make this more generic for any Runnable
+type noLeaderElectionController struct {
```
skriss, line 19: worth some comments, either here or where it's used, on why we have this.

internal/debug/debug.go
```diff
@@ -30,12 +31,16 @@ type Service struct {
 	Builder *dag.Builder
 }
 
+func (svc *Service) NeedLeaderElection() bool {
+	return false
+}
+
 // Start fulfills the g.Start contract.
```
skriss, line 38: nit, this comment is now out-of-date.

internal/httpsvc/http.go
```diff
@@ -43,9 +43,13 @@ type Service struct {
 	http.ServeMux
 }
 
+func (svc *Service) NeedLeaderElection() bool {
+	return false
+}
+
 // Start fulfills the g.Start contract.
```
skriss, line 50: nit, comment out of date.

cmd/contour/serve.go
```diff
@@ -169,7 +166,6 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 }
 
 type Server struct {
-	group      workgroup.Group
```
skriss, line 172: looks like we could almost drop the `workgroup` package entirely, but it's still used in the featuretests framework. We could do the work to fully drop it in a separate PR.

cmd/contour/serve.go
```diff
@@ -380,46 +383,48 @@ func (s *Server) doServe() error {
...
 	eventHandler := &contour.EventRecorder{
 		Next:    contourHandler,
 		Counter: contourMetrics.EventHandlerOperations,
 	}
 
-	// Register leadership election.
-	if s.ctx.DisableLeaderElection {
-		contourHandler.IsLeader = disableLeaderElection(s.log)
-	} else {
-		contourHandler.IsLeader = setupLeadershipElection(&s.group, s.log, s.ctx.Config.LeaderElection, s.coreClient, contourHandler.UpdateNow)
```
skriss, line 412: this code used to trigger a DAG rebuild immediately when a Contour instance was elected leader, which helped avoid stale status issues - I'm not seeing anything that will do that now, but could be missing it.

internal/featuretests/v3/featuretests.go
```diff
@@ -160,15 +155,15 @@ func setup(t *testing.T, opts ...interface{}) (cache.ResourceEventHandler, *Cont
 		}()
 		return srv.Serve(l) // srv now owns l and will close l before returning
 	})
-	g.Add(eh.Start())
+	g.AddContext(eh.Start)
```
skriss, line 158: this is the workgroup that would have to be dropped if we wanted to drop the entire workgroup package.

skriss (changes requested):
Handful of comments, but overall this is looking nice!

examples/contour/02-role-contour.yaml
sunjayBhatia: yep, without it, leader election still works but theres errors in the logs from failures to write events

cmd/contour/serve.go
sunjayBhatia: yeah thats a good point, I'll try to think up a test that exercises that

examples/contour/02-role-contour.yaml
stevesloka: Might be good to add this to a role instead of a cluster role, I think for this specifically, it doesn't need cluster-wide access.

cmd/contour/serve.go
sunjayBhatia: yeah, didnt drop it entirely in case we needed it for some reason, could just use some simple coordination mechanisms to run components in tests w/o the need for the workgroup code

internal/controller/controller.go
```diff
@@ -0,0 +1,25 @@
...
+	controller.Controller
+}
+
+func (*noLeaderElectionController) NeedLeaderElection() bool {
+	return false
```
stevesloka, line 24: Does this just fulfill an interface or something?

sunjayBhatia: yep, see https://github.com/kubernetes-sigs/controller-runtime/blob/7e7bf8c1e1df991b957dfaa8b36d975740f309ab/pkg/manager/manager.go#L301-L306

could do something to ensure the type is assignable to a particular interface to do a compilation-time check, but also would need to check the return value of this method since it matters as well, which is a little unfortunate how they implemented it

examples/contour/02-role-contour.yaml
sunjayBhatia: hm yeah thats a good point, might open a separate PR to address this since all of the RBAC for leaderelection resources could be namespace scoped, but it makes the role generation a little different

sunjayBhatia: heres what things might look like if we switched to more scoped resources for the leaderelection resource RBAC: https://github.com/projectcontour/contour/pull/4204

stevesloka: Yup I think that makes sense, the less permissions we need to have the better overall.

stevesloka (approved):
/lgtm with @skriss's comments. Nice work @sunjayBhatia!

internal/featuretests/v3/featuretests.go
skriss: https://github.com/projectcontour/contour/issues/4206

cmd/contour/serve.go
skriss: https://github.com/projectcontour/contour/issues/4206

internal/controller/gateway.go
```diff
@@ -45,35 +44,36 @@ type gatewayReconciler struct {
...
+	c, err := controller.NewUnmanaged("gateway-controller", mgr, controller.Options{Reconciler: r})
 	if err != nil {
-		return nil, err
+		return err
+	}
+	if err := mgr.Add(&noLeaderElectionController{c}); err != nil {
+		return err
 	}
 
 	if err := c.Watch(
```
sunjayBhatia, line 71: also note that these sort of watches *should* work with dynamic updates to contour configuration since we pass in a predicate func that has access to the reconciler's data, we should be able to update that data with correct concurrency/locking mechanisms (could be other solutions, this first springs to mind, theoretically should be cheap since we shouldnt get too many config crd updates)

sunjayBhatia: a point in favor of doubling down on controller-runtime managed controllers/runnables, in case that was a concern @projectcontour/maintainers

cmd/contour/serve.go
sunjayBhatia: the e2e test i can think of is a little racy:
1. delete leader pod
2. make a change to an httpproxy/gateway/gatewayclass/etc. that would make it invalid
3. eventually resource should have correct status

however, 2. needs to be timed just right, after the original leader has stopped but before a new leader has been elected

sunjayBhatia: simple fix would be to replicate what we had before, and pass in the manager "elected" channel to the event handler, so it re-runs on leader election

longer-term, seems we might need to have coverage over this logic with a controller-runtime style test of the controller itself: https://github.com/projectcontour/contour/blob/7bf2309d411a11a80b1e0f75918f9ea5b8932d78/internal/controller/gateway.go#L89-L107 (and propagate it to other controllers, not just gateway and gatewayclass)

this way we can more easily control and test how these processes are coordinated

we might also want to possibly move other things to controllers rather than informer-based, e.g. httpproxy/ingress/secrets etc. to standardize testing

skriss: One idea for the E2E, could you manually create the leader-elect configmap before starting up Contour, indicating some non-existent pod holds the lease? That way no Contour pod would acquire it for at least a few seconds. Still some raciness, but maybe would be easier than having to delete a pod. Food for thought.

sunjayBhatia: another race i think is that even if you do trigger a dag update/status update handler update on leader election, if that isn't running yet (bc some processes start later), you might miss a write to appropriate channels and miss an update

skriss:
@sunjayBhatia I was playing with this locally and had a couple issues:
- proxies created *after* my local Contour acquires leadership do not seem to trigger a DAG rebuild/get status
- the process is not shutting down cleanly on SIGTERM

I imagine the first is related to the E2E failures. I haven't had a chance to dig into these any further yet, just wanted to put the reports here. I'll look some more later this AM.

sunjayBhatia:
> @sunjayBhatia I was playing with this locally and had a couple issues:
> 
> * proxies created _after_ my local Contour acquires leadership do not seem to trigger a DAG rebuild/get status
> * the process is not shutting down cleanly on SIGTERM
> 
> I imagine the first is related to the E2E failures. I haven't had a chance to dig into these any further yet, just wanted to put the reports here. I'll look some more later this AM.

yep, same issues here, e2e tests do some capturing of the issues but a little worried about other components besides just the main status updater (e.g load balancer status updater), needs a little more work

cmd/contour/serve.go
skriss: Another idea for the implementation here would be to bring back the event handler's `UpdateNow` function, then define a new leader-elected Runnable that, when started, just calls `eventHandler.UpdateNow()`. Shouldn't be much different from what you have now, but maybe a little simpler?

sunjayBhatia:
> `Contour server failed" error="failed waiting for all runnables to end within grace period of 30s: context deadline exceeded` on shutdown, could adjust to not use graceful shutdown as well

i think fixing my silly channels mistake also helped with shutdown issues

internal/k8s/status_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+package k8s
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestStatusUpdateHandlerRequiresLeaderElection(t *testing.T) {
```
skriss, line 22: It might be good to also verify in these tests that e.g. `StatusUpdateHandler` implements the `LeaderElectionRunnable` interface, just to ensure no silly typos or anything

skriss (commented):
One thought on the tests, otherwise I think this looks good, doing some ad-hoc testing locally  too

cmd/contour/serve.go
sunjayBhatia: https://github.com/projectcontour/contour/pull/4202/commits/602f8f7c1d1258136d3d7c4ff2f686c7865513c0

wdyt, used that idea ^ rather than having channels flying around

internal/leadership/notifier.go
```diff
@@ -0,0 +1,39 @@
...
+	ToNotify []NeedLeaderElectionNotification
+}
+
+func (n *Notifier) NeedLeaderElection() bool {
+	return true
+}
+
+func (n *Notifier) Start(ctx context.Context) error {
+	for _, t := range n.ToNotify {
+		go t.OnElectedLeader()
```
sunjayBhatia, line 35: ran these non-blocking, dont think this should cause any issues, but worth thinking about 🤔

skriss: I can't think of any issues with it 👍

skriss (approved):
:+1: local testing all looks good to me, as does the new leader election notification abstraction

sunjayBhatia:
Tests need a small fix to make them more robust now we’re deleting contour and restarting in incluster suite

skriss (approved):
LGTM, not sure if there were any other outstanding threads here though

sunjayBhatia:
> LGTM, not sure if there were any other outstanding threads here though

i think we got everything!

PR #4203, opened 2021-12-01
test/scripts: ignore pre-release tags when getting upgrade version

skriss:
Excludes tags with "alpha", "beta" or "rc" when
getting the tag to upgrade from for upgrade tests.

### DIFF
--- a/test/scripts/get-contour-upgrade-from-version.sh
+++ b/test/scripts/get-contour-upgrade-from-version.sh
@@ -6,12 +6,12 @@ set -o pipefail
 
 if CURRENT_TAG=$(git describe --tags --exact-match 2>/dev/null); then
   # We are on a tag, so find previous tag to this one.
-  git tag -l --sort=-v:refname | grep -A1 $CURRENT_TAG| tail -1
+  git tag -l --sort=-v:refname | grep -A10 $CURRENT_TAG | grep -v 'alpha\|beta\|rc' | head -1
 elif git branch --show-current | grep -q release; then
   # We are on a release branch, so find tag.
   git describe --tags --abbrev=0
 else
   # We are likely on main or some other checkout, just use latest tag.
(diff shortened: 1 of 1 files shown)


test/scripts/get-contour-upgrade-from-version.sh
```diff
@@ -6,12 +6,12 @@ set -o pipefail
...
   # We are on a tag, so find previous tag to this one.
-  git tag -l --sort=-v:refname | grep -A1 $CURRENT_TAG| tail -1
+  git tag -l --sort=-v:refname | grep -A10 $CURRENT_TAG | grep -v 'alpha\|beta\|rc' | head -1
 elif git branch --show-current | grep -q release; then
   # We are on a release branch, so find tag.
   git describe --tags --abbrev=0
 else
   # We are likely on main or some other checkout, just use latest tag.
   # If needed, user can override this version with environment variables.
-  git describe --tags $(git rev-list --tags --max-count=1)
```
skriss, line 16: I noticed this was incorrectly pulling "v1.18.3" right now for commits on main, which is chronologically the newest tag but not the highest semver tag, so switched it to use the semver-based sorting logic

skriss: See e.g. https://github.com/projectcontour/contour/runs/4383967716?check_suite_focus=true#step:7:160

sunjayBhatia (approved):
tried it out, looks good!

I guess the assumption with `-A10` is we wouldn't have more than 10 alpha/beta/rc tags, which makes sense to me as reasonable

skriss:
> I guess the assumption with `-A10` is we wouldn't have more than 10 alpha/beta/rc tags, which makes sense to me as reasonable

Yeah, seemed like a safe #

stevesloka (approved):

PR #4204, opened 2021-12-01
Switch to Role for leaderelection resource RBAC

sunjayBhatia:
- New RoleBinding for new contour Role, named contour-rolebinding since we can't modify the existing one
- RBAC for leaderelection moved from contour ClusterRole to projectcontour/contour Role
- make appropriate changes to e2e tests


sunjayBhatia:
impetus was from https://github.com/projectcontour/contour/pull/4202#discussion_r760395054

switching to a namespace scoped role makes sense, though does require some changes to our example deployment and good release notes/expectation setting and downstream consumers knowing about the change

the namespace for the role is hardcoded in our code, could be swapped out by someone manually creating their own manifests or with further tooling (ytt overlays, etc?)

sunjayBhatia:
would argue if we do it, this should maybe be a "major" change

sunjayBhatia:
Need to work out the upgrade tests, current failure is:

```
RoleBinding.rbac.authorization.k8s.io "contour" is invalid: roleRef: Invalid value: rbac.RoleRef{APIGroup:"rbac.authorization.k8s.io", Kind:"Role", Name:"contour"}: cannot change roleRef
```

sunjayBhatia:
Will probably have to fiddle with the naming of resources since it seems modifying a rolebinding is a no-go, unless it is ok to delete and re-create as part of an upgrade

stevesloka (approved):
I think this is a great tiny security thing that helps out! One dependency I can think of is the helm chart needing an update possibly to handle the namespace.

youngnick:
This is looking great! I think the changelog/release note is really important to get right on this one, I'll approve once that's done.

sunjayBhatia:
might need a check to see if there are any merge issues, but should be good to review

skriss (commented):
Change LGTM. We'll need to follow up with updates to several downstream consumers:
- operator
- Helm chart
- TCE
- anything else?

skriss (approved):

youngnick (approved):

sunjayBhatia:
I'll open some issues on the relevant consumers so they are updated, once the next release is cut with this change

sunjayBhatia:
https://github.com/projectcontour/contour-operator/pull/488 operator change, waiting on this one

PR #4209, opened 2021-12-02
internal/dag: set HTTPRoute condition for invalid path match

### LINKED ISSUE #3906: HTTPRoute processing should ensure `/` is prefixed onto path matches
Create an HTTPRoute with a path match without a `/` prefix, requests to route are 404ed by Envoy

HTTP requests routed correctly *or* the HTTPRoute is marked as invalid with an condition describing what is wrong

skriss:
Sets an HTTPRoute condition of "ValidMatches: false" with
reason "InvalidPathMatch" when a path match does not start
with '/' or contains consecutive '/' characters.

Closes #3906.


internal/dag/gatewayapi_processor.go
```diff
@@ -896,30 +895,48 @@ func (p *GatewayAPIProcessor) validateBackendRef(backendRef gatewayapi_v1alpha2.
...
-		return &PrefixMatchCondition{Prefix: "/"}, nil
+		return &PrefixMatchCondition{Prefix: "/"}, true
 	}
 
 	path := pointer.StringDeref(match.Value, "/")
 
 	// If path match type is not defined, default to 'PathPrefix'.
 	if match.Type == nil || *match.Type == gatewayapi_v1alpha2.PathMatchPathPrefix {
+		if !strings.HasPrefix(path, "/") {
+			routeAccessor.AddCondition(status.ConditionValidMatches, metav1.ConditionFalse, status.ReasonInvalidPathMatch, "Match.Path.Value must start with '/'.")
```
skriss, line 908: note, this is a Contour-specific condition type for now as there is not one defined upstream.

sunjayBhatia (approved):

stevesloka (approved):
/lgtm

Feels like we should add some rules upstream to also handle this. I'll open an issue there.

PR #4216, opened 2021-12-06
test/e2e/httpproxy: Loosen cookie matching to a contains match

sunjayBhatia:
In case cookie attributes are in different order

Re:
https://github.com/projectcontour/contour/runs/4435347950?check_suite_focus=true


skriss (approved):
LGTM pending CI, nice find

youngnick (approved):
Nice find!

PR #4227, opened 2021-12-09
experimenting with reorganizing builder test cases

skriss:
This was one idea I had for making `builder_test.go` more manageable. Many times, we have a set of test cases that "go together", because they're testing a particular field or piece of functionality, and typically the test cases are copy-pasted, with small variations made by hand in each one to exercise a slightly different function or error case. What if instead of copy-pasting the whole case by hand, we programmed the deltas?

I think there are a few nice benefits here:
- generally less code (this simple proof of concept was able to drop almost 200 lines of code by organizing just 2 groups of test cases this way) which is easier to read and work with
- better logical organization makes it easier to find relevant tests, etc.
- cases that exercise error conditions are less likely to pass for the wrong reason. I've found a few builder tests that are passing for the wrong reason, usually because we're trying to test one particular error condition but actually there's another error condition in the test data that's causing nothing to be added to the DAG. This is an inevitable outcome of massive copy-paste.

Interested in thoughts here. This is definitely just a proof of concept, and actually completing a refactoring along these lines would be quite a bit more work.

### DIFF
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -289,53 +289,6 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 		},
 	}
 
-	gatewayTLSPassthroughSameNamespace := &gatewayapi_v1alpha2.Gateway{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      "contour",
-			Namespace: "projectcontour",
-		},
-		Spec: gatewayapi_v1alpha2.GatewaySpec{
-			GatewayClassName: gatewayapi_v1alpha2.ObjectName(validClass.Name),
-			Listeners: []gatewayapi_v1alpha2.Listener{{
-				Port:     80,
-				Protocol: gatewayapi_v1alpha2.TLSProtocolType,
-				TLS: &gatewayapi_v1alpha2.GatewayTLSConfig{
-					Mode: gatewayapi.TLSModeTypePtr(gatewayapi_v1alpha2.TLSModePassthrough),
-				},
-				AllowedRoutes: &gatewayapi_v1alpha2.AllowedRoutes{
-					Namespaces: &gatewayapi_v1alpha2.RouteNamespaces{
-						From: gatewayapi.FromNamespacesPtr(gatewayapi_v1alpha2.NamespacesFromSame),
-					},
-				},
-			}},
-		},
-	}
-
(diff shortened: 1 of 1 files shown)


sunjayBhatia:
> cases that exercise error conditions are less likely to pass for the wrong reason

+1 for helping with that!

sunjayBhatia (commented):
i like this!

it might even make it a little easier to see/perform a red-green-refactor TDD approach

one other idea to explore that may help avoid false positives could be to separate out in the `add` func where we do the change to the input and expected output to avoid mistakes/copy paste? maybe the mutator func could become two, one for the input and one for the output, if that isn't too complicated (not really a fully baked idea, what you have already is great, just maybe the concept of the lines of code where the input and expected output being separated makes me think we would be more careful when making additions)

youngnick:
I only just had a look at this, it looks really good. I used a conceptually similar approach for the tests in #4292, this approach seems even easier to parse.

tsaarni:
Following is not directly related, but I have a question regarding issue I have been wondering about these long table driven tests:

When troubleshooting a test case I often end up needing to set a breakpoint to step through the code. This becomes bit harder to do from IDE (such as vscode) when table driven subtests are used - IDE does not know about subtests and runs the whole test case. Setting a breakpoint in the test loop will of course stop at every subtest and not only the one you wanted to troubleshoot. I often revert back in commenting out other subtests to let me debug under IDE focusing only on single subtest.  

In some parts of the code the convention is turned around: instead of having a table and for loop with `t.Run()` the loop is unrolled and `Run()` is called via wrapper function like done [here](https://github.com/projectcontour/contour/blob/d797ae3154443343362909e191c6921f61038b7e/cmd/contour/shutdownmanager_test.go#L138-L192). To start debugging specific subtest I can then put a breakpoint at the `run()` line.

How do you usually deal with this?

youngnick:
I guess I avoid this problem by not using an IDE properly - I never use breakpoints or similar, I tend to use print statements. I guess too much time spent using non-breakpointable languages. :)

skriss:
Yeah, same for me, I don't really use a debugger these days; I will comment out the irrelevant test cases on occasion if needed. However, I think the point remains that some of the table driven tests, particularly the ones that have dozens/hundreds of complex entries, become pretty unwieldy for editing **and** debugging. Definitely something we should keep in mind if/when we put significant effort into refactoring.

PR #4228, opened 2021-12-09
internal/dag: move some builder tests into different files

skriss:
Moves a few test functions from builder_test.go into
different test files to better align with what they're
testing. This is just straight copy-paste from one
place to another.


youngnick (approved):
That's a huge yes from me, anything that makes builder_test.go smaller is good.

sunjayBhatia (approved):

PR #4230, opened 2021-12-09
add AuthBeforeRateLimiting field to AuthorizationServer

### LINKED ISSUE #3726: Allow configuration of ext_authz, global_rate_limit HTTP filter priority
Hi, loving Contour, great work!

We've been using the external authz support with great success and recently investigated taking advantage of the new Global Rate Limiter configuration and discovered that the global rate limit filter is hardcoded to being applied after any external authz filter. This means that using the global rate limiter for DDOS protection is not going to protect the external authorization implementation.

The ext_authz filter is added to the filters in

https://github.com/projectcontour/contour/blob/4c0b72ee9d14fe19eaf835417661b9ffeb9ad4de/internal/xdscache/v3/listener.go#L487

while the global rate limiter is added in

https://github.com/projectcontour/contour/blob/4c0b72ee9d14fe19eaf835417661b9ffeb9ad4de/internal/xdscache/v3/listener.go#L499

Note that `ext_authz` is added after _local_ rate limiting (which is part of `DefaultFilters`):

Naturally it would be _best_ if the user could configure the order of these for their own use case, since it meaningfully impacts the meaning of these filters.

skriss:
Adds an AuthBeforeRateLimiting field, which allows
the user to control whether auth happens first, or
rate limiting happens first.

This also flips the default behavior to rate limit
before auth.

Closes #3726.

### DIFF
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -175,6 +175,14 @@ type AuthorizationServer struct {
 	// WithRequestBody specifies configuration for sending the client request's body to authorization server.
 	// +optional
 	WithRequestBody *AuthorizationServerBufferSettings `json:"withRequestBody,omitempty"`
+
+	// AuthBeforeRateLimiting defines whether the authorization server
+	// will be consulted before applying any rate limiting to a
+	// request. The default is false, i.e. to apply rate-limiting first,
+	// in order to protect the authorization server.
+	//
+	// +optional
+	AuthBeforeRateLimiting bool `json:"authBeforeRateLimiting,omitempty"`
 }
 
 // AuthorizationServerBufferSettings enables ExtAuthz filter to buffer client request data and send it as part of authorization request
--- a/internal/xdscache/v3/listener.go
+++ b/internal/xdscache/v3/listener.go
@@ -427,8 +427,8 @@ func (c *ListenerCache) OnChange(root *dag.DAG) {
 				if vh.AuthorizationService != nil && vh.AuthBeforeRateLimiting {
 					hcmBuilder.
 						AddFilter(envoy_v3.FilterMisdirectedRequests(vh.VirtualHost.Name)).
-						DefaultFilters(). // TODO this includes the local rate limit filter, which should go after the auth filter in this case
-						AddFilter(authFilter).
+						DefaultFilters().
+						AddFilterBefore(authFilter, "local_ratelimit").
 						AddFilter(envoy_v3.OriginalIPDetectionFilter(cfg.XffNumTrustedHops)).
 						AddFilter(envoy_v3.GlobalRateLimitFilter(envoyGlobalRateLimitConfig(cfg.RateLimitConfig)))
 				} else {
--- a/internal/xdscache/v3/listener.go
+++ b/internal/xdscache/v3/listener.go
@@ -411,11 +411,8 @@ func (c *ListenerCache) OnChange(root *dag.DAG) {
 				// metrics prefix to keep compatibility with previous
 				// Contour versions since the metrics prefix will be
 				// coded into monitoring dashboards.
-				cm := envoy_v3.HTTPConnectionManagerBuilder().
+				hcmBuilder := envoy_v3.HTTPConnectionManagerBuilder().
 					Codec(envoy_v3.CodecForVersions(cfg.DefaultHTTPVersions...)).
-					AddFilter(envoy_v3.FilterMisdirectedRequests(vh.VirtualHost.Name)).
-					DefaultFilters().
-					AddFilter(authFilter).
 					RouteConfigName(path.Join("https", vh.VirtualHost.Name)).
 					MetricsPrefix(listener.Name).
 					AccessLoggers(cfg.newSecureAccessLog()).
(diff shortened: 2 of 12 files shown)


internal/xdscache/v3/listener.go
```diff
@@ -425,12 +422,25 @@ func (c *ListenerCache) OnChange(root *dag.DAG) {
...
-					AddFilter(envoy_v3.OriginalIPDetectionFilter(cfg.XffNumTrustedHops)).
-					AddFilter(envoy_v3.GlobalRateLimitFilter(envoyGlobalRateLimitConfig(cfg.RateLimitConfig))).
-					Get()
+					AllowChunkedLength(cfg.AllowChunkedLength)
+
+				if vh.AuthorizationService != nil && vh.AuthBeforeRateLimiting {
+					hcmBuilder.
+						AddFilter(envoy_v3.FilterMisdirectedRequests(vh.VirtualHost.Name)).
+						DefaultFilters().
+						AddFilterBefore(authFilter, "local_ratelimit").
```
skriss, line 431: This `AddFilterBefore` function is a little ugly, but pulling the local rate limit filter out of `DefaultFilters()` would've been painful. Open to any other ideas.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -171,6 +171,14 @@ type AuthorizationServer struct {
...
 	// +optional
 	FailOpen bool `json:"failOpen,omitempty"`
+
+	// AuthBeforeRateLimiting defines whether the authorization server
+	// will be consulted before applying any rate limiting to a
+	// request. The default is false, i.e. to apply rate-limiting first,
+	// in order to protect the authorization server.
+	//
+	// +optional
+	AuthBeforeRateLimiting bool `json:"authBeforeRateLimiting,omitempty"`
```
stevesloka, line 181: I wonder if we should make this a `pritority` level or something? This would allow additional filters to be applied in different orders when/if we add them. For instance, maybe you want the lua filters to run before auth but after rate limiting. 

This priority might just be a int which the lower priority puts it higher in the list.

stevesloka (commented):
Just a quick question about changing the bool to an int to handle future use-cases. I think change the current default around also makes sense as a first step if we wanted to just do that and come back to the configuration later on.

apis/projectcontour/v1/httpproxy.go
skriss: There was some discussion on this in the issue (https://github.com/projectcontour/contour/issues/3726#issuecomment-987204370 and following), agree it's a more general solution but also I think coming up with a good API is pretty tricky for the reasons I mentioned in the issue which is why I didn't go down that path. If you have thoughts on how it would work across all the filters, including those configured in HTTPProxy, in the config file, and not explicitly configured at all, I'm happy to hear the ideas.

skriss: Would you put `priority` fields in all the different places filters have configuration? E.g. in your example, global rate limiting is configured in the config file, so you'd need one priority there; auth is in `HTTPProxy.Spec.VirtualHost.Authorization`, so another there; and then Lua (assuming you're talking about the cookie rewrite stuff?) is in `HTTPProxy.Spec.Routes.CookieRewritePolicies`, so you'd need another priority there?

skriss: > I don't think gatwayapi solves it with filters since their order isn't determined.

Gateway API does have an explicit list of filters in the API, though, so it seems easier to solve there since filters are explicit concepts that are defined together in one place, so ordering could be defined by the user either by reordering items in the list, or by using a `priority` field on the Filter structs.

internal/xdscache/v3/listener.go
sunjayBhatia: yeah only other thing i can think of besides pulling out that filter is a bigger refactor, so that the order you call `AddFilter` doesn't matter, we maybe always have a categorized sort order for filters (that can be slightly switched based on `AuthBeforeRateLimiting` ofc), but that could be messy and inflexible later

sunjayBhatia (approved):

sunjayBhatia:
needs a `main` merge but lgtm

apis/projectcontour/v1/httpproxy.go
stevesloka: > Would you put priority fields in all the different places filters have configuration? E.g. in your example, global rate limiting is configured in the config file, so you'd need one priority there; auth is in HTTPProxy.Spec.VirtualHost.Authorization, so another there; and then Lua (assuming you're talking about the cookie rewrite stuff?) is in HTTPProxy.Spec.Routes.CookieRewritePolicies, so you'd need another priority there?

Yeah I agree this isn't great, just feels like we're immediately causing a big api issue later on by making this so specific. 

Would we need to add a `authBeforeLua` field? I guess if that was true and `authBeforeRateLimiting` was true that would be them before RateLimiting, but what order would Lua or Auth be?

skriss: > Yeah I agree this isn't great, just feels like we're immediately causing a big api issue later on by making this so specific.

Yeah, I don't love the API implemented in this PR either, but I guess my hope was that we wouldn't need to add more filter reordering options, at least not until a v2 API where maybe we can reorganize things to make it easier.

I'll think about this some more and see if I can come up with any other options.

sunjayBhatia: something else to maybe consider: https://kubernetes.slack.com/archives/C8XRH2R4J/p1639502957273600

the newer matching API that is coming out has some interesting features (match action to skip a filter, action could be to a route, etc.)

youngnick: I think that adding in filter priority is a really big API change, big enough to warrant an API rev, and I would question whether we should invest in a HTTPProxy API rev at this point, or work on making the Gateway API support these use cases instead.

I agree with @stevesloka that having this be a boolean is limiting, but I don't think that we can solve that without a really big revamp of HTTPProxy.

youngnick (approved):
This change LGTM, it feels like our choice here is between going with a quick-and-dirty fix for HTTPProxy now, or basically implementing a list of filters like Gateway API does for HTTPProxy, which is a big enough change to need an API rev (that is, `v2alpha1` etc).

I'm okay with fixing up HTTProxy for now and waiting to see what we need to do for Gateway API.

apis/projectcontour/v1/httpproxy.go
xaleeks: Echo this comment to build it as int if possible so we get the optionality to add more levers later on.  I can see applying certain transformations both before and after auth or rate limiting.  But leave this decision up to you guys

PR #4232, opened 2021-12-10
update Go version to 1.17.5

skriss:

### DIFF
--- a/.github/workflows/prbuild.yaml
+++ b/.github/workflows/prbuild.yaml
@@ -60,7 +60,7 @@ jobs:
             ${{ runner.os }}-${{ github.job }}-go-
       - uses: actions/setup-go@v2
         with:
-          go-version: '1.17.3'
+          go-version: '1.17.5'
       - name: add deps to path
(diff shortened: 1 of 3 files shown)


stevesloka (approved):

tsaarni:
@skriss Sorry! I had missed this one and created yet another PR #4238. 

I also got quite confused about the two http server implementations `net/http` and `golang.org/x/net/http`. We include both modules as dependencies. There was vulnerability CVE-2021-44716 which was fixed in both implementations.  

We run `net/http` but I think it is still good to bump `golang.org/x/net/http` as well to avoid false positives. Maybe we can keep #4238 for that.

tsaarni (approved):

tsaarni:
WDYT, should the changelog entry mention [CVE-2021-44716](https://groups.google.com/g/golang-announce/c/hcmEScgc00k) since this change mitigates that vulnerability.

xref #4237

skriss:
@tsaarni thanks, updated the relnote

PR #4236, opened 2021-12-13
set Gateway status address from Envoy LB service

### LINKED ISSUE #3828: Gateway API: set Gateway status address from Envoy LB service
To support status addresses, Contour should get the LB address of the Envoy service, similarly to how it's done for Ingress and HTTPProxy.

skriss:
Sets the Gateway's status address from the
Envoy LoadBalancer service address, similarly
to how it's done for Ingress and HTTPProxy.

Closes #3828.


cmd/contour/ingressstatus.go
```diff
@@ -80,6 +84,12 @@ func (isw *loadBalancerStatusWriter) Start(ctx context.Context) error {
 		&networking_v1.Ingress{},
 	}
 
+	// Only create Gateway informer if a controller name was provided,
+	// otherwise the API may not exist in the cluster.
+	if len(isw.gatewayControllerName) > 0 {
```
stevesloka, line 89: Would you get updates if the controller name wasn't defined?

skriss: Without this check, we'd try to set up an informer for Gateways whether the user had configured Gateway API or not, but if the API wasn't even installed in the cluster, informer setup would fail.

stevesloka: Oh that's right, there's a seperate informer which gets all the resources over again.

stevesloka (approved):

internal/k8s/statusaddress_test.go
```diff
@@ -341,6 +345,254 @@ func TestStatusAddressUpdater(t *testing.T) {
...
+	for name, tc := range testCases {
+		t.Run(name+" OnAdd", func(t *testing.T) {
+			suc := StatusUpdateCacher{}
+			assert.True(t, suc.Add(tc.preop.Name, tc.preop.Namespace, tc.preop), "unable to add object to cache")
+
+			mockCache := &mocks.Cache{}
+			mockCache.
+				On("Get", mock.Anything, client.ObjectKey{Name: string(tc.preop.Spec.GatewayClassName)}, mock.Anything).
+				Run(func(args mock.Arguments) {
+					args[2].(*gatewayapi_v1alpha2.GatewayClass).Spec.ControllerName = gatewayapi_v1alpha2.GatewayController(tc.gatewayClassControllerName)
```
sunjayBhatia, line 550: nit: maybe add a comment for what this is doing? since `Get` mutates the passed in pointer (similar to below for `OnUpdate`)

sunjayBhatia (approved):
nice!

internal/k8s/statusaddress_test.go
skriss: :+1: added

sunjayBhatia (approved):

PR #4247, opened 2021-12-17
test/scripts: Add script to deploy gcp benchmarking cluster

sunjayBhatia:
Has some parameters that can be adjusted, namely number of nodes and
machine type

Intended to deploy Contour on its own workers and Envoy/apps on other
node pool to ensure Contour can be controlled with resource limits for
more exact experiments (vs. running side by side with other workloads).


test/scripts/deploy-gcp-bench-cluster.sh
```diff
@@ -0,0 +1,41 @@
+#! /usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+CLUSTER_NAME=contour-bench
```
skriss, line 7: would it make sense to add a random component to the name to avoid conflicts?

skriss (approved):
LGTM; presumably we'll want a teardown script as well and possibly randomized names to avoid conflicts, but can add subsequently.

test/scripts/deploy-gcp-bench-cluster.sh
sunjayBhatia: yep, can add that!

was going to have this be used in a "nightly" CI run since the full tests are relatively expensive and takes a while so *hopefully* collisions won't be an issue

skriss (approved):
:+1:

stevesloka (approved):

PR #4251, opened 2021-12-20
test/e2e/bench: Add bench test suite

sunjayBhatia:
Based on #4247

Includes test for scenario from #4058

- Deploys Contour with resource limit to cluster and node selector for
contour nodes
- Deploys Envoys and test apps with node selector for app workload nodes
- Deploys apps + HTTPProxy sequentially, times time to successful
request as "ready"
- Outputs CSV of results and PNG scatter plot
- Exposes Envoy metrics port in daemonset and Contour metrics port to be used with `PodMonitor`
(see
    https://cloud.google.com/stackdriver/docs/solutions/gke/managing-metrics#configure-workload-metrics) for monitoring metrics
- Doesn't make any assertions on the results as of yet


sunjayBhatia:
have to increase the number of nodes to support the 4000 instance case, GKE has a limit of 110 pods per node

sunjayBhatia:
also considering adding a warmup to ensure the initial opening on Envoy ports doesn't interfere with the actual data collection

stevesloka:
> have to increase the number of nodes to support the 4000 instance case, GKE has a limit of 110 pods per node

Do you need pods to test? You could just create services + endpoints manually to create load, but I guess then you couldn't test the actual routing.

sunjayBhatia:
> > have to increase the number of nodes to support the 4000 instance case, GKE has a limit of 110 pods per node
> 
> Do you need pods to test? You could just create services + endpoints manually to create load, but I guess then you couldn't test the actual routing.

yeah to mirror the knative test scenario from https://github.com/projectcontour/contour/issues/4058 we would need to measure how long it takes to an app being routable

a separate test could be to measure how long till HTTPProxy "valid" which wouldn't require actual app deployments

Makefile
```diff
@@ -325,6 +325,17 @@ check-ingress-conformance: | install-contour-working run-ingress-conformance cle
...
+.PHONY: deploy-gcp-bench-cluster
+deploy-gcp-bench-cluster:
+	./test/scripts/gcp-bench-cluster.sh deploy
+
+.PHONY: run-bench
+run-bench:
+	ginkgo -tags=e2e -mod=readonly -keep-going -randomize-suites -randomize-all -slow-spec-threshold=4h -timeout=5h -r -v ./test/e2e/bench
+
+.PHONY: bench
+bench: deploy-gcp-bench-cluster run-bench
```
skriss, line 337: do we want to delete the cluster after running the test?

skriss (commented):
This looks good, have you been able to run it to completion / do you have a sample result graph that you could share?

I can't remember off the top of my head if the knative test actually ran requests against the backends or just waited for some other indicator of the config being programmed (HTTPProxy status or checking Envoy config). Relatedly, since knative does scale to 0, I don't think that their test actually ended up keeping 4000 backend pods around on the cluster. May be something to keep an eye on.

sunjayBhatia:
> have you been able to run it to completion / do you have a sample result graph that you could share?

yeah i can re-rerun to get the graph back, the full 4000 instance case is pretty slow/flaky :(

sunjayBhatia:
> since knative does scale to 0, I don't think that their test actually ended up keeping 4000 backend pods around on the cluster

yeah thats a fair point, could kick off a scale down to 0/deployment deletion async once we get success on a particular instance, which would also reduce the endpoint processing/traffic etc.

sunjayBhatia:
heres a partial graph, keep getting flakes and failures unfortunately, this one got to 860 instances

sequential-service-creation

results summary from ginkgo

```
      sequential-service-creation
      Name                     | N   | Min    | Median | Mean    | StdDev  | Max
      ===============================================================================
      time_to_ready [duration] | 860 | 36.6ms | 736ms  | 916.3ms | 743.2ms | 3.7352s
                               |     | echo-8 |        |         |         | echo-638
```

sunjayBhatia:
doing some more experimenting, adding do not merge for a bit

sunjayBhatia:
> I can't remember off the top of my head if the knative test actually ran requests against the backends or just waited for some other indicator of the config being programmed

I believe the knative "ready" status is set once their probing mechanism determines a backend is reachable

Makefile
sunjayBhatia: added that, though with the makefile mechanics if youre relying on a generated name its sorta broken, you can pass a cluster name via env var though that should work

youngnick:
I think the approach is looking great, but I think we should do the initial version with a number of pods that succeeds reliably somehow.

sunjayBhatia:
> I think the approach is looking great, but I think we should do the initial version with a number of pods that succeeds reliably somehow.

running again to see what we can go with reliably

youngnick (approved):
LGTM as long as 1000 services seems reasonably stable. Let's not let the perfect be the enemy of the good.

Makefile
```diff
@@ -326,6 +326,22 @@ check-ingress-conformance: | install-contour-working run-ingress-conformance cle
...
+teardown-gcp-bench-cluster:
+	./test/scripts/gcp-bench-cluster.sh teardown
+
+.PHONY: run-bench
+run-bench:
+	ginkgo -tags=e2e -mod=readonly -keep-going -randomize-suites -randomize-all -slow-spec-threshold=4h -timeout=5h -r -v ./test/e2e/bench
+
+.PHONY: bench
+bench: deploy-gcp-bench-cluster teardown-gcp-bench-cluster
+# bench: deploy-gcp-bench-cluster run-bench teardown-gcp-bench-cluster
```
skriss, line 343: Looks like we want L343 instead of L342 here so the test runs

test/scripts/gcp-bench-cluster.sh
```diff
@@ -5,47 +5,46 @@ set -o nounset
...
+readonly CONTOUR_BENCH_NUM_APP_POOL_NODES=${CONTOUR_BENCH_NUM_APP_POOL_NODES:-10}
 
 # TODO: If credentials available in environment, log in.
 
 function deploy() {
-  gcloud container clusters create ${CLUSTER_NAME} \
-    --cluster-version=${CLUSTER_VERSION} \
-    --zone=${CLUSTER_ZONE} \
+  num_nodes=$CONTOUR_BENCH_NUM_APP_POOL_NODES
+  ((num_nodes+=1))
```
skriss, line 18: what's this for?

skriss (changes requested):
pretty much LGTM, couple minor things.

test/scripts/gcp-bench-cluster.sh
sunjayBhatia: previously i had two node pools to separate the worker nodes from contour nodes but when you add a node pool GKE always automatically repairs your cluster so its not usable really right after its created

this way you ask for some number of worker nodes and we add 1 which contour is deployed to (to separate contour and possibly allow us to add more specific resource constraints to test it in isolation, rather than it run on a node with other variable workloads)

skriss: Ah, makes sense, a short comment here would be helpful.

skriss: :+1:

skriss (approved):

PR #4252, opened 2021-12-22
Add CRL to http proxy client validation

artificial-aidan:
Signed-off-by: Aidan Jensen <aidan@artificial.com>

Wanted to run this by the maintainers before I put the effort into implementation. Have a possible need to add a certificate revocation list to our http proxies. It gets passed in as a k8s secret just like a CA Certificate, so I would just follow that functionality.

Thoughts or objections?

### DIFF
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -929,12 +929,6 @@ type DownstreamValidation struct {
 	// +kubebuilder:validation:MinLength=1
 	CACertificate string `json:"caSecret,omitempty"`
 
-	// Name of a Kubernetes secret that contains a pem encoded crl.
-	// This will be used to verify that a client certificate has not been revoked
-	// +optional
-	// +kubebuilder:validation:MinLength=1
-	CACertificateRevocationList string `json:"crlSecret,omitempty"`
-
 	// SkipClientCertValidation disables downstream client certificate
 	// validation. Defaults to false. This field is intended to be used in
 	// conjunction with external authorization in order to enable the external
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -933,7 +933,7 @@ type DownstreamValidation struct {
 	// This will be used to verify that a client certificate has not been revoked
 	// +optional
 	// +kubebuilder:validation:MinLength=1
-	CACertificateRevocationList string `json:"caCrlSecret,omitempty"`
+	CACertificateRevocationList string `json:"crlSecret,omitempty"`
 
 	// SkipClientCertValidation disables downstream client certificate
 	// validation. Defaults to false. This field is intended to be used in
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -929,6 +929,12 @@ type DownstreamValidation struct {
 	// +kubebuilder:validation:MinLength=1
 	CACertificate string `json:"caSecret,omitempty"`
 
+	// Name of a Kubernetes secret that contains a pem encoded crl.
+	// This will be used to verify that a client certificate has not been revoked
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	CACertificateRevocationList string `json:"caCrlSecret,omitempty"`
+
 	// SkipClientCertValidation disables downstream client certificate
 	// validation. Defaults to false. This field is intended to be used in
 	// conjunction with external authorization in order to enable the external
--- a/design/downstream-crl-design.md
+++ b/design/downstream-crl-design.md
@@ -12,6 +12,7 @@ Currently Contour allows specifying a CA Certificate bundle that is used to vali
 This is done using Envoy's Common TLS Context inside of a Downstream TLS context.
 Envoy's Common TLS Context also supports a CRL data source, which is a set of CRL's that are used when validating a clients certificate.
 This CRL data source can be exposed the same way that we expose the CA Certificate data source, through Contour's `DownstreamValidation` structure.
+This data source will give admins the ability to have finer grained control over their ingress, and enable CRLs for HTTPProxies that use client cert authentication
 
 ## Goals
 - Allow a Kubernetes secret containing a PEM encoded CRL to be used when validating client connections
@@ -72,24 +73,7 @@ kubectl create secret generic crl-for-client-validation --from-file=./crl.pem
 ```go
 // DownstreamValidation defines how to verify the client certificate.
 type DownstreamValidation struct {
-	// Name of a Kubernetes secret that contains a CA certificate bundle.
-	// The client certificate must validate against the certificates in the bundle.
-	// If specified and SkipClientCertValidation is true, client certificates will
-	// be required on requests.
-	// +optional
-	// +kubebuilder:validation:MinLength=1
-	CACertificate string `json:"caSecret,omitempty"`
-
-	// SkipClientCertValidation disables downstream client certificate
-	// validation. Defaults to false. This field is intended to be used in
-	// conjunction with external authorization in order to enable the external
-	// authorization server to validate client certificates. When this field
-	// is set to true, client certificates are requested but not verified by
-	// Envoy. If CACertificate is specified, client certificates are required on
-	// requests, but not verified. If external authorization is in use, they are
-	// presented to the external authorization server.
-	// +optional
-	SkipClientCertValidation bool `json:"skipClientCertValidation"`
+  ...
 
 	// Name of a Kubernetes Opaque secret that contains a concatenated list of
   // pem encoded crls.
@@ -117,18 +101,28 @@ Validation is also performed to make sure that the CA Bundle has the correct PEM
 The same validation would be performed, but checking for type `CRL`
 
 ## Alternatives Considered
+
 N/A
 
 ## Security Considerations
-N/A
+
+Secret must be in the same namespace as the HTTPProxy, similar requirements to the CA when using client cert authentication
 
 ## Compatibility
+
 This change should be additive, so there should be no compatibility issues
 
 ## Implementation
-N/A
+
+E2E tests required:
+  * Happy path
+    * 1 CRL to one certificate
+    * 2+ CRL to 2+ certificates (same length)
+  * Errors
+    * Mismatch in CRL/Cert chain length
 
 ## Open Issues
 
+- K8s secrets are limited in size, so there will be an inherent limitation to the CRL bundle in this design
 - How to handle an error where a user supplies a CRL for one certificate in a chain, but not all. In which case Envoy will fail to verify
   - Short of decoding both the certificate list and CRL list, and making sure one exists for both, I don't have another answer. This seems out of scope for Contour here.
--- a/design/downstream-crl-design.md
+++ b/design/downstream-crl-design.md
@@ -1,6 +1,6 @@
 # Downstream Certificate Revocation List Proposal
 
-Status: Draft
+Status: Accepted
 
 ## Abstract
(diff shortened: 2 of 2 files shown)


tsaarni:
Hi @artificial-aidan 

For me, the idea sounds reasonable.

One thought: In theory CRLs could be large. Using secret as storage could reach [1MB limit](https://kubernetes.io/docs/concepts/configuration/secret/#restrictions). I guess this might happen somewhere around 20000 revoked certs.  Maybe the CRL data could also grow to be substantial part of XDS traffic sent from Contour to Envoys.  Alternative that could scale better, would be to configure file path to CRL but that has its own logistical problems, so secret does feel easier to use.  If the use case is just to revoke few client certs and we can live with that limitation, then I think secret can work.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -929,6 +929,12 @@ type DownstreamValidation struct {
 	// +kubebuilder:validation:MinLength=1
 	CACertificate string `json:"caSecret,omitempty"`
 
+	// Name of a Kubernetes secret that contains a pem encoded crl.
```
tsaarni, line 932: I guess CRL file should allow [one or more](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto#envoy-v3-api-field-extensions-transport-sockets-tls-v3-certificatevalidationcontext-crl) PEM encoded CRLs concatenated together.

artificial-aidan:
> Hi @artificial-aidan
> 
> For me, the idea sounds reasonable.
> 
> One thought: In theory CRLs could be large. Using secret as storage could reach [1MB limit](https://kubernetes.io/docs/concepts/configuration/secret/#restrictions). I guess this might happen somewhere around 20000 revoked certs. Maybe the CRL data could also grow to be substantial part of XDS traffic sent from Contour to Envoys. Alternative that could scale better, would be to configure file path to CRL but that has its own logistical problems, so secret does feel easier to use. If the use case is just to revoke few client certs and we can live with that limitation, then I think secret can work.

I think CRL's have their own set of problems, and at a large scale like that you are probably better off using a remote list and querying it somehow. If you are managing 20,000 certs as a k8s secret, I think you will have other problems as well :D

tsaarni:
> I think CRL's have their own set of problems, and at a large scale like that you are probably better off using a remote list and querying it somehow. If you are managing 20,000 certs as a k8s secret, I think you will have other problems as well :D

I agree :)

sunjayBhatia:
I would say maybe this should go through a formal design process, just because there are some details here to get right and agree on, but generally yeah agree that is a useful feature

Some details a design could address:
- specifics on how a CRL in a secret is provided (a separate secret as you have here or a field in the existing `caSecret` field for downstream validation)
- how the CRL is validated
- error handling etc. given the documentation on the Envoy field:
```
	// An optional `certificate revocation list
	// <https://en.wikipedia.org/wiki/Certificate_revocation_list>`_
	// (in PEM format). If specified, Envoy will verify that the presented peer
	// certificate has not been revoked by this CRL. If this DataSource contains
	// multiple CRLs, all of them will be used. Note that if a CRL is provided
	// for any certificate authority in a trust chain, a CRL must be provided
	// for all certificate authorities in that chain. Failure to do so will
	// result in verification failure for both revoked and unrevoked certificates
	// from that chain.
	Crl *v3.DataSource `protobuf:"bytes,7,opt,name=crl,proto3" json:"crl,omitempty"`
```

artificial-aidan:
> I would say maybe this should go through a formal design process, just because there are some details here to get right and agree on, but generally yeah agree that is a useful feature
> 
> Some details a design could address:
> 
> * specifics on how a CRL in a secret is provided (a separate secret as you have here or a field in the existing `caSecret` field for downstream validation)
> * how the CRL is validated
> * error handling etc. given the documentation on the Envoy field:
> 
> ```
> 	// An optional `certificate revocation list
> 	// <https://en.wikipedia.org/wiki/Certificate_revocation_list>`_
> 	// (in PEM format). If specified, Envoy will verify that the presented peer
> 	// certificate has not been revoked by this CRL. If this DataSource contains
> 	// multiple CRLs, all of them will be used. Note that if a CRL is provided
> 	// for any certificate authority in a trust chain, a CRL must be provided
> 	// for all certificate authorities in that chain. Failure to do so will
> 	// result in verification failure for both revoked and unrevoked certificates
> 	// from that chain.
> 	Crl *v3.DataSource `protobuf:"bytes,7,opt,name=crl,proto3" json:"crl,omitempty"`
> ```

It would also make sense to include [`only_verify_leaf_cert_crl`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto#envoy-v3-api-msg-extensions-transport-sockets-tls-v3-certificatevalidationcontext) in some form. 

I don't know if contour should do CRL validation, does it do CA validation right now? Or does it just pass it to envoy

What is the formal design process? Is that documented somewhere?

sunjayBhatia:
> What is the formal design process? Is that documented somewhere?

https://github.com/projectcontour/community/blob/main/GOVERNANCE.md#proposal-process examples of existing designs in the `design` directory!

sunjayBhatia:
> I don't know if contour should do CRL validation, does it do CA validation right now? Or does it just pass it to envoy

we do some minimal validation to ensure we don't send invalid config to envoy, heres where secret validation is done: https://github.com/projectcontour/contour/blob/main/internal/dag/secret.go

artificial-aidan:
> > I don't know if contour should do CRL validation, does it do CA validation right now? Or does it just pass it to envoy
> 
> we do some minimal validation to ensure we don't send invalid config to envoy, heres where secret validation is done: https://github.com/projectcontour/contour/blob/main/internal/dag/secret.go

Ok, seems reasonable to add a similar validation to CRL bundles, as long as it's easily done through go (assuming it is, haven't tried)

tsaarni:
> Ok, seems reasonable to add a similar validation to CRL bundles, as long as it's easily done through go (assuming it is, haven't tried)

I think for CA bundle we just parse the PEM blocks, not the actual certificates. So I think almost the same code could be used for CRLs as well, but checking for `X509 CRL` PEM block type instead of `CERTIFICATE`. There is also CRL parser in Go, if we would like to go that deep.

artificial-aidan:
Alright, took a stab at a design proposal. First time with this, so let me know if anything obvious is missing.

artificial-aidan:
Bumping this. Any thoughts on the design proposal?

sunjayBhatia (commented):
just added some comments around adding some more color to the design, but generally looks great!

design/downstream-crl-design.md
```diff
@@ -0,0 +1,128 @@
...
+  name: client-validation-example
+  namespace: default
+spec:
+  virtualhost:
+    fqdn: example.com
+    tls:
+      secretName: server-credentials
+      clientValidation:
+        caSecret: ca-cert-for-client-validation
+        caCrlSecret: crl-for-client-validation
```
tsaarni, line 46: Could the name be just `crlSecret`? The CRLs are of course published by CAs but I don't recall bumping into CRL being referred as CA CRL.

design/downstream-crl-design.md
```diff
@@ -0,0 +1,128 @@
...
+    * 1 CRL to one certificate
+    * 2+ CRL to 2+ certificates (same length)
+  * Errors
+    * Mismatch in CRL/Cert chain length
+
+## Open Issues
+
+- K8s secrets are limited in size, so there will be an inherent limitation to the CRL bundle in this design
+- How to handle an error where a user supplies a CRL for one certificate in a chain, but not all. In which case Envoy will fail to verify
+  - Short of decoding both the certificate list and CRL list, and making sure one exists for both, I don't have another answer. This seems out of scope for Contour here.
```
tsaarni, line 128: Are you referring to the case which the parameter `only_verify_leaf_cert_crl` addresses (at some level) in the future?  Yes, I agree it is outside of scope for Contour for now - the CRL check is done at level where Contour is not actively involved.

design/downstream-crl-design.md
```diff
@@ -0,0 +1,128 @@
...
+  - `internal.envoy.v3.auth.validationContext()` will be updated to accept a `crl` the same way it accepts a `ca`
+  - If empty it will not set the `crl` field, maintaining backward compatibility
+
+### Secret validation
+
+Currently basic validation is done on Opaque secrets with a `ca.crt` key, just to make sure that the length is non-zero.
+This validation would also be performed for the CRL.
+
+Validation is also performed to make sure that the CA Bundle has the correct PEM header and is of type `CERTIFICATE`.
+The same validation would be performed, but checking for type `CRL`
```
tsaarni, line 101: I think the type is `X509 CRL` https://datatracker.ietf.org/doc/html/rfc7468#section-4

tsaarni:
I've reviewed the design doc and looks good to me as well!

apis/projectcontour/v1/httpproxy.go
```diff
@@ -929,6 +929,12 @@ type DownstreamValidation struct {
...
+	// This will be used to verify that a client certificate has not been revoked
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	CACertificateRevocationList string `json:"caCrlSecret,omitempty"`
```
skriss, line 936: In light of the awkward acronyms, maybe just `crlSecret` (as you have in the design doc) would be better for the YAML, no need to include `ca` as a prefix.

design/downstream-crl-design.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Detailed Design
+
+### CRL Secret
+
+The same approach shall be followed for configuring revocation lists as is used currently to store the CA certificates for client validation
+
+The CRL is stored in an opaque Kubernetes secret.
+The secret will be stored in the same namespace as the corresponding `HTTPProxy` object.
+The secret object shall contain entry named `crl.pem`.
```
skriss, line 61: I think @sunjayBhatia mentioned this in one of his comments, but would storing this in the CA cert secret make any sense, or do you prefer to keep it separate? Keeping it separate definitely has the advantage of giving you more space to work with, but just wondering if you'd considered the alternative.

skriss (commented):
This proposal makes sense to me and the implementation looks fairly straight-forward. Just had a couple small questions/comments. Thanks @artificial-aidan!

apis/projectcontour/v1/httpproxy.go
artificial-aidan: Yep, forgot to update this

design/downstream-crl-design.md
artificial-aidan: I thought about it, but I think its an added bit of confusion. Having it be a separate secret is clear from a configuration point of view. I guess technically the user could put the same secret for both?

skriss: Fair point, as long as the validation allows it which it currently does. In any case I am fine with the separate secret.

skriss (commented):
Design doc LGTM - if you drop the API change from this PR, then we can merge just the design, otherwise can wait for the implementation

artificial-aidan:
I'll drop the API change soon. Work got busy, hoping to chip away at it next week.

artificial-aidan:
> Design doc LGTM - if you drop the API change from this PR, then we can merge just the design, otherwise can wait for the implementation

API Change dropped. Ready to merge

sunjayBhatia (approved):

sunjayBhatia:
hey @artificial-aidan sorry to get back to you late, but this looks good to me 👍🏽 any other @projectcontour/maintainers have comments?

tsaarni (approved):
Looks good to me too!  

BTW: `only_verify_leaf_cert_crl` option now became available in Envoy 1.21.

skriss:
cc @youngnick planning to merge this tomorrow unless you have any feedback on it.

youngnick (approved):
LGTM, this is a great design doc, thanks @artificial-aidan!

tsaarni:
Hi @artificial-aidan! I'd be curious to know if you have plans to take a stab at implementing the CRL design proposal? Would be cool to have the feature :)

artificial-aidan:
Priorities shifted at my day job and I have not had a chance to work on it sadly :(

tsaarni:
xref #4252

PR #4257, opened 2022-01-04
hack/release: add release notes generation tool

### LINKED ISSUE #4205: create tooling for generating release changelog from individual changelog files
As a follow-up to #4064, we need some tooling for generating a complete release changelog from the files. This could be a bash script, a go program, or something else.

skriss:
Adds a tool for generating the full release notes
for a release based on the individual files in
changelogs/unreleased plus the release notes template.

Closes #4205.


hack/release/releasenotes/make-release-notes.go
```diff
@@ -0,0 +1,145 @@
+// Copyright Project Contour Authors
```
skriss, line 1: This can probably just be folded into the existing `prepare-release.go`, no need to have it be a separate tool. Looking into that change now.

skriss: OK, second commit moves it into prepare-release.go so it can all be run in one shot.

hack/release/prepare-release.go
```diff
@@ -188,4 +199,116 @@ func main() {
...
+			Content:  strings.TrimSpace(string(contents)),
+		}
+
+		switch strings.ToLower(nameParts[2]) {
+		case "major":
+			d.Major = append(d.Major, entry)
+		case "minor":
+			d.Minor = append(d.Minor, entry)
+		case "small":
+			entry.Content = strings.TrimSpace(entry.Content)
```
sunjayBhatia, line 253: redundant based on line 244?

sunjayBhatia (commented):
should this tool also remove the `unreleased` release notes as well?

hack/release/prepare-release.go
skriss: yep, removed

hack/release/prepare-release.go
```diff
@@ -188,4 +199,116 @@ func main() {
...
+// recordContributor adds contributor to contributors if they
+// are not already present and are not a maintainer.
+func recordContributor(contributors []string, contributor string) []string {
+	for _, existing := range contributors {
+		if contributor == existing {
+			return contributors
+		}
+	}
+
+	for _, maintainer := range []string{"@skriss", "@stevesloka", "@sunjayBhatia", "@tsaarni", "@youngnick"} {
```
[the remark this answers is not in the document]
skriss: switched to a map for maintainers, though given this is just a release tool + the small volume of data this will be processing it really shouldn't matter much either way

skriss:
> should this tool also remove the `unreleased` release notes as well?

Yeah, added. Originally I didn't do that since I wasn't checking for pre-releases, and we don't want to delete the files when creating a pre-release version, but now trivial to do properly.

sunjayBhatia (approved):

stevesloka (approved):

PR #4263, opened 2022-01-06
update daily E2E run for Kubernetes 1.23

skriss:
Also adds support for running it manually.


sunjayBhatia (approved):

sunjayBhatia:
we could also maybe limit this to 1 k8s version, since esp. the knob it is modifying isn't really relevant to test diff k8s versions on

PR #4264, opened 2022-01-06
remove dependency from daily build

skriss:
Fixes error seen with daily build:
```
The workflow is not valid. .github/workflows/build_daily.yaml (Line: 14, Col: 3): The workflow must contain at least one job with no dependencies.
```

We can just run against the last built main image.


skriss:
Trivial change, power-merging.

sunjayBhatia:
whoops! thanks for fixing

PR #4265, opened 2022-01-07
Simplify daily build

sunjayBhatia:
Remove matrix strategy as current daily build doesn't really need it
(don't need to test against multiple k8s versions to test envoy xDS
server).

Also remove image build/load to speed up CI run


skriss (approved):

sunjayBhatia:
merging with powers

PR #4266, opened 2022-01-10
internal/k8s: StatusUpdateWriter only writes to update channel when leader

sunjayBhatia:
Otherwise, when not leader, update buffered channel can be filled up,
since StatusUpdateHandler does not run. This deadlocks the EventHandler.

Shows up in e2e tests in intermittent in-cluster smoke-test failures.
More prevalent with more instances so modified smoke test to test more
apps. The failure occurs when the contour instance Envoy connects to is
not the leader.

### DIFF
--- a/internal/k8s/status.go
+++ b/internal/k8s/status.go
@@ -62,6 +62,7 @@ func (m StatusMutatorFunc) Mutate(old client.Object) client.Object {
 type StatusUpdateHandler struct {
 	log           logrus.FieldLogger
 	client        client.Client
+	sendUpdates   chan struct{}
 	updateChannel chan StatusUpdate
 }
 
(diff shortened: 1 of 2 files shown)


skriss (approved):
LGTM, good catch!

internal/k8s/status.go
```diff
@@ -109,6 +111,9 @@ func (suh *StatusUpdateHandler) Start(ctx context.Context) error {
 	suh.log.Info("started status update handler")
 	defer suh.log.Info("stopped status update handler")
 
+	// Enable StatusUpdaters to start sending updates to this handler.
+	close(suh.sendUpdates)
```
stevesloka, line 115: What would happen if this instance looses the leader and a new one is elected. Would't you have 2 updates to the status now?

sunjayBhatia: if this instances loses leadership the process will shut down so only one should be running, the context governing the runnable should be cancelled

stevesloka (approved):

PR #4273, opened 2022-01-18
Bump Envoy to v1.21.0

sunjayBhatia:
See release notes:
https://www.envoyproxy.io/docs/envoy/v1.21.0/version_history/current

Also bumps go-control-plane to version matching Envoy version APIs


sunjayBhatia:
Running into a similar thing to when we bumped to v1.20.x, some new fields/types were added at the same time as existing fields were deprecated, we can't immediately switch over but will have to skip linting etc. for them and update in the next Contour release

skriss (approved):
LGTM, I guess only question is if we see any issues with jumping 2 minor Envoy versions (1.19.x -> 1.21), but I guess the upgrade tests are passing so seems fine

sunjayBhatia:
> LGTM, I guess only question is if we see any issues with jumping 2 minor Envoy versions (1.19.x -> 1.21), but I guess the upgrade tests are passing so seems fine

yeah, would be reasonable to be conservative and wait on this until contour 1.21 and stay on envoy 1.20.x for contour 1.20.x, though i dont see too much in the release notes that seems controversial

(also side note, funny we're almost matching Envoy version numbers)

internal/envoy/v3/auth.go
```diff
@@ -81,6 +81,7 @@ func validationContext(ca []byte, subjectName string, skipVerifyPeerCert bool) *
 	}
 
 	if len(subjectName) > 0 {
+		// nolint:staticcheck
```
youngnick, line 84: How will we make sure we come back and take these nolints out later? Should we have another TODO-style comment here or something?

youngnick (approved):
This LGTM, I think that jumping a couple of Envoy versions is not ideal, but hopefully this will be the only time that it happens now that we are quarterly, we should track Envoy versions from now on.

internal/envoy/v3/auth.go
sunjayBhatia: added https://github.com/projectcontour/contour/issues/4275 to the milestone and just added a note as a reminder to remove

Issue #4102, opened 2021-10-12
docs: Document default Ingress headers


stevesloka:
PR #4018 added the ability to default headers for all Ingress objects, but we need some docs to help describe the feature, how it works, how to implement, etc.

PR #4279, opened 2022-01-19
site: Add docs for header manipualtion with Ingress

sunjayBhatia:
Fixes: #4102

### DIFF
--- a/changelogs/unreleased/4279-sunjayBhatia-docs.md
+++ b/changelogs/unreleased/4279-sunjayBhatia-docs.md
@@ -1,0 +1,1 @@
+Adds documentation for header manipulation when using Ingress v1 resources.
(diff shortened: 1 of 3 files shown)


skriss (commented):
Changes LGTM, however while reviewing I couldn't find a link to this page in the TOC - am I missing it? If it's not there, we should add, otherwise I'm not sure how folks find this page.

sunjayBhatia:
> Changes LGTM, however while reviewing I couldn't find a link to this page in the TOC - am I missing it? If it's not there, we should add, otherwise I'm not sure how folks find this page.

no i think i forgot to add it when i added this page, adding now

skriss (approved):
LGTM

skriss:
Looks like this page needs an update for recently-merged #4271 too, I'll put a PR up for that.

youngnick (approved):

PR #4281, opened 2022-01-19
internal/certgen: Fix error handling when writing to k8s

sunjayBhatia:
Previously if creating a secret failed, we fell back to attempting to
update the secret, however if it didn't already exist that would also
fail, shadowing the original error.

Instead we fail fast if the failure to create a secret was not because
it already exists.


skriss (approved):
LGTM

Issue #4246, opened 2021-12-16
global rate limiting E2E test flake


skriss:
https://github.com/projectcontour/contour/runs/4553732342?check_suite_focus=true

```
Error Trace:	global_rate_limiting_test.go:110
  	            				suite.go:724
  	            				asm_amd64.s:1581
  	Error:      	Should be true
  	Test:       	HTTPProxy global rate limiting with namespace: httpproxy-global-rate-limiting-vhost-non-tls with rate limit service global rate limit policy set on non-TLS virtualhost is applied
  	Messages:   	expected 429 response code, got 200
```

skriss:
It's possible this has been fixed by https://github.com/projectcontour/contour/pull/5210, will be testing without flake retries.

PR #4284, opened 2022-01-19
hack/release: allow GitHub usernames with '-'

### LINKED ISSUE #4280: release notes tooling will break on GitHub username with '-'
Just realized that the release notes tooling that splits on '-' to parse the changelog file names will break if user names have '-' in them. https://github.com/projectcontour/contour/blob/main/hack/release/prepare-release.go#L231-L237

skriss:
Fixes a bug where GitHub usernames containing
a '-' would break the changelog generator.

Closes #4280.

Tested manually, we now have a changelog file with a hyphenated username, all looks good.

### DIFF
--- a/hack/release/prepare-release.go
+++ b/hack/release/prepare-release.go
@@ -231,6 +231,7 @@ func generateReleaseNotes(version, kubeMinVersion, kubeMaxVersion string) error 
 		entry, err := parseChangelogFilename(dirEntry.Name())
 		if err != nil {
 			fmt.Printf("Skipping changelog file: %v\n", err)
+			continue
 		}
 
 		contents, err := ioutil.ReadFile(filepath.Join("changelogs", "unreleased", dirEntry.Name()))
--- a/hack/release/prepare-release.go
+++ b/hack/release/prepare-release.go
@@ -228,11 +228,9 @@ func generateReleaseNotes(version, kubeMinVersion, kubeMaxVersion string) error 
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


youngnick (approved):
Oh wow, that will teach me to think more carefully about dashes as separators, thanks.

hack/release/prepare-release.go
```diff
@@ -228,25 +228,19 @@ func generateReleaseNotes(version, kubeMinVersion, kubeMaxVersion string) error
...
 		}
 
-		nameParts := strings.Split(strings.TrimSuffix(dirEntry.Name(), ".md"), "-")
-
-		if len(nameParts) != 3 {
-			fmt.Printf("Skipping changelog file with invalid name format %q\n", dirEntry.Name())
-			continue
+		entry, err := parseChangelogFilename(dirEntry.Name())
+		if err != nil {
+			fmt.Printf("Skipping changelog file: %v\n", err)
```
sunjayBhatia, line 233:
```suggestion
			fmt.Printf("Skipping changelog file: %v\n", err)
			continue
```

sunjayBhatia (commented):
i think just one missing `continue` but otherwise LGTM

hack/release/prepare-release.go
skriss: 🤦  good catch

sunjayBhatia (approved):

skriss:
Gonna merge this one as-is since it's definitely not causing the global rate limiting test flake. xref #4246

PR #4288, opened 2022-01-21
Further updates to getting started guide

skriss:


skriss:
OK, I'm pretty happy with this now in that it covers:
- kind cluster
- "regular" cluster with LB service support

And:
- YAML install
- Helm install
- Operator install

There are unfortunately a few variations on the steps for some of the combinations, but I tried to keep it as simple as possible. Feedback welcome!

https://deploy-preview-4288--projectcontour.netlify.app/getting-started/

site/content/getting-started/_index.md
```diff
@@ -1,237 +1,188 @@
...
+- 1+ Envoy pod(s), each with the status **Running** and 1/1 **Ready**  
+ 
+## Test it out!
 
 Congratulations, you have installed Contour and Envoy! Let's install a web application workload and get some traffic flowing to the backend.
- 
-## 3. Install a kuard workload (simple application)  
-This section installs [kuard][9] to verify web traffic is flowing to the pods through envoy.
 
+TODO is there a different sample workload then?
```
skriss, line 140: Need to resolve this; given the next line says not to expose this publicly we probably want some other workload.

youngnick: Maybe one of the echoservers or `httpbin`? It just needs to show enough to validate the thing works, not be fancy.

youngnick (approved):
This looks good to me, I agree that using Kuard is not ideal, but it's been like this for ages already, so I won't block my approval on fixing it. 😄

site/content/getting-started/_index.md
skriss: Easy enough, switched it over.

sunjayBhatia (approved):

Issue #4062, opened 2021-09-29
Pre-1.21 release: ContourConfiguration and ContourDeployment CRD review


sunjayBhatia:
This is a holding issue to document that we should do a pre-1.21 release review to ensure we are happy with the state of the new ContourConfiguration and ContourDeployment CRDs.

Steve S did some great work to get these CRDs going and we've had some in-person discussions on how we might future proof things from as they are initially implemented, restructure, follow best practices, etc. Since some of the implementation has started as well, we are getting a better sense of how the design will work in practice. Even though these CRDs are in the v1alpha1 group and we have had a design review process, we should ensure we get to as close to a final state as we can at first so we do not have to change things drastically down the line.

Things we discussed we should review include:
- kubebuilder validations (there were discussions on how to do some of these things and various people had ideas to try to propose etc.)
- in-code validation
- boolean field naming (defaulting to secure, polarity of field names e.g. "DisableFoo" vs. "EnableFoo" consistency)
- other defaulting (defaulting to secure, recommended options, etc.)
- general CRD structure, grouping of fields etc.
- etc.

My suggestion is each maintainer does a review of the API and we hold the release until we all comment with an approval

youngnick:
Great idea, thanks @sunjayBhatia. Totally agreed, this is very similar to how upstream Kubernetes does API reviews.

youngnick:
I haven't responded to this before now, but I think these are all great points and I think that we should address them. I'm not sure what to do yet though(which is why I hadn't repsonded until now). Thanks for the review @tsaarni.

sunjayBhatia:
> 2. When setting single field manually e.g. spec.envoy.metrics I got error that field spec.envoy.cluster must be set. The default were not applied anymore. I believe the reason is that the cluster defaults are defined only within field spec.envoy up in the hierarchy and not repeated again on spec.envoy.cluster. Action: All defaults need to be repeated consistently on each level of the struct hierarchy.

Having an issue with this as well creating a resource from Go rather than just YAML

b/c some fields are 'optional' but not pointers+tagged with `omitempty`, they are filled in with empty go structs rather than a `nil` so they are always supplied with zero values

e.g. a configuration struct like:

```
		contourConfig = &contour_api_v1alpha1.ContourConfiguration{
			ObjectMeta: metav1.ObjectMeta{
				Name:      "incluster-test",
				Namespace: f.Deployment.Namespace.Name,
			},
		}
```


results in the error:

```
[...]

PR #4292, opened 2022-01-24
Update ContourConfig definitions

youngnick:
* Remove enum defnitions, replace with godoc for allowed values and documentation of constants for string alias types
* Simplify defaults
* Update defaulting logic for TLS Ciphersuites
* Update servecontext_test.go tests to reduce boilerplate
* Update API Reference docs to list available constant values for string alias types

Here I'm proposing swapping everything to optional, in line with my comments in #4062. While working on that, I've cleaned up a few other things I've found, as listed above.


youngnick:
With the current spec as I write this, applying a blank ContourConfiguration object:

```yaml
kind: ContourConfiguration
metadata:
  namespace: projectcontour
  name: blank2
spec: {}
```

will get you this:
```yaml
kind: ContourConfiguration
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"projectcontour.io/v1alpha1","kind":"ContourConfiguration","metadata":{"annotations":{},"name":"blank2","namespace":"projectcontour"},"spec":{}}
  creationTimestamp: "2022-01-24T05:07:57Z"
  generation: 1
  name: blank2
  namespace: projectcontour
  resourceVersion: "461305830"
  uid: 791c3e8e-8d65-4788-9ed2-1b9e2259badb
spec:
  debug:
    kubernetesLogLevel: 0
    logLevel: info
  enableExternalNameService: false
  envoy:
    cluster:
      dnsLookupFamily: auto
    defaultHTTPVersions:
    - HTTP/1.1
    - HTTP/2
    health:
      address: 0.0.0.0
      port: 8002
    http:
      accessLog: /dev/stdout
      address: 0.0.0.0
      port: 8080
    https:
      accessLog: /dev/stdout
      address: 0.0.0.0
      port: 8443
    listener:
      tls:
        minimumProtocolVersion: "1.2"
      useProxyProtocol: false
    logging:
      accessLogFormat: envoy
    metrics:
      address: 0.0.0.0
      port: 8002
    network:
      adminPort: 9001
    service:
      name: envoy
      namespace: projectcontour
  health:
    address: 0.0.0.0
    port: 8002
  httpproxy:
    disablePermitInsecure: false
  metrics:
    address: 0.0.0.0
    port: 8002
  xdsServer:
    address: 0.0.0.0
    port: 8001
    tls:
      caFile: /certs/ca.crt
      certFile: /certs/tls.crt
      insecure: false
      keyFile: /certs/tls.key
    type: contour
```

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -223,13 +274,22 @@ type EnvoyConfig struct {
...
-	// strings of the form "HTTP/xx". Supported versions are
+	// strings of the form "HTTP/xx".
 	// "HTTP/1.1" and "HTTP/2".
-	DefaultHTTPVersions []HTTPVersionType `json:"defaultHTTPVersions"`
+	//
+	// Values: `HTTP/1.1`, `HTTP/2`
+	//
+	// Other values will produce an error.
+	// The default includes both values.
+	// +kubebuilder:default="HTTP/1.1";"HTTP/2"
```
skriss, line 290: Seems like one that'll change over time, how do we handle?

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -287,26 +361,59 @@ type EnvoyListenerConfig struct {
...
+	// Values: (empty string): use the default ConnectionBalancer, `exact`: use the Exact ConnectionBalancer
+	//
+	// Other values will produce an error.
+	// +optional
+	ConnectionBalancer string `json:"connectionBalancer,omitempty"`
 
 	// TLS holds various configurable Envoy TLS listener values.
-	TLS EnvoyTLS `json:"tls"`
+	// +optional
+	// +kubebuilder:default={minimumProtocolVersion: "1.2"}
```
skriss, line 379: another that'll change over time, how to handle?

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -287,26 +361,59 @@ type EnvoyListenerConfig struct {
...
-	// negotiate. Valid options are `1.2` (default) and `1.3`.
-	// +kubebuilder:validation:Enum="1.2";"1.3"
+	// negotiate.
+	//
+	// Values:
+	// * `1.2` (default)
+	// * `1.3`
+	//
+	// Other values will produce an error.
+	// +kubebuilder:default="1.2"
```
skriss, line 416: another that'll change over time

skriss (commented):
Added a few specific comments but probably due for an in-person discussion

apis/projectcontour/v1alpha1/contourconfig.go
youngnick: This one is probably worth looking at the docstring again and making sure we explain that you *can* choose 1.1, but we don't turn it on by default.

youngnick:
In our maintainer's meeting today, we discussed this, @skriss is going to summarize on #4062. I'll leave this as is for now.

sunjayBhatia:
> In our maintainer's meeting today, we discussed this, @skriss is going to summarize on #4062. I'll leave this as is for now.

also a TODO item related to this is to write an e2e test to validate kubebuilder defaults, this will ensure if defaults change in a PR, we have a signal and can write a release note etc.

trick will be to make sure new fields are added to this test over time, to ensure we maintain coverage

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -75,31 +74,48 @@ type ContourConfigurationSpec struct {
...
 	// Metrics defines the endpoint Contour uses to serve metrics.
 	// +optional
-	// +kubebuilder:default={address: "0.0.0.0", port: 8000}
-	Metrics MetricsConfig `json:"metrics"`
+	// +kubebuilder:default={port: 8002}
+	Metrics *MetricsConfig `json:"metrics"`
 }
 
 // XDSServerType is the type of xDS server implementation.
 type XDSServerType string
```
sunjayBhatia, line 82: for this sort of thing I also see `+enum` as an annotation (e.g. [here](https://github.com/kubernetes/api/blob/e6d62ddcb18470ddd34d9c4cd0e3178ab58cfaeb/core/v1/types.go#L2610)), which is described [here](https://github.com/kubernetes/enhancements/issues/2887), not sure if we care to do this or not, might make things nice for downstream consumers? (but also not clear to me if this locks you into the same thing as kubebuilder enum, that adding values is a major change)

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -539,7 +673,7 @@ type ContourConfigurationStatus struct {
 	// +patchStrategy=merge
 	// +listType=map
 	// +listMapKey=type
-	Conditions []contour_api_v1.DetailedCondition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
```
sunjayBhatia, line 676: just wanted to open a thread on this one, to see what the reasoning behind this change was? the standard `metav1.Condition` list may be sufficient?

youngnick: Ah, thanks for catching this, I had intended to put a comment in here.

I think DetailedCondition was a mistake on my part. Unfixable in the v1 APIs, but we aren't really using the `Errors` and `Warnings` fields in DetailedCondition, so I think we should move towards using the upstream struct instead.

apis/projectcontour/v1alpha1/contourconfig.go
youngnick: The argument against using Enums is that Clients can (and should) assume that they know all possible values of the enum, so adding or removing fields can break clients unless it's made very clear that the enum could be changed. Making the type an aliased string leaves the door open for other values than the list at any particular version.

It feels to me like this applies no matter how the enum is specified.

Does this really apply here? I think so, but I suppose that a lot of these fields are unlikely to have further values added. I think XDSServerType is a good example of one where we could conceivably add further values (like maybe `incremental` or something) as we add further functionality and want to have a chance to get it in people's hands before it's as stable as the rest.

I agree that having to document this stuff manually sucks, but this feels to me like another time when the rule has been written in upstream maintainer pain.

apis/projectcontour/v1alpha1/contourconfig.go
youngnick: This could (and probably should) be a separate PR though.

youngnick:
With conversation still ongoing in #4062, I'll leave this alone for now. Once we have more consensus there, I'll come back here and update.

youngnick:
TODO list here, from #4062 
- [ ] Remove all the kubebuilder defaults from the ContourConfig CRD. Not validation, just default values. Everything should still be optional though.
- [ ] Replace them with Contour-managed defaults that are documented in the relevant godoc, so that it shows up on the website and in kubectl describe.
- [ ] Provide an example YAML for ContourConfig with everything commented out.

youngnick:
Superseded by #4451

PR #4293, opened 2022-01-24
test/e2e/incluster: Dump Contour pod logs after each test

sunjayBhatia:
To help debug flakes and failures in CI

Logs will only be output on failures since we don't use Ginkgo verbose
mode.


sunjayBhatia:
re-running to see if my fix is robust enough

sunjayBhatia:
still issues with pods being deleted simultaneously with log collection, might consider ignoring errors and making it best-effort

sunjayBhatia:
ready for review, relaxed the assertions and made it so we wait for pods to go away between tests

youngnick (approved):
LGTM, nice change!

skriss (approved):

PR #4303, opened 2022-01-27
update netlify redirects for v1.20.0

skriss:

cc @tsaarni @OrlinVasilev 

This is the output of "Update quickstart YAML redirects" under https://projectcontour.io/resources/release-process/#minor-release-process. We usually do this as a separate PR from the website/docs that's merged *after* the release branches are created, so the redirects are always valid.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -50,7 +50,7 @@ HUGO_ENABLEGITINFO = "true"
 # kubectl apply https://projectcontour.io/quickstart/contour.yaml
 [[redirects]]
   from = "/quickstart/contour.yaml"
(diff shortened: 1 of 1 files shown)


tsaarni (approved):
LGTM
> We usually do this as a separate PR from the website/docs that's merged after the release branches are created, so the redirects are always valid.

Makes sense. Netlify will automatically rebuild the site when new commits land on main?

skriss:
> LGTM
> 
> > We usually do this as a separate PR from the website/docs that's merged after the release branches are created, so the redirects are always valid.
> 
> Makes sense. Netlify will automatically rebuild the site when new commits land on main?

Yep, that's right, it's integrated with the GitHub repo.

OrlinVasilev (approved):
LGTM

PR #4304, opened 2022-01-27
internal/dag: don't check annotations for uncached objects

### LINKED ISSUE #4239: Error messages related to ignored Ingresses.
When contour is runned with a specific ingress class (----ingress-class-name=something) contour shall ignore all ingresses not annotated or marked with the same ingress class.

If any ingress resource has an annotation that is not supported by contour, it will show an error message about the annotation being ignored, for example:
ERRO[0000] ignoring invalid or unsupported annotation    annotation=ingress.kubernetes.io/body-size.

The problem is: this error message is logged for **any** ingress resources, not just the ones that are having the same ingress class.
This means contour does use resources to check these annotations on "throwaway" ingress resources, and also logs these errors needlessly. 
In debug logs it is shown that the resource is ignored but only after the error message:
DEBU[0000] ignoring object with unmatched ingress class.
This happens every time an ingress resource is created, or modified in the whole cluster.

This is true for ingress v1 and ingress v1beta1 resources, although it is curious that in both cases the log says
version=networking.k8s.io/v1

Replication steps:
Install contour with a random ingress class to look for.
Install or change any Ingress resource, that has an annotation that contour does not handle: 
For example: 
ingress.kubernetes.io/ssl-redirect: "true"
ingress.kubernetes.io/ssl-redirect: "true"

Example ingress manifest included.
[ingress.txt](https://github.com/projectcontour/contour/files/7713166/ingress.txt)

I was expecting the ingresses not using the right ingress class not to be processed at all.

**Environment:**

- Contour version: 1.19.1 but seen in 1.15.2 too
- Kubernetes installer & version:  v1.20.4

skriss:
In the DAG cache, only check for invalid annotations
on objects that are being cached, rather than all
objects.

Closes #4239.


youngnick (approved):
LGTM, this is a nice optimization.

sunjayBhatia (approved):
LGTM, yeah much nicer since we wont get extraneous logs, also since it was a warning anyway totally fine to change the order 👍🏽

PR #4310, opened 2022-01-28
hack/actions: fix release image tag fetching

### LINKED ISSUE #4309: "build and push a release" workflow is broken
https://github.com/projectcontour/contour/runs/4983758966?check_suite_focus=true

Our v1.20.0 build failed. We've switched from `actions/checkout@v1` to `actions/checkout@v2` recently, and it seems like the checkout behavior has changed such that it now checks out the tag itself, which then leads to an error when our release script [tries to fetch all tags](https://github.com/projectcontour/contour/blob/main/hack/actions/build-and-push-release-images.sh#L31).

skriss:
Adds `--force` to `git fetch --tags` to
work with the v2 checkout action.

Closes #4309.

I was able to repro the error locally by running through the relevant git commands, and to confirm that with this change, the error no longer occurs. Won't be able to exercise the full workflow without pushing a dummy tag, though.

### DIFF
--- a/hack/actions/build-and-push-release-images.sh
+++ b/hack/actions/build-and-push-release-images.sh
@@ -28,7 +28,7 @@ fi
 
 # Fetch all tags so we can check if the current tag
 # is the highest semver.
-git fetch --tags
+git fetch --tags --force
 
 HIGHEST_SEMVER_TAG=""
 


sunjayBhatia (approved):

skriss:
Upgrade tests are going to fail for now since we have a v1.20.0 tag but no corresponding image.

youngnick (approved):

PR #4312, opened 2022-01-28
cmd/contour: Disable unused listeners

sunjayBhatia:
controller-runtime manager provided listeners for metrics and health
(both are implemented via our own mechanisms, health was already
disabled but this explicitly does so)

Helps with e2e tests especially, ensure Contour can start up and doesn't
conflict on binding to the same port (e.g. default metrics port 8080)
if a Contour instance does not exit properly.


skriss (approved):

sunjayBhatia:
upgrade tests failing due to missing 1.20.0 image

PR #4316, opened 2022-01-31
internal/dag: removed duplicated function for secret validation

### LINKED ISSUE #3503: Secret validation duplicated in cache & dag layers
We sanitize secrets when they arrive at the cache layer (https://github.com/projectcontour/contour/blob/main/internal/dag/secret.go#L34:6), however, that logic is duplicated in the dag package (https://github.com/projectcontour/contour/blob/main/internal/dag/accessors.go#L225). 

I theory we should be able to rely on the secrets being valid from the cache bit, but in any case should centralize this logic so we don't get differing results on what is valid or not.

alessandroargentieri:
internal/dag: Removed duplicated function for secret validation in dag layer

Removed validSecret function from accessors.go in the dag package.
This function checked if a secret is a valid TLS secret and returned an error.
Substituted with the validTLSSecret in secret.go in the dag package,
function that wraps the isValidSecret function by adding some logic on it.
This way the duplication of code is avoided and the feature of the deleted function is maintained.

Fixes #3503

Signed-off-by: alessandroargentieri <alexmawashi87@gmail.com>

### DIFF
--- a/internal/dag/secret.go
+++ b/internal/dag/secret.go
@@ -27,6 +27,15 @@ import (
 // CACertificateKey is the key name for accessing TLS CA certificate bundles in Kubernetes Secrets.
 const CACertificateKey = "ca.crt"
 
+// validTLSSecret returns an error if the Secret is not of type TLS or if it doesn't contain certificate and private key material.
+func validTLSSecret(s *v1.Secret) error {
+	if s.Type != v1.SecretTypeTLS {
+		return fmt.Errorf("secret type is not %q", v1.SecretTypeTLS)
+	}
+	_, err := isValidSecret(s)
+	return err
+}
+
 // isValidSecret returns true if the secret is interesting and well
 // formed. TLS certificate/key pairs must be secrets of type
 // "kubernetes.io/tls". Certificate bundles may be "kubernetes.io/tls"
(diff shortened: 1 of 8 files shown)


internal/dag/secret.go
```diff
@@ -27,6 +27,15 @@ import (
 // CACertificateKey is the key name for accessing TLS CA certificate bundles in Kubernetes Secrets.
 const CACertificateKey = "ca.crt"
 
+// validTLSSecret returns true if the Secret contains certificate and private key material.
```
skriss, line 30:
```suggestion
// validTLSSecret returns true if the Secret is of type TLS and contains certificate and private key material.
```

skriss (changes requested):
@alessandroargentieri thanks for this! Looks pretty good to me, just one tiny nit on the godoc. Also, please add a file `changelogs/unreleased/4316-alessandroargentieri-small.md` with a brief changelog entry mentioning the refactoring so you get credit in the next release's release notes :) There are a couple examples in that directory that you can refer to.

internal/dag/secret.go
alessandroargentieri: @skriss I've applied the changes, hope they're good

skriss (approved):
thanks for the updates, LGTM!

youngnick (approved):
LGTM, nice work and thanks @alessandroargentieri!

PR #4323, opened 2022-02-05
Sanitize dot-graph output

### LINKED ISSUE #4328: Dot graph exporter `debug/dag/` does not escape label values
`CONTOUR_POD=$(kubectl -n envoyweb get pod -l app=contour -o jsonpath='{.items[0].metadata.name}')`  
`kubectl -n envoyweb port-forward $CONTOUR_POD 6060`  
`curl -s localhost:6060/debug/dag | dot`
> Error: <stdin>: syntax error in line 5 near ':'

I found there's special characters (`"`, `<`, `>`) in label values carried over from our listener config.

Output from the Graphql `dot` rendering tool should help visualize the contour DAG.

**Environment:**

- Contour version: Fork of v0.12.1
- `dot` version: dot - graphviz version 2.50.0 (20211204.2007)

kb000:
This change html-escapes text fields in the `/debug/dag` `dot` graph output.

The dot-graph output of contour's dag state can allow HTML-unsafe characters, which leads to failures rendering the graph in the `dot` utility.   (I found this issue in my enterprise's fork, where the issue is more severe, but I figured I'd upstream the fix, too.)

Added unit tests and refactored a bit to allow unit testing.

Fixes #4328


kb000:
Addresses https://github.com/projectcontour/contour/issues/4328

skriss:
Thanks for the PR @kb000! Will review soon.

internal/dag/builder.go
```diff
@@ -35,6 +35,11 @@ func (pf ProcessorFunc) Run(dag *DAG, source *KubernetesCache) {
 	}
 }
 
+type BuilderType interface {
+	// Build builds and returns a new DAG
+	Build() *DAG
+}
+
```
skriss, line 42: I might move this into `internal/debug` where it's required, and then can rename to `DAGBuilder` or similar.

internal/debug/dot.go
```diff
@@ -24,21 +25,33 @@ import (
...
 
-	nodes := map[interface{}]bool{}
-	edges := map[pair]bool{}
+	printNodes(nodes, w)
+	printEdges(edges, w)
+
+	fmt.Fprintln(w, "}")
+}
+
+func collectDag(b dag.BuilderType) (nodes nodeCollection, edges edgeCollection) {
```
skriss, line 49: we generally don't use named return parameters, let's drop the names here and declare them below.

skriss (changes requested):
This looks great to me @kb000, just a couple small nits and I think we can get it in. Thanks again for the PR!

internal/dag/builder.go
kb000: Great suggestion. Duck interfacing blowing my mind here.

skriss (approved):

youngnick (approved):
LGTM, I think that the use of `mockery` is new, but maybe we should consider it for more places?

sunjayBhatia:
> LGTM, I think that the use of `mockery` is new, but maybe we should consider it for more places?

Yeah we added it recently and this was nice to see additional usage from a contributor 🤙🏽

PR #4325, opened 2022-02-07
Add query parameter hash policy

### LINKED ISSUE #4319: Add query parameter hash policy

Right now `projectcontour` implements routing/lb policies based on:
- cookie
- header
- sourceIP

Envoy supports hash policy on query parameters, [here](https://www.envoyproxy.io/docs/envoy/v1.17.0/api-v3/config/route/v3/route_components.proto#config-route-v3-routeaction-hashpolicy-queryparameter)
It would be nice to have it, as some services out in the wild use all types of identity mappings, not only header-based.

pkit:
fixes #4319

fixed some small typos in docs too


site/content/docs/main/config/request-routing.md
```diff
@@ -283,15 +284,64 @@ spec:
         terminal: true
       - headerHashOptions:
           headerName: User-Agent
+```
+In this example, if a client request contains the `X-Some-Header` header, the value of the header will be hashed and used to route to an upstream Endpoint. This could be used to implement a similar workflow to cookie-based session affinity by passing a consistent value for this header. If it is present, because it is set as a `terminal` hash option, Envoy will not continue on to process to `User-Agent` header or source IP to calculate a hash. If `X-Some-Header` is not present, Envoy will use the `User-Agent` header value to make a routing decision along with the source IP of the client making the request. These policies can be used alone or as shown for an advanced routing decision.
+
```
pkit, line 289: From envoy docs and the actual code behavior here I assume that you cannot ever use multiple hash strategies in one route, therefore docs here were altered to show that only one of `headerHashOptions`, `hashSourceIP` or `queryParameterHashOptions` can be configured. Correct me if I'm wrong.

sunjayBhatia: doesn't mean we have to undo what you have here, but might be good to add a little blurb saying they can be used together

sunjayBhatia:
@pkit looks like you might have to regenerate some YAML for this change, let us know if you need anything here

also could you add a release note, following https://github.com/projectcontour/contour/blob/main/CONTRIBUTING.md#commit-message-and-pr-guidelines

thanks!

will continue with a more detailed review as well

apis/projectcontour/v1/httpproxy.go
```diff
@@ -864,6 +864,17 @@ type HeaderHashOptions struct {
...
 
+// QueryParameterHashOptions contains options to configure a query parameter based hash
+// policy, used in request attribute hash based load balancing.
+type QueryParameterHashOptions struct {
+	// QueryParameterName is the name of the HTTP request query parameter that will be used to
+	// calculate the hash key. If the query parameter specified is not present on a
+	// request, no hash will be produced.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	QueryParameterName string `json:"queryParameterName,omitempty"`
```
sunjayBhatia, line 875: maybe simplify to just `ParameterName` and json field `parameterName`? a little more succinct (in retrospect `HeaderName` in `HeaderHashOptions` could have just been `Name`)

changelogs/unreleased/4325-pkit-minor.md
```diff
@@ -0,0 +1,5 @@
+## Query parameter hash based load balancing
+
+Contour users can now configure their load balancing policies on `HTTPProxy` resources to hash the query parameter on a request to ensure consistent routing to a backend service instance.
+
+See [this page](https://projectcontour.io/docs/v1.20.0/config/request-routing/#load-balancing-strategy) for more details on this feature.
```
sunjayBhatia, line 5:
```suggestion
See [this page](https://projectcontour.io/docs/v1.21.0/config/request-routing/#load-balancing-strategy) for more details on this feature.
```

since this will be out in that release

internal/dag/policy.go
```diff
@@ -665,16 +665,20 @@ func loadBalancerRequestHashPolicies(lbp *contour_api_v1.LoadBalancerPolicy, val
...
+		queryParameterHashPolicies := map[string]bool{}
 		for _, hashPolicy := range lbp.RequestHashPolicies {
 			rhp := RequestHashPolicy{
 				Terminal: hashPolicy.Terminal,
 			}
 
 			// Ensure hashing for exactly one request attribute is set.
-			if (!hashPolicy.HashSourceIP && hashPolicy.HeaderHashOptions == nil) ||
-				(hashPolicy.HashSourceIP && hashPolicy.HeaderHashOptions != nil) {
+			if (!hashPolicy.HashSourceIP && hashPolicy.HeaderHashOptions == nil && hashPolicy.QueryParameterHashOptions == nil) ||
```
sunjayBhatia, line 676: another way to skin this is to have a counter for number of options set and individually check if each option is set, incrementing if so

if counter is 0 or greater than 1, then write the warning

we will probably have to do something like that if we add any more options and to simplify what we already have

internal/dag/policy.go
```diff
@@ -665,16 +665,20 @@ func loadBalancerRequestHashPolicies(lbp *contour_api_v1.LoadBalancerPolicy, val
 		hashSourceIPSet := false
 		// Map of unique header names.
 		headerHashPolicies := map[string]bool{}
+		// Map of unique query parameter names.
+		queryParameterHashPolicies := map[string]bool{}
```
sunjayBhatia, line 669: nit: could make this and `headerHashPolicies` a `map[string]struct{}{}` so the value takes up no space, don't know why i didn't do that to start with

youngnick:
Hi @pkit, small bump here in case this has fallen off your radar.

pkit:
Just got out of Ukraine.
You know, wrong place, wrong time.
Will continue working on that one as soon as I get to some kind of workplace.

youngnick:
Oh wow. Glad you are okay!

This is not that important, we will keep it open for you until you can get to it.

site/content/docs/main/config/request-routing.md
pkit: I'm not sure how it works in the field. I've tried to configure both "header" and "sourceIP" policy for a route and the result was: no hash.

apis/projectcontour/v1/httpproxy.go
pkit: Done

internal/dag/policy.go
pkit: Done

site/content/docs/main/config/request-routing.md
sunjayBhatia: hey @pkit had a quick question on this if you had a chance to check it out

sorry for the delay reviewing, everything largely looks good

sunjayBhatia:
hey all, fixing this up in another PR to make sure it gets in to the release, will give @pkit credit of course!

sunjayBhatia:
closing as https://github.com/projectcontour/contour/pull/4508 is merged

PR #4326, opened 2022-02-07
Added configurable upstream connect timeout

### LINKED ISSUE #2264: Unable to change connect_timeout in Envoy config
Running into the Envoy issue reported [here](https://github.com/envoyproxy/envoy/issues/3492)

Based on this [comment](https://github.com/envoyproxy/envoy/issues/3492#issuecomment-419906021), I'd like to update the connect_timeout setting in the envoy.json config file. However, due to the way Contour bootstraps Envoy, this isn't easily configurable. 

There is another issue open pertaining to Envoy defaults [here](https://github.com/projectcontour/contour/issues/1375) that would have the default increased to 15 seconds from 250ms.

There's also another issue pertaining to the configurability of these knobs [here](https://github.com/projectcontour/contour/issues/2225).

**Environment:**

- Contour version: 1.1.0
- Kubernetes version: (use `kubectl version`): 1.17.2
- Kubernetes installer & version: kubeadm
- Cloud provider or hardware configuration: GCP
- OS (e.g. from `/etc/os-release`): Ubuntu 18.04

tsaarni:
This change adds optional configurable timeout for upstream TCP connections.  Currently Envoy is programmed with a hardcoded value of 2 seconds. With this change, user will be able to change this global timeout by setting a variable in the config file or `ContourConfiguration`.

Fixes #2264


pkg/config/parameters.go
```diff
@@ -463,6 +471,10 @@ func (t TimeoutParameters) Validate() error {
 		return fmt.Errorf("connection shutdown grace period %q: %w", t.ConnectionShutdownGracePeriod, err)
 	}
 
+	if err := v(t.ConnectTimeout); err != nil {
```
sunjayBhatia, line 474: using `v()` here will allow `infinity` as a valid timeout, which we probably don't want, as you mentioned elsewhere

tsaarni: Missed that, thanks @sunjayBhatia! Now fixed.

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -412,6 +412,14 @@ type TimeoutParameters struct {
 	// for more information.
 	// +optional
 	ConnectionShutdownGracePeriod *string `json:"connectionShutdownGracePeriod,omitempty"`
+
+	// ConnectionTimeout defines how long the proxy should wait when establishing connection to upstream service.
```
skriss, line 416: nit:
```suggestion
	// ConnectTimeout defines how long the proxy should wait when establishing connection to upstream service.
```

pkg/config/parameters.go
```diff
@@ -422,6 +422,14 @@ type TimeoutParameters struct {
 	// See https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-drain-timeout
 	// for more information.
 	ConnectionShutdownGracePeriod string `yaml:"connection-shutdown-grace-period,omitempty"`
+
+	// ConnectionTimeout defines how long the proxy should wait when establishing connection to upstream service.
```
skriss, line 426:
```suggestion
	// ConnectTimeout defines how long the proxy should wait when establishing connection to upstream service.
```

skriss (changes requested):
Looks reasonable to me, just a couple godoc nits

internal/envoy/v3/cluster_test.go
```diff
@@ -508,6 +508,22 @@ func TestCluster(t *testing.T) {
 				),
 			},
 		},
+		"cluster with connect timeout set": {
+			cluster: &dag.Cluster{
+				Upstream:       service(s1),
+				ConnectTimeout: 2 * time.Second,
```
sunjayBhatia, line 514: nit: since the test merges the want and default cluster config, maybe use a different value to ensure theres no chance a mistake in implementation still makes this test pass

skriss (approved):
LGTM pending Sunjay's comment

youngnick:
Let's bring this in without a kubebuilder default for now, and I'll come back to what we can do there  next week.

internal/envoy/v3/cluster_test.go
tsaarni: Good catch, changed to 10 seconds.

tsaarni:
In addition to fixing @sunjayBhatia's comment I've also added connect timeout to `contourconfiguration_test.go` for coverage improvement.

youngnick (approved):

sunjayBhatia (approved):

PR #4331, opened 2022-02-09
Add configurable log level for access log.

### LINKED ISSUE #4321: Less verbose access logging.
Currently Contour sets up envoy to log every request to /dev/stdout or alternatively if
--envoy-http-access-log=/dev/null
--envoy-https-access-log=/dev/null
parameters are used the access logs can be turned off completely.
The logs can help us debugging, although on a busy system the amount of generated logs can be a big task to log collecting subsystems

It would be great to have a solution between fully on and fully off. For example an error level logging of access: only logging the requests that are not successful. This could be a good middle ground.
This can be achieved with envoys own access_log_filters 

I was able to have some filtering with 
              filter:
                status_code_filter:
                  comparison:
                    op: GE
                    value:
                      default_value : 300
                      runtime_key: 'something'
in envoys access_log configuration.
Alternatively the response_flag_filter could be useful also, although I was unable to create a filter that trigger to everything only leaving 200 - lines out.

It would be great to have this feature be parameterized in configmap for exmaple:
by default: no filtering
<operand: EQ LE GE>:<response code>

tsaarni:
This proposed change introduces new config file field called `accesslog-level` and corresponding CRD field `ContourConfiguration.spec.envoy.logging.accessLogLevel`. The fields are used to set verbosity level for the access logs. 

The logging levels are:
* `info`: every request is logged. This is the default value and corresponds to the existing behavior.
* `error`: log only errors. In practice it means requests with status code equal or greater than `300` and requests that have [envoy response flag](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage#config-access-log-format-response-flags) set. Filtering is done by using [`AccessLogFilter`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/accesslog/v3/accesslog.proto#config-accesslog-v3-accesslogfilter) feature in access logger.
* `disabled`: do not log requests at all.

The motivation for this feature is to provide way for user to reduce the number of logs - to avoid overwhelming the centralized log collection system they might have.

Fixes #4321.


changelogs/unreleased/4331-tsaarni-minor.md
```diff
@@ -0,0 +1,4 @@
+## Configurable access log level
+
+The verbosity of HTTP and HTTPS access logs can now be configured to one of: `info` (default), `error`, `disabled`.
+The verbosity level is set with `accesslog-level` field in the [configuration file](https://projectcontour.io/docs/main/configuration/#configuration-file).
```
youngnick, line 4: Maybe mention ContourConfig here?

youngnick (approved):
LGTM with one minor comment.

changelogs/unreleased/4331-tsaarni-minor.md
tsaarni: Missed that, thanks! Added now.

Issue #1393, opened 2019-08-27
Move leader election to use the new Lease object


youngnick:
As of Kubernetes 1.15, the new Lease object is in beta, and can be used as the lock for leader election.

Once the Lease object is in beta in all our supported versions of Kubernetes (ie in the 1.17 or 1.18 timeframe), we should move the leader election code to use it.

This will mean that Contour won't need Update permissions on the leader election configmap (which is currently the `config` ConfigMap in `heptio-contour`).

sunjayBhatia:
we've started doing this as part of moving to using more controller-runtime managed runnables/leader election

release 1.21.0 can remove use of the configmap for leader election

skriss:
Bumping this to 1.21 for the remaining work.

Issue #4258, opened 2022-01-04
add a "deprecation" changelog type


skriss:
We should add a "deprecation" changelog type, for calling out any deprecations or removals that users need to be aware of. Including these as changelog files as part of the relevant PR makes it more likely that it'll be remembered, and therefore included in the release changelog.

I could see these being in addition to a regular changelog (i.e. one PR could have a major/minor/small changelog plus a deprecation changelog), since one PR may have a noteworthy change in addition to a deprecation.

PR #4332, opened 2022-02-09
Leader election only uses Lease object

sunjayBhatia:
- Removes RBAC for ConfigMaps
- Also removes mentions of leader election ConfigMaps from examples
- Updates e2e tests

Fixes: #1393

### DIFF
--- a/changelogs/unreleased/4332-sunjayBhatia-minor.md
+++ b/changelogs/unreleased/4332-sunjayBhatia-minor.md
@@ -1,0 +1,6 @@
+## Leader election now only uses Lease object
+
+Contour now only uses the Lease object to coordinate leader election.
+RBAC in example manifests has been updated accordingly.
+
+**Note:** Upgrading to this version of Contour will explicitly require you to upgrade to Contour v1.20.0 *first* to ensure proper migration of leader election coordination resources.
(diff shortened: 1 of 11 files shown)


sunjayBhatia:
the config file still has leaderelection configuration here: https://projectcontour.io/docs/main/configuration/#leader-election-configuration

and it mentions a configmap

we note that it is "deprecated" but when can we remove that?

sunjayBhatia:
will need to also update the operator once this merges

skriss:
> the config file still has leaderelection configuration here: https://projectcontour.io/docs/main/configuration/#leader-election-configuration
> 
> and it mentions a configmap
> 
> we noted last release that it is "deprecated" but when can we remove the option entirely?

https://projectcontour.io/resources/deprecation-policy/ does have some general guidance ("choose your own timeline"), in this case IMHO it's fine to drop in 1.21 as long as we explicitly call it out in the release notes. Maybe a good time to get #4258 going so we don't forget to call it out?

sunjayBhatia:
> > the config file still has leaderelection configuration here: https://projectcontour.io/docs/main/configuration/#leader-election-configuration
> > and it mentions a configmap
> > we noted last release that it is "deprecated" but when can we remove the option entirely?
> 
> https://projectcontour.io/resources/deprecation-policy/ does have some general guidance ("choose your own timeline"), in this case IMHO it's fine to drop in 1.21 as long as we explicitly call it out in the release notes. Maybe a good time to get #4258 going so we don't forget to call it out?

yeah https://github.com/projectcontour/contour/releases/tag/v1.20.0 calls it out as deprecated, though I didn't add a timeline in that release note

I'll work on the deprecations/removals release note tag, thats good to have

skriss:
I'll fold the RBAC update into https://github.com/projectcontour/contour-operator/pull/487 as well so we don't need another PR

skriss (approved):
LGTM but fine if you also want to drop the config file fields in this PR

sunjayBhatia:
> LGTM but fine if you also want to drop the config file fields in this PR

I'll do that in a separate PR once https://github.com/projectcontour/contour/pull/4333 is merged to ensure it all works

PR #4333, opened 2022-02-09
release-note/deprecation PR label validated by tooling

sunjayBhatia:
Can be used alone or in conjunction with another label.

Also makes it so release-note and release-note-action-required legacy
labels have to be alone, for simplicity.

Fixes: #4258

### DIFF
--- a/.github/workflows/label_check.yaml
+++ b/.github/workflows/label_check.yaml
@@ -9,15 +9,31 @@ env:
   GOPROXY: https://proxy.golang.org/
   SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
 jobs:
+  # Ensures correct release-note labels are set:
+  # - At least one label
+  # - At most one of the main category labels
+  # - A deprecation label alone or with something other than "none-required"
+  # Ensures you can have a change that is just a deprecation, or include a
+  # deprecation with another release note.
   check-label:
     name: Check release-note label set
     runs-on: ubuntu-latest
     steps:
       - uses: mheap/github-action-required-labels@v1
         with:
-          mode: exactly
+          mode: minimum
           count: 1
-          labels: "release-note/major, release-note/minor, release-note/small, release-note/docs, release-note/infra, release-note/none-required"
+          labels: "release-note/major, release-note/minor, release-note/small, release-note/docs, release-note/infra, release-note/deprecation, release-note, release-note-action-required, release-note/none-required"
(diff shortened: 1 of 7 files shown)


sunjayBhatia:
WIP need to test it out further

.github/workflows/label_check.yaml
```diff
@@ -9,22 +9,51 @@ env:
...
     name: Check release-note label set
     runs-on: ubuntu-latest
     steps:
       - uses: mheap/github-action-required-labels@v1
         with:
-          mode: exactly
+          mode: minimum
           count: 1
-          labels: "release-note/major, release-note/minor, release-note/small, release-note/docs, release-note/infra, release-note/none-required"
+          labels: "release-note/major, release-note/minor, release-note/small, release-note/docs, release-note/infra, release-note/deprecation, release-note, release-note-action-required, release-note/none-required"
```
skriss, line 26: do we want to drop `release-note` and `release-note-action-required`? I think they've basically been replaced by the new ones.

.github/workflows/label_check.yaml
```diff
@@ -9,22 +9,51 @@ env:
...
+      - uses: mheap/github-action-required-labels@v1
+        with:
+          mode: maximum
+          count: 1
+          labels: "release-note/major, release-note/minor, release-note/small, release-note/docs, release-note/infra, release-note, release-note-action-required, release-note/none-required"
```
skriss, line 31: ditto

.github/workflows/label_check.yaml
youngnick: Yeah +1 from me, we want to make sure people don't use those labels.

.github/workflows/label_check.yaml
```diff
@@ -9,22 +9,51 @@ env:
...
+          count: 1
+          labels: "release-note/deprecation, release-note/none-required"
   check-changelog:
     name: Check for changelog file
     needs:
       - check-label
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+    - uses: actions/cache@v2
```
youngnick, line 44: Does this speed up the check-changefile-exists.go run?

youngnick (approved):
LGTM with a couple of small questions.

.github/workflows/label_check.yaml
sunjayBhatia: Yeah that’s the intent, so each run doesn’t have so download/recompile go dependencies

skriss (approved):
LGTM!

PR #4337, opened 2022-02-10
test/e2e: when creating namespaces, wait for terminating ones to delete

skriss:
Gives terminating namespaces time to delete before
attempting to create a new one of the same name.

### DIFF
--- a/test/e2e/framework.go
+++ b/test/e2e/framework.go
@@ -340,6 +340,18 @@ func (f *Framework) CreateNamespace(name string) {
 			Labels: map[string]string{"contour-e2e-ns": "true"},
 		},
 	}
+	key := client.ObjectKeyFromObject(ns)
+
+	existing := &corev1.Namespace{}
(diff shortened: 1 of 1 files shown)


youngnick (approved):

sunjayBhatia (approved):

sunjayBhatia:
we also had an option in the namespace deletion helper to wait for deletion, in case that works too, though this strategy might be a little faster since you might not wait every single time: https://github.com/projectcontour/contour/blob/0d42d2eeee3fbae38f6022944ef932c028da3237/test/e2e/framework.go#L348

skriss:
Yeah, my thinking was to "lazily" wait -- I imagine waiting proactively for every deletion will slow things down a bunch, whereas this approach only waits if needed, which *should* only be in the case of retrying or re-running a test case (e.g. for the flake retries as you pointed out Sunjay, or for another PR I'm working on that ends up running certain tests twice with different configurations).

PR #4350, opened 2022-02-17
internal/envoy: fix cluster deduplication

### LINKED ISSUE #4242: Dynamic clusters clashing between Ingress and HttpProxy
Install Contour normally.
Have a http1 AND http2c compliant backend. 
Have the following not common but doable config:
Set up an Ingress: for example example.com with / going to a backend 80.
Set up an HTTPPProxy: example2.com with / going to the same backend 80 but with h2c protocol options.

Both example.com and example2.com will be server to the same backend BUT both will be served over http1.1 between envoy and the backend, meaning the backend service options for the service 80 port is dropped from the httpproxy config.

If I remember correctly the order does not matter, regardless of which is applied first Ingress wins.

Both fqdn-s to be served exactly as their respective resource is set up.

I attach the related service description 
[kahttp-service.txt](https://github.com/projectcontour/contour/files/7717754/kahttp-service.txt)

Ingress HTTPproxy manifests:
[ingress.txt](https://github.com/projectcontour/contour/files/7717770/ingress.txt)

The used backend will include the used http version in its output (response sent out to the client: Proto: HTTP/1.1)

**Environment:**

- Contour version: 1.15.2
- Kubernetes version: (use `kubectl version`): 1.21.4

### LINKED ISSUE #4348: Upstream SNI value is not configured correctly when two HTTPProxies refer to same upstream service
One of the `dag.Cluster` fields is [`SNI`](https://github.com/projectcontour/contour/blob/aa134a2a8a9f9050191e8cdbe4bf453675935a12/internal/dag/dag.go#L668-L673). It should be used when Envoy communicates towards the upstream service with TLS.  The value is derived from `HTTProxy.spec.routes.requestHeadersPolicy` `Host` key.

When Envoy clusters are created, unique name is generated by [hashing relevant fields](https://github.com/projectcontour/contour/blob/aa134a2a8a9f9050191e8cdbe4bf453675935a12/internal/envoy/cluster.go#L27-L56
) that should identify the cluster as unique. Since `SNI` is not part of the hash, the following otherwise equal `HTTPProxies` will result in same name and end up overwriting each other. The result is that only one cluster gets configured to Envoy:

```yaml
kind: HTTPProxy
metadata:
  name: my-httpproxy-1
spec:
  virtualhost:
    fqdn: test-1.127-0-0-101.nip.io
  routes:
    - services:
        - name: my-backend-service
          port: 8000
          protocol: tls
          validation:
            caSecret: my-ca-secret
            subjectName: my-backend-service
      requestHeadersPolicy:
        set:
          - name: Host
            value: "my-backend-service-1"  # SNI value
---
kind: HTTPProxy
metadata:
  name: my-httpproxy-2
spec:
  virtualhost:
    fqdn: test-2.127-0-0-101.nip.io
  routes:
    - services:
        - name:  my-backend-service
          port: 8000
          protocol: tls
          validation:
            caSecret: my-ca-secret
            subjectName: my-backend-service
      requestHeadersPolicy:
        set:
          - name: Host
            value: "my-backend-service-2"  # SNI value
```

There could be also `HTTPProxy` with no `requestHeaderPolicy` at all, which would mean no SNI. That also gets collapsed into  single cluster. What ends up in upstream TLS SNI depends on which cluster gets in first.

tsaarni:
Include SNI and Protocol fields into cluster name generation to ensure that different cluster will be generated if two clusters are otherwise equal except for those fields.  Previously clusters overwrote each other and only one got configured to Envoy.  

Fixes #4242 
Fixes #4348

### DIFF
--- a/changelogs/unreleased/4350-tsaarni-small.md
+++ b/changelogs/unreleased/4350-tsaarni-small.md
@@ -1,0 +1,1 @@
+Fixed a bug where upstream TLS SNI (`HTTProxy.spec.routes.requestHeadersPolicy` `Host` key) and protocol fields might not take effect when e.g. two `HTTPProxies` were otherwise equal but differed only on those fields.
--- a/internal/envoy/cluster.go
+++ b/internal/envoy/cluster.go
@@ -24,6 +24,10 @@ import (
 )
 
 // Clustername returns the name of the CDS cluster for this service.
+//
+// Note: Cluster name is used to deduplicate clusters.
+// When for example two HTTPProxies result in Clusters with equal name, only single cluster will be configured to Envoy.
+// Therefore the generated name must contain all relevant fields that make the cluster unique.
 func Clustername(cluster *dag.Cluster) string {
 	service := cluster.Upstream
 	buf := cluster.LoadBalancerPolicy
@@ -46,6 +50,7 @@ func Clustername(cluster *dag.Cluster) string {
 		buf += uv.CACertificate.Object.ObjectMeta.Name
 		buf += uv.SubjectName
 	}
+	buf += cluster.Protocol + cluster.SNI
 
 	// This isn't a crypto hash, we just want a unique name.
 	hash := sha1.Sum([]byte(buf)) // nolint:gosec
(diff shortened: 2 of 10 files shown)


youngnick:
I commented on the issue too, but just in case you miss it there, I think that #4242 is very similar?

tsaarni:
@youngnick it sure was thanks for noticing!  :blush:  I've updated the PR.

youngnick (approved):
LGTM, nice work.

skriss (approved):
LGTM 👍

PR #4356, opened 2022-02-21
Added idle connection timeout to HTTPProxy

### LINKED ISSUE #3313: Upstream Idle Connection Timeout Setting
**Need to understand, is there any configurable upstream idle connection setting at contour:**

As per envoy proxy timeout settings [envoy timeouts](https://www.envoyproxy.io/docs/envoy/latest/faq/configuration/timeouts#connection-timeouts) , upstream idle connections  use the common_http_protocol_options field in the Cluster’s extension_protocol_options, keyed by envoy.extensions.upstreams.http.v3.HttpProtocolOptions,  didn't find any corresponding option at contour to override via httpproxy or contour configuration 

We have services running on tomcat application server in kubernetes and idle connections being closed by tomcat with  keepalive setting 20 sec but still envoy using those connections to send request resulting upstream_cx_destroy_remote_with_active_rq 503 error and tomcat is closing connections before envoy 

Route level timeoutPolicy idle timeout default 5m only dealing with active stream but not idle connections, we have tried with route level idle timeout to match tomcat keep alive settings but it is not working  and still there are upstream_cx_destroy_remote_with_active_rq 503 errors as  tomcat is closing connections before envoy 

Flow:  Client --> Envoy(Kubernetes)-->Tomact App(Kubernetes)  503 with upstream to envoy 

**We need to have some configurable setting for upstream idle connections via contour configuration or httpproxy, and please provide details on what is the envoy default upstream idle connection timeout value  and how we can override it from contour/httpproxy**

tsaarni:
This change adds field `HTTPProxy.spec.routes.timeoutPolicy.idleConnection`. The field sets timeout for how long the upstream connection will be kept idle between requests before disconnecting it. This [connection timeout](https://www.envoyproxy.io/docs/envoy/latest/faq/configuration/timeouts#connection-timeouts) is different than previously existing field `timeoutPolicy.idle`, which controls the [stream timeout](https://www.envoyproxy.io/docs/envoy/latest/faq/configuration/timeouts#stream-timeouts): a timeout for single request-response. Altough connection timeout might be what user wanted in the first place, the existing field and functionality is kept for backwards compatibility and new field was introduced for the connection timeout.

When `timeoutPolicy.idleConnection` is not set then Envoy defaults to [1 hour](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#config-core-v3-httpprotocoloptions). User can also set `infinite` to disable the timeout

Timeout is set in Envoy `Cluster` by using [`HttpProtocolOptions`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/upstreams/http/v3/http_protocol_options.proto.html). If two `HTTPProxies` refer to single upstream service and are otherwise equal except for the `timeoutpolicy.idleConnection` value, then separate `Clusters` will be created for both to allow different timeouts.

Fixes #3313
    
### DIFF
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -11388,9 +11388,6 @@ func TestDAGInsert(t *testing.T) {
 
 	for name, tc := range tests {
 		t.Run(name, func(t *testing.T) {
-			if name != "insert httpproxy w/ infinite timeoutpolicy" {
-				return
-			}
 			builder := Builder{
 				Source: KubernetesCache{
 					FieldLogger: fixture.NewTestLogger(t),
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -8274,10 +8274,8 @@ func TestDAGInsert(t *testing.T) {
 					VirtualHosts: virtualhosts(
 						virtualhost("bar.com", &Route{
 							PathMatchCondition: prefixString("/"),
-							Clusters:           clustermap(s1),
-							TimeoutPolicy: TimeoutPolicy{
-								ResponseTimeout: timeout.DurationSetting(90 * time.Second),
-							},
+							Clusters:           withTimeout(clustermap(s1), TimeoutPolicy{ResponseTimeout: timeout.DurationSetting(90 * time.Second)}),
+							TimeoutPolicy:      TimeoutPolicy{ResponseTimeout: timeout.DurationSetting(90 * time.Second)},
 						}),
 					),
 				},
@@ -8337,10 +8335,8 @@ func TestDAGInsert(t *testing.T) {
 					VirtualHosts: virtualhosts(
 						virtualhost("bar.com", &Route{
 							PathMatchCondition: prefixString("/"),
-							Clusters:           clustermap(s1),
-							TimeoutPolicy: TimeoutPolicy{
-								ResponseTimeout: timeout.DisabledSetting(),
-							},
+							Clusters:           withTimeout(clustermap(s1), TimeoutPolicy{ResponseTimeout: timeout.DisabledSetting()}),
+							TimeoutPolicy:      TimeoutPolicy{ResponseTimeout: timeout.DisabledSetting()},
 						}),
 					),
 				},
@@ -12517,4 +12513,11 @@ func withMirror(r *Route, mirror *Service) *Route {
 		},
 	}
 	return r
+}
+
+func withTimeout(clusters []*Cluster, timeout TimeoutPolicy) []*Cluster {
+	for _, c := range clusters {
+		c.TimeoutPolicy = timeout
+	}
+	return clusters
 }
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -8233,7 +8233,7 @@ func TestDAGInsert(t *testing.T) {
 						virtualhost("*", &Route{
 							PathMatchCondition: prefixString("/"),
 							Clusters:           clustermap(s1),
-							TimeoutPolicy: TimeoutPolicy{
+							TimeoutPolicy: RouteTimeoutPolicy{
 								ResponseTimeout: timeout.DurationSetting(90 * time.Second),
 							},
 						}),
@@ -8254,7 +8254,7 @@ func TestDAGInsert(t *testing.T) {
 						virtualhost("*", &Route{
 							PathMatchCondition: prefixString("/"),
 							Clusters:           clustermap(s1),
-							TimeoutPolicy: TimeoutPolicy{
+							TimeoutPolicy: RouteTimeoutPolicy{
 								ResponseTimeout: timeout.DurationSetting(90 * time.Second),
 							},
 						}),
@@ -8274,8 +8274,8 @@ func TestDAGInsert(t *testing.T) {
 					VirtualHosts: virtualhosts(
 						virtualhost("bar.com", &Route{
 							PathMatchCondition: prefixString("/"),
-							Clusters:           withTimeout(clustermap(s1), TimeoutPolicy{ResponseTimeout: timeout.DurationSetting(90 * time.Second)}),
-							TimeoutPolicy:      TimeoutPolicy{ResponseTimeout: timeout.DurationSetting(90 * time.Second)},
+							Clusters:           clustermap(s1),
+							TimeoutPolicy:      RouteTimeoutPolicy{ResponseTimeout: timeout.DurationSetting(90 * time.Second)},
 						}),
 					),
 				},
@@ -8294,7 +8294,7 @@ func TestDAGInsert(t *testing.T) {
 						virtualhost("*", &Route{
 							PathMatchCondition: prefixString("/"),
 							Clusters:           clustermap(s1),
-							TimeoutPolicy: TimeoutPolicy{
+							TimeoutPolicy: RouteTimeoutPolicy{
 								ResponseTimeout: timeout.DisabledSetting(),
 							},
 						}),
@@ -8315,7 +8315,7 @@ func TestDAGInsert(t *testing.T) {
 						virtualhost("*", &Route{
 							PathMatchCondition: prefixString("/"),
 							Clusters:           clustermap(s1),
-							TimeoutPolicy: TimeoutPolicy{
+							TimeoutPolicy: RouteTimeoutPolicy{
 								ResponseTimeout: timeout.DisabledSetting(),
 							},
 						}),
@@ -8335,8 +8335,8 @@ func TestDAGInsert(t *testing.T) {
 					VirtualHosts: virtualhosts(
 						virtualhost("bar.com", &Route{
 							PathMatchCondition: prefixString("/"),
-							Clusters:           withTimeout(clustermap(s1), TimeoutPolicy{ResponseTimeout: timeout.DisabledSetting()}),
-							TimeoutPolicy:      TimeoutPolicy{ResponseTimeout: timeout.DisabledSetting()},
+							Clusters:           clustermap(s1),
+							TimeoutPolicy:      RouteTimeoutPolicy{ResponseTimeout: timeout.DisabledSetting()},
 						}),
 					),
 				},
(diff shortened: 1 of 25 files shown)


youngnick (approved):
LGTM, nice change.

tsaarni:
Since this is `HTTPProxy` API change, I'd appreciate another pair of eyes on the PR :smiley:  
Will merge after that.

internal/dag/policy.go
```diff
@@ -405,8 +405,8 @@ func ingressTimeoutPolicy(ingress *networking_v1.Ingress, log logrus.FieldLogger
 		response = annotation.ContourAnnotation(ingress, "request-timeout")
 		if len(response) == 0 {
 			return TimeoutPolicy{
-				ResponseTimeout: timeout.DefaultSetting(),
-				IdleTimeout:     timeout.DefaultSetting(),
+				ResponseTimeout:   timeout.DefaultSetting(),
+				IdleStreamTimeout: timeout.DefaultSetting(),
```
sunjayBhatia, line 409: missing `IdleConnectionTimeout` here?

internal/dag/dag.go
```diff
@@ -862,6 +869,9 @@ type ExtensionCluster struct {
 
 	// ConnectTimeout defines how long the proxy should wait when establishing connection to upstream service.
 	ConnectTimeout time.Duration
+
+	// IdleConnectionTimeout defines how long the proxy should keep idle connection to upstream service before disconnecting.
+	IdleConnectionTimeout timeout.Setting
```
sunjayBhatia, line 874: ExtensionCluster already has a TimeoutPolicy field, could we collapse this and other timeout fields into there? just did a quick look, not sure if it totally makes sense etc.

sunjayBhatia (commented):
just a couple small things, otherwise looks good, nice addition + naming/description of the field

internal/dag/policy_test.go
```diff
@@ -232,7 +232,7 @@ func TestTimeoutPolicy(t *testing.T) {
 				Idle: "900s",
 			},
 			want: TimeoutPolicy{
-				IdleTimeout: timeout.DurationSetting(900 * time.Second),
+				IdleStreamTimeout: timeout.DurationSetting(900 * time.Second),
```
skriss, line 235: would be nice to add cases for invalid & valid idle connection timeouts here, for consistency

skriss (commented):
Looks good 👍 , one inline comment on test coverage; I also noticed there's a `timeoutpolicy_test.go` in featuretests, would be nice to extend that to cover this new field for completeness.

internal/dag/dag.go
tsaarni: There was small "side-effect" that maybe makes this change bit illogical:  

`TimeoutPolicy` was previously only for `Route` but now after repurposing the same struct for `Cluster` as well, we will have complete set of timeout values included in both `dag.Route` and `dag.Cluster`. To my surprise, two tests in `builder_test.go` started failing:  the tests were about changing response timeout, which previously meant `Route.TimeoutPolicy.ResponseTimeout` was expected to be set. Now after this change, the test must also expect `Cluster.TimeoutPolicy.ResponseTimeout` to be set, even if response timeout makes no sense in context of `dag.Cluster`.

I'm OK with this small illogicality, but I made the change as separate commit so I can easily revert it as well, if wanted!

internal/dag/policy.go
tsaarni: Now added.

internal/dag/policy_test.go
tsaarni: Added tests for invalid, valid timeout and infinite.

tsaarni:
Thank you for review, good comments!

> Looks good 👍 , one inline comment on test coverage; I also noticed there's a `timeoutpolicy_test.go` in featuretests, would be nice to extend that to cover this new field for completeness.

I've now added test case to featuretests as well,

internal/dag/dag.go
sunjayBhatia: hm yeah maybe not worth the confusion, thanks for looking into that, ok if you would rather keep it as-is and revert

skriss: Yeah I'm fine backing it out too, but can live with either way.

youngnick: It doesn't seem like any of these are good options, so I guess we're choosing the least worst?

- Keep what we have here (timeoutpolicy the same, ResponseTimeout must be set for tests)
- revert what we have here, what happens?

Is there a better way to do this? Should we have two timeoutpolicy structs that are mostly the same except for this?

tsaarni: > Is there a better way to do this? Should we have two timeoutpolicy structs that are mostly the same except for this?

Exactly @youngnick, I got stalled because I started to think that would be the preferred way: specific timeout struct for route, another for cluster. I then got distracted by other work, so sorry for delay. I should try arrange some time soon! :)

youngnick: Awesome, I'll keep un-staling this then :)

tsaarni: Splitting `TimeoutPolicy` into `RouteTimeoutPolicy` and `ClusterTimeoutPolicy` worked OK, but I noticed that `dag.ExtensionCluster` behaves bit like a mixture of route and cluster: in addition of creating a cluster it is also used to configure `ExtAuthz_GrpcService` which  has response timeout just like `Route` does. So to follow the DRY principle, I ended up with following:

* `Route` has timeout policy attribute of type `RouteTimeoutPolicy`
* `Cluster` has `ClusterTimeoutPolicy`
* `ExtensionCluster` has both `RouteTimeoutPolicy` and `ClusterTimeoutPolicy`

and now `Route` does not contain timeout parameters that belong to `Cluster` and the other way around - with the negative side effect that `ExtensionCluster` needed both, but what do you think?

internal/dag/builder_test.go
```diff
@@ -11388,6 +11388,9 @@ func TestDAGInsert(t *testing.T) {
 
 	for name, tc := range tests {
 		t.Run(name, func(t *testing.T) {
+			if name != "insert httpproxy w/ infinite timeoutpolicy" {
+				return
+			}
```
skriss, line 11393: guessing this is left over from debugging :)

internal/dag/extension_processor.go
```diff
@@ -84,7 +84,7 @@ func (p *ExtensionServiceProcessor) buildExtensionService(
 	ext *contour_api_v1alpha1.ExtensionService,
 	validCondition *contour_api_v1.DetailedCondition,
 ) *ExtensionCluster {
-	tp, err := timeoutPolicy(ext.Spec.TimeoutPolicy, p.ConnectTimeout)
+	tp, _, err := timeoutPolicy(ext.Spec.TimeoutPolicy, p.ConnectTimeout)
```
skriss, line 87: should we be getting and using the cluster timeout policy here too? I don't see anywhere where it's being set otherwise

skriss (commented):
I think this breakout makes sense and is helpful for organizing things. One question on the ExtensionCluster impl.

internal/dag/builder_test.go
tsaarni: Sorry, my bad! Glad you spotted it, now fixed 😊

internal/dag/dag.go
```diff
@@ -861,6 +861,7 @@ type ExtensionCluster struct {
 	LoadBalancerPolicy string
 
 	// RouteTimeoutPolicy specifies how to handle timeouts to this extension.
+	// Defines default response timeout unless the
```
skriss, line 864: oops 😄

internal/dag/extension_processor.go
skriss: 😄  no problem, we've all done it. Updates look good!

skriss (approved):
LGTM, just one tiny godoc nit above

internal/dag/dag.go
tsaarni: Uhh :) I've now removed that unfinished sentense.  

I was originally going to point out there in comment that the response timeout can come from that struct (originating from `ExtensionService.spec.timeoutPolicy.response`) or it may be overwritten by value from `HTTPProxy.spec.virtualhost.authorization.responseTimeout`. This can be seen in [code here](https://github.com/projectcontour/contour/blob/2b7e6ccdc9fdcce787bab0e0bf3e753fd62a41b9/internal/dag/httpproxy_processor.go#L302-L306) and it was existing functionality but it confused me a bit - and without this logic, which I think is not described in API docs, the `RouteTimeoutPolicy` could be even removed from `ExtensionCluster`.

sivakrishna-git:
@tsaarni : Please update documentation https://projectcontour.io/docs/main/config/request-routing/#response-timeouts as well

internal/dag/policy.go
```diff
@@ -396,55 +396,67 @@ func ingressRetryPolicy(ingress *networking_v1.Ingress, log logrus.FieldLogger)
...
-		return TimeoutPolicy{
-			ResponseTimeout: timeout.DefaultSetting(),
-			IdleTimeout:     timeout.DefaultSetting(),
-		}, nil
+		return RouteTimeoutPolicy{
+				ResponseTimeout:   timeout.DefaultSetting(),
+				IdleStreamTimeout: timeout.DefaultSetting(),
+			}, ClusterTimeoutPolicy{
+				IdleConnectionTimeout: timeout.DefaultSetting(),
+				ConnectTimeout:        0,
```
ravilr, line 433: @tsaarni @sunjayBhatia @skriss while upgrade testing from contour-v1.19 to latest v1.22, seeing this issue where ContourConfig file has timeout.connect-timeout set to 5s, but the final envoy config rendered by the xdscache keeps fluctuating certain Cluster's connect_timeout b/w 5s and 2s on any onChanges(). it is expected to be 5s for all dynamic Cluster's connect_timeout.

I think this resetting of Cluster's ConnectTimeout=0 seems to be causing this. Why is this resetting Cluster's ConnectTimeout when a Route's TimeoutPolicy isn't set ? shouldn't this be setting Cluster's ConnectTimeout to passed func arg connectTimeout also ? (just like in the final return below).

tsaarni: @ravilr You are correct, thanks for spotting this!

PR #4363, opened 2022-02-22
internal/envoy: Make merge_slashes option configurable

### LINKED ISSUE #4339: support disabling Envoy's merge_slashes option
An upstream resource with URL path `/wiki/Mitou_Shoukan://Blood-Sign_-_Tập_1_Ending_1` becomes inaccessible when exposed via Contour - the path is normalized to `/wiki/Mitou_Shoukan:/Blood-Sign_-_Tập_1_Ending_1` in the upstream request.

The URL should be accessible.

~~It seems this is due to https://github.com/projectcontour/contour/pull/988 enabling `normalize_path` for connection managers as a mitigation for https://github.com/envoyproxy/envoy/security/advisories/GHSA-xcx5-93pw-jw2w. Currently, this value is not configurable.~~

Edit: The cause is [`merge_slashes`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-merge-slashes) rather than `normalize_path`.

**Environment:**

- Contour version: 1.20.0

mszabo-wikia:
#2406 enabled Envoy's non-standard `merge_slashes` path transformation
option[1] by default, which strips duplicate slashes from request URLs.
This can cause issues with upstream applications that do not use them
as an RFC 3986 delimiter,[2] so expose a configuration setting to toggle
this behavior. This toggle defaults to true to maintain compatibility
with existing installs.

Fixes #4339

---
[1] https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-merge-slashes
[2] https://datatracker.ietf.org/doc/html/rfc3986#section-2.2


mszabo-wikia:
Looks like I missed some signature changes during rebase, looking at that now.

mszabo-wikia:
With the changelog file name corrected, this should hopefully be ready now :)

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -289,6 +289,10 @@ type EnvoyListenerConfig struct {
 	// See: https://github.com/projectcontour/contour/issues/3221
 	DisableAllowChunkedLength bool `json:"disableAllowChunkedLength"`
 
+	// MergeSlashes toggles Envoy's non-standard merge_slashes path transformation option
+	// which strips duplicate slashes from request URL paths.
+	MergeSlashes bool `json:"mergeSlashes"`
```
skriss, line 294: I think this should be switched to `DisableMergeSlashes`, following a similar pattern as for the above field `DisableAllowChunkedLength`, with a default value of `false`. This way, Contour's default behavior doesn't change, and the user only needs to specify something if they want to change the default behavior. Otherwise, we run into problems where we can't distinguish between this field not being specified, and being explicitly specified as `false`.

cmd/contour/serve.go
```diff
@@ -335,6 +335,7 @@ func (s *Server) doServe() error {
 		Timeouts:                     timeouts,
 		DefaultHTTPVersions:          parseDefaultHTTPVersions(contourConfiguration.Envoy.DefaultHTTPVersions),
 		AllowChunkedLength:           !contourConfiguration.Envoy.Listener.DisableAllowChunkedLength,
+		MergeSlashes:                 contourConfiguration.Envoy.Listener.MergeSlashes,
```
skriss, line 338: Assuming we switch to `DisableMergeSlashes`, this is where I'd apply the negation (just like the line just above for `DisableAllowChunkedLength`), so that the `xdscache_v3.ListenerConfig` still just has a `MergeSlashes` field.

pkg/config/parameters.go
```diff
@@ -613,6 +613,10 @@ type Parameters struct {
 	// See: https://github.com/projectcontour/contour/issues/3221
 	DisableAllowChunkedLength bool `yaml:"disableAllowChunkedLength,omitempty"`
 
+	// MergeSlashes toggles Envoy's non-standard merge_slashes path transformation option
+	// which strips duplicate slashes from request URL paths.
+	MergeSlashes bool `yaml:"mergeSlashes,omitempty"`
```
skriss, line 618: Would also switch this one to `DisableMergeSlashes` with a default value of `false`. Otherwise, if this field is not specified in the config file, it's indistinguishable from it being set to `false` which would change the default behavior for users.

changelogs/unreleased/4363-mszabo-wikia-small.md
```diff
@@ -0,0 +1,2 @@
+Envoy's [`merge_slashes`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-merge-slashes) option that enables
+a non-standard path transformation option to replace multiple consecutive slashes in an URL path with a single slash is now configurable.
```
skriss, line 2:
```suggestion
a non-standard path transformation option to replace multiple consecutive slashes in an URL path with a single slash can now be disabled by setting the `DisableMergeSlashes` option in the Contour config file or ContourConfiguration custom resource.
```

skriss (changes requested):
Thanks for the PR @mszabo-wikia! I have one piece of feedback on the polarity of the config field, let's wait for @youngnick's input as well before switching everything just to make sure he agrees with me.

apis/projectcontour/v1alpha1/contourconfig.go
mszabo-wikia: Thanks, this sounds good. I thought applying the `kubebuilder` default would have sufficed for that but I see now that it only controls the generated example resources.

mszabo-wikia:
Thanks for the feedback! I've amended the PR accordingly

skriss (approved):
Thanks for the updates @mszabo-wikia, LGTM! Let's wait for one more maintainer to review as well before merging.

sunjayBhatia (approved):
LGTM, fan of the field naming/polarity for the config for users 👍🏽

youngnick (approved):
This is great, thanks @mszabo-wikia !

mszabo-wikia:
Thanks for the review and guidance! :)

PR #4367, opened 2022-02-24
Add docs for redirect policy

### LINKED ISSUE #4366: Examples of redirecting example.com to www.example.com
I believe the HTTPRequestRedirectPolicy can be used to redirect requests to example.com to www.example.com

Are there any examples of this being implemented?

we have 
kind: HTTPProxy
metadata:
  name: frontend
spec:
  ingressClassName: contour
  virtualhost:
    fqdn: www.example.com
  routes:
    - conditions:
        - prefix: /
      services:
        - name: web
          port: 80

We would like for request to example.com to be forwarded to this virtual host

Thanks

sunjayBhatia:
Fixes #4366


skriss (approved):
nice, LGTM. Could also copy this content into the v1.20.0 docs for posterity if you wanted, but non-blocking.

PR #4368, opened 2022-02-24
Add question issue template

sunjayBhatia:
We get a lot of issues tagged as bugs that are just questions. Since
reporters have no blank template to use, it appears the bug template is
the easiest to repurpose. This change adds a question template so that
contributor questions can be categorized appropriately.


skriss (approved):
Good idea 👍

youngnick (approved):
Great idea, thanks!

PR #4372, opened 2022-02-24
Fix docs link rendering

sunjayBhatia:


skriss (commented):
What was the issue? All I'm seeing is the number changed? Probably missing it

skriss (approved):

sunjayBhatia:
Also missed adding the actual link on some older pages, figured I would do some better numbering too

youngnick (approved):

Issue #4241, opened 2021-12-14
Support for regex rewrite in contour


gsadasiv:
Looking for a capability to rewrite path in HTTP proxy.

e.g. /f1/<f1-id>/f2/<f2-id>/f3/<f3-id>/…  -> /f1/<f1-id>/f3/<f3-id>/…

Envoy seems to have this capability through regex_rewrite. https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-r[…]v3-routeaction-regex-rewrite

Can this be exposed in Contour?

youngnick:
Contour already supports rewriting the request path in HTTPProxy, see https://projectcontour.io/docs/v1.19.1/config/request-rewriting/ for the current-stable docs on this.

Notably, however, it does *not* support regexes for either field - they must be exact matches.

There are two reasons for this:
- Firstly, it's difficult to validate that the regex will be accepted by Envoy and because of #1176, if we send Envoy an update that it can't accept, it may stop Envoy accepting config changes from Contour until the change is reverted. Basically, we can't validate that the regex will be accepted, and if a user put in an invalid regex, Envoy would silently stop accepting updates until the invalid regex was fixed. And you'd have no way to know, because Contour currently can't easily tell you that it was that config that produced the problem.
- Secondly, even if we fix those problems, Envoy keeps track of a metric called "regex size", which is a measure of the computational complexity of the regex. If the regex exceeds that size, it may also be rejected. #2043 has the details.

The tl;dr here is that:
- you can already do some path replacement
- path replacement must use literal strings, not a regex, unfortunately.

Hopefully this helps @gsadasiv.

skriss:
From @pavelnikolov in #4364:

> **Please describe the problem you have**
> 
> **tl;dr** I am trying to rewrite my path from `/prometheus/account/*/api/v1/query` to `/admin/prometheus/api/v1/query`.
> 
> I am using Contour as ingress controller in my Kubernetes cluster. I have a multi-tenanted Prometheus and I want to allow my users to access three paths `/api/v1/query`, `/api/v1/query_range` and `/federate`. I am using [PromAuthProxy](https://github.com/cherti/PromAuthProxy) to restrict the returned metrics if the user is authorized to access the metrics for the given account ID.
> 
> This is my `HttpProxy` configuration:
> 
> ```yaml
> # inside the root HttpProxy
>   ...
>   includes:
>     ...
[...]

Issue #3452, opened 2021-03-08
Guidance on setting Envoy runtime keys


sunjayBhatia:
> @stevesloka  @skriss  Envoy still enforces max_program_size for re2 regexes via run time keys (defaults to 100): See the discussion here : https://github.com/envoyproxy/envoy/pull/11187.     it doesn't matter if using xDS v3 api.
whereas Contour since v1.10 have completely removed setting max_program_size for regexes in xDS route configuration, essentially breaking all users of regex path since Contour v1.10. 

> Any guidance on how to set Envoy's runtime key in the Envoy bootstrap config ?

_Originally posted by @ravilr in https://github.com/projectcontour/contour/issues/2933#issuecomment-791154676_

This issue seems like a generic one that could be expanded to allow support or at least guidance in docs for using Envoy runtime keys with Contour bootstrap, or even Contour dynamically sending config w/ RTDS

sunjayBhatia:
@ravilr To start you could not rely on `contour bootstrap` and use a modified version of what that outputs as your Envoy bootstrap configuration (it isn't required that you use the output of that tool as your Envoy bootstrap config, you can certainly roll your own)

But it seems useful to have something in Contour that helps with setting up Envoy runtime config, even if it is just a guide/docs

youngnick:
I think there are two things for us to do here:
- Fix the max_program_size as soon as we can, somehow. Istio did it by using the deprecated fields temporarily. (istio/istio#26153)
- Figure out a more general thing to do about runtime config of Envoy. I think that RTDS is probably the way to go here, so that you can flip things around without needing to restart Envoy.

ravilr:
@sunjayBhatia  sure.  for others who stumble on to this due to the broken regex path in Ingress support, till an official Contour release gets this, you can either add the below layered runtime config snippet to envoy_config.json( passed as bootstrap config to envoy -c ):  

```
 "layered_runtime": {
     "layers": [
      {
       "name": "base",
       "static_layer": {
        "re2.max_program_size.error_level": 1048576
       }
      },
      {
       "name": "admin",
       "admin_layer": {}
      }
     ]
    }
```

or, if willing to run a patched version, below patch against contour-v1.10.1:
```
--- a/internal/envoy/v3/bootstrap.go
+++ b/internal/envoy/v3/bootstrap.go
@@ -33,10 +33,13 @@ import (
        envoy_matcher_v3 "github.com/envoyproxy/go-control-plane/envoy/type/matcher/v3"
[...]

PR #4379, opened 2022-02-28
internal/envoy/v3: Reenable larger max regex program size

sunjayBhatia:
Was inadvertently reverted previously to Envoy's default of 100. This
change sets the error limit of a regex program size back to 1048576
(2^20).

Also adds warning threshold that Envoy will log and collect stats about.

Updates: #4241
Updates: #3452

### DIFF
--- a/internal/envoy/v3/bootstrap.go
+++ b/internal/envoy/v3/bootstrap.go
@@ -175,6 +175,11 @@ func bootstrapConfig(c *envoy.BootstrapConfig) *envoy_bootstrap_v3.Bootstrap {
 						},
 					},
 				},
+				// Admin layer needs to be included here to maintain ability to
+				// modify runtime settings via admin console. We have it as the
+				// last layer so changes made via admin console override any
+				// settings from previous layers.
+				// See https://www.envoyproxy.io/docs/envoy/latest/configuration/operations/runtime#admin-console
 				{
 					Name: "admin",
 					LayerSpecifier: &envoy_bootstrap_v3.RuntimeLayer_AdminLayer_{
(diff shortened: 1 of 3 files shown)


internal/envoy/v3/bootstrap.go
```diff
@@ -156,6 +162,27 @@ func bootstrap(c *envoy.BootstrapConfig) ([]bootstrapf, error) {
...
+								"re2.max_program_size.warn_level":  {Kind: &structpb.Value_NumberValue{NumberValue: maxRegexProgramSizeWarn}},
+							},
+						},
+					},
+				},
+				{
+					Name: "admin",
+					LayerSpecifier: &envoy_bootstrap_v3.RuntimeLayer_AdminLayer_{
+						AdminLayer: &envoy_bootstrap_v3.RuntimeLayer_AdminLayer{},
+					},
```
skriss, line 182: Does this layer need to be specified?

skriss (approved):
LGTM, just one question

internal/envoy/v3/bootstrap.go
sunjayBhatia: Yeah, by default if you dont specify any runtime layers, Envoy inserts a default admin layer which should let you modify runtime values via the admin endpoint. If we omit it and specify other layers, you lose that functionality. (see https://www.envoyproxy.io/docs/envoy/latest/configuration/operations/runtime#admin-console)

sunjayBhatia: I can add a comment to that effect ^

skriss: Perfect, thanks!

skriss (approved):

youngnick (approved):

PR #4380, opened 2022-02-28
Add base RTDS implementation

sunjayBhatia:
Part of gradual implementation of RTDS, first step to allow configurable
runtime values. Returns an empty set of Runtime layers (instead of just
returning unimplemented to avoid issues in future upgrades). Next step
will be to add ability to dynamically fetch Runtime layer in bootstrap.

Updates: #3452


youngnick (approved):

skriss (approved):
LGTM

sunjayBhatia:
unrelated osx flake

PR #4387, opened 2022-03-03
Add dynamic Runtime layer to bootstrap

sunjayBhatia:
RuntimeCache returns only the dynamic layer if requested

Next step: add configurability

PR #4388, opened 2022-03-03
API additions for configurable Envoy Runtime fields

sunjayBhatia:
- Follow up to #4387
- Adds Runtime section to Envoy configuration in ContourConfiguration
CRD (to be copied over to config file parameters)
- Adds regex program size limit/warning threshold (naming TBD)

Further notes:
- This approach would mean we (Contour maintainers) have to be the
guards of what Runtime values are made configurable
- Not sure if there would be pushback, since that is what we have today for most things,
but this theoretically could be an opportunity to give users more
free-ranging control if we offered a more generic way to expose Envoy runtime fields
- API could look like (basically mirroring protobuf structs):
```
type EnvoyConfig struct {
  ...
  RuntimeSettings []RuntimeSetting
}

type RuntimeSetting struct {
  Key string
  Value RuntimeValue
}

// Only one field can be set
type RuntimeValue struct {
  Number *float64
  String *string
  Bool *bool
}
```
- Some drawbacks of this second approach:
  - its not easy to get a set of valid runtime keys to validate against, however this may sort of not matter since Envoy actually will accept whatever as a runtime key and save the value, regardless if it knows about it
  - there are some runtime keys that may interfere with Contour features, we could maybe try to prevent use of these via validation, but again might not be feasible/overly complex

Would appreciate any thoughts people have on this


youngnick (commented):
Personally, I'm in favor of the maintainers gating and validating what runtime features we allow the configuration of, because of the possibility for bad interactions with the stuff in there already, so I'm a -1 on having something more generic than what you have here.

What you have here looks good to me though, and has a pretty clear way for us to add more runtime settings.

PR #4391, opened 2022-03-07
internal/k8s: Set logger for controller-runtime in InitLogging

sunjayBhatia:
Ensures log messages from controller-runtime manager/child components
are logged and visible. Previously, these logs were actually written to
a null logger since the controller-runtime logger was not initialized
(see
https://github.com/kubernetes-sigs/controller-runtime/blob/9ee63fc65a9720568f74df9901b16883621636b4/pkg/log/log.go#L77-L85)

### DIFF
--- a/internal/k8s/log.go
+++ b/internal/k8s/log.go
@@ -78,9 +78,14 @@ func InitLogging(options ...LogOption) {
 		// Use the LogSink from a logrusr Logger, but wrapped in
 		// an adapter that always returns true for Enabled() since
 		// we rely on klog to do log level filtering.
+		logger := logrusr.New(p.log, logrusr.WithReportCaller())
 		klog.SetLogger(logr.New(&alwaysEnabledLogSink{
-			LogSink: logrusr.New(p.log, logrusr.WithReportCaller()).GetSink(),
+			LogSink: logger.GetSink(),
 		}))
+
+		// Also set the controller-runtime logger to the same
+		// concrete logger.
+		// controller_runtime_log.SetLogger(logger)
 	}
 }
 
--- a/internal/k8s/log.go
+++ b/internal/k8s/log.go
@@ -23,6 +23,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/sirupsen/logrus"
 	klog "k8s.io/klog/v2"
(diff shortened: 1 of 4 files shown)


internal/k8s/log.go
```diff
@@ -78,9 +79,14 @@ func InitLogging(options ...LogOption) {
...
 		// we rely on klog to do log level filtering.
+		logger := logrusr.New(p.log, logrusr.WithReportCaller())
 		klog.SetLogger(logr.New(&alwaysEnabledLogSink{
-			LogSink: logrusr.New(p.log, logrusr.WithReportCaller()).GetSink(),
+			LogSink: logger.GetSink(),
 		}))
+
+		// Also set the controller-runtime logger to the same
+		// concrete logger.
+		controller_runtime_log.SetLogger(logger)
```
skriss, line 89: do we need the same "always enabled" shenanigans here as above? I'm not sure

sunjayBhatia: if we want the `--kubernetes-debug` flag to apply to this logger properly, I think we have to do some custom logic in another wrapper since just wrapping a `logrus` logger with `logrusr` makes it so this logic applies: https://github.com/bombsimon/logrusr/blob/52223a38abaa5b1fa059a3cfebd121d7c307725f/logrusr.go#L80-L87

```
// Enabled tests whether this Logger is enabled. It will return true if the
// logrus.Logger has a level set to logrus.InfoLevel or higher (Warn/Panic).
// According to the documentation, level V(0) should be equivalent as calling
// Info() directly on the logger. To ensure this the constant `logrusDiffToInfo`
// will be added to all passed values so that V(0) creates a logger with level
// logrus.InfoLevel and V(2) would create a logger with level logrus.TraceLevel.
// This menas that if logrus is  set to logrus.InfoLevel or **higher** this
// method will return true, otherwise false.
```

if we set to always enabled, there wouldn't be anything stopping everything being logged i believe, since there is no klog in the middle

i made a copy of the existing log level test that shows this

sunjayBhatia: so its a little messier than I thought it would be... controller-runtime not using klog is a bit of a bummer

youngnick (approved):

PR #4392, opened 2022-03-07
internal/dag: Avoid error log for all Opaque secrets that do not contain ca.crt

jerome-quere:
When validating secrets contour would print an error log for every Opaque cluster secret that doesn't contain a ca.crt key.
This MR make sure that an error is printed only when a ca.crt key is present AND it's value is not a valid PEM certificate list.

Signed-off-by: Jerome Quere <jerome.quere@bodyguard.ai>

### DIFF
--- a/internal/dag/secret.go
+++ b/internal/dag/secret.go
@@ -74,9 +74,16 @@ func isValidSecret(secret *v1.Secret) (bool, error) {
 			return false, nil
 		}
 
+		data, ok := secret.Data[CACertificateKey]
+
+		// Secret is not a CA
+		if !ok {
+			return false, nil
+		}
+
 		// If there's an Opaque Secret with a `ca.crt` key, and it's zero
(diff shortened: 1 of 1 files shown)


sunjayBhatia:
thanks for the PR @jerome-quere, it would be great if you could add a test case here with this change: https://github.com/projectcontour/contour/blob/c1094dbae2b501fdc633ea65b427c178d9f634b4/internal/dag/secret_test.go#L209

sunjayBhatia (commented):
also if you could add a changelog following this guide: https://github.com/projectcontour/contour/blob/main/CONTRIBUTING.md#commit-message-and-pr-guidelines

that would be great!

skriss:
@jerome-quere just a nudge here to take a look at the feedback above, we have a new release coming up at the end of April so would be great to get this in before then, if possible.

skriss:
Updated this in #4528, closing this one.

PR #4394, opened 2022-03-08
certgen: add optional --name-prefix flag

skriss:
Adds an optional --name-prefix flag to the
certgen command that, if specified, adds a
prefix to the outputted Kubernetes secret
names, e.g. prefix-contourcert.

### DIFF
--- a/changelogs/unreleased/4394-skriss-small.md
+++ b/changelogs/unreleased/4394-skriss-small.md
@@ -1,0 +1,1 @@
+Adds an optional `--name-prefix` flag to the `contour certgen` command which, if specified, will be added as a prefix to the names of the generated Kubernetes secrets (e.g. `myprefix-contourcert` and `myprefix-envoycert`).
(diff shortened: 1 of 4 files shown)


sunjayBhatia (approved):

youngnick (approved):
LGTM regardless of if it's a prefix or suffix.

Feels like prefix has the advantage that the resources will be grouped by install/deployment in lists, suffix means they'll be grouped by type (envoy or contour).

skriss:
> Feels like prefix has the advantage that the resources will be grouped by install/deployment in lists, suffix means they'll be grouped by type (envoy or contour).

Yeah, I think that's the main difference. After thinking about it a bit I guess I'm still leaning towards prefix, so in `kubectl get` output things will be grouped first by the prefix (i.e. Contour instance name), but TBH I don't think it matters too much as long as we're consistent. If anyone else has thoughts, LMK.

skriss:
I'll merge as-is, there's plenty of time to switch to suffix before the next release if we decide that's better.

Issue #4390, opened 2022-03-06
Requests shouldn't be reevaluated after path rewrite


rodrigo14miguel:
## Problem

Path rewrite reevaluates HTTPProxy's conditions after the rewrite, instead of delivering the request directly to the service.
The request URL path should be rewritten and delivered to the service without addition condition evaluation.

## Reproduce

I have two deployments, **deploy1** served by service **svc1** and **deploy2** served by service **svc2**.
Both deployments are exposing their website with prefix **/** in the container.
Expose both services under the same FQDN with one HTTPproxy:

```yaml
apiVersion: projectcontour.io/v1
kind: HTTPProxy
metadata:
  name: myapp-httpproxy
spec:
  virtualhost:
    fqdn: myapp.example.com
    tls:
      secretName: server-tls
  routes:
  - services:
    - name: svc1
      port: 80
    conditions: #condition svc1
    - prefix: /
  - services:
    - name: svc2
      port: 80
    conditions: #condition svc2
    - prefix: /secondsvc
    pathRewritePolicy:
      replacePrefix:
      - prefix: /secondsvc
        replacement: /
```

## Expected behavior

When accessing **myapp.example.com/** I should reach **svc1/**. 
When accessing **myapp.example.com/secondsvc** I should reach **svc2/** (the request prefix **/secondsvc** should be rewritten as **/** and delivered directly to **svc2** without any additional condition reevaluation). 

## Actual behavior

When accessing **myapp.example.com/** I reach **svc1/**. 
When accessing **myapp.example.com/secondsvc** I reach **svc1/**. 
The **/secondsvc** prefix is being rewritten to **/** and then there is some kind of reevaluation against the HTTPproxy conditions. Because there is only one condition with prefix **/** and it belongs to **svc1**, the request is delivered to **svc1** instead of **svc2**.

**Note:** I know that the stated example will never work with the current behavior of contour. Because the rewritten request is reevaluated against the available conditions and **svc2** does not have any prefix **/** exposed, the rewritten request will never reach **svc2**. But I can't just have two conditions with the same **/** prefix, each pointing to different services, this creates a conflict.  

## Use case

Exposing multiple apps under the same FQDN when all the apps are using the same prefix internally.

rodrigo14miguel:
[...]

PR #4395, opened 2022-03-08
test/e2e: Add path prefix rewrite test

sunjayBhatia:
To demonstrate it works, and to give a framework to use if/when we do
path regex rewriting


sunjayBhatia:
wrote this up when looking at https://github.com/projectcontour/contour/issues/4390 since we didnt have something for it and it was easier to demonstrate it this way, ok if this is extra coverage that doesn't need to be merged, but figured I would put up a PR

skriss (approved):
LGTM and matches my expectations of this feature, will be interesting to see what comes of #4390

PR #4402, opened 2022-03-16
Remove the hack in scripts for certgen job

### LINKED ISSUE #4318: certgen job can use an IfNotPresent image pull policy
Historically the certgen job used an `Always` image pull policy because it was configured to use the `latest` image tag. However, it now is named uniquely for each release and uses a specific version as an image tag. As such, it can use an `IfNotPresent` image pull policy like all other images now, meaning the special cases for it in our tooling can be removed.

rajatvig:
examples/render: Set the `imagePullPolicy` for certen job to `IfNotPresent`

Set the `imagePullPolicy` for certen job to `IfNotPresent`

Fixes #4318 

Signed-off-by: Rajat Vig <rvig@etsy.com>

### DIFF
--- a/examples/render/contour-deployment.yaml
+++ b/examples/render/contour-deployment.yaml
@@ -4908,7 +4908,7 @@ spec:
       containers:
       - name: contour
         image: ghcr.io/projectcontour/contour:main
-        imagePullPolicy: Always
+        imagePullPolicy: IfNotPresent
         command:
         - contour
         - certgen
--- a/changelogs/unreleased/4402-rajatvig-small.md
+++ b/changelogs/unreleased/4402-rajatvig-small.md
@@ -1,0 +1,1 @@
+Removed the hack for ImagePullPolicy for certgen
--- a/examples/render/contour-gateway.yaml
+++ b/examples/render/contour-gateway.yaml
@@ -4911,7 +4911,7 @@ spec:
       containers:
       - name: contour
         image: ghcr.io/projectcontour/contour:main
-        imagePullPolicy: Always
+        imagePullPolicy: IfNotPresent
         command:
         - contour
         - certgen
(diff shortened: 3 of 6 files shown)


youngnick (approved):
Nice work, thanks!

rajatvig:
@youngnick Would you know why the build failed for release notes? Do not see any options to add release note labels.

examples/render/contour-deployment.yaml
```diff
@@ -4908,7 +4908,7 @@ spec:
       containers:
       - name: contour
         image: ghcr.io/projectcontour/contour:main
-        imagePullPolicy: Always
+        imagePullPolicy: IfNotPresent
```
stevesloka, line 4911: This was set this way on the `main` branch because if not you'd never get an updated version of it. The versioned examples do change this to `Always`.

skriss: @stevesloka I think you're suggesting that we move [this logic](https://github.com/projectcontour/contour/blob/main/hack/generate-deployment.sh#L56-L66) out of this script that generates the rendered manifests, and into https://github.com/projectcontour/contour/blob/main/hack/release/make-release-tag.sh so it only applies to versioned manifests?

stevesloka: I guess my point was on the `main` branch, if you used this code (which you really shouldn't but only in a dev/test env), then the 2nd time you updated you wouldn't get the new version of Contour since there's a version already present on the machine. =)

skriss: Yep makes sense -- this has been true for the contour serve, bootstrap and shutdown manager image refs for a long time so odds are nobody's actually using this manifest anyway, but filed https://github.com/projectcontour/contour/issues/4404 to track the additional change

PR #4405, opened 2022-03-16
internal/envoy: Fix disableMergeSlashes config option propagation

mszabo-wikia:
In 335e816ec14de417abebcb3c387c462c0470143d we made Envoy's
merge_slashes option configurable via the ContourConfiguration CRD and
the main configuration file, but the value from the configuration isn't
actually propagated to listeners. I did not notice this in our original
testing pass as that was performed before the polarity of the config
option was inverted.

As a fix, ensure we correctly pass the disableMergeSlashes config option
value to Envoy listener configs and add E2E test cases to verify the
default behavior as well as the config option.

### DIFF
--- a/cmd/contour/servecontext.go
+++ b/cmd/contour/servecontext.go
@@ -448,6 +448,7 @@ func (ctx *serveContext) convertToContourConfigurationSpec() contour_api_v1alpha
 			Listener: &contour_api_v1alpha1.EnvoyListenerConfig{
 				UseProxyProto:             &ctx.useProxyProto,
 				DisableAllowChunkedLength: &ctx.Config.DisableAllowChunkedLength,
+				DisableMergeSlashes:       &ctx.Config.DisableMergeSlashes,
 				ConnectionBalancer:        ctx.Config.Listener.ConnectionBalancer,
 				TLS: &contour_api_v1alpha1.EnvoyTLS{
 					MinimumProtocolVersion: ctx.Config.TLS.MinimumProtocolVersion,


sunjayBhatia:
ah good catch, yeah an e2e test makes 100% all the code paths are covered, we have some work to do for coverage closer to the `cmd` package

sunjayBhatia:
hm, every few times i run the test locally looks like one of them fails, havent looked too deeply into why yet

mszabo-wikia:
@sunjayBhatia Yeah, this is odd... I followed the same pattern as for other tests that exercise custom config options so not entirely sure what could be the issue yet

mszabo-wikia:
Maybe I should explicitly set the config options to `true` in the other case in case they leak from the other test, is my only guess

skriss (commented):
Thanks for catching this and fixing @mszabo-wikia, the fix LGTM. I can observe the test flake @sunjayBhatia pointed out too, but I'm stumped as to where it's coming from as well. My hunch is if instead of just checking for a 200 response in the `RequestUntil` method, we also checked for the expected request path, this would succeed, but still not exactly sure what sequence of events is going on.

test/e2e/ingress/disable_merge_slashes_test.go
```diff
@@ -0,0 +1,99 @@
...
+				Condition: e2e.HasStatusCode(200),
+			})
+			require.NotNil(t, res, "request never succeeded")
+			require.Truef(t, ok, "expected 200 response code, got %d", res.StatusCode)
+
+			if disableMergeSlashes {
+				assert.Equal(t, "/path/with//multiple/slashes", f.GetEchoResponseBody(res.Body).Path)
+			} else {
+				assert.Equal(t, "/path/with/multiple/slashes", f.GetEchoResponseBody(res.Body).Path)
+			}
```
skriss, line 96: Something like this seems to get rid of the test flake for me (unfortunately `f.HTTP.RequestUntil` doesn't allow you to check conditions on the response body, something to fix but don't need to do it this PR, hence why it's wrapped in a `require.Eventually`).

I'm wondering if sometimes the Route update gets to Envoy before the Listener update, so the request returns a 200 but the listener hasn't actually been reconfigured yet? @sunjayBhatia does that sound possible to you?

```suggestion
			require.Eventually(t, func() bool {
				res, ok := f.HTTP.RequestUntil(&e2e.HTTPRequestOpts{
					Host:      i.Spec.Rules[0].Host,
					Path:      "/path/with//multiple/slashes",
					Condition: e2e.HasStatusCode(200),
				})
				if !ok || res == nil {
					return false
				}

				if disableMergeSlashes {
					return "/path/with//multiple/slashes" == f.GetEchoResponseBody(res.Body).Path
				} else {
					return "/path/with/multiple/slashes" == f.GetEchoResponseBody(res.Body).Path
				}
			}, time.Minute, time.Second)
```

sunjayBhatia: yeah, since the Listener resource stream and Route resource streams are separate, you can get into that case (which is an other advertisement for using the ADS version of xDS so things are ordered properly for us)

might be able to fix it by making this a TLS enabled Ingress so requests have to go through a different filter chain to be accepted, so the listener update has to happen before a request is successful

sunjayBhatia:
oh also we do have this existing test in the httpproxy package that we could piggy back off of: https://github.com/projectcontour/contour/blob/ce4ef88a22a7188adcc8964e2e82030e7e9e3ea0/test/e2e/httpproxy/merge_slash_test.go

skriss:
> oh also we do have this existing test in the httpproxy package that we could piggy back off of: https://github.com/projectcontour/contour/blob/ce4ef88a22a7188adcc8964e2e82030e7e9e3ea0/test/e2e/httpproxy/merge_slash_test.go (same thing basically)

Ah nice, yeah we should de-dupe these, totally forgot we had this one!

test/e2e/ingress/disable_merge_slashes_test.go
skriss: FYI we've merged https://github.com/projectcontour/contour/pull/4409, so you could now implement the change above directly within the `Condition` function in the `f.HTTP.RequestUntil` call and drop the `require.Eventually`, or @sunjayBhatia's suggestion is another viable path.

skriss:
@mszabo-wikia would you be able to circle back and resolve the test flake and merge conflicts here? We should definitely get this fix in for the upcoming 1.21 release (RC date is Apr 21).

mszabo-wikia:
Hey @skriss , thanks for the heads-up - I was originally waiting on #4409 but then haven't been able to follow up due to a bout of Covid. Hopefully I can pick this PR up again in a few days' time.

mszabo-wikia:
@skriss Given the proximity of the 1.21 RC, what do you think of landing this fix separately and then updating the test suite as a fast-follow?

skriss:
> @skriss Given the proximity of the 1.21 RC, what do you think of landing this fix separately and then updating the test suite as a fast-follow?

Works for me, I don't want to not get the fix in because we're blocking on tests.

mszabo-wikia:
Awesome, thank you

skriss:
Got tests updated in #4495 so we can get this fix in before the upcoming release, closing this one out in favor of #4495.

PR #4409, opened 2022-03-17
test/e2e: minor fixes and improvements

skriss:
- Fixes a bug in the code that waits for an existing
terminating namespace to be gone before creating a
new one for E2E tests which caused the create to
fail.
- Changes the type of the condition func used
by HTTP.RequestUntil to allow checking for
properties of the response body.

### DIFF
--- a/test/e2e/framework.go
+++ b/test/e2e/framework.go
@@ -348,7 +348,7 @@ func (f *Framework) CreateNamespace(name string) {
 		// Got an existing namespace and it's terminating: give it a chance to go
 		// away.
 		require.Eventually(f.t, func() bool {
-			return api_errors.IsNotFound(f.Client.Get(context.TODO(), key, ns))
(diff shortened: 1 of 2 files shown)


test/e2e/framework.go
```diff
@@ -348,7 +348,7 @@ func (f *Framework) CreateNamespace(name string) {
 		// Got an existing namespace and it's terminating: give it a chance to go
 		// away.
 		require.Eventually(f.t, func() bool {
-			return api_errors.IsNotFound(f.Client.Get(context.TODO(), key, ns))
+			return api_errors.IsNotFound(f.Client.Get(context.TODO(), key, existing))
```
youngnick, line 351: :oof:

youngnick (approved):
Nice finds!

sunjayBhatia (approved):

PR #4410, opened 2022-03-18
support processing a specific Gateway

skriss:
Adds a configuration option to process a specific
Gateway. This is an optional alternate way to run
Contour for Gateway API, vs. the existing method
of configuring a GatewayClass controller string, 
and having Contour detect the first GatewayClass
and associated Gateway for that controller string.


apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -108,8 +108,17 @@ type GatewayConfig struct {
...
+	// +optional
+	ControllerName string `json:"controllerName,omitempty"`
+
+	// GatewayName defines a specific Gateway that this Contour
+	// instance corresponds to. If set, Contour will reconcile
+	// only this gateway, and will not reconcile any gateway
+	// classes.
+	// Exactly one of ControllerName or GatewayName must be set.
+	// +optional
+	GatewayName *NamespacedName `json:"gatewayName,omitempty"`
```
skriss, line 121: The naming here definitely stutters a bit, but couldn't think of a better alternative. In config this would look like:
```yaml
gateway:
  gatewayName:
    namespace: foo
    name: bar
```

One alternate option would be to call this field `specificGateway`, so:
```yaml
gateway:
  specificGateway:
    namespace: foo
    name: bar
```

Open to input!

internal/dag/cache.go
```diff
@@ -146,11 +154,45 @@ func (kc *KubernetesCache) Insert(obj interface{}) bool {
...
+				}
+
+				kc.gateway = obj
+
+				gatewayClass := &gatewayapi_v1alpha2.GatewayClass{}
+				if err := kc.Client.Get(context.Background(), client.ObjectKey{Name: string(kc.gateway.Spec.GatewayClassName)}, gatewayClass); err != nil {
+					kc.WithError(err).Errorf("error getting gatewayclass for gateway %s/%s", kc.gateway.Namespace, kc.gateway.Name)
+				} else {
+					kc.gatewayclass = gatewayClass
+				}
```
skriss, line 188: This pattern (fetch related resource B when resource A is added to the cache) is a little different than what we have elsewhere, but it seemed OK, particularly given the controller-runtime client will do reads from its cache, so we shouldn't actually be hitting the API server here, thoughts?

youngnick: This seems like a good approach to me, I think something like this pattern is common in other controller-runtime clients.

apis/projectcontour/v1alpha1/contourconfig.go
youngnick: I like `specificGateway`, or even just `specific`. But I think people will understand if we just use `gatewayName` as well.

internal/dag/cache.go
```diff
@@ -47,6 +49,10 @@ type KubernetesCache struct {
 	// cached.
 	IngressClassNames []string
 
+	// Gateway is the optional name of the specific Gateway to cache.
+	// If set, only the Gateway with this namespace/name will be kept.
+	Gateway *types.NamespacedName
```
sunjayBhatia, line 54: nit: reading some of the tests this `Gateway` field and the cached `gateway` field were a little confusing at first

sunjayBhatia, line 54: nit: reading some of the tests this `Gateway` field and the cached `gateway` field were a little confusing at first

sunjayBhatia (commented):
just one nit on field naming, otherwise a general question on informer vs the controller pattern

do you have any thoughts on sticking with the controllers, since we only have the for the gateway resources? kind of in an in-between state rn. we could either expand usage to all resources and add the "specific gateway" logic in the relevant controllers or move back to all informers w/ the dag cache insertion being the place to do filtering?

skriss:
> do you have any thoughts on sticking with the controllers, since we only have the for the gateway resources? kind of in an in-between state rn. we could either expand usage to all resources and add the "specific gateway" logic in the relevant controllers or move back to all informers w/ the dag cache insertion being the place to do filtering?

Yeah, that's a good question. I think I went with informers here because the filtering logic needed was minimal and it meant I didn't have to touch the existing GatewayClass/Gateway controllers which are already somewhat complex and insert a bunch of conditionals, but you're right that we continue to be in an in-between state and maybe that wasn't the right choice.

I've always liked the simplicity of directly using informers, it's less boilerplate and less runtime overhead than controllers, and in general we're not really reconciling individual objects which controllers would be better for, instead processing the whole state of the world on every relevant event, but I think either pattern can work. My main concern with moving to controllers for every resource has been that without benchmarking, we wouldn't know if we were introducing performance regressions or not.

I think it'd be worth taking a look at some other non-Gateway resources, e.g. Services, and doing a spike on what that would look like with the controller pattern, to get a better feel for if we want to move everything over or if we're OK living with both patterns existing side-by-side.

apis/projectcontour/v1alpha1/contourconfig.go
skriss: I don't have a strong preference here between these options; @sunjayBhatia any thoughts?

sunjayBhatia:
> I think it'd be worth taking a look at some other non-Gateway resources, e.g. Services, and doing a spike on what that would look like with the controller pattern, to get a better feel for if we want to move everything over or if we're OK living with both patterns existing side-by-side.

yeah thats a good idea to see what it would look like to use the controller pattern expanded more, but i agree generally informers feel a little lighter-weight and fit into what we already have integrated with the `ResourceEventHandler` interface

apis/projectcontour/v1alpha1/contourconfig.go
sunjayBhatia: only thing i can think of is:

```
gateway:
  controlledGateway:
    namespace: foo
    name: bar
```

or

```
gateway:
  gatewayRef:
    namespace: foo
    name: bar
```

becomes maybe a little better if the struct name changes?

```
gatewayAPI:
  gatewayRef:
    namespace: foo
    name: bar
```

not a too strong opinion, your other options are good too, favorite of mine is maybe `gatewayRef`

sunjayBhatia:
> I think I went with informers here because the filtering logic needed was minimal and it meant I didn't have to touch the existing GatewayClass/Gateway controllers which are already somewhat complex and insert a bunch of conditionals

im ok with keeping as you have it not to block progress and just logging an issue to refactor this

apis/projectcontour/v1alpha1/contourconfig.go
youngnick: I am +1 on `gatewayRef` - `xRef` is the preferred upstream way to name this field, so it's consistent with upstream.

youngnick:
I agree that we should try and be consistent, either controllers or informers. My understanding was that controllers made some of the cache problems that informers have go away, and any additional complexity was worth it on that front, but my information is old, so I am very supportive of spiking something and checking.

apis/projectcontour/v1alpha1/contourconfig.go
skriss: :+1: I like that, good idea @sunjayBhatia! Have updated.

skriss:
Filed https://github.com/projectcontour/contour/issues/4417 to track the informer vs. controller item.

sunjayBhatia (approved):

youngnick (approved):

PR #4411, opened 2022-03-18
test/e2e: add test for Gateway with multiple HTTPS listeners

skriss:
Adds a test that demonstrates a Gateway
with multiple HTTPS Listeners, each with
a different hostname and TLS cert.

Also updated some of the Describe text for clarity.

### DIFF
--- a/test/e2e/framework.go
+++ b/test/e2e/framework.go
@@ -18,6 +18,8 @@ package e2e
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
 	"net/http"
 	"os"
@@ -521,4 +523,17 @@ func DetailedConditionInvalid(conditions []contourv1.DetailedCondition) bool {
 		}
 	}
 	return false
+}
+
+// VerifyTLSServerCert returns a TLS config functional
+// option that enables verifying the TLS server cert using
+// the provided CA cert.
+func VerifyTLSServerCert(caCert []byte) func(*tls.Config) {
+	return func(c *tls.Config) {
+		certPool := x509.NewCertPool()
+		certPool.AppendCertsFromPEM(caCert)
+
+		c.RootCAs = certPool
+		c.InsecureSkipVerify = false
+	}
 }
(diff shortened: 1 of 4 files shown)


skriss:
I have to say, I did not quite realize that this worked this way.

sunjayBhatia:
unrelated flake looks like

sunjayBhatia:
> I have to say, I did not quite realize that this worked this way.

yeah i agree, i never really thought of it like this, makes sense though rereading [this](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1alpha2.GatewaySpec) and makes sense it works how we implemented it since for each you would add another secure vhost in the dag, and theres no conflicts

sunjayBhatia (approved):

skriss:
> > I have to say, I did not quite realize that this worked this way.
> 
> yeah i agree, i never really thought of it like this, makes sense though rereading [this](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1alpha2.GatewaySpec) and makes sense it works how we implemented it since for each you would add another secure vhost in the dag, and theres no conflicts

Yep -- I think we do have some additional validation to do to ensure that all the HTTPS listeners are truly compatible (same port, no conflicting hostnames, etc), but nice that it mostly already works. I guess one way to think about it is that a Gateway API Listener is more akin to an Envoy filter chain, than an Envoy listener.

youngnick:
> I guess one way to think about it is that a Gateway API Listener is more akin to an Envoy filter chain, than an Envoy listener.

In this case, yeah. But sadly, not always. These are examples of _mergeable_ listeners, which act more like HCMs with their own filter set in Envoy terms, and like you said, _mergeable_ listeners have to have no conflicts on a bunch of values. Listeners *can* also be used to describe completely separate ports etc, which would act more like Envoy listeners.

youngnick (approved):

youngnick:
This seems like something we should probably have upstream conformance tests on too, now that I think about it.

skriss:
> This seems like something we should probably have upstream conformance tests on too, now that I think about it.

Yeah I'd love to contribute a few more of the E2E's that we have upstream, just need to find some time for it. Ultimately it'd be awesome if we could mostly get rid of ours and rely on the conformance tests instead.

Issue #4423, opened 2022-03-26
Envoy is not sending metrics on GRPC


rajatvig:
We are using Contour as an Ingress over GRPC Services and started to notice that the underlying service returns an error, Envoy reports 200 in the access log and there are no GRPC metrics.

After looking up Envoy documentation, there is a [grpc_stats_filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/grpc_stats_filter) but it is not configured in the listener.

PR #4424, opened 2022-03-26
Add grpc_stats filter to listener

rajatvig:
Add grpc_stats filter to envoy. Fixes issue #4423.

Signed-off-by: Rajat Vig <rvig@etsy.com>


rajatvig:
While the code looks like it would work, am unsure how to verify it. Is there any documentation on how to test features like this locally?

youngnick:
In the past, when testing stats, we've just done local smoke tests by running a local build, running the right traffic through, then checking that the stats increment.

Feels like it should be testable, but there's a lot of moving parts involved in setting up a gRPC service, that I don't think we have yet.

rajatvig:
```json
{
  "name": "grpc_stats",
  "typed_config": {
   "@type": "type.googleapis.com/envoy.extensions.filters.http.grpc_stats.v3.FilterConfig",
   "emit_filter_state": true,
   "enable_upstream_stats": true
  }
}
```

is the filter in the `config_dump` on the Envoy pod. 

Unsure, if there is any value to emit metrics per path. That might need an option on the Cluster itself and likely another pull request.

rajatvig:
@youngnick Does this look good?

rajatvig:
Could this be looked at please?

examples/contour/01-contour-config.yaml
```diff
@@ -95,6 +95,7 @@ data:
     #   - "upstream_service_time"
     #   - "user_agent"
     #   - "x_forwarded_for"
+    #   - "grpc_status"
```
sunjayBhatia, line 98: in order for `grpc_status` to be a default field logged with JSON logging, you will have to add to this list: https://github.com/projectcontour/contour/blob/667dc3ff0f0b9ad545bb9746be8232b968c22e8c/pkg/config/accesslog.go#L17

sunjayBhatia: also will require an addition to the testing here and the accompanying implementation: https://github.com/projectcontour/contour/blob/667dc3ff0f0b9ad545bb9746be8232b968c22e8c/pkg/config/parameters_test.go#L194

otherwise actually using that field will not succeed, it will be rejected as invalid

sunjayBhatia: similarly for https://github.com/projectcontour/contour/blob/667dc3ff0f0b9ad545bb9746be8232b968c22e8c/pkg/config/parameters_test.go#L506

changelogs/unreleased/4424-rajatvig-small.md
```diff
@@ -0,0 +1 @@
+Add grpc_stats filter for Envoy
```
sunjayBhatia, line 1: might want to also add a note that this will give you the ability to log grpc status codes in access logs as well

internal/envoy/v3/listener.go
```diff
@@ -25,6 +25,7 @@ import (
 	envoy_listener_v3 "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
 	envoy_compressor_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/compressor/v3"
 	envoy_config_filter_http_ext_authz_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/ext_authz/v3"
+	envoy_config_filter_http_grpc_stats "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/grpc_stats/v3"
```
sunjayBhatia, line 28: nit: we've typically added the `v*` to the import path, would be good to do here as well for consistency

internal/envoy/v3/listener.go
```diff
@@ -50,10 +51,11 @@ const (
 	HTTPVersion2    HTTPVersionType = http.HttpConnectionManager_HTTP2
 	HTTPVersion3    HTTPVersionType = http.HttpConnectionManager_HTTP3
 
-	HTTPFilterRouter  = "type.googleapis.com/envoy.extensions.filters.http.router.v3.Router"
-	HTTPFilterCORS    = "type.googleapis.com/envoy.extensions.filters.http.cors.v3.Cors"
-	HTTPFilterGrpcWeb = "type.googleapis.com/envoy.extensions.filters.http.grpc_web.v3.GrpcWeb"
-	HTTPFilterGzip    = "type.googleapis.com/envoy.extensions.compression.gzip.compressor.v3.Gzip"
+	HTTPFilterRouter    = "type.googleapis.com/envoy.extensions.filters.http.router.v3.Router"
+	HTTPFilterCORS      = "type.googleapis.com/envoy.extensions.filters.http.cors.v3.Cors"
+	HTTPFilterGrpcStats = "type.googleapis.com/envoy.extensions.filters.http.grpc_stats.v3.FilterConfig"
```
sunjayBhatia, line 56: do we need this? doesn't look like it is used anywhere?

sunjayBhatia (changes requested):
some comments around the access logging changes etc.

examples/contour/01-contour-config.yaml
rajatvig: Done. Added the test cases and to the default list.

changelogs/unreleased/4424-rajatvig-small.md
rajatvig: Done

internal/envoy/v3/listener.go
rajatvig: Done

internal/envoy/v3/listener.go
rajatvig: Removed.

internal/envoy/v3/listener_test.go
```diff
@@ -22,6 +22,7 @@ import (
 	envoy_listener_v3 "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
 	envoy_compressor_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/compressor/v3"
 	envoy_config_filter_http_ext_authz_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/ext_authz/v3"
+	envoy_config_filter_http_grpc_stats "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/grpc_stats/v3"
```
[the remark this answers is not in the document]
rajatvig: Done

sunjayBhatia (commented):
About the access logging change: Since we're providing `grpc_status` and `%GRPC_STATUS%` access log fields that print the textual representation of the status code (e.g. `Unavailable` for status code 14), might we also want to provide the code itself in the future

Envoy 1.22 looks like it merged a change to print the number in access logs as well: https://github.com/envoyproxy/envoy/pull/20678

rajatvig:
That sounds great. I assume that'd be a different PR to do when 1.22 releases though, right?

Can create the issue so that I can pick it up once 1.22 releases and Contour supports it. What do you think?

sunjayBhatia:
> That sounds great. I assume that'd be a different PR to do when 1.22 releases though, right?
> 
> Can create the issue so that I can pick it up once 1.22 releases and Contour supports it. What do you think?

yeah, due to upgrade safety we will likely have to merge that into contour after we release a version (likely 1.21) of contour that is bundled with Envoy 1.22 (so we make sure during an upgrade any new features we use aren't sent via xDS to older Envoys)

creating an issue to track it would be great, I can add that issue to the Contour 1.22 milestone so we make sure to get it in then

sunjayBhatia (approved):

rajatvig:
Actually, just checked that 1.22 was released 3 days back. Would Contour 1.21 be supporting 1.22? If so, I can update the PR to include the changes.

sunjayBhatia:
> Actually, just checked that 1.22 was released 3 days back. Would Contour 1.21 be supporting 1.22? If so, I can update the PR to include the changes.

yeah, since this is a totally new access log format string that will be supported in contour, we can support it in contour 1.21 since we will also be bumping to Envoy 1.22 (if it was a different scenario, we may not be able to merge it until Contour 1.22 but in this case it is a totally new feature that wasn't available before so it works)

sunjayBhatia:
> > Actually, just checked that 1.22 was released 3 days back. Would Contour 1.21 be supporting 1.22? If so, I can update the PR to include the changes.
> 
> yeah, since this is a totally new access log format string that will be supported in contour, we can support it in contour 1.21 since we will likely also be bumping to Envoy 1.22 (if it was a different scenario, we may not be able to merge it until Contour 1.22 but in this case it is a totally new feature that wasn't available before so it works)

but maybe do this as a follow up that can be merged after we bump Envoy, in case we happen to hold it off, don't want to block this change as the grpc stats filter is generally helpful anyway

PR #4426, opened 2022-03-29
internal/provisioner: support more than one instance per namespace

### LINKED ISSUE #4418: Gateway API: support >1 Contour instance per namespace
Gateway API has no restrictions on the # of Gateways per namespace, therefore Contour's implementation should support >1 Contour Gateway per namespace.

skriss:
Adds support for provisioning more than one Contour+Envoy instance per namespace.

Closes #4418.

### DIFF
--- a/internal/provisioner/model/names.go
+++ b/internal/provisioner/model/names.go
@@ -1,0 +1,117 @@
+// Copyright Project Contour Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package model
+
+import (
+	"fmt"
+	"strings"
+)
+
+// ConfigMapName returns the name of the Contour ConfigMap resource.
+func (c *Contour) ConfigMapName() string {
+	return "contour-" + c.Name
+}
+
+// ContourServiceName returns the name of the Contour Service resource.
+func (c *Contour) ContourServiceName() string {
+	return "contour-" + c.Name
+}
+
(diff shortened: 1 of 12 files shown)


internal/provisioner/model/names.go
```diff
@@ -0,0 +1,117 @@
...
+
+// ContourCertsSecretName returns the name of the Contour xDS TLS certs Secret resource.
+func (c *Contour) ContourCertsSecretName() string {
+	return c.Name + "-contourcert"
+}
+
+// EnvoyCertsSecretName returns the name of the Envoy xDS TLS certs Secret resource.
+func (c *Contour) EnvoyCertsSecretName() string {
+	return c.Name + "-envoycert"
+}
```
skriss, line 59: I have a subsequent PR that flips these names around to use the Gateway name as a suffix rather than a prefix, for consistency. Will be done as part of directly generating the certs rather than using the Job.

skriss:
Rebased.

internal/provisioner/objects/configmap/configmap.go
```diff
@@ -150,6 +143,8 @@ accesslog-format: envoy
 #   Configure the number of additional ingress proxy hops from the
 #   right side of the x-forwarded-for HTTP header to trust.
 #   num-trusted-hops: 0
+# Name of the envoy service to inspect for Ingress status details.
```
sunjayBhatia, line 146: ah, one thing i forgot to check on from the previous PR was if this was synced with what we have in the example configmap vs. still the string from the operator which might be out of date

hopefully this problem goes away when we use just the contourconfiguration CRD

skriss: You mean all the commented out example config? Yeah, it is definitely missing some things, but yeah this will switch over to using ContourConfiguration instead of the ConfigMap pretty soon (I have some WIP code for it already just need the defaulting/validation revisions to land first) so I'm not worrying about it.

sunjayBhatia (approved):

internal/provisioner/objects/configmap/configmap.go
sunjayBhatia: > You mean all the commented out example config

yep

> this will switch over to using ContourConfiguration instead of the ConfigMap pretty soon

👍🏽 love it, sounds good

PR #4432, opened 2022-03-30
generate xDS TLS certs directly from provisioner

### LINKED ISSUE #4419: Gateway API: provisioner should directly provision xDS TLS certs instead of using Job
It's simpler and less error-prone to directly provision xDS TLS certs using `pkg/certs`, rather than creating a certgen Job to run `contour certgen`. One thing to keep in mind is that they should probably still be rotated once for each release.

skriss:
In the Gateway provisioner, generate xDS TLS
certs directly by using the certgen package,
rather than creating a certgen Job to create
the secrets. This is simpler and less error-
prone. Certs are still rotated with each new
release.

Also updates the certgen command/package
to allow a suffix rather than a prefix, to be
consistent with other provisioner-generated
names. Since this option was only added after
the last release, there are no backwards
compatibility issues here.

Closes #4419.


cmd/contour/certgen.go
```diff
@@ -44,7 +44,7 @@ func registerCertGen(app *kingpin.Application) (*kingpin.CmdClause, *certgenConf
 	certgenApp.Flag("certificate-lifetime", "Generated certificate lifetime (in days).").Default(strconv.Itoa(certs.DefaultCertificateLifetime)).UintVar(&certgenConfig.Lifetime)
 	certgenApp.Flag("overwrite", "Overwrite existing files or Secrets.").BoolVar(&certgenConfig.Overwrite)
 	certgenApp.Flag("secrets-format", "Specify how to format the generated Kubernetes Secrets.").Default("legacy").StringVar(&certgenConfig.Format)
-	certgenApp.Flag("secrets-name-prefix", "Specify a prefix to be used as the first part of the generated Kubernetes secrets' names.").StringVar(&certgenConfig.NamePrefix)
+	certgenApp.Flag("secrets-name-suffix", "Specify a suffix to be appended to the generated Kubernetes secrets' names.").StringVar(&certgenConfig.NameSuffix)
```
sunjayBhatia, line 47: should we do some validation this is a valid string that can be included as part of a name? (i think this may be one of the few user inputs that actually end up in a k8s resource name?)

also maybe less important, but do we expect people to start this with a `-` or do we want to insert it for them? (maybe overkill but just occurred to me)

skriss: > should we do some validation this is a valid string that can be included as part of a name? (i think this may be one of the few user inputs that actually end up in a k8s resource name?)

Probably not a bad idea; namespace should be validated as well, it's not currently. Worth noting that if you output directly to a cluster, e.g. `--kube`, you already get a decent error message from the API server when the create fails:
```
FATA[0000] failed output certificates                    error="failed to write certificates to \"foo\": Secret \"contourcert%!\" is invalid: metadata.name: Invalid value: \"contourcert%!\": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')"
```

But if you were outputting to YAML, obviously you wouldn't get this until later when you actually applied it.

> also maybe less important, but do we expect people to start this with a `-` or do we want to insert it for them? (maybe overkill but just occurred to me)

Originally when I implemented the prefix flag I had the `-` automatically added, then thought that maybe some users wouldn't want a `-`, so I switched to just literally appending the provided suffix. I'm still leaning towards not adding it automatically, just because then there's no way to get rid of it if someone doesn't want it, but open to input.

skriss: Looking at where this could be implemented, I _guess_ it'd go in the `AsSecrets` / `AsLegacySecrets` functions (https://github.com/projectcontour/contour/blob/63342278690d2aacf02b4a1b65c36a5131336de1/internal/certgen/certgen.go#L151-L207), since those are the ones that deal with generating Kubernetes resources, and also are called both by the cmd-line and by the provisioner directly.  Would require adding an error return param. Does that sound reasonable to you @sunjayBhatia?

skriss: Added validation/tests in https://github.com/projectcontour/contour/pull/4432/commits/eec57cd82b7bf02e9899e4fb4765bfb26d01dba9

internal/certgen/certgen_test.go
```diff
@@ -0,0 +1,66 @@
...
+			wantErrs:   1,
+		},
+		"invalid name suffix characters": {
+			namespace:  "valid-namespace",
+			nameSuffix: "-invalid-namesuffix-chars$",
+			wantErrs:   1,
+		},
+		"name suffix too long": {
+			namespace:  "valid-namespace",
+			nameSuffix: "this-name-suffix-is-way-too-long-and-surely-wont-be-allowed-by-the-api-server-but-wow-secret-names-are-allowed-to-be-really-long-so-this-test-data-has-to-be-extremely-lengthy-in-order-to-trigger-an-error-running-out-of-words-now-but-this-will-just-keep-going-until-i-see-a-green-check",
```
sunjayBhatia, line 45: hahaha

cmd/contour/certgen.go
sunjayBhatia: looks and works great! 👍🏽 

didnt really think about it before but since the suffix in the provisioner path comes from the gateway resource name  it all works great and should have no chance of an invalid string there

sunjayBhatia (approved):

tsaarni (approved):
LGTM!

Couple of general comments though. They are not relevant to this change, but I wanted to ask your thoughts:

I think in the long run it would be nice to see that Gateway provisioner could also support certificates provided by some other system, which does proper time-based rotation.  Long time ago I criticized that the hardcoded 365 day certgen certificates are a time-bomb waiting to destroy the deployments. Even with the upgrade based rotation, it still a gamble. Although, to my surprise I have not seen people submitting issues about that so far, so maybe the situation was not that bad :).

To the point: in case of static provisioning certgen could be easily replaced with a proper system (such as cert-manager) but it seems to me, it is not possible with dynamic provisioning yet, or with the old operator either.

Another small issue is related to the fact that we delete the CA key. New CA gets generated at every upgrade. I think it might cause a small glitch in the xDS communication before the new CA cert is taken into use by both ends. Depending on the order of pod-vs-secret creation, there is a chance that the old certs are written (by kubelet) to the new pods. It might take couple of minutes for the update to propagate via volume mounts. But at the end, I think both Contour and Envoy should pick up the updates and xDS should come up after rotation. Maybe you tested this already?

internal/certgen/certgen_test.go
skriss: Gotta have a little fun sometimes :)

skriss:
@tsaarni thanks for the review & feedback, as always excellent comments to consider!

> I think in the long run it would be nice to see that Gateway provisioner could also support certificates provided by some other system, which does proper time-based rotation. Long time ago I criticized that the hardcoded 365 day certgen certificates are a time-bomb waiting to destroy the deployments. Even with the upgrade based rotation, it still a gamble. Although, to my surprise I have not seen people submitting issues about that so far, so maybe the situation was not that bad :).
> 
> To the point: in case of static provisioning certgen could be easily replaced with a proper system (such as cert-manager) but it seems to me, it is not possible with dynamic provisioning yet, or with the old operator either.

Yep, I absolutely agree with this. It would be fairly straightforward to provide an option to use cert-manager; if we wanted a more generic extension point, we'd need to think some more about what the design would look like. At a high level, the way I'd see this working is that the `ContourDeployment` CRD (which will function as parameters for a GatewayClass) would add some options to control how certs are provisioned, and then the provisioner would look at those to decide what resources to provision, and how.

> Another small issue is related to the fact that we delete the CA key. New CA gets generated at every upgrade. I think it might cause a small glitch in the xDS communication before the new CA cert is taken into use by both ends. Depending on the order of pod-vs-secret creation, there is a chance that the old certs are written (by kubelet) to the new pods. It might take couple of minutes for the update to propagate via volume mounts. But at the end, I think both Contour and Envoy should pick up the updates and xDS should come up after rotation. Maybe you tested this already?

Another excellent point; there are almost certainly some scenarios where the different sides are not in sync for a period of time. For minor version upgrades, I *think* it should generally work well though; the updated secrets get written before the deployment/daemonset are updated, so the new pods coming up should pick up the new secrets. There is still a chance that some of the older pods could pick up an updated secret prematurely, though. And if Envoy's version is not bumped, then the existing pods would take some time for the updated secrets to propagate. I think this is definitely an area for some more R&D to be done.

PR #4436, opened 2022-03-31
test/e2e/provisioner: validate routing via provisioned instances

skriss:
Adds basic routing validation to provisioned
instances to ensure that Envoy is actually
being programmed correctly.

Note, some of this may end up being thrown away when we fold the Gateway API conformance tests into our CI since they'll cover a lot of this, but for now this is a good thing to add to ensure that not only is the Gateway CR's status being set correctly, but also that Contour<->Envoy communication is happening correctly.

### DIFF
--- a/test/e2e/provisioner/provisioner_test.go
+++ b/test/e2e/provisioner/provisioner_test.go
@@ -17,12 +17,14 @@
 package provisioner
 
 import (
+	"fmt"
 	"testing"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/projectcontour/contour/internal/gatewayapi"
 	"github.com/projectcontour/contour/test/e2e"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


test/e2e/provisioner/provisioner_test.go
```diff
@@ -91,62 +93,139 @@ var _ = Describe("Gateway provisioner", func() {
...
-								Namespaces: &gatewayapi_v1alpha2.RouteNamespaces{
-									From: gatewayapi.FromNamespacesPtr(gatewayapi_v1alpha2.NamespacesFromSame),
+			res, ok := f.HTTP.RequestUntil(&e2e.HTTPRequestOpts{
+				OverrideURL: "http://" + gateway.Status.Addresses[0].Value,
+				Host:        string(route.Spec.Hostnames[0]),
+				Path:        "/prefix/match",
+				Condition:   e2e.HasStatusCode(200),
+			})
+			require.Truef(f.T(), ok, "expected 200 response code, got %d", res.StatusCode)
+			require.NotNil(f.T(), res)
```
sunjayBhatia, line 133: nit: might need to reverse the order of 132 and 133 since the error message uses a field in the `res` variable

skriss: ah yeah good call!

PR #4438, opened 2022-03-31
clean up provisioner RBAC and reuse Contour RBAC for it

skriss:
Tidies up provisioner RBAC markers and reuses
the Contour RBAC file when generating the
provisioner RBAC YAML instead of requiring all
markers to be duplicated in the provisioner's
file.

### DIFF
--- a/internal/provisioner/objects/rbac/clusterrole/cluster_role.go
+++ b/internal/provisioner/objects/rbac/clusterrole/cluster_role.go
@@ -62,6 +62,7 @@ func desiredClusterRole(name string, contour *model.Contour) *rbacv1.ClusterRole
 	var (
 		createGetUpdate = []string{"create", "get", "update"}
 		getListWatch    = []string{"get", "list", "watch"}
+		update          = []string{"update"}
 	)
 
 	policyRuleFor := func(apiGroup string, verbs []string, resources ...string) rbacv1.PolicyRule {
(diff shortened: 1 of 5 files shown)


internal/provisioner/objects/rbac/clusterrole/cluster_role.go
```diff
@@ -87,7 +88,7 @@ func desiredClusterRole(name string, contour *model.Contour) *rbacv1.ClusterRole
 			// Gateway API resources.
 			// Note, ReferencePolicy does not currently have a .status field so it's omitted from the status rule.
 			policyRuleFor(gatewayv1alpha2.GroupName, getListWatch, "gatewayclasses", "gateways", "httproutes", "tlsroutes", "referencepolicies"),
-			policyRuleFor(gatewayv1alpha2.GroupName, createGetUpdate, "gatewayclasses/status", "gateways/status", "httproutes/status", "tlsroutes/status"),
+			policyRuleFor(gatewayv1alpha2.GroupName, update, "gatewayclasses/status", "gateways/status", "httproutes/status", "tlsroutes/status"),
```
skriss, line 91: This was reduced to just `update` to match https://github.com/projectcontour/contour/blob/main/internal/k8s/rbac.go#L23

internal/provisioner/rbac/rbac.go
```diff
@@ -13,22 +13,29 @@
...
+// ---
+
+// RBAC for core Contour resources to be provisioned.
+// +kubebuilder:rbac:groups="",resources=configmaps;secrets;services;serviceaccounts,verbs=get;list;watch;create;update;delete
+// +kubebuilder:rbac:groups=apps,resources=deployments;daemonsets,verbs=get;list;watch;create;update;delete
+// +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles;clusterrolebindings;roles;rolebindings,verbs=get;list;watch;create;update;delete
+// ---
+
+// RBAC for leader election for the provisioner.
 // +kubebuilder:rbac:groups="",resources=events,verbs=create;get;update,namespace=projectcontour
```
sunjayBhatia, line 33: should be removable? tried w/o these couple lines and it didnt change the generated role

skriss: Yeah I think this one is effectively duplicated with https://github.com/projectcontour/contour/blob/main/internal/k8s/rbac.go#L27-L29, hence why no change in the generated role, but I left it in here because the provisioner itself can do leader election, so even if Contour (for some reason) stopped doing leader election, the provisioner would still need it itself.

sunjayBhatia: ah right, good to eventually note somewhere if you want to deploy the provisioner in a custom namespace you'll have to change that part of the manifest 👍🏽

sunjayBhatia: idk if we properly documented that except maybe in release notes for the xds server itself

sunjayBhatia (approved):

Issue #4431, opened 2022-03-30
E2E flake: Gateway TLSRoute Terminate


skriss:
Have seen this one flake a few times:

https://github.com/projectcontour/contour/runs/5759542857?check_suite_focus=true#step:7:133

```
Error Trace:	tlsroute_test.go:139
  	            				suite.go:601
  	            				asm_amd64.s:1571
Error:      	Condition satisfied
Test:       	Gateway API Reconcile mode controller Gateway with one TLS listener with Mode=Terminate with namespace: gateway-tlsroute-mode-terminate with gateway gateway-tlsroute-mode-terminate/tls-terminate, controllerName: projectcontour.io/ingress-controller-378 with TLS secret gateway-tlsroute-mode-terminate/tlscert for hostname tlsroute.gatewayapi.projectcontour.io TLS requests terminate via SNI at Envoy and then are routed to a service
```

skriss:
I'm wondering if https://github.com/projectcontour/contour/blob/main/test/e2e/gateway/tlsroute_test.go#L82-L90 and https://github.com/projectcontour/contour/blob/main/test/e2e/gateway/tlsroute_test.go#L149-L158 could inadvertently be causing some test pollution by accepting any hostname; it seems like changes to Listeners are a little slower to propagate than just route changes, so perhaps these are lingering in Envoy after one test completes and causing issues for the next? Just a guess, not too sure at this point.

skriss:
This test was dropped in favor of the upstream conformance test, closing this out!

PR #4441, opened 2022-04-01
site: add idleConnection to request routing doc

tsaarni:
Updates #3313.

### DIFF
--- a/site/content/docs/main/config/request-routing.md
+++ b/site/content/docs/main/config/request-routing.md
@@ -230,6 +230,7 @@ spec:
   - timeoutPolicy:
       response: 1s
       idle: 10s
+      idleConnection: 60s
     retryPolicy:
       count: 3
       perTryTimeout: 150ms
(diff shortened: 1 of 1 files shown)


sunjayBhatia (approved):

skriss:
E2E failure on main looks like https://github.com/projectcontour/contour/issues/4431, need to dig into that one a bit but for now we can probably add the standard flake handling to it to retry inline.

tsaarni:
For the record, the error was  

```
Error Trace:	cookie_rewrite_test.go:410
			suite.go:601
			asm_amd64.s:1571
Error:      	map[string]struct {}{"echo":struct {}{}} does not contain "echo-other"
Test:       	HTTPProxy cookie-rewriting with namespace: app-cookie-rewrite cookies from app can be rewritten
  
In [It] at: /home/runner/work/contour/contour/test/e2e/httpproxy/cookie_rewrite_test.go:410
```

https://github.com/projectcontour/contour/runs/5787579923?check_suite_focus=true

Issue #4235, opened 2021-12-13
Gateway API: add Gateway.Spec.Addresses support


skriss:
ref. https://github.com/projectcontour/contour/issues/3828#issue-926892960

To support spec address, Contour should use the address when creating the Envoy service.

~~Blocked by #3545~~

skriss:
This can be at least partially supported now via the provisioner, by using the first Gateway.Spec.Addresses value to populate the Envoy service's loadBalancerIP field during creation.

PR #4443, opened 2022-04-01
basic Gateway address support

skriss:
Adds provisioner support for a single
Gateway address of type IP or hostname.
The value will be used for the Envoy
service's spec.loadBalancerIP field.

Updates #4235.

Still need to work out exactly what to do with Conditions here for the non-happy path, but wanted to put up the draft as an example of using a fake controller-runtime client for testing which I think works pretty well.


internal/k8s/statusaddress.go
```diff
@@ -168,22 +168,6 @@ func (s *StatusAddressUpdater) OnAdd(obj interface{}) {
...
-			}
-		}
-		if !ready {
-			s.Logger.
-				WithField("name", o.Name).
-				WithField("namespace", o.Namespace).
-				Debug("Gateway is not ready, not setting address")
-			return
-		}
-
```
skriss, line 186: I can't remember why I included this logic here, but it creates an issue in the case where an address is requested in the Gateway spec, because it's considered "Not Ready" until that address shows up in the status, which this would block. Tests seem OK without it so hoping its removal doesn't cause issues..

skriss:
Now reading through all the threads in https://github.com/projectcontour/contour-operator/pull/336 and realizing this implementation may be too simplistic in that it won't work for certain infrastructure providers. Leaving in draft while I do some more investigation..

skriss:
> Now reading through all the threads in [projectcontour/contour-operator#336](https://github.com/projectcontour/contour-operator/pull/336) and realizing this implementation may be too simplistic in that it won't work for certain infrastructure providers. Leaving in draft while I do some more investigation..

After looking at those thread some more, I think this still makes sense as an initial implementation that will be useful for some but not all infrastructure providers. We'll also look to add support for setting annotations on the Envoy service which, when combined with this PR, should cover most use cases.

youngnick:
Agreed with the approach, having *some* support is better than none.

skriss:
Rebased this on #4448 to make use of the testing approach.

sunjayBhatia:
needs a rebase/merge but from first look looks pretty good

sunjayBhatia (approved):

PR #4448, opened 2022-04-04
internal/provisioner: add Reconcile tests for controllers

skriss:

Replaces small unit tests of some of the individual helpers with full tests of the Reconcile() method that use a fake controller-runtime client.

### DIFF
--- a/internal/provisioner/controller/gatewayclass_test.go
+++ b/internal/provisioner/controller/gatewayclass_test.go
@@ -14,102 +14,313 @@
 package controller
 
 import (
+	"context"
 	"testing"
 
+	"github.com/go-logr/logr"
+	contourv1alpha1 "github.com/projectcontour/contour/apis/projectcontour/v1alpha1"
 	"github.com/projectcontour/contour/internal/gatewayapi"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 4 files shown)


skriss:
GatewayClass coverage should be pretty good; will look to add more scenarios to the Gateway tests (all the resource creation/etc) but figured I'd get any input on the setup first.

internal/provisioner/controller/gatewayclass_test.go
```diff
@@ -14,102 +14,310 @@
...
+			}
+
+			r := &gatewayClassReconciler{
+				gatewayController: "projectcontour.io/gateway-provisioner",
+				client:            client.Build(),
+				log:               logr.Discard(),
+			}
+
+			var req reconcile.Request
+			if tc.req != nil {
```
sunjayBhatia, line 306: nice! so its a bit less repetitive

skriss:
> maybe premature optimisation even though this file isnt that long is if you wanted to make it even more concise, could have the assertion funcs on status as a var (or even a helper that takes a type+status+reason and returns a func of appropriate assertions)

Yeah good idea, that part is pretty repetitive - see latest commit, I think that's along the lines of what you were thinking?

sunjayBhatia (approved):

PR #4451, opened 2022-04-06
Make all ContourConfig CRD fields optional

skriss:
- Makes all ContourConfig CRD fields optional
- Applies defaults internally, and overrides them with any user-specified settings
- Documents the Contour defaults in the CRD field godoc

TODOs
- [ ] remove kubebuilder enum constraints, replace with validation functions (copy from pkg/config) -- will likely do as a separate PR
- [ ] more testing


apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -55,14 +53,13 @@ type ContourConfigurationSpec struct {
...
+	HTTPProxy *HTTPProxyConfig `json:"httpproxy,omitempty"`
 
 	// EnableExternalNameService allows processing of ExternalNameServices
-	// Defaults to disabled for security reasons.
+	//
+	// Contour's default is false for security reasons.
 	// +optional
-	// +kubebuilder:default=false
-	EnableExternalNameService bool `json:"enableExternalNameService"`
+	EnableExternalNameService *bool `json:"enableExternalNameService,omitempty"`
```
skriss, line 62: I ended up making all boolean fields pointers. The reasoning is that, if we ever wanted to change the internal default for one of these to `true`, then if this weren't a pointer we'd have no way to distinguish between "user didn't provide a value for the field" and "user explicitly specified false". However, we do generally follow a pattern of boolean config fields always defaulting to false, so I'm definitely open to leaving these as non-pointers that always must default to false (it makes them easier to work with in code, fewer derefs). Looking for input.

More generally, my thinking with primitive types is - if the zero value is a valid option for the field, then the field probably should be a pointer to distinguish between "not specified" and "explicitly specified as the zero value". If the zero value is invalid, then the field doesn't need a pointer since we can assume anything with the zero value was not specified.

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -34,19 +33,18 @@ type ContourConfigurationSpec struct {
 	// Debug contains parameters to enable debug logging
 	// and debug interfaces inside Contour.
 	// +optional
-	// +kubebuilder:default={logLevel: "info", kubernetesLogLevel: 0}
-	Debug DebugConfig `json:"debug"`
+	Debug *DebugConfig `json:"debug,omitempty"`
 
 	// Health defines the endpoints Contour uses to serve health checks.
+	//
+	// Contour's default is { address: "0.0.0.0", port: 8000 }.
```
skriss, line 40: Since `HealthConfig` is a shared type, these defaults have to be documented at this level, which is a little inconsistent with other non-shared struct fields, but I think it's the best we can do?

cmd/contour/servecontext_test.go
```diff
@@ -363,288 +364,165 @@ func TestParseHTTPVersions(t *testing.T) {
...
+				Address: "0.0.0.0",
+				Port:    8000,
+			},
+		}
+	}
+
+	cases := map[string]struct {
+		getServeContext         func(ctx *serveContext) *serveContext
+		getContourConfiguration func(cfg contour_api_v1alpha1.ContourConfigurationSpec) contour_api_v1alpha1.ContourConfigurationSpec
+	}{
```
skriss, line 500: I switched this test to use the pattern of "have a base config, then each test adds its own modification to that config" which I think works really well here for clarity & maintaintability and saves us ~600 lines of copy-pasted config.

internal/contourconfig/contourconfiguration.go
```diff
@@ -17,10 +17,147 @@ import (
...
+		},
+		Envoy: &contour_api_v1alpha1.EnvoyConfig{
+			Listener: &contour_api_v1alpha1.EnvoyListenerConfig{
+				UseProxyProto:             pointer.Bool(false),
+				DisableAllowChunkedLength: pointer.Bool(false),
+				DisableMergeSlashes:       pointer.Bool(false),
+				ConnectionBalancer:        "",
+				TLS: &contour_api_v1alpha1.EnvoyTLS{
+					MinimumProtocolVersion: "1.2",
+					CipherSuites: []contour_api_v1alpha1.TLSCipherType{
```
sunjayBhatia, line 81: nit: could we consolidate with https://github.com/projectcontour/contour/blob/6add750633e3b8d20890d10a92e65e033df45b16/pkg/config/ciphersuites.go#L33 maybe?

internal/contourconfig/contourconfiguration.go
```diff
@@ -17,10 +17,147 @@ import (
...
+}
+
+func UInt32Ptr(val uint32) *uint32 {
+	return &val
+}
+
+func OverlayOnDefaults(spec contour_api_v1alpha1.ContourConfigurationSpec) (contour_api_v1alpha1.ContourConfigurationSpec, error) {
+	res := Defaults()
+
+	if err := mergo.Merge(&res, spec, mergo.WithOverride); err != nil {
```
sunjayBhatia, line 37: 💯

cmd/contour/servecontext_test.go
sunjayBhatia: yep, good stuff

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -232,13 +255,19 @@ type EnvoyConfig struct {
...
 
 	// DefaultHTTPVersions defines the default set of HTTPS
 	// versions the proxy should accept. HTTP versions are
 	// strings of the form "HTTP/xx". Supported versions are
 	// "HTTP/1.1" and "HTTP/2".
-	DefaultHTTPVersions []HTTPVersionType `json:"defaultHTTPVersions"`
+	//
+	// Contour's default list is:
+	//   - "HTTP/1.1"
+	//	 - "HTTP/2"
```
sunjayBhatia, line 268: tiny nit: spacing?

sunjayBhatia (commented):
i think the rules you described are reasonable 👍🏽 good to reduce pointer derefs/nil checks where possible and where zero values wouldn't work at all, but maintain ability to use zero values if they do make sense

skriss:
I was able to get the provisioner updated to generate a ContourConfiguration instead of a ConfigMap, based on this PR (the provisioner work is still WIP but wasn't working prior to the ContourConfig changes): https://github.com/skriss/contour/tree/gw-provisioner-contourconfig

skriss:
This one should be ready for official review, ad-hoc testing looks good as well as all the E2E's and the WIP branch for the provisioner that makes use of this (and was previously not working).

I'll tackle the dropping of the kubebuilder enum constraints in a separate PR.

skriss:
rebased.

PR #4454, opened 2022-04-07
internal/provisioner: provision a ContourConfiguration resource

### LINKED ISSUE #4421: Gateway API: provisioner should generate a ContourConfiguration instead of a ConfigMap
The provisioner should move to generating a ContourConfiguration CR, and configuring Contour to use it, instead of the ConfigMap. This will make supporting GatewayClass params easier.

skriss:
Switches from provisioning a ConfigMap to
a ContourConfiguration CRD to specify config
for Contour.

Closes #4421.


internal/provisioner/objects/contourconfig/contourconfig.go
```diff
@@ -0,0 +1,124 @@
...
+		if maybeUpdated.Spec.Envoy == nil {
+			maybeUpdated.Spec.Envoy = &contour_api_v1alpha1.EnvoyConfig{}
+		}
+		maybeUpdated.Spec.Envoy.Service = &contour_api_v1alpha1.NamespacedName{
+			Namespace: contour.Namespace,
+			Name:      contour.EnvoyServiceName(),
+		}
+
+		if !equality.Semantic.DeepEqual(current, maybeUpdated) {
+			return cli.Update(ctx, maybeUpdated)
```
sunjayBhatia, line 82: i forget what the plan around updating and owner labels are but in the future for this and other resources, but a thing to solve in the future (e.g. if someone removes owner labels what do we do)

sunjayBhatia (approved):

internal/provisioner/objects/contourconfig/contourconfig.go
skriss: Yeah, right now the existence of the owner labels is pretty essential to the provisioner controlling resources. There was some talk in the operator repo of switching to owner references, which could also work except there are some complications with cluster-scoped resources. IMHO it's OK to say that the resources must have the labels in order to be controlled by the provisioner; I do think the behavior if someone removes the labels from a managed resource may be weird, though (seems like the provisioner may try to create a new instance of the resource, which would fail because a resource with that name already exists). So probably an area for improvement.

PR #4459, opened 2022-04-11
internal/provisioner: support using ContourDeployment.Config

skriss:
When a GatewayClass has a parametersRef
to a ContourDeployment resource that has
Config specified, use that Config when
provisioning Gateways and their associated
ContourConfigurations.

### DIFF
--- a/internal/provisioner/controller/gateway.go
+++ b/internal/provisioner/controller/gateway.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 
 	"github.com/go-logr/logr"
+	contour_api_v1alpha1 "github.com/projectcontour/contour/apis/projectcontour/v1alpha1"
 	"github.com/projectcontour/contour/internal/provisioner/model"
 	"github.com/projectcontour/contour/internal/provisioner/objects/contourconfig"
 	"github.com/projectcontour/contour/internal/provisioner/objects/daemonset"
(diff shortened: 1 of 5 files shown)


internal/provisioner/controller/gateway.go
```diff
@@ -323,3 +332,22 @@ func (r *gatewayReconciler) ensureContourDeleted(ctx context.Context, contour *m
 
 	return errs
 }
+
+func (r *gatewayReconciler) getGatewayClassParams(ctx context.Context, gatewayClass *gatewayapi_v1alpha2.GatewayClass) (*contour_api_v1alpha1.ContourDeployment, error) {
+	if !isContourDeploymentRef(gatewayClass.Spec.ParametersRef) {
```
sunjayBhatia, line 337: is this case possible? since GatewayClasses won't be accepted if they have an invalid parametersref and we won't reconcile a Gateway for a GatewayClass that has not been accepted?

skriss: Yeah probably not, I think this function call is effectively just handling nil-checking on the parametersref right now. Can simplify to just do that directly.

skriss: Thinking about it some more, there probably is a (admittedly pretty hard-to-trigger) race condition here, if a GatewayClass has been updated from a valid ContourDeployment ref to a ref to some other resource type but hasn't yet reconciled / marked "Accepted: false", and ~simultaneously a Gateway reconcile is triggered that uses that GatewayClass. Based on that I'm thinking it's safer to keep the check here (it's just checking fields, no cache/API calls so not expensive to execute), but I'll add some comments.

sunjayBhatia: ah yep that is a good point about the race, yeah worth adding some comments (and maybe a test if possible) as to what is expected in this edge case

sunjayBhatia (approved):

PR #4461, opened 2022-04-11
test/e2e: fix TLSRoute test flake

skriss:
Drops parts of the TLSRoute tests that remove
all SNI filtering, to avoid flakes. Also uses
unique hostnames between the tests.

Updates #4431.

I've run the E2E's a bunch of times with this change and so far haven't seen any occurrences of that test failure, will repeat some more times on this PR.

### DIFF
--- a/test/e2e/gateway/tlsroute_test.go
+++ b/test/e2e/gateway/tlsroute_test.go
@@ -17,8 +17,6 @@
 package gateway
 
 import (
-	"context"
-	"crypto/tls"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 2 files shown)


test/e2e/gateway/tlsroute_test.go
```diff
@@ -72,37 +67,14 @@ func testTLSRoutePassthrough(namespace string) {
 		require.Never(f.T(), func() bool {
 			_, err := f.HTTP.SecureRequest(&e2e.HTTPSRequestOpts{
 				Host: "something.else.not.matching",
-				TLSConfigOpts: []func(*tls.Config){
```
skriss, line 75: this line is redundant since the SNI gets set to the Host if not otherwise specified, so just cleaning this up.

skriss:
Alright, this has run without fail a whole bunch of times (on the PR, on the branch, locally) so I'm fairly confident this was the root cause.

sunjayBhatia (approved):

PR #4462, opened 2022-04-13
validate Gateway listener protocols/ports/hostnames

skriss:
Ensures that there is at most one insecure
port and one secure port across all Gateway
listeners, and that hostnames are unique
within the groups of insecure and secure
listeners, respectively.

Leaving as a draft for now to do some more manual testing + ensure coverage is good, also feel free to poke holes in the model, definitely possible I've misinterpreted something.

### DIFF
--- a/internal/gatewayapi/listeners.go
+++ b/internal/gatewayapi/listeners.go
@@ -1,0 +1,96 @@
+// Copyright Project Contour Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package gatewayapi
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	gatewayapi_v1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+)
+
+type ValidateListenersResult struct {
+	HTTPPort                      int
+	ValidHTTPListeners            []gatewayapi_v1alpha2.Listener
+	InvalidHTTPListenerConditions map[gatewayapi_v1alpha2.SectionName]metav1.Condition
+}
+
+func ValidateListeners(listeners []gatewayapi_v1alpha2.Listener) ValidateListenersResult {
+
+	// All listeners with a protocol of "HTTP" must use the same port number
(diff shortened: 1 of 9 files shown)


skriss:
Looking pretty good in manual testing as well, marking ready for review.

internal/provisioner/controller/gateway_test.go
```diff
@@ -541,6 +542,142 @@ func TestGatewayReconcile(t *testing.T) {
...
+						{
+							Name:     "listener-3",
+							Protocol: gatewayv1alpha2.HTTPProtocolType,
+							Port:     80,
+						},
+						// listener-4 will be ignored because it's an unsupported protocol
+						{
+							Name:     "listener-4",
+							Protocol: gatewayv1alpha2.TCPProtocolType,
+							Port:     83,
```
sunjayBhatia, line 656: nit: maybe make the port 82 to make sure it only differs on protocol?

internal/gatewayapi/listeners.go
```diff
@@ -0,0 +1,136 @@
...
+// It returns the insecure & secure ports to use, as well as conditions for all invalid listeners.
+// If a listener is not in the "InvalidListenerConditions" map, it is assumed to be valid according
+// to the above rules.
+func ValidateListeners(listeners []gatewayapi_v1alpha2.Listener) ValidateListenersResult {
+	result := ValidateListenersResult{
+		InvalidListenerConditions: map[gatewayapi_v1alpha2.SectionName]metav1.Condition{},
+	}
+
+	// All listeners with a protocol of "HTTP" must use the same port number
+	// Heuristic: the first port number encountered is allowed, any other listeners with a different port number are marked "Detached" with "PortUnavailable"
```
sunjayBhatia, line 44: this may be something to call out in a doc somewhere, i dont think this specific grouping/heuristic is in the gateway docs themselves (looking at the `Listeners` field documentation here https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.Gateway)

skriss: Yeah, this is really a Contour-specific thing since we only support 2 ports total. Agree we should call this out somewhere - need to figure out where to put docs for all this, but let me add this to a new issue.

sunjayBhatia (approved):

sunjayBhatia:
looks good! simple logic for determining conflicts/choosing a port 👍🏽 seems right that the provisioner will do similar collection of listeners to program an envoy service but contour can do the more detailed validation and set status

skriss:
> seems right that the provisioner will do similar collection of listeners to program an envoy service but contour can do the more detailed validation and set status

Yeah originally i was thinking they'd both need to set status, which requires more care in terms of ensuring they're not fighting with each other (a la the discussions we were having ~18 months ago), though I think would've been doable given it's all now a single codebase, but realized that it was probably fine to just have Contour itself set the status, the provisioner just needs to find out what ports to configure. We'll see if any issues pop up with that model, but I think it works pretty well actually.

Issue #4427, opened 2022-03-29
flesh out ContourDeployment spec


skriss:
The definition of `ContourDeployment` that we have in repo is minimal:

```yaml
kind: ContourDeployment
apiVersion: projectcontour.io/v1alpha1
metadata:
  namespace: projectcontour
  name: example
spec:
  replicas: 2
  config:
    ...
```

The structure of the rest of the spec needs to be fleshed out.

One key decision to make here: originally in the [ContourConfiguration/ContourDeployment design](https://github.com/projectcontour/contour/blob/main/design/configuration-crd.md), since the `contour serve` command was going to manage Envoy, any details of the Envoy installation went in `ContourConfiguration`. However, with the current plans around the Gateway provisioner, I think it would make more sense for these Envoy details to go directly under `ContourDeployment` for the provisioner to use. One thing we could do is collect all Envoy-related options under a top-level `envoy` field, so if/when we did decide to move Envoy management into `contour serve`, we could pick that section of `ContourDeployment` up and move it into `ContourConfiguration` wholesale. A rough sketch of what that could look like, including many of the fields we know we want to support:

```yaml
kind: ContourDeployment
apiVersion: projectcontour.io/v1alpha1
metadata:
  namespace: projectcontour
  name: example
spec:
  # Settings for the Contour part of the installation.
  contour:
    replicas: 2
    nodePlacement:
      nodeSelector:
        foo: bar
      tolerations: []
  # Settings for the Envoy part of the installation.
  envoy:
    workloadType: Deployment
    networkPublishing:
      type: LoadBalancerService
      serviceAnnotations:
        foo: bar
    nodePlacement:
      nodeSelector:
        foo: bar
      tolerations: []
  # The ContourConfigurationSpec. Arguably could move under contour.
  config:
    enableExternalNameService: true    
```

skriss:
Calling this done by #4467 for now, can always review and revise before promotion to beta or GA.

PR #4467, opened 2022-04-14
define ContourDeployment v1alpha1 spec

skriss:
Consider this a design proposal in PR form; this seems
like a reasonable initial organization to me for now (it's
alpha, so we reserve the right to make breaking changes
per https://projectcontour.io/resources/deprecation-policy/)
but open to all feedback on this.

This adds most of the options that were previously
present in the `Contour` CR, but organized by
`contour` vs `envoy` settings at the top level.

Note that `envoy.networkPublishing.serviceAnnotations`
has been added in place of a provider-specific API
model for now; this is the lowest common denominator
and will give folks full flexibility to specify what they
need while we decide what we want to do about those
provider-specific items longer-term.

Updates #4427.

Only change from what I proposed in #4427 was to move the ContourConfigurationSpec under the `contour` section, so YAML would now look like:

```yaml
kind: ContourDeployment
metadata:
  namespace: projectcontour
  name: example
spec:
  # Settings for the Contour part of the installation.
  contour:
    replicas: 2
    nodePlacement:
      nodeSelector:
        foo: bar
      tolerations: []
    # The ContourConfigurationSpec.
    config:
      enableExternalNameService: true    
  # Settings for the Envoy part of the installation.
  envoy:
    workloadType: Deployment
    networkPublishing:
      type: LoadBalancerService
      serviceAnnotations:
        foo: bar
    nodePlacement:
      nodeSelector:
        foo: bar
      tolerations: []
```


sunjayBhatia (commented):
one other thing is (see comment in snippet that sort of gives it a little stutter? idk):

```
kind: ContourDeployment
metadata:
  namespace: projectcontour
  name: example
spec:
  # Settings for the Contour part of the installation.
  contour:
    replicas: 2
    nodePlacement:
      nodeSelector:
        foo: bar
      tolerations: []
    # The ContourConfigurationSpec.
    config:
      envoy: # --> another envoy substruct
        ...
      metrics:
        ...
      enableExternalNameService: true    
  # Settings for the Envoy part of the installation.
  envoy:
    workloadType: Deployment
    networkPublishing:
      type: LoadBalancerService
      serviceAnnotations:
        foo: bar
    nodePlacement:
      nodeSelector:
        foo: bar
      tolerations: []
```

only other suggestion/idea i had might be to name/separate things with a name like `deployment` vs `runtime` (or something similar) config? 

```
kind: ContourDeployment
metadata:
  namespace: projectcontour
  name: example
spec:
  deploymentSettings: # --> or "infrastructureSettings" or something
    contour:
      ...
    envoy:
      ...
  runtimeSettings: # --> this would be the contourconfigurationspec type maybe
    ...
```

but other than moving to that, this seems like a good solution

skriss:
Yeah, that is a fair point with moving the ContourConfigurationSpec under Contour. I think the deployment vs. runtime split could work - can you think of anything else that'd go under runtime, or would it just be the ContourConfigurationSpec? We could also just go back to contour, envoy, and config, I think that's more or less fine too.

sunjayBhatia:
> I think the deployment vs. runtime split could work - can you think of anything else that'd go under runtime, or would it just be the ContourConfigurationSpec

yeah i was thinking the `runtimeXXX` field would just be a naming (rather than `config`) thing to make it clearer what the field is a bit and the type would be `ContourConfigurationSpec`

no strong opinion on the nesting part, was mostly just playing around with it, just the naming of whatever the `ContourConfigurationSpec` field i think could change IMO

skriss:
> > I think the deployment vs. runtime split could work - can you think of anything else that'd go under runtime, or would it just be the ContourConfigurationSpec
> 
> yeah i was thinking the `runtimeXXX` field would just be a naming (rather than `config`) thing to make it clearer what the field is a bit and the type would be `ContourConfigurationSpec`
> 
> no strong opinion on the nesting part, was mostly just playing around with it, just the naming of whatever the `ContourConfigurationSpec` field i think could change IMO

Gotcha yeah that makes sense `config` is definitely an overloaded term.

skriss:
Did a round of updates here to do the rename you suggested @sunjayBhatia. If we're reasonably happy with this, I'd say let's merge it so I can get going on the implementations for these fields (really want to get it all into 1.21), and we can always circle back with a final API review to reorg/rename before release, that's not too taxing to do after the fact given we haven't released it yet.

skriss:
I'll do releases notes on the PR(s) that have implementations for these fields, so not bothering with one for this PR.

sunjayBhatia (approved):

PR #4472, opened 2022-04-15
implement ContourDeployment spec

skriss:


internal/provisioner/controller/gateway.go
```diff
@@ -237,25 +214,77 @@ func (r *gatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+				contourModel.Spec.NodePlacement.Contour = &model.ContourNodePlacement{
+					NodeSelector: nodePlacement.NodeSelector,
+					Tolerations:  nodePlacement.Tolerations,
+				}
+			}
+		}
+
+		if gatewayClassParams.Spec.Envoy != nil {
+			// Workload type
+			// TODO where to do validation and surface errors?
```
sunjayBhatia, line 256: Yeah this a good point, depends how we want to behave in general for these sorts of config errors that we can correct. In the rare case someone messes this field up the most strict option might be to put Accepted=false and reason InvalidParameters on the GatewayClass, but that seems a bit drastic when we can choose a default for the user that is sensible and will likely still work.

At least logging that we are going to override the mistake for now with something might be good? Unless you think it's better to be strict to start and loosen it up later.

IMO might be ok to be strict here since this one is hard to mess up, but 🤷🏽

sunjayBhatia (approved):

internal/provisioner/controller/gateway.go
skriss: Yeah, I think it's probably most conformant to validate these up-front with the GatewayClass and set Accepted=false on an invalid value. Let me take a look at doing that, shouldn't be too difficult to add I don't think.

skriss: Alright, https://github.com/projectcontour/contour/pull/4472/commits/5a8d06ca9c46625e9b16d295110a013a34fcfe83 should get the job done for now 🤞

sunjayBhatia (approved):

PR #4474, opened 2022-04-15
update default controller name to projectcontour.io/gateway-controller

skriss:
This more general controller name makes sense
given that the provisioner is just one piece
of Gateway control logic; the xDS server also
plays a major role.


sunjayBhatia (approved):
yeah i like that name change

PR #4477, opened 2022-04-16
internal/envoy: use Envoy's path-based prefix matching

### LINKED ISSUE #4442: make use of Envoy's path-based prefix matching
Now that https://github.com/envoyproxy/envoy/issues/18148 has been implemented in https://github.com/envoyproxy/envoy/pull/20145, we should look to move our Ingress and HTTPRoute implementations to use this new prefix matching mode rather than our custom regex. We can implement this once Envoy 1.22 comes out.

mmalecki:
Update envoy's `go-control-plane` to latest.
Update the tests.
Remove the "path prefix match segment with regex meta char" - it testing
possible ill-behavior of the RegExp prefix implementation.

Fixes #4442.

Signed-off-by: Maciej Małecki <me@mmalecki.com>

### DIFF
--- a/internal/envoy/v3/route.go
+++ b/internal/envoy/v3/route.go
@@ -149,10 +149,6 @@ func routeAuthzContext(settings map[string]string) *any.Any {
 	)
 }
 
-const prefixPathMatchSegmentRegex = `(?:[\/].*)*`
-
-var _ = regexp.MustCompile(prefixPathMatchSegmentRegex)
-
 // RouteMatch creates a *envoy_route_v3.RouteMatch for the supplied *dag.Route.
 func RouteMatch(route *dag.Route) *envoy_route_v3.RouteMatch {
 	switch c := route.PathMatchCondition.(type) {
--- a/internal/envoy/v3/route.go
+++ b/internal/envoy/v3/route.go
@@ -169,10 +169,8 @@ func RouteMatch(route *dag.Route) *envoy_route_v3.RouteMatch {
 		switch c.PrefixMatchType {
 		case dag.PrefixMatchSegment:
 			return &envoy_route_v3.RouteMatch{
-				PathSpecifier: &envoy_route_v3.RouteMatch_SafeRegex{
-					// Add anchor since we have a string literal prefix.
-					// Reduces regex program size so Envoy doesn't reject long prefix matches.
-					SafeRegex: SafeRegexMatch("^" + regexp.QuoteMeta(c.Prefix) + prefixPathMatchSegmentRegex),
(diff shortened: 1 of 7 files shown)


sunjayBhatia:
since this uses a feature from Envoy 1.22, we will have to hold off merging this until we have a Contour version released with example manifests that use that version of Envoy to ensure as seamless of upgrades as possible (most likely we release Contour 1.21 without this and with Envoy 1.22 and Contour 1.22 with this feature merged)

internal/envoy/v3/route.go
```diff
@@ -169,10 +169,8 @@ func RouteMatch(route *dag.Route) *envoy_route_v3.RouteMatch {
 		switch c.PrefixMatchType {
 		case dag.PrefixMatchSegment:
 			return &envoy_route_v3.RouteMatch{
-				PathSpecifier: &envoy_route_v3.RouteMatch_SafeRegex{
-					// Add anchor since we have a string literal prefix.
-					// Reduces regex program size so Envoy doesn't reject long prefix matches.
-					SafeRegex: SafeRegexMatch("^" + regexp.QuoteMeta(c.Prefix) + prefixPathMatchSegmentRegex),
```
skriss, line 175: nit: we can drop the `prefixPathMatchSegmentRegex` declaration now.

skriss (changes requested):
Thanks for this @mmalecki! Generally looks pretty good; one nit plus looks like you may need to run `go mod tidy` and commit the result.

internal/envoy/v3/route.go
mmalecki: Ah, very good point (sorry, no language server yelling at me about that yet!). Removed now.

mmalecki:
Thanks for the review @skriss. I've pushed a `go mod tidy` squashed into the initial commit (as it touches `go.{mod,sum}` already), and a commit removing `prefixPathMatchSegmentRegex` separately.

skriss:
Thanks for the updates @mmalecki, those look good. Looks like the linter's still unhappy about formatting - running `gofmt -s -w .` in the root of the project and pushing the result should do the trick.

mmalecki:
Done! Was missing some formatting indeed, apologies for the back-and-forth on this.

> Also note we end up squashing all commits in the PR down to 1 on merge to main, so don't worry too much about how they're arranged.

Ah, that's good to know! (I still prefer them neat, for review-ability, but yay for not needing to obsess about 'em ;) )

skriss (approved):
Thanks @mmalecki, LGTM!

sunjayBhatia (approved):

Issue #3807, opened 2021-06-14
Migrate Gateway API Validation Logic Out of DAG


danehans:
**Please describe the problem you have**

Step 1: Move DAG validation logic to the [validation](https://github.com/projectcontour/contour/tree/main/internal/validation) package.
Step 2: Non-Contour-specific validation should be moved [upstream](https://github.com/kubernetes-sigs/gateway-api/tree/master/apis/v1alpha2/validation).
Step 3: Add support for the upstream validating webhook.

youngnick:
Excellent plan, strong 👍 from me.

xaleeks:
Step 2 is especially important, thinking about demarcating some of the roles and responsibilities of various pieces of contour between what project Contour should be doing and what upstream gateway API should be doing.  

I tagged it for v1.18

skriss:
xref https://github.com/kubernetes-sigs/gateway-api/issues/926

youngnick:
This one should now mostly be covered by making the upstream validation webhook be part of our install, which is actually required.

See kubernetes-sigs/gateway-api#1005 for more details.

skriss:
Working on getting this done for 1.21.

skriss:
Ran into issues on Kube 1.23, filed https://github.com/kubernetes-sigs/gateway-api/issues/1125.

skriss:
Issue was fixed on Gateway API's `main`, so I'll use that for now and then will update this when we go to Gateway API 0.5.

skriss:
Tracking https://github.com/kubernetes-sigs/gateway-api/pull/1126 for a potential v0.4.3 that we could pull in for the upcoming Contour release.

PR #4489, opened 2022-04-21
Gateway API: update to v0.4.3, add validating webhook

skriss:
Updates the version of Gateway API to
v0.4.3 and adds the validating webhook
from that version to Contour's example
YAML files and E2E tests.

Closes https://github.com/projectcontour/contour/issues/3807.


Makefile
```diff
@@ -230,10 +232,17 @@ generate-crd-yaml:
 	@echo "Generating Contour CRD YAML documents..."
 	@./hack/generate-crd-yaml.sh
 
-.PHONY: generate-gateway-crd-yaml
-generate-gateway-crd-yaml:
+# TODO switch to version-specific YAML for the validating webhook
+# once v0.5.0 comes out (v0.4.x is using a certgen tool that's 
+# incompatible with Kubernetes 1.22+).
```
skriss, line 237: I'll file an issue for this; we may get a v0.4.3 with the fix too before our next Contour release.

test/e2e/gateway/invalid_forward_to_test.go
```diff
@@ -63,28 +63,6 @@ func testInvalidForwardTo(namespace string) {
 							},
 						},
 					},
-
```
skriss, line 66: This route rule without a port in the backend ref is now rejected by the webhook, so dropping it to allow the test to pass.

sunjayBhatia (approved):

skriss:
Thinking we should hold this one until there is a release-tagged image for the webhook, otherwise we're using a `latest` image tag which will not be stable over time (https://github.com/projectcontour/contour/blob/5255e0f2f8d51f06abc7a49e3e71aa95e86467ef/examples/gateway/01-admission_webhook.yaml#L64).

youngnick:
When Rob cuts a 0.4.3 Gateway API release, we will get a versioned webhook, so I agree we should hold until then.

Makefile
skriss: Now using versioned webhook YAML.

skriss:
> When Rob cuts a 0.4.3 Gateway API release, we will get a versioned webhook, so I agree we should hold until then.

This is now updated for Gateway API v0.4.3

PR #4491, opened 2022-04-23
internal/provisioner: Refactor for objects/service

izturn:
Refactor for objects/service

### DIFF
--- a/internal/provisioner/objects/service/service.go
+++ b/internal/provisioner/objects/service/service.go
@@ -151,9 +151,7 @@ func ensureServiceDeletedHelper(ctx context.Context, cli client.Client, contour 
 	if err := cli.Delete(ctx, svc); err == nil || errors.IsNotFound(err) {
 		return nil
 	}
-
 	return err
-
 }
 
 // EnsureContourServiceDeleted ensures that a Contour Service for the
--- a/internal/provisioner/objects/service/service.go
+++ b/internal/provisioner/objects/service/service.go
@@ -111,7 +111,7 @@ var (
 // EnsureContourService ensures that a Contour Service exists for the given contour.
 func EnsureContourService(ctx context.Context, cli client.Client, contour *model.Contour) error {
 	desired := DesiredContourService(contour)
-	current, err := currentContourService(ctx, cli, contour)
+	current, err := currentService(ctx, cli, contour.Namespace, contour.ContourServiceName())
 	if err != nil {
 		if errors.IsNotFound(err) {
 			return createService(ctx, cli, desired)
@@ -127,7 +127,7 @@ func EnsureContourService(ctx context.Context, cli client.Client, contour *model
 // EnsureEnvoyService ensures that an Envoy Service exists for the given contour.
 func EnsureEnvoyService(ctx context.Context, cli client.Client, contour *model.Contour) error {
 	desired := DesiredEnvoyService(contour)
-	current, err := currentEnvoyService(ctx, cli, contour)
+	current, err := currentService(ctx, cli, contour.Namespace, contour.EnvoyServiceName())
 	if err != nil {
 		if errors.IsNotFound(err) {
 			return createService(ctx, cli, desired)
@@ -319,39 +319,12 @@ func DesiredEnvoyService(contour *model.Contour) *corev1.Service {
 	return svc
 }
 
+// currentContourService returns the current Contour/Envoy Service for the provided contour.
 func currentService(ctx context.Context, cli client.Client, namespace, name string) (*corev1.Service, error) {
 	current := &corev1.Service{}
 	key := types.NamespacedName{
 		Namespace: namespace,
 		Name:      name,
-	}
-	err := cli.Get(ctx, key, current)
-	if err != nil {
-		return nil, err
-	}
-	return current, nil
-}
-
-// currentContourService returns the current Contour Service for the provided contour.
-func currentContourService(ctx context.Context, cli client.Client, contour *model.Contour) (*corev1.Service, error) {
-	current := &corev1.Service{}
-	key := types.NamespacedName{
-		Namespace: contour.Namespace,
-		Name:      contour.ContourServiceName(),
-	}
-	err := cli.Get(ctx, key, current)
-	if err != nil {
-		return nil, err
-	}
-	return current, nil
-}
-
-// currentEnvoyService returns the current Envoy Service for the provided contour.
-func currentEnvoyService(ctx context.Context, cli client.Client, contour *model.Contour) (*corev1.Service, error) {
-	current := &corev1.Service{}
-	key := types.NamespacedName{
-		Namespace: contour.Namespace,
-		Name:      contour.EnvoyServiceName(),
 	}
 	err := cli.Get(ctx, key, current)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


izturn:
ping @stevesloka @skriss ?

skriss:
Sorry for the delay, will review soon.

internal/provisioner/objects/service/service.go
```diff
@@ -108,78 +108,62 @@ var (
...
+func EnsureContourService(ctx context.Context, cli client.Client, contour *model.Contour) error {
+	return ensureServiceHelper(ctx, cli, contour, DesiredContourService, currentContourService, updateContourServiceIfNeeded)
+}
+
 // EnsureEnvoyService ensures that an Envoy Service exists for the given contour.
 func EnsureEnvoyService(ctx context.Context, cli client.Client, contour *model.Contour) error {
-	desired := DesiredEnvoyService(contour)
-	current, err := currentEnvoyService(ctx, cli, contour)
+	return ensureServiceHelper(ctx, cli, contour, DesiredEnvoyService, currentEnvoyService, updateEnvoyServiceIfNeeded)
+}
```
skriss, line 138: Personally I would probably skip this change (the EnsureContourService and EnsureEnvoyService refactoring) - yes, it removes a few lines of duplication, but it adds a bunch of indirection that may not be worth it. I would rather hold off and see if we can find an abstraction for reconciling (create/update/delete) that can be applied across a number of different resources within `internal/provisioner/objects`, not just the services.

internal/provisioner/objects/service/service.go
```diff
@@ -108,78 +108,62 @@ var (
...
+
+func ensureServiceDeletedHelper(ctx context.Context, cli client.Client, contour *model.Contour, currentSvc currentService) error {
```
skriss, line 140: Let's call this just `ensureServiceDeleted`. Also, instead of taking a `currentService` func, let's just take a namespaced name

internal/provisioner/objects/service/service.go
```diff
@@ -303,17 +287,20 @@ func DesiredEnvoyService(contour *model.Contour) *corev1.Service {
...
-			for _, p := range contour.Spec.NetworkPublishing.Envoy.NodePorts {
-				if p.PortNumber != nil {
-					for i, q := range svc.Spec.Ports {
-						if q.Name == p.Name {
-							svc.Spec.Ports[i].NodePort = *p.PortNumber
-						}
-					}
+		if len(contour.Spec.NetworkPublishing.Envoy.NodePorts) == 0 {
+			break
+		}
```
skriss, line 292: no need to explicitly check for an empty slice here, if `NodePorts` is nil/empty the subsequent loop will be a no-op anyway.

internal/provisioner/objects/service/service.go
```diff
@@ -331,12 +318,11 @@ func DesiredEnvoyService(contour *model.Contour) *corev1.Service {
 	return svc
 }
 
-// currentContourService returns the current Contour Service for the provided contour.
-func currentContourService(ctx context.Context, cli client.Client, contour *model.Contour) (*corev1.Service, error) {
+func currentContourServiceHelper(ctx context.Context, cli client.Client, namespace, name string) (*corev1.Service, error) {
```
skriss, line 321:
```suggestion
func currentService(ctx context.Context, cli client.Client, namespace, name string) (*corev1.Service, error) {
```

internal/provisioner/objects/service/service.go
```diff
@@ -369,42 +351,46 @@ func createService(ctx context.Context, cli client.Client, svc *corev1.Service)
...
+		return nil
+	}
+	_, updated := equality.ClusterIPServiceChanged(current, desired)
+	if updated {
+		if err := cli.Update(ctx, desired); err != nil {
+			return fmt.Errorf("failed to update service %s/%s: %w", desired.Namespace, desired.Name, err)
 		}
 	}
 	return nil
+
```
skriss, line 364: To be consistent with the change you made just above:
```suggestion
	_, updated := equality.ClusterIPServiceChanged(current, desired)
	if !updated {
	    return nil
	}

	
	if err := cli.Update(ctx, desired); err != nil {
		return fmt.Errorf("failed to update service %s/%s: %w", desired.Namespace, desired.Name, err)
	}
```

skriss (changes requested):
@izturn happy to accept some of these changes, but some I think we can skip for now, see inline comments.

internal/provisioner/objects/service/service.go
izturn: got it, done

izturn:
> @izturn happy to accept some of these changes, but some I think we can skip for now, see inline comments.

thx, done
btw, how to make codecov/patch happy?

internal/provisioner/objects/service/service.go
```diff
@@ -345,18 +332,14 @@ func currentContourService(ctx context.Context, cli client.Client, contour *mode
...
-	key := types.NamespacedName{
-		Namespace: contour.Namespace,
-		Name:      contour.EnvoyServiceName(),
-	}
-	err := cli.Get(ctx, key, current)
-	if err != nil {
-		return nil, err
-	}
-	return current, nil
+	return currentService(ctx, cli, contour.Namespace, contour.EnvoyServiceName())
```
skriss, line 342: let's drop these and just use the `currentService` func with the right name param where they're used, they're not adding much.

internal/provisioner/objects/service/service.go
```diff
@@ -369,42 +352,48 @@ func createService(ctx context.Context, cli client.Client, svc *corev1.Service)
...
-			updated, needed = equality.LoadBalancerServiceChanged(current, desired)
-		}
-		if needed {
-			if err := cli.Update(ctx, updated); err != nil {
-				return fmt.Errorf("failed to update service %s/%s: %w", desired.Namespace, desired.Name, err)
-			}
-			return nil
+	if !labels.Exist(current, model.OwnerLabels(contour)) {
+		return nil
+
```
skriss, line 375: nit: rm blank line

internal/provisioner/objects/service/service.go
```diff
@@ -369,42 +352,48 @@ func createService(ctx context.Context, cli client.Client, svc *corev1.Service)
...
+	// Add additional network publishing types as they are introduced.
+	default:
+		// LoadBalancerService is the default network publishing type.
+		updated, needed = equality.LoadBalancerServiceChanged(current, desired)
+	}
+	if needed {
+		if err := cli.Update(ctx, updated); err != nil {
+			return fmt.Errorf("failed to update service %s/%s: %w", desired.Namespace, desired.Name, err)
 		}
+
```
skriss, line 396: nit: rm blank line

skriss:
> btw, how to make codecov/patch happy?

Our coverage in this package is not great, I think in this case since the PR is not decreasing overall coverage, we can overlook it and it can merge even if that check is not passing.

internal/provisioner/objects/service/service.go
skriss: Sorry, what I meant here was to delete the `currentContourService` and `currentEnvoyService` function definitions, and wherever they're used, just call `currentService(...)` directly, passing in the appropriate namespaced name for the service.

izturn: done, sorry, I misunderstood it

skriss (approved):
Thanks, LGTM!

PR #4495, opened 2022-04-25
fix DisableMergeSlashes propagation, add tests

skriss:
Fixes a bug where the DisableMergeSlashes option
was not being propagated from the config file and
adds unit and E2E tests.

Co-authored-by: Máté Szabó <mszabo@fandom.com>

cc @mszabo-wikia, got this updated so we can get it in for the upcoming release. I've included you as a co-author here.


sunjayBhatia (approved):

mszabo-wikia:
Thanks!

PR #4496, opened 2022-04-25
design: Proposal for proxy2gateway tool

sunjayBhatia:


sunjayBhatia:
notes from maintainer meeting:
- possibly a generic translation abstraction to convert other resources to Gateway API (e.g. Ingress or otherwise)
- show an example with a base Gateway that already has listeners etc.
- flesh out example w/o base Gateway, maybe this is the preferred first option?
- think about the practice around where hostname goes (e.g. on Listener, on Route, both?)
  - bc HTTPProxy supports the infra team controlling model, adding hostnames to the Gateway Listener seems how we may be opinionated about this (rather than adding to the Route)
- could be useful as a controller in-cluster?
- "audit" mode, where you can see what resources will be generated cleanly, generate a report of what will apply cleanly
- testing: integration test that converts things and sees if they actually work

sunjayBhatia:
added some more to the `proxy2gateway` tests/example scenarios: https://github.com/sunjayBhatia/proxy2gateway/commit/e4e3b34bd53f069b392a0bc1cf3f45d247d5bd6d

Issue #4468, opened 2022-04-14
update Gateway API docs


skriss:
Docs for Gateway API need to be updated to account for changes over the last release, primarily covering the provisioner.

Specifics:
- Dynamic provisioning (using the provisioner) vs. static provisioning ("normal" Contour install)
- Static provisioning: configuring a controller name vs. a gateway ref
- Contour's constraints on listeners (HTTP/HTTPS/TLS only, 2 ports max, but can have multiple listeners differentiated on hostname)
  - [ ] https://github.com/projectcontour/contour/pull/4462#discussion_r850637270
- Dynamic provisioning:
  - GatewayClasses & Params
  - Reconciliation model
  - ...

sunjayBhatia:
I would add the getting started guide as well since the gateway API site refers to this page, maybe add the provisioner above the operator option (or remove the operator option completely): https://projectcontour.io/guides/gateway-api/

Rycieos:
> Contour's constraints on listeners (HTTP/HTTPS/TLS only, 2 ports max, but can have multiple listeners differentiated on hostname) 

This part did not make it into the docs. I only discovered this limitation after trying to instantiate a Gateway, and received an error message saying as such.

Not related to the missing docs, but this limitation is very disappointing, as having HTTPS and TLS endpoints on the same Load Balancer IP was the feature I was most excited by in the Gateway API, and I thought I had found a project that could deliver that. I can't find any issue in this project about supporting both HTTPRoutes and TLSRoutes on the same Gateway; does one exist already?

skriss:
@Rycieos I don't believe we have an issue for this (more generally I'd state it as supporting an arbitrary number of Listeners per Gateway). It'd be great if you could open one and add as many details as you can about your use cases!

Rycieos:
Thanks @skriss, I opened #4960

PR #4501, opened 2022-04-26
update quickstart and Gateway API guide

skriss:
Updates #4468.

Since the quickstart and guides aren't versioned, holding this until the 1.21 GA release.

### DIFF
--- a/site/content/getting-started/_index.md
+++ b/site/content/getting-started/_index.md
@@ -74,6 +74,11 @@ You should see the following:
 
 ### Option 3: Contour Gateway Provisioner (alpha)
 
+The Gateway provisioner watches for the creation of [Gateway API][31] `Gateway` resources, and dynamically provisions Contour+Envoy instances based on the `Gateway's` spec.
+Note that although the provisioning request itself is made via a Gateway API resource (`Gateway`), this method of installation still allows you to use *any* of the supported APIs for defining virtual hosts and routes: `Ingress`, `HTTPProxy`, or Gateway API's `HTTPRoute` and `TLSRoute`.
+In fact, this guide will use an `Ingress` resource to define routing rules, even when using the Gateway provisioner for installation.
+
+
 Deploy the Gateway provisioner:
 ```bash
 $ kubectl apply -f https://projectcontour.io/quickstart/gateway-provisioner.yaml
@@ -242,3 +247,4 @@ If you encounter issues, review the [troubleshooting][17] page, [file an issue][
 [28]: /guides/kind
 [29]: https://helm.sh/docs/intro/install/
 [30]: /guides/kind/#kind-configuration-file
+[31]: https://gateway-api.sigs.k8s.io/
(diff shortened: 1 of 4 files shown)


site/content/getting-started/_index.md
```diff
@@ -163,19 +169,21 @@ kubectl patch ingress httpbin -p '{"spec":{"ingressClassName": "contour"}}'
 
 Now we're ready to send some traffic to our sample application, via Contour & Envoy.
```
sunjayBhatia, line 170: maybe a future change would be to do this via the Gateway API example modified in this PR as well, but probably requires a little refactor of the flow of this doc

site/content/guides/gateway-api.md
```diff
@@ -39,80 +39,158 @@ The following prerequisites must be met before using Gateway API with Contour:
 
 ## Deploying Contour with Gateway API
 
-### Option #1: Gateway API with Contour
+Contour supports two modes of provisioning for use with Gateway API: **static** and **dynamic**.
 
-Refer to the [contour][6] design for additional background on the Gateway API implementation.
+In **static** provisioning, the platform operator defines a `Gateway` resource, and then manually deploys a Contour instance corresponding to that `Gateway` resource. It is up to the platform operator to ensure that all configuration matches between the `Gateway` and the Contour/Envoy resources.
```
sunjayBhatia, line 44: should there be some mention of the GatewayClass controller string? since the example for Option 1 uses that method

site/content/guides/gateway-api.md
```diff
@@ -127,23 +205,67 @@ httproute.gateway.networking.k8s.io/kuard   ["local.projectcontour.io"]
...
-$ curl -H "Host: local.projectcontour.io" -s -o /dev/null -w "%{http_code}" "http://$GATEWAY/"
+$ curl -i http://local.projectcontour.io:8888
+```
+You should receive a 200 response code along with the HTML body of the main `kuard` page.
+
+You can also open http://local.projectcontour.io:8888/ in a browser.
+
+## Next Steps
+
+### Customizing your dynamically provisioned Contour instances
```
sunjayBhatia, line 225: nice!

sunjayBhatia (approved):
some small discussion/comments, overall LGTM 👍🏽

site/content/getting-started/_index.md
skriss: Yeah actually what's interesting about how it is right now is that even if we're using the Gateway provisioner to provision Contour/Envoy, we still use Ingress to define routing rules. I should call that out a bit more explicitly, to make it clear that using the Gateway provisioner for provisioning does not mean you have to use HTTPRoute/TLSRoute for defining routes. If we did want to use `HTTPRoute`, then yeah we'd have to split up some of the "Test it out" section depending on the installation method.

site/content/guides/gateway-api.md
skriss: Yeah good call, let me work that in.

skriss:
Eventually it'd be nice to have a whole section of docs for Gateway API so we can elaborate on some of this in more detail, but that will require some planning to figure out how to structure, plus we don't want to duplicate everything from https://gateway-api.sigs.k8s.io/.

youngnick (approved):
This looks great to me!

PR #4504, opened 2022-04-29
WIP: Release 1.21 cherry picks

sunjayBhatia:
Mostly to use as a sanity check/to collect the changes together, might not need to actually use this PR to merge them since we may rather want separate commits for each, so we can use this branch to merge separately and push the `release-1.21` branch

Includes:
- https://github.com/projectcontour/contour/pull/4486
- https://github.com/projectcontour/contour/pull/4505
- https://github.com/projectcontour/contour/pull/4508
- https://github.com/projectcontour/contour/pull/4509

PR #4505, opened 2022-04-29
fix Gateway API hostname intersection logic

skriss:
Updates the Gateway API hostname intersection
logic to match the spec by ignoring any route
hostnames that don't intersect with the listener
hostname, dropping support for "*" as a
valid hostname and improving the reason and
message on the "Accepted: false" route condition
when there is no intersection.

Ref. https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1alpha2.HTTPRoute (hostnames docs):

>  If a hostname is specified by both the Listener and HTTPRoute, there must be at least one intersecting hostname for the HTTPRoute to be attached to the Listener. For example:
> 
> - A Listener with test.example.com as the hostname matches HTTPRoutes that have either not specified any hostnames, or have specified at least one of test.example.com or *.example.com.
> - A Listener with *.example.com as the hostname matches HTTPRoutes that have either not specified any hostnames or have specified at least one hostname that matches the Listener hostname. For example, test.example.com and *.example.com would both match. On the other hand, example.com and test.example.net would not match.
> 
> If both the Listener and HTTPRoute have specified hostnames, any HTTPRoute hostnames that do not match the Listener hostname MUST be ignored. For example, if a Listener specified *.example.com, and the HTTPRoute specified test.example.com and test.example.net, test.example.net must not be considered for a match.
> 
> If both the Listener and HTTPRoute have specified hostnames, and none match with the criteria above, then the HTTPRoute is not accepted. The implementation must raise an ‘Accepted’ Condition with a status of False in the corresponding RouteParentStatus.


youngnick (approved):
Love it, nice work.

Could you take the comment about what to do if syntactically invalid hostnames make it through back to the gateway-api repo?

skriss:
> Could you take the comment about what to do if syntactically invalid hostnames make it through back to the gateway-api repo?

Will do - looking at it again I realized it is specified for Listeners, just not for Routes, so I added the proper Listener condition setting and addressed one of my TODOs.

sunjayBhatia (approved):

skriss:
Everyone OK with including this in 1.21 as a bug fix?

sunjayBhatia:
> Everyone OK with including this in 1.21 as a bug fix?

yeah sounds good to me 👍🏽

sunjayBhatia:
added to https://github.com/projectcontour/contour/pull/4504

PR #4512, opened 2022-05-04
internal/dag: check routes for intersecting hostnames across all listeners

### LINKED ISSUE #4507: Gateway API: routes are marked "Accepted: false" if their hostnames don't intersect with any one of the listeners they select
Even with https://github.com/projectcontour/contour/pull/4505 in place, there is still one more issue with the Gateway API hostname intersection logic. If there's a Gateway with e.g. multiple HTTP listeners, each with a different hostname, and an HTTPRoute selects the entire Gateway but doesn't have intersecting hostnames with _any one_ of those listeners, the route will get marked "Accepted: false" due to the non-intersecting hostnames, even if that HTTPRoute _does_ intersect with some other listener.

Per Gateway API spec:

> For the purpose of status, an attachment is considered successful if at least one section in the parent resource accepts it. For example, Gateway listeners can restrict which Routes can attach to them by Route kind, namespace, or hostname. If 1 of 2 Gateway listeners accept attachment from the referencing Route, the Route MUST be considered successfully attached. If no Gateway listeners accept attachment from this Route, the Route MUST be considered detached from the Gateway.

We need to reorganize our processing loops in the Gateway API processor, so that instead of going listener-by-listener, we collect *all* listeners that each route potentially selects, and then checking across that entire set of listeners to see if there's an intersecting hostname or not.

skriss:
Checks HTTPRoutes and TLSRoutes for intersecting
hostnames across all Listeners, and only sets an
"Accepted: false" condition if the route does not
intersect with any Listener.

Closes #4507.


skriss:
This now passes the WIP Gateway API conformance test I'm working on, that originally detected this issue.

sunjayBhatia (approved):
LGTM, only change would maybe be to add a comment on `computeListener` that it collects routes that are selected by/selct the listener in addition to what it did before (set status on num attached routes, validate tls etc.)

skriss:
> LGTM, only change would maybe be to add a comment on `computeListener` that it collects routes that are selected by/selct the listener in addition to what it did before (set status on num attached routes, validate tls etc.)

:+1: done

PR #4486, opened 2022-04-19
cmd/contour: Add support for json log formatter

tsaarni:
This change enables Contour logs in JSON format as discussed in https://github.com/projectcontour/contour/issues/4414#issuecomment-1087109701.

The PR proposes new command line argument `--log-format` over generally preferred configuration CRD or config file field. The reasoning for new command line argument is to allow setting formatter for all sub-commands, not only `serve`, and setting it as early as possible, before CR is fetched from the API server or before config file is read from filesystem.

Fixes #4414 

Signed-off-by: Tero Saarni <tero.saarni@est.tech>

PR #4518, opened 2022-05-09
rename DebugLogLevel to LogLevel

skriss:

### DIFF
--- a/apis/projectcontour/v1alpha1/contourconfig.go
+++ b/apis/projectcontour/v1alpha1/contourconfig.go
@@ -319,14 +319,14 @@ type DebugConfig struct {
 	// +optional
 	Port int `json:"port,omitempty"`
 
-	// DebugLogLevel defines the log level which Contour will
+	// LogLevel defines the log level which Contour will
 	// use when outputting log information.
 	//
 	// Values: `info` (default), `debug`.
 	//
 	// Other values will produce an error.
 	// +optional
-	DebugLogLevel LogLevel `json:"logLevel,omitempty"`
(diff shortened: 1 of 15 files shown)


tsaarni (commented):
This change itself looks OK to me, but I started to wonder: where is `ContourConfiguration.Spec.Debug.(Debug)LogLevel` used?  The orignal command line switch `--debug` and the configuration file field, which are booleans, were [used here](https://github.com/projectcontour/contour/blob/3fae0d635fffb323063e40d7c2278ce2d16c61e9/cmd/contour/contour.go#L121-L123). To me it looks like the CR field is not used at all 🤔

skriss:
> This change itself looks OK to me, but I started to wonder: where is `ContourConfiguration.Spec.Debug.(Debug)LogLevel` used? The orignal command line switch `--debug` and the configuration file field, which are booleans, were [used here](https://github.com/projectcontour/contour/blob/3fae0d635fffb323063e40d7c2278ce2d16c61e9/cmd/contour/contour.go#L121-L123). To me it looks like the CR field is not used at all 🤔
> 
> The CR field might have been left unimplemented because of similar problem as in #4486 - the CR can be fetched relatively late in the startup process, after logs were already needed.

That is a great point, it indeed does not seem to be being used. I only noticed the wording when making unrelated changes to the CRD but didn't actually look at its usage. Let me look at it a bit more and see if it makes sense to remove instead.

skriss:
Closing this one, will submit a new PR to remove these unused ContourConfiguration fields.

PR #4526, opened 2022-05-12
add HTTPDirectResponsePolicy 

### LINKED ISSUE #2515: Add a "route destination" to directly return a status code with body , for example "403/forbidden"
Most of my workloads expose a `/healthz` , `/readiness` , `/metrics` endpoint on the same port as the main service ( Which i understand is not great but i have no way of changing ) 

This causes every publicly exposed workload to also expose those internal endpoint to the internet, I would like a way to specify a way to return `403 forbidden` directly as a route destination ( similar to what [gloo does](https://docs.solo.io/gloo/latest/guides/traffic_management/request_processing/direct_response_action/) ) 

probably related to https://github.com/projectcontour/contour/issues/1627

a syntax like :
```
routes:
     - conditions:
      - prefix: /readiness
      directRouteAction:
          status: 403
          body: "forbidden"
```

At the moment my *workaround* is to run an httpbin service and use `httproxy inclusion` and `pathrewrite` to achieve this 
```
---
kind: HTTPProxy
metadata:
  name: forbid-generic-403
  namespace: httpbin
  annotations:
    projectcontour.io/ingress.class: contour
spec:
  routes:
    - services:
        - name: httpbin
          port: 8080
      pathRewritePolicy:
        replacePrefix:
        - replacement: /status/403
---
kind: HTTPProxy
metadata:
  name: podinfo-tls
  annotations:
    projectcontour.io/ingress.class: contour
spec:
  virtualhost:
    fqdn: podinfo-tls.example.com
    tls:
      secretName: wildcard
  includes:
    - name: forbid-generic-403
      namespace: httpbin
      conditions:
      - prefix: /metrics
    - name: forbid-generic-403
      namespace: httpbin
      conditions:
      - prefix: /healthz
    - name: forbid-generic-403
      namespace: httpbin
      conditions:
      - prefix: /readiness
    #- name: forbid
    #  namespace: httpbin
  routes:
    - services:
        - name: podinfo
          port: 8443
```

yangyy93:
1、add a HTTPDirectResponsePolicy to the HTTPProxy spec which allows for HTTPProxy.Routes to specify a directResponsePolicy. This policy will allow a direct response to be configured for a specific set of Conditions within a single route.
2、must set exactly one of route.services or route.requestRedirectPolicy or route.directResponsePolicy
fix #2515

Signed-off-by: yy <yang.yang@daocloud.io>


test/e2e/httpproxy/httpproxy_test.go
```diff
@@ -104,8 +104,8 @@ var _ = Describe("HTTPProxy", func() {
 	AfterEach(func() {
 		require.NoError(f.T(), f.Deployment.StopLocalContour(contourCmd, contourConfigFile))
 	})
+	f.NamespacedTest("httpproxy-direct-response-policy", testDirectResponseRule)
 
-	f.NamespacedTest("httpproxy-request-redirect-policy", testRequestRedirectRule)
```
skriss, line 108: Looks like maybe this was removed inadvertently, if so please add back.

internal/featuretests/v3/directresponsepolicy_test.go
```diff
@@ -0,0 +1,128 @@
+// Copyright Project Contour Authors
```
skriss, line 1: This file needs to be formatted with `goimports` - `goimports -w internal/featuretests/v3/directresponsepolicy_test.go` should do it.

skriss (commented):
Thanks for the PR @yangyy93! I will review in detail soon, but a couple initial things to get the basic CI checks passing. 

In addition to the specific comments below, please run `make generate` and commit the result.

test/e2e/httpproxy/request_redirect_test.go
```diff
@@ -43,15 +43,6 @@ func testRequestRedirectRuleNoService(namespace string) {
...
 
-func testRequestRedirectRule(namespace string) {
-	Specify("redirects can be specified on route rule", func() {
-		t := f.T()
-
-		proxy := getRedirectHTTPProxy(namespace, false)
-		doRedirectTest(namespace, proxy, t)
-	})
-}
-
```
skriss, line 54: Per previous comment, this needs to be added back along with the call to it in `httpproxy_test.go`.

internal/dag/builder_test.go
```diff
@@ -10298,6 +10280,139 @@ func TestDAGInsert(t *testing.T) {
 								PortNumber: 443,
 								StatusCode: 301,
 							},
+						},
+					)),
+				},
+			),
+		},
+		"HTTPProxy request DirectResponse policy - code 200": {
```
skriss, line 10288:
```suggestion
		"HTTPProxy DirectResponse policy - code 200": {
```

internal/dag/builder_test.go
```diff
@@ -10298,6 +10280,139 @@ func TestDAGInsert(t *testing.T) {
...
+							DirectResponse: &DirectResponse{
+								StatusCode: 200,
+								Body:       "success",
+							},
+						},
+					)),
+				},
+			),
+		},
+		"HTTPProxy request DirectResponse policy - no body": {
```
skriss, line 10328:
```suggestion
		"HTTPProxy DirectResponse policy - no body": {
```

internal/dag/builder_test.go
```diff
@@ -10298,6 +10280,139 @@ func TestDAGInsert(t *testing.T) {
...
+							PathMatchCondition: prefixString("/"),
+							DirectResponse: &DirectResponse{
+								StatusCode: 503,
+							},
+						},
+					)),
+				},
+			),
+		},
+		"HTTPProxy request DirectResponse policy with multiple matches": {
```
skriss, line 10366:
```suggestion
		"HTTPProxy DirectResponse policy with multiple matches": {
```

internal/dag/httpproxy_processor.go
```diff
@@ -468,6 +469,11 @@ func (p *HTTPProxyProcessor) computeRoutes(
 	}
 
 	for _, route := range proxy.Spec.Routes {
+		if err := routeActionCountValid(route); err != nil {
+			validCond.AddError(contour_api_v1.ConditionTypeRouteError, "routeActionCountNotValid", err.Error())
```
skriss, line 473:
```suggestion
			validCond.AddError(contour_api_v1.ConditionTypeRouteError, "RouteActionCountNotValid", err.Error())
```

apis/projectcontour/v1/httpproxy.go
```diff
@@ -377,6 +377,23 @@ type Route struct {
...
+	// StatusCode is  the HTTP response status to be returned.
+	// +required
+	// +kubebuilder:validation:Minimum=200
+	// +kubebuilder:validation:Maximum=599
+	StatusCode *int `json:"statusCode"`
+
+	// Body is the content of the response body.
+	// If this setting is omitted, no body is included in the generated response.
+	// +optional
+	Body *string `json:"body,omitempty"`
```
skriss, line 396: This can be changed to just a `string` instead of a string pointer, since we don't need to distinguish between "not supplied" and an empty string.

internal/dag/status_test.go
```diff
@@ -2623,6 +2623,64 @@ func TestDAGStatus(t *testing.T) {
...
+		},
+	}
+	run(t, "No routeAction specified is invalid", testcase{
+		objs: []interface{}{noRouteAction},
+		want: map[types.NamespacedName]contour_api_v1.DetailedCondition{
+			{Name: noRouteAction.Name, Namespace: noRouteAction.Namespace}: fixture.NewValidCondition().
+				WithError(contour_api_v1.ConditionTypeRouteError, "routeActionCountNotValid",
+					"must set exactly one of route.services or route.requestRedirectPolicy or route.directResponsePolicy"),
+		},
+	})
```
skriss, line 2650: This test case looks identical to "invalid HTTPProxy due to empty route.service", let's drop this new one and just revise the existing one's name to indicate that it covers no route actions.

test/e2e/httpproxy/direct_response_test.go
```diff
@@ -0,0 +1,99 @@
...
+		t := f.T()
+		proxy := getDirectResponseHTTPProxy(namespace)
+		doDirectTest(namespace, proxy, t)
+	})
+}
+
+func doDirectTest(namespace string, proxy *contour_api_v1.HTTPProxy, t GinkgoTInterface) {
+	f.Fixtures.Echo.Deploy(namespace, "echo")
+
+	f.CreateHTTPProxyAndWaitFor(proxy, e2e.HTTPProxyValid)
```
skriss, line 41: I know we haven't done this consistently elsewhere, but let's do it right here:
```suggestion
	_, ok := f.CreateHTTPProxyAndWaitFor(proxy, e2e.HTTPProxyValid)
	require.True(f.T(), ok)
```

skriss (changes requested):
Thanks again for the PR @yangyy93, overall it looks good, mainly just a bunch of little nits to address.

yangyy93:
@skriss Thanks again for your review, I have optimized the code following some of your suggestions

test/e2e/httpproxy/request_redirect_test.go
skriss: Ah, got it, that makes sense.

skriss (approved):
LGTM

apis/projectcontour/v1/httpproxy.go
```diff
@@ -377,6 +377,23 @@ type Route struct {
...
+type HTTPDirectResponsePolicy struct {
+	// StatusCode is the HTTP response status to be returned.
+	// +required
+	// +kubebuilder:validation:Minimum=200
+	// +kubebuilder:validation:Maximum=599
+	StatusCode int `json:"statusCode"`
+
+	// Body is the content of the response body.
+	// If this setting is omitted, no body is included in the generated response.
+	// +optional
```
sunjayBhatia, line 395: should we set a kubebuilder validation max length on this? something somewhat arbitrary for now to be defensive?

yangyy93: > should we set a kubebuilder validation max length on this? something somewhat arbitrary for now to be defensive?

@sunjayBhatia It is a good idea to limit the maximum length of the the direct response body, do you have any good suggestions for this?

yangyy93: > should we set a kubebuilder validation max length on this? something somewhat arbitrary for now to be defensive?

What do you think about setting the max length to 1048576？

yangyy93: > hm you may hit the default etcd/k8s max object size of 1mb before getting to that 1mb limit on a string field, which does limit how large of a body you can serve anyways
> 
> since contour will have to load that resource into memory and also send that over to envoy, maybe it would be better to discourage such a large body to start and in fact choose something quite a bit smaller, since it can significantly impact resource usage, especially in a multitenant installation of contour

@sunjayBhatia  Yes, 1mb is sometimes a waste of resources for configuration distribution. But if this value is set too small, it may not meet the requirements.
This requires us to strike a balance between the use of resources and the user experience.

skriss: I think a high-ish limit plus a note in the field description makes sense -- it's worth specifically calling out that the contents will have to be sent over xDS *every time* there is any config change, since we're doing SOTW, so could lead to increased latency & resource utilization. I do see your point @sunjayBhatia about a multi-tenant install, but on balance I guess I'd lean towards letting folks have a large response if they need it, provided we've documented what they should watch out for.

Maybe the best of all worlds would be to have a configurable size limit, so Contour admins could limit the max size to something smaller if needed. Probably could be a follow-on enhancement though, if we wanted to do this.

yangyy93: Adding a note to this field is a great way, I will add relevant note in this field. It can be discussed later whether it needs to be limited by configuration or gatekeeper.

yangyy93: ping @sunjayBhatia @skriss

sunjayBhatia: added https://github.com/projectcontour/contour/issues/4547 for a follow up

sunjayBhatia (approved):

PR #4532, opened 2022-05-17
site/guides: fix incorrect resource URLs

zwtop:


skriss (approved):
Thanks for catching this @zwtop, LGTM

PR #4534, opened 2022-05-17
apis: remove ContourConfiguration DebugLogLevel and KubernetesDebugLogLevel

skriss:
Removes these two fields from the ContourConfiguration
spec as they were unused since they are required to be
specified via CLI flag.


sunjayBhatia (commented):
looks like one more removal needed in the e2e tests otherwise lgtm

sunjayBhatia (approved):

PR #4529, opened 2022-05-16
test/e2e: e2e tests can be run against an ipv6 cluster

sunjayBhatia:
Need to set the CONTOUR_E2E_LOCAL_HOST to the correct IP and set
IPV6_CLUSTER=true

Also works with USE_CONTOUR_CONFIGURATION_CRD=true

Fixes: #3284

PR #4535, opened 2022-05-17
internal/envoy/v3: Use path_config_source for config source specifier

sunjayBhatia:
"path" config source is deprecated for path_config_source which also
adds the option of a watched directory that can be for different uses

Noticed when working on https://github.com/projectcontour/contour/pull/4529


skriss (approved):
LGTM, looks like we hit a flake in the merge slash test though :/ thought we had that fixed but guess not.

PR #4544, opened 2022-05-20
internal/dag: Fix private key validation logic

sunjayBhatia:
Make sure we don't return early on parsing a valid PKCS1 RSA or PKCS8
key


skriss (approved):

skriss:
Gateway API conformance expected to fail until the upstream test bug is fixed, LGTM 👍

Issue #4543, opened 2022-05-20
Test failing with latest gateway-api conformance suite


tsaarni:
CI got broken after some of the [recent changes](https://github.com/kubernetes-sigs/gateway-api/pulls?q=is%3Apr+is%3Aclosed) in gateway-api conformance test suite. Following error now happens consistently e.g. https://github.com/projectcontour/contour/runs/6525677808?check_suite_focus=true#step:7:426

```
        httproute-invalid-reference-policy.go:93: 
            	Error Trace:	helpers.go:243
            	            				helpers.go:140
            	            				httproute-invalid-reference-policy.go:93
            	            				suite.go:185
            	            				suite.go:140
            	Error:      	Received unexpected error:
            	            	timed out waiting for the condition
            	Test:       	TestConformance/HTTPRouteInvalidReferencePolicy
            	Messages:   	error waiting for HTTPRoute to have parents matching expectations
```

skriss:
I'm taking a look at this, looks like a combo of a fun upstream change and something we now need to update in Contour:

- https://github.com/kubernetes-sigs/gateway-api/pull/1170 (conformance test bugs)
- https://github.com/kubernetes-sigs/gateway-api/pull/1151 (change to spec that we need to implement)

skriss:
Upstream PR has been merged and we're back to green: https://github.com/projectcontour/contour/runs/6565697126?check_suite_focus=true

PR #4545, opened 2022-05-20
Gateway API: return a 404 when no valid backend refs

skriss:
The Gateway API spec has been updated to
require returning a 404 instead of a 503
when there are no valid backend refs on
an HTTPRoute rule. Updates Contour to match
the revised spec.

Gateway API ref: https://github.com/kubernetes-sigs/gateway-api/pull/1151

Updates #4543.


sunjayBhatia (approved):

skriss:
Gateway API conformance expected to fail until the test bug is fixed, ignoring.

PR #4548, opened 2022-05-24
cmd/contour: change shutdown --check-delay default to 0s

### LINKED ISSUE #4422: consider changing "contour envoy shutdown" --check-delay default to 0s
Currently, the `contour envoy shutdown` command has a default `--check-delay` value of `60s`. This means that, after triggering a deletion of Envoy, nothing happens for 60 seconds, at which point the command starts polling for remaining open connections.

I don't see any reason not to start checking for open connections right away. In the case where are no open connections, this means the Envoy pod can shut down immediately, rather than having to wait 60 seconds. The polling continues indefinitely, until there are few enough connections (default: none) to shutdown, so not sure what the delayed start is buying us.

skriss:
Changes the default check delay on the
shutdown command from 60s to 0s, since there
is no harm in checking immediately for
all connections to be closed, and it may
lead to faster shutdown when there are no
connections.

Closes #4422.

Reviewers, please check me on logic here; I don't see any harm in starting to check for no open connections immediately, but maybe I'm missing something.

### DIFF
--- a/changelogs/unreleased/4548-skriss-small.md
+++ b/changelogs/unreleased/4548-skriss-small.md
@@ -1,0 +1,1 @@
+Changes the `contour envoy shutdown` command's `--check-delay` default to `0s` from `60s`, allowing Envoy pods to shut down more quickly when there are no open connections.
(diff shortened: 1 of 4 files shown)


stevesloka (approved):
/lgtm 

Was originally meant to be a tunable knob, but will still work for any user by defaulting smaller. 🚢

PR #4551, opened 2022-05-27
Bump gopkg.in/yaml.v3 to address CVE-2022-28948

tsaarni:
Update gopkg.in/yaml.v3 to v3.0.1 to address CVE-2022-28948.

PR #4552, opened 2022-05-31
pkg/config: Migrate from gopkg.in/yaml.v2 to v3

tsaarni:
Use gopkg.in/yaml.v3 for config file parsing.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,8 +30,7 @@ require (
 	google.golang.org/grpc v1.45.0
 	google.golang.org/protobuf v1.28.0
 	gopkg.in/alecthomas/kingpin.v2 v2.2.6
-	gopkg.in/yaml.v2 v2.4.0
-	gopkg.in/yaml.v3 v3.0.1 // indirect
+	gopkg.in/yaml.v3 v3.0.1
 	k8s.io/api v0.24.0
 	k8s.io/apiextensions-apiserver v0.24.0
 	k8s.io/apimachinery v0.24.0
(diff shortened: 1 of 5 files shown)


tsaarni:
@skriss after seeing https://github.com/projectcontour/contour/pull/4551#issuecomment-1142226645 I wanted to test if we can remove `gopkg.in/yaml.v2` and include only one version in the binary. But the change in this PR did not remove it, still there in the compiled binary

```console
$ go version -m contour | grep gopkg.in/yaml
        dep     gopkg.in/yaml.v2        v2.4.0  h1:D8xgwECY7CYvx+Y2n4sBz93Jn9JRvxdiyyo8CTfuKaY=
        dep     gopkg.in/yaml.v3        v3.0.1  h1:fxVm/GzAzEWqLHuvctI91KS9hhNmmWOoWu0XTYJS7CA=
```

I wanted to find out why, but I do not understand the results:

```console
$ go mod why gopkg.in/yaml.v2
# gopkg.in/yaml.v2
github.com/projectcontour/contour/hack/release
sigs.k8s.io/kustomize/kyaml/yaml
gopkg.in/yaml.v2
```

Any clues what is going on?

tsaarni:
With `go mod graph` I see the full graph with numerous dependencies to `gopkg.in/yaml.v2` from different packages, but not in optimal format to track which one(s) are bringing it to the binary. If I'd find the relevant ones, I could check if there is update that has also migrated to `v3`. But there is so many dependencies that it might not be very probable we get rid of it completely.

skriss (approved):

PR #4557, opened 2022-06-06
Gateway API: add request mirror filter support

### LINKED ISSUE #3557: Support HTTPRoute requestMirror RouteFilterType [extended]
Support HTTPRoute requestMirror RouteFilterType when defined on an HTTPRoute.

//ref:https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.HTTPRouteFilter

(Support: Extended)

sepaper:
Adds support for the "RequestMirror" HTTPRoute
filter type at the rule level.

Closes #3557.

Signed-off-by: Sejong Kim <sepaper@naver.com>


test/e2e/gateway/request_mirror_test.go
```diff
@@ -0,0 +1,93 @@
...
+		mirrorLogRegexp := regexp.MustCompile(`Echoing back request made to \/mirror to client`)
+
+		logs, err := f.Fixtures.Echo.DumpEchoLogs(namespace, "echo-shadow")
+		require.NoError(f.T(), err)
+		for _, log := range logs {
+			if mirrorLogRegexp.MatchString(string(log)) {
+				mirrored = true
+			}
+		}
+		require.True(t, mirrored, "expected the request to be mirrored")
```
skriss, line 91: Instead of using a static sleep here, you could move this logic within a `require.Eventually(...)` call - we use that in a few other places in the E2E's that you can use as a reference

skriss (commented):
Thanks for the PR @sepaper! It looks pretty good at first glance, one suggestion on the E2E test, but I'll do a more thorough review tomorrow.

skriss:
You'll also need to add a changelog file, see https://github.com/projectcontour/contour/blob/main/CONTRIBUTING.md#commit-message-and-pr-guidelines for details.

sepaper:
Hi, @skriss. Thanks for the review and suggestion!
I missed the necessity of the waiting time for echo deployment availability, so I added some logic for that. Thanks to your suggestion, I could easily implement that. Thank you so much :)

internal/dag/gatewayapi_processor.go
```diff
@@ -1009,9 +1010,31 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1alpha2.HTTPRo
...
 					redirect = filter.RequestRedirect
 				}
+			case gatewayapi_v1alpha2.HTTPRouteFilterRequestMirror:
+				// Get the mirror filter if there is one. If there are more than one
+				// mirror filters, "NotImplemented" condition on the Route is set to
+				// status: True, with the "NotImplemented" reason.
+				if mirrorPolicy != nil {
+					routeAccessor.AddCondition(status.ConditionNotImplemented, metav1.ConditionTrue, status.ReasonNotImplemented, "HTTPRoute.Spec.Rules.Filters: Only one mirror filter is supported.")
+					continue
+				}
```
skriss, line 1020: Ultimately I think it makes sense to support multiple RequestMirror filters per rule, since the semantics seem pretty clear: if you have >1 RequestMirror filter, then each one defines a backend to mirror traffic to. Envoy easily supports this (see `request_mirror_policies` in https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-routeaction), and the change to Contour should be pretty straightforward too: change the `dag.Route.MirrorPolicy` to be a slice and then update the places that use it appropriately.

I'm OK if we log an issue and tackle this improvement separately, or if you want to include it in this PR, great!

test/e2e/fixtures.go
```diff
@@ -26,9 +27,13 @@ import (
 	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
skriss, line 30: nit, let's use the `corev1` alias for this pkg where it's used so we avoid the double import.

test/e2e/fixtures.go
```diff
@@ -180,6 +186,65 @@ func (e *Echo) DeployN(ns, name string, replicas int32) func() {
...
+	config, err := clientcmd.BuildConfigFromFlags("", e.kubeConfig)
+	if err != nil {
+		return nil, err
+	}
+	coreClient, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+
+	pods := new(v1.PodList)
```
skriss, line 215: nit:
```suggestion
	pods := new(corev1.PodList)
```

test/e2e/fixtures.go
```diff
@@ -180,6 +186,65 @@ func (e *Echo) DeployN(ns, name string, replicas int32) func() {
...
+	podListOptions := &client.ListOptions{
+		LabelSelector: labels.SelectorFromSet(map[string]string{"app.kubernetes.io/name": name}),
+		Namespace:     ns,
+	}
+	if err := e.client.List(context.TODO(), pods, podListOptions); err != nil {
+		return nil, err
+	}
+
+	podLogOptions := &v1.PodLogOptions{
```
skriss, line 224: nit:
```suggestion
	podLogOptions := &corev1.PodLogOptions{
```

skriss (changes requested):
Thanks again for the PR @sepaper, nice work, particularly with thorough testing!

test/e2e/fixtures.go
sepaper: done

test/e2e/fixtures.go
sepaper: done

test/e2e/fixtures.go
sepaper: done

internal/dag/gatewayapi_processor.go
skriss: ref. #4566

skriss (approved):
LGTM, thanks again @sepaper!

sepaper:
@skriss Thanks again for the review!

Issue #4554, opened 2022-06-03
Gateway API: change wildcard hostnames to be suffix-matching, not single segment-matching


skriss:
Gateway API is changing the meaning of hostname wildcards to be prefix matching instead of single segment matching (ref. https://github.com/kubernetes-sigs/gateway-api/pull/1173#issuecomment-1145508508). Contour has already implemented them as single segment matches, so we'll need to update. This means we don't need to use the `:authority` header match trick that we use for `Ingress` anymore for Gateway API. We'll also need to revise the Gateway<->Route hostname intersection logic.

PR #4558, opened 2022-06-06
Gateway API: fix bug with routes having multiple listener refs

skriss:
Fixes a bug where a route with multiple parent
refs to listeners would not attach to all
listeners correctly.


sunjayBhatia (approved):
makes sense! shouldnt return early and should check all parent refs

skriss:
Gateway conformance is expected to fail, found this bug while working on https://github.com/projectcontour/contour/issues/4554 to get back to green there

PR #4559, opened 2022-06-06
Gateway API: allow wildcard hostnames to match more than one DNS label

### LINKED ISSUE #4554: Gateway API: change wildcard hostnames to be suffix-matching, not single segment-matching

skriss:
Changes Gateway API hostname logic to allow
wildcard hostnames to match more than one
DNS label per clarification in the upstream
spec.

Closes #4554.

### DIFF
--- a/internal/dag/gatewayapi_processor_test.go
+++ b/internal/dag/gatewayapi_processor_test.go
@@ -207,7 +207,7 @@ func TestComputeHosts(t *testing.T) {
 			want:      sets.NewString("very.many.labels.projectcontour.io"),
 			wantError: nil,
 		},
-		"listener wildcard host doesn't match hostnames with many labels host": {
+		"listener wildcard host matches hostnames with many labels host": {
 			listenerHost: "*.projectcontour.io",
 			hostnames: []gatewayapi_v1alpha2.Hostname{
 				"very.many.labels.projectcontour.io",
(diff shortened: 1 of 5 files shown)


internal/dag/gatewayapi_processor_test.go
```diff
@@ -199,20 +199,20 @@ func TestComputeHosts(t *testing.T) {
...
+		"listener host with many labels matches hostnames wildcard host": {
+			listenerHost: "very.many.labels.projectcontour.io",
 			hostnames: []gatewayapi_v1alpha2.Hostname{
 				"*.projectcontour.io",
 			},
-			want:      nil,
+			want:      sets.NewString("very.many.labels.projectcontour.io"),
 			wantError: nil,
 		},
 		"listener wildcard host doesn't match hostnames with many labels host": {
```
sunjayBhatia, line 210: should this test case description be updated? "listener wildcard host matches hostnames with many labels host"

skriss: fixed!

sunjayBhatia (approved):

PR #4567, opened 2022-06-09
Bump Envoy to 1.21.1

sunjayBhatia:
(Docker images haven't been pushed yet, but getting this out early)


sunjayBhatia:
(ignore the branch name that was a mistake)

tsaarni:
It seems that currently the link https://www.envoyproxy.io/docs/envoy/v1.22.1/version_history/current is broken, but it is indeed correct (from Envoy release notes). There was an issue at Envoy github about this as well with fix coming later.

sunjayBhatia:
> It seems that currently the link https://www.envoyproxy.io/docs/envoy/v1.22.1/version_history/current is broken, but it is indeed correct (from Envoy release notes). There was an issue at Envoy github about this as well with fix coming later.

Ah yeah saw that too, hopefully will be fixed before we release

Issue #4555, opened 2022-06-03
Gateway API: ReferencePolicy -> ReferenceGrant rename


skriss:
Gateway API is renaming the `ReferencePolicy` resource to `ReferenceGrant`. We'll need to update Contour for this.

ref. https://github.com/kubernetes-sigs/gateway-api/pull/1179

youngnick:
As of today, ReferencePolicy will be still included in v0.5.0, so it's not super urgent, but we should move before v0.6.0, when it will be removed (and most likely, moved to beta tbh).

skriss:
Still to do: drop ReferencePolicy when we upgrade to v0.6.0.

skriss:
Moving to 1.24 since it's unlikely we'll get Gateway API v0.6 into 1.23.

skriss:
- [ ] drop ReferencePolicy support (#4830)
- [ ] update ReferenceGrant to v1beta1

PR #4580, opened 2022-06-15
Gateway API: add support for ReferenceGrant

skriss:
Adds support for the renamed ReferenceGrant
resource while retaining support for
ReferencePolicy, to allow for a seamless
migration. ReferencePolicy support will
be removed in a subsequent release.

Updates https://github.com/projectcontour/contour/issues/4555.

### DIFF
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -1424,7 +1424,7 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 				},
 			),
 		},
-		"HTTPRoute references a backend in a different namespace, no ReferencePolicy": {
+		"HTTPRoute references a backend in a different namespace, no ReferenceGrant": {
 			gatewayclass: validClass,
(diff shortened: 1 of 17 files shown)


internal/dag/builder_test.go
```diff
@@ -1424,7 +1424,7 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 				},
 			),
 		},
-		"HTTPRoute references a backend in a different namespace, no ReferencePolicy": {
+		"HTTPRoute references a backend in a different namespace, no ReferenceGrant": {
```
skriss, line 1427: I decided to just switch these tests over rather than duplicate them, since the old ones will go away soon, but I can do the former if folks think it's necessary.

sunjayBhatia (approved):
LGTM

just thinking out loud, since referencepolicy/grant is an "allow list" rather than a block list, there can't be really any conflicts between different instances of either, so seems pretty simple to have both working etc. and remove referencepolicy next next (1.23.0) version

PR #4584, opened 2022-06-15
test/e2e/incluster: Contour/Envoy listening on ipv6 in an ipv4 cluster

sunjayBhatia:
This works since we set up Envoy to listen on ipv6 with ipv4
compatibility and we set up Contour to listen via golang net.Listen with
the "tcp" network, which when using an ipv6 address, golang
automatically sets ipv6/ipv4 compatible socket options.

This configuration can be an option we advertise to users who want a configuration
they can use across ipv4, dual-stack, and/or ipv6 clusters.

### DIFF
--- a/test/e2e/incluster/incluster_test.go
+++ b/test/e2e/incluster/incluster_test.go
@@ -18,6 +18,7 @@ package incluster
 
 import (
 	"context"
+	"os"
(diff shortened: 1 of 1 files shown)


skriss (approved):
LGTM!

PR #4586, opened 2022-06-18
internal/provisioner: Use local variable  to replace the long access chain of fields

izturn:
it seems the fields of object nest too deeply, so use  a local variable to replace it to make read code easy

### DIFF
--- a/test/e2e/deployment.go
+++ b/test/e2e/deployment.go
@@ -339,9 +339,6 @@ func (d *Deployment) WaitForContourDeploymentUpdated() error {
 		if err := d.client.List(context.TODO(), pods, labelSelectAppContour); err != nil {
 			return false, err
 		}
-		if pods == nil {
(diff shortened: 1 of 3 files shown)


youngnick (commented):
The change itself looks fine to me, could you please run `gofmt -s` and check that `make lint` succeeds?

izturn:
thx @youngnick  done, but how can i set the release note label?

test/e2e/deployment.go
```diff
@@ -339,9 +339,6 @@ func (d *Deployment) WaitForContourDeploymentUpdated() error {
 		if err := d.client.List(context.TODO(), pods, labelSelectAppContour); err != nil {
 			return false, err
 		}
-		if pods == nil {
```
tsaarni, line 342: I wonder what was the reason for removing this?

youngnick (approved):
LGTM, thanks @izturn!

Issue #4278, opened 2022-01-19
malformed authority header for ExtensionService


cortopy:
**What steps did you take and what happened:**

I wrote an `AuthorizationServer` using Rust. In Rust, the main package to handle http2 connections is called `h2`, which happens to be a very strict implementation of the http2 protocol.

I then registered my server as an `ExtensionService`

When Envoy makes a request, the authority header is set to:

```
"extension/{{ namespace }}/{{ ExtensionService name }}"
```

That is an invalid authority header according to the specs. As a result, `h2` refuses to handle the request as reported [here a while ago](https://github.com/hyperium/h2/issues/525)

**What did you expect to happen:**

A valid authority header is sent to the server. Maybe only the ExtensionService name, or the virtual host fqdn?

**Anything else you would like to add:**

I've tried with h2 and h2c protocols without any diffrence. 

I also tried not providing a protocol to see if maybe a http1 request could be handled (Rust has utilities already to convert http1/grpc-web on the fly and it could work) but, in this case, no protocol is set and the server doesn't know what to do with the TCP request

**Environment:**

- Contour version: 1.19.1 
- Kubernetes version: (use `kubectl version`): 1.23
- Kubernetes installer & version: k0s
- Cloud provider or hardware configuration: bare metal
- OS (e.g. from `/etc/os-release`): Ubuntu

sunjayBhatia:
It is likely Envoy is using the cluster name Contour creates (with slashes etc.) as the `:authority` header, we can probably look at changing that around

sunjayBhatia:
I would think this issue may apply to any `h2` service, surprising it has only come up now

cortopy:
@sunjayBhatia thanks for looking into this. 

I'm also using Contour to expose grpc-web services. I'm using `h2c` protocol in the `HTTPProxy` and this works fine even with http2. In what I've tested, this only happens with the slashes in the authority sent by the `ExtensionService`

travisghansen:
If anyone needs a quick/dirty solution to this I wrote this for similar issues (typically involving uds): https://github.com/democratic-csi/csi-grpc-proxy which can be used to simply proxy the request and rewrites (unconditionally essentially) the `:authority` header to `localhost`.

travisghansen:
Is there an envoy bug we're aware of that has been created to address this issue?

youngnick:
@sunjayBhatia would be the one to know here.

sunjayBhatia:
[...]

PR #4587, opened 2022-06-21
All gRPC Services have a sanitized authority

sunjayBhatia:
- so that cluster name, which may not be a valid host, is not used
- if SNI is set on ExtensionService, it is passed through as authority
- otherwise, / is replaced with . in extension cluster name to be used
    as authority

Fixes: #4278


youngnick (approved):
This LGTM, thanks for sorting this out.

PR #4588, opened 2022-06-22
Gateway API: implement HTTP query param matching

### LINKED ISSUE #4089: Gateway API: implement HTTP query param matching [extended]
ref. https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.HTTPQueryParamMatch

extended support in v1alpha2.

skriss:
Closes #4089.

### DIFF
--- a/changelogs/unreleased/4588-skriss-minor.md
+++ b/changelogs/unreleased/4588-skriss-minor.md
@@ -1,0 +1,30 @@
+## Gateway API: implement HTTP query parameter matching
+
+Contour now implements Gateway API's [HTTP query parameter matching](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.HTTPQueryParamMatch).
+Only `Exact` matching is supported.
+For example, the following HTTPRoute will send a request with a query string of `?animal=whale` to `s1`, and a request with a querystring of `?animal=dolphin` to `s2`.
+
+```yaml
+apiVersion: gateway.networking.k8s.io/v1alpha2
+kind: HTTPRoute
+metadata:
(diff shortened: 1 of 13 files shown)


sunjayBhatia (commented):
looks good overall, I think maybe only thing to add would be to check for duplicate requested matches for the same key (and set status accordingly), and maybe offer an e2e test/document how the match will work with Envoy if a URL is presented that has the same query key multiple times (e.g. `/foo?bar=1&bar=2`), since as far as I can tell its server dependent and not in an RFC

skriss:
> looks good overall, I think maybe only thing to add would be to check for duplicate requested matches for the same key (and set status accordingly), 

Good point, I don't see anything in Gateway API right now that specifies behavior, but for HTTP Header matches which are structured similarly, it says:
> If multiple entries specify equivalent header names, only the first entry with an equivalent name MUST be considered for a match. Subsequent entries with an equivalent header name MUST be ignored. Due to the case-insensitivity of header names, “foo” and “Foo” are considered equivalent.

So I think it probably makes sense to apply similar logic here, and I'll raise something upstream to clarify in the spec.

> and maybe offer an e2e test/document how the match will work (match on first occurrence of key, last, etc.) with Envoy if a URL is presented that has the same query key multiple times (e.g. `/foo?bar=1&bar=2`), since as far as I can tell its server dependent and not in an RFC

Good idea, let me test this out.

skriss:
@sunjayBhatia see https://github.com/projectcontour/contour/pull/4588/commits/0fd420f5699962fc34fea7f8e96e893065247c68, let me know if this makes sense to you!

sunjayBhatia (approved):
looks good!

PR #4592, opened 2022-06-23
Adds support for client certificate revocation.

### LINKED ISSUE #4590: Implement revocation support for client certificate validation
Implement optional CRL checking that can be used to deny downstream clients that present revoked certificates when client validation is enabled.

This is issue for implementing [design/downstream-crl-design.md](https://github.com/projectcontour/contour/blob/main/design/downstream-crl-design.md) which was kindly contributed by @artificial-aidan #4252.

tsaarni:
This change adds optional support to validate client certificates against CRL.

Implements [downstream-crl-design.md](https://github.com/projectcontour/contour/blob/main/design/downstream-crl-design.md) with the addition of `only_verify_leaf_cert_crl` which is now available in Envoy.

Fixes #4590


internal/envoy/v3/auth.go
```diff
@@ -91,6 +91,15 @@ func validationContext(ca []byte, subjectName string, skipVerifyPeerCert bool) *
 		}
 	}
 
+	if len(crl) > 0 {
+		vc.ValidationContext.Crl = &envoy_api_v3_core.DataSource{
+			Specifier: &envoy_api_v3_core.DataSource_InlineBytes{
+				InlineBytes: crl,
```
sunjayBhatia, line 97: nit: could you add a TODO comment to move this to SDS (similar to https://github.com/projectcontour/contour/blob/f44a2125a71c240fca1fa03aee2fa57694c5a28e/internal/envoy/v3/auth.go#L74) which we should do eventually

tsaarni: I moved the comment to function level, since it seems the whole `ValidationContext` seems to be delivered via SDS ([doc](https://www.envoyproxy.io/docs/envoy/latest/configuration/security/secret#sds-configuration)) and not just the DataSource(s): CA and CRL.

tsaarni (commented):
Spotted two typos. Will be fixing those and improving test coverage a small bit.

skriss (approved):
LGTM!

PR #4593, opened 2022-06-23
Gateway API: only keep first match for a given header name

### LINKED ISSUE #4591: Gateway API: ignore all but the first header match for a given header name
See https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.HTTPHeaderMatch

Currently we're not doing this: https://github.com/projectcontour/contour/blob/main/internal/dag/gatewayapi_processor.go#L1230-L1249

skriss:
Adds logic to only keep the first HTTP header match
with a given name (case-insensitive) for each HTTP
route rule, per the Gateway API spec.

Closes #4591.

### DIFF
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -1237,20 +1237,28 @@ func gatewayPathMatchCondition(match *gatewayapi_v1alpha2.HTTPPathMatch, routeAc
 
 func gatewayHeaderMatchConditions(matches []gatewayapi_v1alpha2.HTTPHeaderMatch) ([]HeaderMatchCondition, error) {
 	var headerMatchConditions []HeaderMatchCondition
+	seenNames := sets.String{}
 
 	for _, match := range matches {
-		// HeaderMatchTypeExact is the default if not defined in the object.
-		headerMatchType := HeaderMatchTypeExact
-		if match.Type != nil {
-			switch *match.Type {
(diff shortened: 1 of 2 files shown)


sunjayBhatia (approved):

internal/dag/gatewayapi_processor.go
```diff
@@ -1237,20 +1237,28 @@ func gatewayPathMatchCondition(match *gatewayapi_v1alpha2.HTTPPathMatch, routeAc
...
+		// the only supported match type.
+		if match.Type != nil && *match.Type != gatewayapi_v1alpha2.HeaderMatchExact {
+			return nil, fmt.Errorf("HTTPRoute.Spec.Rules.Matches.Headers: Only Exact match type is supported")
+		}
+
+		// If multiple match conditions are found for the same header name (case-insensitive),
+		// use the first one and ignore subsequent ones.
+		upperName := strings.ToUpper(string(match.Name))
+		if seenNames.Has(upperName) {
+			continue
```
sunjayBhatia, line 1253: a shame maybe that the spec doesn't have a status warning/condition/reason etc. to set for this sort of thing, but could be added if people start getting confused about how things work

skriss: Yeah, that could be added -- I raised an issue upstream about the query param thing and there's been some discussion there on adding validation for these things to the webhook, so adding a Condition could be another possible outcome there, I'll keep an eye on it.

Issue #4556, opened 2022-06-06
Unable to use contour for typical wildcard TLS setup: `IngressSpec.tls.secretName` seems to be required, proxy protocol support is not working


artiommocrenco:
**What steps did you take and what happened:**
Not willing to use contour CRDs.

Using the same `networking.k8s.io/v1` Ingress resources that worked with nginx-ingress. Expected them to work with contour.

I am testing out bitnami/contour chart with the following helm values:
```yaml
contour:
  extraArgs:
    - --use-proxy-protocol
configInline:
  network:
    num-trusted-hops: 1
  tls:
    fallback-certificate:
      name: wildcard-tls
      namespace: contour
```

where `wildcard-tls` secret is created by cert-manager for `*.example.com`

I have Ingress with similar IngressSpec:

```
rules:
- host: app.example.com
  http:
    paths:
    - backend:
        service:
          name: app-ui
          port:
            name: http
      path: /
      pathType: Prefix
tls:
- hosts:
  - app.example.com
```

This produces the following:

1. `unresolved secret reference` contour error

```
time="2022-06-06T12:06:34Z" level=error msg="unresolved secret reference" context=IngressProcessor error="Secret not found" name=app-ui namespace=app secret=app/
```

2. `envoy.http.original_ip_detection.xff` envoy error

```
[2022-06-06 11:51:46.111][1][warning][config] [source/common/config/grpc_subscription_impl.cc:126] gRPC config for type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error adding/updating listener(s) ingress_http: Didn't find a registered implementation for name: 'envoy.http.original_ip_detection.xff'
```

3. Envoy HTTPS endpoints are not starting at all, only HTTP endpoints do (and only if I remove `configInline.network.num-trusted-hops` setting)

**What did you expect to happen:**

1. Not specifying `IngressSpec.tls.secretName` should not produce a log message of type `error`. `IngressSpec.tls.secretName` is optional as per [Kubernetes API Reference](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/ingress-v1/#IngressSpec). 
2. There should not be any `envoy.http.original_ip_detection.xff` envoy-related errors
3. Envoy HTTPS endpoints must be functional and use the configured wildcard certificate as per `fallback-certificate`
4. App must then be available both via `http://app.example.com/` and `https://app.example.com/`

**Environment:**

- Contour version: 
```
contour-7.10.1              	1.20.1     
```
[...]

Issue #2453, opened 2020-04-21
support fallback certificate in v1beta1/Ingress


stevesloka:
#2428 has a design to enable fallback certs in `HTTPProxy`. This issue is to follow up with thoughts around how to apply to Ingress v1beta1.

_Originally posted by @stevesloka in https://github.com/projectcontour/contour/pull/2428_

ravilr:
I guess the logic to support fallback certificate feature in Ingress resource wouldn't be that different from HTTPProxy.
If ingressResource.Spec.TLS is not set and an fallback annotation is set, add the hosts and corresponding route rules from that ingressResource to ingress_fallback route config during dag visit.

echupriyanov:
Any updates on this feature?

sunjayBhatia:
Cc @xaleeks if this is something on your priority list

If we were to support this on Ingress, would have to be via an annotation to enable it, however which vhost to apply a fallback cert to is problematic as an Ingress could have many

youngnick:
Yes, I agree that it would need an annotation, and would also need clear documentation about what would happen in various Ingress cases (like multiple vhosts).

sunjayBhatia:
same boat as https://github.com/projectcontour/contour/issues/4556 moving out of a milestone until we get more info

PR #4595, opened 2022-06-24
Add fallback cert support to Ingress v1

sunjayBhatia:
- when annotation projectcontour.io/enable-fallback-certificate == true,
fallback configured globally will be added
- since TLS secret is optional, if not provided then fallback used as
main secret and fallback
- if fallback configuration invalid, try to keep main TLS configuration
still configured
- errors surfaced via logs since we don't have status

Fixes: #2453
Fixes: #4556


sunjayBhatia:
WIP, wanted to get it out for review early, still need to test things out more and write an e2e test

main odd thing to review is to look at what happens when no TLS secret is provided in Ingress TLS section (since it is technically optional) and fallback is enabled, right now I have it so that the fallback cert is added as the main vhost TLS secret *and* the fallback (for requests w/o SNI)

internal/dag/ingress_processor.go
```diff
@@ -77,34 +82,102 @@ func (p *IngressProcessor) Run(dag *DAG, source *KubernetesCache) {
...
+				fallbackCertEnabled = false
+			}
+		}
+
+		for _, tls := range ing.Spec.TLS {
+			var secret *Secret
+			if tls.SecretName == "" {
+				if fallbackCertEnabled {
+					// If no provided secret, use fallback cert as main vhost secret.
+					secret = fallbackCert
```
skriss, line 142: Is this the same behavior as for HTTPProxy? (started looking but if you've already tracked this down I'll spare myself the spelunk).

skriss:
> main odd thing to review is to look at what happens when no TLS secret is provided in Ingress TLS section (since it is technically optional) and fallback is enabled, right now I have it so that the fallback cert is added as the main vhost TLS secret and the fallback (for requests w/o SNI)

I haven't spent a lot of time with this code, so I might be missing something, but it seems to me like you could either (a) do what you've done here; or (b) not add a filterchain for the specific vhost, and you'd effectively end up with the same user experience (since in b, requests for that vhost would match the fallback filterchain which would have the same cert and still have all the routes). Does that sound right to you? If so, then I think what you've got here makes sense and is probably a bit more efficient than (b).

internal/dag/ingress_processor.go
sunjayBhatia: HTTPProxy requires you to specify a TLS cert so you always get a "main" TLS cert for most requests and the fallback for ones w/o SNI

with Ingress you don't have to provide a cert so this solution is kind of stretching of the meaning of "fallback" to extend to that use case

sunjayBhatia:
> > main odd thing to review is to look at what happens when no TLS secret is provided in Ingress TLS section (since it is technically optional) and fallback is enabled, right now I have it so that the fallback cert is added as the main vhost TLS secret and the fallback (for requests w/o SNI)
> 
> I haven't spent a lot of time with this code, so I might be missing something, but it seems to me like you could either (a) do what you've done here; or (b) not add a filterchain for the specific vhost, and you'd effectively end up with the same user experience (since in b, requests for that vhost would match the fallback filterchain which would have the same cert and still have all the routes). Does that sound right to you? If so, then I think what you've got here makes sense and is probably a bit more efficient than (b).

yeah, you would get the same ux with (a) and (b) for a user making requests, with just a minor hit that you end up going through the catchall "any TLS" filterchain+fallback route

originally I sortof tried to do (b) but we would need a few small/fiddly logic changes to the dag and xdscache to make it ok to have a secure vhost that does not do passthrough TLS or have a secret

tsaarni:
In the [docs]( https://projectcontour.io/docs/main/config/tls-termination/#fallback-certificate) the fallback certificate is currently explained through the use case where client does not set SNI. Using the fallback cert also when SNI is set would be small exception to that, a bit of a stretch like said previously, but of course it is just one definition.

Technically user could configure `secretName` with a reference to the fallback certificate secret (assuming they know the name of it), instead of configuring `projectcontour.io/enable-fallback-certificate: true` annotation, like they need to do for `HTTPProxy` (as stated previously). But if I understood if correctly, the motivation with the automatic behavior is to avoid an error case since the field is optional?  One can still have similar errors if (1) leaving `secretName` field empty and not setting `projectcontour.io/enable-fallback-certificate: true` or if (2) fallback certificate secret name has not been configured for the deployment or delegation rule is not there.

The description for leaving field `Ingress.spec.tls.secretName` empty says _"Field is left optional to allow TLS routing based on SNI hostname alone"_ so I wonder if that option is really meant for TLS passthrough?

That said, I do not have any objections for the suggested implementation! :)  It seems to match what [NGINX](https://kubernetes.github.io/ingress-nginx/user-guide/tls/#default-ssl-certificate) does, if I understood the description correctly (did not test).

buroa:
Ping @sunjayBhatia, can we get this merged?

Issue #4560, opened 2022-06-08
update to Gateway API v0.5.0/v1beta1


skriss:
Once Gateway API v0.5.0 is released, Contour needs to be updated to use it, as well as the v1beta1 API.

ref. #4555

skriss:
I'm discovering that updating to `v1beta1` types is problematic, specifically for HTTPRoute, because TLSRoute is not moving to beta, but we have a lot of code that's shared across those two types for hostnames, status, etc. Since v1alpha2 and v1beta1 are effectively identical for HTTPRoute (and both will be present), it may make sense to only update GatewayClass and Gateway to v1beta1 for now, and wait until TLSRoute graduates to update the route code.

youngnick:
That makes sense for now, but we may have to make that shared logic a bit smarter - I suspect that it may be a while before we move TLSRoute to v1beta1.

skriss:
Hey @carlisia!

> > Since v1alpha2 and v1beta1 are effectively identical for HTTPRoute (and both will be present), it may make sense to only update GatewayClass and Gateway to v1beta1 for now, and wait until TLSRoute graduates to update the route code.
> 
> Hey @skriss (👋!!) just wanted to be sure: Although you have a bump to Gateway API release v0.5.0-rc1 on your latest Contour release (Contour v1.21.1), you are still only using the `v1alpha2` api all across, is this correct? 

We've updated Contour's `main` branch to use Gateway API v0.5.0-rc1, but that hasn't made it into any release and we'll update to v0.5.0 before our next release. Correct, we are still using the `v1alpha2` API version, even in `main`.

> And you will be updating maybe only GatewayClass and Gateway to v1beta1 in a future release? The next one possibly? Am > I reading this correctly?

I think we'll _probably_ update Contour code to use `v1beta1` for GatewayClass, Gateway, and HTTPRoute before Contour's next release (although since v1alpha2 and v1beta1 are automatically converted between by the API server, users can still use either API version), though I'm still sorting out exactly how much of a pain it will be to update HTTPRoute code :)

> 
> Addendum: and you are not using any of the `standard` Gateway API CRDs in the currently release, right?

Contour 1.21.1 uses Gateway API v0.4.3 which didn't have release channels yet; on `main` (using v0.5.0-rc1) we've decided to use the experimental channel since we've already implemented ReferencePolicy, and that's only in the experimental channel right now.

[...]

PR #4596, opened 2022-06-27
Gateway API: use v1beta1 API for HTTPRoute

skriss:
Updates HTTPRoute code to use the v1beta1
API since this resource has moved to beta
in the v0.5.0 Gateway API release.

Updates #4560.


sunjayBhatia (approved):
lgtm, package/import replacement and the upgrade/downgrade logic to cohabitate with the v1alpha2 TLSRoute makes sense to me!

PR #4597, opened 2022-06-28
Added support for overload manager

### LINKED ISSUE #1794: Investigate enabling overload manager for Envoy
As mentioned in #1375, Envoy has an overload manager that we should investigate.

It's pretty generic by design, but currently is effectively a memory (that is, heap size) manager. You can configure a maximum heap size, and then different actions to take at different percentages of that heap size.

Sample from the Envoy PR:
```yaml
overload_manager:
  refresh_interval: 0.25s
  resource_monitors:
  - name: "envoy.resource_monitors.fixed_heap"
    config:
      # TODO: Tune for your system.
      max_heap_size_bytes: 2147483648 # 2 GiB
  actions:
  - name: "envoy.overload_actions.shrink_heap"
    triggers:
    - name: "envoy.resource_monitors.fixed_heap"
      threshold:
        value: 0.95
  - name: "envoy.overload_actions.stop_accepting_requests"
    triggers:
    - name: "envoy.resource_monitors.fixed_heap"
      threshold:
        value: 0.98
```

tsaarni:
This change adds minimal support for Envoy's [overload manager](https://www.envoyproxy.io/docs/envoy/latest/configuration/operations/overload_manager/overload_manager) to avoid cases where Envoy process is terminated by the out-of-memory killer, which results in traffic distrubances.

This PR proposes that administrator can (optionally) configure maximum amount of heap that Envoy is allowed to reserve. It does not allow the overload [actions](https://www.envoyproxy.io/docs/envoy/latest/configuration/operations/overload_manager/overload_manager#overload-actions)  to be added or configured in any way. Instead, it configures default actions which are set according to example in ["Configuring Envoy as an edge proxy"](https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/edge#configuring-envoy-as-an-edge-proxy) best practices doc: `shrink_heap` action is executed when 95% and `stop_accepting_requests` action when 98% of configured maximum heap is reached.

The configuration of maximum heap is (unfortunately) again a new command line flag. This is the same for all other bootstrap paramters so far as well. The reasoning is that `contour bootsrap` is executed inside Envoy init container, where we do not have Contour config file or capability to read `ContourConfiguration` CR from the API server. Or at least we have not done this so far. 

There is a major conflict between overload manager and how we expose `/ready` and `/stats` by setting up a proxy to serve these endpoints! 

While the real admin API at `/admin/admin.sock` still works during overload, the requests via the "proxied versions" of `/ready` and `/stats` served via TCP socket will be rejected when `stop_accepting_requests` is active.  As a result, Envoy will be removed from the endpoints of the service. Since the Envoy instance was not accepting new requests anyways, maybe this will not make the overload any worse for the other Envoys. However, another side-effect is that administrator cannot monitor the Envoy instance anymore since stats endpoint will not be served either. Especially the memory related metrics would be of interest, since metrics will show that the `stop_accepting_requests` action is active and the heap numbers explaining why. When admin sets max heap too low, they will not be able to find that out by checking metrics - since metrics are not served due to heap being low :thinking: 

The feature itself seems very useful as it can avoid the OOM killer but I'd like to hear your opinion about the limitations.

Fixes #1794

As a workaround, I found out following commands helpful to access the "real" admin API when the "proxied" admin API endpoints are rejecting requests due to overload:

```console
sudo curl --silent --unix-socket /proc/$(pidof envoy)/root/admin/admin.sock http://localhost/stats | grep -E "^overload|^server.memory"
sudo curl --silent --unix-socket /proc/$(pidof envoy)/root/admin/admin.sock http://localhost/memory # tcmalloc metrics
```

These need to be executed on the worker node, or just on the dev host when running Kind.

Envoy's `fixed_heap` monitor uses the tcmalloc metrics at `/memory` and following formula to calculate the overload percentage: `(heap_size - pageheap_unmapped) / maximum_heap`


youngnick:
I think in cases that the heap is low, getting metrics is definitely less of a big deal than being oomkilled. It seems like this is about as good a compromise as we're going to be able to do, sadly.

It's unfortunate that we have to make the heap size a bootstrap cmdline param, but I don't see any other way to do it.

I also think that this feature has to come with a bunch of warnings about being careful with your sizing, making sure that it matches up with any Pod requests and limits you've put on Envoy, and so on.

I'll give the PR a more detailed review soon, sorry about the delay @tsaarni.

youngnick (commented):
I just had a better look at this PR, it looks pretty reasonable, but seems like it's missing the documentation page? Did you intend to include that here @tsaarni?

tsaarni:
I will come back with some documentation shortly.

tsaarni:
Missing documentation `site/content/docs/main/config/overload-manager.md` added.

tsaarni:
This PR is ready for review.

skriss:
Sorry for the delay on this @tsaarni, planning to take a look soon!

sunjayBhatia (approved):
LGTM, just one comment on the documentation page!

We can do a follow up issue to add this new bootstrap flag to the dynamic gateway provisioning method of deploying Contour: https://github.com/projectcontour/contour/blob/main/apis/projectcontour/v1alpha1/contourdeployment.go

tsaarni:
Thanks @sunjayBhatia for the review!

> this and the below might need an update to match the bootstrap config (looks like 90% and 98% for these two actions)

Thanks for spotting this! I went the other direction and changed the bootstrap config to match the documentation, since values came from [Envoy best practices document](https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/edge#configuring-envoy-as-an-edge-proxy). I think I had no real reason to use 90% instead of 95%.

tsaarni:
If there are no further questions, I'll merge this tomorrow.

changelogs/unreleased/4597-tsaarni-minor.md
```diff
@@ -0,0 +1,4 @@
+## Overload Manager
```
skriss, line 1: I'd bump this changelog up to "major", it will definitely be one of the big features in the release :)

internal/envoy/v3/bootstrap.go
```diff
@@ -251,6 +253,51 @@ func bootstrapConfig(c *envoy.BootstrapConfig) *envoy_bootstrap_v3.Bootstrap {
...
+			},
+			Actions: []*envoy_config_overload_v3.OverloadAction{
+				{
+					Name: "envoy.overload_actions.shrink_heap",
+					Triggers: []*envoy_config_overload_v3.Trigger{
+						{
+							Name: "envoy.resource_monitors.fixed_heap",
+							TriggerOneof: &envoy_config_overload_v3.Trigger_Threshold{
+								Threshold: &envoy_config_overload_v3.ThresholdTrigger{
+									Value: 0.95,
```
skriss, line 278: I could see folks wanting to tune these thresholds via flag, but I'm fine leaving statically defined for now since we can always add flags later if/when needed.

site/content/docs/main/config/overload-manager.md
```diff
@@ -0,0 +1,30 @@
...
+Overload manager is disabled by default.
+It can be enabled at deployment time by using `--overload-max-heap=[MAX_BYTES]` command line flag in [`contour bootstrap`][2] command.
+The bootstrap command is executed in [init container of Envoy pod][3] to generate initial configuration for Envoy.
+To enable overload manager, modify the deployment manifest and add for example `--overload-max-heap=2147483648` to set maximum heap size to 2 GiB.
+The appropriate number of bytes can be different from system to system.
+
+After the feature is enabled, following two overload actions are configured to Envoy:
+
+* Shrink heap action is executed when 95% of the maximum heap size is reached.
+* Envoy will stop accepting requests when 99% of the maximum heap size is reached.
```
skriss, line 17: off-by-one 😄  
```suggestion
* Envoy will stop accepting requests when 98% of the maximum heap size is reached.
```

skriss (changes requested):
Just a couple tiny things but this looks good to me. The issue with the admin endpoint is unfortunate, and maybe we can do some more thinking there on if there's something else we can do, but I don't think it needs to block getting the initial PR in, seems like a net improvement.

site/content/docs/main/config/overload-manager.md
tsaarni: Thanks, apparently I'm not any good with percentages :blush: Corrected !

internal/envoy/v3/bootstrap.go
tsaarni: True, these could be exposed in future, though I guess it means few more command line switches again...

tsaarni:
@skriss Thank you for the review! 

> Just a couple tiny things but this looks good to me. The issue with the admin endpoint is unfortunate, and maybe we can do some more thinking there on if there's something else we can do, but I don't think it needs to block getting the initial PR in, seems like a net improvement.

I agree. I could not figure anything that could be done on Contour side, except removing the proxy from the admin API, but that is there for a reason. Overload manager cannot be applied directly on named listeners only, or the other way around: it cannot be configured to ignore certain listeners...

skriss (approved):

PR #4602, opened 2022-06-29
Update contour cli commands to be a more correct xDS implementation

youngnick:
This PR updates `contour cli` with better logging and a correct implementation of the full ACK and NACK flow for both SOTW and Delta xDS types.

The `watchstream` and `watchDeltaStream` functions in `contour/cmd/cli.go` are extremely similar, but the types required (`DiscoveryRequest` and `DiscoveryResponse` for SOTW and `DeltaDiscoveryRequest` and `DeltaDiscoveryResponse` for Delta) are juuuust different enough that I felt that the brevity loss is made up for in readability.

### DIFF
--- a/cmd/contour/cli.go
+++ b/cmd/contour/cli.go
@@ -17,18 +17,22 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"fmt"
 	"io/ioutil"
-	"os"
 
+	corev3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	envoy_service_cluster_v3 "github.com/envoyproxy/go-control-plane/envoy/service/cluster/v3"
 	envoy_discovery_v3 "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
 	envoy_service_endpoint_v3 "github.com/envoyproxy/go-control-plane/envoy/service/endpoint/v3"
 	envoy_service_listener_v3 "github.com/envoyproxy/go-control-plane/envoy/service/listener/v3"
 	envoy_service_route_v3 "github.com/envoyproxy/go-control-plane/envoy/service/route/v3"
-	"github.com/golang/protobuf/proto"
+	"github.com/sirupsen/logrus"
+	grpc_code "google.golang.org/genproto/googleapis/rpc/code"
+	"google.golang.org/genproto/googleapis/rpc/status"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/protobuf/encoding/protojson"
 	kingpin "gopkg.in/alecthomas/kingpin.v2"
 )
 
@@ -38,6 +42,10 @@ type Client struct {
 	CAFile      string
 	ClientCert  string
 	ClientKey   string
+	Nack        bool
+	Delta       bool
+	NodeId      string
+	Log         *logrus.Logger
 }
 
 func (c *Client) dial() *grpc.ClientConn {
@@ -52,16 +60,20 @@ func (c *Client) dial() *grpc.ClientConn {
 		}
 		// Load the client certificates from disk
 		certificate, err := tls.LoadX509KeyPair(c.ClientCert, c.ClientKey)
-		kingpin.FatalIfError(err, "failed to load certificates from disk")
+		if err != nil {
+			c.Log.WithError(err).Fatal("failed to load certificates from disk")
+		}
 		// Create a certificate pool from the certificate authority
 		certPool := x509.NewCertPool()
 		ca, err := ioutil.ReadFile(c.CAFile)
-		kingpin.FatalIfError(err, "failed to read CA cert")
+		if err != nil {
+			c.Log.WithError(err).Fatal("failed to read CA cert")
+		}
 
 		// Append the certificates from the CA
 		if ok := certPool.AppendCertsFromPEM(ca); !ok {
 			// TODO(nyoung) OMG yuck, thanks for this, crypto/tls. Suggestions on alternates welcomed.
-			kingpin.Fatalf("failed to append CA certs")
+			c.Log.Fatal("failed to append CA certs")
 		}
 
 		creds := credentials.NewTLS(&tls.Config{
@@ -80,7 +92,9 @@ func (c *Client) dial() *grpc.ClientConn {
 	}
 
 	conn, err := grpc.Dial(c.ContourAddr, options...)
-	kingpin.FatalIfError(err, "failed connecting Contour Server")
+	if err != nil {
+		c.Log.WithError(err).Fatal("failed connecting Contour Server")
+	}
 
 	return conn
 }
@@ -88,28 +102,36 @@ func (c *Client) dial() *grpc.ClientConn {
 // ClusterStream returns a stream of Clusters using the config in the Client.
 func (c *Client) ClusterStream() envoy_service_cluster_v3.ClusterDiscoveryService_StreamClustersClient {
 	stream, err := envoy_service_cluster_v3.NewClusterDiscoveryServiceClient(c.dial()).StreamClusters(context.Background())
-	kingpin.FatalIfError(err, "failed to fetch stream of Clusters")
+	if err != nil {
+		c.Log.WithError(err).Fatal("failed to fetch stream of Clusters")
+	}
 	return stream
 }
 
 // EndpointStream returns a stream of Endpoints using the config in the Client.
 func (c *Client) EndpointStream() envoy_service_endpoint_v3.EndpointDiscoveryService_StreamEndpointsClient {
 	stream, err := envoy_service_endpoint_v3.NewEndpointDiscoveryServiceClient(c.dial()).StreamEndpoints(context.Background())
-	kingpin.FatalIfError(err, "failed to fetch stream of Endpoints")
+	if err != nil {
+		c.Log.WithError(err).Fatal("failed to fetch stream of Endpoints")
+	}
 	return stream
 }
 
 // ListenerStream returns a stream of Listeners using the config in the Client.
 func (c *Client) ListenerStream() envoy_service_listener_v3.ListenerDiscoveryService_StreamListenersClient {
 	stream, err := envoy_service_listener_v3.NewListenerDiscoveryServiceClient(c.dial()).StreamListeners(context.Background())
-	kingpin.FatalIfError(err, "failed to fetch stream of Listeners")
+	if err != nil {
+		c.Log.WithError(err).Fatal("failed to fetch stream of Listeners")
+	}
 	return stream
 }
 
 // RouteStream returns a stream of Routes using the config in the Client.
 func (c *Client) RouteStream() envoy_service_route_v3.RouteDiscoveryService_StreamRoutesClient {
 	stream, err := envoy_service_route_v3.NewRouteDiscoveryServiceClient(c.dial()).StreamRoutes(context.Background())
-	kingpin.FatalIfError(err, "failed to fetch stream of Routes")
+	if err != nil {
+		c.Log.WithError(err).Fatal("failed to fetch stream of Routes")
+	}
 	return stream
 }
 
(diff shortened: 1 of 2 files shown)


youngnick:
This fell out of some work I'm doing on implementing incremental xDS properly in Contour.

cmd/contour/cli.go
```diff
@@ -118,21 +140,242 @@ type stream interface {
...
+	req := &envoy_discovery_v3.DiscoveryRequest{
+		TypeUrl:       typeURL,
+		ResourceNames: resources,
+		VersionInfo:   currentVersion,
+		Node: &corev3.Node{
+			Id: NodeID,
+		},
 	}
+	log.WithField("currentVersion", currentVersion).Info("Sending discover request")
+	fmt.Println(m.Format(req))
```
youngnick, line 162: I've used `fmt.Println` for printing out the marshaled Envoy messages to make them easier for a human to read. `contour cli` isn't really intended to be machine-consumable anyway, it's a troubleshooting tool for humans. 😄

cmd/contour/cli.go
```diff
@@ -118,21 +140,242 @@ type stream interface {
...
 		}
-		err := st.Send(req)
-		kingpin.FatalIfError(err, "failed to send Discover Request")
+
+		currentVersion = resp.VersionInfo
+
+		if nack {
+			// We'll NACK the response we just got.
+			// The ResponseNonce field is what makes it an ACK,
+			// and the VersionInfo field must match the one in the response we
```
sunjayBhatia, line 190: I think the version info in a NACK or ACK should match the last good response, so in this NACK case should still return an empty version: https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#nack

its a little funny in the NACK case for the CLI to send NACKs forever and starting with the first request, maybe we could add a "NACK sentinel" file so we can test turning NACKs on and off?

sunjayBhatia:
playing around with this change and the contour server in `envoy` xDS mode w/ delta, when creating and deleting an HTTPProxy it looks like after the initial snapshot is sent, I'm not seeing any updates in LDS or CDS so some details might need to be ironed out either with the client, the server, etc.

cmd/contour/cli.go
```diff
@@ -118,21 +140,242 @@ type stream interface {
...
+		if err != nil {
+			log.WithError(err).Fatal("failed to marshal incremental Discovery Response")
+		}
+
+		currentVersion = resp.SystemVersionInfo
+
+		if nack {
+			// We'll NACK the response we just got.
+			// The ResponseNonce field is what makes it an ACK,
+			// and the VersionInfo field must match the one in the response we
```
sunjayBhatia, line 331: similar thing here for the version info which is less important in the delta case and isn't part of the request, could probably remove some of the commentary about it here: https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#incremental-xds

sunjayBhatia: versions and response nonces will likely end up diverging since nonces are per-client req/resp and the snapshot/system version info is more based on config changes as well

cmd/contour/cli.go
youngnick: The NACK sentinel file idea is a good idea, we could also take a keyboard input maybe? Then we can toggle on the NACK, and toggle it off again.

I'll fix the version, but I think we should merge this in, then come back and add in some sort of runtime NACK switch.

cmd/contour/cli.go
youngnick: Yeah, fair enough, I'll remove the versioninfo stuff, but I think having some information about what makes it an ACK is useful.

cmd/contour/cli.go
youngnick: If we remove sending the version here, go-control-plane will respond to the NACK by immediately sending a DiscoveryResponse, which will mean that we'll spam NACKs and responses back and forth as fast as the wire can handle.

I think we need to think of the VersionInfo as "the last config *seen*", not "the last config *ACKed*", at least until we add better NACK handling to go-control-plane.

youngnick:
> playing around with this change and the contour server in envoy xDS mode w/ delta, when creating and deleting an HTTPProxy it looks like after the initial snapshot is sent, I'm not seeing any updates in LDS or CDS so some details might need to be ironed out either with the client, the server, etc.

> listeners wont change in this case, but we should be making sure clusters get removed so they don't stick around

Hmm, I've been checking with an Ingress, and definitely saw clusters being added and removed. I'll do a quick check with a HTTPProxy.

youngnick:
Hmm, it seems to work for me.
I ran `contour` with `contour serve --contour-config-name=blank --debug --insecure --xds-address=0.0.0.0 --xds-port=8001`, and the `blank` ContourConfig sets the `envoy` xDS server.

Then, I ran `contour cli cds --node-id "ContourCLI" --delta`, and while it was running, edited the HTTPProxy that was using the `echo-service` Service, changing it to `httpbin` as well, and got the following output:

```
INFO[0000] Sending incremental discover request          currentVersion=0
{
  "node": {
    "id": "ContourCLI"
  },
  "type_url": "type.googleapis.com/envoy.config.cluster.v3.Cluster"
}
INFO[0000] Received Discovery Response                   currentVersion=0 nonce=1 resp_system_version_info=5
{
  "system_version_info": "5",
  "resources": [
    {
      "name": "default/echo-service/80/da39a3ee5e",
      "version": "f483c35d3c4d1a0abb90dc115f77c297e3222c0d6d7844226a72f60a4603d9e4",
      "resource": {
        "@type": "type.googleapis.com/envoy.config.cluster.v3.Cluster",
        "name": "default/echo-service/80/da39a3ee5e",
        "alt_stat_name": "default_echo-service_80",
        "type": "EDS",
        "eds_cluster_config": {
          "eds_config": {
            "api_config_source": {
              "api_type": "GRPC",
              "transport_api_version": "V3",
              "grpc_services": [
                {
                  "envoy_grpc": {
                    "cluster_name": "contour",
                    "authority": "contour"
                  }
                }
              ]
            },
            "resource_api_version": "V3"
          },
          "service_name": "default/echo-service/http-port"
        },
        "connect_timeout": "2s",
        "common_lb_config": {
          "healthy_panic_threshold": {}
        }
      }
    },
    {
      "name": "default/httpbin/80/da39a3ee5e",
      "version": "b8215f4680562ad9b9554aa58df94d810330accae5af0301bac25246369f43e6",
      "resource": {
        "@type": "type.googleapis.com/envoy.config.cluster.v3.Cluster",
        "name": "default/httpbin/80/da39a3ee5e",
        "alt_stat_name": "default_httpbin_80",
        "type": "EDS",
        "eds_cluster_config": {
          "eds_config": {
            "api_config_source": {
              "api_type": "GRPC",
              "transport_api_version": "V3",
              "grpc_services": [
                {
                  "envoy_grpc": {
                    "cluster_name": "contour",
                    "authority": "contour"
                  }
                }
              ]
            },
            "resource_api_version": "V3"
          },
          "service_name": "default/httpbin"
        },
        "connect_timeout": "2s",
        "common_lb_config": {
          "healthy_panic_threshold": {}
        }
      }
    }
  ],
  "type_url": "type.googleapis.com/envoy.config.cluster.v3.Cluster",
  "nonce": "1"
}
INFO[0000] Sending incremental ACK discover request      currentVersion=5 response_nonce=1 version_info=5
{
  "node": {
    "id": "ContourCLI"
  },
  "type_url": "type.googleapis.com/envoy.config.cluster.v3.Cluster",
  "response_nonce": "1"
}
INFO[0015] Received Discovery Response                   currentVersion=5 nonce=2 resp_system_version_info=6
{
  "system_version_info": "6",
  "type_url": "type.googleapis.com/envoy.config.cluster.v3.Cluster",
  "removed_resources": [
    "default/echo-service/80/da39a3ee5e"
  ],
  "nonce": "2"
}
INFO[0015] Sending incremental ACK discover request      currentVersion=6 response_nonce=2 version_info=6
{
  "node": {
    "id": "ContourCLI"
  },
  "type_url": "type.googleapis.com/envoy.config.cluster.v3.Cluster",
  "response_nonce": "2"
}
```

I guess I'm missing something though?

sunjayBhatia:
> Hmm, it seems to work for me. I ran `contour` with `contour serve --contour-config-name=blank --debug --insecure --xds-address=0.0.0.0 --xds-port=8001`, and the `blank` ContourConfig sets the `envoy` xDS server.
> 
> Then, I ran `contour cli cds --node-id "ContourCLI" --delta`, and while it was running, edited the HTTPProxy that was using the `echo-service` Service, changing it to `httpbin` as well, and got the following output:
> 
> ```
> INFO[0000] Sending incremental discover request          currentVersion=0
> {
>   "node": {
>     "id": "ContourCLI"
>   },
>   "type_url": "type.googleapis.com/envoy.config.cluster.v3.Cluster"
> }
> INFO[0000] Received Discovery Response                   currentVersion=0 nonce=1 resp_system_version_info=5
> {
>   "system_version_info": "5",
>   "resources": [
>     {
>       "name": "default/echo-service/80/da39a3ee5e",
>       "version": "f483c35d3c4d1a0abb90dc115f77c297e3222c0d6d7844226a72f60a4603d9e4",
>       "resource": {
>         "@type": "type.googleapis.com/envoy.config.cluster.v3.Cluster",
>         "name": "default/echo-service/80/da39a3ee5e",
>         "alt_stat_name": "default_echo-service_80",
>         "type": "EDS",
>         "eds_cluster_config": {
>           "eds_config": {
>             "api_config_source": {
>               "api_type": "GRPC",
>               "transport_api_version": "V3",
>               "grpc_services": [
>                 {
>                   "envoy_grpc": {
>                     "cluster_name": "contour",
>                     "authority": "contour"
>                   }
>                 }
>               ]
>             },
>             "resource_api_version": "V3"
>           },
>           "service_name": "default/echo-service/http-port"
>         },
>         "connect_timeout": "2s",
>         "common_lb_config": {
>           "healthy_panic_threshold": {}
>         }
>       }
>     },
>     {
>       "name": "default/httpbin/80/da39a3ee5e",
>       "version": "b8215f4680562ad9b9554aa58df94d810330accae5af0301bac25246369f43e6",
>       "resource": {
>         "@type": "type.googleapis.com/envoy.config.cluster.v3.Cluster",
>         "name": "default/httpbin/80/da39a3ee5e",
>         "alt_stat_name": "default_httpbin_80",
>         "type": "EDS",
>         "eds_cluster_config": {
>           "eds_config": {
>             "api_config_source": {
>               "api_type": "GRPC",
>               "transport_api_version": "V3",
>               "grpc_services": [
>                 {
>                   "envoy_grpc": {
>                     "cluster_name": "contour",
>                     "authority": "contour"
>                   }
>                 }
>               ]
>             },
>             "resource_api_version": "V3"
>           },
>           "service_name": "default/httpbin"
>         },
>         "connect_timeout": "2s",
>         "common_lb_config": {
>           "healthy_panic_threshold": {}
>         }
>       }
>     }
>   ],
>   "type_url": "type.googleapis.com/envoy.config.cluster.v3.Cluster",
>   "nonce": "1"
> }
> INFO[0000] Sending incremental ACK discover request      currentVersion=5 response_nonce=1 version_info=5
> {
>   "node": {
>     "id": "ContourCLI"
>   },
>   "type_url": "type.googleapis.com/envoy.config.cluster.v3.Cluster",
>   "response_nonce": "1"
> }
> INFO[0015] Received Discovery Response                   currentVersion=5 nonce=2 resp_system_version_info=6
> {
>   "system_version_info": "6",
>   "type_url": "type.googleapis.com/envoy.config.cluster.v3.Cluster",
>   "removed_resources": [
>     "default/echo-service/80/da39a3ee5e"
>   ],
>   "nonce": "2"
> }
> INFO[0015] Sending incremental ACK discover request      currentVersion=6 response_nonce=2 version_info=6
> {
>   "node": {
>     "id": "ContourCLI"
>   },
>   "type_url": "type.googleapis.com/envoy.config.cluster.v3.Cluster",
>   "response_nonce": "2"
> }
> ```
> 
> I guess I'm missing something though?

Ah I saw that behavior but also when just deleting the HTTPProxy entirely the clusters didn’t change. Maybe a general issue with Contour rather than a xDS one but it shows regardless

youngnick:
I agree that we'll need to come back and do more about this, but what do you think about getting this in as-is now @sunjayBhatia?

sunjayBhatia (approved):
yeah we can get this in as-is and refine as we go, implementing delta + nack handling

PR #4603, opened 2022-06-29
Add support for AppProtocol as well as service annotations

### LINKED ISSUE #2431: Support Service AppProtocol field
Kubernetes 1.8 adds an [appProtocol](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#serviceport-v1-core) field to the Service. This is an optional feature that is enabled with the ServiceAppProtocol [feature gate](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/).

We already support this concept in a Contour-specific way, so we should support this standard field in Contour too. We need to be careful how we resolve this against the equivalent field in labels and HTTPProxy.

evankanderson:
---
name: Pull request
about: Support using AppProtocol to detect backend protocol (pass-through current "tls", "h2", "h2c" values) in addition to projectcontour.io/upstream-protocol.X annotations
labels: ["release-note/small"]
---

Fixes #2431

Note that this does not do any particular validation of the AppProtocol field, but it does add tests. :grin:

### DIFF
--- a/internal/dag/accessors.go
+++ b/internal/dag/accessors.go
@@ -99,6 +99,9 @@ func upstreamProtocol(svc *v1.Service, port v1.ServicePort) string {
 	if protocol == "" {
 		protocol = up[strconv.Itoa(int(port.Port))]
 	}
+	if protocol == "" && port.AppProtocol != nil {
+		protocol = *port.AppProtocol
+	}
 	return protocol
 }
 
--- a/internal/dag/accessors_test.go
+++ b/internal/dag/accessors_test.go
@@ -23,6 +23,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/pointer"
 )
 
 func TestBuilderLookupService(t *testing.T) {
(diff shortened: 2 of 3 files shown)


internal/dag/accessors_test.go
```diff
@@ -75,10 +76,38 @@ func TestBuilderLookupService(t *testing.T) {
 		},
 	}
 
+	annotatedService := &v1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        "annotatedservice",
+			Namespace:   "default",
+			Annotations: map[string]string{"projectcontour.io/upstream-protocol.tls": "6441,6442,ssl"},
```
sunjayBhatia, line 83: is this intended to have 6442 here? none of the ports below are 6442 (tests still work I believe bc we look up w/ name and/or number)

evankanderson: Yes, this is intentionally an extra value to ensure that extra values were ignored.

evankanderson:
> I think we might want to add some validation here since the `AppProtocol` field could be a wide range of values and we assume internally that protocol really only is one of `tls` `h2` or `h2c`

This seems to only be used here:
https://github.com/projectcontour/contour/blob/main/internal/envoy/v3/cluster.go#L81

Which only uses certain specific values, and ignores the rest. I suppose we might be able to save a little memory by not copying these strings around.

internal/dag/accessors.go
```diff
@@ -99,6 +99,9 @@ func upstreamProtocol(svc *v1.Service, port v1.ServicePort) string {
 	if protocol == "" {
 		protocol = up[strconv.Itoa(int(port.Port))]
 	}
+	if protocol == "" && port.AppProtocol != nil {
```
youngnick, line 102: Shouldn't this check be before the previous one? Since the previous one will change the value of protocol if it's empty, and this one expects protocol to be empty, so it will never get hit?

youngnick (changes requested):
The idea looks good, but I'm unclear how the logic in `upstreamProtocol` works. If it's just me, maybe a comment to explain it?

evankanderson:
(Let me know if I should continue with this, or try a different avenue)

sunjayBhatia:
We may want to reconsider whether we want to do this at all since this provision Gateway API GEP implies that `appProtocol` is a pretty broad field that Gateway API does not intend to use to configure backend app details: https://github.com/kubernetes-sigs/gateway-api/pull/1333 (https://github.com/kubernetes-sigs/gateway-api/pull/1333/files#diff-5ea234dff68be219d19428c1c6cee202f7df6a867a11d294bb16b0dff0c081aeR26-R30)

PR #4604, opened 2022-06-29
Gateway API: set route parent statuses correctly

### LINKED ISSUE #4600: Gateway API: set route status correctly when routes attach to specific listeners
Right now, Contour assumes that a route's parent is a Gateway, and status is set as such. However, it's possible for routes to specify that they attach to specific listeners within a Gateway, and the status should reflect attachment to that specific listener.

https://github.com/kubernetes-sigs/gateway-api/pull/1237 implemented a verification of this in the upstream conformance tests and we are now failing that test.

skriss:
Rearranges the Gateway API DAG processor to
process each route parent ref atomically,
to enable properly setting status for each
of them.

Closes #4600.

Probably can still improve test coverage a bit, but putting it up for a first look anyway. This turned into an unfortunately large refactor, but I think the new structure is a bit clearer (and needed to fix the issue). The idea is to first compute listener conditions and resolve some of their references, and the move on to computing routes, where for each route you compute for one parentRef at a time. That's needed since we have to set status for each route parent ref (could be a ref to the entire Gateway, i.e. all listeners, or to specific listener(s)).

Also note that this will fail Gateway conformance because there's a bug in the v0.5.0-rc1 conformance tests related to this, but it's been fixed in Gateway API `main` (hence why our nightly build started failing). I've verified this PR passes the `main` Gateway conformance tests and we'll get that fix in the upcoming v0.5.0 build.


youngnick (approved):
LGTM, nice work.

skriss:
https://github.com/projectcontour/contour/pull/4604/commits/f266ba2394539d18e792ba63c5a6d4f769ed7ea3 just adds back and fixes up a unit test that I had deleted, to improve coverage.

I'll hold this until after we get Gateway API v0.5.0 so that we can fully verify it still passes all the conformance tests, no rush on it.

skriss:
Rebased to pick up Gateway API v0.5.0-rc2 but this'll still fail conformance until I can get one more PR up to fix route error codes and how the Accepted condition is set.

PR #4607, opened 2022-07-01
internal/status: Avoid meaningless Condition changes

evankanderson:
No longer automatically perform a Status update on _any_ call to AddCondition.

I recently noticed in my cluster that my HTTPRoute objects had updates to `status.parents.conditions[Valid].LastTransitionTime` once per second. In turn, this triggered my own reconcilers to check out the changes, adding load to the cluster and churn on etcd.

I could see an argument for updating the anonymous function in `RouteConditionsAccessor` in `cache.go` instead of `AddCondition` -- I don't have a strong feeling one way or the other.

### DIFF
--- a/changelogs/unreleased/4607-evankanderson-small.md
+++ b/changelogs/unreleased/4607-evankanderson-small.md
@@ -1,0 +1,1 @@
+`LastTransitionTime` should only update for object transitions when the Condition has actually changed value.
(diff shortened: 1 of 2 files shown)


skriss (approved):
LGTM, thanks @evankanderson!

youngnick:
The change makes sense to me, seems like it should be more efficient. I'll wait for @evankanderson to update before pushing the approve button though.

youngnick:
ping @evankanderson - do you still want to go ahead here?

PR #4610, opened 2022-07-07
update envoy log template keywords

yangyy93:
Updated the list of know supported envoy log template keywords according to the log fields supported by the envoy doc
https://www.envoyproxy.io/docs/envoy/v1.22.2/intro/arch_overview/observability/access_logging.html

Signed-off-by: yy <yang.yang@daocloud.io>


tsaarni (approved):
LGTM :+1:

youngnick (approved):

PR #4615, opened 2022-07-12
internal/dag: set Listener condition for invalid namespace selector

skriss:
Moves the logic for parsing/validating a Listener
namespace label selector into computeListener, and
sets Listener conditions appropriately when the
selector is invalid.


sunjayBhatia (approved):

youngnick (approved):
LGTM, nice work.

PR #4618, opened 2022-07-13
CI: don't run build/test for changelogs or site changes

skriss:
Excludes changes to the changelogs/ and site/
directories from triggering the build/test
workflow. This saves some time on the PR
process since typically a changelog is added
as a second commit, that previously would
trigger the build/test workflow a second time.

### DIFF
--- a/changelogs/unreleased/4618-skriss-small.md
+++ b/changelogs/unreleased/4618-skriss-small.md
@@ -1,0 +1,1 @@
+Don't run build/test workflow for changes to the `changelogs/` or `site/` directories.


sunjayBhatia (approved):

sunjayBhatia:
great call 👍🏽

skriss:
🤔 maybe this doesn't actually work for PRs, looks like on pushing a second commit with only a changelog, the build/test workflow still gets run since the original commit has non-excluded files. I'll do some more experimenting on my fork with this.

PR #4621, opened 2022-07-18
update Envoy to 1.23.0

skriss:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ IMAGE := $(REGISTRY)/$(PROJECT)
 SRCDIRS := ./cmd ./internal ./apis
 LOCAL_BOOTSTRAP_CONFIG = localenvoyconfig.yaml
 SECURE_LOCAL_BOOTSTRAP_CONFIG = securelocalenvoyconfig.yaml
-ENVOY_IMAGE = docker.io/envoyproxy/envoy:v1.22.2
+ENVOY_IMAGE = docker.io/envoyproxy/envoy:v1.23.0
 GATEWAY_API_VERSION ?= $(shell grep "sigs.k8s.io/gateway-api" go.mod | awk '{print $$2}')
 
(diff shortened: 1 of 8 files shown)


changelogs/unreleased/4621-skriss-minor.md
```diff
@@ -0,0 +1,4 @@
+## Update to Envoy 1.23.0
+
+Contour now uses Envoy 1.23.0.
+See [the Envoy changelog](https://www.envoyproxy.io/docs/envoy/v1.23.0/version_history/v1.23/v1.23.0) for more information on the contents of the release.
```
sunjayBhatia, line 4: one thing to note here that may be interesting for operators is to take note of some of the stats name changes

from the release notes:
```
stats listener: fixed metric tag extraction so that [stat_prefix](https://www.envoyproxy.io/docs/envoy/v1.23.0/api-v3/config/listener/v3/listener.proto#envoy-v3-api-field-config-listener-v3-listener-stat-prefix) is properly extracted. This changes the Prometheus name from envoy_listener_myprefix_downstream_cx_overflow{} to envoy_listener_downstream_cx_overflow{envoy_listener_address="myprefix"}. This does not affect the Prometheus name if stat_prefix is not set.
```

sunjayBhatia (approved):
lgtm! one small comment on the release note

changelogs/unreleased/4621-skriss-minor.md
skriss: Seems like this would only affect listeners that specify a `stat_prefix` (https://github.com/envoyproxy/envoy/issues/16103), which ours don't as far as I see, they're getting the default `address_port`

sunjayBhatia: ah yeah good call, i'm confusing the stat prefix from other places!

PR #4626, opened 2022-07-19
Consolidate access log and tls cipher suite validation logic

sunjayBhatia:
Currently it is duplicated between pkg/config and apis/projectcontour/v1alpha1 between the config file and config crd respectively. This change moves it all to one place, under apis/projectcontour/v1alpha1 (while retaining the exported code in pkg/config) since that is what we intended to move to for configuration.

The second commit in the PR has the most interesting content and commit message with some details.


apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -361,26 +361,6 @@ type EnvoyListenerConfig struct {
...
-// * ECDHE-RSA-AES128-SHA
-// * AES128-GCM-SHA256
-// * AES128-SHA
-// * ECDHE-ECDSA-AES256-GCM-SHA384
-// * ECDHE-RSA-AES256-GCM-SHA384
-// * ECDHE-ECDSA-AES256-SHA
-// * ECDHE-RSA-AES256-SHA
-// * AES256-GCM-SHA384
-// * AES256-SHA
-type TLSCipherType string
```
sunjayBhatia, line 382: removed this type since it didnt seem to be getting us anything, string list seems like it should be just fine to use

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -449,7 +445,7 @@ type EnvoyLogging struct {
 	// AccessLogFields sets the fields that JSON logging will
 	// output when AccessLogFormat is json.
 	// +optional
-	AccessLogFields AccessLogFields `json:"jsonFields,omitempty"`
+	AccessLogJSONFields AccessLogJSONFields `json:"accessLogJSONFields,omitempty"`
```
sunjayBhatia, line 448: rename for consistency

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -449,7 +445,7 @@ type EnvoyLogging struct {
 	// AccessLogFields sets the fields that JSON logging will
```
skriss, line 445:
```suggestion
	// AccessLogJSONFields sets the fields that JSON logging will
```

apis/projectcontour/v1alpha1/contourconfig.go
skriss: I'm OK with this given our stated policy that we can rename things in alpha APIs, let's just make sure we call it out as a breaking change in the release notes.

apis/projectcontour/v1alpha1/contourconfig_helpers.go
```diff
@@ -106,22 +104,51 @@ func (e *EnvoyConfig) Validate() error {
...
 
+// SanitizedCipherSuites returns a deduplicated list of TLS ciphers.
+func (e *EnvoyTLS) SanitizedCipherSuites() []string {
+	if len(e.CipherSuites) == 0 {
+		return DefaultTLSCiphers
+	}
+
+	// Not using the set abstraction here so we can maintain the
+	// list order.
+	uniqueCiphers := map[string]struct{}{}
```
skriss, line 141: Guessing the code changed after the comment was added here, since you're maintaining the ordered list in a slice -- can delete the comment, and optionally actually use a sets.String if you want (slightly nicer API but either way's fine).

skriss (changes requested):
Nice, this is a big improvement. Couple little nits

apis/projectcontour/v1alpha1/contourconfig_helpers_test.go
```diff
@@ -11,121 +11,210 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package v1alpha1
+package v1alpha1_test
```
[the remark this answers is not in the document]
sunjayBhatia: yeah just wanted to make sure no unexported helpers were being used in the tests, just to make sure i got everything right when pulling stuff over from the other package, not a big deal to go back if we need to

apis/projectcontour/v1alpha1/contourconfig_helpers.go
sunjayBhatia: ah, this is how it originally was in the pkg/config version and I tried using the sets.String api but that gives you back a sorted list of set elements (or "unsorted"/how they come out of the internal map), so the comment was a bit of a warning to not try to use the sets package

apis/projectcontour/v1alpha1/contourconfig_helpers_test.go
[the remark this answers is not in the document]
skriss: I guess it doesn't really matter, fine as-is

apis/projectcontour/v1alpha1/contourconfig_helpers.go
skriss: Yeah I just meant you could replace the usage of the map with a sets.String, but agree you'll need to keep the slice too in order to preserve order.

skriss: I think it's fine as-is, if anyone has to touch this they can figure it out.

sunjayBhatia: ah i see what you mean now

apis/projectcontour/v1alpha1/contourconfig.go
sunjayBhatia: mentioned in release notes!

apis/projectcontour/v1alpha1/contourconfig_helpers.go
```diff
@@ -131,18 +133,18 @@ func (e *EnvoyTLS) Validate() error {
...
 func (e *EnvoyTLS) SanitizedCipherSuites() []string {
 	if len(e.CipherSuites) == 0 {
 		return DefaultTLSCiphers
 	}
 
-	// Not using the set abstraction here so we can maintain the
-	// list order.
-	uniqueCiphers := map[string]struct{}{}
+	uniqueCiphers := sets.NewString()
+	// We also use a list so we can maintain the order.
```
skriss, line 143: :+1:

skriss (approved):

PR #4630, opened 2022-07-20
internal/provisioner: Unify the ensure objects deleted

izturn:
There are so many EnsureXDeleted in the sub-packages of objects , so unify them to EnsureObjectDelete to let maintain work easy


internal/provisioner/objects/object.go
```diff
@@ -38,3 +45,25 @@ func NewUnprivilegedPodSecurity() *corev1.PodSecurityContext {
...
+		}
+		return err
+	}
+	if !labels.Exist(obj, model.OwnerLabels(contour)) {
+		return nil
+	}
+	if err := cli.Delete(ctx, obj); err == nil || errors.IsNotFound(err) {
+		return nil
+	}
+	return err
```
skriss, line 68: It looks like this is returning the wrong `err` if there was an error deleting the object, i.e. the one declared on L55 rather than the one returned from the delete. Let's invert this to avoid such issues:

```suggestion
	if err := cli.Delete(ctx, obj); err != nil && !errors.IsNotFound(err) {
		return err
	}
	
	return nil
```

changelogs/unreleased/4630-izturn-minor.md
```diff
@@ -0,0 +1,3 @@
+## Unify the ensure objects deleted
```
skriss, line 1: Let's switch this to a `small` changelog (file name needs to be updated), and drop the header here, just include the line of text below.

internal/provisioner/objects/contourconfig/contourconfig.go
```diff
@@ -88,32 +88,19 @@ func setGatewayConfig(config *contour_api_v1alpha1.ContourConfiguration, contour
...
-		if err := cli.Delete(ctx, current); err != nil {
-			if errors.IsNotFound(err) {
-				return nil
-			}
-			return err
-		}
+	getter := func(ctx context.Context, cli client.Client, namespace, name string) (client.Object, error) {
+		return current(ctx, cli, namespace, name)
 	}
+	return objects.EnsureObjectDeleted(ctx, cli, contour, contour.ContourConfigurationName(), getter)
```
skriss, line 94: I don't think we need to declare `getter` here, do we? This should work:
```suggestion
	return objects.EnsureObjectDeleted(ctx, cli, contour, contour.ContourConfigurationName(), current)
```

internal/provisioner/objects/object.go
```diff
@@ -38,3 +45,25 @@ func NewUnprivilegedPodSecurity() *corev1.PodSecurityContext {
 		RunAsNonRoot: &nonRoot,
 	}
 }
+
+// objectGetter get the object with name
```
skriss, line 49:
```suggestion
// objectGetter gets an object given a namespace and name.
```

internal/provisioner/objects/object.go
```diff
@@ -38,3 +45,25 @@ func NewUnprivilegedPodSecurity() *corev1.PodSecurityContext {
...
+type objectGetter func(ctx context.Context, cli client.Client, namespace, name string) (client.Object, error)
+
+// EnsureObjectDeleted ensures that a object with the name is deleted
```
skriss, line 52:
```suggestion
// EnsureObjectDeleted ensures that an object with the given namespace and name is deleted
```

internal/provisioner/objects/service/service.go
```diff
@@ -141,20 +141,10 @@ func EnsureEnvoyService(ctx context.Context, cli client.Client, contour *model.C
...
-	if !labels.Exist(svc, model.OwnerLabels(contour)) {
-		return nil
-	}
-	if err := cli.Delete(ctx, svc); err == nil || errors.IsNotFound(err) {
-		return nil
+	getter := func(ctx context.Context, cli client.Client, namespace, name2 string) (client.Object, error) {
+		return currentService(ctx, cli, namespace, name2)
 	}
-	return err
+	return objects.EnsureObjectDeleted(ctx, cli, contour, name, getter)
```
skriss, line 147: Should be able to drop the `getter` wrapper and just pass `currentService` directly into `objects.EnsureObjectDeleted(...)`

skriss (changes requested):

internal/provisioner/objects/object.go
izturn: oops, the L65's err is shadowed, maybe unshadowing it is better for read?

internal/provisioner/objects/contourconfig/contourconfig.go
izturn: no, if don't have it, the compiler is not happy with:
>
> cannot use current (value of type func(ctx context.Context, cli client.Client, namespace string, name string) (*v1alpha1.ContourConfiguration, error)) as type objects.objectGetter in argument to objects.EnsureObjectDeleted
>

changelogs/unreleased/4630-izturn-minor.md
izturn: done

internal/provisioner/objects/object.go
izturn: done

internal/provisioner/objects/object.go
izturn: done

internal/provisioner/objects/service/service.go
izturn: same as above

internal/provisioner/objects/contourconfig/contourconfig.go
skriss: Ah, must be because the return types are different.

PR #4631, opened 2022-07-20
internal/dag: fix Gateway API HTTPS/TLS protocol handling

### LINKED ISSUE #4629: Gateway API Listeners with protocol TLS should only support TLS passthrough
ref. https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.ProtocolType
ref. https://gateway-api.sigs.k8s.io/guides/tls/#clientserver-and-tls

Listeners with a protocol of `TLS` accept TLS-encrypted TCP traffic (**not** assumed to be HTTP). Gateway API allows for TLS to either be passed through to the backend, or terminated at the proxy for this protocol. In the first case (TLS passthrough), `TLSRoute` must be used to route traffic to the backend, since it's still TLS-encrypted. In the second case (TLS termination), `TCPRoute` must be used to route traffic, since it may be non-HTTP traffic.

Contour does not and will not support `TCPRoute`, so as a result we should only be supporting TLS passthrough on Listeners using the TLS protocol.

Note that the HTTPS protocol covers the case of TLS-encrypted HTTP traffic with TLS terminating at the proxy.

skriss:
Requires the HTTPS listener protocol to
be used with TLS mode Terminate (compatible
with HTTPRoutes) and the TLS protocol to
be used with TLS mode Passthrough (compatible
with TLSRoutes).

Closes #4629.

The actual change is pretty straightforward, but a bunch of our tests were inadvertently using invalid combos of protocols and modes that had to be cleaned up.

For reference, https://gateway-api.sigs.k8s.io/guides/tls/ provides the best explanation of how this is intended to work.

### DIFF
--- a/changelogs/unreleased/4631-skriss-minor.md
+++ b/changelogs/unreleased/4631-skriss-minor.md
@@ -1,0 +1,5 @@
+## Gateway API: enforce correct TLS modes for HTTPS and TLS listener protocols
+
+Contour now enforces that the correct TLS modes are used for the HTTPS and TLS listener protocols.
+For an HTTPS listener, the TLS mode "Terminate" must be used (this is compatible with HTTPRoutes).
+For a TLS listener, the TLS mode "Passthrough" must be used (this is compatible with TLSRoutes).
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -345,31 +345,6 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 		Data: secretdata(fixture.CERTIFICATE, fixture.RSA_PRIVATE_KEY),
 	}
 
-	gatewayTLSTerminateAllNamespaces := &gatewayapi_v1beta1.Gateway{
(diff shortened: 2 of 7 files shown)


sunjayBhatia (approved):
looks good! also makes tests etc. a lot simpler which is nice

skriss:
Gonna leave this open so @youngnick can take a look and make sure I'm not missing anything re: the spec.

youngnick (approved):
This LGTM, but I would ask if @skriss thinks that we could clean up the Gateway spec here? Is it clear enough that certain Protocols map to certain Routes?

skriss:
> This LGTM, but I would ask if @skriss thinks that we could clean up the Gateway spec here? Is it clear enough that certain Protocols map to certain Routes?

https://gateway-api.sigs.k8s.io/guides/tls/ does a good job of laying it out, but I don't think it's super clear in the API documentation itself. It's an intersection of protocol, TLS mode, and route types, so would have to think about the best place to update. I will take a look.

PR #4640, opened 2022-07-26
Gateway API: run conformance via a Go test

### LINKED ISSUE #4616: run Gateway API conformance from our own Go test, using the GWAPI conformance module
Currently we're running the Gateway API conformance tests by [cloning the Gateway API repo and running the tests from there](https://github.com/projectcontour/contour/blob/main/test/scripts/run-gateway-conformance.sh#L44-L48), however as I understand it the intended way of running them is to have our own Go test that mirrors the one [in the Gateway API repo](https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/conformance_test.go#L33-L57), and uses the `conformance` Go module. Importantly, this will allow us to customize the [SupportedFeatures](https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/conformance_test.go#L51-L53); we support HTTPRoute query param matching and should [opt into the conformance test for it](https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/utils/suite/suite.go#L49-L50).

skriss:
Switches to running Gateway API conformance
via a Go test that uses the conformance
module. This allows us to customize the
settings for the run, including opting into
tests for extended conformance features.

Closes #4616.


test/scripts/run-gateway-conformance.sh
```diff
@@ -41,8 +41,22 @@ spec:
...
+
+if [ "$GATEWAY_API_VERSION" = "$GO_MOD_GATEWAY_API_VERSION" ]; then
+  go test -tags conformance ./test/conformance/gatewayapi --gateway-class=contour
+else 
+  cd $(mktemp -d)
+  git clone https://github.com/kubernetes-sigs/gateway-api
+  cd gateway-api
+  git checkout "${GATEWAY_API_VERSION}"
+  go test ./conformance -gateway-class=contour
+fi
```
skriss, line 62: Re: https://github.com/projectcontour/contour/issues/4616#issuecomment-1185932732, this seems like the best option for now, as I definitely want to (a) opt into the tests for extended features that we support; and (b) continue running new `main` conformance tests on a nightly basis as they're added upstream.

sunjayBhatia: do we have somewhere we explicitly set `GATEWAY_API_VERSION`, from https://github.com/projectcontour/contour/blob/e895c86a85eb6ba076276fe04dea83ddc029bc29/Makefile#L10 seems like it gets set by the version in the module too so they won't diverge? I'm probably missing some place where this works when we bump the version on `main` or something

skriss: Yeah, for the daily build that runs the conformance tests from `main` it's set here:
https://github.com/projectcontour/contour/blob/main/.github/workflows/build_daily.yaml#L141-L144

sunjayBhatia (approved):

test/scripts/run-gateway-conformance.sh
sunjayBhatia: oh duh, was looking at the wrong workflow file 🙃

PR #4646, opened 2022-07-27
test/scripts: fix upgrade from version logic when on a tag

skriss:
Updates the "upgrade from" version logic for when
on a tag to get the correct previous tag.

### DIFF
--- a/test/scripts/get-contour-upgrade-from-version.sh
+++ b/test/scripts/get-contour-upgrade-from-version.sh
@@ -6,7 +6,7 @@ set -o pipefail
 
 if CURRENT_TAG=$(git describe --tags --exact-match 2>/dev/null); then
   # We are on a tag, so find previous tag to this one.
-  git tag -l --sort=-v:refname | grep -A10 $CURRENT_TAG | grep -v 'alpha\|beta\|rc' | head -1
+  git tag -l --sort=-v:refname | grep -v 'alpha\|beta\|rc' | grep -A1 -x $CURRENT_TAG | grep -v "$CURRENT_TAG" | head -1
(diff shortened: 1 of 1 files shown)


skriss:
The upgrade tests for the build of the `v1.22.0` tag failed, they were trying to upgrade from v1.22.0 to v1.22.0 which is not right, so this updates the logic to pull e.g. `v1.21.1` when on the `v1.22.0` tag.

test/scripts/get-contour-upgrade-from-version.sh
```diff
@@ -6,7 +6,7 @@ set -o pipefail
 
 if CURRENT_TAG=$(git describe --tags --exact-match 2>/dev/null); then
   # We are on a tag, so find previous tag to this one.
-  git tag -l --sort=-v:refname | grep -A10 $CURRENT_TAG | grep -v 'alpha\|beta\|rc' | head -1
+  git tag -l --sort=-v:refname | grep -v 'alpha\|beta\|rc' | grep -A1 -x $CURRENT_TAG | grep -v "$CURRENT_TAG" | head -1
```
sunjayBhatia, line 9:
```suggestion
  git tag -l --sort=-v:refname | grep -v 'alpha\|beta\|rc' | grep -A1 -x $CURRENT_TAG | tail -1
```

nit: might be able to simplify a little but otherwise 👍🏽

sunjayBhatia (approved):

test/scripts/get-contour-upgrade-from-version.sh
skriss: ah yeah good point, let me make that change.

Issue #1894, opened 2019-11-13
Instructions for presenting grpc services needs a guide


davecheney:
We need a guide that concisely describes the HTTPProxy configuration and service annotations required to proxy grpc through contour.

twfx7758:
me too

reagul:
I can help with this. Can i get assigned ?

stevesloka:
Awesome all yours @reagul! 

Some work steps in my head if this helps:
- I think this would be a good "guide" where you'd add a new file here: https://github.com/projectcontour/contour/tree/master/site/_guides
- If you want to test what it looks like on your machine you can run: `make site-devel` and open up a browser to `localhost:4000`
- It would be good to link to this guide from the `annotations` page as well as the `HTTPProxy`

JasonMorgan:
hey folks! Any progress on this? I was just looking for this exact info. Happy to help if anyone can point me at a starting point.

jpeach:
> hey folks! Any progress on this? I was just looking for this exact info. Happy to help if anyone can point me at a starting point.

To expose gRPC services, you would need to enable TLS on the [virtual hosts](https://projectcontour.io/docs/v1.9.0/api/#projectcontour.io/v1.VirtualHost) and set the `h2` protocol on any upstream [service](https://projectcontour.io/docs/v1.9.0/api/#projectcontour.io/v1.Service) entries.

If you'd like to contribute a specific user guide, that would be great. Steve's comment above and the [site contribution doc](https://github.com/projectcontour/contour/blob/main/SITE_CONTRIBUTION.md) should help you get started with the docs.

reagul:
@JasonMorgan i am still working on this one. I will post something soon.

danthegoodman1:
Any update on this?

youngnick:
Hmm, it doesn't look like there's much movement here. I'd agree that this one is prime candidate for a fresh guide. Perhaps @OrlinVasilev or @jonasrosland could raise this one at a Tech Docs WG meeting?

OrlinVasilev:
@reagul did you make any progress on that or we can restart that topic?

samelie:
A guide would be really helpful. Am stumbling piecing together what's required in the contour+envoy config to make this work

OrlinVasilev:
@jpeach - are you still interested in that or if anyone else wants to take it ?

youngnick:
I think that James has moved on and will not be interested in updating here.

samelie:
[...]

PR #4648, opened 2022-07-27
test/e2e/httpproxy: Add test for gRPC server behind Contour with h2c

sunjayBhatia:
Adds test verification for plaintext gRPC server

Part of sequence of PRs that will add test verification for various gRPC
configurations, and a guide to set up gRPC with Contour.

Uses test fixture from github.com/mhausenblas/yages which may be
replaced with a more fully featured option.

Updates: #1894


sunjayBhatia:
hm, thought the retries and timeouts would fix flakiness, gotta do a little more investigation, maybe add some backoffs to the retries

skriss (commented):
This looks great to me -- I'm not too familiar with the gRPC-web stuff but looks like you've got it sorted out. Are they passing consistently or is there still some flakiness?

sunjayBhatia:
> This looks great to me -- I'm not too familiar with the gRPC-web stuff but looks like you've got it sorted out. Are they passing consistently or is there still some flakiness?

i think with these retry options its looking pretty stable: https://github.com/projectcontour/contour/pull/4648/files#diff-ee633bab9384727cc410aa9165052e4aa38763741c18bdd1609c33a9b9159048R98-R104, the main flakiness i was seeing earlier was the service not being healthy so this mimics our requestuntil logic (maybe even more forgiving w/ exponential backoff)

skriss (approved):
LGTM, this is great to get codified!

sunjayBhatia:
> LGTM, this is great to get codified!

if we do end up allowing the grpc-web filter to be disabled, should help there too

docs guide coming soon as well

PR #4649, opened 2022-07-28
test/e2e/upgrade: remove WaitForEnvoyDaemonSetOutOfDate check

### LINKED ISSUE #4647: upgrade test flakes after releasing v1.22.0
I've seen several (but not all) of the upgrade test runs fail after releasing v1.22.0. The failures can be seen at the top of https://github.com/projectcontour/contour/actions/workflows/prbuild.yaml. They all point to the code that waits for the Envoy DaemonSet to be out-of-date, and then updated. I'm wondering if the upgrade is a no-op so it never becomes out-of-date, though that doesn't explain why it still succeeds sometimes. Needs a little more digging into.

skriss:
Removes the check that the Envoy DaemonSet
has become out-of-date after the upgrade,
since during the time that we're waiting
for the Contour deployment to update, the
Envoy DS is likely to become up-to-date
again.

Closes #4647.

### DIFF
--- a/test/e2e/deployment.go
+++ b/test/e2e/deployment.go
@@ -367,20 +367,6 @@ func (d *Deployment) EnsureEnvoyDeployment() error {
 	return d.ensureResource(d.EnvoyDeployment, new(apps_v1.Deployment))
 }
 
-// Wait for Envoy daemonset update to go out of date due to the
-// update propagating. Possibly needed before calling WaitForEnvoyDaemonSetUpdated
-// to ensure that does not pass without waiting for update to propagate.
-func (d *Deployment) WaitForEnvoyDaemonSetOutOfDate() error {
-	daemonSetUpdated := func() (bool, error) {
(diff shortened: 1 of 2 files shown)


test/e2e/deployment.go
```diff
@@ -394,7 +380,8 @@ func (d *Deployment) waitForEnvoyDaemonSetUpdated() error {
 		if err := d.client.Get(context.TODO(), client.ObjectKeyFromObject(d.EnvoyDaemonSet), tempDS); err != nil {
 			return false, err
 		}
-		return tempDS.Status.NumberAvailable > 0 &&
+		return tempDS.Status.ObservedGeneration == tempDS.Generation &&
```
skriss, line 383: Added in the observed generation vs. generation check which I think should mostly prevent any race conditions here.

sunjayBhatia (approved):

PR #4656, opened 2022-08-03
Switch to image with better healthchecking

sunjayBhatia:
Rather than TCP check, use grpc-health-probe

Should hopefully help with flakes since upstream won't go healthy until the gRPC server is actually up

See https://github.com/sunjayBhatia/yages for fork

This will be the image/code used for adding upstream TLS tests as well

Updates #1894

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,9 +16,9 @@ require (
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
 	github.com/imdario/mergo v0.3.12
 	github.com/jetstack/cert-manager v1.5.1
-	github.com/mhausenblas/yages v0.0.0-20181104055001-8d495d7135de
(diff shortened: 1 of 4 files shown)


skriss:
WDYT about moving the fork into the projectcontour org?

sunjayBhatia:
> WDYT about moving the fork into the projectcontour org?

sounds good to me, wasn't sure if i should fork it here to start but can figure out how to transfer/re-fork it over

sunjayBhatia:
moved over to https://github.com/projectcontour/yages, worth someone taking a look at the repo+goreleaser settings+code to see its all sane

skriss (approved):
Took a quick look at the yages fork as well and it all looks reasonable to me, :shipit:!

PR #4661, opened 2022-08-09
Bump golangci-lint and fix a few lint issues

sunjayBhatia:

Issue #4349, opened 2022-02-17
The go.mod file is go version 1.15，but the go version has been upgraded to 1.17.5


yankay:
According the PR: https://github.com/projectcontour/contour/pull/4232 , the golang version has been upgrade to 1.17.5.

But there are still 1.15 in go.mod file: https://github.com/projectcontour/contour/blob/main/go.mod#L3.

Should we upgrade it to make the upgrade completely

sunjayBhatia:
The version in the go.mod file represents the minimum version of Go needed to compile the contour module. It’s a bit older than the latest to ensure we don’t cause compatibility issues as I believe we don’t use any language features only in later versions that would force people to upgrade their Go technically

yankay:
Thanks.

PR #4662, opened 2022-08-10
Make lint/staticcheck happy

izturn:
Package io/ioutil is deprecated: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code.

### DIFF
--- a/.github/workflows/prbuild.yaml
+++ b/.github/workflows/prbuild.yaml
@@ -18,8 +18,8 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v2
         with:
-          version: v1.45.0
-          args: --build-tags=e2e
+          version: v1.48.0
+          args: --build-tags=e2e,conformance,gcp,oidc,tools,none
       - uses: act10ns/slack@v1
         with:
           status: ${{ job.status }}
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,7 +1,5 @@
 run:
   timeout: 10m
-  skip-files:
-    - apis/projectcontour/v1/kubeconditions.go # copy-pasted from k8s.io/apimachinery, has invalid header
 
 linters:
   enable:
@@ -47,7 +45,8 @@ issues:
   # And we can't nolint them by line, because https://github.com/golangci/golangci-lint/issues/741
   - linters:
     - staticcheck
-    text: "SA1019: package github.com/golang/protobuf"
+    text: >-
+      SA1019: "github.com/golang/protobuf
   - linters:
     - golint
     text: "var `ingress_https` should be `ingressHTTPS`"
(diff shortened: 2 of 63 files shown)


tsaarni (approved):

tsaarni (changes requested):
Shouldn't we step minumum go version in `go.mod` to 1.16 since the replacement functions for the deprecated `ioutil` were also introduced in that version?

In the PR's CI checks [lint gets killed with error code 137](https://github.com/projectcontour/contour/runs/7760857339?check_suite_focus=true#step:3:87) but no other output for some reason, possibly https://github.com/golangci/golangci-lint/issues/576?

lewgun:
i think we should upgrade to go1.18, because with the **go release policy**  the other versions below it will not be supported by go team, @tsaarni  what's your opinion ?

tsaarni:
> i think we should upgrade to go1.18, because with the **go release policy** the other versions below it will not be supported by go team, @tsaarni what's your opinion ?

I dont have strict opinion but I do recall [this response](https://github.com/projectcontour/contour/issues/4349#issuecomment-1043029475) from @sunjayBhatia which makes sense to me. Though I have not tested if the code actually compiles with old compilers? :)

lewgun:
got it, so let's upgrade it to go1.16?

izturn:
done

tsaarni (approved):
LGTM :+1:

I still wonder why the lint step in CI check now constantly [fails with error 137](https://github.com/projectcontour/contour/runs/7763283989?check_suite_focus=true#step:3:87)? I don't think there is anything wrong in the PR and running golangci-lint-1.45.0 locally for it succeeded. I wonder if this is rather related to github's environment. BTW @sunjayBhatia is in the middle of updating the version of golanci-lint https://github.com/projectcontour/contour/pull/4661 but I don't know if that would fix this.

tsaarni:
>   wonder if this is rather related to github's environment. BTW @sunjayBhatia is in the middle of updating the version of golanci-lint #4661 but I don't know if that would fix this.

The same happens with an "empty" PR :thinking: Since @sunjayBhatia 's PR #4661 seemed to be running ok with newer golangci-lint, maybe that needs to go in first.

sunjayBhatia:
@izturn you should be able to rebase or merge `main` and focus this change on just switching from `ioutil` to the `os` package and bumping the go.mod go version to 1.16, other lint issues have been taken care of

lewgun:
@sunjayBhatia  only ioutil => os, not ioutil => io ?

lewgun:
done

izturn:
@sunjayBhatia unfortunately, after merging with **main**, there are more issues about **github.com/goalng/protobuf**, so I exclude them, in **.golangcl.yml**, it's passed in my local dev env, but why it fall in the **action**?

izturn:
done

sunjayBhatia (approved):

PR #4669, opened 2022-08-13
cmd/gateway-provisioner: Add flag: leader-election-namespace

### LINKED ISSUE #4668: gateway-provisioner: create manager fails when enabling leader election
run contour gateway-provisioner in local dev like this:
```
 ./contour  gateway-provisioner --metrics-addr=127.0.0.1:8080 --enable-leader-election
```
will get the error:
```
time="2022-08-10T08:46:56+08:00" level=error msg="failed to create contour gateway provisioner" caller="contour.go:87" context=kubernetes error="failed to create manager: unable to find leader election namespace: not running in-cluster, please specify LeaderElectionNamespace" logger=setup
```
this can be fixed by add a new flag: `--leader-election-namespace=xxx`

fix by #4669 

**Environment:**

- Contour version: v1.22
- Kubernetes version: (use `kubectl version`): v1.23.1
- Kubernetes installer & version: kind: v0.14.0
> NAME="Ubuntu"
> VERSION="20.04.3 LTS (Focal Fossa)"

izturn:
Add a new flag `leader-election-namspace` 
fix #4668


cmd/contour/gatewayprovisioner.go
```diff
@@ -52,6 +52,10 @@ func registerGatewayProvisioner(app *kingpin.Application) (*kingpin.CmdClause, *
 		Default(config.metricsBindAddress).
 		StringVar(&config.metricsBindAddress)
 
+	cmd.Flag("leader-election-namespace", "The namespace in which the leader election resource will be created.").
+		Default(config.leaderElectionNamespace).
```
sunjayBhatia, line 56: hm, right now the default is empty since we just added this field in this PR and havent set a default here: https://github.com/projectcontour/contour/blob/a20cac5011b5c2a3614b542dd988a6fb5369d625/cmd/contour/gatewayprovisioner.go#L34-L41

since the example YAML configuration etc. sets up the provisioner to run in the `projectcontour` namespace by default we can probably set that as the default, (similar to the contour serve controller here https://github.com/projectcontour/contour/blob/a20cac5011b5c2a3614b542dd988a6fb5369d625/cmd/contour/serve.go#L130)

we should also note that the RBAC configuration needs to be updated if this flag is used in-cluster

izturn: ok, i have not set the default value to `projectcontour`, is just want to keep it as before, I will update it

izturn: > we should also note that the RBAC configuration needs to be updated if this flag is used in-cluster

I can't understand this, pls give me more info about it

izturn:
@sunjayBhatia  any advise about how to make the e2e `Build and Test Pull Request / e2e (kubernetes:n-1, ContourConfiguration) (pull_request) ` pass?, i dig into the source code but get nothing

sunjayBhatia:
> @sunjayBhatia any advise about how to make the e2e `Build and Test Pull Request / e2e (kubernetes:n-1, ContourConfiguration) (pull_request) ` pass?, i dig into the source code but get nothing

Looks like a flake in an unrelated test, rerunning the jobs

cmd/contour/gatewayprovisioner.go
```diff
@@ -41,25 +42,29 @@ func registerGatewayProvisioner(app *kingpin.Application) (*kingpin.CmdClause, *
...
+		StringVar(&conf.envoyImage)
 
 	cmd.Flag("metrics-addr", "The address the metric endpoint binds to. It can be set to 0 to disable serving metrics.").
-		Default(config.metricsBindAddress).
-		StringVar(&config.metricsBindAddress)
+		Default(conf.metricsBindAddress).
+		StringVar(&conf.metricsBindAddress)
+
+	cmd.Flag("leader-election-namespace", "The namespace in which the leader election resource will be created.").
+		Default(config.GetenvOr("CONTOUR_NAMESPACE", "projectcontour")).
```
sunjayBhatia, line 57: we don't currently set the namespace in the example manifest for the gateway provision, so i think this can be simplified a bit
```suggestion
		Default("projectcontour").
```

izturn:
the `Build and Test Pull Request / Codespell (pull_request) ` is unhappy, but in my local env, it's ok, so should I try to fix them in this/another pr?

sunjayBhatia:
> the `Build and Test Pull Request / Codespell (pull_request) ` is unhappy, but in my local env, it's ok, so should I try to fix them in this/another pr?

You should be able to get codespell to pass by merging main since https://github.com/projectcontour/contour/commit/491e263e2ab3f4b6e6c0fa629b580491a03ab2da should fix it

cmd/contour/gatewayprovisioner.go
sunjayBhatia: > > we should also note that the RBAC configuration needs to be updated if this flag is used in-cluster
> 
> I can't understand this, pls give me more info about it

See https://github.com/projectcontour/contour/blob/491e263e2ab3f4b6e6c0fa629b580491a03ab2da/examples/gateway-provisioner/01-roles.yaml#L179

When this flag is added, users should be aware this resource will have to change if they are using this flag in cluster etc., not a big thing but maybe nice to find a place to note it

izturn:
done

sunjayBhatia (approved):

changelogs/unreleased/4669-izturn-small.md
```diff
@@ -0,0 +1 @@
+Add a new flag `leader-election-namspace` for gateway-provisioner
```
skriss, line 1:
```suggestion
Add a new flag `leader-election-namespace` for gateway-provisioner
```

cmd/contour/gatewayprovisioner.go
```diff
@@ -52,6 +53,10 @@ func registerGatewayProvisioner(app *kingpin.Application) (*kingpin.CmdClause, *
 		Default(config.metricsBindAddress).
 		StringVar(&config.metricsBindAddress)
 
+	cmd.Flag("leader-election-namespace", "The namespace in which the leader election resource will be created.").
+		Default(config.leaderElectionNamespace).
```
skriss, line 57: I'd prefer to use the Downward API to set an env var containing the namespace that the gateway provisioner is deployed in, and then use the value of that env var as the default here. That way if the user deploys the gateway provisioner into a different namespace, they don't also need to set this flag. This is how Contour itself does it -- see https://github.com/projectcontour/contour/blob/main/examples/contour/03-contour.yaml#L77-L82 for defining the env var using the downward API, and then https://github.com/projectcontour/contour/blob/a20cac5011b5c2a3614b542dd988a6fb5369d625/cmd/contour/serve.go#L130 for how it's used as the default.

skriss (changes requested):

cmd/contour/gatewayprovisioner.go
sunjayBhatia: yeah thats a good point, was thinking we might do that in a follow up (previous changes in the PR had something similar) but might as well do it here

maybe use `CONTOUR_PROVISIONER_NAMESPACE` or something similar as the env var

changelogs/unreleased/4669-izturn-small.md
izturn: done

cmd/contour/gatewayprovisioner.go
izturn: done

cmd/contour/gatewayprovisioner.go
```diff
@@ -41,25 +42,29 @@ func registerGatewayProvisioner(app *kingpin.Application) (*kingpin.CmdClause, *
...
+		StringVar(&provisionerConfig.envoyImage)
 
 	cmd.Flag("metrics-addr", "The address the metric endpoint binds to. It can be set to 0 to disable serving metrics.").
-		Default(config.metricsBindAddress).
-		StringVar(&config.metricsBindAddress)
+		Default(provisionerConfig.metricsBindAddress).
+		StringVar(&provisionerConfig.metricsBindAddress)
+
+	cmd.Flag("leader-election-namespace", "The namespace in which the leader election resource will be created.").
+		Default(config.GetenvOr("CONTOUR_PROVISIONER_NAMESPACE", "projectcontour")).
```
sunjayBhatia, line 57: to accompany this change, I think we would also like to add an env var configuration like this: https://github.com/projectcontour/contour/blob/491e263e2ab3f4b6e6c0fa629b580491a03ab2da/examples/contour/03-contour.yaml#L77-L82

to [this file](https://github.com/projectcontour/contour/blob/main/examples/gateway-provisioner/03-gateway-provisioner.yaml)

izturn: done

sunjayBhatia:
one more thing @izturn you'll have to re-run `make generate` and commit the changes to the example YAMLs

izturn:
done

sunjayBhatia (approved):
lgtm, ill let @skriss have another look if needed before merging

skriss (approved):
LGTM, thanks @izturn!

PR #4674, opened 2022-08-18
crd/contourDeployments: Add field for set the flag "--ingress-class-name"

### LINKED ISSUE #4673: crd/ContourDeployment: Add field for set '--ingress-class-name'
 There is no way for setting the flag `--ingress-class-name`, when creating `contour` by `gateway-provisioner`, so we need to add a field to crd/ContourDeployment to fill the gap

fix by #4674

izturn:
fix #4673 


izturn:
sorry, it‘s my fault, it‘s can be pass by RuntimeSettings

PR #4676, opened 2022-08-18
gateway provisioner: support changing Contour log level

izturn:
add the flag `--debug` to enable debug logging


apis/projectcontour/v1alpha1/contourdeployment.go
```diff
@@ -59,6 +59,11 @@ type ContourSettings struct {
 	//
 	// +optional
 	NodePlacement *NodePlacement `json:"nodePlacement,omitempty"`
+
+	// Enable debug logging.
+	//
+	// +optional
+	Debug bool `json:"debug,omitempty"`
```
sunjayBhatia, line 66: personally would prefer using this type https://github.com/projectcontour/contour/blob/491e263e2ab3f4b6e6c0fa629b580491a03ab2da/apis/projectcontour/v1alpha1/contourconfig.go#L297-L306

rather than a boolean here

something like:

```suggestion
	LogLevel LogLevel `json:"debug,omitempty"`
```

sunjayBhatia: with a default of `info`

izturn: i just use `bool` here, is just want to match the 
```
```
in serve.go, if you think `LogLevel` is better, I will change it

sunjayBhatia: same as the other PR:

yeah I would prefer the rename: personally don't love the existing flag name and I would like to new field to be a bit more readable

no need to match the field name exactly with the flag, as people using the provisioner don't need to know the exact flag name, we can make it configurable with something more clear what the meaning is

izturn: done

changelogs/unreleased/4676-izturn-small.md
```diff
@@ -1 +1 @@
-Add the flag to enable debug logging
\ No newline at end of file
+Add the flag to set the log level for Contour
```
sunjayBhatia, line 1:
```suggestion
Add Contour log level configurability to ContourDeployment resource.
```

izturn: done

skriss (commented):
Sorry for the belated review, but this looks pretty good to me - just need to update codegen by running `make generate`. Had one small nit that hopefully you can address as well. Thanks!

(see https://github.com/projectcontour/contour/pull/4677#issuecomment-1234408686 for info on the `make generate` step)

internal/provisioner/objects/deployment/deployment.go
```diff
@@ -98,6 +99,11 @@ func DesiredDeployment(contour *model.Contour, image string) *appsv1.Deployment
 		fmt.Sprintf("--leader-election-resource-name=%s", contour.LeaderElectionLeaseName()),
 		fmt.Sprintf("--envoy-service-name=%s", contour.EnvoyServiceName()),
 	}
+
+	if contour.Spec.LogLevel == v1alpha1.DebugLog {
+		args = append(args, "--debug")
+	}
```
skriss, line 105: Please add some unit test cases to cover this -- looks like you can do something similar to https://github.com/projectcontour/contour/blob/main/internal/provisioner/objects/deployment/deployment_test.go#L156.

apis/projectcontour/v1alpha1/contourdeployment.go
```diff
@@ -59,6 +59,13 @@ type ContourSettings struct {
 	//
 	// +optional
 	NodePlacement *NodePlacement `json:"nodePlacement,omitempty"`
+
+	// LogLevel sets the log level for Contour
+	// Allowed values are "info", "debug".
+	//
+	// +optional
+	// +kubebuilder:default=info
+	LogLevel LogLevel `json:"logLevel,omitempty"`
```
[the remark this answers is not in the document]
sunjayBhatia: since we're not using kubebuilder defaults anywhere, should we remove here as well?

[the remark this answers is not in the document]
skriss: Yeah I guess that makes sense since we're applying them internally.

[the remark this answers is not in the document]
izturn: done

internal/provisioner/objects/deployment/deployment.go
izturn: done

internal/provisioner/model/model.go
```diff
@@ -673,6 +679,7 @@ func New(cfg Config) *Contour {
 		},
 		Spec: ContourSpec{
 			ContourReplicas: cfg.Replicas,
+			LogLevel:        contourv1alpha1.DebugLog,
```
skriss, line 682: I think we should leave the default as info, switching to debug logging should be opt-in.

izturn: i find the New only used in unittests, so set the `LogLevel` to `DebugLog`, I will change it

skriss:
Looks like there are some merge conflicts to resolve here.

izturn:
@skriss  PTAL

skriss (approved):

PR #4681, opened 2022-08-19
crd/ContourDeployment: Add field 'podAnnotations'

izturn:
add the field `podAnnotations` to `crd/ContourDeployment` to enable adding more annotations to `pod/envoy`


izturn:
ping @skriss @tsaarni

izturn:
ping again @skriss @tsaarni

sunjayBhatia (changes requested):
would prefer this field on the `EnvoySettings`

also would be good to add some unit test coverage for the changes in internal/provisioner/objects/dataplane

sunjayBhatia:
needs a `main` merge as well

internal/provisioner/controller/gateway.go
```diff
@@ -292,6 +292,10 @@ func (r *gatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 					Tolerations:  nodePlacement.Tolerations,
 				}
 			}
+
+			for k, v := range gatewayClassParams.Spec.Envoy.PodAnnotations {
```
sunjayBhatia, line 296: hm, we don't actually have any coverage of the actual logic in this change which is unfortunate, this would have to be covered by a test in this package

probably important to test that we're appending on top of the existing annotations since that ends up becoming API surface

do we want to prevent overwriting of existing annotations?

changelogs/unreleased/4681-izturn-small.md
```diff
@@ -0,0 +1 @@
+add the field podAnnotations to crd/ContourConfiguration to enable adding more annotations to pod/envoy
```
sunjayBhatia, line 1: worth mentioning in this PR and others youve submitted what the behavior of the added configuration is, what the defaults are, and how user configuration interacts with the defaults

izturn: done

apis/projectcontour/v1alpha1/contourdeployment.go
```diff
@@ -127,6 +127,10 @@ type EnvoySettings struct {
 	//
 	// +optional
 	NodePlacement *NodePlacement `json:"nodePlacement,omitempty"`
+
+	// PodAnnotations holds the annotations that will be add to the envoy‘s pod.
```
skriss, line 131: Little nit on wording (will need to re-run `make generate` after this as well):
```suggestion
	// PodAnnotations defines annotations to add to the Envoy pods.
```

skriss (changes requested):
One small nit, otherwise I think we can get this in.

apis/projectcontour/v1alpha1/contourdeployment.go
izturn: done

izturn:
`codegen` is not happy, what can i do?

skriss:
> `codegen` is unhappy, what can i do?

I re-ran it and it went through successfully

izturn:
> > `codegen` is unhappy, what can i do?
> 
> I re-ran it and it went through successfully

ok, thx

skriss (approved):
LGTM pending CI, thanks @izturn

skriss:
@sunjayBhatia mind taking a quick look here? OK to merge from my PoV.

sunjayBhatia (approved):

PR #4682, opened 2022-08-19
Correct spelling mistages catched by newer version of codespell

tsaarni:
Corrected some spelling mistakes found by codespell and added few workarounds for false positives, such as unfortunate character combination `bU` in base64 encoded test data which was taken for misspelled `by`, fixed now by just re-generating new certificate with less unfortunate encoding.

List of errors corrected by this PR is:

```
Error: ./internal/featuretests/kubernetes.go:82: bU ==> by
Error: ./internal/fixture/certificate_constants.go:65: bU ==> by
Error: ./internal/xdscache/v3/secret_test.go:618: bU ==> by
Error: ./internal/dag/policy_test.go:35: anotations ==> annotations
Error: ./internal/dag/policy.go:61: ro ==> to, row, rob, rod, roe, rot
Error: ./internal/dag/policy.go:62: ro ==> to, row, rob, rod, roe, rot
Error: ./internal/dag/policy.go:66: ro ==> to, row, rob, rod, roe, rot
Error: ./internal/dag/policy.go:67: ro ==> to, row, rob, rod, roe, rot
Error: ./test/e2e/bench/bench_test.go:254: warmup ==> warm up, warm-up
Error: ./test/e2e/bench/bench_test.go:255: warmup ==> warm up, warm-up
Error: ./test/e2e/bench/bench_test.go:272: warmup ==> warm up, warm-up
Error: ./design/ingress-status-loadbalancer.md:41: administator ==> administrator
```


tsaarni:
Long time ago I changed some test certs to be generated on the fly by using https://github.com/tsaarni/certyaml.  I was thinking if I should replace rest now but there are some use cases like validating PKCS#1 encoded private keys `RSA PRIVATE KEY` and `EC PARAMETERS` which could not be generated with certyaml currently. So I took the simple route and just generated new certificate and private key, and copied it where ever we had the old one.

I was also wondering where this problem with codespell originates.  New version coming from actions? Would it be possible to get a fixed version? I checked https://github.com/codespell-project/actions-codespell/ and there seems to be tag `v1.0` while we have https://github.com/projectcontour/contour/blob/87fc2ffba5b636e836f3c2cd04d57a767c244c32/.github/workflows/prbuild.yaml#L36

It seems to me that `v1.0` might be moving target too.

skriss (approved):
LGTM, yeah the `v1.0` action tag does look to be mutable, but probably makes sense to switch to that rather than `master` anyway, so we're at least pinning to a released version.

skriss (approved):

PR #4690, opened 2022-08-28
internal/dag: set connect timeout when route policy is unset

### LINKED ISSUE #4688: Global connect-timeout is ignored when route's timeoutPolicy is not present
As found by @ravilr

> @tsaarni @sunjayBhatia @skriss while upgrade testing from contour-v1.19 to latest v1.22, seeing this issue where ContourConfig file has timeout.connect-timeout set to 5s, but the final envoy config rendered by the xdscache keeps fluctuating certain Cluster's connect_timeout b/w 5s and 2s on any onChanges(). it is expected to be 5s for all dynamic Cluster's connect_timeout.
>
> I think this resetting of Cluster's ConnectTimeout=0 seems to be causing this. Why is this resetting Cluster's ConnectTimeout when a Route's TimeoutPolicy isn't set ? shouldn't this be setting Cluster's ConnectTimeout to passed func arg connectTimeout also ? (just like in the final return below).

https://github.com/projectcontour/contour/blob/a6036886fb5eceebe9b4dd5a0aa4b16919f6eb12/internal/dag/policy.go#L427-L436

tsaarni:
The global connect-timeout configuration value was not taking effect for clusters when timeoutPolicy was not set for the route.

The bug was found and pinpointed by @ravilr https://github.com/projectcontour/contour/commit/0e3d735feb66c0041d4400925cf8b7fd293bc430#r82376547

Fixes #4688

### DIFF
--- a/changelogs/unreleased/4690-tsaarni-small.md
+++ b/changelogs/unreleased/4690-tsaarni-small.md
@@ -1,0 +1,1 @@
+The global connect-timeout configuration value was not taking effect for routes that did not have timeoutPolicy set.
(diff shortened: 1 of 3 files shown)


sunjayBhatia (approved):

skriss (approved):
LGTM, thanks @tsaarni!

PR #4700, opened 2022-09-05
apis/projectcontour: add gvr for contourdeployments

izturn:
add gvr definition for ContourDeployments


skriss (changes requested):
Rebasing/merging main should get rid of the codegen issue, I've merged a PR that fixes that.

We can also drop the changelog file from this one, typically we only need changelogs for user-facing changes.

skriss (approved):

Issue #4499, opened 2022-04-26
gateway provisioner: when using a NodePortService, Gateway's port numbers should define nodeport numbers


skriss:
**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
- Installed the provisioner: 
  ```bash
  kubectl apply -f examples/gateway-provisioner/
  ```
- Created a GatewayClass with the envoy network publishing type set to `NodePortService`:
  ```yaml
  ---
  kind: GatewayClass
  apiVersion: gateway.networking.k8s.io/v1alpha2
  metadata:
    name: contour
  spec:
    controllerName: projectcontour.io/gateway-controller
    parametersRef:
      group: projectcontour.io
      kind: ContourDeployment
      namespace: projectcontour
      name: contour-params
  ---
  kind: ContourDeployment
  apiVersion: projectcontour.io/v1alpha1
  metadata:
    namespace: projectcontour
    name: contour-params
  spec:
    envoy:
      networkPublishing:
        type: NodePortService
  ```
- created a Gateway using that GatewayClass:
  ```yaml
  kind: Gateway
  apiVersion: gateway.networking.k8s.io/v1alpha2
  metadata:
    name: gateway-1
    namespace: projectcontour
  spec:
    gatewayClassName: contour
    listeners:
      - name: http
        protocol: HTTP
        port: 30080
        allowedRoutes:
          namespaces:
            from: All
  ```
- observed that the provisioned Envoy service does *not* have a NodePort value of `30080`, instead it's used for the service port value:
  ```bash
  kubectl -n projectcontour get service/envoy-gateway-1 -o yaml
  ```

**What did you expect to happen:**
When a GatewayClass is parameterized to provision envoy services of type NodePort, I think it's logical that the Gateway's port values should be used for NodePort values.

In this scenario, it probably makes sense to default the service port values to 80/443.

I guess one challenge with this approach is that, because port numbers are required on Gateway listeners, this means you can't get auto-allocated NodePort values, which might be undesirable for some users. There is an open issue upstream (https://github.com/kubernetes-sigs/gateway-api/issues/1061) to allow port numbers to be optional, which would be the best of both worlds here.

Perhaps we should hold changing this until that issue has landed.

skriss:
[...]

PR #4705, opened 2022-09-08
crd/ContourDeployment: Add field 'Ports'

izturn:
add the field `Ports` to `crd/ContourDeployment` to enable set containerPort & nodePort for envoy service 


skriss:
Could you explain what you're trying to accomplish with this change? Need to consider how this relates to Gateway Listeners' concept of Ports, which define the port a client would use to connect to the Gateway.

izturn:
the main reason is to let user can set the `nodePort` number for  ServiceType: NodePort

skriss:
xref https://github.com/projectcontour/contour/issues/4499

skriss:
@izturn ack, will try to revisit this in the coming weeks. If you could add some more detail on your use case, that'd be helpful.

izturn:
ping

izturn:
@skriss @sunjayBhatia ping

izturn:
ping @skriss @sunjayBhatia

skriss:
I'm not crazy about specifying some port info in `ContourDeployment`, because I think it belongs with the Listener logically. Right now, a user can change the Listener definitions on the Gateway, and Contour will update the Service definition accordingly. However, ContourDeployment changes typically do _not_ result in changes to the underlying infra, due to the second paragraph [here](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.GatewayClass). That's just one illustration of why having port info stored in two different places can be problematic.

What if alternately, we did the following:
- we change the [behavior described here](https://github.com/projectcontour/contour/issues/4499#issue-1216076991) today, to set the NodePort value to the Listener's PortNumber value when a NodePortService is requested.
- if we did the above, then we could document that when requesting a NodePort service, it's not possible to get auto-assigned ports due to limitations in the Gateway API spec itself. If this is problematic for folks, then we have some justification to change the API spec upstream.

cc @sunjayBhatia interested in your thoughts here.

skriss:
@izturn I am going to close this PR out and will keep an eye out for a new PR per https://github.com/projectcontour/contour/pull/4705#issuecomment-1371921283, thanks for the patience on this one!

PR #4709, opened 2022-09-09
crd/ContourDeployment: Add field 'ComponentLabels"

izturn:
add the field `ComponentLabels` to crd/ContourDeployment to enable add labels for `Contour` components


internal/provisioner/model/names.go
```diff
@@ -82,17 +82,19 @@ func (c *Contour) EnvoyRBACNames() RBACNames {
...
-		"app.kubernetes.io/managed-by": "contour-gateway-provisioner",
+		"app.kubernetes.io/instance": c.Name,
 	}
 
 	// Add owner labels
 	for k, v := range OwnerLabels(c) {
 		labels[k] = v
 	}
 
+	// Add user-defined labels
```
sunjayBhatia, line 93: this should probably come after the contour-defined owner labels so the owner ones override things set by the user, in case they conflict

unless of course we see a compelling reason to let a user override these labels

a unit test for this would be great to add

internal/provisioner/model/names.go
```diff
@@ -82,17 +82,19 @@ func (c *Contour) EnvoyRBACNames() RBACNames {
 // ComponentLabels returns labels for a Contour component.
 func (c *Contour) ComponentLabels() map[string]string {
 	labels := map[string]string{
-		"app.kubernetes.io/name":       "contour",
-		"app.kubernetes.io/instance":   c.Name,
-		"app.kubernetes.io/component":  "ingress-controller",
-		"app.kubernetes.io/managed-by": "contour-gateway-provisioner",
+		"app.kubernetes.io/instance": c.Name,
```
sunjayBhatia, line 85: any particular reason to remove the `name` or `managed-by` labels as things that come by default? I can understand changing the value of the `component` label to something else

sunjayBhatia (changes requested):
I think its reasonable to have a set of labels that will be applied to the managed components for the provisioner, I just wonder if we might end up needing individual labels per-resource or if this will be enough for most people?

Also a couple changes:
- unit testing of label application, particularly making sure the precedence order of user created vs. implementation-specified works properly
- a mention in the release notes what labels are created by the provisioner, might want to even expand the release note level to "minor" to provide more context there

internal/provisioner/model/names.go
izturn: I changed from `componentLabels` to  `commonLabels`, with my use-case, I must filter out the resources which created by myself, not only by `provisioner`,  TBH, I don't know whether  it is a common demand or not

internal/provisioner/model/names.go
izturn: this is a wrong change, i reverted it

internal/provisioner/model/names.go
izturn: i will adjust the codes after you and @skriss  and someone else, think it's a good idea to add this atrribute

izturn:
hey @sunjayBhatia @skriss, let's discuss moving forward #4724 from this pr?

apis/projectcontour/v1alpha1/contourdeployment.go
```diff
@@ -66,6 +66,10 @@ type ContourDeploymentSpec struct {
 	//
 	// +optional
 	RuntimeSettings *ContourConfigurationSpec `json:"runtimeSettings,omitempty"`
+
+	// CommonLabels is a set of labels to add to the provisioned Contour resources.
+	// +optional
+	CommonLabels map[string]string `json:"commonLabels,omitempty"`
```
skriss, line 72: I'm thinking maybe `ResourceLabels` would be a slightly more descriptive name, @sunjayBhatia WDYT?

internal/provisioner/model/model.go
```diff
@@ -343,6 +347,8 @@ type EnvoyNetworkPublishing struct {
 	ServiceAnnotations map[string]string
 }
 
+//
+
```
skriss, line 351: nit: can rm this empty comment/whitespace.

internal/provisioner/objects/secret/secret.go
```diff
@@ -57,11 +57,11 @@ func EnsureXDSSecrets(ctx context.Context, cli client.Client, contour *model.Con
 	}
 
 	for _, secret := range secrets {
-		// Add owner labels.
+		// Add owner & user-definied labels.
```
skriss, line 60:
```suggestion
		// Add owner & user-defined labels.
```

internal/provisioner/objects/dataplane/dataplane.go
```diff
@@ -421,7 +421,7 @@ func desiredDeployment(contour *model.Contour, contourImage, envoyImage string)
 						"prometheus.io/port":   "8002",
 						"prometheus.io/path":   "/stats/prometheus",
 					},
-					Labels: EnvoyPodSelector(contour).MatchLabels,
+					Labels: envoyPodLabels(contour),
```
skriss, line 424: What's the reasoning for this change to add the component labels to the pod selector? The pod selector should already be unique per Gateway, not sure this is necessary.

internal/provisioner/model/names.go
```diff
@@ -82,17 +82,33 @@ func (c *Contour) EnvoyRBACNames() RBACNames {
 // ComponentLabels returns labels for a Contour component.
 func (c *Contour) ComponentLabels() map[string]string {
```
skriss, line 83: I'd suggest a rename to `AppLabels`, along with an update to the godoc indicating that this is specifically for the Deployment/DaemonSet, to help differentiate from the below func.

internal/provisioner/model/names.go
```diff
@@ -82,17 +82,33 @@ func (c *Contour) EnvoyRBACNames() RBACNames {
...
-		"app.kubernetes.io/name":       "contour",
 		"app.kubernetes.io/instance":   c.Name,
+		"app.kubernetes.io/name":       "contour",
 		"app.kubernetes.io/component":  "ingress-controller",
 		"app.kubernetes.io/managed-by": "contour-gateway-provisioner",
 	}
 
+	for k, v := range CommonLabels(c) {
+		labels[k] = v
+	}
```
skriss, line 93: Agree with @sunjayBhatia's previous comment that we probably don't want users to be able to overwrite the above labels, so I'd start with the user-provided labels, then add the app.kubernetes.io ones.

skriss (changes requested):
I think this one is pretty reasonable to get in, did have some comments inline though, and looks like you'll need to rebase / rerun `make generate`.

apis/projectcontour/v1alpha1/contourdeployment.go
sunjayBhatia: yep, i like that name 👍🏽

internal/provisioner/objects/dataplane/dataplane.go
izturn: because when working with `helm`, it will auto add the labels(can't be removed)
```
  "app.kubernetes.io/name":       "xxx",
  "app.kubernetes.io/instance":   "yyy",
```
 to the service's `selector`, if not add the two labels to pod(s), the endpoint will be empty

internal/provisioner/model/model.go
izturn: done

apis/projectcontour/v1alpha1/contourdeployment.go
izturn: done

internal/provisioner/objects/secret/secret.go
izturn: done

internal/provisioner/model/names.go
izturn: done

internal/provisioner/model/names.go
izturn: done

internal/provisioner/model/names.go
```diff
@@ -79,20 +79,35 @@ func (c *Contour) EnvoyRBACNames() RBACNames {
...
 		labels[k] = v
 	}
 
+	// Add user-defined labels
+	for k, v := range c.Spec.ResourceLabels {
+		labels[k] = v
+	}
+
 	return labels
 }
```
skriss, line 112: I'd switch the order of these too, so user-provided labels are unable to overwrite the generated labels:

```suggestion
func CommonLabels(c *Contour) map[string]string {
	labels := map[string]string{}

	// Add user-defined labels
	for k, v := range c.Spec.ResourceLabels {
		labels[k] = v
	}

	// Add owner labels
	for k, v := range OwnerLabels(c) {
		labels[k] = v
	}

	return labels
}
```

skriss:
@sunjayBhatia if you don't have any further comments we can merge this as-is and I can do a quick follow-up on my above comment.

internal/provisioner/objects/dataplane/dataplane_test.go
```diff
@@ -154,6 +154,9 @@ func checkDaemonSecurityContext(t *testing.T, ds *appsv1.DaemonSet) {
 func TestDesiredDaemonSet(t *testing.T) {
 	name := "ds-test"
 	cntr := model.Default(fmt.Sprintf("%s-ns", name), name)
+	cntr.Spec.ResourceLabels = map[string]string{
```
sunjayBhatia, line 157: we can do this in a follow up, but I don't think this line is really testing anything: https://github.com/projectcontour/contour/blob/6fc0daecce9df626085a92872791440889a0e887/internal/provisioner/objects/deployment/deployment_test.go#L143, its just checking if the labels on the resource match against itself...

we should probably write out explicitly what labels we expect

for this and the deployment test

sunjayBhatia:
> @sunjayBhatia if you don't have any further comments we can merge this as-is and I can do a quick follow-up on my above comment.

yep. sounds good, just added one more comment on the existing test, but thats not actually an issue with this PR

skriss (approved):
LGTM, will follow up on two minor things. Thanks @izturn!

skriss:
Oops, missed that there's a merge conflict. @izturn if you can resolve that, then we can get this in.

internal/provisioner/model/names.go
izturn: done

Issue #4666, opened 2022-08-11
Need contour to support regex for CORS- alloworigin field


sri-rajar:
Contour alloworigin field for CORS policy is not supporting pattern matching on domains (like *.vmware.com) .

Envoy has ability to provide regex  support https://www.envoyproxy.io/docs/envoy/latest/api-v3/type/matcher/v3/string.proto#envoy-v3-api-msg-type-matcher-v3-stringmatcher  with safe_regex  option

The ask is for contour to support regex pattern matching for the allowOrigin field in CORS policy.

skriss:
Tentatively adding to 1.23 milestone.

sunjayBhatia:
This looks doable, other than piping config through the only other significant piece of work is to ensure we properly validate regex patterns passed in by users and differentiate them from specific origins

since the `allowOrigin` field is a string slice, not a structured type, we have to be mindful of that, and *maybe* have to deprecate this field and introduce a more structured type for this field if necessary, but hopefully we can do it in-line

sunjayBhatia:
@sri-rajar check out this draft pr: https://github.com/projectcontour/contour/pull/4710 for an early preview!

PR #4710, opened 2022-09-09
CORSPolicy AllowOrigin can be configured as a regex

sunjayBhatia:
This will allow more flexible CORS Allowed Origins with a regex pattern, rather than having to allow all (w/ "*") or enumerating all exact matches.

We attempt to parse each allow origin match as an Origin header (using go url parsing) to check if we get a scheme://host[:port]. If the string doesnt parse as a URL or the format is invalid for an Origin header exact match, then we attempt to parse it as a regex.

Fixes: https://github.com/projectcontour/contour/issues/4666

### DIFF
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -293,27 +293,32 @@ type CORSHeaderValue string
 // CORSPolicy allows setting the CORS policy
 type CORSPolicy struct {
 	// Specifies whether the resource allows credentials.
-	//  +optional
+	// +optional
 	AllowCredentials bool `json:"allowCredentials,omitempty"`
 	// AllowOrigin specifies the origins that will be allowed to do CORS requests. "*" means
 	// allow any origin.
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
 	AllowOrigin []string `json:"allowOrigin"`
 	// AllowMethods specifies the content for the *access-control-allow-methods* header.
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
 	AllowMethods []CORSHeaderValue `json:"allowMethods"`
 	// AllowHeaders specifies the content for the *access-control-allow-headers* header.
-	//  +optional
+	// +optional
+	// +kubebuilder:validation:MinItems=1
 	AllowHeaders []CORSHeaderValue `json:"allowHeaders,omitempty"`
 	// ExposeHeaders Specifies the content for the *access-control-expose-headers* header.
-	//  +optional
+	// +optional
+	// +kubebuilder:validation:MinItems=1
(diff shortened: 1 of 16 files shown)


sunjayBhatia:
Still TODO: add a status test that covers the CORS bits, since that doesnt already exist

sunjayBhatia:
I have a e2e test I used to verify things, not quite decided on whether I want to include it or not, since an e2e test is really mostly testing the Envoy filter, which we don't need to do

apis/projectcontour/v1/httpproxy.go
```diff
@@ -293,27 +293,32 @@ type CORSHeaderValue string
 // CORSPolicy allows setting the CORS policy
 type CORSPolicy struct {
 	// Specifies whether the resource allows credentials.
-	//  +optional
+	// +optional
 	AllowCredentials bool `json:"allowCredentials,omitempty"`
 	// AllowOrigin specifies the origins that will be allowed to do CORS requests. "*" means
 	// allow any origin.
```
skriss, line 299: Probably worth adding some more godoc here on the input formats that are valid.

changelogs/unreleased/4710-sunjayBhatia-minor.md
```diff
@@ -0,0 +1,4 @@
+## HTTPProxy CORS policy supports regex matching on Allowed Origins
+
+The AllowedOrigin field of the HTTPProxy CORSPolicy can be configured as a regex to enable more flexibility for users.
```
skriss, line 3:
```suggestion
The AllowOrigin field of the HTTPProxy CORSPolicy can be configured as a regex to enable more flexibility for users.
```

internal/dag/httpproxy_processor_test.go
```diff
@@ -131,3 +135,160 @@ func TestEnforceRoute(t *testing.T) {
...
+					{Type: CORSAllowOriginMatchRegex, Value: "http://baz.com?query1=2"},
+					{Type: CORSAllowOriginMatchRegex, Value: "http://example.org#fragment"},
+				},
+				ExposeHeaders: []string{},
+			},
+		},
+		"allow origin regex valid": {
+			cp: &contour_api_v1.CORSPolicy{
+				AllowMethods: []contour_api_v1.CORSHeaderValue{"GET"},
+				AllowOrigin:  []string{`.*\.foo\.com`, `https://example\.bar-[0-9]+\.org`},
```
skriss, line 219: I'd say the format for wildcard domain matching in particular should be documented as an example in the godoc for AllowOrigin, since folks may think they can just put `*.foo.com` which is not quite right.

skriss (changes requested):
This approach makes sense to me and gives users a lot of flexibility here. Just a couple minor comments. Other action item would be to add some words/examples to the existing documentation: https://projectcontour.io/docs/v1.22.1/config/cors/

skriss (approved):
:+1:

Issue #4636, opened 2022-07-25
Contour support for Envoy's JWT validation filter


fcalvet:
Envoy has a filter to perform client authorization through a JSON Web Token: [jwt_authn](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/jwt_authn_filter)

Support for this was previously asked for in https://github.com/projectcontour/contour/issues/748, implementation of Envoy's ext_authz filter [in Contour](https://projectcontour.io/docs/main/config/client-authorization/) was preferred.

While this allows checking JWT, it requires to maintain an external service to do so, while Envoy's JWT filter allows to do so without any additional service.

skriss:
Tentatively adding to 1.23 milestone, at least for some PoC/design work.

PR #4711, opened 2022-09-09
design: add JWT verification design doc

skriss:
Updates #4636.

### DIFF
--- a/design/jwt-verification-design.md
+++ b/design/jwt-verification-design.md
@@ -1,0 +1,205 @@
+# JWT Verification Support
+
+## Abstract
+This document describes a design for performing JSON Web Token (JWT) verification for requests to virtual hosts hosted by Contour.
+
+## Background
+JSON Web Token (JWT) is an open standard (RFC 7519) that defines a compact and self-contained way for securely transmitting information between parties as a JSON object. 
+This information can be verified and trusted because it is digitally signed. (ref. https://jwt.io/introduction)
+
+Envoy Proxy has built-in support for verifying JWTs that are attached to incoming requests, via the [JWT Authentication HTTP filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/jwt_authn_filter#config-http-filters-jwt-authn). (Note, this document will use the term **JWT verification** to describe this process).
+Specifically, Envoy can verify the signature, audience, issuer and time restrictions of a JWT.
+If verification fails, the request will be rejected.
+It's important to note that this filter does not itself obtain a JWT for an incoming request; the request must already have one attached.
+
+Contour does not currently have support for configuring the JWT authentication filter.
+This document proposes a design for adding that support to Contour's custom resource, `HTTPProxy`.
+
+## Goals
+- JWT verification for requests to TLS-enabled virtual hosts.
+- Expose a subset of the Envoy filter configuration to cover the most common use cases.
+- Be able to easily expose additional configuration settings if/when needed in the future.
+
+## Non Goals
+- JWT verification for requests to non-TLS enabled virtual hosts.
+- Exposing all possible Envoy configuration settings.
+- Supporting end-to-end OAuth2/OIDC flows.
+
+## High-Level Design
+Contour's `HTTPProxy` resource will get a new optional field, `spec.virtualhost.jwtProviders`, to define the details of how to verify JWTs for requests to a given virtual host.
+This field will only be supported for virtual hosts for which Envoy is terminating TLS.
+The structure of this field will be similar to the [Envoy filter's providers field](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#envoy-v3-api-msg-extensions-filters-http-jwt-authn-v3-jwtauthentication), with some simplifications. 
+Specifically, a provider will define an issuer, 0+ audiences, and a JSON Web Key Set (JWKS) that can be used to verify a JWT (see [the Envoy documentation](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#envoy-v3-api-msg-extensions-filters-http-jwt-authn-v3-jwtprovider) for more information).
+Any number of providers can be defined for a virtual host, to allow different routes to be verified differently.
+
+`HTTPProxy` routes will also get a new optional field, `spec.routes.jwtProvider`, to name a provider defined in the virtual host that should be used to verify JWTs on requests handled by the route.
+If this field is absent, JWTs will not be verified for requests handled by the route.
+
+Contour will validate the contents of `spec.virtualhost.jwtProviders` and `spec.routes.jwtProvider` if present, and will configure the JWT authentication filter on the HTTP Connection Manager for the relevant virtual host.
+Contour will also add a CDS cluster for the remote JWKS, as required by [the Envoy configuration](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#envoy-v3-api-msg-extensions-filters-http-jwt-authn-v3-remotejwks).
+
+
+## Detailed Design
+The detailed structure of the new fields is shown via YAML below:
+
+```yaml
+apiVersion: projectcontour.io/v1
+kind: HTTPProxy
+metadata:
+  name: jwt-verification
+spec:
+  virtualhost:
+    fqdn: example.com
+    tls:
+      secretName: tls-cert
+    jwtProviders:
+      - 
+        # name is a unique name for the provider.
+        name: provider-1
+        # issuer (optional) must match the "iss" field in the JWT.
+        # If not specified, the "iss" field is not checked.
+        issuer: foo.com
+        # audiences (optional) allowlist for the "aud" field in the JWT.
+        # If not specified, the "aud" field is not checked.
+        audiences:
+          - audience-1
+          - audience-2
+        # remoteJWKS is an HTTP endpoint that returns the JWKS
+        # to use to verify the JWT signature.
+        remoteJWKS:
+          httpURI:
+            uri: https://example.com/jwks.json
+            timeout: 1s
+          # cacheDuration is how long to cache the fetched JWKS
+          # locally.
+          cacheDuration: 5m
+        # localJWKS can be used instead of remoteJWKS and defines
+        # an in-cluster secret containing the JWKS for this provider.
+        localJWKS:
+          secretName: my-jwks
+          key: jwks.json
+  routes:
+    # This route specifies jwtProvider: provider-1, which requires requests
+    # to have a JWT has issuer=foo.com, audience of either "audience-1" or 
+    # "audience-2", and signature must be able to be verified using the JWKS
+    # at https://example.com/jwks.json).
+    - conditions:
+        - prefix: /
+      jwtProvider: provider-1
+      services:
+        - name: s1
+          port: 80
+    # This route does *not* have a jwtProvider defined, so requests
+    # with paths starting with /js are excluded from JWT verification.
+    # Note that Contour orders routes such that longer prefixes take
+    # priority over shorter prefixes, so requests starting with /js will
+    # be handled by this route rather than the previous catch-all route.
+    - conditions:
+        - prefix: /js
+      services:
+        - name: s1
+          port: 80
+```
+
+It is worth highlighting some of the configuration options that Envoy's filter has, that Contour will *not* expose, at least not initially:
+- non-default extract locations for the JWT (the default locations are (1) the `Authorization` header using the Bearer schema; and (2) the `access_token` query parameter, in that order).
+- complex rule requirements (e.g. `requires_any`, `requires_all`)
+- rule route matches are limited to what Contour already supports for routing (path prefix and some header matching)
+
+Contour's API is structured such that these and other more complex/less common options may be added at a later date, if there is user demand for them.
+
+A complete/valid HTTPProxy using JWT verification is shown below:
+
+```yaml
+apiVersion: projectcontour.io/v1
+kind: HTTPProxy
+metadata:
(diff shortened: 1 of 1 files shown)


design/jwt-verification-design.md
```diff
@@ -0,0 +1,205 @@
...
+    - conditions:
+        - prefix: /js
+      services:
+        - name: s1
+          port: 80
+```
+
+It is worth highlighting some of the configuration options that Envoy's filter has, that Contour will *not* expose, at least not initially:
+- non-default extract locations for the JWT (the default locations are (1) the `Authorization` header using the Bearer schema; and (2) the `access_token` query parameter, in that order).
+- complex rule requirements (e.g. `requires_any`, `requires_all`)
```
tsaarni, line 106: Just a question to clarify to myself: we will use simple requirement rules such as below?

```yaml
- match:
    prefix: /
  requires: { provider_name: provider-A }
```

design/jwt-verification-design.md
```diff
@@ -0,0 +1,205 @@
...
+        - prefix: /
+      services:
+        - name: s1
+          port: 80
+```
+
+This design has the potential benefit of reducing duplication in the routing rules, by not requiring an entire route to be duplicated in order to express an "except" condition for JWT verification.
+It also allows the user to have direct control over the order in which JWT rules are applied, versus being subject to Contour's rules for sorting routes.
+
+However, this design results in two separate places where route-related behavior are defined, and also creates cognitive overhead for users by having two separate ways of ordering match criteria.
```
tsaarni, line 201: Another question: I have not studied `jwt_authn` filter at this level before. Are these two API alternatives also reflecting the alternatives where we can add `jwt_authn` filter in the Envoy config? 

I did not see full example in Envoy docs but I guess it could be added in various places? It could be added to `http_filters` of `HttpConnectionManager` , or to `typed_per_filter_config` of `config.route.v3.VirtualHost `, or `typed_per_filter_config` of `config.route.v3.Route`

Requirements uses `RouteMatch` to apply verification, similarly as `Route` applies them for routing requests. Does this imply following: if `jwt_authn` filter is added inside `config.route.v3.Route`, which itself already has matched the URL for routing, the match requirement inside  `jwt_authn` does not matter much anymore?  The requirement could simply match `/` since route level matching has already happened before we reach `jwt_authn` filter . On the other hand, if `jwt_authn` is used inside virtualhost (before routing), it requires duplicating all match rules that need to be also present in virtualhost's Routes?

tsaarni (commented):
Great work and glad to see this feature coming! I added few questions.

design/jwt-verification-design.md
```diff
@@ -0,0 +1,205 @@
...
+- Be able to easily expose additional configuration settings if/when needed in the future.
+
+## Non Goals
+- JWT verification for requests to non-TLS enabled virtual hosts.
+- Exposing all possible Envoy configuration settings.
+- Supporting end-to-end OAuth2/OIDC flows.
+
+## High-Level Design
+Contour's `HTTPProxy` resource will get a new optional field, `spec.virtualhost.jwtProviders`, to define the details of how to verify JWTs for requests to a given virtual host.
+This field will only be supported for virtual hosts for which Envoy is terminating TLS.
```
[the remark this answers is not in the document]
sunjayBhatia: Mostly because there is only one filter chain for all plain HTTP virtualhosts, so only one filter instance and no "straightforward" way to configure different providers for different vhosts

(if we want to offer this, maybe we could combine the providers for different vhosts all together in the same filter configuration, but make their names unique etc.)

design/jwt-verification-design.md
```diff
@@ -0,0 +1,205 @@
...
+The structure of this field will be similar to the [Envoy filter's providers field](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#envoy-v3-api-msg-extensions-filters-http-jwt-authn-v3-jwtauthentication), with some simplifications. 
+Specifically, a provider will define an issuer, 0+ audiences, and a JSON Web Key Set (JWKS) that can be used to verify a JWT (see [the Envoy documentation](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#envoy-v3-api-msg-extensions-filters-http-jwt-authn-v3-jwtprovider) for more information).
+Any number of providers can be defined for a virtual host, to allow different routes to be verified differently.
+
+`HTTPProxy` routes will also get a new optional field, `spec.routes.jwtProvider`, to name a provider defined in the virtual host that should be used to verify JWTs on requests handled by the route.
```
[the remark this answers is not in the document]
skriss: I think that's accurate, I can add a sentence or two about it. Do you see any issues with that model?

design/jwt-verification-design.md
```diff
@@ -0,0 +1,205 @@
...
+If this field is absent, JWTs will not be verified for requests handled by the route.
+
+Contour will validate the contents of `spec.virtualhost.jwtProviders` and `spec.routes.jwtProvider` if present, and will configure the JWT authentication filter on the HTTP Connection Manager for the relevant virtual host.
+Contour will also add a CDS cluster for the remote JWKS, as required by [the Envoy configuration](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#envoy-v3-api-msg-extensions-filters-http-jwt-authn-v3-remotejwks).
```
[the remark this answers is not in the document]
skriss: Yes, that's a good point. I was actually considering whether or not it'd make sense to make use of an `ExtensionService` here, which already has fields for defining a number of these cluster properties (including upstream validation). I don't think it actually ends up being a great fit here, but we may want to take some of the properties defined for an `ExtensionService` and add them here as optional settings.

design/jwt-verification-design.md
```diff
@@ -0,0 +1,205 @@
...
+        remoteJWKS:
+          httpURI:
+            uri: https://example.com/jwks.json
+            timeout: 1s
+          # cacheDuration is how long to cache the fetched JWKS
+          # locally.
+          cacheDuration: 5m
+        # localJWKS can be used instead of remoteJWKS and defines
+        # an in-cluster secret containing the JWKS for this provider.
+        localJWKS:
```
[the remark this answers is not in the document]
skriss: :+1: Let me call that out more explicitly.

design/jwt-verification-design.md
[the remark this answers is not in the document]
tsaarni: Seems good and useful behavior to me! Just asking since it was not mentioned :smiley:

design/jwt-verification-design.md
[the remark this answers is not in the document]
skriss: Yeah it's really for both of those reasons. I do think we _could_ make it work for plain HTTP -- if we add an `:authority` header match for each requirement, then we should be able to scope things for the right vhosts - but we'd have to be careful to ensure no pollution across vhosts. My preference would be to not implement it unless there's a strong use case for it.

design/jwt-verification-design.md
skriss: Yep, that's the plan.

design/jwt-verification-design.md
```diff
@@ -0,0 +1,240 @@
...
+        - name: s1
+          port: 80
+```
+
+This design has the potential benefit of reducing duplication in the routing rules, by not requiring an entire route to be duplicated in order to express an "except" condition for JWT verification.
+It also allows the user to have direct control over the order in which JWT rules are applied, versus being subject to Contour's rules for sorting routes.
+
+However, this design results in two separate places where route-related behavior are defined, and also creates cognitive overhead for users by having two separate ways of ordering match criteria.
+
+### Options for working with HTTPProxy Inclusion
```
skriss, line 211: Added a few different options for how this feature interacts with inclusion here, input welcome.

design/jwt-verification-design.md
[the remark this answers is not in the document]
skriss: Spent some more time thinking about this and wrote up a few different options: https://github.com/projectcontour/contour/pull/4711/files#r968997784.

design/jwt-verification-design.md
sunjayBhatia: > On the other hand, if jwt_authn is used inside virtualhost (before routing), it requires duplicating all match rules that need to be also present in virtualhost's Routes?

yep, exactly, which is why IMO reducing duplication and keeping the opt-in/out of JWT verification for a Route is a bit cleaner, we can reduce what users need to configure and simply have them use the existing route matching rules

skriss: Yeah, this is a good point, having the JWT providers specified on the routing rules themselves should actually simplify how we configure Envoy since we can put the configuration in the routes themselves in per-route config. I need to play around with it some since it's not how I initially implemented it, and the docs are not super-detailed, but definitely a good though to pursue.

sunjayBhatia: yeah the per-route config seems like where you can set what requirement you want to be verified (by name), but the definition of the requirements/providers still has to go in the filter

design/jwt-verification-design.md
sunjayBhatia: hm, so the tradeoff with these seems like single jwt provider vs. root propagating the jwt requirement

would it be too complicated to do some hybrid where you can still configure multiple providers but have a setting (bool flag in the provider type for example) where you can specify one to propagate down to children (the bool would only be set to true in one provider in the list)

you could also have the opt in/specific provider field per-route to override this

might be then useful to offer the ability to configure multiple jwt requirements on a route? but could be overkill

sunjayBhatia (commented):
I'm a fan of the proposed structure around inclusion since it leaves the defaulting in the hands of the owner of the root httpproxy, owners of included child httpproxies can still opt out, or be restricted by gatekeeper policies, and still offers the flexibility of allowing multiple jwt providers to be available

I think that and using route matching to configure jwt verification opt in/out work great together

anything outstanding left to sort out? i think theres coverage of validating the jwks endpoint w/ a CA etc., and the remaining discussions are around possible additional features we may add? (one small thing there is if someone were running an authorization server in-cluster they may want to use a service to generate the Envoy cluster, so that could be a note for a later feature to add)

design/jwt-verification-design.md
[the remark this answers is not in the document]
skriss: I've now made a revision to the design to implement this a bit differently -- namely, one JWT provider can optionally be tagged as the "default", in which case it auto-applies to all routes in the root proxy and any child proxies. Individual routes can then override the default provider, or opt out of JWT verification. If no provider is set as the default, then routes can opt into verification by explicitly specifying a provider. I think this covers all of the use cases and applies consistently for single HTTPProxies as well as parent/child proxies using includes.

design/jwt-verification-design.md
skriss: I made this switch in my WIP branch and it seems to be working well. The other nice benefit it has is it avoids us needing to separately sort a list of JWT requirements -- now the sorting happens as part of the normal route sorting we do. Helps avoid some ugly code duplication in our sorter package.

design/jwt-verification-design.md
skriss: I've updated the doc with a revision along the lines of Sunjay's above comment.

skriss:
> anything outstanding left to sort out? i think theres coverage of validating the jwks endpoint w/ a CA etc., and the remaining discussions are around possible additional features we may add? (one small thing there is if someone were running an authorization server in-cluster they may want to use a service to generate the Envoy cluster, so that could be a note for a later feature to add)

I added some fields to JWT provider for JWKS TLS validation. Good point on the in-cluster JWKS endpoint, I think that could be a good follow-on feature to add support for. I've tried to leave room in the API for additions so that we can tackle additional features as needed, just wanted to start with the basics.

sunjayBhatia (approved):

design/jwt-verification-design.md
```diff
@@ -0,0 +1,471 @@
...
+Contour's `HTTPProxy` resource will get a new optional field, `spec.virtualhost.jwtProviders`, to define the details of how to verify JWTs for requests to a given virtual host.
+This field will only be supported for virtual hosts for which Envoy is terminating TLS.
+The structure of this field will be similar to the [Envoy filter's providers field](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#envoy-v3-api-msg-extensions-filters-http-jwt-authn-v3-jwtauthentication), with some simplifications. 
+Specifically, a provider will define an issuer, 0+ audiences, and a JSON Web Key Set (JWKS) that can be used to verify a JWT (see [the Envoy documentation](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#envoy-v3-api-msg-extensions-filters-http-jwt-authn-v3-jwtprovider) for more information).
+Any number of providers can be defined for a virtual host, to allow different routes to be verified differently.
+At most one provider can be marked as the "default", meaning it will automatically be applied as a requirement to all routes unless they explicitly opt out (more below).
```
tsaarni, line 34: What do you think if instead of separate field `default: true` we would pick the default provider by name `name: default`. It would avoid the error case where more than one providers have `default: true` set. However, the provider named `default` then becomes "magically" applied by default, which of course could be considered as bit surprising by users? I'm unsure myself. Maybe putting a special meaning to a name is a bad idea, but I wanted to hear what you think?

skriss: Yeah, I've definitely used that pattern before for similar situations -- it does have the advantage of not requiring a separate flag which is nice. When I used it before, it was for the names of CR's themselves, which meant that the API server would prevent multiple resources from being created with the name "default"; in this case, though, a user could name multiple JWT providers "default" so we'd still need code to check for that situation & set an error condition. I think since we'd have to do that runtime check regardless, I probably lean towards keeping the flag as I think it is a little less "magical", but open to others' thoughts here too.

skriss:
Looks like this is ready to merge; any further feedback can always be addressed with the implementation.

PR #4720, opened 2022-09-13
Update Gateway YAML generation for extracted namespace file

### LINKED ISSUE #4519: Deploying Gateway Provisioner with kustomize fails with accumulation error
Rather than deploy by hand with `kubectl apply -f https://projectcontour.io/quickstart/contour-gateway-provisioner.yaml` I took it the next step to a committable version using kustomize, which looks very similar:

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
  - https://raw.githubusercontent.com/projectcontour/contour/release-1.21/examples/render/contour-gateway-provisioner.yaml
```

(Note that I tried both with and without the fancy url above, the have the same result, and I'd prefer using something with explicit versioning).
This generally should work in the same way but at times kustomize is a little more rigid about what it accepts. In this case it complains that it gets two attempts to create the namespace `gateway-api`. On looking further, sure enough it exists twice in the [rendered file](https://github.com/projectcontour/contour/blob/3fae0d635fffb323063e40d7c2278ce2d16c61e9/examples/render/contour-gateway-provisioner.yaml) and given the [build file](https://github.com/projectcontour/contour/blob/3fae0d635fffb323063e40d7c2278ce2d16c61e9/hack/generate-provisioner-deployment.sh#L17-L18) that loads https://github.com/projectcontour/contour/blob/3fae0d635fffb323063e40d7c2278ce2d16c61e9/examples/gateway/01-admission_webhook.yaml and https://github.com/projectcontour/contour/blob/3fae0d635fffb323063e40d7c2278ce2d16c61e9/examples/gateway/02-certificate_config.yaml each of which attempt to create that namespace. 

As a result kustomize complains

```
error: accumulating resources: accumulation err='accumulating resources from 'https://raw.githubusercontent.com/projectcontour/contour/release-1.21/examples/render/contour-gateway-provisioner.yaml': may not add resource with an already registered id: ~G_v1_Namespace|~X|gateway-api': git cmd = '/opt/homebrew/bin/git fetch --depth=1 origin HEAD': exit status 128
```

which can be understood to mean what I said above.

kustomize would install in the same way as apply.

Changing one or both of those files to not include that extraneous namespace creation would fix it, as would adding a render step to remove one (though that is a bit overkill I'd think). Simply having only one step that builds the namespace should be sufficient, though perhaps it could be its own file before the two offenders.

**Environment:**

- Contour version: 1.21
```
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.2", GitCommit:"8b5a19147530eaac9476b0ab82980b4088bbc1b2", GitTreeState:"clean", BuildDate:"2021-09-15T21:31:32Z", GoVersion:"go1.16.9", Compiler:"gc", Platform:"darwin/arm64"}
Server Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.8+rke2r1", GitCommit:"7061dbbf75f9f82e8ab21f9be7e8ffcaae8e0d44", GitTreeState:"clean", BuildDate:"2022-03-16T22:51:09Z", GoVersion:"go1.16.15b7", Compiler:"gc", Platform:"linux/amd64"}
```
- Kubernetes installer & version: rke2 (something recent)
- Cloud provider or hardware configuration: 4-node metal
- OS (e.g. from `/etc/os-release`): server os Ubuntu, client macos

sunjayBhatia:
Closes #4519.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -237,6 +237,7 @@ generate-gateway-yaml:
 	@echo "Generating Gateway API CRD YAML documents..."
 	@kubectl kustomize -o examples/gateway/00-crds.yaml "github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=${GATEWAY_API_VERSION}"
 	@echo "Generating Gateway API webhook documents..."
+	@curl -s -o examples/gateway/00-namespace.yaml https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/${GATEWAY_API_VERSION}/config/webhook/0-namespace.yaml
 	@curl -s -o examples/gateway/01-admission_webhook.yaml https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/${GATEWAY_API_VERSION}/config/webhook/admission_webhook.yaml
 	@curl -s -o examples/gateway/02-certificate_config.yaml https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/${GATEWAY_API_VERSION}/config/webhook/certificate_config.yaml
 	
--- a/examples/gateway/00-namespace.yaml
+++ b/examples/gateway/00-namespace.yaml
@@ -1,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: gateway-system
--- a/examples/gateway/01-admission_webhook.yaml
+++ b/examples/gateway/01-admission_webhook.yaml
@@ -1,8 +1,3 @@
-apiVersion: v1
-kind: Namespace
-metadata:
-  name: gateway-system
----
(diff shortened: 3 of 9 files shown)


skriss (approved):
LGTM

skriss:
Filed https://github.com/projectcontour/contour/issues/4721 to look at simplifying some of this code.

skriss:
Looks like needs to be added to https://github.com/projectcontour/contour/blob/main/test/scripts/make-kind-cluster.sh#L122-L126 as well.

skriss:
I'm going to merge as-is with the test flake/timeout since this passed most of the upgrade/E2E matrix runs.

PR #4723, opened 2022-09-14
JWT verification

### LINKED ISSUE #4636: Contour support for Envoy's JWT validation filter

skriss:
Closes #4636.


apis/projectcontour/v1/httpproxy.go
```diff
@@ -244,6 +244,64 @@ type VirtualHost struct {
...
+	// Remote JWKS to use for verifying JWT signatures.
+	// +kubebuilder:validation:Required
+	RemoteJWKS RemoteJWKS `json:"remoteJWKS"`
+}
+
+// RemoteJWKS defines how to fetch a JWKS from an HTTP endpoint.
+type RemoteJWKS struct {
+	// The URI and associated info for the JWKS.
+	// +kubebuilder:validation:Required
+	HTTPURI HTTPURI `json:"httpURI"`
```
skriss, line 287: We could just inline the two fields that HTTPURI has (`URI` and `Timeout`) here to simplify the structure a little -- I probably lean towards that, thoughts?

internal/dag/dag.go
```diff
@@ -636,6 +640,41 @@ type SecureVirtualHost struct {
...
+
+type JWTProvider struct {
+	Name       string
+	Issuer     string
+	Audiences  []string
+	RemoteJWKS RemoteJWKS
+}
+
+type RemoteJWKS struct {
+	HTTPURI       HTTPURI
```
skriss, line 656: can do the same inlining as with the API here, the nesting is not useful for us

apis/projectcontour/v1/httpproxy.go
```diff
@@ -244,6 +244,64 @@ type VirtualHost struct {
...
+	// +optional
+	Audiences []string `json:"audiences,omitempty"`
+
+	// Remote JWKS to use for verifying JWT signatures.
+	// +kubebuilder:validation:Required
+	RemoteJWKS RemoteJWKS `json:"remoteJWKS"`
+}
+
+// RemoteJWKS defines how to fetch a JWKS from an HTTP endpoint.
+type RemoteJWKS struct {
```
skriss, line 284: TLS validation fields & impl still need to be added, I was thinking I'd do that in a follow-up PR.

skriss:
Marked this ready for review as the code should be in pretty good shape, still need to do docs and ensure there are no follow-ups based on design doc discussions.

apis/projectcontour/v1/httpproxy.go
sunjayBhatia: yep, that works for me

internal/dag/dag.go
sunjayBhatia: 👍🏽

apis/projectcontour/v1/httpproxy.go
```diff
@@ -391,6 +449,27 @@ type Route struct {
...
+}
+
+type JWTVerificationPolicy struct {
+	// Require names a specific JWT provider (defined in the virtual host)
+	// to require for the route. If specified, this field overrides the
+	// default provider if one exists. If this field is not specified,
+	// the default provider will be required if one exists. At most one of
+	// this field or the "disabled" field can be specified.
+	// +optional
+	Require string `json:"require,omitempty"`
```
sunjayBhatia, line 465: thinking out loud: we could pull this out into a struct to make it easier in the future to add more a advanced requirement pattern, but i think that would still maybe require deprecating fields so probably not worth it, just a small thing to think about again in case anyone finds it useful

internal/envoy/v3/cluster.go
```diff
@@ -162,6 +162,48 @@ func ExtensionCluster(ext *dag.ExtensionCluster) *envoy_cluster_v3.Cluster {
...
+	switch c.Scheme {
+	case "http":
+		port = 80
+	case "https":
+		port = 443
+		// TODO TLS validation
+		transportSocket = UpstreamTLSTransportSocket(UpstreamTLSContext(nil, c.Address, nil))
+	}
+
+	cluster.LoadAssignment = &envoy_endpoint_v3.ClusterLoadAssignment{
```
sunjayBhatia, line 186: nit: sort of duplicated with the `StaticClusterLoadAssignment` helper, which does take a dag.Service rather than a host/port but maybe could be de-duped?

sunjayBhatia: also could maybe use the `Endpoints` helper?

apis/projectcontour/v1/httpproxy.go
skriss: I had been thinking that the `JWTVerificationPolicy` struct itself could add `RequiresAll`, `RequiresAny`, etc. fields down the road if needed, with only one of them being allowed to be specified but yeah we could also encapsulate all of those into a separate struct

internal/envoy/v3/cluster.go
skriss: Yeah, good point, let me take another look at this -- initially I didn't want to start changing the existing helpers since they were pretty tied to the dag.Service, but there's probably some opportunity for reuse.

internal/dag/httpproxy_processor.go
```diff
@@ -331,6 +343,78 @@ func (p *HTTPProxyProcessor) computeHTTPProxy(proxy *contour_api_v1.HTTPProxy) {
...
+					Issuer:    jwtProvider.Issuer,
+					Audiences: jwtProvider.Audiences,
+					RemoteJWKS: RemoteJWKS{
+						URI:     jwtProvider.RemoteJWKS.URI,
+						Timeout: jwksTimeout,
+						Cluster: DNSNameCluster{
+							Address:         jwksURL.Hostname(),
+							Scheme:          jwksURL.Scheme,
+							DNSLookupFamily: string(p.DNSLookupFamily),
+						},
```
tsaarni, line 413: Shouldn't we allow user to set non-default port number for jwksURL? That is, should we add Port field that can override default 80 and 443?

internal/envoy/v3/cluster.go
skriss: Take a look at https://github.com/projectcontour/contour/pull/4723/commits/1e1810cb37630d1806c12ef5ea1af94afe7096c5, does some refactoring across this area.

internal/dag/httpproxy_processor.go
skriss: Yep, we should definitely support that.

skriss: Implemented in https://github.com/projectcontour/contour/pull/4723/commits/479c8099f0dd76d756ac31c0d44b3d8dd8d0a4e1

skriss:
@sunjayBhatia ready for another look here when you get a chance.

internal/envoy/v3/route.go
```diff
@@ -151,16 +164,25 @@ func routeAuthzContext(settings map[string]string) *any.Any {
...
+func PathRouteMatch(pathMatchCondition dag.MatchCondition) *envoy_route_v3.RouteMatch {
+	switch c := pathMatchCondition.(type) {
 	case *dag.RegexMatchCondition:
 		return &envoy_route_v3.RouteMatch{
 			PathSpecifier: &envoy_route_v3.RouteMatch_SafeRegex{
 				// Add an anchor since we at the very least have a / as a string literal prefix.
 				// Reduces regex program size so Envoy doesn't reject long prefix matches.
 				SafeRegex: SafeRegexMatch("^" + c.Regex),
 			},
-			Headers:         headerMatcher(route.HeaderMatchConditions),
```
sunjayBhatia, line 162: nice!

sunjayBhatia (approved):
LGTM, great doc page and test coverage!

internal/envoy/v3/route.go
skriss: :+1: turns out this wasn't actually necessary since the rules moved to inline in the routes, but it's a good cleanup anyway so will leave it in place.

PR #4729, opened 2022-09-19
docs: fix miss

kahirokunn:


skriss:
@kahirokunn I am fine with the first commit, but I'd rather not touch all these design docs just to remove whitespace. Could you drop the whitespace commit from the PR? Thanks.

skriss:
Actually, in re-reviewing the first commit, I don't think we need to change this doc - it's part of the historical record of how we were thinking about things at the time. If you're looking for more up-to-date documenation, you can see the more recent design doc: https://github.com/projectcontour/contour/blob/main/design/automated-provisioning-design.md

And also up-to-date/versioned API documentation for e.g. ContourDeployment: https://projectcontour.io/docs/v1.22.1/config/api/#projectcontour.io/v1alpha1.ContourDeployment

PR #4735, opened 2022-09-20
Set grpc fixture imagepullpolicy to ifnotpresent

sunjayBhatia:
A shot in the dark as to why CI randomly fails w/ gRPC error code unimplemented


skriss (approved):
Worth a shot. As discussed, other options could be adding code to wait for the deployment to be up and ready before attempting gRPC requests, and/or adding the "Unimplemented" code to the list of retryable codes.

PR #4741, opened 2022-09-21
support TLS validation for JWKS servers

### LINKED ISSUE #4733: add upstream TLS validation to JWT verification JWKS cluster
TLS validation fields & impl still need to be added, I was thinking I'd do that in a follow-up PR.

_Originally posted by @skriss in https://github.com/projectcontour/contour/pull/4723#discussion_r973093233_

skriss:
Adds TLS server cert validation for JWKS
servers as part of JWT verification.

Closes #4733.


site/content/docs/main/config/jwt-verification.md
```diff
@@ -78,6 +77,13 @@ spec:
 In the above example, the default route requires requests to carry JWTs that can be verified using provider-1.
 The second route _excludes_ requests to paths starting with `/css` from JWT verification, because it does not have a JWT verification policy.
 
+### Configuring TLS validation for the JWKS server
```
skriss, line 80: Copied this content mostly verbatim from the same subject on the upstream-tls.md page.

internal/dag/status_test.go
```diff
@@ -3356,6 +3356,251 @@ func TestDAGStatus(t *testing.T) {
...
+					"RemoteJWKSUpstreamValidationInvalid",
+					"Spec.VirtualHost.JWTProviders.RemoteJWKS.UpstreamValidation is invalid: invalid CA Secret \"roots/nonexistent\": Secret not found",
+				),
+		},
+	})
+
+	jwksInvalidCACert := &v1.Secret{
+		ObjectMeta: fixture.ObjectMeta("roots/cacert"),
+		Data: map[string][]byte{
+			"wrong-key": []byte(fixture.CERTIFICATE),
```
sunjayBhatia, line 3477: i guess this case causes the secret to never be added to the cache, which results in a not found error similar to the above? I'm not sure if we can even test something distinct to that, since we do so much checking of secret validity when adding to the cache

the check for a valid CA when looking it up from the cache is almost moot it seems

might be ok to remove this test case

skriss: Yep, I realized the same when I implemented this test case. I figured we may as well leave it so it's covered in case of future changes in this area.

sunjayBhatia (approved):

PR #4744, opened 2022-09-22
internal/contour: ignore status-only Gateway API updates

skriss:
Ignores status-only updates on Gateway API
resources when determining whether to
re-process the DAG. This avoids unnecessary
processing and xDS traffic.


sunjayBhatia (approved):
good catch! didnt realize we did have this set up already

PR #4745, opened 2022-09-22
internal/k8s: check Gateway API resource status equality

skriss:
Adds all Gateway API resources with status to
the isStatusEqual func, which helps avoid
extraneous status update calls to the apiserver
when status has not changed.


skriss:
Follow-on to #4744, another place that filters out no-ops that we forgot to add all Gateway API types to.

sunjayBhatia (approved):

PR #4749, opened 2022-09-23
design: Contour utilizing Gateway API Policy Attachment and Route Filters

sunjayBhatia:
Design to outline how Contour development around GW API Policy Attachment and Route Filters should proceed

Attempting to generate guidelines/patterns/etc. we can agree on before starting this work

Will be accompanied with a spike example

### DIFF
--- a/design/gw-api-policy-and-filter-design.md
+++ b/design/gw-api-policy-and-filter-design.md
@@ -15,11 +15,17 @@ We've outlined the feature parity gap between Contour's HTTPProxy and the existi
 Gateway API does offer a few extension points intended to support many of the features Contour and other ingress controllers provide:
 - [Policy Attachment](https://gateway-api.sigs.k8s.io/references/policy-attachment/)
 - [HTTPRoute filters](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteFilter)
+- Using a different type than Service in a `BackendRef`
 - [GatewayClass parameters](https://gateway-api.sigs.k8s.io/api-types/gatewayclass/#gatewayclass-parameters)
 
 These configuration points are intended for different classes of configuration.
-We will mainly focus on the first two above, though the third is worth mentioning as an option.
-The Contour Gateway Provisioner uses GatewayClass parameters for passing general configuration for an instance of Contour.
+We will mainly focus on the first two above, though the next two are worth mentioning as options.
+
+The Contour Gateway Provisioner uses GatewayClass parameters for passing general configuration for an instance of Contour and is not really a focus of this design.
+
+It is also possible for us to use a different type than Service as the target of a `BackendRef` [`BackendObjectReference`](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.BackendObjectReference).
+For example, this may be a "wrapper" type around a Service that offers additional configuration Service does not have.
+The current [Backend Capabilities GEP](https://github.com/kubernetes-sigs/gateway-api/pull/1430) seems to be using this method to configuring things like TLS between a proxy and backend (which is a bit out of scope of this design, but possible relevant).
 
 Policy attachment is described in detail [here](https://gateway-api.sigs.k8s.io/references/policy-attachment/).
 Policy resources must follow a standard structure, referencing the resource they apply to via a [`targetRef`](https://gateway-api.sigs.k8s.io/references/policy-attachment/#target-reference-api) field.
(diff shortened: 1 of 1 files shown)


sunjayBhatia:
WIP for now, just to save it and get some eyes on it to start if anyone is interested

design/gw-api-policy-and-filter-design.md
```diff
@@ -0,0 +1,126 @@
...
+  - These will hopefully be covered by [this GEP](https://gateway-api.sigs.k8s.io/geps/gep-1282/) and subsequent features
+- Contour flags we've added for smaller surface area coverage or convention etc.
+  - e.g. `contourconfiguration.spec.envoy.listener.disableAllowChunkedLength`
+
+## High-Level Design
+Summary TBD
+
+## Detailed Design
+
+### Categorizing features between Policies or Filters
```
skriss, line 56: Some useful quotes from the spec:

> Filters define processing steps that must be completed during the request or response lifecycle. Filters act as an extension point to express additional processing that may be performed in Gateway implementations. Some examples include request or response modification, implementing authentication strategies, rate-limiting, and traffic shaping.

(ref. https://gateway-api.sigs.k8s.io/api-types/httproute/#filters-optional)

> Custom Route filters provide a way to configure request/response modifiers or middleware embedded inside Route rules or backend references.

> Policy attachment is more broad in scope. In contrast with filters, policies can be attached to a wide variety of Gateway API resources, and include a concept of hierarchical defaulting and overrides. Although Policy attachment can be used to target an entire Route or Backend, it cannot currently be used to target specific Route rules or backend references. If there are sufficient use cases for this, policy attachment may be expanded in the future to support this fine grained targeting.

(ref. https://gateway-api.sigs.k8s.io/references/policy-attachment/#interaction-with-custom-route-filters)

design/gw-api-policy-and-filter-design.md
```diff
@@ -0,0 +1,126 @@
...
+- direct HTTP response configuration (HTTPProxy directResponsePolicy, could also be contributed to GW API upstream?)
+
+Things not in Contour yet:
+- tracing
+- IP allow/block lists
+- etc.
+
+We can categorize these a few ways, but to differentiate what should be configured in a Gateway API Policy vs. Filter, I propose that we consider two categories:
+
+#### Features that modify content in the data path
```
skriss, line 72: Probably another way to think about this is: configuration that is likely to be specific to an individual route, and unlikely to generally apply across an entire Listener/Gateway. I think both the examples below fit this well.

design/gw-api-policy-and-filter-design.md
```diff
@@ -0,0 +1,126 @@
...
+
+## High-Level Design
+Summary TBD
+
+## Detailed Design
+
+### Categorizing features between Policies or Filters
+Here is a list of features we want to configure likely outside scope of the Gateway API standardized features that are already in Contour:
+- rate limiting
+- auth (external auth and/or built-in Envoy filters)
```
skriss, line 59: Rate limiting and auth are interesting -- although they're specifically called out in the description for filters, and I can easily see them being implemented as filters, they feel like things where we'd want to be able to apply defaults at the Gateway/Listener level, and also be able to override at the route level. I guess that pushes them towards Policy Attachment.

design/gw-api-policy-and-filter-design.md
skriss, line 56: Even just looking at the built-in filters -- e.g. the request header modifier -- in Contour/HTTPProxy, we support both global default header modifications, and per-route modifications. So to implement that in Gateway API, you'd need to use policy attachment, except there's also the built-in filter, and having both contradicts the guidance in https://gateway-api.sigs.k8s.io/references/policy-attachment/#2-custom-filters-and-policies-should-not-overlap. 🤔

sunjayBhatia: yep, that and what you mention with rate limiting + auth are what im struggling with a little, though there is a guideline about separating concerns, it doesnt seem like it is really that cut and dry with these examples

maybe something to bring up upstream

skriss:
Local rate limiting could be a good simple one to spike both ways, if we had something functional that we could show it might help with an upstream discussion

sunjayBhatia:
some discussion at the end of this about a "decision tree" or similar about when to use policy attachment etc.: https://www.youtube.com/watch?v=WMns9-_D9o0&list=PL69nYSiGNLP1GgO7k02ipPGZUFpSzGaHH&index=40&ab_channel=Kubernetes

In the context of this design around backend app/protocol details: https://docs.google.com/document/d/1qWy18JijbYZw70AS9Gt7AcL2O07p8FQWlkHs5agl99c/edit#heading=h.5p4hft7tu75r

design/gw-api-policy-and-filter-design.md
```diff
@@ -0,0 +1,238 @@
...
+- HTTP header policy
+- TLS parameters
+- Envoy "Cluster" settings
+- Envoy "Listener" settings
+- etc.
+
+## Security Considerations
+
+### Cross-namespace references
+It is not specified explicitly in the Gateway API documentation, but if Policy or Filter custom resources live in a separate namespace from the resource it references/is referenced by, we will likely need to ensure the appropriate ReferenceGrant is present that allows the resources to be "attached."
```
danehans, line 210: [ExtensionRef](https://github.com/kubernetes-sigs/gateway-api/blob/v0.5.1/apis/v1beta1/httproute_types.go#L601-L609) is a LocalObjectRef so the ref'd resource should be in the same ns.

sunjayBhatia: ah yep, good catch, should be corrected to match the `Rate Limiting: Filter` section which mentions this

PR #4752, opened 2022-09-26
Add schedule for staling issues

sunjayBhatia:
Stale after 60 days, closed after 90

Exempted a few labels from being staled as well.

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -14,12 +14,16 @@ jobs:
     - uses: actions/stale@v4
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
+        days-before-pr-stale: 14
+        days-before-pr-close: 30
+        stale-pr-label: 'lifecycle/stale'
         stale-pr-message: >
           Marking this PR stale since there has been no activity for 14
-          days. It will be closed if there is no activity for another
-          30 days.
-        stale-pr-label: 'lifecycle/stale'
-        days-before-stale: 14
-        days-before-close: 30
-        days-before-issue-stale: -1 # disable marking issues as stale
-        days-before-issue-close: -1 # disable closing stale issues
+          days. It will be closed if there is no activity for 30 days.
+        days-before-issue-stale: 60
+        days-before-issue-close: 90
+        stale-issue-label: 'lifecycle/stale'
+        exempt-issue-labels: 'help wanted,good first issue'
+        stale-issue-message: >
+          Marking this Issue stale since there has been no activity for 60
+          days. It will be closed if there is no activity for 90 days.


.github/workflows/stale.yaml
```diff
@@ -14,12 +14,16 @@ jobs:
...
-        stale-pr-label: 'lifecycle/stale'
-        days-before-stale: 14
-        days-before-close: 30
-        days-before-issue-stale: -1 # disable marking issues as stale
-        days-before-issue-close: -1 # disable closing stale issues
+          days. It will be closed if there is no activity for 30 days.
+        days-before-issue-stale: 60
+        days-before-issue-close: 90
+        stale-issue-label: 'lifecycle/stale'
+        exempt-issue-labels: 'help wanted,good first issue'
```
sunjayBhatia, line 26: also considered adding the `blocked,blocked/*` labels as well as possibly the `priority/important-longterm` label, thoughts?

sunjayBhatia: we're gonna use `priority/important-soon` for things "in the backlog" that we think we should do in the next couple releases

going to audit that label and add it to the exempt category

.github/workflows/stale.yaml
```diff
@@ -14,12 +14,16 @@ jobs:
...
         stale-pr-message: >
           Marking this PR stale since there has been no activity for 14
-          days. It will be closed if there is no activity for another
-          30 days.
-        stale-pr-label: 'lifecycle/stale'
-        days-before-stale: 14
-        days-before-close: 30
-        days-before-issue-stale: -1 # disable marking issues as stale
-        days-before-issue-close: -1 # disable closing stale issues
+          days. It will be closed if there is no activity for 30 days.
```
skriss, line 22: I actually think the previous message was correct, see e.g. https://github.com/projectcontour/contour/pull/4603 which was marked stale 26 days ago but hasn't been closed yet. Seems like the "close" counter starts when the issue/PR is labeled stale.

.github/workflows/stale.yaml
```diff
@@ -14,12 +14,16 @@ jobs:
...
+        days-before-issue-stale: 60
+        days-before-issue-close: 90
```
skriss, line 24: In light of the above comment, probably makes sense to set this to 30 (so 60 days to stale, then another 30 to close)?

skriss (changes requested):

.github/workflows/stale.yaml
sunjayBhatia: oh oops i totally screwed up the math there whoops, will put back and fix the increments

sunjayBhatia:
modified the project board to move existing `priority/important-soon` items to `priority/important-longterm`, planned things on the project board now have `priority/important-soon`

.github/workflows/stale.yaml
```diff
@@ -14,12 +14,18 @@ jobs:
...
-        days-before-issue-stale: -1 # disable marking issues as stale
-        days-before-issue-close: -1 # disable closing stale issues
+        days-before-issue-stale: 60
+        days-before-issue-close: 30
+        stale-issue-label: 'lifecycle/stale'
+        exempt-issue-labels: 'help wanted,good first issue,priority/important-soon'
+        stale-issue-message: >
+          Marking this Issue stale since there has been no activity for 60
+          days. It will be closed if there is no activity for another
+          30 days.
```
skriss, line 31: I'm thinking it'd be helpful to expand this message a bit, to be more like what upstream k8s projects use (e.g. https://github.com/kubernetes-sigs/gateway-api/issues/1180#issuecomment-1250439469):

- explain why we have the stale bot
- explain what the stale bot does (this is covered already)
- explain what actions the user can take (comment, close, contribute)

I think the extra bit of context is a little more friendly and gives the user specific steps they can take if they want.

skriss (approved):
LGTM, thanks for updating the messages.

I'm in favor of doing this, along with better triage of issues as they're opened, to better manage our backlog and reflect to our users what our priorities are. Having a smaller/more focused backlog will be more useful for both maintainers and users, while not preventing anyone from "helping themselves" if they have a specific issue they want to contribute on.

PR #4753, opened 2022-09-27
internal/provisioner: fix current & expect ports len not equal bug

izturn:
when  `len(range current.Spec.Ports)` > `len(expected.Spec.Ports[i])` it will crashed
```
	for i, p := range current.Spec.Ports {
		if !apiequality.Semantic.DeepEqual(p, expected.Spec.Ports[i]) {
			updated.Spec.Ports = expected.Spec.Ports
			changed = true
		}
	}
```


internal/provisioner/equality/equality.go
```diff
@@ -194,6 +194,7 @@ func NodePortServiceChanged(current, expected *corev1.Service) (*corev1.Service,
 	if len(current.Spec.Ports) != len(expected.Spec.Ports) {
 		updated.Spec.Ports = expected.Spec.Ports
 		changed = true
+		return updated, changed
```
sunjayBhatia, line 197: could you add a test that covers this?

sunjayBhatia (changes requested):

internal/provisioner/equality/equality.go
izturn: done

internal/provisioner/equality/equality_test.go
```diff
@@ -531,6 +531,13 @@ func TestNodePortServiceChanged(t *testing.T) {
 			},
 			expect: true,
 		},
+		{
+			description: "if the ports number changed",
```
sunjayBhatia, line 535: just a tiny nit on the description
```suggestion
			description: "if the number of ports changed",
```

sunjayBhatia (commented):
just a nit on naming/the branching logic and needs a rebase or main merge and it looks good to go

internal/provisioner/equality/equality.go
```diff
@@ -194,6 +194,7 @@ func NodePortServiceChanged(current, expected *corev1.Service) (*corev1.Service,
...
 	}
 
 	for i, p := range current.Spec.Ports {
```
sunjayBhatia, line 200: to fit into the structure of this function, with only 1 location for a return, might be better to do an `else` case for the check for ranging through the list of ports, rather than return early

internal/provisioner/equality/equality_test.go
```diff
@@ -531,6 +531,13 @@ func TestNodePortServiceChanged(t *testing.T) {
...
+			mutate: func(svc *corev1.Service) {
+				svc.Spec.Ports = svc.Spec.Ports[:len(svc.Spec.Ports)-1]
```
sunjayBhatia, line 537: also I think this test passes without any change to the implementation

you need to append to this list to get it to fail as you describe

sunjayBhatia (changes requested):
wait actually i misread this test

internal/provisioner/equality/equality_test.go
izturn: done

izturn:
the wrong signoff, closed and recreate as #4760

PR #4760, opened 2022-09-29
 internal/provisioner: fix current & expect number of  ports changed bug

izturn:
recreate of #4753 

### DIFF
--- a/internal/provisioner/equality/equality.go
+++ b/internal/provisioner/equality/equality.go
@@ -194,38 +194,38 @@ func NodePortServiceChanged(current, expected *corev1.Service) (*corev1.Service,
 	if len(current.Spec.Ports) != len(expected.Spec.Ports) {
 		updated.Spec.Ports = expected.Spec.Ports
 		changed = true
-	}
+	} else {
+		for i, p := range current.Spec.Ports {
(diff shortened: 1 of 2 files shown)


izturn:
@sunjayBhatia  PTAL

internal/provisioner/equality/equality.go
```diff
@@ -194,38 +194,38 @@ func NodePortServiceChanged(current, expected *corev1.Service) (*corev1.Service,
...
 		updated.Spec.Ports = expected.Spec.Ports
 		changed = true
-	}
+	} else {
+		for i, p := range current.Spec.Ports {
+			if !apiequality.Semantic.DeepEqual(p, expected.Spec.Ports[i]) {
+				updated.Spec.Ports = expected.Spec.Ports
+				changed = true
+			}
+		}
```
skriss, line 203: The `else` block should end here -- we should only do the port-by-port comparison if current and expected have the same number of ports, but then all the subsequent comparisons of spec fields should be done regardless.

izturn: thx, my fault

sunjayBhatia (approved):

Issue #4715, opened 2022-09-12
Gateway API: fix route sorting


skriss:
Contour is currently failing the `main` conformance test defined for route sorting: https://github.com/kubernetes-sigs/gateway-api/pull/1370

(Note, there is an issue filed to ensure that the spec is unambiguous: https://github.com/kubernetes-sigs/gateway-api/issues/1381)

We need to track the outcome of this and ensure Contour implements the correct behavior.

Tentatively adding to 1.23 for tracking.

skriss:
@sunjayBhatia's PR to clarify this upstream has been merged: https://github.com/kubernetes-sigs/gateway-api/pull/1401

So looks like now we need to fix Contour to pass the conformance test!

sunjayBhatia:
For context, this comment/function is useful: https://github.com/projectcontour/contour/blob/5af0a4263d73bd24bb2dabf163c7c0e39814de4a/internal/xdscache/v3/route.go#L136-L154

also route sorter: https://github.com/projectcontour/contour/blob/5af0a4263d73bd24bb2dabf163c7c0e39814de4a/internal/sorter/sorter.go#L147

route sorter helper for comparing header + query matches: https://github.com/projectcontour/contour/blob/5af0a4263d73bd24bb2dabf163c7c0e39814de4a/internal/sorter/sorter.go#L113

we first sort the header matches within each dag.Route (should probably do the same for query matches)

then we sort the dag.Routes, which has a precedence first for the path match type and length of path, then orders them based on number of header and query matches

header matches here are then compared between routes, if theyre the same length the longer one wins, if the same, theyre compared w/ the header name alphabetically

sunjayBhatia:
in the failing conformance test, we have two route matches that could apply to the request, both with the same path and number of header matches so the first one in the list should win

however, in contour currently the one with the header name that sorts first wins instead

sunjayBhatia:
a couple ideas come to mind to fix this:
- add an option to dag.Route, used only in routes generated by HTTPRoute processing, to signify we will not sort on header name alphabetical order to retain passed in order
- just never sort routes based on header name alphabetical order, and retain the passed in order from the HTTPRoute/HTTPProxy etc.
- add an option to dag.Route that would add a priority field, could sort on path match+length, then header length, then query length, then priority, and so-on

there is also a question of how to sort things that may overlap between *different HTTPRoutes* but that is another issue, to do with resource precedence (age, etc.)

sunjayBhatia:
[...]

PR #4762, opened 2022-09-29
internal/sorter: Routes w/ query param matches are sorted based on query param name

sunjayBhatia:
To ensure stable ordering of the xDS config sent to Envoy

Added test will fail sometimes w/o this change

Updates: #4715


skriss (approved):
LGTM, yeah agree no relnote needed.

PR #4763, opened 2022-09-29
Fix Gateway API route sorting when conflicts within an HTTPProxy

sunjayBhatia:
This change adds a Priority field to dag.Route so that routes that come from different HTTPRouteRules within an HTTPRoute can be differentiated so that the first one in the list gets precedence.

Fixes: #4715

Passes Gateway API conformance on main

~~Still need to add more unit tests to the dag package, but figured this would serve as a good point to start to review, to see if we like the approach~~

~~Has the side effect that if an HTTPRoute and HTTPProxy (or Ingress) generate a conflicting dag.Route, the ~~HTTPProxy~~ HTTPRoute one should always win~~

We have a separate concern about conflicts between HTTPRoutes that also needs a conformance test upstream and some more consideration in Contour.


internal/dag/gatewayapi_processor.go
```diff
@@ -1154,15 +1155,23 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1beta1.HTTPRou
 			}
 		}
 
+		// Priority is used to ensure if there are multiple matching route rules
+		// within an HTTPRoute, the one that comes first in the list has
+		// precedence. We subtract the index of the rule from MaxUint8 to ensure
+		// rules that come first have a higher priority. All dag.Routes generated
+		// from this HTTPRoute rule matches have the same priority.
+		priority := math.MaxUint8 - uint8(ruleIndex)
```
sunjayBhatia, line 1163: an alternative that would mean we don't have to change a bunch of tests would be to say a lower value means a higher priority

this would then also mean the side-effect of HTTPRoute generated dag.Routes don't automatically get more precedence than conflicting HTTPProxy generated ones

internal/sorter/sorter.go
```diff
@@ -116,6 +116,17 @@ func longestRouteByHeaderAndQueryParamConditions(lhs, rhs *dag.Route) bool {
 		return len(lhs.HeaderMatchConditions) > len(rhs.HeaderMatchConditions)
 	}
 
+	// One route has a longer QueryParamMatchConditions slice.
```
sunjayBhatia, line 119: moving this up to accommodate the GW API spec also changes sort order a little since length of both header and query param matches is checked before comparing item by item

internal/dag/gatewayapi_processor.go
sunjayBhatia: (i havent changed the dag package unit tests yet for this reason)

internal/dag/dag.go
```diff
@@ -228,6 +228,10 @@ type Route struct {
 	// match on the querystring parameters.
 	QueryParamMatchConditions []QueryParamMatchCondition
 
+	// Priority specifies the relative priority of the Route when compared to other
+	// Routes that may have equivalent match conditions.
+	Priority uint8
```
skriss, line 233: Worth adding a note indicating that a higher number means it will sort first.

internal/sorter/sorter.go
skriss: should be fine, we only support query param matches via Gateway API anyway.

skriss (commented):
Approach LGTM. I have a feeling that if you're trying to mix Gateway API/HTTPRoute and HTTPProxy, you're probably in for some unexpected behavior no matter what, so I'm not too concerned about exactly how conflicts between them are sorted.

skriss:
> Has the side effect that if an HTTPRoute and HTTPProxy (or Ingress) generate a conflicting dag.Route, the HTTPProxy one should always win

Isn't it the opposite? HTTPRoute ones always win because they have a non-zero priority?

sunjayBhatia:
> > Has the side effect that if an HTTPRoute and HTTPProxy (or Ingress) generate a conflicting dag.Route, the HTTPProxy one should always win
> 
> Isn't it the opposite? HTTPRoute ones always win because they have a non-zero priority?

ha, yeah I think i just mixed up the names there, HTTPRoute should always win

internal/dag/gatewayapi_processor.go
skriss: Yeah I think this works fine, and a lot fewer tests to have to add a non-zero priority to

sunjayBhatia: changed to the inverse, lower number means higher priority

sunjayBhatia:
@skriss you think this needs a release note?

skriss:
> @skriss you think this needs a release note? i guess it is user facing but the intent was mostly for conformance, idk!

Probably worth a note just to say that we now sort Gateway API routes in a conformant way.

skriss (approved):
LGTM pending relnote, nice work!

Issue #4579, opened 2022-06-14
remove Contour E2E's that overlap with Gateway API conformance


skriss:
As upstream Gateway API conformance tests are built out, we can probably get rid of overlapping Contour E2E's that validate the same functionality. This will save some CI time and reduce maintenance burden. As a starting point, once we upgrade to Gateway API v0.5.0 and pin to the conformance tests included there, we should do an audit of Contour's E2E's and drop any redundant ones.

sunjayBhatia:
unstaling, this seems worthwhile to do

sunjayBhatia:
some removals going in here: https://github.com/projectcontour/contour/pull/4765

skriss:
Adding this to the current milestone for tracking, we'll likely carry it forward.

skriss:
Bumping to 1.24.0 for continued work.

skriss:
@sunjayBhatia anything else here for 1.24, or shall we move this over to 1.25 for continued work?

sunjayBhatia:
looks like none of the gateway tests are removable due to overlap with v0.7.1 or v0.8.0-rc1 tags as far as I can tell, can call this done for v1.26.0 IMO

skriss:
I'm going to close this out as we've done most/all of this work by now. Can file specific issues as anything else comes up.

PR #4765, opened 2022-09-29
Remove e2e gateway tests that overlap with Gateway API conformance

sunjayBhatia:
To help w/ flakes in CI and reduce duplication

- query param match test should be covered by
https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/tests/httproute-query-param-matching.go
- header match test should be covered by https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/tests/httproute-header-matching.go
- invalid backend ref test should be covered by
  - https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/tests/httproute-invalid-backendref-unknown-kind.go
  - https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/tests/httproute-invalid-backendref-nonexistent.go
- request header modifier (on route rule) test covered by https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/tests/httproute-request-header-modifier.go
- route parent refs test should be covered by:
  - https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/tests/httproute-simple-same-namespace.go
  - https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/tests/httproute-cross-namespace.go
  - https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/tests/httproute-invalid-cross-namespace-parent-ref.go

Updates: #4579


test/e2e/gateway/query_param_match_test.go
```diff
@@ -1,111 +0,0 @@
...
-					},
-				},
-			},
-		}
-		f.CreateHTTPRouteAndWaitFor(route, httpRouteAccepted)
-
-		cases := map[string]string{
-			"/?animal=whale":                     "echo-1",
-			"/?animal=whale&foo=bar":             "echo-1", // extra irrelevant parameters have no impact
-			"/?animal=whale&animal=dolphin":      "echo-1", // the first occurrence of a given key in a querystring is used for matching
```
sunjayBhatia, line 81: this isn't explicitly covered since it is implementation dependent how to deal with repeated query params

test/e2e/gateway/query_param_match_test.go
```diff
@@ -1,111 +0,0 @@
...
-			"/?animal=dolphin":                   "echo-2",
-			"/?animal=dolphin&animal=whale":      "echo-2", // the first occurrence of a given key in a querystring is used for matching
```
sunjayBhatia, line 83: same here

skriss (commented):
The other thing we could do is change this behavior:

https://github.com/projectcontour/contour/blob/main/test/e2e/gateway/gateway_test.go#L69-L76

We currently run all the tests twice, once with a Contour configured with a GatewayClass controller string, and once with a specific Gateway. I think we could move the "controller string" method to be a daily run instead of a PR-blocking run. The provisioner uses the "specific gateway" method, so better to keep in the PR runs.

sunjayBhatia:
> The other thing we could do is change this behavior:
> 
> https://github.com/projectcontour/contour/blob/main/test/e2e/gateway/gateway_test.go#L69-L76
>
> We currently run all the tests twice, once with a Contour configured with a GatewayClass controller string, and once with a > specific Gateway. I think we could move the "controller string" method to be a daily run instead of a PR-blocking run. The > provisioner uses the "specific gateway" method, so better to keep in the PR runs.

yee thats a good idea, I'll do that in a separate change since its a little bigger

skriss (approved):

PR #4767, opened 2022-09-30
internal/provisioner: use generics for Ensure functions

skriss:
cc @sunjayBhatia, let me know what you think


internal/provisioner/objects/object.go
```diff
@@ -47,16 +47,16 @@ func NewUnprivilegedPodSecurity() *corev1.PodSecurityContext {
...
 // EnsureObjectDeleted ensures that an object with the given namespace and name is deleted
 // if Contour owner labels exist.
-func EnsureObjectDeleted(ctx context.Context, cli client.Client, contour *model.Contour, name string, getter ObjectGetter) error {
-	obj, err := getter(ctx, cli, contour.Namespace, name)
+func EnsureObjectDeleted[T client.Object](
+	ctx context.Context,
+	cli client.Client,
+	contour *model.Contour,
+	name string,
+	getObject func(ctx context.Context, cli client.Client, namespace, name string) (T, error),
```
skriss, line 57: we can probably get rid of passing the `getObject` func in here since all the implementations are the same just issuing a `client.Get(...)` call, will mess around some more.

skriss:
Still experimenting here to see how much we can consolidate down into the EnsureObject/EnsureObjectDeleted funcs, and how much Go generics let us get away with.

izturn:
nice, I had wanted to do this, but I was afraid you all didn't want to use generics

sunjayBhatia (commented):
has a good advantage of being able to write a specific updater etc. func for the type you want to update since that does require some type-specific info, with a generic helper that calls it, the type constraint on the generic function only needs to be for the interface that is used

messed around with this a bit but didnt make any real meaningful changes on top of this, nice!

skriss:
Will get this dusted off and ready for official review.

skriss:
This should be ready for review, I will also self-review again.

skriss:
@sunjayBhatia I re-reviewed as well as did some manual testing, so I think everything's good to go here.

skriss:
rebased

skriss:
@sunjayBhatia no huge rush but it'd be nice to get this one in before too long, thanks in advance!

sunjayBhatia:
> @sunjayBhatia no huge rush but it'd be nice to get this one in before too long, thanks in advance!

looking now! sorry for the delay

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/projectcontour/contour
 
-go 1.16
+go 1.19
```
sunjayBhatia, line 3: CONTRIBUTING.md has a specific ref to Go 1.16, might want to just make that say to look in this file for the minimum Go version

internal/provisioner/objects/object_test.go
```diff
@@ -0,0 +1,281 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	pkgclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
```
sunjayBhatia, line 33: nice tests!

sunjayBhatia (approved):
lgtm, definitely an improvement!

go.mod
skriss: Ah good call

PR #4770, opened 2022-10-03
test/e2e: Gateway tests run in "gateway" reconcile mode by default

sunjayBhatia:
Adds a daily build run that runs them in "controller" mode, and ability to focus the package to run


skriss (approved):
LGTM

skriss:
Unfortunately now looks like our upgrade test is flaking, probably something to look into before 1.23 release in case there's an actual upgrade issue

PR #4772, opened 2022-10-04
Added support for Envoy slow start mode.

### LINKED ISSUE #2296: Progressive traffic increase for new Pods
We have a JVM-based web app behind Contour/Envoy/NLB, with horizontal pod auto scaling in place.
When a new pod gets created due to auto scaling, Contour/Envoy directs a proportional amount of traffic on that new pod. However, because the app is cold, we're seeing consistent **timeouts** until it warms up.

Screenshot 2020-02-28 17 36 47

We tried the same scenario by using a Service type LoadBalancer, in EKS (with an **Elastic Load Balancer** in front) and we don't see the same issue in this scenario. This seems to be because the ELB is doing a progressive traffic increase on the new pod, as the graph seen below.
Screenshot 2020-02-28 17 34 56

Is there any plan to support something similar in Contour? I see we have the possibility to set weights for different services in an IngressRoute. Would it be something to consider to set some weifghts at pod level for a given service, based on their age? (or is something like this available today)?

### LINKED ISSUE #4687: Gradual traffic increase for cold starting endpoints
Some of our backends require a slow start mechanism for Java based new cold pods as they have a special compile on demand approach, where the required logic is compiled on first use. 

For such backends increased connect timeouts and retry policies did not help.
It would be good to optionally use Envoy-s slow start mode for such nodes
more info about the feature:
https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/slow_start.html

This could let new endpoints get a smaller percentage of the traffic at first, then over a set time increase their share to the original percentage.
This could further minimise traffic loss caused by newly appearing slow starting pods.

The feature could be governed on a backend cluster basis in HTTPProxy.

tsaarni:
This PR adds support for slow start mode, described [here](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/slow_start) in Envoy documentation. Envoy to gradually increase the amount of traffic targeted to a newly added upstream endpoint. This can be useful for example with JVM based applications, that might otherwise get overwhelmed during JIT warm-up period.

The feature is configured via struct `HTTPProxy.spec.routes.services.slowStart`:

```yaml
kind: HTTPProxy
metadata:
  name: slow-start
spec:
  virtualhost:
    fqdn: www.example.com
  routes:
    - services:
        - name: java-app
          port: 80
          slowStart:
            window: 3s
            aggression: 1.0
            minWeightPercent: 10
```

Slow start is off by default. To enable it, only `window` is mandatory. Other parameters are optional. 

Fixes #2296
Fixes #4687


tsaarni:
Manual test procedure for this PR is described here https://gist.github.com/tsaarni/a22e3d69bf6c31ff0597e9f78454868a.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -1168,3 +1171,34 @@ type HTTPProxyList struct {
...
+	//
+	// More info: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/slow_start
+	//
+	// +optional
+	// +kubebuilder:default=`1.0`
+	// +kubebuilder:validation:Pattern=`^([0-9]+([.][0-9]+)?|[.][0-9]+)$`
+	Aggression string `json:"aggression"`
+
+	// The minimum percentage of origin weight that avoids too small new weight, which may cause endpoints in slow start mode receive no traffic in slow start window.
+	// If not specified, the default is 10%.
```
skriss, line 1198: Is this basically saying that the endpoint will get at least 10% of traffic, from the very beginning? I know this is copied mostly verbatim from envoy docs, but I think we can probably make it a little more clear, just need to read up a bit more to make sure I'm understanding exactly what it means.

internal/dag/httpproxy_processor.go
```diff
@@ -870,6 +870,13 @@ func (p *HTTPProxyProcessor) computeRoutes(
 				}
 			}
 
+			slowStartConfig, err := slowStartConfig(service.SlowStart)
+			if err != nil {
+				validCond.AddErrorf(contour_api_v1.ConditionTypeServiceError, "SlowStartInvalid",
+					"%s on slow start", err)
+				return nil
+			}
```
skriss, line 878: it'd be nice to add some test cases to status_test.go to cover the error case here and make codecov stop complaining.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -1168,3 +1171,34 @@ type HTTPProxyList struct {
 	metav1.ListMeta `json:"metadata"`
 	Items           []HTTPProxy `json:"items"`
 }
+
+// SlowStart will gradually increase amount of traffic to a newly added endpoint.
+// It works only with RoundRobin and WeightedLeastRequest load balancing strategies.
```
skriss, line 1176: Maybe worth explicitly stating that it will be _ignored_ if using a different load balancing strategy.

skriss (changes requested):
@tsaarni just a couple nits but overall this looks great and pretty straight-forward, thanks! I think we can include this in the upcoming 1.23 release.

apis/projectcontour/v1/httpproxy.go
skriss, line 1198: I'm thinking something like:

```suggestion
	// The minimum or starting percentage of traffic to send to new endpoints. A non-zero value helps avoid a too small initial weight, which may cause endpoints in slow start mode receive no traffic in the slow start window.
	// If not specified, the default is 10%.
```

apis/projectcontour/v1/httpproxy.go
tsaarni: I have added a note that the configuration is ignored for other strategies.  But on second thought, I started to question myself, so I want to ask: should we ignore (like it is done now) or make HTTPProxy invalid?

apis/projectcontour/v1/httpproxy.go
```diff
@@ -811,6 +811,9 @@ type Service struct {
 	// The policies for rewriting Set-Cookie header attributes.
 	// +optional
 	CookieRewritePolicies []CookieRewritePolicy `json:"cookieRewritePolicies,omitempty"`
+	// Slow start will gradually increase amount of traffic to a newly added endpoint.
+	// +optional
+	SlowStart *SlowStart `json:"slowStart,omitempty"`
```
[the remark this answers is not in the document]
tsaarni: Name changed!

internal/dag/httpproxy_processor.go
tsaarni: Test added.  I also changed the printout a little so that the error we surface will not contain too much unnecessary details (which the error string directly from `strconv.ParseFloat()` had).

apis/projectcontour/v1/httpproxy.go
tsaarni: Yes exactly.  This was clearly visible when I tested with only 2 replicas and set 50% as `MinimumWeightPercent` with long 30s `window`: the pod immediately got 50% of the traffic after container restart.

skriss: Looks good, I think this is pretty clear now, thanks!

skriss (approved):
thanks for the updates @tsaarni, LGTM. I'll leave for @sunjayBhatia to take a look if he wants.

internal/envoy/v3/cluster.go
```diff
@@ -312,3 +331,17 @@ func protocolOptions(explicitHTTPVersion HTTPVersionType, idleConnectionTimeout
...
 	}
 }
+
+// slowStartConfig returns the slow start configuration.
+func slowStartConfig(slowStartConfig *dag.SlowStartConfig) *envoy_cluster_v3.Cluster_SlowStartConfig {
+	return &envoy_cluster_v3.Cluster_SlowStartConfig{
+		SlowStartWindow: protobuf.Duration(slowStartConfig.Window),
+		Aggression: &envoy_core_v3.RuntimeDouble{
+			DefaultValue: slowStartConfig.Aggression,
+			RuntimeKey:   "contour.slowstart.aggression",
```
sunjayBhatia, line 341: interesting this is also configurable via RTDS, not sure if we will need to take advantage of that though

site/content/docs/main/config/slow-start.md
```diff
@@ -0,0 +1,38 @@
...
+Slow start mode is a configuration setting that is used to gradually increase the amount of traffic targeted to a newly added upstream endpoint.
+By default, the amount of traffic will increase linearly for the duration of time window set by `window` field, starting from 10% of the target load balancing weight and increasing to 100% gradually.
+The easing function for the traffic increase can be adjusted by setting optional field `aggression`.
+A value above 1.0 results in a more aggressive increase initially, slowing down when nearing the end of the time window.
+Value below 1.0 results in slow initial increase, picking up speed when nearing the end of the time window.
+Optional field `minWeightPercent` can be set to change the minimum percent of target weight.
+It is used to avoid too small new weight, which may cause endpoint to receive no traffic in beginning of the slow start window.
+
+Slow start mode can be useful for example with JVM based applications, that might otherwise get overwhelmed during JIT warm-up period.
+Application might perform worse immediately after pod start or after container restarts.
```
sunjayBhatia, line 12: nit

```suggestion
Such applications may respond to requests slowly or return errors immediately after pod start or after container restarts.
User impact of this behavior can be mitigated by using slow start configuration to gradually increase traffic to recently started service endpoints.
```

sunjayBhatia (approved):
LGTM, just made a little comment on doc wording 👍🏽

internal/envoy/v3/cluster.go
tsaarni: True, I did not see how to take advantage of it either, but it seems that (almost) all `Runtime*` message types have validation rules in `.proto` which makes the `RuntimeKeys` required.  For example in this case [`RuntimeDouble.runtime_key`](https://github.com/envoyproxy/envoy/blob/8f2674fa6a024bf0eee502834b93e6e82f957bd1/api/envoy/config/core/v3/base.proto#L287) has validation rule:

```
string runtime_key = 2 [(validate.rules).string = {min_len: 1}];
```

that causes config to be rejected if runtime key is not given:

```
[2022-10-06 06:03:23.608][1][warning][config] [source/common/config/grpc_subscription_impl.cc:126] 
gRPC config for type.googleapis.com/envoy.config.cluster.v3.Cluster rejected:
Proto constraint validation failed
(ClusterValidationError.RoundRobinLbConfig: embedded message failed validation | 
caused by RoundRobinLbConfigValidationError.SlowStartConfig: embedded message failed validation |
caused by SlowStartConfigValidationError.Aggression: embedded message failed validation |
caused by RuntimeDoubleValidationError.RuntimeKey: value length must be at least 1 characters)
```

So I have assumed these are mandatory, even if we would never use them.

Curiously, there are at least some cases where it is not required, like when we configure ratelimit [here](https://github.com/projectcontour/contour/blob/d42b7f419050a33625adf691e9321c543acedf5f/internal/envoy/v3/ratelimit.go#L49-L54). In this case the message definition is [missing the validation rule](https://github.com/envoyproxy/envoy/blob/8f2674fa6a024bf0eee502834b93e6e82f957bd1/api/envoy/config/core/v3/base.proto#L486-L495). Maybe that is just by mistake, since that is the only `runtime_key` that is missing validation in [`base.proto`](https://github.com/envoyproxy/envoy/blob/main/api/envoy/config/core/v3/base.proto).

site/content/docs/main/config/slow-start.md
tsaarni: Fixed!

tsaarni:
> LGTM, just made a little comment on doc wording 👍🏽

Very good clarification, thank you for the review!

PR #4775, opened 2022-10-05
Gateway API: Spike on RateLimitFilter

sunjayBhatia:
Shows how a HTTPRoute Filter based implementation might look for rate limiting

so far just "local" rate limiting implemented

Example usable via manifest:

```yaml
---
apiVersion: v1
kind: Namespace
metadata:
  name: demo
---
kind: GatewayClass
metadata:
  name: example
spec:
  controllerName: projectcontour.io/projectcontour/contour
---
kind: Gateway
metadata:
  name: example
  namespace: demo
spec:
  gatewayClassName: example
  listeners:
  - name: http
    protocol: HTTP
    port: 80
    allowedRoutes:
      namespaces:
        from: All
---
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: echo
  name: echo
  namespace: demo
spec:
  replicas: 3
  selector:
    matchLabels:
      app: echo
  template:
    metadata:
      labels:
        app: echo
    spec:
      containers:
      - image: gcr.io/k8s-staging-ingressconformance/echoserver:v20210922-cec7cf2
        name: echo
        ports:
        - name: "http"
          containerPort: 3000
---
apiVersion: v1
kind: Service
metadata:
  labels:
    app: echo
  name: echo
  namespace: demo
spec:
  ports:
  - port: 80
    protocol: TCP
    targetPort: 3000
  selector:
    app: echo
  sessionAffinity: None
  type: ClusterIP
---
kind: RateLimitFilter
metadata:
  name: local-ratelimit-example
  namespace: demo
spec:
  local:
    requests: 10
    unit: minute
---
kind: HTTPRoute
metadata:
  name: echo
  namespace: demo
spec:
  parentRefs:
    kind: Gateway
    name: example
    namespace: demo 
  hostnames:
  - "filter-example.com"
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /
    backendRefs:
    - kind: Service
      name: echo
      port: 80
    filters:
    - type: ExtensionRef
      extensionRef:
        group: projectcontour.io
        kind: RateLimitFilter
        name: local-ratelimit-example
```


skriss:
This looks great, nice to see it in action. I guess this one is an easy one to do as a per-HTTPRoute filter, because Envoy actually supports per-route config for it. Nice that you could define a single RateLimitFilter resource, and then reference it from a number of different routes, if you wanted the same rate limits to apply across multiple routes. You could even have an admission controller that automatically adds it to routes, though that kind of behavior starts to overlap with the "Policy" side of things.

Would be interesting to look across the various features that HTTPProxy has and see how many of them could be implemented like this.

Thinking through this some more, it also seems far easier to implement a Filter, than to implement a Policy, since you don't have to implement all of the defaults/overrides across various layers of the resource graph.

PR #4781, opened 2022-10-07
fixes to enable development on arm64

skriss:
- remove default build platform of `linux/amd64` (current versions of Docker should automatically populate this based on the host)
- remove pre-loading of E2E test fixture images into kind (probably not necessary anymore given we're no longer using httpbin, which was huge; the current images are small #'s of MB -- hitting what looks like https://github.com/kubernetes-sigs/kind/issues/2549 which is actually a containerd issue, but things work fine if the images are just pulled normally)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
-ARG BUILDPLATFORM=linux/amd64
 ARG BUILD_BASE_IMAGE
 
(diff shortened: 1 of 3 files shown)


skriss:
Re-running E2E's on a few different dev machines, but at least `make container` seems to be working fine everywhere.

sunjayBhatia (approved):
dont have an arm64 machine to test on but seems reasonable!

Issue #4206, opened 2021-12-01
drop internal/workgroup package


skriss:
looks like we could almost drop the `workgroup` package entirely, but it's still used in the featuretests framework. We could do the work to fully drop it in a separate PR.

_Originally posted by @skriss in https://github.com/projectcontour/contour/pull/4202#discussion_r760353704_

https://github.com/projectcontour/contour/pull/4202#discussion_r760363268

skriss:
Throwing this into 1.24, nice small cleanup item.

PR #4783, opened 2022-10-10
Fix workgroup AddContext bug

chattarajoy:
In case functions are added to workgroup using AddContext method, the AddContext method doesn't exit if the corresponding function exits. This change makes sure the Run flow exits in case the function running has exited.

### DIFF
--- a/internal/workgroup/group.go
+++ b/internal/workgroup/group.go
@@ -40,6 +40,7 @@ func (g *Group) Add(fn func(<-chan struct{}) error) {
 func (g *Group) AddContext(fn func(context.Context) error) {
 	g.fn = append(g.fn, func(stop <-chan struct{}) error {
 		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
 		res := make(chan error)
 
 		// run the function & send the result on res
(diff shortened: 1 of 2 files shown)


skriss:
Thanks for the PR @chattarajoy, however what I would really like to do is drop the workgroup package entirely since it's only being used by test code at this point and I don't think we need our own custom goroutine manager anymore -- see #4206. Would you be interested in taking that work on?

sunjayBhatia (commented):
yep agreed, we should focus on possibly removing the package since it's only used in tests

chattarajoy:
Hi @skriss , since I was using this module in my own project (copied the code, not importing using go mod) I found this bug and fixed it in my project as well as thought of raising a fix here. I am sorry I won't be able to pick up the deletion of this model and remove its usage.

skriss:
> Hi @skriss , since I was using this module in my own project (copied the code, not importing using go mod) I found this bug and fixed it in my project as well as thought of raising a fix here. I am sorry I won't be able to pick up the deletion of this model and remove its usage.

@chattarajoy fair enough - I think we can merge this as-is and still look to do #4206 subsequently.

skriss (approved):

PR #4788, opened 2022-10-12
internal/dag: validate Secrets only when used

### LINKED ISSUE #2616: Misleading Proxy status for secrets type opaque

### LINKED ISSUE #2908: internal: Secrets that aren't referenced from an ingress object are still validated
Any secret that exists in the cluster is validated to make sure it's a good cert to use inside Envoy, however, Contour shouldn't process each secret, but only those that are referenced from any ingress object.

skriss:
Changes Secret validation to happen when Secrets
are used by processors, rather than as they are added
to the DAG cache. Also caches the validation results
on the DAG objects for efficiency.

This eliminates the need for Contour to inspect Secrets
that are not used by an Ingress/HTTPProxy/Gateway. It
also replaces the confusing "Secret not found" errors
with specific validation errors when a Secret does exist,
but is not valid.

This also allows Secret validation logic to be more
precise, since we know what the processors are trying
to use them for.

Fixes #2616.
Fixes #2908.

### DIFF
--- a/internal/dag/cache.go
+++ b/internal/dag/cache.go
@@ -516,18 +516,63 @@ func isRefToSecret(ref gatewayapi_v1beta1.SecretObjectReference, secret *v1.Secr
 		string(ref.Name) == secret.Name
 }
 
-// LookupSecret returns a Secret if present or nil if the underlying kubernetes
-// secret fails validation or is missing.
-func (kc *KubernetesCache) LookupSecret(name types.NamespacedName, validate func(*Secret) error) (*Secret, error) {
+func (kc *KubernetesCache) LookupTLSSecret(name types.NamespacedName) (*Secret, error) {
 	sec, ok := kc.secrets[name]
 	if !ok {
 		return nil, fmt.Errorf("Secret not found")
 	}
 
-	if err := validate(sec); err != nil {
-		return nil, err
+	// Compute and store the validation result if not
+	// already stored.
+	if sec.ValidTLSSecret == nil {
+		sec.ValidTLSSecret = &SecretValidationStatus{
+			Error: validTLSSecret(sec.Object),
+		}
 	}
 
+	if err := sec.ValidTLSSecret.Error; err != nil {
+		return nil, err
+	}
+	return sec, nil
+}
+
+func (kc *KubernetesCache) LookupCASecret(name types.NamespacedName) (*Secret, error) {
+	sec, ok := kc.secrets[name]
+	if !ok {
+		return nil, fmt.Errorf("Secret not found")
+	}
+
+	// Compute and store the validation result if not
+	// already stored.
+	if sec.ValidCASecret == nil {
+		sec.ValidCASecret = &SecretValidationStatus{
+			Error: validCASecret(sec.Object),
+		}
+	}
+
+	if err := sec.ValidCASecret.Error; err != nil {
+		return nil, err
+	}
+	return sec, nil
+}
+
+func (kc *KubernetesCache) LookupCRLSecret(name types.NamespacedName) (*Secret, error) {
+	sec, ok := kc.secrets[name]
+	if !ok {
+		return nil, fmt.Errorf("Secret not found")
+	}
+
+	// Compute and store the validation result if not
+	// already stored.
+	if sec.ValidCRLSecret == nil {
+		sec.ValidCRLSecret = &SecretValidationStatus{
+			Error: validCRLSecret(sec.Object),
+		}
+	}
+
+	if err := sec.ValidCRLSecret.Error; err != nil {
+		return nil, err
+	}
 	return sec, nil
 }
 
@@ -537,7 +582,7 @@ func (kc *KubernetesCache) LookupUpstreamValidation(uv *contour_api_v1.UpstreamV
 		return nil, nil
 	}
 
-	cacert, err := kc.LookupSecret(caCertificate, validCA)
+	cacert, err := kc.LookupCASecret(caCertificate)
 	if err != nil {
 		// UpstreamValidation is requested, but cert is missing or not configured
 		return nil, fmt.Errorf("invalid CA Secret %q: %s", caCertificate, err)
--- a/internal/dag/cache.go
+++ b/internal/dag/cache.go
@@ -103,18 +103,6 @@ func (kc *KubernetesCache) Insert(obj interface{}) bool {
 	maybeInsert := func(obj interface{}) bool {
 		switch obj := obj.(type) {
 		case *v1.Secret:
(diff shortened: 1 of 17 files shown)


skriss:
Still testing this, but putting up as a draft in case anyone wants to check out the approach.

internal/dag/cache.go
```diff
@@ -103,19 +103,9 @@ func (kc *KubernetesCache) Insert(obj interface{}) bool {
...
-						WithField("version", k8s.VersionOf(obj)).
-						Error(err)
-				}
-				return false
-			}
-
-			kc.secrets[k8s.NamespacedNameOf(obj)] = obj
+			// Secret validation status is intentionally cleared, it needs
+			// to be re-validated after an insert.
+			kc.secrets[k8s.NamespacedNameOf(obj)] = &Secret{Object: obj}
```
skriss, line 108: we end up caching more Secrets this way, but we never look at them unless they're actually referenced. I'm not sure it matters all that much memory-wise, since these are all in the controller-runtime cache already anyway, and we're just storing pointers to them.

internal/dag/cache.go
```diff
@@ -528,7 +518,7 @@ func isRefToSecret(ref gatewayapi_v1beta1.SecretObjectReference, secret *v1.Secr
 
 // LookupSecret returns a Secret if present or nil if the underlying kubernetes
 // secret fails validation or is missing.
-func (kc *KubernetesCache) LookupSecret(name types.NamespacedName, validate func(*v1.Secret) error) (*Secret, error) {
+func (kc *KubernetesCache) LookupSecret(name types.NamespacedName, validate func(*Secret) error) (*Secret, error) {
```
skriss, line 521: I'm considering splitting this method out into `LookupTLSSecret`, `LookupCASecret`, and `LookupCRLSecret` and getting rid of the indirection of the passed-in validation function, thoughts? IMO it's a little simpler to navigate that way.

skriss: Went ahead with this in https://github.com/projectcontour/contour/pull/4788/commits/849d3f6aa8c8bc3000c4e2b3de50018952db53e4

sunjayBhatia: yeah, i like the option you chose, also makes it a little clearer in reading calling code what exactly is happening etc.

internal/dag/cache.go
sunjayBhatia: yeah even if it stores a few more references, we do gain some information since you have the validation status cached which stops redoing some work if a secret is used in multiple resources

skriss:
Rebased, should be ready for review.

internal/dag/secret_test.go
```diff
@@ -473,24 +341,31 @@ func TestIsValidSecret(t *testing.T) {
...
-				Err   error
-			}
-
-			want := Result{Valid: tc.valid, Err: tc.err}
-
-			valid, err := isValidSecret(tc.secret)
-			got := Result{Valid: valid, Err: err}
-
-			assert.Equal(t, want, got)
+			assert.Equal(t, tc.tlsSecretError, validTLSSecret(tc.secret))
```
sunjayBhatia, line 366: nice way of testing this, since the way the Secret type is constructed it means we can have a single secret used in different ways

internal/dag/dag.go
```diff
@@ -928,7 +928,10 @@ func (s *ServiceCluster) Rebalance() {
 // Secret represents a K8s Secret for TLS usage as a DAG Vertex. A Secret is
 // a leaf in the DAG.
 type Secret struct {
-	Object *v1.Secret
+	Object         *v1.Secret
+	ValidTLSSecret *SecretValidationStatus
```
sunjayBhatia, line 932: I guess this has the wrapper around the error since we need the distinction between not yet set, no error, and error

sunjayBhatia (approved):
to summarize:
- instead of saving secrets directly, we save secrets with their validation errors
- secrets can simultaneously be of different types
- secrets no longer validated when inserted to dag, which means more secret references
- secrets validated when validating resources that reference them

internal/dag/cache.go
ravilr: @skriss @sunjayBhatia 

Noticing a regression in Contour's resource usage after upgrading to contour-v1.24.1 (from contour-v1.22.x) . IMO, above change could be the reason ? Contour's memory usage in clusters having very large secret resources (like secrets of type `helm.sh/release.v1` )  keeps growing and leading to OOM due to reaching container's resources.memory.limits.  contour-v1.22.x on the same cluster (with similar K8s resources profile and large secrets) didn't had this issue.

1.) heap pprof from a Contour-v1.24.1 instance shows 3GB+ in use memory towards v1.Secrets resources unmarshalling ( the above change is holding back references to secrets in informer cache, preventing the unmarshall'ed bytes from GC'ed ?) :
contour-heap

2.) memory usage graph (before 03/06 12 pm and after):

Screen Shot 2023-03-07 at 11 24 14 PM

3.) Also, the cpu usage of Contour has gone up considerably since the v1.22.x to v1.24.1 upgrade (before 03/06 12pm and after) :

Screen Shot 2023-03-07 at 11 24 26 PM


Also cc @tsaarni (who's related work on filtering out unwanted Secrets from the informer cache is also relevant, IMO).

skriss: @ravilr sounds like a definite possibility. One change we could make here is to re-introduce a filter to only cache Secrets of type Opaque and TLS. I will also take a further look at @tsaarni's work here.

ravilr: On further debugging, this turned out to be a [bug in a controller causing it to patch Secret resources in a tight loop(more than 20 times per second)](https://github.com/cert-manager/cert-manager/issues/5755).  since the cluster in question went through the upgrade of multiple components including Contour,  the change in contour's resource usage was wrongly attributed to contour upgrade change. 

but, IMO, contour should still filter out unrelated secrets like helm release secrets which are very large, from its secret  informer cache, to prevent contour being affected with high resource usage and event processing overhead from those unrelated Secrets resources in a cluster. So, I hope we can take [Tero's PR ](https://github.com/projectcontour/contour/pull/5099)forward.

PR #4796, opened 2022-10-17
Add support for optional certificate validation

gautierdelorme:
This is required to support optional mTLS so that apps can use different authentication schemes (e.g. mTLS, cookies, tokens, etc...).
I initially named the new option `OnlyVerifyClientCertIfGiven` but this was confusing when `SkipClientCertValidation` is also enabled. I decided to go with `OnlyRequestClientCert` (and not `RequireClientCert` for example) since all the other option are opt-in by setting them to true so I would like to avoid changing this assumption with this new option. The other problem was at the Go API level `DownstreamValidation.RequireClientCert` default value would be `false` since that's the null value for booleans which would change the current behavior and I wanted to avoid implementing it as a pointer.

Signed-off-by: Gautier Delorme <gautier.delorme@gmail.com>

### DIFF
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -1131,6 +1131,14 @@ type DownstreamValidation struct {
 	// certificate chain will be subject to validation by CRL.
 	// +optional
 	OnlyVerifyLeafCertCrl bool `json:"crlOnlyVerifyLeafCert"`
+
+	// OnlyRequestClientCert when set to true will request a client certificate
+	// but allow the connection to continue if the client does not provide one.
+	// If a client certificate is sent, it will be verified according to the
+	// other properties, which includes disabling validation if
+	// SkipClientCertValidation is set. Defaults to false.
+	// +optional
+	OnlyRequestClientCert bool `json:"onlyRequestClientCert"`
 }
 
 // HTTPProxyStatus reports the current state of the HTTPProxy.
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -1132,13 +1132,13 @@ type DownstreamValidation struct {
 	// +optional
 	OnlyVerifyLeafCertCrl bool `json:"crlOnlyVerifyLeafCert"`
 
-	// OnlyRequestClientCert when set to true will request a client certificate
+	// OptionalClientCertificate when set to true will request a client certificate
 	// but allow the connection to continue if the client does not provide one.
 	// If a client certificate is sent, it will be verified according to the
 	// other properties, which includes disabling validation if
 	// SkipClientCertValidation is set. Defaults to false.
 	// +optional
-	OnlyRequestClientCert bool `json:"onlyRequestClientCert"`
+	OptionalClientCertificate bool `json:"optionalClientCertificate"`
 }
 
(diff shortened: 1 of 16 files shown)


gautierdelorme:
Do you think this change could make it to the 1.23 release?

tsaarni (commented):
Looks very good to me! :+1:  Small comment about the field name

apis/projectcontour/v1/httpproxy.go
```diff
@@ -1125,6 +1125,14 @@ type DownstreamValidation struct {
...
 	// +optional
 	OnlyVerifyLeafCertCrl bool `json:"crlOnlyVerifyLeafCert"`
+
+	// OnlyRequestClientCert when set to true will request a client certificate
+	// but allow the connection to continue if the client does not provide one.
+	// If a client certificate is sent, it will be verified according to the
+	// other properties, which includes disabling validation if
+	// SkipClientCertValidation is set. Defaults to false.
+	// +optional
+	OnlyRequestClientCert bool `json:"onlyRequestClientCert"`
```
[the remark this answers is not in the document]
tsaarni: My thinking was that I'd highlight the security implications first: after enabling this option the proxy will not enforce the client to present certificate to be validated, which I think `OptionalClientCertificate` could communicate bit clearer. You are right of course, that the point for the use case is that certificate is still requested from client. I'd keep that clearly stated in the description.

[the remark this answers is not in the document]
gautierdelorme: @tsaarni I'm sold 👍 Renamed in https://github.com/projectcontour/contour/pull/4796/commits/0f03cb1b58645144b6fe63b8e3ba149de9eeac32.

changelogs/unreleased/4796-gautierdelorme-minor.md
```diff
@@ -0,0 +1,5 @@
+## Optional Client Certificate Validation
+
+By default, client certificates are required but some applications might support different authentication schemes.
```
skriss, line 3:
```suggestion
By default, when client certificate validation is configured, client certificates are required.
However, some applications might support different authentication schemes.
```

skriss (changes requested):
@gautierdelorme nice work, everything looks great to me, just one minor wording suggestion on the changelog. Thanks for the PR!

changelogs/unreleased/4796-gautierdelorme-minor.md
gautierdelorme: @skriss 👍 fixed in https://github.com/projectcontour/contour/pull/4796/commits/d910de6ea00b8f2d3e3e406736b82dca7894e926

skriss: Looks like "However, some applications might support different authentication schemes." got dropped, was that intended? I'd keep that sentence in for clarity.

gautierdelorme: Good catch, it wasn't intended. Fixed in https://github.com/projectcontour/contour/pull/4796/commits/45a2250fc53d087bb4b053479492e88df0fa9a6c.

skriss (approved):
LGTM pending CI, will leave for a second maintainer to approve though. Thanks again @gautierdelorme!

tsaarni (approved):
Great to have this feature, thank you for contributing this @gautierdelorme!

PR #4798, opened 2022-10-17
Update API doc generation links to recent k8s version

sunjayBhatia:


skriss (approved):
nice, I guess we need to add this to our Kube bumps going forward

PR #4803, opened 2022-10-19
crd/ContourDeployment: Add field for set externalTrafficPolicy for service/envoy

izturn:
add the field to `crd/ContourDeployment` to support changing `Service/Envoy`'s ExternalTrafficPolicy


internal/provisioner/controller/gateway_test.go
```diff
@@ -1015,7 +1015,8 @@ func TestGatewayReconcile(t *testing.T) {
 				Spec: contourv1alpha1.ContourDeploymentSpec{
 					Envoy: &contourv1alpha1.EnvoySettings{
 						NetworkPublishing: &contourv1alpha1.NetworkPublishing{
-							Type: contourv1alpha1.ClusterIPServicePublishingType,
+							Type:                  contourv1alpha1.ClusterIPServicePublishingType,
+							ExternalTrafficPolicy: corev1.ServiceExternalTrafficPolicyTypeCluster,
```
skriss, line 1019: Please add an assertion that this field is set on the Envoy service.

internal/provisioner/controller/gatewayclass.go
```diff
@@ -180,6 +182,14 @@ func (r *gatewayClassReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 					invalidParamsMessages = append(invalidParamsMessages, msg)
 				}
+
+				switch params.Spec.Envoy.NetworkPublishing.ExternalTrafficPolicy {
+				case "", corev1.ServiceExternalTrafficPolicyTypeCluster, corev1.ServiceExternalTrafficPolicyTypeLocal:
+				default:
+					msg := fmt.Sprintf("invalid ContourDeployment spec.envoy.networkPublishing.externalTrafficPolicy %q, must be Local or Cluster",
+						params.Spec.Envoy.NetworkPublishing.ExternalTrafficPolicy)
+					invalidParamsMessages = append(invalidParamsMessages, msg)
+				}
```
skriss, line 192: Please add test cases for this to gatewayclass_test.go

internal/provisioner/objects/service/service.go
```diff
@@ -250,7 +250,7 @@ func DesiredEnvoyService(contour *model.Contour) *corev1.Service {
 	epType := contour.Spec.NetworkPublishing.Envoy.Type
 	if epType == model.LoadBalancerServicePublishingType ||
 		epType == model.NodePortServicePublishingType {
-		svc.Spec.ExternalTrafficPolicy = corev1.ServiceExternalTrafficPolicyTypeLocal
+		svc.Spec.ExternalTrafficPolicy = contour.Spec.NetworkPublishing.Envoy.ExternalTrafficPolicy
```
skriss, line 253: Please add test cases for this to service_test.go (both for the default option and the non-default option).

skriss (changes requested):

internal/provisioner/objects/service/service.go
```diff
@@ -196,10 +196,11 @@ func DesiredEnvoyService(contour *model.Contour) *corev1.Service {
...
 		Spec: corev1.ServiceSpec{
-			Ports:           ports,
-			Selector:        dataplane.EnvoyPodSelector(contour).MatchLabels,
-			SessionAffinity: corev1.ServiceAffinityNone,
-			LoadBalancerIP:  contour.Spec.NetworkPublishing.Envoy.LoadBalancer.LoadBalancerIP,
+			Ports:                 ports,
+			Selector:              dataplane.EnvoyPodSelector(contour).MatchLabels,
+			SessionAffinity:       corev1.ServiceAffinityNone,
+			LoadBalancerIP:        contour.Spec.NetworkPublishing.Envoy.LoadBalancer.LoadBalancerIP,
+			ExternalTrafficPolicy: corev1.ServiceExternalTrafficPolicyTypeLocal,
```
skriss, line 203: We need to drop this line, since we can only set this field if the Service type is NodePort or LoadBalancer. It's already being set in those cases on L254.

internal/provisioner/objects/service/service_test.go
```diff
@@ -177,14 +177,19 @@ func TestDesiredEnvoyService(t *testing.T) {
 	checkServiceHasPortName(t, svc, "https")
 	checkServiceHasPortProtocol(t, svc, corev1.ProtocolTCP)
 
+	cntr.Spec.NetworkPublishing.Envoy.Type = model.ClusterIPServicePublishingType
+	svc = DesiredEnvoyService(cntr)
+	checkServiceHasExternalTrafficPolicy(t, svc, corev1.ServiceExternalTrafficPolicyTypeLocal)
```
skriss, line 182: In this case there should be _no_ external traffic policy, since it's invalid to set it for a ClusterIP service. Please modify the assertion to ensure that the field is not set.

internal/provisioner/controller/gateway_test.go
```diff
@@ -1015,7 +1015,8 @@ func TestGatewayReconcile(t *testing.T) {
 				Spec: contourv1alpha1.ContourDeploymentSpec{
 					Envoy: &contourv1alpha1.EnvoySettings{
 						NetworkPublishing: &contourv1alpha1.NetworkPublishing{
-							Type: contourv1alpha1.ClusterIPServicePublishingType,
+							Type:                  contourv1alpha1.ClusterIPServicePublishingType,
```
skriss, line 1018: Change this to a NodePort service since it's not valid to set an ExternalTrafficPolicy on a ClusterIP service:
```suggestion
							Type:                  contourv1alpha1.NodePortServicePublishingType,
```

skriss (changes requested):

skriss (approved):
LGTM, will leave for @sunjayBhatia

sunjayBhatia (approved):

Issue #4802, opened 2022-10-18
Handle Envoy 1.24.0 deprecations


sunjayBhatia:
See: https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.24/v1.24.0#deprecated

Assuming 1.24.0 will be picked up in Contour 1.23.0, we should make these changes in Contour 1.24.0 (to ensure a smooth upgrade), since they seem to include deprecations paired with additions to the now-preferred fields

skriss:
Should update the go-control-plane module dep as part of this.

PR #4804, opened 2022-10-19
update Envoy to v1.24.0

### LINKED ISSUE #4736: Pick up Envoy 1.24.0
Envoy 1.24.0 is scheduled to be out in the first half of October. We should include it in the upcoming Contour 1.23 assuming it passes CI and there are no obvious compat issues in the changelog.

skriss:
Closes #4736.


skriss:
I ended up not bumping go-control-plane here because the newly deprecated fields are triggering linter errors, I figure we can do that when we address #4802, let me know if you see any issue with that.

sunjayBhatia (approved):
LGTM, only nit would be to add a link to the envoy release notes in our changelog

and yeah on the deprecations, seems totally reasonable to not bump go-control-plane, we're not using any of the new apis yet, unless we want to add explicit nolint annotations everywhere but that might be a pain

skriss:
> unless we want to add explicit nolint annotations everywhere but that might be a pain

Yeah thought about it but didn't seem worth it.

PR #4806, opened 2022-10-20
Updated actions/checkout and actions/cache from version 2 to version 3

fMurugi:
Updated actions/checkout and actions/cache from version 2 to version 3


.github/workflows/prbuild.yaml
```diff
@@ -14,9 +14,9 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v2
+        uses: golangci/golangci-lint-action@v3
```
skriss, line 19: Looks like this change is triggering some new lint errors that need to be looked into. Can either do that in this PR, or remove the version bump of golangci-lint from this PR and deal with that separately.

fMurugi: what do you mean remove the version bump?

skriss: You got it 👍

skriss (approved):
Thanks @fMurugi, LGTM. We can use the DCO you applied to your first commit when we squash/merge.

PR #4811, opened 2022-10-20
use Docker cache mounts for go build & mod cache

skriss:
This significantly speeds up repeat builds
on the same host, and is more effective at
caching go modules than the Docker layer
caching technique that was being used
previously since a single module change
does not invalidate the entire cache.

Note, this won't make CI builds any faster
since they run on a fresh host every time,
but makes repeat local dev builds way faster.

ref. https://docs.docker.com/engine/reference/builder/#run---mount


sunjayBhatia (approved):
very cool!

PR #4813, opened 2022-10-21
Doc updates for "FIPS 140-2 in Contour"

### LINKED ISSUE #4794: [Doc] The "goboring-version" in "FIPS 140-2 in Contour" doc is outdated.
**On Page:** [FIPS 140-2 in Contour](https://projectcontour.io/guides/fips/)

The specific paragraphs are as follows: 

```
We can use the Google-provided Go implementation that has patches on top of standard Go to enable integrating BoringCrypto. This is available to us in the [goboring/golang](https://hub.docker.com/r/goboring/golang/) container image we can use as a build base. In addition, to ensure we can statically compile the contour binary when it is linked with the BoringCrypto C library, we must pass some additional arguments to the make container target.
To perform the Contour image build with BoringCrypto, change directories to where you have the Contour source code checked out and run the following (replacing <goboring-version-tag> with the appropriate version of Go and BoringCrypto, see [here](https://go.googlesource.com/go/+/dev.boringcrypto/misc/boring/README.md#version-strings) for version specifics):
```

However, when I checked [the version list of goboring/golang on Docker hub](https://hub.docker.com/r/goboring/golang/tags), the latest version is [1.16.7b7](https://hub.docker.com/layers/goboring/golang/1.16.7b7/images/sha256-ad86974b2b260afa11293748b6c1e0fe0edbeb5874f8850d108cdcdf0a5795c1?context=explore). The Go 1.17+ versions are not there according to [the image overview](https://go.googlesource.com/go/+/dev.boringcrypto/README.boringcrypto.md).

```
This branch is no more. GOEXPERIMENT=boringcrypto on the standard branch does the job now.
```

So there is no need to use `goboring/golang` for BoringCrypto enable, instead, we can use the standard `golang` image with `GOEXPERIMENT=boringcrypto`. I've tested the output container image in this way and it seems to work well. Test according to:

```
To be fully sure the produced contour binary has been compiled with BoringCrypto you must remove the -s flag from the base Contour Makefile to stop stripping symbols and run through the build process above. Then you will be able to inspect the contour binary with go tool nm to check for symbols containing the string _Cfunc__goboringcrypto_.
```

So maybe the doc needs some updates, I will be glad to submit a PR if it is convenient.

moeyui1:
Some updates for [FIPS 140-2 in Contour](https://projectcontour.io/guides/fips/), see also #4794 :

1. Add description for Go 1.19+ to enable integrating BoringCrypto;
2. Add [rsc.io/goversion](https://godoc.org/rsc.io/goversion) as an another way to validate whether the `contour` binary has been compiled with BoringCrypto.

Closes #4794.


site/content/guides/fips.md
```diff
@@ -47,8 +47,9 @@ The Contour [Dockerfile][8] uses a multistage build that performs compilation in
 In order to minimize the `projectcontour/contour` image footprint, the final output image only consists of a single layer, containing a lone file: the statically compiled `contour` binary.
 The standard Contour build uses the upstream `golang` image as a build base, however we will have to swap that out to build Contour with BoringCrypto.
 
-We can use the Google-provided Go implementation that has patches on top of standard Go to enable integrating BoringCrypto.
-This is available to us in the [`goboring/golang`][9] container image we can use as a build base.
+Starting with Go 1.19, you can simply add [`GOEXPERIMENT=boringcrypto`][18] to enable integrating BoringCrypto for standard Go. For the Go version under 1.19, we can use the Google-provided Go implementation that has patches on top of standard Go to enable integrating BoringCrypto.
```
skriss, line 50: It'd be great if you could add an example of the command needed when using this Go 1.19 method. One option for organization would be to split the `Building Contour` section into multiple subsections, i.e. one for Go 1.19+ and one for older Go versions.

skriss (dismissed):
Thanks @moeyui1 for the updates here! One comment from me. cc @sunjayBhatia

site/content/guides/fips.md
```diff
@@ -68,7 +69,7 @@ The command above can be broken down as follows:
 The container image build process should fail before export of the `contour` binary to the final image if the compiled binary is not statically linked.
 
 To be fully sure the produced `contour` binary has been compiled with BoringCrypto you must remove the `-s` flag from the base Contour `Makefile` to stop stripping symbols and run through the build process above.
-Then you will be able to inspect the `contour` binary with `go tool nm` to check for symbols containing the string `_Cfunc__goboringcrypto_`.
+Then you will be able to inspect the `contour` binary with `go tool nm` to check for symbols containing the string `_Cfunc__goboringcrypto_`. Also, you can use the program [rsc.io/goversion][21]. It will report the crypto implementation used by a given binary when invoked with the `-crypto` flag.
```
sunjayBhatia, line 72: tiny nit: we tend to separate sentences to one per line in docs like these to make diffs easier to read
```suggestion
Then you will be able to inspect the `contour` binary with `go tool nm` to check for symbols containing the string `_Cfunc__goboringcrypto_`.
Also, you can use the program [rsc.io/goversion][21]. It will report the crypto implementation used by a given binary when invoked with the `-crypto` flag.
```

sunjayBhatia:
thanks for the PR @moeyui1, running through the update instructions now 👍🏽

site/content/guides/fips.md
sunjayBhatia: note the additional flags and enabling CGO from previous versions of Go don't seem to be necessary, so having two separate sections to make that clear would be nice

moeyui1: Thanks for the suggestion and supplement. 👍  I will add them to the doc later.
But with CGO enabled, I got this output running `goversion` for inspection:

> ./contour go1.19.2 X:boringcrypto (boring crypto)

while with CGO disabled:

>./contour go1.19.2 X:boringcrypto (standard crypto)

So it seems that CGO enabled is still necessary for `boringcrypto`. And the same as the additional flags for CGO enabled: 

> BUILD_EXTRA_GO_LDFLAGS="-linkmode=external -extldflags=-static"

sunjayBhatia: ah good call, I was passing the `-v` flag not `-crypto`, running:

```
$ make container BUILD_GOEXPERIMENT=boringcrypto BUILD_CGO_ENABLED=1 BUILD_EXTRA_GO_LDFLAGS="-linkmode=external -extldflags=-static"
```

you should get the proper output:

```
$ goversion -crypto ./contour
./contour go1.19.2 X:boringcrypto (boring crypto)
```

sunjayBhatia: no issues so far, so maybe can be ignored, but something to keep an eye on if there are any other relevant warnings in the future

moeyui1:
A little busy these days, will complete this a few days later.

site/content/guides/fips.md
```diff
@@ -47,8 +47,107 @@ The Contour [Dockerfile][8] uses a multistage build that performs compilation in
...
 
-We can use the Google-provided Go implementation that has patches on top of standard Go to enable integrating BoringCrypto.
-This is available to us in the [`goboring/golang`][9] container image we can use as a build base.
+### Go with BoringCrypto
+
+#### Go version 1.19 and higher
+
+Starting with Go 1.19, you can simply add [`GOEXPERIMENT=boringcrypto`][18] to enable integrating BoringCrypto for standard Go.
+
+First of all we need to change the `Dockerfile`, to declare a new arg:
```
sunjayBhatia, line 56: ah sorry for the confusion, but I intended us to actually make these changes to the Makefile/Dockerfile so users don't have to modify them themselves, could you move these changes there? should also simplify this section

sunjayBhatia (changes requested):

site/content/guides/fips.md
moeyui1: Oh I see, never mind :)

sunjayBhatia (approved):

Issue #4814, opened 2022-10-21
update more github actions to address deprecations


skriss:
- [x] [actions/stale](https://github.com/projectcontour/contour/blob/main/.github/workflows/stale.yaml#L14)
- [x] [docker/setup-buildx-action](https://github.com/projectcontour/contour/blob/main/.github/workflows/prbuild.yaml#L88)
- [x] [docker/login-action](https://github.com/projectcontour/contour/blob/main/.github/workflows/build_main.yaml#L20)
- [x] [golangci-lint-action](https://github.com/projectcontour/contour/blob/main/.github/workflows/prbuild.yaml#L19)
- [ ] [mheap/github-action-required-labels](https://github.com/projectcontour/contour/blob/main/.github/workflows/label_check.yaml#L23) (waiting on a new version of the action; setup-go also needs to be bumped in the label_check workflow when this is done)

fMurugi:
@skriss can i update this also?
And how do i know the versions to update to? Thank you.

skriss:
@fMurugi definitely, thanks!

If you find the GitHub repo for each action (e.g. https://github.com/docker/setup-buildx-action), we can use the newest major version for each. The README should have some examples that show which version to use (e.g. for the buildx action, it'd be `@v2`). Let me know if you need further info.

skriss:
Also note to do a search across all of our workflows in `.github/workflows` for usages of the above actions, some may be used more than once in more than one file.

fMurugi:
Okay thank you for the info.

fMurugi:
Am not finding the latest version of actions/stale 
here is where am looking https://github.com/marketplace/actions/stale-for-actions

skriss:
That one is actually https://github.com/actions/stale, looks like `@v6` is the latest

fMurugi:
Saw it thanks

skriss:
We've now got dependabot in place which will update our actions for us automatically going forward. Thanks again @fMurugi for all your work here.

PR #4821, opened 2022-10-25
update golangci-lint version and action version

skriss:
Updates #4814.

I'll create a separate issue for possibly addressing the issue reported by gosec for the shutdown manager.


sunjayBhatia (approved):
yeah gosec is now v strict about http servers/timeouts!

PR #4824, opened 2022-10-26
Fix upgrade from version tag finding

### LINKED ISSUE #4809: upgrade tests getting wrong "from" version on release branch before release tag created
https://github.com/projectcontour/contour/actions/runs/3290540851/jobs/5423694404#step:7:142

This is on a release branch, but before a tag has been created on that branch. It should be upgrading from the latest tag in this case (i.e. v1.22.1).

https://github.com/projectcontour/contour/blob/main/test/scripts/get-contour-upgrade-from-version.sh#L10-L12 is the code in question. Needs to handle both (a) release branch that doesn't yet have a tag created on it; and (b) older release branches, i.e. `release-1.22`, where the "from" should be the latest v1.22.x tag.

sunjayBhatia:
- logic for when we're checked out on a tag remains the same
- Contour v1.2.0 was improperly tagged on main, so next case is when we're checked out on a release branch with a previous tag, we ignore v1.2.0 and use the existing tag
  - no branches other than release branches should have tags
- remaining cases are for release branches w/ tags, main, etc. so find the latest tag

Fixes #4809

### DIFF
--- a/test/scripts/get-contour-upgrade-from-version.sh
+++ b/test/scripts/get-contour-upgrade-from-version.sh
@@ -7,11 +7,22 @@ set -o pipefail
 if CURRENT_TAG=$(git describe --tags --exact-match 2>/dev/null); then
   # We are on a tag, so find previous tag to this one.
   git tag -l --sort=-v:refname | grep -v 'alpha\|beta\|rc' | grep -A1 -x $CURRENT_TAG | tail -1
-elif git branch --show-current | grep -q release; then
-  # We are on a release branch, so find tag.
-  git describe --tags --abbrev=0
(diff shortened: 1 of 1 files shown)


sunjayBhatia:
this branch
```
$ git branch --show-current
fix-upgrade-from-version-rel-branch
$ ./test/scripts/get-contour-upgrade-from-version.sh
v1.23.0
```

`main` (w/ these changes cherry-picked)
```
$ git branch --show-current
main
$ ./test/scripts/get-contour-upgrade-from-version.sh
v1.23.0
```

skriss (approved):
Nice, LGTM

PR #4827, opened 2022-10-28
Service updates/deletes only trigger DAG rebuilds if referenced by an HTTPProxy

### LINKED ISSUE #4793: updates/deletes to unrelated Services trigger DAG rebuilds
Spun out from https://github.com/projectcontour/contour/issues/4386, it looks like the same thing happens for Services since we are not calling `serviceTriggersRebuild` on removals (which happen for both updates and deletes).

Vishal-Chdhry:
Deleting only those services that is referenced by an HTTPProxy, triggers a rebuild

Closes #4793 

Signed-off-by: Vishal Choudhary <contactvishaltech@gmail.com>


internal/featuretests/v3/cluster_test.go
```diff
@@ -656,22 +666,48 @@ func TestUnreferencedService(t *testing.T) {
...
+	// does not trigger a rebuild
+	rh.OnDelete(s2)
+	res = c.Request(clusterType)
+	res.Equals(&envoy_discovery_v3.DiscoveryResponse{
+		Resources: resources(t,
+			cluster("default/kuard/80/da39a3ee5e", "default/kuard", "default_kuard_80"),
+		),
+		TypeUrl: clusterType,
+	})
+	assertEqualVersion(t, "2", res)
```
skriss, line 700: Per the failing test, looks like the expected version here is actually "1" - the initial add of `s1` does not trigger a DAG build since it's not _yet_ referenced, then the add of the `default/simple` HTTPProxy _does_ trigger a DAG build, then adding/deleting s2 should _not_ trigger a rebuild, leaving us at a version of "1".

internal/featuretests/v3/cluster_test.go
```diff
@@ -656,22 +666,48 @@ func TestUnreferencedService(t *testing.T) {
...
+
+	// verifying that deleting a Service that is referenced by an HTTPProxy,
+	// triggers a rebuild
+	rh.OnDelete(s1)
+	res = c.Request(clusterType)
+	res.Equals(&envoy_discovery_v3.DiscoveryResponse{
 		Resources: resources(t,
 			cluster("default/kuard/80/da39a3ee5e", "default/kuard", "default_kuard_80"),
```
skriss, line 708: since we deleted `s1` just above, we no longer expect a cluster here

internal/featuretests/v3/cluster_test.go
```diff
@@ -656,22 +666,48 @@ func TestUnreferencedService(t *testing.T) {
...
 		),
 		TypeUrl: clusterType,
 	})
+	assertEqualVersion(t, "3", res)
```
skriss, line 712: I believe this should be "2" per the above comment.

skriss:
Thanks for the PR @Vishal-Chdhry! The code change looks good, added a couple notes re: the failing tests. Please also add a changelog file, per https://github.com/projectcontour/contour/actions/runs/3346425314/jobs/5543220033.

internal/featuretests/v3/cluster_test.go
```diff
@@ -656,22 +666,41 @@ func TestUnreferencedService(t *testing.T) {
...
 	})
 
-	c.Request(clusterType).Equals(&envoy_discovery_v3.DiscoveryResponse{
+	res := c.Request(clusterType)
+	res.Equals(&envoy_discovery_v3.DiscoveryResponse{
 		Resources: resources(t,
 			cluster("default/kuard/80/da39a3ee5e", "default/kuard", "default_kuard_80"),
 		),
 		TypeUrl: clusterType,
 	})
```
skriss, line 676:
```suggestion
assertEqualVersion(t, "1", res)
```

internal/featuretests/v3/cluster_test.go
```diff
@@ -656,22 +666,41 @@ func TestUnreferencedService(t *testing.T) {
...
 
-	c.Request(clusterType).Equals(&envoy_discovery_v3.DiscoveryResponse{
+	res = c.Request(clusterType)
+	res.Equals(&envoy_discovery_v3.DiscoveryResponse{
+		Resources: resources(t,
+			cluster("default/kuard/80/da39a3ee5e", "default/kuard", "default_kuard_80"),
+		),
+		TypeUrl: clusterType,
+	})
+
```
skriss, line 689:
```suggestion
assertEqualVersion(t, "1", res)
```

internal/featuretests/v3/cluster_test.go
```diff
@@ -656,22 +666,41 @@ func TestUnreferencedService(t *testing.T) {
...
 			cluster("default/kuard/80/da39a3ee5e", "default/kuard", "default_kuard_80"),
 		),
 		TypeUrl: clusterType,
 	})
+	assertEqualVersion(t, "1", res)
+
+	// verifying that deleting a Service that is referenced by an HTTPProxy,
+	// triggers a rebuild
+	rh.OnDelete(s1)
+	assertEqualVersion(t, "2", res)
```
skriss, line 705:
```suggestion
        res = c.Request(clusterType)
	assertEqualVersion(t, "2", res)
```

internal/featuretests/v3/cluster_test.go
```diff
@@ -623,9 +624,18 @@ func TestUnreferencedService(t *testing.T) {
...
-	rh.OnAdd(fixture.NewService("kuard").
-		WithPorts(v1.ServicePort{Port: 80, TargetPort: intstr.FromString("8080")}),
-	)
+	// Equals(...) only checks resources, so explicitly
+	// check version & nonce here and subsequently.
+	assertEqualVersion := func(t *testing.T, expected string, r *Response) {
+		t.Helper()
+		assert.Equal(t, expected, r.VersionInfo, "got unexpected VersionInfo")
+		assert.Equal(t, expected, r.Nonce, "got unexpected Nonce")
+	}
```
skriss, line 633: Not strictly needed for this PR, but it would be nice to move this to be a method on the `Response` struct, so you could do `r.EqualVersion("1")` similarly to how you can do `r.Equals(...)` currently. That would replace the `assertEqualVersion` helper here and in `secrets_test.go`.

skriss (changes requested):
@Vishal-Chdhry few more comments on the feature test

skriss (approved):
Thanks for the updates @Vishal-Chdhry, LGTM!

Issue #4738, opened 2022-09-20
Pick up Gateway API v0.6.0


skriss:
Contour 1.24 should pick up Gateway API v0.6.0.

Sub-tasks:
- [x] https://github.com/projectcontour/contour/issues/4555
- [x] Gateway Scheduled -> Accepted condition (https://github.com/kubernetes-sigs/gateway-api/pull/1447)
- [x] Listener Detached: true -> Accepted: false (Detached) (https://github.com/kubernetes-sigs/gateway-api/pull/1446)

PR #4839, opened 2022-11-07
[dev-gwapi-0.6] update to latest Gateway API commit

skriss:
- Update to v1beta1 ReferenceGrant
- HTTPRequestHeaderFilter -> HTTPHeaderFilter
- GatewayConditionScheduled -> GatewayConditionAccepted
- ListenerConditionDetached -> ListenerConditionAccepted

Updates #4555.
Updates #4738.


skriss:
This was everything needed to get Contour compiling + passing the linter against the current Gateway API HEAD.

skriss:
I've made the assumption here that we don't need to continue setting the old/deprecated conditions along with the new, which is much more straight-forward implementation-wise. Let me know if you think we should do something different!

sunjayBhatia (approved):

Issue #4473, opened 2022-04-15
provisioner: fix reconciliations to only update fields that should change over time


skriss:
With Gateway API, parameters defined by the GatewayClass should apply at the time of Gateway instantiation, but subsequent updates to those parameters should not flow through to already instantiated resources. The provisioner is doing this partly correctly, and partly incorrectly, since it was based on existing code that did ~full reconciliation on updates. We need to go through each resource's "Ensure" function and make sure that only expected updates are flowing through to provisioned resources (e.g. listeners, addresses).

Rycieos:
This just bit me hard. I was following the docs on proxy-proto (https://projectcontour.io/docs/main/guides/proxy-proto/), and added these values to my `ContourDeployment`:

```yaml
spec:
  runtimeSettings:
    envoy:
      listener:
        useProxyProtocol: true
  envoy:
    networkPublishing:
      serviceAnnotations:
        service.beta.kubernetes.io/aws-load-balancer-backend-protocol: tcp
        service.beta.kubernetes.io/aws-load-balancer-proxy-protocol: '*'
```

As I had read in [the GatewayClass guide](https://projectcontour.io/docs/main/config/gateway-api/#customizing-a-gatewayclass):
> Contour follows the recommended behavior, meaning changes to a GatewayClass and its parameters are not propagated down to existing Gateways.

I expected none of those changes to propagate to my current Gateways. While the `runtimeSettings` did not, after a few minutes the `serviceAnnotations` did, at which point AWS started speaking PROXY protocol, and envoy did not know how to handle it, breaking all connections.

I get that a full dig through the project for places this expectation is broken is a difficult task, so maybe changing the quoted section of the docs to add some warning would be a good interim solution.

PR #4841, opened 2022-11-09
internal/provisioner: fix EnsureContourConfig bug

izturn:


skriss:
@izturn this is actually working as intended. Per the [Gateway API spec](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.GatewayClass):

> It is recommended that this resource be used as a template for Gateways. This means that a Gateway is based on the state of the GatewayClass at the time it was created and changes to the GatewayClass or associated parameters are not propagated down to existing Gateways. This recommendation is intended to limit the blast radius of changes to GatewayClass or associated parameters.

So changes to the ContourDeployment/ContourConfiguration associated with a GatewayClass should _not_ be applied to existing Gateways, instead they should only be used for new Gateways going forward.

skriss:
Also see https://github.com/projectcontour/contour/issues/4473 for fixing this across other parts of the provisioner code.

izturn:
thx for your explanation @skriss

PR #4849, opened 2022-11-15
[dev-gwapi-0.6] replace Gateway/Listener Ready conditions with Programmed

skriss:
Replaces the "Ready" condition used for Gateways
and Listeners with "Programmed".

Updates #4848.

### DIFF
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -93,11 +93,12 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
 	)
 	defer commit()
 
-	var gatewayNotReadyCondition *metav1.Condition
+	var gatewayNotProgrammedCondition *metav1.Condition
 
 	if !isAddressAssigned(p.source.gateway.Spec.Addresses, p.source.gateway.Status.Addresses) {
-		gatewayNotReadyCondition = &metav1.Condition{
-			Type:    string(gatewayapi_v1beta1.GatewayConditionReady),
+		// TODO(sk) resolve condition type-reason mismatch
+		gatewayNotProgrammedCondition = &metav1.Condition{
+			Type:    string(gatewayapi_v1beta1.GatewayConditionProgrammed),
 			Status:  metav1.ConditionFalse,
 			Reason:  string(gatewayapi_v1beta1.GatewayReasonAddressNotAssigned),
 			Message: "None of the addresses in Spec.Addresses have been assigned to the Gateway",
@@ -148,7 +149,7 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
 				routeParentStatusAccessor := routeAccessor.StatusUpdateFor(routeParentRef)
 
 				// If the Gateway is invalid, set status on the route and we're done.
-				if gatewayNotReadyCondition != nil {
+				if gatewayNotProgrammedCondition != nil {
 					routeParentStatusAccessor.AddCondition(gatewayapi_v1beta1.RouteConditionAccepted, metav1.ConditionFalse, status.ReasonInvalidGateway, "Invalid Gateway")
 					continue
 				}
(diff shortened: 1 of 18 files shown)


internal/dag/gatewayapi_processor.go
```diff
@@ -93,11 +93,12 @@ func (p *GatewayAPIProcessor) Run(dag *DAG, source *KubernetesCache) {
 	)
 	defer commit()
 
-	var gatewayNotReadyCondition *metav1.Condition
+	var gatewayNotProgrammedCondition *metav1.Condition
 
 	if !isAddressAssigned(p.source.gateway.Spec.Addresses, p.source.gateway.Status.Addresses) {
-		gatewayNotReadyCondition = &metav1.Condition{
-			Type:    string(gatewayapi_v1beta1.GatewayConditionReady),
+		// TODO(sk) resolve condition type-reason mismatch
```
skriss, line 99: The "AddressNotAssigned" reason did not move over to the "Programmed" condition, it's still documented as being for the "Ready" condition (ref. https://github.com/kubernetes-sigs/gateway-api/blob/eb822e981bd8f7b4f52e89cafbe2d0397b5f01fb/apis/v1beta1/gateway_types.go#L643-L648), so need to figure out what to do here. It would be a little odd to set "Ready: false" for this scenario, but never set "Ready: true" (since we're not currently able to meet the requirements for the "Ready: true" condition).

internal/dag/gatewayapi_processor.go
```diff
@@ -881,39 +882,40 @@ func (p *GatewayAPIProcessor) computeGatewayConditions(gwAccessor *status.Gatewa
...
 				break
 			}
 		}
 
-		if !allListenersReady {
-			// If we have invalid listeners, set Ready=false.
-			gwAccessor.AddCondition(gatewayapi_v1beta1.GatewayConditionReady, metav1.ConditionFalse, gatewayapi_v1beta1.GatewayReasonListenersNotValid, "Listeners are not valid")
+		if !allListenersProgrammed {
+			// If we have invalid listeners, set Programmed=false.
+			// TODO(sk) resolve condition type-reason mismatch
```
skriss, line 914: Similarly here, the "ListenersNotValid" reason did not move over to the "Programmed" condition (ref. https://github.com/kubernetes-sigs/gateway-api/blob/eb822e981bd8f7b4f52e89cafbe2d0397b5f01fb/apis/v1beta1/gateway_types.go#L634-L637), so need to decide what to do here.

skriss (commented):
This was mostly straightforward, but there are a couple questionable combos of Condition type + reason that need to be resolved.

internal/dag/gatewayapi_processor.go
skriss: I'm feeling like this is probably fine as-is for now since it's not breaking any conformance tests, but possibly something to follow up on down the road.

internal/dag/gatewayapi_processor.go
sunjayBhatia: yeah the wording on that Reason makes it sounds like it should be used with the Programmed Condition

`one or more Listeners have an invalid or unsupported configuration and cannot be configured on the Gateway.`

good for us to keep now and open an upstream issue

internal/dag/gatewayapi_processor.go
sunjayBhatia: hm yeah this one is a little weird, I could go either way on whether it belongs with the Programmed or Ready condition, but seems find to leave as you have it here until an upstream issue/conformance test changes things

sunjayBhatia (approved):

internal/dag/gatewayapi_processor.go
skriss: :+1: yeah agree this one could potentially go either way. I'll raise an issue upstream for it.

internal/dag/gatewayapi_processor.go
skriss: Agreed. Will raise an upstream issue.

PR #4853, opened 2022-11-24
Internal Redirect Support in HTTPProxy

Jean-Daniel:
Hello,

I'm new to the contour community, but I would like to propose to add support for http internal redirect.

This is a feature we need to migrate our infrastructure to Contour, and something Envoy support for a very long time (v1.10).

Here is a first draft proposal, and I also have a working implementation to test it.

I'm looking for comments, especially for the API design part as I don't have many experience designing CRDs.

Thanks.


tsaarni (commented):
Welcome @Jean-Daniel to the Contour community! Thank you for your contribution!

design/internal-redirect_design.md
```diff
@@ -0,0 +1,212 @@
...
+         - name: echo
+           port: 8080
+       internalRedirectPolicy:
+         maxInternalRedirects: 5
+         redirectResponseCodes: [ 301, 302, 303 ]
+         predicates:
+           - name: SafeCrossScheme
+           - name: PreviousRoutes
+           - name: AllowListedRoutes
+             allowedRouteNames: [ "other" ]
```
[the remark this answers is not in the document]
Jean-Daniel: To be honest, I'm not satisfy by the `name` + `allowedRoutesNames` tuple too.
I prefer your suggestion. As internal redirect predicates [by design can only reject a route](https://github.com/envoyproxy/envoy/blob/bba0a84834859426e05ea912f0b2f1a2c79d0920/source/common/router/router.cc#L1737), the predicate evaluation order does not really matter and we can probably use a simpler representation like:

```yaml
predicates: 
  safeCrossScheme: false
  previousRoutes: false
  allowedRouteNames: []
```

and generate the AllowListedRoutes predicate only if the list is not empty.

design/internal-redirect_design.md
```diff
@@ -0,0 +1,212 @@
...
+## Background
+Internal redirect is a way to let the proxy server intercept 3xx redirect response, synthesizing a new request, sending it to the upstream specified by the new route match, and returning the redirected response as the response to the original request.
+Internal redirect is supported by Envoy since v1.10, but there is no way to configure an internal redirect policy using Contour.
+
+## Goals
+- Configuring per route internal redirect policy.
+- Supporting envoy builtin 'internal redirect policy' predicates.
+
+## Non Goals
+- Supporting user defined 'internal redirect policy' predicate.
```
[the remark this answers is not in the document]
Jean-Daniel: Predicates are Typed Extension, and you can use the standard (but poorly documented) [envoy extension machinery](https://www.envoyproxy.io/docs/envoy/latest/extending/extending) to create new predicates.

design/internal-redirect_design.md
```diff
@@ -0,0 +1,212 @@
+# Internal Redirect
+
+Status: Draft
+
+## Abstract
+This document describes the API changes needed to support internal redirect policy support in contour HTTPProxy.
+
+## Background
+Internal redirect is a way to let the proxy server intercept 3xx redirect response, synthesizing a new request, sending it to the upstream specified by the new route match, and returning the redirected response as the response to the original request.
```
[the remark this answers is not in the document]
Jean-Daniel: That's a good question. As authorisations are envoy filters, they should be executed for the redirect. The Internal Redirect documentation says that:

> The altered request headers will then have a new route selected, be sent through a new filter chain, and then shipped upstream with all of the normal Envoy request sanitization taking place.

design/internal-redirect_design.md
```diff
@@ -0,0 +1,179 @@
...
+    // AllowedRouteNames is the list of routes that’s allowed as redirect target
+    // by this predicate, identified by the route’s name.
+    //
+    // +optional
+    AllowedRouteNames []string `json:"allowedRouteNames,omitempty"`
+}
+```
+
+Update `Route` accordingly. 
+The `AllowedRouteNames` predicates declares a list of allowed route's name, so a `Name` field is required on `Route` to be able to use it.
```
sunjayBhatia, line 105: if this directly translates to this field: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-route-name

we will likely need to ensure uniqueness across all routes to make sure routes are configured correctly redirect predicates work correctly

Jean-Daniel: Yes, it translate to envoy route name field.

Is uniqueness required by envoy ? 

Regarding the internal redirect use, nothing prevent you to use the same name for multiple routes. For instance, if you have 2 routes that point to the same resource, you may use the same `Name`, and only declare a single entry in `AllowedRouteNames`.

Redirect predicates should work correctly, as its only job is to check if the route name is in the route allowed list.

sunjayBhatia: in a multi-tenant scenario where multiple teams/users are configuring HTTPProxies (or there are multiple HTTPProxies using the feature) we should make sure names don't conflict between users, that could generate unexpected results for people if route names are reused

sunjayBhatia: we'll need to think about this in the implementation section of this design/feature, something to keep track of here to make sure we provide a general enough API that is safe to use

tsaarni: It seems the route name feature originates from debugging/troubleshooting use case ([see here](https://github.com/envoyproxy/envoy/issues/6364)) for printing to access log with `%ROUTE_NAME%` macro. Quickly browsing the code, I did not find any uniqueness requirements. 

Can allow list point even to a route name on different virtual host? 

In context of multi-tenant scenario, what would be natural scope for the allow list? For example: single namespace or e.g. root `HTTPProxy` (virtualhost) and its child proxies (possibly in other namespaces)?

Jean-Daniel: > In context of multi-tenant scenario, what would be natural scope for the allow list?

The problem with scoping the route name is that you limit the possible use cases. Team A may legitimately want to configure internal redirect for a route on Team B proxy, and configuring an internal redirect pointing on a different virtual host may also be very useful.

And the risk if we do not enforce route uniqueness is that in case of multiple teams provisioning the same proxy, team A configures an internal redirect on a route to handle a specific redirect response, and the team A service behind that route may also return a redirect response with the same status code an a location that matches a team B route with the same name. In such case, this other redirect is handled internally instead of being return to the client.

If the team A service does not returns redirect response pointing on Team B route (or use a different status code for internal redirects and other redirects), having duplicated routes on team B side will not have any impact for team A.

We have to find the right balance between risk mitigation and reducing the capability of this feature.

Jean-Daniel: Thinking more about it, enforcing route uniqueness may be the way to go.

It does not limit the possibility to use `Allow List` with redirects accros virtual hosts and accros namespaces (as long as teams agree on a route name), and the only drawback is that you may have to declare multiple routes instead of one in the allowed list.

Moreover, enforcing route name uniqueness will make the name more useful for debugging/troubleshooting routing issues.

design/internal-redirect_design.md
```diff
@@ -0,0 +1,182 @@
...
+    // +optional
+    Predicates HTTPInternalRedirectPredicate `json:"predicates,omitempty"`
+
+    // AllowCrossSchemeRedirect Allow internal redirect to follow a target URI with a different scheme than the value of x-forwarded-proto.
+    // +optional
+    AllowCrossSchemeRedirect bool `json:"allowCrossSchemeRedirect,omitempty"`
+}
+
+// HTTPInternalRedirectPredicate defines the predicate used to accept or reject an internal redirection.
+type HTTPInternalRedirectPredicate struct {
```
sunjayBhatia, line 87: would it make sense to get rid of this additional type and pull these up to the HTTPInternalRedirectPolicy struct? I'm not sure how much value having these nested gives us in this API

we don't necessarily need to use "predicate" in the naming to make things a little clearer for users, I think the Envoy names also may be a little opaque

e.g SafeCrossScheme -> something that conveys "allow mismatched downstream and redirect target scheme" (this one is harder to come up with a good name 😅 ), or PreviousRoutes -> AllowRepeatedRouteRedirect

Jean-Daniel: This would also be an opportunity to solve the configuration issue where `SafeCrossScheme` being meaningful only if `AllowCrossSchemeRedirect` redirect is `true`. 

It think both fields should be merged into something like `CrossSchemeRedirect` and take an enum like `always`, `never` and `safeOnly`.

Jean-Daniel: I have updated the proposal by removing the nested type, and trying to simplify the API.

skriss (commented):
Thanks for the design doc @Jean-Daniel and apologies for the delay in reviewing.

One question I have: is support for the `allow_listed_routes` predicate a hard requirement for you? Or would it be possible to ship this feature without it? That individual predicate seems like the thorniest part of this design since it requires route names, so just wondering whether it's a must-have or not.

design/internal-redirect_design.md
```diff
@@ -0,0 +1,170 @@
...
+N/A
+
+## Security Considerations
+
+When targeting an untrusted backend which can return any value in the `Location` header, the gateway operator may want to be able to limit the scope of the internal redirect to a list of predefined routes by populating the `allowedRouteNames` field.
+But in the case of multitenant Contour deployment, 2 teams may declare routes using the same name, defeating the purpose of the allowed routes list.
+
+To prevent unexpected redirect behavior in that case, Contour should ensure route name uniqueness.
+
+## Compatibility
```
sunjayBhatia, line 159: might be worth considering if there are other *Policy fields that this is not valid to or doesn't make sense to configure with on a HTTPProxy Route, e.g. RequestRedirectPolicy or DirectResponsePolicy

Jean-Daniel:
> One question I have: is support for the `allow_listed_routes` predicate a hard requirement for you? 

To be honest, I hesitate to remove it from the first draft at some point for this exact reason (route name requirement). When using trusted backend, it's already possible to limit the scope of the internal redirect using an unusual redirect code.

Having these use case covered will be enough for me, and if someone wants to extend the feature, it will always be possible  afterward.

skriss:
:+1: I think that's our best plan to move forward then.

design/internal-redirect_design.md
Jean-Daniel: At the Envoy level, `Route`, `RequestRedirect` and `DirectResponse` are 3 mutually exclusive members and `InternalRedirect` is a member of `Route`.

It should behave like any other member of `HTTPProxy.Route` that is ignored if `RequestRedirectPolicy` or `DirectResponsePolicy` policy is defined (`TimeoutPolicy`, `RetryPolicy`, …).

skriss (approved):
I don't have any further questions/comments, this looks implementable to me! Will hold for another maintainer. Thanks again @Jean-Daniel

design/internal-redirect_design.md
```diff
@@ -0,0 +1,159 @@
+# Internal Redirect
+
+Status: Draft
```
sunjayBhatia, line 3: can move this to Accepted!

sunjayBhatia (approved):
LGTM, nice work! looks clear and implementable

design/internal-redirect_design.md
Jean-Daniel: Done

sunjayBhatia:
thanks @Jean-Daniel! next steps will be to get an implementation PR up, feel free to put it as a draft if you want to get feedback, notes on testing, etc.

PR #4855, opened 2022-11-28
Create welcome-new-contributors.yaml

pnbrown:
Creates a GitHub action that will welcome new contributors the first time they file an issue or open a pull request.

Signed-off-by: nigel <nigelb@vmware.com>

### DIFF
--- a/.github/workflows/welcome-new-contributors.yaml
+++ b/.github/workflows/welcome-new-contributors.yaml
@@ -1,0 +1,26 @@
+name: 'Welcome New Contributors'
+
+on:
+  issues:
+    types: [opened]
+  pull_request:
+    types: [opened]
+
+jobs:
+  welcome-new-contributor:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Greet the contributor'
+        uses: garg3133/welcome-new-contributors@a38583ed8282e23d63d7bf919ca2d9fb95300ca6  # v1.2 
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          issue-message: >
+            Hey @contributor_name! Thanks for opening your first issue. We appreciate your contribution and welcome you to our community! 
+            We are glad to have you here and to have your input on Contour. 
+            You can also join us on [our mailing list](https://groups.google.com/g/project-contour) and [in our channel](https://kubernetes.slack.com/archives/C8XRH2R4J) 
+            in the [Kubernetes Slack Workspace](https://communityinviter.com/apps/kubernetes/community)
+          pr-message: >
+            Hi @contributor_name! Welcome to our community and thank you for opening your first Pull Request. 
+            Someone will review it soon. Thank you for committing to making Contour better.
+            You can also join us on [our mailing list](https://groups.google.com/g/project-contour) and [in our channel](https://kubernetes.slack.com/archives/C8XRH2R4J) 
+            in the [Kubernetes Slack Workspace](https://communityinviter.com/apps/kubernetes/community)


.github/workflows/welcome-new-contributors.yaml
```diff
@@ -0,0 +1,25 @@
...
+    types: [opened]
+  pull_request_target:
+    types: [opened]
+
+jobs:
+  welcome-new-contributor:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Greet the contributor'
+        uses: garg3133/welcome-new-contributors@v1.2
```
tsaarni, line 14: Is this particular action needed? I ask because there seems to be similar action from "verified creator" here https://github.com/marketplace/actions/first-interaction, which might also be better maintained than https://github.com/garg3133/welcome-new-contributors which has not seen updates anymore. The "first-interaction" action seems to lack the `@contributor_name` macro though.

.github/workflows/welcome-new-contributors.yaml
```diff
@@ -0,0 +1,25 @@
...
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          issue-message: >
+            Hey @contributor_name! Thanks for opening your first issue. We appreciate your contribution and welcome you to our community! 
+            We are glad to have you here and to have your input on Contour. 
+            You can also join us on [our mailing list](https://groups.google.com/g/project-contour) and [in our channel](https://kubernetes.slack.com/archives/C8XRH2R4J) 
+            in the [Kubernetes Slack Workspace](https://communityinviter.com/apps/kubernetes/community)
+          pr-message: >
+            Hi @contributor_name! And thank you for opening your first Pull Request. Someone will review it soon. Thank you for committing to making Contour better.
```
tsaarni, line 23: Should we have "welcome to our community" sentence in PRs message too?

tsaarni (commented):
Looks good to me! Just couple of questions.

.github/workflows/welcome-new-contributors.yaml
pnbrown: That's a good question. I'll look into it. I've previously used the currently linked one in other repositories.

.github/workflows/welcome-new-contributors.yaml
pnbrown: Thanks for the callout. Added.

tsaarni (approved):

.github/workflows/welcome-new-contributors.yaml
pnbrown: I think adding a little personal touch of having the user's name is worth using the older action. It seems to stand alone well enough and appears to lack a need to be updated.

What are your thoughts?


An alternate solution is to fork that repository. Would that be preferable?

tsaarni: I'm ok with this action, but since it is given `GITHUB_TOKEN` it would be bit safer if we could pin the hash. In that case the developer cannot publish a new version with the same tag, that does something bad.

Right now I don't find particularly good instructions for pinning action versions but here is one https://infra.apache.org/github-actions-policy.html.

pnbrown: So in this case we would pin to this release?
https://github.com/garg3133/welcome-new-contributors/commit/a38583ed8282e23d63d7bf919ca2d9fb95300ca6

garg3133/welcome-new-contributors@a38583ed8282e23d63d7bf919ca2d9fb95300ca6

pnbrown: Ahhh. Awesome. Looks like GitHub did its magic on mine and shortened it. I'll push a new commit.

.github/workflows/welcome-new-contributors.yaml
```diff
@@ -0,0 +1,26 @@
+name: 'Welcome New Contributors'
+
+on:
+  issues:
+    types: [opened]
+  pull_request_target:
```
tsaarni, line 6: Sorry @pnbrown for coming back with one more late comment, but I noticed `pull_request_target` which is suggested in https://github.com/garg3133/welcome-new-contributors but it is discouraged in context of `GITHUB_TOKEN` in various places: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target and https://securitylab.github.com/research/github-actions-preventing-pwn-requests/.  I'm not too familiar with the details (still reading) and it might not apply here since there is no checkout in this workflow, but just a heads up.

tsaarni: From what I'm reading, this use seems safe to me since the workflow does not run untrusted code (other than the action itself). However this is the first use of `pull_request_target` in this repo. The other similar action ([first-interaction](https://github.com/marketplace/actions/first-interaction)) seems to suggest `pull_request` instead of `pull_request_target`, see example [here](https://github.com/actions/first-interaction/blob/main/.github/workflows/first-interaction.yml). 

The problem with `pull_request_target` seems to be that`GITHUB_TOKEN` will have _write access_ to the repo, unlike the `GITHUB_TOKEN` which is issued in case of `pull_request`. I'm unsure if `pull_request_target` is really needed in order to comment on an issue and pull request? I guess the use of `pull_request_target` could be just mistake on [welcome-new-contributors](https://github.com/garg3133/welcome-new-contributors) docs.

pnbrown: It looks like the other uses just pull_request

https://github.com/actions/first-interaction/blob/main/.github/workflows/first-interaction.yml

tsaarni: Nice 👍   Do we have someone to create their first issue/pull request, to check that the action will still work? 😄

pnbrown: I'm asking around 😅

skriss: @tsaarni thanks for digging into this, I've been reading through the links you provided. From my read (particularly based on https://securitylab.github.com/research/github-actions-preventing-pwn-requests/), it seems like we will need to use `pull_request_target` to allow comments on PRs opened from forks, but I'm also happy to try it using `pull_request` first to see if it does work since the docs aren't 100% clear.

If we do need to end up switching back to `pull_request_target`, I agree that this usage seems safe since it's not executing any untrusted code. We could also use https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions to pare down the write permissions that are granted to just issues and PRs.

skriss (approved):
LGTM, let's see if this works with the `pull_request` trigger.

pnbrown:
It didn't work https://github.com/projectcontour/contour/pull/4860

pnbrown:
https://github.com/projectcontour/contour/actions/runs/3572754189/jobs/6008787066#step:2:1

tsaarni:
I see `Error: Resource not accessible by integration` :( 

I think there must be some middle ground by giving some permissions more, but still not full privileges... I'll experiment a bit myself with a different action that does not require creating a new github account on every round :-D

pnbrown:
😊 sounds good.

Yeah. I should have posted the error message. My bad.

pnbrown:
This should work for issues for now though

tsaarni:
There seems to be something magical in github tokens since in my test it worked with just `pull_request` alone, but I think there is different permissions in play depending on where the PR is coming from. I only tested creating PRs from/to my own repo so it does not reflect the real word.

I don't want to block adding this but I wonder could you possibly still try one more thing: adding write permission explicitly for `issues` like this

```yaml
jobs:
  welcome-new-contributor:
    runs-on: ubuntu-latest
    permissions:
      issues: write
```

skriss:
@tsaarni your vigilance here is much appreciated! I'm :+1: to using `pull_request_target` along with explicit `permissions` to limit to the minimal privileges needed (probably read+write on both issues and prs), and to adding warnings/links to the workflow file

PR #4872, opened 2022-11-29
update to cert-manager v1.10.1

skriss:


skriss:
Wasn't sure dependabot could handle this one given the module name change

sunjayBhatia:
> Wasn't sure dependabot could handle this one given the module name change

yeah looks like its failing on this one

skriss:
> > Wasn't sure dependabot could handle this one given the module name change
> 
> yeah looks like its failing on this one

Where do you see that?

PR #4873, opened 2022-11-29
Add trivy scan workflow

sunjayBhatia:
Runs on push to main and release branches, weekly, and can be triggered manually

Uploads a code scan result similar to CodeQL

### DIFF
--- a/.github/workflows/trivy-scan.yaml
+++ b/.github/workflows/trivy-scan.yaml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: aquasecurity/trivy-action@master
+      - uses: aquasecurity/trivy-action@v0.8.0
         with:
           scan-type: 'fs'
           format: 'sarif'
--- a/.github/workflows/trivy-scan.yaml
+++ b/.github/workflows/trivy-scan.yaml
@@ -25,7 +25,7 @@ jobs:
           format: 'sarif'
           output: 'trivy-results.sarif'
           ignore-unfixed: true
-          severity: 'HIGH'
+          severity: 'HIGH,CRITICAL'
       - uses: github/codeql-action/upload-sarif@v2
         with:
           sarif_file: 'trivy-results.sarif'
--- a/.github/workflows/trivy-scan.yaml
+++ b/.github/workflows/trivy-scan.yaml
@@ -1,0 +1,31 @@
+name: Trivy Scan
+
+on:
+  push:
+    # Run on main and supported release branches
+    branches:
+      - main
+      - release-1.23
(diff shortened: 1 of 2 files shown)


.github/workflows/trivy-scan.yaml
```diff
@@ -0,0 +1,31 @@
...
+  push:
+    # Run on main and supported release branches
+    branches:
+      - main
+      - release-1.23
+      - release-1.22
+      - release-1.21
+  # Run weekly
+  schedule:
+    - cron: '0 12 * * 1'
```
skriss, line 13: Unfortunate that we can't schedule runs against the release branches -- docs say "Scheduled workflows run on the latest commit on the default or base branch".

.github/workflows/trivy-scan.yaml
```diff
@@ -0,0 +1,31 @@
...
+  # Allow manual runs
+  workflow_dispatch:
+
+jobs:
+  trivy-scan:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: aquasecurity/trivy-action@master
```
skriss, line 22: any particular reason to use `master` vs the latest tagged release (https://github.com/aquasecurity/trivy-action/releases/tag/0.8.0)?

.github/workflows/trivy-scan.yaml
sunjayBhatia: yeah we could maybe try something that has a matrix of branch names to do the checkout/scan/upload steps and only have it run on a schedule and manually rather than on push?

sunjayBhatia: or have multiple workflows which is a little unfortunate

skriss: Yeah that sounds reasonable, I'm fine getting this in as an initial step though since we can run manually.

.github/workflows/trivy-scan.yaml
```diff
@@ -0,0 +1,31 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: aquasecurity/trivy-action@v0.8.0
+        with:
+          scan-type: 'fs'
+          format: 'sarif'
+          output: 'trivy-results.sarif'
+          ignore-unfixed: true
+          severity: 'HIGH'
```
skriss, line 28: should this be 'CRITICAL,HIGH'? (ref. https://github.com/aquasecurity/trivy-action#scan-ci-pipeline)

sunjayBhatia: ah yeah good catch, thought it was this level and higher, but should include all

skriss (approved):
🚀

PR #4874, opened 2022-11-29
release-1.23: Bump ginkgo and gomega to get transitive deps updated

sunjayBhatia:


skriss:
Is this all that's needed to get a clear scan on this branch?

sunjayBhatia:
> Is this all that's needed to get a clear scan on this branch?

I *think* so, these two modules have explicit dependencies on new versions of x/text and x/net

skriss (approved):
Yep, looks like (assuming I'm running it correctly) this change clears the scan.

PR #4876, opened 2022-11-29
trivy: Only perform vulnerability scanning

sunjayBhatia:
github should already be checking repo for secrets etc. and trivy is finding test fixtures as issues.

An alternative to this would be to skip directories containing secrets/fixtures using the skip-dirs option.


skriss (approved):

sunjayBhatia:
merging w/ powers since this is a workflow change

PR #4877, opened 2022-11-29
trivy: Add scan to 1.23 release branch

sunjayBhatia:
Needed so we can trigger workflow manually

### DIFF
--- a/.github/workflows/trivy-scan.yaml
+++ b/.github/workflows/trivy-scan.yaml
@@ -1,0 +1,32 @@
+name: Trivy Scan
+
+on:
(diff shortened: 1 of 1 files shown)


sunjayBhatia:
this would be the alternative we need to do for all supported release branches to enable running the action manually (vs. having a single action that runs on a schedule/manually that checks out the various branches itself)

skriss (approved):

sunjayBhatia:
still messing around with a matrix version

sunjayBhatia:
closing since we like the matrix approach better

PR #4879, opened 2022-11-30
trivy: Use matrix/ref checkout in scheduled runs

sunjayBhatia:
Removes run on push and alternative that adds workflow to all release branches

We may not need to run the scan on every commit really so this is a more passive alternative

### DIFF
--- a/.github/workflows/trivy-scan.yaml
+++ b/.github/workflows/trivy-scan.yaml
@@ -1,13 +1,6 @@
 name: Trivy Scan
 
 on:
-  push:
-    # Run on main and supported release branches
-    branches:
-      - main
-      - release-1.23
-      - release-1.22
-      - release-1.21
   # Run weekly
   schedule:
     - cron: '0 12 * * 1'
@@ -16,9 +9,18 @@ on:
 
 jobs:
   trivy-scan:
+    strategy:
+      matrix:
+        branch:
+        - main
+        - release-1.23
+        - release-1.22
+        - release-1.21
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          ref: ${{ matrix.branch }}
       - uses: aquasecurity/trivy-action@0.8.0
         with:
           security-checks: vuln


skriss (approved):
:+1: yeah I think this is probably the most useful configuration

skriss:
I guess the one question would be whether GitHub will understand which branch alerts are coming from when using this approach -- i.e. we don't want issues on `release-1.21` showing up as issues for `main` in https://github.com/projectcontour/contour/security/code-scanning. Not totally sure off the top of my head how GitHub keeps track of which branch results are for.

tsaarni:
> Not totally sure off the top of my head how GitHub keeps track of which branch results are for.

I have not tried this and the sarif file itself does not have this info, but it seems the codeql upload action might [get the ref](https://github.com/github/codeql-action/blob/160613c380ece8936349330e7b7b51ba4ea53698/src/actions-util.ts#L497-L505) and put it in [uploaded payload](https://github.com/github/codeql-action/blob/160613c380ece8936349330e7b7b51ba4ea53698/src/upload-lib.ts#L349-L360), in `ref` parameter ([doc](https://docs.github.com/en/rest/code-scanning?apiVersion=2022-11-28#upload-an-analysis-as-sarif-data))

> ref string Required
> The full Git reference, formatted as `refs/heads/<branch name>`, `refs/pull/<number>/merge`, or `refs/pull/<number>/head`

tsaarni (approved):

sunjayBhatia:
still messing with this on my fork to see if it shows issues correctly, so far it is not, see:
- workflow run: https://github.com/sunjayBhatia/contour/actions/runs/3579346286
- 1.21 release branch should have some issues: https://github.com/sunjayBhatia/contour/security/code-scanning?query=is%3Aopen+branch%3Arelease-1.21

sunjayBhatia:
It does show in an individual issue that the affected branches are the release-* ones on my fork: https://github.com/sunjayBhatia/contour/security/code-scanning/10 so maybe this is working as expected, gonna try it out

PR #4882, opened 2022-11-30
Dependency updates for release-1.22

sunjayBhatia:
Bump go, golang.org/x/net+text, github.com/emicklei/go-restful manually


sunjayBhatia:
flake downloading kustomize, will re-run

skriss:
Does it make sense to bump ginkgo/gomega here like you did for release-1.23?

sunjayBhatia:
> Does it make sense to bump ginkgo/gomega here like you did for release-1.23?

yeah we could def do that, though the thought was to be a little more conservative on the older releases since the version changes in those tools was broader/to avoid debugging something unrelated

skriss:
> > Does it make sense to bump ginkgo/gomega here like you did for release-1.23?
> 
> yeah we could def do that, though the thought was to be a little more conservative on the older releases since the version changes in those tools was broader/to avoid debugging something unrelated

Fair enough. I see there's no explicit version override for `golang.org/x/net` so that must be getting bumped as a result of one of the other ones?

sunjayBhatia:
> > > Does it make sense to bump ginkgo/gomega here like you did for release-1.23?
> > 
> > 
> > yeah we could def do that, though the thought was to be a little more conservative on the older releases since the version changes in those tools was broader/to avoid debugging something unrelated
> 
> Fair enough. I see there's no explicit version override for `golang.org/x/net` so that must be getting bumped as a result of one of the other ones?

yeah the x/net bump makes the x/text bump a no-op

skriss (approved):
LGTM, upgrade test failures looked like probably flakes to me

PR #4885, opened 2022-11-30
release-1.21: Bump module version and transitive deps

sunjayBhatia:
Bumped golang.org/x/crypto+net+text


skriss (approved):
LGTM pending CI

.golangci.yml
```diff
@@ -47,7 +47,8 @@ issues:
   # And we can't nolint them by line, because https://github.com/golangci/golangci-lint/issues/741
   - linters:
     - staticcheck
-    text: "SA1019: package github.com/golang/protobuf"
+    text: >-
+      SA1019: "github.com/golang/protobuf
```
skriss, line 51: this change looks a little funny, the lint errors I'm seeing seem to match the old text?

sunjayBhatia: yeah this change was for a newer version of golangci-lint, just downgraded again so its back to what it was 🤞🏽 this is it

test/scripts/run-gateway-conformance.sh
```diff
@@ -40,7 +40,10 @@ spec:
   controllerName: projectcontour.io/gateway-controller
 EOF
 
+GO_MOD_GATEWAY_API_VERSION=$(grep "sigs.k8s.io/gateway-api" go.mod | awk '{print $2}')
+
 cd $(mktemp -d)
 git clone https://github.com/kubernetes-sigs/gateway-api
 cd gateway-api
+git checkout "${GO_MOD_GATEWAY_API_VERSION}"
 go test ./conformance -gateway-class=contour
```
skriss, line 49: Oof, looks like release-1.21 uses Gateway API v0.4.3: https://github.com/projectcontour/contour/blob/release-1.21/go.mod#L43

Which doesn't have conformance tests: https://github.com/kubernetes-sigs/gateway-api/tree/v0.4.3/conformance

Which I guess was why we were running them from `main` back then.

Probably OK to just drop the gateway-conformance run entirely?

sunjayBhatia: yeah I spent a little too long trying to figure out what version would make this work and be what we were using back then, but not super important tbh, so gonna drop it

PR #4886, opened 2022-12-01
Use dependabot to keep github actions up to date

sunjayBhatia:
Should help with action deprecation by being proactive about updating them


skriss (approved):
nice!

PR #4889, opened 2022-12-01
Bump Envoy to 1.22.6

sunjayBhatia:
Release notes should be populated here eventually: https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.22/v1.22.6


skriss (approved):
LGTM, I think the upgrade test failure is just a flake that we've subsequently fixed

PR #4892, opened 2022-12-02
patch release 1.21.2: Docs update

sunjayBhatia:


versions.yaml
```diff
@@ -52,8 +52,19 @@ versions:
...
     supported: "true"
+    dependencies:
+      envoy: "1.22.6"
+      kubernetes:
+        - "1.23"
+        - "1.22"
+        - "1.21"
+      gateway-api:
+        - v1alpha2
+      contour-operator: v1.21.2
```
sunjayBhatia, line 65: assuming we will still release an operator, but that might be a bit of more work to do

sunjayBhatia: not sure how out of date CI etc. is there

skriss (approved):

sunjayBhatia:
merging w/ powers since this is just a docs update

PR #4905, opened 2022-12-06
Use generics for Gateway API pointer helpers

sunjayBhatia:
The `internal/gatewayapi` package has a lot of helper functions to get pointers to Gateway API types that are based on primitives.

We can use a generic helper function to implement this instead:

```
func To[T any](v T) *T {
       return &v
}
```

e.g. `ref.To(<some primitive value>)`

This came from the fact that Gateway API 0.6.0 review is considering removing the helpers added to the API, see: https://github.com/kubernetes-sigs/gateway-api/pull/1538/files/a5c0a509d597cc9a99e91f4629257105ad536877#r1026842150

We never started using them, but we can just get rid of the code we had in favor of a generics approach

The first commit shows a smaller example set of removals, the second finishes it up


internal/ref/ref.go
```diff
@@ -0,0 +1,22 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package contour contains the translation business logic that listens
+// to Kubernetes ResourceEventHandler events and translates those into
+// additions/deletions in caches connected to the Envoy xDS gRPC API server.
```
skriss, line 16: can rm this

sunjayBhatia: oops good catch, copy paste error

skriss (changes requested):
One nit, otherwise LGTM, nice cleanup! As discussed, after this goes in we can also look to get rid of our dependency on `k8s.io/utils/pointer`.

skriss (approved):

Issue #4359, opened 2022-02-22
Ability to disable Envoy adding server headers to responses


youngnick:
From #3692, @LiddersUK:
> My security team and the external pen testers continue to raise this as an (admittedly low impact) issue as if an attacker knows the technology that we employ then it gives them an attack vector.
> Their recommendation is for us to remove any server banners from response headers.

According to envoyproxy/envoy#14421, it seems like we will need to set the `server_header_transformation` param on the HTTPConnectionManager top `PASS_THROUGH`.

This will require a new global config boolean that defaults to `true` that will allow disabling of Envoy's server header responses.

chris93111:
Hi any update of this ?

Vishal-Chdhry:
Hi @sunjayBhatia! I would love to implement this, can you please guide me on where to start?

Vishal-Chdhry:
@sunjayBhatia, Sorry for the delay on this issue, i was busy in college exams.
 
why cant we directly use an enum, i am having trouble with figuring out where it should be an enum and where should i convert it to the boolean type.

I have to add it as an enum here
https://github.com/projectcontour/contour/blob/057bac9a7180ab87a80c6436cc8d22f8de43f0ba/internal/envoy/v3/listener.go#L150

and convert it to an enum as well probably by doing something like this
https://github.com/projectcontour/contour/blob/057bac9a7180ab87a80c6436cc8d22f8de43f0ba/internal/envoy/v3/listener.go#L239

I have opened a PR on this as well

PR #4906, opened 2022-12-07
Ability to disable Envoy adding server headers to responses

Vishal-Chdhry:
Signed-off-by: Vishal Choudhary <contactvishaltech@gmail.com>

adds `server_header_transformation` on HTTPConnectionManager
Fixes: #4359


Vishal-Chdhry:
Screenshot 2022-12-07 at 23 57 32

`make generate` gives me this error as well

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -336,6 +336,14 @@ type EnvoyListenerConfig struct {
 	// +optional
 	DisableMergeSlashes *bool `json:"disableMergeSlashes,omitempty"`
 
+	// DisableServerHeaderTransformation signifies we will not modify the Server header.
```
sunjayBhatia, line 339:
```suggestion
	// DisableServerHeaderTransformation signifies Envoy will not modify the Server header.
	// If set to false, the Server header will be set to "envoy". If set to true, the Server header
	// returned from the backend will not be modified.
```

I think you can remove the next two lines with the specific Envoy enum fields, we don't need to expose those to the user

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -336,6 +336,14 @@ type EnvoyListenerConfig struct {
...
+	// false: OVERWRITE
+	// true:	PASS_THROUGH
+	//
+	// Contour's default is false.
+	// +optional
+	DisableServerHeaderTransformation *bool `json:"DisableServerHeaderTransformation,omitempty"`
```
sunjayBhatia, line 345:
```suggestion
	DisableServerHeaderTransformation *bool `json:"disableServerHeaderTransformation,omitempty"`
```

pkg/config/parameters.go
```diff
@@ -485,6 +485,9 @@ type Parameters struct {
 	// which strips duplicate slashes from request URL paths.
 	DisableMergeSlashes bool `yaml:"disableMergeSlashes,omitempty"`
 
+	// DisableServerHeaderTransformation signifies we will not modify the Server header.
+	DisableServerHeaderTransformation bool `yaml:"DisableServerHeaderTransformation,omitempty"`
```
sunjayBhatia, line 489:
```suggestion
	DisableServerHeaderTransformation bool `yaml:"disableServerHeaderTransformation,omitempty"`
```

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -336,6 +336,14 @@ type EnvoyListenerConfig struct {
 	// +optional
 	DisableMergeSlashes *bool `json:"disableMergeSlashes,omitempty"`
 
+	// DisableServerHeaderTransformation signifies Envoy will not modify the Server header.
+	// false: OVERWRITE
+	// true:	PASS_THROUGH
+	//
+	// Contour's default is false.
+	// +optional
+	DisableServerHeaderTransformation *bool `json:"disableServerHeaderTransformation,omitempty"`
```
skriss, line 345: @sunjayBhatia given that this is an enum in Envoy, what do you think about making it the same here, with the default being "OVERWRITE"? My worry is that if we make it a bool, but then someone wants the "APPEND_IF_ABSENT" behavior, then we need to change the field type to accomodate.

sunjayBhatia: yep thats a fair point, makes the questions about plumbing a boolean through easier as well

skriss: @Vishal-Chdhry let's go with the above approach, thanks!

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -336,6 +337,13 @@ type EnvoyListenerConfig struct {
 	// +optional
 	DisableMergeSlashes *bool `json:"disableMergeSlashes,omitempty"`
 
+	// DisableServerHeaderTransformation signifies Envoy will not modify the Server header.
+	// It has 3 possible values: OVERWRITE, PASS_THROUGH, APPEND_IF_ABSENT
+	//
+	// Contour's default is OVERWRITE.
+	// +optional
+	DisableServerHeaderTransformation *http.HttpConnectionManager_ServerHeaderTransformation `json:"disableServerHeaderTransformation,omitempty"`
```
skriss, line 345: Let's change the name of this field to `ServerHeaderTransformation` since it's now an enum rather than a bool (including all godoc, JSON tags, etc)

skriss, line 345: This field should be structured like [DNSLookupFamily](https://github.com/projectcontour/contour/blob/5e3b212a990ee7d52a744291e19ad4cd97af6e43/apis/projectcontour/v1alpha1/contourconfig.go#L528-L565), i.e. with a Contour-defined "enum" type. Please format the godoc similarly as well, and include an explanation on the behavior of each option.

pkg/config/parameters.go
```diff
@@ -489,6 +490,9 @@ type Parameters struct {
 	// which strips duplicate slashes from request URL paths.
 	DisableMergeSlashes bool `yaml:"disableMergeSlashes,omitempty"`
 
+	// DisableServerHeaderTransformation signifies Envoy will not modify the Server header.
+	DisableServerHeaderTransformation http.HttpConnectionManager_ServerHeaderTransformation `yaml:"disableServerHeaderTransformation,omitempty"`
```
skriss, line 494: Please rename this field as well

skriss, line 494: Please also structure this field similarly to [DNSLookupFamily](https://github.com/projectcontour/contour/blob/5e3b212a990ee7d52a744291e19ad4cd97af6e43/pkg/config/parameters.go#L79-L95)

skriss (commented):
- Also need to add the new field to  https://github.com/projectcontour/contour/blob/main/site/content/docs/main/configuration.md for the config file
- We'll need to add some unit testing as well, can add pointers in the next round of review

skriss:
@Vishal-Chdhry just wanted to let you know that we are planning to cut a release candidate for 1.24 ~next Monday, so if possible it'd be great to get this updated by then.

Vishal-Chdhry:
> @Vishal-Chdhry just wanted to let you know that we are planning to cut a release candidate for 1.24 ~next Monday, so if possible it'd be great to get this updated by then.

sure, i will be on it right away

Vishal-Chdhry:
@skriss how should I add the unit tests?

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -336,6 +336,18 @@ type EnvoyListenerConfig struct {
...
+	// When configured as OVERWRITE, Overwrites any Server header with the contents of server_name.
+	// When configured as APPEND_IF_ABSENT, ⁣If no Server header is present, append Server server_name If a Server header is present, pass it through.
+	// When configured as PASS_THROUGH, ⁣Pass through the value of the server header, and do not append a header if none is present.
+	//
+	// Values: `OVERWRITE` (default), `APPEND_IF_ABSENT`, `PASS_THROUGH`
+	//
+	// Other values will produce an error.
+	// Contour's default is OVERWRITE.
+	// +optional
+	ServerHeaderTransformation ServerHeaderTransformationType `json:"ServerHeaderTransformation,omitempty"`
```
skriss, line 349:
```suggestion
	ServerHeaderTransformation ServerHeaderTransformationType `json:"serverHeaderTransformation,omitempty"`
```

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -532,6 +544,21 @@ const (
 	AllClusterDNSFamily ClusterDNSFamilyType = "all"
 )
 
+// ServerHeaderTransformation defines the action to be applied to the Server header on the response path
+type ServerHeaderTransformationType int32
```
skriss, line 548: Let's make this a string:
```suggestion
type ServerHeaderTransformationType string
```

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -532,6 +544,21 @@ const (
...
+
+const (
+	// Overwrite any Server header with the contents of server_name.
+	// This is the default value
+	OVERWRITE ServerHeaderTransformationType = 0
```
skriss, line 553: Let's format these constants like the following (please make similar changes for the other consts as well):
```suggestion
	// Overwrite any Server header with "envoy".
	// This is the default value.
	OverwriteServerHeader ServerHeaderTransformationType = "overwrite"
```

pkg/config/parameters.go
```diff
@@ -93,6 +93,22 @@ const IPv4ClusterDNSFamily ClusterDNSFamilyType = "v4"
 const IPv6ClusterDNSFamily ClusterDNSFamilyType = "v6"
 const AllClusterDNSFamily ClusterDNSFamilyType = "all"
 
+// ServerHeaderTransformation defines the action to be applied to the Server header on the response path
+type ServerHeaderTransformationType int32
```
skriss, line 97: Similar changes as for contourconfiguration.go:
```suggestion
type ServerHeaderTransformationType string
```

pkg/config/parameters.go
```diff
@@ -93,6 +93,22 @@ const IPv4ClusterDNSFamily ClusterDNSFamilyType = "v4"
...
+func (s ServerHeaderTransformationType) Validate() error {
+	switch s {
+	case OVERWRITE, APPEND_IF_ABSENT, PASS_THROUGH:
+		return nil
+	default:
+		return fmt.Errorf("invalid server header transformation %q", s)
+	}
+}
+
+const OVERWRITE ServerHeaderTransformationType = 0
```
skriss, line 108: Again, please make similar changes for the other constants:
```suggestion
const OverwriteServerHeader ServerHeaderTransformationType = "overwrite"
```

pkg/config/parameters.go
```diff
@@ -489,6 +505,14 @@ type Parameters struct {
 	// which strips duplicate slashes from request URL paths.
 	DisableMergeSlashes bool `yaml:"disableMergeSlashes,omitempty"`
 
+	// Defines the action to be applied to the Server header on the response path
+	// When configured as OVERWRITE, Overwrites any Server header with the contents of server_name.
+	// When configured as APPEND_IF_ABSENT, ⁣If no Server header is present, append Server server_name If a Server header is present, pass it through.
+	// When configured as PASS_THROUGH, ⁣Pass through the value of the server header, and do not append a header if none is present.
+	//
+	// Contour's default is OVERWRITE.
+	ServerHeaderTransformation ServerHeaderTransformationType `yaml:"ServerHeaderTransformation,omitempty"`
```
skriss, line 514:
```suggestion
	ServerHeaderTransformation ServerHeaderTransformationType `yaml:"serverHeaderTransformation,omitempty"`
```

site/content/docs/main/configuration.md
```diff
@@ -77,7 +77,8 @@ Where Contour settings can also be specified with command-line flags, the comman
...
 | disableAllowChunkedLength | boolean                | `false`                                                                                              | If this field is true, Contour will disable the RFC-compliant Envoy behavior to strip the `Content-Length` header if `Transfer-Encoding: chunked` is also set. This is an emergency off-switch to revert back to Envoy's default behavior in case of failures.
 | disableMergeSlashes       | boolean                | `false`                                                                                              | This field disables Envoy's non-standard
-merge_slashes path transformation behavior that strips duplicate slashes from request URL paths.
+| serverHeaderTransformation       | string                | `OVERWRITE`                                                                                              | This field defines the action to be applied to the Server header on the response path. Values: `OVERWRITE` (default), `APPEND_IF_ABSENT`, `PASS_THROUGH`
+merge_slashes path transformation behavior that strips duplicate slashes from request URL paths.
```
skriss, line 81: Looks like this line needs to move up above the `serverHeaderTransformation` entry

skriss (changes requested):

skriss:
Unit tests to add:
- https://github.com/projectcontour/contour/blob/main/cmd/contour/servecontext_test.go#L365 (new test case for this field)
- https://github.com/projectcontour/contour/blob/main/internal/envoy/v3/listener_test.go#L529 (new test case for this field)
- https://github.com/projectcontour/contour/blob/main/internal/xdscache/v3/listener_test.go#L143 (new test case for this field)
- https://github.com/projectcontour/contour/blob/main/pkg/config/parameters_test.go (new test function testing the Validate method)
- https://github.com/projectcontour/contour/blob/main/internal/featuretests/v3/listeners_test.go#L1245-L1297 (should be able to follow the pattern used for the highlighted test to add a new test for this field)

Vishal-Chdhry:
> Unit tests to add:

@skriss done

internal/xdscache/v3/listener.go
```diff
@@ -125,6 +125,9 @@ type ListenerConfig struct {
 	// MergeSlashes toggles Envoy's non-standard merge_slashes path transformation option for all listeners.
 	MergeSlashes bool
 
+	// ServerHeaderTransformation signifies we will not modify the Server header.
```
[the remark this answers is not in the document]
skriss:
```suggestion
ServerHeaderTransformation defines the action to be applied to the Server header on the response path.
```

internal/envoy/v3/listener.go
```diff
@@ -241,6 +243,20 @@ func (b *httpConnectionManagerBuilder) MergeSlashes(enabled bool) *httpConnectio
...
+func (b *httpConnectionManagerBuilder) ServerHeaderTransformation(value contour_api_v1alpha1.ServerHeaderTransformationType) *httpConnectionManagerBuilder {
+	switch value {
+	case contour_api_v1alpha1.OverwriteServerHeader:
+		b.serverHeaderTransformation = http.HttpConnectionManager_OVERWRITE
+	case contour_api_v1alpha1.AppendIfAbsentServerHeader:
+		b.serverHeaderTransformation = http.HttpConnectionManager_APPEND_IF_ABSENT
+	case contour_api_v1alpha1.PassThroughServerHeader:
+		b.serverHeaderTransformation = http.HttpConnectionManager_PASS_THROUGH
+	default:
+		return nil
```
skriss, line 255: You can drop the default case here, it's causing panics when no server header transformation is configured. We want to return `b` here which is already being done below.

internal/xdscache/v3/listener_test.go
```diff
@@ -2631,6 +2632,61 @@ func TestListenerVisit(t *testing.T) {
 				SocketOptions: envoy_v3.TCPKeepaliveSocketOptions(),
 			}),
 		},
+		"httpproxy with server_header_transformation set to overwrite in listener config": {
+			ListenerConfig: ListenerConfig{
+				ServerHeaderTransformation: v1alpha1.OverwriteServerHeader,
```
skriss, line 2637: Please test a non-default value here.

internal/envoy/v3/listener_test.go
```diff
@@ -1182,6 +1183,57 @@ func TestHTTPConnectionManager(t *testing.T) {
 				},
 			},
 		},
+		"server header transform set to overwrite": {
+			routename:                 "default/kuard",
+			accesslogger:              FileAccessLogEnvoy("/dev/stdout", "", nil, v1alpha1.LogLevelInfo),
+			serverHeaderTranformation: v1alpha1.OverwriteServerHeader,
```
skriss, line 1189: Please test a non-default value here.

internal/featuretests/v3/listeners_test.go
```diff
@@ -1295,3 +1295,57 @@ func TestHTTPProxyXffNumTrustedHops(t *testing.T) {
 		TypeUrl: listenerType,
 	})
 }
+
+func TestHTTPProxyServerHeaderTransformation(t *testing.T) {
+	rh, c, done := setup(t, func(conf *xdscache_v3.ListenerConfig) {
+		conf.ServerHeaderTransformation = contour_api_v1alpha1.OverwriteServerHeader
```
skriss, line 1301: Please test a non-default value here

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -336,6 +336,18 @@ type EnvoyListenerConfig struct {
 	// +optional
 	DisableMergeSlashes *bool `json:"disableMergeSlashes,omitempty"`
 
+	// Defines the action to be applied to the Server header on the response path
+	// When configured as OVERWRITE, Overwrites any Server header with the contents of server_name.
+	// When configured as APPEND_IF_ABSENT, ⁣If no Server header is present, append Server server_name If a Server header is present, pass it through.
+	// When configured as PASS_THROUGH, ⁣Pass through the value of the server header, and do not append a header if none is present.
+	//
+	// Values: `OVERWRITE` (default), `APPEND_IF_ABSENT`, `PASS_THROUGH`
```
skriss, line 344: @sunjayBhatia WDYT about casing here of the values here? For DNS lookup family we lowercased, but I don't see existing multi-word options. I guess we could switch these to lowercase, but honestly not sure if there's any point in doing so.

changelogs/unreleased/4906-Vishal-Chdhry-small.md
```diff
@@ -0,0 +1 @@
+adds server_header_transformation on HTTPConnectionManager
```
skriss, line 1: Please update this to a `minor` changelog with some more info:

```
## Enable configuring Server header transformation 

Envoy's treatment of the Server header on responses can now be configured in the Contour config file or ContourConfiguration CRD.
When configured as `OVERWRITE`, Envoy overwrites any Server header with "envoy".
When configured as `APPEND_IF_ABSENT`, ⁣if a Server header is present, Envoy will pass it through, otherwise, it will set it to "envoy".
When configured as `PASS_THROUGH`, Envoy passes through the value of the Server header and does not append a header if none is present.
```

skriss (changes requested):
@Vishal-Chdhry one more round of minor feedback and then I think this is about ready to merge.

skriss:
Looks like you need to re-run `make generate` and commit the results as well.

apis/projectcontour/v1alpha1/contourconfig.go
Vishal-Chdhry: actually it's an integer in envoy

sunjayBhatia: hm yeah as you say many if not all the other relevant `const`s in this package and `pkg/config` we define are lowercase, so for consistency we could switch to that, since we're not really going to pass the value straight through to Envoy it doesn't really matter

skriss (approved):
LGTM other than the enum value casing change, which we can change in a follow-up PR so we can get this in. Thanks for the work on this @Vishal-Chdhry!

skriss:
I pushed fixes so we can get this in for the RC.

skriss (approved):
LGTM, will wait for @sunjayBhatia to take one more look

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -336,6 +336,18 @@ type EnvoyListenerConfig struct {
 	// +optional
 	DisableMergeSlashes *bool `json:"disableMergeSlashes,omitempty"`
 
+	// Defines the action to be applied to the Server header on the response path
+	// When configured as overwrite, overwrites any Server header with the contents of server_name.
+	// When configured as append_if_absent, ⁣If no Server header is present, append Server server_name If a Server header is present, pass it through.
+	// When configured as pass_through, pPass through the value of the server header, and do not append a header if none is present.
```
sunjayBhatia, line 342:
```suggestion
	// When configured as pass_through, pass through the value of the server header, and do not append a header if none is present.
```

pkg/config/parameters.go
```diff
@@ -489,6 +505,14 @@ type Parameters struct {
 	// which strips duplicate slashes from request URL paths.
 	DisableMergeSlashes bool `yaml:"disableMergeSlashes,omitempty"`
 
+	// Defines the action to be applied to the Server header on the response path
+	// When configured as overwrite, overwrites any Server header with the contents of server_name.
```
sunjayBhatia, line 509: similar comment about `server_name`

sunjayBhatia (approved):

PR #4908, opened 2022-12-07
Implement support for ResponseHeaderModifier

### LINKED ISSUE #4819: Gateway API: support ResponseHeaderModifier filter (extended/experimental)
https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteFilter

Should be fairly straight-forward since we already have support for this via HTTPProxy, and it's very similar to the RequestHeaderModifier logic.

sunjayBhatia:
One note: Host header is allowed in response modifier filter since it is not significant in a response

Enables Gateway API conformance test for response header modifier, bumps echoserver image in e2e tests, and adds tests for filter per-backend since this is missing from upstream tests (should add this soon).

Fixes #4819

### DIFF
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -1060,10 +1060,9 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1beta1.HTTPRou
 
 		// Process rule-level filters.
 		var (
-			requestHeaderPolicy, responseHeaderPolicy             *HeadersPolicy
-			requestHeaderModifierSeen, responseHeaderModifierSeen bool
-			redirect                                              *gatewayapi_v1beta1.HTTPRequestRedirectFilter
-			mirrorPolicy                                          *MirrorPolicy
+			requestHeaderPolicy, responseHeaderPolicy *HeadersPolicy
+			redirect                                  *gatewayapi_v1beta1.HTTPRequestRedirectFilter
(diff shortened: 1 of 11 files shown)


internal/dag/gatewayapi_processor.go
```diff
@@ -1060,10 +1060,10 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1beta1.HTTPRou
 
 		// Process rule-level filters.
 		var (
-			headerPolicy       *HeadersPolicy
-			headerModifierSeen bool
-			redirect           *gatewayapi_v1beta1.HTTPRequestRedirectFilter
-			mirrorPolicy       *MirrorPolicy
+			requestHeaderPolicy, responseHeaderPolicy             *HeadersPolicy
+			requestHeaderModifierSeen, responseHeaderModifierSeen bool
```
skriss, line 1064: not directly related to your PR but seems like we could drop the bools and just nil-check the policy vars

skriss (approved):
LGTM, one optional nit.

May also make sense to hold this until after we merge dev-gwapi-0.6 into main, so it can keep its own changelog.

skriss:
Looks like it needs a kick to re-trigger CI..

skriss (approved):

skriss:
This will close #4819, right? Can update https://github.com/projectcontour/contour/pull/4908#issue-1482900640 if so.

PR #4930, opened 2022-12-15
Remove usage of k8s utils package (including pointer* helpers)

sunjayBhatia:
Replaces all usage with our own generic based helpers

Also removes usage of clock/fake clock interface since it was not being used with any effect in testing

### DIFF
--- a/internal/status/gatewayclassconditions_test.go
+++ b/internal/status/gatewayclassconditions_test.go
@@ -19,8 +19,6 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	realclock "k8s.io/utils/clock"
-	fakeclock "k8s.io/utils/clock/testing"
 	gatewayapi_v1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 )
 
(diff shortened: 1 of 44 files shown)


skriss (approved):
Nice cleanup! 🚀

sunjayBhatia:
taking a look again at that status package

internal/status/gatewayclassconditions_test.go
```diff
@@ -180,23 +170,27 @@ func TestMergeConditions(t *testing.T) {
...
-	// Simulate the passage of time between original condition creation
-	// and update processing
-	fakeClock.SetTime(later)
-
 	for _, tc := range testCases {
 		got := mergeConditions(tc.current, tc.updates...)
-		if conditionChanged(tc.expected[0], got[0]) {
-			assert.Equal(t, tc.expected, got, tc.name)
-		}
+		assert.Equal(t, tc.expected, got, tc.name)
```
sunjayBhatia, line 193: rather than use `conditionChanged` which doesnt check the transition time, just use `assert.Equal` to check equality

also means as per how the tests were originally set up, we needed to change the implementation to only update the transition time if the status changed

internal/status/gatewayclassconditions.go
```diff
@@ -61,7 +61,10 @@ func mergeConditions(conditions []metav1.Condition, updates ...metav1.Condition)
 					conditions[j].Reason = update.Reason
 					conditions[j].Message = update.Message
 					conditions[j].ObservedGeneration = update.ObservedGeneration
-					conditions[j].LastTransitionTime = update.LastTransitionTime
+					// Only update the transition time if Status changes.
+					if cond.Status != update.Status {
```
skriss, line 65: Makes sense and consistent with the godoc for `LastTransitionTime`, which says "lastTransitionTime is the last time the condition transitioned from one status to another."

Also just noting for reference that https://github.com/projectcontour/contour/blob/main/internal/k8s/helpers.go#L29 is used when determining whether to actually write status updates to the apiserver, to filter out transition time-only changes. But it's an improvement to not have changes like that in the first place.

skriss (approved):

sunjayBhatia:
upgrade flake downloading 1.23.2 manifest, will rerun

PR #4931, opened 2022-12-15
internal/dag: Fix comparison of include match conditions

### LINKED ISSUE #4907: Root proxy is invalid with DuplicateMatchConditions error when there are multiple inclusions with header match conditions
Root proxy gets into invalid state with DuplicateMatchConditions error when there are multiple inclusions with header match conditions. The header matches have multiple conditions, and few of these conditions are same across includes. But as all match conditions in include are ANDed, all the match conditions under include should be considered for detecting duplicate match conditions, and not just individual ones. Refer below for an example.

Since match conditions under include are ANDed, code should check for uniqueness of the ANDed condition and across all includes not just the consecutive one. The logic used here to detect duplicates seems incorrect: https://github.com/projectcontour/contour/blob/main/internal/dag/httpproxy_processor.go#L1383

NOK
====
Root HTTPProxy with below spec goes to invalid state with DuplicateMatchConditions error.

spec:
  includes:
  - conditions:
    - prefix: /offer1
    name: child1
    namespace: offer1
  - conditions:
    - header:
        exact: xxx
        name: x-header
    - header:
        exact: yyy
        name: y-header
    name: child1
    namespace: offer1
  - conditions:
    - header:
        exact: xxx
        name: x-header
    - header:
        exact: zzz
        name: z-header
    name: child1
    namespace: offer1

Status:
  Conditions:
    Errors:
      Message:             duplicate conditions defined on an include
      Reason:              DuplicateMatchConditions
      Status:              True
      Type:                IncludeError
    Last Transition Time:  2022-12-07T19:14:58Z
    Message:               At least one error present, see Errors for details
    Observed Generation:   1
    Reason:                ErrorPresent
    Status:                False
    Type:                  Valid
  Current Status:          invalid
  Description:             At least one error present, see Errors for details

OK
====
Root HTTPProxy with similar spec with rearranged includes (so those with header matches are not next to each other) is in valid state.

spec:
  includes:
  - conditions:
    - header:
        exact: xxx
        name: x-header
    - header:
        exact: yyy
        name: y-header
    name: child1
    namespace: offer1
  - conditions:
    - prefix: /offer1
    name: child1
    namespace: offer1
  - conditions:
    - header:
        exact: xxx
        name: x-header
    - header:
        exact: zzz
        name: z-header
    name: child1
    namespace: offer1

Status:
  Conditions:
    Last Transition Time:  2022-12-07T19:25:10Z
    Message:               Valid HTTPProxy
    Observed Generation:   1
    Reason:                Valid
    Status:                True
    Type:                  Valid
  Current Status:          valid
  Description:             Valid HTTPProxy

**Environment:**

- Contour version: 1.23

sunjayBhatia:
Previously includeMatchConditionsIdentical was just comparing adjacent includes and comparing conditions element by element rather than as a whole, ANDed together

Fixes #4907


sunjayBhatia:
might do another pass on the implementation to see if we can do something more elegant

currently uses the existing conversions to dag types for match conditions that are then sorted and stringified so they can be compared

since this happens before the match conditions are deemed valid or not, this is a little odd maybe

internal/dag/httpproxy_processor.go
```diff
@@ -1381,17 +1380,21 @@ func toStringSlice(hvs []contour_api_v1.CORSHeaderValue) []string {
...
-	j := 0
-	for i := 1; i < len(includes); i++ {
-		// Now compare each include's set of conditions
-		for _, cA := range includes[i].Conditions {
-			for _, cB := range includes[j].Conditions {
-				if (cA.Prefix == cB.Prefix) && equality.Semantic.DeepEqual(cA.Header, cB.Header) {
-					return true
-				}
-			}
+	seenIncludeConditions := map[string]struct{}{}
```
skriss, line 1383: I was never crazy about our implementation of converting sets of conditions to strings to check for equality. I think we'd probably be able to come up with some adversarial inputs that would cause it to fail, since we're not using "safe" delimiters. But, practically speaking, it probably gets the job done for most normal inputs. I don't have an issue with using the DAG structs here, better than having to duplicate the stringification logic.

sunjayBhatia: another option after working on this for a little, still messy but no string comparisons:

```go
	seenConds := map[string][][]HeaderMatchCondition{}
	for _, include := range includes {
		pathPrefix := mergePathMatchConditions(include.Conditions).Prefix

		includeHeaderConds := mergeHeaderMatchConditions(include.Conditions)
		sort.SliceStable(includeHeaderConds, func(i, j int) bool {
			if includeHeaderConds[i].MatchType != includeHeaderConds[j].MatchType {
				return includeHeaderConds[i].MatchType < includeHeaderConds[j].MatchType
			}
			if includeHeaderConds[i].Invert != includeHeaderConds[j].Invert {
				return includeHeaderConds[i].Invert
			}
			if includeHeaderConds[i].Name != includeHeaderConds[j].Name {
				return includeHeaderConds[i].Name < includeHeaderConds[j].Name
			}
			if includeHeaderConds[i].Value != includeHeaderConds[j].Value {
				return includeHeaderConds[i].Value < includeHeaderConds[j].Value
			}
			return false
		})

		seenHeaderConds, pathSeen := seenConds[pathPrefix]
		if !pathSeen {
			seenConds[pathPrefix] = [][]HeaderMatchCondition{
				includeHeaderConds,
			}
			continue
		}

		for _, headerConds := range seenHeaderConds {
			if len(headerConds) != len(includeHeaderConds) {
				seenConds[pathPrefix] = append(seenConds[pathPrefix], includeHeaderConds)
				continue
			}

			headerCondsIdentical := true
			for i := range headerConds {
				if headerConds[i] != includeHeaderConds[i] {
					seenConds[pathPrefix] = append(seenConds[pathPrefix], includeHeaderConds)
					headerCondsIdentical = false
					break
				}
			}
			if headerCondsIdentical {
				return true
			}
		}
	}

	return false
```

sunjayBhatia: a little better than comparing each condition set for an include to each and every other one, but the header conditions comparison still has to collect all unique sets and compare them

sunjayBhatia:
from https://github.com/projectcontour/contour/pull/4931/commits/37edb30c7511ff454e144a8b8e170721b6a5786a

```
The included proxies that come before we find a duplicate are now not
marked as orphaned, which is a small behavior change.
```

maybe not desirable, would be good to hear what people thing about this

skriss:
> from [37edb30](https://github.com/projectcontour/contour/commit/37edb30c7511ff454e144a8b8e170721b6a5786a)
> 
> ```
> The included proxies that come before we find a duplicate are now not
> marked as orphaned, which is a small behavior change.
> ```
> 
> maybe not desirable, would be good to hear what people thing about this

If I'm reading it correctly, if we do end up finding a duplicate set of conditions on any two included proxies, then we end up not programming routes for any of the includes on the root HTTPProxy. Is that right? In that case, it does seem to me like all of those included proxies that didn't end up getting programmed should probably be called orphaned.

skriss:
Possibly an easy fix would be on https://github.com/projectcontour/contour/blob/dc0d593f4174cbc7fc88bbae04072fa3d6b5db83/internal/dag/httpproxy_processor.go#L675 instead of deleting from `p.orphaned`, just keep track of a list of child proxies that we want to eventually remove from `p.orphaned`, then if we get to https://github.com/projectcontour/contour/blob/dc0d593f4174cbc7fc88bbae04072fa3d6b5db83/internal/dag/httpproxy_processor.go#L677 we know we didn't return early due to duplicate conditions, so go ahead and delete from `p.orphaned` there

sunjayBhatia:
> If I'm reading it correctly, if we do end up finding a duplicate set of conditions on any two included proxies, then we end up not programming routes for any of the includes on the root HTTPProxy. Is that right? In that case, it does seem to me like all of those included proxies that didn't end up getting programmed should probably be called orphaned.

ah good point, I was kinda thinking of the orphaned relationship similar to the Gateway/Route `attachedRoutes` idea so it might be ok, however in this case orphaned means a bit more than just the reference between the resources is good but that config has been generated, which you're right in this case will not be

skriss:
> > If I'm reading it correctly, if we do end up finding a duplicate set of conditions on any two included proxies, then we end up not programming routes for any of the includes on the root HTTPProxy. Is that right? In that case, it does seem to me like all of those included proxies that didn't end up getting programmed should probably be called orphaned.
> 
> ah good point, I was kinda thinking of the orphaned relationship similar to the Gateway/Route `attachedRoutes` idea so it might be ok, however in this case orphaned means a bit more than just the reference between the resources is good but that config has been generated, which you're right in this case will not be

Yeah I think in this case it's probably safest to stick to the existing semantics, where child proxies that have an invalid include of them are called orphaned, rather than some form of attached + invalid.

sunjayBhatia:
we can preserve some more of the includes/routes etc. that are valid by not returning early which would mean some traffic is still able to be served: https://github.com/projectcontour/contour/pull/4931/commits/3ff0eaab09de8c1579afd407a93f8cec77e379e9

internal/dag/builder_test.go
```diff
@@ -10913,7 +10945,26 @@ func TestDAGInsert(t *testing.T) {
 			objs: []interface{}{
 				proxy108, proxy108a, proxy108b, s1, s12, s13,
 			},
-			want: listeners(),
```
skriss, line 10916: I'd call out this behavior change explicitly in the changelog (i.e. that the first include with a given set of conditions is programmed/not orphaned, only subsequent ones with dupe conditions are now not programmed).

internal/dag/status_test.go
```diff
@@ -1534,6 +1534,12 @@ func TestDAGStatus(t *testing.T) {
 				Conditions: []contour_api_v1.MatchCondition{{
 					Prefix: "/blog",
 				}},
+			}, {
```
skriss, line 1537: It'd be good to keep the original test case around too, i.e. a "simple" inclusion of two different child proxies with duplicate conditions, where now the first one is still programmed, and the second one is marked orphaned/not programmed.

skriss (changes requested):

internal/dag/builder_test.go
sunjayBhatia: let me know what you think!

skriss (approved):
thanks for the updates, LGTM!

PR #4933, opened 2022-12-16
Update Gateway API to 0.6.0-rc2

sunjayBhatia:
Unfortunately includes a bump to controller-runtime and k8s API etc. since Gateway API bumped those as dependencies as well.

Can split them off into their own PRs if we need to.


skriss:
LGTM once lint errors are addressed

sunjayBhatia:
ah looks like might be blocked on https://github.com/projectcontour/contour/pull/4937

skriss:
Found https://github.com/kubernetes/kubernetes/pull/114585 upstream, so maybe we'll get an upstream patch for this

I guess if we have to we could hack our `generate-crds` step to insert the needed YAML patch..

sunjayBhatia:
> Found [kubernetes/kubernetes#114585](https://github.com/kubernetes/kubernetes/pull/114585) upstream, so maybe we'll get an upstream patch for this
> 
> I guess if we have to we could hack our `generate-crds` step to insert the needed YAML patch..

oh sweet good find, yeah might have to do that in the meantime

skriss:
> > Found [kubernetes/kubernetes#114585](https://github.com/kubernetes/kubernetes/pull/114585) upstream, so maybe we'll get an upstream patch for this
> > I guess if we have to we could hack our `generate-crds` step to insert the needed YAML patch..
> 
> oh sweet good find, yeah might have to do that in the meantime

Yeah that works for me for now and we'll keep an eye on the upstream PR

skriss (approved):

PR #4956, opened 2023-01-04
internal/envoy: fix no router filter crash

### LINKED ISSUE #4840: [Internal] Potential problem of listener.go#AddFilter

I am looking at the source code and making some private changes. Actually it is adding a custom http filter in 
[`internal\xdscache\v3\listener.go`](https://github.com/projectcontour/contour/blob/2ba1559ee5746c8616f6ec37f0905411f4bb03e4/internal/xdscache/v3/listener.go#L436) just like this:

```go
//...
cm := envoy_v3.HTTPConnectionManagerBuilder().
	Codec(envoy_v3.CodecForVersions(cfg.DefaultHTTPVersions...)).
	AddFilter(envoy_v3.FilterMisdirectedRequests(vh.VirtualHost.Name)).
        // Here is my custom filter, I add it here follow the FilterRewriteHostForCafe filter above
	AddFilter(envoy_v3.MyCustomHTTPFilter()).
	DefaultFilters().
	AddFilter(authFilter).
	RouteConfigName(path.Join("https", vh.VirtualHost.Name)).
//...
```

But got `Index out of range` exception when running (sorry I forget to capture the log but it's very easy to reproduce). The   exception was thrown from [`internal\envoy\v3\listener.go#AddFilter` ](https://github.com/projectcontour/contour/blob/2ba1559ee5746c8616f6ec37f0905411f4bb03e4/internal/envoy/v3/listener.go#L381) as follows:

```go

func (b *httpConnectionManagerBuilder) AddFilter(f *http.HttpFilter) *httpConnectionManagerBuilder {
	if f == nil {
		return b
	}

	b.filters = append(b.filters, f)

        // The first call of this function return directly without checking router filter, so it works well.
	if len(b.filters) == 1 {
		return b
	}

	lastIndex := len(b.filters) - 1
	routerIndex := -1
	for i, filter := range b.filters {
		if filter.GetTypedConfig().MessageIs(&envoy_extensions_filters_http_router_v3.Router{}) {
			routerIndex = i
			break
		}
	}

	if routerIndex != -1 && f.GetTypedConfig().MessageIs(&envoy_extensions_filters_http_router_v3.Router{}) {
		panic("Can't add more than one router to a filter chain")
	}

	if routerIndex != lastIndex {
                // Here we got an "index out of range" exception when the routerFilter is not found, because the routerIndex=-1
		routerFilter := b.filters[routerIndex]
		b.filters = append(b.filters[:routerIndex], b.filters[routerIndex+1])
		b.filters = append(b.filters, routerFilter)
	}

	return b
}

```

So if the router filter is not found, it will throw the exception. I found the router filter is added by `DefaultFilters()` function in [`internal\xdscache\v3\listener.go`](https://github.com/projectcontour/contour/blob/2ba1559ee5746c8616f6ec37f0905411f4bb03e4/internal/xdscache/v3/listener.go#L436). Which means the `AddFilter` function called before `DefaultFilter` won't find the router filter such that will throw an exception. But logically the `FilterMisdirectedRequests` is still OK because the first call of `AddFilter` doesn't check the router filter.

Maybe we should specify the order bethween `AddFilter` and `DefaultFilter`, or modify the `AddFilter` function to prevent this potential problem.

**Environment:**

- Contour version: latest

izturn:

fix #4840

### DIFF
--- a/internal/envoy/v3/listener.go
+++ b/internal/envoy/v3/listener.go
@@ -368,19 +368,20 @@ func (b *httpConnectionManagerBuilder) AddFilter(f *http.HttpFilter) *httpConnec
 		}
 	}
 
-	// We can't add more than one router entry, and there should be no way to do it.
-	// If this happens, it has to be programmer error, so we panic to tell them
-	// it needs to be fixed. Note that in hitting this case, it doesn't matter we added
-	// the second one earlier, because we're panicking anyway.
-	if routerIndex != -1 && f.GetTypedConfig().MessageIs(&envoy_extensions_filters_http_router_v3.Router{}) {
-		panic("Can't add more than one router to a filter chain")
-	}
-
-	if routerIndex != -1 && routerIndex != lastIndex {
-		// Move the router to the end of the filters array.
-		routerFilter := b.filters[routerIndex]
-		b.filters = append(b.filters[:routerIndex], b.filters[routerIndex+1])
-		b.filters = append(b.filters, routerFilter)
+	if routerIndex != -1 {
+		// We can't add more than one router entry, and there should be no way to do it.
+		// If this happens, it has to be programmer error, so we panic to tell them
+		// it needs to be fixed. Note that in hitting this case, it doesn't matter we added
+		// the second one earlier, because we're panicking anyway.
+		if f.GetTypedConfig().MessageIs(&envoy_extensions_filters_http_router_v3.Router{}) {
+			panic("Can't add more than one router to a filter chain")
+		}
+		if routerIndex != lastIndex {
+			// Move the router to the end of the filters array.
+			routerFilter := b.filters[routerIndex]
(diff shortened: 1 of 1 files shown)


internal/envoy/v3/listener.go
```diff
@@ -376,7 +376,7 @@ func (b *httpConnectionManagerBuilder) AddFilter(f *http.HttpFilter) *httpConnec
 		panic("Can't add more than one router to a filter chain")
 	}
 
-	if routerIndex != lastIndex {
+	if routerIndex != -1 && routerIndex != lastIndex {
```
sunjayBhatia, line 379: could make line 375 `if routerIndex != -1 {` and have that if wrap that one and this one to make it so its a little less redundant

izturn: done

sunjayBhatia (approved):

Issue #4324, opened 2022-02-06
design: Update ExtensionService to split out each type of extension


youngnick:
After working on #399 for a while, @skriss suggested (in https://github.com/projectcontour/contour/pull/3802#discussion_r660084154) that we look at updating ExtensionService to have a separate CRD for each type of extension service.

This issue is to cover that work.

From https://github.com/projectcontour/contour/pull/3802#discussion_r660084154:
> I think if we moved to a CRD per service type (ExtAuthService, RateLimitService, TracingService), we could make this work since each type could be significantly different - and tracing would need to be significantly different than the others given that it's not an xDS gRPC service, and, depending on the provider, does not make use of an Envoy cluster at all (which is effectively what an ExtensionService is right now). However, I think it's going to be difficult to move from the generic ExtensionService to a CRD per service type since it'll require what I believe is a breaking API change for ExtAuth (HTTPProxy currently has an implicitly-typed reference to an ExtensionService), and disruptive/breaking config file changes for rate limiting. If we wanted to do that, it might need to wait for a v2 API/v2 Contour.
>Another option would be to add type-specific subfields to ExtensionService; however, because tracing is so different from the existing ExtensionService types, many of the existing fields in ExtensionService don't apply to tracing, so we end up with a messy API where many fields just don't apply in certain circumstances. I'm not a huge fan of this option given where we're starting from -- even though ExtensionService is "alpha", we don't want to break users if we can avoid it, which makes it harder to make in-place changes.
>We could introduce just a TracingService CRD in the v1alpha1 group which would give us a chance to experiment with having a CRD per service type, and then look to migrate the other existing service types to specific CRDs down the road. I like this approach in theory, though there are some challenges specifically around OpenCensus since it can't be configured more than once per Envoy lifetime (that would apply even if we used the config file).

pavelnikolov:
This PR was sent last week and I am looking forward to having this merged and implemented in contour as well.
https://github.com/envoyproxy/envoy/pull/20281

sunjayBhatia:
at least in the context of OpenTelemetry and enabling tracic, this may not be necessary as a prerequisite for that work as the OpenTelemetry filter just takes a gRPC service so an ExtensionService based approach will still likely work

skriss:
> at least in the context of OpenTelemetry and enabling tracic, this may not be necessary as a prerequisite for that work as the OpenTelemetry filter just takes a gRPC service so an ExtensionService based approach will still likely work

[...]

PR #4957, opened 2023-01-04
design: add tracing design doc

### LINKED ISSUE #3792: write design doc for tracing support

yangyy93:
Add tracing design support to export tracing data to OpenTelemetry
ref #399 
ref #3802
close #3792 
@skriss @sunjayBhatia 

Signed-off-by: yy [yang.yang@daocloud.io]

### DIFF
--- a/design/tracing-design.md
+++ b/design/tracing-design.md
@@ -1,0 +1,170 @@
+# Design for Supporting Tracing in Contour
+
+Status: Draft
+
+## Abstract
+Envoy has rich support for [distributed tracing][1].
+However, Contour does not currently offer a way to configure Envoy for tracing.Exporting trace data to OpenTelemetry has been supported since envoy v1.23. This design document provides the implementation of exporting tracing data to [OpenTelemetry][2] in Contour.
+
+## Background
+Envoy's [documentation on tracing][1] says:
+> Distributed tracing allows developers to obtain visualizations of call flows in large service oriented architectures. It can be invaluable in understanding serialization, parallelism, and sources of latency.
+ 
+Envoy supports:
+- generating traces for incoming requests
+- joining existing traces when a client sets the `x-client-trace-id` header
+- exporting trace data to various third-party providers (Zipkin, Jaeger, Datadog, etc.)
+
+The tracing ecosystem is complex right now.
+However, a few observations can be made:
+- [OpenTelemetry][2] is a CNCF project which is working to become a standard in the space. It was formed as a merger of the OpenTracing and OpenCensus projects. It supports ingesting trace data in a variety of formats, transforming it, and exporting it to a variety of backends.
+- Added support for exporting tracing data to opentelemetry since envoy v1.23
+
+## Goals
+- **Support a single global trace configuration** that applies to all requests.
+- **Support Opentelemetry formats**, Because Opentelemetry is an observability standard. Data can be ingested, transformed, and sent to an observability backend.
+- **Support custom tags to send additional span data**, including getting data from environment variables, request headers, etc.
+
+## Non Goals
+- **Per-HTTPProxy trace configuration.** There is a technical limitation to enabling per-proxy configuration. Trace settings are configured on the HTTP connection manager (HCM). Each TLS-enabled virtual host uses its own HCM, but *all* non-TLS virtual hosts share a single HCM. As such, while it's possible to have unique trace settings for each TLS virtual host, it's not possible to do the same for non-TLS virtual hosts.  Note that this same limitation has come up when designing external authorization and global rate limiting as well.
+- **Trace formats other than OpenTelemetry.** While Contour *may* choose to add direct support for other trace formats in the future, our hope is that by supporting the OpenTelemetry collector, users are able to export traces to the backend of their choice without having to directly configure it in Contour/Envoy.
+
+## High-Level Design
+The Contour configuration file will be extended with a new optional `tracing` section. This configuration block, if present, will enable tracing and will define the trace format and other properties needed to generate and export trace data.
+
+The `OpenTelemetry` trace format will be supported.
+
+The parameters defined in the configuration file will be used to configure each [HTTP connection manager's `tracing` section][3].
+
+Since the tracing backend is pluggable, Contour will not package any particular backend.
+The user is responsible for deploying and operating the tracing backend of their choice, and configuring Contour to make use of it.
+
+## Detailed Design
+
+### otel-collector ExtensionService
+
+When using Tracing,first,an ExtensionService must be defined for the otel-collector with the cluster-level details for the otel-collector itself.
+For example:
+
+```yaml
+apiVersion: projectcontour.io/v1alpha1
+kind: ExtensionService
+metadata:
+  namespace: projectcontour
+  name: otel-collector
+spec:
+  protocol: h2c
+  services:
+    - name: otel-collector
+      port: 4317
+  timeoutPolicy:
+    response: 50ms
+```
+
+### Contour Configuration
+
+The `tracing` section of the Contour config file will look like:
+
+```yaml
+tracing:
+  # overallSampling defines the sampling rate of trace data.
+  overallSampling: 100
+  # maxPathTagLength defines maximum length of the request path to extract and include in the HttpUrl tag.
+  maxPathTagLength: 256
+  # customTags defines a list of custom tags with unique tag name.
+  # A customTag can only set one of literal, environment and requestHeaderName
+  customTags:
+    # tagName used to populate the tag name
+    - tagName: literal
+      # literal is a static custom tag value.
+      literal: 'this is literal'
+    - tagName: environment
+      # environment indicates that the label value is obtained from the environment variable.
+      environment: HOST
+    - tagName: requestHeaderName
+      # requestHeaderName indicates which request header the label value is obtained from.
+      requestHeaderName: ':path'
+  # extensionService Identifies the extension service defining the openTelemetry collector service.
+  # formatted as <namespace>/<name>.    
+  extensionService: projectcontour/otel-collector
+```
+
+### HCM Configuration
+
+In the presence of the `tracing` block above, all HTTP connection managers (HCM) will be configured for tracing.
+
+A Tracing HCM will include the following as an example:
+
+```go
+...
+Tracing: &http.HttpConnectionManager_Tracing{
+            OverallSampling: &envoy_type.Percent{
+                Value: 100,
+            },
+            MaxPathTagLength: wrapperspb.UInt32(256),
+            CustomTags:       []*envoy_trace_v3.CustomTag{
+                {
+                    Tag: "literal",
+                    Type: &envoy_trace_v3.CustomTag_Literal_{
+                        Literal: &envoy_trace_v3.CustomTag_Literal{
+                            Value: "this is literal",
+                        },
+                    },
+                },
+                {
+                    Tag: "environment",
+                    Type: &envoy_trace_v3.CustomTag_Environment_{
+                        Environment: &envoy_trace_v3.CustomTag_Environment{
+                            Name: "HOST",
+                        },
+                    },
+                },
+                {
+                    Tag: "requestHeaderName",
+                    Type: &envoy_trace_v3.CustomTag_RequestHeader{
+                        RequestHeader: &envoy_trace_v3.CustomTag_Header{
+                        Name:         ":path",
+                        },
+                    },
+                },
+            },
+            Provider: &envoy_config_trace_v3.Tracing_Http{
+                Name: "envoy.tracers.opentelemetry",
+                ConfigType: &envoy_config_trace_v3.Tracing_Http_TypedConfig{
+                    TypedConfig: protobuf.MustMarshalAny(&envoy_config_trace_v3.OpenTelemetryConfig{
+                        GrpcService: &envoy_core_v3.GrpcService{
+                            TargetSpecifier: &envoy_core_v3.GrpcService_EnvoyGrpc_{
+                                EnvoyGrpc: &envoy_core_v3.GrpcService_EnvoyGrpc{
+                                    ClusterName: "extension/projectcontour/otle-collector",
+                                    Authority:   "extension/projectcontour/otle-collector",
+                                },
+                            },
+                            Timeout: durationpb.New(time.Millisecond * 50),
+                        },
+                        ServiceName: "contour-envoy-otel.otel-collector",
+                    }),
+                },
+            },
+        },
+...
+```
+
+## Alternatives Considered
+If there are alternative high level or detailed designs that were not pursued they should be called out here with a brief explanation of why they were not pursued.
+
+## Security Considerations
+If this proposal has an impact to the security of the product, its users, or data stored or transmitted via the product, they must be addressed here.
+
+## Compatibility
+A discussion of any compatibility issues that need to be considered
+
+## Implementation
+A description of the implementation, timelines, and any resources that have agreed to contribute.
+
+## Open Issues
+- Envoy supports custom tracing sampling rate and custom label override hcm configuration in route - should contour be supported in the future?
+
+
+[1]: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/observability/tracing
+[2]: https://opentelemetry.io/
+[3]: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-msg-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-tracing
--- a/design/tracing-design.md
+++ b/design/tracing-design.md
@@ -1,6 +1,6 @@
 # Design for Supporting Tracing in Contour
 
-Status: Draft
+Status: Accepted
 
 ## Abstract
 Envoy has rich support for [distributed tracing][1].
@@ -25,7 +25,7 @@ However, a few observations can be made:
 ## Goals
 - **Support a single global trace configuration** that applies to all requests.
 - **Support OpenTelemetry formats**, Because OpenTelemetry is an observability standard. Data can be ingested, transformed, and sent to an observability backend.
-- **Support custom tags to send additional span data**, including getting data from environment variables, request headers, etc.
+- **Support custom tags to send additional span data**, including getting data from request headers, etc.
 
 ## Non Goals
 - **Per-HTTPProxy trace configuration.** There is a technical limitation to enabling per-proxy configuration. Trace settings are configured on the HTTP connection manager (HCM). Each TLS-enabled virtual host uses its own HCM, but *all* non-TLS virtual hosts share a single HCM. As such, while it's possible to have unique trace settings for each TLS virtual host, it's not possible to do the same for non-TLS virtual hosts.  Note that this same limitation has come up when designing external authorization and global rate limiting as well.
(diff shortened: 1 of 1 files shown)


design/tracing-design.md
```diff
@@ -0,0 +1,170 @@
...
+Envoy has rich support for [distributed tracing][1].
+However, Contour does not currently offer a way to configure Envoy for tracing.Exporting trace data to OpenTelemetry has been supported since envoy v1.23. This design document provides the implementation of exporting tracing data to [OpenTelemetry][2] in Contour.
+
+## Background
+Envoy's [documentation on tracing][1] says:
+> Distributed tracing allows developers to obtain visualizations of call flows in large service oriented architectures. It can be invaluable in understanding serialization, parallelism, and sources of latency.
+ 
+Envoy supports:
+- generating traces for incoming requests
+- joining existing traces when a client sets the `x-client-trace-id` header
```
pavelnikolov, line 15: Is this header allowed by default?

design/tracing-design.md
```diff
@@ -0,0 +1,170 @@
...
+      environment: HOST
+    - tagName: requestHeaderName
+      # requestHeaderName indicates which request header the label value is obtained from.
+      requestHeaderName: ':path'
+  # extensionService Identifies the extension service defining the openTelemetry collector service.
+  # formatted as <namespace>/<name>.    
+  extensionService: projectcontour/otel-collector
+```
+
+### HCM Configuration
```
sunjayBhatia, line 92: can move this down to the "Implementation" section

design/tracing-design.md
```diff
@@ -0,0 +1,170 @@
...
+If this proposal has an impact to the security of the product, its users, or data stored or transmitted via the product, they must be addressed here.
+
+## Compatibility
+A discussion of any compatibility issues that need to be considered
+
+## Implementation
+A description of the implementation, timelines, and any resources that have agreed to contribute.
+
+## Open Issues
+- Envoy supports custom tracing sampling rate and custom label override hcm configuration in route - should contour be supported in the future?
```
sunjayBhatia, line 165: this makes sense as a future enhancement/follow on after the initial implementation if it is too big a chunk to do in one go

design/tracing-design.md
```diff
@@ -0,0 +1,170 @@
...
+  # maxPathTagLength defines maximum length of the request path to extract and include in the HttpUrl tag.
+  maxPathTagLength: 256
+  # customTags defines a list of custom tags with unique tag name.
+  # A customTag can only set one of literal, environment and requestHeaderName
+  customTags:
+    # tagName used to populate the tag name
+    - tagName: literal
+      # literal is a static custom tag value.
+      literal: 'this is literal'
+    - tagName: environment
```
sunjayBhatia, line 81: how much will users in the context of Contour/Envoy use this? is it common to set custom environment variables on the Envoy container?

design/tracing-design.md
```diff
@@ -0,0 +1,170 @@
...
+                        ServiceName: "contour-envoy-otel.otel-collector",
+                    }),
+                },
+            },
+        },
+...
+```
+
+## Alternatives Considered
+If there are alternative high level or detailed designs that were not pursued they should be called out here with a brief explanation of why they were not pursued.
```
sunjayBhatia, line 153: pretty much covered in non-goals

design/tracing-design.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Security Considerations
+If this proposal has an impact to the security of the product, its users, or data stored or transmitted via the product, they must be addressed here.
+
+## Compatibility
+A discussion of any compatibility issues that need to be considered
```
sunjayBhatia, line 159: a discussion here about what tracing provider services the Envoy tracing filter is compatible with would be good to mention

this is a new opt-in feature that will be enabled on all http connection managers, so shouldn't have compatibility concerns outside of the above

skriss:
cc @rajatvig I know you've been interested in this feature as well, would appreciate your 👀 on this PR!

design/tracing-design.md
```diff
@@ -0,0 +1,170 @@
+# Design for Supporting Tracing in Contour
+
+Status: Draft
+
+## Abstract
+Envoy has rich support for [distributed tracing][1].
+However, Contour does not currently offer a way to configure Envoy for tracing.Exporting trace data to OpenTelemetry has been supported since envoy v1.23. This design document provides the implementation of exporting tracing data to [OpenTelemetry][2] in Contour.
```
skriss, line 7:
```suggestion
Exporting trace data to OpenTelemetry has been supported since envoy v1.23.
However, Contour does not currently offer a way to configure Envoy for tracing.
This design document provides the implementation of exporting tracing data to [OpenTelemetry][2] in Contour.
```

design/tracing-design.md
```diff
@@ -0,0 +1,170 @@
...
+ 
+Envoy supports:
+- generating traces for incoming requests
+- joining existing traces when a client sets the `x-client-trace-id` header
+- exporting trace data to various third-party providers (Zipkin, Jaeger, Datadog, etc.)
+
+The tracing ecosystem is complex right now.
+However, a few observations can be made:
+- [OpenTelemetry][2] is a CNCF project which is working to become a standard in the space. It was formed as a merger of the OpenTracing and OpenCensus projects. It supports ingesting trace data in a variety of formats, transforming it, and exporting it to a variety of backends.
+- Added support for exporting tracing data to opentelemetry since envoy v1.23
```
skriss, line 21:
```suggestion
- Envoy has had support for exporting tracing data to OpenTelemetry since v1.23.
```

design/tracing-design.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Goals
+- **Support a single global trace configuration** that applies to all requests.
+- **Support Opentelemetry formats**, Because Opentelemetry is an observability standard. Data can be ingested, transformed, and sent to an observability backend.
```
skriss, line 25:
```suggestion
- **Support OpenTelemetry formats**, Because OpenTelemetry is an observability standard. Data can be ingested, transformed, and sent to an observability backend.
```

design/tracing-design.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Non Goals
+- **Per-HTTPProxy trace configuration.** There is a technical limitation to enabling per-proxy configuration. Trace settings are configured on the HTTP connection manager (HCM). Each TLS-enabled virtual host uses its own HCM, but *all* non-TLS virtual hosts share a single HCM. As such, while it's possible to have unique trace settings for each TLS virtual host, it's not possible to do the same for non-TLS virtual hosts.  Note that this same limitation has come up when designing external authorization and global rate limiting as well.
+- **Trace formats other than OpenTelemetry.** While Contour *may* choose to add direct support for other trace formats in the future, our hope is that by supporting the OpenTelemetry collector, users are able to export traces to the backend of their choice without having to directly configure it in Contour/Envoy.
+
+## High-Level Design
+The Contour configuration file will be extended with a new optional `tracing` section. This configuration block, if present, will enable tracing and will define the trace format and other properties needed to generate and export trace data.
```
skriss, line 33:
```suggestion
The Contour configuration file and ContourConfiguration CRD will be extended with a new optional `tracing` section. This configuration block, if present, will enable tracing and will define the trace format and other properties needed to generate and export trace data.
```

design/tracing-design.md
```diff
@@ -0,0 +1,170 @@
...
+The parameters defined in the configuration file will be used to configure each [HTTP connection manager's `tracing` section][3].
+
+Since the tracing backend is pluggable, Contour will not package any particular backend.
+The user is responsible for deploying and operating the tracing backend of their choice, and configuring Contour to make use of it.
+
+## Detailed Design
+
+### otel-collector ExtensionService
+
+When using Tracing,first,an ExtensionService must be defined for the otel-collector with the cluster-level details for the otel-collector itself.
```
skriss, line 46:
```suggestion
When using tracing, the user must first define an ExtensionService with the cluster-level details for the otel-collector itself.
```

design/tracing-design.md
```diff
@@ -0,0 +1,170 @@
...
+            },
+            Provider: &envoy_config_trace_v3.Tracing_Http{
+                Name: "envoy.tracers.opentelemetry",
+                ConfigType: &envoy_config_trace_v3.Tracing_Http_TypedConfig{
+                    TypedConfig: protobuf.MustMarshalAny(&envoy_config_trace_v3.OpenTelemetryConfig{
+                        GrpcService: &envoy_core_v3.GrpcService{
+                            TargetSpecifier: &envoy_core_v3.GrpcService_EnvoyGrpc_{
+                                EnvoyGrpc: &envoy_core_v3.GrpcService_EnvoyGrpc{
+                                    ClusterName: "extension/projectcontour/otle-collector",
+                                    Authority:   "extension/projectcontour/otle-collector",
```
skriss, line 139:
```suggestion
                                    ClusterName: "extension/projectcontour/otel-collector",
                                    Authority:   "extension/projectcontour/otel-collector",
```

skriss (changes requested):
Thanks for the design doc @yangyy93! Some initial nits to improve readability but overall the direction of this looks good to me.

skriss:
cc @yoitsro I know you've been interested in this feature as well, would appreciate your 👀 on this PR!

design/tracing-design.md
sunjayBhatia: also do you happen to have a working end to end example/spike?

design/tracing-design.md
```diff
@@ -0,0 +1,170 @@
...
+                                },
+                            },
+                            Timeout: durationpb.New(time.Millisecond * 50),
+                        },
+                        ServiceName: "contour-envoy-otel.otel-collector",
```
rajatvig, line 144: This should not be the name of the collector service but a user defined service name via the configuration. The default could be the `ingressClassName` of Contour.

design/tracing-design.md
```diff
@@ -0,0 +1,170 @@
...
+  timeoutPolicy:
+    response: 50ms
+```
+
+### Contour Configuration
+
+The `tracing` section of the Contour config file will look like:
+
+```yaml
+tracing:
```
rajatvig, line 69: Add a `serviceName` option with default to the `ingressClassName` or just `contour`.

rajatvig (commented):
This is great. Added some comments

design/tracing-design.md
```diff
@@ -0,0 +1,170 @@
...
+## Detailed Design
+
+### otel-collector ExtensionService
+
+When using Tracing,first,an ExtensionService must be defined for the otel-collector with the cluster-level details for the otel-collector itself.
+For example:
+
+```yaml
+apiVersion: projectcontour.io/v1alpha1
+kind: ExtensionService
```
rajatvig, line 51: Thinking about the design a little more, it would be nice if there was a specific `TracingConfig` or a similar CRD instead of a generic `ExtensionService`.

That way the configuration portion (mentioned later) that currently is coded as part of Contour Configuration becomes one large CRD with its own lifecycle. 

It could be something like

```yaml
kind: TracingConfiguration
metadata:
  namespace: projectcontour
  # the name here will be used as the name of the service to report
  name: otel-collector
spec:
  # overallSampling defines the sampling rate of trace data.
  overallSampling: 100
  # maxPathTagLength defines maximum length of the request path to extract and include in the HttpUrl tag.
  maxPathTagLength: 256
  # customTags defines a list of custom tags with unique tag name.
  # A customTag can only set one of literal, environment and requestHeaderName
  customTags:
    # tagName used to populate the tag name
    - tagName: literal
      # literal is a static custom tag value.
      literal: 'this is literal'
    - tagName: environment
      # environment indicates that the label value is obtained from the environment variable.
      environment: HOST
    - tagName: requestHeaderName
      # requestHeaderName indicates which request header the label value is obtained from.
      requestHeaderName: ':path'
  service:
    protocol: h2c
    services:
      - name: otel-collector
        port: 4317
    timeoutPolicy:
      response: 50ms
```

The bit I am not sure with this however, how would it work multiple Contour installations. 
Would it be better to specify the `TracingConfiguration` name in the Contour Configuration block or specify the `ingressClassName` in the `TracingConfiguration`?

pavelnikolov: I think that service name, const labels, sampling, force tracing, traceID header etc might be configurable in Contour as a tracing section in the config and the otel-collector can be just an Extension service. This way each instance of Contour would have its own configuration that lives with it. Then the otel-collector can handle tracing specific configuration. So we might have the tracing config, but if the extension service is not defined, the tracing is just no-op.

design/tracing-design.md
sunjayBhatia: hm yeah looks like something that should possibly be configurable so traces can be identified how the user want to, will need to think about this more, the idea of having `contour-<ingressclassname>` is interesting for a default

design/tracing-design.md
yangyy93: ok

design/tracing-design.md
yangyy93: Obtaining the value of an environment variable is useful for some needs. For example, when there are multiple envoy instances, I can know which gateway generated the tracing through the HOST environment variable

design/tracing-design.md
```diff
@@ -0,0 +1,170 @@
...
+                },
+            },
+        },
+...
+```
+
+## Alternatives Considered
+If there are alternative high level or detailed designs that were not pursued they should be called out here with a brief explanation of why they were not pursued.
+
+## Security Considerations
```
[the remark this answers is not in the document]
yangyy93: ok

design/tracing-design.md
yangyy93: Yes, this is a good idea, serviceName becomes configurable, default value is contour-<ingressClassName>

design/tracing-design.md
yangyy93: Yes, this is a good idea, serviceName becomes configurable, default value is `contour-<ingressClassName>`

design/tracing-design.md
yangyy93: thanks for your suggestion

design/tracing-design.md
yangyy93: yes，It is allowed by default.
https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/runtime#config-http-conn-man-runtime-client-enabled

design/tracing-design.md
rajatvig: That works great.

skriss: We have definitely thought about splitting out a separate CRD per Extension type (see https://github.com/projectcontour/contour/issues/4324), that would include more than just the basic gRPC service/cluster info. However, in the interest of making progress here, I'd prefer to stick to the current structure we already have, where you have a generic ExtensionService to identify the otel collector, and then tracing-specific config lives in the config file/ContourConfiguration CRD. This matches the pattern we're using for global rate limiting, for example.

design/tracing-design.md
sunjayBhatia: is the persona configuring tracing tags etc. going to be the same as the persona deploying Envoy? wondering if we should consider that these are likely not the same people in many cases so it doesn't quite make sense for them to have to have this cross-functional knowledge of deployment details

this might be something we can offer some syntactic sugar for or do by default for users rather than giving the generic ability to add tags based on environment variables

yangyy93: Envoy itself supports obtaining the value of custom tags from environment variables. I think it is a good enhancement for us to provide this ability for users to use,and how to use it can be configured by the administrator.

design/tracing-design.md
yangyy93: The ingressClassName setting supported by contour is an array type. If the configured ingressClassName has multiple values, I don't think of a good way to deal with it, so I want to change the default value from `contour-<ingressClassName>` to `contour`.

design/tracing-design.md
wilsonwu: From some usecases, I think get tag from envs is good feature, for people who do the configurations for Contour/Envoy should have knowledge of this TracingConfiguration CRD, and most time for debug or issue tracking we need to know which Envoy container cause the problem, and if can't read tag from env, I think find pod name from trace is a hard work.

sunjayBhatia: providing the pod name/hostname in the trace labels makes a lot of sense, I'm just wondering if it is something we can provide automatically without exposing more surface area than we need to, in case there are secrets etc. in the environment

sunjayBhatia: one option would be to include the pod/hostname automatically for everyone, or offer a flag to enable it in the tag configuration e.g. `includePodDetails` that might add the name and namespace of the origin Envoy pod automatically, or even have a special tag type we can provide that is syntactic sugar for including origin envoy pod details

sunjayBhatia: this way also you don't have to redeploy Envoy if you want to modify what tags need to be set on the trace

skriss (approved):
no further comments from me, I think this is about ready to go once any outstanding threads from @sunjayBhatia are resolved. Thanks @yangyy93!

design/tracing-design.md
sunjayBhatia: Yep that makes sense @yangyy93

design/tracing-design.md
sunjayBhatia: if the suggestion [here](https://github.com/projectcontour/contour/pull/4957/files#r1094639206) works and removing the `environment` tag value source, this LGTM

should add a note we would like to include the envoy pod name/namespace in the default tags

sunjayBhatia: we can also implement that specific feature as a follow-up

yangyy93: This is a great suggestion, I'll update the design docs later

design/tracing-design.md
```diff
@@ -0,0 +1,181 @@
...
+
+The tracing ecosystem is complex right now.
+However, a few observations can be made:
+- [OpenTelemetry][2] is a CNCF project which is working to become a standard in the space. It was formed as a merger of the OpenTracing and OpenCensus projects. It supports ingesting trace data in a variety of formats, transforming it, and exporting it to a variety of backends.
+- Envoy has had support for exporting tracing data to OpenTelemetry since v1.23.
+
+## Goals
+- **Support a single global trace configuration** that applies to all requests.
+- **Support OpenTelemetry formats**, Because OpenTelemetry is an observability standard. Data can be ingested, transformed, and sent to an observability backend.
+- **Support custom tags to send additional span data**, including getting data from environment variables, request headers, etc.
```
sunjayBhatia, line 28:
```suggestion
- **Support custom tags to send additional span data**, including getting data from request headers, etc.
```

design/tracing-design.md
```diff
@@ -0,0 +1,181 @@
...
+  # includePodDetail defines a flag. If it is true, contour will add the pod name and namespace to the span of the trace. the default is true
+  includePodDetail: true
+  # serviceName defines a configurable service name.the default is contour
+  serviceName: contour
+  # overallSampling defines the sampling rate of trace data.
+  overallSampling: 100
+  # maxPathTagLength defines maximum length of the request path to extract and include in the HttpUrl tag.
+  maxPathTagLength: 256
+  # customTags defines a list of custom tags with unique tag name.
+  # A customTag can only set one of literal, environment and requestHeaderName
```
sunjayBhatia, line 82:
```suggestion
  # A customTag can only set one of literal or requestHeaderName
```

design/tracing-design.md
```diff
@@ -0,0 +1,181 @@
+# Design for Supporting Tracing in Contour
+
+Status: Draft
```
sunjayBhatia, line 3:
```suggestion
Status: Accepted
```

sunjayBhatia (approved):
LGTM, some tiny changes to make including marking this as "Accepted"!

design/tracing-design.md
yangyy93: ok,thanks

PR #4958, opened 2023-01-04
cmd/contour:  sort the option flags

### LINKED ISSUE #2397: Contour CLI help should sort the option flags.

When a Contour subcommand has a log of option flags, it is hard to find the one you want because the `--help` output doesn't sort them. Sorting the flags by name would make it easier to read and to find a flags in the category you are looking for.

izturn:

sort the option flags
fix #2397

### DIFF
--- a/cmd/contour/bootstrap.go
+++ b/cmd/contour/bootstrap.go
@@ -24,6 +24,8 @@ func registerBootstrap(app *kingpin.Application) (*kingpin.CmdClause, *envoy.Boo
 	var config envoy.BootstrapConfig
 
 	bootstrap := app.Command("bootstrap", "Generate bootstrap configuration.")
+	bootstrap.Arg("path", "Configuration file ('-' for standard output).").Required().StringVar(&config.Path)
+
 	bootstrap.Flag("admin-address", "Path to Envoy admin unix domain socket.").Default("/admin/admin.sock").StringVar(&config.AdminAddress)
 	bootstrap.Flag("admin-port", "DEPRECATED: Envoy admin interface port.").IntVar(&config.AdminPort)
 	bootstrap.Flag("dns-lookup-family", "Defines what DNS Resolution Policy to use for Envoy -> Contour cluster name lookup. Either v4, v6, auto, or all.").StringVar(&config.DNSLookupFamily)
@@ -32,10 +34,10 @@ func registerBootstrap(app *kingpin.Application) (*kingpin.CmdClause, *envoy.Boo
 	bootstrap.Flag("envoy-key-file", "Client key filename for Envoy secure xDS gRPC communication.").Envar("ENVOY_KEY_FILE").StringVar(&config.GrpcClientKey)
 	bootstrap.Flag("namespace", "The namespace the Envoy container will run in.").Envar("CONTOUR_NAMESPACE").Default("projectcontour").StringVar(&config.Namespace)
 	bootstrap.Flag("overload-max-heap", "Defines the maximum heap size in bytes until overload manager stops accepting new connections.").Uint64Var(&config.MaximumHeapSizeBytes)
-	bootstrap.Arg("path", "Configuration file ('-' for standard output).").Required().StringVar(&config.Path)
 	bootstrap.Flag("resources-dir", "Directory where configuration files will be written to.").StringVar(&config.ResourcesDir)
 	bootstrap.Flag("xds-address", "xDS gRPC API address.").StringVar(&config.XDSAddress)
 	bootstrap.Flag("xds-port", "xDS gRPC API port.").IntVar(&config.XDSGRPCPort)
 	bootstrap.Flag("xds-resource-version", "The versions of the xDS resources to request from Contour.").Default("v3").StringVar((*string)(&config.XDSResourceVersion))
+
 	return bootstrap, &config
 }
--- a/cmd/contour/bootstrap.go
+++ b/cmd/contour/bootstrap.go
@@ -24,18 +24,18 @@ func registerBootstrap(app *kingpin.Application) (*kingpin.CmdClause, *envoy.Boo
 	var config envoy.BootstrapConfig
 
 	bootstrap := app.Command("bootstrap", "Generate bootstrap configuration.")
-	bootstrap.Arg("path", "Configuration file ('-' for standard output).").Required().StringVar(&config.Path)
-	bootstrap.Flag("resources-dir", "Directory where configuration files will be written to.").StringVar(&config.ResourcesDir)
 	bootstrap.Flag("admin-address", "Path to Envoy admin unix domain socket.").Default("/admin/admin.sock").StringVar(&config.AdminAddress)
 	bootstrap.Flag("admin-port", "DEPRECATED: Envoy admin interface port.").IntVar(&config.AdminPort)
-	bootstrap.Flag("xds-address", "xDS gRPC API address.").StringVar(&config.XDSAddress)
-	bootstrap.Flag("xds-port", "xDS gRPC API port.").IntVar(&config.XDSGRPCPort)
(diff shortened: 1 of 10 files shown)


cmd/contour/bootstrap.go
```diff
@@ -24,18 +24,18 @@ func registerBootstrap(app *kingpin.Application) (*kingpin.CmdClause, *envoy.Boo
...
 	bootstrap.Flag("envoy-cert-file", "Client certificate filename for Envoy secure xDS gRPC communication.").Envar("ENVOY_CERT_FILE").StringVar(&config.GrpcClientCert)
 	bootstrap.Flag("envoy-key-file", "Client key filename for Envoy secure xDS gRPC communication.").Envar("ENVOY_KEY_FILE").StringVar(&config.GrpcClientKey)
 	bootstrap.Flag("namespace", "The namespace the Envoy container will run in.").Envar("CONTOUR_NAMESPACE").Default("projectcontour").StringVar(&config.Namespace)
-	bootstrap.Flag("xds-resource-version", "The versions of the xDS resources to request from Contour.").Default("v3").StringVar((*string)(&config.XDSResourceVersion))
-	bootstrap.Flag("dns-lookup-family", "Defines what DNS Resolution Policy to use for Envoy -> Contour cluster name lookup. Either v4, v6, auto, or all.").StringVar(&config.DNSLookupFamily)
 	bootstrap.Flag("overload-max-heap", "Defines the maximum heap size in bytes until overload manager stops accepting new connections.").Uint64Var(&config.MaximumHeapSizeBytes)
+	bootstrap.Arg("path", "Configuration file ('-' for standard output).").Required().StringVar(&config.Path)
```
skriss, line 35: This one is a positional argument, not a flag, so I'd keep it at the top separate from the flags.

cmd/contour/certgen.go
```diff
@@ -35,19 +35,18 @@ func registerCertGen(app *kingpin.Application) (*kingpin.CmdClause, *certgenConf
...
 	certgenApp.Flag("incluster", "Use in cluster configuration.").BoolVar(&certgenConfig.InCluster)
+	certgenApp.Flag("kube", "Apply the generated certs directly to the current Kubernetes cluster.").BoolVar(&certgenConfig.OutputKube)
 	certgenApp.Flag("kubeconfig", "Path to kubeconfig (if not in running inside a cluster).").Default(filepath.Join(os.Getenv("HOME"), ".kube", "config")).StringVar(&certgenConfig.KubeConfig)
 	certgenApp.Flag("namespace", "Kubernetes namespace, used for Kube objects.").Default(certs.DefaultNamespace).Envar("CONTOUR_NAMESPACE").StringVar(&certgenConfig.Namespace)
-	// NOTE: --certificate-lifetime can be used to accept Duration string once certificate rotation is supported.
-	certgenApp.Flag("certificate-lifetime", "Generated certificate lifetime (in days).").Default(strconv.Itoa(certs.DefaultCertificateLifetime)).UintVar(&certgenConfig.Lifetime)
+	certgenApp.Arg("outputdir", "Directory to write output files into (default \"certs\").").Default("certs").StringVar(&certgenConfig.OutputDir)
```
skriss, line 44: Same here, this is an arg not a flag so keep it separate.

skriss (changes requested):

skriss:
Also, please add a "small" changelog file for this, just to be safe since it is a user-facing change.

sunjayBhatia (commented):
should we add an e2e test for this so any future additions are properly ordered? e.g. do something like here: https://github.com/projectcontour/contour/blob/76521b7dbc562815083ec207abef686bc02614ce/test/e2e/framework.go#L198 (compile the contour binary and make some assertions about the flag order)

skriss:
> should we add an e2e test for this so any future additions are properly ordered? e.g. do something like here:
> 
> https://github.com/projectcontour/contour/blob/76521b7dbc562815083ec207abef686bc02614ce/test/e2e/framework.go#L198
> 
> (compile the contour binary and make some assertions about the flag order)

That's a good idea, otherwise this will likely just fall out of order again.

cmd/contour/bootstrap.go
izturn: done

cmd/contour/certgen.go
izturn: done

izturn:
> > should we add an e2e test for this so any future additions are properly ordered? e.g. do something like here:
> > https://github.com/projectcontour/contour/blob/76521b7dbc562815083ec207abef686bc02614ce/test/e2e/framework.go#L198
> > 
> > (compile the contour binary and make some assertions about the flag order)
> 
> That's a good idea, otherwise this will likely just fall out of order again.

To ensure they all are sorted with parse the command line's output is too complicated, maybe a ut is enough

cmd/contour/contour_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"sort"
+	"testing"
+
+	"github.com/sirupsen/logrus"
+	kingpin "gopkg.in/alecthomas/kingpin.v2"
+)
+
+func testOptionFalgsAreSortedHelper(t *testing.T, cmd *kingpin.CmdClause) {
```
skriss, line 24:
```suggestion
func assertOptionFlagsAreSorted(t *testing.T, cmd *kingpin.CmdClause) {
```

cmd/contour/contour_test.go
```diff
@@ -0,0 +1,66 @@
...
+		flags = append(flags, v.Name)
+	}
+	if !sort.StringsAreSorted(flags) {
+		t.Errorf("the flags for subcommand: '%s' aren't sorted", cmd.Model().Name)
+
+		sort.Strings(flags)
+		for _, v := range flags {
+			println(v)
+		}
+	}
```
skriss, line 37: Can simplify this down to:
```suggestion
	assert.Truef(t, sort.StringsAreSorted(flags), "the flags for subcommand %q aren't sorted: %v", cmd.Model().Name, flags)
```

cmd/contour/contour_test.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func TestOptionFlagsAreSorted(t *testing.T) {
```
skriss, line 40: The one downside to this approach is that if we ever add a new subcommand, we'll have to remember to add it to this test, but I think that's pretty rare in practice so I'm OK with this.

skriss (changes requested):
Thanks for the updates @izturn, had a couple minor nits but otherwise looks good to me.

sunjayBhatia:
> > > should we add an e2e test for this so any future additions are properly ordered? e.g. do something like here:
> > > https://github.com/projectcontour/contour/blob/76521b7dbc562815083ec207abef686bc02614ce/test/e2e/framework.go#L198
> > > 
> > > (compile the contour binary and make some assertions about the flag order)
> > 
> > 
> > That's a good idea, otherwise this will likely just fall out of order again.
> 
> To ensure they all are sorted with parse the command line's output is too complicated, maybe a ut is enough

yeah that works for me 👍🏽

cmd/contour/contour_test.go
izturn: sorry for commit the test code

cmd/contour/contour_test.go
izturn: done

cmd/contour/contour_test.go
izturn: done

cmd/contour/contour_test.go
izturn: done

izturn:
i don't know why `codegen` is not happy again, :(

skriss (approved):
Thanks for the updates @izturn, LGTM!

sunjayBhatia (approved):

PR #4961, opened 2023-01-05
test/e2e/gateway: Remove redundant tests

sunjayBhatia:
Changes:
- Remove test cases for path matching
  - Covered by conformance tests for exact path matching and prefix path matching
  - Still need to contribute a couple prefix match cases back upstream before we can delete this test completely
- Remove basic path redirect test
  - Covered in Gateway API v0.6.0
  - More advanced features are not covered in the latest release but are on main, so we cannot remove those cases yet.

Updates #4579


sunjayBhatia:
See https://github.com/kubernetes-sigs/gateway-api/pull/1627 for PR to gw api main to add more path matching cases

skriss (approved):

PR #4962, opened 2023-01-05
Gateway API: support URLRewrite filter

### LINKED ISSUE #4300: Gateway API: support HTTPURLRewriteFilter (extended/experimental)
Add support for the Gateway API [HTTPURLRewriteFilter](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPURLRewriteFilter). This is currently extended support/experimental, but seems worth adding.

skriss:
Closes #4300.

~~Implements prefix and full path rewriting; doesn't do host rewriting (will do that in a separate PR).~~

~~This currently passes the conformance test (with a fix for https://github.com/kubernetes-sigs/gateway-api/pull/1622/files#r1062855963 in place), but it is not working properly for the case where you want to remove the prefix entirely (e.g. `/prefix/foo` -> `/foo`). The issue is described in https://github.com/envoyproxy/envoy/issues/23130; looks like we might need to generate multiple Envoy routes to properly handle that case. Putting this up as a draft for now while I think about how to best handle that case.~~

### DIFF
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -1487,6 +1487,23 @@ func (p *GatewayAPIProcessor) clusterRoutes(routeNamespace string, matchConditio
 	// the matches is satisfied." To implement this,
 	// we create a separate route per match.
 	for _, mc := range matchConditions {
+		// Handle the case where the prefix is supposed to be rewritten to "/", i.e. removed.
+		// This doesn't work out of the box in Envoy with path_separated_prefix
+		// and prefix_rewrite, so we have to use a regex. Specifically, for a prefix
+		// match of "/foo", a prefix rewrite of "/", and a request to "/foo/bar", Envoy
+		// will rewrite the request path to "//bar" which is invalid. The regex handles matching
+		// and removing any trailing slashes.
+		//
+		// This logic is implemented here rather than in internal/envoy because there
+		// is already special handling at the DAG level for similar issues for HTTPProxy.
+		if pathRewritePolicy != nil && pathRewritePolicy.PrefixRewrite == "/" {
+			prefixMatch, ok := mc.path.(*PrefixMatchCondition)
+			if ok {
+				pathRewritePolicy.PrefixRewrite = ""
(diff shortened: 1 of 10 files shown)


skriss:
https://github.com/projectcontour/contour/pull/4962/commits/94e897728c42a32e6d5536aac3a9579c24c9ce3e should address the prefix removal case -- it's worked for the cases I've tested, but please check me on the logic.

skriss:
I went ahead and added support for Hostname rewrites as well since it was fairly small. This now passes the conformance tests that exist for this filter (they're in `main` but not yet released).

internal/dag/gatewayapi_processor.go
```diff
@@ -1430,6 +1510,23 @@ func (p *GatewayAPIProcessor) clusterRoutes(routeNamespace string, matchConditio
...
+		// will rewrite the request path to "//bar" which is invalid. The regex handles matching
+		// and removing any trailing slashes.
+		//
+		// This logic is implemented here rather than in internal/envoy because there
+		// is already special handling at the DAG level for similar issues for HTTPProxy.
+		if pathRewritePolicy != nil && pathRewritePolicy.PrefixRewrite == "/" {
+			prefixMatch, ok := mc.path.(*PrefixMatchCondition)
+			if ok {
+				pathRewritePolicy.PrefixRewrite = ""
+				pathRewritePolicy.PrefixRegexRemove = "^" + regexp.QuoteMeta(prefixMatch.Prefix) + "/*"
```
sunjayBhatia, line 1526: I guess this is `/*` not `/?` for the case there are be multiple slashes in a request, e.g. `/foo///bar`, esp. given we have now the option to disable merging slashes

if thats right, minor thing but might be worth a note here (or maybe a conformance test? not sure if this test will get accepted in upstream conformance)

sunjayBhatia (approved):
LGTM!

internal/dag/gatewayapi_processor.go
skriss: Yeah that was the thinking, TBH I'm not 100% sure it's the right semantics but I think it's fine for now

PR #4963, opened 2023-01-06
test/e2e: App poller logs request/response errors

sunjayBhatia:
Should help in seeing which stage of upgrade test causes errors and what they are


skriss (approved):

sunjayBhatia:
looks like this PR itself shows the issue, gonna maybe switch to an HTTPProxy and use some resp header operators to see if we can get some more info in the response we can log

sunjayBhatia:
if there is an unexpected status code, errors look like:

```
unexpected status code: 503 response flags: [UH]
```

Issue #4851, opened 2022-11-15
shutdown-manager issues


skriss:
xref. #3192 
xref. #4322 
xref. #4812 

We've encountered a number of interrelated issues with the graceful Envoy shutdown workflow that I want to capture in one place.

The first issue goes something like the following:
1. _Something_ triggers Kubernetes to terminate/restart the shutdown-manager sidecar container in an Envoy pod. We've seen definite evidence of its liveness probes failing due to timeouts, which could be caused by the container not having any resource requests and getting CPU-throttled.
2. When the shutdown-manager container is terminated, its preStop hook is triggered, which runs the `contour envoy shutdown` command. This command tells the associated Envoy to start draining all Listeners, including the Listener that is the target of the Envoy container's readiness probe. So now, Envoy is draining HTTP/S connections as well as reporting unready.
3. The shutdown-manager's preStop hook eventually returns once enough connections have been closed, and the shutdown-manager container restarts successfully.
4. The Envoy container is stuck indefinitely in a "draining" state, failing its readiness probe, but never restarting because there is no liveness probe or anything else to trigger a restart.

A few thoughts about this issue:
- adding resource requests to the shutdown-manager (and really, all containers) may help minimize the occurrence of this issue since they should ensure the container(s) get the CPU they need to avoid being restarted
- adding a liveness probe to the Envoy container may help avoid it getting permanently stuck by restarting it when it's not healthy, as long as its implementation plays nicely with the whole graceful shutdown workflow
- a restart of the shutdown-manager sidecar should probably not be triggering a drain of Envoy listeners. That should _only_ be triggered by the Envoy container itself terminating/restarting.
- https://github.com/projectcontour/contour/security/advisories/GHSA-mjp8-x484-pm3r needs to be kept in mind as we potentially modify this workflow

Secondly, #4322 describes issues with draining nodes due to the emptyDir that is used to allow the shutdown-manager and envoy containers to communicate (via UDS for the envoy admin interface, and by file to communicate when the Listener drain is done). I won't repeat that discussion here, just linking it for reference.

skriss:
A few more thoughts:
- it seems like the *best* solution here would be to build our own Envoy image, that includes the Contour binary, and then use the `contour envoy shutdown` command as a preStop hook for the Envoy container (that would trigger Envoy to drain connections and wait until they fall below the threshold).  This would eliminate the sidecar, the emptyDir volume, and any need for communicating across containers.
[...]

PR #4967, opened 2023-01-09
remove shutdown-manager liveness probe

skriss:
The probe can currently cause problems when it fails by causing the shutdown-manager container to be restarted by itself, which then results in the envoy container getting stuck in a "DRAINING" state indefinitely.

Not having the probe is less bad overall because envoy pods are less likely to get stuck in "DRAINING", and the worst case without it is that shutdown-manager is truly unresponsive during a pod termination, in which case the envoy container will simply terminate without first draining active connections.

Updates #4851.

### DIFF
--- a/changelogs/unreleased/4967-skriss-minor.md
+++ b/changelogs/unreleased/4967-skriss-minor.md
@@ -4,4 +4,6 @@ The liveness probe has been removed from the Envoy pods' shutdown-manager sideca
 This change is to mitigate a problem where when the liveness probe fails, the shutdown-manager container is restarted by itself.
 This ultimately has the unintended effect of causing the envoy container to be stuck indefinitely in a "DRAINING" state and not serving traffic.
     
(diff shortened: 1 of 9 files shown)


changelogs/unreleased/4967-skriss-minor.md
```diff
@@ -0,0 +1,7 @@
+## shutdown-manager sidecar container liveness probe removed
+
+The liveness probe has been removed from the Envoy pods' shutdown-manager sidecar container.
+This change is to mitigate a problem where when the liveness probe fails, the shutdown-manager container is restarted by itself.
+This ultimately has the unintended effect of causing the envoy container to be stuck indefinitely in a "DRAINING" state and not serving traffic.
+    
+Overall, not having the liveness probe on the shutdown-manager container is less bad because envoy pods are less likely to get stuck in "DRAINING", and the worst case without it is that shutdown-manager is truly unresponsive during a pod termination, in which case the envoy container will simply terminate without first draining active connections.
```
sunjayBhatia, line 7: nit on the wording, might be good to explicitly add that the termination/restart of the container caused by the liveness probe will ensure the envoy pod is back in the set of ready envoys to load balance traffic to

sunjayBhatia (approved):

PR #4970, opened 2023-01-09
site: add troubleshooting doc for unready Envoy

skriss:
Updates #4851.


site/data/docs/main-toc.yml
```diff
@@ -97,6 +97,8 @@ toc:
         url: /troubleshooting/profiling-contour
       - page: Contour Operator
         url: /troubleshooting/operator
+      - page: Envoy Container Stuck Unready
```
sunjayBhatia, line 100: tiny nit
```suggestion
      - page: Envoy Container Stuck in Unready State
```

site/content/docs/main/troubleshooting/envoy-container-draining.md
```diff
@@ -0,0 +1,28 @@
...
+livenessProbe:
+  httpGet:
+    path: /ready
+    port: 8002
+  initialDelaySeconds: 15
+  periodSeconds: 5
+  failureThreshold: 6
+```
+
+This will help the envoy container recover if it does become stuck in this state.
```
sunjayBhatia, line 25: maybe note here that recover = restart/be recreated

i forget off the top of my head how graceful draining will/will not work

skriss: > i forget off the top of my head how graceful draining will/will not work in this case

Yeah, it depends on what the exact sequence of operations was that preceded the container getting stuck in this state. If it was the shutdown-manager sidecar being restarted immediately beforehand, then _that_ would have triggered a graceful drain, followed by the liveness probe then restarting the envoy container, which is more or less OK, just stretched out over a slightly longer time frame than if the whole Pod had been terminated.

skriss: I added a vague comment related to this

sunjayBhatia (approved):

PR #4971, opened 2023-01-10
expose configuration for envoy's RateLimitedAsResourceExhausted

### LINKED ISSUE #4901: Question: Returning custom gRPC header or status in rate limiting

I'm curious to know if there is a way to set a custom gRPC status code and/or header to be returned if a rate limit is exceeded via the contour rate limiting system, rather than simply returning 429. The 429 error gets mapped by gRPC into an `UNAVAILABLE` error without much user-visible context, and we'd like to provide a nicer error message.

vroldanbet:
Closes https://github.com/projectcontour/contour/issues/4901

The Rate Limit filter in Envoy translates a `429 HTTP` response code to `UNAVAILABLE` as specified in the [gRPC mapping document](https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md). Alternatively, Envoy allows translating it to a less ambiguous `RESOURCE_EXHAUSTED`. 

This commit introduces a new setting to allow contour to forward the same parameter introduced in https://github.com/envoyproxy/envoy/pull/4879

The default value is disabled to retain the original behaviour of returning `UNAVAILABLE`, as changing it would be a breaking change.

This PR is largely inspired in the work done at https://github.com/projectcontour/contour/pull/3457

FYI @sunjayBhatia as we discussed about this in https://github.com/projectcontour/contour/issues/4901

Closes #4901.


skriss:
Thanks for the PR @vroldanbet! Please run `make generate` and commit the results to get the codegen CI check to pass.

skriss:
Please also address https://github.com/projectcontour/contour/actions/runs/3895937483/jobs/6651861120, the output has instructions to follow. Thanks!

vroldanbet:
@skriss done! Ready for another CI run!

examples/contour/01-crds.yaml
```diff
@@ -1006,7 +1011,7 @@ spec:
                           feature gate. \n This field is immutable."
                         items:
                           x-kubernetes-map-type: atomic
-                          description: ResourceClaim references one entry in PodSpec.ResourceClaims.
+description: ResourceClaim references one entry in PodSpec.ResourceClaims.
```
skriss, line 1014: We just merged https://github.com/projectcontour/contour/pull/4972 to fix this whitespace problem for macOS, if you could rebase your branch on `main` and re-run `make generate` it should go away.

skriss (changes requested):

skriss:
Looks like one UT still needs to be updated for the new field (https://github.com/projectcontour/contour/blob/main/cmd/contour/servecontext_test.go#L623-L645), otherwise this looks good to me

vroldanbet:
@skriss just pushed the fix, thanks for your patience and prompt feedback 🙏🏻

sunjayBhatia (approved):

skriss (approved):
LGTM, thanks @vroldanbet!

PR #4972, opened 2023-01-11
update generate-crd-yaml.sh to be macOS-compatible

skriss:


skriss:
Apparently `\s` is a GNU sed extension not recognized by BSD/macOS sed, and was resulting in this: https://github.com/projectcontour/contour/blob/147084db2cc368476abfa31e3231d411e49a077f/examples/contour/01-crds.yaml#L1014

sunjayBhatia (approved):
whoops that was me!

skriss:
> whoops that was me!

nbd, the GNU vs BSD diffs are always a pain!

PR #4978, opened 2023-01-13
Add additional go build environment variables to image build

sunjayBhatia:
To accommodate consumers who may have custom build requirements


Makefile
```diff
@@ -46,9 +46,15 @@ BUILD_BASE_IMAGE ?= golang:1.19.4
 # Enable build with CGO.
 BUILD_CGO_ENABLED ?= 0
 
+# Specify private modules.
+BUILD_GOPRIVATE ?= ""
```
sunjayBhatia, line 50: added this since it might be useful, though in practice unlikely to be used

skriss (approved):

Issue #4964, opened 2023-01-06
Gateway API: support HTTPRequestRedirectFilter path rewriting


skriss:
Contour already supports all non-experimental fields of [HTTPRequestRedirectFilter](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRequestRedirectFilter). It would be nice to also support the `path` field, which allows for path rewriting during redirect.

See #4962 for an implementation of this for the URLRewrite filter.

skriss:
Adding this to 1.25.0 but if someone can get to it for 1.24.0 that'd be even better!

skriss:
> Hi @skriss! May I pick this up?

@Vishal-Chdhry yes, thank you! Assigning you now.

PR #4982, opened 2023-01-15
Gateway API: support RequestRedirect path rewrite filter 

Vishal-Chdhry:
Signed-off-by: Vishal Choudhary <sendtovishalchoudhary@gmail.com>

Adds support for path rewriting during redirect

Closes: #4964


Vishal-Chdhry:
@skriss are these all the required changes?

internal/dag/gatewayapi_processor.go
```diff
@@ -1119,9 +1119,47 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1beta1.HTTPRou
...
+						gatewayapi_v1beta1.RouteReasonUnsupportedValue,
+						fmt.Sprintf("HTTPRoute.Spec.Rules.Filters.RequestRedirect.Path.Type: invalid type %q: only ReplacePrefixMatch and ReplaceFullPath are supported.", filter.RequestRedirect.Path.Type),
+					)
+					continue
+				}
+
+				pathRewritePolicy = &PathRewritePolicy{
+					PrefixRewrite:   prefixRewrite,
+					FullPathRewrite: fullPathRewrite,
+				}
```
skriss, line 1162: We should not be using `pathRewritePolicy` here since that gets applied for a Route that proxies to an upstream; instead, these rewrite fields need to be used for constructing the Redirect action which is happening [here](https://github.com/projectcontour/contour/blob/ed3dbafa3e372913bf5d5325e63012c9bbc73ed2/internal/dag/gatewayapi_processor.go#L1630-L1635) for the DAG and then [here](https://github.com/projectcontour/contour/blob/ed3dbafa3e372913bf5d5325e63012c9bbc73ed2/internal/envoy/v3/route.go#L242) to generate Envoy config.

What I would probably suggest is moving the instantiation of the `dag.Redirect` that is currently happening [here](https://github.com/projectcontour/contour/blob/ed3dbafa3e372913bf5d5325e63012c9bbc73ed2/internal/dag/gatewayapi_processor.go#L1630-L1635) up to here, and setting the appropriate rewrite fields on it as part of that. Then, pass that `dag.Redirect` into `p.redirectRoutes` instead of the `gatewayapi_v1beta1.HTTPRequestRedirectFilter`.

Note that `dag.Redirect` and the `routeRedirect` function both already have support for prefix and full path rewriting since we [already support this for HTTPProxy](https://projectcontour.io/docs/v1.23.2/config/api/#projectcontour.io/v1.HTTPRequestRedirectPolicy). However, I think it's possible there's a bug in that implementation that doesn't handle completely removing the path, i.e. replacing it with `/`. You may need to use similar logic as https://github.com/projectcontour/contour/blob/ed3dbafa3e372913bf5d5325e63012c9bbc73ed2/internal/dag/gatewayapi_processor.go#L1551-L1568 to make this work for rewrites during redirects.

skriss (changes requested):
Thanks for the PR @Vishal-Chdhry!

internal/dag/gatewayapi_processor.go
```diff
@@ -1119,9 +1119,47 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1beta1.HTTPRou
...
+						gatewayapi_v1beta1.RouteReasonUnsupportedValue,
+						fmt.Sprintf("HTTPRoute.Spec.Rules.Filters.RequestRedirect.Path.Type: invalid type %q: only ReplacePrefixMatch and ReplaceFullPath are supported.", filter.RequestRedirect.Path.Type),
+					)
+					continue
+				}
+
+				if redirect != nil {
+					redirect.Path.ReplaceFullPath = &fullPathRewrite
+					redirect.Path.ReplacePrefixMatch = &prefixRewrite
+				}
```
Vishal-Chdhry, line 1162: @skriss should this not be enough as if `redirect != nil` we are calling this some lines below
https://github.com/projectcontour/contour/blob/e511e8129006b371a81f8e9df5f982bb0b67baab/internal/dag/gatewayapi_processor.go#L1230-L1254

skriss:
Hey @Vishal-Chdhry I'm thinking this one would be easier for me to take over, since it's somewhat intricate and I already worked on https://github.com/projectcontour/contour/pull/4962 so I have all the context. It's a little hard to explain short of just writing the code. I'll give you co-author credit on my work, to credit you for the work you did so far.

internal/dag/gatewayapi_processor.go
skriss: > by the way, apologies for being so late on this, I was busy in college work and with some other organisations as i was trying to get into LFX and GSoC for internships opportunities

Not a problem, hope all turns out well for you!

Vishal-Chdhry:
sure please go ahead

PR #4989, opened 2023-01-19
update to Kubernetes 1.26.1

skriss:
Also removes fix for ResourceClaim in CRDs since
it's been fixed upstream.

Also updates gomega and genproto deps.


skriss:
> could also bump kubectl here or in a separate PR, maybe once the kind images come out

Good call, will do along with the kind image when available

Issue #3693, opened 2021-05-19
Add IP allowlist and blocklist functionality


youngnick:
As either a cluster admin or application developer, I want to be able to restrict access to services behind Contour by source IP address.

This issue covers:
- identifying the exact use cases
- designing a solution to cover the use cases
- implementing that solution.

The original source issue for this was #62 (way back!), and we have a few things to decide, and I can think of a few things for us to consider.

Firstly, where and how could we implement this?
- As an annotation on Ingress objects (like in #62 )
- As a field in HTTPProxy
- In the Gateway API (although this would also include designing something for upstream)

When implementing it, we need to make sure that we call out the interactions with upstream load balancers, X-Forwarded-For, and Client IPs etc, which can complicate this sort of thing a bit for proxies.

I suspect the answer here is that the IP address we care about is whatever Envoy thinks the Client IP is, but we will need to document how to be sure you have the right one a bit more than we do now, I think.

xaleeks:
Great idea. Just want to add we need to support both IPv4 and v6, specified in both single IPs and CIDR blocks.  Another comment is we can either make this available as part of the existing `spec.rateLimitPolicy` or have this interaction with rate limiting possible.  My preference, as with a lot of our features, is to have this configurable globally by default that cascades to all HTTPProxy objects in existence and able to set it per HTTPProxy that overrides global config.

Should we allow setting a deny list instead of an allow list? Or both.
  
@youngnick  Regarding the options you listed, I think we have to think about adding these sorts of capabilities to both HTTPProxy and HTTPRoute in the meantime.  The best we can do for anything in gateway API is to land a feature request in their upstream for now.  

I moved this to 1.19, does that feel appropriate?

xaleeks:
> When implementing it, we need to make sure that we call out the interactions with upstream load balancers, X-Forwarded-For, and Client IPs etc, which can complicate this sort of thing a bit for proxies.
> 
> I suspect the answer here is that the IP address we care about is whatever Envoy thinks the Client IP is, but we will need to document how to be sure you have the right one a bit more than we do now, I think.
Does the `X-Forwarded-For` not carry the source IP on it?  I think we discussed this before as well, but an alternative is to leverage the HAProxy Proxy Protocol for this, and this would cover any TCP stream.  

I also found this extension for retrieving the original IP address.  Can we leverage this if the above aren’t sufficient?

[...]

PR #4990, opened 2023-01-19
Add ip filter design doc

ecordell:
I took a stab at a proposal to address https://github.com/projectcontour/contour/issues/3693

I haven't looked deeply at the envoy filters available for this, but I wrote down what looked like would work based on the docs.

### DIFF
--- a/design/ip-filtering-design.md
+++ b/design/ip-filtering-design.md
@@ -1,0 +1,172 @@
+# IP Filtering 
+
+Status: Approved 
+
+## Abstract
+
+Allow or deny traffic to services behind contour with an IP allowlist or an IP denylist.
+
+## Background
+
+There have been [long standing requests](https://github.com/projectcontour/contour/issues/62) to support some form of IP filtering.
+IP filtering is a useful tool for network segregation that is supported by other ingress controllers like [ingress-nginx](https://github.com/kubernetes/ingress-nginx/blob/87d1b8bbf28265386b339e65d8943d8c3f8582ed/docs/user-guide/nginx-configuration/annotations.md#whitelist-source-range) and [ambassador](https://www.getambassador.io/docs/edge-stack/latest/topics/running/ambassador#ip-allow-and-deny), and is a [common](https://github.com/projectcontour/contour/issues/62#issuecomment-698141472) [request](https://github.com/projectcontour/contour/issues/62#issuecomment-843193956) among the community, often listed as a [blocker](https://github.com/projectcontour/contour/issues/3693#issuecomment-1317049708) for adopting contour.
+
+## Goals
+
+- Support ipv4 and ipv6 allowlists and denylists via Contour APIs.
+- Support filtering on request IP or on forwarded IP (i.e. via `X-Forwarded-For`)
+
+## Non Goals
+
+- Define Gateway API support for ip filtering. This is an obvious next step, but requires broader agreement. 
+- Define a global IP filter that works across multiple Routes or HTTPProxy objects.
+- Supporting envoy's network rbac filter (this design is for the http filter only).
+
+## High-Level Design
+
+The `virtualhost` and `route` sections of an `HTTPProxy` will each have two new fields added: `ipAllowPolicy` and `ipDenyPolicy`.
+
+Example `HTTPProxy`:
+
+```yaml
+apiVersion: projectcontour.io/v1
+kind: HTTPProxy
+metadata:
+  name: basic
+spec:
+  virtualhost:
+    fqdn: foo-basic.bar.com
+    ipAllowPolicy:
+      # traffic is allowed if it came from localhost (i.e. co-located reverse proxy)
+      - cidr: 127.0.0.1/32
+        source: Peer
+  routes:
+    - conditions:
+      - prefix: /
+      services:
+        - name: s1
+          port: 80
+      # route-level ip filters overrides the virtualhost-level filters
+      ipAllowPolicy:
+      # traffic is allowed if it came from localhost (i.e. co-located reverse proxy)
+      - cidr: 127.0.0.1/32
+        source: Peer
+      # and the request originated from an IP in this range 
+      - cidr: 99.99.0.0/16
+        source: Remote
+```
+
+## Detailed Design
+
+
+### API
+
+`ipAllowPolicy`/`ipDenyPolicy` has the following structure:
+
+```yaml
+ipAllowPolicy:
+  - cidr: 127.0.0.1/32
+    source: Peer
+  - cidr: 99.99.0.0/16
+    source: Remote
+```
+
+```yaml
+ipDenyPolicy:
+  - cidr: 127.0.0.1/32
+    source: Peer
+  - cidr: 99.99.0.0/16
+    source: Remote
+```
+
+Note that:
+- Allow rules deny all requests that don't match a rule.
+- Deny rules allow all requests that don't match a rule.
+- Rules defined on the virtualhost can be overridden by rules defined on a route.
+- Allow rules cannot be mixed with Deny rules: it won't be evident if a deny rule should be an exception to an allow rule, or if an allow rule should be an exception to a deny rule.
+- Multiple allow/deny CIDR ranges can be specified by repeating a key.
+- `Peer` refers to the request IP
+- `Remote` refers to the originating IP, i.e. `X-Forwarded-For`. `numTrustedHops` may also need to be set to filter on the desired IP.
+
+### Processing
+
+`HTTPProxyProcessor` and `dag.Route` will store the ip filtering rules in a new field.
+
+These settings will be converted convert to a per-route or per-virtualhost envoy filter, i.e.:
+
+```yaml
+routes:
+  - match:
+      prefix: /
+    route:
+      cluster: example-cluster 
+    typed_per_filter_config:
+      envoy.filters.http.rbac:
+        '@type': type.googleapis.com/envoy.extensions.filters.http.rbac.v3.RBACPerRoute
+        rbac:
+          rules:
+            # action is ALLOW if ipAllowPolicy is used, otherwise DENY is used
+            action: ALLOW
+            policies:
+              ip-rules:
+                permissions:
+                  - any: true
+                principals:
+                  # direct_remote_ip is used if `source: Peer` is specified
+                  - direct_remote_ip:
+                      address_prefix: 127.0.0.1
+                      prefix_len: 32
+                  # remote_ip is used if `source: Remote` is specified
+                  - remote_ip:
+                      address_prefix: 99.99.0.0
+                      prefix_len: 16
+```
+
+Please see [the Envoy docs](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/rbac/v3/rbac.proto) for more detailed explanations of the filter behavior.
+
+## Alternatives
+
+The ip filtering could be a top-level field on the `HTTPProxy`.
+
+This would allow building "trees" of ip filters via `includes`:
+```yaml
+apiVersion: projectcontour.io/v1
+kind: HTTPProxy
+metadata:
+  name: basic
+spec:
+  virtualhost:
+    fqdn: foo-basic.bar.com
+  ipAllowPolicy:
+  # traffic is allowed if it came from localhost (i.e. co-located reverse proxy)
+  - peer: 127.0.0.1
+  # and the request originated from an IP in this range 
+  - remote: 99.99.0.0/16
+  includes:
+  - name: service2
+    conditions:
+      - prefix: /service2
+  routes:
+    - conditions:
+      - prefix: /
+      services:
+        - name: s1
+          port: 80
+---
+apiVersion: projectcontour.io/v1
+kind: HTTPProxy
+metadata:
+  name: service2
+spec:
+  ipDenyPolicy:
+    # if the request made it through the first filter, it could still be denied here
+    - remote: 99.99.0.1
+  routes:
+    - conditions:
+        - prefix: /service2
+      services:
+        - name: s2
+          port: 80
+```
+
+But no other request filtering options work this way today, so this approach seems like a mismatch with current Contour.


sunjayBhatia (commented):
it would be great to add some implementation notes about how the Envoy filters might work, once we agree on a general shape of the API

i.e. mention the RBAC filter: https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/rbac_filter and how configuring Prinicipals might work: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/rbac/v3/rbac.proto#envoy-v3-api-msg-config-rbac-v3-principal

some examples of configuration and requests (including source ip and/or x-forwarded-for header) that would be allowed/denied would be great to add

the Envoy filter allows full per-route configuration so the per-route configuration makes a lot of sense to me

design/ip-filtering-design.md
```diff
@@ -0,0 +1,140 @@
...
+There have been [long standing requests](https://github.com/projectcontour/contour/issues/62) to support some form of IP filtering.
+IP filtering is a useful tool for network segregation that is supported by other ingress controllers like [ingress-nginx](https://github.com/kubernetes/ingress-nginx/blob/87d1b8bbf28265386b339e65d8943d8c3f8582ed/docs/user-guide/nginx-configuration/annotations.md#whitelist-source-range) and [ambassador](https://www.getambassador.io/docs/edge-stack/latest/topics/running/ambassador#ip-allow-and-deny), and is a [common](https://github.com/projectcontour/contour/issues/62#issuecomment-698141472) [request](https://github.com/projectcontour/contour/issues/62#issuecomment-843193956) among the community, often listed as a [blocker](https://github.com/projectcontour/contour/issues/3693#issuecomment-1317049708) for adopting contour.
+
+## Goals
+
+- Support ip allowlists and denylists via Contour APIs.
+- Support filtering on request IP or on forwarded IP (i.e. via `X-Forwarded-For`)
+
+## Non Goals
+- Define Gateway API support for ip filtering. This is an obvious next step, but requires broader agreement.
```
sunjayBhatia, line 20: we'll likely need to write up a design for a CR to be used as an extension Filter or Policy for Gateway API

design/ip-filtering-design.md
```diff
@@ -0,0 +1,140 @@
...
+```yaml
+ipDenyPolicy:
+- peer: 127.0.0.1 
+- remote: 99.99.0.0/16
+```
+
+Note that:
+- Allow rules cannot be mixed with Deny rules: it won't be evident if a deny rule should be an exception to an allow rule, or if an allow rule should be an exception to a deny rule.
+- Multiple allow/deny CIDR ranges can be specified by repeating a key.
+- `peer` refers to the request IP
```
sunjayBhatia, line 71: writing down how exactly these concepts fold into the Envoy RBAC configuration fields would be good to make sure to head off any confusion early

ecordell:
Thanks for the review @sunjayBhatia - I'll work on ironing out the envoy details and update here.

design/ip-filtering-design.md
ecordell: Are you saying that you think we should bring the Gateway API into scope? Or just that that's the likely direction if/when this is approached in for Gateway API?

sunjayBhatia: a bit of both actually, probably good to get a baseline of what we want to support done in HTTPProxy as well as acknowledging that unless the upstream Gateway API project supports it natively we will have to write our own CR to handle that feature

tsaarni (commented):
Great design doc @ecordell, Thank You for working on this! I've added some comments/questions.

design/ip-filtering-design.md
ecordell: Yep, I took that from Ambassador (it was linked in one of the original GH issues on ip filtering). I don't have a strong opinion on the fields names, if `direct_remote` and `remote` sound better.

ecordell:
@sunjayBhatia @tsaarni I addressed most of the feedback and made a couple of updates to the proposed API.

I've implemented this https://github.com/projectcontour/contour/pull/5008 with the understanding that we may need to adjust it if we change anything here first. Implementing it was the easiest way to suss out exactly how envoy's rbac filter actually works in practice.

design/ip-filtering-design.md
```diff
@@ -0,0 +1,140 @@
...
+
+```yaml
+apiVersion: projectcontour.io/v1
+kind: HTTPProxy
+metadata:
+  name: basic
+spec:
+  virtualhost:
+    fqdn: foo-basic.bar.com
+  routes:
```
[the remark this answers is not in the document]
ecordell: I can test this - I'm not 100% sure what happens, but I would defer to whatever envoy does here.

ecordell:
Since I think the API design is what probably requires the most discussion, here are some alternatives to think about:

1. What's in the proposal right now:

```yaml
      ipAllowPolicy:
      - cidr: 127.0.0.1/32
        source: Peer
      #or 
      ipDenyPolicy:
      - cidr: 99.99.0.0/16
        source: Remote
```
- separate top-level fields for allow vs. deny
- separate field for cidr range and ip source
- uses `Remote` / `Peer` terminology instead of `Remote` and `DirectRemote`

2. An option with only one top-level field: 

```yaml
      ipFilterPolicy:
      - allow: 127.0.0.1/32
        source: Peer
      - allow: 99.99.0.0/16
        source: Peer
      #or 
      ipFilterPolicy:
      - deny: 99.99.0.0/16
        source: Remote
```
- single top-level field for filters, each filter specifies allow vs. deny
- separate field for cidr range and ip source
- uses `Remote` / `Peer` terminology instead of `Remote` and `DirectRemote`

3. Using envoy terminology for remotes:

```yaml
      ipFilterPolicy:
      - allow: 127.0.0.1/32
        source: DirectRemote
      - allow: 99.99.0.0/16
        source: Remote
      #or 
      ipFilterPolicy:
      - deny: 99.99.0.0/16
        source: Remote
```
- single top-level field for filters, each filter specifies allow vs. deny
- separate field for cidr range and ip source
- uses `Remote` and `DirectRemote`

4. If the top-level field indicates allow vs. deny, we can get by with only one field for the actual filters:

```yaml
      ipAllowPolicy:
      - directRemote: 127.0.0.1/32
      - remote: 99.99.0.0/16
      #or 
      ipDenyPolicy:
      - remote: 99.99.0.0/16
```

- separate top-level fields for allow vs. deny
- single field for ip source and range

Another api question is weather bare ips should be interpreted as `/32` cidrs. Right now I have everything written as an explicit cidr range.

design/ip-filtering-design.md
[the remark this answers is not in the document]
sunjayBhatia: based on how we program match conditions the more specific prefix will get the routing priority, so since we're using per-route filters the IP policy should be applied after the route is chosen

sunjayBhatia:
> Another api question is weather bare ips should be interpreted as /32 cidrs. Right now I have everything written as an explicit cidr range.

I'm in favor of being more explicit, I don't think it's too bad to ask people to add `/32` or `/128` if they want to allow/deny a particular IP

sunjayBhatia:
Also I'm somewhat in favor of the options that have only one top-level field, just so there are less keys to have to configure and people don't get the impression just looking at the shape of the API that they could combine allow and deny policies on a route

ecordell:
> I'm in favor of being more explicit, I don't think it's too bad to ask people to add /32 or /128 if they want to allow/deny a particular IP

I tend to agree, but for what it's worth the nginx whitelist feature supports single-ips in the filter.

izturn:
@ecordell  any progress?

tsaarni:
> > I'm in favor of being more explicit, I don't think it's too bad to ask people to add /32 or /128 if they want to allow/deny a particular IP
> 
> I tend to agree, but for what it's worth the nginx whitelist feature supports single-ips in the filter.

`iptables -s address[/mask]` also allows using single IP address, from [man page](https://man7.org/linux/man-pages/man8/iptables.8.html):  "_Address can be either a network name, a hostname, a network IP address (with /mask), or a plain IP address._". It seems to be pretty common shorthand with some routers as well (like [here](https://www.juniper.net/documentation/us/en/software/junos/security-policies/topics/topic-map/security-policy-configuration.html#id-dynamic-address-groups-in-security-policies)).

ecordell:
@tsaarni I agree, allowing bare IPs probably makes the most sense. 

@izturn I believe I've addressed all of the feedback in this PR; I'm just waiting for further review and approval. Is there something specific you'd like to see addressed?

izturn:
@ecordell, thx for your work, I can't wait to use it, btw I saw there are two branches: `ipfilter`, `annotation-ipfilter` in your fork, which one is your prefer?

ecordell:
Thanks @izturn!

`ipfilter` is the branch we should look at - `annotation-ipfilter` is a fork that uses annotations to drive the ipfiltering so that we can use the ipfiltering features before this proposal is approved and merged.

izturn:
got it

design/ip-filtering-design.md
```diff
@@ -0,0 +1,166 @@
...
+Example `HTTPProxy`:
+
+```yaml
+apiVersion: projectcontour.io/v1
+kind: HTTPProxy
+metadata:
+  name: basic
+spec:
+  virtualhost:
+    fqdn: foo-basic.bar.com
```
wilsonwu, line 38: Could we support allow and deny IP list for virtualhost level?
```suggestion
    fqdn: foo-basic.bar.com
    ipAllowPolicy:
      # traffic is allowed if it came from localhost (i.e. co-located reverse proxy)
      - cidr: 127.0.0.1/32
        source: Peer
      # and the request originated from an IP in this range 
      - cidr: 99.99.0.0/16
        source: Remote
```

izturn: @ecordell

ecordell:
@sunjayBhatia I'll work on updating for virtualhost filtering, sorry for the silence

sunjayBhatia:
> Late to the party on this one. Is supporting the `ingress.kubernetes.io/whitelist-source-range` annotation on ingress objects in the scope of this design? If we are adding IP filtering then it would be great to be a drop in replacement for other ingress replacements. That annotation is the only reason I keep an nginx controller running in our clusters.

not at the moment, no, but if you're willing to contribute an addendum to this document or an additional feature once this is merged/agreed upon we will of course consider it

ecordell:
@sunjayBhatia doc is updated for virtualhost-level ip filtering

It's implemented in https://github.com/projectcontour/contour/pull/5008 as well

wilsonwu (approved):
lgtm

design/ip-filtering-design.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Background
+
+There have been [long standing requests](https://github.com/projectcontour/contour/issues/62) to support some form of IP filtering.
+IP filtering is a useful tool for network segregation that is supported by other ingress controllers like [ingress-nginx](https://github.com/kubernetes/ingress-nginx/blob/87d1b8bbf28265386b339e65d8943d8c3f8582ed/docs/user-guide/nginx-configuration/annotations.md#whitelist-source-range) and [ambassador](https://www.getambassador.io/docs/edge-stack/latest/topics/running/ambassador#ip-allow-and-deny), and is a [common](https://github.com/projectcontour/contour/issues/62#issuecomment-698141472) [request](https://github.com/projectcontour/contour/issues/62#issuecomment-843193956) among the community, often listed as a [blocker](https://github.com/projectcontour/contour/issues/3693#issuecomment-1317049708) for adopting contour.
+
+## Goals
+
+- Support ipv4 and ipv6 allowlists and denylists via Contour APIs.
+- Support filtering on request IP or on forwarded IP (i.e. via `X-Forwarded-For`)
```
izturn, line 17: From the change log:
>Filters can indicate whether the direct IP should be used or whether a reported IP from `PROXY` or `X-Forwarded-For` should be used instead.

so what's about `PROXY`?

design/ip-filtering-design.md
```diff
@@ -0,0 +1,172 @@
...
+    source: Remote
+```
+
+Note that:
+- Allow rules deny all requests that don't match a rule.
+- Deny rules allow all requests that don't match a rule.
+- Rules defined on the virtualhost can be overridden by rules defined on a route.
+- Allow rules cannot be mixed with Deny rules: it won't be evident if a deny rule should be an exception to an allow rule, or if an allow rule should be an exception to a deny rule.
+- Multiple allow/deny CIDR ranges can be specified by repeating a key.
+- `Peer` refers to the request IP
```
izturn, line 88: `Peer`/`Remote` OR `direct_remote_ip`/`remote_ip` ?

design/ip-filtering-design.md
izturn: > Is the terminology `peer` and `remote` originating from Ambassador? Envoy seems to use [`direct_remote_ip`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/rbac/v3/rbac.proto#envoy-v3-api-field-config-rbac-v3-principal-direct-remote-ip) and [`remote_ip`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/rbac/v3/rbac.proto#envoy-v3-api-field-config-rbac-v3-principal-remote-ip). Which one is better to align?

I like the latter

izturn (approved):
/lgtm

design/ip-filtering-design.md
ecordell: > maybe we need to unify them all

Not sure I'm following your suggestion, would you mind rephrasing it?

izturn: sorry for my pool english, i means the descriptions in these two places need to be consistent

design/ip-filtering-design.md
```diff
@@ -0,0 +1,172 @@
+# IP Filtering 
+
+Status: Draft
```
sunjayBhatia, line 3: can switch this to approved

sunjayBhatia (approved):
Taking this and the implementation in #5008 together I think this is good to approve, the open questions I had that weren't explicitly covered here were around how exactly the overrides would work (virtualhost config overridden per-route) and how includes would work, but looks like that is already tested in the implementation so we can discuss there

I'll leave this open for a little longer in case anyone else wants to chime in

sunjayBhatia (commented):
one more thought I just had would be to nail down what the precedence of the RBAC filter is when combined with other filters

i.e. should IP filtering happen before rate limiting, ExtAuth etc.

since the filter ends up getting applied as a per-route filter this might mean it is evaluated later than people would expect

design/ip-filtering-design.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Non Goals
+
+- Define Gateway API support for ip filtering. This is an obvious next step, but requires broader agreement. 
+- Define a global IP filter that works across multiple Routes or HTTPProxy objects.
+- Supporting envoy's network rbac filter (this design is for the http filter only).
+
+## High-Level Design
+
+The `route` section of an `HTTPProxy` will have two new fields added: `ipAllowPolicy` and `ipDenyPolicy`.
```
skriss, line 27:
```suggestion
The `virtualhost` and `route` sections of an `HTTPProxy` will each have two new fields added: `ipAllowPolicy` and `ipDenyPolicy`.
```

skriss (commented):
One small clarification for posterity

PR #4993, opened 2023-01-20
internal/xds/v3: Improve logging on connection close

sunjayBhatia:
- contour xDS server does not log errors if connection was canceled by client, detected via gRPC code
- when connection is closed change to debug from info level to avoid log noise
- add stream open/close logging callbacks for go-control-plane implementation


skriss (approved):
Nice, LGTM

PR #4994, opened 2023-01-23
feat: Add HTTP support for External Auth

clayton-gonsalves:
This change adds external authorization support for HTTP upstreams.
Fixes: #4954 

Signed-off-by: Clayton Gonsalves <claytonivorgonsalves@gmail.com>

Issue #4954, opened 2023-01-03
Support for external auth with HTTP virtual hosts


claytonig:
**Please describe the problem you have**
I am evaluating contour as a potential candidate for our ingress across our tech stack. We terminate TLS at the cloud provider level, which is a requirement. The cloud provider load balancers do not forward the SNI information if TLS is terminated at the load balancer.  

External authorization requires a virtual host to have HTTPS enabled, which doesn't work with the above mentioned constraints. 

**Current setup**
Cloud provider: AWS 
Helm chart: [Bitnami](https://github.com/bitnami/charts/tree/main/bitnami/contour)

I tried the setup both with an NLB and an ELB. Listing the annotations of the `LoadBalancer` service as rest of the chart was pretty much default. 

ELB
```
  annotations:
    service.beta.kubernetes.io/aws-load-balancer-backend-protocol: https
    service.beta.kubernetes.io/aws-load-balancer-ssl-cert: arn:aws:acm:xxx
```

NLB
```
  annotations:
     service.beta.kubernetes.io/aws-load-balancer-type: nlb
     service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: ip
     service.beta.kubernetes.io/aws-load-balancer-scheme: internet-facing
     service.beta.kubernetes.io/aws-load-balancer-ssl-ports: "443"
     service.beta.kubernetes.io/aws-load-balancer-ssl-cert: "arn:aws:acm:xxx"
     service.beta.kubernetes.io/aws-load-balancer-backend-protocol: ssl
```

In both cases i got the following response `curl: (52) Empty reply from server`

sunjayBhatia:
one solution, which the community would need to indicate if it is acceptable could be to have a globally configured ExtAuth Service that all HTTP vhosts would use and could be the default for all non-HTTP vhosts but overridable in HTTPS vhosts only

this would need some careful documentation etc. (and some more validation) to make sure people understand how it works

claytonig:
Yeah, I was also thinking along the same lines. Here's what I had in mind. 


GlobalHttpExtAuth dial on the contour configuration. 

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: contour
  namespace: projectcontour
data:
  contour.yaml: |
    GlobalHttpExtAuth:
      # extensionService is the <namespace>/<name>
      # of the ExtensionService we created in the
      # previous step.
      extensionService: projectcontour/externalAuth
      # failOpen is whether to allow requests through
[...]

PR #4995, opened 2023-01-23
Design: ExtAuth HTTP support

clayton-gonsalves:
Design proposal to address #4954 

There is a WIP draft [PR](https://github.com/projectcontour/contour/pull/4994) with some of the intended code changes. 
Signed-off-by: Clayton Gonsalves <claytonivorgonsalves@gmail.com>

### DIFF
--- a/design/global-external-authorization-design.md
+++ b/design/global-external-authorization-design.md
@@ -1,0 +1,384 @@
+# Global External Authorization
+
+Status: Accepted
+
+## Abstract
+
+HTTP Support for External Authorization.
+
+## Background
+See [#4954](https://github.com/projectcontour/contour/issues/4954) for context.
+
+## Goals
+- Support Envoy's L7 External Authorization filter for HTTP Virtual Hosts.
+- Extend the functionality of existing External Authorization.
+
+## Non Goals
+- Supporting more than one External Authorization Configuration for HTTP endpoints.
+
+## High-Level Design
+
+Contour will add HTTP support for Envoy's External Authorozation.
+
+Currently, Contour supports External Authorization only over HTTPS. External Authorization allows for requests to be authenticated before they are forwarded upstream. However, External authorization over HTTPS is incompatible with setups that terminate TLS at the cloud provider's load balancer layer because the Envoy HCM relies on the SNI to choose the correct filter chain. 
+
+A new type, `GlobalExtAuth`, will be defined as part of the `ContourConfiguration`.
+A `GlobalExtAuth` config would define a global external authorization configuration for all hosts and routes. 
+
+### Opting out from Global External Auth 
+Global external auth would enable auth on all virtual hosts by default. 
+However, individual HTTPProxy owners will have the option to toggle this setting. This can be done in one of the following ways. 
+
+#### Disabling External Auth
+
+##### Virtual Host level 
+Disable global external auth on the virtual host. This setting would disable all routes on said virtual host. 
+
+```yaml
+kind: HTTPProxy
+...
+spec:
+  virtualhost:
+    fqdn: local.projectcontour.io
+    authorization:
+      globalExtAuthDisabled: true
+...
+```
+
+##### Route level
+For finer grain toggles, global external auth can also be toggled on an individual route. This is similar to the existing per route authentication policy and we will reuse the same dials.
+
+```yaml 
+kind: HTTPProxy
+metadata:
+  name: echo
+spec:
+...
+  routes:
+    - authPolicy:
+        context:
+          route: apex
+        disabled: false
+      conditions:
+        - prefix: /
+      services:
+        - name: ingress-conformance-echo
+          port: 80
+    - authPolicy:
+        context:
+          route: subpage
+        disabled: true
+      conditions:
+        - prefix: /subpage
+      services:
+        - name: ingress-conformance-echo
+          port: 80
+```
+
+#### HTTPS Override
+If a HTTPS virtual host has external authorization enabled, that configuration will override the global external authorization.
+This gives an individual service owner finer grain control of their services in case they have different requirements than the settings in the global external authorization. 
+
+## Detailed Design
+
+### Contour Configuration changes
+An external authorization service can be configured in the Contour config file.
+This External authorization configuration will be used for all HTTP routes. 
+
+```yaml
+globalExtAuth:
+  extensionService: projectcontour-auth/htpasswd
+  failOpen: false
+  authPolicy:
+    disabled: false
+    context:
+      header1: value1
+      header2: value2
+  responseTimeout: 1s
+  withRequestBody:
+    maxRequestBytes: 1024
+```
+
+```go
+type Parameters struct {
+  ...
+  // GlobalExtAuth optionally holds properties of the global external auth configuration.
+	GlobalExtAuth GlobalExtAuth `yaml:"globalExtAuth,omitempty"`
+  ...
+}
+
+// GlobalExtAuth defines properties of a global external auth service.
+type GlobalExtAuth struct {
+	// ExtensionService identifies the extension service responsible for the authorization.
+	// formatted as <namespace>/<name>.
+	ExtensionService string `yaml:"extensionService,omitempty"`
+
+
+	// AuthPolicy sets a default authorization policy for client requests.
+	// This policy will be used unless overridden by individual routes.
+	//
+	// +optional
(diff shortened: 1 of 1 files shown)


design/external-authorization-http-support-design.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+### HTTPS Override
+If an upstream has HTTPS enabled, the HTTPS configuration will take precedence. 
+i.e., if global HTTP external authorization is enabled and an upstream is configured with HTTPS without external authorization. Then that particular upstream would not use envoy's external authorization filter. 
+
+## Alternatives Considered
+
+### Envoy Filter Matching API 
+Having a separate external authorization configuration for each upstream is more desirable. We  can achieve this by leveraging Envoy's [Matching API]()
```
davinci26, line 59: nit: docs are missing here

design/external-authorization-http-support-design.md
```diff
@@ -0,0 +1,165 @@
...
+  context:
+    header1: value1
+    header2: value2
+    # ResponseTimeout configures maximum time to wait for a check response from the authorization server
+  responseTimeout: 1s
+
+```
+
+### HTTPS Override
+If an upstream has HTTPS enabled, the HTTPS configuration will take precedence.
```
davinci26, line 53: is this basically because they will be on different filters? we should probably follow the precedent of the project. 

Does contour have any other configurations that can be applied both with `configmap` and with `annotation`?

davinci26 (approved):
Some minor nits. My opinion doesn't hold much weight in this project but imho this is a great feature for single tenant clusters that want a uniform behaviour enforced by a single infra team.

design/external-authorization-http-support-design.md
sunjayBhatia: we do have precedent for doing some global configuration with a per-route override, we allow configuration of reqest/response header modification at the [config file](https://projectcontour.io/docs/v1.23.2/configuration/#policy-configuration) and [per-route](https://projectcontour.io/docs/v1.23.2/config/api/#projectcontour.io/v1.Route), with the aim being the former is configured by the persona operating the contour installation, and the latter for any app-operator/developer specific overrides

when we support this feature in Gateway API seems like there may be room here to implement this as a Policy resource and use the override capabilities there

design/external-authorization-http-support-design.md
```diff
@@ -0,0 +1,165 @@
...
+      authPolicy:
+        context:
+          header2: value2
+  routes:
+  - services:
+    - name: internal.projectcontour.io
+      port: 80
+```
+
+With this proposal, contour will generate the envoy configuration snippet below.
```
sunjayBhatia, line 113: I'm not entirely sure this example will work, I think the [ServerNameInput](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/matching/common_inputs/network/v3/network_inputs.proto#extensions-matching-common-inputs-network-v3-servernameinput) is for matching SNI. You would have to match on the `:authority` header

Haven't played around with the `Composite` and `ExecuteFilterAction` filters/actions so curious to see how that works

not a big deal to generate a working example here so we can omit it if needed, but if we do want to provide an example we should validate it works

sunjayBhatia: Also I think the TLS-vhost auth filter configuration would still go on the HTTPS listener and specific filter chain, rather than be combined with the HTTP one

clayton-gonsalves: Fair, I should've validated it. 
I made some changes; the new config works with the `host` header and the `HttpRequestHeaderMatchInput` matching API. 

Also, agree with the TLS-vhost filter, which would be on the HTTPS listener. I added a note to make that clear.

design/external-authorization-http-support-design.md
```diff
@@ -0,0 +1,165 @@
...
+
+### HTTPS Override
+If an upstream has HTTPS enabled, the HTTPS configuration will take precedence. 
+i.e., if global HTTP external authorization is enabled and an upstream is configured with HTTPS without external authorization. Then that particular upstream would not use envoy's external authorization filter. 
+
+## Alternatives Considered
+
+### Envoy Filter Matching API 
+Having a separate external authorization configuration for each upstream is more desirable. We  can achieve this by leveraging Envoy's [Matching API](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/matching/matching_api)
+Note: This feature is an Alpha release, and there are no guarantees that future changes will maintain backward compatibility.
```
skriss, line 60: Agree that this is probably the more desirable functionality, especially since it is consistent for all types of virtual hosts. Unfortunately I don't think we'll be comfortable making use of it until the matching API becomes stable.

design/external-authorization-http-support-design.md
```diff
@@ -0,0 +1,165 @@
...
+Currently, Contour supports External Authorization only over HTTPS. External Authorization allows for requests to be authenticated before they are forwarded upstream. However, External authorization over HTTPS is incompatible with setups that terminate TLS at the cloud provider's load balancer layer because the Envoy HCM relies on the SNI to choose the correct filter chain. 
+
+A new type, `GlobalHTTPExtAuth`, will be defined as part of the `ContourConfiguration`.
+A `GlobalHTTPExtAuth` config would define a global external authorization configuration for all the HTTP routes. 
+
+## Detailed Design
+
+### Contour Configuration changes
+By using global external authorization, an external authorization service can be configured in the Contour config file.
+This External authorization configuration will be used for all HTTP routes.
```
skriss, line 32: One use case we should consider is whether individual HTTP vhosts should be able to opt out of ExtAuth, if it's enabled globally. This is a little tricky since different orgs may view this functionality differently -- is it defining a global **requirement** that all HTTP vhosts must use a given ExtAuth server, or a global **default** ExtAuth config that can be opted out of by individual sites?

Personally I think I lean towards the latter, particularly since with TLS vhosts, it's up to the individual root HTTPProxy owner to decide whether to enable ExtAuth or not. And, if we move to using the matching API to enable per-HTTP vhost ExtAuth config, then we'd again support opting into ExtAuth on a per-vhost basis.

If we _do_ want to support the use case of opting out of the global config, perhaps we could do something like:

```yaml
kind: HTTPProxy
metadata:
  name: opt-out-of-ext-auth
spec:
  virtualhost:
    fqdn: local.projectcontour.io
    authorization:
      disabled: true
```

clayton-gonsalves: I see your point; it would be nice to have the flexibility to opt-out if a service owner doesn't want auth. 
This should be achievable by doing it on a per-route basis, but that could get tedious if there are a bunch of them.

clayton-gonsalves: I have updated the design and added two toggles to disable auth; per route and per virtual host

sunjayBhatia (commented):
I think this LGTM, we can work out some more fine details as needed in PR review of the actual implementation

Just needs a `main` merge (just to get CI happy) and should be good to go

skriss (approved):
:+1: to moving ahead with this. I did have one final question but it doesn't need to block starting to make progress here. Thanks for the work here @clayton-gonsalves!

design/external-authorization-http-support-design.md
```diff
@@ -0,0 +1,381 @@
...
+spec:
+  virtualhost:
+    fqdn: local.projectcontour.io
+    authorization:
+      ...
+      globalExtAuthDisabled: true
+      ...
+```
+
+### HTTPS Override
```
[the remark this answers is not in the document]
skriss: So the general pattern would be:
- globally specified config acts as a _default_ for all vhosts
- individual vhosts can opt out of the default via an HTTPProxy flag
- where supported, per-vhost config for the same feature overrides global config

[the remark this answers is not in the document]
sunjayBhatia: Ah I think I assumed the behavior was as you describe here @skriss I totally glossed over line 199 (`Then that particular upstream would not...` eluded me). I agree that your description above is more logically consistent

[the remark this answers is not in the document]
sunjayBhatia: @clayton-gonsalves wdyt about this distinction?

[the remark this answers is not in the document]
clayton-gonsalves: I agree! It does make sense to have it the default for all vhosts. It makes the behavior consistent. I have changed the wording to convey that.
Thanks for the reviews and support :)

skriss (approved):

sunjayBhatia (approved):

PR #5000, opened 2023-01-24
internal/dag: Add exact path match condition in HttpProxy

### LINKED ISSUE #4069: Exact path match for HTTPProxy
It'd be nice to have exact patch match for HTTPProxy.
HTTPProxy seems to support only prefix path match, even though exact path match is included in [this design document](https://github.com/projectcontour/contour/blob/main/design/routing-design.md#exact-match).
Now, I need to configure a route for exact '/' and a route for prefix '/' at the same time, and I have no idea how to do it with HTTPProxy.

arjunsalyan:
This adds support for exact path matching in the HttpProxy along side prefix matching

fixes #4069
Signed-off-by: Arjun Salyan <iarjunsalyan@gmail.com>

### DIFF
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -61,11 +61,15 @@ type Include struct {
 }
 
 // MatchCondition are a general holder for matching rules for HTTPProxies.
-// One of Prefix, Header or QueryParameter must be provided.
+// One of Prefix, Exact, Header or QueryParameter must be provided.
 type MatchCondition struct {
 	// Prefix defines a prefix match for a request.
 	// +optional
 	Prefix string `json:"prefix,omitempty"`
+
+	// Exact defines a exact match for a request.
+	// +optional
+	Exact string `json:"exact,omitempty"`
 
 	// Header specifies the header condition to match.
 	// +optional
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -68,6 +68,7 @@ type MatchCondition struct {
 	Prefix string `json:"prefix,omitempty"`
 
 	// Exact defines a exact match for a request.
+	// This field is not allowed in include match conditions.
 	// +optional
 	Exact string `json:"exact,omitempty"`
 
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -9117,6 +9117,86 @@ func TestDAGInsert(t *testing.T) {
 		},
 	}
 
+	// Invalid because has exact in include match conditions
+	proxy113 := &contour_api_v1.HTTPProxy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "example113-com",
+			Namespace: s1.Namespace,
+		},
+		Spec: contour_api_v1.HTTPProxySpec{
+			VirtualHost: &contour_api_v1.VirtualHost{
+				Fqdn: "example113.com",
+			},
+			Includes: []contour_api_v1.Include{{
+				Name: "kuarder",
+				Conditions: []contour_api_v1.MatchCondition{{
+					Exact: "/kuarder",
+				}},
+			}},
+		},
+	}
+
+	proxy114 := &contour_api_v1.HTTPProxy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "example114-com",
+			Namespace: s1.Namespace,
+		},
+		Spec: contour_api_v1.HTTPProxySpec{
+			VirtualHost: &contour_api_v1.VirtualHost{
+				Fqdn: "example114.com",
+			},
+			Includes: []contour_api_v1.Include{{
+				Name:      "proxy114a",
+				Namespace: s1.Namespace,
+				Conditions: []contour_api_v1.MatchCondition{{
+					Prefix: "/foo",
+				}},
+			}, {
+				Name:      "proxy114b",
+				Namespace: s2.Namespace,
+				Conditions: []contour_api_v1.MatchCondition{{
+					Prefix: "/bar",
+				}},
+			}},
+		},
+	}
+
+	proxy114a := &contour_api_v1.HTTPProxy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "proxy114a",
+			Namespace: s1.Namespace,
+		},
+		Spec: contour_api_v1.HTTPProxySpec{
+			Routes: []contour_api_v1.Route{{
+				Conditions: []contour_api_v1.MatchCondition{{
+					Exact: "/exact",
+				}},
+				Services: []contour_api_v1.Service{{
+					Name: s1.Name,
+					Port: 8080,
+				}},
+			}},
+		},
+	}
+
+	proxy114b := &contour_api_v1.HTTPProxy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "proxy114b",
+			Namespace: s2.Namespace,
+		},
+		Spec: contour_api_v1.HTTPProxySpec{
+			Routes: []contour_api_v1.Route{{
+				Conditions: []contour_api_v1.MatchCondition{{
+					Prefix: "/prefix",
+				}},
+				Services: []contour_api_v1.Service{{
+					Name: s2.Name,
+					Port: 8080,
+				}},
+			}},
+		},
+	}
+
 	cookieRewritePoliciesRoute := &contour_api_v1.HTTPProxy{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "example-com",
@@ -11799,6 +11879,51 @@ func TestDAGInsert(t *testing.T) {
 								},
 							),
 							routeCluster("/kuarder/withavengeance",
+								&Cluster{
+									Upstream: &Service{
+										Weighted: WeightedService{
+											Weight:           1,
+											ServiceName:      s2.Name,
+											ServiceNamespace: s2.Namespace,
+											ServicePort:      s2.Spec.Ports[0],
+											HealthPort:       s2.Spec.Ports[0],
+										},
+									},
+								},
+							),
+						),
+					),
+				},
+			),
+		},
+		"invalid httpproxy with exact condition in include match conditions": {
+			objs: []interface{}{proxy113, s1},
+			want: listeners(),
+		},
+		"httpproxy with include, exact and prefix on included proxy": {
+			objs: []interface{}{
+				proxy114, proxy114a, proxy114b, s1, s2,
+			},
+			want: listeners(
+				&Listener{
+					Name: HTTP_LISTENER_NAME,
+					Port: 8080,
+					VirtualHosts: virtualhosts(
+						virtualhost("example114.com",
+							routeClusterExact("/foo/exact",
+								&Cluster{
+									Upstream: &Service{
+										Weighted: WeightedService{
+											Weight:           1,
+											ServiceName:      s1.Name,
+											ServiceNamespace: s1.Namespace,
+											ServicePort:      s1.Spec.Ports[0],
+											HealthPort:       s1.Spec.Ports[0],
+										},
+									},
+								},
+							),
+							routeCluster("/bar/prefix",
 								&Cluster{
 									Upstream: &Service{
 										Weighted: WeightedService{
@@ -14738,6 +14863,13 @@ func routeCluster(prefix string, first *Cluster, rest ...*Cluster) *Route {
 	}
 }
 
+func routeClusterExact(exact string, first *Cluster, rest ...*Cluster) *Route {
+	return &Route{
+		PathMatchCondition: exactString(exact),
+		Clusters:           append([]*Cluster{first}, rest...),
+	}
+}
+
 func routeUpgrade(prefix string, first *Service, rest ...*Service) *Route {
 	r := prefixroute(prefix, first, rest...)
 	r.HTTPSUpgrade = true
@@ -14957,6 +15089,9 @@ func listeners(ls ...*Listener) []*Listener {
 
 func prefixString(prefix string) MatchCondition {
 	return &PrefixMatchCondition{Prefix: prefix, PrefixMatchType: PrefixMatchString}
+}
+func exactString(exact string) MatchCondition {
+	return &ExactMatchCondition{Path: exact}
 }
 func prefixSegment(prefix string) MatchCondition {
 	return &PrefixMatchCondition{Prefix: prefix, PrefixMatchType: PrefixMatchSegment}
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -14863,9 +14863,9 @@ func routeCluster(prefix string, first *Cluster, rest ...*Cluster) *Route {
 	}
 }
 
-func routeClusterExact(exact string, first *Cluster, rest ...*Cluster) *Route {
+func routeClusterExact(path string, first *Cluster, rest ...*Cluster) *Route {
 	return &Route{
-		PathMatchCondition: exactString(exact),
+		PathMatchCondition: exact(path),
 		Clusters:           append([]*Cluster{first}, rest...),
 	}
 }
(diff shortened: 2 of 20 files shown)


arjunsalyan:
This is by no means near completion. Still need to cover some edge cases as I feel I must have missed them, and work on the documentation. Any reviews and feedback would be very helpful. Also, please bear with me as this is my first PR.

design/httpproxy-status-conditions.md
```diff
@@ -277,7 +277,7 @@ IncludeError
 "include: %s", err (pathConditionsValid, path conditions check)
 "route: %s", err (route conditions check using pathConditionsValid)
   - "prefix conditions must start with /, %s was supplied", cond.Prefix
-  - "more than one prefix is not allowed in a condition block"
+  - "more than one prefix or exact is not allowed in a condition block"
```
sunjayBhatia, line 280: no need to modify old design docs for this change

sunjayBhatia (commented):
re the rules around merging path conditions in the tree of HTTPProxy includes:

```
// If the root condition is prefix, and included condition is prefix then after merging we get one prefix condition.
// If the root condition is prefix, and included condition is exact then after merging we get one exact condition.
// If the root condition is exact, and included condition is exact then after merging we get one exact condition.
// If the root condition is exact, and included condition is prefix then after merging we get one prefix condition.
```

an alternate interpretation of the "exact" match in this context could be that such a match is terminal, i.e. it must come as a leaf of the tree of match conditions, since as it stands above if an exact match is at the root or middle of an include tree it effectively becomes a prefix match. if we want to keep these things distinct/clear that might be a way to go

internal/dag/httpproxy_processor.go
```diff
@@ -1410,24 +1422,24 @@ func includeMatchConditionsIdentical(include contour_api_v1.Include, seenConds m
 		return false
 	})
 
-	seenHeaderConds, pathSeen := seenConds[pathPrefix]
+	seenHeaderConds, pathSeen := seenConds[mergedPath]
```
sunjayBhatia, line 1425: the `seenConds` map key will need some changes to differentiate they type of the merged path, since it is valid to have a prefix or an exact match for the same path string

internal/dag/conditions.go
```diff
@@ -22,37 +22,66 @@ import (
...
+// If the root condition is prefix, and included condition is exact then after merging we get one exact condition.
+// If the root condition is exact, and included condition is exact then after merging we get one exact condition.
+// If the root condition is exact, and included condition is prefix then after merging we get one prefix condition.
+// Hence this can be concluded that the last rule in the delegation chain decides the path type
+//
 // pathMatchConditionsValid guarantees that if a prefix is present, it will start with a
 // / character, so we can simply concatenate.
-func mergePathMatchConditions(conds []contour_api_v1.MatchCondition) *PrefixMatchCondition {
-	prefix := ""
+func mergePathMatchConditions(conds []contour_api_v1.MatchCondition) MatchCondition {
```
sunjayBhatia, line 35: we'll definitely need some thorough unit tests to make sure this logic is well defined and has proper coverage

sunjayBhatia:
Thanks for the PR @arjunsalyan I think this is a good start, but we will definitely need some more unit test coverage, we can give pointers where to look to add these if you like

we should already have logic further in our processing logic to sort the generated routes properly so that exact matches are considered by Envoy first, so this should mostly be covered by httpproxy processor/dag tests and some end to end tests

internal/dag/conditions.go
arjunsalyan: > re the rules around merging path conditions in the tree of HTTPProxy includes:
> 
> ```
> // If the root condition is prefix, and included condition is prefix then after merging we get one prefix condition.
> // If the root condition is prefix, and included condition is exact then after merging we get one exact condition.
> // If the root condition is exact, and included condition is exact then after merging we get one exact condition.
> // If the root condition is exact, and included condition is prefix then after merging we get one prefix condition.
> ```
> 
> an alternate interpretation of the "exact" match in this context could be that such a match is terminal, i.e. it must come as a leaf of the tree of match conditions, since as it stands above if an exact match is at the root or middle of an include tree it effectively becomes a prefix match. if we want to keep these things distinct/clear that might be a way to go

Agreed @sunjayBhatia, If you see the next line ahead of these comments I have mentioned the interpretation. These comments are just to explain why we reached the conclusion that the leaf decides the path type.

```
// Hence this can be concluded that the last rule in the delegation chain decides the path type
```

Let me know if you meant something else.

arjunsalyan:
> re the rules around merging path conditions in the tree of HTTPProxy includes:
> 
> ```
> // If the root condition is prefix, and included condition is prefix then after merging we get one prefix condition.
> // If the root condition is prefix, and included condition is exact then after merging we get one exact condition.
> // If the root condition is exact, and included condition is exact then after merging we get one exact condition.
> // If the root condition is exact, and included condition is prefix then after merging we get one prefix condition.
> ```
> 
> an alternate interpretation of the "exact" match in this context could be that such a match is terminal, i.e. it must come as a leaf of the tree of match conditions, since as it stands above if an exact match is at the root or middle of an include tree it effectively becomes a prefix match. if we want to keep these things distinct/clear that might be a way to go

Agreed @sunjayBhatia, If you see the next line ahead of these comments I have mentioned the interpretation. These comments are just to explain why we reached the conclusion that the leaf decides the path type.

```
// Hence this can be concluded that the last rule in the delegation chain decides the path type
```

Let me know if you meant something else.

sunjayBhatia:
> > re the rules around merging path conditions in the tree of HTTPProxy includes:
> > ```
> > // If the root condition is prefix, and included condition is prefix then after merging we get one prefix condition.
> > // If the root condition is prefix, and included condition is exact then after merging we get one exact condition.
> > // If the root condition is exact, and included condition is exact then after merging we get one exact condition.
> > // If the root condition is exact, and included condition is prefix then after merging we get one prefix condition.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > an alternate interpretation of the "exact" match in this context could be that such a match is terminal, i.e. it must come as a leaf of the tree of match conditions, since as it stands above if an exact match is at the root or middle of an include tree it effectively becomes a prefix match. if we want to keep these things distinct/clear that might be a way to go
> 
> Agreed @sunjayBhatia, If you see the next line ahead of these comments I have mentioned the interpretation. These comments are just to explain why we reached the conclusion that the leaf decides the path type.
> 
> ```
> // Hence this can be concluded that the last rule in the delegation chain decides the path type
> ```
> 
> Will improve the wording of this statement. Let me know if you meant something else.

Above I meant that an option we can take is to not allow and mark an HTTPProxy as Invalid if you try to use an exact match in an include match condition

arjunsalyan:
> Above I meant that an option we can take is to not allow and mark an HTTPProxy as Invalid if you try to use an exact match in an include match condition

Got it. I have some subjective thoughts here, especially for the case of delegation of a path to teams. Included routes should have the flexibility to use exact matches. But open to any other views on this.

I have added unit tests and e2e trying to cover most of the things. Will start with documentation once these things come to good shape.

arjunsalyan:
For now I have added the docs considering my initial theory about routes delegation, just to make this PR ready for a review. Although we can revisit it.

sunjayBhatia (commented):
I think my interpretation of "Exact" path matching in conjunction with the Inclusion feature is that if you are using Inclusion, you really want a "Prefix" match (because the Included child proxy can then do different matching) so "Exact" matches should be disallowed on the path match conditions on an Include

cc @skriss @tsaarni if you have any thoughts

arjunsalyan:
> > I think my interpretation of "Exact" path matching in conjunction with the Inclusion feature is that if you are using Inclusion, you really want a "Prefix" match (because the Included child proxy can then do different matching) so "Exact" matches should be disallowed on the path match conditions on an Include
> > cc @skriss @tsaarni if you have any thoughts
> 
> Yeah, that is how I had been thinking about it too, should be disallowed as an Include condition.

Alright, I am starting to see the point here. But before we reach a conclusion, please allow me to put one last statement in the favour of the proposed merging algorithm.

- Since we have a strict rule of one fqdn - one httpproxy, inclusion becomes the only way to route traffic to services in different namespaces from the same fqdn. To adopt a general approach, we currently follow a methodology where we create an orphan httpproxy without any rules in all of the namespaces. Then we create a root httpproxy in one of the namespaces and define all the rules in that root httpproxy. Not having the exact condition in include takes away this flexibility.

There are definitely other ways to handle this, but just putting a use case.

Anyways, I am now aligned on not allowing the `Exact` in the Include condition. Will push the changes soon.

arjunsalyan:
Hi @skriss @sunjayBhatia 

The changes for disallowing the `exact` condition except for the child/leaf httproxy are done.

site/content/docs/main/config/request-routing.md
```diff
@@ -80,6 +86,15 @@ Up to one prefix condition may be present in any condition block.
 
 Prefix conditions **must** start with a `/` if they are present.
 
+#### Exact conditions
+
+Paths defined are matched using exact conditions.
+Up to one exact condition may be present in any condition block. Any condition block can
+either have an exact condition or prefix condition, but not both together. Exact conditions are
+not allowed in root httpproxies or in includes blocks.
```
sunjayBhatia, line 94:
```suggestion
only allowed in route match conditions and not in include match conditions.
```

site/content/docs/main/config/inclusion-delegation.md
```diff
@@ -22,10 +22,11 @@ This process is recursive.
...
 When conditions are combined through inclusion Contour merges the conditions inherited via inclusion with any conditions specified on the route.
-This may result in duplicates, for example two `prefix:` conditions, or two header match conditions with the same name and value.
+This may result in duplicates, for example two `prefix:` conditions, mix of both `prefix:` and `exact` conditions, or two header match conditions with the same name and value.
 To resolve this Contour applies the following logic.
 
 - `prefix:` conditions are concatenated together in the order they were applied from the root object. For example the conditions, `prefix: /api`, `prefix: /v1` becomes a single `prefix: /api/v1` conditions. Note: Multiple prefixes cannot be supplied on a single set of Route conditions.
+- `exact:` conditions are also concatenated just like `prefix:` conditions, but `exact:` conditions are not allowed in the root httpproxy or in the inlcusion block. If the child httpproxy has `exact:` condition then after concatenation, it becomes a single `exact:` condition. For example, `prefix: /static` and `exact: /main.js` become a single `exact: /static/main.js` condition.
```
sunjayBhatia, line 29:
```suggestion
- `exact:` conditions are also concatenated just like `prefix:` conditions, but `exact:` conditions are not allowed in include match conditions. If the child httpproxy has `exact:` condition then after concatenation, it becomes a single `exact:` condition. For example, `prefix: /static` and `exact: /main.js` become a single `exact: /static/main.js` condition.
```

apis/projectcontour/v1/httpproxy.go
```diff
@@ -61,12 +61,16 @@ type Include struct {
...
 
 // MatchCondition are a general holder for matching rules for HTTPProxies.
-// One of Prefix, Header or QueryParameter must be provided.
+// One of Prefix, Exact, Header or QueryParameter must be provided.
 type MatchCondition struct {
 	// Prefix defines a prefix match for a request.
 	// +optional
 	Prefix string `json:"prefix,omitempty"`
 
+	// Exact defines a exact match for a request.
```
sunjayBhatia, line 70: would be good to note here that this field cannot be used in an include match condition

internal/dag/httpproxy_processor.go
```diff
@@ -624,6 +624,12 @@ func (p *HTTPProxyProcessor) computeRoutes(
 			namespace = proxy.Namespace
 		}
 
+		if err := includeMatchConditionsValid(include.Conditions); err != nil {
+			validCond.AddErrorf(contour_api_v1.ConditionTypeIncludeError, "PathMatchConditionsNotValid",
```
sunjayBhatia, line 628: we should add a test in `status_test.go` for this error case

internal/dag/httpproxy_processor.go
```diff
@@ -624,6 +624,12 @@ func (p *HTTPProxyProcessor) computeRoutes(
 			namespace = proxy.Namespace
 		}
 
+		if err := includeMatchConditionsValid(include.Conditions); err != nil {
```
sunjayBhatia, line 627: we should add a test in `builder_test.go` for the changes in this file, there should be example tests in there for prefix matching that can be used as a template

sunjayBhatia: in general since codespell/codegen have failed, we aren't seeing the coverage report which I think will show we need some more unit tests for changes in this file

internal/dag/httpproxy_processor.go
```diff
@@ -1412,7 +1423,15 @@ type matchConditionAggregate struct {
 }
 
 func includeMatchConditionsIdentical(includeConds []contour_api_v1.MatchCondition, seenConds map[string][]matchConditionAggregate) bool {
-	pathPrefix := mergePathMatchConditions(includeConds).Prefix
+	mergedPath := ""
+
```
sunjayBhatia, line 1427: we should add some unit testing to `TestIncludeMatchConditionsIdentical` for the changes in this function

arjunsalyan: I was wondering if this function even needs to be changed now. Because include conditions will only have prefix conditions and not exact conditions, I don't see a need to change this function. But since `mergePathMatchConditions` has the possibility to return an `ExactMatchCondition`, so we should just add something to handle that part.

sunjayBhatia: I think we may be able to rely on the fact that include match conditions should never have exact path matches in them, validation should happen before this is called so defo fair point there shouldn't really be changes to this

arjunsalyan:
@sunjayBhatia Added the generated sources, which were leading to the failed test.

arjunsalyan:
This is strange

```
Build and Test Pull Request / test-linux (pull_request)
Build and Test Pull Request / test-osx (pull_request)
```

These tests pass all fine when I run them locally. Looking into it

sunjayBhatia:
> This is strange
> 
> ```
> Build and Test Pull Request / test-linux (pull_request)
> Build and Test Pull Request / test-osx (pull_request)
> ```
> 
> These tests pass all fine when I run them locally. Looking into it

you'll want to merge/rebase on main, and update the port in that test (github is merging the PR with the base branch when checking out the code)

arjunsalyan:
> you'll want to merge/rebase on main, and update the port in that test (github is merging the PR with the base branch when checking out the code)

Ah! The tests were updated. Fixed

test/e2e/httpproxy/exact_path_condition_match_test.go
```diff
@@ -0,0 +1,158 @@
...
+							},
+						},
+					},
+				},
+			},
+		}
+
+		f.CreateHTTPProxyAndWaitFor(serviceProxy, e2e.HTTPProxyValid)
+
+		cases := map[string]string{
```
sunjayBhatia, line 132: nit: I think this looks correct, but it's not super intuitive to read

not a huge deal but would be nice to either add some comments above some of the cases or use naming of the routes/services that make it clearer

test/e2e/httpproxy/include_exact_condition_test.go
```diff
@@ -0,0 +1,215 @@
...
+							},
+						},
+					},
+				},
+			},
+		}
+		f.CreateHTTPProxyAndWaitFor(baseProxy, e2e.HTTPProxyValid)
+		f.CreateHTTPProxyAndWaitFor(invalidRootProxy, e2e.HTTPProxyInvalid)
+
+		cases := map[string]string{
```
sunjayBhatia, line 191: similar nit to the above on comments/naming to make these test cases super clear (looks correct though)

internal/dag/builder_test.go
```diff
@@ -14958,6 +15090,9 @@ func listeners(ls ...*Listener) []*Listener {
 func prefixString(prefix string) MatchCondition {
 	return &PrefixMatchCondition{Prefix: prefix, PrefixMatchType: PrefixMatchString}
 }
+func exactString(exact string) MatchCondition {
```
sunjayBhatia, line 15093: nit: the `prefixString` was named since it represented a prefix match that would basically match a path using "string prefix" vs. matching path segments, we could change this to just `exact` here

sunjayBhatia (approved):
looks good overall, just some tiny nits

approving but will leave for another maintainer to have a pass over as well before merge

internal/dag/builder_test.go
arjunsalyan: Makes sense, rather `exact()` is already there making it completely redundant.

test/e2e/httpproxy/exact_path_condition_match_test.go
arjunsalyan: Can agree to that, it is just the nature of a few cases that I could not find better naming. Have tried something with comments, not sure how much value that adds though.

skriss (approved):
LGTM, thanks @arjunsalyan!

Issue #2497, opened 2020-05-04
Support for running Contour in a single namespace


josephglanville:
We would like to deploy independent Contour deployments per namespace. Our use case for this is we have a (set of) multi-tenant clusters with namespaces divided by teams. Each team has a large degree of freedom in which tools they wish to adopt. This is ok for the vast majority of tools which can be deployed in modes that don't require cluster wide access (except maybe an admin to create Custom Resource Definitions).

One of the teams has already adopted Contour and did their best to deploy it in a way that would allow others to but unfortunately they miss-understood the nature of `--root-namespaces`, the problems with which I will describe below.

Currently Contour is designed and only capable of running as a cluster-wide service. It expects to be able to list all of it's managed resources in all namespaces. This requires it to have a ClusterRole with broad permissions. In addition this makes it very impractical to run more than one Contour instance in a cluster because both will be attempting to manage all resources and if they differ in `--root-namespaces` settings will mark each others root `HTTPProxy` objects as invalid.

This cannot be solved with RBAC in the current form as if Contour fails to list resources in all namespaces it will fail to operate correctly.

Our ideal solution would be to deploy Contour + Envoy isolated in each namespace without conflicts from each other with only the CRDs being defined globally, i.e using basic Role instead of ClusterRole.

This feature is already present in the other solution we are looking at, Ambassador: https://www.getambassador.io/docs/latest/topics/running/running/ via it's `AMBASSADOR_SINGLE_NAMESPACE` environment variable and largely inspired my approach in https://github.com/projectcontour/contour/pull/2492 as an attempt to fufil this requirement.

stevesloka:
Hey @josephglanville! This has come up before with similar requirements. One thing you could do today is to add `ingress.class` to your resources. If you set that arg on Contour, it will only watch for ingress resources that match your class. I understand that's not ideal, but might help get through your current needs. 

I think having a flag is good to restrict the surface area that Contour needs to watch but might make it a set of namespaces, rather than a single one. This way we could fit the needs of other users who might need to restrict to a few. 

We do that now with `secrets` when using the `root-namespaces` arg. If you restrict where your roots live, then Contour only watches for secrets in those corresponding namespaces. 

I'd like to see what @youngnick & @jpeach think about this so we can come up with a way to move forward.

josephglanville:
If I am not mistaken `ingress.class` only works with `Ingress` resources. We would like to use the `HTTPProxy` CRDs. I don't believe this annotation is honored on those.

[...]

Issue #4899, opened 2022-12-06
Question about "installing multiple Contour instances"


izturn:
**What question do you have?:**
     With https://github.com/projectcontour/contour/pull/4832, we can installing multiple Contour instances per cluster, but the `informOnResource` for resources such as `ExtensionService` had not namespaced. It leads to one contour watch the events for the others.
     Should we namespaced for them, like for `Secret`?
     If you all like this idea, I will send a pr.

sunjayBhatia:
If we do this we will have to have some sort of flag letting Contour know it is intended to only run in a single namespace. We will also need to make a decision on and document what it means for inclusion/delegation, TLSCertificateDelegation, GW API ReferenceGrants, etc. when running in this mode, since those are intended cross-namespace reference mechanisms.

skriss:
Note, this is being addressed in #5214

PR #5002, opened 2023-01-25
Allow limiting K8s watches on specific namespaces.

tsaarni:
This PR adds an option for users to limit which namespaces `contour serve` will attempt to access. Following use cases could benefit from this feature:

* Allows Contour to be used in a restricted environment where it is not acceptable to grant `ClusterRole` and `ClusterRoleBinding` to Contour.
* It could be used to limit the number of resources processed by Contour.

An example of the first use case: Users want to install their own dedicated instances of Contour together with their application into their own namespace in a soft-multitenant cluster. Cluster administrator might be willing to installing the CRDs for users, but due to security reasons it is unacceptable to grant `ClusterRoles` that give users access to resources in other namespaces, most importantly it would allow the user to read *all* `Secrets` in the cluster. 

The implementation uses the controller-runtime feature to set the watched `Namespace` option when only single namespace needs to be watched, or alternatively it uses `MultiNamespacedCacheBuilder` when more than one namespace needs to be watched. With some preliminary testing, it seems to work. To test the feature, kustomize file is added to the examples. It simply changes the existing `ClusterRole` and `ClusterRoleBinding` to namespaced `Role` and `RoleBinding`.

The same feature seems to be supported by NGINX with `-watch-namespace` (see [link](https://docs.nginx.com/nginx-ingress-controller/installation/running-multiple-ingress-controllers/)).

The issue has been previously discussed in #2497 and https://github.com/projectcontour/contour-operator/issues/95.

**Marking the PR as draft, since it is just a PoC, but I'd be curious to hear any feedback.**

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -156,6 +156,7 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 	serve.Flag("leader-election-retry-period", "The interval which Contour will attempt to acquire leadership lease.").Default("2s").DurationVar(&ctx.LeaderElection.RetryPeriod)
 
 	serve.Flag("root-namespaces", "Restrict contour to searching these namespaces for root ingress routes.").PlaceHolder("<ns,ns>").StringVar(&ctx.rootNamespaces)
+	serve.Flag("watch-namespaces", "Restrict contour to search resources in these namespaces only.").PlaceHolder("<ns,ns>").StringVar(&ctx.watchNamespaces)
 
 	serve.Flag("stats-address", "Envoy /stats interface address.").PlaceHolder("<ipaddr>").StringVar(&ctx.statsAddr)
 	serve.Flag("stats-port", "Envoy /stats interface port.").PlaceHolder("<port>").IntVar(&ctx.statsPort)
@@ -196,11 +197,26 @@ func NewServer(log logrus.FieldLogger, ctx *serveContext) (*Server, error) {
 		return nil, fmt.Errorf("unable to create scheme: %w", err)
 	}
 
+	var cacheFunc ctrl_cache.NewCacheFunc
+	var namespaces string
+	if watchedNamespaces := ctx.watchedNamespaces(); watchedNamespaces != nil {
+		if len(watchedNamespaces) == 1 {
+			namespaces = watchedNamespaces[0]
+		}
+		if len(watchedNamespaces) > 1 {
+			cacheFunc = ctrl_cache.MultiNamespacedCacheBuilder(watchedNamespaces)
+		}
+		log.WithField("context", "watched-namespaces").Infof("watching namespaces %q", watchedNamespaces)
+
+	}
+
 	// Instantiate a controller-runtime manager.
 	options := manager.Options{
 		Scheme:                 scheme,
 		MetricsBindAddress:     "0",
 		HealthProbeBindAddress: "0",
+		Namespace:              namespaces,
+		NewCache:               cacheFunc,
 	}
 	if ctx.LeaderElection.Disable {
 		log.Info("Leader election disabled")
--- a/cmd/contour/servecontext.go
+++ b/cmd/contour/servecontext.go
@@ -62,6 +62,9 @@ type serveContext struct {
 	// httpproxy root namespaces
 	rootNamespaces string
 
+	// Watch only these namespaces to allow running with limited RBAC permissions.
+	watchNamespaces string
+
 	// ingress class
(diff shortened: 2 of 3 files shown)


tsaarni:
While following is related to `Secrets` only and not all the other resources that the `ClusterRole` gives permission to, there is existing discussion about limited RBAC role in relation to "root namespaces" feature ([link](https://github.com/projectcontour/contour/blob/40df85985c0ec632fa57b04a1a77a8c1f3c217c7/site/content/docs/main/config/virtual-hosts.md#L130-L132)). It also includes example RBAC change with namespaced access for `Secrets` ([link](https://github.com/projectcontour/contour/blob/40df85985c0ec632fa57b04a1a77a8c1f3c217c7/examples/root-rbac/rbac.yaml#L28-L41)).

> Additionally, when defined, Contour will only watch for Kubernetes secrets in these namespaces ignoring changes in all other namespaces. Proper RBAC rules should also be created to restrict what namespaces Contour has access matching the namespaces passed to the command line flag.

This is not true anymore, not in the way it was written.  Back then filtering was based on creating `client-go` informer that was limited to namespace https://github.com/projectcontour/contour/commit/3cc619dae6220b7593bd11e32f002d709dd54537. After controller-runtime was taken into use the filtering is done at client end ([link](https://github.com/projectcontour/contour/blob/40df85985c0ec632fa57b04a1a77a8c1f3c217c7/internal/k8s/filter.go#L26-L44)). Controller-runtime will attempt to access `Secrets` regardless of namespace and it fails with the example namespaced Role.

skriss:
cc @izturn

nsimons:
I've been testing this change and discussed with @tsaarni offline that I can try to finalize this item.

Basically we have three kinds of use cases to take into consideration
1) inform on multiple namespaces (`--watch-namespaces=foo,bar,baz`)
2) only inform on a single namespace (`--watch-namespaces=foo`)
3) current Contour default, inform on all namespaces

We also have the `--root-namespaces` flag to throw into the mix; i.e. further restrict the namespaces where root `HTTPProxies` and their associated `Secrets` can live. To me it has become clear that the namespaces defined by `--root-namespaces` need to be a subset of the namespaces that we inform on (e.g. `--watch-namespaces=foo,bar`, `--root-namespaces=foo`), otherwise Contour wouldn't know about them(!). This warrants a check in the code at startup and some documentation. In (2) it wouldn't make sense to set `--root-namespaces`, but it shouldn't harm either.

I've also been investigating the things that @sunjayBhatia mentioned in #4899:

> We will also need to make a decision on and document what it means for inclusion/delegation, TLSCertificateDelegation, GW API ReferenceGrants, etc. when running in this mode, since those are intended cross-namespace reference mechanisms.

For a single namespace (2) it does not make sense to create `TLSCertificateDelegation/ReferenceGrant` resources; everything happens within the same namespace. I think this should be obvious to the user without the need for any extra documentation (but we could of course also mention it). Inclusion/delegation would work as usual, i.e. you can reference root/leaf resources within the same namespace.

For multiple namespaces (1) (well technically (2) as well) there is a risk that the user tries to reference resources in a namespace not covered by the informers (not in `--watch-namespaces`), and creating resources in those namespaces. 

I guess the question is, should we have some separate error message for this scenario?

With the patch as is, this scenario manifests itself as a kind of not-found error in the resource statuses. E.g. root `HTTPProxy` reports `IncludeNotFound` if the leaf `HTTPProxy` is in the wrong namespace (the leaf being in `NotReconciled` status), or `SecretNotValid` for a secret. In Gateway API the `Gateway` will be OK, but the `HTTPRoute` in the wrong namespace will not report any status.

I'm inclined to answer no, with this reasoning:
- Currently, in (3) with `--root-namespaces` defined, a secret referenced in a namespace not covered by `--root-namespaces` will simply be reported as `SecretNotValid`.
- `TLSCertificateDelegation/ReferenceGrant` resources does not seem to get any status field updates, so these statuses are empty regardless. Also there are no errors if you give a non-existent name/namespace.
- `HTTPRoute` with wrong `parentRefs` will have empty status

So it seems to me that we are already expecting the user to "tackle" these issues without an explicit error that pin-points the different problem.

Any thoughts on this?

I was also thinking a bit how the different users in the cluster would know which namespaces are being watched by Contour, but I guess this info has to travel in a backchannel of sorts? Similar to how an application developer would know the correct `parentRefs` name to point towards in the `HTTPRoute`.

skriss:
@nsimons what you describe sounds reasonable to me.

tsaarni:
The development of this feature continues in #5214.

PR #5004, opened 2023-01-25
internal/metrics: drop empty labels from dagRebuildGauge

### LINKED ISSUE #2238: internal/metrics: contour_dagrebuild_timestamp should not be defined as a GaugeVec
The `contour_dagrebuild_timestamp` is defined as a Guage Vec, it should probably not be.

skriss:
The dagRebuildGauge metric was being constructed
as a GaugeVec, but with an empty label slice. This 
drops the empty label slice and changes to a plain Gauge.
There is no impact for users.

Closes #2238.

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -33,7 +33,7 @@ type Metrics struct {
 	proxyValidGauge     *prometheus.GaugeVec
 	proxyOrphanedGauge  *prometheus.GaugeVec
 
-	dagRebuildGauge             *prometheus.GaugeVec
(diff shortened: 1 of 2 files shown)


sunjayBhatia (approved):

skriss:
@sunjayBhatia same q on this one, are you OK with merging this now for inclusion in 1.24? Just wanted to check since we've already cut the RC.

sunjayBhatia:
> @sunjayBhatia same q on this one, are you OK with merging this now for inclusion in 1.24? Just wanted to check since we've already cut the RC.

yeah seems low risk IMO

PR #5008, opened 2023-01-25
httpproxy: Add support for ip-based filtering

### LINKED ISSUE #3693: Add IP allowlist and blocklist functionality

ecordell:
Configures Envoy's `envoy.filters.http.rbac` per route via HTTPProxy.

See #4990 for design 

Fixes #3693


sunjayBhatia (commented):
this looks great so far, awesome to have an implementation to play with

ecordell:
I'll rebase this when the design is finalized

skriss:
@ecordell thanks for the PR! One thing to put on your radar is that we will want to add some user documentation as part of this PR -- should be a new page in `site/content/docs/main/config`, and added to the TOC via `site/data/docs/main-toc.md`.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -550,6 +574,26 @@ type JWTVerificationPolicy struct {
...
+	// Source indicates how to determine the ip address to filter on, and can be
+	// one of two values:
+	//  - `Remote` filters on the ip address of the client, accounting for PROXY and
+	//    X-Forwarded-For as needed.
+	//  - `Peer` filters on the ip of the network request, ignoring PROXY and
+	//    X-Forwarded-For.
+	Source IPFilterSource `json:"source"`
+
+	// CIDR is a CIDR block of a ipv4 or ipv6 addresses to filter on.
+	CIDR string `json:"cidr,omitempty"`
```
skriss, line 594: Unless I'm mistaken this field is required, so let's drop the `omitempty`.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -550,6 +574,26 @@ type JWTVerificationPolicy struct {
 	Disabled bool `json:"disabled,omitempty"`
 }
 
+type IPFilterSource string
```
skriss, line 577: Let's add kubebuilder enum validation here:

`// +kubebuilder:validation:Enum=Peer;Remote`

test/e2e/httpproxy/ip_filtering_test.go
```diff
@@ -0,0 +1,399 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/util/retry"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	contourv1 "github.com/projectcontour/contour/apis/projectcontour/v1"
+	"github.com/projectcontour/contour/test/e2e"
+)
+
+func testIPFilterPolicy(namespace string) {
+	FSpecify("requests can be filtered by ip address", func() {
```
skriss, line 34:
```suggestion
	Specify("requests can be filtered by ip address", func() {
```

skriss (changes requested):

ecordell:
@skriss thanks for the initial review! addressed everything that you noted.

test/e2e/httpproxy/ip_filtering_test.go
```diff
@@ -0,0 +1,399 @@
...
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/util/retry"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	contourv1 "github.com/projectcontour/contour/apis/projectcontour/v1"
+	"github.com/projectcontour/contour/test/e2e"
+)
+
+func testIPFilterPolicy(namespace string) {
```
sunjayBhatia, line 33: these tests will all run in the same namespace and the HTTPProxies have the same fqdn, so I fear we will get some flakiness in CI long-term due to cross-test pollution (configuration for one test is not updated when another is running causing it to fail)

could be a cause of this run of failures, but I would have to look into it more: https://github.com/projectcontour/contour/actions/runs/4579282622/jobs/8086947055?pr=5008

test/e2e/httpproxy/ip_filtering_test.go
```diff
@@ -0,0 +1,399 @@
...
+			Condition: e2e.HasStatusCode(403),
+			RequestOpts: []func(*http.Request){
+				e2e.OptSetHeaders(map[string]string{"X-Forwarded-For": "10.10.10.0"}),
+			},
+		})
+		require.NotNil(t, res, "request never succeeded")
+		require.Truef(t, ok, "expected 403 response code, got %d", res.StatusCode)
+	})
+
+	Specify("requests can be filtered by ip address in included routes", func() {
```
sunjayBhatia, line 278: maybe better as a featuretest, but can we add a test where both the root and child proxy have ip allow/deny rules? who takes precedence in that case/what are the semantics there?

internal/dag/httpproxy_processor.go
```diff
@@ -171,7 +172,12 @@ func (p *HTTPProxyProcessor) computeHTTPProxy(proxy *contour_api_v1.HTTPProxy) {
 				"Spec.VirtualHost.JWTProviders can only be defined for root HTTPProxies that terminate TLS")
 			return
 		}
+	}
 
+	if len(proxy.Spec.VirtualHost.IPAllowFilterPolicy) > 0 && len(proxy.Spec.VirtualHost.IPDenyFilterPolicy) > 0 {
+		validCond.AddError(contour_api_v1.ConditionTypeIPFilterError, "IncompatibleIPAddressFilters",
```
sunjayBhatia, line 178: some tests in `status_test.go` would be nice for coverage over the error cases we've added

test/e2e/httpproxy/ip_filtering_test.go
ecordell: > but can we add a test where both the root and child proxy have ip allow/deny rules? who takes precedence in that case/what are the semantics there?

Only the parent httproxy can have virtualhost-level rules, and any rules in child httpproxies will be per-route (and unique, otherwise validation on the overall httproxy would fail).

There's no contour-level de-tangling of ip filter rules if that makes sense

test/e2e/httpproxy/ip_filtering_test.go
skriss: Yep, it does like like (a) giving each of the root proxies a unique FQDN; and (b) giving each proxy a unique name makes these tests pass.

skriss:
@ecordell just a friendly ping that the end of this release cycle is rapidly approaching; if you could address the above comments and fix the small merge conflict then we can take another look here. Thanks!

apis/projectcontour/v1/httpproxy.go
```diff
@@ -550,6 +574,28 @@ type JWTVerificationPolicy struct {
...
+type IPFilterPolicy struct {
+	// Source indicates how to determine the ip address to filter on, and can be
+	// one of two values:
+	//  - `Remote` filters on the ip address of the client, accounting for PROXY and
+	//    X-Forwarded-For as needed.
+	//  - `Peer` filters on the ip of the network request, ignoring PROXY and
+	//    X-Forwarded-For.
+	Source IPFilterSource `json:"source"`
+
+	// CIDR is a CIDR block of a ipv4 or ipv6 addresses to filter on.
```
skriss, line 595: It's worth explicitly noting here that this can also be a single IP

test/e2e/httpproxy/ip_filtering_test.go
skriss: From my testing, it looks like the presence of any ip allow/deny policies at the route level completely override any vhost-level settings, so if you specify anything for a route, the vhost-level settings are effectively ignored for that route. There's no merging.

ecordell:
@skriss Addressed the comments and rebased, please let me know if you see anything else I need to do to get this into the next release

test/e2e/httpproxy/ip_filtering_test.go
SAMZONG: maybe this result need into projectcoutour document.

ecordell: There's a section in the user docs for this now: https://github.com/projectcontour/contour/blob/eff7460f1389a2e1b63a82eda33d4379f91188e4/site/content/docs/main/config/ip-filtering.md

ecordell:
@sunjayBhatia @skriss any final comments? I want to make sure I've covered everything so this can get into the next release, feel free to ping me in kube slack (same username) if there's anything else you'd like me to address.

go.mod
```diff
@@ -2,6 +2,9 @@ module github.com/projectcontour/contour
 
 go 1.19
 
+// remove once https://github.com/cert-manager/cert-manager/issues/5953 is fixed
+replace github.com/Venafi/vcert/v4 => github.com/jetstack/vcert/v4 v4.9.6-0.20230127103832-3aa3dfd6613d
```
sunjayBhatia, line 6: huh never end up running `go list ...` very often but confirmed this is an issue

sunjayBhatia (approved):

skriss:
> @skriss Addressed the comments and rebased, please let me know if you see anything else I need to do to get this into the next release

Thanks @ecordell! Looks like @sunjayBhatia approved and I will take a final look as soon as I can (at KubeCon EU right now). We definitely want to get this into the upcoming release as well.

skriss (approved):
thanks for the updates @ecordell, LGTM. One last tiny merge conflict to resolve and this should be ready for merge

PR #5009, opened 2023-01-26
add summary metric for DAG rebuild duration

### LINKED ISSUE #4059: add metric for DAG processing duration
It'd be nice to have a metric capturing the duration of DAG rebuilds. We currently have metrics counting the number of DAG rebuilds and capturing the timestamp of the last DAG rebuild, but nothing that captures durations. This would be useful for understanding & debugging performance isuses.

skriss:
Adds a Summary metric contour_dagrebuild_seconds
to capture DAG rebuild duration quantiles.

Closes #4059.


internal/dag/builder.go
```diff
@@ -65,6 +70,11 @@ func (b *Builder) Build() *DAG {
 		StatusCache:        status.NewCache(gatewayNSName, gatewayController),
 	}
 
+	if b.Metrics != nil {
```
skriss, line 73: This was the easiest way to not have to fix a bunch of tests; alternately, can create a wrapper struct that handles the metrics part like is done for the xds cache update metrics, but it requires switching a bunch of stuff to interfaces and this seemed simpler. Let me know if anyone prefers otherwise.

internal/metrics/metrics.go
```diff
@@ -134,6 +136,24 @@ func NewMetrics(registry *prometheus.Registry) *Metrics {
 				Help: "Total number of times DAG has been rebuilt since startup",
 			},
 		),
+		DAGRebuildSeconds: prometheus.NewSummary(
+			prometheus.SummaryOpts{
+				Name: DAGRebuildSeconds,
+				Help: "Duration in seconds of DAG rebuilds",
+				Objectives: map[float64]float64{
+					0.1:  0.01,
```
sunjayBhatia, line 144: thoughts on choice/granularity of quantiles here? i'm not sure how useful some of the in between percentiles might be, wdyt about doing something like 0, .25, .5, .75, .9, .95, .99, 1? (so a little less granularity but still a good set of summary buckets) could omit the 0 and 1 quantiles if we dont care about max/min values

skriss: No particular reasoning on the chosen values, was just looking to have enough data to understand what's happening -- happy to switch to yours

skriss: See https://github.com/projectcontour/contour/pull/5009/commits/5389bd0f2a5ca20d529dcc524911fa0e7699e11b

skriss:
Also just FWIW as a reference point, I was looking at these on a cluster with ~thousands of HTTProxies and Services, and seeing P99 DAG build times under half a second.

PR #5010, opened 2023-01-26
Internal redirect implementation

### LINKED ISSUE #4843: Add support for Envoy internal_redirect_action

Envoy provide a way to configure [Internal redirect](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/http_connection_management#arch-overview-internal-redirects) policy. This is used to provide a way for an upstream server to return content from an other server.

For instance, you may have an upstream server that compute a S3 object storage URL, and instructs envoy to return the content of this URL as response to the downstream request. This is especially useful when the S3 storage is not publicly exposed and can only be accessed from internal server.

Today, I don't see how to configure a [`internal-redirect-policy`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routeaction-internal-redirect-policy) for a route defined by either an `HTTPRoute`, or an `HTTPProxy` object.

Jean-Daniel:
This is the implementation of the "internal redirect" specification. 

I split it in 3 commit to simplify review. 

The E2E tests requires a service that returns a redirect result code. I don't think this is something the echo image can do, and to avoid having to build a custom image just for that use case, the test instead rely on contour `RedirectResponse` and create a Service pointing to itself.

Closes #4843.


sunjayBhatia:
> The E2E tests requires a service that returns a redirect result code. I don't think this is something the echo image can do, and to avoid having to build a custom image just for that use case, the test instead rely on contour RedirectResponse and create a Service pointing to itself.

You should be able to use the `/status/<code>` endpoint of the echo server: https://github.com/kubernetes-sigs/ingress-controller-conformance/blob/7ee2f3e2d283acb203be862cacd593c71afb1311/images/echoserver/echoserver.go#L94

Jean-Daniel:
> You should be able to use the /status/<code> endpoint 

Unfortunately not. Internal redirect tests require that the `Location` header be set in the response,  and the echo server `/status` endpoint does not support `X-Echo-Set-Header`.

sunjayBhatia:
> > You should be able to use the `/status/<code>` endpoint of the echo server
> 
> Unfortunately not. Internal redirect tests require that the `Location` header be set in the response, and the echo server `/status` endpoint does not support `X-Echo-Set-Header`.

ah yeah good point, I think we can probably try to get that change into the echo server as well in the near-term, can bother the maintainers of that repo to accept some PRs

internal/dag/dag.go
```diff
@@ -213,6 +213,40 @@ type Redirect struct {
 	Prefix string
 }
 
+const (
+	// HeaderMatchTypeExact matches a header value exactly.
```
skriss, line 217: Please update the godoc :)

internal/dag/dag.go
```diff
@@ -213,6 +213,40 @@ type Redirect struct {
...
+	InternalRedirectCrossSchemeNever = "never"
+
+	// HeaderMatchTypeContains matches a header value if it contains the
+	// provided value.
```
skriss, line 221: Please update the godoc :)

internal/dag/dag.go
```diff
@@ -213,6 +213,40 @@ type Redirect struct {
...
+	InternalRedirectCrossSchemeSafeOnly = "safeOnly"
+
+	// HeaderMatchTypePresent matches a header if it is present in a request.
```
skriss, line 224: Please update the godoc :)

skriss (changes requested):
Thanks @Jean-Daniel! Had a couple minor initial comments, but overall this looks good.

The other thing we'll want to do is add some documentation for this new feature. It could possibly go on the existing [Request Routing page](https://github.com/projectcontour/contour/blob/main/site/content/docs/main/config/request-routing.md), or could be a new page under https://github.com/projectcontour/contour/tree/main/site/content/docs/main/config with a new entry in the [nav menu](https://github.com/projectcontour/contour/blob/main/site/data/docs/main-toc.yml).

skriss:
There is also one minor whitespace-related merge conflict to resolve.

skriss:
And you'll also need to add a changelog file per https://github.com/projectcontour/contour/actions/runs/4086120214/jobs/7045008032

skriss:
Updates LGTM.

Still need to add some user documentation:
> The other thing we'll want to do is add some documentation for this new feature. It could possibly go on the existing [Request Routing page](https://github.com/projectcontour/contour/blob/main/site/content/docs/main/config/request-routing.md), or could be a new page under https://github.com/projectcontour/contour/tree/main/site/content/docs/main/config with a new entry in the [nav menu](https://github.com/projectcontour/contour/blob/main/site/data/docs/main-toc.yml).

skriss:
@Jean-Daniel gentle nudge on the request for documentation; it'd be great to get this merged soon!

Jean-Daniel:
Thanks for the nudge, I didn't had much time these days. 
I just wrote a simple paragraph in the routing page with pointer to the Envoy doc. Is it enough ?

skriss (approved):
> I just wrote a simple paragraph in the routing page with pointer to the Envoy doc. Is it enough ?

Looks good to me, thanks for the updates!

apis/projectcontour/v1/httpproxy.go
```diff
@@ -616,6 +620,33 @@ type HTTPRequestRedirectPolicy struct {
...
 
+type HTTPInternalRedirectPolicy struct {
+	// MaxInternalRedirects An internal redirect is not handled, unless the number of previous internal
+	// redirects that a downstream request has encountered is lower than this value.
+	// +optional
+	MaxInternalRedirects uint32 `json:"maxInternalRedirects,omitempty"`
+
+	// RedirectResponseCodes If unspecified, only 302 will be treated as internal redirect.
+	// Only 301, 302, 303, 307 and 308 are valid values.
+	// +optional
```
sunjayBhatia, line 631: I think we're missing validation (via kubebuilder or otherwise in the httpproxy processor) that valid status codes hasve been set here

apis/projectcontour/v1/httpproxy.go
```diff
@@ -616,6 +620,33 @@ type HTTPRequestRedirectPolicy struct {
...
+	RedirectResponseCodes []uint32 `json:"redirectResponseCodes,omitempty"`
+
+	// AllowCrossSchemeRedirect Allow internal redirect to follow a target URI with a different scheme
+	// than the value of x-forwarded-proto.
+	// SafeOnly allows same scheme redirect and safe cross scheme redirect, which means if the downstream
+	// scheme is HTTPS, both HTTPS and HTTP redirect targets are allowed, but if the downstream scheme
+	// is HTTP, only HTTP redirect targets are allowed.
+	// +kubebuilder:validation:Enum=Always;Never;SafeOnly
```
sunjayBhatia, line 639: cc @skriss I think we at some point stopped using Enums and defaults via kubebuilder and preferred to do validation in the various processors?

skriss: We definitely did for ContourConfiguration, but HTTPProxy uses it in a number of places, so I was fine with this.

sunjayBhatia:
other than the small kubebuilder defaulting/validation note, functionality looks good!

apis/projectcontour/v1/httpproxy.go
Jean-Daniel: According to their documentation, Envoy simply ignores other status code, so no validation is "required", but maybe we should warn if an other status code is used ?

skriss: We use kubebuilder validation for this elsewhere (https://github.com/projectcontour/contour/blob/main/apis/projectcontour/v1/httpproxy.go#L596) so I think that would work here as well.

Jean-Daniel: `type RetryOn string` is not exactly a type alias, but a new type. When creating the DAG representation, the `RetryOn` slice must be explicitly converted to a `string[]` by creating a new `slice` and populating it with `string(RetryOn)` values.

kubebuilder does not support type alias (`type RedirectResponseCode = uint32`), so I will have to handle the conversion explicitly too.

apis/projectcontour/v1/httpproxy.go
Jean-Daniel: Talking about validation, my understanding was that the annotation where use only to generate the rules in the CRDs, but it still requires a ValidatingWebHook to perform the validation. 

Where is the validation performed in Contour ?

sunjayBhatia (approved):
one nit on removing usage of `pointer` package, but can do in a follow-up

Issue #4820, opened 2022-10-25
Gateway API: Implement GRPCRoute


sunjayBhatia:
**Please describe the problem you have**
[GRPCRoute](https://gateway-api.sigs.k8s.io/api-types/grpcroute/) is an experimental channel resource in Gateway API for: `specifying routing behavior of gRPC requests from a Gateway listener to an API object, i.e. Service.`

Given we've had a lot of requests for gRPC-backend app features etc. it would be nice to support gRPC in a more first-class manner with this resource.

sunjayBhatia:
Proposal to move this into 1.24 or next release milestone, Gateway API upstream wants to graduate this resource and need implementations to start using it!

skriss:
cc @vmw-yingy @fangfpeng

skriss:
ref. https://projectcontour.io/guides/grpc/

PR #5011, opened 2023-01-26
GRPCRoute: Add GRPCRoute support

fangfpeng:
Add kind and cache support for GRPCRoute

Updates #4820

Signed-off-by: Fang Peng <fpeng@vmware.com>

### DIFF
--- a/internal/k8s/kind.go
+++ b/internal/k8s/kind.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes/scheme"
+	gatewayapi_v1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 	gatewayapi_v1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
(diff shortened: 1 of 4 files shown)


internal/k8s/kind.go
```diff
@@ -45,6 +46,8 @@ func KindOf(obj interface{}) string {
 			return "HTTPProxy"
 		case *gatewayapi_v1beta1.HTTPRoute:
 			return "HTTPRoute"
+		case *gatewayapi_v1alpha2.GRPCRoute:
+			return "GRPCRoute"
```
skriss, line 50: Will have to look to see exactly where this is used; I'm noticing we don't have TLSRoute here.

skriss (approved):
LGTM!

internal/k8s/kind.go
fangfpeng: Right, TLSRoute is not there. For routes, KindOf() is used in cache.go, statusaddress.go and metrics.go for some debug/error message or metric counter.

skriss: :+1: thanks, yeah makes sense to add GRPCRoute here, we should follow up with adding any other missing types for now, and/or see if we can implement this functionality differently such that we don't need to maintain a hardcoded list.

Issue #5007, opened 2023-01-25
add metric(s) for status update load


sunjayBhatia:
**Please describe the problem you have**
When debugging issues like #5001 it isn't clear at the moment via metrics how the status update components of Contour are working. This includes knowing the frequency/duration/etc. of status updates so we can find where bottlenecks might be. Updating status is often a resource intensive operation for the leader instance of Contour so the more information we can get, the better.

We should add some metrics that can help us get a picture of what is happening, some ideas:
- [x] status update count total (per resource kind)
- [x] status update count succeeded (per resource kind)
- [x] status update failure count (per resource kind)
- [x] summary/histogram of status update duration (segmented between failures and success?)
- [x] number of status updates per dag run
- [x] total duration of status updates per dag run

skriss:
Other ideas to look at:
- number of status updates (excluding no-ops) per DAG run
- total duration of status updates per DAG run

sunjayBhatia:
> number of status updates (excluding no-ops) per DAG run

I *think* when aggregating/visualizing this is a little tricky to do "per DAG run" given the nature? would we want to capture a distribution for this vs. what we get in https://github.com/projectcontour/contour/pull/5037 would be to look at the counters `status_update_total - status_update_noop_total` (I think we could correlate the latter with plotting `contour_dagrebuild_total` to figure out when DAG rebuilds happen? would have to experiment a bit)

sunjayBhatia:
> total duration of status updates per DAG run

I wonder if we can extrapolate this out indirectly from `contour_status_update_duration_seconds_sum` etc., or of it's just better to collect the total metric, will have to play around a bit

skriss:
> look at the counters status_update_total - status_update_noop_total (I think we could correlate the latter with plotting contour_dagrebuild_total to figure out when DAG rebuilds happen? would have to experiment a bit)

Yeah, that's fair, that would probably get us close enough anyway. 

> I wonder if we can extrapolate this out indirectly from contour_status_update_duration_seconds_sum etc., or of it's just better to collect the total metric, will have to play around a bit

Yeah, if we can find a way to derive it that's fine, I think this would just help indicate whether the queue is backing up or not -- individual status updates would still be ~quick, but if there are a large # in the queue, and we're only doing 5 per second, then updates will get backed up.

sunjayBhatia:
Doing the above in a follow-up PR ^

skriss:
Bumping to 1.27

[...]

PR #5012, opened 2023-01-26
internal/k8s: Updates to KindOf helper

sunjayBhatia:
Adds support for kinds that were missing

Relevant to #5011 and to work on #5007

For now a draft while we think about how to not have to manage this manually


skriss:
Need to add `Namespace` as well.

Let's get this merged, then can revisit if there's a refactor we can do here to avoid having to manually maintain this.

skriss:
Now ready for rebase

skriss (approved):

PR #5017, opened 2023-01-27
internal/dag: Loosen restrictions on duplicate include detection

sunjayBhatia:
Empty or lone prefix path matches on "/" are now not considered duplicates, due to complications with existing examples and possible wide range of users relying on the previous less restrictive behavior.

Updates release note on duplicate include detection to major and adds some more detail


internal/dag/httpproxy_processor.go
```diff
@@ -1392,8 +1392,12 @@ func toStringSlice(hvs []contour_api_v1.CORSHeaderValue) []string {
 
 func includeMatchConditionsIdentical(include contour_api_v1.Include, seenConds map[string][][]HeaderMatchCondition) bool {
 	pathPrefix := mergePathMatchConditions(include.Conditions).Prefix
-
 	includeHeaderConds := mergeHeaderMatchConditions(include.Conditions)
+
+	if pathPrefix == "/" && len(includeHeaderConds) == 0 {
```
sunjayBhatia, line 1397: the header condition check is covered by some existing tests, which was nice

skriss, line 1397: Worth a comment here for posterity

skriss:
Yep, this looks sensible for now and we'll revisit more broadly with 1.25.

skriss (approved):

PR #5021, opened 2023-01-28
internal/dag: Add GRPCRoute support in gatewayapi processor

fangfpeng:
Implement gatewayapi_processor.go for building GRPCRoute dag.

todos:
- add GRPCRoute into allowed route kinds for listener with HTTP/HTTPS protocol This needs quite some changes on status_test.go for wanted SupportedKinds for listeners with http/http protocols.
- add unit tests

Updates #4820

Signed-off-by: Fang Peng <fpeng@vmware.com>

### DIFF
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -38,6 +38,7 @@ import (
 const (
 	KindHTTPRoute = "HTTPRoute"
 	KindTLSRoute  = "TLSRoute"
+	KindGRPCRoute = "GRPCRoute"
 	KindGateway   = "Gateway"
 )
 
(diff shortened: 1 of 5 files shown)


internal/dag/gatewayapi_processor.go
```diff
@@ -1502,6 +1786,89 @@ func (p *GatewayAPIProcessor) clusterRoutes(routeNamespace string, matchConditio
...
+		// Keep track of all the weights for this set of backend refs. This will be
+		// used later to understand if all the weights are set to zero.
+		totalWeight += routeWeight
+
+		// https://github.com/projectcontour/contour/issues/3593
+		service.Weighted.Weight = routeWeight
+		clusters = append(clusters, &Cluster{
+			Upstream:              service,
+			Weight:                routeWeight,
+			Protocol:              service.Protocol,
```
sunjayBhatia, line 1860: We probably want to make a little change here:
- The protocol here comes from the referenced backend Service annotations if they are set (see: https://github.com/projectcontour/contour/blob/19a4c9f0a372838782d30ffa78c3c261cfaf6f34/internal/dag/accessors.go#L64)
- If the Service annotations are set, we should probably respect them, for backwards compatibility etc.
- If not set, currently we will configure a HTTP/1 cluster 
  - https://github.com/projectcontour/contour/blob/19a4c9f0a372838782d30ffa78c3c261cfaf6f34/internal/envoy/v3/cluster.go#L115
  - https://github.com/projectcontour/contour/blob/19a4c9f0a372838782d30ffa78c3c261cfaf6f34/internal/envoy/v3/cluster.go#L308

I think if the protocol is not set on the Service, we should maybe assume a protocol that matches what listener the route was attached to, e.g. `h2` for HTTPS listener and `h2c` for HTTP, given the user is explicitly asking for a gRPC route not a general HTTP route

internal/dag/gatewayapi_processor.go
```diff
@@ -1256,6 +1339,206 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1beta1.HTTPRou
...
+	}
+
+	// Type specifies how to match against the service and/or method.
+	// Support: Core (Exact with service and method specified)
+	// Not Support: Implementation-specific (Exact with method specified but no service specified)
+	// Not Support: Implementation-specific (RegularExpression)
+
+	// Support "Exact" match type only. If match type is not specified, use "Exact" as default.
+	if match.Type != nil && *match.Type != gatewayapi_v1alpha2.GRPCMethodMatchExact {
+		routeAccessor.AddCondition(status.ConditionValidMatches, metav1.ConditionFalse, status.ReasonMethodMatchType, "GRPCRoute.Spec.Rules.Matches.Method: Only Exact match type is supported.")
```
skriss, line 1498: I think we can use an upstream condition type and reason here: `gatewayapi_v1beta1.RouteConditionAccepted` and `gatewayapi_v1beta1.RouteReasonUnsupportedValue`. It's not explicitly spelled out in the API spec (yet) for GRPCRoute, but this is what's typically used for other route types in these scenarios.

internal/dag/gatewayapi_processor.go
```diff
@@ -1256,6 +1339,206 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1beta1.HTTPRou
...
+		return nil, false
+	}
+
+	if match.Service == nil || isBlank(*match.Service) || match.Method == nil || isBlank(*match.Method) {
+		routeAccessor.AddCondition(status.ConditionValidMatches, metav1.ConditionFalse, status.ReasonInvalidMethodMatch, "GRPCRoute.Spec.Rules.Matches.Method: Both Service and Method need be configured.")
```
skriss, line 1503: Can use gatewayapi_v1beta1.RouteConditionAccepted as the condition type but the reason is less obvious here, fine with the custom one for now.

internal/dag/gatewayapi_processor.go
```diff
@@ -1256,6 +1339,206 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1beta1.HTTPRou
...
+		for _, match := range rule.Matches {
+			// Convert method match to path match
+			pathMatch, ok := gatewayGRPCMethodMatchCondition(match.Method, routeAccessor)
+			if !ok {
+				continue
+			}
+
+			headerMatches, err := gatewayGRPCHeaderMatchConditions(match.Headers)
+			if err != nil {
+				routeAccessor.AddCondition(status.ConditionNotImplemented, metav1.ConditionTrue, status.ReasonHeaderMatchType, err.Error())
```
skriss, line 1371: Let's use `gatewayapi_v1beta1.RouteConditionAccepted` and `gatewayapi_v1beta1.RouteReasonUnsupportedValue` here as well.

skriss (changes requested):

internal/dag/gatewayapi_processor.go
```diff
@@ -1256,6 +1339,206 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1beta1.HTTPRou
...
+		return nil, false
+	}
+
+	if match.Service == nil || isBlank(*match.Service) || match.Method == nil || isBlank(*match.Method) {
+		routeAccessor.AddCondition(status.ConditionValidMatches, metav1.ConditionFalse, status.ReasonInvalidMethodMatch, "GRPCRoute.Spec.Rules.Matches.Method: Both Service and Method need be configured.")
+		return nil, false
+	}
+
+	// Convert service and method into path
+	path := *match.Service + "/" + *match.Method
```
vmw-yingy, line 1508: Shall we add "/" between service and method?
https://gateway-api.sigs.k8s.io/api-types/grpcroute/ 
gives an example of method:"The com.example.v2.User.Login method."

sunjayBhatia: yeah there should be a `/` between service and method, for that example in `com.example.v2.User`, `com.example.v2` is a package path,`User` is the service name, and `Login` is the RPC method

see https://projectcontour.io/docs/v1.24.0/guides/grpc/#httpproxy-configuration for an example as well matching on service `yages.Echo` and method `Ping`

skriss:
I think this is probably good to merge into the dev branch so we can work on adding more tests if you are ready @fangfpeng

internal/status/routeconditions.go
```diff
@@ -37,10 +37,12 @@ const (
 	ReasonHeaderMatchType               gatewayapi_v1beta1.RouteConditionReason = "HeaderMatchType"
 	ReasonQueryParamMatchType           gatewayapi_v1beta1.RouteConditionReason = "QueryParamMatchType"
 	ReasonHTTPRouteFilterType           gatewayapi_v1beta1.RouteConditionReason = "HTTPRouteFilterType"
+	ReasonMethodMatchType               gatewayapi_v1beta1.RouteConditionReason = "MethodMatchType"
```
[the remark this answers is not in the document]
fangfpeng: will take care of this in a next CL.

fangfpeng:
> I think this is probably good to merge into the dev branch so we can work on adding more tests if you are ready @fangfpeng

@skriss Yes, I am ready for the merge. We will add more test cases in next CLs.

PR #5023, opened 2023-01-29
Automatically set GOMAXPROCS to match the number of CPUs available to the container

### LINKED ISSUE #5022: Set GOMAXPROCS to match the number of CPUs actually available
From https://pkg.go.dev/runtime: "The GOMAXPROCS variable limits the number of operating system threads that can execute user-level Go code simultaneously.".

By default GOMAXPROCS is set to the number of CPUs visible to the process but when running in containers this is the number of CPUs of the node which is sometimes much higher than the number of CPUs available to processes running in the container.  On bare metal nodes for example there is a significant difference between the two.

Setting GOMAXPROCS higher than the number of CPUs available can hurt performance - for example see https://blog.devgenius.io/know-gomaxprocs-before-deploying-your-go-app-to-kubernetes-7a458fb63af1 and https://github.com/uber-go/automaxprocs.

While it is possible to manually set the GOMAXPROCS environment variable in the contour container to match resources.request.cpu it would be much better if this value was automatically set to match the number of CPUs available to the container like automaxprocs does.

erwbgy:
Use go.uber.org/automaxprocs to automatically set GOMAXPROCS to match the number of CPUs available to the container.

Fixes #5022.

### DIFF
--- a/cmd/contour/contour.go
+++ b/cmd/contour/contour.go
@@ -16,6 +16,9 @@ package main
 import (
 	"os"
 
+	// Automatically set GOMAXPROCS to match the number of CPUs available to the container
+	_ "go.uber.org/automaxprocs"
+
 	resource_v3 "github.com/envoyproxy/go-control-plane/pkg/resource/v3"
 	"github.com/projectcontour/contour/internal/build"
 	"github.com/projectcontour/contour/internal/envoy"
--- a/changelogs/unreleased/5023-erwbgy-small.md
+++ b/changelogs/unreleased/5023-erwbgy-small.md
@@ -1,0 +1,1 @@
+Contour now sets GOMAXPROCS to match the number of CPUs available to the container which results in lower and more stable CPU usage under high loads and where the container and node CPU counts differ significantly.
--- a/go.mod
+++ b/go.mod
@@ -25,6 +25,7 @@ require (
 	github.com/stretchr/testify v1.8.1
 	github.com/tsaarni/certyaml v0.9.1
 	github.com/vektra/mockery/v2 v2.16.0
+	go.uber.org/automaxprocs v1.5.1
 	golang.org/x/oauth2 v0.4.0
 	gonum.org/v1/plot v0.12.0
 	google.golang.org/genproto v0.0.0-20230117162540-28d6b9783ac4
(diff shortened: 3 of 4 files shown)


sunjayBhatia (commented):
doesn't look like there is a way to override the auto setting, in case this isn't something you want to apply in your specific cluster for whatever reason (you may already have a manual GOMAXPROCS customized setting that works, or this new feature is causing issues so you want to revert to the old behavior)

if we do change the default behavior, we should call it out in at least a minor change maybe even major

sunjayBhatia:
> doesn't look like there is a way to override the auto setting, in case this isn't something you want to apply in your specific cluster for whatever reason (you may already have a manual GOMAXPROCS customized setting that works, or this new feature is causing issues so you want to revert to the old behavior)
> 
> if we do change the default behavior, we should call it out in at least a minor change maybe even major

ah nevermind, it's not documented in the package but you can override by setting GOMAXPROCS yourself: https://github.com/uber-go/automaxprocs/blob/58f5aee66e9fc9138dd8ff4f966f101b51b7db63/maxprocs/maxprocs.go#L101

so in that case, we should call that out in our release notes that you can revert to original behavior/override the auto setting in case you need to

cmd/contour/contour.go
```diff
@@ -16,6 +16,9 @@ package main
 import (
 	"os"
 
+	// Automatically set GOMAXPROCS to match the number of CPUs available to the container
+	_ "go.uber.org/automaxprocs"
```
sunjayBhatia, line 20: i think maybe instead of implementing it this way with a side effect import, let's be more explicit about using the package so we don't accidentally undo the behavior we introduce in the future etc.

maybe see if we can write our own init function or something that hooks up our logger to the `maxprocs .Set` function

erwbgy: That makes sense @sunjayBhatia. I'll rework it to use `maxprocs.Set` so it is explicit and we get logging.

rajatvig: The uber library is relatively well known as intentionally been created to set `GOMAXPROCS` with as little fuss as possible. 

Would recommend reusing the code as is vs writing a custom implementation.

sunjayBhatia: the suggestion here isn't to rewrite the library, but to directly use the helper function rather than the side-effect import so we have more visibility and don't accidentally undo any changes here, get logging from the library, etc.

rajatvig: Ok if I add it in?

rajatvig: Created https://github.com/projectcontour/contour/pull/5211 as this PR ran into quite a few conflicts due to age.

skriss:
Closing in favor of #5211

PR #5032, opened 2023-01-30
drop unneeded toolchain installations

skriss:
Drops kubebuilder and kustomize toolchain
installations since they're not used. Also
switches to using go run for the ginkgo CLI
to avoid having a hardcoded version outside
of go.mod.


sunjayBhatia (approved):
good call, yeah the manual ginkgo bumps were getting old!

PR #5036, opened 2023-01-31
HTTPProxy: Implement HTTP query param matching

### LINKED ISSUE #5033: Support query parameter route matching in HTTPProxy
Supporting query parameter route matching would enable users to make use of an existing Envoy capability that is already present for the Gateway API HTTPRoute resource.

relu:
Closes #5033


changelogs/unreleased/5036-relu-minor.md
```diff
@@ -0,0 +1,66 @@
+## HTTPProxy: implement HTTP query parameter matching
```
skriss, line 1: This is a fantastic changelog, thank you!

internal/dag/dag.go
```diff
@@ -156,20 +156,45 @@ func (hc *HeaderMatchCondition) String() string {
...
+	// given regular expression.
+	QueryParamMatchTypeRegex = "regex"
+
+	// QueryParamMatchTypeContains matches a querystring parameter value
+	// contains the given string.
+	QueryParamMatchTypeContains = "contains"
+
+	// QueryParamMatchTypeIgnoreCase enables case insensitive matching of the
+	// querystring parameter value.
+	QueryParamMatchTypeIgnoreCase = "ignorecase"
```
skriss, line 178: I think we can drop this one since it has its own field in the QueryParamMatchCondition, correct me if I'm missing something though

internal/featuretests/v3/envoy.go
```diff
@@ -110,7 +110,7 @@ func routeCluster(cluster string, opts ...func(*envoy_route_v3.Route_Route)) *en
 	return r
 }
 
-func routePrefix(prefix string, headers ...dag.HeaderMatchCondition) *envoy_route_v3.RouteMatch {
+func routePrefixWithHeaderConditions(prefix string, headers ...dag.HeaderMatchCondition) *envoy_route_v3.RouteMatch {
```
skriss, line 113: It looks like most uses of this function don't actually use any header conditions, apart from in `headercondition_test.go` -- for readability I'm thinking it'd be nice to keep a func named `routePrefix` that _only_ takes a prefix, and use that where appropriate, then have`routePrefixWithHeaderConditions` and `routePrefixWithQueryParameterConditions` separately to be used where needed.

skriss (changes requested):
Thanks for the PR @relu, this is great work. I'm still making my way through a few things but here are my comments so far. In addition, please add some documentation to https://github.com/projectcontour/contour/blob/main/site/content/docs/main/config/request-routing.md#conditions covering this new functionality. Thanks again!

test/e2e/httpproxy/query_parameter_condition_match_test.go
```diff
@@ -0,0 +1,349 @@
...
+		}
+		f.CreateHTTPProxyAndWaitFor(p, e2e.HTTPProxyValid)
+
+		type scenario struct {
+			queryParams    map[string]string
+			expectResponse int
+			expectService  string
+		}
+
+		cases := []scenario{
```
sunjayBhatia, line 229: similar to the gw api testing, would be good to have coverage to demonstrate the case where a request has multiple query params with the same key so we have it defined how contour will behave, e.g. https://github.com/projectcontour/contour/blob/052a12d88decdb068926333b9f90ae77cb650a5c/test/e2e/gateway/query_param_match_test.go#L69-L73

internal/dag/httpproxy_processor.go
```diff
@@ -1390,9 +1399,10 @@ func toStringSlice(hvs []contour_api_v1.CORSHeaderValue) []string {
 	return s
 }
 
-func includeMatchConditionsIdentical(include contour_api_v1.Include, seenConds map[string][][]HeaderMatchCondition) bool {
+func includeMatchConditionsIdentical(include contour_api_v1.Include, seenHeaderConds map[string][][]HeaderMatchCondition, seenQueryParamConds map[string][][]QueryParamMatchCondition) bool {
```
sunjayBhatia, line 1402: it would be great to have some coverage over the logic here similar to this test case and a few below it: https://github.com/projectcontour/contour/blob/052a12d88decdb068926333b9f90ae77cb650a5c/internal/dag/status_test.go#L1731

changelogs/unreleased/5036-relu-minor.md
sunjayBhatia: +1 from me

internal/dag/dag.go
relu: Yep, seems I forgot to remove this from an earlier stage.

internal/dag/httpproxy_processor.go
relu: Good catch, totally missed this one!

skriss (approved):
LGTM, thanks @relu!

skriss:
@sunjayBhatia leaving this for you since you had a few comments from previous review.

sunjayBhatia (approved):
LGTM, thanks @relu 

I will follow up with some more unit test coverage to add to this

PR #5041, opened 2023-02-01
Fix the upgrade e2e test flakiness

relu:
The upgrade e2e test can fail in some situations because the envoy waiter is incorrectly returning true before the StatefulSet even manages to finish the rollout since it only checks the StatefulSet state on properties that are inconsistently populated due to timing differences.

This fix addressed that by using a similar condition validation method as the one already used for the Contour Deployment rollout.

### DIFF
--- a/test/e2e/deployment.go
+++ b/test/e2e/deployment.go
@@ -44,6 +44,7 @@ import (
 	api_errors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	apimachinery_util_yaml "k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/client-go/kubernetes"
@@ -328,31 +329,9 @@ func (d *Deployment) WaitForContourDeploymentUpdated() error {
 	if len(d.ContourDeployment.Spec.Template.Spec.Containers) != 1 {
 		return errors.New("invalid Contour Deployment containers spec")
 	}
-	contourPodImage := d.ContourDeployment.Spec.Template.Spec.Containers[0].Image
+	labelSelectAppContour := labels.SelectorFromSet(d.ContourDeployment.Spec.Selector.MatchLabels)
 	updatedPods := func() (bool, error) {
-		pods := new(v1.PodList)
-		labelSelectAppContour := &client.ListOptions{
-			LabelSelector: labels.SelectorFromSet(d.ContourDeployment.Spec.Selector.MatchLabels),
-			Namespace:     d.ContourDeployment.Namespace,
-		}
-		if err := d.client.List(context.TODO(), pods, labelSelectAppContour); err != nil {
-			return false, err
-		}
-
-		updatedPods := 0
-		for _, pod := range pods.Items {
-			if len(pod.Spec.Containers) != 1 {
-				return false, errors.New("invalid Contour Deployment pod containers")
-			}
-			if pod.Spec.Containers[0].Image != contourPodImage {
-				continue
-			}
-			for _, cond := range pod.Status.Conditions {
-				if cond.Type == v1.PodReady && cond.Status == v1.ConditionTrue {
-					updatedPods++
-				}
-			}
-		}
+		updatedPods := d.getUpdatedPods(labelSelectAppContour, d.EnvoyDaemonSet.Namespace)
 		return updatedPods == int(*d.ContourDeployment.Spec.Replicas), nil
 	}
 	return wait.PollImmediate(time.Millisecond*50, time.Minute, updatedPods)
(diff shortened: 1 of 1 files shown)


test/e2e/deployment.go
```diff
@@ -374,17 +374,31 @@ func (d *Deployment) WaitForEnvoyUpdated() error {
...
+				}
+			}
+		}
+		return updatedPods == int(*&d.EnvoyDaemonSet.Status.DesiredNumberScheduled), nil
 	}
-	return wait.PollImmediate(time.Millisecond*50, time.Minute*3, daemonSetUpdated)
+	return wait.PollImmediate(time.Millisecond*50, time.Minute*3, updatedPods)
 }
 
 func (d *Deployment) waitForEnvoyDeploymentUpdated() error {
```
skriss, line 404: I suppose it would make sense to use similar pod-level logic within this method too, for when we're testing envoy as a Deployment.

skriss (commented):
Thanks for this @relu!

test/e2e/deployment.go
```diff
@@ -374,17 +374,31 @@ func (d *Deployment) WaitForEnvoyUpdated() error {
...
 			return false, err
 		}
-		return tempDS.Status.ObservedGeneration == tempDS.Generation &&
-			tempDS.Status.NumberAvailable > 0 &&
-			tempDS.Status.NumberAvailable == tempDS.Status.DesiredNumberScheduled &&
-			tempDS.Status.UpdatedNumberScheduled == tempDS.Status.DesiredNumberScheduled, nil
+
+		updatedPods := 0
+		for _, pod := range pods.Items {
+			if pod.Spec.InitContainers[0].Image != contourPodImage {
```
sunjayBhatia, line 390: yeah this makes sense as the indicator to ensure the pods actually have been updated with the new contour

test/e2e/deployment.go
```diff
@@ -374,17 +374,31 @@ func (d *Deployment) WaitForEnvoyUpdated() error {
 }
 
 func (d *Deployment) waitForEnvoyDaemonSetUpdated() error {
-	daemonSetUpdated := func() (bool, error) {
-		tempDS := new(apps_v1.DaemonSet)
-		if err := d.client.Get(context.TODO(), client.ObjectKeyFromObject(d.EnvoyDaemonSet), tempDS); err != nil {
+	contourPodImage := d.ContourDeployment.Spec.Template.Spec.Containers[0].Image
+	updatedPods := func() (bool, error) {
+		pods := new(v1.PodList)
+		labelSelectAppContour := &client.ListOptions{
```
sunjayBhatia, line 380: nit, naming on this change to Envoy specific

test/e2e/deployment.go
sunjayBhatia: ah, looks like we're getting a panic here since we don't have an init container in many of our e2e test suites: https://github.com/projectcontour/contour/actions/runs/4067086318/jobs/7004630297#step:7:160

see: https://github.com/projectcontour/contour/blob/052a12d88decdb068926333b9f90ae77cb650a5c/test/e2e/deployment.go#L557 (we remove the init container in that case since we just run `contour bootstrap` locally)

test/e2e/deployment.go
```diff
@@ -374,28 +353,62 @@ func (d *Deployment) WaitForEnvoyUpdated() error {
...
 
 func (d *Deployment) waitForEnvoyDeploymentUpdated() error {
-	deploymentUpdated := func() (bool, error) {
-		tempDeploy := new(apps_v1.Deployment)
-		if err := d.client.Get(context.TODO(), client.ObjectKeyFromObject(d.EnvoyDeployment), tempDeploy); err != nil {
-			return false, err
+	labelSelectAppEnvoy := labels.SelectorFromSet(d.EnvoyDeployment.Spec.Selector.MatchLabels)
+	updatedPods := func() (bool, error) {
+		updatedPods := d.getUpdatedPods(labelSelectAppEnvoy, d.EnvoyDeployment.Namespace)
+		return updatedPods == int(*d.EnvoyDeployment.Spec.Replicas) &&
```
skriss, line 376: I'm not seeing exactly what the linter is complaining about here; `// nolint(gocritic)` is always an option if this is a false positive.

test/e2e/deployment.go
```diff
@@ -374,28 +353,62 @@ func (d *Deployment) WaitForEnvoyUpdated() error {
...
+			updatedPods == int(d.EnvoyDeployment.Status.ReadyReplicas) &&
+			int(d.EnvoyDeployment.Status.UnavailableReplicas) == 0, nil
+	}
+	return wait.PollImmediate(time.Millisecond*50, time.Minute*3, updatedPods)
+}
+
+func (d *Deployment) getUpdatedPods(labelSelector labels.Selector, namespace string) int {
+	contourPodImage := d.ContourDeployment.Spec.Template.Spec.Containers[0].Image
```
skriss, line 384: Looks like we'll need to pass the desired image in as an arg, since it'll be the Envoy image when used within `WaitForEnvoy...` method

test/e2e/deployment.go
relu: Maybe my last push will do it, changed the order of the operands.

test/e2e/deployment.go
relu: Not quite, it's supposed to check the contour image (which exists as a sidecar in envoy as well) because that's the main image that changes after the upgrade.

skriss: Ahh, I see. That works!

skriss (approved):
LGTM, will leave for @sunjayBhatia. Thanks @relu!

sunjayBhatia (approved):

PR #5042, opened 2023-02-02
HCM: add support for 'always_set_request_id_in_response'

### LINKED ISSUE #2911: Expose X-Request-Id header in downstream responses
This is the issue described in envoyproxy/envoy#5828. The simple solution is to set `always_set_request_id_in_response: true` on the `HttpConnectionManager` filter, but I think this is only available in xDS API v3 (see #1898).

This may not be desirable in all installations of Contour and could be configurable at various levels:

- globally, via Contour config
- locally, via `HTTPProxy`

Conceptually, I'd say in belongs in the same API vicinity as `HeadersPolicy`.

izturn:

add support for `always_set_request_id_in_response`
fix #2911


izturn:
ping @skriss @sunjayBhatia

sunjayBhatia (commented):
so you can sort of get this behavior today by modifying the global response header policy like so:

```yaml
    policy:
      # default headers to set on all responses (unless set/removed on the HTTPProxy object itself)
      response-headers:
        set:
          X-Request-Id: "%REQ(X-Request-Id)%"
```

this global response header policy is overridable in an individual HTTPProxy and does not currently get applied to any direct responses (e.g. the 502 we program if part of a HTTPProxy route is invalid)

it would be good to validate if a particular HTTPProxy responseHeadersPolicy can override the "AlwaysSetRequestIDInResponse" setting and document that behavior

Also as a follow up or additional issue we could consider if global response headers can/should be set on direct responses we program. We could actually implement this "instead" of this feature to hopefully provide the same functionality w/o adding a new config knob.

izturn:
got it

PR #5043, opened 2023-02-02
Add tracing support

### LINKED ISSUE #399: Tracing Configuration

yangyy93:
Add support for exporting tracing data to opentelemetry
close #399 

Signed-off-by: yy [yang.yang@daocloud.io]


apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -293,6 +293,8 @@ type EnvoyConfig struct {
 	// Network holds various configurable Envoy network values.
 	// +optional
 	Network *NetworkParameters `json:"network,omitempty"`
+
+	Tracing *TracingConfig `json:"tracing,omitempty"`
```
skriss, line 297: I think it probably makes sense for this to be a top-level field on the spec, i.e. as a sibling of `RateLimitService`.

yangyy93: ok

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -654,6 +657,53 @@ type RateLimitServiceConfig struct {
 	EnableResourceExhaustedCode *bool `json:"enableResourceExhaustedCode,omitempty"`
 }
 
+// TracingConfig defines properties for exporting trace data to OpenTelemetry.
+type TracingConfig struct {
+	// IncludePodDetail defines a flag.
+	// If it is true, contour will add the pod name and namespace to the span of the trace.
+	// the default is true.
+	// +optional
+	IncludePodDetail *bool `json:"IncludePodDetail,omitempty"`
```
pavelnikolov, line 666: Is this field name starting with a capital letter intentionally? Let's use lowercase to make it consistent with the rest. If you decide to leave it uppercase first letter, then the below would be equivalent:
```go
	IncludePodDetail *bool `json:",omitempty"`
```

yangyy93: This is a mistake, it should start with a lowercase letter

yangyy93:
@skriss any suggestions?

pavelnikolov:
What is remaining to be done before this PR can be merged?

pavelnikolov:
@yangyy93 it looks like a changelog file is missing:
```
Run go run ./hack/actions/check-changefile-exists.go
go: downloading github.com/google/go-github/v48 v48.2.0
go: downloading github.com/sirupsen/logrus v1.[9](https://github.com/projectcontour/contour/actions/runs/4382443387/jobs/7671496430#step:5:10).0
go: downloading golang.org/x/oauth2 v0.6.0
go: downloading golang.org/x/sys v0.6.0
go: downloading github.com/google/go-querystring v1.1.0
go: downloading golang.org/x/crypto v0.6.0
FATA[0000] 
Thanks for your PR.
For a PR to be accepted to Contour, it must have:
- at least one release-note label set
- a file named changelogs/unreleased/PR#-author-category,
  where category matches the release-note/category label you apply.

Error: Missing changelog file at ./changelogs/unreleased/5043-yangyy93-major.md
	
Please see the "Commit message and PR guidelines" section of CONTRIBUTING.md,
or https://github.com/projectcontour/contour/blob/main/design/changelog.md for background. 
exit status 1
```

changelogs/unreleased/5043-yangyy93-minor.md
```diff
@@ -0,0 +1,14 @@
+## Add Tracing Support
```
pavelnikolov, line 1: This file needs to be renamed to
```
./changelogs/unreleased/5043-yangyy93-major.md
```

sidharthramesh:
OMG is this happening!? Contour is a wonderful except for the tracing support, it’s got everything I need. Thank you for this.

changelogs/unreleased/5043-yangyy93-minor.md
yangyy93: Yes, this feature should belong to the major, I will update the changelog and add the corresponding documentation

cmd/contour/serve.go
```diff
@@ -599,6 +605,75 @@ func (s *Server) doServe() error {
...
+	}
+
+	var sni string
+	if extensionSvc.Spec.UpstreamValidation != nil {
+		sni = extensionSvc.Spec.UpstreamValidation.SubjectName
+	}
+
+	var customTags []*dag.CustomTag
+
+	if pointer.BoolDeref(tracingConfig.IncludePodDetail, true) {
```
skriss, line 642: We have a generic internal package for this -- see `ref.Val`. Please use that instead.

cmd/contour/servecontext_test.go
```diff
@@ -699,6 +699,50 @@ func TestConvertServeContext(t *testing.T) {
...
+							RequestHeaderName: ":method",
+						},
+					},
+					ExtensionService: "otel/otel-collector",
+				}
+				return ctx
+			},
+			getContourConfiguration: func(cfg contour_api_v1alpha1.ContourConfigurationSpec) contour_api_v1alpha1.ContourConfigurationSpec {
+				cfg.Tracing = &contour_api_v1alpha1.TracingConfig{
+					ServiceName:      pointer.String("contour"),
```
skriss, line 725:
```suggestion
					ServiceName:      ref.To("contour"),
```

pkg/config/parameters.go
```diff
@@ -557,6 +557,51 @@ type Parameters struct {
...
+
+	// OverallSampling defines maximum length of the request path
+	// to extract and include in the HttpUrl tag.
+	// the default value is 256
+	MaxPathTagLength uint32 `yaml:"maxPathTagLength,omitempty"`
+
+	// CustomTags defines a list of custom tags with unique tag name.
+	CustomTags []CustomTag `yaml:"customTags,omitempty"`
+
+	// ExtensionService identifies the extension service defining the otle-collector,
```
skriss, line 588:
```suggestion
	// ExtensionService identifies the extension service defining the otel-collector,
```

internal/dag/dag.go
```diff
@@ -701,6 +701,44 @@ type SecureVirtualHost struct {
 
 	// JWTProviders specify how to verify JWTs.
 	JWTProviders []JWTProvider
+
+	// Tracing defines custom tags and the service to send trace data
+	Tracing TracingConfig
```
skriss, line 706: I don't think this is being used anywhere, is it? Since tracing is a global setting, it's being directly configured on the Listener via the Listener xdscache, and doesn't need to be represented in the DAG. If that's accurate, then we can revert the additions to internal/dag.

skriss (changes requested):
@yangyy93 just starting to go through this, had a few initial comments.

We will also need to add documentation for this great new feature, as a new page under `site/content/docs/main/config`. Please see other pages for an idea of what we're looking for.

internal/dag/dag.go
```diff
@@ -701,6 +701,44 @@ type SecureVirtualHost struct {
...
+	OverallSampling float64
+
+	MaxPathTagLength uint32
+
+	CustomTags []*CustomTag
+}
+
+type CustomTag struct {
+	// TagName is the unique name of the custom tag.
+	TagName string `json:"tagName"`
```
skriss, line 727: no need for json tags on this struct

cmd/contour/serve.go
```diff
@@ -599,6 +605,75 @@ func (s *Server) doServe() error {
...
+		customTags = append(customTags, &dag.CustomTag{
+			TagName:         "podName",
+			EnvironmentName: "HOSTNAME",
+		}, &dag.CustomTag{
+			TagName:         "podNamespaceName",
```
skriss, line 647: My preference for the tag name would be:
```suggestion
			TagName:         "podNamespace",
```

skriss:
FYI here's a very basic config that folks can use for testing:

Install otel operator:
```
kubectl apply -f https://github.com/open-telemetry/opentelemetry-operator/releases/latest/download/opentelemetry-operator.yaml
```

Install an otel collector instance, with verbose logging exporter enabled:
```yaml
apiVersion: opentelemetry.io/v1alpha1
kind: OpenTelemetryCollector
metadata:
  name: simplest
spec:
  config: |
    receivers:
      otlp:
        protocols:
          grpc:
          http:
    processors:

    exporters:
      logging:
        verbosity: detailed

    service:
      pipelines:
        traces:
          receivers: [otlp]
          processors: []
          exporters: [logging]
```

Define extension service:
```yaml
kind: ExtensionService
metadata:
  name: otel-collector
spec:
  protocol: h2c
  services:
    - name: simplest-collector
      port: 4317
```

Update Contour config (needs a Contour deployment restart afterwards):
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: contour
  namespace: projectcontour
data:
  contour.yaml: |
    tracing:
      extensionService: default/otel-collector
      serviceName: some-service-name
      customTags:
      - tagName: custom-tag
        literal: foo
      - tagName: header-tag
        requestHeaderName: X-Custom-Header
```

Now you should be able to see traces in the logs of the otel collector, e.g.:
`kubectl logs deploy/simplest-collector`

@yangyy93, it would be nice to include something similar to the above in `examples`, which could also be referenced from the documentation.

cmd/contour/serve.go
yangyy93: ok

skriss:
@yangyy93 this will need a rebase to resolve merge conflicts with recent merges, thanks.

yangyy93:
@skriss Thanks for your review, I will modify the code and add documentation according to your comments

changelogs/unreleased/5043-yangyy93-minor.md
skriss: This still needs to be changed to major

cmd/contour/serve.go
```diff
@@ -24,6 +24,8 @@ import (
 	"time"
 
 	"github.com/alecthomas/kingpin/v2"
+	"k8s.io/utils/pointer"
```
skriss, line 27: Can remove this import now to address the lint error

skriss (changes requested):

pavelnikolov:
Is this PR ready to be merged?

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -659,6 +662,53 @@ type RateLimitServiceConfig struct {
...
+	// contour's default is 256.
+	// +optional
+	MaxPathTagLength *uint32 `json:"maxPathTagLength,omitempty"`
+
+	// CustomTags defines a list of custom tags with unique tag name.
+	// +optional
+	CustomTags []*CustomTag `json:"customTags,omitempty"`
+
+	// ExtensionService identifies the extension service defining the otel-collector.
+	ExtensionService NamespacedName `json:"extensionService,omitempty"`
```
skriss, line 693: This one's required, can drop the `omitempty` tag.

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -659,6 +662,53 @@ type RateLimitServiceConfig struct {
...
+}
+
+// CustomTag defines custom tags with unique tag name
+// to create tags for the active span.
+type CustomTag struct {
+	// TagName is the unique name of the custom tag.
+	TagName string `json:"tagName,omitempty"`
```
skriss, line 700: This field is required, can drop `omitempty`

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -659,6 +662,53 @@ type RateLimitServiceConfig struct {
 	EnableResourceExhaustedCode *bool `json:"enableResourceExhaustedCode,omitempty"`
 }
 
+// TracingConfig defines properties for exporting trace data to OpenTelemetry.
+type TracingConfig struct {
```
skriss, line 666: Please add some validation logic for this struct to `contourconfig_helpers.go` in this same package. In particular, things I see that should be validated:
- OverallSampling must be convertible to a float
- each CustomTag must have exactly one of Literal or RequestHeaderName

(This can look similar to the validation you have in pkg/config).

site/content/docs/main/config/tracing.md
```diff
@@ -0,0 +1,117 @@
...
+- [Overview](#overview)
+- [Tracing-config](#tracing-config)
+
+## Overview
+
+Envoy has rich support for [distributed tracing][1]，and supports exporting data to third-party providers (Zipkin, Jaeger, Datadog, etc.)
+
+[OpenTelemetry][2] is a CNCF project which is working to become a standard in the space. It was formed as a merger of the OpenTracing and OpenCensus projects.
+
+Contour supports configuring envoy to export data to open, and allows users to customize some configurations.
```
skriss, line 12:
```suggestion
Contour supports configuring envoy to export data to OpenTelemetry, and allows users to customize some configurations.
```

apis/projectcontour/v1alpha1/contourconfig_helpers.go
```diff
@@ -48,6 +52,51 @@ func (c *ContourConfigurationSpec) Validate() error {
 	return nil
 }
 
+func (t *TracingConfig) Validate() error {
+	if t.OverallSampling == nil && t.MaxPathTagLength == nil && t.ExtensionService == nil && t.CustomTags == nil {
+		return nil
+	}
```
skriss, line 58: Is there any specific reason to return nil for this case? I think that if TracingConfig is non-nil, then it should require at least an ExtensionService to be specified. I'd just drop these lines, and adjust the associated unit test case

pkg/config/parameters.go
```diff
@@ -689,6 +737,47 @@ func (p *MetricsParameters) Validate() error {
 	return nil
 }
 
+func (t *Tracing) Validate() error {
+	if t == nil {
+		return nil
+	}
+	if t.OverallSampling == 0 && t.MaxPathTagLength == 0 && t.ExtensionService == "" && t.CustomTags == nil {
+		return nil
+	}
```
skriss, line 746: Same comment here, I think if Tracing is non-nil, then we should require ExtensionService, so just drop these lines and let the rest of the logic catch errors.

skriss: I see that we're somewhat inconsistent in our existing code here, but I think it makes the most sense to use a `*Tracing` field that can be nil on the top-level Parameters struct, and not initialize it in the `Defaults` function.

skriss:
Just doing a little more testing on the latest version of code, but I think this is looking pretty good, thanks for the work on it @yangyy93!

internal/envoy/v3/tracing.go
```diff
@@ -0,0 +1,112 @@
...
+		if traceCustomTag := customTag(tag); traceCustomTag != nil {
+			customTags = append(customTags, traceCustomTag)
+		}
+	}
+
+	return &http.HttpConnectionManager_Tracing{
+		OverallSampling: &envoy_type.Percent{
+			Value: tracing.OverallSampling,
+		},
+		MaxPathTagLength: wrapperspb.UInt32(tracing.MaxPathTagLength),
```
skriss, line 46: When I'm testing with a Contour config file, and I don't specify any `maxPathTagLength`, it's getting passed to Envoy as an explicit `0` which has the effect of always stripping the path from the `http.url` tag. Looks like this is becaue the field on Parameters is not a pointer, so the code can't distinguish between "not provided" and "explicitly set to 0". So, we should change that field to a pointer to be able to differentiate the two cases, and then _only_ pass a value to Envoy if the user explicitly specified a value. If they didn't specify a value, we don't need to explicitly default to 256, we can just not pass the field to Envoy and let its own default apply.

yangyy93: Yes, this problem does occur, thanks for your review

skriss (approved):
Thanks for the updates @yangyy93, testing all checks out and this LGTM! Will leave for one more maintainer to approve.

sidharthramesh:
Yay!

RmStorm:
Is this RC release coming up sometime soon? I'm looking forward to trying it!

skriss:
> Is this RC release coming up sometime soon? I'm looking forward to trying it!

@RmStorm yes; we're a little behind schedule due to KubeCon but I hope to get it out today or tomorrow. Keep an eye out!

sunjayBhatia:
@RmStorm Hey there, thanks for testing out the RC! Looks like you're using the Gateway Provisioner to create Contour instances? If so would be good to confirm you've got all the CRDs and versions of images correct as a sanity check

Also, if you're modifying the ContourDeployment pointed to by a GatewayClass ParametersRef after you've already created a Gateway, you won't see the Contour installation for that Gateway get changed (in case that is what is going on here) you'll have to create a fresh Gateway after the change (due to the fact that the GatewayClass parameters are supposed to be a "snapshot" of what config to use and changing it according to the API guidelines should not cause a re-reconcile of any generated resources, this is a rough edge in Gateway API).

RmStorm:
Thanks that is very helpful. Manually updating the `contourConfig` and then rolling over the contour pods got the tracing config into envoy. :tada: Very cool.

Now I ran head first into the next problem. I'm getting the traces from Envoy in my trace backend but none of the underlying services are picking up on them. I have inspected some requests and it is because the `traceparent` header is not set. Is Envoy supposed to set trace headers? Or are my downstream services supposed to pay attention to the value of `x-request-id` and consider that the traceparent?

If that's the case I have a bit of a problem because one of the services is [opa](https://github.com/open-policy-agent/opa) and they rely on the [traceparent header](https://github.com/open-policy-agent/opa/blob/main/runtime/runtime.go#L369) being set.

I guess concretely my question is, can I get Envoy to set headers for tracepropagation or must I use `x-request-id`? and if `x-request-id` is the way to go, what is the common way of going about it.

Issue #4960, opened 2023-01-05
Support an arbitrary number of Listeners per Gateway


Rycieos:
This is an enhancement request, but it could also be considered a bug, as the Kubernetes Gateway API requires this support.

### Description
A single Gateway object needs to be able to support an arbitrary number of Listeners, including multiple Listeners of the same type.

### User story
I have an application that speaks both HTTP as well as a nonstandard protocol over TCP to the backend servers. I also want all traffic wrapped in TLS. To greatly simplify environment creation, I want all HTTPS and TCP traffic to be handled on the same FQDN, meaning the same IP address, meaning the same Gateway object. Since the HTTPS and TLS wrapped TCP traffic do not differ in FQDN, it is not possible to have them on the same port.

This can be specified with this example Gateway Spec:
```yaml
apiVersion: gateway.networking.k8s.io/v1beta1
kind: Gateway
spec:
  listeners:
  - allowedRoutes:
      kinds:
      - group: gateway.networking.k8s.io
        kind: HTTPRoute
    name: https
    port: 443
    protocol: HTTPS
  - allowedRoutes:
      kinds:
      - group: gateway.networking.k8s.io
        kind: TLSRoute
    name: tls
    port: 5000
    protocol: TLS
```

While currently I do not need multiple listeners of the same type on different ports, it should also be possible to do that as well.

### Issue
Currently, if this Gateway is created, Contour returns an error with the message:
https://github.com/projectcontour/contour/blob/d7d401280417e9d0721fd927bc1e364abbe63b67/internal/gatewayapi/listeners.go#L128

skriss:
xref #3616

skriss:
@Rycieos FYI I am going to start looking into this to see how we'd implement it. We agree that ultimately, not supporting this use case means we are not Gateway API conformant so we'd like to address it.

skriss:
A few things that need to be sorted out here:
- how to map arbitrary Gateway Listener port numbers to port numbers that Envoy listens on inside the container (external Listener port numbers could be e.g. 81, 5081, 32081)
- how does this interact with Ingress/HTTPProxy resources, which currently assume just 2 ports, HTTP and HTTPS? The easiest option is to say they are incompatible, at least for now (could set a condition on HTTPProxies if Contour is configured for an incompatible Gateway)

skriss:
@Rycieos if you're interested in trying out an early dev build, I have a branch that seems to be basically working:

[...]

PR #5044, opened 2023-02-02
multiple Listeners design doc

skriss:
Updates #4960.

### DIFF
--- a/design/multiple-listeners-design.md
+++ b/design/multiple-listeners-design.md
@@ -90,6 +90,73 @@ N/A
 
 ## Compatibility
 
+### Static Gateway mode
+
+It's also possible to configure Contour to correspond to a particular Gateway, while not using the Gateway provisioner to dynamically manage infrastructure.
+This is described in detail [here](https://projectcontour.io/docs/v1.24.1/guides/gateway-api/#option-1-statically-provisioned).
+In this case, the user is responsible for defining the Envoy Service YAML, including exposing any relevant ports.
+
+For example, given the following Gateway:
+
+```yaml
+kind: Gateway
+apiVersion: gateway.networking.k8s.io/v1beta1
+metadata:
+  name: contour
+  namespace: projectcontour
+spec:
+  gatewayClassName: example
+  listeners:
+    - name: http-1
+      protocol: HTTP
+      port: 80
+    - name: http-2
+      protocol: HTTP
+      port: 80
+    - name: https-1
+      protocol: HTTPS
+      port: 443
+      tls:
+         ...
+    - name: https-2
+      protocol: HTTPS
+      port: 444
+      tls:
+         ...
+```
+
+The user would need to define a corresponding Envoy Service:
+
+```yaml
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: envoy
+  namespace: projectcontour
+spec:
+  ports:
+  - port: 80
+    name: http-1
+    protocol: TCP
+    targetPort: 64592 # matches Contour's mapping scheme described above
+  - port: 81
+    name: http-2
+    protocol: TCP
+    targetPort: 64593 # matches Contour's mapping scheme described above
+  - port: 443
+    name: https-1
+    protocol: TCP
+    targetPort: 64955 # matches Contour's mapping scheme described above
+  - port: 444
+    name: https-2
+    protocol: TCP
+    targetPort: 64956 # matches Contour's mapping scheme described above
+  selector:
+    app: envoy
+  type: LoadBalancer
+```
+
 ### Gateway mode with Ingress/HTTPProxy
 
 When using Gateway mode, it is possible to use Ingress and/or HTTPProxy to define routing config.
--- a/design/multiple-listeners-design.md
+++ b/design/multiple-listeners-design.md
@@ -14,7 +14,7 @@ It's important to distinguish between **Envoy Listeners** and **Gateway Listener
 
 An **Envoy Listener** can be thought of simply as Envoy listening on a given port.
 
-A **Gateway Listener** describes listening on a given port, for requests with a given hostname, and optionally with certain TLS details.
+A **Gateway Listener** encompasses listening on a certain "exposed" port (typically defined on an Envoy LoadBalancer service), which is mapped to an underlying Envoy Listener, for requests with a given hostname, and optionally with certain TLS details.
 In Envoy terminology, this more closely corresponds to the combination of a Listener, a FilterChain, and a RouteConfiguration virtual host.
 
 ## Goals
@@ -79,6 +79,8 @@ Each generated Envoy Listener will have one or more corresponding Envoy RouteCon
 - HTTPS Listeners will have multiple corresponding RouteConfigurations:
     - each attached virtual host will have a RouteConfiguration. For non-Gateway mode, it will be named `https/<fqdn>`, and for Gateway mode, it will be named `<listener name>/<fqdn>`.
     - if any attached virtual host has the fallback certificate enabled, an additional RouteConfiguration will be created. For non-Gateway mode, it will be named `ingress_fallbackcert`, and for Gateway mode, it will be named `<listener name>/fallbackcert`.
+
+For both non-Gateway and Gateway mode, the stats prefix for the HTTP connection manager will equal the Envoy Listener's name, i.e. `ingress_http` and `ingress_https` for non-Gateway mode, and `<http|https>_<port number>` for Gateway mode.
 
 ## Alternatives Considered
 N/A
--- a/design/multiple-listeners-design.md
+++ b/design/multiple-listeners-design.md
@@ -1,0 +1,147 @@
+# Support for multiple Listeners
+
+Status: Draft
+
+## Abstract
+This document proposes changes to Contour in order to support an arbitrary number of Listeners.
+
+## Background
+Historically, Contour has always supported configuring two Envoy Listeners: one for HTTP, and one for HTTPS.
+Root HTTPProxies without TLS enabled have been configured on the HTTP Listener, and those with TLS enabled have been configured on the HTTPS Listener.
+Now that Contour implements Gateway API, it's necessary to support any number of Listeners in order to be fully conformant with the Gateway API spec.
+
+It's important to distinguish between **Envoy Listeners** and **Gateway Listeners**, since they are related but not the same.
+
+An **Envoy Listener** can be thought of simply as Envoy listening on a given port.
+
+A **Gateway Listener** describes listening on a given port, for requests with a given hostname, and optionally with certain TLS details.
+In Envoy terminology, this more closely corresponds to the combination of a Listener, a FilterChain, and a RouteConfiguration virtual host.
+
+## Goals
+- Support Gateways, and associated HTTPRoutes/TLSRoutes, with Listeners configured on >2 ports.
+- Don't break existing functionality for HTTPProxy/Ingress.
+
+## Non Goals
+- Support for HTTPProxy/Ingress attaching to additional Listeners.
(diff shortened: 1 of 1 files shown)


design/multiple-listeners-design.md
```diff
@@ -0,0 +1,147 @@
...
+- Each DAG Listener results in an Envoy Listener being configured. The naming scheme will be described in the Detailed Design section below. 
+- Envoy RouteConfigurations will be generated for each Envoy Listener. The naming scheme will be described in the Detailed Design section below. 
+
+For Gateway mode, we must also consider what happens when HTTPProxy or Ingress is used for routing configuration.
+The details of this use case are described in the Compatibility section below.
+
+## Detailed Design
+
+### internal/gatewayapi
+The `ValidateListeners` function logic must change, to allow any number of unique ports, while still validating protocols and hostnames for each port.
```
Rycieos, line 49: > allow any number of unique ports

The spec _allows_ for ports on different listeners to be the same:
https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.GatewaySpec

> Each listener in a Gateway must have a unique combination of Hostname, Port, and Protocol.
> 
> An implementation MAY group Listeners by Port and then collapse each group of Listeners into a single Listener if the implementation determines that the Listeners in the group are “compatible”. An implementation MAY also group together and collapse compatible Listeners belonging to different Gateways.
> 
> For example, an implementation might consider Listeners to be compatible with each other if all of the following conditions are met:
> 
> 1. Either each Listener within the group specifies the “HTTP” Protocol or each Listener within the group specifies either the “HTTPS” or “TLS” Protocol.
> 2. Each Listener within the group specifies a Hostname that is unique within the group.
> 3. As a special case, one Listener within a group may omit Hostname, in which case this Listener matches when no other Listener matches.

The language makes it seem that this is not required by implementation, but that it should be done where possible.

For clarity, this example Gateway spec should be supported:
```yaml
kind: Gateway
spec:
  listeners:
  - allowedRoutes:
      kinds:
        kind: HTTPRoute
    name: https
    hostname: foo.example.com
    port: 443
    protocol: HTTPS
  - allowedRoutes:
      kinds:
        kind: TLSRoute
    name: tls
    hostname: bar.example.com
    port: 443
    protocol: TLS
```

design/multiple-listeners-design.md
```diff
@@ -0,0 +1,147 @@
...
+## Background
+Historically, Contour has always supported configuring two Envoy Listeners: one for HTTP, and one for HTTPS.
+Root HTTPProxies without TLS enabled have been configured on the HTTP Listener, and those with TLS enabled have been configured on the HTTPS Listener.
+Now that Contour implements Gateway API, it's necessary to support any number of Listeners in order to be fully conformant with the Gateway API spec.
+
+It's important to distinguish between **Envoy Listeners** and **Gateway Listeners**, since they are related but not the same.
+
+An **Envoy Listener** can be thought of simply as Envoy listening on a given port.
+
+A **Gateway Listener** describes listening on a given port, for requests with a given hostname, and optionally with certain TLS details.
```
sunjayBhatia, line 17: maybe add here that this also involves an Envoy Service port as well

design/multiple-listeners-design.md
```diff
@@ -0,0 +1,147 @@
...
+
+### internal/provisioner
+The Gateway controller will use the results of the `ValidateListeners` function (above) to add all relevant ports to `contourModel.Spec.NetworkPublishing.Envoy.Ports`.
+These ports will be populated on the Envoy service and deployment/daemonset.
+
+Currently, the single insecure port is always mapped to port 8080 inside the Envoy container, and the single secure port is always mapped to 8443.
+Supporting any number of Listeners requires a more flexible port mapping scheme, since the requested Gateway Listener ports could be privileged ports (e.g. 1-1023), non-privileged ports (e.g. 8080, 8443), and/or node ports (e.g. 32080).
+The main constraint is that, since Envoy runs as non-root, it cannot bind to privileged ports within the container.
+As such, any requested privileged Gateway Listener port must be mapped to a non-privileged port inside the Envoy container.
+The scheme proposed is as follows:
```
sunjayBhatia, line 60: though conflicts here might be an edge case, we could try to prevent them entirely by setting container ports to increment with the index of the unique Gateway Listener ports, e.g. if they are unique `spec.listeners[0]` gets port 1024 in the container, `spec.listeners[1]` gets 1025, and so on.

this might be a little harder to keep straight in terms of debugging etc., but would reduce some error handling complication etc. since [you can only have up to 64 Listeners at the moment](https://github.com/kubernetes-sigs/gateway-api/blob/e1b21a5ba165c19fc2be50a3599b607e4834cf1a/apis/v1beta1/gateway_types.go#L113) so running out of ports is unlikely

design/multiple-listeners-design.md
```diff
@@ -0,0 +1,147 @@
...
+This automatic redirect functionality is only supported for Ingress and HTTPProxy, while Gateway API can be explicitly configured by the user to do a similar redirect if desired.
+For the purposes of this design, we propose not changing the implementation of this functionality.
+The redirect will continue to be automatically enabled (for Ingress, only if `ingress.kubernetes.io/force-ssl-redirect` is set to `true`).
+It will assume that the "outside" ports are the standard HTTP/HTTPS ports 80 and 443, and will not work properly otherwise.
+
+## Implementation
+An initial PR will refactor Contour internals to allow for any number of Envoy Listeners, while still only programming the two default ones.
+A subsequent PR will make the necessary Gateway mode-related changes to actually program all Gateway Listeners.
+
+## Open Issues
```
sunjayBhatia, line 146: There may be edge cases in the current proposal and the above idea around port mapping where Gateway Listeners change order, are added/removed etc.

This may cause some expensive xDS configuration reloads in Envoy since they involve draining listeners, possibly listening on a new socket, etc.

something to keep in mind when we have added surface area for making this occur more often, not something we have to worry about too much with only 2 static listeners

design/multiple-listeners-design.md
```diff
@@ -0,0 +1,147 @@
...
+The DAG will now store desired Listener names, addresses and "inside" ports.
+Subsequently, in the Ingress/HTTPProxy/Gateway API processors, virtual hosts will be added directly to the appropriate DAG Listener(s), rather than to the root of the DAG.
+The final step of DAG processing will still prune and sort Listeners and virtual hosts.
+
+The Listener naming scheme will be as follows:
+- for non-Gateway mode, the default Listeners will continue to be named `ingress_http` and `ingress_https` for compatibility.
+- for Gateway mode, Listeners will be named `<http|https>_<port number>`, e.g. `http_80` or `https_8443`.
+
+### internal/xdscache
+All DAG Listeners will be converted to Envoy Listeners, using the DAG Listeners' name, address and port.
```
sunjayBhatia, line 76: might want to mention the stat prefix for the new Gateway mode Listeners here too, in case there are any changes

design/multiple-listeners-design.md
skriss: Agreed, Contour should already allow this for the 2 ports that it does permit, this is covered by this test case: https://github.com/projectcontour/contour/blob/main/internal/gatewayapi/listeners_test.go#L172-L208 (though not 100% sure we've confirmed routing via both an HTTPRoute and TLSRoute works in this scenario).

This line in the doc is indicating that we now need to allow this for >2 ports.

design/multiple-listeners-design.md
skriss: Hmm, worth considering, though yeah I do worry about the associated cost if Listeners come and go, and that triggering port re-mapping. The nice thing about a deterministic mapping based on port number is that they will never change.

design/multiple-listeners-design.md
```diff
@@ -0,0 +1,147 @@
...
+The Gateway controller will use the results of the `ValidateListeners` function (above) to add all relevant ports to `contourModel.Spec.NetworkPublishing.Envoy.Ports`.
+These ports will be populated on the Envoy service and deployment/daemonset.
+
+Currently, the single insecure port is always mapped to port 8080 inside the Envoy container, and the single secure port is always mapped to 8443.
+Supporting any number of Listeners requires a more flexible port mapping scheme, since the requested Gateway Listener ports could be privileged ports (e.g. 1-1023), non-privileged ports (e.g. 8080, 8443), and/or node ports (e.g. 32080).
+The main constraint is that, since Envoy runs as non-root, it cannot bind to privileged ports within the container.
+As such, any requested privileged Gateway Listener port must be mapped to a non-privileged port inside the Envoy container.
+The scheme proposed is as follows:
+- any privileged Gateway Listener port (1-1023) is mapped to the range 64513-65535 within the container.
+- any other Gateway Listener port (1024-65535) is mapped as-is within the container.
```
skriss, line 62: Something we have to think about within this range is the admin and stats Listeners that Contour always creates that take up ports 9001 and 8000(?)

design/multiple-listeners-design.md
skriss: The other tricky thing is that multiple Gateway Listeners may correspond to a single port / Envoy Listener, so you'd have to figure out how to handle that case. It's not always 1:1.

skriss: Another option here is to add a static value to all Gateway Listener ports that avoids privileged ports plus the two static Listener ports, so e.g. +9001. I think this is what EG does.

sunjayBhatia: yeah that makes a lot of sense too, not really a chance to run out of ports since a single Gateway has a pretty low limit

design/multiple-listeners-design.md
```diff
@@ -0,0 +1,149 @@
...
+
+## High-Level Design
+Broadly speaking, Contour can operate in one of two modes, which we will call **Gateway mode** and **non-Gateway mode**.
+In Gateway mode, Contour is configured to process a Gateway API Gateway, and gets its Listener configuration from there.
+Non-Gateway mode is the traditional mode of operation, where Listener configuration comes from CLI flags, a config file, and/or a ContourConfiguration resource, plus the Envoy service YAML, and HTTPProxy and/or Ingress is used for routing configuration.
+
+For non-Gateway mode, this design should be fully backwards-compatible.
+There should be no changes observed by a user of Contour in non-Gateway mode.
+
+For Gateway mode, the following changes are needed:
```
sunjayBhatia, line 36: maybe add a little comment on "Gateway mode" which I assume is when you're using the Gateway provisioner etc., vs. "non-Gateway" which is when you're not (using "static" Gateway mode etc.)

(sidebar, maybe some compatibility notes to add about "static" mode?)

sunjayBhatia (approved):
Since we've cut some scope that was holding up the discussion of this before (separated "gateway" and "non-Gateway" mode and not attempted to add redirect support to the former) I think this seems implementable especially since you've done https://github.com/projectcontour/contour/pull/5128 to do this incrementally!

I'm cool to approve, we can keep this open for a little in case others want to comment further

design/multiple-listeners-design.md
sunjayBhatia: regardless of what we pick, we're aware of the fact we need to map ports and avoid conflicts etc. 👍🏽

Rycieos (approved):
LGTM

design/multiple-listeners-design.md
skriss: > maybe add a little comment on "Gateway mode" which I assume is when you're using the Gateway provisioner etc., vs. "non-Gateway" which is when you're not (using "static" Gateway mode etc.)

Are you looking for something more than L29-31? 

> (sidebar, maybe some compatibility notes to add about "static" mode?)

Yep good callout, the tricky thing here is that users will have to update the Envoy service to expose the correct ports, which means they need to understand the port mapping Contour will use.

PR #5049, opened 2023-02-06
Bump Go to 1.20.0

sunjayBhatia:
See release notes: https://go.dev/doc/go1.20


skriss (approved):

internal/featuretests/v3/featuretests.go
```diff
@@ -102,7 +102,8 @@ func setup(t *testing.T, opts ...interface{}) (cache.ResourceEventHandler, *Cont
 
 	registry := prometheus.NewRegistry()
 
-	rand.Seed(time.Now().Unix())
+	// nolint:gosec
+	rand.New(rand.NewSource(time.Now().Unix()))
```
tsaarni, line 106: I think that `rand.New()` returns a new Rand object which is meant to be used like this

```go
r := rand.NewRand(...)
randomInteger := r.Int()
``` 

If I interpret [the doc](https://pkg.go.dev/math/rand#Seed) correctly, that should be used for programs that require a specific (predictable) pseudo-random sequence.  Since we had time as the seed, we did not depend on a specific sequence.

The document also says "_If Seed is not called, the generator is seeded randomly at program startup_" so I think correct approach would be to just remove the call to `Seed()`.

skriss (approved):

PR #5060, opened 2023-02-06
site: generalize upgrading instructions

### LINKED ISSUE #3223: rationalize upgrading instructions

skriss:
Generalizes the upgrading instructions so
they no longer need to be updated for every
Contour version.

Closes #3223.

### DIFF
--- a/site/content/resources/upgrading.md
+++ b/site/content/resources/upgrading.md
@@ -9,6 +9,74 @@ This document describes the changes needed to upgrade your Contour installation.
 
 <div id="toc" class="navigation"></div>
 
+# Before you start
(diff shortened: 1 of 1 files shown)


skriss:
Should've done this long ago, I think these still read fine and shouldn't require any per-version updates!

sunjayBhatia (approved):
👍🏽

PR #5062, opened 2023-02-07
Use http constants instead of string

Fish-pro:
Signed-off-by: Fish-pro <zechun.chen@daocloud.io>

When I read the code, I noticed that some special library calls used custom strings as arguments. Therefore, it is recommended to use package constants instead of strings

I hope the above suggestions can help contribute to the community. Thank you


skriss (approved):
LGTM, thanks @Fish-pro

PR #5063, opened 2023-02-07
Gateway API: Bump to v0.6.1 and run all conformance tests

sunjayBhatia:
~~So we can see in CI what we are currently missing and close the gaps~~

Skips redirect tests that check for original request port in returned
Location header, failing likely due to:
https://github.com/envoyproxy/envoy/issues/17318

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.19
 require (
 	github.com/ahmetb/gen-crd-api-reference-docs v0.3.0
 	github.com/bombsimon/logrusr/v2 v2.0.1
-	github.com/cert-manager/cert-manager v1.11.0-alpha.2
+	github.com/cert-manager/cert-manager v1.11.0
 	github.com/davecgh/go-spew v1.1.1
 	github.com/docker/distribution v2.8.1+incompatible
 	github.com/envoyproxy/go-control-plane v0.10.3-0.20221018195204-799a7af9e5b9
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.19
 require (
 	github.com/ahmetb/gen-crd-api-reference-docs v0.3.0
 	github.com/bombsimon/logrusr/v2 v2.0.1
-	github.com/cert-manager/cert-manager v1.11.0
+	github.com/cert-manager/cert-manager v1.11.0-alpha.2
 	github.com/davecgh/go-spew v1.1.1
 	github.com/docker/distribution v2.8.1+incompatible
 	github.com/envoyproxy/go-control-plane v0.10.3-0.20221018195204-799a7af9e5b9
@@ -39,7 +39,7 @@ require (
 	k8s.io/klog/v2 v2.90.0
 	sigs.k8s.io/controller-runtime v0.14.4
 	sigs.k8s.io/controller-tools v0.11.3
-	sigs.k8s.io/gateway-api v0.6.0
+	sigs.k8s.io/gateway-api v0.6.1
 	sigs.k8s.io/kustomize/kyaml v0.14.0
 )
 
--- a/test/conformance/gatewayapi/gateway_conformance_test.go
+++ b/test/conformance/gatewayapi/gateway_conformance_test.go
@@ -40,24 +40,21 @@ func TestGatewayConformance(t *testing.T) {
 	require.NoError(t, v1beta1.AddToScheme(client.Scheme()))
 
 	cSuite := suite.New(suite.Options{
-		Client:               client,
-		GatewayClassName:     *flags.GatewayClassName,
-		Debug:                *flags.ShowDebug,
-		CleanupBaseResources: *flags.CleanupBaseResources,
-		// Keep the list of supported features in sync with
+		Client:                     client,
+		GatewayClassName:           *flags.GatewayClassName,
+		Debug:                      *flags.ShowDebug,
+		CleanupBaseResources:       *flags.CleanupBaseResources,
+		EnableAllSupportedFeatures: true,
+		// Keep the list of skipped features in sync with
 		// test/scripts/run-gateway-conformance.sh.
-		SupportedFeatures: map[suite.SupportedFeature]bool{
-			suite.SupportReferenceGrant:                     true,
-			suite.SupportTLSRoute:                           true,
-			suite.SupportHTTPRouteQueryParamMatching:        true,
-			suite.SupportHTTPRouteMethodMatching:            true,
-			suite.SupportHTTPResponseHeaderModification:     true,
-			suite.SupportRouteDestinationPortMatching:       true,
-			suite.SupportGatewayClassObservedGenerationBump: true,
-
-			// TODO uncomment the below once they are included in a Gateway API release
-			// suite.HTTPRoutePathRewrite: true,
-			// suite.HTTPRouteHostRewrite: true,
+		SkipTests: []string{
+			// These are skipped because the tests check for the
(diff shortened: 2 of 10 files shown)


skriss:
:+1: do you happen to know if this will be available in GWAPI v0.6.1? Might be best to wait for that + pin to it vs. tracking `main`.

sunjayBhatia:
> 👍 do you happen to know if this will be available in GWAPI v0.6.1? Might be best to wait for that + pin to it vs. tracking `main`.

ah yeah will change to the release branch: https://github.com/kubernetes-sigs/gateway-api/commits/release-0.6

sunjayBhatia:
modulo the invalid/incorrect tests @skriss found these are the failures so far:

```
    --- FAIL: TestConformance/HTTPRouteRedirectHostAndStatus (1.03s)
        --- FAIL: TestConformance/HTTPRouteRedirectHostAndStatus/0_request_to_'/hostname-redirect'_should_receive_a_302 (30.00s)
        --- FAIL: TestConformance/HTTPRouteRedirectHostAndStatus/2_request_to_'/host-and-status'_should_receive_a_301 (30.00s)
    --- FAIL: TestConformance/HTTPRouteRedirectPath (1.03s)
        --- FAIL: TestConformance/HTTPRouteRedirectPath/3_request_to_'/path-and-status'_should_receive_a_301 (30.00s)
        --- FAIL: TestConformance/HTTPRouteRedirectPath/2_request_to_'/path-and-host'_should_receive_a_302 (30.00s)
        --- FAIL: TestConformance/HTTPRouteRedirectPath/1_request_to_'/full/path/original'_should_receive_a_302 (30.00s)
        --- FAIL: TestConformance/HTTPRouteRedirectPath/5_request_to_'/full-path-and-status'_should_receive_a_301 (30.00s)
        --- FAIL: TestConformance/HTTPRouteRedirectPath/4_request_to_'/full-path-and-host'_should_receive_a_302 (30.00s)
        --- FAIL: TestConformance/HTTPRouteRedirectPath/0_request_to_'/original-prefix/lemon'_should_receive_a_302 (30.00s)
    --- FAIL: TestConformance/HTTPRouteRedirectScheme (1.03s)
        --- FAIL: TestConformance/HTTPRouteRedirectScheme/3_request_to_'/scheme-and-host-and-status'_should_receive_a_302 (30.00s)
        --- FAIL: TestConformance/HTTPRouteRedirectScheme/0_request_to_'/scheme'_should_receive_a_302 (30.00s)
        --- FAIL: TestConformance/HTTPRouteRedirectScheme/1_request_to_'/scheme-and-host'_should_receive_a_302 (30.00s)
        --- FAIL: TestConformance/HTTPRouteRedirectScheme/2_request_to_'/scheme-and-status'_should_receive_a_301 (30.00s)
```

skriss:
Re: route redirect path, the implementation is still in progress: https://github.com/projectcontour/contour/issues/4964

sunjayBhatia:
the remaining failures all seem to be around the port the request was made with (e.g. `172.18.255.200:80`) not being preserved in the response `Location` header (error message in tests `expected redirected port to be 80, got  `)

@skriss found https://github.com/envoyproxy/envoy/issues/17318 which looks like the cause, though we may also want to dispute if the test should be checking the port value for standard ports 80/443

Vishal-Chdhry:
> Re: route redirect path, the implementation is still in progress: https://github.com/projectcontour/contour/issues/4964 / https://github.com/projectcontour/contour/pull/4982

@skriss I will get on it right away

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.19
 require (
 	github.com/ahmetb/gen-crd-api-reference-docs v0.3.0
 	github.com/bombsimon/logrusr/v2 v2.0.1
-	github.com/cert-manager/cert-manager v1.11.0
+	github.com/cert-manager/cert-manager v1.11.0-alpha.2
```
skriss, line 8: Can this be reverted?

test/conformance/gatewayapi/gateway_conformance_test.go
```diff
@@ -40,24 +40,21 @@ func TestGatewayConformance(t *testing.T) {
...
-			suite.SupportHTTPRouteQueryParamMatching:        true,
-			suite.SupportHTTPRouteMethodMatching:            true,
-			suite.SupportHTTPResponseHeaderModification:     true,
-			suite.SupportRouteDestinationPortMatching:       true,
-			suite.SupportGatewayClassObservedGenerationBump: true,
-
-			// TODO uncomment the below once they are included in a Gateway API release
-			// suite.HTTPRoutePathRewrite: true,
-			// suite.HTTPRouteHostRewrite: true,
+		SkipTests: []string{
```
skriss, line 50: 👍

skriss (changes requested):

skriss:
> the remaining failures all seem to be around the port the request was made with (e.g. `172.18.255.200:80`) not being preserved in the response `Location` header (error message in tests `expected redirected port to be 80, got `)
> 
> @skriss found [envoyproxy/envoy#17318](https://github.com/envoyproxy/envoy/issues/17318) which looks like the cause, though we may also want to dispute if the test should be checking the port value for standard ports 80/443
> 
> **Note** changing the test to use a port other than 80 for the relevant Listener also fails, so the linked Envoy issue is a more general problem, see test error log below:
> 
> ```
> === NAME  TestConformance/HTTPRouteRedirectPath/1_request_to_'/full/path/original'_should_receive_a_302
>     http.go:186: Response expectation failed for request: {{http   172.18.255.200:8080 /full/path/original  false false   }  HTTP GET map[X-Echo-Set-Header:[]] true [] [] }  not ready yet: expected redirected port to be 8080, got  (after 22.036988368s)
> ```

I'm thinking we can work around this by explicitly setting the redirect port to the Listener's port, if one is not specified in the filter (similar but different to the workaround described in the Envoy issue). Testing out a fix for this now.

skriss (approved):

sunjayBhatia:
> > the remaining failures all seem to be around the port the request was made with (e.g. `172.18.255.200:80`) not being preserved in the response `Location` header (error message in tests `expected redirected port to be 80, got `)
> > @skriss found [envoyproxy/envoy#17318](https://github.com/envoyproxy/envoy/issues/17318) which looks like the cause, though we may also want to dispute if the test should be checking the port value for standard ports 80/443
> > Note changing the test to use a port other than 80 for the relevant Listener also fails, so the linked Envoy issue is a more general problem, see test error log below:
> > ```
> > === NAME  TestConformance/HTTPRouteRedirectPath/1_request_to_'/full/path/original'_should_receive_a_302
> >     http.go:186: Response expectation failed for request: {{http   172.18.255.200:8080 /full/path/original  false false   }  HTTP GET map[X-Echo-Set-Header:[]] true [] [] }  not ready yet: expected redirected port to be 8080, got  (after 22.036988368s)
> > ```
> 
> I'm thinking we can work around this by explicitly setting the redirect port to the Listener's port, if one is not specified in the filter (similar but different to the workaround described in the Envoy issue). Testing out a fix for this now.
> 
> (Or maybe this is the intended way to use the Envoy config after all)

Also checking on if the `HttpConnectionManager.strip_port_mode` field is having any effect here

skriss:
Ah good call, that does seem likely!

sunjayBhatia:
> Ah good call, that does seem likely!

Hm, if I'm not doing something wrong, doesn't seem like it is having any effect on the redirects to turn that off unfortunately

PR #5064, opened 2023-02-07
Avoid comparing full objects during update to optimize performance

### LINKED ISSUE #5024: cmp.Equal() cost too much CPU

i have a k8s env deployed, has below resources from contour metrics
"
contour_eventhandler_operation_total{kind="Endpoints",op="add"} 103
contour_eventhandler_operation_total{kind="Endpoints",op="update"} 2
contour_eventhandler_operation_total{kind="HTTPProxy",op="add"} 1
contour_eventhandler_operation_total{kind="Ingress",op="add"} 4
contour_eventhandler_operation_total{kind="IngressClass",op="add"} 1
contour_eventhandler_operation_total{kind="Secret",op="add"} 521
contour_eventhandler_operation_total{kind="Secret",op="update"} 4
contour_eventhandler_operation_total{kind="Service",op="add"} 103
"
and i observed CPU burst to very high(300M+) for every hours, then i capture the cpu profile and found that the below result when CPU burst happens
"
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top20 --cum
Showing nodes accounting for 6.88s, 45.81% of 15.02s total
Dropped 267 nodes (cum <= 0.08s)
Showing top 20 nodes out of 68
      flat  flat%   sum%        cum   cum%
         0     0%     0%     14.34s 95.47%  github.com/projectcontour/contour/internal/contour.(*EventHandler).Start
     0.01s 0.067% 0.067%     14.34s 95.47%  github.com/projectcontour/contour/internal/contour.(*EventHandler).onUpdate
         0     0% 0.067%     14.34s 95.47%  sigs.k8s.io/controller-runtime/pkg/manager.(*runnableGroup).reconcile.func1
     1.05s  6.99%  7.06%     14.30s 95.21%  github.com/google/go-cmp/cmp.(*state).compareAny
         0     0%  7.06%     14.30s 95.21%  github.com/google/go-cmp/cmp.(*state).comparePtr
         0     0%  7.06%     14.30s 95.21%  github.com/google/go-cmp/cmp.(*state).compareStruct
         0     0%  7.06%     14.30s 95.21%  github.com/google/go-cmp/cmp.Equal
         0     0%  7.06%     14.23s 94.74%  github.com/google/go-cmp/cmp.(*state).compareMap
     0.10s  0.67%  7.72%     14.23s 94.74%  github.com/google/go-cmp/cmp.(*state).compareSlice
     0.11s  0.73%  8.46%      9.05s 60.25%  github.com/google/go-cmp/cmp.(*state).statelessCompare
     0.14s  0.93%  9.39%      7.87s 52.40%  github.com/google/go-cmp/cmp.(*state).tryOptions
     1.17s  7.79% 17.18%      7.63s 50.80%  github.com/google/go-cmp/cmp.Options.filter
     0.48s  3.20% 20.37%      6.46s 43.01%  github.com/google/go-cmp/cmp.pathFilter.filter
     1.81s 12.05% 32.42%      5.98s 39.81%  github.com/google/go-cmp/cmp/cmpopts.structFilter.filter
     0.04s  0.27% 32.69%      4.70s 31.29%  github.com/google/go-cmp/cmp/internal/diff.Difference
     0.04s  0.27% 32.96%      4.63s 30.83%  github.com/google/go-cmp/cmp.(*state).compareSlice.func2
"

i also found that k8s client api (kubernetes-sigs/controller-runtime ) will reload all k8s resources for every 10 hours(default), it means when 10 hour reaches, hundreds of k8s resources update events come to EventHandler.onUpdate() --> cmp.Equal()

here is the description of golang cmp
"
This package is intended to be a more powerful and safer alternative to reflect.DeepEqual for comparing whether two values are semantically equal. It is intended to only be used in tests, as performance is not a goal and it may panic if it cannot compare the values. Its propensity towards panicking means that its unsuitable for production environments where a spurious panic may be fatal.
"

i also notice that for endpoints handler, it use " if oldObj == newObj " directly, so it seems endpoints update has no performance problem
i guess the low performance of cmp.Equal is reflect

but why contour use this low performance library, any improvement could be

no CPU burst happen if no k8s resource update

**Environment:**

- Contour version: 

tsaarni:
This change optimizes processing of `HTTPProxy`, `Secret` and other objects by avoiding full object comparison. This reduces CPU usage during object updates.

Comparing the full objects with `cmp.Equal()` during `EventHandler.onUpdate()` was very costly, which was especially visible in big clusters with a lot of objects. In general, the load caused by `cmp.Equal()` during update was overwhelming compared to the load that Contour otherwise generates in total.

For objects that implement `generation` it can be used, instead of comparing the full object. Generation is updated only when object spec is updated. Updates to `status` or `annotations` fields will not trigger new generation, which needs to be considered separately for objects where those changes matter.

Fixes #5024.

### DIFF
--- a/internal/contour/handler.go
+++ b/internal/contour/handler.go
@@ -18,18 +18,13 @@ package contour
 
 import (
 	"context"
+	"reflect"
 	"time"
 
-	"github.com/google/go-cmp/cmp"
-	"github.com/google/go-cmp/cmp/cmpopts"
-	contour_api_v1 "github.com/projectcontour/contour/apis/projectcontour/v1"
-	contour_api_v1alpha1 "github.com/projectcontour/contour/apis/projectcontour/v1alpha1"
 	"github.com/projectcontour/contour/internal/dag"
 	"github.com/projectcontour/contour/internal/k8s"
 	"github.com/sirupsen/logrus"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	gatewayapi_v1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
-	gatewayapi_v1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 type EventHandlerConfig struct {
@@ -189,23 +184,31 @@ func (e *EventHandler) onUpdate(op interface{}) bool {
 	case opAdd:
 		return e.builder.Source.Insert(op.obj)
 	case opUpdate:
-		if cmp.Equal(op.oldObj, op.newObj,
-			cmpopts.IgnoreFields(contour_api_v1.HTTPProxy{}, "Status"),
-			cmpopts.IgnoreFields(contour_api_v1alpha1.ExtensionService{}, "Status"),
-			cmpopts.IgnoreFields(gatewayapi_v1beta1.GatewayClass{}, "Status"),
-			cmpopts.IgnoreFields(gatewayapi_v1beta1.Gateway{}, "Status"),
-			cmpopts.IgnoreFields(gatewayapi_v1beta1.HTTPRoute{}, "Status"),
-			cmpopts.IgnoreFields(gatewayapi_v1alpha2.TLSRoute{}, "Status"),
-			cmpopts.IgnoreFields(gatewayapi_v1alpha2.GRPCRoute{}, "Status"),
-			cmpopts.IgnoreFields(metav1.ObjectMeta{}, "ResourceVersion"),
-			cmpopts.IgnoreFields(metav1.ObjectMeta{}, "ManagedFields"),
-		) {
-			e.WithField("op", "update").Debugf("%T skipping update, only status has changed", op.newObj)
-			return false
+		old, oldOk := op.oldObj.(client.Object)
+		new, newOk := op.newObj.(client.Object)
+		if oldOk && newOk {
+			equal, err := k8s.IsObjectEqual(old, new)
+			// Error is returned if there was no support for comparing equality of the specific object type.
+			// We can still process the object but it will be always considered as changed.
+			if err != nil {
+				e.WithError(err).WithField("op", "update").
+					WithField("name", new.GetName()).WithField("namespace", new.GetNamespace()).
+					WithField("gvk", reflect.TypeOf(new)).Errorf("error comparing objects")
+			}
+			if equal {
+				// log the object name and namespace to help with debugging.
+				e.WithField("op", "update").
+					WithField("name", old.GetName()).WithField("namespace", old.GetNamespace()).
+					WithField("gvk", reflect.TypeOf(new)).Debugf("skipping update, no changes to relevant fields")
+				return false
+			}
+			remove := e.builder.Source.Remove(op.oldObj)
+			insert := e.builder.Source.Insert(op.newObj)
+			return remove || insert
 		}
-		remove := e.builder.Source.Remove(op.oldObj)
-		insert := e.builder.Source.Insert(op.newObj)
-		return remove || insert
+		// This should never happen.
+		e.WithField("op", "update").Errorf("%T skipping update, object is not a client.Object", op.newObj)
+		return false
 	case opDelete:
 		return e.builder.Source.Remove(op.obj)
 	case bool:
--- a/internal/k8s/helpers.go
+++ b/internal/k8s/helpers.go
@@ -14,12 +14,18 @@
 package k8s
 
 import (
+	"fmt"
+	"reflect"
+
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	contour_api_v1 "github.com/projectcontour/contour/apis/projectcontour/v1"
 	contour_api_v1alpha1 "github.com/projectcontour/contour/apis/projectcontour/v1alpha1"
+	v1 "k8s.io/api/core/v1"
 	networking_v1 "k8s.io/api/networking/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	gatewayapi_v1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 	gatewayapi_v1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 )
@@ -82,4 +88,71 @@ func isStatusEqual(objA, objB interface{}) bool {
 		}
 	}
 	return false
+}
+
+// IsObjectEqual checks if objects received during update are equal.
+//
+// Make an attempt to avoid comparing full objects since it can be very CPU intensive.
+// Prefer comparing Generation when only interested in spec changes.
+func IsObjectEqual(old, new client.Object) (bool, error) {
+
+	// Fast path for any object: when ResourceVersions are equal, the objects are equal.
+	// NOTE: This optimizes the case when controller-runtime executes full sync and sends updates for all objects.
+	if isResourceVersionEqual(old, new) {
+		return true, nil
+	}
+
+	switch old := old.(type) {
+
+	// Fast path for objects that implement Generation and where only spec changes matter.
+	// Status/annotations/labels changes are ignored.
+	// Generation is implemented in CRDs, Ingress and IngressClass.
+	case *contour_api_v1alpha1.ExtensionService,
+		*contour_api_v1.TLSCertificateDelegation:
+		return isGenerationEqual(old, new), nil
+
+	case *gatewayapi_v1beta1.GatewayClass,
+		*gatewayapi_v1beta1.Gateway,
+		*gatewayapi_v1beta1.HTTPRoute,
+		*gatewayapi_v1alpha2.TLSRoute,
+		*gatewayapi_v1beta1.ReferenceGrant,
+		*gatewayapi_v1alpha2.GRPCRoute:
+		return isGenerationEqual(old, new), nil
+
+	// Slow path: compare the content of the objects.
+	case *contour_api_v1.HTTPProxy,
+		*networking_v1.Ingress:
+		return isGenerationEqual(old, new) &&
+			apiequality.Semantic.DeepEqual(old.GetAnnotations(), new.GetAnnotations()), nil
+	case *v1.Secret:
+		if new, ok := new.(*v1.Secret); ok {
+			return reflect.DeepEqual(old.Data, new.Data), nil
+		}
+	case *v1.Service:
+		if new, ok := new.(*v1.Service); ok {
+			return apiequality.Semantic.DeepEqual(old.Spec, new.Spec) &&
+				apiequality.Semantic.DeepEqual(old.Status, new.Status) &&
+				apiequality.Semantic.DeepEqual(old.GetAnnotations(), new.GetAnnotations()), nil
+		}
+	case *v1.Endpoints:
+		if new, ok := new.(*v1.Endpoints); ok {
+			return apiequality.Semantic.DeepEqual(old.Subsets, new.Subsets), nil
+		}
+	case *v1.Namespace:
+		if new, ok := new.(*v1.Namespace); ok {
+			return apiequality.Semantic.DeepEqual(old.Labels, new.Labels), nil
+		}
+	}
+
+	// ResourceVersions are not equal and we don't know how to compare the object type.
+	// This should never happen and indicates that new type was added to the code but is missing in the switch above.
+	return false, fmt.Errorf("do not know how to compare %T", new)
+}
+
+func isGenerationEqual(a, b client.Object) bool {
+	return a.GetGeneration() == b.GetGeneration()
+}
+
+func isResourceVersionEqual(a, b client.Object) bool {
+	return a.GetResourceVersion() == b.GetResourceVersion()
 }
--- a/changelogs/unreleased/5064-tsaarni-small.md
+++ b/changelogs/unreleased/5064-tsaarni-small.md
@@ -1,0 +1,1 @@
+Optimize processing of HTTPProxy, Secret and other objects by avoiding full object comparison. This reduces CPU usage during object updates.
--- a/internal/k8s/helpers_test.go
+++ b/internal/k8s/helpers_test.go
@@ -1,0 +1,195 @@
+// Copyright Project Contour Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package k8s
+
+import (
+	"os"
+	"strings"
+	"testing"
+
+	contour_api_v1 "github.com/projectcontour/contour/apis/projectcontour/v1"
+	contour_api_v1alpha1 "github.com/projectcontour/contour/apis/projectcontour/v1alpha1"
+	"github.com/stretchr/testify/assert"
+	v1 "k8s.io/api/core/v1"
+	networking_v1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gatewayapi_v1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+	gatewayapi_v1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+)
+
+func TestIsObjectEqual(t *testing.T) {
+	tests := []struct {
+		name     string
+		filename string
+		equals   bool
+	}{
+		{
+			name:     "Secret with content change",
+			filename: "testdata/secret-content-change.yaml",
+			equals:   false,
+		},
+		{
+			name:     "Secret with metadata change",
+			filename: "testdata/secret-metadata-change.yaml",
+			equals:   true,
+		},
+		{
+			name:     "Service with status change",
+			filename: "testdata/service-status-change.yaml",
+			equals:   false,
+		},
+		{
+			name:     "Service with annotation change",
+			filename: "testdata/service-annotation-change.yaml",
+			equals:   false,
+		},
+		{
+			name:     "Endpoint with content change",
+			filename: "testdata/endpoint-content-change.yaml",
+			equals:   false,
+		},
+		{
+			name:     "HTTPProxy with annotation change",
+			filename: "testdata/httpproxy-annotation-change.yaml",
+			equals:   false,
+		},
+		{
+			name:     "Ingress with annotation change",
+			filename: "testdata/ingress-annotation-change.yaml",
+			equals:   false,
+		},
+		{
+			name:     "Namespace with label change",
+			filename: "testdata/namespace-label-change.yaml",
+			equals:   false,
+		},
+	}
+
+	scheme := runtime.NewScheme()
+	_ = v1.AddToScheme(scheme)
+	_ = networking_v1.AddToScheme(scheme)
+	_ = contour_api_v1.AddKnownTypes(scheme)
+
+	deserializer := serializer.NewCodecFactory(scheme).UniversalDeserializer()
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			buf, err := os.ReadFile(tc.filename)
+			assert.NoError(t, err)
+
+			// Each file contains two YAML records, which should be compared with each other.
+			objects := strings.Split(string(buf), "---")
+			assert.Equal(t, 2, len(objects), "expected 2 objects in file")
+
+			// Decode the objects.
+			old, _, err := deserializer.Decode([]byte(objects[0]), nil, nil)
+			assert.NoError(t, err)
+			new, _, err := deserializer.Decode([]byte(objects[1]), nil, nil)
+			assert.NoError(t, err)
+
+			got, err := IsObjectEqual(old.(client.Object), new.(client.Object))
+			assert.NoError(t, err)
+			assert.Equal(t, tc.equals, got)
+		})
+	}
+}
+
+func TestIsEqualForResourceVersion(t *testing.T) {
+	old := &v1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            "test",
+			Namespace:       "default",
+			ResourceVersion: "123",
+		},
+		Data: map[string][]byte{
+			"foo": []byte("bar"),
+		},
+	}
+
+	new := old.DeepCopy()
+
+	// Objects with equal ResourceVersion should evaluate to true.
+	got, err := IsObjectEqual(old, new)
+	assert.NoError(t, err)
+	assert.True(t, got)
+
+	// Differences in data should be ignored.
+	new.Data["foo"] = []byte("baz")
+	got, err = IsObjectEqual(old, new)
+	assert.NoError(t, err)
+	assert.True(t, got)
+}
+
+// TestIsEqualFallback compares with ConfigMap objects, which are not supported.
+func TestIsEqualFallback(t *testing.T) {
+	old := &v1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            "test",
+			Namespace:       "default",
+			ResourceVersion: "123",
+		},
+		Data: map[string]string{
+			"foo": "bar",
+		},
+	}
+
+	new := old.DeepCopy()
+
+	// Any object (even unsupported types) with equal ResourceVersion should evaluate to true.
+	got, err := IsObjectEqual(old, new)
+	assert.NoError(t, err)
+	assert.True(t, got)
+
+	// Unsupported types with unequal ResourceVersion should return an error.
+	new.ResourceVersion = "456"
+	got, err = IsObjectEqual(old, new)
+	assert.Error(t, err)
+	assert.False(t, got)
+}
+
+func TestIsEqualForGeneration(t *testing.T) {
+	run := func(t *testing.T, old client.Object) {
+		t.Helper()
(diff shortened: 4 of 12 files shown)


tsaarni:
Notes for reviewers:

- Tests that update objects will now need unique `generation` to trigger update. Those test cases have been updated by calling a function that gives unique `generation`.
- There remains almost 1500 places in test code that still constructs `ObjectMeta` in-place and will have `generation: 0`. I left them untouched since replacing them requires a lot of manual work.
- Status updates still use `cmp.Equal()` ([link](https://github.com/projectcontour/contour/blob/main/internal/k8s/helpers.go)) and there doesn't seem to be straightforward way to avoid comparing the complete object. However, according to profiling this does not seem to be a big issue even for clusters with many objects.
- As mentioned in #5024, the Overview text in [go-cmp](https://pkg.go.dev/github.com/google/go-cmp/cmp) discourages its use for production.

internal/contour/handler.go
```diff
@@ -189,19 +183,12 @@ func (e *EventHandler) onUpdate(op interface{}) bool {
...
-			cmpopts.IgnoreFields(contour_api_v1alpha1.ExtensionService{}, "Status"),
-			cmpopts.IgnoreFields(gatewayapi_v1beta1.GatewayClass{}, "Status"),
-			cmpopts.IgnoreFields(gatewayapi_v1beta1.Gateway{}, "Status"),
-			cmpopts.IgnoreFields(gatewayapi_v1beta1.HTTPRoute{}, "Status"),
-			cmpopts.IgnoreFields(gatewayapi_v1alpha2.TLSRoute{}, "Status"),
-			cmpopts.IgnoreFields(metav1.ObjectMeta{}, "ResourceVersion"),
-			cmpopts.IgnoreFields(metav1.ObjectMeta{}, "ManagedFields"),
-		) {
+		// Ignore if generation is the same.
+		if op.oldObj.(client.Object).GetGeneration() == op.newObj.(client.Object).GetGeneration() {
```
skriss, line 187: I'm observing that in the old code, we did not ignore status for things like Services, Secrets, etc. Wondering if there's any particular need to trigger DAG rebuilds when status for those referenced resources changes? The only thing I can think of off the top of my head is the Envoy service's status.loadBalancer changing which should trigger an update to Ingress/HTTPProxy status, but that goes through a different code path than this.

skriss (commented):
Thanks @tsaarni!

internal/contour/handler.go
tsaarni: True, it was unfortunately not simple as this. Not all objects have generation field set at all.

tsaarni:
After some digging, I've ended up enumerating each type and associated comparison method. For example, all CRDs can be compared by `generation` field, but for some others I used `resourceVersion`. I'm not sure if this makes sense either. The unit tests are not that useful here since the objects are crafted by hand. Preferably updates would be all tested in e2e environment.

Since I now enumerate types, I added default/fallback implementation - this is to catch corner cases where new types are added later but update for `IsObjectEqual()` was missed. The intention is that a full comparison is made in that case.

I experimented with `reflect.DeepEquals()` and compared the performance to `cmp.Equal()`. Curiously, it seemed to me that `reflect.DeepEquals()` does not suffer from the performance problem, even if I let it process thousands of `Secrets` by full comparison. This requires more investigation. Maybe in reality it did not make full comparison in my test...

tsaarni:
For the curious, here is benchmarking:


```go

const numBytes = 5000

func BenchmarkCmpEqual(b *testing.B) {
	a := v1.Secret{
			Name: "test",
		},
		Data: map[string][]byte{
			"test": make([]byte, numBytes),
		},
	}
	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		cmp.Equal(&a, &a)
	}
}

func BenchmarkDeepEqual(b *testing.B) {
	a := v1.Secret{
			Name: "test",
		},
		Data: map[string][]byte{
			"test": make([]byte, numBytes),
		},
	}
	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		reflect.DeepEqual(&a, &a)
	}
}
```

Results:

```
# numBytes = 5000
BenchmarkCmpEqual-20     	     430	   2964049 ns/op	  319721 B/op	     168 allocs/op
BenchmarkDeepEqual-20    	27470511	        40.72 ns/op	       0 B/op	       0 allocs/op

# numBytes = 50000
BenchmarkCmpEqual-20     	      40	  28284435 ns/op	 4606866 B/op	     210 allocs/op
BenchmarkDeepEqual-20    	28174531	        40.81 ns/op	       0 B/op	       0 allocs/op
```

For 5000-byte secret, `cmp.Equal()` took 2.96ms, for 50000-byte secret it took 10 times that. An example of a big secret (50000 bytes and waaay above) is for example helm manifest secrets of type `type: helm.sh/release.v1`.  When installing via helm, it will update the secret which will trigger considerable CPU peak right at the startup.

internal/k8s/helpers.go
```diff
@@ -83,3 +87,60 @@ func isStatusEqual(objA, objB interface{}) bool {
...
+	case *gatewayapi_v1beta1.GatewayClass:
+		return isGenerationEqual(a, b)
+	case *gatewayapi_v1beta1.Gateway:
+		return isGenerationEqual(a, b)
+	case *gatewayapi_v1beta1.HTTPRoute:
+		return isGenerationEqual(a, b)
+	case *gatewayapi_v1alpha2.TLSRoute:
+		return isGenerationEqual(a, b)
+
+	// Fallback implementation: compare the full object if not handled above.
```
skriss, line 125: I'm thinking we could probably just use ResourceVersion comparison as the fallback strategy for any resources that Contour is not updating the status of, WDYT? Would be nice and simple. I don't really see a need to deep-compare.

internal/k8s/helpers.go
```diff
@@ -83,3 +87,60 @@ func isStatusEqual(objA, objB interface{}) bool {
...
+	switch a.(type) {
+	case *v1.Secret:
+		return isResourceVersionEqual(a, b)
+	case *v1.Service:
+		return isResourceVersionEqual(a, b)
+	case *v1.Endpoints:
+		return isResourceVersionEqual(a, b)
+
+	case *networking_v1.Ingress:
+		return isGenerationEqual(a, b)
```
skriss, line 106: Ingress seems like the only non-CRD here; does it implement Generation?

internal/k8s/helpers.go
tsaarni: Yes, I agree.

tsaarni:
I'm still missing filtering of [`ManagedFields`](https://kubernetes.io/docs/reference/using-api/server-side-apply/#field-management) which had caused performance issues previously https://github.com/projectcontour/contour/pull/2987.

I was also thinking that for the types that have only `ResourceVersion`, we could use that for "fast path" decision, taking care of the periodic controller-runtime sync, but then make a full comparison of e.g. `Secret.Data` with `reflect.DeepEqual()` to filter out e.g. annotation/label changes etc?  

```go
return isResourceVersionEqual(a, b) || reflect.DeepEqual(a.(*v1.Secret).Data, b.(*v1.Secret).Data)
```

internal/k8s/helpers.go
tsaarni: > I'm thinking we could probably just use ResourceVersion comparison as the fallback strategy for any resources that Contour is not updating the status of, WDYT? Would be nice and simple. I don't really see a need to deep-compare.

Since it is just fallback, it would be OK even though [`ManagedFields`](https://kubernetes.io/docs/reference/using-api/server-side-apply/#field-management) had caused performance issues previously https://github.com/projectcontour/contour/pull/2987?

tsaarni:
I've added some fine-grained comparin logic that should approximate the previous behavior, with the exception that "fast path" is taken when possible.  For "slow path" I changed to `reflect.DeepEquals()` which is not that bad.

I've added unit tests (not e2e test) but some are based on manifests dumped from real cluster. I'm not sure if that is worth it or not, since we can of course craft those objects by hand as well. 

I have not replaced `cmp.Equal()` from [`k8s.isStatusEqual()`](https://github.com/projectcontour/contour/blob/d044dbd7b72587f4e40052af13f76439f25ed363/internal/k8s/helpers.go#L29) but it may be preferable to remove that too.

It would be also nice to investigate filtering `Secrets` of other types than `Opaque` and `kubernetes.io/tls` no controller-runtime level, so that e.g. huge helm manifest secrets `helm.sh/release.v1` would be dropped as early as possible.

sunjayBhatia (commented):
If possible could we pull the `ObjectMeta` and other helper changes into their own PR? just in case we want to backport or revert etc. the specific changes around improving the object comparison bits

tsaarni:
> If possible could we pull the `ObjectMeta` and other helper changes into their own PR? just in case we want to backport or revert etc. the specific changes around improving the object comparison bits

I created https://github.com/projectcontour/contour/pull/5094 and rebased this PR on top of that. I hope github does not get confused :)

tsaarni:
I've now looked at the remaining use of `cmp.Equal()` which we have in [`k8s.isStatusEqual()`](https://github.com/projectcontour/contour/blob/d044dbd7b72587f4e40052af13f76439f25ed363/internal/k8s/helpers.go#L29).  I'm proposing leaving that untouched for now, even if use of `cmp.Equal()` is not recommended in production environment according to the [package description](https://pkg.go.dev/github.com/google/go-cmp/cmp).

Reasoning:
* It does not seem to cause any considerable CPU churn, not in any scenario that I managed to simulate. 
* Ignoring `LastTransitionTime` fields in various objects looks complicated to me unless it could be done like [here](https://github.com/kubernetes/kubernetes/blob/b0ed87078e4042c12e456c8427dee0fb947a724c/pkg/kubelet/kubelet_node_status.go#L811-L816) in similar code in Kubernetes, simply by ignoring `Conditions` completely?

If you have ideas around this, I'd gladly work on it. Otherwise I'm concluding this PR.

internal/k8s/helpers.go
```diff
@@ -83,3 +89,77 @@ func isStatusEqual(objA, objB interface{}) bool {
...
+
+	// Fast path for any object: when ResourceVersions are equal, the objects are equal.
+	// NOTE: This optimizes the case when controller-runtime executes full sync and sends updates for all objects.
+	if isResourceVersionEqual(old, new) {
+		return true, nil
+	}
+
+	switch old := old.(type) {
+
+	// Fast path for objects that implement Generation and where only spec changes matter.
```
skriss, line 107: This seems like a reasonable comparison for this set of resources given that status/annotation/label changes shouldn't need to trigger rebuilds for them.

internal/k8s/helpers.go
```diff
@@ -83,3 +89,77 @@ func isStatusEqual(objA, objB interface{}) bool {
...
+		return isGenerationEqual(old, new), nil
+
+	case *gatewayapi_v1beta1.GatewayClass:
+		return isGenerationEqual(old, new), nil
+	case *gatewayapi_v1beta1.Gateway:
+		return isGenerationEqual(old, new), nil
+	case *gatewayapi_v1beta1.HTTPRoute:
+		return isGenerationEqual(old, new), nil
+	case *gatewayapi_v1alpha2.TLSRoute:
+		return isGenerationEqual(old, new), nil
```
skriss, line 123: Need to add `*gatewayapi_v1beta1.ReferenceGrant` here as well

internal/k8s/helpers.go
```diff
@@ -83,3 +89,77 @@ func isStatusEqual(objA, objB interface{}) bool {
...
+
+	// Slow path: compare the content of the objects.
+	case *contour_api_v1.HTTPProxy:
+		if new, ok := new.(*contour_api_v1.HTTPProxy); ok {
+			return isGenerationEqual(old, new) &&
+				apiequality.Semantic.DeepEqual(old.ObjectMeta.Annotations, new.ObjectMeta.Annotations), nil
```
skriss, line 129: I _think_ this is reasonable comparison logic, though worth noting that previously, Label changes (for example) would trigger a DAG rebuild whereas now they won't. I can't think of a reason that they would need to trigger a rebuild, but it is a slight change in behavior. Similarly for some of the below objects.

internal/k8s/helpers.go
```diff
@@ -83,3 +89,77 @@ func isStatusEqual(objA, objB interface{}) bool {
...
+		if new, ok := new.(*v1.Endpoints); ok {
+			return apiequality.Semantic.DeepEqual(old.Subsets, new.Subsets), nil
+		}
+
+	default:
+
+	}
+
+	// ResourseVersions are not equal and we don't know how to compare the object type.
+	return false, fmt.Errorf("do not know how to compare %s", old.(client.Object).GetObjectKind().GroupVersionKind().String())
```
skriss, line 156: Do you think it's necessary to return an error here? We're effectively falling back to ResourceVersion comparison, which seems fine. I guess the question is, does logging this have any benefit for the user/operator?

skriss (commented):
Thanks @tsaarni, I'm in favor of this change in general, just being paranoid about any behavior changes given the importance of this logic.

internal/k8s/helpers.go
tsaarni: I was thinking how to alert the developer who might miss adding detailed comparison when introducing new type. I wonder, can it be said that for every handled type, there should be explicit comparison? 

If yes, then without error printout, it might be likely that we only notice the problem much later e.g. when performance impact becomes inevitable.

If no, then I can remove the error.

internal/k8s/helpers.go
tsaarni: Fixed!

internal/k8s/helpers.go
tsaarni: I can add more fields, but it seems that only annotations would get used by Contour. And e.g. object name is immutable, so it is safe to exclude.

Adding everything is bit problematic since `DeepEqual()` and alikes work in opposite way to `cmp.Equal()` which gave the option to ignore named fields.

sunjayBhatia (commented):
this looks great!

I think in some places we could maybe do away with the inline type assertions etc. and also ensure consistent types between `old` and `new` objects with type constraints etc.

I did a little spike on using generics on top of your changes here: https://github.com/sunjayBhatia/contour/commit/92c3b9ecec4f72d1bcc5af988a1da351b5646bdd (still has a hack to use the type assertion switch to get at the internal fields)

let me know what you think (this doesn't necessarily have to be incorporated in this PR)

internal/k8s/helpers.go
skriss: I think it's OK as is; just need to squint at each individual object to make sure we're comparing the full set of things.

internal/k8s/helpers.go
skriss: > I was thinking how to alert the developer who might miss adding detailed comparison when introducing new type. I wonder, can it be said that for every handled type, there should be explicit comparison?

I think this is probably right. In that case, I wonder if it'd be better to use a panic on an unknown object type? This would cause E2E tests to fail, which would then force the code to be updated before getting merged, whereas the error log could always be missed.

As an illustrative example, I know we need to add Namespace and GRPCRoute to this function still, but they are only used for Gateway API so you wouldn't even see the errors in the log unless you were testing with Gateway API. But, with a panic in place, the E2E's would fail.

I know intentionally adding `panic()` to code doesn't sit right with everyone, so curious what folks think.

Another option would be to look to enforce this at compile time, by introducing an interface that all types we cache would be required to implement. I'd like to look into this some more to see if/how it could work, but it doesn't need to be done as part of this PR.

internal/k8s/helpers.go
```diff
@@ -83,3 +89,81 @@ func isStatusEqual(objA, objB interface{}) bool {
...
+// Prefer comparing Generation when only interested in spec changes.
+func IsObjectEqual(old, new interface{}) (bool, error) {
+
+	// Fast path for any object: when ResourceVersions are equal, the objects are equal.
+	// NOTE: This optimizes the case when controller-runtime executes full sync and sends updates for all objects.
+	if isResourceVersionEqual(old, new) {
+		return true, nil
+	}
+
+	switch old := old.(type) {
```
skriss, line 105: need to add handling for `Namespace` and `GRPCRoute`. In the case of Namespace, the Gateway API processor actually looks at the labels, so those need to taken into account for the comparison. GRPCRoute _should_ be like the other two Gateway API route types.

internal/k8s/helpers.go
sunjayBhatia: > Another option would be to look to enforce this at compile time, by introducing an interface that all types we cache would be required to implement. I'd like to look into this some more to see if/how it could work, but it doesn't need to be done as part of this PR.

yeah this is sorta what i was heading towards with the generics approach

we would probably need to add some wrapper types of our own around the k8s core resources or GW API resources if we need to get at some of the internals (spec field, status field, etc.) with a `.Equal` method or something similar and make that interface a constraint

the way i did it doesn't go this far ^ and still has switches on types in the comparison helper, so could be done better with some more thinking

skriss: @sunjayBhatia yeah I like that idea. I'm fine with the impl as-is for now as long as we're looking into some improvements here down the road, since I do think it's super-easy to miss all the various spots to update for new types right now. But, in this case, falling back to ResourceVersion comparison isn't so bad either.

tsaarni: Thank You @sunjayBhatia and @skriss for looking into this!

I must confess that I have not looked much into Go generics yet, so I'm likely not knowledgeable enough to grasp how it could be done :) To me it seems impossible to influence the interfaces in such a way that the problem is converted into compiler time error, because the objects are Kubernetes types, constructed by controller-runtime.

internal/k8s/helpers.go
tsaarni: Thank you for spotting this, I've now added these types!

tsaarni:
>I think in some places we could maybe do away with the inline type assertions etc. and also ensure consistent types between old and new objects with type constraints etc.
>
> I did a little spike on using generics on top of your changes here: [sunjayBhatia@92c3b9e](https://github.com/sunjayBhatia/contour/commit/92c3b9ecec4f72d1bcc5af988a1da351b5646bdd) (still has a hack to use the type assertion switch to get at the internal fields)

Thank you @sunjayBhatia! 

You added great cleanups that I've now included! But I did not understand what the generic `IsObjectEqual()` is adding, if we still need to have the switch-case to compare relevant bits of each type? It does not look very generic :sweat_smile:

BTW I assume we could just trust `old` and `new` having consistent type, even without checking?

tsaarni:
> I wonder if it'd be better to use a panic on an unknown object type?
>...
> falling back to ResourceVersion comparison isn't so bad either.

@skriss I'm torn between these two :thinking:  

Panic would be the correct approach in my mind. No one will read error printouts from e2e tests, and they are not automatically causing test cases to fail either.  As you pointed out, I already missed some types.  But do I then dare to add it with the risk of panicking in production? :-)

skriss:
> > I wonder if it'd be better to use a panic on an unknown object type?
> > ...
> > falling back to ResourceVersion comparison isn't so bad either.
> 
> @skriss I'm torn between these two 🤔
> 
> Panic would be the correct approach in my mind. No one will read error printouts from e2e tests, and they are not automatically causing test cases to fail either. As you pointed out, I already missed some types. But do I then dare to add it with the risk of panicking in production? :-)

Yeah, I feel similarly. I think let's hold off on the panic for now; @sunjayBhatia and I discussed some ideas about at least minimizing the number of places where we have to do type-switching, and then using some wrapper structs + interfaces in other places, which I think might provide the same benefit without the risk of somehow causing a panic in production. We can look into that as a followup. In the mean time, we'll just have to remember to update this for any new types.

tsaarni:
I guess one option could be also to look closer into `cmp.Equal()` and `apiequality.Semantic.DeepEqual()` and try figure out if it is possible to have comparison that combines the flexibility of the former with the speed of the latter.

PS: Sorry, I had some freak accident with the github GUI and ended up editing your previous response although I'm pretty sure I did not go through edit menu :open_mouth: I did edit it back to original, since github does not allow deleting latest edit.

internal/k8s/helpers.go
sunjayBhatia: yeah for sure, the controller-runtime/client-go controller/reconciler/handler interfaces all use `interface{}` there will likely be at the very least 1 place we need to do a type assertion

sunjayBhatia:
> > I think in some places we could maybe do away with the inline type assertions etc. and also ensure consistent types between old and new objects with type constraints etc.
> > I did a little spike on using generics on top of your changes here: [sunjayBhatia@92c3b9e](https://github.com/sunjayBhatia/contour/commit/92c3b9ecec4f72d1bcc5af988a1da351b5646bdd) (still has a hack to use the type assertion switch to get at the internal fields)
> 
> Thank you @sunjayBhatia!
> 
> You added great cleanups that I've now included! But I did not understand what the generic `IsObjectEqual()` is adding, if we still need to have the switch-case to compare relevant bits of each type? It does not look very generic 😅
> 
> BTW I assume we could just trust `old` and `new` having consistent type, even without checking?

yeah we likely need to add wrapper types around the resource types to make things fully generic, esp. to make sure we don't have to do the type assertions you mention to get at the internals of the struct fields

i have some ideas around this area which will require wider changes so no sweat to disregard the mish-mash of generics additions

internal/contour/handler.go
```diff
@@ -189,23 +183,26 @@ func (e *EventHandler) onUpdate(op interface{}) bool {
...
+		new, newOk := op.newObj.(client.Object)
+		if oldOk && newOk {
+			equal, err := k8s.IsObjectEqual(old, new)
+			// Error is returned if there was no support for comparing equality of the specific object type.
+			// We can still process the object but it will be always considered as changed.
+			if err != nil {
+				e.WithError(err).WithField("op", "update").Errorf("%T error comparing objects", op.newObj)
+			}
+			if equal {
+				e.WithField("op", "update").Debugf("%T skipping update, only status has changed", op.newObj)
```
skriss, line 196: It'd be nice to:
- log the namespace/name of the resource here as well, for debugging purposes
- adjust the message to be a little more general since it's not 100% accurate -- something like "skipping update, no changes to relevant fields"

internal/k8s/helpers.go
```diff
@@ -83,3 +89,73 @@ func isStatusEqual(objA, objB interface{}) bool {
...
+		if new, ok := new.(*v1.Endpoints); ok {
+			return apiequality.Semantic.DeepEqual(old.Subsets, new.Subsets), nil
+		}
+	case *v1.Namespace:
+		if new, ok := new.(*v1.Namespace); ok {
+			return apiequality.Semantic.DeepEqual(old.Labels, new.Labels), nil
+		}
+
+	default:
+
```
skriss, line 147: This can be dropped.

internal/k8s/helpers.go
```diff
@@ -83,3 +89,73 @@ func isStatusEqual(objA, objB interface{}) bool {
...
+	}
+
+	// ResourseVersions are not equal and we don't know how to compare the object type.
```
skriss, line 150: nit:
```suggestion
	// ResourceVersions are not equal and we don't know how to compare the object type.
```

skriss (changes requested):
Couple last comments then I think I am comfortable with going ahead with this, thanks @tsaarni!

internal/contour/handler.go
tsaarni: I've added this. 

I also noticed that `obj.GetObjectKind().GroupVersionKind()` is (mostly) empty due to [kubernetes/client-go#541](https://www.github.com/kubernetes/client-go/issues/541) so I switched to printing type of the object.

internal/k8s/helpers.go
tsaarni: Fixed.

internal/k8s/helpers.go
tsaarni: Fixed

skriss (approved):
LGTM, thanks @tsaarni!

Will leave for @sunjayBhatia to take another look as well.

sunjayBhatia (approved):

sunjayBhatia:
thanks for the great work @tsaarni!

ravilr:
Related to https://github.com/projectcontour/contour/pull/4788#discussion_r1129065936

Did anything change with respect to object comparisons b/w contour release v1.22.x and v1.24.1 ? 

we're noticing increased CPU usage (in addition to memory usage) after upgrading to v1.24.1 and cpu profile shows `google/go-cmp` methods in the top samples:

```
$ go tool pprof contourcpu.pprof 
File: contour
Type: cpu
Time: Mar 8, 2023 at 12:31am (PST)
Duration: 120.17s, Total samples = 136.15s (113.30%)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top 20
Showing nodes accounting for 101.59s, 74.62% of 136.15s total
Dropped 482 nodes (cum <= 0.68s)
Showing top 20 nodes out of 66
      flat  flat%   sum%        cum   cum%
    20.06s 14.73% 14.73%     52.57s 38.61%  github.com/google/go-cmp/cmp/cmpopts.structFilter.filter
    14.08s 10.34% 25.08%     30.12s 22.12%  reflect.(*rtype).AssignableTo
    13.69s 10.06% 35.13%     75.38s 55.37%  github.com/google/go-cmp/cmp.Options.filter
     9.04s  6.64% 41.77%     61.61s 45.25%  github.com/google/go-cmp/cmp.pathFilter.filter
     6.09s  4.47% 46.24%      6.15s  4.52%  reflect.implements
     5.76s  4.23% 50.47%      9.89s  7.26%  reflect.directlyAssignable
     4.86s  3.57% 54.04%      4.86s  3.57%  aeshashbody
     4.52s  3.32% 57.36%    119.68s 87.90%  github.com/google/go-cmp/cmp.(*state).compareAny
     3.27s  2.40% 59.76%     14.09s 10.35%  runtime.gcDrain
     2.57s  1.89% 61.65%      9.53s  7.00%  runtime.mapdelete
     2.38s  1.75% 63.40%      2.38s  1.75%  reflect.(*rtype).hasName (inline)
     2.36s  1.73% 65.13%      7.06s  5.19%  runtime.mapassign
     2.16s  1.59% 66.72%      2.16s  1.59%  reflect.(*rtype).Kind (partial-inline)
     2.01s  1.48% 68.20%      8.41s  6.18%  runtime.scanobject
     1.98s  1.45% 69.65%      2.02s  1.48%  runtime.pageIndexOf (inline)
     1.71s  1.26% 70.91%      1.71s  1.26%  runtime.memhash64
     1.38s  1.01% 71.92%      4.98s  3.66%  runtime.mapaccess2
     1.28s  0.94% 72.86%      1.57s  1.15%  runtime.findObject
     1.24s  0.91% 73.77%     76.86s 56.45%  github.com/google/go-cmp/cmp.(*state).tryOptions
     1.15s  0.84% 74.62%      2.86s  2.10%  runtime.typehash
```

tsaarni:
> Did anything change with respect to object comparisons b/w contour release v1.22.x and v1.24.1 ?

@ravilr Hmm, I cannot spot anything specific with a quick glance, but there is a lot of changed code. At least some new resource types seem to have been introduced, mainly related to Gateway API. The comparison was based on the same go-cmp library even previously, but the library version has changed too. I did not compare different versions when benchmarking.

PR #5068, opened 2023-02-08
Gateway API: support RequestRedirect path rewrite

skriss:
Adds support for the Path field on the HTTPRoute
RequestRedirect filter.

Closes #4964.

Signed-off-by: Steve Kriss <krisss@vmware.com>
Co-authored-by: Vishal Choudhary <sendtovishalchoudhary@gmail.com>

I have manually run the Gateway API conformance tests against this PR, with the host check removed since we're having some issues with that, and can confirm all tests pass.

PR #5070, opened 2023-02-08
Gateway API: set redirect port to listener port if not specified

skriss:
(See https://github.com/projectcontour/contour/commit/dca8657f9e7e79b524ca1752c1774f186488f1b8 only)

So this _works_, it passes the conformance tests and works for non-standard ports as well. I'd say we should maybe not explicitly set the port redirect if the Listener is using a standard 80/443 port, since it results in `:80` / `:443` always being appended to the redirect location, except that would cause us to fail the conformance test since it requires a port on the redirect location whether it's a standard port number or not, though that test behavior is debatable.

### DIFF
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -1121,6 +1121,15 @@ func (p *GatewayAPIProcessor) computeHTTPRoute(route *gatewayapi_v1beta1.HTTPRou
 				// custom conformance.", here we choose to just select the first one.
 				if redirect == nil && filter.RequestRedirect != nil {
 					redirect = filter.RequestRedirect
+
+					// Ref. https://github.com/envoyproxy/envoy/issues/17318, Envoy drops
+					// the request port from the redirect location if a port is not specified
+					// in the redirect config, so if the redirect filter does not specify
+					// a port, use the Listener's port.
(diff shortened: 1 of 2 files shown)


skriss:
Rebased this one on main so we can see the conformance results independent of #5068

sunjayBhatia:
> One issue here is that this doesn't play nicely with `scheme` redirects. Need to think about that case some more. Perhaps if we did omit the `port` redirect setting for Listeners using standard ports, it would solve the problem, since presumably if you are doing `scheme` redirects with non-standard ports, then you're also explicitly specifying the `port` as part of your filter config.

would have to take a look at the exact conformance test but yeah this is a little weird, i would guess the tests are incidentally expecting an `https` scheme w/ port 80 in some cases

sunjayBhatia:
One other note from the spec: https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRequestRedirectFilter

`Port is the port to be used in the value of the Location header in the response. When empty, port (if specified) of the request is used.`

There is no conformance test for that case, all of the tests use the GW address+Listener port to construct the request, so if that is added we would still fail if we always set the port to the listener port unfortunately

skriss:
Sounds like an Envoy change to preserve the original request port if not otherwise specified is going to be necessary here.

skriss:
Closing this out for now, will revisit pending outcome of upstream discussions.

PR #5072, opened 2023-02-09
internal/provisioner: set nodePort for LoadBalancerServicePublishingType

izturn:


izturn:
I'd like to set the specific node port for the service instead of generate it by random

izturn:
@skriss  any advise about this use case?

sunjayBhatia (commented):
We don't have a good way to represent both the "outside" port and node ports in play here (where "outside" is the port a cloud provider will configure a load balancer to listen on) since the Gateway Listener only gives us one port to specify

I don't think it makes a whole lot of sense to put NodePort configuration in the GatewayClass params object, since that is cluster-scoped and would require some coordination between ports on the Gateway as well

This idea is something that maybe needs to be brought up in a discussion/issue upstream

izturn:
@sunjayBhatia got it, thx

PR #5076, opened 2023-02-09
internal/sorter: Fix sorting for Query parameter matches

sunjayBhatia:
Needs to take into account match type, value, case sensitivity

Also ensures query param match conditions on routes in sorted order

Fixes a small issue in the header condition sorter as well.

### DIFF
--- a/internal/sorter/sorter_test.go
+++ b/internal/sorter/sorter_test.go
@@ -30,15 +30,12 @@ import (
 	"google.golang.org/protobuf/types/known/wrapperspb"
 )
 
-func shuffleRoutes(routes []*dag.Route) []*dag.Route {
-	shuffled := make([]*dag.Route, len(routes))
-
-	copy(shuffled, routes)
-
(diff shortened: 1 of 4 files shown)


internal/sorter/sorter_test.go
```diff
@@ -507,10 +619,7 @@ func TestSortClusters(t *testing.T) {
 		{Name: "second"},
 	}
 
-	have := []*envoy_cluster_v3.Cluster{
-		want[1],
-		want[0],
-	}
+	have := shuffleSlice(want)
```
skriss, line 622: I like this change in general but with only 2 items in the slice, we have a 50-50 chance of the shuffled slice already being ordered correctly. Maybe we should run all of these tests that involve randomness a bunch of times to avoid any flakes due to that. Could just wrap the shuffles + asserts in a loop that does it 10 times.

skriss (approved):
Thanks for the update, LGTM

Issue #5077, opened 2023-02-10
Lint errors with go1.20 and golangci-lint v1.51.1


tsaarni:
After the linter was updated https://github.com/projectcontour/contour/pull/5050 some new errors appered. However, these do not seem to occur in CI, only for me locally? :thinking:

```console
$ go version
go version go1.20 linux/amd64
$ golangci-lint --version
golangci-lint has version v1.51.1 built from (unknown, mod sum: "h1:N5HD/x0ZrhJYsgKWyz7yJxxQ8JKR0Acc+FOP7QtGSAA=") on (unknown)

$ make lint
Running Go linter ...
tools.go:1: : # github.com/projectcontour/contour
<unknown line number>: internal compiler error: have package "main" (0xc000439bd0), want package "github.com/ahmetb/gen-crd-api-reference-docs" (0xc000439b80)

Please file a bug report including a short program that triggers the error.
https://go.dev/issue/new (typecheck)
//go:build tools
internal/featuretests/v3/featuretests.go:105:2: SA1019: rand.Seed has been deprecated since Go 1.20 and an alternative has been available since Go 1.0: Programs that call Seed and then expect a specific sequence of results from the global random source (using functions such as Int) can be broken when a dependency changes how much it consumes from the global random source. To avoid such breakages, programs that need a specific result sequence should use NewRand(NewSource(seed)) to obtain a random generator that other packages cannot access. (staticcheck)
        rand.Seed(time.Now().Unix())
        ^
make: *** [Makefile:198: lint-golint] Error 1
```

tsaarni:
The errors are related to go1.20.  I've created https://github.com/golangci/golangci-lint/issues/3583 for the internal compiler error.

sunjayBhatia:
i had opened https://github.com/projectcontour/contour/pull/5049 to work this out, hadn't gotten to fixing it up completely, but we can drop the tools package for now while waiting for a fix

tsaarni:
> i had opened #5049 to work this out, hadn't gotten to fixing it up completely, but we can drop the tools package for now while waiting for a fix

:+1: Will you add that to #5049 or would you like me to create separate PR?

PR #5078, opened 2023-02-10
tests: remove call to deprecated rand.Seed

tsaarni:
Remove call to deprecated `rand.Seed()` from the test code, for go1.20 compatibility. https://pkg.go.dev/math/rand#Seed

Updates #5077


tsaarni:
Closing this as duplicate for #5049

PR #5080, opened 2023-02-10
cmd/contour: optional CRDs

### LINKED ISSUE #4684: Optional CRD-s
Rationale: We are using Contour in a multi tenant environment, usually more than one Contour is installed on the same cluster, also different versions are in play.
To solve helm ownership issues on resources, we are handling the required CRD-s in it's own lifecycle as a separate helm package. For that reason we are trying to minimize the amount of CRD-s to take care of, only introduce them to the lifecycle when absolutely required.
As an example: so far have not needed extensionservices, contourconfiguration or contourdeloyment resources.

Currently Contour has a hard dependency on CRD-s, without all of them the Contour process fails to start.
I would propose to optionally create the informers on all CRD resources, only warn on the resources missing instead of a full stop.
That way if some of the CRD-s are missing, Contour can start up, only the related functionalities are left out of use.

nsimons:
Allow disabling certain informers for CRDs by passing a command line flag to the serve command. This makes the corresponding CRD effectively optional.

I'd like some feedback on the terminology. Would the audience of the documents necessarily know what an `informer` is? Also `feature` is quite generic, but I tried to explain the intended meaning in this context. I'm open to other suggestions.

I tried to look for a good place to write some tests, but I didn't find much related to the flags. I have tested manually with the CRDs present/missing in my cluster.

Fixes #4684

Signed-off-by: Niklas Simons <niklas.simons@est.tech>

### DIFF
--- a/changelogs/unreleased/5080-nsimons-minor.md
+++ b/changelogs/unreleased/5080-nsimons-minor.md
@@ -1,0 +1,8 @@
+## Allow Disabling Features
+
+The `contour serve` command takes a new optional flag, `--disable-feature`, that allows disabling
+certain features.
+
+Currently this flag can be used to disable the informer for ExtensionService resources,
+effectively making the ExtensionService CRD optional in the cluster.
+To do this, use the flag as follows: `--disable-feature=extensionservices`
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -126,6 +126,7 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 	serve.Flag("debug", "Enable debug logging.").Short('d').BoolVar(&ctx.Config.Debug)
 	serve.Flag("debug-http-address", "Address the debug http endpoint will bind to.").PlaceHolder("<ipaddr>").StringVar(&ctx.debugAddr)
 	serve.Flag("debug-http-port", "Port the debug http endpoint will bind to.").PlaceHolder("<port>").IntVar(&ctx.debugPort)
+	serve.Flag("disable-feature", "Do not start an informer for the specified resources.").PlaceHolder("<extensionservices>").EnumsVar(&ctx.disabledFeatures, "extensionservices")
 	serve.Flag("disable-leader-election", "Disable leader election mechanism.").BoolVar(&ctx.LeaderElection.Disable)
 
 	serve.Flag("envoy-http-access-log", "Envoy HTTP access log.").PlaceHolder("/path/to/file").StringVar(&ctx.httpAccessLog)
@@ -469,14 +470,22 @@ func (s *Server) doServe() error {
 		Counter: contourMetrics.EventHandlerOperations,
 	}
 
-	// Inform on default resources.
-	for name, r := range map[string]client.Object{
+	// Start to build informers.
+	informerResources := map[string]client.Object{
 		"httpproxies":               &contour_api_v1.HTTPProxy{},
 		"tlscertificatedelegations": &contour_api_v1.TLSCertificateDelegation{},
 		"extensionservices":         &contour_api_v1alpha1.ExtensionService{},
 		"services":                  &corev1.Service{},
 		"ingresses":                 &networking_v1.Ingress{},
-	} {
+	}
+
+	// Some of the resources are optional and can be disabled, do not create informers for those.
+	for _, feat := range s.ctx.disabledFeatures {
+		delete(informerResources, feat)
+	}
+
+	// Inform on the remaining resources.
+	for name, r := range informerResources {
 		if err := informOnResource(r, eventHandler, s.mgr.GetCache()); err != nil {
 			s.log.WithError(err).WithField("resource", name).Fatal("failed to create informer")
 		}
--- a/site/content/docs/main/configuration.md
+++ b/site/content/docs/main/configuration.md
@@ -50,6 +50,7 @@ Many of these flags are mirrored in the [Contour Configuration File](#configurat
 | `--use-proxy-protocol`                                   | Use PROXY protocol for all listeners                                   |
 | `--accesslog-format=<envoy\|json>`                       | Format for Envoy access logs                                           |
 | `--disable-leader-election`                              | Disable leader election mechanism                                      |
+| `--disable-feature=<extensionservices>`                  | Do not start an informer for the specified resources.                  |
 | `--leader-election-lease-duration`                       | The duration of the leadership lease.                                  |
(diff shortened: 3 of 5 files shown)


tsaarni (commented):
This looks good to me 👍 

Regarding terminology: "informer" to me is terminology from client-go, while "watch" could be more of an API level term used in this context? 

In theory, disabled feature _could_ be something else than CRD as well, like in [Kubernetes --feature-gates option](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/) which could be seen conceptually similar to this feature. However, the ask here is of course for optional CRDs specifically (for the capability of skipping install of the CRDs), and the only features that can be disabled now are CRDs.

Some reasoning about potentially disabled features: I was commenting previously https://github.com/projectcontour/contour/issues/4684#issuecomment-1387168532 that the features included in this PR seem to me the only ones that make sense to be disabled currently because

* ContourDeployement CRD needs to be installed only if executing contour gateway-provisioner. Executing it without the CRD does not make much sense either, so users make their choice by executing the provisioner.
* The Gateway API CRDs (multiple CRDs) need to be installed only if Gateway API is configured. Optionality exists there via [the config](https://projectcontour.io/docs/main/configuration/#configuration-example). So, no need for e.g. `--disable-feature=gatewayapi`

Reasoning to add a new command line option, instead of having it in config, could be that this information is needed before ContourConfig CR is looked up.

There might be one small corner case now, if user would add both `-disable-feature=contourconfigurations` and `--contour-config-name=myconfig` where the latter would trigger initial lookup for the CR (code [here](https://github.com/projectcontour/contour/blob/7b2050e1b23df95dc0eddb83cf526eeb4832b91b/cmd/contour/serve.go#L254-L272)) but user then asked for ignore updates to the CR. 

Leaving for others to comment!

nsimons:
Thanks for the feedback @tsaarni!

"watch" could be better, it's a more familiar term since that is also used e.g. with kubectl when watching for changes in objects. I'll leave this one open for now, if there are other opinions.

Nice catch with the corner case, disabling the feature while trying to use it would not make sense. Validation added.

tsaarni (approved):
LGTM :+1:

nsimons:
@skriss as you were involved in the original issue, would you like to review this as well?

cmd/contour/serve.go
```diff
@@ -88,6 +88,14 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 			return fmt.Errorf("cannot specify both %s and %s", "--contour-config", "-c/--config-path")
 		}
 
+		if ctx.contourConfigurationName != "" {
```
skriss, line 91: I believe we can drop this now and just let the Get call fail if the user has specified the flag but doesn't have ContourConfiguration CRDs installed.

cmd/contour/serve.go
```diff
@@ -126,6 +134,7 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 	serve.Flag("debug", "Enable debug logging.").Short('d').BoolVar(&ctx.Config.Debug)
 	serve.Flag("debug-http-address", "Address the debug http endpoint will bind to.").PlaceHolder("<ipaddr>").StringVar(&ctx.debugAddr)
 	serve.Flag("debug-http-port", "Port the debug http endpoint will bind to.").PlaceHolder("<port>").IntVar(&ctx.debugPort)
+	serve.Flag("disable-feature", "Do not start an informer for the specified resource.").PlaceHolder("<extensionservices|contourconfigurations>").EnumsVar(&ctx.disabledFeatures, "extensionservices", "contourconfigurations")
```
skriss, line 137: Drop contourconfigurations

site/content/docs/main/deploy-options.md
```diff
@@ -201,6 +201,11 @@ Next, pass `--envoy-service-http-port=80 --envoy-service-https-port=443` to the
 This is best paired with a DaemonSet (perhaps paired with Node affinity) to ensure that a single instance of Contour runs on each Node.
 See the [AWS NLB tutorial][10] as an example.
 
+## Disabling Features
+
+You can run Contour with certain features disabled by passing `--disable-feature=extensionservices` and/or `--disable-feature=contourconfigurations` flags to the Contour `serve` command.
```
skriss, line 206: drop contourconfigurations

site/content/docs/main/configuration.md
```diff
@@ -16,50 +16,51 @@ The `contour serve` command is the main command which is used to watch for Kuber
...
+| `--envoy-service-http-port=<port>`                              | Kubernetes Service port for HTTP requests                              |
+| `--envoy-service-https-port=<port>`                             | Kubernetes Service port for HTTPS requests                             |
+| `--envoy-service-name=<name>`                                   | Name of the Envoy service to inspect for Ingress status details.       |
+| `--envoy-service-namespace=<namespace>`                         | Envoy Service Namespace                                                |
+| `--use-proxy-protocol`                                          | Use PROXY protocol for all listeners                                   |
+| `--accesslog-format=<envoy\|json>`                              | Format for Envoy access logs                                           |
+| `--disable-leader-election`                                     | Disable leader election mechanism                                      |
+| `--disable-feature=<extensionservices\|contourconfigurations>`  | Do not start an informer for the specified resource.                   |
```
skriss, line 53: drop contourconfigurations

changelogs/unreleased/5080-nsimons-minor.md
```diff
@@ -0,0 +1,5 @@
+## Allow Disabling Features
```
skriss, line 1: drop contourconfigurations

skriss (changes requested):
I have just removed the ContourConfiguration informer entirely in #5133, since the informer wasn't being used and we don't have plans to use it any time soon. So, ContourConfiguration should now be optional in that if you don't use the `--contour-config-name` flag, then the CRD is not required to be present. I think because of that we can drop the special handling for that resource.

Even though this will now just apply to `ExtensionService`, I think it is reasonable to have a way to not require alpha resources to be installed, so still :+1: for going ahead with this.

cmd/contour/serve.go
nsimons: Makes sense!

nsimons:
Thanks @skriss. Fixed the comments, I also double-checked that `ContourConfiguration` is now indeed optional by default.

tsaarni (approved):
Thank you @nsimons! Ignoring coverage targets for this one due to the nature of the change and merging since requested changes were made.

Issue #4952, opened 2022-12-31
support getting LoadBalancerStatus from an Ingress (e.g. AWS ALB) or Gateway


kahirokunn:
I am using Ingress + NodePortS without LoadBalancerService.
Therefore, it is useful to have envoy-ingress-name and envoy-ingress-namespace.
Similarly, it may be useful to have envoy-gateway-name and envoy-gateway-namespace as well, since the Gateway API is expected to become more active in the future.

related Issues: https://github.com/projectcontour/contour/issues/4951

sunjayBhatia:
the `envoy-service-name/namespace` flags are used for identifying which LoadBalancer Service External IP to watch in order to populate the  LoadBalancer IP on Ingress/HTTPProxy resources. I'm not sure how that fits in with the proposed flags you mention here?

kahirokunn:
> LoadBalancer Service External IP to watch in order to populate the LoadBalancer IP on Ingress/HTTPProxy resources.

@sunjayBhatia I also think that flags will be needed to identify which external IP of the Ingress (eg. aws-load-balancer-controller) to monitor.

sunjayBhatia:
I'm not sure what you mean by that, can you clarify a bit? How is your ALB set up to interact with Contour and what information on which resources are you missing?

kahirokunn:
@sunjayBhatia Sorry for the lack of explanation.

1. LoadBalancer Service -> envoy is the default configuration
2. Ingress (ingressClass alb) -> NodePort Service -> envoy is the configuration I use

The reason I use ALB (Ingress + NodePort Service) instead of NLB (LoadBalancer Service) is to use AWS WAF.
https://aws.amazon.com/waf/?nc1=h_ls

Many companies have places where they would like to use AWS WAF.
However, in that case, contour cannot configure the HTTPProxy status.loadbalancer.
If status.loadbalancer is not set, External DNS, etc. will not work, and it will be difficult to find out to which ingress (ALB) it is connected.

sunjayBhatia:
Thanks for spelling that out, makes sense

We have not put too much work into Contour's UX when using an ALB setup so we'll have to think a little about how we want to have this work

kahirokunn:
I see.
Contour is the best product out there, so I'd be very happy if you'd consider the ALB experience as well!
Thx!!!

kahirokunn:
@sunjayBhatia @skriss 
I want to make contour easy to use in ALB as well.
[...]

Issue #2539, opened 2020-05-19
Ingress addresses not set unless using a LoadBalancer


al45tair:
**What steps did you take and what happened:**
Installed Contour and Envoy in a cluster that doesn't have LoadBalancer support. Per the instructions, I configured Envoy to run in a DaemonSet on all the nodes and to use host networking, which is basically the same thing that the community nginx-ingress controller does in the same situation.

Then I created an Ingress to test it.

**What did you expect to happen:**
I expected the Ingress to list the cluster IPs. Instead I saw

    status:
      loadBalancer: {}

**Anything else you would like to add:**
The problem appears to be that Contour watches the "envoy" service to get the addresses to apply to ingresses. If it can't find the "envoy" service, then unless you either add annotations to your Ingress or use the `--ingress-status-address` option, Contour sets an empty LoadBalancerStatus.

Adding annotations is annoying (you don't need to do that with nginx-ingress, after all), and the `--ingress-status-address` option only supports a single address (so, for a multi-node cluster where more than one node can serve for the ingress, you're stuck using a DNS name, which would make External DNS provision a CNAME record, which is undesirable because CNAMEs clash with everything). I'll file a separate issue about `--ingress-status-address` only supporting a single address.

**Environment:**

- Contour version: e1961860
- Kubernetes version: (use `kubectl version`): Client Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.6-beta.0", GitCommit:"e7f962ba86f4ce7033828210ca3556393c377bcc", GitTreeState:"clean", BuildDate:"2020-01-15T08:26:26Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.5", GitCommit:"e0fccafd69541e3750d460ba0f9743b90336f24f", GitTreeState:"clean", BuildDate:"2020-04-16T11:35:47Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/arm64"}
- Kubernetes installer & version: kubeadm version: &version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.5", GitCommit:"e0fccafd69541e3750d460ba0f9743b90336f24f", GitTreeState:"clean", BuildDate:"2020-04-16T11:41:38Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/arm64"}
[...]

PR #5083, opened 2023-02-13
RFC: Add `--envoy-ingress-name`, and `--envoy-ingress-namespace` flags. (#4952)

kahirokunn:
https://github.com/projectcontour/contour/issues/4952

I want to add `--envoy-ingress-name` and `--envoy-ingress-namespace`.
I'm thinking something like this, what do you think?
If you like it, I can implement the rest.

### DIFF
--- a/apis/projectcontour/v1alpha1/contourconfig.go
+++ b/apis/projectcontour/v1alpha1/contourconfig.go
@@ -234,6 +234,12 @@ type EnvoyConfig struct {
 	// +optional
 	Service *NamespacedName `json:"service,omitempty"`
 
+	// Ingress holds Envoy service parameters for setting Ingress status.
+	//
+	// Contour's default is { namespace: "projectcontour", name: "envoy" }.
+	// +optional
+	Ingress *NamespacedName `json:"ingress,omitempty"`
+
 	// Defines the HTTP Listener for Envoy.
 	//
 	// Contour's default is { address: "0.0.0.0", port: 8080, accessLog: "/dev/stdout" }.
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -136,6 +136,8 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 	serve.Flag("envoy-service-https-port", "Kubernetes Service port for HTTPS requests.").PlaceHolder("<port>").IntVar(&ctx.httpsPort)
 	serve.Flag("envoy-service-name", "Name of the Envoy service to inspect for Ingress status details.").PlaceHolder("<name>").StringVar(&ctx.Config.EnvoyServiceName)
 	serve.Flag("envoy-service-namespace", "Envoy Service Namespace.").PlaceHolder("<namespace>").StringVar(&ctx.Config.EnvoyServiceNamespace)
+	serve.Flag("envoy-ingress-name", "Name of the Envoy ingress to inspect for Ingress status details.").PlaceHolder("<name>").StringVar(&ctx.Config.EnvoyIngressName)
+	serve.Flag("envoy-ingress-namespace", "Envoy Ingress Namespace.").PlaceHolder("<namespace>").StringVar(&ctx.Config.EnvoyIngressNamespace)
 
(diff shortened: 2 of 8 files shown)


kahirokunn:
keep

kahirokunn:
keep

skriss:
@kahirokunn first of all I want to apologize for how long it's taken us to review this. We appreciate your contributions to the project!

One thought I have here - maybe instead of having a pair of flags/config fields for Service, Ingress, Gateway, etc., we could just have three flags/fields -- resource type, namespace, and name. This would help avoid flag/config sprawl, and would be more extensible for any future ways of sourcing the LoadBalancerStatus. If we do like this pattern, then we need to decide whether we want to keep the existing fields for Service namespace/name, or deprecate them and replace them with more generally named fields. Deprecation is possible but requires a series of changes over multiple releases to allow folks a chance to migrate to the replacements. What are your thoughts on this approach?

kahirokunn:
@skriss I agree, because that approach worked in all cases with minimal code.
If you want, I can implement it myself.

skriss:
> @skriss I agree, because that approach worked in all cases with minimal code. If you want, I can implement it myself.

:+1: I'd like to get @sunjayBhatia and/or @tsaarni's input here as well before we move forward. Also, if you could take a look at #2539 and see if anything there would impact our approach here, that would be great! I haven't had a chance to review that issue in detail myself yet.

kahirokunn:
Keep

kahirokunn:
@skriss https://github.com/projectcontour/contour/issues/2539
This issue does not affect my problem, but if I implement your suggestion the issue will be resolved.

> we could just have three flags/fields -- resource type, namespace, and name

kahirokunn:
@sunjayBhatia @tsaarni's Could you give me your opinion?

kahirokunn:
keep

tsaarni (commented):
I'm not too familiar with this area, but the idea looks OK to me. 

To me it seems aligned with what we already have for `Service`, and I cannot see it conflicting with having future status discovery/update methods like #2539, since I assume they always need to be mutually exclusive: only one method should be used at a time. Arming two watchers that fight to update status should be considered (and handled early) as a misconfiguration, right? That could be clearer in future documentation too.

BTW, I needed to read your comment here https://github.com/projectcontour/contour/issues/4952#issuecomment-1374330673 to understand the use of this functionality! :smile:

kahirokunn:
keep

kahirokunn:
keep

kahirokunn:
Keep

kahirokunn:
keep

kahirokunn:
keep

tsaarni:
Since there were no objections to suggestion https://github.com/projectcontour/contour/pull/5083#issuecomment-1498347249 it would be great to have short API proposal here - to get this PR unblocked! :sweat_smile: 

@skriss wrote: 
> One thought I have here - maybe instead of having a pair of flags/config fields for Service, Ingress, Gateway, etc., we could just have three flags/fields -- resource type, namespace, and name. This would help avoid flag/config sprawl, and would be more extensible for any future ways of sourcing the LoadBalancerStatus. 

One more thought: what do people think about single switch with format like `DISCOVERY_SCHEME:PARAMS`, for example `--loadbalancer-status=service:namespace/resource`, `--loadbalancer-status=ingress:namespace/resource`, `--loadbalancer-status=hostname:fqdn,hostname:fqdn` ...

kahirokunn:
keep

kahirokunn:
keep

kahirokunn:
keep

skriss:
I like @tsaarni's suggestion above for the flag design.

@kahirokunn to move this forward, could you write up a brief summary of the changes we plan here, taking into account the comments above? In short, I think we would be adding two new generic flags that would provide the existing functionality for getting the IP from a Service, adding the functionality to alternate get it from an Ingress, and then deprecating the existing `--envoy-service-name/namespace` flags (but we can't remove them for a couple of releases).

Assuming everyone supports that approach, then we can move forward with the implementation.

kahirokunn:
@skriss Sorry, I agree with you.
However, I have lost track of time and it looks like I will not be able to proceed with this implementation.
I would appreciate it if someone else could implement a new one instead.

PR #5084, opened 2023-02-13
drop old +build tags

skriss:
The +build syntax has been replaced by go:build
and the old tags are no longer needed.

### DIFF
--- a/cmd/contour/gcp.go
+++ b/cmd/contour/gcp.go
@@ -12,10 +12,9 @@
 // limitations under the License.
 
 //go:build gcp
-// +build gcp
 
 package main
 
-// This file is protected by the +build gcp tag above to prevent
+// This file is protected by the go:build gcp tag above to prevent
 // the gcp dependencies from being part of the standard contour image.
 import _ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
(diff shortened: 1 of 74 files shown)


cmd/contour/gcp.go
```diff
@@ -12,10 +12,9 @@
 // limitations under the License.
 
 //go:build gcp
-// +build gcp
 
 package main
 
-// This file is protected by the +build gcp tag above to prevent
+// This file is protected by the go:build gcp tag above to prevent
 // the gcp dependencies from being part of the standard contour image.
```
skriss, line 19: This comment led me down a bit of a rabbit-hole, since as far as I knew, we _do_ specify these build tags when building the image:

- https://github.com/projectcontour/contour/blob/main/Makefile#L84
- https://github.com/projectcontour/contour/blob/main/Makefile#L112-L113
- https://github.com/projectcontour/contour/blob/main/Dockerfile#L33

It appears that the original intent was to not include these in the actual image builds, since they're only needed for running the binary locally in certain environments, not in an in-cluster setting:
- https://github.com/projectcontour/contour/pull/170

However, during some subsequent Dockerfile refactoring we moved to using the `make` targets within the Dockerfile, which had the side effect of starting to include these:
- https://github.com/projectcontour/contour/pull/2383

So, I'll file a separate issue to decide what to do here. For now just doing the mechanical update to the comment and ignoring the fact that it's not currently accurate.

sunjayBhatia (approved):

PR #5087, opened 2023-02-13
test/e2e/gateway: Remove more duplicate tests

sunjayBhatia:
- Remove path condition match test, duplicate of test added in https://github.com/kubernetes-sigs/gateway-api/commit/be595a279711926ef08b538630d2f0ea6c65878b
- Remove TLSRoute test, covered by test added in https://github.com/kubernetes-sigs/gateway-api/commit/a2014943b5c960f6d125bc4267536926665a25a4


test/e2e/gateway/tlsroute_test.go
```diff
@@ -1,74 +0,0 @@
...
-		res, ok := f.HTTP.SecureRequestUntil(&e2e.HTTPSRequestOpts{
-			Host:      "passthrough.tlsroute.gatewayapi.projectcontour.io",
-			Condition: e2e.HasStatusCode(200),
-		})
-		require.NotNil(t, res, "request never succeeded")
-		assert.Truef(t, ok, "expected 200 response code, got %d", res.StatusCode)
-		assert.Equal(t, "echo", f.GetEchoResponseBody(res.Body).Service)
-
-		// Ensure request doesn't route when non-matching SNI is provided
-		require.Never(f.T(), func() bool {
```
sunjayBhatia, line 67: the gw api test doesn't do this, so we can keep it if we think its important (and/or add it upstream)

with these negative cases I definitely get the intent, but to have a fully sufficient test that tests it all the way would be v difficult to add

skriss (approved):
LGTM, I think we can live without the negative case.

PR #5092, opened 2023-02-13
don't generate ingress_http RouteConfig if no vhosts

skriss:

Ran into this while working on some refactoring for multiple Listeners, and I'm not sure why we generate/send an empty `ingress_http` RouteConfiguration if there are no vhosts, since we're not generating a Listener/HCM to reference the RouteConfiguration in this case. For HTTPS, we're only generating the RouteConfigs if there are vhosts.

Does anyone see an issue with changing this behavior? I'll have a bunch of unit tests to update but E2E's look like they pass just fine with the change in place.

### DIFF
--- a/internal/featuretests/v3/corspolicy_test.go
+++ b/internal/featuretests/v3/corspolicy_test.go
@@ -461,10 +461,5 @@ func TestCorsPolicy(t *testing.T) {
 
 	rh.OnAdd(invvhost)
 
-	c.Request(routeType).Equals(&envoy_discovery_v3.DiscoveryResponse{
-		Resources: resources(t,
-			envoy_v3.RouteConfiguration("ingress_http")),
-		TypeUrl: routeType,
-	}).Status(invvhost).IsInvalid()
-
+	c.Request(routeType).HasNoResources().Status(invvhost).IsInvalid()
 }
--- a/internal/featuretests/v3/directresponsepolicy_test.go
+++ b/internal/featuretests/v3/directresponsepolicy_test.go
@@ -118,10 +118,5 @@ func TestDirectResponsePolicy_HTTProxy(t *testing.T) {
 
 	rh.OnUpdate(proxyNobody, proxyInvalid)
 
-	c.Request(routeType).Equals(&envoy_discovery_v3.DiscoveryResponse{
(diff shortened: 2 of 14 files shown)


skriss:
@sunjayBhatia when you get a chance would appreciate your thoughts on this

sunjayBhatia:
in general i think this change makes sense, no need to send over extra data etc. if there are no routes

when you remove all HTTProutes for example with/without this change the `ingress_http` listener should be removed completely:
```
$ curl -s localhost:9001/listeners
envoy-admin::127.0.0.1:9001
stats-health::0.0.0.0:8002
```

though the listener config dump does show the listener in draining state still referencing the `ingress_http` route config:
```
    {
     "name": "ingress_http",
     "draining_state": { <-------
      "version_info": "1",
      "listener": {
       "@type": "type.googleapis.com/envoy.config.listener.v3.Listener",
       "name": "ingress_http",
       "address": {
        "socket_address": {
         "address": "0.0.0.0",
         "port_value": 8080
        }
       },
       "filter_chains": [
        {
         "filters": [
          {
           "name": "envoy.filters.network.http_connection_manager",
           "typed_config": {
            "@type": "type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager",
            "stat_prefix": "ingress_http",
            ...
             "route_config_name": "ingress_http" <--------
            },
```

when applying this change I do still see the dynamic route config:
```
   "dynamic_route_configs": [
    {
     "version_info": "4",
     "route_config": {
      "@type": "type.googleapis.com/envoy.config.route.v3.RouteConfiguration",
      "name": "ingress_http"
     },
     "last_updated": "2023-02-16T17:40:07.899Z"
    }
   ]
```

which I think comes from here: https://github.com/projectcontour/contour/blob/687a910f5fa7c60fd0f2b2210d4b1bd4cef1c7b4/internal/xdscache/v3/route.go#L69-L77

so a slight change in behavior in practice w/ Envoy, would have to look at history to see why we do that

skriss:
Closing this out for now, I have a way to not change this behavior at the moment and I'd rather keep the impact of the multiple Listeners refactoring as small as I can.

PR #5099, opened 2023-02-16
Reduce the memory footprint by optimizing Kubernetes informer cache

tsaarni:
This PR tries to avoid storing data in controller-runtime / client-go cache when we know for sure we do not need it.

While working on https://github.com/projectcontour/contour/pull/5064#issuecomment-1424616124 it was discovered that considerable time was spent on processing `Secrets` of type `helm.sh/release.v1`. These secrets contain the helm manifest, and they are just noise for Contour. Besides CPU impact, there is also memory footprint impact, since these `Secrets` can get large. See related discussion [cert-manager/cert-manager#5220](https://www.github.com/cert-manager/cert-manager/issues/5220).

I was looking for ways to avoid caching objects that fail to fulfil given criteria. I found one: [`SelectorByObject`](https://github.com/kubernetes-sigs/controller-runtime/blob/master/designs/use-selectors-at-cache.md). It is based on filtering at the API server side by using [selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/). Unfortunately, selectors will not be flexible enough for this use case.

Another approach was making the objects smaller by providing custom [transform function](https://www.github.com/kubernetes-sigs/controller-runtime/pull/1805). The change presented in this PR uses transform function to set `Secret.data` to `nil` for objects that clearly are not relevant for Contour.

Yet another approach, although more complicated: [metadata-only watches](https://www.github.com/kubernetes-sigs/controller-runtime/pull/1174) are interesting from security perspective. It is not desirable that a process holds _all_ `Secrets` of the cluster needlessly in memory. As a side effect, the approach could also lower memory footprint. Curiously, the "actual" data can find its way to metadata too, in a form of `last-applied-configuration` annotation, see for example [here](https://www.github.com/kubernetes/kubernetes/issues/23564) and [here](https://www.github.com/kubernetes/kubernetes/issues/29923). To solve this problem, one could combine metadata-only watches with transform function.

**NOTE1** I have not found an obvious way to prove if there is a meaningful impact to memory footprint from this change.

**NOTE2** If this optimization has been discussed already before, sorry for the repeat!

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -216,6 +216,46 @@ func NewServer(log logrus.FieldLogger, ctx *serveContext) (*Server, error) {
 		Scheme:                 scheme,
 		MetricsBindAddress:     "0",
 		HealthProbeBindAddress: "0",
+		NewCache: ctrl_cache.BuilderWithOptions(ctrl_cache.Options{
+			// TransformByObject is a function that allows changing incoming objects before they are cached by the informer.
+			// This is useful for saving memory by removing fields that are not needed by Contour.
+			TransformByObject: ctrl_cache.TransformByObject{
+				&corev1.Secret{}: func(obj interface{}) (interface{}, error) {
+					secret, ok := obj.(*corev1.Secret)
+					// TransformFunc should handle the tombstone of type cache.DeletedFinalStateUnknown
+					if !ok {
+						return obj, nil
+					}
+
+					// Do not touch Secrets that might be needed.
+					if secret.Type == corev1.SecretTypeTLS || secret.Type == corev1.SecretTypeOpaque {
+						return obj, nil
+					}
+
+					// Other types of Secrets will never be referred to, so we can remove all data.
+					// For example Secrets of type helm.sh/release.v1 can be quite large.
+					// Last-applied-configuration annotation might contain a copy of the complete data.
+					secret.Data = nil
+					secret.SetManagedFields(nil)
+					secret.SetAnnotations(nil)
+
+					// Returning error will avoid handlers from being called.
+					// This does not avoid caching.
+					return nil, fmt.Errorf("ignoring secret %s/%s of type %s", secret.Namespace, secret.Name, secret.Type)
+				},
+			},
+			// DefaultTransform is called for objects that do not have a TransformByObject function.
+			DefaultTransform: func(obj interface{}) (interface{}, error) {
+				o, ok := obj.(client.Object)
+				// TransformFunc should handle the tombstone of type cache.DeletedFinalStateUnknown
+				if !ok {
+					return obj, nil
+				}
+
+				o.SetManagedFields(nil)
+				return o, nil
+			},
+		}),
 	}
 	if ctx.LeaderElection.Disable {
 		log.Info("Leader election disabled")
--- a/changelogs/unreleased/5099-tsaarni-small.md
+++ b/changelogs/unreleased/5099-tsaarni-small.md
@@ -1,0 +1,1 @@
+Optimized the memory usage when handling Secrets in Kubernetes client informer cache.


ravilr:
xref https://github.com/projectcontour/contour/pull/4788#discussion_r1129065936

ravilr:
> I was looking for ways to avoid caching objects that fail to fulfil given criteria. I found one: [SelectorByObject](https://github.com/kubernetes-sigs/controller-runtime/blob/master/designs/use-selectors-at-cache.md). It is based on filtering at the API server side by using [selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/). Unfortunately, selectors will not be flexible enough for this use case.


Helm created secrets of type `helm.sh/release.v1` have the `owner: helm` label.  Traefik (another ingress controller) uses this `owner!=helm` as a selector to its secrets informer cache: https://github.com/traefik/traefik/blob/56f7515ecd8572c6a4c73f35e54b33414e6efb79/pkg/provider/kubernetes/ingress/client.go#L186

IMO, selector approach would result in filtering at the Kubernetes APIServer itself, so that the Contour's informer cache and eventHandler wouldn't even see them.  the transformer approach is effectively nullifying it at the client cache side (meaning still work for the contour controller side).

tsaarni:
> Helm created secrets of type helm.sh/release.v1 have the owner: helm label. Traefik (another ingress controller) uses this owner!=helm as a selector to its secrets informer cache: https://github.com/traefik/traefik/blob/56f7515ecd8572c6a4c73f35e54b33414e6efb79/pkg/provider/kubernetes/ingress/client.go#L186

Thank you @ravilr, great finding! :+1:  

I did not consider specifically addressing Helm before, but you are correct, doing that would allow using label selector -based approach. Since Helm Secrets can be the biggest and most well-known problem, it can make sense to address that specifically.  Even though from a security perspective it still would be desirable not to keep _any_ extra Secrets resident in the process's memory.

I'll make an attempt with selectors to address the Helm issue.

ravilr:
Another option is to filter on field selector, but it appears client-go doesn't support set-based requirements for field selector( `type IN (Opaque, kubernetes.io/tls)`). 

But, this seems to work:
`kubectl get secrets -A --field-selector=type!=helm.sh/release.v1,type!=kubernetes.io/service-account-token` 

Above will also filter out ServiceAccount tokens which are auto-refreshed every hour and could have significant events traffic in some clusters with large number of ServiceAccount resources.

sunjayBhatia:
We got some reports of high memory usage between Contour versions and doing some experiments, it doesn't seem the Helm secrets being added to the DAG cache don't make any difference:

Setup:
- Kind cluster
- Contour v1.23.0 in one ns
- Contour v1.24.3 in another
- One HTTPProxy (reconciled by both)
- Enable metrics-server

Initial resource usage via `kubectl top pod`:

v1.23.0:
```
NAME                      CPU(cores)   MEMORY(bytes)
contour-547659bf5-qqscz   14m          21Mi
contour-547659bf5-sp5dw   21m          25Mi
envoy-gxvnm               7m           32Mi
```

v1.24.3:
```
NAME                     CPU(cores)   MEMORY(bytes)
contour-6f944569-crrn7   25m          26Mi
contour-6f944569-tzdvz   15m          22Mi
envoy-kz2ct              7m           35Mi
```

Add secrets of type `helm.sh/release.v1`, 50 at a time, each with data field of
size 2^17 bytes (131072), largest power of 2 that would let secret be created.

After 600 secrets created in batches, and the cluster sitting idle for a few minutes:

v1.23.0:
```
NAME                      CPU(cores)   MEMORY(bytes)
contour-547659bf5-qqscz   14m          258Mi
contour-547659bf5-sp5dw   20m          299Mi
envoy-gxvnm               6m           34Mi
```

v1.24.3:
```
NAME                     CPU(cores)   MEMORY(bytes)
contour-6f944569-crrn7   23m          301Mi
contour-6f944569-tzdvz   14m          259Mi
envoy-kz2ct              7m           36Mi
```

After 1000 secrets created in batches, and the cluster sitting idle for a few minutes:

v1.23.0:
```
NAME                      CPU(cores)   MEMORY(bytes)
contour-547659bf5-qqscz   13m          393Mi
contour-547659bf5-sp5dw   22m          434Mi
envoy-gxvnm               6m           34Mi
```

v1.24.3:
```
NAME                     CPU(cores)   MEMORY(bytes)
contour-6f944569-crrn7   23m          435Mi
contour-6f944569-tzdvz   14m          395Mi
envoy-kz2ct              6m           36Mi
```

After 3000 secrets created in batches, and the cluster sitting idle for a few minutes:

v1.23.0:
```
NAME                      CPU(cores)   MEMORY(bytes)
contour-547659bf5-qqscz   13m          1051Mi
contour-547659bf5-sp5dw   21m          1070Mi
envoy-gxvnm               6m           34Mi
```

v1.24.3:
```
NAME                     CPU(cores)   MEMORY(bytes)
contour-6f944569-crrn7   23m          1105Mi
contour-6f944569-tzdvz   14m          1070Mi
envoy-kz2ct              6m           36Mi
```

tsaarni:
@sunjayBhatia Do you refer to https://github.com/projectcontour/contour/pull/4788 and do you mean that unrelated secrets are part of DAG cache as well as in client-go cache?  Could the reason for smallish difference be that they both share the underlying data?

Still the figures you included seemed to indicate an increase of footprint from initial ~25Mi to ~1000Mi while just adding unrelated secrets to the cluster, so from that perspective the impact of caching unrelated objects on client-go level would be huge. 

I found it a bit surprising that I did not find convenient method to write a "predicate" function that could control it on the client end.

sunjayBhatia:
> @sunjayBhatia Do you refer to #4788 and do you mean that unrelated secrets are part of DAG cache as well as in client-go cache? Could the reason for smallish difference be that they both share the underlying data?
> 
> Still the figures you included seemed to indicate an increase of footprint from initial ~25Mi to ~1000Mi while just adding unrelated secrets to the cluster, so from that perspective the impact of caching unrelated objects on client-go level would be huge.
> 
> I found it a bit surprising that there was no method to write a "predicate" function that could control caching on the client end. Or ways to manage cache footprint in general, besides the ones I mentioned in the description.

sorry yes, us explicitly adding references to the secrets does not seem to make a difference in memory usage vs. when we do not, the client-go informer cache still contains this content

agree that we should try to make sure unrelated content doesn't show up in the informer caches, but it would seem simply adding a filter to what is added to the DAG cache doesn't really have a material impact on memory usage

sunjayBhatia:
With PR 5099 (rebased on `main`):

Initial resource usage:
```
NAME                      CPU(cores)   MEMORY(bytes)
contour-869c8b576-h6gqc   1m           14Mi
contour-869c8b576-xhzcw   2m           17Mi
envoy-z6ffk               8m           29Mi
```

After 600 secrets created in batches, and the cluster sitting idle for a few minutes:
```
NAME                      CPU(cores)   MEMORY(bytes)
contour-869c8b576-h6gqc   1m           15Mi
contour-869c8b576-xhzcw   2m           18Mi
envoy-z6ffk               8m           29Mi
```

After 1000 secrets created in batches, and the cluster sitting idle for a few minutes:
```
NAME                      CPU(cores)   MEMORY(bytes)
contour-869c8b576-h6gqc   1m           15Mi
contour-869c8b576-xhzcw   2m           19Mi
envoy-z6ffk               7m           29Mi
```

After 3000 secrets created in batches, and the cluster sitting idle for a few minutes:
```
NAME                      CPU(cores)   MEMORY(bytes)
contour-869c8b576-h6gqc   1m           16Mi
contour-869c8b576-xhzcw   2m           19Mi
envoy-z6ffk               7m           29Mi
```

During secret creation, memory usage does increase slightly, but I never
observed anything > 30Mi

cmd/contour/serve.go
```diff
@@ -215,6 +215,24 @@ func NewServer(log logrus.FieldLogger, ctx *serveContext) (*Server, error) {
...
+					secret.Data = nil
+					// Returning error will avoid handlers from being called.
+					// Object will still be cached.
+					return nil, fmt.Errorf("ignoring secret %s/%s of type %s", secret.Namespace, secret.Name, secret.Type)
+				},
+			},
+			DefaultTransform: func(obj interface{}) (interface{}, error) {
+				obj.(client.Object).SetManagedFields(nil)
+				return obj, nil
+			},
```
skriss, line 234: Based on godoc it seems like this won't get applied to Secrets, so would be good to clear ManagedFields for all of them as well

skriss:
> I can resurrect this PR if wanted?

+1, I think this would be a great optimization to get in

tsaarni:
> FYI I'm planning to merge #5214 shortly and looks like this'll conflict. Actually, before I merge, let's confirm that these two are actually compatible with each other?

Apparently they are not compatible :(  

`BuilderWithOptions` combines user provided options with the default options ([link](https://github.com/kubernetes-sigs/controller-runtime/blob/53057ba616d1ba05ea6e295c1da870a555d9988d/pkg/cache/cache.go#L250-L257)).
`MultiNamespacedCacheBuilder` does not seem to provide means to pass these extra parameters ([link](https://github.com/kubernetes-sigs/controller-runtime/blob/53057ba616d1ba05ea6e295c1da870a555d9988d/pkg/cache/multi_namespace_cache.go#L48-L49)). I need to check tomorrow how to compose these to create `NewCacheFunc` that considers both aspects.

Meanwhile I will push rebased and cleaned up version of this PR.

tsaarni:
There is forthcoming breaking change in controller-runtime API. To my surprise, both multinamespace and transform function use cases seem to be impacted by it. For example, [kubernetes-sigs/controller-runtime/pull/2157](https://www.github.com/kubernetes-sigs/controller-runtime/pull/2157) deprecates `MultiNamespacedCacheBuilder()` which is used in #5214. The API change was merged, and further refactoring has happened since then. For example, the API used by this PR for transformations seems changed too. 

I did not yet test with controller-runtime main, but it seems feasible that all these changes might work in the best way after all, removing any conflict that may arise between this PR and #5214. None of this seems to target latest 0.14 release track which Contour is using currently.

cmd/contour/serve.go
```diff
@@ -216,6 +216,36 @@ func NewServer(log logrus.FieldLogger, ctx *serveContext) (*Server, error) {
 		Scheme:                 scheme,
 		MetricsBindAddress:     "0",
 		HealthProbeBindAddress: "0",
+		NewCache: ctrl_cache.BuilderWithOptions(ctrl_cache.Options{
+			// TransformByObject is a function that allows changing incoming objects before they are cached by the informer.
+			// This is useful for saving memory by removing fields that are not needed by Contour.
+			TransformByObject: ctrl_cache.TransformByObject{
+				&corev1.Secret{}: func(obj interface{}) (interface{}, error) {
+					secret, _ := obj.(*corev1.Secret)
```
skriss, line 224: The godoc for `TransformFunc` says:

> // TransformFunc (similarly to ResourceEventHandler functions) should be able
> // to correctly handle the tombstone of type cache.DeletedFinalStateUnknown

See e.g. https://github.com/projectcontour/contour/blob/main/internal/dag/cache.go#L279-L280 for how we handle that in the DAG cache currently - I guess we should do something similar here.

cmd/contour/serve.go
```diff
@@ -216,6 +216,36 @@ func NewServer(log logrus.FieldLogger, ctx *serveContext) (*Server, error) {
...
+					secret.SetAnnotations(nil)
+
+					// Returning error will avoid handlers from being called.
+					// This does not avoid caching.
+					return nil, fmt.Errorf("ignoring secret %s/%s of type %s", secret.Namespace, secret.Name, secret.Type)
+				},
+			},
+			// DefaultTransform is called for objects that do not have a TransformByObject function.
+			DefaultTransform: func(obj interface{}) (interface{}, error) {
+				obj.(client.Object).SetManagedFields(nil)
```
skriss, line 245: I'm thinking we should check whether the type-assertion succeeds, and do _something_ if it doesn't -- probably just return the object as-is. Somewhat related to the above comment, where we might get a `cache.DeletedFinalStateUnknown`.

skriss (changes requested):

cmd/contour/serve.go
tsaarni: Thanks, I had missed this! 

I added check for the success of type assertion for both `TransformByObject` and `DefaultTransform`. I just pass the original object if it fails.

In case of `TransformByObject` the function is explicitly assigned for type `corev1.Secret` but I did not find anything that suggests that it could not get passed an `cache.DeletedFinalStateUnknown` as well, so it makes sense to add this. 

Though, I'm bit confused why the [`TransformFunc`](https://github.com/kubernetes/client-go/blob/8cbca742aebe24b24f7f4e32fd999942fa9133e8/tools/cache/controller.go#L356-L365) would be passed `cache.DeletedFinalStateUnknown` in any situation, if the tombstone object is associated to [`OnDelete`](https://github.com/kubernetes/client-go/blob/8cbca742aebe24b24f7f4e32fd999942fa9133e8/tools/cache/controller.go#L209-L212)? I thought transformation would happen when new objects are introduced into the cache, not when they are removed. Maybe the transformers are called also at delete 🤨.

cmd/contour/serve.go
tsaarni: Fixed

skriss (approved):
LGTM, thanks @tsaarni! My preference would be to get this PR merged for the upcoming 1.25 release, and then to follow up later with any refactoring needed once the new controller-runtime version comes out to enable this plus #5214.

sunjayBhatia (approved):

PR #5101, opened 2023-02-17
internal/provisioner: expose the metrics port for envoy

### LINKED ISSUE #5232: [provisioner] envoy should expose the metrics port.

When deploying Envoy with the contour provisioner, the generated pods expose a single port:

```yaml
    ports:
    - containerPort: 8080
      name: http
      protocol: TCP
```

It should also expose the metric port, as it is required to monitor the envoy pods using a `PodMonitor` or a `ServiceMonitor`.

```yaml
    ports:
    - containerPort: 8080
      name: http
      protocol: TCP
    - containerPort: 8002
      name: metrics
      protocol: TCP
```

The port number should be derived from `ContourDeployment.spec.runtimeSettings.envoy.metrics.port`

Deploy envoy using contour provisioner and gateway, and be able to monitor Envoy using a PodMonitor, which required that the monitoring port be exposed by the target pod.

**Environment:**

- Contour version: 1.24.2

izturn:
Expose the default/changed metric port for envoy

Closes #5232.


izturn:
ping @sunjayBhatia @skriss

Jean-Daniel:
Is the usage of --stats-port flag required ?
The last time I tried, I saw that contour was directly using the value from its `contourConfiguration` to configure the envoy metric listener. 

So I don't think there is a need to add an extra flag to the contour deployment.

internal/provisioner/controller/gateway.go
```diff
@@ -236,6 +236,20 @@ func (r *gatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			contourModel.Spec.RuntimeSettings.Envoy.Metrics.Port != 0 {
+
+			port := contourModel.Spec.RuntimeSettings.Envoy.Metrics.Port
+			contourModel.Spec.NetworkPublishing.Envoy.Ports = append(
+				contourModel.Spec.NetworkPublishing.Envoy.Ports,
+				model.Port{
+					Name:          "metrics",
+					ServicePort:   int32(port),
+					ContainerPort: int32(port),
+				})
```
skriss, line 251: This has the effect of exposing the metrics port on the Service, which we don't do when using the default port and I don't think we want to do for custom ports. Were you intentionally trying to expose it on the Service? If so, could explain the thinking further

izturn:
ok, i will upgrade it

izturn:
gentle ping  @skriss @sunjayBhatia

internal/provisioner/objects/deployment/deployment.go
```diff
@@ -105,6 +105,7 @@ func DesiredDeployment(contour *model.Contour, image string) *appsv1.Deployment
 		switch {
 		case port.Name == "http" && port.ContainerPort != objects.EnvoyInsecureContainerPort:
 			args = append(args, fmt.Sprintf("--envoy-service-http-port=%d", port.ContainerPort))
+
```
skriss, line 108: Let's revert the no-op changes to deployment.go and deployment_test.go, for clarity in the commit.

internal/provisioner/objects/deployment/deployment_test.go
```diff
@@ -152,7 +152,8 @@ func TestDesiredDeployment(t *testing.T) {
 		"key": "value",
 	}
 
-	// Use non-default container ports to test that --envoy-service-http(s)-port
+	// Use non-default container ports to test that
```
skriss, line 155: Let's revert the no-op changes to deployment.go and deployment_test.go, for clarity in the commit.

skriss (changes requested):
Couple nits, but otherwise looks reasonable 👍

skriss:
Please add a changelog file as well.

changelogs/unreleased/5101-izturn-small.md
```diff
@@ -0,0 +1 @@
+Expose the default/changed metric port for envoy
```
skriss, line 1:
```suggestion
Gateway provisioner: add a container port to the Envoy daemonset/deployment for the metrics port.
```

internal/provisioner/objects/dataplane/dataplane.go
```diff
@@ -136,14 +136,32 @@ func desiredContainers(contour *model.Contour, contourImage, envoyImage string)
 		ports = append(ports, p)
 	}
 
-	healthPort := 8002
+	var (
+		healthPort  = 8002
```
skriss, line 140: while we're here, it'd be nice to move this constant into `object.go` along with the other port numbers, similar to `EnvoyMetricsPort`.

skriss (changes requested):

skriss (approved):
LGTM, thanks @izturn

PR #5107, opened 2023-02-20
Update gatewayapi_processor.go

### LINKED ISSUE #5103: set condition of Accepted: false with reason UnsupportedValue for unknown HTTPRoute path match types
Currently, when an HTTPRoute has an unknown path match type (not `PathPrefix` or `Exact`), Contour sets a condition of "NotImplemented: true" with reason "PathMatchType". See:
- https://github.com/projectcontour/contour/blob/main/internal/dag/gatewayapi_processor.go#L1432

Per [the Gateway API spec](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.PathMatchType):

> Unknown values here must result in the implementation setting the Accepted Condition for the Route to status: False, with a Reason of UnsupportedValue.

Thanks to @fangfpeng for finding this!

It would be good to audit other uses of the custom `NotImplemented` condition and see if there are other places where we should make a similar switch, it looks like there probably are.

**Environment:**
- Contour version: main

nerdynikhil:
This PR fixes #5103

### DIFF
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -1429,7 +1429,7 @@ func gatewayPathMatchCondition(match *gatewayapi_v1beta1.HTTPPathMatch, routeAcc
 		return &ExactMatchCondition{Path: path}, true
 	}
 
-	routeAccessor.AddCondition(status.ConditionNotImplemented, metav1.ConditionTrue, status.ReasonPathMatchType, "HTTPRoute.Spec.Rules.PathMatch: Only Prefix match type and Exact match type are supported.")
+	routeAccessor.AddCondition(status.ConditionAccepted, metav1.ConditionFalse, status.ReasonUnsupportedValue, "Unsupported HTTPRoute path match type.")
 	return nil, false
 }
 
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -1429,7 +1429,7 @@ func gatewayPathMatchCondition(match *gatewayapi_v1beta1.HTTPPathMatch, routeAcc
 		return &ExactMatchCondition{Path: path}, true
 	}
 
-	routeAccessor.AddCondition(status.ConditionAccepted, metav1.ConditionFalse, status.ReasonUnsupportedValue, "Unsupported HTTPRoute path match type.")
+	routeAccessor.AddCondition(gatewayapi_v1beta1.ConditionAccepted, metav1.ConditionFalse, gatewayapi_v1beta1.ReasonUnsupportedValue, "Unsupported HTTPRoute path match type.")
 	return nil, false
 }
 


internal/dag/gatewayapi_processor.go
```diff
@@ -1429,7 +1429,7 @@ func gatewayPathMatchCondition(match *gatewayapi_v1beta1.HTTPPathMatch, routeAcc
 		return &ExactMatchCondition{Path: path}, true
 	}
 
-	routeAccessor.AddCondition(status.ConditionNotImplemented, metav1.ConditionTrue, status.ReasonPathMatchType, "HTTPRoute.Spec.Rules.PathMatch: Only Prefix match type and Exact match type are supported.")
+	routeAccessor.AddCondition(status.ConditionAccepted, metav1.ConditionFalse, status.ReasonUnsupportedValue, "Unsupported HTTPRoute path match type.")
```
skriss, line 1432:
```suggestion
	routeAccessor.AddCondition(gatewayapi_v1beta1.ConditionAccepted, metav1.ConditionFalse, gatewayapi_v1beta1.ReasonUnsupportedValue, "Unsupported HTTPRoute path match type.")
```

internal/dag/gatewayapi_processor.go
```diff
@@ -1429,7 +1429,7 @@ func gatewayPathMatchCondition(match *gatewayapi_v1beta1.HTTPPathMatch, routeAcc
 		return &ExactMatchCondition{Path: path}, true
 	}
 
-	routeAccessor.AddCondition(status.ConditionNotImplemented, metav1.ConditionTrue, status.ReasonPathMatchType, "HTTPRoute.Spec.Rules.PathMatch: Only Prefix match type and Exact match type are supported.")
+	routeAccessor.AddCondition(gatewayapi_v1beta1.ConditionAccepted, metav1.ConditionFalse, gatewayapi_v1beta1.ReasonUnsupportedValue, "Unsupported HTTPRoute path match type.")
```
skriss, line 1432: Looks like I gave you the wrong names here in my initial suggestion, it should be:

```suggestion
	routeAccessor.AddCondition(gatewayapi_v1beta1.RouteConditionAccepted, metav1.ConditionFalse, gatewayapi_v1beta1.RouteReasonUnsupportedValue, "Unsupported HTTPRoute path match type.")
```

https://github.com/projectcontour/contour/blob/main/CONTRIBUTING.md will walk you through how to ensure the updated code builds locally and passes all unit tests. I'm guessing there is probably a test case or two to update here.

skriss: Please also leave the Condition message as it was previously - `HTTPRoute.Spec.Rules.PathMatch: Only Prefix match type and Exact match type are supported.`

skriss:
I ended up tackling this more holistically in #5125 so I'm going to close out this PR. Thanks for the input regardless @nerdynikhil.

PR #5109, opened 2023-02-20
add gauge metric for DAG cache items

### LINKED ISSUE #5005: add metrics for DAG cache contents
It'd be nice to have ~~counter~~ gauge metrics indicating how many items are in the DAG cache (by kind at least). This would be helpful for ensuring we aren't holding onto resources we shouldn't be.

Suggested structure would be a single GaugeVec metric, named `contour_dag_cache_object_count`, with at least one label, `kind`, and potentially another for `namespace`. This metric should be set each time an object is added to or removed from the DAG cache, based on the current length of the relevant map.

izturn:

fix #5005 

BTW: 
 the prometheus complain with:
 > contour_dag_cache_object_count: non-histogram and non-summary metrics should not have "_count" suffix, 

so remove the suffix: `_count`


changelogs/unreleased/5109-izturn-small.md
```diff
@@ -0,0 +1 @@
+Adds a new gauge metric, `contour_dag_cache_object`, to indicate the total number of items are in the DAG cache at current.
```
skriss, line 1:
```suggestion
Adds a new gauge metric, `contour_dag_cache_object`, to indicate the total number of items that are currently in the DAG cache.
```

internal/metrics/metrics.go
```diff
@@ -34,6 +34,7 @@ type Metrics struct {
 	proxyOrphanedGauge  *prometheus.GaugeVec
 
 	dagRebuildGauge             prometheus.Gauge
+	dagCacheOjbectGauge         *prometheus.GaugeVec
```
skriss, line 37:
```suggestion
	dagCacheObjectGauge         *prometheus.GaugeVec
```

internal/metrics/metrics.go
```diff
@@ -66,6 +67,7 @@ const (
 	HTTPProxyValidGauge     = "contour_httpproxy_valid"
 	HTTPProxyOrphanedGauge  = "contour_httpproxy_orphaned"
 
+	DAGCacheOjbectGauge         = "contour_dag_cache_object"
```
skriss, line 70:
```suggestion
	DAGCacheObjectGauge         = "contour_dag_cache_object"
```

internal/metrics/metrics.go
```diff
@@ -222,6 +233,14 @@ func (m *Metrics) SetDAGRebuiltTotal() {
 	m.dagRebuildTotal.Inc()
 }
 
+// SetDAGCacheOjbectMetric records the total number of items in cache at current
+func (m *Metrics) SetDAGCacheOjbectMetric(kind string, count int) {
```
skriss, line 237:
```suggestion
func (m *Metrics) SetDAGCacheObjectMetric(kind string, count int) {
```

internal/metrics/metrics_test.go
```diff
@@ -690,3 +691,60 @@ func TestRemoveProxyMetric(t *testing.T) {
 		})
 	}
 }
+
+func TestSetDAGCacheOjbectMetric(t *testing.T) {
```
skriss, line 695:
```suggestion
func TestSetDAGCacheObjectMetric(t *testing.T) {
```

internal/metrics/metrics.go
```diff
@@ -130,6 +132,13 @@ func NewMetrics(registry *prometheus.Registry) *Metrics {
 				Help: "Timestamp of the last DAG rebuild.",
 			},
 		),
+		dagCacheOjbectGauge: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: DAGCacheOjbectGauge,
+				Help: "Total number of items are in the DAG cache at current.",
```
skriss, line 138:
```suggestion
				Help: "Total number of items that are currently in the DAG cache.",
```

skriss (changes requested):
Thanks for the PR @izturn! A few small things to address at first.

internal/dag/cache.go
```diff
@@ -179,39 +191,47 @@ func (kc *KubernetesCache) Insert(obj interface{}) bool {
...
 	}
 
-	if maybeInsert(obj) {
+	if ok, count := maybeInsert(obj); ok {
 		// Only check annotations if we actually inserted
 		// the object in our cache; uninteresting objects
 		// should not be checked.
 		if obj, ok := obj.(metav1.Object); ok {
 			kind := k8s.KindOf(obj)
+			kc.Metrics.SetDAGCacheObjectMetric(kind, count)
```
skriss, line 233: We need to set this regardless of whether `ok` was true or not, since `ok` represents whether the DAG needs to be rebuilt or not, _not_ whether something was added to the cache or not.

internal/dag/cache.go
```diff
@@ -244,89 +264,104 @@ func (kc *KubernetesCache) Remove(obj interface{}) bool {
 
 	switch obj := obj.(type) {
 	default:
-		return kc.remove(obj)
+		ok, count := kc.remove(obj)
+		if ok {
+			kc.Metrics.SetDAGCacheObjectMetric(k8s.KindOf(obj), count)
```
skriss, line 269: same here, need to set the metric whether `ok` was true or not.

internal/dag/cache.go
```diff
@@ -122,11 +130,12 @@ func (kc *KubernetesCache) Insert(obj interface{}) bool {
 					WithField("ingress-class-name", ref.Val(obj.Spec.IngressClassName, "")).
 					WithField("target-ingress-classes", kc.IngressClassNames).
 					Debug("ignoring Ingress with unmatched ingress class")
-				return false
+				return false, 0
```
skriss, line 133:
```suggestion
				return false, len(kc.ingresses)
```

internal/dag/cache.go
```diff
@@ -137,37 +146,40 @@ func (kc *KubernetesCache) Insert(obj interface{}) bool {
 					WithField("ingress-class-name", obj.Spec.IngressClassName).
 					WithField("target-ingress-classes", kc.IngressClassNames).
 					Debug("ignoring HTTPProxy with unmatched ingress class")
-				return false
+				return false, 0
```
skriss, line 149:
```suggestion
				return false, len(kc.httpproxies)
```

internal/dag/cache.go
```diff
@@ -137,37 +146,40 @@ func (kc *KubernetesCache) Insert(obj interface{}) bool {
...
+			return true, len(kc.tlscertificatedelegations)
+
 		case *gatewayapi_v1beta1.GatewayClass:
 			switch {
 			// Specific gateway configured: make sure the incoming gateway class
 			// matches that gateway's.
 			case kc.ConfiguredGatewayToCache != nil:
 				if kc.gateway == nil || obj.Name != string(kc.gateway.Spec.GatewayClassName) {
-					return false
+					return false, 0
```
skriss, line 165: The returned value here should be 1 if `kc.gatewayclass` is not nil

internal/dag/cache.go
```diff
@@ -137,37 +146,40 @@ func (kc *KubernetesCache) Insert(obj interface{}) bool {
...
 			}
+
 		case *gatewayapi_v1beta1.Gateway:
 			switch {
 			// Specific gateway configured: make sure the incoming gateway
 			// matches, and get its gateway class.
 			case kc.ConfiguredGatewayToCache != nil:
 				if k8s.NamespacedNameOf(obj) != *kc.ConfiguredGatewayToCache {
-					return false
+					return false, 0
```
skriss, line 182: The returned value here should be 1 if `kc.gateway` is not nil

internal/dag/cache.go
```diff
@@ -244,89 +264,104 @@ func (kc *KubernetesCache) Remove(obj interface{}) bool {
...
 		delete(kc.tlscertificatedelegations, m)
-		return ok
+		return ok, len(kc.tlscertificatedelegations)
+
 	case *gatewayapi_v1beta1.GatewayClass:
 		switch {
 		case kc.ConfiguredGatewayToCache != nil:
 			if kc.gatewayclass == nil || obj.Name != kc.gatewayclass.Name {
-				return false
+				return false, 0
```
skriss, line 316: 1 if kc.gatewayclass is not nil, otherwise 0

internal/dag/cache.go
```diff
@@ -244,89 +264,104 @@ func (kc *KubernetesCache) Remove(obj interface{}) bool {
...
-			return true
+			return true, 0
 		}
+
 	case *gatewayapi_v1beta1.Gateway:
 		switch {
 		case kc.ConfiguredGatewayToCache != nil:
 			if kc.gateway == nil || k8s.NamespacedNameOf(obj) != k8s.NamespacedNameOf(kc.gateway) {
-				return false
+				return false, 0
```
skriss, line 329: 1 if `kc.gateway` is not nil, otherwise 0

internal/dag/cache.go
```diff
@@ -179,39 +191,47 @@ func (kc *KubernetesCache) Insert(obj interface{}) bool {
...
 		}
 	}
 
-	if maybeInsert(obj) {
+	if ok, count := maybeInsert(obj); ok {
 		// Only check annotations if we actually inserted
 		// the object in our cache; uninteresting objects
 		// should not be checked.
 		if obj, ok := obj.(metav1.Object); ok {
 			kind := k8s.KindOf(obj)
```
skriss, line 232: Please double-check that `KindOf` actually covers all the resource types we cache -- IIRC it was missing some (ref. #5012)

skriss (changes requested):

internal/dag/cache.go
izturn: yes, the `KindOf` not covers the all resource types now, but I don't know whether I should update it or leave it to #5012

skriss: OK let's get #5012 merged and then we'll be good here.

skriss:
@izturn this needs a rebase as we added GRPCRoute support, please make sure we have a metric for that in the DAG cache as well!

izturn:
@skriss PTAL

skriss (approved):
Tested this out and LGTM, thanks @izturn!

skriss:
Sorry, I should've held my #5133 until this merged, but now a minor merge conflict to resolve (we no longer are informing on ContourConfiguration, so don't need the logic for it in the DAG cache)

internal/dag/cache.go
```diff
@@ -181,42 +198,51 @@ func (kc *KubernetesCache) Insert(obj interface{}) bool {
...
 			// not an interesting object
 			kc.WithField("object", obj).Error("insert unknown object")
-			return false
+			return false, 0
 		}
 	}
 
-	if maybeInsert(obj) {
+	ok, count := maybeInsert(obj)
+	kind := k8s.KindOf(obj)
```
sunjayBhatia, line 239: if we keep it with this kind of implementation, what about instead of finding the count of resources in `maybeInsert`, we just take the kind and take the length of the map of the appropriate resource here (or even just iterate over all the kinds and set them all)

finding the number of resources to set the metric to above seems like it's easy to make a small mistake that is hard to catch, esp., in cases with special handling of resources that we don't necessarily want to add etc.

it would be good to add some testing to make sure we don't make any such mistakes

sunjayBhatia (commented):
Apologies, I'm a little late on this one but this seems like an opportunity to pull this out into it's own component/abstraction so we can test things and so this functionality doesn't interfere with the dag cache directly and we don't need to change the dag itself to get this metric

e.g. in other places we use the `ResourceEventHandler` or `dag.Observer` interface to compose components together

we could have an `Observer` or other abstraction that runs after the cache is updated to take stock of what is in the cache and send metrics

see https://github.com/projectcontour/contour/blob/main/internal/contour/metrics.go

internal/dag/cache.go
izturn: @sunjayBhatia  thx,  @skriss WDYT? I will resend a new pr or update this one after we make a decision

sunjayBhatia: chatting with @skriss offline, I'm revising my position and in support of this impl:
- with my suggestion if any new resource kinds are added to the cache you will have to remember to update the component that sets the cache metrics, this impl forces you to b/c of the signature of `maybeInsert` etc.
- in the future we plan to attempt to utilize go generics more which should give us opportunities to refactor this

sunjayBhatia: when I get some time I'm planning on experimenting with making KubernetesCache use generics to simplify things and reduce type assertions/switches etc.

sunjayBhatia (approved):

PR #5116, opened 2023-02-22
internal/dag: fix method match of GRPCRoute

fangfpeng:
- need to add a leading slash in the path
- need to add a prefix match if method match is not specified.

Updates #4820

### DIFF
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -1432,6 +1432,13 @@ func (p *GatewayAPIProcessor) computeGRPCRoute(route *gatewayapi_v1alpha2.GRPCRo
 			})
 		}
 
+		// If no matches are specified, the implementation MUST match every gRPC request.
+		if rule.Matches == nil {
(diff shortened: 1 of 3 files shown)


skriss (approved):
LGTM, is this OK to merge? (it's marked as a draft right now)

fangfpeng:
> LGTM, is this OK to merge? (it's marked as a draft right now)

Not yet, I will add a patch for the 0 matches case soon.

internal/dag/gatewayapi_processor.go
```diff
@@ -1432,6 +1432,13 @@ func (p *GatewayAPIProcessor) computeGRPCRoute(route *gatewayapi_v1alpha2.GRPCRo
 			})
 		}
 
+		// If no matches are specified, the implementation MUST match every gRPC request.
+		if rule.Matches == nil {
```
skriss, line 1436: Small nit, this will handle both a nil and an empty slice:
```suggestion
		if len(rule.Matches) == 0 {
```

skriss (approved):
LGTM!

PR #5117, opened 2023-02-22
changelogs: add changelog

fangfpeng:
Also made some changes based on review comment
on gatewayapi_processor.go.


changelogs/unreleased/5114-fangfpeng-small.md
```diff
@@ -0,0 +1 @@
+Gateway API: support the GRPCRoute.
```
skriss, line 1: Let's bump this up to a minor, and include a bit more information (this will appear in the release notes) -- a link to https://gateway-api.sigs.k8s.io/api-types/grpcroute/ would be helpful, and maybe also mention that the "core conformance" parts of the spec are implemented. It probably also makes sense to include a link to https://projectcontour.io/docs/v1.25.0/guides/grpc/#gateway-api-configuration (doesn't exist yet but will be created as a copy of https://projectcontour.io/docs/main/guides/grpc/#gateway-api-configuration when we release) since we'll want to update that guide to show how to use GRPCRoute.

skriss: See e.g. https://github.com/projectcontour/contour/blob/main/changelogs/unreleased/5036-relu-minor.md for an example of another minor changelog.

skriss (approved):

PR #5114, opened 2023-02-21
Gateway API: add GRPCRoute support

fangfpeng:
Closes #4820.

PR #5122, opened 2023-02-24
Gateway API: update grpc.md for GRPCRoute usage

vmw-yingy:
Adding a section for GRPCRoute configuration

Updates #4820


site/content/docs/main/guides/grpc.md
```diff
@@ -144,10 +144,43 @@ spec:
 
 ## Gateway API Configuration
 
-At the moment, configuring gRPC routes with Gateway API resources is achieved by the same method as with Ingress v1: annotation to select a protocol and port on a Service referenced by HTTPRoute `spec.rules[].backendRefs`.
+Gateway API now supports a specific resource [GRPCRoute][6] for routing gRPC requests.
 
-Gateway API does include a specific resource [GRPCRoute][6] for routing gRPC requests.
-This may be supported in future versions of Contour.
+## GRPCRoute Configuration
```
skriss, line 149: I think we can drop this section header, this content can all go directly under "Gateway API Configuration"

site/content/docs/main/guides/grpc.md
```diff
@@ -144,10 +144,43 @@ spec:
...
+      port: 9000
+```
+Using the sample deployment above along with this GRPCRoute example, you can test calling this plaintext gRPC server with the same grpccurl command:
+
+```yaml
+grpcurl -plaintext -authority=my-grpc-service.foo.com <load balancer IP and port if needed> yages.Echo/Ping
+```
+Note that the second matching method for service of ServerReflection is required by grpcurl command.
+
+In addition to using the GRPCRoute resource, configuring gRPC routes with Gateway API resources can also be achieved by the same method as with Ingress v1: annotation to select a protocol and port on a Service referenced by HTTPRoute `spec.rules[].backendRefs`.
```
skriss, line 183: I don't think it's necessary to retain this content about how to use HTTPRoute to route to gRPC services, since GRPCRoute is the preferred way.

However, I would replace it with a sentence or two indicating that the user _should_ annotate their Service similarly to when using Ingress, to indicate the protocol to use when connecting to the backend Service (h2 or h2c), and also that if it's not specified, Contour will infer the protocol based on the Gateway Listener protocol (HTTP -> h2c, HTTPS -> h2)

skriss (changes requested):

sunjayBhatia (commented):
looks good modulo Steve's comments!

skriss:
Since we've got approvals on #5114 I'm going to merge that one now, and retarget this against main.

vmw-yingy:
> Since we've got approvals on #5114 I'm going to merge that one now, and retarget this against main.

OK, I will update the grpc.md on the main branch.

skriss:
Replaced by #5126, closing.

PR #5124, opened 2023-02-27
internal/dag: merge common route processing code

skriss:
Extracts a helper method `processRoute` that can
be used for any route type and implements common
logic around hostnames, parent attachment, etc.


sunjayBhatia (approved):
nice! diff is a little hard to read of course but looking through the file itself looks like a great change 👍🏽

skriss:
> nice! diff is a little hard to read of course but looking through the file itself looks like a great change 👍🏽

Yeah, I figured all the related testing served as a decent backstop to the diff being challenging

PR #5132, opened 2023-03-01
Use constants 'http.MethodGet' instead of string 'GET'

yulng:
Use package constants instead of strings
I hope the above suggestions can help contribute to the community
Thank you


skriss (approved):
Thanks @yulng, LGTM

PR #5133, opened 2023-03-01
remove informer for ContourConfiguration

skriss:
ContourConfiguration is not being cached
in the DAG cache so the informer is not
needed.

Also adds ContourConfiguration and
ContourDeployment to the KindOf function
since they are known types.

I noticed this while testing #5109 because I had a ContourConfiguration in my cluster, and it was showing up in the metric output as an unknown object.


sunjayBhatia (approved):

sunjayBhatia:
we can always bring this back if we need to do dynamic config updates etc., but as of now we are not planning on it

PR #5151, opened 2023-03-06
Update team.html

pnbrown:
Move Steve to Tech Lead and Nick Young to Emeritus Maintainer.

### DIFF
--- a/changelogs/unreleased/5151-pnbrown-docs.md
+++ b/changelogs/unreleased/5151-pnbrown-docs.md
@@ -1,0 +1,1 @@
+Updates Steve Kriss as Tech Lead and moves Nick Young to Emeritus Maintainer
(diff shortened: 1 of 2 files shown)


tsaarni (approved):
Looks good to me! 

One question: I assume it was agreed that Nick is not removed from the web site? I ask, because there have been cases before where emeritus maintainers were "moved" to this file from the web page https://github.com/projectcontour/community/blob/main/MAINTAINERS.md

sunjayBhatia (commented):
yeah @tsaarni has a point about keeping the front page for active maintainers

fine with either way, but if we want to keep emeritus maintainers here we probably should bring back others, probably easier to leave the emeritus maintainers in the community repo

skriss:
+1, I think we can drop Nick here for consistency since he's still listed in the community repo.

pnbrown:
Understood. That change is made.

skriss (approved):

PR #5155, opened 2023-03-09
internal/dag: finish the TODO for `EnsureService`

izturn:


internal/dag/ingress_processor.go
```diff
@@ -154,13 +154,11 @@ func (p *IngressProcessor) computeIngressRule(ing *networking_v1.Ingress, rule n
 		be := httppath.Backend
 		m := types.NamespacedName{Name: be.Service.Name, Namespace: ing.Namespace}
 
-		var port intstr.IntOrString
+		port := int(be.Service.Port.Number)
 		if len(be.Service.Port.Name) > 0 {
-			port = intstr.FromString(be.Service.Port.Name)
-		} else {
-			port = intstr.FromInt(int(be.Service.Port.Number))
+			_, svcPort, _ := p.source.LookupService(m, intstr.FromString(be.Service.Port.Name))
```
skriss, line 159: we need to handle the error here

izturn: If there is an error, the `svcPort`'s Port will be 0, it will lead to `EnsureService` returns a error, so  ignore it here

izturn: ok

PR #5156, opened 2023-03-09
internal/dag: simplify ut for dag/status

izturn:


internal/dag/gatewayapi_processor.go
```diff
@@ -1412,19 +1412,20 @@ func (p *GatewayAPIProcessor) validateBackendRef(backendRef gatewayapi_v1beta1.B
 	return p.validateBackendObjectRef(backendRef.BackendObjectReference, "Spec.Rules.BackendRef", routeKind, routeNamespace)
 }
 
+func resolvedRefsFalse(reason gatewayapi_v1beta1.RouteConditionReason, msg string) *metav1.Condition {
```
skriss, line 1415: I'd return a non-pointer here since we're dereferencing the result everywhere:
```suggestion
func resolvedRefsFalse(reason gatewayapi_v1beta1.RouteConditionReason, msg string) metav1.Condition {
```

izturn: The `resolvedRefsFalse`  extract from `validateBackendObjectRef`, so both the ut & the method can share it; if we need to create a new one for ut?

izturn:
ping @skriss

internal/dag/gatewayapi_processor.go
skriss: Let's do this:
- return a `metav1.Condition` from this function so it's consistent with the other condition helpers that are used in status_test.go
- where used in `gatewayapi_processor.go`, wrap in a `ref.To(...)`

izturn: done

izturn:
@skriss  PTAL

skriss (approved):
LGTM, thanks @izturn

PR #5160, opened 2023-03-09
support many Gateway Listeners

### LINKED ISSUE #4960: Support an arbitrary number of Listeners per Gateway

skriss:
Adds support for programming an arbitrary number
of Gateway listeners in Envoy and the Envoy service.

Closes #4960.

### DIFF
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -14161,7 +14161,7 @@ func TestGatewayWithHTTPProxyAndIngress(t *testing.T) {
 
 	tests := map[string]struct {
 		gateway *gatewayapi_v1beta1.Gateway
-		objs    []interface{}
+		objs    []any
 		want    []*Listener
 	}{
 		"HTTPProxy attached to HTTP-only Gateway": {
@@ -14186,7 +14186,7 @@ func TestGatewayWithHTTPProxyAndIngress(t *testing.T) {
 					},
 				},
 			},
-			objs: []interface{}{
+			objs: []any{
 				kuardService,
 				&contour_api_v1.HTTPProxy{
 					ObjectMeta: metav1.ObjectMeta{
@@ -14256,7 +14256,7 @@ func TestGatewayWithHTTPProxyAndIngress(t *testing.T) {
 					},
 				},
 			},
-			objs: []interface{}{
+			objs: []any{
 				kuardService,
 				&contour_api_v1.HTTPProxy{
 					ObjectMeta: metav1.ObjectMeta{
@@ -14322,7 +14322,7 @@ func TestGatewayWithHTTPProxyAndIngress(t *testing.T) {
 					},
 				},
 			},
-			objs: []interface{}{
+			objs: []any{
 				kuardService,
 				&contour_api_v1.HTTPProxy{
 					ObjectMeta: metav1.ObjectMeta{
@@ -14395,7 +14395,7 @@ func TestGatewayWithHTTPProxyAndIngress(t *testing.T) {
 					},
 				},
 			},
-			objs: []interface{}{
+			objs: []any{
 				kuardService,
 				sec1,
 				&contour_api_v1.HTTPProxy{
@@ -14475,7 +14475,7 @@ func TestGatewayWithHTTPProxyAndIngress(t *testing.T) {
 					},
 				},
 			},
-			objs: []interface{}{
+			objs: []any{
 				kuardService,
 				sec1,
 				&contour_api_v1.HTTPProxy{
@@ -14545,7 +14545,7 @@ func TestGatewayWithHTTPProxyAndIngress(t *testing.T) {
 					},
 				},
 			},
-			objs: []interface{}{
+			objs: []any{
 				kuardService,
 				sec1,
 				&contour_api_v1.HTTPProxy{
@@ -14603,7 +14603,7 @@ func TestGatewayWithHTTPProxyAndIngress(t *testing.T) {
 					},
 				},
 			},
-			objs: []interface{}{
+			objs: []any{
 				kuardService,
 				&networking_v1.Ingress{
 					ObjectMeta: metav1.ObjectMeta{
@@ -14669,7 +14669,7 @@ func TestGatewayWithHTTPProxyAndIngress(t *testing.T) {
 					},
 				},
 			},
-			objs: []interface{}{
+			objs: []any{
 				kuardService,
 				&networking_v1.Ingress{
 					ObjectMeta: metav1.ObjectMeta{
@@ -14731,7 +14731,7 @@ func TestGatewayWithHTTPProxyAndIngress(t *testing.T) {
 					},
 				},
 			},
-			objs: []interface{}{
+			objs: []any{
 				kuardService,
 				&networking_v1.Ingress{
 					ObjectMeta: metav1.ObjectMeta{
@@ -14800,7 +14800,7 @@ func TestGatewayWithHTTPProxyAndIngress(t *testing.T) {
 					},
 				},
 			},
-			objs: []interface{}{
+			objs: []any{
 				kuardService,
 				sec1,
 				&networking_v1.Ingress{
@@ -14879,7 +14879,7 @@ func TestGatewayWithHTTPProxyAndIngress(t *testing.T) {
 					},
 				},
 			},
-			objs: []interface{}{
+			objs: []any{
 				kuardService,
 				sec1,
 				&networking_v1.Ingress{
@@ -14948,7 +14948,7 @@ func TestGatewayWithHTTPProxyAndIngress(t *testing.T) {
 					},
 				},
 			},
-			objs: []interface{}{
+			objs: []any{
 				kuardService,
 				sec1,
 				&networking_v1.Ingress{
--- a/internal/dag/builder_test.go
+++ b/internal/dag/builder_test.go
@@ -1988,9 +1988,10 @@ func TestDAGInsertGatewayAPI(t *testing.T) {
 			},
 			want: listeners(
 				&Listener{
-					Name:    "http-443",
-					Address: "0.0.0.0",
-					Port:    64955,
+					Name:     "http-443",
+					Address:  "0.0.0.0",
+					Port:     64955,
+					Protocol: "HTTP",
 					VirtualHosts: virtualhosts(
 						virtualhost("test.projectcontour.io",
 							prefixrouteHTTPRoute("/", service(kuardService)),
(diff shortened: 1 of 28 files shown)


internal/provisioner/objects/dataplane/dataplane.go
```diff
@@ -212,7 +202,6 @@ func desiredContainers(contour *model.Contour, contourImage, envoyImage string)
 				SuccessThreshold:    int32(1),
 				TimeoutSeconds:      int32(1),
 			},
-			Ports: ports,
```
skriss, line 215: In testing this, I realized that every time a new Gateway Listener with a new port is added, it triggers a rollout of Envoy because the container ports change. Here, I've stopped setting container ports on the Envoy workload entirely, since it's not required in order to expose the port anyway and mostly serves as documentation. Definitely interested in folks' thoughts here.

internal/gatewayapi/listeners.go
```diff
@@ -18,144 +18,174 @@ import (
...
+		}
+
+		if !found {
+			// Map privileged ports (1-1023) to the range 64513-65535 within
+			// the container. All other ports can be used as-is inside the
+			// container.
+			containerPort := listener.Port
+			if containerPort < 1024 {
+				containerPort += 64512
+			}
```
skriss, line 178: I went with the original port mapping scheme from the design doc here, but could still change this to something else.

skriss:
Still need to look at the upgrade path here; since the ports used internally are changing, the Envoy service needs to be updated as part of the upgrade and I'm guessing there will be some expected downtime.

skriss:
This is now passing the upstream dynamic listener ports conformance PR with no modifications thanks to @sunjayBhatia's work to upgrade to v0.6.2.

internal/provisioner/objects/dataplane/dataplane.go
skriss: @izturn @wilsonwu @Jean-Daniel I'd be interested in any input you have here. Let me know if you need more context!

internal/gatewayapi/listeners.go
skriss: @izturn @wilsonwu @Jean-Daniel same here - let me know if you have any input on the port mapping scheme being proposed here

skriss: The other option I'm considering is adding 8000 to the Listener's port to determine the port Envoy should listen on (80 -> 8080, 443 -> 8443, etc), wrapping back around to the beginning of the port range for very high ports and skipping ports 1-1023 (since Envoy can't bind to them by default). The nice thing about this is that the standard ports continue to map to the same container ports, which should help avoid some upgrade issues for the normal case.

skriss:
I've been doing some upgrade testing to test out the different listener port mapping schemes, but because of https://github.com/projectcontour/contour/issues/5375, I think that actually there's currently some downtime for any provisioner-driven upgrade. I'll see if I can get at least a spike of a fix for #5375 implemented, which then should allow for a better comparison between the port mapping schemes.

skriss:
Rebased on main to pick up provisioner upgrade test, I believe it'll fail because with this port mapping scheme which results in changes for how ports 80/443 are mapped, there is some downtime.

internal/gatewayapi/listeners.go
skriss: I've now switched to [the scheme described above](https://github.com/projectcontour/contour/pull/5160#discussion_r1191702512) in https://github.com/projectcontour/contour/pull/5160/commits/2d65eb99eb3df2702f06511e87e13323c508037e, as it's more backwards-compatible and results in less downtime during upgrades when using the standard ports.

Jean-Daniel: I find it more understandable and cleaner, as standard use cases will have well known values and the limitation (58559-59581 and 57536-58558 are overlapping) would probably almost never have an impact on real use cases.

internal/gatewayapi/listeners.go
```diff
@@ -18,144 +18,190 @@ import (
...
+				Status:  metav1.ConditionFalse,
+				Reason:  string(gatewayapi_v1beta1.ListenerReasonUnsupportedProtocol),
+				Message: fmt.Sprintf("Listener protocol %q is unsupported, must be one of HTTP, HTTPS or TLS", listener.Protocol),
+			}
+			continue
+		}
+
+		// Check for conflicts with other listeners.
+		conflicted := false
+		for j, otherListener := range listeners {
```
sunjayBhatia, line 94: i think re-looping around all listeners will have us comparing against potentially invalid elements that we could probably ignore and doing a sort of two-way comparison when there are conflicts when we only need to do it once? wdyt about saving the valid listeners (indexed maybe with `envoyListenerName` from below since that ends up capturing protocol/port) in a map and then each listener contains a list of unique hostnames assigned to it, e.g. a structure like

```golang
type ValidateListenersResult struct {
  ListenerNames map[string]string

  // maps name (protocol-port, e.g. http-80 to port details)
  // could maybe just use the Gateway Listener port number
  Ports map[string]ListenerPort

  InvalidListenerConditions map[gatewayapi_v1beta1.SectionName]metav1.Condition
}

type ListenerPort struct {
  port int32
  containerPort int32
  protocol string
  hostnames map[string]struct{}
}
```

this way once a Gateway Listener is completely valid and unique, it is added to the map, and subsequent Listeners are compared against the good ones only, with list order precedence (which is how Routes solve conflicts, not explicitly specified anywhere on Gateways but good for consistency?)

sunjayBhatia (commented):
🤩 

still digesting, some comments so far!

internal/dag/httpproxy_processor.go
```diff
@@ -228,7 +228,13 @@ func (p *HTTPProxyProcessor) computeHTTPProxy(proxy *contour_api_v1.HTTPProxy) {
 				return
 			}
 
-			svhost := p.dag.EnsureSecureVirtualHost(HTTPS_LISTENER_NAME, host)
+			listener, err := p.dag.GetSingleListener("HTTPS")
```
skriss, line 231: Note to self: we talked about using a custom Listener protocol for this case, e.g. `projectcontour.io/https` or something along those lines, to indicate the TLS details will be contained in an HTTPProxy or Ingress rather than the Listener itself. This is probably more intuitive than using protocol = HTTPS which then requires mode: passthrough in order to not specify certificate details in the Listener. ref. https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1beta1/gateway_types.go#L247-L251

sunjayBhatia: related to migration/integration I had and wanted to maybe spike on seeing how complicated it would be to be able to reference HTTPProxies/Ingresses from a Gateway Listener, could be made to be compatible with the method for specifying a custom protocol you mention in this thread and a migration step for users as they move to Gateway API

sunjayBhatia: would make integrating with multiple listener ports a little easier in theory

skriss: OK, with https://github.com/projectcontour/contour/pull/5160/commits/2b7f149edf9239f43c1bff64a37ad1a0f8a32ebc, the following Gateway spec can be used with HTTPProxies/Ingresses that terminate TLS:

```yaml
kind: Gateway
metadata:
  name: contour
  namespace: projectcontour
spec:
  gatewayClassName: contour
  listeners:
    - name: http
      port: 80
      protocol: HTTP
      allowedRoutes:
        namespaces:
          from: All
    - name: https
      port: 443
      protocol: projectcontour.io/https # using this protocol means `tls` does not need to be specified for this Listener
      allowedRoutes:
        namespaces:
          from: All
```

internal/dag/ingress_processor.go
```diff
@@ -104,7 +104,16 @@ func (p *IngressProcessor) computeSecureVirtualhosts() {
 			// ahead and create the SecureVirtualHost for this
 			// Ingress.
 			for _, host := range tls.Hosts {
-				svhost := p.dag.EnsureSecureVirtualHost(HTTPS_LISTENER_NAME, host)
+				listener, err := p.dag.GetSingleListener("HTTPS")
+				if err != nil {
+					p.WithError(err).
+						WithField("name", ing.GetName()).
+						WithField("namespace", ing.GetNamespace()).
+						Errorf("listener not found")
```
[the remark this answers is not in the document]
skriss: Addressed in https://github.com/projectcontour/contour/pull/5160/commits/c515e9ff0dd11bd4e05b016b8efcf8a11aefb314

internal/dag/status_test.go
```diff
@@ -4556,6 +4556,167 @@ func TestDAGStatus(t *testing.T) {
 				),
 		},
 	})
+
+	run(t, "HTTPProxy cannot attach to a Gateway with >1 HTTP Listener", testcase{
```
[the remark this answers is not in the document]
skriss: :+1:, I also need to add some persistent documentation for how to use Gateway/the provisioner with HTTPProxy or Ingress that covers this as well.

internal/gatewayapi/listeners.go
skriss: I think that's reasonable, will work on refactoring.

skriss: Started working on this and it seemed easiest to just change the inner loop to only compare a given Listener to other Listeners before it in list order, which has the same logical effect and requires less code shuffling 😁  I also cleaned up a few other things to help readability. LMK if this works for you! https://github.com/projectcontour/contour/pull/5160/commits/6ab84ac2655d6377e2cf9f50a95779717adf2f3a

sunjayBhatia (commented):
did some basic manual testing (by writing some quick provisioner e2e tests) on top of gw conformance tests and looks like things are working well

as we progress i think we can have a follow up to add testing (unit, e2e where applicable) for the mixed modes described here and the docs PR (GW provisioner w/ Ingress/HTTPProxy) to flush out any integration issues (so far things seem to be working great 💯 )

lmk when you want a final detailed pass, but overall looks great 👍🏽

internal/dag/status_test.go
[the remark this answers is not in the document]
skriss: Updated changelog in https://github.com/projectcontour/contour/pull/5160/commits/8aa1bf0b0d223cad02a7214503ab5c6f8ff61e4f.

skriss:
> as we progress i think we can have a follow up to add testing (unit, e2e where applicable) for the mixed modes described here and the docs PR (GW provisioner w/ Ingress/HTTPProxy) to flush out any integration issues (so far things seem to be working great 💯 )

+1, I have at least manually tested config like this: https://gist.github.com/skriss/1f3745f8dabb63035588fabe5795a164 (basically one vhost with TLS using HTTPProxy, one vhost with TLS using Gateway/HTTPRoute)

Would be good to identify what other use cases we want to support, for both mixed mode (HTTPProxy and Gateway/HTTPRoute co-existing) and migration purposes.

Rycieos:
@skriss this is looking great! Sorry I was never able to give feedback earlier, but I can test now. I'm assuming you are still using the same image tag, so I'll try testing with your [steveheptio/contour:many-listeners](https://hub.docker.com/layers/steveheptio/contour/many-listeners/images/sha256-1ffec68ee6cc214e7b28ef44b8a75fc16a36d6fcb64ba94957baaa283ca924bf?context=explore) image.

skriss:
> @skriss this is looking great! Sorry I was never able to give feedback earlier, but I can test now. I'm assuming you are still using the same image tag, so I'll try testing with your [steveheptio/contour:many-listeners](https://hub.docker.com/layers/steveheptio/contour/many-listeners/images/sha256-1ffec68ee6cc214e7b28ef44b8a75fc16a36d6fcb64ba94957baaa283ca924bf?context=explore) image.

@Rycieos thanks! Let me push the latest image, I'm not sure which version that tag currently refers to. Will update here in a sec.

skriss:
OK, latest versions have been pushed:

- [docker.io/steveheptio/contour:many-listeners](https://hub.docker.com/layers/steveheptio/contour/many-listeners/images/sha256-9e4e0ad1aa53ec6217d7d15de04c30f4f11c494499b069a422ec9bf515a96f52?context=repo)
- [docker.io/steveheptio/contour:many-listeners-tcproute](https://hub.docker.com/layers/steveheptio/contour/many-listeners-tcproute/images/sha256-274aa8b38617f7c487873598ba35a12633415c864bcfab4085a03fe8b2ebf681?context=repo) (this adds experimental TCPRoute support on top of the many listeners work, can't recall if you were interested in this too or not)

Rycieos:
> this adds experimental TCPRoute support on top of the many listeners work, can't recall if you were interested in this too or not

I was not; I need TLSRoutes, but not TCP.

internal/gatewayapi/listeners_test.go
```diff
@@ -23,339 +23,493 @@ import (
...
+			},
+		}
+
+		res := ValidateListeners(listeners)
+		assert.Len(t, res.InvalidListenerConditions, 0)
+		assert.Len(t, res.Ports, 1)
+		assert.Len(t, res.ListenerNames, 3)
+	})
+
+	t.Run("Conflicting protocols on a port", func(t *testing.T) {
```
sunjayBhatia, line 376: i dont know if this is covered elsewhere but seems like our new custom protocol could cause a conflict with a HTTP port as well

sunjayBhatia: implementation above looks correct, but a test case to make sure we don't cause any issues in the future would be great!

internal/dag/builder_test.go
```diff
@@ -14217,6 +14135,903 @@ func TestDAGInsert(t *testing.T) {
 	}
 }
 
+func TestGatewayWithHTTPProxyAndIngress(t *testing.T) {
```
sunjayBhatia, line 14138: nice!

internal/dag/builder_test.go
```diff
@@ -14217,6 +14135,903 @@ func TestDAGInsert(t *testing.T) {
...
+							Name:     "https-1",
+							Protocol: gatewayapi_v1beta1.HTTPSProtocolType,
+							Port:     443,
+							AllowedRoutes: &gatewayapi_v1beta1.AllowedRoutes{
+								Namespaces: &gatewayapi_v1beta1.RouteNamespaces{
+									From: ref.To(gatewayapi_v1beta1.NamespacesFromAll),
+								},
+							},
+							TLS: &gatewayapi_v1beta1.GatewayTLSConfig{
+								Mode: ref.To(gatewayapi_v1beta1.TLSModePassthrough),
```
sunjayBhatia, line 14392: i assume this is because we don't have a secret that makes sense in terminate mode since TLS is specified on the HTTPProxy?

now i remember discussing this a bit before but seeing it just now threw me for a second, could be a little counterintuitive for users, still on the fence

sunjayBhatia (approved):
Just a couple small comments, but overall looks great! nice test coverage and looks like the gateway conformance is passing 👍🏽

Rycieos:
Unfortunately, I am unable to test this; due to this limitation on `TLSRoute`s I was unaware of:
> Listener.TLS.Mode must be "Passthrough" when protocol is "TLS".

I need to `Terminate` TLS at the Gateway. Is this an intentional limitation? The Gateway spec seems to suggest both must be supported, but it is not clear. I couldn't find any open issues on this project about this. I would be happy to open one.

But since this is a different message than I got last time, I guess you fixed that problem. Sorry I can't report better results.

skriss:
OK, sounds like we have some more work to do to fully support your use case (though this PR is a big step in the right direction).

There's some conflicting information upstream about the exact combos of listener protocol and route type that should be supported:
- https://gateway-api.sigs.k8s.io/guides/tls/
- https://gateway-api.sigs.k8s.io/concepts/api-overview/#route-summary-table
- https://github.com/kubernetes-sigs/gateway-api/issues/1515

So will need to track down exactly what _should_ be supported. I think thus far we had implemented on the basis of https://gateway-api.sigs.k8s.io/guides/tls/, which indicates TLSRoute is for Passthrough mode.

From my perspective, it seems like TLS termination should be compatible with both TLSRoute (if you need to make a routing decision on the basis of SNI) and TCPRoute (if there's no routing decision to be made, all traffic arriving on the listener port will be proxied to the service(s) in the TCPRoute).

Since you say

internal/dag/builder_test.go
skriss: Ah, this test case can be switched to use the custom protocol, in which case the whole `TLS` block can be removed (this is pretty much the point of supporting the custom protocol, so users _don't_ have to specify tls mode: passthrough which as you say is counterintuitive)

skriss: Oh, I guess I have a test case below that uses the custom protocol, so this is more just documentation that this is a valid alternate way of configuring things.

skriss: The basic problem is that if you specify a protocol of HTTPS, then the Gateway API webhook requires TLS mode to be specified, and if you set it to terminate, then it further requires a secret to be specified. So this is all just to work around the webhook logic.

Rycieos:
> Since you say you need TLSRoute - does that imply you need to make a routing decision based on SNI? If not, then TLS Termination + TCPRoute should be sufficient.

That is correct. Routing by SNI is an extremely useful feature, and as far as I can tell, only Traefik supports that part of the Gateway API currently. Or at least is the only provider that supports `TLSRoutes` **and** their Termination.

> If you could open a new issue to track this work, that'd be great. 

I opened #5461

PR #5166, opened 2023-03-13
fix: migrate from k8s.gcr.io to registry.k8s.io

ahmedwaleedmalik:
Kubernetes is migrating its image registry from k8s.gcr.io to registry.k8s.io.

Part of kubernetes/k8s.io#4780

### DIFF
--- a/examples/gateway/02-certificate_config.yaml
+++ b/examples/gateway/02-certificate_config.yaml
@@ -88,7 +88,7 @@ spec:
     spec:
       containers:
         - name: create
-          image: k8s.gcr.io/ingress-nginx/kube-webhook-certgen:v1.1.1
+          image: registry.k8s.io/ingress-nginx/kube-webhook-certgen:v1.1.1
           imagePullPolicy: IfNotPresent
           args:
             - create
@@ -122,7 +122,7 @@ spec:
     spec:
       containers:
         - name: patch
-          image: k8s.gcr.io/ingress-nginx/kube-webhook-certgen:v1.1.1
+          image: registry.k8s.io/ingress-nginx/kube-webhook-certgen:v1.1.1
           imagePullPolicy: IfNotPresent
           args:
             - patch


liangyuanpeng (approved):

sunjayBhatia (commented):
This is a generated file pulled from https://github.com/kubernetes-sigs/gateway-api so once that releases a new version, this will be updated accordingly

ahmedwaleedmalik:
> This is a generated file pulled from https://github.com/kubernetes-sigs/gateway-api so once that releases a new version, this will be updated accordingly

If I were you, I wouldn't really wait for their release to make this switch. But I'd leave that up to you. As long as you are not affected by https://kubernetes.io/blog/2023/03/10/image-registry-redirect/

skriss:
We can't just update this file statically because it will cause our codegen CI to fail. However, I'm thinking it might be a good idea to add a temporary step to https://github.com/projectcontour/contour/blob/main/Makefile#L253-L259 to replace the old registry with the new. I'm not concerned about it on `main` since we should get an updated Gateway API version before Contour's next release, but it might be good to backport the change to previous releases, just to avoid any issues due to the upcoming redirect. @sunjayBhatia WDYT?

sunjayBhatia:
> We can't just update this file statically because it will cause our codegen CI to fail. However, I'm thinking it might be a good idea to add a temporary step to https://github.com/projectcontour/contour/blob/main/Makefile#L253-L259 to replace the old registry with the new. I'm not concerned about it on `main` since we should get an updated Gateway API version before Contour's next release, but it might be good to backport the change to previous releases, just to avoid any issues due to the upcoming redirect. @sunjayBhatia WDYT?

yeah that sounds like a good idea!

skriss:
@ahmedwaleedmalik I've opened #5175 for this and added you as a co-author, will be closing this one out. Thanks!

PR #5194, opened 2023-03-15
Bump gateway-api to v0.6.2

sunjayBhatia:


sunjayBhatia:
left as a draft for now to show failing conformance tests which we can fix here or in separate dedicated PRs

skriss:
TODOs I see:
- [ ] For routes, **always** set the ResolvedRefs condition, even if the route can't attach to a parent
- [ ] Set the Accepted condition to true on Listeners unless it's explicitly been set to false

sunjayBhatia:
* [ ]  Set the Accepted condition to true on Listeners unless it's explicitly been set to false - covered by https://github.com/projectcontour/contour/pull/5220

sunjayBhatia:
rebased this branch on main w/ commits from open PRs to fix the TODOs

skriss:
>  For routes, always set the ResolvedRefs condition, even if the route can't attach to a parent

I took a quick look at this - looks like naively setting this condition to true for _every_ route+parentRef allows us to pass all of the 0.6.2 tests, but this is obviously not 100% correct. To really implement this correctly we'll need to do some reorganization of our logic, since right now we only process route rules for valid parent refs, and if we do process them, then they get programmed in Envoy. We'll need to be able to validate all routes/rules, but only add them to the DAG if they're for valid parent attachments.

I'm not proceeding further on this for now, just wanted to record my notes.

sunjayBhatia:
> > For routes, always set the ResolvedRefs condition, even if the route can't attach to a parent
> 
> I took a quick look at this - looks like naively setting this condition to true for _every_ route+parentRef allows us to pass all of the 0.6.2 tests, but this is obviously not 100% correct. To really implement this correctly we'll need to do some reorganization of our logic, since right now we only process route rules for valid parent refs, and if we do process them, then they get programmed in Envoy. We'll need to be able to validate all routes/rules, but only add them to the DAG if they're for valid parent attachments.
> 
> I'm not proceeding further on this for now, just wanted to record my notes.

Implemented it here: https://github.com/projectcontour/contour/pull/5259, will add more context/notes there

skriss (approved):
🎉

PR #5195, opened 2023-03-17
HTTPProxy: support Host header rewrites per-service

### LINKED ISSUE #4002: support Host header rewrites per-service
Currently, Contour supports rewriting the `Host` header at the route level only. It's explicitly disallowed at the per-service level, because Envoy did not historically support it. As of Envoy 1.19.0, however, setting `host_rewrite_literal` for an individual weighted cluster is now supported (see https://www.envoyproxy.io/docs/envoy/v1.19.0/version_history/current#new-features), so Contour should expose this functionality.

fangfpeng:
Closes #4002

### DIFF
--- a/internal/dag/policy_test.go
+++ b/internal/dag/policy_test.go
@@ -607,7 +607,7 @@ func TestHeadersPolicy(t *testing.T) {
 	}
 	for name, tc := range tests {
 		t.Run(name, func(t *testing.T) {
-			got, gotErr := headersPolicyService(&tc.dhp, tc.hp, dynamicHeaders)
+			got, gotErr := headersPolicyService(&tc.dhp, tc.hp, true, dynamicHeaders)
 			if tc.wantErr {
 				assert.Error(t, gotErr)
 			} else {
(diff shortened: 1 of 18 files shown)


skriss:
Can also drop this line now: https://github.com/projectcontour/contour/blob/main/apis/projectcontour/v1/httpproxy.go#L885

(will need to run `make generate` afterwards)

internal/envoy/v3/route_test.go
```diff
@@ -66,7 +66,7 @@ func TestRouteRoute(t *testing.T) {
 				ServiceNamespace: s1.Namespace,
 				ServicePort:      s1.Spec.Ports[0],
 			},
-		},
+		}
```
skriss, line 69: Looks like we need the comma back!

internal/dag/policy_test.go
```diff
@@ -1217,7 +1217,68 @@ func TestValidateHeaderAlteration(t *testing.T) {
...
 		t.Run(test.name, func(t *testing.T) {
-			got, gotErr := headersPolicyService(test.dhp, test.in, test.dyn)
+			got, gotErr := headersPolicyService(test.dhp, test.in, false, test.dyn)
+			assert.Equal(t, test.want, got)
+			assert.Equal(t, test.wantErr, gotErr)
+		})
+	}
+}
+
+func TestAllowHostHeaderRewrite(t *testing.T) {
```
skriss, line 1227: Since we're using row-based testing here, it looks like we should be able to fold these test cases into `TestHeadersPolicy` -- `in` -> `hp`, can drop `dyn` entirely since it's not used in these test cases.

sunjayBhatia (approved):

PR #5210, opened 2023-03-20
test/e2e: clone DefaultTransport and disable keep-alives

skriss:


test/e2e/http.go
```diff
@@ -103,7 +103,17 @@ func (h *HTTP) RequestUntil(opts *HTTPRequestOpts) (*HTTPResponse, bool) {
...
 	}
 
-	client := &http.Client{}
+	// Clone the DefaultTransport and disable keep alives
+	// so we don't reuse connections within this method or
+	// across multiple calls to this method. This helps
+	// prevent requests from inadvertently being made to
+	// a draining Listener.
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	transport.DisableKeepAlives = true
```
sunjayBhatia, line 112: yeah i think the previous line is helpful but this one adds more assurance (also in the https request version) since even within a test, you could be updating a listener, so the below `requestUntil` could reuse a connection

PR #5211, opened 2023-03-20
Automatically set GOMAXPROCS to match the number of CPUs available to the container

rajatvig:
Follow up to https://github.com/projectcontour/contour/pull/5023 to address https://github.com/projectcontour/contour/issues/5022

### DIFF
--- a/changelogs/unreleased/5211-rajatvig-small.md
+++ b/changelogs/unreleased/5211-rajatvig-small.md
@@ -1,0 +1,1 @@
+Contour now sets GOMAXPROCS to match the number of CPUs available to the container which results in lower and more stable CPU usage under high loads and where the container and node CPU counts differ significantly.
--- a/changelogs/unreleased/5211-rajatvig-small.md
+++ b/changelogs/unreleased/5211-rajatvig-small.md
@@ -1,1 +1,2 @@
-Contour now sets GOMAXPROCS to match the number of CPUs available to the container which results in lower and more stable CPU usage under high loads and where the container and node CPU counts differ significantly.
+Contour now sets `GOMAXPROCS` to match the number of CPUs available to the container which results in lower and more stable CPU usage under high loads and where the container and node CPU counts differ significantly.
+This is the default behavior but can be overridden by specifying `GOMAXPROCS` to a fixed value as an environment variable.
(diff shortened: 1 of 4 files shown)


rajatvig:
Unsure how to go about fixing the code coverage drop.

changelogs/unreleased/5211-rajatvig-small.md
```diff
@@ -0,0 +1 @@
+Contour now sets GOMAXPROCS to match the number of CPUs available to the container which results in lower and more stable CPU usage under high loads and where the container and node CPU counts differ significantly.
```
sunjayBhatia, line 1: small change in wording but something to the effect of saying this is the new default behavior and you can still override it by setting `GOMAXPROCS` yourself would be good

rajatvig: Does the update look good?

cmd/contour/contour.go
```diff
@@ -29,6 +31,9 @@ func main() {
 	log := logrus.StandardLogger()
 	k8s.InitLogging(k8s.LogWriterOption(log.WithField("context", "kubernetes")))
 
+	// set GOMAXPROCS
+	_, _ = maxprocs.Set(maxprocs.Logger(log.Printf))
```
sunjayBhatia, line 35: i think we do want be defensive and handle the error here, its unlikely to occur but good to still check

something like
```
if err != nil {
  log.WithError(err).Fatal("failed to set GOMAXPROCS")
}
```

sunjayBhatia (changes requested):
lgtm other than the one comment

logs look like this with no resources defined:
```
time="2023-03-23T13:39:58Z" level=info msg="maxprocs: Leaving GOMAXPROCS=8: CPU quota undefined"
```

when defined:
```
time="2023-03-23T13:43:42Z" level=info msg="maxprocs: Updating GOMAXPROCS=1: determined from CPU quota"
```

when set in the environment:
```
time="2023-03-23T13:49:02Z" level=info msg="maxprocs: Honoring GOMAXPROCS=\"4\" as set in environment"
```

sunjayBhatia (approved):

skriss (approved):

PR #5214, opened 2023-03-22
Allow limiting Contour to watch certain namespaces

### LINKED ISSUE #2497: Support for running Contour in a single namespace

### LINKED ISSUE #4899: Question about "installing multiple Contour instances"

nsimons:
This PR is a continuation of #5002 with added checks, test cases and documentation.

Fixes #2497
Fixes #4899

Signed-off-by: Niklas Simons [niklas.simons@est.tech](mailto:niklas.simons@est.tech)


tsaarni (commented):
Looks very good to me :+1:  Few questions

test/e2e/httpproxy/httpproxy_test.go
```diff
@@ -444,7 +444,7 @@ descriptors:
 				}
 			})
 
-			f.NamespacedTest("root-ns-crd", testRootNamespaces(rootNamespaces))
+			f.NamespacedTest("nonroot-ns-crd", testRootNamespaces(rootNamespaces))
```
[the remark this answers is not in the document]
nsimons: Yeah it was a bit confusing for me at first as well, hence the renaming. I'll refactor it further so that it's no longer a magic string.

examples/namespaced/kustomization.yaml
```diff
@@ -0,0 +1,54 @@
+resources:
```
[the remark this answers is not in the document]
nsimons: Sure!

site/content/docs/main/deploy-options.md
```diff
@@ -219,6 +219,9 @@ Also see the [upgrade guides][15] on steps to roll out a new version of Contour.
 
 It's possible to run multiple instances of Contour within a single Kubernetes cluster.
 This can be useful for separating external vs. internal ingress, for having separate ingress controllers for different ingress classes, and more.
+Each Contour instance can also be configured via the `--watch-namespaces` flag to handle their own namespaces. This allows the Kubernetes RBAC objects
```
skriss, line 222: A nice follow-up to this work would be to provide a way to set this flag & adjust the RBAC resources via the Gateway provisioner, for folks who are using the provisioner. I'll file a separate issue for that.

examples/namespaced/kustomization.yaml
```diff
@@ -0,0 +1,61 @@
+# This kustomize file can be used as an example for deploying Contour without ClusterRole and
```
skriss, line 1: Noting that this does not work with Gateway API, since that requires watching Namespaces and GatewayClasses which are cluster-scoped. I think we can address this (if needed) along with the follow-up for the Gateway provisioner noted separately.

test/e2e/httpproxy/namespaces_test.go
```diff
@@ -0,0 +1,216 @@
...
+			// Proxy in non-watched namespace should not be handled
+			deployEchoServer(f.T(), f.Client, nonWatchedNS, "echo")
+			p := newEchoProxy("proxy", nonWatchedNS)
+			err := f.CreateHTTPProxy(p)
+			require.Nil(f.T(), err, "could not create httpproxy")
+
+			// wait.Poll automatically waits the interval before first poll.
+			// If the resource is still NotReconsiled we can be fairly sure that the proxy is not handling it
+			interval := 10 * time.Second
+			ok := wait.Poll(interval, f.RetryTimeout, func() (bool, error) {
```
skriss, line 57: We use `require.Never` in a few places for this sort of thing - poll periodically up to a time limit that a condition does not occur. In this case we could either check that it never becomes valid, or that it never loses the NotReconciled status. I have a slight preference for that approach for consistency.

test/e2e/httpproxy/namespaces_test.go
```diff
@@ -0,0 +1,216 @@
...
+				Condition: e2e.HasStatusCode(200),
+			})
+			require.NotNil(f.T(), res)
+			require.Truef(f.T(), ok, "expected 200 response code, got %d", res.StatusCode)
+
+			// Root proxy in non-watched namespace should fail
+			deployEchoServer(f.T(), f.Client, nonWatchedNS, "echo")
+			p = newEchoProxy("root-proxy", nonWatchedNS)
+			_, ok = f.CreateHTTPProxyAndWaitFor(p, e2e.HTTPProxyNotReconciled)
+			require.Truef(f.T(), ok, "expected HTTPProxy to have status NotReconciled")
```
skriss, line 143: Would be good to add (brief) polling here too

skriss (changes requested):
Thanks @nsimons, this looks pretty good to me, couple comments below

test/e2e/httpproxy/namespaces_test.go
nsimons: That's neat, thanks!

examples/namespaced/kustomization.yaml
nsimons: Indeed, this kustomization file is now focusing on the "standard" Contour configuration (render/contour.yaml), which does not include Gateway API configuration. We can create a new kustomization file for namespaced Gateway-enabled Contour separately.

skriss:
Thanks for the updates @nsimons - I think this looks good, would just like to poke around with testing a little bit more, but we should be able to get this into the upcoming release

nsimons:
Hmm, right.
I don't know if there is any clean way to deal with this. The cache needs to cover the namespace where Envoy's service resides (same as Contour, or given by `--envoy-service-namespace`), otherwise the status updater will not get any events.

We could force the user to give the namespace in `--watch-namespaces` (similar as envoy client certificate and fallback cert), but if the user does not really care about it then it's a superfluous restriction. Or we can add a warning that LoadBalancerStatus is not populated if the namespace is not watched. I'm leaning towards the latter.

It depends which use cases we want to support. From the related issues I think the most common scenario will be that everything is located within the same namespace (Contour, Envoy, application). Another one that we can foresee is that Contour+Envoy is in an infra namespace and the app is somewhere else. In that case it's probably okay that Contour watches both infra and app namespaces, even if we don't expect any resources in infra ns.

skriss:
Yeah, thinking some more I think it's fair to say that if you want the Envoy service to be watched, you need to include it. Let's just see if we can find a place to explicitly document this -- I'm thinking maybe `deploy-options.md` is the most logical place to cover it.

skriss:
xref https://github.com/projectcontour/contour/pull/5099#issuecomment-1503973546 -- so looks like:
- currently, this PR (#5214) and #5099 are incompatible with each other
- there are upcoming changes in controller-runtime that will break both, but not sure when the new version that includes those changes will be out

tsaarni:
> xref [#5099 (comment)](https://github.com/projectcontour/contour/pull/5099#issuecomment-1503973546) -- so looks like:
> 
> * currently, this PR ([Allow limiting Contour to watch certain namespaces #5214](https://github.com/projectcontour/contour/pull/5214)) and [Reduce the memory footprint by optimizing Kubernetes informer cache #5099](https://github.com/projectcontour/contour/pull/5099) are incompatible with each other
> * there are upcoming changes in controller-runtime that will break both, but not sure when the new version that includes those changes will be out

I did not find a way to achieve both goals with the current version.  I submitted question https://github.com/kubernetes-sigs/controller-runtime/issues/2273 asking for help.

I tested the main branch of controller-runtime and I can confirm it solves the problem.  I stored a diff of my experiment here https://gist.github.com/tsaarni/94184af509664334e0b3350c9c0ac86a. The project is going through API refactoring and few changes were required around the code base.

tsaarni:
> I did not find a way to achieve both goals with the current version. I submitted question [kubernetes-sigs/controller-runtime#2273](https://github.com/kubernetes-sigs/controller-runtime/issues/2273) asking for help.

I got confirmation that it is not possible as of today.  Though I'm not absolutely 100% certain if one could hack around the limitation, maybe one could reproduce some larger part of the builder logic on Contour side...
 
The API in `main` has not yet been settled and there will be another round of changes, but it looks like we will be able to have namespace restricted watches + transform functions in the final API.

skriss:
So sounds like we may have to pick/choose between #5214 and #5099 for the upcoming Contour 1.25, and then plan to refactor later to support both. @tsaarni @nsimons what are your thoughts on priority between the two? From my perspective #5099 would be higher priority, but  would like your take as well.

sunjayBhatia:
> So sounds like we may have to pick/choose between #5214 and #5099 for the upcoming Contour 1.25, and then plan to refactor later to support both. @tsaarni @nsimons what are your thoughts on priority between the two? From my perspective #5099 would be higher priority, but would like your take as well.

personally +1 on #5099 since that can have a wider benefit for users, but can be swayed as well

tsaarni:
We discussed this with @nsimons. Maybe there could be one option that would allow doing both, with one limitation:

**Background**

Currently controller-runtime allows setting **single** namespace `manager.Options{Namespace: "foo"}` while still using the `BuilderWithOptions()` which is needed to implement the memory footprint optimization. 

**Proposal** 

In the 1st phase we could limit #5214 to **single** namespace. When controller-runtime releases the new API, we could have 2nd phase that allows **multiple** namespaces to be watched. Multiple namespaces can be useful since it allows "cross-namespace" features to work, even when Contour is not granted cluster roles.

Alternative 1: We could keep `--watch-namespaces` (plural) but have it print an error "currently only single namespace is allowed" if more than one namespace is given. In future, we can allow multiple namespaces without breaking the API.

Alternative 2: We could change it to `--watch-namespace` and introduce new command line option later, when (or if) watching multiple namespaces will be added.

For our own use case the limitation of single namespace would be completely OK. If we agree on the approach, @nsimons promised to make the changes to this PR on top of #5099, so we could work on the memory footprint issue first.

sunjayBhatia:
> **Proposal**
> 
> In the 1st phase we could limit #5214 to **single** namespace. When controller-runtime releases the new API, we could have 2nd phase that allows **multiple** namespaces to be watched. Multiple namespaces can be useful since it allows "cross-namespace" features to work, even when Contour is not granted cluster roles.
> 
> Alternative 1: We could keep `--watch-namespaces` (plural) but have it print an error "currently only single namespace is allowed" if more than one namespace is given. In future, we can allow multiple namespaces without breaking the API.
> 
> Alternative 2: We could change it to `--watch-namespace` and introduce new command line option later, when (or if) watching multiple namespaces will be added.
> 
> For our own use case the limitation of single namespace would be completely OK. If we agree on the approach, @nsimons promised to make the changes to this PR on top of #5099, so we could work on the memory footprint issue first.

I'm down with this proposal, and Alternative 1 (if we want to do 2 we could make the flag like `--disable-feature` flag which can be specified multiple times)

sunjayBhatia (commented):
relevant to this PR, this issue has some comments and links with details on the next controller-runtime release date and release notes: https://github.com/kubernetes-sigs/controller-runtime/issues/2295

nsimons:
Thanks Sunjay. I think it's better that we wait for controller-runtime 0.15.0 which contains the full multi-namespace support. Then we can merge this PR almost as-is.

It should be released quite soon as far as I understood. I started to bump and test the alpha version in a related PR.

tsaarni (approved):
LGTM :+1:

skriss (approved):
LGTM, thanks @nsimons!

sunjayBhatia (approved):

PR #5217, opened 2023-03-23
internal/envoy/v3: Trim trailing slashes from path separated prefix match

sunjayBhatia:
Found when testing out
https://github.com/kubernetes-sigs/gateway-api/pull/1855

We could also do this higher up in dag processing but this seems like the common/failsafe option.

### DIFF
--- a/internal/envoy/v3/route.go
+++ b/internal/envoy/v3/route.go
@@ -194,7 +194,9 @@ func PathRouteMatch(pathMatchCondition dag.MatchCondition) *envoy_route_v3.Route
 		case dag.PrefixMatchSegment:
 			return &envoy_route_v3.RouteMatch{
 				PathSpecifier: &envoy_route_v3.RouteMatch_PathSeparatedPrefix{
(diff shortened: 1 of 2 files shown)


internal/envoy/v3/route.go
```diff
@@ -194,7 +194,9 @@ func PathRouteMatch(pathMatchCondition dag.MatchCondition) *envoy_route_v3.Route
 		case dag.PrefixMatchSegment:
 			return &envoy_route_v3.RouteMatch{
 				PathSpecifier: &envoy_route_v3.RouteMatch_PathSeparatedPrefix{
-					PathSeparatedPrefix: c.Prefix,
+					// Trim trailing slash as PathSeparatedPrefix expects
+					// no trailing slashes.
+					PathSeparatedPrefix: strings.TrimSuffix(c.Prefix, "/"),
```
skriss, line 199: I think it makes sense to use `strings.TrimRight` here to handle removing any number of trailing slashes, just to be safe since there is no particular validation on the field

PR #5233, opened 2023-03-29
internal/provisioner: Do not use hard-coded values, but configuration defined ports.

Jean-Daniel:
This fix an issue with envoy pods generated by contour provisioner.

Envoy pods are never declared `Ready` when using a custom health port.
Envoy exposes the wrong port in prometheus annotations when using a custom metrics port.

### DIFF
--- a/internal/provisioner/objects/dataplane/dataplane.go
+++ b/internal/provisioner/objects/dataplane/dataplane.go
@@ -136,6 +136,14 @@ func desiredContainers(contour *model.Contour, contourImage, envoyImage string) 
 		ports = append(ports, p)
 	}
 
+	healthPort := 8002
+	if contour.Spec.RuntimeSettings != nil &&
+		contour.Spec.RuntimeSettings.Envoy != nil &&
+		contour.Spec.RuntimeSettings.Envoy.Health != nil &&
+		contour.Spec.RuntimeSettings.Envoy.Health.Port > 0 {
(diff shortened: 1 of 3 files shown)


internal/provisioner/objects/dataplane/dataplane.go
```diff
@@ -204,7 +204,7 @@ func desiredContainers(contour *model.Contour, contourImage, envoyImage string)
 					HTTPGet: &corev1.HTTPGetAction{
 						Scheme: corev1.URISchemeHTTP,
 						Path:   "/ready",
-						Port:   intstr.IntOrString{IntVal: int32(8002)},
+						Port:   intstr.IntOrString{IntVal: int32(contour.Spec.RuntimeSettings.Envoy.Health.Port)},
```
skriss, line 207: This reference is nil by default which is resulting in a panic. We'll need to add some nil-checking and defaulting here. Please also add unit tests that cover this for the default and non-default cases.

skriss (changes requested):

skriss:
Thanks for the update @Jean-Daniel -- last two things are to [add a changelog file](https://github.com/projectcontour/contour/actions/runs/4621735162/jobs/8173532461?pr=5233) and [add DCO signoff to your commits](https://github.com/projectcontour/contour/pull/5233/checks?check_run_id=12532013872).

skriss (approved):

sunjayBhatia (approved):

PR #5234, opened 2023-03-29
Expose metric port in generated envoy containers

Jean-Daniel:
This is a fix for #5232


Jean-Daniel:
This is a duplicate of #5101

PR #5108, opened 2023-02-20
add install-provisioner-working

nerdynikhil:
Updates #5102 

Added setup script for Contour Provisioner and updated Makefile to build the latest version of Contour Provisioner image

PR #5238, opened 2023-03-31
docs: updated contributing instructions

tsaarni:
This is a proposal for minimal update on contributor docs. It could be useful for KubeCon EU 2023 ContribFest participants. Any further ideas are welcome!

Some notes:

- ~When user runs `make install-{contour,provisioner}-working` it exposes the cluster not only to localhost but to the network their laptop is on. https://github.com/projectcontour/contour/blob/3a85420d0b97155629fa18ca03db01db2c65438e/test/scripts/kind-expose-port.yaml#L9 Can we change this without impacting e2e tests?~ Now using 127.0.0.1 instead.
- I added support for running `codespell` as a container, using image from the codespell actions, allowing people to run `make lint` without python requirement
- I removed `sed -i` in-place edit of manifests, so that people do not accidentally end up committing them.
- We require `jq` but also requiring `yq` stands out a bit in my opinion - too many little utilities ;) I know it was just recently added https://github.com/projectcontour/contour/pull/5108#discussion_r1139401694. For ubuntu it is not available as `.deb`, it is available as snap. Some people may dislike installing snaps ;) I could imagine maybe kustomize to be used to achieve the same, when sed is not wanted.
- It could be nice if people could run unit tests too inside build container, but maybe that is for the future.
- ~Tagging each image with PID of the build script is not that nice ([link](https://github.com/projectcontour/contour/blob/7730dbbedebcefab5135b30be7574908fa719ee3/test/scripts/install-contour-working.sh#L48)): user's machine `docker image ls` will quickly fill up with contour images. Using some other convention would be better.~ Edit: now using github hash + dirty flag

### DIFF
--- a/test/scripts/install-contour-working.sh
+++ b/test/scripts/install-contour-working.sh
@@ -44,8 +44,10 @@ if ! kind::cluster::exists "$CLUSTERNAME" ; then
     exit 2
 fi
 
-# Build the current version of Contour.
-VERSION="v$$"
+# Set the image tag to match the current git hash + dirty flag.
+VERSION=$(git describe --exclude="*" --always --dirty)
+
+# Build the image.
 make -C ${REPO} container IMAGE=ghcr.io/projectcontour/contour VERSION=${VERSION}
 
 # Push the Contour build image into the cluster.
--- a/test/scripts/install-contour-working.sh
+++ b/test/scripts/install-contour-working.sh
@@ -44,15 +44,6 @@ if ! kind::cluster::exists "$CLUSTERNAME" ; then
     exit 2
 fi
 
-# Wrap sed to deal with GNU and BSD sed flags.
-run::sed() {
-    local -r vers="$(sed --version < /dev/null 2>&1 | grep -q GNU && echo gnu || echo bsd)"
-    case "$vers" in
-        gnu) sed -i "$@" ;;
-        *) sed -i '' "$@" ;;
-    esac
-}
-
 # Build the current version of Contour.
 VERSION="v$$"
 make -C ${REPO} container IMAGE=ghcr.io/projectcontour/contour VERSION=${VERSION}
@@ -73,17 +64,13 @@ for file in ${REPO}/examples/contour/02-job-certgen.yaml ${REPO}/examples/contou
   # Set image pull policy to IfNotPresent so kubelet will use the
   # images that we loaded onto the node, rather than trying to pull
   # them from the registry.
-  run::sed \
-    "-es|imagePullPolicy: Always|imagePullPolicy: IfNotPresent|" \
-    "$file"
-
   # Set the image tag to $VERSION to unambiguously use the image
   # we built above.
-  run::sed \
+  sed \
+    "-es|imagePullPolicy: Always|imagePullPolicy: IfNotPresent|" \
     "-es|image: ghcr.io/projectcontour/contour:.*$|image: ghcr.io/projectcontour/contour:${VERSION}|" \
-    "$file"
-
-  ${KUBECTL} apply -f "$file"
+    "$file" | \
+  ${KUBECTL} apply -f -
 done
 
 # Wait for Contour and Envoy to report "Ready" status.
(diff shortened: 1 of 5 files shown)


tsaarni:
BTW: I've myself used following port mappings with kind

```yaml
- role: worker
  extraPortMappings:
  - containerPort: 80
    hostPort: 80
    listenAddress: "127.0.0.101"
  - containerPort: 443
    hostPort: 443
    listenAddress: "127.0.0.101"
```

I'd have to check if this approach works for macOS at all (when docker is in VM), but I've liked this since it allows me to use the well-known ports, eliminating one detail that can go wrong. Using well-known ports on 127.0.0.1 might bit obtrusive, so I've used randomly picked another localhost address like 127.0.0.101. Instead of `local.projectcontour.io` I've used `my-host-xyz.127-0-0-101.nip.io` which has the added benefit that one can test virtualhosts too.

tsaarni:
> I'd have to check if this approach works for macOS at all (when docker is in VM)

I've checked and it does work, but with complications: unlike Linux, macOS does not respond to loopback addresses 127.0.0.0/8 other than 127.0.0.1, unless one creates an alias `sudo ifconfig lo0 alias 127.0.0.101 up`. There was also a prompt for admin password when Docker attempts binding privileged ports (80 and 443), something that does not happen on Linux.

Another observation I had when setting up environment for my Macbook: Contour's scripts only work with Docker Desktop but not podman (kind can work with podman). People might be unable to install Docker Desktop due to its license terms. 

One more macOS issue: I wonder are people doing CRD generation on macOS?  For me `make generate-crd-deepcopy` fails due to [kubernetes-sigs/controller-tools/issues/734](https://www.github.com/kubernetes-sigs/controller-tools/issues/734) 

```
Error: unable to parse option "object:headerFile=/var/folders/3f/08k867j16n1c0wvrxzp19b_m0000gn/T/tmp.076ezrxD": [invalid digit '8' in octal literal (at <input>:1:28) exponent has no digits (at <input>:1:64)] 
```

controller-gen does not like the names that `mktemp` on macOS creates
 https://github.com/projectcontour/contour/blob/3a85420d0b97155629fa18ca03db01db2c65438e/hack/generate-crd-deepcopy.sh#L37

tsaarni:
I took some detailed notes on how to start from ~clean install of Ubuntu 22.04 (or 20.04) https://gist.github.com/tsaarni/d9dea0a869be8a68401b5c3fb5f2915d

skriss (approved):
Changes LGTM, thanks @tsaarni!

FWIW, I am doing Contour dev locally on my MacBook (M1) these days. I'll see if I can work on adding a section with some macOS-specific tips and tricks (metallb + docker-mac-net-connect is the biggest one I can think of, that avoids having to use kind port mappings to access Envoy)

test/scripts/install-contour-working.sh
```diff
@@ -44,17 +44,10 @@ if ! kind::cluster::exists "$CLUSTERNAME" ; then
...
-# Wrap sed to deal with GNU and BSD sed flags.
-run::sed() {
-    local -r vers="$(sed --version < /dev/null 2>&1 | grep -q GNU && echo gnu || echo bsd)"
-    case "$vers" in
-        gnu) sed -i "$@" ;;
-        *) sed -i '' "$@" ;;
-    esac
-}
+# Set the image tag to match the current git hash + dirty flag.
+VERSION=$(git describe --exclude="*" --always --dirty)
```
skriss, line 48: I noticed an issue with this change, repeat runs of `make install-contour-working` if you've made local changes no longer trigger a Deployment rollout because nothing has changed in the spec, whereas previously the tag was changing. We could explicitly run a `kubectl rollout restart`, or set an annotation on the Deployment, or explicitly delete the pods, or ...

skriss: No worries, just realized it today when troubleshooting why my updated code wasn't running!

Yeah, maybe it's best to go back to random for now to keep that repeated deployment use case working, and we can revisit a long-term solution that addresses all the issues.

skriss: Fixed for now in #5278

PR #5257, opened 2023-04-06
Make alpha Gateway APIs optional

### LINKED ISSUE #5247: allow alpha Gateway API resource informers to be disabled
Currently, Contour packages the experimental channel of Gateway API, which includes both alpha and beta resources. However, some users may like to only have standard channel (beta) resources in their cluster. Contour should allow disabling the informers for alpha resources in that case.

We should be able to make use of the `--disable-feature` flag to `contour serve` which already allows disabling Contour's alpha ExtensionService resource informer.

nsimons:
Fixes #5247

Signed-off-by: Niklas Simons [niklas.simons@est.tech](mailto:niklas.simons@est.tech)

### DIFF
--- a/changelogs/unreleased/5080-nsimons-minor.md
+++ b/changelogs/unreleased/5080-nsimons-minor.md
@@ -3,6 +3,10 @@
 The `contour serve` command takes a new optional flag, `--disable-feature`, that allows disabling
 certain features.
 
-Currently this flag can be used to disable the informer for ExtensionService resources,
-effectively making the ExtensionService CRD optional in the cluster.
-To do this, use the flag as follows: `--disable-feature=extensionservices`
+The flag is used to disable the informer for a custom resource, effectively making the corresponding
+CRD optional in the cluster. You can provide the flag multiple times.
+
+Current options include `extensionservices` and the experimental Gateway API features `tlsroutes` and
+`grpcroutes`.
+
+For example, to disable ExtensionService CRD, use the flag as follows: `--disable-feature=extensionservices`.
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -126,7 +126,7 @@ func registerServe(app *kingpin.Application) (*kingpin.CmdClause, *serveContext)
 	serve.Flag("debug", "Enable debug logging.").Short('d').BoolVar(&ctx.Config.Debug)
 	serve.Flag("debug-http-address", "Address the debug http endpoint will bind to.").PlaceHolder("<ipaddr>").StringVar(&ctx.debugAddr)
 	serve.Flag("debug-http-port", "Port the debug http endpoint will bind to.").PlaceHolder("<port>").IntVar(&ctx.debugPort)
-	serve.Flag("disable-feature", "Do not start an informer for the specified resources.").PlaceHolder("<extensionservices>").EnumsVar(&ctx.disabledFeatures, "extensionservices")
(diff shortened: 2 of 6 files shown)


nsimons:
This should be good for review now.
Few points:
- the kustomization file requires `--load-restrictor=LoadRestrictionsNone` when rendered due to our unconventional kustomization folder structure, we can discuss/change this in scope of #5258 
- I decided to change my previous PRs changelog (#5080) instead of making a new one, as I don't think it has been released yet (?). This makes the coming release note a bit cleaner IMO. Let me know if I should do it differently

sunjayBhatia:
> * I decided to change my previous PRs changelog ([cmd/contour: optional CRDs #5080](https://github.com/projectcontour/contour/pull/5080)) instead of making a new one, as I don't think it has been released yet (?). This makes the coming release note a bit cleaner IMO. Let me know if I should do it differently

sounds good to me 👍🏽

skriss:
I tried this out in a cluster with a [standard channel Gateway API install](https://gateway-api.sigs.k8s.io/guides/#install-standard-channel) and all looks good to me.

skriss (approved):

sunjayBhatia (approved):
could also modify the ClusterRole in the optional-features kustomization to remove the rules for the relevant resources to offer that as an example

sunjayBhatia:
One additional thought on this is what happens when TLSRoute, GRPCRoute etc. are moved to the "standard" support channel. If we plan on disabling the ability to opt-out of reconciling these CRDs we should call that out in the release notes.

skriss:
> One additional thought on this is what happens when TLSRoute, GRPCRoute etc. are moved to the "standard" support channel. If we plan on disabling the ability to opt-out of reconciling these CRDs we should call that out in the release notes.

Hmm, yeah good question. I had been thinking of this as a way to opt out of the alpha resources, but I could see an argument for continuing to want to opt out of them even after they become beta/GA, if you're only interested in HTTPRoute.

nsimons:
Yeah I agree, this can be useful even in the future for some folks that don't want/need to install all CRDs, like in the original issue #4684.

nsimons:
> could also modify the ClusterRole in the optional-features kustomization to remove the rules for the relevant resources to offer that as an example

I investigated this a bit, sure it's possible but kustomize is not that well-equipped to do it cleanly. Using JSON patches we can do it with this kind of syntax:

```yaml
  - patch: |-
      - op: remove
        path: /rules/1/resources/2
      - op: remove
        path: /rules/1/resources/4
    target:
      group: rbac.authorization.k8s.io
      kind: ClusterRole
      name: contour
      version: v1
```
The above snippet looks for the ClusterRole, and from the rule with index 1 removes resources with indexes 2 and 4 (corresponding to `grpcroutes` and `tlsroutes` respectively).
This is obviously quite brittle, as is will break if the item order changes.

We can also add `test` directives, which can check what we are about to `remove`:

```yaml
  - patch: |-
      - op: test
        path: /rules/1/resources/2
        value: grpcroutes
      - op: remove
        path: /rules/1/resources/2

      - op: test
        path: /rules/1/resources/4
        value: tlsroutes
      - op: remove
        path: /rules/1/resources/4
```

and we could have a simple script that tries to execute `kustomize` any time we re-generate the examples. In that way we can catch the breakage and fix it.

I think this is something we could discuss later, since there is a maintenance aspect to it. We could also clean up the RBAC for `examples/render/contour.yaml` since that one contains all Gateway permissions, but doesn't render any other Gateway-related resources.

sunjayBhatia:
> I think this is something we could discuss later, since there is a maintenance aspect to it. We could also clean up the RBAC for `examples/render/contour.yaml` since that one contains all Gateway permissions, but doesn't render any other Gateway-related resources.

yep sounds good, thanks for looking into this, sounds like not straightforward to just plop in now so we can discuss further in https://github.com/projectcontour/contour/issues/5258

PR #5259, opened 2023-04-07
internal/dag: Always set resolved refs in GW API processor

sunjayBhatia:
If no matching listeners, we still walk the route and resolve any backendrefs

To cover GW conformance for v0.6.2


skriss (approved):
LGTM, this seems like the most practical approach given the current code structure.

sunjayBhatia:
> LGTM, this seems like the most practical approach given the current code structure.

yeah, tried a few other experiments but they got pretty messy

PR #5270, opened 2023-04-11
minor reorganization for CONTRIBUTING.md

skriss:
Did a few more small edits to try to make the happy path as clear as possible:
- start with the basic `go build` and `go test` commands since they're fastest and require the least tooling
- combine some of the other `make` target docs into a single section on pre-submit checks using `make checkall`
- clarify that running the Contour binary locally against a kind cluster is an alternative approach to `make install-contour-working` (I think at this point the latter is efficient enough on repeat builds that it'll work for most folks)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,18 +31,41 @@ This section describes how to build Contour from source.
 
 ### Fetch the source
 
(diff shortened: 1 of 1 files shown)


tsaarni (approved):
LGTM :+1:

PR #5296, opened 2023-04-19
internal/dag: Fix for duplicate include detection

sunjayBhatia:
If we have an include on a certain path, then a different valid include condition on that same path, duplicates of the later include conditions were not detected.

This actually does not need to be backported to the 1.24 release as that introduced the improved duplicate detection behavior, but did add the seen path/header conditions to the map, see here: https://github.com/projectcontour/contour/blob/v1.24.3/internal/dag/httpproxy_processor.go#L1441


skriss (approved):
LGTM

PR #5297, opened 2023-04-19
[release-1.24] internal/dag: Fix bug in duplicate include match detection that caused memory usage spike

sunjayBhatia:
While looping through seen include match conditions for a particular path, code was appending to the seen list 😅.

Also using a slice of slices causes a lot more allocations as you increase includes and header match conditions.

Was able to repro a huge memory spike with a single root HTTPProxy with 500 includes with 5 header match conditions each.

No release note but will add when cutting the patch release that contains this change


sunjayBhatia:
@skriss added the e2e test from https://github.com/projectcontour/contour/pull/5323 to this PR since it seems to pretty reliably prove out the change

PR #5301, opened 2023-04-20
chore: remove double package import

### LINKED ISSUE #5300: remove double import in internal/envoy/v3/listener.go
In internal/envoy/v3/listener.go, the package `github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/router/v3` is imported twice under two different aliases (`envoy_extensions_filters_http_router_v3` and `envoy_router_v3`). 

The import using `envoy_extensions_filters_http_router_v3` should be removed, and the references to it should be updated to use `envoy_router_v3`.

padlar:
fixes #5300


sunjayBhatia (approved):

pnbrown (commented):
It looks good to me

PR #5302, opened 2023-04-20
Add an option to GatewayProvisioner CRD to watch subset of namespaces

### LINKED ISSUE #5256: Support namespace-scoped Contour instances via Gateway provisioner
As a follow-up to #5214, Contour's Gateway provisioner should support deploying Contour scoped to watching a subset of namespaces or one namespace. This requires adding an option to ContourDeployment for specifying the namespaces to watch, and then setting the `--watch-namespaces` flag on `contour serve` and adjusting the RBAC resources to be 1+ Roles and RoleBindings instead of a ClusterRole and ClusterRoleBinding.

_Originally posted by @skriss in https://github.com/projectcontour/contour/pull/5214#discussion_r1157821393_

padlar:
fixes #5256

### DIFF
--- a/apis/projectcontour/v1alpha1/contourdeployment.go
+++ b/apis/projectcontour/v1alpha1/contourdeployment.go
@@ -114,6 +114,10 @@ type ContourSettings struct {
 	// Deployment describes the settings for running contour as a `Deployment`.
 	// +optional
 	Deployment *DeploymentSettings `json:"deployment,omitempty"`
+
+	// WatchNamespaces restricts watching contour instances to a subset of namespaces.
+	// +optional
+	WatchNamespaces string `json:"watchNamespaces,omitempty"`
 }
 
 // DeploymentSettings contains settings for Deployment resources.
--- a/changelogs/unreleased/5302-padlar-small.md
+++ b/changelogs/unreleased/5302-padlar-small.md
@@ -1,0 +1,1 @@
+Gateway API: Adds an optional field to GatewayProvisioner CRD to watch resources in subset of namespaces. When unset, resources are watched in all the namespaces.
--- a/examples/contour/01-crds.yaml
+++ b/examples/contour/01-crds.yaml
@@ -1205,6 +1205,10 @@ spec:
                           Limits. More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/'
                         type: object
                     type: object
+                  watchNamespaces:
+                    description: WatchNamespaces restricts watching contour instances
+                      to a subset of namespaces.
+                    type: string
                 type: object
               envoy:
                 description: Envoy specifies deployment-time settings for the Envoy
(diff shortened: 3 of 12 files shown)


pnbrown:
As discussed in person, this will need to wait for #5214, but can be lined up now by basing the branch off the other PR.

skriss:
Thanks @padlar!

Once #5214 merges, we'll need to rebase this on upstream/main.

This will also need: 
-`make generate` to be run and the results committed
- unit testing
- a changelog file per https://github.com/projectcontour/contour/actions/runs/4755394119/jobs/8449483023?pr=5302

padlar:
> * `make generate` to be run and the results committed
I have run `make generate` bunch of times with no changes. Not sure why the CI thinks otherwise. I tried with the same go version v1.20.3 with no success.

skriss:
@padlar we've just merged #5214, so this should be ready for a rebase onto main. I'd suggest something like:

```
git fetch upstream main
git rebase -i upstream/main
<in the interactive rebase prompt, delete the lines for any commits that aren't yours>
<resolve any conflicts>
git push -f
```

apis/projectcontour/v1alpha1/contourdeployment.go
```diff
@@ -114,6 +114,10 @@ type ContourSettings struct {
 	// Deployment describes the settings for running contour as a `Deployment`.
 	// +optional
 	Deployment *DeploymentSettings `json:"deployment,omitempty"`
+
+	// WatchNamespaces restricts watching contour instances to a subset of namespaces.
```
sunjayBhatia, line 118: I think we want the documentation here and the release note to state that this field is for limiting the namespaces the *Contour instance* will watch for resource updates (and not what the Gateway Provisioner will watch)

skriss:
One thing I'm realizing we haven't been great about as we add more options to ContourDeployment, is keeping in mind this section from the Gateway API spec:

> It is recommended that [GatewayClass] be used as a template for Gateways. This means that a Gateway is based on the state of the GatewayClass at the time it was created and changes to the GatewayClass or associated parameters are not propagated down to existing Gateways. This recommendation is intended to limit the blast radius of changes to GatewayClass or associated parameters. If implementations choose to propagate GatewayClass changes to existing Gateways, that MUST be clearly documented by the implementation.

(ref. https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.GatewayClass)

So basically, any settings from the GatewayClass/ContourDeployment should be applied at the time of creating a Gateway's Kubernetes resources, but not subsequently updated if the GatewayClass/ContourDeployment change.

There is also https://github.com/kubernetes-sigs/gateway-api/pull/1868 (GEP) to consider, which has recently been merged upstream but has not yet been added to the API.

sunjayBhatia:
Hey @padlar looks like we'll have to move this to the next release milestone since 1.26.0 is coming up

We'll have to discuss how to handle @skriss' comment https://github.com/projectcontour/contour/pull/5302#issuecomment-1601579805 and how we want to move this one forward

Issue #5292, opened 2023-04-19
internal/envoy/v3: add test coverage for SlowStartConfig with Least Request LB policy


skriss:
[Slow start config has test coverage for when used with a round robin LB policy](https://github.com/projectcontour/contour/blob/main/internal/envoy/v3/cluster_test.go#L577-L610), however coverage is missing for when used with [least request LB policy](https://github.com/projectcontour/contour/blob/main/internal/envoy/v3/cluster.go#L119-L124). Add a test case similar to the existing one, for a cluster with a "least request" LB policy.

tsaarni:
Tip: To check the coverage results, run `make check-coverage` and open file `coverage.html`

owayss:
working on this

PR #5303, opened 2023-04-20
internal/envoy/v3: adds a testcase for slow start mode with non-default LB policy

owayss:
internal/envoy/v3: adds a testcase for slow start mode with non-default LB policy.

fixes https://github.com/projectcontour/contour/issues/5292


skriss:
Thanks @owayss! Will review this as soon as we can and provide feedback or merge!

skriss (approved):
Thanks @owayss, LGTM! :shipit:

PR #5305, opened 2023-04-20
De-dupe ExtensionService resolution code

### LINKED ISSUE #5295: cmd/contour/serve.go: de-dupe ExtensionService resolution code
In cmd/contour/serve.go, we have three different ExtensionServices that are being resolved:
- Tracing
- Global rate limiting
- Global ExtAuth

There is some code duplication between these that would be good to extract a helper function for. Specifically, the following blocks of code all do the same thing:
- https://github.com/projectcontour/contour/blob/main/cmd/contour/serve.go#L675-L702
- https://github.com/projectcontour/contour/blob/main/cmd/contour/serve.go#L720-L748
- https://github.com/projectcontour/contour/blob/main/cmd/contour/serve.go#L606-L631

Extract a helper for this logic that can be reused across the three `setupXXX` methods.

shayyxi:
Closes #5295.

### DIFF
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -638,36 +638,40 @@ func (s *Server) doServe() error {
 	return s.mgr.Start(signals.SetupSignalHandler())
 }
 
-func (s *Server) setupTracingService(tracingConfig *contour_api_v1alpha1.TracingConfig) (*xdscache_v3.TracingConfig, error) {
-	if tracingConfig == nil {
-		return nil, nil
-	}
-
-	// ensure the specified ExtensionService exists
+func getExtensionSvcProperties(name string, namespace string) (client.ObjectKey, timeout.Setting, string, error) {
 	extensionSvc := &contour_api_v1alpha1.ExtensionService{}
 	key := client.ObjectKey{
-		Namespace: tracingConfig.ExtensionService.Namespace,
-		Name:      tracingConfig.ExtensionService.Name,
+		Namespace: namespace,
+		Name:      name,
 	}
-	// Using GetAPIReader() here because the manager's caches won't be started yet,
-	// so reads from the manager's client (which uses the caches for reads) will fail.
-	if err := s.mgr.GetAPIReader().Get(context.Background(), key, extensionSvc); err != nil {
-		return nil, fmt.Errorf("error getting tracing extension service %s: %v", key, err)
-	}
-	// get the response timeout from the ExtensionService
+
 	var responseTimeout timeout.Setting
 	var err error
 
 	if tp := extensionSvc.Spec.TimeoutPolicy; tp != nil {
 		responseTimeout, err = timeout.Parse(tp.Response)
 		if err != nil {
-			return nil, fmt.Errorf("error parsing tracing extension service %s response timeout: %v", key, err)
+			return client.ObjectKey{}, timeout.Setting{}, "", fmt.Errorf("error parsing tracing extension service %s response timeout: %v", key, err)
 		}
 	}
 
 	var sni string
 	if extensionSvc.Spec.UpstreamValidation != nil {
 		sni = extensionSvc.Spec.UpstreamValidation.SubjectName
+	}
+
+	return key, responseTimeout, sni, nil
+}
+
+func (s *Server) setupTracingService(tracingConfig *contour_api_v1alpha1.TracingConfig) (*xdscache_v3.TracingConfig, error) {
+	if tracingConfig == nil {
+		return nil, nil
+	}
+
+	// ensure the specified ExtensionService exists
+	key, responseTimeout, sni, err := getExtensionSvcProperties(tracingConfig.ExtensionService.Name, tracingConfig.ExtensionService.Namespace)
+	if err != nil {
+		return nil, err
 	}
 
 	var customTags []*xdscache_v3.CustomTag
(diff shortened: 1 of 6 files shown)


skriss:
Thanks for the PR @shayyxi!

My one suggestion here: if we create an `xdscache_v3.ExtensionServiceConfig` struct that contains these three fields (key, timeout, SNI), and modify the existing `xdscache_v3` structs (TracingConfig, RateLimitConfig, GlobalExternalAuthConfig) to embed an `xdscache_v3.ExtensionServiceConfig`, then we can change the signature of the new helper to return a `xdscache_v3.ExtensionServiceConfig` instead of three separate fields:

```go
func getExtensionSvcProperties(name string, namespace string) (*xdscache_v3.ExtensionServiceConfig, error)
```

As you pointed out in the ContribFest, that many return parameters is a little bit challenging, and they are all used in the same place so are pretty cohesive, so I think this change would encapsulate them better.

Let me know if you need any more clarity on that refactor.

shayyxi:
Hi. Just to update. I would resume my work on this PR after kubecon from next week. :-) Thanks for an amazing session once again.

skriss:
> Hi. Just to update. I would resume my work on this PR after kubecon from next week. :-) Thanks for an amazing session once again.

@shayyxi sounds good, will keep an eye out. Thanks for attending and participating!

shayyxi:
Just an update. After the Kubecon I am catching up with my work so couldn't manage to work on it. Will surely start working on it either by tomorrow otherwise from Monday.

sunjayBhatia (commented):
hey @shayyxi just a friendly check-in to see if you need any more support on this change? thanks for taking the time to contribute!

shayyxi:
@sunjayBhatia Hey Sunjay! Thanks for asking. It is very clear to me so far what needs to be done. I just got busy with job tasks. I will be resuming this PR from today or tomorrow. And will try to finalize it within next week.

shayyxi:
@skriss Based on my understanding of your comment. I have done the changes. Please review them!

shayyxi:
@skriss That's the thing I was little bit worried about while implementing the suggestion you gave in the review comment. This change is causing a domino like effect as one struct change is causing alot of changes in the methods where that struct is being used. Am I doing something wrong in the implementation?

Or would you agree to stick to these 3 values return?

Or should we still progress with this refactoring and do the changes everywhere needed?

skriss:
@shayyxi I'll have to look a little more closely, but I think that if you _embed_ the new struct rather than defining it as a field, then it should be mostly compatible for readers, e.g.:

```go
type TracingConfig struct {
	ExtensionServiceConfig
        ...
}
```

shayyxi:
@skriss I tried that already but that didn’t work too. However to be really sure I would give that one more try.

shayyxi:
In my last comment, “that didn’t work” means that I still had to change the usage of struct in some methods because of incompatibility between fields.

skriss:
@shayyxi I was thinking something like this: https://github.com/skriss/contour/commit/05d069655a64633cacff3ff840898aee97ad3fa1

It compiles and unit tests pass (haven't run E2E's yet but I think they should be fine). Let me know what you think

cmd/contour/serve.go
```diff
@@ -638,30 +638,20 @@ func (s *Server) doServe() error {
...
-	// get the response timeout from the ExtensionService
+
 	var responseTimeout timeout.Setting
 	var err error
 
 	if tp := extensionSvc.Spec.TimeoutPolicy; tp != nil {
 		responseTimeout, err = timeout.Parse(tp.Response)
 		if err != nil {
-			return nil, fmt.Errorf("error parsing tracing extension service %s response timeout: %v", key, err)
+			return &xdscache_v3.ExtensionServiceConfig{}, fmt.Errorf("error parsing tracing extension service %s response timeout: %v", key, err)
```
skriss, line 654:
```suggestion
			return &xdscache_v3.ExtensionServiceConfig{}, fmt.Errorf("error parsing extension service %s response timeout: %v", key, err)
```

cmd/contour/serve.go
```diff
@@ -638,30 +638,20 @@ func (s *Server) doServe() error {
...
-		Namespace: tracingConfig.ExtensionService.Namespace,
-		Name:      tracingConfig.ExtensionService.Name,
+		Namespace: namespace,
+		Name:      name,
 	}
-	// Using GetAPIReader() here because the manager's caches won't be started yet,
-	// so reads from the manager's client (which uses the caches for reads) will fail.
-	if err := s.mgr.GetAPIReader().Get(context.Background(), key, extensionSvc); err != nil {
-		return nil, fmt.Errorf("error getting tracing extension service %s: %v", key, err)
-	}
```
skriss, line 656: This block of code needs to be included in `getExtensionSvcProperties` as well. You could either pass in the APIReader as an argument for the function to use, or change `getExtensionSvcProperties` to be a method on `s` so that it can access `s.mgr` directly. I'd be fine with either approach.

shayyxi:
Thanks for the quick suggestions @skriss 
I will resume the work tomorrow and will get back with my changes.

shayyxi:
Actually, I saw these tests in the build failure log. But I wasn't sure if such deep changes would be feasible in a refactoring (which seemed simple and small 😆).

shayyxi:
@skriss @sunjayBhatia can you please provide an approval to run build workflows?

internal/featuretests/v3/global_authorization_test.go
```diff
@@ -14,6 +14,7 @@
 package v3
 
 import (
+	"github.com/projectcontour/contour/internal/k8s"
```
skriss, line 17: This needs to move back down to the below import group where it was previously

internal/xdscache/v3/listener.go
```diff
@@ -14,6 +14,7 @@
 package v3
 
 import (
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
skriss, line 17: Move down to below import group

cmd/contour/serve.go
```diff
@@ -638,30 +638,26 @@ func (s *Server) doServe() error {
...
-		Name:      tracingConfig.ExtensionService.Name,
+		Namespace: namespace,
+		Name:      name,
 	}
+
 	// Using GetAPIReader() here because the manager's caches won't be started yet,
 	// so reads from the manager's client (which uses the caches for reads) will fail.
 	if err := s.mgr.GetAPIReader().Get(context.Background(), key, extensionSvc); err != nil {
-		return nil, fmt.Errorf("error getting tracing extension service %s: %v", key, err)
+		return xdscache_v3.ExtensionServiceConfig{}, fmt.Errorf("error parsing extension service %s: %v", key, err)
```
skriss, line 651:
```suggestion
		return xdscache_v3.ExtensionServiceConfig{}, fmt.Errorf("error getting extension service %s: %v", key, err)
```

skriss (changes requested):
Thanks for the updates @shayyxi, just a couple minor things

skriss (changes requested):
Thanks for the updates @shayyxi, just a couple minor things

shayyxi:
Done with fixing review comments. Please approve for workflows @skriss 😅

cmd/contour/serve.go
skriss: Looks like you changed the wrong one; this one should be "error getting..." and the next one (for response timeout) should be "error parsing..."

shayyxi:
In this particular PR/case, are those warnings given by codecov/patch relevant?

If not, then isn't there any mechanism to somehow make them intelligent or suppress them for specific lines?

cmd/contour/serve.go
shayyxi: doing this...

shayyxi:
@skriss Done with the last comment....

skriss (approved):
Thanks @shayyxi, LGTM!

internal/xdscache/v3/listener.go
```diff
@@ -134,15 +134,17 @@ type ListenerConfig struct {
 	TracingConfig *TracingConfig
 }
 
+type ExtensionServiceConfig struct {
+	ExtensionService client.ObjectKey
```
sunjayBhatia, line 138: we could probably change this (back) to be a NamespacedName (even though ObjectKey is an alias to it) just to limit import blast radius etc.

sunjayBhatia: (so we don't import controller-runtime packages into this one unnecessarily for example)

shayyxi: Looking into it....

shayyxi: Fixed...

sunjayBhatia (approved):

shayyxi:
@skriss can we please merge this PR?

PR #5307, opened 2023-04-21
chore: extend tracing test for custom tags

### LINKED ISSUE #5291: tracing: add feature test for custom tags
Tracing has a [feature test](https://github.com/projectcontour/contour/blob/main/internal/featuretests/v3/tracing_test.go), however it is missing test coverage for [custom tags](https://github.com/projectcontour/contour/blob/main/internal/xdscache/v3/listener.go#L150). Coverage should be added in the feature test that ensures custom tags specified in the config are properly set on the tracing filter.

This could be done by adding some custom tags in [the existing TracingConfig](https://github.com/projectcontour/contour/blob/main/internal/featuretests/v3/tracing_test.go#L34-L40), and then verifying that the [emitted tracing filter config](https://github.com/projectcontour/contour/blob/main/internal/featuretests/v3/tracing_test.go#L105-L111) has the appropriate tags set.

padlar:
fixes #5291

### DIFF
--- a/internal/featuretests/v3/tracing_test.go
+++ b/internal/featuretests/v3/tracing_test.go
@@ -37,6 +37,20 @@ func TestTracing(t *testing.T) {
 		Timeout:          timeout.DefaultSetting(),
 		OverallSampling:  100,
 		MaxPathTagLength: 256,
+		CustomTags: []*xdscache_v3.CustomTag{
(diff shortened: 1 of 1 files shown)


skriss (approved):
LGTM, thanks @padlar!

PR #5310, opened 2023-04-24
support query param regex matching in Gateway API

### LINKED ISSUE #5262: Gateway API: support query param regex matching
As a follow-up to https://github.com/projectcontour/contour/pull/5239, [query parameter matching also allows for RegularExpression matches as custom conformance](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.QueryParamMatchType). It would be great to add support for this.

padlar:
fixes #5262

### DIFF
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -1673,11 +1673,11 @@ func gatewayQueryParamMatchConditions(matches []gatewayapi_v1beta1.HTTPQueryPara
 			matchType = HeaderMatchTypeExact
 		case gatewayapi_v1beta1.QueryParamMatchRegularExpression:
 			if err := ValidateRegex(match.Value); err != nil {
-				return nil, fmt.Errorf("FIXME: Invalid value for RegularExpression match type is specified")
+				return nil, fmt.Errorf("HTTPRoute.Spec.Rules.Matches.QueryParams: Invalid value for RegularExpression match type is specified")
 			}
 			matchType = HeaderMatchTypeRegex
 		default:
-			return nil, fmt.Errorf("FIXME: Only support Exact or RegularExpression match types")
+			return nil, fmt.Errorf("HTTPRoute.Spec.Rules.Matches.QueryParams: Only Exact and RegularExpression match types are supported")
 		}
 
 		// If multiple match conditions are found for the same value,
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -1667,10 +1667,17 @@ func gatewayQueryParamMatchConditions(matches []gatewayapi_v1beta1.HTTPQueryPara
 	seenNames := sets.New[string]()
 
 	for _, match := range matches {
-		// "Exact" is the default if not defined in the object, and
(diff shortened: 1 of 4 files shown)


skriss:
Thanks @padlar! Will review soon.

internal/dag/gatewayapi_processor.go
```diff
@@ -1667,10 +1667,17 @@ func gatewayQueryParamMatchConditions(matches []gatewayapi_v1beta1.HTTPQueryPara
...
-		// the only supported match type.
-		if match.Type != nil && *match.Type != gatewayapi_v1beta1.QueryParamMatchExact {
-			return nil, fmt.Errorf("HTTPRoute.Spec.Rules.Matches.QueryParams: Only Exact match type is supported")
+		var matchType string
+		switch ref.Val(match.Type, gatewayapi_v1beta1.QueryParamMatchExact) {
+		case gatewayapi_v1beta1.QueryParamMatchExact:
+			matchType = HeaderMatchTypeExact
+		case gatewayapi_v1beta1.QueryParamMatchRegularExpression:
+			if err := ValidateRegex(match.Value); err != nil {
+				return nil, fmt.Errorf("FIXME: Invalid value for RegularExpression match type is specified")
```
skriss, line 1676:
```suggestion
				return nil, fmt.Errorf("HTTPRoute.Spec.Rules.Matches.QueryParams: Invalid value for RegularExpression match type is specified")
```

internal/dag/gatewayapi_processor.go
```diff
@@ -1667,10 +1667,17 @@ func gatewayQueryParamMatchConditions(matches []gatewayapi_v1beta1.HTTPQueryPara
...
+			}
+			matchType = HeaderMatchTypeRegex
+		default:
+			return nil, fmt.Errorf("FIXME: Only support Exact or RegularExpression match types")
```
skriss, line 1680:
```suggestion
			return nil, fmt.Errorf("HTTPRoute.Spec.Rules.Matches.QueryParams: Only Exact and RegularExpression match types are supported")
```

skriss, line 1680: Let's add a test case to status_test.go that covers this branch (unsupported query param match type)

skriss (changes requested):
Thanks @padlar! This is looking pretty good, just a few comments. Please also add some test cases to internal/dag/builder_test.go in the `TestDAGInsertGatewayAPI` test -- you should be able to find some existing HTTPRoute query param tests to base the new test cases off of.

internal/dag/gatewayapi_processor.go
padlar: Added few more test cases

skriss:
Thanks for the updates @padlar. You'll also need to add a changelog file, per https://github.com/projectcontour/contour/actions/runs/4824508057/jobs/8594362044?pr=5310 and https://github.com/projectcontour/contour/blob/main/CONTRIBUTING.md#commit-message-and-pr-guidelines, so users can read about this in the release notes!

skriss (approved):
LGTM, thanks @padlar!

Issue #5311, opened 2023-04-24
HTTPProxy: Add support for regex based path routing and Request Header routing


clayton-gonsalves:
**Please describe the problem you have**
The HTTProxy currently allows regex based routing for request query params. This should be extended to also paths and request headers.

PR #5319, opened 2023-04-26
HTTPProxy: add regex support for path matching

clayton-gonsalves:
Fixes: #5311


clayton-gonsalves:
@skriss @sunjayBhatia This is ready for review. Thanks

sunjayBhatia (commented):
Overall looks great! just some minor comments so far

One more minor unit test to add would be to have an include prefix with a trailing slash and a regex leaf match condition just to get coverage over that slash de-duplication (e.g. prefix include match prefix`/api/` and route match `/v[0-9]+)`)

Might want to make sure our strategy of merging slashes doesn't interfere with any possible valid regexes as well: https://github.com/projectcontour/contour/blob/9bcc69e20cab473ea5d9aa61f56d248c38e4f406/internal/dag/conditions.go#L48-L49 (I don't *think* it will but worth another look)

internal/dag/conditions_test.go
```diff
@@ -824,3 +872,46 @@ func TestValidateQueryParameterMatchConditions(t *testing.T) {
...
+		},
+		"Regex and Prefix conditions set": {
+			matchconditions: []contour_api_v1.MatchCondition{{
+				Regex: "/.*/api",
+			}, {
+				Prefix: "/v1",
+			}},
+			want: false,
+		},
+		"invalid regex": {
```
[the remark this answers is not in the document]
clayton-gonsalves: 👍

internal/dag/builder_test.go
```diff
@@ -13884,6 +13964,51 @@ func TestDAGInsert(t *testing.T) {
...
+				proxyRegexPath, child1, child2, s1, s2,
+			},
+			want: listeners(
+				&Listener{
+					Name: HTTP_LISTENER_NAME,
+					Port: 8080,
+					VirtualHosts: virtualhosts(
+						virtualhost("example.valid.regexpath.com",
+							routeClusterRegex("/foo/regex/.*",
+								&Cluster{
```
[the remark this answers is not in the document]
clayton-gonsalves: added the `service` but `clusterweight()` didnt match

sunjayBhatia (changes requested):
just the one thing to fix but otherwise lgtm

sunjayBhatia (approved):

apis/projectcontour/v1/httpproxy.go
```diff
@@ -54,14 +54,14 @@ type Include struct {
 	// structs.
 	// When applied, they are merged using AND, with one exception:
 	// There can be only one Prefix MatchCondition per Conditions slice.
-	// More than one Prefix, or contradictory Conditions, will make the
+	// More than one Prefix or Regex, or contradictory Conditions, will make the
```
skriss, line 57: regex match isn't allowed for includes, so can drop this comment change.

internal/dag/conditions.go
```diff
@@ -105,6 +126,9 @@ func includeMatchConditionsValid(conds []contour_api_v1.MatchCondition) error {
 		if cond.Exact != "" {
 			return fmt.Errorf("exact conditions are not allowed in includes block")
 		}
+		if cond.Regex != "" {
+			return fmt.Errorf("regex conditions are not allowed in an included HTTPProxy")
```
skriss, line 130: I think it'd be a little more clear to use similar wording as for exact match above, since this is about not allowing regex in the include conditions themselves, not in the conditions within the included HTTPProxy's routes:
```suggestion
			return fmt.Errorf("regex conditions are not allowed in includes block")
```

site/content/docs/main/config/request-routing.md
```diff
@@ -76,7 +76,7 @@ spec:
 
 Each Route entry in a HTTPProxy **may** contain one or more conditions.
 These conditions are combined with an AND operator on the route passed to Envoy.
-Conditions can be either a `prefix`, `exact`, `header` or a `queryParameter` condition. `prefix` and `exact`
+Conditions can be either a `prefix`, `exact`, `regex`, `header` or a `queryParameter` condition. `prefix` and `exact`
 conditions cannot be used together in one condition block.
```
skriss, line 80:
```suggestion
Conditions can be either a `prefix`, `exact`, `regex`, `header` or a `queryParameter` condition. At most one of `prefix`, `exact` or `regex` can be used in one condition block.
```

skriss (changes requested):
Thanks @clayton-gonsalves! Just a couple small things but overall looks great.

skriss:
Looks like this also closes https://github.com/projectcontour/contour/issues/1179

clayton-gonsalves:
@skriss, the osx ci step fails, but looking into it, I am unsure if it's related.

apis/projectcontour/v1/httpproxy.go
skriss, line 57:
```suggestion
	// More than one Prefix, or contradictory Conditions, will make the
```

Alternately, you could add some text here also describing that exact & regex are not supported at all for an `Include`, but that's already covered on the fields in `MatchCondition` below so I don't think it's strictly necessary.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -481,7 +491,7 @@ type Route struct {
 	// Conditions are a set of rules that are applied to a Route.
 	// When applied, they are merged using AND, with one exception:
 	// There can be only one Prefix MatchCondition per Conditions slice.
-	// More than one Prefix, or contradictory Conditions, will make the
+	// More than one Prefix or Regex, or contradictory Conditions, will make the
```
skriss, line 494: Not directly related to your PR, but noticed we're missing exact here:
```suggestion
	// More than one Prefix, Exact or Regex, or contradictory Conditions, will make the
```

skriss:
> @skriss, the osx ci step fails, but looking into it, I am unsure if it's related.

It flakes sometimes, I re-ran it.

skriss (approved):
We can merge and follow up with some godoc tweaks. Thanks again @clayton-gonsalves!

skriss:
ref. https://github.com/projectcontour/contour/pull/5413 for a few tweaks to the godoc.

PR #5323, opened 2023-04-27
test/e2e/incluster: Add memory usage test for header includes

sunjayBhatia:
Tests that Contour reconciles a root properly that has many children with header match include conditions and that the Contour pods are not restarted (which would happen due to OOM since we set a memory limit).

Cherry-picking this commit against the release-1.24 branch and it fails (the root HTTPProxy never is valid in the runs I have done, but watching the container restarts, there are a few due to OOM)


skriss (approved):
Verified on my dev env it fails when cherry-picked to release-1.24, and passes with the fix in place. I like this approach! We could use this for some more CPU/memory regression tests.

PR #5331, opened 2023-04-28
Gateway provisioner: add field to opt out of reconciling optional CRDs

### LINKED ISSUE #5276: Gateway Provisioner: Add ability to opt-out of reconciling optional CRDs
As a follow up to https://github.com/projectcontour/contour/pull/5257, user of the Gateway Provisioner should have the ability to provision contour instances that have disabled reconciliation of some optional CRDs (especially those that are "experimental" channel Gateway API CRDs). We can probably also remove the RBAC rules to disable access to these CRDs for installations that have opted out.

It seems like this should be configurable via a new field on the ContourDeployment CRD so that this setting can be applied to all contour instances for a particular GatewayClass.

padlar:
fixes #5276


skriss:
https://github.com/projectcontour/contour/pull/5302#issuecomment-1601579805 applies here as well.

PR #5334, opened 2023-05-01
fix upgrade from version for RC tags

### LINKED ISSUE #5328: get-contour-upgrade-from-version.sh doesn't work for RC tags
ref. https://github.com/projectcontour/contour/actions/runs/4832306037/jobs/8610980307

It's hitting this case: https://github.com/projectcontour/contour/blob/main/test/scripts/get-contour-upgrade-from-version.sh#L7-L9

Problem is the first `grep` is filtering out the current tag, so then the second `grep` doesn't find anything. We want to get the first non-prerelease tag _before_ the current (RC) tag.

skriss:
Closes #5328.

Ran this logic on a bunch of different pre-release and release tags and all looks good:

```
v1.25.0-rc.2 -> v1.24.4
v1.25.0-rc.1 -> v1.24.4
v1.24.4 -> v1.24.3
v1.24.0 -> v1.23.5
```

### DIFF
--- a/test/scripts/get-contour-upgrade-from-version.sh
+++ b/test/scripts/get-contour-upgrade-from-version.sh
@@ -5,8 +5,13 @@ set -o nounset
 set -o pipefail
 
 if CURRENT_TAG=$(git describe --tags --exact-match 2>/dev/null); then
-  # We are on a tag, so find previous tag to this one.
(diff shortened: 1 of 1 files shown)


sunjayBhatia (approved):
tried it hopefully this is the one 🤞🏽

PR #5347, opened 2023-05-08
update to Go 1.20.4

skriss:


Makefile
```diff
@@ -41,7 +41,7 @@ endif
 IMAGE_PLATFORMS ?= linux/amd64,linux/arm64
 
 # Base build image to use.
-BUILD_BASE_IMAGE ?= golang:1.20.3
+BUILD_BASE_IMAGE ?= golang:1.20.4
```
skriss, line 44: Wondering if we should think about switching to e.g. `golang:1.20` to get patch version bumps for free, would reduce the toil here a bit.

PR #5352, opened 2023-05-10
Added config for socket options for listeners

### LINKED ISSUE #4605: Make configurable Envoy listener socketOptions to allow DSCP marking for QoS purposes
Hi Ladies and Gents,

DSCP (differentiated services code point) is a mechanism used for classifying/prioritizing network traffic on IP networks [DSCP_wiki](https://en.wikipedia.org/wiki/Differentiated_services). Since Contour is used as O&M external interface provider for microservices at our customer, request arrived to support DSCP marking. Therefore all microservice O&M HTTP interface DSCP marking can be done in a unified place.

DiffServ uses 6-bit differentiated services code point (DSCP) in the 8-bit differentiated services field (DS field) in the IP header for packet classification purposes (last 2 bits are reserved - CU). Please see picture with details:

image2021-11-24_14-39-42

Example Envoy IPv4 listener configuration to set CS2 selection via Envoy [Socket Option](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/socket_option.proto)

```
static_resources:
  listeners:
  - name: listener_0
    address:
      socket_address: { address: 100.109.96.4, port_value: 10000 }
    socket_options:
    - level: 0           #  0 for IPPROTO_IP
      name: 1            #  1 for IP_TOS
      int_value: 64
      state: STATE_LISTENING
```
int_value 64 will be converted into 8 bits: 0 1 0 0 0 0 0 0 and used to fill the given IP header section of DS according to the picture.
Since the last 2 bits are reserved, this will end up in configuring 0 1 0 0 0 0  as Class Selector that is CS2 according to [Class Selector mapping](https://www.ccexpert.us/traffic-shaping-2/the-class-selector-phb-and-dscp-values.html)

This Class Selector can also be captured in IP packets received from Envoy:
wireshark

Captured packets:
[dscp_CS2_settings.pcapng.tar.gz](https://github.com/projectcontour/contour/files/9019120/dscp_CS2_settings.pcapng.tar.gz)

Example Envoy IPv6 listener configuration:
```
        socket_options:   #  this values need to be double checked!
        - level: 41        #  41 for IPPROTO_IPv6
          name: 67         #  67 for IPV6_TCLASS
          int_value: 32 
          state: STATE_LISTENING 
```

Example to configure both IPv4 and IPv6 with different DSCP values:
```
static_resources:
  listeners:
  - name: listener_0
    address:
      socket_address: { address: "::", port_value: 10000, ipv4_compat: true }
    socket_options:
    - level: 0
      name: 1
      int_value: 32
      state: STATE_LISTENING
    - level: 41
      name: 67
      int_value: 64
      state: STATE_LISTENING
```

**Reproduction steps (for Ubuntu):**
For reproduction purposes the following configuration of Envoy can be used:
[base_envoy_config_socket_options_CS2.txt](https://github.com/projectcontour/contour/files/9018977/base_envoy_config_socket_options_CS2.txt)

Create a kind network:
```
docker network create -d=bridge -o com.docker.network.bridge.enable_ip_masquerade=true -o com.docker.network.driver.mtu=1400 --opt com.docker.network.bridge.name=docker1 --subnet 100.109.100.0/20 kind
```

Start Envoy container:
```
docker run -it --network kind --ip 100.109.96.4 --mount type=bind,source=<PLACEOFTHECONFIGURATIONFILE>base_envoy_config_socket_options_CS2.txt,target=/config.yaml envoyproxy/envoy-dev:latest /usr/local/bin/envoy -c config.yaml --service-cluster front-proxy
```
Copy go, [myserver.go.tar.gz](https://github.com/projectcontour/contour/files/9019199/myserver.go.tar.gz) (simple echo application) into container, install tcpdump and start myserver.go:
```
export ENVOY_CONTAINER=$(docker container ls -l -q)
docker cp /usr/local/go $ENVOY_CONTAINER:/usr/local/
docker cp myserver.go $ENVOY_CONTAINER:/tmp
docker exec -it $ENVOY_CONTAINER bash
export PATH=$PATH:/usr/local/go/bin
cd /tmp
apt-get update
apt-get install -y tcpdump
go run myserver.go
```
Start tcpdump in another shell:
```
export ENVOY_CONTAINER=$(docker container ls -l -q)
docker exec -it $ENVOY_CONTAINER bash
cd /tmp
tcpdump -i any -w dscp
```
Create direct entry in /etc/hosts file that targets the Envoy container:
```
100.109.96.4   primary.com
```
In another shell execute curl command:
```
curl primary.com:10000
```
Results can be copied and analyzed:
```
export ENVOY_CONTAINER=$(docker container ls -l -q)
docker cp $ENVOY_CONTAINER:/tmp/dscp .
```

Other notes:
- Request would be to make configurable int_value via Contour for IPv4 and IPv6 use cases as well
- Different Class Selectors shall be able to be configured on the same listener for IPv4 and IPv6
- Configurability for specific Class Selectors (e.g.: only to support CS1, CS2 and CS3) shall not be limited, since the Class Selectors can be changed in the future.
- Configuring int_value 32 ended up in 0 0 1 0 0 0 as DSCP value that verified to be Class Selector 1 by tcpdump file
- IPv6 listener configuration parameters haven't been verified yet. Once this feature request accepted, can be tried out as well.
- TCP KeepAlive is also implemented via socketOption: [2638](https://github.com/projectcontour/contour/pull/2638), might be interesting to have some general function to configure socket options

tsaarni:
This PR adds new config field to support DSCP marking for outbound IP packets, for both IPv4 (TOS field) and IPv6 (Traffic Class field). This is done using existing `listener` stanza:

Example config file:

```yaml
listener:
  socket-options:
   tos: 64
   traffic-class: 64
```

Example ContourConfiguration CR

```yaml
kind: ContourConfiguration
metadata:
  name: contour
  namespace: projectcontour
spec:
  envoy:
   listener:
    socketOptions:
     tos: 64
     trafficClass: 64
```

The DSCP field is the 6 most significant bits of the exposed fields. However, the PR proposes exposing the socket options as they are exposed in Envoy and the socket API (complete TOS and TrafficClass bytes). For example, to set DSCP value `16` user needs to bit shift the value to 6 most significant bits of the byte and set the `tos` and/or `trafficClass` to value `64`.

The PR proposes moving the existing TCP keepalive options into a generic socket option file. If more socket options are added in the future, the new file would be the place to add them - regardless of being TCP or IP level options.

**Limitations:**

Since the two new socket options depend on IP version, the user needs to know in advance which one they want to set. 

For example, if the worker nodes where Envoy runs have IPv6 support enabled in the kernel, and Envoy was configured to listen to IPv6 address, then `traffic-class` can be set. If not, setting `traffic-class` will cause failure to configure listeners to Envoy.  Envoy will call `setsockopt()` to set IPv6 socket option on a socket that does not support IPv6, the syscall returns error and Envoy rejects the whole listener configuration. Contour cannot validate in advance if this syscall will succeed on the nodes where Envoy runs, so unfortunately it will be runtime failure.

The documentation for the config has a note about this limitation.

Fixes #4605


tsaarni:
I've now resolved merge conflicts that had appeared during recent changes in `main`. The PR is now again ready for review.

tsaarni:
Catching up with some more merge conflicts. These changes seem to be on a path that is bit prone to merge conflicts over the time, so reviews would be appreciated! :smile: :pray:

pkg/config/parameters.go
```diff
@@ -499,6 +502,37 @@ func (p *ListenerParameters) Validate() error {
 		return fmt.Errorf("invalid per connections buffer limit bytes value %q set on listener, minimum value is 1", *p.PerConnectionBufferLimitBytes)
 	}
 
+	return p.SocketOptions.Validate()
+}
+
+// SocketOptions defines configurable socket options for Envoy listeners.
+type SocketOptions struct {
```
sunjayBhatia, line 509: (w/o knowing the details of TOS vs. Traffic Class) is it valid to auto detect the Listener ip family to be able to collapse these two into one field?

sunjayBhatia (commented):
overall looks correct from a quick pass over the code, seems straightforward enough to set some socket options on the Envoy Listeners

should maybe add a featuretest in lieu of an e2e test to make sure config gets plumbed all the way through correctly

this looks like it will intersect a bit with https://github.com/projectcontour/contour/pull/5523 but I can sort that out if this merges first 👍🏽

pkg/config/parameters.go
tsaarni: Reliably auto-detecting could be challenging when listener address is set as `::`. We use [`Ipv4Compat: true`](https://github.com/projectcontour/contour/blob/984e42b99601f2a413708795f87ba780ba5f1a68/internal/envoy/v3/listener.go#L670) so Envoy accepts also IPv4 clients - unless IPv4 is disabled on the host where Envoy is running, I think.

tsaarni:
Thank You for review @sunjayBhatia!

> should maybe add a featuretest in lieu of an e2e test to make sure config gets plumbed all the way through correctly

So far, I've checked the values manually from IP header using Wireshark. It would be great to have e2e test, but I think it would require a test client that processes HTTP responses at IP packet level. Normally the values are processed only by the routers & host IP stacks, but the receiving application is unaware of the values.

sunjayBhatia:
> Thank You for review @sunjayBhatia!
> 
> > should maybe add a featuretest in lieu of an e2e test to make sure config gets plumbed all the way through correctly
> 
> So far, I've checked the values manually from IP header using Wireshark. It would be great to have e2e test, but I think it would require a test client that processes HTTP responses at IP packet level. Normally the values are processed only by the routers & host IP stacks, but the receiving application is unaware of the values.
> 
> Did you have some other approach in mind for e2e?

yeah totally agree, I think just an additional explicit test in internal/featuretest (unless I missed it is already there) for the new socket options added in this PR to make sure the config is generated properly when all the different internal components are hooked up should be sufficient on top of that validation you've done

pkg/config/parameters.go
sunjayBhatia: ah fair point, if we wanted to be clever could we maybe do something like "if listen address == :: then set the ipv4 and ipv6 options, otherwise set the appropriate option for the address family"?

sunjayBhatia: if in practice users may want to set different values between ipv4 and ipv6 seems more correct to keep these options/configurable values separate

sunjayBhatia (commented):
Looks like some conflicts to resolve and just to close out this thread https://github.com/projectcontour/contour/pull/5352#discussion_r1286254257

Otherwise 👍🏽

sunjayBhatia (approved):

Issue #5357, opened 2023-05-12
Set Global RateLimit Policy inside Contour's Config


shadialtarsha:
**Please describe the problem you have**
Currently, the `rateLimitPolicy` is configured on a per HTTPProxy level. While this allows for flexible config per service, it is also a drawback when we need to have the same config everywhere (global configs, single tenant use cases)
Contour should be able to have the option to specify a `ratelimitPolicy` on the contour configuration. This policy should be overridable by individual virtual hosts.

PR #5359, opened 2023-05-12
Design: General Global RateLimit Policy

shadialtarsha:
Design proposal to address https://github.com/projectcontour/contour/issues/5357

Signed-off-by: Shadi Altarsha [shadi.altarsha.94@gmail.com](mailto:shadi.altarsha.94@gmail.com)


design/general-global-rate-limit-policy.md
```diff
@@ -0,0 +1,100 @@
...
+## Non Goals
+- General rate limit policy for Route level `rateLimitPolicy`.
+- L4 rate limiting support. 
+
+## High-Level Design
+Currently, the global `rateLimitPolicy` for `VirtualHost` is set per HTTPProxy. This option allows each service to configure its own global `rateLimitPolicy`. However, for a single tenant setup/general usecase, this means you have to edit all the HTTPProxy objects with the same ratelimit policies.
+
+A new field, `GeneralRateLimitPolicy` (optional), will be added to `RateLimitServiceConfig` which is part of `ContourConfigurationSpec`. This field will define rate limit descriptors that will be added to every `VirtualHost`.
+
+HTTPProxy has to opt-in **explicitly** to use the general rate limit policy using `rateLimitPolicy.generalRateLimitPolicyEnabled` flag and still can have its own global `rateLimitPolicy` which overrides the general one.
```
skriss, line 23: I think it's going to be more consistent to do this the other way, i.e. to allow individual HTTPProxies to **opt out** of the general rate limit policy if it's defined. This is what we did with the recent global ExtAuth work, for example. Is there a specific reason you proposed using **opt-in** here?

design/general-global-rate-limit-policy.md
```diff
@@ -0,0 +1,100 @@
...
+- Define a general L7 global rate limit policy which is used by all HTTPProxy on the `VirtualHost` level.
+
+## Non Goals
+- General rate limit policy for Route level `rateLimitPolicy`.
+- L4 rate limiting support. 
+
+## High-Level Design
+Currently, the global `rateLimitPolicy` for `VirtualHost` is set per HTTPProxy. This option allows each service to configure its own global `rateLimitPolicy`. However, for a single tenant setup/general usecase, this means you have to edit all the HTTPProxy objects with the same ratelimit policies.
+
+A new field, `GeneralRateLimitPolicy` (optional), will be added to `RateLimitServiceConfig` which is part of `ContourConfigurationSpec`. This field will define rate limit descriptors that will be added to every `VirtualHost`.
```
skriss, line 21: Personally I'd prefer "default rate limit policy" as the name for this. Also need to decide if we want to include "global" as part of the name, to differentiate from local rate limiting which is not affected by this proposal. `DefaultGlobalRateLimitPolicy` ends up being a  pretty long name though; open to ideas here.

skriss, line 21: I'm also thinking about whether it'd make sense to put this new field under the `Policy` field in config which we document as "default policy applied if not overridden by the user", since it's not strictly a property of the rate limit service itself, and does logically make sense under `Policy`. The downside is that rate limit-related config is split across two separate places. Interested in others' thoughts here.

skriss (changes requested):
Thanks for the design doc @shadialtarsha! I had a few initial comments & thoughts.

design/general-global-rate-limit-policy.md
shadialtarsha: Ah!! thank you for choosing a good name for this 😄 
I will use that actually. I was struggling to come up with a better word.

design/general-global-rate-limit-policy.md
shadialtarsha: I was just thinking about not introducing a regression. That is it.

And then I was like If I am gonna add `DefaultGlobalRateLimitPolicy` I properly would like this policy to be applied directly without the need to edit all the HTTPProxy and opt them in manually.
So I don't really mind opt-in by default (in fact I would prefer this option) But yeah, the main motivation is just not introducing regression.

design/general-global-rate-limit-policy.md
shadialtarsha: Ahh, I didn't actually notice the `Policy` section, thanks for pointing it out. 
So from a developer's perspective, I would like to configure everything related to rate limiting in one place. It is a bit more intuitive to do it in the `rateLimitServiceConfig` that `Policy`
But no strong opinion on this.

design/general-global-rate-limit-policy.md
davinci26: +1 on `opt-out` as the default but I do see Shadi's point on breaking changes. What is the pattern here?

design/default-global-rate-limit-policy.md
```diff
@@ -0,0 +1,100 @@
...
+#### HTTPProxy
+```yaml
+apiVersion: projectcontour.io/v1
+kind: HTTPProxy
+metadata:
+  name: echo
+spec:
+  virtualhost:
+    fqdn: local.projectcontour.io
+    rateLimitPolicy:
```
davinci26, line 56: What would happen if we have different descriptors defined in `defaultGlobalRateLimitPolicy` and in the `HTTPProxy`?

design/default-global-rate-limit-policy.md
```diff
@@ -0,0 +1,100 @@
...
+### Contour Configuration changes
+
+A new field, `DefaultGlobalRateLimitPolicy` of type `GlobalRateLimitPolicy` (optional), will be added to `RateLimitServiceConfig` as part of `ContourConfigurationSpec`
+```go
+...
+type RateLimitService struct {
+	...
+	// DefaultGlobalRateLimitPolicy allows setting a default global rate limit policy for all HTTPProxy
+	// HTTPProxy can overwrite this configuration.
+	DefaultGlobalRateLimitPolicy contour_api_v1.GlobalRateLimitPolicy `yaml:"defaultGlobalRateLimitPolicy,omitempty"`
```
davinci26, line 77: We never define in the doc `contour_api_v1.GlobalRateLimitPolicy`? What fields does this struct have?

davinci26 (commented):
design overall makes sense to me, I would love some examples that would help me understand how we reconcile the  Global policy with the local.

Looking at the current implementation:

```
type RateLimitPolicy struct {
	Local *LocalRateLimitPolicy `json:"local,omitempty"`
	Global *GlobalRateLimitPolicy `json:"global,omitempty"`
}
```

Doesn't give much functionality so I wonder how we effectively override global options. Imagine the following scenarios:

* Can one service have it's own policy entirely? What would happen
* Can one service `fail_closed` whereas others `fail_open`

I am not suggesting that we should support all these scenarios but we should consider them and trying to see how they fit in

davinci26 (commented):
design overall makes sense to me, I would love some examples that would help me understand how we reconcile the  Global policy with the local.

Looking at the current implementation:

```
type RateLimitPolicy struct {
	Local *LocalRateLimitPolicy `json:"local,omitempty"`
	Global *GlobalRateLimitPolicy `json:"global,omitempty"`
}
```

Doesn't give much functionality so I wonder how we effectively override global options. Imagine the following scenarios:

* Can one service have it's own policy entirely? What would happen
* Can one service `fail_closed` whereas others `fail_open`

I am not suggesting that we should support all these scenarios but we should consider them and trying to see how they fit in

davinci26 (commented):
design overall makes sense to me, I would love some examples that would help me understand how we reconcile the  Global policy with the local.

Looking at the current implementation:

```
type RateLimitPolicy struct {
	Local *LocalRateLimitPolicy `json:"local,omitempty"`
	Global *GlobalRateLimitPolicy `json:"global,omitempty"`
}
```

Doesn't give much functionality so I wonder how we effectively override global options. Imagine the following scenarios:

* Can one service have it's own policy entirely? What would happen
* Can one service `fail_closed` whereas others `fail_open`

I am not suggesting that we should support all these scenarios but we should consider them and trying to see how they fit in

davinci26 (commented):
design overall makes sense to me, I would love some examples that would help me understand how we reconcile the  Global policy with the local.

Looking at the current implementation:

```
type RateLimitPolicy struct {
	Local *LocalRateLimitPolicy `json:"local,omitempty"`
	Global *GlobalRateLimitPolicy `json:"global,omitempty"`
}
```

Doesn't give much functionality so I wonder how we effectively override global options. Imagine the following scenarios:

* Can one service have it's own policy entirely? What would happen
* Can one service `fail_closed` whereas others `fail_open`

I am not suggesting that we should support all these scenarios but we should consider them and trying to see how they fit in

design/general-global-rate-limit-policy.md
skriss: In this case, since `DefaultGlobalRateLimitPolicy` will be a _new_ config field that is not populated by default, it should be backwards-compatible for folks who aren't using the feature, so I'd say having it apply to all HTTPProxies by default (i.e. opt-out behavior) if it is populated is fine and probably a better UX overall, as long as there's also a way to opt individual HTTPProxies out. cc @sunjayBhatia @tsaarni for other thoughts.

design/general-global-rate-limit-policy.md
skriss: Let's see if @sunjayBhatia @tsaarni have any thoughts here before we make a final decision.

design/default-global-rate-limit-policy.md
shadialtarsha: I added its definition.

shadialtarsha:
@davinci26 
I added some examples to explain how both policies can co-exist together.

Regarding your questions:
> Can one service have it's own policy entirely? What would happen?

Yes when a service defines its own global `rateLimitPolicy`, the default global one won't be considered at all!

> Can one service fail_closed whereas others fail_open

The failure mode is defined on the rate limit service configuration and not on the host. In theory, you should be able to do this by having multiple rateLimit service configurations and assigning them to hosts as you want, but I think Contour's configuration allows for only one `rateLimitService` configuration right now.

design/general-global-rate-limit-policy.md
sunjayBhatia: yep +1 to that ^ 

generally we've tried to have the global configs set in config file/CRD as operator owned defaults that can be overridden for individual HTTPProxies as needed so that should be consistent

design/general-global-rate-limit-policy.md
sunjayBhatia: seems like we're a little inconsistent here, for global extAuth we have `AuthorizationServer` which contains the `AuthPolicy` field and some others not specifically about the auth server

tsaarni: The generic `Policy` type has also field [`applyToIngress`](https://github.com/projectcontour/contour/blob/b5469efa7d6725a9e182fd226e333adc1337174e/apis/projectcontour/v1alpha1/contourconfig.go#L725-L729), which currently means that (header) policies are applied to `Ingresses`. If adding rate limiting to the type, should that too apply to `Ingresses`?

design/general-global-rate-limit-policy.md
shadialtarsha: Yeah I think the flag of opting out should be part of the Global rate limit policy for consistency

design/default-global-rate-limit-policy.md
```diff
@@ -0,0 +1,180 @@
...
+metadata:
+  name: contour
+  namespace: projectcontour
+data:
+  contour.yaml: |
+    rateLimitService:
+      extensionService: projectcontour/ratelimit
+      domain: contour
+      failOpen: false
+      defaultGlobalRateLimitPolicy:
```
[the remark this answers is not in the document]
shadialtarsha: I like the idea of having a flexible solution and it can be useful for sure, but the trade-offs of implementing `APPEND` vs. its benefits might not be balanced. 

Adding `APPEND` means we must maintain many cases for [actions](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-ratelimit-action) that users can set on Envoy. Some of these actions will generate descriptors with custom keys and values, other actions only support setting descriptors value, and Envoy sets the descriptor key. 

So yeah I am not really sure if we want to go this road.

shadialtarsha:
@skriss @tsaarni @sunjayBhatia Thank you all for your input. I made the changes and would appreciate another round.

sunjayBhatia (commented):
I think I would reverse the polarity on the new field on HTTPProxy to match other similar usage (AuthorizationPolicy, JWTVerificationPolicy), also since it makes defaulting a little easier you can just leave it unset to keep things as-is or set it to opt out (boolean default false vs. default true)

Otherwise seems reasonable to me 👍🏽

sunjayBhatia (approved):

shadialtarsha:
@skriss, @tsaarni, and @davinci26 can I get a final round, please?

davinci26 (approved):

skriss (approved):
LGTM, thanks @shadialtarsha!

PR #5362, opened 2023-05-15
Bump client-go, controller-runtime and related dependencies

nsimons:
Needed for #5214.

Some breaking changes have happened in these dependencies which I have addressed, biggest one being the new boolean parameter in ResourceEventHandler.OnAdd that can be used to know whether we are doing an initial sync or not. I believe we can pass `false` in order to retain old behavior, since we are not using it to know the sync status (_yet_). For the featuretests I created a simple wrapper to hide this parameter and make the change more manageable.

Keeping in draft until controller-runtime 0.15.0 stable version is released.

Signed-off-by: Niklas Simons <niklas.simons@est.tech>


cmd/contour/serve.go
```diff
@@ -217,32 +217,31 @@ func NewServer(log logrus.FieldLogger, ctx *serveContext) (*Server, error) {
...
-					// Other types of Secrets will never be referred to, so we can remove all data.
-					// For example Secrets of type helm.sh/release.v1 can be quite large.
-					// Last-applied-configuration annotation might contain a copy of the complete data.
-					secret.Data = nil
-					secret.SetManagedFields(nil)
-					secret.SetAnnotations(nil)
-
-					// Returning error will avoid handlers from being called.
-					// This does not avoid caching.
-					return nil, fmt.Errorf("ignoring secret %s/%s of type %s", secret.Namespace, secret.Name, secret.Type)
```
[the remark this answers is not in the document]
skriss: xref https://github.com/projectcontour/contour/pull/5366#issuecomment-1548660061 / https://github.com/projectcontour/contour/pull/5366#issuecomment-1550289190 -- agreed, seems like not returning an error is probably the way to go here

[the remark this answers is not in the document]
tsaarni: When the nillified secret is propagated to the handler, there might be risk it will attempt to access `secret.Data` as a map? I guess [this](https://github.com/projectcontour/contour/blob/9605b63d5c0aa9f1802a642c5ce5b1c9e298dc5a/internal/dag/secret.go#L39) check saves it this time, but would it be safer to use empty map instead of nil, so that it would pass as a valid (empty) Secret?

```diff
- secret.Data = nil
+ secret.Data = map[string][]byte{}
```

skriss (changes requested):
Thanks @nsimons! This mostly LGTM, just a couple tiny nits

internal/controller/controller_test.go
```diff
@@ -22,12 +22,11 @@ import (
 	"github.com/projectcontour/contour/internal/fixture"
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
+	"sigs.k8s.io/controller-runtime/pkg/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
-
 	// This package is deprecated and will be removed in future versions of
```
sunjayBhatia, line 27: can remove this comment too!

internal/provisioner/controller/gateway.go
```diff
@@ -133,7 +133,7 @@ func (r *gatewayReconciler) isGatewayClassReconcilable(obj client.Object) bool {
 	return accepted
 }
 
-func (r *gatewayReconciler) getGatewayClassGateways(gatewayClass client.Object) []reconcile.Request {
+func (r *gatewayReconciler) getGatewayClassGateways(_ context.Context, gatewayClass client.Object) []reconcile.Request {
```
sunjayBhatia, line 136: same comment on ctx getting passed through here too

internal/featuretests/v3/featuretests.go
```diff
@@ -66,7 +66,15 @@ const (
 	secretType   = resource.SecretType
 )
 
-func setup(t *testing.T, opts ...interface{}) (cache.ResourceEventHandler, *Contour, func()) {
+// ResourceEventHandlerWrapper wraps the ResourceEventHandler interface from client-go/tools/cache.
+// The OnAdd function has a boolean parameter that we do not need for testing.
+type ResourceEventHandlerWrapper interface {
```
sunjayBhatia, line 71: ah unfortunate client-go doesn't have this interface, though they would since they do have this: https://github.com/kubernetes/client-go/blob/4cb373f7caddf872f8dd0e0adbed3d0fe7c6aaf7/tools/cache/controller.go#L234-L238

internal/controller/controller_test.go
nsimons: Whoopsie!

internal/controller/gateway.go
```diff
@@ -121,7 +121,7 @@ func (r *gatewayReconciler) OnElectedLeader() {
 	}
 }
 
-func (r *gatewayReconciler) mapGatewayClassToGateways(gatewayClass client.Object) []reconcile.Request {
+func (r *gatewayReconciler) mapGatewayClassToGateways(_ context.Context, gatewayClass client.Object) []reconcile.Request {
 	var gateways gatewayapi_v1beta1.GatewayList
 	if err := r.client.List(context.Background(), &gateways); err != nil {
```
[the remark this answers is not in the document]
nsimons: Right, makes sense.

sunjayBhatia (approved):
LGTM, looks like all the relevant deps have been bumped in this PR

I'll add a follow up issue to see if we can take advantage of the new addition to the client-go `ResourceEventHandler` loop to make our processing of the initial resource list/startup better

internal/featuretests/v3/featuretests.go
tsaarni: @sunjayBhatia I wonder if this kind of abrupt non-backwards compatible change (the addition of `isInInitialList bool` parameter) in a hugely popular package could be a sign that client-go developers did not expect people to depend on `ResourceEventHandler` interface and redefine their own `OnAdd` etc methods directly? It seemed strange decision to break clients just because of new parameter like that.  Maybe they expected users only to instantiate `ResourceEventHandlerFuncs` struct which you point out, and not define their own like Contour does.

sunjayBhatia: I think the docs on the `*Funcs` structs imply theyre for usage when you don't need to implement all the methods, so I would expect the authors to know people will rely on the interface itself

there is also `*DetailedFuncs` here: https://github.com/kubernetes/client-go/blob/4cb373f7caddf872f8dd0e0adbed3d0fe7c6aaf7/tools/cache/controller.go#L264

regardless, unfortunate there was such a change w/o a major version revision

PR #5373, opened 2023-05-17
[wait for Contour deployment to update before upgrading Envoy

skriss:
(Testing in GHA)


sunjayBhatia (approved):
nice! yeah good change to make sure the data plane is fully rolled out (not just started rolling out) before starting the envoy roll out process

PR #5381, opened 2023-05-19
test/e2e: provisioner upgrade test

skriss:
Adds an upgrade test for Contour provisioned
by the Gateway provisioner.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -341,7 +341,7 @@ load-contour-image-kind: ## Load Contour image from building working source or p
 	./test/scripts/kind-load-contour-image.sh
 
 .PHONY: upgrade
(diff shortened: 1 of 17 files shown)


sunjayBhatia (approved):
nice!

tiny nit: maybe we could reuse some of the "create and wait for status" helpers that might currently exist unexported in another e2e test package, but not a huge deal tbh

skriss:
> tiny nit: maybe we could reuse some of the "create and wait for status" helpers that might currently exist unexported in another e2e test package, but not a huge deal tbh

Ah yeah, forgot I had copy-pasted those. Looks like we are proliferating them across `test/e2e/gateway`, `test/e2e/provisioner`, and now `test/e2e/upgrade`. I'll move them into a common file in just `test/e2e` for use across all the packages.

PR #5386, opened 2023-05-22
[provisioner] Add `ipFamilyPolicy` field to NetworkPublishing.

### LINKED ISSUE #5308: [provisioner] add a way to generate dualstack envoy service

Our setup actually requires to deploy envoy behind a Dual Stack LoadBalancer service.
While it is possible to create a `LoadBalancer` service by providing the Envoy Networking Type, Kubernetes defaults to create such `Service` with `ipFamilyPolicy` set to `SingleStack`.

To create a Dual Stack service, `ipFamilyPolicy` [must be explicitly](https://kubernetes.io/docs/concepts/services-networking/dual-stack/#service-type-loadbalancer) set to `PreferDualStack` or `RequireDualStack `, which is currently not possible.

Would it make sense to add such setting in `ContourDeployment.spec.envoy.networkPublishing` ?

Jean-Daniel:
Add ipFamilyPolicy field to Envoy NetworkPublishing definition.

Fixes #5308

### DIFF
--- a/internal/provisioner/equality/equality.go
+++ b/internal/provisioner/equality/equality.go
@@ -98,6 +98,11 @@ func ClusterIPServiceChanged(current, expected *corev1.Service) (*corev1.Service
 		changed = true
 	}
 
+	if !apiequality.Semantic.DeepEqual(current.Spec.IPFamilyPolicy, expected.Spec.IPFamilyPolicy) {
+		updated.Spec.IPFamilyPolicy = expected.Spec.IPFamilyPolicy
+		changed = true
+	}
+
 	if !apiequality.Semantic.DeepEqual(current.Spec.SessionAffinity, expected.Spec.SessionAffinity) {
 		updated.Spec.SessionAffinity = expected.Spec.SessionAffinity
 		changed = true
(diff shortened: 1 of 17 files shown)


changelogs/unreleased/small-ipFamilyPolicy.md
```diff
@@ -0,0 +1 @@
+Gateway API: Add `ipFamilyPolicy` field to `ContourDeployment.spec.envoy.networkPublishing` to control dual-stack-ness of the generated service.
```
skriss, line 1: rename to `5386-Jean-Daniel-small.md`.

internal/provisioner/equality/equality.go
```diff
@@ -157,6 +157,11 @@ func LoadBalancerServiceChanged(current, expected *corev1.Service) (*corev1.Serv
 		changed = true
 	}
 
+	if !apiequality.Semantic.DeepEqual(current.Spec.IPFamilyPolicy, expected.Spec.IPFamilyPolicy) {
+		updated.Spec.IPFamilyPolicy = expected.Spec.IPFamilyPolicy
+		changed = true
+	}
```
skriss, line 163: I believe we'll want this change for `ClusterIPServiceChanged` and `NodePortServiceChanged` as well.

skriss (approved):
LGTM, thanks @Jean-Daniel!

internal/provisioner/objects/service/service.go
```diff
@@ -250,6 +250,10 @@ func DesiredEnvoyService(contour *model.Contour) *corev1.Service {
 		}
 	}
 
+	if contour.Spec.NetworkPublishing.Envoy.IPFamilyPolicy != "" {
+		svc.Spec.IPFamilyPolicy = &contour.Spec.NetworkPublishing.Envoy.IPFamilyPolicy
```
sunjayBhatia, line 254: i *think* this could technically cause a memory leak since we're keeping a reference to a struct field so the containing struct may not be GCed, could instead use `ref.To` helper which would make a copy of the value of the field only and return a reference to the copy

there may be other places in the codebase we should fix this as well

sunjayBhatia (approved):

PR #5388, opened 2023-05-22
test/e2e: disable keep-alives in Poller

skriss:


sunjayBhatia (approved):

sunjayBhatia:
aha good catch! (I wonder if there should be a linter that prevents using http.DefaultTransport as-is 👀, maybe something to follow up with at some point)

skriss:
> aha good catch! (I wonder if there should be a linter that prevents using http.DefaultTransport as-is 👀, maybe something to follow up with at some point)

Yeah that's a good idea, for E2E's anyway it seems like using the default with keep alives enabled causes more trouble than it's worth.

PR #5353, opened 2023-05-10
Bump Gateway API to v0.7.1

sunjayBhatia:
To ensure compatibility with upcoming release

Fixes: #5336
Fixes: #5414

PR #5390, opened 2023-05-22
internal/envoy/v3: Loosen FilterMisdirectedRequests helper

sunjayBhatia:
Previously would configure the Lua script to match on requestedServerName when fqdn had a prefix of "*.". Now Gateway API allows a complete wildcard listener hostname match of "*" so this needs to be amended. Gateway API 0.7.x conformance tests surfaced this.


sunjayBhatia:
no release note as this is to support Gateway API conformance

should allow more tests to pass in https://github.com/projectcontour/contour/pull/5353

also upstream tests to be added as part of https://github.com/kubernetes-sigs/gateway-api/issues/2040

skriss (approved):

Issue #5375, opened 2023-05-18
Gateway provisioner: update control plane before data plane


skriss:
The [documented approach to upgrading Contour](https://projectcontour.io/resources/upgrading/) is to upgrade the control plane first, including allowing it to fully roll out, before upgrading the data plane. This helps minimize any downtime/failed requests during the upgrade. The Gateway provisioner does not currently implement this logic; it more or less updates both at the same time, which can lead to more failed requests during the upgrade. We should look at whether we can orchestrate these upgrades better in order to minimize downtime.

sunjayBhatia:
For testing, we could add an upgrade e2e test (w/ app polling etc.) with some relevant change to a Gateway that causes data plane and/or control plane re-rollout

sunjayBhatia:
as a follow up to https://github.com/projectcontour/contour/pull/5409, when we solve this issue we can see if we can re-increase the success threshold of the e2e upgrade test(s)

PR #5409, opened 2023-05-25
test/e2e: lower provisioner upgrade threshold

skriss:
Lowers the provisioner upgrade successful requests threshold to accommodate the fact that the
provisioner does not fully upgrade the control plane before the data plane.

I saw this flake at least once, so doing this for now until #5375 can be completed.


sunjayBhatia (approved):
lgtm, seems reasonable since it is a known issue that we are tracking 👍🏽

Issue #5411, opened 2023-05-26
Envoy FIPS build doesn't produce tarball


kyle-ntx:
**What question do you have?:**
I've been attempting to follow the instructions on the [FIPS 140-2 in Contour](https://projectcontour.io/guides/fips/) documentation page, however, I am struggling to create the envoy_binary.tar.gz file mentioned in the section about building Envoy.  

When I run the command `BAZEL_BUILD_EXTRA_OPTIONS="--define boringssl=fips" ENVOY_DOCKER_BUILD_DIR=~/envoyBuild ./ci/run_envoy_docker.sh './ci/do_ci.sh bazel.release.server_only'` from within the envoy directory (created by git cloning the envoy repo), the build appears to run without error, however, it does not produce the expected tar file. 

When I look in the specified ENVOY_DOCKER_BUILD_DIR directory, I find 3 folders: bazel_root, envoy, repository_cache.  Digging down into the 'envoy' folder I eventually find an executable called 'envoy' and another called envoy.dwp.  They are found at this path: `/envoyBuild/envoy/x64/source/exe/envoy`

Does anyone have any idea what I am doing wrong here?  My gut feeling on this is that either I've missed something obvious or the Envoy build process has changed and the command supplied in the documentation no longer works the way it used to. 

**Environment:**

- Contour version: latest. 
- Kubernetes version: (use `kubectl version`): N/A
- Kubernetes installer & version: N/A
- Cloud provider or hardware configuration: Azure  F32s_v2
- OS (e.g. from `/etc/os-release`):
Build Environment:
Ubuntu Server 22.04.2 LTS
Docker Engine 24.0.1

sunjayBhatia:
Yep the build process in Envoy has changed so our doc needs a bit of an update, we can use this issue to track making the updates

sunjayBhatia:
Take a look at the update in https://github.com/projectcontour/contour/pull/5415 to see if it works! Would appreciate any comments in case you run into any issues

kyle-ntx:
Thanks for the super fast response! I've taken a look at the PR and left a comment.

PR #5415, opened 2023-05-26
site: Update guide on building Envoy in FIPS compliant mode

sunjayBhatia:
Updates docs for main and 1.25

Fixes: #5411

### DIFF
--- a/site/content/docs/main/guides/fips.md
+++ b/site/content/docs/main/guides/fips.md
@@ -104,6 +104,8 @@ We can first compile the Envoy binary by running the following in a `bash` shell
 BAZEL_BUILD_EXTRA_OPTIONS="--define boringssl=fips" ENVOY_DOCKER_BUILD_DIR=<envoy-output-dir> ./ci/run_envoy_docker.sh './ci/do_ci.sh bazel.release //test/exe:envoy_static_test'
 ```
 
+*This command mimics the Envoy release CI process with the target `bazel.release` but differs in only running a single test for brevity. You may omit the `//test/exe:envoy_static_test` test entirely to run the full suite of Envoy tests.*
+
 Replace `<envoy-output-dir>` with a directory you would like the build output to be placed on your host computer.
 Once that build completes, you should have a file named `envoy_binary.tar.gz` in your specified output directory.
 
--- a/site/content/docs/main/guides/fips.md
+++ b/site/content/docs/main/guides/fips.md
@@ -101,7 +101,7 @@ Note that if you do build with FIPS mode outside of the build container, you can
 We can first compile the Envoy binary by running the following in a `bash` shell from the Envoy source directory:
 
 ```bash
-BAZEL_BUILD_EXTRA_OPTIONS="--define boringssl=fips" ENVOY_DOCKER_BUILD_DIR=<envoy-output-dir> ./ci/run_envoy_docker.sh './ci/do_ci.sh bazel.release.server_only'
+BAZEL_BUILD_EXTRA_OPTIONS="--define boringssl=fips" ENVOY_DOCKER_BUILD_DIR=<envoy-output-dir> ./ci/run_envoy_docker.sh './ci/do_ci.sh bazel.release //test/exe:envoy_static_test'
 ```
 
 Replace `<envoy-output-dir>` with a directory you would like the build output to be placed on your host computer.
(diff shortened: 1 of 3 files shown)


site/content/docs/main/guides/fips.md
```diff
@@ -101,7 +101,7 @@ Note that if you do build with FIPS mode outside of the build container, you can
 We can first compile the Envoy binary by running the following in a `bash` shell from the Envoy source directory:
 
 ```bash
-BAZEL_BUILD_EXTRA_OPTIONS="--define boringssl=fips" ENVOY_DOCKER_BUILD_DIR=<envoy-output-dir> ./ci/run_envoy_docker.sh './ci/do_ci.sh bazel.release.server_only'
+BAZEL_BUILD_EXTRA_OPTIONS="--define boringssl=fips" ENVOY_DOCKER_BUILD_DIR=<envoy-output-dir> ./ci/run_envoy_docker.sh './ci/do_ci.sh bazel.release //test/exe:envoy_static_test'
```
tsaarni, line 104: Just a question: what are these [test targets](https://github.com/envoyproxy/envoy/blob/main/test/exe/BUILD)? I've used `//source/exe:envoy-static` instead (though I never dealt with FIPS builds).

sunjayBhatia: that particular test compiles the envoy binary target and then checks it is statically linked, the `bazel.release` "subcommand" of that script takes a test target to run so i chose that one as a minimal test that wouldn't add too much overhead to the build (bazel should cache the test compilation of the envoy binary and reuse it in the final binary build step since it is using all the same parameters)

sunjayBhatia: I decided to omit commenting on these details from this change but can add it if we think its useful

tsaarni: Thanks for the info! Maybe it could be good to add some short version, since for someone who does not know (like me!) it might look a bit like a debug build due to the target name 😄

tsaarni (commented):
Great explanation, it makes it clear now!

Curiously, if I run the command, I do not get `envoy_binary.tar.gz` either, but I get this

```
$ tar tf output/envoy/x64/bin/release.tar.zst
dbg/
dbg/envoy
dbg/envoy-contrib
dbg/envoy-contrib.dwp
dbg/envoy.dwp
envoy
envoy-contrib
utils/
utils/su-exec
```

When uncompressing the package, the binaries did not have execute flags, but otherwise they seemed ok...

tsaarni (approved):
> $ tar tf output/envoy/x64/bin/release.tar.zst

I realized this was quite recent change https://www.github.com/envoyproxy/envoy/pull/27271. The .tar.gz file likely still exists in the most recent released versions, so approving the PR.  

On the long run it might be bit quite difficult to keep up with Envoy, at least when reusing the CI scripts as the basis for the build.

site/content/docs/main/guides/fips.md
```diff
@@ -101,9 +101,11 @@ Note that if you do build with FIPS mode outside of the build container, you can
...
 ```
 
+*This command mimics the Envoy release CI process with the target `bazel.release` but differs in only running a single test for brevity. You may omit the `//test/exe:envoy_static_test` test entirely to run the full suite of Envoy tests.*
+
 Replace `<envoy-output-dir>` with a directory you would like the build output to be placed on your host computer.
 Once that build completes, you should have a file named `envoy_binary.tar.gz` in your specified output directory.
```
kyle-ntx, line 110: `envoy_binary.tar.gz` is not created. 
`<envoy-output-dir>/envoy/x64/bin/release.tar.zst` is created instead and needs to be copied to the <envoy-repo-dir>/linux/amd64/  directory. After this we can run the `docker build -f ./ci/Dockerfile-envoy .`  command as specified in the existing documentation.

kyle-ntx: Just to clarify, we don't need to extract `release.tar.zst` the `[Dockerfile is expecting that file](https://github.com/envoyproxy/envoy/blob/af8aef48b5395d32d4bbac5aa3ed0a84d3bf31d1/ci/Dockerfile-envoy#L12)`.

sunjayBhatia: ah thats right, forgot to update the copy instructions, will add that as well 👍🏽

sunjayBhatia: updated the copy instructions, please take another look!

kyle-ntx: Thanks @sunjayBhatia! 
I got it working, however one thing that tripped me up is that I was running an older version of docker and the docker build steps failed.  Took me ages to figure out that the docker version had anything to do with the problem. Do you think it's worth adding a note to suggest the build is carried out with the latest version of docker?

kyle-ntx: oh man, I just noticed there is a new note there. All good from me then

kyle-ntx (approved):

skriss (approved):
I have not done any testing here but given other feedback, seems like it's ready for merge

Issue #5401, opened 2023-05-25
Allow for advanced configuration on listeners and proxies. 


clayton-gonsalves:
## Summary 
Envoy supports many knobs to tune the behaviour of listeners and clusters. 
Contour today already supports some of them. Example `idle_timeout` [here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-msg-config-core-v3-httpprotocoloptions) 

Supporting more of the knobs for both the clusters and listeners would be nice. 

###Suggestion:
 
#### Listener  
Under the [EnvoyConfig](https://projectcontour.io/docs/main/config/api/#projectcontour.io/v1alpha1.EnvoyConfig) 
`max_requests_per_connection`

##### Per Cluster 
Under the [EnvoyConfig.Cluster](https://projectcontour.io/docs/main/config/api/#projectcontour.io/v1alpha1.ClusterParameters)
`max_requests_per_connection`

##### All clusters
Similar to how we have a Policy defined for an HTTPProxy, we can have a `ConfigurationsPolicy` under the HTTPProxy virtual host.

PR #5417, opened 2023-05-26
Add max_requests_per_connection for cluster and listener config

clayton-gonsalves:
This Change Adds toggling the `maxRequestsPerConnection` flag on the listener and/or the Clusters

Fixes: https://github.com/projectcontour/contour/issues/5401


internal/xdscache/v3/listener_test.go
```diff
@@ -3401,6 +3402,142 @@ func TestListenerVisit(t *testing.T) {
 				SocketOptions: envoy_v3.TCPKeepaliveSocketOptions(),
 			}),
 		},
+		"httpproxy with MaxRequestsPerConnection set in listener config": {
```
davinci26, line 3405: do we have a test where we assert that when we don't set it in the config we don't set it at all?

davinci26 (commented):
In what scenario do we need the listener to limit connections?

internal/xdscache/v3/listener_test.go
clayton-gonsalves: There are other tests (such as test with defaults and few others) which would fail as the setting I introduced is off by default.

clayton-gonsalves:
> In what scenario do we need the listener to limit connections?
it is more common to do this with the upstream. but one that I can think of are limiting long-lived connections.

I added the setting to the listener/downstream also because envoy exposes this in both places, so made sense to introduce it in contour at both places.

davinci26 (approved):
Minor preference for not adding stuff to the API unless we know that we need them because we would need to maintain them 'forever' but won't block on that

internal/dag/dag.go
```diff
@@ -985,6 +985,9 @@ type Cluster struct {
 	TimeoutPolicy ClusterTimeoutPolicy
 
 	SlowStartConfig *SlowStartConfig
+
+	// MaxRequestsPerConnection defines the maximum number of requests per connection to the upstream before it is closed.
+	MaxRequestsPerConnection *uint32
```
skriss, line 990: I don't think this is ever being set for Clusters? It'll need to be plumbed through to the HTTPProxy (at least) DAG processor so it can be set on the dag.Cluster, similar to DNSLookupFamily

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -589,6 +596,13 @@ type ClusterParameters struct {
 	// Other values will produce an error.
 	// +optional
 	DNSLookupFamily ClusterDNSFamilyType `json:"dnsLookupFamily,omitempty"`
+
+	// Defines the maximum requests for upstream connections. If not specified, there is no limit.
+	// see https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-msg-config-core-v3-httpprotocoloptions
+	// for more information.
+	//
+	// +optional
+	MaxRequestsPerConnection *uint32 `json:"maxRequestsPerConnection,omitempty"`
```
skriss, line 605: I'm curious if you considered making this configurable per-cluster, rather than having a global setting that applies to all clusters? I don't have a problem with it as-is, but could imagine some users wanting to be able to change on a per-cluster basis as well.

internal/envoy/v3/cluster.go
```diff
@@ -339,6 +340,10 @@ func protocolOptions(explicitHTTPVersion HTTPVersionType, idleConnectionTimeout
 		options.CommonHttpProtocolOptions = &envoy_core_v3.HttpProtocolOptions{IdleTimeout: durationpb.New(idleConnectionTimeout.Duration())}
 	}
 
+	if maxRequestsPerConnection != nil {
+		options.CommonHttpProtocolOptions = &envoy_core_v3.HttpProtocolOptions{MaxRequestsPerConnection: wrapperspb.UInt32(*maxRequestsPerConnection)}
```
sunjayBhatia, line 344: looks like this will overwrite the above setting for idle timeout if both are set

sunjayBhatia: can add a test where both are set to make sure we cover this

clayton-gonsalves: added

internal/dag/dag.go
clayton-gonsalves: oof, fixed.

apis/projectcontour/v1alpha1/contourconfig.go
clayton-gonsalves: Hey Skriss, 
yeah, i did think about it. 

I was thinking of something like a `AdvancedConfigPolicy` on the HTTPProxy. 

However, I wanted to run that by the design review process. Also, I want to make sure we design it with extensibility in mind as I envision all the cluster-specific config to live in that, e.g. access logs per cluster, config knobs etc.

clayton-gonsalves: Also, from the discussion with Sotiris above, the `max_requests_per_connection` is a rather uncommon use case on the listener. I added it for the sake of uniformity, but if we think its not needed at this time, I am happy to revert those changes.

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -366,6 +366,13 @@ type EnvoyListenerConfig struct {
 	// +optional
 	ConnectionBalancer string `json:"connectionBalancer,omitempty"`
 
+	// Defines the maximum requests for downstream connections. If not specified, there is no limit.
+	// see https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-msg-config-core-v3-httpprotocoloptions
+	// for more information.
+	//
+	// +optional
+	MaxRequestsPerConnection *uint32 `json:"maxRequestsPerConnection,omitempty"`
```
sunjayBhatia, line 374: we might want to make these `int32` instead and add a minimum of 0

see https://github.com/projectcontour/contour/issues/2123

clayton-gonsalves: not sure setting as min of 0 would work, by default, this is unset on envoy. adding 0 might result in undefined behaviour.

sunjayBhatia: ah fair point, this field is not like a timeout where setting to 0 means to disable it, so a minimum of 1 would make sense then to make sure we're not getting an invalid value (negative or 0), the fact the field is a pointer is our signal if it has been set or not

hm, maybe ok to leave as-is actually we do have uints scattered around

sunjayBhatia: i guess generally we should have a kubebuilder validation minimum of 1 for this field since 0 doesn't make any sense

pkg/config/parameters.go
```diff
@@ -413,6 +413,13 @@ type ClusterParameters struct {
 	// See https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto.html#envoy-v3-api-enum-config-cluster-v3-cluster-dnslookupfamily
 	// for more information.
 	DNSLookupFamily ClusterDNSFamilyType `yaml:"dns-lookup-family"`
+
+	// Defines the maximum requests for upstream connections. If not specified, there is no limit.
+	// see https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-msg-config-core-v3-httpprotocoloptions
+	// for more information.
+	//
+	// +optional
+	MaxRequestsPerConnection *uint32 `yaml:"max-requests-per-connection,omitempty"`
```
sunjayBhatia, line 422: sorry for the runaround but a similar minimum value of 1 validation in this package would be good to add

sunjayBhatia: ah i wasn't clear here, but we'll need to add validation in the go code itself, this is not a CRD so kubebuilder validations don't apply, we'll need to add something here https://github.com/projectcontour/contour/blob/4333ed3bbab811a4517854f2d892028f2e268df8/pkg/config/parameters.go#L825 and a test to verify it

clayton-gonsalves: oops, sorry for the back and forth. hopefully this last commit should take care of it.

pkg/config/parameters.go
```diff
@@ -436,6 +455,14 @@ type ListenerParameters struct {
 	// See https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/listener.proto#envoy-api-msg-listener-connectionbalanceconfig
 	// for more information.
 	ConnectionBalancer string `yaml:"connection-balancer"`
+
+	// Defines the maximum requests for downstream connections. If not specified, there is no limit.
+	// see https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-msg-config-core-v3-httpprotocoloptions
+	// for more information.
+	//
+	// +kubebuilder:validation:Minimum=1
```
sunjayBhatia, line 463: nit: lets remove this and the above instance of this

internal/dag/httpproxy_processor.go
```diff
@@ -948,20 +951,21 @@ func (p *HTTPProxyProcessor) computeRoutes(
...
+				RequestHeadersPolicy:     reqHP,
+				ResponseHeadersPolicy:    respHP,
+				CookieRewritePolicies:    cookieRP,
+				Protocol:                 protocol,
+				SNI:                      determineSNI(r.RequestHeadersPolicy, reqHP, s),
+				DNSLookupFamily:          string(p.DNSLookupFamily),
+				ClientCertificate:        clientCertSecret,
+				TimeoutPolicy:            ctp,
+				SlowStartConfig:          slowStart,
+				MaxRequestsPerConnection: p.MaxRequestsPerConnection,
```
sunjayBhatia, line 968: this isn't exclusive to this change, but we have various cluster-level settings that are supposed to be global that do not get applied to Ingress or Gateway API generated clusters

probably worth a follow up issue so we don't block this but we should make this consistent

cc @skriss

sunjayBhatia (commented):
just one nit to clean up but otherwise looks good to me, once that change is made i think im good to approve

internal/dag/httpproxy_processor.go
skriss: xref #5458

skriss (approved):
LGTM, thanks @clayton-gonsalves!

PR #5422, opened 2023-05-30
refactor to fix shadow-vars & dup import

izturn:

refactor to fix shadow variable: `any`  and import the packages more then once.


sunjayBhatia (approved):
nice cleanups 👍🏽

sunjayBhatia:
hm wonder why this rule was not triggered for these: https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#duplicated-imports

maybe something to look at as a follow up, to make sure enabled linters are what we want and all the settings for rules are correct

skriss (approved):

PR #5424, opened 2023-05-30
e2e: add linter for http.DefaultTransport

### LINKED ISSUE #5389: add linter to prevent using http.DefaultTransport for E2E's
aha good catch! (I wonder if there should be a linter that prevents using http.DefaultTransport as-is 👀, maybe something to follow up with at some point)

_Originally posted by @sunjayBhatia in https://github.com/projectcontour/contour/issues/5388#issuecomment-1557323194_

izturn:

fix #5389

### DIFF
--- a/test/e2e/http.go
+++ b/test/e2e/http.go
@@ -19,7 +19,6 @@ import (
 	"context"
 	"crypto/tls"
 	"io"
-	"net"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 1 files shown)


sunjayBhatia (approved):
tested it out and this added linter works as it says on the tin 👍🏽

test/e2e/http.go
```diff
@@ -333,3 +333,20 @@ func HasStatusCode(code int) func(*HTTPResponse) bool {
...
+			Timeout:   30 * time.Second,
+			KeepAlive: 30 * time.Second,
+		}).DialContext,
+		ForceAttemptHTTP2:     true,
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
+		DisableKeepAlives:     true,
+	}
```
skriss, line 351: Let's do this so we don't have to worry about keeping in sync (can remove the comment too):
```suggestion
        // nolint:forbidigo
        transport := http.DefaultTransport.(*http.Transport).Clone()
	transport.DisableKeepAlives = true
	return transport
```

PR #5427, opened 2023-05-31
cmd/contour: Failure setting GOMAXPROCS is non-fatal

sunjayBhatia:
This is an issue in certain local development and other environments where cgroup namespaces are distinct per container. The automaxprocs library errors with 'path <foo> is not a descendent of mount point root <foo>/kubelet and cannot be exposed from "/sys/fs/cgroup/misc/kubelet"'


sunjayBhatia:
Candidate to backport to 1.25

PR #5432, opened 2023-06-01
collapse 1.21 and 1.20 docs

### LINKED ISSUE #5197: docs: collapse more sets of docs into one per major+minor version
Contour recently moved from a set of documentation per patch release, to one set per unique major+minor version (see https://github.com/projectcontour/contour/pull/5163). As part of the above PR, the docs for all of the patch releases for each of the three supported minor versions at the time (1.24, 1.23, 1.22) were combined into a single set per major+minor version. However, our docs site shows documentation for versions going back further than this (see the dropdown at the top left of https://projectcontour.io/docs/1.24/), and so still shows patch release versions in the dropdown.

It would be nice to combine the docs for a few more minor versions into a single set per minor version, and to remove a few of the older minor versions from the docs dropdown, such that the dropdown _only_ shows major+minor versions.

The PR linked above can be used as a reference, but the steps to be taken for each major+minor version are:
- rename `site/content/docs/<latest patch release for major+minor>` to `site/content/docs/<major.minor>`
- delete the older patch release docs for the given major+minor
- Update `site/config.yaml` -- params.docs_versions - to remove the patch version entries, and include the major+minor version
- rename `site/data/docs/<toc mapping file for patch release>` to `site/data/docs/<major-minor>-toc.yml`
- update `site/data/docs/toc-mapping.yml` to remove the entries for the patch releases, and include the major+minor version
- update `site/content/docs/<major.minor>/_index.md` to set version = "<major.minor>", and add "branch: release-<major.minor>"
- Within `site/content/docs/<major.minor>`, find all uses of `< param version >` **that are linking to the GitHub repository**, and replace it with `< param branch >`

Run `make site-devel` to build & run the website locally (requires Hugo), and ensure everything looks good.

skriss:
- Collapses the 1.21 and 1.20 docs into a single set for each minor version
- Removes docs for versions earlier than 1.20 from the dropdown.

Closes #5197.


sunjayBhatia (approved):
as far as i can tell this looks right 👍🏽 only minor concern would have been if more recent patch versions had doc updates the older ones did not, but I don't think that exists for these versions

skriss:
> only minor concern would have been if more recent patch versions had doc updates the older ones did not, but I don't think that exists for these versions

Yeah -- I kept the latest patch version's docs for each, and I think if there _were_ any diffs, it'd just be fixes that should apply to any patch version.

PR #5434, opened 2023-06-02
internal/sorter: Routes with method matching get precedence

sunjayBhatia:
This is to support Gateway API conformance, release 0.7.1 dictates that HTTP method comes before length of header matches in the precendence order for choosing routes.

Since we currently implement method matching in the DAG by adding a :method header match, this solution finds any :method headers being matched on and first determines precedence of two routes being compared based on that. If method matches are equal (both present or not present), then we go ahead with comparing the remaining header matches.

Since HTTPProxy provides this ability as well, this will affect priority of routes from HTTPProxies, not just Gateway API HTTPRoute.

Alternately we could add a method field directly to the dag.Route struct and implement method matching that way. We may want to in that case also sanitize the header matches to make sure we properly collect matches on the :method header, or disallow using them in HTTPRoute, etc.

See gateway API docs: https://github.com/kubernetes-sigs/gateway-api/blob/v0.7.1/apis/v1beta1/httproute_types.go#L163-L167

### DIFF
--- a/changelogs/unreleased/5434-sunjayBhatia-minor.md
+++ b/changelogs/unreleased/5434-sunjayBhatia-minor.md
@@ -1,0 +1,6 @@
+## Routes with HTTP Method matching have higher precedence
+
+For conformance with Gateway API v0.7.1+, routes that utilize HTTP method matching now have an explicit precedence over routes with header/query matches.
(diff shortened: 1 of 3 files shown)


skriss (commented):
LGTM, I think this makes sense. Let's just call out the potential change in route priority for this corner case in the release note.

PR #5437, opened 2023-06-05
internal/envoy/v3: RouteConfiguration ignores port when choosing virtualhost

sunjayBhatia:
Previously we stripped the port from the Host header when matching a virtualhost, which also resulted in the header being modified for the downstream. For Gateway API conformance, we no longer want to do this.

This basically reverts #3458 and uses the newer RouteConfiguration field that allows us to ignore port when choosing a virtualhost. We reintroduce the logic in the SNI/Host misdirected request Lua to strip the port when checking the hostname against SNI.


sunjayBhatia:
huh the ingress test fails but almost identical gateway one passes

sunjayBhatia:
ah, the header regex match on a single DNS label seems to be doing it

skriss (approved):
LGTM assuming E2E's are happy now, `:authority` header match change make sense

PR #5457, opened 2023-06-08
Wait for gateway-api admission server to roll out in install script

sunjayBhatia:
Motivation for this change is flaky upgrade tests

When running upgrade tests, we install and setup kind which installs the current version of gateway api admission server and then in the tests, install the latest release of contour + the supported gateway api components, which might mean a redeploy of the admission webhook. Since we don't wait for the webhook to roll out, creating gateway resources causes the tests to fail if the webhook isnt up yet or some pods are terminating (also why waiting for the deployment to be available is a little problematic).


skriss (approved):
Ah, this makes sense, LGTM! 💯

sunjayBhatia:
huh looks like this didnt fully fix the issue: https://github.com/projectcontour/contour/actions/runs/5216288625/jobs/9419468641

PR #5463, opened 2023-06-12
Fix dag rebuild trigger when Secret object changes

### LINKED ISSUE #5462: HTTPProxy not reconciled on client-clr secret changes
Noticed that if client-crl secret is configured in HTTPProxy, that proxy is not reconciled/updated on any changes related to the client-crl secret.

A minimal example:
Create a HTTPProxy with external tls and client-clr configured.
Apply external tls secret
Apply proxy
Check proxy description: Says client-crl secret missing. (So far good)
Apply client-clr secret
Check proxy description: Says client-crl secret missing. (Not true anymore)

If the proxy is reapplied the problem is solved.
Same happens on other secret related activities: apply, delete, patch

Any change related to the client-crl secret to have an affect on the proxy and traffic.

**Environment:**

- Contour version: 1.25
- Kubernetes installer & version: 1.23.6
- OS (e.g. from `/etc/os-release`): Suse

### LINKED ISSUE #5538: secretTriggersRebuild does not handle the projectcontour.io/tls-cert-namespace Ingress annotation
As of Ingress v1, the `/` character is no longer allowed in the Ingress TLS secret name field, so the `projectcontour.io/tls-cert-namespace` annotation is supported to specify the namespace where the TLS secret is located, if different than the namespace than the Ingress. The `secretTriggersRebuild` function has not been updated to account for this, though, which means that changes to a Secret won't trigger a DAG rebuild if only used by an Ingress in a different namespace.

tsaarni:
Update of CRL Secret did not trigger DAG rebuild when it was not co-located in the same Secret with CA certificate. The effect was that CRL secret was not recognized if it was created after the HTTPProxy and updates to CRL content were not taken into use.

Trigger for DAG rebuild was also missing for Ingress when using certificate delegation via `projectcontour.io/tls-cert-namespace` annotation (Ingress v1 does not anymore allow `/` as part of secret name). It is also fixed in this PR.

Fixes #5462
Fixes #5538

### DIFF
--- a/internal/dag/cache.go
+++ b/internal/dag/cache.go
@@ -520,7 +520,7 @@ func (kc *KubernetesCache) secretTriggersRebuild(secretObj *v1.Secret) bool {
 
 	for _, ingress := range kc.ingresses {
 		for _, tls := range ingress.Spec.TLS {
-			if secret == k8s.NamespacedNameFrom(tls.SecretName, k8s.TLSCertAnnotationNamespace(ingress),  k8s.DefaultNamespace(ingress.Namespace)) {
+			if secret == k8s.NamespacedNameFrom(tls.SecretName, k8s.TLSCertAnnotationNamespace(ingress), k8s.DefaultNamespace(ingress.Namespace)) {
 				return true
 			}
 		}
--- a/internal/dag/cache.go
+++ b/internal/dag/cache.go
@@ -501,8 +501,8 @@ func isRefToService(ref gatewayapi_v1beta1.BackendObjectReference, service *v1.S
 // secretTriggersRebuild returns true if this secret is referenced by an Ingress
 // or HTTPProxy object, or by the configuration file. If the secret is not in the same namespace
 // it must be mentioned by a TLSCertificateDelegation.
-func (kc *KubernetesCache) secretTriggersRebuild(secret *v1.Secret) bool {
-	if _, isCA := secret.Data[CACertificateKey]; isCA {
+func (kc *KubernetesCache) secretTriggersRebuild(secretObj *v1.Secret) bool {
+	if _, isCA := secretObj.Data[CACertificateKey]; isCA {
 		// locating a secret validation usage involves traversing each
 		// proxy object, determining if there is a valid delegation,
 		// and if the reference the secret as a certificate. The DAG already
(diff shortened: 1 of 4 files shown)


internal/dag/cache.go
```diff
@@ -532,6 +532,10 @@ func (kc *KubernetesCache) secretTriggersRebuild(secret *v1.Secret) bool {
 				return true
 			}
 		}
+		cv := tls.ClientValidation
+		if cv != nil && proxy.Namespace == secret.Namespace && cv.CertificateRevocationList == secret.Name {
```
sunjayBhatia, line 536: also add the delegation check like here? https://github.com/projectcontour/contour/blob/d16a115d1c49583309707f8aebb3b48b9f54e71a/internal/dag/cache.go#L525-L534

tsaarni: I've now fixed this.

I suggest some refactoring for the comparison logic for delegations since in my opinion the old one was bit difficult to understand. 

How I _think_ the old logic worked: 

It built map `delegations` where it "projects" all delegated secrets like those secrets would have been in the target namespace ([link](https://github.com/projectcontour/contour/blob/d0f1171bd31f8aebf1fb80a6991138f396470d37/internal/dag/cache.go#L517-L523). For example, delegation of secret `foo` stored in `ns1` and delegated for namespace `targetns` would yield map entry for secret `targetns/foo`. I think of this as "projecting" the secret `foo` into target namespace.

Consider example where there were two secrets both called `foo`, one in `ns1` and another in `ns2`. Only `foo` in `ns1` was delegated. Now imagine that user refers to `ns2/foo` (which was not delegated).  The `delegations` list has the "projected" secret `targetns/foo` so the code accepts this and proceeds to compare secret reference `ns2/foo` - even if `ns2` was not delegated ([link](https://github.com/projectcontour/contour/blob/d0f1171bd31f8aebf1fb80a6991138f396470d37/internal/dag/cache.go#L565-L569)).  This is not that important; it would just trigger extra DAG rebuild when it does not need to.

In the refactored version, I keep the delegation checks, which complicates the logic but avoids the extra DAG rebuilds.

internal/dag/cache.go
```diff
@@ -511,38 +511,56 @@ func (kc *KubernetesCache) secretTriggersRebuild(secret *v1.Secret) bool {
...
-				if tls.SecretName == secret.Namespace+"/"+secret.Name {
-					return true
-				}
+		return false
+	}
+
+	for _, ingress := range kc.ingresses {
+		for _, tls := range ingress.Spec.TLS {
+			if compare(secret, k8s.NamespacedNameFrom(tls.SecretName, k8s.DefaultNamespace(ingress.Namespace)), ingress.Namespace) {
+				return true
```
skriss, line 563: Two thoughts here:
- When we're checking each Ingress, HTTPProxy, etc. to see if it references `secretObj`, the _first_ thing we should do is see if the namespace/name of the reference match the namespace/name of `secretObj` (taking into account defaulting of the reference's namespace to the Ingress/HTTPProxy's namespace if it's not explicitly specified). If not, then the Ingress/HTTPProxy does not reference the Secret in question, so we don't have to do any further checks for delegation/etc. That'll be the most common case (most Ingresses/Proxies won't have a reference to the Secret in question), so we'll want to exit early if it's hit. 
- For checking if a reference is allowed, I think we should be able to use the existing `delegationPermitted` function here; it'd be nice to only have one (well-tested) version of that logic. Since we'll only hit this branch for Ingresses/HTTPProxies that _do_ reference the Secret, I'm not worried about small differences in performance.

So, for Ingress for example, I'm thinking something like:

```go
for _, ingress := range kc.ingresses {
	for _, tls := range ingress.Spec.TLS {
		// Get a NamespacedName for the Secret that the Ingress references, to make comparisons easier.
		ingressSecretRef := k8s.NamespacedNameFrom(tls.SecretName, k8s.DefaultNamespace(ingress.Namespace))

		// Secret is not referenced by this Ingress TLS: continue.
		if ingressSecretRef != secret {
			continue
		}

		// Check if Secret is allowed to be referenced by this Ingress
		// (this handles both same-namespace and cross-namespace cases).
		if kc.delegationPermitted(secret, ingress.Namespace) {
			return true
		}
	}
}```

Also note https://github.com/projectcontour/contour/issues/5538 which I noticed while looking at this; we can fix it separately.

Comment (tsaarni) [reply internal/dag/cache.go:563]: I agree it would be best to reuse `delegationPermitted()` but I was concerned it was very sub-optimal performance-wise for this particular use case: it would repeatedly try to find the same secret from all delegations, as many times as we have ingresses/HTTPProxies. Comparing the names first makes the call less frequent (assuming delegation is used rarely, which is likely the case) so maybe the performance does not matter?  

I used reverse order since as generally permission check would be nice to do first, but it does not matter in this case. However, preferably, I would like to combine the name comparison with permission check in a single function call, so that it will become less likely that any future edits will miss the permission check.

As last point, as I mentioned, I was also thinking of simplifying this all and completely removing the delegation check. Little harm would be done by extra DAG rebuild, even if it would happen for rare references that were invalid because of missing delegation?

Comment (skriss) [reply internal/dag/cache.go:563]: > As last point, as I mentioned, I was also thinking of simplifying this all and completely removing the delegation check. Little harm would be done by extra DAG rebuild, even if it would happen for rare references that were invalid because of missing delegation?

That's fair, I think that would be perfectly fine too and would avoid all of this!

Comment (tsaarni) [reply internal/dag/cache.go]: I went ahead changed the PR and removed the delegation checks. I changed the test cases a bit, removing some delegation checks from `TestSecretTriggersRebuild` which do not anymore have much meaning, but I kept the `TLSCertificateDelegation` objects in that test case for completeness, even though they are now ignored.

**Comment by tsaarni** [comment]
I've updated this PR to address also #5538.

File: internal/dag/cache.go [hunk 3]
```diff
@@ -511,38 +513,15 @@ func (kc *KubernetesCache) secretTriggersRebuild(secret *v1.Secret) bool {
...
-			}
-		}
-
-		if delegations["*/"+secret.Name] {
-			for _, tls := range ingress.Spec.TLS {
-				if tls.SecretName == secret.Namespace+"/"+secret.Name {
-					return true
-				}
+		for _, tls := range ingress.Spec.TLS {
+			if secret == k8s.NamespacedNameFrom(tls.SecretName, k8s.TLSCertAnnotationNamespace(ingress)) {
```
skriss, line 523: Still need to fall back to the ingress's namespace as a last resort:
```suggestion
			if secret == k8s.NamespacedNameFrom(tls.SecretName, k8s.TLSCertAnnotationNamespace(ingress),  k8s.DefaultNamespace(ingress.Namespace)) {
```

skriss (changes requested):

internal/dag/cache.go
tsaarni: This was first time I dared to try click "commit suggestion" but I failed :-D

skriss: Sorry, my suggestion was invalid I guess! Looks like extra space

internal/dag/cache_test.go
```diff
@@ -122,10 +122,13 @@ func TestKubernetesCacheInsert(t *testing.T) {
 					ObjectMeta: metav1.ObjectMeta{
 						Name:      "www",
 						Namespace: "extra",
+						Annotations: map[string]string{
+							"projectcontour.io/tls-delegation-namespace": "default",
```
skriss, line 126:
```suggestion
							"projectcontour.io/tls-cert-namespace": "default",
```

internal/dag/cache_test.go
```diff
@@ -160,10 +163,13 @@ func TestKubernetesCacheInsert(t *testing.T) {
...
```
skriss, line 167:
```suggestion
							"projectcontour.io/tls-cert-namespace": "default",
```

skriss (approved):
LGTM pending CI, we can ignore the DCO check since your initial commit has it and just ensure it's on the squashed commit

Thanks @tsaarni!

Issue #5264, opened 2023-04-10
Allow Cluster per connection buffer limit to be configurable


skriss:
Per https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/edge, when running Envoy as an edge proxy, it's recommended to set [Listener per-connection buffer limits](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener.proto#envoy-v3-api-field-config-listener-v3-listener-per-connection-buffer-limit-bytes) to 32KiB.

There are two changes needed here:
~~- [ ] set the default value within [the clusterDefaults func](https://github.com/projectcontour/contour/blob/main/internal/envoy/v3/cluster.go#L35)~~
- [ ] add a tuneable option within [ClusterParameters](https://github.com/projectcontour/contour/blob/main/pkg/config/parameters.go#L400) (for the Contour config file) and [ClusterParameters](https://github.com/projectcontour/contour/blob/main/apis/projectcontour/v1alpha1/contourconfig.go#L568) (for the ContourConfiguration CRD).

These can be done separately (in separate PRs), as the first item is smaller than the second.

This also needs some level of testing to ensure it does not adversely affect existing installations.

(xref #1408, creating a new issue for clarity for ContribFest).

Issue #1795, opened 2019-10-25
Investigate setting default HTTP/2 options for upstream clusters


youngnick:
In #1375, a PR to Envoy suggested setting HTTP/2 options as follows:

```yaml
    http2_protocol_options:
      max_concurrent_streams: 100
      initial_stream_window_size: 65536 # 64 KiB
      initial_connection_window_size: 1048576 # 1 MiB
```
This ticket is to investigate setting these.

youngnick:
xref #1375

rajatvig:
Would it be ok for me to work on making these as the default? Or is it desired to keep the values configurable?

sunjayBhatia:
I think the first action item here is to just document for posterity and so we have a sort of decision record what the impact of changing these settings is (what the difference from not setting them is, any memory etc. impacts), and then we can proceed with a change (to change the defaults or make configurable etc.)

PR #5464, opened 2023-06-12
Set per_connection_buffer_limit_bytes to 32768 as a default

rajatvig:
Fixes part 1 of https://github.com/projectcontour/contour/issues/5264


rajatvig:
> Looks good to me! 👍
> 
> One question: This might be unnecessary worry, but since there is quite a difference from the original defaults, is there a chance that some users would need the configurability from day one to avoid abrupt change? BTW the same question applies to #5330, and possibly it was the concern that @skriss had in mind here [#5330 (comment)](https://github.com/projectcontour/contour/pull/5330#issuecomment-1548526162)

I can follow it up with an option to configure it using the contour configuration. Was planning to work on https://github.com/projectcontour/contour/issues/1795 as well which I assume would also need defaults to match what exists and overrides where desired.

tsaarni:
> I can follow it up with an option to configure it using the contour configuration.

Sounds great, thank you!

> Was planning to work on #1795 as well which I assume would also need defaults to match what exists and overrides where desired.

Nice! I'll add comment to that issue about some potential Envoy & HTTP/2 TLS issues to look for, that I did not see mentioned/linked so far.

tsaarni (approved):

rajatvig:
@tsaarni I worked on making it configurable and realised it might be a cleaner implementation and avoids surprises vs setting a default and then adding in configuration.

https://github.com/projectcontour/contour/pull/5493 is the PR.

rajatvig:
Closing in favour of https://github.com/projectcontour/contour/pull/5493

PR #5472, opened 2023-06-13
internal/dag: GetServiceClusters no longer returns ExternalName clusters

sunjayBhatia:
This method is used by the EndpointsTranslator on dag change to recalculate the relevant endpoints to send to Envoy. Since we do not use EDS to configure endpoints for Clusters that come from ExternalName Services, we no longer collect ServiceClusters associated with these Clusters. This is required for ADS, as go-control-plane will not respond to requests if the snapshot contains resources that were not requested.

### DIFF
--- a/internal/dag/accessors.go
+++ b/internal/dag/accessors.go
@@ -241,6 +241,14 @@ func (d *DAG) GetServiceClusters() []*ServiceCluster {
 	var res []*ServiceCluster
 
 	for _, cluster := range d.GetClusters() {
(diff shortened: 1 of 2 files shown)


sunjayBhatia:
This was causing this test to fail when enabling ADS with go-control-plane: https://github.com/projectcontour/contour/blob/76df45e774d0f015b3f85e388931e4a7c7ec9319/test/e2e/httpproxy/internal_redirect_test.go#L93

PR #5474, opened 2023-06-14
internal/k8s: Fix log test pollution

sunjayBhatia:
Fixes flaky test/test pollution from test that output to stdout/err. Uses test logger instead.


sunjayBhatia:
From test failures such as https://github.com/projectcontour/contour/actions/runs/5261137167/jobs/9508931469

tsaarni (approved):

PR #5477, opened 2023-06-14
fix new codespell issues

skriss:


sunjayBhatia (approved):
huh i guess there might have been changes to the codespell dictionary/allowed typos

skriss:
> huh i guess there might have been changes to the codespell dictionary/allowed typos

Yeah I think it happens with minor/patch bumps sometimes, 🤷

PR #5484, opened 2023-06-16
Remove dots in certgen job name

### LINKED ISSUE #5483: Certgen job name with dots causes warning when deployed
Since Kubernetes v1.27 following warning is printed when deploying contour with the example manifests

```console
$ kubectl apply -f https://projectcontour.io/quickstart/contour.yaml
...
Warning: metadata.name: this is used in Pod names and hostnames, which can result in surprising behavior; a DNS label is recommended: [must not contain dots]
job.batch/contour-certgen-v1.25.0 created
```

The job name fails [this check](
https://github.com/kubernetes/kubernetes/blob/fa78f28f0a011e9922cf12145dd602c5a63981e8/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L178-L203) in Kubernetes, introduced by [this change](https://www.github.com/kubernetes/kubernetes/pull/114412). The version number suffix like `-v1.25.0` is added to the job name by the release script [here](https://github.com/projectcontour/contour/blob/45c36efd4fe37917690ea1bb2e35cecfe9b9dd7f/hack/release/make-release-tag.sh#L49-L55).

Note that this is just a warning and the deployment is still successful.

tsaarni:
Avoids Kubernetes warning when deploying contour.

Fixes #5483

### DIFF
--- a/hack/release/make-release-tag.sh
+++ b/hack/release/make-release-tag.sh
@@ -47,10 +47,12 @@ for example in examples/contour/03-envoy.yaml examples/deployment/03-envoy-deplo
 done
 
 # Update the certgen job name to the new version.
+# Dot in the version is replaced with a dash to make it a valid pod name.
 for example in examples/contour/02-job-certgen.yaml ; do
     run::sed \
-        "-es|contour-certgen-main|contour-certgen-$NEWVERS|" \
-        "-es|contour-certgen-$OLDVERS|contour-certgen-$NEWVERS|" \
+        "-es|contour-certgen-main|contour-certgen-${NEWVERS//./-}|" \
+        "-es|contour-certgen-$OLDVERS|contour-certgen-${NEWVERS//./-}|" \
+        "-es|contour-certgen-${OLDVERS//./-}|contour-certgen-${NEWVERS//./-}|" \
         "$example"
 done
 
--- a/changelogs/unreleased/5484-tsaarni-small.md
+++ b/changelogs/unreleased/5484-tsaarni-small.md
@@ -1,0 +1,1 @@
+Certgen job name in example deployment manifest is now generated using pattern `contour-certgen-v1-26-0` instead of `contour-certgen-v1.26.0` to follow Kubernetes pod naming rules.


hack/release/make-release-tag.sh
```diff
@@ -47,10 +47,12 @@ for example in examples/contour/03-envoy.yaml examples/deployment/03-envoy-deplo
...
 
 # Update the certgen job name to the new version.
+# Dot in the version is replaced with a dash to make it a valid pod name.
 for example in examples/contour/02-job-certgen.yaml ; do
     run::sed \
-        "-es|contour-certgen-main|contour-certgen-$NEWVERS|" \
-        "-es|contour-certgen-$OLDVERS|contour-certgen-$NEWVERS|" \
+        "-es|contour-certgen-main|contour-certgen-${NEWVERS//./-}|" \
+        "-es|contour-certgen-$OLDVERS|contour-certgen-${NEWVERS//./-}|" \
+        "-es|contour-certgen-${OLDVERS//./-}|contour-certgen-${NEWVERS//./-}|" \
```
tsaarni, line 55: I'm not completely sure if both these are needed? We can have old format with dots when releasing minor release from "old" branch, new version with dash in the future release branches?

sunjayBhatia: maybe be nice to do for older release branches, 1.25 supports k8s 1.27 so this warning would apply to that minor release and future patches

sunjayBhatia (commented):
I think worth adding a `small` release note for this just to call it out, in case any users have any automation etc. based on the job name

tsaarni:
@sunjayBhatia Just curious since the upgrade tests seem to fail pretty consistently with this webhook error https://github.com/projectcontour/contour/actions/runs/5292276775/jobs/9579049910, do you know what this is about? Maybe it was already discussed somewhere, but I did not find it by a quick search.

sunjayBhatia:
> @sunjayBhatia Just curious since the upgrade tests seem to fail pretty consistently with this webhook error https://github.com/projectcontour/contour/actions/runs/5292276775/jobs/9579049910, do you know what this is about? Maybe it was already discussed somewhere, but I did not find it by a quick search.

yeah, I tried to fix it with https://github.com/projectcontour/contour/pull/5457 which has some context, but turns out it wasn't sufficient, we might need to wait longer or for another condition on the webhook deployment 🤷🏽 

I'm considering maybe optionally removing the install of the gateway api webhook from the base kind cluster setup so it doesn't get reinstalled in the upgrade test scenario, not sure yet tbh

tsaarni:
> yeah, I tried to fix it with #5457 which has some context, but turns out it wasn't sufficient, we might need to wait longer or for another condition on the webhook deployment 🤷🏽

Could it make the wait to work if the admission-server had probes?

tsaarni:
> tested against a branch off the various combinations (-main, -v9.100.0, -v9-100-1) and it works 👍🏽

I'm going to merge and ignore the flaking upgrade test case.

Issue #5480, opened 2023-06-15
Gateway API: Listeners should work in IPv4 and IPv6 environments


sunjayBhatia:
**What steps did you take and what happened:**
Gateway API e2e tests in the daily build against an ipv6 cluster are failing, see https://github.com/projectcontour/contour/actions/runs/5278512063/jobs/9547950783 and other builds

This is due to here: https://github.com/projectcontour/contour/blob/45c36efd4fe37917690ea1bb2e35cecfe9b9dd7f/internal/dag/listener_processor.go#L42

At the moment all listeners listen on IPv4 only.

**What did you expect to happen:**
Gateway API tests pass and Contour with Gateway API can be run in IPv6 clusters

**Anything else you would like to add:**
- We solved this in the simple 2 listener case with the flags `--envoy-service-http[s]-address`, we could offer a new similar GatewayClass params config option to specify the listener address or repurpose these flags somehow
- One option would be to always listen on `::` since we configure Envoy listeners with IPv4 compatibility, however this may be restrictive for some environments with nodes that do not have IPv6 kernel modules enabled

PR #5486, opened 2023-06-16
internal/dag: Listener processor uses HTTP/S addresses for Gateway Listeners

sunjayBhatia:
Use existing config flags/fields for HTTP/S listen addresses to configure Envoy Listeners when in Gateway mode.

Fixes: #5480


skriss (approved):
nice, LGTM

PR #5491, opened 2023-06-16
Added probes for gateway-api-admission-server

tsaarni:


tsaarni:
@sunjayBhatia I thought simply adding probes and then waiting for available condition could fix it, but it did not. Just TCP probes due to lack of suitable HTTP end-point, but it should not matter. Quite a mystery...

tsaarni:
So is it so that Kubernetes still sends the admission control request to the old, decommissioned pod somehow?

tsaarni:
After having a better look, I suspected that the job that [patches](https://github.com/projectcontour/contour/blob/bf5cb53c309b76aa442f94570b1ad71e5120247e/examples/gateway/02-certificate_config.yaml#L124-L134) the [`ValidatingWebhookConfiguration...caBundle`](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#configure-admission-webhooks-on-the-fly) with CA certificate ([link](https://github.com/jet/kube-webhook-certgen/blob/cf887ac4a6190e91a053da8364578659cb03a1bf/pkg/k8s/k8s.go#L53)) has not completed during upgrade and therefore Kubernetes could attempt to connect to the new NEW admission-server and validate its certificate with OLD CA. "Failed to call webhook" would be result of TLS cert validation.

But 3rd attempt to re-run the test suite failed again with the same error, so the theory did not hold...

tsaarni:
Failing at 5th attempt: highly available admission server with probes + waiting for available condition AND 2 replicas.

skriss:
Just thinking about the order of operations here:

- `make setup-kind-cluster` (run as part of `make upgrade`) is installing Gateway API v0.7.1 including webhook
- The provisioner upgrade test is then running `make install-provisioner-release`, which will install the latest released provisioner (v1.25.0), including Gateway API v0.6.2 including the webhook. This is effectively _downgrading_ Gateway API and the webhook
- The provisioner upgrade test is subsequently upgrading the provisioner itself, but looks like not Gateway API, so it's remaining on v0.6.2

This is definitely a little funky, perhaps avoiding that unintentional Gateway API downgrade would help but would need to do some more experimenting.

skriss:
Another thought is perhaps at some point, certs are being re-generated, but they aren't reloaded in the admission server Deployment for "awhile" since the secret is mounted as a volume which doesn't refresh instantly when the secret itself changes.

sunjayBhatia:
yeah I was thinking about making the Gateway CRD/resource install optional in kind cluster setup so we could disable it in the upgrade test pipeline to see if that would help, assuming the downgrade/redeploy/etc. is related to the issue

tsaarni:
I experimented with the CA mismatch idea a bit outside of this PR and realized that there would be proper error message from Kubernetes if there was validation error.

If there is a downgrade scenario, and webhook could be called during that, then adding probes and setting replicas=2 for "to" version might not help until also the "from" version has them. 

BTW I wanted to instrument the code with some more debug printouts, in case there is something fundamental that is not known yet, but then the upgrade tests started to work all the sudden 😕

I guess we could simply put a sleep there to hide the problem. But in general, the gateway-api-admission-server should of course be improved if it going to be recommended for production use. It might be people bump into this problem too. 
I would assume any admission controller should be highly available component, should for example implement health endpoint.

tsaarni:
> BTW I wanted to instrument the code with some more debug printouts, in case there is something fundamental that is not known yet, but then the upgrade tests started to work all the sudden 😕

Now it started failing often again. Here is a summary [failing sequence](https://github.com/projectcontour/contour/actions/runs/5326737552/jobs/9649272977?pr=5491):

| timestamp | description |
|-----------|----------|
| 19:53:42.4271395Z | started to deploy "from" release |
| 19:53:48.520981 | admission controller running, it printed into logs `server started and listening on :8443`  |
| 19:53:48.8381070Z | `install-provisioner-release.sh` finished deployment  |
| 19:53:48.9834715Z | service endpoints now:`10.244.2.13:8443` which is the IP of the pod that is listening `:8443`|
| 19:53:49.4266897Z | create gateway class |
| 19:53:59.4316038Z |  Kubernetes times out the admission controller webhook call after 10 second timeout |

tsaarni:
Two obvious but curious observation:

1) Kubernetes does not simply get "connection refused" but the connection is really pending for 10 seconds before timeout.
2) The webhook call does not always fail with same error, for example these have happened
- failed to call webhook: Post "[https://gateway-api-admission-server.gateway-system.svc:443/validate?timeout=10s](https://gateway-api-admission-server.gateway-system.svc/validate?timeout=10s)": dial tcp 10.96.215.53:443: i/o timeout
- failed to call webhook: Post "https://gateway-api-admission-server.gateway-system.svc:443/validate?timeout=10s": context deadline exceeded

This does not seem to be unique to this case, other projects, with other admission controllers have been having tough time too https://cert-manager.io/docs/troubleshooting/webhook/

tsaarni:
With following conditions
* The server process is running, it is listening to the server port.
* Pod IP address is added to the service.
* The webhook request fails in a way that might suggest that there is some delay in setting up IP forwarding rules in the kernel for the service.

It does not seem to me like anything more could be done on our side, except sleeping for a while before triggering webhook calls.

Following would not help solving this but it would be curious to try polling the server via service IP and pod IP to see the delay in service to start working. Or attempt to use headless service to avoid iptables / virtual IP / cluster IP to be used for the service but resolve to pod IPs instead.

PR #5492, opened 2023-06-17
Fix configmap word in docs.

yanggangtony:
name: Pull request

about: The configmap is k8s's terminology , and can not be separated.
docs change 'config map to configmap , its will be more k8s terminology word.'

labels: ["release-note/not-required"]


skriss (approved):

yanggangtony:
Thanks

Issue #5458, opened 2023-06-08
Global Contour configurations should be applied in all processors


sunjayBhatia:
This isn't exclusive to this PR #5417, but we have various cluster-level settings that are supposed to be global that do not get applied to Ingress or Gateway API generated clusters. Similarly, global extauth or other features do not get applied to routes/clusters/etc. generated from Gateway API/etc. resources (see https://github.com/projectcontour/contour/discussions/5435)

We should make sure the story here is consistent and document what is not supported and/or implement these features across the board.

Notes:
- It may be more ideal to abstract some of these global configs to be applied regardless of the processors in place, so duplication of code doesn't have to happen

skriss:
xref https://github.com/projectcontour/contour/pull/5493#discussion_r1237383250

skriss:
> It may be more ideal to abstract some of these global configs to be applied regardless of the processors in place, so duplication of code doesn't have to happen

I like this idea, otherwise we get duplication of code _and_ it's easy to miss certain places that need updates.

PR #5493, opened 2023-06-17
Allow changing per_connection_buffer_limit_bytes using contour configuration for clusters

### LINKED ISSUE #1408: Lowering Envoy's cluster per-connection buffer to 32kb

rajatvig:
Fixes https://github.com/projectcontour/contour/issues/5264
Fixes #1408


tsaarni:
Looks very good! 

One quick comment that I noticed while quickly browsing the PR: new parameters in `parameters.go` would be nice to have in `site/content/docs/main/configuration.md` ([link]( https://projectcontour.io/docs/1.25/configuration/#configuration-file)).  BTW I see there are some misses from previous PRs and we are alrady getting out of sync. Too many different ways to configure I guess 😅

rajatvig:
> Looks very good!
> 
> One quick comment that I noticed while quickly browsing the PR: new parameters in `parameters.go` would be nice to have in `site/content/docs/main/configuration.md` ([link](https://projectcontour.io/docs/1.25/configuration/#configuration-file)). BTW I see there are some misses from previous PRs and we are alrady getting out of sync. Too many different ways to configure I guess 😅

Just updated the docs and also added some docs for the `max-requests-per-connection` which was not documented. 

The difference between the two PRs is that the older one had a default so a lot of tests changed. On the other hand, this one uses configuration to pass and only sets the value when it is set by the operator in the configmap or CRD which results in fewer test changes.

Just realised that the same [field](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener.proto#envoy-v3-api-field-config-listener-v3-listener-per-connection-buffer-limit-bytes) is supposed to set on the listener as well which I will follow up in a different PR.

rajatvig:
https://github.com/projectcontour/contour/pull/5513 is the other PR to set the same parameter on the Listener. Have updated links on this to point to the cluster configuration vs the listener one as earlier.

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -605,6 +605,15 @@ type ClusterParameters struct {
 	// +kubebuilder:validation:Minimum=1
 	// +optional
 	MaxRequestsPerConnection *uint32 `json:"maxRequestsPerConnection,omitempty"`
+
+	// Defines the soft limit on size of the listener’s new connection read and write buffers in bytes.
```
skriss, line 609:
```suggestion
	// Defines the soft limit on size of the cluster's new connection read and write buffers in bytes.
```

pkg/config/parameters.go
```diff
@@ -420,6 +420,13 @@ type ClusterParameters struct {
 	//
 	// +optional
 	MaxRequestsPerConnection *uint32 `yaml:"max-requests-per-connection,omitempty"`
+
+	// Defines the soft limit on size of the listener’s new connection read and write buffers
```
skriss, line 424:
```suggestion
	// Defines the soft limit on size of the cluster's new connection read and write buffers
```

skriss (changes requested):
Thanks @rajatvig! Looks great, just a couple nits

pkg/config/parameters.go
rajatvig: Fixed

site/content/docs/main/configuration.md
```diff
@@ -167,9 +167,11 @@ _This is Envoy's default setting value and is not explicitly configured by Conto
...
+| Field Name                        | Type   | Default | Description                                                                                                                                                                     |
+|-----------------------------------|--------|---------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| dns-lookup-family                 | string | auto    | This field specifies the dns-lookup-family to use for upstream requests to externalName type Kubernetes services from an HTTPProxy route. Values are: `auto`, `v4`, `v6`, `all` |
+| max-requests-per-connection       | int    | none*   | This field specifies the maximum requests for upstream connections. If not specified, there is no limit                                                                         |
+| per-connection-buffer-limit-bytes | int    | none*   | This field specifies the soft limit on size of the listener’s new connection read and write buffer. If not specified, Envoy defaults of 1MiB apply                              |
```
skriss, line 174:
```suggestion
| per-connection-buffer-limit-bytes | int    | none*   | This field specifies the soft limit on size of the cluster's new connection read and write buffer. If not specified, Envoy defaults of 1MiB apply                              |
```

site/content/docs/main/configuration.md
```diff
@@ -376,6 +378,12 @@ data:
     #   configure the cluster dns lookup family
     #   valid options are: auto (default), v4, v6, all
     #   dns-lookup-family: auto
+    #   the maximum requests for upstream connections.
+    #   If not specified, there is no limit.
+    #   Setting this parameter to 1 will effectively disable keep alive
+    #   max-requests-per-connection: 0
+    #   the soft limit on size of the listener’s new connection read and write buffers
```
skriss, line 385:
```suggestion
    #   the soft limit on size of the cluster's new connection read and write buffers
```

internal/dag/httpproxy_processor.go
```diff
@@ -100,6 +100,9 @@ type HTTPProxyProcessor struct {
 
 	// MaxRequestsPerConnection defines the maximum number of requests per connection to the upstream before it is closed.
 	MaxRequestsPerConnection *uint32
+
+	// PerConnectionBufferLimitBytes defines the soft limit on size of the listener’s new connection read and write buffers.
+	PerConnectionBufferLimitBytes *uint32
```
sunjayBhatia, line 105: I think we need to add this config to the Ingress and Gateway API processors as well to ensure this setting is applied to all generated Clusters

rajatvig: Nice catch. Just updated the PR to include both the processors.

cmd/contour/serve.go
```diff
@@ -1124,24 +1126,27 @@ func (s *Server) getDAGBuilder(dbc dagBuilderConfig) *dag.Builder {
...
+			DisablePermitInsecure:         dbc.disablePermitInsecure,
+			FallbackCertificate:           dbc.fallbackCert,
+			DNSLookupFamily:               dbc.dnsLookupFamily,
+			ClientCertificate:             dbc.clientCert,
+			RequestHeadersPolicy:          &requestHeadersPolicy,
+			ResponseHeadersPolicy:         &responseHeadersPolicy,
+			ConnectTimeout:                dbc.connectTimeout,
+			GlobalExternalAuthorization:   dbc.globalExternalAuthorizationService,
+			MaxRequestsPerConnection:      dbc.maxRequestsPerConnection,
+			PerConnectionBufferLimitBytes: dbc.perConnectionBufferLimitBytes,
```
skriss, line 1139: Need to set these on the `IngressProcessor` above as well

skriss (changes requested):

cmd/contour/serve.go
rajatvig: Fixed

site/content/docs/main/configuration.md
```diff
@@ -167,9 +167,11 @@ _This is Envoy's default setting value and is not explicitly configured by Conto
...
+| Field Name                        | Type   | Default | Description                                                                                                                                                                     |
+|-----------------------------------|--------|---------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| dns-lookup-family                 | string | auto    | This field specifies the dns-lookup-family to use for upstream requests to externalName type Kubernetes services from an HTTPProxy route. Values are: `auto`, `v4`, `v6`, `all` |
+| max-requests-per-connection       | int    | none*   | This field specifies the maximum requests for upstream connections. If not specified, there is no limit                                                                         |
+| per-connection-buffer-limit-bytes | int    | none*   | This field specifies the soft limit on size of the cluster’s new connection read and write buffer. If not specified, Envoy defaults of 1MiB apply                               |
```
skriss, line 174: I think the idea with the `*` next to defaults was to indicate defaults that _Envoy_ applied, rather than Contour -- so in this case, `per-connection-buffer-limit-bytes` should list a default of `1MiB` with a `*`, and then `max-requests-per-connection` does not need a `*` since there's no default.

If we're going to use that, please copy the footnote from `Timeout Configuration` to this section as well. It also looks like the leading `*` in that footnote was dropped at some point - please add it back so it's clear what that note refers to.

skriss (changes requested):
One more docs nit, otherwise LGTM. I think we can see if there are any opportunities for further refactoring with #5458

sunjayBhatia (approved):

PR #5512, opened 2023-06-20
Skip Gateway API install in upgrade tests

sunjayBhatia:
Test setup installs Gateway API CRDs and webhook in the initial Contour version it installs in the cluster so allow skipping cluster setup install.


skriss:
Re-running CI a few times..

skriss (approved):
upgrades looking good 💯

Issue #5263, opened 2023-04-10
Allow Listener per connection buffer limit to be configurable


skriss:
Per https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/edge, when running Envoy as an edge proxy, it's recommended to set [Listener per-connection buffer limits](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener.proto#envoy-v3-api-field-config-listener-v3-listener-per-connection-buffer-limit-bytes) to 32KiB.

There are two changes needed here:
~~- [ ] set the default value within [the Listener func](https://github.com/projectcontour/contour/blob/main/internal/envoy/v3/listener.go#L133)~~
- [ ] add a tuneable option within [ListenerParameters](https://github.com/projectcontour/contour/blob/main/pkg/config/parameters.go#L434) (for the Contour config file) and [EnvoyListenerConfig](https://github.com/projectcontour/contour/blob/main/apis/projectcontour/v1alpha1/contourconfig.go#L319) (for the ContourConfiguration CRD).

These can be done separately, in two different PRs, since the first is smaller than the second.

This also needs some level of testing to ensure it does not adversely affect existing installations.

(xref #1407, creating a new issue for clarity for ContribFest).

umutkocasarac:
I am working on it

bhavaniMiro:
I'll update the fork with tests passing too now

PR #5513, opened 2023-06-21
Allow changing per_connection_buffer_limit_bytes using contour configuration for listeners

### LINKED ISSUE #1407: Lower the per connection buffer to 64k at the listener level

rajatvig:
Fixes https://github.com/projectcontour/contour/issues/5263
Fixes #1407


internal/xdscache/v3/listener.go
```diff
@@ -383,6 +387,7 @@ func (c *ListenerCache) OnChange(root *dag.DAG) {
 				ServerHeaderTransformation(cfg.ServerHeaderTransformation).
 				NumTrustedHops(cfg.XffNumTrustedHops).
 				MaxRequestsPerConnection(cfg.MaxRequestsPerConnection).
+				PerConnectionBufferLimitBytes(cfg.PerConnectionBufferLimitBytes).
```
skriss, line 390: There are two more places in this file where an `HTTPConnectionManagerBuilder` is used and needs this to be called (one for TLS, one for TLS+fallback)

skriss (changes requested):

internal/xdscache/v3/listener.go
rajatvig: Updated

internal/envoy/v3/listener.go
```diff
@@ -279,6 +280,12 @@ func (b *httpConnectionManagerBuilder) MaxRequestsPerConnection(maxRequestsPerCo
 	return b
 }
 
+// PerConnectionBufferLimitBytes the soft limit on size of the listener’s new connection read and write buffers
+func (b *httpConnectionManagerBuilder) PerConnectionBufferLimitBytes(perConnectionBufferLimitBytes *uint32) *httpConnectionManagerBuilder {
+	b.perConnectionBufferLimitBytes = perConnectionBufferLimitBytes
```
skriss, line 285: The `Get` method also needs to be modified to generate the xds config for this setting

site/content/docs/main/configuration.md
```diff
@@ -188,9 +188,13 @@ The network configuration block can be used to configure various parameters netw
...
+|-----------------------------------|--------|---------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| connection-balancer               | string | `""`    | This field specifies the listener connection balancer. If the value is `exact`, the listener will use the exact connection balancer to balance connections between threads in a single Envoy process. See [the Envoy documentation][14] for more information. |
+| max-requests-per-connection       | int    | none    | This field specifies the maximum requests for upstream connections. If not specified, there is no limit                                                                                                                                                       |
```
sunjayBhatia, line 194:
```suggestion
| max-requests-per-connection       | int    | none    | This field specifies the maximum requests for downstream connections. If not specified, there is no limit                                                                                                                                                       |
```

internal/envoy/v3/listener.go
skriss: Yep, unfortunately we've got a lot of this

rajatvig: Had been thinking why nothing failed and suspected I am missing something. Will take a look and change.

rajatvig: Changed the code to set it on the Listener

skriss (approved):
LGTM, thanks @rajatvig!

sunjayBhatia (approved):

PR #5516, opened 2023-06-22
Enable HTTPProxy Fractional Mirroring

### LINKED ISSUE #1889: Mirroring doesn't work correctly with weights
I specified this proxy:
```YAML
kind: HTTPProxy
metadata:
  name: artifactory-v2
spec:
  routes:
    services:
    - name: artifactory-prod
      port: 80
      weight: 95
    - service: artifactory-dev
      port: 80
      weight: 4
    - service: artifactory-sampler
      port: 80
      weight: 1
      mirror: true
```

And my expectation was that 1% of traffic would be mirrored to `artifactory-sampler`. However, the envoy configuration ends up like this:
```json
         "routes": [
            {
              "match": {
                "prefix": "/v2/token/"
              },
              "route": {
                "weighted_clusters": {
                  "clusters": [
                    {
                      "name": "default/artifactory-dev/80/da39a3ee5e",
                      "weight": 4
                    },
                    {
                      "name": "default/artifactory-prod/80/da39a3ee5e",
                      "weight": 95
                    }
                  ],
                  "total_weight": 99
                },
                "request_mirror_policy": {
                  "cluster": "default/artifactory-sampler/80/da39a3ee5e"
                }
              }
            },
            {
              "match": {
                "prefix": "/v1/token/"
              },
              "route": {
                "weighted_clusters": {
                  "clusters": [
                    {
                      "name": "default/artifactory-dev/80/da39a3ee5e",
                      "weight": 4
                    },
                    {
                      "name": "default/artifactory-prod/80/da39a3ee5e",
                      "weight": 95
                    }
                  ],
                  "total_weight": 99
                },
                "request_mirror_policy": {
                  "cluster": "default/artifactory-sampler/80/da39a3ee5e"
                }
              }
            }
          ]
        },
```

Which, IIUC, sends mirrors all the traffic from the route. If weights can't be combined with mirroring, we should emit a status condition to tell the user that. However, I think that we can look at the weights and automatically set a `runtime_fraction` for the mirror, so that this behaves the way it looks.

https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route.proto#route-routeaction-requestmirrorpolicy

### LINKED ISSUE #5495: Request for Fractional Mirroring
Hello,

Envoy allows for fractional request mirroring: [Envoy Docs](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-routeaction-requestmirrorpolicy) & [Envoy proto](https://github.com/envoyproxy/envoy/blob/908af8a94fac533e693c396bad6bbbb00cc10b6f/api/envoy/config/route/v3/route_components.proto#L809).

Currently in Contour without the ability to set this attribute it defaults to mirroring 100% of traffic.

Mirroring is particularly useful when developing new applications, but being forced to mirror 100% of (usually prod) traffic is too burdensome / expensive in high RPS contexts. Being able to control the mirroring percentage would allow us to use the mirroring functionality in **many** more places.

Could a workaround be to create a mirror (to a dummy service) and use fractional splitting on that mirror to get the desired % to the target service? Would that work? Are there other options to get the same effect?

If this needs to be implemented, am happy to work on this and submit a PR.

Love to chat more about reactions to this request. Thank you,

hshamji:
Fixes #1889
Fixes #5495

internal/dag: & internal/envoy/v3:
- Updated internal MirrorPolicy struct to with a `weight` field;
- The `weight` attribute on an HTTPProxy mirrored service is overloaded to also mean an integer percent of traffic to be sent to the mirrored service;
- The GatewayAPI does not describe fractional mirroring so mirroring is kept at 100%;
- Added test confirming that HTTProxy mirror `weight` is correctly propagated to Envoy config;
- Updated docs;

Signed-off-by: Hassan Shamji [hshamji@etsy.com](mailto:hshamji@etsy.com)

### DIFF
--- a/changelogs/unreleased/5516-hshamji-small.md
+++ b/changelogs/unreleased/5516-hshamji-small.md
@@ -1,0 +1,1 @@
+Enable HTTPProxy Fractional Mirroring
--- a/changelogs/unreleased/5516-hshamji-small.md
+++ b/changelogs/unreleased/5516-hshamji-small.md
@@ -1,1 +1,1 @@
-Enable HTTPProxy Fractional Mirroring
+HTTPProxy: support fractional mirroring to a service by specifying `mirror: true` and a `weight` of 1-100. See the [traffic mirroring docs](https://projectcontour.io/docs/1.26/config/request-routing/#traffic-mirroring) for more information.
--- a/examples/contour/01-crds.yaml
+++ b/examples/contour/01-crds.yaml
@@ -5637,13 +5637,15 @@ spec:
                             minimum: 1
                             type: integer
                           mirror:
-                            description: If Mirror is true the Service will receive
+                            description: 'If Mirror is true the Service will receive
                               a read only mirror of the traffic for this route. If
-                              Mirror is true, then fractional mirroring is enabled
-                              by setting the Weight field. Omitting the Weight field
-                              or setting it to 0 will result in 100% mirroring. Setting
-                              it to an integer less than or equal to 100 will result
-                              in corresponding percentage of traffic being mirrored.
+                              Mirror is true, then fractional mirroring can be enabled
+                              by optionally setting the Weight field. Legal values
+                              for Weight are 1-100. Omitting the Weight field will
+                              result in 100% mirroring. NOTE: Setting Weight explicitly
+                              to 0 will unexpectedly result in 100% traffic mirroring.
+                              This occurs since we cannot distinguish omitted fields
+                              from those explicitly set to their default values'
                             type: boolean
                           name:
                             description: Name is the name of Kubernetes service to
@@ -6029,13 +6031,15 @@ spec:
                           minimum: 1
                           type: integer
                         mirror:
-                          description: If Mirror is true the Service will receive
+                          description: 'If Mirror is true the Service will receive
                             a read only mirror of the traffic for this route. If Mirror
-                            is true, then fractional mirroring is enabled by setting
-                            the Weight field. Omitting the Weight field or setting
-                            it to 0 will result in 100% mirroring. Setting it to an
-                            integer less than or equal to 100 will result in corresponding
-                            percentage of traffic being mirrored.
+                            is true, then fractional mirroring can be enabled by optionally
+                            setting the Weight field. Legal values for Weight are
+                            1-100. Omitting the Weight field will result in 100% mirroring.
+                            NOTE: Setting Weight explicitly to 0 will unexpectedly
+                            result in 100% traffic mirroring. This occurs since we
+                            cannot distinguish omitted fields from those explicitly
+                            set to their default values'
                           type: boolean
                         name:
                           description: Name is the name of Kubernetes service to proxy
--- a/examples/contour/01-crds.yaml
+++ b/examples/contour/01-crds.yaml
@@ -5638,7 +5638,12 @@ spec:
                             type: integer
                           mirror:
                             description: If Mirror is true the Service will receive
-                              a read only mirror of the traffic for this route.
+                              a read only mirror of the traffic for this route. If
+                              Mirror is true, then fractional mirroring is enabled
+                              by setting the Weight field. Omitting the Weight field
+                              or setting it to 0 will result in 100% mirroring. Setting
+                              it to an integer less than or equal to 100 will result
+                              in corresponding percentage of traffic being mirrored.
                             type: boolean
                           name:
                             description: Name is the name of Kubernetes service to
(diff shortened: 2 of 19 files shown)


internal/dag/httpproxy_processor.go
```diff
@@ -993,8 +993,18 @@ func (p *HTTPProxyProcessor) computeRoutes(
 				return nil
 			}
 			if service.Mirror {
-				r.MirrorPolicy = &MirrorPolicy{
-					Cluster: c,
+				// Default value if weight is omitted is 0. Setting this to 100 to retain backward-compat behavior.
+				// Edge case is a user intentionally/explicitly setting the Weight to 0 (ie. disabling the mirror).
```
skriss, line 997: Let's add some godoc to `apis/v1/projectcontour/httpproxy.go` on the Weight/Mirror fields documenting this behavior, along with the "normal" weighted mirror behavior, for users.

You'll need to run `make generate` and commit the results afterwards.

site/content/docs/1.20/config/request-routing.md
```diff
@@ -192,6 +192,7 @@ The mirror traffic is considered _read only_, any response by the mirror will be
 
 This service can be useful for recording traffic for later replay or for smoke testing new deployments.
 
+Without specifying an integer `weight` for the service, 100% of traffic will be mirrored. A non-zero integer weight <=100 directs the corresponding percentage of traffic to the mirror service.
```
skriss, line 195: Please make this change in `site/content/docs/main/config/request-routing.md` instead of this file. It'll then get copied to version-specific docs when we release.

skriss (changes requested):
Thanks for the PR @hshamji! This looks pretty good, added a couple minor comments. Will do a little manual testing as well.

skriss:
Please also see https://github.com/projectcontour/contour/actions/runs/5349229918/jobs/9700161065?pr=5516 for details on how to add a changelog file -- I've tagged this one as "small" but "minor" is also fine if you prefer.

internal/dag/httpproxy_processor.go
skriss, line 997: I'd also clarify the comment here since the user setting the weight to 0 will _not_ disable the mirror since it's indistinguishable from the "weight not specified" case; they'd need to remove the service from the route.

skriss:
Manual testing LGTM!

internal/dag/httpproxy_processor.go
hshamji: Updated and generated docs

site/content/docs/1.20/config/request-routing.md
hshamji: Removed this change and applied it to correct file

internal/dag/httpproxy_processor.go
hshamji: Clarified the edge case in the comment.

changelogs/unreleased/5516-hshamji-small.md
```diff
@@ -0,0 +1 @@
+Enable HTTPProxy Fractional Mirroring
```
skriss, line 1: Adding a few more details to the changelog:
```suggestion
HTTPProxy: support fractional mirroring to a service by specifying `mirror: true` and a `weight` of 1-100. See the [traffic mirroring docs](https://projectcontour.io/docs/1.26/config/request-routing/#traffic-mirroring) for more information.
```

hshamji: Updated. Thanks!

examples/contour/01-crds.yaml
```diff
@@ -6025,7 +6030,12 @@ spec:
                           type: integer
                         mirror:
                           description: If Mirror is true the Service will receive
-                            a read only mirror of the traffic for this route.
+                            a read only mirror of the traffic for this route. If Mirror
+                            is true, then fractional mirroring is enabled by setting
+                            the Weight field. Omitting the Weight field or setting
+                            it to 0 will result in 100% mirroring. Setting it to an
```
harshita-meena, line 6036: shouldn't we set it to 100 to send 100% traffic ?

harshita-meena: In that case, can you explicitly mention the edge case. Is there no way around the 0 value, just that the difference between a value of 0 and 1 is drastic and opposite, can we denote omission of the field by Null value instead of 0, or maybe not allowing a 0 value altogether. I can't think of a case where you will mention mirroring with 0 weights.

hshamji: I do agree; it does tempt the principle of least astonishment.

Being more explicit about that edge is a good idea regardless, ie. stating that legal values are 1-100.

I don't see an easy workaround otherwise, this overloading of the `weight` field was always a bit of a compromise:
- there isn't a non-0 null type for the go int/float; 
- changing the field type (and to what?) would break backwards-compat; 
- adding a new field explicitly, ex. `mirroring_fraction`, seemed unnecessarily complicating;

Have updated the comments in the docs & code comments. Lmk.

skriss (approved):
LGTM, thanks @hshamji!

examples/contour/01-crds.yaml
harshita-meena: comment looks perfect! allowing values from 1-100 make sense! go for the merge 🎉

sunjayBhatia (approved):

PR #5522, opened 2023-06-23
Update request-rewriting.md

BartNetJS:
fixing indent

### DIFF
--- a/site/content/docs/v1.12.0/config/request-rewriting.md
+++ b/site/content/docs/v1.12.0/config/request-rewriting.md
@@ -23,9 +23,9 @@ spec:
   - services:
     - name: s1
       port: 80
-    pathRewritePolicy:
-      replacePrefix:
-      - replacement: /new/prefix
+      pathRewritePolicy:
+        replacePrefix:
+        - replacement: /new/prefix
 ```
 
 The `replacePrefix` field accepts an array of possible replacements.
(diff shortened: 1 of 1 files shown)


sunjayBhatia:
Hey @BartNetJS I think we're going to decline this change, it is for quite an old version of these docs and is not quite correct. `pathRewritePolicy` does not live under the `service` object, it lives on the `route` (see the "Appears On" note here: https://projectcontour.io/docs/1.25/config/api/#projectcontour.io/v1.PathRewritePolicy)

BartNetJS:
Hi @sunjayBhatia , thanks for reviewing this.
I'm new to contour, I didn't notice I was working on an old version.
On the page for version 1.25 I couldn't find any information about path rewrites on the HTTPProxy: https://projectcontour.io/docs/1.25/config/fundamentals/
Are there somewhere else good examples?

I'm using currently this, and is working, but it was with trail and error:
```
kind: HTTPProxy
metadata:
  name: basic
  namespace: {{ .Values.namespace }}
spec:
  virtualhost:
    fqdn: {{.Values.hostName}}
  routes:
    - services:
      - name: jp-tracetest
        port: 11633
        conditions:
        - prefix: /tracetest
        pathRewritePolicy:
          replacePrefix:
          - replacement: /
```
That's why I proposed the change

PR #5524, opened 2023-06-23
enable websockets for Gateway listeners

### LINKED ISSUE #5241: Gateway API with websocket
When using the Kubernetes Gateway API, I don't see any obvious option to enable websocket support for contour. This is fairly annoying as I'd need to switch to the "traditional" deployment mode just for that. Is there any way to get websockets working while using the Gateway API?

skriss:
Closes #5241.

I had another look at all the context around this, and I can't see any issue with just enabling this by default, which tells envoy to pass the `Connection: Upgrade` and `Upgrade: websocket` headers to the upstream rather than returning a 403. However, given that the classic/non-Gateway API websockets behavior is and has been opt-in for a long time, I don't think we should necessarily change that now.

### DIFF
--- a/changelogs/unreleased/5524-skriss-small.md
+++ b/changelogs/unreleased/5524-skriss-small.md
@@ -1,0 +1,1 @@
+Gateway API: enable websockets upgrade by default.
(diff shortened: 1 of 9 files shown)


sunjayBhatia (approved):
seems reasonable to allow upgrading websocket upgrade generically, we can consider changing the default for other resource kinds in a future release if we want to as well for uniformity

I don't know if anyone will want to disable websocket upgrade, but we can see if that comes up

PR #5528, opened 2023-06-26
HTTPProxy: support health check status ranges

### LINKED ISSUE #5289: HTTPProxy: HTTP healthcheck expected status code should be configurable
In some cases, the [HTTPHealthCheckPolicy API](https://projectcontour.io/docs/1.24/config/api/#projectcontour.io/v1.HTTPHealthCheckPolicy) provided by HTTPProxy is not sufficient, users may want the healthcheck to pass with a status code other than 200.

An open question here is whether this should be a single status code or a list

Notes:
- Envoy API field `ExpectedStatuses`: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/health_check.proto#envoy-v3-api-field-config-core-v3-healthcheck-httphealthcheck-expected-statuses

skriss:
Adds configurable status ranges that are considered healthy for HTTP health checking.

Closes #5289.


internal/envoy/v3/healthcheck.go
```diff
@@ -32,23 +32,39 @@ func httpHealthCheck(cluster *dag.Cluster) *envoy_core_v3.HealthCheck {
 		host = hc.Host
 	}
 
-	// TODO(dfc) why do we need to specify our own default, what is the default
-	// that envoy applies if these fields are left nil?
```
skriss, line 36: Envoy does not apply defaults, these are required fields.

internal/featuretests/v3/cluster_test.go
```diff
@@ -587,35 +588,116 @@ func TestClusterWithHealthChecks(t *testing.T) {
...
+	rh.OnAdd(proxy1)
+
+	c.Request(clusterType).Equals(&envoy_discovery_v3.DiscoveryResponse{
+		Resources: resources(t,
+			clusterWithHealthCheck("default/kuard/80/bc862a33ca", "default/kuard", "default_kuard_80", "/healthz", nil),
+		),
+		TypeUrl: clusterType,
+	})
+
+	// proxy2 has valid expected status ranges.
```
skriss, line 618: I'm experimenting with feature tests only, per discussion this morning, but I can circle back and add more test cases for individual functions if we think that's worthwhile. This does a good job coverage-wise, and is (at least for this feature) more pleasant to write IMO.

site/content/docs/main/config/health-checks.md
```diff
@@ -44,6 +44,42 @@ Health check configuration parameters:
 - `timeoutSeconds`: The time to wait (seconds) for a health check response. If the timeout is reached the health check attempt will be considered a failure. Defaults to 2 seconds if not set.
 - `unhealthyThresholdCount`: The number of unhealthy health checks required before a host is marked unhealthy. Note that for http health checking if a host responds with 503 this threshold is ignored and the host is considered unhealthy immediately. Defaults to 3 if not defined.
 - `healthyThresholdCount`: The number of healthy health checks required before a host is marked healthy. Note that during startup, only a single successful health check is required to mark a host healthy.
+- `expectedStatuses`: An optional list of HTTP status ranges that are considered healthy. Ranges follow half-open semantics, meaning the start is inclusive and the end is exclusive. Statuses must be between 100 (inclusive) and 600 (exclusive).
+
+### Non-default expected statuses
+
+To consider status codes other than 200 healthy, specify `expectedStatuses` like the following:
```
sunjayBhatia, line 51: nit: one tiny thing to note would be that if you still want 200 to be valid you must make sure that is in your specified range(s)

sunjayBhatia: maybe also show an example of what the range should look like for allowing a specific status code (i.e what a range that consists of a solitary status code looks like)?

sunjayBhatia (approved):
LGTM, nice! just comments on documentation to head off some questions we might get

site/content/docs/main/config/health-checks.md
skriss: Updated in https://github.com/projectcontour/contour/pull/5528/commits/b4f3752ce777da3f970c618708bea972c6978608

PR #5533, opened 2023-06-29
make the maximum TLS version is configurable

### LINKED ISSUE #5281: Envoy Listener maximum TLS version should be configurable
To work around issues such as [this](https://github.com/envoyproxy/envoy/issues/19548), where configuring the *minimum* TLS version is not sufficient, we should also allow configuration of the maximum TLS version on Envoy Listeners.

Notes:
- validation that min and max are valid together (i.e. max is not smaller than min)
- default to 1.3 for backwards compatibility
- should go alongside:
  -  https://github.com/projectcontour/contour/blob/5b4f4eb303766dc76ae045a9be006943c6f76111/pkg/config/parameters.go#L184
  - https://github.com/projectcontour/contour/blob/5b4f4eb303766dc76ae045a9be006943c6f76111/apis/projectcontour/v1alpha1/contourconfig.go#L383
  - https://github.com/projectcontour/contour/blob/5b4f4eb303766dc76ae045a9be006943c6f76111/apis/projectcontour/v1/httpproxy.go#L390

izturn:

fix #5281


pkg/config/parameters_test.go
```diff
@@ -391,10 +391,12 @@ tls:
 
 	check(func(t *testing.T, conf *Parameters) {
 		assert.Equal(t, "1.3", conf.TLS.MinimumProtocolVersion)
+		assert.Equal(t, "1.2", conf.TLS.MaximumProtocolVersion)
 		assert.Equal(t, TLSCiphers{"ECDHE-RSA-AES256-GCM-SHA384"}, conf.TLS.CipherSuites)
 	}, `
 tls:
   minimum-protocol-version: 1.3
+  maximum-protocol-version: 1.2
```
sunjayBhatia, line 399: I think we should have validation that minimum cannot be > maximum

internal/dag/httpproxy_processor.go
```diff
@@ -243,7 +243,10 @@ func (p *HTTPProxyProcessor) computeHTTPProxy(proxy *contour_api_v1.HTTPProxy) {
 			svhost := p.dag.EnsureSecureVirtualHost(listener.Name, host)
 			svhost.Secret = sec
 			// default to a minimum TLS version of 1.2 if it's not specified
-			svhost.MinTLSVersion = annotation.MinTLSVersion(tls.MinimumProtocolVersion, "1.2")
+			svhost.MinTLSVersion = annotation.TLSVersion(tls.MinimumProtocolVersion, "1.2")
+
+			// default to a maximum TLS version of 1.3 if it's not specified
+			svhost.MaxTLSVersion = annotation.TLSVersion(tls.MaximumProtocolVersion, "1.3")
```
sunjayBhatia, line 249: I think we'll have to validate min is not > max here

internal/dag/ingress_processor.go
```diff
@@ -121,7 +121,9 @@ func (p *IngressProcessor) computeSecureVirtualhosts() {
 				svhost := p.dag.EnsureSecureVirtualHost(listener.Name, host)
 				svhost.Secret = sec
 				// default to a minimum TLS version of 1.2 if it's not specified
-				svhost.MinTLSVersion = annotation.MinTLSVersion(annotation.ContourAnnotation(ing, "tls-minimum-protocol-version"), "1.2")
+				svhost.MinTLSVersion = annotation.TLSVersion(annotation.ContourAnnotation(ing, "tls-minimum-protocol-version"), "1.2")
+				// default to a maximum TLS version of 1.3 if it's not specified
+				svhost.MaxTLSVersion = annotation.TLSVersion(annotation.ContourAnnotation(ing, "tls-maximum-protocol-version"), "1.3")
```
sunjayBhatia, line 126: same validation issue here as well

izturn:
@sunjayBhatia  PTAL

internal/annotation/annotations.go
```diff
@@ -56,6 +56,7 @@ var annotationsByKind = map[string]map[string]struct{}{
 		"projectcontour.io/response-timeout":             {},
 		"projectcontour.io/retry-on":                     {},
 		"projectcontour.io/tls-minimum-protocol-version": {},
+		"projectcontour.io/tls-maximum-protocol-version": {},
```
skriss, line 59: Please document this in https://projectcontour.io/docs/main/config/annotations/#contour-specific-ingress-annotations

internal/dag/httpproxy_processor.go
```diff
@@ -250,8 +250,17 @@ func (p *HTTPProxyProcessor) computeHTTPProxy(proxy *contour_api_v1.HTTPProxy) {
...
+			if svhost.MaxTLSVersion < svhost.MinTLSVersion {
+				validCond.AddError(contour_api_v1.ConditionTypeTLSError, "TLSConfigNotValid",
+					"Spec.Virtualhost.TLS the minimum protocol version is greater than the maximum protocol version")
+				return
+			}
```
skriss, line 263: Let's add a test case to either status_test.go or to a feature test that covers this.

internal/dag/ingress_processor.go
```diff
@@ -126,7 +126,19 @@ func (p *IngressProcessor) computeSecureVirtualhosts() {
...
+
+				if svhost.MaxTLSVersion < svhost.MinTLSVersion {
+					p.WithError(err).
+						WithField("name", ing.GetName()).
+						WithField("namespace", ing.GetNamespace()).
+						WithField("minTLSVersion", svhost.MinTLSVersion).
+						WithField("maxTLSVersion", svhost.MaxTLSVersion).
+						Errorf("error TLS protocol version")
+					return
+				}
```
skriss, line 141: Let's add a test case (status_test.go or a feature test)

internal/dag/ingress_processor.go
```diff
@@ -126,7 +126,19 @@ func (p *IngressProcessor) computeSecureVirtualhosts() {
...
+				// default to a maximum TLS version of 1.3 if it's not specified
+				svhost.MaxTLSVersion = annotation.TLSVersion(annotation.ContourAnnotation(ing, "tls-maximum-protocol-version"), "1.3")
+
+				if svhost.MaxTLSVersion < svhost.MinTLSVersion {
+					p.WithError(err).
+						WithField("name", ing.GetName()).
+						WithField("namespace", ing.GetNamespace()).
+						WithField("minTLSVersion", svhost.MinTLSVersion).
+						WithField("maxTLSVersion", svhost.MaxTLSVersion).
+						Errorf("error TLS protocol version")
```
skriss, line 139: Please make the error message more specific similar to what's used in the HTTPProxy processor

internal/xdscache/v3/listener.go
```diff
@@ -485,11 +498,13 @@ func (c *ListenerCache) OnChange(root *dag.DAG) {
 			// Secret is provided when TLS is terminated and nil when TLS passthrough is used.
 			if vh.Secret != nil {
 				// Choose the higher of the configured or requested TLS version.
-				vers := max(cfg.minTLSVersion(), envoy_v3.ParseTLSVersion(vh.MinTLSVersion))
+				minVer := max(cfg.minTLSVersion(), envoy_v3.ParseTLSVersion(vh.MinTLSVersion))
+				maxVer := max(cfg.maxTLSVersion(), envoy_v3.ParseTLSVersion(vh.MaxTLSVersion))
```
skriss, line 502: The logic for minimum TLS version enforces that the minimum TLS version on a specific HTTPProxy is _at least_ as high as the min TLS version configured in the global Contour config.

For max TLS version, I think the corresponding logic would be that the global Contour config defines a ceiling on the max TLS version, and individual HTTPProxies can't exceed that max -- e.g. if the global config says max TLS version = 1.2, and a specific proxy says 1.3, then the effective max TLS version for that proxy is 1.2.

So, we'd want to use a `min` function to get the effective max TLS version.

cc @sunjayBhatia -- does this sound reasonable to you, given how the min TLS version logic works?

pkg/config/parameters.go
```diff
@@ -183,6 +184,7 @@ func (n NamespacedName) Validate() error {
 // TLSParameters holds configuration file TLS configuration details.
 type TLSParameters struct {
 	MinimumProtocolVersion string `yaml:"minimum-protocol-version"`
+	MaximumProtocolVersion string `yaml:"maximum-protocol-version"`
```
skriss, line 187: Please document this in https://projectcontour.io/docs/main/configuration/#tls-configuration

skriss (changes requested):
Thanks for the PR @izturn, had a few comments

internal/xdscache/v3/listener.go
izturn: @sunjayBhatia  WDYT?

sunjayBhatia: yep the logic @skriss describes sounds reasonable to me 👍🏽

izturn: ok

site/content/docs/main/configuration.md
```diff
@@ -111,7 +111,8 @@ Contour should provision TLS hosts.
...
-| minimum-protocol-version | string   | `1.2`                                                                                                             | This field specifies the minimum TLS protocol version that is allowed. Valid options are `1.2` (default) and `1.3`. Any other value defaults to TLS 1.2.                                                                                                                                                                                                                                                                                                                                                                                                                                                                         |
+| minimum-protocol-version | string   | `1.2`                                                                                                             | This field specifies the maximum TLS protocol version that is allowed. Valid options are `1.2` (default) and `1.3`. Any other value defaults to TLS 1.2.
```
sunjayBhatia, line 114:
```suggestion
| minimum-protocol-version | string   | `1.2`                                                                                                             | This field specifies the minimum TLS protocol version that is allowed. Valid options are `1.2` (default) and `1.3`. Any other value defaults to TLS 1.2.                                                                                                                                     
```

internal/xdscache/v3/listener_test.go
```diff
@@ -902,9 +913,11 @@ func TestListenerVisit(t *testing.T) {
 				SocketOptions: envoy_v3.TCPKeepaliveSocketOptions(),
 			}),
 		},
-		"tls-min-protocol-version from config overridden by httpproxy": {
+
+		"tls-protocol-version from config overridden by httpproxy": {
```
sunjayBhatia, line 917: i think we're missing a case where the maximum is 1.2 but an httpproxy tries to override it to 1.3, the resulting config should still have a max of 1.2

pkg/config/parameters_test.go
```diff
@@ -390,15 +393,22 @@ tls:
...
 	}, `
 tls:
-  minimum-protocol-version: 1.3
+  minimum-protocol-version: 1.2
+  maximum-protocol-version: 1.3
   cipher-suites:
   - ECDHE-RSA-AES256-GCM-SHA384
 `)
 
+	checkFail(`
```
sunjayBhatia, line 407: for full coverage looks like we can add a couple cases where min/max tls protocols provided are invalid

internal/featuretests/v3/tlsprotocolversion_test.go
```diff
@@ -87,29 +87,42 @@ func TestTLSMinimumProtocolVersion(t *testing.T) {
...
-				SecretName: sec1.Name,
-			}},
-			Rules: []networking_v1.IngressRule{{
-				Host: "kuard.example.com",
-				IngressRuleValue: networking_v1.IngressRuleValue{
-					HTTP: &networking_v1.HTTPIngressRuleValue{
-						Paths: []networking_v1.HTTPIngressPath{{
-							Backend: *featuretests.IngressBackend(s1),
-						}},
+	makeIngrgess := func(minVer, maxVer string) *networking_v1.Ingress {
```
sunjayBhatia, line 90:
```suggestion
	makeIngress := func(minVer, maxVer string) *networking_v1.Ingress {
```

internal/dag/builder_test.go
```diff
@@ -12460,6 +12499,7 @@ func TestDAGInsert(t *testing.T) {
 				},
 			),
 		},
+
```
sunjayBhatia, line 12502: it would be great if we could omit these extra whitespace changes

changelogs/unreleased/5533-izturn-small.md
```diff
@@ -0,0 +1 @@
+Let the maximum TLS version on Envoy Listeners is configurable, the valid options are `1.2` and `1.3`, it's defaults to TLS 1.3.
```
sunjayBhatia, line 1:
```suggestion
The maximum TLS version on Envoy Listeners can be configured. Valid options are TLS versions `1.2` and `1.3` with a default of `1.3`.
```

sunjayBhatia (commented):
overall good, very close, just a few test nitpicks and wording changes

one more thing I think Steve mentioned would be to have an explicit test in status_test.go for the new status reason/message added in the httpproxy processor, its important to have some explicit testing for these since it effectively becomes API surface once we expose things in status

izturn:
> overall good, very close, just a few test nitpicks and wording changes
> 
> one more thing I think Steve mentioned would be to have an explicit test in status_test.go for the new status reason/message added in the httpproxy processor, its important to have some explicit testing for these since it effectively becomes API surface once we expose things in status

thx, done

sunjayBhatia (approved):
rebased on main and did a little refactor/fix of my own to prevent another round trip of comments, LGTM but should get another review from @projectcontour/maintainers

skriss (approved):
LGTM, thanks @izturn and @sunjayBhatia!

Issue #4637, opened 2022-07-25
Contour support for Envoy's  stats per route


izturn:
**Please describe the problem you have**
     At https://github.com/envoyproxy/envoy/issues/3351 @stevesloka  advise envoy to expose metrics per vhost, now this feature has been released along with envoy v1.23 as [route-stat-prefix](https://www.envoyproxy.io/docs/envoy/v1.23.0/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-route-stat-prefix) (the pr is https://github.com/envoyproxy/envoy/pull/21302), shall we want to support it too?

wilsonwu:
Any plan for this feature support?

sunjayBhatia:
Seems reasonable to support with a few considerations:
- document which stats will be enabled (https://www.envoyproxy.io/docs/envoy/v1.23.0/configuration/http/http_filters/router_filter#config-http-filters-router-vcluster-stats)
- we should document the resource impact this will have to each instance of Envoy (from the Envoy docs `We do not recommend setting up a stat prefix for every application endpoint. This is both not easily maintainable and statistics use a non-trivial amount of memory(approximately 1KiB per route).`)
- do a bench/load test to see if it has a noticeable impact the way Contour programs routes etc.
  - this will mean 
- consider if this should be an opt-in feature

We would definitely take community contributions to help speed this up, otherwise we've got this prioritized for 1.24.0 currently

wilsonwu:
We are planning to do tests for this feature, update later.

wilsonwu:
Load test results come:

We did two rounds load test:
1. 10k routes without vhost metrics
2. 10k routes with vhost metrics

All tests env is 1 instance 4C4G envoy v1.23

Test results:
The 1st round:
Below image show envoy start CPU and memory: CPU 2%, memory 6% (almost 250m)
<img width="1032" alt="image" src="https://user-images.githubusercontent.com/1269496/197752257-1309f9f8-2772-4e57-b216-9add6474a74a.png">
After sent requests to 10k routes randomly, CPU and memory like below: CPU 400%, memory 8% (almost 350m)
<img width="1063" alt="image" src="https://user-images.githubusercontent.com/1269496/197756882-76adfef3-f121-4b68-a730-118b1bea2bf4.png">

The 2nd round:
Below image show envoy start CPU and memory: CPU 2% - 3%, memory 7.5% (almost 330m)
[...]

PR #5535, opened 2023-06-29
Add envoy stat_prefix support for httpproxy

therealak12:
This pull request addresses [issue#4637](https://github.com/projectcontour/contour/issues/4637).
I'm not sure if this is a perfect implementation but I thought receiving feedback could help me implement it more properly.

Currently it adds a field in the envoy stanza of contour configuration, which, if set, will also be set in the envoy route config.

Another possible implementation would be to add a field to HTTPProxy spec, allowing users to set different `StatPrefix`es per route.

Also, as a third way to implement, we can set `resource.name._resource.kind_resource.namespace` as the stat_prefix!


therealak12:
Screenshot from 2023-06-29 17-51-42

Here is a demo of the current implementation.

izturn:
I prefer this to  xref: https://github.com/projectcontour/contour/pull/5539

therealak12:
> Do we think users will need to configure this per-route stat prefix? Is there any value in making this configurable? If not we can probably get away with just adding a boolean flag and configuring the stat prefix using some details of the Envoy (hashing the route match parameters for example) route to ensure uniqueness/ability to differentiate between different routes
> 
> In general this seems like it achieves what #4637 requires, an opt-in ability to signal Envoy to collect per-route upstream stats

[This PR](https://github.com/projectcontour/contour/pull/5539) does something like what you've explained.

sunjayBhatia:
> > Do we think users will need to configure this per-route stat prefix? Is there any value in making this configurable? If not we can probably get away with just adding a boolean flag and configuring the stat prefix using some details of the Envoy (hashing the route match parameters for example) route to ensure uniqueness/ability to differentiate between different routes
> > In general this seems like it achieves what #4637 requires, an opt-in ability to signal Envoy to collect per-route upstream stats
> 
> [This PR](https://github.com/projectcontour/contour/pull/5539) does something like what you've explained.

I think it would be useful to modify that PR with a generated stat prefix that differentiates between different Envoy routes (maybe in addition to or instead of the existing stat prefix that differentiates between originating resources) and test it out with a couple different HTTPProxies/Ingresses/HTTPRoutes with multiple route rules

(I would guess that using different stat prefixes could possibly expand the memory footprint that was analysed in https://github.com/projectcontour/contour/issues/4637 as well, but would have to confirm to be sure)

cc @izturn @wilsonwu from #4637 is the goal to see stats between different route matches (e.g. see the upstream stats for requests to /foo vs /bar), if so I don't think the existing PRs do so?

therealak12:
@sunjayBhatia 
Does it look good to you?

image

Please don't worry about the prefix section for now. It is just a temporary thing that I put in place to make the PoC work. I will improve the `String()` method if it is accepted.

These are the stats for an httpproxy with the following spec:
```yaml
routes:
- conditions:
  - prefix: /etc
  permitInsecure: true
  services:
  - name: python-http-server
    port: 8000
- conditions:
  - prefix: /tmp
  permitInsecure: true
  services:
  - name: python-http-server
    port: 8000
virtualhost:
  fqdn: some-random-path.apps.private.okd4.ts-2.staging-snappcloud.io
```

It sets StatPrefix to a string built like this:
```go
fmt.Sprintf("%s_%s_%s_%s", httpRoute.GetNamespace(), "httproute", httpRoute.GetName(), r.PathMatchCondition.String())
```

sunjayBhatia:
this seems more usable and now does separate the stats for different route rules

we'll have to figure out a stat prefix we can generate that can express uniqueness, but also can be used by an operator to figure out what route theyre looking at (e.g. not just an opaque hash)

would love to get @izturn and @wilsonwu's thoughts since they originally requested this

therealak12:
> this seems more usable and now does separate the stats for different route rules
> 
> we'll have to figure out a stat prefix we can generate that can express uniqueness, but also can be used by an operator to figure out what route theyre looking at (e.g. not just an opaque hash)
> 
> would love to get @izturn and @wilsonwu's thoughts since they originally requested this

We had a conversation about the hash which I've replied here:
https://kubernetes.slack.com/archives/C8XRH2R4J/p1689159578352139?thread_ts=1688670204.661949&cid=C8XRH2R4J

It seems the _60 characters limit_ can be safely removed now.

therealak12:
One point to add here,

It's not that trivial to include the `matchCondition` in the stats.

There are various cases such as regex, exact, and prefix match. To distinguish between two routes, one having `exactMatch` set to `/sample` and the other having `prefix` set to `/sample` we should include the type of the match condition in the stat too.

Besides, there are query and header match conditions too.

A possible format for the stat_prefix would be:

```
namespace_ingressResourceType_ingressResourceName_matchType_matchValue
```

where `matchType` can be one of regex, prefix, or exact.

Although this supports all 3 match types, it doesn't count headers or query matches in. Trying to put those in the prefix makes it too complex and even useless.

It would be nice if we agree on a format before going on.

izturn:
@therealak12 , thx for your pr, i am ok with it

sunjayBhatia:
> @therealak12 , thx for your pr, i am ok with it

can you clarify a bit more @izturn how do you intend to use the new stats? specifically around the last bullet here: https://github.com/projectcontour/contour/pull/5535#issuecomment-1638789265

wilsonwu:
Sorry for the late reply, I think per route stats is enough for our current situation, if some use cases for per virtualhost, just calculated by metrics system is OK.

izturn:
@sunjayBhatia Sorry for the late reply, As @wilsonwu said, I think per route stats is enough for our use case

therealak12:
As @sunjayBhatia pointed above, the per-route metrics are located at the highest level of the granularity pyramid:
> do we need the granularity of per-Route rule stats

So let's discuss whether **per-virtualhost** metrics is enough or if we should agree on a format for presenting **per-route** metrics. In our use case, we are happy with only including the `prefix` from `matchCondition`s. In general, there are several `matchConditions` that can differentiate routes. Including all of them results in excessively lengthy metric names.

Your valuable comment will be crucial in reaching the best decision. Thanks in advance.

@wilsonwu
@izturn

wilsonwu:
> As @sunjayBhatia pointed above, the per-route metrics are located at the highest level of the granularity pyramid:
> 
> > do we need the granularity of per-Route rule stats
> 
> So let's discuss whether **per-virtualhost** metrics is enough or if we should agree on a format for presenting **per-route** metrics. In our use case, we are happy with only including the `prefix` from `matchCondition`s. In general, there are several `matchConditions` that can differentiate routes. Including all of them results in excessively lengthy metric names.
> 
> Your valuable comment will be crucial in reaching the best decision. Thanks in advance.
> 
> @wilsonwu @izturn

Thanks for these information, I our use case, we make envoy as an API gateway, so users pay more attention to the API, and as my early reply I thnk if we have per route metrics, the per virtualhost and per envoy metrics can be caculated be themselves, anyway, support per virtualhost metrics directly is better.

sunjayBhatia:
> @sunjayBhatia Based on the discussion, it seems providing per-virtual-host metrics is enough. I'd like to hear your thoughts on this.

to summarize:
- it looks like the "per route" metrics that are important to @wilsonwu /@izturn are in fact actually "per vhost", so aggregating request metrics needs to be done with that in mind when making up the stat prefix
- the new stats will be rooted at `vhost.<virtual host name>.route.<stat_prefix>` so a static `stat_prefix` seems sufficient, since we don't need to add any further granularity
- to be conservative and make this an additive change, how users opt-in to these new stats could be a boolean or letting users configure the route `stat_prefix`, we can discuss the merits of either approach but I'm leaning towards a bool opt-in to require less validation etc.
  - furthermore, is it sufficient to have this as a global opt-in or per HTTPProxy? I would argue for it to be a global config since this is an operator-level configuration that could affect Envoy's performance
- as an aside, if we did want to add *some additional granularity* we could (generate the stat prefix based on the resource that originated the Route for example), and users who don't care about that additional detail can maybe aggregate the statistic themselves (though we would have to think about that UX in more detail)

izturn:
hey @therealak12, thx 4  ur work,  any progress?

therealak12:
> hey @therealak12, thx 4 ur work, any progress?

Hey.
I'm busy these days. I'll update this thread soon.

We need to agree on some design decisions first.

- I agree with having a boolean to enable/disable the `stat_prefix`. However, I think a global one is better than adding a per-resource field. The reason is that it fits our requirements better. This is the first thing to decide.

- Also, I think including the name of the resource in the stat_prefix is fine and better than not doing that. This is the second decision.

If these decisions are accepted, I'll update this MR or create a new one.

wilsonwu:
> > hey @therealak12, thx 4 ur work, any progress?
> 
> Hey. I'm busy these days. I'll update this thread soon.
> 
> We need to agree on some design decisions first.
> 
> * I agree with having a boolean to enable/disable the `stat_prefix`. However, I think a global one is better than adding a per-resource field. The reason is that it fits our requirements better. This is the first thing to decide.
> * Also, I think including the name of the resource in the `stat_prefix` is fine and better than not doing that. This is the second decision.
> 
> If these decisions are accepted, I'll update this MR or create a new one.

- Agree the global enable/disable for `stat_prefix`.
- And for include a resource name or not, I think currently the path is OK, no more voice for this requirement, I am fine for it.

therealak12:
So I've just (force) pushed a new commit. If the `enableStatPrefix` configuration is true, it sets `stat_prefix` to `<resource_namespace>_<resource_name>`.

izturn:
@therealak12 any progress?

therealak12:
> @therealak12 any progress?

This is actually ready for review. I'll resolve merge conflicts soon.

therealak12:
Mark as fresh👋

therealak12:
Mark as fresh again👋

davinci26:
@therealak12 are you still motivated to push this through? I can probably shepherd it.

cc @sunjayBhatia

therealak12:
Absolutely @davinci26 ! I'm open to any modifications.

The current version of the code contains the latest changes as per consensus in the discussion, however, feel free to also 
suggest or implement changes as you see fit.

davinci26 (commented):
I am checking this out locally and the first comment is that we should also add this flag in `contour/pkg/config/parameters.go` so users can configure this with `contour-config.yaml`

davinci26:
there is a bug with envoy `docker.io/envoyproxy/envoy:v1.27.2` that doesn't set those as label correctly:

`envoy_vhost_bar_com_route_default_basic_upstream_rq_retry{envoy_virtual_host="foo-basic"} 0` note `default_basic` is part of the stat name and not part of a `label` so that would require `Envoy` cleanup otherwise the cardinality might be too high for large clusters

davinci26:
I feel like we are circling back a bit but I would think that a prefix in the form `namespace_ingressResourceType_ingressResourceName_matchType_matchValue` would be more useful. Let me catch up tho and see why we decided against it

therealak12:
> I feel like we are circling back a bit but I would think that a prefix in the form `namespace_ingressResourceType_ingressResourceName_matchType_matchValue` would be more useful. Let me catch up tho and see why we decided against it

This is the way we're using it now in [another fork](https://github.com/snapp-incubator/contour/blob/snappcloud/internal/dag/httpproxy_processor.go#L856). As we use the `path` matchType only, we include it in the stat prefix.

Including all match types can make the metric name too large. There might be several query or header matches. This is why this PR doesn't include them in the stat prefix.

therealak12:
> I am checking this out locally and the first comment is that we should also add this flag in `contour/pkg/config/parameters.go` so users can configure this with `contour-config.yaml`

Thanks. I thought the `contourconfiguration` resource had deprecated the configuration file, but I've just read the docs and there's no deprecation warning. So I'll add it to the `parameters.go` too.

davinci26:
I see the `envoy` bug, which we should either fix in envoy and wait for the release or work around it here.

The name: `foo-basic.bar.com` of the route is usually the `fqdn` which contains `.`, those dots are not handled correctly by envoy `regex` for transforming the `statsd` to `prometheus` format so the generated stat name is messed up

davinci26:
> This is the way we're using it now in [another fork](https://github.com/snapp-incubator/contour/blob/snappcloud/internal/dag/httpproxy_processor.go#L856). As we use the path matchType only, we include it in the stat prefix.

I wonder if we can create a `route_id` that is shorter, I think it would be useful to have this as a label actually

davinci26 (changes requested):
I wanted to consolidate a bit my feedback on the PR:

1. we should fix `contourconfiguration` 
2. On the route id I think we should:

* work around the Envoy problem with `dots` on the `stat_prefix` by offering some guardrails.

Given that we have many users wanted the `stat_prefix` to be different and have slightly different requirements and appetite for customizations I wonder if this already a good hint that we should delegate this to the user.

I am thinking that we can have a `stat_prefix` format as a global input that would allow us to customize the `stat_prefix` depending on how much cardinality we want in each cluster.

What are you thinking @projectcontour/maintainers  around the `api` part?

izturn:
We are trying to use this feature in our products. Yes, the `dot` in fqdn may cause the generated metrics  to be a bit strange, but I think this is also acceptable.

eg: 
```go
envoy_vhost_abc_def_com_route_default_home_upstream_rq_xx{envoy_response_code_class="2",envoy_virtual_host="xyz-123"}  

// fqdn: xyz-123.abc-def.com 
// namespace: default
// name: home
```

therealak12:
> > This is the way we're using it now in [another fork](https://github.com/snapp-incubator/contour/blob/snappcloud/internal/dag/httpproxy_processor.go#L856). As we use the path matchType only, we include it in the stat prefix.
> 
> I wonder if we can create a `route_id` that is shorter, I think it would be useful to have this as a label actually

What do you mean by `route_id`? If you mean creating something to use instead of the route name, it won't work for us. We need the names.
If something else, could you please elaborate?

therealak12:
> I see the `envoy` bug, which we should either fix in envoy and wait for the release or work around it here.
> 
> The name: `foo-basic.bar.com` of the route is usually the `fqdn` which contains `.`, those dots are not handled correctly by envoy `regex` for transforming the `statsd` to `prometheus` format so the generated stat name is messed up

Envoy replaces many symbols (e.g. `/` and `:`) with an underscore. This should and can be fixed in envoy, not in contour.

therealak12:
> I am checking this out locally and the first comment is that we should also add this flag in `contour/pkg/config/parameters.go` so users can configure this with `contour-config.yaml`

I've just made a change to address your comment. I'd appreciate it if you take a look. Thanks.

dnbn:
Thanks for the PR @therealak12, I'am also very much interested by this feature.

chaosbox:
Thankyou for the PR, and have been waiting for a while now, we would be happy to test / provide initial feedback if reqd.

izturn:
any progress?

therealak12:
Dear @izturn @chaosbox,
this feature is a controversial one. We've proceeded with a custom implementation that satisfies our minimum requirements on a separate fork.
IMO, implementing a format that satisfies every project's requirements is not possible.

therealak12:
Going to close the PR.
Kindly please open a new one if you're interested in implementing the feature.

PR #5539, opened 2023-07-01
Set specific stat_prefix for each resource

therealak12:
This is another way of setting StatPrefix as mentioned in [pull#5535](https://github.com/projectcontour/contour/pull/5535).

It sets `namespace_resourcetype_resourcename` as the StatPrefix.


wilsonwu:
Seems need to reopen.

therealak12:
> Seems need to reopen.

I've applied the changes in PR#5535. This is not required anymore.

Issue #5542, opened 2023-07-04
Execution order of HTTP External Auth


yangyy93:
**What steps did you take and what happened:**
After I enabled GlobalExternalAuthorization in contourConfig, I found that the execution order of[ the ext_authz filter in the http filter is after the global-ratelimit filter](https://github.com/projectcontour/contour/blob/main/internal/xdscache/v3/listener.go#L387-L388), and the [execution order in the https filter is before the global-ratelimit ](https://github.com/projectcontour/contour/blob/main/internal/xdscache/v3/listener.go#L440). I think it is better to have a uniform order of execution.

sunjayBhatia:
Yep that is a good point, we should ensure these are consistent

sunjayBhatia:
See also https://github.com/projectcontour/contour/issues/3726 (most directly related) and https://github.com/projectcontour/contour/issues/4826 if you have any input there as well

PR #5559, opened 2023-07-17
Fix: Make order of Ext auth and Global ratelimit uniform

clayton-gonsalves:
This PR ensures that the order of execution of extauth and global ratelimit is the same across HTTP and HTTPS virtual hosts, which is Auth goes first, then Global rate limit.

Fixes: https://github.com/projectcontour/contour/issues/5542


shadialtarsha (approved):
We can add a comment to the if statement saying that the order is respected here.

changelogs/unreleased/5544-clayton-gonsalves-small.md
```diff
@@ -0,0 +1,4 @@
+## Fix order of Global ExtAuth and Global Ratelimit
```
skriss, line 1: We use PR #'s in the changelog filenames, so for consistency, please rename to `5559-clayton-gonsalves-small.md`.

internal/xdscache/v3/listener.go
```diff
@@ -389,8 +391,8 @@ func (c *ListenerCache) OnChange(root *dag.DAG) {
 				NumTrustedHops(cfg.XffNumTrustedHops).
 				MaxRequestsPerConnection(cfg.MaxRequestsPerConnection).
 				Tracing(envoy_v3.TracingConfig(envoyTracingConfig(cfg.TracingConfig))).
```
skriss, line 393: Tangentially related, looks like tracing is also in a different order between the two; for HTTP vhosts tracing comes before ExtAuth whereas for TLS vhosts it's after ExtAuth. Ideally we can make that consistent as well; @yangyy93 do you have thoughts on the preferred order for tracing?

yangyy93: The [order of envoy's filters](https://www.envoyproxy.io/docs/envoy/v1.26.3/intro/arch_overview/http/http_filters#arch-overview-http-filters-ordering) is important, because envoy will process traffic according to the order of filters. It is good to process auth first, and then global rate limit. Tracing configuration takes effect in envoy and the code order in contour is irrelevant. It would be better if the sequence consistency can be maintained in http vhosts and tls vhosts.

skriss: Ah good call, I was thinking tracing was an HTTP filter but that's not the case.

skriss (approved):

sunjayBhatia (approved):

PR #5565, opened 2023-07-17
test/conformance: only skip specific failing test case

skriss:
Lets us run the rest of the test cases in this particular test.


skriss:
Heh, this is definitely not supported by Gateway API :/

PR #5566, opened 2023-07-18
test/conformance: only skip specific failing test case

skriss:
Take 2 on this, using `go test -skip`. It's a little uglier, but IMO worth it to get the rest of the test coverage from that test.

See https://github.com/skriss/contour/actions/runs/5591856217/jobs/10223702289 for a verbose run that shows the individual test case being skipped.

### DIFF
--- a/test/scripts/run-gateway-conformance.sh
+++ b/test/scripts/run-gateway-conformance.sh
@@ -52,15 +52,22 @@ EOF
 GO_MOD_GATEWAY_API_VERSION=$(grep "sigs.k8s.io/gateway-api" go.mod | awk '{print $2}')
 
 if [ "$GATEWAY_API_VERSION" = "$GO_MOD_GATEWAY_API_VERSION" ]; then
-  go test -timeout=40m -tags conformance ./test/conformance/gatewayapi --gateway-class=contour
+  # Skipped test checks for the original request port in the returned Location
+  # header which Envoy is stripping.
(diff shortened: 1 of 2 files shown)


test/scripts/run-gateway-conformance.sh
```diff
@@ -52,15 +52,22 @@ EOF
 GO_MOD_GATEWAY_API_VERSION=$(grep "sigs.k8s.io/gateway-api" go.mod | awk '{print $2}')
 
 if [ "$GATEWAY_API_VERSION" = "$GO_MOD_GATEWAY_API_VERSION" ]; then
-  go test -timeout=40m -tags conformance ./test/conformance/gatewayapi --gateway-class=contour
+  # Skipped test checks for the original request port in the returned Location
+  # header which Envoy is stripping.
+  # See: https://github.com/envoyproxy/envoy/issues/17318
+  # Skip here instead of in the suite options since it's a subtest.
+  go test -timeout=40m -tags conformance ./test/conformance/gatewayapi --gateway-class=contour \
+    -skip TestGatewayConformance/HTTPRouteRedirectPortAndScheme/http-listener-on-8080/0_request_to_\'\/scheme-nil-and-port-nil\'_should_receive_a_302
```
sunjayBhatia, line 60: We'll have to see how this works with the new conformance reporting suite if the report can still be generated accurately

PR #5585, opened 2023-07-25
Cleanup: Use Envoy contains for header routing

### LINKED ISSUE #5572: Cleanup: Use Envoy `contains` for header routing

In `func containsMatch(s string, ignoreCase bool) *envoy_route_v3.HeaderMatcher_StringMatch`  https://github.com/projectcontour/contour/blob/cca06915b4c3a603d87ee7857cafaba19e4bac74/internal/envoy/v3/route.go#L826

we implement it with `safe_regex` matching but we can switch to use `contains` which is the native way of envoy to do this 

See https://www.envoyproxy.io/docs/envoy/latest/api-v3/type/matcher/v3/string.proto

davinci26:
Fixes #5572

### DIFF
--- a/internal/envoy/v3/route.go
+++ b/internal/envoy/v3/route.go
@@ -17,7 +17,6 @@ import (
 	"bytes"
 	"fmt"
 	"net/http"
-	"regexp"
 	"sort"
 	"strings"
 	"text/template"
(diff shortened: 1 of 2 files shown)


davinci26:
I marked as [release-note/none-required](https://github.com/projectcontour/contour/labels/release-note%2Fnone-required) to make the CI happy, but happy to add one if you think it is needed

skriss (approved):
LGTM, did some ad-hoc testing and all behavior seems to stay consistent

skriss:
Not sure why gateway-conformance is giving so much trouble on this one; I was able to run successfully locally. Will give it one more try.

PR #5597, opened 2023-07-26
Add dependabot config for release branches

sunjayBhatia:
Add go and github action bumps for only patch versions on release branches

Also updates release process for minor releases to mention this config should be updated

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,9 @@
 
 version: 2
 updates:
-- package-ecosystem: "gomod"
+# main branch targets
+- target-branch: main
+  package-ecosystem: "gomod"
   directory: "/"
   schedule:
     interval: "weekly"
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -22,3 +25,92 @@ updates:
...
+# release branch N targets
+- target-branch: release-1.25
+  package-ecosystem: "gomod"
+  directory: "/"
+  schedule:
+    interval: "weekly"
+  ignore:
+  - update-types:
+    - "version-update:semver-major"
+    - "version-update:semver-minor"
```
skriss, line 37: docs aren't totally clear, but do we need need to add a `dependency-name: *` here? ref. https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#specifying-dependencies-and-versions-to-ignore

sunjayBhatia: probably won't hurt to do so, let's try it!

skriss (approved):
nice, LGTM!

PR #5651, opened 2023-08-02
Fix Gateway Status Addresses Updater with multiple Ingress entries.

### LINKED ISSUE #5650: [provisioner] Gateway.status.addresses is truncated to the first address only.

Unlike `HttpProxy.status` which is populated with all public addresses, the `Gateway.status.addresses` field is only populated with the first address.

HttpProxy.status:
```yaml
    loadBalancer:
      ingress:
      - ip: 192.168.1.2
      - ip: 2a0c:abcd:ef01::1
``` 

Gateway.status:
```yaml
    addresses:
    - type: IPAddress
      value: 192.168.1.2
```

That the `Gateway` object is updated to expose all IPs declared on the envoy service, like `HttpProxy` does.

Status update is done here:

https://github.com/projectcontour/contour/blob/d60959ec98d46df61de6127464677fa571fd2dae/internal/k8s/statusaddress.go#L187-L195

**Environment:**

- Contour version: 1.26.0-rc.1

Jean-Daniel:
That field is used by other components like k8s external-dns, and should contains all information provided by the envoy service load balancer status.

It also fixes the `Gateway.Status.Addresses` field update when all Ingress entries are remove from the envoy service.

Fixes #5650

### DIFF
--- a/changelogs/unreleased/5651-Jean-Daniel-small.md
+++ b/changelogs/unreleased/5651-Jean-Daniel-small.md
@@ -1,0 +1,1 @@
+Gateway provisioner: Expose all Envoy service IPs and Hostname in `Gateway.Status.Addresses`. Only the first IP was exposed before, even for dual-stack service.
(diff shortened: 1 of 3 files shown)


tsaarni:
Thank you @Jean-Daniel! Would you also add a changelog entry?

tsaarni (approved):

tsaarni:
Thanks for the contribution @Jean-Daniel!

PR #5657, opened 2023-08-07
Setting `disabled = true` on a route should disable the virtual host global rate limit policy

shadialtarsha:
Envoy has[ extensions.filters.http.ratelimit.v3.RateLimitPerRoute](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ratelimit/v3/rate_limit.proto#envoy-v3-api-field-extensions-filters-http-ratelimit-v3-ratelimitperroute-vh-rate-limits) API which allows control over the `Vhost Rate Limits` on the route level. 

This is very useful to disable the vhost rate limits for example on specific routes like `/health` routes.

HTTPProxy example:
```yaml
kind: HTTPProxy
metadata:
  name: echo
spec:
  virtualhost:
    fqdn: local.projectcontour.io
    rateLimitPolicy:
      global:
        disabled: true
      local:
        requests: 100
        unit: hour
        burst: 20
  routes:
    - conditions:
        - prefix: /
      services:
        - name: ingress-conformance-echo
          port: 80
    - conditions:
        - prefix: /foo
      rateLimitPolicy:
        global:
          disabled: true
      services:
        - name: ingress-conformance-echo
          port: 80
```

Fixes https://github.com/projectcontour/contour/issues/5685


changelogs/unreleased/5657-shadialtarsha-minor.md
```diff
@@ -0,0 +1,42 @@
...
+  routes:
+    - conditions:
+        - prefix: /
+      services:
+        - name: ingress-conformance-echo
+          port: 80
+    - conditions:
+        - prefix: /foo
+      rateLimitPolicy:
+        vhRateLimits: "Ignore"
```
davinci26, line 31: the name seems a bit awk to me here for some reason:

* `vhRateLimits`: is `vh` a common thing in the api?
* there is a different object in the CRD called `virtualhost` so this is a bit confusing

I think `routeRateLimitPolicy` (or just rateLimit) would be better imho

davinci26: I think here is you define the `merge` strategy actually, so I guess maybe something along that lines

changelogs/unreleased/5657-shadialtarsha-minor.md
```diff
@@ -0,0 +1,42 @@
...
+apiVersion: projectcontour.io/v1
+kind: HTTPProxy
+metadata:
+  name: echo
+spec:
+  virtualhost:
+    fqdn: local.projectcontour.io
+    rateLimitPolicy:
+      global:
+        disabled: true
```
davinci26, line 17: In hindsight it seems to be that the api here should be the same and a bit recursive.

I think it should have been:

```
    rateLimitPolicy:
        globalPolicy: "Ignore|include|Override"
 ```

Comment (davinci26) [reply changelogs/unreleased/5657-shadialtarsha-minor.md:17]: In general we have 3 layers:

* Global
* HTTProxy
* Route

And each one can be merged with the above policy but the APIs are not the same which makes things a bit confusing imho. We should aim for consistency here

File: internal/envoy/v3/ratelimit.go [hunk 3]
```diff
@@ -163,3 +163,12 @@ func enableXRateLimitHeaders(enable bool) ratelimit_filter_v3.RateLimit_XRateLim
 	}
 	return ratelimit_filter_v3.RateLimit_OFF
 }
+
+// rateLimitPerRoute returns a per-route config to configure vhost rate limits.
+func rateLimitPerRoute(mode int) *anypb.Any {
```
davinci26, line 168: nit we should probably cast to int here and change the function signature to take a `dagRoute.RateLimitPerRoute.VhRateLimits`

internal/dag/policy_test.go
```diff
@@ -994,7 +994,358 @@ func TestRateLimitPolicy(t *testing.T) {
...
+			want: &RateLimitPolicy{
+				Local: &LocalRateLimitPolicy{
+					MaxTokens:          10,
+					TokensPerFill:      10,
+					FillInterval:       time.Minute,
+					ResponseStatusCode: 431,
+				},
+			},
+		},
+		"local - custom response headers to add": {
```
davinci26, line 1070: why do we add this code? I dont get how this test is part of this PR?

clayton-gonsalves (commented):
Question: doesnt this https://projectcontour.io/docs/main/config/api/#projectcontour.io/v1.GlobalRateLimitPolicy
`disabled` solve the use case you are trying to achieve?

changelogs/unreleased/5363-shadialtarsha-minor.md
```diff
@@ -1,28 +0,0 @@
-## Default Global RateLimit Policy
```
skriss, line 1: Please bring this changelog file back :)

changelogs/unreleased/5657-shadialtarsha-minor.md
```diff
@@ -0,0 +1,36 @@
...
```
skriss, line 17: Seems like a better example of this fix would be having a non-disabled virtualhost-level global rate limit policy (rather than having `disabled: true` here), and then disabling it for a specific route like you have below (unless I'm misunderstanding the issue here).

changelogs/unreleased/5657-shadialtarsha-minor.md
```diff
@@ -0,0 +1,36 @@
+## Disable the virtualhost's Global RateLimit policy
```
skriss, line 1:
```suggestion
## Specific routes can now opt out of the virtual host's global rate limit policy
```

changelogs/unreleased/5657-shadialtarsha-minor.md
```diff
@@ -0,0 +1,36 @@
+## Disable the virtualhost's Global RateLimit policy 
+
+Setting `global.disabled` flag to false on a specific route should disable the vhost global rate limit policy.
```
skriss, line 3:
```suggestion
Setting `rateLimitPolicy.global.disabled` flag to true on a specific route now disables the global rate limit policy inherited from the virtual host for that route.
```

skriss (changes requested):
Thanks @shadialtarsha, I think this makes sense, had a few clarifications for the changelog

changelogs/unreleased/5657-shadialtarsha-minor.md
shadialtarsha: Thanks for the suggestion!

internal/dag/dag.go
```diff
@@ -335,6 +335,9 @@ type Route struct {
 	// RateLimitPolicy defines if/how requests for the route are rate limited.
 	RateLimitPolicy *RateLimitPolicy
 
+	// RateLimitPerRoute defines how the route should handle rate limits defined by the virtual host.
+	RateLimitPerRoute *RateLimitPerRoute
```
clayton-gonsalves, line 339: I think we should be able to use reuse  `RateLimitPolicy.disabled`. that will allow us to not introduce new fields.

internal/dag/dag.go
```diff
@@ -571,6 +574,24 @@ type HeaderValueMatchDescriptorEntry struct {
...
 }
 
+type VhRateLimitsType int
+
+const (
+	// VhRateLimitsOverride (Default) will use the virtual host rate limits unless the route has a rate limit policy.
+	VhRateLimitsOverride VhRateLimitsType = iota
+
+	// VhRateLimitsInclude will use the virtual host rate limits even if the route has a rate limit policy.
+	VhRateLimitsInclude
```
clayton-gonsalves, line 584: Do we need `VhRateLimitsInclude` for this case?

test/e2e/httpproxy/global_rate_limiting_test.go
```diff
@@ -413,3 +413,210 @@ func testGlobalRateLimitingRouteTLS(namespace string) {
...
+
+			// Add a global rate limit policy on the virtual host.
+			p.Spec.VirtualHost.RateLimitPolicy = &contourv1.RateLimitPolicy{
+				Global: &contourv1.GlobalRateLimitPolicy{
+					Descriptors: []contourv1.RateLimitDescriptor{
+						{
+							Entries: []contourv1.RateLimitDescriptorEntry{
+								{
+									GenericKey: &contourv1.GenericKeyDescriptor{
+										Value: "vhostlimit",
```
skriss, line 474: Let's use a different value string for this descriptor to avoid any interaction with existing tests (the new value will also need to be set up with the rate limit service itself in `httpproxy_test.go`)

test/e2e/httpproxy/global_rate_limiting_test.go
```diff
@@ -413,3 +413,210 @@ func testGlobalRateLimitingRouteTLS(namespace string) {
...
```
skriss, line 576: Same comment - use a unique descriptor value here to avoid interactions with other tests already using this descriptor value.

test/e2e/httpproxy/default_global_rate_limiting_test.go
```diff
@@ -362,3 +365,84 @@ func testDefaultGlobalRateLimitingVirtualHostTLS(namespace string) {
...
+
+		// Wait until we get a 429 from the proxy confirming
+		// that we've exceeded the rate limit.
+		res, ok := f.HTTP.RequestUntil(&e2e.HTTPRequestOpts{
+			Host:      p.Spec.VirtualHost.Fqdn,
+			Condition: e2e.HasStatusCode(429),
+			Path:      "/echo",
+			RequestOpts: []func(*http.Request){
+				e2e.OptSetHeaders(map[string]string{
+					"X-Default-Header": "test_value_1",
```
skriss, line 411: Same comment, use a unique header value here

test/e2e/httpproxy/default_global_rate_limiting_test.go
```diff
@@ -362,3 +365,84 @@ func testDefaultGlobalRateLimitingVirtualHostTLS(namespace string) {
...
+
+		// We set vh_rate_limits to ignore, which means the route should ignore any rate limit policy
+		// set by the virtual host. Make another request to confirm 200.
+		res, ok = f.HTTP.RequestUntil(&e2e.HTTPRequestOpts{
+			Host:      p.Spec.VirtualHost.Fqdn,
+			Path:      "/echo",
+			Condition: e2e.HasStatusCode(200),
+			RequestOpts: []func(*http.Request){
+				e2e.OptSetHeaders(map[string]string{
+					"X-Default-Header": "test_value_1",
```
skriss, line 441: Will need to match the new unique header value chosen above

internal/dag/dag.go
shadialtarsha: So `dag. RateLimitPolicy` doesn't really have `disabled` on it, this noob is only on the API.

I don't believe introducing this switch to DAG is a good idea as DAG is closer to Envoy than to the HTTPProxy API and also `RateLimitPerRoute` is native to Envoy's struct for doing exactly this.

internal/dag/dag.go
shadialtarsha: I don't have a strong opinion here. 
I kept them just in case someone wants to extend the implementation and supports another mode.

shadialtarsha:
Sorry took me a bit of time to address changes here as we were busy with many other things. 
will follow up immediately with the next round

clayton-gonsalves (approved):

shadialtarsha:
@skriss would appreciate another round once you have time 🙏🏻

shadialtarsha:
Hey folks sorry another ping for this @skriss and @sunjayBhatia 🙏🏻

skriss (changes requested):

skriss:
Otherwise LGTM, thanks @shadialtarsha

skriss (approved):
LGTM, thanks @shadialtarsha

Issue #5127, opened 2023-02-28
HTTPProxy: Configuring/disabling sending Envoy client cert to upstream services


sunjayBhatia:
**Please describe the problem you have**
Contour currently allows configuration of a global client certificate Envoy will present to upstream services: https://projectcontour.io/docs/v1.24.1/configuration/#envoy-client-certificate

This is a global configuration that cannot be overridden for a particular service. Users may want to override or disable sending the client certificate for a particular application in a multi-tenant scenario to prevent changing the global client cert and impacting the rest of the cluster.

Some options here:
- Add to the `UpstreamValidation` struct: https://projectcontour.io/docs/v1.24.1/config/api/#projectcontour.io/v1.UpstreamValidation
  - add options to disable sending client cert, customizing a particular client cert, etc.
  - since this type is used in a few places, have to be sure any changes are applicable in those uses
- rather add a top-level field on the Service?

Relevant code:
- https://github.com/projectcontour/contour/blob/fea8d13050727f2d488573f1f985eeae3ba8bb14/internal/dag/httpproxy_processor.go#L919-L927
- https://github.com/projectcontour/contour/blob/fea8d13050727f2d488573f1f985eeae3ba8bb14/internal/envoy/v3/auth.go#L29-L35

tsaarni:
+1 
I've also heard requests at my organization for an option to set different client certs per target service. If interpreting _very strictly_, client authentication related parameters does not fit into `services.validation` since it is not about validation. For symmetry against downstream, it could have been `services.tls.secretName` but since we already have `services.validation` it probably would look strange too :)

sunjayBhatia:
moving to 1.27.0 as 1.26.0 is coming up soon

PR #5666, opened 2023-08-09
Allow TLSv1.3 between Envoy and upstream service

### LINKED ISSUE #5501: Support of minimumProtocolVersion when Envoy acts as client
Hi Ladies and Gents,

With this feature request I would ask for supporting to configure TLS 1.3 when Envoy acts as client towards upstream server.
As far as we observed, it supports only TLS 1.2 currently, and when upstream forces to use the TLS 1.3, the session initiation ends up in TLS protocol version alert.

Similar functionality is already supported in TLS termination when Envoy acts as server: spec.virtualhost.tls.minimumProtocolVersion.

**- 1. -**
Please note that there is an issue with Envoy when it acts as client and TLS 1.3 is configured by default:
https://github.com/envoyproxy/envoy/issues/9300
The edge case is not detailed about what behavior will be observed in Envoy, just mentioning that can happen when Envoy configured to use client certificate as well.
Therefore it can happen that the combination of these features will not work.

**- 2. -**
Also based on the above ticket, both tls_maximum_protocol_version and tls_minimum_protocol_version proposed to be configured:
```
            '@type': type.googleapis.com/envoy.api.v2.auth.UpstreamTlsContext
            common_tls_context:
              tls_params:
                tls_maximum_protocol_version: TLSv1_3
                tls_minimum_protocol_version: TLSv1_3
```

Many thanks in advance!

BR.
Gergo

tsaarni:
Envoy currently defaults to min/max TLSv1.2 for upstream connections. This change sets maximum to TLSv1.3 to allow both TLSv1.2 and TLSv1.3 between Envoy and the upstream service. The upstream service is responsible for picking one of the protocol versions.

The suggested change does not make the upstream min/max TLS version configurable via Contour. If we need min/max as configurable, then we can consider adding new field(s) either inside `httpproxy.spec.routes.services.validation`, which is existing struct with TLS related parameters, but the name implies it is only for validation parameters, or we could introduce new structure `httpproxy.spec.routes.services.tls`, and include other TLS parameters as well, such as the ones discussed in https://github.com/projectcontour/contour/issues/5127.

There are some implications from the suggested change that the users will be able to observe:  

When an upstream service using TLSv1.3 does not accept the client certificate that Envoy presents, the TLS alert message about authentication failure is not propagated to downstream client. See https://github.com/envoyproxy/envoy/issues/9300 for further information. 

Secondly, Envoy will not retry connection attempt after authentication failure, when TLSv1.3 is used (not 100% sure if it does for TLSv1.2 but it is implied in the Envoy issue). This should be OK since upstream service consists of replicas of single deployment and all pods are expected to share their configuration, including the authentication parameters. Therefore, authentication would fail towards another replica as well.

Fixes #5501

### DIFF
--- a/changelogs/unreleased/5666-tsaarni-small.md
+++ b/changelogs/unreleased/5666-tsaarni-small.md
@@ -1,0 +1,1 @@
+Allow TLSv1.3 between Envoy and upstream services.
--- a/internal/envoy/v3/auth.go
+++ b/internal/envoy/v3/auth.go
@@ -38,6 +38,12 @@ func UpstreamTLSContext(peerValidationContext *dag.PeerValidationContext, sni st
 		CommonTlsContext: &envoy_v3_tls.CommonTlsContext{
 			AlpnProtocols:                  alpnProtocols,
 			TlsCertificateSdsSecretConfigs: clientSecretConfigs,
+			TlsParams: &envoy_v3_tls.TlsParameters{
+				// Envoy defaults to TLSv1.2 for both minimum and maximum protocol version, when acting as a client.
+				// Override the maximum to allow upstream server to use TLSv1.3.
+				// See https://github.com/envoyproxy/envoy/issues/9300 for further information.
+				TlsMaximumProtocolVersion: envoy_v3_tls.TlsParameters_TLSv1_3,
+			},
 		},
 		Sni: sni,
 	}
--- a/internal/envoy/v3/auth_test.go
+++ b/internal/envoy/v3/auth_test.go
@@ -45,7 +45,11 @@ func TestUpstreamTLSContext(t *testing.T) {
 	}{
 		"no alpn, no validation": {
 			want: &envoy_v3_tls.UpstreamTlsContext{
-				CommonTlsContext: &envoy_v3_tls.CommonTlsContext{},
+				CommonTlsContext: &envoy_v3_tls.CommonTlsContext{
+					TlsParams: &envoy_v3_tls.TlsParameters{
+						TlsMaximumProtocolVersion: envoy_v3_tls.TlsParameters_TLSv1_3,
+					},
+				},
 			},
 		},
 		"h2, no validation": {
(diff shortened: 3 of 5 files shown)


tsaarni:
> Secondly, Envoy will not fail after authentication failure in case of TLSv1.3. It should be OK since upstream service consists of replicas of single deployment - all pods are expected to share their configuration, including the authentication parameters. Therefore, the lack of retrying request towards another replica instance should not matter.

Can this be problem for externalName services? Do we end up loadbalancing between the IP addresses that external address resolves to? The external upstream servers might have different authentication configuration.

sunjayBhatia:
> > Secondly, Envoy will not retry connection attempt after authentication failure, when TLSv1.3 is used.
> 
> Can this be problem for externalName services? Do we end up loadbalancing between the IP addresses that external DNS address resolves to? The external upstream servers might potentially have different authentication configuration and therefore retry might be succesful.

We do end up configuring Envoy to resolve DNS itself when we have an ExternalName service:
- https://github.com/projectcontour/contour/blob/d21ea00f15d4672484cc7f5dd683f7a078d7d71d/internal/envoy/v3/cluster.go#L64-L73
- https://github.com/projectcontour/contour/blob/d21ea00f15d4672484cc7f5dd683f7a078d7d71d/internal/envoy/v3/endpoint.go#L78-L86

I believe retries happen once Envoy's router has already selected an endpoint: https://www.envoyproxy.io/docs/envoy/latest/intro/life_of_a_request#overview

sunjayBhatia (commented):
the majority of users/use cases should be fine with this default changing, but given the limitations/to avoid issues/support requests it might be good to validate what retry modes don't work as expected, esp. since the majority of what is available in [RetryPolicy](https://github.com/projectcontour/contour/blob/135c8a96514a598fd54d5045c1ed4bb3ad279535/apis/projectcontour/v1/httpproxy.go#L1121-L1162) is at the HTTP level

based on that, we may consider changing this to be an opt-in rather than always set

tsaarni:
Here are the results:

The corner case where Envoy will have problem with TLSv1.3 handshake is when upstream does not accept client certificate and retry policy for HTTP connection manager is configured with `connection-failure`. Connection attempt towards other cluster endpoints (when they exists) is not triggered in case of TLSv1.3 while it is with TLSv1.2.

The case where TCP proxy _would_  have the same problem is when `TcpProxy.max_connect_attempts` is greater than 1.  We currently use default value (1) and there is no retry, regardless of the protocol version.

PR #5668, opened 2023-08-09
fix [Gateway API]set Listener ResolvedRefs=true if not set to false

### LINKED ISSUE #5648: Gateway API: set Listener ResolvedRefs=true if not set to false
ref. https://github.com/kubernetes-sigs/gateway-api/pull/2247

We need to set the Listener's ResolvedRefs=true if we haven't otherwise set it to false.

Should be similar to what we do for Accepted/Programmed -- see https://github.com/projectcontour/contour/blob/d60959ec98d46df61de6127464677fa571fd2dae/internal/dag/gatewayapi_processor.go#L435

HeavenTonight:
fixed #5648
fix [Gateway API]set Listener ResolvedRefs=true if not set to false

### DIFF
--- a/internal/dag/gatewayapi_processor.go
+++ b/internal/dag/gatewayapi_processor.go
@@ -603,7 +603,7 @@ func (p *GatewayAPIProcessor) getListenerRouteKinds(listener gatewayapi_v1beta1.
 			gwAccessor.AddListenerCondition(
 				string(listener.Name),
 				gatewayapi_v1beta1.ListenerConditionResolvedRefs,
-				metav1.ConditionFalse,
+				metav1.ConditionTrue,
 				gatewayapi_v1beta1.ListenerReasonInvalidRouteKinds,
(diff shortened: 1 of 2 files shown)


sunjayBhatia (changes requested):
This is not quite correct, we do not want to set the ResolvedRefs condition to `true` in all cases, rather we want to set it to `true` when other validation has not already added the condition to ensure it is always set (similar to the logic here for other conditions: https://github.com/projectcontour/contour/blob/d60959ec98d46df61de6127464677fa571fd2dae/internal/dag/gatewayapi_processor.go#L435-L484)

I think we may want to hold off on implementing this until the conversation on https://github.com/kubernetes-sigs/gateway-api/pull/2247 is resolved and conformance test changes are merged, the conversation there may cause us to set a condition in a different place (possibly at the top level of the Gateway)

skriss:
@HeavenTonight did you want to work on revising this per above comments? The upstream conformance has now merged so this should be testable.

PR #5690, opened 2023-08-19
Support %RESP(Header-Name)% dynamic response headers

### LINKED ISSUE #5689: Support %RESP(Header-Name)% dynamic response headers
This feature is very useful for me 
[#3234 ](https://github.com/projectcontour/contour/issues/3234)

However i don't know why %RESP(Header-Name)% is not in part of the list as it is the same than %REQ(Header-Name).
Here the doc:
[Access Log Usage](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage
)

For me it is useful to add this dynamic value as it was done for REQ
It will be a quick win.

```
var validRespEnvoyVar = regexp.MustCompile(`%(%RESP\([\w-]+\)%)%`)
escapedValue = validRespEnvoyVar.ReplaceAllString(escapedValue, "$1")
```

Here the line where it is done:
[REQ line](https://github.com/projectcontour/contour/commit/911f1bf49ba211cad56f3a86bd80413c9fdd9ce1#diff-558cac45415d3c8c6037c1226818eaa38b5342897b7a6448425d34e7b2f7ab61R162)

hmouaad:
Fixes #5689 

Quick win to add new dynamic header value
Please feel free to add any comment on this pull request

### DIFF
--- a/internal/dag/policy.go
+++ b/internal/dag/policy.go
@@ -315,6 +315,9 @@ func escapeHeaderValue(value string, dynamicHeaders map[string]string) string {
 	// REQ(header-name)
 	var validReqEnvoyVar = regexp.MustCompile(`%(%REQ\(:?[\w-]+(\?:?[\w-]+)?\)(:\d+)?%)%`)
 	escapedValue = validReqEnvoyVar.ReplaceAllString(escapedValue, "$1")
+	// RESP(header-name) similar to REQ
+	var validRespEnvoyVar = regexp.MustCompile(`%(%RESP\(:?[\w-]+(\?:?[\w-]+)?\)(:\d+)?%)%`)
+	escapedValue = validRespEnvoyVar.ReplaceAllString(escapedValue, "$1")
 	return escapedValue
 }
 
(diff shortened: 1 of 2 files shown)


hmouaad:
Someone can reveiw this pull request ?

internal/dag/policy.go
```diff
@@ -315,6 +315,9 @@ func escapeHeaderValue(value string, dynamicHeaders map[string]string) string {
 	// REQ(header-name)
 	var validReqEnvoyVar = regexp.MustCompile(`%(%REQ\(:?[\w-]+(\?:?[\w-]+)?\)(:\d+)?%)%`)
 	escapedValue = validReqEnvoyVar.ReplaceAllString(escapedValue, "$1")
+	// RESP(header-name) similar to REQ
+	var validRespEnvoyVar = regexp.MustCompile(`%(%RESP\(:?[\w-]+(\?:?[\w-]+)?\)(:\d+)?%)%`)
```
skriss, line 319: @hmouaad I would think we'd only want to allow this in a ResponseHeadersPolicy? I'm not sure off the top of my head what Envoy's behavior would be if this was specified in `request_headers_to_add`, but seems like good validation for Contour to just not allow it. That will require some further refactoring here.

PR #5700, opened 2023-08-22
update to Kubernetes 1.28

skriss:


skriss:
If CI looks good I'd say we may as well get this in ahead of the Contour 1.26 release

sunjayBhatia (approved):

PR #5703, opened 2023-08-28
Design: Endpoint slice support

clayton-gonsalves:
Design proposal to address https://github.com/projectcontour/contour/issues/2696

### DIFF
--- a/design/endpoint-slice-support.md
+++ b/design/endpoint-slice-support.md
@@ -1,0 +1,101 @@
+# Contour k8s endpoint slice support
+
+Contour currently supports Kubernetes endpoints.
+This document proposes a process to enable support for k8s endpoint slices and, eventually, migrate to exclusively using it.
+
+## Goals
+
+- Contour will support k8s endpoint slices.
+- The endpoint slices will be gated by a feature flag in the Contour config.
+- Kubernetes endpoints will still be the default mode of operation.
+
+## Non-goals
+
+- Removal of k8s endpoints.
+- Support advanced features such as topology aware routing/hints offered by k8s endpoints.
+
+## Background
+
+See [#2696](https://github.com/projectcontour/contour/issues/2696)
+
+Before the introduction of the endpoint slices, the only method for a service to expose the IP addresses and ports of the pads it fronted was through Kubernetes endpoints.
+
+However, Kubernetes endpoints have a few challenges with scaling, such as
+- transmission of the entire endpoint object even if one endpoint changes, leading to network contention and heavy recalculation on the consumers of this data.
+- As of k8s v1.22, a hard limit of 1000 addresses per endpoint object.
+- No topology aware information for more intelligent routing.
+
+To solve the above problems, endpoint slices was introduced in 1.16 and, as of k8s v1.21 in [stable state](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/).
+
+
+## High-Level Design
+
+A feature flag will gate the usage of endpoint slices, and the implementation will use the same pattern as used in the `endpointtranslator.go` with a few differences listed in the detailed design.
+
+## Detailed Design
+
+### Contour Configuration changes
+Endpoint slices will be configured by a boolean feature flag, which defaults to false and allows users to opt to use endpoint slices.
+
+```yaml
+...
+server:
+    xds-server-type: contour
+use-endpoint-slice: true
+...
+```
+
+```go
+type Parameters struct {
+    // useEndpointSlice configures contour to fetch endpoint data
+    // from k8s endpoint slices. defaults to false and reading endpoint
+    // data from the k8s endpoints.
+    useEndpointSlice bool
+}
+```
+### Implementation
+
+#### Context
+The implementation of k8s endpoint slices is fundamentally different from how endpoint slices work.
+
+Below is an example of a Kubernetes service represented in as endpoint and endpoint slice objects
+
+endpoints
+```
+NAME                       ENDPOINTS                                                          AGE
+ingress-conformance-echo   10.244.0.10:3000,10.244.0.153:3000,10.244.0.154:3000 + 7 more...   3d
+```
+
+endpoint slice
+```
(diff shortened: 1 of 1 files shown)


design/endpoint-slice-support.md
```diff
@@ -0,0 +1,101 @@
+# Contour k8s endpoint slice support
+
+Contour currently supports Kubernetes endpoints.
+This document proposes a process to enable support for k8s endpoint slices and, eventually, migrate to exclusively using it.
+
+## Goals
+
+- Contour will support k8s endpoint slices.
+- The endpoint slices will be gated by a feature flag in the Contour config.
+- Kubernetes endpoints will still be the default mode of operation.
```
skriss, line 10: IMO, once we're confident in the EndpointSlices implementation, we should move to using it as the default, and eventually remove the Endpoint option/code altogether. Let's call that out here as well (assuming folks agree) so it's clear what the plan is.

design/endpoint-slice-support.md
```diff
@@ -0,0 +1,101 @@
...
+- No topology aware information for more intelligent routing.
+
+To solve the above problems, endpoint slices was introduced in 1.16 and, as of k8s v1.21 in [stable state](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/).
+
+
+## High-Level Design
+
+A feature flag will gate the usage of endpoint slices, and the implementation will use the same pattern as used in the `endpointtranslator.go` with a few differences listed in the detailed design.
+
+## Detailed Design
```
skriss, line 35: Should/will any of the [EndpointSlice conditions](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#conditions) be taken into account?

design/endpoint-slice-support.md
```diff
@@ -0,0 +1,101 @@
...
+    endpointSlice: map[types.NamespacedName]map[string]*discoveryv1.EndpointSlice{},
+}
+```
+
+The above structure lets us keep track of each endpoint slice and change/update/delete an endpoint without shuffling all the existing endpoints.
+
+## Tests
+
+To ensure the feature is well tested, this feature will have
+- e2e tests
```
skriss, line 96: I'd suggest adding a run to our [daily build](https://github.com/projectcontour/contour/blob/main/.github/workflows/build_daily.yaml) that configures Contour to use EndpointSlices and then runs the full E2E suite. We've done this for other major config variations (e.g. running Envoy as a Deployment) that we want to test regularly/with full coverage but that don't necessarily need to go into the per-PR suite.

design/endpoint-slice-support.md
skriss, line 35: Could you add some info about how [FQDN address types](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#address-types) are handled? Off the top of my head, I think we'd want to handle them similarly to ExternalName services (i.e. disabled by default since there are some security concerns), but would need to look at it a little more to be sure.

skriss (changes requested):
Thanks for the design doc @clayton-gonsalves, apologies for the delay in reviewing. Added a few comments/questions.

design/endpoint-slice-support.md
clayton-gonsalves: added a small migration section in https://github.com/projectcontour/contour/pull/5703/commits/63fac731b8b745fbe893ca69e3739cd3b21641f5

design/endpoint-slice-support.md
clayton-gonsalves: I've chosen to stick with just the `ready` condition to keep things simpler for initial implementation and have feature parity with endpoints. Added a section regarding that in https://github.com/projectcontour/contour/pull/5703/commits/63fac731b8b745fbe893ca69e3739cd3b21641f5

design/endpoint-slice-support.md
clayton-gonsalves: That is a great idea. added that to the test section in https://github.com/projectcontour/contour/pull/5703/commits/63fac731b8b745fbe893ca69e3739cd3b21641f5

design/endpoint-slice-support.md
clayton-gonsalves: So, I looked into this and am leaning towards not supporting it. 

To use the endpointSlice FQDN address type, a user needs to create the EndpointSlice object manually and attach it to a service via labels, something like this 

```
apiVersion: v1
kind: Service
metadata:
  name: fqdn-test
spec:
  ports:
    - name: http
      port: 80
      targetPort: 80
---
apiVersion: discovery.k8s.io/v1
kind: EndpointSlice
metadata:
  name:  fqdn-test
  labels:
    kubernetes.io/service-name: fqdn-test
ports:
  - name: http
    appProtocol: http
    protocol: TCP
    port: 80
endpoints:
  - addresses:
      - "projectocontour.io"
```

While the [PR](https://github.com/projectcontour/contour/pull/5745) I have open should work with this out of the box, 

I have the below reasoning to disable it explicitly. 
- I think there is possibly a small set use-cases that this solves that externalName services don't solve. 
- Limit the scope of features to avoid the introduction of confounding variables. 

wdyt?

skriss (approved):
LGTM, thanks @clayton-gonsalves

tsaarni (commented):
Since the processing of `EndpointSlices` and `Endpoints` would be mutually exclusive maybe it could be mentioned that `Endpoints` where user has [disabled mirroring](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#endpointslice-mirroring) will not be processed by Contour.

clayton-gonsalves:
> Since the processing of `EndpointSlices` and `Endpoints` would be mutually exclusive maybe it could be mentioned that `Endpoints` where user has [disabled mirroring](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#endpointslice-mirroring) will not be processed by Contour.

addressed in https://github.com/projectcontour/contour/pull/5703/commits/45a8a96d2e9eaed5c05a9fc0e93106b5d57b17e4

tsaarni (approved):
Looks good to me!

PR #5722, opened 2023-08-28
dependabot: group core k8s dependency updates

skriss:
ref. https://github.blog/changelog/2023-06-30-grouped-version-updates-for-dependabot-public-beta/ and https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

This _should_ result in just one PR total for the core k8s dependency updates.


sunjayBhatia (approved):
+1, in practice these updates (esp. for client-go which triggers others) should all come together

PR #5735, opened 2023-09-01
remove Gateway API webhook

### LINKED ISSUE #5732: remove Gateway API webhook from example manifests

Now that Gateway API uses CEL, and Contour's supported Kubernetes versions are all 1.25+, we should be able to drop the Gateway API webhook from our example installs.

ref. https://kubernetes.io/blog/2023/08/29/gateway-api-v0-8/#cel-validation

skriss:
Drops the Gateway API webhook from example
manifests and testing since validations are
now implemented in Common Expression Language
(CEL).

Closes #5732.


sunjayBhatia (commented):
nice, looks like the examples just have CRDs now 👍🏽 

one more place that could be simplified? https://github.com/projectcontour/contour/blob/6a1ee75be5e1c7bbb11c1790ed55123946d1587e/Makefile#L205-L210

PR #5745, opened 2023-09-05
Add support for endpoint slices

clayton-gonsalves:
This PR adds support for Kubernetes endpoint slices. 

Fixes https://github.com/projectcontour/contour/issues/2696


internal/xdscache/v3/endpointslicetranslator.go
```diff
@@ -0,0 +1,458 @@
...
+	discoveryv1 "k8s.io/api/discovery/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/cache"
+)
+
+// RecalculateEndpointsSlice generates a slice of LoadBalancingEndpoint
+// resources by matching the given service port to the given discoveryv1.EndpointSlice.
+// endpointSliceMap may be nil, in which case, the result is also nil.
+func RecalculateEndpointsSlice(port, healthPort v1.ServicePort, endpointSliceMap map[string]*discoveryv1.EndpointSlice) []*LoadBalancingEndpoint {
+	var lb []*LoadBalancingEndpoint
```
davinci26, line 41: Do we want a prom stat counter here? I am thinking that it might be nice for debugging to be able to see the rate of recalculations so we can correlated with issues

davinci26 (commented):
lgtm, one question and one test request

test/e2e/infra/endpointslice_test.go
```diff
@@ -0,0 +1,126 @@
...
+								Name: "echo",
+								Port: 80,
+							},
+						},
+					},
+				},
+			},
+		}
+
+		f.CreateHTTPProxyAndWaitFor(p, e2e.HTTPProxyValid)
```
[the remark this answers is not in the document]
clayton-gonsalves: It would be hard to achieve a case where we can deterministically say during a test if the endpoints are ready or not. sure, w could remove the sleep function to catch some unready pods but that might lead to flakey tests

internal/xdscache/v3/endpointslicetranslator.go
```diff
@@ -0,0 +1,458 @@
...
+	uniqueEndpoints := make(map[string]struct{}, 0)
+	var healthCheckPort int32
+
+	for _, enpointSlice := range endpointSliceMap {
+		sort.Slice(enpointSlice.Endpoints, func(i, j int) bool {
```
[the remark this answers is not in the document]
clayton-gonsalves: This is from a carry over from the endpoint slice implementation. I think we can get rid of this.

internal/xdscache/v3/endpointslicetranslator.go
clayton-gonsalves: rechecking the code, there metrics should be there by default.

clayton-gonsalves:
@skriss @sunjayBhatia  this is ready for the first pass

internal/xdscache/v3/endpointslicetranslator.go
```diff
@@ -0,0 +1,458 @@
...
+func (c *EndpointSliceCache) UpdateEndpointSlice(endpointSlice *discoveryv1.EndpointSlice) bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	name := types.NamespacedName{Namespace: endpointSlice.Namespace, Name: endpointSlice.Labels[discoveryv1.LabelServiceName]}
+
+	if c.endpointSlices[name] == nil {
+		c.endpointSlices[name] = make(map[string]*discoveryv1.EndpointSlice)
+	}
+	c.endpointSlices[name][endpointSlice.DeepCopy().Name] = endpointSlice.DeepCopy()
```
skriss, line 227: Let's just do a single DeepCopy() here, for the value.

internal/xdscache/v3/endpointslicetranslator.go
```diff
@@ -0,0 +1,458 @@
...
+	// needs to be recalculated. Clusters can be added to the stale
+	// slice due to changes in Endpoints or due to a DAG rebuild.
+	stale []*dag.ServiceCluster
+
+	// Index of ServiceClusters. ServiceClusters are indexed
+	// by the name of their Kubernetes Services. This makes it
+	// easy to determine which Endpoints affect which ServiceCluster.
+	services map[types.NamespacedName][]*dag.ServiceCluster
+
+	// Cache of endpointsSlices, indexed by Namespaced name.
```
skriss, line 119: The key is the NamespacedName of the _Service_, not the EndpointSlice -- let's note that here.

internal/xdscache/v3/endpointslicetranslator.go
```diff
@@ -0,0 +1,458 @@
...
+	v1 "k8s.io/api/core/v1"
+	discoveryv1 "k8s.io/api/discovery/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/cache"
+)
+
+// RecalculateEndpointsSlice generates a slice of LoadBalancingEndpoint
+// resources by matching the given service port to the given discoveryv1.EndpointSlice.
+// endpointSliceMap may be nil, in which case, the result is also nil.
+func RecalculateEndpointsSlice(port, healthPort v1.ServicePort, endpointSliceMap map[string]*discoveryv1.EndpointSlice) []*LoadBalancingEndpoint {
```
skriss, line 40: I'd say make this a method on EndpointSliceCache so we can name it just `RecalculateEndpoints` and it doesn't conflict with the existing one, since what it's recalculating is just (LoadBalancing) Endpoints.

internal/xdscache/v3/endpointslicetranslator.go
```diff
@@ -0,0 +1,458 @@
...
+	var lb []*LoadBalancingEndpoint
+	uniqueEndpoints := make(map[string]struct{}, 0)
+	var healthCheckPort int32
+
+	for _, enpointSlice := range endpointSliceMap {
```
skriss, line 45:
```suggestion
	for _, endpointSlice := range endpointSliceMap {
```

internal/xdscache/v3/endpointslicetranslator.go
```diff
@@ -0,0 +1,458 @@
...
+		}
+	}
+
+	c.stale = clusters
+	c.services = serviceIndex
+
+	return nil
+}
+
+// UpdateEndpoint adds endpointSlice to the cache, or replaces it if it is
```
skriss, line 214:
```suggestion
// UpdateEndpointSlice adds endpointSlice to the cache, or replaces it if it is
```

internal/xdscache/v3/endpointslicetranslator.go
```diff
@@ -0,0 +1,458 @@
...
+	// all as stale.
+	if affected := c.services[name]; len(affected) > 0 {
+		c.stale = append(c.stale, affected...)
+		return true
+	}
+
+	return false
+}
+
+// DeleteEndpoint deletes endpointSlice from the cache. Any ServiceClusters
```
skriss, line 239:
```suggestion
// DeleteEndpointSlice deletes endpointSlice from the cache. Any ServiceClusters
```

internal/xdscache/v3/endpointslicetranslator.go
```diff
@@ -0,0 +1,458 @@
...
+// that are backed by a Service that endpointSlice belongs to, become stale. Returns
+// a boolean indicating whether any ServiceClusters use endpointSlice or not.
+func (c *EndpointSliceCache) DeleteEndpoint(endpointSlice *discoveryv1.EndpointSlice) bool {
```
skriss, line 242:
```suggestion
func (c *EndpointSliceCache) DeleteEndpointSlice(endpointSlice *discoveryv1.EndpointSlice) bool {
```

skriss (changes requested):
Thanks @clayton-gonsalves, this is looking pretty good - first pass just getting a few nits out of the way. Still looking into a few of the details

changelogs/unreleased/5745-clayton-gonsalves-minor.md
```diff
@@ -0,0 +1 @@
+Add Kubernetes Endpoint Slice support.
```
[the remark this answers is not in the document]
skriss:
```suggestion
## Add Kubernetes Endpoint Slice support
```

internal/xdscache/v3/endpointslicetranslator.go
```diff
@@ -0,0 +1,458 @@
...
+
+			// Range over each port. We want the resultant endpoints to be a
+			// a cartesian product MxN where M are the endpoints and N are the ports.
+			for _, p := range enpointSlice.Ports {
+				// Nil check for the port.
+				if p.Port == nil {
+					continue
+				}
+
+				if p.Protocol == nil || (healthPort.Protocol != *p.Protocol || port.Protocol != *p.Protocol) && *p.Protocol != v1.ProtocolTCP {
```
skriss, line 64: Still digesting the logic in this line, I feel like the `&&` should be an `||` (I know this matches what was in EndpointsTranslator though). We can address separately if it should be changed, since it would need to change in both if so.

skriss (changes requested):

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -84,6 +84,11 @@ type ContourConfigurationSpec struct {
 
 	// Tracing defines properties for exporting trace data to OpenTelemetry.
 	Tracing *TracingConfig `json:"tracing,omitempty"`
+
+	// useEndpointSlice configures contour to fetch endpoint data
+	// from k8s endpoint slices. defaults to false and reading endpoint
+	// data from the k8s endpoints.
+	UseEndpointSlice bool `json:"useEndpointSlice,omitempty"`
```
skriss, line 91: I'd be interested to hear what the other maintainers think about adding a more generic `FeatureFlags` field instead.  It could be used for other things, and deprecating a value for it should be simpler than deprecating an entire field.

clayton-gonsalves: I think thats a good idea, rather than polluting the config every time we want to add/deprecate a field.

internal/xdscache/v3/endpointslicetranslator.go
clayton-gonsalves: yeah looking at it again, it feels like it should be an `||` because we only want to support TCP, we should short-circuit it if the protocol is anything else .

apis/projectcontour/v1alpha1/contourconfig.go
clayton-gonsalves: i made the change in https://github.com/projectcontour/contour/pull/5745/commits/1fd6eea12689e2de160d46537422f595af150d9b but will also wait for feedback from the other maintainers.

cmd/contour/serve.go
```diff
@@ -588,12 +599,21 @@ func (s *Server) doServe() error {
...
-		Counter: contourMetrics.EventHandlerOperations,
-	}, s.mgr.GetCache()); err != nil {
-		s.log.WithError(err).WithField("resource", "endpoints").Fatal("failed to create informer")
+	// Inform on endpoints/endpointSlices.
+	if contourConfiguration.UseEndpointSlice {
+		if err := informOnResource(&discoveryv1.EndpointSlice{}, &contour.EventRecorder{
+			Next:    endpointHandler,
+			Counter: contourMetrics.EventHandlerOperations,
+		}, s.mgr.GetCache()); err != nil {
+			s.log.WithError(err).WithField("resource", "endpoints").Fatal("failed to create informer")
```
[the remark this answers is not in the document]
clayton-gonsalves: fixed in https://github.com/projectcontour/contour/pull/5745/commits/1fd6eea12689e2de160d46537422f595af150d9b

cmd/contour/serve.go
```diff
@@ -588,12 +599,21 @@ func (s *Server) doServe() error {
...
 
-	// Inform on endpoints.
-	if err := informOnResource(&corev1.Endpoints{}, &contour.EventRecorder{
-		Next:    endpointHandler,
-		Counter: contourMetrics.EventHandlerOperations,
-	}, s.mgr.GetCache()); err != nil {
-		s.log.WithError(err).WithField("resource", "endpoints").Fatal("failed to create informer")
+	// Inform on endpoints/endpointSlices.
+	if contourConfiguration.UseEndpointSlice {
+		if err := informOnResource(&discoveryv1.EndpointSlice{}, &contour.EventRecorder{
```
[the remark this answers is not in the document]
clayton-gonsalves: added in https://github.com/projectcontour/contour/pull/5745/commits/1fd6eea12689e2de160d46537422f595af150d9b

internal/xdscache/v3/endpointslicetranslator.go
```diff
@@ -0,0 +1,458 @@
...
+	"github.com/projectcontour/contour/internal/sorter"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/protobuf/proto"
+	v1 "k8s.io/api/core/v1"
+	discoveryv1 "k8s.io/api/discovery/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/cache"
+)
+
+// RecalculateEndpointsSlice generates a slice of LoadBalancingEndpoint
```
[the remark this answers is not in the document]
clayton-gonsalves: fixed in https://github.com/projectcontour/contour/pull/5745/commits/1fd6eea12689e2de160d46537422f595af150d9b

internal/xdscache/v3/endpointslicetranslator.go
clayton-gonsalves: addressed in https://github.com/projectcontour/contour/pull/5745/commits/5f090766d8b3f26fbbb09740a202f797949c04df

internal/xdscache/v3/endpointslicetranslator.go
clayton-gonsalves: I havent changed this yet, but happy to change this here and create an issue for `endpoints`

apis/projectcontour/v1alpha1/contourconfig.go
skriss: I was actually thinking a string slice, rather than a concrete field per feature, so that as features come and go it doesn't require any changes to the type itself, just to the values that are accepted in the field. But, would like to get a second opinion before making this change.

changelogs/unreleased/5745-clayton-gonsalves-minor.md
```diff
@@ -0,0 +1,4 @@
+## Add Kubernetes Endpoint Slice support
+
+This change optionally enables Contour to consume the kubernetes endpointslice API to determine the endpoints to configure Envoy with.
+Note: This change is off by default and is gated by the feature flag `useEndpointSlice`.
```
skriss, line 4: Let's add a sentence about this becoming the default down the road:

```suggestion
Note: This change is off by default and is gated by the feature flag `useEndpointSlice`.
This feature will be enabled by default in a future version on Contour once it has had sufficient bake time in production environments.
```

skriss (changes requested):

internal/xdscache/v3/endpointslicetranslator.go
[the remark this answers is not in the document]
skriss: I think it's desirable to sort, to produce stable results across runs. I forget the exact details but impacts performance and/or load balancing algorithms. So I'd say let's leave this in place.

skriss:
Modulo the feature flags spec, this is looking good to me, no further comments at the moment

apis/projectcontour/v1alpha1/contourconfig.go
sunjayBhatia: +1 for string slice, similar in spirit to the `--disable-feature` flag which is more generic than adding a specific flag per thing we want to disable

I guess that naming makes it sound more generic than it is, it's really for disabling reconciling some optional extended support CRDs, i dunno if we should repurpose the existing flag, have to think about it a bit

skriss: Yeah, I was looking at that flag as well for possible use, but since its semantics are opt-out rather than opt-in, it seemed hard to use here (would you opt out of using Endpoints which would result in opting into EndpointSlices? Seems unintuitive). I also don't quite remember why we went with a flag vs. a config file/ContourConfig field for that, will have to look back. 

If we do just go ahead with a config file/ContourConfig field for opt-in features, then down the road we could always look at deprecating the `--disable-feature` flag and replacing with something more aligned with the opt-in feature flags.

changelogs/unreleased/5745-clayton-gonsalves-minor.md
clayton-gonsalves: addressed in https://github.com/projectcontour/contour/pull/5745/commits/ee4aa23bff2d152d469d36cf1d6ccf7dd8ce4777

cmd/contour/serve.go
```diff
@@ -588,12 +599,21 @@ func (s *Server) doServe() error {
...
+			Counter: contourMetrics.EventHandlerOperations,
+		}, s.mgr.GetCache()); err != nil {
+			s.log.WithError(err).WithField("resource", "endpointslices").Fatal("failed to create informer")
+		}
+	} else {
+		if err := informOnResource(&corev1.Endpoints{}, &contour.EventRecorder{
+			Next:    endpointHandler,
+			Counter: contourMetrics.EventHandlerOperations,
+		}, s.mgr.GetCache()); err != nil {
+			s.log.WithError(err).WithField("resource", "endpointslices").Fatal("failed to create informer")
```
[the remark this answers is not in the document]
clayton-gonsalves: addressed in https://github.com/projectcontour/contour/pull/5745/commits/ee4aa23bff2d152d469d36cf1d6ccf7dd8ce4777

apis/projectcontour/v1alpha1/contourconfig.go
clayton-gonsalves: added a string slice in https://github.com/projectcontour/contour/pull/5745/commits/ee4aa23bff2d152d469d36cf1d6ccf7dd8ce4777

is this what y'all had in mind?

internal/xdscache/v3/endpointslicetranslator.go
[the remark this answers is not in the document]
clayton-gonsalves: just want to call out, that here the sorting happens only per `slice`, since we store all the slices in a map it's not possible to sort the combination of all the endpoints.

changelogs/unreleased/5745-clayton-gonsalves-minor.md
```diff
@@ -0,0 +1,6 @@
...
```
sunjayBhatia, line 4: nit
```suggestion
Note: This change is off by default and is gated by the feature flag `useEndpointSlices`.
```

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -84,8 +84,19 @@ type ContourConfigurationSpec struct {
 
 	// Tracing defines properties for exporting trace data to OpenTelemetry.
 	Tracing *TracingConfig `json:"tracing,omitempty"`
+
+	// FeatureFlags defines toggle to enable new contour features.
+	// available toggles are
```
sunjayBhatia, line 89: nit:
```suggestion
	// Available toggles are:
	// useEndpointSlices - Configures contour to fetch endpoint data
	// from k8s EndpointSlices instead of the default behavior which reads
	// endpoint data from Endpoint resources.
```

skriss:
@clayton-gonsalves unfortunately looks like there are a couple small merge conflicts to resolve

apis/projectcontour/v1alpha1/contourconfig.go
skriss: Looks reasonable to me!

skriss (commented):
No further comments from me, ready for this to go once the merge conflicts and latest comments are resolved

tsaarni (changes requested):
Could you add the `featureFlags` also in Contour config file documentation here [`site/content/docs/main/configuration.md`](site/content/docs/main/configuration.md)? It often tends to get out-of-sync and the naming for attributes is not very consistent (dash vs camelcase). 

There is also example config in the same doc page, and same example config is in deployment manifest [`examples/contour/01-contour-config.yaml`](examples/contour/01-contour-config.yaml).

clayton-gonsalves:
> Could you add the `featureFlags` also in Contour config file documentation here [`site/content/docs/main/configuration.md`](site/content/docs/main/configuration.md)? It often tends to get out-of-sync and the naming for attributes is not very consistent (dash vs camelcase).
> 
> There is also example config in the same doc page, and same example config is in deployment manifest [`examples/contour/01-contour-config.yaml`](examples/contour/01-contour-config.yaml).

I added it to the `configuration.md, I would prefer keeping the example configs as is because this feature is still opt in. I don't want folks copy pasting the config and enabling something by default.

clayton-gonsalves:
@skriss @sunjayBhatia I think all comments are resolved. is this good to go?

tsaarni (commented):
Just couple more questions from me, nothing important though!

internal/xdscache/v3/endpointslicetranslator.go
```diff
@@ -0,0 +1,458 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/cache"
+)
+
+// RecalculateEndpoints generates a slice of LoadBalancingEndpoint
+// resources by matching the given service port to the given discoveryv1.EndpointSlice.
+// endpointSliceMap may be nil, in which case, the result is also nil.
+func (c *EndpointSliceCache) RecalculateEndpoints(port, healthPort v1.ServicePort, endpointSliceMap map[string]*discoveryv1.EndpointSlice) []*LoadBalancingEndpoint {
+	var lb []*LoadBalancingEndpoint
+	uniqueEndpoints := make(map[string]struct{}, 0)
```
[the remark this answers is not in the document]
clayton-gonsalves: It;s out of habit but I reason that a `map[T]struct{}` is intuitive as a set compared to a `map[T]bool`. A `map[T]bool` could have an element as `false` which doesn't make much sense and means you need to have an if check to be sure. 

If I am not mistaken, an empty struct also doesn't allocate memory. although at the scale we are using it, it won't make that big of a difference. If there is a precedence to use `map[T]bool` in this project I am happy to change it.

internal/xdscache/v3/endpointslicetranslator.go
```diff
@@ -0,0 +1,458 @@
...
+				}
+
+				// Match by port name.
+				if port.Name != "" && p.Name != nil && port.Name != *p.Name {
+					continue
+				}
+
+				// we can safely take the first element here.
+				// Refer to: https://issue.k8s.io/106267
+				addr := envoy_v3.SocketAddress(endpoint.Addresses[0], int(*p.Port))
```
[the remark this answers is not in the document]
clayton-gonsalves: done

internal/xdscache/v3/endpointslicetranslator.go
```diff
@@ -0,0 +1,458 @@
...
+// endpointSliceMap may be nil, in which case, the result is also nil.
+func (c *EndpointSliceCache) RecalculateEndpoints(port, healthPort v1.ServicePort, endpointSliceMap map[string]*discoveryv1.EndpointSlice) []*LoadBalancingEndpoint {
+	var lb []*LoadBalancingEndpoint
+	uniqueEndpoints := make(map[string]struct{}, 0)
+	var healthCheckPort int32
+
+	for _, endpointSlice := range endpointSliceMap {
+		sort.Slice(endpointSlice.Endpoints, func(i, j int) bool {
+			return endpointSlice.Endpoints[i].Addresses[0] < endpointSlice.Endpoints[j].Addresses[0]
+		})
```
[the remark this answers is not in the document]
clayton-gonsalves: There is a discussion around it [here](https://github.com/projectcontour/contour/pull/5745#discussion_r1324600660),. 
I tried to follow the pattern from the endpoint implementation.

skriss:
@tsaarni @sunjayBhatia anything else blocking here? I'd definitely like to get this into the upcoming 1.27 RC. Can always address minor stuff between RC and GA if needed.

skriss (approved):
LGTM, will get this into the upcoming 1.27 RC

sunjayBhatia:
> @tsaarni @sunjayBhatia anything else blocking here? I'd definitely like to get this into the upcoming 1.27 RC. Can always address any remaining non-blocking items between RC and GA if needed.

nothing blocking from me 👍🏽

PR #5752, opened 2023-09-13
Fixup: Sort paths based on length rather than lexi

davinci26:
Since Envoy is greedy matching path routes, order is important. Contour decides to sort the routes in a way that is not really intuitive and can lead to suprises.

In particular even tho the comment in the code state that routes are ordered based on legnth the reality is that they are sorted based on string comparison. This PR fixes this.

* I think the current behaviour doesnt make much sense and it is a bit brittle.
* Updating the behaviour has significant update risk since there might be folks that rely on this routing behaviour without really knowing it.
* Should we even merge this PR? I am of two minds and I would like some input:

1. Option (1): Merge it as and make a clear changelog/announcement about the fix
2. Option (2): Create a config flag with a feature-flag e.g. `route_sorting_strategy` and switch the implementation to not do sorting when the flag is present. That way it allows folks to opt-out from the sorting as they need to.

Longest path based matching kinda makes sense to me now that I know about it, but it is rough edge than needs users to be familiar with contour and it is harder to socialize in larger teams.

### DIFF
--- a/changelogs/unreleased/5752-davinci26-major.md
+++ b/changelogs/unreleased/5752-davinci26-major.md
@@ -1,8 +1,8 @@
 ## Fix bug with algorithm used to sort Envoy regex/prefix path rules
 
-Envoy is greedy matching path routes and as a result order is important. Contour to produce consistent routing tables in the presence of HTTPProxy inclusion and Ingress Conformance sorts the routes before updating Envoy.
+Envoy greedy matches routes and as a result the order route matches are presented to Envoy is important. Contour attempts to produce consistent routing tables so that the most specific route matches are given preference. This is done to facilitate consistency when using HTTPProxy inclusion and provide a uniform user experience for route matching to be inline with Ingress and Gateway API Conformance.
 
-This changes fixes the sorting algorithm used for `Prefix` and `Regex` based path matching. Previously the algorithm lexicographically sorted the regex routes instead of sorting them based on the length of the `Prefix`|`Regex`.
+This changes fixes the sorting algorithm used for `Prefix` and `Regex` based path matching. Previously the algorithm lexicographically sorted based on the path match string instead of sorting them based on the length of the `Prefix`|`Regex`.
 
 ### How to update safely
 
@@ -10,5 +10,5 @@ Caution is advised if you update Contour and you are operating large routing tab
 
 1. Deploy a duplicate Contour installation that parses the same CRDs
 2. Port-forward to the Envoy admin interface [docs](https://projectcontour.io/docs/v1.3.0/troubleshooting/)
-3. Access `http://127.0.0.1:9001/config_dump` and compare the configuration of Envoy. In particular the routes and their order.
+3. Access `http://127.0.0.1:9001/config_dump` and compare the configuration of Envoy. In particular the routes and their order. The prefix routes might be changing in order, so if they are you need to verify that the route matches as expected.
 
--- a/changelogs/unreleased/5752-davinci26-major.md
+++ b/changelogs/unreleased/5752-davinci26-major.md
@@ -4,6 +4,7 @@ Envoy greedy matches routes and as a result the order route matches are presente
 
 This changes fixes the sorting algorithm used for `Prefix` and `Regex` based path matching. Previously the algorithm lexicographically sorted based on the path match string instead of sorting them based on the length of the `Prefix`|`Regex`. i.e. Longer prefix/regexes will be sorted first in order to give preference to more specific routes, then lexicographic sorting for things of the same length.
 
+Note that for prefix matching, this change is _not_ expected to change the relative ordering of more specific prefixes vs. less specific ones when the more specific prefix match string has the less specific one as a prefix, e.g. `/foo/bar` will continue to sort before `/foo`. However, relative ordering of other combinations of prefix matches may change per the above description.
 ### How to update safely
 
 Caution is advised if you update Contour and you are operating large routing tables. We advise you to:
--- a/changelogs/unreleased/5752-davinci26-major.md
+++ b/changelogs/unreleased/5752-davinci26-major.md
@@ -1,6 +1,6 @@
 ## Fix bug with algorithm used to sort Envoy regex/prefix path rules
 
-Envoy is greedy matching path routes and as a result order is important. Contour to produce consistent routing tables in the precence of HTTPProxy inclusion and Ingress Conformance sorts the routes before updating Envoy.
+Envoy is greedy matching path routes and as a result order is important. Contour to produce consistent routing tables in the presence of HTTPProxy inclusion and Ingress Conformance sorts the routes before updating Envoy.
 
 This changes fixes the sorting algorithm used for `Prefix` and `Regex` based path matching. Previously the algorithm lexicographically sorted the regex routes instead of sorting them based on the length of the `Prefix`|`Regex`.
 
--- a/changelogs/unreleased/5752-davinci26-major.md
+++ b/changelogs/unreleased/5752-davinci26-major.md
@@ -2,7 +2,7 @@
 
 Envoy greedy matches routes and as a result the order route matches are presented to Envoy is important. Contour attempts to produce consistent routing tables so that the most specific route matches are given preference. This is done to facilitate consistency when using HTTPProxy inclusion and provide a uniform user experience for route matching to be inline with Ingress and Gateway API Conformance.
 
-This changes fixes the sorting algorithm used for `Prefix` and `Regex` based path matching. Previously the algorithm lexicographically sorted based on the path match string instead of sorting them based on the length of the `Prefix`|`Regex`.
+This changes fixes the sorting algorithm used for `Prefix` and `Regex` based path matching. Previously the algorithm lexicographically sorted based on the path match string instead of sorting them based on the length of the `Prefix`|`Regex`. i.e. Longer prefix/regexes will be sorted first in order to give preference to more specific routes, then lexicographic sorting for things of the same length.
 
 ### How to update safely
 
--- a/changelogs/unreleased/5752-davinci26-major.md
+++ b/changelogs/unreleased/5752-davinci26-major.md
@@ -1,0 +1,14 @@
+## Fix bug with algorithm used to sort Envoy regex/prefix path rules
+
+Envoy is greedy matching path routes and as a result order is important. Contour to produce consistent routing tables in the precence of HTTPProxy inclusion and Ingress Conformance sorts the routes before updating Envoy.
+
+This changes fixes the sorting algorithm used for `Prefix` and `Regex` based path matching. Previously the algorithm lexicographically sorted the regex routes instead of sorting them based on the length of the `Prefix`|`Regex`.
+
+### How to update safely
+
+Caution is advised if you update Contour and you are operating large routing tables. We advise you to:
+
+1. Deploy a duplicate Contour installation that parses the same CRDs
+2. Port-forward to the Envoy admin interface [docs](https://projectcontour.io/docs/v1.3.0/troubleshooting/)
+3. Access `http://127.0.0.1:9001/config_dump` and compare the configuration of Envoy. In particular the routes and their order.
+
--- a/internal/featuretests/v3/replaceprefix_test.go
+++ b/internal/featuretests/v3/replaceprefix_test.go
@@ -478,21 +478,30 @@ func artifactoryDocker(t *testing.T) {
 		Resources: resources(t,
 			envoy_v3.RouteConfiguration("ingress_http",
 				envoy_v3.VirtualHost("artifactory.projectcontour.io",
-
+					&envoy_route_v3.Route{
+						Match: routePrefix("/v2/container-external/"),
+						Action: withPrefixRewrite(routeCluster("artifactory/service/8080/da39a3ee5e"),
+							"/artifactory/api/docker/container-external/v2/"),
+					},
 					&envoy_route_v3.Route{
 						Match: routePrefix("/v2/container-sandbox/"),
 						Action: withPrefixRewrite(routeCluster("artifactory/service/8080/da39a3ee5e"),
 							"/artifactory/api/docker/container-sandbox/v2/"),
 					},
 					&envoy_route_v3.Route{
-						Match: routePrefix("/v2/container-sandbox"),
-						Action: withPrefixRewrite(routeCluster("artifactory/service/8080/da39a3ee5e"),
-							"/artifactory/api/docker/container-sandbox/v2"),
-					},
-					&envoy_route_v3.Route{
 						Match: routePrefix("/v2/container-release/"),
 						Action: withPrefixRewrite(routeCluster("artifactory/service/8080/da39a3ee5e"),
 							"/artifactory/api/docker/container-release/v2/"),
+					},
+					&envoy_route_v3.Route{
+						Match: routePrefix("/v2/container-external"),
+						Action: withPrefixRewrite(routeCluster("artifactory/service/8080/da39a3ee5e"),
+							"/artifactory/api/docker/container-external/v2"),
+					},
+					&envoy_route_v3.Route{
+						Match: routePrefix("/v2/container-sandbox"),
+						Action: withPrefixRewrite(routeCluster("artifactory/service/8080/da39a3ee5e"),
+							"/artifactory/api/docker/container-sandbox/v2"),
 					},
 					&envoy_route_v3.Route{
 						Match: routePrefix("/v2/container-release"),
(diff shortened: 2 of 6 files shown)


davinci26:
Unclear if the CI will pass, will fix tomorrow tho. Looking for your feedback folks!

internal/featuretests/v3/replaceprefix_test.go
```diff
@@ -478,46 +478,45 @@ func artifactoryDocker(t *testing.T) {
 		Resources: resources(t,
 			envoy_v3.RouteConfiguration("ingress_http",
 				envoy_v3.VirtualHost("artifactory.projectcontour.io",
-
 					&envoy_route_v3.Route{
-						Match: routePrefix("/v2/container-sandbox/"),
+						Match: routePrefix("/v2/container-external/"),
```
sunjayBhatia, line 482: though this makes the order different, that is basically just cosmetic, should still be a no-op change for prefix match routes since the matching mechanism does a lexical/simple string comparison match against what is configured here) (as opposed to a regex match)

internal/sorter/sorter.go
```diff
@@ -296,30 +296,36 @@ func (s routeSorter) Less(i, j int) bool {
...
+					return a.PrefixMatchType == dag.PrefixMatchSegment
 				}
-				// Segment prefixes sort first as they are more specific.
-				return a.PrefixMatchType == dag.PrefixMatchSegment
 			}
 		}
 	case *dag.RegexMatchCondition:
 		switch b := s[j].PathMatchCondition.(type) {
 		case *dag.RegexMatchCondition:
-			cmp := strings.Compare(a.Regex, b.Regex)
```
sunjayBhatia, line 317: yeah i think this was a long-standing bug, we always sorta said that length of the regex was a proxy for complexity/specificity, its sort of a rough guess/heuristic, but this original implementation does not achieve what the stated API was

internal/sorter/sorter.go
```diff
@@ -296,30 +296,36 @@ func (s routeSorter) Less(i, j int) bool {
...
-			switch cmp {
-			case 1:
+			switch {
+			case len(a.Regex) > len(b.Regex):
 				// Sort longest regex first.
 				return true
-			case -1:
+			case len(a.Regex) < len(b.Regex):
 				return false
 			default:
```
sunjayBhatia, line 330: since we store routes and access them as a map, i think we will need some sort of fallback to sort them stably if the lengths are the same, otherwise we may have routes change order

sunjayBhatia (commented):
I'm currently leaning toward marking this as a major changelog and proceeding, since this really is a bug fix to get back to how the intended behavior is described, but would appreciate further thoughts

sunjayBhatia:
Also @davinci26 if you had an example that caused you to run into the problem we discussed on regex match ordering, would be great to add to the tests if not already there at least for documentation/posterity purposes, can of course change any specific strings from your exact example

internal/featuretests/v3/replaceprefix_test.go
davinci26: prefix did similar `lexi` sorting which I guess is also not correct

davinci26:
> Also @davinci26 if you had an example that caused you to run into the problem we discussed on regex match ordering, would be great to add to the tests if not already there at least for documentation/posterity purposes, can of course change any specific strings from your exact example

I added the test in `shorter_test.go`

```
// Note that regex matches sort before prefix matches.
{
	PathMatchCondition: matchRegex("/athis/is/the/longest"),
},
{
	PathMatchCondition: matchRegex(`/foo((\/).*)*`),
},
```
See: https://go.dev/play/p/nIn3eyd-XIX

internal/featuretests/v3/replaceprefix_test.go
davinci26: I might be missing the point comment so if I do let me know 😅 But basically prefix routing also had lexical/simple string comparison so I fixed that part as well. Since it is important in both cases, both regexp and prefix. So this is why we needed to change this test

davinci26:
@sunjayBhatia all comments addressed, the CI should go green and it should be ready for merge

internal/sorter/sorter.go
clayton-gonsalves: nit: can we add a test case for this scenario?

clayton-gonsalves (commented):
lgtm, module one question/request

internal/sorter/sorter.go
davinci26: I believe a test already exists in `TestSortRoutesLongestHeaders` where everything is `PathMatchCondition: matchRegex("/pathregex"),`

clayton-gonsalves (approved):

internal/sorter/sorter.go
sunjayBhatia: see this test case:

```diff
diff --git a/internal/sorter/sorter_test.go b/internal/sorter/sorter_test.go
index 90a158c8..98256a84 100644
--- a/internal/sorter/sorter_test.go
+++ b/internal/sorter/sorter_test.go
@@ -284,6 +284,12 @@ func TestSortRoutesPathMatch(t *testing.T) {
                {
                        PathMatchCondition: matchRegex(`/foo((\/).*)*`),
                },
+               {
+                       PathMatchCondition: matchRegex("/bar.*"),
+               },
+               {
+                       PathMatchCondition: matchRegex("/foo.*"),
+               },
                {
                        PathMatchCondition: matchRegex("/"),
                },
@@ -712,6 +718,6 @@ func shuffleAndCheckSort[T any](t *testing.T, want []T) {
                have := shuffleSlice(want)

                sort.Stable(For(have))
-               assert.Equal(t, want, have)
+               assert.Equalf(t, want, have, "failed during trial %d", i)
        }
 }
```

we run all these sorter tests through a bunch of trials and you'll see if you run `TestSortRoutesPathMatch` that the test fails a few times but not all with either ordering of those two route matches, since those regexes are the same length

can be fixed if we have a fallback to sorting lexicographically like we do with prefix matches (and having the correct expected order)

internal/featuretests/v3/replaceprefix_test.go
sunjayBhatia: oh yeah my comment was just for context for any other readers, to offer some justification for why this change happens in the ordering

changelogs/unreleased/5752-davinci26-major.md
```diff
@@ -0,0 +1,14 @@
+## Fix bug with algorithm used to sort Envoy regex/prefix path rules
+
+Envoy is greedy matching path routes and as a result order is important. Contour to produce consistent routing tables in the presence of HTTPProxy inclusion and Ingress Conformance sorts the routes before updating Envoy.
```
sunjayBhatia, line 3:
```suggestion
Envoy greedy matches routes and as a result the order route matches are presented to Envoy is important. Contour attempts to produce consistent routing tables so that the most specific route matches are given preference. This is done to facilitate consistency when using HTTPProxy inclusion and provide a uniform user experience for route matching to be inline with Ingress and Gateway API Conformance.
```

changelogs/unreleased/5752-davinci26-major.md
```diff
@@ -0,0 +1,14 @@
...
+
+This changes fixes the sorting algorithm used for `Prefix` and `Regex` based path matching. Previously the algorithm lexicographically sorted the regex routes instead of sorting them based on the length of the `Prefix`|`Regex`.
```
sunjayBhatia, line 5:
```suggestion
This changes fixes the sorting algorithm used for `Prefix` and `Regex` based path matching. Previously the algorithm lexicographically sorted based on the path match string instead of sorting them based on the length of the `Prefix`|`Regex`.
```

changelogs/unreleased/5752-davinci26-major.md
```diff
@@ -0,0 +1,14 @@
...
+
+### How to update safely
+
+Caution is advised if you update Contour and you are operating large routing tables. We advise you to:
+
+1. Deploy a duplicate Contour installation that parses the same CRDs
+2. Port-forward to the Envoy admin interface [docs](https://projectcontour.io/docs/v1.3.0/troubleshooting/)
+3. Access `http://127.0.0.1:9001/config_dump` and compare the configuration of Envoy. In particular the routes and their order.
```
sunjayBhatia, line 13: It may be surprising to users that their prefix route matches change order since we are now sorting on length, so maybe good to mention here if we're suggesting inspecting the routing table

changelogs/unreleased/5752-davinci26-major.md
sunjayBhatia: Maybe add a sentence spelling out what the change will be: longer prefix/regexes will be sorted first in order to give preference to more specific routes, then lexicographic sorting for things of the same length

changelogs/unreleased/5752-davinci26-major.md
davinci26: fixed in de92073f94f2426279a4f99d39029d97355b61d2

changelogs/unreleased/5752-davinci26-major.md
davinci26: fixed in de92073f94f2426279a4f99d39029d97355b61d2

internal/sorter/sorter.go
davinci26: I think I stabilized it now with de92073f94f2426279a4f99d39029d97355b61d2

changelogs/unreleased/5752-davinci26-major.md
davinci26: fixed in a4ef0f6d

davinci26:
@sunjayBhatia ptal, this should be good now

internal/sorter/sorter.go
```diff
@@ -296,33 +297,44 @@ func (s routeSorter) Less(i, j int) bool {
 	switch a := s[i].PathMatchCondition.(type) {
 	case *dag.PrefixMatchCondition:
 		if b, ok := s[j].PathMatchCondition.(*dag.PrefixMatchCondition); ok {
-			cmp := strings.Compare(a.Prefix, b.Prefix)
```
skriss, line 299: I think despite how any comments/docs may explain this, the important property that is being expressed here is that if there is a route with a prefix match of `/foo` and a route with a prefix match of `/foo/bar`, then the latter will sort first/take precedence when matching. In other words, more specific prefix matches take precedence over less specific ones. The current implementation actually implements that logic fine, since reverse lexicographic order will put more specific prefixes first.

Are there actually any scenarios where reverse lexicographic prefix match sorting does not give the desired result, but reverse length sorting does? As far as I can tell there shouldn't be, since for routing purposes we only care about relative ordering between pairs of prefixes where one wholly contains the other (e.g. `/foo/bar` vs. `/foo`).

Either way, I think the code change is fine/good since it's a little more clear, and will be more consistent with sorting for other match types, docs, comments, etc., but if the above is true then I would say the changelog should focus mainly on the behavior change for **regex** matching since that will definitely impact users. We can note that there will be some changes in prefix match ordering as well, but that they aren't expected to be impactful for routing behavior.

I can suggest some words for the changelog if we're on the same page about this, just want to get folks' input here first.

internal/sorter/sorter.go
```diff
@@ -296,33 +297,44 @@ func (s routeSorter) Less(i, j int) bool {
...
+			case len(a.Prefix) < len(b.Prefix):
 				return false
 			default:
-				if a.PrefixMatchType == b.PrefixMatchType {
-					return compareRoutesByMethodHeaderQueryParams(s[i], s[j])
+				cmp := strings.Compare(a.Prefix, b.Prefix)
+				switch cmp {
+				case 1:
+					return true
+				case 2:
```
skriss, line 311: I think this should be `case -1`?

davinci26: yes

davinci26: ty

internal/sorter/sorter.go
davinci26: What you are saying makes sense, I am just not sure if there is any case we might be missing

davinci26: fair point, I dont think I can think of anything that would affect this case

internal/sorter/sorter.go
```diff
@@ -296,33 +297,44 @@ func (s routeSorter) Less(i, j int) bool {
...
+			switch {
+			case len(a.Regex) > len(b.Regex):
 				// Sort longest regex first.
 				return true
-			case -1:
+			case len(a.Regex) < len(b.Regex):
 				return false
 			default:
-				return compareRoutesByMethodHeaderQueryParams(s[i], s[j])
+				v, err := compareRoutesByMethodHeaderQueryParams(s[i], s[j])
```
sunjayBhatia, line 333: found another test case that does not consistently sort with this implementation but did with the previous (/pathregex2 always sorted before all the /pathregex1 matches):
```diff
diff --git a/internal/sorter/sorter_test.go b/internal/sorter/sorter_test.go
index aac862a4..3958bca4 100644
--- a/internal/sorter/sorter_test.go
+++ b/internal/sorter/sorter_test.go
@@ -392,25 +392,31 @@ func TestSortRoutesLongestHeaders(t *testing.T) {
                        PathMatchCondition: matchExact("/pathexact"),
                },
                {
-                       PathMatchCondition: matchRegex("/pathregex"),
+                       PathMatchCondition: matchRegex("/pathregex1"),
                        HeaderMatchConditions: []dag.HeaderMatchCondition{
                                exactHeader("header-name", "header-value"),
                        },
                },
                {
-                       PathMatchCondition: matchRegex("/pathregex"),
+                       PathMatchCondition: matchRegex("/pathregex1"),
                        HeaderMatchConditions: []dag.HeaderMatchCondition{
                                presentHeader("header-name"),
                        },
                },
                {
-                       PathMatchCondition: matchRegex("/pathregex"),
+                       PathMatchCondition: matchRegex("/pathregex1"),
                        HeaderMatchConditions: []dag.HeaderMatchCondition{
                                exactHeader("long-header-name", "long-header-value"),
                        },
                },
                {
-                       PathMatchCondition: matchRegex("/pathregex"),
+                       PathMatchCondition: matchRegex("/pathregex2"),
+                       HeaderMatchConditions: []dag.HeaderMatchCondition{
+                               presentHeader("header-name"),
+                       },
+               },
+               {
+                       PathMatchCondition: matchRegex("/pathregex1"),
                },
                {
                        PathMatchCondition: matchPrefixSegment("/path"),
@@ -718,6 +724,6 @@ func shuffleAndCheckSort[T any](t *testing.T, want []T) {
                have := shuffleSlice(want)

                sort.Stable(For(have))
-               assert.Equal(t, want, have)
+               assert.Equalf(t, want, have, "failed trial #%d", i)
        }
 }
```

davinci26: I would expect this case `compareRoutesByMethodHeaderQueryParams` to return an error

davinci26: and to fallback to lexi sorting

sunjayBhatia: you can imagine another example of regexes of the same length where you don't want the matches to be "interleaved" like could happen with what we have in this PR (e.g. say path matches for `/foo.*` and `/.*foo` with/without header matches)

I was able to get the test to consistently sort (and IMO sort in an order that makes a little more sense) by doing the lexical comparison before comparing header/etc. matches
```diff
diff --git a/internal/sorter/sorter.go b/internal/sorter/sorter.go
index 3a6f4de1..d2d65b69 100644
--- a/internal/sorter/sorter.go
+++ b/internal/sorter/sorter.go
@@ -330,11 +330,16 @@ func (s routeSorter) Less(i, j int) bool {
                        case len(a.Regex) < len(b.Regex):
                                return false
                        default:
-                               v, err := compareRoutesByMethodHeaderQueryParams(s[i], s[j])
-                               if err != nil {
-                                       return strings.Compare(a.Regex, b.Regex) == -1
+                               cmp := strings.Compare(a.Regex, b.Regex)
+                               switch cmp {
+                               case 1:
+                                       return true
+                               case -1:
+                                       return false
+                               default:
+                                       v, _ := compareRoutesByMethodHeaderQueryParams(s[i], s[j])
+                                       return v
                                }
-                               return v
                        }
                case *dag.PrefixMatchCondition:
                        return true
```
(I think you can get rid of the added error return too since we should never get to a case where the route match conditions are fully exactly equal)

davinci26: > you can imagine another example of regexes of the same length where you don't want the matches to be "interleaved" like could happen with what we have in this PR (e.g. say path matches for /foo.* and /.*foo with/without header matches)

This is how I had initially but I thought that we wanted `compareRoutesByMethodHeaderQueryParams` for more specificity first

davinci26: But I agree this change helps quite a bit with stabilizing the sorting algo

davinci26: fixed in b7f8c9eff14519b21bdf30c927b89637a288649c

sunjayBhatia (approved):
Will leave open for a bit so @skriss and others can get more thoughts in

But overall I think this looks good, LGTM, thanks for fixing this up!

changelogs/unreleased/5752-davinci26-major.md
```diff
@@ -0,0 +1,14 @@
+## Fix bug with algorithm used to sort Envoy regex/prefix path rules
+
+Envoy greedy matches routes and as a result the order route matches are presented to Envoy is important. Contour attempts to produce consistent routing tables so that the most specific route matches are given preference. This is done to facilitate consistency when using HTTPProxy inclusion and provide a uniform user experience for route matching to be inline with Ingress and Gateway API Conformance.
+
+This changes fixes the sorting algorithm used for `Prefix` and `Regex` based path matching. Previously the algorithm lexicographically sorted based on the path match string instead of sorting them based on the length of the `Prefix`|`Regex`. i.e. Longer prefix/regexes will be sorted first in order to give preference to more specific routes, then lexicographic sorting for things of the same length.
+
```
skriss, line 6:
```suggestion

Note that for prefix matching, this change is _not_ expected to change the relative ordering of more specific prefixes vs. less specific ones when the more specific prefix match string has the less specific one as a prefix, e.g. `/foo/bar` will continue to sort before `/foo`. However, relative ordering of other combinations of prefix matches may change per the above description.
```

skriss:
Otherwise LGTM

davinci26:
`DCO` is not enforced with commit suggestions

davinci26:
Unless there are any flakes @skriss this should be good-to-go

PR #5755, opened 2023-09-15
chore: replace deprecated function

yyzxw:
replace `instr.FromInt` to `instr.FormInt32`, FromInt has been deprecated

### DIFF
--- a/apis/projectcontour/v1alpha1/contourdeployment.go
+++ b/apis/projectcontour/v1alpha1/contourdeployment.go
@@ -77,7 +77,7 @@ type ContourDeploymentSpec struct {
 type ContourSettings struct {
 	// Deprecated: Use `DeploymentSettings.Replicas` instead.
 	//
-	// Replicas is the desired number of Contour replicas. If if unset,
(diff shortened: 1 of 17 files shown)


sunjayBhatia (changes requested):
from a quick search, seems like there are still a bunch of places where this function is still used (e.g. `internal/xdscache/v3/route_test.go`), would be great to fix those up as well

apis/projectcontour/v1alpha1/contourdeployment.go
```diff
@@ -77,7 +77,7 @@ type ContourDeploymentSpec struct {
 type ContourSettings struct {
 	// Deprecated: Use `DeploymentSettings.Replicas` instead.
 	//
-	// Replicas is the desired number of Contour replicas. If if unset,
+	// Replicas is the desired number of Contour replicas. If unset,
```
skriss, line 80: You will need to run `make generate` and commit the results to reflect this change in the CRD YAML and get CI to pass.

Issue #5137, opened 2023-03-03
Envoy Thrift Proxy


shadialtarsha:
**What question do you have?:**
I was wondering if there are any plans to add support for [Envoy Thrift Proxy](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/thrift_proxy/v3/thrift_proxy.proto) to Contour.
Thank you!

sunjayBhatia:
We do not have plans to do so at this time, but would accept community contributions to start experimenting and designing such a feature!

shadialtarsha:
Hey @sunjayBhatia 
Thanks for replying! 
We are evaluating Contour as a potential ingress controller and having a Thrift proxy is an essential feature for us and we are happy to contribute! 
Let me open a design proposal for this!

PR #5769, opened 2023-09-19
design: Thrift Proxy Support

shadialtarsha:
Design: Adding Thrift Proxy Support to Contour

Addresses: https://github.com/projectcontour/contour/issues/5137

### DIFF
--- a/design/thrift-proxy-design.md
+++ b/design/thrift-proxy-design.md
@@ -1,0 +1,215 @@
+# Thrift Proxy Design
+
+# Abstract
+
+Thrift Proxy is an L7 Envoy proxy that is used for routing Thrift protocol. 
+This document outlines the rationale, objectives, architectural design, and technical details of integrating Thrift Proxy with Contour. 
+The integration of Thrift Proxy into Contour opens up new possibilities for routing and managing Thrift-based applications.
+
+
+# Background
+
+The Thrift Proxy is an essential component in the Envoy ecosystem, offering the ability to route Thrift protocol traffic at the application layer (Layer 7). 
+This functionality is particularly valuable for organizations that rely on the Thrift Protocol as their primary RPC implementation.
+
+Key features of the Envoy Thrift Proxy include:
+
+* Route Matching: The Thrift Proxy enables granular route matching based on various criteria, such as the Thrift method name, Thrift service and name, and a set of headers. 
+This level of specificity ensures accurate routing of Thrift requests to their intended destinations.
+* Rate Limiting: It supports rate limit filtering within the Thrift filter chain.
+* Cluster Management: The Thrift Proxy offers support for weighted clusters and mirrored clusters, enabling efficient load balancing and fail-over strategies.
+* Route Discovery: It facilitates route discovery mechanisms, making it easier to dynamically configure and manage Thrift routing rules.
+
+### Contour's Current Capabilities
+
+As of the current Contour release, the project provides "TCP Proxying" functionality. 
+While this is valuable, it does not include native support for Thrift Proxying.
+
+### The Need for Thrift Proxy in Contour
+
+The absence of Thrift Proxy support in Contour creates limitations for organizations that heavily rely on the Thrift Protocol. 
+As the Thrift Protocol continues to be a popular choice for inter-service communication, integrating Thrift Proxy into Contour is a logical step forward. 
+This integration will address the following key needs:
+* Protocol Support: Adding Thrift Proxy support ensures that Contour can seamlessly route and manage Thrift-based applications alongside its existing capabilities for HTTP and gRPC traffic.
+* Enhanced Routing Control: With Thrift Proxy, Contour gains the ability to implement fine-grained routing logic for Thrift requests, enabling intelligent routing decisions based on Thrift-specific criteria.
+* Interoperability: Thrift Proxy ensures that Contour can easily integrate with systems and services utilizing the Thrift Protocol.
+
+By introducing Thrift Proxy support into Contour, we aim to elevate Contour's versatility, making it a more comprehensive solution for modern, distributed application architectures.
+
+## Goals
+
+* Implement an equivalent to HTTPProxy, called ThriftProxy, which allows users to define their Thrift routing logic and services: This goal aims to provide users with a dedicated resource, ThriftProxy, 
+where they can specify custom Thrift routing rules, enabling fine-grained control over how Thrift traffic is handled within Contour.
+* Add a new Listener for the Thrift protocol: This goal involves enhancing Contour's capabilities by introducing a new Listener specifically designed to handle Thrift protocol traffic.
+* Implement ThriftRoute discovery: ThriftRoute discovery mechanisms will allow for dynamic configuration and management of Thrift routing rules.
+
+## Non Goals
+
+* Thrift protocol will be only supported by the ThriftProxy CRD, excluding support for K8s ingresses: In this iteration, the focus is solely on the ThriftProxy CRD for defining Thrift routing logic. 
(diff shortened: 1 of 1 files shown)


design/thrift-proxy-design.md
```diff
@@ -0,0 +1,215 @@
...
+}
+```
+* ThriftRouteMatch: Defines the matching rules that incoming Thrift traffic must satisfy to be routed according to this rule. These rules can be based on factors such as the Thrift method name and custom headers.
+* ThriftRouteAction: Specifies the cluster that should handle the Thrift traffic that matches the given rule. This cluster configuration includes details such as load balancing policies, fail-over strategies, and rate limiting.
+
+## Aggregated Service Discovery
+Aggregated Service Discovery (ASD) is a dynamic resource management mechanism that allows Envoy to fetch configuration resources from multiple sources, 
+aggregate them, and ensure that they are consistently and efficiently distributed to proxy instances. In the context of our project, ASD serves as the backbone of Thrift route discovery.
+
+### Enabling Aggregated Service Discovery
```
sunjayBhatia, line 186: We've wanted to take advantage of go-control-plane's xDS server implementation (to utilize ADS etc.) for quite a while, but haven't had the time to iron out details relating to making it the default. I would prefer to pursue that avenue to enable ADS rather than adding to the existing contour-specific implementation of the xDS server which we intended to get rid of at some point if ADS is required to make this work.

design/thrift-proxy-design.md
```diff
@@ -0,0 +1,215 @@
...
+```
+&envoy_thrift_v3.Route{
+    Match: ThriftRouteMatch(rule),
+    Route: ThriftRouteAction(rule),
+}
+```
+* ThriftRouteMatch: Defines the matching rules that incoming Thrift traffic must satisfy to be routed according to this rule. These rules can be based on factors such as the Thrift method name and custom headers.
+* ThriftRouteAction: Specifies the cluster that should handle the Thrift traffic that matches the given rule. This cluster configuration includes details such as load balancing policies, fail-over strategies, and rate limiting.
+
+## Aggregated Service Discovery
```
sunjayBhatia, line 182: Huh interesting that TRDS is limited to ADS basically, might be worth adding a link to the field where it says that: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/thrift_proxy/v3/thrift_proxy.proto#extensions-filters-network-thrift-proxy-v3-trds (we can't use the plain gRPC SOTW api config source)

design/thrift-proxy-design.md
```diff
@@ -0,0 +1,215 @@
...
+By introducing Thrift Proxy support into Contour, we aim to elevate Contour's versatility, making it a more comprehensive solution for modern, distributed application architectures.
+
+## Goals
+
+* Implement an equivalent to HTTPProxy, called ThriftProxy, which allows users to define their Thrift routing logic and services: This goal aims to provide users with a dedicated resource, ThriftProxy, 
+where they can specify custom Thrift routing rules, enabling fine-grained control over how Thrift traffic is handled within Contour.
+* Add a new Listener for the Thrift protocol: This goal involves enhancing Contour's capabilities by introducing a new Listener specifically designed to handle Thrift protocol traffic.
+* Implement ThriftRoute discovery: ThriftRoute discovery mechanisms will allow for dynamic configuration and management of Thrift routing rules.
+
+## Non Goals
```
sunjayBhatia, line 46: Would add Thrift over HTTP is a non-goal, since you can use existing HTTP routing capabilities of Contour

design/thrift-proxy-design.md
```diff
@@ -0,0 +1,215 @@
...
+### Thrift Listener
+A dedicated Thrift Listener is responsible for receiving incoming Thrift traffic on a distinct port separate from gRPC and HTTP protocols.
+### Thrift RouteConfiguration
+Thrift RouteConfiguration defines the Envoy configurations that correspond to user-defined routing rules.
+### Aggregated Service Discovery
+Aggregated Service Discovery (ADS) to enable Thrift route discovery.
+
+# Detailed Design
+
+## ThriftProxy CRD
```
sunjayBhatia, line 103: Would it be possible for you all to use Gateway API and a new *Route CRD instead? (i.e. ThriftRoute CRD)

I say this for a couple reasons, but mainly because if we expect a unique Envoy Listener specific to Thrift Proxy routes (i.e. not the existing two Listeners on plain HTTP or HTTPS ports), we need to leverage Contour's dynamic Listener support which we only really intended to support with Gateway API.

We should probably add some discussion to this design about Envoy Listeners, how we expect Thrift Proxy support to interact with them.

design/thrift-proxy-design.md
```diff
@@ -0,0 +1,215 @@
+# Thrift Proxy Design
+
+# Abstract
+
+Thrift Proxy is an L7 Envoy proxy that is used for routing Thrift protocol. 
+This document outlines the rationale, objectives, architectural design, and technical details of integrating Thrift Proxy with Contour. 
+The integration of Thrift Proxy into Contour opens up new possibilities for routing and managing Thrift-based applications.
+
+
+# Background
```
sunjayBhatia, line 10: A useful link I found looking into envoy + Thrift Proxy history, the original proposal to add Thrift support to Envoy: https://docs.google.com/document/d/1QuiZMT1ruetavVHsj4Sv55YpBFVxTEzJj-ZpOJqAboc/edit

design/thrift-proxy-design.md
shadialtarsha: Thanks for pointing out the Gateway API point, and I agree with you on adding ThriftRoute CRD for the listeners.

I haven't gotten my hands dirty with the Gateway API stuff, and we don't use it in my company yet. That is why I tried to be explicit about not having it as a Goal of this design doc.

However, I can over 2 options here:
- Open an issue to track adding support for ThriftRoute to Gateway API
- I can include it in this design if we agree that it is important and we should cover it in early stages.

design/thrift-proxy-design.md
shadialtarsha: Just to make sure I understand correctly what we are talking about here. 
We are talking about this snippet of code, right?
```
	switch x.config.Type {
	case contour_api_v1alpha1.EnvoyServerType:
		v3cache := contour_xds_v3.NewSnapshotCache(false, log)
		x.snapshotHandler.AddSnapshotter(v3cache)
		contour_xds_v3.RegisterServer(envoy_server_v3.NewServer(ctx, v3cache, contour_xds_v3.NewRequestLoggingCallbacks(log)), grpcServer)
	case contour_api_v1alpha1.ContourServerType:
		contour_xds_v3.RegisterServer(contour_xds_v3.NewContourServer(log, xdscache.ResourcesOf(x.resources)...), grpcServer)
	default:
		// This can't happen due to config validation.
		log.Fatalf("invalid xDS server type %q", x.config.Type)
	}
```

I didn't know that Contour has its own implementation of the xDS before you mentioned it now 😅 

So yeah you probably read below, ADS is required for this work. 

Now I am not sure if such a migration should happen in this design or not.

I am happy to work on making this migration in another work thread if you have an issue open or a design doc for the outcome.

sunjayBhatia: Yep that is the switch between the two xDS server implementations

design/thrift-proxy-design.md
sunjayBhatia: I think this is still an open point to add to this doc, the above sections mention adding a dedicated Thrift Listener but would be good to add some more detail to this Detailed Design section on this topic (how will the port used for Thrift be configured, what does the filter chain/match look like, etc.)

sunjayBhatia (commented):
left an additional comment, but should get some more thoughts from @projectcontour/maintainers

PR #5772, opened 2023-09-21
Stops route sorting on `HTTPProxy` CRD

davinci26:
Stops route sorting on `HTTPProxy` CRD

### Goal:

Contour generates the Envoy routes for each virtual host in the same
order as they are in the `HTTPProxy`. This benefits users from having no
suprises between the Envoy actual routing table and what they described in the
CRD.

---

### Core changes:

* Change the underlying data structure in `dag.go` and `VirtualHost` to be a slice such that we can maintain the order.
* Change the sorting to be conditional on the caller wanting it. This is used to implement sorting for virtual hosts which are "partially" controlled by an Ingress/ApiGateway so we can have conformance.
* Change the implementation of `expandPrefixMatches` to maintain the order of the original list.

### Work Pending

- [x] Add better testing to ensure routes are sorted on the default case 

Signed-off-by: Sotiris Nanopoulos <sotiris.nanopoulos@reddit.com>


davinci26:
Not stale

cmd/contour/serve.go
```diff
@@ -1197,6 +1199,7 @@ func (s *Server) getDAGBuilder(dbc dagBuilderConfig) *dag.Builder {
 			GlobalRateLimitService:        dbc.globalRateLimitService,
 			PerConnectionBufferLimitBytes: dbc.perConnectionBufferLimitBytes,
 			SetSourceMetadataOnRoutes:     true,
+			ShouldSortRoutes:              dbc.sortHTTProxyRoutes,
```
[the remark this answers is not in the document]
shadialtarsha: having the default behavior to `sort` would save you some changes in the code later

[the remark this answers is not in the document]
davinci26: > So sortHTTPRoutes will be false by default (boolean zero value is false in go)
Is that intended?

Yes, the reason being that it facilitates testing which constructs `HTTPProxyProcessors` directly and we want those tests to test the not sorted case by default as it helps quite a bit with test coverage

changelogs/unreleased/5772-davinci26-small.md
```diff
@@ -0,0 +1,5 @@
+## Allow cluster operators to disable route sorting with `HTTPProxy`
+
+This change allows cluster admins to turn on a `feature-flag` that disables route sorting. When this feature flag is turned on routes are sent to Envoy in the same
+order as they are described in the `HTTPProxy` CRD. This allows operators to build more complex routing tables but they need to be careful with changes since now order
+becomes important. Includes are resolved in a depth first fashion.
```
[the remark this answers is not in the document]
davinci26: I think you are right

shadialtarsha (approved):

internal/dag/httpproxy_processor.go
```diff
@@ -112,6 +112,10 @@ type HTTPProxyProcessor struct {
 	// configurable and off by default in order to support the feature
 	// without requiring all existing test cases to change.
 	SetSourceMetadataOnRoutes bool
+
+	// Allows cluster operators to prevent the processor from sorting the routes
+	// before sending them to Envoy.
+	ShouldSortRoutes bool
```
sunjayBhatia, line 118: I would like to try to keep this logic out of the various processors as much as possible and try to only make changes at the dag or xdscache level

e.g. we could have this sort of config on the `dag.Builder` that would get passed down to `dag.DAG.EnsureVirtualHost` and eventually `dag.VirtualHost.AddRoute` to abstract this away from the processors (which tbh should maybe be in their own package at this point)

theoretically the processors should already be adding routes in list-order, and I thought we already did process includes in depth-first order, based on this recursion: https://github.com/projectcontour/contour/blob/6917d85ce41f3050107994c06e6bb5b7039c99ed/internal/dag/httpproxy_processor.go#L726

sunjayBhatia: based on the different dag builder/DAG options, we could maybe keep both a list and map structures for routes on a vhost and only use one of the two based on the passed in config

just off the top of my head thoughts, not fully formed yet

davinci26: I am open to the idea, still need to marinate the ramifications on test coverage for this. That being said I am not sure 
how this would work because:

`EnsureVirtualHost` which effectively constructs the vhost is not aware of the caller. And we have 3 cases

* The caller is an HTTPProxy with controller in `omitRoute` mode: in which case it should set it to `false`
* The caller is an Ingress with controller in `omitRoute` mode: in which case it should set it to `true`
* The caller is an Ingress and an HTTPProxy with controller in `omitRoute` mode: in which case it should set it to `false`

So in that case it seems to me that it is the responsibility of the processor to communicate its preference. Basically the processor is `aware` of the conformance requirements and declares if it needs them for the `vhost` that it manages

davinci26: > theoretically the processors should already be adding routes in list-order, and I thought we already did process includes in depth-first order, based on this recursion:

😅 actually they werent because `expandPrefixMatches` was using a match internally and that re-ordered all routes

PR #5798, opened 2023-10-03
fix spelling errors

skriss:


.codespell.ignorewords
```diff
@@ -6,3 +6,4 @@ od
 als
 wit
 aks
+immediatedly
```
skriss, line 9: this one's in upstream (Kubernetes) godoc/CRD YAML so just ignoring for now.

sunjayBhatia (approved):

PR #5802, opened 2023-10-04
add per-httpproxy http-version support

### LINKED ISSUE #5822: Add Per-HTTPProxy HTTP-Version Support to Address HTTP2 Coalescing Issues with Wildcards
### Description:

With the emergence of the `HTTP2 Coalescing` issue, where using a wildcard can result in complications in SNI-based routing (as fully discussed in multiple issues including https://github.com/projectcontour/contour/issues/1493), we've recognized a critical need to have a per-HTTPProxy mechanism to specify HTTP versions. This issue is exacerbated by browsers like Safari (or potentially underlining WebKit/CFNetwork) which does not retry upon receiving an HTTP 421 response. The proposed workaround for this issue is to universally disable http2 as mentioned in https://github.com/projectcontour/contour/issues/2619 and implemented in https://github.com/projectcontour/contour/pull/2622

We propose, adding a new field to the HTTPProxy spec allowing the specification of HTTP versions to offer for that HTTPProxy. This field comes into play only when `spec.tls` is activated and `spec.tcpproxy` is deactivated.

#### Critical Use Case:
In scenarios where multiple HTTPProxies are served, and a subset of them use the same wildcard certificate, it becomes imperative to have the flexibility to disable http/2 for specific proxies to avoid the mentioned issue. Meanwhile, we want to retain the performance advantages of http/2 for others, and also to be able to use gRPC routing for routes.

### Proposed Solution:

1. Selectively disable http/2 for HTTPProxies using the wildcard certificate, thus bypassing the 421 issue for problematic browsers.
2. Retain the http/2 functionality for other routes to capitalize on grpc and enhanced performance.

### Request:

Providing users the granularity to specify HTTP versions on a per-proxy basis, addressing the HTTP2 Coalescing challenge while balancing performance considerations. Furthermore, to ensure backward compatibility, if the new field in the HTTPProxy spec is not explicitly set, the behavior would default to using `default-http-versions` from the global configuration, provided it's not empty.

therealak12:
This PR adds a new field to HTTPProxy spec which specifies the HTTP versions to offer for that HTTPProxy. It's used only when `spec.tls` is set and `spec.tcpproxy` is not.

A critical use case for the field is when we're serving multiple HTTPProxies where a subset of them use the same wildcard certificate. We can disable http/2 for those utilizing this field while keeping it enabled for others.

Closes #5822


therealak12:
Just to make this active.

therealak12:
Could a maintainer please add a `release-note/small` label to this PR? Appears I don't have access to do so.


image

apis/projectcontour/v1/httpproxy.go
```diff
@@ -18,6 +18,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// HttpVersion is an alias to enforce validation
+// +kubebuilder:validation:Enum=h2;http/1.1
+type HttpVersion string
```
skriss, line 23: Please use the following casing for the Go type/field:
```suggestion
type HTTPVersion string
```

apis/projectcontour/v1/httpproxy.go
```diff
@@ -18,6 +18,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// HttpVersion is an alias to enforce validation
+// +kubebuilder:validation:Enum=h2;http/1.1
```
skriss, line 22: Since we already have these constants defined in the ContourConfiguration struct (in v1alpha1) and the Contour config file, let's use the same values for them here for consistency:
```suggestion
// +kubebuilder:validation:Enum=HTTP/2;HTTP/1.1
```

Yes, this will then require a translation to the Envoy-accepted values, but we already have code for this in `cmd/contour/parseDefaultHTTPVersions` and `internal/envoy/v3/ProtoNamesForVersions` that we should be able to reuse, possibly with some refactoring required.

apis/projectcontour/v1/httpproxy.go
```diff
@@ -40,6 +44,11 @@ type HTTPProxySpec struct {
 	// is given precedence over this field.
 	// +optional
 	IngressClassName string `json:"ingressClassName,omitempty"`
+
+	// HttpVersions specify the http versions to offer for this HTTPProxy.
+	// If empty, the DefaultHTTPVersions from v1alpha1.EnvoyConfig will be used.
+	// It is ignored when TCPProxy is set.
+	HttpVersions []HttpVersion `json:"httpVersions,omitempty"`
```
skriss, line 51: I think this should move at least under the `VirtualHost` struct, since it is only applicable for root HTTPProxies that define a virtualhost. Still thinking about whether it should move under `VirtualHost.TLS` since it is really only relevant for TLS vhosts AFAIK. cc @sunjayBhatia and @tsaarni for any thoughts here.

internal/xdscache/v3/listener.go
```diff
@@ -502,7 +502,11 @@ func (c *ListenerCache) OnChange(root *dag.DAG) {
 
 				filters = envoy_v3.Filters(cm)
 
-				alpnProtos = envoy_v3.ProtoNamesForVersions(cfg.DefaultHTTPVersions...)
+				if len(vh.HttpVersions) != 0 {
+					alpnProtos = vh.HttpVersions
```
skriss, line 506: In this case where the vhost specifies HTTP versions, I think we can also use the virtual host's HTTP versions for the HTTPConnectionManager's codec, above, since it's specific to the vhost.

internal/dag/httpproxy_processor.go
```diff
@@ -1473,6 +1474,21 @@ func (p *HTTPProxyProcessor) GlobalAuthorizationContext() map[string]string {
 	return nil
 }
 
+// getSortedHttpVersions returns and empty slice or ["h2", "http/1.1"] or ["http/1.1"].
+// This is done to conform with how envoy expects AlpnProtocols in tlsv3.CommonTlsContext.
+func (p *HTTPProxyProcessor) getSortedHttpVersions(proxy *contour_api_v1.HTTPProxy) []string {
```
skriss, line 1479: It'd be nice to reduce the duplication between this new method, `cmd/contour/parseDefaultHTTPVersions` and `internal/envoy/v3/ProtoNamesForVersions`. I'll have to look some more for specific ideas on how to refactor for reuse but if you have ideas in the meantime they are welcome.

skriss (changes requested):
Thanks @therealak12, and apologies for the delayed review here.

In addition to the inline comments:
- I'm not sure why codecov didn't run for the PR, but we'll need to ensure some test coverage for the new code paths. Either a combination of unit tests in `internal/dag` and  `internal/xdscache/v3`, or a "feature test" in `internal/featuretests/v3` should cover it. Let me know if you need more direction here.
- It'd be good to update https://projectcontour.io/resources/faq/ (the question about Safari & 421's) to mention this

apis/projectcontour/v1/httpproxy.go
sunjayBhatia: +1 for moving this under VirtualHost

apis/projectcontour/v1/httpproxy.go
therealak12: The `ProtoNamesForVersions` function converts from `int32` codecs to `["http/1.1", "h2"]` .

The `parseDefaultHTTPVersions` converts from `["HTTP/1.1", "HTTP/2"]`  to `int32` codecs.

Note that the desired format for Envoy configuration is ["http/1.1", "h2"]. So, why use the ["HTTP/1.1", "HTTP/2"] format at all instead of directly using ["http/1.1", "h2"]? 😅 

Additionally, those functions cannot be easily imported since they're private, or importing them leads to import cycles unless we move them into a util-like package.

therealak12: This [comment](https://github.com/projectcontour/contour/pull/5802#discussion_r1442300257) already answers my question. So thanks. I'll refactor.

internal/dag/httpproxy_processor.go
therealak12: Do agree to move `parseDefaultHTTPVersions` and `ProtoNamesForVersions` into a util package?

internal/xdscache/v3/listener.go
```diff
@@ -502,7 +502,11 @@ func (c *ListenerCache) OnChange(root *dag.DAG) {
 
 				filters = envoy_v3.Filters(cm)
 
-				alpnProtos = envoy_v3.ProtoNamesForVersions(cfg.DefaultHTTPVersions...)
+				if len(vh.HTTPVersions) != 0 {
```
sunjayBhatia, line 505: we need to make sure to have unit tests for this precedence behavior, i.e. when http versions are set in the virtualhost, they override the default

internal/xdscache/v3/listener.go
sunjayBhatia: This comment still needs to be addressed

therealak12: Sure. I'll do when there's an agreement on [this](https://github.com/projectcontour/contour/pull/5802#discussion_r1443650964).

lubronzhan:
Kindly ping @therealak12 Could you resolve what Sunjay mentioned. Thank you!

therealak12:
> Kindly ping @therealak12 Could you resolve what Sunjay mentioned. Thank you!

Waiting for this:
https://github.com/projectcontour/contour/pull/5802#discussion_r1443650964

PR #5804, opened 2023-10-05
internal/dag: default Listener ResolvedRefs to true

### LINKED ISSUE #5648: Gateway API: set Listener ResolvedRefs=true if not set to false

skriss:
Sets Gateway Listeners' ResolvedRefs condition
to true by default, to meet conformance.

Closes #5648.

Ran Gateway API main branch conformance manually, confirmed it fails without this change and passes with it.

### DIFF
--- a/test/scripts/run-gateway-conformance.sh
+++ b/test/scripts/run-gateway-conformance.sh
@@ -60,7 +60,7 @@ else
   git checkout "${GATEWAY_API_VERSION}"
   # Keep the list of skipped features in sync with
   # test/conformance/gatewayapi/gateway_conformance_test.go.
-  go test -timeout=40m ./conformance -gateway-class=contour -all-features \
+  go test -timeout=40m ./conformance -run TestConformance -gateway-class=contour -all-features \
     -exempt-features=Mesh \
     -skip-tests=HTTPRouteRedirectPortAndScheme
 fi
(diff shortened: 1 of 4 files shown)


test/scripts/run-gateway-conformance.sh
```diff
@@ -60,7 +60,7 @@ else
   git checkout "${GATEWAY_API_VERSION}"
   # Keep the list of skipped features in sync with
   # test/conformance/gatewayapi/gateway_conformance_test.go.
-  go test -timeout=40m ./conformance -gateway-class=contour -all-features \
+  go test -timeout=40m ./conformance -run TestConformance -gateway-class=contour -all-features \
```
skriss, line 63: added this to avoid running TestExperimentalConformance here.

sunjayBhatia (approved):
looks just like what i did to fix the tests when running them to review 👍🏽

skriss:
> looks just like what i did to fix the tests when running them to review 👍🏽

OK did not realize there was another in-flight Gateway API PR around this, will re-run once more to make sure since that just merged 10min ago

skriss:
OK all good ✅

PR #5813, opened 2023-10-09
Adds `log-path` to the Contour cli

davinci26:
I am debugging a `shutdown` issue and getting logs from `shutdown` manager is generally tricky because `preStop` command logs don't appear in `kubectl logs` or generally anywhere unless there is an error. 

To debug this issue I use this PR in our local PR that adds a `log-path` parameter to contour and redirected the output of the preStop hook to `/proc/1/fd/1`. See this relevant [stackoverflow thread](https://stackoverflow.com/questions/46525317/how-to-monitor-executing-of-prestop-command)

If I found it useful I believe other people might so I pushed it here as well in case you believe it is generally helpful

### DIFF
--- a/cmd/contour/contour.go
+++ b/cmd/contour/contour.go
@@ -45,7 +45,7 @@ func main() {
 
 	// Log-format applies to log format of all sub-commands.
 	logFormat := app.Flag("log-format", "Log output format for Contour. Either text or json.").Default("text").Enum("text", "json")
-
+	logPath := app.Flag("log-path", "Log output path.").Default("").String()
 	bootstrap, bootstrapCtx := registerBootstrap(app)
 
 	certgenApp, certgenConfig := registerCertGen(app)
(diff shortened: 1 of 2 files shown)


cmd/contour/contour.go
```diff
@@ -45,7 +45,7 @@ func main() {
 
 	// Log-format applies to log format of all sub-commands.
 	logFormat := app.Flag("log-format", "Log output format for Contour. Either text or json.").Default("text").Enum("text", "json")
-
+	logPath := app.Flag("log-path", "Log output path.").Default("").String()
```
tsaarni, line 48: The description could mention also that when path is not defined, the logs are printed in stderr.

davinci26: Yeah makes sense, done

davinci26:
@projectcontour/maintainers are you interested in the PR above? I am done debugging the issue I was facing, I used this in our fork to test things out and it was a nice QOL improvement for me. I would recommend we put in a path to merge but if you y'all are not interested I will close it

Issue #5482, opened 2023-06-15
xDS Snapshot versions should be properly generated on startup when used with go-control-plane


sunjayBhatia:
**What steps did you take and what happened:**
The `infra` package e2e tests often fail in our daily build (see example here https://github.com/projectcontour/contour/actions/runs/5269544981/jobs/9527825382) particularly every time (reproed locally) the HTTPS metrics test runs *before* the plain HTTP metrics test.

This is for a few reasons:
- when the HTTPS test runs, the Listener references SDS secrets that Contour serves
- when the HTTP test runs, that Listener is still in Envoy's config so as Envoy reconnects to the Contour instance for the next test, it re-requests the Listener and secrets (with resource version `1`)
- Contour does generate a plain HTTP Listener, however the *version* it generates a snapshot with is `1`, the *same` as the Listener resource version Envoy knows about
- go-control-plane does not respond to the Listener request since the versions match, it will only respond if a snapshot is set with a different version (our own xDS server starts at version `0` so avoids this exact issue, and also still responds to requests unlike how go-control-plane functions)
- (Since SOTW xDS especially does not have a way to *remove* resources, that complicates things. Delta xDS can, which would make this better.)

**What did you expect to happen:**
- `infra` tests consistently pass
- when Contour is restarted, or a new instance starts, any xDS requests that come in for existing resources (from a previous instance of Contour that Envoy was connected to etc.) should be responded to, with the updated/possibly same config generated by this new instance

**Anything else you would like to add:**
Since the `version` in an xDS request/response is just an opaque string, one quick solution (which I quickly tested and works) is to use a uuid or other randomly generated ID that would be unique on each snapshot.

PR #5805, opened 2023-10-05
test/e2e: Add race detection in e2e tests

sunjayBhatia:
Compile contour binary with -race flag and look for "DATA RACE" in stderr. Fails test if found.

PR #5819, opened 2023-10-09
internal/xdscache: Generate uuid for snapshot version

sunjayBhatia:
Snapshotter had a data race reading/writing the snapshot version between threads. This version is not in practice used for the contour xDS server DiscoveryResponse versions but is in the go-control-plane version.

Fixes: #5482


sunjayBhatia:
see https://github.com/projectcontour/contour/pull/5805#issuecomment-1753619118 for data race found

skriss (approved):

PR #5826, opened 2023-10-11
Set stream limits for HTTP2 protocol - CVE CVE-2023-44487

akshaysngupta:
https://github.com/envoyproxy/envoy/pull/30055

1. Set `max_concurrent_stream=100` to limit streams per connection. Helps limit memory usage.
2. Set `http.max_requests_per_io_cycle=1` to limit requests processed in one IO cycle.
3. Set `overload.premature_reset_total_stream_count=50`. This kept low enough to reduce impact.
4. Set `overload.premature_reset_min_stream_lifetime_seconds=1`.

### DIFF
--- a/internal/envoy/v3/listener.go
+++ b/internal/envoy/v3/listener.go
@@ -486,6 +486,12 @@ func (b *httpConnectionManagerBuilder) Get() *envoy_listener_v3.Filter {
 			AcceptHttp_10:      true,
 			AllowChunkedLength: b.allowChunkedLength,
 		},
+		Http2ProtocolOptions: &envoy_core_v3.Http2ProtocolOptions{
+			MaxConcurrentStreams:        wrapperspb.UInt32(100),
+			HpackTableSize:              wrapperspb.UInt32(4096),
+			InitialStreamWindowSize:     wrapperspb.UInt32(65535),
+			InitialConnectionWindowSize: wrapperspb.UInt32(65535),
+		},
 
 		UseRemoteAddress:  wrapperspb.Bool(true),
 		XffNumTrustedHops: b.numTrustedHops,
--- a/internal/envoy/v3/runtime.go
+++ b/internal/envoy/v3/runtime.go
@@ -36,8 +36,11 @@ func RuntimeLayers() []*envoy_service_runtime_v3.Runtime {
 func baseRuntimeLayer() *structpb.Struct {
 	return &structpb.Struct{
 		Fields: map[string]*structpb.Value{
-			"re2.max_program_size.error_level": {Kind: &structpb.Value_NumberValue{NumberValue: maxRegexProgramSizeError}},
-			"re2.max_program_size.warn_level":  {Kind: &structpb.Value_NumberValue{NumberValue: maxRegexProgramSizeWarn}},
+			"re2.max_program_size.error_level":                     structpb.NewNumberValue(maxRegexProgramSizeError),
+			"re2.max_program_size.warn_level":                      structpb.NewNumberValue(maxRegexProgramSizeWarn),
+			"overload.premature_reset_total_stream_count":          structpb.NewNumberValue(50),
+			"overload.premature_reset_min_stream_lifetime_seconds": structpb.NewNumberValue(1),
+			"http.max_requests_per_io_cycle":                       structpb.NewNumberValue(10),
 		},
 	}
 }


internal/envoy/v3/runtime.go
```diff
@@ -36,8 +36,11 @@ func RuntimeLayers() []*envoy_service_runtime_v3.Runtime {
 func baseRuntimeLayer() *structpb.Struct {
 	return &structpb.Struct{
 		Fields: map[string]*structpb.Value{
-			"re2.max_program_size.error_level": {Kind: &structpb.Value_NumberValue{NumberValue: maxRegexProgramSizeError}},
-			"re2.max_program_size.warn_level":  {Kind: &structpb.Value_NumberValue{NumberValue: maxRegexProgramSizeWarn}},
+			"re2.max_program_size.error_level":                     structpb.NewNumberValue(maxRegexProgramSizeError),
+			"re2.max_program_size.warn_level":                      structpb.NewNumberValue(maxRegexProgramSizeWarn),
+			"overload.premature_reset_total_stream_count":          structpb.NewNumberValue(50),
```
sunjayBhatia, line 41: This and the below have defaults that Envoy sets so I'm not sure we need to set these ourselves. 50 also is a lot lower than the Envoy default of 500, can you offer some reasoning behind setting this to 50?

see https://www.envoyproxy.io/docs/envoy/v1.27.1/version_history/v1.27/v1.27.1.html

internal/envoy/v3/runtime.go
```diff
@@ -36,8 +36,11 @@ func RuntimeLayers() []*envoy_service_runtime_v3.Runtime {
...
+			"overload.premature_reset_min_stream_lifetime_seconds": structpb.NewNumberValue(1),
+			"http.max_requests_per_io_cycle":                       structpb.NewNumberValue(10),
```
sunjayBhatia, line 43: this overlaps with https://github.com/projectcontour/contour/pull/5827, I would prefer to make it configurable so we do not impact existing deployments that may be taking advantage of not having a limit here for valid traffic

internal/envoy/v3/listener.go
```diff
@@ -486,6 +486,12 @@ func (b *httpConnectionManagerBuilder) Get() *envoy_listener_v3.Filter {
 			AcceptHttp_10:      true,
 			AllowChunkedLength: b.allowChunkedLength,
 		},
+		Http2ProtocolOptions: &envoy_core_v3.Http2ProtocolOptions{
+			MaxConcurrentStreams:        wrapperspb.UInt32(100),
+			HpackTableSize:              wrapperspb.UInt32(4096),
+			InitialStreamWindowSize:     wrapperspb.UInt32(65535),
```
sunjayBhatia, line 492: Envoy already defaults to this value

internal/envoy/v3/listener.go
```diff
@@ -486,6 +486,12 @@ func (b *httpConnectionManagerBuilder) Get() *envoy_listener_v3.Filter {
...
+			InitialConnectionWindowSize: wrapperspb.UInt32(65535),
```
sunjayBhatia, line 493: Envoy already defaults to this value

internal/envoy/v3/listener.go
sunjayBhatia: see https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-msg-config-core-v3-http2protocoloptions

internal/envoy/v3/listener.go
```diff
@@ -486,6 +486,12 @@ func (b *httpConnectionManagerBuilder) Get() *envoy_listener_v3.Filter {
 			AcceptHttp_10:      true,
 			AllowChunkedLength: b.allowChunkedLength,
 		},
+		Http2ProtocolOptions: &envoy_core_v3.Http2ProtocolOptions{
+			MaxConcurrentStreams:        wrapperspb.UInt32(100),
+			HpackTableSize:              wrapperspb.UInt32(4096),
```
sunjayBhatia, line 491: Envoy already defaults to this value

internal/envoy/v3/listener.go
```diff
@@ -486,6 +486,12 @@ func (b *httpConnectionManagerBuilder) Get() *envoy_listener_v3.Filter {
 			AcceptHttp_10:      true,
 			AllowChunkedLength: b.allowChunkedLength,
 		},
+		Http2ProtocolOptions: &envoy_core_v3.Http2ProtocolOptions{
+			MaxConcurrentStreams:        wrapperspb.UInt32(100),
```
sunjayBhatia, line 490: we may want to make this configurable on the Listener since the limit we have now is the Envoy one, which is the max allowed, users may be relying on this and we don't want to disturb working/valid traffic in an upgrade

sunjayBhatia (commented):
I was working on https://github.com/projectcontour/contour/pull/5827 while this was submitted and looks like there is some overlap

It seems pretty much all of these are set via Envoy defaults already

internal/envoy/v3/runtime.go
akshaysngupta: I tested perf during the attack by setting `premature_reset_total_stream_count` to 50,100 and 500.
Roughly, Envoy served 4 times more RPS for legitimate traffic when value was set to `50` vs `500`.

sunjayBhatia: In that case, I think we can follow up on https://github.com/projectcontour/contour/pull/5827 with a change to make this runtime field configurable, as it seems it is something that needs tuning to a particular environment/usage

PRs welcome for that change

internal/envoy/v3/listener.go
sunjayBhatia: pulled this into https://github.com/projectcontour/contour/issues/5846

internal/envoy/v3/runtime.go
sunjayBhatia: added https://github.com/projectcontour/contour/issues/5847 to cover this

sunjayBhatia:
Hey @akshaysngupta I don't think we will merge this PR as structured, but would rather prefer PRs that solve the targeted issues I just created and linked. I personally would not consider those blockers for a release since there are other mitigations for CVE-2023-44487 in place with defaults and other Contour changes. IMO they are a nice to have that we would appreciate community contributions for.

akshaysngupta:
@sunjayBhatia those other defaults are not enough. Contour-managed Envoy are getting OOM killed within the first 5 seconds of the attack without a limit on the max streams. But it also makes sense to keep this configurable. Sadly, I don't have the bandwidth right now to make the contribution.

What are the options that users have until this setting exposed to mitigate the attack?
Should users build their own custom Contour images to mitigate?

egerkke:
Hi @akshaysngupta ,

We are also investigating how new Contour parameters shall be introduced due to the rapid reset attack. And you mentioned:

> Contour-managed Envoy are getting OOM killed within the first 5 seconds (depends on memory allocated) of the attack without a limit on the max streams.

Did you refer to [max-concurrent-streams](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-field-config-core-v3-http2protocoloptions-max-concurrent-streamsl) of Envoy? If yes then can you please explain a little bit how it is related to the rapid reset attack? My understanding is that this is actually one of the main points of the attack that prematurely closed streams by RSTs are not counted  in the [max-concurrent-streams](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-field-config-core-v3-http2protocoloptions-max-concurrent-streamsl).

Or exposing this parameter in general is useful since its high default value can be exploited easily even without the rapid reset attack?

akshaysngupta:
That's right. Rapid resets don't increase the stream count. This helps with later where attacker creates too many streams.

PR #5827, opened 2023-10-11
Add configurability for HTTP requests per IO cycle

sunjayBhatia:


akshaysngupta:
Can you take reference from this PR: https://github.com/projectcontour/contour/pull/5826 and set other values.

This PR still doesn't safeguard against OOM kills when too many streams are initiated on the connection.

sunjayBhatia:
> Can you take reference from this PR: #5826 and set other values.
> 
> This PR still doesn't safeguard against Envoy OOM kills when too many streams are initiated on the connection.

this PR is not intended to cover all of the configurable fields related to this CVE, this is a narrowly scoped PR by design

pkg/config/parameters.go
```diff
@@ -484,6 +484,13 @@ type ListenerParameters struct {
 
 	// SocketOptions is used to set socket options for listeners.
 	SocketOptions SocketOptions `yaml:"socket-options"`
+
+	// Defines the limit on number of HTTP requests that Envoy will process from a single
+	// connection in a single I/O cycle. Requests over this limit are processed in subsequent
+	// I/O cycles. Can be used as a mitigation for CVE-2023-44487 when abusive traffic is
+	// detected. Configures the http.max_requests_per_io_cycle Envoy runtime setting. The default
+	// value when this is not set is no limit.
+	MaxRequestsPerIOCycle *uint32 `yaml:"max-requests-per-io-cycle"`
```
tsaarni, line 493: Missing this description also in [site/content/docs/main/configuration.md](site/content/docs/main/configuration.md)?

internal/xdscache/v3/runtime.go
```diff
@@ -20,30 +20,46 @@ import (
...
+
+// NewRuntimeCache builds a RuntimeCache with the provided runtime
+// settings that will be set in the runtime layer configured by Contour.
+func NewRuntimeCache(runtimeSettings ConfigurableRuntimeSettings) *runtimeCache {
+	runtimeKV := make(map[string]*structpb.Value)
+	if runtimeSettings.MaxRequestsPerIOCycle != nil && *runtimeSettings.MaxRequestsPerIOCycle > 0 {
+		runtimeKV["http.max_requests_per_io_cycle"] = structpb.NewNumberValue(float64(*runtimeSettings.MaxRequestsPerIOCycle))
+	}
+	return &runtimeCache{runtimeKV: runtimeKV}
 }
```
tsaarni, line 44: Question: would it make sense to move the existing (fixed) `re2.max_program_size.error_level` and `re2.max_program_size.warn_level` runtime parameters from `internal/envoy/v3/runtime.go` and set them to `runtimeKV` here as well? In that case all the runtime keys could be found in single file.

sunjayBhatia: At the moment we also use this helper from `internal/envoy/v3` to build up the bootstrap config base RTDS layer: https://github.com/projectcontour/contour/blob/f52ad069b1090aaabf8e05866af4ef4e1e4bdb31/internal/envoy/v3/bootstrap.go#L169

We may be able to get rid of that static bootstrap runtime layer since it was introduced in https://github.com/projectcontour/contour/pull/4379 and just serve dynamic layers instead (added in https://github.com/projectcontour/contour/pull/4387) which would mean we can refactor and consolidate a bit.

I think I would prefer to do that refactor in a separate commit/PR though

tsaarni: > At the moment we also use this helper from internal/envoy/v3 to build up the bootstrap config

Ah right, that makes sense. It seemed unnecessarily complicated, when not realizing that :smile:

pkg/config/parameters.go
sunjayBhatia: yep good catch 👍🏽

internal/xdscache/v3/runtime.go
sunjayBhatia: added https://github.com/projectcontour/contour/issues/5845 as a cleanup issue

tsaarni (approved):
Looks good to me! :+1:

PR #5828, opened 2023-10-11
allow configuration of upstream TLS connection settings

### LINKED ISSUE #3574: Should globally configurable TLS options should be set on upstream TLS contexts?

### LINKED ISSUE #5501: Support of minimumProtocolVersion when Envoy acts as client

KauzClay:
Just as you can configure TLS settings for downstream connections, like setting min/max TLS versions and cipher suites, you should be able to do that for upstream connections too.

For upstream connections, I would like the default Max TLS version to be 1.3, instead of the [current Envoy default of 1.2](https://github.com/envoyproxy/go-control-plane/blob/b65248991f33ae59b75bcc80a589c79e9735fe03/envoy/extensions/transport_sockets/tls/v3/common.pb.go#L213-L214)

release-note/minor

Fixes #5501 
Fixes #3574


tsaarni:
Hi @KauzClay 👋 Linking some previous discussions related to upstream TLS just for reference info: #5501, #5666

KauzClay:
@sunjayBhatia , it is still pretty rough, but I've attempted to wire everything up for this new param.

One question I had though, was how do I enforce defaults for min and max TLS version? I would like the Contour default for max version to be different than Envoy's default.

changelogs/unreleased/5828-KauzClay-minor.md
```diff
@@ -0,0 +1,6 @@
+## Allow Configuration of Upstream TLS Options
+
+In a similar way to how Contour users can configure Min/Max TLS version and
+Cipher Suites for Envoy's listeners, this change allows users to specify the
+same information for upstream connections. This change also defaults the Max TLS
+version for upstream connections to 1.3, instead of the Envoy default of 1.2.
```
tsaarni, line 6: It would be nice to mention here how user can achieve this, that is, refer to the ContourConfiguration and config parameter names.

tsaarni (commented):
Added some comments inline, but looks good to me in general!

I think it would be good to have some reasoning here about the corner cases that have led Envoy not to enable TLSv1.3 for upstream, and why it is OK for Contour to do so - something that answers to https://github.com/projectcontour/contour/pull/5666#pullrequestreview-1570779029 

One more thought: It would be nice to have e2e test for the protocol version.  The `echoserver` used for e2e suite returns a JSON response that includes the negotiated TLS version (see [here](https://github.com/kubernetes-sigs/ingress-controller-conformance/blob/d920ed36a0076e169a9a329a850844ab3a695ae8/images/echoserver/echoserver.go#L231C6-L246)). We could check that TLSv1.3 gets returned. Check [`test/e2e/httpproxy/backend_tls_test.go`](https://github.com/projectcontour/contour/blob/main/test/e2e/httpproxy/backend_tls_test.go) on similar assertion checks done for the response. In `responseTLSDetails` the client certificate was parsed from response document and compared to what we had configured.

pkg/config/parameters.go
```diff
@@ -429,6 +429,11 @@ type ClusterParameters struct {
 	//
 	// +optional
 	PerConnectionBufferLimitBytes *uint32 `yaml:"per-connection-buffer-limit-bytes,omitempty"`
+
+	// UpstreamTLS contains the TLS policy parameters for upstream connections
+	//
+	// +optional
+	UpstreamTLS TLSParameters `yaml:"upstream-tls,omitempty"`
 }
```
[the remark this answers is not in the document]
sunjayBhatia: can also get rid of the `+optional` tag here, thats just a kubebuilder thing (and we should remove it from all these fields but can do that separately for other fields)

test/e2e/httpproxy/httpproxy_test.go
```diff
@@ -193,93 +193,109 @@ var _ = Describe("HTTPProxy", func() {
...
+		})
+		Context("with backend tls", func() {
 			testBackendTLS(namespace)
 		})
+		Context("with backend tls version configured via Contour ConfigMap", func() {
+			protocolVersion := "TLSv1.3"
+
+			Context("via Contour ConfigMap", func() {
+				BeforeEach(func() {
+					contourConfig.Cluster.UpstreamTLS.MinimumProtocolVersion = protocolVersion
```
sunjayBhatia, line 287: seems like this should this be `1.3` rather than `TLSv1.3` as above? (though echoserver might return `TLSv1.3`, i don't remember exactly)

test/e2e/httpproxy/httpproxy_test.go
```diff
@@ -193,93 +193,109 @@ var _ = Describe("HTTPProxy", func() {
 	})
 
 	f.NamespacedTest("httpproxy-backend-tls", func(namespace string) {
```
sunjayBhatia, line 195: hm, we might need to separate out the tests into different `NamespacedTest` blocks to make sure no test pollution

if we see flakiness in this area (creating/deleting resources etc.) this would be something to look into

test/e2e/httpproxy/httpproxy_test.go
KauzClay: yeah, I had a feeling that might not align.

Based on [this block](https://github.com/kubernetes-sigs/ingress-controller-conformance/blob/d920ed36a0076e169a9a329a850844ab3a695ae8/images/echoserver/echoserver.go#L231C6-L246) that @tsaarni shared with me, I'm pretty sure echoserver returns "TLSv1.3".

I don't think that would also be acceptable for configuring Contour though

KauzClay: I just wanted to commit something and see all the tests go (my machine suffers when I try to run tests locally lol)

internal/envoy/v3/cluster.go
```diff
@@ -177,6 +179,7 @@ func ExtensionCluster(ext *dag.ExtensionCluster) *envoy_cluster_v3.Cluster {
 				ext.UpstreamValidation,
 				ext.SNI,
 				ext.ClientCertificate,
+				ext.UpstreamTLS,
```
tsaarni, line 182: Is this set anywhere? I believe it should be set here https://github.com/KauzClay/contour/blob/5233293fc71feaf8461e619d6ca12257f46c78de/internal/dag/extension_processor.go#L103-L117

internal/envoy/v3/cluster.go
```diff
@@ -207,7 +210,7 @@ func DNSNameCluster(c *dag.DNSNameCluster) *envoy_cluster_v3.Cluster {
 
 	var transportSocket *envoy_core_v3.TransportSocket
 	if c.Scheme == "https" {
-		transportSocket = UpstreamTLSTransportSocket(UpstreamTLSContext(c.UpstreamValidation, c.Address, nil))
+		transportSocket = UpstreamTLSTransportSocket(UpstreamTLSContext(c.UpstreamValidation, c.Address, nil, c.UpstreamTLS))
```
tsaarni, line 213: Is `c.UpstreamTLS` unset? I believe it should be set here https://github.com/KauzClay/contour/blob/5233293fc71feaf8461e619d6ca12257f46c78de/internal/dag/httpproxy_processor.go#L490-L495

pkg/config/parameters.go
```diff
@@ -429,6 +429,9 @@ type ClusterParameters struct {
 	//
 	// +optional
 	PerConnectionBufferLimitBytes *uint32 `yaml:"per-connection-buffer-limit-bytes,omitempty"`
+
+	// UpstreamTLS contains the TLS policy parameters for upstream connections
+	UpstreamTLS TLSParameters `yaml:"upstream-tls,omitempty"`
```
tsaarni, line 434: Should we avoid reusing the `TLSParameters` struct here since it contains some parameters that can be meaningful only for downstream or upstream TSL? 

For example field `TLSParameters.FallbackCertificate` is used as server certificate, which appears as configuration file field  `tls.fallback-certificate`. When reusing the struct for upstream, it could be also defined as `cluster.upstream-tls.fallback-certificate` according purely to syntax, but if defined there, it would be ignored. 

Similarly, config file field `tls.envoy-client-certificate` is used as optional client cert in upstream TLS. Reusing the struct would mean field `cluster.upstream-tls.envoy-client-certificate` is syntactically correct, but actually it will not get used.

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -656,6 +656,11 @@ type ClusterParameters struct {
 	// +kubebuilder:validation:Minimum=1
 	// +optional
 	PerConnectionBufferLimitBytes *uint32 `json:"per-connection-buffer-limit-bytes,omitempty"`
+
+	// UpstreamTLS contains the TLS policy parameters for upstream connections
+	//
+	// +optional
+	UpstreamTLS *EnvoyTLS `json:"upstream-tls,omitempty"`
```
tsaarni, line 663: Question: What do people prefer: nesting TLS config in `ContourConfiguration.spec.envoy.cluster.upstream-tls` (like suggested here) or bumping it one level `ContourConfiguration.spec.envoy.upstream-tls`? 

Maybe bit unfortunate but there is existing upstream TLS parameter directly under `ContourConfiguration.spec.envoy`, which is `clientCertificate`. It's bit unfortunate that the upstream TLS settings will be split in two places, but hopefully not too far from each other.

The corresponding downstream TLS setting is `ContourConfiguration.spec.envoy.listener.tls`. I guess perfect symmetry across TLS settings is not achievable anymore, not even in CRD config which was the fresh start :)

skriss (changes requested):

KauzClay:
hey @skriss regarding:
>Given dowstream TLS settings are nested under envoy.listener, I think putting upstream settings under envoy.cluster is logical.
>
> Re: clientCertificate -- the CRD is still v1alpha1, so we could make a breaking change if we wanted to reorganize. Seems like moving clientCertificate into UpstreamTLS might make sense?

just want to make sure I'm following. I think you're saying that what I have with envoy.cluster.upstreamTLS is okay.

And then the second part is to move envoy.ClientCertificate into envoy.cluster.upstreamTLS?

apis/projectcontour/v1alpha1/contourconfig.go
KauzClay: I could move clientCertificate into UpstreamTLS. OR would you prefer that be in a separate MR?

internal/envoy/v3/cluster.go
KauzClay: thank you for pointing this out, I had a feeling some things were missing, but I got a bit lost in the code.

pkg/config/parameters.go
KauzClay: hmmm I see. I went with TLSParameters originally because it had what I needed and I didn't want to create a similar-but-not-quite-the-same struct.

What would you say to creating a new struct that just contains `{MinimumProtocolVersion,MaximumProtocolVersion,CipherSuites}`, and then making that struct a part of TLSParameters?

apis/projectcontour/v1alpha1/contourconfig.go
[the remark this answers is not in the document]
skriss: yeah, it's fine in `pkg/config`, we have some inconsistency there but `upstream-tls` matches the most common style.

skriss: Let's hold off for now to see what others think about that change, we can do it in a separate PR if we decide to proceed.

sunjayBhatia: +1 for keeping it under cluster

test/e2e/httpproxy/httpproxy_test.go
```diff
@@ -282,6 +282,116 @@ var _ = Describe("HTTPProxy", func() {
...
+					Name:      "backend-client-cert",
+				},
+			}
+
+			contourConfiguration.Spec.Envoy.ClientCertificate = &contour_api_v1alpha1.NamespacedName{
+				Name:      "backend-client-cert",
+				Namespace: namespace,
+			}
+		})
+		Context("with backend tls version configured via Contour ConfigMap", func() {
```
sunjayBhatia, line 369: I think we can combine these two contexts (configmap and contourconfig-specific wrappers) and just set the same options in the BeforeEach just as we do here: https://github.com/projectcontour/contour/blob/4d48db7a095888cdec109bc4f90f87a65013082b/test/e2e/httpproxy/httpproxy_test.go#L115-L118

we have different test runs for using each type of config so we should only need one instance of the test

sunjayBhatia (commented):
Functionalityl ooks good and change overall looks good modulo whatever existing nits there are and questions about API types

Once those are resolved I'm good to go with this change 👍🏽

changelogs/unreleased/5828-KauzClay-minor.md
```diff
@@ -0,0 +1,8 @@
+## Allow Configuration of Upstream TLS Options
+
+In a similar way to how Contour users can configure Min/Max TLS version and
+Cipher Suites for Envoy's listeners, this change allows users to specify the
+same information for upstream connections. In the ContourConfiguration, this is
+available under `spec.Envoy.Cluster.UpstreamTLS`. The equivalent config
```
skriss, line 6:
```suggestion
available under `spec.envoy.cluster.upstreamTLS`. The equivalent config file
```

internal/dag/extension_processor.go
```diff
@@ -114,6 +118,7 @@ func (p *ExtensionServiceProcessor) buildExtensionService(
 		ClusterTimeoutPolicy: ctp,
 		SNI:                  "",
 		ClientCertificate:    clientCertSecret,
+		UpstreamTLS:          (*UpstreamTLS)(p.UpstreamTLS),
```
skriss, line 121: Huh, I did not realize this was valid Go. Looks like since the API struct and the DAG struct have exactly the same fields, casting is allowed.

internal/dag/httpproxy_processor.go
```diff
@@ -1026,6 +1031,7 @@ func (p *HTTPProxyProcessor) computeRoutes(
 				SlowStartConfig:               slowStart,
 				MaxRequestsPerConnection:      p.MaxRequestsPerConnection,
 				PerConnectionBufferLimitBytes: p.PerConnectionBufferLimitBytes,
+				UpstreamTLS:                   determineUpstreamTLS(p.UpstreamTLS),
```
skriss, line 1034: I _think_ you can use the same casting you're using in some of the other processors and drop the helper function entirely, though functionally it's just fine as-is.

pkg/config/parameters.go
skriss: :+1: that sounds good to me

skriss (changes requested):

pkg/config/parameters.go
KauzClay: Okay, I ended up creating a new smaller struct called ProtocolParameters. I can always change it if you think a different name is better

test/e2e/httpproxy/httpproxy_test.go
KauzClay: okay I combined them into 1 before each. Does it look how you're imagining now?

pkg/config/parameters.go
```diff
@@ -182,8 +182,7 @@ func (n NamespacedName) Validate() error {
 
 // TLSParameters holds configuration file TLS configuration details.
 type TLSParameters struct {
-	MinimumProtocolVersion string `yaml:"minimum-protocol-version"`
-	MaximumProtocolVersion string `yaml:"maximum-protocol-version"`
+	ProtocolParameters ProtocolParameters `yaml:"protocol"`
```
skriss, line 185: We'll need to embed the new struct here so we don't make a breaking change to the config file YAML representation:
```suggestion
	ProtocolParameters
```

skriss (changes requested):

pkg/config/parameters.go
KauzClay: hmmm I see. I almost did that too lol but went for adding a new field instead

skriss:
I'll take another pass today but I think this is looking pretty good

internal/featuretests/v3/upstreamprotocol_test.go
skriss, line 1: It'd be ideal to add some test coverage here for when upstream TLS settings are specified as well, since these are much faster to run than the E2E's. See e.g. https://github.com/projectcontour/contour/blob/4d48db7a095888cdec109bc4f90f87a65013082b/internal/featuretests/v3/fallbackcert_test.go#L34-L47 for an example of how to set some non-default settings on the various processors, then the feature test case(s) can ensure that those flow through to the created xDS clusters.

As a general note, typically we try to include either (a) internal/dag/builder_test.go and internal/dag/status_test.go tests, or (b) internal/featuretests/v3 tests for new features like this. This would be in addition to E2E's, for optimal coverage.

skriss (changes requested):
Sorry, a couple last things, but I think that's it from me.

internal/featuretests/v3/upstreamprotocol_test.go
KauzClay: thanks for explaining all this, there's so much to change!

Anyway, I added a new test, I think it is pretty simple. I have one for httpproxy, and for an extension service.

Would you also want others?

I was thinking of gatewayAPI, but you'd need a tlsRoute to do this, right? I'm under the impression tlsRoute isn't quitet there yet upstream?

skriss: It'd be great to have one for Ingress as well, but I won't block the PR on it since it's lower priority

cmd/contour/serve.go
```diff
@@ -1194,6 +1200,7 @@ func (s *Server) getDAGBuilder(dbc dagBuilderConfig) *dag.Builder {
 			FieldLogger:       s.log.WithField("context", "ExtensionServiceProcessor"),
 			ClientCertificate: dbc.clientCert,
 			ConnectTimeout:    dbc.connectTimeout,
+			UpstreamTLS:       dbc.upstreamTLS,
```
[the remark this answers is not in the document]
skriss: Sorry to backtrack on this, but I'm thinking we can actually back out the Gateway API change for now, and tackle it when we implement https://github.com/projectcontour/contour/issues/5929. I _believe_  we'll want these to apply for HTTPRoutes, but honestly I need to re-read the spec to make sure, so OK to just skip for now so we can get this merged.

skriss (approved):
This LGTM, will leave for another maintainer to take another pass, thanks @KauzClay!

internal/featuretests/v3/upstreamprotocol_test.go
KauzClay: @skriss  perhaps this is my lack of understanding showing, but how would you do it for ingress?

You can't indicate that you wanna do upstreamTLS like you can on an HTTPProxy, right?

Is the presence of the backend service port being 443 enough to make envoy do TLS to it?

skriss: You can annotate the target Service with `projectcontour.io/upstream-protocol.tls: 443` which would then apply for an Ingress (or HTTPProxy) that references it, ref. https://projectcontour.io/docs/1.27/config/annotations/#contour-specific-service-annotations and https://projectcontour.io/docs/1.27/config/upstream-tls/

KauzClay: ah okay, thanks!

tsaarni (changes requested):
Looks good to me too! Just one more small request. If I'm not mistaken, I think we can avoid the typecasts to `dag.UpstreamTLS`

cmd/contour/serve.go
```diff
@@ -560,6 +560,11 @@ func (s *Server) doServe() error {
 		globalRateLimitService:             contourConfiguration.RateLimitService,
 		maxRequestsPerConnection:           contourConfiguration.Envoy.Cluster.MaxRequestsPerConnection,
 		perConnectionBufferLimitBytes:      contourConfiguration.Envoy.Cluster.PerConnectionBufferLimitBytes,
+		upstreamTLS: &contour_api_v1alpha1.EnvoyTLS{
+			MinimumProtocolVersion: annotation.TLSVersion(contourConfiguration.Envoy.Cluster.UpstreamTLS.MinimumProtocolVersion, "1.2"),
+			MaximumProtocolVersion: annotation.TLSVersion(contourConfiguration.Envoy.Cluster.UpstreamTLS.MaximumProtocolVersion, "1.3"),
+			CipherSuites:           contourConfiguration.Envoy.Cluster.UpstreamTLS.SanitizedCipherSuites(),
+		},
```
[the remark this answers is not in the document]
KauzClay: oh okay, yeah I can do that.

I originally went with the envoyTLS object because it seemed like other things in the dagBuilderConfig used objects from `contour_api_v1alpha1`, but this seems nicer

skriss:
Looks like we're good to merge here, thanks @KauzClay! Will merge later today.

tsaarni (approved):

PR #5840, opened 2023-10-12
httpproxy: fix the order for filter: AuthZ & AuthN

### LINKED ISSUE #4826: (Re)consider the order of JWT verification vs external authorization filters
When both JWT verification and external authorization are enabled on the same route, it seems that external authorization is executed before JWT verification https://github.com/projectcontour/contour/blob/01164fe9f75f471361de08cd758d5ea372ba6020/internal/xdscache/v3/listener.go#L438-L439

Maybe combining both was not explicitly considered during the introduction of the feature, but it could be useful combination: validate user with JWT authentication and then authorize the request with external authorization. In that case, wouldn't it be more useful if the order of filters was the other way around? Authorization decision should not be requested unless request was successfully authenticated.

This issue that was brought up by @derekperkins in a discussion on Slack https://kubernetes.slack.com/archives/C8XRH2R4J/p1666799473264549.

izturn:
put the filter: Authz after AuthN
fix #4826 


tsaarni:
Hi @izturn! :wave:  You could add `closes #4826` to the description.

izturn:
@tsaarni sorry, i don't see it, i find it when do ext_proc, ;-)

tsaarni:
@izturn please add changelog entry. This is change to external behavior and it is good to notify users in release notes.

I'm in favor of this PR as discussed previously in #4826.

izturn:
@tsaarni PTAL

changelogs/unreleased/5840-izturn-small.md
```diff
@@ -0,0 +1 @@
+Rearrange the order of filters, put the external authorization filter after the JWT verification filter.
```
skriss, line 1: Let's bump this up to minor, and re-word a bit for clarity for users:

```suggestion
## JWT Authentication happens before External Authorization

Fixes a bug where when the external authorization filter and JWT authentication filter were both configured, the external authorization filter was executed _before_ the JWT authentication filter.  Now, JWT authentication happens before external authorization when they are both configured.
```

internal/xdscache/v3/listener.go
```diff
@@ -481,8 +481,8 @@ func (c *ListenerCache) OnChange(root *dag.DAG) {
 					Codec(envoy_v3.CodecForVersions(cfg.DefaultHTTPVersions...)).
 					AddFilter(envoy_v3.FilterMisdirectedRequests(vh.VirtualHost.Name)).
 					DefaultFilters().
-					AddFilter(authFilter).
-					AddFilter(envoy_v3.FilterJWTAuth(vh.JWTProviders)).
+					AddFilter(envoy_v3.FilterJWTAuthN(vh.JWTProviders)).
```
skriss, line 484: Ideally we can add a test case for this. I won't block the PR on it, but it would be very nice to have.

skriss (changes requested):

izturn:
@sunjayBhatia  @skriss  PTAL

skriss (approved):
LGTM, will leave for another maintainer to take a second look as well, thanks @izturn

tsaarni (approved):

Issue #5846, opened 2023-10-12
Make Listener MaxConcurrentStreams for HTTP/2 configurable


sunjayBhatia:
We should make the max concurrent streams the Envoy Listeners expose for HTTP/2 configurable so that users can tune this value. This can help with performance, memory usage, etc.

We shouldn't change the default from what is there (the Envoy basically unlimited default) so we do not impact existing deployments/traffic.

              we may want to make this configurable on the Listener since the limit we have now is the Envoy one, which is the max allowed, users may be relying on this and we don't want to disturb working/valid traffic in an upgrade

_Originally posted by @sunjayBhatia in https://github.com/projectcontour/contour/pull/5826#discussion_r1355751569_

PR #5850, opened 2023-10-13
HTTP/2 max concurrent streams can be configured

sunjayBhatia:
Adds a global Listener configuration field for admins to be able to protect their installations of Contour/Envoy with a limit. Default is no limit to ensure existing behavior is not impacted for valid traffic. This field can be used for tuning resource usage or mitigating DOS attacks like in CVE-2023-44487.

Fixes: #5846


akshaysngupta:
thanks! lgtm

tsaarni (approved):
Looks good to me! 👍

PR #5851, opened 2023-10-13
release-1.26: Backport #5827

sunjayBhatia:


pkg/config/parameters.go
```diff
@@ -484,6 +484,13 @@ type ListenerParameters struct {
 
 	// SocketOptions is used to set socket options for listeners.
 	SocketOptions SocketOptions `yaml:"socket-options"`
+
+	// Defines the limit on number of HTTP requests that Envoy will process from a single
+	// connection in a single I/O cycle. Requests over this limit are processed in subsequent
+	// I/O cycles. Can be used as a mitigation for CVE-2023-44487 when abusive traffic is
+	// detected. Configures the http.max_requests_per_io_cycle Envoy runtime setting. The default
+	// value when this is not set is no limit.
+	MaxRequestsPerIOCycle *uint32 `yaml:"max-requests-per-io-cycle"`
```
sunjayBhatia, line 493: this ends up being a new configuration field available if we do backport this change

worth a discussion whether this is desirable or we should only include this new mitigation/feature in the latest release

tsaarni: Envoy also ended up introducing this new configuration in backports, which is maybe bit unusual, but I do not have strong opinion. It sounds ok for me. 

That said, the Envoy version needs to get bumped. I guess to each respective release tracks according to https://projectcontour.io/resources/compatibility-matrix/ vs https://github.com/envoyproxy/envoy/security/advisories/GHSA-jhv4-f7mr-xx76.

tsaarni (approved):

PR #5852, opened 2023-10-13
release-1.25: Backport #5827 and #5850

sunjayBhatia:
Commit messages:

---

HTTP/2 max concurrent streams can be configured (#5850)

Adds a global Listener configuration field for admins to be able to
protect their installations of Contour/Envoy with a limit. Default is no
limit to ensure existing behavior is not impacted for valid traffic.
This field can be used for tuning resource usage or mitigated DOS
attacks like in CVE-2023-44487.


tsaarni (commented):
CI in 1.25 and older is blocked by lint error

`unexported-return: exported func NewRuntimeCache returns unexported type *v3.runtimeCache, which can be annoying to use (revive)`

Strangely the only difference in `.golangci.yml` for `revive` seems unrelated (`use-any`)? Could be just version difference as well, though seems unlikely (golangci-lint v1.51.1 vs v1.51.2)

sunjayBhatia:
> CI in `release-1.25` and older is blocked by lint error
> 
> ```
> unexported-return: exported func NewRuntimeCache returns unexported type *v3.runtimeCache, which can be annoying to use (revive)
> ```
> 
> Strangely the only difference in `.golangci.yml` for `revive` seems unrelated (`use-any`)? Could be just version difference as well, though seems unlikely (golangci-lint v1.51.1 vs v1.51.2)

Looks like the issue might be because we've enabled *only* a specific linter in 1.26+ so only the `use-any` rule is running for the `revive` linter, see https://revive.run/docs#golangci-lint

tsaarni (approved):

PR #5854, opened 2023-10-14
Improve tagging & deployment for install-contour/provisioner-working scripts

### LINKED ISSUE #5277: improve tagging & deployment for install-contour/provisioner-working
Come up with a tagging and deployment scheme for install-contour-working and install-provisioner-working that enables rapid re-deployment of updated code for both the static Contour and the provisioner use cases, and also doesn't fill up the dev environment with randomly-tagged images.

harshil1973:
closes #5277

### DIFF
--- a/test/scripts/install-contour-working.sh
+++ b/test/scripts/install-contour-working.sh
@@ -44,15 +44,13 @@ if ! kind::cluster::exists "$CLUSTERNAME" ; then
     exit 2
 fi
 
-# Set (pseudo) random image tag to trigger restarts at every deployment.
-# TODO: Come up with a scheme that doesn't fill up the dev environment with randomly-tagged images.
-VERSION="v$$"
+docker rmi ghcr.io/projectcontour/contour:dev --force
 
 # Build the image.
-make -C ${REPO} container IMAGE=ghcr.io/projectcontour/contour VERSION=${VERSION}
+make -C ${REPO} container IMAGE=ghcr.io/projectcontour/contour VERSION=dev
 
 # Push the Contour build image into the cluster.
-kind::cluster::load::docker ghcr.io/projectcontour/contour:${VERSION}
+kind::cluster::load::docker ghcr.io/projectcontour/contour:dev
 
 # Install Contour
 ${KUBECTL} apply -f ${REPO}/examples/contour/00-common.yaml
@@ -71,10 +69,13 @@ for file in ${REPO}/examples/contour/02-job-certgen.yaml ${REPO}/examples/contou
   # we built above.
   sed \
     "-es|imagePullPolicy: Always|imagePullPolicy: IfNotPresent|" \
-    "-es|image: ghcr.io/projectcontour/contour:.*$|image: ghcr.io/projectcontour/contour:${VERSION}|" \
+    "-es|image: ghcr.io/projectcontour/contour:.*$|image: ghcr.io/projectcontour/contour:dev|" \
     "$file" | \
   ${KUBECTL} apply -f -
 done
+
+# Patching the deployment with timestamp will trigger pod restart.
+${KUBECTL} patch deployment -n projectcontour contour --patch '{"spec": {"template": {"metadata": {"annotations": {"timestamp": "'$(date +%s)'"}}}}}'
 
 # Wait for Contour and Envoy to report "Ready" status.
 ${KUBECTL} wait --timeout="${WAITTIME}" -n projectcontour -l app=contour deployments --for=condition=Available
--- a/test/scripts/install-provisioner-working.sh
+++ b/test/scripts/install-provisioner-working.sh
@@ -49,6 +49,11 @@ ${KUBECTL} apply -f <(cat examples/gateway-provisioner/03-gateway-provisioner.ya
     yq eval '.spec.template.spec.containers[0].imagePullPolicy = "IfNotPresent"' - | \
     yq eval '.spec.template.spec.containers[0].args += "--contour-image="+env(CONTOUR_IMG)' -)
 
+# Patching the deployment with timestamp will trigger pod restart.
+for deployment in $(kubectl get deployment -n projectcontour -l app.kubernetes.io/managed-by=contour-gateway-provisioner -o name) ; do
+  ${KUBECTL} patch $deployment -n projectcontour --patch '{"spec": {"template": {"metadata": {"annotations": {"timestamp": "'$(date +%s)'"}}}}}'
+done
+
 # Wait for the provisioner to report "Ready" status.
 ${KUBECTL} wait --timeout="${WAITTIME}" -n projectcontour -l control-plane=contour-gateway-provisioner deployments --for=condition=Available
 ${KUBECTL} wait --timeout="${WAITTIME}" -n projectcontour -l control-plane=contour-gateway-provisioner pods --for=condition=Ready


test/scripts/install-contour-working.sh
```diff
@@ -71,11 +69,13 @@ for file in ${REPO}/examples/contour/02-job-certgen.yaml ${REPO}/examples/contou
   # we built above.
   sed \
     "-es|imagePullPolicy: Always|imagePullPolicy: IfNotPresent|" \
-    "-es|image: ghcr.io/projectcontour/contour:.*$|image: ghcr.io/projectcontour/contour:${VERSION}|" \
+    "-es|image: ghcr.io/projectcontour/contour:.*$|image: ghcr.io/projectcontour/contour:latest|" \
     "$file" | \
   ${KUBECTL} apply -f -
 done
 
+${KUBECTL} patch deployment -n projectcontour contour --patch '{"spec": {"template": {"metadata": {"annotations": {"timestamp": "'$(date +%s)'"}}}}}'
```
tsaarni, line 77: There could be small comment to explain the purpose: patching the deployment with timestamp will trigger pod restart.

test/scripts/install-provisioner-working.sh
```diff
@@ -49,6 +47,8 @@ ${KUBECTL} apply -f <(cat examples/gateway-provisioner/03-gateway-provisioner.ya
     yq eval '.spec.template.spec.containers[0].imagePullPolicy = "IfNotPresent"' - | \
     yq eval '.spec.template.spec.containers[0].args += "--contour-image="+env(CONTOUR_IMG)' -)
 
+${KUBECTL} patch deployment -n projectcontour contour-gateway-provisioner --patch '{"spec": {"template": {"metadata": {"annotations": {"timestamp": "'$(date +%s)'"}}}}}'
```
tsaarni, line 50: There could be small comment to explain the purpose: patching the deployment with timestamp will trigger pod restart.

tsaarni (changes requested):
Thank you @harshil1973!

Looks good to me. Small remark from me inline.

The side effect of patching with timestamp label is that also at the initial `make install-*` (when cluster is being created) the `kubectl patch` will trigger an extra restart that would not be necessary. But at least to me, this seems small "price" to pay for not spamming the developer's disk with temp images.

test/scripts/install-contour-working.sh
```diff
@@ -44,15 +44,13 @@ if ! kind::cluster::exists "$CLUSTERNAME" ; then
     exit 2
 fi
 
-# Set (pseudo) random image tag to trigger restarts at every deployment.
-# TODO: Come up with a scheme that doesn't fill up the dev environment with randomly-tagged images.
-VERSION="v$$"
+docker rmi ghcr.io/projectcontour/contour:latest --force
```
skriss, line 47: Since `latest` is a [real tag that we publish](https://github.com/projectcontour/contour/pkgs/container/contour/138343235?tag=latest), I'd suggest using something like `dev` or `working` instead.

test/scripts/install-provisioner-working.sh
```diff
@@ -28,16 +28,16 @@ if ! kind::cluster::exists "$CLUSTERNAME" ; then
...
-make -C ${REPO} container IMAGE=ghcr.io/projectcontour/contour VERSION=latest
+make -C ${REPO} container IMAGE=ghcr.io/projectcontour/contour VERSION=dev
 
 # Push the Contour Provisioner image into the cluster.
-kind::cluster::load::docker ghcr.io/projectcontour/contour:latest
+kind::cluster::load::docker ghcr.io/projectcontour/contour:dev
 
 # Install the Gateway provisioner using the loaded image.
-export CONTOUR_IMG=ghcr.io/projectcontour/contour:latest
+export CONTOUR_IMG=ghcr.io/projectcontour/contour:dev
```
skriss, line 40: I think there's an issue here where even though the provisioner itself will be updated to use the newly built `-dev` image, any Contour instances managed by it don't get updated, since the tag hasn't changed, which can be problematic for the dev workflow.

harshil1973: In that case, maybe an old pseudo-random image tag schema would be nice with some changes like we can add the prefix `dev` to it so that we can remove all the images starting with the `dev` tag.

skriss:
One path forward here could be to just make this change for the non-provisioner use case, which will at least solve some of the problem. We can continue to work on the provisioner scenario separately.

harshil1973:
@skriss Done 👍

tsaarni:
If removing provisioner-working from the scope then I think `test/scripts/install-provisioner-working.sh` should not patch the timestamp at all (like it currently still does), but instead we'd just relay on random PID based image tag. Otherwise it will be restarted twice :smile: 

But I wonder, isn't it possible to follow your original idea for managed contour's as well by something like:

```diff
diff --git a/test/scripts/install-provisioner-working.sh b/test/scripts/install-provisioner-working.sh
@@ -51,6 +52,9 @@ ${KUBECTL} apply -f <(cat examples/gateway-provisioner/03-gateway-provisioner.ya

 # Patching the deployment with timestamp will trigger pod restart.
 ${KUBECTL} patch deployment -n projectcontour contour-gateway-provisioner --patch '{"spec": {"template": {"metadata": {"annotations": {"timestamp": "'$(date +%s)'"}}}}}'
+for deployment in $(kubectl get deployment -n projectcontour -l app.kubernetes.io/managed-by=contour-gateway-provisioner -o name) ; do
+    ${KUBECTL} patch $deployment -n projectcontour --patch '{"spec": {"template": {"metadata": {"annotations": {"timestamp": "'$(date +%s)'"}}}}}'
+done

 # Wait for the provisioner to report "Ready" status.
 ${KUBECTL} wait --timeout="${WAITTIME}" -n projectcontour -l control-plane=contour-gateway-provisioner deployments --for=condition=Available
```

If there are not any (Gateway instance not created yet) the loop would be just skipped.

tsaarni:
There was small confusion, the loop that starts managed contours should have applied to `install-provisioner-working.sh` which is for testing contour provisioner & managed contours.  

On the other hand `install-contour-working.sh` which has just single (unmanaged) contour, what you had previously was correct.

I'll include diff here on top of the current PR that shows what I meant:

```diff
diff --git a/test/scripts/install-contour-working.sh b/test/scripts/install-contour-working.sh
index f5ce10d6..88e595bc 100755
--- a/test/scripts/install-contour-working.sh
+++ b/test/scripts/install-contour-working.sh
@@ -75,9 +75,7 @@ for file in ${REPO}/examples/contour/02-job-certgen.yaml ${REPO}/examples/contou
 done

 # Patching the deployment with timestamp will trigger pod restart.
-for deployment in $(kubectl get deployment -n projectcontour -l app.kubernetes.io/managed-by=contour-gateway-provisioner -o name) ; do
-  ${KUBECTL} patch $deployment -n projectcontour --patch '{"spec": {"template": {"metadata": {"annotations": {"timestamp": "'$(date +%s)'"}}}}}'
-done
+${KUBECTL} patch deployment -n projectcontour contour --patch '{"spec": {"template": {"metadata": {"annotations": {"timestamp": "'$(date +%s)'"}}}}}'

 # Wait for Contour and Envoy to report "Ready" status.
 ${KUBECTL} wait --timeout="${WAITTIME}" -n projectcontour -l app=contour deployments --for=condition=Available
diff --git a/test/scripts/install-provisioner-working.sh b/test/scripts/install-provisioner-working.sh
index ddbe02e8..70ec461d 100755
--- a/test/scripts/install-provisioner-working.sh
+++ b/test/scripts/install-provisioner-working.sh
@@ -33,13 +33,13 @@ fi
 VERSION="v$$"

 # Build the Contour Provisioner image.
-make -C ${REPO} container IMAGE=ghcr.io/projectcontour/contour VERSION=${VERSION}
+make -C ${REPO} container IMAGE=ghcr.io/projectcontour/contour VERSION=dev

 # Push the Contour Provisioner image into the cluster.
-kind::cluster::load::docker ghcr.io/projectcontour/contour:${VERSION}
+kind::cluster::load::docker ghcr.io/projectcontour/contour:dev

 # Install the Gateway provisioner using the loaded image.
-export CONTOUR_IMG=ghcr.io/projectcontour/contour:${VERSION}
+export CONTOUR_IMG=ghcr.io/projectcontour/contour:dev

 ${KUBECTL} apply -f examples/gateway-provisioner/00-common.yaml
 ${KUBECTL} apply -f examples/gateway-provisioner/01-roles.yaml
@@ -52,6 +52,11 @@ ${KUBECTL} apply -f <(cat examples/gateway-provisioner/03-gateway-provisioner.ya
 # Patching the deployment with timestamp will trigger pod restart.
 ${KUBECTL} patch deployment -n projectcontour contour-gateway-provisioner --patch '{"spec": {"template": {"metadata": {"annotations": {"timestamp": "'$(date +%s)'"}}}}}'

+# Trigger restart of all pods managed by the provisioner.
+for deployment in $(kubectl get deployment -n projectcontour -l app.kubernetes.io/managed-by=contour-gateway-provisioner -o name) ; do
+  ${KUBECTL} patch $deployment -n projectcontour --patch '{"spec": {"template": {"metadata": {"annotations": {"timestamp": "'$(date +%s)'"}}}}}'
+done
+
 # Wait for the provisioner to report "Ready" status.
 ${KUBECTL} wait --timeout="${WAITTIME}" -n projectcontour -l control-plane=contour-gateway-provisioner deployments --for=condition=Available
 ${KUBECTL} wait --timeout="${WAITTIME}" -n projectcontour -l control-plane=contour-gateway-provisioner pods --for=condition=Ready
```

That is,  `install-contour-working.sh` is changed back to what you had previously and  `install-provisioner-working.sh` applies the idea to provisioner use case by patching both provisioner and the managed contours.

PR #5857, opened 2023-10-16
golangci-lint: Fix revive rules

sunjayBhatia:
When we enabled the use-any rule we disabled all the default rules that are run by revive (see: https://revive.run/docs#golangci-lint)

This change grabs all the default rules from
https://github.com/mgechev/revive/blob/master/defaults.toml and adds the use-any rule

### DIFF
--- a/test/e2e/incluster/leaderelection_test.go
+++ b/test/e2e/incluster/leaderelection_test.go
@@ -30,7 +30,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func testLeaderElection(namespace string) {
+func testLeaderElection() {
 	// This test is solely a check on the fact that we have set up leader
 	// election resources as expected. This does not test that internal
 	// components (e.g. status writers) are set up properly given a contour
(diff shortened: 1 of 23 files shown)


sunjayBhatia:
The first commit will fail, showing all the revive lint errors that have been missed since we accidentally disabled all the linting rules

test/e2e/incluster/leaderelection_test.go
```diff
@@ -30,7 +30,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func testLeaderElection(namespace string) {
```
skriss, line 33: Should `namespace` be getting used below instead of `f.Deployment.Namespace.Name`? Seems like the latter would just be `projectcontour` which is not where this test is supposed to be getting run

sunjayBhatia: I think this test is just checking leases/events are created as expected on leader changes and these resources should just exist in the namespace the contour is running in so there isn't actually any need to namespace-scope this one

skriss: Ah yeah OK nvm

skriss (approved):

PR #5864, opened 2023-10-18
internal/envoy: fix wrong check for filter/router

izturn:


internal/envoy/v3/listener.go
```diff
@@ -418,7 +418,7 @@ func (b *httpConnectionManagerBuilder) AddFilter(f *http.HttpFilter) *httpConnec
 		// If this happens, it has to be programmer error, so we panic to tell them
 		// it needs to be fixed. Note that in hitting this case, it doesn't matter we added
 		// the second one earlier, because we're panicking anyway.
-		if f.GetTypedConfig().MessageIs(&envoy_router_v3.Router{}) {
+		if f.GetTypedConfig().MessageIs(&envoy_router_v3.Router{}) && routerIndex != lastIndex {
```
skriss, line 421: It looks like this doesn't actually affect Contour today, because the Router filter is added via DefaultFilters which doesn't use this code path, but this change does make sense regardless. If you already had a Router, and added another, then `routerIndex` would capture the index of the first one.

Please add a test case that covers this -- at least one that covers that a single Router can be added to an existing non-empty list of filters without fail.

izturn: done, find it when implement ext_proc.

sunjayBhatia (approved):
test fails (with panic) w/o this change, thanks!

Issue #5123, opened 2023-02-27
[Feature Discussion] The ways to extend/custom functionalities for Envoy: WASM vs. Lua vs. External Processing vs. GO filter


wilsonwu:
This is a summarize and discussion of the ways to extend or custom Envoy functionalities, let's compare below 4 methods, we don't talk about details of the implementation, just compare and choose which we want to make Contour to support.

Comparison:
### WASM:
ref. https://github.com/projectcontour/contour/issues/4276
Security: Medium
For Envoy, it provide sandbox to make WASM run in an isolation area, the Envoy main process will not be affected.

Extendibility: High
Users can implement any logic in WASM and set the WASM run in any stage of Envoy.

Performance: Low
The WASM performance is not good, from an Istio expert, the WASM only has 50% performance of Envoy native logic.

Others: WASM will increase the NACKs, because load WASM depends on WASM file and from configuration to load file, there is more chances to NACKs.

### Lua:
ref. https://github.com/projectcontour/contour/issues/3006
Security: Low
If user write a bad script, Envoy will be affected.

Extendibility: High
As same as WASM user can create any logic by Lua.

Performance: Medium / Low
From some Istio usage experience, for performance Lua is better then WASM, and from gateway use case, the performance still not good.

### External Processing
ref. https://github.com/projectcontour/contour/issues/5038
Security: High
Like ExAuth and Traffic Ratelimit, use external processing to extend envoy functionalities is a good way, because there is no logic code running in Envoy, just get response from a external service (like auth server), so if the external service has problem, for Envoy, it works well.

Extendibility: Low
For External Processing, must design by the standard, if user want to do anything not related the rules, maybe that can't work.

Performance: High
The external service can deploy more replica and assign more resources to improve performance. 

### GO filter
Security: Low
As my understanding the filter of envoy be extended will inject envoy process, if there is any problem, it will affect envoy.

Extendibility: High
As same as WASM and Lua, we can code anything we want.

Performance: High
For GOLANG the performance is higher than WASM and Lua, but still need to test.

wilsonwu:
Vote from me:
WASM +1
External Processing +1

clayton-gonsalves:
Thanks for adding this, a few comments: 

Adding the design and implementation efforts for each of these would also be nice. 

For example,
- WASM requires significant research and handling of the NACKs before we can move forward. 
- ExtProc has similar patterns with rate limiting and ExtAuthz. The design effort won't be as much. 
[...]

PR #5866, opened 2023-10-19
Design: External Processing support

izturn:

Design proposal to address #5123


design/external-processing-design.md
```diff
@@ -0,0 +1,619 @@
...
+            requestTrailerMode: 2
+            responseBodyMode: 0
+            responseHeaderMode: 1
+            responseTrailerMode: 2
+...
+```
+
+##### Route level
+
+For more precise control, the `Global` and/or `VirtualHost` external processing can also be **overrideed/toggled** on an individual route. the priority is as follows: `Global` < `VirtualHost` < `Route`
```
tsaarni, line 124: It is not completely clear to me when processors are suggested to be combined. User must grasp the processors defined on different levels of configuration, how overrides and priority impacts their execution. The syntax also becomes quite verbose when external processing is configured on a route. 

I wonder if there would be simpler approach?

For example, in case of JWT verification the route-level verification policy refers to policies which are all defined at virtualhost level ([link](https://projectcontour.io/docs/1.27/config/jwt-verification/)). Also in case of external authorization, the extension service itself is not repeated on route level, even though [some finer grained](https://projectcontour.io/docs/1.27/config/client-authorization/#scoping-authorization-policy-settings) contols are provided there.

izturn:
Sorry,  I am very busy this month, But this task will be my main focus next month

design/external-processing-design.md
```diff
@@ -0,0 +1,619 @@
...
+}
+
+// ProcessingPhase define the phase in the filter chain where the external processing filter will be injected
+type ProcessingPhase string
+
+const (
+	// UnspecifiedPhase decides where to insert the external processing service.
+	// This will generally be at the end of the filter chain, right before the Router
+	UnspecifiedPhase ProcessingPhase = "UnspecifiedPhase"
+
```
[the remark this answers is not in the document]
izturn: Derived from [PluginPhase](https://istio.io/latest/docs/reference/config/proxy_extensions/wasm-plugin/#PluginPhase), yes,  the `Default` maybe better

design/external-processing-design.md
izturn: I designed `extProc` based on these two implementations as above. I divided its functions into three levels: `global`, `virtualhost`, and `route`. The priorities of these levels increase successively. The latter two levels also provide options to disable the corresponding functions. This design makes the functions complete, but also complex.

To be honest, I am not particularly clear how users will use them in combination (that's why I came up with this design). But based on our company's product requirements, this function is mainly needed at the `route level`. So, to make it easier to understand and use, we can just implement the `route-level` function in the first phase.  @skriss @sunjayBhatia @tsaarni   WDYT?

tsaarni: Sorry for the delay @izturn! 😳

In case of external authorization there is one reference to `ExtensionService` from top level  `HTTPProxy.spec.virtualhost.authorization`. On the route level `HTTPProxy.spec.routes.authPolicy` one can opt out, and add key:value pairs that will provide [context for the external authorization server](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto#envoy-v3-api-field-extensions-filters-http-ext-authz-v3-checksettings-context-extensions) about the requests.

If copying the same idea here, could it be like this:

* Extension service would be configured only once under `HTTPProxy.spec.virtualhost.extProc` and it automatically applies to all routes
* Under `HTTPProxy.spec.routes.extProcPolicy` overrides could have only minor changes, such as disabling external processing for specific route but not e.g. replace the gRPC service with another one.

As far as I can see, the overrides in external authz are [just these](https://github.com/projectcontour/contour/blob/080940778e4995aa310e8d5e9caab5a0f2747c08/internal/envoy/v3/route.go#L184-L207).

izturn: After our internal discussion and exploration, we found that the global ext_proc may not have much significance at the moment, so I have updated  the design. PTAL @tsaarni

izturn: hey @tsaarni   Based on the positioning of ext_proc as an escape hatch, as well as discussions and explorations within our team, I've simplified the design: only support up to one ext_proc at each level, removing the design of adding multiple ext_proc(s) at different phases of the filter by priority, PTAL

design/external-processing-design.md
```diff
@@ -0,0 +1,648 @@
...
+## High-Level Design
+
+Contour will add HTTP support for Envoy's External Processing.
+
+new type: `ExternalProcessor` and its friends: `ExtProc`, `ExtProcOverride`, `ProcessingMode`, `HeaderMutationRules`,`GRPCService`, `ExtProcPolicy` and more, will be defined for implement the design.
+
+In this design, the configuration is divided into three levels: `Global`, `VirtualHost`, `Route`, each level can be set up to one External processing; each level has a `disabled` option, but at different levels, it has different meanings.
+
+If the external procssing is added to the filter chain(s), it will be inserted just before the `Router` filter.
```
clayton-gonsalves, line 46:
```suggestion
If the external processing is added to the filter chain(s), it will be inserted just before the `Router` filter.
```

design/external-processing-design.md
```diff
@@ -0,0 +1,648 @@
...
+
+### Global level
+
+At the `Global` level, there is at most one external processing configured, and if the `globalExtProc` is NOT nil, and the `processor` is set but `disabled == false`, then it will be append to the filter chain for HTTP and the default chain for HTTPS if this VirtualHost has enabled the fallback certificate; but for the normal HTTPS it varies depending on the configuration at the `VirtualHost` level(see below). If `disabled == true`, it will be ignored.
```
clayton-gonsalves, line 50: why do we need `disabled == false`? cant we check the for the existence  of `globalExtProc` to make a decision whether to enable or disable `globalExtProc`?

design/external-processing-design.md
```diff
@@ -0,0 +1,648 @@
...
+	//
+	// if both disabled and overrides are set. use disabled.
+	//
+	// +optional
+	Disabled bool `json:"disabled,omitempty"`
+
+	// Overrides aspects of the configuration for this route.
+	//
+	// +optional
+	Overrides *ExtProcOverride `json:"overrides,omitempty"`
```
clayton-gonsalves, line 431: why dont we embed `ExtProc` here instead of adding a new type

design/external-processing-design.md
```diff
@@ -0,0 +1,648 @@
...
+	// set in most cases.
+	//
+	// +optional
+	FailOpen bool `json:"failOpen,omitempty"`
+}
+
+// ExtProc defines the envoy External Processing filter which allows an external service to act on HTTP traffic in a flexible way
+// The external server must implement the v3 Envoy external processing GRPC protocol
+// (https://www.envoyproxy.io/docs/envoy/v1.27.0/api-v3/extensions/filters/http/ext_proc/v3/ext_proc.proto).
+type ExtProc struct {
```
clayton-gonsalves, line 364: I think we should add the field `allow_mode_override`

clayton-gonsalves (changes requested):

design/external-processing-design.md
izturn: Just for convenience, because the extProc configuration item is a bit complicated, so that in the test or some other situations, you do not have to repeatedly add and remove the corresponding extProc, for whether to remove it, my attitude is open.

clayton-gonsalves (approved):

design/external-processing-design.md
tsaarni: Currently route level override allows overriding everything. For example, it is possible to point each route to be processed by different external processor gRPC server. I'm curious, do you need this flexibility?

design/external-processing-design.md
```diff
@@ -0,0 +1,639 @@
...
+
+An operator configures external processing on a root `HTTPProxy` by setting the `VirtualHost.ExtProc` field.
+Setting this field without also setting the `TLS` field is an error.
+
+### Progressing Flow
+
+This chart (copy from [External Processing Filter][10]) shows the simplest possible implementation of the filter -- a filter server receives the HTTP request headers, decides to accept the response (and can optionally modify the headers) so it closes the stream cleanly. At this point it is no longer involved in filter processing. see [External Processing Filter][10] for more information.
+
+![drawing](images/ext_proc_flow.png)
+
```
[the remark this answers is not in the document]
izturn: nice

wilsonwu:
For some real use cases my meet, I think API/Route level customize logic is useful, because in some tranditional company, some of thier applications are not follow the standard of microservices and may have strange logics, so they need the Route level customize process to help only one API to do something, and rest of APIs are all work fine without this process, so I think if we can support Route level ext process, it will better.

design/external-processing-design.md
izturn: https://github.com/projectcontour/contour/pull/5866#issuecomment-2021856492 @tsaarni

design/external-processing-design.md
```diff
@@ -0,0 +1,660 @@
...
+	// if both disabled and overrides are set. use disabled.
+	//
+	// +optional
+	Disabled bool `json:"disabled,omitempty"`
+
+	// Overrides aspects of the configuration for this route.
+	//
+	// +optional
+	Overrides *ExtProc `json:"overrides,omitempty"`
+}
```
clayton-gonsalves, line 422: Looking at the implementation in PR I feel the HTTPProxy struct extproc options feel a bit fragmented 

is there a reason we cant bundle them up into one struct member on the `type VirtualHost struct {}`

This is what i was thinking 

```
type ExternalProcessingServer struct {
	// ExtensionServiceRef specifies the extension resource that will handle the client requests.
	//
	// +optional
	ExtensionServiceRef ExtensionServiceReference `json:"extensionRef,omitempty"`

	// ResponseTimeout sets how long the proxy should wait for responses.
	// Timeout durations are expressed in the Go [Duration format](https://godoc.org/time#ParseDuration).
	// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
	// The string "infinity" is also a valid input and specifies no timeout.
	//
	// +optional
	// +kubebuilder:validation:Pattern=`^(((\d*(\.\d*)?h)|(\d*(\.\d*)?m)|(\d*(\.\d*)?s)|(\d*(\.\d*)?ms)|(\d*(\.\d*)?us)|(\d*(\.\d*)?µs)|(\d*(\.\d*)?ns))+|infinity|infinite)$`
	ResponseTimeout string `json:"responseTimeout,omitempty"`

	// If FailOpen is true, the client request is forwarded to the upstream service
	// even if the server fails to respond. This field should not be
	// set in most cases.
	//
	// +optional

	// ProcessingMode describes which parts of an HTTP request and response are sent to a remote server
	// and how they are delivered.
	//
	// +optional
	ProcessingMode *ProcessingMode `json:"processingMode,omitempty"`

	// MutationRules specifies what headers may be manipulated by a processing filter.
	// This set of rules makes it possible to control which modifications a filter may make.
	//
	// for Overrides is must be nil
	//
	// +optional
	MutationRules *HeaderMutationRules `json:"mutationRules,omitempty"`

	// If true, the filter config processingMode can be overridden by the response message from the external processing server `mode_override``.
	// If false, `mode_override` API in the response message will be ignored.
	//
	// +optional
	AllowModeOverride bool `json:"allowModeOverride,omitempty"`
	
	ExtProcPolicy *ExtProcPolicy  `json:"extProcPolicy,omitempty"`
}


...


type VirtualHost struct {
...
ExtProcServer ExternalProcessingServer `json:"extProcServer,omitempty"`
...
}
```

clayton-gonsalves: this also keeps the UX similar to what we have for ratelimiting and extauth

design/external-processing-design.md
clayton-gonsalves: +1 to what @tsaarni says, do we want to define also a separate servers? having the route setting override makes sense

>For some real use cases I meet, I think API/Route level customize logic is useful, because in some tranditional company, some of thier applications are not follow the standard of microservices and may have strange logics, so they need the Route level customize process to help only one API to do something, and rest of APIs are all work fine without this process, so I think if we can support Route level ext process, it will better.

would this not be achievable with just the overrides + 1 globally defined ext proc server?

clayton-gonsalves (changes requested):
left a few comments on the configuration approach

design/external-processing-design.md
izturn: @clayton-gonsalves PTAL, BTW let our review the design first, then the implemention ;-)

izturn:
ping @clayton-gonsalves

design/external-processing-design.md
```diff
@@ -0,0 +1,651 @@
...
+            namespace: extproc-test
+          failOpen: true
+          responseTimeout: 31s
+          processingMode:
+            requestBodyMode: NONE
+            requestHeaderMode: SKIP
+            requestTrailerMode: SKIP
+            responseBodyMode: NONE
+            responseHeaderMode: SEND
+            responseTrailerMode: SKIP
```
tsaarni, line 178: In my opinion, repeating the whole `ExternalProcessing` struct at route level is not optimal API - it gets too verbose and I think for other similar cases we have used different approach earlier.

What would you think about (re)using the approach that [JWT verification](https://projectcontour.io/docs/1.28/config/jwt-verification/) implements, where we define one or more named processors at virtualhost level and then at route level refer to one processor by its name? In that case, one or more routes can refer to a processor without repeating the full set of parameters at each route.

izturn: I considered this approach, but did not adopt it for the following reasons:

1. The current design is based on the **auth** implementation, not the **jwt**.

2. From the Envoy configuration definition, the structure of **jwt** and **ext_proc** is very different. In jwt  the external server(s) defined at the **virtualhost level** are only used as a temporary storage place for server(s) (and then referenced at the route level) and do not have a real effect. However, this is not the case with **ext_proc**. Therefore, if multiple server(s) are defined at the **virtualhost level**, then at **each route level**, the corresponding server(s) must be explicitly enabled or disabled, which increases the difficulty of configuration.

3. The **jwt scheme** is similar to the scheme we designed in the first version. In our actual use, we have found that while this is flexible, it is more difficult to understand and use.

Based on the above reasons and our internal practices, I personally believe that the current design is more reasonable.

tsaarni: >  In jwt the external server(s) defined at the virtualhost level are only used as a temporary storage place for server(s) (and then referenced at the route level) and do not have a real effect. However, this is not the case with ext_proc. Therefore, if multiple server(s) are defined at the virtualhost level, then at each route level, the corresponding server(s) must be explicitly enabled or disabled, which increases the difficulty of configuration.

I might be missing something, since I do not see the difference:

The JWT API defines providers as temporary storage for server info as you say, but it also allows setting one as the [default provider](https://projectcontour.io/docs/main/config/jwt-verification/#setting-a-default-provider) which is then automatically applied to all routes. Routes can still opt-out by disabling JWT verification explicitly, or require a specific (non-default) provider.

On the other hand, if I understood the current proposal correctly, for (non-default) route level external processor, the current API requires user to re-define the external processor server parameters again on each route, even in the case when two separate routes would share single processor. Wouldn't "distributing" the configuration like that lead into duplication of configuration?

clayton-gonsalves: The current design prioritizes flexibility but results in excessive verbosity. It might be beneficial to introduce a vhost-level extproc setting to alleviate redundancy, especially within the same vhost/proxy.

The configuration on the config seems a bit in between to me where we provide flexibilty but in a sense limited. Historically, we've maintained global defaults in the configuration, managed by admins or cluster admins. IMO, service-specific settings within it feels somewhat out of place

Issue #5773, opened 2023-09-21
Should we sort routes produced by `HTTPPRoxy`


davinci26:
## Abstract

Since Envoy is greedy matching path routes, order is important. Contour decides to sort the routes in a way that is not really intuitive and can lead to suprises. In particular with the sorter.go we choose to sort the routing table based on heuristics. This approach can lead to confusion since sorting is low-level implementation primitive of Contour and most users will not know about.

## Personas

There are two main personas in question are:

- Cluster Administrator: A cluster administrator is responsible for providing core functionality to a k8s cluster. For the purposes of this document the cluster administrator installs Contour to the cluster and exposes to the internet. The administrator is responsible for the health of the Contour services.

- ResourceOwner: Someone who manages an HTTPProxy resource inside the Kubernetes cluster configuring the Route/Service combination to allow users of the application to access. A resource owner might not be familiar with Envoy, doesn't have access to the Envoy admin interface and has surface level understanding of how the the HTPProxy Routes are enforced.

## Problem

Imagine the following scenario where you have an HTTPProxy that looks like the following

```
apiVersion: projectcontour.io/v1
kind: HTTPProxy
metadata:
  name: foo
spec:
  ingressClassName: foo
  - authPolicy:
      disabled: true
    conditions:
    - regex: /(?i).foo2.*
    - name: service-foo
      port: 80
  - authPolicy:
      disabled: true
    conditions:
    - regex: /(?i).foo2bar.*
    services:
    - name: service-bar
      port: 80
```

The `HTTPProxy` above will generate a route where the route matching `/(?i).foo2bar.*` will be before the route matching `/(?i).foo2.*`. This is due to the [sorter.go](https://github.com/projectcontour/contour/blob/main/internal/sorter/sorter.go) implementation. The behaviour of the sorter is problematic for the following cases:

* HTTPProxy is written/maintained/debugged by `ResourceOwner` which are not familiar with the internals of Contour and the sorting. If a routing rule doesn't behave as expected the `Cluster Administrator` (or some tooling) would need to verify that the order is correct and there is no other issue.
* Both `Cluster Administrator` and `ResourceOwner` have a hard time reasoning about the order through pure code inspection. This makes debugging issues more complicated.
* It is hard to write tooling/abstractions on top of `HTTPProxy` without leaking this implementation detail to the users.

In more generic terms the sorter is too magical and it can work in unexpected ways.

## What do you want the project to decide?
[...]

PR #5890, opened 2023-10-23
Design Doc: Omit Route Sorting

davinci26:
Design doc for: https://github.com/projectcontour/contour/issues/5773

### DIFF
--- a/design/httpproxy-route-sorting.md
+++ b/design/httpproxy-route-sorting.md
@@ -17,7 +17,7 @@ Since Envoy is greedy matching path routes, order is important. Contour sorts th
 And it breaks ties in the following way:
 
 1. Matches of the same type are ordered by their length.
-2. Matches of the same length are sorted by the presense of additional matching conditions(e.g. if a particular header exists)
+2. Matches of the same length are sorted by the presence of additional matching conditions(e.g. if a particular header exists)
 3. Matches of the same type, length and match conditions are sorted lexicographically.
 
 ## Personas
--- a/design/httpproxy-route-sorting.md
+++ b/design/httpproxy-route-sorting.md
@@ -1,0 +1,170 @@
+# Omit Route Sorting for `HTTPProxy`
+
+Status: In-Review
+
+## Abstract
+
+This document describes a feature flag that allows operators to disable the default route sorting algorithm used for `HTTPProxy`.
+
+## Background
+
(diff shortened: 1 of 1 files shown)


design/httpproxy-route-sorting.md
```diff
@@ -0,0 +1,170 @@
...
+
+```
+
+
+### Route Matching for Objects managed by both an HTTPProxy and Ingress (or Gateway)
+
+Since `Gateway` and `Ingress` have routing conformance which are defined outside of the repository any `virtualHost` that is managed by both `CRDs` will follow the `Gateway`/`Ingress` sorting regardless of
+the configuration.
+
+## Alternatives Considered
```
sunjayBhatia, line 156: Another alternative is to just offer guidance via documentation about what users configuring contour should do to realize certain routing schemes.

e.g. add guidance from this comment: https://github.com/projectcontour/contour/issues/5773#issuecomment-1779332101

this may take the form of generating additional routing rules with specific matches to get the precedence behavior desired

pros:
- no changes to contour
- contour routing stays similar to the ingress/gateway api standard for all users so migration is not an issue

cons:
- could lead to proliferation of special case rules that are hard to reason about for humans
- no guarantee the exact routing specificity can be acheived

sunjayBhatia: would be nice to include this in the document at least for discussion purposes/posterity

davinci26: yeah fair comment, added

design/httpproxy-route-sorting.md
```diff
@@ -0,0 +1,186 @@
...
+| Request                          | With Sorting  | Without Sorting |
+| -------------                    | ------------- | --------------- |
+| GET -H "x-experiment: bar" /bar  | b2            | b1              |
+| GET /bar | Content Cell          | b2            | b2              |
+| GET /                            | b3            | b3              |
+
+
+### Route Matching with Include Statements
+
+Include statements are resolved in a depth first fashion.
```
skriss, line 88: Here the "include" routes are sorted before the routes directly defined in the root HTTPProxy; any particular justification for that order vs. putting them after the directly-defined routes?

davinci26: > Here the "include" routes are sorted before the routes directly defined in the root HTTPProxy; any particular justification for that order vs. putting them after the directly-defined routes?

I dont 100% follow, does `any particular justification for that order` refer to the example below or the `depth first` resolution?

skriss: Sorry, put the comment on the wrong line. It's about the example below, where the route(s) defined by `includes` come before all the routes defined directly on the proxy in `routes` (i.e. `/blog` comes first in the Envoy route table, followed by `/bar` and `/`).

The depth-first bit for included routes makes sense to me.

Issue #5569, opened 2023-07-19
Gateway API conformance: Run conformance report generation in CI


sunjayBhatia:
Gateway API has an experimental version of the test suite that can be run to generate a conformance report, see:
- https://github.com/kubernetes-sigs/gateway-api/blob/main/geps/gep-917.md
- https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/utils/suite/experimental_suite.go
- https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/experimental_conformance_test.go

We should generate the conformance report in CI so we can publish it to the Gateway API repo (and in the future to whatever other conformance reporting infrastructure might exist), e.g. https://github.com/kubernetes-sigs/gateway-api/pull/2192

Open for discussion where/when we should run the reporting and exactly the output, my initial thoughts:
- Should output a file containing the conformance report YAML (possibly in addition to printing the output to the action logs)
- Run on pushes to main and release tags
- Reported Contour component version for non-release versions should use the commit SHA
- Don't run on PRs, just run the normal test suite

sunjayBhatia:
spike branch on what things might look like: https://github.com/sunjayBhatia/contour/blob/6bf46b64005439d8d4010cfe4d6b41683a82bacc/test/conformance/gatewayapi/gateway_conformance_test.go#L53-L89

sunjayBhatia:
Still various issues to work out:
- https://github.com/kubernetes-sigs/gateway-api/issues/2205
- https://github.com/kubernetes-sigs/gateway-api/issues/2206

PR #5893, opened 2023-10-24
Run Gateway API conformance and generate report on tagged builds

sunjayBhatia:
Fixes: #5569


skriss (approved):
Nice! Will test with 1.27.0-rc.1

PR #5894, opened 2023-10-24
Minor fixes to Gateway API conformance on release tag

sunjayBhatia:
- Don't build/load contour image to kind as we should be using the image we just pushed
- Fill in Gateway API version from our environment var

### DIFF
--- a/.github/workflows/build_tag.yaml
+++ b/.github/workflows/build_tag.yaml
@@ -71,7 +71,7 @@ jobs:
           GENERATE_GATEWAY_CONFORMANCE_REPORT: "true"
         run: |
           export CONTOUR_E2E_IMAGE="ghcr.io/projectcontour/contour:$(git describe --tags)"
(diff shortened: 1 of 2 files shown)


sunjayBhatia:
noticed when looking at logs for https://github.com/projectcontour/contour/actions/runs/6632046855/job/18017005554

skriss (approved):

sunjayBhatia:
The build in CI on the 1.27.0 rc-1 tag worked, but it did an extra step to build the local contour and load it into the kind cluster, this change just omits that

skriss:
I see `gatewayAPIVersion: TODO` in the report, is that expected?

sunjayBhatia:
> I see `gatewayAPIVersion: TODO` in the report, is that expected?

yeah, that comes from the suite itself, though maybe we can put in the right version ourselves, I'll push a quick commit

PR #5896, opened 2023-10-25
Use necojackarc/auto-request-review action to request reviews from non-maintainers

sunjayBhatia:
CODEOWNERS only allows requesting reviews from users/teams with owner permission on the repo

Use this action which should not require the reviewers github team to have owner permissions

### DIFF
--- a/.github/reviewers.yaml
+++ b/.github/reviewers.yaml
@@ -1,0 +1,8 @@
+reviewers:
+  defaults:
+    - team:projectcontour/maintainers
+    - team:projectcontour/reviewers
+
+options:
+  ignore_draft: true
+  number_of_reviewers: 2
--- a/.github/reviewers.yaml
+++ b/.github/reviewers.yaml
@@ -1,7 +1,7 @@
 reviewers:
   defaults:
-    - team:projectcontour/maintainers
-    - team:projectcontour/reviewers
+    - team:maintainers
+    - team:contour-reviewers
 
 options:
   ignore_draft: true
--- a/.github/reviewers.yaml
+++ b/.github/reviewers.yaml
@@ -1,8 +1,7 @@
 reviewers:
   defaults:
-    - team:maintainers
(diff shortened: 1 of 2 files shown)


sunjayBhatia:
As part of this we might need to disable reviews on the @projectcontour/maintainers team to let this action handle all the reviewer logic

sunjayBhatia:
Also may want to unset this: https://github.com/projectcontour/contour/blob/6917d85ce41f3050107994c06e6bb5b7039c99ed/.github/dependabot.yml#L17

sunjayBhatia:
unfortunately to get access to the secret needed to access teams, had to switch to the `pull_request_target` event to trigger the workflow on, which then won't run on this PR since the base branch doesn't include the workflow

sunjayBhatia:
we'll likely have to merge this w/o knowing if it works so that PRs have access to the secret/etc. to run this action

.github/reviewers.yaml
```diff
@@ -0,0 +1,8 @@
+reviewers:
+  defaults:
+    - team:maintainers
+    - team:contour-reviewers
+
+options:
+  ignore_draft: true
+  number_of_reviewers: 2
```
skriss, line 8: Is there any way to ensure that at least one maintainer is selected?

sunjayBhatia: ah yeah there might be a way to do so with the "groups" feature, but it might be simpler to keep the CODEOWNERS and just specify reviewers with this action

skriss: Circling back here, I'm in favor of keeping CODEOWNERS for existing maintainers, and using this workflow for assigning non-maintainer reviewers.

sunjayBhatia:
I'll need to regenerate the secret referenced in the workflow as I believe that has expired

skriss (approved):

sunjayBhatia:
Ok regenerated token and merging

PR #5898, opened 2023-10-26
update to Gateway API 1.0

### LINKED ISSUE #5918: Gateway API: bump Go module to 1.0 and get CI passing
This issue covers updating the Gateway API Go module version and doing the ~minimum required to get Contour building and unit/E2E/conformance tests passing. Further work to implement new features, fix failing conformance tests, etc. will be covered in other issues.

skriss:
Just putting this up as a draft for visibility, looks like this is ~the minimum required to bump the dependency and get all tests to pass. It's mainly bumping the API version used for constants (ones that have moved to v1 no longer have v1beta1 equivalents), plus disabling a few newly failing conformance tests for now until we can address them. There's potentially a bunch more to do to get to a good end-state, but this is probably a good start.

Closes #5918.


skriss:
The latest commits show what it takes to update informers and all internal code to use `v1` where possible. Unfortunately, this causes the upgrade tests to fail -- I believe it's because those tests aren't actually applying the latest Gateway API CRD's as part of the upgrade, so they're technically wrong. But, this does demonstrate why it might be better to continue informing on `v1beta1`, at least for a release or so, to be more forgiving around upgrades.

skriss:
This is ready for review now that 1.0 is out, I think getting this merged should unblock most of the rest of the work.

sunjayBhatia (approved):
Looks good!

Regarding the v1beta1/v1 cross-compatibility, looks like we're leaving the e2e tests to create/act on v1beta1 resources and gateway conformance uses v1 so that should cover the two cases (unless we wanted to add something more explicit)

PR #5901, opened 2023-10-27
Make overload.premature_reset_total_stream_count runtime setting configurable

### LINKED ISSUE #5847: Make overload.premature_reset_total_stream_count runtime setting configurable
This is a parameter that can be tuned to help mitigate CVE-2023-44487

Envoy has a default value of 500 but users may want to tune this to prevent resource starvation during an attack.

              In that case, I think we can follow up on https://github.com/projectcontour/contour/pull/5827 with a change to make this runtime field configurable, as it seems it is something that needs tuning to a particular environment/usage

              PRs welcome for that change

_Originally posted by @sunjayBhatia in https://github.com/projectcontour/contour/pull/5826#discussion_r1355816748_

harshil1973:
Fixes #5847


harshil1973:
@tsaarni Can you look into this draft and suggest some changes also I want to test my changes locally in the kind cluster. Can you provide some steps to do it?

PR #5915, opened 2023-10-30
remove operator references

skriss:
- removes operator from compatibility matrix
- removes operator netlify redirects


netlify.toml
```diff
@@ -88,97 +88,6 @@
   to = "https://raw.githubusercontent.com/projectcontour/contour/:splat/examples/render/contour-gateway-provisioner.yaml"
   status = 302
 
-
-##################################
```
skriss, line 92: We can always leave these around if anyone thinks we should, but given 1.24 is now out of support it seemed like a good time to drop 'em.

PR #5927, opened 2023-11-01
clarify logic in endpointstranslator.go

### LINKED ISSUE #5800: clarify logic in endpointstranslator.go
ref. https://github.com/projectcontour/contour/pull/5745#discussion_r1334684738
ref. https://github.com/projectcontour/contour/blob/7feb49e86b3598e78421f6afe249ff6740cd6d9b/internal/xdscache/v3/endpointstranslator.go#L57
ref. https://github.com/projectcontour/contour/pull/5745#discussion_r1344514952

This line, and the (soon-to-be-merged) same line in endpointslicetranslator.go, could be rewritten for clarity. I'd suggest possibly breaking it into multiple independent short-circuits.

xiangpingjiang:
fix #5800

### DIFF
--- a/internal/xdscache/v3/endpointslicetranslator.go
+++ b/internal/xdscache/v3/endpointslicetranslator.go
@@ -55,31 +55,27 @@ func (c *EndpointSliceCache) RecalculateEndpoints(port, healthPort v1.ServicePor
 
 			// Range over each port. We want the resultant endpoints to be a
 			// a cartesian product MxN where M are the endpoints and N are the ports.
-			for _, p := range endpointSlice.Ports {
+			for _, endpointPort := range endpointSlice.Ports {
 				// Nil check for the port.
-				if p.Port == nil {
+				if endpointPort.Port == nil {
 					continue
 				}
 
-				if p.Protocol == nil {
+				if endpointPort.Protocol == nil {
 					continue
 				}
 
-				if *p.Protocol != v1.ProtocolTCP {
-					continue
-				}
-
-				if port.Protocol != *p.Protocol || healthPort.Protocol != *p.Protocol {
+				if *endpointPort.Protocol != v1.ProtocolTCP {
 					continue
 				}
 
 				// Set healthCheckPort only when port and healthPort are different.
-				if p.Name != nil && (healthPort.Name != "" && healthPort.Name == *p.Name && port.Name != healthPort.Name) {
-					healthCheckPort = *p.Port
+				if endpointPort.Name != nil && (healthPort.Name != "" && healthPort.Name == *endpointPort.Name && port.Name != healthPort.Name) {
+					healthCheckPort = *endpointPort.Port
 				}
 
 				// Match by port name.
-				if port.Name != "" && p.Name != nil && port.Name != *p.Name {
+				if port.Name != "" && endpointPort.Name != nil && port.Name != *endpointPort.Name {
 					continue
 				}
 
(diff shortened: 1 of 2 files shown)


internal/xdscache/v3/endpointslicetranslator.go
```diff
@@ -61,7 +61,7 @@ func (c *EndpointSliceCache) RecalculateEndpoints(port, healthPort v1.ServicePor
 					continue
 				}
 
-				if p.Protocol == nil || (healthPort.Protocol != *p.Protocol || port.Protocol != *p.Protocol) && *p.Protocol != v1.ProtocolTCP {
+				if p.Protocol == nil || healthPort.Protocol != *p.Protocol || port.Protocol != *p.Protocol || *p.Protocol != v1.ProtocolTCP {
```
skriss, line 64: I'd like to see this split up into a couple different if statements, perhaps one checking for nil, one checking for non-TCP, and one checking for non-matching protocols on the port or healthPort. Same for endpointstranslator. Thanks!

xiangpingjiang: hello @skriss 

Do you mean this way ?
```Go
	if p.Protocol == nil {
		if *p.Protocol != v1.ProtocolTCP {
			if healthPort.Protocol != *p.Protocol || port.Protocol != *p.Protocol {

			}

		}

	}

```

This way chage the "||" logic

skriss: I was thinking like:

```go
if p.Protocol == nil {
        continue
}

if *p.Protocol != v1.ProtocolTCP {
        continue
}

// similarly, check if *p.Protocol matches port or healthPort protocol
```

xiangpingjiang: Is there some differences between one statement with "||" and splitting up into a couple different if statements ? IMO,it's the same

skriss: IMO it improves readability to split up and have several smaller bits of early-exit logic. Of course, logically they are the same, but the issue was about readability.

internal/xdscache/v3/endpointstranslator.go
```diff
@@ -54,10 +54,13 @@ func RecalculateEndpoints(port, healthPort v1.ServicePort, eps *v1.Endpoints) []
...
 
 		for _, p := range s.Ports {
-			if (healthPort.Protocol != p.Protocol || port.Protocol != p.Protocol) && p.Protocol != v1.ProtocolTCP {
+			if p.Protocol != v1.ProtocolTCP {
 				// NOTE: we only support "TCP", which is the default.
 				continue
 			}
+			if healthPort.Protocol != p.Protocol || port.Protocol != p.Protocol {
+				continue
+			}
```
skriss, line 63: Digging into the unit test failures, I think it's fine to drop this if statement here. This code is intended to just skip over any endpoints ports that are not using TCP protocol, since that's all we support. During the DAG build we already ensure that `healthPort.Protocol` and `port.Protocol` are TCP, so as long as we're filtering out endpoint ports with non-TCP protocol (which we're doing above), then this check on L61-63 is not needed.

internal/xdscache/v3/endpointslicetranslator.go
```diff
@@ -61,7 +61,15 @@ func (c *EndpointSliceCache) RecalculateEndpoints(port, healthPort v1.ServicePor
...
+					continue
+				}
+
+				if *p.Protocol != v1.ProtocolTCP {
+					continue
+				}
+
+				if port.Protocol != *p.Protocol || healthPort.Protocol != *p.Protocol {
 					continue
 				}
```
skriss, line 74: Same comment here, digging into the unit test failures, I think it's fine to drop this if statement here. This code is intended to just skip over any endpoints ports that are not using TCP protocol, since that's all we support. During the DAG build we already ensure that healthPort.Protocol and port.Protocol are TCP, so as long as we're filtering out endpoint ports with non-TCP protocol (which we're doing above), then this check on L61-63 is not needed.

skriss (changes requested):
commented re: unit test failures, cc @sunjayBhatia to check my logic here

skriss:
@xiangpingjiang will you be able to address the failing unit tests here?

xiangpingjiang:
> @xiangpingjiang will you be able to address the failing unit tests here?

yes, I will do it this week

internal/xdscache/v3/endpointstranslator.go
sunjayBhatia: Yeah in practice we'll never create a cluster with a service port or health port that is not TCP (the `LookupService` helper prevents this), so it seems valid to drop this.

The tests are failing because we don't set a health port or even a port number in the `ServicePort` for most of them (and `dag.ServiceCluster.Services[].HealthPort` is an empty value of `corev1.ServicePort` rather than a nil pointer). In these tests as written `healthPort.Protocol` (empty) will never be equal to `p.Protocol` (TCP) and in many of them `port.Protocol` will also be empty. In practice we always have a full value set for `dag.ServiceCluster.Services[].HealthPort` (either set to the same as `ServicePort` or an independent valid value).

sunjayBhatia: I suggest also for readability, renaming `p` to `endpointPort` (or something similar, though `EndpointPort` is the type so seems like a good fit) and `port` to `servicePort`

sunjayBhatia: We could of course fix all the tests to fill in proper values everywhere to keep the original intent (health port and service port have to match the endpoint port protocol), and/or use an empty string check to make the tests happy (e.g. `if (healthPort.Protocol != "" && healthPort.Protocol != p.Protocol) || (servicePort.Protocol != "" && servicePort.Protocol != p.Protocol) { ... }`)

sunjayBhatia (commented):
left a comment on the open thread, seems like what @skriss suggested should work just fine, tested locally myself and with a few other options suggested all get the units to pass

skriss (approved):
LGTM, thanks @xiangpingjiang

PR #5934, opened 2023-11-03
Gateway API: support backend appProtocol

### LINKED ISSUE #5897: Gateway API: support backend appProtocol
ref. https://github.com/kubernetes-sigs/gateway-api/pull/1979

izturn:

fix #5897


internal/dag/gatewayapi_processor.go
```diff
@@ -1761,9 +1766,43 @@ func (p *GatewayAPIProcessor) validateBackendObjectRef(backendObjectRef gatewaya
...
+	}
+
 	return service, nil
 }
 
+// the ServicePort's AppProtocol must be one of the these.
+const (
+	protoK8sH2C     = "kubernetes.io/h2c"
+	protoK8sWS      = "kubernetes.io/ws"
+	protoContourH2C = "projectcontour.io/h2c"
```
sunjayBhatia, line 1780: hm, I don't know if we need to redefine our own h2c protocol here since there is an upstream one available

internal/dag/accessors_test.go
```diff
@@ -151,6 +207,34 @@ func TestBuilderLookupService(t *testing.T) {
...
+			port:           8443,
+			want:           appProtcolService(annotatedService, "tls"),
+		},
+		"lookup service by port number with k8s app protocol: h2c": {
+			NamespacedName: types.NamespacedName{Name: appProtoService.Name, Namespace: appProtoService.Namespace},
+			port:           8443,
+			want:           appProtcolService(appProtoService, "h2c"),
+		},
+
+		"lookup service by port number with unsupported k8s app protocol: wss": {
```
sunjayBhatia, line 221: looks like port 8444 in `appProtoService` actually has protocol `ws`

internal/dag/gatewayapi_processor.go
```diff
@@ -1761,9 +1766,43 @@ func (p *GatewayAPIProcessor) validateBackendObjectRef(backendObjectRef gatewaya
 		return nil, ref.To(resolvedRefsFalse(gatewayapi_v1beta1.RouteReasonBackendNotFound, fmt.Sprintf("service %q is invalid: %s", meta.Name, err)))
 	}
 
+	if err = validateAPPProtocol(&service.Weighted.ServicePort); err != nil {
+		return nil, ref.To(resolvedRefsFalse(gatewayapi_v1.RouteReasonUnsupportedProtocol, err.Error()))
```
sunjayBhatia, line 1770: would be good to add a test to `status_test.go` for this case

sunjayBhatia:
upgraded the release note to `minor`

internal/dag/gatewayapi_processor.go
izturn: i define it here, just to match: httpproxy's protocol: h2c

izturn: @skriss  WDYT?

izturn:
btw the check: `gateway-conformance` is good and bad at times

internal/dag/gatewayapi_processor.go
skriss: I don't think we need to define our own since these are standard upstream values. Let's just stick with the `kubernetes.io/*` ones.

skriss (changes requested):

skriss:
> btw the check: `gateway-conformance` is good and bad at times

If you see tests related to this change flaking, that'd be good to know, otherwise yeah we have seen some other flakes from time to time, we're trying to track them down and resolve them though.

internal/dag/accessors.go
```diff
@@ -103,13 +103,38 @@ func validateExternalName(svc *v1.Service, enableExternalNameSvc bool) error {
 	return nil
 }
 
+// the ServicePort's AppProtocol must be one of the these.
+const (
+	protoK8sH2C     = "kubernetes.io/h2c"
+	protoK8sWS      = "kubernetes.io/ws"
+	protoContourH2  = "projectcontour.io/h2"
+	protoContourTLS = "projectcontour.io/tls"
```
skriss, line 111: Per previous thread, can drop these.

internal/dag/accessors.go
```diff
@@ -103,13 +103,38 @@ func validateExternalName(svc *v1.Service, enableExternalNameSvc bool) error {
...
+)
+
+func toContourProtocol(appProtocol string) (string, bool) {
+	proto, ok := map[string]string{
+		// *NOTE: for gateway-api: the websocket is enabled by default
+		protoK8sWS:  "",
+		protoK8sH2C: "h2c",
+
+		protoContourH2:  "h2",
+		protoContourTLS: "tls",
```
skriss, line 121: Per previous thread, can drop these.

internal/dag/accessors.go
```diff
@@ -103,13 +103,38 @@ func validateExternalName(svc *v1.Service, enableExternalNameSvc bool) error {
...
+	protoK8sH2C     = "kubernetes.io/h2c"
+	protoK8sWS      = "kubernetes.io/ws"
+	protoContourH2  = "projectcontour.io/h2"
+	protoContourTLS = "projectcontour.io/tls"
+)
+
+func toContourProtocol(appProtocol string) (string, bool) {
+	proto, ok := map[string]string{
+		// *NOTE: for gateway-api: the websocket is enabled by default
+		protoK8sWS:  "",
```
skriss, line 117: Thinking about this some more, websocket upgrades are already enabled by default for Gateway API, and this PR does not enable using `kubernetes.io/ws` for HTTPProxy or Ingress, so I think we should _either_:
- not mention `kubernetes.io/ws` in the changelog, and instead just focus it on supporting the `h2c` value; _or_ 
- implement support for `kubernetes.io/ws` for Ingress and HTTPProxy in this PR to be consistent with the changelog

@izturn which would you like to do here?

skriss (changes requested):

internal/dag/accessors.go
izturn: I think we can focus on passing the conformance test first, and then I can send a change that implements option 2 later.

changelogs/unreleased/5934-izturn-minor.md
```diff
@@ -0,0 +1,3 @@
+## Backend Protocol Selection
```
skriss, line 1:
```suggestion
## Gateway API Backend Protocol Selection
```

changelogs/unreleased/5934-izturn-minor.md
```diff
@@ -0,0 +1,3 @@
...
+
+Contour now enables end-users to specify backend protocols by setting the backend Service's [ServicePort.AppProtocol](https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol) parameter. The accepted values for it is `kubernetes.io/h2c`. It's important to note that the `kubernetes.io/ws` is enabled default for `GatewayAPI`, If `AppProtocol` is set, any other configurations, such as the annotation: `projectcontour.io/upstream-protocol.{protocol}` will be disregarded.
```
skriss, line 3:
```suggestion
For Gateway API, Contour now enables end-users to specify backend protocols by setting the backend Service's [ServicePort.AppProtocol](https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol) parameter. The accepted values are `kubernetes.io/h2c` and `kubernetes.io/ws`. Note that websocket upgrades are already enabled by default for Gateway API. If `AppProtocol` is set, any other configurations, such as the annotation: `projectcontour.io/upstream-protocol.{protocol}` will be disregarded.
```

skriss (changes requested):

izturn:
ping @skriss  @sunjayBhatia

skriss (approved):
LGTM

PR #5940, opened 2023-11-07
GatewayAPI: Support timeout

### LINKED ISSUE #5921: Gateway API: implement timeout support
Timeout support has been added to the API (ref. https://gateway-api.sigs.k8s.io/geps/gep-1742/). There are conformance tests that cover this functionality. This feature should be implemented.

izturn:

fix #5921


internal/dag/gatewayapi_processor.go
```diff
@@ -1102,6 +1103,33 @@ func (p *GatewayAPIProcessor) resolveRouteRefs(route any, routeAccessor *status.
 	}
 }
 
+func validateTimeout(timeout *gatewayapi_v1.HTTPRouteTimeouts) (*RouteTimeoutPolicy, *RetryPolicy, error) {
```
skriss, line 1106:
```suggestion
func parseHTTPRouteTimeouts(timeout *gatewayapi_v1.HTTPRouteTimeouts) (*RouteTimeoutPolicy, *RetryPolicy, error) {
```

changelogs/unreleased/5940-izturn-minor.md
```diff
@@ -0,0 +1,4 @@
+## HTTPRoute Timeouts
```
skriss, line 1: Please add a link to https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPRouteTimeouts somewhere in the changelog, as a reference for readers.

internal/dag/gatewayapi_processor.go
```diff
@@ -1102,6 +1103,33 @@ func (p *GatewayAPIProcessor) resolveRouteRefs(route any, routeAccessor *status.
...
+	br := &contour_api_v1.RetryPolicy{}
+	if timeout.BackendRequest != nil {
+		br.PerTryTimeout = string(*timeout.BackendRequest)
+	}
+
+	rp := retryPolicy(br)
+	if rp.PerTryTimeout.Duration() > rtp.ResponseTimeout.Duration() {
+		rtp.ResponseTimeout = rp.PerTryTimeout
+	}
+	return &rtp, rp, nil
```
skriss, line 1130: I _think_ this implementation is fine for now while Gateway API does not yet support retries, but some observations for reviewers/others on how this is working:

- `retryPolicy(...)` enables a single retry (i.e. up to two total backend requests) by default. However, since for Gateway API we don't specify any retriable conditions, I don't think Envoy will actually retry any requests
- given that the conformance tests pass, I _assume_ that the per-try timeout is still taking effect for that first request
- I think it'd be good to set `br.NumRetries` to -1 to explicitly disable retries for now, assuming that the per-try timeout still takes effect on the first backend request in that case. It's a little bit more self-documenting in the Envoy config that way.
- an alternate way we could implement this now is just to set the route "response" timeout to the smaller of the Request or BackendRequest timeout. It might be a little clearer/more obvious for now, but once Gateway API supports retries, then this code will become necessary again. So, I'm fine keeping the implementation as-is given the above notes.

skriss (changes requested):

skriss (changes requested):
I'd also like to see either some builder tests or feature tests that cover this functionality (since you added status tests, builder tests are probably easiest).

izturn:
> I'd also like to see either some builder tests or feature tests that cover this functionality (since you added status tests, builder tests are probably easiest).

done, PTAL

internal/dag/gatewayapi_processor.go
skriss: I did some testing, and it looks like the current code actually does enable retries here (which is a behavior change), since we default to retrying 5xx codes if no other retry conditions are set, and Envoy defaults to a single retry if no other retry count is specified.

The other option would be to not enable an automatic retry, in which case it doesn't really make sense to support `timeouts.backendRequest` right now since without retries it's the same as `timeouts.request`. Once retries are added to the Gateway API spec, then we can circle back and implement them properly, along with `timeouts.backendRequest`.

I'm leaning towards the latter option, to keep things simple and avoid the behavior change of automatically adding a retry, which would likely change again when Gateway API gets proper retry support.

cc @sunjayBhatia would appreciate your thoughts here as well.

izturn: ping @sunjayBhatia

changelogs/unreleased/5940-izturn-minor.md
```diff
@@ -1,4 +1,3 @@
 ## HTTPRoute Timeouts
 
-Contour now enables end-users to specify timeout by configuring the parameter `HTTPRouteRule.Timeouts`, Please ensure the value of `BackendRequest` must be <= the value of `Request` timeout. otherwise `Request` will be set to the same value as `BackendRequest`
-
+Contour now enables end-users to specifys timeout by setting the [HTTPRouteRule.Timeouts](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPRouteTimeouts) parameter, Please ensure the value of `BackendRequest` must be <= the value of `Request` timeout. otherwise `Request` will be set to the same value as `BackendRequest`
```
sunjayBhatia, line 3: nit:

```suggestion
Contour now enables end-users to specify timeouts by setting the [HTTPRouteRule.Timeouts](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPRouteTimeouts) parameter. Please ensure the value of `BackendRequest` must be <= the value of `Request` timeout, otherwise `Request` will be set to the same value as `BackendRequest`.
```

internal/dag/gatewayapi_processor.go
```diff
@@ -1145,14 +1175,24 @@ func (p *GatewayAPIProcessor) computeHTTPRouteForListener(route *gatewayapi_v1be
...
-			redirect             *Redirect
-			mirrorPolicies       []*MirrorPolicy
 			pathRewritePolicy    *PathRewritePolicy
-			urlRewriteHostname   string
+			requestTimeoutPolicy *RouteTimeoutPolicy
 		)
 
+		requestTimeoutPolicy, retryPolicy, err = parseHTTPRouteTimeouts(rule.Timeouts)
+		if err != nil {
+			routeAccessor.AddCondition(gatewayapi_v1beta1.RouteConditionResolvedRefs, metav1.ConditionFalse, status.ReasonInvalidTimeout, err.Error())
```
sunjayBhatia, line 1192: lets add some tests for this added status condition for completeness

also I'm not sure we want to use the ResolvedRefs condition, I think we probably want to use the Programmed condition instead since we're actively not going to program this route rule given the `continue` below

I could even see an argument for setting `Accepted` to false given this would mean the route is not syntactically valid

internal/dag/builder_test.go
```diff
@@ -15936,6 +15977,18 @@ func prefixroute(prefix string, first *Service, rest ...*Service) *Route {
 	}
 }
 
+func timeoutsHTTPProute(prefix string, timeouts *gatewayapi_v1.HTTPRouteTimeouts, first *Service, rest ...*Service) *Route {
+	services := append([]*Service{first}, rest...)
+	rtp, rp, _ := parseHTTPRouteTimeouts(timeouts)
```
sunjayBhatia, line 15982: hm, I don't feel great about using a helper that is being tested to generate the expected output for it's test, lets do this differently so we're not using real code here

changelogs/unreleased/5940-izturn-minor.md
sunjayBhatia: the spec mentions 

```
Because the Request timeout encompasses the BackendRequest timeout, the value of BackendRequest must be <= the value of Request timeout.
```

I'm not sure we should be resetting the `request` timeout to `backendRequest` if the latter is larger, rather the opposite (resetting `backendRequest` down to `request`)

internal/dag/gatewayapi_processor.go
sunjayBhatia: +1 for not changing retry behavior and keeping things as is until there is an upstream retry API

if we can set a backend request timeout without enabling retries at all that would be ideal

cases to consider (and have tests for):
- both timeout values are set
- only `request` timeout is set: seems straightforward
- only `backendRequest` is set

sunjayBhatia: if we want to explicitly say we don't support the `backendRequest` timeout we can do so with status conditions or just fail the conformance test for now

skriss:
merged #5997 which is based on this PR, closing this one out. Thanks @izturn

Issue #3837, opened 2021-06-24
test/e2e: NamespacedTest helper should be able to create/delete multiple namespaces


sunjayBhatia:
Some tests need to manage multiple namespaces to create resources in.

Currently the `NamespacedTest` helper and associated `NamespacedTestBody` signatures are as follows:

```
type NamespacedTestBody func(string)

func (f *Framework) NamespacedTest(namespace string, body NamespacedTestBody) {
```

We're allowed to specify a namespace and any additional namespaces need to be managed in the test.

We could expand this to support an optional list of namespaces to ensure test boilerplate of managing namespaces is minimized/accidental pollution does not happen.

Ideas:
- change signature to `func (f *Framework) NamespacedTest(body NamespacedTestBody, namespaces ...string) {`
  - Assumption is that first ns is the "main" one to use
- change signature to `func (f *Framework) NamespacedTest(namespace string, body NamespacedTestBody, additionalNamespaces ...string) {`
  - Gives ability to keep syntax as-is, only tests that need additional nses will use it

See https://github.com/projectcontour/contour/pull/3803

sunjayBhatia:
Here's a diff of what the second option above starts to look like:

```
diff --git a/test/e2e/framework.go b/test/e2e/framework.go
index 8eb42e0b..dc81705e 100644
--- a/test/e2e/framework.go
+++ b/test/e2e/framework.go
@@ -200,9 +200,9 @@ func (f *Framework) T() ginkgo.GinkgoTInterface {
        return f.t
 }

-type NamespacedTestBody func(string)
+type NamespacedTestBody func(namespaces string, additionalNamespaces ...string)

-func (f *Framework) NamespacedTest(namespace string, body NamespacedTestBody) {
+func (f *Framework) NamespacedTest(body NamespacedTestBody, namespace string, additionalNamespaces ...string) {
        ginkgo.Context("with namespace: "+namespace, func() {
                ginkgo.BeforeEach(func() {
                        f.CreateNamespace(namespace)
@@ -211,7 +211,7 @@ func (f *Framework) NamespacedTest(namespace string, body NamespacedTestBody) {
                        f.DeleteNamespace(namespace, false)
                })

-               body(namespace)
+               body(namespace, additionalNamespaces...)
        })
 }

diff --git a/test/e2e/ingress/001_tls_wildcard_host_test.go b/test/e2e/ingress/001_tls_wildcard_host_test.go
index a70dc8be..139b89d9 100644
[...]

PR #5942, opened 2023-11-07
usage of nameSpacedTest instead of Create/delete

satyazzz123:
Using `NamespacedTest` instead of `CreateNamespace` and `DeleteNamespace`

updates : #3837


test/e2e/httpproxy/httpproxy_test.go
```diff
@@ -582,119 +582,100 @@ descriptors:
...
-					f.DeleteNamespace(ns, false)
-				}
+				// No need to handle namespace deletion here
 			})
-
+	
+			// Call NamespacedTest with specific namespace names
+			f.NamespacedTest(rootNamespaces[0], testRootNamespaces(rootNamespaces))
+			f.NamespacedTest(rootNamespaces[1], testRootNamespaces(rootNamespaces))
 			f.NamespacedTest(nonRootNamespace, testRootNamespaces(rootNamespaces))
```
sunjayBhatia, line 600: we only want to run this actual test once, this should rather look something like this:
```suggestion
			f.NamespacedTest(nonRootNamespace, testRootNamespaces(rootNamespaces), rootNamespaces...)
```

test/e2e/httpproxy/httpproxy_test.go
```diff
@@ -677,24 +674,21 @@ descriptors:
...
-					f.CreateNamespace(ns)
-				}
+				
 				additionalContourArgs = []string{
 					"--watch-namespaces=" + strings.Join(watchNamespaces, ","),
 					"--root-namespaces=" + strings.Join(rootNamespaces, ","),
 				}
 			})
 
 			AfterEach(func() {
```
sunjayBhatia, line 684: can remove this since it is empty

test/e2e/httpproxy/include_exact_condition_test.go
```diff
@@ -36,8 +36,7 @@ func testIncludeExactCondition(namespace string) {
 		)
 
 		for _, ns := range []string{appNamespace, adminNamespace} {
-			f.CreateNamespace(ns)
-			defer f.DeleteNamespace(ns, false)
+			defer f.NamespacedTest(ns, nil)
```
sunjayBhatia, line 39: we need to make sure this test is called with the NamespacedTest helper with the correct namespaces for it to create/delete

we can then get rid of this whole loop (here and other places) since the NamespacedTest helper should be actually calling this test function and doing the create/delete before/after the test for us

sunjayBhatia (changes requested):
looks like there are some syntax errors here that need fixing

PR #5947, opened 2023-11-08
feature(httpproxy): Enhance retry policy

davinci26:
Enhances retry policy with `retry_host_predicate` which allows users to eject the failing host from Envoy retries.

This pattern is described in https://www.envoyproxy.io/docs/envoy/latest/faq/load_balancing/transient_failures.html as a way to have more reliable retries.

### DIFF
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -1175,6 +1175,7 @@ type RetryPolicy struct {
 	// SkipPreviousHost configures Envoy to use Previous Host Retry Predicate
 	// that allows it to choose a different host than the host where previous request
 	// has failed.
+	// +optional
 	SkipPreviousHost bool `json:"skipPreviousHost"`
 }
 
--- a/apis/projectcontour/v1/httpproxy.go
+++ b/apis/projectcontour/v1/httpproxy.go
@@ -1171,6 +1171,11 @@ type RetryPolicy struct {
 	// This field is only respected when you include `retriable-status-codes` in the `RetryOn` field.
 	// +optional
 	RetriableStatusCodes []uint32 `json:"retriableStatusCodes,omitempty"`
+
+	// SkipPreviousHost configures Envoy to use Previous Host Retry Predicate
+	// that allows it to choose a different host than the host where previous request
+	// has failed.
+	SkipPreviousHost bool `json:"skipPreviousHost"`
 }
 
 // ReplacePrefix describes a path prefix replacement.
--- a/changelogs/unreleased/5947-davinci26-minor.md
+++ b/changelogs/unreleased/5947-davinci26-minor.md
@@ -1,0 +1,6 @@
+### feature(httpproxy): Enhance retry policy
+
+Enhances retry policy with retry_host_predicate which allows users to eject the failing host from Envoy retries.
+
+This pattern is described in https://www.envoyproxy.io/docs/envoy/latest/faq/load_balancing/transient_failures.html as a way to have more reliable retries.
+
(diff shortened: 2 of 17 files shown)


davinci26:
@seth-epps also PTAL

apis/projectcontour/v1/httpproxy.go
```diff
@@ -1171,6 +1171,12 @@ type RetryPolicy struct {
 	// This field is only respected when you include `retriable-status-codes` in the `RetryOn` field.
 	// +optional
 	RetriableStatusCodes []uint32 `json:"retriableStatusCodes,omitempty"`
+
+	// SkipPreviousHost configures Envoy to use Previous Host Retry Predicate
+	// that allows it to choose a different host than the host where previous request
+	// has failed.
+	// +optional
+	SkipPreviousHost bool `json:"SkipPreviousHost,omitempty"`
```
seth-epps, line 1179: the json struct tag should be camelcase

internal/dag/dag.go
```diff
@@ -429,6 +429,12 @@ type RetryPolicy struct {
 	// PerTryTimeout specifies the timeout per retry attempt.
 	// Ignored if RetryOn is blank.
 	PerTryTimeout timeout.Setting
+
+	// SkipPreviousHost configures Envoy to use Previous Host Retry Predicate
+	// that allows it to choose a different host than the host where previous request
+	// has failed.
+	// +optional
+	SkipPreviousHost bool `json:"SkipPreviousHost,omitempty"`
```
seth-epps, line 437: Can remove the struct tag and the `+optional`

clayton-gonsalves (changes requested):
minor nit 
https://github.com/projectcontour/contour/blob/main/site/content/docs/main/config/request-routing.md

can you update the above doc under `retryPolicy`

internal/dag/dag.go
davinci26: fixed in 0641aa3d74a102e52039fa9a4f51a0349060eae5

davinci26:
@clayton-gonsalves & @seth-epps  all the comments fixed in 0641aa3d74a102e52039fa9a4f51a0349060eae5

PTAL

clayton-gonsalves (approved):

seth-epps (approved):

internal/dag/dag.go
davinci26: I actually reverted that change because we want the `+optional` such that the generated CRDs have this marked as optional

davinci26: otherwise all `httpproxies` that dont have this `flag` fail the CRD validation

changelogs/unreleased/5947-davinci26-minor.md
```diff
@@ -0,0 +1,6 @@
+### feature(httpproxy): Enhance retry policy
+
+Enhances retry policy with retry_host_predicate which allows users to eject the failing host from Envoy retries.
+
+This pattern is described in https://www.envoyproxy.io/docs/envoy/latest/faq/load_balancing/transient_failures.html as a way to have more reliable retries.
+
```
sunjayBhatia, line 6: something to allude to in a note here and maybe in the field documentation would be that the `host_selection_retry_max_attempts` field of the retry policy defaults to 1 so there will only be 1 try to avoid a previous host per retry

if you increase your retry attempts or have many hosts in your backend service this might shake itself out just fine, but if there are few backends and/or a low number of retry attempts, even with this on you may unexpectedly still be retrying to the same unhealthy instances

internal/dag/dag.go
davinci26: fixed

PR #5959, opened 2023-11-13
Use Info instead of Printf logs for dag build info lines

sunjayBhatia:
These log lines are useful to see by default and do not show up with the default log level when using log.Printf


skriss (approved):
LGTM, good catch

PR #5961, opened 2023-11-14
Gateway API: fix for failing GatewayWithAttachedRoutes conformance test

### LINKED ISSUE #5920: Gateway API: fix for failing GatewayWithAttachedRoutes conformance test
As of Gateway API v1.0.0-rc2, the GatewayWithAttachedRoutes conformance test has started failing. This is related to how attached routes are counted when a Gateway is invalid. ref. https://github.com/kubernetes-sigs/gateway-api/pull/2477

### LINKED ISSUE #5922: Gateway API: fix for GatewayStaticAddresses conformance test
In Gateway API 1.0, a GatewayStaticAddresses conformance test has been added. Contour is not currently passing this test, however Contour does have basic support for Gateways to specify addresses and have them be reflected in the Envoy service. This issue is to investigate the failure and get the test passing (or add detail as to why it can't).

izturn:

fix #5920


internal/dag/status.go
```diff
@@ -18,6 +18,7 @@ import (
 
 	contour_api_v1 "github.com/projectcontour/contour/apis/projectcontour/v1"
 	"github.com/projectcontour/contour/internal/status"
+
```
sunjayBhatia, line 21: nit: can remove this diff

sunjayBhatia (commented):
just a requested change on the release note so far but from a quick look at the test changes/code changes this looks like it achieves what looks like the right outcome for the most part, though ends up being a little repetitive

With this change, we're processing Listeners/Routes multiple times now, once looping through the Listeners for finding which ones are valid, then going through each Listener/Route another time for seeing if the Routes can attach/intersect hostnames, then doing the Listener/Route analysis *again* for the purposes of calculating attached route counts. It would be good to coalesce some of this logic together since it seems like extra processing we could reduce.

Maybe this looks like not returning early from `computeListener` and including the error of why the listener is invalid/if it should still be allowed to have attached routes in the returned `listenerInfo` struct

This isn't something the upstream conformance tests are checking yet, but if this doesn't exist would be good to ensure a test exists for a Gateway Listener that has an invalid configuration in its route selection (e.g. invalid namespace selector due to a field missing or invalid route kind) but a Route with a parent ref that would be valid should still not count

izturn:
ping @sunjayBhatia

internal/dag/status_test.go
```diff
@@ -9262,6 +9263,100 @@ func TestGatewayAPIHTTPRouteDAGStatus(t *testing.T) {
 		wantGatewayStatusUpdate: validGatewayStatusUpdate("http", gatewayapi_v1.HTTPProtocolType, 1),
 	})
 
+	run(t, "HTTP listener, route's parent ref route kind is invalid", testcase{
```
sunjayBhatia, line 9266: nit
```suggestion
	run(t, "HTTP listener with invalid AllowedRoute kind referenced by route parent ref", testcase{
```

examples/example-workload/gatewayapi/blue-green/blue-green.yaml
```diff
@@ -118,8 +118,8 @@ spec:
     - kind: Service
       name: green
       port: 80
-      weight: 70
+      weight: 90
```
skriss, line 121: Let's drop this file's change from this PR, looks inadvertent.

internal/dag/gatewayapi_processor.go
```diff
@@ -200,16 +197,17 @@ func (p *GatewayAPIProcessor) processRoute(
...
 			continue
 		}
 
-		// Get the list of listeners that are (a) included by this parent ref, and
-		// (b) allow the route (based on kind, namespace).
-		allowedListeners := p.getListenersForRouteParentRef(routeParentRef, route.GetNamespace(), routeKind, readyListeners, routeParentStatus)
+		// Get the list of listeners that are
+		// (a) included by this parent ref, and
+		// (b) allow the route (based on kind, namespace)
+		// (c) pass the all check for it
```
skriss, line 209: I'm not sure what this (c) means - could you be more specific?

internal/dag/gatewayapi_processor.go
```diff
@@ -421,7 +429,12 @@ func (p *GatewayAPIProcessor) computeListener(
 	listener gatewayapi_v1beta1.Listener,
 	gwAccessor *status.GatewayStatusUpdate,
 	validateListenersResult gatewayapi.ValidateListenersResult,
-) (bool, *listenerInfo) {
+) (info *listenerInfo) {
```
skriss, line 432: We prefer not to use named return params, let's switch this to non-named and just return the local variable where needed.

skriss (changes requested):
Couple nits but otherwise looks pretty good to me (I'm heavily relying on the tests here for correctness!)

skriss (approved):
LGTM, will leave for @sunjayBhatia

sunjayBhatia (approved):

PR #5963, opened 2023-11-14
Use setup-go action before codeql

sunjayBhatia:
So the go version used is kept up to date, to prevent actions warnings

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -9,6 +9,9 @@ on:
   schedule:
     - cron: '0 10 * * 1'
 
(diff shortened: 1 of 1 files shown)


sunjayBhatia:
should be able to see in the `Perform CodeQL Analysis` job of the codeql action workflow that there is no warning that says a newer go version is needed

sunjayBhatia:
Looks like the message about needing a newer go version is gone in this PRs workflow run: https://github.com/projectcontour/contour/actions/runs/6867930700/job/18677315815?pr=5963

skriss (approved):

PR #5968, opened 2023-11-15
implement Gateway infrastructure labels/annotations

### LINKED ISSUE #5967: Gateway API: implement Gateway infrastructure labels/annotations
ref. https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.GatewayInfrastructure
ref. https://gateway-api.sigs.k8s.io/geps/gep-1762/

skriss:
Closes #5967.

### DIFF
--- a/internal/provisioner/objects/dataplane/dataplane.go
+++ b/internal/provisioner/objects/dataplane/dataplane.go
@@ -341,9 +341,10 @@ func DesiredDaemonSet(contour *model.Contour, contourImage, envoyImage string) *
 
 	ds := &appsv1.DaemonSet{
 		ObjectMeta: metav1.ObjectMeta{
-			Namespace: contour.Namespace,
-			Name:      contour.EnvoyDataPlaneName(),
-			Labels:    contour.AppLabels(),
+			Namespace:   contour.Namespace,
+			Name:        contour.EnvoyDataPlaneName(),
+			Labels:      contour.WorkloadLabels(),
+			Annotations: contour.CommonAnnotations(),
 		},
 		Spec: appsv1.DaemonSetSpec{
 			RevisionHistoryLimit: ref.To(int32(10)),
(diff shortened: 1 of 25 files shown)


internal/provisioner/objects/dataplane/dataplane.go
```diff
@@ -341,9 +341,10 @@ func DesiredDaemonSet(contour *model.Contour, contourImage, envoyImage string) *
 
 	ds := &appsv1.DaemonSet{
 		ObjectMeta: metav1.ObjectMeta{
-			Namespace: contour.Namespace,
-			Name:      contour.EnvoyDataPlaneName(),
-			Labels:    contour.AppLabels(),
+			Namespace:   contour.Namespace,
+			Name:        contour.EnvoyDataPlaneName(),
+			Labels:      contour.WorkloadLabels(),
+			Annotations: contour.CommonAnnotations(),
```
skriss, line 347: It _probably_ makes sense to propagate these annotations to the pod template as well - thoughts?

internal/provisioner/objects/deployment/deployment.go
```diff
@@ -195,9 +195,10 @@ func DesiredDeployment(contour *model.Contour, image string) *appsv1.Deployment
 	}
 	deploy := &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
-			Namespace: contour.Namespace,
-			Name:      contour.ContourDeploymentName(),
-			Labels:    contour.AppLabels(),
+			Namespace:   contour.Namespace,
+			Name:        contour.ContourDeploymentName(),
+			Labels:      contour.WorkloadLabels(),
+			Annotations: contour.CommonAnnotations(),
```
skriss, line 201: ditto re: pod template annotations

internal/provisioner/objects/dataplane/dataplane.go
sunjayBhatia: yeah seems right to me to propagate these down in addition to the other existing labels/annotations we place on the pod template(s)

sunjayBhatia (commented):
overall looks good, as the GEP says we may end up wanting to filter some annotations out from being applied to all resources that we know only apply to certain ones (like Service annotations etc.) but can do that as the need arises

makes sense to me to add the configured labels/annotations to the contour/envoy pod templates too

sunjayBhatia (approved):
LGTM 👍🏽

PR #5969, opened 2023-11-15
Gateway API: add Gateway label to generated resources

skriss:
Adds the gateway.networking.k8s.io/gateway-name label
to generated resources.


internal/provisioner/model/model.go
```diff
@@ -24,9 +24,13 @@ import (
 )
 
 const (
-	// OwningGatewayNameLabel is the owner reference label used for a Contour
-	// created by the gateway provisioner. The value should be the name of the Gateway.
-	OwningGatewayNameLabel = "projectcontour.io/owning-gateway-name"
+	// ContourOwningGatewayNameLabel is the Contour-defined owner reference label applied
+	// to generated resources. The value should be the name of the Gateway.
+	ContourOwningGatewayNameLabel = "projectcontour.io/owning-gateway-name"
```
skriss, line 29: We can probably deprecate/remove this one eventually since it serves the same purpose, just need to be careful about compatibility

internal/provisioner/model/model.go
```diff
@@ -596,18 +604,10 @@ const (
...
-		},
-	}
-}
-
 // OwnerLabels returns owner labels for the provided contour.
 func OwnerLabels(contour *Contour) map[string]string {
 	return map[string]string{
-		OwningGatewayNameLabel: contour.Name,
+		ContourOwningGatewayNameLabel:    contour.Name,
+		GatewayAPIOwningGatewayNameLabel: contour.Name,
```
skriss, line 611: ref. https://github.com/projectcontour/contour/issues/5970

sunjayBhatia: ha, on this note we should probably vet the label values to make sure theyre <= 63 chars

sunjayBhatia (approved):

PR #5972, opened 2023-11-16
fix go-control-plane instantiation order

skriss:
Previously, the go-control-plane snapshot cache was not instantiated and wired up to the snapshot handler (via a snapshotter) until *after* the first DAG run was completed and during the xDS server startup. As a result, the first DAG run never populated the go-control plane snapshot cache, so the go-control-plane xDS server had nothing to respond to initial discovery requests with.

This moves the wiring of the snapshot handler to the snapshot cache up to where the snapshot handler itself is instantiated, before any runnables are started. It also removes the unneeded snapshotter layer of indirection. The snapshot handler now directly generates a go-control-plane snapshot and stores it in the snapshot cache.


.github/workflows/prbuild.yaml
```diff
@@ -169,6 +169,7 @@ jobs:
           NODEIMAGE: ${{ matrix.node_image }}
           LOAD_PREBUILT_IMAGE: "true"
           USE_CONTOUR_CONFIGURATION_CRD: ${{ matrix.use_config_crd }}
+          CONTOUR_E2E_XDS_SERVER_TYPE: envoy
```
skriss, line 172: just for testing on the PR, need to revert before merge

cmd/contour/serve.go
```diff
@@ -482,8 +482,13 @@ func (s *Server) doServe() error {
 		}),
 	}
 
-	// snapshotHandler is used to produce new snapshots when the internal state changes for any xDS resource.
-	snapshotHandler := xdscache.NewSnapshotHandler(resources, s.log.WithField("context", "snapshotHandler"))
+	// snapshotHandler triggers go-control-plane Snapshots based on
+	// the contents of the Contour xDS caches after the DAG is built.
+	snapshotHandler := xdscache_v3.NewSnapshotHandler(
+		resources,
+		envoy_cache_v3.NewSnapshotCache(false, &contour_xds_v3.Hash, s.log.WithField("context", "snapshotCache")),
```
skriss, line 489: This is the important change, the snapshot handler is now connected to the snapshot cache before any runnables are started.

internal/xds/v3/snapshotter.go
```diff
@@ -1,56 +0,0 @@
-// Copyright Project Contour Authors
```
skriss, line 1: This seemed like an unnecessary layer of indirection between the snapshot handler and snapshot cache. Those two are now directly connected.

skriss:
Ran through the E2E's a number of times on this PR and https://github.com/projectcontour/contour/pull/5973 (which includes this change), had a few seemingly unrelated flakes but overall looks good. I'm going to revert the E2E config change now.

sunjayBhatia (approved):
Been playing around with this change this morning, adding some log lines in various places to watch contour startup in more detail, and as far as I can tell this looks correct
 👍🏽 Should be able to fix the daily build e2e failures that happen on startup of contour. Nice work!

PR #5973, opened 2023-11-16
cmd/contour: only create snapshot handler for envoy xDS server

skriss:
Since the snapshot handler is only used for the envoy
xDS server type, only instantiate it and wire it up
when using that server type.


sunjayBhatia (approved):
LGTM, makes sense to not add extra observers etc. when the snapshotHandler is not actually needed

PR #5975, opened 2023-11-17
test/e2e: replace fixed sleeps with require.Eventually

### LINKED ISSUE #5974: update endpoint slice E2E to use require.Eventually(...) instead of fixed sleeps
I've seen the endpoint slice E2E flake a few times recently. Upon review, https://github.com/projectcontour/contour/blob/main/test/e2e/infra/endpointslice_test.go#L63-L99 is using fixed 10-second sleeps to wait for convergence, which can be fragile. Typically we use `require.Eventually` or similar constructs to handle these situations. This should make the test less likely to flake.

skriss:
Makes the endpoint slice E2E test faster and
less likely to flake.

Closes #5974.

### DIFF
--- a/test/e2e/infra/endpointslice_test.go
+++ b/test/e2e/infra/endpointslice_test.go
@@ -16,7 +16,8 @@
 package infra
 
 import (
-	"time"
+	"slices"
+	"sort"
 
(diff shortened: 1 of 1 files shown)


sunjayBhatia (approved):
LGTM, should be more resilient to flakes/eventual convergence and also less lines of code to read 👍🏽

PR #5987, opened 2023-11-25
Inspect loadbalancer address from Envoy ingress objects

### LINKED ISSUE #4952: support getting LoadBalancerStatus from an Ingress (e.g. AWS ALB) or Gateway

hligit:
Contour sets load balancer address in Ingress/HTTPProxy object's status. The load balancer address is extracted from Envoy `Service` object's status. This doesn't work in configuration that Envoy `Service` is of type `ClusterIP` and an Envoy `Ingress` object is used for inbound. This patch adds support for inspecting the Envoy `Ingress` object for setting load balancer address.

A new  flag `--load-balancer-status` is introduced as suggested by @tsaarni  in https://github.com/projectcontour/contour/pull/5083#issuecomment-1700491125

Flag `--load-balancer-status` takes value in one of below formats:

1. `hostname:fqdn1[,fqdn2]`:  This provides the same functionality as `--ingress-status-address`.
2. `sevice:namespace/name`: same as flags `envoy-service-name` and `envoy-service-namespace` combined.
3. `ingress:namespace/name`: new feature for getting address from Ingress objects.

Since the new flag covers functionality of some existing flags, if both `--load-balancer-status` and `--ingress-status-address` or `envoy-service-name/namespace` are specified, only value of `--load-balancer-status` is used and a warning message is logged.

This is based on [kahirokunn](https://github.com/kahirokunn)'s PR https://github.com/projectcontour/contour/pull/5083 .

Fixes #4952

### DIFF
--- a/apis/projectcontour/v1alpha1/contourconfig.go
+++ b/apis/projectcontour/v1alpha1/contourconfig.go
@@ -239,6 +239,12 @@ type EnvoyConfig struct {
 	// +optional
 	Service *NamespacedName `json:"service,omitempty"`
 
+	// Ingress holds Envoy service parameters for setting Ingress status.
+	//
+	// Contour's default is { namespace: "projectcontour", name: "envoy" }.
+	// +optional
+	Ingress *NamespacedName `json:"ingress,omitempty"`
+
 	// Defines the HTTP Listener for Envoy.
 	//
 	// Contour's default is { address: "0.0.0.0", port: 8080, accessLog: "/dev/stdout" }.
--- a/apis/projectcontour/v1alpha1/contourconfig.go
+++ b/apis/projectcontour/v1alpha1/contourconfig.go
@@ -239,11 +239,15 @@ type EnvoyConfig struct {
 	// +optional
 	Service *NamespacedName `json:"service,omitempty"`
 
-	// Ingress holds Envoy service parameters for setting Ingress status.
+	// LoadBalancer specifies how Contour should set the ingress status address.
+	// If provided, the value can be in one of the formats:
+	//   - address:<address,...>: Contour will use the provided comma separated list of addresses directly. The address can be a fully qualified domain name or an IP address.
+	//   - service:<namespace>/<name>: Contour will use the address of the designated service.
+	//   - ingress:<namespace>/<name>: Contour will use the address of the designated ingress.
 	//
-	// Contour's default is { namespace: "projectcontour", name: "envoy" }.
+	// Contour's default is an empty string.
 	// +optional
-	Ingress *NamespacedName `json:"ingress,omitempty"`
+	LoadBalancer string `json:"loadBalancer,omitempty"`
 
 	// Defines the HTTP Listener for Envoy.
 	//
--- a/cmd/contour/serve.go
+++ b/cmd/contour/serve.go
@@ -20,6 +20,7 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/alecthomas/kingpin/v2"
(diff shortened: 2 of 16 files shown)


cmd/contour/serve.go
```diff
@@ -684,29 +687,84 @@ func (s *Server) doServe() error {
...
-	// Register an informer to watch envoy's service if we haven't been given static details.
-	if lbAddress := contourConfiguration.Ingress.StatusAddress; len(lbAddress) > 0 {
-		s.log.WithField("loadbalancer-address", lbAddress).Info("Using supplied information for Ingress status")
-		lbsw.lbStatus <- parseStatusFlag(lbAddress)
-	} else {
-		serviceHandler := &k8s.ServiceStatusLoadBalancerWatcher{
-			ServiceName: contourConfiguration.Envoy.Service.Name,
-			LBStatus:    lbsw.lbStatus,
-			Log:         s.log.WithField("context", "serviceStatusLoadBalancerWatcher"),
+	if contourConfiguration.Envoy.LoadBalancer != "" {
```
tsaarni, line 690: When user does not create `ContourConfig` CR, this will still never be empty string: It gets unconditionally set [here](https://github.com/hligit/contour/blob/98835b0e0a3a397eadfcb5096ff41cefc05fbcd0/cmd/contour/servecontext.go#L580) with [this](https://github.com/hligit/contour/blob/98835b0e0a3a397eadfcb5096ff41cefc05fbcd0/pkg/config/parameters.go#L1034) value via [this](https://github.com/hligit/contour/blob/98835b0e0a3a397eadfcb5096ff41cefc05fbcd0/cmd/contour/serve.go#L349-L352) execution path which fills in the defaults. As result, old options never take effect, even when the new option was not explicitly set.

tsaarni (changes requested):
Hi @hligit, thanks for contributing, and continuing @kahirokunn's work!

I had small initial finding while doing a quick manual testing. Please check comments inline!

> Since the new flag covers functionality of some existing flags, if both --load-balancer-status and --ingress-status-address or envoy-service-name/namespace are specified, only value of --load-balancer-status is used and a warning message is logged.

I think that for backwards compatibility, even in case of configuration options like these, it is "safer" to give priority for the old method and not the other way around. Often it may not make big difference, but in this case as the new option got assigned with default values, old option just stops working. It may be simplest just to swap the precedence and prioritise old options first.

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -253,6 +253,12 @@ type EnvoyConfig struct {
 	// +optional
 	Service *NamespacedName `json:"service,omitempty"`
 
+	// LoadBalancer holds load balancer parameters for Ingress status.
+	//
+	// Contour's default is "service:projectcontour/envoy"
+	// +optional
+	LoadBalancer string `json:"loadBalancer,omitempty"`
```
tsaarni, line 260: For the documentation, it would be great to include examples here on the supported "kinds", to give overall idea what is possible to configure. This comment will be basis of generated API docs. Besides the generated API docs, the new option should be documented also in [`site/content/docs/main/configuration.md`](https://github.com/projectcontour/contour/blob/main/site/content/docs/main/configuration.md).

tsaarni (changes requested):

cmd/contour/serve.go
hligit: Good catch! Thanks! I have removed setting value in `Defaults()`.
Also changed the order of flags by inspecting if ingress-status-address is empty, than the new flag and lastly the `envoy-service-name` and `envoy-service-namespace`, because they always have [default values set](https://github.com/hligit/contour/blob/a216475ca1000ad26659de09a64edff4090b8bc2/pkg/config/parameters.go#L1028-L1029).

Do you think this is good enough or we should check if the envoy service flags are explicitly set and should take precedence of the new flags?

apis/projectcontour/v1alpha1/contourconfig.go
hligit: Thanks for calling this out! I've updated both.

hligit:
Thanks @tsaarni for reviewing and the guidance! Sorry for my late response!

I've addressed the comments in amended commits. There is one issue regarding flag precedence in my [reply](https://github.com/projectcontour/contour/pull/5987#discussion_r1416336251) of your inline comment.

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -253,6 +253,16 @@ type EnvoyConfig struct {
 	// +optional
 	Service *NamespacedName `json:"service,omitempty"`
 
+	// LoadBalancer specifies how Contour should set the ingress status address.
+	// If provided, the value can be in one of the formats:
+	//   - hostname:<address,...>: Contour will use the provided comma separated list of addresses directly.
```
sunjayBhatia, line 258: currently we support setting ingress status address via config flag/parameter here and it can be an IP or hostname

would be good to keep that functionality (and not just support hostnames) if we intend to deprecate the old method of configuring this

looks like that might just mean a naming change here and elsewhere since existing helpers are being used

kahirokunn:
keep

kahirokunn:
keep

kahirokunn:
keep

kahirokunn:
keep

kahirokunn:
keep 😢

kahirokunn:
keep

kahirokunn:
keep 😢

tsaarni:
I'll make an attempt to pick up this PR to the finish line, xref  #7595

kahirokunn:
keep

kahirokunn:
Keep

Issue #6001, opened 2023-12-09
Support override the defaults for `circuit breakers`


davinci26:
The way https://projectcontour.io/docs/1.27/config/annotations/ work right now is that if they are not set in the service annotations as expected then they are not set letting Envoy default them.

```
	if envoy.AnyPositive(service.MaxConnections, service.MaxPendingRequests, service.MaxRequests, service.MaxRetries) {
		cluster.CircuitBreakers = &envoy_cluster_v3.CircuitBreakers{
			Thresholds: []*envoy_cluster_v3.CircuitBreakers_Thresholds{{
				MaxConnections:     protobuf.UInt32OrNil(service.MaxConnections),
				MaxPendingRequests: protobuf.UInt32OrNil(service.MaxPendingRequests),
				MaxRequests:        protobuf.UInt32OrNil(service.MaxRequests),
				MaxRetries:         protobuf.UInt32OrNil(service.MaxRetries),
			}},
		}
	}
```

This works great but in bigger clusters the Envoy default might not be good enough. Then it is becomes the responsibility of every service owner to set the annotations in their service correctly which can be error prone.

We would like to add a `global` default such that we can perform logic in the form of `service.MaxConnections` or `GlobalDefault`

davinci26:
This is a bit more long term but related we should bring those into the `HTTPProxy` CRD, it feels like they are annotations due to legacy reasons

therealak12:
Also, solves https://github.com/projectcontour/contour/issues/1192. Doesn't it?

PR #5950, opened 2023-11-11
enabled trackRemaining in Envoy

satyazzz123:
Enabled TrackRemaining in Envoy via Contour proxy or annotations.

updates : #5794

PR #6013, opened 2023-12-12
Adds support for global circuit budget

davinci26:
Solves #6001


davinci26:
removing @clayton-gonsalves since he is on PAT leave (if you see this feel free to review anyway).

Adding @shadialtarsha and @seth-epps as well

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -107,6 +112,17 @@ const (
 	EnvoyServerType XDSServerType = "envoy"
 )
 
+type GlobalCircuitBreakerDefaults struct {
+	// +optional
```
davinci26, line 116: this requires a doc string

cmd/contour/servecontext.go
```diff
@@ -584,13 +584,14 @@ func (ctx *serveContext) convertToContourConfigurationSpec() contour_api_v1alpha
...
-		RateLimitService:            rateLimitService,
-		Policy:                      policy,
-		Metrics:                     &contourMetrics,
-		Tracing:                     tracingConfig,
-		FeatureFlags:                ctx.Config.FeatureFlags,
+		EnableExternalNameService:    &ctx.Config.EnableExternalNameService,
+		GlobalExternalAuthorization:  globalExtAuth,
+		RateLimitService:             rateLimitService,
+		Policy:                       policy,
+		GlobalCircuitBreakerDefaults: ctx.Config.GlobalCircuitBreakerDefaults,
```
shadialtarsha, line 591: I would pass this as a pointer instead of relying on implicit coercion.

shadialtarsha (commented):
I think the code makes sense to me. 
I have 2 questions:
1. What will happen if not all `GlobalCircuitBreakerDefaults` attributes are defined? I think they will have the value 0 and we will default to Envoy settings again which makes sense but I would clarify this in the doc.
2. Wondering if we can write e2e tests for this feature. I know it might be tricky.

apis/projectcontour/v1alpha1/contourconfig.go
```diff
@@ -76,6 +76,11 @@ type ContourConfigurationSpec struct {
 	// +optional
 	Policy *PolicyConfig `json:"policy,omitempty"`
 
+	// GlobalCircuitBreakerDefaults specifies default circuit breaker budget across all services.
+	// If defined, this will be used as the default for all services.
+	// +optional
+	GlobalCircuitBreakerDefaults *GlobalCircuitBreakerDefaults `json:"circuitBreaker,omitempty"`
```
skriss, line 82: I think it would make sense to move this to under `ClusterParameters` (so `spec.envoy.cluster.circuitBreakers`)

skriss, line 82: Let's pluralize the YAML name:
```suggestion
	// GlobalCircuitBreakerDefaults specifies default circuit breaker budget across all services.
	// If defined, this will be used as the default for all services.
	// +optional
	GlobalCircuitBreakerDefaults *GlobalCircuitBreakerDefaults `json:"circuitBreakers,omitempty"`
```

pkg/config/parameters.go
```diff
@@ -827,6 +830,11 @@ type MetricsParameters struct {
 	Envoy   MetricsServerParameters `yaml:"envoy,omitempty"`
 }
 
+// RedditParameters
+type RedditParameters struct {
+	CircuitBreaker *contour_api_v1alpha1.GlobalCircuitBreakerDefaults `yaml:"circuitBreaker,omitempty"`
+}
+
```
skriss, line 837: Can drop this :)

pkg/config/parameters.go
```diff
@@ -669,6 +669,9 @@ type Parameters struct {
 	// GlobalExternalAuthorization optionally holds properties of the global external authorization configuration.
 	GlobalExternalAuthorization GlobalExternalAuthorization `yaml:"globalExtAuth,omitempty"`
 
+	// GlobalCircuitBreakerDefaults holds configurable global defaults for the circuit breakers.
+	GlobalCircuitBreakerDefaults *contour_api_v1alpha1.GlobalCircuitBreakerDefaults `yaml:"circuitBreaker,omitempty"`
```
skriss, line 673: Same comments here - let's move under `ClusterParameters`, and pluralize YAML name to `circuitBreakers`.

skriss, line 673: Please document the new settings in https://github.com/projectcontour/contour/blob/main/site/content/docs/main/configuration.md#configuration-file as well.

cmd/contour/serve.go
```diff
@@ -1209,6 +1211,7 @@ func (s *Server) getDAGBuilder(dbc dagBuilderConfig) *dag.Builder {
 			GlobalRateLimitService:        dbc.globalRateLimitService,
 			PerConnectionBufferLimitBytes: dbc.perConnectionBufferLimitBytes,
 			SetSourceMetadataOnRoutes:     true,
+			GlobalCircuitBreakerDefaults:  dbc.globalCircuitBreakerDefaults,
```
skriss, line 1214: I think it makes sense to have these global defaults take effect for all three processors (Ingress, HTTPProxy, Gateway) -- so need to add the implementation to `internal/dag/gatewayapi_processor.go`, and then make sure we're passing the defaults into all three processors here when we instantiate them.

internal/metrics/metrics.go
```diff
@@ -87,6 +88,8 @@ const (
 	statusUpdateConflict        = "contour_status_update_conflict_total"
 	statusUpdateNoop            = "contour_status_update_noop_total"
 	statusUpdateDurationSeconds = "contour_status_update_duration_seconds"
+
+	ContourCircuitBreakerSettings = "contour_dag_circuit_breaker_settings"
```
skriss, line 92: Could you explain why you want to expose this as a metric? We don't really have any other metrics like this, so it's a little unusual.

skriss (changes requested):
Thanks for the PR @davinci26! Overall this makes sense to me, added some comments.

internal/metrics/metrics.go
davinci26: > Could you explain why you want to expose this as a metric? 

We want to alert and build graphs on this as a defence in depth. We have discovered during  a deployment that we accidentally deleted those annotations and that caused issues and that was really painful to debug.

Basically it helps with the discoverability of what people are setting across the cluster.

cmd/contour/serve.go
davinci26: Yeah I did for `ingress` and `httpproxy`, I wasnt aware if there was anything specific to gateway api spec that made this feature not that useful. Will add it

cmd/contour/servecontext.go
shadialtarsha: actually that is not true, both are pointers 😅

internal/dag/policy.go
```diff
@@ -809,3 +812,49 @@ func loadBalancerRequestHashPolicies(lbp *contour_api_v1.LoadBalancerPolicy, val
...
+		s.MaxPendingRequests = cb.MaxPendingRequests
+	}
+
+	if s.MaxRequests == 0 && cb != nil {
+		s.MaxRequests = cb.MaxRequests
+	}
+
+	if s.MaxRetries == 0 && cb != nil {
+		s.MaxRetries = cb.MaxRetries
+	}
```
seth-epps, line 835: If the global circuit breaker isn't `nil`, but the fields are omitted (since they're optional), are we relying on the envoy cluster defaults?

Based on the envoy cluster configs here, https://github.com/projectcontour/contour/blob/main/internal/envoy/v3/cluster.go#L81-L90, I'm worried that any provided value will change those defaults since they're not passed as `nil`, but instead the default `uint32` 0

davinci26: good catch!

davinci26: Now I remember, darn I felt into this rabbit `hole` again

```
// UInt32OrNil returns a wrapped UInt32Value. If val is 0, nil is returned
func UInt32OrNil(val uint32) *wrapperspb.UInt32Value {
	switch val {
	case 0:
		return nil
	default:
		return wrapperspb.UInt32(val)
	}
}
```

handles 0 alright

pkg/config/parameters.go
davinci26: fixed in bdf1de3b2658964a0b571ef62674caaa8efe3ffe

davinci26: fixed in bdf1de3b2658964a0b571ef62674caaa8efe3ffe

apis/projectcontour/v1alpha1/contourconfig.go
davinci26: fixed in bdf1de3b2658964a0b571ef62674caaa8efe3ffe

davinci26:
@skriss this is ready for another review round

cmd/contour/serve.go
skriss, line 1214: Still need to pass `dbc.globalCircuitBreakerDefaults` into the `IngressProcessor`, above.

site/content/docs/main/configuration.md
```diff
@@ -157,10 +157,14 @@ The cluster configuration block can be used to configure various parameters for
 |-----------------------------------|--------|---------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | dns-lookup-family                 | string | auto    | This field specifies the dns-lookup-family to use for upstream requests to externalName type Kubernetes services from an HTTPProxy route. Values are: `auto`, `v4`, `v6`, `all` |
 | max-requests-per-connection       | int    | none    | This field specifies the maximum requests for upstream connections. If not specified, there is no limit                                                                         |
+| circuit-breakers       | [CircuitBreakers](###circuit-breakers)    | none    | This field specifies the default value for [circuit-breaker-annotations](https://projectcontour.io/docs/main/config/annotations/) for services that don't specify them.                                                                    |
```
skriss, line 160:
```suggestion
| circuit-breakers       | [CircuitBreakers](#circuit-breakers)    | none    | This field specifies the default value for [circuit-breaker-annotations](https://projectcontour.io/docs/main/config/annotations/) for services that don't specify them.                                                                    |
```

changelogs/unreleased/6013-davinci26-small.md
```diff
@@ -0,0 +1,4 @@
+## Support for Global Circuit Breaker Policy
```
skriss, line 1: Let's bump this up to `minor` since it has a header and more than a sentence of description, so need to rename the file and update the PR label.

changelogs/unreleased/6013-davinci26-small.md
```diff
@@ -0,0 +1,4 @@
...
+
+The way [circuit-breaker-annotations](https://projectcontour.io/docs/1.27/config/annotations/) work currently is that when not present they are being defaulted to Envoy defaults. The Envoy defaults can be quite low for larger clusters with more traffic so if
+a user accidentally deletes them or unset them this cause an issue. With this change we are providing contour administrators the ability to provide global defaults that are good. In that case even if the user forgets to set them or deletes them they can have the safety net of good defaults.
```
skriss, line 4: Let's mention that these are set in the config file via `cluster.circuit-breakers`, and add a link to https://projectcontour.io/docs/1.28/configuration/#circuit-breakers release, or via ContourConfiguration CRD in `spec.envoy.cluster.circuitBreakers` with a link to https://projectcontour.io/docs/1.28/config/api/#projectcontour.io/v1alpha1.GlobalCircuitBreakerDefaults

It's fine that the links are broken right now, they'll resolve when the release is cut.

skriss (changes requested):

cmd/contour/serve.go
davinci26: thanks nice catch, I fixed the `cluster_test` and I had a false sense of confidence, I codified this in `serve_test` as well

changelogs/unreleased/6013-davinci26-small.md
davinci26: fixed in 4151035aa07d44ff5b162b52dd73935da4f7cda2

changelogs/unreleased/6013-davinci26-small.md
davinci26: fixed in 4151035aa07d44ff5b162b52dd73935da4f7cda2

site/content/docs/main/configuration.md
davinci26: fixed in 4151035aa07d44ff5b162b52dd73935da4f7cda2

cmd/contour/serve.go
davinci26: fixed in 4151035aa07d44ff5b162b52dd73935da4f7cda2 + added tests

davinci26:
@skriss unless there is an embarrassing bug somewhere this is ready

skriss:
Looks like we got some merge conflicts, if you can resolve those I'll take another look

davinci26:
@skriss merge conflicts resolved

internal/metrics/metrics.go
```diff
@@ -239,6 +242,10 @@ func NewMetrics(registry *prometheus.Registry) *Metrics {
 			},
 			[]string{"kind", "error"},
 		),
+		CircuitBreakerSettings: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: ContourCircuitBreakerSettings,
+			Help: "Number of times a worker process has been restarted",
```
skriss, line 247: Help text should be updated here to be relevant

internal/metrics/metrics.go
```diff
@@ -239,6 +242,10 @@ func NewMetrics(registry *prometheus.Registry) *Metrics {
 			},
 			[]string{"kind", "error"},
 		),
+		CircuitBreakerSettings: prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
skriss, line 245: Please also add the new metric to the `Zero()` function and re-run `make generate` to generate docs for it.

skriss (approved):
This LGTM, I'm approving with comments that need to be addressed but can be done in a follow-up PR if preferred. Will leave for another maintainer to take a look as well. Thanks @davinci26!

internal/metrics/metrics.go
davinci26: yeah yeah you right, fixed now!

internal/metrics/metrics.go
davinci26: Cool I didn't know about this

davinci26:
@projectcontour/maintainers can someone PTAL as well

internal/metrics/metrics.go
sunjayBhatia: Hm, I would expect something like this to be exposed via resource Status or something similar rather than a Contour xDS server metric

Also I believe we have an issue/PR to expand on cluster circuit breaker metrics: https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_stats#circuit-breakers-statistics (here: https://github.com/projectcontour/contour/pull/5950)

I don't want to hold up the rest of the change, so maybe lets pull the metrics part out into another PR?

davinci26: > Hm, I would expect something like this to be exposed via resource Status or something similar rather than a Contour xDS server metric


Personally I take a more liberal approach to metrics in the sense that, if `contour` makes the decision to set a specific value it would be useful for operators to know the decision at runtime, especially if it affect things

davinci26: > I don't want to hold up the rest of the change, so maybe lets pull the metrics part out into another PR?

I won't die on this hill :D, So I  can separate the changes into two PRs. But my general take is that adding the metric is pretty "cheap" and overall useful. I.e. there have been cases where I wish I had this metric tracked.

sunjayBhatia (commented):
i think everything LGTM minus the metric part, I wonder if we can pull that out into a separate PR so we can discuss a little more?

davinci26:
> i think everything LGTM minus the metric part, I wonder if we can pull that out into a separate PR so we can discuss a little more?

Will do it in a separate PR

davinci26:
there is more cleanup to be made

davinci26:
I think I got all the references now

internal/metrics/metrics.go
```diff
@@ -87,6 +87,8 @@ const (
 	statusUpdateConflict        = "contour_status_update_conflict_total"
 	statusUpdateNoop            = "contour_status_update_noop_total"
 	statusUpdateDurationSeconds = "contour_status_update_duration_seconds"
+
+	ContourCircuitBreakerSettings = "contour_dag_circuit_breaker_settings"
```
sunjayBhatia, line 91: last thing to remove i think here

site/content/docs/v1.9.0/annotations.md
```diff
@@ -23,7 +23,7 @@ In both cases, they will behave as follows, by default:
 * If specified as `kubernetes.io/ingress.class: contour`, then Contour serves the Ingress.
 * If any other value, Contour ignores the Ingress definition.
 
-You can override the default class `contour` by providing the `--ingress-class-name` flag to Contour. 
+You can override the default class `contour` by providing the `--ingress-class-name` flag to Contour.
```
sunjayBhatia, line 26: can probably get rid of this change being part of this PR too, since this is quite an old version etc.

sunjayBhatia (approved):

therealak12:
Also, solves https://github.com/projectcontour/contour/issues/1192. Doesn't it?

skriss:
> Also, solves #1192. Doesn't it?

Looks like it, closing that one out.

PR #6016, opened 2023-12-14
internal/dag: Add support for per-host max connections

### LINKED ISSUE #6015: Support per_host_thresholds Envoy circuit breakers
Having the ability to set per-host circuit breaker thresholds would be very useful. At the moment, it is only possible to set cluster scoped thresholds.

Note that only max_connections are supported currently by envoy.

I would like to work on this feature addition.

relu:
Support per-host max connections circuit breaker threshold enabling Envoy to enforce a maximum number of connections for each individual Kubernetes service endpoint using a new service-level annotation: `projectcontour.io/per-host-max-connections`.

Resolves #6015


davinci26:
Envoy docs are a bit confusing to me, does `host` refer to cluster endpoint in this case?

davinci26 (commented):
We need a release note for this. 

https://github.com/projectcontour/contour/pull/6013 is also conflicting. I think we need to be able to set a global default for this.

davinci26:
otherwise the PR lgtm, thanks for the feature, we might even put it into use actually

relu:
> does host refer to cluster endpoint in this case?

That is what I understand from [here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/circuit_breaker.proto#config-cluster-v3-circuitbreakers)

> I think we need to be able to set a global default for this.

That could make sense if one wants to configure a safe default max-connections limit covering all ingress configurations.

Sorry, it seems I forgot to include the release note. I'll add it in a few minutes. Thank you!

davinci26 (approved):
LGTM if you merge before https://github.com/projectcontour/contour/pull/6013 gets in then you have to do the defaults otherwise I can do it

davinci26:
@projectcontour/maintainers can you PTAL so we can merge it.

It has some conflicts with https://github.com/projectcontour/contour/pull/6013 so whoever merges it first need to handle it

skriss:
@davinci26 thanks for reviewing, I'll take a look this week

skriss (approved):
thanks @relu, LGTM

skriss:
@sunjayBhatia will give you a chance to take a look if you want

PR #6024, opened 2023-12-18
chore: replace deprecated function

yyzxw:
replace instr.FromInt to instr.FormInt32, FromInt has been deprecated

### DIFF
--- a/apis/projectcontour/v1alpha1/contourdeployment.go
+++ b/apis/projectcontour/v1alpha1/contourdeployment.go
@@ -80,7 +80,7 @@ type ContourDeploymentSpec struct {
 type ContourSettings struct {
 	// Deprecated: Use `DeploymentSettings.Replicas` instead.
 	//
-	// Replicas is the desired number of Contour replicas. If if unset,
+	// Replicas is the desired number of Contour replicas. If unset,
 	// defaults to 2.
 	//
 	// if both `DeploymentSettings.Replicas` and this one is set, use `DeploymentSettings.Replicas`.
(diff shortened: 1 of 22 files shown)


internal/xds/v3/contour.go
```diff
@@ -81,7 +82,7 @@ func (s *contourServer) stream(st grpcStream) error {
 	done := func(log logrus.FieldLogger, err error) error {
 		// If the stream has been closed by the client "gracefully",
 		// do not log as an error.
-		if err != nil && err != context.Canceled && status.Code(err) != codes.Canceled {
+		if err != nil && !errors.Is(err, context.Canceled) && status.Code(err) != codes.Canceled {
```
davinci26, line 85: can we revert this change since this unrelated to the goal of this PR? I am worried that if there is something wrong here it would be hard to parse through the commit history

davinci26 (commented):
Thanks for the cleanup! Appreciate it

I am ok with both the `docs` change and the `replacement` in a single PR but I would defer to the maintainers to make the final call. 

If we decide to keep take the PR as is tho we should update the title and PR description such that it is clear what this PR does.

internal/xds/v3/contour.go
yyzxw: done

sunjayBhatia (changes requested):
I believe these comments from a previous iteration of this PR still apply to this one:
- https://github.com/projectcontour/contour/pull/5755#pullrequestreview-1631780001
- https://github.com/projectcontour/contour/pull/5755#issuecomment-1790986012

PR #6032, opened 2023-12-20
remove disabled ingress-conformance run from CI

skriss:
Upstream Ingress conformance is flaky/poorly maintained and
unlikely to reach a state where it can reliably be run in CI.

### DIFF
--- a/.github/workflows/prbuild.yaml
+++ b/.github/workflows/prbuild.yaml
@@ -241,7 +241,8 @@ jobs:
           steps: ${{ toJson(steps) }}
           channel: '#contour-ci-notifications'
         if: ${{ failure() && github.ref == 'refs/heads/main' }}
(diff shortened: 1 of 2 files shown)


sunjayBhatia (approved):
LGTM

We might just remove the ingress conformance tests from the pipelines too tbh and just leave the makefile task to run them manually

no-one really reviews changes to the repo and since theyre so flaky (no retries/timeouts configurable on http requests) i dont think anyone is actually running them in CI

skriss:
> LGTM
> 
> We might just remove the ingress conformance tests from the pipelines too tbh and just leave the makefile task to run them manually
> 
> no-one really reviews changes to the repo and since theyre so flaky (no retries/timeouts configurable on http requests) i dont think anyone is actually running them in CI

SGTM, let me make that change here

PR #6033, opened 2023-12-20
Promote upstream TLS changelog entry from minor to major

tsaarni:
After #5828 was merged (Thank You @KauzClay!) it occurred to me that we could promote the change from `minor` to `major` category, to call out the maximum TLS protocol version update. All information was there already, but I think it could be good to highlight it as a major change, since users might observe some differences with their upstream services after upgrade, when TLS handshake selects TLSv1.3. In worst case users could even have some minor problems due to the gaps in Envoy's retry handling, described in [envoyproxy/envoy#9300](https://www.github.com/envoyproxy/envoy/issues/9300).

### DIFF
--- a/changelogs/unreleased/5828-KauzClay-major.md
+++ b/changelogs/unreleased/5828-KauzClay-major.md
@@ -1,0 +1,9 @@
+## Upstream TLS now supports TLS 1.3 and TLS parameters can be configured
+
+The default maximum TLS version for upstream connections is now 1.3, instead of the Envoy default of 1.2.
+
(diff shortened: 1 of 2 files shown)


tsaarni:
@KauzClay could you review this since I'm proposing change to your changelog entry? Thank you 😄

KauzClay:
@tsaarni cool, I see what you're saying about potential surprised when connections opt to 1.3.

Your changes make sense to me 👍

skriss (approved):

PR #6046, opened 2023-12-27
feat: add no response config

yyzxw:


sunjayBhatia:
Looks like this project is deprecated: https://github.com/probot/no-response

We already have a stale issue workflow, can you add some more info on motivation for this?

PR #6048, opened 2023-12-29
Add action for generating OpenSSF scorecard

sunjayBhatia:
Generates scorecard to ensure repo/project is passing open source best practices.

### DIFF
--- a/.github/workflows/openssf-scorecard.yaml
+++ b/.github/workflows/openssf-scorecard.yaml
@@ -1,0 +1,46 @@
+name: OpenSSF Scorecard
+on:
+  branch_protection_rule:
+  # Run weekly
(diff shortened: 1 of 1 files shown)


sunjayBhatia:
Taken as an example from https://github.com/envoyproxy/envoy/blob/main/.github/workflows/scorecard.yml and https://github.com/envoyproxy/go-control-plane/pull/845

sunjayBhatia:
Ran the scorecard check locally using instructions from here https://securityscorecards.dev/#using-the-cli and looks like we have some things to assess and improve

skriss (approved):
LGTM, will have to squint at details of the report to see what we can improve.

PR #6051, opened 2024-01-03
Use exact version number for ossf/scorecard-action

sunjayBhatia:


sunjayBhatia:
Merging with powers as this is just a action version fix

PR #6052, opened 2024-01-03
Pin github actions using shas

sunjayBhatia:
Dependabot should be able to update these and preserve the tag in the comment, as per: https://github.com/dependabot/dependabot-core/pull/5951

Workflows updated with: https://app.stepsecurity.io/secureworkflow


sunjayBhatia:
🤞🏽 this doesn't break any actions

skriss (approved):
LGTM, glad dependabot will be updating these for us

PR #6053, opened 2024-01-03
Bump go to 1.21.5

sunjayBhatia:


sunjayBhatia:
debugging macos test races which i cant repro locally with 1.21.5, unlikely that this fixes it but worth a bump

seems like the recent grpc-go bump might be related, see https://github.com/grpc/grpc-go/issues/6485 and https://github.com/grpc/grpc-go/pull/6489 I don't see any failures before bumping to grpc-go 1.60.0

skriss (approved):

PR #6058, opened 2024-01-06
feat: add max-connections-per-listener config option

### LINKED ISSUE #5654: Make downstream Listener connection limit configurable

This is often useful for large installations in order to manage resources, prevent DDOS, etc. and falls in line with Envoys edge proxy guidelines: https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/edge

This is an explicit follow on from https://github.com/projectcontour/contour/issues/4717 to make key edge best practices configurable

flawedmatrix:
Adds a setting for `max-connections-per-listener` to limit the number of active connections to a listener.

fixes #5654


sunjayBhatia:
breadcrumb to think about when I get a chance to do a full review: think about admin/stats/etc. Listeners

sunjayBhatia:
> breadcrumb to think about when I get a chance to do a full review: think about admin/stats/etc. Listeners

so far my thought is to exclude those Listeners from this change since the shouldn't be at as much risk from a DOS attack (should not be exposed outside of the cluster, or even node), but we could do a follow on to set a reasonable default number of max connections for these listeners (or even make the limit configurable)

If we do want to ever set the global connection limit, that applies to all Listeners including the admin one, stats, etc., so we would have to make sure those don't get starved, and maybe set the "ignore global limit" flag for them

changelogs/unreleased/6058-flawedmatrix-minor.md
```diff
@@ -0,0 +1,3 @@
+## Allow setting connection limit per listener
+
+Adds a `max-connections-per-listener` config option to limit the number of active connections to a listener. The default, if unset, is unlimited.
```
skriss, line 3: It'd be nice to spell out the full YAML path of the config option for both the ContourConfiguration CRD (e.g. `spec.envoy.listener.maxConnectionsPerListener`) and the Contour config file (`listener.max-connections-per-listener`) here for clarity.

skriss (changes requested):
One little changelog nit, otherwise LGTM. I agree with @sunjayBhatia that it should be fine to not apply this setting to the admin/stats listeners.

flawedmatrix:
Yeah, that makes sense. If we ever need to set a listener connection limit on the admin and stats endpoint that can be a follow-on change.

skriss (approved):
LGTM, will leave for @sunjayBhatia to review as well. Thanks @flawedmatrix and @christianang!

sunjayBhatia (approved):

PR #6061, opened 2024-01-08
Add instructions for docker-mac-net-connect and installation instructions to contributing guide

### LINKED ISSUE #5269: add links/scripts for installing development tools
https://github.com/projectcontour/contour/blob/main/CONTRIBUTING.md#prerequisites lists the tools that are needed for Contour development. It'd be great to add links to install instructions for all the tools. Additionally, if possible, it'd be nice to have scripts for installing on common development environments (e.g. for macOS, most/all of the tools should be able to be installed via `brew install`).

### LINKED ISSUE #5294: document using docker-mac-net-connect when using kind on macOS
`make install-contour-working` and `make install-provisioner-working` both create a kind cluster that has metallb installed. This adds support for LoadBalancer services, so when the envoy service is deployed, it gets an external IP. This external IP can be directly accessed on Linux, but on macOS, it's not directly accessible from the host because docker runs in a VM.

I personally have used https://github.com/chipmk/docker-mac-net-connect with great success to solve this problem. When installed on the mac, it creates a wireguard tunnel which allows IPs from within the kind cluster to be accessed directly from the host. This means that I can directly curl the envoy service's external IP to send traffic to Envoy, just like I would if I was running Contour in a cloud provider cluster.

It would be great to test this out on a fresh install on macOS, and document it in https://github.com/projectcontour/contour/blob/main/CONTRIBUTING.md for macOS users.

christianang:
Docker Mac Net Connect allows you to connect directly to Docker-for-Mac containers using their ip address instead of binding ports.

This also adds more helpful instructions for installation of tools, either tells them the recommended way to install the tool or links to the installation instructions of the specific tool.

Fixes #5294
Fixes #5269

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,7 +19,35 @@ This section describes how to build Contour from source.
 
     The easiest way to experiment with Contour is to build it in a container and run it locally in [kind](https://kind.sigs.k8s.io/) cluster.
 
-2. Install tools: `git`, `make`, [`kind`](https://kind.sigs.k8s.io/docs/user/quick-start/), `kubectl`, `jq` and `yq`
+    On Mac, you can follow instructions to download and install [Docker Desktop](https://www.docker.com/products/docker-desktop/).
+
+    On Linux, you can follow one of the installation methods provided by [Docker](https://docs.docker.com/desktop/install/linux-install/).
+
+2. Install tools: `git`, `make`, `gcc` (used by `go test -race`), [`kind`](https://kind.sigs.k8s.io/docs/user/quick-start/), `kubectl`, `jq` and `yq`
+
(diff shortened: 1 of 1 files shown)


skriss (approved):
LGTM, should be helpful, thanks!

christianang:
@skriss decided to add to this PR for #5269. So could use another look. @flawedmatrix and I decided against adding a script to install tools. Our thinking is that every development environment might be a bit different and they may choose to install a tool differently. Providing instruction in the guide isn't too different from running a script so a user should still get most of the value compared to a script.

sunjayBhatia (approved):
nice!

skriss:
> hm havent seen this failure before: https://github.com/projectcontour/contour/actions/runs/7454768792/job/20306875474?pr=6061

Yeah, also never seen it, something to keep an eye on but clearly not introduced by this docs PR.

Issue #5845, opened 2023-10-12
Consider removing static "base" Runtime layer from bootstrap


sunjayBhatia:
At the moment we also use this helper from `internal/envoy/v3` to build up the bootstrap config base RTDS layer: https://github.com/projectcontour/contour/blob/f52ad069b1090aaabf8e05866af4ef4e1e4bdb31/internal/envoy/v3/bootstrap.go#L169

We may be able to get rid of that static bootstrap runtime layer since it was introduced in https://github.com/projectcontour/contour/pull/4379 and just serve dynamic layers instead (added in https://github.com/projectcontour/contour/pull/4387) which would mean we can refactor and consolidate a bit.

_Originally posted by @sunjayBhatia in https://github.com/projectcontour/contour/pull/5827#discussion_r1356959974_

Issue #6055, opened 2024-01-04
Featuretests test data race/panic due to Testing.Log called after test finishes


sunjayBhatia:
See test runs:
- https://github.com/projectcontour/contour/actions/runs/7403392541/job/20143161478
- https://github.com/projectcontour/contour/actions/runs/7400733328/job/20135033752
- https://github.com/projectcontour/contour/actions/runs/7400814171/job/20135279982
- https://github.com/projectcontour/contour/actions/runs/7251025980/job/19752564722
- https://github.com/projectcontour/contour/actions/runs/7425728318/job/20208024297
- https://github.com/projectcontour/contour/actions/runs/7449199586/job/20265372911
- https://github.com/projectcontour/contour/actions/runs/7478586959/job/20353923206

Tests have a data race/panic because `Testing.Logf` is called after the test ends

The offending call comes from here: https://github.com/projectcontour/contour/blob/8d63b86cb195fdd109ced0aad82922f74ebf0897/internal/xds/v3/contour.go#L88

We set up and wait for teardown of the Contour gRPC server here: https://github.com/projectcontour/contour/blob/8d63b86cb195fdd109ced0aad82922f74ebf0897/internal/featuretests/v3/featuretests.go#L162-L211

`GracefulStop()` and `Serve` *should* be waiting for all server handler goroutines to exit before returning, the test cleanup function should then be waiting for these to return before letting the test end, e.g. https://github.com/projectcontour/contour/blob/8d63b86cb195fdd109ced0aad82922f74ebf0897/internal/featuretests/v3/extensionservice_test.go#L427-L428

However as can be seen in the tests above, this is not strictly the case, it seems the server is allowed to exit before some of the handlers fully return. Repro by running `make check-test-race`, modified with `-count=50` (we can exacerbate this by adding a sleep to the `done` function in the contour xDS server).

This issue seems to have popped up post bumping grpc-go to 1.60.0+, I cannot repro it on 1.59.0

Ironically, this was an issue that was supposedly addressed in 1.60.0:
- https://github.com/grpc/grpc-go/issues/6485
- https://github.com/grpc/grpc-go/pull/6489

sunjayBhatia:
Minimal repro here: https://github.com/sunjayBhatia/grpc-go-issue-repro/tree/main/shutdown-cleanup

Tracked the offending commit down to https://github.com/grpc/grpc-go/pull/6489

sunjayBhatia:
[...]

PR #6063, opened 2024-01-10
Remove static base runtime layer from bootstrap

lubronzhan:
Right now dynamic runtime layer is available, they add the same config, so the static layer could be removed
Fix https://github.com/projectcontour/contour/issues/5845


sunjayBhatia:
Can ignore the test-osx failure, it is a race related to: https://github.com/projectcontour/contour/issues/6055

sunjayBhatia (approved):

PR #6068, opened 2024-01-10
Use minimum permissions for github token in PR workflows

sunjayBhatia:
To address OpenSSF security scan/best practices

Also does a little whitespace cleanup/standardizes YAML indenting

### DIFF
--- a/.github/workflows/welcome-new-contributors.yaml
+++ b/.github/workflows/welcome-new-contributors.yaml
@@ -3,12 +3,14 @@ name: 'Welcome New Contributors'
 on:
   issues:
     types: [opened]
+  # Workloads with pull_request_target and the GitHub Token secret should never include executing untrusted code
+  # See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target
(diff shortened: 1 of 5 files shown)


sunjayBhatia:
Doing this in pieces to make sure I don't break CI

.github/workflows/welcome-new-contributors.yaml
```diff
@@ -3,12 +3,14 @@ name: 'Welcome New Contributors'
...
+  # See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target
+  # And https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
   pull_request_target:
     types: [opened]
 
-# Workloads with pull_request_target and the GitHub Token secret should never include executing untrusted code
-# See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target
-# And https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
+permissions:
+  contents: read
```
skriss, line 13: I guess this gets overridden by the job-level config below that needs write access, but still a good idea to default to read-only and opt into read/write where needed.

skriss (approved):

